PR #1288, opened 2023-01-12
Bitcoind rpc url handling cleanup

dpc:

PR #1289, opened 2023-01-12
Pass the bitcoind rpc url via env variable

dpc:
On top of #1288

### DIFF
--- a/fedimint-api/src/bitcoin_rpc.rs
+++ b/fedimint-api/src/bitcoin_rpc.rs
@@ -1,1 +1,31 @@
+use std::ffi::OsStr;
 
+use url::Url;
+
+/// Name of the env value used for passing bitcoind rpc url to modules that need it
+pub const FM_BITCOIND_RPC_ENV: &str = "FM_BITCOIND_RPC";
+
+/// Default url that will be used if [`FM_BITCOIND_RPC_ENV`] is not set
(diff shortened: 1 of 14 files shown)


fedimint-api/src/bitcoin_rpc.rs
```diff
@@ -1 +1,31 @@
+use std::ffi::OsStr;
 
+use url::Url;
+
+/// Name of the env value used for passing bitcoind rpc url to modules that need it
+pub const FM_BITCOIND_RPC_ENV: &str = "FM_BITCOIND_RPC";
+
+/// Default url that will be used if [`FM_BITCOIND_RPC_ENV`] is not set
+pub const FM_BITCOIND_RPC_DEFAULT_FALLBACK: &str = "http://127.0.0.1:8332";
```
justinmoon, line 9: why default to mainnet?

justinmoon (approved):

fedimint-api/src/bitcoin_rpc.rs
justinmoon: i guess this is the default of bitcoin-cli ...

just seem like kinda arbitrary default.

dpc: Because that's the intended setup it's meant to be used with. Anyone not running it against mainnet, is a developer, not user, and they do an extra step, I think.

If anyone has good arguments against it, we can change. :)

justinmoon: Can't argue with that!

PR #1269, opened 2023-01-09
Split `ModuleKey` into `ModuleInstanceId` & `ModuleKind`

dpc:
### Update

So I followed through and I think I'm almost done, but I've got tired

Missing pieces:

* [ ] ~~`fedimint-dbdump` needs to load the config to know instande ids to create all the decoders; easy~~ I would prefer doing it in a follow-up, as I think we will need to switch to clap, share some constants and logic of loading config between fedimintd-dbump and fedimimntd. 

Some thoughts:

* Quite a bit of changes, but mostly mechanical.
* Client and tests code is all non-modularized and in quite few places I had to hardcode the 1:1 mapping between kind and instance id. We know the initialization order.
* In some places doing this effort actually made things cleaner&better.

### Previous description

So far we've been conflating the two, thinking that a federation will run each module type exactly once, using a `u16` value as a "globally unique ID".

After the modularization is complete, it became more clear that actually we want a seprate id for a "module type" (named here "kind" to avoid confussion with Rust "type"), and a separate id for each module federation decided to "instantiate".

Each fedimint binary (like `fedimintd` will register all available module kinds) early in `main` function.

```rust
fn main() {
    FedimintServer::builder()
        .datadir("/var/lib/fedimintd")
        .with_module::<WalletInit>()
        .with_module::<MintInit>()
        .with_module::<LightningInit>()
        .build()
        .run()
        .await
}
```

The the users will define module instances they actually want to run as a part of the federation (we don't need this part yet, and we can just assume each module kind is insantiated exactly once).

The DKG will generate federation config that will have module configs keyed by the `ModuleInstanceId`, and containing `kind: ModuleKind` for each instance, so like:

```
{
  // ...

  modules: {
    1: {
      kind: "mint",
      // ..
    },
    2: {
      kind: "wallet",
    }
  }
  // ...
}
```

On start, when config is initialized `fedimintd` will use the config to know for which id, to instantiate what module with what config.

In the future we might want to implement ability to change the composition of module instances. So the Federation can disable certain module, and replace it with a different one, etc.

It's undecided yet, are we going to use it for module version change coordination, or just to give ability to run same module multiple times with different configs (might be useful for 3rd party modules).

One way or another the architecture seems better if the id of the instance is separate from it's "type".

This change only does the minimum rename and is to gather approval for the idea. In follow-up changes we will have to actually replace certain parts of code-base that are using strings, where they should be using ids, etc.

PR #1290, opened 2023-01-12
Fix tmuxinator after instance-id change

dpc:
Broken in #1269

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -1,5 +1,7 @@
 # shellcheck shell=bash
 
+export LEGACY_HARDCODED_INSTANCE_ID_WALLET="2"
+
 function mine_blocks() {
     PEG_IN_ADDR="$($FM_BTC_CLIENT getnewaddress)"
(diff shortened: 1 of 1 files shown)


justinmoon (approved):

scripts/lib.sh
```diff
@@ -93,7 +95,7 @@ function gw_connect_fed() {
 }
 
 function get_finality_delay() {
-    cat $FM_CFG_DIR/client.json | jq -e -r '.modules.wallet.finality_delay'
```
elsirion, line 96: Why remove `-e`? This will likely lead to silent test failures again.

It seems this comes from the "merge master into this branch" commit even though I don't see it in the GH UI?! We need to be careful with these GH UI merge/rebase actions. Better to do this locally and pay attention when both master and feature branch had changes in the same spot.

okjodom (approved):

elsirion (commented):
Merged a bit too fast :grimacing:

scripts/lib.sh
justinmoon: ohh oops. I did this via the UI. Sorry about that.

PR #1294, opened 2023-01-12
Improve error handling for setup UI

### LINKED ISSUE #1183: distributedgen create-cert fails if `--out-dir` doesn't exist
```
$ rm -rf /var/lib/fedimint
$ distributedgen create-cert --out-dir /var/lib/fedimint/ --base-port 5000 --name alpha --password password  --announce-address <address>

thread 'main' panicked at 'write error: Os { code: 2, kind: NotFound, message: "No such file or directory" }', fedimintd/src/bin/distributedgen.rs:146:72
```

In this case I would expect it to create `/var/lib/fedimint` if it doesn't already exist

### LINKED ISSUE #1211: fedimintd/src/distributedgen.rs shouldn't panic
Perhaps it was OK when it was only a CLI, but now that it's a module that is also consumed by the admin UI we should be returning `Result`s.

jkitman:
No longer panic from DKG setup UI, but displays the error message, fixes #1211.

Also fixes #1183 by displaying the missing directory.

Would be cool to auto-populate more fields, esp for testing.  I'll try that in another PR.

### DIFF
--- a/fedimintd/src/bin/distributedgen.rs
+++ b/fedimintd/src/bin/distributedgen.rs
@@ -129,7 +129,7 @@ enum Command {
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
     tracing_subscriber::fmt()
         .with_env_filter(
             EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
@@ -153,8 +153,8 @@ async fn main() {
             name,
             password,
         } => {
-            let config_str = create_cert(dir_out_path, p2p_url, api_url, name, password);
-            println!("{}", config_str);
+            let config_str = create_cert(dir_out_path, p2p_url, api_url, name, password)?;
+            Ok(println!("{}", config_str))
         }
         Command::Run {
             dir_out_path,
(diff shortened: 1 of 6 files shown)


fedimintd/src/bin/distributedgen.rs
```diff
@@ -129,7 +129,7 @@ enum Command {
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
```
dpc, line 132: :clap:

dpc (approved):

fedimintd/src/distributedgen.rs
```diff
@@ -102,26 +102,26 @@ pub async fn run_dkg(
...
-pub fn parse_peer_params(url: String) -> PeerServerParams {
+pub fn parse_peer_params(url: String) -> anyhow::Result<PeerServerParams> {
     let split: Vec<&str> = url.split('@').collect();
-    assert_eq!(split.len(), 4, "Cannot parse cert string");
-    let p2p_url = split[0].parse().expect("could not parse URL");
-    let api_url = split[1].parse().expect("could not parse URL");
-    let hex_cert = hex::decode(split[3]).expect("cert was not hex encoded");
-    PeerServerParams {
+
+    ensure!(split.len() == 4, "Cert string has wrong number of fields");
```
justinmoon, line 115: didn't know about `ensure` ... that's nice

PR #1248, opened 2023-01-09
request payment of multiple invoices in a stream

okjodom:
Inspired by [this discussion](https://github.com/fedimint/fedimint/pull/1224#discussion_r1064215899). Allows lnrpc clients used by `gatewayd` to batch pay invoice operations and stream them to `pay_invoice`

PR #1224, opened 2023-01-06
Gateway cln extension service

okjodom:
Defining Fedimint CLN plugin (extension) to provide [GatewayLightning](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/proto/gatewaylnrpc.proto#L7:L19) rpc services.
A gateway lightning RPC client calls into this service to complete any desired lightning functionality. RPC interfaces exposed and implemented are:
- `get_pubkey` to get the public key of the lightning node backing this service
- `pay_invoice` to pay an invoice on behalf of the requesting client
- `subscribe_intercept_htlcs` to get a filtered stream of HTLCs for a federation

This extension is packaged as a binary named `gateway-cln-extension`

image

closes #1103

PR #1295, opened 2023-01-12
Gateway CLN extension

### LINKED ISSUE #1103: Gateway CLN Plugin Architecture
From @elsirion  cc @okjodom @justinmoon 

Re: gateway : I had a nice call with cdecker, he proposed to build the CLN integration as a very minimal plugin that can intercept HTLCs etc. and a separate daemon that speaks to it via some RPC interface. The plugin can decide what to do if the daemon isn't available (probably just tell CLN to proceed as if it wasn't intercepting them). This would give us multiple wins:

1. The safet-critical code that makes CLN not close channels would be a small RPC forwarder plugin
2. The daemon would be easy to restart without touching CLN
3. We wouldn't have to pass logs through CLN which tends to mangle them, attaching debuggers becomes easier
4. Dealing with CLN and LND would become much more similar

I'm not sure if you already came up with that independently, I think it was at least discussed at some point. But it seems like a solid plan, especially since it's coming from the author of the plugin system.

okjodom:
Defining Fedimint CLN plugin (extension) to provide [GatewayLightning](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/proto/gatewaylnrpc.proto#L7:L19) rpc services.
A gateway lightning RPC client calls into this service to complete any desired lightning functionality. RPC interfaces exposed and implemented are:
- `get_pubkey` to get the public key of the lightning node backing this service
- `pay_invoice` to pay an invoice on behalf of the requesting client
- `subscribe_intercept_htlcs` to get a filtered stream of HTLCs for a federation

This extension is packaged as a binary named `gateway-cln-extension`, and is currently **deployed on the second lightning node in the dev environment**

image

closes #1103


gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -166,7 +166,28 @@ impl GatewayLightning for ClnRpcService {
...
+            .call(cln_rpc::Request::Getinfo(
+                model::requests::GetinfoRequest {},
+            ))
+            .await
+            .map(|response| {
+                let pub_key = match response {
+                    cln_rpc::Response::Getinfo(model::responses::GetinfoResponse {
+                        id, ..
+                    }) => id,
+                    _ => panic!("Unexpected response from cln_rpc"),
```
elsirion, line 167: I'd prefer to avoid new `panic`s at this point when it isn't a logic error on our side. This one is borderline ok because we might not want to run the GW anyway if the CLN API schema changed. thoughts @dpc?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -190,11 +191,70 @@ impl GatewayLightning for ClnRpcService {
...
+
+        let mut outcomes: Vec<Result<PayInvoiceResponse, tonic::Status>> = Vec::new();
+
+        // Since the cln_rpc client is not `Sync`, we cannot actually lazily pocess the requests into response stream
+        // TODO: Fix clnrpc lifetime to muntithreading and lazy processing
+        while let Some(PayInvoiceRequest {
+            invoice,
+            max_delay,
+            max_fee_percent,
+        }) = stream.message().await?
```
elsirion, line 197: It's unclear to me why invoices to be paid have to come in a stream instead of sigle requests? Does it have to do with the client being in a mutex? If so, why is that? CLN is perfectly able to pay things in parallel.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -190,11 +191,70 @@ impl GatewayLightning for ClnRpcService {
             })
     }
 
+    // type PayInvoiceStream = tonic::Streaming<PayInvoiceResponse>;
```
elsirion, line 179: commented out code should be removed

gateway/ln-gateway/Cargo.toml
```diff
@@ -22,6 +22,7 @@ path = "src/bin/cln_extension.rs"
 
 [dependencies]
 anyhow = "1.0.66"
+async-stream = "0.3.3"
```
elsirion, line 25: not a fan of this additional dep, will it go away once the TODO further down is addressed?

https://github.com/fedimint/fedimint/pull/1295/commits/3d67279b989159f09e5add6812077e3b1f873ec0#diff-11314b558aa2437b3f9376371d29c660caf580b98f1e52fe9f7f548d7cc851d0R206-R207

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -295,24 +304,100 @@ pub enum ClnRpcError {
...
+        serde_json::json!({ "result": "continue" })
     }
 
-    async fn intercept_htlc(&self, _payload: HtlcAccepted) -> serde_json::Value {
-        unimplemented!("TODO: Implement HTLC filtering using short channel ids")
+    // Await outcome of HTLC processing by the gateway.
+    // If the gateway does not respond within the HTLC expiry `cltv_expiry_relative`,
+    // Automatically respond with a failure message.
+    async fn await_htlc_processing(
+        &self,
```
elsirion: Docstrings should begin with `///` so they are rendered on https://docs.fedimint.org

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -295,24 +304,100 @@ pub enum ClnRpcError {
...
+                    // Open a channel to receive the outcome of the HTLC processing
+                    let (sender, receiver) = oneshot::channel::<Action>();
+                    self.outcomes.lock().await.insert(short_channel_id, sender);
+
+                    // If the gateway does not respond within the HTLC expiry,
+                    // Automatically respond with a failure message.
+                    return tokio::time::timeout(
+                        Duration::from_secs(5),
+                        self.await_htlc_processing(receiver, payment_hash.clone()),
+                    )
```
elsirion: TODO: don't block here, HTLCs can be processed in parallel imo. Not urgent though.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -295,24 +304,100 @@ pub enum ClnRpcError {
...
+                    .await
+                    .unwrap_or_else(|e| {
+                        error!("await_htlc_processing error {:?}", e);
+
+                        // Bolt 4: https://github.com/lightning/bolts/blob/master/04-onion-routing.md#failure-messages
+                        // 2002 error code reports general temporary failure of this processing node.
+                        serde_json::json!({
+                            "result": "fail",
+                            "failure_message": "2002"
+                        })
```
elsirion: Can we make this a `const` somewhere?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -336,32 +338,33 @@ impl ClnHtlcInterceptor {
...
+                    let (sender, receiver) = oneshot::channel::<serde_json::Value>();
                     self.outcomes.lock().await.insert(short_channel_id, sender);
 
                     // If the gateway does not respond within the HTLC expiry,
                     // Automatically respond with a failure message.
-                    return tokio::time::timeout(
-                        Duration::from_secs(5),
-                        self.await_htlc_processing(receiver, payment_hash.clone()),
-                    )
+                    return tokio::time::timeout(Duration::from_secs(5), async {
```
elsirion, line 424: This should probably be more than 5s. It would be totally fine even if it is 1h in most cases. I'd go for 30s or so as a worst case. The GW needs to buy the preimage (1 epoch) and the federation needs to decrypt the preimage (1+ up to f epochs).

We should also think about what to do when shut down because CLN shuts down. Can we subscribe to the appropriate signals and cancel all HTLC in that case? Idk how CLN shuts down, best to speak to the CLN team.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -277,13 +278,100 @@ impl GatewayLightning for ClnRpcService {
...
+                .await
+                .outcomes
+                .lock()
+                .await
+                .remove(&hash)
+            {
+                // Translate action request into a cln rpc response for `htlc_accepted` event
+                let htlca_res = match action {
+                    Some(Action::Settle(Settle { preimage })) => {
+                        if let Ok(pk) = XOnlyPublicKey::from_slice(&preimage) {
```
elsirion, line 300: It's very annoying we have to deal with all that parsing boilerplate here. Is there a way to get protobuf structs parsed into properly typed structs using some sort of derive macro?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -277,13 +278,100 @@ impl GatewayLightning for ClnRpcService {
...
+            responses.push(Err(Status::internal(
+                "No reference found for this intercepted and processed htlc",
+            )))
+        }
+
+        let stream = async_stream::try_stream! {
+            for response in responses {
+                yield response?;
+            }
+        };
```
elsirion, line 352: I'm very sure there's a better way to get a stream from a `Vec`. E.g. `into_iter`+[`tokio_stream::iter`](https://docs.rs/tokio-stream/latest/tokio_stream/fn.iter.html)

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -27,31 +26,34 @@ use tokio_stream::wrappers::ReceiverStream;
 use tonic::{transport::Server, Status};
 use tracing::error;
 
+#[derive(Parser)]
+pub struct ClnExtensionOpts {
+    /// Gateway CLN extension service bind address
+    #[arg(long = "addr", env = "GW_CLN_EXTENSION_BIND_ADDRESS")]
+    pub addr: SocketAddr,
+}
```
elsirion, line 34: Optionally? Sure. But this will be horribly annoying to set up with e.g. nix-bitcoin where I can't simply inject env vars into the cln service afaik. Can we just make it a cln config option and maybe allow overriding via env? thoughts @dpc? I agree with the removal of the config file.

elsirion (commented):
I saw you addressed some of my earlier comments in later commits. Just ignore these.

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: Ack. Marking this for quick [follow ups](https://github.com/fedimint/fedimint/pull/1224#issuecomment-1380933799). We need to do several iterations on this base implementation

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: We'd have to define a default address in case a user doesn't care to supply cln config. Also, it might be a painful experience to always require a the cln plugin config even when env var is set

elsirion: We wouldn't require it, just make it optional by giving it a clearly invalid default. If neither is specified we can either fall back to a default or stop with an error message. We'll need some more config anyway to authenticate requests imo.

gateway/ln-gateway/Cargo.toml
okjodom: Dependency on `async_stream` can go away now if I find an easier way of producing Streams.
Addressing TODO opens up optimizations on processing, where we can send the same cln rpc instance to different threads to execute pay invoice requests

gateway/ln-gateway/src/bin/cln_extension.rs
dpc: Seems like this error  mean that our implementation is inherently incorrect? (wrong schema w.r.t. to node we're paired with? In which case this is excusable (or maybe even a good idea).

So... put a longer panic message and good enough?

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: We [defined the grpc spec](https://github.com/fedimint/fedimint/pull/1248) to allow the client to send a stream of invoices to be paid.
While unnecessary at this point, it allows for nice batching of requests by a client, i.e. the client can do `service.pay_invoice([invoice, invoice])` instead of multiple calls to `service.pay_invoice(invoice)`.

This spec only tells the service to assume/expect multiple invoices with every request.
- It's up to the service to optimize how it pays the invoices.
- It might consider parallel processing on every single request (stream from a single client) and across several requests (multiple streams, possibly from multiple clients) should be explored.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -0,0 +1,238 @@
...
+// Note: One this binary is stable, we should be able to remove current 'ln_gateway'
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let service = ClnRpcService::new()
+        .await
+        .expect("Failed to create cln rpc service");
+    let dir = service.dir.clone();
+
+    // Read configurations
+    let gw_cfg_path = dir.join("lnrpc.config");
```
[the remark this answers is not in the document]
justinmoon: What do we think about using the `.config` extension for a JSON file? IMO it would be better to just use the standard `.json` extension here and for `gateway.config`.

Also, do we need another config file? Since this just has a bind address can we pass it in via CLI or env?

gateway/ln-gateway/src/bin/cln_extension.rs
justinmoon: Curious here, too

okjodom: Two aspects of concern:

- Should the invoices to be paid come in a stream? : Yes - so a client can minimize it's requests (round trips) to the extension service. Sending a stream of invoices to be paid is the spec we want for that. See
`rpc PayInvoice(stream PayInvoiceRequest) returns (stream PayInvoiceResponse)`
- Can / should the extension service pay multiple invoices in parallel? Hopefully yes - Would be cool to parallel process pay invoice requests, but it's all up to the extension impl

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: Marked with a TODO.

I didn't yet look into how to make CLN process HTLCs in parallel. This is a simple migration from our `htlc_accepted` hook pattern, where we block while processing every HTLC intercepted

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: We need an iter `Stream`. Implements `poll_next`.

Instead of `async_stream` dependency, I ended up with a custom `impl Stream` for a `Vec` which is useful for now as we are not really parallel processing these requests.

okjodom: Generators might come in handy in future iterations

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -0,0 +1,238 @@
...
+use thiserror::Error;
+use tokio::{
+    io::{stdin, stdout},
+    sync::{mpsc, Mutex},
+};
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::{transport::Server, Status};
+use tracing::error;
+
+// Note: One this binary is stable, we should be able to remove current 'ln_gateway'
```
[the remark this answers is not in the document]
okjodom: oh I missed it again

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub cltv_expiry: u32,
+    pub cltv_expiry_relative: u32,
+    pub payment_hash: bitcoin_hashes::sha256::Hash,
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct Onion {
+    pub payload: String,
+    #[serde(rename = "type")]
+    pub type_: String,
```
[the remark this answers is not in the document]
okjodom: Updated. The doc doesn't mention back compat. I've synced the types in this PR. These type mappings on the new interceptor service are isolated. However, if we bump the lib version, we may be at risk with our ["in production" type mappings](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/cln.rs#L33:L60).

Following up there

justinmoon:
Review call notes:
- @elsirion concerned about streams. Adds complexity because we're doing this more asynchronously. And we won't be experiencing heavy load or batched requests any time soon. Doesn't have to change for this PR, but would be nice to reduce complexity here.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -279,29 +289,98 @@ pub enum ClnExtensionError {
...
         }
     }
 
-    async fn intercept_htlc(&self, _payload: HtlcAccepted) -> serde_json::Value {
-        unimplemented!("TODO: Implement HTLC filtering using short channel ids")
+    async fn intercept_htlc(&self, payload: HtlcAccepted) -> serde_json::Value {
+        let short_channel_id = payload.onion.short_channel_id;
+        let htlc_expiry = payload.htlc.cltv_expiry;
+
+        if let Some(subscription) = self.subscriptions.lock().await.get(&short_channel_id) {
```
elsirion, line 404: Can we hold this lock for a bit shorter by restructuring the code? Right now it will be held up to 30s and block every other intercepted HTLC. @dpc is that assessment correct? I'm not sure why the lock would stick around, but if it didn't I don't see how that code would work without a call to `Sender::clone()`.

How I see it:

```rust
self.subscriptions.lock() // impl Future<Output = MutexGuard<'_, HashMap<ShortChannelId, Sender>>>
    .await // MutexGuard<'_, HashMap<ShortChannelId, Sender>>
    .get(&short_channel_id) // &Sender
```

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -421,3 +504,15 @@ impl Stream for StreamProcessor<PayInvoiceResponse> {
         }
     }
 }
+
+impl Stream for StreamProcessor<CompleteHtlcsResponse> {
+    type Item = Result<CompleteHtlcsResponse, tonic::Status>;
```
elsirion, line 509: Is there any reason to not implement this generically: `impl<T> Stream for StreamProcessor<T>`? (ignoring if it's a good idea to do so)

elsirion (changes requested):
I think `StreamProcessor` has to go, too easy to misuse and fixing it isn't really worth it given that there's a good alternative.

Other comments were less severe iirc.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -250,3 +256,39 @@ impl ClnHtlcInterceptor {
...
+        self.responses.is_empty()
+    }
+}
+
+impl Stream for StreamProcessor<PayInvoiceResponse> {
+    type Item = Result<PayInvoiceResponse, tonic::Status>;
+
+    fn poll_next(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if self.is_empty() {
+            Poll::Ready(None)
```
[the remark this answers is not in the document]
okjodom: Life save. Haha! 🚀

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: [pending] will look into easier type mapping between grpc and our rust types

elsirion (commented):
First half

elsirion (approved):

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -0,0 +1,457 @@
...
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let service = ClnRpcService::new()
+        .await
+        .expect("Failed to create cln rpc service");
+
+    // Parse configurations
+    let bind_address = match ClnExtensionOpts::try_parse() {
+        Ok(opts) => opts.addr,
+        Err(_) => service.addr,
```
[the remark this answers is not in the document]
okjodom: ack.

PR #1299, opened 2023-01-12
Rename intercepted HTLC handler function

elsirion:
It's definitely not an incoming invoice that this fn processes. I'd like to call it `handle_intercepted_htlc`, but I see how that might not get the high level functionality across as well (and would require more renames of structs).


okjodom:
I did an initial rename a while back, but missed this function name. Your suggestion is definitely better. Ref #885

okjodom (approved):

PR #1300, opened 2023-01-12
Show `FM_BITCOIND_RPC` in a setup UI

dpc:
On top of #1289 

Screenshot from 2023-01-12 13-34-20

### DIFF
--- a/fedimintd/src/ui.rs
+++ b/fedimintd/src/ui.rs
@@ -214,10 +214,21 @@ async fn post_guardians(
 
 #[derive(Template)]
 #[template(path = "params.html")]
-struct UrlConnection {}
+struct UrlConnection {
+    ro_bitcoind_rpc: String,
+}
 
 async fn params_page(Extension(_state): Extension<MutableState>) -> UrlConnection {
-    UrlConnection {}
+    UrlConnection {
+        ro_bitcoind_rpc: fedimint_api::bitcoin_rpc::read_bitcoin_rpc_env_from_global_env()
+            .as_ref()
+            .map(ToString::to_string)
+            .ok()
+            .unwrap_or(format!(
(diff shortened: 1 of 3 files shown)


dpc:
I wanted (and still want to) to add some tooltips or something like that give some helpful infomation, without cluttering the UI, but bootstrap (or rather popper.js failed me: the tooltips were off to the center, and gave up on figuring out why after a short struggle).

dpc:
@justinmoon as it's mostly your code I'm messing with :)

justinmoon:
will review this evening

elsirion:
Re screenshot: p2p we don't speak websocket but a custom binary protocol over TLS, thus the schema should probably be `fedimint://…`

fedimintd/src/ui.rs
```diff
@@ -214,10 +214,21 @@ async fn post_guardians(
 
 #[derive(Template)]
 #[template(path = "params.html")]
-struct UrlConnection {}
+struct UrlConnection {
+    ro_bitcoind_rpc: String,
```
elsirion, line 218: Can't this be an `Url`? Iirc that implements `Display`.

elsirion (approved):
I love how the re-arranged form fields look. It's becoming really professional!

fedimintd/src/ui.rs
dpc: We don't always have `Url`, and it's `Option<Url>` and then askama's template seems too simplistic to work with it... 

Maybe there's a better way.

elsirion: You could have a `match` statement in askama iirc, but I agree that's probably not much better at that point (although you could hide other HTML elements like labels and headings depending on it).

Issue #1232, opened 2023-01-06
Cleanup naming in module types.


dpc:
* [ ] `ModuleInit` -> `ModuleConstructor` (or `ModuleGen`?)
* [ ] Find&replace all `ConfigGen` and `config_gen` to match.
* Consider `I` prefix in all module traits that we use in `Arc<dyn Trait>` and alikes. Just for consistency.
* Probably some other stuff. Go around and think about it.

justinmoon:
I find the name `ServerModulePlugin` kind of weird. `Plugin` and `module` in my mind mean roughly the same thing. So seeing both next to each other is odd.

dpc:
Yeap. This was a temporary meassure as we have:

```rust
struct Foo; // the dyn newtype wrapper
```

```
trait Foo; // type erased module trait
```

```
trait FooPlugin; // version of `Foo` with associated types that auto-impls to `Foo`
```

Better name suggestions appreciated.

dpc:
@elsirion Maybe `ModuleInit` should be named `ModuleKind` or `ModuleType`. That's kind of a design that is build in my head.

The binary app has to just list all the module "types" it supports (`ModuleKindRegistry`). Then the federation members build the actual list of module instances with its configuration, where each instance is of type from the supported kinds.

mxz42:
Not sure if it directly refers to this, but it's about naming, too:

I think as someone who wants to implement a module, it would "sound" and "feel right" to just write 
```rust
impl ServerModule for MyGreatNewServerModule
```

naming convention may look like:

```rust
trait IServerModule;  // interface
struct DynIServerModule<Arc<dyn IServerModule>>;  // dyn new type
trait ServerModule; //  trait with associated types implemented by modules
```

---

> I find the name ServerModulePlugin kind of weird. Plugin and module in my mind mean roughly the same thing. 

I agree. With less work one could just rename `ServerModulePlugin` to `ServerModuleTyped`, though, that would carry the "unnecessary" information `Typed` for an implementor.

elsirion:
@dpc Instead of a `ModuleKindRegistry` we could have a `FedimintdBuilder` and a `DistributedgenBuilder` to which these `ModuleKind` types could be mounted to respectively.

dpc:
@elsirion Ack on the builder, though I'm not sure about the composition of the `.run` and the UI and other misc code in the `main()` function now. We'll see what we can do.

@mxz42 

[...]

PR #1301, opened 2023-01-12
Rename symbols around modules

mxz42:
part of #1232

The `ModuleConsensusItem` still contains `Module` as there is the `ConsensusItem` enum as well. But I think one could work around that and just rename that enum either in the definition or when importing it (if `ModuleConsensusItem` should be renamed to `ConsensusItem`.


dpc (approved):

dpc:
LGTM, but ping @elsirion to make sure we're on the same page.

elsirion (approved):

PR #1305, opened 2023-01-13
Fix logging a lot of raw byte debug strings in certain places

dpc:
Without some kind of abbreviation reading through log outputs is going to be painful, and usually first bytes are the most significant ones.

### DIFF
--- a/fedimint-api/src/fmt_utils.rs
+++ b/fedimint-api/src/fmt_utils.rs
@@ -1,0 +1,16 @@
+use std::fmt;
+
+/// Use for displaying bytes in the logs
+pub struct AbbreviateHexBytes<'a>(pub &'a [u8]);
+
+impl<'a> fmt::Display for AbbreviateHexBytes<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.0.len() <= 64 {
+            f.write_str(&hex::encode(self.0))?;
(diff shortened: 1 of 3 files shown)


dpc:
@m1sterc001guy you might be one that is going to use it most often. :D

fedimint-api/src/fmt_utils.rs
```diff
@@ -0,0 +1,16 @@
...
+/// Use for displaying bytes in the logs
+pub struct AbbreviateHexBytes<'a>(pub &'a [u8]);
+
+impl<'a> fmt::Display for AbbreviateHexBytes<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.0.len() <= 64 {
+            f.write_str(&hex::encode(self.0))?;
+        } else {
+            f.write_str(&hex::encode(&self.0[..64]))?;
+            f.write_fmt(format_args!("-{}", self.0.len()))?;
```
elsirion, line 12: Future TODO: Would be cool to have an env var to deactivate this behavior.

elsirion (approved):

fedimint-api/src/fmt_utils.rs
dpc: Good idea. https://github.com/fedimint/fedimint/pull/1312

Issue #1303, opened 2023-01-12
`ModuleGen` should have a typed and erased versions, like all other module types.


dpc:
Currently it has only erased version and it's kind of weird, that the user needs to deal with conversion to erased types in this one place.

dpc:
I also need it to properly fix #1302

dpc:
@mxz42 It might be already on your radar?

dpc:
@mxz42 Right.

You don't mention a blanket `impl<T> IModuleGen for T where T: ModuleGen...` so mentioning it for completeness.

mxz42:
this can be closed? @dpc

PR #1314, opened 2023-01-13
Create interface for module generation trait

mxz42:
part of #1303

- creates an `IModuleGen` for `ModuleGen` and the blanket impl
- `ModuleGen` is now the typed trait with associated types

- [x] is the method `module_kind()` always supposed to be implemented? or default impl? see TODO comment  (edit: yes, default impl)

### DIFF
--- a/fedimint-api/src/module/mod.rs
+++ b/fedimint-api/src/module/mod.rs
@@ -267,7 +267,7 @@ pub trait ModuleGen: Debug + Sized {
 
     type Decoder: Decoder;
 
-    fn module_kind(&self) -> ModuleKind {
+    fn module_kind() -> ModuleKind {
         Self::KIND
     }
 
@@ -318,7 +318,7 @@ where
     }
 
     fn module_kind(&self) -> ModuleKind {
-        <Self as ModuleGen>::module_kind(self)
+        <Self as ModuleGen>::module_kind()
     }
 
     async fn init(
--- a/fedimint-api/src/module/mod.rs
+++ b/fedimint-api/src/module/mod.rs
@@ -5,10 +5,12 @@ pub mod registry;
 use std::collections::{BTreeMap, HashSet};
 use std::ffi::OsString;
 use std::fmt::Debug;
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use futures::future::BoxFuture;
 use secp256k1_zkp::XOnlyPublicKey;
+use serde_json::Value;
 use thiserror::Error;
 
 use crate::cancellable::Cancellable;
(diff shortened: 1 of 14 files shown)


fedimint-api/src/module/mod.rs
```diff
@@ -248,6 +250,132 @@ pub trait ModuleGen {
...
+#[async_trait]
+impl<T> IModuleGen for T
+where
+    T: ModuleGen + 'static + Sync,
+{
+    fn decoder(&self) -> DynDecoder {
+        DynDecoder::from_typed(ModuleGen::decoder(self))
+    }
+
+    fn module_kind(&self) -> ModuleKind {
```
dpc, line 320: same here

fedimint-api/src/module/mod.rs
```diff
@@ -248,6 +250,132 @@ pub trait ModuleGen {
...
+///
+/// For examples, take a look at one of the `MintConfigGenerator`, `WalletConfigGenerator`, or
+/// `LightningConfigGenerator` structs.
+#[async_trait]
+pub trait ModuleGen: Debug + Sized {
+    const KIND: ModuleKind;
+
+    type Decoder: Decoder;
+
+    fn module_kind(&self) -> ModuleKind {
```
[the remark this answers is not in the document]
mxz42: yep, true. I remove it.

fedimint-api/src/module/mod.rs
mxz42: I think this does though. otherwise `module_kind` wouldn't be object safe and I couldn't do `<Self as ModuleGen>`

mxz42:
Thanks for the review @dpc . I'm not sure if there are changes in the `<Module>ConfigGenerator` necessary, but I don't think so?

modules/fedimint-dummy/src/lib.rs
```diff
@@ -46,19 +46,25 @@ pub struct DummyOutputConfirmation;
...
+    type Decoder = DummyDecoder;
 
-    fn module_kind(&self) -> ModuleKind {
-        const KIND: &str = "dummy";
-        ModuleKind::from_static_str(KIND)
+    fn decoder(&self) -> DummyDecoder {
+        DummyDecoder
     }
 
+    // TODO: is this method always required to be implemented? Or maybe have default impl that
```
dpc: We should use the default. Or maybe even remove the default method and rellay on `Self::KIND` everywhere. Yes. That was my mistake. But maybe in another PR.

dpc (approved):

dpc:
@mxz42 Thanks for taking care of it! Please follow-up with deletion of these extra static default methods. The less is the better here, so that users don't get confused or make needless mistakes.

> Thanks for the review @dpc . I'm not sure if there are changes in the <Module>ConfigGenerator necessary, but I don't think so?

Do we still have these? I thought `ModuleGen` take care of all things config now. Are they leftovers?

dpc:
@mxz42 Oh. The `MintConfigGenerator` *is* the `MintGen` now. We should rename them, I guess and make it just `MintGen`. Similiarily `MintConfigGenParams` should become just `MintGenParams` maybe?

mxz42:
> @mxz42 Oh. The `MintConfigGenerator` _is_ the `MintGen` now. We should rename them, I guess and make it just `MintGen`. Similiarily `MintConfigGenParams` should become just `MintGenParams` maybe?

Sounds good to me :+1:

PR #1317, opened 2023-01-13
Remove `module_kind` method in typed `ModuleGen` trait

mxz42:
follow up on #1314 @dpc 

I think this is what you meant?


elsirion (approved):

dpc:
Yes. :+1: 

Having that default impl was nice for the caller, but it gives user extra place that they don't have to, but *can* make a mistake or get confused about.

PR #1320, opened 2023-01-13
Rename module generation structs for `ModuleGen` consistency

mxz42:
folluw up for #1314  @dpc , also ACK @elsirion ?

Renaming the `MintConfigGeneration` to `MintGen` for consistency with `ModuleGen`.
Same for `MintConfigGenerationParams` -> `MintGenParams`.

same applies to the other modules.

### DIFF
--- a/client/client-lib/src/mint/mod.rs
+++ b/client/client-lib/src/mint/mod.rs
@@ -669,9 +669,7 @@ mod tests {
     use fedimint_core::modules::ln::{ContractAccount, LightningGateway};
     use fedimint_core::modules::mint::config::MintClientConfig;
     use fedimint_core::modules::mint::db::ECashUserBackupSnapshot;
-    use fedimint_core::modules::mint::{
-        Mint, MintConfigGenParams, MintConfigGenerator, MintOutput,
-    };
+    use fedimint_core::modules::mint::{Mint, MintGen, MintGenParams, MintOutput};
     use fedimint_core::modules::wallet::PegOutFees;
     use fedimint_core::outcome::{SerdeOutputOutcome, TransactionStatus};
     use fedimint_core::transaction::legacy::Input;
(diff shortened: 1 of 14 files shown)


client/client-lib/src/mint/mod.rs
```diff
@@ -801,14 +799,14 @@ mod tests {
             FakeFed::<Mint>::new(
                 4,
                 |cfg, _db| async move { Ok(Mint::new(cfg.to_typed().unwrap())) },
-                &ConfigGenParams::new().attach(MintConfigGenParams {
+                &ConfigGenParams::new().attach(MintGenParams {
```
elsirion, line 802: These parameters are only used to generate config, not to generate the module instance itself. So keeping the `Config` part in the name might be reasonable. But I'd be ok with it for brevity.

elsirion (approved):
Unclear on the params, otherwise LGTM. Leaving the merge to @dpc

client/client-lib/src/mint/mod.rs
mxz42: good point. Might be worth leaving as is. There is also the trait that should be renamed if we rename the struct.
```rust
impl ModuleConfigGenParams for MintGenParams {  // previoulsy MintConfigGenParams
    const MODULE_NAME: &'static str = "mint";
}
```

So, I think it's usefull to keep `config` in the name then? @dpc

dpc: Generating the config in a way is generating the module instance, at least from the higher level PoV.

"The module (instance)" is uniquely defined by the `(type, id, config)` tuple. (there's env, but it's more like injecting external connectivity runtime settings, so can be ignored.

So I think it's fair to say that "when we generate config, we generate the module".

Plus it's shorter. :shrug:

dpc (approved):

client/client-lib/src/mint/mod.rs
mxz42: ok. 
then the trait `ModuleConfigGenParams` should be renamed to `ModuleGenParams`, so we have a consistent `impl ModuleGenParams for MintGenParams`

dpc: @mxz42 :+1: I leave it up to you.

dpc:
@mxz42 I've just noticed that mints `OutputOutcome` is called ... `OutputOutcome` which creates a confusion. Seems like for all other stuff we name them `LightningFoo`, so at least for consistency we should fix it as well.

PR #1321, opened 2023-01-13
`dbtool` to read, write and delete raw data from our DBs

elsirion:
This is the lower level alternative to `dbdump` to manually salvage DBs.

This is a draft so far since it's missing tests and I could not figure out how to make write operations persist so far.

### DIFF
--- a/fedimint-dbtool/src/main.rs
+++ b/fedimint-dbtool/src/main.rs
@@ -1,0 +1,73 @@
+use anyhow::Result;
+use bitcoin_hashes::hex::ToHex;
+use bytes::Bytes;
+use clap::{Parser, Subcommand};
+use fedimint_api::db::Database;
+use fedimint_api::module::registry::ModuleDecoderRegistry;
+
+#[derive(Debug, Clone, Parser)]
+struct Options {
+    database: String,
+    #[command(subcommand)]
+    command: DbCommand,
+}
+
+#[derive(Debug, Clone, Subcommand)]
+enum DbCommand {
+    List {
+        #[arg(value_parser = hex_parser)]
+        prefix: Bytes,
+    },
+    Write {
+        #[arg(value_parser = hex_parser)]
+        key: Bytes,
+        #[arg(value_parser = hex_parser)]
+        value: Bytes,
+    },
+    Delete {
+        #[arg(value_parser = hex_parser)]
+        prefix: Bytes,
+    },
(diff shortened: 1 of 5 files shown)


dpc (dismissed):

fedimint-dbtool/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    let rocksdb = fedimint_rocksdb::RocksDb::open(path)?;
+    Ok(Database::new(rocksdb, ModuleDecoderRegistry::default()))
+}
+
+fn print_kv(key: &[u8], value: &[u8]) {
+    println!("{} {}", key.to_hex(), value.to_hex());
+}
+
+#[tokio::main]
+async fn main() {
```
m1sterc001guy, line 48: Should we merge the functionality of this and dbdump so we only have 1 binary? I need to do some refactoring of dbdump anyway

fedimint-dbtool/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+            dbtx.raw_insert_bytes(&key, value.into())
+                .await
+                .expect("DB error");
+        }
+        DbCommand::Delete { prefix: key } => {
+            dbtx.raw_remove_entry(&key).await.expect("DB error");
+        }
+    }
+
+    dbtx.commit_tx().await.expect("DB error");
```
m1sterc001guy, line 72: Hmm I'm not seeing anything obvious on why this wouldn't be persisting values. If you use this tool with a new, empty db, does that work? Is commit crashing?

elsirion: Lol, the commit tracking is only implemented in the "normal", get/delete/… functions and not the `raw_*` versions.

That shouldn't be a problem, but apparently commits are only done if the commit tracker thinks something happened, which is an unnecessary and imo dangerous optimization. Fixed in 9a6f594c3db59379b8aaecf348a3e2f2317cf21a.

fedimint-dbtool/src/main.rs
elsirion: Eventually maybe when you have time. For now that works. I'd also like to discuss modularization of the DB tools. This one doesn't have to care yet fortunately xD

dpc (approved):

justinmoon:
This will be very neat to play around with

PR #1323, opened 2023-01-14
Change how the `IFederationApi` works

### LINKED ISSUE #1169: Change how `IFederationApi` works.
The `IFederationApi` doesn't scale well (we have to keep making `unimplemented!()` in every mock implementations), and I'm quite convinced it's not the right interface to put there. And it's up to the caller to know both the call meaning and the implementation (like types to be serialized, but also who to call and when exactly ("strategy")).

It seems to me that what we really need is:

```
pub trait IFederationApi {
    async fn request_member_raw<R>(
        &self,
        peer_id: PeerId,
        method: &str,
        params: &[serde_json::Value],
    ) -> Result<serde_json::Value>;

    async fn request_union(&self, method: &str, params: &[serde_json::Value]) -> Vec<serde_json::Value> {
       /* default impl using Self::request_member_raw and union strategy */
     }
    async fn request_current_consensus(&self, method: &str, params: &[serde_json::Value]) -> Vec<serde_json::Value> {
       /* default impl using Self::request_member_raw and current consensus strategy */
    }
    /* ... */
}
```

+/- some tweaks and helper functions to auto-convert stuff to/from `serde_json::Value`

The mocking will have to happen on string level (intercept whole `request_member_raw` and `match` `method` etc.)

This way picking the strategy, dealing with serialization types and any other logic is entirely on the caller of  `IFederationApi`, while the implementation takes care of delivering the requests, and possibly faking responses.

Feedback appreciated.

dpc:
List of follow-ups (before lands):

* [x] that `T: Deref<...>` is stupid there, isn't it?
* [x] fix wasm, probably painful

List of follow-ups (after lands):
* [ ] name errors (and results) `MemberError` and `FederationError`
* [ ] I made passing everything by value to dying in a dyn+async+lifetimes hell. Revisit after landing.
* [ ] move module Api traits to corresponding modules
* [ ] rename `IFederationApiFed` to `FederationApiExt`

Close  #1169

### DIFF
--- a/client/client-lib/src/api.rs
+++ b/client/client-lib/src/api.rs
@@ -1,10 +1,12 @@
-use std::cmp::min;
-use std::fmt::Debug;
+use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::{self, Debug};
+use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
+use std::{cmp, result};
 
 use async_trait::async_trait;
-use bitcoin::{Address, Amount};
+use bitcoin::Address;
 use bitcoin_hashes::sha256::Hash as Sha256Hash;
 use fedimint_api::config::ClientConfig;
 use fedimint_api::core::{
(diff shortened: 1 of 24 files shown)


dpc:
@elsirion @jkitman OK, so I think I have the core elements in place, so you should be able to judge the general idea. But I still need to redo dozen of api calls, and fix and cleanup bunch of stuff. So while I do the boring part, feel free to give feedback already.

client/client-lib/src/api.rs
```diff
@@ -1,171 +1,575 @@
...
+                        }
+                    };
+                    match strategy.process(PeerResponse { peer, result }) {
+                        QueryStep::Retry(peers) => {
+                            for retry_peer in peers {
+                                peer_errors.remove(&retry_peer);
+                                futures.push(Box::pin(async {
+                                    // Note: we need to sleep inside the retrying future,
+                                    // so that `futures` is being polled continously
+                                    sleep(Duration::from_millis(delay_ms)).await;
```
dpc, line 133: @jkitman @elsirion If my understanding is correct, this is actually pretty big footgun that we've overlooked in the existing code. If this `sleep` is outside of the future we spawn, then we are not doing `futures.next().await` which means other futures are not making progress.

dpc: I am also thinking about keeping a per-peer counter, so the exponential increase is per-peer.

client/client-lib/src/api.rs
```diff
@@ -1,171 +1,575 @@
...
+pub trait IFederationApi: Debug {
+    /// List of available federation members
+    fn members(&self) -> &BTreeSet<PeerId>;
+    /// Make request to a specific federation member by `peer_id`
+    async fn request_raw(&self, peer_id: PeerId, method: &str, params: &[Value]) -> Result<Value>;
+}
 
-    /// Submit a transaction to all federation members
-    async fn submit_transaction(&self, tx: LegacyTransaction) -> Result<TransactionId>;
+async fn with_erased_conversion<Params, Res, E, F>(params: &Params, f: F) -> result::Result<Res, E>
```
dpc, line 47: Note to self: we might only need with_erased_params

client/client-lib/src/api.rs
```diff
@@ -1,171 +1,575 @@
...
+use crate::LegacyTransaction;
 
 #[cfg_attr(target_family = "wasm", async_trait(? Send))]
 #[cfg_attr(not(target_family = "wasm"), async_trait)]
-pub trait IFederationApi: Debug + Send + Sync {
-    /// Fetch the outcome of an entire transaction
-    async fn fetch_tx_outcome(&self, tx: TransactionId) -> Result<TransactionStatus>;
+pub trait IFederationApi: Debug {
+    /// List of available federation members
+    fn members(&self) -> &BTreeSet<PeerId>;
```
elsirion, line 42: Unclear if this would be the ones we currently have a connection to or the ones we could open one to.

client/client-lib/src/api.rs
```diff
@@ -1,171 +1,575 @@
...
+    async fn request_raw(&self, peer_id: PeerId, method: &str, params: &[Value]) -> Result<Value>;
+}
 
-    /// Submit a transaction to all federation members
-    async fn submit_transaction(&self, tx: LegacyTransaction) -> Result<TransactionId>;
+async fn with_erased_conversion<Params, Res, E, F>(params: &Params, f: F) -> result::Result<Res, E>
+where
+    Params: Serialize + Send + Sync,
+    Res: DeserializeOwned,
+    F: FnOnce(Vec<Value>) -> Pin<Box<dyn Future<Output = result::Result<Value, E>> + Send>> + Send,
```
elsirion, line 51: What would the conversion be used for?

client/client-lib/src/api.rs
elsirion: Right, this needs to be a select, possibly even over all the other futures in parallel too?

elsirion (commented):
Didn't have time for a thorough review, but looks directionally right.

client/client-lib/src/api.rs
dpc: Oh, I thought it's already all and the underlying impl is responsible for maintaining connectivity. Will add a comment.

dpc:
Awesome. That's exactly what i need right now, before i draw the rest of the owl.

client/client-lib/src/api.rs
elsirion: I mean it kinda is responsible, but not always able to, so this fn could mean either.

dpc:
> Right, this needs to be a select, possibly even over all the other futures in parallel too?

I think the way I wrote it is fine now. The `futures.next().await` is effectively a "select" tha polls all futures in there towards completion. We just don't want to be awaiting on something else as it prevents everything in `futures` from making progres.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -216,18 +218,21 @@ pub async fn fixtures(num_peers: u16) -> anyhow::Result<Fixtures> {
 
             let user_db = if env::var("FM_CLIENT_SQLITE") == Ok(s) {
                 let db_name = format!("client-{}", rng().next_u64());
-                Database::new(sqlite(dir.clone(), db_name).await, module_decode_stubs())
+                Database::new(sqlite(dir.clone(), db_name).await, decoders.clone())
             } else {
-                Database::new(rocks(dir.clone()), module_decode_stubs())
+                Database::new(rocks(dir.clone()), decoders.clone())
```
dpc, line 223: I should have not went on the mission to eliminate the `module_decode_stubs` usage here. I didn't understand the scope of work required. But I guess no point in reverting now.

gateway/tests/tests/tests.rs
```diff
@@ -17,10 +17,17 @@ use ln_gateway::{
     utils::retry,
 };
 use mint_client::api::WsFederationConnect;
+use tracing_subscriber::EnvFilter;
 use url::Url;
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_gateway_authentication() -> Result<()> {
+    tracing_subscriber::fmt()
```
dpc, line 25: Had to do some debugging and decided to leave the debugging in.

client/client-lib/src/api.rs
dpc: `futures.await` is a select in parallel

I did the per-peer retry counters as well, BTW.

client/client-lib/src/api.rs
dpc: I can no longer view the code you anchored for some reason, so I have no idea which function was it.

client/client-lib/src/query.rs
```diff
@@ -253,40 +238,44 @@ impl<R: Eq + Clone + Debug> QueryStrategy<R> for CurrentConsensus<R> {
...
 
 /// Results from the strategy handling a response from a peer
 ///
-/// `Retry` sending requests to some of the peers
-/// `Continue` awaiting and handling responses
-/// `Finished` return a final result
+/// Note that the implementation driving the [`QueryStrategy`] returning
+/// [`QueryStep`] is responsible from remembering and collecting errors
+/// for each peer.
+#[derive(Debug)]
```
dpc, line 269: I didn't want to touch it, but change in the interface involved change in the way errors are handled, and that forced me to go over stuff here and tweak it a bit to fit.

client/client-lib/src/query.rs
```diff
@@ -9,16 +9,17 @@ use jsonrpsee_types::error::CallError as RpcCallError;
...
 /// Returns a result from the first responding peer
 pub struct TrustAllPeers;
 
 impl<R> QueryStrategy<R> for TrustAllPeers {
-    fn process(&mut self, response: FedResponse<R>) -> QueryStep<R> {
-        let FedResponse { peer: _, result } = response;
-        QueryStep::Finished(result.map_err(ApiError::RpcError))
+    fn process(&mut self, peer: PeerId, result: api::Result<R>) -> QueryStep<R> {
+        match result {
+            Ok(o) => QueryStep::Success(o),
```
dpc, line 20: @jkitman I changed this a little. I don't think we want to return a random failure, just because it was the first one to arrive. Also query strategy returns a `FedResponse` which is a collection of errors, not just one error. So we keep waiting until someone returns a success or everyone return a failure.

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+                JsonRpcError::Internal(_) => true,
+                JsonRpcError::Call(jsonrpsee_types::error::CallError::Custom(e)) => e.code() == 404,
+                _ => false,
+            },
+        }
+    }
+}
+
+#[derive(Debug, Error)]
+pub struct FedApiError(BTreeMap<PeerId, ApiError>);
```
dpc, line 90: Neat part: now that we have a different member vs federation api error, we will report responses from all peers if the call fail etc. Should help the debugging.

client/client-lib/src/api.rs
```diff
@@ -582,55 +1033,55 @@ fn url_to_string_with_default_port(url: &Url) -> String {
...
+    //                     sleep(Duration::from_millis(delay_ms)).await;
+    //                     delay_ms = min(max_delay_ms, delay_ms * 2);
+    //                 }
+    //                 QueryStep::Continue => {}
+    //                 QueryStep::Finished(result) => return result,
+    //             },
+    //             None => return Err(ApiError::NoResult),
+    //         }
+    //     }
+    // }
```
elsirion: You forgot this here?

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+pub type FedResult<T> = result::Result<T, FedApiError>;
+
+pub mod fake;
+
+// TODO: rename to `MemberApiError`
+#[derive(Debug, Error)]
+pub enum ApiError {
+    // TODO: this sticks out like a sore thumb
+    // #[error("Params serialization error: {0}")]
+    // ParamsSerialization(anyhow::Error),
```
elsirion, line 65: Just remove this?

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+        }
+        f.write_str(")")?;
+        Ok(())
+    }
+}
+
+impl FedApiError {
+    fn is_retryable(&self) -> bool {
+        self.0.iter().any(|(_, e)| e.is_retryable())
+    }
```
elsirion, line 109: This might have to be renamed eventually to `is_any_retryable` since it might make sense to ask if a certain threshold is retryable (if it isn't there might not be a chance for success even if the retryable ones return `Ok` eventually).

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+}
 
 #[cfg_attr(target_family = "wasm", async_trait(? Send))]
 #[cfg_attr(not(target_family = "wasm"), async_trait)]
-pub trait IFederationApi: Debug + Send + Sync {
-    /// Fetch the outcome of an entire transaction
-    async fn fetch_tx_outcome(&self, tx: TransactionId) -> Result<TransactionStatus>;
+pub trait IFederationApi: Debug {
+    /// List of all federation members for the purpuse
```
elsirion, line 143: typo, also which purpose?

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+    ) -> result::Result<Value, jsonrpsee_core::Error>;
+}
 
-    /// Submit a transaction to all federation members
-    async fn submit_transaction(&self, tx: LegacyTransaction) -> Result<TransactionId>;
+/// Run `f` with no `params`
+///
+/// ## Panics
+/// When `params` can't serialize to `serd_json::Value` (which should be never)
+#[cfg(target_family = "wasm")]
```
elsirion, line 164: I wonder if we can build a proc macro that can inject `+Send+Sync` in marked places in case we are not building for WASM. The code duplication is getting out of hand.

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+                                        PeerResponse {
+                                            peer: retry_peer,
+                                            result: self
+                                                .request_raw(retry_peer, method, params)
+                                                .await,
+                                        }
+                                    }
+                                }));
+                            }
+                        }
```
elsirion, line 358: This can potentially lead to multiple parallel requests to the same peer from different retries and could lead to an exponential number of parallel reqeuests. Maybe we should remove pending requests to the peers to which we start new ones?

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+
+    /// Submit a transaction for inclusion
+    async fn submit_transaction(&self, tx: LegacyTransaction) -> FedResult<TransactionId> {
+        with_erased_single_param(&SerdeTransaction::from(&tx.into_type_erased()), |params| {
+            Box::pin(async move {
+                self.request_current_consensus("/transaction".to_owned(), params)
+                    .await
+            })
+        })
+        .await
```
elsirion, line 489: Wouldn't it be easier to have a fn that generates params from serializable types instead of also making it call a closure?

client/client-lib/src/lib.rs
```diff
@@ -627,7 +647,7 @@ impl Client<UserClientConfig> {
         &self,
         node_pub_key: Option<secp256k1::PublicKey>,
     ) -> Result<LightningGateway> {
-        let gateways = self.fetch_registered_gateways().await?;
+        let gateways = dbg!(self.fetch_registered_gateways().await)?;
```
elsirion, line 650: Should use tracing

elsirion (approved):
Looks mostly good, some minor nits. Ideally @jkitman could take a look at the query strategy changes. @Maan2003 might also find this one interesting.

client/client-lib/src/api.rs
dpc: @elsirion I don't like the word `retryable` BTW. We need to be more specific. I don't think there's one "retryable" qualifier. It's up to the caller to decide what is worth retrying. We are probably thinking of "is potentially intermittent(?) / can go away on it's own?"

client/client-lib/src/api.rs
dpc: Was thinking the same thing. https://github.com/fedimint/fedimint/issues/1358

client/client-lib/src/api.rs
```diff
@@ -1,148 +1,615 @@
...
+pub trait IFederationApiFed: IFederationApi {
+    async fn request_fed_with_strategy<MemberRet: serde::de::DeserializeOwned, FedRet: Debug>(
         &self,
-        address: &Address,
-        amount: &Amount,
-    ) -> Result<Option<PegOutFees>>;
+        mut strategy: impl QueryStrategy<MemberRet, FedRet> + Send,
+        method: String,
+        params: Vec<Value>,
+    ) -> FedResult<FedRet> {
```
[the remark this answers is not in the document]
dpc: Hard to say. Its just an extension trait for `IFederationApi` ... so maybe it's OK? It will be a bit weird if they are named completely differently, I think.

client/client-lib/src/api.rs
dpc: That would only happen if the strategy returned `Retry` for the same peer, multiple times, before getting response from it first. Which is kind of a bug.

It would be better to remove the previous future, and it might even be useful for certain strategies, but I don't see an easy way to do https://docs.rs/futures/latest/futures/prelude/stream/struct.FuturesUnordered.html ATM.

client/client-lib/src/api.rs
dpc: I think you're right. I started with a function that converted params to Value, and then returned Value into the desired type. But then it turned out that strategies need a return value to be already deserialized so they can do smarter things on it... but I left the thing as a closure.

Will do.

client/client-lib/src/api.rs
elsirion: The reload code when you pushed the fixup might have messed it up, it was about the functions that took a closure and called it with a value. You changed them since then, so it's no longer relevant.

client/client-lib/src/api.rs
elsirion: As with all naming questions, I'm very open to better suggestions :sweat_smile:

client/client-lib/src/api.rs
dpc: Less relevant right now as these functions no longer take closures. Yay!

client/client-lib/src/api.rs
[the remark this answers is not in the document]
elsirion: I'm ok with it. Might call it `FederationApiExt`, but not a blocker.

client/client-lib/src/api.rs
elsirion: > Would have to write own `FutureUnordered`-like thing or look for something.

I don't think so. [`select_all`](https://docs.rs/futures/latest/futures/future/fn.select_all.html) should do exactly what we require here: wait for any future to complete and give back a vec of the remaining ones.

client/client-lib/src/api.rs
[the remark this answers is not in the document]
mxz42: if it's not object safe, I'd be in favor of `FederationApiExt`. It's still short and it's similar to `IFederationApi`

client/client-lib/src/api.rs
dpc: @elsirion It might be the primitive behind `FutureUnordered` but then we would need to track and remove them as well, also wrap each in some `FutureWithId` ... 

... it's doable, I'm sure. :)

client/client-lib/src/api.rs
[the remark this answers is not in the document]
mxz42: wait, why isn't `IFederationApiFed` object safe? @elsirion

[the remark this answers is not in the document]
dpc: @mxz42 `<MemberRet: serde::de::DeserializeOwned, FedRet: Debug>` I think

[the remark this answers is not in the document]
dpc: Will do a rename to `FederationApiExt` in post-PR, along with other renames.

elsirion (approved):

client/client-lib/src/api.rs
justinmoon: That's great

client/client-lib/src/query.rs
jkitman: My thinking was that if enough peers agree on the same error then we can trust that is the "correct" error, e.g. a 404 indicates a tx id was not found.  Although perhaps we should avoid treating errors as informative for the client.

PR #1324, opened 2023-01-14
Fix integration tests with real fixures

m1sterc001guy:
Running integration tests with real fixtures currently has two problems:

1. Environment variables are not passed to the wallet module, causing the RPCs to bitcoind to fail.
2. The task group running DKG is not shut down causing a port conflict when running the federation.

This PR fixes both by setting the regtest bitcoin connection address, passing the environment variables to the wallet module, and shutting down DKG so the federation can re-use the port.

### DIFF
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -4,6 +4,7 @@ use std::env;
 use std::future::Future;
 use std::iter::repeat;
 use std::net::SocketAddr;
+use std::os::unix::prelude::OsStrExt;
 use std::path::PathBuf;
 use std::pin::Pin;
 use std::rc::Rc;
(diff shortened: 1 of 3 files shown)


m1sterc001guy:
Hmmm some tests seem to be getting stalled during DKG, will need more investigation. At least some of the tests are passing now though.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -169,16 +170,21 @@ pub async fn fixtures(num_peers: u16) -> anyhow::Result<Fixtures> {
     match env::var("FM_TEST_DISABLE_MOCKS") {
         Ok(s) if s == "1" => {
             info!("Testing with REAL Bitcoin and Lightning services");
+            let mut config_task_group = TaskGroup::new();
```
justinmoon, line 173: I'm not an expert here, but I think if we already have a task group we should spawn new ones from it so in case we need to shut everything down we still can:

```
`config_task_group = task_group.make_subgroup()`
```

m1sterc001guy: Will fix

m1sterc001guy:
> Hmmm some tests seem to be getting stalled during DKG, will need more investigation. At least some of the tests are passing now though.

This is stalling because one of the federation members is choosing the same port as bitcoind. Will need to check that the ports do not collide to fix this.

dpc:
> This is stalling because one of the federation members is choosing the same port as bitcoind. Will need to check that the ports do not collide to fix this.

Maybe we should move the allocated ports to 20k-30k range?

m1sterc001guy:
> Maybe we should move the allocated ports to 20k-30k range?

This is the easiest fix. Changed the port range to be +20000 from the default port and now all tests are passing.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1026,6 +1032,12 @@ impl FederationTest {
             let mut override_modules = override_modules(cfg.clone(), db.clone()).await;
 
             let mut modules = BTreeMap::new();
+            // We currently have no way to enforce that modules are not reading
```
dpc, line 1035: I... recognize this code. If you want to re-use it, probably better to put original one in some method you can call here that returns that btreemap.

m1sterc001guy: :) Fixed

Issue #1307, opened 2023-01-13
Use `bitcoin_hashes::hex` instead of `hex`


dpc:
We trust and use `bitcoin-rust` code anyway, so by not using hex, we could get rid of one dependency.

And also `hex` doesn't seem to be able to write stuff out without allocating, while `bitcoin_hashes::hex` can https://docs.rs/bitcoin_hashes/latest/bitcoin_hashes/hex/fn.format_hex.html

mxz42:
I'd like to take a look at this

mxz42:
hex is still used here:

```sh
$ rg '^hex'
client/client-lib/Cargo.toml
24:hex = "0.4.3"

fedimint-server/Cargo.toml
23:hex = "0.4.2"

fedimintd/Cargo.toml
33:hex = "0.4.2"

fedimint-core/Cargo.toml
21:hex = "0.4.2"

fedimint-api/Cargo.toml
21:hex = { version = "0.4.3", features = [ "serde" ] }

modules/fedimint-wallet/Cargo.toml
22:hex = "0.4.3"

modules/fedimint-mint/Cargo.toml
21:hex = "0.4.2"

fedimint-dbdump/Cargo.toml
18:hex = { version = "0.4.3", features = [ "serde"] }

gateway/ln-gateway/Cargo.toml
29:hex = "0.4.3"
```

I'll go there next, I guess

mxz42:
PR #1472 removed all *direct* dependencies on the `hex` crate. 
@elsirion  had a good hint though: Next time we look at removing a crate, we could check first which transitive dependencies exist for that crate; meaning which other crates depend on that crate internally. 

in general this works with `cargo tree --invert hex`. currently we still depend transitively on `hex` through `cln-rpc` and `sqlx`

```
$ cargo tree --invert hex
hex v0.4.3
├── cln-rpc v0.1.2
│   └── ln-gateway v0.1.0 (/home/mxz/dev/mxz/fm/gateway/ln-gateway)
│       └── gateway-cli v0.1.0 (/home/mxz/dev/mxz/fm/gateway/cli)
│       [dev-dependencies]
│       ├── fedimint-tests v0.1.0 (/home/mxz/dev/mxz/fm/integrationtests)
│       └── gateway-tests v0.1.0 (/home/mxz/dev/mxz/fm/gateway/tests)
│   [dev-dependencies]
│   └── fedimint-tests v0.1.0 (/home/mxz/dev/mxz/fm/integrationtests)
[...]

PR #1325, opened 2023-01-14
Remove `hex` dependency from crypto crates

mxz42:
part of #1307 

I couldn't put this part into one line. Not exactly sure why.
```rust
// current state
let deser: String = Deserialize::deserialize(d)?;
Vec::<u8>::from_hex(&deser).map_err(serde::de::Error::custom)?

// does compile but tests fail:
Vec::<u8>::from_hex(Deserialize::deserialize(d)?).map_err(serde::de::Error::custom)?
```

example test failure message: 
```
---- set_lightning_invoice_expiry stdout ----
thread 'set_lightning_invoice_expiry' panicked at 'configuration mismatch: invalid type: string "a7d06980c1c68997f040d35e2abc52d1a30eb1dd760c607ca53d7089fffcc55ff737b7e7c641b1c5c226d20023290a0d0c6cc12c95da3c5f9c19480ae2f079cf483d71f55c0fbb386659df645482840468927fed439226d51ab7a8fd2d882445", expected a borrowed string', fedimint-server/src/config.rs:180:14
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1104,7 +1104,6 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "fedimint-api",
- "hex",
  "hkdf",
  "ring",
  "secp256k1-zkp",
@@ -3952,11 +3951,11 @@ name = "tbs"
 version = "0.1.0"
 dependencies = [
  "bincode",
+ "bitcoin_hashes 0.11.0",
  "bls12_381",
  "clap",
  "ff",
  "group",
- "hex",
  "rand",
  "rand_chacha",
  "serde",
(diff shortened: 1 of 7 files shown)


elsirion (approved):

elsirion:
You could do it in one line afaik, but you'd nee to do

```rust
Vec::<u8>::from_hex(&<String as Deserialize>::deserialize(d)?).map_err(serde::de::Error::custom)?
```

otherwise it would try to deserialize into a `&str`, which doesn't always work for JSON (e.g. in case of escape characters where the underlying byte sequence needs to be transformed).

mxz42:
ah, I see. yes, that works. thanks!

Issue #1318, opened 2023-01-13
Crash recovery, rejoining and banned peers


elsirion:
In a recent deployment #1316 lead to a panic of one of the guardians.

The logs indicate that at least the network stack kept running since otherwise `systemd` typically prints a restart message. Secondly, it appears our peers banned us (not that this is a 3-of-3 federation for *#reasons* so only two peers banned us, but that should not matter). 

We need to:
1. Ensure the federation shuts down completely so that it can attempt to re-join in such a case
2. Figure out why that peer got banned. It seems like the epoch signature the previously failed peer contributed was for an earlier epoch.
3. Unban when peers wen a node restarts itself, so that a restart of an entire federation can quick-fix certain bugs (this would still require coordination, so would not hide the problem)

Furthermore, I'd like to simulate a failure of this kind where an epoch is lost and see if our current rejoin code could handle it gracefully.

## Logs

<details>
  <summary>Log of the failed guardian (charlie)</summary>

```
Jan 12 23:20:14 charlie fedimintd[32185]: - Epoch: 510 [PeerId(0), PeerId(1), PeerId(2)] -
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 0: Transaction
Jan 12 23:20:14 charlie fedimintd[32185]:     Input: Lightning Contract f3dd1a942454997378c29814088d167470a45811ac636457a2f0ccd4ebf63730 with amount 40000000 msat
Jan 12 23:20:14 charlie fedimintd[32185]:     Output: Mint Coins 40000000 msat
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 0: Module CI: module=0 ci=Wallet Block Height 771650
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 0: Outcome Signature
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 1: Module CI: module=0 ci=Wallet Block Height 771650
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 1: Outcome Signature
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 2: Module CI: module=0 ci=Wallet Block Height 771650
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 2: Outcome Signature
Jan 12 23:20:14 charlie fedimintd[32185]: thread 'main' panicked at 'DB Error: Resource busy: ', /build/source/fedimint-server/src/consensus/mod.rs:302:36
[...]

PR #1327, opened 2023-01-14
DO NOT MERGE: Simulate DB failure during tx processsing

jkitman:
Simulates the failure in #1318.

One of the peers only partially updates the DB, leading to them failing to submit blind sig shares and getting banned by peers.


justinmoon:
Should this stay open?

jkitman:
No we can close it, unfortunately I wasn't able to reproduce although we fixed some related issues.

PR #1328, opened 2023-01-14
Don't create zero amount change ouputs

### LINKED ISSUE #1319: Don't attach e-cash outputs to LN offer creation transactions
```
Jan 13 19:37:15 alpha fedimintd[42194]:   Peer 2: Transaction
Jan 13 19:37:15 alpha fedimintd[42194]:     Output: LN offer for 1000 msat with hash be7b2e8660b5ca7975ebe1805060ee67f2df28237ca402f39a2b1553273d3c7a
Jan 13 19:37:15 alpha fedimintd[42194]:     Output: Mint Coins 0 msat
```

justinmoon:
Closes #1319

When creating lightning offers, we request zero change, which adds a zero amount output.

### DIFF
--- a/client/client-lib/src/mint/mod.rs
+++ b/client/client-lib/src/mint/mod.rs
@@ -207,7 +207,10 @@ impl MintClient {
 
         let mut change_outputs: Vec<(usize, NoteIssuanceRequests)> = vec![];
         let notes_per_denomination = self.notes_per_denomination(&mut dbtx).await;
-        for amount in change {
+        for amount in change.clone() {
+            if amount.msats == 0 {
+                continue;
+            }
             let (issuances, nonces) = self
                 .create_ecash(amount, notes_per_denomination, &mut dbtx)
                 .await;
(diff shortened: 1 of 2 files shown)


justinmoon:
Any idea why CI isn't passing? I get same thing locally ...

dpc:
There's a test that relies on an empty thing your now ignoring as a new epoch trigger to make progress @justinmoon

elsirion:
I can't tell you the why, @jkitman might have an idea, but it seems processing the offer takes one less epoch, so you have to reduce the 2 to a 1 here: 

https://github.com/fedimint/fedimint/blob/13e9b2a3b2fe5f616fa43b573d0dab90f89ac33b/integrationtests/tests/tests.rs#L442

elsirion:
> There's a test that relies on an empty thing your now ignoring as a new epoch trigger to make progress @justinmoon

I think that was actually more accidental. I get it now. Even an empty e-cash output would run one more consensus round (to exchange an empty signature share struct) before there was nothing left to do anymore. We don't need that extra epoch imo (at least in the case of `lightning_gateway_pays_internal_invoice` which I was just debugging).

justinmoon:
Waiting one less epoch fixed it. Thanks guys!

elsirion (approved):

Issue #1265, opened 2023-01-09
Partition Database so Modules can re-use key prefixes


m1sterc001guy:
We don't currently have a great way of iterating through all of the keys in the database (i.e SELECT * in SQL land). This is because different parts of the code have reserved key prefixes, so to iterate through all the keys we need to know about all of the DbKeyPrefix enums in different parts of the code. This is what dbdump does currently.

I want to iterate through all the keys so that I can implement the db migrations by copying all the key/value pairs to a new db (we can also support an in-place upgrade too, but making a new db is safer). 

One way to do this would be to refactor all of the prefixes into a giant enum and simply iterate through that. This is not great for modularization though since it requires all modules to reserve a key prefix space within the database.

A better solution is to use RocksDB's column families. We can partition the database for each module (each module can be a different column family). That means that key prefixes can then overlap between modules and we will still be able to read the key/value pairs properly from the database.

For the client, if its using SQLite instead of RocksDB, we can use different tables instead of Column Families.

dpc:
I was thinking that we should put modules data in a pre-prefix.

So we would hava a global `u8` enum for everything, where one thing would be:

```
#[repr(u8)]
enum Prefix {
  //...
  Module,
}
```

and then database would handle it somehow to prefix all operations from a module with `Prefix::Module` (1 bytes), `ModuleKey` (2 bytes) both on reads and writes.

This would isolate the modules from each other and put all the module data behind a common prefix.

m1sterc001guy:
I suppose the downside of using column families is that in order to scan all of the keys in the database, we now need to iterate over all of the modules, then iterate over all of the prefixes.

dpc:
> If we create a pre-prefix, are we concerned about performance at all? This essentially lumps all of the keys together in the same LSM tree,

My understanding is that this won't have any effect. The database just maintains a sorted list of all key-value pairs. It doesn't matter if the list is `a, b, c, d, e, f, g`, or `a, b, mc, md, me, mf, g`.  They sort the same, they will be split into segments all the same. The only difference is that the keys will be slightly longer.

m1sterc001guy:
> My understanding is that this won't have any effect. The database just maintains a sorted list of all key-value pairs. It doesn't matter if the list is `a, b, c, d, e, f, g`, or `a, b, mc, md, me, mf, g`. They sort the same, they will be split into segments all the same. The only difference is that the keys will be slightly longer.

[...]

PR #1331, opened 2023-01-15
Inject module_instance_id as prefix into database operations

m1sterc001guy:
This PR introduces a wrapper struct to wrap the DatabaseTransaction into an IsolatedDatabaseTransaction. IsolatedDatabaseTransaction will now always prepend a 2 byte "module_prefix", which is the ModuleInstanceId, when inserting and reading from the database. The purpose of this is to logically partition the database so modules can use keys that have conflicting prefixes. This also allows us to use the same key for different modules (ex: ModuleDatabaseVersion).

When passing a DatabaseTransaction to a module, it is first wrapped using with_module_prefix, which wraps the existing DatabaseTransaction inside an IsolatedDatabaseTransaction. This means they cannot overwrite keys from other modules. The fedimint server is now responsible for committing the wrapped DatabaseTransactions, because the individual modules cannot commit because they do not own the transaction. This approach also allows the fedimint server to create a DatabaseTransaction that spans over multiple modules and is also atomic.

[Original issue](https://github.com/fedimint/fedimint/issues/1265)

### DIFF
--- a/client/client-lib/src/ln/mod.rs
+++ b/client/client-lib/src/ln/mod.rs
@@ -362,7 +362,7 @@ mod tests {
             .await
             .members
             .iter()
-            .map(|(peer_id, _, _)| *peer_id)
+            .map(|(peer_id, _, _, _)| *peer_id)
             .collect();
         FederationApiFaker::new(fed, members)
             // TODO: is the output here is supposed to be a mint or wallet?
@@ -390,9 +390,15 @@ mod tests {
                     Ok(mint
                         .lock()
                         .await
-                        .fetch_from_all(|m, db| async {
-                            m.get_contract_account(&mut db.begin_transaction().await, contract)
-                                .await
+                        .fetch_from_all(|m, db, module_instance_id| async {
+                            m.get_contract_account(
+                                &mut db
+                                    .begin_transaction()
+                                    .await
+                                    .with_module_prefix(*module_instance_id),
+                                contract,
+                            )
+                            .await
                         })
                         .await
                         .unwrap())
--- a/fedimint-api/src/core/server.rs
+++ b/fedimint-api/src/core/server.rs
@@ -52,7 +52,11 @@ pub trait IServerModule: Debug {
     fn as_any(&self) -> &dyn Any;
 
     /// Blocks until a new `consensus_proposal` is available.
-    async fn await_consensus_proposal(&self, dbtx: &mut DatabaseTransaction<'_>);
+    async fn await_consensus_proposal(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        module_instance_id: ModuleInstanceId,
+    );
 
     /// This module's contribution to the next consensus proposal
     async fn consensus_proposal(
@@ -69,6 +73,7 @@ pub trait IServerModule: Debug {
         &self,
         dbtx: &mut DatabaseTransaction<'a>,
         consensus_items: Vec<(PeerId, DynModuleConsensusItem)>,
+        module_instance_id: ModuleInstanceId,
     );
 
     /// Some modules may have slow to verify inputs that would block transaction processing. If the
@@ -87,6 +92,7 @@ pub trait IServerModule: Debug {
         dbtx: &mut DatabaseTransaction<'_>,
         verification_cache: &DynVerificationCache,
         input: &DynInput,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<InputMeta, ModuleError>;
 
     /// Try to spend a transaction input. On success all necessary updates will be part of the
@@ -102,6 +108,7 @@ pub trait IServerModule: Debug {
         dbtx: &mut DatabaseTransaction<'c>,
         input: &'b DynInput,
         verification_cache: &DynVerificationCache,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<InputMeta, ModuleError>;
 
     /// Validate a transaction output before submitting it to the unconfirmed transaction pool. This
@@ -112,6 +119,7 @@ pub trait IServerModule: Debug {
         &self,
         dbtx: &mut DatabaseTransaction,
         output: &DynOutput,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<TransactionItemAmount, ModuleError>;
 
     /// Try to create an output (e.g. issue notes, peg-out BTC, …). On success all necessary updates
@@ -129,6 +137,7 @@ pub trait IServerModule: Debug {
         dbtx: &mut DatabaseTransaction<'a>,
         output: &DynOutput,
         out_point: OutPoint,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<TransactionItemAmount, ModuleError>;
 
     /// This function is called once all transactions have been processed and changes were written
@@ -140,6 +149,7 @@ pub trait IServerModule: Debug {
         &self,
         consensus_peers: &HashSet<PeerId>,
         dbtx: &mut DatabaseTransaction<'a>,
+        module_instance_id: ModuleInstanceId,
     ) -> Vec<PeerId>;
 
     /// Retrieve the current status of the output. Depending on the module this might contain data
@@ -156,7 +166,12 @@ pub trait IServerModule: Debug {
     ///
     /// Summing over all modules, if liabilities > assets then an error has occurred in the database
     /// and consensus should halt.
-    async fn audit(&self, dbtx: &mut DatabaseTransaction<'_>, audit: &mut Audit);
+    async fn audit(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        audit: &mut Audit,
+        module_instance_id: ModuleInstanceId,
+    );
 
     /// Returns a list of custom API endpoints defined by the module. These are made available both
     /// to users as well as to other modules. They thus should be deterministic, only dependant on
@@ -183,8 +198,16 @@ where
     }
 
     /// Blocks until a new `consensus_proposal` is available.
-    async fn await_consensus_proposal(&self, dbtx: &mut DatabaseTransaction<'_>) {
-        <Self as ServerModule>::await_consensus_proposal(self, dbtx).await
+    async fn await_consensus_proposal(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        module_instance_id: ModuleInstanceId,
+    ) {
+        <Self as ServerModule>::await_consensus_proposal(
+            self,
+            &mut dbtx.with_module_prefix(module_instance_id),
+        )
+        .await
     }
 
     /// This module's contribution to the next consensus proposal
@@ -193,11 +216,14 @@ where
         dbtx: &mut DatabaseTransaction<'_>,
         module_instance_id: ModuleInstanceId,
     ) -> Vec<DynModuleConsensusItem> {
-        <Self as ServerModule>::consensus_proposal(self, dbtx)
-            .await
-            .into_iter()
-            .map(|v| DynModuleConsensusItem::from_typed(module_instance_id, v))
-            .collect()
+        <Self as ServerModule>::consensus_proposal(
+            self,
+            &mut dbtx.with_module_prefix(module_instance_id),
+        )
+        .await
+        .into_iter()
+        .map(|v| DynModuleConsensusItem::from_typed(module_instance_id, v))
+        .collect()
     }
 
     /// This function is called once before transaction processing starts. All module consensus
@@ -208,10 +234,11 @@ where
         &self,
         dbtx: &mut DatabaseTransaction<'a>,
         consensus_items: Vec<(PeerId, DynModuleConsensusItem)>,
+        module_instance_id: ModuleInstanceId,
     ) {
         <Self as ServerModule>::begin_consensus_epoch(
             self,
-            dbtx,
+            &mut dbtx.with_module_prefix(module_instance_id),
             consensus_items
                 .into_iter()
                 .map(|(peer, item)| {
@@ -255,11 +282,12 @@ where
         dbtx: &mut DatabaseTransaction<'_>,
         verification_cache: &DynVerificationCache,
         input: &DynInput,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<InputMeta, ModuleError> {
         <Self as ServerModule>::validate_input(
             self,
             interconnect,
-            dbtx,
+            &mut dbtx.with_module_prefix(module_instance_id),
             verification_cache
                 .as_any()
                 .downcast_ref::<<Self as ServerModule>::VerificationCache>()
@@ -286,11 +314,12 @@ where
         dbtx: &mut DatabaseTransaction<'c>,
         input: &'b DynInput,
         verification_cache: &DynVerificationCache,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<InputMeta, ModuleError> {
         <Self as ServerModule>::apply_input(
             self,
             interconnect,
-            dbtx,
+            &mut dbtx.with_module_prefix(module_instance_id),
             input
                 .as_any()
                 .downcast_ref::<<Self as ServerModule>::Input>()
@@ -312,10 +341,11 @@ where
         &self,
         dbtx: &mut DatabaseTransaction,
         output: &DynOutput,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<TransactionItemAmount, ModuleError> {
         <Self as ServerModule>::validate_output(
             self,
-            dbtx,
+            &mut dbtx.with_module_prefix(module_instance_id),
             output
                 .as_any()
                 .downcast_ref::<<Self as ServerModule>::Output>()
@@ -339,10 +369,11 @@ where
         dbtx: &mut DatabaseTransaction<'a>,
         output: &DynOutput,
         out_point: OutPoint,
+        module_instance_id: ModuleInstanceId,
     ) -> Result<TransactionItemAmount, ModuleError> {
         <Self as ServerModule>::apply_output(
             self,
-            dbtx,
+            &mut dbtx.with_module_prefix(module_instance_id),
             output
                 .as_any()
                 .downcast_ref::<<Self as ServerModule>::Output>()
@@ -361,8 +392,14 @@ where
         &self,
         consensus_peers: &HashSet<PeerId>,
         dbtx: &mut DatabaseTransaction<'a>,
+        module_instance_id: ModuleInstanceId,
     ) -> Vec<PeerId> {
-        <Self as ServerModule>::end_consensus_epoch(self, consensus_peers, dbtx).await
+        <Self as ServerModule>::end_consensus_epoch(
+            self,
+            consensus_peers,
+            &mut dbtx.with_module_prefix(module_instance_id),
+        )
+        .await
     }
 
     /// Retrieve the current status of the output. Depending on the module this might contain data
@@ -374,17 +411,31 @@ where
         out_point: OutPoint,
         module_instance_id: ModuleInstanceId,
     ) -> Option<DynOutputOutcome> {
-        <Self as ServerModule>::output_status(self, dbtx, out_point)
-            .await
-            .map(|v| DynOutputOutcome::from_typed(module_instance_id, v))
+        <Self as ServerModule>::output_status(
+            self,
+            &mut dbtx.with_module_prefix(module_instance_id),
+            out_point,
+        )
+        .await
+        .map(|v| DynOutputOutcome::from_typed(module_instance_id, v))
     }
 
     /// Queries the database and returns all assets and liabilities of the module.
     ///
     /// Summing over all modules, if liabilities > assets then an error has occurred in the database
     /// and consensus should halt.
-    async fn audit(&self, dbtx: &mut DatabaseTransaction<'_>, audit: &mut Audit) {
-        <Self as ServerModule>::audit(self, dbtx, audit).await
+    async fn audit(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        audit: &mut Audit,
+        module_instance_id: ModuleInstanceId,
+    ) {
+        <Self as ServerModule>::audit(
+            self,
+            &mut dbtx.with_module_prefix(module_instance_id),
+            audit,
+        )
+        .await
     }
 
     fn api_endpoints(&self) -> Vec<ApiEndpoint<DynServerModule>> {
@@ -395,12 +446,13 @@ where
                 handler: Box::new(
                     move |module: &DynServerModule,
                           dbtx: fedimint_api::db::DatabaseTransaction<'_>,
-                          value: serde_json::Value| {
+                          value: serde_json::Value,
+                          module_instance_id: Option<ModuleInstanceId>| {
                         let typed_module = module
                             .as_any()
                             .downcast_ref::<T>()
                             .expect("the dispatcher should always call with the right module");
-                        Box::pin(handler(typed_module, dbtx, value))
+                        Box::pin(handler(typed_module, dbtx, value, module_instance_id))
                     },
                 ),
             })
--- a/fedimint-api/src/core/server.rs
+++ b/fedimint-api/src/core/server.rs
@@ -193,14 +193,11 @@ where
         dbtx: &mut DatabaseTransaction<'_>,
         module_instance_id: ModuleInstanceId,
     ) -> Vec<DynModuleConsensusItem> {
-        <Self as ServerModule>::consensus_proposal(
-            self,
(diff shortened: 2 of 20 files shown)


fedimint-api/src/db/mod.rs
```diff
@@ -169,6 +169,93 @@ impl Drop for CommitTracker {
...
 
+pub struct ModuleDatabaseTransaction<'a, 'b: 'a> {
+    tx: &'a mut DatabaseTransaction<'b>,
+    module_prefix: Vec<u8>,
+}
+
+impl<'a, 'b> ModuleDatabaseTransaction<'a, 'b> {
+    pub fn new(
+        dbtx: &'a mut DatabaseTransaction<'b>,
+        module_prefix: u16,
```
dpc, line 180: We should just take `ModuleInstanceId` I think.

fedimint-api/src/db/mod.rs
```diff
@@ -169,6 +169,93 @@ impl Drop for CommitTracker {
...
+    ) -> ModuleDatabaseTransaction<'a, 'b> {
+        ModuleDatabaseTransaction {
+            tx: dbtx,
+            module_prefix: vec![(module_prefix >> 8) as u8, (module_prefix & 0xff) as u8],
```
dpc, line 184: I thought the prefix is going to be `X <module_id>` where `X` would be some designed prefix for all module data. Otherwise we would have to be careful for the `module_prefix` not to conflict with any other global keys we have. Which is possible maybe if we deem that extra byte not worth it.

fedimint-api/src/db/mod.rs
```diff
@@ -214,7 +301,19 @@ impl<'a> DatabaseTransaction<'a> {
     where
         K: DatabaseKey + DatabaseKeyPrefixConst,
     {
-        let key_bytes = key.to_bytes();
+        self.module_get_value(key, Vec::new()).await
+    }
+
+    pub async fn module_get_value<K>(
```
dpc, line 307: I was hoping we don't have to introduce any new extra methods, and the `ModuleDatabaseTransaction` will just pass already concatendated key to the `DatabaseTransaction` so it doesn't even have to know about it.

In that case we can serialize the `K` straight into the `Vec<u8>` that already contains the prefix, to avoid copying everything twice (once serializing, then concatenating).

fedimint-api/src/db/mod.rs
```diff
@@ -232,17 +331,32 @@ impl<'a> DatabaseTransaction<'a> {
...
+        module_prefix.append(&mut key_prefix.to_bytes());
+        let prefix_bytes = &module_prefix;
         self.tx
             .raw_find_by_prefix(&prefix_bytes)
             .await
             .map(move |res| {
                 res.and_then(|(key_bytes, value_bytes)| {
-                    let key = KP::Key::from_bytes(&key_bytes, &decoders)?;
+                    let key = KP::Key::from_bytes(
+                        &key_bytes[(module_prefix.len() - 1)..].to_vec(),
```
dpc, line 357: This `.to_vec` seems needless, BTW.

fedimint-api/src/db/mod.rs
```diff
@@ -169,6 +169,93 @@ impl Drop for CommitTracker {
...
+    pub fn new_without_prefix(
+        dbtx: &'a mut DatabaseTransaction<'b>,
+    ) -> ModuleDatabaseTransaction<'a, 'b> {
+        ModuleDatabaseTransaction {
+            tx: dbtx,
+            module_prefix: vec![],
+        }
+    }
+
+    pub async fn get_value<K>(&mut self, key: &K) -> Result<Option<K::Value>>
```
dpc, line 197: We should not be implementing these functions, and only `raw` ones, and call into the `self.tx` only on raw ones as well.

dpc: So we prefix the prefix, call `self.tx.<the_same_raw_method>`, strip the prefix, and return back.

dpc: `raw_find_by_prefix` might be a bit of a problem, so we might need to build in suffix stripping right into `PrefixIter`, to avoid another layer of dynamic polymorphism, just to substitute it with a prefix-stripping implementation..

dpc:
Sems OK, but IMO, we need to do the prefix adding and removal on the lower level APIs. The higher APIs are just a convenience layer.

BTW. We can still pass to the modules the a `DatabaseTransaction`. It's just that the `DatabaseTransaction::tx` should now hold Box<dyn X>` where X is the `ModuleDatabaseTransaction`.

The way I see it:

```
struct ModuleDatabaseTransaction<'a> {
  tx: &'a dyn IDatabaseTransaction<'a>
  prefix: Vec<u8>
}

impl IDatabaseTransaction<'a> for ModuleDatabaseTransaction<'a> {
   /* impl all the raw methods by adding suffix to the key and removing if needed on the way back (like in PrefixIter) */
}
```

then maybe a helper:

```rust
impl<'a> DatabaseTranstion<'a> {
   fn with_module_prefix(self, module_id: Vec<u8>) -> &'a ModuleDatabaseTransaction<'a> {
      ModuleDatabaseTransaction::new(self, module_id_to_db_byte_prefix(module_id)))
   }
}
```

so that in the code we can quickly turn 
```rust
for (module_id, module) in self.modules {
   module.some_method(&dbtx.with_module_prefix(module_id), other_params).await?;
}
```

As you can see `tx: &'a dyn IDatabaseTransaction<'a>`, to avoid loosing the ownership of the original wrapped transaction, and get it right back after the call has ended.

There might be something I'm missing here, and async/await might get it the way, so please ping me if anything is unclear or you get stuck with some async/await + lifetimes problems.

client/client-lib/src/ln/mod.rs
```diff
@@ -399,9 +399,11 @@ mod tests {
                 .mint
                 .lock()
                 .await
-                .fetch_from_all(|m, db| async {
-                    m.get_contract_account(&mut db.begin_transaction().await, contract)
-                        .await
+                .fetch_from_all(|m, db, module_instance_id| async {
+                    let mut dbtx = db.begin_transaction().await;
+                    let mut module_dbtx =
+                        ModuleDatabaseTransaction::new(&mut dbtx, module_instance_id.clone());
```
elsirion, line 405: Can we make the API more like:

```rust
let mut dbtx = db.begin_transaction().await;
let mut module_dbtx = dbtx.module_tx(module_instance_id.clone());
```

fedimint-api/src/core/server.rs
```diff
@@ -395,12 +397,13 @@ where
                 handler: Box::new(
                     move |module: &DynServerModule,
                           dbtx: fedimint_api::db::DatabaseTransaction<'_>,
-                          value: serde_json::Value| {
+                          value: serde_json::Value,
+                          module_instance_id: Option<ModuleInstanceId>| {
```
elsirion, line 401: I don't think we need to add this arg here in `api_endpoints` and could just inject the appropriate module id into the DB tx in.

https://github.com/fedimint/fedimint/blob/2c75860db88e0903e40a4ebd6c9f263553435a09/fedimint-server/src/net/api.rs#L133-L180

fedimint-api/src/db/mod.rs
elsirion, line 184: Can we just use the `consensus_encode` impl for `module_prefix` here?

elsirion (commented):
Agree with @dpc and added some of my own nits

client/client-lib/src/ln/mod.rs
elsirion: Oh, I see dpc commented the same in the thread :laughing:

m1sterc001guy: Yeah I tried this initially and was running into lifetime problems. I'll try and do the refactoring @dpc suggested and ping if I'm still running into issues

fedimint-api/src/db/mod.rs
m1sterc001guy: I added MODULE_GLOBAL_PREFIX (0xff) which is prepended to all module prefixes. As long as the consensus key/values don't use 0xff as a prefix, the data should be separated from the modules.

fedimint-api/src/db/mod.rs
m1sterc001guy: This should be fixed now. I ended up doing the pre-pending and stripping of the prefix in raw_find_by_prefix of IsolatedDatabaseTransaction

fedimint-api/src/db/mod.rs
m1sterc001guy: Done

dpc: > As long as the consensus key/values don't use 0xff as a prefix

Don't we have like an enum or list of prefixes somewhere? We should reserve it just like any other one we use there.

m1sterc001guy: Ah yes, I can add it to the enum of prefixes

m1sterc001guy: > Ah yes, I can add it to the enum of prefixes

Actually, it looks like that enum is in fedimint_server, not fedimint_api where the IsolatedDatabaseTransaction impl is.

Maybe we should move the enum from fedimint_server/src/db.rs to fedimint_api?

dpc: Maybe we could nest them somehow?

dpc:
```

#[repr(u8)]
#[derive(Clone, Debug)]
pub enum DbKeyPrefix {
    ProposedTransaction = 0x01,
    AcceptedTransaction = 0x02,
    DropPeer = 0x03,
    RejectedTransaction = 0x04,
    EpochHistory = 0x05,
    Common(DbKeyPrefixCommon),
}


#[repr(u8)]
#[derive(Clone, Debug)]
pub enum DbKeyPrefixCommon {
    Module = 0xff
}
```

Haven't checked if `EnumIter` is going to complain, but I would throw it away if it does and impl whatever it does manually.

dpc: THough I haven't checked if this deserializes to u8...

dpc: Or just move to api, or define some as consts in api, and as enum in server

m1sterc001guy: Kept the constant in api but added a new entry in fedimint-server that references that constant.

fedimint-api/src/db/mod.rs
```diff
@@ -169,10 +175,86 @@ impl Drop for CommitTracker {
...
+        let mut prefix_with_module = self.prefix.clone();
+        prefix_with_module.append(&mut key_prefix.to_vec());
+        let raw_prefix = self
+            .tx
+            .raw_find_by_prefix(prefix_with_module.as_slice())
+            .await;
+        Box::new(raw_prefix.map(|pair| match pair {
+            Ok(kv) => {
+                let key = kv.0;
+                let stripped_key = &key[(self.prefix.len())..];
```
elsirion, line 233: observation: Assumes prefix-freeness of `ModuleInstanceId` encodings.

fedimint-api/src/db/mod.rs
```diff
@@ -169,10 +175,86 @@ impl Drop for CommitTracker {
...
+                Ok((stripped_key.to_vec(), kv.1))
+            }
+            _ => pair,
+        }))
+    }
+
+    async fn commit_tx(self: Box<Self>) -> Result<()> {
+        tracing::warn!("DatabaseTransaction inside modules cannot be committed");
+        Ok(())
+    }
```
elsirion, line 243: Wondering if warn or panic is better here …

fedimint-api/src/module/mod.rs
```diff
@@ -81,7 +81,7 @@ pub trait TypedApiEndpoint {
 
     async fn handle<'a, 'b>(
         state: &'a Self::State,
-        dbtx: fedimint_api::db::DatabaseTransaction<'b>,
+        dbtx: &'a mut fedimint_api::db::DatabaseTransaction<'b>,
```
elsirion, line 84: :heart_eyes: I disliked the owned DB tx here since forever but gave up the lifetime fight back then.

modules/fedimint-wallet/src/lib.rs
```diff
@@ -1536,7 +1536,13 @@ pub fn is_address_valid_for_network(address: &Address, network: Network) -> bool
 #[instrument(level = "debug", skip_all)]
 pub async fn run_broadcast_pending_tx(db: Database, rpc: DynBitcoindRpc, tg_handle: &TaskHandle) {
     while !tg_handle.is_shutting_down() {
-        broadcast_pending_tx(db.begin_transaction().await, &rpc).await;
+        broadcast_pending_tx(
+            db.begin_transaction()
+                .await
+                .with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_WALLET),
```
elsirion, line 1542: Needs to be fixed. I'm ok with a follow-up, but outside tests and the old client we should not use legacy ids.

elsirion (dismissed):

elsirion:
@dpc @jkitman do you want to take another look or can we go ahead and merge?

fedimint-api/src/core/server.rs
```diff
@@ -69,6 +73,7 @@ pub trait IServerModule: Debug {
         &self,
         dbtx: &mut DatabaseTransaction<'a>,
         consensus_items: Vec<(PeerId, DynModuleConsensusItem)>,
+        module_instance_id: ModuleInstanceId,
```
dpc, line 76: We should not inject module instance id to the modules. In principle they shouldn't need to know, and by not giving it to them we are makinig sure they are oblivious to it and can't do weird stuff with it.

fedimint-api/src/core/server.rs
```diff
@@ -183,8 +198,16 @@ where
     }
 
     /// Blocks until a new `consensus_proposal` is available.
-    async fn await_consensus_proposal(&self, dbtx: &mut DatabaseTransaction<'_>) {
```
dpc, line 186: This should already be prefixed here.

fedimint-api/src/core/server.rs
```diff
@@ -183,8 +198,16 @@ where
...
-        <Self as ServerModule>::await_consensus_proposal(self, dbtx).await
+    async fn await_consensus_proposal(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        module_instance_id: ModuleInstanceId,
+    ) {
+        <Self as ServerModule>::await_consensus_proposal(
+            self,
+            &mut dbtx.with_module_prefix(module_instance_id),
```
dpc, line 208: So we shouldn't need to change anything here.

fedimint-api/src/core/server.rs
```diff
@@ -255,11 +282,12 @@ where
         dbtx: &mut DatabaseTransaction<'_>,
         verification_cache: &DynVerificationCache,
         input: &DynInput,
+        module_instance_id: ModuleInstanceId,
```
dpc, line 285: Same here. The fedimint core code should do it.

fedimint-api/src/db/mod.rs
```diff
@@ -169,10 +175,86 @@ impl Drop for CommitTracker {
     }
 }
 
+struct IsolatedDatabaseTransaction<'isolated, 'parent: 'isolated> {
+    tx: &'isolated mut DatabaseTransaction<'parent>,
+    prefix: Vec<u8>,
+}
+
+impl<'isolated, 'parent: 'isolated> IsolatedDatabaseTransaction<'isolated, 'parent> {
+    pub fn new(
```
dpc, line 184: In the future we might want to inject prefixes for other stuff, so something more generic than `ModuleInstanceId` might be needed, but good enough for now, and simpler. Just mentioning it - this struct could be used to "isolated by prefix" things other than modules just fine.

fedimint-api/src/core/server.rs
elsirion: It's just the `IServerModule` trait though that no one will ever impl. It doesn't make a huge difference, I'm ok with either.

I think I proposed something along the lines of what got implemented but was thinking about carrying the instance id in the `DynServerModule` struct to avoid having to pass it around.

fedimint-api/src/db/mod.rs
```diff
@@ -169,10 +175,86 @@ impl Drop for CommitTracker {
     }
 }
 
+struct IsolatedDatabaseTransaction<'isolated, 'parent: 'isolated> {
+    tx: &'isolated mut DatabaseTransaction<'parent>,
```
dpc, line 179: No need to change it, but IIRC `inner` is preferable from repeating the same name.

fedimint-api/src/db/mod.rs
```diff
@@ -169,10 +175,86 @@ impl Drop for CommitTracker {
...
+    }
+}
+
+#[async_trait]
+impl<'isolated, 'parent> IDatabaseTransaction<'isolated>
+    for IsolatedDatabaseTransaction<'isolated, 'parent>
+{
+    async fn raw_insert_bytes(&mut self, key: &[u8], value: Vec<u8>) -> Result<Option<Vec<u8>>> {
+        let mut key_with_prefix = self.prefix.clone();
+        key_with_prefix.append(&mut key.to_vec());
```
dpc, line 205: Calling to `.to_vec()` just to append it to another vec seems like a waste (even if compiler can optimize). `.extend_from_slice`?

fedimint-api/src/db/mod.rs
dpc: How come? The `self.prefix` is the only thing we appended, and only thing we should remove. If we nest isolated classess each one will take care of its part of the prefix.

dpc: Had we change the defintion of `PrefixIter` to return `&[u8]` we might get rid of the `to_vec()` and since we are probably going to decode to the destination soon anyway, the higher level code would not care.

fedimint-api/src/db/mod.rs
dpc: We could make

```
trait FullDBTransactoin : BTransactoin {
   fn commit_tx(self) -> {   }
}
```

To make commits just not available .

But that would require another dyn newtype I think. Probably the most proper way to do it. But extra work, so could be done later.

dpc (commented):
Something tells me that it's best to avoid that instance_id in that `IServerModule` instance, even if just to keep the typed and untyped version the same.

I think it should be calling code's responsibility to do this. I don't think there will be many places doing it, and it's not something that can be easily overlooked.

m1sterc001guy:
> Something tells me that it's best to avoid that instance_id in that `IServerModule`, even if just to keep the typed and untyped version the same. It might come later and make us have to insert some dummy values where otherwise we wouldn't have to, or be a gateway to breaking the module-id obliviousness of the module implementation.
> 
> I think it should be calling code's responsibility to do this. I don't think there will be many places doing it, and it's not something that can be easily overlooked.
> 
> @elsirion What do you think?

I'll make the change, I agree it's not too difficult to do and the calling code should be responsible.

jkitman:
Easier and safer for module writers to not worry about db txs :+1:

fedimint-api/src/db/mod.rs
```diff
@@ -169,10 +175,86 @@ impl Drop for CommitTracker {
     }
 }
 
+struct IsolatedDatabaseTransaction<'isolated, 'parent: 'isolated> {
```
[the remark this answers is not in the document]
m1sterc001guy: Added some docs describing the purpose and implementation of IsolatedDatabaseTransaction

fedimint-api/src/db/mod.rs
m1sterc001guy: I changed it to a panic for now.

fedimint-api/src/core/server.rs
m1sterc001guy: Removed module_instance_id from server.rs where I could. fedimint-server now creates the wrapped DatabaseTransaction

fedimint-api/src/core/server.rs
```diff
@@ -374,9 +377,13 @@ where
         out_point: OutPoint,
         module_instance_id: ModuleInstanceId,
     ) -> Option<DynOutputOutcome> {
-        <Self as ServerModule>::output_status(self, dbtx, out_point)
-            .await
-            .map(|v| DynOutputOutcome::from_typed(module_instance_id, v))
+        <Self as ServerModule>::output_status(
+            self,
+            &mut dbtx.with_module_prefix(module_instance_id),
```
m1sterc001guy, line 382: One more to remove here

m1sterc001guy: Removed

fedimint-api/src/db/mod.rs
m1sterc001guy: Changed to use extend_from_slice

fedimint-api/src/db/mod.rs
m1sterc001guy: Changed to inner_tx

fedimint-api/src/db/mod.rs
m1sterc001guy: I took a swing at this. IsolatedDatabaseTransaction now accepts anything that is Send + Encodable.

fedimint-api/src/core/server.rs
m1sterc001guy: Removed

fedimint-api/src/core/server.rs
m1sterc001guy: This should be fixed. No longer wrapping the transaction in server.rs

modules/fedimint-wallet/src/lib.rs
m1sterc001guy: Hmm yes I think this is best as a follow up. I think the issue here is that, in order to not use the hardcoded constant, we'd need to pass the module_instance_id all the way down from ModuleGen::init. We can't create a database transaction during ModuleGen::init because this thread is polling the DB and needs to create a new transaction every time. Passing it down all the way through ModuleGen::init though means individual modules are now aware of their ID, which I think we're trying to avoid. Maybe this would also be a good candidate for the notification work @Maan2003 

@dpc is there an existing issue for removing the legacy harded instance ids or should I make a new one?

fedimint-api/src/db/mod.rs
m1sterc001guy: Is the concern with to_vec just that it is an extra copy? I tried changing this to &[u8] but ran into some lifetime issues, since the returned key/value pairs are temporary values with a lifetime of the existing function. I think I would need to use a Box here if we wanted to return &[u8]?

fedimint-api/src/db/mod.rs
```diff
@@ -169,10 +175,99 @@ impl Drop for CommitTracker {
     }
 }
 
+/// IsolatedDatabaseTransaction is a wrapper around DatabaseTransaction that is responsible for
```
elsirion, line 178: nit: using inline code formatting for type names would look nicer in rendered docs

fedimint-api/src/db/mod.rs
```diff
@@ -205,6 +300,26 @@ impl<'a> DatabaseTransaction<'a> {
...
+    ) -> DatabaseTransaction<'isolated>
+    where
+        'parent: 'isolated,
+    {
+        let decoders = self.decoders.clone();
+        let isolated = Box::new(IsolatedDatabaseTransaction::new(self, module_instance_id));
+        DatabaseTransaction {
+            tx: isolated,
+            decoders,
+            // DatabaseTransaction passed to modules cannot be committed, so the commit tracker is set to committed to surpress the warning
```
elsirion, line 315: The parent tx will take care of it though, right?

elsirion (approved):
LGTM

fedimint-api/src/db/mod.rs
m1sterc001guy: Yes, parent transaction will keep track of writes and the commit tracker. I can update the comment to reflect

elsirion:
Can we merge @dpc @jkitman ?

modules/fedimint-wallet/src/lib.rs
dpc: Wait, what? This is a module code already. Whatever this module is holding in that `db: Database` should be already isolated.

dpc: If we need to pass modules a whole db handle, we might need a similar wrapper for the whole `Database`

dpc: The module must never get it's unprivileged hands on anything remotely resembling unisolated access to the database. :D

dpc (approved):
I'm fine with landing already but we need to follow up with doing the same isolation technique for `db: Database` that we give that wallet.

m1sterc001guy:
> I'm fine with landing already but we need to follow up with doing the same isolation technique for db: Database that we give that wallet.

Ah yes, I'll address that next. Will rebase then merge

elsirion (approved):

PR #1337, opened 2023-01-15
Gatewayd

okjodom:
Gateway daemon that uses a [GatewayLightning](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/proto/gatewaylnrpc.proto#L7:L19) rpc client for lightning functions

- The production `gatewayd` instance uses a network gateway lightning rpc client which calls into a remote node to
  - get node pubkey for constructing invoices
  - pay invoice on behalf of user
  - subscribe to htlcs directed at a federation
  - process and respond to htlcs intercepted
- RPC calls are agnostic of the lightning node implementation

- #1295 is an example of gateway lightning extension this client can call into

image

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -38,6 +38,11 @@ windows:
           - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN2_DIR --addr=127.0.0.1:9001 --plugin=$FM_BIN_DIR/gateway-cln-extension &
           - echo $! >> $FM_PID_FILE
           - fg
+        - gateway:
+          - sleep 7 # wait for gateway-cln-extension to start
+          - $FM_BIN_DIR/gatewayd
+          - echo $! >> $FM_PID_FILE
+          - fg
         - federation:
           - sleep 1 # wait for bitcoind
           - ./scripts/start-fed.sh
--- a/gateway/ln-gateway/src/bin/gatewayd.rs
+++ b/gateway/ln-gateway/src/bin/gatewayd.rs
@@ -10,17 +10,17 @@ use fedimint_api::{
     module::{registry::ModuleDecoderRegistry, DynModuleGen},
     task::TaskGroup,
 };
-use fedimint_server::modules::{
-    ln::{common::LightningDecoder, LightningGen},
-    mint::{common::MintDecoder, MintGen},
-    wallet::{common::WalletDecoder, WalletGen},
-};
 use ln_gateway::{
     client::{DynGatewayClientBuilder, RocksDbFactory, StandardGatewayClientBuilder},
     gatewayd::{
         gateway::Gateway,
         lnrpc_client::{DynLnRpcClient, NetworkLnRpcClient},
     },
+};
+use mint_client::modules::{
+    ln::{common::LightningDecoder, LightningGen},
+    mint::{common::MintDecoder, MintGen},
+    wallet::{common::WalletDecoder, WalletGen},
 };
 use tracing::{error, info};
 use url::Url;
--- a/gateway/ln-gateway/src/bin/gatewayd.rs
+++ b/gateway/ln-gateway/src/bin/gatewayd.rs
@@ -1,0 +1,122 @@
+use std::{net::SocketAddr, path::PathBuf};
+
+use clap::Parser;
+use fedimint_api::{
+    config::ModuleGenRegistry,
+    core::{
+        LEGACY_HARDCODED_INSTANCE_ID_LN, LEGACY_HARDCODED_INSTANCE_ID_MINT,
+        LEGACY_HARDCODED_INSTANCE_ID_WALLET,
+    },
+    module::{registry::ModuleDecoderRegistry, DynModuleGen},
+    task::TaskGroup,
+};
+use fedimint_server::modules::{
+    ln::{common::LightningDecoder, LightningGen},
+    mint::{common::MintDecoder, MintGen},
+    wallet::{common::WalletDecoder, WalletGen},
+};
+use ln_gateway::{
+    client::{DynGatewayClientBuilder, RocksDbFactory, StandardGatewayClientBuilder},
+    gatewayd::{
+        gateway::Gateway,
+        lnrpc_client::{DynLnRpcClient, NetworkLnRpcClient},
+    },
+};
+use tracing::{error, info};
+use url::Url;
+
+#[derive(Parser)]
+pub struct GatewayOpts {
+    /// Path to folder containing gateway config and data files
+    #[arg(long = "data-dir", env = "FM_GATEWAY_DATA_DIR")]
+    pub data_dir: PathBuf,
+
+    /// Gateway webserver listen address
+    #[arg(long = "listen", env = "FM_GATEWAY_LISTEN_ADDR")]
+    pub listen: SocketAddr,
+
+    /// Public URL from which the webserver API is reachable
+    #[arg(long = "api-addr", env = "FM_GATEWAY_API_ADDR")]
+    pub api_addr: Url,
+
+    /// Gateway webserver authentication password
+    #[arg(long = "password", env = "FM_GATEWAY_PASSWORD")]
+    pub password: String,
+
+    /// Public URL to a Gateway Lightning rpc service
+    #[arg(long = "lnrpc-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
+    pub lnrpc_addr: Url,
+}
+
+// Fedimint Gateway Binary
+///
+/// This binary runs a webserver with an API that can be used by Fedimint clients to request routing of payments through the Lightning Network.
+/// It uses a `GatewayLightningClient`, an rpc client to communicate with a remote Lightning node accessible through a `GatewayLightningServer`.
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let mut args = std::env::args();
+
+    if let Some(ref arg) = args.nth(1) {
+        if arg.as_str() == "version-hash" {
+            println!("{}", env!("GIT_HASH"));
+            return Ok(());
+        }
+    }
+
+    // Read configurations
+    let GatewayOpts {
+        data_dir,
+        listen,
+        api_addr,
+        lnrpc_addr,
+        password,
+    } = GatewayOpts::parse();
+
+    info!(
+        "Starting gateway with these configs \n data directory: {:?},\n listen: {},\n api address: {},\n lnrpc address: {} ",
+        data_dir, listen, api_addr, lnrpc_addr
+    );
+
+    // Create federation client builder
+    let client_builder: DynGatewayClientBuilder =
+        StandardGatewayClientBuilder::new(data_dir.clone(), RocksDbFactory.into(), api_addr).into();
+
+    // Create task group for controlled shutdown of the gateway
+    let task_group = TaskGroup::new();
+
+    // Create a lightning rpc client
+    let lnrpc: DynLnRpcClient = NetworkLnRpcClient::new(lnrpc_addr).await?.into();
+
+    // Create module decoder registry
+    let decoders = ModuleDecoderRegistry::from_iter([
+        (LEGACY_HARDCODED_INSTANCE_ID_LN, LightningDecoder.into()),
+        (LEGACY_HARDCODED_INSTANCE_ID_MINT, MintDecoder.into()),
+        (LEGACY_HARDCODED_INSTANCE_ID_WALLET, WalletDecoder.into()),
+    ]);
+
+    // Create module generator registry
+    let module_gens = ModuleGenRegistry::from(vec![
+        DynModuleGen::from(WalletGen),
+        DynModuleGen::from(MintGen),
+        DynModuleGen::from(LightningGen),
+    ]);
+
+    // Create gateway instance
+    let gateway = Gateway::new(
+        lnrpc,
+        client_builder,
+        decoders,
+        module_gens,
+        task_group.clone(),
+    )
+    .await;
+
+    if let Err(e) = gateway.run(listen, password).await {
+        task_group.shutdown_join_all(None).await?;
+
+        error!("Gateway stopped with error: {}", e);
+        return Err(e.into());
+    }
+
+    Ok(())
+}
--- a/gateway/ln-gateway/src/gatewayd/actor.rs
+++ b/gateway/ln-gateway/src/gatewayd/actor.rs
@@ -14,7 +14,13 @@ use mint_client::{GatewayClient, PaymentParameters};
 use rand::{CryptoRng, RngCore};
 use tracing::{debug, info, instrument, warn};
 
-use crate::{ln::LnRpc, rpc::FederationInfo, utils::retry, LnGatewayError, Result};
+use crate::{
+    gatewayd::lnrpc_client::DynLnRpcClient,
+    gatewaylnrpc::{PayInvoiceRequest, PayInvoiceResponse},
+    rpc::FederationInfo,
+    utils::retry,
+    LnGatewayError, Result,
+};
 
 /// How long a gateway announcement stays valid
 const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
@@ -88,7 +94,7 @@ impl GatewayActor {
     #[instrument(skip_all, fields(%contract_id))]
     pub async fn pay_invoice(
         &self,
-        ln_rpc: Arc<dyn LnRpc>,
+        lnrpc: DynLnRpcClient,
         contract_id: ContractId,
     ) -> Result<OutPoint> {
         debug!("Fetching contract");
@@ -130,7 +136,7 @@ impl GatewayActor {
             self.buy_preimage_internal(&payment_params.payment_hash, &payment_params.invoice_amount)
                 .await
         } else {
-            self.buy_preimage_external(ln_rpc, contract_account.contract.invoice, &payment_params)
+            self.buy_preimage_external(lnrpc, contract_account.contract.invoice, &payment_params)
                 .await
         };
 
@@ -183,26 +189,23 @@ impl GatewayActor {
 
     pub async fn buy_preimage_external(
         &self,
-        ln_rpc: Arc<dyn LnRpc>,
+        lnrpc: DynLnRpcClient,
         invoice: lightning_invoice::Invoice,
         payment_params: &PaymentParameters,
     ) -> Result<Preimage> {
-        match ln_rpc
-            .pay(
-                invoice,
-                payment_params.max_delay,
-                payment_params.max_fee_percent(),
-            )
+        match lnrpc
+            .pay(PayInvoiceRequest {
+                invoice: invoice.to_string(),
+                max_delay: payment_params.max_delay,
+                max_fee_percent: payment_params.max_fee_percent(),
+            })
             .await
         {
-            Ok(preimage) => {
-                debug!(?preimage, "Successfully paid LN invoice");
-                Ok(preimage)
+            Ok(PayInvoiceResponse { preimage, .. }) => {
+                let slice: [u8; 32] = preimage.try_into().expect("Failed to parse preimage");
+                Ok(Preimage(slice))
             }
-            Err(e) => {
-                warn!("LN payment failed, aborting");
-                Err(LnGatewayError::CouldNotRoute(e))
-            }
+            Err(e) => Err(e),
         }
     }
 
--- a/gateway/ln-gateway/src/gatewayd/actor.rs
+++ b/gateway/ln-gateway/src/gatewayd/actor.rs
@@ -211,11 +211,7 @@ impl GatewayActor {
     }
 
     #[instrument(skip_all, fields(%contract_id))]
-    pub async fn pay_invoice(
-        &self,
-        lnrpc: DynLnRpcClient,
-        contract_id: ContractId,
-    ) -> Result<OutPoint> {
+    pub async fn pay_invoice(&self, contract_id: ContractId) -> Result<OutPoint> {
         debug!("Fetching contract");
         let rng = rand::rngs::OsRng;
         let contract_account = self.client.fetch_outgoing_contract(contract_id).await?;
@@ -255,7 +251,7 @@ impl GatewayActor {
             self.buy_preimage_internal(&payment_params.payment_hash, &payment_params.invoice_amount)
                 .await
         } else {
-            self.buy_preimage_external(lnrpc, contract_account.contract.invoice, &payment_params)
+            self.buy_preimage_external(contract_account.contract.invoice, &payment_params)
                 .await
         };
 
@@ -308,11 +304,11 @@ impl GatewayActor {
 
     pub async fn buy_preimage_external(
         &self,
-        lnrpc: DynLnRpcClient,
         invoice: lightning_invoice::Invoice,
         payment_params: &PaymentParameters,
     ) -> Result<Preimage> {
-        match lnrpc
+        match self
+            .lnrpc
             .pay(PayInvoiceRequest {
                 invoice: invoice.to_string(),
                 max_delay: payment_params.max_delay,
--- a/gateway/ln-gateway/src/gatewayd/actor.rs
+++ b/gateway/ln-gateway/src/gatewayd/actor.rs
@@ -1,7 +1,7 @@
 use std::{sync::Arc, time::Duration};
 
 use bitcoin::{Address, Transaction};
-use bitcoin_hashes::sha256;
+use bitcoin_hashes::{sha256, Hash};
 use fedimint_api::{task::TaskGroup, Amount, OutPoint, TransactionId};
 use mint_client::modules::{
     ln::{
@@ -12,11 +12,15 @@ use mint_client::modules::{
 };
 use mint_client::{GatewayClient, PaymentParameters};
 use rand::{CryptoRng, RngCore};
-use tracing::{debug, info, instrument, warn};
+use tracing::{debug, error, info, instrument, warn};
 
 use crate::{
     gatewayd::lnrpc_client::DynLnRpcClient,
-    gatewaylnrpc::{PayInvoiceRequest, PayInvoiceResponse},
+    gatewaylnrpc::{
+        complete_htlcs_request::{Action, Cancel, Settle},
+        CompleteHtlcsRequest, PayInvoiceRequest, PayInvoiceResponse,
+        SubscribeInterceptHtlcsRequest, SubscribeInterceptHtlcsResponse,
+    },
     rpc::FederationInfo,
     utils::retry,
     LnGatewayError, Result,
@@ -25,12 +29,20 @@ use crate::{
 /// How long a gateway announcement stays valid
 const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
 
+#[derive(Clone)]
 pub struct GatewayActor {
     client: Arc<GatewayClient>,
+    lnrpc: DynLnRpcClient,
+    task_group: TaskGroup,
 }
 
 impl GatewayActor {
-    pub async fn new(client: Arc<GatewayClient>, route_hints: Vec<RouteHint>) -> Result<Self> {
+    pub async fn new(
+        client: Arc<GatewayClient>,
+        lnrpc: DynLnRpcClient,
+        route_hints: Vec<RouteHint>,
+        task_group: TaskGroup,
+    ) -> Result<Self> {
         let register_client = client.clone();
         tokio::spawn(async move {
             loop {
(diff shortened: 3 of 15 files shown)


gateway/ln-gateway/src/gateway.rs
```diff
@@ -44,29 +50,43 @@ impl Gateway {
...
-        // TODO: consider encapsulating message channel within LnGateway
-        sender: mpsc::Sender<GatewayRequest>,
-        receiver: mpsc::Receiver<GatewayRequest>,
         task_group: TaskGroup,
     ) -> Self {
+        // Create message channels for the webserver
+        let (sender, receiver) = mpsc::channel::<GatewayRequest>(100);
+
+        // Start gateway without an lnrpc client
+        let lnrpc: Option<DynLnRpcClient> = None;
```
elsirion, line 63: Should be initialized here imo.

Also, if there is only one LN rpc client, then we actually don't need the factory. Allowing hot swaps of the underlying LN node is not a current goal imo. That would require a whole lot of complicated shutdown logic that we don't have and we should keep it as simple as possible till we actually want that feature.

gateway/ln-gateway/src/gateway.rs
```diff
@@ -44,29 +50,43 @@ impl Gateway {
     pub async fn new(
         config: GatewayConfig,
         decoders: ModuleDecoderRegistry,
-        ln_rpc: Arc<dyn LnRpc>,
+        lnrpc_factory: DynLnRpcClientFactory,
```
elsirion, line 55: This argument could just be a `DynLnRpcCLient`. We could still inject a mock client that way.

gateway/ln-gateway/src/actor_fork.rs
```diff
@@ -44,7 +55,112 @@ impl GatewayActor {
...
+                        payment_hash,
+                        outgoing_amount_msat,
+                        intercepted_htlc_id,
+                        ..
+                    }) = match stream.message().await {
+                        Ok(Some(msg)) => Some(msg),
+                        Ok(None) => {
+                            warn!("Stream closed");
+                            None
+                        }
```
elsirion, line 101: These nested loops don't make any sense. If the stream closes the outer `loop` will lead to an infinite loop using 100% CPU without re-establishing the stream. As long as the stream is active the outer `loop` is not needed.

gateway/ln-gateway/src/actor_fork.rs
```diff
@@ -44,7 +55,112 @@ impl GatewayActor {
...
+                        };
+
+                        htlc_outcomes.push(outcome);
+                    }
+
+                    if let Err(e) = lnrpc.complete_htlcs(htlc_outcomes).await {
+                        error!("Failed to complete HTLCs: {:?}", e);
+                        // NOTE: This is a potential loss of funds for the gateway.
+                        // We should consider a retry of this operation, cancel the HTLCs or reclaim funds.
+                    }
```
elsirion, line 157: This will only be hit once the stream closes, which also means that `complete_htlcs` may never be called and `htlc_outcomes` could grow indefinitely.

I already [mentioned this DoS vector in a previous iteration of the PR](https://github.com/fedimint/fedimint/pull/1198#discussion_r1069643531) and did not receive a rebuttal.

elsirion (changes requested):
Architecturally I only got some small nits.

The HTLC processing needs some clarification. Either I fundamentally misunderstand something about it or `tonic::Streaming` or the current implementation is broken.

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -0,0 +1,97 @@
...
+    gateway::Gateway,
+    rpc::lnrpc_client::{DynLnRpcClientFactory, NetworkLnRpcClientFactory},
+};
+use tracing::{error, info};
+
+#[derive(Parser)]
+pub struct GatewayOpts {
+    /// Path to folder containing gateway config and data files
+    #[arg(long = "cfg-dir", env = "GW_CONFIG_DIR")]
+    pub cfg_dir: PathBuf,
```
mxz42, line 30: maybe call this `data_dir` as well. makes it consistent with `fedimintd` [here](https://github.com/fedimint/fedimint/blob/master/fedimintd/src/bin/main.rs#L27)

gateway/ln-gateway/src/actor_fork.rs
okjodom: > This will only be hit once the stream closes, which also means that complete_htlcs may never be called and htlc_outcomes could grow indefinitely.
> I already mentioned this DoS vector in a previous iteration of the PR and did not receive a rebuttal.

I agree there is a high severity logical bug here. Seems like we need additional markers at which we flush the `htlc_outcomes`, and also guarantee flushing of any contents just before we quit and close our subscription

[pending]

gateway/ln-gateway/src/actor_fork.rs
okjodom: I was quite convinced we needed both loops. The outer one for keeping a persistent and long running subscription for htlcs. The inner While loop for reading htlc messages from the stream. Probably carried over these notions from working with plain mpsc channel subscriptions

[pending, will fix]

gateway/ln-gateway/src/actor_fork.rs
elsirion: I'd just call `complete_htlcs` after every successful buying of a preimage. We don't need batching for lack of expected load and additional latency might cost us actual money (30s timeout).

My proposal: get rid of outer loop, rename `complete_htlcs` to `complete_htlc` and make it take only one completion event instead of a vector, find a way to re-subscribe to HTLC events if the stream ever dies.

Also, what are the delivery guarantees if a stream dies because of e.g. an internet interruption? Do we have to resubmit all pending HTLCs?

okjodom: You know if I just got rid of this batching and streaming in all places except subscribe htlcs, my life would be way simpler!
On it!

justinmoon: Nice

gateway/ln-gateway/src/actor_fork.rs
```diff
@@ -42,7 +53,110 @@ impl GatewayActor {
...
+        actor.subscribe_htlcs().await?;
+
+        Ok(actor)
+    }
+
+    async fn subscribe_htlcs(&self) -> Result<()> {
+        let actor = self.to_owned();
+        let lnrpc = self.lnrpc.to_owned();
+        let short_channel_id = self.client.config().mint_channel_id;
+        let mut tg = self.task_group.clone();
```
mxz42, line 71: IMHO, this seems kind of weird to always (saw this now multiple times in the `GatewayActor`) clone required variables, especially cloning `self`. 
I'd love to avoid this if possible. 
For instance, we should be able to take a `&mut self` in the method itself, so we can just use `self.task_group.spawn()` and don't need to clone it. 
On the other hand, for the `lnrpc` we can create a clone (that is also just a copy of the `Arc`) which can be passed into the `task_group.spawn` closure. 
 
something like 
```rust
{ // open scope for variables moved into the task group closure
     let lnrpc_copy = self.lnrpc.clone();
     self.task_group.spawn(....) {
        ....
        lnrpc_copy.complete_htlc(...)
    }

} // end of scope

File: gateway/ln-gateway/src/actor_fork.rs [hunk 7]
```diff
@@ -42,7 +53,110 @@ impl GatewayActor {
...
+                                .await
+                            {
+                                error!("Failed to complete HTLC: {:?}", e);
+                                // Note: To prevent loss of funds for the gateway,
+                                // we should either retry completing the htlc or reclaim funds from the federation
+                            };
+                        }
+                        Err(e) => {
+                            error!("Failed to process intercepted HTLC: {:?}", e);
+                            let _ = lnrpc
```
mxz42, line 144: why are we ignoring the `Result` returned by `complete_htlc()`?

gateway/ln-gateway/src/rpc/lnrpc_client.rs
```diff
@@ -0,0 +1,117 @@
...
+    LnGatewayError, Result,
+};
+
+#[async_trait]
+pub trait ILnRpcClient: Debug + Send + Sync {
+    /// Get the public key of the lightning node
+    async fn get_pubkey(&self) -> Result<GetPubKeyResponse>;
+
+    /// Attempt to pay an invoice using the lightning node
+    async fn pay_invoice(&self, invoice: PayInvoiceRequest) -> Result<PayInvoiceResponse>;
```
mxz42, line 28: perhaps rename this to `pay(invoice)` as in `get(amount)` [here](https://github.com/fedimint/fedimint/blob/1d3758de7d6f8a0e6a466ee1f8e99cd65f0d90a5/fedimint-api/src/tiered_multi.rs#L132)

mxz42 (commented):
wow, huge PR! great work!
I left a couple nits here and there. There are quite a lot of `clone`s in the `Gateway` implementation. Maybe we can get rid of some of them. 

I'm also a huge fan of documentation. It only takes a couple minutes now, but saves a lot of time and effort later when you haven't look at the code for a while. And, of course, it helps other people who are reading this the first time. 

Take a look at the [`async recv()`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv) for the `mpsc::Receiver` as well. I think we can use it and get rid of the manual sleep (unless there is a reason that I'm not aware of?).

okjodom:
Great pointers, @mxz42 . 
I won't be able to get to this PR this week. Marking as draft until I can come back and polish it

gateway/ln-gateway/src/gateway.rs
```diff
@@ -0,0 +1,390 @@
...
+        loop {
+            // Shut down main loop if requested
+            if loop_ctrl.is_shutting_down() {
+                break;
+            }
+
+            let least_wait_until = Instant::now() + Duration::from_millis(100);
+
+            // Handle messages from webserver and plugin
+            while let Ok(msg) = self.receiver.try_recv() {
```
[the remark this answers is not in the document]
mxz42: well, I guess you used the manual sleep to periodically check the `is_shutting_down()` method. My thought was to use [`tokio::select` somewhat like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=851813e40e886fa28bff85e6cda917ab).  However, I'm not sure if our task group lets us do that. perhaps @dpc or @elsirion could chime in?

gateway/ln-gateway/src/rpc/lnrpc_client.rs
```diff
@@ -40,3 +47,51 @@ impl DynLnRpcClient {
         DynLnRpcClient(client)
     }
 }
+
+/// An `ILnRpcClient` that wraps around `GatewayLightningClient` for convenience,
+/// and makes real RPC requests over the wire to a remote lightning node.
+/// The lightnign node is exposed via a corresponding `GatewayLightningServer`.
+#[derive(Debug)]
+pub struct NetworkLnRpcClient {
+    client: GatewayLightningClient<Channel>,
```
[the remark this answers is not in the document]
okjodom: Maybe rename `NetworkLnRpcClient` to `NetworkLightningRpcClinet`?

`GatewayLightningClient` is a gRPC autogen type, having [defined `GatewayLightning` here](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/proto/gatewaylnrpc.proto#L7)

gateway/ln-gateway/src/rpc/lnrpc_client.rs
```diff
@@ -40,3 +47,51 @@ impl DynLnRpcClient {
...
+    }
+}
+
+#[async_trait]
+impl ILnRpcClient for NetworkLnRpcClient {
+    async fn get_pubkey(&self) -> Result<GetPubKeyResponse> {
+        unimplemented!()
+    }
+
+    async fn pay_invoice(&self, _invoice: PayInvoiceRequest) -> Result<PayInvoiceResponse> {
```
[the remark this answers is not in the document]
okjodom: Shortened forms are neat. But seems limiting on the get side because we cannot then have `get(A)`, `get(B)`, no ?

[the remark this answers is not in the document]
okjodom: Opting to go with `long_snake_api()` definitions for now. We can polish names

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Gateway webserver authentication password
+    #[arg(long = "password", env = "FM_GATEWAY_PASSWORD")]
+    pub password: String,
+
+    /// Public URL to a Gateway Lightning rpc service
+    #[arg(long = "lnrpc-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
+    pub lnrpc_addr: Url,
+}
+
+/// Fedimint Gateway Binary
```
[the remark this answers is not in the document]
okjodom: Went all out. This is the entry point to gatewayd :)

gateway/ln-gateway/src/rpc/lnrpc_client.rs
okjodom: I like the short form. But kept long snake form for consistency with the getter apis, because I don't see how to shorten two `get_x()` methods that take no params

gateway/ln-gateway/src/gateway.rs
```diff
@@ -0,0 +1,390 @@
...
+
+            for config in configs {
+                let client = self
+                    .client_builder
+                    .build(config.clone(), decoders.clone(), module_gens.clone())
+                    .await
+                    .expect("Could not build federation client");
+
+                if let Err(e) = self.connect_federation(Arc::new(client)).await {
+                    error!("Failed to connect federation: {}", e);
```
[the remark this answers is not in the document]
okjodom: yeah, telemetry pipeline from error. Let the gateway admins tend to the fact that federations are starved of gateway goodness :)

Applying improved error message

gateway/ln-gateway/src/actor_fork.rs
okjodom: ack. using `to_owned` as a safer copy method. Recommendation from @dpc :)

gateway/ln-gateway/src/actor_fork.rs
okjodom: this specific complete htlc  requires no futher action. If we fail to send the complete htlc message, lightning node will still cancel HTCL after expiry period lapses. Result can be safely ignored.

Added a note and TODO comment

gateway/ln-gateway/src/rpc/lnrpc_client.rs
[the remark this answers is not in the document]
okjodom: [pending]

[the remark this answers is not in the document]
mxz42: hm, I see. I'd probably prefer to leave it as it is then. That'd be consistent we most of the rest of the code base: 
```sh
$ rg '(struct|enum|trait|type) \w*?Ln\w+'
client/client-lib/src/ln/mod.rs
40:pub struct LnClient {
316:pub enum LnClientError {

client/client-lib/src/api.rs
528:pub trait LnFederationApi {

gateway/ln-gateway/src/ln.rs
6:pub trait LnRpc: Send + Sync + 'static {

integrationtests/tests/fixtures/utils.rs
11:pub struct LnRpcAdapter {

gateway/ln-gateway/src/rpc/lnrpc_client.rs
23:pub trait ILnRpcClient: Debug + Send + Sync {
55:pub struct NetworkLnRpcClient {

gateway/ln-gateway/src/lib.rs
55:pub struct LnGateway {
410:pub enum LnGatewayError {

gateway/tests/tests/fixtures/ln.rs
12:pub struct MockLnRpc {
```

Arguably, there are still some types which contain `Lightning` but I think we should move to `Ln` as it's shorter and still unique. Hence, having this struct is fine, as we will probably work the the "convenience wrapper" `NetworkLnRpcClientd` anyway (if I understood it correctly). 
@dpc any opinion on the naming here?

fyi, the types that currently contain `Lightning`: 

```sh
$ rg '(struct|enum|trait|type) \w*?Lightning\w+'
client/client-lib/src/ln/db.rs
101:pub struct LightningGatewayKey;
110:pub struct LightningGatewayKeyPrefix;

integrationtests/tests/fixtures/fake.rs
20:pub struct FakeLightningTest {

modules/fedimint-ln/src/common.rs
11:pub struct LightningDecoder;

modules/fedimint-ln/src/config.rs
15:pub struct LightningConfig {
23:pub struct LightningConfigConsensus {
38:pub struct LightningConfigPrivate {
51:pub struct LightningClientConfig {

modules/fedimint-ln/src/db.rs
123:pub struct LightningGatewayKey(pub PublicKey);
132:pub struct LightningGatewayKeyPrefix;

gateway/ln-gateway/src/ln.rs
24:pub struct LightningError(pub Option<i32>);

integrationtests/tests/fixtures/real.rs
24:pub struct RealLightningTest {

integrationtests/tests/fixtures/mod.rs
480:pub trait LightningTest {

modules/fedimint-ln/src/lib.rs
94:pub struct LightningInput {
126:pub enum LightningOutput {
185:pub enum LightningOutputOutcome {
209:pub struct LightningGateway {
227:pub struct LightningConsensusItem {
239:pub struct LightningVerificationCache;
242:pub struct LightningGen;
1063:pub enum LightningError {
```

gateway/ln-gateway/src/rpc/lnrpc_client.rs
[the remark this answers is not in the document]
mxz42: > Shortened forms are neat. But seems limiting on the get side because we cannot then have `get(A)`, `get(B)`, no ?

hm :thinking: don't understand what you mean

gateway/ln-gateway/src/actor_fork.rs
mxz42: ah, thanks. :+1:  follow-up: this expiry time is usually rather short, correct?  Does the expiry correspond to [this field](https://github.com/fedimint/fedimint/blob/cd413f1598b4ee209a19a9f3ba8db59cf0828be8/gateway/ln-gateway/proto/gatewaylnrpc.proto#L81)?

gateway/ln-gateway/src/rpc/lnrpc_client.rs
mxz42: right, now I see what you mean. there is `get_balance` and `get_info`. hm, I guess we could do something like `pay(invoice)`, `fed_info()`, and `balance()`

okjodom:
@mxz42 :)

> wow, huge PR! great work! I left a couple nits here and there. There are quite a lot of `clone`s in the `Gateway` implementation. Maybe we can get rid of some of them.

ack!

> I'm also a huge fan of documentation. It only takes a couple minutes now, but saves a lot of time and effort later when you haven't look at the code for a while. And, of course, it helps other people who are reading this the first time.

Some documentation to set stage for this big architectural change: #1559 

> Take a look at the [`async recv()`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv) for the `mpsc::Receiver` as well. I think we can use it and get rid of the manual sleep (unless there is a reason that I'm not aware of?).

Deferring this to #1603  - I suspect this would bring a nice atomic and observable improvement to gateway perf. This PR is already doing a lot of changes, of which I'm trying to prioritize and focus on architectural changes first.

gateway/ln-gateway/src/gatewayd/actor.rs
```diff
@@ -57,7 +75,114 @@ impl GatewayActor {
...
         });
 
-        Ok(Self { client })
+        let actor = Self {
+            client,
+            lnrpc,
+            task_group,
+        };
+
+        actor.subscribe_htlcs().await?;
```
mxz42, line 84: (just writing out thoughts) 
I wonder if we should separate the `new()` and the `subscribe_htlcs()` method.  

`new` just creates registered GatewayActor and returns it and then we would call `.subscribe_htlcs` on it. 
However, it is "one step" and would require to make the `subscribe_htlcs` method public which we might not want to do. 

opinion?

gateway/ln-gateway/src/gatewayd/actor.rs
```diff
@@ -57,7 +75,114 @@ impl GatewayActor {
...
+                    Ok(Some(msg)) => Some(msg),
+                    Ok(None) => {
+                        warn!("HTLC stream closed by service");
+                        None
+                    }
+                    Err(e) => {
+                        error!("HTLC stream closed with error: {:?}", e);
+                        None
+                    }
+                } {
```
mxz42, line 117: IMHO, the struct unpacking and matching in one big block looks sort of weird. Maybe it's considered idiomatic rust, maybe not; I don't know. But I'd prefer to structure this a bit different (see next comment)

gateway/ln-gateway/src/gatewayd/actor.rs
```diff
@@ -57,7 +75,114 @@ impl GatewayActor {
...
+                    if subscription.is_shutting_down() {
```
mxz42, line 118: If I understand this correctly, we only check if our task group is shutting down after we asynchronously received a message with `stream.message()`. Hence, as long as we didn't receive a message, we're not checking for shutdown. 

I think we want to do it as in the [`peers.rs`](https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/net/peers.rs#L163-L169). We create a shutdown handler and then use `tokio::select` to choose where to make progress; also note the `while` loop for the shutdown check.
[Another similar example is used in `task.rs`](https://github.com/fedimint/fedimint/blob/master/fedimint-api/src/task.rs#L126-L148)

gateway/ln-gateway/src/gatewayd/lnrpc_client.rs
```diff
@@ -0,0 +1,130 @@
...
+
+// TODO: Issue 1554: Define gatewaylnpc spec for getting route hints
+pub struct GetRouteHintsResponse {
+    pub route_hints: Vec<RouteHint>,
+}
+
+#[async_trait]
+pub trait ILnRpcClient: Debug + Send + Sync {
+    /// Get the public key of the lightning node
+    async fn get_pubkey(&self) -> Result<GetPubKeyResponse>;
```
mxz42, line 31: nit: just naming suggestion. but feel free to move it to a later point I guess.
- `pubkey()`
- `route_hints()`
- `pay(invoice)`  (we had that already :'D)

gateway/ln-gateway/src/gatewayd/lnrpc_client.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// Get route hints to the lightning node
+    async fn get_route_hints(&self) -> Result<GetRouteHintsResponse>;
+
+    /// Attempt to pay an invoice using the lightning node
+    async fn pay_invoice(&self, invoice: PayInvoiceRequest) -> Result<PayInvoiceResponse>;
+
+    /// Subscribe to intercept htlcs that belong to a specific mint identified by `short_channel_id`
+    async fn subscribe_intercept_htlcs(
```
mxz42, line 40: again just typing out thoughts: 
What we're really doing here is subscribing to a `channel_id`. I was wondering if we could make that clearer by naming this `subscribe(channel_id)` and constructing the `SubscribeInterceptHtlcRequest` inside that function?

gateway/ln-gateway/src/lib.rs
```diff
@@ -456,6 +453,8 @@ impl Drop for LnGateway {
 pub enum LnGatewayError {
     #[error("Federation client operation error: {0:?}")]
     ClientError(#[from] ClientError),
+    #[error("Lightning rpc operation error: {0:?}")]
+    LnrpcError(#[from] tonic::Status),
```
mxz42, line 457: just a nit: since `ln` and `rpc` are two different nouns, maybe call this `LnRpcError`

gateway/ln-gateway/src/gatewayd/actor.rs
okjodom: At this moment, I think an actor is useless without an active subscription for htlcs. For this reason, `subscribe_htlcs` made sense an important pre-condition to bootstrapping and returning a new, valid actor.

We can make it one step at a later time, when we have reasons for actors to be valid without active subs for htlcs.

gateway/ln-gateway/src/gatewayd/lnrpc_client.rs
okjodom: adopted. Plus, renamed `subscribe_intercept_htlcs` to `subscribe_htlcs`. simple

gateway/ln-gateway/src/gatewayd/lnrpc_client.rs
okjodom: I don't see how it makes it clearer.
Also, we'd need to be consistent across the interfaces provided by `ILnRpcClient` definition, of which some of those types like `CompleteHtlcsRequest` are complex, and we can just ask the users to just construct the types when making the calls.

**wontfix**

gateway/ln-gateway/src/lib.rs
okjodom: 💯

gateway/ln-gateway/src/gatewayd/lnrpc_client.rs
mxz42: ack

gateway/ln-gateway/src/gatewayd/actor.rs
mxz42: sounds reasonable :+1:

gateway/ln-gateway/src/gatewayd/actor.rs
mxz42: I'm certainly not against unpacking, that's a pretty cool feature and I'm quite sure there are no performance costs either. 

I think we'll see what this loop ends up with, when `await`ing the `stream.message()` and the shutdown handler of the task group

gateway/ln-gateway/src/gatewayd/actor.rs
okjodom: [pending]

gateway/ln-gateway/src/rpc/lnrpc_client.rs
```diff
@@ -0,0 +1,50 @@
...
+#[async_trait]
+pub trait ILnRpcClient: Debug + Send + Sync {
+    /// Get the public key of the lightning node
+    async fn get_pubkey(&self) -> Result<GetPubKeyResponse>;
+
+    /// Attempt to pay an invoice using the lightning node
+    async fn pay_invoice(
+        &self,
+        invoices: Vec<PayInvoiceRequest>,
+    ) -> Result<Streaming<PayInvoiceResponse>>;
```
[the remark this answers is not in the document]
okjodom: After #1426, only `subscribe_intercept_htlcs` exposes `tonic::Streaming` type.

Now, `Streaming` has a complex `Item = std::result::Result<SubscribeInterceptHtlcsResponse, Status>` where `Status` is yet another nuanced tonic type we'd need to mock redefine.

To make mocking of `ILnRpcClient` easy, I aliased the tonic type : `pub type HtlcStream = Streaming<SubscribeInterceptHtlcsResponse>;`. This way, testing libs don't have to explicitly take a dependency on tonic.

Mocking `Ok` results should be easy. Mocking `Err` results might still be challenging since response would need to be a tonic `Status`.

gateway/ln-gateway/src/rpc/lnrpc_client.rs
```diff
@@ -101,3 +101,26 @@ impl ILnRpcClient for NetworkLnRpcClient {
         unimplemented!()
     }
 }
+
+/// A generic factory trait for creating `DynLnRpcClient` instances.
+#[async_trait]
+pub trait ILnRpcClientFactory: Debug {
```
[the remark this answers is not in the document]
okjodom: For the `NetworkLnRpcClient` which sends rpc requests over the wire, we may need a way of spawning a new client for every actor / actor rpc scenario because each rpc call is actually blocking on the client [ _read somewhere, looking for the right docs for reference_ ]

- If this is the case, we'd end up with a perf bottleneck, situation similar to #1577 
- It's okay to start off simple, with a single `DynLnRpcClient` instantiated and provided to gatewayd at boot. The factory pattern might be a solution down the road if we observe issues with the rpc client perf.

gateway/ln-gateway/src/actor_fork.rs
okjodom: Yes.
We'll need to dig into expiry a little more down the road for optimizations of intercepted htlc processing.
For example, the gateway might want to reject / ignore a htlc in cases where that field indicates the htlc is super close to expiry...

gateway/ln-gateway/src/rpc/lnrpc_client.rs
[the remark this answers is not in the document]
okjodom: #1636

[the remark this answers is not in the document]
okjodom: #1636

gateway/ln-gateway/src/gateway.rs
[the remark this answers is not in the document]
okjodom: defer to #1603

gateway/ln-gateway/src/actor_fork.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub async fn restore(&self) -> Result<()> {
+        // TODO: get the task group from `self`
+        let mut task_group = TaskGroup::new();
+
+        self.client
+            .mint_client()
+            .restore_ecash_from_federation(10, &mut task_group)
+            .await
+            .map_err(LnGatewayError::Other)?
+            .map_err(|e| LnGatewayError::Other(e.into()))?;
```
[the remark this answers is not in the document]
okjodom: Believe it's because `.restore_ecash_from_federation` returns a cancellable, which we observe with the second `.map_err`

gateway/ln-gateway/src/gatewayd/actor.rs
elsirion: We could solve this as a follow-up. It's still broken everywhere else too.

elsirion (approved):
On a high level this PR looks like it moved into the right direction, maybe @mxz42 can take a closer look again.

mxz42:
alright, if you could take a look at the [comment](https://github.com/fedimint/fedimint/pull/1337#discussion_r1103121577)  @okjodom   and this has not effect, then LGTM.  

Then the rough skeleton is there. Let's fix the message loop in a separate PR. also makes it easier to review

.tmuxinator.yml
```diff
@@ -38,6 +38,11 @@ windows:
           - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN2_DIR --addr=127.0.0.1:9001 --plugin=$FM_BIN_DIR/gateway-cln-extension &
           - echo $! >> $FM_PID_FILE
           - fg
+        - gateway:
+          - sleep 7 # wait for gateway-cln-extension to start
+          - $FM_BIN_DIR/gatewayd
+          - echo $! >> $FM_PID_FILE
+          - fg
```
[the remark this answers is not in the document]
elsirion: (you are adding a pane, so the index might change here, my guess would be `1.4` -> `1.5`)

mxz42 (approved):

elsirion:
Is `gatewayd` supposed to work in `tmuxinator`? 

```
source .tmpenv; alias ln1="\$FM_LN1"; alias ln2="\$FM_LN2"; alias bitcoin-cli="\$FM_BTC_CLIENT"; alias fedimint-cli="\$FM_MINT_CLIENT"; alias gateway-cli="\$FM_GATEWAY_CLI"; alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"; alias dbtool="\$FM_DB_TOOL"; alias restart="./scripts/restart-tmux.sh"

[user@minimint:~/minimint]$ source .tmpenv; alias ln1="\$FM_LN1"; alias ln2="\$FM_LN2"; alias bitcoin-cli="\$FM_BTC_CLIENT"; alias fedimint-cli="\$FM_MINT_CLIENT"; alias gateway-cli="\$FM_GATEWAY_CLI"; alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"; alias dbtool="\$FM_DB_TOOL"; alias restart="./scripts/restart-tmux.sh"

[user@minimint:~/minimint]$ sleep 7
$FM_BIN_DIR/gatewayd
echo $! >> $FM_PID_FILE
fg

[user@minimint:~/minimint]$ $FM_BIN_DIR/gatewayd
Error: Other: Failed to connect to lnrpc server

Caused by:
    Failed to connect to lnrpc server

[user@minimint:~/minimint]$ echo $! >> $FM_PID_FILE

[user@minimint:~/minimint]$ fg
-bash: fg: current: no such job
```

Also, `restart` doesn't work at all, did anyone ever test it?

elsirion:
After playing around with it a bit I think there is a mistake in the way `gatewayd` is stated (doesn't wait for CLN to come online), but even if that wasn't the case it hits some `unimplemented!()` call. I'm ok with going ahead in that unfinished state, will take some iterations to make `gatewayd` really work.

elsirion (approved):

okjodom:
> Is `gatewayd` supposed to work in `tmuxinator`?
> 
> ```
> source .tmpenv; alias ln1="\$FM_LN1"; alias ln2="\$FM_LN2"; alias bitcoin-cli="\$FM_BTC_CLIENT"; alias fedimint-cli="\$FM_MINT_CLIENT"; alias gateway-cli="\$FM_GATEWAY_CLI"; alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"; alias dbtool="\$FM_DB_TOOL"; alias restart="./scripts/restart-tmux.sh"
> 
> [user@minimint:~/minimint]$ source .tmpenv; alias ln1="\$FM_LN1"; alias ln2="\$FM_LN2"; alias bitcoin-cli="\$FM_BTC_CLIENT"; alias fedimint-cli="\$FM_MINT_CLIENT"; alias gateway-cli="\$FM_GATEWAY_CLI"; alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"; alias dbtool="\$FM_DB_TOOL"; alias restart="./scripts/restart-tmux.sh"
> 
> [user@minimint:~/minimint]$ sleep 7
> $FM_BIN_DIR/gatewayd
> echo $! >> $FM_PID_FILE
> fg
> 
> [user@minimint:~/minimint]$ $FM_BIN_DIR/gatewayd
> Error: Other: Failed to connect to lnrpc server
> 
> Caused by:
>     Failed to connect to lnrpc server
> 
> [user@minimint:~/minimint]$ echo $! >> $FM_PID_FILE
> 
> [user@minimint:~/minimint]$ fg
> -bash: fg: current: no such job
> ```


I have a recent `unimplemeneted` placeholder for getting route hints. This is on gatewayd boot path so it breaks running gatewayd on tmux. Fixing with #1554

> > Also, `restart` doesn't work at all, did anyone ever test it (unrelated to this PR, your change works @okjodom, the restart script just lacks permissions afaik)? Tracked in #1650 now.

True. restart just lacks permissions

mxz42:
> > > Also, `restart` doesn't work at all, did anyone ever test it (unrelated to this PR, your change works @okjodom, the restart script just lacks permissions afaik)? Tracked in #1650 now.
> 
> True. restart just lacks permissions

yep, had a PR submitted this afternoon. Is now merged. 

As I didn't want to leave it as partially fixed (regarding correct shebang for nix environments) I added another PR for the rest of #1650. See PR #1651

PR #1345, opened 2023-01-16
cleanup: move code out of api_endpoint macro

maan2003:


elsirion:
Great change! I had trouble with that quite often when debugging lifetimes inside a macro.

PR #1348, opened 2023-01-16
cleanup: remove serialize to string and deserialize hack

Maan2003:
this is no longer needed because use `SerdeTransaction` now.
earlier some types from hbbft could not be deserialized directly.


dpc (approved):

elsirion:
Which other change does this relate to, that made the hack unnecessary?

Maan2003:
using `SerdeTransaction`

earlier `Transaction` had types that could not be directly deserialized, but now `SerdeTransaction` deserializes to bytes first then use `Decodable` trait instead.

PR #1349, opened 2023-01-16
Replace uses of `hex` dependency

mxz42:
part of #1307

more work to remove the `hex` dependency crate. I can't remove the hex dependency in cargo.toml in `fedimint-api` yet as it is needed because the mint module uses it.
used [here](https://github.com/fedimint/fedimint/blob/8d4ef26a814e53ba7012b1b50fb3afec4382c920/modules/fedimint-mint/src/common.rs#L17) and [here](https://github.com/fedimint/fedimint/blob/8d4ef26a814e53ba7012b1b50fb3afec4382c920/modules/fedimint-mint/src/db.rs#L158)

example: 
```rust
#[derive(Debug, Serialize, Deserialize, Encodable, Decodable)]
pub struct BackupRequest {
    pub id: secp256k1::XOnlyPublicKey,
    #[serde(with = "hex::serde")]
    pub payload: Vec<u8>,
    pub timestamp: std::time::SystemTime,
}
```

thoughts on replacing this? writing a deserialize and serialize function for `Vec<u8>` that de/serializes to hex and use it with `#[serde(serialize_with = ...)]` ? It is still wanted I guess?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2598,7 +2598,6 @@ dependencies = [
  "fedimint-rocksdb",
  "fedimint-testing",
  "futures",
- "hex",
  "hkdf",
  "js-sys",
  "jsonrpsee-core",
--- a/client/client-lib/Cargo.toml
+++ b/client/client-lib/Cargo.toml
@@ -21,7 +21,6 @@ bincode = "1.3.1"
 bitcoin = "0.29.2"
 bitcoin_hashes = "0.11.0"
 futures = "0.3.24"
-hex = "0.4.3"
 hkdf = { path = "../../crypto/hkdf" }
 lightning-invoice = "0.21.0"
 lightning = "0.0.113"
--- a/client/client-lib/src/mint/mod.rs
+++ b/client/client-lib/src/mint/mod.rs
@@ -5,6 +5,7 @@ use std::fmt;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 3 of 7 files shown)


dpc:
> writing a deserialize and serialize function for Vec<u8> that de/serializes to hex and use it with #[serde(serialize_with = ...)]

That would be a way to go. Maybe `rust-bitcoin` already have one? If not, maybe you'd like to contribute to it?

If you put both functions in a module, we could get

```
#[serde(with = "fedimint_api::hex::serde")]
```

now and

```
#[serde(with = "bitcoin_hashes::hex::serde")]
```

later.

dpc:
> Do you think it would be generally useful to have an implementation that de/serializes a Vec<u8> to/from hex?

@mxz42 I don't understand. Isn't that what we're doing here all over? :D

elsirion (approved):

mxz42:
> > Do you think it would be generally useful to have an implementation that de/serializes a Vec to/from hex?
> 
> @mxz42 I don't understand. Isn't that what we're doing here? :D

yes :D sorry, vague question. My question is if it is useful for `bitcoin_hashes` to have a de/serialize() to/from `Vec<u8>`.  But I guess it is :)

dpc:
If they don't have already, sure it is. It was the only part of `hex` dependency that I liked. :D

PR #1350, opened 2023-01-16
fix: rust_log_full_enabled check env var only once

Maan2003:
LocalKey::with executes the closure everytime :shrug: 

thread_local is lazy by default. So we can do non const initialization.

### DIFF
--- a/fedimint-api/src/fmt_utils.rs
+++ b/fedimint-api/src/fmt_utils.rs
@@ -1,18 +1,14 @@
-use std::cell::Cell;
 use std::fmt;
 use std::thread_local;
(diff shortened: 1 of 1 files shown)


dpc:
Oooooooooooh. Good catch!

dpc (approved):

elsirion:
Good catch, I actually thought the same at first but then read the `with` docs wrong :see_no_evil: 

> This will lazily initialize the value if this thread has not referenced this key yet.

https://doc.rust-lang.org/std/thread/struct.LocalKey.html#method.with

PR #1351, opened 2023-01-16
Re-try committing DB tx when processing consensus outcomes

### LINKED ISSUE #1316: Consensus DB commits can panic
We just experienced a panic in a live deployment on this line:

https://github.com/fedimint/fedimint/blob/73d851fc208f4bb0dadd65c148800a9c753a30d7/fedimint-server/src/consensus/mod.rs#L320

```
Jan 12 23:20:14 charlie fedimintd[32185]: 2023-01-12T23:20:14.174355Z  INFO fedimint_server:
Jan 12 23:20:14 charlie fedimintd[32185]: - Epoch: 510 [PeerId(0), PeerId(1), PeerId(2)] -
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 0: Transaction
Jan 12 23:20:14 charlie fedimintd[32185]:     Input: Lightning Contract f3dd1a942454997378c29814088d167470a45811ac636457a2f0ccd4ebf63730 with amount 40000000 msat
Jan 12 23:20:14 charlie fedimintd[32185]:     Output: Mint Coins 40000000 msat
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 0: Module CI: module=0 ci=Wallet Block Height 771650
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 0: Outcome Signature
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 1: Module CI: module=0 ci=Wallet Block Height 771650
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 1: Outcome Signature
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 2: Module CI: module=0 ci=Wallet Block Height 771650
Jan 12 23:20:14 charlie fedimintd[32185]:   Peer 2: Outcome Signature
Jan 12 23:20:14 charlie fedimintd[32185]: thread 'main' panicked at 'DB Error: Resource busy: ', /build/source/fedimint-server/src/consensus/mod.rs:302:36
Jan 12 23:20:14 charlie fedimintd[32185]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It is unclear to me why this error happened, there should not be any conflicting, parallel writes. That's why we need to review each call of `commit()` in the server code base and replace it with a loop that will re-try the transaction (re-creating it) so that this can not happen anymore. The consensus code should not have any side-effects not captured by the DB tx, so this seems safe (although inefficient, but it only seems to happen rarely).

elsirion:
fixes #1316

### DIFF
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -245,6 +245,8 @@ impl FedimintConsensus {
             .flat_map(|(peer, cis)| cis.into_iter().map(move |ci| (peer, ci)))
             .unzip_consensus_item();
 
+        let mut dbtx = self.db.begin_transaction().await;
+
         // Begin consensus epoch
         {
             let per_module_cis: HashMap<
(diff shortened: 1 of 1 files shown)


dpc:
@elsirion LGTM, but we should really split this function into sub-functions that take are explicit about their inputs&output, so it's easier to reason about this loop. :D

dpc (approved):

Issue #1340, opened 2023-01-16
Replace `get_last_epoch` api with `get_epoch_count`


dpc:
The problem with the current API is that it's inclusive bound:  it has no way to represent "no epochs yet" (would be 0 with exclusive one).

Prior art: `bitcoind` has a `getblockcount`, not `getlastblock` :)

It might lead to subtle errors.

dpc:
Same in the `IBitcoindRpc`

dpc:
@jp1ac4 , please see what the existing endpoint is doing. What this issue is about is basically making it return `0` when it was previously returning `None` (or error) and `N + 1` where it was previously returning `Some(N)` + renaming it.

jp1ac4:
Hi @dpc, thanks for the hint. Are any changes required to `fetch_last_epoch`, which I believe is using the `/epoch` endpoint?

dpc:
We should rename everything to `fetch_epoch_count` and similar (to contain "count" and not "last"), just like `getblockcount` of `bitcoind`, then fix the code.

ngutech21:
Issue can be closed because branch got merged

PR #1363, opened 2023-01-17
Add electrum support

dpc:
This adds (not really tested properly yet) support for electrum as backend.

Fixes some issues I've found along the way.

### DIFF
--- a/fedimint-api/src/bitcoin_rpc.rs
+++ b/fedimint-api/src/bitcoin_rpc.rs
@@ -5,20 +5,52 @@ use url::Url;
 /// Name of the env value used for passing bitcoind rpc url to modules that need it
 pub const FM_BITCOIND_RPC_ENV: &str = "FM_BITCOIND_RPC";
 
+/// Name of the env value used for passing electrum rpc url to modules that need it
+pub const FM_ELECTRUM_RPC_ENV: &str = "FM_ELECTRUM_RPC";
+
 /// Default url that will be used if [`FM_BITCOIND_RPC_ENV`] is not set
 pub const FM_BITCOIND_RPC_DEFAULT_FALLBACK: &str = "http://127.0.0.1:8332";
+
+/// Bitcoin RPC backend
+pub enum BitcoindRpcBackend {
+    /// Bitcoin Core RPC
+    Bitcoind(Url),
+    /// Electrum RPC
+    Electrum(Url),
+}
+
+pub enum BitcoinRpcBackendType {
+    Bitcoind,
+    Electrum,
+}
 
 /// Get the value of url the module would use by reading it from process environemnt
 ///
 /// Should be used in test code only to mimick prod code behavior
-pub fn read_bitcoin_rpc_env_from_global_env() -> anyhow::Result<Url> {
-    fm_bitcoind_rpc_env_value_to_url(std::env::var_os(FM_BITCOIND_RPC_ENV).as_deref())
+pub fn read_bitcoin_backend_from_global_env() -> anyhow::Result<BitcoindRpcBackend> {
+    select_bitcoin_backend_from_envs(
+        std::env::var_os(FM_BITCOIND_RPC_ENV).as_deref(),
+        std::env::var_os(FM_ELECTRUM_RPC_ENV).as_deref(),
+    )
+}
+
+pub fn select_bitcoin_backend_from_envs(
+    bitcoind_rpc: Option<&OsStr>,
+    electrum_rpc: Option<&OsStr>,
+) -> anyhow::Result<BitcoindRpcBackend> {
+    Ok(if let Some(val) = bitcoind_rpc {
+        BitcoindRpcBackend::Bitcoind(fm_bitcoind_rpc_env_value_to_url(Some(val))?)
+    } else if let Some(val) = electrum_rpc {
+        BitcoindRpcBackend::Electrum(fm_electrum_rpc_env_value_to_url(val)?)
+    } else {
+        BitcoindRpcBackend::Bitcoind(fm_bitcoind_rpc_env_value_to_url(None)?)
+    })
 }
 
 /// Get the value of bitcoin rpc url to use, from the value of env variable
 ///
 /// Useful in places where variable value is already available.
-pub fn fm_bitcoind_rpc_env_value_to_url(value: Option<&OsStr>) -> anyhow::Result<Url> {
+fn fm_bitcoind_rpc_env_value_to_url(value: Option<&OsStr>) -> anyhow::Result<Url> {
     Ok(if let Some(url_str) = value {
         Url::parse(
             url_str
@@ -28,4 +60,10 @@ pub fn fm_bitcoind_rpc_env_value_to_url(value: Option<&OsStr>) -> anyhow::Result
     } else {
         Url::parse(FM_BITCOIND_RPC_DEFAULT_FALLBACK)?
     })
+}
+
+pub fn fm_electrum_rpc_env_value_to_url(value: &OsStr) -> anyhow::Result<Url> {
+    Ok(Url::parse(value.to_str().ok_or_else(|| {
+        anyhow::format_err!("Url not ascii text")
+    })?)?)
 }
--- a/fedimint-bitcoind/src/bitcoincore_rpc.rs
+++ b/fedimint-bitcoind/src/bitcoincore_rpc.rs
@@ -1,7 +1,13 @@
+use std::fmt;
+
 use ::bitcoincore_rpc::bitcoincore_rpc_json::EstimateMode;
 use ::bitcoincore_rpc::jsonrpc::error::RpcError;
 use ::bitcoincore_rpc::{jsonrpc, Auth, RpcApi};
-use anyhow::format_err;
+use anyhow::{bail, format_err, Context};
+use bitcoin::consensus::Encodable;
+use bitcoin_hashes::hex::ToHex;
+use electrum_client::ElectrumApi;
+use fedimint_api::bitcoin_rpc::BitcoindRpcBackend;
 use fedimint_api::module::__reexports::serde_json::Value;
 use jsonrpc::error::Error as JsonError;
 use serde::Deserialize;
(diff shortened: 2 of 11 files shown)


fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -162,3 +180,76 @@ where
...
+                .get(0)
+                .ok_or_else(|| format_err!("empty block headers response"))?
+                .block_hash())
+        })
+    }
+
+    async fn get_block(&self, _hash: &BlockHash) -> Result<Block> {
+        // TODO: electrum does not seem to be able to return whole blocks
+        // do we **really** need it?
+        unimplemented!()
```
dpc, line 233: This is a blocker. Seems like we're using it for checking peg-out tx status?

What if we checked instead the history of the status of the output? https://docs.rs/electrum-client/0.12.0/electrum_client/trait.ElectrumApi.html#tymethod.script_get_history

elsirion: This is kinda problematic: when targeting core we should not assume an address index or something like this, while electrum apparently can't serve whole blocks. Maybe we can re-define the API boundary? 

Caution: When implementing using e.g. `script_get_history` we still need to wait till consensus height reaches the block height of the returned UTXOs.

elsirion: Maybe Electrum is the wrong way to go and esplora would be much easier? Their API also has an easy way to get a txoutproof and not just an electrum merkleproof (they can probably be converted, but the tree algos necessary are a bit annoying).

fedimint-api/src/bitcoin_rpc.rs
```diff
@@ -5,20 +5,59 @@ use url::Url;
...
 ///
 /// Should be used in test code only to mimick prod code behavior
-pub fn read_bitcoin_rpc_env_from_global_env() -> anyhow::Result<Url> {
-    fm_bitcoind_rpc_env_value_to_url(std::env::var_os(FM_BITCOIND_RPC_ENV).as_deref())
+pub fn read_bitcoin_backend_from_global_env() -> anyhow::Result<BitcoindRpcBackend> {
+    select_bitcoin_backend_from_envs(
+        std::env::var_os(FM_BITCOIND_RPC_ENV).as_deref(),
+        std::env::var_os(FM_ELECTRUM_RPC_ENV).as_deref(),
+    )
+}
```
elsirion, line 30: Why make this two different env vars instead of e.g. `electrum://ip:port`, `electrumssl://ip:port` or `bitcoind://ip:port`?

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -162,3 +180,76 @@ where
...
+#[async_trait]
+impl IBitcoindRpc for ElectrumClient {
+    async fn get_network(&self) -> Result<Network> {
+        let resp = fedimint_api::task::block_in_place(|| {
+            self.0.server_features().map_err(anyhow::Error::from)
+        })?;
+        Ok(match resp.genesis_hash.to_hex().as_str() {
+            "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f" => Network::Bitcoin,
+            n => panic!("Unknown Genesis Hash \"{}\"", n),
+        })
```
elsirion, line 209: I wonder if we can do better with the tools rust-bitcoin gives us? At least we could do a linear search over networks and their chain hashes iirc?

fedimint-bitcoind/src/bitcoincore_rpc.rs
dpc: I guess it's not really a blocker. Just a missing feature. :D

fedimint-api/src/bitcoin_rpc.rs
dpc: @elsirion Not sure. Then we kind of invent new URL schemas which I don't think are official. Or are they?

dpc: I consider electrum support to be a bit of a hack/wheelchairs, and maybe it's better to have a separate env? If someone meant bitcoind, it shouldn't possibly switch to electrum etc.

I don't know. Mixed feelings. One one hand two env vars are kind of meh, on the other - cramming them together also kind of meh. :D

fedimint-bitcoind/src/bitcoincore_rpc.rs
elsirion: How so? Without it we won't recognize change we send to ourselves. Only workaround seems defining an API both core and electrum client can implement (probably with quite some complexity in these impls due to mismatching APIs).

dpc: I have fixed it now, but I was thinking that - it will just crash on peg-out, so it "supports only federations that never peg-out". :D

dpc:
@elsirion I think I've fixed all the blocking issues. We're planning to do some federation setup testing tomorrow or on Wednesday, so we'll give it some proper testing then. If possible, and LGTM, we could land already - will make testing easier if it's already there.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -162,3 +182,81 @@ where
...
+    async fn submit_transaction(&self, transaction: Transaction) -> Result<()> {
+        fedimint_api::task::block_in_place(|| {
+            let mut bytes = vec![];
+            transaction
+                .consensus_encode(&mut bytes)
+                .expect("can't fail");
+            let _txid = self.0.transaction_broadcast_raw(&bytes)?;
+            Ok(())
+        })
+    }
```
dpc, line 261: WTF! Where did the call go?!

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -162,3 +184,111 @@ where
...
+
+    async fn was_transaction_confirmed_in(
+        &self,
+        transaction: &Transaction,
+        height: u64,
+    ) -> Result<bool> {
```
dpc, line 271: Please verify my thinking here.

dpc: OK. Some of the tooling tricked me and stashed my change.

fedimint-bitcoind/src/bitcoincore_rpc.rs
dpc: OK. Some of the tooling tricked me and stashed my change. I think I recovered everything, but be double suspicious of my code.

fedimint-bitcoind/src/bitcoincore_rpc.rs
elsirion: "federations that never peg-out" sounds like "write only memory" :laughing:

fedimint-bitcoind/src/lib.rs
```diff
@@ -17,6 +18,13 @@ pub mod bitcoincore_rpc;
 /// Functions may panic if if the bitcoind node is not reachable.
 #[async_trait]
 pub trait IBitcoindRpc: Debug + Send + Sync {
+    /// `true` if it's real-bitcoin (not electrum) backend and thus supports `get_block` call
+    ///
+    /// This is a bit of a workaround to support electrum.
+    fn is_bitcoind(&self) -> bool {
+        true
+    }
```
elsirion, line 26: Could we make this a function that returns an enum instead that tells us which tx confirmation query strategy to use? That would be more general. E.g. an esplora client might act like bitcoind, but certainly isn't.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -162,3 +184,111 @@ where
         })
     }
 }
+
+pub struct ElectrumClient(electrum_client::Client);
```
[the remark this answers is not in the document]
dpc: Agreed.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -162,3 +184,111 @@ where
...
+    }
+
+    async fn get_fee_rate(&self, confirmation_target: u16) -> Result<Option<Feerate>> {
+        fedimint_api::task::block_in_place(|| {
+            Ok(Some(Feerate {
+                sats_per_kvb: (self.0.estimate_fee(confirmation_target as usize)? * 100_000_000f64)
+                    .ceil() as u64,
+            }))
+        })
+    }
```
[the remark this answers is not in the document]
dpc: You're right this is slightly incorrect, but I don't think we're going to get vbytes from electrum, and these two should somewhat close to each other. If anything per kv will be slightly higher, so  federation will slightly overpay, not discounting segregated witnesses by 4.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -162,3 +184,111 @@ where
...
+            Ok(())
+        })
+    }
+
+    async fn was_transaction_confirmed_in(
+        &self,
+        transaction: &Transaction,
+        height: u64,
+    ) -> Result<bool> {
+        if self.get_block_height().await? <= height {
```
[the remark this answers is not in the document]
dpc: This call under the hood is actually `getblockcount` . So if block count is 10, then we have blocks from 0 to 9, so the user can't ask for block 10. So `10 <= 10`

Please double check me, just had my morning coffee. :D

Related: https://github.com/fedimint/fedimint/issues/1340

fedimint-bitcoind/src/lib.rs
dpc: Good idea.

fedimint-bitcoind/src/bitcoincore_rpc.rs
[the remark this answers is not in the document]
elsirion: Can we build a test for that with regtest electrs?

elsirion (approved):

fedimint-bitcoind/src/bitcoincore_rpc.rs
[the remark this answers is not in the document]
dpc: https://github.com/fedimint/fedimint/issues/1383

PR #1286, opened 2023-01-12
Unify `*Registry` types with a common macro 

dpc:
So they can maintain consistent apis (where desired).

On top of #1285. Please review that first, then click Update here. :)

PR #1364, opened 2023-01-17
Unify `ModuleRegistry*` types

dpc:
Second attempt at #1286

### DIFF
--- a/fedimint-api/src/module/registry.rs
+++ b/fedimint-api/src/module/registry.rs
@@ -4,6 +4,7 @@ use crate::core::DynDecoder;
 pub use crate::core::ModuleInstanceId;
 use crate::server::DynServerModule;
 
+/// Module Registry hold module-specific data `M` by the `ModuleInstanceId`
 #[derive(Debug, Clone)]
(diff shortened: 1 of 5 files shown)


fedimint-api/src/module/registry.rs
```diff
@@ -19,28 +20,34 @@ impl<M> From<BTreeMap<ModuleInstanceId, M>> for ModuleRegistry<M> {
...
-    /// Return the server module belonging to the module identified by the supplied `module_key`
+    /// Return the module data belonging to the module identified by the supplied `module_id`
     ///
     /// # Panics
     /// If the module isn't in the registry
-    pub fn get(&self, module_key: ModuleInstanceId) -> &M {
-        self.0
-            .get(&module_key)
-            .expect("CIs were decoded, so the module exists")
+    pub fn get_expect(&self, id: ModuleInstanceId) -> &M {
```
dpc, line 49: Seem a bit odd that `get` can panic and it might lead to bugs, but we don't want to add `expect` all over the code so we do need a method that will do it. `get_expect`... well... it does the job done.  :D

elsirion (dismissed):

elsirion (approved):

PR #1369, opened 2023-01-17
Enforce import granularity "module" via rustfmt

### LINKED ISSUE #258: Enforce import structure through `rustfmt`
Currently different editors lead to different import structures, we [should enforce one](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#Module%5C%3A). Would also be good to explore rustfmt options a bit further.

dpc:
This one seems best compromise between compact size and being merge-conflict-friendly.

Fix #258


dpc:
It's actually less massive then I expected.

dpc:
Let's land it in two weeks after things slow down a bit.

dpc:
Done already.

PR #1372, opened 2023-01-17
chrome tracing for integration tests

Maan2003:
Just as suggestion to try out.

It is much more user friendly than open telemetry for debugging.   

Instructions are in docs/telemetry.md

### DIFF
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -77,6 +77,7 @@ use rand::RngCore;
 use real::{RealBitcoinTest, RealLightningTest};
 use tokio::sync::Mutex;
 use tracing::{debug, info};
+use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
 use url::Url;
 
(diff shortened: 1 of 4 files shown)


dpc (dismissed):
This is sweet! :heart: 

We should probably make it compilation-conditional, so we don't need to add this code to our dependencies for production builds (at least by default?) and put into more places, but can done in a follow-up PRs.

Maan2003:
> and put into more places

it doesn't work for long living processes, because it doesn't live update (writes on exit)

Maan2003:
Intermittant CI failures :/

integrationtests/tests/fixtures/mod.rs
```diff
@@ -149,13 +150,25 @@ pub async fn fixtures(num_peers: u16) -> anyhow::Result<Fixtures> {
...
-                    .unwrap_or_else(|_| EnvFilter::new("info,fedimint::consensus=warn")),
-            )
-            .init();
+        if env::var_os("FEDIMINT_TRACE_CHROME").map_or(false, |x| !x.is_empty()) {
+            let (cr_layer, gaurd) = tracing_chrome::ChromeLayerBuilder::new()
+                .include_args(true)
+                .build();
+            // drop gaurd cause file to written and closed
+            // in this case file will closed after exit of program
+            std::mem::forget(gaurd);
```
dpc, line 159: How does it work? Without a guard the rest of the code there doesn't know that we're exiting? It might be that it mostly works, but it can truncate the output.

IMO, you should return `Option<Guard>` from both sides here, `Some` from here, `None` from the other one.

dpc: Oh. Given that we run all tests in parallel, I'm not even sure how useful this will be with all tracing mixed together.

A thread_local holding a `Weak<Guard>` that tests acquire and upgrade to `Arc<Guard>` (and initialize if empty) would probably do the right thing, though the problem of inter-mixing logging would still be there.

dpc: Then people can just filter the test they are interested in, and/or run them serially.

Maan2003: Yes, it is only useful in single test mode. But then `Drop` for thread local doesn't run for main thread on linux :/

dpc: That's why we use `Weak` in the thread local - so it doesn't hold any references and everything drops after last test failed.

elsirion: Do you think @dpc's approach is viable? Sounds a bit cleaner tbh. Otoh I don't mind this hack in test code :laughing: Can not running `drop` cause problems in any way? Like for example not flushing buffers?

dpc:
@elsirion Fine with me. Can be done later or never if this one works with no issues.

Issue #664, opened 2022-09-26
Lightning gateway serves multiple federations


justinmoon:
Making a new issue based inspired by https://github.com/fedimint/fedimint/issues/613#issuecomment-1257633238

What needs to happen in order to support this usecase?

elsirion:
Braindump:

* Add a hop to the route hint so that the route becomes `[gateway, federation, user]` where the user key is randomly generated by the user and the federation key is generated once by the gateway. This is important since the GW needs to be able to decrypt the onion that would be sent to the "federation" hop (but is just processed by the GW to determine from where to get the preimage). Any ideas where else to hide a few bits to select the federation welcome (hash/user pk grinding?)
* The GW needs to maintain separate e-cash wallets per federation
* "Internal routing" can now happen between federations and becomes more complex

okjodom:
> Add a hop to the route hint so that the route becomes [gateway, federation, user] where the user key is randomly generated by the user and the federation key is generated once by the gateway. This is important since the GW needs to be able to decrypt the onion that would be sent to the "federation" hop (but is just processed by the GW to determine from where to get the preimage). Any ideas where else to hide a few bits to select the federation welcome (hash/user pk grinding?)

Ack on this.

okjodom:
> The GW needs to maintain separate e-cash wallets per federation

Supported by #714 , by fact of the each registered federation getting a dedicated **GatewayActor**

okjodom:
> "Internal routing" can now happen between federations and becomes more complex
Semantics: We probably should distinguish between payments/routing by giving definitions for "internal payments", "external payments" and "self payments"

My take on it is that:

1. internal payments
- happens between two clients of the same federation
- implementation as discussed in #613 , fully scopes all interactions between clients and the federation they belong t
- gateways are NOT involved in routing

2. external payments
- happens between a client in a federation, and some other lightning entity outside this federation
- entity outside this federation might be in another federation, but it doesn't matter unless it's an "self payment" case, defined below
- gateways are involved in routing

3. self payments
- special case of external payments
- happens when a gateway serves two federations, between which two clients want to complete a transaction
- self payment logic is fully scoped within the gateway

current issue is about **external payments** and **self payments**

okjodom:
[...]

PR #1381, opened 2023-01-17
Fix CI not runing tests

dpc:
The meaning of `cargoBuild` changed at some point in crane, and was
ignoring previous overrides to run the tests instead of just doing
`cargo build`.

Couple of minor fixes along with it.

Fix #1379

PR #1382, opened 2023-01-17
Filter cln htlcs on intercept

okjodom:
Instead of processing all htlcs to gateway for processing, filter only the relevant ones, and forward all other intercepted htlcs. This rightly assumes the gateway is only serving one federation, and clients in that federation use `short_channel_id=0` in the route hint

**TODO**: Expand this filtering logic to work with other short channel ids, as part of completing #664

### DIFF
--- a/gateway/ln-gateway/src/cln.rs
+++ b/gateway/ln-gateway/src/cln.rs
@@ -8,7 +8,6 @@ use cln_rpc::{model, ClnRpc, Request, Response};
 use fedimint_api::Amount;
 use fedimint_server::modules::ln::contracts::Preimage;
 use serde::{Deserialize, Deserializer, Serialize};
-use serde_json::json;
 use tokio::io::{stdin, stdout};
 use tokio::sync::Mutex;
 use tracing::{debug, error, instrument};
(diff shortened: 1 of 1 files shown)


dpc:
Cherry-picked into https://github.com/fedimint/fedimint/pull/1381 to see if helps.

okjodom:
> Cherry-picked into #1381 to see if helps.

@justinmoon, @dpc, you can cherry pick this latest, or review and merge. I figured the correct bolt 4 failure message code to use in case we fail to buy a preimage. Now the node sends this message back to LN in case it needs to gracefully cancel HTLCs

```
2023-01-18T01:40:51.301Z INFO    020901ddfafcad5adc88757f9a09f80f8d1660f29f3461677413df191143fe9623-chan#1: htlc 0 failed from 0th node with code 0x1639 (INVALID 5689)
2023-01-18T01:40:51.314Z UNUSUAL plugin-pay: cmd 234 partid 0: Intermediate node 020901ddfafcad5adc88757f9a09f80f8d1660f29f3461677413df191143fe9623 reported strange error code 1639 (UPDATE|1593)
2023-01-18T01:40:51.314Z INFO    plugin-pay: cmd 234 partid 0: failed: INVALID 5689 (reply from remote)
```

gateway/ln-gateway/src/cln.rs
```diff
@@ -133,22 +132,51 @@ impl LnRpc for Mutex<cln_rpc::ClnRpc> {
...
+/// BOLT 4: https://github.com/lightning/bolts/blob/master/04-onion-routing.md#failure-messages
+/// 16399 error code reports unknown payment details.
+///
+/// TODO: We should probably use a more specific error code based on htlc processing fail reason
+fn htlc_processing_failure() -> serde_json::Value {
+    serde_json::json!({
+        "result": "fail",
+        "failure_message": "1639"
+    })
+}
```
justinmoon, line 144: Can this be a `const`?

justinmoon (approved):
LGTM

gateway/ln-gateway/src/cln.rs
okjodom: Tried a const but the strings need to be statics at best. Didn't know how to make it a const

PR #1281, opened 2023-01-11
Add and verify consensus config hash

jkitman:
Makes consensus and client configs implement `Encodable` so they can be deterministically hashed and validated.

Adds the consensus hash and an API endpoint for retrieving it with the client config (next step will be to save and return the signed hash of the client config).

Fixes #950, will be used to solve #1262

PR #1385, opened 2023-01-18
Revert "Add and verify consensus config hash (#1281)"

dpc:
This reverts commit 7299e5e6ff22991975164654e321e1477fd18754.

We had cli tests disabled in CI, and it seems to affect both tmuxinator and cli tests.

Let's revert while we're trying to get it under control.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "0.7.19"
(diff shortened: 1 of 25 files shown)


justinmoon (approved):

dpc:
@jkitman Sorry. Bit of my fault for not having tests running. We can land again after we get things green again.

justinmoon:
As context, tmuxinator (i didn't test cli-tests.sh but assume it was same) would just get stuck waiting for consensus on the config. Couldn't use the federation.

PR #1386, opened 2023-01-18
Receive HTLC to default fed

okjodom:
Since we don't fully support multiple federations, we need to *guess* the intended destination actor for every incoming payment received via intercepted htlc. This pr removes any reference of `default_federation` in gateway configs, and instead, picks a listed actor in the gateway, then attempts to settle payment into the federation represented by this actor.

#1337 proposes a long term fix for this issue, by actively streaming intercepted htlcs into a subscribing actor, therefore avoiding the need to select actors for every intercepted htlc

See #664 for details on supporting multiple federations

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -220,9 +220,11 @@ impl LnGateway {
 
         // FIXME: Issue 664: We should avoid having a special reference to a federation
         // all requests, including `ReceivePaymentPayload`, should contain the federation id
-        // TODO: Parse federation id from routing hint in htlc_accepted message
-        self.select_actor(self.config.default_federation.clone())
-            .await?
+        //
+        // We use a random federation as the default (works because we only have one federation registered)
+        //
+        // TODO: Use subscribe intercept htlc streams to avoid actor selection with every intercepted htlc!
+        self.actors.lock().await.values().collect::<Vec<_>>()[0]
             .buy_preimage_internal(&payment_hash, &invoice_amount)
             .await
     }
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -96,8 +96,6 @@ async fn main() {
                     announce_address,
                     // TODO: Generate a strong random password
                     password: source_password(cli.rpcpassword),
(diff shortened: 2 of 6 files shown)


dpc (approved):

justinmoon:
So this code assumes that there is exactly 1 gateway actor. How are we sure this will be true?

It seems that [filtering htlcs](https://github.com/fedimint/fedimint/pull/1382/files#diff-4dc1d69fdaa69b89bc846149cfaf8ff5b802853d13c5049362d5c934258cab50R153-R179) ensures this code **never** executes before a gateway actor has been created.

But how do we know we will never have 2 gateway actors? Perhaps `gateway-cli connect-fed` should refuse adding a second actor until we have proper multi-federation support?

dpc:
> But how do we know we will never have 2 gateway actors? P

If I'm reading it right, if we have 2, we will arbitrarily route to first one on the list.

This is just a temporary fix, until proper routing is implemented.

okjodom:
> > But how do we know we will never have 2 gateway actors? P
> 
> If I'm reading it right, if we have 2, we will arbitrarily route to first one on the list.
> 
> This is just a temporary fix, until proper routing is implemented.

Keen getting a permanent fix shortly.

Now if we had two gateway actors, and selected the wrong one (we select at random because hashmaps are not ordered), we would attempt to settle htlc to it but will not find the corresponding incoming transaction on the federation. Gateway will fail the htlc

gateway/ln-gateway/src/lib.rs
```diff
@@ -220,9 +220,11 @@ impl LnGateway {
...
         // FIXME: Issue 664: We should avoid having a special reference to a federation
         // all requests, including `ReceivePaymentPayload`, should contain the federation id
-        // TODO: Parse federation id from routing hint in htlc_accepted message
-        self.select_actor(self.config.default_federation.clone())
-            .await?
+        //
+        // We use a random federation as the default (works because we only have one federation registered)
+        //
+        // TODO: Use subscribe intercept htlc streams to avoid actor selection with every intercepted htlc!
+        self.actors.lock().await.values().collect::<Vec<_>>()[0]
```
elsirion, line 227: nit: We don't need to `collect` here, `.next().expect("…")` would have been enough.

PR #1388, opened 2023-01-18
Add and verify consensus config hash

jkitman:
Add back this commit, fixes an issue with passing params to the API (also API was not returning the error).


client/client-lib/src/api.rs
```diff
@@ -283,7 +283,7 @@ pub trait FederationApiExt: IFederationApi {
                             for (failed_peer, error) in failed {
                                 member_errors.insert(failed_peer, error);
                             }
-                            return Err(FederationError(BTreeMap::new()));
+                            return Err(FederationError(member_errors));
```
dpc, line 286: 💯

dpc (approved):

client/client-lib/src/api.rs
```diff
@@ -381,6 +381,9 @@ pub trait GlobalFederationApi {
         timeout: Duration,
         decoders: &ModuleDecoderRegistry,
     ) -> OutputOutcomeResult<R>;
+
+    /// Fetch verifiable client configuration info
+    async fn download_client_config(&self) -> FederationResult<ConfigResponse>;
```
justinmoon, line 386: We should remove `get_client_config()`. It doesn't work anymore.

PR #1394, opened 2023-01-18
Add federation info to cli info command

### LINKED ISSUE #1384: `fedimint-cli info` should include `federation_id`

asifshiraz:
Fixed #1384

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -6,7 +6,7 @@ use std::path::PathBuf;
 use std::process::exit;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


justinmoon (approved):

client/cli/src/main.rs
```diff
@@ -83,7 +83,9 @@ enum CliOutput {
     },
 
     Info {
-        network: String,
+        federation_id: String,
```
elsirion, line 86: Can we impl `Serialize` for the federation id struct such that we don't have to manually turn it into a `String` here?

Issue #1342, opened 2023-01-16
Renames around `OutputOutcome` in mint module


mxz42:
#1320 

> @mxz42 I've just noticed that mints OutputOutcome is called ... OutputOutcome which creates a confusion. Seems like for all other stuff we name them LightningFoo, so at least for consistency we should fix it as well.

 @dpc  might be a bit complicated as `MintOutputOutcome` already exists. and quite a few other similar output types.

I think the `MintOutputConfirmation` can be renamed to the `MintConsensusItem` as it is used as such in the `ServerModule` impl associated type `OutputOutcome`. 

`MintOutputOutcome` is currently an `Option<OutputOutcome>` ... 

Do you have thoughts on making this more consistent? 
```
/modules/fedimint-mint/src (master *$) | rg '^pub struct' lib.rs
71:pub struct Mint {
80:pub struct MintOutputConfirmation {                     --> MintConsensusItem 
90:pub struct OutputConfirmationSignatures(
98:pub struct OutputOutcome(pub TieredMulti<tbs::BlindedSignature>);
113:pub struct Note(pub Nonce, pub tbs::Signature);
121:pub struct Nonce(pub secp256k1_zkp::XOnlyPublicKey);
131:pub struct BlindNonce(pub tbs::BlindedMessage);
134:pub struct VerifiedNotes {
139:pub struct MintGen;
299:pub struct MintGenParams {
311:pub struct MintInput(pub TieredMulti<Note>);
323:pub struct MintOutput(pub TieredMulti<BlindNonce>);
333:pub struct MintOutputOutcome(pub Option<OutputOutcome>);
1081:pub struct MintShareErrors(pub Vec<(PeerId, PeerErrorType)>);

/modules/fedimint-ln/src (master *$) | rg '^pub struct' lib.rs
87:pub struct Lightning {
92:pub struct LightningInput {
171:pub struct ContractOutput {
177:pub struct ContractAccount {
207:pub struct LightningGateway {
218:pub struct LightningConsensusItem {
230:pub struct LightningVerificationCache;
233:pub struct LightningGen;

/modules/fedimint-wallet/src (master *$) | rg '^pub struct' lib.rs
104:pub struct RoundConsensusItem {
111:pub struct PegOutSignatureItem {
117:pub struct RoundConsensus {
124:pub struct Wallet {
131:pub struct SpendableUTXO {
139:pub struct PendingTransaction {
163:pub struct UnsignedTransaction {
193:pub struct PegOutFees {
[...]

PR #1396, opened 2023-01-18
Renames in mint module

mxz42:
part of #1342

The first commit does renames agreed upon in #1342 .

The second commit renames `OutputConfirmationSignatures` to `MintOutputSignatureShare`. according to the comment above the struct, it will contain (at some point) only the `tbs::BlindedSignatureShare` and the `tbs::BlindedMessage` will be remembered by the mint. If you don't agree I remove the second commit.

```rust
// FIXME: optimize out blinded msg by making the mint remember it
/// Blind signature share from one Federation peer for a single [`MintOutput`]
#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
pub struct MintOutputSignatureShare(
    pub TieredMulti<(tbs::BlindedMessage, tbs::BlindedSignatureShare)>,
);
```

thoughts @dpc @elsirion 

--- 

I rebased this branch on #1389, otherwise some tests failed with `No gateway` or something similar.

### DIFF
--- a/modules/fedimint-mint/src/lib.rs
+++ b/modules/fedimint-mint/src/lib.rs
@@ -81,13 +81,13 @@ pub struct MintConsensusItem {
     /// Reference to a Federation Transaction containing an [`MintOutput`] with `BlindNonce`s the signatures` are for
     pub out_point: OutPoint,
     /// (Partial) signatures
-    pub signatures: OutputConfirmationSignatures,
+    pub signatures: MintOutputSignatureShare,
 }
 
 // FIXME: optimize out blinded msg by making the mint remember it
 /// Blind signature share from one Federation peer for a single [`MintOutput`]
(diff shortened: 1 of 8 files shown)


dpc:
Probably want to rebase.

mxz42:
you mean rebase on master or should I just wait until #1389 is merged?

dpc:
image

Seems like much. Do you need all of them?

modules/fedimint-mint/src/lib.rs
```diff
@@ -359,7 +359,7 @@ impl ServerModule for Mint {
     type Input = MintInput;
     type Output = MintOutput;
     type OutputOutcome = MintOutputOutcome;
-    type ConsensusItem = MintOutputConfirmation;
+    type ConsensusItem = MintConsensusItem;
```
elsirion, line 362: :+1:

elsirion (approved):

Issue #1353, opened 2023-01-16
Stop using the consensus DB for submitted transactions


elsirion:
Using the DB we also use for consensus to keep unconfirmed transactions is a potential security risk in combination with other bugs (e.g. #1316). To be on the safe side we should migrate to an in-memory unconfirmed transaction cache that:

1. Allows async submission and notification (probably via a channel)
2. Keeps transactions till they are confirmed in an epoch
3. Prevents re-submission of already confirmed transactions

dpc:
Question: If the user submited a tx to `threshold_required` members, then one of them crashed and lost it (as it's not persisted now), is it still going to get confirmed? If not, then clients should re-submit, or we should implement fetching the "mempool" from other peers on start.

elsirion:
> If the user submited a tx to `threshold_required` members, then one of them crashed and lost it (as it's not persisted now), is it still going to get confirmed?

Technically the one that crashed counts as malicious for security analysis, so assuming no one else is malicious it will get confirmed. I others are malicious consensus might stall completely (if f=1 for example).

> If not, then clients should re-submit

Clients need to handle tx that simply never happened (re-submit) or get rejected (try to re-claim inputs).

PR #1402, opened 2023-01-19
Prevent transaction replays

### LINKED ISSUE #1347: Prevent transaction replays
For the mint use case this hasn't been relevant so far, but other modules (e.g. account systems) would benefit from a no-replay guarantee.

elsirion:
Fixes #1347

### DIFF
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -530,6 +530,16 @@ impl FedimintConsensus {
 
         let tx_hash = transaction.tx_hash();
 
+        // Prevent transaction replays
+        if dbtx
+            .get_value(&AcceptedTransactionKey(tx_hash))
+            .await
+            .expect("DB error")
+            .is_some()
+        {
+            return Err(TransactionSubmissionError::TransactionReplayError(tx_hash));
+        }
+
         let mut pub_keys = Vec::new();
         for input in transaction.inputs.iter() {
             let meta = self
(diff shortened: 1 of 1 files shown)


jkitman (approved):
Should work, although it'd be nice to have a integration test for this.

fedimint-server/src/consensus/mod.rs
```diff
@@ -530,6 +530,16 @@ impl FedimintConsensus {
...
         let tx_hash = transaction.tx_hash();
 
+        // Prevent transaction replays
+        if dbtx
+            .get_value(&AcceptedTransactionKey(tx_hash))
+            .await
+            .expect("DB error")
+            .is_some()
+        {
+            return Err(TransactionSubmissionError::TransactionReplayError(tx_hash));
```
dpc, line 540: Should it be an error? Maybe just an OK and do nothing? In that case the call is idempotent.

dpc: I wasn't aware we are already storing all accepted txes.

elsirion: > Should it be an error? Maybe just an OK and do nothing? In that case the call is idempotent.

I think that would mess with your recovery code since the tx would not be marked as invalid in the epoch history.

The alternative I see is checking outside this fn, right before it is called. That way we don't write a tx failed to the DB but can still mark it as invalid in the epoch history. That actually sounds better to me.

elsirion:
Will get back to this in a bit

fedimint-server/src/consensus/mod.rs
dpc: Oh. This is already on processing consensus. OK then.

I was thinking that we could do this check on submission and then return `Ok`.

elsirion: Right, on submission we want a similar check, that's part of #1353. But in case guardians are malicious we definitely need the consensus check here.

PR #1404, opened 2023-01-19
fix: implement variantless enum encode/decode derive

### LINKED ISSUE #1398: Derive encodable doesn't work with `enum NotPossible {}`.
Can be useful for modules that don't use `Input/Ouput/etc.`

elsirion:
fixes #1398

### DIFF
--- a/fedimint-derive/src/lib.rs
+++ b/fedimint-derive/src/lib.rs
@@ -121,7 +121,16 @@ pub fn derive_encodable(input: TokenStream) -> TokenStream {
             }
         }
         syn::Data::Enum(DataEnum { variants, .. }) => {
-            let match_arms = variants.iter().enumerate().map(|(variant_idx, variant)| {
+            if variants.is_empty() {
+                quote! {
+                    impl Encodable for #ident {
+                        fn consensus_encode<W: std::io::Write>(&self, _writer: &mut W) -> std::result::Result<usize, std::io::Error> {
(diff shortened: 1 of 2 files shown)


fedimint-derive/src/lib.rs
```diff
@@ -121,7 +121,16 @@ pub fn derive_encodable(input: TokenStream) -> TokenStream {
             }
         }
         syn::Data::Enum(DataEnum { variants, .. }) => {
-            let match_arms = variants.iter().enumerate().map(|(variant_idx, variant)| {
+            if variants.is_empty() {
+                quote! {
+                    impl Encodable for #ident {
+                        fn consensus_encode<W: std::io::Write>(&self, _writer: &mut W) -> std::result::Result<usize, std::io::Error> {
+                            unreachable!("Enum without variants can't exist")
```
Maan2003, line 128:
```suggestion
                            match self { }
```

dpc: LGTM, but this one worth fixing before landing.

elsirion: Done. Most interestingly I had to dereference the enum or [I would get this error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021):

Screenshot 2023-01-19 at 14-05-56 Rust Playground

dpc (approved):

Issue #1093, opened 2022-12-10
Get rid of "coin" and "token" terminology


justinmoon:
A couple months ago we decided to get rid of "coin" and "token" terminology in favor of "note". But now we use all 3 terms in the codebase which is even worse. I imagine this could be very confusing to a new contributor. Let's finish the job!

elsirion:
![standards_2x](https://user-images.githubusercontent.com/84478054/206826698-732a8516-5579-497a-95fa-e8a4eae5a6a0.png)
https://xkcd.com/927/

asifshiraz:
Anyone working on this? Seems low priority but might be good to see all the coin related logic while doing the renames, so if it's open, I can take.

dpc:
@asifshiraz Please go ahead!

dpc:
Somewhat related, but I was wondering if we could synchronize with: https://mobile.twitter.com/callebtc/status/1614372111878848516

asifshiraz:
I noticed a comment https://github.com/fedimint/fedimint/blob/master/client/cli/src/main.rs#L228 which says "Issue tokens in exchange for a peg-in proof (not yet implemented, just creates notes)"
Seems like tokens was intended to be a separate concept. Would you still like a straight away replacement of this term with notes, or is there more nuance to it.

justinmoon:
@asifshiraz AFAIK there's no difference in these terms. That's a confusing doc string. I'd say just change it to ` /// Issue notes in exchange for a peg-in proof`

elsirion:
That comment is [an artifact of Fedimint's ancient pre-history](https://github.com/fedimint/fedimint/commit/6ac914adcef80d8a386f0a6c1cec870219d514d3#diff-f3a1ebcd411a317d0f272f6686eb1d3257391202e04a803e5d6eafa7c3055eb8R20) :laughing:. Back in these days the command actually created e-cash notes out of thin air (see that it only took an amount argument) and the naming was inconsistent (tokens=coins=notes).

https://github.com/fedimint/fedimint/blob/6ac914adcef80d8a386f0a6c1cec870219d514d3/mint-client/src/main.rs#L58-L65

PR #1407, opened 2023-01-19
Change coin terminology to notes in mint and client modules

asifshiraz:
#1093


justinmoon:
For future reference, it's ok to push up a PR that's still failing. You can mark it as "draft" if it's not ready for review. And you can force-push to it re-run CI.

PR #1411, opened 2023-01-19
Change coin terminology to notes in mint and client modules

asifshiraz:


asifshiraz:
Something wrong with the pipeline I guess. Seems stuck.

PR #1419, opened 2023-01-19
Hide user+password in serialized URLs in UI

### LINKED ISSUE #1403: Do not show the `user:pass@` from bitcoind node in the setup UI
This is a subset of https://github.com/fedimint/fedimint/issues/1298 but important and urgent

elsirion:
Fixes #1403

### DIFF
--- a/fedimint-core/src/util.rs
+++ b/fedimint-core/src/util.rs
@@ -1,0 +1,87 @@
+use std::borrow::Cow;
+use std::fmt::{Debug, Display, Formatter};
+
+use url::Url;
+
+// TODO: make fully RFC1738 conformant
+/// Wrapper for `Url` that only prints the scheme, domain, port and path portion of a `Url` in its
+/// `Display` implementation. This is useful to hide private information like user names and
+/// passwords in logs or UIs.
+///
+/// The output is not fully RFC1738 conformant but good enough for our current purposes.
+pub struct SanitizedUrl<'a>(Cow<'a, Url>);
+
+impl<'a> SanitizedUrl<'a> {
+    pub fn new_owned(url: Url) -> SanitizedUrl<'static> {
+        SanitizedUrl(Cow::Owned(url))
+    }
+
+    pub fn new_borrowed(url: &'a Url) -> SanitizedUrl<'a> {
+        SanitizedUrl(Cow::Borrowed(url))
+    }
+}
+
+impl<'a> Display for SanitizedUrl<'a> {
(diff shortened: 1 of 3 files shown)


fedimint-core/src/util.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    pub fn new_borrowed(url: &'a Url) -> SanitizedUrl<'a> {
+        SanitizedUrl(Cow::Borrowed(url))
+    }
+}
+
+impl<'a> Display for SanitizedUrl<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}://", self.0.scheme())?;
+
```
dpc, line 27: Might be better to print `REDACTED` or something when user or password are set.

dpc (approved):

fedimint-core/src/util.rs
```diff
@@ -0,0 +1,87 @@
...
+        let test_cases = vec![
+            ("http://1.2.3.4:80/foo", "http://1.2.3.4/foo", "SanitizedUrl(http://1.2.3.4/foo, has_password=false, has_username=false)"),
+            ("http://1.2.3.4:81/foo", "http://1.2.3.4:81/foo", "SanitizedUrl(http://1.2.3.4:81/foo, has_password=false, has_username=false)"),
+            ("fedimint://1.2.3.4:1000/foo", "fedimint://1.2.3.4:1000/foo", "SanitizedUrl(fedimint://1.2.3.4:1000/foo, has_password=false, has_username=false)"),
+            ("fedimint://foo:bar@domain.com:1000/foo", "fedimint://domain.com:1000/foo", "SanitizedUrl(fedimint://domain.com:1000/foo, has_password=true, has_username=true)"),
+            ("fedimint://foo@1.2.3.4:1000/foo", "fedimint://1.2.3.4:1000/foo", "SanitizedUrl(fedimint://1.2.3.4:1000/foo, has_password=false, has_username=true)"),
+        ];
+
+        for (url_str, sanitized_display_expected, sanitized_debug_expected) in test_cases {
+            dbg!(&url_str);
```
elsirion, line 70: Please complain about such `dbg!` statements, that was left here by accident.

dpc: I haven't noticed, but if it's in tests, then I guess it's fine? :D


Maybe we should have a pre-commit check looking for `dbg!`. I leave them accidentally all the time. :D

elsirion: I ran the tests after the merge and it just looked ugly :laughing: thx for adding the check!

PR #1420, opened 2023-01-19
Don't hardcode module gens in the ui

dpc:
A papercut when working on your own module.


elsirion (approved):
We still don't support params for any additional modules, but this at least makes it easier to wire new modules into fedimint.

PR #1421, opened 2023-01-19
Don't flicker the setup UI after dkg is done.

### LINKED ISSUE #1406: After creating the federation succesfully the UI flickers
After successfully creating federation the UI seems to be flickering (like the page is continuously restarting) and Copy button does seem to work (copying manually works thought).

dpc:
Fix #1406

### DIFF
--- a/fedimintd/templates/run.html
+++ b/fedimintd/templates/run.html
@@ -1,7 +1,11 @@
 {% extends "base.html" %}
 
 {% block title %} Fedimint {% endblock %}
-{% block head %} <meta http-equiv="refresh" content="5" /> {% endblock %}
+{% block head %}
+{% if !has_connection_string %}
+ <meta http-equiv="refresh" content="5" />
+{% endif %}
+{% endblock %}
 
 {% block content %}
 <div class="container text-center mt-5 p-3">


dpc:
Looks to me like it's working.

justinmoon (approved):

fedimintd/templates/run.html
```diff
@@ -1,7 +1,6 @@
 {% extends "base.html" %}
 
 {% block title %} Fedimint {% endblock %}
-{% block head %} <meta http-equiv="refresh" content="5" /> {% endblock %}
```
elsirion, line 4: Doesn't this need to be conditional on `has_connection_string` instead of removing it outright? How will we get to the QR after the DKG has finished? If you tested with 1-of-1 latency might have hidden that problem.

elsirion (changes requested):

fedimintd/templates/run.html
dpc: I think you're right in the principle, and I didn't fully understand what this code is doing. But it also looks like we actually not displaying this page until dkg finished? @justinmoon

justinmoon:
I'll review this soon

fedimintd/templates/run.html
elsirion: In that case we should remove `has_connection_string` and the branch.

dpc: We should make it show something again while the dkg is running. But that could be in another PR. I've make it conditional now. :)

elsirion (approved):

dpc:
@justinmoon Took care of it in #1429

PR #1422, opened 2023-01-19
Graceful shutdown fixes

### LINKED ISSUE #1418: Does not terminate (hangs) on panic
On error I've spotted in https://github.com/fedimint/fedimint/issues/1417#issue-1549749091 the fedimint killed all the peer connections, but then did not stop, preventing `systemd` from starting it again.

elsirion:
- [x] Move main task into task group
- [x] Limit time `shutdown` can take

Fixes #1418

### DIFF
--- a/fedimint-api/src/task.rs
+++ b/fedimint-api/src/task.rs
@@ -1,11 +1,13 @@
 #![cfg_attr(target_family = "wasm", allow(dead_code))]
+
+use std::collections::VecDeque;
 use std::future::Future;
+use std::pin::Pin;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::SeqCst;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
-use anyhow::anyhow;
 use futures::future::BoxFuture;
 use futures::lock::Mutex;
 pub use imp::*;
@@ -14,8 +16,8 @@ use thiserror::Error;
 use tokio::sync::oneshot;
 #[cfg(not(target_family = "wasm"))]
 use tokio::task::JoinHandle;
-use tracing::debug;
-use tracing::info;
+use tracing::{debug, warn};
+use tracing::{error, info};
 #[cfg(target_family = "wasm")]
 type JoinHandle<T> = futures::future::Ready<anyhow::Result<T>>;
 
@@ -29,7 +31,7 @@ struct TaskGroupInner {
     is_shutting_down: AtomicBool,
     #[allow(clippy::type_complexity)]
     on_shutdown: Mutex<Vec<Box<dyn FnOnce() -> BoxFuture<'static, ()> + Send + 'static>>>,
-    join: Mutex<Vec<(String, JoinHandle<()>)>>,
+    join: Mutex<VecDeque<(String, JoinHandle<()>)>>,
 }
 
 impl TaskGroupInner {
@@ -107,9 +109,12 @@ impl TaskGroup {
         self.inner.shutdown().await
     }
 
-    pub async fn shutdown_join_all(self) -> anyhow::Result<()> {
+    pub async fn shutdown_join_all(
+        self,
+        join_timeout: Option<Duration>,
+    ) -> Result<(), anyhow::Error> {
         self.shutdown().await;
-        self.join_all().await
+        self.join_all(join_timeout).await
     }
 
     #[cfg(not(target_family = "wasm"))]
@@ -168,7 +173,7 @@ impl TaskGroup {
         if let Some(handle) = self::imp::spawn(async move {
             f(handle).await;
         }) {
-            self.inner.join.lock().await.push((name, handle));
+            self.inner.join.lock().await.push_back((name, handle));
         }
         guard.completed = true;
     }
@@ -192,7 +197,7 @@ impl TaskGroup {
         if let Some(handle) = self::imp::spawn_local(async move {
             f(handle).await;
         }) {
-            self.inner.join.lock().await.push((name, handle));
+            self.inner.join.lock().await.push_back((name, handle));
         }
         guard.completed = true;
     }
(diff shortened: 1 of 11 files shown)


fedimintd/src/bin/main.rs
```diff
@@ -114,11 +153,6 @@ async fn main() -> anyhow::Result<()> {
     let key = get_key(opts.password, salt_path)?;
     let cfg = read_server_configs(&key, opts.data_dir.clone())?;
 
-    let local_task_set = tokio::task::LocalSet::new();
```
elsirion, line 117: I'm not sure what the `local_task_set` is good for. Doesn't seem to get used.

fedimintd/src/bin/main.rs
```diff
@@ -72,7 +75,43 @@ async fn main() -> anyhow::Result<()> {
...
 
-    let mut task_group = TaskGroup::new();
+    let mut root_task_group = TaskGroup::new();
+    root_task_group.install_kill_handler();
+
+    let task_group = root_task_group.clone();
+    let (main_shutdown_tx, main_shutdown_rx) = tokio::sync::oneshot::channel::<()>();
+
+    root_task_group
+        .spawn_local("main", move |_task_handle| async move {
```
dpc, line 85: IIRC you can get the `TaskGroup` from the `TaskHandle` to avoid clone.

fedimintd/src/bin/main.rs
```diff
@@ -72,7 +75,43 @@ async fn main() -> anyhow::Result<()> {
...
+            match run(opts, task_group).await {
+                Ok(()) => {
+                    info!("Shutting down");
+                }
+                Err(e) => {
+                    error!(?e, "Main task returned error, shutting down");
+                }
+            }
+
+            main_shutdown_tx
```
dpc, line 95: Seems like we should `task_group.shutdown();`here. (If we are out of run, we want everything else to finish).

dpc: So we wouldn't need  `main_shutdown_rx.await` (or that `oneshot` channel at all)` and we could just do `root_task_group.join_all().await`

fedimintd/src/bin/main.rs
```diff
@@ -72,7 +75,43 @@ async fn main() -> anyhow::Result<()> {
...
+                .send(())
+                .expect("Could not send shutdown signal");
+        })
+        .await;
+
+    main_shutdown_rx.await.expect("Main task panicked");
+
+    root_task_group.shutdown().await;
```
dpc, line 103: As mentioned above  - this should be `join_all`, and we don't want limit the time out `shutdown()` (if finishes right away), but the time of `join_all` **after** `is_shutting_down` flag was set.

fedimintd/src/bin/main.rs
dpc: Oh damn. This has something to do with `consensus` not being `Send` or something, and the fact that we need to `spawn_local` it. This was just creating some set of threads for it to run in a non-send mode(?), then "enter" it, and then check the results from it. Bleh.

fedimintd/src/bin/main.rs
dpc: Oh... This is bad.

Possibly we should have a `while` loop there and lock, pop one, wait one it, until we don't have any more?

dpc: I wouldn't be surprised if this wasn't one of the root issues. Good catch.

Who wrote it? What was he thinking. :D

elsirion: I'll take a look tomorrow with a fresh mind. The while loop seems ok, we'd want a `VecDequeue` though to pop from the opposite side than where we push. Since we don't seem to do random access we might as well use an unbounded channel.

fedimintd/src/bin/main.rs
elsirion: The thing is, it was never used. The only side effect was some guard that was spawned. I removed it and it didn't seem any different, but I did have to run the future in a `spawn_local` anyway.

fedimintd/src/bin/main.rs
elsirion: I didn't find a way, but not too bad imo, just an `Arc::clone`.

fedimintd/src/bin/main.rs
elsirion: I ended up making it a `VecDequeue` since the channel would have meant two member vars and a mutex around the receiver still. We should come back and refactor the `TaskGroup` such that handles can spawn, `TaskGroup` is not clone anymore and is the only thing that can `join_all`. That way consuming `self` would actually make sense (when it's just an `Arc` that can be cloned consuming self means ~nothing).

fedimintd/src/bin/main.rs
elsirion: Lol, it does have side effects …

fedimint-api/src/task.rs
```diff
@@ -221,12 +224,35 @@ impl TaskGroup {
...
-    pub async fn join_all(self) -> anyhow::Result<()> {
-        for (name, join) in self.inner.join.lock().await.drain(..) {
+    pub async fn join_all(self, join_timeout: Option<Duration>) -> anyhow::Result<()> {
+        while let Some((name, join)) = self.inner.join.lock().await.pop_front() {
             debug!("Waiting for {name} task to finish");
-            join.await
-                .map_err(|e| anyhow!("Thread {name} panicked with: {e}"))?;
-            debug!("{name} task finished.");
+
+            #[cfg(not(target_family = "wasm"))]
```
dpc, line 231: I would not bother with timeouts here.

Instead, I would go for:

```
impl TaskGroup {
  fn install_shutdown_deadline_terminator(&self) {
      let s_clone = self.clone();

      task::spawn(|| async { // standalone non-task-group thread
          // wait for shutdown flag being set
          while !s_clone.is_shutting_down() {
                sleep(1).await; // could be waiting for the shutting down callback or something instead of polling, but whatever
          }

          sleep(30).await; // give em some time
          

          if !s_clone.had_successful_shutdown() {
              std::process:exit(-10); // party is over
           }
      ||);
  }
}

dpc:
Looks reasonable, but it is all kind of tricky, and I'd have to sit in front of an editor for 30 minutes - 1h to really get an educated opinion. :D

dpc (approved):

elsirion:
Tested it with a panic in consensus and it works as intended: https://gist.github.com/elsirion/7aae29939c68e5d9df6c0a785add5d85

We should investigate why certain tasks are not shutting down by themselves, but that's out of scope for this PR.

PR #1352, opened 2023-01-16
WIP: DB commit retry

elsirion:
Alternative, more generic attempt at #1351, but abandoned for now because of lifetime issues. If anyone has some ideas how to fix please lmk.

PR #1423, opened 2023-01-19
Lifetimes for autocommit feature in db

mxz42:
generic autocommit version for fix in pr #1351 that fixed issue #1316 

Builds on top of #1352 

still needs some work now to integrate, but the function works now _I think_.

### DIFF
--- a/fedimint-api/src/db/mod.rs
+++ b/fedimint-api/src/db/mod.rs
@@ -1,5 +1,7 @@
 use std::error::Error;
 use std::fmt::Debug;
+use std::future::Future;
+use std::pin::Pin;
 use std::sync::Arc;
 
 use anyhow::Result;
@@ -58,6 +60,29 @@ struct DatabaseInner<Db: IDatabase + ?Sized> {
     db: Db,
 }
 
+/// Error returned when the autocommit function fails
+#[derive(Debug, Error)]
+pub enum AutocommitError<E> {
+    /// Committing the transaction failed too many times, giving up
+    CommitFailed {
+        /// Number of retries
+        retries: usize,
+        /// Last error on commit
+        last_error: anyhow::Error,
+    },
+    /// Error returned by the closure provided to `autocommit`. If returned no commit was attempted
+    /// in that round
+    ClosureError {
+        /// Retry on which the closure returned an error
+        ///
+        /// Values other than 0 typically indicate a logic error since the closure given to
+        /// `autocommit` should not have side effects and thus keep succeeding if it succeeded once.
+        retries: usize,
+        /// Error returned by the closure
+        error: E,
+    },
+}
+
 impl Database {
     pub fn new(db: impl IDatabase + 'static, module_decoders: ModuleDecoderRegistry) -> Self {
         let inner = DatabaseInner {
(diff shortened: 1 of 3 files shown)


fedimint-api/src/db/mod.rs
```diff
@@ -88,6 +113,63 @@ impl Database {
...
+    /// Runs a closure with a reference to a database transaction and tries to commit the
+    /// transaction if the closure returns `Ok` and rolls it back otherwise. If committing fails the
+    /// closure is run again for up to `retries` times. If retries is `None` it will run
+    /// `usize::MAX` times which is close enough to infinite times.
+    ///
+    /// The closure `tx_fn` provided should not have side effects outside of the database
+    /// transaction provided, or if it does these should be idempotent, since the closure might be
+    /// run multiple times.
+    ///
+    /// # Lifetime Parameters
```
dpc, line 126: :+1:

elsirion:
It's probably best to remove [11d2e05](https://github.com/fedimint/fedimint/pull/1423/commits/11d2e051211e0d122a37e5f3e24835c068b78563) and squash the remaining two and then add another one that uses the function in `process_consensus_outcome`.

fedimint-api/src/db/mod.rs
```diff
@@ -71,6 +96,64 @@ impl Database {
     pub async fn begin_transaction(&self) -> DatabaseTransaction {
         DatabaseTransaction::new(self.0.db.begin_transaction().await, &self.0.module_decoders)
     }
+
+    /// Runs a closure with a reference to a database transaction and tries to commit the
+    /// transaction if the closure returns `Ok` and rolls it back otherwise. If committing fails the
+    /// closure is run again for up to `retries` times. If retries is `None` it will run
+    /// `usize::MAX` times which is close enough to infinite times.
```
elsirion, line 103: running `usize::MAX` times was true at one point, but isn't anymore. We should probably make it true again since after `usize::MAX` rounds there will be an overflow/panic in debug mode and a silent overflow (afaik) in release mode.

fedimint-api/src/db/mod.rs
```diff
@@ -71,6 +96,64 @@ impl Database {
...
+                            // try again
+                        }
+                    }
+                }
+                Err(e) => {
+                    return Err(AutocommitError::ClosureError { retries, error: e });
+                }
+            };
+
+            retries += 1;
```
elsirion, line 154: This might overflow after `usize::MAX` rounds. For `usize=u32` that could potentially happen (1ms per iteration would mean it happens after 50 days), for `usize=u64` probably not. Then again, if your federation is stuck for so long trying to commit a DB tx, do you really care?

elsirion (approved):
Love it! What do you think about the overflow consideration @dpc?

dpc:
Use u64 or checking_add and panic on overflow

m1sterc001guy (approved):

mxz42:
good point. 
I kept the `usize` and used `checked_add` and an expect with a comment above explaining the situation.

works for all? @dpc @elsirion

elsirion (approved):

PR #1424, opened 2023-01-20
Make sure other peers received all our dkg messages

### LINKED ISSUE #1405: Dkg via UI terminates too fast
Take it with a grain of salt, because maybe I mis-analysed or did something wrong, but ...

I just did a test on localhost trying to make a 2of2 federation and I've got one peer finish and start querying the fullnode, while the other somehow missed the last message and never finished.

Over the network the latency might make it so that this almost never happens, but ideally we would have one more "All other guardians finished" button before we terminate dkg or some kind of best effort ping-pong message to each peer, just to make sure every peer had a chance to receive all messages.

dpc:
Fix #1405

### DIFF
--- a/fedimintd/src/distributedgen.rs
+++ b/fedimintd/src/distributedgen.rs
@@ -2,11 +2,12 @@ use std::collections::BTreeMap;
 use std::fs;
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
+use std::time::Duration;
 
 use anyhow::ensure;
 use fedimint_api::module::DynModuleGen;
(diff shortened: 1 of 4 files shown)


elsirion (approved):

fedimintd/src/distributedgen.rs
```diff
@@ -101,6 +102,12 @@ pub async fn run_dkg(
     )
     .await?;
 
+    // When testing on localhost, it's possible to get
```
dpc, line 105: I'm so sloppy. I wrote all these flushing, to avoid this hack and yet I still pushed it.

PR #1425, opened 2023-01-20
Complain about leftover `dbg!` in `.rs` files

dpc:
Seems like I'm not the only one, so we could say it's endemic now.

### DIFF
--- a/misc/git-hooks/pre-commit
+++ b/misc/git-hooks/pre-commit
@@ -13,6 +13,7 @@ git diff-files --quiet
 is_unclean=$?
 set -e
 
+
 # Stash pending changes and revert them when script ends
 if [ $is_unclean -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


misc/git-hooks/pre-commit
```diff
@@ -22,24 +23,29 @@ fi
 
 git_ls_files="$(git ls-files)"
 
->&2 echo "Checking *.nix files..."
```
dpc, line 25: These were just noise, so removed them, so only errors are displayed.

elsirion (approved):

justinmoon:
To anyone else watching, `dbg!(<variable>)` is really helpful for debugging. You can wrap parameters to function calls with it. Shorter to type than `tracing::info!()` or whatever. I had never used it until doing some pair programming with @dpc yesterday!

PR #1426, opened 2023-01-20
Trim gateway rpc streams

okjodom:
Simplify gateway < > lightning extension rpc spec by defining unary calls for `pay_invoice` and `complete_htlc`.
We don't need streams right now, and should have started off with this simpler rpc spec

### DIFF
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -1,16 +1,14 @@
 use std::{
-    collections::HashMap, net::SocketAddr, path::PathBuf, pin::Pin, str::FromStr, sync::Arc,
-    time::Duration,
+    array::TryFromSliceError, collections::HashMap, net::SocketAddr, path::PathBuf, str::FromStr,
+    sync::Arc, time::Duration,
 };
 
 use anyhow::anyhow;
-use bitcoin::XOnlyPublicKey;
-use bitcoin_hashes::{sha256, Hash};
+use bitcoin_hashes::{hex::ToHex, sha256, Hash};
 use clap::Parser;
 use cln_plugin::{options, Builder, Plugin};
 use cln_rpc::{model, ClnRpc};
 use fedimint_api::{task::TaskGroup, Amount};
-use futures::Stream;
 use ln_gateway::gatewaylnrpc::{
     complete_htlcs_request::{Action, Cancel, Settle},
     gateway_lightning_server::{GatewayLightning, GatewayLightningServer},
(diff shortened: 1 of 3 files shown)


gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -202,63 +202,51 @@ impl GatewayLightning for ClnRpcService {
...
-                .map_err(|e| {
-                    error!("cln pay rpc returned error {:?}", e);
-                    tonic::Status::internal(e.to_string())
-                });
+        } = request.into_inner();
 
-            results.push(outcome);
-        }
+        // Since the cln_rpc client is not `Sync`, we cannot actually lazily pocess the requests into response stream
+        // The current approach will process all requests within the incoming stream, before returning a response stream
```
elsirion, line 214: This documentation seems outdated?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -262,86 +260,72 @@ impl GatewayLightning for ClnRpcService {
...
+                return Err(Status::invalid_argument(e.to_string()));
             }
+        };
+
+        // TODO: Implement non-blocking access to the outcomes map sowe can multithread complete_htlcs
+        if let Some(outcome) = self.interceptor.outcomes.lock().await.remove(&hash) {
+            // Translate action request into a cln rpc response for `htlc_accepted` event
+            let htlca_res = match action {
+                Some(Action::Settle(Settle { preimage })) => {
+                    if let Ok(pk) = XOnlyPublicKey::from_slice(&preimage) {
```
elsirion, line 285: We actually don't care if it's a valid `XOnlyPublicKey` in the extension afaik, that's the problem of the gateway. We could make it a `[u8; 32]` in a follow-up.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -352,11 +336,13 @@ pub enum ClnExtensionError {
 }
 
 /// BOLT 4: https://github.com/lightning/bolts/blob/master/04-onion-routing.md#failure-messages
-/// 2002 error code reports general temporary failure of this processing node.
-fn temp_node_failure() -> serde_json::Value {
+/// 16399 error code reports unknown payment details.
+///
+/// TODO: We should probably use a more specific error code based on htlc processing fail reason
```
elsirion, line 341: Do you want that TODO in the rendered docs?

elsirion (commented):
Thx, I think this makes the code easier to reason about. I just have some minor nits :)

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: `htlc_processing_failure` failure message is not part of  public gateway api. Doesn't need to be in rendered docs

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: This type was originally a `[u8; 32]` Preimage in our definitions. Sending it as Protobuf type over grpc loses the sizing information, but we only need to coerce it into a string to send to cln-pay.

opted to go with `String::from_utf8(preimage)`

elsirion: Why a string though? CLN guarantees this to be 32 hex encoded bytes. My opposition to xonly pub key was that CLN doesn't guarantee its a well-formed pub key and that this is just a fedimint convention and should probably be checked in gatewayd.

elsirion: I mean, technically yes, but only because JSON is shit and doesn't have a native binary data type :laughing: it's a 32bytes hex-encoded array

okjodom: @elsirion, there's no blocking change required here, right? Can we close pr?

elsirion: Just had another look and I think I confused the flow direction of data here actually, so sorry if I caused any confusion.

While I still think not parsing into `XOnlyPublicKey` is semantically more correct, the previous version at least lead to a valid encoding (by accident?) because [`XOnlyPublicKey`'s `Display` impl serializes to hex](https://docs.rs/secp256k1/latest/src/secp256k1/key.rs.html#1127-1129). `String::from_utf8(preimage)` does not achieve this afaik (unless we are transmitting hex encoded data as raw bytes, which would be even more broken).

What we want is taking the `&[u8]` from protobuf and converting it into a hex encoded `[u8; 32]`. So the right move is probably adding an assert that `preimage` is 32bytes long and then using the hex impl form `bitcoin_hashes` (as @mxz42 did all over fedimint recently, e.g. #1472) to generate the string that can be given to CLN.

okjodom: 🔥 . Would have been a bad bug. Good catch on the utf8 stuff

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -260,85 +258,73 @@ impl GatewayLightning for ClnRpcService {
...
+        if let Some(outcome) = self.interceptor.outcomes.lock().await.remove(&hash) {
+            // Translate action request into a cln rpc response for `htlc_accepted` event
+            let htlca_res = match action {
+                Some(Action::Settle(Settle { preimage })) => {
+                    let assert_pk: Result<[u8; 32], TryFromSliceError> =
+                        preimage.as_slice().try_into();
+                    if let Ok(pk) = assert_pk {
+                        serde_json::json!({ "result": "resolve", "payment_key": pk.to_hex() })
+                    } else {
+                        htlc_processing_failure()
```
elsirion, line 287: Could be an `expect`/`panic` since as @dpc noted on a previous PR e.g. sending a too short/long preimage would be a programmer error from the gateway side.

Follow up: at least log with `error` level.

elsirion (approved):

PR #1427, opened 2023-01-20
Delete unnecessary workaround

dpc:
In #1424 I first wrote this workaround, then decided it's meh, then wrote rest of the #1424 with a proper solution, and forgot to delete the workaround.


elsirion (approved):

elsirion:
Lol, I actually wondered about that but was like "better safe than sorry" :laughing:

PR #1429, opened 2023-01-20
Don't hang the UI during dkg

dpc:
Follow up to #1421

I tested the success case and getting back to `/` after restart, but I'm unable to trigger error handling I've added. Seems like nothing from what is being entered to the UI is actually compared, trying to make the settings wrong results in dkg not being able to connect or panics, and never actually a failure, so :shrug: for now.

At least you'll get

image

when dkg is running, which switches automatically to connection string when dkg completes, which is better then previous behavior where call would hang until dkg is complete.

### DIFF
--- a/fedimintd/src/ui.rs
+++ b/fedimintd/src/ui.rs
@@ -52,26 +52,43 @@ async fn home_page(axum::extract::State(_): axum::extract::State<MutableState>) 
 #[derive(Template)]
 #[template(path = "run.html")]
 struct RunTemplate {
-    connection_string: String,
-    has_connection_string: bool,
+    state: RunTemplateState,
+}
+
+enum RunTemplateState {
+    DkgNotStarted,
+    DkgInProgress,
+    DkgDone(String),   // connnection string
+    DkgFailed(String), // error
+    LocalIoError(String),
 }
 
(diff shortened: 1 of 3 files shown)


fedimintd/src/ui.rs
```diff
@@ -52,26 +52,42 @@ async fn home_page(axum::extract::State(_): axum::extract::State<MutableState>)
...
+                            let connect_info = WsFederationConnect::from(&cfg);
+
+                            RunTemplateState::DkgDone(
+                                serde_json::to_string(&connect_info).expect("should deserialize"),
+                            )
+                        }
+                        Err(e) => RunTemplateState::LocalIoError(e.to_string()),
+                    },
+                    Err(e) => RunTemplateState::LocalIoError(e.to_string()),
+                }
```
elsirion, line 85: Could this be an extra fn with early exit to avoid the nesting?

elsirion (approved):
Looks good, just one nit, feel free to address or ignore.

fedimintd/src/ui.rs
dpc: Right, but I'll ignore as I don't have time to retest it right now. Added a TODO comment to refactor.

justinmoon (approved):

PR #1454, opened 2023-01-21
Introducing gateway admin UI (aka, Mintgate)

okjodom:
Gateway admin UI previously under active development in [GETLN/mintgate](https://github.com/GETLN/mintgate) now lives within Fedimint source tree.

This PR parents a "stable" version of the UI under `/gateway/ui` directory. Co-locating the UI and the Gateway server means better guarantees of API compatibility and opens room for wider contribution to the UI experience.

Thanks for initial Mintgate contributors. After this PR lands, we can fully shift future development into this repo.

Next Steps:
- [ ] #2091
- [ ] #2092


okjodom:
Kudos to @IgboPharaoh, for his fantastic past and ongoing contributions to [mintgate](https://github.com/GETLN/mintgate)

🍾 🎊

elsirion:
What is the goal of making it a submodule? I'd be happy to move the repo into the org, but unless we want to package it in one of our binaries having it as a submodule doesn't make much sense imo (for the annoyance that submodules typically cause).

okjodom:
> What is the goal of making it a submodule? I'd be happy to move the repo into the org, but unless we want to package it in one of our binaries having it as a submodule doesn't make much sense imo (for the annoyance that submodules typically cause).

Submodule so it lives super close to the gateway binary source tree. _And we can define build steps that packages a headless vs headed versions of the gateway binary?_

This is tangential to which org hosts the repo. The submodule can point to a repo in fedimint `url = https://github.com/fedimint/mintgate.git`, vs an external repo `url = https://github.com/GETLN/mintgate.git`

Our inspiration and closest approximation is the [web admin set-up in sensei](https://github.com/L2-Technology/sensei/tree/main/web-admin), where they develop the admin ui right within sensei source tree

> annoyance that submodules typically cause

Watching out for pain here. If it's too painful, we could adopt the sensei method. Or, go the other direction and never package UI with the binary.

okjodom:
> EDIT: adding the index over our docs dir is really useful!
🥳 

> EDIT2: maybe having the compatible versions would be nice, but I would not block fedimint of mintgate catching up, so maybe it's more practical for mintgate to state the version of fedimint it works with?

Yes. mintgate might declare versions of compatibility. Or, gatewayd would lock the versions it pulls in. Have a slower upgrade cadence on its mintgate ui. For example, we can define a build process that pulls in latest master from mintgate but only builds using git tags

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -31,9 +32,16 @@ pub async fn run_webserver(
         .route("/connect", post(connect))
         .layer(RequireAuthorizationLayer::bearer(&authkey));
 
+    // TODO: Provide admin UI path from configurations
+    let admin_ui_path = Path::new("../../../admin-ui/build");
+
+    // Serve admin UI
+    let admin_ui = SpaRouter::new("/", admin_ui_path);
```
justinmoon, line 39: this is really neat!

justinmoon:
Recapping [this thread](https://discord.com/channels/990354215060795454/1066444049436905562/1066444054054850580) in discord:

- Submodules suck. Just check the code into `gateway/ui` or folder like this. Will help us rapidly iterate towards stable and usable first release of fedimint. Nix can help us make isolated nix environment so users don't need to touch node.js build if they don't want to. When everything matures, we can revisit moving UIs to separate repos.
- Don't get blocked on the nix setup. OK to expect user to run `npm install` and `npm run build` etc on their own node.js toolchain. If a build folder is found, admin UI is served. If no build folder is found, no UI is served.
- Over time convert the fedimintd admin UI to a single page app using same tooling and conventions as gateway UI. So it's easy for devs to contribute to both.

gateway/ln-gateway/src/rpc/rpc_server.rs
okjodom: Not sure we'll go with this, given the need for bundlng UI files with the binary

m1sterc001guy (approved):
LGTM but want to emphasize that I've never written Typescript before :)

My only concern is additional review burden and code that most maintainers aren't familiar with. Perahaps its possible to add maintainers for just the gateway/ui subdirectory?

okjodom:
> My only concern is additional review burden and code that most maintainers aren't familiar with. Perahaps its possible to add maintainers for just the gateway/ui subdirectory?

Ack. I'm sure we'll build strong front-end capacity on the project, particularly to support Mintgate and the Setup UI. They'll all be built on the same Typescript + React stack

docs/gateway.md
```diff
@@ -88,6 +92,26 @@ As described in [Running Fedimint for dev testing](./dev-running.md#using-the-ga
...
+```
+
+Install the required dependencies and run the Mintgate client
+
+```sh
+yarn install
+yarn start
+```
+
+You should see a running Mintgate instance in your local browser. Probably at `https://localhost:3000`
```
justinmoon, line 111: We should change this port to something in the `81xx` range that we normally use. 3000 likely to cause conflicts.

gateway/ui/src/Admin.tsx
```diff
@@ -0,0 +1,123 @@
...
+			const result = fedListCopy.sort((a, b) =>
+				a.details.date_created < b.details.date_created
+					? 1
+					: a.details.date_created > b.details.date_created
+						? -1
+						: 0
+			);
+
+			return setFedlist(result);
+		}
```
justinmoon, line 77: Are we using tabs here? Might render on GitHub better if we're using spaces?

justinmoon (approved):
Nice!

docs/gateway.md
```diff
@@ -72,9 +72,13 @@ Options:
   -V, --version                    Print version information
 ```
 
-### mintgate
+### Mintgate
```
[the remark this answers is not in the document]
okjodom: Ack. The pet name is nice, but not great for wider audience

PR #1466, opened 2023-01-21
Change `btc_client` to `bitcoin-cli` for consistency

mxz42:
I'm aware that this alias "shadows" the real `bitcoin-cli`. But I'd argue it's fine to do that in the tmuxinator? 
alias points to: 
```
bitcoin-cli -regtest -rpcuser=bitcoin -rpcpassword=bitcoin
```

If you think that's a problem, then I'd go for the name `btc-cli`. That gets rid of the annyoing to type underscore (:smile:)  and doesn't shadow the real `bitcoin-cli`

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -5,7 +5,7 @@ pre_window:
   - source .tmpenv
   - alias ln1="\$FM_LN1"
   - alias ln2="\$FM_LN2"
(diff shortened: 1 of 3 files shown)


elsirion (approved):
Looks good, did you manually verify it still works?

justinmoon:
Love it!

mxz42:
yes, run tmuxinator and a couple commands to verify @elsirion

PR #1471, opened 2023-01-21
Allow federation to send additional config values to clients

### LINKED ISSUE #1432: Allow additional fields to be transmitted to clients in config

elsirion:
Fixes #1432

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -12,7 +12,9 @@ use clap::{Parser, Subcommand};
 use fedimint_core::api::{
     FederationApiExt, GlobalFederationApi, IFederationApi, WsClientConnectInfo, WsFederationApi,
 };
-use fedimint_core::config::{load_from_file, ClientConfig, ModuleGenRegistry};
+use fedimint_core::config::{
+    load_from_file, ClientConfig, ModuleGenRegistry, META_FEDERATION_NAME_KEY,
+};
 use fedimint_core::core::{
     LEGACY_HARDCODED_INSTANCE_ID_LN, LEGACY_HARDCODED_INSTANCE_ID_MINT,
     LEGACY_HARDCODED_INSTANCE_ID_WALLET,
@@ -533,7 +535,7 @@ async fn handle_command(
                 .config()
                 .as_ref()
                 .meta
-                .get("federation_name")
+                .get(META_FEDERATION_NAME_KEY)
                 .ok_or_else(|| {
                     CliError::from(
                         CliErrorKind::GeneralFederationError,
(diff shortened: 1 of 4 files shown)


client/cli/src/main.rs
```diff
@@ -542,9 +542,24 @@ async fn handle_command(
                 .iter()
                 .map(|(amount, coins)| (amount.to_owned(), coins.len()))
                 .collect();
+
+            let federation_name = client
+                .config()
+                .as_ref()
+                .meta
+                .get("federation_name")
```
jkitman, line 550: Can we make this a `const`?

elsirion: Done.

client/cli/src/main.rs
```diff
@@ -542,9 +542,24 @@ async fn handle_command(
                 .iter()
                 .map(|(amount, notes)| (amount.to_owned(), notes.len()))
                 .collect();
+
+            let federation_name = client
+                .config()
+                .as_ref()
+                .meta
+                .get(META_FEDERATION_NAME_KEY)
+                .ok_or_else(|| {
```
dpc, line 551: If it's required... why make it an "additional config value"?

elsirion: I think it was a suggestion by @justinmoon to make it not required anymore. There could be a BIP-like process to define meta-data extensions. And even though most federations will have a name, it's not *really* necessary.

dpc: Makes sense, but then we should remove this error here? Or default to "unnamed" for now?

elsirion: Right, the client needs to be able to just display all the meta fields.

elsirion:
One limitation is that it only allows `String -> String` maps, so more complex data structures would need some additional encoding. This is due to the need to hash the config for signing, so if we wanted more general data types we'd need to create a `DynMetaEntry` struct with the `trait IMetaEntry: Serialize + Deserialize + Encodable` so that it's both compatible with `serde` and our own encoding API.

client/cli/src/main.rs
elsirion: Done.

fedimint-api/src/config.rs
```diff
@@ -0,0 +1 @@
+
```
justinmoon, line 1: What it the purpose of this empty file?

justinmoon (approved):

fedimint-api/src/config.rs
elsirion: Good catch, rebasing artifact.

justinmoon (approved):

dpc (approved):

PR #1472, opened 2023-01-21
Remove the `hex` crate

mxz42:
This is the final PR to remove the `hex` crate. 
- all uses are replaced with the traits `ToHex` and `FromHex` from `bitcoin_hashes::hex`

Unfortunately, rust-bitcoin [removed the `ToHex` trait recently](https://github.com/rust-bitcoin/rust-bitcoin/pull/1531/files).  The functionality was moved to the `bitcoin-internals` crates which is supposed to be used only internally by them. However, there is a [discussion about exposing this publicly](https://github.com/rust-bitcoin/rust-bitcoin/discussions/1568#discussioncomment-4743062). Time will tell.

if you'd like to use the `hex::serde` functionality: you can replace this currently with `fedimint_api::hex::serde` ([code](https://github.com/fedimint/fedimint/blob/6cf12b0a53386908f64c043a46d79cfcbdf23d00/fedimint-api/src/hex/serde.rs)). 
[example](https://github.com/fedimint/fedimint/blob/6cf12b0a53386908f64c043a46d79cfcbdf23d00/modules/fedimint-mint/src/common.rs#L17)

---

WIP: currently looking at the error handling of some of the deserializers. specifically at code like: 

```rust
        let bytes = hex::decode(hex_str.as_ref()).map_err(|_e| D::Error::custom("Invalid hex"))?;
// or 
        bincode::deserialize(&bytes).map_err(|e| serde::de::Error::custom(format!("{:?}", e)))
```

I think (know) that all these errors implement the `Display` trait, so the following is shorter and works, too: 

```rust
        let bytes = Vec::from_hex(&hex_str).map_err(serde::de::Error::custom)?;
```

and while our error `invalid hex` is correct, the errors from [`bitcoin_hashes::hex`](https://docs.rs/bitcoin_hashes/latest/bitcoin_hashes/hex/enum.Error.html#impl-Error)  are rather nice as they also contain more information: 
```rust
impl fmt::Display for Error {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match *self {
            Error::InvalidChar(ch) => write!(f, "invalid hex character {}", ch),
            Error::OddLengthString(ell) => write!(f, "odd hex string length {}", ell),
            Error::InvalidLength(ell, ell2) => write!(f, "bad hex string length {} (expected {})", ell2, ell),
        }
    }
}
```

is there a reason why we sometimes use `format!` with the `:?` debug print? 

What am I missing here?


elsirion (approved):
Sweet! After this PR only `fedimint-sqlite` (through `sqlx`) and the gateway (throught `cln-rpc`) depend on `hex`. This means if we can upstream a similar fix to `cln-rpc` (they are already part of the wider rust-bitcoin ecosystem) only the client will ever depend on `hex` (as the main consumer of `fedimit-sqlite`).

PR #1476, opened 2023-01-22
Add backup&restore to ln-gateway (and gateway-cli)

dpc:
It didn't take me long before I corrupted first mainnet ln-gateway I've set up by doing a `gateway-cli deposit` one block too early. Since our client code does db modifications first, and asks federation to make a tx second, the gateway is now stuck looping and waiting for an outcome of a transaction that didn't ever happen. :S

After I realized what happened I first made a loudest facepalm in the world. Then I started thinking if I could easily recover from this corruption, and it occurred to me that a e-cash restore that is already implemented in the client would probably fix it. So here it is - just enough code to be able to make gateway do `restore` (and `backup`, as I'm already doing it).

Considering the state of the `client` consensus state handling, we probably really, really need this to work.

### DIFF
--- a/gateway/ln-gateway/src/actor.rs
+++ b/gateway/ln-gateway/src/actor.rs
@@ -2,7 +2,7 @@ use std::{sync::Arc, time::Duration};
 
 use bitcoin::{Address, Transaction};
 use bitcoin_hashes::sha256;
-use fedimint_api::{Amount, OutPoint, TransactionId};
+use fedimint_api::{task::TaskGroup, Amount, OutPoint, TransactionId};
 use fedimint_server::modules::{
     ln::contracts::{ContractId, Preimage},
     wallet::txoproof::TxOutProof,
@@ -227,6 +227,35 @@ impl GatewayActor {
             .await
             .map_err(LnGatewayError::ClientError)
             .map(|out_point| out_point.txid)
+    }
+
+    pub async fn backup(&self) -> Result<()> {
+        self.client
+            .mint_client()
+            .back_up_ecash_to_federation()
+            .await
+            .map_err(LnGatewayError::Other)?;
+
+        Ok(())
+    }
+
+    pub async fn restore(&self) -> Result<()> {
+        // TODO: get the task group from `self`
+        let mut task_group = TaskGroup::new();
+
+        self.client
+            .mint_client()
+            .restore_ecash_from_federation(10, &mut task_group)
+            .await
+            .map_err(LnGatewayError::Other)?
+            .map_err(|e| LnGatewayError::Other(e.into()))?;
+
+        task_group
+            .join_all(None)
+            .await
+            .map_err(LnGatewayError::Other)?;
+
+        Ok(())
     }
 
     pub async fn get_balance(&self) -> Result<Amount> {
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -7,8 +7,8 @@ use fedimint_server::modules::wallet::txoproof::TxOutProof;
 use ln_gateway::{
     config::GatewayConfig,
     rpc::{
-        rpc_client::RpcClient, BalancePayload, ConnectFedPayload, DepositAddressPayload,
-        DepositPayload, WithdrawPayload,
+        rpc_client::RpcClient, BackupPayload, BalancePayload, ConnectFedPayload,
+        DepositAddressPayload, DepositPayload, RestorePayload, WithdrawPayload,
     },
 };
 use mint_client::utils::from_hex;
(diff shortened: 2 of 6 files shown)


dpc:
@elsirion @okjodom @justinmoon Please sanity-check it. I'm going to sleep now, but tomorrow once I find some time I'm going to make copy of rocksdb state and attempt recovery using this code.

okjodom:
@dpc, I found some missing connectors and boilerplate: See commit [b0bb0944](https://github.com/dpc/fedimint/pull/51/commits/b0bb0944e686e460d1664af811a06c6782eb9235)

gateway/ln-gateway/src/actor.rs
```diff
@@ -229,6 +229,33 @@ impl GatewayActor {
             .map(|out_point| out_point.txid)
     }
 
+    pub async fn backup(&self) -> Result<()> {
```
okjodom, line 232: should we be looking to call backup periodically, or even best-effort as the gateway is looking to shut down?
why does the client need to initiate backup, and not have that happen automatically?

gateway/ln-gateway/src/actor.rs
```diff
@@ -229,6 +229,33 @@ impl GatewayActor {
...
+        self.client
+            .mint_client()
+            .back_up_ecash_to_federation()
+            .await
+            .map_err(LnGatewayError::Other)?;
+
+        Ok(())
+    }
+
+    pub async fn restore(&self) -> Result<()> {
```
okjodom, line 242: should be calling restore on actor start. Just in case there was previous backed up state?

gateway/ln-gateway/src/actor.rs
dpc: We could. We never wired it anywhere.

dpc: Now I'm a bit worried about making backup of corrupted state. :D

gateway/ln-gateway/src/actor.rs
dpc: No. That's not the intended use and it might be way too slow. In normal circumstances the client should track their state on their own.

dpc:
> @dpc, I found some missing connectors and boilerplate: See commit [b0bb0944](https://github.com/dpc/fedimint/pull/51/commits/b0bb0944e686e460d1664af811a06c6782eb9235)

Thanks!

I've cherry-picked, squashed and pushed again.

gateway/ln-gateway/src/actor.rs
justinmoon: Yea. this is a nice change, but I don't think it will fix your problem. Your backup will still contain invalid `NotIssuanceRequest`s which will prevent your valid `NotIssuanceRequest`s from completing.

I think fixing `https://github.com/fedimint/fedimint/issues/1479` would solve the problem. You'd still have invalid `NotIssuanceRequest`s in your DB, but they would just cause a little delay when fetching balance.

gateway/ln-gateway/src/actor.rs
justinmoon: Yea. Restore is when you lost your gateway client db and want to restore it from seed phrase. Should be done manually by the CLI.

gateway/ln-gateway/src/actor.rs
justinmoon: Actually I think this change could work on its own, but only if you never made a backup to the federation and restored by downloading entire epoch history. Then it wouldn't accidentally pick up the invalid `NotIssuanceRequest`s.

dpc: https://github.com/fedimint/fedimint/blob/73b79347e370594794a0500cbb39dac9468c5756/client/client-lib/src/mint/backup.rs#L114

We do back-up, track, wipe old & restore from backup https://github.com/fedimint/fedimint/blob/73b79347e370594794a0500cbb39dac9468c5756/client/client-lib/src/mint/backup.rs#L114 `NoteIssuanceRequest`s

justinmoon: Nevermind, I think this will work!

dpc:
https://i.imgur.com/bz07n.gif

justinmoon (approved):
LGTM. I'll defer the merge to @elsirion because I'm unfamiliar with backup code.

okjodom:
We can batch these backup / restore operations on a set of federations, but that is a UI overlay over this feature.
Looking good!

okjodom (approved):

dpc:
@okjodom The way things are currently, it's a bit risky to run `recover` on a federation that is currently running. We have no global locking, etc. As things are right now, please consider these as mostly 'for emergency' purposes. :D

Issue #1245, opened 2023-01-08
Upstream HtlcAccepted, Onion, and Htlc structs to cln_plugin


justinmoon:
https://github.com/fedimint/fedimint/blob/b85aa04d76f5bbbe664453d9918fac06588dfc3a/gateway/ln-gateway/src/cln.rs#L33-L60

mubarak23:
i want to believe this has been done @justinmoon

elsirion:
From all I can find in https://github.com/ElementsProject/lightning/tree/master/cln-rpc this is still an open issue. 

Currently, `cln_rpc` uses functions that [consume and output JSON for hooks](https://github.com/ElementsProject/lightning/blob/21fcd4594b9db99017e6c6f4cca750566d1b2969/plugins/src/lib.rs#L198-L211). Ideally there'd be a way to register typed hooks where decoding/encoding is handled by the library and the library also supplies the structs representing specific hook callbacks.

```rust
    pub fn hook<C, F>(mut self, hookname: &str, callback: C) -> Self
    where
        C: Send + Sync + 'static,
        C: Fn(Plugin<S>, Request) -> F + 'static,
        F: Future<Output = Response> + Send + 'static,
    {
        self.hooks.insert(
            hookname.to_string(),
            Hook {
                callback: Box::new(move |p, r| Box::pin(callback(p, r))),
            },
        );
        self
    }
```

I don't know what a good API would look like, maybe a `Hook` trait that is implemented on the input type and has an associated type specifying the output type. I also don't know how all that fits into CLN's code generation philosophy, but worth a try imo if you have some experience designing rust APIs.

For reference: we use the `hook` feature for HTLC interception:

https://github.com/fedimint/fedimint/blob/b85aa04d76f5bbbe664453d9918fac06588dfc3a/gateway/ln-gateway/src/cln.rs#L138-L152

m1sterc001guy:
We are planning on removing Core Lightning support post v0.5 https://github.com/fedimint/fedimint/issues/6218. 

Closing

PR #1483, opened 2023-01-22
Unify CLN plugin types between old and new plugin

elsirion:
6723518fa2b018686ddb5570f9e97fd23fedb38a changed the type definition of `Htlc` and `Onion` to be compliant with the current spec, but only for the new CLN plugin. The old one remained out of spec, which this commit fixes.

As a second step we should either upstream these types (together with tests, so that the CLN team notices when they break their API) or put them in a common crate so we don't have to maintain this redundancy.

### DIFF
--- a/gateway/ln-gateway/src/cln.rs
+++ b/gateway/ln-gateway/src/cln.rs
@@ -30,10 +30,12 @@ where
 }
 
 // TODO: upstream these structs to cln-plugin
+// See: https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md#htlc_accepted
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Htlc {
+    pub short_channel_id: String,
     #[serde(deserialize_with = "as_fedimint_amount")]
-    pub amount: Amount,
+    pub amount_msat: Amount,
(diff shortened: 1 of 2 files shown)


justinmoon (dismissed):

justinmoon:
Can we remove these structs, too? https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/bin/cln_extension.rs#L70-L91

Got an issue for upstreaming these https://github.com/fedimint/fedimint/issues/1245

okjodom:
> As a second step we should either upstream these types (together with tests, so that the CLN team notices when they break their API) or put them in a common crate so we don't have to maintain this redundancy.

cool.  I can now prioritize #1245

okjodom (dismissed):

justinmoon (approved):

elsirion:
> Can we remove these structs, too? https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/bin/cln_extension.rs#L70-L91. They're both in the same crate so I don't see why not.

I removed the commit doing so again since I had to change back the short channel id to a string. Once we impl a parsing fn we can change it back, but I merely want to fix a bug here.

okjodom (approved):

Issue #1322, opened 2023-01-13
`raw_find_by_prefix` should return a stream instead of an iterator


elsirion:
If we want to integrate async DBs (why all the DB functions are async now) streaming data from the DB will probably also be async. Thus a stream fits better than an iterator.

https://github.com/fedimint/fedimint/blob/872b7ead0c32326c2bc56fad119aea2c09a88382/fedimint-api/src/db/mod.rs#L128

https://github.com/fedimint/fedimint/blob/872b7ead0c32326c2bc56fad119aea2c09a88382/fedimint-api/src/db/mod.rs#L60

dpc:
Theoretically yes but streams are such PITA to work with (as async is half-baked), and it seems even rocksdb doesn't implement it as a stream https://docs.rs/rocksdb/latest/rocksdb/struct.DBIteratorWithThreadMode.html#impl-Iterator-for-DBIteratorWithThreadMode%3C%27a%2C%20D%3E so :shrug: for now? :D

dpc:
Oh, wait ... https://github.com/fedimint/fedimint/blob/master/fedimint-rocksdb/src/lib.rs#L185 rocksdb doesn't do anything async? Which means we are doing blocking IO in our async code all the time anyway?

m1sterc001guy:
It looks like Rocksdb does do some asynchronous I/O, but they only expose a synchronous API to the application

From docs:

> One way to mitigate the impact of storage latency is to read asynchronously and in parallel as much as possible, in order to hide IO latency. We have implemented this in RocksDB in Iterators and MultiGet. In Iterators, we prefetch data asynchronously in the background for each file being iterated on, unlike the current implementation that does prefetching synchronously, thus blocking the iterator thread. In MultiGet, we determine the set of files that a given batch of keys overlaps, and read the necessary data blocks from those files in parallel using an asynchronous file system API. These optimizations have significantly decreased the overall latency of the RocksDB MultiGet and iteration APIs on slower storage compared to local flash.

> The optimizations described here are in the internal implementation of Iterator and MultiGet in RocksDB. The user API is still synchronous, so existing code can easily benefit from it. We might consider async user APIs in the future.

dpc:
@m1sterc001guy Oh dear. We should wrap everything in `spawn_blocking`, but I be this is going to conflict with lifetimes and such.


[...]

PR #1486, opened 2023-01-23
Isolate each Database when initializing the module

m1sterc001guy:
As a follow up to https://github.com/fedimint/fedimint/pull/1331, this PR allows for entire Databases to be isolated instead of just individual DatabaseTransactions. We want the ability to isolate Databases and DatabaseTransactions because there are some db transactions that need to atomically span multiple modules, but we also sometimes give each module a handle to the database, such as during module_initialization (this db handle is only used in the wallet module currently). 

Isolated databases have the same API as normal databases accept that begin_transaction now returns a DatabaseTransaction that is isolated, so all operations will have a prefix prepended to each key to keep the data separate.

### DIFF
--- a/fedimint-api/src/db/mem_impl.rs
+++ b/fedimint-api/src/db/mem_impl.rs
@@ -170,10 +170,15 @@ impl Iterator for MemDbIter {
 #[cfg(test)]
 mod tests {
     use super::MemDatabase;
-    use crate::{db::Database, module::registry::ModuleDecoderRegistry};
+    use crate::{core::ModuleInstanceId, db::Database, module::registry::ModuleDecoderRegistry};
 
     fn database() -> Database {
         Database::new(MemDatabase::new(), ModuleDecoderRegistry::default())
+    }
+
+    fn module_database(module_instance_id: ModuleInstanceId) -> Database {
+        let db = Database::new(MemDatabase::new(), ModuleDecoderRegistry::default());
+        db.new_isolated(module_instance_id)
     }
 
     #[test_log::test(tokio::test)]
@@ -234,5 +239,10 @@ mod tests {
     #[test_log::test(tokio::test)]
     async fn test_module_dbtx() {
         fedimint_api::db::verify_module_prefix(database()).await;
+    }
+
+    #[test_log::test(tokio::test)]
+    async fn test_module_db() {
+        fedimint_api::db::verify_module_db(database(), module_database(1)).await;
     }
 }
--- a/fedimint-api/src/db/mod.rs
+++ b/fedimint-api/src/db/mod.rs
@@ -53,7 +53,10 @@ pub trait IDatabase: Debug + Send + Sync {
 }
 
 #[derive(Clone, Debug)]
-pub struct Database(Arc<DatabaseInner<dyn IDatabase>>);
+pub struct Database {
+    inner_db: Arc<DatabaseInner<dyn IDatabase>>,
+    module_instance_id: Option<ModuleInstanceId>,
+}
 
 // NOTE: `Db` is used instead of just `dyn IDatabase`
 // because it will impossible to construct otherwise
@@ -93,14 +96,30 @@ impl Database {
             module_decoders,
         };
 
-        Self(Arc::new(inner))
+        Self {
+            inner_db: Arc::new(inner),
+            module_instance_id: None,
+        }
+    }
+
+    pub fn new_isolated(&self, module_instance_id: ModuleInstanceId) -> Self {
+        let db = self.inner_db.clone();
+        Self {
+            inner_db: db,
+            module_instance_id: Some(module_instance_id),
+        }
     }
 
     pub async fn begin_transaction(&self) -> DatabaseTransaction {
-        DatabaseTransaction::new(
-            self.0.db.begin_transaction().await,
-            self.0.module_decoders.clone(),
-        )
+        let dbtx = DatabaseTransaction::new(
+            self.inner_db.db.begin_transaction().await,
+            self.inner_db.module_decoders.clone(),
+        );
+
+        match self.module_instance_id {
+            Some(module_instance_id) => dbtx.new_module_tx(module_instance_id),
+            None => dbtx,
+        }
     }
 
     /// Runs a closure with a reference to a database transaction and tries to commit the
@@ -234,6 +253,7 @@ pub trait IDatabaseTransaction<'a>: 'a + Send {
 
 // TODO: use macro again
 #[doc = " A handle to a type-erased database implementation"]
+#[derive(Clone)]
 pub struct CommitTracker {
     is_committed: bool,
     has_writes: bool,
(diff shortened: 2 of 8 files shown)


fedimint-api/src/db/mod.rs
```diff
@@ -247,6 +261,77 @@ impl Drop for CommitTracker {
...
+            .await
+    }
+
+    async fn raw_find_by_prefix(&mut self, key_prefix: &[u8]) -> PrefixIter<'_> {
+        Box::new(
+            self.dbtx
+                .with_module_prefix(self.prefix)
+                .raw_find_by_prefix(key_prefix)
+                .await
+                .collect::<Vec<_>>()
```
m1sterc001guy, line 311: I don't think this is the greatest, but we'll probably need to wait to fix this until https://github.com/fedimint/fedimint/issues/1322 is fixed.

fedimint-api/src/db/mod.rs
```diff
@@ -53,7 +53,10 @@ pub trait IDatabase: Debug + Send + Sync {
 }
 
 #[derive(Clone, Debug)]
-pub struct Database(Arc<DatabaseInner<dyn IDatabase>>);
+pub struct Database(
+    Arc<DatabaseInner<dyn IDatabase>>,
+    pub Option<ModuleInstanceId>,
```
Maan2003, line 58: should use a named struct now.

fedimint-api/src/db/mem_impl.rs
```diff
@@ -170,12 +170,17 @@ impl Iterator for MemDbIter {
...
     use super::MemDatabase;
-    use crate::{db::Database, module::registry::ModuleDecoderRegistry};
+    use crate::{core::ModuleInstanceId, db::Database, module::registry::ModuleDecoderRegistry};
 
     fn database() -> Database {
         Database::new(MemDatabase::new(), ModuleDecoderRegistry::default())
     }
 
+    fn module_database(module_instance_id: ModuleInstanceId) -> Database {
+        let db = Database::new(MemDatabase::new(), ModuleDecoderRegistry::default());
```
dpc, line 180: Why default? Empty decoder registries make dpc sad. 

Please no more empty decoder registries. Take as an argument.

fedimint-api/src/db/mod.rs
```diff
@@ -247,6 +261,77 @@ impl Drop for CommitTracker {
     }
 }
 
+/// ModuleDatabaseTransaction is a wrapper around IsolatedDatabaseTransaction that
```
dpc, line 264: Why do we need ... a wrapper around a wrapper? Could this be just 

```
struct ModuleDatabaseTransaction<'a> {
    dbtx: DatabaseTransaction<'a>,
}
```

where `dbtx` is already wrapped?

fedimint-api/src/db/mod.rs
```diff
@@ -247,6 +261,77 @@ impl Drop for CommitTracker {
...
+            self.dbtx
+                .with_module_prefix(self.prefix)
+                .raw_find_by_prefix(key_prefix)
+                .await
+                .collect::<Vec<_>>()
+                .into_iter(),
+        )
+    }
+
+    async fn commit_tx(self: Box<Self>) -> Result<()> {
```
dpc, line 316: If we have `ModuleDatabaseTransaction` we might want to not have `commit_tx` on them at all.

fedimint-api/src/db/mod.rs
dpc: We could already implement it (probably iface + dyn-newtype, again), and eventually we  will just turn it from `impl Iterator ...` to `impl Stream ...`.

fedimint-api/src/db/mod.rs
m1sterc001guy: I'm actually running into problems when I do that, because the `with_module_prefix` function takes a mutable reference to self, so it is still borrowed when trying to return the new ModuleDatabaseTransaction (that has a wrapped dbtx).

cannot return value referencing function parameter `dbtx`
returns a value referencing data owned by the current function

Let me know if I'm missing something but the current way seems the easiest to me?

fedimint-api/src/db/mod.rs
m1sterc001guy: I changed the implementation to panic, similar to IsolatedDatabaseTransaction

fedimint-api/src/db/mod.rs
m1sterc001guy: I'll leave for a follow up to keep this PR focused

fedimint-api/src/db/mod.rs
dpc: It should work - in similar way that it works for `IsolatedDatabaseTransaction` to wrap the inner transaction.

Possibly here you need `dbtx: &mut dyn IDatabaseTransaction<'a>` as well - same way?

m1sterc001guy: I don't think that works either, otherwise I could just reuse IsolatedDatabaseTransaction and I wouldn't need another wrapper. I believe the issue is that I need to create a DatabaseTransaction in begin_transaction that is wrapped, from nothing. I can't borrow a DatabaseTransaction that was created in begin_transaction and then return the new wrapped one, because the new one still has a reference to the old one that has now dropped out of scope.

elsirion:
I feel like eventually we'll have to think about these new requirements holistically and come up with a better overall architecture. But for now this seems ready for mergig except for a few minor nits. Thoughts @dpc?

dpc (dismissed):
Good enough. These owned vs borrowed seems weirdly inconsistent, but if it works, then we're good for now.

fedimint-api/src/db/mod.rs
```diff
@@ -247,6 +261,77 @@ impl Drop for CommitTracker {
     }
 }
 
+/// ModuleDatabaseTransaction is a wrapper around IsolatedDatabaseTransaction that
+/// consumes an existing DatabaseTransaction. This allows entire Databases to be
+/// isolated and calling begin_transaction will always produce a ModuleDatabaseTransaction,
+/// which is isolated from other modules by prepending a prefix to each key.
+struct ModuleDatabaseTransaction<'a> {
+    dbtx: DatabaseTransaction<'a>,
```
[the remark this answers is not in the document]
m1sterc001guy: Yeah I initially tried this, but ran into issue with the borrow checker when returning the new DatabaseTransaction (since IsolatedDatabaseTransaction borrows an exisitng DatabaseTransaction).

This is the only workable solution I've been able to come up with but if anyone comes up with a different solution we can certainly refactor.

fedimint-api/src/db/mod.rs
```diff
@@ -93,14 +96,30 @@ impl Database {
...
+            module_instance_id: None,
+        }
+    }
+
+    pub fn new_isolated(&self, module_instance_id: ModuleInstanceId) -> Self {
+        let db = self.inner_db.clone();
+        Self {
+            inner_db: db,
+            module_instance_id: Some(module_instance_id),
+        }
```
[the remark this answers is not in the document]
m1sterc001guy: Yeah. I'll make the change.

fedimint-api/src/db/mod.rs
```diff
@@ -392,6 +483,20 @@ impl<'parent> DatabaseTransaction<'parent> {
...
+        let decoders = self.decoders.clone();
+        let commit_tracker = self.commit_tracker.clone();
+        let wrapped = ModuleDatabaseTransaction::new(self, module_instance_id);
+        DatabaseTransaction {
+            tx: Box::new(wrapped),
+            decoders,
+            commit_tracker,
+        }
+    }
+
```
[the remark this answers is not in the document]
m1sterc001guy: Hmm that's how its coded right now, but I'm up for discussing whether or not they should.

IsolatedDatabaseTransactions can never be committed because they do not own the transaction that can be committed. But this API allows a module (assuming it has access to a Database handle, which it does during ModuleGen), to own a transaction. Right now this is only used in the wallet module and is read-only, so its not a problem, but perhaps other modules will want to write into the database during ModuleGen. What do you guys thinks?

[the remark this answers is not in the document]
m1sterc001guy: Thinking about this a bit more, I think we should let them commit. That's what makes 'ModuleDatabaseTransaction' and 'IsolatedDatabaseTransaction' different. 'IsolatedDatabaseTransaction' needs to be created from outside the module and borrowed when passed into the module. `ModuleDatabaseTransaction` allows a module to own the transaction so it should be able to commit it if it needs to.

I'll add this to the documentation.

fedimint-api/src/db/mod.rs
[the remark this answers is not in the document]
m1sterc001guy: Added a test as well.

elsirion (approved):

PR #1489, opened 2023-01-23
Makes fetch notes parallel and eliminates unnecessary retry

### LINKED ISSUE #1479: `MintClient::fetch_all_notes` hangs forever if DB contains invalid `NoteIssuanceRequest`
If DB contains invalid `NotIssuanceRequest` entries, [MintClient::fetch_all_notes](https://github.com/fedimint/fedimint/blob/5aefd0b57b6e38a109d8a6428ee44524a73a9cd2/client/client-lib/src/mint/mod.rs#L480-L511) will hang forever.

Instead, we should probably enforce a "max retries" to calls to `fetch_coins`.

Also, it would be nice if these calls could be more concurrent. Imagine if the database somehow accidentally has 10 invalid `NotIssuanceRequests`. We fetch each one with a timeout of 1 second, and max retries of 3. That would take 30 seconds currently, but only 3 seconds if calls could be done concurrently.

jkitman:
Since we should retry 404s automatically based on our strategy, we can eliminate the infinite loop here.

In the future once polling is replaced by notifications this becomes further irrelevant.

Fixes #1479

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -517,20 +517,20 @@ async fn handle_command(
         ),
         Command::Fetch => {
             let mut result = Vec::<OutPoint>::new();
-            let mut has_error = false;
+            let mut last_error = None;
             for fetch_result in client.fetch_all_notes().await {
                 match fetch_result {
                     Ok(v) => result.push(v),
-                    Err(_) => {
-                        has_error = true;
+                    Err(e) => {
+                        last_error = Some(e);
                     }
                 }
             }
-            if has_error {
+            if let Some(error) = last_error {
                 Err(CliError::from(
                     CliErrorKind::GeneralFederationError,
                     "failed to fetch notes",
-                    None,
+                    Some(Box::new(error)),
                 ))
             } else {
                 Ok(CliOutput::Fetch { issuance: (result) })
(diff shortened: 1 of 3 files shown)


justinmoon:
Looks good. I think we should also apply a timeout to the `fetch_notes` call. I've observed this hanging forever.

jkitman:
> Looks good. I think we should also apply a timeout to the `fetch_notes` call. I've observed this hanging forever.

This should fix that because it removes the 404 retry from calls to `fetch_output_outcome` which is where it's retrying forever.

dpc (approved):

dpc:
> This should fix that because it removes the 404 retry from calls to `fetch_output_outcome` which is where it's retrying forever.

We generally should wrap every networking operation in some form of a timeout at some point. Any network connection can hang indefinitely in certain (uncommon, but still) circumstances.

But that's more of a technicality that should not block this improvement.

jkitman:
> We generally should wrap every networking operation in some form of a timeout at some point. Any network connection can hang indefinitely in certain (uncommon, but still) circumstances.

Hopefully moving from polling to event subscriptions will avoid a lot of the issues we get related to retries.

Issue #1493, opened 2023-01-23
All binaries should register `tracing-subscriber` with `env` feature and default logging level (`info`?)


dpc:
I've just noticed that not all binaries are registering logging. Having logging is very useful once things don't go smoothly and investigation needed. We should go through every binary and make sure it has logging set properly.

Maan2003:
Maybe a helper in fedimint-api?

So logging is consistent everywhere.

dpc:
> Maybe a helper in fedimint-api?

We don't want to pull in extra deps into `fedimint-api`. That crate is on a diet.

It's a responsibility of each binary to set up its logging. It just so happen that in our project we will pretty much always use `tracing-subsriber` with `RUST_LOG` controlled log levels, and something like `info` as default level, as a convention.

Maan2003:
> We don't want to pull in extra deps into fedimint-api. That crate is on a diet.

cargo features will work fine.

> It's a responsibility of each binary to set up its logging. It just so happen that in our project we will pretty much always use tracing-subsriber with RUST_LOG controlled log levels, and something like info as default level, as a convention.

Say, I want to have conditional tracing-chrome or opentelemetry. if it is in fedimint-api, we just have it as cargo feature of fedimint-api, instead of having it in all the binaries.

dpc:
> cargo features will work fine.

Agreed. I think that would work.

I just don't think architecturally it belongs in `fedimint-api` which is to be public-facing core interoperability crate. The smaller we can make it, the better. I'd like to throw out anything that is not absolutely necessary.

It will be consumed about anything fedimint related including 3rd party modules. Optional dependencies and features are a complexity, all just to save couple of lines in a handful of binaries that barely ever change. It might also create downstream problems with feature unification, or users wanting to use different (patched) version of `tracing-subsriber`. 

I think it all falls into "duplication is cheaper than the wrong abstraction" umbrella.

If at any point we feel like we really would like a common function, there's probably going to be a room for `fedimint-utils` (placeholder name) that will be intended for exactly this sort of things: common code, utilities, etc. that we use across the project, that we do not want to promote to be publicly visible. I would have no qualms with such a code living there.

If you want to have it already - please do, I'm sure we'll find bunch of other code that we could move there. Maybe we should even more `crypto/aead` to be a module there behind a cargo feature etc.

PR #1494, opened 2023-01-23
Enable debug tracing on gateway binaries

okjodom:
add tracing subscriber on `gateway-cl-extension` and `ln_gateway` binaries.

resolution for #1493 for the gateway

### DIFF
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -27,6 +27,7 @@ use tokio::{
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::{transport::Server, Status};
 use tracing::{debug, error};
+use tracing_subscriber::EnvFilter;
 
 #[derive(Parser)]
 pub struct ClnExtensionOpts {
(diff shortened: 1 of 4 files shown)


dpc (approved):

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -38,6 +39,12 @@ pub struct ClnExtensionOpts {
 // Note: Once this binary is stable, we should be able to remove current 'ln_gateway'
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
+        )
+        .init();
+
```
elsirion, line 47: You can't use the default subscriber that prints to stdout in CLN plugins since they use stdin/out for communication with CLN. [`cln-pugin` brings its own subscriber which formats log events as valid CLN JSON messages. We should try to understand it better.](https://github.com/ElementsProject/lightning/blob/master/plugins/src/logging.rs)

elsirion (changes requested):

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: Just realized this when debugging changes in this PR!
From what @dpc has pointed out, the log level flag is controllable with `CLN_PLUGIN_LOG`

Validating. And closing this PR as no-fix

dpc: `tracing_subscriber` defaults to stdout? Bad default.

PR #1495, opened 2023-01-23
finalize_change shouldn't commit to the database

### LINKED ISSUE #1478: MintClient::finalize_change shouldn't commit to the database
Having `MintClient::finalize_change` commit to the database assumes that the transaction being built will be confirmed by the federation. If it isn't, then the database will now contain an `NotIssuanceRequest` where not corresponding transaction exists in the federation. Calls to `fetch_all_notes()` will currently hang until the transaction is confirmed.

Instead, commit **after** `submit_transaction` has succeeded.

jkitman:
Only writes change to the DB once a transaction has been submitted successfully, adds integration test coverage as well.

Note that this does not handle the case where the tx is rejected during consensus.

Fixes #1478

### DIFF
--- a/client/client-lib/src/lib.rs
+++ b/client/client-lib/src/lib.rs
@@ -341,13 +341,19 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
         self.submit_tx_with_change(tx, &mut rng).await
     }
 
-    async fn submit_tx_with_change<R: RngCore + CryptoRng>(
+    /// Submits a transaction to the fed, making change using our change module
+    ///
+    /// If the submission fails, no changes to the DB will occur.
+    pub async fn submit_tx_with_change<R: RngCore + CryptoRng>(
         &self,
         tx: TransactionBuilder,
         rng: R,
     ) -> Result<TransactionId> {
-        let final_tx = tx.build(self, rng).await;
-        Ok(self.context.api.submit_transaction(final_tx).await?)
+        let mut dbtx = self.context.db.begin_transaction().await;
+        let final_tx = tx.build(self, &mut dbtx, rng).await;
+        let result = self.context.api.submit_transaction(final_tx).await?;
+        dbtx.commit_tx().await.expect("DB Error");
+        Ok(result)
     }
 
     /// Spent some [`SpendableNote`]s to receive a freshly minted ones
@@ -522,15 +528,10 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
             notes
         } else {
             let mut tx = TransactionBuilder::default();
-            let change = vec![amount, notes.total_amount() - amount];
 
             let (mut keys, input) = MintClient::ecash_input(notes)?;
             tx.input(&mut keys, input);
-            let tx = tx
-                .build_with_change(self.mint_client(), rng, change, &self.context.secp)
-                .await;
-
-            self.context.api.submit_transaction(tx).await?;
+            self.submit_tx_with_change(tx, rng).await?;
             self.fetch_all_notes().await;
             self.mint_client().select_notes(amount).await?
         };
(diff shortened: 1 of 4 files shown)


client/client-lib/src/lib.rs
```diff
@@ -341,13 +341,19 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
...
+    pub async fn submit_tx_with_change<R: RngCore + CryptoRng>(
         &self,
         tx: TransactionBuilder,
         rng: R,
     ) -> Result<TransactionId> {
-        let final_tx = tx.build(self, rng).await;
-        Ok(self.context.api.submit_transaction(final_tx).await?)
+        let mut dbtx = self.context.db.begin_transaction().await;
+        let final_tx = tx.build(self, &mut dbtx, rng).await;
+        let result = self.context.api.submit_transaction(final_tx).await?;
```
dpc, line 354: This has an opposite problem. submit can succeed, while commit might fail / never happen.

dpc:
For full correctness we can't avoid some kind of log/WaL/outbox pattern thing, where we first make one big db tx that transitions the state and writes out all side effects first, then in a separate step (even thread) we deliver side effect to outher systems.

We should probably think about it form ground up for the client state machines 
@mxz42 . This *very* common consideration in distributed systems.

Check out https://www.youtube.com/watch?v=tcePbob8rrY - the exact two same lines are being talked over here. (I recommend the whole channel)


I generally thing that log-based (aka event based) is the best and simplest approach to managing this sort of stuff in distributed systems like fedimint is. In place mutations are just to fragile and bug-prone. I'm not sure if you're familiar with even-sourcing, but I'm tempted to recommend using (or something based on the idea) it for client side state management. If anyone wants to talk about it, let me know. (also ping @elsirion since it's an important architectural consideration)

client/client-lib/src/lib.rs
dpc: Note: Even though not "correct", this change is probably better than what we have before. So I'm not saying that we shouldn't land it, just that it's not exactly right approach either.

elsirion:
That event sourcing approach sounds like "state machine arhitecture that keeps previous states and their transitions in a log". Fine by me (I thought about deleting old states after transition, but why not keep a log just in case).

elsirion (approved):
Looks good.

elsirion:
@justinmoon, is that what you meant?

dpc:
> That event sourcing approach sounds like "state machine arhitecture that keeps previous states and their transitions in a log". Fine by me (I thought about deleting old states after transition, but why not keep a log just in case).


In event sourcing one only keeps the events in the database and not the state itself. The state is rebuilt from the events each time. That's the most important characteristic. When this becomes infeasible, snapshots are used to periodically compact the very old events.  The advantages of this approach is relative rigid simplicity, resilience against race conditions, deterministic replay, ability to debug things, ability to fix the state from existing events, etc.

But it's one of the most heavy approaches mentioned here. I don't know if ES is really necessary for us. I just think it's worth knowing about.

justinmoon:
> @justinmoon, is that what you meant?

Yes

jkitman:
Yeah to do this properly we need to model the state machine better.  The client should store outpoints before submission, resubmit them if no response was received, and remove them if the response is an error.

justinmoon:
After thinking about this behavior a bit more, I wonder if this behavior is even worse. The old behavior would pollute our DB with issuances that would never be completed if transactions fails. But this could lead to valid issuances in the federation which our DB doesn't know about.

dpc:
For pegins it might be much worse, as we don't even have the tweak in the database to recover from.

jkitman:
Feels like we should undo this change since with #1489 the client won't block forever, it'll just wait 2 secs and return the errors.

But please don't revert the commit, I'd prefer to keep the refactoring and just change the order of things.

justinmoon:
@jkitman agreed. After https://github.com/fedimint/fedimint/pull/1489 the problem is much less severe.

PR #1496, opened 2023-01-23
Don't fail if gw LN node is direct recipient of payment

elsirion:
The first commit reproduces the problem in CI, the second one fixes it, the third one is out of abundance of caution.

### DIFF
--- a/gateway/ln-gateway/src/cln.rs
+++ b/gateway/ln-gateway/src/cln.rs
@@ -44,7 +44,8 @@ pub struct Htlc {
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Onion {
     pub payload: String,
-    pub short_channel_id: String,
+    #[serde(default)]
+    pub short_channel_id: Option<String>,
     #[serde(deserialize_with = "as_fedimint_amount")]
     pub forward_msat: Amount,
     pub outgoing_cltv_value: u32,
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/cln.rs
```diff
@@ -33,24 +33,18 @@ where
 // See: https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md#htlc_accepted
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Htlc {
-    pub short_channel_id: String,
     #[serde(deserialize_with = "as_fedimint_amount")]
     pub amount_msat: Amount,
+    // TODO: use these to validate we can actually redeem the HTLC in time
     pub cltv_expiry: u32,
```
elsirion, line 39: @okjodom We need to verify that we have enough time to buy the preimage by checking the CLTV expiry here. I don't understand the details, it's probably best to read the BOLTs and talk to LDK.

gateway/ln-gateway/src/cln.rs
```diff
@@ -44,7 +44,8 @@ pub struct Htlc {
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Onion {
     pub payload: String,
-    pub short_channel_id: String,
+    #[serde(default)]
+    pub short_channel_id: Option<String>,
```
elsirion, line 47: @okjodom Be sure to do the same for the new GW.

gateway/ln-gateway/src/cln.rs
okjodom: ack.

gateway/ln-gateway/src/cln.rs
okjodom: Removed all other fields because we don't use them?

okjodom (approved):

gateway/ln-gateway/src/cln.rs
justinmoon: For a gateway we're running right,`short_channel_id` is occasionally missing. This leads to a panic because `serde_json` expects it to be there.

Removing other fields because we don't use them and don't want to panic in a similar way in case they're not there.

PR #1497, opened 2023-01-23
CLN extension allows node to receive payment

okjodom:
Applies fix introduced by #1496 to allow gateway node to receive payment not meant for federations

Bonus commit to make a note on how to enable debug logging

### DIFF
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -36,6 +36,7 @@ pub struct ClnExtensionOpts {
 }
 
 // Note: Once this binary is stable, we should be able to remove current 'ln_gateway'
+// Use CLN_PLUGIN_LOG=<log-level> to enable debug logging from within cln-plugin
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
     let (service, listen) = ClnRpcService::new()
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -81,13 +81,10 @@ pub struct Htlc {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Onion {
-    pub payload: String,
-    pub short_channel_id: u64,
+    #[serde(default)]
+    pub short_channel_id: Option<u64>,
     #[serde(deserialize_with = "as_fedimint_amount")]
     pub forward_msat: Amount,
-    pub outgoing_cltv_value: u32,
-    pub shared_secret: bitcoin_hashes::sha256::Hash,
-    pub next_onion: String,
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/bin/ln_gateway.rs
```diff
@@ -29,8 +29,8 @@ use tokio::sync::mpsc;
 use tracing::error;
 
 /// Fedimint gateway packaged as a CLN plugin
+// Use CLN_PLUGIN_LOG=1 to enable debug logging from within cln-plugin
```
dpc, line 32: I think it take same values as `RUST_LOG` (just under different name). https://github.com/ElementsProject/lightning/blob/acfb63e4bf19751167fbc69448f236ae413e690f/plugins/src/logging.rs#L56

okjodom: Ah cool. Give it a log level: https://docs.rs/log/latest/log/enum.Level.html

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -81,13 +81,10 @@ pub struct Htlc {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Onion {
-    pub payload: String,
-    pub short_channel_id: u64,
+    #[serde(default)]
+    pub short_channel_id: Option<u64>,
     #[serde(deserialize_with = "as_fedimint_amount")]
     pub forward_msat: Amount,
```
dpc, line 88:
```
2023-01-24T00:18:39.224Z **BROKEN** plugin-ln_gateway-logging: htlc_accepted error missing field `forward_msat`
```

:shrug:

dpc: I was trying without this field either now, but I think I got rate-limited now for too many failed payments.

okjodom: > ```
> 2023-01-24T00:18:39.224Z **BROKEN** plugin-ln_gateway-logging: htlc_accepted error missing field `forward_msat`
> ```
> 
> shrug

`cln-extension` is out of path for your testing. yes, our current version of `cln-plugn` used in `ln-gateway` (that you have deployed) will this this error if you expected `forward_msat` in onion

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -81,13 +81,10 @@ pub struct Htlc {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Onion {
-    pub payload: String,
-    pub short_channel_id: u64,
+    #[serde(default)]
+    pub short_channel_id: Option<u64>,
```
elsirion, line 86: FYI, you'll have to add a custom parsing fn here since we get the scid in `<block>x<tx>x<out_idx>` format and not as a string.

gateway/ln-gateway/src/bin/cln_extension.rs
elsirion: ~@dpc this log message seems pre-#1496? @okjodom is right, the code touched by this PR is not used anywhere.~ 

But it's a good point: why not remove `forward_msat` just like in #1496?

okjodom: It'd be a bigger change touching subscribe intercept grpc spec!

Proper fix which would save us a lot of code churn, should start with #1245.

So I'm with keeping that problematic field, or stashing this PR for a while then fix they whole type issue on one go

elsirion (approved):

gateway/ln-gateway/src/bin/cln_extension.rs
elsirion: Follow up: look into if `forward_msat` also has to be optional. Maybe make the fields only present on intermediate hops a separate struct inside an `Option`. Would be good to get feedback from CLN, which #1245 will help getting.

gateway/ln-gateway/src/bin/cln_extension.rs
okjodom: ack. pending

PR #1499, opened 2023-01-24
Change UI form field from finality delay to block confirmations

### LINKED ISSUE #1277: Setup UI should ask for "block confirmations required" instead of "finality delay"
AFAICT you need need to wait N+1 blocks to execute a peg-in with a `finality_delay` of N

asifshiraz:
Fixes #1277

### DIFF
--- a/fedimintd/src/ui.rs
+++ b/fedimintd/src/ui.rs
@@ -269,7 +269,7 @@ pub struct ParamsForm {
     network: Network,
     /// The number of confirmations a deposit transaction requires before accepted by the
     /// federation
-    finality_delay: u32,
+    block_confirmations: u32,
 }
 
 #[debug_handler]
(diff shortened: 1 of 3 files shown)


jkitman (dismissed):
LGTM, assuming you tested it manually?

asifshiraz:
> LGTM, assuming you tested it manually?

I got a seemingly unrelated error when testing, so I assumed this is working okay. I got an error "Cert string has wrong number of fields".

fedimintd/src/ui.rs
```diff
@@ -301,7 +301,8 @@ async fn post_federation_params(
             name: form.guardian_name,
             tls_connect_string,
         },
-        finality_delay: form.finality_delay,
+        // finality delay is always one less than required block confirmations
+        finality_delay: form.block_confirmations - 1,
```
elsirion, line 305: This will panic if `block_confirmations` is 0. Front end validation should not be considered secure. Use [`checked_sub`](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_sub) instead to generate an error return value instead of a panic.

elsirion (changes requested):

elsirion (approved):

PR #1505, opened 2023-01-24
Fixes the reconnect test failures

jkitman:
Fixes flakiness in our reconnect due to not giving a peer enough time to rejoin before shutting down the others.


elsirion:
So this is essentially just a way to trigger more epochs?

elsirion (approved):

PR #1518, opened 2023-01-24
Saves to DB before tx is submitted, undoing #1495

jkitman:
Saves to DB before tx is submitted, undoing #1495.

Assumes that we will eventually have a better fix, but this is safer for now since whenever we submit a peg-in we need to make sure our notes are persisted first.

### DIFF
--- a/integrationtests/tests/tests.rs
+++ b/integrationtests/tests/tests.rs
@@ -143,6 +143,21 @@ async fn peg_outs_are_only_allowed_once_per_epoch() -> Result<()> {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn peg_ins_that_are_unconfirmed_are_rejected() -> Result<()> {
+    test(2, |_fed, user, bitcoin, _, _| async move {
+        let peg_in_address = user.client.get_new_pegin_address(rng()).await;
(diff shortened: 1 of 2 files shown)


integrationtests/tests/tests.rs
```diff
@@ -142,6 +142,21 @@ async fn peg_outs_are_only_allowed_once_per_epoch() -> Result<()> {
...
 
+#[tokio::test(flavor = "multi_thread")]
+async fn peg_ins_that_are_unconfirmed_are_rejected() -> Result<()> {
+    test(2, |_fed, user, bitcoin, _, _| async move {
+        let peg_in_address = user.client.get_new_pegin_address(rng()).await;
+        let (proof, tx) = bitcoin.send_and_mine_block(&peg_in_address, Amount::from_sat(10000));
+        let result = user.client.peg_in(proof, tx, rng()).await;
+
+        // TODO make return error more useful
+        assert!(result.is_err());
```
elsirion, line 153: The `result` is an error because no epochs are run, so the federation doesn't know about the new blocks yet I assume?

elsirion (approved):

elsirion:
Either way it's broken, this way around seems a little safer.

integrationtests/tests/tests.rs
jkitman: Right and enough blocks didn't pass.

PR #1523, opened 2023-01-25
Modularize dbdump

m1sterc001guy:
This PR refactors the dbdump tool in the following ways:

- Moves the logic to dbtool and removes dbdump. "Dump" is now a command of dbtool.
- Modularizes the dumping of the server module databases. dump_module_db was added as a function in ModuleGen. Each module must implement it now (or can just return an empty BTreeMap).
- Logic for dumping each module's database was moved into each module.


modules/fedimint-ln/src/lib.rs
```diff
@@ -342,6 +345,83 @@ impl ModuleGen for LightningGen {
     ) -> anyhow::Result<bitcoin_hashes::sha256::Hash> {
         serde_json::from_value::<LightningClientConfig>(config)?.consensus_hash()
     }
+
+    async fn dump_module_database(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        prefixes: Vec<String>,
```
dpc, line 352: Should we use `Vec<u8>` for prefixes?

m1sterc001guy: It's actually just a string that refers to a specific prefix (e.g "Accepted Transaction"). Figured it was easier to write on the command line than translating what the actual prefix is.

Maybe column is a better name?

m1sterc001guy: Changed to prefix_names

modules/fedimint-wallet/src/lib.rs
```diff
@@ -359,6 +365,81 @@ impl ModuleGen for WalletGen {
...
     }
+
+    async fn dump_module_database(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        prefix_names: Vec<String>,
+    ) -> Box<dyn Iterator<Item = (String, Box<dyn erased_serde::Serialize + Send>)> + '_> {
+        let mut wallet: BTreeMap<String, Box<dyn erased_serde::Serialize + Send>> = BTreeMap::new();
+        for table in DbKeyPrefix::iter() {
+            filter_prefixes_by_name!(table, prefix_names);
```
dpc, line 376: This macro expanding to something that alerts the control flow (`continue`) is very sneaky. IMO, the macro should contain only the check, `if matches_smth_smth!(...) { continue; }`.

fedimint-api/src/db/mod.rs
```diff
@@ -708,6 +708,52 @@ impl DecodingError {
     }
 }
 
+#[macro_export]
+macro_rules! filter_prefixes_by_name {
+    ($prefix:ident, $prefixes:ident) => {
+        if !$prefixes.is_empty() && !$prefixes.contains(&$prefix.to_string().to_lowercase()) {
```
dpc, line 714: This `is_empty` check is completely redundant, no? If so I wonder if there's even a point to have this macro.

fedimint-api/src/module/mod.rs
```diff
@@ -353,6 +359,12 @@ pub trait ModuleGen: Debug + Sized {
     fn validate_config(&self, identity: &PeerId, config: ServerModuleConfig) -> anyhow::Result<()>;
 
     fn hash_client_module(&self, config: serde_json::Value) -> anyhow::Result<sha256::Hash>;
+
+    async fn dump_module_database(
```
dpc, line 363: This shouldn't be on `ModuleGen`, but on `Module`, no? We want to dump things from a specific module instance.

dpc: Maybe add `asyn fn list_db_prefix_keys() -> Vec<String>` so the user can just list these.

dpc: `_module_` here is redundant, IMO.

fedimint-dbtool/src/main.rs
```diff
@@ -33,6 +39,15 @@ enum DbCommand {
         #[arg(value_parser = hex_parser)]
         key: Bytes,
     },
+    Dump {
```
dpc, line 42: Docstring missing.

fedimint-api/src/module/mod.rs
m1sterc001guy: > This shouldn't be on `ModuleGen`, but on `Module`, no? We want to dump things from a specific module instance.

Hmm do we need to? I think if we want to move it to `Module`, we'd need to instantiate each module (`FedimintConsensus::new`), which for some modules, like the wallet module, instantiates new threads and does other initialization. In this case we just want to read the database, we don't actually need to instantiate the module. The current code does dump things from a specific module instance, it just grabs it from the server config.

I don't have a strong opinion but I will change it if you feel its a better design.

fedimint-api/src/db/mod.rs
m1sterc001guy: I removed the entire macro

modules/fedimint-wallet/src/lib.rs
m1sterc001guy: Macro removed

fedimint-api/src/module/mod.rs
dpc: That's true. But then - shouldn't we pass the instance id here, so the implementation knows *which* module to dump (there might be multiple of the same kind). Or is this already handled by `dbtx` being isolated?

modules/fedimint-wallet/src/lib.rs
m1sterc001guy: Yeah, think I accidentally lost a commit. Will fix.

fedimint-api/src/module/mod.rs
m1sterc001guy: Yep, this is handled by `dbtx` since it is isolated. When the returned data is added to the BTreeMap that is eventually serialized as JSON, it is added as "<kind>-<module instance id>" so when looking at the dump output it is easy to tell the difference (e.g. Mint-2)

dpc (dismissed):

elsirion (approved):

Issue #1205, opened 2023-01-05
New stable threshold key/id for authenticating configuration


jkitman:
I've been trying to aggregate issues related to the idea of adding a new stable threshold key/id to the federation for authenticating configuration.  We can handle these as tasks in roughly the following order:
- [x] Generate new threshold `auth_pk_set` (exchanging TLS certs would still be necessary for bootstrapping DKG)
- [x] Use the `auth_pk_set` as a stable id that all peers have contributed to
- #754
- [x] Expose an endpoint for retrieving consensus/client configs (which includes version) signed by the `auth_pk_set`
- #950
- [ ] Clients can consider the `auth_pk_set` the only "trusted" information, minimizing the serialized "connect info" (although they also need an endpoint to download the configs from)
- #614
- #755
- [ ] Support updating p2p certs or endpoints using the auth key share (cryptographic id) for each peer
- #744
- [ ] Changing peer membership could be handled by threshold-signing the new `auth_pk_set`

dpc:
:100: :partying_face:

PR #1527, opened 2023-01-25
Verify client config

### LINKED ISSUE #755: How can user authenticate a federation?
Currently a user acquires a list of guardian server websocket addresses, connects to them, and downloads a config containing all details to interact with the federation.

But how should the user validate that these servers are actually controlled by the federation guardians and not sybils?

jkitman:
This code has the federation sign a hash of the client config so that the config can be validated with only the federation id.

Related to #1205
Fixes #755


fedimint-server/src/consensus/mod.rs
```diff
@@ -443,6 +453,69 @@ impl FedimintConsensus {
...
+        dbtx: &mut DatabaseTransaction<'_>,
+        outcome: &HbbftConsensusOutcome,
+        drop_peers: &mut Vec<PeerId>,
+    ) {
+        let config = self.get_config_with_sig(dbtx).await;
+
+        if config.client_hash_signature.is_none() {
+            let maybe_client_hash = config.client.consensus_hash(&self.module_inits);
+            let client_hash = maybe_client_hash.expect("hashes");
+            let peers: Vec<PeerId> = outcome.contributions.keys().cloned().collect();
```
elsirion, line 469: Var name could be more descriptive, I suspected a bug where we ban non-contributing peers at first.

elsirion (approved):

elsirion:
Great change, just going ahead with merging :)

PR #1535, opened 2023-01-26
Remove confirmed BTC transactions from db

### LINKED ISSUE #1534: Peg-out transaction is not being removed from DB after it matures

elsirion:
Once we have recognized one of our transactions on-chain and have saved change to the DB we can stop trying to submit the transaction to our bitcoin node and forget about it.

Fixes #1534


dpc (approved):

jkitman (approved):
Looks good, since it's consensus height the block should have enough confirmations.

Issue #172, opened 2022-06-24
Disentangle API from `MinimintConsensus`


elsirion:
Currently the API task holds a reference to `MinimintConsensus` which thus has to be `Sync`. This is problematic for #79 and generally not a nice design imo. There should be some sort of accessor object that only requires access to the DB (which is already `Sync`).

justinmoon:
What's the status here?

elsirion:
It's still TODO as we are still passing module references/the consensus struct into the API handler closures, e.g.:

https://github.com/fedimint/fedimint/blob/08c3c35c6997c52fd7c443fcea6e91a70eec65ba/modules/fedimint-ln/src/lib.rs#L859-L865

PR #1536, opened 2023-01-26
Moves tx submission into a channel

jkitman:
This code has the federation sign a hash of the client config so that the config can be validated with only the federation id.

Attempts to use `Peekable`.

Related to https://github.com/fedimint/fedimint/issues/1205
Fixes https://github.com/fedimint/fedimint/issues/755


fedimint-server/src/consensus/mod.rs
```diff
@@ -153,14 +162,19 @@ impl FedimintConsensus {
...
+        (
+            Self {
+                modules,
+                cfg,
+                module_inits,
+                db,
+                tx_sender,
+            },
+            tx_receiver,
+        )
```
elsirion, line 177: I feel like we need to overthink the `FedimintConsensus` and `FedimintServer` border. Returning the receiver separately is very weird. No blocker for this PR though.

fedimint-server/src/consensus/mod.rs
```diff
@@ -239,18 +253,26 @@ impl FedimintConsensus {
...
+    /// For saving a tx, should be done prior to making a consensus proposal
+    pub async fn save_transaction_to_db(&self, transaction: Transaction) {
+        let mut dbtx = self.db.begin_transaction().await;
+
         let new = dbtx
-            .insert_entry(&ProposedTransactionKey(tx_hash), &transaction)
+            .insert_entry(&ProposedTransactionKey(transaction.tx_hash()), &transaction)
             .await
             .expect("DB error");
         dbtx.commit_tx().await.expect("DB Error");
```
elsirion, line 271: Follow up: We should probably make every `commit_tx` an `autocommit` call eventually.

elsirion (approved):
This got so much nicer with `peek` and `next().now_or_never()`!

elsirion:
Sorry, apparently there's a conflict now :(

fedimint-server/src/consensus/mod.rs
jkitman: Agreed, could not put receiver in `FedimintConsensus` without wrapping it in a `Mutex` which seems messier.  Right now feels like it could be broken into an interface between the API thread / consensus thread AND the outcome processing code.

fedimint-server/src/consensus/mod.rs
jkitman: Why is that?

justinmoon (approved):

fedimint-server/src/consensus/mod.rs
elsirion: See #172

fedimint-server/src/consensus/mod.rs
dpc: Always better to report an error and retry then crash, IMO.

PR #1539, opened 2023-01-26
fix: Remove unused /offers module from lightning module

justinmoon:


okjodom (approved):

modules/fedimint-ln/src/lib.rs
```diff
@@ -872,12 +872,6 @@ impl ServerModule for Lightning {
                         .ok_or_else(|| ApiError::not_found(String::from("Contract not found")))
                 }
             },
-            api_endpoint! {
-                "/offers",
```
dpc, line 876: Please check if don't have corresponding method in the `IFedimintApi` super-traits.

justinmoon: I grepped the codebase for `/offers` and this was the only hit

PR #1529, opened 2023-01-26
ci: update to use dtolnay instead of actions-rs

oleonardolima:
## Description

As described in https://github.com/rust-bitcoin/rust-bitcoin/pull/1509 and https://github.com/actions-rs/toolchain/issues/216 the actions-rs is unmaintained.

A different [Github Action](https://github.com/dtolnay/rust-toolchain ) maintained by dtolnay rust developer can be used instead.

This PR updates to use the `stable` and `nightly` toolchain instead of the action-rs one, and also need to run a `cargo build` and `cargo clippy` command as the dtolnay's rust-toolchain action does not have a cargo build/clippy one.

Fixes #1172

PR #1541, opened 2023-01-27
ci: update to use dtolnay instead of actions-rs

oleonardolima:
ci: update to use dtolnay instead of actions-rs

ci: use nightly instead of passing as `with`


elsirion:
Superseded by #1529

PR #1542, opened 2023-01-27
Adds a test and better error for spend_ecash

jkitman:
We saw an issue where `spend_ecash` was called from multiple threads which was unsafe.  But this adds better error message and test coverage.

### DIFF
--- a/client/client-lib/src/lib.rs
+++ b/client/client-lib/src/lib.rs
@@ -61,6 +61,7 @@ use fedimint_core::{
 use fedimint_derive_secret::{ChildId, DerivableSecret};
 use futures::stream::FuturesUnordered;
 use futures::StreamExt;
+use itertools::{Either, Itertools};
 use lightning::ln::PaymentSecret;
 use lightning::routing::gossip::RoutingFees;
 use lightning::routing::router::{RouteHint, RouteHintHop};
@@ -545,6 +546,7 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
         rng: R,
     ) -> Result<TieredMulti<SpendableNote>> {
         let notes = self.mint_client().select_notes(amount).await?;
+        let mut dbtx = self.context.db.begin_transaction().await;
 
         let final_notes = if notes.total_amount() == amount {
             notes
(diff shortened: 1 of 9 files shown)


client/client-lib/src/lib.rs
```diff
@@ -551,11 +551,9 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
             self.fetch_all_notes().await;
```
elsirion, line 551: I think there is at least another, if not a totally different root cause: `fetch_all_notes` times out now, so if the federation is too slow we don't get our freshly-split e-cash notes and then select a bigger amount.

jkitman: Good point, I'll convert `fetch_all_notes` to return a `Result` to force callers to handle possible failures with it.

client/client-lib/src/lib.rs
```diff
@@ -554,14 +555,12 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
...
+            self.fetch_all_notes().await?;
             self.mint_client().select_notes(amount).await?
         };
-        assert_eq!(
-            final_notes.total_amount(),
-            amount,
-            "should have exact change"
-        );
+        if final_notes.total_amount() != amount {
+            return Err(ClientError::SpendReusedNote);
```
elsirion, line 562: I don't get how this is the appropriate error here, might be useful to add a comment? How would `SpendReuseNote` lead to `final_notes.total_amount() != amount`?

client/client-lib/src/lib.rs
```diff
@@ -554,14 +555,12 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
             let (mut keys, input) = MintClient::ecash_input(notes)?;
             tx.input(&mut keys, input);
             self.submit_tx_with_change(tx, rng).await?;
-            self.fetch_all_notes().await;
+            self.fetch_all_notes().await?;
```
elsirion, line 558: Failing if a part fails is not optimal here imo:

1. The failure could be from another tx
2. The timeout is too tight to call fetch right after submit (2s, in real scenarios epochs take longer), so this will always fail. Needs to re-try imo.

client/client-lib/src/lib.rs
jkitman: If we fetched all the notes the only reason we wouldn't have enough at this point is if they were spent concurrently.

This is the failure that @justinmoon saw in the wild.

client/client-lib/src/lib.rs
elsirion: What I'd do:

1. Only fetch the two outputs that were just generated
2. Increase timeout so that the necessary epochs can happen in that time in realistic scenarios

elsirion (approved):

Issue #1530, opened 2023-01-26
Gateway handles requests serially


justinmoon:
Right now the gateway can't do 2 things at once.

For example, put a long sleep in `/pay_invoice`, pay an invoice via CLI inside tmuxinator, then call `gateway-cli balance`. It will hang until the sleep ends.

I'm guessing that `gateway-cli` hangs when anything is happening, and that perhaps the whole gateway can't handle concurrently requests.

dpc:
Is it because there is only one actor handling traffic and it's handling everything serially?

justinmoon:
I think so

dpc:
Spawning multiple handling loops (in a loop) should take care of it and seems like an easy fix. We will have to pick some "concurrency level".

dpc:
```
-        loop {                                                                                                
-            // Shut down main loop if requested                                                               
-            if loop_ctrl.is_shutting_down() {                                                                 
-                break;         
-            }                                                                                                 
                                                                                                               
-            let least_wait_until = Instant::now() + Duration::from_millis(100);
+        for _ in 0..32 {                                                                                      
+            loop {                                                                                            
+                // Shut down main loop if requested                                                           
+                if loop_ctrl.is_shutting_down() {  
+                    break;                                                                                    
+                }                  
                    
```

justinmoon:
This is risky because the existing client was never intended for concurrent usage

okjodom:
>```
> +        for _ in 0..32 {                                                                                      
> +            loop {                                                                                            
> +                // Shut down main loop if requested                                                           
> +                if loop_ctrl.is_shutting_down() {  
> +                    break;                                                                                    
> +                }    
>```
@dpc, this suggestion is not applicable unless we unless we change out messaging channel from webserver to processors, to use a 'many-to-many' pattern. We presently use an [mpsc::Receiver](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html) which can have may senders, but the receiver can only be owned by one thread. A choking point, maybe

m1sterc001guy:
This is fixed now :) https://github.com/fedimint/fedimint/pull/2654

PR #1543, opened 2023-01-27
Make the gateway fast

okjodom:
#1530 reports a critical limitation in the gateways, where requests like `/pay_invoice` are handled serially, and so the gateway is not ready to serve multiple users at once!

## Experiment.

Make gateway `/info` super slow, and try to issue multiple requests from `gateway-cli info` tool.

- I used a `10s` sleep in gateway handler for info requests
```fedimint_api::task::sleep_until(Instant::now() + Duration::from_millis(10000)).await;```
- Issued 3 requests to source gateway info. Numbers are representative. _Please account for some few seconds of human switching shell tabs to press ENTER on `gw info (gateway-cli info)` command_

## Observations:
- Before gateway request handler changes included in this PR, all the requests would be handled serially!
image
  - Client requests are queued up, with longer and longer wait times for each subsequent request
  - Would be a terrible experience for clients that are calling into the gateway for services like `/pay_invoice`
  
- After gateway request handler changes proposed  in this PR, the requests are now parallel
image
  - Client requests are handled in parallel using threads spawned by the gateway
  - Requests take as long as the logic of processing needs to operate. Simulated by the `10s` sleep
  
## Blocker
- To apply these changes as an improvement to the gateway, we need some client logic changes to allow for thread safety.
- https://github.com/fedimint/fedimint/blob/master/client/client-lib/src/lib.rs#L145-L150
```
#[derive(Copy, Clone)]
pub struct Client<C> {
    config: C,
    context: Arc<ClientContext>,
    #[allow(unused)]
    root_secret: DerivableSecret,
}
++++
```

### DIFF
--- a/client/client-lib/src/lib.rs
+++ b/client/client-lib/src/lib.rs
@@ -143,6 +143,7 @@ impl From<GatewayClientConfig> for LightningGateway {
     }
 }
 
+#[derive(Copy, Clone)]
 pub struct Client<C> {
     config: C,
     context: Arc<ClientContext>,
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -31,7 +31,6 @@ use mint_client::{
     api::WsFederationConnect, ln::PayInvoicePayload, mint::MintClientError, ClientError,
     GatewayClient,
 };
-use rpc::{BackupPayload, RestorePayload};
 use thiserror::Error;
 use tokio::sync::{mpsc, Mutex};
 use tracing::{debug, error, warn};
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by dpc** [comment]
Here is a script I used to to test parallel request handling in tmuxinator:

```
> cat stress.sh 
#!/usr/bin/env bash

set -e

export connection_string
connection_string="$($FM_MINT_CLIENT connect-info | jq -e -r '.connect_info')"

function pingpong() {

  local cli
  local dir
  
  dir="$TMP/cli-${RANDOM}${RANDOM}"
  cli="./target/debug/fedimint-cli --workdir $dir"

  echo "join"
  $cli join-federation "${connection_string}" > /dev/null

  for loop_i in $(seq 10) ; do
    echo "recv $loop_i"
    local invoice
    invoice=$($cli ln-invoice 13 "foo${RANDOM}${RANDOM}" | jq -e -r '.invoice')

    $FM_LN2 pay "${invoice}" >/dev/null || echo "recv: FAILED ln2 pay ${invoice}"

    local ok
    for _retry_i in $(seq 20) ; do
      if $cli wait-invoice "${invoice}" | jq -e -r '.paid_in_tx' > /dev/null ; then
        ok="true"
        break
      fi
      sleep .2
     done
  
    if [[ "$ok" != "true" ]] ; then
       echo "recv: FAILED wait: ${invoice}"
    fi

    $cli fetch > /dev/null

    echo "send $loop_i"
    invoice="$($FM_LN2 invoice 12 "foo${RANDOM}${RANDOM}" "bar" | jq -e -r '.bolt11')"
    $cli ln-pay "$invoice" > /dev/null || echo "send: FAILED: ${invoice}"

  done
  echo "done"
}

for _i in $(seq 30) ; do
  pingpong &
done

wait
```

Maybe it will come handy.

dpc:
@elsirion Has patches for this that go max further already.

okjodom:
Proposing as an improvement to on/after the new gateway design at #1337

PR #1549, opened 2023-01-27
fix: autocommit re-tries failed DB tx

elsirion:
We got a comparison the wrong way around, this commit adds a test so it doesn't break again and fixes the issue.

### DIFF
--- a/fedimint-api/src/db/mod.rs
+++ b/fedimint-api/src/db/mod.rs
@@ -1368,6 +1368,7 @@ mod tests {
         test_dbtx.commit_tx().await.expect("DB Error");
     }
 
+    #[cfg(test)]
     #[tokio::test]
     async fn test_autocommit() {
         use std::marker::PhantomData;
--- a/fedimint-api/src/db/mod.rs
+++ b/fedimint-api/src/db/mod.rs
@@ -164,7 +164,7 @@ impl Database {
                         Ok(()) => {
                             return Ok(val);
                         }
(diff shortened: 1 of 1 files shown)


dpc (approved):

fedimint-api/src/db/mod.rs
```diff
@@ -709,9 +709,17 @@ impl DecodingError {
 }
 
 mod tests {
```
dpc, line 711: `#[cfg(tests)]` missing (test? or tests?)

elsirion: It's not missing on the `mod`, maybe I should create a separate `mod` though. The problem is the existing test `fn`s aren't standalone, but just a template to test all other database implementatios.

elsirion:
Sorry for pulling you in @justinmoon, I think Intellij had a bug and added you as a commit author :laughing:

fedimint-api/src/db/mod.rs
elsirion: Clippy is happy after I put the relevant `use`s into the test `fn`.

dpc: I'm not sure if I get it. Should we rename this module if it's not a module with tests? `dbtest` or `testutil`? All other `mod tests` are exclusively for tests.

dpc (approved):

mxz42:
oh, wow. my bad. great that it's fixed and that there's a test now

justinmoon:
@elsirion I will take credit for your work any time!

PR #1550, opened 2023-01-28
Fix u32 overflow 

mxz42:
Fixes the u32 overflow mentioned in #1545

### DIFF
--- a/gateway/ln-gateway/src/utils.rs
+++ b/gateway/ln-gateway/src/utils.rs
@@ -5,15 +5,15 @@ use tracing::info;
 
 /// Retry an operation util the operation succeeds, OR
 /// The maximum number of attempts are made without success
-pub async fn retry<F, R, T>(
+pub async fn retry<F, Fut, T>(
     op_name: String,
     op_fn: F,
     wait: Duration,
     max_attempts: u32,
 ) -> Result<T, anyhow::Error>
 where
-    F: Fn() -> R,
-    R: Future<Output = Result<T, anyhow::Error>>,
+    F: Fn() -> Fut,
+    Fut: Future<Output = Result<T, anyhow::Error>>,
 {
     assert_ne!(max_attempts, 0, "max_attempts must be greater than 0");
     let mut attempts = 0;
--- a/gateway/ln-gateway/src/utils.rs
+++ b/gateway/ln-gateway/src/utils.rs
@@ -9,29 +9,29 @@ pub async fn retry<F, R, T>(
     op_name: String,
     op_fn: F,
     wait: Duration,
-    max_retries: u32,
+    max_attempts: u32,
 ) -> Result<T, anyhow::Error>
 where
     F: Fn() -> R,
     R: Future<Output = Result<T, anyhow::Error>>,
 {
-    let mut att = 0;
+    assert_ne!(max_attempts, 0, "max_attempts must be greater than 0");
+    let mut attempts = 0;
     loop {
+        attempts += 1;
         match op_fn().await {
             Ok(result) => return Ok(result),
-            Err(e) => {
-                att += 1;
-                if att > max_retries {
-                    return Err(e);
-                }
+            Err(err) if attempts < max_attempts => {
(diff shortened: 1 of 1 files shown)


mxz42:
> buildjet  lost communication with the server 

please restart CI. @elsirion tests run successfully locally

justinmoon:
Re-ran CI. In future you can just amend you last commit (`git commit --amend`), change the commit message slightly, and force-push to this branch and CI will re-run.

gateway/ln-gateway/src/utils.rs
```diff
@@ -5,23 +5,22 @@ use tracing::info;
...
 {
-    let mut att = 0;
+    let mut retries = 0;
     loop {
         match op_fn().await {
             Ok(result) => return Ok(result),
             Err(e) => {
-                att += 1;
-                if att > max_retries {
+                if retries >= max_retries {
```
dpc, line 23: https://llewellynfalco.blogspot.com/2016/02/dont-use-greater-than-sign-in.html

gateway/ln-gateway/src/utils.rs
```diff
@@ -5,23 +5,22 @@ use tracing::info;
...
+    F: Fn() -> Fut,
+    Fut: Future<Output = Result<T, anyhow::Error>>,
 {
-    let mut att = 0;
+    let mut retries = 0;
     loop {
         match op_fn().await {
             Ok(result) => return Ok(result),
             Err(e) => {
-                att += 1;
```
dpc, line 23: I thought "attempts" is preferable then "retries" and someone (you?) were fixing it in another place already, and I agreed with it.

`max_attempts` would be better, IMO. And incrementing right after (or even before `op_fn` make sure that the counter gets increased even if someone uses `continue` somewhere etc.

gateway/ln-gateway/src/utils.rs
mxz42: that is a fantastic post. thanks! will adhere to it in the future

gateway/ln-gateway/src/utils.rs
mxz42: right, yes, that was me if you refer to #1545. I didn't know you agreed! After thinking about this a bit longer I came to the conclusion that this might just be the way these `retry` function are supposed to work in fedimint. Always run once, and then up to `max_retries` times.

It was just counter-intuitive for me when I looked at it the first time, so I thought I found an off-by-one error. Though, as explained in the linked PR, it technically is not.

so, TLDR, I'll redo these changes with `max_attempts` and reopen the other PR #1545 and adjust the behavior there for consistency as well. 

Off the top of your head, do you know any other functions that haver similar behavior?

mxz42:
> Re-ran CI. In future you can just amend you last commit (`git commit --amend`), change the commit message slightly, and force-push to this branch and CI will re-run.

thanks! will do

gateway/ln-gateway/src/utils.rs
dpc: https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/config/retry/struct.RetryConfig.html#method.with_max_attempts

mxz42: cool. Is that just FYI or is there a suggestion for a change? Should we do something like a `RetryConfig` struct? 
I see they're using  exponential backoff (with jitter) for their `interval` between retries. That's kinda cool. 

When building the config, you can set the `max_attempts`. The comment mentions that this value has to be greater than zero. I don't see that they enforce it anywhere. I used an `assert_ne` for that. I think we want to have that?

```rust
    /// Set the maximum number of times a request should be tried, including the initial attempt.
    /// This value must be greater than zero.
    pub fn with_max_attempts(mut self, max_attempts: u32) -> Self {
        self.max_attempts = max_attempts;
        self
    }
```

mxz42:
in case you didn't notice, I pushed some changes @dpc

elsirion (approved):

PR #1551, opened 2023-01-28
Gatewayd Compat Setup

okjodom:
Cherry pick back-compatibility changes that will allow us to introduce the proposed `gatewayd` (#1337 ) while still supporting the current `ln_gateway`.

### Why make a fork of `LnGateway` : copy code of `src/lib.rs` to `src/gatewayd/gateway.rs` ?
- Note: `LnGateway` is renamed to `Gateway` in this new file
- so we can change the constructor signature of `LnGateway`
- so we can change how the gateway gets HTLC intercepts. `Gateway` in `src/gateway.rs` does not implement `handle_receive_payment`. Instead, actors directly subscribe to htlc intercepts on an entirely new pattern. More on this below

### Why make a fork of `GatewayActor` : copy code of `src/actor.rs` to `src/gatewayd/actor.rs` ?
- So we can change constructor signature of `GatewayActor`. Have the new actor take a reference to lnrpc, which is 
- So we can enable direct subscription to intercepted HTLCs using the owned lnrpc
- API signature of the `GatewayActor` changes because it owns a reference to lnrpc, rather than expec these to be passed in at api calls.
- Note: Changes incoming. You can preview them at #1337

### DIFF
--- a/gateway/ln-gateway/src/gatewayd/mod.rs
+++ b/gateway/ln-gateway/src/gatewayd/mod.rs
@@ -1,0 +1,13 @@
+// ### Why make a fork of `LnGateway` : copy code of `../src/lib.rs` to `./gateway.rs` ?
+// - Note: `LnGateway` is renamed to `Gateway` in this new file
+// - so we can change the constructor signature of `LnGateway`
+// - so we can change how the gateway gets HTLC intercepts. `Gateway` in `src/gateway.rs` does not implement `handle_receive_payment`. Instead, actors directly subscribe to htlc intercepts on an entirely new pattern. More on this below
+
+// ### Why make a fork of `GatewayActor` : copy code of `../src/actor.rs` to `./actor.rs` ?
+// - So we can change constructor signature of `GatewayActor`. Have the new actor take a reference to lnrpc, which is
+// - So we can enable direct subscription to intercepted HTLCs using the owned lnrpc
+// - API signature of the `GatewayActor` changes because it owns a reference to lnrpc, rather than expec these to be passed in at api calls.
+// - Note: Changes incoming. You can preview them at #1337
+
+pub mod actor;
+pub mod gateway;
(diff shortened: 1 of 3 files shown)


gateway/ln-gateway/src/gatewayd/mod.rs
```diff
@@ -0,0 +1,13 @@
+// ### Why make a fork of `LnGateway` : copy code of `../src/lib.rs` to `./gateway.rs` ?
+// - Note: `LnGateway` is renamed to `Gateway` in this new file
+// - so we can change the constructor signature of `LnGateway`
+// - so we can change how the gateway gets HTLC intercepts. `Gateway` in `src/gateway.rs` does not implement `handle_receive_payment`. Instead, actors directly subscribe to htlc intercepts on an entirely new pattern. More on this below
+
+// ### Why make a fork of `GatewayActor` : copy code of `../src/actor.rs` to `./actor.rs` ?
+// - So we can change constructor signature of `GatewayActor`. Have the new actor take a reference to lnrpc, which is
+// - So we can enable direct subscription to intercepted HTLCs using the owned lnrpc
+// - API signature of the `GatewayActor` changes because it owns a reference to lnrpc, rather than expec these to be passed in at api calls.
+// - Note: Changes incoming. You can preview them at #1337
```
justinmoon, line 10: Nit: I think we want 3 `///` in here so it would show up on documentation site.

justinmoon (approved):
I'm going to go ahead and merge this. While it isn't complete or tested yet, it doesn't interfere with anything else, is isolated to one module, and will allow @okjodom to iterate towards the new gateway design.

gateway/ln-gateway/src/gatewayd/mod.rs
okjodom: I actually didn't want these showing up in rust docs

comment says nothing about the structs. Just why we have them, relative to what we already have

PR #1552, opened 2023-01-28
fix(tracing): span request at trace level

Maan2003:


justinmoon:
As context, `ln_gateway` really spams `INFO    plugin-ln_gateway: request; peer=0` log messages. It makes the logs difficult to read.

For instance, search for `request;` in this CI just run just a little while ago: https://github.com/fedimint/fedimint/actions/runs/4033050341/jobs/6933489268

I see 60 occurrences of this log statement. And `cli-test.sh` just does one incoming and one outgoing lightning payment via gateway.

client/client-lib/src/api.rs
```diff
@@ -828,7 +828,7 @@ pub struct PeerResponse<R> {
 }
 
 impl<C: JsonRpcClient> FederationMember<C> {
-    #[instrument(fields(peer = %self.peer_id), skip_all)]
+    #[instrument(level = "trace", fields(peer = %self.peer_id), skip_all)]
```
elsirion, line 831: Since we are at it, what makes that log line especially annoying is not knowing anything about the request sent. Logging the `method` field would be very useful imo.

elsirion (approved):
Looks good

elsirion (approved):

PR #1556, opened 2023-01-31
feat: Fetch epochs during ecash recovery on 8 threads

dpc:
It seems that in practice fetching epochs take a long time leading to a bad user-experience. Fetching epochs from federation has a large latency, so it makes sense to do in on multiple-threads.

I selected 8 as a good compromise that should not create issues, like too much load, while hopefully giving most of the benefit of paralellism.

### DIFF
--- a/client/client-lib/src/mint/backup.rs
+++ b/client/client-lib/src/mint/backup.rs
@@ -219,12 +219,18 @@ impl MintClient {
     async fn fetch_epochs(
         &self,
         epoch_range: RangeInclusive<u64>,
+        divisor: u64,
+        reminder: u64,
         sender: mpsc::Sender<api::FederationResult<SignedEpochOutcome>>,
         task_handle: &TaskHandle,
     ) {
-        for epoch in epoch_range {
+        for epoch in epoch_range.clone() {
             if task_handle.is_shutting_down() {
                 break;
+            }
+
+            if (epoch - *epoch_range.start()) % divisor != reminder {
+                continue;
             }
 
             info!(epoch, "Fetching epoch");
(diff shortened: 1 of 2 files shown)


client/client-lib/src/mint/backup.rs
```diff
@@ -219,14 +219,20 @@ impl MintClient {
     async fn fetch_epochs(
         &self,
         epoch_range: RangeInclusive<u64>,
+        divisor: u64,
```
dpc, line 222: This is lame, but relatively simple.

client/client-lib/src/mint/backup.rs
```diff
@@ -219,14 +219,20 @@ impl MintClient {
...
+        reminder: u64,
```
elsirion, line 223: `s/reminder/remainder/`

client/client-lib/src/mint/backup.rs
```diff
@@ -272,19 +278,28 @@ impl MintClient {
...
+                    move |task_handle| async move {
+                        self_clone
+                            .fetch_epochs(epoch_range, num_threads, task_i, tx, &task_handle)
+                            .await
+                    }
+                })
+                .await;
+
+            rx_channels.push(rx);
+        }
```
elsirion, line 302: Instead of statically dispatching epochs to specific threads, creating a stream of epoch fetch futures and then calling [`buffered`](https://docs.rs/futures/0.3.26/futures/stream/trait.StreamExt.html#method.buffered) or [`buffer_unordered`](https://docs.rs/futures/0.3.26/futures/stream/trait.StreamExt.html#method.buffer_unordered) with the intended parallelism as argument seems way more elegant.

elsirion (commented):
While this *works* it's not elegant given all the async rust tooling at our disposal.

client/client-lib/src/mint/backup.rs
dpc: Oh, yeah. I just wasn't familiar with these and short on time, so did it "manually". I'll have to look into it and educate myself a bit.

dpc:
https://github.com/fedimint/fedimint/pull/1582 is better.

PR #1564, opened 2023-02-01
feat: add esplora block source support

### LINKED ISSUE #1187: Alternative block sources
We should support an option to use Esplora / Electrum as block sources. This will help us deploy our early "dogfooding" federations more quickly.

- [x] electrum support https://github.com/fedimint/fedimint/pull/1363
- [ ] esplora support

oleonardolima:
## Description

The bitcoin backend currently supported are both bitcoind and electrum (without integration tests at the moment), as mentioned in #1187 having multiple, easy to use, bitcoin rpc backend support is useful for dogfooding federations and easy setup/deploy.

That said, this PR aims to implement and support [Esplora](https://github.com/Blockstream/esplora) support, although as it's for Electrum it currently does not have any integration test (I'm working on that 😁).

## How
This adds a new implementation for the [`IBitcoindRpc`](https://github.com/fedimint/fedimint/blob/13a0e24e902f63fae682c4e4af6e02069621086a/fedimint-bitcoind/src/lib.rs#L27-L58) trait for Esplora client. However, it could be implemented by using `reqwest` and calling each endpoint available on the Esplora API, this PR uses the [rust-esplora-client](https://github.com/bitcoindevkit/rust-esplora-client) as it's abstract all that logic and makes it easy to use, which is maintained by BDK team.

fixes #1187

### DIFF
--- a/fedimint-bitcoind/src/bitcoincore_rpc.rs
+++ b/fedimint-bitcoind/src/bitcoincore_rpc.rs
@@ -56,6 +56,9 @@ pub fn make_bitcoin_rpc_backend(
             .context("bitcoind rpc backend initialization failed"),
         BitcoindRpcBackend::Electrum(url) => make_electrum_rpc(url, task_handle)
             .context("electrum rpc backend initialization failed"),
+        BitcoindRpcBackend::Esplora(_) => {
+            todo!("esplora rpc backend is currently not implemented yet")
+        }
     }
 }
 
--- a/fedimint-bitcoind/src/bitcoincore_rpc.rs
+++ b/fedimint-bitcoind/src/bitcoincore_rpc.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::fmt;
 
 use ::bitcoincore_rpc::bitcoincore_rpc_json::EstimateMode;
@@ -56,9 +57,7 @@ pub fn make_bitcoin_rpc_backend(
             .context("bitcoind rpc backend initialization failed"),
         BitcoindRpcBackend::Electrum(url) => make_electrum_rpc(url, task_handle)
             .context("electrum rpc backend initialization failed"),
-        BitcoindRpcBackend::Esplora(_) => {
-            todo!("esplora rpc backend is currently not implemented yet")
-        }
+        BitcoindRpcBackend::Esplora(url) => make_esplora_rpc(url, task_handle),
     }
 }
 
@@ -76,6 +75,13 @@ pub fn make_bitcoind_rpc(url: &Url, task_handle: TaskHandle) -> Result<DynBitcoi
 
 pub fn make_electrum_rpc(url: &Url, _task_handle: TaskHandle) -> Result<DynBitcoindRpc> {
     Ok(ElectrumClient::new(url)?.into())
+}
+
+pub fn make_esplora_rpc(url: &Url, task_handle: TaskHandle) -> Result<DynBitcoindRpc> {
+    let esplora_client = EsploraClient::new(url)?;
+    let retry_client = RetryClient::new(esplora_client, task_handle);
+
+    Ok(retry_client.into())
 }
 
 /// Wrapper around [`bitcoincore_rpc::Client`] logging failures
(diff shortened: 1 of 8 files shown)


fedimint-api/src/bitcoin_rpc.rs
```diff
@@ -67,3 +80,11 @@ pub fn fm_electrum_rpc_env_value_to_url(value: &OsStr) -> anyhow::Result<Url> {
         anyhow::format_err!("Url not ascii text")
     })?)?)
 }
+
+// TODO: (@oleonardolima) Should we use a single fn instead of 3, such as: `fm_backend_rpc_env_value_to_url` ?
```
dpc, line 84: Yeah. One function for all, then `FM_BITCOIND_RPC_DEFAULT_FALLBACK` moved to the `else` in `select_bitcoin_backend_from_envs`

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -75,6 +77,11 @@ pub fn make_electrum_rpc(url: &Url, _task_handle: TaskHandle) -> Result<DynBitco
     Ok(ElectrumClient::new(url)?.into())
 }
 
+// TODO: (@oleonardolima) Should we use a retry mechanism for esplora client ?
```
dpc, line 80: IIRC `ElectrumClient` already had one, so I skipped it. Check if `EsploraClient` has one too. 

We should probably unify the behavior there and just keep trying forever for all clients.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -292,3 +299,112 @@ impl IBitcoindRpc for ElectrumClient {
...
+impl IBitcoindRpc for EsploraClient {
+    fn backend_type(&self) -> BitcoinRpcBackendType {
+        BitcoinRpcBackendType::Esplora
+    }
+
+    async fn get_network(&self) -> Result<Network> {
+        let genesis_height: u32 = 0;
+        let genesis_hash = self.0.get_block_hash(genesis_height).await?;
+
+        // TODO: How could we extract the BlockHash to a constant, and use something like the `bitcoin::blockdata::constants::genesis_block` from rust-bitcoin ?
```
dpc, line 330: `const MAINNET_GENESIS_BLOCK_HEX : &'str = "..."`? Seems like a good idea to share. If `rust-bitcoin` has one, even better.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -292,3 +299,112 @@ impl IBitcoindRpc for ElectrumClient {
...
+        &self,
+        transaction: &Transaction,
+        height: u64,
+    ) -> Result<bool> {
+        if self.get_block_height().await? <= height {
+            bail!("Esplora backend does not contain the block at {height} yet");
+        }
+
+        let txid = transaction.txid();
+        match self.0.get_tx_status(&txid).await? {
```
dpc, line 405: This has different semantics then other implementations. This will respond `true` even if `transaction` was confirmed in a different block (even later one).

`was_transaction_confirmed_in` was a workaround for lack of `get_block` in Electrum. You don't have to implement it if `get_block` is available

modules/fedimint-wallet/src/lib.rs
```diff
@@ -1223,12 +1225,25 @@ impl Wallet {
...
                             .was_transaction_confirmed_in(&transaction.1.tx, height as u64)
                             .await
-                            .expect("bitcoin rpc backend failed")
+                            .expect("bitcoin electrum rpc backend failed")
                         {
                             self.recognize_change_utxo(dbtx, transaction.1).await;
                         }
                     }
                 }
+                BitcoinRpcBackendType::Esplora => {
```
dpc, line 1234: Do what `Bitcoind` is doing instead.

fedimint-bitcoind/src/bitcoincore_rpc.rs
elsirion: I don't think we need to re-try client creation, I didn't see any possibly transient error cases. After it's created we can just wrap it in a retry client.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -292,3 +299,112 @@ impl IBitcoindRpc for ElectrumClient {
...
+    fn new(url: &Url) -> anyhow::Result<Self> {
+        let builder = esplora_client::Builder::new(url.as_str());
+        let client = builder.build_async()?;
+        Ok(Self(client))
+    }
+}
+
+impl fmt::Debug for EsploraClient {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("EsploraClient")
```
elsirion, line 316: Would be interesting to see the API url if possible

fedimint-bitcoind/src/bitcoincore_rpc.rs
elsirion: You could actually filter by block height here, but I agree it's safer to make this `unimplemented`.

modules/fedimint-wallet/src/lib.rs
elsirion: Should be

```rust
BitcoinRpcBackendType::Bitcoind | BitcoinRpcBackendType::Esplora => {
```

above I guess?

fedimint-bitcoind/src/bitcoincore_rpc.rs
oleonardolima: > I don't think we need to re-try client creation, I didn't see any possible transient error cases. After it's created we can just wrap it in a retry client.

Agreed, as dpc mentioned it probably should unify and wrap it in a retry client. After checking it on esplora crate I'll just wrap it in a retry client.

fedimint-bitcoind/src/bitcoincore_rpc.rs
oleonardolima: Great point, I'll try to check if that's possible 😁

fedimint-bitcoind/src/bitcoincore_rpc.rs
oleonardolima: I know that rust-bitcoin exposes the fn based on the `Network` as a parameter, returning the `Block` 🤔.
I'll double-check if rust-bitcoin exposes its constants, otherwise, I'll add the constants.

fedimint-bitcoind/src/bitcoincore_rpc.rs
oleonardolima: Ah, totally missed the implementation semantics 😅. Sure, I'll use the `get_block` instead and keep `was_transaction_confirmed_in` unimplemented.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -292,3 +301,95 @@ impl IBitcoindRpc for ElectrumClient {
...
+        let genesis_height: u32 = 0;
+        let genesis_hash = self.0.get_block_hash(genesis_height).await?;
+
+        // TODO: How could we extract the BlockHash to a constant, and use something like the `bitcoin::blockdata::constants::genesis_block` from rust-bitcoin ?
+        let network = match genesis_hash.to_hex().as_str() {
+            "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f" => Network::Bitcoin,
+            "000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943" => Network::Testnet,
+            "00000008819873e925422c1ff0f99f7cc9bbb232af63a077a480a3633bee1ef6" => Network::Signet,
+            hash => {
+                warn!("Unknown genesis hash {hash} - assuming regtest");
```
justinmoon, line 337: This might be a bad assumption if people start running private signets. Don't have a better suggestion, though.

elsirion: It might not even be a problem, alt-net address prefixes are a mess: 

https://github.com/fedimint/fedimint/blob/2af4e6a02f8ca4fb42165bfed263c681c225d9f4/modules/fedimint-wallet/src/lib.rs#L1638-L1647

See also the [cleaner impl landed in rust-bitcoin by now](https://docs.rs/bitcoin/0.29.2/src/bitcoin/util/address.rs.html#818-830).

flake.nix
```diff
@@ -895,6 +894,8 @@
                 pkgs.rnix-lsp
                 pkgs-unstable.convco
                 pkgs.nodePackages.bash-language-server
+              ] ++ lib.optionals (!stdenv.isAarch64 || !stdenv.isDarwin) [
+                pkgs.semgrep
```
elsirion, line 898: Was this an accidental change by rebasing?

oleonardolima: I'll remove this by rebasing the branch, I've cherry-picked this one from #1600 because I needed it to use `git commit` and it hasn't landed at the time.

fedimint-bitcoind/src/bitcoincore_rpc.rs
oleonardolima: During development, I was wondering if this method would be needed if the network were an env var but I'm not sure if this is a better alternative and/or has already been a discussion in the past.

oleonardolima: Great, I'll take a look at the impl for the address and see if I can figure out an alternative based on that, otherwise, I'll just extract to a const for now.

fedimint-bitcoind/src/bitcoincore_rpc.rs
elsirion: Bump

oleonardolima: @elsirion I've tried to work on this one, it's possible to get the underlying URL, e.g https://blockstream.info/testnet/api as it can be accessed by the `self.0.url()` from `AsyncClient`.

But having access to each different path it's not possible, as it's only accessible from the [`Request` struct in `reqwest`](https://docs.rs/reqwest/latest/src/reqwest/async_impl/request.rs.html#526-549)

https://github.com/seanmonstar/reqwest/blob/bdd4db02447c68b6be528c25cde6fa1fc47bf8e1/src/async_impl/request.rs#L526-L549

Do you think that adding only the underlying API url would be enough?

dpc: I don't get it.

Does this work:

```
pub struct EsploraClient {
  client: esplora_client::AsyncClient,
  sanitized_url: String,
}
```

?

dpc: For `bitcoind`, I used:

```
 let retry_client = RetryClient::new(
        Client(ErrorReporting::new(url, bitcoind_client)),
        task_handle,
    );
```

to wrap the original client into something that reports error. You could do the similiar thing here.

oleonardolima: I mean this is possible:
``` rust
impl fmt::Debug for EsploraClient {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_fmt(format_args!("EsploraClient [url: {}]", self.0.url()))
    }
}
```

But it will only show the URL (eg. https://blockstream.info/testnet/api) and not the path being called.

dpc (dismissed):
LGTM, though probably best to squash, at least somewhat.

oleonardolima:
I'm not sure why requesting a re-review for one user is removing the request for another one 🤯

dpc (approved):

justinmoon (approved):

PR #1566, opened 2023-02-01
Use epoch count instead of last epoch

jp1ac4:
Hi, I've started working on this PR to address #1340.

I hope I'm on the right track. I'm quite new to this project and Rust in general so I'll be very grateful for any feedback :)

### DIFF
--- a/client/client-lib/src/api.rs
+++ b/client/client-lib/src/api.rs
@@ -368,7 +368,7 @@ pub trait GlobalFederationApi {
         decoders: &ModuleDecoderRegistry,
     ) -> FederationResult<SignedEpochOutcome>;
 
-    async fn fetch_last_epoch(&self) -> FederationResult<u64>;
+    async fn fetch_epoch_count(&self) -> FederationResult<u64>;
 
     async fn fetch_output_outcome<R>(
         &self,
@@ -457,8 +457,8 @@ where
         .await
     }
 
-    async fn fetch_last_epoch(&self) -> FederationResult<u64> {
-        self.request_eventually_consistent("/epoch".to_owned(), erased_no_param())
+    async fn fetch_epoch_count(&self) -> FederationResult<u64> {
+        self.request_eventually_consistent("/fetch_epoch_count".to_owned(), erased_no_param())
             .await
     }
 
--- a/client/client-lib/src/mint/backup.rs
+++ b/client/client-lib/src/mint/backup.rs
@@ -10,7 +10,7 @@
 use std::{
     cmp::{max, Reverse},
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
-    ops::RangeInclusive,
+    ops::Range,
 };
 
 use anyhow::Result;
@@ -138,7 +138,7 @@ impl MintClient {
             .collect();
 
         // Use the newest (highest epoch)
-        responses.sort_by_key(|backup| Reverse(backup.epoch));
+        responses.sort_by_key(|backup| Reverse(backup.epoch_count));
 
         Ok(responses.into_iter().next())
     }
@@ -170,7 +170,7 @@ impl MintClient {
 
     async fn prepare_plaintext_ecash_backup(&self) -> Result<PlaintextEcashBackup> {
         // fetch consensus height first - so we dont miss anything when scanning
-        let epoch = self.context.api.fetch_last_epoch().await?;
+        let epoch_count = self.context.api.fetch_epoch_count().await?;
 
         let mut dbtx = self.start_dbtx().await;
         let notes = self.get_available_notes(&mut dbtx).await;
@@ -191,7 +191,7 @@ impl MintClient {
             notes,
             pending_notes,
             next_note_idx,
-            epoch,
+            epoch_count,
         })
     }
 
@@ -217,7 +217,7 @@ impl MintClient {
     /// TODO: could be internal to recovery_loop?
     fn fetch_epochs_stream(
         &self,
-        epoch_range: RangeInclusive<u64>,
+        epoch_range: Range<u64>,
     ) -> impl futures::Stream<
         Item = (
             u64,
@@ -246,18 +246,17 @@ impl MintClient {
         backup: PlaintextEcashBackup,
         gap_limit: usize,
     ) -> Result<Cancellable<EcashRecoveryFinalState>> {
-        let end_epoch = match self.context.api.fetch_last_epoch().await {
+        let current_epoch_count = match self.context.api.fetch_epoch_count().await {
             Ok(v) => v,
             Err(e) => {
                 return Err(e.into());
             }
         };
-        let epoch_range = backup.epoch..=end_epoch;
+        // TODO: This -1 is probably not necessary, as it should be enough to start from the exact epoch the snapshot was taken, but it is harmless to start from any epoch in the past, and starting a bit earlier makes it more robust in face of some inconsistency that we've missed.
+        let start_epoch = backup.epoch_count.saturating_sub(1);
+        let epoch_range = start_epoch..current_epoch_count;
 
-        info!(
-            start_epoch = backup.epoch,
-            end_epoch, "Recovering from snapshot"
-        );
+        info!(start_epoch, current_epoch_count, "Recovering from snapshot");
 
         let mut tracker = EcashRecoveryTracker::from_backup(
             backup,
@@ -306,7 +305,7 @@ impl MintClient {
 pub struct PlaintextEcashBackup {
     notes: TieredMulti<SpendableNote>,
     pending_notes: Vec<(OutputFinalizationKey, NoteIssuanceRequests)>,
-    epoch: u64,
+    epoch_count: u64,
     next_note_idx: Tiered<NoteIndex>,
 }
 
@@ -316,7 +315,7 @@ impl PlaintextEcashBackup {
         Self {
             notes: TieredMulti::default(),
             pending_notes: vec![],
-            epoch: 0,
+            epoch_count: 0,
             next_note_idx: Tiered::default(),
         }
     }
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -545,14 +545,15 @@ impl FedimintConsensus {
         client
     }
 
-    pub async fn get_last_epoch(&self) -> Option<u64> {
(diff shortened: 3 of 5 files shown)


client/client-lib/src/api.rs
```diff
@@ -452,7 +452,7 @@ where
         .await
     }
 
-    async fn fetch_last_epoch(&self) -> FederationResult<u64> {
+    async fn fetch_epoch_count(&self) -> FederationResult<u64> {
         self.request_eventually_consistent("/epoch".to_owned(), erased_no_param())
```
dpc, line 456: We should rename the api call as well while at it.

client/client-lib/src/mint/backup.rs
```diff
@@ -259,18 +259,20 @@ impl MintClient {
...
-        let end_epoch = match self.context.api.fetch_last_epoch().await {
+        let epoch_count = match self.context.api.fetch_epoch_count().await {
             Ok(v) => v,
             Err(e) => {
                 return Err(e.into());
             }
         };
-        let epoch_range = backup.epoch..=end_epoch;
+        let start_epoch = if backup.epoch_count > 0 {
+            backup.epoch_count - 1
```
dpc, line 269: Handle underflow with `saturating_sub`

fedimint-server/src/consensus/mod.rs
```diff
@@ -538,14 +538,20 @@ impl FedimintConsensus {
         client
     }
 
-    pub async fn get_last_epoch(&self) -> Option<u64> {
-        self.db
+    pub async fn get_epoch_count(&self) -> u64 {
+        let maybe_e = self
```
dpc, line 542: `e` is a confusing name here.

fedimint-server/src/consensus/mod.rs
```diff
@@ -538,14 +538,20 @@ impl FedimintConsensus {
...
+            .db
             .begin_transaction()
             .await
             .get_value(&LastEpochKey)
             .await
-            .expect("db query must not fail")
-            .map(|e| e.0)
+            .expect("db query must not fail");
+
+        if let Some(e) = maybe_e {
```
dpc, line 550: `.unwrap_or` would be shorter

client/client-lib/src/api.rs
jp1ac4: I've now renamed it to `/fetch_epoch_count`. Could also just use `/epoch_count` if you prefer?

client/client-lib/src/mint/backup.rs
jp1ac4: Thanks for this tip. I've changed that now.

fedimint-server/src/consensus/mod.rs
jp1ac4: I've changed it to `ep_hist_key` when using `.map` below.

fedimint-server/src/consensus/mod.rs
jp1ac4: Thanks, I've rewritten this part now.

client/client-lib/src/mint/backup.rs
```diff
@@ -259,18 +259,16 @@ impl MintClient {
...
-        let end_epoch = match self.context.api.fetch_last_epoch().await {
+        let epoch_count = match self.context.api.fetch_epoch_count().await {
             Ok(v) => v,
             Err(e) => {
                 return Err(e.into());
             }
         };
-        let epoch_range = backup.epoch..=end_epoch;
+        let start_epoch = backup.epoch_count.saturating_sub(1);
+        let epoch_range = start_epoch..epoch_count;
```
dpc, line 269: This reads confusingly now, IMO. 

```
let epoch_range = backup.epoch_count.saturating_sub(1)..current_epoch_count;
```

would be best, I guess. We might need to rename the field in the `PlaintextEcashBackup`, but I think that's a good idea anyway, as it is more precise.

dpc: BTW, that `saturating_sub(1)` is probably unnecessary, but I added it because I didn't have confidence this is the case. I should have probably added a comment.

Can you please add `// TODO: This -1 is probably not neccessary, as it should be enough to start from the exact epoch the snapshot was taken, but it is harmless to start from any epoch in the past.`. Thanks.

dpc (commented):
Other than this one readability comment, LGTM. Please mark as "Ready for Review" when done and lets have someone else take a look as well.

client/client-lib/src/api.rs
dpc: We'll have to rename all of these soon anyway. There's no reason for them to even start with a `/` to begin with, and I think I'd prefer to make them look as much as bitcoin core as possible, so `getepochcount`. But probably not a job for this PR.

jp1ac4: OK thanks for explaining that. I'll keep it as `/fetch_epoch_count` for this PR.

client/client-lib/src/mint/backup.rs
jp1ac4: Thanks for your comments.

Regarding the `-1`, if I understand correctly the backup contains `backup.epoch_count` epochs, with the last backup epoch being `backup.epoch_count - 1`, and so with the `-1` we'll start from the same epoch the backup was taken. If I'm reading your suggested comment correctly, it implies that by using `-1` we'll start in an earlier epoch than when the backup was taken, which I don't think is the case. On a related note, could we also start from the next epoch after the backup was taken, so without the `-1`?

Another thought I had was whether it makes sense to exit early in case the backup and current `epoch_count` are the same. Even though `epoch_range` would be empty, we might do some extra steps before we realize.

Regarding removing the `start_epoch` variable, I'm currently using it in the `info!` call below, which I could also replace with `backup.epoch_count.saturating_sub(1)`, but I thought it might be better to avoid repeating that.

Sorry, I hope I'm not overcomplicating this, but just wanted to make sure :)

dpc: The backup of the notes from the database is taken at the epoch count N, which means any changes to note state could have only happen in the next epoch (so epoch N), so starting from just N (instead of N - 1) *should* be OK. I wast just not sure if they weren't any weird races or corner cases.

> Another thought I had was whether it makes sense to exit early in case the backup and current epoch_count are the same.

Probability of this is almost zero in practice, so not worth special-casing. Generally decent logic should not need any `if something.is_empty() { /* early exit */ return; }`. Rather common anti pattern.

> Regarding removing the start_epoch variable, I'm currently using it in the info! call below, which I could also replace with backup.epoch_count.saturating_sub(1), but I thought it might be better to avoid repeating that.

I guess I only care about `epoch_count` being named `current_epoch_count`.

dpc:
image

When working on PRs, I never merge, only `git fetch upstream && git rebase -i upstream/master`. Dealing with merge commits is a pain, only useful when the structure of the changes needs to be preserved (e.g. commit signatures, or just the whole history). In a PR that's almost never the case.

client/client-lib/src/mint/backup.rs
jp1ac4: Ah yes I see, thanks for clarifying about the backup epoch. I've now added your comment to the code.

> Probability of this is almost zero in practice, so not worth special-casing.

OK thanks for confirming, and also thanks for the tip about the anti-pattern :)

> I guess I only care about `epoch_count` being named `current_epoch_count`.

Ah sorry, I hadn't spotted this renaming in your previous comment. Have now made this change and left `start_epoch` unchanged.

jp1ac4:
> When working on PRs, I never merge

Ah OK. I was planning to rebase (and squash commits) at the end once all the discussion had been finalized. I can rebase now. Shall I also squash all commits now or only once PR has been approved?

jp1ac4:
Thanks for your review of the draft PR, @dpc, and your helpful comments. I've now squashed and rebased my commits and marked this as Ready for Review.

dpc:
> Shall I also squash all commits now or only once PR has been approved

Personally, I don't care, but some reviewers might like to have each fix in a different commit and a final squash.

Generally, it's best to keep changes small so it doesn't matter. I think this one is small enough to just be one piece.

client/client-lib/src/api.rs
```diff
@@ -457,8 +457,8 @@ where
         .await
     }
 
-    async fn fetch_last_epoch(&self) -> FederationResult<u64> {
-        self.request_eventually_consistent("/epoch".to_owned(), erased_no_param())
+    async fn fetch_epoch_count(&self) -> FederationResult<u64> {
+        self.request_eventually_consistent("/fetch_epoch_count".to_owned(), erased_no_param())
```
elsirion, line 461: nit: we might want some unified naming scheme. `get` seems more common. thoughts @mxz42? Not a blocker though.

client/client-lib/src/mint/backup.rs
```diff
@@ -246,18 +246,17 @@ impl MintClient {
...
     ) -> Result<Cancellable<EcashRecoveryFinalState>> {
-        let end_epoch = match self.context.api.fetch_last_epoch().await {
+        let current_epoch_count = match self.context.api.fetch_epoch_count().await {
             Ok(v) => v,
             Err(e) => {
                 return Err(e.into());
             }
         };
-        let epoch_range = backup.epoch..=end_epoch;
+        // TODO: This -1 is probably not necessary, as it should be enough to start from the exact epoch the snapshot was taken, but it is harmless to start from any epoch in the past, and starting a bit earlier makes it more robust in face of some inconsistency that we've missed.
```
elsirion, line 255: Is this claim true that double-processing of epochs is harmless @dpc?

elsirion (approved):
LGTM, would like an ACK from @dpc though.

client/client-lib/src/api.rs
mxz42: yeah, as dpc [mentioned](https://github.com/fedimint/fedimint/pull/1566#discussion_r1094912734) `getepochcount` would be best to be sort of close to the bitcoin cli

client/client-lib/src/mint/backup.rs
dpc: Yes. Just like you can always start restore from zero, you could start it from 1, or any other number in between. Restore code looks for things it expects and ignores everything else.

elsirion: Would be good to just have a test to see if we need the sub or not (comment @mxz42).

PR #1567, opened 2023-02-01
Dummy Module Database Migration

m1sterc001guy:
This PR adds:

- Persisting the database version of each module and checking if a db migration needs to occur.
- A DB Migration of the Dummy Module integration test.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::error::Error;
 use std::fmt::Debug;
 use std::marker::PhantomData;
@@ -8,10 +9,12 @@ use anyhow::Result;
 use async_trait::async_trait;
 use futures::future::BoxFuture;
 use futures::{stream, Stream, StreamExt};
+use serde::Serialize;
+use strum_macros::EnumIter;
 use thiserror::Error;
-use tracing::{debug, instrument, trace, warn};
+use tracing::{debug, info, instrument, trace, warn};
 
-use crate::core::ModuleInstanceId;
+use crate::core::{ModuleInstanceId, ModuleKind};
 use crate::encoding::{Decodable, Encodable};
 use crate::fmt_utils::AbbreviateHexBytes;
 
@@ -740,7 +743,7 @@ where
 #[macro_export]
 macro_rules! impl_db_prefix_const {
     (key = $key:ty, value = $val:ty, db_prefix = $db_prefix:expr $(, query_prefix = $query_prefix:ty)* $(,)?) => {
-        impl fedimint_core::db::DatabaseKeyPrefixConst for $key {
+        impl $crate::db::DatabaseKeyPrefixConst for $key {
             const DB_PREFIX: u8 = $db_prefix as u8;
             type Key = Self;
             type Value = $val;
@@ -754,6 +757,46 @@ macro_rules! impl_db_prefix_const {
             }
         )*
     };
+}
+
+#[derive(Debug, Encodable, Decodable, Serialize)]
+pub struct DatabaseVersionKey;
+
+#[derive(Debug, Clone, Copy, Encodable, Decodable)]
+pub struct DatabaseVersionKeyPrefix;
+
+#[derive(Debug, Encodable, Decodable, Serialize, Clone, PartialOrd, Ord, PartialEq, Eq)]
+pub struct DatabaseVersion(pub u64);
+
+impl_db_prefix_const!(
+    key = DatabaseVersionKey,
+    value = DatabaseVersion,
+    db_prefix = DbKeyPrefix::DatabaseVersion,
+    query_prefix = DatabaseVersionKeyPrefix
+);
+
+impl std::fmt::Display for DatabaseVersion {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl DatabaseVersion {
+    pub fn increment(&mut self) {
+        self.0 += 1;
+    }
+}
+
+impl std::fmt::Display for DbKeyPrefix {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+#[repr(u8)]
+#[derive(Clone, EnumIter, Debug)]
+pub enum DbKeyPrefix {
+    DatabaseVersion = 0x50,
 }
 
 #[derive(Debug, Error)]
@@ -831,11 +874,77 @@ macro_rules! push_db_key_items {
     };
 }
 
-mod tests {
-    use futures::StreamExt;
+/// MigrationMap is a BTreeMap that maps DatabaseVersions to async functions.
+/// These functions are expected to "migrate" the database from the keyed
+/// DatabaseVersion to DatabaseVersion + 1.
+pub type MigrationMap<'a> = BTreeMap<
+    DatabaseVersion,
+    for<'b> fn(
+        &'b mut DatabaseTransaction<'a>,
+    ) -> Pin<Box<dyn futures::Future<Output = anyhow::Result<()>> + Send + 'b>>,
+>;
 
-    use super::Database;
+/// `apply_migrations` iterates from the on disk database version for the module
+/// up to `target_db_version` and executes all of the migrations that exist in
+/// the `MigrationMap`. Each migration in `MigrationMap` updates the database to
+/// have the correct on-disk structures that the code is expecting. The entire
+/// migration process is atomic (i.e migration from 0->1 and 1->2 happen
+/// atomically). This function is called before the module is initialized and as
+/// long as the correct migrations are supplied in `MigrationMap`, the module
+/// will be able to read and write from the database successfully.
+pub async fn apply_migrations<'a>(
+    db: &'a Database,
+    kind: ModuleKind,
+    target_db_version: DatabaseVersion,
+    migrations: MigrationMap<'a>,
+) -> Result<(), anyhow::Error> {
+    let mut dbtx = db.begin_transaction().await;
+    let disk_version = dbtx.get_value(&DatabaseVersionKey).await?;
+    let db_version = if let Some(disk_version) = disk_version {
+        let mut current_db_version = disk_version;
+
+        if current_db_version > target_db_version {
+            return Err(anyhow::anyhow!(format!(
+                "On disk database version for module {kind} was higher than the code database version."
+            )));
+        }
+
+        while current_db_version < target_db_version {
+            if let Some(migration) = migrations.get(&current_db_version) {
+                migration(&mut dbtx).await?;
+            } else {
+                panic!("Missing migration for version {current_db_version}");
+            }
+
+            current_db_version.increment();
+            dbtx.insert_entry(&DatabaseVersionKey, &current_db_version)
+                .await?;
+        }
+
+        current_db_version
+    } else {
+        dbtx.insert_entry(&DatabaseVersionKey, &target_db_version)
+            .await?;
+        target_db_version
+    };
+
+    dbtx.commit_tx().await?;
+    info!("{} module db version: {}", kind, db_version);
+    Ok(())
+}
+
+#[allow(unused_imports)]
+mod tests {
+    use futures::{FutureExt, StreamExt};
+
+    use super::{
+        apply_migrations, Database, DatabaseTransaction, DatabaseVersion, DatabaseVersionKey,
+        MigrationMap,
+    };
+    use crate::core::ModuleKind;
+    use crate::db::mem_impl::MemDatabase;
     use crate::encoding::{Decodable, Encodable};
+    use crate::module::registry::ModuleDecoderRegistry;
 
     #[repr(u8)]
     #[derive(Clone)]
@@ -856,6 +965,19 @@ mod tests {
         value = TestVal,
         db_prefix = TestDbKeyPrefix::Test,
         query_prefix = DbPrefixTestPrefix
+    );
+
+    #[derive(Debug, Encodable, Decodable)]
+    struct TestKeyV0(u64, u64);
+
+    #[derive(Debug, Encodable, Decodable)]
+    struct DbPrefixTestPrefixV0;
+
+    impl_db_prefix_const!(
+        key = TestKeyV0,
+        value = TestVal,
+        db_prefix = TestDbKeyPrefix::Test,
+        query_prefix = DbPrefixTestPrefixV0
     );
 
     #[derive(Debug, Encodable, Decodable)]
@@ -1485,6 +1607,74 @@ mod tests {
         assert_eq!(test_dbtx.get_value(&TestKey(101)).await.unwrap(), None);
 
         test_dbtx.commit_tx().await.expect("DB Error");
+    }
+
+    #[cfg(test)]
+    #[tokio::test]
+    pub async fn verify_test_migration() {
+        // Insert a bunch of old dummy data that needs to be migrated to a new version
+        let db = Database::new(MemDatabase::new(), ModuleDecoderRegistry::default());
+        let expected_test_keys_size: usize = 100;
+        let mut dbtx = db.begin_transaction().await;
+        for i in 0..expected_test_keys_size {
+            dbtx.insert_new_entry(&TestKeyV0(i as u64, (i + 1) as u64), &TestVal(i as u64))
+                .await
+                .expect("DB Error");
+        }
+
+        dbtx.insert_new_entry(&DatabaseVersionKey, &DatabaseVersion(0))
+            .await
+            .expect("DB Error");
+        dbtx.commit_tx().await.expect("DB Error");
+
+        let mut migrations = MigrationMap::new();
+
+        migrations.insert(DatabaseVersion(0), move |dbtx| {
+            migrate_test_db_version_0(dbtx).boxed()
+        });
+
+        apply_migrations(
+            &db,
+            ModuleKind::from_static_str("TestModule"),
+            DatabaseVersion(1),
+            migrations,
+        )
+        .await
+        .expect("Error applying migrations for TestModule");
+
+        // Verify that the migrations completed successfully
+        let mut dbtx = db.begin_transaction().await;
+
+        // Verify Dummy module migration
+        let test_keys = dbtx
+            .find_by_prefix(&DbPrefixTestPrefix)
+            .await
+            .collect::<Vec<_>>()
+            .await;
+        let test_keys_size = test_keys.len();
+        assert_eq!(test_keys_size, expected_test_keys_size);
+        for test_key in test_keys {
+            let (key, val) = test_key.unwrap();
+            assert_eq!(key.0, val.0 + 1);
+        }
+    }
+
+    #[allow(dead_code)]
+    async fn migrate_test_db_version_0<'a, 'b>(
+        dbtx: &'b mut DatabaseTransaction<'a>,
+    ) -> Result<(), anyhow::Error> {
+        let example_keys_v0 = dbtx
+            .find_by_prefix(&DbPrefixTestPrefixV0)
+            .await
+            .collect::<Vec<_>>()
+            .await;
+        dbtx.remove_by_prefix(&DbPrefixTestPrefixV0).await?;
+        for pair in example_keys_v0 {
+            let (key, val) = pair?;
+            let key_v2 = TestKey(key.1);
+            dbtx.insert_new_entry(&key_v2, &val).await?;
+        }
+        Ok(())
     }
 
     #[cfg(test)]
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -11,7 +11,7 @@ use std::sync::Mutex;
 
 use fedimint_core::config::{ConfigResponse, ModuleGenRegistry};
 use fedimint_core::core::ModuleInstanceId;
-use fedimint_core::db::{Database, DatabaseTransaction};
+use fedimint_core::db::{apply_migrations, Database, DatabaseTransaction};
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::epoch::*;
 use fedimint_core::module::audit::Audit;
@@ -148,10 +148,19 @@ impl FedimintConsensus {
             };
             info!(module_instance_id = *module_id, kind = %kind, "Init module");
 
+            let isolated_db = db.new_isolated(*module_id);
+            apply_migrations(
+                &isolated_db,
+                init.module_kind(),
+                init.database_version(),
+                init.get_database_migrations(),
+            )
+            .await?;
+
             let module = init
                 .init(
                     cfg.get_module_config(*module_id)?,
-                    db.new_isolated(*module_id),
+                    isolated_db,
                     &env,
                     task_group,
                 )
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -1229,10 +1229,21 @@ impl FederationTest {
                     modules.insert(id, module);
                 } else {
                     info!(module_instance_id = id, kind = %kind, "Init module");
+
+                    let isolated_db = db.new_isolated(id);
+                    fedimint_core::db::apply_migrations(
+                        &isolated_db,
+                        gen.module_kind(),
+                        gen.database_version(),
+                        gen.get_database_migrations(),
+                    )
+                    .await
+                    .unwrap();
+
                     let module = gen
                         .init(
                             cfg.get_module_config(id).unwrap(),
-                            db.new_isolated(id),
+                            isolated_db,
                             &env_vars,
                             &mut task_group,
                         )
--- a/modules/fedimint-dummy/src/db.rs
+++ b/modules/fedimint-dummy/src/db.rs
@@ -1,7 +1,30 @@
+use fedimint_core::db::DatabaseTransaction;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::impl_db_prefix_const;
+use futures::StreamExt;
 use serde::Serialize;
 use strum_macros::EnumIter;
+
+/// Example function that will migrate the Dummy Module's database from
+/// version 0 to version 1. This function selects all of the ExampleKeyV0
+/// and inserts a new String to construct ExampleKeys, deletes the old
+/// ExampleKeyV0, then inserts the new ExampleKeys.
+pub async fn migrate_dummy_db_version_0<'a, 'b>(
+    dbtx: &'b mut DatabaseTransaction<'a>,
+) -> Result<(), anyhow::Error> {
+    let example_keys_v0 = dbtx
+        .find_by_prefix(&ExampleKeyPrefixV0)
+        .await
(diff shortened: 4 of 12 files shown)


fedimint-api/src/core/server.rs
```diff
@@ -51,6 +54,10 @@ pub trait IServerModule: Debug {
 
     fn as_any(&self) -> &dyn Any;
 
+    fn database_version(&self) -> DatabaseVersion;
```
dpc, line 57: I wonder if we should expose this, or just give modules some functions to deal with it internally.

fedimint-api/src/core/server.rs
```diff
@@ -51,6 +54,10 @@ pub trait IServerModule: Debug {
...
+
+    async fn migrate_database(&self, db: &Database, db_version: u64) -> Result<(), anyhow::Error>;
```
dpc, line 59: Unclear if `db_version` is "from" or "to" or what exactly.

modules/fedimint-ln/src/lib.rs
```diff
@@ -441,6 +444,114 @@ impl ServerModule for Lightning {
...
+
+    async fn migrate_database(&self, db: &Database, db_version: u64) -> Result<(), anyhow::Error> {
+        let mut current_migration_version = db_version;
+        while current_migration_version < self.database_version().version {
+            match current_migration_version {
+                0 => {
+                    let mut read_dbtx = db.begin_transaction().await;
+                    let mut write_dbtx = db.begin_transaction().await;
+
+                    // Upgrade MigratableKeyV1 and MigratableValueV1 to MigratableKeyV2 and MigratableValueV2
```
dpc, line 459: I would move each migration logic to separate closure/function and store them in some `BTreeMap<DatabaseVersion, Box<FnOnce(from: &DatabaseTransaction, to: &DatabaseTransaction)> -> Box<Future<Item = Result<...>>);` kind of thing, so this can become just a loop, lookup, call.

fedimint-api/src/db/mod.rs
```diff
@@ -684,6 +687,124 @@ where
...
+}
+
+#[derive(Debug, Encodable, Decodable)]
+struct MigratableKeyPrefix;
+
+#[derive(Debug, Encodable, Decodable, Eq, PartialEq, DatabaseVersion)]
+#[struct_version(3, "value_1", "value_2", "value_3")]
+#[struct_version(2, "value_1", "value_2")]
+#[struct_version(1, "value_1")]
+pub struct MigratableValue {
```
dpc, line 757: I kind of just expected a `MigratableValuev1`, `MigratableValuev2`, etc.

This way new code can just copy&paste old structs, add `vX` suffix and never have to touch them.

fedimint-api/src/db/mod.rs
```diff
@@ -684,6 +687,124 @@ where
     }
 }
 
+#[derive(Debug, Encodable, Decodable, Serialize)]
+pub struct DatabaseVersionKey;
+
+#[derive(Debug, Encodable, Decodable, Serialize, Clone)]
+pub struct DatabaseVersion {
```
dpc, line 694: `DatabaseVersion(u64)`? That `version` field is redundant.

modules/fedimint-mint/src/lib.rs
```diff
@@ -441,6 +441,18 @@ impl ServerModule for Mint {
...
     }
 
+    fn database_version(&self) -> DatabaseVersion {
+        DatabaseVersion { version: 1 }
+    }
+
+    async fn migrate_database(
+        &self,
+        _db: &Database,
+        _db_version: u64,
```
dpc, line 451: I'm surprised this parameter is needed at all. Shouldn't each module source code know what its latest database schema version is, and be able to read the current schema from the database?


I kind of think that we should just let modules call `migrate_database` automatically when instantiatied. But maybe there are some reasons not to. But we can add it easily when we need it, I guess.

fedimint-api/src/db/mod.rs
```diff
@@ -684,6 +687,124 @@ where
...
+    pub key_2: u64,
+
+    pub key_3: u64,
+}
+
+#[derive(Debug, Encodable, Decodable)]
+struct MigratableKeyPrefix;
+
+#[derive(Debug, Encodable, Decodable, Eq, PartialEq, DatabaseVersion)]
+#[struct_version(3, "value_1", "value_2", "value_3")]
```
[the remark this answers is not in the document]
m1sterc001guy: Haha probably should have added some documentation :)

Essentially it's just a macro that makes generating "Versioned" structs easier. Can just use this one line and list the subset of the fields that you want in your old version.

The thinking is that when adding a new field (or modifying an existing one), you can add this macro and keep the current version, then just modify the current struct with the change.

Perhaps this is overall and too confusing, but I was trying to avoid re-defining the same structs everywhere just to add one field.

fedimint-api/src/db/mod.rs
m1sterc001guy: See my other comment. If we think its overkill I can remove it and just rely on developers to create new versioned structs.

modules/fedimint-mint/src/lib.rs
m1sterc001guy: The module source code will know the current database version (hardcoded in the code) and it will also know the database version that it read from disk (that's what this is, it needs a better name). Perhaps this should just be read from disk here instead of being passed as a parameter.

The module will need to use the database version from disk to know how to read the structs from disk because they might be in a different format from the current code.

I'm indifferent if we should explicitly call `migrate_database` or just let the modules do it. I suppose by adding it to the ServerModule trait, it would force module developers to think about database migrations.

fedimint-api/src/core/server.rs
m1sterc001guy: I think exposing this forces developers to think about database migrations, but I'm ok with changing it too.

fedimint-api/src/db/mod.rs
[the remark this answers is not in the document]
dpc: Possibly some redundancy is going to be less error-prone then smart logic. :crossed_fingers:

fedimint-api/src/db/mod.rs
m1sterc001guy: Removed the custom derive macro

fedimint-api/src/db/mod.rs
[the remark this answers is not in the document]
m1sterc001guy: Removed custom derive macro

fedimint-api/src/core/server.rs
m1sterc001guy: Removed

modules/fedimint-ln/src/lib.rs
m1sterc001guy: Hmm were you thinking of instantiating this statically? I ran into issues doing that with a BTreeMap and a async function as the value. 

Anyway, I just refactored the logic here into another function, perhaps that's good enough for now. Let me know if you see a better way of doing it.

modules/fedimint-mint/src/lib.rs
m1sterc001guy: Removed migrate_database from the trait and calling it from init instead

modules/fedimint-dummy/src/db.rs
```diff
@@ -1,12 +1,43 @@
...
 use serde::Serialize;
 use strum_macros::EnumIter;
 
+pub const DATABASE_VERSION: DatabaseVersion = DatabaseVersion(1);
+
+pub async fn migrate_dummy_db_version_0(db: &Database) -> Result<DatabaseVersion, anyhow::Error> {
+    let mut read_dbtx = db.begin_transaction().await;
+    let mut write_dbtx = db.begin_transaction().await;
+
+    let example_keys_v1 = read_dbtx
```
dpc, line 13: Now that you've implemented streams here, we could do it via stream to avoid allocating memory for all of them at the same time?

dpc: Worth checking if it works. :)

modules/fedimint-mint/src/lib.rs
```diff
@@ -371,6 +372,33 @@ impl ModuleGen for MintGen {
...
+            tracing::info!("Mint module no database version found.");
+            dbtx.insert_entry(&DatabaseVersionKey, &DATABASE_VERSION)
+                .await?;
+            dbtx.commit_tx().await?;
+            DATABASE_VERSION
+        };
+
+        tracing::info!("Mint module db version: {}", mint_db_version);
+        Ok(())
+    }
```
elsirion, line 399: This code seems very repetitive. I think there should be one function implementing this which is called by all modules in their `init` function. It should take the migration map mentioned by @dpc as argument.


```rust
type MigrationMap = BTreeMap<Version, fn(&mut DatabaseTransaction) -> Box<dyn Future<Output = anyhow::Result<()>>>>

pub fn migrate_db(db: Database, migrations: MigrationMap) -> anyhow::Result<()> {
    …
}
```

This function should run all necessary migrations from that map and update the version key on each step.

There will likely be some lifetime issues around futures capturing a DB tx reference, but see #1423 for how it might be solved.

modules/fedimint-dummy/src/db.rs
elsirion: This is unlikely to work since reading requires a mut ref to the DB tx as well as writing.

m1sterc001guy: Yep I did try to use for_each first like this:

`example_keys_v1.for_each(|(key, _)| async {
        let key_v2 = ExampleKey(key.0, "Example String".to_string());
        write_dbtx.insert_new_entry(&key_v2, &()).await.expect("DB Error");
    });`

I tried making the closure `async` and `asycn move` but I'm hitting issues since `write_dbtx` needs to be mut.

modules/fedimint-mint/src/lib.rs
m1sterc001guy: Sure, I can refactor this. I think the `DatabaseTransaction` in the `BTreeMap` doesn't need to be a reference, since we can do each migration as a separate transaction, which would mean we're updating each version atomically instead of doing the entire migration atomically. That might help with the lifetime issues.

Where do you think this map should be populated? In `init`? I tried initializing it statically but ran into problem storing async function inside the `BTreeMap`.

modules/fedimint-ln/src/lib.rs
m1sterc001guy: Nevermind, got this to work :)

modules/fedimint-mint/src/lib.rs
m1sterc001guy: Nevermind, got this to work :)

integrationtests/tests/tests.rs
```diff
@@ -1203,3 +1207,35 @@ async fn cannot_replay_transactions() -> Result<()> {
...
     .await
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn verify_dummy_migration() -> Result<()> {
+    let fed_db = |decoders| async {
+        // Insert a bunch of old dummy data that needs to be migrated to a new version
+        let db = Database::new(MemDatabase::new(), decoders);
+        let dbtx = db.begin_transaction().await;
+        let mut module_dbtx = dbtx.new_module_tx(LEGACY_HARDCODED_INSTANCE_ID_DUMMY);
```
m1sterc001guy, line 1217: I'm not super crazy about re-adding the LEGACY_HARDCODED_INSTANCE_ID, even if its just for the dummy module. But I couldn't figure out a way to "inject" data into the db before the federation has started without adding specialized logic into the `Fixtures`, which didn't seem like the right solution.

m1sterc001guy: I'm not actually sure if its possible to not use it, since by definition, when the federation hasn't started yet, the module instance ids haven't been assigned yet.

modules/fedimint-mint/src/lib.rs
```diff
@@ -154,10 +154,11 @@ impl ModuleGen for MintGen {
     async fn init(
         &self,
         cfg: ServerModuleConfig,
-        _db: Database,
+        db: Database,
         _env: &BTreeMap<OsString, OsString>,
         _task_group: &mut TaskGroup,
     ) -> anyhow::Result<DynServerModule> {
+        self.migrate_database(db).await?;
```
elsirion, line 161: I'm still not super happy with manually calling and defining `migrate_database`. 

My ideal: 
* let modules define a fn that returns the migration map
* Default-implement `migrate_database` on the trait
* Let `FedimintConsensus` call it (or init? idk tbh)

elsirion (dismissed):
Looks pretty good as a first step. We can iterate in follow-ups as we gain experience with migrations.

modules/fedimint-mint/src/lib.rs
m1sterc001guy: Ha, I actually had it in the trait initially and moved it out. I can move it back.

> Let `FedimintConsensus` call it (or init? idk tbh)

I just realized that it needs to be called in `init`, since the database is passed in as a parameter there. Can't be called after since it is possible that the database is used in `init`, and can't be called before (if we want it on the trait) since the module isn't initialized yet.

fedimint-api/src/db/mod.rs
```diff
@@ -789,6 +833,60 @@ macro_rules! push_db_key_items {
...
+        let mut current_dbversion = ondisk_version;
+
+        if current_dbversion > target_dbversion {
+            return Err(anyhow::anyhow!(format!(
+                "On disk database version for module {kind} was higher than the code database version."
+            )));
+        }
+
+        while current_dbversion < target_dbversion {
+            if let Some(migration) = migration_map.get(&current_dbversion) {
```
dpc, line 869: Shouldn't we require migration for every version bump? After all - if there was a version bump, there had to be a reason for it.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -822,6 +880,31 @@ impl FederationTest {
         user.client.fetch_all_notes().await.unwrap();
     }
 
+    /// Verifies that the database migration for the dummy module completed successfully
+    pub async fn verify_dummy_module_database_migration(&self, expected_example_key_size: usize) {
+        for server in &self.servers {
+            block_on(async {
```
dpc, line 886: AFAIR, these `block_on` calls are a bad workaround for warning about holding `borrow`/`borrow_mut` of `RefCell` across `yield`? They are not making the problem go away, just mask it. It would be more honest to just use `#[allow(...)]` with a comment.

After https://github.com/fedimint/fedimint/pull/1599 the need for them should no longer be there.

BTW. We should never need to use `Rc<RefCell<_>>` that a type that should (almost) never appear in a multi-threaded runtime async code.

integrationtests/tests/tests.rs
dpc: You should not need to add `LEGACY_HARDCODED_INSTANCE_ID`. You can make up any id, as the module under test can't even check it.

fedimint-api/src/db/mod.rs
m1sterc001guy: That's a good point. I can add a panic which should break our tests if that ever happens.

integrationtests/tests/fixtures/mod.rs
m1sterc001guy: Yes this was just done to appease the compiler :) 

Should I just wait until https://github.com/fedimint/fedimint/pull/1599 lands then fix it here?

dpc: Don't wait, but after #1599 would be nice if you could clean it up here and whenever else (should be just `grep` away).

m1sterc001guy: Changed to `#[allow(...)]` for now, can remove in follow up

fedimint-api/src/db/mod.rs
m1sterc001guy: Added `panic!` if there is no migration for a db version

modules/fedimint-mint/src/lib.rs
m1sterc001guy: Moved into `ModuleGen` trait

fedimint-api/src/module/mod.rs
```diff
@@ -386,6 +386,7 @@ pub struct ApiVersion {
 #[async_trait]
 pub trait ModuleGen: Debug + Sized {
     const KIND: ModuleKind;
+    const DATABASE_VERSION: DatabaseVersion;
```
dpc, line 389: The version of a database is a property of the module instance.

fedimint-api/src/module/mod.rs
```diff
@@ -411,6 +412,13 @@ pub trait ModuleGen: Debug + Sized {
         task_group: &mut TaskGroup,
     ) -> anyhow::Result<DynServerModule>;
 
+    /// Construct the MigrationMap that contains the migrations from each DatabaseVersion to
+    /// DatabaseVersion + 1. Then apply the migrations to the database atomically.
+    async fn migrate_database(&self, db: Database) -> Result<(), anyhow::Error> {
+        let migrations = MigrationMap::new();
```
dpc, line 418: You probably want `fn ModuleGen::get_database_migrations(&self) -> MigrationMap` and then use it from `fn ModuleGen::init`?

dpc: Some stuff here is tricky...

integrationtests/tests/tests.rs
m1sterc001guy: The issue is that the integration test needs to know which module instance id is assigned to the dummy module, so that it can backfill some test data before the federation starts that will later be migrated when the federation starts. If we don't use the correct ID, then the dummy module won't even see the test data (since it is logically isolated), the migration won't execute and the test will fail.

fedimint-api/src/module/mod.rs
m1sterc001guy: +1 yes that is better

dpc: We can play with it over time, no need to stress about it.

fedimint-api/src/module/mod.rs
m1sterc001guy: Perhaps this is a misnomer. This represents the database version that the current code works with. So I believe its actually a static property of the code which we will need to bump when there's a change to the db schema of a module.

The database version that is persisted to the database is a property of the module instance though (this is already isolated using the `IsolatedDatabaseTransaction` mechanism.

m1sterc001guy: Added comment in `ModuleGen` to clarify its purpose

integrationtests/tests/tests.rs
dpc: I think kitman just today put module instance ids into `FederationTest`, so we don't have to hardcode them. (grep `mint_id`). Would that help?

fedimint-api/src/module/mod.rs
dpc: Oh, you're right.

dpc:
OK, so things that I realized just recently:

* the "outside" (generic code) should persist the current version of db schema for each module instance
* user provides map of migrations in `ModuleGen` and a static current desired db version
* "the outside" should be responsible for calling all the migration between the current db schema version and desired db version, probably in the code that does the `Module::init` of all the modules

integrationtests/tests/tests.rs
m1sterc001guy: Unfortunately I don't think it does, since the module instance ids are assigned when the federation is created, and for this test I need to inject dummy data into the database before those ids are assigned so that the migration can be triggered.

m1sterc001guy:
Latest version should do most of the above things you mentioned:

> the "outside" (generic code) should persist the current version of db schema for each module instance, so the module never even sees it;
See `apply_migrations` in `fedimint_api::db::mod.rs`. This is outside modules and handles the `DatabaseVersion` for each module.

> user provides map of migrations in ModuleGen and a static current desired db version
Latest code contains `get_migration_map` which is part of the `ModuleGen` trait and allows modules to define which migrations are needed. `DATABASE_VERSION` constant forces modules to define which version the code is compatible with.

> "the outside" should be responsible for calling all the migration between the current db schema version and desired db version, probably in the code that does the Module::init of all the modules
This one is tricky. `apply_migrations` needs to be called before `init`, since `db` is one of the parameters to `init` and it is unsafe to access the `db` before the migrations have occurred. However, we can't call one of the trait functions in `FedimintConsensus::new` until after `init` since that's what creates the instance of the module in the first place. Perhaps we should statically iterate through all of the "registered modules" (i.e LightningGen, WalletGen, etc) and call `apply_migrations` on each one? That way we could do the migrations before calling `init`

m1sterc001guy:
> However, we can't call one of the trait functions in FedimintConsensus::new until after init since that's what creates the instance of the module in the first place.

Ignore this, I misinterpreted the code. This should be fixed now, `apply_migrations` is called before `init` in `FedimintConsensus::new`

fedimint-api/src/db/mod.rs
```diff
@@ -789,6 +833,69 @@ macro_rules! push_db_key_items {
...
+
+        current_dbversion
+    } else {
+        dbtx.insert_entry(&DatabaseVersionKey, &target_dbversion)
+            .await?;
+        target_dbversion
+    };
+
+    dbtx.commit_tx().await?;
+    tracing::info!("{} module db version: {}", migration_data.kind, db_version);
```
dpc, line 895: nit: just qualify to be `info!(...)`

modules/fedimint-dummy/src/lib.rs
```diff
@@ -74,15 +79,25 @@ impl ModuleGen for DummyConfigGenerator {
...
+    fn get_database_migration_data(&self) -> DatabaseMigrationData {
+        let mut migrations = MigrationMap::new();
+
+        migrations.insert(DatabaseVersion(0), move |dbtx| {
+            migrate_dummy_db_version_0(dbtx).boxed()
+        });
+
+        DatabaseMigrationData {
+            kind: Self::KIND,
+            target_db_version: Self::DATABASE_VERSION,
```
dpc, line 91: Both of `kind` and `target_db_version` are statics. I think we should just have a standalone methods to return them and avoid the vague "*Data" aggregate struct.

fedimint-server/src/consensus/mod.rs
```diff
@@ -143,10 +143,14 @@ impl FedimintConsensus {
             };
             info!(module_instance_id = *module_id, kind = %kind, "Init module");
 
+            let isolated_db = db.new_isolated(*module_id);
+            fedimint_api::db::apply_migrations(&isolated_db, init.get_database_migration_data())
```
dpc, line 147: We could probably move `fedimint_api::db:apply_migrations` to `fedimint-server` now.

fedimint-api/src/module/mod.rs
```diff
@@ -278,6 +280,10 @@ pub trait IModuleGen: Debug {
         task_group: &mut TaskGroup,
     ) -> anyhow::Result<DynServerModule>;
 
+    /// Retreives the `DatabaseMigrationData` from the module to be applied to the database
+    /// before the module is initialized.
+    fn get_database_migration_data(&self) -> DatabaseMigrationData;
```
dpc, line 285: "data" is a word to avoid. Everything in SWE is data, so it doesn't bring any information.

fedimint-api/src/db/mod.rs
```diff
@@ -789,6 +833,69 @@ macro_rules! push_db_key_items {
...
+            return Err(anyhow::anyhow!(format!(
+                "On disk database version for module {0} was higher than the code database version.", migration_data.kind
+            )));
+        }
+
+        while current_dbversion < target_dbversion {
+            if let Some(migration) = migration_data.migrations.get(&current_dbversion) {
+                migration(&mut dbtx).await?;
+            } else {
+                panic!("DatabaseVersion {current_dbversion} does not contain a migration.");
```
dpc, line 879: `"Missing migration for version {}"`?

BTW. I wonder if we should index on the `to_version` or `from_version`. I guess `from` (that you currently use) have a nice property that it starts from `0`...

integrationtests/tests/tests.rs
dpc: I think we posted at the same time:

> I think kitman just today put module instance ids into FederationTest, so we don't have to hardcode them. (grep mint_id). Would that help?

integrationtests/tests/fixtures/mod.rs
justinmoon: #1599 is merged

m1sterc001guy: Removed `#[allow(...)]`

modules/fedimint-dummy/src/lib.rs
m1sterc001guy: Removed `DatabaseMigrationData` struct

fedimint-server/src/consensus/mod.rs
m1sterc001guy: Moved to `fedimint-server`

fedimint-api/src/module/mod.rs
m1sterc001guy: Removed :)

fedimint-api/src/db/mod.rs
m1sterc001guy: This was my thinking as well. I originally had it the other way but ended up confusing myself since things appeared to be off by 1.

integrationtests/tests/tests.rs
m1sterc001guy: Removed the integration test and created an equivalent unit test instead

fedimint-server/src/consensus/mod.rs
m1sterc001guy: I moved `apply_migrations` back to `fedimint_api`. I moved it back so that the unit test could be in `fedimint_api`. It also occurred to me that the client will want to have the ability to call `apply_migrations` too, so I think keeping it in `fedimint_api` is ok.

dpc (dismissed):

dpc (approved):

elsirion:
Since @dpc worked with you on this most I'd like to see his ACK first, then I can do a final review, then maybe squash some of the fixups if applicable and then re-ACK+merge.

dpc (approved):

fedimint-core/src/db/mod.rs
```diff
@@ -831,11 +874,77 @@ macro_rules! push_db_key_items {
...
 }
 
+/// MigrationMap is a BTreeMap that maps DatabaseVersions to async functions.
+/// These functions are expected to "migrate" the database from the keyed
+/// DatabaseVersion to DatabaseVersion + 1.
+pub type MigrationMap<'a> = BTreeMap<
+    DatabaseVersion,
+    for<'b> fn(
+        &'b mut DatabaseTransaction<'a>,
+    ) -> Pin<Box<dyn futures::Future<Output = anyhow::Result<()>> + Send + 'b>>,
```
mxz42, line 884: nit: I think we can use a `BoxFuture<'b, T>` here  [docuementation link](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html)
makes it shorter :) I wanted to go through the code base and replace all these patterns anyway at some point.

mxz42 (approved):

fedimint-core/src/db/mod.rs
dpc: `semgrep` would probably be able to detect it. :)

mxz42: :+1:  added to my todo

Issue #1528, opened 2023-01-25
Remove blocking IO inside async code


justinmoon:
AFAIK we currently block during async writes do at least one of our database implementations. IIRC, one blocking IO operation could stall the whole async runtime?

@dpc can explain more.

dpc:
Doing blocking IO in async code can lead to the whole tokio executor thread getting stack for some time, not letting (some) other futures/tasks making progress.

I think @m1sterc001guy is aware of this. We might even have an issue for it already.

m1sterc001guy:
Yep, started on working on this today. This issue was also mentioned [here](https://github.com/fedimint/fedimint/issues/1322) but it's good to have its own issue to track.

m1sterc001guy:
Rocksdb just merged support for enabling async IO in their C API: https://github.com/facebook/rocksdb/commit/ff04fb154bd74fd0681baa83b478095207e2719d

The rust-rocksdb library is linking against a version of rocksdb that does not have this API available yet though. For a proof of concept, I forked the repo, updated rocksdb to master, built from source, and exposed the API in rust. https://github.com/m1sterc001guy/rust-rocksdb/tree/async_io

This is a (hacky) example of how to consume it in fedimint. https://github.com/m1sterc001guy/fedimint/tree/async_rocksdb

I think we should wait for a new rocksdb release, try to upstream the rust changes, and use this so that our reads don't block the CPU thread.

> AFAIK we currently block during async writes do at least one of our database implementations. IIRC, one blocking IO operation could stall the whole async runtime?

Small nit, but writes don't block in Rocksdb. They always write to memory then do compaction in the background. If it can't write to memory because the buffer is full, the write will fail. If this happens, ideally we should backoff for a little bit and re-try (more discussion here: https://github.com/fedimint/fedimint/discussions/1370)

m1sterc001guy:
Best explanation of what rocksdb async IO actually does: http://rocksdb.org/blog/2022/10/07/asynchronous-io-in-rocksdb.html

dpc:
@m1sterc001guy I might be reading it wrong but their async and Rust's async might be two completely different things.

m1sterc001guy:
Hmm ok agreed on that. The terminology is definitely misleading :( I think we might still benefit from this but it looks like we'll need to implement our own mechanism for unblocking the thread that does the read.

dpc:
>  we'll need to implement our own mechanism for unblocking the thread that does the read.


[...]

PR #1568, opened 2023-02-01
Use block_in_place for synchronous Rocksdb operations

m1sterc001guy:
This PR wraps all Rocksdb I/O operations in fedimint_api::task::block_in_place so that the executor thread is not blocked on I/O. Rocksdb does not do I/O during writes (it always writes to memory or it fails), but insert/remove operations return the previous value, so they have a chance of doing I/O while reading for the previous value.

Issue: https://github.com/fedimint/fedimint/issues/1528

### DIFF
--- a/client/client-lib/src/mint/mod.rs
+++ b/client/client-lib/src/mint/mod.rs
@@ -897,7 +897,7 @@ mod tests {
     }
 
     #[allow(clippy::needless_collect)]
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread")]
     async fn test_parallel_issuance() {
         const ITERATIONS: usize = 10_000;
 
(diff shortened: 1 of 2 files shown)


m1sterc001guy:
@dpc what are your thoughts on how to validate that this unblocks the executor? I'm not sure how much overhead block_in_place adds and if the channel implementation that you mentioned [here](https://github.com/fedimint/fedimint/issues/1528) is superior.

dpc:
> I'm not sure how much overhead block_in_place adds and if the channel implementation that you mentioned https://github.com/fedimint/fedimint/issues/1528 is superior.

Under the hood `block_in_place` uses a threadpool and sends it stuff over channel. So it is probably the same. If everything works and there is no lifetime issues, go ahead.

dpc (approved):

Issue #1469, opened 2023-01-21
Server side logging improvements


dpc:
The request logging should have higher logging level

On the server side being able to see the requests that are being made is one of the most useful debugging tools.

Our default logging level is `info`, so we possibly should make it `info`. If anyone will not want to log them, they can override.

https://github.com/fedimint/fedimint/pull/1357/files#r1083315034

Also, we should use the `target` feature more often, and define `const LOG_TARGET_X : &str = ` and document (`./docs/...md`) a list of `targets` that one can enable and disable. AFAIU, the default "target" is the module name which is not very useful in real settings.

We should phase out usage of non-target-taged logging statements.

dpc:
@Maan2003 @justinmoon @elsirion  Thoughts?

elsirion:
> is the module name which is not very useful in real settings.

Why? So far I found it quite nice to work with, typically what I'm interested in is in one or two modules.

> Our default logging level is info, so we possibly should make it info. If anyone will not want to log them, they can override.

To me that feels too noisy, but useful. Maybe just make it debug and activate by default in our deployments?

I think eventually we might just want to log everything anyway and just filter after the fact (especially if we can send the logs to some log server with a nice UI). Maybe the `trace` and `debug` ones should get pruned more quickly in that case.

justinmoon:
I defer to you guys on this. We need documentation about debugging tips and tricks. [I just learned of this `target` trick.](https://github.com/fedimint/fedimint/pull/1357/files#r1083334600)

dpc:
> Why? So far I found it quite nice to work with, typically what I'm interested in is in one or two modules.

This works for developers like us, but not for external users (mint admins) that might not be able to navigate the source code.

While module is somewhat a category, there will be plenty of categories that are cross-module.

Also by committing to stable targets in constants, we make them more like an API that we consciously evolve. Renamings and refactoring will not break filters set by users (as much).

dpc:
Ideally I'd like to be able to do `env RUST_LOG=db=trace,request=trace,debug` etc. to eg. correlate database queries with requests, and get very detailed information about each (arguments, results).

dpc:
IIRC, everything can be hierarchical, so we can have

* consensus
* peers
  * send
  * recv
* db
* api

and so on.

dpc:
#1571 is implementing the idea. Please provide feedback.

dpc:
Closing. Followup moved to #2257

PR #1571, opened 2023-02-03
Logging improvements starter pack

dpc:
This is to start the ball rolling on #1469 .

New constants are defined, committing us to certain logging categories, which we will treat as a semi-public API. This way users can set their logging levels, and they will not change out of the blue, just because we shuffled some code around.

Then, as a showcase,  I change the `error!` and `warn!` use to utilize this. This does add a bit of noise, but I think it's reasonable. We could introduce things like `error_net_peer!(...)`, `warn_consensus!(...)` wrappers etc. if we decide that we don't like the noise.

Lastly, we introduce `semgrep` rules and add them to git pre-commit hook, to enforce correct usage.

I limited the scope of changes to one package and two logging levels, to keep the change small and get some feedback first, before attempting to go all in.

While at it, I changed some logging levels that I've found wrong. Let me know if you disagree.

### DIFF
--- a/.semgrep.yaml
+++ b/.semgrep.yaml
@@ -1,0 +1,12 @@
+rules:
+- id: use-log-targets
+  languages:
+    - rust
+  message: Log statements should have static target defined to aid filtering
+  patterns:
+    # semgrep struggles with finding boundries of rust macro invocation when elipsis (`...`) is used
+    # so matching this one with raw regexes instead
+    - pattern-regex: (error|warn)!\([^;]*?\);
+    - pattern-not-regex: '(error|warn)!\(\s*target: [^;]*?\);'
+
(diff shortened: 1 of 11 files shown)


elsirion (dismissed):
I like it, but it will be a bit of a battle to change it everywhere.

elsirion:
This closes #1574 too I assume?

elsirion (approved):

dpc:
> This closes #1574 too I assume?

I think I'll follow through and convert couple more crates & logging levels first, the close it maybe?

elsirion:
> I think I'll follow through and convert couple more crates & logging levels first, the close it maybe?

No, the one I mentioned is about parallel pre-commit runs :laughing:

PR #1573, opened 2023-02-03
feat: enforce conventional commit format using `convco`

dpc:
Currently only via git commit check hook, but will add a proper CI test after we try it out, and decide that we want to keep it.

* https://www.conventionalcommits.org
* https://github.com/convco/convco/

Needed to pull in `nixpkgs-unstable` to get a version of the linter that supports stdin. It doesn't seem to increase the amount of things being downloaded much, so should be OK, and we can remove it in the future.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -139,6 +139,22 @@
         "type": "github"
       }
     },
+    "nixpkgs-unstable": {
+      "locked": {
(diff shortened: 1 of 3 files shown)


elsirion (approved):

elsirion:
Let's try this out!

PR #1575, opened 2023-02-03
Clean up associated types in ServerModule

dpc:


fedimint-api/src/module/mod.rs
```diff
@@ -425,8 +425,7 @@ where
 
 #[async_trait]
 pub trait ServerModule: Debug + Sized {
-    const KIND: ModuleKind;
-
+    type Gen: ModuleGen;
```
elsirion, line 425: Now I'm unsure which way around this reference makes the most sense … but I think this should work (just changes if you attach modules or module gens to the federation builder eventually).

fedimint-api/src/module/mod.rs
```diff
@@ -427,10 +424,6 @@ where
 pub trait ServerModule: Debug + Sized {
     type Gen: ModuleGen;
     type Decoder: Decoder;
-    type Input: Input;
-    type Output: Output;
-    type OutputOutcome: OutputOutcome;
-    type ConsensusItem: ModuleConsensusItem;
```
elsirion, line 434: :+1: discussed that with @mxz42 a while ago

elsirion (approved):

mxz42 (approved):
nice!

fedimint-api/src/module/mod.rs
dpc: Might be both.

Issue #1576, opened 2023-02-03
Review and avoid use of locks (`Mutex`, `RwLock`)


elsirion:
If not handled with a lot of care ([always locking in same order, …](https://softwareengineering.stackexchange.com/a/418597)) `Mutex`es can easily lead to seemingly random deadlocks that are hard to debug. Especially with the code base evolving rapidly this is a major challenge since one might not always be aware of locks that are being acquired in other parts of the code when introducing changes.

We should remove all locks not strictly necessary for core features and formally reason about the ones that remain, explaining in comments why deadlocks cannot occur.

EDIT: more on deadlocks https://www.scs.stanford.edu/12au-cs140/notes/deadlock.pdf

justinmoon:
Idea: lock mutex / guard in logger so when a deadlock happens at least we know what caused it

dpc:
Rots in the backlog and if we have bugs of this kind we'll just have to fix them. I also don't recall any suspicious hangs for a long, long time.

PR #1577, opened 2023-02-03
fix: avoid lock around CLN RPC client

elsirion:
Related to #1576

### DIFF
--- a/gateway/ln-gateway/src/cln.rs
+++ b/gateway/ln-gateway/src/cln.rs
@@ -6,13 +6,14 @@ use async_trait::async_trait;
 use cln_plugin::{anyhow, options, Builder, Error, Plugin};
 use cln_rpc::model::{ListchannelsRequest, ListpeersPeersChannelsState, ListpeersRequest};
 use cln_rpc::primitives::ShortChannelId;
-use cln_rpc::{model, ClnRpc, Request, Response};
+use cln_rpc::{model, Request, Response};
 use fedimint_api::Amount;
 use fedimint_server::modules::ln::contracts::Preimage;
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/cln.rs
```diff
@@ -58,18 +59,40 @@ pub struct HtlcAccepted {
...
+    }
+
+    /// Creates a new RPC client for a request.
+    ///
+    /// This operation is cheap enough to do it for each request since it merely connects to a UNIX
+    /// domain socket without doing any further initialization.
+    async fn rpc_client(&self) -> Result<cln_rpc::ClnRpc, LightningError> {
+        cln_rpc::ClnRpc::new(&self.socket)
+            .await
+            .map_err(|_e| LightningError(None))
```
dpc, line 85: If we are going to swallow that error, I would rather log it. Would be even better to return it somehow. Can't we return `anyhow::Result` here anyway?

dpc (commented):
One nit, LGTM (and a big improvement) otherwise.

gateway/ln-gateway/src/cln.rs
elsirion: Good catch! I missed that one when tidying up the original change. I went for the logging approach since changing the error type would require far-reaching API changes in parts that are likely to be changed in the GW refactoring effort anyway. 

@okjodom: What's our error strategy for the new gateway? Can we have more expressive errors while still supporting LN error codes?

okjodom: Well define strict Ln error codes over grpc spec.

Logging within gatewayd is more open, should conform to logging pattern we define everywhere else

okjodom (approved):
lgtm!

dpc (approved):

PR #1578, opened 2023-02-03
Fix MacOS build failure on master

elsirion:


dpc (approved):

dpc:
It's not macos. It's latest Rust being used in non-nix build. I'll update `fenix` in the flake.lock.

PR #1581, opened 2023-02-03
Log code version on startup

### LINKED ISSUE #1408: On start fedimintd should log its version (as `info!`)
Helps a lot debugging.

asifshiraz:
Fixes #1408

### DIFF
--- a/fedimintd/src/bin/main.rs
+++ b/fedimintd/src/bin/main.rs
@@ -53,6 +53,8 @@ async fn main() {
         }
     }
 
+    info!("Starting fedimintd (version: {CODE_VERSION})");
+
     let opts: ServerOpts = ServerOpts::parse();
     let fmt_layer = tracing_subscriber::fmt::layer();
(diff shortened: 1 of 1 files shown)


fedimintd/src/bin/main.rs
```diff
@@ -53,6 +53,8 @@ async fn main() {
         }
     }
 
+    info!("Starting fedimintd (version: {})", CODE_VERSION);
```
dpc, line 56: Clippy will complain new new version of Rust that you can put `CODE_VERSION` into `{}` directly, so you might as well fix it already. :)

elsirion: @dpc Can we add this new clippy version to CI run on PRs? It seems to run for MacOS, so it would fail on master after merge.

asifshiraz: Done

dpc (approved):

justinmoon:
@asifshiraz Could you have `ln_gateway` also print out it code version?

PR #1582, opened 2023-02-04
 feat: Fetch epochs in parallel during ecash restoration (stream version)

dpc:
Replaces #1556 

It seems that in practice fetching epochs take a long time leading to a bad user-experience. Fetching epochs from federation has a large latency, so it makes sense to do in on multiple-threads.

I selected 8 as a good compromise that should not create issues, like too much load, while hopefully giving most of the benefit of parallelism.


elsirion (approved):
This looks so much cleaner now :heart_eyes:

dpc:
I does. Next time I'll know how to do it well. :)

Issue #1110, opened 2022-12-13
Parallelize everything


justinmoon:
(discussed in dev call)

We should parallelize everything we can.

@elsirion: can you fill in the description here.

elsirion:
@dpc and I also noticed that we need to wrap these places into `spawn_blocking` blocks so we don't block `tokio` while `par_iter` is running.

justinmoon:
Update: new module interface for AlephBFT may regress here

elsirion:
While the initial switch will reduce parallelism, this shouldn't be a problem in practice for small base2 denomination federations as they are currently standard. Once the migration is complete we will take a look again at how to parallelize heavy crypto operations most elegantly. @joschisan

joschisan:
The collection and combination of ecash signature shares will most likely be pushed out to the clients. There is no need to order the shares. Should cut compute and space required by the mint module down to a fraction of what it is right now.

justinmoon:
What's the status here @dpc @elsirion @joschisan? Still useful issue?

joschisan:
So this issue should likely be updated as follows: combining shares should be pushed to the clients; Signing does not need to be paralleled for now as it takes only 1/7 of the verification time. finally verification, if done properly right now, would have parallelise over three axis to be performant for all valid transactions: 
* parallelise over ordered transactions, if we have many transactions with few notes
* parallelise over a transactions inputs, if we have transactions with many inputs but few notes per input
* parallelise within a single ecash input, if we have transactions with many notes and only a few input.

Only the latter is done right now. However, the current plan is to only allow a single ecash note per input to simplify the mint module and also remove one axis of parallelisation. Therefore we would only have to parallelise over the first and second axis.

justinmoon:
dev call: @elsirion said this can be done backwards-compatibly so id could come in 2.0.x

elsirion:
Not urgently needed after partial impl and many optimizations by @joschisan.

PR #1585, opened 2023-02-04
refactor: Make `FedimintServer` `Send` and `Sync`

dpc:
I'm working on something else, and this is blocking me.

Dealing with `Unpin` and `!Send` things in async seems like something we should avoid anyway.

### DIFF
--- a/fedimint-api/src/net/peers.rs
+++ b/fedimint-api/src/net/peers.rs
@@ -1,4 +1,4 @@
-use std::ops::{Deref, DerefMut};
+use std::ops::Deref;
 use std::sync::Arc;
 
 use async_trait::async_trait;
@@ -12,19 +12,13 @@ use crate::cancellable::Cancellable;
 pub mod fake;
 
 /// Owned [`PeerConnections`] trait object type
-pub struct PeerConnections<Msg>(Box<dyn IPeerConnections<Msg> + Send + Unpin + 'static>);
+pub struct PeerConnections<Msg>(Arc<dyn IPeerConnections<Msg> + Send + Sync + 'static>);
 
 impl<Msg> Deref for PeerConnections<Msg> {
-    type Target = dyn IPeerConnections<Msg> + Send + Unpin + 'static;
+    type Target = dyn IPeerConnections<Msg> + Send + Sync + 'static;
 
     fn deref(&self) -> &Self::Target {
         &*self.0
-    }
-}
-
-impl<Msg> DerefMut for PeerConnections<Msg> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut *self.0
     }
 }
 
@@ -42,26 +36,26 @@ impl<Msg> DerefMut for PeerConnections<Msg> {
 #[async_trait]
 pub trait IPeerConnections<Msg>
 where
-    Msg: Serialize + DeserializeOwned + Unpin + Send,
+    Msg: Serialize + DeserializeOwned + Sync + Send,
 {
     /// Send a message to a specific peer.
     ///
     /// The message is sent immediately and cached if the peer is reachable and only cached
     /// otherwise.
-    async fn send(&mut self, peers: &[PeerId], msg: Msg) -> Cancellable<()>;
+    async fn send(&self, peers: &[PeerId], msg: Msg) -> Cancellable<()>;
 
     /// Await receipt of a message from any connected peer.
-    async fn receive(&mut self) -> Cancellable<(PeerId, Msg)>;
+    async fn receive(&self) -> Cancellable<(PeerId, Msg)>;
 
     /// Removes a peer connection in case of misbehavior
-    async fn ban_peer(&mut self, peer: PeerId);
+    async fn ban_peer(&self, peer: PeerId);
 
     /// Converts the struct to a `PeerConnection` trait object
     fn into_dyn(self) -> PeerConnections<Msg>
     where
-        Self: Sized + Send + Unpin + 'static,
+        Self: Sized + Send + Sync + 'static,
     {
-        PeerConnections(Box::new(self))
+        PeerConnections(Arc::new(self))
     }
 }
 
@@ -87,8 +81,8 @@ impl<MuxKey, Msg> Deref for MuxPeerConnections<MuxKey, Msg> {
 /// channel at the same time.
 pub trait IMuxPeerConnections<MuxKey, Msg>
 where
-    Msg: Serialize + DeserializeOwned + Unpin + Send,
-    MuxKey: Serialize + DeserializeOwned + Unpin + Send,
+    Msg: Serialize + DeserializeOwned + Sync + Send,
+    MuxKey: Serialize + DeserializeOwned + Sync + Send,
 {
     /// Send a message to a specific destination at specific peer.
     async fn send(&self, peers: &[PeerId], mux_key: MuxKey, msg: Msg) -> Cancellable<()>;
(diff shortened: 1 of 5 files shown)


elsirion (changes requested):
Imo making the connection manager sync and sharable is the wrong way. 

There should be exactly one place where sending and receiving happens, I don't see why it should be `Sync` or in an `Arc`. Having a bunch of `Mutex`es in a very critical part of the system to make it act as if it was more flexible than it can be seems like asking for trouble (#1576). 

Maybe you can describe the problem you are trying to solve in an issue so we can try to find another solution? Maybe it's related to  #172?

dpc:
> Maybe you can describe the problem you are trying to solve in an issue so we can try to find another solution? Maybe it's related to  #172?

Rust's async pretty much forces into Sync, or at least Send, even when it's not needed. In tests we're using Rc<RefCell<>> over FedimintServer  and block_on hacks because of it. 

I agree that sending from one thread is best and safest. We can still do it, and keep this struct Send and Sync for async and testing sake. The lock is taken there in very few places and it's easy to prove that that it won't lead to a deadlocks.

dpc:
These block_on hacks btw totally don't work and just mask legitimate errors of holding RefCell borrowed over await points. Sending them into block_on doesn't change anything. At least that's my understanding.

With Send FedimintServer I want to refactor tests to speed then up significantly by running per federation ops in separate tasks, and not just joins, because joins on cpu intensive operations don't give any parallelism . For this I need Sync, or I'll have to invent even more weird juggling and hacks.

dpc:
These block_on hacks btw totally don't work and just mask legitimate errors of holding RefCell borrowed over await points. Sending them into block_on doesn't change anything. At least that's my understanding.

With Send FedimintServer I want to refactor tests to speed then up significantly by running per federation ops in separate tasks, and not just joins, because joins on cpu intensive operations don't give any parallelism . For this I need Sync, or I'll have to invent even more weird juggling and hacks.

I'll give you Nx speedup in test times, but I need this struct to be at least Send. There slow crypto is not a problem. The problem is that all runs on one thread.

dpc:
These block_on hacks btw totally don't work and just mask legitimate errors of holding RefCell borrowed over await points. Sending them into block_on doesn't change anything. At least that's my understanding.

With Send FedimintServer I want to refactor tests to speed then up significantly by running per federation ops in separate tasks, and not just joins, because joins on cpu intensive operations don't give any parallelism . For this I need Sync, or I'll have to invent even more weird juggling and hacks.

I'll give you Nx speedup in test times, but I need this struct to be at least Send. There slow crypto is not a problem. The problem is that all runs on one thread.

Anyway, I'll have to test PR speedup in top of it, we can be bargain then.

elsirion:
I agree send-ness should be a goal. Let's see how we can make it work without making the network stack sync.

dpc:
@elsirion 

> I agree send-ness should be a goal. Let's see how we can make it work without making the network stack sync.

Actually, now that I think about it. I need it `Sync`. In tests we keep list of `FederationServers` as a collection. If we want to run operations on them actually in parallel (even CPU bound, like crypot), we need to spawn some tasks. 

The nicest way to do that is just to have them `Sync` in `Arc`, so the `Arc<FederationServer>` is `Send`. Then we can `clone`, send to a task, `join_all` on all the tasks, and job done. At no point anything synchronous is done on these types.


In production code, you don't even need to put `FederationServer` in a `Arc`, and keep using it from a single thread-only, in a serialized way. The fact that is `Sync`, can be a little secret.


When you look at what was used to achieve `Sync`, it's only:

```
    outgoing: Mutex<Option<Sender<T>>>,
    incoming: Mutex<Option<Receiver<T>>>,
```

So the lock is taken when doing `outgoing.lock().await.send()` and `incoming.lock().await.recv()` (ignoring `Option` for now).

AFAICT this does not change the blocking behavior in any meaningful way, and can't introduce any deadlocks, even if it was to be used from multiple threads (which again, is not, and we don't need to do it).

One exception here is a `ban` operation, which (if were to be ran from another thread), could block, if any other message was received.

The `Mutex` on `outgoing` is even unnecessary for sending, as this type is `Sync` anyway. I only added it there to allow setting the whole thing to `None` and disconnect the peer tread.

If you are really worried about someone accidentally changing architecture and starting new threads because `FedimintServer` is now `Sync`, we could possibly wrap it in some `struct IForbidSync` that has some `!Sync` field, and `Deref`s to `FedimintServer`.

I guess it might be possible to still spawn tasks per `FedimintServer` without that `Sync`, e.g. by disassembling the container holding them, taking ownership, sending to a task, and then return ownership back to the main thread doing the `join_all`, but that seems terrible. Or maybe putting `FedimintServer`s in their own threads to begin with, and then sending them commands to do? Again - seems like a lot of complexity, where an `Arc` would work just fine.

> I'll give you Nx speedup in test times

I actually have that change and will clean it up and push, and it does parallelize per-server epoch processing, but given that it's only one of many things going on, it is not as significant as I thought. Only some tests get 20% faster or so.

However even without the speedup removing `Rc<RefCell<_>>` and `block_on` calls is quite nice with it.

dpc:
I was also considering if putting the heaviest crypto operations into `spawn_blocking` wouldn't be a good idea generally. https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code

elsirion:
> Actually, now that I think about it. I need it `Sync`. In tests we keep list of `FederationServers` as a collection. If we want to run operations on them actually in parallel (even CPU bound, like crypot), we need to spawn some tasks.

That's exactly why I wanted to get to the bottom of the problem. [I don't think you even need `Sync`ness for this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2ebfe4147eae351badb9e6e11d6780d5) and you don't need to spawn tasks thanks to async rust's primitives.

```rust
let mut container = vec![1, 2, 3, 4];

{
    let mut mutating_futures = container.iter_mut()
        .map(|mut_ref| async move {
            sleep(Duration::from_millis(10 - *mut_ref)).await;
            println!("Doubling number {mut_ref}");
            *mut_ref *= 2;
        })
        .collect::<futures::stream::FuturesUnordered<_>>();

    let before = std::time::Instant::now();
    while let Some(()) = mutating_futures.next().await {}
    println!("Operation took {}ms (would take 30ms without parallelism)", before.elapsed().as_millis());
}

println!("{container:?}");
```

Output:
```
Doubling number 4
Doubling number 3
Doubling number 2
Doubling number 1
Operation took 9ms (would take 30ms without parallelism)
[2, 4, 6, 8]
```

And even if it was useful for tests, if there is no clear general architectural benefit to a change (that writing tests sometimes leads to), then the tests should adapt instead of the production code. Making the main code more fragile, less misuse-resistant or less readable (with the mentioned workarounds) just to make the test suite a bit faster/easier to write would be to cut off your nose to spite your face imo.

dpc:
We discussed on Discord, and I'll pursue a route of making it just `Send`, and then using `Arc<Mutex<Send>>` instead, which will allow me to get rid of `Rc<RefCell<FedmingServerver>>` and using `spawn` just fine.

We probably want to use `spawn_blocking` for CPU-heavy workloads as well.

> then the tests should adapt instead of the production code.

A bit of an off-topic point , but can't agree. :P . You can't test a non-testable code well with no amount of work on the test side. 

The code under test needs to go hand in hand with the needs of the test, otherwise it's a dissaster. That was/is a whole premise of TDD - to drive a good software architecture early on, by enforcing test-ability constrains very early on (upfront).

elsirion:
> We probably want to use `spawn_blocking` for CPU-heavy workloads as well.

Related to #1110.

Closing since the new approach will be fundamentally different and this discussion will be mostly unrelated, so better to make it a new PR.

elsirion:
#1599 is (part of) the outcome of this discussion.

Issue #1553, opened 2023-01-28
Add a `restart` command to tmuxinator environment


justinmoon:
Currently if you are running the tmuxinator dev environment, edit some code, and want to re-run it, you either need to kill and restart tmuxinator (slow) or manually to stop, rebuild, and restart what you changed (annoying). We should instead have one command that's easy to remember and run that just handles this for the developer.

- [x] Kill `fedimintd`, rebuild it, and re-run it **in the same pane** it currently runs in. See [this message](https://discord.com/channels/990354215060795454/1060246795004944466/1064727278568878160). I'm not sure exactly how this would work. This shell process would need to subscribe to some kind of signal to know `restart` has been requested. (#1589)
- [x] Stop the CLN plugin, rebuild it, and re-run it. See [this script](https://github.com/fedimint/fedimint/blob/master/scripts/gw-reload.sh).

The user should be able to run this command from any tmux pane.

cdmoss:
Regarding the second item, I included gw-reload.sh at the bottom of the new restart-tmux.sh script, since that's script @justinmoon refers to. Is there something else that should be done?

elsirion:
I don't think so, good point. Can re-open if there is anything left.

PR #1589, opened 2023-02-05
feat: add restart command for tmuxinator

cdmoss:
A first draft to address #1553. Adds a tmuxinator alias `restart` that runs the new script "restart-tmux", which sends the commands referenced in this [message](https://discord.com/channels/990354215060795454/1060246795004944466/1064727278568878160) and runs gw-reload.sh. Not sure if this is the best approach so I definitely welcome suggested alternatives.

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -10,6 +10,7 @@ pre_window:
   - alias gateway-cli="\$FM_GATEWAY_CLI"
   - alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"
   - alias dbtool="\$FM_DB_TOOL"
+  - alias restart="./scripts/restart-tmux.sh"
 tmux_detached: false
 windows:
   - main:
--- a/scripts/restart-tmux.sh
+++ b/scripts/restart-tmux.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+ 
+tmux send-keys -t fedimint-dev:1.4 "pkill -9 fedimintd" C-m
+tmux send-keys -t fedimint-dev:1.4 "cargo build --bin fedimintd" C-m
+tmux send-keys -t fedimint-dev:1.4 "./scripts/start-fed.sh" C-m
+
+./scripts/gw-reload.sh


cdmoss:
I just noticed it accidentally deleted one of the old aliases in the tmuxinator.yml file, I'll fix that right away.

cdmoss:
I just noticed I deleted one of the original aliases in .tmuxinator.yml, I'll fix that right away.

scripts/restart-tmux.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+ 
+tmux send-keys -t fedimint-dev:1.4 "cargo build --bin fedimintd" C-m
+tmux send-keys -t fedimint-dev:1.4 "pkill -9 fedimintd" C-m
```
elsirion, line 4: Is that the right way around? This mixes federation and build logs, so if the build fails new epoch logs might hide that fact.

elsirion (changes requested):

elsirion:
Needs rebase.

scripts/restart-tmux.sh
cdmoss: Point taken on the ordering, I switched it around. I'm not sure if the convenience tradeoff from using -9 is worth it, so I'll defer to your judgement on that - I can certainly change it to something more graceful.

elsirion: :+1: Let's leave it till graceful shutdown is faster (fix some stuck tasks)

elsirion:
Can you rebase on the current master please to remove the merge conflict?

.tmuxinator.yml
```diff
@@ -10,6 +10,8 @@ pre_window:
   - alias gateway-cli="\$FM_GATEWAY_CLI"
   - alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"
   - alias fedimint-dbdump="\$FM_DB_DUMP"
+  - alias restart="./scripts/restart-tmux.sh"
+  - alias dbtool="\$FM_DB_TOOL"
```
elsirion, line 14: This line seems unrelated? Did it get added by a rebase gone wrong?

elsirion (approved):

.tmuxinator.yml
cdmoss: I accidentally deleted the dbtool line early on, and added it back in with a commit that I squashed in the rebase. I thought I had rebased properly, but I could have messed something up, I'm kind of new to maintaining a tidy history... Thanks for approval! :)

PR #1591, opened 2023-02-05
feat: Ability for `spawn` to return a value

dpc:


fedimint-api/src/task.rs
```diff
@@ -8,6 +8,8 @@ use std::sync::atomic::Ordering::SeqCst;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
+#[cfg(target_family = "wasm")]
+use futures::channel::oneshot;
```
Maan2003, line 12: you can use `tokio::sync` types on wasm now, btw

Maan2003 (approved):
LGTM

fedimint-api/src/task.rs
dpc: Oh, good to know. We don't even seem to include `tokio` in wasm builds of `fedimint-api`. If anything changed recently that makes things easier, we should probably do a whole separate PR with a pass over all these `cfg`s and try to unify them.

elsirion (approved):

PR #1592, opened 2023-02-06
feat: Run `rust-tests.sh` in parallel

dpc:
This change is trying to take care of a biggest pain point of our current state of testing suite - the long time it takes us to execute the integration tests with real LN and `bitcoind`.

Before:

image

After:

image

After quite a bit of investigation I realized that it should be possible to retain the shared infrastructure (which is great for keeping the performance high and resource usage low) yet run tests in parallel with relatively minor changes.

The biggest assumption that is being broken is strict control over block height. As multiple tests are sharing `bitcoind`, the block height can increase at any time, which will lead to random epochs being introduced.

The common pattern this breaks is:

```
join_all([
  operation,
  await_consensus_epoch(exact_amount_of_epochs_required)
])
```

where `operation` triggers new epochs, and blocks until they are processed, somewhere inside itself, while `await_consensus_epoch` is triggering the epoch processing.

This worked before because `await_consensus_epoch` blocks waiting for events triggering epochs. Since now new block heights can happen randomly, this leads to flakiness.

I refactor affected code by breaking `operation`s into smaller pieces to allow calling `await_consensus_epoch` exactly when required.

In few places this didn't seem easily doable, so I have added an ability to force lock over whole "real" `bitcoind` client to prevent random events. This serializes the execution of given tests, but better few exceptions than all tests.

Couple of minor facilities are added where necessary.

### DIFF
--- a/client/client-lib/src/lib.rs
+++ b/client/client-lib/src/lib.rs
@@ -577,6 +577,47 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
         Ok(final_notes)
     }
 
+    /// For tests only: Select notes of a given amount, and then remint them,
+    /// remove the amount of notes from the database and return it to the user.
+    ///
+    /// This is a respin of `spent_ecash` for tests, where it is neccessary
+    /// to process epochs after `self.submit_tx_with_change`. Then
+    /// `remint_ecash_await` can be called to do the rest.
+    ///
+    /// TODO: Like `spend_ecash`, I think this function works in tests mostly
+    /// by accident.
+    pub async fn remint_ecash<R: RngCore + CryptoRng>(&self, amount: Amount, rng: R) -> Result<()> {
+        let notes = self.mint_client().select_notes(amount).await?;
+
+        let mut tx = TransactionBuilder::default();
+
+        let (mut keys, input) = MintClient::ecash_input(notes)?;
+        tx.input(&mut keys, input);
+        self.submit_tx_with_change(tx, rng).await?;
+
+        Ok(())
+    }
+
+    /// Continuation of `remint_notes`
+    pub async fn remint_ecash_await(&self, amount: Amount) -> Result<TieredMulti<SpendableNote>> {
+        self.fetch_all_notes().await;
+        let notes = self.mint_client().select_notes(amount).await?;
+        assert_eq!(notes.total_amount(), amount, "should have exact change");
+
+        let mut dbtx = self.context.db.begin_transaction().await;
+        for (amount, note) in notes.iter_items() {
+            dbtx.remove_entry(&NoteKey {
+                amount,
+                nonce: note.note.0,
+            })
+            .await
+            .expect("DB Error");
+        }
+        dbtx.commit_tx().await.expect("DB Error");
+
+        Ok(notes)
+    }
+
     /// Tries to fetch e-cash notes from a certain out point. An error may just mean having queried
     /// the federation too early. Use [`MintClientError::is_retryable`] to determine
     /// if the operation should be retried at a later time.
@@ -850,15 +891,49 @@ impl Client<UserClientConfig> {
             .map_err(|_| ClientError::Timeout)?
     }
 
-    pub async fn generate_invoice<R: RngCore + CryptoRng>(
+    pub async fn generate_confirmed_invoice<R: RngCore + CryptoRng>(
         &self,
         amount: Amount,
         description: String,
         mut rng: R,
         expiry_time: Option<u64>,
     ) -> Result<ConfirmedInvoice> {
-        let gateway = self.fetch_active_gateway().await?;
+        let (txid, invoice, payment_keypair) = self
+            .generate_unsigned_invoice_and_submit(amount, description, &mut rng, expiry_time)
+            .await?;
+
+        self.await_invoice_confirmation(txid, invoice, payment_keypair)
+            .await
+    }
+    pub async fn generate_unsigned_invoice_and_submit<R: RngCore + CryptoRng>(
+        &self,
+        amount: Amount,
+        description: String,
+        mut rng: R,
+        expiry_time: Option<u64>,
+    ) -> Result<(TransactionId, Invoice, KeyPair)> {
         let payment_keypair = KeyPair::new(&self.context.secp, &mut rng);
+        let (invoice, ln_output) = self
+            .generate_unsigned_invoice(amount, description, payment_keypair, &mut rng, expiry_time)
+            .await?;
+
+        // There is no input here because this is just an announcement
+        let mut tx = TransactionBuilder::default();
+        tx.output(ln_output);
+        let txid = self.submit_tx_with_change(tx, &mut rng).await?;
+
+        Ok((txid, invoice, payment_keypair))
+    }
+
+    pub async fn generate_unsigned_invoice<R: RngCore + CryptoRng>(
+        &self,
+        amount: Amount,
+        description: String,
+        payment_keypair: KeyPair,
+        mut rng: R,
+        expiry_time: Option<u64>,
+    ) -> Result<(Invoice, Output)> {
+        let gateway = self.fetch_active_gateway().await?;
         let raw_payment_secret: [u8; 32] = payment_keypair.x_only_public_key().0.serialize();
         let payment_hash = bitcoin::secp256k1::hashes::sha256::Hash::hash(&raw_payment_secret);
         let payment_secret = PaymentSecret(raw_payment_secret);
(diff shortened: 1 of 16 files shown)


dpc:
There is still some minor flakiness in there:

```
---- lightning_gateway_pays_outgoing_invoice stdout ----
thread 'lightning_gateway_pays_outgoing_invoice' panicked at 'called `Result::unwrap()` on an `Err` value: CouldNotRoute(LightningError(Some(210)))', integrationtests/tests/tests.rs:595:14


failures:
    lightning_gateway_pays_outgoing_invoice
```

`lightningd` refusing to route at random? Possibly a simple retry somewhere will help.

This took a lot of effort, and I'm getting tired, at least for tonight. Please let me know if this seams generally reasonable, and point out any issues.

In a typical to me fashion, I just started banging at it with a hammer, jumping from the issue to the issue until it worked. Had I knew what exactly is going to be needed upfront I'd split it into smaller commits.

integrationtests/tests/tests.rs
```diff
@@ -473,17 +503,27 @@ async fn lightning_gateway_pays_internal_invoice() -> Result<()> {
...
-                fed.await_consensus_epochs(4).await.unwrap();
-            }
-        )
-        .0
-        .unwrap();
+                .pay_invoice_buy_preimage_finalize_and_claim(contract_id, buy_preimage)
+                .await
+                .unwrap()
+        };
+        info!("THERE");
```
m1sterc001guy, line 522: Remove?

integrationtests/tests/tests.rs
```diff
@@ -511,7 +551,9 @@ async fn lightning_gateway_pays_internal_invoice() -> Result<()> {
         gateway.user.assert_total_notes(sats(2010)).await; // gateway routed internally and earned fee
         receiving_user.assert_total_notes(sats(1000)).await; // this user received the 1000 sat invoice
 
-        assert_eq!(lightning.amount_sent().await, sats(0)); // We did not route any payments over the lightning network
+        if !lightning.is_shared() {
```
m1sterc001guy, line 554: This is necessary because it is possible another test sent some sats at the same time?

integrationtests/tests/tests.rs
```diff
@@ -893,22 +948,23 @@ async fn runs_consensus_if_tx_submitted() -> Result<()> {
...
-        bitcoin.mine_blocks(100);
-        let (proof, tx) = bitcoin.send_and_mine_block(&peg_in_address, Amount::from_sat(1000));
+        bitcoin.mine_blocks(100).await;
+        let (proof, tx) = bitcoin
+            .send_and_mine_block(&peg_in_address, Amount::from_sat(1000))
+            .await;
         fed.run_consensus_epochs(1).await;
 
-        // If epochs run before the blocks are mined, user won't be able to peg-in
-        join_all(vec![
```
m1sterc001guy, line 902: When writing integration tests for the future, is the general guidance to assume that another test could be running at the same time, so waiting on global state should be discouraged? (or take a global lock).

These changes are great but we should probably outline some guidelines so new tests don't replicate these issues

fedimint-testing/src/btc/fixtures.rs
```diff
@@ -113,15 +120,15 @@ impl BitcoinTest for FakeBitcoinTest {
         )
     }
 
-    fn get_new_address(&self) -> Address {
+    async fn get_new_address(&self) -> Address {
```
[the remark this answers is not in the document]
dpc: Actually no. I was doing locking differently at first, and then I needed it `async`.
 
Thought ... We should really put these blocking IO calls to bitcoind in `spawn_blocking` anyway ... so maybe I should just keep it?

gateway/ln-gateway/src/lib.rs
```diff
@@ -273,8 +273,14 @@ impl LnGateway {
         // We use a random federation as the default (works because we only have one federation registered)
         //
         // TODO: Use subscribe intercept htlc streams to avoid actor selection with every intercepted htlc!
-        self.actors.lock().await.values().collect::<Vec<_>>()[0]
-            .buy_preimage_internal(&payment_hash, &invoice_amount)
+        let lock = &self.actors.lock().await;
```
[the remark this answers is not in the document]
dpc: I don't know. I just extended the lifetime of a guard in here, the locking was already in place.

integrationtests/tests/tests.rs
dpc: I think so, yes.

integrationtests/tests/tests.rs
dpc: See the PR description for explanation. I could put it somewhere in the repo itself, but the question is where exactly. Root docstring of the testing crate maybe?

dpc:
> Removing the joins defeats the purpose of the runs_consensus_if_tx_submitted and runs_consensus_if_new_block tests.

Oh. That is good to know. I have removed that joins before adding ability to lock the node, and without a good understanding of the test. Will revert it there.

dpc:
@jkitman Actually, now that I look at them I'm a bit suspicious of them. Can you hop with on a call sometime and explain them to me? Maybe we could avoid locking there somehow.

jkitman:
> Actually, now that I look at them I'm a bit suspicious of them. Can you hop with on a call sometime and explain them to me? Maybe we could avoid locking there somehow.

Admittedly they were a bit hacky since they relied on a `sleep` to see if an epoch gets automatically triggered.  However, relying on `has_pending_epoch` isn't sufficient, since there are multiple ways to trigger an epoch https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/lib.rs#L452-L457

dpc:
> since there are multiple ways to trigger an epoch https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/lib.rs#L452-L457

We are in control of everything except bitcoind and LN nodes. Bitcoind is locked, so it won't generate any new heights, and other tests won't pay anything to this test. Federation itself and gateway is per-tests, so they are not shared. 

I think it should be OK then. Anything I missed?

jkitman:
> I think it should be OK then. Anything I missed?

The tests can still serve the purpose of confirming an epoch gets triggered I suppose.  We are only missing a check that epochs don't always get triggered (e.g. by introducing something like the dummy module which doesn't implement anything for `await_consensus_proposal`).

gateway/ln-gateway/src/lib.rs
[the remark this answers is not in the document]
justinmoon: @okjodom?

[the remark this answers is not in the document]
okjodom: lock because you are selecting an actor to use. actors is a mutex because we can create and reference new actors, or remove existing actors at runtime

PS: Actor selection and TODO described in line 275 should be complete very soon! I'm okay with the lock for now

client/client-lib/src/lib.rs
```diff
@@ -577,6 +577,43 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
         Ok(final_notes)
     }
 
+    /// Select notes of a given amount, and then remint them
+    ///
+    /// This is a respin of `spent_ecash`, and I can't make sense
+    /// of the purpose. IMO, the whole client API needs deep
+    /// rethinking. --dpc
```
elsirion, line 584: Yes, very much so :laughing:

client/client-lib/src/lib.rs
```diff
@@ -577,6 +577,43 @@ impl<T: AsRef<ClientConfig> + Clone> Client<T> {
...
+        self.submit_tx_with_change(tx, rng).await?;
+
+        Ok(())
+    }
+
+    /// Continuation of `remint_notes`
+    pub async fn remint_ecash_await(&self, amount: Amount) -> Result<TieredMulti<SpendableNote>> {
+        self.fetch_all_notes().await;
+        let notes = self.mint_client().select_notes(amount).await?;
+        assert_eq!(notes.total_amount(), amount, "should have exact change");
```
elsirion, line 601: 1. `should have exact change` isn't super correct, it's not change but "input notes" we want to have in the right denominations
2. Compared to `spend_ecash` it seems accidental that we get the right notes (mostly because we ensure to always have at least n of each denomination after @jkitman's change)

But it seems to work and a new client will be able to model this way better. I would advise against using this in production though. Maybe add some comment that it's tests-only?

scripts/rust-tests.sh
```diff
@@ -2,9 +2,9 @@
...
 set -euxo pipefail
-export RUST_LOG=info
+export RUST_LOG="${RUST_LOG:-info}"
 
-source ./scripts/setup-tests.sh
+source ./scripts/setup-tests.sh ""
 
 export FM_TEST_DISABLE_MOCKS=1
-cargo test -p fedimint-tests -- --test-threads=1
+cargo test -p fedimint-tests "$@"
```
elsirion, line 10: :tada:

elsirion (approved):
Awesome! I'd like for @jkitman to sign off since he understands the testing framework best. @okjodom seems to already have taken a look at the LN part.

client/client-lib/src/lib.rs
dpc: I thought I copied `spend_ecash` 1:1, just split it into two parts.

okjodom (dismissed):
:rocket:

integrationtests/tests/tests.rs
```diff
@@ -984,16 +1047,22 @@ async fn can_get_signed_epoch_history() -> Result<()> {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn rejoin_consensus_single_peer() -> Result<()> {
-    test(4, |fed, user, bitcoin, _, _| async move {
-        bitcoin.mine_blocks(110);
+    test(4, |fed, user, bitcoin, _, ln| async move {
+        if ln.is_shared() {
```
jkitman, line 1051: Why doesn't it work?

dpc: I should turn to a locked test.

elsirion (approved):
Will take a follow-up PR to iron out some kinks as discussed on discord, but this will make iterating on it so much faster that we want to go ahead.

PR #1596, opened 2023-02-06
fix: Use futures::Stream instead of Iterator for find_by_prefix

m1sterc001guy:
Changes find_by_prefix and raw_find_by_prefix functions in the DatabaseTransaction trait to return a Stream instead of an iterator.

https://github.com/fedimint/fedimint/issues/1322


elsirion (approved):
Looks good to go after #1592 lands. Fixes #1322.

PR #1597, opened 2023-02-06
Greedy note select

jkitman:
Transferring ecash offline takes a long time ~15-30 sec because of the all the data that needs to be transferred.

Part of the problem is our note selection algorithm isn't greedy, resulting in many notes to be transferred.

This reduces the avg number of selected notes from ~70 to ~10, tested using 100 different amounts from 1k to 100k sats.

BTW if we didn't use millisats (only sat denominations) that drops from 10 to 6.

Minor improvements in our latency tests as well:
AVG REISSUE TIME: 1.037 seconds
AVG LN SEND TIME: 3.452 seconds
AVG LN RECEIVE TIME: 3.685 seconds

Previously:
AVG REISSUE TIME: 1.420 seconds
AVG LN SEND TIME: 3.552 seconds
AVG LN RECEIVE TIME: 3.942 seconds

### DIFF
--- a/fedimint-api/src/tiered_multi.rs
+++ b/fedimint-api/src/tiered_multi.rs
@@ -370,11 +370,11 @@ mod test {
     }
 
     #[test]
-    fn select_notes_returns_exact_amount() {
+    fn select_notes_returns_exact_amount_with_minimum_notes() {
         let starting = notes(vec![
-            (Amount::from_sats(1), 5),
-            (Amount::from_sats(5), 5),
-            (Amount::from_sats(20), 5),
+            (Amount::from_sats(1), 10),
+            (Amount::from_sats(5), 10),
+            (Amount::from_sats(20), 10),
         ]);
 
         assert_eq!(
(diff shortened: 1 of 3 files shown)


fedimint-api/src/tiered_multi.rs
```diff
@@ -370,11 +370,11 @@ mod test {
     }
 
     #[test]
-    fn select_notes_returns_exact_amount() {
+    fn select_notes_returns_exact_amount_with_minimum_notes() {
```
elsirion, line 374: Validated that this test fails before the second commit :+1:

fedimint-api/src/tiered_multi.rs
```diff
@@ -149,27 +149,28 @@ where
...
-            .collect::<TieredMulti<C>>();
+        for (note_amount, note) in self.iter_items().rev() {
+            remaining -= note_amount;
+
+            if note_amount <= amount {
+                amount -= note_amount;
+                selected.push((note_amount, (*note).clone()))
+            } else if remaining < amount {
+                // we can't make exact change, so just use this note
+                amount = Amount::ZERO;
```
elsirion, line 168: This seems equivalent to a `break` of the for loop? If `amount == 0` then neither the first nor second if-else branch will ever be hit and subtracting from `remaining` also become redundant due to that.

fedimint-api/src/tiered_multi.rs
```diff
@@ -149,27 +149,28 @@ where
...
+                selected.push((note_amount, (*note).clone()))
```
elsirion, line 169: Tests still run with

```suggestion
                // we can't make exact change, so just use this note
                selected.push((note_amount, (*note).clone()));
                break;
```

elsirion (approved):

PR #1586, opened 2023-02-04
fix: Delete leftover `block_on`s blocks

dpc:
I don't think these are neccessary anymore.

PR #1599, opened 2023-02-07
feat: In test run consensus operations in parallel

dpc:
Note: Let's land #1592 first, as these two are going to create a merge conflict.

While `join` provides concurrency, it does not provide parallelism. Operations that are CPU heavy, will block other futures from running, and multiple cores will be under-utilized.

This change uses `spawn` to create multiple tasks for certain heavy and common operations, which should help utilize multiple-cores better.

### DIFF
--- a/fedimint-api/src/db/mod.rs
+++ b/fedimint-api/src/db/mod.rs
@@ -1,11 +1,11 @@
 use std::fmt::Debug;
-use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::{error::Error, marker::PhantomData};
 
 use anyhow::Result;
 use async_trait::async_trait;
+use futures::future::BoxFuture;
 use futures::{stream, Stream, StreamExt};
 use thiserror::Error;
 use tracing::{debug, instrument, trace, warn};
@@ -151,9 +151,7 @@ impl Database {
         max_retries: Option<usize>,
     ) -> Result<T, AutocommitError<E>>
     where
-        for<'a> F: Fn(
-            &'a mut DatabaseTransaction<'dt>,
-        ) -> Pin<Box<dyn Future<Output = Result<T, E>> + 'a>>,
+        for<'a> F: Fn(&'a mut DatabaseTransaction<'dt>) -> BoxFuture<'a, Result<T, E>>,
     {
         let mut retries: usize = 0;
         loop {
--- a/fedimint-server/src/lib.rs
+++ b/fedimint-server/src/lib.rs
@@ -280,7 +280,8 @@ impl FedimintServer {
         // once we produce an outcome we no longer need to rejoin
         self.rejoin_at_epoch = None;
 
-        for epoch_num in self.next_epoch_to_process()..=last_outcome.epoch {
+        let next_epoch_to_process = self.next_epoch_to_process();
+        for epoch_num in next_epoch_to_process..=last_outcome.epoch {
             let (items, epoch, prev_epoch_hash, rejected_txs, at_know_trusted_checkpoint) =
                 if epoch_num == last_outcome.epoch {
                     (
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -1,4 +1,3 @@
-use std::cell::RefCell;
 use std::collections::{BTreeMap, HashMap};
 use std::env;
 use std::future::Future;
@@ -6,9 +5,8 @@ use std::iter::repeat;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::pin::Pin;
-use std::rc::Rc;
-use std::sync::atomic::AtomicU16;
 use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicI64, AtomicU16};
 use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
@@ -60,6 +58,7 @@ use fedimint_wallet::{SpendableUTXO, WalletGen};
 use futures::executor::block_on;
 use futures::future::{join_all, select_all};
 use futures::FutureExt;
+use futures::StreamExt;
 use hbbft::honey_badger::Batch;
 use itertools::Itertools;
 use lightning_invoice::Invoice;
@@ -80,6 +79,7 @@ use mint_client::{
 use rand::rngs::OsRng;
 use rand::RngCore;
 use real::{RealBitcoinTest, RealLightningTest};
+use tokio::sync::Mutex;
 use tracing::info;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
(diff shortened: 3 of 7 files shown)


integrationtests/tests/fixtures/mod.rs
```diff
@@ -945,16 +958,18 @@ impl FederationTest {
...
-            )
-            .await
-            {
-                maybe_cancelled?;
+            let mut task_group = TaskGroup::new();
+            for (i, server) in self.servers.iter().enumerate() {
+                let server = server.clone();
+                task_group
+                    .spawn(format!("server-{i}-consensu_epoch"), move |_| async {
+                        Self::consensus_epoch(server, Duration::from_millis(0)).await
```
dpc, line 966: This was the main goal here. `await_consensus_epochs` is used in a serial fashion all over the test suite and has a significant impact.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -755,17 +761,27 @@ impl FederationTest {
...
-            self.await_consensus_epochs(2)
-        )
-        .0
-        .unwrap()
+        let mut task_group = TaskGroup::new();
+
+        let client = user.client.clone();
+
+        let res = task_group
+            .spawn("spend ecash", move |_| async move {
```
dpc, line 769: I might need to redo / scrap this one, after #1592 lands.

fedimint-server/src/lib.rs
```diff
@@ -280,7 +280,8 @@ impl FedimintServer {
         // once we produce an outcome we no longer need to rejoin
         self.rejoin_at_epoch = None;
 
-        for epoch_num in self.next_epoch_to_process()..=last_outcome.epoch {
+        let next_epoch_to_process = self.next_epoch_to_process();
+        for epoch_num in next_epoch_to_process..=last_outcome.epoch {
```
dpc, line 284: I don't understand yet why this is needed. https://users.rust-lang.org/t/i-dont-understand-this-lifetime-async-quirk/88883

elsirion: Fascinating problem!

integrationtests/tests/fixtures/mod.rs
elsirion, line 966: I don't think you need to spawn here? Wouldn't [`join_all`](https://docs.rs/futures/latest/futures/future/fn.join_all.html#) over an iterator of these futures also do the trick?

elsirion (approved):
Only a nit, looks great!

fedimint-server/src/lib.rs
dpc: https://github.com/rust-lang/rust/issues/57017

integrationtests/tests/fixtures/mod.rs
dpc: `join_all` will not create separate tasks, so concurrency, but no parallelism.

elsirion: … which we need till other problems are fixed (moving heavy compute off tokio).

integrationtests/tests/fixtures/mod.rs
dpc: I scrapped it.

fedimint-api/src/db/mod.rs
```diff
@@ -153,7 +153,7 @@ impl Database {
     where
         for<'a> F: Fn(
             &'a mut DatabaseTransaction<'dt>,
-        ) -> Pin<Box<dyn Future<Output = Result<T, E>> + 'a>>,
+        ) -> Pin<Box<dyn Future<Output = Result<T, E>> + Send + 'a>>,
```
elsirion, line 156: Now that it's `Send` it looks like a `BoxFuture`?

integrationtests/tests/fixtures/mod.rs
elsirion: But isn't the number of executor threads determined by the runtime anyway? Calling `tokio::spawn` does not mean tokio creates an actual thread afaik, just that this new task will be scheduled on available threads independent of the current task.

elsirion (approved):
LGTM

integrationtests/tests/fixtures/mod.rs
dpc: Yes. And chances are other runtime threads have nothing better to do. :)

elsirion: Ah, I think I get your point now, even if compute is negligible/wouldn't *fill up* a thread it might lead to more latency if it is very bursty and happens at the same time.

elsirion (approved):

mxz42 (approved):
cool!

mxz42 (approved):

integrationtests/tests/fixtures/mod.rs
```diff
@@ -780,7 +786,7 @@ impl FederationTest {
     }
 
     /// Helper to issue change for a user
-    pub async fn spend_ecash<C: AsRef<ClientConfig> + Clone>(
+    pub async fn spend_ecash<C: AsRef<ClientConfig> + Clone + Send + Sync + 'static>(
```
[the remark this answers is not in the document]
dpc: I must have gotten confused.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1129,44 +1138,72 @@ impl FederationTest {
...
 
-    fn update_last_consensus(&self) {
-        let new_consensus = self
-            .servers
-            .iter()
-            .flat_map(|s| s.borrow().last_consensus.clone())
-            .max_by_key(|c| c.epoch)
+    async fn update_last_consensus(&self) {
+        let new_consensus = futures::stream::iter(self.servers.iter())
+            .then(|s| s.lock())
```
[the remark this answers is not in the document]
dpc: Yes. It will become a part of some big thing that implements `Stream` that will have `.fold` ran on it, which consumes it, which calls drops on everything inside it.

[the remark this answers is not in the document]
mxz42: you can't answer an option A *or* option B question with "yes" :'D 

so the lock is dropped at the end of `flat_map` after the value is cloned because the look is not needed afterwards, right?

mxz42 (approved):

integrationtests/tests/fixtures/mod.rs
```diff
@@ -840,7 +846,7 @@ impl FederationTest {
 
         for server in &self.servers {
             block_on(async {
-                let svr = server.borrow_mut();
+                let svr = server.lock().await;
                 let mut dbtx = svr.database.begin_transaction().await;
```
elsirion, line 850: This loop is similar in nature to the one parallelized. Probably lower impact though, just noticed while looking at #1586.

elsirion (approved):

integrationtests/tests/fixtures/mod.rs
dpc: @elsirion Yeah, but only does IO, and database ops already yield executor so I don't think there's a reason to complicate with `spawn`.

PR #1602, opened 2023-02-07
Fixes some of the flakiness of reconnecting

jkitman:
Realized that sometimes the reconnect can stall because the `run_empty_epochs` may be overridden with a smaller number.  This change makes that additive.

### DIFF
--- a/fedimint-server/src/lib.rs
+++ b/fedimint-server/src/lib.rs
@@ -480,8 +480,11 @@ impl FedimintServer {
                 Ok(self.handle_step(step).await?)
             }
             (_, EpochMessage::RejoinRequest(epoch)) => {
-                info!("Requested to run {} epochs", epoch);
-                self.run_empty_epochs = min(NUM_EPOCHS_REJOIN_AHEAD, epoch);
+                self.run_empty_epochs += min(NUM_EPOCHS_REJOIN_AHEAD, epoch);
+                info!(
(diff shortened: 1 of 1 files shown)


elsirion (approved):
Seems reasonable, we need to take more care not to accidentally spam rejoin messages after this change though since they aren't idempotent anymore (well, never truly were, but now the consequences would be bigger).

fedimint-server/src/lib.rs
```diff
@@ -480,8 +480,11 @@ impl FedimintServer {
                 Ok(self.handle_step(step).await?)
             }
             (_, EpochMessage::RejoinRequest(epoch)) => {
-                info!("Requested to run {} epochs", epoch);
-                self.run_empty_epochs = min(NUM_EPOCHS_REJOIN_AHEAD, epoch);
+                self.run_empty_epochs += min(NUM_EPOCHS_REJOIN_AHEAD, epoch);
```
dpc, line 483: This function, this field, this message - none of them have a comment. `NUM_EPOCHS_REJOIN_AHEAD` does have a comment "how many epochs ahead of consensus to rejoin".

I don't understand why we would rejoin 'ahead' of consensus? Like, to epochs that are not even made yet?

I don't understand. :(

jkitman: I'll try to add a better comment around that.

PR #1605, opened 2023-02-07
Caches the submitted txs instead of using DB

jkitman:
Avoids putting pending txs into the DB.

We have to use a `Mutex` around the cache, or alternatively we could put the cache into `FedimintServer` which seems worse.

### DIFF
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -7,6 +7,7 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::ffi::OsString;
 use std::iter::FromIterator;
 use std::os::unix::prelude::OsStrExt;
+use std::sync::Mutex;
 
 use fedimint_api::config::{ConfigResponse, ModuleGenRegistry};
 use fedimint_api::core::ModuleInstanceId;
@@ -33,8 +34,7 @@ use crate::config::ServerConfig;
 use crate::consensus::interconnect::FedimintInterconnect;
 use crate::db::{
     AcceptedTransactionKey, ClientConfigSignatureKey, DropPeerKey, DropPeerKeyPrefix,
-    EpochHistoryKey, LastEpochKey, ProposedTransactionKey, ProposedTransactionKeyPrefix,
-    RejectedTransactionKey,
+    EpochHistoryKey, LastEpochKey, RejectedTransactionKey,
 };
 use crate::logging::LOG_CONSENSUS;
 use crate::transaction::{Transaction, TransactionError};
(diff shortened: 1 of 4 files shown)


fedimint-server/src/consensus/mod.rs
```diff
@@ -7,6 +7,7 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::ffi::OsString;
 use std::iter::FromIterator;
 use std::os::unix::prelude::OsStrExt;
+use std::sync::Mutex;
```
justinmoon, line 10: Do we want tokio mutex?

jkitman: I don't need to keep it locked across an `await` point, according to the docs it's more lightweight just to use the std one. https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html

elsirion: Not only that, but also the fact that there should never be contention, so it isn't blocking for too long I assume?

fedimint-server/src/consensus/mod.rs
```diff
@@ -647,15 +634,14 @@ impl FedimintConsensus {
...
-            .map(|res| {
-                let (_key, value) = res.expect("DB error");
-                ConsensusItem::Transaction(value)
-            })
-            .collect()
-            .await;
+        let mut items: Vec<ConsensusItem> = self
+            .tx_cache
+            .lock()
+            .unwrap()
```
elsirion, line 640: nit: prefer `expect` over `unwrap`

elsirion (approved):

fedimint-server/src/consensus/mod.rs
jkitman: I think `.lock()` is a special case that doesn't allow `expect`

elsirion: 🤨

dpc: No way. `Result` is a `Result`, has both `expect` and `unwrap`.

fedimint-server/src/consensus/mod.rs
dpc: It's correct, but the idea of keeping track which Mutex is being used and making sure it's correct is making me sceptical. I'm tempted to ban `std::sync::Mutex` altogether somehow. :D

fedimint-server/src/consensus/mod.rs
```diff
@@ -406,9 +394,8 @@ impl FedimintConsensus {
             let span = info_span!("Processing transaction");
             async {
                 trace!(?transaction);
-                dbtx.remove_entry(&ProposedTransactionKey(txid))
-                    .await
-                    .expect("DB Error");
+                let mut tx_cache = self.tx_cache.lock().unwrap();
```
dpc, line 397: Holding a blocking mutex across `.await`. :fearful: 

We're lucky that compiler complained about it when I was rebasing #1599

jkitman: Not sure what the alternative is though in this case...

elsirion: tokio mutex?

dpc: Nothing comes to my mind that would be better than the existing Rust lints. I discovered this because the whole future now needs to be `Send`, etc. Seems like not an easy problem to solve.

PR #1608, opened 2023-02-07
feat: Add macro for implementing DatabaseKeyPrefixConst

m1sterc001guy:
Database code currently contains a lot of boilerplate code for reading/writing keys, key prefixes, and their associated values. Once every DB struct is versioned, the amount of boilerplate code will become a lot.

This PR attempts to make a code a bit more readable and just add a macro for generating the common key/value and key prefix/value implementations needed for reading/writing to the database.

### DIFF
--- a/client/client-lib/src/mint/db.rs
+++ b/client/client-lib/src/mint/db.rs
@@ -33,7 +33,12 @@ pub struct NoteKey {
 #[derive(Debug, Clone, Encodable, Decodable)]
 pub struct NoteKeyPrefix;
 
-impl_db_prefix_const!(NoteKey, NoteKeyPrefix, SpendableNote, DbKeyPrefix::Note);
+impl_db_prefix_const!(
+    key = NoteKey,
+    value = SpendableNote,
+    prefix = DbKeyPrefix::Note,
+    key_prefix = NoteKeyPrefix
+);
 
 #[derive(Debug, Clone, Encodable, Decodable, Serialize)]
 pub struct PendingNotesKey(pub TransactionId);
@@ -42,10 +47,10 @@ pub struct PendingNotesKey(pub TransactionId);
 pub struct PendingNotesKeyPrefix;
 
 impl_db_prefix_const!(
-    PendingNotesKey,
-    PendingNotesKeyPrefix,
-    TieredMulti<SpendableNote>,
-    DbKeyPrefix::PendingNotes
+    key = PendingNotesKey,
+    value = TieredMulti<SpendableNote>,
+    prefix = DbKeyPrefix::PendingNotes,
+    key_prefix = PendingNotesKeyPrefix
 );
 
 #[derive(Debug, Clone, PartialEq, Eq, Encodable, Decodable, Serialize, Deserialize)]
@@ -55,10 +60,10 @@ pub struct OutputFinalizationKey(pub OutPoint);
 pub struct OutputFinalizationKeyPrefix;
 
 impl_db_prefix_const!(
-    OutputFinalizationKey,
-    OutputFinalizationKeyPrefix,
-    NoteIssuanceRequests,
-    DbKeyPrefix::OutputFinalizationData
+    key = OutputFinalizationKey,
+    value = NoteIssuanceRequests,
+    prefix = DbKeyPrefix::OutputFinalizationData,
+    key_prefix = OutputFinalizationKeyPrefix
 );
 
 #[derive(Debug, Clone, Encodable, Decodable, Serialize)]
@@ -68,21 +73,13 @@ pub struct NextECashNoteIndexKey(pub Amount);
 pub struct NextECashNoteIndexKeyPrefix;
 
 impl_db_prefix_const!(
-    NextECashNoteIndexKey,
-    NextECashNoteIndexKeyPrefix,
-    u64,
-    DbKeyPrefix::NextECashNoteIndex
+    key = NextECashNoteIndexKey,
+    value = u64,
+    prefix = DbKeyPrefix::NextECashNoteIndex,
+    key_prefix = NextECashNoteIndexKeyPrefix
 );
 
 #[derive(Debug, Clone, Encodable, Decodable, Serialize)]
 pub struct NotesPerDenominationKey;
 
-#[derive(Debug, Clone, Encodable, Decodable, Serialize)]
-pub struct NotesPerDenominationKeyPrefix;
-
-impl_db_prefix_const!(
-    NotesPerDenominationKey,
-    NotesPerDenominationKeyPrefix,
-    u16,
-    0
-);
+impl_db_prefix_const!(key = NotesPerDenominationKey, value = u16, prefix = 0);
--- a/client/client-lib/src/mint/db.rs
+++ b/client/client-lib/src/mint/db.rs
@@ -1,4 +1,3 @@
-use fedimint_api::db::DatabaseKeyPrefixConst;
 use fedimint_api::encoding::{Decodable, Encodable};
 use fedimint_api::impl_db_prefix_const;
 use fedimint_api::{Amount, OutPoint, TieredMulti, TransactionId};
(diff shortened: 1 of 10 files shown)


fedimint-api/src/db/mod.rs
```diff
@@ -680,6 +680,23 @@ where
     }
 }
 
+#[macro_export]
+macro_rules! impl_db_prefix_const {
```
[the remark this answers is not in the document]
dpc: Not nice - a must! :D 

I still don't understand what and why w.r.t. all these traits, and now it's going to be hidden behind and opaque macro.

Please. We need at least a one line on each of the major item, and possibly some explanation of how things works together in some places.

client/client-lib/src/mint/db.rs
```diff
@@ -29,80 +30,59 @@ pub struct NoteKey {
...
 #[derive(Debug, Clone, Encodable, Decodable, Serialize)]
 pub struct NotesPerDenominationKey;
 
-impl DatabaseKeyPrefixConst for NotesPerDenominationKey {
-    const DB_PREFIX: u8 = 0;
-    type Key = Self;
-    type Value = u16;
-}
+#[derive(Debug, Clone, Encodable, Decodable, Serialize)]
+pub struct NotesPerDenominationKeyPrefix;
```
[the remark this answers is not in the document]
m1sterc001guy: Great suggestion, thank you

fedimint-api/src/db/mod.rs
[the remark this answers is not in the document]
m1sterc001guy: Added rustdoc on usage as description of parameters, including examples.

fedimint-api/src/db/mod.rs
```diff
@@ -680,6 +680,46 @@ where
...
+    ($key:ty, $val:ty, $prefix:expr) => {
+        impl DatabaseKeyPrefixConst for $key {
+            const DB_PREFIX: u8 = $prefix as u8;
+            type Key = Self;
+            type Value = $val;
+        }
+    };
+    ($key:ty, $val:ty, $prefix:expr, $($prefix_ty:ty),+) => {
```
Maan2003, line 706: any reason for multiple rules?

use named arguments in macros whenever the order can be not obvious. 
```suggestion
    (key = $key:ty, value = $val:ty, prefix = $prefix:expr $(, key_prefix = $prefix_ty:ty)* $(,)?) => {
```

fedimint-api/src/db/mod.rs
```diff
@@ -680,6 +680,40 @@ where
...
+///
+/// Examples:
+///
+/// `impl_db_prefix_const!(key = TestKey, value = TestVal, prefix = TestDbKeyPrefix::Test);`
+///
+/// `impl_db_prefix_const!(key = AltTestKey, value = TestVal, prefix = TestDbKeyPrefix::AltTest, key_prefix = AltDbPrefixTestPrefix);`
+#[macro_export]
+macro_rules! impl_db_prefix_const {
+    (key = $key:ty, value = $val:ty, prefix = $prefix:expr $(, key_prefix = $prefix_ty:ty)* $(,)?) => {
+        impl DatabaseKeyPrefixConst for $key {
```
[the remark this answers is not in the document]
Maan2003: @m1sterc001guy $crate is actually real syntax (refers to crate where macro is defined)

dpc (dismissed):

elsirion:
Needs a rebase after #1625 to make CI happy.

modules/fedimint-mint/src/db.rs
```diff
@@ -28,91 +28,66 @@ impl std::fmt::Display for DbKeyPrefix {
...
-    type Key = ReceivedPartialSignatureKey;
-    type Value = MintOutputSignatureShare;
-}
+impl_db_prefix_const!(
+    key = ReceivedPartialSignatureKey,
+    value = MintOutputSignatureShare,
+    prefix = DbKeyPrefix::ReceivedPartialSig,
+    key_prefix = ReceivedPartialSignaturesKeyPrefix,
+    key_prefix = ReceivedPartialSignatureKeyOutputPrefix
+);
```
elsirion, line 76: Sweet!

elsirion (approved):

fedimint-api/src/db/mod.rs
```diff
@@ -680,6 +680,40 @@ where
     }
 }
 
+/// This is a helper macro that generates the implementations of
+/// DatabaseKeyPrefixConst necessary for reading/writing to the
+/// database and fetching by prefix.
+///
+/// - `key`: This is the type of struct that will be used as the key into the database.
+/// - `value`: This is the type of struct that will be used as the value into the database.
+/// - `prefix`: Enum expression that is represented as a `u8` that is the prefix prepended to this key.
```
mxz42, line 689: is there a reason why this is not `db_prefix` or `prefix_db` anymore?

fedimint-api/src/db/mod.rs
```diff
@@ -680,6 +680,40 @@ where
...
+/// - `key_prefix`: Optional type of struct that can be passed multiple times if necessary. These will be used
```
mxz42, line 690: assuming we called the `prefix` -> `prefix_db`, then we could call the `key_prefix`  `prefix_query` or similar. As it's used to *query* for a specific key prefixed first with the `prefix_db` itself.  

it's `<prefix_db><prefix_query><rest_of_key_bytes>`, right?

mxz42 (approved):
Wow, that's so much cleaner now. :+1: 

I only have a nit and question about the naming. With this PR I think I finally understood what all the *prefixes* are doing and how they work together. 
And the one prefix (called by me `prefix_query`) is only used by `find_by_prefix()`? 

The renaming may make it easier to distinguish between the prefix for the DB itself and zero or more query prefixes for the key.

fedimint-api/src/db/mod.rs
m1sterc001guy: `prefix_query` is just the struct that holds the prefix byte and defines the types for the `Key` and `Value`. I enabled auto-merge so this already got merged, but we can rename them in a follow up if you'd like.

On disk I believe it looks something like this:
`<Global Module Prefix Byte><Module Prefix Byte Upper><Module Prefix Byte Lower><prefix_db><rest of key bytes>` 

The global module prefix byte and module prefix bytes are appended even before `prefix_db` when the key belongs to a module.

fedimint-api/src/db/mod.rs
m1sterc001guy: No reason. We can rename in a follow up if you'd like

fedimint-api/src/db/mod.rs
mxz42: so just for completeness as we talked off-thread. the `<module byte upper>`  and `<module byte lower>` is the module instance ID, which is a `u16` so 2 bytes.  The global module prefix is currently defined as `0xff`. 
so
`<Global Module Prefix><Module Instance Prefix><prefix_db><key_bytes>`

thanks @m1sterc001guy 

@ others: also take a look at [the DB documentation](https://github.com/fedimint/fedimint/blob/master/docs/database.md#server-db-layout)

fedimint-api/src/db/mod.rs
mxz42: what do you think of the idea? 

IMHO, it may improve readability and reduce the time to understand how these prefixes work together. The `db_prefix` is always there as it is essential for the type. And then there usually is at least one `query_prefix` that returns all the values for a type.  (how is that enforced in the macro actually? is `$(, key_prefix = $prefix_ty:ty)* $(,)?` *zero or more* or *one or more*?)    

rename 

- `prefix` to `db_prefix`
- `key_prefix` to `query_prefix`

m1sterc001guy: > (how is that enforced in the macro actually? is $(, key_prefix = $prefix_ty:ty)* $(,)? zero or more or one or more?)

Yes exactly.

I'm good with all of the changes proposed here, feel free to open PR or I will get to it soon :)

mxz42: I can do it if you like :) depends on what soon is :'D

PR #1611, opened 2023-02-07
fix: conventional commit checking when using `git commit -m`

dpc:
The approach I used before (starting editor to edit commit message) does not play well with how git commit hooks work (IMO, poorly designed).

Instead of opening an editor, print an error message and introduce a helper to try again without figuring out how to recover the old message.

### DIFF
--- a/misc/git-hooks/commit-msg
+++ b/misc/git-hooks/commit-msg
@@ -10,44 +10,21 @@ function rm_tmp_file {
 
 trap rm_tmp_file EXIT
 
-while : ; do
-  # Sanitize file first, by removing leading lines that are empty or start with a hash,
-  # as `convco` currently does not do it automatically (but git will)
-  echo -n "" > "$tmp_file"
-  body_detected=""
-  while read -r line ; do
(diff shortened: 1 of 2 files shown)


misc/git-hooks/commit-msg
```diff
@@ -10,44 +10,21 @@ function rm_tmp_file {
...
+# Sanitize file first, by removing leading lines that are empty or start with a hash,
+# as `convco` currently does not do it automatically (but git will)
+echo -n "" > "$tmp_file"
+body_detected=""
+while read -r line ; do
+  # skip any initial comments (possibly from previous run)
+  if [ -z "${body_detected:-}" ] && { [[ "$line" =~ ^#.*$ ]] || [ "$line" == "" ]; }; then
+    continue
   fi
+  body_detected="true"
```
elsirion, line 22: Where is `body_detected` used later on? I can't find it. Is there some implicit transfer to the caller?

dpc: Leftover that I overlooked.

elsirion (approved):

PR #1616, opened 2023-02-08
fix: Print lightning gateway version at startup

### LINKED ISSUE #1607: `ln_gateway` should print out its code version
@asifshiraz Could you have `ln_gateway` also print out it code version?

_Originally posted by @justinmoon in https://github.com/fedimint/fedimint/issues/1581#issuecomment-1421136411_

asifshiraz:
fixes #1607

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -81,6 +81,8 @@ impl LnGateway {
         receiver: mpsc::Receiver<GatewayRequest>,
         task_group: TaskGroup,
     ) -> Self {
+        info!(version = env!("GIT_HASH"), "Starting lightning gateway");
+
         let mut num_retries = 0;
         let route_hints = loop {
             let route_hints = ln_rpc


gateway/ln-gateway/src/lib.rs
```diff
@@ -118,6 +118,9 @@ impl LnGateway {
             .load_federation_actors(decoders, module_gens, route_hints)
             .await;
 
+        let version = env!("GIT_HASH").to_string();
+        info!("Starting lightning gateway (version: {version}");
```
mxz42, line 122: This currently prints 

```
2023-02-08T16:44:31.786Z INFO    plugin-ln_gateway: Starting lightning gateway (version: 04a59457a253243428dab6dc4aa844d1ac4577a9
```
There's a missing closing `)` or better remove the opening `(` I guess :)

Also, since we only print the version hash, we don't need to allocate via `.to_string()`. We can just take the `&str` from the `env` macro.

dpc, line 122: BTW. Just looking at it now, we should use structural logging like this: ` info!(version, "Starting lightning gateway");` (or `%version`).

mxz42:
just another thought: maybe move the logging statement to the start of the `new` method, so that it's the first logging statement that appears for ln_gateway before `LnRpc::route_hints;` and `LnRpc::pubkey;` and so on. So that even if one of these calls fail, the version is already printed.

gateway/ln-gateway/src/lib.rs
asifshiraz: Done

dpc (dismissed):

mxz42 (approved):

dpc (approved):

mxz42 (approved):

mxz42:
thanks! @asifshiraz

PR #1619, opened 2023-02-08
Replay prevention2

### LINKED ISSUE #1347: Prevent transaction replays

### LINKED ISSUE #1353: Stop using the consensus DB for submitted transactions

jkitman:
Prevents the replay of transactions with an error at the API and P2P levels with integration tests.  Small refactor of some test code as a separate commit.

Should close #1402
Last bit of work to close #1353
Also should close #1347

### DIFF
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -22,9 +22,10 @@ use fedimint_api::cancellable::Cancellable;
 use fedimint_api::config::{ClientConfig, ModuleGenRegistry};
 use fedimint_api::core;
 use fedimint_api::core::{
-    DynModuleConsensusItem as PerModuleConsensusItem, ModuleConsensusItem,
+    ModuleConsensusItem,
     LEGACY_HARDCODED_INSTANCE_ID_MINT, LEGACY_HARDCODED_INSTANCE_ID_WALLET,
 };
+use fedimint_api::core::{DynModuleConsensusItem, ModuleInstanceId};
 use fedimint_api::db::mem_impl::MemDatabase;
 use fedimint_api::db::Database;
 use fedimint_api::module::registry::{ModuleDecoderRegistry, ModuleRegistry};
@@ -47,6 +48,7 @@ use fedimint_server::multiplexed::PeerConnectionMultiplexer;
 use fedimint_server::net::connect::mock::MockNetwork;
 use fedimint_server::net::connect::{Connector, TlsTcpConnector};
 use fedimint_server::net::peers::PeerConnector;
+use fedimint_server::transaction::legacy::Transaction;
 use fedimint_server::{consensus, EpochMessage, FedimintServer};
 use fedimint_testing::btc::{fixtures::FakeBitcoinTest, BitcoinTest};
 use fedimint_wallet::config::WalletConfig;
@@ -68,6 +70,7 @@ use ln_gateway::{
     LnGateway,
 };
 use mint_client::module_decode_stubs;
+use mint_client::transaction::TransactionBuilder;
 use mint_client::{
     api::WsFederationApi, mint::SpendableNote, Client, GatewayClient, GatewayClientConfig,
     UserClient, UserClientConfig,
@@ -75,7 +78,7 @@ use mint_client::{
 use rand::rngs::OsRng;
 use rand::RngCore;
 use real::{RealBitcoinTest, RealLightningTest};
-use tracing::{debug, info};
+use tracing::{info};
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
 use url::Url;
@@ -607,6 +610,22 @@ impl<T: AsRef<ClientConfig> + Clone> UserTest<T> {
         UserTest { client, config }
     }
 
+    /// Creates a transaction for testing submissions to the federation
+    pub async fn tx_with_change(&self, builder: TransactionBuilder, change: Amount) -> Transaction {
+        let mint = self.client.mint_client();
+        let mut dbtx = mint.start_dbtx().await;
+
+        builder
+            .build_with_change(
+                self.client.mint_client(),
+                &mut dbtx,
+                rng(),
+                vec![change],
+                &secp(),
+            )
+            .await
+    }
+
     /// Helper for readability
     pub async fn set_notes_per_denomination(&self, notes: u16) {
         self.client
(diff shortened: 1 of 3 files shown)


integrationtests/tests/fixtures/mod.rs
```diff
@@ -670,19 +690,18 @@ struct ServerTest {
...
+    /// Returns the first item with the given module id from the last consensus outcome
+    pub fn find_module_item(&self, id: ModuleInstanceId) -> Option<DynModuleConsensusItem> {
+        self.last_consensus
+            .borrow()
+            .clone()
             .contributions
             .values()
             .flat_map(|items| items.clone())
-            .collect()
+            .find_map(|ci| match ci {
```
dpc, line 701: Might be good to assert there was only one to prevent any accidental non-determinism?

integrationtests/tests/fixtures/mod.rs
```diff
@@ -832,6 +851,35 @@ impl FederationTest {
...
+    /// has already been spent
+    pub async fn clear_ecash_nonces(&self) {
+        for server in &self.servers {
+            block_on(async {
+                let svr = server.borrow_mut();
+                let mut dbtx = svr.database.begin_transaction().await;
+
+                {
+                    let mut module_dbtx =
+                        dbtx.with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_MINT);
```
dpc, line 864: :(

Maybe we could make a `mint_module_id: ModuleInstanceId` in `FederationTest` and start using that to not proliferate this hardcoding that is supposed to go away?

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1210,18 +1262,11 @@ impl FederationTest {
...
-}
-
-pub fn assert_module_ci<M: ModuleConsensusItem>(mci: &PerModuleConsensusItem) -> &M {
-    debug!(
-        module_instance_id = mci.module_instance_id(),
-        "Checking module consensus item"
-    );
-    mci.as_any().downcast_ref().unwrap()
+/// Unwraps a dyn consensus item into a specific one for making assertions
+pub fn unwrap_item<M: ModuleConsensusItem>(mci: &Option<DynModuleConsensusItem>) -> &M {
```
dpc, line 1266: Might add `#[track_caller]` so when it panics, it panics with the place calling it, so it's easier to figure out which type was it.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1210,18 +1262,11 @@ impl FederationTest {
...
+    mci.as_ref()
+        .expect("Module item exists")
```
dpc, line 1268: If you require it to exist, why take `Option`, and not just `&DynModuleConsensusItem`.

integrationtests/tests/tests.rs
```diff
@@ -1199,3 +1164,38 @@ async fn verifies_client_configs() -> Result<()> {
     })
     .await
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn cannot_replay_transactions() -> Result<()> {
+    test(4, |fed, user, bitcoin, _, _| async move {
+        fed.mine_and_mint(&user, &*bitcoin, sats(5000)).await;
```
dpc, line 1171: Need to call `fed.prepare_funding_wallet` to ensure funds are available.

Not sure if we should just do it from `test`, because 95% of tests will want it...

integrationtests/tests/tests.rs
```diff
@@ -1199,3 +1164,38 @@ async fn verifies_client_configs() -> Result<()> {
...
+        let tx_typed = user.tx_with_change(builder, sats(5000)).await;
+        let tx = tx_typed.into_type_erased();
+        let mint_id = LEGACY_HARDCODED_INSTANCE_ID_MINT;
+
+        // submit the tx successfully
+        let response = fed.submit_transaction(tx.clone()).await;
+        assert_matches!(response, Ok(()));
+        fed.run_empty_epochs(2).await;
+        assert!(fed.find_module_item(mint_id).is_some());
+        fed.clear_ecash_nonces().await;
```
dpc, line 1186: Could we verify that it failed at API level by comparing error message instead of clearing spent nonces?

RN if both `fed.clear_ecash_nonces` and replay prevention logic are buggy, then you might still get an error and not notice, so the failure mode is not ideal.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -832,6 +851,35 @@ impl FederationTest {
         }
     }
 
+    /// Removes the ecash nonces from the fed DB to simulate the fed losing track of what ecash
+    /// has already been spent
+    pub async fn clear_ecash_nonces(&self) {
```
dpc, line 856: `clear_spent_mint_nonces` (I miss the word "spent" most)

integrationtests/tests/fixtures/mod.rs
```diff
@@ -832,6 +851,35 @@ impl FederationTest {
...
+        for server in &self.servers {
+            block_on(async {
+                let svr = server.borrow_mut();
+                let mut dbtx = svr.database.begin_transaction().await;
+
+                {
+                    let mut module_dbtx =
+                        dbtx.with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_MINT);
+
+                    let nonces: Vec<NonceKey> = module_dbtx
```
dpc, line 866: We have `remove_by_prefix` now, so this could be done in op?

integrationtests/tests/fixtures/mod.rs
```diff
@@ -832,6 +851,35 @@ impl FederationTest {
...
+                        .find_by_prefix(&NonceKeyPrefix)
```
dpc, line 867: BTW. Should we rename this key to `SpentNonceKeyPrefix` (it's a critical distinction, and I think it would make understanding code much easier @m1sterc001guy )

elsirion (approved):

integrationtests/tests/fixtures/mod.rs
jkitman: Sure I can add that

integrationtests/tests/fixtures/mod.rs
jkitman: I'll do that as a separate refactor since it occurs in several of the tests

integrationtests/tests/fixtures/mod.rs
jkitman: It's really a convenience method for the results of `find_module_item`

integrationtests/tests/tests.rs
jkitman: Yeah, let's do that as a follow-up refactor, for now I'll add it to the test.

integrationtests/tests/tests.rs
jkitman: I do check for that at the API level, but at the P2P level I'm not sure how to easily get at that error.

integrationtests/tests/fixtures/mod.rs
jkitman: I'd be down for that in a separate PR

dpc (dismissed):

integrationtests/tests/tests.rs
dpc: Oh, OK.

integrationtests/tests/fixtures/mod.rs
jkitman: Hmm actually it doesn't really work since you get an item for every peer, plus sometimes you are asserting that no item exists.  I think it's fine because it works similar to how `find` does.

dpc (approved):

Issue #865, opened 2022-10-31
Investigate adding support for `tokio-console` and `tokio-backtrace`


dpc:
* https://tokio.rs/blog/2022-10-announcing-async-backtrace
* https://docs.rs/tokio-console/latest/tokio_console/

Both could help debugging issues and optimizing code.

elsirion:
For Jäger see also: #107

ngutech21:
Support for Jäger and tokio-console has been added. Can this issue be closed or are we still waiting for adding async-backtrace?

PR #1620, opened 2023-02-08
feat: `tokio-console` support

dpc:
The whole feature is said to be "unstable", but it might turned out very useful, and it's enabled only optionally, so should be OK.

Run `fedimintd` with an extra `--tokio-console-bind=127.0.0.1:1111` to enable. Connect with `tokio-console http://127.0.0.1:1111` (don't forget about `http://` part or it will not work, but not give you a plain error either).

Preview:

image

Re: #865

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,9 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
+
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
-rustflags = ["-C", "link-arg=-fuse-ld=lld"]
+rustflags = ["-C", "link-arg=-fuse-ld=lld", "--cfg", "tokio_unstable"]
 
 # mold is currently broken on MacOS
 # [target.aarch64-apple-darwin]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -569,6 +569,42 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d7b894f5411737b7867f4827955924d7c254fc9f4d91a6aad6b097804b1018b"
 
 [[package]]
+name = "console-api"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e57ff02e8ad8e06ab9731d5dc72dc23bef9200778eae1a89d555d8c42e5d4a86"
+dependencies = [
+ "prost",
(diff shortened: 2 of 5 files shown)


dpc:
@Maan2003 I'm out of context w.r.t how did `cln-plugin` work, and IIRC it doesn't even use `tracing` yet? So I guess we can't enable it for `ln-gateway` RN, but would be a good argument to convince upstream to switch to `tracing`?

.cargo/config.toml
```diff
@@ -1,6 +1,9 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
+
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
-rustflags = ["-C", "link-arg=-fuse-ld=lld"]
+rustflags = ["-C", "link-arg=-fuse-ld=lld", "--cfg", "tokio_unstable"]
```
elsirion, line 6: Will this be baked into release builds? We should probably avoid having unstable features in releases. Right now Fedimint is unstable itself, but I fear that we will forget about it.

elsirion:
> @Maan2003 I'm out of context w.r.t how did `cln-plugin` work, and IIRC it doesn't even use `tracing` yet? So I guess we can't enable it for `ln-gateway` RN, but would be a good argument to convince upstream to switch to `tracing`?

I think @Maan2003 created a "log to disk" patch which used tracing iirc. Maybe we can adapt that. We should be able to both use the tracing -> log and the console subscriber at the same time afaik.

.cargo/config.toml
dpc: AFAIU "unstable" here means, that *compilation* could break in minor releases, not that it's dangerous at runtime: https://github.com/tokio-rs/tokio/blob/1dcfe1cc9bb32fdc57ead3ff642d58bc8f282fd4/tokio/Cargo.toml#L114

Having it in prod builds is where I actually want to check it out. :)

elsirion (approved):

Maan2003:
`tracing` support in cln-plugin: https://github.com/ElementsProject/lightning/issues/5954

> I think @Maan2003 created a "log to disk"

log to disk has really bad perf in cloud environment. But yes adapting that should be possible (without file part).

dpc:
For `tokio-console` what's needed is initialization of another `tracing` subscriber layer that exposes `tokio=trace,runtime=trace` is kind of all that is needed at runtime.

PR #1621, opened 2023-02-08
Base for state machine client

elsirion:
This PR implements the interfaces for module state machines and implements an executor for these. There are a few TODOs left for after this design is validated, but seems to work in a [basic test](https://github.com/fedimint/fedimint/blob/f3c54150dd0684f7c769e08ad580327081e768c1/fedimint-client/src/sm/executor.rs#L411-L614) so far.

I can live with either keeping it a PR or merging, I think there shouldn't be too much trouble rebasing since I'm not touching the main code base too much. But feedback is definitely welcome at this point.

Builds on top of #1782 to make CI happy.

### DIFF
--- a/fedimint-client/Cargo.toml
+++ b/fedimint-client/Cargo.toml
@@ -13,9 +13,7 @@ path = "src/lib.rs"
 
 
 [dependencies]
-bitcoin = "0.29.2"
 fedimint-core  = { path = "../fedimint-core/" }
-thiserror = "1.0.30"
-
-[dev-dependencies]
-fedimint-core ={ path = "../fedimint-core/" }
+futures = "0.3.26"
+serde = "1.0.152"
+serde_json = "1.0.91"
--- a/fedimint-client/Cargo.toml
+++ b/fedimint-client/Cargo.toml
@@ -13,7 +13,15 @@ path = "src/lib.rs"
 
 
 [dependencies]
+anyhow = "1.0.69"
 fedimint-core  = { path = "../fedimint-core/" }
+fedimint-logging = { path = "../fedimint-logging" }
 futures = "0.3.26"
+rand = "0.8.5"
 serde = "1.0.152"
 serde_json = "1.0.91"
+tokio = { version = "1.25.0", features = [ "time" ] }
+tracing = "0.1.37"
+
+[dev-dependencies]
+tracing-test = "0.2.4"
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1,4 +1,19 @@
 //! Client library for fedimintd
 
+use fedimint_core::core::Decoder;
+
+use crate::sm::State;
+
 /// Client state machine interfaces and executor implementation
 pub mod sm;
+
+/// Fedimint module client
+pub trait ClientModule {
+    /// Data and API clients available to state machine transitions
+    type StateMachineContext: Clone;
+    /// All possible states this client can submit to the executor
+    type States: State<ModuleContext = Self::StateMachineContext>;
+
+    fn decoder(&self) -> Decoder;
+    fn context(&self) -> Self::StateMachineContext;
+}
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -9,11 +9,20 @@ pub mod sm;
 
 /// Fedimint module client
 pub trait ClientModule {
-    /// Data and API clients available to state machine transitions
-    type StateMachineContext: Clone;
+    /// Data and API clients available to state machine transitions of this
+    /// module
+    type ModuleStateMachineContext;
+
+    /// Data and API clients available to state machine transitions of all
+    /// modules
+    type GlobalStateMachineContext;
+
     /// All possible states this client can submit to the executor
-    type States: State<ModuleContext = Self::StateMachineContext>;
+    type States: State<
+        Self::GlobalStateMachineContext,
+        ModuleContext = Self::ModuleStateMachineContext,
+    >;
 
     fn decoder(&self) -> Decoder;
-    fn context(&self) -> Self::StateMachineContext;
+    fn context(&self) -> Self::ModuleStateMachineContext;
 }
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1,3 +1,4 @@
-//! This is a sketch/frame of future "modularized" Fedimint client library
-//!
-//! It's not being used by anything yet.
+//! Client library for fedimintd
+
+/// Client state machine interfaces and executor implementation
+pub mod sm;
--- a/fedimint-client/src/sm/executor.rs
+++ b/fedimint-client/src/sm/executor.rs
@@ -1,13 +1,16 @@
 use std::collections::BTreeMap;
+use std::fmt::Debug;
+use std::io::{Error, Read, Write};
+use std::marker::PhantomData;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
 use anyhow::bail;
 use fedimint_core::core::{IntoDynInstance, ModuleInstanceId};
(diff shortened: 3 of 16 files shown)


dpc:
I checked bfd0623 and it looks reasonable.

fedimint-client/src/sm/executor.rs
```diff
@@ -0,0 +1,614 @@
...
+        let mut executor_builder = Executor::builder();
+        executor_builder.with_module::<MockClientModule>(42, module);
+        let executor = executor_builder.build(tg, db.clone(), Arc::new(())).await;
+
+        info!("Initialized test executor");
+        (executor, broadcast, db)
+    }
+
+    #[tokio::test]
+    #[tracing_test::traced_test]
```
justinmoon, line 563: What does this do? Don't see any assertions about log output ...

fedimint-client/src/sm/executor.rs
```diff
@@ -0,0 +1,614 @@
...
+/// machines a different [execution context](super::state::Context) depending on
+/// the owning module, making it very flexible.
+pub struct Executor {
+    inner: Arc<ExecutorInner>,
+}
+
+struct ExecutorInner {
+    db: Database,
+    context: GlobalContext,
+    module_contexts: BTreeMap<ModuleInstanceId, DynContext>,
```
justinmoon, line 49: We use a map to allow the executor to access context of different module instance than it's own?

Specifically wondering how "making change" will work in a generalized way that doesn't depend on some specific module that can do this -- e.g. our ecash module at X git hash.

fedimint-client/src/sm/executor.rs
```diff
@@ -0,0 +1,614 @@
...
+        // TODO build await fn+timeout or allow manual driving of executor
+        tokio::time::sleep(Duration::from_secs(1)).await;
+        sender.send(0).unwrap();
+        tokio::time::sleep(Duration::from_secs(2)).await;
+
+        assert!(
+            executor
+                .contains_inactive_state(MOCK_INSTANCE_1, MockStateMachine::Final)
+                .await,
+            "State was written to DB and waits for broadcast"
```
justinmoon, line 611: Are the custom messages in `assert!` supposed to describe the success or failure of the assertion? These describe success, but I would expect the message to describe the failure since that's when it shows up in test logs ...

fedimint-client/Cargo.toml
```diff
@@ -13,7 +13,13 @@ path = "src/lib.rs"
...
+anyhow = "1.0.69"
 fedimint-core  = { path = "../fedimint-core/" }
+fedimint-logging = { path = "../fedimint-logging" }
 futures = "0.3.26"
+rand = "0.8.5"
 serde = "1.0.152"
 serde_json = "1.0.91"
+tokio = { version = "1.25.0", features = [ "time" ] }
+tracing = "0.1.37"
+tracing-test = "0.2.4"
```
justinmoon, line 25: Should this be a `dev-dependency`?

justinmoon (commented):
Super cool.

Left some nits and dumb questions :)

If anyone wants to play with this, try running `cargo test -p fedimint-client test_executor -- --nocapture` to just run the one executor test. One way to complete the state machine is to send `0` which is tested. There's another way to complete the state machine that isn't tested. Try to find it!

fedimint-client/src/sm/executor.rs
elsirion: I wanted to see log output, what did we use for tests elsewhere?

fedimint-client/src/sm/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+pub(super) mod state;
+
+use std::sync::Arc;
+
+pub use state::{Context, DynContext, DynState, State, StateTransition};
+
+// TODO: fill in useful data or make everything generic over a global context
+// type parameter
+/// Context that every state machine is supplied by its executor
+pub type GlobalContext = Arc<()>;
```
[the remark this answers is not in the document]
elsirion: Most importantly a handle that lets you create transactions, i.e. you give it inputs and outputs and it makes sure that it's sufficiently funded and creates change if necessary using one of the modules that is selected for such tasks.

So e.g. the LN module would create an LN input when claiming an incoming contract and then just give it to this handle. Unbeknownst to the LN module the top level client will use the mint client to create an e-cash output to balance the transaction and create the state machines necessary to submit it.

fedimint-client/src/sm/executor.rs
elsirion: The executor is an executor for all modules. Depending with which module's state machine it is dealing it will select the corresponding context. The executor also isn't a full client yet, just the core piece.

fedimint-client/src/sm/executor.rs
elsirion: Good question, opinions @dpc?

fedimint-client/src/lib.rs
```diff
@@ -1,3 +1,19 @@
...
+
+/// Client state machine interfaces and executor implementation
+pub mod sm;
+
+/// Fedimint module client
+pub trait ClientModule {
+    /// Data and API clients available to state machine transitions
+    type StateMachineContext: Clone;
+    /// All possible states this client can submit to the executor
+    type States: State<ModuleContext = Self::StateMachineContext>;
```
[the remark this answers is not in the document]
elsirion: I was flipping between `State` and `StateMachine` a lot ^^ technically the `States` associated type is supposed to be an enum over all possible states of a module. These states may belong to different state machines. So I could imagine the actual enum to look as follows:

```rust
enum MintStateMachines {
    InputStateMachine(InputStateMachine),
    OutputStateMachine(OutputStateMachine),
}

enum OutputStateMachine {
    State1(…),
    State2(…),
    …
}

enum InputStateMachine {
    State1(…),
    State2(…),
    …
}
```

fedimint-client/Cargo.toml
elsirion: Done

fedimint-client/src/sm/state.rs
```diff
@@ -6,17 +6,14 @@ use std::sync::Arc;
...
-    dyn_newtype_define_with_instance_id, dyn_newtype_impl_dyn_clone_passhthrough_with_instance_id,
-    module_dyn_newtype_impl_encode_decode, newtype_impl_eq_passthrough_with_instance_id,
-};
 use futures::future::BoxFuture;
 
 use crate::sm::{GlobalContext, OperationId};
 
 /// Implementors act as state machines that can be executed
-pub trait State:
+pub trait State<GC>:
```
elsirion, line 16: I'm not 100% sure if it should be an associated type instead, but I think this will work for now and we can always change it.

fedimint-client/src/sm/executor.rs
elsirion: That would mean the current version is ok? It describes why it shouldn't panic/what we assert.

fedimint-client/src/sm/executor.rs
justinmoon: I always just use `-- --nocapture`. This is good for now.

fedimint-client/src/lib.rs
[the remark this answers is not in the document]
justinmoon: i'll think about it. This should block this PR, though.

fedimint-client/src/sm/mod.rs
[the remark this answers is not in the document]
justinmoon: Ahh cool. That also answers my other question about how change will be made ...

fedimint-client/src/sm/executor.rs
dpc: I guess. Though I would consider doing it the longer way: "foobar shouldn't fail because the state blablablah".

dpc (approved):

justinmoon (approved):
LGTM

PR #1624, opened 2023-02-08
fix: include the given maximum amount in tier generation

mxz42:
Practically, that's likely not that relevant as larger values are given. 
Though, intuitively I'd expect that the given amount should be part of the tiers if an exact value is given. 

@jkitman thoughts?

### DIFF
--- a/fedimint-api/src/tiered.rs
+++ b/fedimint-api/src/tiered.rs
@@ -87,12 +87,12 @@ impl Tiered<SecretKeyShare> {
 }
 
 impl Tiered<()> {
-    /// Generates denominations as powers of 2 until a `max`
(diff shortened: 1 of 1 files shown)


dpc:
I guess I would expect inclusive so :+1:

jkitman:
Sure, that makes sense.  In practice, I think you'd want to set the max based on how much you'd expect the avg user to hold to keep the anonymity sets vs. speed trade-off good.

jkitman (approved):

elsirion (approved):

jkitman (approved):

dpc (approved):

PR #1623, opened 2023-02-08
Tests refactor

jkitman:
Refactoring of the tests to make things a bit more clear.

Some of the feedback from @dpc

PR #1625, opened 2023-02-09
fix: Fix test epoch handling

dpc:
* Make the module explicitily return proposals with a flag that indicate if it should trigger a new epoch, or not.
* Rename relevant test helper functions and add comments.
* Fix some bugs.

### DIFF
--- a/fedimint-api/src/module/mod.rs
+++ b/fedimint-api/src/module/mod.rs
@@ -440,6 +440,66 @@ where
     }
 }
 
+pub enum ConsensusProposal<CI> {
+    /// Trigger new epoch immediately including these consensus items
+    Trigger(Vec<CI>),
+    /// Contribute consensus items if other module triggers an epoch
+    // TODO: turn into `(Vec<CI>, Update)` where `Updates` is a future
+    // that will return a new `ConsensusProposal` when updates are available.
+    // This wake we can get rid of `await_consensus_proposal`
+    Contribute(Vec<CI>),
+}
+
+impl<CI> ConsensusProposal<CI> {
+    pub fn empty() -> Self {
+        ConsensusProposal::Contribute(vec![])
+    }
+
+    /// Trigger new epoch if contains any elements, otherwise contribute nothing.
+    pub fn new_auto_trigger(ci: Vec<CI>) -> Self {
+        if ci.is_empty() {
+            Self::Contribute(vec![])
+        } else {
+            Self::Trigger(ci)
+        }
+    }
+
+    pub fn map<F, CIO>(self, f: F) -> ConsensusProposal<CIO>
+    where
+        F: FnMut(CI) -> CIO,
+    {
+        match self {
+            ConsensusProposal::Trigger(items) => {
+                ConsensusProposal::Trigger(items.into_iter().map(f).collect())
+            }
+            ConsensusProposal::Contribute(items) => {
+                ConsensusProposal::Contribute(items.into_iter().map(f).collect())
+            }
+        }
+    }
+
+    pub fn forces_new_epoch(&self) -> bool {
+        match self {
+            ConsensusProposal::Trigger(_) => true,
+            ConsensusProposal::Contribute(_) => false,
+        }
+    }
+
+    pub fn items(&self) -> &[CI] {
+        match self {
+            ConsensusProposal::Trigger(items) => items,
+            ConsensusProposal::Contribute(items) => items,
+        }
+    }
+
+    pub fn into_items(self) -> Vec<CI> {
+        match self {
+            ConsensusProposal::Trigger(items) => items,
+            ConsensusProposal::Contribute(items) => items,
+        }
+    }
+}
+
 #[async_trait]
 pub trait ServerModule: Debug + Sized {
     type Gen: ModuleGen;
@@ -460,7 +520,7 @@ pub trait ServerModule: Debug + Sized {
     async fn consensus_proposal<'a>(
         &'a self,
         dbtx: &mut DatabaseTransaction<'_>,
-    ) -> Vec<<Self::Decoder as Decoder>::ConsensusItem>;
+    ) -> ConsensusProposal<<Self::Decoder as Decoder>::ConsensusItem>;
 
     /// This function is called once before transaction processing starts. All module consensus
     /// items of this round are supplied as `consensus_items`. The database transaction will be
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -56,10 +56,10 @@ use fedimint_testing::btc::{fixtures::FakeBitcoinTest, BitcoinTest};
 use fedimint_wallet::config::WalletConfig;
 use fedimint_wallet::db::UTXOKey;
 use fedimint_wallet::Wallet;
-use fedimint_wallet::WalletConsensusItem;
 use fedimint_wallet::{SpendableUTXO, WalletGen};
 use futures::executor::block_on;
 use futures::future::{join_all, select_all};
+use futures::FutureExt;
 use hbbft::honey_badger::Batch;
 use itertools::Itertools;
 use lightning_invoice::Invoice;
@@ -728,6 +728,8 @@ impl FederationTest {
             let proposal = ConsensusProposal {
                 items,
                 drop_peers: vec![],
+                // if we force it, we want to trigger an epoch
+                force_new_epoch: true,
             };
 
             server.borrow_mut().override_proposal = Some(proposal.clone());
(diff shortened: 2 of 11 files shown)


dpc:
Rebased on top of #1623 . I think we need both.

fedimint-api/src/module/mod.rs
```diff
@@ -440,6 +440,68 @@ where
     }
 }
 
+pub enum ConsensusProposal<CI> {
```
jkitman, line 443: Maybe `Force` -> `TriggerNewEpoch` and `Lazy` -> `ContributeToEpoch`

dpc: I'm not into neither current ones, nor these ones, but the current ones are shorter. :D

Let's get more opinions and I'm fine with whatever.

fedimint-api/src/module/mod.rs
```diff
@@ -440,6 +440,68 @@ where
...
+    Force(Vec<CI>),
+    /// Do not trigger new epoch, but if other module does, use these
+    /// consensus items.
+    // TODO: turn into `(Vec<CI>, Update)` where `Updates` is a future
+    // that will return a new `ConsensusProposal` when updates are available.
+    // This wake we can get rid of `await_consensus_proposal`
+    Lazy(Vec<CI>),
+}
+
+impl<CI> From<Vec<CI>> for ConsensusProposal<CI> {
```
jkitman, line 454: Wonder if it is safer to force modules to implement this themselves to avoid accidentally always triggering epochs.

fedimint-api/src/module/mod.rs
elsirion: `TriggerEpoch` and `Contribute`?

fedimint-api/src/module/mod.rs
elsirion: I don't like hiding this kind of logic in a `From`/`into`. Should be more explicit like `new_auto_lazy(…)`

integrationtests/tests/fixtures/mod.rs
```diff
@@ -973,22 +974,26 @@ impl FederationTest {
...
+    ///
+    /// Wallet chain height can trigger epoch randomly, but since
+    /// they will be processed along-side any other epochs you
+    /// already expect, nothing is being done about it. It should
+    /// not matter, but is worth pointing out.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there's an empty proposal.
     #[allow(clippy::await_holding_refcell_ref)]
```
elsirion, line 990: We don't need this anymore, right?

dpc: https://github.com/fedimint/fedimint/pull/1599 fixes it

fedimint-api/src/module/mod.rs
dpc: You're right. It did feel fishy. Done.

fedimint-api/src/module/mod.rs
dpc: I went with `Trigger` and `Contributed`. Because now we can say that epochs get:

image

dpc:
@elsirion @jkitman Not 100% stable, but from my tests in CI pipeline, much, much better.

fedimint-api/src/module/mod.rs
jkitman: How about `Force` -> `Trigger` at least?  That is consistent with how we refer to triggering epochs elsewhere in the code.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1020,60 +1038,59 @@ impl FederationTest {
...
+            // Pending transactions will trigger an epoch
+            if let Ok(tx) = fedimint_server.tx_receiver.get_mut().as_mut().try_recv() {
+                fedimint_server
+                    .consensus
+                    .tx_sender
+                    .send(tx)
+                    .await
+                    .expect("Can send");
+                return true;
+            }
```
elsirion, line 1057: `fedimint_server.tx_receiver` is a `Peekable` to avoid exactly this hack. You should be able to avoid re-sending and potentially re-ordering transactions this way:

```rust
if let Some(Some(_)) fedimint_server.tx_receiver.peek().now_or_never() {
    return true;
}
```

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1020,60 +1038,59 @@ impl FederationTest {
...
+
+            let consensus_proposal = fedimint_server.consensus.get_consensus_proposal().await;
+            if !consensus_proposal.force_new_epoch {
+                continue;
+            }
+
+            if !consensus_proposal.items.is_empty() {
+                return true;
+            }
```
elsirion, line 1066: Is this still correct? Shouldn't there just always be an epoch if `force_new_epoch == true`? It will never be a problem currently because the wallet always contributes opportunistically, but if that ever changes this might lead to some annoying debugging (e.g. rejoin-epochs might be empty but will be forced).

modules/fedimint-dummy/src/lib.rs
```diff
@@ -211,8 +212,8 @@ impl ServerModule for Dummy {
     async fn consensus_proposal(
         &self,
         _dbtx: &mut DatabaseTransaction<'_>,
-    ) -> Vec<DummyConsensusItem> {
-        vec![]
+    ) -> ConsensusProposal<DummyConsensusItem> {
+        ConsensusProposal::empty()
```
elsirion, line 216: nit: making `await_consensus_proposal` dependent on `consensus_proposal` in the dummy module would show developers using the dummy module as a reference how it should be done.

elsirion (commented):
Great progress!

fedimint-api/src/module/mod.rs
dpc: That's what it is now, no? `Trigger` and `Contribute`.

integrationtests/tests/fixtures/mod.rs
dpc: Agreed. Fixed.

integrationtests/tests/fixtures/mod.rs
dpc: Done.

elsirion (approved):

integrationtests/tests/fixtures/mod.rs
```diff
@@ -973,22 +975,26 @@ impl FederationTest {
         }
     }
 
-    /// Has every federation node send new consensus proposals then process the outcome.
-    /// If the epoch has empty proposals (no new information) then panic
+    /// Runs `n` epochs in the federation (each guardian node)
+    ///
+    /// Call in tests when new epoch conditions are already
+    /// in place, to trigger new epoch generation.
```
mxz42, line 981: maybe clarify  a bit to : 
```
Call this method in tests if some conditions that trigger a new epoch already exist
```

integrationtests/tests/fixtures/mod.rs
```diff
@@ -998,13 +1004,26 @@ impl FederationTest {
...
+    ///
     /// If proposals are empty you will need to run a concurrent task that triggers a new epoch or
-    /// it will wait forever
-    pub async fn await_consensus_epochs(&self, epochs: usize) -> anyhow::Result<()> {
+    /// it will wait forever.
+    ///
+    /// When conditions triggering proposals are already in place, calling this functions has
+    /// the same effect as calling [`run_consensus_epochs`], as blocking conditions can't
+    /// happen. However in that situation calling [`run_consensus_epochs`] is preferable, as
+    /// it will snappily panic, instead of handing in case of bugs.
```
mxz42, line 1018: I don't understand the above sentence. Do you mean something along the lines of
```
 However,  in that situation calling [`run_consensus_epochs`] is preferred because it will panic instead of (waiting|handling) in case of bugs.
```
Not sure what you mean by *handing*?

justinmoon (approved):

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1020,60 +1039,52 @@ impl FederationTest {
...
-                    }
-                    _ => return false,
-                },
-                ConsensusItem::EpochOutcomeSignatureShare(_) => continue,
-                _ => return false,
+    /// Get the consensus items proposed by all the peers
+    ///
+    /// Notably, unlike [`has_pending_epoch`] this does not take into account
+    /// pending transactions or ignore consensus items that do not trigger
+    /// epoch on its own.
```
mxz42, line 1069: confusing language. 
So, this method returns all proposals including pending transactions(?) and ConsensusItems that do not trigger an epoch on their own?

mxz42 (commented):
only a  couple things in the comment  that I didn't quite understand. 

I don't have an opinion on `Trigger` and `Contribute` or if it should be called `Force` or similar. The doc comments above them state what they do in a clear way.  
If we use the notation of "forcing a epoch" in many places, we may want to go for that as @jkitman stated but I didn't look in the code base.

integrationtests/tests/fixtures/mod.rs
dpc: https://github.com/fedimint/fedimint/pull/1634

PR #1626, opened 2023-02-09
Verifies the client config in the API

jkitman:
Creates a generic `VerifySigned` query strategy that can be used to validate the client configs are signed.

Next step towards verifiable client config downloads from any number of URLs using the `WsFederationConnect` JSON.

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -375,6 +375,12 @@ async fn main() {
             );
         };
     } else {
+        let module_gens = ModuleGenRegistry::from(vec![
+            DynModuleGen::from(WalletGen),
+            DynModuleGen::from(MintGen),
+            DynModuleGen::from(LightningGen),
+        ]);
+
         let cli = Cli::parse();
         if let Command::JoinFederation { connect } = cli.command {
             let connect_obj: WsFederationConnect = serde_json::from_str(&connect)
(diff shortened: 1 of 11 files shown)


client/client-lib/src/query.rs
```diff
@@ -23,30 +22,54 @@ impl<R> QueryStrategy<R> for TrustAllPeers {
...
+    /// * `pk`: Public key used to verify the response
+    /// * `required`: How many responses until a failure or success is returned
+    /// * `allow_consensus_fallback`: Returns a success if cannot verify but `required` agree
+    /// * `verifier`: Function that verifies the data with the public key
+    pub fn new(
+        pk: PublicKey,
+        required: usize,
+        allow_consensus_fallback: bool,
+        verifier: impl Fn(&PublicKey, &R) -> bool + Send + Sync + 'static,
+    ) -> Self {
```
elsirion, line 46: nit: `pk` could just be captured by the `verifier` closure, one less arg.

client/cli/src/main.rs
```diff
@@ -375,20 +375,25 @@ async fn main() {
             );
         };
     } else {
+        let module_gens = ModuleGenRegistry::from(vec![
+            DynModuleGen::from(WalletGen),
+            DynModuleGen::from(MintGen),
+            DynModuleGen::from(LightningGen),
+        ]);
```
elsirion, line 382: A bit off topic, but couldn't we generate the `decoders` further down from these `module_gens`?

elsirion (dismissed):
Looks good! One little nit, but that's very much personal preference I think.

client/cli/src/main.rs
jkitman: They are declared further down, or do you mean right after?

client/client-lib/src/query.rs
jkitman: Excellent point, this would allow for other kinds of verifiers as well.

client/cli/src/main.rs
elsirion: Instead of declaring them they could be auto-generated I assume?

elsirion (approved):

dpc (approved):

PR #1637, opened 2023-02-10
refactor: `RealBitcoinTest` locking

dpc:
Please https://github.com/fedimint/fedimint/pull/1599 first.

Instead of having bunch of ad-hoc `xyz_no_lock`, make them a separate lowest-level (locking-oblivious) `struct`, and let the higher level `struct`s wrap it.

### DIFF
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -854,26 +854,24 @@ impl FederationTest {
             .unwrap();
 
         for server in &self.servers {
-            block_on(async {
-                let svr = server.lock().await;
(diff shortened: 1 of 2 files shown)


jkitman (dismissed):
Is there now some reasoning we can use to determine when a test needs to acquire a lock?

dpc:
> Is there now some reasoning we can use to determine when a test needs to acquire a lock?

So far it's been: "we need to test the *lack* of pending epochs, so we can't tolerate other tests changing the blockchain height".

jkitman (approved):

elsirion (approved):

Issue #1379, opened 2023-01-17
CI currently isn't running shell scripts


jkitman:
@dpc Not sure what went wrong, but seems for at least a week our GH actions aren't running the shell script based tests.  You can verify this by looking at the logs or modifying the scripts to fail.

dpc:
Thanks for spotting it! I've got a fix.

jkitman:
Now we are uncovering the silent errors :smile: 

`WIRE_UNKNOWN_NEXT_PEER` error is one I've been seeing in the cli-tests, maybe @okjodom can help fix?

dpc:
He's on it. Pinged him on Discord.

justinmoon:
This happens in tmuxinator when you create an invoice with `fedimint-cli` and pay it with `ln2`. 

![image](https://user-images.githubusercontent.com/4335621/213028032-884b0a60-4b10-41de-8cff-ea29f1ad8649.png)

This hack fixes it

![image](https://user-images.githubusercontent.com/4335621/213028351-3222a59a-4151-4649-9c9c-82ebd55baa09.png)

It seems that the `htlc_accepted` message we get from CLN changed. I experienced this in another deployment and was baffled because I didn't think I had updated CLN.

Did Nix update our CLN recently? I wonder if our change to short channel id's caused CLN to give us HTLCs in a different format?

okjodom:
@justinmoon we last bumped to `cln-plugin` last month: See #1049 

If it's the case that [v0.1.1](https://crates.io/crates/cln-plugin/0.1.1) already included the breaking changes to `htlc_accepted`, then we'd need to redefine `HtlcAccepted` struct to this:

```
// TODO: Keep in sync with cln-plugin:
// See: https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md#htlc_accepted
#[derive(Clone, Serialize, Deserialize, Debug)]
pub struct Htlc {
    pub short_channel_id: u64,
    #[serde(deserialize_with = "as_fedimint_amount")]
    pub amount_msat: Amount,
    pub cltv_expiry: u32,
    pub cltv_expiry_relative: u32,
    pub payment_hash: bitcoin_hashes::sha256::Hash,
}

#[derive(Clone, Serialize, Deserialize, Debug)]
pub struct Onion {
    pub payload: String,
    pub short_channel_id: u64,
    #[serde(deserialize_with = "as_fedimint_amount")]
    pub forward_msat: Amount,
[...]

PR #1639, opened 2023-02-10
fix: code coverage

### LINKED ISSUE #1622: Code coverage isn't being uploaded
At some point we stopped uploading our test coverage to codecov.io. It appears that the output directory of `nix build -L .#debug.workspaceCov` doesn't have a `lcov.info` file anymore and it also can't be found in `target.tar.zst` which is the only file present. Looking at the logs I'm not even sure coverage tests are run. Could this be another silent breakage from `crane` changing its API similar to #1379 @dpc?

dpc:
At some point the code coverage was aftected by the same issue that affected tests in #1379 .

Fix #1622

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -385,25 +385,22 @@
 
         # Build only deps, but with llvm-cov so `workspaceCov` can reuse them cached
         workspaceDepsCov = craneLib.buildDepsOnly (commonArgs // {
-          pname = commonArgs.pname + "-lcov";
+          pnameSuffix = "-lcov-deps";
           src = filterWorkspaceDepsBuildFiles ./.;
-          cargoBuildCommand = "cargo llvm-cov --workspace --profile $CARGO_PROFILE";
+          buildPhaseCargoCommand = "cargo llvm-cov --workspace --profile $CARGO_PROFILE --no-report";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -385,27 +385,24 @@
...
+          buildPhaseCargoCommand = "cargo llvm-cov --workspace --profile $CARGO_PROFILE --no-report";
+          cargoBuildCommand = "dontuse";
+          cargoCheckCommand = "dontuse";
           nativeBuildInputs = commonArgs.nativeBuildInputs ++ [ cargo-llvm-cov ];
           doCheck = false;
         });
 
-        workspaceCov = craneLib.cargoBuild (commonArgs // {
-          pname = commonArgs.pname + "-lcov";
+        workspaceCov = craneLib.buildPackage (commonArgs // {
```
elsirion, line 397: I got the "use `workspaceCov` instead of `workspaceTestCov`" part when working on it myself, but didn't know the difference between `cargoBuild` and `buildPackage` (or even considered it).

elsirion (approved):
Thx a lot! I couldn't figure it out myself yesterday.

jkitman (approved):

PR #1640, opened 2023-02-10
feat: share dependency builds between different packages

dpc:
By compiling multiple packages together we can re-use their
dependency builds.

Unfortunately this can't be done just "one for all" as some
targets can compile only a subset of packages (wasm), etc.
so we need to be mindful about grouping them.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -442,12 +442,18 @@
         });
 
 
-        pkg = { name, dirs, defaultBin ? null }:
+        # Compile a group of packages together
+        #
+        # This unifies their cargo features and avoids building common dependencies mulitple
+        # times, but will produce a derivation with all listed packages.
+        pkgsBuild = { name, pkgs, dirs, defaultBin ? null }:
           let
+            # "--package x --package y" args passed to cargo
+            pkgsArgs = lib.strings.concatStringsSep " " (lib.mapAttrsToList (name: value: "--package ${name}") pkgs);
             deps = craneLib.buildDepsOnly (commonArgs // {
               src = filterWorkspaceDepsBuildFiles ./.;
-              pname = "pkg-${name}-deps";
-              buildPhaseCargoCommand = "cargo build --profile $CARGO_PROFILE --package ${name}";
+              pname = name;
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -471,8 +471,15 @@
             craneLib = craneLibCross.${target.attr};
             deps = craneLib.buildDepsOnly (commonArgs // {
               src = filterWorkspaceDepsBuildFiles ./.;
-              pname = "pkg-${name}-${target.attr}-deps";
-              buildPhaseCargoCommand = "cargo build --profile $CARGO_PROFILE --target ${target.name} --package ${name}";
+              pname = "fedimint-packages-${target.attr}";
+              # workaround: on wasm, we can't compile all deps, so narrow dependency build
+              # to ones used by the client package only
+              buildPhaseCargoCommand =
+                if target.name == "wasm32-unknown-unknown" then
```
jkitman, line 478: where does `wasm32-unknown-unknown` come from?

jkitman (dismissed):

flake.nix
dpc: From a cross-toolchain defintion somewhere in the file. But that's the "triple" target used for wasm. Just like `x86_64` is a common name for intel cpu architecture.

dpc:
Doesn't look like it is working. The dependency builds are not being re-used.

dpc:
OK, I've changed the approach to building in groups and that speeds things up.

flake.nix
```diff
@@ -499,11 +513,19 @@
...
 
-        fedimintd = pkg {
-          name = "fedimintd";
+        fedimint-pkgs = pkgsBuild {
+          name = "fedimint-pkgs";
+
+          pkgs = {
+            fedimintd = { };
+            fedimint-cli = { };
+            fedimint-tests = { };
```
elsirion, line 522: We might want to avoid test-only features being active in release builds eventually to reduce attack surface. But ok for now imo.

elsirion (approved):

jkitman (approved):

PR #1641, opened 2023-02-10
refactor: rename variable in isolated database transaction

mxz42:
very small nit; thought I submit it anyway.

just came across when reading about `IsolatedDatabaseTransaction`s. Currently, `new` is only called with the `module_instance_id` as param. If that will be the main value for the `new` function, I think it's worth to rename that parameter. If it will be called with other prefixes in the future (not only module instance id), then just close this PR @m1sterc001guy


m1sterc001guy (approved):
Rename is fine with me

jkitman (approved):
Are there other instances that should be renamed?

dpc (approved):

mxz42:
> Are there other instances that should be renamed?

hm, maybe. I just came across these today and renamed them to follow the `IThing` and `DynThing` pattern

PR #1642, opened 2023-02-10
feat: add `just` command to execute real tests (no fixtures)

mxz42:
sometimes I just want to run the real tests without firing up a bash shell, load the setup script... 

I use the `bash` wrapper as my default shell is `fish`. Should be fine? 

thoughts? 

output is something like this:
```
$  just test-real
bash -c 'source ./scripts/setup-tests.sh >/dev/null && export FM_TEST_DISABLE_MOCKS=1 && export RUST_BACKTRACE=1 && export RUST_LOG=none && cargo test -p fedimint-tests '
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
error: timeout on transient error: Could not connect to the server 127.0.0.1:18443

Make sure the bitcoind server is running and that you are connecting to the correct RPC port.
LN gateway 1 not ready yet. Waiting ...
    Finished test [unoptimized + debuginfo] target(s) in 0.14s
     Running tests/tests.rs (target/debug/deps/fedimint_tests-77077150af491434)

running 31 tests
test audit_negative_balance_sheet_panics - should panic ... ok
test can_have_federations_with_one_peer ... ok
test ecash_in_wallet_can_sent_through_a_tx ... ok
test lightning_gateway_cannot_claim_invalid_preimage ... ok
test ecash_cannot_double_spent_with_different_nodes ... ok
test can_get_signed_epoch_history ... ok
test lightning_gateway_pays_outgoing_invoice ... ok
[...]
```


mxz42:
> Originally, the command to startup the tests is documented as cargo test -p fedimint-tests -- --test-threads=1.
With the parallelization it seem to work now without --test-threads=1?

oh, and if that's true I might also update the [integration test documentation here](https://github.com/fedimint/fedimint/tree/master/integrationtests#running-with-real-services)?

dpc:
@mxz42 We've just parallelized these tests this week. `./scripts/rust-tests.sh` is the current command we use, so we should avoid duplicating that call IMO. Either everything in justfile, or justfile should call that script.

mxz42:
oh, wow, I honestly didn't know this script existed because I always looked [at this description](https://github.com/fedimint/fedimint/tree/master/integrationtests#running-with-real-services)

dpc:
> oh, wow, I honestly didn't know this script existed because I always looked [at this description](https://github.com/fedimint/fedimint/tree/master/integrationtests#running-with-real-services)

It's hard to keep documentation up to date. Hopefully putting everything useful in a `justfile` will help discoverability.

scripts/rust-tests.sh
```diff
@@ -1,10 +1,18 @@
...
 # Runs the all the Rust integration tests
 
-set -euxo pipefail
+set -euo pipefail
 export RUST_LOG="${RUST_LOG:-info}"
 
-source ./scripts/setup-tests.sh ""
+# Convert RUST_LOG to lowercase
+# if RUST_LOG is none, don't show output of test setup
+if [ "${RUST_LOG,,}" = "none" ]; then
```
dpc, line 9: Why convert it to lowercase? Why would would anyone ... `RUST_LOG=nOnE`?

mxz42: I've seen people do all uppercase, so `RUST_LOG=NONE`.

dpc (approved):

scripts/rust-tests.sh
mxz42: I agree :D especially because you have to hold the shift key while typing (nobody is using capslock hopefully :sweat_smile: )

justfile
```diff
@@ -10,6 +10,9 @@ build:
 test: build
   cargo test
 
+test-real:
+  ./scripts/rust-tests.sh
```
justinmoon, line 14: nit: I think the script and the just command should have the same name

justinmoon (approved):

PR #1643, opened 2023-02-10
202302/rename to `Dyn`s

mxz42:
Adding the `Dyn` prefix to 2 structs

### DIFF
--- a/fedimint-api/src/config.rs
+++ b/fedimint-api/src/config.rs
@@ -35,7 +35,7 @@ use threshold_crypto::Signature;
 use url::Url;
 
 use crate::module::DynModuleGen;
-use crate::net::peers::MuxPeerConnections;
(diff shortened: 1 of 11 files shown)


dpc:
Do we want `Dyn` on everything? We needed it most on module types, as there we wanted the shortest name to go to 3rd party module writers.

Here we can use `Foo` + `IFoo` and the fact that it's `Dyn` is not something I need reminder in name, IMO.

mxz42:
hm, I just came across these today and thought that may fight the pattern.

valid argument, though. 

thoughts? @elsirion @jkitman

elsirion:
Imo consistency would be nice and seeing on first glance that it's a mockable interface would be good too, but I see your point of longer names @dpc (and would add that it is also more churn).

mxz42:
closing this in favor of the more general issue #2055

PR #1646, opened 2023-02-10
fix: add execute permission to script

mxz42:


dpc (approved):

elsirion (approved):
I believe this is only a partial fix to #1650. In a pure nix environment it would still fail.

Issue #1356, opened 2023-01-16
Move `hbbft` dependency from `fedimint-api` to `fedimint-server`


dpc:
We need to minimize dependencies on `fedimint-api` as everything is using it.

thessc1:
Seems like a good first issue. Happy to tackle this one and #295. Any other comments?

justinmoon:
It's all yours @thessc1!

thessc1:
Is the ideal state for each module to not have any dependencies on the others? I can do a review to see if there are any others along with hbbft.

elsirion:
> Is the ideal state for each module to not have any dependencies on the others?

It's a bit more nuanced. We did have some extended discussions a while ago from which this ideal structure emerged: https://github.com/fedimint/fedimint/blob/master/docs/modular-architecture.md

It's probably not a good task for someone new to the project since making the right decisions requires a lot of (sometimes historical) context.

What would be easier and also desired is looking into 3rd party dependencies that are only depended on by one or two crates (you can use `cargo tree` and `cargo tree -i package`) and figuring out if they can be removed by turning of unnecessary default features of dependencies.

thessc1:
Alright, I'll stick to the 3rd party dependencies + hbbft for this one. Thanks!

thessc1:
Question here, are we trying to remove nested dependencies, e.g. bitcoin is an explicit dependency, but also nested in lightning and miniscript? So, in this case would we want to turn off the defaults for lightning and miniscript and just use the explicit dependency?

Or alternatively, there's a dependency within most of these dependencies called proc-macro2. Do we want to remove that nested dependency and either make it explicit or only leave it within 1 dependency?

dpc:
@thessc1 We should re-export the most common 3rd party crates that are part of the `fedimint-api`, and use these re-exports in downstream crates (everything else). `bitcoin`, `secp256k1`, `serde_json` - anything that is `pub` in `fedimint-api`.

thessc1:
Just for clarification, we want to re-export the most commonly used 3rd party ones, e.g. maybe used in 3+ places. Anything else that is needed, but only used in a few places, we leave nested. And then anything that is part of a default, but isn't used at all, we just eliminate?

dpc:
I think we can sort out the details of minor deps later. But the major ones that are used all over the public apis, yes.

thessc1:
Cool, I'll pull out the major one's then.

dpc:
[...]

PR #1648, opened 2023-02-11
Re: #1356 reexport 3rd party crates in fedimint-api

thessc1:
#1356 

If this is on the right track, I can do the same exercise for the rest of the modules.


dpc:
LGTM, but please use informative PR titles and commits.

Also please don't use `git merge`. Rebase your change instead with `git fetch upstream && git rebase upstream/master`.

PR #1651, opened 2023-02-11
fix: add correct shebang for nix environments

### LINKED ISSUE #1650: `restart` command in `tmuxinator` doesn't work
Two problems:
* The file is not executable, setting `chmod +x` fixes that
* The interpreter of `scripts/restart-tmux.sh` (`/bin/bash`) specified in the first line doesn't exist in the nix env, needs to be `#!/usr/bin/env bash`. The same goes for `scripts/gw-reload.sh` which is also called as part of `restart`.

mxz42:
fixes #1650

script permission was added in #1646

### DIFF
--- a/scripts/always-fail-test.sh
+++ b/scripts/always-fail-test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # CI to check if this actually failed
 exit 1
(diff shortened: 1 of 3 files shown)


dpc (approved):
This is the way

justinmoon (approved):

PR #1652, opened 2023-02-11
feat: spell-checking via `cargo spellcheck`

dpc:
It's now possible to run `just spell` to find spelling mistakes only in files modified
by us. Much more reasonable then trying to fix the whole codebase. But if you want to try
`just spell-fix-all` is still there.

**Previous message**:
I got curious and wanted to try it out, but I'm not sure if it's worth following through.

First, `cargo spellcheck` seem rather buggy and slow.

Second, it's a lot of effort. Hunspell doesn't know *a lot* of the terms we're using.

I rather get some feedback before investing any more time in it.


justinmoon:
Seems like a good idea. We have had lots of typos in our codebase. This seemed to catch a lot of them. Perhaps someone else can take this over if there's a bunch of manual work remaining?

dpc:
We should discuss on dev call, but maybe in next, next one as I'm going miss one.

elsirion:
I don't think this is urgent, but when we find the time would be nice :)

elsirion (approved):

mxz42:
I think that would certainly be nice. Currently, I have a manual list where I add spelling mistakes I come across in the code base. Wanted to submit a PR at some point fixing only typos :'D

dpc:
@mxz42 approve & merge then and try it out! :)

justinmoon (approved):

PR #1654, opened 2023-02-11
Client connect string

### LINKED ISSUE #614: Best way to serialize "connect info"?
Currently we use the following JSON encoding to allow clients to "join a federation":

```
{"members":[[0,"wss://fm-signet.sirion.io:443"]],"max_evil":0}
```

This basically just JSON-encodes the parameters to `WsFederationApi::new()`. Once a client has a `WsFederationApi`, it can download configs from all federation servers?

What is the optimal way to encode this information?

- `max_evil` can be derived from the peer count via `hbbft::util::max_faulty()`, so that really doesn't need to be included. (could be removed from `WsFederationApi::new()` as well ...
- The `0` in `[0,"wss://fm-signet.sirion.io:443"]` is just a "peer id" which can also be implied by order of the array.
- So really all we need is the websocket URLs ...

Might be easier to copy/paste if we bech32m-encoded it? Maybe the error-correcting would help, too? But then it's harder to inspect it visually...

### LINKED ISSUE #1205: New stable threshold key/id for authenticating configuration

jkitman:
Refactors the client connection string to use `one_honest` number of peers.  Now that we verify against the `federation_id` we can use a smaller number of urls or even non-federation urls to download from, and the urls don't need to be trusted.

JSON still seems like the best format since it's easy to construct and read, I tried other formats and they don't compress the information much.  Bytes for 10 guardians goes down from `649` to `242` with this change.

Example config:
```
{
  "urls": [
    "ws://127.0.0.1:18174/",
    "ws://127.0.0.1:18184/",
  ],
  "id": "ab011cbb464a4b71d90bc1f08a9034e43365a3620a72b76fd06e07d90221c5510d8a7489e78d4375114aae30ff3250b4"
}
```

Fixes #614
Will also close #1205 as most of the stable fed id work is done, will open new issue for updating P2P federation connections.

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -19,7 +19,7 @@ use fedimint_api::module::DynModuleGen;
 use fedimint_api::task::TaskGroup;
 use fedimint_api::{Amount, OutPoint, TieredMulti, TransactionId};
 use fedimint_core::api::{
-    FederationApiExt, GlobalFederationApi, IFederationApi, WsFederationApi, WsFederationConnect,
+    FederationApiExt, GlobalFederationApi, IFederationApi, WsClientConnectInfo, WsFederationApi,
 };
 use fedimint_core::config::load_from_file;
 use fedimint_core::query::EventuallyConsistent;
@@ -108,7 +108,7 @@ enum CliOutput {
     },
 
     ConnectInfo {
-        connect_info: WsFederationConnect,
+        connect_info: WsClientConnectInfo,
     },
 
     JoinFederation {
@@ -383,9 +383,9 @@ async fn main() {
 
         let cli = Cli::parse();
         if let Command::JoinFederation { connect } = cli.command {
-            let connect_obj: WsFederationConnect = serde_json::from_str(&connect)
+            let connect_obj: WsClientConnectInfo = serde_json::from_str(&connect)
                 .or_terminate(CliErrorKind::InvalidValue, "invalid connect info");
-            let api = Arc::new(WsFederationApi::new(connect_obj.members))
+            let api = Arc::new(WsFederationApi::from_urls(&connect_obj))
                 as Arc<dyn IFederationApi + Send + Sync + 'static>;
             let cfg: ClientConfig = api
                 .download_client_config(&connect_obj.id, module_gens.clone())
@@ -625,7 +625,7 @@ async fn handle_command(
             )
         }
         Command::ConnectInfo => {
-            let info = WsFederationConnect::from(client.config().as_ref());
+            let info = WsClientConnectInfo::from(client.config().as_ref());
             Ok(CliOutput::ConnectInfo {
                 connect_info: (info),
             })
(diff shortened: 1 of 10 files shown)


elsirion (approved):

elsirion:
As a separate matter, now that we have the federation id in there taking up most of the bytes it might make sense to explore bech32 encoding of the whole string. This would get rid of json special characters and be more compact for binary data too. If bech32 is serialized in upper case it only uses 5.5 bit per character (so only an overhead of 10% for binary data compared to 100% for hex encoding+special characters forcing binary encoding).

jkitman:
@elsirion we could support that for QR codes or copy+pasting.

dpc:
BTW. I'm not sure if there's a point of having it human readable. Sure it's appealing for developers and SWE like us, but if it was just consensus-encoded, then base64 or something, then it would be shorter and easier to copy&paste without mistakes. Some short checksum to detect truncation and other copy&paste mistakes would be OK.

The client UI should be able to display URLs to the user if it's desired. Or if it can't connect it can display human readable errors, etc.

jkitman:
@dpc I'd prefer to merge this change as-is then do the serialization in a separate PR.

But are you suggesting that even in the CLI / stored on the filesystem we use bech32 encoding?  Right now those seem like developer tools.

The bech32 encoding would be useful for storing/displaying in a UI for non-technical users to copy.

justinmoon (approved):

dpc:
@jkitman I don't understand what's the point of bech32. It seems useful for bitcoin addresses that are (somewhat) short and can be spelled over a phone or wrote done on a piece of paper and error detection and correction are useful. We should optimize for short encoding, with some minimal sanity checksum that will detect copy&paste fails. 

I would like to use the same some base56 or whatever that is best for copy&paste all over the UI, where copy&paste is involved. The only criteria is ease of copy&paste: no weird characters that can make software consider the string as separate words, and the shorter the better.

elsirion:
@dpc Bech32 has the nice property (if all uppercase) that it fits into the alphanumeric charset of QR codes. This means each bech32 character takes 5.5 bits to encode in a QR code while itself encoding 5bit, so wasting merely 10% for binary data while still being copyable. Compare that to e.g. base64 encoding where each character encodes 6bit but takes 8bit to encode in a QR code due to the larger charset, that's an overhead of 50% for binary data (base58 even worse).

PR #1655, opened 2023-02-11
Refactor aead encryption

### LINKED ISSUE #974: Refactor symmetric encryption helpers into a shared dependency
I think we could refactor the encryption helpers into a shared dependency.

_Originally posted by @jkitman in https://github.com/fedimint/fedimint/pull/928#discussion_r1031601017_

jkitman:
Finish moving encryption helpers to the aead crate, trying to keep `fedimintd` as small a crate as possible.

Fixes  #974


dpc (approved):

elsirion (approved):

crypto/aead/Cargo.toml
```diff
@@ -12,5 +12,7 @@ path = "src/lib.rs"
 
 [dependencies]
 anyhow = "1.0.65"
+rpassword = "7.2.0"
 ring = "0.16.20"
+hex = "0.4.3"
```
mxz42, line 17: instead of `hex` we should use `bitcoin_hashes::hex`  and the `ToHex` trait. currently, upstream rust-bitcoin is redoing their hashing anyway, so as soon as [this PR for the new hex crate lands](https://github.com/rust-bitcoin/rust-bitcoin/pull/1573) we can depend on that.

PR #1656, opened 2023-02-11
[DONT MERGE] WASM Fix Send requirements

Maan2003:
Database incorrectly requires `: Send`. This tries to fix that but in a generic way.


fedimint-api/src/task.rs
```diff
@@ -445,3 +445,76 @@ mod imp {
         }
     }
 }
+
+#[macro_export]
+macro_rules! async_trait {
```
dpc, line 450: Having it named the same as underlying will confuse people, imo.

Maan2003: `async_trait_maybe_send` ?

elsirion: sounds better and easier to find in the code base

elsirion:
Concept ACK, most of the module code won't need it anymore as soon as we untangle client and server part.

Maan2003:
Weird warning:
```
fedimint-workspace-clippy>    --> fedimint-core/src/db/mod.rs:263:14
fedimint-workspace-clippy>     |
fedimint-workspace-clippy> 252 | pub trait IDatabaseTransaction<'a>: 'a + MaybeSend {
fedimint-workspace-clippy>     |           -------------------- this trait cannot be made into an object...
fedimint-workspace-clippy> ...
fedimint-workspace-clippy> 263 |     async fn raw_remove_by_prefix(&mut self, key_prefix: &[u8]) -> Result<()> {
fedimint-workspace-clippy>     |              ^^^^^^^^^^^^^^^^^^^^ ...because method `raw_remove_by_prefix` references the `Self` type in its `where` clause
fedimint-workspace-clippy>     = help: consider moving `raw_remove_by_prefix` to another trait
fedimint-workspace-clippy>     = note: `-D where-clauses-object-safety` implied by `-D warnings`
fedimint-workspace-clippy> error: could not compile `fedimint-core` due to 2 previous errors
fedimint-workspace-clippy> warning: build failed, waiting for other jobs to finish...
fedimint-workspace-clippy> error: could not compile `fedimint-core` due to 2 previous errors
```

https://github.com/fedimint/fedimint/actions/runs/4197884089/jobs/7280789847#step:6:58

fedimint-core/src/db/mod.rs
```diff
@@ -966,7 +968,7 @@ pub async fn apply_migrations<'a>(
 }
 
 #[allow(unused_imports)]
-mod tests {
+pub mod tests {
```
m1sterc001guy, line 971: Does this need to be public?

fedimint-core/src/task.rs
```diff
@@ -446,3 +446,104 @@ mod imp {
         }
     }
 }
+
+/// async trait that use MaybeSend
```
m1sterc001guy, line 450: Can we add an explanation for why this is necessary? Will be more obvious for someone new to the codebase why these macros were created

m1sterc001guy:
> Concept ACK, most of the module code won't need it anymore as soon as we untangle client and server part, till then it's an improvement though

Should we make an issue for this so we remember to do the refactor?

Also, going forward, is the recommendation to always send `MaybeSend` in place of `Send`? What is the best way of preventing new code from being merged that breaks the pattern you're introducing here?

client/client-lib/src/api.rs
```diff
@@ -33,11 +32,10 @@ pub trait LnFederationApi {
     async fn offer_exists(&self, payment_hash: Sha256Hash) -> FederationResult<bool>;
 }
 
-#[cfg_attr(target_family = "wasm", async_trait(? Send))]
-#[cfg_attr(not(target_family = "wasm"), async_trait)]
+#[apply(async_trait_maybe_send!)]
```
dpc, line 35: Is `apply` something we're going to use for other stuff? If not maybe something like:

```
#[async_trait_ext(maybe_send)]
```

BTW. That `!` is easy to overlook and not clear (is it a `!Send` or some macro invocation)? If we need "not" in there, maybe let's include it as `maybe_not_send`?

fedimint-core/src/db/mem_impl.rs
```diff
@@ -3,11 +3,12 @@ use std::fmt::Debug;
 use std::sync::Mutex;
 
 use anyhow::Result;
-use async_trait::async_trait;
 use bitcoin_hashes::hex::ToHex;
 use futures::stream;
+use macro_rules_attribute::apply;
```
dpc, line 8: Oh, that's where it is coming from. Oh well. OK then.

client/client-lib/src/api.rs
dpc: I see. So `apply` is from some 3rd party library, and then `async_trait_maybe_send!` is a macro invocation. 

Oh well. Will learn how to live it it.

Maan2003: It is a tradeoff, the definition of macro is really simple. but the usage is a bit weird.

I can write a proc macro attribute if you want.

dpc: Nah, good enough.

Maan2003:
> MaybeSend in place of Send

Yes, use `MaybeSend` whenever possible, sometimes it is not. Example: anyhow::from requires `Send`

fedimint-core/src/db/mod.rs
Maan2003: yes completely accidental.

fedimint-core/src/task.rs
Maan2003: Will add more docs for MaybeSend. Explaining the reason.

Maan2003:
> help: consider moving `raw_remove_by_prefix` to another trait
>  = note: `-D where-clauses-object-safety` implied by `-D warnings`

see https://github.com/dtolnay/async-trait/issues/228

this is a false positive, fixed in Rust 1.69
solution is slapping `#![allow(where_clauses_object_safety)]` on every crate 
OR
 allowing it CI (passing `-- -A where_clauses_object_safety` to clippy)

dpc:
> solution is slapping `#![allow(where_clauses_object_safety)]` on every crate

Only ones that are using these macros?

These warning will be annoying for users building locally, so I think this is the route we need to pursue.

m1sterc001guy (approved):

elsirion (approved):

m1sterc001guy (approved):

PR #1657, opened 2023-02-11
Re: issues 1356. Eliminated reexport of serde_json from a number of m…

thessc1:
…odules. Also,reexported a number of 3rd party dependencies for fedimint-api.

How does this work?


.gitignore
```diff
@@ -8,3 +8,4 @@ it
 .tmpenv
 .direnv
 .vscode
+.DS_Store
```
elsirion, line 11: Please do not make unrelated edits in PRs. You can ignore paths locally by adding them to `.git/info/exclude`. If you think this should be added to the global `.gitignore` please open a separate PR (although I feel more like removing idea and vscode looking at it, these are just dev-specific and should not be global).

elsirion (commented):
Otherwise looks good :)

.gitignore
thessc1: Roger. Removed.

fedimint-server/Cargo.toml
```diff
@@ -20,12 +20,14 @@ bitcoin = "0.29.2"
...
+fedimint-api = { path = "../fedimint-api" }
+fedimint-core = { path = "../fedimint-core" }
+fedimint-wallet = { path = "../modules/fedimint-wallet", features = ["native"] }
 futures = "0.3.24"
 impl-tools = "0.6.1"
 itertools = "0.10.5"
 jsonrpsee = { version = "0.16.2", features = ["server"] }
-fedimint-api = { path = "../fedimint-api" }
-fedimint-core = { path = "../fedimint-core" }
+mint-client = { path = "../client/client-lib" }
```
elsirion, line 30: Why does the server rely on the mint client? Maybe that's a bit too involved and should go into a separate PR, but would be good to know. (pinging @jkitman who also worked on modularization)

jkitman: We just finished a refactor to avoid the `fedimint-server` depending on modules, we should keep it that way.

thessc1:
Hey folks, anything else needed here? Happy to resolve the conflicts if needed.

dpc (approved):

jkitman (changes requested):
Beyond the conflicts we cannot have `fedimint-server` depend on `mint-client` again.

dpc:
>

PR #1660, opened 2023-02-12
bech32 encode connect string

### LINKED ISSUE #1659: Bech32 encoding for connect string
> As a separate matter, now that we have the federation id in there taking up most of the bytes it might make sense to explore bech32 encoding of the whole string. This would get rid of json special characters and be more compact for binary data too. If bech32 is serialized in upper case it only uses 5.5 bit per character (so only an overhead of 10% for binary data compared to 100% for hex encoding+special characters forcing binary encoding).

_Originally posted by @elsirion in https://github.com/fedimint/fedimint/issues/1654#issuecomment-1426882261_

See also  #614

jkitman:
Reduction is 242 -> 229 bytes for 10 peers.

Fixes #1659

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -1,11 +1,17 @@
 use std::collections::{BTreeMap, BTreeSet};
-use std::fmt::{self, Debug};
+use std::fmt::{self, Debug, Display, Formatter};
+use std::io::{Cursor, Read};
 use std::pin::Pin;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 use std::{cmp, result};
 
+use anyhow::ensure;
 use async_trait::async_trait;
+use bech32::Variant::Bech32m;
+use bech32::{FromBase32, ToBase32};
+use bitcoin::consensus::ReadExt;
 use bitcoin_hashes::sha256;
 use fedimint_core::config::{
     ApiEndpoint, ClientConfig, ConfigResponse, FederationId, ModuleGenRegistry,
@@ -28,7 +34,7 @@ use jsonrpsee_ws_client::{WsClient, WsClientBuilder};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use thiserror::Error;
-use threshold_crypto::PublicKey;
+use threshold_crypto::{PublicKey, PK_SIZE};
 use tracing::{debug, error, instrument, trace, warn};
 use url::Url;
 
@@ -584,7 +590,7 @@ struct FederationMember<C> {
 /// Information required for client to construct [`WsFederationApi`] instance
 ///
 /// Can be used to download the configs and bootstrap a client
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct WsClientConnectInfo {
     /// Urls that support the federation API (expected to be in PeerId order)
     pub urls: Vec<Url>,
@@ -610,6 +616,64 @@ impl WsClientConnectInfo {
         let honest: Vec<_> = all.into_iter().take(num_honest).collect();
 
         WsClientConnectInfo::new(&config.federation_id, &honest)
+    }
+}
+
+/// We can represent client connect info as a bech32 string for compactness and
+/// error-checking
+///
+/// Human readable part (HRP) includes the version
+/// ```txt
+/// [ hrp (4 bytes) ] [ id (48 bytes) ] ([ url len (2 bytes) ] [ url bytes (url len bytes) ])+
+/// ```
+const BECH32_HRP: &str = "fed1";
+
+impl FromStr for WsClientConnectInfo {
+    type Err = anyhow::Error;
+
+    fn from_str(encoded: &str) -> Result<Self, Self::Err> {
+        let (hrp, data, variant) = bech32::decode(encoded)?;
+
+        ensure!(hrp == BECH32_HRP, "Invalid HRP in bech32 encoding");
+        ensure!(variant == Bech32m, "Expected Bech32m encoding");
+
+        let bytes: Vec<u8> = Vec::<u8>::from_base32(&data)?;
+        let total_len = bytes.len() as u64;
+        let mut cursor = Cursor::new(bytes);
+        let mut id_bytes = [0; PK_SIZE];
+        cursor.read_exact(&mut id_bytes)?;
+
+        let mut urls = vec![];
+        while cursor.position() < total_len {
+            let len = cursor.read_u16()? as usize;
+            let mut url_bytes = vec![0; len];
+            cursor.read_exact(&mut url_bytes)?;
+
+            let url = std::str::from_utf8(&url_bytes)?;
+            urls.push(url.parse()?);
+        }
+
+        Ok(Self {
+            urls,
+            id: FederationId(PublicKey::from_bytes(id_bytes)?),
+        })
+    }
+}
+
+/// Parses the connect info from a bech32 string
+impl Display for WsClientConnectInfo {
+    fn fmt(&self, formatter: &mut Formatter<'_>) -> fmt::Result {
+        let mut data = vec![];
+        data.extend(self.id.0.to_bytes());
+        for url in &self.urls {
+            let url_bytes = url.as_str().as_bytes();
+            data.extend((url_bytes.len() as u16).to_le_bytes());
+            data.extend(url.as_str().as_bytes());
+        }
+        let encode =
+            bech32::encode(BECH32_HRP, data.to_base32(), Bech32m).map_err(|_| fmt::Error)?;
+
+        formatter.write_str(&encode)
     }
 }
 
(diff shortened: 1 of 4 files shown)


fedimint-core/src/api.rs
```diff
@@ -595,6 +599,40 @@ impl WsClientConnectInfo {
...
+        for url in url_str.split(';') {
+            urls.push(url.parse()?);
+        }
+
+        Ok(Self {
+            urls,
+            id: FederationId(PublicKey::from_bytes(id_bytes)?),
+        })
+    }
+
```
elsirion, line 635: If this becomes the canonical user-facing representation it should probably rather be implemented as `Display` and `FromStr`.

fedimint-core/src/api.rs
```diff
@@ -595,6 +599,40 @@ impl WsClientConnectInfo {
...
     }
 
+    /// Converts the connect info into a bech32 string for compact display / copying
+    pub fn to_bech32(&self) -> anyhow::Result<String> {
+        let mut data = vec![];
+        data.extend(self.id.0.to_bytes());
+        for url in &self.urls {
+            data.extend(";".as_bytes());
+            data.extend(url.as_str().as_bytes());
+        }
```
elsirion, line 609: I'd much prefer having a length-prefixed encoding than relying on a delimiter that may very well be part of a valid URL. 

An alternative encoding:
```
[ id (48 bytes) ] ([ url len (2 bytes) ] [ url bytes (url len bytes) ])+
```

elsirion (changes requested):

dpc:
Why bech32? I don't think we need error detection&correction, no one will spell it over the phone or write down on a piece of paper. Also 32-based is not as compact as base58 or base64 would be.

For connection string all we care about is how well they copy&paste, so we want characters that don't confuse devices to consider them separate words, and then the shorter the better. Short checksum to detect copy&paste errors would be enough.

elsirion:
> @dpc Bech32 has the nice property (if all uppercase) that it fits into the alphanumeric charset of QR codes. This means each bech32 character takes 5.5 bits to encode in a QR code while itself encoding 5bit, so wasting merely 10% for binary data while still being copyable. Compare that to e.g. base64 encoding where each character encodes 6bit but takes 8bit to encode in a QR code due to the larger charset, that's an overhead of 50% for binary data (base58 even worse).

https://github.com/fedimint/fedimint/pull/1654#issuecomment-1427259762

dpc:
@elsirion That's a property of base32, not just bech32. If it's better for some reason, sure we can use it. But I don't see a benefit of extra error detection&correction, which only enlarge the encoding.

I might be wrong, but the idea with the QR codes might be that addresses might often be included in some larger body of text encoded in a QR code (like URLs), and QR code might have some special mode for when only text is used. I don't see that being the case for us though. I would be guessing that QR codes should have a binary-mode encoding, and then we should avoid double-encoding things:  we would go straight from binary data to QR code. If possible - I'm making a guess that QR encoding must have support for encoding binary data.

dpc:
Actually, we might want to invent url/uri for connecting string. `fedimint:...`, and then multiple urls might get large so maybe base32 is a good idea.

jkitman:
For QR codes use base32 and copy+pasting use base64 if we are optimizing for size of both?  Either way, I don't think we need to over optimize much...the size of the string is not too large for most federations and we haven't really figured out the best UX flow for how this should work.

Perhaps in the long-run there will be trusted sites that host links to known federation ids posted on social media / websites?  

And the download of configs can happen independently of that verification?

elsirion:
> I would be guessing that QR codes should have a binary-mode encoding, and then we should avoid double-encoding things

I think that's not a great UX in many cases @dpc. E.g. I often use another QR reader if the app-provided one doesn't work well, for that the encoded data should be human readable. The base32 alphabet chosen by bech32 works very well for that with minimal overhead.

> Actually, we might want to invent url/uri for connecting string. fedimint:..., and then multiple urls might get large so maybe base32 is a good idea.

That will still fit into the alphanumeric charset, so bech32 still makes sense (also why LNURL uses it afaik, even though I find it annoying there).

For reference:

Screenshot 2023-02-13 at 10-27-01 QR code - Wikipedia

dpc:
OK, so base32. But what do we need that overhead of bech32 for? AFAIU QR codes already have error correction, and no one will be spelling these connection strings over the phone, I guess.

elsirion:
The error correction overhead of bech32 is quite small. I wouldn't mind using a custom base32 encoding without it, but using a standard is kinda nice because libraries already exist for all kinds of platforms.

dpc:
OK then.

elsirion:
But we should probably use bech32m, the newest version of it that protects against extensions.

fedimint-core/src/api.rs
```diff
@@ -613,6 +617,51 @@ impl WsClientConnectInfo {
...
+
+        let bytes: Vec<u8> = Vec::<u8>::from_base32(&data)?;
+        let id_bytes: [u8; PK_SIZE] = bytes[..PK_SIZE].try_into()?;
+        let mut urls = vec![];
+        let mut index = PK_SIZE;
+
+        while index < bytes.len() {
+            let len = u16::from_be_bytes(bytes[index..index + 2].try_into()?) as usize;
+            index += 2;
+            let url = std::str::from_utf8(&bytes[index..index + len])?;
```
elsirion, line 637: This code crashes with malicious input (len greater than total byte len). We should just use a cursor instead of incrementing `index` and then use it as a `Read` object. I think that would also give us helpers for integer decoding. That should automatically give us sane errors when trying to read too much.

elsirion (changes requested):

jkitman:
@dpc @elsirion since it sounds like we are good with bech32m, this PR is ready to go now

fedimint-core/src/api.rs
```diff
@@ -613,6 +619,54 @@ impl WsClientConnectInfo {
...
+        ensure!(variant == Bech32m, "Expected Bech32m encoding");
+
+        let bytes: Vec<u8> = Vec::<u8>::from_base32(&data)?;
+        let total_len = bytes.len() as u64;
+        let mut cursor = Cursor::new(bytes);
+        let mut id_bytes = [0; PK_SIZE];
+        cursor.read_exact(&mut id_bytes)?;
+
+        let mut urls = vec![];
+        while cursor.position() < total_len {
```
elsirion, line 638: https://doc.rust-lang.org/std/io/struct.Cursor.html#method.is_empty

jkitman: Needed to update rust, thought it was still unstable.

jkitman: Oh wait still is...

fedimint-core/src/api.rs
```diff
@@ -613,6 +619,54 @@ impl WsClientConnectInfo {
...
+impl FromStr for WsClientConnectInfo {
+    type Err = anyhow::Error;
+
+    fn from_str(encoded: &str) -> Result<Self, Self::Err> {
+        let (hrp, data, variant) = bech32::decode(encoded)?;
+
+        ensure!(hrp == "fed", "Invalid HRP in bech32 encoding");
+        ensure!(variant == Bech32m, "Expected Bech32m encoding");
+
+        let bytes: Vec<u8> = Vec::<u8>::from_base32(&data)?;
```
dpc, line 631: Can we add one bytes for "version", in case we ever need to change anything?

jkitman: Sure, I think the standard way is to put it into the HRP, so make it "fed1"

dpc: :+1: That would be pretty.

fedimint-core/src/api.rs
```diff
@@ -613,6 +619,54 @@ impl WsClientConnectInfo {
     }
 }
 
+impl FromStr for WsClientConnectInfo {
+    type Err = anyhow::Error;
+
+    fn from_str(encoded: &str) -> Result<Self, Self::Err> {
```
mxz42, line 625: can we add the expected representation as documentation somewhere? I think it'd be nice to have it

```
[ id (48 bytes) ] ([ url len (2 bytes) ] [ url bytes (url len bytes) ])+
```

fedimint-core/src/api.rs
```diff
@@ -613,6 +619,63 @@ impl WsClientConnectInfo {
     }
 }
 
+/// We can represent client connect info as a bech32 string for compactness and
+/// error-checking
+///
+/// Human readable part (HRP) includes the version
+/// [ hrp (4 bytes) ] [ id (48 bytes) ] ([ url len (2 bytes) ] [ url bytes (url
+/// len bytes) ])+
```
mxz42, line 627: let's put than in a code block so it's not getting formatted

```rust 

/// ```txt
/// [ hrp (4 bytes) ] [ id (48 bytes) ] ([ url len (2 bytes) ] [ url bytes (url len bytes) ])+
/// ```

```

mxz42 (approved):

dpc (approved):

PR #1661, opened 2023-02-13
Make `Decoder` extensible

### LINKED ISSUE #1633: Make `Decoder` API extensible
I have to add new module associated types for the client state machine (#1447). Since we want to disentangle client and server code base (will help a lot with WASM) adding decoding functions to the current `Decoder` trait is not a good option.

Instead I plan to make the decoder a map from `TypeId` to decode fn returning `Box<Any>`, which in then casted to the appropriate `Dyn*` type. All this can be abstracted away with a generic fn imo.

elsirion:
Currently the decoder can only decode 4 specific module associated types/their `Dyn*`-type abstractions that got defined by the server module.

This is problematic for the client that needs to define its own module associated types. This PR allows arbitrary such types to be added easily without extending the `(I)Decoder` trait.

Fixes #1633


fedimint-core/src/module/mod.rs
```diff
@@ -618,10 +619,28 @@ impl<CI> ConsensusProposal<CI> {
...
+        decoder_builder.with_decodable_type::<Self::Output>();
+        decoder_builder.with_decodable_type::<Self::OutputOutcome>();
+        decoder_builder.with_decodable_type::<Self::ConsensusItem>();
+        decoder_builder
+    }
+}
+
 #[apply(async_trait_maybe_send!)]
 pub trait ServerModule: Debug + Sized {
+    type ModuleTypes: ModuleTypes;
```
dpc, line 641: What's the root reason why we need another level of indirection instead of inline? because it is shared between client and server? In that case maybe this should be `Common: ModuleCommon`?

dpc (approved):

fedimint-core/src/module/mod.rs
elsirion: Yes, pretty much. I want to get rid of the client module having an associated type which is the server module type. Naming can be changed.

elsirion: Renamed.

dpc (approved):

fedimint-core/src/core.rs
```diff
@@ -110,11 +108,8 @@ macro_rules! module_dyn_newtype_impl_encode_decode {
                 r: &mut R,
                 modules: &$crate::module::registry::ModuleDecoderRegistry,
             ) -> Result<Self, DecodeError> {
-                $crate::core::encode::module_decode_key_prefixed_decodable(
-                    r,
-                    modules,
-                    |r, m, id| m.$decode_fn(r, id),
-                )
+                let key = ModuleInstanceId::consensus_decode(r, modules)?;
+                modules.get_expect(key).decode(r, key)
```
mxz42, line 112: nit: imho, variable names should have at least 3 letters. `item, err, ctx, ...`.  It makes things easier to read. Even for short methods like this; especially as methods may be extended in the future and suddenly a variable name is used in a long block of code. 
Feel free to ignore though, it's a bit unrelated to the changes.

fedimint-core/src/core.rs
```diff
@@ -240,192 +267,99 @@ macro_rules! newtype_impl_display_passthrough_with_instance_id {
...
-    /// Decode `Output` compatible with this module, after the module key prefix
-    /// was already decoded
-    fn decode_output(
-        &self,
-        r: &mut dyn io::Read,
-        instance_id: ModuleInstanceId,
-    ) -> Result<DynOutput, DecodeError>;
+/// A type that has a `Dyn*`, type erased version of itself
+pub trait IntoDynInstance {
+    /// The type erased version of this type
```
mxz42, line 272: when I read this initially, it was a bit unclear to me what `this` refers to.
nit: `/// The type erased version of the type implementing this trait`

fedimint-core/src/core.rs
```diff
@@ -240,192 +267,99 @@ macro_rules! newtype_impl_display_passthrough_with_instance_id {
...
 
-    fn decode_consensus_item(
+    /// Decodes a specific `DynType` from the `reader` byte stream.
+    ///
+    /// # Panics
+    /// * If no decoder is registered for the `DynType`
+    pub fn decode<DynType: Any>(
         &self,
-        r: &mut dyn Read,
+        reader: &mut dyn Read,
```
mxz42, line 347: oh, cool! you actually changed the `r` to `reader` here :+1:

modules/fedimint-mint/src/lib.rs
```diff
@@ -455,16 +452,21 @@ impl std::fmt::Display for MintConsensusItem {
     }
 }
 
+pub struct MintModuleTypes;
```
mxz42, line 455: i was wondering if I could come up with a better name that does not include `types` as it's a bit generic but I couldn't. And perhaps generic/or general is what we want here, as this represents the common types/structs/required consensus types shared between the client and server part of a module. Also thinking with regards to module implementors what `ModuleCommon` is and the trait documentation says that, so :+1:

mxz42 (approved):
Cool PR!
was also very nice to follow along the commits and see how the code developed

fedimint-core/src/core.rs
elsirion: renamed `r` to `reader` (I just didn't change it previously)

fedimint-core/src/core.rs
elsirion: fixed

fedimint-core/src/core.rs
```diff
@@ -240,192 +267,99 @@ macro_rules! newtype_impl_display_passthrough_with_instance_id {
...
+    ///
+    /// **Caution**: One `Decoder` object should only contain decoders that
+    /// belong to the same [*module kind*](fedimint_core::core::ModuleKind).
+    ///
+    /// # Panics
+    /// * If multiple `Types` with the same `DynType` are added
+    pub fn with_decodable_type<Type>(&mut self)
+    where
+        Type: IntoDynInstance + Decodable,
+        Type::DynType: 'static,
```
mxz42, line 308: just a question: is `DynType` always `'static` (so it could be moved to the associated type definition in the trait  itself)? 
if my regex isn't lying to me, `IntoDynInstance` is currently implemented for these types:

```sh
$ (pr/1661 *%) | rg -i 'impl.*?IntoDynInstance'
fedimint-core/src/core.rs
178:        impl fedimint_core::core::IntoDynInstance for $input {
188:        impl fedimint_core::core::IntoDynInstance for $output {
198:        impl fedimint_core::core::IntoDynInstance for $outcome {
208:        impl fedimint_core::core::IntoDynInstance for $ci {
```

e.g. the `DynInput` is expanded: 

```rust
pub struct DynInput(Box<dyn IInput + 'static + Send + Sync>, ::fedimint_core::core::ModuleInstanceId);
```
so, they're always `'static`, right? 

do we might have cases in the future where  `DynType` would not be `'static`  and, therefore, would be limiting to put the `type DynType: 'static;` bound in the trait itself?

just wondering

mxz42 (approved):

fedimint-core/src/core.rs
dpc: non-`'static` would mean that the given value holds some references to other things, which would get in a way of treating it as "opaque value", so I don't think we would ever want it.

dpc (approved):

fedimint-core/src/core.rs
mxz42: right, so we could move the `'static` bound into the trait `IntoDynInstance` where the associated type is defined: `type DynType: 'static`?

dpc: Possibly yes. Might be a good idea.

mxz42: I know that it compiles because I tried it. 
I was especially interested about the reasoning. So yeah, as we treat `DynType` as an "opaque value" I think it sounds like a good idea to have in the trait definition

Issue #1658, opened 2023-02-11
Switch `is_valid_for_network` to `rust-bitcoin` implementation


elsirion:
We have an old version from before the [patch landed in rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/pull/618):

https://github.com/fedimint/fedimint/blob/2af4e6a02f8ca4fb42165bfed263c681c225d9f4/modules/fedimint-wallet/src/lib.rs#L1638-L1647

But we should switch to [the rust-bitcoin version](https://docs.rs/bitcoin/0.29.2/src/bitcoin/util/address.rs.html#818-830) now.

jp1ac4:
Hi, is the idea here to replace the contents of the `is_address_valid_for_network` function with a call to rust-bitcoin's `is_valid_for_network` function or rather to remove this function entirely and call the rust-bitcoin function directly elsewhere, e.g. in the `validate_output` function?

elsirion:
Remove function entirely and just use the rust-bitcoin version of it instead.

PR #1662, opened 2023-02-13
Use rust-bitcoin's `is_valid_for_network`

jp1ac4:
Hi, this PR is for #1658. Thanks.

### DIFF
--- a/modules/fedimint-wallet/src/lib.rs
+++ b/modules/fedimint-wallet/src/lib.rs
@@ -15,8 +15,8 @@ use bitcoin::util::psbt::raw::ProprietaryKey;
 use bitcoin::util::psbt::{Input, PartiallySignedTransaction};
 use bitcoin::util::sighash::SighashCache;
 use bitcoin::{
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet/src/lib.rs
```diff
@@ -690,7 +690,10 @@ impl ServerModule for Wallet {
         dbtx: &mut DatabaseTransaction,
         output: &WalletOutput,
     ) -> Result<TransactionItemAmount, ModuleError> {
-        if !is_address_valid_for_network(&output.recipient, self.cfg.consensus.network) {
+        if !&output
```
elsirion, line 693: Why is that `&` here? function calls typically do all the referencing/dereferencing themselves.

elsirion (commented):
Thx, looks good, but we should avoid that random-looking reference op.

modules/fedimint-wallet/src/lib.rs
jp1ac4: Ah, thanks for spotting this. I've now removed it.

justinmoon (approved):

PR #1664, opened 2023-02-13
Refactor DKG into fedimint-server

jkitman:
This refactor moves the DKG code out of `fedimintd` and `fedimint-core` into `fedimint-server` where it belongs.

For now, we need a `server` feature flag on the modules until the client lib is rewritten to not depend on the server modules.

Will need to continue to follow-up with more refactors to get to the proper code organization:

- `fedimint-server` should contain the server-only library code
- Server modules should depend on `fedimint-server`
- Minimize what is in `fedimintd` it should only contain code that plugs together server modules


fedimint-server/src/config/distributedgen.rs
```diff
@@ -0,0 +1,480 @@
...
+    threshold: usize,
+    f1_poly: Poly<Scalar, Scalar>,
+    f2_poly: Poly<Scalar, Scalar>,
+    hashed_commits: BTreeMap<PeerId, Sha256>,
+    commitments: BTreeMap<PeerId, Vec<G>>,
+    sk_shares: BTreeMap<PeerId, Scalar>,
+    pk_shares: BTreeMap<PeerId, Vec<G>>,
+}
+
+/// Implementation of "Secure Distributed Key Generation for Discrete-Log Based Cryptosystems"
```
dpc, line 39: I don' `impl` blocks can have docstrings. This should be on `struct`? Does this work? How is this not triggering the compiler. What is happening? I need a coffee.

dpc (dismissed):

dpc (approved):

justinmoon (approved):

PR #1667, opened 2023-02-13
refactor: clarify parameters of db macro

mxz42:
implement feedback from [this conversation](https://github.com/fedimint/fedimint/pull/1608#discussion_r1103269619)

@m1sterc001guy

### DIFF
--- a/fedimint-api/src/db/mod.rs
+++ b/fedimint-api/src/db/mod.rs
@@ -695,38 +695,61 @@ where
 }
 
 /// This is a helper macro that generates the implementations of
-/// DatabaseKeyPrefixConst necessary for reading/writing to the
+/// [`DatabaseKeyPrefixConst`] necessary for reading/writing to the
 /// database and fetching by prefix.
 ///
 /// - `key`: This is the type of struct that will be used as the key into the
-///   database.
+///   database
 /// - `value`: This is the type of struct that will be used as the value into
-///   the database.
-/// - `prefix`: Enum expression that is represented as a `u8` that is the prefix
-///   prepended to this key.
-/// - `key_prefix`: Optional type of struct that can be passed multiple times if
-///   necessary. These will be used
-/// as the key prefix when querying the database via `find_by_prefix`.
+///   the database
+/// - `db_prefix`: Required enum expression that is represented as a `u8` and is
+///   prepended to this key
+/// - `query_prefix`: Optional type of struct that can be passed zero or more
+///   times. Every query prefix can be used to query the database via
+///   `find_by_prefix`
 ///
 /// Examples:
 ///
-/// `impl_db_prefix_const!(key = TestKey, value = TestVal, prefix =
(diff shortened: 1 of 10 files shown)


justinmoon:
Merge conflicts after rustfmt change

fedimint-api/src/db/mod.rs
```diff
@@ -695,38 +695,57 @@ where
...
 ///
-/// `impl_db_prefix_const!(key = AltTestKey, value = TestVal, prefix =
-/// TestDbKeyPrefix::AltTest, key_prefix = AltDbPrefixTestPrefix);`
+/// ```
+/// use crate::fedimint_api::impl_db_prefix_const;
+/// impl_db_prefix_const!(
+///     key = TestKey,
+///     value = TestVal,
+///     db_prefix =
+/// TestDbKeyPrefix::Test
```
m1sterc001guy, line 721: Was this automated by `cargo fmt`? Feels a little weird to have this on a newline

m1sterc001guy (dismissed):
LGTM

fedimint-api/src/db/mod.rs
mxz42: yeah, good catch. it was definitely auto-formatted. perhaps because I didn't import the type.

fedimint-api/src/db/mod.rs
```diff
@@ -695,38 +695,61 @@ where
...
-///   the database.
-/// - `prefix`: Enum expression that is represented as a `u8` that is the prefix
-///   prepended to this key.
-/// - `key_prefix`: Optional type of struct that can be passed multiple times if
-///   necessary. These will be used
-/// as the key prefix when querying the database via `find_by_prefix`.
+///   the database
+/// - `db_prefix`: Required enum expression that is represented as a `u8` and is
+///   prepended to this key
+/// - `query_prefix`: Optional type of struct that can be passed zero or more
```
elsirion, line 707: Not sure if `query_prefix` is any more clear, but fine by me.

elsirion (approved):

m1sterc001guy (approved):

fedimint-api/src/db/mod.rs
mxz42: I felt that the word `key` and `prefix` is a bit overloaded in this context, so I thought that this might clear things up. 

My reasoning/thinking process was

- `db_prefix` is the prefix of the `struct Thing` in the db and is always required
- `query_prefix` 
  - is optional
  - will very likely  (always, I guess) contain the `db_prefix` and then some other fields of the `key` (e.g. the `struct Thing`) to narrow down the returned values. It has nothing to do with the `db_prefix` of the key itself (remember, `query_prefix` was  previously called `key_prefix`). And I think that was the confusing part for me: are we prefixing the `key`, so the `db_prefix`? and we're not. 

When we think in database and key-value storage terms, the term `key_prefix` does make sense, of course, as we operate on a database key.

PR #1674, opened 2023-02-13
Refactor the config io to simplify

jkitman:
Small refactor to remove the prompt password which we never use and probably shouldn't have deep in the lib code.

Also simplifies some of the shared config io code into `read_server_config` and `write_server_config`


justinmoon (approved):
Much simpler. We should probably remove rpassword from gateway-cli, too, @okjodom.

elsirion (approved):

PR #1675, opened 2023-02-13
fix: rename database traits

m1sterc001guy:
Some offline discussion with @dpc and @mxz42 has shown that there is some confusion on the purpose of some database traits. Currently there are many traits named similar things, so their purpose isn't immediately clear. This PR renames the database traits with the intention of making it more clear on their purpose.

- `DatabaseKeyPrefixConst` -> `DatabaseRecord` and `DatabaseLookup`. `DatabaseRecord` is for individual db records and `DatabaseLookup` is for multi-record lookups.
- `SerializableDatabaseValue` was collapsed into `DatabaseValue`
- `DatabaseRecord` and `DatabaseLookup` derive from `DatabaseKeyPrefix` since both of those need the ability to prepend a prefix. This is the same functionality as before but the syntax is a bit cleaner.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -26,26 +26,44 @@ use crate::module::registry::ModuleDecoderRegistry;
 
 pub const MODULE_GLOBAL_PREFIX: u8 = 0xff;
 
-pub trait DatabaseKeyPrefixConst {
+pub trait DatabaseKeyPrefix: Debug {
+    fn to_bytes(&self) -> Vec<u8>;
+}
+
+/// A key + value pair in the database with a unique prefix
+/// Extends `DatabaseKeyPrefix` to prepend the key's prefix.
+pub trait DatabaseRecord: DatabaseKeyPrefix {
     const DB_PREFIX: u8;
     type Key: DatabaseKey + Debug;
     type Value: DatabaseValue + Debug;
 }
 
-pub trait DatabaseKeyPrefix: Debug {
-    fn to_bytes(&self) -> Vec<u8>;
+/// A key that can be used to query one or more `DatabaseRecord`
+/// Extends `DatabaseKeyPrefix` to prepend the key's prefix.
+pub trait DatabaseLookup: DatabaseKeyPrefix {
+    type Record: DatabaseRecord;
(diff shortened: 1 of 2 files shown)


dpc (dismissed):

justinmoon:
LGTM. Will defer the merge to @mxz42.

mxz42 (approved):
I love it! Great PR

mxz42:
damn, was too late. has a merge conflict now @m1sterc001guy

fedimint-core/src/db/mod.rs
```diff
@@ -710,50 +728,68 @@ where
...
 ///
 /// impl_db_prefix_const!(
-///     key = AltTestKey,
-///     value = TestVal,
-///     db_prefix = TestDbKeyPrefix::AltTest,
-///     query_prefix = AltDbPrefixTestPrefix
+///     key = MyKey,
+///     value = MyValue,
+///     db_prefix = DbKeyPrefix::MyKey,
+///     query_prefix = MyKeyPrefix
```
mxz42, line 777: I like it!

fedimint-core/src/db/mod.rs
```diff
@@ -686,18 +714,8 @@ where
...
-{
-    fn from_bytes(data: &[u8], modules: &ModuleDecoderRegistry) -> Result<Self, DecodingError> {
-        T::consensus_decode(&mut std::io::Cursor::new(data), modules)
-            .map_err(|e| DecodingError::Other(e.0))
-    }
-}
-
 /// This is a helper macro that generates the implementations of
-/// [`DatabaseKeyPrefixConst`] necessary for reading/writing to the
+/// DatabaseRecord necessary for reading/writing to the
```
mxz42, line 718: nit: at some point add backticks around the type

mxz42 (approved):

fedimint-core/src/db/mod.rs
m1sterc001guy: Whoops, missed this. Fixed now

m1sterc001guy:
@jkitman doc strings should be fixed with this PR

justinmoon (approved):

mxz42 (approved):

PR #1677, opened 2023-02-14
Setup UI fixes

justinmoon:
trim whitespace and dash-separate guardian names because they are treated as DNS names which can't contain spaces

### DIFF
--- a/fedimintd/src/ui.rs
+++ b/fedimintd/src/ui.rs
@@ -125,8 +125,13 @@ async fn post_guardians(
     let state_copy = state.clone();
     let mut state = state.lock().await;
     let params = state.params.clone().expect("invalid state");
-    let mut connection_strings: Vec<String> =
-        serde_json::from_str(&form.connection_strings).expect("not json");
+    let connection_strings: Vec<String> = serde_json::from_str(&form.connection_strings)
+        .map_err(|_| anyhow::anyhow!("Invalid connection string"))?;
+    let mut connection_strings: Vec<String> = connection_strings
+        .into_iter()
+        .map(|s| s.trim().to_string())
+        .collect();
+
     connection_strings.push(params.guardian.tls_connect_string);
 
     // Don't allow re-running DKG if configs already exist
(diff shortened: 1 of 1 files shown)


dpc (approved):
Should we sanity check for invalid characters?

justinmoon:
> Should we sanity check for invalid characters?

Yea. Should probably do this on all form inputs. What characters should be considered invalid?

dpc:
> Yea. Should probably do this on all form inputs. What characters should be considered invalid?

> DNS names can contain only alphabetical characters (A-Z), numeric characters (0-9), the minus sign (-), and the period (.). Period characters are allowed only when they are used to delimit the components of domain style names.

elsirion:
Given that we want to migrate an API+PWA approach I'd like to avoid having too much cleverness on the server side that might creep into the API eventually. Couldn't we apply these simple fixes on the client side? Especially in the case of federation name that would be more transparent to the user too (live replace?).

dpc:
If the server side requires dns-safe characters to work, it should do the input sanitization on the server side anyway, no? I agree that UI-side sanitization would be great, but probably in addition.

elsirion:
Imo the server should check and fail if invalid, not apply random fixes. The client can try to nudge the user to do the right thing.

mxz42:
> Imo the server should check and fail if invalid, not apply random fixes. The client can try to nudge the user to do the right thing.

Yes, the server *must validate* and fails if it gets invalid data.

> UI-side sanitization would be great, but probably in addition.

The client *should* also validate, so that a user has a good UX and doesn't enter all the details and then gets a bad response from the server. 


Sorry, missing a bit of context: So, we're using the guardian name as DNS name and TLS Server name indication. If we only need to derive the DNS name from the guardian name, then I'd probably be fine with replacing spaces with `-` or `_` to make it a valid DNS name *if and only if* the rest of the name contains valid DNS characters that need to be validated by the server first  (and client-side to have good UX).

okjodom (approved):

dpc (approved):

justinmoon:
> Couldn't we apply these simple fixes on the client side? 

Yea I could do that

elsirion:
Needs rebase

PR #1678, opened 2023-02-14
Fixes up some docstrings that used fedimint-api

jkitman:
Small fixup


jkitman:
@m1sterc001guy still some problems in `cargo test --doc` I'm having trouble fixing...

m1sterc001guy:
Ah yeah, that example references a struct that is not accessible from that module.

Feel free to just delete and I'll fix in this PR since I'm doing renames anyway: https://github.com/fedimint/fedimint/pull/1675#issuecomment-1429962029

Issue #1532, opened 2023-01-26
Setup UI should accept the default values of fields from the command-line


dpc:
That would make setup scriptable and less tiresome, while still testing the UI.

jp1ac4:
Hi, I could give this a try. Is it related to the `run-ui.sh` script?

dpc:
@jp1ac4 Go for it. Yes, from the looks of it `run-ui.sh` starts N federations with their setup UI. 

We would just like to have env variables handled by clap that would carry all over the UI screen with connections strings etc. and populate them.

Note: We will probably at some point change how the UI will work altogether, but until then this would be a great improvement.

justinmoon:
@jp1ac4 I can help you here. Have some changes that partially implement this. Can help you finish them off. Can also just explain how the setup UI works in general to help you get started with it. Ping me on discord and let's do a quick call.

dpc:
Lots of this is already working, but there were few fields missing, IIRC.

dpc:
:+1:

dpc:
@jp1ac4 Awesome job!

jp1ac4:
Thanks, @dpc :)

PR #1683, opened 2023-02-14
WIP: Get default values from env variables

jp1ac4:
Hi, I've started making some changes towards #1532 based on a discussion with @justinmoon.

Values for the networking parameters are read from env variables and passed to the params page. I haven't handled them using clap as suggested in https://github.com/fedimint/fedimint/issues/1532#issuecomment-1428317094, but can try adding that if required.

After initial feedback, I can do the same for the other parameters too.

Thanks.

### DIFF
--- a/fedimintd/src/ui.rs
+++ b/fedimintd/src/ui.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -220,6 +221,10 @@ async fn post_guardians(
 struct UrlConnection {
     ro_bitcoin_rpc_type: &'static str,
     ro_bitcoin_rpc_url: String,
+    bind_p2p: String,
+    p2p_url: String,
+    bind_api: String,
+    api_url: String,
 }
 
 async fn params_page(
@@ -244,6 +249,10 @@ async fn params_page(
     UrlConnection {
         ro_bitcoin_rpc_type,
         ro_bitcoin_rpc_url,
+        bind_p2p: env::var("FM_BIND_P2P").unwrap_or("".to_string()),
+        p2p_url: env::var("FM_P2P_URL").unwrap_or("".to_string()),
+        bind_api: env::var("FM_BIND_API").unwrap_or("".to_string()),
+        api_url: env::var("FM_API_URL").unwrap_or("".to_string()),
     }
 }
 
--- a/fedimintd/templates/params.html
+++ b/fedimintd/templates/params.html
@@ -45,6 +45,7 @@ content %}
         id="bind_p2p"
         name="bind_p2p"
         placeholder="127.0.0.1:8173"
+        value="{{bind_p2p}}"
       />
     </div>
     <div class="form-group">
@@ -55,6 +56,7 @@ content %}
         id="p2p_url"
         name="p2p_url"
         placeholder="wss://p2p.your-website.com"
+        value="{{p2p_url}}"
       />
     </div>
     <div class="form-group">
@@ -65,6 +67,7 @@ content %}
         id="bind_api"
         name="bind_api"
         placeholder="127.0.0.1:8174"
+        value="{{bind_api}}"
       />
     </div>
     <div class="form-group">
@@ -75,6 +78,7 @@ content %}
         id="api_url"
         name="api_url"
         placeholder="wss://api.your-website.com"
+        value="{{api_url}}"
       />
     </div>
     <h5>Wallet settings</h5>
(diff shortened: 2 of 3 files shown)


fedimintd/templates/params.html
```diff
@@ -54,7 +54,7 @@ <h5>Network connectivity</h5>
         class="form-control"
         id="p2p_url"
         name="p2p_url"
-        placeholder="wss://p2p.your-website.com"
```
justinmoon, line 57: Can we keep these placeholders? If `""` gets passed in the placeholder should still display.

scripts/run-ui.sh
```diff
@@ -6,11 +6,18 @@ rm -r $PWD/fed-ui
...
+export FM_BITCOIND_RPC="http://bitcoin:bitcoin@127.0.0.1:18443"
 bitcoind -regtest -daemon -fallbackfee=0.0004 -txindex -server -rpcuser=bitcoin -rpcpassword=bitcoin
 
 # start guardians
 for ((ID = 0; ID < $FM_FED_SIZE; ID++)); do
   mkdir $PWD/fed-ui/mock-$ID
+  p2p_port=$((10000 + 10000 * $ID))
+  api_port=$((10001 + 10000 * $ID))
+  export FM_BIND_P2P=127.0.0.1:$p2p_port
+  export FM_P2P_URL=ws://127.0.0.1:$p2p_port
```
justinmoon, line 18: Can you change this to `export FM_P2P_URL=fedimint://127.0.0.1:$p2p_port`

We use a custom protocol so `fedimint://` is a better prefix than `ws://`.

justinmoon:
Looks good. I ran it and it worked as expected. Network settings pre-filled from environment variable. 2 small comments then it should be good to go IMO.
image

fedimintd/templates/params.html
jp1ac4: Sure, I've added the placeholders back in, which will be displayed if the corresponding value is empty.

Regarding the values, I've included two different approaches. One checks for an empty string and the other checks for `Some`/`None`. You can let me know which way you prefer and I can make them all consistent.

scripts/run-ui.sh
jp1ac4: I've changed that now. Does this also apply to `FM_API_URL`?

jp1ac4:
Thanks for your feedback, @justinmoon. I've made the changes you requested and added a couple of follow-up questions to your comments.

Once it's all ready, I'll squash my commits.

fedimintd/src/ui.rs
```diff
@@ -236,6 +241,13 @@ async fn params_page(
     UrlConnection {
         ro_bitcoin_rpc_type,
         ro_bitcoin_rpc_url,
+        bind_p2p: match env::var("FM_BIND_P2P") {
+            Ok(s) => Some(s),
+            _ => None,
+        },
+        p2p_url: env::var("FM_P2P_URL").unwrap_or("".to_string()),
```
jkitman, line 248: Seems unsafe to default to an empty string, better to return an error.

justinmoon: This just pre-fills a value in the HTML form based on an environment variable. If no environment variable is set, then the form field is empty and user needs to fill it in themselves.

fedimintd/templates/params.html
justinmoon: I don't think you even need the if-else. Just do it like this:

```
        value={{s}}
        placeholder="..."
```

fedimintd/src/ui.rs
jkitman: Could be nice if we auto-generated the bind addresses.

justinmoon: What do you mean auto-generated? Choose a random open port?

fedimintd/templates/params.html
jp1ac4: Ah yes, you're right. When I initially tried that the value was showing up as `/`, but I think it's because I wasn't enclosing it in quotes. I've now made this change using quotes around the value.

I've also removed the use of `Option` and just default to an empty string again.

fedimintd/src/ui.rs
jp1ac4: Shall I try adding auto-generation as part of this PR or should that be done separately?

justinmoon: Problem with that is that this port needs to be opened or forwarded somehow. And I think this port is also used for DKG, right? So unless the user had basically all their ports open or opened the randomly selected port after submitting this form, then DKG wouldn't complete?

justinmoon (approved):
Just tried this and it worked very nicely.

If I removed the exported environment variables from `run-ui.sh`, then the form just behaved as it used to with placeholders showing. So it works with or without environment variables set.

scripts/run-ui.sh
jp1ac4: Ah OK, thanks for clarifying that.

Sure, I'll create a separate PR for the placeholder and link it to the issue you've just created (#1744).

elsirion (approved):

PR #1684, opened 2023-02-14
refactor: remove unnecessary method to start db transaction

mxz42:
Is there a reason we have that wrapper? 

- most of the time we use `db.begin_transaction()` already, only in some cases that wrapper was used
- and I like `begin_transaction()` more as the name is more expressive than `database_transaction`


justinmoon (approved):
LGTM

dpc (approved):

PR #1687, opened 2023-02-15
test: add tests for retry function

mxz42:
Add two tests for the retry function. The first test should run the closure only once as it immediately succeeds. The closure of the second test always fails, hence the closure should be run `max_attempt` times.

### DIFF
--- a/gateway/ln-gateway/src/utils.rs
+++ b/gateway/ln-gateway/src/utils.rs
@@ -43,3 +43,41 @@ where
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::atomic::{AtomicU8, Ordering};
+    use std::time::Duration;
+
+    use anyhow::anyhow;
+
+    use super::retry;
+
+    #[tokio::test]
+    async fn retry_succeed_with_one_attempt() {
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/utils.rs
```diff
@@ -43,3 +43,40 @@ where
...
+    use anyhow::anyhow;
+
+    use super::retry;
+
+    #[tokio::test]
+    async fn retry_one_attempt() {
+        let counter = AtomicU8::new(0);
+        let closure = || async {
+            counter.fetch_add(1, Ordering::SeqCst);
+            Ok(42)
```
m1sterc001guy, line 61: Why 42?

dpc: Because it's the best number?

mxz42: > Because it's the best number?

yeah, it's literally that @m1sterc001guy  
no real reason at all ;D

mxz42: done

dpc (approved):

PR #1688, opened 2023-02-15
fix: apply database migrations for the consensus keys and values

m1sterc001guy:
Through some refactors this code was accidentally dropped from https://github.com/fedimint/fedimint/pull/1567

We need to apply database migrations to the consensus keys/values as well to ensure that these stay up to date with the code, in addition to the modules.

### DIFF
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -35,7 +35,7 @@ use fedimint_server::config::{connect, ServerConfig, ServerConfigParams};
 use fedimint_server::consensus::{
     ConsensusProposal, FedimintConsensus, HbbftConsensusOutcome, TransactionSubmissionError,
 };
-use fedimint_server::db::CONSENSUS_DATABASE_VERSION;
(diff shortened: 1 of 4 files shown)


integrationtests/tests/fixtures/mod.rs
```diff
@@ -1206,6 +1207,15 @@ impl FederationTest {
             let mut modules = BTreeMap::new();
             let env_vars = FedimintConsensus::get_env_vars_map();
 
+            fedimint_core::db::apply_migrations(
+                &db,
+                "Consensus".to_string(),
```
dpc, line 1212: Should it be "consensus" or some kind of "fedimint-core" / "global" or something like that? It's not just consensus data we might migrate. Anything that generic fedimint server is storing might be subject to migration.

m1sterc001guy: Yeah "Global" is probably better

dpc (approved):

justinmoon (approved):

PR #1689, opened 2023-02-15
Add tracing to binaries, make telemetry non-default

### LINKED ISSUE #1493: All binaries should register `tracing-subscriber` with `env` feature and default logging level (`info`?)

jkitman:
Fixes #1493

Some binaries were too trivial for tracing or did not have the dep.

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 use bitcoin::{Address, Amount, Transaction};
 use clap::{Parser, Subcommand};
 use fedimint_core::config::FederationId;
+use fedimint_server::logging::TracingSetup;
 use ln_gateway::config::GatewayConfig;
 use ln_gateway::rpc::rpc_client::RpcClient;
 use ln_gateway::rpc::{
@@ -77,7 +78,9 @@ pub enum Commands {
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
+    TracingSetup::default().init()?;
+
     let cli = Cli::parse();
     let client = RpcClient::new(cli.address);
 
@@ -210,6 +213,8 @@ async fn main() {
             print_response(response).await;
         }
     }
+
+    Ok(())
 }
 
 pub async fn print_response(response: reqwest::Response) {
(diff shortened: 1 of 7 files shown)


dpc (approved):

gateway/cli/src/main.rs
```diff
@@ -77,7 +78,9 @@ pub enum Commands {
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
+    TracingSetup::default().init()?;
```
okjodom, line 82: @jkitman for the gateway, we probably want tracing on the following key binaries:
- [gatewayd](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/bin/gatewayd.rs)
- [cln_extension](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/bin/cln_extension.rs)

The following binaries don't seem that important / are about to be phased out
- [ln_gateway](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/bin/ln_gateway.rs) - to be phased out. superseded by `gatewayd`
- [cli](https://github.com/fedimint/fedimint/blob/master/gateway/cli/src/main.rs) - current file. Just a cli tool

okjodom (dismissed):
approved, with suggestions

gateway/cli/src/main.rs
jkitman: Can't do plugins, have to use the env var, but added to gatewayd.

okjodom (approved):

gateway/cli/src/main.rs
okjodom: ack. `cln_extension` plugs in to CLNs logging system

dpc (approved):

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -10,6 +10,7 @@ use fedimint_core::core::{
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::module::DynModuleGen;
 use fedimint_core::task::TaskGroup;
+use fedimint_server::logging::TracingSetup;
```
elsirion, line 13: Separate issue since the dep was already there, but `gatewayd` should not depend on `fedimint-server` imo.

Issue #1635, opened 2023-02-10
Fix issues with `DatabaseTransaction:commit_tx` taking `self` by move.


dpc:
I remember there were some patterns that were not possible due to that (maybe @Maan2003 remembers).

Two possible (somewhat orthogonal) things we could do:

First, make `commit_tx` take `&mut self` in the trait, but keep it `self` on the `DatabaseTransaction` newtype wrapper. 

This would allow us to always wrap inner `IDatabaseTranasction` via `&mut dyn IDatabaseTranasction`, avoiding allocations. Under the hood the implementations often need to consume the transaction (that's why we made `commit_tx` take `self` by move. But they could use `Option` and panic if called twice (the dyn newtype will ensure this can't happen anyway), or even use `mem::replace` to replace the old tx with a new one (if this doesn't cost us any performance) 

Second, split the `commit_tx` into `trait IDatabaseTransaction : IDatabaseTransactionRef`. There are some places where we explicitily don't want the user of `DatabaseTransaction` (module code) to be allowed to call `commit_tx` so this is probably necessary as well. Maybe just doing this part will make the first one unnecessary.

m1sterc001guy:
> Second, split the commit_tx into trait IDatabaseTransaction : IDatabaseTransactionRef. There are some places where we explicitily don't want the user of DatabaseTransaction (module code) to be allowed to call commit_tx so this is probably necessary as well. Maybe just doing this part will make the first one unnecessary.

If we do this, I think we need to introduce a new wrapper type. Right now, both "Isolated" db transactions and "Commitable" transactions are the same wrapper type (`DatabaseTransaction`). Isolated transactions just panic if `commit_tx` is called. If we want to prevent the modules from ever committing, I think we'd need a new wrapper type that didn't have `commit_tx` as a function option.

dpc:
> If we do this, I think we need to introduce a new wrapper type.

Yes.

Maan2003:
usage: avoid extra box in https://github.com/fedimint/fedimint/pull/1197/files#diff-b8d3e00d0239ca0f82aaa8fb95c1b63eb2cce27f2108f6c10457acdeb06d0138R107

PR #1197, opened 2023-01-04
Remove polling

Maan2003:
Remain tasks:
- [x] debug test running forever
- [x] unit tests
- [x] docs
- [x] integrating task groups
---

Currently, we use polling to check for updates(say "fetch_transaction").

What we need is to subscribe for change in web server task, and push notification for these changes in the main task.
But we can't immediately notify the web server task because the transaction is not yet committed. And the database won't contain the value. 

So mainly I see two solution:
- have "post commit hooks", function that get called after transaction is committed. But then we don't access to the state of the module.
- build notifications into the database using database keys, which is much more clean.

Another Attempt #273

PR #1695, opened 2023-02-16
IDatabaseTransaction taking self by move during commit

m1sterc001guy:
This PR attempts to fix https://github.com/fedimint/fedimint/issues/1635

Right now additional allocations need to be taken because `commit_tx` takes `self` by move and not by `&mut` like the other functions in the trait. This PR introduces a new trait that contains an `Option` of the database transaction and consumes it during commit, and panics if the transaction is `None`. Additional wrapper was also introduced to accomodate the new trait.

This will directly conflict with https://github.com/fedimint/fedimint/pull/1197

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -70,7 +70,7 @@ pub type PrefixStream<'a> = Pin<Box<dyn Stream<Item = (Vec<u8>, Vec<u8>)> + Send
 
 #[async_trait]
 pub trait IDatabase: Debug + Send + Sync {
-    async fn begin_transaction<'a>(&'a self) -> Box<dyn IDatabaseTransaction<'a>>;
+    async fn begin_transaction<'a>(&'a self) -> Box<dyn ISingleUseDatabaseTransaction<'a>>;
 }
 
 #[derive(Clone, Debug)]
@@ -271,7 +271,7 @@ pub trait IDatabaseTransaction<'a>: 'a + Send {
         Ok(())
     }
 
-    async fn commit_tx(self: Box<Self>) -> Result<()>;
+    async fn commit_tx(self) -> Result<()>;
 
     async fn rollback_tx_to_savepoint(&mut self);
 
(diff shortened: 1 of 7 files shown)


m1sterc001guy:
I need to squash commits and add documentation for the new trait

fedimint-core/src/db/mod.rs
```diff
@@ -271,7 +271,34 @@ pub trait IDatabaseTransaction<'a>: 'a + Send {
...
+    /// was created.
+    ///
+    /// Warning: Avoid using this in fedimint client code as not all database
+    /// transaction implementations will support setting a savepoint during
+    /// a transaction.
+    async fn set_tx_savepoint(&mut self);
+}
+
+#[async_trait]
+pub trait IDatabaseTransaction<'a>: 'a + Send {
```
dpc, line 290: I though we're going to `trait IDatabaseTransaction: IDatabaseTransactionNoCommit` or something like that.

dpc: One way or another `Core` doesn't tell much.

fedimint-core/src/db/mod.rs
```diff
@@ -286,6 +313,79 @@ pub trait IDatabaseTransaction<'a>: 'a + Send {
     async fn set_tx_savepoint(&mut self);
 }
 
+pub struct IDatabaseTransactionWrapper<'a, Tx: IDatabaseTransactionCore<'a> + Send>(
+    Option<Tx>,
```
dpc, line 317: `Wrapper` also not very informative.

And it shouldn't be named `*I*Database...` as it is no longer an interface/trait?

`SingleUseDatabaseTransaction<Tx>` maybe?

fedimint-core/src/db/mod.rs
```diff
@@ -286,6 +313,79 @@ pub trait IDatabaseTransaction<'a>: 'a + Send {
...
+}
+
+#[async_trait]
+impl<'a, Tx: IDatabaseTransactionCore<'a> + Send> IDatabaseTransaction<'a>
+    for IDatabaseTransactionWrapper<'a, Tx>
+{
+    async fn raw_insert_bytes(&mut self, key: &[u8], value: Vec<u8>) -> Result<Option<Vec<u8>>> {
+        match &mut self.0 {
+            Some(tx) => tx.raw_insert_bytes(key, value).await,
+            None => panic!("Cannot insert into already consumed transaction"),
```
dpc, line 334: I wonder if it wouldn't be better to return an error instead...

fedimint-core/src/db/mod.rs
dpc: Ah, this not about lack of commit, but `&mut self` vs `self`.

fedimint-core/src/db/mod.rs
m1sterc001guy: Changed to return error where possible.

fedimint-core/src/db/mod.rs
m1sterc001guy: Ugh yeah, was struggling to come up with the name. I like `SingleUseDatabaseTransaction`. Here's what I renamed everything to:

`IDatabaseTransactionCore` -> `IDatabaseTransaction`
`IDatabaseTransaction` -> `ISingleUseDatabaseTransaction`
`IDatabaseTransactionWrapper` -> `SingleUseDatabaseTransaction`

Hopefully that's a bit more clear.

fedimint-core/src/db/mod.rs
m1sterc001guy: Yes this only handles `&mut self` vs `self`.

We can still split the trait up into `IDatabaseTransaction` and `IDatabaseTransactionCommit` if we want, but I worry introducing more traits will just make it more confusing. Perhaps the better solution is just to introduce the new public wrapper struct that can't commit and give that to the modules. I think this pretty much accomplishes the same goal, but I'm open to suggestions.

fedimint-core/src/db/mod.rs
```diff
@@ -286,6 +286,115 @@ pub trait IDatabaseTransaction<'a>: 'a + Send {
     async fn set_tx_savepoint(&mut self);
 }
 
+/// `ISingleUseDatabaseTransaction` re-defines the functions from
+/// `IDatabaseTransaction` but does not consumed `self` when committing to the
+/// database. This allows for wrapper structs to more easily borrow
+/// `ISingleUseDatabaseTransaction` without needing to make additional
+/// allocations.
+#[async_trait]
+pub trait ISingleUseDatabaseTransaction<'a>: 'a + Send {
```
Maan2003, line 295: why is this `pub`?

fedimint-core/src/db/mod.rs
```diff
@@ -286,6 +286,115 @@ pub trait IDatabaseTransaction<'a>: 'a + Send {
...
+#[async_trait]
+impl<'a, Tx: IDatabaseTransaction<'a> + Send> ISingleUseDatabaseTransaction<'a>
+    for SingleUseDatabaseTransaction<'a, Tx>
+{
+    async fn raw_insert_bytes(&mut self, key: &[u8], value: Vec<u8>) -> Result<Option<Vec<u8>>> {
+        match &mut self.0 {
+            Some(tx) => tx.raw_insert_bytes(key, value).await,
+            None => Err(anyhow::anyhow!(
+                "Cannot insert into already consumed transaction"
+            )),
```
Maan2003, line 335: try `self.0.as_mut().context(...).func(....)`

context is from `anyhow::Context`

fedimint-core/src/db/mod.rs
```diff
@@ -358,7 +467,14 @@ impl<'a> IDatabaseTransaction<'a> for ModuleDatabaseTransaction<'a> {
...
 
-    async fn commit_tx(self: Box<Self>) -> Result<()> {
+    async fn raw_remove_by_prefix(&mut self, key_prefix: &[u8]) -> Result<()> {
+        self.dbtx
+            .with_module_prefix(self.prefix)
+            .raw_remove_by_prefix(key_prefix)
+            .await
+    }
+
+    async fn commit_tx(self) -> Result<()> {
```
Maan2003, line 477: all internal wrapper would work with and implement `ISingleUseDatabaseTransaction` directly, IMO

fedimint-core/src/db/mod.rs
```diff
@@ -467,13 +587,13 @@ impl<'isolated, 'parent, T: Send + Encodable + 'isolated> IDatabaseTransaction<'
 
 #[doc = " A handle to a type-erased database implementation"]
 pub struct DatabaseTransaction<'a> {
-    tx: Box<dyn IDatabaseTransaction<'a>>,
+    tx: Box<dyn ISingleUseDatabaseTransaction<'a>>,
```
Maan2003, line 590: can we make this `&'a mut ISingleUseDatabaseTransaction<'a>` and save some boxing?

fedimint-core/src/db/mod.rs
dpc: Crates like `fedimint-rocksdb` use it wrap their normal implementation with this "one time commit"-guard.

fedimint-core/src/db/mod.rs
dpc: Yeah, I'm also wondering if this wouldn't be simpler. A bit more redundancy, but one less thing in a common code. Some db trait implementation crates might not even need an `Option`.

fedimint-core/src/db/mod.rs
dpc: Ideally this would be `tx: &dyn IDatabaseTransaction<'a>,`. Here we have no need to know about the wrapper anymore, no?

fedimint-core/src/db/mod.rs
Maan2003: ah the `fn begin_transaction`

fedimint-core/src/db/mod.rs
Maan2003: IDatabaseTransaction still has `commit(self)` right?

m1sterc001guy: So, I think the issue is that `DatabaseTransaction` needs to hold on to the actual boxxed `ISingleDatabaseTransaction`, otherwise (when holding onto a reference) we are creating the `DatabaseTransaction` and returning a value referencing a temporary object.

I think this can be worked around if we introduce a new public wrapper type instead of reusing `DatabaseTransaction` for everything. 

Example with bad name: `NewModuleDatabaseTransaction` could be the db transaction we pass to the modules, it would not have a `commit_tx` function and it would hold a `&mut dyn ISingleUseDatabaseTransaction` that referenced the existing db transaction. `DatabaseTransaction` would stay the way it is right now and always hold a `Box`.

Perhaps I can do this in a different PR if you guys agree.

fedimint-core/src/db/mod.rs
m1sterc001guy: Changed all wrappers to implement `ISingleUseDatabaseTransaction`

fedimint-core/src/db/mod.rs
m1sterc001guy: Fixed

dpc (dismissed):

fedimint-core/src/db/mod.rs
```diff
@@ -437,43 +541,47 @@ impl<'isolated, 'parent, T: Send + Encodable + 'isolated> IDatabaseTransaction<'
...
     decoders: ModuleDecoderRegistry,
     commit_tracker: CommitTracker,
 }
 
 impl<'a> std::ops::Deref for DatabaseTransaction<'a> {
-    type Target = dyn IDatabaseTransaction<'a>;
+    type Target = dyn ISingleUseDatabaseTransaction<'a>;
```
Maan2003, line 584: this is not a good idea. This adds `DatabaseTransaction::commit` which takes `&mut self` and leaves DatabaseTransaction in unusable state. 
`DatabaseTransaction` should implement its own `commit(self)`

m1sterc001guy: Removed the deref altogether

dpc (approved):

Maan2003 (approved):

PR #1698, opened 2023-02-16
fix: don't fail when external websites are down

dpc:
One of the websites we link to had a temporary problem, and I don't want this to block our PRs.

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -5,6 +5,7 @@ name: Markdown Link Check
 jobs:
   markdown-link-check:
     runs-on: ubuntu-22.04
+    continue-on-error: true # can fail for external reasons, so don't be a blocker
     steps:
       - uses: actions/checkout@v3
       - uses: gaurav-nelson/github-action-markdown-link-check@v1


justinmoon (approved):

.github/workflows/linkcheck.yml
```diff
@@ -5,6 +5,7 @@ name: Markdown Link Check
 jobs:
   markdown-link-check:
     runs-on: ubuntu-22.04
+    continue-on-error: true # can fail for external reasons, so don't be a blocker
```
mxz42, line 8: the external reason would be that a website returns a 404 error? 

could it fail for a "internal" reason as well?

mxz42 (approved):

.github/workflows/linkcheck.yml
dpc: I can't see a way to differentiate between these two. When the link is dead there's no way to tell if the link is wrong, or the server temporary down.

IIUC this change will mark the job as passed, so we'll see in the UI that there are some bad links, but it won't block the CI on it.

mxz42: ok, sounds good

PR #1699, opened 2023-02-17
fix: `SerdeEncodable` should impl `Serialize` and `Deserialize`

dpc:
Sometimes it's necessary to use type using it in a context that needs to be `Serialize`&`Deserialize`.

### DIFF
--- a/fedimint-core/src/encoding/mod.rs
+++ b/fedimint-core/src/encoding/mod.rs
@@ -15,6 +15,7 @@ use anyhow::format_err;
 use bitcoin_hashes::sha256::HashEngine;
 use bitcoin_hashes::{sha256, Hash};
 pub use fedimint_derive::{Decodable, Encodable, UnzipConsensus};
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use url::Url;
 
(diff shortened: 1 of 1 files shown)


jkitman (approved):
Can you provide a bit more context on why?

dpc:
> Can you provide a bit more context on why?

Sometimes it's necessary to use type using it in a context that needs to be `Serialize`&`Deserialize`.

fedimint-core/src/encoding/mod.rs
```diff
@@ -532,7 +533,7 @@ where
 }
 /// Wrappers for `T` that are `De-Serializable`, while we need them in
 /// `Encodable` contex
-#[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Ord, Eq, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Ord, Eq, Hash, Serialize, Deserialize)]
 pub struct SerdeEncodable<T>(pub T);
```
Maan2003, line 537: #[serde(tranparent)]
but it doesn't matter for json

Maan2003 (approved):

PR #1700, opened 2023-02-17
Simplify config gen and fix error handling

jkitman:
Simplifies the code so that users construct a `ServerConfigParams` which gets passed into config gen, rather than having `run_dkg` generate it.


fedimint-server/src/config/mod.rs
```diff
@@ -449,28 +452,20 @@ impl ServerConfig {
...
-                        module_instance_id,
-                        peers,
-                        &params.modules,
-                        task_group,
-                    )
-                    .await?
-                {
-                    cfgs
-                } else {
-                    return Ok(Err(Cancelled));
```
dpc, line 472: We used to this custom handling, so a canceled operation is not considered an error and user pressing ctrl+c does not lead to panics and/or screen full of errors.

fedimint-server/src/config/mod.rs
```diff
@@ -519,7 +513,9 @@ impl ServerConfig {
             "Distributed key generation has completed successfully!"
         );
 
-        Ok(Ok(server))
+        drop(connections);
```
dpc, line 516: Seems unnecessary. It will be dropped at the end of the scope anyway.

dpc (commented):
Roughly LGTM, but comments about `Cancellable`.

fedimint-server/src/config/mod.rs
jkitman: Okay, wonder why this was in the original code...

fedimint-server/src/config/mod.rs
jkitman: We assume a failure in the middle of a DKG run is a cancellation?  Doesn't this swallow the underlying error if there is one?

jkitman: Can we make a determination whether or not to display an error based on the `TaskGroup`?

fedimint-server/src/config/mod.rs
```diff
@@ -610,6 +606,43 @@ impl ServerConfigParams {
...
+
+        let key = get_key(password, dir_out_path.join(SALT_FILE))?;
+        let tls_pk = encrypted_read(&key, dir_out_path.join(TLS_PK))?;
+        let cert_string = fs::read_to_string(dir_out_path.join(TLS_CERT))?;
+
+        let our_params = parse_peer_params(cert_string)?;
+        let our_id = peers
+            .iter()
+            .find(|(_peer, params)| params.cert == our_params.cert)
+            .map(|(peer, _)| *peer)
```
mxz42, line 632: nit, feel free to ignore
```rust
            .find_map(|(&peer, params)| {
               if params.cert == our_params.cert {
                   Some(peer)
               } else {
                   None
               }
```

mxz42 (approved):
AFAIU, looks good :+1:

fedimint-server/src/config/io.rs
```diff
@@ -1,27 +1,20 @@
...
 
-use crate::config::{
-    connect, gen_cert_and_key, PeerServerParams, ServerConfig, ServerConfigParams,
-};
-use crate::fedimint_core::net::peers::IMuxPeerConnections;
-use crate::multiplexed::PeerConnectionMultiplexer;
+use crate::config::{gen_cert_and_key, PeerServerParams, ServerConfig};
+
+/// Version of the server code (should be the same among peers)
+pub const CODE_VERSION: &str = env!("GIT_HASH");
```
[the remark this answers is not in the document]
dpc: Not sure if this is a job for this PR, but indeed we want to switch to version & consensus comparisons.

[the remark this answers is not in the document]
jkitman: Perhaps we should rename to `CODE_VERSION`

fedimint-server/src/config/mod.rs
dpc: `Cancellable<Result<X>>` makes a distinction between operation being cancelled via ctrl+c : `Err(Cancelled)` or hitting an error `OK(Err(...))`. If it was cancelled we are not supposed to print errors, etc. If it failed, we handle like normally we treat errors.

It's a bit inconvenient, because we have to "unpeel" the inner error in calling places, but it forces handling it correctly, and there isn't many places we need to take care of it.

jkitman: But as far as I can tell we just return `Cancelled` if `run_g1` ever returns an error, regardless of whether this was a cancellation or an actual error with DKG.

jkitman: Perhaps we should spawn every instance of DKG as a separate subtask and swallow the errors if the task is cancelled?

dpc: > But as far as I can tell we just return Cancelled if run_g1 ever returns an error, regardless of whether this was a cancellation or an actual error with DKG.

Possibly it got broken during recent refactoring, or I didn't do it correctly there when I first introduced it.

> Perhaps we should spawn every instance of DKG as a separate subtask and swallow the errors if the task is cancelled?

Swallowing errors leads to hard to debug issues when actual errors actually happen. Treating cancellation as an error leads to bad UX.

The only way to handle it robustly is to make a distinction between cancellation and an error (which is what `Cancellable` is about) and handle it properly everywhere.

jkitman: I see...I think it's not working properly now, let me try to fix.

dpc (approved):

justinmoon (approved):

PR #1710, opened 2023-02-17
fix: Improve tmuxinator environment variable filtering

justinmoon:
I had an environment variable with `FM_` in the middle of the value, which caused an "invalid variable" error. Instead, we look for `FM_` only at the beginning of an environment variable.

See log https://gist.github.com/justinmoon/2480ba4fa8ce486e60d610f6e3a04fe0 for full error

### DIFF
--- a/scripts/tmuxinator.sh
+++ b/scripts/tmuxinator.sh
@@ -18,7 +18,7 @@ export FM_VERBOSE_OUTPUT=0
 source scripts/build.sh
 echo "Running in temporary directory $FM_TEST_DIR"
 
-env | sed -En 's/(FM_[^=]*).*/\1/gp' | while read var; do printf 'export %s=%q\n' "$var" "${!var}"; done > .tmpenv
+env | sed -En 's/^(FM_[^=]*).*/\1/gp' | while read var; do printf 'export %s=%q\n' "$var" "${!var}"; done > .tmpenv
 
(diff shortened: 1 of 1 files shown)


scripts/tmuxinator.sh
```diff
@@ -18,7 +18,7 @@ export FM_VERBOSE_OUTPUT=0
 source scripts/build.sh
 echo "Running in temporary directory $FM_TEST_DIR"
 
-env | sed -En 's/(FM_[^=]*).*/\1/gp' | while read var; do printf 'export %s=%q\n' "$var" "${!var}"; done > .tmpenv
+env | sed -En 's/(^FM_[^=]*).*/\1/gp' | while read var; do printf 'export %s=%q\n' "$var" "${!var}"; done > .tmpenv
```
dpc, line 21: I wonder if there's a difference if `^` is in front or inside the `(...)`.

dpc (dismissed):

scripts/tmuxinator.sh
justinmoon: Probably safer to put it outside?

justinmoon: Put it outside

dpc (approved):

justinmoon:
If https://github.com/fedimint/fedimint/pull/1711 works for everyone, this is no longer needed

elsirion (approved):

PR #1711, opened 2023-02-17
fix: remove .tmpenv file

justinmoon:
Tmuxinator works just fine without this file ***for me***. Seems that environment is passed to the tmuxinator environment by default?

### DIFF
--- a/scripts/tmuxinator.sh
+++ b/scripts/tmuxinator.sh
@@ -18,13 +18,9 @@ export FM_VERBOSE_OUTPUT=0
 source scripts/build.sh
 echo "Running in temporary directory $FM_TEST_DIR"
 
-env | sed -En 's/(FM_[^=]*).*/\1/gp' | while read var; do printf 'export %s=%q\n' "$var" "${!var}"; done > .tmpenv
-
 export FM_BITCOIND_RPC="http://bitcoin:bitcoin@127.0.0.1:18443"
 
 SHELL=$(which bash) tmuxinator local
 tmux -L fedimint-dev kill-session -t fedimint-dev || true
 pkill bitcoind
 pkill lightningd
-
-rm .tmpenv
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,5 @@ ln1
 ln2
 it
 .idea
-.tmpenv
 .direnv
 .vscode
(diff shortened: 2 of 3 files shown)


dpc:
Seems like this was used to pass only `FM_*` environment variables into tmuxinator. I think  @elsirion introduced it, so he might know how useful it is.

scripts/tmuxinator.sh
```diff
@@ -18,13 +18,9 @@ export FM_VERBOSE_OUTPUT=0
 source scripts/build.sh
 echo "Running in temporary directory $FM_TEST_DIR"
 
-env | sed -En 's/(FM_[^=]*).*/\1/gp' | while read var; do printf 'export %s=%q\n' "$var" "${!var}"; done > .tmpenv
-
 export FM_BITCOIND_RPC="http://bitcoin:bitcoin@127.0.0.1:18443"
```
mxz42, line 21: if I compare the `.tmpenv` file with the env variables set in the tmux session (`env | rg '^FM_.*'`) then there is also this `FM_BITCOIND_RPC` variable (which is not in .tmpenv).

mxz42:
So I think this works for me, too. The only problem (but not related to our tmuxinator setup) I have sometimes with my tmux setup is that if I change global env variables in my shell, they are not propagated into tmux; even if I `pkill tmux` and restart it.

elsirion:
I think the propagation did not work for me back then, but possibly because of NixOS weirdness? 

When trying it now if fails with a lot of weird errors (first `bitcoind` didn't start, now the wallet isn't loaded), why would sourcing the `.tmpenv` file have *any* such side effects?! 

I'm testing in a relatively bare-bones NixOS container with `tmux` installed in the user env iirc. If anyone with NixOS could reproduce that would be awesome:

```nix
{ config, lib, pkgs, ... }:

with lib;

{ boot.isContainer = true;
  networking.hostName = mkDefault "fedimint";
  networking.useDHCP = false;
  
  bindMounts."/home/user/fedimint" = {
    hostPath = "/home/user/projects/fedimint";
    isReadOnly = false;
  };
}
```

justinmoon:
> When trying it now if fails with a lot of weird errors (first bitcoind didn't start, now the wallet isn't loaded) ...

Just to be clear, this happens when you run `scripts/tmuxinator.sh` on this PR?

elsirion:
Well, `just tmuxinator`, but yes.

justinmoon:
OK. Thinking it's probably best to just close this. What we have works for everyone and this doesn't.

PR #1712, opened 2023-02-17
Move module parameterization to fedimintd

jkitman:


fedimint-core/src/config.rs
```diff
@@ -231,29 +228,42 @@ impl ClientConfig {
 ///
 /// The same `ModuleKind` may have multiple instances with different settings
 #[derive(Debug, Clone, Default)]
-pub struct ConfigGenParams(BTreeMap<String, serde_json::Value>);
+pub struct ConfigGenParams {
+    modules: BTreeMap<ModuleInstanceId, JsonWithKind>,
+    pub registry: ModuleGenRegistry,
```
dpc, line 233: This being here doesn't look correct. Convenient maybe somewhere, but how is "module gen registry" a part of "config gen params"?

fedimint-core/src/config.rs
```diff
@@ -231,29 +228,42 @@ impl ClientConfig {
...
+            .iter()
+            .map(|(id, kind)| (*id, self.registry.get(kind.kind()).expect("Exists").clone()))
+            .collect()
+    }
+
+    /// Add a module with configuration parameters
     pub fn attach<P: ModuleGenParams>(mut self, module_params: P) -> Self {
-        self.0.insert(
-            P::MODULE_NAME.to_string(),
+        let next_id = self.modules.len() as u16;
```
dpc, line 250: BTW. module instance id don't have to be continuous. In the future, we might allow federation to turn of a certain module and leave it without anything under that ID.

fedimint-core/src/module/mod.rs
```diff
@@ -292,6 +292,7 @@ pub trait IModuleGen: Debug {
     fn trusted_dealer_gen(
         &self,
         peers: &[PeerId],
+        module_id: ModuleInstanceId,
```
dpc, line 295: Modules were not supposed to touch their ids (and be oblivious to them, just in case).

fedimintd/src/bin/distributedgen.rs
```diff
@@ -162,16 +162,13 @@ async fn main() -> anyhow::Result<()> {
...
-                {
-                    Ok(server) => server,
-                    Err(DkgError::Cancelled(_)) => return Ok(info!("DKG cancelled")),
-                    Err(DkgError::Failed(err)) => return Err(err),
-                };
-
-            write_server_config(&server, dir_out_path, &password, &module_registry())
+            let server = match ServerConfig::distributed_gen(&params, &mut task_group).await {
+                Ok(server) => server,
+                Err(DkgError::Cancelled(_)) => return Ok(info!("DKG cancelled")),
```
dpc, line 167: LOL. Putting that `info!(...)` in there feels like cheating. :D

fedimintd/src/lib.rs
```diff
@@ -2,31 +2,33 @@ use bitcoin::Network;
...
+        .attach(LightningGenParams)
         .attach(MintGenParams {
             mint_amounts: Tiered::gen_denominations(max_denomination)
                 .tiers()
                 .cloned()
                 .collect(),
         })
+        .attach(WalletGenParams {
+            network,
+            // TODO this is not very elegant, but I'm planning to get rid of it in a next commit
```
dpc, line 28: BTW. "the next commit" did not materialize. :D

modules/fedimint-mint/src/lib.rs
```diff
@@ -179,9 +179,12 @@ impl ModuleGen for MintGen {
     fn trusted_dealer_gen(
         &self,
         peers: &[PeerId],
+        module_id: ModuleInstanceId,
         params: &ConfigGenParams,
     ) -> BTreeMap<PeerId, ServerModuleConfig> {
-        let params = params.get::<MintGenParams>().expect("Invalid mint params");
+        let params = params
+            .get::<MintGenParams>(&module_id)
```
dpc, line 186: The module should get its own params, and not the whole map of params and it's id to look it up.

fedimint-core/src/config.rs
```diff
@@ -231,29 +228,42 @@ impl ClientConfig {
 ///
 /// The same `ModuleKind` may have multiple instances with different settings
 #[derive(Debug, Clone, Default)]
-pub struct ConfigGenParams(BTreeMap<String, serde_json::Value>);
```
dpc, line 234: The idea here was that in the UI configuration we might not always have all the available fields (because there will be module we didn't even know about) etc. exposed, so we should have some form of free-form

```
some_key1: some_json_value1,
some_key2: some_json_value2,
...
```

configuration that user would be able to provide per-module.

So `ConfigGenParams` was already per-module-instance and `String` was just one key that user might give to the module.


`registry` really doesn't belong here, and then having the module look up it's own config is very undesireable.

dpc: The `IModuleGen::trusted_dealer_gen` can take `ConfigGenParams` as `BTreeMap<String, serde_json::Value>`, but `ModuleGen::trusted_dealer_gen` (static typed version) could have `type ConfigGenParams` and then:

```
#[async_trait]
impl<T> IModuleGen for T
where
    T: ModuleGen + 'static + Sync,
{
```

would cast the untyped `BTreeMap<String, serde_json::Value>` to typed `ModuleGen::ConfigGenParams`, hiding the typed-erased json from the module author.

fedimint-core/src/config.rs
jkitman: Doesn't have to be but it ensures every configured module is also registered, so the developer only needs to do it in one place and doesn't accidentally miss registering a module.

fedimint-core/src/config.rs
jkitman: Makes sense, ordering them seemed the most simple, plus they need to be consistent with the client for now.

fedimint-core/src/module/mod.rs
jkitman: But we pass it into `distributed_config_gen` for each module?

fedimintd/src/bin/distributedgen.rs
jkitman: Haha copying a pattern from elsewhere in the code, not sure who did it, might have been me.

fedimintd/src/lib.rs
jkitman: Yeah what was the context on the comment?

modules/fedimint-mint/src/lib.rs
jkitman: So pass in the JSON value instead?  This is how we were handling it in distributed_gen.

fedimint-core/src/config.rs
jkitman: I thought `ModuleInstanceId` identified each module instance, not a `String`?

fedimint-core/src/config.rs
dpc: This can surely be achieved in some other ways. Again. This `struct` used to mean "extra arguments for a single module instance", and now without changing the name (or even how it used) it is working completely differently.

fedimint-core/src/config.rs
dpc: That map here was a map of 0 or more values, keyed on arbitrary string. So e.g. if someone added a `FederatedFilesystemModule` in the future, our UI would not know about it, so we would have no nice configuration boxes for it, but yet the user could add configuration manually for that particular module instance they are configuring in the config/UI and type in:

```
storage: "aws"
aws_credentials:
   token_id: ASDFADFADSF
   token_secret: j8j9sjfojasdodfjksadjf
```

and the whole thing would be put into the `ConfigGenParams` as a "map" and passed as a value to the `FederatedFilesystemModuleGen::distributed_gen` for that module.

The `Sting` in that map is a key: `storage` and `aws_credentials` in this case. We could make it into one big `serde_json::Value`, I supposed, but we never really implemented any of this, and it seems it confused everyone that looked at it.

fedimint-core/src/module/mod.rs
dpc: Yes. That what `ConfigGenParams` was supposed to be. (see other comment)

fedimint-core/src/config.rs
jkitman: So it sounds like what we need is something to store `ModuleInstanceId` -> `ConfigGenParams`?

And then each module instance gets its own `ConfigGenParams`?  Not sure that `ConfigGenParams` needs to be a map, it could just be a `JsonWithKind`.

dpc: > So it sounds like what we need is something to store `ModuleInstanceId` -> `ConfigGenParams`?

Possibly, yes.

> And then each module instance gets its own `ConfigGenParams`? 

Yes.

> Not sure that `ConfigGenParams` needs to be a map, it could just be a `JsonWithKind`.

Possibly, yes.  There are some considerations. `serde_json::Value` possibly being an array, null etc, which makes less sense, but whatever. It's not really used for anything yet, so we can play with it once we do.

PR #1713, opened 2023-02-18
fix: unify cross and native package builds

dpc:
This fixes `nix build .#client-pkgs`

I had renamed the cross compilation target from made up "shot-names" to just full triple the compiler actually uses. It was just getting in away and confusing.

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -118,9 +118,9 @@ jobs:
     strategy:
       matrix:
         build:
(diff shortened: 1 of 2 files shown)


elsirion (approved):

dpc:
@elsirion looks like that due to it changing the workflow jobs it will not pass CI and will need admin merge. So while at it, I'll add a rename of one job there.

elsirion (approved):

Issue #613, opened 2022-09-22
Federation facilitates internal payments w/o involving gateway


justinmoon:
https://github.com/fedimint/fedimint/issues/461 implemented basic internal payments support by outsourcing it to the gateway. Perhaps we could have the sender buy the preimage directly from the federation?

<details>
  <summary>Edit: Long, outdated original post</summary>
https://github.com/fedimint/fedimint/issues/461 implemented basic internal payments support by outsourcing it to the gateway. But the gateway still charges a fee without really providing any service that the federation couldn't itself, which is worse from the user POV. 

Not sure if / when we want to implement this, but wanted to write down how we could do it just to share the knowledge:

1. Recipient creates an invoice. This basically sets up an account in the federation that they will be able to sweep once funds are deposited to it.
2. Recipient sends invoice to sender
3. Sender attempts to pay invoice. First step is depositing ecash into an `OutgoingContract` in the federation which is spendable by anyone who presents the preimage corresponding for this invoice. At this point the user would normally tell the gateway about this escrow contract and the gateway would go buy the preimage over lightning. But if the recipient's account is also in the federation, then the federation itself could just settle between the two accounts here. Notes:
  - It would do this by checking the database [here](https://github.com/fedimint/fedimint/blob/2e4a1d48cd1d0d659b6a71659e4b36bae36e79a4/modules/fedimint-ln/src/lib.rs#L317) for an `IncomingContract` with the same `hash` parameter as the `OutgoingContract`. If such a contract exists, deposit `invoice.amount` into the `IncomingContract` and `ContractOutput.amount - invoice.amount` to `ContractAccount` which the sender can sweep immediately. Otherwise, proceed as we currently do by funding the outgoing account. This way the federation wouldn't make any profit (which complicates things) and user wouldn't unnecessarily pay fees to the gateway.
  - Whatever fee the sender had deposited to the outgoing escrow account in excess of the invoice's amount could remain in this account and sender could sweep it back. This alleviates a previous concern voiced by @Kodylow and others that having the federation settle internally could lead to it making profit which just complicates things right now.
  - Clients that pay invoices (e.g. [ln-pay cli command](https://github.com/fedimint/fedimint/blob/2e4a1d48cd1d0d659b6a71659e4b36bae36e79a4/client/cli/src/main.rs#L454-#L481)) would need to be able to recognize that internal settlement happened. Before contacting the gateway perhaps they could check to see if their outgoing account already had `invoice.amount` deducted? Something like that ...
[...]

PR #1717, opened 2023-02-18
Use new gateway code in cli + rust tests

justinmoon:
In place of the existing `ln_gateway`, we integrated `gatewayd` + `gateway_cln_extension` to the existing CLI and rust tests. This implies the new gateway architecture has parity of test coverage as we had on the old gateway, and we can proceed to remove the old gateway at will.

Next comes #1718

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -28,14 +28,12 @@ windows:
           - fg
         - ln1:
           - sleep 5 # wait for bitcoind and federation
-          - gateway-cli generate-config 127.0.0.1:8175 'http://127.0.0.1:8175' $FM_CFG_DIR #generate gateway config
-          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN1_DIR --addr=127.0.0.1:9000 --plugin=$FM_BIN_DIR/ln_gateway --fedimint-cfg=$FM_CFG_DIR &
+          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN1_DIR --addr=127.0.0.1:9000 --plugin=$FM_BIN_DIR/gateway-cln-extension &
           - echo $! >> $FM_PID_FILE
           - fg
         - ln2:
           - sleep 5 # wait for bitcoind and federation
-          - export GW_CLN_EXTENSION_LISTEN_ADDRESS=0.0.0.0:54321
-          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN2_DIR --addr=127.0.0.1:9001 --plugin=$FM_BIN_DIR/gateway-cln-extension &
+          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN2_DIR --addr=127.0.0.1:9001
           - echo $! >> $FM_PID_FILE
           - fg
         - gateway:
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -77,13 +77,13 @@ where
     ))
 }
 
-// TODO: Keep in sync with cln-plugin:
+// TODO: upstream these structs to cln-plugin
 // See: https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md#htlc_accepted
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Htlc {
-    pub short_channel_id: u64,
     #[serde(deserialize_with = "as_fedimint_amount")]
     pub amount_msat: Amount,
+    // TODO: use these to validate we can actually redeem the HTLC in time
     pub cltv_expiry: u32,
     pub cltv_expiry_relative: u32,
     pub payment_hash: bitcoin_hashes::sha256::Hash,
@@ -92,7 +92,7 @@ pub struct Htlc {
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Onion {
     #[serde(default)]
-    pub short_channel_id: Option<u64>,
+    pub short_channel_id: Option<String>,
     #[serde(deserialize_with = "as_fedimint_amount")]
     pub forward_msat: Amount,
 }
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -38,7 +38,7 @@ use tracing::{debug, error, trace, warn};
 #[derive(Parser)]
 pub struct ClnExtensionOpts {
     /// Gateway CLN extension service listen address
-    #[arg(long = "listen", env = "GW_CLN_EXTENSION_LISTEN_ADDRESS")]
+    #[arg(long = "listen", env = "FM_CLN_EXTENSION_LISTEN_ADDRESS")]
     pub listen: SocketAddr,
 }
 
@@ -153,7 +153,7 @@ impl ClnRpcService {
                     Some(options::Value::String(listen)) => {
                         if listen == "default-dont-use" {
                             panic!(
-                                "Gateway cln extension is missing a listen address configuration. You can set it via GW_CLN_EXTENSION_LISTEN_ADDRESS env variable, or by adding a --listen config option to the cln plugin"
+                                "Gateway cln extension is missing a listen address configuration. You can set it via FM_CLN_EXTENSION_LISTEN_ADDRESS env variable, or by adding a --listen config option to the cln plugin"
                             )
                         } else {
                             SocketAddr::from_str(&listen).expect("invalid listen address")
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -507,7 +507,14 @@ impl ClnHtlcInterceptor {
         let htlc_expiry = payload.htlc.cltv_expiry;
 
         let short_channel_id = match payload.onion.short_channel_id {
-            Some(scid) => scid,
+            Some(scid) => match ShortChannelId::from_str(&scid) {
+                Ok(scid) => scid_to_u64(scid),
+                Err(_) => {
+                    // Ignore invalid SCID
+                    error!("Received invalid short channel id {:?}", scid);
+                    return serde_json::json!({ "result": "continue" });
+                }
+            },
             None => {
                 // This is a HTLC terminating at the gateway node. DO NOT intercept
                 return serde_json::json!({ "result": "continue" });
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -33,7 +33,7 @@ use tokio::sync::{mpsc, oneshot, Mutex};
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::transport::Server;
 use tonic::Status;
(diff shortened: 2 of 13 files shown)


okjodom (approved):

gateway/ln-gateway/src/gatewayd/actor.rs
```diff
@@ -33,6 +33,12 @@ pub struct GatewayActor {
     task_group: TaskGroup,
 }
 
+#[derive(Debug, Clone)]
+pub enum BuyPreimage {
```
mxz42, line 37: when we use this struct, have we not already bought the preimage? so `BoughtPreimage`?

.tmuxinator.yml
```diff
@@ -28,14 +28,12 @@ windows:
...
-          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN1_DIR --addr=127.0.0.1:9000 --plugin=$FM_BIN_DIR/ln_gateway --fedimint-cfg=$FM_CFG_DIR &
+          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN1_DIR --addr=127.0.0.1:9000 --plugin=$FM_BIN_DIR/gateway-cln-extension &
           - echo $! >> $FM_PID_FILE
           - fg
         - ln2:
           - sleep 5 # wait for bitcoind and federation
-          - export GW_CLN_EXTENSION_LISTEN_ADDRESS=0.0.0.0:54321
-          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN2_DIR --addr=127.0.0.1:9001 --plugin=$FM_BIN_DIR/gateway-cln-extension &
+          - lightningd --dev-fast-gossip --dev-bitcoind-poll=1 --network regtest --bitcoin-rpcuser=bitcoin --bitcoin-rpcpassword=bitcoin --lightning-dir=$FM_LN2_DIR --addr=127.0.0.1:9001
```
mxz42, line 36: do we need the second lightning node?

mxz42:
Also, I'm a bit confused about all these `pay_invoice_.*`  methods :sweat_smile: 


```sh
$ rg 'fn .*?pay_invoice[^ ]+'
gateway/ln-gateway/src/gatewayd/gateway.rs
239:    async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<()> {

gateway/ln-gateway/src/gatewayd/actor.rs
255:    pub async fn pay_invoice(&self, contract_id: ContractId) -> Result<OutPoint> {
264:    pub async fn pay_invoice_buy_preimage(&self, contract_id: ContractId) -> Result<BuyPreimage> {
315:    pub async fn pay_invoice_buy_preimage_finalize(
329:    pub async fn pay_invoice_buy_preimage_finalize_and_claim(

gateway/ln-gateway/src/bin/cln_extension.rs
323:    async fn pay_invoice(


# EDIT: The following are part of the old ln_gateway and will soon be removed, right?
gateway/ln-gateway/src/actor.rs
97:    pub async fn pay_invoice(
110:    pub async fn pay_invoice_buy_preimage(
169:    pub async fn pay_invoice_buy_preimage_finalize(
183:    pub async fn pay_invoice_buy_preimage_finalize_and_claim(

gateway/ln-gateway/src/lib.rs
286:    async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<()> {

gateway/ln-gateway/src/rpc/rpc_server.rs
109:async fn pay_invoice(
```

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -92,7 +92,7 @@ pub struct Htlc {
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Onion {
     #[serde(default)]
-    pub short_channel_id: Option<u64>,
+    pub short_channel_id: Option<String>,
```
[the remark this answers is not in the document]
okjodom: `cln-plugin` returns the scid in human readable format. I initially defined the spec here to expect a u64 which is what is defined in bolt spec

.tmuxinator.yml
okjodom: Yeah, we open channels between the nodes so we can test incoming vs outgoing payment scenarios on the gateway.

okjodom:
> Also, I'm a bit confused about all these `pay_invoice_.*` methods sweat_smile
> 
> ```shell
> $ rg 'fn .*?pay_invoice[^ ]+'
> gateway/ln-gateway/src/gatewayd/gateway.rs
> 239:    async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<()> {
> 
> gateway/ln-gateway/src/gatewayd/actor.rs
> 255:    pub async fn pay_invoice(&self, contract_id: ContractId) -> Result<OutPoint> {
> 264:    pub async fn pay_invoice_buy_preimage(&self, contract_id: ContractId) -> Result<BuyPreimage> {
> 315:    pub async fn pay_invoice_buy_preimage_finalize(
> 329:    pub async fn pay_invoice_buy_preimage_finalize_and_claim(
> 
> gateway/ln-gateway/src/bin/cln_extension.rs
> 323:    async fn pay_invoice(
> 
> 
> # EDIT: The following are part of the old ln_gateway and will soon be removed, right?
> gateway/ln-gateway/src/actor.rs
> 97:    pub async fn pay_invoice(
> 110:    pub async fn pay_invoice_buy_preimage(
> 169:    pub async fn pay_invoice_buy_preimage_finalize(
> 183:    pub async fn pay_invoice_buy_preimage_finalize_and_claim(
> 
> gateway/ln-gateway/src/lib.rs
> 286:    async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<()> {
> 
> gateway/ln-gateway/src/rpc/rpc_server.rs
> 109:async fn pay_invoice(
> ```

The `pay_invoice` methods on `GatewayActor` really fragmented by #1592 in order to allow running integration tests in parallel.  Similar changes in this PR, just ports the same changes from the old `ln_gateway::GatewayActor` to the new `ln_gateway::gatewayd::GatewayActor`. Ideally, the tests would not dictate the api surface of something like the actor.

For next steps, we should do the following:
- Change how we run integration tests on the gatewayd. Instead of testing via direct method calls on a `GatewayActor`, we should make realistic api calls over gatewayd. WIP proposal in #1702

gateway/ln-gateway/src/gatewayd/actor.rs
okjodom: Id want to get rid of the struct all together. `BuyPreimage::Internal` should not be a part of gateways functionality, so we really don't need this enum.

Issue #613 is relevant

gateway/ln-gateway/src/gatewayd/actor.rs
```diff
@@ -147,7 +154,36 @@ impl GatewayActor {
...
+                                // cancel HTCL after expiry period lapses.
+                                // Result can be safely ignored.
+                                // TODO: make sure this succeeded?
+                                let _ = lnrpc_copy
+                                    .complete_htlc(CompleteHtlcsRequest {
+                                        intercepted_htlc_id,
+                                        action: Some(Action::Cancel(Cancel {
+                                            reason: e.to_string(),
+                                        })),
+                                    })
```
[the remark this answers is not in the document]
okjodom: hm. I don't think `buy_preimage_internal` should have anything to do with `complete_htlc!` cc @justinmoon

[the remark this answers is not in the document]
okjodom: fixing

.tmuxinator.yml
justinmoon: Yes. You can't test a lightning payment if you only have 1 node. CLN can't send to itself.

mxz42: cool, thanks :+1:

gateway/ln-gateway/src/gatewayd/actor.rs
mxz42: thanks for the reference

gateway/ln-gateway/src/gatewayd/actor.rs
```diff
@@ -252,57 +296,93 @@ impl GatewayActor {
...
+    #[instrument(skip(self), ret, err)]
     pub async fn buy_preimage_internal(
         &self,
         payment_hash: &sha256::Hash,
         invoice_amount: &Amount,
-    ) -> Result<Preimage> {
+    ) -> Result<(OutPoint, ContractId)> {
+        let mut rng = rand::rngs::OsRng;
+
+        info!("buy_preimage_internal");
```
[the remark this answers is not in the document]
okjodom: removed the info traces, which were meant to be temporary

gateway/ln-gateway/src/gatewayd/actor.rs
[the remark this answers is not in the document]
mxz42: what do you mean by "fixing"?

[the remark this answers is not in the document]
okjodom: > Re: "fixing"

I was concerned by the application of `complete_htlc` rpc call to cancel htlc in case `buy_preimage_internal` failed. My intuition was, _`buy_preimage_internal` should not have anything to do with `complete_htlc` rpc_. So **fixing** meant trying to resolve the failed buy preimage by some other means, other than complete htlc

As a follow up, I figured we actually have to call `complete_htlc`, otherwise we'd have a long pending htlc intercept scenario, before the htlc lapses it's natural expiry and gets cancelled by the LN node.

Further follow up will involve removing internal payments as a feature of gateways. See #613

gateway/ln-gateway/src/bin/cln_extension.rs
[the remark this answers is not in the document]
justinmoon: Using `u64` was just incorrect. Core Lightning has never returned a `u64` for this field. We really need to upstream these structs to avoid errors like this in the future: https://github.com/fedimint/fedimint/issues/1245.

gateway/ln-gateway/src/gatewayd/actor.rs
justinmoon, line 174: In the future we should add test coverage to trigger these failure case and test the it behaves as expected. But I don't think this should block this PR because it does pass our existing test suite which means we aren't testing this sort of incoming HTLC failures.

Also, we shouldn't cheat with `let _ = <result>` in the future.

justinmoon:
@okjodom and I did a review session and I think this looks good-to-go. Since it's my PR I can't accept ...

gateway/ln-gateway/src/gatewayd/actor.rs
okjodom: retry of the rpc call could come in handy

mxz42: > retry of the rpc call could come in handy. We should retry sending the rpc call to cancel htlc

:+1: 

the `utils::retry` function might be useful here, right?

mxz42 (approved):

integrationtests/tests/fixtures/mod.rs
```diff
@@ -195,12 +193,18 @@ pub async fn fixtures(num_peers: u16) -> anyhow::Result<Fixtures> {
...
             let bitcoin = RealBitcoinTest::new(&bitcoin_rpc_url);
+
             let socket_gateway = PathBuf::from(dir.clone()).join("ln1/regtest/lightning-rpc");
             let socket_other = PathBuf::from(dir.clone()).join("ln2/regtest/lightning-rpc");
             let lightning =
                 RealLightningTest::new(socket_gateway.clone(), socket_other.clone()).await;
-            let gateway_lightning_rpc = ClnRpc::new(socket_gateway.clone());
-            let lightning_rpc_adapter = LnRpcAdapter::new(Box::new(gateway_lightning_rpc));
+
+            let lnrpc_addr = env::var("FM_GATEWAY_LIGHTNING_ADDR")
```
jkitman, line 202: Is this only used for testing?  If not, perhaps it should be defined elsewhere?

jkitman (approved):
Seems like this PR does a lot more than just run the new gateway in tests?  But if it is passing with the new gateway that is good enough for me.

PR #1721, opened 2023-02-20
Fix connect info serde impl and accepted string in cli commands

### LINKED ISSUE #1716: The connection string that setup UI generates is bunch of hex while `fedimint-cli` still expects ... some json?
I've just set up a regtest federation and the UI shows me QR code an base32 encoded blow, but `fedimint-cli join-federation` does not like it, and from I can tell still wants json.

Is there a way to convert between these?

@jkitman

Maan2003:
Fixes #1716

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -4,6 +4,7 @@ use std::error::Error;
 use std::fmt::Debug;
 use std::path::PathBuf;
 use std::process::exit;
+use std::str::FromStr;
 use std::sync::Arc;
 
 use bitcoin::{secp256k1, Address, Network, Transaction};
@@ -326,12 +327,12 @@ trait ErrorHandler<T, E> {
         F: Fn(T) -> CliOutput;
 }
 
-impl<T, E: Error + 'static> ErrorHandler<T, E> for Result<T, E> {
+impl<T, E: Into<Box<dyn Error>>> ErrorHandler<T, E> for Result<T, E> {
     fn or_terminate(self, err: CliErrorKind, msg: &str) -> T {
         match self {
             Ok(v) => v,
             Err(e) => {
-                let cli_error = CliError::from(err, msg, Some(Box::new(e)));
+                let cli_error = CliError::from(err, msg, Some(e.into()));
                 eprintln!("{cli_error}");
                 exit(1);
             }
@@ -343,7 +344,7 @@ impl<T, E: Error + 'static> ErrorHandler<T, E> for Result<T, E> {
     {
         match self {
             Ok(v) => Ok(success(v)),
-            Err(e) => Err(CliError::from(err, msg, Some(Box::new(e)))),
+            Err(e) => Err(CliError::from(err, msg, Some(e.into()))),
         }
     }
 }
@@ -387,7 +388,8 @@ async fn main() {
 
         let cli = Cli::parse();
         if let Command::JoinFederation { connect } = cli.command {
-            let connect_obj: WsClientConnectInfo = serde_json::from_str(&connect)
+            let connect_obj: WsClientConnectInfo = WsClientConnectInfo::from_str(&connect)
+                .map_err(Box::<dyn Error>::from)
                 .or_terminate(CliErrorKind::InvalidValue, "invalid connect info");
             let api = Arc::new(WsFederationApi::from_urls(&connect_obj))
                 as Arc<dyn IFederationApi + Send + Sync + 'static>;
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{self, Debug, Display, Formatter};
 use std::io::{Cursor, Read};
@@ -590,7 +591,7 @@ struct FederationMember<C> {
 /// Information required for client to construct [`WsFederationApi`] instance
 ///
 /// Can be used to download the configs and bootstrap a client
-#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct WsClientConnectInfo {
     /// Urls that support the federation API (expected to be in PeerId order)
     pub urls: Vec<Url>,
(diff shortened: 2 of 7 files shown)


client/cli/src/main.rs
```diff
@@ -326,12 +327,12 @@ trait ErrorHandler<T, E> {
         F: Fn(T) -> CliOutput;
 }
 
-impl<T, E: Error + 'static> ErrorHandler<T, E> for Result<T, E> {
+impl<T, E: Into<Box<dyn Error>>> ErrorHandler<T, E> for Result<T, E> {
```
Maan2003, line 330: because anyhow::Error does not impl Error

fedimint-core/Cargo.toml
```diff
@@ -41,6 +41,7 @@ hbbft = { git = "https://github.com/jkitman/hbbft", branch = "upgrade-threshold-
 rand = "0.8.5"
 miniscript = { version = "7.0.0", git = "https://github.com/rust-bitcoin/rust-miniscript/", rev = "2f1535e470c75fad85dbad8633986aae36a89a92", features = [ "compiler", "serde" ] }
 secp256k1-zkp = { version = "0.7.0", features = [ "use-serde", "bitcoin_hashes", "global-context" ] }
+serde_with = "2.2.0"
```
mxz42, line 44: that would be the introduction of a new dependency; though, I guess that'd be fine?

fedimint-server/src/config/io.rs
```diff
@@ -154,6 +156,13 @@ fn plaintext_json_write<T: Serialize + DeserializeOwned>(
     Ok(())
 }
 
+fn plaintext_display_write<T: Display>(obj: &T, path: PathBuf) -> anyhow::Result<()> {
+    let mut file = fs::File::create(path.clone())
```
mxz42, line 160: nit: we can take reference here `&path`

scripts/cli-test.sh
```diff
@@ -25,7 +25,7 @@ start_gateway
 #### BEGIN TESTS ####
 
 # test the fetching of client configs
-CONNECT_STRING=$(cat $FM_CFG_DIR/client-connect.json)
+CONNECT_STRING=$(cat $FM_CFG_DIR/client-connect)
```
mxz42, line 28: new ending to indicate filetype?  `.bech32` or similar?

mxz42 (commented):
this change would also write the `client-connect.json` as bech32. 
previously you could open it and see something like: 

```json
{
  "urls": [
    "ws://127.0.0.1:18174/",
    "ws://127.0.0.1:18184/"
  ],
  "id": "82fa9f96e15668ef7a47034067632da29de736dc9388bb2aa21f562f76826fa7ee929af0ee28bd3324f01545cca0203a"
}
```

now the content would be

```
fed11kuncff50fcp3msef3aypv39jn78n405mw7gmal6uvpft0rw90dq95h8dgfhdzy2wuyx0gagqa5xpx9gqwaen5te0xyerwt3s9cczuvf6xyurzde59u2sqamn8ghj7vfjxuhrqt3s9ccn5vfcxyurgtcuz8av9
```

we're ok with that change?

fedimint-core/Cargo.toml
Maan2003: I can implement it manually, if we don't want to be have this dependency.

scripts/cli-test.sh
Maan2003: IMO extensionless is fine. json is a well know format support by many editors, bech32 is not.

fedimint-core/src/api.rs
```diff
@@ -1099,7 +1100,11 @@ mod tests {
         };
 
         let bech32 = connect.to_string();
+        let json = serde_json::to_string(&connect).unwrap();
+        assert_eq!(json, format!("\"{bech32}\""));
```
dpc, line 1104: json and bech32 being interwined just make reading harder.

I'm fine with one test, but first render and compare bech32, then render and compare json (or the other way around). Just not two at the same time.

justinmoon:
> we're ok with that change?

Yes. This is our new connection string format. Is there an easy way to convert this bech32 back into JSON in CLI for debugging?

Tried `hal` but I just get hex:

```
% hal bech32 decode fed11kuncff50fcp3msef3aypv39jn78n405mw7gmal6uvpft0rw90dq95h8dgfhdzy2wuyx0gagqa5xpx9gqwaen5te0xyerwt3s9cczuvf6xyurzde59u2sqamn8ghj7vfjxuhrqt3s9ccn5vfcxyurgtcuz8av9
{
  "bech32": "fed11kuncff50fcp3msef3aypv39jn78n405mw7gmal6uvpft0rw90dq95h8dgfhdzy2wuyx0gagqa5xpx9gqwaen5te0xyerwt3s9cczuvf6xyurzde59u2sqamn8ghj7vfjxuhrqt3s9ccn5vfcxyurgtcuz8av9",
  "hrp": "fed1",
  "payload": "161c13180909140f091801111b101909111d04010c110512131e0713150f141b0e1e081b1d1f1a1c0c01090b0f030e050f0d00051417070d0809170d02040a0e1c04060f081d08001d140601060508000e1d1913140b190f060419030e0b1110051818021c0c091a06041c03020d1914051c0a10001d1b13070817121e0c0912061c1703000b111005181813140c091806041c03080b18"
}
```

fedimint-core/src/api.rs
dpc, line 1104: Wait. I don't get it. What is going on here? json formatting is `\"{bech32}\"`? How does this ... what?

Cargo.lock
```diff
@@ -733,6 +765,85 @@ dependencies = [
...
+name = "cxx"
+version = "1.0.91"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "86d3488e7665a7a483b57e25bdd90d0aeb2bc7608c8d0346acf2ad3f1caf1d62"
+dependencies = [
+ "cc",
+ "cxxbridge-flags",
+ "cxxbridge-macro",
+ "link-cplusplus",
+]
```
justinmoon, line 778: I suspect this was from another change. Could you update `Cargo.lock` to only include changes with `serde_with`?

dpc:
> @dpc would this require updating the API versions?

I don't think so. The original change was, but it was landed around the same time we landed versioning.

fedimint-core/src/api.rs
Maan2003: \ is just escape here. so format is `"{bech32}"`

Cargo.lock
Maan2003: I am not sure how this is coming in. But this change still stays when regenerating cargo.lock

Maan2003: I think this is due to optional deps of serde_with, just writing manually impl.

fedimint-core/src/api.rs
dpc: Thank you. I need more coffee.

dpc (approved):

scripts/cli-test.sh
mxz42: yah, I agree. 
My thinking was more in regards to indicate what the numbers and letters are in this file. Putting `.bech32` at the end means someone does not have to look at (perhaps lacking/outdated) documentation or figure out what it is

I'm fine with either way, though

dpc (approved):

elsirion (approved):

PR #1735, opened 2023-02-22
feat: fedimintd as a library

dpc:
Refactor `fedimint` crate to make `fedimintd` and `distributedgen` small
builder-based libraries.

This allows people who want to run fedimint with their own modules
to create a small binary project, attach standard and 3rd party modules,
and link them together into their own `fedimintd` "fork".

Thanks to this running 3rd party module will not require forking the
whole fedimintd.

The whole `fedimintd`-fork will look like this:

```rust
use fedimint_ln::LightningGen;
use fedimint_mint::MintGen;
use fedimint_wallet::WalletGen;
use fedimintd::fedimintd::Fedimintd;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    Fedimintd::new()?
        .attach(WalletGen)
        .attach(MintGen)
        .attach(LightningGen)
        .run()
        .await
}
```

along with `Cargo.toml` to import `fedimintd` library and any sub-modules.

The builder might be a bit simplistic as it is RN, but it should be a good start.


fedimintd/src/lib.rs
```diff
@@ -1,12 +1,15 @@
...
-pub mod ui;
+mod ui;
+
+/// Module for creating `distributetgen` binary with custom modules
+pub mod distributed_gen;
+/// Module for creating `fedimintd` binary with custom modules
+pub mod fedimintd;
 
 /// Generates the configuration for the modules configured in the server binary
 pub fn configure_modules(
```
jkitman, line 15: Should this module-specific code also be in the `bin`?

jkitman (approved):
I like the direction we are going in, slowly moving the module-specific code to the smaller outer dependencies.

fedimintd/src/lib.rs
dpc: The "fork"-binary will only link module kinds with the rest of `fedimintd` code. Everything else should stay in the general code.

These should come from the config/UI eventually.

elsirion:
Why wouldn't the `fedimintd` library part live in `fedimint-server`?

elsirion (approved):
Love it!

PR #1737, opened 2023-02-22
Fixes peg-out being under dust limit

### LINKED ISSUE #1693: Check peg-out tx validity
When generating a transaction for a peg-out we should check if bitcoind accepts it. Currently we dont prevent user generated dust outputs for example.

jkitman:
Simple fix for the dust limit error.

Fixes #1693 at least for this case, not sure whether we can generalize, especially given different bitcoin rpc backends.

### DIFF
--- a/modules/fedimint-wallet/src/lib.rs
+++ b/modules/fedimint-wallet/src/lib.rs
@@ -168,7 +168,7 @@ impl Serialize for PendingTransaction {
 
 /// A PSBT that is awaiting enough signatures from the federation to becoming a
 /// `PendingTransaction`
-#[derive(Clone, Debug, Encodable, Decodable)]
+#[derive(Clone, Debug, Eq, PartialEq, Encodable, Decodable)]
 pub struct UnsignedTransaction {
     pub psbt: PartiallySignedTransaction,
     pub signatures: Vec<(PeerId, PegOutSignatureItem)>,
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet/src/lib.rs
```diff
@@ -193,9 +193,9 @@ impl Serialize for UnsignedTransaction {
 }
 
 struct StatelessWallet<'a> {
-    descriptor: &'a Descriptor<CompressedPublicKey>,
-    secret_key: &'a secp256k1::SecretKey,
-    secp: &'a secp256k1::Secp256k1<secp256k1::All>,
+    pub descriptor: &'a Descriptor<CompressedPublicKey>,
+    pub secret_key: &'a secp256k1::SecretKey,
+    pub secp: &'a secp256k1::Secp256k1<secp256k1::All>,
```
elsirion, line 198: Why do these need to be pub? (not that we'd have great pub hygiene elsewhere, but would be nice to become more restrictive over time to reduce public APIs we are committed to)

elsirion (approved):

modules/fedimint-wallet/src/lib.rs
jkitman: To be constructed in the `mod tests`.  Is there a better way around this?

elsirion: Can't submodules always access private fields of structs in super? Maybe I'm completely off here.

dpc: They should, AFAIK.

There's also `pub(crate)` and for integration tests we can always make some "hidden constructor" `#[doc(hidden)]` to cheat, yet not expose publicly stuff we don't want as a stable API.

dpc (approved):

modules/fedimint-wallet/src/lib.rs
jkitman: Ah you are right, not sure why I got a compile error earlier.

elsirion (approved):

PR #1745, opened 2023-02-23
chore: re-export `fedimint_core` from `fedimint_server`

dpc:
This is to save module developers having to import fedimint deps multiple times.


dpc:
This works but `derive(Encodable, Decodable)` want `fedimint_core` anyway, so :shrug:

PR #1747, opened 2023-02-24
refactor: split `impl_db_record` into two macros

dpc:
Why? Because the types these macro `impl`s traits for can come from different crates making it impossible to use it.

Currently this does not happen, but I'm experimenting with 3rd party module building and spliting module into server/client/common sub-parts and this becomes a blocker.

Also, it seems it's easier to use and understand when each macro is smaller (functionaly).

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -860,7 +860,7 @@ where
 ///     MyKey = 0x50,
 /// }
 ///
-/// impl_db_prefix_const!(key = MyKey, value = MyValue, db_prefix = DbKeyPrefix::MyKey);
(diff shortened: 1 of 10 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -884,22 +884,24 @@ where
...
             const DB_PREFIX: u8 = $db_prefix as u8;
             type Key = Self;
             type Value = $val;
         }
+    };
+}
 
+#[macro_export]
+macro_rules! impl_db_lookup{
+    (key = $key:ty $(, query_prefix = $query_prefix:ty)* $(,)?) => {
```
elsirion, line 904: Needs docs

elsirion (approved):
I don't see anything wrong with it, although I haven't come up with a case yet where it would be needed

m1sterc001guy (approved):

PR #1748, opened 2023-02-24
fix: remove unsued macro argument

dpc:


elsirion (approved):
Shouldn't there be a lint against this?

justinmoon (approved):

PR #1749, opened 2023-02-24
refactor: ServerModule::distributed_gen interface

dpc:
We're passing a lot of arguments there, most of which should not be accessible to the module itself. The module should have no access to its instance id, or raw dkg connection.

I think we'll need to define more operations, and possibly refactor it all even more, but this should be a good starting step in the right direction.

### DIFF
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -304,10 +304,7 @@ pub trait IModuleGen: Debug {
 
     async fn distributed_gen(
         &self,
-        connections: &MuxPeerConnections<ModuleInstanceId, DkgPeerMsg>,
-        our_id: &PeerId,
-        module_id: ModuleInstanceId,
-        peers: &[PeerId],
+        peers: &PeerHandle,
         params: &ConfigGenParams,
     ) -> DkgResult<ServerModuleConfig>;
 
@@ -458,10 +455,7 @@ pub trait ModuleGen: Debug + Sized {
 
     async fn distributed_gen(
         &self,
-        connections: &MuxPeerConnections<ModuleInstanceId, DkgPeerMsg>,
-        our_id: &PeerId,
-        module_id: ModuleInstanceId,
-        peers: &[PeerId],
+        peer: &PeerHandle,
         params: &ConfigGenParams,
     ) -> DkgResult<ServerModuleConfig>;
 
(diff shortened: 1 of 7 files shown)


fedimint-core/src/module/mod.rs
```diff
@@ -809,3 +799,44 @@ impl<T: Encodable + Decodable> SerdeModuleEncoding<T> {
...
+
+impl<'a> DkgHandle<'a> {
+    pub fn peers(&self) -> &[PeerId] {
+        self.peers.as_slice()
+    }
+}
+
+impl<'a> DkgHandle<'a> {
+    pub fn new(
+        connections: &'a MuxPeerConnections<ModuleInstanceId, DkgPeerMsg>,
```
elsirion, line 830: Why separate?

fedimint-server/src/config/distributedgen.rs
```diff
@@ -473,3 +476,74 @@ mod tests {
         keys
     }
 }
+
+// TODO: this trait is only needed to break the `DkgHandle` impl
+// from it's defintion that is still in `fedimint-core`
+#[async_trait]
+pub trait DkgHandleOps {
+    async fn run_g1<T>(&self, v: T) -> DkgResult<HashMap<T, DkgKeys<G1Projective>>>
```
elsirion, line 484: We might want to keep this trait for mocking in tests, making it a `dyn` interface.

elsirion (dismissed):

elsirion:
> […] although that means modules will not be able to implement their own versions of DKG.

Modules could have their own `DkgHandleOps` extension trait they impl for `DkgHandle` that supports their custom DKG. We might actually want to move the BLS DKG into its own crate eventually on which both `fedimint-server` (HBBFT), `fedimint-mint` (e-cash, backups) and `fedimint-ln` (threshold encrypted preimages) depend, but e.g. not `fedimint-wallet`.

dpc:
> I guess if we go this route we can abstract DKG from the modules, although that means modules will not be able to implement their own versions of DKG.

I'm fine with giving modules even methods for sending arbitrary messages and so on. I think my most important goal is not passing them `module_instance_id` (so if they send are sending messages `module_instance_id` is injected for them). Cutting on the arguments is another perk.

Maybe we should rename the thing to not be `Dkg` specific. `PeerCommHandle` or something.

fedimint-server/src/config/distributedgen.rs
```diff
@@ -473,3 +476,74 @@ mod tests {
...
+
+    async fn exchange_pubkeys(
+        &self,
+        key: secp256k1::PublicKey,
+    ) -> DkgResult<BTreeMap<PeerId, secp256k1::PublicKey>>;
+}
+
+#[async_trait]
+impl<'a> DkgHandleOps for DkgHandle<'a> {
+    async fn run_g1<T>(&self, v: T) -> DkgResult<HashMap<T, DkgKeys<G1Projective>>>
```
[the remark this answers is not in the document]
dpc: In a different PR. :)

dpc:
Renamed this handle to be `peers: &PeerHandle` everywhere, which should make more sense.

elsirion (approved):

jkitman (approved):

PR #1750, opened 2023-02-24
feat: `just final-check`

dpc:


elsirion (approved):

justinmoon:
How does this compare with https://github.com/fedimint/fedimint/blob/master/scripts/final-checks.sh

IMO we should only have one "final check". Justfile is the right place to define it because it's more discoverable there.

dpc:
> IMO we should only have one "final check". Justfile is the right place to define it because it's more discoverable there.

Yeah, but they are doing slightly different things, and I'm not sure if they are reasons for it. I think we can kill `./script/final-checks.sh` after we have some time getting used to `just final-check` and making sure it works OK.

justinmoon:
Works for me. I think @jkitman is the main user of `final-checks.sh` script ...

justinmoon (approved):

jkitman:
So this `final-check` builds wasm and runs tests?  Perhaps a better name is in order?

I'm okay if you want to remove the shell script, I can just script it locally.

justinmoon:
> I'm okay if you want to remove the shell script, I can just script it locally.

If you find it useful, I think we should have a just command for it. Probably best to have just commands for all developer-facing scripts so that they're more discoverable.

Issue #1730, opened 2023-02-21
Add connect string encoder/decoder to `fedimint-cli`


elsirion:
Since we now use bech32 encoding for federation connect strings we should have a way to inspect and manipulate them.

justinmoon:
What would an "encoder" take as input? JSON?

elsirion:
Probably best to do it that way, just the JSON representation of `WsClientConnectInfo`. You are right that this probably doesn't generalize well.

justinmoon:
@jp1ac4 just start with the decoding side of this. Let us know if you have any questions.

jp1ac4:
Is the encoding part also required? Perhaps it could take `federation_id` and a list of `url`s as arguments?

justinmoon:
That sounds useful @jp1ac4. I'd say give it a try.

justinmoon:
@jp1ac4 Closing this because it looks done. Let me know if I'm wrong!

jp1ac4:
Hi @justinmoon, I think the only thing that might remain is to optionally pass JSON to the encoder instead of the `--id` and `--urls` arguments, but not sure if this is required.

PR #1754, opened 2023-02-24
feat: add connect string decoder to `fedimint-cli`

jp1ac4:
I've made an initial attempt at the decoder part of #1730.

It would be great if you could check I'm on the right track. Thanks.

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -12,7 +12,7 @@ use clap::{Parser, Subcommand};
 use fedimint_core::api::{
     FederationApiExt, GlobalFederationApi, IFederationApi, WsClientConnectInfo, WsFederationApi,
 };
-use fedimint_core::config::{load_from_file, ClientConfig, ModuleGenRegistry};
+use fedimint_core::config::{load_from_file, ClientConfig, FederationId, ModuleGenRegistry};
 use fedimint_core::db::Database;
 use fedimint_core::module::DynModuleGen;
 use fedimint_core::query::EventuallyConsistent;
@@ -32,6 +32,7 @@ use mint_client::utils::{
 use mint_client::{module_decode_stubs, Client, UserClientConfig};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
+use url::Url;
 
 #[derive(Serialize)]
 #[serde(rename_all(serialize = "snake_case"))]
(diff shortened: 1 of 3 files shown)


client/cli/src/main.rs
```diff
@@ -103,6 +104,11 @@ enum CliOutput {
         connect_info: WsClientConnectInfo,
     },
 
+    DecodeConnectInfo {
+        urls: Vec<Url>,
+        id: FederationId,
```
elsirion, line 109: Maybe this can just wrap the `WsClientConnectInfo` transparently? Not sure if that would need [additional annotation](https://serde.rs/attr-flatten.html) to work or if `DecodeConnectInfo(WsClientConnectInfo)` would do the trick.

elsirion (approved):
Looks good! Maybe add a test in the shell script integration tests that checks that it actually works.

client/cli/src/main.rs
jp1ac4: My attempts along these lines don't seem to work as they always output the bech32 string. Maybe I need to add something to `WsClientConnectInfo` so it can output its JSON representation directly?

jp1ac4:
Thanks for your feedback, @elsirion. I'll see if I can add a test to the integration tests.

jp1ac4:
I've added some checks in `cli-test.sh`.

In order for the script to reach my checks, I had to comment out a couple of lines above them that were causing an error:
```
start_gatewayd
./scripts/pegin.sh $PEG_IN_AMOUNT 1 # peg in gateway
```
But I've left those lines unchanged in my commit.

justinmoon:
You could try this to simplify it a bit more https://github.com/fedimint/fedimint/commit/dffacc4718143bd1a59df599faf5093074bad2e6

jp1ac4:
Thanks for your suggestion, @justinmoon. I've now made those changes.

dpc (approved):

dpc:
Needs a rebase.

client/cli/src/main.rs
elsirion: Ah, we probably defined a custom serde serialization, then keep it as is :)

elsirion (approved):
:tada:

elsirion:
I'd say this is merge-ready. Is there anything you still want to do before merging @jp1ac4? Otherwise we can mark it as ready for review :)

justinmoon (approved):

justinmoon:
Thanks for the PR @jp1ac4

PR #1718, opened 2023-02-19
Remove `ln_gateway`

justinmoon:
Removes any types and implementations in support of `ln_gateway`
After PR #1717 , we can fully delete the old gateway, as proposed with this change

PR #1760, opened 2023-02-24
feat: impl Display for FederationId

justinmoon:
Inspired by https://github.com/fedimint/fedimint/pull/1718#pullrequestreview-1314059849

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -1,5 +1,5 @@
 use std::collections::BTreeMap;
-use std::fmt::Debug;
+use std::fmt::{Debug, Display};
 use std::hash::Hash;
 use std::ops::Mul;
 use std::path::Path;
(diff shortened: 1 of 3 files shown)


dpc (dismissed):

fedimint-core/src/config.rs
```diff
@@ -138,6 +138,13 @@ pub struct ConfigResponse {
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, Hash, PartialEq, Encodable)]
 pub struct FederationId(pub threshold_crypto::PublicKey);
 
+impl Display for FederationId {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let hex = self.0.to_bytes().to_hex();
+        write!(f, "{hex}")
+    }
```
mxz42, line 145: we should be able to use  [`format_hex`](https://docs.rs/bitcoin_hashes/latest/bitcoin_hashes/hex/fn.format_hex.html) here.
it's more efficient than going through `ToHex`.  [we also do this here](https://github.com/fedimint/fedimint/blob/master/fedimint-core/src/fmt_utils.rs#L22-L25)

justinmoon: TIL about `format_hex`. Updated to use it.

mxz42 (approved):
nice :+1:

elsirion (approved):

PR #1762, opened 2023-02-24
chore: improve `just final-check`

dpc:


flake.nix
```diff
@@ -777,6 +783,13 @@
                 + wasm32CrossEnvVars;
             });
 
+            # Like `cross` but only with wasm
```
justinmoon, line 786: `link`?

justinmoon (approved):

flake.nix
dpc: The shell right above this one is called `cross`.

justfile
```diff
@@ -30,7 +30,8 @@ lint:
   cargo doc --profile dev --no-deps --document-private-items
 
 final-check: lint
-  nix build -L .#cross.wasm32-unknown-unknown.client-pkgs
+  cargo test --doc
+  nix develop .#crossWasm -c cargo check --target wasm32-unknown-unknown --package mint-client
```
mxz42, line 34: do you want to add a comment above `final-check` when this is supposed to be used?

elsirion (approved):

justfile
dpc: #1767

PR #1765, opened 2023-02-24
fix: shutdown gateway-cln-extension on lightningd exit

okjodom:
-Works when lightningd requests plugin stop, and when the task group is generally shutting down
 
 > Re: _cdecker, "the shutdown signal is not reliable (if the main daemon crashes or aborts the plugin's don't get the signal), as such we must be robust against sudden disappearances"_

we should observe the reliability of `cln-plugin` shutdown subscription, so we can be resilient to lightningd crashes.

PR #1770, opened 2023-02-25
gateway-clin-extension shuts down properly (uses fork of cln_plugin)

justinmoon:
Built on https://github.com/fedimint/fedimint/pull/1765 to avoid merge conflicts. Both are useful. https://github.com/fedimint/fedimint/pull/1765 for graceful exits, this for non-graceful exits.

Also, don't call `lightning-cli plugin stop gateway-cln-extension` in `lib.sh`. `cln_plugin` now exits the process whenever stdin disconnects https://github.com/ElementsProject/lightning/pull/6036

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -531,8 +531,7 @@ dependencies = [
 [[package]]
 name = "cln-plugin"
 version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "49be99e6e5ad55d420884b5b2a68aca890bcd1a1540ed6d2892363623a60f538"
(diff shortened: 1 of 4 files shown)


justinmoon:
@okjodom @mxz42 could you guys check that this fixes the problem of `cln-gateway-extension` continuing to run after `lightningd` exits, e.g. after `scripts/cli-test.sh` finishes?

justinmoon:
Left the PR as a "draft" because it changes CLN to a git dependency. Will move the CLN fork to fedimint org if we proceed, and pin to an exact commit.

okjodom:
> @okjodom @mxz42 could you guys check that this fixes the problem of `cln-gateway-extension` continuing to run after `lightningd` exits, e.g. after `scripts/cli-test.sh` finishes?

confirmed this fixes the problem

PR #1772, opened 2023-02-25
gateway-clin-extension shuts down properly (doesn't need fork of cln_plugin)

justinmoon:
Alternative to #1770 that doesn't require change to `cln_plugin`

Calls `plugin.join()` to wait for the plugin to say it encountered an error and is shutting down. Once this happens, it tell the task group to shutdown, then returns from `Server::serve_with_shutdown()` which will stop the gRPC server and exit from the `main` function of the plugin.


justinmoon:
but it doesn't work quite yet ...

PR #1773, opened 2023-02-25
 feat: clients with a subset of federation module support 

dpc:
See commit message.


elsirion (approved):
Looks good, it's kind of a merkelized config hash. Eventually we might want to merkelize single values instead of module configs to support backwards compatible config extensions (adding a field old clients don't care about).

m1sterc001guy (approved):

jkitman (approved):

Issue #1766, opened 2023-02-24
Our test scripts shouldn't stop the cln-gateway-extension plugin directly


justinmoon:
Instead, we should let `lightningd` stop it. But currently we're stopping it ourselves because lightnignd is failing to stop it when we kill `lightningd`.

https://github.com/fedimint/fedimint/pull/1765#discussion_r1117672753

PR #1775, opened 2023-02-25
Shutdown gateway-cln-plugin when lightningd stops running

justinmoon:
Replaces https://github.com/fedimint/fedimint/pull/1772 and https://github.com/fedimint/fedimint/pull/1770.

Closes https://github.com/fedimint/fedimint/issues/1766.

Switched to our [own fork of core-lightning](https://github.com/fedimint/lightning/tree/fedimint) to get [this change](https://github.com/ElementsProject/lightning/pull/6041).

### DIFF
--- a/gateway/ln-gateway/Cargo.toml
+++ b/gateway/ln-gateway/Cargo.toml
@@ -29,7 +29,7 @@ bitcoin_hashes = "0.11.0"
 bitcoin = { version = "0.29.2", features = ["serde"] }
 clap = { version = "4.1.6", features = ["derive", "std", "help", "usage", "error-context", "suggestions", "env"], default-features = false }
 cln-rpc = "0.1.1"
-cln-plugin = "0.1.1"
+cln-plugin = { git = "https://github.com/fedimint/lightning", rev = "2db131d5" }
 futures = "0.3.24"
 lightning-invoice = "0.21.0"
 fedimint-core ={ path = "../../fedimint-core" }
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -49,7 +49,7 @@ pub struct ClnExtensionOpts {
 async fn main() -> Result<(), anyhow::Error> {
     let tg = TaskGroup::new();
 
-    let (service, listen) = ClnRpcService::new(tg.clone())
+    let (service, listen, plugin) = ClnRpcService::new()
         .await
         .expect("Failed to create cln rpc service");
 
@@ -61,10 +61,10 @@ async fn main() -> Result<(), anyhow::Error> {
     Server::builder()
         .add_service(GatewayLightningServer::new(service))
         .serve_with_shutdown(listen, async {
-            // Wait for a shutdown signal received from the controlling task group
-            // The plugin will be shutdown when this callback returns
-            let shutdown_rx = tg.make_handle().make_shutdown_rx().await;
-            let _ = shutdown_rx.await;
+            // Wait for plugin to signal it's shutting down
+            // Shut down everything else via TaskGroup regardless of error
+            let _ = plugin.join().await;
+            tg.shutdown().await;
         })
         .await
         .map_err(|e| ClnExtensionError::Error(anyhow!("Failed to start server, {:?}", e)))?;
@@ -116,12 +116,12 @@ pub struct ClnRpcClient {}
 pub struct ClnRpcService {
     socket: PathBuf,
     interceptor: Arc<ClnHtlcInterceptor>,
-    task_group: TaskGroup,
 }
 
 impl ClnRpcService {
-    pub async fn new(task_group: TaskGroup) -> Result<(Self, SocketAddr), ClnExtensionError> {
-        let interceptor = Arc::new(ClnHtlcInterceptor::new(task_group.clone()));
+    pub async fn new(
+    ) -> Result<(Self, SocketAddr, Plugin<Arc<ClnHtlcInterceptor>>), ClnExtensionError> {
+        let interceptor = Arc::new(ClnHtlcInterceptor::new());
 
         if let Some(plugin) = Builder::new(stdin(), stdout())
             .option(options::ConfigOption::new(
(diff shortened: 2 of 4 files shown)


justinmoon:
Actually if we don't fork `cln_plugin` and don't handle the `shutdown` notification, the plugin still gets shutdown. The only problem is that sometimes it panics due to https://github.com/ElementsProject/lightning/issues/6039:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: SendError(LogEntry { level: Debug, message: "-- runtime.resource;" })', /Users/justin/.cargo/registry/src/github.com-1ecc6299db9ec823/cln-plugin-0.1.2/src/logging.rs:91:18
```

I don't think this is a problem. Plugin stops and that's the goal. But from node operator POV it could be confusing to see a panic in the logs.

okjodom (approved):

justinmoon:
> hm, I'm not receiving the shutdown hook when running `cli-test.sh`. I should right?

No, I stopped sending it https://github.com/fedimint/fedimint/pull/1775/files#diff-5838719667e50a69f27dfddac131cad8ed4e11bee597a64a1b48454efdfad40bL89. Think it's a little better to test that plugin gets shut down when lightningd is killed. Ideally we'd also test the `shutdown` event but I'm now sure where we'd do that right now ...

mxz42:
> > hm, I'm not receiving the shutdown hook when running `cli-test.sh`. I should right?
> 
> No, I stopped sending it https://github.com/fedimint/fedimint/pull/1775/files#diff-5838719667e50a69f27dfddac131cad8ed4e11bee597a64a1b48454efdfad40bL89. Think it's a little better to test that plugin gets shut down when lightningd is killed. Ideally we'd also test the `shutdown` event but I'm now sure where we'd do that right now ...

hm, ok, but then we [don't need to subscribe to the shutdown event](https://github.com/fedimint/fedimint/pull/1775/files#diff-11314b558aa2437b3f9376371d29c660caf580b98f1e52fe9f7f548d7cc851d0R153-R157) at all if AFAICU?

I'm not sure if that's what we want. Is [your PR](https://github.com/ElementsProject/lightning/pull/6041) going to be merged? Did you speak to cdecker? 

If I understand this whole thing correctly: 
- there is the shutdown hook. we as a plugin should subscribe to that shutdown hook so that `lightningd` can notify us 
   - if `lightningd` is shutting down. e.g. via `lightningd-stop`. 
   - if the `lightning.. plugin stop ` command is issued to our plugin. 

That's the proper way to shutdown. In both cases we also can only do clean up on our part because `lightningd` won't accept any RPC calls (only "id-less" logging and notifications; whatever that exactly means). quote: [shutdown notication documentation](https://lightning.readthedocs.io/PLUGINS.html?highlight=shutdown#shutdown)

> In the shutdown case, plugins should not interact with lightnind except via (id-less) logging or notifications. New rpc calls will fail with error code -5 and (plugin’s) responses will be ignored. Because lightningd can crash or be killed, a plugin cannot rely on the shutdown notification always been send.


There is also the possibility that `lightningd` crashes, or is killed by the OOM killer, or whatever else may happen to the lightningd process so that *lightningd does not have the possibility to notify us (the plugin) to shutdown*. That's the case where we may want to check stdin to notice that `lightningd` is not there anymore. And then we exit. We can only do cleanup on your part as `lightnind` is not there anymore (if we had to clean up anything). 

is that correct so far?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -153,12 +153,8 @@ impl ClnRpcService {
...
                 "shutdown",
                 |plugin: Plugin<Arc<ClnHtlcInterceptor>>, _: serde_json::Value| async move {
-                    info!("Received shutdown event ... gracefully shutting down the plugin");
-                    let handle = tokio::spawn(async move {
-                        plugin.state().shutdown().await;
-                        Ok(())
-                    });
-                    handle.await?
+                    info!("Received \"shutdown\" notification from lightningd ... requesting cln_plugin shutdown");
+                    plugin.shutdown()
```
elsirion, line 157: I wonder if this should be a default subscription we should upstream?

gateway/ln-gateway/Cargo.toml
```diff
@@ -29,7 +29,7 @@ bitcoin_hashes = "0.11.0"
 bitcoin = { version = "0.29.2", features = ["serde"] }
 clap = { version = "4.1.6", features = ["derive", "std", "help", "usage", "error-context", "suggestions", "env"], default-features = false }
 cln-rpc = "0.1.1"
-cln-plugin = "0.1.1"
+cln-plugin = { git = "https://github.com/fedimint/lightning", rev = "2db131d5" }
```
elsirion, line 32: Bing upstreamed in https://github.com/ElementsProject/lightning/pull/6041

elsirion (approved):

justinmoon:
> Is https://github.com/ElementsProject/lightning/pull/6041 going to be merged? Did you speak to cdecker?

Don't know. My last PR took months to merge.

> is that correct so far?

Yes

## more info

Let me describe a little more what this PR does. `cln_plugin`, our dependency for building a core-lightning plugin, has a receiver internally which returns when STDIN goes away which we can await by calling `.join()` ([source](https://github.com/ElementsProject/lightning/blob/master/plugins/src/lib.rs#L690-L701)).

I didn't want to handle shutdown and cleanup in 2 places: `shutdown` notification and awaiting this `plugin.join()`. It just complicates the code and requires passing task groups around in plugin structs.

Instead, I made a [PR](https://github.com/ElementsProject/lightning/pull/6041) to `cln_plugin` which add a simple hook to the plugin to send a value to the aforementioned receiver whenever we receive `shutdown` notification, causing it to return. Now we can just handle shutdown whenever `plugin.join()` returns because `shutdown` notification causes `plugin.join()` to return.

elsirion (approved):

mxz42:
@justinmoon 

thanks for the longer explanation :+1:  sounds good. 

> I have manually verified that this does trigger cleanup via plugin.join().

That's also great.

> One confusing thing is that once we return from the shutdown handler, it seems like lightningd still waits 30 seconds for our plugin to stop.

Did you find out more about this?

justinmoon:
> Did you find out more about this?

I don't remember. I haven't noticed this at all recently. But we don't send `shutdown` anywhere in our tests or anything ...

PR #1776, opened 2023-02-25
fix: shut down CLN plugin when lightningd stops running

justinmoon:
Closes https://github.com/fedimint/fedimint/issues/1766.

Alternative to https://github.com/fedimint/fedimint/pull/1775 which doesn't fork `cln_plugin` dependency. Sometimes it hits logging-related panics on shutdown, but I don't think this is a problem https://github.com/fedimint/fedimint/pull/1775#issuecomment-1445201224.

Whenever `cln_plugin` stops, the `plugin.join()` future will resolve. Do our `TaskGroup`-related cleanup at this point.

### DIFF
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -49,7 +49,7 @@ pub struct ClnExtensionOpts {
 async fn main() -> Result<(), anyhow::Error> {
     let tg = TaskGroup::new();
 
-    let (service, listen) = ClnRpcService::new(tg.clone())
+    let (service, listen, plugin) = ClnRpcService::new()
         .await
         .expect("Failed to create cln rpc service");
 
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -145,22 +145,6 @@ impl ClnRpcService {
                     handle.await?
                 },
             )
-            // Shutdown the plugin when lightningd is shutting down or when the plugin is stopped
-            // via `plugin stop` command. There's a chance that the subscription is never called in
-            // case lightningd crashes or aborts.
-            // For details, see documentation for `shutdown` event notification:
-            // https://lightning.readthedocs.io/PLUGINS.html?highlight=shutdown#shutdown
-            .subscribe(
```
okjodom, line 153: Simple. I like it!

How come we no longer need to subscribe to `shutdown` event?

justinmoon: If we receive a `shutdown` then  lightningd will kill our plugin regardless of whether we handle if AFAIK.

Our problem is when lightningd is killed and doesn't seem to shutdown our plugin at all. But this is solved by awaiting `plugin.join()` and shutting down when that resolves.

justinmoon:
Let's go with #1775 because it's more explicit about handling the `shutdown` notification. I'm more confident it will work consistently because of that.

PR #408, opened 2022-08-18
Rename MINIMINT_TEST_REAL to MINIMINT_TEST_DISABLE_FIXTURES

dpc:
IMO, "real" does not convey the meaning well. "Against real
nodes" would be OK, but seems to long.

PR #1778, opened 2023-02-25
fix: Rename FM_TEST_DISABLE_MOCKS to FM_TEST_WITH_REAL_FIXTURES

justinmoon:
The double-negative in `FM_TEST_DISABLE_MOCKS` variable always confuses me.

Not sure if `FM_TEST_WITH_REAL_FIXTURES` is the best name, but it confuses me less.

What do you think?

### DIFF
--- a/integrationtests/README.md
+++ b/integrationtests/README.md
@@ -31,7 +31,7 @@ Tests run by default with fake Lightning and Bitcoin services for fast concurren
 
 To run the tests in parallel against fake versions of Lightning and Bitcoin:
 ```shell
-export FM_TEST_DISABLE_MOCKS=0
+export FM_TEST_USE_REAL_DAEMONS=0
 cargo test -p fedimint-tests
 ```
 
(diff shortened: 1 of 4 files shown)


justinmoon:
We’ve done this bike shed before so feel free to close! https://github.com/fedimint/fedimint/pull/408#issuecomment-1219835105

mxz42:
I think too that the *double negative* is confusing. 
after going through #408 I also agree with dpc that the default tests (without env var set) should be done on the fixtures ("not the real services"). 

I assumed that "fixtures" would also be something that is mocked but apparently that's wrong. so maybe go without the word fixtures as it seems to cause confusion?

imho, we should always avoid double negatives. it usually makes things weird or you have to think twice about what something means.

@dpc suggested 
- `FM_TEST_USE_REAL_BITCOIND_AND_LIGHTNINGD` which is nice but sort of long

maybe a bit more concise
- `FM_TEST_USE_REAL_SERVICES`
- `FM_TEST_USE_REAL_DAEMONS` (I'd prefer that one as both, `bitcoind` and `lightningd` are daemons, it's not a double negative, and it's shorter than dpc's good suggestion)

dpc (approved):

justinmoon:
rebased and took @mxz42  suggestion to use `FM_TEST_USE_REAL_DAEMONS`. That is very clear IMO.

dpc (approved):

mxz42 (approved):
tACK

PR #1781, opened 2023-02-26
fix: Remove unnecessary sleep and await_fedimint_block_sync calls

justinmoon:
In the process of trying to get `gateway-cli-extension` to reliably stop after running `cli-test.sh`, I noticed that we had an unnecessary sleep in `cli-test.sh`. We can just poll `getreceivedbyaddress` which is the next line in the script.

### DIFF
--- a/scripts/cli-test.sh
+++ b/scripts/cli-test.sh
@@ -47,8 +47,6 @@ $FM_MINT_CLIENT fetch
 # peg out
 PEG_OUT_ADDR="$($FM_BTC_CLIENT getnewaddress)"
 $FM_MINT_CLIENT peg-out $PEG_OUT_ADDR 500
-sleep 5 # FIXME wait for tx to be included
-await_fedimint_block_sync
 until [ "$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR 0)" == "0.00000500" ]; do
   sleep $POLL_INTERVAL
 done


justinmoon:
rebased

scripts/cli-test.sh
```diff
@@ -47,8 +47,6 @@ $FM_MINT_CLIENT fetch
 # peg out
 PEG_OUT_ADDR="$($FM_BTC_CLIENT getnewaddress)"
 $FM_MINT_CLIENT peg-out $PEG_OUT_ADDR 500
-sleep 5 # FIXME wait for tx to be included
-await_fedimint_block_sync
 until [ "$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR 0)" == "0.00000500" ]; do
   sleep $POLL_INTERVAL
```
dpc, line 51: Not something to change in this PR, but IMO we should always have `&>2 echo "Waiting for xyz..." before sleep in these loops to have a visual clue about what is taking how much etc.

scripts/cli-test.sh
```diff
@@ -47,8 +47,6 @@ $FM_MINT_CLIENT fetch
 # peg out
 PEG_OUT_ADDR="$($FM_BTC_CLIENT getnewaddress)"
 $FM_MINT_CLIENT peg-out $PEG_OUT_ADDR 500
-sleep 5 # FIXME wait for tx to be included
-await_fedimint_block_sync
```
elsirion, line 51: Do we know why the `await_fedimint_block_sync` was there and why it's ok to remove? Intuitively this seems ok, I just hope there isn't some arcane dependency on it a fe lines further.

elsirion (approved):

jkitman (approved):

scripts/cli-test.sh
justinmoon: I'm not sure why it's there. We mine more blocks a few lines lower and don't do anything with fedimint in between. So doesn't seem useful at all.

PR #1786, opened 2023-02-27
feat: Combine `fetch` and `wait-invoice` CLI commands

### LINKED ISSUE #1463: Combine `wait-invoice` and `fetch` CLI commands
First time users of our CLI alway get confused by having to receive money over LN, which requires: 

(1) `ln-invoice` to generate invoice
(2) `wait-invoice` to wait for it to be paid and claim the funds in the contract
(3) `fetch` to get your balance to update

Perhaps we should just have `fetch` look for any `ConfirmedInvoies` in the DB and try to complete them. Then the flow would just require steps (1) and (3).

jp1ac4:
I've started work on this PR to resolve #1463 following the suggested approach of checking for unexpired `ConfirmedInvoice`s in the DB.

I haven't yet added any tests or removed the `wait-invoice` command (we may want to keep it still?).

There may be a more efficient way to iterate over the contracts to claim.

The `fetch` command now tries to claim contracts for all unexpired `ConfirmedInvoice`s as a first step, but I haven't added anything to check the output of this step or return it to the user.

I'll be grateful for your feedback. Thanks.

### DIFF
--- a/client/client-lib/src/lib.rs
+++ b/client/client-lib/src/lib.rs
@@ -56,7 +56,7 @@ use lightning::ln::PaymentSecret;
 use lightning::routing::gossip::RoutingFees;
 use lightning::routing::router::{RouteHint, RouteHintHop};
 use lightning_invoice::{CreationError, Invoice, InvoiceBuilder, DEFAULT_EXPIRY_TIME};
-use ln::db::LightningGatewayKey;
+use ln::db::{ConfirmedInvoiceKey, ConfirmedInvoiceKeyPrefix, LightningGatewayKey};
 use ln::PayInvoicePayload;
 use mint::NoteIssuanceRequests;
 use modules::mint::MintOutputOutcome;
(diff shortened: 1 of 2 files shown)


dpc:
> I haven't yet added any tests or removed the `wait-invoice` command (we may want to keep it still?).

For scripting we do want ability to do each phase separately, either by a separate command or combination of flags on other command. (probably most clear if we use separate commands), so yeah ... let's keep it.

jp1ac4:
Hi @justinmoon, do you think the output of the `fetch`command should be modified to incorporate the `wait-invoice` output?

justinmoon:
> Hi @justinmoon, do you think the output of the `fetch`command should be modified to incorporate the `wait-invoice` output?

Yes let's update the output

jp1ac4:
Hi @justinmoon, I've modified the `fetch` output to include information about the claimed contracts.

If there are several contracts, I only return an error if they all fail to be claimed. This is because any contracts that have been processed previously will return an error and then it would not be possible to claim a new contract until the previous ones expired.

Please let me know if this is what you had in mind or if you have any suggestions. Thanks.

client/cli/src/main.rs
```diff
@@ -524,11 +525,21 @@ async fn handle_command(
...
+                Ok(result) => Ok(CliOutput::Fetch {
+                    claimed_incoming_contracts: contracts,
+                    issuance: (result),
+                }),
+                Err(error) => Err(CliError::from(
+                    CliErrorKind::GeneralFederationError,
+                    "failed to fetch notes",
+                    Some(Box::new(error)),
+                )),
+            },
```
elsirion, line 539: Nesting `match`es isn't nice, consider putting the fetch code into a function that resturns a `Result` and then using `map_err` and the `?` operator instead.

client/client-lib/src/lib.rs
```diff
@@ -1102,6 +1102,56 @@ impl Client<UserClientConfig> {
...
+    pub async fn list_unexpired_confirmed_invoices(&self) -> Vec<(ContractId, ConfirmedInvoice)> {
+        self.context
+            .db
+            .begin_transaction()
+            .await
+            .find_by_prefix(&ConfirmedInvoiceKeyPrefix)
+            .await
+            .filter_map(
+                |(ConfirmedInvoiceKey(contract_id), confirmed_invoice)| async move {
+                    if !confirmed_invoice.invoice.is_expired() {
```
elsirion, line 1115: Just because an invoice expired doesn't mean it wasn't paid before that deadline and we checked it since then.

client/cli/src/main.rs
jp1ac4: OK sure, will give that a try thanks.

client/client-lib/src/lib.rs
jp1ac4: Ah yes, I see. I think the idea of checking for expired invoices was as a simpler alternative to saving in the DB whether the invoice was paid, but perhaps we should go with the DB approach instead (see also https://github.com/fedimint/fedimint/issues/1463#issuecomment-1444509118).

client/cli/src/main.rs
jp1ac4: I've removed the `match` expressions now. Happy to refactor further if required.

client/client-lib/src/lib.rs
jp1ac4: No problem at all. I've removed the filtering now.

Note that in case `fetch` is run when there are no new contracts to claim, a "transaction already processed" error will be returned for each previously claimed contract, which could grow large over time if there is no filtering on the contracts we attempt to claim. Perhaps I need to change the error handling, but I guess the command will typically be run when there is a new contract to claim.

jp1ac4:
As per https://github.com/fedimint/fedimint/issues/1463#issuecomment-1519048279, this is no longer required.

PR #1788, opened 2023-02-27
Adds tests for electrum

jkitman:
Runs wallet tests using `electrs` fixing a bug where the fee rate could be below the min relay fee.

Note that we should do the same for `esplora` eventually, although it's not in Nix packages.

Also we should handle errors more intelligently in general I think rather than just retrying all the time.


dpc (approved):

mxz42:
btw, as I just saw the commit: would be nice if that would use the conventional commit style. I guess `test: ` would be a good fit here

elsirion (approved):

Issue #1723, opened 2023-02-20
Remove fedimint_server dependency from gateway crates 


justinmoon:
https://github.com/fedimint/fedimint/issues/1722 helps here

PR #1790, opened 2023-02-27
feat: check (ban) architectually unwanted dependencies

dpc:


misc/git-hooks/pre-commit
```diff
@@ -33,6 +33,13 @@ fi
 export git_ls_files
 git_ls_files="$(git ls-files)"
 
+function check_forbiden_depenencies() {
+  if grep -E "(fedimint-mint|fedimint-wallet|fedimint-ln)" fedimint-server/Cargo.toml >&2 ; then
+    >&2 echo "fedimint-server/Cargo.toml must not depend on modules"
+    return 1
+  fi
+}
+export -f check_forbiden_depenencies
```
elsirion, line 42: Can we do the same for others like new client <-> server?

elsirion (approved):

misc/git-hooks/pre-commit
```diff
@@ -33,6 +33,13 @@ fi
 export git_ls_files
 git_ls_files="$(git ls-files)"
 
+function check_forbiden_depenencies() {
```
okjodom, line 36: nit: *dependencies

misc/git-hooks/pre-commit
dpc: Which crates shouldn't include which exactly?

okjodom: Citing recent example, `gateway/*` should not include `fedimint-server`. #1723

jkitman:
@dpc can we enforce something similar to this? https://github.com/fedimint/fedimint/blob/master/docs/modular-architecture.md

dpc:
> @dpc can we enforce something similar to this? https://github.com/fedimint/fedimint/blob/master/docs/modular-architecture.md

We can ban unwanted deps, but that's not exactly the same as enfocing that chart.

I added all the suggestions. We can always add more as we spot them.

okjodom (approved):
🚀

justinmoon (approved):

PR #1793, opened 2023-02-27
fix: make SystemTime serde repr same on wasm and native

Maan2003:
the serde repr for SystemTime was different on wasm because it wrapped a `Duration`.

### DIFF
--- a/fedimint-core/src/time.rs
+++ b/fedimint-core/src/time.rs
@@ -27,21 +27,22 @@ mod wasm {
         Encodable,
         Decodable,
     )]
-    pub struct SystemTime(Duration);
+    pub struct SystemTime(std::time::SystemTime);
 
     impl SystemTime {
-        pub const UNIX_EPOCH: SystemTime = SystemTime(Duration::from_secs(0));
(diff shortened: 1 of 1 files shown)


justinmoon (approved):

fedimint-core/src/time.rs
```diff
@@ -27,21 +27,22 @@ mod wasm {
         Encodable,
         Decodable,
     )]
-    pub struct SystemTime(Duration);
+    pub struct SystemTime(std::time::SystemTime);
```
elsirion, line 30: Do we actually still need the newtype? Seems like we could just use std `SystemTime` directly? Maybe except for `now` which we could amend using a custom trait.

elsirion (approved):
Generally ok, merge if my comment is invalid.

fedimint-core/src/time.rs
Maan2003: IMO switching from std SystemTime to fedimint SystemTime should not require any other changes. If we add a trait that also needs to be imported :shrug:

dpc (approved):

PR #1794, opened 2023-02-27
fix: log to stderr and other minor logging issues

dpc:


fedimint-logging/src/lib.rs
```diff
@@ -52,7 +53,9 @@ impl TracingSetup {
     pub fn init(&self) -> anyhow::Result<()> {
         let filter_layer =
             EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
-        let fmt_layer = tracing_subscriber::fmt::layer().with_filter(filter_layer);
+        let fmt_layer = tracing_subscriber::fmt::layer()
+            .with_writer(io::stderr)
```
garyray-k, line 57: Is there a reason tracing is written to stderr and not stdout?

dpc: Logging to stdout can interfere with normal output of each program.

Maan2003: :eyes: fmt layer uses stdout by default?

Maan2003 (approved):

elsirion (approved):

fedimint-logging/src/lib.rs
dpc: @Maan2003 It seems like it. I filled https://github.com/tokio-rs/tracing/issues/2492 to complain / ask about it.

PR #1798, opened 2023-02-28
fix: ln client doesn't retry on preimage pending decryption

### LINKED ISSUE #1795: `await_preimage_decryption` doesn't retry on `Pending`
`await_preimage_decryption` calls `await_output_outcome` which will return:

```
    LN(                                                                                                                                                         
        Contract {                                                                                                                                                          id: be3d072c81b3efc01c4ad8fa51ca274c3f61f21db09ce4f1b0945dd9041afb82,                                                                               
            outcome: Incoming(                                                                                                                                                  Pending,                                                                                                                                        
            ),                                                                                                                                                  
        },                                                                                                                                                          ),   
```

which technically is not an output outcome error. After receiving it the `buy_preimage_from_federation_await_decryption` immediately fails and runs `refund_incoming_contract` .

dpc:
This is a source of test flakiness.

Fix #1795

### DIFF
--- a/client/client-lib/src/lib.rs
+++ b/client/client-lib/src/lib.rs
@@ -13,8 +13,9 @@ pub mod modules {
 
 use std::fmt::{Debug, Formatter};
 use std::iter::once;
+use std::ops::Add;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use api::{LnFederationApi, WalletFederationApi};
 use bitcoin::util::key::KeyPair;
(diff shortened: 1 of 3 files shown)


client/client-lib/src/lib.rs
```diff
@@ -1415,16 +1416,39 @@ impl Client<GatewayClientConfig> {
...
-        Ok(self
-            .context
-            .api
-            .await_output_outcome::<OutputOutcome>(
-                outpoint,
-                Duration::from_secs(10),
-                &self.context.decoders,
+        let deadline = Instant::now().add(Duration::from_secs(30));
+
+        let poll = || async {
```
okjodom, line 1421: Poll looks like a variant of retry. Is our existing [retry utility ](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/utils.rs#L16) applicable for this?

dpc: It has to be deadline based, not attempt-s based, so I don't think so.

dpc: I don't know if it's worth adding one for it either. Maybe if we converted more code to use deadlines it would make sense.

justinmoon:
Is there anywhere else in the codebase where we should be polling with a deadline?

justinmoon (approved):

dpc:
> Is there anywhere else in the codebase where we should be polling with a deadline?
> 
> Looked a little and couldn't find anything ...

We should always use deadlines. Deadlines are the best. But not in this PR. :D

okjodom (approved):

PR #1799, opened 2023-02-28
refactor: split CommonModuleGen from ServerModuleGen

dpc:
On top of #1773

I'd like to actually split server & client sides of modules into sub-crates to stop `ModuleGen`'s server-side functionality on the client. To do so, we need to split `ModulGen`.


dpc:
Not sure if this is the way to go yet.

dpc:
Now that it's ready it's probably better to just land #1805 instead.

Issue #1387, opened 2023-01-18
Mint makes a lot of `getblockcount` calls


dpc:
When debugging tests with env RUST_LOG=debug

```
mint 0: 2023-01-18T06:56:39.601556Z DEBUG bitcoincore_rpc: JSON-RPC request: estimatesmartfee [10,"CONSERVATIVE"]                                                                                                                                            
mint 1: 2023-01-18T06:56:39.601998Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
mint 3: 2023-01-18T06:56:39.602096Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
mint 2: 2023-01-18T06:56:39.602057Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
mint 1: 2023-01-18T06:56:39.602671Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []             
mint 3: 2023-01-18T06:56:39.602707Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
mint 0: 2023-01-18T06:56:39.602789Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
mint 3: 2023-01-18T06:56:39.603127Z DEBUG bitcoincore_rpc: JSON-RPC request: estimatesmartfee [10,"CONSERVATIVE"]                                                                                                                                            
mint 1: 2023-01-18T06:56:39.603230Z DEBUG bitcoincore_rpc: JSON-RPC request: estimatesmartfee [10,"CONSERVATIVE"]                                                                                                                                            
mint 0: 2023-01-18T06:56:39.603530Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
mint 0: 2023-01-18T06:56:39.603970Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
mint 0: 2023-01-18T06:56:39.604256Z DEBUG bitcoincore_rpc: JSON-RPC request: estimatesmartfee [10,"CONSERVATIVE"]                                                                                                                                            
mint 3: 2023-01-18T06:56:39.604454Z DEBUG bitcoincore_rpc: JSON-RPC request: getblockcount []                                                                                                                                                                
[...]

PR #1802, opened 2023-03-01
Add debug logging to bitcoincore rpc

garyray-k:
Created in reference to #1387
Simply logs whether an RPC call was successful or not.

### DIFF
--- a/fedimint-bitcoind/src/bitcoincore_rpc.rs
+++ b/fedimint-bitcoind/src/bitcoincore_rpc.rs
@@ -12,7 +12,7 @@ use fedimint_core::bitcoin_rpc::BitcoindRpcBackend;
 use fedimint_core::module::__reexports::serde_json::Value;
 use jsonrpc::error::Error as JsonError;
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -152,34 +154,35 @@ where
 
     async fn get_block_height(&self) -> Result<u64> {
```
Maan2003, line 155: you can use tracing #instrument instead of manual logging
```suggestion
    #[instrument(name = "BitcoindRpc::get_block_height", ret, err)]
    async fn get_block_height(&self) -> Result<u64> {
```

garyray-k: Thanks for the recommendation. I have used the macro instead of manual logging.

Maan2003:
the CI failure is unrelated, will rerun it.

Maan2003 (approved):
LGTM

dpc (approved):

PR #1803, opened 2023-03-01
test: lightning module database migration test

m1sterc001guy:
Right now, we do not currently have a mechanism for catching if someone updated a database key or value, but forgot to update the corresponding database migration. This PR aims to add test coveage to catch this scenario, starting with the lightning module.

There are two components to testing database migrations:

1. Creating a backup. This can be done by setting `export FM_TEST_DB_BACKUP_DIR` to a directory then running `cargo test prepare_migration_snapshots`. This will create database backups in the directory specified by the environment variable. This directory can then be saved/zipped for step 2 that can be run at a later date.

2. Run the database migration tests against a backup. The database migration tests are only run if the FM_TEST_DB_BACKUP_DIR has been set (otherwise they just pass). The database migration tests will then apply the database migrations to every database and then read every key/value from every database. If a database migration has failed, or the key/value type has been changed but no migration has been specified, these tests should fail.

The idea is that steps 1 and 2 will be run at different times. Step 1 should be run when doing releases (or ideally, anytime a database key/value changes). Step 2 should ideally be run as part of CI or a daily cadence so that mistakes are caught early.

Next: Update CI to automatically run step 2 from a pre-saved backup.

### DIFF
--- a/fedimint-testing/src/lib.rs
+++ b/fedimint-testing/src/lib.rs
@@ -1,14 +1,16 @@
 use std::collections::HashSet;
 use std::fmt::Debug;
 use std::future::Future;
+use std::path::Path;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
+use std::{env, fs, io};
 
 use async_trait::async_trait;
 use fedimint_core::config::{ClientModuleConfig, ConfigGenParams, ServerModuleConfig};
 use fedimint_core::core::{ModuleInstanceId, LEGACY_HARDCODED_INSTANCE_ID_WALLET};
 use fedimint_core::db::mem_impl::MemDatabase;
-use fedimint_core::db::{Database, ModuleDatabaseTransaction};
+use fedimint_core::db::{Database, DatabaseTransaction, ModuleDatabaseTransaction};
 use fedimint_core::module::interconnect::ModuleInterconect;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::module::{
@@ -16,6 +18,10 @@ use fedimint_core::module::{
     TransactionItemAmount,
 };
 use fedimint_core::{OutPoint, PeerId, ServerModule};
+use fedimint_rocksdb::RocksDb;
+use futures::future::BoxFuture;
+use rand::rngs::OsRng;
+use rand::RngCore;
 
 pub mod btc;
 pub mod ln;
@@ -338,6 +344,92 @@ where
     }
 
     first
+}
+
+/// Creates the database backup directory by appending the `snapshot_name`
+/// to the `FM_TEST_DB_BACKUP_DIR`. Then this function will execute the provided
+/// `prepare_fn` which is expected to populate the database with the appropriate
+/// data for testing a migration. If `FM_TEST_DB_BACKUP_DIR` is not set, this
+/// function doesn't do anything.
+pub async fn prepare_snapshot<F>(
+    snapshot_name: &str,
+    prepare_fn: F,
+    decoders: ModuleDecoderRegistry,
+) where
+    F: for<'a> Fn(DatabaseTransaction<'a>) -> BoxFuture<'a, ()>,
+{
+    if let Ok(parent_dir) = env::var("FM_TEST_DB_BACKUP_DIR") {
+        let snapshot_dir = Path::new(&parent_dir).join(snapshot_name);
+        let db = Database::new(RocksDb::open(snapshot_dir).unwrap(), decoders);
+        let dbtx = db.begin_transaction().await;
+        prepare_fn(dbtx).await;
+    }
+}
+
+/// Iterates over all of the databases supplied in the database backup
+/// directory, which is specified by `FM_TEST_DB_BACKUP_DIR` environment
+/// variable. First, a temporary database will be created and the contents will
+/// be populated from the database backup directory. Next, this function will
+/// execute the provided `validate` closure. The `validate` closure is expected
+/// to do any validation necessary on the temporary database, such as applying
+/// the appropriate database migrations and then reading all of the data to
+/// verify the migrations were successful.
+pub async fn validate_migrations<F, Fut>(validate: F, decoders: ModuleDecoderRegistry)
+where
+    F: Fn(Database) -> Fut,
+    Fut: futures::Future<Output = ()>,
+{
+    if let Ok(parent_dir) = env::var("FM_TEST_DB_BACKUP_DIR") {
+        let db_dir = Path::new(&parent_dir);
+        let files = fs::read_dir(db_dir).unwrap();
+        for file in files.flatten() {
+            let name = file.file_name().into_string().unwrap();
+            let temp_db = open_temp_db_and_copy(
+                format!("{}-{}", name.as_str(), OsRng.next_u64()),
+                &file.path(),
+                decoders.clone(),
+            );
+            validate(temp_db).await;
+        }
+    }
+}
+
+/// Open a temporary database located at `temp_path` and copy the contents from
+/// the folder `src_dir` to the temporary database's path.
+fn open_temp_db_and_copy(
+    temp_path: String,
+    src_dir: &Path,
+    decoders: ModuleDecoderRegistry,
+) -> Database {
+    // First copy the contents from src_dir to the path where the database will be
+    // opened
+    let path = tempfile::Builder::new()
+        .prefix(temp_path.as_str())
+        .tempdir()
+        .unwrap();
+    copy_directory(src_dir, path.path()).expect("Error copying database to temporary directory");
+
+    Database::new(RocksDb::open(path).unwrap(), decoders)
+}
+
+/// Helper function that recursively copies all of the contents from
+/// `src` to `dst`.
+pub fn copy_directory(src: &Path, dst: &Path) -> io::Result<()> {
+    // Create the destination directory if it doesn't exist
+    fs::create_dir_all(dst)?;
+
+    for entry in fs::read_dir(src)? {
+        let entry = entry?;
+        let path = entry.path();
+        if path.is_dir() {
+            copy_directory(&path, &dst.join(entry.file_name()))?;
+        } else {
+            let dst_path = dst.join(entry.file_name());
+            fs::copy(&path, &dst_path)?;
+        }
+    }
+
+    Ok(())
 }
 
 struct FakeInterconnect(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1402,6 +1402,7 @@ dependencies = [
  "fedimint-bitcoind",
  "fedimint-core",
  "fedimint-ln-client",
+ "fedimint-rocksdb",
  "fedimint-server",
  "fedimint-wallet-client",
  "futures",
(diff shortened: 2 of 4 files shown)


modules/fedimint-ln/src/db.rs
```diff
@@ -118,3 +118,143 @@ impl_db_lookup!(
...
+        LightningGatewayKeyPrefix, ProposeDecryptionShareKeyPrefix,
+    };
+    use crate::LightningGen;
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn verify_lightning_database_migration() {
+        // Only run the database migration test if the database backup directory
+        // environment variable has been defined
+        let backup_dir = env::var("FM_TEST_DB_BACKUP_DIR");
+        if let Ok(backup_dir) = backup_dir {
```
dpc, line 146: The indentation level here is getting out of control while the whole block is very long, so maybe `let Some(backup_dir) = backup_dir else { return };` https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html could help.

modules/fedimint-ln/src/db.rs
```diff
@@ -118,3 +118,143 @@ impl_db_lookup!(
...
+            let mut migrated_values = BTreeMap::new();
+            apply_to_databases(
```
dpc, line 148: `apply_to_databases` doesn't seem informative. I'm confused what's going on, and will have to read impl of `apply_to_databases` first.

modules/fedimint-ln/src/db.rs
```diff
@@ -118,3 +118,143 @@ impl_db_lookup!(
...
+                Path::new(&backup_dir),
+                |db| async move {
+                    // First apply all of the database migrations so that the data can be properly
+                    // read.
+                    let module = DynModuleGen::from(LightningGen);
+                    let isolated_db = db.new_isolated(LEGACY_HARDCODED_INSTANCE_ID_LN);
```
dpc, line 154: Could we always use `0` as the module id here? The module shouldn't care or even know which one we're using.

modules/fedimint-ln/src/db.rs
```diff
@@ -118,3 +118,143 @@ impl_db_lookup!(
...
+                    .await
+                    .expect("Error applying migrations to temp database");
+
+                    // Verify that all of the data from the lightning namespace can be read. If a
+                    // database migration failed or was not properly supplied,
+                    // this will fail.
+                    let mut migrated_pairs: BTreeMap<u8, usize> = BTreeMap::new();
+                    let mut dbtx = isolated_db.begin_transaction().await;
+
+                    for prefix in DbKeyPrefix::iter() {
```
dpc, line 170: Move to another function maybe to make more readable?

modules/fedimint-ln/src/db.rs
```diff
@@ -118,3 +118,143 @@ impl_db_lookup!(
...
+                                    .await;
+                                let num_gateways = gateways.len();
+                                for gateway in gateways {
+                                    gateway.expect("Error deserializing LightningGateway");
+                                }
+                                migrated_pairs
+                                    .insert(DbKeyPrefix::LightningGateway as u8, num_gateways);
+                            }
+                            DbKeyPrefix::Offer => {
+                                // TODO: Offers are temporary database entries
```
dpc, line 224: That's why I think module developer should write manually their own logic to populate the database.

Migrations must work in any state, even if federation got shut down/crashed in them middle of something weird.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -103,6 +104,44 @@ pub struct Fixtures {
...
+///
+/// <backup_parent>
+///     test-X
+///         db-A-B
+///         db-C-D
+/// X, B, and D are random numbers. A and C are the PeerId's of the federation
+/// members.
+fn backup_databases(backup_parent: &Path) -> anyhow::Result<()> {
+    let backup_dir = backup_parent.join(format!("test-{}", rng().next_u64()));
+    let test_dir = env::var("FM_TEST_DIR")?;
```
dpc, line 127: Would be probably more composable if this function took `test_dir` as an argument.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -103,6 +104,44 @@ pub struct Fixtures {
...
+
+    let files = fs::read_dir(db_dir)?;
+    for file in files.flatten() {
+        let name = file.file_name().into_string().unwrap();
+        if name.starts_with("db-") {
+            let src = file.path();
+            let dst = backup_dir.join(name);
+            tracing::info!("Copying from {:?} to {:?}", src, dst);
+            copy_directory(&src, &dst)?;
+            fs::remove_dir_all(&src)?;
```
dpc, line 138: If it's "backup", why does it remove?

integrationtests/tests/fixtures/mod.rs
```diff
@@ -127,6 +166,11 @@ where
         fixtures.lightning,
     )
     .await;
+
+    if let Ok(backup_dir) = env::var("FM_TEST_DB_BACKUP_DIR") {
```
dpc, line 170: I generally don't like the part of taking a snapshot from integration tests etc. I don't think we are going to get good coverage from it.

Developers of every module needs to prepare their own snapshot to test from on their own, seeding with intelligently crafted data, IMO.

dpc (commented):
Some minor suggestions, but main point is that I think trying to couple it with integration tests is not a good idea. Taking snapshots of integration tests are not going to give us good data to work with, and introduces coupling with integration test module set composition.

It should all be done per-module, with module developer seeding the database snapshot with meaningful data manually.

Otherwise - how is 3rd party module developer supposed to test migrations anyway? We can't expect ever 3rd party module to add themselves to our integration tests.

dpc:
As a 3rd party module developer I'd like to (in my module code) write something like:

```
mod tests {
   fn seed_db_with_v0_data(db: &Database) -> Result<()> {
      // use datastructures with v0 vesion and write some example data to `db`
   }

  #[tokio::something::whatver::test]
   fn prepare_migration_snapshots() {
      // this runs only if certain env var is set or something
      fedimint-db-testing::prepare_snapshot(
       "v0",
       seed_db_with_v0_data
      );

      fedimint-db-testing::prepare_snapshot(
       "v8",
       seed_db_with_v8_data
      );
   }


  #[tokio::something::whatver::test]
   fn test_migrations() {
     fedimint-db-testing::test_migrations( /* create module gen or whatever is required to let `test_migrations` do it's job */);
   }
}
```

where

`fedimint-db-testing::prepare_snapshot` would take care of dumping database snapshots somewhere, to be checked-in into git.

and

`fedimint-db-testing::test_migrations` run the migrations against each previous prepared and stored snapshot.

fedimint-testing/src/lib.rs
```diff
@@ -429,6 +445,25 @@ fn open_temp_db_and_copy(temp_path: String, src_dir: &Path) -> Database {
     Database::new(RocksDb::open(path).unwrap(), decoders)
 }
 
+pub fn open_temp_db(path: &Path) -> Database {
+    let decoders = ModuleDecoderRegistry::from_iter([
```
dpc, line 449: Let's at least use a function where we hardcode it (if there isn't one already).

modules/fedimint-ln/src/db.rs
```diff
@@ -121,140 +121,221 @@ impl_db_lookup!(
...
+
+        dbtx.commit_tx()
+            .await
+            .expect("Error committing to database");
+        Ok(())
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn prepare_migration_snapshots() {
+        if let Ok(parent_dir) = env::var("FM_TEST_DB_BACKUP_DIR") {
```
dpc, line 238: We should move this to a function that will generate the destination, prepare an empty db etc. that just takes the name of the snapshot and a function that will populate it.

modules/fedimint-ln/src/db.rs
```diff
@@ -121,140 +121,221 @@ impl_db_lookup!(
...
+    };
     use crate::db::{
         AgreedDecryptionShareKeyPrefix, ContractKeyPrefix, ContractUpdateKeyPrefix, DbKeyPrefix,
-        LightningGatewayKeyPrefix, ProposeDecryptionShareKeyPrefix,
+        LightningGatewayKeyPrefix, OfferKeyPrefix, ProposeDecryptionShareKeyPrefix,
     };
-    use crate::LightningGen;
+    use crate::{ContractAccount, LightningGateway, LightningGen, LightningOutputOutcome};
+
+    async fn create_db_with_v0_data(db: Database) -> anyhow::Result<()> {
```
dpc, line 158: Function populating the empty db with data should probably take a `DatabaseTransaction` just to cut on the common code.

modules/fedimint-ln/src/db.rs
m1sterc001guy: Removed

modules/fedimint-ln/src/db.rs
m1sterc001guy: Outdated, backups are no longer made from running the integration tests

integrationtests/tests/fixtures/mod.rs
m1sterc001guy: Removed

integrationtests/tests/fixtures/mod.rs
m1sterc001guy: Removed

integrationtests/tests/fixtures/mod.rs
m1sterc001guy: Removed

fedimint-testing/src/lib.rs
m1sterc001guy: Moved the decoders to be an argument instead of geneating them here

modules/fedimint-ln/src/db.rs
```diff
@@ -121,140 +121,221 @@ impl_db_lookup!(
...
+            id: OfferId::from_str(
+                "0123456789012345678901234567890101234567890123456789012345678901",
+            )
+            .unwrap(),
+        };
+        dbtx.insert_new_entry(&contract_update_key, &lightning_output_outcome)
+            .await
+            .expect("Error inserting ContractUpdate");
+
+        let propose_decryption_share_key = ProposeDecryptionShareKey(contract_id);
```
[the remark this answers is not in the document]
m1sterc001guy: Tried to inline as best I could

modules/fedimint-ln/src/db.rs
m1sterc001guy: Done

modules/fedimint-ln/src/db.rs
m1sterc001guy: Done

modules/fedimint-ln/src/db.rs
m1sterc001guy: I think the indendation is much better after reducing the scope of the PR, hopefully its easier to read

m1sterc001guy:
I reduced the scope of the PR to add just one migration test for the Lightning module since I've been hitting a lot of merge conflicts while trying to do all the tests at once. I'll open separate PRs for the other modules and the global DB key/values.

modules/fedimint-ln/src/db.rs
m1sterc001guy: Do you think this is still needed? A macro might help with readability here, similar to what we did in `dbtool`

dpc (approved):

elsirion (approved):

PR #1805, opened 2023-03-02
feat: introduce ClientModuleGenRegistry

dpc:
We can't have client code use server-side module generators forever. This change introduces everything necessary to support "client side module gens" with ability to downgrade `ServerModuleGen` to `ClientModuleGen`.

### DIFF
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -28,7 +28,7 @@ use crate::module::audit::Audit;
 use crate::module::interconnect::ModuleInterconect;
 use crate::net::peers::MuxPeerConnections;
 use crate::server::{DynServerModule, VerificationCache};
-use crate::task::{MaybeSend, TaskGroup};
+use crate::task::{MaybeSend, MaybeSync, TaskGroup};
 use crate::{
     apply, async_trait_maybe_send, dyn_newtype_define, maybe_add_send, maybe_add_send_sync, Amount,
     OutPoint, PeerId,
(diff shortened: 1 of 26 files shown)


fedimint-core/src/module/mod.rs
```diff
@@ -263,6 +263,75 @@ where
     }
 }
 
+// TODO: II is ridicolous
```
m1sterc001guy, line 266: nit: typo

fedimint-core/src/module/mod.rs
```diff
@@ -263,6 +263,75 @@ where
...
+pub trait ICommonModuleGen: Debug {
```
m1sterc001guy, line 267: Add docs to trait?

fedimint-core/src/module/mod.rs
```diff
@@ -263,6 +263,75 @@ where
...
+    fn module_kind(&self) -> ModuleKind {
+        <Self as ICommonModuleGen>::module_kind(self)
+    }
+
+    fn hash_client_module(&self, config: serde_json::Value) -> anyhow::Result<sha256::Hash> {
+        <Self as ICommonModuleGen>::hash_client_module(self, config)
+    }
+}
+
+pub trait IClientModuleGen: Debug + MaybeSend + MaybeSync {
```
m1sterc001guy, line 327: Perhaps a naive question. I see `IClientModuleGen` and `ICommonModuleGen` share some function definitions? Can one trait derive from the other? Although I suppose it would be weird to do `ICommonModuleGen : IClientModuleGen`

dpc: I forgot to polish this part, and I've renamed some things and added some comments to make it more clear. Thanks for pointing it out!

fedimint-core/src/module/mod.rs
```diff
@@ -416,19 +511,18 @@ pub trait ModuleGen: Debug + Sized {
     /// to move from the previous database version to the current version.
     const DATABASE_VERSION: DatabaseVersion;
```
m1sterc001guy, line 512: At some point `ClientModuleGen` will also need to have Database Migrations right?

m1sterc001guy: Either way, I guess the `DATABASE_VERSION` shouldn't be shared

modules/fedimint-mint/src/lib.rs
```diff
@@ -145,17 +145,38 @@ pub struct VerifiedNotes {
...
+#[derive(Debug)]
+pub struct MintClientGen;
+
+#[apply(async_trait_maybe_send!)]
+impl ClientModuleGen for MintClientGen {
+    type Common = MintCommonGen;
+}
+
+#[derive(Debug)]
+pub struct MintGen;
```
m1sterc001guy, line 173: Nit: `MintServerGen` is probably better here

modules/fedimint-wallet/src/lib.rs
```diff
@@ -229,17 +229,36 @@ impl std::fmt::Display for WalletOutputOutcome {
...
+#[derive(Debug)]
+pub struct WalletClientGen;
+
+#[apply(async_trait_maybe_send!)]
+impl ClientModuleGen for WalletClientGen {
+    type Common = WalletCommonGen;
+}
+
+#[derive(Debug)]
+pub struct WalletGen;
```
m1sterc001guy, line 254: Nit: Same comment as `MintGen`

m1sterc001guy (approved):
LGTM, most comments are nits or clarifications on my part

fedimint-core/src/module/mod.rs
dpc: Yes. We'll have to figure out the details with time. The exactly client modules will probably have to come from the new client code.

modules/fedimint-mint/src/lib.rs
dpc: I'll rename in another PR to avoid too many merge conflicts.

elsirion (approved):

PR #1807, opened 2023-03-02
add esplora tests to CI

### LINKED ISSUE #1383: Add regtest electrs tests for peg-ins and peg-outs.
We have electrs support but it's not well tested well, and without automated tests it is going to be hard to keep it functional.

https://github.com/fedimint/fedimint/pull/1363#discussion_r1072901329

jkitman:
Adds tests for the esplora client (after fixing some issues with it).  @oleonardolima

Fixes #1383

### DIFF
--- a/fedimint-bitcoind/src/bitcoincore_rpc.rs
+++ b/fedimint-bitcoind/src/bitcoincore_rpc.rs
@@ -358,7 +358,10 @@ pub struct EsploraClient(esplora_client::AsyncClient);
 
 impl EsploraClient {
     fn new(url: &Url) -> anyhow::Result<Self> {
-        let builder = esplora_client::Builder::new(url.as_str());
+        // Url needs to have any trailing path including '/' removed
+        let without_trailing = url.as_str().trim_end_matches('/');
+
+        let builder = esplora_client::Builder::new(without_trailing);
         let client = builder.build_async()?;
         Ok(Self(client))
     }
--- a/scripts/setup-tests.sh
+++ b/scripts/setup-tests.sh
@@ -25,12 +25,16 @@ done
 $FM_BTC_CLIENT createwallet ""
 
 # Start electrs
-ELECTRS_CFG="$FM_ELECTRS_DIR"/electrs.toml
+ELECTRS_CFG="$FM_TEST_DIR/electrs.toml"
 echo 'auth = "bitcoin:bitcoin"' > $ELECTRS_CFG
 echo 'daemon_rpc_addr = "127.0.0.1:18443"' >> $ELECTRS_CFG
 echo 'daemon_p2p_addr = "127.0.0.1:18444"' >> $ELECTRS_CFG
 echo 'electrum_rpc_addr = "127.0.0.1:50001"' >> $ELECTRS_CFG
-electrs --network "regtest" --conf "$ELECTRS_CFG" --db-dir "$FM_ELECTRS_DIR" &
+electrs --network "regtest" --conf "$ELECTRS_CFG" --db-dir "$FM_TEST_DIR/electrs" --daemon-dir "$FM_BTC_DIR" &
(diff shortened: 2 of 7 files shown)


fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -358,7 +358,11 @@ pub struct EsploraClient(esplora_client::AsyncClient);
 
 impl EsploraClient {
     fn new(url: &Url) -> anyhow::Result<Self> {
-        let builder = esplora_client::Builder::new(url.as_str());
+        // Urls add a trailing '/' that needs to be removed
+        let mut without_trailing = url.to_string();
```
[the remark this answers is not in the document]
jkitman: Think I can write this in a safer way using `.origin()`

scripts/setup-tests.sh
```diff
@@ -25,12 +25,16 @@ done
...
 echo 'auth = "bitcoin:bitcoin"' > $ELECTRS_CFG
 echo 'daemon_rpc_addr = "127.0.0.1:18443"' >> $ELECTRS_CFG
 echo 'daemon_p2p_addr = "127.0.0.1:18444"' >> $ELECTRS_CFG
 echo 'electrum_rpc_addr = "127.0.0.1:50001"' >> $ELECTRS_CFG
-electrs --network "regtest" --conf "$ELECTRS_CFG" --db-dir "$FM_ELECTRS_DIR" &
+electrs --network "regtest" --conf "$ELECTRS_CFG" --db-dir "$FM_TEST_DIR/electrs" --daemon-dir "$FM_BTC_DIR" &
+echo $! >> $FM_PID_FILE
+
+# Start esplora
+esplora --cookie "bitcoin:bitcoin" --network "regtest" --daemon-dir "$FM_BTC_DIR" --http-addr "127.0.0.1:50002" --daemon-rpc-addr "127.0.0.1:18443" --monitoring-addr "127.0.0.1:50003" --db-dir "$FM_TEST_DIR/esplora" &
```
oleonardolima, line 37: @jkitman using [`--lightmode`](https://github.com/Blockstream/electrs#light-mode) would make it faster for the tests ?

dpc (dismissed):

scripts/setup-tests.sh
elsirion: I don't think it makes much of a difference since we spawn a fresh `regtest` network each time, so the indices are empty and even at the end of testing will only contain a handful of items.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -358,7 +358,10 @@ pub struct EsploraClient(esplora_client::AsyncClient);
 
 impl EsploraClient {
     fn new(url: &Url) -> anyhow::Result<Self> {
-        let builder = esplora_client::Builder::new(url.as_str());
+        // Url needs to have any trailing path including '/' removed
+        let without_trailing = url.origin().ascii_serialization();
```
elsirion, line 362: This will remove the path component of the URL altogether which can be problematic if the API is accessible at a non-root path. E.g. blockstream's testnet API:

```
https://blockstream.info/testnet/api/tx/<txid>/hex
```

elsirion (changes requested):
I think we should fix the URL serialization since it could lead to really weird bugs when someone tries to use this outside of tests.

fedimint-bitcoind/src/bitcoincore_rpc.rs
jkitman: Fixed this, looks like we have a markdown check failure on master though.

elsirion (approved):
LGTM. Never mind the docs failure, I made it an optional test. #1820 fixes docs once merged.

dpc (approved):

Issue #1447, opened 2023-01-20
New Client Library Tracking Issue


justinmoon:
## Goals
* Parallelism
* Operation-separation (Tor circuit support)
* Modularity
* Correctness in all cases (crashes, restarts, …)

## Sub-tasks
- [x] Generalized `Decode` trait (#1633)
- [x] Define interfaces and implement state machine executor (#1621)
- [x] Move `ClientModuleGen` to `fedimint-client` #1828 
- [x] Ecash client
- [ ] Wallet client
- [ ] Lightning client
  - [x] Send
  - [x] Receive
  - [ ] Gateway
- [ ] Join federation
- [ ] Check for compatibility with federation via [versions](https://github.com/fedimint/fedimint/pull/2265) RPC 
- [ ] Report if consensus isn't running via https://github.com/fedimint/fedimint/pull/2272
- [ ] `fedimint-cli ng` can display some form of transaction history

justinmoon:
Close in favor of https://github.com/fedimint/fedimint/issues/2514

PR #1810, opened 2023-03-03
Split some modules (dummy & mint) into subcrates

dpc:
On top #1805 

Gradually removing client dependency on server side module code.


client/cli/Cargo.toml
```diff
@@ -20,7 +20,8 @@ lightning-invoice = { version = "0.21.0", features = [ "serde" ] }
 mint-client = { path = "../client-lib" }
 fedimint-core ={ path = "../../fedimint-core" }
 fedimint-rocksdb = { path = "../../fedimint-rocksdb" }
-fedimint-mint = { path = "../../modules/fedimint-mint" }
+fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
```
dpc, line 23: We only depend on client side here, yay!

client/client-lib/Cargo.toml
```diff
@@ -64,3 +64,4 @@ test-log = { version = "0.2", features = [ "trace" ], default-features = false }
 tracing-subscriber = { version = "0.3.16", features = [ "env-filter" ] }
 fedimint-testing = { path = "../../fedimint-testing" }
 once_cell = "1.16.0"
+fedimint-mint-server = { path = "../../modules/fedimint-mint-server" }
```
dpc, line 67: dev-deps only, as tests use it.

fedimintd/Cargo.toml
```diff
@@ -41,7 +41,7 @@ fedimint-rocksdb = { path = "../fedimint-rocksdb" }
 fedimint-server = { path = "../fedimint-server" }
 fedimint-logging = { path = "../fedimint-logging", features = ["telemetry"] }
 fedimint-wallet = { path = "../modules/fedimint-wallet", features = ["native", "server"] }
-fedimint-mint= { path = "../modules/fedimint-mint", features = ["server"] }
+fedimint-mint-server = { path = "../modules/fedimint-mint-server" }
```
dpc, line 44: No more conditional compilation.

jkitman (approved):
So does `fedimint-server` no longer depend on `fedimint-mint-client`?

Will this be possible with to perform on the other modules as well?

dpc:
> So does fedimint-server no longer depend on fedimint-mint-client?

It does, but that doesn't bother me. If fedimint-server needs to use client-side calls, then it can depend on client module.

It's the other way around that is more of a problem.

> Will this be possible with to perform on the other modules as well?

That's the plan, but I rather rebase smaller commits than one large one.

elsirion (approved):

elsirion:
Great work, that will come in handy in #1447.

Issue #1729, opened 2023-02-21
New DatabaseTransaction wrapper type that does not have commit


m1sterc001guy:
As described in this [issue](https://github.com/fedimint/fedimint/issues/1635) and discussed in this [PR](https://github.com/fedimint/fedimint/pull/1695), we should introduce a new public wrapper type for DatabaseTransaction that does not have `commit_tx`. We would be able to give these db transactions to the modules so that the modules can insert/delete data from the database, but they would not have the ability to commit. This new type will make this explicit since the `commit_tx` won't even be available on the database transaction wrapper.

PR #1811, opened 2023-03-03
refactor: expose ModuleDatabaseTransaction for modules

m1sterc001guy:
Today we wrap `IsolatedDatabaseTransaction` inside `DatabaseTransaction` and only expose `DatabaseTransaction` to the modules. This has some downsides, such as the fact that `commit_tx` is exposed as a possible function for the modules to call. This should not be allowed though, since modules should not be allowed to commit to the database because the fedimint server is responsible for managing the lifetime of a database transaction. Today if this happens, `commit_tx` will panic.

This PR is a refactor to expose `ModuleDatabaseTransaction` to the modules and add the public facing database transaction API to `ModuleDatabaseTransaction`, but removed the functions that modules should not be allowed to call. This allows us to keep the wrapper functionality that we have introduced in previous PRs, but also address the issue of not exposing the full database transaction API.

Fixes: https://github.com/fedimint/fedimint/issues/1729

### DIFF
--- a/fedimint-client/src/sm/state.rs
+++ b/fedimint-client/src/sm/state.rs
@@ -5,7 +5,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use fedimint_core::core::ModuleInstanceId;
-use fedimint_core::db::DatabaseTransaction;
+use fedimint_core::db::IsolatedDatabaseTransaction;
 use fedimint_core::dyn_newtype_define_with_instance_id;
 use fedimint_core::encoding::{Decodable, DynEncodable, Encodable};
 use futures::future::BoxFuture;
@@ -83,7 +83,11 @@ where
 type TriggerFuture = Pin<Box<dyn Future<Output = serde_json::Value> + Send + 'static>>;
 // TODO: remove Arc, maybe make it a fn pointer?
 type StateTransitionFunction<S> = Arc<
-    dyn for<'a> Fn(&'a mut DatabaseTransaction<'_>, serde_json::Value, S) -> BoxFuture<'a, S>
+    dyn for<'a> Fn(
+            &'a mut IsolatedDatabaseTransaction<'_, '_, ModuleInstanceId>,
+            serde_json::Value,
+            S,
+        ) -> BoxFuture<'a, S>
         + Send
         + Sync,
 >;
@@ -129,7 +133,11 @@ impl<S> StateTransition<S> {
         S: Send + Sync + Clone + 'static,
         V: serde::Serialize + serde::de::DeserializeOwned + Send,
         Trigger: Future<Output = V> + Send + 'static,
-        TransitionFn: for<'a> Fn(&'a mut DatabaseTransaction<'_>, V, S) -> BoxFuture<'a, S>
+        TransitionFn: for<'a> Fn(
+                &'a mut IsolatedDatabaseTransaction<'_, '_, ModuleInstanceId>,
+                V,
+                S,
+            ) -> BoxFuture<'a, S>
             + Send
             + Sync
             + Copy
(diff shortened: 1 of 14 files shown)


fedimint-core/src/core/server.rs
```diff
@@ -204,14 +211,17 @@ where
     }
 
     /// Blocks until a new `consensus_proposal` is available.
-    async fn await_consensus_proposal(&self, dbtx: &mut DatabaseTransaction<'_>) {
```
dpc, line 207: Do we want IsolatedDatabaseTransaction here? It is exposing outside the implementation detail of modules getting isolated. Is it an attempt to make sure we never forget about isolation on the caller side?

dpc: Oh. I thought I'm looking at a different PR of yours ...

dpc:
Seems to me that a proper way to handle this is to have have separate newtype that just doesn't expose `commit_tx`. 

This way user (module developer) code is not tied to a concrete type and an implementation detail.

m1sterc001guy:
> Seems to me that a proper way to handle this is to have have separate newtype that just doesn't expose `commit_tx`.
> 
> This way user (module developer) code is not tied to a concrete type and an implementation detail.

But the implementation details of `IsolatedDatabaseTransaction` are hidden right? I believe the only public functions are `get_value`, `insert_entry`, `insert_new_entry`, `remove_entry`, `find_by_prefix`, and `remove_by_prefix`. I don't think the modules can call the raw functions, which is where the details on how the prefix is appended/prepended are handled.

dpc:
> But the implementation details of `IsolatedDatabaseTransaction` are hidden right?

Yes, but the `IsolatedDatabaseTransaction` has the implementation details kind of right there in the name. :)

m1sterc001guy:
> Yes, but the IsolatedDatabaseTransaction has the implementation details kind of right there in the name. :)

Do you think just renaming `IsolatedDatabaseTransaction` would be sufficient? I suppose we could introduce a new type but I'm worried too many wrapper types will become confusing.

Perhaps:

`IsolatedDatabaseTransaction` becomes `ModuleDatabaseTransaction`
The existing `ModuleDatabaseTransaction` becomes `WrappedIsolatedDatabaseTransaction`. This will be private since it is only used inside of fedimint_core::db::mod.rs.

I'm terrible at coming up with names so those are probably still not the best.

dpc:
> `IsolatedDatabaseTransaction` becomes `ModuleDatabaseTransaction`
> The existing `ModuleDatabaseTransaction` becomes `WrappedIsolatedDatabaseTransaction`. This will be private since it is only used inside of fedimint_core::db::mod.rs.

`ModuleDatabaseTransaction` sounds reasonable to me. Though my main concern is just avoiding tying the type we use in the module API to any concrete implementation, as in the future we might want to add some other functionality etc. that would be otherwise invisible to the user.

elsirion:
Needs rebase after #1810

m1sterc001guy:
> Though my main concern is just avoiding tying the type we use in the module API to any concrete implementation, as in the future we might want to add some other functionality etc. that would be otherwise invisible to the user.

Ok I'm decently happy with the most recent renames. Here is a description of each type and its purpose:

`DatabaseTransaction`: existing public type, used for atomic db transactions across modules
`ModuleDatabaseTransaction`: new public type, cannot commit to db, doesn't contain implementation details on isolation
`IsolatedDatabaseTransaction`: existing private type, contains implementation details on prepending/striping prefixes
`CommitableIsolatedDatabaseTransaction`: renamed private type, used for atomic db transactions within a module. This type will always be wrapped inside `DatabaseTransaction` so modules aren't aware of it.

Comments in code have been updated

dpc (approved):

fedimint-client/src/sm/state.rs
```diff
@@ -84,7 +84,7 @@ type TriggerFuture = Pin<Box<dyn Future<Output = serde_json::Value> + Send + 'st
 // TODO: remove Arc, maybe make it a fn pointer?
 type StateTransitionFunction<S> = Arc<
     dyn for<'a> Fn(
-            &'a mut IsolatedDatabaseTransaction<'_, '_, ModuleInstanceId>,
+            &'a mut ModuleDatabaseTransaction<'_, ModuleInstanceId>,
```
elsirion, line 87: nit: `ModuleDatabaseTrabsaction` could be a type def instead:

```rust
type ModuleDatabaseTransaction<'a, 'b> = IsolatedDatabaseTransaction<'a, 'b, ModuleInstanceId>;
```

elsirion (approved):

fedimint-client/src/sm/state.rs
dpc: That might be a little bit leaky though.

PR #1815, opened 2023-03-05
added log statements before sleep in tests

### LINKED ISSUE #1792: Any sleep or wait in tests should be preceeded by log statement
https://github.com/fedimint/fedimint/pull/1781#discussion_r1119029229

PaarthAgarwal:
An attempt to fix #1792.
cc: @justinmoon


dhruv-1001:
@justinmoon according to the linked issue, I believe you meant to add logs before sleep only in test-scripts, right?

dhruv-1001:
Just a quick fix: your commit message should follow the convention, which is:

1. Start with a tag (in your case chore:)
2. Present-tense statement answering the question: "What does this commit do?"

So for this one it would look something like this: `chore: add log statements before sleep in tests`

PR #1818, opened 2023-03-06
chore: replace link checking action

dpc:
The one we're using has terrible output.


dpc:
Damn it. This one seems only able to crawl web-server. :/

PR #1813, opened 2023-03-03
fix: identifiable contract trait typo

oleonardolima:
- rename all `IdentifyableContract` trait references to`IdentifiableContract`

~**NOTE:** needs to land on top of #1812~ not anymore

PR #1819, opened 2023-03-06
fix: continue-on-error for link checker

dpc:


dpc:
It still shows up as a Required here, but maybe it's because it takes setting from `master` branch?

elsirion:
> It still shows up as a Required here

I think this is a GitHub UI setting and not defined in the code? Afaik continue on error just makes the job not fail if the step fails, that's not what we want though. I will make it non-required in the UI, also see #1820 for a fix for the current problem.

elsirion:
Done, closing this one.

Screenshot 2023-03-06 at 11-15-35 fix identifiable contract trait typo by oleonardolima · Pull Request #1813 · fedimint_fedimint

PR #1825, opened 2023-03-07
Add RBF support

### LINKED ISSUE #233: Improve peg-outs with dynamic fees
**Currently peg-outs have several serious issues that need to be addressed:**
- Fed can pay more fees than users contribute resulting in a negative balance (fed crashes)
- Fed can accept peg-outs that depend on unconfirmed change (fed crashes)
- No way to bump fees on unconfirmed transactions (if on-chain fees suddenly go up, peg-outs can get stuck indefinitely)

**Step 1: Add user-determined fees.**
- Replace the fixed user peg-out fee with an exact fee based on the current mempool (eliminating peg-out profits/losses for the fed):
- Expose a wallet API that given a `(tx amount, peg-out address)` returns the `PegOutFees = (total sats, sat/byte)`
- Users include the `PegOutFees` in their `PegOut` requests which can be rejected during consensus proposal if the fed cannot fulfill the request (due to not enough fees or `SpendableUTXO`)
- The fed requires `FeeConsensus < actual fees == PegOutFees` (target 10 blocks to confirm)
- Peg-outs are immediately signed and broadcast (rather than aggregated and waiting for an urgency heuristic), so there will be one tx per peg-out request
- Since subsequent peg-outs may need to wait for change (~10-20 blocks), it is recommended that the fed maintains a few `SpendableUTXO` to handle concurrent requests, although peg-out volume should be low for early use-cases
- Fixed by https://github.com/fedimint/fedimint/pull/276

**Step 2: Allow for user RBF**
- Allow users to output additional fees to existing transactions either via a new `BumpFee` output type or by modifying the existing `PegOut`
- When a `BumpFee` is accepted the federation should sign and RBF the existing peg-out transaction with the additional fees
- After the tx reaches confirmation finality any fees that did not get confirmed are made redeemable to the user through a mechanism where the user proves they created the initial `PegOut`

**Step 3: Aggregate or Queue transactions**
- Long-term if peg-out volume increases eventually users may have to wait for change because the fed cannot maintain enough `SpendableUTXO`
- In this case either transactions will need to be queued or aggregated so that users do not have to wait around and continually try to resubmit peg-outs
- In either case, RBF will need to be available so that users can adjust fees and receive change as we cannot know ahead of time which transactions will be confirmed on-chain or how fees might increase after a peg-out is submitted.

This issue replaces #1 and #196 with a more up-to-date roadmap for peg-outs.

### LINKED ISSUE #1480: Handling peg-out failures
I was just chatting with @elsirion about how the client library handles peg-out failures. For example, let's say the fee-rate changes and it never confirms?

He mentioned that he had chatted with @jkitman about this some time ago, but wasn't sure if the changes were ever implemented.

jkitman:
- Allows clients to submit RBF transactions to bump the fees of a stuck transaction
- Note there is a risk the fed could bank a small profit if in a rare case the non-RBF tx confirms first

**Follow-up fee fix**
- Need to add checks to avoid fees set below `minRelayFee` that could cause `submit_transaction` to fail
- Switch to sat/vb for clarity
- Confirm that the actual tx weight matches the `PegOutFees` paid for (could not if it pulls in an additional UTXO)

Fixes #1480
Fixes #233

### DIFF
--- a/client/client-lib/Cargo.toml
+++ b/client/client-lib/Cargo.toml
@@ -31,6 +31,7 @@ fedimint-ln-client = { path = "../../modules/fedimint-ln-client" }
 fedimint-logging = { path = "../../fedimint-logging" }
 fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
 fedimint-wallet-client = { path = "../../modules/fedimint-wallet-client", default-features = false }
+fedimint-wallet-common = { path = "../../modules/fedimint-wallet-common", default-features = false }
 itertools = "0.10.5"
 rand = "0.8"
 ring = "0.16.20"
--- a/integrationtests/tests/tests.rs
+++ b/integrationtests/tests/tests.rs
@@ -6,7 +6,7 @@ use anyhow::Result;
 use assert_matches::assert_matches;
 use bitcoin::{Amount, KeyPair};
 use fedimint_core::task::TaskGroup;
-use fedimint_core::{msats, sats, TieredMulti};
+use fedimint_core::{msats, sats, Feerate, TieredMulti};
 use fedimint_ln_client::contracts::{Preimage, PreimageDecryptionShare};
 use fedimint_ln_client::LightningConsensusItem;
 use fedimint_logging::LOG_TEST;
@@ -16,8 +16,8 @@ use fedimint_server::consensus::TransactionSubmissionError::{
 };
 use fedimint_server::epoch::ConsensusItem;
 use fedimint_server::transaction::TransactionError::UnbalancedTransaction;
-use fedimint_wallet_server::common::PegOutSignatureItem;
 use fedimint_wallet_server::common::WalletConsensusItem::PegOutSignature;
+use fedimint_wallet_server::common::{PegOutFees, PegOutSignatureItem, Rbf};
 use fixtures::{rng, secp, sha256};
 use futures::future::{join_all, Either};
 use mint_client::mint::MintClient;
@@ -25,6 +25,7 @@ use mint_client::transaction::legacy::Output;
 use mint_client::transaction::TransactionBuilder;
 use mint_client::{ClientError, ConfigVerifyError};
 use threshold_crypto::{SecretKey, SecretKeyShare};
+use tracing::log::warn;
 use tracing::{debug, info, instrument};
 
 use crate::fixtures::{peers, test, unwrap_item, FederationTest};
@@ -78,7 +79,7 @@ async fn wallet_peg_in_and_peg_out_with_fees() -> Result<()> {
             bitcoin.mine_block_and_get_received(&peg_out_address).await,
             sats(peg_out_amount)
         );
-        user.assert_total_notes(sats(peg_in_amount - peg_out_amount) - fees)
+        user.assert_total_notes(sats(peg_in_amount - peg_out_amount) - fees.amount().into())
             .await;
         assert_eq!(fed.max_balance_sheet(), 0);
     })
@@ -115,6 +116,7 @@ async fn wallet_peg_outs_are_only_allowed_once_per_epoch() -> Result<()> {
 
         fed.mine_and_mint(&user, &*bitcoin, sats(5000)).await;
         let (fees, _) = user.peg_out(1000, &address1).await;
+        let fees = fees.amount().into();
         user.peg_out(1000, &address2).await;
         info!(target: LOG_TEST, ?fees, "Tx fee");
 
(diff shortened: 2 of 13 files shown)


modules/fedimint-wallet/src/lib.rs
```diff
@@ -1351,20 +1417,52 @@ impl Wallet {
...
+                tx.selected_utxos.extend(available_utxos);
+
+                self.offline_wallet().create_tx(
+                    tx.peg_out_amount,
+                    tx.destination,
+                    tx.selected_utxos,
+                    tx.fees.fee_rate,
+                    &change_tweak,
+                    Some(rbf.clone()),
+                )
```
elsirion, line 1463: Afaik there isn't any interface contract in `create_tx` that says the first UTXOs will always be chosen. Maybe we should split the argument into "for sure" inputs and "potential additional inputs"? There should also be a comment about the dangers of not using the previous UTXOs (double spends, Fedimint forgetting about them, …) so someone doesn't "optimize" it later.

On another note, the ordering by amount should probably happen inside `create_tx` since the algorithm in there likely assumes that to be the case.

modules/fedimint-wallet/src/lib.rs
```diff
@@ -1450,8 +1550,6 @@ impl<'a> StatelessWallet<'a> {
         let mut selected_utxos: Vec<(UTXOKey, SpendableUTXO)> = vec![];
         let mut fees = fee_rate.calculate_fee(total_weight);
 
-        // When selecting UTXOs we select from largest to smallest amounts
-        utxos.sort_by_key(|(_, utxo)| utxo.amount);
```
elsirion, line 1454: Ah I see, you just moved sorting outside.

modules/fedimint-wallet/src/lib.rs
```diff
@@ -1687,24 +1789,31 @@ pub async fn run_broadcast_pending_tx(db: Database, rpc: DynBitcoindRpc, tg_hand
...
+        if !rbf_txids.contains(&tx.txid()) {
+            debug!(
+                tx = %tx.txid(),
+                weight = tx.weight(),
+                output = ?tx.output,
+                "Broadcasting peg-out",
+            );
+            trace!(transaction = ?tx);
+            let _ = rpc.submit_transaction(tx).await;
+        }
```
elsirion, line 1816: This should be perfomant enough for now, eventually we could split the RBFed and non-RBFed txns into different DB "tables".

modules/fedimint-wallet/src/lib.rs
```diff
@@ -699,29 +733,31 @@ impl ServerModule for Wallet {
...
+                .into_module_error_other();
+            }
+            let consensus_fee_rate = self.current_round_consensus(dbtx).await.unwrap().fee_rate;
+            if pegout.fees.fee_rate < consensus_fee_rate {
+                return Err(WalletError::PegOutFeeRate(
+                    pegout.fees.fee_rate,
+                    consensus_fee_rate,
+                ))
+                .into_module_error_other();
+            }
```
elsirion, line 754: Why don't we validate at least that the RBFed tx exists in case of `WalletOutput::Rbf`?

elsirion (changes requested):

elsirion:
Looks really good overall, just some concerns around API misuse resistance since this is one of the most crucial parts of the code.

modules/fedimint-wallet/src/lib.rs
jkitman: Sure, did think about doing it that way, just thought this was a simpler interface, but does require the caller to order the UTXOs themselves.

modules/fedimint-wallet/src/lib.rs
jkitman: That will fail during `create_peg_out_tx` a few lines down.

elsirion (dismissed):

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1091,6 +1093,42 @@ impl Wallet {
...
+
+            // Search for tx that wanted to RBF the `removed` one
+            for tx in all_transactions.values() {
+                if let Some(rbf) = &tx.rbf {
+                    if rbf.txid == removed.tx.txid() {
+                        pending_to_remove.push(tx.clone());
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
jkitman: I'll refactor to a separate function, it's complicated because you basically need to search for all related txs in case of multiple RBFs.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -474,29 +475,31 @@ impl ServerModule for Wallet {
...
-            .into_module_error_other();
-        }
-        let consensus_fee_rate = self.current_round_consensus(dbtx).await.unwrap().fee_rate;
-        if output.fees.fee_rate < consensus_fee_rate {
-            return Err(WalletError::PegOutFeeRate(
-                output.fees.fee_rate,
-                consensus_fee_rate,
-            ))
-            .into_module_error_other();
+        if let WalletOutput::PegOut(pegout) = output {
```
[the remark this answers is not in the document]
jkitman: Yep, that's validated in `create_peg_out_tx`.  I'm going to follow-up with a PR to add/refactor/test the validations.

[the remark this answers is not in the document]
justinmoon: Ahh ok. Looks good, then.

client/client-lib/Cargo.toml
```diff
@@ -31,6 +31,7 @@ fedimint-ln-client = { path = "../../modules/fedimint-ln-client" }
 fedimint-logging = { path = "../../fedimint-logging" }
 fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
 fedimint-wallet-client = { path = "../../modules/fedimint-wallet-client", default-features = false }
+fedimint-wallet-common = { path = "../../modules/fedimint-wallet-common", default-features = false }
```
dpc, line 34: Not a blocker but my plan was for client and server always to re-export common, so downstream users should always import just client or server.

dpc (approved):

integrationtests/tests/tests.rs
```diff
@@ -141,6 +143,65 @@ async fn wallet_peg_outs_are_only_allowed_once_per_epoch() -> Result<()> {
...
+
+        assert_eq!(
+            bitcoin.mine_block_and_get_received(&address).await,
+            sats(1000)
+        );
+        bitcoin
+            .mine_blocks(fed.wallet.consensus.finality_delay as u64)
+            .await;
+        fed.run_consensus_epochs(1).await;
+        assert_eq!(fed.max_balance_sheet(), 0);
```
m1sterc001guy, line 200: Why this check? Does the test assume it is sweeping all the funds from the federation?

jkitman: Good question!  If the federation doesn't set fees, it should never be at a profit.

This check verifies RBF `PendingTransactions` are counted as liabilities rather than assets (in the normal case `PendingTransactions` represent a change UTXO asset we will receive)

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -660,13 +653,15 @@ impl ServerModule for Wallet {
             .add_items(dbtx, &UTXOPrefixKey, |_, v| v.amount.to_sat() as i64 * 1000)
             .await;
         audit
-            .add_items(dbtx, &UnsignedTransactionPrefixKey, |_, v| {
-                v.change.to_sat() as i64 * 1000
+            .add_items(dbtx, &UnsignedTransactionPrefixKey, |_, v| match v.rbf {
+                None => v.change.to_sat() as i64 * 1000,
+                Some(rbf) => rbf.fees.amount().to_sat() as i64 * -1000,
```
m1sterc001guy, line 658: Do we have a millisats to sats constant? Might be more clear here.

Also, I assume the negative here is what makes the `PendingTransaction` a liability?

justinmoon: I found this whole function a little confusing as well with the i64 and negatives ...

m1sterc001guy (approved):

modules/fedimint-wallet-server/src/lib.rs
jkitman: Negatives are liabilities, positives are assets, everything is millisat denominated so we need to multiply by 1000.

Could definitely use comments.

dpc: Oh. Makes me think we should `TryFrom` instead of `as` to eliminate any chance for silent integer conversion issues.

jkitman: Don't think it should be an issue for `u64` as we'd need to have more sats than could ever exist :smile:

dpc: With msats and some feerate calculations, maybe, maybe it would be possible to theoretically? Anyway - I think it's just for the habit of it. :D

Issue #231, opened 2022-07-09
Remove polling


Maan2003:
- [x] database notifcations
- [x] no polling for transaction output.
- [ ] no polling for modules endpoints

dpc:
Being worked on in #1197

Maan2003:
This is completed.

Issue #796, opened 2022-10-15
I'm concerned about `bitvec` dependency


dpc:
See: https://github.com/zkcrypto/ff/issues/91

dpc:
The author responded, there has been some progress. Nothing actionable anymore.

PR #1826, opened 2023-03-07
feat: add notifications in database for modifying keys

Maan2003:
separated from #231

motivation: useful in new client.

### DIFF
--- a/fedimint-core/Cargo.toml
+++ b/fedimint-core/Cargo.toml
@@ -43,6 +43,7 @@ rand = "0.8.5"
 miniscript = { version = "7.0.0", git = "https://github.com/rust-bitcoin/rust-miniscript/", rev = "2f1535e470c75fad85dbad8633986aae36a89a92", features = [ "compiler", "serde" ] }
 secp256k1-zkp = { version = "0.7.0", features = [ "use-serde", "bitcoin_hashes", "global-context" ] }
 macro_rules_attribute = "0.1.3"
+bitvec = "1.0.1"
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 jsonrpsee-ws-client = "0.16.2"
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -13,7 +13,7 @@ use macro_rules_attribute::apply;
 use serde::Serialize;
 use strum_macros::EnumIter;
 use thiserror::Error;
-use tracing::{debug, info, instrument, trace, warn};
+use tracing::{debug, error, info, instrument, trace, warn};
 
 use crate::core::ModuleInstanceId;
 use crate::encoding::{Decodable, Encodable};
@@ -22,9 +22,11 @@ use crate::task::{MaybeSend, MaybeSync};
 use crate::{async_trait_maybe_send, maybe_add_send};
 
 pub mod mem_impl;
+pub mod notifications;
 
-pub use tests::*;
+pub use test_utils::*;
 
+use self::notifications::{Notifications, NotifyingTransaction};
 use crate::module::registry::ModuleDecoderRegistry;
 
 pub const MODULE_GLOBAL_PREFIX: u8 = 0xff;
@@ -37,6 +39,7 @@ pub trait DatabaseKeyPrefix: Debug {
 /// Extends `DatabaseKeyPrefix` to prepend the key's prefix.
 pub trait DatabaseRecord: DatabaseKeyPrefix {
     const DB_PREFIX: u8;
+    const NOTIFY_ON_MODIFY: bool = false;
     type Key: DatabaseKey + Debug;
     type Value: DatabaseValue + Debug;
 }
@@ -58,8 +61,19 @@ where
 /// `DatabaseKey` that represents the lookup structure for retrieving key/value
 /// pairs from the database.
 pub trait DatabaseKey: Sized {
+    /// Send a notification to tasks waiting to be notified if the value of
+    /// `DatabaseKey` is modified
+    ///
+    /// For instance, this can be used to be notified when a key in the
+    /// database is created. It is also possible to run a closure with the
+    /// value of the `DatabaseKey` as parameter to verify some changes to
+    /// that value.
+    const NOTIFY_ON_MODIFY: bool = false;
     fn from_bytes(data: &[u8], modules: &ModuleDecoderRegistry) -> Result<Self, DecodingError>;
 }
+
+/// Marker trait for `DatabaseKey`s where `NOTIFY` is true
+pub trait DatabaseKeyWithNotify {}
 
 /// `DatabaseValue` that represents the value structure of database records.
 pub trait DatabaseValue: Sized + Debug {
@@ -84,6 +98,7 @@ pub struct Database {
 // because it will impossible to construct otherwise
 #[derive(Debug)]
 struct DatabaseInner<Db: IDatabase + ?Sized> {
+    notifications: Notifications,
     module_decoders: ModuleDecoderRegistry,
     db: Db,
 }
@@ -116,6 +131,7 @@ impl Database {
     pub fn new(db: impl IDatabase + 'static, module_decoders: ModuleDecoderRegistry) -> Self {
         let inner = DatabaseInner {
             db,
+            notifications: Notifications::new(),
             module_decoders,
         };
 
@@ -141,6 +157,7 @@ impl Database {
         let dbtx = DatabaseTransaction::new(
             self.inner_db.db.begin_transaction().await,
             self.inner_db.module_decoders.clone(),
+            &self.inner_db.notifications,
         );
 
         match self.module_instance_id {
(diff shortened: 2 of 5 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -37,6 +39,7 @@ pub trait DatabaseKeyPrefix: Debug {
 /// Extends `DatabaseKeyPrefix` to prepend the key's prefix.
 pub trait DatabaseRecord: DatabaseKeyPrefix {
     const DB_PREFIX: u8;
+    const NOTIFY: bool = false;
```
elsirion, line 42: Please document what it does. I assume it activates notifications for this key?

fedimint-core/src/db/mod.rs
```diff
@@ -761,9 +866,10 @@ impl<'parent> DatabaseTransaction<'parent> {
     pub fn new(
         dbtx: Box<dyn ISingleUseDatabaseTransaction<'parent>>,
         decoders: ModuleDecoderRegistry,
+        notifications: &'parent Notifications,
     ) -> DatabaseTransaction<'parent> {
         DatabaseTransaction {
-            tx: dbtx,
+            tx: Box::new(NotifyingTransactionWrapper::new(dbtx, notifications)),
```
elsirion, line 872: With all these wrappers having some architecture diagram in the docs would be very useful. Maybe you can work together with @m1sterc001guy on that?

It's becoming increasingly hard to keep the full architecture in my mind and up to date, so I imagine it's even worse for new contributors. [Mermaid](https://github.com/mermaid-js/mermaid) in markdown docs works quite well for these kind of things since it's easily diffable.

fedimint-core/src/db/mod.rs
```diff
@@ -216,6 +223,80 @@ impl Database {
...
+    ) -> Result<(T, DatabaseTransaction<'a>)>
+    where
+        K: DatabaseKey + DatabaseRecord,
+    {
+        // not compile time yet :/
+        assert!(
+            <K as DatabaseKey>::NOTIFY,
+            "wait_key called with key({}) without NOTIFY flag",
+            std::any::type_name::<K>()
+        );
```
elsirion, line 244: I wonder if `const _: () = assert!(<K as DatabaseKey>::NOTIFY);` would work ([source stackoverflow](https://stackoverflow.com/a/70925999)).

Alternatively there's always type-level booleans, not that they'd be a good idea typically :laughing: 

```rust
pub trait Bool {
    fn to_bool() -> bool;
}

pub struct True {}
pub struct False {}

impl Bool for True {
    fn to_bool() -> bool {
        true
    }
}
impl Bool for False {
    fn to_bool() -> bool {
        false
    }
}
```

elsirion (approved):
Looks good, but I'd like @m1sterc001guy to have a thorough look again too and maybe also @dpc.

fedimint-core/src/db/mod.rs
```diff
@@ -216,6 +223,80 @@ impl Database {
                 .expect("db autocommit retry counter overflowed");
         } // end of loop
     }
+
+    /// Waits for key to be notified.
+    ///
+    /// Calls the `f` on every notifications. Return the value when `f` returns
```
m1sterc001guy, line 229: nit grammar: "notifications"

fedimint-core/src/db/mod.rs
```diff
@@ -216,6 +223,80 @@ impl Database {
...
+            <K as DatabaseKey>::NOTIFY,
+            "wait_key called with key({}) without NOTIFY flag",
+            std::any::type_name::<K>()
+        );
+        let key_bytes = if let Some(module_id) = self.module_instance_id {
+            let mut prefix_bytes = vec![MODULE_GLOBAL_PREFIX];
+            module_id
+                .consensus_encode(&mut prefix_bytes)
+                .expect("Error encoding module instance id as prefix");
+            prefix_bytes.extend(key.to_bytes());
```
m1sterc001guy, line 250: Can we make this a function? `IsolatedDatabaseTransaction` uses very similar logic

fedimint-core/src/db/mod.rs
```diff
@@ -396,6 +483,13 @@ impl<'a, Tx: IDatabaseTransaction<'a> + Send> ISingleUseDatabaseTransaction<'a>
             .await;
         Ok(())
     }
+
+    fn add_notification_key(&mut self, key: &[u8]) -> Result<()> {
+        self.0
+            .as_mut()
+            .context("Cannot set a tx savepoint on an already consumed transaction")?
```
m1sterc001guy, line 490: Update error message

fedimint-core/src/db/mod.rs
m1sterc001guy: +1 Created an issue: https://github.com/fedimint/fedimint/issues/1831

fedimint-core/src/db/mod.rs
```diff
@@ -2070,3 +2196,138 @@ mod tests {
...
+    async fn test_wait_key_no_transaction() {
+        let db = Database::new(MemDatabase::new(), ModuleDecoderRegistry::default());
+
+        let key_task = waiter(&db, TestKey(1)).await;
+        assert_eq!(
+            future_returns_shortly(async { key_task.await.unwrap() }).await,
+            None,
+            "should not notify"
+        );
+    }
```
m1sterc001guy, line 2332: Can we add a test that tests using an IsolatedDatabase? (i.e `db.new_isolated(<module_instance_id>)`)

fedimint-core/src/db/notifications.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    pub fn add<K>(&mut self, key: &K)
+    where
+        K: Hash,
+    {
+        let mut hasher = DefaultHasher::new();
+        key.hash(&mut hasher);
+        let hash_value = hasher.finish();
+        let slot_idx = (hash_value % (self.buckets.len() as u64)) as usize;
```
m1sterc001guy, line 102: Maybe make the above a helper function? Its repeated in a few places

m1sterc001guy: `get_slot_index` ?

fedimint-core/src/db/notifications.rs
```diff
@@ -0,0 +1,236 @@
...
+        self.buckets.set(slot_idx, true);
+    }
+}
+
+/// Wrapper to implements `add_notification_key`
+pub struct NotifyingTransactionWrapper<'a> {
```
m1sterc001guy, line 108: Not sure if I like "Wrapper". Maybe `NotifyingDatabaseTransaction`? That could be clear

m1sterc001guy (commented):
Generally looks good to me, left a couple suggested changes

fedimint-core/Cargo.toml
```diff
@@ -43,6 +43,7 @@ rand = "0.8.5"
 miniscript = { version = "7.0.0", git = "https://github.com/rust-bitcoin/rust-miniscript/", rev = "2f1535e470c75fad85dbad8633986aae36a89a92", features = [ "compiler", "serde" ] }
 secp256k1-zkp = { version = "0.7.0", features = [ "use-serde", "bitcoin_hashes", "global-context" ] }
 macro_rules_attribute = "0.1.3"
+bitvec = "1.0.1"
```
justinmoon, line 46: @dpc we're not concerned about this anymore? (https://github.com/fedimint/fedimint/issues/796)

fedimint-core/src/db/mod.rs
```diff
@@ -216,6 +223,80 @@ impl Database {
...
+    /// a `Some(T)`
+    pub async fn wait_key_check<'a, K, T>(
+        &'a self,
+        key: &K,
+        f: impl Fn(Option<K::Value>) -> Result<Option<T>>,
+    ) -> Result<(T, DatabaseTransaction<'a>)>
+    where
+        K: DatabaseKey + DatabaseRecord,
+    {
+        // not compile time yet :/
```
justinmoon, line 239: Maybe make this a `FIXME` just so it's easier to search for ...

fedimint-core/src/db/notifications.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    use futures::Future;
+
+    use super::*;
+
+    async fn future_returns_shortly<F: Future>(sub: F) -> bool {
+        tokio::time::timeout(Duration::from_millis(10), sub)
+            .await
+            .is_ok()
+    }
```
justinmoon, line 190: Nit: This exact same function is defined twice. Perhaps we could move it to some kind of shared test utils crate.

fedimint-core/src/db/mod.rs
```diff
@@ -214,6 +225,74 @@ impl Database {
                 .expect("db autocommit retry counter overflowed");
         } // end of loop
     }
+
+    /// Waits for key to be notified.
+    ///
+    /// Calls the `f` on every notifications. Return the value when `f` returns
+    /// a `Some(T)`
+    pub async fn wait_key_check<'a, K, T>(
```
Maan2003, line 233: Needs to handle shutdowns. How to do this? should I just add argument `tg_handle: ..`

elsirion: Why would it need special handling here? The caller can just put a `select!` around the main future/loop and the stop signal at a higher level assuming this function is cancellation safe. I think we should handle shutdowns at the highest possible level.

fedimint-core/src/db/mod.rs
```diff
@@ -58,9 +61,14 @@ where
 /// `DatabaseKey` that represents the lookup structure for retrieving key/value
 /// pairs from the database.
 pub trait DatabaseKey: Sized {
+    /// Whether to send notifcation for modification of the key.
```
Maan2003, line 64: help needed for better doc

fedimint-core/Cargo.toml
dpc: The author is reputable and did some push to sorting outstanding issues some time after pointed out that they are lingering. 

The reality is that most of these crates have unpaid maintainers that are spread thin.

If a dependency can be avoided (especially non-trivial), it's better one way or another though.

fedimint-core/src/db/mod.rs
elsirion: @mxz42 to the rescue? :smile:

fedimint-core/src/db/notifications.rs
```diff
@@ -0,0 +1,242 @@
...
+    {
+        self.buckets.set(slot_index_for_key(key), true);
+    }
+}
+
+/// Wrapper to implements `add_notification_key`
+pub struct NotifyingTransaction<'a> {
+    // TODO: try removing Box
+    dbtx: Box<dyn ISingleUseDatabaseTransaction<'a>>,
+    // notifications to be submited after commit
```
mxz42, line 110: nit: typo: 
- `Wrapper to implement `add...`
- submitted

fedimint-core/src/db/mod.rs
```diff
@@ -214,6 +225,74 @@ impl Database {
                 .expect("db autocommit retry counter overflowed");
         } // end of loop
     }
+
+    /// Waits for key to be notified.
+    ///
+    /// Calls the `f` on every notification. Return the value when `f` returns
+    /// a `Some(T)`
+    pub async fn wait_key_check<'a, K, T>(
```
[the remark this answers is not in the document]
Maan2003: > explicit `'a`

IMO explicit is more readable in this case.

> `f` function. Could be `func`tion or `clo`sure.

`f` is very common in rust (see std). But `checker` is better in this case.

Maan2003:
rebase on #1829 and removed `Result`s

fedimint-core/src/db/mod.rs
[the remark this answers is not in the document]
mxz42: I think having lifetimes is never more readable than without? But it's certainly no blocker.


Yes, `f` is very common. though, as this `f` decides if we return or keep waiting, I agree that anything more expressive than `f` is better.

mxz42 (approved):

fedimint-core/src/db/mod.rs
```diff
@@ -214,6 +231,75 @@ impl Database {
                 .expect("db autocommit retry counter overflowed");
         } // end of loop
     }
+
+    /// Waits for key to be notified.
+    ///
+    /// Calls the `f` when value of the key may have changed.
```
m1sterc001guy, line 237: Update comment since named changed

m1sterc001guy (approved):

fedimint-core/src/db/mod.rs
```diff
@@ -58,9 +61,20 @@ where
...
 /// pairs from the database.
 pub trait DatabaseKey: Sized {
+    /// Send a notification to tasks waiting to be notified if the value of
+    /// `DatabaseKey` is modified
+    ///
+    /// For instance, this can be used to be notified when a key in the
+    /// database is created. It is also possible to run a closure with the
+    /// value of the `DatabaseKey` as parameter to verify some changes to
+    /// that value.
+    const NOTIFY_ON_MODIFY: bool = false;
```
elsirion, line 71: Love how this name rolls off the tongue :rofl:

elsirion (approved):

PR #1828, opened 2023-03-07
Move `ClientModuleGen` into `fedimint-client`

elsirion:
This is the second part of the Client NG effort (after #1621) and moves the `ClientModuleGen` trait into `fedimint-client` such that the `ClientModule` trait can stay there (would need to move into `fedimint-core` otherwise).

While this is far from pretty and doesn't represent a complete feature or anything I'd like to merge it sooner rather than later due to its invasiveness.


dpc (commented):
LGTM so far

dpc (approved):

justinmoon (approved):

PR #1832, opened 2023-03-07
Add checks, refactor, test the wallet validation

jkitman:
As a follow-up to RBF, add checks for if peg-outs `fee < min_relay_fee` or the `fee_tx_weight != actual_tx_weight`.

Refactors all the validation into a separate function in `StatelessWallet` for easier unit testing.

### DIFF
--- a/client/client-lib/src/wallet/mod.rs
+++ b/client/client-lib/src/wallet/mod.rs
@@ -340,10 +340,7 @@ mod tests {
         let output = PegOut {
             recipient: addr.clone(),
             amount,
-            fees: PegOutFees {
-                fee_rate: Feerate { sats_per_kvb: 1000 },
-                total_weight: 0,
-            },
+            fees: PegOutFees::new(1000, 883),
         };
 
         // agree on output
--- a/modules/fedimint-wallet-common/src/lib.rs
+++ b/modules/fedimint-wallet-common/src/lib.rs
@@ -152,6 +152,13 @@ pub struct PegOutFees {
 }
 
 impl PegOutFees {
+    pub fn new(sats_per_kvb: u64, total_weight: u64) -> Self {
+        PegOutFees {
+            fee_rate: Feerate { sats_per_kvb },
+            total_weight,
+        }
+    }
+
     pub fn amount(&self) -> Amount {
         self.fee_rate.calculate_fee(self.total_weight)
     }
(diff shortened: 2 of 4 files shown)


client/client-lib/src/wallet/mod.rs
```diff
@@ -340,10 +340,7 @@ mod tests {
         let output = PegOut {
             recipient: addr.clone(),
             amount,
-            fees: PegOutFees {
-                fee_rate: Feerate { sats_per_kvb: 1000 },
-                total_weight: 0,
-            },
+            fees: PegOutFees::new(1000, 883),
```
justinmoon, line 343: This scares me a little since it would be very easy to reverse the order of these arguments ...

modules/fedimint-wallet-common/src/lib.rs
```diff
@@ -302,16 +309,18 @@ pub enum WalletError {
     PegInProofError(#[from] PegInProofError),
     #[error("The peg-in was already claimed")]
     PegInAlreadyClaimed,
-    #[error("Peg-out fee rate {0:?} is set below consensus {1:?}")]
-    PegOutFeeRate(Feerate, Feerate),
+    #[error("Peg-out fee rate {0} is set below consensus {1}")]
+    PegOutFeeBelowConsensus(u64, u64),
```
elsirion, line 313: These fee rates don't have/display a unit anymore now

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1305,8 +1289,58 @@ struct StatelessWallet<'a> {
 }
 
 impl<'a> StatelessWallet<'a> {
+    /// Given a tx created from an `WalletOutput`, validate there will be no
+    /// issues submitting the transaction to the Bitcoin network
+    fn validate_tx(
+        &self,
+        tx: &UnsignedTransaction,
+        output: &WalletOutput,
```
elsirion, line 1297: It feels a bit wrong making the `StatelessWallet` aware of of the wallet module, originally it was intended as a place to put all the BTC logic that can be tested separately. This is still kinda the case since `WalletOutput` is only a data struct and I don't have a better solution to offer rn :/

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1305,8 +1289,58 @@ struct StatelessWallet<'a> {
...
+            return Err(WalletError::PegOutFeeBelowConsensus(
+                tx.fees.fee_rate.sats_per_kvb,
+                consensus_fee_rate.sats_per_kvb,
+            ));
+        }
+
+        // Validate added fees are above the min relay tx fee
+        let fees = match output {
+            WalletOutput::PegOut(pegout) => pegout.fees.clone(),
+            WalletOutput::Rbf(rbf) => rbf.fees.clone(),
```
elsirion, line 1326: `rbf.fees` is the *delta* of fees added and we enforce [rule 4 of BIP 125 here](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki#implementation-details). Maybe adding some reference to the BIP would be good here for others to know what's going on.

elsirion (dismissed):
Looks good, just two nits

modules/fedimint-wallet-server/src/lib.rs
jkitman: > originally it was intended as a place to put all the BTC logic that can be tested separately

This is still true, but perhaps we need an easier way of unit testing the module functions.  `FakeFed` is a little too involved for unit testing.

modules/fedimint-wallet-server/src/lib.rs
jkitman: Sure I will add a comment

elsirion (approved):

Issue #1829, opened 2023-03-07
Clean up DB expect boilerplate


jkitman:
Currently we have `213` instances of `.expect("DB error")` style calls in our code, this introduces a lot of useless boilerplate in our code.

In most cases we just want to get an `Option<(K, V)>` or a `Iter<(K, V)>` so perhaps we can add those as helper functions to our database trait?

@elsirion @m1sterc001guy What do you guys think about this?

m1sterc001guy:
> The thinking behind setting a flag during insert and checking it during commit is that this would allow us to support other DBs that don't support optimistic concurrency control. Essentially the public wrappers would swallow any error during insert time and always fail at commit time, which would expose consistent semantics to the higher layers, regardless of the underlying db.

Thinking about this a tad more, one scenario where this might be problematic is if there is some code that relies on reading from the existing transaction before it is committed. If `insert_entry` fails silently, this might result in the wrong thing happening. Perhaps it is better just to panic there as well.

elsirion:
I'm very much in favor of this and optimistic DBs don't really work with the current code base anyway. The question would be if bubbling up DB error everywhere just to support optimistic DB tx is worth it, I kinda doubt it.

PR #1834, opened 2023-03-07
refactor: remove  return values in public database transaction functions

m1sterc001guy:
Currently in our database code we propagate all errors via `Result`s all the way up to the application code. This was done partly because there was a discussion on where database errors should be handled https://github.com/fedimint/fedimint/discussions/1370

In practice, we were just calling `expect` on database errors, which makes the code much more verbose and difficult to read. This PRs refactors the error handling so that the public database transaction functions do not return `Result`, except for `commit_tx`. Right now database error handling is done after attempting to commit to the database. Database implementations that fail prematurely (e.g sqlite) can swallow their failures and fail during `commit_tx` as well, so that the failure semantics are similar across database implementations.

Fixes: https://github.com/fedimint/fedimint/issues/1829

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -520,17 +520,23 @@ impl<'isolated, T: Send + Encodable> ModuleDatabaseTransaction<'isolated, T> {
     }
 
     #[instrument(level = "debug", skip_all, fields(?key), ret)]
-    pub async fn get_value<K>(&mut self, key: &K) -> Result<Option<K::Value>>
+    pub async fn get_value<K>(&mut self, key: &K) -> Option<K::Value>
     where
         K: DatabaseKey + DatabaseRecord,
     {
         let key_bytes = key.to_bytes();
-        let value_bytes = match self.isolated_tx.raw_get_bytes(&key_bytes).await? {
-            Some(value) => value,
-            None => return Ok(None),
-        };
-
-        Ok(Some(decode_value::<K::Value>(&value_bytes, self.decoders)?))
+        let value_bytes = self
+            .isolated_tx
+            .raw_get_bytes(&key_bytes)
+            .await
+            .expect("Unrecoverable error when reading from database");
+        match value_bytes {
+            Some(value_bytes) => Some(
+                decode_value::<K::Value>(&value_bytes, self.decoders)
+                    .expect("Unrecoverable error when decoding the database value"),
+            ),
+            None => None,
+        }
     }
 
     #[instrument(level = "debug", skip_all, fields(key = ?key_prefix))]
@@ -538,10 +544,10 @@ impl<'isolated, T: Send + Encodable> ModuleDatabaseTransaction<'isolated, T> {
         &mut self,
         key_prefix: &KP,
     ) -> impl Stream<
-        Item = Result<(
+        Item = (
             KP::Record,
             <<KP as DatabaseLookup>::Record as DatabaseRecord>::Value,
-        )>,
+        ),
     > + '_
     where
         KP: DatabaseLookup,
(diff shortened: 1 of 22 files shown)


dpc:
>  Database implementations that fail prematurely (e.g sqlite) can swallow their failures and fail during commit_tx as well,

Does it work for reads? If we are to continue and pretend there was no error, we need to be able to return *something*. After a failed write, can one still issue new reads on the same opened transaction? Even then - we would probably get stale reads, which might be a problem in case the code has some assertions where it verifies reading own writes, even unintentionally.

fedimint-core/src/db/mod.rs
```diff
@@ -555,79 +561,89 @@ impl<'isolated, T: Send + Encodable> ModuleDatabaseTransaction<'isolated, T> {
...
                     target: LOG_DB,
                     "Database overwriting element when expecting insertion of new entry. Key: {:?}",
                     key
                 );
-                Ok(None)
+                None
             }
-            None => Ok(None),
+            None => None,
         }
```
justinmoon, line 613: This could just be `if let Some(_) = ...`

Also, if we're always returning `None` do we even need a return value?

fedimint-core/src/db/mod.rs
```diff
@@ -811,32 +827,43 @@ impl<'parent> DatabaseTransaction<'parent> {
         return self.tx.commit_tx().await;
     }
 
+    pub async fn expect_commit_tx(mut self) {
```
justinmoon, line 830: If I was brand new to this project, it would be a little weird to call `expect_commit_tx` to commit to the database. I would probably reach for `commit_tx` because it sounds more like what I'm looking to do.

I wonder if this function could just be `commit_tx` and the current `commit_tx` could be `commit_tx_result` or something? 

Also thinking that the one that returns a result could probably be private now? But seems like we use it for test assertions so maybe not ...

fedimint-core/src/db/mod.rs
```diff
@@ -850,77 +877,89 @@ impl<'parent> DatabaseTransaction<'parent> {
...
+            .expect("Unrecoverable error while inserting into the database")
         {
-            Some(old_val_bytes) => Ok(Some(decode_value(&old_val_bytes, &self.decoders)?)),
-            None => Ok(None),
+            Some(old_val_bytes) => Some(
+                decode_value(&old_val_bytes, &self.decoders)
+                    .expect("Unrecoverable error while decoding the database value"),
+            ),
+            None => None,
         }
```
justinmoon, line 905: Could we just call `.map` here instead of `match`?

justinmoon (commented):
Looks good. A couple little nits ...

fedimint-core/src/db/mod.rs
m1sterc001guy: Good catch, removed return value

dpc:
> In my opinion, its difficult to architect the code to be totally agnostic of the database's behavior, as it forces us to handle the cases where the db transaction fails optimistically or pessimistically.

Generally true.

Though in this the pessimistic locking is a more general case of optimistic locking. We could handle these errors everywhere and retry on them and that would work for both sqlite and rocksdb. If we were to use sqlite as a default we would have to do that anyway.

More work and complexity, but it could be done. Though if we were using rocksdb as default, these codepaths wouldn't we well tested, so :shrug: .

I guess with this change, we are accepting that rocksdb will be the only db we use. Both on server and client. These panics in certain scenarios will be a problem in rare, but inevitable cases if we are supporting 3rd party modules. So I think we need to be explicit about it - unless we enforce some event-based append-only model of client side operations or something like that, I think this change will make certain scenarios panic randomly on sqlite, making it not very workable for client.

It's probably for the best, though I wish it wasn't a C++ project. :/

elsirion (approved):
So much cleaner now :heart_eyes:

elsirion:
@dpc I hope we might be able to find a good rust native KV DB for at least the client eventually, but supporting wildly different transaction semantics sounds like too much effort for now. This PR at least gets us to *some* local maximum.

effort

m1sterc001guy:
Agree with everything mentioned above. I don't think the issues with sqlite are worth blocking this PR on but at some point in the future I'd like to spend some time to add the appropriate error handling changes to make it work properly. For now I think this refactor PR is an improvement.

I'll start a discussion on the different DB implementations that we've looked at and their tradeoffs, that will probably be useful in the future.

dpc (approved):

PR #1835, opened 2023-03-08
fix: shorter FM_TMP_DIR

dpc:
On some systems tmp dir is still too long, and shorter tmp dir template would help.


elsirion (approved):

justinmoon (approved):
Nice. Shaved 9 characters off my tmpdir.

PR #1839, opened 2023-03-09
chore: remove duplicate code from dummy-module

ngutech21:
Cleans up duplicated code in the dummy-modules:
* Remove ServerModuleGen and ServerModule from dummy-common
* Remove DummyVerificationCache from dummy-server and use the DummyVerificationCache from the dummy-common module instead

### DIFF
--- a/modules/fedimint-dummy-common/src/lib.rs
+++ b/modules/fedimint-dummy-common/src/lib.rs
@@ -1,45 +1,22 @@
-use std::collections::{BTreeMap, HashSet};
-use std::ffi::OsString;
 use std::fmt;
 
 use async_trait::async_trait;
 use bitcoin_hashes::sha256;
-use fedimint_core::config::{
-    ConfigGenParams, DkgResult, ModuleConfigResponse, ModuleGenParams, ServerModuleConfig,
-    TypedServerModuleConfig, TypedServerModuleConsensusConfig,
-};
+use fedimint_core::config::ModuleGenParams;
 use fedimint_core::core::{Decoder, ModuleInstanceId, ModuleKind};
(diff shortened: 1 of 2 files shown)


modules/fedimint-dummy-server/src/lib.rs
```diff
@@ -247,8 +247,3 @@ impl Dummy {
         Dummy { cfg }
     }
 }
-
-#[derive(Debug, Clone)]
-pub struct DummyVerificationCache;
```
dpc, line 252: Verification cache should be a part of a server. Client has no need for it. So I would delete the duplicate in common.

dpc (commented):
No idea how this happen. Thanks for taking care of it. Other than verification cache thing, LGTM

modules/fedimint-dummy-server/src/lib.rs
ngutech21: Yes makes sense

modules/fedimint-dummy-common/src/lib.rs
```diff
@@ -94,12 +91,11 @@ impl ModuleCommon for DummyModuleTypes {
     type ConsensusItem = DummyConsensusItem;
 }
 
-plugin_types_trait_impl!(
```
dpc, line 335: We need to delete `plugin_types_trait_impl` BTW. Only `plugin_types_trait_impl_common` should be needed now.

dpc (dismissed):

elsirion (dismissed):
That's the best kind of PR: removing instead of adding code ^^

elsirion:
Oh, needs rebase :(

dpc (approved):

elsirion (approved):

justinmoon:
Nice PR!

Issue #1841, opened 2023-03-09
Authenticated API endpoints


elsirion:
Some functionality should only be available to the guardian running `fedimintd`. The easiest way to achieve this is via some bearer token, i.e. a random password, that is sent per connection open *or* per request (tbd).

The question is how to generate this token and communicate it to the guardian's client. Below I include a possible flow that I came up with a while ago:

1. Guardian generates registration QR (in terminal/web/whatever)
1.1 fedimintd generates short term secret and saves it in the DB
1.2 fedimintd encodes the secret together with the other join info in the QR
2. Guardian App scans QR
2.1 The app already has its root secret generated in the background
2.2 It derives an authentication key from the root secret
2.3 The authentication key is sent to this guardian's fedimintd, the request is authenticated with the short term secret from the QR
2.4 fedimintd checks if the short term secret used in the request is in the DB and removes it
2.4.1 If successful the authentication key (=long term secret) is saved to the DB
2.4.2 A second request would fail since the short term secret was removed
3. Any authenticated API requests can now be authenticated with the authentication key (=long term secret)

Ideally this functionality would be handled by Fedimint such that the API endpoint implementation merely receives a boolean stating if the request was made by a guardian (alternative: API endpoints include a flag if they are public or private). That way modules could also implement private API endpoints.

jkitman:
Since the setup/admin UI will probably be browser-based, guardians may be logging in on a computer without camera support.

We already require the user to create a strong password when starting `fedimintd` for encrypting configs.  A salted hash of this password can be used to authenticate when logging into the setup/admin UI (since the admin endpoint might be exposed on a public IP).

The client UI could store this hash temporarily during each admin session, requiring users to enter their password periodically to login.  I think this works better from a UX perspective.  Eventually we could support two-factor auth if we want greater security for the admin UI.

elsirion:
Sounds good to me, but also means we should enforce strong passwords.

jkitman:
Yep you can also drop or ban IPs that attempt with too many bad passwords if brute-forcing via the API is a concern.

Issue #948, opened 2022-11-21
Shutdown at specific epoch when applying new consensus rules


justinmoon:
When making an upgrade, we need all peers to shutdown at the same epoch. Otherwise, different peers might end up in different states.

Then everyone switches out fedimintd and applies the new rules at the same time.

jkitman:
Peers can signal a desire to shutdown, perhaps passing in an optional consensus config hash, and when a threshold of peers are signaling in consensus, we automatically shutdown.

dpc:
Shutdown and then get restarted automatically by systemd? I don't think the shutdown changes anything here. If anything the fedimintd should just refuse to generate any new epochs and wait (for operator, or some other things unblocking it).

I guess we either invest into writing proper "migration" logic, where fedimintd can support both the previous and new consensus rules, and switch at a given (probably `ConsensusItem` driven) epoch automatically (more work for fedimint devs and module developers, smoother experience for the guardians) or just stop and have all the guardians coordinate simultanous restart.

I wonder if we could even have an easy hybrid. We could have two sets of consensus ports. Before upgrade the guardian would run the fedimintd like this:


```
fedimintd-0.14 --upgrade-new-bin fedimintd-0.15
```

After reaching upgrade epoch, fedimintd-0.14 would restart itself. On start it would detect that it is already at upgrade epoch and start operating in a read-only mode on an alternative set of ports, it would also start `fedimintd-0.15` that would take over normal operation.

Each fedimintd would know that if it is connecting to a peer that reports higher versions then their own, they should try alternative set of ports.

This way the old version can stick around indefinitely and allow any peers that were down etc. to finish syncing to the upgrade epoch when possible. While at no point we need a binary that can support both old and new protocol.

jkitman:
> After reaching upgrade epoch, fedimintd-0.14 would restart itself. On start it would detect that it is already at upgrade epoch and start operating in a read-only mode on an alternative set of ports, it would also start fedimintd-0.15 that would take over normal operation.

Yeah, I was imagining something similar like that would be good to minimize downtime, since the guardians might not be able to manually coordinate the upgrade within a short time window.

elsirion:
I don't think we should focus on uncoordinated upgrades too much yet. Being able to upgrade *at all* would be a huge improvement and is far lower complexity. Adding automatic switchover or even running two federations in parallel can build on top of that and would require far more work right now imo.

jkitman:
[...]

PR #1842, opened 2023-03-09
Support shutting-down at an epoch

jkitman:
This is the preliminary code for supporting shutdown at an epoch #948

What remains is the client code + tests to enable guardians to actually trigger the shutdown.  Since that requires developing an authenticated API it'll be done in a follow-up PR. See #1841

### DIFF
--- a/fedimint-core/src/epoch.rs
+++ b/fedimint-core/src/epoch.rs
@@ -12,13 +12,25 @@ use threshold_crypto::{PublicKey, PublicKeySet, Signature, SignatureShare};
 
 use crate::transaction::Transaction;
 
+/// All the items that may be produced during a consensus epoch
 #[derive(Debug, Clone, Eq, PartialEq, Hash, UnzipConsensus, Encodable, Decodable)]
 pub enum ConsensusItem {
+    /// Fed shutdown occurs once a threshold want to upgrade
+    ConsensusUpgrade(ConsensusUpgrade),
+    /// Threshold sign the configs for verification via the API
     ClientConfigSignatureShare(SerdeSignatureShare),
+    /// Threshold sign the epoch history for verification via the API
     EpochOutcomeSignatureShare(SerdeSignatureShare),
+    /// Fedimint tx that contains module inputs and outputs that are net
+    /// equal
     Transaction(Transaction),
+    /// Any data that modules require consensus on
     Module(ModuleConsensusItem),
 }
+
+/// May eventually contains consensus info about the upgrade
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Encodable, Decodable)]
+pub struct ConsensusUpgrade;
 
 pub type SerdeConsensusItem = SerdeModuleEncoding<ConsensusItem>;
 
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -9,6 +9,7 @@ use std::iter::FromIterator;
 use std::os::unix::prelude::OsStrExt;
 use std::sync::Mutex;
 
+use anyhow::format_err;
 use fedimint_core::config::{ConfigResponse, ServerModuleGenRegistry};
 use fedimint_core::core::ModuleInstanceId;
 use fedimint_core::db::{
@@ -24,7 +25,7 @@ use fedimint_core::module::{ModuleError, TransactionItemAmount};
 use fedimint_core::outcome::TransactionStatus;
 use fedimint_core::server::{DynServerModule, DynVerificationCache};
 use fedimint_core::task::TaskGroup;
-use fedimint_core::{Amount, OutPoint, PeerId, TransactionId};
+use fedimint_core::{Amount, NumPeers, OutPoint, PeerId, TransactionId};
 use fedimint_logging::{LOG_CONSENSUS, LOG_CORE};
 use futures::future::select_all;
 use futures::StreamExt;
@@ -39,9 +40,9 @@ use crate::config::ServerConfig;
 use crate::consensus::interconnect::FedimintInterconnect;
 use crate::consensus::TransactionSubmissionError::TransactionReplayError;
 use crate::db::{
-    get_global_database_migrations, AcceptedTransactionKey, ClientConfigSignatureKey, DropPeerKey,
-    DropPeerKeyPrefix, EpochHistoryKey, LastEpochKey, RejectedTransactionKey,
-    GLOBAL_DATABASE_VERSION,
+    get_global_database_migrations, AcceptedTransactionKey, ClientConfigSignatureKey,
+    ConsensusUpgradeKey, DropPeerKey, DropPeerKeyPrefix, EpochHistoryKey, LastEpochKey,
+    RejectedTransactionKey, GLOBAL_DATABASE_VERSION,
 };
 use crate::transaction::{Transaction, TransactionError};
 
@@ -331,6 +332,7 @@ impl FedimintConsensus {
                             epoch_outcome_signature_share: _epoch_outcome_signature_share_cis,
                             client_config_signature_share: _client_config_signature_share_cis,
                             transaction: transaction_cis,
+                            consensus_upgrade: consensus_upgrade_cis,
                             module: module_cis,
                         } = consensus_outcome
                             .contributions
@@ -339,6 +341,7 @@ impl FedimintConsensus {
                             .unzip_consensus_item();
 
                         self.process_module_consensus_items(dbtx, &module_cis).await;
+                        self.process_upgrade_items(dbtx, &consensus_upgrade_cis).await;
 
                         let rejected_txs = self
                             .process_transactions(dbtx, epoch, &transaction_cis)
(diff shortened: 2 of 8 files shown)


fedimint-core/src/epoch.rs
```diff
@@ -12,14 +12,29 @@ use threshold_crypto::{PublicKey, PublicKeySet, Signature, SignatureShare};
 
 use crate::transaction::Transaction;
 
+/// All the items that may be produced during a consensus epoch
 #[derive(Debug, Clone, Eq, PartialEq, Hash, UnzipConsensus, Encodable, Decodable)]
 pub enum ConsensusItem {
+    /// Fed shutdown occurs once a threshold of peers signal (for consensus
+    /// upgrades)
+    ShutdownSignal(ShutdownReason),
```
dpc, line 20: I was imagining that this will be more of a `ConsensusUpgrade {  versions: GlobalConsensusVersion, height: EpochHeight }` kind of thing, that will cause the `fedimintd` to quit when out of data.

I don't think we need/want a "general purpose shutdown mechanism" in the consensus. IMO it should only contain consensus-related items (which version upgrade is very much a part of).

jkitman: I could rename it, there probably aren't other consensus shutdown cases.  Where would the `GlobalConsensusVersion` come from?

dpc: > I could rename it, there probably aren't other consensus shutdown cases. 

IMO, the way to think about it is in consensus terms. Eventually we might not even require a shutdown and fedimint might just smoothly switch to new consensus versions, etc. Shutdown is just an implementation detail of how we handle consensus upgrades currently, no?

>  Where would the GlobalConsensusVersion come from?

Hmmm... Maybe it should not be a version, but `ConsensusChange { consensus_hash: }`? With a whole consensus config hash? That would allow consensus changes without consensus upgrades, which is more general.

I'm not deep into discussion here. Eventually I guess guardian will just go to the UI, click some "propose consensus rule change", get some UI to add modules, change config values, add/remove members, etc. This will also include an epoch (date estimate). The proposal will be send to other fed members and trigger a notification to their guardians and they will see the proposal and either accept and reject? When quorum is reached the upgrade is locked and all other fedimints will follow (even if their guardians don't act / reject it).

dpc: But currently we haven no such UI, so I'm not sure how we are planning to trigger it. Some `fedimint-cli` where one passes an epoch + new config?

jkitman: Yes, eventually we'd want it to be built into the UI.

For now, users will probably need to run a script that will backup and update any consensus config changes.  Similar to DB changes, you'd need to be able to roll-back and some migrations can be automated (also some will require running DKG).

Perhaps we should run these migrations while the fed is still running (generating new configs without using them) so that we can all confirm we have the same consensus hash?  But that would also require the new binary is available to run them while the old binary is still running.  For now it's easier just to agree to shutdown and coordinate the changes out-of-band.

dpc: >  For now it's easier just to agree to shutdown and coordinate the changes out-of-band.

I see. OK. I guess by making it a consensus item, we can make even peers that are not around stop.

In that case I would suggest introducing a `shutdown-id`.

After performing upgrade procedure, guardian would have to restart `fedimintd` with a `--consensus-continue <shutdown-id>` or something like that to make sure operators don't get it restarted by accident. Unknown `--consensus-continue <shutdown-id>` should be ignored (with a warning) and after consensus was restarted, `--consensus-continue` should no longer be required.

The "Reason" in the shutdown seems kind of meh. If everything is out of band, I don't see why this wouldn't be.

dpc: > After performing upgrade procedure, guardian would have to restart `fedimintd` with a `--consensus-continue <shutdown-id>` or something like that to make sure operators don't get it restarted by accident. Unknown `--consensus-continue <shutdown-id>` should be ignored (with a warning) and after consensus was restarted, `--consensus-continue` should no longer be required.

Otherwise, I don't see currently anything that would prevent `fedimintd` from just being restarted by `systemd` or other scripts.

fedimint-core/src/epoch.rs
```diff
@@ -12,14 +12,29 @@ use threshold_crypto::{PublicKey, PublicKeySet, Signature, SignatureShare};
...
+    /// Any data that modules require consensus on
     Module(ModuleConsensusItem),
 }
 
+/// Contains consensus info about the shutdown reason
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Encodable, Decodable)]
+pub enum ShutdownReason {
+    Upgrade,
+}
+
```
elsirion, line 37: Agree with DPC for now an enum with 1 variant is kinda unnecessary. In terms of upgradeability it also doesn't add much without `#[non_exhaustive]` and afaik `ShutdownReason` shouldn't be part of the public API anyway? I'd say leave it out ofr now and add it if we ever have another reason.

fedimint-core/src/epoch.rs
elsirion: I second that idea, automatic systemd restarts are a concern.

fedimint-server/src/consensus/mod.rs
```diff
@@ -547,6 +549,31 @@ impl FedimintConsensus {
...
+    async fn process_shutdown_signals(
+        &self,
+        dbtx: &mut DatabaseTransaction<'_>,
+        shutdown_signals: &[(PeerId, ShutdownReason)],
+    ) {
+        if !shutdown_signals.is_empty() {
+            let maybe_exists = dbtx.get_value(&ShutdownSignalKey).await;
+            let mut peers = maybe_exists.unwrap_or_default();
+            peers.extend(shutdown_signals.iter().map(|(peer, _)| peer));
+            dbtx.insert_entry(&ShutdownSignalKey, &peers).await;
```
elsirion, line 562: Personally not a big fan of editing DB entries, but saves a few DB key/prefix structs here. With DB transactions it's not a problem anymore.

fedimint-server/src/lib.rs
```diff
@@ -352,6 +367,10 @@ impl FedimintServer {
             }
         }
 
+        if self.consensus.is_at_shutdown_threshold().await {
+            self.task_group.shutdown().await;
```
elsirion, line 371: Did you test if this works? Some tasks were not adhering to the shutdown signal yet …

dpc: BTW. My consensus processing logic here is limited so I'm not sure if there's anything to do anymore about it, but:

We should gate the epoch processing on lack of blocking shutdown item explicitly and not relay on task shutdown system to enforce it because it might be replaced / refactored / buggy.

The consensus shutdown should be somewhere in the most common code before processing the epoch height scheduled for shutdown (in `self.consensus.process_consensus_outcome`?). *Before*, so it can be never accidentally skipped. If it's after then one can process the epoch, write state to the db, get restarted and then never run the check altogether, potentially processing more epochs than other peers, etc. Or some corner case (rejoining consensus, etc.) can run through bunch of epochs without doing the check etc.

jkitman: If it's before then we will ignore any other items in that epoch?  That might cause issues for clients who think their transactions were accepted.

fedimint-core/src/epoch.rs
jkitman: Trying to understand the process we are agreeing upon:

1. Guardians manually trigger an upgrade via an authenticated API, submitted as a consensus item
2. Once a threshold is reached a deterministic shutdown id is persisted in the DB and `fedimintd` shuts down for everyone at the same epoch
3. Guardians manually run the upgrade (config/binary changes) then pass in the shutdown id as an arg to start `fedimintd` again

fedimint-server/src/lib.rs
jkitman: We could process the items in the shutdown epoch and the shutdown id will be persisted in the same DB transaction.

As a fail-safe we can check for the non-existence of the shutdown id upon start-up or epoch processing (would need to be manually removed by passing it in as a start-up parameter)

fedimint-core/src/epoch.rs
elsirion: Exactly, that's how I understand it too, this will prevent accidental restarts

fedimint-server/src/lib.rs
dpc: > If it's before then we will ignore any other items in that epoch? That might cause issues for clients who think their transactions were accepted.

We will exit.

So an epoch that had consensus item that made the shutdown reach a threshold is the last one that was processed, and it will update the db (atomically: in the same db transactions as all other side-effects of that epoch). And then the next epoch won't be able to process due to the check right before it.

jkitman:
I think I've taken your comments into account.  The process we are aiming for is now:
- Guardians manually trigger an upgrade via an authenticated API, submitted as a consensus item
- Once a threshold is reached, `fedimintd` shuts down for everyone at the same epoch
- Guardians manually run the upgrade (config/binary changes) then pass in the upgrade epoch id as an arg to start `fedimintd` again (to avoid accidental restarts)

Without the authenticated API unfortunately I cannot test things yet, but we could land this then I'll take a look at implementing that next.

dpc (commented):
LGTM, one nit.

dpc (approved):

elsirion (approved):

PR #1844, opened 2023-03-09
Simpler Core-Lightning plugin

justinmoon:
- Doesn't use gRPC at all
- `ILnRpcClient` implementation for CLN does direct RPC with CLN instead of using gRPC to proxy the requests
- The plugin forwards HTLC to `gatewayd` using the same authenticated HTTP POST mechanism that gateway-cli uses
- The gateway's `run()` loop does SCID filtering to figure out whether an HTLC is relevant to us, and if it is, which federation "actor" to process it with. gRPC implementation did it inside the plugin.
- If we changed LND to use this pattern, then `Gateway::new()` would call `ILnRpcClient.subscribe_htlcs(sender)` which would just send the HTLC to the `Gateway.run()` loop like the HTTP POST method does for CLN. So we'd have 1 HTLC subscription for the whole gateway, instead of 1 per actor (which doesn't seem to work with the gRPC library we've been using).

I think we should still keep the gRPC mechanism because it's a nice way interface for people to write their own custom gateway implementations against. Just don't use it for Core Lightning because a simpler approach works just fine.


elsirion:
> The plugin forwards HTLC to `gatewayd` using the same authenticated HTTP POST mechanism that gateway-cli uses

That's not a nice architecture, I mean it works for local deployments, but now both sides have to kinda know/reach each other …

justinmoon:
> > The plugin forwards HTLC to `gatewayd` using the same authenticated HTTP POST mechanism that gateway-cli uses
> 
> That's not a nice architecture, I mean it works for local deployments, but now both sides have to kinda know/reach each other …

Extension needs to know hostname and password for gatewayd webserver. Gatewayd doesn't need to know anything about extension. It's just the reverse of the current requirement that the address of the extension's gRPC server. No password because this connection is un-authenticated currently.

PR #1850, opened 2023-03-10
fix: Remove outdated entries from .gitignore

justinmoon:
I think the integration tests used to store stuff there. But that's all in a tmp directory now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,5 @@
 /target
 /result
-cfg
-ln1
-ln2
-it
 .idea
 .tmpenv
 .direnv


dpc (approved):

.gitignore
```diff
@@ -1,9 +1,5 @@
 /target
 /result
-cfg
```
elsirion, line 3: I actually still use that one :stuck_out_tongue_closed_eyes: (adding it to `.git/info/exclude` as I should have from the beginning)

.gitignore
```diff
@@ -1,9 +1,5 @@
...
-ln1
-ln2
-it
 .idea
```
elsirion, line 3: We might also think about removing setup-specific entries and document `.git/info/exclude` somewhere

elsirion (approved):

PR #1851, opened 2023-03-10
Add config files for bitcoind and electrs

justinmoon:
... for tests and tmuxinator.

Esplora doesn't support config files.

Added config files for LND and Core Lighnting in #1852

I think the config files make the shell scripts a bit more readable ...

### DIFF
--- a/misc/test/electrs.toml
+++ b/misc/test/electrs.toml
@@ -1,0 +1,5 @@
+auth = "bitcoin:bitcoin"
+daemon_rpc_addr = "127.0.0.1:18443"
+daemon_p2p_addr = "127.0.0.1:18444"
+electrum_rpc_addr = "127.0.0.1:50001"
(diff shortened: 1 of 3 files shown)


elsirion (dismissed):
Great idea!

justinmoon:
Fixed conflicts with https://github.com/fedimint/fedimint/pull/1852 since that also added the bitcoin.conf.

dpc (approved):
LGTM, though I would use `-conf=path/to/bitcoind.conf` to avoid mutability (copying files).

justinmoon:
> LGTM, though I would use `-conf=path/to/bitcoind.conf` to avoid mutability (copying files).

Made an issue and assigned myself https://github.com/fedimint/fedimint/issues/1857

elsirion (approved):

PR #1719, opened 2023-02-20
LND support

justinmoon:
Builds on #1718 and #1717

So far this implements outgoing payments and tests them in cil-test.sh and has fully working tmuxinator environment, too.

TODO:

- [ ] Route hints
- [x] Subscribe HTLCs
- [ ] Intercept HTLCs

Notes:

- Split out tmuxinator windows more: 1 cli window, 1 fedimintd window, 1 lightning window with cln. + lnd + gatewayd for both, and 1 bitcoind window
- Got rid of "ln1" and "ln2" and instead calling them roughly "cln" and "lnd".
- I'm finding inding the `ILnRpcClient` very clunky because it's really intended to speak to cln-extension. Nothing like this exists for LND. I think we should consider refactoring this trait to not use cln-extension GRPC types as arguments and return types.

PR #1852, opened 2023-03-10
Run LND in tests and tmuxinator

justinmoon:
First chunk broken off https://github.com/fedimint/fedimint/pull/1719.

Just run LND as non-gateway node in tests and tmuxinator.

In follow-up PR I'll turn this node into a gateway as well.

Also contains commits from https://github.com/fedimint/fedimint/pull/1851 which could be merged separately.

### DIFF
--- a/misc/test/bitcoin.conf
+++ b/misc/test/bitcoin.conf
@@ -1,0 +1,9 @@
+regtest=1
+fallbackfee=0.0004
+txindex=1
+server=1
+rpcuser=bitcoin
+rpcpassword=bitcoin
+zmqpubrawblock=tcp://127.0.0.1:38332
+# commonly used port 28333 collides with dkg ports
+zmqpubrawtx=tcp://127.0.0.1:38333
--- a/scripts/cli-test.sh
+++ b/scripts/cli-test.sh
@@ -54,26 +54,39 @@ mine_blocks 10
 RECEIVED=$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR)
 [[ "$RECEIVED" = "0.00000500" ]]
 
-# outgoing lightning
-INVOICE="$($FM_LN2 invoice 100000 test test 1m | jq -e -r '.bolt11')"
-await_cln_block_processing
+# lightning tests
+await_lightning_node_block_processing
+
+# outgoing lightning payment via gateway
+ADD_INVOICE="$($FM_LNCLI addinvoice --amt_msat 100000)"
+INVOICE="$(echo $ADD_INVOICE| jq -e -r '.payment_request')"
+PAYMENT_HASH="$(echo $ADD_INVOICE| jq -e -r '.r_hash')"
 $FM_MINT_CLIENT ln-pay $INVOICE
 # Check that ln-gateway has received the ecash notes from the user payment
 # 100,000 sats + 100 sats without processing fee
 LN_GATEWAY_BALANCE="$($FM_GATEWAY_CLI balance $FED_ID | jq -e -r '.balance_msat')"
 [[ "$LN_GATEWAY_BALANCE" = "100100000" ]]
-INVOICE_RESULT="$($FM_LN2 waitinvoice test)"
-INVOICE_STATUS="$(echo $INVOICE_RESULT | jq -e -r '.status')"
+INVOICE_STATUS="$($FM_LNCLI lookupinvoice $PAYMENT_HASH | jq -e -r '.state')"
+[[ "$INVOICE_STATUS" = "SETTLED" ]]
+
+# incoming lightning payment via gateway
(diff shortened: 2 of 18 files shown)


misc/test/bitcoin.conf
```diff
@@ -0,0 +1,9 @@
+regtest=1
+fallbackfee=0.0004
+txindex=1
+server=1
+rpcuser=bitcoin
+rpcpassword=bitcoin
+zmqpubrawblock=tcp://127.0.0.1:38332
+# commonly used port 28333 collides with dkg ports
+zmqpubrawtx=tcp://127.0.0.1:38333
```
elsirion, line 9: Will this conflict with #1851? Might want to be clever about merge order.

scripts/lib.sh
```diff
@@ -8,17 +8,19 @@ function mine_blocks() {
...
+    LN_ADDR="$($FM_LIGHTNING_CLI newaddr | jq -e -r '.bech32')"
     $FM_BTC_CLIENT sendtoaddress $LN_ADDR 1
     mine_blocks 10
-    FM_LN2_PUB_KEY="$($FM_LN2 getinfo | jq -e -r '.id')"
-    export FM_LN2_PUB_KEY
-    FM_LN1_PUB_KEY="$($FM_LN1 getinfo | jq -e -r '.id')"
-    export FM_LN1_PUB_KEY
-    $FM_LN1 connect $FM_LN2_PUB_KEY@127.0.0.1:9001
-    until $FM_LN1 -k fundchannel id=$FM_LN2_PUB_KEY amount=0.1btc push_msat=5000000000; do sleep $POLL_INTERVAL; done
+    $FM_LNCLI getinfo
```
elsirion, line 17: This `getinfo` without processing its output seems unnecessary/left from debugging something?

scripts/cli-test.sh
```diff
@@ -54,26 +54,39 @@ mine_blocks 10
...
+PAYMENT_HASH="$(echo $ADD_INVOICE| jq -e -r '.r_hash')"
 $FM_MINT_CLIENT ln-pay $INVOICE
 # Check that ln-gateway has received the ecash notes from the user payment
 # 100,000 sats + 100 sats without processing fee
 LN_GATEWAY_BALANCE="$($FM_GATEWAY_CLI balance $FED_ID | jq -e -r '.balance_msat')"
 [[ "$LN_GATEWAY_BALANCE" = "100100000" ]]
-INVOICE_RESULT="$($FM_LN2 waitinvoice test)"
-INVOICE_STATUS="$(echo $INVOICE_RESULT | jq -e -r '.status')"
-[[ "$INVOICE_STATUS" = "paid" ]]
+INVOICE_STATUS="$($FM_LNCLI lookupinvoice $PAYMENT_HASH | jq -e -r '.state')"
```
elsirion, line 69: Using lookup instead of waiting for the invoice works because we wait for the payment to succeed previously (i.e. the preimage reached the payer). This means the LN node *should* recognize the invoice as paid, since it gave away the preimage.

scripts/tmux-user-shell.sh
```diff
@@ -31,7 +32,8 @@ echo
 echo "This shell provides the following aliases:"
 echo ""
 echo "  fedimint-cli   - cli client to interact with the federation"
-echo "  ln1, ln2       - cli clients for the two lightning nodes (1 is gateway)"
+echo "  lightning-cli  - cli client for Core Lightning"
+echo "  lncli          - cli client for LND"
```
elsirion, line 36: docs: which one is the gateway?

elsirion (approved):
Some minor nits, but looks like a really good step into the direction of gateway support for multiple LN implementations.

okjodom (approved):

scripts/cli-test.sh
justinmoon: Correct. Once `fedimint-cli ln-pay <invoice>` completes, that means that the LND node received the payment, so it should have marked that invoice as paid.

misc/test/bitcoin.conf
justinmoon: I think we can merge whichever is approved first, and I can fix conflicts if there are any.

scripts/lib.sh
justinmoon: Good catch. I'm going to merge this (since it's approved) and then make a follow-up PR to remove this.

justinmoon: https://github.com/fedimint/fedimint/pull/1856

scripts/tmux-user-shell.sh
elsirion: How will users choose which gateway to use in that future? We might want to add the ability to name gateways.

justinmoon: To get the tests passing I defined these https://github.com/fedimint/fedimint/pull/1719/files#diff-5838719667e50a69f27dfddac131cad8ed4e11bee597a64a1b48454efdfad40bR167-R187

But yea naming them might be interesting, too.

PR #1855, opened 2023-03-10
Client NG III

elsirion:
* [x] define `PrimaryClientModule` interface
* [x] Build transaction funding/change functionality
* [x] Also expose it via the global context so state machines can generate refund txns
* [x] Remove circular dependency when initializing executor+client
* [x] Add test for tx submission SM
* [x] Implement `Client` builder


fedimint-client/src/module/mod.rs
```diff
@@ -55,9 +68,174 @@ where
...
+    /// Creates an output of **exactly** `amount` that will pay into the
+    /// holdings managed by the module.
+    ///
+    /// It returns:
+    /// * The output of **exactly** `amount`.
+    /// * A closure that generates states belonging to the output. This closure
+    ///   takes the transaction id of the transaction in which the output was
+    ///   used and the output index as input since these cannot be known at time
+    ///   of calling `create_change_output` and have to be injected later.
+    async fn create_change_output(
```
dpc, line 146: `create_exact_output`? Sure, we'll probably use it for creating change, but on a higher level we only care about module being able to create an exact arbitrary value output, and you even made that world bold already - why not just put it in a name then?

fedimint-client/src/module/mod.rs
```diff
@@ -55,9 +68,174 @@ where
...
+    ///   larger, the caller has to handle this case and possibly generate
+    ///   change using `create_change_output`.
+    /// * A closure that generates states belonging to the input. This closure
+    ///   takes the transaction id of the transaction in which the input was
+    ///   used and the input index as input since these cannot be known at time
+    ///   of calling `create_funding_input` and have to be injected later.
+    ///
+    /// The function returns an error if the client's funds are not sufficient
+    /// to create the requested input.
+    async fn create_funding_input(
```
dpc, line 127: Similarly here. `create_sufficient_input` maybe?

justinmoon (approved):
LGTM. Don't have a super strong understanding here yet, but I think we'll continue reviewing this code as new modules are built.

PR #1861, opened 2023-03-10
fix: just test-ci-all behavior on ctrl+c

dpc:


scripts/test-ci-all.sh
```diff
@@ -12,6 +11,12 @@ if [ -n "${HOME:-}" ] && [ -d "$HOME" ]; then
   touch "$HOME/.parallel/will-cite"
 fi
 
+
+# Avoid re-building workspace in parallel in all test derivations
+>&2 echo "### Making sure workspace is built..."
+nix build -L .#debug.workspaceBuild 2>&1 | ts -s
```
elsirion, line 17: What was the rationale to move this block?

elsirion (approved):

scripts/test-ci-all.sh
dpc: Actually moving the block above (making gnu parallel silent) all the way up, just to get it out of the way, as it has nothing to do with the core logic.

justinmoon (approved):

PR #1865, opened 2023-03-11
fix: don't manually pkill processes after tmuxinator exits

justinmoon:
`kill_fedimint_processes` should already clean up everything


mxz42 (approved):
also run tmuxinator; no `bitcoind|lightnind` exists after exiting  tmux with `:kill-session`

okjodom (approved):

PR #1868, opened 2023-03-11
Bash functions to start all our daemons

justinmoon:
I was noticing some inconsistencies between how daemons are spawned and disposed (https://github.com/fedimint/fedimint/pull/1865) in tests vs tmuxinator and thought I'd make it more consistent, readable, and easier to reason about.

* If a daemon depends on another, I try to wait for it in the start function.
* `start_bitcoind` creates RPC wallet and mines 101 blocks.
* `start_gatewayd` connects to federation.
* Also, updated `gw_connect_fed` to retry if it fails. That should help address [this failure](https://discord.com/channels/990354215060795454/990354215878688860/1083640931556806737).

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -11,6 +11,7 @@ pre_window:
   - alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"
   - alias dbtool="\$FM_DB_TOOL"
   - alias restart="./scripts/restart-tmux.sh"
+  - source scripts/lib.sh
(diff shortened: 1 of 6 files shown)


dpc (approved):

okjodom (approved):

elsirion (approved):
We accumulated a scary amount of bash :sweat_smile:

justinmoon:
I think all of this should be rewritten in rust, but cleaning it up conceptually in bash should help that task ...

PR #1869, opened 2023-03-11
Improve gateway instantiation

justinmoon:
- Don't panic
- Don't retry fetching route hints if empty vec is returned. I don't think this solves anything, and it confuses control flow because the constructor can just block forever if you have a node with no channels. IMO that's just weird behavior, but maybe I'm missing something.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -167,18 +167,6 @@ impl Gateway {
         Ok(())
     }
 
-    async fn select_actor(&self, federation_id: FederationId) -> Result<Arc<GatewayActor>> {
-        self.actors
-            .lock()
-            .await
-            .get(&federation_id.to_string())
-            .cloned()
-            .ok_or(GatewayError::Other(anyhow::anyhow!(
-                "No federation with id {}",
-                federation_id.to_string()
-            )))
-    }
-
     pub async fn load_actor(
         &self,
         client: Arc<GatewayClient>,
@@ -199,6 +187,18 @@ impl Gateway {
             actor.clone(),
         );
         Ok(actor)
+    }
+
+    async fn select_actor(&self, federation_id: FederationId) -> Result<Arc<GatewayActor>> {
+        self.actors
+            .lock()
+            .await
+            .get(&federation_id.to_string())
+            .cloned()
+            .ok_or(GatewayError::Other(anyhow::anyhow!(
+                "No federation with id {}",
+                federation_id.to_string()
+            )))
     }
 
     async fn handle_connect_federation(
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -108,12 +108,12 @@ impl Gateway {
             module_gens: module_gens.clone(),
         };
 
-        gw.load_federation_actors(decoders, module_gens).await?;
+        gw.load_actors(decoders, module_gens).await?;
 
         Ok(gw)
     }
 
-    async fn load_federation_actors(
+    async fn load_actors(
         &self,
         decoders: ModuleDecoderRegistry,
         module_gens: ClientModuleGenRegistry,
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -96,15 +96,38 @@ impl Gateway {
         // Create message channels for the webserver
         let (sender, receiver) = mpsc::channel::<GatewayRequest>(100);
 
-        // Source route hints form the LN node
+        let gw = Self {
+            lnrpc,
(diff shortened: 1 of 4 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -93,28 +90,13 @@ impl Gateway {
...
-            info!(
-                ?num_retries,
-                "LN node returned no route hints, trying again in {}s",
-                ROUTE_HINT_RETRY_SLEEP.as_secs()
-            );
-            num_retries += 1;
-            tokio::time::sleep(ROUTE_HINT_RETRY_SLEEP).await;
-        };
+        // Fetch route hints form the LN node
+        let route_hints: Vec<RouteHint> = lnrpc
```
okjodom: lgtm. I'll let @elsirion comment on the value of the retry. Seems like this only applied to some aspect of CLN plugin

okjodom (approved):
lgtm

gateway/ln-gateway/src/lib.rs
```diff
@@ -87,33 +86,10 @@ impl Gateway {
...
-            let route_hints: Vec<RouteHint> = lnrpc
-                .routehints()
-                .await
-                .expect("Could not fetch route hints")
-                .try_into()
-                .expect("Could not parse route hints");
-
-            if !route_hints.is_empty() || num_retries == ROUTE_HINT_RETRIES {
-                break route_hints;
-            }
```
elsirion, line 106: It doesn't block *forever*, currently only 20s. The reason for this is that only online channels are returned (to avoid returning permanently offline/closed ones) and after a node comes online it may take a moment to get its channels up. Iirc this was a fix to an actual problem that happened infrequently and only in production (in tests envs channels come oneline super quickly).

The real fix would be re-fetching this list on every re-registration. Then we wouldn't mind as much there being no route hints for the first 5 minutes or so after restart.

gateway/ln-gateway/src/lib.rs
```diff
@@ -126,18 +102,24 @@ impl Gateway {
...
-    ) {
+    ) -> Result<()> {
+        // Fetch route hints form the LN node
+        let route_hints: Vec<RouteHint> = self
+            .lnrpc
+            .routehints()
+            .await
+            .map_err(|_| GatewayError::FailedToFetchRouteHints)?
+            .try_into()
+            .map_err(|_| GatewayError::FailedToFetchRouteHints)?;
```
elsirion, line 122: Why would different GW actors talking to the same LN node, being loaded a few ms after each other request route hints multiple times instead of only once?

elsirion (changes requested):
* I support getting rid of the expects
* The reason for moving route hint fetching to actor initialization is unclear to me
* Finding a better way of dealing with temp offline channels during startup would be nice, just removing the current logic could lead to quite some pain during the next deployment.

gateway/ln-gateway/src/lib.rs
justinmoon: This makes sense. I didn't understand the offline channel part. I figured the only change would be if a new channel showed up, which seemed unlikely in such a short timespan.

gateway/ln-gateway/src/lib.rs
justinmoon: They don't. This function is called once in `Gateway::new()` -- not per-actor.

The reason I moved if from the constructor to this method was because this method was the only part of `Gateway::new()` that cared about route hints. Seems more self-contained to call it inside this function.

justinmoon: I think it might be better if the gateway had the route hints attribute, and the actors were simpler clients without this state.

gateway/ln-gateway/src/lib.rs
elsirion: The plugin starts alongside CLN and right after restart all channels are offline because CLN didn't have time to get them all up. They will connect ion a few seconds typically if they aren't Tor-connected channels.

gateway/ln-gateway/src/lib.rs
elsirion: Then we'll need some synchronization primitive again that preferably won't lead to deadlocks in some obscure way :grimacing: But it would make things easier, like updating the available route hints every now and then.

elsirion (dismissed):

gateway/ln-gateway/src/lib.rs
elsirion: > They don't. This function is called once in `Gateway::new()` -- not per-actor.

Also sorry for the confusion, should have read more carefully.

justinmoon:
marking draft until i fix ci

justinmoon:
@elsirion @okjodom fixed ci

elsirion (approved):

Issue #665, opened 2022-09-26
Generic `await` bash function


justinmoon:
See https://github.com/fedimint/fedimint/pull/661#discussion_r979843619

By my count our [lib.sh](https://github.com/fedimint/fedimint/blob/master/scripts/lib.sh) has 5 separate `until ... do sleep $POLL_INTERVAL done` blocks. We could write a single `await <condition>` function which does the sleeping and call this directly in these 5 cases.

blakejakopovic:
I've taken a stab at this and have a forked [commit](https://github.com/blakejakopovic/fedimint/commit/c714189ca87142626db5a2e9eae1916c5e8c47dc) ([patch](https://github.com/blakejakopovic/fedimint/commit/c714189ca87142626db5a2e9eae1916c5e8c47dc.patch)) pending feedback before I submit a PR here. Was this is intended approach? Do we know why a mix of single and double square brackets were used for the conditions? - happy to standardise on either.

I've tested it a few ways in isolation, however I don't have full fedimint dev environment setup - so it would benefit from further testing. The following scripts (possibly incomplete) appear to be impacted by the following `lib.sh` refactor - I'd be grateful if anyone can assist in testing while I get a dev environment setup.

```
scripts/cli-test.sh
scripts/clientd-tests.sh
scripts/setup-tests.sh
scripts/tmux-user-shell.sh
```

dpc:
We're rewriting in Rust, we can have a real `await` now.

PR #1870, opened 2023-03-11
run_dkg and start_federation bash functions

justinmoon:
More attempts to simplify how we start services in tests & tmuxinator. The reason I'm doing this is that it was kind of difficult to reason about all the dependencies in these scripts while trying to modify them to [run LND](https://github.com/fedimint/fedimint/pull/1852).

I removed the `scripts/start-fed.sh`. I think @jkitman used to call that one directly sometimes so maybe we should keep it? Main reason I chose to delete it is because it now assumes `FM_FED_SIZE` is set so couldn't be called directly before this PR.

### DIFF
--- a/docs/dev-env.md
+++ b/docs/dev-env.md
@@ -167,7 +167,7 @@ write the files to the host file-system (e.g. config generation). `-p` is used t
 container's port 17240.
 
 To generate federation configs and run the federation please see the `run_dkg` and `start_federation` helper functions in [lib.sh](https://github.com/fedimint/fedimint/blob/master/scripts/lib.sh), as well as the `--help` output of
-`distributedgen` and `fedimintd`:
+`distributedgen` and `fedimintd`. If you'd like to run function like `run_dkg` and `start_federation` from inside nix shell, run `source scripts/build.sh` and then `source scripts/lib.sh` and they will be available.
 
 Note that you can also start a "fake" 1-of-1 "federation" that will allow you to test most aspects of Fedimint without
 having to run e.g. 4 instances.


scripts/build.sh
```diff
@@ -5,7 +5,6 @@ echo "Run with 'source ./scripts/build.sh [fed_size] [dir]"
 
 # allow for overriding arguments
 export FM_FED_SIZE=${1:-4}
-BASE_PORT=$((8173 + 10000))
```
justinmoon, line 8: This variable was only used for DKG so moved it to be internal to that function.

scripts/build.sh
```diff
@@ -20,6 +19,7 @@ else
 fi
 export FM_TMP_DIR
 export FM_TEST_FAST_WEAK_CRYPTO="1"
+export FM_POLL_INTERVAL=1
```
justinmoon, line 22: Define this in one place instead of a few places. For tmuxinator it was set to `0.5` and elsewhere it was `1`. Is setting to `1` ok?

scripts/lib.sh
```diff
@@ -146,25 +146,59 @@ function show_verbose_output()
...
+  DKG_PIDS=""
+  for ((ID=0; ID<FM_FED_SIZE; ID++));
+  do
+    export FM_PASSWORD="pass$ID"
+    fed_port=$(echo "$BASE_PORT + $ID * 10" | bc -l)
+    api_port=$(echo "$BASE_PORT + $ID * 10 + 1" | bc -l)
+    $FM_BIN_DIR/distributedgen run  --bind-p2p 127.0.0.1:$fed_port --bind-api 127.0.0.1:$api_port --out-dir $FM_CFG_DIR/server-$ID --certs $CERTS &
+    DKG_PIDS="$DKG_PIDS $!"
+  done
+  wait $DKG_PIDS
```
justinmoon, line 180: Now we can't just `wait` for all we've processes spawned, so instead we keep track of just the DKG PIDs and only wait for them.

justinmoon:
Seems like VSCode's website is down (https://code.visualstudio.com/) causing docs CI failure ...

jkitman (approved):
LGTM

justinmoon:
converted to draft because it should be failing CI ...

justinmoon:
Everything is passing for me locally. Right now I don't this CI is working properly, though https://discord.com/channels/990354215060795454/1084235802088849448

scripts/lib.sh
```diff
@@ -16,15 +16,15 @@ function open_channel() {
...
+    until [[ $($FM_LIGHTNING_CLI listpeers | jq -e -r ".peers[] | select(.id == \"$LND_PUBKEY\") | .channels[0].state") = "CHANNELD_NORMAL" ]]; do sleep $FM_POLL_INTERVAL; done
 }
 
 function await_bitcoin_rpc() {
     until $FM_BTC_CLIENT getblockchaininfo 1>/dev/null 2>/dev/null ; do
         >&2 echo "Bitcoind rpc not ready yet. Waiting ..."
-        sleep "$POLL_INTERVAL"
+        sleep "$FM_POLL_INTERVAL"
     done
 }
```
elsirion, line 29: I wish we had #665, then we could be fancy here without too much overhead/copy paste and just add one period every second and not print a new line.

scripts/lib.sh
```diff
@@ -172,29 +175,38 @@ function start_lightningd() {
...
   gw_connect_fed
+  echo "started gatewayd"
 }
 
 function start_electrs() {
+  echo "starting electrs"
   await_bitcoin_rpc
   electrs --conf-dir "$FM_ELECTRS_DIR" --db-dir "$FM_ELECTRS_DIR" --daemon-dir "$FM_BTC_DIR" &
   echo $! >> $FM_PID_FILE
+  echo "started electrs"
```
elsirion, line 236: for future optimization: should we make this a [done] at he end of the line? The first echo would just not print a newline.

docs/dev-env.md
```diff
@@ -166,10 +166,8 @@ $ docker run -it -v $PWD/demo:/var/fedimint -p 17240:17240 fedimintd:iqviraxy2cz
 write the files to the host file-system (e.g. config generation). `-p` is used to bind the host's port 17240 as the
 container's port 17240.
 
-To generate federation config and run the federation please see the integration tests and the `--help` output of
+To generate federation configs and run the federation please see the `run_dkg` and `start_federation` helper functions in [lib.sh](https://github.com/fedimint/fedimint/blob/master/scripts/lib.sh), as well as the `--help` output of
```
elsirion, line 169: Should we mention `source`ing here? By default we don't source lib in the dev env afaik (only in tmuxinator iirc).

elsirion (dismissed):

elsirion:
> LGTM, but maybe squash the commits.

It was very pleasant to follow along while reviewing, what would be the benefit of squashing?

scripts/lib.sh
justinmoon: I tried implementing this, but don't know bash well enough, and that frustration got me to try https://github.com/fedimint/fedimint/pull/1878.

scripts/lib.sh
justinmoon: https://github.com/fedimint/fedimint/issues/1882

docs/dev-env.md
justinmoon: Added a sentence about sourcing `build.sh` and `lib.sh` here. This doesn't work for me, but if I hardcode the `SRC_DIR` parameter it does  (https://github.com/fedimint/fedimint/issues/1883).

justinmoon:
everything here except log update got merged in https://github.com/fedimint/fedimint/pull/1878

elsirion (approved):

dpc (approved):

PR #1871, opened 2023-03-11
run-ui.sh uses common start_federation shell function

justinmoon:
Builds on #1870. Last 4 commits are new.

The test script for running a federation without doing DKG now starts the federation with the same mechanics as tests and tmuxinator do.

Along the way made a `setup_fedimintd_env` function which sets all environment variables that `distributedgen` or `fedimintd` depend on, as well as creates the data dir.

After this PR I'm finished with this shell refactoring and will return to LND. Just wanted to fix some of the annoyances I encountered while still fresh in my mind. Not touching [start-containers.sh](https://github.com/fedimint/fedimint/blob/master/scripts/start-containers.sh) because it's not tested or anything ...


dpc (approved):

scripts/lib.sh
```diff
@@ -184,6 +184,25 @@ function run_dkg() {
...
+  export FM_P2P_URL=fedimint://127.0.0.1:$P2P_PORT
+  export FM_BIND_API=127.0.0.1:$API_PORT
+  export FM_API_URL=ws://127.0.0.1:$API_PORT
+  export FM_LISTEN_UI=127.0.0.1:$UI_PORT
+  export FM_PASSWORD="pass$ID"
+  export FM_FEDIMINTD_DATA_DIR="$FM_CFG_DIR/server-$ID"
+
+  # ensure datadir exists ... pipe to `true` because run_dkg() and setup_federation()
+  # both call this so the directory might already be here ...
+  mkdir $FM_FEDIMINTD_DATA_DIR || true
```
elsirion, line 197: This will potentially hide some errors (e.g. no space left on device, lack of permissions), but seems good enough.

elsirion (approved):

scripts/lib.sh
justinmoon: Good point, but I think that's unlikely. This will always run `build.sh` which makes directories inside the tmpdir ...

PR #1877, opened 2023-03-12
Remove flaky sleeps from tmuxinator config

justinmoon:
Based on #1870

Now the bash functions wait for anything they depend on, so sleeps are no longer necessary.


scripts/lib.sh
```diff
@@ -82,6 +82,10 @@ function kill_fedimint_processes {
   rm -f $FM_PID_FILE
 }
 
+function await_gateway_cln_extension() {
+  while ! echo exit | nc localhost 8177; do sleep $FM_POLL_INTERVAL; done
```
elsirion, line 86: How does this work? :open_mouth: What does sending exit do, just cooperatively closing the connection? Is this gRPC?

elsirion (approved):

scripts/lib.sh
justinmoon: I just copied it from here https://unix.stackexchange.com/a/5279. `nc` will read from stdin and block. echoing `exit` and piping it to `nc` will immediately exit. This is a great example of where it would be more readable if we just wrote this in rust!

dpc (approved):

PR #1879, opened 2023-03-12
fix: remove unused `Transaction` struct

elsirion:
Was probably an oversight from when I was finishing server modularization.

### DIFF
--- a/fedimint-core/src/core.rs
+++ b/fedimint-core/src/core.rs
@@ -19,7 +19,6 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     dyn_newtype_define_with_instance_id, dyn_newtype_impl_dyn_clone_passhthrough_with_instance_id,
-    ModuleDecoderRegistry,
 };
 
 pub mod client;
@@ -359,7 +358,8 @@ impl Debug for Decoder {
     }
 }
 
-/// Something that can be an [`DynInput`] in a [`Transaction`]
+/// Something that can be an [`DynInput`] in a
+/// [`Transaction`](fedimint_core::transaction::Transaction)
 ///
 /// General purpose code should use [`DynInput`] instead
 pub trait IInput: Debug + Display + DynEncodable {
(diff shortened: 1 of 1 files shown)


justinmoon (approved):

justinmoon:
doc test failure

dpc (approved):

mxz42 (dismissed):

mxz42:
doc test fails only because of ```[`Transaction`]``` is not linkable in `core.rs`. 

I think - while on it - you could also update the link in `docs/lifecycle.md` as the link is pinned to a hash.

```sh
pr/1879 %) | rgi '`transaction`'
docs/database.md
81:| OutgoingPaymentClaim    |     `0x24`    | contract id (sha256)               | `Transaction`                |

docs/lifecycle.md
3:- A [`Transaction`](https://github.com/fedimint/fedimint/blob/a1f57e3c6ff860a9c4a998bf88ebad73ebdb67c9/fedimint-core/src/transaction.rs#L12) Consists of inputs and outputs

docs/architecture.md
32:The API task in `net:api:run_server` allows clients to submit a `Transaction` and retrieve the `TransactionStatus`.
48:Any module can contribute inputs and outputs in the same `Transaction`.
63:A submitted `Transaction` often requires multiple epochs to become spendable, usually because they require signatures from a quorum of federation members.

fedimint-core/src/core.rs
361:/// Something that can be an [`DynInput`] in a [`Transaction`]
380:    /// An owned, immutable input to a [`Transaction`]
391:/// Something that can be an [`DynOutput`] in a [`Transaction`]
402:    /// An owned, immutable output of a [`Transaction`]
432:    /// An owned, immutable output of a [`Transaction`] before it was finalized
458:    /// An owned, immutable output of a [`Transaction`] before it was finalized

modules/fedimint-mint-common/src/lib.rs
83:/// [`Note`]s as an `Input`s of `Transaction` with new [`Note`]s being created

```

dpc:
@elsirion let's land (doc link fix needed). Already bit me. :D

elsirion:
> doc test fails only because of `` [`Transaction`] `` is not linkable in `core.rs`.

Fixed with the latest push I think.

> I think - while on it - you could also update the link in `docs/lifecycle.md` as the link is pinned to a hash.

I'll leave it as is, the document assumes a pre-`Dyn*`-type code base, would be very confusing to link to the current one. Eventually we'll need to remove or restructure that document. But it still works well to get the concept across (and then you tell people it's much more complicated in reality :laughing: ).

> @elsirion let's land (doc link fix needed). Already bit me. :D

Given how long nobody noticed I assumed it wasn't as high prion xD

justinmoon (approved):

mxz42 (approved):

PR #1886, opened 2023-03-13
Fix min value for Argon2

jkitman:
Looks like our CI is broken (always passing) so I didn't catch this issue that will cause the fed to fail to start.

### DIFF
--- a/crypto/aead/src/lib.rs
+++ b/crypto/aead/src/lib.rs
@@ -3,7 +3,7 @@ use std::path::PathBuf;
 
 use anyhow::{bail, format_err, Result};
 use argon2::password_hash::{Salt, SaltString};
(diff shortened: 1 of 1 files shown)


mxz42 (approved):
ah damn, missed that, too, as I didn't start tmuxinator.
Startup and interaction e.g. `fedimint-cli info` works.

I think unrelated to this, tmuxinator hangs for me at 
```sh
Bitcoind rpc not ready yet. Waiting ...
lightning-cli: Moving into '/tmp/nix-shell.KVqh0J/fm-fbOh/cln/regtest': No such file or directory
```

For anyone else as well?

elsirion (approved):

justinmoon:
@mxz42 setup fixtures heavily in flux right now. let me know if you keep seeing that ...

PR #1887, opened 2023-03-13
test: mint and wallet database migration tests

m1sterc001guy:
Adding tests for the mint and wallet modules that test database migrations. The intention of this PR is to catch when a change to a database key/value has occurred, but the corresponding migration has not been added. These tests create some dummy data and insert it into the database to be consumed later. The `test_migrations` test then reads these databases, applies the database migrations, and then validates that it can read all of the data.

Lightning module PR: https://github.com/fedimint/fedimint/pull/1803


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1626,3 +1626,314 @@ mod tests {
...
+            input: vec![TxIn {
+                previous_output: utxo.0,
+                script_sig: Default::default(),
+                sequence: Sequence::ENABLE_RBF_NO_LOCKTIME,
+                witness: bitcoin::Witness::new(),
+            }],
+            output,
+        };
+
+        let inputs = vec![Input {
```
m1sterc001guy, line 1711: Hmm do you guys have thoughts on how we should handle persisted structs from external crates? I think there is a small chance that if we bump the version of the external crate, the structure has changed, but now it becomes difficult to write a db migration since the external crate likely doens't expose the old structure as well.

dpc: Part of the reason we need migration tests. IMO - developers will have to import old versions of their dependencies (in parallel to new ones, possibly) for as long as they want to maintain migration window.

elsirion: The good thing is we use BTC consensus encoding here, which is unlikely to change. But agree with @dpc in the general case.

elsirion (approved):
To be clear, we can only test if a migration specified in the code did not apply, right? Eventually we might want to checkout a previous version, generate the DB, then run tests with the current version. I guess that's what @dpc means with migration tests.

m1sterc001guy:
> To be clear, we can only test if a migration specified in the code did not apply, right? Eventually we might want to checkout a previous version, generate the DB, then run tests with the current version. I guess that's what @dpc means with migration tests.

Correct. Today if no migrations are applied, but the data can still be read, the tests will still pass. Once we start making releases or adding migrations, we should definitely be taking db snapshots and run those against the current version. That process is manual right now, perhaps we can automate in the future.

While adding a new migration, developers should ideally add their own migration test that does validate the migration applied correctly, similar to what I did in the dummy module.

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -1170,3 +1170,229 @@ pub struct VerifiedNotes {
...
+            data: BYTE_32.to_vec(),
+        };
+        dbtx.insert_new_entry(&backup_key, &ecash_backup).await;
+
+        dbtx.commit_tx().await;
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn prepare_migration_snapshots() {
+        prepare_snapshot(
```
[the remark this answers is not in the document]
m1sterc001guy: Done

dpc (approved):

elsirion (approved):

PR #1888, opened 2023-03-13
DO NOT MERGE: trigger false CI pass

justinmoon:
Over the weekend this CI run passed but the latency tests clearly had an error and should have caused it to fail https://github.com/fedimint/fedimint/actions/runs/4393951590/jobs/7694690151#step:10:1729

This PR adds `exit 1` to `latency-test.sh` and CI still passes, but it shouldn't ...

### DIFF
--- a/scripts/latency-test.sh
+++ b/scripts/latency-test.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 # Runs a test to determine the latency of certain user actions
 
+exit 1 # this should cause CI to fail
+
 set -eu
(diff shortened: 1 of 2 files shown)


justinmoon:
rebased on https://github.com/fedimint/fedimint/pull/1890. Hopefully this triggers a failure. Then we know https://github.com/fedimint/fedimint/pull/1890 works.

justinmoon:
Closing now that https://github.com/fedimint/fedimint/pull/1890 is merged which seems to fix the problem ...

PR #1890, opened 2023-03-13
chore: fix test-ci-all ignoring failures

dpc:
set -o pipefail gets reset when parallel invokes a bash function.


justinmoon:
I think this fixes CI. I rebased https://github.com/fedimint/fedimint/pull/1888 on top of it, and it fails as expected. Before rebasing, it passed even though `exit 1` was at the top of `latency-tests.sh`.

justinmoon (approved):

m1sterc001guy (approved):

okjodom (approved):

PR #1895, opened 2023-03-13
fix: configure proto3 optionals in grpc proto compiler

okjodom:


elsirion:
This seems to make CI fail, any idea why?

okjodom:
> This seems to make CI fail, any idea why?

Really surprised by the way CI fails. And I see no trace of this change being the cause of failures in `Build/Tests`

justinmoon (approved):

dpc (approved):

PR #1899, opened 2023-03-13
Spawn all daemons via RUST

justinmoon:
Spawn electrs, esplora, fedimintd, gatewayd via rust. Also, run a whole federation or dkg.

After this, whenever you're bored you can just choose some bash function in lib.sh and re-write it in rust. I think all the basic infrastructure is there.

### DIFF
--- a/fedimint-testing/src/bin/fixtures.rs
+++ b/fedimint-testing/src/bin/fixtures.rs
@@ -15,6 +15,8 @@ enum Cmd {
     Bitcoind,
     Lightningd,
     Lnd,
+    Electrs,
+    Esplora,
     Daemons,
 }
 
@@ -55,7 +57,7 @@ async fn run_bitcoind() -> anyhow::Result<()> {
     let mut bitcoind = Command::new("bitcoind")
         .arg(format!("-datadir={btc_dir}"))
         .spawn()
-        .expect("failed to spawn");
+        .expect("failed to spawn bitcoind");
     info!("bitcoind started");
 
     // create client
@@ -91,7 +93,7 @@ async fn run_lightningd() -> anyhow::Result<()> {
         .arg(format!("--lightning-dir={cln_dir}"))
         .arg(format!("--plugin={bin_dir}/gateway-cln-extension"))
         .spawn()
-        .expect("failed to spawn");
+        .expect("failed to spawn lightningd");
     info!("lightningd started");
 
     lightningd.wait().await?;
@@ -109,10 +111,53 @@ async fn run_lnd() -> anyhow::Result<()> {
     let mut lnd = Command::new("lnd")
         .arg(format!("--lnddir={lnd_dir}"))
         .spawn()
-        .expect("failed to spawn");
+        .expect("failed to spawn lnd");
     info!("lnd started");
 
     lnd.wait().await?;
+
+    Ok(())
+}
+
+async fn run_electrs() -> anyhow::Result<()> {
+    // wait for bitcoin RPC to be ready ...
+    await_bitcoin_rpc("electrs").await?;
+
+    let electrs_dir = env::var("FM_ELECTRS_DIR").unwrap();
+
+    // spawn electrs
+    let mut electrs = Command::new("electrs")
+        .arg(format!("--conf-dir={electrs_dir}"))
+        .arg(format!("--db-dir={electrs_dir}"))
+        .spawn()
+        .expect("failed to spawn electrs");
+    info!("electrs started");
+
+    electrs.wait().await?;
+
+    Ok(())
+}
+async fn run_esplora() -> anyhow::Result<()> {
+    // wait for bitcoin RPC to be ready ...
+    await_bitcoin_rpc("esplora").await?;
+
+    let daemon_dir = env::var("FM_BTC_DIR").unwrap();
+    let test_dir = env::var("FM_TEST_DIR").unwrap();
+
+    // spawn esplora
+    let mut esplora = Command::new("esplora")
+        .arg(format!("--daemon-dir={daemon_dir}"))
+        .arg(format!("--db-dir={test_dir}/esplora"))
+        .arg("--cookie=bitcoin:bitcoin")
+        .arg("--network=regtest")
+        .arg("--daemon-rpc-addr=127.0.0.1:18443")
+        .arg("--http-addr=127.0.0.1:50002")
+        .arg("--monitoring-addr=127.0.0.1:50003")
+        .spawn()
+        .expect("failed to spawn esplora");
+    info!("esplora started");
+
+    esplora.wait().await?;
 
     Ok(())
 }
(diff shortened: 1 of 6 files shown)


dpc (approved):

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -1,51 +1,53 @@
...
-    bitcoin_rpc: bool,
+#[derive(Subcommand)]
+enum Cmd {
+    // daemons
+    Bitcoind,
+    Lightningd,
+    Lnd,
+    Electrs,
+    Esplora,
+    Daemons,
```
m1sterc001guy, line 36: Maybe rename this to AllDaemons? I needed to look at the code to determine which daemons this was referring to

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -60,14 +62,76 @@ fn bitcoin_rpc() -> anyhow::Result<Arc<BitcoinClient>> {
     Ok(client)
 }
 
-async fn run_bitcoind(state: MutableState) -> anyhow::Result<()> {
+async fn fedimint_client() -> anyhow::Result<UserClient> {
+    let workdir: PathBuf = env::var("FM_CFG_DIR").unwrap().parse()?;
```
m1sterc001guy, line 66: Can unwrap be changed to ?

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -60,14 +62,76 @@ fn bitcoin_rpc() -> anyhow::Result<Arc<BitcoinClient>> {
...
+    let cfg_path = workdir.join("client.json");
+    let db_path = workdir.join("client.db");
+    let cfg: UserClientConfig = load_from_file(&cfg_path).expect("Failed to parse config");
```
m1sterc001guy, line 69: Can expect be changed to ?

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -60,14 +62,76 @@ fn bitcoin_rpc() -> anyhow::Result<Arc<BitcoinClient>> {
...
+        sleep(Duration::from_secs(1)).await;
+        info!("{waiter_name} waiting for bitcoin rpc ...");
+    }
+
+    Ok(())
+}
+
+async fn await_fedimint_block_sync() -> anyhow::Result<()> {
+    // FIXME: make sure we've had at least 101 blocks ... probably best to call
+    // await_bitcoin_rpc and do it there? but then the function has weird name ...
```
m1sterc001guy, line 109: Could pass a closure or something to await_bitcoin_rpc that would encapsulate that logic. Then not all callers of `await_bitcoin_rpc` would need to wait 101 blocks if its not necessary. Won't block on this though

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -60,14 +62,76 @@ fn bitcoin_rpc() -> anyhow::Result<Arc<BitcoinClient>> {
...
+
+async fn run_bitcoind() -> anyhow::Result<()> {
     let btc_dir = env::var("FM_BTC_DIR").unwrap();
 
     // spawn bitcoind
     let mut bitcoind = Command::new("bitcoind")
         .arg(format!("-datadir={btc_dir}"))
         .spawn()
-        .expect("failed to spawn");
+        .expect("failed to spawn bitcoind");
```
m1sterc001guy, line 134: Change to ?

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -106,65 +167,365 @@ async fn run_lightningd(state: MutableState) -> anyhow::Result<()> {
         .arg(format!("--lightning-dir={cln_dir}"))
         .arg(format!("--plugin={bin_dir}/gateway-cln-extension"))
         .spawn()
-        .expect("failed to spawn");
+        .expect("failed to spawn lightningd");
```
m1sterc001guy, line 170: Can we change to ?

fedimint-testing/src/bin/fixtures.rs
m1sterc001guy, line 1: It might be helpful to add a comment at the top that describes the necessary environment variables that need to be set. In the bash scripts it is somewhat obvious where those are set, but with this it's necessary to look through the code

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -106,65 +167,365 @@ async fn run_lightningd(state: MutableState) -> anyhow::Result<()> {
...
+
+    Ok(())
+}
+
+async fn run_fedimintd(id: usize) -> anyhow::Result<()> {
+    // wait for bitcoin RPC to be ready ...
+    await_bitcoin_rpc(&format!("fedimint-{id}")).await?;
+
+    let bin_dir = env::var("FM_BIN_DIR").unwrap();
+    let cfg_dir = env::var("FM_CFG_DIR").unwrap();
```
m1sterc001guy, line 245: Can we change to ?

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -106,65 +167,365 @@ async fn run_lightningd(state: MutableState) -> anyhow::Result<()> {
...
+
+async fn run_federation(start_id: usize, stop_id: usize) -> anyhow::Result<()> {
+    let mut task_group = TaskGroup::new();
+    task_group.install_kill_handler();
+
+    for id in start_id..stop_id {
+        task_group
+            .spawn(format!("fedimintd-{id}"), move |_| async move {
+                info!("starting fedimintd-{}", id);
+                run_fedimintd(id).await.expect("fedimintd failed");
```
m1sterc001guy, line 290: Can we change expect to ?

fedimint-testing/src/bin/fixtures.rs
justinmoon: Since `run_bitcoind` mines 101 blocks as soon as it starts, I think we should just wait for 101 blocks to be mined in all cases. That won't take long, and it's a nice invariant. For example, then we can send from the default rpc wallet (`pegin.sh` needs this) and LND won't blow up when it tries to start (if bitcoin RPC commands return that they are still loading the block index, it does blow up).

justinmoon (commented):
Thanks for review @m1sterc001guy. This I've addressed everything.

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -106,65 +167,365 @@ async fn run_lightningd(state: MutableState) -> anyhow::Result<()> {
...
+            })
+            .await;
+    }
+
+    task_group.join_all(None).await?;
+
+    Ok(())
+}
+
+fn fedimintd_env(id: usize) -> HashMap<String, String> {
```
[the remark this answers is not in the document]
justinmoon: done

fedimint-testing/src/bin/fixtures.rs
justinmoon: Added a comment at the top, and in `main()` it exits if `FM_TMP_DIR` isn't set -- meaming that `build.sh` wasn't `source`'d.

fedimint-testing/src/bin/fixtures.rs
```diff
@@ -84,87 +158,410 @@ async fn run_bitcoind(state: MutableState) -> anyhow::Result<()> {
...
+    Ok(())
+}
+
+async fn run_dkg(servers: usize) -> anyhow::Result<()> {
+    let root_task_group = TaskGroup::new();
+    root_task_group.install_kill_handler();
+    let mut task_group = root_task_group.make_subgroup().await;
+
+    // let base_port = 8173 + 10000;
+    // let mut certs = vec![];
```
m1sterc001guy, line 438: remove?

dpc (approved):

PR #1904, opened 2023-03-14
Auth api

### LINKED ISSUE #1841: Authenticated API endpoints

jkitman:
This adds optional auth to the APIs available through the `has_auth` parameter in the `api_endpoint!` handler.

The auth is a hashed password stored in the encrypted `ServerConfig` (better performance + security over DB).

Follow up PR will be to add an authenticated API for #948 along with tests.

Fixes #1841

### DIFF
--- a/crypto/aead/src/lib.rs
+++ b/crypto/aead/src/lib.rs
@@ -85,22 +85,15 @@ pub fn encrypted_read(key: &LessSafeKey, file: PathBuf) -> Result<Vec<u8>> {
 ///
 /// * `password` - Strong user-created password
 /// * `salt` - Nonce >8 bytes to discourage rainbow attacks
-fn get_encryption_key(password: &str, salt: &[u8]) -> Result<LessSafeKey> {
+pub fn get_encryption_key(password: &str, salt: &str) -> Result<LessSafeKey> {
     let mut key = [0u8; ring::digest::SHA256_OUTPUT_LEN];
 
     argon2()
-        .hash_password_into(password.as_bytes(), salt, &mut key)
+        .hash_password_into(password.as_bytes(), salt.as_bytes(), &mut key)
         .map_err(|e| format_err!("could not hash password").context(e))?;
     let key = UnboundKey::new(&ring::aead::CHACHA20_POLY1305, &key)
         .map_err(|_| anyhow::Error::msg("Unable to create key"))?;
     Ok(LessSafeKey::new(key))
-}
-
-/// Helper for `get_encryption_key` reading the salt from a path
-pub fn get_encryption_key_with_path(password: &str, salt_path: PathBuf) -> Result<LessSafeKey> {
-    let salt_str = fs::read_to_string(salt_path)?;
-    let salt = hex::decode(salt_str)?;
-    get_encryption_key(password, &salt)
 }
 
 /// Memory-hard Argon2 key stretching for password-based authentication
@@ -138,7 +131,7 @@ mod tests {
     #[test]
     fn encrypts_and_decrypts() {
         let password = "test123";
-        let salt = "salt1235".as_bytes();
+        let salt = "salt1235";
         let message = "hello world";
 
         let key = get_encryption_key(password, salt).unwrap();
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -42,7 +42,7 @@ use tracing::{debug, error, instrument, trace};
 use url::Url;
 
 use crate::epoch::{SerdeEpochHistory, SignedEpochOutcome};
-use crate::module::ApiRequest;
+use crate::module::ApiRequestErased;
 use crate::outcome::TransactionStatus;
 use crate::query::{
     CurrentConsensus, EventuallyConsistent, QueryStep, QueryStrategy, UnionResponses,
@@ -50,8 +50,6 @@ use crate::query::{
 };
 use crate::transaction::{SerdeTransaction, Transaction};
 use crate::CoreError;
-
-type JsonValue = serde_json::Value;
 
 pub type MemberResult<T> = result::Result<T, MemberError>;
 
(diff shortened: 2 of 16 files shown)


fedimint-core/src/api.rs
```diff
@@ -154,21 +155,22 @@ pub trait IFederationApi: Debug + MaybeSend + MaybeSync {
...
+    vec![params_raw]
 }
 
 /// Build a `Vec<json::Value>` that [`IFederationApi::request_raw`] expects when
 /// one argument are passed to the API call
 ///
 /// Notably the caling convention of fedimintd api is a bit weird ATM, so by
 /// using this function you'll make it easier to change it in the future.
-pub fn erased_single_param<Params>(param: &Params) -> Vec<JsonValue>
+pub fn erased_single_param<Params>(params: &Params) -> Vec<JsonValue>
```
m1sterc001guy, line 168: A little weird to call this `params` when the function title is `erased_single_param` :)

jkitman: Yeah I intend to do a follow-up refactor to clean this up, perhaps returning `ApiRequestErased`.

fedimint-server/src/config/mod.rs
```diff
@@ -748,6 +758,7 @@ impl ServerConfigParams {
                 let bind_api = parse_host_port(peer_params[peer].clone().api_url)?;
 
                 let params: ServerConfigParams = Self::gen_params(
+                    ApiAuth("dummy_password".to_string()),
```
m1sterc001guy, line 761: Does this need to be passed from the command line?

fedimintd/src/fedimintd.rs
```diff
@@ -30,17 +30,22 @@ pub struct ServerOpts {
...
     pub data_dir: PathBuf,
     /// Password to encrypt sensitive config files
+    // TODO: should probably never send password to the server directly, rather send the hash via
+    // the API
     #[arg(env = "FM_PASSWORD")]
     pub password: String,
     /// Port to run admin UI on
     #[arg(long = "listen-ui", env = "FM_LISTEN_UI")]
     pub listen_ui: Option<SocketAddr>,
+    /// After an upgrade the epoch must be passed in
```
m1sterc001guy, line 40: Unrelated question to the PR but since I'm thinking of it: if you were a guardian that was not apart of coordinating the shutdown (maybe you were offline), but you wanted to re-join after the upgrade occurred, how would you figure out this number?

m1sterc001guy:
> The auth is a hashed password stored in the encrypted ServerConfig (better performance + security over DB).

Just curious, why is it better perf + security over the DB? RocksDb should keep data in memory after the first read. And if its only storing the hash wouldn't the security be about the same?

I'm cool with it being in `ServerConfig` though. I suppose it is easier to manually change the values in `ServerConfig` by decrypting and re-encrypting the config file (with the db this would be more difficult).

jkitman:
> only storing the hash wouldn't the security be about the same

Guess the performance would be the same.  An attacker with filesystem access without memory access would be able to call authenticated commands on the API using the hashed password, but if it's encrypted they wouldn't.  So a security edge-case.

Also felt it was slightly better code-wise to add it to configs than introduce a new DB key for it, although that's debatable.

fedimint-server/src/config/mod.rs
jkitman: Just for the CI tests so it's okay to have it as a dummy string for now.

fedimintd/src/fedimintd.rs
jkitman: Ideally users wouldn't be calling this from the command-line but via an API call from an admin UI.

Still a lot of work to do on that front.

m1sterc001guy (approved):

fedimint-server/src/config/mod.rs
justinmoon: It's a little weird to have a hard-coded password like this nonetheless ... wouldn't want anyone to unintentionally use this on mainnet someday after we've forgotten about it ...

jkitman: True, this function should be refactored into the test fixtures since that is the only place it's used.

fedimint-core/src/module/mod.rs
```diff
@@ -58,6 +59,44 @@ impl TransactionItemAmount {
...
+    }
+
+    pub fn to_json(&self) -> JsonValue {
+        serde_json::to_value(self).expect("parameter serialization error - this should not happen")
+    }
+}
+
+/// Authentication uses the hashed user password in PHC format
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ApiAuth(pub String);
```
justinmoon, line 98: nit: if it's a hash, maybe we could use a hash type?

fedimint-core/src/module/mod.rs
```diff
@@ -194,15 +216,17 @@ impl ApiEndpoint<()> {
...
         ) -> Result<E::Response, ApiError>
         where
             E: TypedApiEndpoint,
             E::Param: Debug,
             E::Response: Debug,
         {
-            tracing::trace!(target: "fedimint_server::request", ?param, "recieved request");
-            let result = E::handle(state, dbtx, param).await;
+            tracing::trace!(target: "fedimint_server::request", ?request, "recieved request");
+            let has_auth = request.auth == Some(api_auth);
```
elsirion, line 244: So we are handing in `api_auth` from the config just for the comparison here? I'd really hope there is a better way to go about it in the future, but can be done in a separate PR since it would likely involve some more refactoring and isn't high prio imo.

fedimint-core/src/api.rs
```diff
@@ -149,48 +147,6 @@ pub trait IFederationApi: Debug + MaybeSend + MaybeSync {
...
-pub fn erased_multi_param<Params>(params: &Params) -> Vec<JsonValue>
-where
-    Params: Serialize,
-{
-    let params_raw = serde_json::to_value(ApiRequest { auth: None, params })
-        .expect("parameter serialization error - this should not happen");
-
-    vec![params_raw]
-}
-
```
elsirion: Nice!

elsirion (approved):

justinmoon (approved):
LGTM

crypto/aead/src/lib.rs
```diff
@@ -85,24 +85,17 @@ pub fn encrypted_read(key: &LessSafeKey, file: PathBuf) -> Result<Vec<u8>> {
 ///
 /// * `password` - Strong user-created password
 /// * `salt` - Nonce >8 bytes to discourage rainbow attacks
-fn get_encryption_key(password: &str, salt: &[u8]) -> Result<LessSafeKey> {
+pub fn get_encryption_key(password: &str, salt: &str) -> Result<LessSafeKey> {
```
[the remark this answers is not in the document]
jkitman: Wanted to be consistent since the Argon API requires it is a B64 encoded `&str` nonce.

fedimint-core/src/module/mod.rs
jkitman: It's not just a hash, but a string in the PHC format which is how the Argon API returns it

https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md

fedimint-core/src/module/mod.rs
jkitman: I think there might be a way to do it but would require moving the `ServerConfig` into `fedimint-core`.

PR #1907, opened 2023-03-14
Lnd Support

okjodom:
Cherry pick of #1719

### DIFF
--- a/gateway/ln-gateway/src/bin/gatewayd.rs
+++ b/gateway/ln-gateway/src/bin/gatewayd.rs
@@ -82,8 +82,8 @@ async fn main() -> Result<(), anyhow::Error> {
     } = GatewayOpts::parse();
 
     info!(
-        "Starting gateway with these configs \n data directory: {:?},\n listen: {},\n api address: {},\n lnrpc address: {:?} ",
-        data_dir, listen, api_addr, lnrpc_addr
+        "Starting gateway with these base configs \n data directory: {:?},\n listen: {},\n api address: {} ",
+        data_dir, listen, api_addr
     );
 
     // Create federation client builder
@@ -96,6 +96,11 @@ async fn main() -> Result<(), anyhow::Error> {
     let lnrpc = if let Some(lnrpc_addr) = lnrpc_addr {
         if let (Some(lnd_tls_cert), Some(lnd_macaroon)) = (lnd_tls_cert, lnd_macaroon) {
             // Detected LND rpc client configurations
+            info!(
+                "Gateway configured to connect to LND LnRpcClient at \n address: {:?},\n tls cert path: {:?},\n macaroon path: {} ",
+                lnrpc_addr, lnd_tls_cert, lnd_macaroon
+            );
+
             GatewayLndClient::new(
                 lnrpc_addr.into(),
                 lnd_tls_cert,
@@ -106,6 +111,11 @@ async fn main() -> Result<(), anyhow::Error> {
             .into()
         } else {
             // Detected generic network rpc client configurations
+            info!(
+                "Gateway configured to connect to generic remote LnRpcClient at \n lnrpc address: {:?} ",
+                lnrpc_addr
+            );
+
             NetworkLnRpcClient::new(lnrpc_addr).await?.into()
         }
     } else {
--- a/gateway/ln-gateway/src/bin/gatewayd.rs
+++ b/gateway/ln-gateway/src/bin/gatewayd.rs
@@ -1,5 +1,6 @@
 use std::net::SocketAddr;
 use std::path::PathBuf;
+use std::process::exit;
 
 use clap::Parser;
 use fedimint_client::module::gen::{ClientModuleGenRegistry, DynClientModuleGen};
@@ -13,7 +14,7 @@ use fedimint_core::task::TaskGroup;
 use fedimint_logging::TracingSetup;
 use ln_gateway::client::{DynGatewayClientBuilder, RocksDbFactory, StandardGatewayClientBuilder};
 use ln_gateway::lnd::GatewayLndClient;
-use ln_gateway::lnrpc_client::NetworkLnRpcClient;
+use ln_gateway::lnrpc_client::{DynLnRpcClient, NetworkLnRpcClient};
 use ln_gateway::Gateway;
 use mint_client::modules::ln::{LightningClientGen, LightningModuleTypes};
 use mint_client::modules::mint::{MintClientGen, MintModuleTypes};
@@ -42,6 +43,10 @@ pub struct GatewayOpts {
     /// Public URL to a Gateway Lightning rpc service
     #[arg(long = "lnrpc-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
     pub lnrpc_addr: Option<Url>,
+
+    /// LND RPC address
+    #[arg(long = "lnd-rpc-host", env = "FM_LND_RPC_ADDR")]
+    pub lnd_rpc_addr: Option<String>,
 
     /// LND TLS cert file path
     #[arg(long = "lnd-tls-cert", env = "FM_LND_TLS_CERT")]
@@ -77,6 +82,7 @@ async fn main() -> Result<(), anyhow::Error> {
         api_addr,
         password,
         lnrpc_addr,
+        lnd_rpc_addr,
         lnd_tls_cert,
         lnd_macaroon,
     } = GatewayOpts::parse();
@@ -93,33 +99,25 @@ async fn main() -> Result<(), anyhow::Error> {
     // Create task group for controlled shutdown of the gateway
     let task_group = TaskGroup::new();
 
-    let lnrpc = if let Some(lnrpc_addr) = lnrpc_addr {
-        if let (Some(lnd_tls_cert), Some(lnd_macaroon)) = (lnd_tls_cert, lnd_macaroon) {
-            // Detected LND rpc client configurations
-            info!(
-                "Gateway configured to connect to LND LnRpcClient at \n address: {:?},\n tls cert path: {:?},\n macaroon path: {} ",
-                lnrpc_addr, lnd_tls_cert, lnd_macaroon
-            );
-
-            GatewayLndClient::new(
-                lnrpc_addr.into(),
-                lnd_tls_cert,
-                lnd_macaroon,
-                task_group.clone(),
-            )
+    let lnrpc: DynLnRpcClient = if let Some(lnrpc_addr) = lnrpc_addr {
+        info!(
+            "Gateway configured to connect to remote LnRpcClient at \n lnrpc address: {:?} ",
+            lnrpc_addr
+        );
+        NetworkLnRpcClient::new(lnrpc_addr).await?.into()
+    } else if let (Some(lnd_rpc_addr), Some(lnd_tls_cert), Some(lnd_macaroon)) =
+        (lnd_rpc_addr, lnd_tls_cert, lnd_macaroon)
+    {
+        info!(
+            "Gateway configured to connect to LND LnRpcClient at \n address: {:?},\n tls cert path: {:?},\n macaroon path: {} ",
+            lnd_rpc_addr, lnd_tls_cert, lnd_macaroon
+        );
+        GatewayLndClient::new(lnd_rpc_addr, lnd_tls_cert, lnd_macaroon, task_group.clone())
             .await?
             .into()
-        } else {
-            // Detected generic network rpc client configurations
-            info!(
-                "Gateway configured to connect to generic remote LnRpcClient at \n lnrpc address: {:?} ",
-                lnrpc_addr
-            );
-
-            NetworkLnRpcClient::new(lnrpc_addr).await?.into()
-        }
     } else {
-        panic!("No lightning node provided")
+        error!("No lightning node provided. For CLN set FM_GATEWAY_LIGHTNING_ADDR for CLN. For LND set FM_LND_RPC_ADDR, FM_LND_TLS_CERT, and FM_LND_MACAROON");
+        exit(1);
     };
 
     // Create module decoder registry
--- a/gateway/ln-gateway/src/bin/gatewayd.rs
+++ b/gateway/ln-gateway/src/bin/gatewayd.rs
@@ -113,9 +113,14 @@ async fn main() -> Result<(), anyhow::Error> {
             "Gateway configured to connect to LND LnRpcClient at \n address: {:?},\n tls cert path: {:?},\n macaroon path: {} ",
             lnd_rpc_addr, lnd_tls_cert, lnd_macaroon
         );
-        GatewayLndClient::new(lnd_rpc_addr, lnd_tls_cert, lnd_macaroon, task_group.clone())
-            .await?
-            .into()
+        GatewayLndClient::new(
+            lnd_rpc_addr,
+            lnd_tls_cert,
+            lnd_macaroon,
+            task_group.make_subgroup().await,
+        )
+        .await?
+        .into()
     } else {
         error!("No lightning node provided. For CLN set FM_GATEWAY_LIGHTNING_ADDR for CLN. For LND set FM_LND_RPC_ADDR, FM_LND_TLS_CERT, and FM_LND_MACAROON");
         exit(1);
@@ -150,7 +155,7 @@ async fn main() -> Result<(), anyhow::Error> {
         client_builder,
         decoders,
         module_gens,
-        task_group.clone(),
+        task_group.make_subgroup().await,
     )
     .await;
 
--- a/gateway/ln-gateway/src/bin/gatewayd.rs
+++ b/gateway/ln-gateway/src/bin/gatewayd.rs
@@ -12,6 +12,7 @@ use fedimint_core::module::ModuleCommon;
 use fedimint_core::task::TaskGroup;
 use fedimint_logging::TracingSetup;
 use ln_gateway::client::{DynGatewayClientBuilder, RocksDbFactory, StandardGatewayClientBuilder};
+use ln_gateway::lnd::GatewayLndClient;
 use ln_gateway::lnrpc_client::NetworkLnRpcClient;
 use ln_gateway::Gateway;
 use mint_client::modules::ln::{LightningClientGen, LightningModuleTypes};
@@ -41,6 +42,14 @@ pub struct GatewayOpts {
     /// Public URL to a Gateway Lightning rpc service
     #[arg(long = "lnrpc-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
     pub lnrpc_addr: Option<Url>,
+
+    /// LND TLS cert file path
+    #[arg(long = "lnd-tls-cert", env = "FM_LND_TLS_CERT")]
+    pub lnd_tls_cert: Option<String>,
+
+    /// LND macaroon file path
+    #[arg(long = "lnd-macaroon", env = "FM_LND_MACAROON")]
+    pub lnd_macaroon: Option<String>,
 }
 
 // Fedimint Gateway Binary
@@ -66,8 +75,10 @@ async fn main() -> Result<(), anyhow::Error> {
         data_dir,
         listen,
         api_addr,
-        lnrpc_addr,
         password,
+        lnrpc_addr,
+        lnd_tls_cert,
+        lnd_macaroon,
     } = GatewayOpts::parse();
 
     info!(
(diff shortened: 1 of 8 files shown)


okjodom:
@justinmoon, here is a clean cherry pick of core logic changes to support LND. It's hard for me to follow which of the test infra changes are still in progress, to be added to this change before we can activate PR. Do take a look...

integrationtests/tests/fixtures/mod.rs
```diff
@@ -114,31 +124,37 @@ pub struct Fixtures {
...
-    f(
-        fixtures.fed,
-        fixtures.user,
-        fixtures.bitcoin,
-        fixtures.gateway,
-        fixtures.lightning,
-    )
-    .await;
-    fixtures.task_group.shutdown_join_all(None).await
+    let nodes = vec![GatewayNode::Cln, GatewayNode::Lnd];
```
justinmoon, line 138: This just runs each integration test once with LND, once with CLN. I should parameterize this better so that non-lightning tests don't run twice ...

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -40,7 +42,19 @@ pub struct GatewayOpts {
 
     /// Public URL to a Gateway Lightning rpc service
     #[arg(long = "lnrpc-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
-    pub lnrpc_addr: Url,
+    pub lnrpc_addr: Option<Url>,
```
m1sterc001guy, line 45: Why are these options? I see below that if they're none, we just exit with error code 1. Can we make these parameters required?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+use crate::gatewaylnrpc::get_route_hints_response::RouteHint;
+use crate::gatewaylnrpc::{
+    CompleteHtlcsRequest, CompleteHtlcsResponse, GetPubKeyResponse, GetRouteHintsResponse,
+    PayInvoiceRequest, PayInvoiceResponse, SubscribeInterceptHtlcsRequest,
+    SubscribeInterceptHtlcsResponse,
+};
+use crate::lnrpc_client::{HtlcStream, ILnRpcClient};
+use crate::GatewayError;
+
+type OutcomeMap = Arc<Mutex<HashMap<sha256::Hash, (LndSenderRef, Option<CircuitKey>)>>>;
```
m1sterc001guy, line 27: Comment here on what the key/value of the map are would be nice

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+#[async_trait]
+impl ILnRpcClient for GatewayLndClient {
+    async fn pubkey(&self) -> crate::Result<GetPubKeyResponse> {
+        let mut client = self.client.clone();
+
+        let info = client
+            .lightning()
+            .get_info(GetInfoRequest {})
+            .await
+            .expect("failed to get info")
```
m1sterc001guy, line 74: Can these expects be ?

m1sterc001guy: Same with below

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        let pub_key: PublicKey = info.identity_pubkey.parse().expect("invalid pubkey");
+        info!("fetched pubkey {:?}", pub_key);
+
+        Ok(GetPubKeyResponse {
+            pub_key: pub_key.serialize().to_vec(),
+        })
+    }
+
+    async fn routehints(&self) -> crate::Result<GetRouteHintsResponse> {
```
m1sterc001guy, line 85: Maybe add a comment that this isn't properly implemented yet

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+        Ok(GetRouteHintsResponse {
+            route_hints: vec![RouteHint { hops: vec![] }],
+        })
+    }
+
+    async fn pay(&self, invoice: PayInvoiceRequest) -> crate::Result<PayInvoiceResponse> {
+        let mut client = self.client.clone();
```
m1sterc001guy, line 92: Can this be borrowed instead of copied?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+        Ok(PayInvoiceResponse {
+            preimage: send_response.payment_preimage,
+        })
+    }
+
+    async fn subscribe_htlcs<'a>(
+        &self,
+        subscription: SubscribeInterceptHtlcsRequest,
+    ) -> crate::Result<HtlcStream<'a>> {
+        let channel_size = 1024;
```
m1sterc001guy, line 121: Can we use a constant for this? Not sure what the unit is at first glance

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+        let (lnd_tx, lnd_rx) = mpsc::channel::<ForwardHtlcInterceptResponse>(channel_size);
+
+        // Channel to send intercepted htlc to gatewayd for processing
+        let (gwd_tx, gwd_rx) =
+            mpsc::channel::<Result<SubscribeInterceptHtlcsResponse, tonic::Status>>(channel_size);
+
+        let scid = subscription.short_channel_id;
+        let mut client = self.client.clone();
+
+        let mut tg = self.task_group.clone();
```
m1sterc001guy, line 133: If I recall correctly, isn't there a way to make a subgroup instead of cloning the original group?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+        let outcomes = self.outcomes.clone();
+
+        // Spawn an LND interceptor task
```
m1sterc001guy, line 136: Update comment on why it needs to be on a separate thread

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                .router()
+                .htlc_interceptor(ReceiverStream::new(lnd_rx))
+                .await
+                .map_err(|e| {
+                    error!("Failed to connect to lnrpc server: {:?}", e);
+                    GatewayError::Other(anyhow!("Failed to subscribe to LND htlc stream"))
+                }) {
+                Ok(stream) => stream.into_inner(),
+                Err(e) => {
+                    let _ = gwd_tx
```
m1sterc001guy, line 148: remove let _

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                        .send(Err(tonic::Status::new(
+                            tonic::Code::Internal,
+                            e.to_string(),
+                        )))
+                        .await;
+                    return;
```
m1sterc001guy, line 154: Is this necessary?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                }
+            };
+
+            while let Some(htlc) = match htlc_stream.message().await {
+                Ok(htlc) => htlc,
+                Err(e) => {
+                    error!("Error received over HTLC subscriprion: {:?}", e);
+                    let _ = gwd_tx
```
m1sterc001guy, line 162: remove let _

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                }
+            } {
+                trace!("handling htlc {:?}", htlc);
+                let response: Option<ForwardHtlcInterceptResponse> =
+                    if htlc.outgoing_requested_chan_id != scid {
+                        // Pass through: This HTLC doesn't belong to the current subscription
+                        // Forward it to the next interceptor or next node
+                        Some(ForwardHtlcInterceptResponse {
+                            incoming_circuit_key: htlc.incoming_circuit_key,
+                            action: 2,
```
m1sterc001guy, line 178: Would prefer if we can use enums here since 2 is difficult to understand what that means

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                            Err(e) => {
+                                error!("Failed to send HTLC to gatewayd for processing: {:?}", e);
+                                Some(cancel_intercepted_htlc(htlc.incoming_circuit_key))
+                            }
+                        }
+                    };
+
+                if response.is_some() {
+                    // TODO: Consider retrying this if the send fails
+                    let _ = lnd_tx.send(response.unwrap()).await.map_err(|e| {
```
m1sterc001guy, line 218: remove let _ ?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                        error!("Failed to send response to LND: {:?}", e);
+                        // The HTLC will timeout and LND will automatically cancel it.
```
m1sterc001guy, line 220: This could be a long timeout though right?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                return Err(GatewayError::LnRpcError(tonic::Status::invalid_argument(
+                    e.to_string(),
+                )));
+            }
+        };
+
+        info!("Completing htlc with reference, {:?}", hash);
+
+        if let Some((outcome, incoming_circuit_key)) = self.outcomes.lock().await.remove(&hash) {
+            let htlca_res = match action {
```
m1sterc001guy, line 255: nit: would prefer spelling out action, I initially thought the a was a typo :)

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                Some(Action::Cancel(Cancel { reason: _ })) => cancel_intercepted_htlc(None),
+                None => {
+                    error!("No action specified for intercepted htlc id: {:?}", hash);
+                    return Err(GatewayError::LnRpcError(tonic::Status::internal(
+                        "No action specified on this intercepted htlc",
+                    )));
+                }
+            };
+
+            // TODO: Consider retrying this if the send fails
```
m1sterc001guy, line 272: Nit: this looks like the same code as above, maybe we could make it a function

integrationtests/tests/fixtures/mod.rs
m1sterc001guy: Maybe this array should be passed in by the test itself? Otherwise we're going to need some mechanism for determining which test is and isnt a lightning test

integrationtests/tests/fixtures/mod.rs
```diff
@@ -212,15 +229,23 @@ pub async fn fixtures(num_peers: u16) -> anyhow::Result<Fixtures> {
...
+            let lnd_tls_cert = env::var("FM_LND_TLS_CERT").unwrap();
+            let lnd_client = connect(
+                lnd_rpc_addr.clone(),
+                lnd_tls_cert.clone(),
+                lnd_macaroon.clone(),
+            )
+            .await
+            .unwrap();
+            let rpc_lnd = Arc::new(Mutex::new(lnd_client.clone()));
+            let lightning = RealLightningTest::new(rpc_cln, rpc_lnd, gateway_node.clone()).await;
```
m1sterc001guy, line 246: Do we construct both the CLN rpc client and the LND rpc client because both might be used in the test, but only one is used for the gateway?

integrationtests/tests/fixtures/real.rs
```diff
@@ -10,59 +8,104 @@ use async_trait::async_trait;
...
-
-        Invoice::from_str(&invoice_resp.payment_request).unwrap()
+        match self.gateway_node {
+            // If we're using CLN as the gateway, use LND to fetch an invoice
+            GatewayNode::Cln => {
+                info!("fetching invoice from lnd");
+                let mut lnd_rpc = self.rpc_lnd.lock().await;
+                let tonic_invoice = match expiry_time {
+                    Some(expiry) => LndInvoice {
+                        value_msat: amount.msats as i64,
```
m1sterc001guy, line 51: Strange....lnd allows value_msat to be negative?

integrationtests/tests/fixtures/real.rs
```diff
@@ -10,59 +8,104 @@ use async_trait::async_trait;
...
+
+                Invoice::from_str(&invoice_resp.payment_request).unwrap()
+            }
+            // If we're using LND as the gateway, use CLN to fetch an invoice
+            GatewayNode::Lnd => {
+                info!("fetching invoice from cln");
+                let random: u64 = rand::random();
+                let invoice_req = requests::InvoiceRequest {
+                    amount_msat: AmountOrAny::Amount(ClnRpcAmount::from_msat(amount.msats)),
+                    description: "".to_string(),
```
m1sterc001guy, line 75: I haven't been following all the drama all twitter about this....but will the recent CLN description change break us if we use empty string here?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+    PayInvoiceRequest, PayInvoiceResponse, SubscribeInterceptHtlcsRequest,
+    SubscribeInterceptHtlcsResponse,
+};
+use crate::lnrpc_client::{HtlcStream, ILnRpcClient};
+use crate::GatewayError;
+
+type OutcomeMap = Arc<Mutex<HashMap<sha256::Hash, (LndSenderRef, Option<CircuitKey>)>>>;
+
+pub struct GatewayLndClient {
+    client: LndClient,
```
justinmoon, line 30: This should be wrapped in a mutex if it will ever be used concurrently ...

gateway/ln-gateway/src/bin/gatewayd.rs
justinmoon: `lnrpc_addr` uses the our custom gRPC spec to talk to an abstract lightning node. Currently that's just the core-lightning plugin, but in the future it could be an integration to some LSP etc.

The `lnd_` parameters are specifically for LND.

The invariant should be that either `lnrpc_addr` is provided, or all the `lnd_` variables are provided, and not both. One or the other. Line 119 checks whether one set or the other is present, but doesn't complain if both are present. But I think it probably should.

gateway/ln-gateway/src/lnd.rs
justinmoon: maybe we should love if this send fails?

gateway/ln-gateway/src/lnd.rs
justinmoon: log error

gateway/ln-gateway/src/lnd.rs
justinmoon: great point

gateway/ln-gateway/src/lnd.rs
justinmoon: `htlca_res` isn't a great name

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+                    )));
+                }
+            };
+
+            // TODO: Consider retrying this if the send fails
+            let _ = outcome.send(htlca_res).await.map_err(|_| {
+                GatewayError::LnRpcError(tonic::Status::internal(
+                    "Failed to send htlc processing outcome to LND node",
+                ))
+            });
```
justinmoon, line 277: Don't suppress this error.

gateway/ln-gateway/src/lnd.rs
```diff
@@ -0,0 +1,302 @@
...
+            info!("HTLC subscription stream ended");
+        })
+        .await;
+
+        Ok(Box::pin(ReceiverStream::new(gwd_rx)))
+    }
+
+    async fn complete_htlc(
+        &self,
+        outcome: CompleteHtlcsRequest,
```
justinmoon, line 235: re-used variable name `outcomer`

integrationtests/tests/fixtures/mod.rs
justinmoon: Relevant to https://github.com/fedimint/fedimint/pull/1702

integrationtests/tests/fixtures/mod.rs
justinmoon: make a comment about why 2 are required.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -255,6 +280,30 @@ pub async fn fixtures(num_peers: u16) -> anyhow::Result<Fixtures> {
             ));
             user.client.await_consensus_block_height(0).await?;
 
+            // gateway
```
justinmoon, line 283: describe here why we're

integrationtests/tests/fixtures/real.rs
```diff
@@ -10,59 +8,104 @@ use async_trait::async_trait;
...
 
+use super::GatewayNode;
 use crate::fixtures::LightningTest;
 
 #[derive(Clone)]
 pub struct RealLightningTest {
     rpc_cln: Arc<Mutex<ClnRpc>>,
     rpc_lnd: Arc<Mutex<LndClient>>,
     initial_balance: Amount,
+    // Which lightning node to use as the gateway?
```
justinmoon, line 36: ///

integrationtests/tests/fixtures/real.rs
justinmoon: let's just leave it here and update if CI ever breaks. this is just a test fixture so not a big deal.

gateway/ln-gateway/src/lnd.rs
okjodom: [#1936](https://github.com/fedimint/fedimint/issues/1936) is the preferred fix

gateway/ln-gateway/src/lnd.rs
okjodom: tracked a TODO for issue #1953

gateway/ln-gateway/src/bin/gatewayd.rs
okjodom: tracked todo for issue #1955

integrationtests/tests/fixtures/mod.rs
okjodom: wontfix for now

gateway/ln-gateway/src/lnd.rs
okjodom: correct

integrationtests/tests/fixtures/real.rs
okjodom: won't fix, for now

integrationtests/tests/fixtures/real.rs
okjodom: indeed, that's what they have in spec: https://github.com/lightningnetwork/lnd/blob/master/lnrpc/lightning.proto#L3377

opportunity for LND contrib?

integrationtests/tests/fixtures/mod.rs
justinmoon: done

integrationtests/tests/fixtures/mod.rs
justinmoon: done

gateway/ln-gateway/src/lnd.rs
justinmoon: I think this is actually as good as we can do here. This task group is just to spawn the htlc subscription task, so we shouldn't need another child.

I made another commit that makes sure that `GatewayLndClient::new` always gets a new subgroup, and added a comment about the purpose of this task group.

While cloning sucks, this basically just cloning an `Arc` so it's pretty cheap ...

integrationtests/tests/fixtures/mod.rs
justinmoon: I updated this

gateway/ln-gateway/src/lnd.rs
justinmoon: done

gateway/ln-gateway/src/lnd.rs
justinmoon: done

gateway/ln-gateway/src/lnd.rs
justinmoon: done

gateway/ln-gateway/src/lnd.rs
justinmoon: done

justinmoon (approved):

integrationtests/tests/tests.rs
```diff
@@ -235,8 +235,8 @@ async fn wallet_peg_ins_that_are_unconfirmed_are_rejected() -> Result<()> {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn wallet_peg_outs_must_wait_for_available_utxos() -> Result<()> {
-    test(2, |fed, user, bitcoin, _, _| async move {
-        // at least one epoch needed to establish fees
+    non_lightning_test(2, |fed, user, bitcoin, _, _| async move {
+        // at least one epoch needed to estabilish fees
```
m1sterc001guy, line 239: nit: this is introducing a typo, it was correct before :)

m1sterc001guy (approved):
LGTM, let's iterate from here

integrationtests/tests/tests.rs
okjodom: :) will fix!

PR #1913, opened 2023-03-14
 Rename function `max_retries` to `max_attempts` parameter for consistency

mxz42:
Renames `max_retries` to `max_attempts` in the db `autocommit` function.

- This *does* change the behaviour of the function slightly. Previously, if `Some(5)` was passed as `max_retries` the closure would run once, and then up to five times (equal to 6 times in total). Passing `Some(5)` now runs the closure up to five times in total, hence `max_attempts`.
- Move the part to increment the attempt counter to the top part of the loop. This is good practice as that will always ensure that every run of the loop will increment the counter. For instance, a future change may introduce an early `continue` that could have introduced a bug if the counter would not have been incremented manually.

This also makes it consistent with the [`retry` function](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/utils.rs#L20)

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -108,20 +108,20 @@ struct DatabaseInner<Db: IDatabase + ?Sized> {
 pub enum AutocommitError<E> {
     /// Committing the transaction failed too many times, giving up
     CommitFailed {
-        /// Number of retries
-        retries: usize,
+        /// Number of attempts
+        attempts: usize,
         /// Last error on commit
         last_error: anyhow::Error,
     },
     /// Error returned by the closure provided to `autocommit`. If returned no
     /// commit was attempted in that round
     ClosureError {
-        /// Retry on which the closure returned an error
+        /// The attempt on which the closure returned an error
         ///
         /// Values other than 0 typically indicate a logic error since the
         /// closure given to `autocommit` should not have side effects
         /// and thus keep succeeding if it succeeded once.
-        retries: usize,
+        attempts: usize,
         /// Error returned by the closure
         error: E,
     },
@@ -168,8 +168,8 @@ impl Database {
 
     /// Runs a closure with a reference to a database transaction and tries to
     /// commit the transaction if the closure returns `Ok` and rolls it back
-    /// otherwise. If committing fails the closure is run again for up to
-    /// `max_retries` times. If `max_retries` is `None` it will run
+    /// otherwise. If committing fails the closure is run for up to
+    /// `max_attempts` times. If `max_attempts` is `None` it will run
     /// `usize::MAX` times which is close enough to infinite times.
     ///
     /// The closure `tx_fn` provided should not have side effects outside of the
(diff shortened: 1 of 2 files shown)


justinmoon (dismissed):

dpc:
The master branch CI failures should be fixed, so please rebase.

fedimint-core/src/db/mod.rs
```diff
@@ -206,29 +216,29 @@ impl Database {
                         Ok(()) => {
                             return Ok(val);
                         }
-                        Err(e) if max_retries.map(|mr| mr <= retries).unwrap_or(false) => {
```
dpc, line 209: Previous version was shorter and better, IMO. :P

dpc (approved):

fedimint-core/src/db/mod.rs
mxz42: :D It was definitely shorter. It just took me a couple minutes to wrap my head around the line every time I looked at it. I felt that it's kinda like the double-negative example we had with the environment variable for testing. 

I can revert that if you like :)

dpc: > I felt that it's kinda like the double-negative example we had with the environment variable for testing.

But ... there's no negation here, no? "max_retries <= retries if Some, otherwise false"

What I don't like is, personally, is just using the `match x { pattern if expression => ... }` thing altogether. I'd merge the `Err` case and make it a normal:

```
Err(e) => if max_retries.map(|mr| mr <= retries).unwrap_or(false) { 
   // try again
   } else {
      return Err(... stuff... );
   }
```

Ability to add condition to pattern is sometimes nice, but oftentimes it's just extra idiosyncrasy.

dpc: > I can revert that if you like :)

We should let the next reviewer be a tie-breaker. :D

mxz42: hm, right, 
that 
```rust
Err(e) => if max_retries.map(|mr| mr <= retries).unwrap_or(false) { 
   // try again
   } else {
      return Err(... stuff... );
   }
```
would work as well and is a bit more concise which I like.

If nobody steps up with an opinion until tomorrow when I look at this again, I'll change it to that, and we can merge it. I don't have a strong opinion on this and the change is about the `max_attempts` thing anyway :D

jkitman: I vote for the old way or just make max_retries `u64` and default to MAX, so you can remove that comment too.

fedimint-core/src/db/mod.rs
```diff
@@ -191,13 +191,23 @@ impl Database {
...
     {
-        let mut retries: usize = 0;
+        assert_ne!(max_attempts, Some(0));
+        let mut curr_attempts: usize = 0;
+
         loop {
+            // The `checked_add()` function is used to catch the `usize` overflow.
+            // With `usize=32bit` and an assumed time of 1ms per iteration, this would crash
+            // after ~50 days. But if that's the case, something else must be wrong.
+            // With `usize=64bit` it would take much longer, obviously.
```
elsirion, line 206: Love this attention to detail!

m1sterc001guy (dismissed):
LGTM, agree with kitman that the `attempts` variable can just be made a u64 to avoice some of the new syntax.

elsirion:
@mxz42 merge or do you want to address any of the comments?

fedimint-core/src/db/mod.rs
```diff
@@ -191,13 +191,23 @@ impl Database {
...
         &'s self,
         tx_fn: F,
-        max_retries: Option<usize>,
+        max_attempts: Option<usize>,
     ) -> Result<T, AutocommitError<E>>
     where
         for<'a> F: Fn(&'a mut DatabaseTransaction<'dt>) -> BoxFuture<'a, Result<T, E>>,
     {
-        let mut retries: usize = 0;
+        assert_ne!(max_attempts, Some(0));
```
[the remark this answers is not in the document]
mxz42: add a panics section in the function documentation

fedimint-core/src/db/mod.rs
mxz42: thanks :) I think at some point we talked about this a while back

fedimint-core/src/db/mod.rs
mxz42: ok, thanks @jkitman. reverted to more or less the old way that also removes the comment

mxz42:
sorry that it took a bit longer to get back to this. real life intervened. 
Thanks for the review @ all.

> @mxz42 merge or do you want to address any of the comments?

I addressed them.


> LGTM, agree with kitman that the `attempts` variable can just be made a u64 to avoice some of the new syntax.

@m1sterc001guy hm, I don't quite understand what you mean by "new syntax" as the variable was an `Option` before. If you mean my way of stating that `if` expression, then ok, I get it :'D That's gone now. 

In general, would it makes sense to drop the `max_attempts` parameter completely and just default to `usize::MAX` attempts as @jkitman suggested? (that's how I read the suggestion)

jkitman (approved):

dpc (approved):

PR #1917, opened 2023-03-14
Latency tests set `-xo pipefile` and `RUST_LOG=info`

justinmoon:


scripts/latency-test.sh
```diff
@@ -1,12 +1,10 @@
 #!/usr/bin/env bash
 # Runs a test to determine the latency of certain user actions
 
-set -eu
+set -euxo pipefail
+export RUST_LOG=info
 FM_FED_SIZE=${1:-4}
 ITERATIONS=${2:-10}
-if [ -z "${RUST_LOG:-}" ]; then
```
dpc, line 7: No, no. Keep the ability to set custom `RUST_LOG`. It's very handy. Just change the default to `info`

justinmoon: Good point. Updated all the shell-based tests to default to `info` but allow overriding ...

Maan2003 (approved):

Issue #1919, opened 2023-03-14
A command to deserialize and print debug output of a `Transaction` (and possibly other most common structures)


dpc:
If I have a hex output from the CI, would be nice not to have to hack test code to see what's in there.

https://gist.github.com/elsirion/3afaf6da90cec6ccf0ee2457594b4441

In the link some code  that does the same thing.

ngutech21:
I would like to work on this issue. Where do I find more examples of these Transactions in the log? How can I reproduce them?

ngutech21:
This issue can be closed because branch got merged

PR #1928, opened 2023-03-15
feat: add decode-tx command to cli

ngutech21:
fix for #1919

Is there a better way to get the default module decoders?

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -367,7 +367,7 @@ enum Command {
     #[clap(hide = true)]
     WipeNotes,
 
-    /// Decode a transaction hex string into its JSON representation
+    /// Decode a transaction hex string and print it to stdout
     DecodeTransaction { hex_string: String },
 
     /// Signal a consensus upgrade
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -21,7 +21,7 @@ use fedimint_core::api::{
     WsClientConnectInfo, WsFederationApi,
 };
 use fedimint_core::config::{load_from_file, ClientConfig, FederationId};
-use fedimint_core::db::Database;
+use fedimint_core::db::{Database, DatabaseValue};
 use fedimint_core::module::{ApiAuth, ApiRequestErased};
 use fedimint_core::query::EventuallyConsistent;
 use fedimint_core::task::TaskGroup;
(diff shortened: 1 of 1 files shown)


client/cli/src/main.rs
```diff
@@ -738,5 +761,43 @@ async fn handle_command(
                 Some(e.into()),
             )),
         },
+        Command::DecodeTransaction { hex_string } => {
+            let decoders = ModuleDecoderRegistry::from_iter([
```
dpc, line 765: The correct way would be #1931 , but just using `module_decode_stubs` would be a good start.

dpc (changes requested):
PLease use `module_decode_stubs`

client/cli/src/main.rs
ngutech21: thanks, fixed it

dpc (approved):
LGTM, but some unused imports are triggering CI.

dpc:
Please squash (seems small enough to be one atomic commit) and let's land.

ngutech21:
squashed the commits

client/cli/src/main.rs
```diff
@@ -331,6 +335,9 @@ enum Command {
     /// Wipe the notes data from the DB. Useful for testing backup & restore
     #[clap(hide = true)]
     WipeNotes,
+
+    /// Decode a transaction hex string into its JSON representation
```
elsirion, line 339: It's actually not encoded to JSON but debug print, right? We'd need to patch in `Serialize` into the `IInput` and `IOutput` traits for that iirc.

client/cli/src/main.rs
```diff
@@ -127,6 +127,10 @@ enum CliOutput {
     },
 
     Backup,
+
+    DecodeTransaction {
+        transaction: String,
```
elsirion, line 132: If the encoding ever becomes JSON this will have to be a `Value`

elsirion (approved):
LGTM, a minor doc nit that would be nice to fix

elsirion:
Also needs rebase

elsirion (approved):

elsirion:
I think serializing to JSON is a good goal longer term, but will require some changes to the type-erased input/output variants. I think @mxz42 has a good grasp on it, maybe you can work together.

dpc (approved):

ngutech21:
> I think serializing to JSON is a good goal longer term, but will require some changes to the type-erased input/output variants. I think @mxz42 has a good grasp on it, maybe you can work together.

Sounds good to me

PR #1932, opened 2023-03-15
Consensus upgrade

jkitman:
PR for adding consensus upgrades to the API, fixing #948.
- Adds the consensus upgrade signaling code via an authorized API.
- Introduces `ApiEvent` so we can send non-transaction events from the API to consensus thread
- Adds a new integration test for signaling the upgrade
- Eventually the upgrade signaling/start-up should move into an admin UI.

Will follow-up with a refactor suggested by @Maan2003 to introduce `ApiEndpointContext`

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -2,26 +2,30 @@ use core::fmt;
 use std::collections::BTreeMap;
 use std::error::Error;
 use std::fmt::Debug;
+use std::fs;
 use std::io::Write;
 use std::path::PathBuf;
 use std::process::exit;
 use std::str::FromStr;
 use std::sync::Arc;
 
+use anyhow::format_err;
 use bitcoin::{secp256k1, Address, Network, Transaction};
 use clap::{Parser, Subcommand};
+use fedimint_aead::get_password_hash;
 use fedimint_client::module::gen::{
     ClientModuleGenRegistry, ClientModuleGenRegistryExt, DynClientModuleGen,
 };
 use fedimint_core::api::{
-    FederationApiExt, GlobalFederationApi, IFederationApi, WsClientConnectInfo, WsFederationApi,
+    FederationApiExt, FederationError, GlobalFederationApi, IFederationApi, WsAuthenticatedApi,
+    WsClientConnectInfo, WsFederationApi,
 };
 use fedimint_core::config::{load_from_file, ClientConfig, FederationId};
 use fedimint_core::db::Database;
-use fedimint_core::module::ApiRequestErased;
+use fedimint_core::module::{ApiAuth, ApiRequestErased};
 use fedimint_core::query::EventuallyConsistent;
 use fedimint_core::task::TaskGroup;
-use fedimint_core::{Amount, OutPoint, TieredMulti, TransactionId};
+use fedimint_core::{Amount, OutPoint, PeerId, TieredMulti, TransactionId};
 use fedimint_ln_client::LightningClientGen;
 use fedimint_logging::TracingSetup;
 use fedimint_mint_client::MintClientGen;
@@ -31,7 +35,7 @@ use mint_client::modules::wallet::txoproof::TxOutProof;
 use mint_client::modules::wallet::WalletClientGen;
 use mint_client::utils::{
     from_hex, parse_bitcoin_amount, parse_ecash, parse_fedimint_amount, parse_node_pub_key,
-    serialize_ecash,
+    parse_peer_id, serialize_ecash,
 };
 use mint_client::{module_decode_stubs, Client, UserClientConfig};
 use serde::{Deserialize, Serialize};
@@ -127,6 +131,8 @@ enum CliOutput {
     },
 
     Backup,
+
+    SignalUpgrade,
 }
 
 impl fmt::Display for CliOutput {
@@ -156,6 +162,27 @@ struct CliError {
     message: String,
     #[serde(skip_serializing)]
     raw_error: Option<Box<dyn Error>>,
+}
+
+// TODO: Refactor federation API errors to just delegate to this
+impl From<FederationError> for CliError {
+    fn from(e: FederationError) -> Self {
+        CliError::from(
+            CliErrorKind::GeneralFederationError,
+            "Failed API call",
+            Some(e.into()),
+        )
+    }
+}
+
+impl From<anyhow::Error> for CliError {
+    fn from(e: anyhow::Error) -> Self {
+        CliError::from(
+            CliErrorKind::GeneralFederationError,
+            "Failed",
+            Some(e.into()),
+        )
+    }
 }
 
 impl Debug for CliError {
@@ -331,6 +358,18 @@ enum Command {
     /// Wipe the notes data from the DB. Useful for testing backup & restore
     #[clap(hide = true)]
     WipeNotes,
+
+    /// Signal a consensus upgrade
+    SignalUpgrade {
+        /// Location of the salt file
+        salt_path: PathBuf,
+        /// Peer id of the guardian
+        #[arg(value_parser = parse_peer_id)]
+        our_id: PeerId,
+        /// Guardian password for authentication
+        #[arg(env = "FM_PASSWORD")]
+        password: String,
+    },
 }
 
 trait ErrorHandler<T, E> {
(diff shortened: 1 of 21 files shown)


client/cli/src/main.rs
```diff
@@ -331,6 +362,20 @@ enum Command {
     /// Wipe the notes data from the DB. Useful for testing backup & restore
     #[clap(hide = true)]
     WipeNotes,
+
+    /// Signal a consensus upgrade
+    SignalUpgrade {
+        /// Guardian password for authentication
+        password: String,
```
dpc, line 369: Take from the env var via `#[clap(env = "...")]`?

client/cli/src/main.rs
```diff
@@ -331,6 +362,20 @@ enum Command {
...
+        /// Location of the salt file
+        salt_path: PathBuf,
+        /// Peer id of the guardian
+        #[clap(value_parser = parse_peer_id)]
```
dpc, line 373: Probably we should put it into fedimint-core behind a feature flag, gated on optional `clap` dependency. Could be a separate PR.

scripts/upgrade-test.sh
```diff
@@ -0,0 +1,48 @@
...
+server1=$(tail -1 $FM_PID_FILE | head -1)
+
+
+function wait_server_shutdown() {
+  echo "Waiting for $1 to shutdown..."
+  tail --pid=$1 -f /dev/null
+  echo "Server $1 has shutdown."
+}
+
+mine_blocks 110
```
justinmoon, line 21: FYI: we mine 101 blocks on bitcoind startup now so this might not be necessary https://github.com/fedimint/fedimint/blob/37a70ba4ef609a456bfb1323ca2a7b598f3b3d1b/fedimint-testing/src/bin/fixtures.rs#L161

scripts/upgrade-test.sh
```diff
@@ -0,0 +1,48 @@
...
+await_fedimint_block_sync
+await_all_peers
+
+# test a consensus upgrade
+$FM_MINT_CLIENT signal-upgrade "pass0" $FM_CFG_DIR/server-0/private.salt 0
+$FM_MINT_CLIENT signal-upgrade "pass1" $FM_CFG_DIR/server-1/private.salt 1
+
+mine_blocks 10
+await_fedimint_block_sync
+await_all_peers
```
justinmoon, line 31: Why do we do this instead of just having `server-2` signal right away? Comment would be nice ...

client/cli/src/main.rs
jkitman: Yep, I just put the function next to the other ones.

flake.crane.nix
```diff
@@ -269,6 +269,13 @@ rec {
     buildPhaseCargoCommand = "patchShebangs ./scripts ; ./scripts/reconnect-test.sh";
   });
 
+  cliTestUpgrade = craneLib.mkCargoDerivation (commonCliTestArgs // {
```
dpc, line 272: You need to add this to one more place in this file if you want it run. And then to `./script/test-ci-all.sh` to trigger from the CI.

fedimint-core/src/api.rs
```diff
@@ -775,6 +775,30 @@ impl<C> WsFederationApi<C> {
...
+
+    pub async fn signal_upgrade(&self) -> FederationResult<()> {
+        self.inner
+            .request_current_consensus(
+                "/upgrade".to_owned(),
+                ApiRequestErased::default().with_auth(&self.auth),
+            )
+            .await
+    }
+}
```
justinmoon, line 800: nit: would be easier to review if these auth changes were in a separate commit from the signalling changes

scripts/upgrade-test.sh
jkitman: True, I don't need it there.

fedimint-server/src/consensus/mod.rs
```diff
@@ -81,6 +82,13 @@ pub struct ConsensusProposal {
     pub force_new_epoch: bool,
 }
 
+/// Events that can be sent from the API to consensus thread
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub enum ApiEvent {
+    Transaction(Transaction),
+    UpgradeSignal,
+}
+
```
justinmoon, line 91: nit: would be easier to review if this refactor was separate commit

flake.crane.nix
jkitman: Where was the other place in the file?  I didn't see that for the other tests

fedimint-server/src/net/api.rs
```diff
@@ -266,5 +266,16 @@ fn server_endpoints() -> Vec<ApiEndpoint<FedimintConsensus>> {
                 Ok(fedimint.get_config_with_sig(dbtx).await)
             }
         },
+        api_endpoint! {
+            "/upgrade",
+            async |fedimint: &FedimintConsensus, _dbtx, _v: (), has_auth| -> () {
+                if has_auth {
+                    fedimint.signal_upgrade().await.map_err(|_| ApiError::server_error("Unable to send signal to server".to_string()))?;
```
justinmoon, line 273: If we're able to send over this channel, we know that the signaling succeeded? We don't need to wait for any response from `fedimintd`?

jkitman: Yes it means the `ApiEvent` is in channel that triggers an epoch with the signal in the consensus thread.

jkitman: We use the same channel for users who submit txs.

flake.crane.nix
dpc:
```
          cli-test = {
            reconnect = craneBuildNative.cliTestReconnect;
            latency = craneBuildNative.cliTestLatency;
            cli = craneBuildNative.cliTestCli;
            rust-tests = craneBuildNative.cliRustTests;
            always-fail = craneBuildNative.cliTestAlwaysFail;
          };
```

fedimint-server/src/net/api.rs
justinmoon: Cool. Thanks for explaining.

fedimint-core/src/module/mod.rs
```diff
@@ -172,8 +188,9 @@ macro_rules! __api_endpoint {
                 $state: &'a Self::State,
                 $dbtx: &'a mut fedimint_core::db::ModuleDatabaseTransaction<'b, ModuleInstanceId>,
                 $param: Self::Param,
-                has_auth: bool,
+                _has_auth: bool,
             ) -> ::std::result::Result<Self::Response, $crate::module::ApiError> {
+                $(let $auth = _has_auth;)?
```
elsirion, line 193: This looks like a hack. Why not just bind the bool arg to `$auth` if present and `_` otherwise?

elsirion (approved):

fedimint-core/src/module/mod.rs
jkitman: Will get rid of this in a follow-up PR, since we want to put `auth` and `dbtx` into a new `ApiEndpointContext` struct.

scripts/upgrade-test.sh
```diff
@@ -0,0 +1,45 @@
...
+  echo "Waiting for $1 to shutdown..."
+  tail --pid=$1 -f /dev/null
+  echo "Server $1 has shutdown."
+}
+
+await_fedimint_block_sync
+
+# test a consensus upgrade
+FM_PASSWORD="pass0" $FM_MINT_CLIENT signal-upgrade $FM_CFG_DIR/server-0/private.salt 0
+FM_PASSWORD="pass1" $FM_MINT_CLIENT signal-upgrade $FM_CFG_DIR/server-1/private.salt 1
```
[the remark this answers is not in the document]
jkitman: I think the best security option is to put the salt in the admin UI.  Failing that, it would live on fedimintd and not be passed.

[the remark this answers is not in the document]
dpc: Make the `salt == hash(federationid || peer_id)` or something? AFAIK it doesn't have to be secret, just different to prevent rainbow tables on some other forms of mass-breaking of keys, no?

fedimint-server/src/lib.rs
```diff
@@ -496,7 +496,7 @@ impl FedimintServer {
         }
 
         tokio::select! {
-            _peek = Pin::new(&mut self.tx_receiver).peek() => Ok(EpochTriggerEvent::NewTransaction),
+            _peek = Pin::new(&mut self.api_receiver).peek() => Ok(EpochTriggerEvent::NewTransaction),
```
Maan2003, line 499: no longer correct?

jkitman: How so?  This is just a rename since we can receive any `ApiEvent`

scripts/upgrade-test.sh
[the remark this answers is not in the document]
jkitman: It doesn't have to be secret, although if you make it secret it does help users with low-entropy passwords.

Although in all likelihood it's probably more of a foot-gun to depend on users to back it up properly so maybe we should just derive it.

fedimint-server/src/lib.rs
Maan2003: is `EpochTriggerEvent::NewTransaction` correct here?

jkitman: Ah I see, yes that should be renamed too.

Maan2003 (approved):

Issue #1920, opened 2023-03-14
Explicitily test a case where ln gateway checks a contract status and gets "pending decryption"


dpc:
This going wrong leads to hard to debug flaky failures and got broken second time as it's timing dependent (after I fixed it the first time in #1798)

dpc:
@okjodom @jkitman

jkitman:
Looks like the changes you made to `await_preimage_decryption` should work, but is it timing out or something?

dpc:
> 00:01:56 fedimint-test-latency> 2023-03-14T20:26:20.784152Z  WARN buy_preimage_from_federation_await_decryption{out_point=OutPoint { txid: 65cca7f04e57154594504fa6a3d368a7b3fa6eaacf3a3c2b99c7bf2a25e49043, out_idx: 0 } contract_id=8aa0a6a4ae3029a00a6577063fef03e7c3b080d25dfad95aebc3c653aa147132}: ln_gateway::actor: Failed to decrypt preimage. Now requesting a refund error=Output outcome error: Transaction rejected: TransactionReplayError(65cca7f04e57154594504fa6a3d368a7b3fa6eaacf3a3c2b99c7bf2a25e49043)

m1sterc001guy:
We have a variation of this test in LNv2 (where the decryption fails client side). Don't think its worth spending more time on it for LNv1.

PR #1933, opened 2023-03-15
Fix latency tests for real

### LINKED ISSUE #1927: Spurious failed transaction cause latency test to fail
This is what is causing random latency test failures:

```00:01:56 fedimint-test-latency> 2023-03-14T20:26:20.784152Z  WARN buy_preimage_from_federation_await_decryption{out_point=OutPoint { txid: 65cca7f04e57154594504fa6a3d368a7b3fa6eaacf3a3c2b99c7bf2a25e49043, out_idx: 0 } contract_id=8aa0a6a4ae3029a00a6577063fef03e7c3b080d25dfad95aebc3c653aa147132}: ln_gateway::actor: Failed to decrypt preimage. Now requesting a refund error=Output outcome error: Transaction rejected: TransactionReplayError(65cca7f04e57154594504fa6a3d368a7b3fa6eaacf3a3c2b99c7bf2a25e49043)```

Workaround: https://github.com/fedimint/fedimint/pull/1926

But we should get rid of it.

I'm not sure if we should change how rejecting txes work, or just not rely on rejected keys for this.

### PR DESCRIPTION
**Pull request by Maan2003**
fix #1927

### DIFF
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -8,7 +8,6 @@ use std::ffi::OsString;
 use std::iter::FromIterator;
 use std::os::unix::prelude::OsStrExt;
 use std::sync::Mutex;
-use std::time::Duration;
 
 use anyhow::format_err;
 use fedimint_core::config::{ConfigResponse, ServerModuleGenRegistry};
@@ -25,7 +24,7 @@ use fedimint_core::module::registry::{
 use fedimint_core::module::{ModuleError, TransactionItemAmount};
 use fedimint_core::outcome::TransactionStatus;
 use fedimint_core::server::{DynServerModule, DynVerificationCache};
-use fedimint_core::task::{sleep, TaskGroup};
+use fedimint_core::task::TaskGroup;
 use fedimint_core::{Amount, NumPeers, OutPoint, PeerId, TransactionId};
 use fedimint_logging::{LOG_CONSENSUS, LOG_CORE};
 use futures::future::select_all;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: fedimint-server/src/consensus/mod.rs [hunk 1]
```diff
@@ -449,8 +448,12 @@ impl FedimintConsensus {
                             .await
                             .expect("Error rolling back to transaction savepoint");
                         warn!(target: LOG_CONSENSUS, %error, "Transaction failed");
-                        dbtx.insert_entry(&RejectedTransactionKey(txid), &format!("{error:?}"))
-                            .await;
+                        // do not insert a RejectedTransactionKey because there must already be
+                        // AcceptedTransactionKey
+                        if !matches!(error, TransactionReplayError(_)) {
```
dpc, line 453: @jkitman Do we have a good way to test this? Basically `test_does_not_insert_rejected_key_on_duplicate_tx`?

dpc (dismissed):
Some short test would be nice, since it's an important behavior that leads to a lot of problems if not preserved, but LGTM.

Please make sure it passess CI at least 3 times, too. :)

justinmoon:
A test would resolve https://github.com/fedimint/fedimint/issues/1920, right?

dpc:
> A test would resolve https://github.com/fedimint/fedimint/issues/1920, right?

No. #1920 is higher level LN-behavior specific test: exercise that pending decryption gets handled correctly (retried). This one would be just an internal test of this specific invariant (no rejected keys on duplicated tx)

fedimint-server/src/consensus/mod.rs
jkitman: Yes, if you look at `cannot_replay_transactions` test.  We could assert the transaction status = accepted at the bottom of the test.

jkitman (changes requested):
You could assert the transaction status = accepted at the bottom of the test `cannot_replay_transactions`.

dpc (approved):

jkitman (approved):

PR #1934, opened 2023-03-15
fix: crane build cache invalidated by top-level dir name

dpc:
See: https://nix.dev/anti-patterns/language#reproducibility-referencing-top-level-directory-with

### DIFF
--- a/flake.crane.nix
+++ b/flake.crane.nix
@@ -2,8 +2,9 @@
 craneLib:
 let
   # filter source code at path `src` to include only the list of `modules`
-  filterModules = modules: src:
+  filterModules = modules: raw-src:
     let
+      src = builtins.path { path = raw-src; name = "src"; };
       basePath = toString src + "/";
       relPathAllCargoTomlFiles = builtins.filter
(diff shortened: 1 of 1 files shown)


justinmoon (approved):

justinmoon:
Explanation here https://discord.com/channels/990354215060795454/1085661891982008341

Maan2003 (approved):
Works on my machine :tm:

flake.crane.nix
```diff
@@ -2,8 +2,9 @@
 craneLib:
 let
   # filter source code at path `src` to include only the list of `modules`
-  filterModules = modules: src:
+  filterModules = modules: raw-src:
     let
+      src = builtins.path { path = raw-src; name = "src"; };
```
elsirion, line 7: Any reason for the string "src" instead of e.g. "fedimint"? What would make the nox path names a bit easier to distinguish.

dpc: No. I've thought it doesn't matter. Will switch to fedimint, I think it's better indeed.

dpc: https://github.com/fedimint/fedimint/pull/1960

PR #1935, opened 2023-03-15
test: global db migration test

m1sterc001guy:
- Adds test for global db migrations that Fedimint uses for consensus.
- Adds database backups so CI will run the `test_migrations`, which will catch if a database migration hasn't been added.

Builds on top of: https://github.com/fedimint/fedimint/pull/1887


m1sterc001guy:
Still investigating a test failure on this PR. Fails when running with nix but not with cargo. I suspect something is wrong with the database backups files.

flake.crane.nix
```diff
@@ -44,7 +44,7 @@ let
   filterWorkspaceDepsBuildFiles = src: filterSrcWithRegexes [ "Cargo.lock" "Cargo.toml" ".cargo" ".cargo/.*" ".*/Cargo.toml" ".*/proto/.*" ] src;
 
   # Filter only files relevant to building the workspace
-  filterWorkspaceFiles = src: filterSrcWithRegexes [ "Cargo.lock" "Cargo.toml" ".cargo" ".cargo/.*" ".*/Cargo.toml" ".*\.rs" ".*\.html" ".*/proto/.*" ] src;
+  filterWorkspaceFiles = src: filterSrcWithRegexes [ "Cargo.lock" "Cargo.toml" ".cargo" ".cargo/.*" ".*/Cargo.toml" ".*\.rs" ".*\.html" ".*/proto/.*" "db/migrations/.*" ] src;
```
m1sterc001guy, line 47: Test issue was that I didn't add the database backups to nix's workspace files. After adding this line, tests pass fine. Let me know if this isn't the right spot.

elsirion: Classic! That bites me every time I add a new crate :laughing:

elsirion (dismissed):

elsirion:
Maybe squash all the commits, otherwise looks good.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1885,6 +1887,7 @@ mod fedimint_migration_tests {
                     create_db_with_v0_data(dbtx).await;
                 })
             },
+            env::var("DB_MIGRATION_DIR").unwrap_or("../../db/migrations".to_string()),
```
dpc, line 1890: Different PR, but the module developer shouldn't need to worry about env variables.

flake.crane.nix
dpc: Good spot.

fedimint-testing/src/lib.rs
```diff
@@ -354,11 +354,11 @@ where
 pub async fn prepare_snapshot<F>(
     snapshot_name: &str,
     prepare_fn: F,
+    parent_dir: String,
```
dpc, line 357: OK... so it is this PR.

Why? :D

dpc (changes requested):
See comments.

modules/fedimint-wallet-server/src/lib.rs
elsirion: What's your proposal? Make it a public constant? Or supply it as a fn argument?

fedimint-testing/src/lib.rs
elsirion: Do you just mean this should be a `PathBuf`, if so I agree and missed that. Comment doesn't make it super clear.

modules/fedimint-wallet-server/src/lib.rs
dpc: I don't understand why we even need it. All I know is this should be an implementation detail.

Module could pass it's `KIND` or something if some kind of of distinct ID is needed.

If it's about relative directory and what's the PWD when given tests are running, then we can e.g. require that `DB_MIGRATION_DIR` is absolute or see what smart stuff do projects like that are doing.

I'm looking quickly at `cargo-fuzz` and it does something like:

https://github.com/rust-fuzz/cargo-fuzz/blob/fa18f0d77aef036302d9f948e80aa12c6b95d57b/tests/tests/project.rs#L8

dpc: https://doc.rust-lang.org/cargo/reference/environment-variables.html

> Cargo exposes these environment variables to your crate when it is compiled. Note that **this applies for running binaries with cargo run and cargo test as well**. To get the value of any of these variables in a Rust program, do this:

So maybe:

> * `CARGO_MANIFEST_DIR` — The directory containing the manifest of your package.

dpc: @m1sterc001guy I hope this is of any help and i'm not completely confused about the reasons here.

m1sterc001guy: Ha thanks for looking into this! Yes the main issue is agreeing upon a directory for all the modules + the global db. And looking at this again now I realize with the nix change that using the environment variable as I am now will not work.

Sounds like `CARGO_MANIFEST_DIR` is the best option

m1sterc001guy: `CARGO_MANIFEST_DIR` gives the directory of the `Cargo.toml` for the crate, not the project repo which isn't ideal. I found this crate [project_root](https://docs.rs/project-root/latest/project_root/) which is a pretty simple way of getting the root of the repo. I wasn't crazy about adding another dependency, but it's only in one place for testing and it allows us to eliminate using environment variables.

fedimint-testing/src/lib.rs
m1sterc001guy: Got rid of `parent_dir` in both `prepare_snapshot` and `validate_migrations`

modules/fedimint-wallet-server/src/lib.rs
dpc: @m1sterc001guy Seems like you can just copy&paste the logic from there https://docs.rs/project-root/latest/src/project_root/lib.rs.html#19-34 :)

dpc: Hmmm... I wonder i we shouldn't have one directory shared between all crates for all snapshots anyway? If the crate is compiled out of tree (which will be the case of external modules) it will keep this snapshots locally for it's own testing purposes anyway...

Or it is really the simplest to require that env to be absolute and use:

```
DB_MIGRATION_DIR="$(realpath ./db/)"
```

in the scripts.

m1sterc001guy: I think we can do either, the backups don't need to be in the same folder together. If they're not in the same folder though, `validate_migrations` needs to know where to look, so it will require another parameter like I had in earlier commits of this PR. Modules could then tell `validate_migrations` where to look for the backup.

Is there an objection to just copying the single function from `project_root`? That seems simple enough, but I'm cool with either approach

dpc (dismissed):

modules/fedimint-wallet-server/src/lib.rs
dpc: I'm fine with how it is now (`<project-root>/db/migrations/<snapshot-name>`).

Possible extensions:

If we want to move it / make it controllable we will make it `<project-root>/$DB_MIGRATION_DIR/<snapshot-name>` where `DB_MIGRATION_DIR` defaults to `db/migrations` if not set.

We could make the logic detect if `DB_MIGRATION_DIR` is relative or absolute, and use the above on when relative and `$DB_MIGRATION_DIR/<snapshot-name>` if absolute (In other words: ignore <project-root>).

Possibly we could take `ModuleKind` as an argument and append it to the path, so two different modules could use the same name (so it would be `<project-root>/$DB_MIGRATION_DIR/<module-kind>/<snapshot-name>`.

dpc (dismissed):

elsirion (dismissed):

elsirion:
Needs a minor rebase

elsirion (approved):

dpc (approved):

PR #1939, opened 2023-03-15
fix(cli): do not pay a paid invoice

### LINKED ISSUE #608: Mint client will try to pay a paid invoice
```
% cargo run --release --bin mint-client-cli federations/testfed ln-pay <invoice>
{
  "ln_pay": {
    "contract_id": "c6bf42c2bd0afe5c9c51a2a179cb2494f2ef11afcad88974c14b916d478ad893"
  }
}

% cargo run --release --bin mint-client-cli federations/testfed info 
{
  "info": {
    "total_amount": 929320,
    "total_num_notes": 25,
    "details": {
      "10": 2,
      "10000": 2,
      "100": 3,
      "1000": 9,
      "100000": 9
    }
  }
}

% cargo run --release --bin mint-client-cli federations/testfed ln-pay <invoice>
{
  "ln_pay": {
    "contract_id": "5d865102339f51aaba1a08a8109413589b3806463186b229398f07b330f47d27"
  }
}

% cargo run --release --bin mint-client-cli federations/testfed info 
{
  "info": {
    "total_amount": 919220,
    "total_num_notes": 23,
    "details": {
      "10000": 1,
      "100000": 9,
      "100": 2,
      "10": 2,
      "1000": 9
    }
  }
}
```

jp1ac4:
The aim of this PR is to fix #608.

@justinmoon As you suggested in https://github.com/fedimint/fedimint/issues/608#issuecomment-1464758957, I've attempted to store the payment hash of paid invoices and refuse attempts to pay an invoice that has already been paid. I'm not sure if this is in the right place.

I may also need to allow for the case where something goes wrong further downstream when paying an invoice, such as in `await_outgoing_contract_acceptance` or `await_outgoing_contract_execution`, so that those parts can still be retried.

If you could take a look and let me know your thoughts, that would be great thanks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1296,6 +1296,7 @@ dependencies = [
  "fedimint-wallet-server",
  "futures",
  "hex",
+ "lightning-invoice",
  "serde",
  "serde_json",
  "strum",
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -781,7 +781,7 @@ impl FedimintCli {
             Command::LnPay { bolt11 } => {
                 let client = cli.build_client(&self.module_gens).await?;
                 let (contract_id, outpoint) = client
-                    .fund_outgoing_ln_contract(bolt11, &mut rng)
+                    .fund_outgoing_ln_contract(bolt11.clone(), &mut rng)
                     .await
                     .map_err_cli_msg(
                         CliErrorKind::GeneralFederationError,
@@ -791,8 +791,9 @@ impl FedimintCli {
                     .await_outgoing_contract_acceptance(outpoint)
                     .await
                     .map_err_cli_msg(CliErrorKind::Timeout, "contract wasn't accepted in time")?;
+                client.save_outgoing_contract_pending(bolt11.clone()).await;
                 client
-                    .await_outgoing_contract_execution(contract_id, &mut rng)
+                    .await_outgoing_contract_execution(bolt11, contract_id, &mut rng)
                     .await
                     .map(|_| CliOutput::LnPay {
                         contract_id: (contract_id),
--- a/fedimint-client-legacy/src/lib.rs
+++ b/fedimint-client-legacy/src/lib.rs
@@ -73,8 +73,8 @@ use url::Url;
 
 use crate::db::ClientSecretKey;
 use crate::ln::db::{
-    OutgoingContractAccountKey, OutgoingContractAccountKeyPrefix, OutgoingPaymentClaimKey,
-    OutgoingPaymentClaimKeyPrefix, OutgoingPaymentKey,
+    OutgoingContractAccountKey, OutgoingContractAccountKeyPrefix, OutgoingContractPendingKey,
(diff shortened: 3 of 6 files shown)


client/client-lib/src/ln/outgoing.rs
```diff
@@ -35,3 +36,18 @@ impl OutgoingContractAccount {
...
+}
+
+impl Serialize for OutgoingPaidInvoice {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.invoice.to_string().as_str())
+    }
+}
```
justinmoon, line 53: Could we not save the invoice directly? Because it doesn't implement `Serialize` or something?

justinmoon:
Looks good so far

> I may also need to allow for the case where something goes wrong further downstream when paying an invoice, such as in `await_outgoing_contract_acceptance` or `await_outgoing_contract_execution`, so that those parts can still be retried.

Yes, I think we should save to the database **after** `await_outgoing_contract_acceptance`. Then we should delete that key if `try_refund_outgoing_contract` succeeds (this happens when the gateway fails to make the payment). This would allow the user to try paying the invoice again with another gateway.

client/client-lib/src/ln/db.rs
```diff
@@ -104,3 +106,19 @@ impl_db_lookup!(
     key = LightningGatewayKey,
     query_prefix = LightningGatewayKeyPrefix
 );
+
+#[derive(Debug, Encodable, Decodable, Serialize)]
+pub struct OutgoingPaidInvoiceKey(pub Sha256Hash);
```
justinmoon, line 111: Maybe call it something like `OutgoingContractPendingKeyPrefix`? Because at this point the lightning invoice isn't paid yet because the gateway could still fail to pay it ...

client/client-lib/src/ln/outgoing.rs
jp1ac4: I didn't try that before as I was following the `ConfirmedInvoice` approach, but I've just tried it now and it seems to work so will make this change.

client/client-lib/src/ln/db.rs
jp1ac4: Sure, I'll rename that. Should I change `Sha256Hash` to `ContractId` in that case?

jp1ac4:
Thanks for your comments, @justinmoon.

> Yes, I think we should save to the database **after** `await_outgoing_contract_acceptance`. Then we should delete that key if `try_refund_outgoing_contract` succeeds (this happens when the gateway fails to make the payment). This would allow the user to try paying the invoice again with another gateway.

I'll make this change and will let you know if I have any questions.

client/client-lib/src/ln/outgoing.rs
jp1ac4: This has now been updated so that the invoice is saved directly.

jp1ac4:
As per https://github.com/fedimint/fedimint/issues/608#issuecomment-1518386201, this is no longer required.

PR #1940, opened 2023-03-15
chore: rename `aead` *lib* to `fedimint_aead`

dpc:
We previously rename the *crate*, but forgot about renaming the library inside it.


mxz42 (approved):
makes sense

jkitman (approved):

PR #1941, opened 2023-03-15
chore(nix): install `nil` Nix LSP server

dpc:
I'm using it installed locally, but I think everyone should enjoy it.


mxz42 (approved):
yeah, why not. works with helix by default :)

jkitman (approved):

PR #1945, opened 2023-03-16
fix: incorrect `ts` used

dpc:
After refactoring crane build into separate file, some parts of the code picked `ts` from `with pkgs`, and not the wrapper that we want to use.


mxz42 (approved):
looks good. tmuxinator works too.

jkitman:
Merged in #1946

PR #1946, opened 2023-03-16
chore(ci): run tests in parallel more efficiently

dpc:
Instead of running in parallel bunch of `nix build -L ...`, where
each `nix build` invocation will have to load the previously built
`./target` (gigabytes) individually, run one `nix build` that runs
`parallel` of each test.

The major difficulty here is sandboxing each tests (network namespace).
Running in `nix build` was giving us sandboxing for free. Now
I have to resort to weird `unshare ... unshare` invocation that seems
to work, but is not best. It also won't work on Mac, but we mostly
care about CI here anyway.

### DIFF
--- a/scripts/test-ci-all.sh
+++ b/scripts/test-ci-all.sh
@@ -13,35 +13,34 @@ fi
 
 
 # Avoid re-building workspace in parallel in all test derivations
->&2 echo "### Making sure workspace is built..."
-nix build -L .#debug.workspaceBuild 2>&1 | ts -s
-
+# Note: Respect 'CARGO_PROFILE' that crane uses
+cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}}
 
 function cli_test_reconnect() {
   set -eo pipefail # pipefail must be set manually again
   echo "### Starting reconnect test..."
-  nix build -L .#debug.cli-test.reconnect 2>&1 | ts -s
+  unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/reconnect-test.sh" 2>&1 | ts -s
 }
 export -f cli_test_reconnect
 
 function cli_test_latency() {
   set -eo pipefail # pipefail must be set manually again
   echo "### Starting latency test..."
-  nix build -L .#debug.cli-test.latency 2>&1 | ts -s
+  unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/latency-test.sh" 2>&1 | ts -s
 }
 export -f cli_test_latency
 
 function cli_test_cli() {
   set -eo pipefail # pipefail must be set manually again
   echo "### Starting cli test..."
-  nix build -L .#debug.cli-test.cli 2>&1 | ts -s
+  unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/cli-test.sh" 2>&1 | ts -s
 }
 export -f cli_test_cli
 
 function cli_test_rust_tests() {
   set -eo pipefail # pipefail must be set manually again
   echo "### Starting integration test..."
-  nix build -L .#debug.cli-test.rust-tests 2>&1 | ts -s
+  unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/rust-tests.sh" 2>&1 | ts -s
 }
 export -f cli_test_rust_tests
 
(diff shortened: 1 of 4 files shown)


okjodom (approved):

scripts/test-ci-all.sh
```diff
@@ -13,43 +13,42 @@ fi
...
-nix build -L .#debug.workspaceBuild 2>&1 | ts -s
-
+# Note: Respect 'CARGO_PROFILE' that crane uses
+cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}}
 
 function cli_test_reconnect() {
   set -eo pipefail # pipefail must be set manually again
   echo "### Starting reconnect test..."
-  nix build -L .#debug.cli-test.reconnect 2>&1 | ts -s
+  unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/reconnect-test.sh" 2>&1 | ts -s
```
mxz42, line 22: so this creates a namespace with uid and guid mapped to 0 (-r), then creates a nework namespace (-n), sets the lo interface up and creates a new user namespace and runs the test script in it. cool.

mxz42 (commented):
I'm not sure if it works for me.

```sh
(pr/1946 %) | just test-ci-all
./scripts/test-ci-all.sh
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
### Starting all tests in parallel...
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
### Starting always_fail test...
00:00:00 unshare: failed to execute ./scripts/always-fail.sh: No such file or directory
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```

(1) the scripts `scripts/always-fail.sh` is there (exists) but apparently it's not found in the namespace? do we need a different path? 

(2) AFter the command is running for about 10min, I get a lot of output at once. Some test output and at the end this:

```sh
00:10:01 +++ lncli -n regtest --lnddir=/tmp/nix-shell.g5zXVN/fm-xZ6p/lnd --rpcserver=localhost:11009 getinfo
00:10:01 +++ jq -r .synced_to_chain
00:10:01 [lncli] could not load global options: could not load TLS cert file: open /tmp/nix-shell.g5zXVN/fm-xZ6p/lnd/tls.cert: no such file or directory
00:10:01 ++ '[' true == '' ']'
00:10:01 ++ echo sleeping
00:10:01 sleeping
00:10:01 ++ sleep 1
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
error: Recipe `test-ci-all` failed on line 33 with exit code 4

```

jkitman:
Whoops I only checked CI works, I haven't been using `just` to run the tests locally.

dpc:
> (1) the scripts `scripts/always-fail.sh` is there (exists) but apparently it's not found in the namespace? do we need a different path?

https://github.com/fedimint/fedimint/pull/1951 @mxz42

dpc:
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```

This looks like your local system misconfiguration(?). Are you running in `nix develop` or `direnv`?

> (2) AFter the command is running for about 10min, I get a lot of output at once. Some test output and at the end this

Could it be your system can't keep up with running the tests? Also - there's an option there that will not start new tests until loadavg is lower than 150, so you might want to not run anything else in the background.

jkitman:
> This looks like your local system misconfiguration
I see the same warning on my machine in `nix develop`

dpc:
Are you maybe using zsh?

See if any of https://stackoverflow.com/questions/2499794/how-to-fix-a-locale-setting-warning-from-perl helps.

`env LANG=C ./script/testi-ci-all.sh` might be the "easiest" solution.

jkitman:
> Do you have it set?

Yes, it's set.

> env LANG=C ./script/testi-ci-all.sh might be the "easiest" solution.

That works for me!

dpc:
> That works for me!

OK then. Let's just put it there then. https://github.com/fedimint/fedimint/pull/1954

mxz42:
> If I run `nix develop --ignore-environment` the 
> ```
> dpc@mutex:~/lab/fedimint.git/main$ echo $LANG
> en_US.UTF-8
> ```
> 
> Do you have it set?

yes. I'm also in a nix develop shell.

> perl: warning: Falling back to the standard locale ("C").

I just haven't bothered yet to look into it as it uses the fallback. It was just annoying that it prints the messages.


>> env LANG=C ./script/testi-ci-all.sh might be the "easiest" solution.
>
>That works for me!

for me as well.

PR #1947, opened 2023-03-16
fix: in scripts use `set -euo`, not `set -euox`

dpc:
`set -x` produces waay too much output.

### DIFF
--- a/scripts/cli-test.sh
+++ b/scripts/cli-test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # Runs a CLI-based integration test
 
-set -euxo pipefail
+set -euo pipefail
 export RUST_LOG="${RUST_LOG:-info}"
 
 export PEG_IN_AMOUNT=10000
--- a/scripts/latency-test.sh
+++ b/scripts/latency-test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # Runs a test to determine the latency of certain user actions
 
-set -euxo pipefail
+set -euo pipefail
 FM_FED_SIZE=${1:-4}
 ITERATIONS=${2:-10}
 export RUST_LOG="${RUST_LOG:-info}"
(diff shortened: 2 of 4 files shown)


dpc:
@justinmoon :+1: / :-1: :question:

elsirion:
I really like knowing which command produced which output, maybe we can [make it optional and only activate it in debug mode](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging)?

dpc:
> I really like knowing which command produced which output

For the scripts itself? `-x` prints lots of extra stuff. Every `lightning-cli`, `bitcoin-cli`, etc., the responses from (long jsons) etc.

Anyway - if we want it, we can keep them. I don't think we did `-x` everywhere previously, so I thought introducing it in all script was a mistake.

`ACTIONS_STEP_DEBUG` seems like debugging the github action runner itself.

elsirion:
> `ACTIONS_STEP_DEBUG` seems like debugging the github action runner itself.

Maybe I got that wrong, when you re-run GH actions allows to re-run ind debug mode, I thought we could read that env var and activate the extended logging with that.

> I don't think we did `-x` everywhere previously, so I thought introducing it in all script was a mistake.

We did have it everywhere in the beginning (I often include it in my scripts to know what's going on), maybe it was removed in the meantime and now re-introduced? Anyway, we can remove it if you want (at least we come across a bug where nobody knows where a log line might have come from and we re-introduce it :stuck_out_tongue:).

elsirion (approved):

jkitman (approved):
I also use it for debugging the tests often, but it's easy enough to add back I suppose.

dpc:
> I also use it for debugging the tests often, but it's easy enough to add back I suppose.

Yeah. The question is - do you want to see it every time, or are you OK adding `x` manually when you are debugging the bash scripts itself.

PR #1951, opened 2023-03-16
fix: location of "always fail" test

dpc:


jkitman:
Wonder why this works in CI?

dpc:
> Wonder why this works in CI?

Because we test if this one fails. And with a wrong name it does. Just for the wrong reasons.

mxz42 (approved):

PR #1959, opened 2023-03-16
Refactor ApiEndpointContext out for use in api_endpoint!

jkitman:
Refactors `ApiEndpointContext` out so we can pass state to all API endpoints.

### DIFF
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -132,6 +132,26 @@ impl ApiError {
     }
 }
 
+/// State made available to all API endpoints for handling a request
+pub struct ApiEndpointContext<'a> {
+    dbtx: ModuleDatabaseTransaction<'a, ModuleInstanceId>,
+
+    has_auth: bool,
+}
+
+impl<'a> ApiEndpointContext<'a> {
+    /// Database tx handle, will be committed
+    pub fn dbtx(&mut self) -> &mut ModuleDatabaseTransaction<'a, ModuleInstanceId> {
+        &mut self.dbtx
+    }
+
+    /// Whether the request was authenticated as the guardian who controls this
+    /// fedimint server
+    pub fn has_auth(&self) -> bool {
+        self.has_auth
+    }
+}
+
 #[apply(async_trait_maybe_send!)]
 pub trait TypedApiEndpoint {
     type State: Sync;
@@ -144,7 +164,7 @@ pub trait TypedApiEndpoint {
 
     async fn handle<'a, 'b>(
         state: &'a Self::State,
-        dbtx: &'a mut fedimint_core::db::ModuleDatabaseTransaction<'b, ModuleInstanceId>,
+        context: &'a mut ApiEndpointContext<'b>,
         request: Self::Param,
         has_auth: bool,
     ) -> Result<Self::Response, ApiError>;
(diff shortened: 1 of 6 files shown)


fedimint-core/src/module/mod.rs
```diff
@@ -144,7 +153,7 @@ pub trait TypedApiEndpoint {
 
     async fn handle<'a, 'b>(
         state: &'a Self::State,
-        dbtx: &'a mut fedimint_core::db::ModuleDatabaseTransaction<'b, ModuleInstanceId>,
+        dbtx: &'a mut ApiEndpointContext<'b>,
```
Maan2003, line 156:
```suggestion
        ctx: &'a mut ApiEndpointContext<'b>,
```

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -753,18 +753,18 @@ impl ServerModule for Lightning {
         vec![
             api_endpoint! {
                 "/account",
-                async |module: &Lightning, dbtx, contract_id: ContractId| -> ContractAccount {
+                async |module: &Lightning, context, contract_id: ContractId| -> ContractAccount {
```
Maan2003, line 756: lets use `ctx` instead of `context`

elsirion:
Looks good, but agree with @Maan2003

modules/fedimint-ln-server/src/lib.rs
jkitman: What's the advantage of abbreviating?  It's not that long a name and I couldn't easily guess `ctx = context`.

Maan2003: fair enough, I am biased. when I worked in rust gui, we called it ctx or even cx, but gui has way too many contexts.

fedimint-core/src/module/mod.rs
```diff
@@ -132,6 +132,26 @@ impl ApiError {
     }
 }
 
+/// State made available to all API endpoints for handling a request
+pub struct ApiEndpointContext<'a> {
+    dbtx: ModuleDatabaseTransaction<'a, ModuleInstanceId>,
+
```
elsirion, line 138: nit: useless newline

elsirion (approved):

fedimint-core/src/module/mod.rs
```diff
@@ -186,11 +205,10 @@ macro_rules! __api_endpoint {
 
             async fn handle<'a, 'b>(
                 $state: &'a Self::State,
-                $dbtx: &'a mut fedimint_core::db::ModuleDatabaseTransaction<'b, ModuleInstanceId>,
+                $context: &'a mut fedimint_core::module::ApiEndpointContext<'b>,
```
Maan2003, line 208: not introduced by this PR, but this should be `$crate::module::ApiEndpointContext`

Maan2003 (approved):

fedimint-core/src/module/mod.rs
jkitman: Why is that?

Maan2003: $crate:: is more correct than fedimint_core, it even works in case fedimint_core is not directly imported by the crate using the macro (reexported by some other crate)

Issue #1952, opened 2023-03-16
Daily build fails


elsirion:
Both `cargo udeps` and `cargo audit` fail in the [daily check](https://github.com/fedimint/fedimint/actions/runs/4433903728). I noticed it due to #1898 when trying to build all outputs of our flake.

TODO:
- [ ] remove unused deps #1962
- [x] update deps with vulnerabilities

justinmoon:
I asked @GetPsyched to look into this

PR #1962, opened 2023-03-16
fix: `just udeps` failing

dpc:


elsirion (approved):
Partially fixes #1952

elsirion:
I didn't notice this at first, but when checking out the PR I get changes in `Cargo.lock` after running `cargo check`, this would happen in every PR building on top of this one.

Issue #1956, opened 2023-03-16
Allow overwritting modules to be instantiated during dkg


dpc:
Currently when setting up the federation the following code:

```
    /// Return legacy initialization order. See [`LegacyInitOrderIter`].
    pub fn legacy_init_order_iter(&self) -> LegacyInitOrderIter<M>
    where
        M: Clone,
    {
        for hardcoded_module in ["mint", "ln", "wallet"] {
            if !self
                .0
                .contains_key(&ModuleKind::from_static_str(hardcoded_module))
            {
                panic!("Missing {hardcoded_module} module");
            }
        }

        LegacyInitOrderIter {
            next_id: 0,
            rest: self.0.clone(),
        }
    }
```

makes it so that `"mint", "ln", "wallet"` will the 3 module instances. For users of their own modules there is no way currently than changing this code to have their own modules.

We should allow passing some `--extra-module-instance <kind> --extra-module-instance <kind>` in the command line that gets passed as `Vec<ModuleKind>` and chained to the `legacy_init_order_iter` to make things more flexible, at least until the UI allows setting up extra modules.

ayush-ranjan-official:
please assign me, I want to work on this issue.

ayush-ranjan-official:
How can I allow multiple line parsing such as `--extra-module-instance <kind> --extra-module-instance <kind>` using clap. please help.

m1sterc001guy:
> For users of their own modules there is no way currently than changing this code to have their own modules.

For now, won't users still need to change the code to include the actual code for the module (DKG and the module business logic)? Are we planning on supporting some sort of dynamic linking in the future?

elsirion:
Sure, but afaik all the CLI stuff lives in our repo and is just imported when you build your custom `distributedgen` (where you attach all your modules):

https://github.com/fedimint/fedimint/blob/master/fedimintd/src/bin/distributedgen.rs#L1-L6

justinmoon:
call: maybe we should migrate default modules to these mechanics

dpc:
I was mistaken and `LegacyInitOrderIter` actually implements returning all extra modules, so I think we don't have to touch this one for time being.

PR #1971, opened 2023-03-17
 Allowing overwritting modules to be instantiated during dkg 

ayush-ranjan-official:
Fixes issue: #1956
using `std::env` I passed  some `--extra-module-instance <kind>` in the command line that gets passed as `args: vec<ModuleKind>` and eventually gets chained to the `legacy_init_order_iter` with the initial 3 module instances
which are `["mint", "ln", "wallet"]`.
I have also used `if-else` statements to support multiple user generated modules.
I tried to fix the issue with my best understanding of the problem, do correct me if I am missing something as I am relatively new to rust programming. 
Thanks.

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -4,6 +4,7 @@ use std::hash::Hash;
 use std::ops::Mul;
 use std::path::Path;
 use std::str::FromStr;
+use std::env;
 
 use anyhow::{bail, format_err};
 use bitcoin::secp256k1;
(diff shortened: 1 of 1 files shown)


fedimint-core/src/config.rs
```diff
@@ -321,7 +322,35 @@ impl<M> ModuleGenRegistry<M> {
     where
         M: Clone,
     {
-        for hardcoded_module in ["mint", "ln", "wallet"] {
+        let args: Vec<String> = env::args().collect();
```
dpc, line 325: We should parse these via `clap`, and pass here as an argument.

fedimint-core/src/config.rs
```diff
@@ -321,7 +322,35 @@ impl<M> ModuleGenRegistry<M> {
...
+                i += 1; // Move to the next argument.
+            }
+        }
+
+        // Chain the modules instances to the legacy_init_order_iter.
+        let mut legacy_init_order_iter = Vec::new();
+        // fixed 3 module instances
+        let mut initial_module: Vec<String> = ["mint", "ln", "wallet"];
+        //adding all module instances to legacy_init_order_iter.
+        legacy_init_order_iter.append(&mut initial_module);
```
dpc, line 349: `.chain(extra_modules)` instead.

fedimint-core/src/config.rs
dpc: `fedimint-server` already takes care of passing the args, BTW. We can just stick it there, and then just need to carry it to here.

ayush-ranjan-official: Oh! Let me do that again. Thanks.

Issue #1955, opened 2023-03-16
Group clap args for configuring gatewayd with LND


okjodom:
See [`ArgGroup`](https://docs.rs/clap/latest/clap/struct.ArgGroup.html#)

PR #1973, opened 2023-03-18
refactor: use subcommand to group gatewayd parameters together

m1sterc001guy:
Refactor that simplifies the command line arguments, makes it harder to invoke `gatewayd` incorrectly.

Fixes: https://github.com/fedimint/fedimint/issues/1955

### DIFF
--- a/gateway/ln-gateway/src/bin/gatewayd.rs
+++ b/gateway/ln-gateway/src/bin/gatewayd.rs
@@ -45,7 +45,7 @@ pub struct GatewayOpts {
 
 #[derive(Debug, Clone, Subcommand)]
 enum Mode {
-    #[clap(name = "LND")]
+    #[clap(name = "lnd")]
     Lnd {
         /// LND RPC address
         #[arg(long = "lnd-rpc-host", env = "FM_LND_RPC_ADDR")]
@@ -59,10 +59,10 @@ enum Mode {
         #[arg(long = "lnd-macaroon", env = "FM_LND_MACAROON")]
         lnd_macaroon: String,
     },
-    #[clap(name = "CLN")]
+    #[clap(name = "cln")]
     Cln {
-        #[arg(long = "lnrpc-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
-        lnrpc_addr: Url,
+        #[arg(long = "cln-extension-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
+        cln_extension_addr: Url,
     },
 }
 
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -39,23 +41,29 @@ pub struct GatewayOpts {
...
+        lnd_tls_cert: String,
+
+        /// LND macaroon file path
+        #[arg(long = "lnd-macaroon", env = "FM_LND_MACAROON")]
+        lnd_macaroon: String,
+    },
+    #[clap(name = "CLN")]
+    Cln {
+        #[arg(long = "lnrpc-addr", env = "FM_GATEWAY_LIGHTNING_ADDR")]
+        lnrpc_addr: Url,
```
elsirion, line 65: Can we rename this to extension address or something like that? Otherwise it's easy to confuse with the unix socket CLN uses.

m1sterc001guy: Agreed, done.

elsirion (approved):

justinmoon (approved):

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -39,23 +41,29 @@ pub struct GatewayOpts {
...
+        lnd_rpc_addr: String,
+
+        /// LND TLS cert file path
+        #[arg(long = "lnd-tls-cert", env = "FM_LND_TLS_CERT")]
+        lnd_tls_cert: String,
+
+        /// LND macaroon file path
+        #[arg(long = "lnd-macaroon", env = "FM_LND_MACAROON")]
+        lnd_macaroon: String,
+    },
```
okjodom, line 61: Since we use the generic `NetworkLnRpcClient`, should we prefer `Lnd | Generic` modes?

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -39,23 +41,29 @@ pub struct GatewayOpts {
...
+    #[clap(name = "cln")]
```
elsirion, line 62: nit: I think this annotation is redundant now

elsirion (approved):

gateway/ln-gateway/src/bin/gatewayd.rs
elsirion: Can be a follow-up imo, right now there are only these 2 extensions.

PR #1975, opened 2023-03-19
chore: create decoder registry in cli

### LINKED ISSUE #1931: Properly create decoder registry in fedimint-cli
Currently we:

```
        let module_gens = ClientModuleGenRegistry::from(vec![
            DynClientModuleGen::from(WalletClientGen),
            DynClientModuleGen::from(MintClientGen),
            DynClientModuleGen::from(LightningClientGen),
        ]);

        let cli = Cli::parse();
        if let Command::JoinFederation { connect } = cli.command {
            let connect_obj: WsClientConnectInfo = WsClientConnectInfo::from_str(&connect)
                .map_err(Box::<dyn Error>::from)
                .or_terminate(CliErrorKind::InvalidValue, "invalid connect info");
            let api = Arc::new(WsFederationApi::from_urls(&connect_obj))
                as Arc<dyn IFederationApi + Send + Sync + 'static>;
            let cfg: ClientConfig = api
                .download_client_config(&connect_obj.id, module_gens.to_common())
                .await
                .or_terminate(
                    CliErrorKind::NetworkError,
                    "couldn't download config from peer",
                );
            let cfg_path = cli.workdir.join("client.json");
            std::fs::create_dir_all(&cli.workdir)
                .or_terminate(CliErrorKind::IOError, "failed to create config directory");
            let writer = std::fs::File::create(cfg_path)
                .or_terminate(CliErrorKind::IOError, "couldn't create config.json");
            serde_json::to_writer_pretty(writer, &cfg)
                .or_terminate(CliErrorKind::IOError, "couldn't write config");
            println!(
                "{}",
                &CliOutput::JoinFederation { joined: (connect) }.to_string()
            );
            return;
        };

        let cfg_path = cli.workdir.join("client.json");
        let db_path = cli.workdir.join("client.db");
        let cfg: UserClientConfig = load_from_file(&cfg_path).expect("Failed to parse config");
        let db = fedimint_rocksdb::RocksDb::open(db_path)
            .or_terminate(CliErrorKind::IOError, "could not open transaction db");
        let db = Database::new(db, module_decode_stubs());
```

So while we do have `module_gens` and we do fetch config, we still use `module_decode_stubs()` to create decoders.

Instead, we should:

* have `module_gens` ready (we do already)
* get and validate client config (we do already)
* construct decoder registry by iterating over modules from the config and for reach `module_instance_id` get the `kind` and call client module gen of that kind from `ClientModuleRegistry` to give us a decoder, then put all decoders into `DecoderRegistry`

ngutech21:
fixes #1931

gets rid of all module_decode_stubs() calls in fedimint-cli/main.rs

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -22,6 +22,7 @@ use fedimint_core::api::{
 };
 use fedimint_core::config::{load_from_file, ClientConfig, FederationId};
 use fedimint_core::db::{Database, DatabaseValue};
+use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::module::{ApiAuth, ApiRequestErased};
 use fedimint_core::query::EventuallyConsistent;
 use fedimint_core::task::TaskGroup;
@@ -37,7 +38,7 @@ use mint_client::utils::{
     from_hex, parse_bitcoin_amount, parse_ecash, parse_fedimint_amount, parse_node_pub_key,
     parse_peer_id, serialize_ecash,
 };
-use mint_client::{module_decode_stubs, Client, UserClientConfig};
+use mint_client::{Client, UserClientConfig};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 use url::Url;
@@ -479,11 +480,19 @@ async fn main() {
         let cfg: UserClientConfig = load_from_file(&cfg_path).expect("Failed to parse config");
         let db = fedimint_rocksdb::RocksDb::open(db_path)
             .or_terminate(CliErrorKind::IOError, "could not open transaction db");
-        let db = Database::new(db, module_decode_stubs());
+
+        let decoders = ModuleDecoderRegistry::new(cfg.clone().0.modules.into_iter().map(
+            |(id, module_cfg)| {
+                let module_gen = module_gens
+                    .get(module_cfg.kind())
+                    .expect("module kind not found in registry");
+                (id, module_gen.as_ref().decoder())
+            },
+        ));
+
+        let db = Database::new(db, decoders.clone());
 
         let rng = rand::rngs::OsRng;
-
-        let decoders = module_decode_stubs();
 
         let client = Client::new(cfg.clone(), decoders, module_gens, db, Default::default()).await;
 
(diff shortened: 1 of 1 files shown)


client/cli/src/main.rs
```diff
@@ -479,11 +480,19 @@ async fn main() {
...
             .or_terminate(CliErrorKind::IOError, "could not open transaction db");
-        let db = Database::new(db, module_decode_stubs());
 
-        let rng = rand::rngs::OsRng;
+        let decoders = ModuleDecoderRegistry::new(cfg.clone().0.modules.into_iter().map(
+            |(id, module_cfg)| {
+                let module_gen = module_gens
+                    .get(module_cfg.kind())
+                    .expect("module kind not found in registry");
+                (id, module_gen.to_dyn_common().decoder())
```
elsirion, line 489: I think you could get away with just getting a reference to the common module gen instead of cloning, I should have documented it better:

https://github.com/fedimint/fedimint/blob/d20cb9750d8be9d38f9570a120af6d6a130fbf7c/fedimint-core/src/module/mod.rs#L466-L470

The `to_dyn_common` is useful when you can't deal with references, e.g. need to build a whole common registry.

elsirion (approved):

client/cli/src/main.rs
```diff
@@ -479,11 +480,19 @@ async fn main() {
         let cfg: UserClientConfig = load_from_file(&cfg_path).expect("Failed to parse config");
         let db = fedimint_rocksdb::RocksDb::open(db_path)
             .or_terminate(CliErrorKind::IOError, "could not open transaction db");
-        let db = Database::new(db, module_decode_stubs());
 
-        let rng = rand::rngs::OsRng;
+        let decoders = ModuleDecoderRegistry::new(cfg.clone().0.modules.into_iter().map(
```
dpc, line 484: Please make a function. We're going to need it in multiple places.

client/cli/src/main.rs
```diff
@@ -479,11 +480,19 @@ async fn main() {
...
+            |(id, module_cfg)| {
+                let module_gen = module_gens
+                    .get(module_cfg.kind())
+                    .expect("module kind not found in registry");
```
dpc, line 488: `Result`s always preferred.

dpc (approved):

dpc:
@ngutech21 Better than previously, but if you could follow up and make it a reusable function, that would be perfect.

PR #1980, opened 2023-03-20
refactor: clean up ConfigGenParams, add ConfigGenParamsRegistry

dpc:
Introduce `Fedimintd::with_extra_module_gens_params` that allows
passing config gen params to use for any additional modules
registered with `Fedimintd::with_module`.

This is very provisional, only to unblock 3rd party module
out-of-tree testing for time being.

### DIFF
--- a/client/client-lib/src/ln/mod.rs
+++ b/client/client-lib/src/ln/mod.rs
@@ -435,7 +435,7 @@ mod tests {
             FakeFed::<Lightning>::new(
                 4,
                 |cfg, _db| async move { Ok(Lightning::new(cfg.to_typed()?)) },
-                &ConfigGenParams::new(),
+                &ConfigGenParams::null(),
                 &LightningGen,
                 module_id,
             )
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -246,6 +246,12 @@ pub struct ConfigGenParams(serde_json::Value);
 pub type ServerModuleGenRegistry = ModuleGenRegistry<DynServerModuleGen>;
 
 impl ConfigGenParams {
+    /// Null value, used as a config gen parameters for module gens that don't
+    /// need any parameters
+    pub fn null() -> Self {
+        Self(jsonrpsee_core::JsonValue::Null)
+    }
+
     pub fn to_typed<P: ModuleGenParams>(&self) -> anyhow::Result<P> {
         serde_json::from_value(self.0.clone())
             .map_err(|e| anyhow::Error::new(e).context("Invalid module params"))
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -229,37 +229,6 @@ impl ClientConfig {
     }
 }
 
-/// Parameters for generating all module configs
-///
-/// The same `ModuleKind` may have multiple instances with different settings
-#[derive(Debug, Clone, Default)]
-pub struct ConfigGenParams(BTreeMap<String, serde_json::Value>);
-
-impl ConfigGenParams {
-    pub fn new() -> ConfigGenParams {
-        ConfigGenParams::default()
-    }
-
-    /// Add params for a module
-    pub fn attach<P: ModuleGenParams>(mut self, module_params: P) -> Self {
-        self.0.insert(
-            P::MODULE_NAME.to_string(),
-            serde_json::to_value(&module_params).expect("Encoding to value doesn't fail"),
-        );
-        self
-    }
-
-    /// Retrieve a typed config generation parameters for a module
-    pub fn get<P: ModuleGenParams>(&self) -> anyhow::Result<P> {
-        let value = self
-            .0
-            .get(P::MODULE_NAME)
-            .ok_or_else(|| anyhow::anyhow!("No params found for module {}", P::MODULE_NAME))?;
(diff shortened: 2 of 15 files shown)


client/client-lib/src/ln/mod.rs
```diff
@@ -435,7 +435,7 @@ mod tests {
             FakeFed::<Lightning>::new(
                 4,
                 |cfg, _db| async move { Ok(Lightning::new(cfg.to_typed()?)) },
-                &ConfigGenParams::new(),
+                &ConfigGenParams::default(),
```
elsirion, line 438: Is this a general recommendation to use `fedault` over `new`?

fedimint-core/src/config.rs
```diff
@@ -269,10 +238,33 @@ impl<M> Default for ModuleGenRegistry<M> {
...
+}
+
 pub type CommonModuleGenRegistry = ModuleGenRegistry<DynCommonModuleGen>;
 
+/// Configs for each module's DKG
+///
+/// Note: in the future, we should make this one a
+/// ModuleRegistry<ConfigGenParams`, as each module **instance** will need a
+/// distinct config for dkg.
+pub type ConfigGenParamsRegistry = ModuleGenRegistry<ConfigGenParams>;
```
elsirion: I assume this would be somewhat involved rn, hence the note?

fedimintd/src/distributed_gen.rs
```diff
@@ -189,14 +189,21 @@ impl DistributedGen {
                 finality_delay,
                 password,
             } => {
+                let mut module_gens_params = ServerModuleGenParamsRegistry::default();
+                attach_default_module_gen_params(
+                    &mut module_gens_params,
+                    max_denomination,
+                    network,
+                    finality_delay,
+                );
```
elsirion, line 198: We probably want to get rid of this special casing eventually?

fedimint-core/src/config.rs
justinmoon: Also, maybe the `\`` should be a `<`?

fedimintd/src/distributed_gen.rs
dpc: Agreed.

justinmoon (dismissed):

fedimint-core/src/config.rs
```diff
@@ -269,10 +238,33 @@ impl<M> Default for ModuleGenRegistry<M> {
     }
 }
 
+/// Module **generation** (so passed to dkg, not to the module itself) config
+/// parameters
+#[derive(Debug, Clone, Default)]
+pub struct ConfigGenParams(serde_json::Value);
```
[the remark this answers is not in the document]
dpc: As things were currently, the use of this type was wierd, where the keys were used for module kind instead of a keys to multiple values.

We'll have to change how the details here work, but currently I'm only interested in an ability to inject config gen params for a given module kind from the outside to unblock 3rd party out-of-tree building.

@elsirion 
Possibly we should have a "ServerModuleGen::default_params() -> ConfigGenParams` and just allow overrides from the CLI in the form of `KIND.some.js.key = value` and in the UI, just display the `ConfigGenParams` whole, as a serialized json, and allow user to edit them if they wish. Seems like a simple but functional way to handling it all.

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -62,9 +62,7 @@ pub struct MintGenParams {
     pub mint_amounts: Vec<Amount>,
 }
 
-impl ModuleGenParams for MintGenParams {
-    const MODULE_NAME: &'static str = "mint";
-}
+impl ModuleGenParams for MintGenParams {}
```
justinmoon, line 65: this impl that doesn't do anything is a little weird

dpc: It does do something. It makes `MintGenParams` implement `ModuleGenParams` trait and thus ... special. :D

justinmoon (approved):

fedimint-core/src/config.rs
```diff
@@ -263,7 +263,7 @@ pub type CommonModuleGenRegistry = ModuleGenRegistry<DynCommonModuleGen>;
 /// Note: in the future, we should make this one a
 /// ModuleRegistry<ConfigGenParams`, as each module **instance** will need a
 /// distinct config for dkg.
-pub type ConfigGenParamsRegistry = ModuleGenRegistry<ConfigGenParams>;
+pub type ServerModuleGenParamsRegistry = ModuleGenRegistry<ConfigGenParams>;
```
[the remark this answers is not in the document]
jkitman: I'd say just go with `ModuleConfigGenParamsRegistry` for consistency or perhaps we want to rename all `ConfigGenParams` -> `ConfigParams`?

fedimint-core/src/config.rs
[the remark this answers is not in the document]
jkitman: > just display the `ConfigGenParams` whole, as a serialized json, and allow user to edit them if they wish

Yes I think that will be how we should handle it in the UI/API during config setup.  The DKG CLI will eventually just be making API calls, but could support downloading the current config, manually editing, and then uploading.

elsirion (approved):

Issue #163, opened 2022-06-21
Test network logic failure cases


elsirion:
Extend `MockNetwork` to simulate connections dropping or having high latency and implement tests using that. See https://github.com/fedimint/minimint/pull/145#discussion_r902518714.

matthiasdebernardini:
Are there any outstanding issues for this? #145  looks to be complete. Should close.

elsirion:
This issue was actually created as a result of #145 and related PRs that introduced `MockNetwork` but didn't fully exhaust the possibilities of enhancing our test suite using it. `MockNetwork` could be used  to simulate all kinds of network conditions (latency, throughput, interruptions, …) and we should do that eventually.

PR #1982, opened 2023-03-21
feat: test unreliable (mock) network

douglaz:
Addresses https://github.com/fedimint/fedimint/issues/163

Creates a `StreamReliability` configuration for `MockConnector` making connections randomly fail on read/write and be delayed (have latency).

Introduces these failures on `test_connect()` test.

### DIFF
--- a/fedimint-server/src/net/connect.rs
+++ b/fedimint-server/src/net/connect.rs
@@ -251,6 +251,7 @@ pub mod mock {
     use anyhow::Error;
     use fedimint_core::PeerId;
     use futures::{FutureExt, SinkExt, Stream, StreamExt};
+    use rand::Rng;
     use tokio::io::{
         AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt, DuplexStream, ReadHalf, WriteHalf,
     };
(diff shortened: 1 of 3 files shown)


elsirion (approved):
Love it! We might also want to use that for integration tests, maybe not on every PR since it will slow things down, but as a manual test it could give insights into worst case latency if guardians were distributed around the world or using Tor.

fedimint-server/src/net/connect.rs
```diff
@@ -261,13 +262,167 @@ pub mod mock {
...
+        ) -> std::task::Poll<Result<(), std::io::Error>> {
+            match self.shutdown_generator.as_mut().map(|g| g.generate(cx)) {
+                Some(std::task::Poll::Ready(Err(e))) => std::task::Poll::Ready(Err(e)),
+                Some(std::task::Poll::Pending) => std::task::Poll::Pending,
+                Some(std::task::Poll::Ready(Ok(()))) | None => {
+                    Pin::new(&mut self.inner).poll_shutdown(cx)
+                }
+            }
+        }
+    }
```
elsirion, line 416: Wondering if there's a way to not build these futures manually. I'm ok with it in tests, but even a false positive test failure because of a broken futures impl could waste a few hours tracking down.

douglaz:
> What happens if you try to run the integration tests with unreliable connections?

With a low latency/failure rate it works. With higher values it starts to take much longer. I didn't explore it enough to get a configuration that fails just the right amount to be feasible to run it on CI (as it requires a bit of trial and error pushing things here). 

I think we can create a new task to address different testing profiles for different circumstances. Perhaps even collecting metrics to better understand how different latencies/failure rates impact on performance.

jkitman (approved):

PR #1988, opened 2023-03-21
Consumer devex improvements [WIP]

litch:
It looks like the distributedgen function being invoked does not tolerate the bitcoinrpc string anymore.

I'll try to make this a nicely packaged PR after I'm more comfortable with how this whole thing starts up, etc.

### DIFF
--- a/scripts/start-containers.sh
+++ b/scripts/start-containers.sh
@@ -31,7 +31,7 @@ function run_dkg() {
         api_port=$(echo "$BASE_PORT + $ID * 10 + 1" | bc -l)
         echo "  server-$ID:" >> $3
         echo "    image: $2" >> $3
-        echo "    command: distributedgen run --bind-p2p 0.0.0.0:$fed_port --bind-api 0.0.0.0:$api_port --out-dir /var/fedimint --certs $CERTS --bitcoind-rpc http://bitcoind:18443" >> $3
+        echo "    command: distributedgen run --bind-p2p 0.0.0.0:$fed_port --bind-api 0.0.0.0:$api_port --out-dir /var/fedimint --certs $CERTS" >> $3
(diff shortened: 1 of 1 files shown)


dpc (approved):

elsirion (changes requested):
Tested it and it partially works. DJG runs now, but `fedimintd` can't connect to `bitcoind` once started.

When we removed `bitcoind-rpc` as a config field to be given to DKG we added it as an environment variable expected by `fedimintd`. If the variable is not present it will default to `http://127.0.0.1:8332`, which is invalid for regtest. For this to work you need to set `FM_BITCOIND_RPC=http://bitcoind:18443`

elsirion:
Actually, even with my proposed fix it doesn't work for me. Not sure if it's docker weirdness on my end or something more general. You can check with `docker logs --follow fed-server-0-1` if there is connectivity to `bitcoind`. If not the logs will looks as follows:

```
2023-03-22T09:33:57.285659Z  WARN fedimint_bitcoind::bitcoincore_rpc: bitcoind returned error on cmd 'getblockchaininfo': JSON-RPC error: transport error: Didn't receive response data in time, timed out. addr="http://bitcoind:18443"
2023-03-22T09:33:57.285708Z ERROR fedimint_bitcoind::bitcoincore_rpc: error=JSON-RPC error: transport error: Didn't receive response data in time, timed out.
2023-03-22T09:33:57.285720Z  INFO fedimint_bitcoind: Bitcoind error JSON-RPC error: transport error: Didn't receive response data in time, timed out., retrying
```

litch:
Yes I got the error message to move, and was getting a bit of guidance from @dpc on next steps.  I'll make sure it works

justinmoon:
this file has been deleted

PR #2004, opened 2023-03-22
fix: can't use `fedimint-cli decode-connect-info` without `--workdir`

dpc:
Refactor `fedimint-cli` to get rid of the `--workdir` hacks.

The root of the problem is that we don't know if we need a `Client` (and thus `--workdir`) before we know which sub-command we are handling. So lets accept that fact, and just build the client dynamically when we need it, instead of trying to maintain a list of places where we actually don't need it, etc.

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -4,7 +4,7 @@ use std::error::Error;
 use std::fmt::Debug;
 use std::fs;
 use std::io::Write;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::str::FromStr;
 use std::sync::Arc;
@@ -42,6 +42,15 @@ use mint_client::{Client, UserClientConfig};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 use url::Url;
+
+/// List of modules supported by this cli
+fn get_module_gen_registry() -> ClientModuleGenRegistry {
+    ClientModuleGenRegistry::from(vec![
+        DynClientModuleGen::from(WalletClientGen),
+        DynClientModuleGen::from(MintClientGen),
+        DynClientModuleGen::from(LightningClientGen),
+    ])
+}
 
 #[derive(Serialize)]
 #[serde(rename_all(serialize = "snake_case"))]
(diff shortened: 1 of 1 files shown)


client/cli/src/main.rs
```diff
@@ -387,14 +415,19 @@ enum Command {
...
 }
 
-trait ErrorHandler<T, E> {
+/// Extension trait for values that either return `T` or terminate the program
+/// with CLI error
+trait OrTerminate<T> {
     fn or_terminate(self, err: CliErrorKind, msg: &str) -> T;
+}
+
+trait ErrorHandler<T, E>: OrTerminate<T> {
```
dpc, line 424: I kind of don't understand the meaning of this method, there is no comments explaining the abstraction that it's trying to create, so I moved the part that I can make sense of, and left `transform` as was.

client/cli/src/main.rs
```diff
@@ -428,99 +477,51 @@ struct PayRequest {
...
+        }
+    }
+}
+
+async fn handle_command(cli: Opts) -> CliResult {
+    let mut task_group = TaskGroup::new();
+    let mut rng = rand::rngs::OsRng;
+
+    match cli.command.clone() {
+        Command::JoinFederation { connect } => {
```
dpc, line 499: Nothing special about this sub-command, so it's now handled like everything else.

client/cli/src/main.rs
```diff
@@ -43,6 +43,15 @@ use serde::{Deserialize, Serialize};
...
 
+/// List of modules supported by this cli
+fn get_module_gen_registry() -> ClientModuleGenRegistry {
+    ClientModuleGenRegistry::from(vec![
+        DynClientModuleGen::from(WalletClientGen),
+        DynClientModuleGen::from(MintClientGen),
+        DynClientModuleGen::from(LightningClientGen),
+    ])
+}
+
```
elsirion, line 54: After client NG lands we'll probably want to build `fedimint-cli` in a similar way to `fedimintd` with attaching modules?

client/cli/src/main.rs
```diff
@@ -228,27 +237,46 @@ impl Error for CliError {}
...
-#[command(version)]
-struct CliNoWorkdir {
-    #[clap(subcommand)]
-    command: CommandNoWorkdir,
-}
+impl Opts {
+    fn workdir(&self) -> &Path {
+        self.workdir
+            .as_ref()
+            .or_terminate(CliErrorKind::IOError, "`--workdir=` argument not set.")
```
elsirion, line 253: nit: I think there are ways in clap to model dependencies between arguments/argument groups that would allow using clap's built-in help system. Not for you but a good first issue I'll create after merging.

client/cli/src/main.rs
elsirion: Seems to be from @your-sudden-death, maybe he can elaborate.

elsirion (approved):

client/cli/src/main.rs
dpc: We were discussing with @justinmoon that we could have `fedimint-cli module <id/kind> <args>` and module itself would handle commands (`<args>`) to it.

justinmoon: Yea. Ideally modules could define some CLI commands they would like to register to a `fedimint-cli` that includes as many modules as it likes ...

justinmoon (dismissed):

justinmoon (approved):

jkitman (approved):

client/cli/src/main.rs
elsirion: I *think* if we split the commands that need a work dir and the ones who don't into two enums we could do it in a non-brittle way, but I agree this might be overdoing it.

client/cli/src/main.rs
elsirion: Do we want to make `fedimint-cli` such a first-class citizen? Right now it's this hacky tool exposing whatever functionality CI needs.

client/cli/src/main.rs
dpc: We were doing just that, and someone put `decode-connect-info` in the wrong one. I don't blame them - my point is exactly that it's additional complication that is easy to forget about, just to display the exactly same error, just by a different part of code (clap vs our code).

client/cli/src/main.rs
justinmoon: I think we want a "`fedimint-cli`" which is a first-class citizen, but might require something close to a rewrite once we have the new client?

PR #2005, opened 2023-03-23
feat: run integration tests under unreliable network conditions

douglaz:
Also makes the reconnection delay a parameter with low default values on tests to avoid exponential growth in test duration when simulated network failures happen.

Closes https://github.com/fedimint/fedimint/issues/163 by completing the work started on https://github.com/fedimint/fedimint/pull/1982

### DIFF
--- a/fedimint-server/src/net/connect.rs
+++ b/fedimint-server/src/net/connect.rs
@@ -322,6 +322,7 @@ pub mod mock {
         latency: LatencyInterval,
         failure_rate: FailureRate,
         sleep_future: Option<Pin<Box<tokio::time::Sleep>>>,
+        successes: u64,
     }
 
     impl UnreliabilityGenerator {
@@ -330,6 +331,7 @@ pub mod mock {
                 latency,
                 failure_rate,
                 sleep_future: None,
+                successes: 0,
             }
         }
 
@@ -347,13 +349,18 @@ pub mod mock {
                 std::task::Poll::Pending => return std::task::Poll::Pending,
             }
             if self.failure_rate.random_fail() {
-                tracing::debug!("Returning random error on unreliable stream");
-                return std::task::Poll::Ready(Err(std::io::Error::new(
+                tracing::debug!(
+                    "Returning random error on unreliable stream after {} successes",
+                    self.successes
+                );
+                std::task::Poll::Ready(Err(std::io::Error::new(
                     std::io::ErrorKind::Other,
                     "Randomly failed",
-                )));
+                )))
+            } else {
+                self.successes += 1;
+                std::task::Poll::Ready(Ok(()))
             }
-            std::task::Poll::Ready(Ok(()))
         }
     }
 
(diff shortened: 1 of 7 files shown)


fedimint-server/src/net/peers.rs
```diff
@@ -81,12 +81,46 @@ struct PeerConnectionStateMachine<M> {
     state: PeerConnectionState<M>,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct DelayCalculator {
+    scaling_factor: f64,
+}
+
```
elsirion, line 88: nit: Could use some docs.

Really nice abstraction, eventually we might even want to expose it in the config for people to choose reconnect strategies and timeouts.

douglaz: Added minimal documentation

fedimint-server/src/net/connect.rs
```diff
@@ -548,7 +579,7 @@ pub mod mock {
                 let (stream_our, stream_theirs) = tokio::io::duplex(43_689);
                 let mut stream_our = UnreliableDuplexStream::new(stream_our, self.reliability);
                 let stream_theirs = UnreliableDuplexStream::new(stream_theirs, self.reliability);
-                client.send(stream_theirs).await.unwrap();
```
douglaz, line 551: This is a existing error on  `net::peers::tests::test_connect` that could be triggered randomly

fedimint-server/src/net/connect.rs
```diff
@@ -523,6 +530,30 @@ pub mod mock {
             }
         };
 
+        pub const INTEGRATION_TEST: StreamReliability = {
+            // Based on empirical testing: creates errors without causing tests to take
```
dpc, line 534: This part is tricky, as changes to the code down the line might invalidate this heuristic.

In the past for similar reasons I used adaptive failure rates. Eg. time decaying failure rates, or linearly increasing with time, but geometrically slashed on each failure (kind of like TCP: https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease) .

fedimint-server/src/net/peers.rs
```diff
@@ -81,12 +81,47 @@ struct PeerConnectionStateMachine<M> {
     state: PeerConnectionState<M>,
 }
 
+/// Calculates delays for reconnecting to peers
+/// The default values are in the order of seconds
+#[derive(Debug, Clone, Copy)]
```
dpc, line 86: `DelayCalculator` has a name making it seems like a code-object. But it's really just a piece of data. `DelaySettings` maybe or something?
But even then ... it only has `scaling_factor` and `reconnection_delay` method. So maybe `ConnectionSettings` or something? Not sure. Just saying that seems a bit off. I'm big on maintaining clear distinction between "code" and "data".

dpc (commented):
Cool beans. Some comments.

fedimint-server/src/net/connect.rs
douglaz: > This part is tricky, as changes to the code down the line might invalidate this heuristic.

Yes, and it's okay. If tests start to hangup and no bug is found, then the parameters will be scaled down. The default parameter for day-to-day CI activities should be kept on the lower side.

But in future we can create a stress script that tries to run the tests with higher values. Then we can run it periodically (like once a day) or in ad-hoc manner. It could create fancy graphs showing how different metrics change based on different values of failure rate/latency.

> In the past for similar reasons I used adaptive failure rates. Eg. time decaying failure rates, or linearly increasing with time, but geometrically slashed on each failure (kind of like TCP: https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease) .

An adaptive failure rate is a very interesting idea, but I think it would also be tricky to get it right. It would require keeping a state that survives between different (re)connections. For instance, you could start with a higher (or lower) failure rate then get it down (or up) after each reconnection/packet. But this may create a bias where errors are more prone to happen on some moments of the test than others. Also I guess it would still require some tuning for the increase/decrease parameters that may still need some retuning in future. You could end in the same spot but with a more complex model to deal with.

fedimint-server/src/net/peers.rs
dpc: Not a big deal one way or another.

dpc (approved):

elsirion (approved):

Issue #1893, opened 2023-03-13
Separate out an admin UI from fedimintd


jkitman:
Currently our setup UI webserver runs as part of `fedimintd` so we need to:
- Add `fedimint` rpc endpoints (with auth) for running setup (DKG)
- Split the webserver into a separate crate that will only need to depend on the API
- As part of a follow-up issue we should add proper tests and decide on a UI framework

Long-term this gives us advantages:
- Could eventually support other front-ends (mobile or federation-specific)
- Admin UI could run locally for added security (doesn't need to be publicly exposed)

See #2003 for discussions on how to build

isaack-mungui:
I'm interested in contributing to Fedimint as part of Summer of Bitcoin. Is this issue and associated issues related to the Admin/Setup UI available for contribution as part of SoB?

jkitman:
@isaack-njama Yes I believe it could be, please ask in our Discord chat https://discord.gg/se87UJPW

elsirion:
> I'm interested in contributing to Fedimint as part of Summer of Bitcoin. Is this issue and associated issues related to the Admin/Setup UI available for contribution as part of SoB?

Would be a nice SoB project indeed! Please propose it, I'll let Adi know, but I think custom proposals were always welcome.

isaack-njama:
@jkitman @elsirion Thank you for the response. I'll make the proposal in the discord server.

IgboPharaoh:
hello @elsirion , I'll love to contribute to this issue

okjodom:
completed after #2559

Issue #1628, opened 2023-02-09
Use a leader for federation setup


jkitman:
In order to make federation setup easier / less error-prone it makes sense to have a leader that everyone connects to, through this high-level UX flow:

1. All guardians start fedimintd which brings up a UI page
2. A leader publishes a URL to connect to
3. Guardians connect to the URL and send over their connection/auth info
4. The leader configures any consensus values (e.g. federation name, finality delay, etc)
5. After all guardians have connected, the leader shares the final consensus and connection/authentication info with all of them
6. Guardians may validate the connection/auth information of other guardians (published in a human-readable form over other trusted channels)
7. Once all guardians complete validation, they start DKG

Assuming all guardians successfully validated each other's auth info then the outcome of DKG can be trusted and the consensus config can be verified upon start-up, so there is no extra trust placed in the leader.

Tasks

- [x] Build new config API for admins (guardians)
- [ ] Replace DKG CLI with calls to the API
- [ ] Build new UI for #1893

justinmoon:
Comments after another discord discussion:
* Setup flow should be password-protected
* (6) could happen by hashing all pubkeys together and displaying it as BIP39 words or emojis. Everyone confirms this is the same. Or do this in the form of a challenge. Or share connection strings like we do currently.

justinmoon:
@jkitman this one done now?

PR #2010, opened 2023-03-23
Initial config API

jkitman:
Initial stab at the config API.  If we are good with this then it'll be a simple matter of implementing the rest of the endpoints with tests.

Related to #1893 #1628

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -801,12 +801,21 @@ impl WsAuthenticatedApi {
         }
     }
 
+    pub async fn set_password(&self, auth: ApiAuth) -> FederationResult<()> {
+        self.request("/set_password", ApiRequestErased::new(auth))
+            .await
+    }
+
     pub async fn signal_upgrade(&self) -> FederationResult<()> {
+        self.request("/upgrade", ApiRequestErased::default()).await
+    }
+
+    async fn request<Ret>(&self, method: &str, params: ApiRequestErased) -> FederationResult<Ret>
+    where
+        Ret: serde::de::DeserializeOwned + Eq + Debug + Clone + MaybeSend,
+    {
         self.inner
-            .request_current_consensus(
-                "/upgrade".to_owned(),
-                ApiRequestErased::default().with_auth(&self.auth),
-            )
+            .request_current_consensus(method.to_owned(), params.with_auth(&self.auth))
             .await
     }
 }
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -163,7 +163,18 @@ mod tests {
 
     #[test_log::test(tokio::test)]
     async fn test_config_api() {
-        let parent = env::var("FM_TEST_DIR").unwrap_or("/tmp".to_string());
+        let (parent, _maybe_tmp_dir_guard) = match env::var("FM_TEST_DIR") {
+            Ok(directory) => (directory, None),
+            Err(_) => {
+                let guard = tempfile::Builder::new()
+                    .prefix("fm-cfg-api")
+                    .tempdir()
+                    .unwrap();
+                let directory = guard.path().to_str().unwrap().to_owned();
+                (directory, Some(guard))
+            }
+        };
+
         let data_dir = PathBuf::from(parent).join("test-config-api");
         fs::create_dir(data_dir.clone()).expect("Unable to create test dir");
 
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -1,0 +1,180 @@
+use std::net::SocketAddr;
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use async_trait::async_trait;
+use fedimint_core::core::ModuleInstanceId;
+use fedimint_core::db::Database;
+use fedimint_core::module::{
+    api_endpoint, ApiAuth, ApiEndpoint, ApiEndpointContext, ApiError, ApiRequestErased,
+};
+use jsonrpsee::server::{ServerBuilder, ServerHandle};
+use jsonrpsee::RpcModule;
+
+use crate::net::api::{attach_endpoints, HasApiContext, RpcHandlerCtx};
+
+/// Serves the config API endpoints
+pub struct ConfigApi {
+    /// Directory the configs will be created in
+    _data_dir: PathBuf,
+    /// In-memory state machine
+    state: Mutex<ConfigApiState>,
+    /// DB not really used
+    db: Database,
+    /// Bind address for the API
+    _bind_api: SocketAddr,
+}
+
+impl ConfigApi {
+    pub fn new(data_dir: PathBuf, bind_api: SocketAddr, db: Database) -> Self {
+        Self {
+            _data_dir: data_dir,
+            state: Mutex::new(ConfigApiState::SetPassword),
+            db,
+            _bind_api: bind_api,
+        }
+    }
+
+    // Sets the auth and decryption key derived from the password
+    pub fn set_password(&self, auth: ApiAuth) -> Result<(), ApiError> {
+        let mut state = self.state.lock().expect("lock poisoned");
+
+        if *state != ConfigApiState::SetPassword {
+            Err(ApiError::bad_request("Password already set".to_string()))
+        } else {
+            *state = ConfigApiState::SetPeerConnectionInfo(auth);
+            Ok(())
+        }
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+/// State machine for config generation, required in a top-to-bottom order
+pub enum ConfigApiState {
+    /// Guardian must first enter a password for auth/decryption
+    SetPassword,
+    /// Guardian must add connection info for at least 1 peer
+    SetPeerConnectionInfo(ApiAuth),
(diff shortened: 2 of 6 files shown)


fedimint-core/src/api.rs
```diff
@@ -801,12 +801,21 @@ impl WsAuthenticatedApi {
         }
     }
 
+    pub async fn set_password(&self, auth: ApiAuth, key: DecryptionKey) -> FederationResult<()> {
+        self.request("/set_password", ApiRequestErased::new((auth, key)))
+            .await
+    }
+
```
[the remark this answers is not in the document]
jkitman: It would be the same user-created password, but the plan is the plaintext password would never touch `fedimintd` but be hashed client-side to derive the auth token and the decryption key.

fedimint-server/src/config/api.rs
```diff
@@ -0,0 +1,192 @@
...
+    _key: DecryptionKey,
+}
+
+/// The current state of config generation, required in a top-to-bottom order
+#[derive(Debug, Eq, PartialEq)]
+pub enum ConfigGenState {
+    /// Guardian must first enter a password for auth/decryption
+    SetPassword,
+    /// Guardian must add connection info for at least 1 peer
+    SetPeerConnectionInfo,
```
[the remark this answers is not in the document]
jkitman: Yes the new API should still support a 1/1 federation.  The "at least 1 peer" is for when you have a leader, they should be that one.

fedimint-core/src/module/mod.rs
```diff
@@ -110,10 +110,15 @@ impl ApiRequestErased {
...
 
-/// Authentication uses the hashed user password in PHC format
+/// Authentication should be the hashed user password in PHC format
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ApiAuth(pub String);
 
+/// Decryption key should be the hashed user password (different salt than the
+/// `ApiAuth`) to encrypt the private key material in configs
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DecryptionKey(pub [u8; 32]);
```
dpc, line 120: This doesn't say anything about what for is this decryption key, or type. Is it supposed to be a generic newtype over any decryption key we might use?

dpc (approved):
Could use more comments, as it's a yet another area of the system and it uses lots of terms that are already used everywhere else (config, auth, keys).

When someone is browsing the codebase trying to fix something (possibly unrelated), would be great if the top-level module comment would tell them "this is (not) what you're looking for", etc.

This doesn't seem wired into normal code anywhere yet, so not a big deal if we land so approving, but I remember that @elsirion had some requirements for admin auth so you probably should get his feedback.

fedimint-core/src/module/mod.rs
jkitman: The decryption key can be derived client-side using any technique, but it's purpose will be "to encrypt the private key material in configs" as in the comment.  Maybe I should rename it `ConfigDecryptionKey`?

dpc: > Maybe I should rename it ConfigDecryptionKey?

+ a comment and I think that's good.

fedimint-core/src/api.rs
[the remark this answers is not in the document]
elsirion: > This gives clients some flexibility in choosing how to stretch the key

Flexibility is typically not what you want from a cryptographic protocol if you can avoid it. If there are to be multiple ways these should be enumerated in a standards document (so different client impls can follow a common set) and the fedimint instance should tell clients which way to use on setup to avoid a scenario like with WiFis where it's "good luck guessing which combination of options you need to make this enterprise WiFi work even though you know your login data".

fedimint-server/src/config/api.rs
```diff
@@ -0,0 +1,192 @@
...
+use crate::net::api::{attach_endpoints, HasApiContext, RpcHandlerCtx};
+
+/// The config state needed for serving the config API
+pub struct ConfigApi {
+    /// Directory the configs will be created in
+    _data_dir: PathBuf,
+    /// In-memory state machine
+    state: Mutex<ConfigGenState>,
+    /// In-memory keys
+    keys: Mutex<Option<ConfigGenKeys>>,
```
elsirion, line 25: Why two `Mutex`es, this just makes it more likely to accidentally build a deadlock.

fedimint-server/src/config/api.rs
```diff
@@ -0,0 +1,192 @@
...
+    api_endpoint, ApiAuth, ApiEndpoint, ApiEndpointContext, ApiError, ApiRequestErased,
+    DecryptionKey,
+};
+use jsonrpsee::server::{ServerBuilder, ServerHandle};
+use jsonrpsee::RpcModule;
+
+use crate::net::api::{attach_endpoints, HasApiContext, RpcHandlerCtx};
+
+/// The config state needed for serving the config API
+pub struct ConfigApi {
```
elsirion, line 19: It's more of a `ConfigGenApi`, isn't it?

fedimint-server/src/config/api.rs
```diff
@@ -0,0 +1,192 @@
...
+    impl TestConfigApi {
+        async fn set_password(&self) -> FederationResult<()> {
+            self.client
+                .set_password(self.auth.clone(), self.key.clone())
+                .await
+        }
+    }
+
+    async fn server_client(port: u16) -> TestConfigApi {
+        let parent = env::var("FM_TEST_DIR").unwrap_or("../../".to_string());
```
elsirion, line 161: I'm a bit scared what `"../../"` might be when I run the tests :grimacing: we should use tempdir in that case.

elsirion (changes requested):
I think at least the "path traversal in tests" comment needs addressing, some other nits we can discuss.

fedimint-server/src/config/api.rs
jkitman: Wanted to use `ConfigGenState` at the top-level, but I could wrap it in another struct.

fedimint-core/src/api.rs
```diff
@@ -801,12 +801,21 @@ impl WsAuthenticatedApi {
         }
     }
 
+    pub async fn set_password(&self, auth: ApiAuth, key: DecryptionKey) -> FederationResult<()> {
```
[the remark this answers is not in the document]
jkitman: I suppose that keeps it simpler for now.

fedimint-server/src/config/api.rs
Maan2003: just panic might be better. forcing the caller to set `FM_TEST_DIR`

fedimint-server/src/config/api.rs
jkitman: Well technically it will be used after config gen runs as well.  But to address @dpc's point I could prefix these structs with `ConfigApi` to better namespace them.

fedimint-server/src/config/api.rs
jkitman: I could default to creating a test dir.

fedimint-server/src/config/api.rs
```diff
@@ -0,0 +1,192 @@
...
+
+#[async_trait]
+impl HasApiContext<ConfigApi> for ConfigApi {
+    async fn context(
+        &self,
+        request: &ApiRequestErased,
+        id: Option<ModuleInstanceId>,
+    ) -> (&ConfigApi, ApiEndpointContext<'_>) {
+        let dbtx = self.db.begin_transaction().await;
+        let keys = self.keys.lock().expect("locks");
```
Maan2003, line 88:
```suggestion
        let keys = self.keys.lock().expect("lock poisoned");
```

fedimint-core/src/api.rs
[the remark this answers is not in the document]
jkitman: Currently the requirement is it is in PHC format which is the closest thing to a standard for hashed passwords.

The server-side doesn't necessarily need to care about how the password was hashed, but I guess if there are multiple clients that need to be compatible they should use the same hashing algorithm and salt.

elsirion (dismissed):

elsirion (approved):

elsirion:
I added the random tmp dir creation so that test runs are independent and automatically cleaned up. Feel free to squash. 

I validated that the directory and cfg file are created (when inotify is fast enough):

```
[nix-shell:~]$ while RES=$(inotifywait -e create /tmp/nix-shell.VSq5SL/ --format %f .); do inotifywait  /tmp/nix-shell.VSq5SL/$RES; done
Setting up watches.
Watches established.
Setting up watches.
Couldn't watch /tmp/nix-shell.VSq5SL/fm-cfg-apiwpTmAf: No such file or directory
Setting up watches.
Watches established.
Setting up watches.
Watches established.
/tmp/nix-shell.VSq5SL/fm-cfg-apiNBTwiy/ DELETE,ISDIR test-config-api
Setting up watches.
Watches established.
Setting up watches.
Watches established.
/tmp/nix-shell.VSq5SL/fm-cfg-apiXehLQZ/ DELETE,ISDIR test-config-api
```

jkitman:
> I added the random tmp dir creation so that test runs are independent and automatically cleaned up. Feel free to squash.

Thanks, sorry I was being dense about what you were asking for.

dpc (approved):

PR #2012, opened 2023-03-23
add generated dirs to gitignore

jkitman:
There are a couple dirs now generated when running tests that we can ignore.

Talking with @m1sterc001guy `db/` will exist after https://github.com/fedimint/fedimint/pull/1935

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+*target/
 /result
 .idea
 .tmpenv


m1sterc001guy:
Database backups are generated during tests, but should only be generated if the backup doesn't already exist. Are you seeing modifications to these files when running tests?

I'm not sure if we should ignore them from git, because eventually we should check in more backups when the database schema is changed

.gitignore
```diff
@@ -1,4 +1,4 @@
-/target
+*target/
```
dpc, line 1: We shouldn't have `target` dirs in any subdirectories.

dpc: And I don't get it. 
The discussion is about `db/...` but I don't see any of it here... ?

jkitman: Removed the `db/`, we create `fedimint-client/target/`

dpc: > Removed the `db/`, we create `fedimint-client/target/`

How... why? `fedimint-client` seems like a part of the whole workspace ...

dpc: I did have `fedimint-client/target/`. I deleted it. Did `just build` and I no longer see it. Maybe it was already fixed?

jkitman: `just test` seems to produce it for me

dpc:
```
fedimint-client/target/doc/mermaid.min.js
```

Oh... I see.

dpc: https://github.com/mersinvald/aquamarine/blob/636675ba342a37a2a0aa87eb9adeea5c266f505c/src/attrs.rs#L138 :dizzy_face:

dpc: https://github.com/mersinvald/aquamarine/issues/38

elsirion: Let's go with this for now till upstream is fixed

elsirion (approved):

elsirion:
This PR doesn't solve the original DB issue anymore (but that was solved otherwise I assume?), but adding the other target folder for now sounds good till https://github.com/mersinvald/aquamarine/issues/38 is resolved.

PR #2015, opened 2023-03-23
WIP: One Click Digital Ocean Fedimint Deployment

m1sterc001guy:
This is a PR that adds a button to the README that will deploy a docker container on Digital Ocean with one click. This might be useful for people wanting to play around with Fedimint but don't want to build it from source using the developer environment. 

image

This is a work in-progress so there are a number of things that can be improved:

 - Use master (or some "tagged" commit) instead of the hard coded hash.
 - Figure out how to get rid of the Dockerfile at the root of the directory (in my limited testing Digital Ocean wants the Dockerfile at the root)
 - Expose all of the necessary ports (currently only exposing the port for the setup UI)
 - Auto-fill parameters to sensible defaults (bitcoin connection to public electrum, signet/testnet, autofill listen addresses, etc)
 - Deploy a "[shellinabox](https://github.com/shellinabox/shellinabox)" container that has `fedimint-cli` already installed so the user can immediately begun playing with ecash.

### DIFF
--- a/.do/deploy.template.yaml
+++ b/.do/deploy.template.yaml
@@ -1,21 +1,7 @@
-alerts:
-- rule: DEPLOYMENT_FAILED
-- rule: DOMAIN_FAILED
-name: fedimintd
-region: nyc
-services:
-- envs:
-  - key: FM_PASSWORD
-    scope: RUN_AND_BUILD_TIME
-    value: pass0
-  http_port: 8176
-  image:
-    registry: fedimint
-    registry_type: DOCKER_HUB
-    repository: fedimintd
-    tag: master
-  instance_count: 1
-  instance_size_slug: basic-xs
-  name: fedimint-fedimintd
-  routes:
-  - path: /
+spec:
+  name: fedimint
+  services:
+    - name: fedimintd
+      git:
+        branch: digital_ocean_deploy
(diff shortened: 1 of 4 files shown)


dpc:
Interesting. My first thought: we should put some warning for the users in the setup UI to make sure they understand using with real funds is risky.

justinmoon:
Super cool. I suggest we make another repo in the fedimint org containing this stuff, and agree with @dpc that we should heavily discourage mainnet usage for at the start. Maybe even disable it if user doesn't go set special environment variable or something.

I tried running it and I think we'll want to only have 1 container here? 
image. Seems like this might try to load-balance between the 2 somehow? 

But in about 60 seconds I was able to get the setup UI to show up. Pretty sick!

m1sterc001guy:
> Also noticed that it default to 2 containers here. I think we'll only ever want 1.

I think this is an issue with how the app spec is specified + the Dockerfile at the root of the repo. I think it should be possible to have this working without a Dockerfile, but I was hitting errors while testing. Ideally I think it would be nice to keep this on the main repo, but some of these issues need to be solved before doing that.

m1sterc001guy:
Closing this for now as I'm not actively working on it.

PR #2018, opened 2023-03-23
Test tx already in blockchain

### LINKED ISSUE #2017: Add test to check backend behavior on tx already included in the blockchain
Like in https://github.com/fedimint/fedimint/pull/2016 it's common issue that the backend returns an error if the tx is already included in the blockchain.

Take one of the tests already in integrationtests/tests/tests.rs and modify it to:

* make a peg-out
* use btc rpc backend to submit the peg-out transaction again, and check that it succeeds (broadcasted tx should be already included)

Revert #2016 to check if the new tests catches the problem.

jkitman:
Reproduces error in #2017 confirmed on eletrs + esplora

Fixes #2017

### DIFF
--- a/integrationtests/tests/tests.rs
+++ b/integrationtests/tests/tests.rs
@@ -92,7 +92,11 @@ async fn wallet_peg_in_and_peg_out_with_fees() -> Result<()> {
                 ..
             }) if *txid == outcome_txid));
 
-        fed.broadcast_transactions().await;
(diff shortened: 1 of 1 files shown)


elsirion (approved):

mxz42 (approved):

mxz42:
#2017 mentioned to revert #2016 but since that was never merged, there's no need. so I'll merge

Issue #1876, opened 2023-03-12
`gatewayd` / `gateway-cln-extension` startup race condition


justinmoon:
`gatewayd` can't start if `FM_GATEWAY_LIGHTNING_ADDR` set and `gateway-cln-extension` not running. We shouldn't have a race condition like this. Either one should be able to startup first.

elsirion:
I think it comes down to having some reconnect logic, a lot like with the API client. This will also help when CLN+extension get restarted but not `gatewayd` (otherwise `gatewayd` would just crash I assume?).

justinmoon:
note: talk to @Maan2003 about websocket client reconnect struct etc ...

justinmoon:
- would need to retry connecting somewhere in gatewayd
- remove `await_gateway_cln_extension` and put a sleep in `gateway-cln-extension`'s `main()` to trigger this

m1sterc001guy:
I'll look into this

m1sterc001guy:
I think there's probably two improvements that should be done here:

1) Fix the race condition on startup (this is easy, we can add a simple loop that retries).
2) Refactor `gatewayd` to have proper error handling when the connection to the lightning node dies. Currently the connection to the lightning node is a `Arc<ILnRpcConnect>` and the Arc is simply cloned to other actors that need to do lightning RPCs. If the connection dies we don't currently have a way of reconstructing the lightning connection. `Gatewayd` doesn't actually crash, it simply won't work currently.

PR #2019, opened 2023-03-23
feat: allow gatewayd to reconnect to CLN or LND

m1sterc001guy:
PR adds the ability for `gatewayd` to reconnect to CLN or LND if they go down. This API could also be used to switch which lightning implementation is used without needing to restart `gatewayd`.

Manual testing using tmuxinator is working. Integration test for sending an outgoing payment after a disconnection has been added.

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -7,8 +7,9 @@ use fedimint_logging::TracingSetup;
 use ln_gateway::rpc::rpc_client::RpcClient;
 use ln_gateway::rpc::{
     BackupPayload, BalancePayload, ConnectFedPayload, DepositAddressPayload, DepositPayload,
-    RestorePayload, WithdrawPayload,
+    LightningReconnectPayload, RestorePayload, WithdrawPayload,
 };
+use ln_gateway::Mode;
 use mint_client::modules::wallet::txoproof::TxOutProof;
 use mint_client::utils::from_hex;
 use url::Url;
@@ -63,6 +64,11 @@ pub enum Commands {
     Backup { federation_id: FederationId },
     /// Restore ecash from last available snapshot or from scratch
     Restore { federation_id: FederationId },
+    // Reconnect to the Lightning Node
+    Reconnect {
+        #[clap(subcommand)]
+        mode: Mode,
+    },
 }
 
 #[tokio::main]
@@ -165,6 +171,28 @@ async fn main() -> anyhow::Result<()> {
                 )
                 .await?;
 
+            print_response(response).await;
+        }
+        Commands::Reconnect { mode } => {
+            let payload = match mode {
+                Mode::Cln { cln_extension_addr } => LightningReconnectPayload {
+                    node_type: Some(Mode::Cln { cln_extension_addr }),
+                },
+                Mode::Lnd {
+                    lnd_rpc_addr,
+                    lnd_tls_cert,
+                    lnd_macaroon,
+                } => LightningReconnectPayload {
+                    node_type: Some(Mode::Lnd {
+                        lnd_rpc_addr,
+                        lnd_tls_cert,
+                        lnd_macaroon,
+                    }),
+                },
+            };
+            let response = client
+                .reconnect(source_password(cli.rpcpassword), payload)
+                .await?;
             print_response(response).await;
         }
     }
--- a/gateway/ln-gateway/src/actor.rs
+++ b/gateway/ln-gateway/src/actor.rs
@@ -1,16 +1,21 @@
+use std::pin::Pin;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::time::Duration;
 
 use bitcoin::{Address, Transaction};
 use bitcoin_hashes::{sha256, Hash};
-use fedimint_core::task::TaskGroup;
+use fedimint_core::task::{RwLock, TaskGroup};
 use fedimint_core::{Amount, OutPoint, TransactionId};
 use futures::stream::StreamExt;
+use futures::Stream;
 use mint_client::modules::ln::contracts::{ContractId, Preimage};
 use mint_client::modules::ln::route_hints::RouteHint;
 use mint_client::modules::wallet::txoproof::TxOutProof;
 use mint_client::{GatewayClient, PaymentParameters};
 use rand::{CryptoRng, RngCore};
+use tokio::sync::mpsc::{self, Receiver, Sender};
+use tonic::Status;
 use tracing::{debug, error, info, instrument, warn};
 
 use crate::gatewaylnrpc::complete_htlcs_request::{Action, Cancel, Settle};
@@ -18,8 +23,8 @@ use crate::gatewaylnrpc::{
     CompleteHtlcsRequest, PayInvoiceRequest, PayInvoiceResponse, SubscribeInterceptHtlcsRequest,
     SubscribeInterceptHtlcsResponse,
 };
-use crate::lnrpc_client::DynLnRpcClient;
-use crate::rpc::FederationInfo;
+use crate::lnrpc_client::ILnRpcClient;
+use crate::rpc::{FederationInfo, GatewayRpcSender, LightningReconnectPayload};
 use crate::utils::retry;
 use crate::{GatewayError, Result};
 
@@ -29,8 +34,10 @@ const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
 #[derive(Clone)]
 pub struct GatewayActor {
     client: Arc<GatewayClient>,
-    lnrpc: DynLnRpcClient,
+    pub lnrpc: Arc<RwLock<dyn ILnRpcClient>>,
     task_group: TaskGroup,
+    gw_rpc: GatewayRpcSender,
+    sender: Option<Sender<Arc<AtomicBool>>>,
 }
 
 #[derive(Debug, Clone)]
@@ -39,15 +46,25 @@ pub enum BuyPreimage {
     External(Preimage),
 }
 
+type HTLCStream = Pin<
+    Box<
+        dyn Stream<Item = std::result::Result<SubscribeInterceptHtlcsResponse, Status>>
+            + Send
+            + 'static,
+    >,
+>;
+
 impl GatewayActor {
     pub async fn new(
         client: Arc<GatewayClient>,
-        lnrpc: DynLnRpcClient,
+        lnrpc: Arc<RwLock<dyn ILnRpcClient>>,
         route_hints: Vec<RouteHint>,
         task_group: TaskGroup,
+        gw_rpc: GatewayRpcSender,
     ) -> Result<Self> {
         let register_client = client.clone();
-        tokio::spawn(async move {
+        let mut tg = task_group.make_subgroup().await;
+        tg.spawn("Register with federation", |_| async move {
             loop {
                 // Retry gateway registration
                 match retry(
@@ -76,12 +93,15 @@ impl GatewayActor {
                     }
                 }
             }
-        });
+        })
+        .await;
 
-        let actor = Self {
+        let mut actor = Self {
             client,
             lnrpc,
-            task_group,
+            task_group: tg,
+            gw_rpc,
+            sender: None,
         };
 
         actor.subscribe_htlcs().await?;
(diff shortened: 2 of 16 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -252,13 +256,17 @@ impl Gateway {
...
-            .collect();
+        let actors = self.actors.lock().await;
+        let mut federations: Vec<FederationInfo> = Vec::new();
+        for (_, actor) in actors.iter() {
+            federations.push(
+                actor
+                    .read()
+                    .await
+                    .get_info()
+                    .expect("Failed to get actor info"),
```
justinmoon, line 267: Can we get rid of this `.expect` while we're at it?

justinmoon:
Closes https://github.com/fedimint/fedimint/issues/1876

gateway/cli/src/main.rs
```diff
@@ -167,6 +173,29 @@ async fn main() -> anyhow::Result<()> {
...
+                        lnd_rpc_addr,
+                        lnd_tls_cert,
+                        lnd_macaroon,
+                    }),
+                },
+            };
+            let response = client
+                .reconnect(source_password(cli.rpcpassword), payload)
+                .await
+                .expect("Failed to reconnect");
```
dpc, line 196: `expect` :(

gateway/ln-gateway/src/actor.rs
```diff
@@ -76,53 +81,80 @@ impl GatewayActor {
...
         actor.subscribe_htlcs().await?;
 
         Ok(actor)
     }
 
-    async fn subscribe_htlcs(&self) -> Result<()> {
+    pub async fn stop_subscribing_htlcs(&mut self) -> Result<()> {
+        if let Some(sender) = &self.sender {
+            sender
+                .send(())
```
dpc, line 103: If you're sending `()` you probably want to use something else. like `Arc<Atomicbool>`.

gateway/ln-gateway/src/actor.rs
```diff
@@ -76,53 +81,80 @@ impl GatewayActor {
...
                     intercepted_htlc_id,
                     ..
-                }) = match stream.next().await {
-                    Some(msg) => match msg {
-                        Ok(msg) => Some(msg),
-                        Err(e) => {
-                            warn!("Error sent over HTLC subscription: {}", e);
+                }) = tokio::select! {
+                    msg = stream.next() => match msg {
+                        Some(msg) => match msg {
```
dpc, line 138: right margin drift is very high here . Maybe move `while let Some(...)` to a sub-function?

Possibly

```
match msg {
  Some(Ok(msg)) => Some(msg),
  Some(Err(e))) => ...
...
}
```

just to remove one level of nesting?

gateway/ln-gateway/src/lnrpc_client.rs
```diff
@@ -72,64 +65,109 @@ impl NetworkLnRpcClient {
...
+        error!("Error: not connected to CLN extension.");
+        Err(GatewayError::Other(anyhow!(
+            "Error not connected to CLN extension"
+        )))
+    }
+
+    async fn connect(&mut self) -> Result<()> {
+        if self.client.is_none() {
+            let mut res = GatewayLightningClient::connect(self.endpoint.clone()).await;
+            while res.is_err() {
```
dpc, line 155: `loop`s can return value.

```rust
let client = loop {
  match  GatewayLightningClient::connect ... {
     Ok(client) => break client;
     Err(e) => ...,
  }
}
```   

Will avoid `mut` and `unwrap`

gateway/ln-gateway/src/lnrpc_client.rs
```diff
@@ -72,64 +65,109 @@ impl NetworkLnRpcClient {
...
+            let res = client.get_route_hints(req).await?;
 
-        let mut client = self.client.clone();
-        let res = client.get_route_hints(req).await?;
+            return Ok(res.into_inner());
+        }
 
-        Ok(res.into_inner())
+        error!("Error: not connected to CLN extension.");
+        Err(GatewayError::Other(anyhow!(
```
dpc, line 102: Maybe make a `fn GatewayError:other(msg: &str) -> Self` since it repeats many times.

justinmoon:
This PR adds the ability to run `gateway-cli reconnect ...` and connect the gateway to a new lightning node. This will be really nice in tests and stuff, but it's a bit of a foot-gun in production since any clients that had downloaded that gateway's old config (containing the old node's pubkey etc) would now have invalid configs. It's a shortcoming of the old client moreso than this PR since old client isn't smart about downloading new gateway configs when they change. With new client we'll need to improve that.

gateway/ln-gateway/src/lnd.rs
```diff
@@ -314,6 +329,38 @@ impl ILnRpcClient for GatewayLndClient {
             );
         }
     }
+
+    async fn connect(&mut self) -> crate::Result<()> {
+        if self.client.is_none() {
```
dpc, line 334: `while self.client.is_none()`

gateway/ln-gateway/src/lnd.rs
```diff
@@ -314,6 +329,38 @@ impl ILnRpcClient for GatewayLndClient {
...
+            let mut res = connect(
+                self.address.clone(),
+                self.tls_cert.clone(),
+                self.macaroon.clone(),
+            )
+            .await;
+            while res.is_err() {
+                tracing::warn!("Couldn't connect to LND, retrying in 5 seconds...");
+                sleep(Duration::from_secs(5)).await;
+                res = connect(
```
dpc, line 344: This `connect` repeating itself makes it much longer and confusing then needed.

gateway/ln-gateway/src/lib.rs
```diff
@@ -322,14 +361,75 @@ impl Gateway {
...
+            )),
+            None => self.lnrpc.clone(),
+        };
+
+        new_lnrpc.write().await.connect().await?;
+        self.lnrpc = new_lnrpc.clone();
+
+        // Restart the subscription of HTLCs for each actor
+        tracing::info!("Restarting HTLC subscription threads.");
+        for (_, actor_lock) in actors.iter() {
```
dpc, line 426: `iter()` in `for` loop shouldn't be neccessary

m1sterc001guy: I get an error when I remove it. I think its necessary because the `actors` HashMap is guarded by a `Mutex`.

gateway/ln-gateway/src/actor.rs
m1sterc001guy: Done. What's the reason why? Is there some subtleness using () in channels?

gateway/ln-gateway/src/actor.rs
m1sterc001guy: Moved to a function and removed that one level of nesting

gateway/ln-gateway/src/lnrpc_client.rs
```diff
@@ -42,18 +42,10 @@ pub trait ILnRpcClient: Debug + Send + Sync {
...
     /// determining an outcome
     async fn complete_htlc(&self, outcome: CompleteHtlcsRequest) -> Result<CompleteHtlcsResponse>;
-}
 
-dyn_newtype_define!(
-    /// Arc reference to a gateway lightning rpc client
-    #[derive(Clone)]
-    pub DynLnRpcClient(Arc<ILnRpcClient>)
-);
+    async fn connect(&mut self) -> Result<()>;
```
justinmoon, line 46: add docstring please. and to disconnect().

gateway/ln-gateway/src/lib.rs
```diff
@@ -253,15 +292,13 @@ impl Gateway {
...
-            .map(|(_, actor)| actor.get_info().expect("Failed to get actor info"))
-            .collect();
+        let actors = self.actors.lock().await;
+        let mut federations: Vec<FederationInfo> = Vec::new();
+        for (_, actor) in actors.iter() {
+            federations.push(actor.read().await.get_info()?);
+        }
 
-        let ln_info = self.lnrpc.info().await?;
+        let ln_info = self.lnrpc.read().await.info().await?;
```
m1sterc001guy, line 301: Ideally we should check if the lightning node is connected or not. And return a value indicating if its connected or not.

gateway/ln-gateway/src/lib.rs
dpc: Oh. It's because `MutexGuard` needs a dereference (which `.` provides implicitily).

This works:

```
-        for (_, actor) in actors.iter() {
+        for (_, actor) in &*actors {
```

but then clippy complains that you're ignoring keys, so should be doing:

```
-        for (_, actor) in actors.iter() {                                                                                                                                           
+        for actor in actors.values() {
```

instead, which I guess is the right thing to do.

gateway/ln-gateway/src/lib.rs
```diff
@@ -329,14 +375,75 @@ impl Gateway {
...
+                    lnd_tls_cert,
+                    lnd_macaroon,
+                    self.task_group.make_subgroup().await,
+                )
+                .await?,
+            )),
+            None => self.lnrpc.clone(),
+        };
+
+        new_lnrpc.write().await.connect().await?;
```
m1sterc001guy, line 435: This looks like its connecting twice unless you're reusing the existing client (once in the constructor and once here)

gateway/ln-gateway/src/actor.rs
```diff
@@ -76,154 +93,204 @@ impl GatewayActor {
...
+        stream: &mut HTLCStream,
+        receiver: &mut Receiver<Arc<AtomicBool>>,
+        gw_rpc_copy: GatewayRpcSender,
+    ) -> Option<SubscribeInterceptHtlcsResponse> {
+        tokio::select! {
+            msg = stream.next() => match msg {
+                Some(Ok(msg)) => Some(msg),
+                Some(Err(e)) => {
+                    warn!("Error sent over HTLC subscription: {}. Sending reconnect RPC", e);
+                    let reconnect_req = LightningReconnectPayload { node_type: None };
```
m1sterc001guy, line 132: Add comment here

gateway/ln-gateway/src/lib.rs
```diff
@@ -329,14 +375,75 @@ impl Gateway {
...
+        let actors = self.actors.lock().await;
+
+        // Stop all threads that are listening for HTLCs
+        tracing::info!("Stopping all HTLC subscription threads.");
+        for (_, actor) in actors.iter() {
+            actor.write().await.stop_subscribing_htlcs().await?;
+        }
+
+        // Disconnect the lightning connection, then reconnect it
+        self.lnrpc.write().await.disconnect().await?;
```
m1sterc001guy, line 413: Make an issue to see if we can get rid of `disconnect` and `connect`

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Actually I was mistaken on this, this is already taken care of. `info` will check if we're connected and return an error if its not connected.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Made issue, will work on this next: https://github.com/fedimint/fedimint/issues/2069

gateway/ln-gateway/src/lnrpc_client.rs
m1sterc001guy: Done

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Changes to use `values()`

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Fixed this, will probably do more refactoring when I do this: https://github.com/fedimint/fedimint/issues/2069

gateway/ln-gateway/src/lib.rs
```diff
@@ -253,15 +292,13 @@ impl Gateway {
...
-        let federations = self
-            .actors
-            .lock()
-            .await
-            .iter()
-            .map(|(_, actor)| actor.get_info().expect("Failed to get actor info"))
-            .collect();
+        let actors = self.actors.lock().await;
+        let mut federations: Vec<FederationInfo> = Vec::new();
+        for (_, actor) in actors.iter() {
```
dpc, line 297: `.values()` too?

gateway/ln-gateway/src/lib.rs
```diff
@@ -90,7 +126,7 @@ pub struct Gateway {
 impl Gateway {
     #[allow(clippy::too_many_arguments)]
     pub async fn new(
-        lnrpc: DynLnRpcClient,
+        lnrpc: Arc<RwLock<dyn ILnRpcClient>>,
```
dpc, line 129: I wonder if we should use https://crates.io/crates/arc-swap . I wish it was a part of stdlib, as it's a common use-case.

justinmoon (dismissed):
codecov warnings are pretty annoying. make it a lot harder to read the diff.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Ack, but let me do this in the next refactor: https://github.com/fedimint/fedimint/issues/2069

I think once I get rid of the mut functions on `ILnRpcClient` it will become easier to use `ArcSwap`

gateway/ln-gateway/src/lib.rs
```diff
@@ -329,14 +375,79 @@ impl Gateway {
...
+            }
+        };
+
+        self.lnrpc = new_lnrpc.clone();
+
+        // Restart the subscription of HTLCs for each actor
+        tracing::info!("Restarting HTLC subscription threads.");
+        for actor in actors.values() {
+            let mut actor = actor.write().await;
+            actor.lnrpc = new_lnrpc.clone();
```
m1sterc001guy, line 446: I don't think this line is necessary actually

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Actually, does `ArcSwap` only provide the ability to atomically swap out <T> (in this case, `ILnRpcClient`)? I think we actually need the ability to mutate `ILnRpcClient`. Currently we set the underlying client to None when the node is disconnected.

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -74,6 +74,13 @@ pub struct RestorePayload {
     pub federation_id: FederationId,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct LightningReconnectPayload {
```
okjodom, line 78: Since payload is only `Some` when we are looking to reconnect to a node which potentially replaces the one that gatewayd currently knows, maybe we could rename this payload to `ReplaceLightningPayload`

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -34,6 +34,7 @@ pub async fn run_webserver(
         .route("/connect", post(connect))
         .route("/backup", post(backup))
         .route("/restore", post(restore))
+        .route("/reconnect", post(reconnect))
```
okjodom, line 37: `/connect` and `/reconnect` are very confusing and poorly named APIs.
I propose we follow up with a rename like:
- `/connect` > `/connect-fed`
- `/reconnect` > `/connect-ln`

okjodom (approved):
Left some ideas for future follow ups

gateway/ln-gateway/src/rpc/rpc_server.rs
okjodom: #2073

gateway/ln-gateway/src/lib.rs
dpc: `Arc<RwLock<dyn ILnRpcClient>>`) has no `Option`? Last time I checked `ILnRpcClient` had only `&self` methods, so other than ability to swap it whole, I don't see any mutability. Might have missed something.

PR #2021, opened 2023-03-23
fix: ignore unreliable result of `submit_transaction`

dpc:
See comment in the code.

This fixes hangs when backends keep returning error due to transactions being submitted already being included in the blockchain: https://github.com/fedimint/fedimint/actions/runs/4503632304/jobs/7927075007?pr=2004


jkitman (approved):
This seems fine since `run_broadcast_pending_tx` will retry until the transaction is confirmed.

elsirion (approved):

Issue #1857, opened 2023-03-10
Don't copy config files in test environment


justinmoon:
Instead just pass the full path to each invocation https://github.com/fedimint/fedimint/pull/1851#pullrequestreview-1335336570

ayush-ranjan-official:
please assign this issue to me

Maan2003:
this is not a good first issue. Referring to original config file is not easy due to nix and it is horrible to debug.

ayush-ranjan-official:
@elsirion I was currently working on this one, please re-assign me.

dpc:
Detail, not worth keeping around.

PR #2024, opened 2023-03-24
Removed cp misc/test/bitcoin.conf $FM_BTC_DIR

ayush-ranjan-official:
Issue Solved : #1857 
Removed `cp misc/test/bitcoin.conf $FM_BTC_DIR` and directly created `bitcoin.conf `when we `mkdir -p $FM_BTC_DIR` using `cat <<EOF > $FM_BTC_DIR/bitcoin.conf` and added `.arg(format!("-conf={btc_dir}/bitcoin.conf"))` to CLI flag when bitcoin daemon is spawned.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -41,11 +41,21 @@ export FM_ELECTRS_DIR="$FM_TEST_DIR/electrs"
 mkdir -p $FM_CLN_DIR
 mkdir -p $FM_LND_DIR
 mkdir -p $FM_BTC_DIR
+cat <<EOF > $FM_BTC_DIR/bitcoin.conf
+regtest=1
+fallbackfee=0.0004
+txindex=1
+server=1
+rpcuser=bitcoin
+rpcpassword=bitcoin
+zmqpubrawblock=tcp://127.0.0.1:38332
+# commonly used port 28333 collides with dkg ports
+zmqpubrawtx=tcp://127.0.0.1:38333
+EOF
 mkdir -p $FM_CFG_DIR
 mkdir -p $FM_ELECTRS_DIR
 
 # Copy configs to data directories
-cp misc/test/bitcoin.conf $FM_BTC_DIR
 cp misc/test/lnd.conf $FM_LND_DIR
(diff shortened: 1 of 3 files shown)


Maan2003:
Why is this better than cp?

ayush-ranjan-official:
> Why is this better than cp?

We want to avoid mutability (copying files).

scripts/build.sh
```diff
@@ -41,11 +41,21 @@ export FM_ELECTRS_DIR="$FM_TEST_DIR/electrs"
 mkdir -p $FM_CLN_DIR
 mkdir -p $FM_LND_DIR
 mkdir -p $FM_BTC_DIR
+cat <<EOF > $FM_BTC_DIR/bitcoin.conf
```
dpc, line 44: I don't get it. I thought the goal is to point with `-c` at the existing file, and here are we create a new one, which makes it the same as `cp`, just from an inline source?

ayush-ranjan-official: The directory of Existing bitcoin.conf file is not same as that of the fixtures.rs hence we can't directly add "-conf=" at the place where we are spawning the bitcoind. Also in the issue it was asked to get rid of the cp command and instead pass the full path . This the way I could think of to fix it.

justinmoon:
> > Why is this better than cp?
> 
> We want to avoid mutability (copying files).

You are still basically copying a file here. The goal of https://github.com/fedimint/fedimint/issues/1857 is to not write any configs into the `$FM_TMP_DIR` at all. Just refer to the ones in `misc/test/` when spawning daemons.

ayush-ranjan-official:
> > > Why is this better than cp?
> > 
> > 
> > We want to avoid mutability (copying files).
> 
> You are still ~basically copying a file here. The goal of #1857 is to not write any configs into the `$FM_TMP_DIR` at all. Just refer to configs in `misc/test/` when spawning daemons.

I tried doing that, please see this : `https://github.com/fedimint/fedimint/pull/2002`
But I am getting this error : `retrying JSON-RPC error: transport error: Couldn't connect to host: Connection refused (os error 111)`

PR #2026, opened 2023-03-24
fix(cli): re-add default value for invoice description

jp1ac4:
This change was originally made in #1996, but I think was accidentally removed in #2004.

### DIFF
--- a/client/cli/src/main.rs
+++ b/client/cli/src/main.rs
@@ -339,6 +339,7 @@ enum Command {
     LnInvoice {
         #[clap(value_parser = parse_fedimint_amount)]
         amount: Amount,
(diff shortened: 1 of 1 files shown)


dpc (approved):

dpc:
Interesting. No idea how this happened. I had to do a rebase in https://github.com/fedimint/fedimint/pull/2004 and I remember dumping bunch of text to a.txt and b.txt to figure out what caused merge conflict and there was only a spelling difference. I still even have that a.txt and b.txt file.

jp1ac4:
Ah that's strange. I noticed just by chance when trying to create an invoice a little earlier.

justinmoon (approved):

PR #2028, opened 2023-03-24
feat: make fedimint-cli an extendable binary

dpc:
Just like what we did with `fedimintd`, this makes `fedimint-cli` crate contain both a library and a binary. A binary uses default module only, but 3rd party developers can use it to create fedimint-cli with extra modules.

I plan to follow it up with a change that will add `ClientModuleGen` cli handler method, so that:

```
fedimint-cli module <id/kind> <args>
```

can be issued from the command line.


justinmoon (approved):
LGTM

PR #2029, opened 2023-03-24
Second iteration on the config gen API

jkitman:
Second iteration on the config gen API.

Adds the ability of peers to send their server connections to a leader and then the leader to return the list of peers that have connected.

Related to https://github.com/fedimint/fedimint/issues/1893 https://github.com/fedimint/fedimint/issues/1628

### DIFF
--- a/fedimint-core/src/admin_client.rs
+++ b/fedimint-core/src/admin_client.rs
@@ -1,0 +1,80 @@
+use std::fmt::Debug;
+
+use fedimint_core::task::MaybeSend;
+use serde::{Deserialize, Serialize};
+use tokio_rustls::rustls;
+use url::Url;
+
+use crate::api::{DynFederationApi, FederationApiExt, FederationResult, WsFederationApi};
+use crate::module::{ApiAuth, ApiRequestErased};
+use crate::PeerId;
+
+/// For a guardian to communicate with their server
+// TODO: Maybe should have it's own CLI client so it doesn't need to be in core
+pub struct WsAdminClient {
+    inner: DynFederationApi,
+    auth: ApiAuth,
+}
+
+impl WsAdminClient {
+    pub fn new(url: Url, our_id: PeerId, auth: ApiAuth) -> Self {
+        Self {
+            inner: WsFederationApi::new(vec![(our_id, url)]).into(),
+            auth,
+        }
+    }
+
+    pub async fn set_password(&self, auth: ApiAuth) -> FederationResult<()> {
+        self.request("/set_password", ApiRequestErased::new(auth))
+            .await
+    }
+
+    pub async fn signal_upgrade(&self) -> FederationResult<()> {
+        self.request("/upgrade", ApiRequestErased::default()).await
+    }
+
+    async fn request<Ret>(&self, method: &str, params: ApiRequestErased) -> FederationResult<Ret>
+    where
+        Ret: serde::de::DeserializeOwned + Eq + Debug + Clone + MaybeSend,
+    {
+        self.inner
+            .request_current_consensus(method.to_owned(), params.with_auth(&self.auth))
+            .await
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PeerServerParams {
+    #[serde(with = "serde_tls_cert")]
+    pub cert: rustls::Certificate,
+    pub p2p_url: Url,
+    pub api_url: Url,
+    pub name: String,
+}
+
+mod serde_tls_cert {
+    use std::borrow::Cow;
+
+    use bitcoin_hashes::hex::{FromHex, ToHex};
+    use serde::de::Error;
+    use serde::{Deserialize, Deserializer, Serializer};
+    use tokio_rustls::rustls;
+
+    pub fn serialize<S>(certs: &rustls::Certificate, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let hex_str = certs.0.to_hex();
+        serializer.serialize_str(&hex_str)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<rustls::Certificate, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value: Cow<str> = Deserialize::deserialize(deserializer)?;
+        Ok(rustls::Certificate(
+            Vec::from_hex(&value).map_err(D::Error::custom)?,
+        ))
+    }
+}
--- a/fedimint-core/src/admin_client.rs
+++ b/fedimint-core/src/admin_client.rs
@@ -24,16 +24,81 @@ impl WsAdminClient {
         }
     }
 
+    /// Sets the password used to decrypt the configs and authenticate
+    ///
+    /// Must be called first before any other calls to the API
     pub async fn set_password(&self, auth: ApiAuth) -> FederationResult<()> {
-        self.request("/set_password", ApiRequestErased::new(auth))
+        self.request_auth("/set_password", ApiRequestErased::new(auth))
             .await
     }
 
-    pub async fn signal_upgrade(&self) -> FederationResult<()> {
-        self.request("/upgrade", ApiRequestErased::default()).await
+    /// During config gen, sets the server connection containing our endpoints
+    ///
+    /// Optionally sends our server info to the config gen leader using
+    /// `add_config_gen_peer`
+    pub async fn set_config_gen_connections(
+        &self,
+        info: ConfigGenConnectionsRequest,
+    ) -> FederationResult<()> {
+        self.request_auth("/set_config_gen_connections", ApiRequestErased::new(info))
+            .await
     }
 
-    async fn request<Ret>(&self, method: &str, params: ApiRequestErased) -> FederationResult<Ret>
+    /// During config gen, used for an API-to-API call that adds a peer's server
+    /// connection info to the leader.
+    ///
+    /// Note this call will fail until the leader has their API running and has
+    /// `set_server_connections` so clients should retry.
+    ///
+    /// This call is not authenticated because it's guardian-to-guardian
+    pub async fn add_config_gen_peer(&self, peer: PeerServerParams) -> FederationResult<()> {
+        self.inner
+            .request_current_consensus(
+                "/add_config_gen_peer".to_owned(),
+                ApiRequestErased::new(peer),
+            )
+            .await
+    }
+
+    /// During config gen, gets all the server connections we've received from
+    /// peers using `add_config_gen_peer`
+    ///
+    /// Could be called on the leader, so it's not authenticated
+    pub async fn get_config_gen_peers(&self) -> FederationResult<Vec<PeerServerParams>> {
+        self.inner
+            .request_current_consensus(
+                "/get_config_gen_peers".to_owned(),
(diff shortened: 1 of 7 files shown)


fedimint-server/src/config/api.rs
```diff
@@ -37,27 +47,156 @@ impl ConfigApi {
...
+                ConfigApiState::SetConfigGenParams(old) => old.with_request(request),
+                _ => return Err(ApiError::bad_request("Config already set".to_string())),
+            };
+            *state = ConfigApiState::SetConfigGenParams(connection.clone());
+
+            connection
+        };
+
+        if let Some(url) = connection.request.leader_api_url.clone() {
+            // Note PeerIds don't really exist at this point, but id doesn't matter
```
dpc, line 80: This comment should hit a gym, cuz it's not in the best shape.

fedimint-server/src/config/api.rs
```diff
@@ -37,27 +47,156 @@ impl ConfigApi {
...
-            Ok(())
+        match &*state {
+            ConfigApiState::SetPassword => *state = ConfigApiState::SetConnections(auth),
+            _ => return Err(ApiError::bad_request("Password already set".to_string())),
+        }
+
+        Ok(())
+    }
+
+    // Sets our connection info, possibly sending it to a leader
```
dpc, line 61: `///`

dpc (dismissed):

fedimint-server/src/config/api.rs
```diff
@@ -37,27 +47,157 @@ impl ConfigApi {
...
+            };
+            *state = ConfigApiState::SetConfigGenParams(connection.clone());
+
+            connection
+        };
+
+        if let Some(url) = connection.request.leader_api_url.clone() {
+            // Note PeerIds don't really exist at this point, but id doesn't matter because
+            // it's not used in the WS client for anything, perhaps it should be removed
+            let client = WsAuthenticatedClient::new(url, PeerId::from(0), connection.auth.clone());
```
elsirion, line 82: This sounds dangerous, why would we send *our* auth data to the leader? They can't verify it and could log into our guardian later? We could have a psk that is required to connect, but that's not what this seems to be and that should likely be a separate PR.

fedimint-server/src/config/api.rs
```diff
@@ -37,27 +47,157 @@ impl ConfigApi {
...
+            _ => return Self::bad_request("Not ready to receive peer server params"),
+        };
+        connection.peers.insert(peer.cert.clone(), peer);
+        *state = ConfigApiState::SetConfigGenParams(connection);
+
+        Ok(())
+    }
+
+    /// Returns the peers that have called `add_config_gen_peer` on the leader,
+    /// only the leader will need to call this
```
elsirion, line 108: Should this be a subscription based API call? Like, give me all `PeerServerParams` every time they change? That way everyone sees the others connect in real time. There should be another endpoint possibly that returns the full params once the leader clicks a buttons indicating everyone who was expected connected.

fedimint-server/src/config/api.rs
```diff
@@ -37,27 +47,157 @@ impl ConfigApi {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ConfigGenConnectionsRequest {
+    /// Our guardian name
+    our_name: String,
+    /// Bind address for our P2P connection
+    our_p2p_bind: SocketAddr,
+    /// Url for our p2p connection
+    our_p2p_url: Url,
+    /// Url for our API connection
+    our_api_url: Url,
```
elsirion, line 135: Aren't these defined through env vars that aren't available to the client? That would mean `fedimintd` should inject them before sending the info to the leader instead of letting the user enter them.

elsirion (changes requested):

elsirion:
Looks good overall, but some problems worth addressing.

client/cli/Cargo.toml
```diff
@@ -29,6 +29,7 @@ fedimint-rocksdb = { path = "../../fedimint-rocksdb" }
 fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
 fedimint-ln-client = { path = "../../modules/fedimint-ln-client" }
 fedimint-logging = { path = "../../fedimint-logging" }
+fedimint-server = { path = "../../fedimint-server" }
```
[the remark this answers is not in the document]
jkitman: Even a client only used for server admin?  Perhaps it needs to be a different client CLI?

fedimint-server/src/config/api.rs
jkitman: We could do it that way, but if they need to be changed the user needs to login to the `fedimintd` server or redeploy it.  We should assume guardians are non-technical.  Seems easier to configure things via the UI unless there are reasonable defaults.

fedimint-server/src/config/api.rs
jkitman: Yes there will be an API endpoint to return all the configs.  This endpoint I thought would only be used by the leader for displaying in their UI.  The followers would only see that the leader is online and accepted their request, then they would wait for the leader to configure the rest of the config gen params.

What is a good example of the subscription based API call?

fedimint-server/src/config/api.rs
jkitman: Good point, I'll remove the auth from that call.

client/cli/Cargo.toml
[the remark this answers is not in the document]
elsirion: I'd say it still belongs in common, why would such a client care about e.g. consensus, and that's what pulling in `fedimint-server` means.

fedimint-server/src/config/api.rs
elsirion: There are a few endpoints @Maan2003 refactored to wait for a certain event to occur, although none of these is subscription-based yet afaik. E.g. waiting for a transaction to be confirmed or rejected:

https://github.com/fedimint/fedimint/blob/cf9917ef806e19890d0073ddb8e9efe7b6c48b65/fedimint-server/src/net/api.rs#L228-L239

You couldn't rely on the DB and would probably need to use [`Notify`](https://dtantsur.github.io/rust-openstack/tokio/sync/struct.Notify.html) directly.

fedimint-server/src/config/api.rs
elsirion: There are a few endpoints @Maan2003 refactored to wait for a certain event to occur, although none of these is subscription-based yet afaik. E.g. waiting for a transaction to be confirmed or rejected:

https://github.com/fedimint/fedimint/blob/cf9917ef806e19890d0073ddb8e9efe7b6c48b65/fedimint-server/src/net/api.rs#L228-L239

You couldn't rely on the DB and would probably need to use [`Notify`](https://dtantsur.github.io/rust-openstack/tokio/sync/struct.Notify.html) directly.

fedimint-server/src/config/api.rs
elsirion: I mean you need to configure your reverse proxy anyway unless it comes pre-packaged (in which case the "packager" can make sure the env vars are set correctly). @dpc is the expert for these ops questions I think, I mainly want to be consistent about it.

client/cli/Cargo.toml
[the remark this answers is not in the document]
jkitman: For now I'll move the structs into core, but add a comment maybe we should have a new CLI for it.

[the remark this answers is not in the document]
elsirion: I don't mind in which CLI it lives, we might want an `fedimint-cli admin <command>` subcommand, but if you think a separate one might be useful feel free to go with that (just seems like needless sprawl).

fedimint-server/src/config/api.rs
```diff
@@ -9,13 +10,22 @@ use fedimint_core::db::Database;
...
+use tokio_rustls::rustls;
+use url::Url;
 
+use crate::config::{gen_cert_and_key, PeerServerParams};
 use crate::net::api::{attach_endpoints, HasApiContext, RpcHandlerCtx};
+use crate::net::client::WsAuthenticatedClient;
 
-/// Serves the config API endpoints
-pub struct ConfigApi {
+pub type Result<T> = std::result::Result<T, ApiError>;
```
mxz42, line 25: nit: maybe call this `ApiResult<T>` to make it distinguishable from the std `Result`? and then it's logical that the `ApiResult` returns an `ApiError`

elsirion: I don't think that's common, see e.g. `std::fmt::Result`, is there a Rust style guide to consult?

fedimint-server/src/config/api.rs
jkitman: If we let peers call this and want it to block I'm not sure the best way to implement...I guess we'd need a `Map<TlsCert, Notify>` and require peers to pass in their cert?  I'm tempted to leave this for now until we know how the UI will be calling it.

fedimint-server/src/config/api.rs
mxz42: I don't know about a style guide. With a quick search I found [`MathResult` in Rust by example](https://doc.rust-lang.org/rust-by-example/std/result.html) or [`LockResult` in std](https://doc.rust-lang.org/std/sync/type.LockResult.html). 
However, now while writing this, I see that these types always have a concrete type `T`; as does `std::fmt::Result`

fedimint-server/src/config/api.rs
jkitman: Would the guardians be recommended to use DNS?  I figure you'd bind to default ports, setup a DNS provider, and enter your URL in the UI.

fedimint-server/src/config/api.rs
elsirion: I think one notify would be good enough? I was assuming we want to give at least the leader a live view of who is connected, so instead of polling this fn would return each time someone connects.

One possible design:
1. client sends how many peers they are aware of
2. server subscribes to notify
3. server checks if current number of peers is more than what client said, return peers to client if so
4. server waits for notification, then jump to 3

fedimint-server/src/config/api.rs
elsirion: > Would the guardians be recommended to use DNS?

Probably as long as we don't have a Tor integration.

jkitman: I'll change it to be passed in for now, we can change it later.

jkitman:
@elsirion Think I've addressed your feedback now.

fedimint-core/src/admin_client.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub async fn signal_upgrade(&self) -> FederationResult<()> {
+        self.request("/upgrade", ApiRequestErased::default()).await
+    }
+
+    async fn request<Ret>(&self, method: &str, params: ApiRequestErased) -> FederationResult<Ret>
+    where
+        Ret: serde::de::DeserializeOwned + Eq + Debug + Clone + MaybeSend,
+    {
+        self.inner
+            .request_current_consensus(method.to_owned(), params.with_auth(&self.auth))
```
elsirion, line 106: nit: should be at least preceded by an assert that there is only one connected node. Long term the admin client should probably not re-use the normal client to reduce risk of leaking auth info to other guardians.

elsirion (approved):

dpc (approved):

PR #2030, opened 2023-03-24
feat: ability for fedimint-cli to call client modules

dpc:


client/cli/src/lib.rs
```diff
@@ -242,29 +245,51 @@ impl Opts {
...
+        module_gens: &ClientModuleGenRegistry,
+    ) -> ModuleDecoderRegistry {
+        ModuleDecoderRegistry::new(cfg.clone().0.modules.into_iter().filter_map(
             |(id, module_cfg)| {
-                let module_gen = module_gens
+                module_gens
                     .get(module_cfg.kind())
-                    .expect("module kind not found in registry");
-                (id, module_gen.as_ref().decoder())
+                    .map(|module_gen| (id, module_gen.as_ref().decoder()))
```
justinmoon, line 276: Nice, this makes it so unknown modules just get ignored instead of panicing

client/cli/src/lib.rs
```diff
@@ -953,6 +1025,37 @@ impl FedimintCli {
                     .await?;
                 Ok(CliOutput::EpochCount { count })
             }
+            Command::Module { id, arg } => {
+                let cfg = cli.load_config_or_terminate();
+                let decoders = cli.load_decoders(&cfg, &self.module_gens);
+                let db = cli.load_db_or_terminate(&decoders);
+                let (_id, module_cfg) = match id {
+                    ModuleSelector::Id(id) => (
+                        id,
```
justinmoon, line 1034: If this isn't used, maybe don't return it?

justinmoon (approved):
one small nit

client/cli/src/lib.rs
dpc: Then I have to do an extra `.map` to get rid of it on the other case, so this is actually shorter. :D

client/cli/src/lib.rs
```diff
@@ -953,6 +1025,37 @@ impl FedimintCli {
...
+                        cfg.as_ref()
+                            .get_module_cfg(id)
+                            .or_terminate_msg(CliErrorKind::IOError, "Can't load module"),
+                    ),
+                    ModuleSelector::Kind(kind) => cfg
+                        .as_ref()
+                        .get_first_module_by_kind_cfg(kind)
```
jkitman, line 1041: So is getting by kind just to make the command more user-friendly?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -23,7 +23,7 @@ impl ClientModuleGen for MintClientGen {
     type Config = MintClientConfig;
 
     async fn init(&self, _cfg: Self::Config, _db: Database) -> anyhow::Result<Self::Module> {
-        unimplemented!()
+        Ok(MintClientModule {})
```
jkitman, line 26: Why only implement for the mint module?

client/cli/src/lib.rs
dpc: Yeap. I suspect in majority of cases people will have one module instance of a given kind, so referring to it by kind does seem to make things smoother ... .

We could possibly change it to be "by kind, only if it is not ambiguous" or something.

PR #2039, opened 2023-03-27
Fix wallet round consensus

### LINKED ISSUE #2011: Wallet round consensus multiple submissions bug
> Before throwing away the peer id, do we make sure that each peer only contributed one item? Otherwise malicious ones could take over. Needs to be fixed in wallet too.

_Originally posted by @elsirion in https://github.com/fedimint/fedimint/pull/1074#discussion_r1146335734_

jkitman:
Fixes #2011

### DIFF
--- a/modules/fedimint-wallet-server/Cargo.toml
+++ b/modules/fedimint-wallet-server/Cargo.toml
@@ -42,3 +42,4 @@ fedimint-server = { path = "../../fedimint-server" }
 fedimint-testing = { path = "../../fedimint-testing" }
 test-log = { version = "0.2", features = [ "trace" ], default-features = false }
 tracing-subscriber = { version = "0.3.16", features = [ "env-filter" ] }
+tokio = { version = "1.26.0", features = ["full"] }
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -370,34 +370,10 @@ impl ServerModule for Wallet {
         // Save signatures to the database
         self.save_peg_out_signatures(dbtx, peg_out_signatures).await;
 
-        // FIXME: also warn on less than 1/3, that should never happen
-        // Make sure we have enough contributions to continue
-        if round_consensus.is_empty() {
-            panic!("No proposals were submitted this round");
-        }
-
-        let fee_proposals = round_consensus.iter().map(|(_, rc)| rc.fee_rate).collect();
-        let fee_rate = self.process_fee_proposals(fee_proposals).await;
-
-        let height_proposals = round_consensus
-            .iter()
-            .map(|(_, rc)| rc.block_height)
-            .collect();
-        let block_height = self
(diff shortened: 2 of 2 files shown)


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -894,43 +861,51 @@ impl Wallet {
...
+            if dedup.insert(peer, item).is_some() {
+                warn!("Peer {} submitted multiple RoundConsensusItem", peer);
+            }
+        }
+        let items: Vec<_> = dedup.values().into_iter().collect();
 
-        let consensus_height = self.consensus_height(dbtx).await.unwrap_or(0);
+        let mut fees: Vec<_> = items.iter().map(|item| item.fee_rate).collect();
+        fees.sort_unstable();
+        let fee_rate = *fees.get(fees.len() / 2).expect("is non-empty");
```
m1sterc001guy, line 890: Nit: The error message on expect could be more informative here

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1582,12 +1557,61 @@ mod tests {
...
+        // removes duplicate entries for a given peer, keeping last entry
+        let consensus = Wallet::process_round_consensus(
+            0,
+            vec![
+                (PeerId::from(0), round_item(1, 4, 7)),
+                (PeerId::from(1), round_item(2, 5, 8)),
+                (PeerId::from(1), round_item(3, 6, 9)),
+            ],
+        );
+        assert_eq!(consensus, round_consensus(3, 6, 7 ^ 9));
```
m1sterc001guy, line 1613: For the block height and fees here, does the median calculation default to using a specific entry rather than averaging the numbers? (for instance, even when discounting Peer 1's first `RoundConsensusItem`, shouldn't the block height be an average of 1 and 3?)

jkitman: It's a median, so you throw out the duplicate, divide the number of items by 2 (`2/2=1`) so you get the last item.

elsirion: I think there are different definitions of medians, but this one was good enough when I thought it through.

modules/fedimint-wallet-server/Cargo.toml
```diff
@@ -42,3 +42,4 @@ fedimint-server = { path = "../../fedimint-server" }
 fedimint-testing = { path = "../../fedimint-testing" }
 test-log = { version = "0.2", features = [ "trace" ], default-features = false }
 tracing-subscriber = { version = "0.3.16", features = [ "env-filter" ] }
+tokio = { version = "1.26.0", features = ["full"] }
```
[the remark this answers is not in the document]
jkitman: If you want run the tests only in that crate, they require tokio.

m1sterc001guy (approved):
LGTM

PR #2041, opened 2023-03-27
chore: add log message when setup UI is running

m1sterc001guy:
Currently there are not many logs when running `fedimintd` with `--listen-ui` which makes it difficult to know if it is running correctly. This PR slightly modifies the log message to make it easy to tell that the binary is listening on a `SocketAddr`.


dpc (approved):
https://goat.com.au/wp-content/uploads/2019/10/thank-you.gif

justinmoon (approved):
Very nice.

Issue #1995, opened 2023-03-22
Tmuxinator should run a gateway on the LND node, too


justinmoon:
We forgot to add [this](https://github.com/fedimint/fedimint/pull/1719/files#diff-9233a92d25c448c1e76d944bea5aa907318e0754477f53da48ee4df5f1807f9dR49) back when cherry-picking from https://github.com/fedimint/fedimint/pull/1719.

This way you can send lightning payments through either node.

dhruv-1001:
I'll take on this one. Please assign them both to me. Thanks!

Just to clear this up, right now there is only one `gateway` using `cln`, and it have a channel with external `lnd` node. This issue expects to add another `gateway` to the federation that uses `lnd` and have a channel with same external `lnd` node. Am I missing something?

justinmoon:
Add another pane to tmuxinator. Run `gatewayd` and connect it to `lnd` by passing [this subcommand](https://github.com/fedimint/fedimint/blob/55622707fd08e9a1dd1dda9db5da81b2f1b660ec/gateway/ln-gateway/src/bin/gatewayd.rs#L49-L62)

justinmoon:
You'll also want to add the `switch_to_lnd_gateway` / `switch_to_cln_gateway` bash function https://github.com/fedimint/fedimint/pull/1719.

justinmoon:
When you run `switch_to_lnd_gateway` / `switch_to_cln_gateway`, try running `fedimint-cli list-gateways` to see what changes. Ask any questions on discord!

justinmoon:
Just realized the cli-tests don't run a LND gateway either :(

m1sterc001guy:
This PR opens up a API for manually switching which Lightning node `gatewayd` will use without restarting `gatewayd` https://github.com/fedimint/fedimint/pull/2019

justinmoon:
Sorry I took this one over @dhruv-1001. I was debugging gatewayd and noticed that it wasn't working at all for incoming LND payments, and in the process of fixing that I pretty much had this issue all fixed.

dhruv-1001:
No problem @justinmoon. I learned a few things while working on it, so it's all good. 🚀

PR #1910, opened 2023-03-14
Replace Bash with Rust 1

Maan2003:
Rewrite bash scripts to rust. 

Remaining work:

- [x] Tmuxinator
- [ ] Export $FM_LOGS_DIR in github actions
- [x] Remove lightning cli commands and use rpc
- [x] Debug why tests are failing in CI (easier after 1)

in other PRs:
- [ ] Port rest of tests
- [ ] Move most of build.sh to rust.
- [ ] Restore tmux restart
- [ ] Port pre-commit hook to rust

PR #2042, opened 2023-03-27
feat: fixtures can spawn LND gateway

justinmoon:
Step towards https://github.com/fedimint/fedimint/issues/1995


justinmoon:
Actually just going to close this. Better to fix in https://github.com/fedimint/fedimint/pull/1910.

PR #2044, opened 2023-03-27
Api config v3

jkitman:
Next iteration of the API, this one enables the leader to configure the final config gen params and send them to followers.

After this we should be able to run DKG.

### DIFF
--- a/fedimint-core/src/admin_client.rs
+++ b/fedimint-core/src/admin_client.rs
@@ -1,6 +1,7 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Debug;
 
+use bitcoin_hashes::sha256;
 use fedimint_core::task::MaybeSend;
 use serde::{Deserialize, Serialize};
 use tokio_rustls::rustls;
@@ -30,7 +31,7 @@ impl WsAdminClient {
     ///
     /// Must be called first before any other calls to the API
     pub async fn set_password(&self, auth: ApiAuth) -> FederationResult<()> {
-        self.request_auth("/set_password", ApiRequestErased::new(auth))
+        self.request_auth("set_password", ApiRequestErased::new(auth))
             .await
     }
 
@@ -42,7 +43,7 @@ impl WsAdminClient {
         &self,
         info: ConfigGenConnectionsRequest,
     ) -> FederationResult<()> {
-        self.request_auth("/set_config_gen_connections", ApiRequestErased::new(info))
+        self.request_auth("set_config_gen_connections", ApiRequestErased::new(info))
             .await
     }
 
@@ -54,7 +55,7 @@ impl WsAdminClient {
     ///
     /// This call is not authenticated because it's guardian-to-guardian
     pub async fn add_config_gen_peer(&self, peer: PeerServerParams) -> FederationResult<()> {
-        self.request("/add_config_gen_peer", ApiRequestErased::new(peer))
+        self.request("add_config_gen_peer", ApiRequestErased::new(peer))
             .await
     }
 
@@ -63,7 +64,7 @@ impl WsAdminClient {
     ///
     /// Could be called on the leader, so it's not authenticated
     pub async fn get_config_gen_peers(&self) -> FederationResult<Vec<PeerServerParams>> {
-        self.request("/get_config_gen_peers", ApiRequestErased::default())
+        self.request("get_config_gen_peers", ApiRequestErased::default())
             .await
     }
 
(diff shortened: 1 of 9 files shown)


fedimint-server/src/config/mod.rs
```diff
@@ -471,9 +472,7 @@ impl ServerConfig {
         // initially, where we consider "module as a code" as "module as an instance at
         // runtime"
         let null_config_gen = ConfigGenParams::null();
-        for (module_instance_id, (kind, gen)) in registry.legacy_init_order_iter().enumerate() {
```
jkitman, line 474: Had to change this so that we can test `distributed_gen` without modules.  Long-term we need to move towards this for modularization anyway.

`MODULE_INSTANCE_ID_DKG_DONE` is introduced because without modules serving to synchronize the global messages, we could send a done message before DKG completes.

fedimint-core/src/admin_client.rs
```diff
@@ -94,6 +85,64 @@ impl WsAdminClient {
             .await
     }
 
+    /// Gets the default config gen params which can be configured by the
+    /// leader, gives them a template to modify
+    pub async fn get_default_config_gen_params(&self) -> FederationResult<ConfigGenParamsRequest> {
+        self.request_auth(
+            "/get_default_config_gen_params",
```
dpc, line 92: Not something we need to address here, but we really need to get rid of that `/`. These are all jsonrpc calls, and not urls, so why are we doing slashes? :D

Also if we are to keep with the convnentions that bitcoind or LN nodes use, we should probably name `getdefaultconfiggenparams`? I know it looks less readable, but bitcoin uses `getblockcount` and not `get_block_count`.

Ping @Maan2003 for awareness and to tell me in case I'm wrong.

fedimint-core/src/admin_client.rs
```diff
@@ -94,6 +85,64 @@ impl WsAdminClient {
...
+    /// error and config gen must be restarted.
+    pub async fn await_run_dkg(&self) -> FederationResult<ConfigGenParamsConsensus> {
+        self.request_auth("/await_run_dkg", ApiRequestErased::default())
+            .await
+    }
+
+    /// After DKG, returns the hash of the consensus config tweaked with our id.
+    /// We need to share this with all other peers to complete verification.
+    pub async fn get_config_verification(&self) -> FederationResult<sha256::Hash> {
+        self.request_auth("/get_config_verification", ApiRequestErased::default())
```
dpc, line 132: "returns the hash of the consensus config tweaked with our id" so shouldn't it be "get tweaked config hash" or something like that?

fedimint-core/src/admin_client.rs
Maan2003: Agreed, lets not add new `/`s.

fedimint-core/src/admin_client.rs
```diff
@@ -94,6 +85,64 @@ impl WsAdminClient {
...
+            .await
+    }
+
+    /// After we exchange verification hashes with other peers, we call this to
+    /// confirm we all have the same consensus configs.
+    pub async fn verify_configs(
+        &self,
+        user_hashes: BTreeSet<sha256::Hash>,
```
dpc, line 140: I don't get it - if everyone are supposed to have the same hash, shouldn't it be just one `Hash`? And if they are not - doesn't it need to know which peer id is which (in case some peer sent a duplicate or something?). If not, please add comment why it's OK / safe for the uninformed.

fedimint-core/src/admin_client.rs
jkitman: Hmm maybe verification hash?  The tweak is kind of an implementation detail, it's just a hash that must be shared with other guardians.

dpc (commented):
LGTM, just some naming comments.

dpc (approved):

fedimint-core/src/config.rs
```diff
@@ -272,6 +284,40 @@ pub type CommonModuleGenRegistry = ModuleGenRegistry<DynCommonModuleGen>;
...
+    {
+        let json: BTreeMap<ModuleKind, serde_json::Value> = Deserialize::deserialize(deserializer)?;
+        let mut params = BTreeMap::new();
+
+        for (key, value) in json {
+            params.insert(key, ConfigGenParams(value));
+        }
+        Ok(ModuleGenRegistry(params))
+    }
+}
```
Maan2003, line 319: why not separately impl serde for `ModuleGenRegistry<T>` and `ConfigGenParams`

jkitman: I guess I could implement it for `ModuleGenRegistry<T: Serialize>`, but since this is the only place that uses it right now I didn't think it was necessary.

Maan2003 (approved):

PR #2046, opened 2023-03-28
feat: add `federation_id` filter and details on gateway-cli info

### LINKED ISSUE #1725: Implement filtering and details on gatewayd info API
Today, `gateway-cli info` returns high level information on all connected federations. Consider supporting filters on gatway info api, so we can use cli or api call to get information about a specific federation served by the gateway.

**Proposal:**
- `gateway-cli info` continues to **return high level info** about all connected federations
- `gateway-cli info <federation-id>` should **return detailed info** about the federation identified. 
  - Details might include balance, summary of transactions, etc

oleonardolima:
## Description
The gatewayd info currently returns only high-level information related to the gateway and all federations, as discussed in #1725 it would be useful to have a way to get the information related to a specific federation, therefore filtering the information by a `federation_id`.

That said, this PR aims to add this feature by making it possible to get gatewayd info filtered for a federation, when passing a federation_id.

## How
This PR adds:
- gateway-cli: adds a new `FederationId` optional parameter for the `Info` command.
- rpc-client: updates the `get_info` to expect an optional `FederationId` parameter, and calls the rpc-server with a `QueryParam` as `?federation_id=` when it has been passed, with no query param otherwise.
- rpc-server: updates the `/info` route handler to use the `QueryParam` axum extractor, it handles both when param, empty or none param is passed as guided here: https://github.com/tokio-rs/axum/blob/main/examples/query-params-with-empty-strings/src/main.rs
- ln-gateway: updates the `handle_info` to expect a `FederationId` field on the `InfoPayload` parameter, filters the selected actor, and use a new `get_detailed_info` `GatewayActor method if a federation_id is passed, old behavior otherwise.
- **_wip(integration-test):_** adds a new test for getting detailed info, when passing a federation_id.

fixes #1725

### DIFF
--- a/gateway/cli/Cargo.toml
+++ b/gateway/cli/Cargo.toml
@@ -16,10 +16,10 @@ axum = "0.6.4"
 axum-macros = "0.3.1"
 bitcoin = { version = "0.29.2", features = ["serde"] }
 clap = { version = "4.1.6", features = ["derive", "std", "help", "usage", "error-context", "suggestions"], default-features = false }
-ln-gateway = { path= "../ln-gateway" }
 fedimint-client-legacy = { path = "../../fedimint-client-legacy" }
-fedimint-core ={ path = "../../fedimint-core" }
+fedimint-core = { path = "../../fedimint-core" }
 fedimint-logging = { path = "../../fedimint-logging" }
+ln-gateway = { path= "../ln-gateway" }
 reqwest = { version = "0.11.14", features = [ "json", "rustls-tls" ], default-features = false }
 rpassword = "7.2.0"
 serde = { version = "1.0", features = ["derive"] }
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -389,12 +389,23 @@ impl Gateway {
         Ok(federation_info)
     }
 
-    async fn handle_get_info(&self, _payload: InfoPayload) -> Result<GatewayInfo> {
+    async fn handle_get_info(&self, payload: InfoPayload) -> Result<GatewayInfo> {
         let actors = self.actors.read().await;
-        let mut federations: Vec<FederationInfo> = Vec::new();
+        let mut all_federations: Vec<FederationInfo> = Vec::new();
         for actor in actors.values() {
-            federations.push(actor.get_info()?);
+            all_federations.push(actor.get_info()?);
         }
+
+        let federations = payload
+            .federation_id
+            .map(|id| {
+                all_federations
+                    .iter()
+                    .filter(|info| info.federation_id.eq(&id))
+                    .cloned()
+                    .collect::<Vec<_>>()
+            })
+            .unwrap_or(all_federations);
 
         let ln_info = self.lnrpc.read().await.info().await?;
 
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -336,7 +336,7 @@ impl Gateway {
 
         if let Ok(actor) = self.select_actor(connect.id).await {
             info!("Federation {} already connected", connect.id);
-            return actor.get_info();
+            return actor.get_summary_info();
         }
 
         let GetNodeInfoResponse { pub_key, alias: _ } = self.lnrpc.read().await.info().await?;
@@ -384,28 +384,27 @@ impl Gateway {
             );
         }
 
-        let federation_info = actor.get_info()?;
+        let federation_info = actor.get_summary_info()?;
 
         Ok(federation_info)
     }
 
     async fn handle_get_info(&self, payload: InfoPayload) -> Result<GatewayInfo> {
-        let actors = self.actors.read().await;
-        let mut all_federations: Vec<FederationInfo> = Vec::new();
-        for actor in actors.values() {
-            all_federations.push(actor.get_info()?);
-        }
-
-        let federations = payload
-            .federation_id
-            .map(|id| {
-                all_federations
-                    .iter()
-                    .filter(|info| info.federation_id.eq(&id))
-                    .cloned()
-                    .collect::<Vec<_>>()
-            })
-            .unwrap_or(all_federations);
+        let mut federations: Vec<FederationInfo> = Vec::new();
+        match payload.federation_id {
+            Some(federation_id) => federations.push(
+                self.select_actor(federation_id)
+                    .await?
+                    .get_detailed_info()
+                    .await?,
+            ),
+            None => {
+                let actors = self.actors.read().await;
+                for actor in actors.values() {
+                    federations.push(actor.get_summary_info()?);
+                }
+            }
+        };
 
         let ln_info = self.lnrpc.read().await.info().await?;
 
--- a/gateway/ln-gateway/src/rpc/mod.rs
+++ b/gateway/ln-gateway/src/rpc/mod.rs
@@ -122,6 +122,19 @@ pub struct FederationInfo {
     pub federation_id: FederationId,
     /// Information we registered with the fed
     pub registration: LightningGateway,
+    /// The current federation balance
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub balance: Option<Amount>,
+}
+
+impl FederationInfo {
+    pub fn new(federation_id: FederationId, registration: LightningGateway) -> Self {
+        FederationInfo {
+            federation_id,
+            registration,
+            balance: None,
+        }
+    }
 }
 
 #[derive(Debug, Serialize, Deserialize)]
--- a/gateway/ln-gateway/src/rpc/mod.rs
+++ b/gateway/ln-gateway/src/rpc/mod.rs
@@ -65,7 +65,9 @@ pub struct ConnectFedPayload {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-pub struct InfoPayload;
+pub struct InfoPayload {
+    pub federation_id: Option<FederationId>,
+}
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct BackupPayload {
--- a/gateway/ln-gateway/src/rpc/rpc_client.rs
+++ b/gateway/ln-gateway/src/rpc/rpc_client.rs
@@ -1,6 +1,7 @@
 use std::result::Result;
 
 use bitcoin::Address;
+use fedimint_core::config::FederationId;
 use fedimint_core::{Amount, TransactionId};
 use reqwest::StatusCode;
 pub use reqwest::{Error, Response};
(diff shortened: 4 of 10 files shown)


gateway/cli/Cargo.toml
okjodom, line 1: Not necessary, but very nice 👌🏿 
I'd only be wary of this change to sort deps resulting in many merge conflicts for you to resolve

gateway/ln-gateway/src/lib.rs
```diff
@@ -251,19 +252,36 @@ impl Gateway {
...
-        })
+        match payload.federation_id {
+            Some(id) => {
+                // TODO: (@oleonardolima) Improve the filter by federation_id
+                let filtered_federations = federations
+                    .iter()
+                    .filter(|federation_info| federation_info.federation_id == id)
+                    .cloned()
+                    .collect::<Vec<_>>();
+                // let filtered_federations = federations
```
dpc, line 272: Leftover.

gateway/ln-gateway/src/lib.rs
```diff
@@ -251,19 +252,36 @@ impl Gateway {
...
+                //     .drain_filter(|federation_info| federation_info.federation_id == id)
+                //     .collect::<Vec<_>>();
+                Ok(GatewayInfo {
+                    federations: filtered_federations,
+                    version_hash: env!("CODE_VERSION").to_string(),
+                })
+            }
+            None => Ok(GatewayInfo {
```
dpc, line 280: You could probably take the `Ok(GatewayInfo( { ... , version_hash: env!("CODE_VERSION").to_string(), })}` outside of the `match` to make it more compact, then use `payload.federation_id.map(|id| ... ).unwrap_or(federations)` to get rid of `match` altogether. IMO that's nicer, but one might argue.

gateway/ln-gateway/src/lib.rs
```diff
@@ -251,15 +252,26 @@ impl Gateway {
...
             .iter()
             .map(|(_, actor)| actor.get_info().expect("Failed to get actor info"))
             .collect();
 
+        let federations = payload
+            .federation_id
+            .map(|id| {
+                all_federations
+                    .iter()
+                    .filter(|info| info.federation_id.eq(&id))
```
dpc, line 269: Just occurred to me - can there be more federations with the same ID? If not, we should probably use `find` instead of `filter` and respond with an error if there's no match, no?

justinmoon: It's a threshold public key, right? I guess it's possible if someone just took configs and updated networking and ran a second instance with the same configs ... cc @jkitman

jkitman: Yeah it's the threshold pubkey so it should be unique to the federation.  The gateway should probably validate the id before connecting to a new federation to prevent attackers from spoofing as the federation, this should be possible already using the client config.

gateway/ln-gateway/src/lib.rs
oleonardolima: I updated it to filter the actor beforehand and call another method, so I went back for a match clause.

Do you think that it could be refactored somehow ?

gateway/ln-gateway/src/lib.rs
oleonardolima: Thanks, guys! I guess it won't be a problem anymore as I updated it to get a selected actor by the given federation_id.

Do you think that might still be a problem?

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -110,10 +112,23 @@ pub struct WithdrawPayload {
     pub address: Address,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct FederationInfo {
     pub federation_id: FederationId,
     pub mint_pubkey: XOnlyPublicKey,
+    pub balance: Option<Amount>,
+}
```
oleonardolima, line 120: @okjodom I got stuck on a better way for this, do you have anything in mind ?

I was thinking that using something like [serde_with](https://docs.rs/serde_with/2.3.1/serde_with/attr.skip_serializing_none.html) would help to keep a single struct, but I'm not sure if adding a new dependency would be a bad idea 🤔

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -26,6 +31,7 @@ pub async fn run_webserver(
 
     // Authenticated, public routes used for gateway administration
     let admin_routes = Router::new()
+        // TODO: (@oleonardolima) Should the /info be really a POST ?
```
oleonardolima, line 34: @okjodom Was POST being used for any specific reason ? 🤔

okjodom: GET seems more appropriate

gateway/ln-gateway/src/rpc/mod.rs
okjodom: **Describing the scenario to rebuild context**

- When we use `/get_info` API without a filter, we need to return `GatewayInfo` struct which lists partials of `FederationInfo` structs.
- When we use `/get_info` API has a filter, i.e. `/get_info <federation_id>`, we need to return  `GatewayInfo` struct which potentially lists  a single, NON-partial `FederationInfo`.

Making `Option`al fields in `FederationInfo` would my go to for making partials `FederationInfo`. Are there alternative patterns? @Maan2003, @dpc

okjodom: image

Maan2003: use `serde(flatten)`

```rust
struct Partial { ... };

struct Complete {
   #[serde(flatten)]
   partial: Partial,
   ....
};

Comment (Maan2003) [reply gateway/ln-gateway/src/rpc/mod.rs:120]: or the other way around

```rust
struct Foo {
  // requried_fields ...
  complete: Option<Complete>;
};

struct Complete { ... };
```

gateway/ln-gateway/src/lib.rs
jkitman: I think we can assume the federation id is unique.

gateway/ln-gateway/src/rpc/mod.rs
oleonardolima: @okjodom @Maan2003 I was trying this approach but I'm still not sure how the GatewayInfo would be able to handle both `PartialFederationInfo` and `FederationInfo`, as they would be two different types 🤔.

Should the Partial pattern handle that, am I missing something ?

[I've tried to solve it by using a `FederationInfo` enum alongside the Partial structures](https://github.com/fedimint/fedimint/pull/2046/commits/63275fafbbae740935f01304cdf88ce9cefe1360), but it doesn't seem to help.

gateway/ln-gateway/src/lib.rs
oleonardolima: Great! It's safe as it's then 😁

justinmoon:
Note: if you rebase this, all those code coverage comments https://github.com/fedimint/fedimint/pull/2046/files should go away.

gateway/ln-gateway/src/rpc/mod.rs
oleonardolima: @okjodom @Maan2003 fyi, trying to apply the Partial pattern was getting really complicated.

Taking a deeper look at the `serde` docs I've found the skip... attributes, and I've used the [`skip_serializing_if`](https://serde.rs/field-attrs.html#skip_serializing_if) as is really suitable for the expected behavior, but please let me know what you think about it.

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -53,10 +61,64 @@ async fn gatewayd_supports_multiple_federations() -> anyhow::Result<()> {
...
+        .get_balance(BalancePayload {
+            federation_id: fed1_id,
+        })
+        .await
+        .unwrap();
+    let fed1_info = client.get_info(Some(fed1_id)).await.unwrap();
+
+    assert_eq!(fed1_info.federations.len(), 1);
+    assert_eq!(
+        fed1_info.federations.first().unwrap().federation_id,
```
jkitman, line 96: nit: tests would read better if you assign id to a variable or expose a function for getting it.

okjodom:
@oleonardolima, let's collab on this item? Give me a ping when free

okjodom (approved):
afaik, this will not introduce any conflicts with active workstreams in gateway-ng since it only touches on the gateway webserver. the new gateway client will need to support the same info spec when ready

gateway/ln-gateway/src/rpc/rpc_client.rs
```diff
@@ -37,8 +38,16 @@ impl GatewayRpcClient {
...
-    pub async fn get_info(&self) -> GatewayRpcResult<GatewayInfo> {
-        let url = self.base_url.join("/info").expect("invalid base url");
+    pub async fn get_info(
+        &self,
+        federation_id: Option<FederationId>,
+    ) -> GatewayRpcResult<GatewayInfo> {
+        let path = match federation_id {
+            Some(id) => format!("/info?federation_id={id}"),
+            None => "/info".to_string(),
+        };
```
elsirion, line 48: Nit: string interpolation is prone to injection attacks (wrong/missing escaping), probably ok here because the tool is only used by the admin and not malicious 3rd parties.

gateway/ln-gateway/src/lib.rs
```diff
@@ -389,13 +389,24 @@ impl Gateway {
...
+                    .cloned()
+                    .collect::<Vec<_>>()
+            })
+            .unwrap_or(all_federations);
```
elsirion: Why don't we first figure out from which actors we want the info and then fetch it? As done currently passing the argument merely filters the output but doesn't reduce the cost of running the command.

gateway/ln-gateway/src/lib.rs
```diff
@@ -384,28 +384,27 @@ impl Gateway {
...
+                    .get_detailed_info()
+                    .await?,
+            ),
+            None => {
+                let actors = self.actors.read().await;
+                for actor in actors.values() {
+                    federations.push(actor.get_summary_info()?);
+                }
+            }
+        };
```
elsirion: Never mind my previous comment, fixed here. When it comes to PR structure though: it would be nice not to undo changes of previous commits, otherwise a lot of time can be wasted when reviewing commit-by-commit. Instead prefer interactive rebasing/amending commits when you notice something has to be changed in the history.

elsirion (approved):
Looks good enough to me, I leave the merge decision to @justinmoon since he's currently working on the GW.

gateway/ln-gateway/src/rpc/rpc_client.rs
oleonardolima: I wasn't fully aware of it, thanks for the comment. The solution for it is actually pretty straightforward I think. Instead of using query parameters, it could be checked based on the payload `federation_id` field (in order to still use the same route).

gateway/ln-gateway/src/lib.rs
oleonardolima: Sorry about that 😅, I should've squashed/amended some changes. Thanks for the tip, I'll keep that in mind for the next PRs

m1sterc001guy:
@oleonardolima some changes were recently merged to make `gateway-cli info` more readable. Can you evaluate if this change is still necessary?

elsirion:
@m1sterc001guy @oleonardolima Is this PR still relevant?

oleonardolima:
> @m1sterc001guy @oleonardolima Is this PR still relevant?

I don't think so w.r.t having the balance info on the `gateway-cli info` command. 

It's still useful if we'd like to add the filter by `federation-id` feature and not fetch all, but we discussed closing this one in a past dev call.

What do you think @okjodom ? (as you're the original reporter on #1725)

oleonardolima:
Closing as per the discussion above.

Issue #1937, opened 2023-03-15
Create more windows in tmuxinator


justinmoon:
In the original [LND PR](https://github.com/fedimint/fedimint/pull/1719), I split out more windows in tmuxinator and it was very nice:

- user shell
- 4 panes each running their own fedimintd instance
- 4 panes with CLN, LND, and gateways for both
- bitcoind

Could potentially add electrs and esplora, which don't currently run in tmuxinator

justinmoon:
@dhruv-1001 There's one thing left to do here.

Tmuxinator currently runs 1 window for the federation with only 1 pane. Inside that, 4 instances of fedimintd are running. That sucks because you get 3-4 copies of every log statement. It would be better to have 4 separate panes each running an instance of fedimintd. That way the logs are much easier to read. Make sense?

Tricky part here is "where to run `run_dkg`"? That's what generates the config files that `fedimintd` needs to run inside tmuxinator. I think this will be much easier to achieve once https://github.com/fedimint/fedimint/pull/1910 is merged. Could you try running that PR and reviewing it a little? Fine if you don't understand a lot. Feel free to ask questions on PR or in discord. Once it's merged, maybe we can do the 4 tmux pane thing ...

dhruv-1001:
> Tricky part here is "where to run `run_dkg`"? 

The way I solved this is by adding `await_dkg_function`
``` rust
// In the last step of run_dkg, client.json is moved into `/cfg` from `server-0`. And this function checks if this files exists
async fn await_dkg_complete(waiter_name: &str) -> anyhow::Result<()> {
    let cfg_dir: PathBuf = env::var("FM_CFG_DIR")?.parse()?;
    let client_json_dir = cfg_dir.join("client.json"); 

    loop {
        if client_json_dir.exists() {
            break;
        }
        info!("{waiter_name} waiting for dkg ...");
        sleep(Duration::from_secs(1)).await;
    }
    Ok(())
}
```
And started the federation as such 
```
  - federation:
      panes:
        - server-0:
          - start_fedimintd 0
        - server-1:
          - start_fedimintd 1
        - server-2:
          - start_fedimintd 2
        - server-3:
          - start_fedimintd 3
```
(`start_fedimintd` runs the command `fixtures fedimintd <server-id>`)

And in the `fixtures.rs` I had every `fedimintd` instance wait for DKG to complete

justinmoon:
That `await_dkg_complete` looks really nice!

[...]

PR #2052, opened 2023-03-28
WIP: test LND gateway in cli-test.sh

justinmoon:
Closes https://github.com/fedimint/fedimint/issues/1995
Progress towards https://github.com/fedimint/fedimint/issues/1937

- Runs a gateway with LND in relevant tests and tmuxinator. This was an omission from https://github.com/fedimint/fedimint/pull/1719.
- Tmuxinator now has `gateway-lnd` and `gateway-cln` instead of `gateway-cli` so both gateways can be used.


gateway/ln-gateway/src/lib.rs
```diff
@@ -48,6 +48,8 @@ use crate::rpc::{
 
 const ROUTE_HINT_RETRIES: usize = 10;
 const ROUTE_HINT_RETRY_SLEEP: Duration = Duration::from_secs(2);
+/// LND HTLC interceptor can't handle SCID of 0, so start from 1
+const INITIAL_SCID: u64 = 1;
```
justinmoon, line 52: This was a very important change. Without this, you can't receive into the federation via LND gateway because LND can't intercept HTLCs with an SCID of 0 (the previous value).

okjodom (approved):

scripts/lib.sh
```diff
@@ -169,13 +198,36 @@ function start_lnd() {
...
+  # start lnd gw
+  export FM_GATEWAY_DATA_DIR=$FM_TEST_DIR/gw-lnd
+  export FM_GATEWAY_LISTEN_ADDR="127.0.0.1:28175"
+  export FM_GATEWAY_API_ADDR="http://127.0.0.1:28175"
+  $FM_BIN_DIR/fixtures gatewayd lnd &
+  echo $! >> $FM_PID_FILE
+  echo "started lnd gateway"
+
+  connect_lnd_gateway
+}
```
dhruv-1001, line 226: Does it make sense to have these env variables to be passed as arguments by rust? [This](https://github.com/dhruv-1001/fedimint/blob/tmux-test/fedimint-testing/src/bin/fixtures.rs#L397-L416) is what I was working on for this issue.

scripts/cli-test.sh
```diff
@@ -56,26 +56,49 @@ RECEIVED=$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR)
...
 ADD_INVOICE="$($FM_LNCLI addinvoice --amt_msat 100000)"
 INVOICE="$(echo $ADD_INVOICE| jq -e -r '.payment_request')"
 PAYMENT_HASH="$(echo $ADD_INVOICE| jq -e -r '.r_hash')"
 $FM_MINT_CLIENT ln-pay $INVOICE
 # Check that ln-gateway has received the ecash notes from the user payment
 # 100,000 sats + 100 sats without processing fee
-LN_GATEWAY_BALANCE="$($FM_GATEWAY_CLI balance $FED_ID | jq -e -r '.balance_msat')"
-[[ "$LN_GATEWAY_BALANCE" = "100100000" ]]
+# LN_GATEWAY_BALANCE="$($FM_GATEWAY_CLI balance $FED_ID | jq -e -r '.balance_msat')"
+# [[ "$LN_GATEWAY_BALANCE" = "100100000" ]]
```
m1sterc001guy, line 70: Are these still necessary?

m1sterc001guy (approved):
LGTM

scripts/lib.sh
justinmoon: Yea I think that looks nice. The only problem is that https://github.com/fedimint/fedimint/pull/1910 is a huge refactor in this direction. But still a good change that we can port over once #1910 lands.

scripts/cli-test.sh
justinmoon: Fixed in https://github.com/fedimint/fedimint/pull/2058

PR #2054, opened 2023-03-28
chore: make it easier to locate a failed CI tests

dpc:
Print a big `## FAILED: {testname}` after each failed test output.

How it will look:

```
> just test-ci-all
./scripts/test-ci-all.sh
⚠️  ulimit too small. Running 'ulimit -Sn 10000' to avoid problems running tests
    Finished dev [unoptimized + debuginfo] target(s) in 0.46s
### Starting all tests in parallel...
### Starting integration test...
## FAILED: cli_test_rust_tests
### Starting always_fail test...
00:00:00 This test always fails and it is expected
...
```

### DIFF
--- a/scripts/test-ci-all.sh
+++ b/scripts/test-ci-all.sh
@@ -25,6 +25,8 @@ cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} --all --all-targets
 
 function cli_test_reconnect() {
   set -eo pipefail # pipefail must be set manually again
+  trap 'echo "## FAILED: ${FUNCNAME[0]}"' ERR 
+
   echo "### Starting reconnect test..."
   unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/reconnect-test.sh" 2>&1 | ts -s
 }
@@ -32,6 +34,8 @@ export -f cli_test_reconnect
 
 function cli_test_upgrade() {
   set -eo pipefail # pipefail must be set manually again
+  trap 'echo "## FAILED: ${FUNCNAME[0]}"' ERR 
+
   echo "### Starting upgrade test..."
   unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/upgrade-test.sh" 2>&1 | ts -s
 }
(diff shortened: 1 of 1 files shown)


scripts/test-ci-all.sh
```diff
@@ -25,41 +25,54 @@ cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} --all --all-targets
...
+
   echo "### Starting integration test..."
   unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/rust-tests.sh" 2>&1 | ts -s
 }
 export -f cli_test_rust_tests
 
 function cli_test_always_fail() {
   set -eo pipefail # pipefail must be set manually again
+  # skip reporting the failure
+  # trap 'echo "## FAILED: ${FUNCNAME[0]}"' ERR
```
jkitman, line 74: Why is this commented?

dpc: Compromise between consistency (someone might think i missed it here), and not actually printing it...

But now that you mention it ... this test actually doesn't fail because it uses `!` ... so I don't need to comment it out...

dpc: Fixed. Thanks!

scripts/test-ci-all.sh
```diff
@@ -25,41 +25,53 @@ cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} --all --all-targets
 
 function cli_test_reconnect() {
   set -eo pipefail # pipefail must be set manually again
+  trap 'echo "## FAILED: ${FUNCNAME[0]}"' ERR
```
dpc, line 28: Unfortunately, I don't think I can re-use anything here, as it seems that each of these gets executed "from scratch" by gnu parallel so all pre-existing bash options are gone, and there's not enough redundancy yet, to make me want to spend more time investigating best way to reuse (maybe common exported function will work, or maybe I need a file).

jkitman (dismissed):
LGTM.  BTW I added a `fm success` at the end of each script a while ago, but maybe we should remove that and stick to this idiom.

Maan2003 (approved):

justinmoon (approved):

justinmoon:
> LGTM. BTW I added a `fm success` at the end of each script a while ago, but maybe we should remove that and stick to this idiom.

This "fm success" thing was really nice improvement btw. That's part of the reason I suggested this idea to dpc.

PR #2058, opened 2023-03-29
fix: improve lightning payment balance assertions in cli-test.sh

justinmoon:
I commented-out some balance assertions in https://github.com/fedimint/fedimint/pull/2052 because they didn't make sense to me and I wanted to revisit them. But I forgot to. Luckily @m1sterc001guy [caught it in review](https://github.com/fedimint/fedimint/pull/2052#discussion_r1151168497). I merged that PR and fix the issue here.

TLDR: cli-test.sh sends 100,000 msats from user to external lightning node via gateway. Looking at the old code, I mistakenly assumed that the starting balance of the gateway was 100,000,000 msats so checking that the final balance of the gateway increased to 100,100,000 seemed wrong -- what about the [hard-coded 1% fee](https://github.com/fedimint/fedimint/blob/428166662c4ccf7dbb415e6b93ee16fb2d52ade3/client/client-lib/src/ln/mod.rs#L84-L100)?

I have been confused by this before and didn't investigate. What I discovered was that the gateway's initial balance was actually 99,999,000 msats and not 100,000,000 msats. So as expected the 1% fee was being taken.

Refactored the tests to make assertions by subtracting initial and final balances so this is clearer.

### DIFF
--- a/scripts/cli-test.sh
+++ b/scripts/cli-test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # Runs a CLI-based integration test
 
-set -euo pipefail
+set -euxo pipefail
 export RUST_LOG="${RUST_LOG:-info}"
 
 export PEG_IN_AMOUNT=10000
@@ -51,7 +51,10 @@ until [ "$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR 0)" == "0.00000500"
 done
 mine_blocks 10
 RECEIVED=$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR)
-[[ "$RECEIVED" = "0.00000500" ]]
+if [[ "$RECEIVED" != "0.00000500" ]]; then
+  echo "Peg-out address received $RECEIVED, expected 0.00000500"
+  exit 1
+fi
 
 # lightning tests
 await_lightning_node_block_processing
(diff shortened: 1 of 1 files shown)


dpc (approved):

okjodom (approved):
✅

m1sterc001guy (approved):

dpc (changes requested):
Please add `>&2 echo "Assertion failed: blablabla"`

scripts/cli-test.sh
```diff
@@ -60,17 +60,22 @@ await_lightning_node_block_processing
...
-# LN_GATEWAY_BALANCE="$($FM_GATEWAY_CLI balance $FED_ID | jq -e -r '.balance_msat')"
-# [[ "$LN_GATEWAY_BALANCE" = "100100000" ]]
+
 INVOICE_STATUS="$($FM_LNCLI lookupinvoice $PAYMENT_HASH | jq -e -r '.state')"
 [[ "$INVOICE_STATUS" = "SETTLED" ]]
 
+# Assert balances changed by 100000 msat (amount sent) + 1000 msat (fee)
+FINAL_CLIENT_BALANCE="$($FM_MINT_CLIENT info | jq -e -r '.total_amount')"
+FINAL_GATEWAY_BALANCE="$($FM_GWCLI_CLN balance $FED_ID | jq -e -r '.balance_msat')"
+[[ "$(($INITIAL_CLIENT_BALANCE - $FINAL_CLIENT_BALANCE))" = "101000" ]]
```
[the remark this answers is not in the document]
justinmoon: Fixed

dpc (dismissed):

scripts/cli-test.sh
```diff
@@ -51,7 +51,7 @@ until [ "$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR 0)" == "0.00000500"
 done
 mine_blocks 10
 RECEIVED=$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR)
-[[ "$RECEIVED" = "0.00000500" ]]
+[[ "$RECEIVED" != "0.00000500" ]];
```
jkitman, line 54: Why does this assert `!=` instead of the exact `=` amount?

dpc: Yeah, what happened here. :D

Also will fail silently which is not good.

dpc: We now know that we had a fiat CI again.

justinmoon:
Pushed a whitespace commit to see if the erroneous `!=` will trigger CI failure now.

scripts/cli-test.sh
```diff
@@ -51,7 +51,7 @@ until [ "$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR 0)" == "0.00000500"
 done
 mine_blocks 10
 RECEIVED=$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR)
-[[ "$RECEIVED" = "0.00000500" ]]
+[[ "$RECEIVED" == "0.00000500" ]];
```
dpc, line 54: Well, if we're already touching it ... why not a proper error message instead of failing silently? :)

justinmoon:
Passing CI now

dpc (dismissed):

scripts/cli-test.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # Runs a CLI-based integration test
 
-set -euo pipefail
+set -euxo pipefail
```
elsirion, line 4: Sneaking in `-x` again? :P

elsirion (approved):

elsirion:
Requires rebase :(

justinmoon:
Closing this because it's fixed in https://github.com/fedimint/fedimint/pull/2147. CI got the better of me this time 😂

Issue #222, opened 2022-07-06
Intermittant CI failure tracking issue


justinmoon:
We've had a handful of CI failure that seem to disappear when you re-run them. Making an issue to keep track of them.

- `lightning_gateway_pays_invoice` `lightning.amount_sent()` assertion https://github.com/fedimint/minimint/runs/7213579990?check_suite_focus=true#step:6:4320

elsirion:
Maybe just wait/poll till it has the expected value and timeout otherwise?

NicolaLS:
> Here's another one https://github.com/fedimint/minimint/runs/7290083028?check_suite_focus=true#step:7:2246

wanted to post the same in here, I get this a lot on my local setup especially on this [branch](https://github.com/NicolaLS/minimint/commit/d7256c97f4cee1123c5207e69a22fddb9dc4f83c)

NicolaLS:
https://github.com/fedimint/minimint/runs/7468414843?check_suite_focus=true

justinmoon:
https://github.com/fedimint/minimint/runs/7634996611?check_suite_focus=true#step:5:1698

NicolaLS:
https://github.com/fedimint/fedimint/runs/7881960910?check_suite_focus=true

justinmoon:
> https://github.com/fedimint/fedimint/runs/7881960910?check_suite_focus=true

Port 8080 already bound, again https://github.com/fedimint/fedimint/runs/7881960910?check_suite_focus=true#step:5:6101

justinmoon:
latency tests got stuck https://github.com/fedimint/fedimint/runs/7925922988?check_suite_focus=true#step:7:939

justinmoon:
https://github.com/fedimint/fedimint/actions/runs/3049065296/jobs/4914765743

jkitman:
@justinmoon Did using port picker resolve these issues?

justinmoon:
I don't know because it only happens very infrequently. Do you want to just add it? https://github.com/fedimint/fedimint/pull/427

NicolaLS:
@jkitman could this maybe be the same issue I had ? (which got fixed by running the tests with `--release`) I also got the port already in use error (for somereason?) when that happened

jkitman:
@NicolaLS I don't think so because we removed the timeouts which was the issue you were hitting.  Still, shouldn't hurt to add `port-picker` to the integration tests as well...

elsirion:
[...]

PR #2059, opened 2023-03-29
Remove panics from bitcoind RPC

jkitman:
@justinmoon Fixes failure here https://github.com/fedimint/fedimint/issues/222#issuecomment-1487825777 which could could fedimintd to panic.

### DIFF
--- a/fedimint-bitcoind/src/bitcoincore_rpc.rs
+++ b/fedimint-bitcoind/src/bitcoincore_rpc.rs
@@ -215,8 +215,7 @@ where
             self.0
                 .estimate_smart_fee(confirmation_target, Some(EstimateMode::Conservative))
                 .map_err(anyhow::Error::from)
-        })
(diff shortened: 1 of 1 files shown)


justinmoon (approved):

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -215,8 +215,7 @@ where
             self.0
                 .estimate_smart_fee(confirmation_target, Some(EstimateMode::Conservative))
                 .map_err(anyhow::Error::from)
-        })
-        .expect("Bitcoind returned an error") // TODO: implement retry logic in case bitcoind is temporarily unreachable
```
justinmoon, line 219: Is this retry logic still needed?

jkitman: Nope we wrap all bitcoin rpc calls in retries already.

Maan2003 (approved):

Issue #2048, opened 2023-03-28
Update codecov app


dpc:
https://about.codecov.io/blog/codecov-is-updating-its-github-integration/

> If you are a user who is not an organization administrator, there is nothing you need to do after the switch. You will simply be guided through the GitHub App Integration login flow the next time you log in to Codecov. If you are an organization administrator and you have installed the App Integration on your organization, there is no further action required.

justinmoon:
Nothing actionable here yet? Seems like new app isn't available yet?

dpc:
![image](https://user-images.githubusercontent.com/9209/228641492-0e2777e1-155f-4bd8-944d-8fd8751ff5e6.png)


Code Coverage messages in every PR say to install the app. I'm a bit confused what is what here, so just reported it and moved on.

PR #2065, opened 2023-03-29
fix: make new codecov app happy

dpc:
edit(justin): closes https://github.com/fedimint/fedimint/issues/2048

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,13 @@
+codecov:
+  require_ci_to_pass: true
+  coverage:
+    precision: 2
+    round: down
+    range: "60...100"
+    status:
+      project:
+        default:
+          target: auto
+          threshold: 1%
+  notify:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,13 @@
...
+  require_ci_to_pass: true
+  coverage:
+    precision: 2
+    round: down
+    range: "60...100"
+    status:
+      project:
+        default:
+          target: auto
+          threshold: 1%
```
justinmoon, line 11: so if a PR reduces coverage by 1%, CI will fail?

justinmoon (approved):

codecov.yml
dpc: I think so, yes. But it won't block landing the PR.

Maan2003 (approved):

PR #2068, opened 2023-03-30
Api config v4

jkitman:
Adds starting consensus to the config api

### DIFF
--- a/fedimint-core/src/admin_client.rs
+++ b/fedimint-core/src/admin_client.rs
@@ -30,8 +30,8 @@ impl WsAdminClient {
     /// Sets the password used to decrypt the configs and authenticate
     ///
     /// Must be called first before any other calls to the API
-    pub async fn set_password(&self, auth: ApiAuth) -> FederationResult<()> {
-        self.request_auth("set_password", ApiRequestErased::new(auth))
+    pub async fn set_password(&self) -> FederationResult<()> {
+        self.request_auth("set_password", ApiRequestErased::new(self.auth.clone()))
             .await
     }
 
@@ -140,6 +140,13 @@ impl WsAdminClient {
             .await
     }
 
+    /// Reads the configs from the disk, starts the consensus server, and shuts
+    /// down the config gen API to start the Fedimint API
+    pub async fn start_consensus(&self) -> FederationResult<()> {
+        self.request_auth("start_consensus", ApiRequestErased::new(self.auth.clone()))
+            .await
+    }
+
     async fn request_auth<Ret>(
         &self,
         method: &str,
@@ -174,11 +181,16 @@ pub struct ConfigGenConnectionsRequest {
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+/// Connection information sent between peers in order to start config gen
 pub struct PeerServerParams {
+    /// TLS cert is necessary for P2P auth during DKG and  consensus
     #[serde(with = "serde_tls_cert")]
     pub cert: rustls::Certificate,
+    /// P2P is the network for running DKG and consensus
     pub p2p_url: Url,
+    /// API for secure websocket requests
     pub api_url: Url,
+    /// Name of the peer, used in TLS auth
     pub name: String,
 }
 
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -647,7 +647,8 @@ pub async fn run_server(
         let auth = config_generated_rx.recv().await.expect("should not close");
         server_handle.stop().expect("Able to stop server");
         let cfg = read_server_config(&auth.0, data_dir.clone())?;
-        FedimintServer::run(cfg, db.clone(), registry.clone(), None, &mut task_group).await?;
+        FedimintServer::run(cfg, db.clone(), registry.clone(), None, &mut task_group)
+            .await?;
     }
 }
 
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -51,15 +51,9 @@ pub struct ConfigGenApi {
     /// DB not really used
     db: Database,
     /// Our connection info configured locally
-    our_connections: ConfigGenConnections,
+    settings: ConfigGenSettings,
     /// Notify if we receive connections from peer
     notify_peer_connection: Notify,
-    /// The default params for the modules
-    default_params: ConfigGenParamsRequest,
-    /// Modules that will generate configs
-    module_gens: BTreeMap<u16, (ModuleKind, DynServerModuleGen)>,
-    /// Registry for config gen
-    registry: ServerModuleGenRegistry,
     /// Tracks when the config is generated
     config_generated_tx: Sender<ApiAuth>,
 }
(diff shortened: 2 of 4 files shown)


fedimint-server/src/config/api.rs
```diff
@@ -563,15 +597,17 @@ impl HasApiContext<ConfigGenApi> for ConfigGenApi {
...
+            ConfigApiState::VerifyConsensusConfig(cfg) => Some(&cfg.private.api_auth) == auth,
+            ConfigApiState::RunningConsensus(cfg) => Some(&cfg.private.api_auth) == auth,
         };
 
         (self, ApiEndpointContext::new(has_auth, dbtx, id))
     }
 }
 
-/// Starts the configuration server
+/// Starts the config gen server
```
dpc, line 608: Nit, style: Newline between the "one line header" and rest of the docstring recommended.

fedimint-server/src/config/api.rs
```diff
@@ -580,29 +616,42 @@ pub async fn run_server(
...
-        .ping_interval(Duration::from_secs(10))
-        .build(&our_connections.api_bind.to_string())
-        .await
-        .expect("Could not start API server")
-        .start(rpc_module)
-        .expect("Could not start API server")
+    mut task_group: TaskGroup,
+) {
+    loop {
+        let (api_sender, mut api_receiver) = tokio::sync::mpsc::channel(1);
```
dpc, line 622: `oneshot::channel`?

fedimint-server/src/config/api.rs
```diff
@@ -580,29 +616,42 @@ pub async fn run_server(
...
+        let mut rpc_module = RpcModule::new(state);
+
+        attach_endpoints(&mut rpc_module, config_endpoints(), None);
+
+        let server_handle = ServerBuilder::new()
+            .max_connections(10)
+            .ping_interval(Duration::from_secs(10))
+            .build(&our_connections.api_bind.to_string())
+            .await
+            .expect("Could not start API server")
```
dpc, line 643: All these `expect`s ... Wouldn't it be easier to return `anyhow::Result`? Just saying, not necessary in this PR.

fedimint-server/src/config/api.rs
```diff
@@ -755,31 +811,44 @@ mod tests {
...
-            .await;
+            let dir = data_dir.join(name_suffix.to_string());
+            fs::create_dir_all(dir.clone()).expect("Unable to create test dir");
+
+            let connections = our_connections.clone();
+            let mut task_group = TaskGroup::new();
+            let subgroup = task_group.make_subgroup().await;
+            task_group
+                .spawn("run-server", move |_| async move {
+                    run_server(
```
dpc, line 822: The number of arguments calls for either a builder (especially given the amount of `Default::default`, or at least packing some of them into some aggregating `struct`.

fedimint-server/src/config/api.rs
```diff
@@ -755,31 +811,44 @@ mod tests {
...
+                task_group,
                 name,
                 our_connections,
             }
         }
 
-        /// Helper function using the auth we generated
-        async fn set_password(&self) -> FederationResult<()> {
-            self.client.set_password(self.auth.clone()).await
+        /// Helper function using the auth we generated, retries until the API
```
dpc, line 846: :sus: . Seems like some another `oneshot::channel` called `api_read_receiver` could be better. Especially that we can't tell if something else is wrong, or it is just not ready or what exactly.

fedimint-server/src/config/api.rs
```diff
@@ -755,31 +811,44 @@ mod tests {
...
+        /// is up
+        async fn retry_set_password(&self) {
+            while self.client.set_password().await.is_err() {
+                sleep(Duration::from_millis(100)).await;
```
dpc, line 850: `info!("stuff not ready");` would give some indication when things went wrong and this is stuck here.

dpc (commented):
Suggestions.

fedimint-core/src/admin_client.rs
```diff
@@ -174,11 +181,16 @@ pub struct ConfigGenConnectionsRequest {
...
 pub struct PeerServerParams {
+    /// TLS cert is necessary for P2P auth during DKG and  consensus
     #[serde(with = "serde_tls_cert")]
     pub cert: rustls::Certificate,
+    /// P2P is the network for running DKG and consensus
     pub p2p_url: Url,
+    /// API for secure websocket requests
     pub api_url: Url,
+    /// Name of the peer, used in TLS auth
     pub name: String,
```
justinmoon, line 194: Side note: would be nice if this type validated that it was a valid DNS name. I know that in the past I've entered names with spaces or just a number like `3` and I get runtime error that invalid DNS name was entered. Might be better if this was validated up-front and enforced by type system.

fedimint-server/src/config/api.rs
```diff
@@ -303,12 +315,34 @@ impl ConfigGenApi {
...
+
+    /// Creates the Fedimint server, decrypting the config files from disk using
+    /// the auth string.  We send a message to the `run_server` task to shutdown
+    /// the API so that the Fedimint API can start.
+    pub async fn start_consensus(&self, auth: ApiAuth) -> ApiResult<()> {
+        {
+            let cfg = read_server_config(&auth.0, self.data_dir.clone())
+                .map_err(|e| ApiError::bad_request(format!("Unable to decrypt configs {e:?}")))?;
+
+            let mut state = self.state.lock().expect("lock poisoned");
```
justinmoon, line 338: So panicking is the right think to do here because once the mutex is poisoned we can't really recover?

fedimint-server/src/config/api.rs
```diff
@@ -875,12 +943,17 @@ mod tests {
...
-            peer.server.stop().expect("server stops");
+        // start consensus
+        for peer in &followers {
+            peer.client.start_consensus().await.unwrap();
+        }
+
+        // shutdown
+        for peer in &followers {
+            peer.task_group.shutdown().await;
         }
```
justinmoon, line 957: Do we know that any epochs were produced here? Could it be that consensus started but no epochs were produced?

fedimint-server/src/config/api.rs
```diff
@@ -303,12 +315,34 @@ impl ConfigGenApi {
...
+    /// the API so that the Fedimint API can start.
+    pub async fn start_consensus(&self, auth: ApiAuth) -> ApiResult<()> {
+        {
+            let cfg = read_server_config(&auth.0, self.data_dir.clone())
+                .map_err(|e| ApiError::bad_request(format!("Unable to decrypt configs {e:?}")))?;
+
+            let mut state = self.state.lock().expect("lock poisoned");
+            *state = ConfigApiState::RunningConsensus(cfg);
         }
+        self.config_created
```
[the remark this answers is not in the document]
jkitman: I'll go with `config_generated_tx`

fedimint-server/src/config/api.rs
jkitman: Sure, for now that makes sense.

fedimint-server/src/config/api.rs
jkitman: I didn't want to pollute the production code with code that would only be used for this one test.  In practice, clients will have to retry like this once the API server starts.

fedimint-server/src/config/api.rs
jkitman: Yeah, I don't think there is a way to recover from them in a way that makes sense because it means I panicked in some other thread that was holding the lock.  I tried to handle the API calls in ways that don't ever panic, but return errors.

fedimint-server/src/config/api.rs
jkitman: My next PR will attempt to handle the upgrade + shutdown and restart from config API again case.

fedimint-server/src/config/api.rs
jkitman: Good point, I'll refactor

fedimint-server/src/config/api.rs
jkitman: Had some trouble with that since the Api struct holds onto it and `send` consumes the oneshot.

justinmoon (approved):

Maan2003 (commented):
LGTM (not approving to let dpc take a look)

Issue #1990, opened 2023-03-21
Support bitcoin RPC auth via "cookie" file


justinmoon:

oleonardolima:
This one seems pretty straightforward, by using the [rust-bitcoincore-rpc `Auth::CookieFile`](https://docs.rs/bitcoincore-rpc/0.16.0/src/bitcoincore_rpc/client.rs.html#194) [here](https://github.com/fedimint/fedimint/blob/c5cb7dc811ac036cb976d4481471df3ec7491a7f/fedimint-bitcoind/src/bitcoincore_rpc.rs#L33-L59), is that right ?

I guess the problem would be how to get the path for the cookie file, a new environment variable as the `FM_BITCOIND_RPC_ENV`, would be appropriate?

https://github.com/fedimint/fedimint/blob/c5cb7dc811ac036cb976d4481471df3ec7491a7f/fedimint-bitcoind/src/bitcoincore_rpc.rs#L33-L59

anmode:
I think we have to create a cookie file. isn't?                                                                                                                                       
```# Create the cookie file
echo "rpcuser=bitcoin" > $FM_BTC_DIR/.cookie
echo "rpcpassword=bitcoin" >> $FM_BTC_DIR/.cookie
chmod 400 $FM_BTC_DIR/.cookie`

anmode:
Cookie based authentication is used when no rpc password is provided. The 0.12 release of Bitcoin Core had the following to say about it:                                                                                                                                                                                   
 ```When no -rpcpassword is specified, the daemon now uses a special ‘cookie’ file for authentication. This file is generated with random content when the daemon starts, and deleted when it exits. Its contents are used as authentication token. Read access to this file controls who can access through RPC. By default it is stored in the data directory but its location can be overridden with the option -rpccookiefile.```     
                                                                                                                                                                                                                   but here ig we have to create our cookie file like this-                                                                                                                                                                             
 ```# Create the cookie file
echo "rpcuser=bitcoin" > $FM_BTC_DIR/.cookie
echo "rpcpassword=bitcoin" >> $FM_BTC_DIR/.cookie
chmod 400 $FM_BTC_DIR/.cookie

elsirion:
> I guess the problem would be how to get the path for the cookie file, a new environment variable as the `FM_BITCOIND_RPC_ENV`, would be appropriate?

[...]

Issue #1994, opened 2023-03-22
Show module env vars in clap help


elsirion:
Modules that expect environment variables to be present should have to register them with fedimint so that they can be displayed as part of the clap help shown when running `fedimintd --help`. Otherwise they are not very discoverable, especially in the case of the bitcoin RPC (was an issue here for example #1988) where there is a default value users might never learn about the env var.

PR #2075, opened 2023-03-31
Bitcoin RPC Cookie Authentication 

VeronicaTee:
Fixes https://github.com/fedimint/fedimint/issues/1990

Fix:
Add support for bitcoin authentication with a cookie file.
When no rpcpassword is specified, a  ‘cookie’ file is used for authentication. Its contents are used as authentication token. If a cookie file exists, it is read for authentication, if it doesn't exist a 'temporary' one is generated for authentication.
Read access to this file controls who can access through RPC.


fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -47,6 +47,50 @@ pub fn from_url_to_url_auth(url: &Url) -> Result<(String, Auth)> {
         }),
         if url.username().is_empty() {
             Auth::None
+        } else if url.password().is_none() {
```
dpc, line 50: I don't think this is the way to do this. Having no password set doesn't mean the cookie file should be used.

We should probably use cookies if certain env vars are used or something. See `fedimint-core/src/bitcoin_rpc.rs` for how it's handled. I'm not sure how exactly should it work.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -47,6 +47,50 @@ pub fn from_url_to_url_auth(url: &Url) -> Result<(String, Auth)> {
...
+                                };
+                            }
+                            _ => {}
+                        }
+                    }
+                }
+                auth
+            } else {
+                let cookie_file_path = format!("{}/.cookie", url.path());
+                std::fs::write(&cookie_file_path, "rpcuser=bitcoin\nrpcpassword=bitcoin\n")?;
```
dpc, line 87: ?

We shouldn't ever write the cookie file, only read it.

Or am I completely confused about how it works?

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -47,6 +47,50 @@ pub fn from_url_to_url_auth(url: &Url) -> Result<(String, Auth)> {
...
             Auth::None
+        } else if url.password().is_none() {
+            if let Some(cookie_file) = rpc_cookie_file {
+                let cookie = std::fs::read_to_string(cookie_file)?;
+                let mut auth = Auth::None;
+                for line in cookie.lines() {
+                    let tokens: Vec<_> = line.split('=').collect();
+                    if tokens.len() == 2 {
+                        match tokens[0] {
+                            "rpcuser" => {
```
dpc, line 58: I thought the format of cookie file was just one line with `<user>:<password>` or something like that?

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -47,6 +47,50 @@ pub fn from_url_to_url_auth(url: &Url) -> Result<(String, Auth)> {
         }),
         if url.username().is_empty() {
             Auth::None
+        } else if url.password().is_none() {
+            if let Some(cookie_file) = rpc_cookie_file {
+                let cookie = std::fs::read_to_string(cookie_file)?;
```
dpc, line 52: Reading&parsing the cookie file is a natural boundary to make a separate functions (parsing from `&str` should be separate), that can be tested easily in isolation.

fedimint-bitcoind/src/bitcoincore_rpc.rs
VeronicaTee: I followed the conversation on the issue. How do you feel I should implement this instead?

dpc: IMO, some env flag to make it use auth from the cookie file would be a good start. We can figure out the other details later, I guess.

VeronicaTee: That makes sense. So, cookie file path is set in the env to make it use the auth from the cookie file, right?

fedimint-bitcoind/src/bitcoincore_rpc.rs
VeronicaTee: Ok, I will isolate the functions.

oleonardolima: @VeronicaTee AFAIK parsing wouldn't be necessary, you could use the [`bitcoincore_rpc::Auth`](https://docs.rs/bitcoincore-rpc/0.16.0/bitcoincore_rpc/enum.Auth.html) with the [`CookieFile`](https://docs.rs/bitcoincore-rpc/0.16.0/bitcoincore_rpc/enum.Auth.html#variant.CookieFile) variant, and then get the authentication required by it's implemented methods.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -47,6 +48,17 @@ pub fn from_url_to_url_auth(url: &Url) -> Result<(String, Auth)> {
...
+        } else if url.password().is_none() {
+            if let Some(cookie_file) = rpc_cookie_file() {
+                let cookie = std::fs::read_to_string(cookie_file)?;
+                let auth = parse_cookie_string(&cookie).unwrap_or(Auth::None);
+                auth
+            } else {
+                let cookie_file_path = format!("{}/.cookie", url.path());
+                create_cookie_file(&cookie_file_path, "bitcoin", "bitcoin")?;
+                let auth = Auth::UserPass("bitcoin".to_owned(), "bitcoin".to_owned());
+                auth
```
oleonardolima, line 60: @VeronicaTee Is creating the .cookie file needed/appropriate? IIRC the creation should be before `bitcoind` start and then passed to it if the user needs a custom .cookie file.

IIRC also if the user wants to use both a `-rpcpassword`,`-rpcuser` and a .cookie file, it would need to use the `-rpcauth` instead which handles both user/pass and .cookie auth.

fedimint-bitcoind/src/bitcoincore_rpc.rs
VeronicaTee: @oleonardolima Like the issue here https://github.com/fedimint/fedimint/issues/1994 you mean? OK.

oleonardolima: @VeronicaTee I mean, getting the given directory path from environment var would still be needed, but you can pass it as a [`std::path::PathBuf`](https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html) to `bitcoincore_rpc::Auth`.

So it wouldn't need to handle and parse it as it's below 😁 

https://github.com/fedimint/fedimint/blob/787b2a022a05845ed9913de0eadd689cd6e63a08/fedimint-bitcoind/src/bitcoincore_rpc.rs#L53-L83

fedimint-bitcoind/src/bitcoincore_rpc.rs
VeronicaTee: Duly noted. Thank you.

fedimint-bitcoind/src/bitcoincore_rpc.rs
VeronicaTee: Ok, I understand you now. Thanks.

justinmoon:
Hasn't been much progress recently so closing this. If you make more progress feel free to re-open.

Issue #1866, opened 2023-03-11
Panic if gateway-cli tries to connect to federation it's already connected to


justinmoon:
1. Run tmuxinator
2. Run contents of `gw_connect_fed` function

```
thread 'main' panicked at 'Error opening new rocks DB: Error { message: "IO error: lock hold by current process, acquire time 1678503857 acquiring thread 8020458112: /tmp/nix-shell.Y0lKYu/fm-ZkPl/cfg/gateway/aada38a8da939ae320849b6add1609a2e175ccad30d10acbc4a55eb4b3da5c401afe3f76209824f8757678d6a298ec19.db/LOCK: No locks available" }', gateway/ln-gateway/src/client.rs:65:59
```

justinmoon:
https://github.com/fedimint/fedimint/pull/2099 was reverted https://github.com/fedimint/fedimint/pull/2135

jkitman:
Is this required for the alpha release?

okjodom:
Would be nice to make alpha. Let me revisit the reverted change

PR #2083, opened 2023-04-01
gateway: skip connecting to federation if already connected

natto1784:
Just skips connecting to it if the node is already connected
Related: https://github.com/fedimint/fedimint/issues/1866 (needs further looking into)

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -252,6 +252,8 @@ impl Gateway {
             GatewayError::Other(anyhow::anyhow!("Invalid federation member string {}", e))
         })?;
 
+        let _ = self.select_actor(connect.id.clone()).await?;
+
         let GetNodeInfoResponse { pub_key, alias: _ } = self.lnrpc.read().await.info().await?;
         let node_pub_key = PublicKey::from_slice(&pub_key)
             .map_err(|e| GatewayError::Other(anyhow!("Invalid node pubkey {}", e)))?;


gateway/ln-gateway/src/lib.rs
```diff
@@ -252,6 +252,14 @@ impl Gateway {
             GatewayError::Other(anyhow::anyhow!("Invalid federation member string {}", e))
         })?;
 
+        for actor in self.actors.lock().await.values() {
+            if actor.read().await.get_info()?.federation_id == connect.id {
```
oleonardolima, line 256: @natto1784 I guess that using `select_actor` would make it simpler, wdyt ?

natto1784: indeed. my bad, it was right there too

gateway/ln-gateway/src/lib.rs
```diff
@@ -252,6 +252,12 @@ impl Gateway {
             GatewayError::Other(anyhow::anyhow!("Invalid federation member string {}", e))
         })?;
 
+        if self.select_actor(connect.id.clone()).await.is_err() {
```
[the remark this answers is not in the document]
justinmoon: `.is_ok()` is nice way to negate check

gateway/ln-gateway/src/lib.rs
```diff
@@ -252,6 +252,8 @@ impl Gateway {
             GatewayError::Other(anyhow::anyhow!("Invalid federation member string {}", e))
         })?;
 
+        let _ = self.select_actor(connect.id.clone()).await?;
```
okjodom, line 255: Pretty sure this will prevent us from connecting any new federations! Have you tested this?
Can you verbally describe the behavior you are expecting and see if this check matches your expectations?

natto1784: ah youre right, i was kinda sleepy when i did this. anyway, dont think it deserves this much discussion, feel free to add it in some other patch.

PR #2084, opened 2023-04-01
gateway: return FederationInfo for /connect-fed

### LINKED ISSUE #2078: Gateway api /connect-fed should return detailed info about the newly connected federation
At present, [`/connect-fed` would returns a null](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/rpc/rpc_server.rs#L118-L126), meaning the gateway admin would have to call `/info` to get details about the fed they just connected. Let's make it better.

Related to #1725

natto1784:
Fixes #2078

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -247,7 +247,7 @@ impl Gateway {
         &self,
         payload: ConnectFedPayload,
         route_hints: Vec<RouteHint>,
-    ) -> Result<()> {
+    ) -> Result<FederationInfo> {
         let connect = WsClientConnectInfo::from_str(&payload.connect).map_err(|e| {
             GatewayError::Other(anyhow::anyhow!("Invalid federation member string {}", e))
         })?;
@@ -277,9 +277,12 @@ impl Gateway {
                 .expect("Failed to build gateway client"),
         );
 
-        if let Err(e) = self.load_actor(client.clone(), route_hints).await {
-            error!("Failed to connect federation: {}", e);
-        }
+        let actor = self
+            .load_actor(client.clone(), route_hints)
+            .await
+            .map_err(|e| {
+                GatewayError::Other(anyhow::anyhow!("Failed to connect federation {}", e))
+            })?;
 
         if let Err(e) = self.client_builder.save_config(client.config()) {
             warn!(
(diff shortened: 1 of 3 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -288,7 +290,9 @@ impl Gateway {
             );
         }
 
-        Ok(())
+        let federation_info = actor.unwrap().read().await.get_info()?;
```
oleonardolima, line 293: I wonder if the unwrap() is needed? 🤔

gateway/ln-gateway/src/lib.rs
```diff
@@ -277,7 +277,9 @@ impl Gateway {
                 .expect("Failed to build gateway client"),
         );
 
-        if let Err(e) = self.load_actor(client.clone(), route_hints).await {
+        let actor = self.load_actor(client.clone(), route_hints).await;
```
oleonardolima, line 280: I'm not sure if would be a good pattern/standard (@dpc  might correct me here 😁) but you could already get the actor, and info all in a single let federation_info.

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -165,7 +165,11 @@ macro_rules! impl_gateway_request_trait {
 }
 
 impl_gateway_request_trait!(InfoPayload, GatewayInfo, GatewayRequest::Info);
-impl_gateway_request_trait!(ConnectFedPayload, (), GatewayRequest::ConnectFederation);
+impl_gateway_request_trait!(
+    ConnectFedPayload,
+    FederationInfo,
```
oleonardolima, line 170: I think it could be `GatewayInfo` instead, but I'm not sure. @okjodom might know it for sure

gateway/ln-gateway/src/lib.rs
natto1784: thanks, followed up with a fixup

gateway/ln-gateway/src/lib.rs
```diff
@@ -277,18 +277,24 @@ impl Gateway {
                 .expect("Failed to build gateway client"),
         );
 
-        if let Err(e) = self.load_actor(client.clone(), route_hints).await {
-            error!("Failed to connect federation: {}", e);
-        }
-
         if let Err(e) = self.client_builder.save_config(client.config()) {
```
okjodom, line 280: We still want to try load actor before persisting the client configuration using `client_builder.save_config`

okjodom (approved):

jkitman (approved):

PR #2085, opened 2023-04-02
chore: change positional arguments to kwargs

### LINKED ISSUE #1562: Ban usage of positional arguments
When working with real software positional arguments are very inconvenient.

* They are not self describing, so it's just harder to read and understand them.
* When messed up by the user the feedback is poor, as it's hard to determine which thing exactly is missing and where.
* (when available) shell autocompletion can't do as good job with them.
* (AFAIU) They don't support environment variables
* They can't be reodered, which is very handy for scripting, or when one wrote a long command invocation, and forgot only one argument somewhere.

The only time when positional arguments are OK is when the command really does have a one central argument that is very primary and mandatory. Like `cat` takes file paths. It would be somewhat annoying the have to do `cat --path somefile --path someotherfile`. But that's usually only a concern for commands really meant for short-form interactive use.

I'd like us to stop using positional arguments.

GetPsyched:
Fixes #1562 

Please let me know if these changes are appropriate so that I can complete them for the other command definitions!

### DIFF
--- a/client/cli/src/lib.rs
+++ b/client/cli/src/lib.rs
@@ -358,48 +358,53 @@ enum Command {
     /// Send direct method call to the API, waiting for all peers to agree on a
     /// response
     Api {
+        #[clap(long)]
         method: String,
         /// JSON args that will be serialized and send with the request
-        #[clap(default_value = "null")]
+        #[clap(long, default_value = "null")]
         arg: String,
     },
 
     /// Issue notes in exchange for a peg-in proof
     PegIn {
-        #[clap(value_parser = from_hex::<TxOutProof>)]
+        #[clap(long, value_parser = from_hex::<TxOutProof>)]
         txout_proof: TxOutProof,
-        #[clap(value_parser = from_hex::<Transaction>)]
+        #[clap(long, value_parser = from_hex::<Transaction>)]
         transaction: Transaction,
     },
 
     /// Reissue notes received from a third party to avoid double spends
     Reissue {
-        #[clap(value_parser = parse_ecash)]
+        #[clap(long, value_parser = parse_ecash)]
         notes: TieredMulti<SpendableNote>,
     },
 
     /// Validate notes without claiming them (only checks if signatures valid,
     /// does not check if nonce unspent)
     Validate {
-        #[clap(value_parser = parse_ecash)]
+        #[clap(long, value_parser = parse_ecash)]
         notes: TieredMulti<SpendableNote>,
     },
 
     /// Prepare notes to send to a third party as a payment
     Spend {
-        #[clap(value_parser = parse_fedimint_amount)]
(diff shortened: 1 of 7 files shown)


Maan2003 (commented):
LGTM

GetPsyched:
I was wondering what should the abbreviated flag be for any given kwarg. What's the general flow to follow when choosing which arguments get what abbreviated flag while also making sure there are no conflicts?

dpc (dismissed):

Maan2003:
> I was wondering what should the abbreviated flag be for any given kwarg. What's the general flow to follow when choosing which arguments get what abbreviated flag while also making sure there are no conflicts?

lets just skip it for now. The short option is currently only being used once(gateway-cli address).

justinmoon:
Should we use kwargs for commands with 1 argument? This would seem kind of annoying to me ... but it makes sense to commands with multiple arguments. Thought @dpc?

dpc:
I'd say 99.9% named arguments. Give them `clap(env = "...")` so if it annoys anyone, they should just `export -x FM_...=value` and have it even better.

GetPsyched:
In which scenarios should I add the env alternative? IMO, having it set for context specific stuff to avoid repetition would be ideal.

GetPsyched:
I think that would be all the commands covered! Please mention if I've missed any commands/attributes that need to be switched to kwargs, I'll fix it right away.

client/cli/src/lib.rs
```diff
@@ -358,48 +358,53 @@ enum Command {
     /// Send direct method call to the API, waiting for all peers to agree on a
     /// response
     Api {
+        #[clap(long)]
         method: String,
         /// JSON args that will be serialized and send with the request
-        #[clap(default_value = "null")]
+        #[clap(long, default_value = "null")]
         arg: String,
```
Maan2003, line 365: This should be an exception. Positional arguments here feel natural

client/cli/src/lib.rs
```diff
@@ -409,27 +414,34 @@ enum Command {
...
+    WaitInvoice {
+        #[clap(long)]
+        invoice: lightning_invoice::Invoice,
+    },
 
     /// Wait for the fed to reach a consensus block height
-    WaitBlockHeight { height: u64 },
+    WaitBlockHeight {
+        #[clap(long)]
+        height: u64,
```
Maan2003, line 434: Wait block height is sufficient, we don't need extra --height

dpc:
> In which scenarios should I add the env alternative? IMO, having it set for context specific stuff to avoid repetition would be ideal.

Always use for passing secrets, otherwise arguments that are often used.

dpc (approved):

justinmoon (approved):

Maan2003:
I am surprised that this is not failing CLI tests :eyes:

PR #2087, opened 2023-04-03
feat: Add optional peer_id parameter to `fedimint-cli api` command

justinmoon:
Currently in tmuxinator you can use this command to fetch the list of gateways:

```
fedimint-cli api /module/0/list_gateways null
```

But that would check for consensus on the result. With this PR now you can just make the API call to a specific guardian. For example, to ask only the guardian with peer id 1:

```
api /module/0/list_gateways null 1
```

This can be useful to debug API calls where the results aren't enforced by consensus.

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -355,13 +355,17 @@ enum Command {
     /// Generate a new peg-in address, funds sent to it can later be claimed
     PegInAddress,
 
-    /// Send direct method call to the API, waiting for all peers to agree on a
-    /// response
+    /// Send direct method call to the API. If you specify --peer-id, it will
+    /// just ask one server, otherwise it will get consensus from all servers
     Api {
+        /// JSON-RPC method to call
         method: String,
-        /// JSON args that will be serialized and send with the request
+        /// JSON-RPC parameters for the request
         #[clap(default_value = "null")]
-        arg: String,
+        params: String,
+        /// Which server to send requst to
+        #[clap(long = "peer-id")]
+        peer_id: Option<u16>,
     },
 
     /// Issue notes in exchange for a peg-in proof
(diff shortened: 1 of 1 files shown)


client/cli/src/lib.rs
```diff
@@ -362,6 +362,9 @@ enum Command {
         /// JSON args that will be serialized and send with the request
         #[clap(default_value = "null")]
         arg: String,
+        /// Which peer to send to. If not specified, get consensus from all
+        /// peers.
+        peer_id: Option<u16>,
```
dpc, line 367: Please name it. Positional arguments are evil.

client/cli/src/lib.rs
```diff
@@ -362,6 +362,9 @@ enum Command {
         /// JSON args that will be serialized and send with the request
         #[clap(default_value = "null")]
         arg: String,
```
mxz42, line 364: btw, I think `arg` is called `param(s)` in the rest of the code. Perhaps rename it to make it (a) more consistent and (b) more similar to the RPC call `method + param`. 

just a thought, feel free to ignore

mxz42:
also we should update the [documentation of Command::Api](https://github.com/fedimint/fedimint/blob/master/client/cli/src/lib.rs#L358-L359) when this lands as we can then ask only one peer.

```rust
    /// Send direct method call to the API, waiting for all peers to agree on a
    /// response
```

mxz42:
while I was running some api requests I had a typo in the method and main panicked. I think we can wrap your the `.unwrap()` calls into something along the lines of `.map_err_cli_msg(CliErrorKind::InvalidValue, "request failed")?` to fix the first one.

errors:
```sh
# method not found
mxz@frame:~/dev/mxz/fm$ fedimint-cli api /module/0/asdf null 0
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Call(Custom(ErrorObject { code: MethodNotFound, message: "Method not found", data: None }))', /home/mxz/dev/mxz/fm/client/cli/src/lib.rs:620:26

# if the 2nd argument is not 'null'
mxz@frame:~/dev/mxz/fm$ fedimint-cli api /module/0/asdf nasdf 0
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected ident", line: 1, column: 2)', /home/mxz/dev/mxz/fm/client/cli/src/lib.rs:610:61
```

client/cli/src/lib.rs
justinmoon: Great point. Maybe we should also rename `ApiRequestErased` to `RpcParamsErased` or something? It's a little weird that `request_with_strategy` takes `method` and `params` arguments, and `params` is a "request".

justinmoon:
> while I was running some api requests I had a typo in the method and main panicked

I ended up replacing the `unwrap()` to deserialize the JSON. But I didn't replace the the ones for checking that the call actually succeeded. I think more work would be needed there in order to now swallow useful information. In the first one, the "Method not found" in the panic message is very helpful to output. Extracting the server's error message in a more human-readable way would be great but will leave that for another PR.

mxz42 (approved):
tack. 

docs are up to date, `peer-id` is positional and main doesn't panic. 

rpc params is still positional but I'd be fine with that (defaults to null).

mxz42:
cli_test_upgrade fails though

```sh
fedimint-test-all> parallel: This job failed:
fedimint-test-all> cli_test_upgrade
fedimint-test-all> Some tests failed. Full job log:
fedimint-test-all> Seq  Host    Starttime       JobRuntime      Send    Receive Exitval Signal  Command
fedimint-test-all> 1    :       1680749158.628       0.020      0       71      0       0       cli_test_always_success
fedimint-test-all> 6    :       1680749193.752      65.039      0       380653  0       0       cli_test_cli
fedimint-test-all> 3    :       1680749168.672      94.785      0       52954   0       0       cli_test_latency
fedimint-test-all> 4    :       1680749173.705      99.811      0       811930  0       0       cli_test_reconnect
fedimint-test-all> 2    :       1680749163.647     366.313      0       7008128 0       0       cli_test_rust_tests
fedimint-test-all> 5    :       1680749178.768     600.583      0       201865  -1      15      cli_test_upgrade
```

dpc:
> cli_test_upgrade fails though

This is a flake that sometimes happen and is waiting for someone to debug it. Seems like the test hangs sometimes and gets killed by a 10min timeout.

dpc (approved):

mxz42 (approved):

PR #2095, opened 2023-04-03
feat: ability to add metadata to mint backup

dpc:
We've discovered that it's useful to add extra data to backups, and absent any more general purpose backup module it's the easiest to just add a field that clients can use to put whatever they desire.

### DIFF
--- a/integrationtests/tests/tests.rs
+++ b/integrationtests/tests/tests.rs
@@ -23,6 +23,7 @@ use std::time::Duration;
 use anyhow::Result;
 use assert_matches::assert_matches;
 use bitcoin::{Amount, KeyPair};
+use fedimint_client_legacy::mint::backup::Metadata;
 use fedimint_client_legacy::mint::MintClient;
 use fedimint_client_legacy::transaction::legacy::Output;
 use fedimint_client_legacy::transaction::TransactionBuilder;
@@ -43,6 +44,7 @@ use fedimint_wallet_server::common::WalletConsensusItem::PegOutSignature;
 use fedimint_wallet_server::common::{PegOutFees, PegOutSignatureItem, Rbf};
 use fixtures::{rng, secp, sha256};
 use futures::future::{join_all, Either};
+use serde::{Deserialize, Serialize};
 use threshold_crypto::{SecretKey, SecretKeyShare};
 use tracing::log::warn;
 use tracing::{debug, info, instrument};
@@ -1219,6 +1221,41 @@ async fn rejoin_consensus_threshold_peers() -> Result<()> {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn ecash_backup_can_recover_metadata() -> Result<()> {
+    non_lightning_test(2, |_fed, user_send, _bitcoin, _, _| async move {
+        #[derive(Serialize, Deserialize)]
+        struct OurMetadata {
+            name: String,
+        }
+
+        let metadata = Metadata::from_json_serialized(OurMetadata {
+            name: "our_name".into(),
+        });
+
+        user_send
+            .client
+            .mint_client()
+            .back_up_ecash_to_federation(metadata.clone())
+            .await
+            .unwrap();
+
+        let mut task_group = TaskGroup::new();
+
+        assert_eq!(
+            user_send
+                .client
+                .mint_client()
+                .restore_ecash_from_federation(10, &mut task_group)
+                .await
+                .unwrap()
+                .unwrap(),
+            metadata
+        );
+    })
+    .await
+}
+
+#[tokio::test(flavor = "multi_thread")]
 async fn ecash_can_be_recovered() -> Result<()> {
     non_lightning_test(2, |fed, user_send, bitcoin, _, _| async move {
         let user_receive = user_send.new_user_with_peers(peers(&[0, 1, 2])).await;
@@ -1230,7 +1267,7 @@ async fn ecash_can_be_recovered() -> Result<()> {
         user_send
             .client
             .mint_client()
-            .back_up_ecash_to_federation()
+            .back_up_ecash_to_federation(Metadata::empty())
             .await
             .unwrap();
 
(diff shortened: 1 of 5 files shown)


client/cli/src/lib.rs
```diff
@@ -982,3 +991,40 @@ impl FedimintCli {
...
+                .collect::<Vec<String>>()[..]
+            {
+                [] => Err(anyhow::format_err!("Empty metadata argument not allowed")),
+                [key] => Err(anyhow::format_err!("Metadata {key} is missing a value")),
+                [key, val] => Ok((key.clone(), val.clone())),
+                [..] => unreachable!(),
+            }
+        })
+        .collect::<anyhow::Result<_>>()
+        .map_err_cli_msg(CliErrorKind::GeneralFederationError, "failed")?;
```
mxz42, line 1012: `CliErrorKind::InvalidValue` error would be better here I think?

and maybe extend `failed` to `parsing backup metadata failed`? If I understand that correctly we have the other error message regarding what exactly went wrong with the metadata argument as well, but I think it'd be nice if that error message would be a bit more descriptive

client/cli/src/lib.rs
```diff
@@ -870,14 +876,17 @@ impl FedimintCli {
...
+            Command::Backup { metadata } => {
+                let metadata = metadata_from_clap_cli(metadata)?;
+
+                cli.build_client(&self.module_gens)
+                    .await?
+                    .mint_client()
+                    .back_up_ecash_to_federation(Metadata::json_from(metadata))
+                    .await
+                    .map(|_| CliOutput::Backup)
+                    .map_err_cli_msg(CliErrorKind::GeneralFederationError, "failed")
```
mxz42, line 888: Do we know what exactly failed here (upload/network/dunno?)? Otherwise the exact error message is contained in the error itself, right?

client/client-lib/src/mint/backup.rs
```diff
@@ -321,6 +326,49 @@ impl MintClient {
     }
 }
 
+/// Backup metadata
+///
+/// A backup can have a blob of extra data encoded in it.
+/// We provide methods to use json encoding, but in clients
+/// are free to use their own encoding.
```
mxz42, line 333: nit :)

/// A backup can have a blob of extra data encoded in it.
/// We provide methods to use JSON encoding, but ~~in~~ clients
/// are free to use their own encoding.

client/client-lib/src/mint/backup.rs
```diff
@@ -321,6 +326,49 @@ impl MintClient {
...
+        self.0
+    }
+
+    /// Is metadata empty
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Create metadata as json from typed `val`
+    pub fn json_from<T: Serialize>(val: T) -> Self {
```
mxz42, line 357: isn't that usually called `from_json`?

integrationtests/tests/tests.rs
```diff
@@ -1218,6 +1220,41 @@ async fn rejoin_consensus_threshold_peers() -> Result<()> {
     .await
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn ecash_backup_can_recovery_metadata() -> Result<()> {
```
mxz42, line 1224: typo: no `y`?  _can_recover_metadata

integrationtests/tests/tests.rs
```diff
@@ -1218,6 +1220,41 @@ async fn rejoin_consensus_threshold_peers() -> Result<()> {
...
+
+        assert_eq!(
+            user_send
+                .client
+                .mint_client()
+                .restore_ecash_from_federation(10, &mut task_group)
+                .await
+                .unwrap()
+                .unwrap(),
+            metadata
```
mxz42, line 1252: hm, I found that a bit confusing that `restore_ecash_from_federation` returns the metadata now. Previously, it made sense to have `ecash` in the name as is recovered the ecash internally in the function but just return `()` (right?). Now, that the function is returning the metadata, perhaps rename it to `restore_backup_from_federation`? Internally, the functions continues to restore the ecash and it returns the metadata.

mxz42 (commented):
That's a pretty cool idea

client/cli/src/lib.rs
```diff
@@ -459,7 +460,12 @@ enum Command {
     },
 
     /// Upload the (encrypted) snapshot of mint notes to federation
-    Backup,
+    Backup {
+        #[clap(long = "metadata")]
+        /// Backup metadata, encoded as `key=value` (use `--metadata=key=value`,
+        /// possibly multiple times)
+        metadata: Vec<String>,
```
elsirion, line 467: nit: can this be directly parsed into the semantically more correct `*Map<String, String>` structure?

client/client-lib/src/mint/backup.rs
elsirion: It's doing something else semantically: taking an object, encoding it as JSON and putting the resulting bytes into metadata. I might call it `from_json_encoded`.

client/cli/src/lib.rs
dpc: The error does have the `raw_error` so the original error will be visible.

```
    fn map_err_cli_msg(self, kind: CliErrorKind, msg: impl Into<String>) -> Result<O, CliError> {
        self.map_err(|e| CliError {
            kind,
            message: msg.into(),
            raw_error: Some(e.into()),
        })
    }
```

integrationtests/tests/tests.rs
dpc: The whole thing is a bit wonky now and will need to be replaced with a general purpose backup module, so I'll just leave as is for now.

client/cli/src/lib.rs
dpc: I looked a little bit can clap do it, and I didn't see a good way. I'm adding a `TODO` to clean up later. Not a big deal, I think.

client/cli/src/lib.rs
mxz42: :+1: `"failed"` as message in an error seems a bit useless, though

since we call `.back_up_ecash_to_federation(Metadata::json_from(metadata))` perhaps `ecash backup failed`?

mxz42 (dismissed):

mxz42:
@dpc  `## FAILED: cli_test_rust_tests` fails

is that flaky, too?

elsirion:
Requires rebase now, but I think all concerns were addressed.

mxz42 (approved):

justinmoon (approved):

PR #2096, opened 2023-04-03
fix: disable codecov in-code annotations

dpc:
https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations-via-yaml


justinmoon:
That worked. https://github.com/fedimint/fedimint/pull/2095/files was full of "annotations" before rebasing on top of this.

justinmoon (approved):

PR #2104, opened 2023-04-04
WIP: Bundle mintgate

okjodom:
Uses [rust-embed](https://crates.io/crates/rust-embed) crate to bundle `gateway/ui` into `gatewayd` binary. This assumes the i is already built as a bundle-ready assets in `gateway/ui/build`

**Further consideraions**
- How would we trigger automated build of `gateway/ui` as a prerequisite of `gatewayd` build
- Would it be useful to build one binary that bundles the UI and one that doesn't?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2684,6 +2684,7 @@ dependencies = [
  "prost 0.11.6",
  "rand",
  "reqwest",
+ "rust-embed",
  "secp256k1",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


justinmoon:
> Would it be useful to build one binary that bundles the UI and one that doesn't?

I think it would probably be good to have separate nix commands for each case?

> How would we trigger automated build of gateway/ui as a prerequisite of gatewayd build

I what context? 

I don't think there's much benefit in CI unless we're trying to test a running `gatewayd` that is serving the UI. And building the assets will slow down CI.

Once this UI is ready for use, building it into docker containers would be a good idea. We'd probably want an environment variable to enable / disable the UI which could be passed into the container.

dpc:
For starter I'd add a runtime command line option (`--assets-dir=`) to the gateway, to point at it where should it look for files to serve. This will decouple build of gateway and the app.

okjodom:
will revisit this proposed bundling pattern shortly

PR #2112, opened 2023-04-04
chore(ci): more robust failure detection test

dpc:
Test if we correctly detect test failures by introducing an artificial failure and re-running the whole thing. I can't think of any more e2e way.

### DIFF
--- a/flake.crane.nix
+++ b/flake.crane.nix
@@ -313,7 +313,17 @@ rec {
     pname = "${commonCliTestArgs.pname}-all";
     version = "0.0.1";
     cargoArtifacts = workspaceBuild;
-    buildPhaseCargoCommand = "patchShebangs ./scripts ; ./scripts/test-ci-all.sh";
+    # One normal run, then if succeeded, modify the "always success test" to fail,
+    # and make sure we detect it (happened too many times that we didn't).
+    # Thanks to early termination, this should be all very quick, as we actually
+    # won't start other tests.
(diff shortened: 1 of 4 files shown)


flake.crane.nix
```diff
@@ -313,7 +313,17 @@ rec {
...
+    # One normal run, then if succeeded, modify the "always success test" to fail,
+    # and make sure we detect it (happened too many times that we didn't).
+    # Thanks to early termination, this should be all very quick, as we actually
+    # won't start other tests.
+    buildPhaseCargoCommand = ''
+      patchShebangs ./scripts
+      ./scripts/test-ci-all.sh || exit 1
+      sed -i -e 's/exit 0/exit 1/g' scripts/always-success-test.sh
+      echo "Verifing failure detection..."
+      ./scripts/test-ci-all.sh 1>/dev/null 2>/dev/null && exit 1
```
justinmoon, line 325: Why `&& exit `?

justinmoon (approved):

flake.crane.nix
dpc: I'm not sure if `set -e` is set here.

So if the `./scripts/test-ci-all.sh` fails, exit 1 will not run and the whole thing return success.

If `./scripts/test-ci-all.sh` succeeds, then `exit 1` runs and we fail.

In other words: making sure we failed after error was introduced.

Maan2003 (approved):

PR #2118, opened 2023-04-04
feat: Tmuxinator runs 4 panes tailing each fedimintd log

justinmoon:
Closes https://github.com/fedimint/fedimint/issues/1937

image

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -21,8 +21,14 @@ windows:
             - source scripts/tmux-user-shell.sh
   - federation:
       panes:
-        - federation:
-          - tail -n +0 -F $FM_LOGS_DIR/fedimint-*.log
+        - fedimint0:
+          - tail -n +0 -F $FM_LOGS_DIR/fedimintd-0.log
+        - fedimint1:
+          - tail -n +0 -F $FM_LOGS_DIR/fedimintd-1.log
+        - fedimint2:
+          - tail -n +0 -F $FM_LOGS_DIR/fedimintd-2.log
+        - fedimint3:
+          - tail -n +0 -F $FM_LOGS_DIR/fedimintd-3.log
   - lightning:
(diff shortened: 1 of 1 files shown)


.tmuxinator.yml
```diff
@@ -21,8 +21,14 @@ windows:
             - source scripts/tmux-user-shell.sh
   - federation:
       panes:
-        - federation:
-          - tail -n +0 -F $FM_LOGS_DIR/fedimint-*.log
```
justinmoon, line 25: `$FM_LOGS_DIR/fedimint-*.log` was slightly wrong. Logs are in `$FM_LOGS_DIR/fedimintd-*.log`. Missing the `d` in `fedimintd`.

mxz42 (approved):
nice, thanks :+1:

mxz42:
The distributed key generation log is also mentioned in #1937.
just as remark: it can be found in `cat $FM_LOGS_DIR/fedimint-dev.log`

justinmoon:
> The distributed key generation log is also mentioned in #1937. just as remark: it can be found in `cat $FM_LOGS_DIR/fedimint-dev.log`

Yea so after https://github.com/fedimint/fedimint/pull/1910, the "federation" tmuxinator pane no longer spawn dkg itself. Instead, dkg (along with everything else) get's spawned [here](https://github.com/fedimint/fedimint/blob/fd4ca76307ed3dcb415490db3b94f67b6930b64d/scripts/tmuxinator.sh#L25) before tmux itself is every launched. It runs in background and logs to file.

So it's a little weird that the DKG logs don't show up visually in tmuxinator anymore. But this should hopefully never fail in tmuxinator. Perhaps we could have each DKG instance log to `fedimintd-<n>.log` like `fedimintd` does? I think that would be an improvement.

Maan2003 (approved):

PR #2119, opened 2023-04-04
chore: fix failure conditions for discord notification

dpc:
Seems like `if: failure()` is true only if any previous step in the same job failed. :shrug:


justinmoon (approved):
That is very annoying!

m1sterc001guy (approved):

PR #2121, opened 2023-04-04
chore: put aliases in its own file to be easily sourced

mxz42:
When opening a new tmux pane in tmuxinator, you can easily source the aliases file with `. .aliases.sh` or `source .aliases.sh`

The core issue is that the commands in `pre_window` of tmuxinator are only executed when tmux is started, but not when I manually open a new pane or window.
```yaml
pre_window:
  - source .tmpenv
  - source scripts/lib.sh
  - source .aliases.sh
```

I thought that's annoying since I open and close panes in the tmuxinator, too, and these new manually opened panes don't have any aliases set. This way you can just source them. 

I'm not sure if we can automate the sourcing as well? perhaps somehow via nix @dpc? 

thoughts in general?
I also just copy & pasted the aliases. The restart script was (and still is) commented out for a reason?

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -3,16 +3,8 @@ root: .
 socket_name: fedimint-dev
 pre_window:
   - source .tmpenv
-  - alias lightning-cli="\$FM_LIGHTNING_CLI"
-  - alias lncli="\$FM_LNCLI"
-  - alias bitcoin-cli="\$FM_BTC_CLIENT"
-  - alias fedimint-cli="\$FM_MINT_CLIENT"
-  - alias gateway-cln="\$FM_GWCLI_CLN"
-  - alias gateway-lnd="\$FM_GWCLI_LND"
-  - alias mint_rpc_client="\$FM_MINT_RPC_CLIENT"
-  - alias dbtool="\$FM_DB_TOOL"
-  # - alias restart="./scripts/restart-tmux.sh"
   - source scripts/lib.sh
+  - source scripts/aliases.sh
 tmux_detached: false
 windows:
   - main:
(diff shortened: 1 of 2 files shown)


justinmoon:
Great idea. I've run into this as well. In general, I think a better filename than `.aliases.sh` would be `scripts/aliases.sh` so we don't crowd the root of the repo with test-specific code.

I wonder if we could setup a "tmux hook" to automatically source this file when new panes are created? https://github.com/tmuxinator/tmuxinator/issues/803#issuecomment-923265720

mxz42:
> Great idea. I've run into this as well. In general, I think a better filename than `.aliases.sh` would be `scripts/aliases.sh` so we don't crowd the root of the repo with tmuxinator-specific code.

sure, I can move it there.

> I wonder if we could setup a "tmux hook" to automatically source this file when new panes are created? [tmuxinator/tmuxinator#803 (comment)](https://github.com/tmuxinator/tmuxinator/issues/803#issuecomment-923265720)

yeah, that would be cool. 
after looking around a bit I found the following

```sh
tmux set-hook after-new-window 'run-shell "source ./scripts/aliases.sh"'
tmux set-hook after-split-window 'run-shell "source ./scripts/aliases.sh"'
``` 
These are the hooks triggered on a new pane (split-window) and a new window. However, the command with `run-shell` is run in its own shell, so sourcing the aliases there is useless, unfortunately.

I'll see if I can find a way of running commands in the shell of the new pane/window.

justinmoon (approved):

dpc (approved):

dpc:
Made https://github.com/fedimint/fedimint/issues/2134

PR #2123, opened 2023-04-05
Tmuxinator pane titles

justinmoon:
I don't think this is mergeable as-is because it sets global tmux options which will probably interfere with the local setups of people who use tmux for non-tmuxinator dev work. I need to figure out how to do `tmux set` / `tmux set-option` for just the current pane or window. There is a `-p` option that's supposed to do this, but couldn't get that to work. Tmux wizard plz halp.

But I think this is a very nice little improvement. Much easier to understand what a given pane is doing.

Based on [this](https://github.com/tmuxinator/tmuxinator/issues/873)

image

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -21,8 +21,14 @@ windows:
             - source scripts/tmux-user-shell.sh
   - federation:
       panes:
-        - federation:
-          - tail -n +0 -F $FM_LOGS_DIR/fedimint-*.log
+        - fedimint0:
+          - tail -n +0 -F $FM_LOGS_DIR/fedimintd-0.log
+        - fedimint1:
+          - tail -n +0 -F $FM_LOGS_DIR/fedimintd-1.log
+        - fedimint2:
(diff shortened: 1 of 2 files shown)


mxz42:
yes, this is nice. I looked at this about a week ago but I couldn't figure out how the pane title is not changed by the running application after we manually set it. It always showed the working directory somehow. 

When I now run tmuxinator in this branch, I get the same behaviour. 
image

I also removed my personal tmux configuration and run tmuxinator. same behaviour.

why does it stick with your setup? :D
my tmux version is `tmux 3.3a` on linux.

any ideas?

justinmoon:
When I didn't call `tmux_pane_title` in `.tmuxinator.yml` for one of the panes, it would show my hostname as a default like it's showing for you. So it seems like that function is just not having any effect for you. Weird.

dpc:
We should probably just switch to mprocs. :D

mxz42:
> We should probably just switch to mprocs. :D

never used that before

justinmoon:
> We should probably just switch to mprocs. :D

Made an issue https://github.com/fedimint/fedimint/issues/2136

PR #2124, opened 2023-04-05
feat: just exit tmuxinator alias

okjodom:
I just want to exit tmuxinator you know

### DIFF
--- a/justfile
+++ b/justfile
@@ -81,3 +81,7 @@ format:
 # start tmuxinator with a dev federation setup
 tmuxinator:
   ./scripts/tmuxinator.sh
+
+# exit tmuxinator session
+exit-tmuxinator:
+  tmux kill-session -t fedimint-dev


justfile
```diff
@@ -81,3 +81,7 @@ format:
 # start tmuxinator with a dev federation setup
 tmuxinator:
   ./scripts/tmuxinator.sh
+
+# exit tmuxinator session
+exit-tmuxinator:
+  tmux kill-session -t fedimint-dev
```
oleonardolima, line 87: @okjodom I wonder if also adding a new cmd to restart (using the existing script) would also be helpful, wdyt ? 😁

okjodom: You can just propose that alias. Would be useful!

dpc:
LGTM, but @Maan2003 is in charge.

Maan2003 (approved):
LGTM

dpc (approved):

PR #2131, opened 2023-04-05
feat: change integer encoding to big endian

douglaz:
(for context see https://github.com/fedimint/fedimint/pull/2088)

Big endianess allows us to sort the serialized bytes, which is useful for database keys

Some notes:
1) This is backwards incompatible, all databases will need to be recreated from scratch
2) Bitcoin encoding is still little endian, some further work will be required to clearly state this in the type system to avoid problems on database keys


douglaz:
Flaky test? Will rebase and try again

crypto/derive-secret/src/lib.rs
```diff
@@ -98,7 +98,9 @@ impl DerivableSecret {
 fn tagged_derive(tag: &[u8; 8], derivation: ChildId) -> [u8; 16] {
     let mut derivation_info = [0u8; 16];
     derivation_info[0..8].copy_from_slice(&tag[..]);
-    derivation_info[8..16].copy_from_slice(&derivation.0.to_le_bytes()[..]);
+    // The endianness isn't important here because we just need some bytes, but
+    // let's use the default for this project (big endian)
+    derivation_info[8..16].copy_from_slice(&derivation.0.to_be_bytes()[..]);
```
elsirion, line 103: Agree, right call

fedimint-core/src/encoding/mod.rs
```diff
@@ -716,4 +716,67 @@ mod tests {
...
+        // #[derive(Ord, PartialOrd, Eq,
+        // PartialEq, Debug, Encodable, Decodable)]
+        // struct TestNonce(secp256k1_zkp::XOnlyPublicKey);
+        // let nonces = (0..1000)
+        //     .map(|_| {
+        //         let (_, pk) = secp256k1_zkp::generate_keypair(&mut
+        // rand::rngs::OsRng);         TestNonce(pk.x_only_public_key().
+        // 0)     })
+        //     .collect::<Vec<_>>();
+        // keeps_ordering_after_serialization(nonces);
```
elsirion, line 780: Not sure if ordering is a real problem for nonces, they don't really have a semantic ordering, unlike amounts. But having the same ordering in rust using `Ord` and of the serialized value might be good indeed ... Not sure a marker trait would help much though.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -310,8 +309,7 @@ impl IBitcoindRpc for ElectrumClient {
     async fn submit_transaction(&self, transaction: Transaction) {
         fedimint_core::task::block_in_place(|| {
             let mut bytes = vec![];
-            transaction
-                .consensus_encode(&mut bytes)
+            bitcoin::consensus::Encodable::consensus_encode(&transaction, &mut bytes)
```
elsirion, line 312: Why that change? Doesn't the bridge impl still work?

elsirion (approved):
Looks good, one question about using rust-bitcoon encoding directly now

fedimint-core/src/encoding/mod.rs
douglaz: Probably the problem is just hypothetical right know, but eventually it may cause a bug. If `Nonce` isn't of a concern perhaps `TransactionId` or another bitcoin structure is.

dpc: Bitocoin's hashes are a mess one way or the other. Encoded one way, displayed in reverse. The exact expectation of the person writing the code about the ordering can't be assumed.

We could probably have some newtype adapter that wraps a `Encodable` and encodes it in reverse (or forces LE/BE explicitly) for cases like that, but probably not necessary right now. A test that serializes a block and tx hash from `rust-bitcoin` and then makes sure we've sticked to *something* should be enough.

douglaz: I've added some serialization of rust bitcoin structures

douglaz:
Added rust bitcoin encoding tests and removed the FIXME

dpc (approved):

fedimint-bitcoind/src/bitcoincore_rpc.rs
elsirion: Oh, thx for the explanation!

elsirion (approved):

PR #2133, opened 2023-04-05
chore: change positional arguments to kwargs

### LINKED ISSUE #1562: Ban usage of positional arguments

GetPsyched:
Fixes #1562 

Includes missing test script updates from #2085.
Apologies for the hiccup!

### DIFF
--- a/scripts/cli-test.sh
+++ b/scripts/cli-test.sh
@@ -45,7 +45,7 @@ $FM_MINT_CLIENT fetch
 
 # peg out
 PEG_OUT_ADDR="$($FM_BTC_CLIENT getnewaddress)"
-$FM_MINT_CLIENT peg-out $PEG_OUT_ADDR 500
+$FM_MINT_CLIENT peg-out --address $PEG_OUT_ADDR --amount 500sat
 until [ "$($FM_BTC_CLIENT getreceivedbyaddress $PEG_OUT_ADDR 0)" == "0.00000500" ]; do
   sleep $FM_POLL_INTERVAL
 done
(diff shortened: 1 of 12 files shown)


scripts/cli-test.sh
```diff
@@ -45,7 +45,7 @@ $FM_MINT_CLIENT fetch
 
 # peg out
 PEG_OUT_ADDR="$($FM_BTC_CLIENT getnewaddress)"
-$FM_MINT_CLIENT peg-out --address $PEG_OUT_ADDR --satoshis 500
+$FM_MINT_CLIENT peg-out --address $PEG_OUT_ADDR --amount 500
```
dpc, line 48: Are accepting `Amount` here? We should use units, just for clarity, so `500sat`, since `--amount` doesn't tell use the unit anymore. But anyway - not exactly

dpc:
I know it's a pain to get it in, but it's a good effort. Feel free to ping me directly on Discord when you want a review, to avoid needless conflicts.

justinmoon:
> Apologies for the hiccup!

Wasn't your fault. CI swallowed the errors.

scripts/cli-test.sh
GetPsyched: Yup, that makes sense.

dpc (approved):

dpc:
LGTM, passes tests. Ready to land?

GetPsyched:
Oh? When I last pushed, it showed that the build test had failed; that's why I still kept it as a draft. Magic.

dpc:
Not magic, but a friendly reviewer restarting a flaky run (we're working on it)

image

GetPsyched:
Oh, that makes sense. Thanks!

Maan2003 (commented):
just a nit, LGTM otherwise

dpc (dismissed):

Maan2003 (approved):
Great Work! Thanks.

dpc (approved):

PR #2139, opened 2023-04-06
chore: move client/cli to fedimint-cli

justinmoon:
Most of our crates are in the top-level directory, including our new client library, so let's pull this out as well.


dpc:
Markdown docs have some broken links now.

justinmoon:
Doc test failure will be fixed once this gets merged. It's complaining that https://github.com/fedimint/fedimint/actions/runs/4624951670/jobs/8180576877?pr=2139 doesn't exist.

mxz42:
the gateway components stay in `./gateway/` or should these be moved, too?

mxz42 (approved):

justinmoon:
Will fix in a bit

dpc (approved):

justinmoon:
> 
the gateway components stay in ./gateway/ or should these be moved, too?

I don't feel strongly about those. It makes a little more sense because the gateway is kind of it's own thing -- CLI, UI and daemon that are kinda independent of everything else. So a separate folder makes a bit more sense to me. Don't feel strongly about it.

mxz42:
> CLI, UI and daemon that are kinda independent of everything else. So a separate folder makes a bit more sense to me.

I agree :+1:

Issue #2055, opened 2023-03-28
`ModuleGen` is not a good name anymore


dpc:
We started with `ModuleConfiGen` that kind of made sense, then we shortened it to `ModuleGen`, but now these types have methods for db migrations, config validation, and so on, so the name seems off.

IMO we should:

* rename `ModuleKind` to `ModuleKindId` (to free up the name and align with `ModuleInstanceId`)
* rename `ModuleGen` to `ModuleKind`
* rename `Module` to `ModuleInstance`

When say `Module` I mean all variations. `ClientModule`, `ServerModule`, `ModuleRegistry`, `Dyn...`, `I...` everything.

dpc:
Ping @mxz42 who is a veteran of module rename wars of some time ago.

PR #2145, opened 2023-04-06
Configgen refactor

jkitman:
Simplifies and standardizes our config gen parameters.

Redundant `ServerConfigParams` is replaced by the new `ConfigGenParams` struct.

Renames some duplicate names:
- `ConfigGenParams` -> `ConfigGenModuleParams`
- `ApiEndpoint` -> `PeerUrl`

Moves `gen_local` to the test crate so it isn't used accidentally.

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -94,11 +94,10 @@ impl JsonWithKind {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Encodable)]
-pub struct ApiEndpoint {
-    /// The peer's API websocket network address and port (e.g.
-    /// `ws://10.42.0.10:5000`)
+pub struct PeerUrl {
+    /// The peer's public URL (e.g. `wss://fedimint-server-1:5000`)
     pub url: Url,
-    /// human-readable name
+    /// The peer's name
     pub name: String,
 }
 
@@ -110,7 +109,7 @@ pub struct ClientConfig {
     // Stable and unique id and threshold pubkey of the federation for authenticating configs
     pub federation_id: FederationId,
     /// API endpoints for each federation member
-    pub api_endpoints: BTreeMap<PeerId, ApiEndpoint>,
+    pub api_endpoints: BTreeMap<PeerId, PeerUrl>,
     /// Threshold pubkey for authenticating epoch history
     pub epoch_pk: threshold_crypto::PublicKey,
     /// Configs from other client modules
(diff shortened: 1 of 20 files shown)


integrationtests/tests/fixtures/mod.rs
```diff
@@ -460,6 +463,59 @@ pub async fn fixtures(num_peers: u16, gateway_node: GatewayNode) -> anyhow::Resu
     Ok(fixtures)
 }
 
+/// config for servers running on different ports on a local network
```
mxz42, line 466: > local network

meaning localhost?

fedimint-core/src/config.rs
```diff
@@ -94,11 +94,11 @@ impl JsonWithKind {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Encodable)]
-pub struct ApiEndpoint {
-    /// The peer's API websocket network address and port (e.g.
+pub struct PeerUrl {
+    /// The peer's public URL (e.g.
     /// `ws://10.42.0.10:5000`)
     pub url: Url,
```
mxz42, line 100: nit: can we move that on one line and change the websocket IP to an actual *public* one? :sweat_smile: it creates dissonance in my head reading 'public URL' but seeing an IP from 10/8

mxz42 (approved):
nice refactor. 

I get the feeling you don't like the conventional commit style? :'D

also thinking about #2055, but here the `ConfigGenModuleParams` and similar still work, right? @dpc

jkitman:
> conventional commit style
What is the conventional style?

mxz42:
[conventional commits website](https://www.conventionalcommits.org/en/v1.0.0/) and [summary](https://www.conventionalcommits.org/en/v1.0.0/#summary)

> The Conventional Commits specification is a lightweight convention on top of commit messages. It provides an easy set of rules for creating an explicit commit history; which makes it easier to write automated tools on top of. This convention dovetails with [SemVer](http://semver.org/), by describing the features, fixes, and breaking changes made in commit messages.

unless I missed something we wanted to do this?

If you'd be arguing that it's not *that* useful right now, I'd probably agree. (anyone arguing against that?) Many PR and features are moving and getting merged very fast with follow-up PRs and similar. so information belonging to one *feature* (e.g. in the commit with `feat:`) are distributed between multiple PRs/commits. 

I think this will be very helpful as soon as fedimint is used "in the real world".

mxz42 (approved):
:+1:

dpc (approved):

dpc (approved):

dpc:
> also thinking about https://github.com/fedimint/fedimint/issues/2055, but here the ConfigGenModuleParams and similar still work, right? @dpc

LGTM

Issue #2101, opened 2023-04-04
Rewriting Scripts in Rust tracking issue


Maan2003:
We have a bunch of integration tests written in bash. They are flaky, hard to write and update. We want to re-write all these tests in Rust and minimize the amount of bash we depend on.

We've made 2 PRs in this project so far:
- https://github.com/fedimint/fedimint/pull/1899 Added ability to spawn daemons (e.g. `fedimintd` and `lnd` etc) via Rust
- https://github.com/fedimint/fedimint/pull/1910 Can write tests purely in rust. As a proof-of-concept, it re-wrote latency-test.sh: 
  - [Old bash version that was replaced](https://github.com/fedimint/fedimint/blob/5ef8158163c8cfa120b235e24c07999724bc33a0/scripts/latency-test.sh) 
  - [New rust version](https://github.com/fedimint/fedimint/blob/c77e5960255843890da809b3f3ebdcccc47e27e1/fedimint-bin-tests/src/main.rs#L417-L515) which is called [here](https://github.com/fedimint/fedimint/blob/c77e5960255843890da809b3f3ebdcccc47e27e1/scripts/latency-test.sh)

Work remaining (in order of importance):
- [ ] (1) Export `$FM_LOGS_DIR` in github actions. This helps us debug logs of specific daemons. See [upload-artifact](https://github.com/actions/upload-artifact)
- [x] Rewrite the rest of the bash integrations tests:
  - [x] (2a) [scripts/cli-test.sh](https://github.com/fedimint/fedimint/blob/master/scripts/cli-test.sh) (easier)
  - [x] (2b) [reconnect tests](https://github.com/fedimint/fedimint/blob/master/scripts/reconnect-test.sh) (harder b/c need to stop and start fedimintd)
  - [x] (2c) [upgrade tests](https://github.com/fedimint/fedimint/blob/master/scripts/upgrade-test.sh) (harder b/c need to stop and start fedimintd, but 2b has now demonstrated how to do it)
- [x] (3) Replace [scripts/setup-tests.sh](https://github.com/fedimint/fedimint/blob/master/scripts/setup-tests.sh) with a rust-spawned dev federation. For example, something like [this function](https://github.com/fedimint/fedimint/blob/57c55563680d40080c776cfa0f72b3ec7449255f/fedimint-bin-tests/src/main.rs#L557-L570). Perhaps rename to dev_federation or something. blocked on (2b) (2c)
[...]

PR #2146, opened 2023-04-06
feat: rewrite reconnest-test.sh in rust

dhruv-1001:
Step towards completing #2101 (2b). This PR rewrites `reconnect-test.sh` in Rust.

### DIFF
--- a/fedimint-bin-tests/src/federation.rs
+++ b/fedimint-bin-tests/src/federation.rs
@@ -1,7 +1,7 @@
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::ops::Range;
 
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use fedimint_core::core::LEGACY_HARDCODED_INSTANCE_ID_WALLET;
 use fedimint_core::db::mem_impl::MemDatabase;
 use fedimint_core::task::TaskGroup;
@@ -45,7 +45,7 @@ fn fedimint_env(peer_id: usize) -> anyhow::Result<HashMap<String, String>> {
 pub struct Federation {
     // client is only for internal use, use cli commands instead
     client: Arc<UserClient>,
-    members: Vec<Fedimintd>,
+    members: BTreeMap<usize, Fedimintd>,
     bitcoind: Bitcoind,
 }
 
@@ -55,9 +55,9 @@ impl Federation {
         bitcoind: Bitcoind,
         ids: Range<usize>,
     ) -> Result<Self> {
-        let mut members = Vec::new();
+        let mut members = BTreeMap::new();
         for id in ids {
-            members.push(Fedimintd::new(process_mgr, bitcoind.clone(), id).await?);
+            members.insert(id, Fedimintd::new(process_mgr, bitcoind.clone(), id).await?);
         }
 
         let workdir: PathBuf = env::var("FM_DATA_DIR")?.parse()?;
@@ -78,7 +78,30 @@ impl Federation {
         })
     }
 
-    pub fn members(&self) -> &[Fedimintd] {
+    pub async fn start_server(
+        &mut self,
+        process_mgr: &ProcessManager,
+        peer_id: usize,
+    ) -> Result<()> {
+        if self.members.contains_key(&peer_id) {
+            return Err(anyhow!("fedimintd-{} already running", peer_id));
+        }
+        self.members.insert(
+            peer_id,
+            Fedimintd::new(process_mgr, self.bitcoind.clone(), peer_id).await?,
+        );
+        Ok(())
+    }
+
+    pub async fn kill_server(&mut self, peer_id: usize) -> Result<()> {
+        let Some((_, fedimintd)) = self.members.remove_entry(&peer_id) else {
+            return Err(anyhow!("fedimintd-{} does not exist", peer_id));
+        };
+        fedimintd.kill().await?;
+        Ok(())
+    }
+
+    pub fn members(&self) -> &BTreeMap<usize, Fedimintd> {
         &self.members
     }
 
@@ -130,6 +153,17 @@ impl Federation {
         Ok(())
     }
 
+    pub async fn await_all_peers(&self) -> Result<()> {
+        cmd!(
+            self,
+            "api",
+            "/module/{LEGACY_HARDCODED_INSTANCE_ID_WALLET}/block_height"
+        )
+        .run()
+        .await?;
+        Ok(())
+    }
+
     pub async fn use_gateway(&self, gw: &Gatewayd) -> Result<()> {
         let pub_key = match &gw.ln {
             ClnOrLnd::Cln(cln) => cln.pub_key().await?,
@@ -139,12 +173,20 @@ impl Federation {
         info!("Using {name} gateway", name = gw.ln.name());
         Ok(())
     }
+
+    pub async fn generate_epochs(&self, epochs: usize) -> Result<()> {
+        for _ in 0..epochs {
+            self.bitcoind.mine_blocks(10).await?;
+            self.await_block_sync().await?;
+        }
+        Ok(())
+    }
 }
 
 #[derive(Clone)]
 pub struct Fedimintd {
     _bitcoind: Bitcoind,
-    _process: ProcessHandle,
+    process: ProcessHandle,
 }
 
 impl Fedimintd {
@@ -168,7 +210,7 @@ impl Fedimintd {
         )
         .envs(env_vars);
 
-        info!("fedimintd started");
+        info!("fedimintd-{peer_id} started");
         let process = process_mgr
             .spawn_daemon(&format!("fedimintd-{peer_id}"), cmd)
             .await?;
@@ -176,8 +218,13 @@ impl Fedimintd {
         // TODO: wait for federation to start
         Ok(Self {
             _bitcoind: bitcoind,
-            _process: process,
+            process,
         })
+    }
+
+    pub async fn kill(self) -> Result<()> {
+        self.process.kill().await?;
+        Ok(())
     }
 }
 
--- a/fedimint-bin-tests/src/main.rs
+++ b/fedimint-bin-tests/src/main.rs
@@ -516,6 +516,52 @@ async fn latency_tests(dev_fed: DevFed) -> Result<()> {
     Ok(())
 }
 
+async fn reconnect_test(dev_fed: DevFed, process_mgr: &ProcessManager) -> Result<()> {
+    #[allow(unused_variables)]
+    let DevFed {
+        bitcoind,
+        cln,
+        lnd,
+        mut fed,
+        gw_cln,
+        gw_lnd,
+        electrs,
+        esplora,
+    } = dev_fed;
+
+    bitcoind.mine_blocks(110).await?;
+    fed.await_block_sync().await?;
+    fed.await_all_peers().await?;
+
+    // test a peer missing out on epochs and needing to rejoin
+    fed.kill_server(0).await?;
+    // FIXME increase this number once we can avoid processing epoch messages too
+    // far in the future
+    fed.generate_epochs(1).await?;
+
+    fed.start_server(process_mgr, 0).await?;
+    // FIXME increase this number once we can avoid processing epoch messages too
+    // far in the future
+    fed.generate_epochs(1).await?;
+    fed.await_all_peers().await?;
+    info!("Server 0 successfully rejoined!");
+    bitcoind.mine_blocks(100).await?;
+
+    // now test what happens if consensus needs to be restarted
+    fed.kill_server(1).await?;
+    bitcoind.mine_blocks(100).await?;
+    fed.await_block_sync().await?;
+    fed.kill_server(2).await?;
+    fed.kill_server(3).await?;
+
+    fed.start_server(process_mgr, 1).await?;
+    fed.start_server(process_mgr, 2).await?;
+    fed.start_server(process_mgr, 3).await?;
+    fed.await_all_peers().await?;
+    info!("fm success: reconnect-test");
+    Ok(())
+}
+
 struct DevFed {
     bitcoind: Bitcoind,
     cln: Lightningd,
@@ -640,6 +686,7 @@ use clap::{Parser, Subcommand};
 enum Cmd {
     Tmuxinator,
     LatencyTests,
+    ReconnectTest,
 }
 
 #[derive(Parser)]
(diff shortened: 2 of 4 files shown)


fedimint-bin-tests/src/federation.rs
```diff
@@ -139,6 +152,21 @@ impl Federation {
         info!("Using {name} gateway", name = gw.ln.name());
         Ok(())
     }
+
+    pub async fn kill_server(&mut self, peer_id: usize) -> anyhow::Result<()> {
+        let members = &mut self.members;
+        let fedimintd = std::mem::replace(&mut members[peer_id], None);
+        drop(fedimintd);
```
dhruv-1001, line 159: To kill the server, I'm dropping `Fedimintd`, which drops it's `ProcessHandle` and therefore killing the process.

fedimint-bin-tests/src/federation.rs
```diff
@@ -139,6 +152,21 @@ impl Federation {
         info!("Using {name} gateway", name = gw.ln.name());
         Ok(())
     }
+
+    pub async fn kill_server(&mut self, peer_id: usize) -> anyhow::Result<()> {
+        let members = &mut self.members;
+        let fedimintd = std::mem::replace(&mut members[peer_id], None);
```
Maan2003, line 158: nit: let members is needless
nit: mem::take()

fedimint-bin-tests/src/main.rs
```diff
@@ -516,6 +517,51 @@ async fn latency_tests(dev_fed: DevFed) -> Result<()> {
...
+    fed.await_block_sync().await?;
+    fed.await_all_peers().await?;
+
+    // test a peer missing out on epochs and needing to rejoin
+    fed.kill_server(0).await?;
+    // FIXME increase this number once we can avoid processing epoch messages too
+    // far in the future
+    fed.generate_epochs(1).await?;
+
+    let mut fed = Federation::new(&process_mgr, bitcoind.clone(), 0..4).await?;
```
Maan2003, line 544: does this work? 

it shouldn't because fedimintd-1,2,3 are still running

fedimint-bin-tests/src/main.rs
```diff
@@ -525,26 +571,27 @@ struct DevFed {
     gw_lnd: Gatewayd,
     electrs: Electrs,
     esplora: Esplora,
+    process_mgr: ProcessManager,
```
Maan2003, line 574: no, don't add process_mgr here.

just pass it to your function as a argument function.

fedimint-bin-tests/src/main.rs
dhruv-1001: I was confused as well as It's done the same way in `reconnect-test.sh`, it starts all the servers again. Should I change it to starting individual servers? (It does work though)

Maan2003: also important: need to be careful about log files here. 
ProcessManager::spawn creates a log file. maybe it should open file in append mode.

fedimint-bin-tests/src/main.rs
dhruv-1001: I initially did that, I think it makes more sense to have it in `DevFed`. Can you tell me why is it better to be passed as argument?

Maan2003: sorry it didn't explain
in my head, process manager is like a pseudo global, no one owns it. 

I think it is fine to put process manager to be in devfed, but I would prefer not to.

fedimint-bin-tests/src/main.rs
Maan2003: > Should I change it to starting individual servers?

yes that should be fine.

fedimint-bin-tests/src/main.rs
dhruv-1001: Oh okay no problem, i'll pass it as argument.
In my head, devfed was like a controller, with tools to start/stop deamons.

fedimint-bin-tests/src/federation.rs
```diff
@@ -78,7 +80,22 @@ impl Federation {
         })
     }
 
-    pub fn members(&self) -> &[Fedimintd] {
+    pub async fn restart_server(
+        &mut self,
+        process_mgr: &ProcessManager,
+        bitcoind: Bitcoind,
```
Maan2003, line 86: not needed, we already have bitcoind in self

fedimint-bin-tests/src/federation.rs
```diff
@@ -168,7 +204,7 @@ impl Fedimintd {
         )
         .envs(env_vars);
 
-        info!("fedimintd started");
+        info!("fedimintd-{peer_id} started");
```
Maan2003, line 207: nice

fedimint-bin-tests/src/federation.rs
```diff
@@ -78,7 +80,22 @@ impl Federation {
...
-    pub fn members(&self) -> &[Fedimintd] {
+    pub async fn restart_server(
+        &mut self,
+        process_mgr: &ProcessManager,
+        bitcoind: Bitcoind,
+        peer_id: usize,
+    ) -> Result<()> {
+        self.members[peer_id] = Some(Fedimintd::new(process_mgr, bitcoind, peer_id).await?);
+        Ok(())
+    }
```
dhruv-1001, line 91: Restarting the server fails, because it is in the process of shutting down from `kill-server()` function. So we need to wait for the child process to completely shut down. I can think of some solutions but they lead to exposing `Child` field inside `ProcessHandleInner` and that defeats the purpose of them being private. What should we do here?

fedimint-bin-tests/src/federation.rs
```diff
@@ -78,7 +80,22 @@ impl Federation {
...
+
+    pub async fn kill_server(&mut self, peer_id: usize) -> Result<()> {
+        drop(std::mem::take(&mut self.members[peer_id]));
```
Maan2003, line 94: note: drop doesn't wait for server to die.

fedimint-bin-tests/src/federation.rs
Maan2003: add `ProcessHandle::kill()`

fedimint-bin-tests/src/federation.rs
dhruv-1001: Sure, should I introduce a field `clone_cound: usize` inside of `Fedimintd` and assert that to be 0. Or is there any other (and better) way to do this, that I don't know of?

Maan2003: https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.try_unwrap

fedimint-bin-tests/src/federation.rs
```diff
@@ -45,7 +45,7 @@ fn fedimint_env(peer_id: usize) -> anyhow::Result<HashMap<String, String>> {
 pub struct Federation {
     // client is only for internal use, use cli commands instead
     client: Arc<UserClient>,
-    members: Vec<Fedimintd>,
+    members: Vec<Option<Arc<Fedimintd>>>,
```
Maan2003, line 48: no? I meant use ProcessHandle's internal arc in ProcessHandle::kill.

Sorry, if I wasn't clear.

fedimint-bin-tests/src/federation.rs
```diff
@@ -78,7 +80,29 @@ impl Federation {
         })
     }
 
-    pub fn members(&self) -> &[Fedimintd] {
+    pub async fn restart_server(
+        &mut self,
+        process_mgr: &ProcessManager,
+        peer_id: usize,
+    ) -> Result<()> {
+        self.members[peer_id] = Some(Arc::new(
```
Maan2003, line 88: important, check that it is None first.

fedimint-bin-tests/src/util.rs
```diff
@@ -14,17 +14,42 @@ use super::*;
 #[derive(Debug, Clone)]
 pub struct ProcessHandle(Arc<ProcessHandleInner>);
 
+impl ProcessHandle {
+    pub async fn kill_process(self) -> Result<()> {
+        let ProcessHandle(arc_process_handle_inner) = self;
+        let process_handle_inner = &mut Arc::try_unwrap(arc_process_handle_inner).unwrap();
+        let mut child = std::mem::take(&mut process_handle_inner.child)
+            .ok_or_else(|| anyhow!("Unable to own ProcessHandleInner"))?;
```
Maan2003, line 22: please rewrite this, i don't see a reason to mem::take here.

fedimint-bin-tests/src/util.rs
```diff
@@ -14,17 +14,42 @@ use super::*;
 #[derive(Debug, Clone)]
 pub struct ProcessHandle(Arc<ProcessHandleInner>);
 
+impl ProcessHandle {
+    pub async fn kill_process(self) -> Result<()> {
```
Maan2003, line 18: nit: rename to `kill`

fedimint-bin-tests/src/util.rs
```diff
@@ -14,17 +14,42 @@ use super::*;
...
+        child.wait().await?;
+        Ok(())
+    }
+}
+
 #[derive(Debug)]
 pub struct ProcessHandleInner {
     name: String,
-    child: Child,
+    child: Option<Child>,
```
Maan2003, line 36: not needed, you can move out of the struct in fn kill() to avoid the drop being called.

fedimint-bin-tests/src/util.rs
```diff
@@ -41,17 +66,21 @@ impl ProcessManager {
...
+        let path = format!("{logs_dir}/{name}.log");
+        let log = OpenOptions::new()
+            .append(true)
+            .create(true)
+            .open(path)
             .await?
-            .into_std()
-            .await;
+            .try_into_std()
+            .expect("Cannot open log file");
```
Maan2003, line 76: what is the difference?

fedimint-bin-tests/src/util.rs
dhruv-1001: Oh I missed that. I was dropping `ProcessHandle` and therefore I used `take`, or else it would have dropped, and we would not be able to use the Child to wait for shutdown.

Maan2003: I have explained it in a comment bellow

fedimint-bin-tests/src/util.rs
dhruv-1001: I think it is needed. As we are taking out `Fedimintd` from members: `Vec<Option<Fedimintd>>` inside the `kill-server()` method, and so it's value will eventually drop, and `ProcessHandleInner::Drop` will be called and if there is no check, it will try to kill the process again.
Am I missing something?

Maan2003: look up rust partial struct moves.

fedimint-bin-tests/src/util.rs
dhruv-1001: I'll be adding back `.into_std()`

fedimint-bin-tests/src/federation.rs
```diff
@@ -45,7 +45,7 @@ fn fedimint_env(peer_id: usize) -> anyhow::Result<HashMap<String, String>> {
 pub struct Federation {
     // client is only for internal use, use cli commands instead
     client: Arc<UserClient>,
-    members: Vec<Fedimintd>,
+    members: Vec<Option<Fedimintd>>,
```
Maan2003, line 48: nit: make it `BTreeMap<usize, Fedimintd>`

fedimint-bin-tests/src/federation.rs
```diff
@@ -78,7 +80,29 @@ impl Federation {
...
+        self.members[peer_id] =
+            Some(Fedimintd::new(process_mgr, self.bitcoind.clone(), peer_id).await?);
+        Ok(())
+    }
+
+    pub async fn kill_server(&mut self, peer_id: usize) -> Result<()> {
+        if self.members[peer_id].is_none() {
+            return Err(anyhow!("fedimintd-{} does not exist", peer_id));
+        }
+        let fedimintd = std::mem::take(&mut self.members[peer_id]).unwrap();
```
Maan2003, line 100: combine unwrap and the is_none check

fedimint-bin-tests/src/util.rs
```diff
@@ -14,17 +14,44 @@ use super::*;
...
     name: String,
-    child: Child,
+    child: Option<Child>,
 }
 
 impl Drop for ProcessHandleInner {
     fn drop(&mut self) {
+        if self.child.as_ref().is_none() {
+            return;
+        }
```
Maan2003, line 45:
```suggestion
        let Some(child) = &mut self.child else { return; };
```

fedimint-bin-tests/src/util.rs
Maan2003: update: this was just wrong, you can't move out of structs that implement drop :facepalm:

Maan2003:
Nice work!
just some small nits

Maan2003 (approved):

fedimint-bin-tests/src/util.rs
```diff
@@ -14,17 +14,36 @@ use super::*;
...
+        let mut process_handle_inner = match Arc::try_unwrap(arc_process_handle_inner) {
+            Ok(process_handler_inner) => process_handler_inner,
+            Err(_) => return Err(anyhow!("Cannot kill process because of clones")),
+        };
+        let mut child = std::mem::take(&mut process_handle_inner.child).unwrap();
+        info!("killing {}", process_handle_inner.name.clone());
+        let _ = nix::sys::signal::kill(
+            nix::unistd::Pid::from_raw(child.id().expect("pid should be present") as _),
+            nix::sys::signal::Signal::SIGTERM,
+        );
```
Maan2003, line 29: not blocking, but move this to a common function (it is used here and in drop)

fedimint-bin-tests/src/util.rs
```diff
@@ -14,17 +14,36 @@ use super::*;
...
 
+impl ProcessHandle {
+    pub async fn kill(self) -> Result<()> {
+        let arc_process_handle_inner = self.0;
+        let mut process_handle_inner = match Arc::try_unwrap(arc_process_handle_inner) {
+            Ok(process_handler_inner) => process_handler_inner,
+            Err(_) => return Err(anyhow!("Cannot kill process because of clones")),
+        };
+        let mut child = std::mem::take(&mut process_handle_inner.child).unwrap();
+        info!("killing {}", process_handle_inner.name.clone());
```
Maan2003, line 25: nit: no clone

Maan2003 (approved):

dhruv-1001:
That's strange. `CI(nix) / Build` didn't fail last time.

Maan2003:
just Build(nix) is very flaky currently, rerunning

dhruv-1001:
> just Build(nix) is very flaky currently, rerunning

And it fails again 😕

Maan2003:
_just re run till passes_

dhruv-1001:
> _just re run till passes_
I don't think I have access to do that.

Maan2003:
wait it failed again

dpc (approved):

PR #2154, opened 2023-04-06
fix: peg-in LND gateway

m1sterc001guy:
Tmuxinator doesn't current peg-in any ecash for LND, so it can't receive payments when tmuxinator starts. This change just pegs-in some ecash for LND so that it can receive payments over the Lightning Network.

### DIFF
--- a/scripts/pegin.sh
+++ b/scripts/pegin.sh
@@ -9,6 +9,9 @@ source ./scripts/lib.sh
 
 PEG_IN_AMOUNT=${PEG_IN_AMOUNT:-$1}
 USE_GATEWAY=${2:-0}
+GATEWAY_TYPE=${3:-"CLN"}
+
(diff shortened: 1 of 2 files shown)


scripts/pegin.sh
```diff
@@ -9,14 +9,17 @@ source ./scripts/lib.sh
 
 PEG_IN_AMOUNT=${PEG_IN_AMOUNT:-$1}
 USE_GATEWAY=${2:-0}
+USE_LND=${3:-0}
```
dpc, line 12: `GATEWAY_TYPE=` "LND" vs "CLN` would be more readable on the caller side.

dpc (approved):

justinmoon (approved):
Nice. Man we were really sloppy merging this LND stuff in. I know this was working at one point but didn't make it in. Tested and it works.

Issue #2152, opened 2023-04-06
Fix devcontainer build or remote it


justinmoon:
@joschisan said he tried the devcontainer recently and it didn't work. To my knowledge it hasn't been used much or at all in 6 months, and Nix is working pretty well on Mac. The devcontainer was added mostly because a developer was having difficulties running the nix dev environment on Mac.

PR #2156, opened 2023-04-06
chore: delete the dev container

dpc:
During deep dive it was pointed out that macos nix issues are largely resolved, and we don't know of any person actually using it.

It might be an extra source of confusion, and if no one is actively using it might be broken.

Closes https://github.com/fedimint/fedimint/issues/2152


Maan2003 (approved):

Maan2003:
Close #2152

justinmoon (approved):

Issue #2158, opened 2023-04-06
Do we send try to send message to ourself? 


mxz42:
Running the federation with with `RUST_LOG=none,net::peer=trace ./scripts/tmuxinator.sh`. 

I was wondering about the following lines in the fedimintd logs:

```log
2023-04-06T19:59:22.711811Z TRACE net::peer: sending message to list of peers=[PeerId(0), PeerId(1), PeerId(2), PeerId(3)]
2023-04-06T19:59:22.711833Z TRACE net::peer: Sending message to peer_id=PeerId(0)
2023-04-06T19:59:22.711840Z TRACE net::peer: Not sending message to unknown peer (maybe banned) peer=PeerId(0)
2023-04-06T19:59:22.711845Z TRACE net::peer: Sending message to peer_id=PeerId(1)
2023-04-06T19:59:22.711895Z TRACE net::peer: Sending message to peer_id=PeerId(2)
2023-04-06T19:59:22.711942Z TRACE net::peer: Sending message to peer_id=PeerId(3)
```

Aren't we/I the guardian with `PeerId(0)`? So, the connection state machine is trying to send a message to myself but can find the `PeerConnection` as it was never opened ([we filter it out](https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/net/peers.rs#L161-L182))

Do we need our own id (`identity`) in `NetworkConfig::peers` [here](https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/net/peers.rs#L67-L68) or could we remove our own PeerId from the `peers` list when building the `NetworkConfig` so we don't need to filter out our own PeerId, for example. 

<details>

<summary>more context</summary>

```log 
$ tail -n +0 -F $FM_LOGS_DIR/fedimintd-0.log
tail: cannot open '/tmp/nix-shell.lOwFde/fm-ulMs/logs/fedimintd-0.log' for reading: No such file or directory
tail: '/tmp/nix-shell.lOwFde/fm-ulMs/logs/fedimintd-0.log' has appeared;  following new file
2023-04-06T20:26:34.693565Z TRACE net::peer: Creating new PeerConnection id=PeerId(1)
2023-04-06T20:26:34.693736Z TRACE net::peer: Creating new PeerConnection id=PeerId(3)
2023-04-06T20:26:34.693802Z TRACE net::peer: Creating new PeerConnection id=PeerId(2)
[...]

PR #2159, opened 2023-04-06
chore: add tracing logs for peer connections

mxz42:
added some logs to see more context around issue #2158

### DIFF
--- a/fedimint-server/src/net/peers.rs
+++ b/fedimint-server/src/net/peers.rs
@@ -258,6 +258,7 @@ where
 {
     #[must_use]
     async fn send(&mut self, peers: &[PeerId], msg: T) -> Cancellable<()> {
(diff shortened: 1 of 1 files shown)


fedimint-server/src/net/peers.rs
```diff
@@ -258,6 +258,7 @@ where
 {
     #[must_use]
     async fn send(&mut self, peers: &[PeerId], msg: T) -> Cancellable<()> {
+        trace!(target: LOG_NET_PEER, ?peers, "sending message to list of");
```
dpc, line 261: If you intend to land it, the `to list of` could be skipped. Generally these `to`, `with` suffixes seem out of place. The message is "doing x" and doesn't need to refer to associated data.

PR #2160, opened 2023-04-06
feat: module notifications

Maan2003:


elsirion:
CI runs, ready for review?

Maan2003:
/wait_offer is still left. Will do it today.

dpc (approved):

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -748,23 +748,34 @@ impl ServerModule for Lightning {
...
+            api_endpoint! {
+                "/wait_account",
+                async |module: &Lightning, context, contract_id: ContractId| -> ContractAccount {
+                    Ok(module
+                        .wait_contract_account(context, contract_id)
+                        .await)
                 }
             },
             api_endpoint! {
                 "/offer",
```
jkitman, line 766: are we always going to expose a wait and non-wait version of the endpoints?

Maan2003: We can make it a parameter. 
We can also just remove the non wait version and make caller wrap it with a short timeout if it doesn't want to wait the full 60 seconds.

jkitman: Yeah I don't see why we don't always wait, let the client decide whether to timeout.  Would like to avoid the complexity of having 2 endpoints for every call.

fedimint-server/src/config/api.rs
```diff
@@ -521,7 +521,12 @@ impl HasApiContext<ConfigGenApi> for ConfigGenApi {
         request: &ApiRequestErased,
         id: Option<ModuleInstanceId>,
     ) -> (&ConfigGenApi, ApiEndpointContext<'_>) {
-        let dbtx = self.db.begin_transaction().await;
+        let mut db = self.db.clone();
+        let mut dbtx = self.db.begin_transaction().await;
```
m1sterc001guy, line 525: Should we just get rid of the reference to the dbtx and create a new transaction on every API call?

Maan2003: > create a new transaction on every API call

that is already the case

>  get rid of the reference

do you mean remove the `dbtx` argument?

m1sterc001guy: > do you mean remove the dbtx argument?

yeah this is what I meant

Maan2003: I tried that, ran into some lifetimes issue. It is likely possible. But not sure if it is useful.

modules/fedimint-ln-server/src/lib.rs
Maan2003: Lets do this in a follow and not block this.

jkitman (approved):

Issue #1936, opened 2023-03-15
Remove `ILnRpcClient.complete_htlc`


justinmoon:
Instead, `ILnRpcClient.subscribe_htlcs` should be able to return responses itself.

To do this, our gRPC spec will need to create a "bi-directional stream" so that `NetworkLnRpcClient.subscribe_htlcs` can return responses itself.

For LND, we should already be able to do this because the [HtlcInterceptor](https://lightning.engineering/api-docs/api/lnd/router/htlc-interceptor) API already supports a bi-directional stream.

PR #2161, opened 2023-04-06
refactor: remove subscribe htlcs and complete_htlc

m1sterc001guy:
Currently in the gateway, there are two RPCs needed for completing an HTLC: `subscribe_htlcs` and `complete_htlc`. Subscribe returns a stream which the actors use to determine if they need to take an action with the federation. If the gateway is able to get the preimage, it then calls `complete_htlc` to complete the payment. 

However, it is not necessary to have two separate RPCs and this makes the code a bit more complicated. This refactor changes it so there is only one RPC called `route_htlcs`. This RPC takes a stream as input and returns a stream. This way, the actors can send messages using the input stream rather than needing to call a separate RPC.

This allows us to clean up the code a bit. In the LND implementation, we are able to remove the `outcomes` map.

Fixes: https://github.com/fedimint/fedimint/issues/1936

### DIFF
--- a/fedimint-testing/src/ln/fixtures.rs
+++ b/fedimint-testing/src/ln/fixtures.rs
@@ -6,6 +6,7 @@ use bitcoin::hashes::{sha256, Hash};
 use bitcoin::secp256k1::{PublicKey, SecretKey};
 use bitcoin::{secp256k1, KeyPair};
 use fedimint_client_legacy::modules::ln::contracts::Preimage;
+use fedimint_core::task::TaskGroup;
 use fedimint_core::Amount;
 use futures::stream;
 use lightning::ln::PaymentSecret;
@@ -27,6 +28,7 @@ pub struct FakeLightningTest {
     pub gateway_node_pub_key: secp256k1::PublicKey,
     gateway_node_sec_key: secp256k1::SecretKey,
     amount_sent: Arc<Mutex<u64>>,
+    task_group: TaskGroup,
 }
 
 impl FakeLightningTest {
@@ -40,6 +42,7 @@ impl FakeLightningTest {
             gateway_node_sec_key: SecretKey::from_keypair(&kp),
             gateway_node_pub_key: PublicKey::from_keypair(&kp),
             amount_sent,
+            task_group: TaskGroup::new(),
         }
     }
 }
@@ -109,10 +112,22 @@ impl ILnRpcClient for FakeLightningTest {
         })
     }
 
-    async fn route_htlc<'a>(
-        &self,
-        _events: ReceiverStream<RouteHtlcRequest>,
+    async fn route_htlcs<'a>(
+        &mut self,
+        events: ReceiverStream<RouteHtlcRequest>,
     ) -> Result<RouteHtlcStream<'a>, GatewayError> {
+        self.task_group
+            .spawn("FakeRoutingThread", |handle| async move {
+                let mut stream = events.into_inner();
+                while let Some(_route_htlc) = stream.recv().await {
+                    if handle.is_shutting_down() {
+                        break;
+                    }
+                    tracing::info!("FakeLightningTest received HTLC message");
+                }
+            })
+            .await;
+
         Ok(Box::pin(stream::iter(vec![])))
     }
 }
--- a/fedimint-testing/src/ln/fixtures.rs
+++ b/fedimint-testing/src/ln/fixtures.rs
@@ -11,12 +11,13 @@ use futures::stream;
 use lightning::ln::PaymentSecret;
 use lightning_invoice::{Currency, Invoice, InvoiceBuilder, SignedRawInvoice, DEFAULT_EXPIRY_TIME};
 use ln_gateway::gatewaylnrpc::{
-    self, CompleteHtlcsRequest, CompleteHtlcsResponse, GetNodeInfoResponse, GetRouteHintsResponse,
-    PayInvoiceRequest, PayInvoiceResponse, SubscribeInterceptHtlcsRequest,
+    self, GetNodeInfoResponse, GetRouteHintsResponse, PayInvoiceRequest, PayInvoiceResponse,
+    RouteHtlcRequest,
 };
-use ln_gateway::lnrpc_client::{HtlcStream, ILnRpcClient};
+use ln_gateway::lnrpc_client::{ILnRpcClient, RouteHtlcStream};
 use ln_gateway::GatewayError;
 use rand::rngs::OsRng;
+use tokio_stream::wrappers::ReceiverStream;
 
 use super::LightningTest;
 
@@ -26,7 +27,6 @@ pub struct FakeLightningTest {
     pub gateway_node_pub_key: secp256k1::PublicKey,
     gateway_node_sec_key: secp256k1::SecretKey,
     amount_sent: Arc<Mutex<u64>>,
-    is_connected: bool,
 }
 
 impl FakeLightningTest {
@@ -40,7 +40,6 @@ impl FakeLightningTest {
             gateway_node_sec_key: SecretKey::from_keypair(&kp),
             gateway_node_pub_key: PublicKey::from_keypair(&kp),
             amount_sent,
-            is_connected: true,
         }
     }
 }
@@ -58,12 +57,6 @@ impl LightningTest for FakeLightningTest {
         amount: Amount,
         expiry_time: Option<u64>,
     ) -> ln_gateway::Result<Invoice> {
-        if !self.is_connected {
-            return Err(GatewayError::Other(anyhow::anyhow!(
-                "Error not connected to Lightning"
-            )));
-        }
-
         let ctx = bitcoin::secp256k1::Secp256k1::new();
 
         Ok(InvoiceBuilder::new(Currency::Regtest)
@@ -92,12 +85,6 @@ impl LightningTest for FakeLightningTest {
 #[async_trait]
 impl ILnRpcClient for FakeLightningTest {
     async fn info(&self) -> ln_gateway::Result<GetNodeInfoResponse> {
-        if !self.is_connected {
-            return Err(GatewayError::Other(anyhow::anyhow!(
-                "Error not connected to Lightning"
-            )));
-        }
-
         Ok(GetNodeInfoResponse {
             pub_key: self.gateway_node_pub_key.serialize().to_vec(),
             alias: "FakeLightningNode".to_string(),
@@ -105,24 +92,12 @@ impl ILnRpcClient for FakeLightningTest {
     }
 
     async fn routehints(&self) -> ln_gateway::Result<GetRouteHintsResponse> {
-        if !self.is_connected {
-            return Err(GatewayError::Other(anyhow::anyhow!(
-                "Error not connected to Lightning"
-            )));
-        }
-
         Ok(GetRouteHintsResponse {
             route_hints: vec![gatewaylnrpc::get_route_hints_response::RouteHint { hops: vec![] }],
         })
     }
 
     async fn pay(&self, invoice: PayInvoiceRequest) -> ln_gateway::Result<PayInvoiceResponse> {
-        if !self.is_connected {
-            return Err(GatewayError::Other(anyhow::anyhow!(
-                "Error not connected to Lightning"
-            )));
-        }
-
         let signed = invoice.invoice.parse::<SignedRawInvoice>().unwrap();
         *self.amount_sent.lock().unwrap() += Invoice::from_signed(signed)
             .unwrap()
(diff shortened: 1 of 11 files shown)


fedimint-testing/src/ln/fixtures.rs
```diff
@@ -134,29 +109,10 @@ impl ILnRpcClient for FakeLightningTest {
         })
     }
 
-    async fn subscribe_htlcs<'a>(
+    async fn route_htlc<'a>(
```
justinmoon, line 112: Should this be plural: `route_htlcs`?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -149,185 +276,64 @@ impl ILnRpcClient for GatewayLndClient {
...
-        } else {
-            // We should never attempt to complete an HTLC that we didn't intercept
-            panic!(
-                "No interceptor reference found for this processed htlc with id: {intercepted_htlc_id:?}",
-            );
-        }
-    }
 }
 
 fn cancel_intercepted_htlc(key: Option<CircuitKey>) -> ForwardHtlcInterceptResponse {
```
justinmoon, line 339: Could we move this into the `cancel_htlc` method above?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -79,6 +81,131 @@ impl GatewayLndClient {
...
+                while let Some(htlc) = match htlc_stream.message().await {
+                    Ok(htlc) => htlc,
+                    Err(e) => {
+                        error!("Error received over HTLC subscription: {:?}", e);
+                        None
+                    }
+                } {
+                    trace!("handling htlc {:?}", htlc);
+                    let response: Option<ForwardHtlcInterceptResponse> = if let Some(a_tx) =
+                        subs.lock().await.get(&htlc.outgoing_requested_chan_id)
```
justinmoon, line 115: It wasn't immediately clear what this lookup was doing. Perhaps extracting this into a method like `self.can_route_htlc(htlc)` would help?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -19,33 +20,25 @@ use tracing::{error, info, trace};
...
     /// Used to spawn a task handling HTLC subscriptions
     task_group: TaskGroup,
-}
 
-// Reference to a sender that forwards ForwardHtlcInterceptResponse messages to
-// LND
-type LndSenderRef = Arc<mpsc::Sender<ForwardHtlcInterceptResponse>>;
+    subscriptions: Arc<Mutex<HashMap<u64, HtlcSubscriptionSender>>>,
+
+    lnd_tx: mpsc::Sender<ForwardHtlcInterceptResponse>,
```
justinmoon, line 40: These could use doc strings

mxz42 (commented):
started a review yesterday but didn't get quite through. 

Submitting the comments now anyway hoping that they're not all out of date after the push

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -230,6 +234,86 @@ impl ClnRpcService {
...
+        } = complete_request;
+        let hash = match sha256::Hash::from_slice(&intercepted_htlc_id) {
+            Ok(hash) => hash,
+            Err(e) => {
+                error!("Invalid intercepted_htlc_id: {:?}", e);
+                return Err(Status::invalid_argument(e.to_string()));
+            }
+        };
+
+        info!("Completing htlc with reference, {:?}", hash);
```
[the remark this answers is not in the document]
m1sterc001guy: Done

gateway/ln-gateway/src/actor.rs
```diff
@@ -147,153 +143,204 @@ impl GatewayActor {
...
+                    SubscribeInterceptHtlcsRequest { short_channel_id },
+                )),
+            })
+            .await
+            .map_err(|_| GatewayError::Other(anyhow::anyhow!("Failed to subscribe to HTLCs")));
+
+        info!(
+            "Subscribed to HTLCs with short channel ID {:?}",
+            short_channel_id
+        );
```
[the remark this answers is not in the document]
m1sterc001guy: Done

gateway/ln-gateway/src/actor.rs
```diff
@@ -46,13 +47,8 @@ pub enum BuyPreimage {
...
 
-type HTLCStream = Pin<
-    Box<
-        dyn Stream<Item = std::result::Result<SubscribeInterceptHtlcsResponse, Status>>
-            + Send
-            + 'static,
-    >,
->;
+type RouteHTLCStream =
+    Pin<Box<dyn Stream<Item = std::result::Result<RouteHtlcResponse, Status>> + Send + 'static>>;
```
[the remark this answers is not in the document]
m1sterc001guy: Ah thank you, didn't know about this. Yeah I don't think the gateway needs to support wasm. Changed this to use `BoxStream`

gateway/ln-gateway/proto/gatewaylnrpc.proto
```diff
@@ -96,6 +106,17 @@ message SubscribeInterceptHtlcsResponse {
...
+  // Optional CircuitKey that LND needs to complete the HTLC
+  CircuitKey key = 12;
+}
+
+message CircuitKey {
+  // The id of the channel that the is part of this circuit.
+  uint64 chan_id = 1;
+
+  // The index of the incoming htlc in the incoming channel.
+  uint64 htlc_id = 2;
```
elsirion, line 119: Does this abstract well over implementations?

gateway/ln-gateway/src/actor.rs
```diff
@@ -147,153 +143,204 @@ impl GatewayActor {
         }
     }
 
-    pub async fn subscribe_htlcs(&mut self) -> Result<()> {
+    async fn subscribe_to_htlcs(
+        &self,
+        ln_sender: Sender<RouteHtlcRequest>,
+        short_channel_id: u64,
+    ) -> Result<()> {
```
elsirion, line 147: This fn feels like it should be called on `ln_sender` as `self` and return a stream of intercepted HTLCs. But it was like this already before, so not this PRs concern.

m1sterc001guy:
(Adding @okjodom's comments that he sent offline while he didn't have much internet access)

gateway/ln-gateway/proto/gatewaylnrpc.proto
```diff
@@ -96,6 +106,17 @@ message SubscribeInterceptHtlcsResponse {
   // A unique identifier for every intercepted HTLC
   // Used to identify an intercepted HTLC through processing and settlement
   bytes intercepted_htlc_id = 11;
+
+  // Optional CircuitKey that LND needs to complete the HTLC
+  CircuitKey key = 12;
```
m1sterc001guy, line 111: Can we generalize this? We should aim for a proto spec agnostic of LN implementation

gateway/ln-gateway/proto/gatewaylnrpc.proto
m1sterc001guy, line 119: (@okjodom ): Do we still need unique identifiers for intercepted htlcs? I suspect this can go away if we truly are not tracking intermediate state between intercepting the htlc, processing, and then completion.

m1sterc001guy: (@okjodom): Or maybe, we keep an indentifier field for which each LN impl chooses to pass a value. LND might want to store the CircuitKey on that field

m1sterc001guy: It does not currently, it is specific to LND. Let me see if there's an easy refactor so that it is generalized.

gateway/ln-gateway/src/actor.rs
m1sterc001guy: This was easy enough to do. I moved `settle_htlc` and `cancel_htlc` into the new struct too.

gateway/ln-gateway/proto/gatewaylnrpc.proto
m1sterc001guy: I changed the gRPC spec to just include the incoming channel id and the htlc id in the `SubscribeInterceptHtlcsResponse`. This is generic across all implementations. 

In LND we just pass these values onto LND to complete the HTLC. For our CLN implementation, we use it as the key into our `outcomes` map to lookup the channel sender needed to unblock the thread that will complete the HTLC.

gateway/ln-gateway/proto/gatewaylnrpc.proto
m1sterc001guy: See comment above. I removed this structure but directly put the incoming channel id and the htlc id in the `SubscribeInterceptHtlcsResponse`.

gateway/ln-gateway/proto/gatewaylnrpc.proto
m1sterc001guy: Yes, we can remove the `intercepted_htlc_id`. 

However, for our CLN extension we still need a way to uniquely identify HTLCs so we can send the preimage to the thread that is blocked waiting for the preimage. For this, we use a pair of (incoming_chan_id, htlc_id). This is also what LND uses to uniquely identify an HTLC (they call it a `CircuitKey`).

gateway/ln-gateway/src/actor.rs
```diff
@@ -157,153 +235,124 @@ impl GatewayActor {
...
-                                        intercepted_htlc_id,
-                                        action: Some(Action::Cancel(Cancel {
-                                            reason: e.to_string(),
-                                        })),
-                                    })
-                                    .await;
-                                continue;
+                            Some(route_htlc_response::Action::CompleteResponse(_complete_response)) => {
+                                // TODO: Might need to add some error handling here
+                                info!("Successfully handled HTLC");
```
justinmoon, line 346: Might be good to include some information about the HTLC that was handled

gateway/ln-gateway/src/actor.rs
```diff
@@ -46,13 +45,92 @@ pub enum BuyPreimage {
...
+        Ok(stream)
+    }
+
+    async fn settle_htlc(
+        &self,
+        preimage: Preimage,
+        incoming_chan_id: u64,
+        htlc_id: u64,
+    ) -> Result<()> {
+        info!("Successfully processed intercepted HTLC");
```
justinmoon, line 88: Might be good to add some information about the HTLC that was settled

justinmoon (dismissed):
LGTM

gateway/ln-gateway/src/actor.rs
m1sterc001guy: Gonna leave this one as a todo since we don't actually know much about this HTLC (`CompleteResponse` doesn't contain any data currently).

justinmoon (approved):

mxz42 (approved):
big change. left some comments and questions but these are better addressed in a new PR as this one is already rather big.

nice work! @m1sterc001guy

mxz42:
CI audit:
```
fedimint-workspace-audit-audit> error: 1 vulnerability found!
error: builder for '/nix/store/p7qkfpa0mxn3nzz4y303g41d69vf895h-fedimint-workspace-audit-audit-0.0.1.drv' failed with exit code 1;
       last 10 log lines:
       > Dependency tree:
       > libsqlite3-sys 0.24.2
       > └── sqlx-core 0.6.2
       >     ├── sqlx-macros 0.6.2
       >     │   └── sqlx 0.6.2
       >     │       └── fedimint-sqlite 0.1.0
       >     │           └── fedimint-tests 0.1.0
       >     └── sqlx 0.6.2
       >
       > error: 1 vulnerability found!
```

we shouldn't forget about his, either.

gateway/ln-gateway/src/actor.rs
```diff
@@ -46,13 +45,96 @@ pub enum BuyPreimage {
...
-        dyn Stream<Item = std::result::Result<SubscribeInterceptHtlcsResponse, Status>>
-            + Send
-            + 'static,
-    >,
->;
+#[derive(Clone)]
+struct LightningSenderStream {
+    ln_sender: Sender<RouteHtlcRequest>,
+    lnrpc: Arc<RwLock<dyn ILnRpcClient>>,
+}
```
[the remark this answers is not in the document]
m1sterc001guy: I'll add a docstring. My understand is RwLock is actually read-optimizied since it won't block other readers

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -509,67 +558,77 @@ fn htlc_processing_failure() -> serde_json::Value {
...
+type HtlcSubscriptionSender = mpsc::Sender<Result<RouteHtlcResponse, Status>>;
 type HtlcOutcomeSender = oneshot::Sender<serde_json::Value>;
 
 /// Functional structure to filter intercepted HTLCs into subscription streams.
 /// Used as a CLN plugin
 #[derive(Clone)]
 pub struct ClnHtlcInterceptor {
     subscriptions: Arc<Mutex<HashMap<u64, HtlcSubscriptionSender>>>,
-    pub outcomes: Arc<Mutex<HashMap<sha256::Hash, HtlcOutcomeSender>>>,
+    pub outcomes: Arc<Mutex<BTreeMap<(u64, u64), HtlcOutcomeSender>>>,
```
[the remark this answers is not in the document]
m1sterc001guy: Yes. Probably should make a new data structure to encapsulate

gateway/ln-gateway/src/actor.rs
```diff
@@ -96,39 +178,39 @@ impl GatewayActor {
         })
         .await;
 
+        // Create a channel that will be used to shutdown the HTLC thread
+        let (sender, receiver) = mpsc::channel::<Arc<AtomicBool>>(100);
```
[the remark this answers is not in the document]
m1sterc001guy: This is currently somewhat difficult because `GatewayActor` needs to be clonable. After [this refactor](https://github.com/fedimint/fedimint/issues/2192) though I think it should be possible

justinmoon:
> CI audit:
> 
> ```
> fedimint-workspace-audit-audit> error: 1 vulnerability found!
> error: builder for '/nix/store/p7qkfpa0mxn3nzz4y303g41d69vf895h-fedimint-workspace-audit-audit-0.0.1.drv' failed with exit code 1;
>        last 10 log lines:
>        > Dependency tree:
>        > libsqlite3-sys 0.24.2
>        > └── sqlx-core 0.6.2
>        >     ├── sqlx-macros 0.6.2
>        >     │   └── sqlx 0.6.2
>        >     │       └── fedimint-sqlite 0.1.0
>        >     │           └── fedimint-tests 0.1.0
>        >     └── sqlx 0.6.2
>        >
>        > error: 1 vulnerability found!
> ```
> 
> we shouldn't forget about his, either.

This got fixed here https://github.com/fedimint/fedimint/pull/2204

PR #2162, opened 2023-04-06
chore: run 'typos' in commit pre-check & improve it 

dpc:
* check only commited files (git ls-files)
* check only text files

One problem is that it's a bit slower then I'd like, but I"m hoping that https://github.com/crate-ci/typos/issues/707 could improve it.


dpc:
@ngutech21 let me know what you think

ngutech21:
What is the motivation for running typos only on non binary files? I didn't encounter any problems like false positives or performance issues by running typos on the full repository.

dpc:
> What is the motivation for running typos only on non binary files? I didn't encounter any problems like false positives or performance issues by running typos on the full repository.

I routinely keep some uncommitted files in the repo, and I was hitting issues with `typos` thinking they need fixes.

Though I guess filtering only `git ls-files` already solves that problem, and there's `--no-binary` or something like that, which I noticed later, so maybe I could remove that filtering.

flake.nix
```diff
@@ -307,6 +307,7 @@
                     cargo-udeps
                     pkgs.parallel
                     pkgs.just
+                    pkgs.linuxKernel.packages.linux_5_10.perf
```
elsirion, line 310: Is this a remnant from performance testing?

misc/git-hooks/pre-commit
```diff
@@ -151,6 +154,17 @@ function check_cargo_lock() {
 }
 export -f check_cargo_lock
 
+function check_typos() {
+  set -eo pipefail
+
+  if ! echo "$git_ls_nonbinary_files" | parallel typos {} ; then
+    >&2 echo "Typos found: Valid new words can be added to '_typos.toml'"
+    # TODO: not enforcing anything right, just being annoying in the CLI
```
elsirion, line 162: *not enforcing anything right* -> *not enforcing anything right **now***?

justfile
```diff
@@ -67,14 +67,31 @@ check-wasm:
...
+  set -eo pipefail
+
+  git_ls_files="$(git ls-files)"
+  git_ls_nonbinary_files="$(echo "$git_ls_files" | xargs file --mime | grep -v "; charset=binary" | cut -d: -f1)"
+
+  if ! echo "$git_ls_nonbinary_files" | parallel typos -w {} ; then
+    >&2 echo "Typos found: Valid new words can be added to '_typos.toml'"
+    # TODO: not enforcing anything right, just being annoying in the CLI
+    # return 1
+  fi
```
elsirion, line 94: Seems copy-pasted, could it be a function?

flake.nix
dpc: Ha, yes.

justfile
dpc: Will maybe have to move it to a script. But it's actually too slow for my liking, and I have a PR upstream to make it better anyway, so it might this PR might need to change a lot anyway. https://github.com/crate-ci/typos/pull/708

justfile
```diff
@@ -67,14 +67,31 @@ check-wasm:
...
+  set -eo pipefail
+
+  git_ls_files="$(git ls-files)"
+  git_ls_nonbinary_files="$(echo "$git_ls_files" | xargs file --mime | grep -v "; charset=binary" | cut -d: -f1)"
+
+  if ! echo "$git_ls_nonbinary_files" | parallel typos {} ; then
+    >&2 echo "Typos found: Valid new words can be added to '_typos.toml'"
+    # TODO: not enforcing anything right now, just being annoying in the CLI
+    # return 1
+  fi
```
elsirion, line 79: Am I missing something or is this identical to the `check_typos` function?

elsirion (approved):
One nit, but feel free to merge

justfile
dpc: The functionality is the same, but I don't think it's worth to complicate both trying to merge a handful of lines. And if we are going to run it in commit check every time ... then maybe we'll delete this copy soon anyway.

PR #2163, opened 2023-04-06
fix: codecov.yml invalid

dpc:
While at it, rename it to be hidden. (Docs say it's OK)

Turns out one can validate it with:

```
curl -X POST --data-binary @.codecov.yml https://codecov.io/validate
```

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -89,6 +89,9 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
+      - name: Validate codecov.yaml configuration
+        run: nix run nixpkgs#curl -- --fail-with-body -X POST --data-binary @.codecov.yml https://codecov.io/validate
+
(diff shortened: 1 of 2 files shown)


justinmoon (dismissed):

justinmoon:
FOR THE 100TH TIME

.github/workflows/ci-nix.yml
```diff
@@ -89,6 +89,9 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
+      - name: Validate codecov.yaml configuration
+        run: curl --fail-with-body -X POST --data-binary @.codecov.yml https://codecov.io/validate
```
justinmoon, line 93: Do we need to add curl to flake.nix? Or `nix run nixpkgs#curl` ;)

dpc: You're thinking like a Nixer! :+1:

Maan2003 (approved):
_hopefully_ no more codecov error

jkitman (approved):

PR #2169, opened 2023-04-07
fix: handle peer names with special chars

### LINKED ISSUE #2074: Replace invalid chars in peer_name with '_'
Right now, peer names with invalid TLS domain chars will cause failures.  It's a better user experience just to replace them with valid chars.

jkitman:
Fixes  #2074

@SumantxD took a shot at fixing it, using your string replacement code

### DIFF
--- a/fedimint-bin-tests/src/federation.rs
+++ b/fedimint-bin-tests/src/federation.rs
@@ -185,7 +185,7 @@ pub async fn run_dkg(root_task_group: &TaskGroup, servers: usize) -> anyhow::Res
     async fn create_tls(id: usize, sender: Sender<String>) -> anyhow::Result<()> {
         // set env vars
         let bin_dir = env::var("FM_BIN_DIR")?;
-        let server_name = format!("Server-{id}");
+        let server_name = format!("Server {id}!");
         let env_vars = fedimint_env(id)?;
         let p2p_url = env_vars.get("FM_P2P_URL").context("FM_P2P_URL not found")?;
         let api_url = env_vars.get("FM_API_URL").context("FM_API_URL not found")?;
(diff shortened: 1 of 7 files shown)


m1sterc001guy:
Are peer names used anywhere else besides the TLS cert? Only worry with silently replacing the bad characters with _ is if the user actually needs to use the peer name somewhere else, but now it's not what they expect.

m1sterc001guy (approved):

fedimint-bin-tests/src/federation.rs
```diff
@@ -185,7 +185,7 @@ pub async fn run_dkg(root_task_group: &TaskGroup, servers: usize) -> anyhow::Res
     async fn create_tls(id: usize, sender: Sender<String>) -> anyhow::Result<()> {
         // set env vars
         let bin_dir = env::var("FM_BIN_DIR")?;
-        let server_name = format!("Server-{id}");
+        let server_name = format!("Server {id}!");
```
dpc, line 188: I wonder if we should just

```
 info!(peer-id=peer_id, out-dir=out_dir", "creating TLS certs created for peer in {out_dir}");
```

kind of thing, instead of interpolating strings for logging.

jkitman: Isn't this being used as a cmd line arg, not logging?

jkitman:
> Are peer names used anywhere else besides the TLS cert? 

They are only replaced when used as a cert, otherwise they will be kept original everywhere else.  We just use the peer name in the TLS cert arbitrarily, it could be anything.

fedimint-bin-tests/src/federation.rs
elsirion: I don't understand that nit either, looks good to me, just test code running the dkg.

elsirion (approved):
We pin certs to peers, so actually don't care about domain names, otherwise we'd need to be careful with collisions here.

jkitman:
> otherwise we'd need to be careful with collisions here

@elsirion Certs are pinned to the P2P endpoint in config gen, shouldn't that be enough?

PR #2171, opened 2023-04-07
commented exports in script, replaced with defaults in rust in conditionals

debjit-bw:
Here I tried to implement the exports in dev_fed. For the variable vars, I used the defaults for now. #2101

### DIFF
--- a/fedimint-bin-tests/src/main.rs
+++ b/fedimint-bin-tests/src/main.rs
@@ -559,7 +559,6 @@ macro_rules! env {
 }
 
 async fn dev_fed(task_group: &TaskGroup, process_mgr: &ProcessManager) -> Result<DevFed> {
-
     let tmp_dir = "/tmp/fm-XXXXX";
     let test_dir = tmp_dir.to_string();
     let bin_dir = tmp_dir.to_string() + "/target/debug";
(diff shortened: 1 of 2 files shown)


Maan2003:
just dumping some design ideas:

```rust
// returns error on invalid utf8, Ok(None) on var=="" or var is not set. Ok(Some(value)) if it is set
fn env_var(var: String) -> Result<Option<String>>;
```
```rust
// macro to simplifiy creation of envs
macro_rules! env { ... } 

// example:
env!(MainEnv {
   FOO = env_var("FOO")?.unwrap_or("default_value"),
});


// expands to

struct MainEnv {
    FOO: String,
}

impl MainEnv {
    fn init() -> Result<Self> {
         Ok(Self { FOO: env_var("FOO")?.unwrap_or("default_value"), });
    }

   fn set_global(&self) -> Result<()> {
        env::set_var("FOO", self.FOO)?;
        Ok(())
   }
}
```

debjit-bw:
in fields like `FM_TEST_DIR = "/tmp/fm-XXXXX"`, I couldn't set variables dynamically like `FM_TEST_DIR = &test_dir`. It was saying `can't capture dynamic environment in a fn item; use the '|| { ... }' closure form instead` but using `|| test_dir` wasn't working either.

fedimint-bin-tests/src/main.rs
```diff
@@ -527,7 +527,237 @@ struct DevFed {
...
+macro_rules! env {
+    ($name:ident { $( $var:ident = $value:expr ),* $(,)? }) => {
+        struct $name {
+            $( $var: String, )*
+        }
+
+        impl $name {
+            fn init() -> Result<Self, std::env::VarError> {
+                Ok(Self {
+                    $( $var: env_var(stringify!($var).to_owned())?.unwrap_or_else(|| $value.to_owned()), )*
```
Maan2003, line 549: bind these variables to local variables, so you can use previously defined variables here

Maan2003, line 549: I think this should just be $value and `env_var(....).unwrap_or(..)` should be at call site.

PR #2172, opened 2023-04-07
Dump database on failed CI

### LINKED ISSUE #1924: On a failed CI test, dump content of client and server db
Register bash trap that on EXIT with non-zero status code, uses `dbtool <db> list` to dump the content of the DB.

Add some easily grepable headers to make locating it in the output easier.

amit-62:
Register bash trap that on EXIT and added header to find database in output.
fixes #1924


scripts/test-ci-all.sh
```diff
@@ -87,6 +87,8 @@ export -f cli_test_always_success
 tmpdir=$(mktemp --tmpdir -d XXXXX)
 trap 'rm -r $tmpdir' EXIT
 joblog="$tmpdir/joblog"
+trap 'if [ "$?" -ne 0 ]; then echo "##Client Database"; dbtool client list; echo "##Server Database"; dbtool server-0 list; fi' EXIT
```
m1sterc001guy, line 90: Does this work? I thought a path to the DB folder would be needed.

amit-62: Hello @m1sterc001guy, should I add `$FM_DATA_DIR/server-0/database`

m1sterc001guy: Give it a try in tmuxinator. I don't remember the full command, but that looks correct for dumping server-0's db.
Something like this:
`dbtool $FM_DATA_DIR/server-0/database dump -- $FM_DATA_DIR/server-0 pass0`

dpc, line 90: Can't really do it here. It needs to be done per test, as each test will run in it's own temp dir, etc. Probably you want make some

```
fn dump_db_content_on_termination() {
  # trap here
}
```

and call it somewhere early in every test.

amit-62: Hello @dpc, can I create `fn dump_db_content_on_termination()` at [test-ci-all.sh](https://github.com/fedimint/fedimint/blob/master/scripts/test-ci-all.sh#L25) and the call it all other function.

amit-62: Hello @m1sterc001guy, but in the issue it is mentioned to call `dbtool <db> list`. Or should I go ahead with your suggestion.

elsirion:
Please don't merge master into feature branches, user rebasing and force pushing instead.

scripts/test-ci-all.sh
```diff
@@ -23,9 +23,13 @@ fi
...
 cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} --all --all-targets
 
+function dump_db_content_on_termination() {
+  trap 'if [ "$?" -ne 0 ]; then echo "##Client Database"; dbtool $FM_DATA_DIR/client.db dump $FM_DATA_DIR clientpass client; echo "##Server Database"; dbtool $FM_DATA_DIR/server-0/database dump -- $FM_DATA_DIR/server-0 pass0; fi' EXIT
+}
+
 function cli_test_reconnect() {
   set -eo pipefail # pipefail must be set manually again
-  trap 'echo "## FAILED: ${FUNCNAME[0]}"' ERR 
+  trap 'dump_db_on_failure; echo "## FAILED: ${FUNCNAME[0]}"' ERR
```
dpc, line 32: `dump_db_on_failure` vs `dump_db_content_on_termination`? Also you don't need to trap here. `dump_db_content_on_termination` already does. 

For this reason I would consider renaming to `register_trap_dump_db_content_on_failure` or something.

Also you need to export the bash function to call it from these (also exported) test functions. See  e.g. `export -f cli_test_upgrade`

dpc: Should work, but make sure to try it out, by introducing some `exit 1` in one of the test scripts somewhere (after the federation db already exist).

amit-62: > `dump_db_on_failure` vs `dump_db_content_on_termination`? Also you don't need to trap here. `dump_db_content_on_termination` already does.
> 
So I need to just call this function in each test function.

scripts/test-ci-all.sh
```diff
@@ -23,9 +23,13 @@ fi
 # Note: Respect 'CARGO_PROFILE' that crane uses
 cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} --all --all-targets
 
+function dump_db_content_on_termination() {
+  trap 'if [ "$?" -ne 0 ]; then echo "##Client Database"; dbtool $FM_DATA_DIR/client.db dump $FM_DATA_DIR clientpass client; echo "##Server Database"; dbtool $FM_DATA_DIR/server-0/database dump -- $FM_DATA_DIR/server-0 pass0; fi' EXIT
```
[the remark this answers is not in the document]
amit-62: > Probably move the body to another function, so it's easier to see what's going on).
> 
Should I create another function to `trap` it, and then call that function here.
> Detect if the database exists at all and print something sane if it doesn't ("Database does not seem to exist yet, will not dump" or something like that).. If the test failed before even creating a database, we don't want to fail with confusing errors.
>

amit-62:
Hello @dpc, I have updated PR with all the improvement as suggested by you please review it.

PR #2173, opened 2023-04-07
fix: remove gatewayd reconnect expects

m1sterc001guy:
When actors in `gatewayd` are processing htlcs, they wait on two futures (one is a signal that the thread should shutdown, the other is the htlc sent from the lightning node).

There is a race condition currently where we try to send a "shutdown" signal over a channel that then returns an error. This can happen during shutdown, where the webserver and the connection to the lightning node shutdown at the same time. Currently, we have some `expects` that will crash `gatewayd` if it can't send over the channel. This PR changes it to remove the `expects` so that `gatewayd` doesn't crash in this case.

### DIFF
--- a/gateway/ln-gateway/src/actor.rs
+++ b/gateway/ln-gateway/src/actor.rs
@@ -111,10 +111,15 @@ impl GatewayActor {
 
     pub async fn stop_subscribing_htlcs(&mut self) -> Result<()> {
         if let Some(sender) = &self.sender {
-            sender
+            return sender
                 .send(Arc::new(AtomicBool::new(true)))
                 .await
-                .expect("Couldn't send signal to shutdown HTLC thread");
+                .map_err(|e| {
+                    GatewayError::Other(anyhow::anyhow!(
+                        "Couldn't send shutdown signal to HTLC thread: {:?}",
+                        e
+                    ))
+                });
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/actor.rs
```diff
@@ -111,10 +111,12 @@ impl GatewayActor {
 
     pub async fn stop_subscribing_htlcs(&mut self) -> Result<()> {
         if let Some(sender) = &self.sender {
-            sender
+            let _ = sender
```
dpc, line 114: Comment explaining why please. Helps with head scratching in the future.

gateway/ln-gateway/src/actor.rs
```diff
@@ -111,10 +111,12 @@ impl GatewayActor {
...
                 .send(Arc::new(AtomicBool::new(true)))
                 .await
-                .expect("Couldn't send signal to shutdown HTLC thread");
+                .map_err(|e| {
+                    warn!("Couldn't send shutdown signal to HTLC thread: {:?}", e);
+                });
```
elsirion, line 119: Wondering if logging+forwarding the error is an anti-pattern. Either it will be swallowed further up, in which case the swallowing code should log, or it will lead to a panic further up, logging the error too.

m1sterc001guy: Ah yeah, got a little ahead of myself here. Here we should propagate the error up. Fixed.

The other `expect` that I changed I think should be swallowed, and I added a comment. It's running in the event loop thread, so returning `None` is enough to get the thread to exit.

gateway/ln-gateway/src/actor.rs
m1sterc001guy: Added

mxz42 (approved):
:+1: CI build "fails" though

dpc (approved):

PR #2174, opened 2023-04-07
refactor: rename Reconnect to ReconnectLightning

m1sterc001guy:
`Reconnect` is ambiguous because it could mean reconnect to a federation or reconnect to the lightning node. This PR makes it more explicit by renaming structs to include "Lightning" if they're dealing with lightning


jkitman (approved):

elsirion (approved):
Makes sense to disambiguate from federation connections.

PR #2175, opened 2023-04-07
One time join codes

### LINKED ISSUE #1727: One time join codes
Add a mechanism to optionally limit the times a join code can be used. This will never be bulletproof but might be useful to avoid temporary testing federations being used too widely.

jkitman:
Allows all servers to configure a local download token that clients can use to download configs, up to a limited number to avoid too many users joining.  Refactors `WsClientConnectInfo` in a more sensible way, so that it comes from the server and is used to fetch the client config.

Figuring out how to distribute the QR code is outside the scope of this ticket, but could be built into the admin UI.

Fixes #1727


fedimint-server/src/db.rs
```diff
@@ -339,6 +366,19 @@ mod fedimint_migration_tests {
...
                                 assert!(dbtx.get_value(&ConsensusUpgradeKey).await.is_some());
                             }
+                            DbKeyPrefix::ClientConfigDownload => {
+                                let downloads = dbtx
+                                    .find_by_prefix(&ClientConfigDownloadKeyPrefix)
+                                    .await
+                                    .collect::<Vec<_>>()
+                                    .await;
+                                let _downloads_len = downloads.len();
+                                // FIXME: Why does this fail?
```
jkitman, line 376: @m1sterc001guy I could use your help figuring out why this doesn't pass.

fedimint-client-legacy/src/lib.rs
```diff
@@ -265,34 +265,6 @@ impl<T: AsRef<ClientConfig> + Clone + Send> Client<T> {
         self.config.clone()
     }
 
-    /// Verifies the config using the federation id
-    pub async fn verify_config(&self, id: &FederationId) -> Result<()> {
```
jkitman, line 269: We remove this and the tests because validation now happens when the config is downloaded initially.

fedimint-server/src/db.rs
m1sterc001guy: You just need to re-create the backup databases. You can do this by deleting the "global" folder underneath db/, modify the `create_db_with_v0_data` function to add a dummy version of your new db entry, then re-run `cargo test`.

Alternatively you can write a db migration to backup your new db entry but that's probably overkill right now.

scripts/cli-test.sh
```diff
@@ -29,12 +29,9 @@ rm $FM_DATA_DIR/client.json
...
 FED_ID="$(get_federation_id)"
-[[ "$($FM_MINT_CLIENT decode-connect-info "$CONNECT_STRING" | jq -e -r '.id')" = "${FED_ID}" ]]
-# Number required for one honest is ceil(($FM_FED_SIZE-1)/3+1)
-ONE_HONEST=2
-ONE_HONEST_URLS=$(cat $FM_DATA_DIR/client.json | jq --argjson one_honest $ONE_HONEST -e -r '.api_endpoints | to_entries[:$one_honest] | map(.value.url) | join(",")')
-[[ "$($FM_MINT_CLIENT decode-connect-info "$CONNECT_STRING" | jq -e -r '.urls | join(",")')" = "$ONE_HONEST_URLS" ]]
-[[ "$($FM_MINT_CLIENT encode-connect-info --urls $ONE_HONEST_URLS --id $FED_ID | jq -e -r '.connect_info')" = "$CONNECT_STRING" ]]
+URL=$($FM_MINT_CLIENT decode-connect-info "$CONNECT_STRING" | jq -e -r '.url')
+TOKEN=$($FM_MINT_CLIENT decode-connect-info "$CONNECT_STRING" | jq -e -r '.download_token')
+[[ "$($FM_MINT_CLIENT encode-connect-info --url $URL --download-token $TOKEN --id $FED_ID | jq -e -r '.connect_info')" = "${CONNECT_STRING}" ]]
```
dpc, line 34: Please add error messages on failure. I'm not sure why we have so many silent assertions in our shell code. Were they supposed to be debuggable through `set -x` or something?

jkitman: Sure, yeah they fail due to `set -x`.

fedimint-core/src/api.rs
```diff
@@ -616,25 +601,23 @@ impl FromStr for WsClientConnectInfo {
...
+        let url_len = u16::from_be_bytes(url_len).into();
+        let mut url_bytes = vec![0; url_len];
+        cursor.read_exact(&mut url_bytes)?;
+        let mut download_token = [0; CONFIG_DOWNLOAD_TOKEN_BYTES];
+        cursor.read_exact(&mut download_token)?;
 
-            let url = std::str::from_utf8(&url_bytes)?;
-            urls.push(url.parse()?);
-        }
+        let url = std::str::from_utf8(&url_bytes)?;
```
elsirion, line 623: Why aren't we using `Encodable`/`Decodable` here?

jkitman: This was how it was originally implemented, I guess it would be annoying for decoding to need the `ModuleDecoderRegistry`.

fedimint-server/src/consensus/mod.rs
```diff
@@ -615,6 +617,39 @@ impl FedimintConsensus {
...
+                "Download token not found".to_string(),
+            ));
+        }
+
+        let times_used = dbtx
+            .get_value(&ClientConfigDownloadKey(token.clone()))
+            .await
+            .unwrap_or_default();
+
+        dbtx.insert_entry(&ClientConfigDownloadKey(token), &(times_used + 1))
```
m1sterc001guy, line 638: Do we want to increment `times_used` here before checking if its over the limit? I haven't read what the caller is doing yet, but if we're at 101 times used, then this API is called, the request will be rejected but the value will be increment unnecessarily to 102 right?

fedimintd/src/ui.rs
```diff
@@ -68,23 +68,9 @@ async fn run_page(axum::extract::State(state): axum::extract::State<MutableState
     RunTemplate {
         state: match state.dkg_state {
             Some(DkgState::Success) => {
-                let path = state
-                    .data_dir
-                    .join(CONSENSUS_CONFIG)
-                    .with_extension(JSON_EXT);
                 // TODO: refactor be a standalone function
```
m1sterc001guy, line 71: Is this TODO still relevant?

integrationtests/tests/fixtures/mod.rs
```diff
@@ -509,6 +509,7 @@ pub fn gen_local(
                 *peer,
                 peer_params.clone(),
                 federation_name.to_string(),
+                Some(1),
```
m1sterc001guy, line 512: Nit: dont need to fix now, but this struct is getting to have a lot of parameters, difficult to determine which is which at first glance

m1sterc001guy (approved):
looks good, only made small comments

fedimintd/src/ui.rs
jkitman: Not sure on this TODO, but plan on removing this UI eventually.

justinmoon (approved):
LGTM

fedimint-core/src/api.rs
dpc: We can always create empty `ModuleDecoderRegistry` if we know no modules are going to be involved.

PR #2181, opened 2023-04-09
Improve error message

### LINKED ISSUE #2027: Improve error "Failed to start cln plugin"
It doesn't provide a reason. 

for me it was due to running two lightningd in parallel.

https://github.com/fedimint/fedimint/blob/46487a37b463a7f45f10ce7f6218583e8ef5d371/gateway/ln-gateway/src/bin/cln_extension.rs#L194-L196

isaack-njama:
Fixes #2027


gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -197,7 +197,11 @@ impl ClnRpcService {
             ))
         } else {
             Err(ClnExtensionError::Error(anyhow!(
-                "Failed to start cln plugin"
+                "
+                    Failed to start cln plugin - another instance of lightningd may already be running.
```
m1sterc001guy, line 201: Is running 2 lightningd nodes the only possible way this error message can be hit?

justinmoon: It might be better to just print out the actual error message.  But it doesn't seem like we get one? Just a `None`?

dpc, line 201: Possibly this should be just logging message

`error!(Failed to start cln plugin - another instance of lightningd may already be running.);`

?

isaack-njama: Noted

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -197,7 +197,11 @@ impl ClnRpcService {
             ))
         } else {
             Err(ClnExtensionError::Error(anyhow!(
-                "Failed to start cln plugin"
+                "
```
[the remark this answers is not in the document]
isaack-njama: I agree. The multiline string makes it a bit clunky

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: I think you still need to return an `Err` from this function though, since it is expecting a return value

dpc:
Please squash and @ me when it the CI.

Maan2003:
note: doesn't not fix #2027 

while the error message is better. It still doesn't provide the source error.

isaack-njama:
> Please squash and @ me when it the CI.

@dpc

dpc:
image


That's a merge commit. :) Please use rebase on top of `master` branch only.

justinmoon:
@isaack-njama "squashing" means there should just be 1 commit 
image

Maan2003:
needs rebase

gateway/ln-gateway/src/bin/cln_extension.rs
elsirion: I actually don't see why running two CLNs should lead to that error, but apparently that's what @Maan2003 observed, so seems like worth mentioning in that message.

elsirion (approved):
The commit message is still a bit noisy, but good enough for a first PR imo. Let's actually figure out under which conditions the plugin builder returns `None` and if we can patch upstream `cln-plugin` to return a more descriptive error.

Screenshot 2023-04-25 at 15-06-57 Improve error message by isaack-njama · Pull Request #2181 · fedimint_fedimint

gateway/ln-gateway/src/bin/cln_extension.rs
Maan2003: I was incorrectly spawning two cln nodes with same commands (same datadir) concurrently.

PR #2183, opened 2023-04-10
feat: Avoid loading all notes in memory

douglaz:
Stream data from DB in descending denomination order then only uses the required notes. Also calculates and uses a summary of the notes when some metrics like `total amount` are required.

Although this started as solution to https://github.com/fedimint/fedimint/issues/82 and may improve the speed of `select_notes`, the focus is mostly on reducing memory usage.

### DIFF
--- a/fedimint-client-legacy/src/mint/mod.rs
+++ b/fedimint-client-legacy/src/mint/mod.rs
@@ -1,5 +1,6 @@
 pub mod db;
 
+use std::cmp::Ordering;
 use std::fmt;
 use std::pin::Pin;
 use std::sync::Arc;
@@ -14,7 +15,7 @@ use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::module::{ModuleCommon, TransactionItemAmount};
 use fedimint_core::tiered::InvalidAmountTierError;
-use fedimint_core::{Amount, OutPoint, Tiered, TieredMulti, TransactionId};
+use fedimint_core::{Amount, OutPoint, Tiered, TieredMulti, TieredSummary, TransactionId};
 use fedimint_mint_client::MintModuleTypes;
 use futures::{Future, StreamExt};
 use secp256k1_zkp::{KeyPair, Secp256k1, Signing};
@@ -260,14 +261,14 @@ impl MintClient {
     ) -> (NoteIssuanceRequests, TieredMulti<BlindNonce>) {
         let mut amount_requests: Vec<((Amount, NoteIssuanceRequest), (Amount, BlindNonce))> =
             Vec::new();
-        let denominations = TieredMulti::represent_amount(
+        let denominations = TieredSummary::represent_amount(
             amount,
-            &self.notes().await,
+            &self.summary().await,
             &self.config.tbs_pks,
             notes_per_denomination,
         );
         for (amt, num) in denominations.iter() {
-            for _ in 0..*num {
+            for _ in 0..num {
                 let (request, blind_nonce) =
                     self.new_ecash_note(&self.context.secp, amt, dbtx).await;
                 amount_requests.push(((amt, request), (amt, blind_nonce)));
@@ -301,7 +302,7 @@ impl MintClient {
                 if &spend_pub_key == note.note.spend_key() {
                     Ok((note.spend_key, (amt, note.note)))
                 } else {
-                    Err(MintClientError::ReceivedUspendableNote)
+                    Err(MintClientError::ReceivedUnspendableNote)
                 }
             })
             .collect::<Result<Vec<_>>>()?;
@@ -383,13 +384,101 @@ impl MintClient {
         NoteIssuanceRequest::new(ctx, secret)
     }
 
-    pub async fn select_notes(&self, amount: Amount) -> Result<TieredMulti<SpendableNote>> {
-        let notes = self.notes().await;
-        let selected_notes = notes.select_notes(amount).ok_or_else(|| {
-            MintClientError::InsufficientBalance(amount, TieredMulti::total_amount(&notes))
-        })?;
+    pub async fn summary(&self) -> TieredSummary {
+        self.start_dbtx()
+            .await
+            .find_by_prefix(&NoteKeyPrefix)
+            .await
+            .fold(
+                TieredSummary::default(),
+                |mut acc, (key, _note)| async move {
+                    acc.inc(key.amount, 1);
+                    acc
+                },
+            )
+            .await
+    }
 
-        Ok(selected_notes)
+    /// Select notes with total amount of *at least* `amount`. If more than
+    /// requested amount of notes are returned it was because exact change
+    /// couldn't be made, and the next smallest amount will be returned.
+    ///
+    /// The caller can request change from the federation.
+    pub async fn select_notes(&self, amount: Amount) -> Result<TieredMulti<SpendableNote>> {
+        let mut dbtx = self.start_dbtx().await;
+        let note_stream = dbtx
+            .find_by_prefix_sorted_descending(&NoteKeyPrefix)
+            .await
+            .map(|(key, note)| (key.amount, note));
+        Self::select_notes_from_stream(note_stream, amount).await
+    }
+
+    // We are using a greedy algorithm to select notes. We start with the largest
+    // then proceed to the lowest tiers/denominations.
+    // But there is a catch: we don't know if there are enough notes in the lowest
+    // tiers, so we need to save a big note in case the sum of the following
+    // small notes are not enough.
+    async fn select_notes_from_stream<Note>(
+        stream: impl futures::Stream<Item = (Amount, Note)>,
+        requested_amount: Amount,
+    ) -> Result<TieredMulti<Note>> {
+        if requested_amount == Amount::ZERO {
+            return Ok(TieredMulti::default());
+        }
+        let mut stream = Box::pin(stream);
+        let mut selected = vec![];
+        // This is the big note we save in case the sum of the following small notes are
+        // not sufficient to cover the pending amount
+        // The tuple is (amount, note, checkpoint), where checkpoint is the index where
+        // the note should be inserted on the selected vector if it is needed
+        let mut last_big_note_checkpoint: Option<(Amount, Note, usize)> = None;
+        let mut pending_amount = requested_amount;
+        let mut previous_amount: Option<Amount> = None; // used to assert descending order
+        loop {
+            if let Some((note_amount, note)) = stream.next().await {
+                assert!(
+                    previous_amount.map_or(true, |previous| previous >= note_amount),
+                    "notes are not sorted in descending order"
+                );
+                previous_amount = Some(note_amount);
+                match note_amount.cmp(&pending_amount) {
+                    Ordering::Less => {
+                        // keep adding notes until we have enough
+                        pending_amount -= note_amount;
+                        selected.push((note_amount, note))
+                    }
+                    Ordering::Greater => {
+                        // probably we don't need this big note, but we'll keep it in case the
+                        // following small notes don't add up to the
+                        // requested amount
+                        last_big_note_checkpoint = Some((note_amount, note, selected.len()));
+                    }
+                    Ordering::Equal => {
+                        // exactly enough notes, return
+                        selected.push((note_amount, note));
+                        return Ok(selected.into_iter().collect());
+                    }
+                }
+            } else {
+                assert!(pending_amount > Amount::ZERO);
+                if let Some((big_note_amount, big_note, checkpoint)) = last_big_note_checkpoint {
+                    // the sum of the small notes don't add up to the pending amount, remove
+                    // them
+                    selected.truncate(checkpoint);
+                    // and use the big note to cover it
+                    selected.push((big_note_amount, big_note));
+                    // so now we have enough to cover the requested amount, return
+                    return Ok(selected.into_iter().collect());
+                } else {
+                    let total_amount = requested_amount - pending_amount;
+                    // not enough notes, return
+                    return Err(MintClientError::InsufficientBalance(
+                        requested_amount,
+                        total_amount,
+                    ));
+                }
+            }
+        }
     }
 
     pub async fn receive_notes<'a, F, Fut>(
@@ -623,7 +712,7 @@ pub enum MintClientError {
     #[error("The transaction outcome returned by the mint has an invalid type (output {0})")]
     InvalidOutcomeType(OutPoint),
     #[error("One of the notes meant to be spent is unspendable")]
-    ReceivedUspendableNote,
+    ReceivedUnspendableNote,
 }
 
 impl MintClientError {
@@ -662,12 +751,14 @@ mod tests {
     use fedimint_core::db::Database;
     use fedimint_core::module::registry::ModuleDecoderRegistry;
     use fedimint_core::outcome::{SerdeOutputOutcome, TransactionStatus};
-    use fedimint_core::{Amount, OutPoint, ServerModule, Tiered, TransactionId};
+    use fedimint_core::{Amount, OutPoint, ServerModule, Tiered, TieredMulti, TransactionId};
     use fedimint_mint_server::{Mint, MintGen, MintGenParams};
     use fedimint_testing::FakeFed;
     use futures::executor::block_on;
+    use itertools::Itertools;
     use tokio::sync::Mutex;
 
+    use super::*;
     use crate::api::fake::FederationApiFaker;
     use crate::mint::db::NextECashNoteIndexKey;
     use crate::mint::MintClient;
@@ -815,7 +906,11 @@ mod tests {
         const ISSUE_AMOUNT: Amount = Amount::from_sats(12);
         issue_notes(&fed, &client, &context.db, ISSUE_AMOUNT).await;
 
-        assert_eq!(client.notes().await.total_amount(), ISSUE_AMOUNT)
+        assert_eq!(client.summary().await.total_amount(), ISSUE_AMOUNT)
+    }
+
+    fn notes_distribution(summary: &TieredSummary) -> Vec<(Amount, usize)> {
+        summary.iter().collect::<Vec<_>>()
     }
 
     #[test_log::test(tokio::test)]
@@ -833,6 +928,18 @@ mod tests {
         };
 
         issue_notes(&fed, &client, &context.db, SPEND_AMOUNT * 2).await;
+        let summary = client.summary().await;
+        let notes = client.notes().await;
+        assert_eq!(notes.total_amount(), SPEND_AMOUNT * 2);
+        assert_eq!(summary.total_amount(), notes.total_amount());
+        assert_eq!(
+            notes_distribution(&summary),
+            vec![
+                (Amount::from_sats(1), 2),
+                (Amount::from_sats(10), 2),
+                (Amount::from_sats(20), 1)
+            ]
+        );
 
         // Spending works
         let mut dbtx = client.context.db.begin_transaction().await;
@@ -841,6 +948,11 @@ mod tests {
         let _tbs_pks = &client.config.tbs_pks;
         let rng = rand::rngs::OsRng;
         let notes = client.select_notes(SPEND_AMOUNT).await.unwrap();
+        let summary = notes.summary();
+        assert_eq!(
+            notes_distribution(&summary),
+            vec![(Amount::from_sats(1), 1), (Amount::from_sats(20), 1)]
+        );
         let (spend_keys, ecash_input) = MintClient::ecash_input(notes.clone()).unwrap();
 
         builder.input(&mut spend_keys.clone(), ecash_input.clone());
@@ -871,9 +983,13 @@ mod tests {
                 .await
                 .consensus_round(&[input.clone()], &[])
                 .await;
-
+            let summary = client.summary().await;
             // The right amount of money is left
-            assert_eq!(client.notes().await.total_amount(), SPEND_AMOUNT);
+            assert_eq!(summary.total_amount(), SPEND_AMOUNT);
+            assert_eq!(
+                notes_distribution(&summary),
+                vec![(Amount::from_sats(1), 1), (Amount::from_sats(10), 2)]
+            );
 
             // Double spends aren't possible
             assert!(fed.lock().await.verify_input(&input).await.is_err());
@@ -883,6 +999,11 @@ mod tests {
         let mut dbtx = client.context.db.begin_transaction().await;
         let mut builder = TransactionBuilder::default();
         let notes = client.select_notes(SPEND_AMOUNT).await.unwrap();
+        let summary = client.summary().await;
+        assert_eq!(
+            notes_distribution(&summary),
+            vec![(Amount::from_sats(1), 1), (Amount::from_sats(10), 2)]
+        );
         let rng = rand::rngs::OsRng;
         let (spend_keys, ecash_input) = MintClient::ecash_input(notes).unwrap();
 
@@ -910,7 +1031,7 @@ mod tests {
             );
 
             // No money is left
-            assert_eq!(client.notes().await.total_amount(), Amount::ZERO);
+            assert_eq!(client.summary().await.total_amount(), Amount::ZERO);
         }
     }
 
@@ -972,7 +1093,6 @@ mod tests {
                 let output = t.join().unwrap();
                 // Most threads will have produces far less than ITERATIONS items notes due to
                 // database transactions failing
-                output.len();
                 output
             })
             .collect::<Vec<_>>();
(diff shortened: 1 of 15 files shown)


fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -383,13 +384,87 @@ impl MintClient {
...
+        stream: impl futures::Stream<Item = (Amount, Note)>,
+        requested_amount: Amount,
+    ) -> Result<TieredMulti<Note>> {
+        if requested_amount == Amount::ZERO {
+            return Ok(TieredMulti::default());
+        }
+        let mut stream = Box::pin(stream);
+        let mut selected = vec![];
+        let mut total_amount = Amount::ZERO;
+        let mut last_big_note_checkpoint: Option<(Amount, Note, usize)> = None;
```
dpc, line 426: While reading it from top to bottom: looks important, but also confusing. Comment here is probably very valuable.

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -383,13 +384,87 @@ impl MintClient {
...
+                    pending_amount -= note_amount;
+                    selected.push((note_amount, note))
+                }
+                Ordering::Equal => {
+                    // exactly enough notes, return
+                    selected.push((note_amount, note));
+                    return Ok(selected.into_iter().collect());
+                }
+                Ordering::Greater => {
+                    // probably we don't need this big note, but we'll keep it in case the following
```
dpc, line 448: Wait, what? Did 430 assert that they are sorted? So why do we still need checkpoints? If Equal and Greater should both mean "we're done"?

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -383,13 +384,87 @@ impl MintClient {
...
+            assert!(
+                previous_amount.map_or(true, |previous| previous >= note_amount),
+                "notes are not sorted in descending order"
+            );
+            previous_amount = Some(note_amount);
+            total_amount += note_amount;
+            match note_amount.cmp(&pending_amount) {
+                Ordering::Less => {
+                    // keep adding notes until we have enough
+                    pending_amount -= note_amount;
```
dpc, line 439: I don't get why we have both `total` and `pending` and why `pending` is not modified  in `Equal` branch...

dpc: Oh. Because we early-return there anyway...

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -383,13 +384,87 @@ impl MintClient {
...
+        if requested_amount > total_amount {
+            return Err(MintClientError::InsufficientBalance(
+                requested_amount,
+                total_amount,
+            ));
+        }
+        let (big_note_amount, big_note, checkpoint) =
+            last_big_note_checkpoint.expect("to have at least one big note");
+        // we can't make exact change, remove all small notes after the big note
+        selected.truncate(checkpoint);
```
dpc, line 464: I'm confused what's going here. Either there's something wrong here, or this function might need a bit more comments to explain things for a bit slower (yet still valuable!) developers like myself. :D

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -994,4 +1100,109 @@ mod tests {
...
+                MintClient::select_notes_from_stream(stream, Amount::from_sats(multiplier * 1000))
+                    .await;
+            total_notes += select.unwrap().into_iter_items().count();
+        }
+        assert_eq!(total_notes / 100, 10);
+    }
+
+    #[test_log::test(tokio::test)]
+    async fn select_notes_returns_exact_amount_with_minimum_notes() {
+        let f = || {
```
dpc, line 1124: `.clone` a `Vec<_>` instead of lambda?

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -994,4 +1100,109 @@ mod tests {
...
+
+    fn sorted_note_stream(
+        notes: Vec<(Amount, usize)>,
+    ) -> impl futures::Stream<Item = (Amount, usize)> {
+        futures::stream::iter(
+            notes
+                .into_iter()
+                .flat_map(|(amount, number)| vec![(amount, 0usize); number])
+                .sorted()
+                .rev(),
```
dpc, line 1198: `.rev` confuses me here. Said it's sorted... I didn't expect it to be in reversed order.

fedimint-core/src/db/mod.rs
```diff
@@ -351,8 +351,16 @@ pub trait IDatabaseTransaction<'a>: 'a + MaybeSend {
 
     async fn raw_remove_entry(&mut self, key: &[u8]) -> Result<Option<Vec<u8>>>;
 
+    /// Returns an stream of key-value pairs with keys that start with
+    /// `key_prefix`. No particular ordering is guaranteed.
```
dpc, line 355: Do we still need this one? Isn't every sane db going to return in order anyway? Hmmm... Maybe not ... if it was like an SQL doing collscan without a key ... no ... but then we always give a "prefix" *of the key*, right? So I can't imagine this not being in order ...

fedimint-core/src/db/mod.rs
```diff
@@ -578,6 +602,19 @@ impl<'a> ISingleUseDatabaseTransaction<'a> for CommittableIsolatedDatabaseTransa
...
 
+    async fn raw_find_by_prefix_sorted_descending(
+        &mut self,
+        key_prefix: &[u8],
+    ) -> Result<PrefixStream<'_>> {
+        let mut isolated = IsolatedDatabaseTransaction::new(self.dbtx.as_mut(), Some(self.prefix));
+        let stream = isolated
+            .raw_find_by_prefix_sorted_descending(key_prefix)
+            .await?
+            .collect::<Vec<_>>()
```
dpc, line 613: This seems to collect into a `Vec`, which seems undesirable. I see that `raw_find_by_prefix` was doing it already, but isn't this ruining the benefits of streaming? We are still going to load all the matching notes at once into memory, just fake later that it was streaming ...

fedimint-core/src/tiered_multi.rs
```diff
@@ -82,14 +84,14 @@ impl<T> TieredMulti<T> {
...
     pub fn structural_eq<O>(&self, other: &TieredMulti<O>) -> bool {
-        let tier_eq = self.0.keys().eq(other.0.keys());
-        let per_tier_eq = self
-            .0
-            .values()
-            .zip(other.0.values())
-            .all(|(c1, c2)| c1.len() == c2.len());
-
-        tier_eq && per_tier_eq
+        let tier_eq = || self.0.keys().eq(other.0.keys());
```
dpc, line 87: Oh. Lambdas again. You really seem to like lambdas. :)

Is it in hopes of making `per_tier_eq` lazily evaluated in the `&&`? Wouldn't compiler probably be able to see through this expression having no side-effects and optimize anyway? Hmmm... maybe... maybe not ...

Anyway - if there's a good reason, please add a comment, otherwise someone is going to revert it back eventually.

fedimint-rocksdb/src/lib.rs
```diff
@@ -51,6 +51,27 @@ impl From<RocksDb> for rocksdb::OptimisticTransactionDB {
     }
 }
 
+// When finding by prefix iterating in Reverse order, we need to start from
+// "prefix+1" instead of "prefix", using lexicographic ordering. See the tests
+// below
+fn next_prefix(prefix: &[u8]) -> Result<Vec<u8>> {
+    let mut next_prefix = prefix.to_vec();
```
dpc, line 58: Thought - this fn could be performance sensitive ... maybe? ... slightly?

I wonder if passing `prefix: &mut [u8]` to avoid this clone wouldn't make sense?

fedimint-rocksdb/src/lib.rs
```diff
@@ -51,6 +51,27 @@ impl From<RocksDb> for rocksdb::OptimisticTransactionDB {
...
+    let mut increased = false;
+    for i in (0..next_prefix.len()).rev() {
+        if next_prefix[i] < 0xff {
```
dpc, line 61: `if next_prefix[i] == 0xff` seems easier to grasp.

fedimint-rocksdb/src/lib.rs
```diff
@@ -51,6 +51,27 @@ impl From<RocksDb> for rocksdb::OptimisticTransactionDB {
...
+            next_prefix[i] += 1;
```
dpc, line 62: Isn't this going to panic on wrap in debug mode?

fedimint-rocksdb/src/lib.rs
```diff
@@ -51,6 +51,27 @@ impl From<RocksDb> for rocksdb::OptimisticTransactionDB {
...
+            increased = true;
+            break;
+        } else {
+            next_prefix[i] = 0;
+        }
+    }
+    if !increased {
+        bail!("Prefix is already the max one: {prefix:?}")
```
dpc, line 70: I don't think we should panic here. Seems like handling `0xff...` should be taken care of. Return an `Option` maybe?

fedimint-rocksdb/src/lib.rs
dpc: That `increased` is kind of meh name. We always going to increase that number ... kind of ... . Maybe `non_zero_digit`? Especially with my version it seems to be a bit more clear.

fedimint-rocksdb/src/lib.rs
```diff
@@ -51,6 +51,27 @@ impl From<RocksDb> for rocksdb::OptimisticTransactionDB {
     }
 }
 
+// When finding by prefix iterating in Reverse order, we need to start from
+// "prefix+1" instead of "prefix", using lexicographic ordering. See the tests
+// below
+fn next_prefix(prefix: &[u8]) -> Result<Vec<u8>> {
```
dpc, line 57: This is kind of a `increment_bytes` thing, no? Not even prefix specific.

fedimint-sqlite/src/lib.rs
```diff
@@ -136,6 +136,34 @@ impl<'a> IDatabaseTransaction<'a> for SqliteDbTransaction<'a> {
         Box::pin(stream::iter(rows))
     }
 
+    async fn raw_find_by_prefix_sorted_descending(
+        &mut self,
+        key_prefix: &[u8],
+    ) -> Result<PrefixStream<'_>> {
+        let mut str_prefix = "".to_string();
+        for prefix in key_prefix {
+            str_prefix = format!("{str_prefix}{prefix:02X?}");
```
dpc, line 145: This seems like a meh way to format this prefix allocating new string for every byte.

```
format!("{}%", to_hex(key_prefix));
```

kind of thing should work instead?

fedimint-sqlite/src/lib.rs
```diff
@@ -136,6 +136,34 @@ impl<'a> IDatabaseTransaction<'a> for SqliteDbTransaction<'a> {
...
+        }
+        str_prefix = format!("{}{}", str_prefix, "%");
+        let query = "SELECT key, value FROM kv WHERE hex(key) LIKE ? ORDER BY key DESC, value DESC";
+        let query_prepared = sqlx::query(query).bind(str_prefix);
+        let results = self.tx.fetch_all(query_prepared).await;
+
+        if results.is_err() {
+            warn!("sqlite find_by_prefix failed to retrieve key range. Returning empty iterator");
```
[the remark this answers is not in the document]
dpc: It's not mandatory, but please consider making it better (here and in existing version) while at it. :)

dpc (commented):
I had plenty of comments (one of them being the actual streaming of notes not working like streaming and fetching all notes anyway), but putting them aside, on the higher level...

What are we optimize for exactly here, and can we get some benchmark?

I assume that if we are going to stream the notes, then there's going to be some extra overhead of "give me more". When fetching all notes at once, I'd assume sqlite does everything once, returns it and that's about it. With an iterator (depending on how it is actually implement it) it could be: "give me one; here it goes; give me another one; here it goes; and another one please; there you go...". Each back and forth by necessity must have *some* overhead. Maybe negligible, maybe amortized, but still *some*.

With fetching all, we obviously waste (briefly, but still) some memory (allocation), and we the db needs to overfetch some data.

But with iterator we might actually be doing many more IO operations scattered over longer time. Total load on the system might be higher, especially if typically we need to fetch .. donno... 75% of all notes anyway (given that big notes are ... well... big, so there need to be fewer of them to compose a certain amount, one would expect).

If you have `1, 2, 4, 8` notes, you have 15 msats total, but if you make payments in `4-7` range you usually need to fetch all but one note anyway. 

So I think it's fair to ask - is it worth it? The number of notes is not going to be sky high. I suspect 50-150 range, because:

```
pub const DEFAULT_MAX_NOTES_PER_DENOMINATION: u16 = 3;
```

and we power-of-2 tiers by default, so we can't be going all that far w.r.t number of tier.

fedimint-client-legacy/src/mint/mod.rs
dpc: Oh. We're going from the largest notes first ...

dpc:
**Oooohhh... I was confused.** So the algorithm is going from largest to smallest ...

But doesn't have the opposite problem? If you need to make a odd payment ... you're going to need to fetch almost all notes anyway, no? Because you need that `1`-value note, that will come last, no matter what you do...

dpc:
Honestly, it seems to me that instead of streaming notes, if performance is the goal, it would be best to just put an in-memory cache in front of that notes in database thing, and always have it memory after initial load. :shrug:

fedimint-core/src/db/mod.rs
douglaz: Nice catch, this was an oversight.

fedimint-rocksdb/src/lib.rs
douglaz: Not sure, what `increment_bytes`?

fedimint-rocksdb/src/lib.rs
douglaz: This is called just once per find by prefix and has only one or two bytes.

It won't really avoid cloning because we need the original prefix too.

fedimint-rocksdb/src/lib.rs
douglaz: It will never wrap because it's guarded by the `if`

fedimint-rocksdb/src/lib.rs
dpc: `fn increment_bytes(prefix: &[u8])`? Or did we misunderstand each other? :D

fedimint-rocksdb/src/lib.rs
douglaz: It isn't panic, it's an error. In fact it could be a panic because it's a bug (an invalid prefix has been created).

But if we return an option the caller should do the same, IMHO.

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -994,4 +1100,109 @@ mod tests {
...
+        ));
+    }
+
+    fn sorted_note_stream(
+        notes: Vec<(Amount, usize)>,
+    ) -> impl futures::Stream<Item = (Amount, usize)> {
+        futures::stream::iter(
+            notes
+                .into_iter()
+                .flat_map(|(amount, number)| vec![(amount, 0usize); number])
```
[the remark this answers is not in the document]
dpc: Could it be `()` if it doesn't mean anything? Or it has to be `0usize`? :dagger: 

I still don't get it, but whatever - not very important.

fedimint-sqlite/src/lib.rs
douglaz: Yeah, it seems inneficient

fedimint-rocksdb/src/lib.rs
dpc: > It won't really avoid cloning because we need the original prefix too.

Caller can do it if they need to, and it makes the code here easier, IMO. But not a blocker.

fedimint-rocksdb/src/lib.rs
douglaz: Oh, I think it's weird and specific enough to be called `next_prefix` instead of something more generic.

fedimint-sqlite/src/lib.rs
```diff
@@ -136,6 +136,34 @@ impl<'a> IDatabaseTransaction<'a> for SqliteDbTransaction<'a> {
...
+            return Ok(Box::pin(stream::iter(Vec::new())));
+        }
+
+        let rows = results.unwrap().into_iter().map(|row| {
```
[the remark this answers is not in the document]
douglaz: Don't ask me, I've just copied the code :sweat_smile:

douglaz:
> I had plenty of comments (one of them being the actual streaming of notes not working like streaming and fetching all notes anyway), but putting them aside, on the higher level...
> 
> What are we optimize for exactly here, and should we get some benchmarks?
Average/max memory usage and average I/O operations. Benchmarks would be nice, but I guess tricky to get?

> 
> I assume that if we are going to stream the notes, then there's going to be some extra overhead of "give me more". 
This may or may not be not true. But even if it's true, not that the new code won't add more overhead for a rocksdb database. It will just read less of the existing iterator.

> When fetching all notes at once, I'd assume sqlite does everything once, returns it and that's about it. With an iterator (depending on how it is actually implement it) it could be: "give me one; here it goes; give me another one; here it goes; and another one please; there you go...". Each back and forth by necessity must have _some_ overhead. Maybe negligible, maybe amortized, but still _some_.
> 
This is a good point. Some notes:
1) Both the new code and the old code will sort the data on sqlite (But you could argue there is a possible optimization where no sorting could be done)
2) Both the old code and the new code won't really stream on sqlite (I didn't change the fetchall)
3) Given the point above you can see I treated sqlite as a second citizen database. The whole idea was focused on rocksdb-like databases (where data is naturally sorted by key).

So this implementation doesn't really make anything better for databases like sqlite, but it also doesn't make it worse performance-wise.

> With fetching all, we obviously waste (briefly, but still) some memory (allocation), and we the db needs to overfetch some data.
> 
> But with iterator we might actually be doing many more IO operations scattered over longer time. Total load on the system might be higher, especially if typically we need to fetch .. donno... 75% of all notes anyway (given that big notes are ... well... big, so there need to be fewer of them to compose a certain amount, one would expect).
Look at the implementation of `fetch_all`, it's iterators/streams all way down. So it won't require more IO operations.

> If you have `1, 2, 4, 8` notes, you have 15 msats total, but if you make payments in `4-7` range you usually need to fetch all but one note anyway. So I expect, that even while having a decent balance, and making smallish payments, still a good chunk of notes will need to be fetched?
> 
> So I think it's fair to ask - is it worth it? The number of notes is not going to be sky high. I suspect 50-150 range, because:
> 
> ```
> pub const DEFAULT_MAX_NOTES_PER_DENOMINATION: u16 = 3;
> ```
> 
> and we power-of-2 tiers by default, so we can't be going all that far w.r.t number of tier.
If the number of notes is and will always be small then it's fair to say: don't bother. 

Is this really the case? So there is not such thing as a "big wallet"?

douglaz:
> It seems to me that instead of streaming notes, if performance is the goal, it would be best to just put an in-memory write-through cache in front of that notes in database thing, and always have it memory after initial load. shrug

Although the new code have the same or better performance (as far as it can be determined by simple inspection), I can't say it's the goal. See the comment above. But I agree, putting everything in memory will probably be faster if you don't bother amount memory consumption.

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -383,13 +384,87 @@ impl MintClient {
...
+    /// requested amount of notes are returned it was because exact change
+    /// couldn't be made, and the next smallest amount will be returned.
+    ///
+    /// The caller can request change from the federation.
     pub async fn select_notes(&self, amount: Amount) -> Result<TieredMulti<SpendableNote>> {
-        let notes = self.notes().await;
-        let selected_notes = notes.select_notes(amount).ok_or_else(|| {
-            MintClientError::InsufficientBalance(amount, TieredMulti::total_amount(&notes))
-        })?;
+        let mut t = self.start_dbtx().await;
```
m1sterc001guy, line 408: nit: I prefer `dbtx` for var names since fedimint has multiple definitions of a "transaction" and its easy to get confused (likely not in this function but good to be consistent)

fedimint-client-legacy/src/mint/mod.rs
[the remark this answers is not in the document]
m1sterc001guy: Comment would be helpful, I am confused too

fedimint-core/src/db/mod.rs
```diff
@@ -351,8 +351,16 @@ pub trait IDatabaseTransaction<'a>: 'a + MaybeSend {
 
     async fn raw_remove_entry(&mut self, key: &[u8]) -> Result<Option<Vec<u8>>>;
 
+    /// Returns an stream of key-value pairs with keys that start with
+    /// `key_prefix`. No particular ordering is guaranteed.
     async fn raw_find_by_prefix(&mut self, key_prefix: &[u8]) -> PrefixStream<'_>;
```
m1sterc001guy, line 356: Don't need to fix in this PR but we should probably make this return `Result<PrefixStream<'_>> so that the API is consistent.

fedimint-sqlite/src/lib.rs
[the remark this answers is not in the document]
m1sterc001guy: If I remember correctly, it was throwing an error :) There might be a better API to use as opposed to `fetch_all` but at the time this was the easiest to get it to work.

fedimint-sqlite/src/lib.rs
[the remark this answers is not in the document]
m1sterc001guy: As mentioned above, we should probably also make `raw_find_by_prefix` return a `Result` so that we can just return the error here.

m1sterc001guy:
> Given the point above you can see I treated sqlite as a second citizen database. The whole idea was focused on rocksdb-like databases (where data is naturally sorted by key).

The data in Sqlite is also sorted by key. We [create a index](https://github.com/fedimint/fedimint/blob/0e015ce4f92b3ffedd09a80807ffdf80c23c4297/fedimint-sqlite/src/lib.rs#L48) on the key and the hex representation of the key when the database is created. 

> Both the old code and the new code won't really stream on sqlite (I didn't change the fetchall)

Looks like the main limitation is converting from sqlite's row structure to a (key, value) pair. As you pointed out `fetch_all` is already using streams. 

Sqlite definitely needs some love as its not tested much right now. With the new client I hope Sqlite will be easier to support so any improvement in that direction is appreciated :)

fedimint-core/src/db/mod.rs
douglaz: This is trickier than it seems. The borrow checker isn't happy if you remove the collect. I know it is feasible because if I inline the code of `IsolatedDatabaseTransaction` then it works:  https://github.com/fedimint/fedimint/pull/2188

@m1sterc001guy any ideas here?

fedimint-sqlite/src/lib.rs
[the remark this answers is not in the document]
douglaz: I'll take a closer look at sqlite, perhaps we can improve that

fedimint-core/src/db/mod.rs
m1sterc001guy: This is where I got stuck too :) Unless @dpc has ideas I think its fine to just inline the isolated code here

douglaz:
> @douglaz Currently we use a greedy note selection algorithm, but in the future we may want to choose notes that have a large enough anonymity set. How would this approach change?

Just to be clear, this PR opens up possibilities, it doesn't restrict you in any way. More specifically:
1) It formalizes that notes can be streamed up from DB sorted by denomination amount. This was already the case but it was actually broken due to small implementation details.
2) It formalizes a `TieredSummary` when you just need to know how many notes of each denomination you have. This structure was mostly there (without this fancy name), but now the code calculates it without loading the full notes on memory. A further optimization would be, as suggested by @elsirion elsewhere, to skip or delay decoding the full `SpendableNote` on this and other cases.
3) Uses 1) to implement the greedy selection algorithm by descending denomination amount, reading only a subset of the notes in many cases (reducing IO) and keeping the minimal number of notes on memory (reducing memory usage)

Now if you want to implement a algorithm focused on common notes you can use the idea of 1) and 3) but perhaps reading in ascending denomination order, focusing on low denomination notes first. Then you have at least two options:
A) If you have an up to date `TieredSummary`, you iterate over the stream keeping exactly what you need, stopping as soon as you are done
B) Without a `TieredSummary, the accumulator keeps all of read lower denomination notes, stopping as soon as you are done

And of course you can do:
C) Same as B), but always reading until the end, without stopping earlier

Note that the current code on master implements C focusing on high denomination notes. This PR still keeps that possible but, besides implementing 3), it also makes it easier to implement A) and B)

(Other than all that, if you have a `TieredSummary` on hand, there are other possibilities involving more complex DB queries, but I won't detail it now)

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -383,13 +384,87 @@ impl MintClient {
...
+            .find_by_prefix(&NoteKeyPrefix)
+            .await
+            .fold(
+                TieredSummary::default(),
+                |mut acc, (key, _note)| async move {
+                    acc.inc(key.amount, 1);
+                    acc
+                },
+            )
+            .await
```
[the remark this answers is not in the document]
elsirion: Forgot to send, seems you addressed in the meantime.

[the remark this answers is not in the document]
douglaz: > Forgot to send, seems you addressed in the meantime.

Yeah, it's isn't implemented because it would require more refactorings, but it's certainly a possibility to keep in mind.

douglaz:
Rebased

douglaz:
Addressed code review

fedimint-client-legacy/src/mint/mod.rs
douglaz: Take a look at the new comments

fedimint-client-legacy/src/mint/mod.rs
douglaz: It's fine as is

fedimint-client-legacy/src/mint/mod.rs
douglaz: Renamed

fedimint-client-legacy/src/mint/mod.rs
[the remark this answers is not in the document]
douglaz: See now

fedimint-core/src/db/mod.rs
douglaz: Will be addressed on https://github.com/fedimint/fedimint/pull/2202

fedimint-core/src/db/mod.rs
```diff
@@ -1000,6 +1055,36 @@ impl<'parent> DatabaseTransaction<'parent> {
...
+        debug!("find by prefix sorted descending");
+        let decoders = self.decoders.clone();
+        let prefix_bytes = key_prefix.to_bytes();
+        self.tx
+            .raw_find_by_prefix_sorted_descending(&prefix_bytes)
+            .await
+            .expect("Error doing prefix search in database")
+            .map(move |(key_bytes, value_bytes)| {
+                let key = KP::Record::from_bytes(&key_bytes, &decoders)
+                    .expect("Unrecoverable error reading DatabaseKey");
```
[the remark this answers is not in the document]
douglaz: Added context

fedimint-rocksdb/src/lib.rs
douglaz: Code gets ugly if you change that

fedimint-rocksdb/src/lib.rs
douglaz: Take a look now

fedimint-sqlite/src/lib.rs
douglaz: Replaced by `hex::encode`

fedimint-sqlite/src/lib.rs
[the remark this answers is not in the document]
douglaz: I've just added a FIXME, this look suitable to be improved in a future PR

fedimint-sqlite/src/lib.rs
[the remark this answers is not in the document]
douglaz: I've just added a FIXME, this looks suitable to be improved in a future PR

fedimint-core/src/db/mod.rs
douglaz: In fact it's blocked by that PR because `raw_find_by_prefix_sorted_descending` will be implemented in the same way

elsirion: This requires a rebase now I assume?

fedimint-rocksdb/src/lib.rs
elsirion: Huh, why if `0xff` an invalid prefix? We might declare it as such because it's convenient, but it's not obviously true to me. 

Correct me if I'm wrong, but in that case we might want to return the empty prefix `vec![]` so that we start iterating from the absolutely last element when iterating in reverse order. Naturally the `0xff` prefix should come first then.

fedimint-sqlite/Cargo.toml
```diff
@@ -15,6 +15,7 @@ sqlx = { version = "0.7.0-alpha.2", default-features = false, features = ["runti
 fedimint-core ={ path = "../fedimint-core" }
 rand = "0.8"
 tracing = "0.1.37"
+hex = "0.4.3"
```
elsirion, line 18: We tried to get rid of the `hex` dependency some time ago and instead used the impl inside `bitcoin_hashes`. Not sure if worth the hassle though … @mxz42

fedimint-sqlite/src/lib.rs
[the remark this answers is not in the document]
elsirion: You already made the return type a result by now it seems, could we return the error here now?

douglaz:
Rebased

douglaz:
Addressed code review

fedimint-client-legacy/src/mint/mod.rs
```diff
@@ -383,13 +384,107 @@ impl MintClient {
...
+                        Ordering::Equal => {
+                            // exactly enough notes, return
+                            selected.push((note_amount, note));
+                            return Ok(selected.into_iter().collect());
+                        }
+                    }
+                }
+                None => {
+                    assert!(pending_amount > Amount::ZERO);
+                    if requested_amount > total_amount {
```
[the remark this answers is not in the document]
douglaz: I thought it would make things more clearer, but it isn't strictly necessary. Changed to not use it

fedimint-sqlite/Cargo.toml
mxz42: Not sure either, but [at some point](https://github.com/rust-bitcoin/rust-bitcoin/pull/1573) we can depend on their `hex` crate as we depend on `bitcoin_hashes` anyway. 
we also use the [`format_hex`](https://docs.rs/bitcoin_hashes/0.11.0/bitcoin_hashes/hex/fn.format_hex.html) macro in a few places.

elsirion (approved):
Looks good to me, referring to @m1sterc001guy for the DB specific impls.

m1sterc001guy (approved):

PR #2187, opened 2023-04-10
chore: update advisory-db on every run

dpc:
Turns out we can update `advisory-db` before every run and it doesn't cause a whole project rebuild, so we can always run with an up to date state.

However - it is failing and probably will often start failing out of the blue with no immediate fix, so I'm making it a non-required job.

It will show like that:

image

which does not prevent merging a PR, and should not report alerts on Discord `#alert` channel. But it will show every PR status as a red-x (just like codecov often does).

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -59,6 +59,9 @@ jobs:
         run: nix build -L .#ci.workspaceDoc
 
       - name: Run cargo audit
+        # sometimes we can't fix these immediately, yet
+        # we don't want to stop the world because of it
+        continue-on-error: true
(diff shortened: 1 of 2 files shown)


Maan2003 (approved):

jkitman (approved):
Thanks for adding, will be important for this to always pass when we have a real release.

dpc:
> Why does `cargo audit` even need `cargo build` isn't audit just a function of `Cargo.lock`?

I doesn't. I don't think we are running `cargo build` anymore for it.

In the past I missed the fast that crane's audit support doesn't even need workspace build etc. I should have moved it to another CI job earlier.

PR #2188, opened 2023-04-11
fix: Do not collect on raw_find_by_prefix

douglaz:


fedimint-core/src/db/mod.rs
```diff
@@ -569,13 +569,24 @@ impl<'a> ISingleUseDatabaseTransaction<'a> for CommittableIsolatedDatabaseTransa
     }
 
     async fn raw_find_by_prefix(&mut self, key_prefix: &[u8]) -> Result<PrefixStream<'_>> {
-        let mut isolated = IsolatedDatabaseTransaction::new(self.dbtx.as_mut(), Some(self.prefix));
-        let stream = isolated
-            .raw_find_by_prefix(key_prefix)
-            .await?
-            .collect::<Vec<_>>()
```
elsirion, line 576: Wow, I didn't remember this at all :grimacing: great catch!

elsirion (approved):

elsirion:
Just needs a `clippy` fix apparently, then it should be good to go, great find! :)

m1sterc001guy:
+1 Would prefer https://github.com/fedimint/fedimint/pull/2202

douglaz:
https://github.com/fedimint/fedimint/pull/2202 is the way

PR #2193, opened 2023-04-11
fix: broken wasm database

Maan2003:


fedimint-core/src/db/mod.rs
```diff
@@ -393,7 +393,7 @@ pub trait IDatabaseTransaction<'a>: 'a + MaybeSend {
 /// `ISingleUseDatabaseTransaction` without needing to make additional
 /// allocations.
 #[apply(async_trait_maybe_send!)]
-pub trait ISingleUseDatabaseTransaction<'a>: 'a + Send {
+pub trait ISingleUseDatabaseTransaction<'a>: 'a + MaybeSend {
```
m1sterc001guy, line 396: Is it possible to write a wasm test so that we catch these issues in CI going forward?

m1sterc001guy: Let's make an issue so we don't forget

justinmoon (approved):

fedimint-core/src/db/mod.rs
Maan2003: I have mentioned it in #2194

Issue #1603, opened 2023-02-07
Improve performance of the Gateway main loop?


okjodom:
Consider using [async recv()](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv) to remove manual sleep in the [main gateway loop](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/gatewayd/gateway.rs#L407). This might bring a performance improvement to the gateways since messages are handled on arrival, rather than after some arbitrary sleep timer schedule

For additional context, see [this suggestion by @mxz42](https://github.com/fedimint/fedimint/pull/1337#pullrequestreview-1269722967)

mxz42:
I think it would be similar as in [`peers.rs`](https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/net/peers.rs#L163-L169). We create a shutdown handler and then use `tokio::select` to choose where to make progress; also note the `while` loop for the shutdown check.
[Another similar example is used in `task.rs`](https://github.com/fedimint/fedimint/blob/master/fedimint-api/src/task.rs#L126-L148) 

also wrote up a [rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2b09a0ae46dda2e88ca59f99eaed351d). I think it would work somewhat like this.

mxz42:
a similar loop exists in the [HTLC interception](https://github.com/fedimint/fedimint/blob/d3f7d45e1dbf4fc016ba176bbbdd0bcf9af83c56/gateway/ln-gateway/src/gatewayd/actor.rs#L107)

PR #2195, opened 2023-04-11
fix: remove sleep in gatewayd web server event loop

m1sterc001guy:
For `gatewayd`'s web server event loop, we currently have two loops with a `sleep` in between. This is not necessary, it is simplier to just block on the channel that is waiting for messages. This PR adds a new message that is sent during shutdown.

Fixes: https://github.com/fedimint/fedimint/issues/1603

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -16,7 +16,7 @@ use std::net::SocketAddr;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use anyhow::anyhow;
 use axum::http::StatusCode;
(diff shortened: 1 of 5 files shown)


justinmoon (approved):
Nice

gateway/ln-gateway/src/lib.rs
```diff
@@ -511,112 +511,117 @@ impl Gateway {
...
+                    // ignore errors: we don't care if already stopped
+                    let _ = shutdown_sender.send(GatewayRequest::Shutdown()).await;
+                })
+            }))
+            .await;
+
+        // Handle messages from webserver and plugin
+        while let Some(msg) = self.receiver.recv().await {
+            tracing::trace!("Gateway received message {:?}", msg);
+
```
elsirion, line 539: One could create a shutdown receiver instead here, which internally does pretty much the same as the manual `on_shutdown` registration, then once could `select!` over either `self.receiver.recv()` or the shutdown receiver receiving something. But this works just as well.

elsirion (approved):

m1sterc001guy:
Yes something is not working, need to debug further

gateway/ln-gateway/src/lib.rs
```diff
@@ -511,112 +511,117 @@ impl Gateway {
...
+            .on_shutdown(Box::new(move || {
+                Box::pin(async move {
+                    // ignore errors: we don't care if already stopped
+                    let _ = shutdown_sender.send(GatewayRequest::Shutdown()).await;
+                })
+            }))
+            .await;
+
+        // Handle messages from webserver and plugin
+        while let Some(msg) = self.receiver.recv().await {
```
dpc, line 537: Seems to me like we should just relaying on the sender getting dropped would be enough, but I guess this way works as well. I'm not sure which way is better.

If a task is just an async handler for another thread, just relaying on channels being disconnected is the cleanest, IMO. And it doesn't matter if the other side is stuck or whatever, as it will not send new messages if it's stuck.

dpc: Tested 2feb9610bf7c7f99a6a1a67a86ecc537339da3b1 multiple times when debugging the CI pipeline and had no issues.

m1sterc001guy: I tried doing that and was running into issues causing the tests to timeout. What ended up working is implementing `on_shutdown` and sending a signal to shutdown both the event loop and the webserver.

elsirion (approved):

PR #2196, opened 2023-04-11
fix: enable cross-compilation to android from macos

dpc:
Previously we were blocked on on lack of android SDK, but justin pointed me at:

https://github.com/tadfisher/android-nixpkgs

which should do the job.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -23,16 +23,17 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1679862000,
-        "narHash": "sha256-OCxvI6rxfHS4FpfXu6YDfO9FTFiz4E6iDk6pGvhnvas=",
+        "lastModified": 1679862022,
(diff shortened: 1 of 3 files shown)


justinmoon:
I was successfully able to run the following from my m1 mac from the `nix develop --ignore-environment .#cross` shell:

```
cargo check -p fedimint-client-legacy --target aarch64-linux-android
cargo check -p fedimint-client-legacy --target x86_64-linux-android
cargo check -p fedimint-client-legacy --target armv7-linux-androideabi
```

🎉

justinmoon (approved):

elsirion (approved):
Works for me too

justinmoon:
> Works for me too

To be clear, this shouldn't change anything for you unless you're compiling **from** MacOS **targeting** Android

dpc:
@justinmoon We switch to new toolchain for Linux users as well, so there is a chance something would broke on Linux.

Issue #2200, opened 2023-04-12
Get rid of openssl


dpc:
The only thing that prevents us from cross-compiling the whole codebase (instead of just client packages) is openssl (which is harder to get for cross-compilation targets).

Seems like:

https://github.com/bitcoindevkit/rust-esplora-client/pull/44

might fix it. I'm not sure if there's any other dependency that pulls it in.

justinmoon:
I feel like we should almost add a CI check to make sure that openssl isn't in our `Cargo.lock`. It keeps sneaking back in.

oleonardolima:
Unfortunately `clippy` does not have a lint for something like "disallowed dependencies" (https://rust-lang.github.io/rust-clippy/master/index.html#cargo), as it would be simpler with the `.clippy.toml`. I guess it might be a good feature to propose there.

elsirion:
We don't need special tooling for this, `grep`ing the `Cargo.lock` file should be sufficient to make sure we don't depend on openssl.

dpc:
Note: we already have a place that does something kind of line that: https://github.com/fedimint/fedimint/blob/21ffb6deb9a8aa7cb3710674e086cd066b75b6a0/misc/git-hooks/pre-commit#L38

PR #2201, opened 2023-04-12
chore: explicitily import reqwest with rustls support

dpc:
Re #2200


elsirion (approved):
This doesn't really help with getting rid of OpenSSL (`default_features=false` should do that already), but ensures that crates compile on their own and not just because some other crate in the workspace coincidentally requested the rustls feature.

justinmoon (approved):

PR #2202, opened 2023-04-12
fix: Let IsolatedDatabaseTransaction::raw_find_by_prefix stream all the way

douglaz:
`IsolatedDatabaseTransaction::raw_find_by_prefix` will get a `Stream`, transform it into a `Vec`, then transform into a `Stream` again destroying all the good things about streams in the process.
(Note the `.collect::<Vec<_>>()` followed by a `stream::iter(stream)` on the old code)

Why do this? Probably because if you just remove the `collect` then the borrow checker gets very nasty.

Yes, the borrow checker. Our friend, but sometimes our foe.

This solution works around that by decomposing the method into two functions, bypassing the struct ownership.

Probably there is a better solution if you play with the lifetimes. But there are times that our life time is too short.
(But everyone is invited to give it a try)

This was found during the review of https://github.com/fedimint/fedimint/pull/2183 but it's a existing and tricky enough issue to have its own PR.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 use anyhow::{Context, Result};
 use fedimint_core::util::BoxFuture;
 use fedimint_logging::LOG_DB;
-use futures::{stream, Stream, StreamExt};
+use futures::{Stream, StreamExt};
 use macro_rules_attribute::apply;
 use serde::Serialize;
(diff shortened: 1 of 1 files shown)


m1sterc001guy (approved):

douglaz:
Rebased to trigger CI

elsirion (dismissed):
Great solution!

douglaz:
Changed to references

elsirion (approved):

PR #2211, opened 2023-04-13
fix: tests hanging on unterminated jsonrpc connections

dpc:


dpc:
Landed everything elsewhere.

PR #2215, opened 2023-04-14
Implement route hint fetching for LND

justinmoon:
Closes https://github.com/fedimint/fedimint/issues/1953

### DIFF
--- a/gateway/ln-gateway/src/lnd.rs
+++ b/gateway/ln-gateway/src/lnd.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::fmt;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/lnd.rs
```diff
@@ -263,10 +264,71 @@ impl ILnRpcClient for GatewayLndClient {
...
+                .into_inner();
+
+            let policy = match info.node1_policy.clone() {
+                Some(policy) => policy,
+                None => continue,
+            };
+            let src_node_id = PublicKey::from_str(&chan.remote_pubkey)
+                .unwrap()
+                .serialize()
+                .to_vec();
```
justinmoon, line 309: Is there a better way to do this?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -263,10 +264,71 @@ impl ILnRpcClient for GatewayLndClient {
...
+                    ))
+                })?
+                .into_inner();
+
+            let policy = match info.node1_policy.clone() {
+                Some(policy) => policy,
+                None => continue,
+            };
+            let src_node_id = PublicKey::from_str(&chan.remote_pubkey)
+                .unwrap()
```
elsirion, line 307: nit: Should be an expect stating why it's expected to work.

gateway/ln-gateway/src/lnd.rs
elsirion: The way you do it *might* be the right one because it validates that we got a valid pub key. If you don't want that it looks like simply parsing a hex string?

elsirion (approved):
Looks good, LN is in a lot of flux rn anyway, so happy to go ahead. Would be good to test this functionality.

PR #2221, opened 2023-04-14
fix: remove rustls from wasm

Maan2003:


dpc (approved):

elsirion (dismissed):

elsirion:
Needs some conditional imports apparently :(

Screenshot 2023-04-15 at 10-51-22 fix remove rustls from wasm · fedimint_fedimint@9107920


Eventually we might just want to further split core or move tls to server to get rid of that feature flag. Actually, did you check if it might be easier to move tls to server?

fedimint-core/src/encoding/mod.rs
```diff
@@ -5,6 +5,8 @@
 mod btc;
 mod secp256k1;
 mod tbs;
+
+#[cfg(not(target_wasm = "wasm"))]
```
Maan2003, line 9: :smile: `target_wasm = "wasm"`

fedimint-core/src/lib.rs
```diff
@@ -22,6 +22,7 @@ pub use crate::core::server;
 use crate::encoding::{Decodable, DecodeError, Encodable};
 use crate::module::registry::ModuleDecoderRegistry;
 
+#[cfg(not(target_family = "wasm"))]
```
elsirion, line 25: Eventually I assume we'll also want the admin client to be WASM compatible? But I'm ok with getting it ready in a separate PR if it's too big.

elsirion (approved):

Maan2003:
> did you check if it might be easier to move tls to server?

It is not, because we can only implement the Encodable trait for foreign types in fedimint-core

PR #2223, opened 2023-04-14
WIP doc: mermaid diagrams for lightning state machines

m1sterc001guy:
Adding mermaid diagrams for LN module state machines and gateway state machines.

### DIFF
--- a/modules/fedimint-gw-client/Cargo.toml
+++ b/modules/fedimint-gw-client/Cargo.toml
@@ -6,3 +6,6 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+anyhow = "1.0.66"
+aquamarine = "0.3.0"
+fedimint-core ={ path = "../../fedimint-core" }
--- a/modules/fedimint-gw-client/Cargo.toml
+++ b/modules/fedimint-gw-client/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "fedimint-gw-client"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
--- a/modules/fedimint-gw-client/src/lib.rs
+++ b/modules/fedimint-gw-client/src/lib.rs
@@ -1,1 +1,2 @@
-// TODO: Implement GW client
+mod pay;
+mod receive;
--- a/modules/fedimint-gw-client/src/lib.rs
+++ b/modules/fedimint-gw-client/src/lib.rs
@@ -1,0 +1,1 @@
+// TODO: Implement GW client
--- a/modules/fedimint-gw-client/src/pay.rs
+++ b/modules/fedimint-gw-client/src/pay.rs
@@ -1,0 +1,17 @@
+/// State machine that pays a lightning invoice on behalf of a federation user.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    PayInvoice -- validate outgoing contract --> Canceled
+///    PayInvoice -- await preimage offer accepted --> BuyPreimage
+///    PayInvoice -- await preimage offer failure --> Canceled
+///    BuyPreimage -- await federation decryption --> Preimage
+///    BuyPreimage -- external payment --> Preimage
+///    BuyPreimage -- await buy preimage failure --> Refund
+///    Preimage -- await claim outgoing contract --> Outpoint
+///    Preimage -- await claim outgoing contract failure --> Refund
+///    Refund -- await cancel transaction acceptance --> Refunded
+///    Refund -- await cancel transaction failure --> Failure
+/// ```
--- a/modules/fedimint-gw-client/src/pay.rs
+++ b/modules/fedimint-gw-client/src/pay.rs
@@ -1,19 +1,24 @@
 use fedimint_core::encoding::{Decodable, Encodable};
 
+#[cfg_attr(doc, aquamarine::aquamarine)]
 /// State machine that pays a lightning invoice on behalf of a federation user.
 ///
 /// ```mermaid
 /// graph LR
 /// classDef virtual fill:#fff,stroke-dasharray: 5 5
 ///
-///    PayInvoice -- validate outgoing contract --> Canceled
-///    PayInvoice -- await preimage offer accepted --> BuyPreimage
-///    PayInvoice -- await preimage offer failure --> Canceled
-///    BuyPreimage -- await federation decryption --> Preimage
-///    BuyPreimage -- external payment --> Preimage
-///    BuyPreimage -- await buy preimage failure --> Refund
-///    Preimage -- await claim outgoing contract --> Outpoint
-///    Preimage -- await claim outgoing contract failure --> Refund
+///    FetchContract -- fetch/validate contract failed --> Canceled
+///    FetchContract -- fetch and validate contract success --> BuyPreimage
+///    BuyPreimage -- internal --> InternalPayment
+///    BuyPreimage -- external --> ExternalPayment
+///    InternalPayment -- await incoming contract acceptance -->Funded
+///    InternalPayment -- await incoming contract failure --> Canceled
+///    Funded -- await federation decryption --> Preimage
+///    Funded -- await federation decryption failure --> Refund
+///    ExternalPayment -- lightning payment success --> Preimage
+///    ExternalPayment -- lightning payment failure --> Canceled
+///    Preimage -- await claim transaction acceptance --> Claimed
+///    Preimage -- await claim transaction failure --> Failure
 ///    Refund -- await cancel transaction acceptance --> Refunded
 ///    Refund -- await cancel transaction failure --> Failure
 /// ```
--- a/modules/fedimint-gw-client/src/pay.rs
+++ b/modules/fedimint-gw-client/src/pay.rs
@@ -1,3 +1,5 @@
+use fedimint_core::encoding::{Decodable, Encodable};
+
 /// State machine that pays a lightning invoice on behalf of a federation user.
 ///
 /// ```mermaid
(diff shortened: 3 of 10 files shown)


modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
+/// State machine that pays a lightning invoice on behalf of a federation user.
```
elsirion, line 1: Module docs begin with `//! `, `/// ` only works for docs attached to things defined right below (and module comments kinda relate to the whole file instead).

But if you do that the `#[aquamarine]` attribute macro doesn't work. So either you could move the comment to `lib.rs` as a normal `///` doc comment above the `mod` decalaration or you attach them to some struct defined inside the module.

modules/fedimint-gw-client/src/lib.rs
```diff
@@ -0,0 +1 @@
+// TODO: Implement GW client
```
elsirion, line 1: You need to add `(pub) mod …` entries for the files with the diagrams so they are rendered

modules/fedimint-gw-client/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "fedimint-gw-client"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
elsirion, line 8: add `aquamarine = "0.3.0"` dep

modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
...
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    PayInvoice -- validate outgoing contract --> Canceled
```
elsirion, line 7: Do we assume we already have the contract? Otherwise we might prepend a step that first fetches it? That would likely be modeled as one step with validation in the code, but conceptually seems like 2.

modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
...
+///    PayInvoice -- await preimage offer accepted --> BuyPreimage
+///    PayInvoice -- await preimage offer failure --> Canceled
```
elsirion, line 9: What are the two different failure cases here?

modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
+/// State machine that pays a lightning invoice on behalf of a federation user.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    PayInvoice -- validate outgoing contract --> Canceled
+///    PayInvoice -- await preimage offer accepted --> BuyPreimage
```
elsirion, line 8: Not sure what offers do here already? This transition seems more about having validated the outgoing contract and now proceeding to buy the preimage in the manner applicable.

modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
...
+///    PayInvoice -- await preimage offer failure --> Canceled
+///    BuyPreimage -- await federation decryption --> Preimage
+///    BuyPreimage -- external payment --> Preimage
```
elsirion, line 11: This might deserve a step in-between to commit to which branch we are taking (internal/external) to prevent that after a restart the list of available GWs changed and we'd choose differently the second time around (potentially trying to buy the same preimage twice in two different ways).

modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
...
+///    BuyPreimage -- await buy preimage failure --> Refund
+///    Preimage -- await claim outgoing contract --> Outpoint
+///    Preimage -- await claim outgoing contract failure --> Refund
```
elsirion, line 14: How do we get a refund in this case? Seems more like a hard error condition.

modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
+/// State machine that pays a lightning invoice on behalf of a federation user.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    PayInvoice -- validate outgoing contract --> Canceled
+///    PayInvoice -- await preimage offer accepted --> BuyPreimage
+///    PayInvoice -- await preimage offer failure --> Canceled
+///    BuyPreimage -- await federation decryption --> Preimage
```
elsirion, line 10: This does a bit more than just waiting, it actually has to create another transaction funding an incoming contract and then wait for this to trigger a preimage decryption.

modules/fedimint-gw-client/src/pay.rs
```diff
@@ -0,0 +1,17 @@
...
+///    BuyPreimage -- external payment --> Preimage
+///    BuyPreimage -- await buy preimage failure --> Refund
+///    Preimage -- await claim outgoing contract --> Outpoint
```
elsirion, line 13: this is also not just "await", it triggers the transaction claiming the contract and then waits for its acceptance. Might deserve an intermediate state to properly model the failure condition of the tx being rejected for some reason.

elsirion (commented):
Can you make the diagrams render with `cargo doc --document-private-items` please, that will make review easier :)

I left comments for one, but am getting rather tired, will take a look at the rest tomorrow.

modules/fedimint-gw-client/src/receive.rs
```diff
@@ -0,0 +1,15 @@
+/// State machine that handles an intercepted HTLC from the Lightning node.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    HTLCIntercepted -- await preimage offer accepted --> BuyPreimage
+///    HTLCIntercepted -- await preimage offer failed --> CancelHTLC
```
elsirion, line 8: This failure means "offer not found"?

modules/fedimint-gw-client/src/receive.rs
```diff
@@ -0,0 +1,15 @@
...
+///    BuyPreimage -- await preimage decryption --> Preimage
+///    BuyPreimage -- await preimage decryption failure --> CancelHTLC
```
elsirion, line 10: This also needs to generate a refund tx

modules/fedimint-gw-client/src/receive.rs
```diff
@@ -0,0 +1,15 @@
...
+///    Preimage -- settle HTLC --> Settled
+///    Preimage -- settle HTLC failure --> Preimage
```
elsirion, line 12: I wonder if we should give up eventually? At least there should be a delay so we don't starve other state machines.

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Yes I'm making that assumption here. I can split that into two states

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Yeah I've been kind of lumping the "transaction + additional action" into one transition. I'll split into two

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Removed

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Removed

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Agreed, this should be a failure state

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Added some dummy structs so the diagrams appear

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Made into two states

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: I split this into two states. First transition awaits for the incoming contract funding tx to be accepted, which transitions into the Funded state. Then we await preimage decryption after we're in the Funded state.

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Done

modules/fedimint-gw-client/src/receive.rs
m1sterc001guy: Added a new state/transition to make this more clear

modules/fedimint-gw-client/src/receive.rs
m1sterc001guy: Added a timeout transition that moves it to the Refund state.

modules/fedimint-gw-client/src/pay.rs
m1sterc001guy: Hmm actually maybe I didn't understand this earlier. So this is actually "fetch outgoing contract and submit transaction claiming contract"?

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,21 @@
...
+///
+///     CreatedOutgoingLnContract -- await transaction timeout --> Aborted
+///     CreatedOutgoingLnContract -- await transaction acceptance --> Funded    
+///     Funded -- await gateway pay  --> Success
+///     Funded -- tell gateway about contract --> Funded
+///     Funded -- timeout --> Refund
+///     Funded -- gateway issued refunded --> Refund
+///     Refund -- await transaction acceptance --> Refunded
+///     Refund -- await transaction rejected --> Failure
+/// ```
```
m1sterc001guy, line 19: This diagram needs to be updated to contain the case where the client directly buys the preimage from the federation.

PR #2225, opened 2023-04-14
feat: `timing` module for simple performance monitoring and tuning

dpc:


fedimint-core/src/timing.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.level(Level::INFO)
+    }
+
+    pub fn cancel(&mut self) {
+        self.inner.take();
+    }
+
+    pub fn finish(self) {
+        /* drop */
+    }
```
justinmoon, line 135: WHat's the purpose of this???

dpc: Just a shorthand.

SO you can

```rust
timing.finish();
```

instead of `drop(timing);`.

The reporting logic happens on drop.

justinmoon (approved):

PR #2227, opened 2023-04-14
chore: report more timings in major tests

dpc:
On top of #2225

### DIFF
--- a/fedimint-core/src/timing.rs
+++ b/fedimint-core/src/timing.rs
@@ -1,0 +1,144 @@
+use std::borrow::Cow;
+use std::time;
+
+use fedimint_logging::LOG_TIMING;
+use tracing::{debug, info, trace, warn, Level};
+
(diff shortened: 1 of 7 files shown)


fedimint-core/src/timing.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub fn info(self) -> Self {
+        self.level(Level::INFO)
+    }
+
+    pub fn cancel(&mut self) {
+        self.inner.take();
+    }
+
+    pub fn finish(self) {
+        /* drop */
```
justinmoon, line 134: This confused me. `/* drop self because it is taken by move */` would have confused me less.

justinmoon (approved):

fedimint-core/src/timing.rs
Maan2003: just do `drop(self)` when in confusion

PR #2228, opened 2023-04-14
chore: minor db fixes

dpc:


justinmoon (approved):
Seems like an improvement to not have to give up ownership when calling these functions.

Issue #2236, opened 2023-04-15
`FM_PID_FILE` is extremely brittle


dpc:
I'm debugging [consensus upgrade test hanging in CI](https://github.com/fedimint/fedimint/actions/runs/4705699872/jobs/8346429730) and it looks like it comes down to `tail --pid=$pid /dev/null` hanging indefinitely.

I started looking at the code and I'm not 100% sure what's wrong, but the way we're using `FM_PID_FILE` is extremely brittle. Multiple processes are appending text to it, which I don't think is guaranteed to be atomic (outputs from multiple processes could possibly arrived mixed, which can especially mess up with lines of numbers), but at minimum ordering there is not guaranteed in the presence of multiple processes writing to it when running in parallel (`command &`) and having multiple tasks access it as well. 

@jkitman @Maan2003

justinmoon:
It seems like the only things that still need that are `upgrade-test.sh` and `setup-test.sh`. Once they're re-written in rust, i think we can completely remove FM_PID_FILE.

justinmoon:
`$FM_PID_FILE` was removed by https://github.com/fedimint/fedimint/pull/3007

PR #2237, opened 2023-04-15
fix: bandaid for flaky cli_test_upgrade hanging sometimes

dpc:
See https://github.com/fedimint/fedimint/issues/2236

My guess is that the reason the test is failing is that 

https://github.com/fedimint/fedimint/blob/fc7b43c650c4685936d0b19faca070e898372f8b/scripts/upgrade-test.sh#L9

sometimes is able to record pids from `FM_PID_FILE` before
it was updated.


justinmoon (approved):
Worth a try

PR #2246, opened 2023-04-18
Configurable Gateway Fees

okjodom:
Enable configurable gateway fees for routing outgoing payments.
Fees are declared using known [RoutingFees](https://github.com/lightningdevkit/rust-lightning/blob/d7f6e34b73fe92173549a32fd608cdf1a9bf5c38/lightning/src/routing/gossip.rs#L1058-L1066) structure from lightning crate, and can be configured when setting up the gateway.

Example: `export FM_GATEWAY_FEES= "10,2000"` means the gateway will run declare a `10 msat` base fee and a `2%` liquidity based fee to federations served

By default, gatewayd configures a **base fee of 0 msat** and a **marginal fee of 1%**, which is [on par with current behavior](https://github.com/fedimint/fedimint/blob/d1e696d0a5bfae3e2cc23f42eba8739213aefa20/client/client-lib/src/ln/mod.rs#L52).

Next, we need to [introduce a gateway API for configuring routing fees](https://github.com/fedimint/fedimint/issues/2271)

Longer term, we need more comprehensive definition for gateway fees, as tracked by Discussion #2392


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1035,6 +1035,7 @@ mod fedimint_migration_tests {
                 .expect("Could not parse URL to generate GatewayClientConfig API endpoint"),
             route_hints: vec![],
             valid_until: SystemTime::now(),
+            fees_proportional_millionths: 0,
```
m1sterc001guy, line 1038: Why is this 0?

fedimint-client-legacy/src/ln/mod.rs
```diff
@@ -550,7 +553,8 @@ mod tests {
         assert_eq!(contract_acc.contract.gateway_key, gateway.mint_pub_key);
         // TODO: test that the client has its key
 
-        let expected_amount_msat = invoice_amt_msat + (invoice_amt_msat / 100);
+        let expected_amount_msat = invoice_amt_msat
+            + (invoice_amt_msat * (gateway.fees_proportional_millionths / 1000000) as u64);
```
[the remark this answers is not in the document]
okjodom: > Can we use DEFAULT_FEES_PROPORTIONAL_MILLIONTHS constant here?

This value is actually injected (announced) by API call to register the gateway with federations. Client code should not have knowledge of the constant value.

> Ah I see its not the same value. Let's make a new constant then

if you mean we should make the computed value of `expected_amount_msat`, I don't see the sense since clients should expect the result of the computation to be dynamic. Updated with every fresh selection / update of the gateway to use in contracts

modules/fedimint-ln-server/src/lib.rs
okjodom: this test suite doesn't assert on the actual values used. Rather it only asserts on the fields of the struct. Using `0` as a random placeholder value on the new field `fees_proportional_millionths` should be harmless, and a good verification of the test assertions imo

m1sterc001guy: Ah I missed that this was the migration test. Yeah that's fine.

fedimint-client-legacy/src/ln/mod.rs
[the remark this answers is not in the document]
okjodom: Oh. That makes sense. No, it's a const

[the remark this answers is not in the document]
okjodom: See `compute_outgoing_contract_amount`

m1sterc001guy (approved):

fedimint-client-legacy/src/lib.rs
```diff
@@ -133,6 +133,9 @@ pub struct GatewayClientConfig {
     /// `short_channel_id` when creating invoices to be settled by this
     /// gateway.
     pub mint_channel_id: u64,
+    /// Liquidity-based routing fee in millionths of a routed amount.
+    /// In other words, 10000 is 1%.
+    pub fees_proportional_millionths: u32,
```
dpc, line 138: If you expect other fee structure systems, might be worthwhile to turn this into an enum, especially if gets written to database.

```rust
#[non_exhaustive]
pub enum FeeStructure {
  Proportional { per_million : u32 }
}
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -58,6 +58,10 @@ const ROUTE_HINT_RETRY_SLEEP: Duration = Duration::from_secs(2);
 /// LND HTLC interceptor can't handle SCID of 0, so start from 1
 const INITIAL_SCID: u64 = 1;
 
+/// Liquidity-based routing fee in millionths of a routed amount.
+/// The default is 10000 (1%). In other words, 10000 is 1%.
+pub const DEFAULT_FEES_PROPORTIONAL_MILLIONTHS: u32 = 10000;
```
dpc, line 63: Could be a `Default` on `FeeStructure`

fedimint-client-legacy/src/lib.rs
okjodom: Ack. I'm expecting this standard definition of [RoutingFees](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/routing/gossip.rs#L1057-L1065).

okjodom: > Ack. I'm expecting this standard definition of [RoutingFees](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/routing/gossip.rs#L1057-L1065).

@m1sterc001guy, @dpc,
I converted PR to draft while I attempt change with full RoutingFees definition

gateway/ln-gateway/src/lib.rs
okjodom: Reused extern definition of [RoutingFees](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/routing/gossip.rs#L1057-L1065) so we actually need a default const

fedimint-core/src/encoding/mod.rs
```diff
@@ -379,6 +379,29 @@ impl Encodable for lightning_invoice::Invoice {
     }
 }
 
+impl Encodable for lightning::routing::gossip::RoutingFees {
```
dpc, line 382: Not sure if it's a good idea to depend for our own encodings on data-structures from external crates (that possibly could change). 

Also how we are going to implement gateway fees and LN gateway fees, and probably two different things. 

We might have fees paid in different currencies supported by federation, on maybe "no gateway fee tuesdays". That's why I think we should go with an enum. But the main variant of that enum should probably be this thing. Not sure if we should copy&paste or use directly.

gateway/ln-gateway/Cargo.toml
```diff
@@ -43,6 +43,7 @@ fedimint-core = { path = "../../fedimint-core" }
 fedimint-logging = { path = "../../fedimint-logging" }
 fedimint-rocksdb = { path = "../../fedimint-rocksdb" }
 futures = "0.3.24"
+lightning = "0.0.113"
```
dpc, line 46: I would copy&paste this type, just to avoid another dependency.

fedimint-core/src/encoding/mod.rs
okjodom: > RE: Also how we are going to implement gateway fees and LN gateway fees, are probably two different things.
>
> Could have fees paid in different currencies supported by federation, or maybe "no gateway fee tuesdays", or some other things? If not, then whatever, let's just go with this, seems reasonable. If possibly yes, maybe we should go with an enum. But the main variant of that enum should probably be this thing. Not sure if we should copy&paste or use directly.

Fantastic idea IMO. If I adopt it as a proposal, we'd have
```rs
#[non_exhaustive]
pub enum FeeStructure {
  LnRouting { fees: RoutingFees }
}
```
To open up future possibility of:
```rs
#[non_exhaustive]
pub enum FeeStructure {
  LnRouting { fees: RoutingFees },
  Discount { ... }
}
```
Because "Free Gateway Tuesdays" is a fantastic thing :)

okjodom: > RE: Not sure if it's a good idea to depend for our own encodings on data-structures from external crates (that possibly could change).

We are also [guilty in these other accounts,](https://github.com/fedimint/fedimint/blob/master/fedimint-core/src/encoding/mod.rs#L376-L391) and I'm not keen on maintaining a variant of [RoutingFees](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/routing/gossip.rs#L1057-L1065) because we already have a dependency on the crate and type in other scenarios.

Can we fully rely on version control to alert breaks in encodings?  Is maintaining our own types the only way to avoid encoding extern types?

okjodom: Proposing flexible fee structure as a quick follow up to the current PR: #2287

gateway/ln-gateway/Cargo.toml
okjodom: fixed in #2287

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -266,6 +271,58 @@ pub mod route_hints {
...
+// lightning::routing::gossip::RoutingFees
+// See https://github.com/lightningdevkit/rust-lightning/blob/b8ed4d2608e32128dd5a1dee92911638a4301138/lightning/src/routing/gossip.rs#L1057-L1065
+pub mod serde_routing_fees {
+    use anyhow::anyhow;
+    use lightning::routing::gossip::RoutingFees;
+    use serde::ser::SerializeStruct;
+    use serde::{Deserialize, Deserializer, Serializer};
+
+    fn u64_to_u32(x: u64) -> Result<u32, anyhow::Error> {
+        if x > u32::MAX as u64 {
```
douglaz, line 284: why not
```rust
Ok(x.try_into()?)
```

fedimint-core/src/encoding/mod.rs
justinmoon: One way I could also see this working in the future is that clients present the gateway(s) with an invoice _before_ they create the outgoing contract essentially asking them what the fees would be to pay this invoice. Default behavior could be just computing a fee based fixed `base_msat` / `proportional_millionths`. But more sophisticated gateways could probe and guess which channel they would pay across in case they had different channels with different routing fees. So fees would be a function of the invoice being paid.

justinmoon:
Thinking about this a little more, I think we need to achieve the following:

* Clients are presented a fee that will be charged for the payment before creating outgoing / incoming contract . They can use this fee to choose a gateway, display this fee to the user, and calculate the amount needed for the outgoing / incoming contracts.
* For outgoing payments, gateways can set a fee they will charge for routing payments a given payment, _net of routing fees they pay_. For incoming payments, gateways can set a separate fee to receive a payment for the user compensating them for locking up capital in channel capacity and ecash balance.

So the first conclusion is that I think we need separate fees for incoming and outgoing payments. They fundamentally represent costs of different scarce resources: outgoing lightning liquidity, and incoming lightning liquidity + cost and risk of holding ecash.

Another conclusion is that no matter how we express these fees in the outgoing case, we need to take them into consideration when the gateway pays the actual invoice. We’ll need to pass these fee calculations on to [PayInvoiceRequest.max_fee_percent](https://github.com/fedimint/fedimint/blob/75b47e3928dbb3446802d68bff0ed43efb9c903f/gateway/ln-gateway/proto/gatewaylnrpc.proto#L40-L46) field in our gRPC spec. Without doing this, the gateway only controls the fees they charge the end-user, but could still lose money routing the payment if routing fees exceed fees charged to the user.

I’m also not sure if `lightning::routing::gossip::RoutingFees` is the best way to encode our fees, because it seems like it’s intended to express the cost for forwarding a payment across a single channel. But the gateway’s job is different: completing the entire payment no matter how many hops it takes. And if we do use `RoutingFees` to express gateway fees, we'll need another parameter expressing how much of these fees the gateway will reserve for itself when paying the invoice. For example, if `RoutingFees` translates to a fee of `x` for a given invoice, the gateway might want to use a value less than `x` for `PayInvoiceRequest.max_fee_percent` to guarantee that they make a profit.

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -168,6 +169,9 @@ pub struct LightningGateway {
     pub route_hints: Vec<route_hints::RouteHint>,
     /// Limits the validity of the announcement to allow updates
     pub valid_until: SystemTime,
+    // Gateway configured routing fees
```
garyray-k, line 172: Missed a `///`?

okjodom: ack.

modules/fedimint-ln-common/src/lib.rs
okjodom: nice

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -47,6 +49,11 @@ pub struct GatewayOpts {
     /// Gateway webserver authentication password
     #[arg(long = "password", env = "FM_GATEWAY_PASSWORD")]
     pub password: String,
+
+    /// Configured gateway routing fees
+    /// Format: <base_msat>,<proportional_millionths>
```
m1sterc001guy, line 54: Nit: For me I know I would be confused by this format, would probably just prefer separate parameters

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -493,16 +493,23 @@ impl LightningClientModule {
...
-            Amount::from_msats(contract_amount_msat)
+        // Compute amount to lock in the outgoing contract
+        let invoice_amount_msat = invoice
+            .amount_milli_satoshis()
+            .ok_or(anyhow::anyhow!("MissingInvoiceAmount"))?;
+
+        let fees = gateway.fees;
+        let base_fee = fees.base_msat as u64;
+        let margin_fee: u64 = if fees.proportional_millionths > 0 {
+            let fee_percent = 1000000 / fees.proportional_millionths as u64;
```
m1sterc001guy, line 504: What does 1000000 represent? Would prefer constant

okjodom: huh. it's just the number 1 million used to compute percentage here :)

okjodom:
> LGTM, let's iterate on it

thanks for review.

@justinmoon, take a look if you still need this change asap

PR #2248, opened 2023-04-18
client-ng: Don't load all e-cash notes

elsirion:
Ports #2183 to client-ng.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -723,6 +723,28 @@ impl<'isolated, T: MaybeSend + Encodable> ModuleDatabaseTransaction<'isolated, T
             })
     }
 
+    #[instrument(level = "debug", skip_all, fields(key = ?key_prefix))]
+    pub async fn find_by_prefix_sorted_descending<KP>(
+        &mut self,
+        key_prefix: &KP,
+    ) -> impl Stream<
+        Item = (
+            KP::Record,
+            <<KP as DatabaseLookup>::Record as DatabaseRecord>::Value,
+        ),
+    > + '_
+    where
+        KP: DatabaseLookup,
+        KP::Record: DatabaseKey,
+    {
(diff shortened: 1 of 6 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -723,6 +723,28 @@ impl<'isolated, T: MaybeSend + Encodable> ModuleDatabaseTransaction<'isolated, T
             })
     }
 
+    #[instrument(level = "debug", skip_all, fields(key = ?key_prefix))]
+    pub async fn find_by_prefix_sorted_descending<KP>(
```
jkitman, line 727: Guessing keys are stored in order so this is constant time?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -377,29 +392,88 @@ impl MintClientModule {
...
+    // not sufficient to cover the pending amount
+    // The tuple is (amount, note, checkpoint), where checkpoint is the index where
+    // the note should be inserted on the selected vector if it is needed
+    let mut last_big_note_checkpoint: Option<(Amount, Note, usize)> = None;
+    let mut pending_amount = requested_amount;
+    let mut previous_amount: Option<Amount> = None; // used to assert descending order
+    loop {
+        if let Some((note_amount, note)) = stream.next().await {
+            assert!(
+                previous_amount.map_or(true, |previous| previous >= note_amount),
```
jkitman, line 419: Wonder how much this actually saves in practice...I'd suspect in most cases you'd need to load every note down to 1sat in order to make exact change.

fedimint-core/src/db/mod.rs
elsirion: Kind of, afaik all our DB backends have some sort of BTree index, so lookups should be `O(log(n))` and ordered traversal probably `O(1)`? @m1sterc001guy

PR #2250, opened 2023-04-18
refactor: out new client interfaces for testing

jkitman:
In order to test the next gen client, we need to extract interfaces that both the legacy and new client can implement in the interim.  These interfaces can eventually become part of the `client-app` API that is directly used by mobile wallet apps.

The final client would implement `IWalletClient + IMintClient + ILightningClient`.

`ITestClient` contains functions specifically for testing.  `IGatewayClient` contains functions only used by the gateway client.

### DIFF
--- a/integrationtests/tests/fixtures/legacy.rs
+++ b/integrationtests/tests/fixtures/legacy.rs
@@ -1,0 +1,434 @@
+use std::error::Error as StdError;
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use bitcoin::hashes::sha256::Hash as Sha256Hash;
+use bitcoin::{Address, Transaction as BitcoinTransaction, Txid};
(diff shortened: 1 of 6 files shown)


integrationtests/tests/fixtures/legacy.rs
```diff
@@ -0,0 +1,419 @@
+use std::error::Error as StdError;
```
jkitman, line 1: Looks like a lot of code, but it's mostly just delegation to implement the new interfaces.

integrationtests/tests/fixtures/user.rs
```diff
@@ -0,0 +1,319 @@
+use std::iter::repeat;
```
jkitman, line 1: All the new interfaces are contained in this file.

elsirion (approved):

PR #2251, opened 2023-04-19
client-ng: don't make modules aware of their instance id

elsirion:
Builds on  #2248

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 use anyhow::anyhow;
 use fedimint_core::api::{DynFederationApi, IFederationApi, WsFederationApi};
 use fedimint_core::config::{ClientConfig, ModuleGenRegistry};
-use fedimint_core::core::{ModuleInstanceId, ModuleKind};
+use fedimint_core::core::{DynInput, DynOutput, IInput, IOutput, ModuleInstanceId, ModuleKind};
 use fedimint_core::db::{Database, DatabaseTransaction, IDatabase};
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::module::registry::ModuleDecoderRegistry;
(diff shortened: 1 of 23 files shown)


fedimint-core/src/module/mod.rs
```diff
@@ -956,11 +956,32 @@ impl<T: Encodable + Decodable> From<&T> for SerdeModuleEncoding<T> {
...
+
+    /// In cases where we know exactly which module kind we expect but don't
+    /// have access to all decoders this function can be used instead.
+    ///
+    /// Note that it just assumes the decoded module instance id to be valid
+    /// since it cannot validate against the decoder registry. The lack of
+    /// access to a decoder registry also makes decoding structs impossible that
+    /// themselves contain module dyn-types (e.g. a module output containing a
+    /// fedimint transaction).
+    pub fn try_into_inner_known_module_kind(&self, decoder: &Decoder) -> Result<T, DecodeError> {
```
dpc, line 973: If this happens to be needed due to lack of `instance_id`, we could just wrap the instance id in some `OpaqueInstanceId(InstanceId)` with not way for the module code to access what's in there, but be able to pass to APIs that need it. Just pointing it out.

dpc (approved):

fedimint-client/src/lib.rs
```diff
@@ -131,50 +228,93 @@ impl Debug for ClientInner {
...
+        input: InstancelessDynClientInput,
+    ) -> TransactionId {
+        let instance_input = ClientInput {
+            input: DynInput::from_parts(self.module_instance, input.input),
+            keys: input.keys,
+            state_machines: states_add_instance(self.module_instance, input.state_machines),
+        };
+
+        self.client
+            .finalize_and_submit_transaction(
```
m1sterc001guy, line 259: Is the general guidance that new clients shouldn't call `finalize_and_submit_transaction` directly and instead should call one of these wrappers? (`claim_input`, `fund_output`)

fedimint-client/src/lib.rs
```diff
@@ -131,50 +228,93 @@ impl Debug for ClientInner {
     }
 }
 
+/// Global state given to a specific client module and state. It is aware inside
+/// which module instance and operation it is used and to avoid module being
+/// aware of their instance id etc.
+#[derive(Clone, Debug)]
+struct ModuleGlobalClientContext {
+    client: Arc<ClientInner>,
+    module_instance: ModuleInstanceId,
```
m1sterc001guy, line 237: nit: I'd prefer `module_instance_id` here, I got confused what this was when looking at the code

fedimint-client/src/lib.rs
```diff
@@ -351,6 +507,19 @@ impl ClientInner {
...
     }
+
+    async fn transaction_update_stream(
+        &self,
+        operation_id: OperationId,
+    ) -> BoxStream<OperationState<TxSubmissionStates>> {
+        self.executor
+            .notifier()
+            .module_notifier::<OperationState<TxSubmissionStates>>(
+                TRANSACTION_SUBMISSION_MODULE_INSTANCE,
```
m1sterc001guy, line 518: I haven't come across this yet. What does this represent?

m1sterc001guy (approved):
Looks like a good refactor to me

fedimint-client/src/lib.rs
elsirion: Since the executor only supports state machines in a module context I used `u16::MAX` as the fake modiule instance ID there. You can find the state machine responsible for submitting transactions here:

https://github.com/fedimint/fedimint/blob/33ccea92b85783e293c0903c9f7cab379d2b32f7/fedimint-client/src/transaction/sm.rs#L45-L84

fedimint-client/src/lib.rs
elsirion: Will try to stay consistent in the future/fix in another PR.

fedimint-client/src/lib.rs
elsirion: When constructing transactions using module client-generated inputs and outcputs there is only support for `finalize_and_submit_transaction` (that's implemented on `Client`) right now, inside state transitions only the `claim_input` and `fund_output` functions can be used, the `finalize_and_submit_transaction` isn't exposed anymore.

This will become clearer once I add the last commit to #2270

PR #2253, opened 2023-04-19
fix: remove flood of Replacing existing connection in test output

### LINKED ISSUE #2107: Flood of `Replacing existing connection` in test output
```
00:04:46 2023-04-04T16:23:12.683342Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.687260Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.687815Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.691400Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.691413Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.695305Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.695932Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.699354Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.699507Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.703042Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.703574Z  INFO run_io_thread: net::peer: Replacing existing connection                                                                                   
00:04:46 2023-04-04T16:23:12.706870Z  INFO run_io_thread: net::peer: Replacing existing connection        
```

Something somehwere could use a small delay.

jkitman:
Fixes #2107

This was annoying me when debugging as well.  Confirmed the culprit, don't take much of a performance hit for adding normal reconnect delays.

```
time ./scripts/rust-tests.sh > out-delayed.txt 2>&1 

real	5m26.697s
user	24m35.680s
sys	2m34.754s

grep -c "Replacing existing connection" out-delayed.txt
2

...

time ./scripts/rust-tests.sh > out-normal.txt 2>&1 

real	5m6.703s
user	26m43.863s
sys	2m46.803s

grep -c "Replacing existing connection" out-normal.txt
72491
```

### DIFF
--- a/integrationtests/tests/fixtures/mod.rs
+++ b/integrationtests/tests/fixtures/mod.rs
@@ -50,7 +50,7 @@ use fedimint_server::consensus::{
 };
 use fedimint_server::net::connect::mock::{MockNetwork, StreamReliability};
 use fedimint_server::net::connect::{parse_host_port, Connector, TlsTcpConnector};
-use fedimint_server::net::peers::{DelayCalculator, PeerConnector};
+use fedimint_server::net::peers::PeerConnector;
 use fedimint_server::{consensus, FedimintServer};
 use fedimint_testing::btc::bitcoind::FakeWalletGen;
 use fedimint_testing::btc::fixtures::FakeBitcoinTest;
@@ -559,7 +559,7 @@ async fn distributed_config(
             let cfg = ServerConfig::distributed_gen(
                 &our_params,
                 registry.legacy_init_modules(),
-                DelayCalculator::TEST_DEFAULT,
+                Default::default(),
                 &mut task_group,
             );
             (*peer, cfg.await.expect("generation failed"))
@@ -1264,7 +1264,7 @@ impl FederationTest {
                 db.clone(),
                 module_inits.clone(),
                 connect_gen(cfg),
-                DelayCalculator::TEST_DEFAULT,
+                Default::default(),
                 &mut task_group,
             )
             .await


integrationtests/tests/fixtures/mod.rs
```diff
@@ -559,7 +559,7 @@ async fn distributed_config(
             let cfg = ServerConfig::distributed_gen(
                 &our_params,
                 registry.legacy_init_modules(),
-                DelayCalculator::TEST_DEFAULT,
```
dpc, line 562: OK, this is so subtle yet notable difference, that I don't think we should leave it as is.

I mean - `TEST_DEFAULT` was supposed to be ... the default... for the tests... and this is a test file... :D

IMO:

* Maybe we should get rid of the `Default::default()` and `DelayCalculator::new()` on this type; too easy to use them carelessly. I don't think `Default` is needed, and we can have `DelayCalculator::production()` or something?
* Maybe make this exponential or give it a threshold when to stop being so aggressive? It's great that it tries fast initially, but the floods of reconnection logs when something doesn't work well is a problem. @douglaz , any thoughts?

elsirion: Exponential backoff would be great! But this PR already seems to be an improvement.

elsirion (approved):

integrationtests/tests/fixtures/mod.rs
douglaz: The problem with the "production" delay is due to its exponential nature it may lead to very high "tail latencies". By very high I don't mean 5% more in general, but like 2x-3x more 1-5% of time. At least this was true when I tested. This may have changed. Perhaps a better way would to make the "test" delay a fixed (like 1s)

douglaz: > Good find!
> 
> OK, this is so subtle yet notable difference, that I don't think we should leave it as is.
> 
> I mean - `TEST_DEFAULT` was supposed to be ... the default... for the tests... and this is a test file... :D
> 
> IMO:
> 
> * Maybe we should get rid of the `Default::default()` and `DelayCalculator::new()` on this type; too easy to use them carelessly. I don't think `Default` is needed, and we can have `DelayCalculator::production()` or something?

But yes, I this is behavior is okay I would just increase the scaling factor for this non production parameter profile instead of switching to the production one. Perhaps the `default` implementation is really misleading here.

dpc: > The problem with the "production" delay is that due to its exponential nature it may lead to very high "tail latencies".

I don't understand this one.

Usually what I'd implement is exponential + ceiling (like 10/30/60s). So it does 1ms,2ms,4,8 and eventually just constant 10s.

dpc:
Let's land this as is. Created https://github.com/fedimint/fedimint/issues/2263 to do some further fixes.

integrationtests/tests/fixtures/mod.rs
dpc: https://github.com/fedimint/fedimint/issues/2263

jkitman: Yep always have a ceiling for exponential backoffs.

elsirion: >  The problem with the "production" delay is that due to its exponential nature it may lead to very high "tail latencies". By very high I don't mean 5% more in general, but like 2x-3x more 1-5% of time. At least this was true when I tested. This may have changed. Perhaps a better way would to make the "test" delay a fixed one (like 1s)

That sounds kinda ok, the tests would take a bit longer but still succeed?

jkitman: Tests were about 20s slower with this, but we should cap the timeout just in case, don't think it's capped now.

PR #2254, opened 2023-04-19
client-ng: add operation log

elsirion:
Builds on  #2251

This PR adds a log of all operation IDs that also allows the creator to add meta data. This meta data can be used by module clients to indicate the high level intent behind the operation helping with interpreting the associated state machines.


elsirion:
Closing in favor of #2270 which builds on top, so I don't have to rebase two PRs all the time.

PR #2265, opened 2023-04-20
feat: expose `versions` api endpoints

dpc:
For client to be able to fetch the summary of all the versions.


justinmoon:
Just ran `fedimint-cli api version` and got a nice little summery 👍

Issue #2263, opened 2023-04-19
Rethink connection retry timeouts


dpc:
See https://github.com/fedimint/fedimint/pull/2253#discussion_r1171833588

douglaz:
> > The problem with the "production" delay is that due to its exponential nature it may lead to very high "tail latencies".
> 
> I don't understand this one.
> 
> Usually what I'd implement is exponential + ceiling (like 10/30/60s). So it does 1ms,2ms,4,8 and eventually just constant 10s.

My mistake here. It isn't currently exponential (which would make things only worse), but it still keeps increasing and there is a high ceiling, so eventually the test will take much longer than expected.

The ceiling is defined by https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/net/peers.rs#L37 and enforced by https://github.com/fedimint/fedimint/blob/master/fedimint-server/src/net/peers.rs#L435

The maximum delay for a single reconnection on default/production settings right know is:
`4 * 300 seconds = 20 minutes`

Under the `TEST_DEFAULT` profile this was at most:
`4 * 300 * 0.1e-3 seconds = 120ms`

So it would keep a safe limit even under stress scenarios.

Perhaps we don't want <1s reconnection delays because this may trigger some warnings, but for a test you also don't want anything longer than that because a test is more of a "speedrun" than a "normal game".

I'll create a PR with one possible solution.

dpc:
> The maximum delay for a single reconnection on default/production settings right know is:

20 minutes sounds like very much, even for production. Even 1s from peers is not going to cause any strain, 60s would be a reasonable value that would allow quick re-connection after host comes back online.

> Perhaps we don't want <1s reconnection delays because this may trigger some warnings, but for a test you also don't want anything longer than that because a test is more of a "speedrun" than a "normal game".

Ideally, sure. But if we have a bug and for some reason the test can't reconnect, then we get a flood of messages that are making debugging experience rather bad.

I think it would be better if the reconnection was initially fast, but after a decent enough time, slowed down significantly. Using exponents in around 1.1-1.6 might give such a result.

BTW. Would be great to see reconnection counters if we're not printing them already. At least it helps correlated these messages.

douglaz:
> > The maximum delay for a single reconnection on default/production settings right know is:
> 
> 20 minutes sounds like very much, even for production. Even 1s from peers is not going to cause any strain, 60s would be a reasonable value that would allow quick re-connection after host comes back online.
> 
[...]

PR #2261, opened 2023-04-19
 chore: fix code coverage status symbols 

dpc:
https://docs.codecov.com/docs/commit-status#target

Fix https://github.com/fedimint/fedimint/issues/2105

PR #2267, opened 2023-04-20
refactor: limit test reconnection delays

douglaz:
Introduce again limits on test reconnection delay but without introducing the flood solved by https://github.com/fedimint/fedimint/pull/2261

Further context is on https://github.com/fedimint/fedimint/issues/2263

(still WIP)

### DIFF
--- a/fedimint-server/src/net/peers.rs
+++ b/fedimint-server/src/net/peers.rs
@@ -33,9 +33,6 @@ use crate::net::connect::{AnyConnector, SharedAnyConnector};
 use crate::net::framed::AnyFramedTransport;
 use crate::net::queue::{MessageId, MessageQueue, UniqueMessage};
 
-/// Maximum connection failures we consider for our back-off strategy
-const MAX_FAIL_RECONNECT_COUNTER: u64 = 300;
-
 /// Owned [`Connector`](crate::net::connect::Connector) trait object used by
 /// [`ReconnectPeerConnections`]
 pub type PeerConnector<M> = AnyConnector<PeerMessage<M>>;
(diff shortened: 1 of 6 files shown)


fedimint-server/src/net/peers.rs
```diff
@@ -85,36 +82,37 @@ struct PeerConnectionStateMachine<M> {
...
 
 impl DelayCalculator {
-    fn new() -> Self {
-        Self {
-            scaling_factor: 1.0,
-        }
-    }
+    /// Maximum connection failures we consider for our back-off strategy,
+    /// production default
+    const PROD_MAX_FAIL_RECONNECT_COUNTER: u64 = 300;
```
jkitman, line 91: Should this be a `Duration` bound instead, so it's more clear?

If someone changes the `scaling_factor` then `300` might end up being too high.

dpc:
```
> for i in $(seq 1 20) ; echo "$i: $(math "(4 ^ $i) / 1000" )ms" ; end
1: 0.004ms
2: 0.016ms
3: 0.064ms
4: 0.256ms
5: 1.024ms
6: 4.096ms
7: 16.384ms
8: 65.536ms
9: 262.144ms
10: 1048.576ms
11: 4194.304ms
12: 16777.216ms
13: 67108.864ms
14: 268435.456ms
15: 1073741.824ms
16: 4294967.296ms
17: 17179869.184ms
18: 68719476.736ms
19: 274877906.944ms
20: 1099511627.776ms
```

I think this is something that we could just have both in tests and in prod at the same time (with a cap of 10s maybe).

5 very quick ones, another 5 quite fast as well, but then quickly straight to the ceiling.

douglaz: I've created https://github.com/fedimint/fedimint/pull/2285 using these ideas

PR #2269, opened 2023-04-20
chore: increase linux cross compile timeout

elsirion:
15 minutes is not enough from time to time. Example: https://github.com/fedimint/fedimint/actions/runs/4752263039/jobs/8442402557?pr=2251

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -165,7 +165,7 @@ jobs:
           - on: linux
             runs-on: ubuntu-22.04
             build-in-pr: true
-            timeout: 15
+            timeout: 20
           - host: macos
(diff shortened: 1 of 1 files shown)


dpc:
Should we maybe switch these to use buildjet? Part of the reason they're slow. I'm not sure how much it will cost us though. @elsirion

elsirion:
Oh sure, I didn't even notice! Originally we only left the non-time-critical things on GH runners (i.e. the daily/master builds).

One problem we could run into is the maximum parallel core usage limit which is 64 and would be quite expensive to increase unfortunately. We should at least make sure that one PR doesn't require more than 64 buildjet cores, maybe some jobs would also perform ok on 4 core machines?

PR #2270, opened 2023-04-20
client-ng: Mint extension trait

elsirion:
The main change introduced by this PR is the introduction of the `MintClientExt` trait:

https://github.com/fedimint/fedimint/blob/dedf431e2463d644bed2a384194ce7d1606b9c15/modules/fedimint-mint-client/src/lib.rs#L59-L105

It is implemented for the `Client` struct and if in scope assumes that the client contains at least one mint module and that we always want to use the first one. This is the case most of the time and frees the API consumer from worrying about instances.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -241,7 +241,7 @@ impl Debug for ClientInner {
 #[derive(Clone, Debug)]
 struct ModuleGlobalClientContext {
     client: Arc<ClientInner>,
-    module_instance: ModuleInstanceId,
+    module_instance_id: ModuleInstanceId,
     operation: OperationId,
 }
 
@@ -257,9 +257,9 @@ impl IGlobalClientContext for ModuleGlobalClientContext {
         input: InstancelessDynClientInput,
     ) -> TransactionId {
         let instance_input = ClientInput {
-            input: DynInput::from_parts(self.module_instance, input.input),
+            input: DynInput::from_parts(self.module_instance_id, input.input),
             keys: input.keys,
-            state_machines: states_add_instance(self.module_instance, input.state_machines),
+            state_machines: states_add_instance(self.module_instance_id, input.state_machines),
         };
 
         self.client
@@ -278,8 +278,8 @@ impl IGlobalClientContext for ModuleGlobalClientContext {
         output: InstancelessDynClientOutput,
     ) -> anyhow::Result<TransactionId> {
         let instance_output = ClientOutput {
-            output: DynOutput::from_parts(self.module_instance, output.output),
-            state_machines: states_add_instance(self.module_instance, output.state_machines),
+            output: DynOutput::from_parts(self.module_instance_id, output.output),
+            state_machines: states_add_instance(self.module_instance_id, output.state_machines),
         };
 
         self.client
@@ -519,7 +519,7 @@ impl ClientInner {
         Arc::new(move |module_instance, operation| {
             ModuleGlobalClientContext {
                 client: client_inner.clone(),
-                module_instance,
+                module_instance_id: module_instance,
                 operation,
             }
             .into()
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -10,7 +10,7 @@ use anyhow::{anyhow, bail};
 use fedimint_core::api::{DynFederationApi, IFederationApi, WsFederationApi};
 use fedimint_core::config::{ClientConfig, ModuleGenRegistry};
 use fedimint_core::core::{DynInput, DynOutput, IInput, IOutput, ModuleInstanceId, ModuleKind};
-use fedimint_core::db::{Database, DatabaseTransaction, IDatabase};
+use fedimint_core::db::{AutocommitError, Database, DatabaseTransaction, IDatabase};
 use fedimint_core::encoding::{Decodable, DecodeError, Encodable};
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::task::{MaybeSend, MaybeSync, TaskGroup};
@@ -199,7 +199,7 @@ fn states_to_instanceless_dyn<
 >(
     state_gen: StateGenerator<S>,
 ) -> StateGenerator<Box<maybe_add_send_sync!(dyn IState<DynGlobalClientContext> + 'static)>> {
-    Box::new(move |txid, out_idx| {
+    Arc::new(move |txid, out_idx| {
         let states: Vec<S> = state_gen(txid, out_idx);
         states
             .into_iter()
@@ -304,7 +304,7 @@ fn states_add_instance(
         Box<maybe_add_send_sync!(dyn IState<DynGlobalClientContext> + 'static)>,
     >,
 ) -> StateGenerator<DynState<DynGlobalClientContext>> {
-    Box::new(move |txid, out_idx| {
+    Arc::new(move |txid, out_idx| {
         let states = state_gen(txid, out_idx);
         Iterator::collect(
             states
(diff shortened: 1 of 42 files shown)


fedimint-client/src/lib.rs
```diff
@@ -317,20 +323,88 @@ pub type ModuleGlobalContextGen = ContextGen<DynGlobalClientContext>;
...
+        if ClientInner::operation_exists(&mut dbtx, operation_id).await {
+            bail!("There already exists an operation with id {operation_id:?}")
+        }
+
         let txid = self
             .inner
             .finalize_and_submit_transaction(&mut dbtx, operation_id, tx_builder)
             .await?;
+
+        dbtx.insert_entry(
```
dpc, line 403: Use `insert_new_entry` if you don't want to overwrite it. Not sure if it's worth keeping `operation_exists` given that. Might be more human readable, but then it's an extra codepath...

fedimint-client/src/lib.rs
```diff
@@ -317,20 +323,88 @@ pub type ModuleGlobalContextGen = ContextGen<DynGlobalClientContext>;
...
+            &OperationLogKey { operation_id },
+            &OperationLogEntry {
+                operation_type: operation_type.to_string(),
+                meta: serde_json::to_value(operation_meta(txid))?,
+            },
+        )
+        .await;
+        dbtx.insert_entry(
```
dpc, line 412: `_new` as well I guess.

fedimint-client/src/lib.rs
```diff
@@ -317,20 +323,88 @@ pub type ModuleGlobalContextGen = ContextGen<DynGlobalClientContext>;
...
+            },
+            &(),
+        )
+        .await;
+
         dbtx.commit_tx().await;
         Ok(txid)
     }
 
+    // TODO: allow fetching pages
```
dpc, line 422: Note: don't use pages, use cursors. I thought it might be worth mentioning...

BTW. instead of pages&cursor ... just return the iterator directly to the caller maybe (with all the `.map` etc needed)?

fedimint-client/src/lib.rs
```diff
@@ -743,3 +844,36 @@ pub fn client_decoders<'a>(
...
+}
+
+impl Decodable for OperationLogEntry {
+    fn consensus_decode<R: Read>(
+        r: &mut R,
+        modules: &ModuleDecoderRegistry,
+    ) -> Result<Self, DecodeError> {
+        let operation_type = String::consensus_decode(r, modules)?;
+        let meta_str = String::consensus_decode(r, modules)?;
+        let meta = serde_json::from_str(&meta_str).map_err(DecodeError::from_err)?;
```
dpc, line 987: Sneaking in `SerdeEncodable` here like that, just after you removed it? :P

fedimint-client/src/lib.rs
```diff
@@ -426,14 +426,40 @@ impl Client {
...
+            .find(|(_, kind, _module)| *kind == module_kind)
+            .map(|(instance_id, _, _)| instance_id)
     }
 }
 
 struct ClientInner {
     db: Database,
     primary_module: DynPrimaryClientModule,
     primary_module_instance: ModuleInstanceId,
+    primary_module_kind: ModuleKind,
```
dpc, line 509: Could be method? :)

fedimint-core/src/module/registry.rs
```diff
@@ -2,46 +2,50 @@ use std::collections::BTreeMap;
...
 
-use crate::core::Decoder;
 pub use crate::core::ModuleInstanceId;
+use crate::core::{Decoder, ModuleKind};
 use crate::server::DynServerModule;
 
 /// Module Registry hold module-specific data `M` by the `ModuleInstanceId`
 #[derive(Debug, Clone)]
-pub struct ModuleRegistry<M>(BTreeMap<ModuleInstanceId, M>);
+pub struct ModuleRegistry<M>(BTreeMap<ModuleInstanceId, (ModuleKind, M)>);
```
dpc, line 11: Do we really need it? Seems a bit unnecessary. For some types, you can just ask the `M` what's the `ModuleKind`, for some others it shouldn't matter?

dpc: I don't know what's the exact reason for it. Is it for `ModuleDecoderRegistry`? We could add `fn kind(&self) -> ModuleKind` to `Decoder` instead?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -45,6 +49,138 @@ use crate::output::{
...
+            MintMeta::Reissuance { out_point } => out_point,
+        };
+
+        let (_, mint_client) = mint_client(self);
+
+        let tx_accepted_future = self.await_tx_accepted(operation_id, out_point.txid);
+        let output_finalized_future = mint_client.await_output_finalized(operation_id, out_point);
+
+        Ok(Box::pin(stream! {
+            yield ReissueExternalNotesState::Created;
```
dpc, line 200: WTH?! :D

dpc (approved):

fedimint-client/src/lib.rs
elsirion: Good catch! The previous check has to stay in place though since all the `new_` version does is *warn* in the logs if there was a replacement happened.

It might be worth making it an assertion, back in the day when `insert_new`/`delete_existing` were introduced it was really hard to know at times if there was no case in which a key would exist since there weren't any database transactions yet and merely these DB change batches that would be executed at a later point. By now we might be reasonably sure what we expect and could even return `Option<V>` from `insert` and then check at the call site if the returned element is what we expected.

fedimint-client/src/lib.rs
elsirion: I tried, but for that to be possible we need read-only database transactions such that we can read from the prefix stream while also querying the same dbtx for operation details. Currently that's impossible because both the stream and the read fns take mut refs. That's also necessary for a lot of use cases because database transactions aren't sync so non-mut refs are not send. So maybe it wouldn't even work with a read-only database transaction?

fedimint-core/src/module/registry.rs
elsirion: The reason was that I wanted to find modules by kind, we could also just extend the traits  to give us the module kind, just seemed more involved.

modules/fedimint-mint-client/src/lib.rs
elsirion: `yield` via macro ^^ I actually saw it first in an @okjodom PR

fedimint-core/src/module/registry.rs
elsirion: Opened https://github.com/fedimint/fedimint/issues/2319 to track

fedimint-client/src/lib.rs
elsirion: Would be made obsolete by https://github.com/fedimint/fedimint/issues/2319 too

fedimint-client/src/lib.rs
dpc: Created #2321

fedimint-client/src/lib.rs
dpc: @elsirion  I was thinking that we could return a tuple of database transaction and the iterator. I think it should work, where the lifetime of the iterator would be marked as shorter than of the dbtx.

fedimint-client/src/lib.rs
dpc: Logging json as a free form metadata is not unusual, but I think it's funny anyway. :D

fedimint-core/src/module/registry.rs
dpc: > The reason was that I wanted to find modules by kind,

Also a method, using `filter` :)

PR #2272, opened 2023-04-20
Status API endpoint and improved switching between APIs

### LINKED ISSUE #2262: DB Notification system should detect shutdown and fail the call
It seems that jsonrpsee will not shutdown cleanly until all client connections are handled. But it also seems that the client connection handlers don't terminate on shutdown system.

It would be ideal if they did.

### LINKED ISSUE #2264: CI test flake: "test config::api::tests::test_config_api has been running for over 60 seconds"
https://github.com/fedimint/fedimint/actions/runs/4748309711/jobs/8434356580

jkitman:
Adds the `server_status` to the APIs so that we can check whether we've restarted from an upgrade and call `start_consensus`.  Removes the need for passing in an upgrade arg and the `upgrade.sh` test.

Also found a solution that fixes #2264 (figured out how to force the server to stop via the `TaskGroup` handler).  May also fix #2262?


fedimint-server/src/config/api.rs
```diff
@@ -630,6 +656,16 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConfigGenApi>> {
                 config.start_consensus(auth).await
             }
         },
+        api_endpoint! {
+            "server_status",
```
dpc, line 660: Just `status` would be enough, IMO.

fedimint-server/src/lib.rs
```diff
@@ -193,20 +203,40 @@ impl FedimintServer {
...
-            .await
-            .context(format!("Bind address: {}", self.settings.api_bind))
-            .expect("Could not start API server")
-            .start(module)
-            .expect("Could not start API server");
-
-        let handler = FedimintApiHandler { runtime, handle };
+        // TODO: Remove this flag needed only for tests
+        let handler = if force_shutdown {
+            let runtime = Runtime::new().expect("Creates runtime");
```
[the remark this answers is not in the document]
dpc: If it doesn't work in tests, how is it going to work in prod? :(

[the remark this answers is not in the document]
jkitman: Well it works in the test I have against the `test_config_api` test, but there's something about running it in the rust-tests.sh that is causing issues.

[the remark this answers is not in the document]
dpc: I'm not a fan of that workaround, and would rather rely on a proper shutdown detection, but if it unblocks you, let's land. Maybe add some `// Workaround:` comment, link to the discussion and issues` and in the future, once the original root cause is fixed, we might revert that part.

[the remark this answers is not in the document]
dpc: Also - maybe let's fill a `jsonrpcsee` issue to supported forced shutdown? It should be too difficult to implement at this level, and it seems very useful to have.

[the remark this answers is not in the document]
jkitman: Okay just pushed a much better solution without a flag.  Instead I return the handle and the caller deals with it.  I think the callback in the `TaskGroup` must be called by one of the many things in the integration test framework before we are ready to shutdown the API.

fedimint-server/src/config/api.rs
```diff
@@ -765,6 +803,23 @@ mod tests {
                 })
                 .await
         }
+
+        /// Helper for getting server status
+        async fn server_status(&self) -> ServerStatus {
+            while self.client.status().await.is_err() {
```
dpc, line 809: This is reading it twice, with a possibility of second one panicking.

```rust
loop {
match self.client.status().await {
  Ok(o) => return Ok(o);
  Err(e) => /* print, sleep */
}
```

?

fedimint-server/src/lib.rs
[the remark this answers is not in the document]
dpc: I see. So instead of trying to guarantee that server is gone when returning, we just return a handle, and now it's a caller's problem to wait for things to finish?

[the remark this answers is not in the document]
dpc: Yeah, I like that better.

fedimint-server/src/config/api.rs
```diff
@@ -765,6 +803,23 @@ mod tests {
...
+                    target: fedimint_logging::LOG_TEST,
+                    "Test retrying server status"
+                )
+            }
+            self.client.status().await.unwrap()
+        }
+
+        /// Helper for writing the password file to bypass the API
+        fn write_password_file(&self) {
+            fs::write(self.dir.join(PLAINTEXT_PASSWORD), &self.auth.0).unwrap();
```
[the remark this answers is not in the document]
jkitman: I see, fs:write could partially write?

fedimint-server/src/config/api.rs
jkitman: True, main reason the polling is needed is because we are waiting for the API to come online so it shouldn't fail the second time, but I can rewrite it better.

fedimint-server/src/config/api.rs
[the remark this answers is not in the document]
dpc: Yeap. And without `fsync` there's no guarantee it was written to disk, and so so and so. https://lwn.net/Articles/457667/

jkitman:
@dpc think I addressed all your feedback now?

fedimint-server/src/config/api.rs
```diff
@@ -317,17 +318,42 @@ impl ConfigGenApi {
...
+        tx.commit_tx().await;
+
         self.config_generated_tx.send(cfg).await.expect("Can send");
 
         Ok(())
     }
 
+    /// Returns the server status
+    pub async fn server_status(&self) -> ServerStatus {
+        let has_upgrade_flag = { self.has_upgrade_flag().await };
```
elsirion, line 340: The braces seem a little bit unnecessary here?

fedimint-server/src/config/api.rs
```diff
@@ -317,17 +318,42 @@ impl ConfigGenApi {
...
+            _ if has_upgrade_flag => ServerStatus::Upgrading,
+            _ => ServerStatus::GeneratingConfig,
+        }
+    }
+
+    /// Returns true if the upgrade flag is set indicating that the server was
+    /// shutdown due to a planned upgrade
+    pub async fn has_upgrade_flag(&self) -> bool {
+        let mut tx = self.db.begin_transaction().await;
+        tx.get_value(&ConsensusUpgradeKey).await.is_some()
```
elsirion, line 354: nit: `tx` is an ambiguous name, I'd prefer `dbtx`

elsirion (approved):
Some nits, feel free to address or merge.

scripts/rust-tests.sh
```diff
@@ -20,7 +20,7 @@ export FM_TEST_USE_REAL_DAEMONS=1
 
 if [ -z "${FM_TEST_ONLY:-}" ] || [ "${FM_TEST_ONLY:-}" = "bitcoind" ]; then
   >&2 echo "### Testing against bitcoind"
-  env RUST_BACKTRACE=1 cargo test -p fedimint-tests -- --test-threads=$(($(nproc) * 2)) "$@"
+  env RUST_BACKTRACE=1 cargo test -p fedimint-tests -- --test-threads=16 "$@"
```
dpc, line 23: Any specific reason for this?

jkitman: Ah was failing with 64 threads on my machine, didn't mean to commit the change...maybe we can automatically set the ulimit when running `nix develop`?

dpc: Oh, I reverted back to `2 * nproc` BTW. I didn't know it breaks for you, and in CI it is a tiny bit faster...

Is it because not enough memory to run all these tests at once, or something else (ulimit?)?

Yeah, I think we should make it an env variable `${FM_TEST_THREADS:-$(($(nproc) * 2))}` or something.

jkitman: Just the ulimit, I forgot to set it.  CI machines run with 16 normally, but locally I can run with more.

fedimint-server/src/config/api.rs
jkitman: I'm used to doing that to keep the async away from the locks, but I think it is unnecessary...

PR #2279, opened 2023-04-20
feat: Lightning Client NG Pay Invoice

m1sterc001guy:
Lightning Client NG paying a bolt 11 invoice. Adds the `Pay` state machine.


fedimint-cli/src/ng.rs
```diff
@@ -82,6 +97,83 @@ pub async fn handle_ng_command<D: IDatabase>(
...
+                .get_module_client::<MintClientModule>(LEGACY_HARDCODED_INSTANCE_ID_MINT)
+                .unwrap();
+
+            let client_input = mint_client
+                .create_input(
+                    &mut dbtx.with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_MINT),
+                    operation_id,
+                    amount,
+                )
+                .await?;
```
elsirion, line 137: You actually don't need this, `finalize_and_submit_transaction` will add funds as needed.

https://github.com/fedimint/fedimint/blob/fccd7396a5f294e61dc503b5446911823d74e720/fedimint-cli/src/ng.rs#L66-L72

fedimint-cli/src/ng.rs
```diff
@@ -82,6 +97,83 @@ pub async fn handle_ng_command<D: IDatabase>(
...
+                .unwrap();
+            let mut dbtx = client.db().begin_transaction().await;
+            let outgoing_ln_contract = ln_client
+                .create_outgoing_ln_contract(
+                    &mut dbtx.with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_LN),
+                    client.context(),
+                    invoice,
+                    rng,
+                )
+                .await?;
```
elsirion, line 114: Ideally you'd return a `ClientOutput` from this function, much like in the mint module:

https://github.com/fedimint/fedimint/blob/d9cb6c2d44f0f9e659a27e9502735c6fa2e00376/modules/fedimint-mint-client/src/lib.rs#L173-L182

You'd return the strongly typed variant (see above example) and then convert it to the dynamic one using `into_dyn`

https://github.com/fedimint/fedimint/blob/d9cb6c2d44f0f9e659a27e9502735c6fa2e00376/fedimint-client/src/transaction/builder.rs#L39-L52

when you use it in `ng.rs`, just like done here with the `mint_input`:

https://github.com/fedimint/fedimint/blob/d9cb6c2d44f0f9e659a27e9502735c6fa2e00376/fedimint-cli/src/ng.rs#L61-L67

fedimint-cli/src/ng.rs
```diff
@@ -82,6 +97,83 @@ pub async fn handle_ng_command<D: IDatabase>(
...
+                        },
+                        state: LightningPayStates::CreatedOutgoingLnContract(
+                            LightningPayCreatedOutgoingLnContract {
+                                funding_txid,
+                                contract_id,
+                            },
+                        ),
+                    },
+                )]
+            });
```
elsirion, line 154: This should be part of `create_outgoing_ln_contract`, the `ClientOutpput` struct already contains a state machine generator.

elsirion (commented):
Great progress! Just a few high-level comments about the API, will go through the actual state machines more thoroughly tomorrow :)

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,545 @@
...
+                    common: old_state.common,
+                    state: LightningPayStates::Refund(LightningPayRefund { refund_txid: txid }),
+                },
+                Err(_) => LightningPayStateMachine {
+                    common: old_state.common,
+                    state: LightningPayStates::Failure,
+                },
+            }
+        } else {
+            // TODO: Add some sort of delay here so we dont constantly ask for block height
```
m1sterc001guy, line 334: This isn't correct. Right now we just get stuck in the `Refundable` state if anything goes wrong. Maybe this should transition to a separate state for the case where we need to wait for the timeout. Right now its confusing why the payment gets stuck.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -25,65 +57,405 @@ impl ClientModuleGen for LightningClientGen {
...
+                    amount: Amount::ZERO,
+                    fee: Amount::ZERO,
+                }
+            }
+        }
+    }
+}
+
+impl LightningClientModule {
+    pub async fn fetch_active_gateway(
```
m1sterc001guy, line 189: Not thrilled that I needed to replicate these functions to expose them to `LightningClientModule`. Perhaps there's a better spot to put these

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -25,65 +57,405 @@ impl ClientModuleGen for LightningClientGen {
...
+        Ok(LightningClientModule {
+            module_instance_id: instance_id,
+            cfg,
+            notifier,
+        })
+    }
+}
+#[derive(Debug, Clone)]
+pub struct LightningClientContext {
+    module_instance_id: ModuleInstanceId,
```
m1sterc001guy, line 75: Need to rebase when https://github.com/fedimint/fedimint/pull/2251 is merged

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,545 @@
...
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///  CreatedOutgoingLnContract -- await transaction failed --> Aborted
+///  CreatedOutgoingLnContract -- await transaction acceptance --> Funded    
+///  Funded -- await gateway payment success  --> Success
+///  Funded -- gateway not available --> Funded
+///  Funded -- await gateway payment failed --> Refundable
+///  Refundable -- gateway issued refunded --> Refund
```
m1sterc001guy, line 30: Add new state to wait for gateway to become cancellable

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -25,65 +57,405 @@ impl ClientModuleGen for LightningClientGen {
...
+    secp: Secp256k1<All>,
+}
+
+impl Context for LightningClientContext {}
+
+impl LightningClientContext {
+    pub async fn fetch_active_gateway(
```
m1sterc001guy, line 82: Refactor this out of the ClientContext

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Refactored out of `LightningClientContext`

gateway/ln-gateway/src/lib.rs
```diff
@@ -382,6 +382,7 @@ impl Gateway {
     }
 
     async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<()> {
+        info!("JUMOELL pay_invoice: {:?}", payload);
```
m1sterc001guy, line 385: Remove

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,536 @@
...
+            .await
+            {
+                Ok(txid) => LightningPayStateMachine {
+                    common: old_state.common,
+                    state: LightningPayStates::Refund(LightningPayRefund { refund_txid: txid }),
+                },
+                Err(_) => LightningPayStateMachine {
+                    common: old_state.common,
+                    state: LightningPayStates::Failure,
+                },
```
justinmoon, line 320: I think we'll need to retry here if, for example, the refund failed because the federation is offline

justinmoon:
The CLI commands all worked for me 🎉

modules/fedimint-ln-client/src/pay.rs
elsirion: An offline federation should not make the transaction submission state machine fail (that would be a bug). We will keep submitting the refund tx until the federation either tells us that it's invalid or that it was accepted.

fedimint-cli/src/ng.rs
m1sterc001guy: This is fixed

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Rebased

modules/fedimint-ln-client/src/db.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl_db_record!(
+    key = OutgoingPaymentKey,
+    value = OutgoingContractData,
+    db_prefix = DbKeyPrefix::OutgoingPayment,
+);
+impl_db_lookup!(
+    key = OutgoingPaymentKey,
+    query_prefix = OutgoingPaymentKeyPrefix
+);
```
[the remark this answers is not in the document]
m1sterc001guy: I don't think so. Will remove.

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: This state is called `Refundable`

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,548 @@
...
+            .map_err(|_| GatewayPayError::GatewayInternalError)?;
+        let length = preimage.len();
+        Ok(preimage[1..length - 1].to_string())
+    }
+
+    async fn transition_pay_timeout(
+        old_state: LightningPayStateMachine,
+        contract_id: ContractId,
+    ) -> LightningPayStateMachine {
+        // TODO: Retry contacting gateway
```
m1sterc001guy, line 269: I didn't implement this yet. What do we want the policy to be here? Try 5 times then go to `Refundable`? Try forever until we contact the gateway?

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,548 @@
...
+        common: &LightningPayCommon,
+        global_context: &DynGlobalClientContext,
+    ) -> Vec<StateTransition<LightningPayStateMachine>> {
+        let txid = self.funding_txid;
+        let contract_id = self.contract_id.clone();
+        let funded_common = common.clone();
+        let rejected_common = funded_common.clone();
+        let success_context = global_context.clone();
+        let reject_context = global_context.clone();
+        let gateway = self.gateway.clone();
```
m1sterc001guy, line 120: For simplicity's sake while implementing this I just cloned everything. I don't think that's too expensive but its probably not necessary.

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: Manually confirmed that `Refundable` will transition after the block timeout or if the gateway sets the `cancellable` flag

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: Yeah the `Failure` case here is when the refund tx is actually rejected by the federation, so this requires the federation to be online.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,545 @@
...
+    async fn transition_outgoing_contract_funded(
+        old_state: LightningPayStateMachine,
+        common: LightningPayCommon,
+        contract_id: ContractId,
+        gateway: LightningGateway,
+    ) -> LightningPayStateMachine {
+        assert!(matches!(
+            old_state.state,
+            LightningPayStates::CreatedOutgoingLnContract(_)
+        ));
```
elsirion, line 165: If you were using the data from inside `old_state.state` you wouldn't need to pass in the gateway, potentially saving some clones.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -0,0 +1,545 @@
...
+        }
+    }
+
+    async fn await_funding_transaction_rejected(
+        common: LightningPayCommon,
+        global_context: DynGlobalClientContext,
+        txid: TransactionId,
+    ) {
+        global_context
+            .await_tx_rejected(common.operation_id, txid)
```
elsirion, line 181: I think we should refactor the await accepted/rejected API in the global context the same way I did for the client itself (1 fn returning a result):

https://github.com/fedimint/fedimint/blob/094f32c2ee2c675a04b1b53c49a6bd87fb38e51c/fedimint-client/src/lib.rs#L707-L722

That would allow merging the two state transitions, requiring fewer clones of notifiers etc.

elsirion (approved):
LGTM, some nits but it works and we can improve incrementally

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -102,6 +103,14 @@ pub trait MintClientExt {
...
         operation_id: OperationId,
     ) -> anyhow::Result<BoxStream<SpendOOBState>>;
+
+    // Wait for the refund transaction to complete after a lightning invoice
+    // failed to be paid.
+    async fn await_lightning_refund(
+        &self,
+        operation_id: OperationId,
+        refund_txid: TransactionId,
+    ) -> anyhow::Result<()>;
```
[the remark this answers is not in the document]
m1sterc001guy: So this is just a quirk with the CLI that I pinged you about offline. The CLI just needs to be alive long enough for the output to be finalized so the notes can be retrieved. The `Pay` state machine doesn't actually wait on this, it just waits for the refund tx to be accepted then exits. We can refactor this if needed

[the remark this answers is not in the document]
elsirion: Ah, I see, was missing that context. I'll implement an abstraction that removes the dependency on the mint module, but good to go for now.

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: Implemented

elsirion (approved):

Issue #2281, opened 2023-04-20
Review use of `SerdeEncodable`


dpc:
See if you can just implement `En/Decodable` for all foreign types?

elsirion:
It doesn't seem to be used anywhere anymore and can be removed afaik.

dpc:
@elsirion Do we have it implemented for the crypto types encoded in module configs? I had to use it in a couple of places in #2283 . Not sure if I missed anything, but I think they are just not implemented, so removing `SerdeEncodable` now will block me.

elsirion:
Only saw that later, will check if decodable is easy to impl there.

elsirion:
While writing 60767d2da7d710a16f6c9f65517bc3b8bbcf4c8d I noticed the many `tiered.map(|(k, v)| (k, v.0)).collect()` hacks necessary to work with `SerdeEncodable`, this makes me even more certain that we want to get rid of it.

For now I could directly impl `Encodable`/`Decodable` for all structs, but in the future with external modules that won't always be the case and I think I should just add some options to the derive macros that let you specify custom encode/decode functions, just like `#[serde(with = "…")]`.

dpc:
> For now I could directly impl `Encodable`/`Decodable` for all structs, but in the future with external modules that won't always be the case and I think I should just add some options to the derive macros that let you specify custom encode/decode functions, just like `#[serde(with = "…")]`.

IMO, we should have impl `Encodable`/`Decodable` for every crate we are using and need it. We can use conditional compilation and optional modules to make it feasible. After all `fedimint-core` is not general purpose, but fedimint-specific.

elsirion:
> IMO, we should have impl `Encodable`/`Decodable` for every crate we are using and need it.

Sure, but we can only reasonably maintain this for default modules. For external modules the annotation would be useful.

PR #2283, opened 2023-04-20
refactor: encode configs as Encodable, not serde_json::Value

dpc:
The current way of using `serde_json::Value` is hashing serde json
serialization string which is probably a bad idea anyway, and
had other shortcomings.

For readability, we leave a field with (redundant) json encodings,
so that humans can view them.

The patch quickly spiraled into a huge refactoring.

I think some more tweaking might be needed, but if the general approach
seems acceptable, we should land to avoid too many merge conflicts.

Created #2281 to review usage of `SerdeEncodable` which is probably
avoidable in many places.

### DIFF
--- a/fedimint-core/Cargo.toml
+++ b/fedimint-core/Cargo.toml
@@ -26,6 +26,7 @@ serde = { version = "1.0.149", features = [ "derive" ] }
 serde_json = "1.0.91"
 strum = "0.24"
 strum_macros = "0.24"
+hex = { version = "0.4.3", features = [ "serde"] }
 sha3 = "0.10.5"
 tbs = { path = "../crypto/tbs" }
 tokio = { version = "1.26.0", features = ["sync"] }
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -26,7 +26,10 @@ use threshold_crypto::group::{Curve, Group, GroupEncoding};
 use threshold_crypto::{G1Projective, G2Projective, Signature};
 use url::Url;
 
-use crate::module::{DynCommonModuleGen, DynServerModuleGen, IDynCommonModuleGen};
+use crate::encoding::Decodable;
+use crate::module::{
+    DynCommonModuleGen, DynServerModuleGen, IDynCommonModuleGen, ModuleConsensusVersion,
+};
 use crate::task::{MaybeSend, MaybeSync};
 use crate::{maybe_add_send_sync, PeerId};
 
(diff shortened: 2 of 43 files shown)


dpc:
@jkitman @elsirion You're probably the best people to give feedback on it.

fedimint-server/src/config/mod.rs
```diff
@@ -133,8 +133,10 @@ pub struct ServerConfigConsensus {
     #[serde(with = "serde_tls_cert_map")]
     pub tls_certs: BTreeMap<PeerId, rustls::Certificate>,
     /// All configuration that needs to be the same for modules
+    pub modules: BTreeMap<ModuleInstanceId, ClientModuleConfig>,
```
dpc, line 136: This is the core change. `modules` is now basically a `BTreeMap<ModuleInstanceId, (Kind, ConsensuVersion, Vec<u8>)`.

Everything else is to just make the compiler accept this new reality.

I leave `modules_json` like `modules` were before, to help the puny human operators that are not capable of decoding hex strings on the fly.

dpc:
```
        > memory allocation of 1275517187546 bytes failed
       > error: test failed, to rerun pass `-p fedimint-client-legacy --lib`
       >
```

That doesn't sound good.

fedimint-core/src/config.rs
```diff
@@ -578,40 +556,62 @@ impl ClientModuleConfig {
...
     pub fn cast<T: TypedClientModuleConfig>(&self) -> anyhow::Result<T> {
-        Ok(serde_json::from_value(self.config.clone())?)
+        Ok(T::consensus_decode(
+            &mut &self.config[..],
+            &Default::default(),
+        )?)
     }
 }
 
+// #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Encodable)]
```
jkitman, line 573: Need to remove?

jkitman (commented):
Should we just implement `Encodable` rather than wrap all the keys with `SerdeEncodable`?

Ah I see #2281

fedimint-server/src/config/mod.rs
```diff
@@ -307,11 +305,13 @@ impl ServerConfig {
                 local,
                 private,
                 consensus,
+                consensus_json,
             } = config;
 
             self.local.modules.insert(name, local);
             self.private.modules.insert(name, private);
             self.consensus.modules.insert(name, consensus);
+            self.consensus.modules_json.insert(name, consensus_json);
```
jkitman, line 314: Kind of weird to have this duplicated, would it be possible to custom implement `serde` to produce the json versions?

dpc: Not in two separate keys, I think. And it would be an extra complexity. It might be a bit weird, but is there only for human convenience.

modules/fedimint-ln-common/src/config.rs
```diff
@@ -20,18 +20,18 @@ pub struct LightningConfig {
     pub consensus: LightningConfigConsensus,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Encodable)]
+#[derive(Debug, Clone, Serialize, Deserialize, Encodable, Decodable)]
 pub struct LightningConfigConsensus {
     /// The threshold public keys for encrypting the LN preimage
-    pub threshold_pub_keys: threshold_crypto::PublicKeySet,
+    pub threshold_pub_keys: SerdeEncodable<threshold_crypto::PublicKeySet>,
```
elsirion, line 26: I'm pretty sure that if `Encodable` was possible to impl, then we can also impl `Decodable`

elsirion: I noticed there were a few more places where `SerdeEncodable` was used. I removed all of them and implemented en/decoding for the types directly in 60767d2da7d710a16f6c9f65517bc3b8bbcf4c8d.

elsirion:
I think a lot of the changes in 60767d2da7d710a16f6c9f65517bc3b8bbcf4c8d came from undoing your changes, so if it was split up into a bunch of "impl encoding for X" commits and then the final one using `Encodable` for config encoding it would make for quite a tidy PR.

fedimint-core/Cargo.toml
```diff
@@ -26,6 +26,7 @@ serde = { version = "1.0.149", features = [ "derive" ] }
 serde_json = "1.0.91"
 strum = "0.24"
 strum_macros = "0.24"
+hex = { version = "0.4.3", features = [ "serde"] }
```
dpc, line 29: Can't live without that `#[serde(with = "::hex::serde")]`

elsirion (approved):

PR #2284, opened 2023-04-20
chore: pre-compile rust-tests in CI

dpc:
This avoids rebuilding them 3 or 4 times in parallel, in each test run.

### DIFF
--- a/scripts/rust-tests.sh
+++ b/scripts/rust-tests.sh
@@ -20,7 +20,7 @@ export FM_TEST_USE_REAL_DAEMONS=1
 
 if [ -z "${FM_TEST_ONLY:-}" ] || [ "${FM_TEST_ONLY:-}" = "bitcoind" ]; then
   >&2 echo "### Testing against bitcoind"
-  env RUST_BACKTRACE=1 cargo test -p fedimint-tests -- --test-threads=16 "$@"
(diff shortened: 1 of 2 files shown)


scripts/rust-tests.sh
```diff
@@ -20,7 +20,7 @@ export FM_TEST_USE_REAL_DAEMONS=1
 
 if [ -z "${FM_TEST_ONLY:-}" ] || [ "${FM_TEST_ONLY:-}" = "bitcoind" ]; then
   >&2 echo "### Testing against bitcoind"
-  env RUST_BACKTRACE=1 cargo test -p fedimint-tests -- --test-threads=$(($(nproc) * 2)) "$@"
+  env RUST_BACKTRACE=1 cargo test -p fedimint-tests ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} -- --test-threads=$(($(nproc) * 2)) "$@"
```
justinmoon, line 23: Nice. For anyone who didn't get it, we re-run this test multiple times with different backends. If we run with same cargo flags here as we built in `test-ci-all.sh`, then don't need to rebuild.

justinmoon (approved):
Needs rebase, though

elsirion (approved):

PR #2288, opened 2023-04-21
chore: pre-populate setup ui using env vars

### LINKED ISSUE #1532: Setup UI should accept the default values of fields from the command-line

jp1ac4:
This aims to resolve #1532 by pre-populating the remaining fields in the setup UI using environment variables.

@dpc As discussed, I've followed the same approach as used in #1683. Please let me know if more robust checks of the environment variables are required.

I haven't pre-populated the guardian name as I believe that would require changing the code in https://github.com/fedimint/fedimint/blob/master/fedimint-testing/src/bin/fixtures.rs#L345, which might be out of scope here, but please let me know otherwise.

### DIFF
--- a/fedimintd/src/ui.rs
+++ b/fedimintd/src/ui.rs
@@ -225,6 +225,9 @@ async fn post_guardians(
 struct UrlConnection {
     ro_bitcoin_rpc_type: &'static str,
     ro_bitcoin_rpc_url: String,
+    bitcoin_network: String,
+    federation_name: String,
+    guardians_count: String,
     bind_p2p: String,
     p2p_url: String,
     bind_api: String,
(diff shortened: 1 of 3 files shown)


fedimintd/src/ui.rs
```diff
@@ -253,6 +256,9 @@ async fn params_page(
     UrlConnection {
         ro_bitcoin_rpc_type,
         ro_bitcoin_rpc_url,
+        bitcoin_network: env::var("FM_BITCOIND_NETWORK").unwrap_or("".to_string()),
```
dpc, line 259: bitcoin vs BITCOIND

I think we should just go with "bitcoin" (no d). It's about general network selector, and is not bitcoind specific.

dpc (commented):
LGTM, just one typo(?)

fedimintd/src/ui.rs
jp1ac4: Ah yes, I'll change this.

jp1ac4:
I've now renamed `FM_BITCOIND_NETWORK` -> `FM_BITCOIN_NETWORK` and squashed & rebased commits.

elsirion (approved):

Issue #1136, opened 2022-12-16
Network stack fixes


elsirion:
- [x] Do not block if peer is not reading from socket (#1149)
- [x] Do not buffer infinite number of epochs for disconnected peers since that is a DoS vector and will actually not help currently if too many messages are sent and processed at once (from more epochs than max parallel epochs) (#1149)
- [ ] Figure out when to rejoin if too many peers dropped messages for us

I think to remember there was a third problem we discussed @jkitman, feel free to add.

jkitman:
The above two cases should drop the connection, disconnected peers should attempt to reconnect, possibly rejoining if too many connections were dropped.

elsirion:
I think in the first case we'd merely disconnect while in the second we'd also drop all buffers, forcing the peer to rejoin consensus if too many of its peers do the same.

jkitman:
I meant in the first case we need to disconnect and drop buffers only once our buffer fills up.

Issue #1080, opened 2022-12-08
Only receive messages from expected epochs from network stack


elsirion:
If a node is partitioned from the rest of the federation for a while it will receive the messages from many epochs in one burst. While every peer connection is ordered we might receive messages from epoch `n+100` from one peer while we have only received messages up to epoch `n` from others, so are still at epoch `n` ourselves. This would lead us to drop the messages from the future (an anti-DoS measure by HBBFT), losing them.

To fix this problem we have to be able to specify that we only want to receive messages up to a certain epoch (determined by the maximum amount of parallel epochs set in HBBFT).

jkitman:
If we implement filtering out-of-order received messages we will still need to buffer them in memory as we do in `PeerConnectionMultiplexer`.  If that were the case, then we might as well just increase `max_future_epochs` in HBBFT which will likely be a more memory-efficient way to buffer the messages.

So I think we need to implement the buffering on the sender-side by keeping track of what epoch our peers are on and only sending messages from `n` epochs ahead.  If our send buffer gets full then we clear the buffer and the peer will need to restart their server to run the rejoin logic (or perhaps we can add the logic to skip ahead when a peer detects they have fallen too far behind such as if the internet is down for a while).

jkitman:
Summing up discussion on how to improve the network stack:

- [ ] In I/O thread, wrap the underlying connection in a buffer to make sending async, drop the connection if the buffer fills up
- [ ] Don't dump the entire resend buffer at once, just send when the connection buffer has room, drop connection if resend buffer fills up
- [ ] Detect missed epochs on the receive side, use it to determine when to rejoin

jkitman:
Won't fix, since we have a workaround of rejoining at any point.  Plus when we have a new consensus algorithm we won't have epochs anymore and it seems like a lot of unnecessary work to add this to the network stack.

Issue #229, opened 2022-07-08
Drop `ReconnectPeerConnections` message cache on extended network interruption


elsirion:
Too many message could make us run out of memory, it's a DoS risk.

elsirion:
Only the DKG uses the old cached peer connections anymore and there we don't make liveness guarantees.

PR #2289, opened 2023-04-21
fix: rejoin after any number of epochs pass

jkitman:
We originally only let a peer rejoin once, however due to the network buffers it's possible they successfully create an outcome but drop HBBFT messages from future epochs, getting them permanently stuck.

This PR allows them to rejoin even after they produce an outcome so long as they determine from a threshold of peers that they are behind.

We can probably close some old issues such as #1080 and #1136.  However, we still need to do #229 since it can stall the entire federation if a peer is offline for too long.

Once we switch to AlpehBFT we can drop a lot of this hacky logic since we will no longer have to deal with the concept of epochs or verifying signed epochs using signatures from produced outcomes.

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -88,7 +88,7 @@ pub struct ConsensusServer {
     /// The list of all other peers
     pub other_peers: BTreeSet<PeerId>,
     /// If `Some` then we restarted and look for the epoch to rejoin at
-    pub rejoin_at_epoch: Option<HashMap<u64, HashSet<PeerId>>>,
+    pub rejoin_at_epoch: HashMap<u64, HashSet<PeerId>>,
     /// How many empty epochs peers requested we run
     pub run_empty_epochs: u64,
     /// Tracks the last epoch outcome from consensus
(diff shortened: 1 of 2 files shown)


elsirion (approved):
Thx a lot for the fix, I tested with a slightly different test locally and it works.

fedimint-server/src/consensus/server.rs
```diff
@@ -563,20 +563,18 @@ impl ConsensusServer {
...
-                    epoch + NUM_EPOCHS_REJOIN_AHEAD
-                );
-                self.hbbft.skip_to_epoch(epoch + NUM_EPOCHS_REJOIN_AHEAD);
-                self.request_rejoin(NUM_EPOCHS_REJOIN_AHEAD).await;
-            }
+        let peers = self.rejoin_at_epoch.entry(epoch).or_default();
+        peers.insert(peer);
+        let threshold = self.cfg.local.p2p_endpoints.threshold();
+
+        if peers.len() >= threshold && self.hbbft.epoch() < epoch {
```
elsirion, line 570: What do you think about having a threshold here, i.e. only falling in line when lacking behind by a few epochs? Would that create new issues? Otherwise I fear we'll hit this code path quite often?

jkitman: I thought about that, but since we should be receiving messages in-order I don't think we'd hit it often unless messages are dropped.  Since we only hit it after a threshold of peers are ahead, we should already have produced an outcome at that point.

PR #2291, opened 2023-04-21
fix: Ensure federation can handle bigger on chain deposits

### LINKED ISSUE #2258: Ensure federation can handle bigger on chain deposits.
and amounts in general.

In https://github.com/fedimint/fedimint/issues/1448 we've discovered that we might break when too much funds are deposited at once, as the current code can't handle too many notes due to artificial limits.

Write a test depositing 100BTC, some sends between users and fix any errors.

jkitman:
Fixes #2258

### DIFF
--- a/fedimint-core/src/tiered_multi.rs
+++ b/fedimint-core/src/tiered_multi.rs
@@ -124,9 +124,20 @@ impl<T> TieredMulti<T> {
             .flat_map(|(amt, notes)| notes.into_iter().map(move |c| (amt, c)))
     }
 
-    /// Returns the length of the longest vector of all tiers
-    pub fn longest_tier_len(&self) -> usize {
-        self.0.values().map(|notes| notes.len()).max().unwrap_or(0)
+    /// Returns the length of the longest vector of all tiers, ignoring the
+    /// `except` tier
+    pub fn longest_tier_except(&self, except: &Amount) -> usize {
+        self.0
+            .iter()
+            .filter_map(|(amt, notes)| {
+                if amt != except {
+                    Some(notes.len())
+                } else {
+                    None
+                }
+            })
+            .max()
+            .unwrap_or_default()
     }
 
     /// Verifies that all keys in `self` are present in the keys of the given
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 /target/
 # in case it's a symlink
 /target
+fedimint-client/target/
```
dpc, line 4: This should no longer be needed. You might just have an old file from before https://github.com/fedimint/fedimint/pull/2286 . Or the new release didn't actually fix it. But they said they did.

fedimint-core/src/tiered_multi.rs
```diff
@@ -124,9 +124,20 @@ impl<T> TieredMulti<T> {
...
+            .iter()
+            .filter_map(|(amt, notes)| {
+                if amt != except {
+                    Some(notes.len())
+                } else {
+                    None
+                }
+            })
+            .max()
+            .unwrap_or(0)
```
douglaz, line 140: nit:
```suggestion
            .unwrap_or_default()
```

fedimint-core/src/tiered_multi.rs
```diff
@@ -124,9 +124,20 @@ impl<T> TieredMulti<T> {
             .flat_map(|(amt, notes)| notes.into_iter().map(move |c| (amt, c)))
     }
 
-    /// Returns the length of the longest vector of all tiers
-    pub fn longest_tier_len(&self) -> usize {
-        self.0.values().map(|notes| notes.len()).max().unwrap_or(0)
+    /// Returns the length of the longest vector of all tiers, except the max
```
douglaz, line 127:
```suggestion
    /// Returns the length of the longest vector of all tiers, except the `except`
```

.gitignore
jkitman: Didn't seem to fix it for me :frowning_face:

jkitman: Or maybe it did?  I'll remove it from this PR.

elsirion (approved):

PR #2297, opened 2023-04-21
[wip] build.sh => fedimint-bin-test

Maan2003:


fedimint-bin-tests/src/main.rs
```diff
@@ -31,6 +31,8 @@ use tracing::{info, warn};
 mod util;
 use util::*;
 
+mod dirs;
```
justinmoon, line 34: Why call it `dirs`?

justinmoon:
Would be nice to move this function into this `mod` since it also deals with environment variables

https://github.com/fedimint/fedimint/blob/b01810900b6a771644cc9943b460aa1231fbbe35/fedimint-bin-tests/src/federation.rs#L14-L43

fedimint-bin-tests/src/main.rs
Maan2003: Because I chose the name before writing the code. Should be called env or vars

PR #2305, opened 2023-04-23
chore: get rid of openssl

### LINKED ISSUE #2200: Get rid of openssl

jkitman:
Fixes #2200

### DIFF
--- a/fedimint-client-legacy/Cargo.toml
+++ b/fedimint-client-legacy/Cargo.toml
@@ -45,17 +45,18 @@ tbs = { path = "../crypto/tbs" }
 thiserror = "1.0.39"
 tokio = { version = "1.26.0", features = ["sync"] }
 tracing = "0.1.37"
-jsonrpsee-types = "0.16.0"
-jsonrpsee-core = { version = "0.16.2", features = [ "client" ] }
+# TODO: use official release, but right now there's a bug that stalls our client
+jsonrpsee-types = "0.18.0"
+jsonrpsee-core = { version = "0.18.0", features = [ "client" ] }
 serde_json = "1.0.91"
 url = { version = "2.3.1", features = ["serde"] }
 threshold_crypto = { git = "https://github.com/fedimint/threshold_crypto" }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-jsonrpsee-ws-client = "0.16.2"
+jsonrpsee-ws-client = { version = "0.18.0", features = ["webpki-tls"], default-features = false }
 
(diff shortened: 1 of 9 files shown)


fedimint-core/src/api.rs
```diff
@@ -692,7 +691,6 @@ impl JsonRpcClient for WsClient {
     async fn connect(url: &Url) -> result::Result<Self, JsonRpcError> {
         #[cfg(not(target_family = "wasm"))]
         return WsClientBuilder::default()
-            .certificate_store(CertificateStore::WebPki)
```
jkitman, line 695: I guess this can safely be removed?

jkitman:
Fails during `ecash_can_be_recovered` when `client.request` never returns (server works fine).

fedimint-core/Cargo.toml
```diff
@@ -19,8 +19,9 @@ futures = "0.3.24"
 bincode = "1.3.1"
 bech32 = "0.9.1"
 itertools = "0.10.5"
-jsonrpsee-types = "0.16.0"
-jsonrpsee-core = { version = "0.16.2", features = [ "client" ] }
+# TODO: use official release, but right now there's a bug that stalls our client
+jsonrpsee-types = { version = "0.18.0" }
```
justinmoon, line 23: I don't think this comment is relevant anymore?

fedimint-client-legacy/Cargo.toml
```diff
@@ -45,17 +45,18 @@ tbs = { path = "../crypto/tbs" }
 thiserror = "1.0.39"
 tokio = { version = "1.26.0", features = ["sync"] }
 tracing = "0.1.37"
-jsonrpsee-types = "0.16.0"
-jsonrpsee-core = { version = "0.16.2", features = [ "client" ] }
+# TODO: use official release, but right now there's a bug that stalls our client
+jsonrpsee-types = "0.18.0"
```
justinmoon, line 49: can remove this comment

justinmoon (approved):

misc/git-hooks/pre-commit
```diff
@@ -54,6 +54,12 @@ function check_forbidden_dependencies() {
       return 1
     fi
   done
+  find ./ -name Cargo.lock | while read -r cargo_lock ; do
```
dpc, line 57: :heart:

Issue #1437, opened 2023-01-20
gateway-cli: Non informative error when making a typo in password


dpc:
```
Enter gateway password: <MADE A TYPO HERE>

Error: 
```

dpc:
Good first issue: Please verify if still an issue, fix is so, close the issue.

isaack-njama:
Is there a recommended way to provide a password? I mean other than from the command line, considering the warning.

elsirion:
> Is there a recommended way to provide a password? I mean other than from the command line, considering the warning.

Assuming you are talking about the "please don not pass it as a command line option" warning: you can also pass it as an environment variable.

https://github.com/fedimint/fedimint/blob/4c3a782bb1c3d17905abaf11a001c338b5a8b4d6/fedimintd/src/fedimintd.rs#L41-L42

cbergqvist:
The situation regarding this has improved thanks to #2472. Current output is:
```
$ gateway-cli info
Enter gateway password:
Error: Bad status returned 401 Unauthorized
```
Could be closed IMHO.

PR #2308, opened 2023-04-24
refactor: show error message from invalid password

isaack-njama:
Helps with #1437

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -236,11 +236,21 @@ pub async fn print_response(response: reqwest::Response) {
 }
 
 pub fn source_password(rpcpassword: Option<String>) -> String {
-    match rpcpassword {
-        None => rpassword::prompt_password("Enter gateway password:").unwrap(),
-        Some(password) => {
-            eprintln!("WARNING: Passing in a password from the command line may be less secure!");
-            password
+    match env::var("FM_PASSWORD") {
+        Ok(password) => password,
+        Err(_) => {
+            match rpcpassword {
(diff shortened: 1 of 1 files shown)


gateway/cli/src/main.rs
```diff
@@ -236,11 +236,21 @@ pub async fn print_response(response: reqwest::Response) {
 }
 
 pub fn source_password(rpcpassword: Option<String>) -> String {
-    match rpcpassword {
-        None => rpassword::prompt_password("Enter gateway password:").unwrap(),
-        Some(password) => {
-            eprintln!("WARNING: Passing in a password from the command line may be less secure!");
-            password
+    match env::var("FM_PASSWORD") {
```
elsirion, line 239: This should already be handled by clap, no need in reading from env directly.

gateway/cli/src/main.rs
```diff
@@ -236,11 +236,21 @@ pub async fn print_response(response: reqwest::Response) {
 }
 
 pub fn source_password(rpcpassword: Option<String>) -> String {
-    match rpcpassword {
-        None => rpassword::prompt_password("Enter gateway password:").unwrap(),
```
elsirion, line 240: If this `unwrap` was the problem we'd see a stack trace in #1437, the actual password checking happens elsewhere. As the function name implies, this function merely returns the password and doesn't check it, so this PR doesn't fix the problem.

elsirion (changes requested):

justinmoon:
Closing to remove inactive PRs. Feel free to re-open if you work on it again.

PR #2311, opened 2023-04-24
refactor: expose auth in api context

jkitman:
Makes it so we don't have to pass in auth twice in endpoints that need to read it.


elsirion (approved):
Unrelated, but could we make `check_no_auth(context)?;` part of a `authenticated_api_endpoint!` macro?

jkitman:
> could we make check_no_auth(context)?; part of a authenticated_api_endpoint! macro?

@elsirion we could, but it would introduce some duplication of the macro code.  Plus we may want to respond differently rather than reject requests when auth is found.

PR #2322, opened 2023-04-25
chore: expose `fedimint-bin-tests` as a flake output

dpc:
I don't see a big advantage of building it in CI, as it is already used for all tests anyway, but external users might want to consume it as a standalone binary.

Verified with:

```
nix build -L .#fedimint-bin-tests
```


justinmoon (approved):
Worked for me

PR #2330, opened 2023-04-26
 refactor: encode configs as Encodable, not serde_json::Value 

dpc:
The current way of using `serde_json::Value` is hashing serde json
serialization string which is probably a bad idea anyway, and
had other shortcomings.

For readability, we leave a field with (redundant) json encodings,
so that humans can view them.

The patch quickly spiraled into a huge refactoring.

I think some more tweaking might be needed, but if the general approach
seems acceptable, we should land to avoid too many merge conflicts.

Created #2281 to review usage of `SerdeEncodable` which is probably
avoidable in many places.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1241,6 +1241,7 @@ dependencies = [
  "futures",
  "getrandom",
  "gloo-timers",
+ "hex",
  "itertools",
  "js-sys",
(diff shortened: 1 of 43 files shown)


dpc:
Landed in another PR.

Issue #2315, opened 2023-04-24
Modularize integration test framework


justinmoon:
* The [test framework](https://github.com/fedimint/fedimint/tree/master/integrationtests/tests/fixtures) defined inside [integrationtests](https://github.com/fedimint/fedimint/tree/master/integrationtests) should probably become a library that could be consumed by external module developers. Perhaps it could live in the same crate as `"(1)"` defined in [this issue](https://github.com/fedimint/fedimint/issues/2260) about splitting `fedimint-bin-tests`.
* The test framework would need to be modularized. AFAIK there is currently no way to add or remove modules.

jkitman:
Thinking about this we need to go through the following:

- [ ] Remove the old `FakeFed` making sure we already have enough unit/integration test coverage
- [ ] Make `fedimint-testing` module-agnostic
- [ ] Refactor `integrationtests` fixtures into `fedimint-testing` and module-specific tests
- [ ] Improve the fixtures / mocks so we can get more out of our test framework

PR #2331, opened 2023-04-26
Refactor: Make fedimint-testing module-agnostic

jkitman:
A few refactors to make `fedimint-testing` no longer depend on modules.
- **move TxOutProof into fedimint-core**, seems like a reasonable place for it to exist if it's part of our generic `BitcoinTest` fixture
- **move FakeWalletGen into integrationtests**, moving the module-specific code into `integrationtests` for now
- **remove duplicate await block height from fedimint-tests**, we already had this implemented in `fedimint-bin-tests` seems redundant.  Eventually we'll need to migrate `peg-in.sh` entirely to rust to get rid of the bash code.

Subtask of #2315


scripts/lib.sh
```diff
@@ -22,7 +22,11 @@ function open_channel() {
 }
 
 function await_fedimint_block_sync() {
-  fixtures await-fedimint-block-sync
+  BLOCKS="$($FM_BTC_CLIENT getblockchaininfo | jq -e -r '.blocks')"
+  FINALITY_DELAY=$(get_finality_delay)
+  AWAIT="$((BLOCKS - FINALITY_DELAY))"
+  echo "await_fedimint_block_sync $AWAIT"
+  $FM_MINT_CLIENT wait-block-height "$AWAIT"
```
justinmoon, line 29: This wasn't working. @Maan2003 should be able to kill this for good pretty soon!

justinmoon (approved):
LGTM

scripts/lib.sh
jkitman: It was working just redundant to have this in Rust twice.  Think we should just convert the `setup-tmuxinator` to be part of `fedimint-bin-tests` then we can eliminate most of the remaining bash scripts.

PR #2296, opened 2023-04-21
fix: remove unused `SerdeEncodable`

elsirion:
fixes #2281

PR #2333, opened 2023-04-27
fix(lightning): replace bincode encoding impl

elsirion:
Gets rid of one other place where we used `bincode` so far, follow up to #2296.


fedimint-core/src/encoding/tbs.rs
```diff
@@ -97,6 +97,42 @@ impl Decodable for tbs::BlindingKey {
     }
 }
 
+impl Encodable for threshold_crypto::Ciphertext {
+    fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
+        self.to_bytes().consensus_encode(writer)
+    }
+}
+
+impl Decodable for threshold_crypto::Ciphertext {
```
m1sterc001guy, line 106: Just FYI, I needed to do something similar for `threshold_crypto::PublicKey` in https://github.com/fedimint/fedimint/pull/2279

elsirion:
I think this will be the first time for me writing a DB migration judging by the CI failure ^^

elsirion:
300 loc of DB migration code later … :laughing:

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1160,3 +1170,264 @@ mod fedimint_migration_tests {
...
+                    expiry_time: value.expiry_time,
+                };
+
+                (new_key, new_value)
+            })
+            .collect::<Vec<_>>()
+            .await;
+
+        for (key, value) in offers {
+            // TODO: this overwrites, allow writing without reading without the warnings
```
m1sterc001guy, line 1356: Confused by this comment

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1160,3 +1170,264 @@ mod fedimint_migration_tests {
...
+        PreimageDecryptionShare,
+    };
+    use fedimint_ln_common::db::{
+        AgreedDecryptionShareKey, ContractKey, OfferKey, ProposeDecryptionShareKey,
+    };
+    use fedimint_ln_common::ContractAccount;
+    use futures::stream::StreamExt;
+    use serde::{Deserialize, Serialize};
+
+    pub async fn migration_v0(dbtx: &mut DatabaseTransaction<'_>) -> anyhow::Result<()> {
```
m1sterc001guy, line 1195: Let's add a comment here on what the db migration is doing (i.e which structs changed encodings). This is to remove the dependency on `bincode` right?

m1sterc001guy:
Do you think it is worth adding and maintaining this migration code? It might be worth taking advantage of the fact that we're pre-release and can make breaking changes for now :)

If we do want to go forward with it, can you add another backup in `prepare_migration_snapshots`? That way we'll be able to detect a db break if these structs ever change again.

elsirion:
It was mostly a learning exercise for me to see how it would work. Not sure if worth keeping as a reference or if I should remove it. How would one go about simply re-generating the checked-in DB?

dpc:
@elsirion This still alive? Should we just regen db and land?

elsirion:
Yeah, regen+land, I'll take a look later.

elsirion:
> Yeah, regen+land, I'll take a look later.

maxresdefault

Maan2003 (approved):
LGTM

Maan2003 (approved):

Maan2003:
reviews from two tabs 😄

PR #2335, opened 2023-04-27
Refactor: move BTC/LN integrationtests fixtures into fedimint-testing

jkitman:
Moves all BTC/LN fixture code from `integrationtests` to `fedimint-testing`

Related to #2315


m1sterc001guy (approved):
Looks simple enough to me

PR #2340, opened 2023-04-27
feature: remove the state constraints in the config gen API

jkitman:
As discussed, this change enables the leader to set the configs before/after/during receiving connections from peers.


okjodom (approved):
ack.

okjodom:
Ha. Didn't mean to merge :)

PR #2343, opened 2023-04-27
refactor(nix): use scope overrides for cleaner CARGO_PROFILE handling

dpc:
The way overriding `CARGO_PROFILE` (which controls cargo build profile used by `crane`) was implemented was completely backwards and a hack. Instead we now use `overrideScope'` to override `commonProfile` which is used everywhere in `craneCommon.nix`.

Nix scopes are a powerful, though a bit complex mechanism, that is used already by `crane`: https://web.archive.org/web/20230427230255/https://crane.dev/advanced/overriding-function-behavior.html so it makes sense to utilize it.

Some further refactoring to clean things up even more might soon follow, and there are some ideas to make it supported better in crane itself: https://github.com/ipetkov/crane/issues/306


flake.nix
```diff
@@ -194,10 +204,11 @@
           # purposes (like `nix build`).
           legacyPackages =
             let
-
-              overrideCargoProfileRecursively = deriv: profile: deriv.overrideAttrs (oldAttrs: {
```
dpc, line 198: This was a terrible hack and now it's no more. Instead we use `overrideScope'` to override `commonProfile` which is used everywhere in `craneCommon.nix`.

douglaz (approved):
I'm no nix expert but looks good

justinmoon (approved):

Issue #2337, opened 2023-04-27
Build Guardian UI Experience components


isaack-njama:
The Guardian UI comprises pages made of components. Tasks are listed per components:

- [ ] SPA wrapper
- [ ] Guardian Role Selector
- [ ] Guardian Role Option
- [ ] Color Mode Switcher
- [ ] Style Role Selectors

[Figma](https://www.figma.com/file/ioeAVbWBc3ow2RaZETatM7/Fedimint-Setup-Flow?node-id=0-1&t=lWkqGrDVvibwsv1I-0)
[Components](https://excalidraw.com/#room=c09bcc9ab669db604fb0,ir8ObJHojcaQ6uSHtqXK5Q)

PR #2344, opened 2023-04-28
feat: create and reference app wrapper

isaack-njama:
Towards #2337


okjodom (approved):
looking forward to improved styling on the role selectors. 🚀

PR #2346, opened 2023-04-28
fix: avoids eslint and prettier conflict in Guardian UI

garyray-k:
Adds prettier to the eslint config so they do not conflict.


okjodom:
`gateway/ui` could use a similar fix: https://github.com/fedimint/fedimint/blob/master/gateway/ui/.eslintrc.json

okjodom (approved):

PR #2347, opened 2023-04-28
fix(mint-client): allow injecting user-specified meta into operations

elsirion:
Proof-of-concept for adding more meta data to operations, needs to be done for other ext traits too.


justinmoon (approved):
LGTM

PR #2352, opened 2023-04-29
chore: improve naming of mprocs & tmuxinator

justinmoon:
I sometimes forget the name of the mprocs dev tool (`fed-shell`). Renaming it to `mprocs` because that's easier to remember :)

Also, renamed `devimint tmuxinator` to `devimint dev-fed` since this is also used by `mprocs` and it calls a `dev_fed` function internally which is a nice name.

After the refactors I re-ran both `just mprocs` and `just tmuxinator` and both seemed to still work!

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -635,7 +635,7 @@ async fn reconnect_test(dev_fed: DevFed, process_mgr: &ProcessManager) -> Result
 #[derive(Subcommand)]
 enum Cmd {
     ExternalDaemons,
-    Tmuxinator,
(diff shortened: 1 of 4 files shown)


Maan2003:
Should we also rename tmuxinator to tmux?

justinmoon:
> Should we also rename tmuxinator to tmux?

I'm in favor. I was thinking about doing that in this PR, but just did the more obvious mprocs renaming.

Maan2003 (approved):

elsirion (approved):

PR #2353, opened 2023-04-29
Allow external modules using Client NG's ClientModule trait to handle fedimint-cli commands

justinmoon:
We had a `handle_cli_command` method on the old `ClientModule` trait. This PR moves it to the `ClientModule` trait for the new client.

Like the old method, it just passes `&[ffi::OsString]`. It might be better to utilize some kind of clap sub-command. But I figured that could wait for a future PR.

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -8,15 +8,13 @@ use std::path::PathBuf;
 use std::process::exit;
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


justinmoon:
Hmmm ... this doesn't quite work ... now I get what these `XClientExt` traits are how we get access to the whole client to do CLI operations.

justinmoon:
Closing in favor of https://github.com/fedimint/fedimint/pull/2368

PR #2358, opened 2023-04-29
DO NOT MERGE: bug investigation

jkitman:
@m1sterc001guy can you help debug why `test_migration` fails with this small change?


m1sterc001guy:
We synced offline. The fix for this is to just delete the db/migrations/global-v0 folder and re-run `cargo test prepare_migration_snapshots`. 

Alternatively, we could write a db migration to fix the test. But there's not really much point since no one is running the dummy module in prod currently and that would just add to the code that we need to maintain.

Issue #2257, opened 2023-04-19
Convert (most) logging statements to use pre-defined logging targets


dpc:
In #1469 we've introduced a set of pre-defined logging targets, to make log filtering more stable and useful. ([PR](https://github.com/fedimint/fedimint/pull/1571))

While a good start, there's still plenty of code that could use going though it and adjusting `target:`s.

elsirion:
Thx for the reminder, totally forgot about that in client-ng :see_no_evil:

dpc:
We've been gradually using this, and I think most of the code uses this convention now, so no need to keep this issue open.

PR #2368, opened 2023-05-01
feat: Client NG module clients can handle CLI commands

justinmoon:
Up to now, we don't have a way for an externally developed module to register a CLI command. You have needed to fork this repo and edit `fedimint-cli`. 

This PR allows new client modules to handle CLI commands themselves.

One thing that would be eventually be nice is to have client modules define the commands they can handle using `clap` primitives. This would be more ergonomic then using `std::ffi::OsString`, and perhaps we could dynamically attach them as first-level commands. For example, `fedimint-cli ping` instead of `fedimint-cli --id starter --arg ping`. Then we could move [these commands](https://github.com/fedimint/fedimint/blob/acc5a9b76249c6aaf1d9708a1662fe7e5fc4b1ab/fedimint-cli/src/ng.rs#L23-L42) and their handlers into the modules themselves, and out of `fedimint-cli`.

See https://github.com/fedimint/fedimint-custom-modules-example/pull/1 for a demo.


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -469,6 +473,54 @@ impl ClientModule for MintClientModule {
             fee: self.cfg.fee_consensus.note_issuance_abs * (output.0.count_items() as u64),
         }
     }
+
+    async fn handle_cli_command(
```
justinmoon, line 477: Not sure if this is worth keeping.

dpc (dismissed):

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -469,6 +473,54 @@ impl ClientModule for MintClientModule {
...
+                    .subscribe_reissue_external_notes_updates(operation_id)
+                    .await
+                    .unwrap();
+
+                while let Some(update) = updates.next().await {
+                    if let ReissueExternalNotesState::Failed(e) = update {
+                        return Err(anyhow::Error::msg(format!("Reissue failed: {e}")));
+                    }
+
+                    info!("Update: {:?}", update);
```
garyray-k, line 514: https://github.com/fedimint/fedimint/issues/2257

justinmoon:
> I'm happy to get it in, but wouldn't be complete or fully consistent. I think once I'm finished with the wallet client I'll migrate the ng commands to this interface and then we can completely switch over.

Cool. I'm not in a big hurry here. Just wanted to make sure that progress we made wasn't lost.

PR #2373, opened 2023-05-01
feat: Lightning Client Receive State Machine

m1sterc001guy:
Implementation of the lightning client receive state machine that awaits for the LightingGateway to fund an `IncomingContract` that can be claimed by the client.

### DIFF
--- a/fedimint-cli/src/ng.rs
+++ b/fedimint-cli/src/ng.rs
@@ -4,15 +4,15 @@ use std::time::Duration;
 use bitcoin::secp256k1;
 use bitcoin_hashes::hex::ToHex;
 use clap::Subcommand;
-use fedimint_client::ClientBuilder;
+use fedimint_client::{Client, ClientBuilder};
 use fedimint_core::config::ClientConfig;
 use fedimint_core::core::{LEGACY_HARDCODED_INSTANCE_ID_LN, LEGACY_HARDCODED_INSTANCE_ID_MINT};
 use fedimint_core::db::IDatabase;
 use fedimint_core::encoding::Decodable;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::task::TaskGroup;
-use fedimint_core::{Amount, TieredMulti, TieredSummary};
-use fedimint_ln_client::{LightningClientExt, LightningClientGen, LnPayState};
+use fedimint_core::{Amount, ParseAmountError, TieredMulti, TieredSummary};
+use fedimint_ln_client::{LightningClientExt, LightningClientGen, LnPayState, LnReceiveState};
 use fedimint_mint_client::{MintClientExt, MintClientGen, MintClientModule, SpendableNote};
 use fedimint_wallet_client::WalletClientGen;
 use futures::StreamExt;
@@ -29,6 +29,14 @@ pub enum ClientNg {
     },
     Spend {
         amount: Amount,
+    },
+    LnInvoice {
+        #[clap(long, value_parser = parse_fedimint_amount)]
+        amount: Amount,
+        #[clap(long, default_value = "")]
+        description: String,
+        #[clap(long)]
+        expiry_time: Option<u64>,
     },
     LnPay {
         bolt11: lightning_invoice::Invoice,
@@ -64,19 +72,7 @@ pub async fn handle_ng_command<D: IDatabase>(
 
     match command {
         ClientNg::Info => {
-            let mint_client = client
-                .get_module_client::<MintClientModule>(LEGACY_HARDCODED_INSTANCE_ID_MINT)
-                .unwrap();
-            let summary = mint_client
-                .get_wallet_summary(
-                    &mut client.db().begin_transaction().await.with_module_prefix(1),
-                )
-                .await;
-            Ok(serde_json::to_value(InfoResponse {
-                total_msat: summary.total_amount(),
-                denominations_msat: summary,
-            })
-            .unwrap())
+            return get_note_summary(&client).await;
         }
         ClientNg::Reissue { notes } => {
             let amount = notes.total_amount();
@@ -103,6 +99,38 @@ pub async fn handle_ng_command<D: IDatabase>(
 
             Ok(serde_json::to_value(notes).unwrap())
         }
+        ClientNg::LnInvoice {
+            amount,
+            description,
+            expiry_time,
+        } => {
+            let mut dbtx = client.db().begin_transaction().await;
+            let active_gateway = client
+                .fetch_active_gateway(&mut dbtx.with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_LN))
+                .await?;
+            dbtx.commit_tx().await;
+
+            let operation_id = client
+                .create_bolt11_invoice_and_receive(amount, description, expiry_time, active_gateway)
+                .await?;
+            let mut updates = client.subscribe_to_ln_receive_updates(operation_id).await?;
+            while let Some(update) = updates.next().await {
+                match update {
+                    LnReceiveState::Claimed { txid } => {
+                        client.await_claim_notes(operation_id, txid).await?;
+                        return get_note_summary(&client).await;
+                    }
+                    LnReceiveState::Canceled { reason } => {
+                        return Err(reason.into());
+                    }
+                    _ => {}
+                }
+
+                info!("Update: {:?}", update);
+            }
+
+            return Err(anyhow::anyhow!("Unknown Lightning receive state"));
+        }
         ClientNg::LnPay { bolt11 } => {
             let mut dbtx = client.db().begin_transaction().await;
             let active_gateway = client
@@ -126,9 +154,7 @@ pub async fn handle_ng_command<D: IDatabase>(
                         .unwrap());
                     }
                     LnPayState::Refunded { refund_txid } => {
-                        client
-                            .await_lightning_refund(operation_id, refund_txid)
-                            .await?;
+                        client.await_claim_notes(operation_id, refund_txid).await?;
                     }
                     _ => {}
                 }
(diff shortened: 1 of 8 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -190,12 +228,121 @@ impl LightningClientExt for Client {
...
+    async fn create_bolt11_invoice_and_receive(
+        &self,
+        amount: Amount,
+        description: String,
+        expiry_time: Option<u64>,
+        gateway: LightningGateway,
+    ) -> anyhow::Result<OperationId> {
+        let (ln_client_id, ln_client) = ln_client(self);
+        // TODO: Get network from Wallet config
+        let network = Network::Regtest;
```
m1sterc001guy, line 240: Need to implement this so it works on other networks

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -0,0 +1,385 @@
...
+                        keypair,
+                        contract,
+                        dbtx,
+                        global_context.clone(),
+                    ))
+                },
+            ),
+            // TODO: Do we even want this case?
+            // Should it just be set to the invoice expiry?
+            StateTransition::new(Self::await_payment_timeout(), |_dbtx, (), old_state| {
```
m1sterc001guy, line 227: I think the timeout here should always be set to the invoice expiry? Otherwise its possible for the invoice to be paid and the contract funded but the client SM won't be waiting to claim it.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -0,0 +1,385 @@
...
+        invoice: Invoice,
+        payment_keypair: KeyPair,
+    ) -> Result<ConfirmedInvoice, LightningReceiveError> {
+        let outpoint = OutPoint { txid, out_idx: 0 };
+        let timeout = Duration::from_secs(15);
+        global_context
+            .api()
+            .await_output_outcome::<LightningOutputOutcome>(outpoint, timeout, &module_decoder)
+            .await
+            .map_err(|_| LightningReceiveError::Rejected)?;
```
elsirion, line 152: Since we aren't doing anything with the output itself we could just await the transaction being accepted via the `await_tx_accepted` fn on the global context. That way we don't need the decoder.

Even better, that way the `await_invoice_rejection` wouldn't be needed since `await_tx_accepted` returns a result that can be used to decide to which following state to transition.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    async fn await_incoming_contract_account(
+        invoice: Invoice,
+        global_context: DynGlobalClientContext,
+    ) -> IncomingContractAccount {
+        // TODO: Get rid of polling
+        loop {
+            let contract_id = (*invoice.payment_hash()).into();
+            let contract =
+                LightningClientContext::get_incoming_contract(contract_id, global_context.clone())
```
elsirion, line 241: nit: feels like `get_incoming_contract` should be a member of `LnFederationApi`

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -0,0 +1,385 @@
...
+            // need to be created
+            state_machines: Arc::new(|_, _| vec![]),
+        };
+
+        let txid = global_context.claim_input(dbtx, client_input).await;
+        OutPoint { txid, out_idx: 0 }
+    }
+
+    async fn await_payment_timeout(timeout: Duration) {
+        sleep(timeout).await
```
elsirion, line 286: We'll probably want to add some buffer time since clocks tend to not be perfectly synchronized and we'd rather wait for payment for too long, making some useless API requests, than lose money.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -0,0 +1,385 @@
...
+            }
+            Err(_) => {
+                // Claim rejection
+                LightningReceiveStateMachine {
+                    common: old_state.common,
+                    state: LightningReceiveStates::Canceled(LightningReceiveError::ClaimRejected),
+                }
+            }
+        }
+    }
```
elsirion, line 350: Unclear these transitions are needed, but ok :)

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -1,18 +1,44 @@
...
+        .get_first_instance(&WalletCommonGen::KIND)
+        .expect("No ln module attached to client");
+
+    let wallet_client = client
+        .get_module_client::<WalletClientModule>(wallet_client_instance)
+        .expect("Instance ID exists, we just fetched it");
+
+    (wallet_client_instance, wallet_client)
+}
+
```
elsirion, line 41: Slightly confused about these changes, were they needed?

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -190,12 +229,120 @@ impl LightningClientExt for Client {
...
+        let operation_id = invoice.payment_hash().into_inner();
+        let output = ln_client.create_offer_output(
+            operation_id,
+            amount,
+            invoice.clone(),
+            keypair,
+            payment_hash,
+            preimage,
+            expiry_time,
+        )?;
```
elsirion, line 260: Why are `generate_unconfirmed_invoice` and `create_offer_output` two functions? Would we ever want to generate an unconfirmed invoice without also creating the output?

elsirion (commented):
Looks mostly good, great work! Only one bug that needs to be fixed before merge since it could be forgotten too easily and would be flaky since it only occurs under specific timing conditions. Also left a few nits that don't need addressing in this PR, more as a guide for further work.

elsirion:
Also, adding a test would be great!

fedimint-cli/src/ng.rs
```diff
@@ -30,6 +30,14 @@ pub enum ClientNg {
     Spend {
         amount: Amount,
     },
+    LnInvoice {
+        #[clap(long, value_parser = parse_fedimint_amount)]
```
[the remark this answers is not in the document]
m1sterc001guy: Just copied this over from the previous CLI. It allows for an optional denomination in the amount (i.e sat). Probably a nice thing to keep.

modules/fedimint-wallet-client/src/lib.rs
m1sterc001guy: This is just to get the `bitcoin::Network` for the invoice, which is in the `WalletClientConfig`. I think they should be needed once the Wallet client is implemented? If you have a better suggestion for this, I can change it :)

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: This is mainly an artifact of copying the existing code, I will refactor so its clearer. You are correct, two functions are not needed.

modules/fedimint-wallet-client/src/lib.rs
elsirion: This should probably be a) in the global config or b) in the LN config too. We should not make clients dependent on each other. But not a big deal for MVP, maybe add a TODO to refactor.

docs/dev-env.md
```diff
@@ -175,7 +175,7 @@ $ docker run -it -v $PWD/demo:/var/fedimint -p 17240:17240 fedimintd:iqviraxy2cz
 write the files to the host file-system (e.g. config generation). `-p` is used to bind the host's port 17240 as the
 container's port 17240.
 
-To generate federation configs and run the federation please see the `run_dkg` and `start_federation` helper functions in [lib.sh](https://github.com/fedimint/fedimint/blob/master/scripts/lib.sh), as well as the `--help` output of
+To generate federation configs and run the federation please see the `run_dkg` and `start_federation` helper functions in devimint, as well as the `--help` output of
```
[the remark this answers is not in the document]
m1sterc001guy: Rebased

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -0,0 +1,385 @@
...
+    Canceled(LightningReceiveError),
+    ConfirmedInvoice(LightningReceiveConfirmedInvoice),
+    Funded(LightningReceiveFunded),
+    Success(TransactionId),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
+pub struct LightningReceiveCommon {
+    pub operation_id: OperationId,
+}
```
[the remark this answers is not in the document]
m1sterc001guy: Removed `LightningReceiveCommon`

modules/fedimint-ln-client/src/receive.rs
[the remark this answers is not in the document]
m1sterc001guy: This was removed to make it simpler

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    async fn transition_funded(
+        old_state: LightningReceiveStateMachine,
+        keypair: KeyPair,
+        contract: IncomingContractAccount,
+        dbtx: &mut ClientSMDatabaseTransaction<'_, '_>,
+        global_context: DynGlobalClientContext,
+    ) -> LightningReceiveStateMachine {
+        let outpoint = Self::claim_incoming_contract(dbtx, contract, keypair, global_context).await;
```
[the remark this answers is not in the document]
m1sterc001guy: This should be fixed now. I added a check that verifies the Decryption status of the preimage before transitioning

modules/fedimint-ln-client/src/receive.rs
m1sterc001guy: Added 10% buffer to the existing timeout

modules/fedimint-wallet-client/src/lib.rs
m1sterc001guy: I added  a TODO in the `LightningClientModule` where the `bitcoin::Network` is read.

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Refactored into one function

fedimint-cli/src/ng.rs
```diff
@@ -103,6 +99,38 @@ pub async fn handle_ng_command<D: IDatabase>(
...
+            expiry_time,
+        } => {
+            let mut dbtx = client.db().begin_transaction().await;
+            let active_gateway = client
+                .fetch_active_gateway(&mut dbtx.with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_LN))
+                .await?;
+            dbtx.commit_tx().await;
+
+            let operation_id = client
+                .create_bolt11_invoice_and_receive(amount, description, expiry_time, active_gateway)
```
justinmoon, line 114: Would be nice if this function also returned the invoice, too. Then callers could postentially return here and present the invoice, and listen for updates later.

m1sterc001guy: Easy enough change. It is also available in the operation log (so listeners to state machine updates can get it from there)

justinmoon: ahh ok. maybe i just don't understand the API yet.

m1sterc001guy: It's fine to return it here too. I made the change.

fedimint-cli/src/ng.rs
```diff
@@ -103,6 +99,38 @@ pub async fn handle_ng_command<D: IDatabase>(
...
+        ClientNg::LnInvoice {
+            amount,
+            description,
+            expiry_time,
+        } => {
+            let mut dbtx = client.db().begin_transaction().await;
+            let active_gateway = client
+                .fetch_active_gateway(&mut dbtx.with_module_prefix(LEGACY_HARDCODED_INSTANCE_ID_LN))
+                .await?;
+            dbtx.commit_tx().await;
```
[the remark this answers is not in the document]
m1sterc001guy: I refactored this a bit so that it doesn't need to take a `dbtx`. This allowed me to get rid of `LEGACY_HARDCODED_INSTANCE_ID_LN`.

elsirion (approved):

PR #2374, opened 2023-05-01
chore: add logging target to devimint

garyray-k:
https://github.com/fedimint/fedimint/issues/2257


dpc (approved):
I wonder if devimint should have a sub-module `LOT_TEST_DEVIMINT` or something, but a move in the right direction anyway.

PR #2375, opened 2023-05-01
feat: Expose full server status

douglaz:
Open for feedback

Current known issues:
1) Some checks fails (see `TODO: find out why this is flaky`)

Pending:
1) Perhaps a similar version on config api? (returning most null values)
2) More unit tests


fedimint-core/src/admin_client.rs
```diff
@@ -216,6 +216,15 @@ pub enum ServerStatus {
     ConsensusRunning,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FullStatus {
+    pub our_epoch: u64,
```
dpc, line 221: Not sure about `our` part. Just `epoch`, `self_`? `own_`? Bikesheding required.

fedimint-core/src/admin_client.rs
```diff
@@ -216,6 +216,15 @@ pub enum ServerStatus {
...
+    pub last_epoch_by_peer: HashMap<PeerId, u64>,
+    pub peers_offline: u64,
+    pub peers_not_synchronized: u64,
```
dpc, line 224: `peers_behind`, `peers_stale`, `peers_out_of_sync`?

fedimint-core/src/admin_client.rs
```diff
@@ -216,6 +216,15 @@ pub enum ServerStatus {
...
+    pub status: ServerStatus,
+}
```
dpc, line 226: FullStatus vs `statsu` vs `ServerStatus`. Too much status maybe.

fedimint-server/src/config/mod.rs
```diff
@@ -804,9 +804,9 @@ where
     T: std::fmt::Debug + Clone + Serialize + DeserializeOwned + Unpin + Send + Sync + 'static,
 {
     let connector = TlsTcpConnector::new(certs, network.identity).into_dyn();
-    ReconnectPeerConnections::new(network, delay_calculator, connector, task_group)
-        .await
-        .into_dyn()
+    let (connections, _) =
```
dpc, line 807: Nit: Seems unnecessary to have that variable.

fedimint-server/src/net/api.rs
```diff
@@ -69,10 +74,19 @@ pub struct ConsensusApi {
...
     pub api_sender: Sender<ApiEvent>,
-
+    pub peer_status_client: PeerStatusClient,
+    pub rejoin_at_epoch: RejoinAtEpoch,
+    pub full_status_cache: ExpiringCache<ApiResult<FullStatus>>,
     pub supported_api_versions: SupportedApiVersionsSummary,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum FullPeerStatus {
```
dpc, line 84: `PeerConsensusStatus` or something?

fedimint-server/src/net/api.rs
```diff
@@ -296,6 +310,73 @@ impl ConsensusApi {
             .await
             .map_err(|_| ApiError::server_error("Unable send event".to_string()))
     }
+
+    pub async fn build_full_server_status(&self) -> ApiResult<FullStatus> {
```
dpc, line 314: That "full" is meh and it keeps repeating :)

fedimint-server/src/net/api.rs
```diff
@@ -296,6 +310,73 @@ impl ConsensusApi {
...
+        let our_epoch = self.get_epoch_count().await;
+        // Clone to avoid holding the lock while we calculate things
+        let rejoin_at_epoch = self.rejoin_at_epoch.lock().await.clone();
```
dpc, line 317: There's got to be a better name for `rejoin_at_epoch` because when I'm reading it confuses me (I know you didn't introduce it, just commenting as I'm reading).

fedimint-server/src/net/api.rs
```diff
@@ -296,6 +310,73 @@ impl ConsensusApi {
...
+        let last_epoch_by_peer = rejoin_at_epoch.into_iter().fold(
+            HashMap::<PeerId, u64>::new(),
+            |mut acc, (epoch, peers)| {
+                let epoch = epoch + 1;
+                for peer in peers {
+                    match acc.entry(peer) {
```
dpc, line 323: `acc.entry(peer).and_modify(...).or_insert(...)` should be neater.

https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.and_modify

fedimint-core/src/admin_client.rs
douglaz: I would lean into `peers_out_of_sync`. The metric is how many peers there are with `status != FullPeerStatus::Synchronized`

fedimint-core/src/admin_client.rs
douglaz: Naming stuff is hard :sweat_smile:

fedimint-server/src/net/api.rs
dpc: `rejoin` is a verb

fedimint-server/src/net/api.rs
```diff
@@ -296,6 +310,73 @@ impl ConsensusApi {
...
+                        }
+                        std::collections::hash_map::Entry::Vacant(entry) => {
+                            entry.insert(epoch);
+                        }
+                    }
+                }
+                acc
+            },
+        );
+        let peers_status = self.peer_status_client.query_all_status().await;
```
dpc, line 338: Eehh... so we went with the active queering approach? :/

fedimint-server/src/net/api.rs
```diff
@@ -296,6 +310,73 @@ impl ConsensusApi {
...
     }
+
+    pub async fn build_full_server_status(&self) -> ApiResult<FullStatus> {
+        let our_epoch = self.get_epoch_count().await;
+        // Clone to avoid holding the lock while we calculate things
+        let rejoin_at_epoch = self.rejoin_at_epoch.lock().await.clone();
+        let last_epoch_by_peer = rejoin_at_epoch.into_iter().fold(
+            HashMap::<PeerId, u64>::new(),
+            |mut acc, (epoch, peers)| {
+                let epoch = epoch + 1;
```
dpc, line 321: I'm not sure why `+ 1`.

fedimint-server/src/net/api.rs
```diff
@@ -296,6 +310,73 @@ impl ConsensusApi {
...
+
+        let full_peer_status = peers
+            .into_iter()
+            .map(|peer| {
+                let full_status = match (
+                    peers_status.get(&peer),
+                    last_epoch_by_peer.get(&peer).copied(),
+                ) {
+                    (Some(Err(e)), _) => {
+                        warn!(target: LOG_NET_API, %peer, "Unable to get peer status: {e:?}");
```
dpc, line 350: Probably not a "warning" level material. Other peers being down might be not unusual, especially in larger federations.

fedimint-server/src/net/api.rs
dpc: `build` suggest some kind of a builder. While this seems to just "return" (the status).

fedimint-server/src/net/peers.rs
```diff
@@ -78,6 +80,52 @@ struct PeerConnectionStateMachine<M> {
...
+#[derive(Clone)]
+pub struct PeerStatusClient(HashMap<PeerId, PeerStatusChannelSender>);
+
+impl PeerStatusClient {
+    pub async fn query_all_status(&self) -> HashMap<PeerId, anyhow::Result<PeerStatus>> {
+        let mut results = HashMap::new();
+        for (peer_id, sender) in self.0.iter() {
+            let (response_sender, response_receiver) = oneshot::channel();
+            let query = PeerStatusQuery {
+                response: response_sender,
```
dpc, line 105: `response_sender` or `response_tx` or something is better name than `response`, especially in a `struct`, IMO.

fedimint-server/src/net/peers.rs
```diff
@@ -78,6 +80,52 @@ struct PeerConnectionStateMachine<M> {
...
+type PeerStatusChannelSender = Sender<PeerStatusQuery>;
+type PeerStatusChannelReceiver = Receiver<PeerStatusQuery>;
+
+#[derive(Clone)]
+pub struct PeerStatusClient(HashMap<PeerId, PeerStatusChannelSender>);
+
+impl PeerStatusClient {
+    pub async fn query_all_status(&self) -> HashMap<PeerId, anyhow::Result<PeerStatus>> {
+        let mut results = HashMap::new();
+        for (peer_id, sender) in self.0.iter() {
```
dpc, line 102: If I'm seeing it right, this should be possible to be

```
self.0.iter().map(...).collect()
```

but when I was typing it I've noticed that this way (and the way it's written currently), it would be done serially. We should probably send all the requests first, then collect all the responses to let stuff happen in parallel.

fedimint-server/src/net/api.rs
douglaz: Yeah, I thought this would be controversial :smile: 

It was my first idea and It was very similar to one of @elsirion's ideas, so I left it as is.

The only sort of "passive" approach I can think of is to create a "events" channel on `CommonPeerConnectionState` and write to it after each state transition and let a thread accumulate this state. Then we would query this state instead of the original one.

We can try that.

fedimint-server/src/net/api.rs
douglaz: AFAIK, the reason is the same why there is a `+ 1` here:
https://github.com/fedimint/fedimint/blob/60595f294c854a24fad47243ec4d42ad67603a59/fedimint-server/src/net/api.rs#L268

fedimint-server/src/net/api.rs
douglaz: I think this only happens when a peer has been banned. If it's just disconnected, the we get the answer below

fedimint-server/src/net/peers.rs
```diff
@@ -383,6 +442,13 @@ where
                     },
                 }
             },
+            Some(status_query) = self.status_query_receiver.recv() => {
+                if status_query.response.send(PeerStatus::Connected).is_err() {
+                    let peer_id = self.peer;
+                    warn!(target: LOG_NET_PEER, %peer_id, "Could not send peer status response: receiver dropped");
```
dpc, line 448: We generally have this system of signaling global shutdown after which threads start voluntarily finishing, and corresponding senders might not be able to send any more messages

In such circumstances they should either ignore (maybe log `debug!`) not being able to send, or even better exit themselves (not always possible).

integrationtests/tests/fixtures/mod.rs
```diff
@@ -795,6 +797,23 @@ impl FederationTest {
         result
     }
 
+    /// Get fedimint peer status from all federation servers
+    pub async fn full_server_status(&self) -> Vec<Result<FullStatus, ApiError>> {
+        let mut result = Vec::new();
+        for server in &self.servers {
```
dpc, line 803: `iter().map(...).collect()` instead of mutating in a loop preferred.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1025,6 +1044,23 @@ impl FederationTest {
         }
     }
 
+    pub async fn assert_peers_synchronized(&self) {
```
dpc, line 1047: `assert_all_peers_synchronized`?

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1025,6 +1044,23 @@ impl FederationTest {
...
+        let full_server_status = self.full_server_status().await;
+        println!("assert_peers_synchronized: full_server_status: {full_server_status:?}");
+        assert!(full_server_status
+            .into_iter()
+            .all(|s| s.unwrap().peers_not_synchronized == 0));
+    }
+
+    pub async fn assert_peers_not_synchronized(&self) {
```
dpc, line 1055: Isn't this just `!assert_all_peers_synchronized`? Probably not worth having both.

integrationtests/tests/fixtures/mod.rs
```diff
@@ -1025,6 +1044,23 @@ impl FederationTest {
...
+        let full_server_status = self.full_server_status().await;
```
dpc, line 1056: Given the caching, lags etc. I don't think we should use status endpoints for asserting anything in tests.

fedimint-server/src/net/peers.rs
```diff
@@ -78,6 +80,52 @@ struct PeerConnectionStateMachine<M> {
...
+pub enum PeerStatus {
+    Connected,
+    Disconnected,
+}
+
+type PeerStatusChannelSender = Sender<PeerStatusQuery>;
+type PeerStatusChannelReceiver = Receiver<PeerStatusQuery>;
+
+#[derive(Clone)]
+pub struct PeerStatusClient(HashMap<PeerId, PeerStatusChannelSender>);
```
dpc, line 97: Just to make it more self-explanatory I'd make it a :+1: 

```rust
struct PeerStatusClient {
  request_senders : HashMap<PeerId, PeerStatusChannelSender>,
}
```

BTW. The biggest problem with this rpc call approach (vs. just shared Arc<some_atomic_counters>` is that now there's bunch of extra code, and people trying to understand this code might get confused by the word `Client`, etc.

Please add a decently detailed comment that explains what is it for and roughly how it works.

fedimint-server/src/net/peers.rs
dpc: Could be maybe `self.0.iter().map(...send...).collect().into_iter().map(...receive...).collect()`. :)

fedimint-server/src/net/api.rs
```diff
@@ -445,3 +534,35 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
...
+            data: Arc::new(tokio::sync::Mutex::new(None)),
+            duration,
+        }
+    }
+
+    pub async fn get<Fut>(&self, f: impl FnOnce() -> Fut) -> T
+    where
+        Fut: futures::Future<Output = T>,
+    {
+        let mut data = self.data.lock().await;
```
dpc, line 558: Since we're holding this lock over the call to `f()` there is a some room for deadlocks here, if the `f()` somehow triggered a recursive locking here. Which is unlikely, but not impossible as the caller might not realize it.

But I'm not sure if not holding it over `f()` is much better.

integrationtests/tests/fixtures/mod.rs
douglaz: There is no caching (this is the method that the endpoints calls, not the result of the endpoint) and the information returned is the most up to date one regarding the internal state of the guardians.

But I agree that perhaps we shouldn't call it on tests right know because there is some point of synchronization missing (e.g maybe we still need to wait further things to happen until we are sure all the nodes are perfectly connected and have caught up the consensus)

fedimint-server/src/net/api.rs
dpc: Actually, nvmd. This is a a call to send a request to peer threads. I wish we could use some other names than "query" and "client". 

IMO this should be

```rust
self.peer_status.get();
```

or something. The fact that there are some "rpc"-like things there can be an implementation detail.

dpc (commented):
Bunch of nits and naming bikeshedding.

I have no idea about this `rejoin_at_epoch`. @jkitman should look at it. :)

integrationtests/tests/fixtures/mod.rs
dpc: Oh, right. No caching, but indeed some synchronization might be needed. And also - this is supposed to be a diagnostic functionality. By utilizing it in tests, we now couple it with tests and other code, making potential changes more difficult and have a chance of breaking something more important that just status endpoint.

fedimint-server/src/net/api.rs
dpc: Oh. So it's `let epoch_count = epoch + 1;`. Maybe even `epoch` should be `last_epoch` or something. Or maybe not.

fedimint-core/src/admin_client.rs
dpc: The thing is that `_not_` is kind meh, because generally one wants to avoid negations in names. *But* by using a negation we make it easier for simple monitoring tools t match that the response contains `"peers_out_of_sync": 0` literally. So it's preferable that all the stats are `0 == good`.

BTW. Might be worthwhile to put this context in a comment somewhere.

fedimint-server/src/net/api.rs
dpc: That *and* when the whole server is shutting down. https://github.com/fedimint/fedimint/blob/60595f294c854a24fad47243ec4d42ad67603a59/fedimint-server/src/net/peers.rs#L218

When user presses ctrl+c or systemd sends `SIGTERM` we don't want to see panics and random errors, ideally. That's why it's important to handle such conditions gracefully.

fedimint-server/src/net/api.rs
douglaz: If we don't hold the lock then multiple calls of `f()` may happen in parallel, defeating the whole purpose of this mechanism.

But if someone introduces a deadlock then it will probably happen on the first test/execution. It think it will be very clear that you did something wrong.

douglaz:
Addresses many comments

integrationtests/tests/fixtures/mod.rs
douglaz: Generally I agree. But then we are left with the question on how to really test this endpoint. It would be good to be sure it can detect issues on different scenarios. Adding it to existing integration tests seemed the most effective way.

But we can say that, for this particular case, unit tests and manual testing is enough.

fedimint-server/src/net/peers.rs
douglaz: Look now

integrationtests/tests/tests.rs
```diff
@@ -1079,6 +1083,8 @@ async fn rejoin_consensus_single_peer() -> Result<()> {
 
         let new_height = client2.await_consensus_block_height(height).await.unwrap();
         assert_eq!(new_height, height);
+        // TODO: find out why this is flaky
+        // fed.assert_peers_synchronized().await;
```
jkitman, line 1087: These tests intentionally have peers fall behind in epochs.

fedimint-core/src/admin_client.rs
jkitman: I would combine this with the existing status endpoint if possible, it would only be available when running consensus though.

fedimint-server/src/consensus/server.rs
```diff
@@ -213,13 +216,19 @@ impl ConsensusServer {
...
             db: db.clone(),
             modules: modules.clone(),
             client_cfg,
             api_sender,
             supported_api_versions: ServerConfig::supported_api_versions_summary(&modules),
+            rejoin_at_epoch: Arc::clone(&rejoin_at_epoch),
+            peer_status_channels,
+            // keep the status for a short time to protect the system against a denial-of-service
+            // attack
+            full_status_cache: ExpiringCache::new(Duration::from_secs(1)),
```
jkitman, line 231: What's the DoS attack here?  Generally we don't prevent these here, we will have to eventually deal with them at the consensus layer.

integrationtests/tests/tests.rs
douglaz: :thinking: 
But are they not coming back?

douglaz:
Addressed more comments

fedimint-server/src/consensus/server.rs
douglaz: Someone may call this unauthenticated endpoint a lot of times, causing a slowdown in the application. A similar concern exist for similar endpoints like `fetch_epoch_count`.

fedimint-server/src/net/peers.rs
douglaz: There is no client anymore :eyes:

douglaz:
Changed a bit the output, now it's something like:

```json
{
    "last_contribution": 22,
    "peers_flagged": 0,
    "peers_offline": 0,
    "peers_online": 3,
    "status": "ConsensusRunning",
    "status_by_peer": {
      "1": {
        "flagged": false,
        "last_contribution": 22,
        "last_contribution_timestamp_seconds": 1683062231
      },
      "2": {
        "flagged": false,
        "last_contribution": 22,
        "last_contribution_timestamp_seconds": 1683062231
      },
      "3": {
        "flagged": false,
        "last_contribution": 22,
        "last_contribution_timestamp_seconds": 1683062231
      }
    }
  }
```

For HBBFT `contribution` is `epoch count`. 
`flagged` is a warning, it means "this peer may need some attention".

fedimint-server/src/consensus/server.rs
dpc: `fetch_epoch_count` is probably trivially fast. In this case a message is put on the queue to all the peer threads, so there's (some, probably minor) a possibility that someone could cause these to be always full and stale the communication between consensus thread and peer threads.

fedimint-server/src/config/api.rs
```diff
@@ -652,6 +653,15 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConfigGenApi>> {
                 }
             }
         },
+        api_endpoint! {
```
dpc, line 656: This full vs non-full seems unnecessary. Should we merge them?

Maybe this one could be `peer_status` and the other one `dkg_status` or `server_status?

fedimint-server/src/net/api.rs
```diff
@@ -296,6 +305,73 @@ impl ConsensusApi {
             .await
             .map_err(|_| ApiError::server_error("Unable send event".to_string()))
     }
+
+    pub async fn full_server_status(&self) -> ApiResult<FullServerStatus> {
```
dpc, line 309: Same here "full" is just not very descriptive.

fedimint-server/src/net/api.rs
dpc: I know. It would not help with a burst of messages, but it will still short-cut all subsequent messages (for 1s).

But yeah, probably not worth changing.

dpc (approved):
That "full" seems meh, but other than this LGTM. Please get an ACK from @jkitman before landing.

jkitman (changes requested):
Looks better now, my only remaining suggestion is to merge this with the existing status API which should be pretty easy, you just need to modify the existing enum to be `ServerStatus::ConsensusRunning(FullServerStatus)`

fedimint-server/src/consensus/server.rs
```diff
@@ -327,8 +336,12 @@ impl ConsensusServer {
...
         // for checking the hashes of the epoch history
         let mut prev_epoch: Option<SignedEpochOutcome> = self.last_processed_epoch.clone();
-        // we can remove tracking past epochs
-        self.rejoin_at_epoch.retain(|k, _| k > &last_outcome.epoch);
+        // we can remove tracking past epochs, but not completely since this structure
+        // is used to build last_epoch_by_peer
+        self.rejoin_at_epoch
+            .lock()
+            .await
+            .retain(|k, _| k >= &last_outcome.epoch);
```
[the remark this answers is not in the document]
douglaz: It's not using this anymroe

fedimint-core/src/api.rs
```diff
@@ -840,6 +841,28 @@ fn url_to_string_with_default_port(url: &Url) -> String {
 
 impl<C: JsonRpcClient> WsFederationApi<C> {}
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FullServerStatus {
```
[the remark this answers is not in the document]
douglaz: Now there are the details

douglaz:
> It exists for all our endpoints, we should handle it in a uniform way, rather than ad-hoc.
> Isn't this a P2P request as well? So it would only be from another peer?
It exists mostly for unauthenticated endpoints that retrieve information. I think it makes sense to consider adding it to similar endpoints, we can open an issue for that.
Because it's unauthenticated it's a bit easier for someone with network access to the machine to call this. Even though a authenticated endpoint can be a target of a similar attack.

douglaz:
> I'm not sure, still trying to understand how you are calculating "in sync". In general 2f+1 peers will be on the same epoch, but not all of them necessarily will be.

I remove these tests, they would be brittle

douglaz:
> Looks better now, my only remaining suggestion is to merge this with the existing status API which should be pretty easy, you just need to modify the existing enum to be ServerStatus::ConsensusRunning(FullServerStatus)

Will address this soon

jkitman (approved):
Looks good lets add a CLI command for it next!

PR #2376, opened 2023-05-01
chore: add target to bitcoind log macro using constant

garyray-k:
#2257

### DIFF
--- a/fedimint-logging/src/lib.rs
+++ b/fedimint-logging/src/lib.rs
@@ -20,6 +20,7 @@ pub const LOG_TASK: &str = "task";
 pub const LOG_TEST: &str = "test";
 pub const LOG_TIMING: &str = "timing";
 pub const LOG_WALLET: &str = "wallet";
+pub const LOG_BITCOIND: &str = "bitcoind";
 
 /// Consolidates the setup of server tracing into a helper
 #[derive(Default)]
--- a/fedimint-logging/src/lib.rs
+++ b/fedimint-logging/src/lib.rs
@@ -7,6 +7,7 @@ use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Layer};
 
 /// Constants for categorizing the logging type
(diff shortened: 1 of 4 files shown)


fedimint-logging/src/lib.rs
```diff
@@ -19,6 +19,7 @@ pub const LOG_TASK: &str = "task";
 pub const LOG_WALLET: &str = "wallet";
 pub const LOG_TEST: &str = "test";
 pub const LOG_ECASH_RECOVERY: &str = "ecash-recovery";
+pub const LOG_BITCOIND: &str = "bitcoind";
```
dpc, line 22: Not sure if this is waht we want to go with it. Generally we want to have "categories" of stuff here. `net::blockchain` or something maybe?

m1sterc001guy: Should this be refactored into an enum?

garyray-k: Would a nested enum work?
```rs
enum LogTarget {
    Net(NetTarget),
    ....
}

enum NetTarget {
    Blockchain,
    ....
}
```

dpc: Logging targets must be strings, AFAIK, and generally correspond (and default to) Rust module naming, so `foo:bar::xyz` etc. They are only used for logging filtering via `RUST_LOG`. No point in trying to make them anything other than string contants, AFAICT.

garyray-k: With enums it looks like we'd have to impl Display for each variation to convert to strings. Probably not worth it when we can just keep this section of constants tidy and organized.

elsirion: Coming back to the beginning of this discussion, I think a rename to `net::blockchain` would make sense.

garyray-k: 😅
oh yea...
done.

elsirion (approved):

PR #2377, opened 2023-05-01
feat: devimint's run-ui command has an option to drive the new UI

justinmoon:
This new option doesn't set the password because UI will need to set this, or `FM_LISTEN_UI` because this will block the config API from running.

Leaving as "draft" because fedimintd's password is currently required. So not setting `FM_PASSWORD` currently causes `fedimintd` to fail.

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -11,6 +11,8 @@ use tokio::sync::mpsc::{self, Receiver, Sender};
 
 use super::*; // TODO: remove this
 
+type EnvVars = HashMap<String, String>;
+
 /// Create a map of environment variables which fedimintd and DKG can use,
 /// but which can't be defined by `build.sh` because multiple of these daemons
 /// run concurrently with different values.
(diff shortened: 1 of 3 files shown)


elsirion:
> Leaving as "draft" because fedimintd's password is currently required. So not setting `FM_PASSWORD` currently causes `fedimintd` to fail.

Is there another PR that fixes that, which this one is thus dependent on?

justinmoon:
> Is there another PR that fixes that, which this one is thus dependent on?

It's still unclear to me how the new setup UI's password is related to `FM_PASSWORD`. Me and @jkitman talked a bit about it [here](https://discord.com/channels/990354215060795454/1102678024891154492/1102678024891154492).

jkitman:
If you pass in `FM_PASSWORD` it gets written to a file that can be read by the config gen API.  Without it you will need to send an API request with the password everytime the server starts.

justinmoon:
> Without it you will need to send an API request with the password everytime the server starts.

Manual intervention whenever fedimintd restarts sound unworkable. Maybe fedimintd needs it's own required password that will be set in deployment environment, and config API has a separate password is just specific to the API?

jkitman:
> Manual intervention whenever fedimintd restarts sound unworkable

That's what the password file would be for, we could default to creating one for now.  Some users may not wish to store the plaintext password on the fedimintd box, so we can keep it optional.

Maan2003:
conflicts with #2413 but that the conflicts should be fixed in that PR IMO.

Maan2003 (approved):
LGTM

PR #2378, opened 2023-05-01
fix: notifications for ModuleDatabaseTransaction

Maan2003:
The notifications for ModuleDatabaseTransaction don't work at all. It doesn't call add_notification_key in insert,remove like DatabaseTransaction. 

Updated the test to catch the bug.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -2412,11 +2412,12 @@ mod tests {
         let db = Database::new(MemDatabase::new(), ModuleDecoderRegistry::default());
         let db = db.new_isolated(module_instance_id);
 
+        let key_task = waiter(&db, TestKey(1)).await;
+
         let mut tx = db.begin_transaction().await;
         tx.insert_new_entry(&key, &val).await;
         tx.commit_tx().await;
 
-        let key_task = waiter(&db, TestKey(1)).await;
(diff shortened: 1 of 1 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -2431,13 +2432,14 @@ mod tests {
         let val = TestVal(2);
         let db = Database::new(MemDatabase::new(), ModuleDecoderRegistry::default());
 
+        let key_task = waiter(&db.new_isolated(module_instance_id), TestKey(1)).await;
```
m1sterc001guy, line 2435: Was this failing because the db wasn't isolated before?

m1sterc001guy:
Did CI catch this break?

fedimint-core/src/db/mod.rs
Maan2003: no, this doesn't fix the test. this makes the test actually fail.
because notification in moduledatabasetransaction don't work at all.

jkitman:
> Did CI catch this break?

I caught it when I was creating the new dummy module.

jkitman (approved):
Confirmed this works cherry-picking into my module branch. :+1:

elsirion:
CI problems likely related to https://www.githubstatus.com/incidents/lhw4ny5js29b

PR #2379, opened 2023-05-02
feature: modular federation tests

jkitman:
Very basic federation test fixture, mostly copied from our existing tests, but now module-agnostic.

Will run the servers in the background to make writing tests easier and not epoch-dependent.

Related to #2315


elsirion (approved):
I don't get the full vision for the new testing framework yet, but I think a lot of APIs will become clearer once tests get migrated over (curious how the primary client module problem will be solved, I guess just always depending on the mint client for now? or did you make enough progress with the CB module?). LGTM.

justinmoon:
LGTM

Issue #2129, opened 2023-04-05
Gateway UI: Theming


okjodom:
The goal is to create a consistent and visually appealing theme that will enhance the user experience. By introducing a theming framework, we can easily make updates and changes to the app's appearance without having to update every component individually.

Ref:
- https://chakra-ui.com/docs/styled-system/customize-theme
- https://chakra-ui.com/docs/styled-system/advanced-theming

isaack-njama:
I would like to work on this

okjodom:
closing as duplicate of #2464

PR #2383, opened 2023-05-02
Gateway UI Theming

isaack-njama:
This PR introduces theming to Gateway UI, as mentioned in #2129 

Before:
image

After:
image


okjodom:
ack.

- Please add pic of before / after in PR description.
- **let's sync on priority of contributions to the UI projects:** Gateway UI is lower pri at this time, and we need to standardize aspects like theming, fonts, localization, across the different projects.

okjodom (approved):

PR #2384, opened 2023-05-02
Refactor: consensus api

jkitman:
A couple small refactors to the consensus API to eliminate some boilerplate in tests and ensure the check of consensus hash will be tested.


justinmoon (approved):
LGTM

PR #2386, opened 2023-05-02
feature: turn dummy module into simple example module

### LINKED ISSUE #2359: Turn dummy module into simple example module
Currently we don't get a lot of value out of our dummy module, but still need to modify it during refactors.  Instead of stubbing out implementations we could create a simple example module for:

- An ideal starting point for module writers that demonstrates every basic operation once - an API call, a transaction, a consensus item
- Keeping track of the boiler-plate code that we should try to eliminate for module writers
- Enable a "money printer" module that will make it easier to run integration tests
- Getting 90% of the way towards a proper account-based module (requested by some orgs)

This is also a necessary step towards #2315

### A useful example should include:
**Server**
- Config gen with a private and consensus key
- Example of a consensus item and threshold signing
- DB migrations
- Validation of input/outputs amounts and keys
- Audit of assets/liabilities
- API endpoint that notifies consensus
- API endpoint that blocks on db

**Client**
- Primary module impl
- Create a transaction between users
- Parse an output outcome
- Server API calls
- Simple state machine for refunding tx input and claiming tx output

**Tests**
- A user-to-user transaction
- API calls
- Refund / output claiming
- Submitting invalid txs

@elsirion @justinmoon ^ any things I'm missing as far as a useful example is concerned?

jkitman:
Fixes #2359 (see there for details)


modules/fedimint-dummy-tests/tests/tests.rs
```diff
@@ -0,0 +1,34 @@
+use fedimint_core::{sats, Amount};
```
justinmoon, line 1: Does this mean that the fixtures now work with external modules to some extent?

jkitman: Yep this is the first passing test written in the modular test framework :smile: 

Over the next couple weeks I'll be converting our old tests to use the new framework to test the new clients.

justinmoon: Amazing

modules/fedimint-dummy-client/src/lib.rs
```diff
@@ -1 +1,295 @@
...
+        let (id, dummy) = dummy_client(self);
+        let mut dbtx = self.db().begin_transaction().await;
+        let op_id = rand::random();
+
+        // TODO: Building a tx could be easier
+        // Create input using our own account
+        let input = dummy
+            .create_sufficient_input(&mut dbtx.with_module_prefix(id), op_id, amount)
+            .await?;
+        dbtx.commit_tx().await;
```
m1sterc001guy, line 87: Do you want to commit the transaction here? Wouldn't this deduct the balance before the transaction is accepted by the fed?

Not sure how robust we want to make the dummy module, but I think one way you could do it would be to spawn a state machine that waits on the output of this TX and then updates the local balance after verifying that the transaction was successful.

m1sterc001guy: Or maybe just move `commit_tx` underneath `await_tx_accepted`

modules/fedimint-dummy-client/src/states.rs
```diff
@@ -0,0 +1,82 @@
...
+    type ModuleContext = DummyClientContext;
+    type GlobalContext = DynGlobalClientContext;
+
+    fn transitions(
+        &self,
+        _context: &Self::ModuleContext,
+        global_context: &Self::GlobalContext,
+    ) -> Vec<StateTransition<Self>> {
+        match self.clone() {
+            DummyStateMachine::Input(amount, txid, id) => vec![StateTransition::new(
```
m1sterc001guy, line 29: Why do we need both of these? They look the same to me

modules/fedimint-dummy-client/src/states.rs
```diff
@@ -0,0 +1,82 @@
...
+    DummyStateMachine::Done
+}
+
+// TODO: Boiler-plate, should return OutputOutcome
+async fn await_tx_accepted(
+    context: DynGlobalClientContext,
+    id: OperationId,
+    txid: TransactionId,
+) -> Result<(), ()> {
+    context.clone().await_tx_accepted(id, txid).await
```
m1sterc001guy, line 72: Is the `clone` necessary?

modules/fedimint-dummy-common/src/db.rs
```diff
@@ -1,64 +1,107 @@
...
-        dbtx.insert_new_entry(&key_v2, &val).await;
-    }
-    Ok(())
-}
+use crate::{DummyOutputOutcome, DummyPrintMoneyRequest};
 
+/// Namespaces DB keys for this module
 #[repr(u8)]
 #[derive(Clone, EnumIter, Debug)]
 pub enum DbKeyPrefix {
```
m1sterc001guy, line 14: Nit: not a big deal for the dummy module but I think its a bit weird to have a shared db.rs for the client/server since I don't think they shouldn't ever access the same database structures

modules/fedimint-dummy-common/src/db.rs
```diff
@@ -1,64 +1,107 @@
...
+    db_prefix = DbKeyPrefix::DummyOutcome,
+);
+impl_db_lookup!(
+    key = DummyOutcomeKeyV1,
+    query_prefix = DummyOutcomeKeyV1Prefix
+);
+
+/// Lookup funds for a user by key or prefix
+#[derive(Debug, Clone, Encodable, Decodable, Eq, PartialEq, Hash, Serialize)]
+pub struct DummyFundsKeyV1(pub XOnlyPublicKey);
```
m1sterc001guy, line 81: Does `DummyFundsKeyV1` have the same data as `DummyFundsKeyV0`? Can we make them different so that the migration actually changes the data structure?

modules/fedimint-dummy-common/src/lib.rs
```diff
@@ -4,23 +4,104 @@ use async_trait::async_trait;
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DummyConfigGenParams {
+    pub tx_fee: Amount,
+}
+
+/// Input for a fedimint transaction
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
+pub struct DummyInput {
+    pub amount: Amount,
+    // // Associate the input with a user's pubkey
```
m1sterc001guy, line 47: Copy error?

modules/fedimint-dummy-common/src/lib.rs
```diff
@@ -4,23 +4,104 @@ use async_trait::async_trait;
...
+    pub account: XOnlyPublicKey,
+}
+
+/// Output for a fedimint transaction
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
+pub struct DummyOutput {
+    pub amount: Amount,
+    // // Associate the output with a user's pubkey
```
m1sterc001guy, line 55: Copy error?

m1sterc001guy:
Really nice PR! I'll let others take a look before merging since there's a lot of new code added. Most of my comments are nits

modules/fedimint-dummy-client/src/api.rs
```diff
@@ -0,0 +1,35 @@
+use fedimint_core::api::{FederationApiExt, FederationResult, IFederationApi};
+use fedimint_core::module::ApiRequestErased;
+use fedimint_core::task::{MaybeSend, MaybeSync};
+use fedimint_core::{apply, async_trait_maybe_send, Amount};
+use fedimint_dummy_common::{DummyPrintMoneyRequest, LEGACY_HARDCODED_INSTANCE_ID_DUMMY};
+use secp256k1::XOnlyPublicKey;
+
+#[apply(async_trait_maybe_send!)]
+pub trait DummyFederationApi {
```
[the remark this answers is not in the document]
jkitman: Maybe `Example`?  It's 90% of the way to being a proper `Account` based module as well.  I'll defer this to another PR.

modules/fedimint-dummy-client/src/lib.rs
jkitman: Does `finalize_and_submit_transaction` start the input state machine?  If so, there is already a refund in that state machine.

The state machine is the trickest part of understand.

modules/fedimint-dummy-client/src/states.rs
jkitman: It's so the client can track their balance independently of the fed (similar to how the mint module works).

If input fails, we get refunded.  If output fails, we don't.
If input succeeds, nothing happens since we already deducted.  If output succeeds, we add new funds.

modules/fedimint-dummy-common/src/db.rs
jkitman: Was following a pattern in the other modules, but I think you are right.

modules/fedimint-dummy-common/src/db.rs
jkitman: Sure, it's mostly for an example anyway.

modules/fedimint-dummy-client/src/lib.rs
jkitman: I will add tests for these later.

modules/fedimint-dummy-client/src/lib.rs
```diff
@@ -1 +1,295 @@
...
+    ) -> anyhow::Result<OutPoint> {
+        let (id, dummy) = dummy_client(self);
+        let mut dbtx = self.db().begin_transaction().await;
+        let op_id = rand::random();
+
+        // TODO: Building a tx could be easier
+        // Create input using our own account
+        let input = dummy
+            .create_sufficient_input(&mut dbtx.with_module_prefix(id), op_id, amount)
+            .await?;
```
elsirion, line 86: Why are we making the assumption that the user wants to spend from the dummy module wallet? Idiomatically one would create an unfunded transaction here and leave it to `finalize_and_submit_transaction` to fund it using the primary module. That *might* be the dummy module if configured, but not necessarily.

modules/fedimint-dummy-server/src/lib.rs
```diff
@@ -209,24 +364,63 @@ impl ServerModule for Dummy {
...
+                // API allows users ask the fed to print money
+                "print_money",
+                async |module: &Dummy, context, request: DummyPrintMoneyRequest| -> () {
+                    // TODO: A way to send messages to consensus without the DB
+                    let mut dbtx = context.dbtx();
+                    dbtx.insert_entry(&DummyPrintKeyV1(request), &()).await;
+                    module.print_notify.notify_one();
+                    Ok(())
+                }
+            },
```
elsirion, line 397: Why not just use a transaction with a special-cased input/output for that? That might also solve #2387

modules/fedimint-dummy-client/src/lib.rs
```diff
@@ -1 +1,295 @@
...
+pub struct DummyClientModule {
+    cfg: DummyClientConfig,
+    key: KeyPair,
+}
+
+/// Data needed by the state machine
+#[derive(Debug, Clone)]
+pub struct DummyClientContext;
+
+// TODO: Boiler-plate
```
elsirion, line 171: I wonder how much of this could become derive macros …

modules/fedimint-dummy-common/src/lib.rs
```diff
@@ -4,23 +4,103 @@ use async_trait::async_trait;
...
+    pub tx_fee: Amount,
+}
+
+/// Input for a fedimint transaction
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
+pub struct DummyInput {
+    pub amount: Amount,
+    /// Associate the input with a user's pubkey
+    pub account: XOnlyPublicKey,
+}
```
elsirion, line 48: I think it would be prudent to include some sort of nonce, otherwise a two separate transfers between the same accounts of the same value would be identical (same txid) and thus the second would not go through. This won't happen for a while, but when it does it will be hard to debug.

elsirion (approved):
I'm ok with it's current form, some nits and things that could be improved imo, but can happen as we go.

modules/fedimint-dummy-client/src/lib.rs
jkitman: :+1: That could be useful for funding our ecash wallet...

modules/fedimint-dummy-server/src/lib.rs
jkitman: Thought about it, but I wanted an example of how to use consensus items and threshold signatures instead of another transaction example.  Maybe I could demo some other way?

PR #2391, opened 2023-05-03
Refactor: Inline config functions

jkitman:
This refactor inlines some of the config functions to eliminate unnecessary indirection and move implementations to where it makes sense (on the server side).  This will also help us eventually replace the config module boilerplate with macros.


douglaz (approved):
LGTM

justinmoon (approved):

PR #2395, opened 2023-05-03
feature: config gen modifications based on UI feedback

jkitman:
- Makes `get_consensus_config_gen_params` return all peers with `ServerStatus`
- Adds new statuses `SharingConfigGenParams`, `ReadyForConfigGen`, `ConfigGenFailed`, `VerifyingConfigs`
- Doesn’t fail when auth is unnecessarily passed in
- Returns all verification codes, leaves checking verification up to the UI

Also simplifies how we handle state.

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeMap;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::sync::Mutex;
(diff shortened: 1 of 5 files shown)


justinmoon (approved):

fedimint-server/src/config/api.rs
```diff
@@ -803,6 +714,26 @@ mod tests {
...
+                    .filter(|(_, param)| param.status != Some(status.clone()))
+                    .collect();
+                if mismatched.is_empty() {
+                    break;
+                }
+                tracing::info!(
+                    target: fedimint_logging::LOG_TEST,
+                    "Test retrying server status"
+                );
+                sleep(Duration::from_millis(10)).await;
```
garyray-k, line 733: Any thought to backoff on this retry? I see this logging like crazy with a 10ms sleep and log entry every iteration.

jkitman: Just for testing, needs to pass this before DKG finishes, that's why the retry is so low, but it finishes quickly.

PR #2396, opened 2023-05-03
fix: remove flaky lightning reconnect test

m1sterc001guy:
This PR removes the flakey lightning gateway reconnect test and re-writes it using `devimint`.

### DIFF
--- a/devimint/src/lib.rs
+++ b/devimint/src/lib.rs
@@ -25,6 +25,7 @@ use tokio::fs;
 use tokio::sync::{MappedMutexGuard, Mutex, MutexGuard};
 use tokio::time::sleep;
 use tonic_lnd::lnrpc::GetInfoRequest;
+use tonic_lnd::LndClient;
 use tracing::{info, warn};
 
 pub mod util;
@@ -131,14 +132,29 @@ impl Bitcoind {
 #[derive(Clone)]
 pub struct Lightningd {
     rpc: Arc<Mutex<ClnRpc>>,
-    _process: ProcessHandle,
+    process: Option<ProcessHandle>,
     bitcoind: Bitcoind,
 }
 
 impl Lightningd {
     pub async fn new(process_mgr: &ProcessManager, bitcoind: Bitcoind) -> Result<Self> {
         let cln_dir = env::var("FM_CLN_DIR")?;
+        let process = Lightningd::start(process_mgr, cln_dir.clone()).await?;
 
+        let socket_cln = PathBuf::from(cln_dir).join("regtest/lightning-rpc");
+        poll("lightningd", || async {
+            Ok(fs::try_exists(&socket_cln).await?)
+        })
+        .await?;
+        let rpc = ClnRpc::new(socket_cln).await?;
+        Ok(Self {
+            bitcoind,
+            rpc: Arc::new(Mutex::new(rpc)),
+            process: Some(process),
+        })
+    }
+
+    pub async fn start(process_mgr: &ProcessManager, cln_dir: String) -> Result<ProcessHandle> {
         let extension_path = cmd!("which", "gateway-cln-extension")
             .out_string()
             .await
@@ -151,19 +167,14 @@ impl Lightningd {
             "--plugin={extension_path}"
         );
 
-        let process = process_mgr.spawn_daemon("lightningd", cmd).await?;
+        process_mgr.spawn_daemon("lightningd", cmd).await
+    }
 
-        let socket_cln = PathBuf::from(cln_dir).join("regtest/lightning-rpc");
-        poll("lightningd", || async {
-            Ok(fs::try_exists(&socket_cln).await?)
-        })
-        .await?;
-        let rpc = ClnRpc::new(socket_cln).await?;
-        Ok(Self {
-            bitcoind,
-            rpc: Arc::new(Mutex::new(rpc)),
-            _process: process,
-        })
+    pub async fn restart(&mut self, process_mgr: &ProcessManager) -> Result<()> {
+        let cln_dir = env::var("FM_CLN_DIR")?;
+        let process = Lightningd::start(process_mgr, cln_dir).await?;
+        self.process = Some(process);
+        Ok(())
     }
 
     pub async fn request<R: cln_rpc::model::IntoRequest>(&self, request: R) -> Result<R::Response>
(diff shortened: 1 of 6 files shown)


devimint/src/main.rs
```diff
@@ -592,6 +592,52 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
+
+    for mut gw in gateways {
+        // Verify that the gateway can query the lightning node for the pubkey and alias
+        let mut cmd = gw.cmd().await.arg("info");
+        assert_eq!(cmd.run().await.is_ok(), true);
+
+        // Verify that after stopping the lightning node, info no longer returns since
+        // the lightning node is unreachable.
+        gw.stop_lightning_node().await?;
+        let info_fut = cmd.run();
```
Maan2003, line 621: interesting reuse of cmd, confused me a lot.

devimint/src/main.rs
```diff
@@ -592,6 +592,52 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
+    } = dev_fed;
+
+    drop(cln);
+    drop(lnd);
+
+    let gateways = vec![gw_cln, gw_lnd];
+
+    for mut gw in gateways {
+        // Verify that the gateway can query the lightning node for the pubkey and alias
+        let mut cmd = gw.cmd().await.arg("info");
```
Maan2003, line 615:
```suggestion
        let mut cmd = cmd!(gw, "info");
```

devimint/src/lib.rs
```diff
@@ -260,6 +304,14 @@ impl Lnd {
         .await?;
         Ok(())
     }
+
+    pub async fn kill(&mut self) -> Result<()> {
```
Maan2003, line 308: kill should take `self` because a `Lnd` value shouldn't exist without a running lnd

devimint/src/lib.rs
```diff
@@ -389,6 +441,20 @@ impl Gatewayd {
         })
     }
 
+    pub async fn stop_lightning_node(&mut self) -> Result<()> {
+        match &mut self.ln {
```
Maan2003, line 445: I think `self.ln` should be removed, we still a new LnKind type for cmd bellow.

devimint/src/lib.rs
```diff
@@ -131,14 +132,29 @@ impl Bitcoind {
 #[derive(Clone)]
 pub struct Lightningd {
     rpc: Arc<Mutex<ClnRpc>>,
-    _process: ProcessHandle,
+    process: Option<ProcessHandle>,
```
Maan2003, line 135: a `Lightningd` value is used as invariant that a lightningd is running.

Maan2003:
does gatewayd require a lightning node at startup, or does it lazily connect.

Maan2003:
then Gatewayd should have `Option<ClnOrLnd>` and `LnKind`

jkitman:
@m1sterc001guy what makes this less flaky?

devimint/src/lib.rs
m1sterc001guy: I'm going to replace this with a trait, I think it will simplify some things

m1sterc001guy:
> then Gatewayd should have `Option<ClnOrLnd>` and `LnKind`

This is tricky, since when we disconnect and change this to `None`, we've now lost the info that stored which lightning node the gateway was connected to. 

I think a better way to do it is to add a trait for the LightningNodes, and have a `kind` function that returns the type of node. That way, `Gatewayd` can store `Option<dyn LightningNodeTrait>` and match on the kind to know which node to restart.

> what makes this less flaky?
It no longer does a lightning payment so it doesn't depend on liquidity. It just query's the nodes pubkey to validate that its connected.

jkitman:
> It no longer does a lightning payment so it doesn't depend on liquidity. It just query's the nodes pubkey to validate that its connected.

Couldn't we just make the LN channel bigger in tests?

dpc (commented):
Looks OK, but defering to @Maan2003

devimint/src/main.rs
m1sterc001guy: Changed to your suggestion below

devimint/src/lib.rs
m1sterc001guy: Instead of making a new trait, I just saved the name of the Lightning Node that the gateway is currently connected to.

devimint/src/main.rs
```diff
@@ -592,6 +592,78 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
+
+    // Drop other references to CLN and LND so that the test can kill them
+    drop(cln);
+    drop(lnd);
+
+    let gateways = vec![gw_cln, gw_lnd];
+
+    for mut gw in gateways {
+        // Verify that the gateway can query the lightning node for the pubkey and alias
+        let mut cmd = cmd!(gw, "info");
```
Maan2003, line 616: if you are reusing this `cmd` then give it a good name say `info_cmd`. But repeating `cmd!(gw, "info")` everytime is also not bad.

devimint/src/main.rs
```diff
@@ -592,6 +592,78 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
+        // Verify that after stopping the lightning node, info no longer returns since
+        // the lightning node is unreachable.
+        gw.stop_lightning_node().await?;
+        let info_fut = cmd.run();
+
+        // CLN will timeout when trying to retrieve the info, LND will return an
+        // explicit error
+        let expected_timeout_or_failure = tokio::time::timeout(Duration::from_secs(3), info_fut)
+            .await
+            .map_err(|_| anyhow::anyhow!("TimeoutError"))
```
Maan2003, line 637: does this work?
```suggestion
            .map_err(Into::into)
```

devimint/src/main.rs
```diff
@@ -592,6 +592,78 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
+                gw.ln = Some(LightningNode::Cln(new_cln));
+            }
+            "lnd" => {
+                let new_lnd = Lnd::new(process_mgr, bitcoind.clone()).await?;
+                gw.ln = Some(LightningNode::Lnd(new_lnd));
+            }
+            _ => {
+                return Err(anyhow::anyhow!(
+                    "Gatewayd is not connected to Lightning Node"
+                ))
```
Maan2003, line 654:
```suggestion
                unreachable!()
```

devimint/src/main.rs
```diff
@@ -592,6 +592,78 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
+
+        // Restart the Lightning Node
+        match ln_node {
+            "cln" => {
+                let new_cln = Lightningd::new(process_mgr, bitcoind.clone()).await?;
+                gw.ln = Some(LightningNode::Cln(new_cln));
+            }
+            "lnd" => {
+                let new_lnd = Lnd::new(process_mgr, bitcoind.clone()).await?;
+                gw.ln = Some(LightningNode::Lnd(new_lnd));
```
Maan2003, line 649: should this a method? `set_lightning_node()`

Maan2003 (commented):
The approach looks good to me. just some nits:

devimint/src/main.rs
```diff
@@ -592,6 +592,69 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
+        // CLN will timeout when trying to retrieve the info, LND will return an
+        // explicit error
+        let expected_timeout_or_failure = tokio::time::timeout(Duration::from_secs(3), info_fut)
+            .await
+            .map_err(Into::into)
+            .and_then(|result| result);
+        assert!(expected_timeout_or_failure.is_err());
+
+        // Restart the Lightning Node
+        match ln_node.as_str() {
```
douglaz, line 635: This is so ugly, why not use the enum `LightningNode`?

m1sterc001guy: We can't because when we kill the process, the `Option` get set to `None` and we have lost the metadata on which Lightning node this gateway was connected to. We can't keep a copy of `LightningNode` because we need to kill the process, and we can't kill the process if there are references to `Lightnind` or `Lnd`.

I described a "better" solution above but that involved much more refactoring and this was simpler.

douglaz: I see. I would then keep it simple and create a
```rust
enum LightningNodeName {
    Lnd,
    Cln
}
```
Then use that instead of the string.

Maan2003 (approved):
LGTM!

please make sure it passes CI atleast 3 times before merging. I am rerunning it once.

PR #2400, opened 2023-05-03
fix: missing wallet module id in ln client

jkitman:


justinmoon:
We need more tests for the new client!

justinmoon (approved):

PR #2401, opened 2023-05-03
feat: expose federation ID in new client

justinmoon:


m1sterc001guy (approved):
Federation ID is needed for the lightning receive case, this PR would allow us to eliminate a parameter when generating the invoice.

elsirion (approved):

PR #2390, opened 2023-05-03
refactor: remove legacy ids from the apis

jkitman:
This removes the legacy ids from the new client APIs, one step closer to being fully modular.

PR #2402, opened 2023-05-04
fix: missing ln module id in ln client

justinmoon:
Follow on after https://github.com/fedimint/fedimint/pull/2400

https://github.com/fedimint/fedimint/pull/2390 broke new client and we didn't notice it because it's not well-tested yet.

You have to manually test this in mprocs / tmuxinator. Before this commit, `fedimint-cli ng ln-invoice` will never succeed. After this commit, it succeeds right after the invoice is paid.


jkitman (approved):
I'll merge for now, but we should remove these legacy ids, once we have testing.

PR #2404, opened 2023-05-04
feat: module (mint in particular) backup for new client

dpc:
Port of the ecash recovery code to new client, now generalized for all client modules (but implemented only for mint module).

Includes `backup`, `restore`, `wipe` command for `fedimint-cli ng`.

The old-client ecash code is still in place. New client code does not have any tests yet, but since the core logic is largely unmodified it should work OK.

I am able to succesfully backup, wipe and restore with a minor issue of `fedimint-cli` not waiting for all the state machines to finish. But I think we should address it in another PR.

### DIFF
--- a/fedimint-cli/src/ng.rs
+++ b/fedimint-cli/src/ng.rs
@@ -1,9 +1,10 @@
-use std::ffi;
 use std::str::FromStr;
 use std::time::Duration;
 
+use anyhow::bail;
 use bitcoin::secp256k1;
 use clap::Subcommand;
+use fedimint_client::backup::Metadata;
 use fedimint_client::sm::OperationId;
 use fedimint_client::Client;
 use fedimint_core::config::ClientConfig;
@@ -18,7 +19,7 @@ use serde::{Deserialize, Serialize};
 use serde_json::json;
 use tracing::info;
 
-use crate::LnInvoiceResponse;
+use crate::{metadata_from_clap_cli, LnInvoiceResponse};
 
 #[derive(Debug, Clone)]
 pub enum ModuleSelector {
@@ -68,15 +69,23 @@ pub enum ClientNg {
         #[clap(value_parser = parse_node_pub_key)]
         pubkey: secp256k1::PublicKey,
     },
-    /// Call module-specific commands
-    Mod {
-        #[clap(long)]
-        id: ModuleSelector,
-
-        /// Command with arguments to call the module with
-        #[clap(long)]
-        arg: Vec<ffi::OsString>,
+    /// Upload the (encrypted) snapshot of mint notes to federation
+    Backup {
+        #[clap(long = "metadata")]
+        /// Backup metadata, encoded as `key=value` (use `--metadata=key=value`,
+        /// possibly multiple times)
+        // TODO: Can we make it `*Map<String, String>` and avoid custom parsing?
+        metadata: Vec<String>,
     },
+    /// Wipe the state of the client (mostly for testing purposes)
+    #[clap(hide = true)]
+    Wipe {
+        #[clap(long)]
+        force: bool,
+    },
+    /// Restore the previously created backup of mint notes (with `backup`
+    /// command)
+    Restore,
 }
 
 pub fn parse_node_pub_key(s: &str) -> Result<secp256k1::PublicKey, secp256k1::Error> {
@@ -215,17 +224,22 @@ pub async fn handle_ng_command(
             gateway_json["active"] = json!(true);
             Ok(serde_json::to_value(gateway_json).unwrap())
         }
-        ClientNg::Mod { id, arg } => {
-            let (id, _) = match id {
-                ModuleSelector::Id(id) => (id, config.get_module_cfg(id)?),
-                ModuleSelector::Kind(kind) => config.get_first_module_by_kind_cfg(kind)?,
-            };
 
-            let module = client
-                .get_module_client_dyn(id)
-                .expect("Module exists according to cfg");
+        ClientNg::Backup { metadata } => {
+            let metadata = metadata_from_clap_cli(metadata)?;
 
-            Ok(module.handle_cli_command(&client, &arg).await?)
+            client
+                .backup_to_federation(Metadata::from_json_serialized(metadata))
+                .await?;
+            Ok(serde_json::to_value(()).unwrap())
+        }
+        ClientNg::Restore => Ok(client.restore_from_backup().await?.to_json_value()?),
+        ClientNg::Wipe { force } => {
+            if !force {
+                bail!("This will wipe the state of the client irrecoverably. Use `--force` to proceed.")
+            }
+            client.wipe_state().await?;
+            Ok(serde_json::to_value(()).unwrap())
         }
     }
 }
--- a/fedimint-client/src/backup.rs
+++ b/fedimint-client/src/backup.rs
@@ -1,0 +1,421 @@
+use std::cmp::Reverse;
+use std::collections::BTreeMap;
+
+use anyhow::Result;
+use bitcoin::secp256k1;
+use fedimint_core::api::GlobalFederationApi;
+use fedimint_core::core::backup::{BackupRequest, SignedBackupRequest};
+use fedimint_core::core::ModuleInstanceId;
+use fedimint_core::encoding::{Decodable, Encodable};
+use fedimint_core::module::registry::ModuleDecoderRegistry;
+use fedimint_derive_secret::DerivableSecret;
+use fedimint_logging::{LOG_CLIENT, LOG_CLIENT_BACKUP, LOG_CLIENT_RECOVERY};
+use secp256k1_zkp::{KeyPair, Secp256k1};
+use serde::{Deserialize, Serialize};
+use tracing::{debug, info, warn};
+
+use super::Client;
+use crate::secret::DeriveableSecretClientExt;
+
+/// Backup metadata
+///
+/// A backup can have a blob of extra data encoded in it. We provide methods to
+/// use json encoding, but clients are free to use their own encoding.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Encodable, Decodable, Clone)]
+pub struct Metadata(Vec<u8>);
+
+impl Metadata {
+    /// Create empty metadata
+    pub fn empty() -> Self {
+        Self(vec![])
+    }
+
+    pub fn from_raw(bytes: Vec<u8>) -> Self {
+        Self(bytes)
+    }
+
+    pub fn into_raw(self) -> Vec<u8> {
+        self.0
+    }
+
+    /// Is metadata empty
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Create metadata as json from typed `val`
+    pub fn from_json_serialized<T: Serialize>(val: T) -> Self {
+        Self(serde_json::to_vec(&val).expect("serializing to vec can't fail"))
+    }
+
+    /// Attempt to deserialize metadata as typed json
+    pub fn to_json_deserialized<T: serde::de::DeserializeOwned>(&self) -> Result<T> {
+        Ok(serde_json::from_slice(&self.0)?)
+    }
+
+    /// Attempt to deserialize metadata as untyped json (`serde_json::Value`)
+    pub fn to_json_value(&self) -> Result<serde_json::Value> {
+        Ok(serde_json::from_slice(&self.0)?)
+    }
+}
+
+/// Client state backup
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Encodable, Decodable)]
+pub struct ClientBackup {
+    /// Epoch count taken right before taking the backup
+    epoch_count: u64,
+    /// Application metadata
+    metadata: Metadata,
+    /// Module specific-backup (if supported)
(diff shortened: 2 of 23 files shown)


fedimint-client/src/module/mod.rs
```diff
@@ -63,6 +64,27 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
         output: &<Self::Common as ModuleCommon>::Output,
     ) -> TransactionItemAmount;
+
+    fn supports_backup(&self) -> bool {
+        false
+    }
+
+    async fn backup(
+        &self,
+        _dbtx: &mut ModuleDatabaseTransaction<'_>,
```
dpc, line 74: @elsirion 

Does this seem reasonable?

To be able to generically call every module to make a backup, I need to have it as a unified method on the common trait. But then them modules don't seem to have the `db` inside themselves and seem to rely on `*Ext` traits to grab it from the `Client::inner`. So I guess I'll just have to pass it?

A bit unclear to me how to actually get `ModuleDatabaseTransaction<'_>`, but maybe I'm just tired and missed it.

dpc: I don't understand if there's a reason we can't just pass modules api and their isolated db handle in the constructor. This way we wouldn't need `*Ext` traits.

modules/fedimint-mint-client/src/backup.rs
```diff
@@ -0,0 +1,69 @@
...
+        dbtx: &mut ModuleDatabaseTransaction<'_>,
+        api: DynFederationApi,
+        metadata: Metadata,
+    ) -> anyhow::Result<EcashBackup> {
+        // fetch consensus height first - so we dont miss anything when scanning
+        let epoch_count = api.fetch_epoch_count().await?;
+
+        let notes = Self::get_all_spendable_notes(&mut dbtx).await;
+
+        let pending_notes: Vec<_> = dbtx
```
dpc, line 49: How do I get these / how are they named&represented in the new client world? :)

fedimint-client/src/module/mod.rs
dpc: @elsirion  Hmm... I don't think extension traits are worth it if they are there just to "get the default instance id for a given trait". That could be just a method.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -925,6 +969,7 @@ impl State for MintClientStateMachines {
             MintClientStateMachines::Output(issuance_state) => issuance_state.operation_id(),
             MintClientStateMachines::Input(redemption_state) => redemption_state.operation_id(),
             MintClientStateMachines::OOB(oob_state) => oob_state.operation_id(),
+            MintClientStateMachines::Restore(oob_state) => oob_state.operation_id(),
```
elsirion, line 972: It's not an `oob_state`

modules/fedimint-mint-client/src/backup/tracker.rs
```diff
@@ -0,0 +1,716 @@
...
+        epoch_pk: threshold_crypto::PublicKey,
+        secret: DerivableSecret,
+    ) -> Self {
+        assert_eq!(secret.level(), 1);
+        let epoch_range =
+            self.next_epoch..cmp::min(self.next_epoch.wrapping_add(100), self.end_epoch);
+        let mut epoch_stream = Self::fetch_epochs_stream(api, epoch_pk, decoders, epoch_range);
+        while let Some((epoch, epoch_history)) = epoch_stream.next().await {
+            // TODO: should `GlobalClientContext` expose a `TaskHandle` to allow us to react
+            // to shutdown?
```
elsirion, line 179: No, I think the state machine executor would just interrupt the trigger future (needs to be tested).

modules/fedimint-mint-client/src/backup/tracker.rs
```diff
@@ -0,0 +1,716 @@
...
+
+impl MintRestoreFailedState {
+    fn transitions(
+        &self,
+        operation_id: OperationId,
+        context: &MintClientContext,
+        global_context: &DynGlobalClientContext,
+    ) -> Vec<StateTransition<MintRestoreStateMachine>> {
+        vec![]
+    }
```
elsirion, line 667: That's unused, is it?

modules/fedimint-mint-client/src/backup/tracker.rs
```diff
@@ -0,0 +1,716 @@
...
+
+    pub fn handle_input(&mut self, input: &MintInput) {
+        // We attempt to delete any nonce we see as spent, simple
+        for (_amt, note) in input.0.iter_items() {
+            self.spendable_note_by_nonce.remove(&note.0);
+        }
+    }
+
+    pub fn handle_output(&mut self, out_point: OutPoint, output: &MintOutput) {
```
elsirion, line 326: A lot of this logic is copied over, right? Anything I should pay more attention on?

modules/fedimint-mint-client/src/backup.rs
elsirion: We discussed that on a call, just backing up all the output `Created` and  states should suffice.

fedimint-client/src/lib.rs
```diff
@@ -260,8 +270,16 @@ struct ModuleGlobalClientContext {
 
 #[apply(async_trait_maybe_send!)]
 impl IGlobalClientContext for ModuleGlobalClientContext {
-    fn api(&self) -> &(dyn IFederationApi + 'static) {
-        self.client.api.as_ref()
+    fn api(&self) -> DynFederationApi {
+        self.client.api.clone()
+    }
+
+    fn decoders(&self) -> ModuleDecoderRegistry {
```
dpc, line 277: Probably I should just pass `decoders` to recovery method?

fedimint-client/src/lib.rs
```diff
@@ -356,6 +374,10 @@ impl Client {
         self.inner.federation_meta.get(key).cloned()
     }
 
+    fn root_secret(&self) -> DerivableSecret {
```
dpc, line 377: @elsirion Do we already have a way to pass module-specific secret to client module? Should it happen in the constructor (when the module instance is created?).

fedimint-client/src/lib.rs
dpc: @elsirion At some point we've discussed that we should encode inputs/outputs/cis as `Vec<u8>` initially, so code can parse epochs without having to have all decoders. This would be the first place that would benefit from it.

fedimint-client/src/backup.rs
```diff
@@ -0,0 +1,198 @@
...
+    ) -> fedimint_aead::LessSafeKey {
+        fedimint_aead::LessSafeKey::new(
+            secret.child_key(BACKUP_CHILD_ID).to_chacha20_poly1305_key(),
+        )
+    }
+
+    /// Static version of [`Self::get_derived_backup_signing_key`] for testing
+    /// without creating whole `MintClient`
+    fn get_derived_backup_signing_key_static(secret: &DerivableSecret) -> secp256k1_zkp::KeyPair {
+        // TODO: Do we need that one derivation level? This key is already derived for
```
dpc, line 184: TODO: copy&pasted,

fedimint-client/src/lib.rs
```diff
@@ -61,6 +61,12 @@ pub mod sm;
 /// Structs and interfaces to construct Fedimint transactions
 pub mod transaction;
 
+/// Client backup
+pub mod backup;
+
+// TODO: where should this sit?
+pub const BACKUP_CHILD_ID: ChildId = ChildId(1);
```
dpc, line 68: @elsirion What do you think?

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -0,0 +1,757 @@
...
+                        for (amount, note) in finalized.spendable_notes {
+                            let key = NoteKey {
+                                amount,
+                                nonce: note.note.0,
+                            };
+                            dbtx.insert_new_entry(&key, &note).await;
+                        }
+
+                        for (txid, issuance_requests) in finalized.unconfirmed_notes {
+                            // TODO: How do I insert these as new state machines from here?
```
dpc, line 162: @elsirion Please help. Biggest blocker RN.

elsirion: Currently unsupported from inside transitions, we didn't have that use case so far. But since `ModuleGlobalClientContext` has access to the executor via `ClientInner` you should be able to add a method to `IGlobalClientContext` for that.

dpc:
I wish #2470 did all the work needed here. :)

fedimint-client/src/lib.rs
dpc: @elsirion Now we have one backup for all modules.

We should probably have:

* root / TYPE_MODULE / module_id
* root / TYPE_BACKUP 

kind of thing?

fedimint-client/src/lib.rs
dpc: I'm putting `root_secret` into `ClientInner`. I made that comment in a confusing way, as this line of code is about that, no module secrets.

fedimint-client/src/lib.rs
dpc: @elsirion I can't pass `Decoders` in the recovery state itself, because they are not encodable. And I need it in state transition functions to make progress (decode new epochs).

Are you OK with having it here?

fedimint-core/src/backup.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[derive(Debug, Encodable, Decodable)]
+pub struct ClientBackupKeyPrefix;
+
+impl_db_record!(
+    key = ClientBackupKey,
+    value = ClientBackupSnapshot,
+    db_prefix = DbKeyPrefix::ClientBackup,
+);
+impl_db_lookup!(key = ClientBackupKey, query_prefix = ClientBackupKeyPrefix);
```
elsirion, line 21: Doesn't that belong in `fedimint-client`?

elsirion (commented):
One fundamental thing I'm wondering about is if we should allow restoring backups in already initialized clients at all … It seems to me that restoring a backup might be some sort of alternative constructor for a client?

fedimint-core/src/backup.rs
dpc: This is a server side backup. It kind of belongs to `fedimint-server`, but there was some reason why it had to be here.

fedimint-client/src/sm/mod.rs
```diff
@@ -45,3 +46,11 @@ impl GlobalContext for () {}
...
 /// 256bit value collisions are impossible for all intents and purposes.
 pub type OperationId = [u8; 32];
+
+/// Generate random [`OperationId`]
+pub fn random_operation_id() -> OperationId {
+    let mut rng = rand::thread_rng();
+    let mut bytes = [0u8; 32];
+    rng.fill_bytes(&mut bytes);
+    bytes
+}
```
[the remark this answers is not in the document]
dpc: Rebased.

fedimint-cli/src/ng.rs
```diff
@@ -63,15 +47,23 @@ pub enum ClientNg {
...
+    /// Wipe the state of the client (mostly for testing purposes)
+    #[clap(hide = true)]
+    Wipe {
         #[clap(long)]
-        arg: Vec<ffi::OsString>,
+        force: bool,
     },
+    /// Restore the previously created backup of mint notes (with `backup`
+    /// command)
+    Restore,
```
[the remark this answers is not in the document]
dpc: We'll have to tweak things here. We've discussed in a chat, but I think it can be a follow-up.

justinmoon:
Not working for me:

```
fedimint-2 backup-and-restore-02 → fedimint-cli ng info
{
  ...
  "total_msat": 10000000
}

$ fedimint-cli ng wipe --force
2023-05-13T04:42:26.035254Z  INFO fedimint_client::sm::executor: Starting state machine executor task
2023-05-13T04:42:26.035254Z  INFO client: Wiping primary module state module_kind=mint module_id=1
null

$ fedimint-cli ng restore
...
{
  "error": "CliError",
  "kind": "GeneralFailure",
  "message": "failure",
  "raw_error": "EOF while parsing a value at line 1 column 0"
}
```

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -0,0 +1,775 @@
...
+    ) -> Vec<StateTransition<MintRestoreStateMachine>> {
+        let global_context = global_context.clone();
+        let global_context_2 = global_context.clone();
+        let secret = context.secret.clone();
+        let self_clone = self.clone();
+        vec![StateTransition::new(
+            async move {
+                self_clone
+                    .make_progress(
+                        global_context.module_api(),
```
justinmoon, line 145: This is the wrong api. Should be `global_context.api()`. This will send `module_1_fetch_epoch_history` RPC calls. It should send `fetch_epoch_history` RPC calls.

elsirion: Fixed

elsirion: Although it makes me wonder why CI didn't fail …

dpc: There are no tests, and I rebased and fixed merged conflicts, and thought that `module_api` is the same thing I had `api()`.

PR #2407, opened 2023-05-04
feature: improve the dummy module for testing

jkitman:
Improved some of the issues with the dummy module:

- Printing money is now a tx so it can output to any client primary module (e.g. for testing ecash) and will be faster
- The signature shares now sign a returned message to demonstrate consensus signing better

### DIFF
--- a/modules/fedimint-dummy-client/src/api.rs
+++ b/modules/fedimint-dummy-client/src/api.rs
@@ -1,15 +1,14 @@
 use fedimint_core::api::{FederationApiExt, FederationResult, IFederationApi};
+use fedimint_core::epoch::SerdeSignature;
 use fedimint_core::module::ApiRequestErased;
 use fedimint_core::task::{MaybeSend, MaybeSync};
-use fedimint_core::{apply, async_trait_maybe_send, Amount};
-use fedimint_dummy_common::DummyPrintMoneyRequest;
-use secp256k1::XOnlyPublicKey;
+use fedimint_core::{apply, async_trait_maybe_send};
 
 #[apply(async_trait_maybe_send!)]
 pub trait DummyFederationApi {
-    async fn print_money(&self, request: DummyPrintMoneyRequest) -> FederationResult<()>;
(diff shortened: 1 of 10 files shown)


modules/fedimint-dummy-client/src/lib.rs
```diff
@@ -46,28 +47,45 @@ pub trait DummyClientExt {
...
+            keys: vec![fed_key_pair()],
+            state_machines: Arc::new(move |_, _| Vec::<DummyStateMachine>::new()),
+        };
+
+        // Build and send tx to the fed
+        // Will output to our primary client module
+        let tx = TransactionBuilder::new().with_input(input.into_dyn(id));
+        let outpoint = |txid| OutPoint { txid, out_idx: 0 };
+        let txid = self
+            .finalize_and_submit_transaction(op_id, DummyCommonGen::KIND.as_str(), outpoint, tx)
```
m1sterc001guy, line 85: Nit: I think you can use `claim_input` here

modules/fedimint-dummy-client/src/api.rs
```diff
@@ -1,29 +1,28 @@
...
 use fedimint_core::task::{MaybeSend, MaybeSync};
-use fedimint_core::{apply, async_trait_maybe_send, Amount};
-use fedimint_dummy_common::DummyPrintMoneyRequest;
-use secp256k1::XOnlyPublicKey;
+use fedimint_core::{apply, async_trait_maybe_send};
 
 #[apply(async_trait_maybe_send!)]
 pub trait DummyFederationApi {
-    async fn print_money(&self, request: DummyPrintMoneyRequest) -> FederationResult<()>;
+    async fn sign_message(&self, message: String) -> FederationResult<()>;
```
m1sterc001guy, line 9: Does this module still intend to print money?

jkitman: Yes, we print money by allowing users to submit txs with the fed's private key.

This was a change so that we can print money to another primary module (e.g. ecash).

PR #2408, opened 2023-05-04
refactor: Merge `status` and `consensus_status` endpoints

douglaz:
https://github.com/fedimint/fedimint/pull/2375 original goal was to add all the information to a single `status` endpoint, but at the time no good solution was found on how to merge both data. While creating the CLI command this simple and obvious solution appeared:
```rust
pub struct StatusResponse {
    pub server: ServerStatus,
    pub consensus: Option<ConsensusStatus>,
}
```

For instance in json:
```json
{
    "consensus": {
      "last_contribution": 14,
      "peers_flagged": 0,
      "peers_offline": 0,
      "peers_online": 3,
      "status_by_peer": {
        "1": {
          "flagged": false,
          "last_contribution": 14,
          "last_contribution_timestamp_seconds": 1683215023
        },
        "2": {
          "flagged": false,
          "last_contribution": 14,
          "last_contribution_timestamp_seconds": 1683215023
        },
        "3": {
          "flagged": false,
          "last_contribution": 14,
          "last_contribution_timestamp_seconds": 1683215023
        }
      }
    },
    "server": "ConsensusRunning"
  }
```

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -863,7 +863,7 @@ fn url_to_string_with_default_port(url: &Url) -> String {
 impl<C: JsonRpcClient> WsFederationApi<C> {}
 
 /// The status of a server, including how it views its peers
-#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ConsensusStatus {
     /// The last contribution that this server has made to the consensus, it's
     /// equivalent to [`PeerConsensusStatus::last_contribution`] and
@@ -903,6 +903,32 @@ pub struct ConsensusContribution {
     pub value: u64,
     /// When the contribution was received
     pub time: SystemTime,
+}
+
+/// The state of the server returned via APIs
+#[derive(Debug, Clone, Default, Serialize, Deserialize, Eq, PartialEq)]
+pub enum ServerStatus {
+    /// Server needs a password to read configs
+    #[default]
+    AwaitingPassword,
+    /// Waiting for peers to share the config gen params
+    SharingConfigGenParams,
+    /// Ready to run config gen once all peers are ready
+    ReadyForConfigGen,
+    /// We failed running config gen
+    ConfigGenFailed,
+    /// Config is generated, peers should verify the config
+    VerifyingConfigs,
+    /// Restarted from a planned upgrade (requires action to start)
+    Upgrading,
+    /// Consensus is running
+    ConsensusRunning,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct StatusResponse {
+    pub server: ServerStatus,
+    pub consensus: Option<ConsensusStatus>,
 }
 
 #[cfg(test)]
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -9,8 +9,9 @@ use bitcoin_hashes::{sha256, Hash};
 use fedimint_aead::random_salt;
 use fedimint_core::admin_client::{
     ConfigGenConnectionsRequest, ConfigGenParamsConsensus, ConfigGenParamsRequest,
-    PeerServerParams, ServerStatus, WsAdminClient,
+    PeerServerParams, WsAdminClient,
 };
+use fedimint_core::api::{ServerStatus, StatusResponse};
 use fedimint_core::config::ServerModuleGenRegistry;
 use fedimint_core::core::ModuleInstanceId;
 use fedimint_core::db::Database;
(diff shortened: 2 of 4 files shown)


fedimint-server/src/config/api.rs
```diff
@@ -573,12 +574,13 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConfigGenApi>> {
...
-                if context.has_auth() {
-                    Ok(config.status().await)
-                } else {
-                    Err(ApiError::unauthorized())
-                }
+            async |config: &ConfigGenApi, _context, _v: ()| -> StatusResponse {
+                let status = config.status().await;
+                assert_ne!(status, ServerStatus::ConsensusRunning);
+                Ok(StatusResponse {
+                    status,
```
justinmoon, line 581: Maybe this key should be `config_status` instead of just `status`?

douglaz: I think a more long name would something like `{server,general,overall}_status`. It's not specific to config. For instance in future it could say that we are in a `shutdown`/`terminating` state.

But I would like to hear @jkitman  thoughts on that.

jkitman: I would say that status isn't necessarily specific to config gen, just most of the current states are.

For instance we have a `ServerStatus::Upgrade` and `ServerStatus::ConsensusRunning` status.

The only issue with sticking the `Option<ConsensusStatus>` into this struct, is we might eventually need a `Option<UpgradeStatus>` or many other options.  But since this is mostly consumed by the UI, I'll leave it up to @justinmoon to decide whether this seems alright.

douglaz: I think it makes perfect sense to add a new field with a new specific struct when there is more information about a particular status.

fedimint-server/src/config/api.rs
```diff
@@ -573,12 +574,13 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConfigGenApi>> {
...
             "status",
-            async |config: &ConfigGenApi, context, _v: ()| -> ServerStatus {
-                if context.has_auth() {
-                    Ok(config.status().await)
-                } else {
-                    Err(ApiError::unauthorized())
-                }
+            async |config: &ConfigGenApi, _context, _v: ()| -> StatusResponse {
+                let status = config.status().await;
+                assert_ne!(status, ServerStatus::ConsensusRunning);
```
jkitman, line 579: I would not put asserts into production code

fedimint-core/src/api.rs
```diff
@@ -905,6 +905,32 @@ pub struct ConsensusContribution {
...
+    VerifyingConfigs,
+    /// Restarted from a planned upgrade (requires action to start)
+    Upgrading,
+    /// Consensus is running
+    ConsensusRunning,
+}
+
+#[derive(Debug, Serialize, PartialEq, Eq)]
+pub struct StatusResponse {
+    pub status: ServerStatus,
```
dpc, line 930: The whole struct/endpoint is called "status", so it doesn't seem necessary  to repeat that word?

`server: ServerStatus` and `consensus:Option<ConsensusStatus>`?

fedimint-server/src/config/api.rs
justinmoon: I'm good with `status`

douglaz: I think @dpc's suggestion make sense, use `server` for `ServerStatus` and `consensus` for `ConsensusStatus`

fedimint-core/src/api.rs
```diff
@@ -905,6 +905,32 @@ pub struct ConsensusContribution {
...
+    Upgrading,
+    /// Consensus is running
+    ConsensusRunning,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct StatusResponse {
+    pub server: ServerStatus,
+    pub consensus: Option<ConsensusStatus>,
+}
```
elsirion, line 932: Nit: Would it make sense to eventually merge these two structs? E.g. `ConsensusStatus` could be a member of `ServerStatus::ConsensusRunning`. Other states could also carry additional info (e.g. error messages for config gen failed).

fedimint-server/src/config/api.rs
```diff
@@ -573,12 +574,12 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConfigGenApi>> {
...
-                    Ok(config.status().await)
-                } else {
-                    Err(ApiError::unauthorized())
-                }
+            async |config: &ConfigGenApi, _context, _v: ()| -> StatusResponse {
+                let server = config.server_status().await;
+                Ok(StatusResponse {
+                    server,
+                    consensus: None
+                })
```
elsirion, line 582: I think we'll want to reconsider making this a public endpoint eventually, but ok for now.

elsirion (approved):

fedimint-core/src/api.rs
douglaz: This was tried and the resulting serialized json is very hard to parse. I works if you're only using rust, but if you are interoperating with other system you need to simplify the result.

PR #2410, opened 2023-05-04
feat(client-ng): fetch active operations

elsirion:
Allows clients to subscribe to all operations where progress is expected.


m1sterc001guy (approved):
LGTM, looks like there might be a flakey bitcoind test though

PR #2412, opened 2023-05-04
feature: parallelize DKG

### LINKED ISSUE #2313: Parallelize DKG
Currently module DKGs are run consecutively. Since their messages are tagged with their instance IDs anyway we could run them in parallel.

jkitman:
Fixes #2313

Some benchmarks show it helps when we simulate network latency:
```
Before parallelization
3.2 sec with 0 ms network delay, 
18.5 sec with 10 ms network delay, 
78.4 sec with 50 ms network delay

After parallelization
3.8 sec with 0 ms network delay, 
5.8 sec with 10 ms network delay, 
28.5 sec with 50 ms network delay
```


dpc (approved):

fedimint-server/src/multiplexed.rs
```diff
@@ -134,28 +135,33 @@ where
                 // since we've last checked.
                 drop(out_of_order);
 
-                let (peer, new_msg) = connections.receive().await?;
-                if key == new_msg.key {
-                    return Ok((peer, new_msg.msg));
-                }
+                // Avoid holding the connections lock which prevents sends
+                // FIXME: should probably spawn a task for receiving instead
+                if let Some(Ok((peer, new_msg))) = connections.receive().now_or_never() {
```
dpc, line 140: This is kind of a hack.

Architecturally the problem is that we're putting "sending" and "receiving" on the same object, which gets locked by the same lock. Before parallel DKG change I guess nowhere in the code anything expected that sending and receiving can be happening at the same time.

This is similar to e.g. sending and receiving on IO, which then get split by https://docs.rs/tokio/latest/tokio/io/fn.split.html , because they are in fact independent and happen in parallel, so should not be locked by the same thing.


The:

```
pub trait IPeerConnections<Msg>
where
    Msg: Serialize + DeserializeOwned + Unpin + Send,
{
```

should become two traits and the consensus code should hold two separate handles -one for sending, one for receiving. Under the hood we can keep a single IO thread handling each peer.

Alternatively, we could just make `send` and `recv` on `IPeerConnections` take `&self` instead of `&mut` so it wouldn't be necessary to take a lock outside of it, and turn:

```
pub struct ReconnectPeerConnections<T> {
    connections: HashMap<PeerId, PeerConnection<T>>,
}


struct PeerConnection<T> {
    outgoing: Sender<T>,
    incoming: Receiver<T>,
}
```

into:

pub struct ReconnectPeerConnections<T> {
    senders: Mutex<HashMap<PeerId, Sender<T>>>,   
    receivers: Mutex<HashMap<PeerId, Receiver<T>>>,
}
```

so we can lock sending and receiving part independently.

@elsirion what do you think?

Comment (elsirion) [reply fedimint-server/src/multiplexed.rs:140]: I don't think we need mutexes even. There could be a fn `PeerConnections::split(&'a mut self) -> (ReadHalf<'a>, WriteHalf<'a>)` that separately borrows the parts of the connections struct. The impl for `ReconnectPeerConnections` could return a `HashMap<PeerId, &Sender<T>>` and `HashMap<PeerId, &mut Receiver<T>` for read/write respectively and impl send/receive on these.

Comment (dpc) [reply fedimint-server/src/multiplexed.rs:140]: @elsirion Something tells me this is going to difficult to implement and work with due to everything being borrowed, lifetimes, etc.

Right now, since methods on `IPeerConnections` take `&mut self`, they force wrapping in `Arc<Mutex<...>>` if anyone wants to use `IPeerConnections` concurrently (which we now do if we want parallelized DKG). So we effectively do have a `Mutex` already, just on the user's side.

Putting this `Mutex` inside just takes the responsibility of locking from the user to the implementation. And splitting it into two, just enables parallel write and read.

Comment (dpc) [reply fedimint-server/src/multiplexed.rs:140]: The reason IO goes with `split` (despite much worse ergonomics) must be that operations on each end (read / write) don't make much sense with just `&self`, as they are inherently sequential, so no one should want `fn read(&self, &mut [u8])` that they can call in parallel. Not only it makes implementing such interfaces harder/slower, but it's just not useful for the caller. What the caller wants is `fn read(&mut self)` and `fn write(&mut self)`... just on a separate object.

But on peer connection, it does make sense as each message is in itself atomic. Multiple parallel `send`s just doesn't guarantee ordering (which the receiver might just not care about anyway), while multiple parallel `recv`... well.. technically works (again - only loosing ordering), though doesn't seem all that useful functionally in practice, I guess.

I guess `IPeerConnections::split` could be OK, under the condition that that it is not borrowing, but taking `self` by move.  

But then... why even bother with splitting. Just start with two separate objects to begin with, like I mentioned here:

> should become two traits and the consensus code should hold two separate handles -one for sending, one for receiving.

Comment (jkitman) [reply fedimint-server/src/multiplexed.rs]: I looked into splitting, there's not a very clean way to do it without splitting `ReconnectPeerConnections`.

Instead I removed all the `Mutex` and `sleep` and replaced it with channels, which greatly improved performance (see updated metrics in my first comment).

Comment (dpc) [reply fedimint-server/src/multiplexed.rs]: > I looked into splitting, there's not a very clean way to do it without splitting ReconnectPeerConnections.

We should split it, yes.

> Instead I removed all the Mutex and sleep and replaced it with channels, which greatly improved performance in some cases.

Looks like quite a bit overhead, turning it into callback-based rpc system. Not excited about it, though I guess it works. Intuition tells me that 1000-sized bounded channels might lead to deadlocks, through probably only theoretically.

That 10ms sleep there leads to unnecessary delays, especially now that the contention in that place happens more oftent. Thats a good point/find.

File: fedimint-server/src/multiplexed.rs [hunk 2]
```diff
@@ -66,20 +59,99 @@ struct ModuleMultiplexerInner<MuxKey, Msg> {
...
+                    connections.ban_peer(peer).await;
+                }
+                // Receive callbacks are added to callback queue by key
+                receive_callback = receive_callbacks_rx.recv() => {
+                    let (key, callback) = receive_callback.ok_or(Cancelled)?;
+                    out_of_order.callbacks.entry(key.clone()).or_default().push_back(callback);
+                    key_inserted = Some(key);
+                }
+                // Actual received messages are added message queue by key
+                receive = connections.receive() => {
```
dpc, line 124: Since `connections.receive()` is blocking, this treat will do absolutely nothing until next message from the peer is received. This includes sending any outgoing messages to that peer. This can lead to subtle hangs.

fedimint-server/src/multiplexed.rs
dpc: > That 10ms sleep there leads to unnecessary delays, especially now that the contention in that place happens more oftent. Thats a good point/find.

I've put a comment there, that it ideally should be a CondVar-based wait.

fedimint-server/src/multiplexed.rs
dpc, line 124: I see. So this is a part of a select now.

fedimint-server/src/multiplexed.rs
```diff
@@ -66,20 +59,99 @@ struct ModuleMultiplexerInner<MuxKey, Msg> {
...
+        mut out_of_order: ModuleMultiplexerOutOfOrder<MuxKey, Msg>,
+        mut send_requests_rx: Receiver<(Vec<PeerId>, MuxKey, Msg)>,
+        mut receive_callbacks_rx: Receiver<Callback<MuxKey, Msg>>,
+        mut peer_bans_rx: Receiver<PeerId>,
+    ) -> Cancellable<()> {
+        loop {
+            let mut key_inserted: Option<MuxKey> = None;
+            tokio::select! {
+                 // Send requests are forwarded to underlying connections
+                 send_request = send_requests_rx.recv() => {
```
dpc, line 108: User sends the send request over a channel, just so this thread receives it, and then turns over and calls `connections.send` which under the hood puts it on another channel..

fedimint-server/src/multiplexed.rs
dpc: I just realized that this is going to be prone to loosing messages.


https://docs.rs/tokio/latest/tokio/macro.select.html :

> Waits on multiple concurrent branches, returning when the first branch completes, cancelling the remaining branches.

If somewhere inside receiver, between when the messages was actually received and returning the value there's any yield point (lock or whatever), and then a message arrives on any of these channels here, that already-received-the-message `connections.receive()` will get canceled and drop the message.

Even if that's not the case already, it might easily be introduced in the future, leading to weirdly lost peer messages.


I really don't like async/await Rust ... :D

dpc: To fix that, I guess each `connections.receive` would have to be started as a separate task, returning the message via channel, that we can await here. Bleh. :D

fedimint-server/src/multiplexed.rs
elsirion: > if anyone wants to use `IPeerConnections` concurrently

But nobody is supposed to use the read part of peer connections concurrently, this will always lead to weird bugs imo. I think we want a solution for allowing parallel writes though. But it sounds like kitman found another solution, taking a look at that.

elsirion (approved):
I wrote with @jkitman, this PR seems to achieve what it set out to achieve. At some point the whole DKG and multiplexer probably needs a redesign, but that would be too much scope-ceep here.

fedimint-server/src/multiplexed.rs
dpc: > It's the first time I hear such strong aversion to async rust,

@elsirion 

Not exactly uncommon in Rust ecosystem, though we're probably a minority.

With normal Rust it's "if it compiles, it works". With async Rust we're back to "pay close attention to" (during code reviews), which experience shows is not a viable long term strategy for building robust software.

Not to mention all the `Box::pin`, extra generics, extra lifetimes all over the place.

PR #2416, opened 2023-05-05
Client-NG: Deposits

elsirion:
* Minor API fixes
* Deposit state machine
* Deposit method in extension traits

### DIFF
--- a/modules/fedimint-wallet-client/src/deposit.rs
+++ b/modules/fedimint-wallet-client/src/deposit.rs
@@ -1,0 +1,373 @@
+use std::sync::Arc;
+use std::time::{Duration, SystemTime};
+
+use anyhow::anyhow;
+use bitcoin::{Script, Txid};
+use fedimint_client::sm::{ClientSMDatabaseTransaction, OperationId, State, StateTransition};
+use fedimint_client::transaction::ClientInput;
+use fedimint_client::DynGlobalClientContext;
+use fedimint_core::core::LEGACY_HARDCODED_INSTANCE_ID_WALLET;
+use fedimint_core::encoding::{Decodable, Encodable};
+use fedimint_core::task::sleep;
+use fedimint_core::txoproof::TxOutProof;
+use fedimint_core::{OutPoint, TransactionId};
+use fedimint_wallet_common::tweakable::Tweakable;
+use fedimint_wallet_common::txoproof::PegInProof;
+use fedimint_wallet_common::WalletInput;
+use miniscript::ToPublicKey;
+use secp256k1::KeyPair;
+use tracing::{trace, warn};
+
+use crate::api::WalletFederationApi;
+use crate::{WalletClientContext, WalletClientStates};
+
+const TRANSACTION_STATUS_FETCH_INTERVAL: Duration = Duration::from_secs(10);
+
+// FIXME: deal with RBF
+// FIXME: deal with multiple deposits
+#[aquamarine::aquamarine]
+/// The state machine driving forward a deposit (aka peg-in).
+///
+/// ```mermaid
+/// graph LR
+///     Created -- Transaction seen --> AwaitingConfirmations["Waiting for confirmations"]
+///     AwaitingConfirmations -- Confirmations received --> Claiming
+///     AwaitingConfirmations -- "Retransmit seen tx (planned)" --> AwaitingConfirmations
+///     Created -- "No transactions seen for [time]" --> Timeout["Timed out"]
+/// ```
+#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
+pub struct DepositStateMachine {
+    pub(crate) operation_id: OperationId,
+    pub(crate) state: DepositStates,
+}
+
+impl State for DepositStateMachine {
+    type ModuleContext = WalletClientContext;
+    type GlobalContext = DynGlobalClientContext;
+
+    fn transitions(
+        &self,
+        context: &Self::ModuleContext,
+        global_context: &Self::GlobalContext,
+    ) -> Vec<StateTransition<Self>> {
+        match &self.state {
+            DepositStates::Created(created_state) => {
+                vec![
+                    StateTransition::new(
+                        await_created_btc_transaction_submitted(
+                            context.clone(),
+                            created_state.tweak_key,
+                        ),
+                        |_db, (btc_tx, out_idx), old_state| {
+                            Box::pin(transition_tx_seen(old_state, btc_tx, out_idx))
+                        },
+                    ),
+                    StateTransition::new(
+                        await_deposit_address_timeout(created_state.timeout_at),
+                        |_db, (), old_state| Box::pin(transition_deposit_timeout(old_state)),
+                    ),
+                ]
+            }
+            DepositStates::WaitingForConfirmations(waiting_state) => {
+                let global_context = global_context.clone();
+                vec![StateTransition::new(
+                    await_btc_transaction_confirmed(
+                        context.clone(),
+                        global_context.clone(),
+                        waiting_state.clone(),
+                    ),
+                    move |dbtx, txout_proof, old_state| {
+                        Box::pin(transition_btc_tx_confirmed(
+                            dbtx,
+                            global_context.clone(),
+                            old_state,
+                            txout_proof,
+                        ))
+                    },
+                )]
+            }
+            DepositStates::Claiming(_) => {
+                vec![]
+            }
+            DepositStates::TimedOut(_) => {
+                vec![]
+            }
+        }
+    }
+
+    fn operation_id(&self) -> OperationId {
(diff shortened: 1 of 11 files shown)


modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -0,0 +1,368 @@
...
+    fn transitions(
+        &self,
+        context: &Self::ModuleContext,
+        global_context: &Self::GlobalContext,
+    ) -> Vec<StateTransition<Self>> {
+        match &self.state {
+            DepositStates::Created(created_state) => {
+                vec![
+                    StateTransition::new(
+                        await_crated_btc_transaction_submitted(
```
m1sterc001guy, line 55: Nit: typi "crated"

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        sleep(Duration::from_secs(10)).await;
+    }
+}
+
+async fn get_script_history(
+    esplora_server: &str,
+    script: &Script,
+) -> anyhow::Result<Vec<bitcoin::Transaction>> {
+    let esplora_client = esplora_client(esplora_server)?;
```
m1sterc001guy, line 149: Does this only support esplora for now?

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -0,0 +1,368 @@
...
+    old_state: DepositStateMachine,
+    btc_transaction: bitcoin::Transaction,
+    out_idx: u32,
+) -> DepositStateMachine {
+    let DepositStateMachine {
+        operation_id,
+        state: old_state,
+    } = old_state;
+
+    match old_state {
```
m1sterc001guy, line 172: Nit: I thought the linter would yell at you if a match statement is used with only one option (instead of using `if let`

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -0,0 +1,368 @@
...
+                tweak_key: created_state.tweak_key,
+                btc_transaction,
+                out_idx,
+            }),
+        },
+        state => panic!("Invalid previous state: {state:?}"),
+    }
+}
+
+async fn await_deposit_address_timeout(timeout_at: SystemTime) {
```
m1sterc001guy, line 185: Is the timeout here just a conservation of resources? Because a transaction could be broadcasted after this timeout period right?

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -0,0 +1,368 @@
...
+async fn await_btc_transaction_confirmed(
+    context: WalletClientContext,
+    global_context: DynGlobalClientContext,
+    waiting_state: WaitingForConfirmationsDepositState,
+) -> TxOutProof {
+    loop {
+        // Wait for confirmation
+        let consensus_height = match global_context
+            .api()
+            .with_module(LEGACY_HARDCODED_INSTANCE_ID_WALLET)
```
m1sterc001guy, line 212: To get rid of this would we need to pass in the module instance id from the `WalletClient`?

modules/fedimint-wallet-client/src/deposit.rs
elsirion: Oh, yeah, and I just remembered I need to remove the hard-coding …

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -1,28 +1,200 @@
...
+            };
+            yield DepositState::Confirmed;
+
+            match tx_subscriber
+                .await_tx_accepted(claiming.transaction_id)
+                .await
+            {
+                Ok(_) => {
+                    // FIXME: await claiming ecash and avoid this random delay
+                    sleep(Duration::from_secs(4)).await;
```
m1sterc001guy, line 180: I think I'll have a fix for this once I submit a PR for this

m1sterc001guy (approved):
LGTM, I'm ok with merging this an iterating on the TODOs

modules/fedimint-wallet-client/src/deposit.rs
elsirion: I actually use the unexpected state for logging, so it can't be a if-let

modules/fedimint-wallet-client/src/deposit.rs
elsirion: Yes, the timeout should be chosen excessively high but we aren't good at dealing with never-terminating state machines right now.

modules/fedimint-wallet-client/src/deposit.rs
elsirion: It's probably better to give the LN module its own API endpoint longer term.

modules/fedimint-wallet-client/src/lib.rs
elsirion: Fixed

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -0,0 +1,373 @@
...
+            Ok(confirmation_height) => confirmation_height,
+            Err(e) => {
+                warn!("Failed to fetch confirmation height: {e}");
+                sleep(TRANSACTION_STATUS_FETCH_INTERVAL).await;
+                continue;
+            }
+        };
+
+        if !confirmation_height
+            .map(|confirmation_height| consensus_height > confirmation_height)
```
justinmoon, line 242: What about finality delay?

modules/fedimint-wallet-client/src/deposit.rs
m1sterc001guy: Maybe we should log if the timeout ever actually does occur

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -122,6 +124,8 @@ impl ServerModuleGen for WalletGen {
                     params.consensus.network,
                     params.consensus.finality_delay,
                     params.local.bitcoin_rpc.clone(),
+                    // TODO: make configurable
```
justinmoon, line 127: It seems this is now configurable by environment variable?

elsirion: What do you mean by that? I don't think I exposed that anywhere so far since I see including this in the server cfg as a temporary fix only. We need some modularized way of letting clients define their own local config that they don't receive from the server imo.

modules/fedimint-wallet-client/src/deposit.rs
justinmoon: Ahh I see. This will just loop until finally the `consensus_height > confirmation_height`. First time after the transaction has been confirmed we'll likely `continue` here because we'll still be waiting on finality delay many confirmations. So it's this looping construct that actually does the job.

modules/fedimint-wallet-server/src/lib.rs
justinmoon: Yea nevermind. `default_esplora_server` just sets default, doesn't read them from environment.

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -0,0 +1,373 @@
...
+                out_idx,
+            }),
+        },
+        state => panic!("Invalid previous state: {state:?}"),
+    }
+}
+
+async fn await_deposit_address_timeout(timeout_at: SystemTime) {
+    if let Ok(time_until_deadline) = timeout_at.duration_since(fedimint_core::time::now()) {
+        tokio::time::sleep(time_until_deadline).await;
```
justinmoon, line 189: This will break wasm. Use `fedimint_core::task::sleep` instead.

modules/fedimint-wallet-client/src/deposit.rs
elsirion: Yes, far from elegant, but should work imo. Ideally I'd like to use electrum long term, which afaik supports subscriptions, but we need a way to convert the txout proofs it gives us to the ones that core and fedimint use.

modules/fedimint-wallet-client/src/deposit.rs
elsirion: Done. Wonder if we could just use semgrep to ban it …

modules/fedimint-wallet-server/src/lib.rs
elsirion: *If* we want to keep it this way it should be a module config gen parameter.

justinmoon (approved):

modules/fedimint-wallet-client/src/deposit.rs
justinmoon: https://github.com/fedimint/fedimint/pull/2564

Issue #2417, opened 2023-05-05
Client-NG cannot reissue after spend


m1sterc001guy:
Reproduce steps:

First get some ecash notes from the legacy client
`fedimint-cli spend 420`
`fedimint-cli ng reissue <note>`

Then spend the ecash notes from the new client
`fedimint-cli ng spend 42`

Then try to get new notes from the legacy client
`fedimint-cli spend 420`
`fedimint-cli ng reissue <note>`

Looks like a decoding issue is happening in `find_by_prefix`

![image](https://user-images.githubusercontent.com/1859166/236505230-3893bf23-10d1-4638-b041-9f35a017519f.png)

elsirion:
I activated more logging using `` to get the following key that could not be read:

```
a1 = active state
de310462fc8ce5780bfb2205b089d7985ad736cdc9265e6e043deda981ac6c17 = operation id
0001 = mint instance
<missing 8 bytes the error is about> = enum variant of MintClientStateMachines
# Begin of MintInputStateMachine
de310462fc8ce5780bfb2205b089d7985ad736cdc9265e6e043deda981ac6c17 = operation id
e48ea0c1a520345ff0397a41dadaec91e5220890b6d746eb4dfbf5af5d9fdc8e 0000000000000000 = FM txid + input idx
# Begin tiered multi
0000000000000000
0000000000000003
0000000000000002
0000000000000001
d19365156b9c47bae1c7fc871ff6f9298cf7c0c3be52f6ddd92d9910d4928ce2b6190f0ca068ea3fd144efde14eecffd52838ef4b7a66b2b1f001ef3ada49a836f4176abdc8fa998facc000aa7a6ea7315cda7b04de733bb464006f8b6c2a79e6f39f1865844af89e8fb7812c82482b6 = Spendable Note (32b nonce+48b sig+32b spend key)
0000000000000008
0000000000000001
b18aad0a0fedfd30ade28ed5e09f150ba6b9d84eadc5a3892da05a258eba9aba99f27fc1f5ec52bdb3011a2e1423971b04478b08ff9d9e5d60342a50649abec0dd36f73f38ae472fd5055f16b126fb9267cd69113b461a383b8fcfcc9a8cb42909cb2b051a9b4dbeb6b47fc6a1c19061
0000000000000020
0000000000000001
[...]

PR #2422, opened 2023-05-05
fix: Client NG spend persists invalid state machine

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2417

Issue is that `try_cancel_oob_spend` was only creating `MintInputStateMachine` and not `MintClientStateMachines::Input(MintInputStateMachine)`


elsirion (approved):

elsirion:
Great find, thx a lot!

PR #2423, opened 2023-05-06
Improvements to the config APIs

jkitman:
Based on UI feedback:
- Add current_id to “get_consensus_config_params”
- Make “get_verify_config_hash” available after start_consensus
- New “connection_code” endpoint, passed into “config” endpoint
- Write password to disk once configs are verified

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -224,7 +224,12 @@ impl ConfigGenApi {
     /// with other peers
     pub fn get_verify_config_hash(&self) -> ApiResult<BTreeMap<PeerId, sha256::Hash>> {
         let state = self.require_status(ServerStatus::VerifyingConfigs)?;
-        state.get_verification()
+        let config = state
+            .config
+            .clone()
+            .ok_or(ApiError::bad_request("Missing config".to_string()))?;
+
+        Ok(get_verification_hashes(&config))
     }
 
     /// Writes the configs to disk after they are generated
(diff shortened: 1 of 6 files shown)


fedimint-server/src/config/api.rs
```diff
@@ -447,29 +452,23 @@ impl ConfigGenState {
...
+            .consensus
+            .consensus_encode(&mut engine)
+            .expect("hashes");
+        cert.consensus_encode(&mut engine).expect("hashes");
 
-            let hash = sha256::Hash::from_engine(engine);
-            hashes.insert(*peer, hash);
-        }
-        Ok(hashes)
+        let hash = sha256::Hash::from_engine(engine);
```
elsirion, line 470: Side-note on efficient hashing: by cloning the hash engine after feeding the config into it but before encoding the particular peer's cert, we could save on hashing and encoding (not encode the config in the loop but outside).

fedimint-server/src/config/api.rs
```diff
@@ -235,8 +235,10 @@ impl ConfigGenApi {
     /// Writes the configs to disk after they are generated
     fn write_configs(&self, config: &ServerConfig, state: &ConfigGenState) -> ApiResult<()> {
         let auth = config.private.api_auth.0.clone();
-        write_new(self.data_dir.join(SALT_FILE), random_salt())
-            .map_err(|e| ApiError::server_error(format!("Unable to write to data dir {e:?}")))?;
+        let io_error = |e| ApiError::server_error(format!("Unable to write to data dir {e:?}"));
+        // TODO: Make writing password optional
+        write_new(self.data_dir.join(PLAINTEXT_PASSWORD), &auth).map_err(io_error)?;
```
elsirion, line 240: Should we track this in an issue or will it be a follow-up soonish?

elsirion (approved):

fedimint-server/src/config/api.rs
jkitman: Yes I'll create an issue

PR #2430, opened 2023-05-07
Guardian UI v0

justinmoon:
Currently this UI is able to run a federation, but the last screen will display an error message 
image

But the federation does actually run.


justinmoon:
Closing in favor of https://github.com/fedimint/fedimint/pull/2360

PR #2434, opened 2023-05-08
feat: tokio::time::sleep -> fedimint_core::task::sleep in Client NG

justinmoon:
`tokio::time::sleep` doesn't work in WASM

### DIFF
--- a/fedimint-client/src/sm/executor.rs
+++ b/fedimint-client/src/sm/executor.rs
@@ -288,7 +288,7 @@ where
         if active_states.is_empty() {
             // FIXME: what to do in this case? Probably best to subscribe to DB eventually
             debug!("No state transitions available, waiting before re-trying");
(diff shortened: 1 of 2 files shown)


dpc:
Ban via semgrep as well (`.semgrep.all.yaml`)?

dpc (approved):

elsirion:
I wonder why CI didn't catch it, aren't we building client ng for wasm? I thought we do at least in parts …

dpc:
@elsirion 

`nix/craneBuild.nix` we're only building `fedimint-client-legacy`:

```
  client-pkgs = { target ? null }: self.pkgsCrossBuild {
    name = "client-pkgs";
    inherit target;

    pkgs = {
      fedimint-client-legacy = { };
    } // lib.optionalAttrs (target == null || target.name != "wasm32-unknown-unknown") {
      # broken on wasm32
      fedimint-sqlite = { };
    };
  };
```

PR #2446, opened 2023-05-08
feat: Add status command to cli

douglaz:
A follow up of https://github.com/fedimint/fedimint/pull/2375 and https://github.com/fedimint/fedimint/pull/2408


devimint/src/vars.rs
```diff
@@ -93,6 +93,11 @@ declare_vars! {
         FM_ESPLORA_DIR: PathBuf = mkdir(FM_TEST_DIR.join("esplora")).await?;
         FM_READY_FILE: PathBuf = FM_TEST_DIR.join("ready");
 
+        // variables required by the admin client
```
dpc, line 96: There's bunch of a things in this PR that seem unrelated (new env vars, passing `WsAdminClient`). Is this intentional?

justinmoon: We shouldn't hard-code `FM_OUR_ID`

douglaz: I was trying the hard way, which is to add to cli ng, but let's try now the easy way, which is much simpler

douglaz:
Closed in favor of https://github.com/fedimint/fedimint/pull/2446

PR #2454, opened 2023-05-08
feature: default no download limit on configs

jkitman:


justinmoon:
I spawned tmuxinator (needed to cherry-pick [this commit](https://github.com/fedimint/fedimint/pull/2458/commits) for it to work)

```
$  cat $FM_DATA_DIR/client-connect
fed113f2k62wydr2hd5kg2llcxs7xalqr607gerzzjszfw0n23f2rju0kndnsv92y743tpwkx4xsddh7wvqq4waen5te0xyerwt3s9cczuvf6xyurzde59adx7mzylmmcw3v3sne3xmgyqju
$ fedimint-cli decode-connect-info fed113f2k62wydr2hd5kg2llcxs7xalqr607gerzzjszfw0n23f2rju0kndnsv92y743tpwkx4xsddh7wvqq4waen5te0xyerwt3s9cczuvf6xyurzde59adx7mzylmmcw3v3sne3xmgyqju
{
  "url": "ws://127.0.0.1:18174/",
  "download_token": "5a6f6c44fef787459184f313",
  "id": "8a556d29c468d576d2c857ff8343c6efc03d3fc8c8c429404973e6a8a543971f69b67061544f562b0bac6a9a0d6dfce6"
}
```

Did we stop encoding multiple servers in the connection string? I'm a little concerned that in for example a 3/4 federation, this server might be down but the other 3 aren't and now clients can't connect.

I see how we might want guardian-specific connection strings to have non-consensus download tokens, but I think we'll also always want to support connecting to the federation if it's in consensus even if 1 server is down.

jkitman:
@justinmoon The problem with the download tokens is that they are unique per server, so a single server cannot return the connection code of another one.  How do you imagine clients get these codes in general?

justinmoon (approved):

PR #2455, opened 2023-05-08
[WIP] refactor: Make auth per connection

douglaz:


douglaz:
It seems too complicated to be worth it, will not go anywhere for now

PR #2459, opened 2023-05-09
feature: add gateways to test framework

jkitman:
Adds the ability to spawn a gateway and gateway webserver from the test framework.


gateway/ln-gateway/src/lib.rs
```diff
@@ -617,13 +621,8 @@ impl Gateway {
                 }
             }
         }
+        warn!("LOOP3");
```
justinmoon, line 624: remove

justinmoon:
Looks good. Will defer the review to @okjodom because he'll be using it.

fedimint-testing/src/gateway.rs
```diff
@@ -0,0 +1,70 @@
...
+        module_gens: ClientModuleGenRegistry,
+    ) -> Self {
+        let listen: SocketAddr = format!("127.0.0.1:{base_port}").parse().unwrap();
+        let address: Url = format!("http://{listen}").parse().unwrap();
+        let mut task = TaskGroup::new();
+        let password = rand::random::<u64>().to_string();
+
+        // Create federation client builder for the gateway
+        let client_builder: DynGatewayClientBuilder =
+            StandardGatewayClientBuilder::new(PathBuf::new(), MemDbFactory.into(), address.clone())
```
okjodom, line 52: `StandardClientBuilder` reads and writes configs to disk. We may want to use an alternative client builder for test cases.

Here is [`TestGatewayClientBuilder`](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/tests/fixtures/client.rs#LL37C1-L37C58)

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -10,6 +10,22 @@
 //! * `IFederationApi` - fake implementation of `IFederationApi` that simulates
 //!   gateway federation client dependency.
 
+use fedimint_testing::federation::FederationTest;
```
okjodom, line 13: comment above this line is not correct, but I will update

fedimint-testing/src/gateway.rs
```diff
@@ -0,0 +1,70 @@
...
+            Arc::new(RwLock::new(lightning)),
+            client_builder.clone(),
+            decoders.clone(),
+            module_gens.clone(),
+            task.clone(),
+        )
+        .await
+        .unwrap();
+
+        gateway.spawn_webserver(listen, password.clone()).await;
```
m1sterc001guy, line 65: Nit: Doesn't need to be this PR but this should probably be moved inside of `Gateway`.

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -10,6 +10,22 @@
...
+use fedimint_testing::fixtures::Fixtures;
+use fedimint_testing::gateway::GatewayTest;
+use ln_gateway::rpc::rpc_client::RpcClient;
+
+async fn fixtures() -> (FederationTest, FederationTest, GatewayTest, RpcClient) {
```
m1sterc001guy, line 18: Maybe we can rename `RpcClient`? Maybe `GatewayRpcClient`?

fedimint-testing/src/fixtures.rs
```diff
@@ -1,49 +1,87 @@
...
+    /// Starts a new gateway
+    pub async fn new_gateway(&self) -> GatewayTest {
+        // TODO: Make construction easier
+        let server_gens = ServerModuleGenRegistry::from(self.servers.clone());
+        let module_kinds = self.params.iter_modules().map(|(id, kind, _)| (id, kind));
+        let decoders = server_gens.decoders(module_kinds).unwrap();
+
+        GatewayTest::new(
+            BASE_PORT.fetch_add(1, Ordering::Relaxed),
+            FakeLightningTest::new(),
```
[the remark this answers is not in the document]
jkitman: Eventually I'll add support for the real fixtures, just trying to get the mock version working now.

gateway/ln-gateway/tests/integration_tests.rs
jkitman: Will rename in follow-up and also put password inside it.

fedimint-testing/src/gateway.rs
jkitman: The fn is inside of `Gateway`.  We spawn it separately so we can know the API is up before continuing.  I do a similar thing for our federations.

PR #2468, opened 2023-05-09
fix: download_client_config

jkitman:
Fix-up of download_client_config
- Removes some unused methods
- Always returns a signed config (waits for signatures)
- Fix off-by-one error on download tokens


fedimint-server/src/net/api.rs
```diff
@@ -259,7 +260,11 @@ impl ConsensusApi {
             }
         }
 
-        Ok(self.get_config_with_sig(dbtx).await)
+        self.get_config_with_sig(dbtx)
+            .await
+            .ok_or(ApiError::server_error(
+                "Client signature not ready".to_string(),
+            ))
```
elsirion, line 267: note to self: We do not commit the dbtx in case an error is returned, so the counter will only count successful retrievals.

elsirion (approved):

PR #2469, opened 2023-05-09
docs(client-ng): add high-level docs

elsirion:


elsirion:
> I think we also accomplish this by making the interfaces more restrictive / serve as guide rails.

I agree, but that will take a bit too figure out fully.

PR #2471, opened 2023-05-09
refactor: common combine sig code

jkitman:
We combine threshold crypto signatures in a few places, this just refactors them into a single place.

Long-term we might move towards FROST instead or a better maintained library.

### DIFF
--- a/modules/fedimint-dummy-server/src/lib.rs
+++ b/modules/fedimint-dummy-server/src/lib.rs
@@ -9,7 +9,7 @@ use fedimint_core::config::{
     ServerModuleConsensusConfig, TypedServerModuleConfig, TypedServerModuleConsensusConfig,
 };
 use fedimint_core::db::{Database, DatabaseVersion, MigrationMap, ModuleDatabaseTransaction};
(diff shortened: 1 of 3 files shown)


justinmoon (approved):

justinmoon:
LGTM. Would be nice if you could take a look tomorrow @elsirion because I'm super familiar with this part of the code.

modules/fedimint-dummy-server/src/lib.rs
```diff
@@ -245,14 +245,20 @@ impl ServerModule for Dummy {
...
-        ConsensusProposal::new_auto_trigger(items.collect().await)
+        let sign_requests: Vec<_> = dbtx
+            .find_by_prefix(&DummySignV1Prefix)
+            .await
+            .collect()
+            .await;
+
+        let consensus_items = sign_requests
+            .into_iter()
+            .filter(|(_, sig)| sig.is_none())
```
elsirion, line 256: You didn't do this filtering before, but I think it's ok? (also only in dummy)

elsirion (commented):
I actually had started reviewing yesterday but got distracted :( nothing major, some nits

PR #2472, opened 2023-05-09
refactor: clean up gateway rpc client

jkitman:
Some refactors to improve gateway testing:
- Can run gateway against feds with 1 peer (faster)
- Rename `RpcClient` to `GatewayRpcClient`
- Have `GatewayRpcClient` store the password and return types for easier testing

### DIFF
--- a/docs/dev-running.md
+++ b/docs/dev-running.md
@@ -104,7 +104,7 @@ The [lightning gateway](../gateway/ln-gateway) connects the federation to the li
 $ gateway-cln balance <FEDERATION-ID>
 
 {
-  "balance_msat": 30000000
+  30000000
 }
 ```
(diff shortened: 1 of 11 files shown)


okjodom (approved):

docs/dev-running.md
```diff
@@ -104,7 +104,7 @@ The [lightning gateway](../gateway/ln-gateway) connects the federation to the li
 $ gateway-cln balance <FEDERATION-ID>
 
 {
-  "balance_msat": 30000000
+  30000000
 }
```
justinmoon, line 108: this section isn't quite right. it needs `--federation-id` before the federation id, and it just returns the balance without braces around it

PR #2477, opened 2023-05-10
feat(client)!: client backup secret derivation

dpc:
In the Client NG we would like one backup for all client modules.

BREAKING CHANGE: Derivation path for module secret changed. Now it is prefixed with an additional level, to allow other derivation types without risking a conflict with any module instance id.

### DIFF
--- a/fedimint-client/src/secret.rs
+++ b/fedimint-client/src/secret.rs
@@ -1,0 +1,23 @@
+use fedimint_core::core::ModuleInstanceId;
+use fedimint_derive_secret::{ChildId, DerivableSecret};
+
+const TYPE_MODULE: ChildId = ChildId(0);
+const TYPE_BACKUP: ChildId = ChildId(1);
+
+pub trait DeriveableSecretClientExt {
+    fn derive_module_secret(&self, module_instance_id: ModuleInstanceId) -> DerivableSecret;
+    fn derive_backup_secret(&self) -> DerivableSecret;
+}
+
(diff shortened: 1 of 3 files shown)


fedimint-client/src/secret.rs
```diff
@@ -0,0 +1,23 @@
+use fedimint_core::core::ModuleInstanceId;
+use fedimint_derive_secret::{ChildId, DerivableSecret};
+
+const TYPE_MODULE: ChildId = ChildId(0);
+const TYPE_BACKUP: ChildId = ChildId(1);
+
+pub trait DeriveableSecretClientExt {
+    fn derive_module_secret(&self, module_instance_id: ModuleInstanceId) -> DerivableSecret;
+    fn derive_backup_secret(&self) -> DerivableSecret;
+}
```
elsirion, line 10: Probably self-documenting enough

elsirion (approved):
Very nice way to encapsulate the details of derivation of different secrets. The era of `*Ext` traits has begun!

dpc:
> Very nice way to encapsulate the details of derivation of different secrets. The era of *Ext traits has begun!

I like `*Ext` traits, but their discoverability is not ideal. But still worth it, IMO.

fedimint-client/src/secret.rs
justinmoon: I think doc strings would be nice. Especially on the struct itself.

PR #2479, opened 2023-05-10
Faucet

Maan2003:
motivation: when you don't cli access. 
example: deployed testing federation, wasm test (so wasm can pay and create invoices)
`curl $FM_FAUCET_BIND_ADDR/invoice -d 21`
`curl $FM_FAUCET_BIND_ADDR/pay -d ln......`


devimint/src/vars.rs
```diff
@@ -104,6 +105,7 @@ declare_vars! {
 
         FM_CLN_EXTENSION_LISTEN_ADDRESS: String = "0.0.0.0:8177";
         FM_GATEWAY_LIGHTNING_ADDR: String = "http://localhost:8177";
+        FM_FAUCET_BIND_ADDR: String = "0.0.0.0:15243";
```
Maan2003, line 108: port number generated by mashing buttons.

dpc (approved):
:fire: 

Haven't tested, but LGTM, and if there are issue, we can follow up with fixes.

devimint/src/vars.rs
dpc: We might go with something more memorable if we expect users to want to type it in their browser. `19999` or something?

justinmoon:
In particular, this will help us write WASM tests that run in a headless browser using a tool like [wasm-bindgen-test](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html)

dpc (approved):

PR #2482, opened 2023-05-10
[WIP][EXPERIMENTAL] feat: load test tool

douglaz:


fedimint-load-test-tool/src/main.rs
```diff
@@ -0,0 +1,627 @@
...
+
+            let gateway_public_key = if let Some(gateway_public_key) = gateway_public_key {
+                Some(gateway_public_key)
+            } else if let Some(generate_invoice_with) = generate_invoice_with {
+                Some(get_gateway_public_key(generate_invoice_with).await?)
+            } else {
+                None
+            };
+            if generate_invoice_with.is_none() {
+                warn!("No --generate-invoice-with given, not LN/gateway tests will be run")
```
dpc, line 132: Sounds more like a `info`

fedimint-load-test-tool/src/main.rs
```diff
@@ -0,0 +1,627 @@
...
+            }
+            run_economy(
```
dpc, line 134: economy? :D

fedimint-load-test-tool/src/main.rs
```diff
@@ -0,0 +1,627 @@
...
+                generate_invoice_with,
+                gateway_public_key,
+                event_sender.clone(),
+            )
+            .await?
+        }
+    };
+    let summary_handle = tokio::spawn(async move {
+        let mut results = HashMap::new();
+        while let Some(event) = event_receiver.recv().await {
```
dpc, line 147: We should probably aggregate here (eg. each time there's more than N events, aggregate to it's own list), otherwise this might quickly balloon, but good enough for time being.

fedimint-load-test-tool/src/main.rs
```diff
@@ -0,0 +1,627 @@
...
+            let entry = results.entry(event.0).or_insert_with(Vec::new);
+            entry.push(event.1);
+        }
+        for (k, mut v) in results {
+            v.sort();
+            let len = v.len();
+            let max = v.iter().last().unwrap();
+            let min = v.first().unwrap();
+            let median = v[len / 2];
+            let sum: Duration = v.iter().sum();
```
dpc, line 157: Nit: since you already need to `iter()` whole thing, you could `.fold` to `(sum, min, max)` in one go.

fedimint-load-test-tool/src/main.rs
```diff
@@ -0,0 +1,627 @@
...
+        #[arg(
+            long,
+            help = "Whether to use lncli or lightning-cli to generate invoices to be paid through the gateway. If none, no gateway/LN tests will be run. Note that you can't generate an invoice using same lightning node used by the gateway (i.e self payment is forbidden)"
+        )]
+        generate_invoice_with: Option<LnType>,
+    },
+}
+
+#[derive(Debug, Clone)]
+struct Event(String, Duration);
```
dpc, line 89: Name the fields. Will make the code using it easier to understand.

dpc (commented):
Looking great.

fedimint-load-test-tool/src/main.rs
```diff
@@ -0,0 +1,627 @@
...
+    Ok(futures)
+}
+
+async fn do_user_task(
+    client: Client,
+    notes: TieredMulti<SpendableNote>,
+    generate_invoice_with: Option<LnType>,
+    event_sender: mpsc::UnboundedSender<Event>,
+) -> anyhow::Result<()> {
+    let amount = notes.total_amount();
```
dpc, line 254: Shouldn't this be in some kind of a loop so each user does this things multiple times over certain time?

douglaz: Ideally yes, in practice 1) this is tricky (you can't reissue the same notes twice and it's very slow to split the initial big notes in small notes) and 2) doing once with dozens of users is enough to measure how slow is the system

fedimint-load-test-tool/src/main.rs
douglaz: This is really far from being an issue right know

fedimint-load-test-tool/src/main.rs
dpc: Lots of systems can sustain a burst of load initially, but then slow down significantly after caches and write buffers get full, deallocators and garbage collectors start hitting fragmentation and other resource reclamation and so and so.

Splitting notes could be the part of natural user flow maybe. E.g. the user could always send 20% of what they have, etc.

douglaz: Added a `--notes-per-user`. So the user now can reissue many notes instead of just one

elsirion (commented):
Ideally there'd be a basic, very short CLI test that just checks if the tool still runs (doesn't need to do an actual load test, that's probably a bit too much for CI).

douglaz:
@elsirion added some cli tests

dpc (approved):

dpc:
Looks good. IMO best to land early, as it's non-production code, and will be easier to review any improvements as follow-ups.

fedimint-load-test-tool/src/common.rs
```diff
@@ -0,0 +1,393 @@
...
+impl ToCmdExt for GatewayClnCli {
+    type Fut = std::future::Ready<devimint::util::Command>;
+
+    fn cmd(self) -> Self::Fut {
+        // try to use alias if set
+        let gw = std::env::var("FM_GWCLI_CLN")
+            .map(|s| s.split_whitespace().map(ToOwned::to_owned).collect())
+            .unwrap_or_else(|_| vec!["gateway-cln".into()]);
+        let mut cmd = tokio::process::Command::new(&gw[0]);
+        cmd.args(&gw[1..]);
```
Maan2003, line 369: nit: this logic should be reused in all three commands

fedimint-load-test-tool/src/common.rs
```diff
@@ -0,0 +1,393 @@
...
+        std::future::ready(devimint::util::Command {
+            cmd,
+            args_debug: lightning_cli,
+        })
+    }
+}
+
+pub struct GatewayClnCli;
+impl ToCmdExt for GatewayClnCli {
+    type Fut = std::future::Ready<devimint::util::Command>;
```
Maan2003, line 361: BTW, you don't need to impl this trait
```suggestion
impl GatewayClnCli {
    pub async fn cmd(&self) -> devimint::util::Command;
```

Maan2003:
just nits, not blocking

fedimint-load-test-tool/src/common.rs
```diff
@@ -0,0 +1,393 @@
...
+
+pub async fn get_notes_cli(amount: &Amount) -> anyhow::Result<TieredMulti<SpendableNote>> {
+    cmd!(FedimintCli, "fetch").out_string().await?;
+    cmd!(FedimintCli, "spend", amount.msats.to_string())
+        .out_json()
+        .await?["note"]
+        .as_str()
+        .map(parse_ecash)
+        .transpose()?
+        .ok_or_else(|| anyhow::anyhow!("missing notes output"))
```
Maan2003, line 41: `context` (use anyhow::Context) is really neat here
```suggestion
        .context("missing notes output")
```

elsirion (approved):
Agree with @dpc, LGTM

PR #2483, opened 2023-05-10
feature: begin to add ln tests

jkitman:
Adds first test for the new LN client, changes some of the gateway API to be easier to test/use.


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -120,61 +119,42 @@ pub enum LnReceiveState {
 
 #[apply(async_trait_maybe_send!)]
 impl LightningClientExt for Client {
-    async fn fetch_active_gateway(&self) -> anyhow::Result<LightningGateway> {
+    async fn get_active_gateway(&self) -> anyhow::Result<LightningGateway> {
```
justinmoon, line 122: Might be more natural to return `Option` here

fedimint-testing/src/gateway.rs
```diff
@@ -18,32 +19,42 @@ use crate::federation::FederationTest;
...
+        self.feds.push(
+            client
+                .connect_federation(ConnectFedPayload { connect })
+                .await
+                .unwrap(),
+        );
+    }
+
+    /// Returns the last registration we sent to a fed
+    pub fn registration(&self) -> LightningGateway {
```
m1sterc001guy, line 48: nit: `last_registration` might be more descriptive?

fedimint-testing/src/gateway.rs
```diff
@@ -73,11 +84,17 @@ impl GatewayTest {
         })
         .await;
 
-        Self {
+        let mut gateway = Self {
+            feds: vec![],
             password,
-            address,
+            api: address,
             _config_dir,
```
m1sterc001guy, line 91: Can these just be deleted if they're not used?

gateway/ln-gateway/src/actor.rs
```diff
@@ -38,6 +39,7 @@ pub struct GatewayActor {
     task_group: TaskGroup,
     gw_rpc: GatewayRpcSender,
     pub sender: Sender<Arc<AtomicBool>>,
+    registration: LightningGateway,
```
m1sterc001guy, line 42: I know you're just refactoring this but its a bit confusing to me. Why does the "actor" need a handle to the `LightningGateway`? Actors ideally should just be federation clients. Is this needed for the federation public key?

gateway/ln-gateway/src/actor.rs
```diff
@@ -149,6 +151,10 @@ impl GatewayActor {
         gw_rpc: GatewayRpcSender,
     ) -> Result<Self> {
         let register_client = client.clone();
+        let registration = register_client
+            .config()
+            .to_gateway_registration_info(route_hints.clone(), GW_ANNOUNCEMENT_TTL);
+        let registration_sent = registration.clone();
         task_group
             .spawn("Register with federation", |_| async move {
```
m1sterc001guy, line 159: Nit: this code always confused me, probably should return to this sometime and fix it. I don't understand why it is safe to do this in a separate thread and continue, even if the registration with the federation fails.

modules/fedimint-wallet-tests/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "fedimint-wallet-tests"
+version = "0.1.0"
+authors = ["The Fedimint Developers"]
+edition = "2021"
+description = "fedimint-wallet is a n on-chain bitcoin wallet module. It uses a key-value store and is not a standard HD wallet."
```
m1sterc001guy, line 6: update?

gateway/ln-gateway/src/actor.rs
m1sterc001guy: Oh I remember, it needs to periodically register with the federation. We should probably move this to a new function to be clearer. Not blocking on this, just thinking out loud

m1sterc001guy (approved):

fedimint-testing/src/gateway.rs
jkitman: The config dir we need to hold a handle to otherwise it'll drop which auto deletes the config dirs.

gateway/ln-gateway/src/actor.rs
jkitman: Yeah might want to change the retry logic, possibly block here.

m1sterc001guy (approved):

PR #2486, opened 2023-05-10
fix: LN Outgoing Payments not idempotent

m1sterc001guy:
The current LN outgoing payment state machine has a few bugs:

 - It calls the gateway's `pay_invoice` in the trigger function. Trigger functions need to be idempotent. Paying an invoice by calling the gateway is not idempotent. This can cause `pay_invoice` to be called multiple times if a different state is executed before `pay_invoice` completes. 
 - `LnPay` does not wait for the change from the `MintOutput`. When executing from the CLI, this will cause the change to not be reflected after the outgoing payment, which causes the test to fail.

This PR changes:
 - Moves calling the gateway's `pay_invoice` from the trigger function to the transition function.
 - Outgoing payment state machine will not transition to funded until after the `LightningOutpoutOutcome` is accepted, so that the workflow can call `pay_invoice` immediately after (instead of just waiting for the tx to be accepted).
 - Adds LN outgoing tests to the CLI tests.

### DIFF
--- a/modules/fedimint-ln-client/src/pay.rs
+++ b/modules/fedimint-ln-client/src/pay.rs
@@ -6,6 +6,7 @@ use fedimint_client::transaction::{ClientInput, TxSubmissionError};
 use fedimint_client::DynGlobalClientContext;
 use fedimint_core::api::GlobalFederationApi;
 use fedimint_core::config::FederationId;
+use fedimint_core::core::Decoder;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::task::sleep;
 use fedimint_core::{Amount, OutPoint, TransactionId};
@@ -74,7 +75,7 @@ impl State for LightningPayStateMachine {
     ) -> Vec<StateTransition<Self>> {
         match &self.state {
             LightningPayStates::CreatedOutgoingLnContract(created_outgoing_ln_contract) => {
-                created_outgoing_ln_contract.transitions(&self.common, &context, global_context)
+                created_outgoing_ln_contract.transitions(&self.common, context, global_context)
             }
             LightningPayStates::Canceled => {
                 vec![]
@@ -144,7 +145,7 @@ impl LightningPayCreatedOutgoingLnContract {
             .api()
             .await_output_outcome::<LightningOutputOutcome>(
                 out_point,
-                Duration::MAX,
+                Duration::from_millis(i32::MAX as u64),
                 &module_decoder,
             )
             .await
@@ -207,14 +208,13 @@ impl LightningPayFunded {
         let gateway = self.gateway.clone();
         let payload = self.payload.clone();
         let contract_id = self.payload.contract_id;
-        let timeout_context = global_context.clone();
         vec![StateTransition::new(
             // Immediately try to pay the invoice by contacting the gateway
             future::ready(()),
             move |_dbtx, (), old_state| {
                 Box::pin(Self::transition_outgoing_contract_execution(
                     old_state,
-                    timeout_context.clone(),
+                    global_context.clone(),
                     contract_id,
                     gateway.clone(),
                     payload.clone(),
--- a/modules/fedimint-ln-client/src/pay.rs
+++ b/modules/fedimint-ln-client/src/pay.rs
@@ -4,14 +4,16 @@ use std::time::Duration;
 use fedimint_client::sm::{ClientSMDatabaseTransaction, OperationId, State, StateTransition};
 use fedimint_client::transaction::{ClientInput, TxSubmissionError};
 use fedimint_client::DynGlobalClientContext;
+use fedimint_core::api::GlobalFederationApi;
 use fedimint_core::config::FederationId;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::task::sleep;
-use fedimint_core::{Amount, TransactionId};
+use fedimint_core::{Amount, OutPoint, TransactionId};
 use fedimint_ln_common::contracts::outgoing::OutgoingContract;
 use fedimint_ln_common::contracts::{ContractId, IdentifiableContract, Preimage};
-use fedimint_ln_common::{LightningGateway, LightningInput};
+use fedimint_ln_common::{LightningGateway, LightningInput, LightningOutputOutcome};
 use fedimint_wallet_client::api::WalletFederationApi;
+use futures::future;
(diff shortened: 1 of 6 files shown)


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -200,6 +208,34 @@ impl MintClientExt for Client {
...
+        if let TransactionStatus::Accepted {
+            epoch: _epoch,
+            outputs,
+        } = tx_outcome
+        {
+            // Assume that if the transaction has more than one output, the output
+            // at index 1 is the change.
+            if outputs.len() > 1 {
+                let (_, mint_client) = mint_client(self);
+                let out_point = OutPoint { txid, out_idx: 1 };
```
m1sterc001guy, line 226: Not sure if this is the most elegant way of awaiting the change from the mint module. It assumes something about the transaction structure. Open to suggestions!

justinmoon: You could pass in an `Outpoint` instead of a `TransactionId`? Then at least you could move the assumption from this utility function to the caller?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -212,6 +220,31 @@ impl MintClientExt for Client {
             .await
     }
 
+    async fn await_mint_change(
```
jkitman, line 223: How come this needs to be a custom function versus just called on the global api?

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -194,35 +210,22 @@ impl LightningPayFunded {
...
-                    Box::pin(Self::transition_pay_timeout(
-                        old_state,
-                        global_context.clone(),
-                        contract_id,
-                    ))
-                },
-            ),
-        ]
+        vec![StateTransition::new(
+            // Immediately try to pay the invoice by contacting the gateway
```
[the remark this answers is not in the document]
m1sterc001guy: It can if we want it to. The thinking behind making it a separate state is if the client crashes/stops after awaiting the Lightning Outcome, but before trying to pay the invoice, we won't need to await the Lightning Outcome again (if it was in the same transition function we'd need to do the whole thing again).

I think it's also a bit more user friendly just to know when the offer has been accepted, but not executed yet

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -133,17 +138,26 @@ impl LightningPayCreatedOutgoingLnContract {
...
         txid: TransactionId,
-    ) -> Result<(), ()> {
+    ) -> Result<(), GatewayPayError> {
+        let out_point = OutPoint { txid, out_idx: 0 };
         global_context
-            .await_tx_accepted(common.operation_id, txid)
+            .api()
+            .await_output_outcome::<LightningOutputOutcome>(
+                out_point,
+                Duration::MAX,
```
elsirion, line 150: Is `Duration::MAX` still a problem for WASM @Maan2003?

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -251,50 +255,19 @@ impl LightningPayFunded {
...
     async fn transition_outgoing_contract_execution(
         old_state: LightningPayStateMachine,
         global_context: DynGlobalClientContext,
-        result: Result<String, GatewayPayError>,
         contract_id: ContractId,
+        gateway: LightningGateway,
+        payload: PayInvoicePayload,
     ) -> LightningPayStateMachine {
-        match result {
+        match Self::gateway_pay_invoice(gateway, payload).await {
```
elsirion, line 264: State transitions should ideally not make API calls since they can't run in parallel and e.g. the API being slow would block all other state machines from making progress. What is the reason for moving the fn here? It will still be run multiple times if the client is restarted at the wrong time, so the API endpoint has to be idempotent anyway (tbf, doing the call here will make this less frequent).

modules/fedimint-mint-client/src/lib.rs
elsirion: This needs a `await_primary_module_output_finalized(OutPoint)` function since we don't even know if the change will always be generated by the mint module. This didn't appear super crucial at first since we know we will get the money once the tx is accepted, but should be patched in. We'd also need to return a change `Option<OutPoint>` whenever one might be added.

elsirion (changes requested):

elsirion:
This is the important one: https://github.com/fedimint/fedimint/pull/2486#discussion_r1191547880

elsirion (approved):
We can probably not fully fix the problem without client-NG on the gateway side, so let's merge this first, it will reduce but not fixthe problem.

modules/fedimint-ln-client/src/pay.rs
Maan2003: Yes, Duration::MAX for wasm needs to be `Duration::from_millis(i32::MAX)`
But lets not block this.

modules/fedimint-ln-client/src/pay.rs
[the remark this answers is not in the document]
m1sterc001guy: As Eric mentioned below, this should ideally move back to the trigger function once we have Client NG on the gateway side.

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: Fixed, thanks for the pointer

modules/fedimint-mint-client/src/lib.rs
m1sterc001guy: Actually to minimize merge conflicts I will do this in a separate PR. Made issue and assigned myself: https://github.com/fedimint/fedimint/issues/2495. Working on this next.

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: As discussed offline, this is difficult to fix without client NG on the gateway side. Made an issue: https://github.com/fedimint/fedimint/issues/2496

modules/fedimint-mint-client/src/lib.rs
jkitman: We also will probably need a `get_total_funds` implemented on all modules so we can easily return that to the client for all modules.

elsirion (approved):

PR #2487, opened 2023-05-10
fix(ln-client): allow injecting user-specified meta into operations

justinmoon:
Extends approach from https://github.com/fedimint/fedimint/pull/2347 to lightning client

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -68,11 +68,12 @@ pub trait LightningClientExt {
     ) -> anyhow::Result<LightningGateway>;
     async fn fetch_registered_gateways(&self) -> anyhow::Result<Vec<LightningGateway>>;
 
-    async fn pay_bolt11_invoice(
+    async fn pay_bolt11_invoice<M: Serialize + Send>(
         &self,
         fed_id: FederationId,
         invoice: Invoice,
         active_gateway: LightningGateway,
+        extra_meta: M,
     ) -> anyhow::Result<OperationId>;
 
     async fn subscribe_ln_pay_updates(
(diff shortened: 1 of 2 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -257,7 +268,16 @@ impl LightningClientExt for Client {
             )
             .await?;
         let tx = TransactionBuilder::new().with_output(output.into_dyn(ln_client_id));
-        let operation_meta_gen = |txid| OutPoint { txid, out_idx: 0 };
```
justinmoon, line 260: Here we were forgetting to add `invoice`

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -272,7 +292,7 @@ impl LightningClientExt for Client {
             &mut dbtx,
             operation_id,
             LightningCommonGen::KIND.as_str(),
-            LightningMeta::Receive {
+            LightningMetaVariants::Receive {
```
m1sterc001guy, line 295: This is the bug I think. `finalize_and_submit_transaction` will actually add an operation log entry so we don't need to do this here. I think if you keep this you should see a warning on the CLI when doing receives

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -418,7 +438,13 @@ async fn ln_operation(client: &Client, operation_id: OperationId) -> anyhow::Res
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-enum LightningMeta {
+pub struct LightningMeta {
+    variant: LightningMetaVariants,
+    extra_meta: serde_json::Value,
```
m1sterc001guy, line 443: What are you thinking of using this for? Just curious

justinmoon: When an outgoing payment reaches a terminal state, it would be nice to know which invoice was paid.

modules/fedimint-ln-client/src/lib.rs
elsirion: Yep, there is a fix in #2486, so there will be a conflict https://github.com/fedimint/fedimint/pull/2486/files#diff-d754f9b1f77285e47d11bdecd4e3df652f846e5cedd12ded095ae84816a64cc2L205-L217

dpc:
Closing as stale. Please re-open if still needed.

elsirion:
Still needed, anyone wants to pick it up? Otherwise should be tracked in an issue.

m1sterc001guy:
Looks pretty straightforward, I can take it on

m1sterc001guy:
Closing in favor of https://github.com/fedimint/fedimint/pull/3276

justinmoon:
Sorry I forgot this was still open :/

PR #2491, opened 2023-05-11
Client refactor

jkitman:
- Returns the tx submit error so clients can see why a tx failed
- Refactors out common instance resources to eliminate boilerplate and help remove more legacy ids

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -114,7 +114,7 @@ use crate::sm::{
 };
 use crate::transaction::{
     tx_submission_sm_decoder, ClientInput, ClientOutput, TransactionBuilder,
-    TransactionBuilderBalance, TxSubmissionContext, TxSubmissionStates,
+    TransactionBuilderBalance, TxSubmissionContext, TxSubmissionError, TxSubmissionStates,
     TRANSACTION_SUBMISSION_MODULE_INSTANCE,
 };
 
@@ -187,6 +187,7 @@ dyn_newtype_define! {
 }
 
 impl DynGlobalClientContext {
+    // TODO: Remove in favor of `await_tx_accepted`
     pub async fn await_tx_rejected(&self, operation_id: OperationId, txid: TransactionId) {
         let update_stream = self.transaction_update_stream(operation_id).await;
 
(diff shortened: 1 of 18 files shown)


modules/fedimint-ln-client/src/pay.rs
```diff
@@ -404,12 +401,10 @@ impl LightningPayRefundable {
...
         // TODO: Remove polling
+        let client = global_context.client();
+        let (_, instance) = client.get_first_module::<LightningClientModule>(&KIND);
         loop {
-            let contract = global_context
-                .api()
-                .with_module(LEGACY_HARDCODED_INSTANCE_ID_LN)
-                .get_outgoing_contract(contract_id)
-                .await;
+            let contract = instance.api.get_outgoing_contract(contract_id).await;
```
elsirion, line 407: 1. The state transition belongs to a *particular instance*, we cannot just get the first instance here
2. There is no need to get a whole client to get an API. The global context already has both the API struct as well as the instance id of the module the state machine belongs to.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -22,23 +20,11 @@ pub trait WalletClientExt {
 
 impl WalletClientExt for Client {
     fn get_network(&self) -> Network {
-        let (_, wallet_client) = wallet_client(self);
-        wallet_client.get_network()
+        let (wallet, _instance) = self.get_first_module::<WalletClientModule>(&KIND);
```
elsirion, line 23: TODO for later: add `KIND` to client module trait such that it deosn't need to be supplied here

elsirion (approved):

fedimint-client/src/lib.rs
```diff
@@ -653,6 +670,16 @@ impl Client {
...
+/// Resources particular to a module instance
+pub struct ClientModuleInstance {
+    /// Instance id of the module
+    pub id: ModuleInstanceId,
+    /// Module-specific DB
+    pub db: Database,
+    /// Module-specific API
+    pub api: DynFederationApi,
+}
+
```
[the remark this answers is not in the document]
jkitman: How would we do that?  Maybe my Rust-fu isn't good enough here.

PR #2493, opened 2023-05-11
fix: Fix Gateway PegIn

m1sterc001guy:
Recent changes to gateway-cli broke the pegin.sh script for gateways. Gateways in tmuxinator won't have ecash notes which makes testing a bit harder.


jkitman (approved):
Ah I broke it, we keep breaking it since it isn't tested.

justinmoon:
The new client should make it much easier to get rid of this script. We can just generate an address, send to it, mine blocks, and call `fedimint-cli ng await-deposit` and the peg-in happen behind the scenes.

PR #2494, opened 2023-05-11
Shared Fedimint-UI workspace : Part 3 / 3

### LINKED ISSUE #2429: Code sharing and common code for UI projects
With #2360 coming close to being mergeable, we're going to have the start of a reusable set of UI components, theme values, and utilities. Since there will be other UI projects in this repository (i.e. `gateway/ui`) it would be ideal if it was easy to share and reuse code across these projects. There are a few ways this could be organized:

1. One folder that all UI projects live in
2. A `package.json` / `yarn.lock` in the root directory that points to the various folders that frontend projects are nested in

Both of these approaches will require some kind of monorepo code sharing solution. Some of the options here are:

* [Yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/)
* [PNPM workspaces](https://pnpm.io/workspaces)
* [Turborepo](https://turbo.build/repo)

The last thing we might want to figure out is publishing a UI library for future fedimint UI projects. This would allow folks working outside of the repository to pull in the Chakra-based UI library and make things that still have the same look & feel of a fedimint UI. I'll defer that to a future issue since that would require setting up an npm organization and coming up with a build & release pipeline. But however we organize the projects now would lay the groundwork for having that package.

okjodom:
In this PR:

- [x] - Extract parts of `apps/guardian-ui` as reusable react components in `@fedimint/ui`
- [x] - Polish library and package docs

Closes #2429


okjodom:
WIP: Figuring out how to bundle svgs in `@fedimint/ui`.

I'm trying to use [tsup](https://tsup.egoist.dev/#what-can-it-bundle) to bundle ts files for the libraries and it works nice, except it seems to have this [deficiency in bundling svg files ](https://github.com/egoist/tsup/discussions/811) - or the doc's are just not very instructive.

I may have three ways forward. Either
- figure out how to use tsup to bundle the svgs in shared react component
- inject svg to reusable components that provide slots (shared component doesn't come with static assets)
- switch bundlers. I'm hoping tsup suffices but may actually have to switch over to something like webpack + svgr-webpack

cc @wbobeirne for ideas

fedimint-ui/apps/guardian-ui/src/components/RunDKG.tsx
```diff
@@ -84,7 +84,7 @@ export const RunDKG: React.FC<Props> = ({ next }) => {
       <CircularProgress
         isIndeterminate={!isWaitingForOthers}
         value={isWaitingForOthers ? progress : undefined}
-        color={theme.colors.blue[400]}
+        color={theme.colors?.blue[400]}
```
wbobeirne, line 87: `colors` should never be undefined, we should figure out why the types for this got loosened.

wbobeirne:
If we want to avoid a bundler, we'd either have to manually create React components for each icon, or come up with a stand-alone svgr part of the build phase that generates the components for us.

If we want to keep it completely separate, we could also split icons out into their own package and do any bundling there in isolation.

fedimint-ui/apps/guardian-ui/src/components/RunDKG.tsx
okjodom: 💯

The shared components in `@fedimint/ui` have an empty theme object at runtime. I suspect I'm missing something so the theme context is appropriately cascaded and observed

okjodom:
> If we want to avoid a bundler, we'd either have to manually create React components for each icon, or come up with a stand-alone svgr part of the build phase that generates the components for us.

I think we already have this. Adding an svgr phase to bundling is my current blocker, which might force me to change bundlers. [tsup](https://tsup.egoist.dev/) is surprising and disappointing me to this end

> If we want to keep it completely separate, we could also split icons out into their own package and do any bundling there in isolation.

Makes sense long term, yes? We'd have a shared icon lib, similar to the shared ui lib introduced here

fedimint-ui/.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+  "tabWidth": 2,
```
EthnTuttle, line 2: This is going to make @justinmoon so happy. 😆

fedimint-ui/apps/gateway-ui/src/index.tsx
```diff
@@ -17,16 +17,16 @@ const mintgate = gateway_api ? new MockMintgate() : new MockMintgate();
...
-		<Fonts />
-		<ChakraProvider theme={theme}>
-			<ApiProvider props={{ mintgate, explorer }}>
-				<Admin />
-			</ApiProvider>
-		</ChakraProvider>
-	</React.StrictMode>
+  <React.StrictMode>
+    <Fonts />
+    <ChakraProvider theme={theme}>
```
EthnTuttle, line 26: Should this be using the `<SharedChakraProvider>` like guardian-ui? Or not yet?

fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -160,7 +159,7 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
             isDisabled={!!statePassword}
           />
           <FormHelperText>
-            {'You\'ll need this every time you visit this page.'}
+            You&#39;ll need this every time you visit this page.
```
EthnTuttle, line 162: No issue, but curious about this change? Why?

fedimint-ui/.prettierrc
```diff
@@ -0,0 +1,9 @@
...
+  "useTabs": false,
+  "printWidth": 80,
+  "semi": true,
+  "trailingComma": "es5",
```
EthnTuttle, line 6: It might not matter much since this is just style but in `fedimint-ui/packages/tsconfig/react-app.json` we have it set to es6.

EthnTuttle (commented):
Just need to run locally and confirm not run-time issues. Otherwise, just a few questions that may or may not be issues.

fedimint-ui/apps/gateway-ui/src/index.tsx
okjodom: Correct, it should be. Didn't bother changing it because gateway-ui doesn't consume any of the shared UI elements just yet

fedimint-ui/.prettierrc
okjodom: def don't need this rule if we are targetting es anyways

fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
okjodom: running `yarn format` on the old string with escaped `\'` would change the string to double quote in order to avoid having the escape.

now fixed by enhancing the eslint `quotes` rule definition. See `eslint-config/index.js`

fedimint-ui/packages/eslint-config/index.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     sourceType: "module",
   },
   rules: {
-    quotes: ["error", "single"],
+    quotes: ["error", "single", { avoidEscape: true, allowTemplateLiterals: true  }],
```
EthnTuttle, line 16: nice

EthnTuttle (approved):
LGTM.

PR #2497, opened 2023-05-11
First mint tests

jkitman:
- Defaults fed size to 1 for mocked tests
- First mint module test for OOB payment
- Fixes LN gateway not waiting to connect to fed, caused a test to fail, should return error and retry client side

### DIFF
--- a/gateway/ln-gateway/src/actor.rs
+++ b/gateway/ln-gateway/src/actor.rs
@@ -15,6 +15,7 @@ use fedimint_ln_common::LightningGateway;
 use futures::stream::{BoxStream, StreamExt};
 use rand::{CryptoRng, RngCore};
 use tokio::sync::mpsc::{self, Receiver, Sender};
+use tokio::sync::Notify;
 use tonic::Status;
 use tracing::{debug, error, info, instrument, warn};
 
@@ -155,6 +156,8 @@ impl GatewayActor {
             .config()
             .to_gateway_registration_info(route_hints.clone(), GW_ANNOUNCEMENT_TTL);
         let registration_sent = registration.clone();
+        let notify = Arc::new(Notify::new());
+        let notfiy_sent = notify.clone();
         task_group
             .spawn("Register with federation", |_| async move {
                 loop {
@@ -174,6 +177,7 @@ impl GatewayActor {
                     {
                         Ok(_) => {
                             info!("Connected with federation");
+                            notfiy_sent.notify_one();
                             tokio::time::sleep(GW_ANNOUNCEMENT_TTL / 2).await;
                         }
                         Err(e) => {
@@ -184,6 +188,8 @@ impl GatewayActor {
                 }
             })
             .await;
+        // Block until we have registered to return
+        notify.notified().await;
 
         // Create a channel that will be used to shutdown the HTLC thread
         let (sender, receiver) = mpsc::channel::<Arc<AtomicBool>>(100);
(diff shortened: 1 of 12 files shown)


modules/fedimint-dummy-client/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub trait DummyClientExt {
 
 #[apply(async_trait_maybe_send!)]
 impl DummyClientExt for Client {
-    async fn print_money(&self, amount: Amount) -> anyhow::Result<()> {
+    async fn print_money(&self, amount: Amount) -> anyhow::Result<(OperationId, OutPoint)> {
```
jkitman, line 65: @elsirion we should be more consistent on when we use `OperationId` or `OutPoint` or `TransactionId`

elsirion: Typically the high-level functions exposed by the `*Ext`-traits only return `OperationId`s since internally there might be multiple transactions belonging to the same operation (e.g. LN submit offer+redeem contract). These `operation_id -> txid` mappings have to be handled internally somehow and only the operation id should be needed to query the high-level state of the operation.

modules/fedimint-mint-tests/tests/tests.rs
```diff
@@ -0,0 +1,41 @@
...
+        .with_module(1, DummyClientGen, DummyGen, DummyConfigGenParams::default())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn sends_ecash_out_of_band() -> anyhow::Result<()> {
+    // Print notes for client1
+    let fed = fixtures().new_fed().await;
+    let (client1, client2) = fed.two_clients().await;
+    let (op, outpoint) = client1.print_money(sats(1000)).await?;
+    client1.await_claim_notes(op, outpoint.txid).await?;
```
elsirion, line 24: As you mentioned, we shouldn't need both out point and txid here. `print_money` could save all required data in the operation meta so that a high-level update stream (or simpler await fn) can be built for it.

gateway/ln-gateway/src/actor.rs
```diff
@@ -147,43 +147,27 @@ impl GatewayActor {
...
-                        Ok(_) => {
-                            info!("Connected with federation");
-                            tokio::time::sleep(GW_ANNOUNCEMENT_TTL / 2).await;
-                        }
-                        Err(e) => {
-                            warn!("Failed to connect with federation: {}", e);
-                            tokio::time::sleep(GW_ANNOUNCEMENT_TTL / 4).await;
-                        }
-                    }
-                }
```
elsirion, line 184: By removing the outer loop you remove re-registration. This means the gateway will disappear for clients after `GW_ANNOUNCEMENT_TTL` since the re-registration after `GW_ANNOUNCEMENT_TTL / 2` doesn't happen.

Maybe use `Notify` to wait for the first successful registration?

elsirion (changes requested):
Re-registration is necessary or the gateway will stop working for clients.

Maybe this actually tells us to test this behavior by reducing the TTL in tests?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -353,6 +358,18 @@ impl MintClientExt for Client {
             }
         }))
     }
+
+    async fn total_notes(&self) -> Amount {
```
justinmoon, line 362: `total_amount` is what we call the analagous function on `TieredMulti`

jkitman: We should eventually add this to all client modules so we can have a total of all modules.

elsirion (approved):

PR #2499, opened 2023-05-11
build(deps): update rocksdb

otech47:
New rust-rocksdb update dropped a few days ago: https://github.com/rust-rocksdb/rust-rocksdb/releases/tag/v0.21.0

Just a minor version bump so shouldn't be too bad but one fix in particular makes `fedimint-rocksdb` more compatible as a dependency when building binaries that target iOS

Probably should review some of the other fixes in the release before bumping this dependency


otech47:
Replaced by #2500

PR #2500, opened 2023-05-11
chore: upgrade rust-rocksdb

justinmoon:


m1sterc001guy (approved):
LGTM, I think this release has async reads which we might want to explore. Not pressing though

PR #2501, opened 2023-05-12
Make client root secret encoding flexible using strategy

### LINKED ISSUE #2478: Client NG: don't keep secret in DB
Get it from the builder instead

elsirion:
Fixes #2478 in a different way than originally intended. The problem we are trying to solve is that there are many possible encoding of a master private key (e.g. BIP39, SLIP39, …) and we don't want to choose for the user. Since we cannot convert our root secret back to these encodings the original encoding needs to be saved.

The simple fix would be making the one constructing a client give us a `DerivableSecret` instead of managing it ourselves, but that puts more work on the implementers of clients and allows for more mistakes

This PR instead introduces a strategy pattern that lets us build strategies using different encoding schemes and let the implementers of client apps choose which one to use (or even roll their own).

### DIFF
--- a/fedimint-bip39/src/lib.rs
+++ b/fedimint-bip39/src/lib.rs
@@ -1,0 +1,48 @@
+//! BIP39 client secret support crate
+
+use std::io::{Read, Write};
+
+use fedimint_client::derivable_secret::DerivableSecret;
+use fedimint_client::secret::RootSecretStrategy;
+use fedimint_core::encoding::{Decodable, DecodeError, Encodable};
+use rand::{CryptoRng, RngCore};
+
+/// BIP39 root secret encoding strategy allowing retrieval of the seed phrase.
+#[derive(Debug)]
+pub struct Bip39RootSecretStrategy<const WORD_COUNT: usize = 12>;
+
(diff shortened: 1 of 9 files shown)


fedimint-bip39/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+use fedimint_client::derivable_secret::DerivableSecret;
+use fedimint_client::secret::RootSecretStrategy;
+use fedimint_core::encoding::{Decodable, DecodeError, Encodable};
+use rand::{CryptoRng, RngCore};
+
+/// BIP39 root secret encoding strategy allowing retrieval of the seed phrase.
+#[derive(Debug)]
+pub struct Bip39RootSecretStrategy<const WORD_COUNT: usize = 12>;
+
+impl<const WORD_COUNT: usize> RootSecretStrategy for Bip39RootSecretStrategy<WORD_COUNT> {
```
dpc, line 14: I can't look deeper into it RN, but couldn't it operate on just `String` (whole phrase), instead of splitting things in words in type system etc.?

justinmoon:
Merged https://github.com/fedimint/fedimint/pull/2529 instead

Issue #2495, opened 2023-05-11
Add await_primary_module_output_finalized(OutPoint)


m1sterc001guy:
The CLI sometimes needs to wait for change if the transaction builder cannot make exact change. The primary module is the module that will add change to the transaction. There should be a generic function that can await the change output being finalized so that the CLI can pickup the change.

Hacky solution for the mint module implemented here: https://github.com/fedimint/fedimint/pull/2486

PR #2502, opened 2023-05-12
feat: await primary module change

m1sterc001guy:
This PR adds a generic way of awaiting a primary modules output to be finalized. Often times, after the output outcome happens on the server side, the client needs to wait on some additional operations on the client side (like claiming ecash notes as change from a Lightning transaction).

Right now we don't have a good way to wait on change. This PR adds a function `await_primary_module_output_finalized` that awaits on an outpoint to be finalized. This PR also changes `finalize_and_submit_transaction` to return an `Option<OutPoint>` which will contain the `OutPoint` of the change (if there is change).

Fixes: https://github.com/fedimint/fedimint/issues/2495

### DIFF
--- a/fedimint-cli/src/ng.rs
+++ b/fedimint-cli/src/ng.rs
@@ -13,7 +13,7 @@ use fedimint_core::config::ClientConfig;
 use fedimint_core::core::{ModuleInstanceId, ModuleKind};
 use fedimint_core::encoding::Decodable;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
-use fedimint_core::{Amount, OutPoint, ParseAmountError, TieredMulti, TieredSummary};
+use fedimint_core::{Amount, ParseAmountError, TieredMulti, TieredSummary};
 use fedimint_ln_client::{LightningClientExt, LnPayState, LnReceiveState};
 use fedimint_mint_client::{MintClientExt, MintClientModule, SpendableNote};
 use futures::StreamExt;
@@ -159,8 +159,7 @@ pub async fn handle_ng_command(
             let mut updates = client.subscribe_to_ln_receive_updates(operation_id).await?;
             while let Some(update) = updates.next().await {
                 match update {
-                    LnReceiveState::Claimed { txid } => {
-                        client.await_claim_notes(operation_id, txid).await?;
+                    LnReceiveState::Claimed => {
                         return get_note_summary(&client).await;
                     }
                     LnReceiveState::Canceled { reason } => {
@@ -177,7 +176,7 @@ pub async fn handle_ng_command(
         ClientNg::LnPay { bolt11 } => {
             client.select_active_gateway().await?;
 
-            let (operation_id, txid) = client
+            let operation_id = client
                 .pay_bolt11_invoice(config.federation_id, bolt11)
                 .await?;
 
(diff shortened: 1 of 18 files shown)


fedimint-cli/src/ng.rs
```diff
@@ -128,7 +128,12 @@ pub async fn handle_ng_command(
             while let Some(update) = updates.next().await {
                 match update {
                     LnReceiveState::Claimed { txid } => {
-                        client.await_claim_notes(operation_id, txid).await?;
+                        client
+                            .await_primary_module_output_finalized(
```
m1sterc001guy, line 132: This should probably be moved to `subscribe_to_ln_receive_updates`

fedimint-client/src/lib.rs
```diff
@@ -166,15 +167,15 @@ pub trait IGlobalClientContext: Debug + MaybeSend + MaybeSync + 'static {
...
-    ) -> TransactionId;
+    ) -> (TransactionId, Option<OutPoint>);
 
     /// This function is mostly meant for internal use, you are probably looking
     /// for [`DynGlobalClientContext::fund_output`].
     async fn fund_output_dyn(
         &self,
         dbtx: &mut ClientSMDatabaseTransaction<'_, '_>,
         output: InstancelessDynClientOutput,
-    ) -> anyhow::Result<TransactionId>;
```
elsirion, line 177: Nit: we should document what the `OutPoint` represents

fedimint-client/src/module/mod.rs
```diff
@@ -287,6 +288,12 @@ pub trait PrimaryClientModule: ClientModule {
         operation_id: OperationId,
         amount: Amount,
     ) -> ClientOutput<<Self::Common as ModuleCommon>::Output, Self::States>;
+
+    async fn await_primary_module_output_finalized(
+        &self,
+        operation_id: OperationId,
+        out_point: OutPoint,
+    ) -> anyhow::Result<()>;
```
elsirion, line 296: I wonder if this should return the amount received? Is that useful or a foot-gun?

elsirion (approved):
API looks good, didn't look at the state machine changes too closely, can do later.

fedimint-client/src/module/mod.rs
m1sterc001guy: I think that would be fine

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -319,42 +336,50 @@ impl LightningClientExt for Client {
...
+            yield LnPayState::Funded;
+
+            match payment_success.await {
+                Ok(preimage) => {
+                    if let Some(change) = change_outpoint {
+                        yield LnPayState::AwaitingChange;
+                        match self.await_primary_module_output_finalized(operation_id, change).await {
+                            Ok(_) => {}
+                            Err(_) => {
+                                yield LnPayState::Failed;
```
elsirion, line 356: There is a nuance here that a malicious federation could block the issuance of e-cash while the contract is funded and the payment succeeds, but probably not a priority to fix.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -319,42 +336,50 @@ impl LightningClientExt for Client {
...
                 }
+                Err(LightningPayError::Refundable(block_height, error)) => {
+                    yield LnPayState::WaitingForRefund{ block_height, gateway_error: error.clone() };
+
+                    if let Ok(refund_txid) = refund_success.await {
+                        // need to await primary module to get refund
+                        if self.await_primary_module_output_finalized(operation_id, OutPoint{ txid: refund_txid, out_idx: 0}).await.is_ok() {
+                            yield LnPayState::Refunded { gateway_error: error };
+                            return;
+                        }
```
elsirion, line 373: I wonder if we could get rid of the `LightningPayStates::Refunded` state now and instead only rely on `await_primary_module_output_finalized` to determine the high-level operation state (for the output to be finalized the tx has to be accepted). Could be a follow-up.

elsirion (approved):
Great work! Left some ideas for folow-ups.

PR #2510, opened 2023-05-13
Integrate backup restore into `ClientBuilder`

elsirion:


modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -614,16 +614,14 @@ impl MintRestoreInProgressState {
...
-        epoch: u64,
         peer_id: PeerId,
         item: &ConsensusItem,
         processed_txs: &mut HashSet<TransactionId>,
         rejected_txs: &BTreeSet<TransactionId>,
         secret: &DerivableSecret,
     ) {
         debug!(target: LOG_ECASH_RECOVERY, ?item, "handling consensus item");
-        assert!(epoch == self.next_epoch);
-        self.next_epoch = epoch;
```
elsirion: This assert made it crash when there wasn't a backup, also seemed quite redundant with the following line.

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -226,13 +226,13 @@ impl MintRestoreInProgressState {
         let mut epoch_stream = Self::fetch_epochs_stream(api, epoch_pk, decoders, epoch_range);
         while let Some((epoch, epoch_history)) = epoch_stream.next().await {
             assert_eq!(epoch_history.outcome.epoch, epoch);
+            self.next_epoch = epoch + 1;
```
elsirion, line 229: Without this the recovery would never end and keep looping in the last epoch.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1001,7 +1000,7 @@ impl MintClientModule {
                 .find_by_prefix(&NoteKeyPrefix)
                 .await
                 .map(|(key, note)| (key.amount, note))
-                .collect::<BTreeMap<_, _>>()
+                .collect::<Vec<_>>()
```
elsirion, line 1003: We were only storing one e-cash not per denomination in the epoch.

dpc:
Landed in  #2528

PR #2511, opened 2023-05-13
feature: start on the ln send test

jkitman:
Starts the ln test, but we can't easily finish it until the gateway is using the new client.

### DIFF
--- a/modules/fedimint-ln-tests/Cargo.toml
+++ b/modules/fedimint-ln-tests/Cargo.toml
@@ -30,7 +30,7 @@ tracing = "0.1.37"
 # Remove once we modularize the gw / ln server
 fedimint-wallet-common = { path = "../fedimint-wallet-common" }
 fedimint-wallet-client = { path = "../fedimint-wallet-client" }
-fedimint-wallet-server = { path = "../fedimint-wallet-server", features = ["native"] }
+fedimint-wallet-server = { path = "../fedimint-wallet-server" }
 fedimint-wallet-tests = { path = "../fedimint-wallet-tests" }
 fedimint-mint-common = { path = "../fedimint-mint-common" }
 fedimint-mint-client = { path = "../fedimint-mint-client" }
--- a/modules/fedimint-ln-tests/Cargo.toml
+++ b/modules/fedimint-ln-tests/Cargo.toml
@@ -12,10 +12,7 @@ path = "tests/tests.rs"
 
 [dependencies]
 anyhow = "1.0.66"
-fedimint-wallet-common = { path = "../fedimint-wallet-common" }
-fedimint-wallet-client = { path = "../fedimint-wallet-client" }
-fedimint-wallet-server = { path = "../fedimint-wallet-server", features = ["native"] }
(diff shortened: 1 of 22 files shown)


modules/fedimint-ln-tests/Cargo.toml
```diff
@@ -27,4 +24,14 @@ fedimint-client = { path = "../../fedimint-client" }
...
 tokio = { version = "1.26.0", features = ["sync"] }
+tracing = "0.1.37"
+# Remove once we modularize the gw / ln server
+fedimint-wallet-common = { path = "../fedimint-wallet-common" }
+fedimint-wallet-client = { path = "../fedimint-wallet-client" }
+fedimint-wallet-server = { path = "../fedimint-wallet-server", features = ["native"] }
+fedimint-wallet-tests = { path = "../fedimint-wallet-tests" }
+fedimint-mint-common = { path = "../fedimint-mint-common" }
+fedimint-mint-client = { path = "../fedimint-mint-client" }
+fedimint-mint-server = { path = "../fedimint-mint-server" }
```
elsirion: Then these would be replaced with `brrrrtimint` aka the dummy module?

modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -13,25 +21,67 @@ fn fixtures() -> Fixtures {
...
+    assert_eq!(next(sub1).await, LnReceiveState::Created);
+    assert_matches!(next(sub1).await, LnReceiveState::WaitingForPayment { .. });
+
+    let (op, _txid) = client2.pay_bolt11_invoice(fed_id, invoice).await.unwrap();
+    let sub2 = &mut client2.subscribe_ln_pay_updates(op).await.unwrap();
+    assert_eq!(next(sub2).await, LnPayState::Created);
+    // TODO: Finish after gw moves from legacy client
+    // assert_eq!(next(sub2).await, LnPayState::Funded);
+    // assert_matches!(next(sub2).await, LnPayState::Success{..});
+    // assert_eq!(next(sub1).await, LnReceiveState::Funded);
```
elsirion, line 84: What is the blocker here for testing that the *user client* does the right thing? That we don't know if the GW is done yet?

elsirion (approved):

modules/fedimint-ln-tests/tests/tests.rs
jkitman: Problem is the gateway needs funds in the form of legacy ecash which isn't so easy to create, I'd have to mix inputs from the new dummy client and outputs from the legacy mint client.

modules/fedimint-ln-tests/Cargo.toml
jkitman: Yes, ideally we will only test modules independently with the dummy to demonstrate they have no dependencies on other modules.

The tests in the gateway crate will probably need to use the wallet and ln modules though to test the deposit RPC.

modules/fedimint-ln-tests/tests/tests.rs
m1sterc001guy: I had the same confusion :)

PR #2513, opened 2023-05-13
Allow variable number of servers in `run-ui`

### LINKED ISSUE #2512: `run-ui` does not respect fed size argument
1. `run-ui.sh` has tails [hard coded for 0 and 1](https://github.com/fedimint/fedimint/blob/master/scripts/run-ui.sh#L10-L13), so it doesn't tail more than 2 federations' logs
2. `run_ui` iters over a [hard coded 2 fedimintds](https://github.com/fedimint/fedimint/blob/master/devimint/src/main.rs#L897), so only 2 start

wbobeirne:
Closes #2512

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -893,9 +893,9 @@ async fn run_ui(
     process_mgr: &ProcessManager,
     task_group: &TaskGroup,
     kind: &RunUiKind,
-    fed_size: usize,
 ) -> Result<()> {
     let bitcoind = Bitcoind::new(process_mgr).await?;
+    let fed_size = process_mgr.globals.FM_FED_SIZE;
     // don't drop fedimintds
     let _fedimintds = futures::future::try_join_all((0..fed_size).map(|peer_id| {
         let bitcoind = bitcoind.clone();
@@ -937,7 +937,7 @@ async fn run_ui(
 use std::fmt::Write;
 
 async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
-    let globals = vars::Global::new(&arg.test_dir, 4).await?;
+    let globals = vars::Global::new(&arg.test_dir, arg.fed_size).await?;
     let log_file = fs::File::create(globals.FM_LOGS_DIR.join("devimint.log"))
         .await?
         .into_std()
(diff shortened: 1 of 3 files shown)


scripts/run-ui.sh
```diff
@@ -4,12 +4,15 @@ export FM_BITCOIN_NETWORK="regtest"
 export FM_UI_KIND=${1:-"old"}
 export FM_FED_SIZE=${2:-2}
 export FM_FED_NAME=${3:-"Cypherpunk Federation"}
+export RUST_BACKTRACE="full"
```
Maan2003, line 7: why full? RUST_BACKTRACE=1 has more readable backtrace.

devimint/src/main.rs
```diff
@@ -974,8 +977,9 @@ async fn main() -> Result<()> {
             task_group.make_handle().make_shutdown_rx().await.await?;
         }
         Cmd::RunUi(kind) => {
+            let fed_size = args.common.fed_size;
             let (process_mgr, task_group) = setup(args.common).await?;
-            run_ui(&process_mgr, &task_group, &kind).await?
+            run_ui(&process_mgr, &task_group, &kind, fed_size).await?
```
Maan2003, line 982: not blocking, 
but you can also use `process_mgr.globals.FM_FED_SIZE` this is set in `setup` function `var::Global::new(..., 4)`

scripts/run-ui.sh
wbobeirne: Whoops, left this in from debugging a panic I ran into while running this. I'll remove.

devimint/src/main.rs
wbobeirne: Got it. New to the Rust side so appreciate the pointers. I just looked for somewhere else where we were accessing an env var and tried to copy the pattern.

Down in globals, it looks to me like it's hard coded to 4? Is there something I'm missing where it's reading in the env var? Or are you recommending I update that `4` to be `arg.fed_size` instead?

Maan2003: yes, that 4 should change to `arg.fed_size`

Maan2003: you can also try replacing other 4s with globals.FM_FED_SIZE  (`rg '\b4\b' devimint/src`)
But can also be done in a followup.

scripts/run-ui.sh
EthnTuttle:
```suggestion
```
Just commenting so I get alerts when it's merged.

devimint/src/main.rs
wbobeirne: Went ahead and did it here. Thanks for the note!

Maan2003 (approved):

Issue #2516, opened 2023-05-14
Test gateway re-registration


justinmoon:
The lightning gateway is supposed to re-register itself every 10 minutes. @okjodom this would be a great thing to cover in your new gateway tests. Should probably make `GW_ANNOUNCEMENT_TTL` a variable passed in to `gatewayd` so we can set the value much lower than 10 minutes for a test.

okjodom:
Covered by `test_gateway_register_with_federation`

PR #2517, opened 2023-05-14
feat: fix gateway registration

justinmoon:
Current code fails because it re-registers with a `valid_until` value that doesn't get updated at all.

@okjodom this would be a great thing to cover in your new gateway tests. Should probably make `GW_ANNOUNCEMENT_TTL` a variable passed in to `gatewayd`. 

This currently doesn't have any test coverage so you have to validate the fix manually (made [issue to add coverage](https://github.com/fedimint/fedimint/issues/2516)). To validate the bug, set `GW_ANNOUNCEMENT_TTL` to 1 second on master and run tmuxinator. Run `gateway-cli list-gateways` and there won't be any. Do the same on this branch and there should be 2.

### DIFF
--- a/gateway/ln-gateway/src/actor.rs
+++ b/gateway/ln-gateway/src/actor.rs
@@ -155,7 +155,6 @@ impl GatewayActor {
         let registration = register_client
             .config()
             .to_gateway_registration_info(route_hints.clone(), GW_ANNOUNCEMENT_TTL);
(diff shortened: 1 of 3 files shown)


okjodom (approved):

elsirion:
I applied a fix for the `can_switch_active_gateway` test since that now got confused about different expiry times.

PR #2518, opened 2023-05-14
feat: mprocs setup for guardian-ui dev

EthnTuttle:
`just run-ui` brings up mprocs with a 2 instance federation via `run-ui.sh` and 2 instances of the Guardian UI

### DIFF
--- a/misc/mprocs-ui.yaml
+++ b/misc/mprocs-ui.yaml
@@ -1,0 +1,17 @@
+procs:
+  run-ui-federation:
+    shell: ./scripts/run-ui.sh new
+  coordinator:
+    cwd: fedimint-ui/apps/guardian-ui/
+    shell: yarn dev
+    env:
+      PORT: "3000"
+      REACT_APP_FM_CONFIG_API: ws://127.0.0.1:18174
+      BROWSER: none
+  follower: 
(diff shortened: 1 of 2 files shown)


elsirion (approved):
Tested, works for me, really cool!

It could use some docs/help in case a user is "holding it wrong" though:

* Needs to be run in `nix develop .#fedimint-ui` shell (can we warn if that's not the case, e.g. `yarn` isn't available)
* `yarn install` needs to be run before (is there a way to check and warn the user if that hasn't been done?)

But that could be a follow-up PR :)

misc/mprocs-ui.yaml
```diff
@@ -0,0 +1,17 @@
+procs:
+  run-ui-federation:
+    shell: ./scripts/run-ui.sh new
+  coordinator:
+    cwd: fedimint-ui/apps/guardian-ui/
+    shell: yarn dev
```
justinmoon, line 6: need yarn install

EthnTuttle:
Do 4 guardians minimum.
Or pass in # of guardians.

elsirion:
I think just having a check in `./scripts/run-ui.sh` and warning the user/pointing into the right direction in the first mprocs window would help a lot

Issue #2474, opened 2023-05-10
Select bitcoin node to use in Guardian Setup


garyray-k:
Enable a Guardian to input a node for their instance to use:

- [ ] input node URL
- [ ] confirm node connection w/ feedback

EthnTuttle:
Related: https://github.com/fedimint/fedimint/pull/2525

justinmoon:
Call: ready to implement this

PR #2519, opened 2023-05-14
refactor: clean up bitcoin rpc

jkitman:
Begin work for #2474

- Standardize our RPC interface (don't expose rpc type to callers)
- Replace env vars with a registry (also enable registering test bitcoind rpc)
- Prepare for eliminating use of env vars in module initialization


okjodom (approved):
✅

devimint/src/vars.rs
```diff
@@ -120,7 +120,8 @@ declare_vars! {
 
         // fedimint config variables
         FM_TEST_BITCOIND_RPC: String = "http://bitcoin:bitcoin@127.0.0.1:18443";
```
EthnTuttle, line 122: I saw this replaced elsewhere, is there a reason it's still around?

jkitman: Yes, in some of our integration tests we have a different RPC interface that communicates directly with bitcoin RPC to perform actions such as mining new blocks.  Since we only implemented that interface for bitcoind we need to pass in the bitcoind URL for tests only.

PR #2520, opened 2023-05-14
fix(devimint): use correct protocol in logging

EthnTuttle:
If it's the new UI, then we want ws:// instead of http://
The logging would be confusing for anyone unfamiliar with this nuance.

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -1045,7 +1045,7 @@ async fn run_ui(
                 Ok(TcpStream::connect(server_addr).await.is_ok())
             })
             .await?;
-            info!(LOG_DEVIMINT, "Started ui/api on http://{server_addr}");
+
             anyhow::Ok(fm)
         }
     }))
--- a/fedimint-server/src/lib.rs
+++ b/fedimint-server/src/lib.rs
@@ -197,6 +197,7 @@ impl FedimintServer {
             .expect("Could not start API server")
             .start(module)
             .expect("Could not start API server");
+        info!(target: LOG_NET_API, "Starting api on ws://{api_bind}");
 
         FedimintApiHandler { handle, runtime }
     }
--- a/fedimintd/src/fedimintd.rs
+++ b/fedimintd/src/fedimintd.rs
@@ -247,6 +247,7 @@ async fn run(
                 .await;
             })
             .await;
+        info!(target: "old-ui", "Started ui on http://{listen_ui}");
 
         // If federation configs (e.g. local.json) missing, wait for admin UI to report
         // DKG completion


devimint/src/main.rs
```diff
@@ -1040,7 +1040,14 @@ async fn run_ui(
                 Ok(TcpStream::connect(server_addr).await.is_ok())
             })
             .await?;
-            info!(LOG_DEVIMINT, "Started ui/api on http://{server_addr}");
+            let url_protocol = match kind {
+                RunUiKind::Old => "http://",
+                RunUiKind::New => "ws://",
```
dpc, line 1045: Don't we want to always use `ws://` or `wss://` ?

EthnTuttle: Good question. I would think `wss` would be the preferred protocol but had not seen it used a lot in the code base so wasn't sure if the API is setup for it. TBH, I wouldn't know where to check if this is the case but will look into it.

elsirion: `wss` only really makes sense in production with a SSL-terminating proxy in front. There should always be two addresses that are configurable:

1. A bind address: That's the one where the server listens on, that's an IP+port and I guess what you are logging here
2. The address the server will be reachable under: that's a protocol+domain/IP+optional port

For this logging we should stick to the unencrypted variant (i.e. `ws://`) since we leave encryption to the deployment (e.g. a NGINX in front).

devimint/src/main.rs
```diff
@@ -1040,7 +1040,14 @@ async fn run_ui(
...
+            };
```
elsirion, line 1046: Could this log be moved inside the UI server impl, close to where the server is actually started? That would make more sense imo since it's more apparent which protocol the server started is running in that context and you wouldn't need to match here.

devimint/src/main.rs
elsirion: > Don't we want to always use ws:// or wss:// ?

As soon as the old UI is removed, yes, but for now it still starts a HTTP server.

devimint/src/main.rs
EthnTuttle: Please see updated code.

EthnTuttle:
You can separately `just run-ui` (from within `nix develop .#fedimint-ui`!!!) and `./scripts/run-ui.sh` to see the updated changes reflected in the logs.

elsirion (approved):

elsirion:
I took the liberty to fix the semgrep lint error:

```
Blocking Findings:

  fedimint-server/src/lib.rs 
     use-log-targets
        Log statements should have static target defined to aid filtering

        200┆ info!(LOG_NET_API, "Starting api on ws://{api_bind}");

Blocking Rules Fired:
   use-log-targets
```

elsirion (approved):

PR #2521, opened 2023-05-15
fix: add `nodejs` to ui dev shell

elsirion:
Otherwise `yarn install` fails, I suppose everyone working on the UI had `nodejs` installed outside nix.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -456,7 +456,7 @@
 
               fedimint-ui = pkgs.mkShell (shellCommonNative
                 // {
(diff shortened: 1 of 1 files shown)


Maan2003 (approved):
yarn install and yarn build works in `nix develop .#fedimint-ui --ignore-enviroment` now.

elsirion:
> yarn install and yarn build works in `nix develop .#fedimint-ui --ignore-enviroment` now.

Could we easily test that in CI? It would seem so unless I'm missing something.

Maan2003:
yes, should be doable. just check exit code of `nix develop .#ui -c 'yarn install && yarn build'`

PR #2529, opened 2023-05-15
feat(client-ng): make client root secret encoding flexible using strategy pattern

dpc:
Rebase #2501

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -89,8 +89,7 @@ use fedimint_core::{
 pub use fedimint_derive_secret as derivable_secret;
 use fedimint_derive_secret::DerivableSecret;
 use futures::StreamExt;
-use rand::distributions::{Distribution, Standard};
-use rand::{thread_rng, Rng};
+use rand::thread_rng;
 use secp256k1_zkp::Secp256k1;
 use secret::DeriveableSecretClientExt;
 use serde::de::DeserializeOwned;
@@ -107,6 +106,7 @@ use crate::module::gen::{
 use crate::module::{
     ClientModule, ClientModuleRegistry, DynPrimaryClientModule, IClientModule, StateGenerator,
 };
+use crate::secret::RootSecretStrategy;
 use crate::sm::executor::{
     ActiveOperationStateKeyPrefix, ContextGen, InactiveOperationStateKeyPrefix,
 };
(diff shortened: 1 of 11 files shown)


justinmoon (approved):
LGTM

fedimint-client/src/lib.rs
```diff
@@ -684,6 +684,15 @@ impl Client {
             .find(|(_, kind, _module)| *kind == module_kind)
             .map(|(instance_id, _, _)| instance_id)
     }
+
+    /// Returns the data from which the client's root secret is derived (e.g.
+    /// BIP39 seed phrase struct).
+    pub async fn root_secret_encoding<S>(&self) -> S::Encoding
```
elsirion, line 690: Btw, I'm very open to a better name

elsirion (commented):
Looks good

fedimint-client/src/lib.rs
dpc: @dpc "root secret encoding" was a bit confusing to me and maybe "encoded secret" or "serialized secret" would be more clear?

elsirion: Well, it's not even encoded really, it's in a form where it can be either converted to a `DerivableSecret` or presumably encoded in a custom way. When you e.g. look at the bip39 impl we have the parsed mnemonic which is the "encoding".

dpc: @elsirion I see. Maybe "secret key input" or "secret key source"?

I would rename the "strategy" too. "strategy" is the name of the pattern which is useful to describe what it's doing, but doesn't have to be the name of the trait. So `s/RootSecretStrategy/RootSecretInput/g`?

elsirion: The problem is that the struct on which the trait is implemented isn't the "root secret input" itself, but rather a ZST that represents the way some other (associated) type has to be encoded. So in a way the ZST is the strategy.

dpc: @elsirion Could it be changed to be "root secret input"?

```rust
trait RootSecretInput : Encodable, Decodable {
  fn to_root_secret() -> DeriveableSecret;
  fn random() -> Self;
}
```

I might be missing something, but the only reason to prefer a ZST over trait over the data itself, is if we needed to store somewhere the code without the actual data.

elsirion: It's a choice between needing to wrap everything in newtypes and the ZST hack. All encoding except `PlainRootSecretStrategy` will be implemented by external crates on types of yet other crates (we can't expect `bip39` to be aware of fedimint, it needs a `fedimint-bip39` connector crate). If they needed to impl that trait directly that wouldn't work.

dpc: @elsirion Is

```rust
struct Bip39Strategy;

impl RootSecretInput {
   type Encoding = bip39::Seedphrase;  

  // ...
}
```

 any better than 

```rust
struct Bip39SecretInput(bip39::Seedphrase);

impl SecretInput {

  // ..
}
```

?

ZST seems about the same, except with the data unnaturally(?) outside.

PR #2147, opened 2023-04-06
Migrate cli-test to Rust

GetPsyched:
Progress in #2101 (2a)

PR #2530, opened 2023-05-15
Fix flaky peg-out test

### LINKED ISSUE #2341: Intermittant cli-test error with string comparison
I have twice seen this error when running `scripts/cli-test.sh`:

```
Error: Peg-out address received 0.000005 BTC, expected 0.00000500
```

The two numbers are equal, but padded differently with zeros.

elsirion:
Fixes #2341. I think the logic is the wrong way around and we test the peg-out not succeeding, which normally works because it's too slow. But if it happens faster than the check code is hit (due to scheduling) then this test fails because the peg-out succeeded.

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -654,14 +654,16 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         .await?
         .as_u64()
         .unwrap();
+
+    let expected_diff = 3030;
     anyhow::ensure!(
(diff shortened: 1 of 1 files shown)


elsirion:
Since this might have been a fundamental misunderstanding of the `anyhow::ensure!()` macro we should probably re-review the entirety of https://github.com/fedimint/fedimint/pull/2147.

Maan2003 (approved):

PR #2534, opened 2023-05-16
Run 4 Servers with `just run-ui`

EthnTuttle:
Builds on: https://github.com/fedimint/fedimint/pull/2513 by @wbobeirne 

and checks for nix shell before running. Also starts 4 fedimintd's and UI's for `just run-ui`

### DIFF
--- a/fedimint-ui/apps/guardian-ui/package.json
+++ b/fedimint-ui/apps/guardian-ui/package.json
@@ -3,13 +3,14 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "react-scripts start",
-    "dev": "react-scripts start",
+    "start": "yarn isNix && react-scripts start",
+    "dev": "yarn isNix && react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint \"src/**/*.ts*\"",
-    "clean": "rm -rf build"
+    "clean": "rm -rf build",
+    "isNix": "if [[ -z \"${IN_NIX_SHELL:-}\" ]]; then echo 'It is recommended to run this command from a Nix dev shell. Use 'nix develop .#fedimint-ui' first'; sleep 3; fi;"
   },
   "dependencies": {
     "@chakra-ui/react": "^2.4.5",
--- a/fedimint-ui/apps/gateway-ui/package.json
+++ b/fedimint-ui/apps/gateway-ui/package.json
@@ -3,13 +3,14 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "react-scripts start",
-    "dev": "react-scripts start",
+    "start": "yarn isNix && react-scripts start",
+    "dev": "yarn isNix && react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint \"src/**/*.ts*\"",
-    "clean": "rm -rf build"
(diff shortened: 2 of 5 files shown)


fedimint-ui/apps/guardian-ui/package.json
```diff
@@ -3,13 +3,14 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "react-scripts start",
-    "dev": "react-scripts start",
+    "start": "yarn isNix && react-scripts start",
```
okjodom, line 6: Two objections with `isNix` check.

- If we have yarn available, do we still need enforce nix shell with `isNix` check?! At present, we only use nix to provide yarn as a standard dev tool, so we don't require ui devs to install yarn globally before they can start the UIs.

- If we still want to enforce nix shell, we should be careful to do so in dev only, but not in prod scenarios

okjodom:
IMO we should just remove `isNix` check in the npm scripts and change would be good to go 🥇

devimint/src/main.rs
```diff
@@ -1010,10 +1012,11 @@ async fn run_ui(
     process_mgr: &ProcessManager,
     task_group: &TaskGroup,
     kind: &RunUiKind,
+    fed_size: usize,
 ) -> Result<()> {
     let bitcoind = Bitcoind::new(process_mgr).await?;
     // don't drop fedimintds
-    let _fedimintds = futures::future::try_join_all((0..2).map(|peer_id| {
+    let _fedimintds = futures::future::try_join_all((0..fed_size).map(|peer_id| {
```
Maan2003, line 1019:
```suggestion
    let _fedimintds = futures::future::try_join_all((0..process_mgr.global.FM_FED_SIZE).map(|peer_id| {
```

fedimint-ui/apps/guardian-ui/package.json
elsirion: It might be related to my comment here: https://github.com/fedimint/fedimint/pull/2518#pullrequestreview-1426268946

I agree we shouldn't just refuse to work if all the tooling is present either way, it would still be nice to point non-ui devs that just want to test if they broke anything the right way. It's quite hard to figure out how to run `just run-ui` right now, it doesn't *just work* :laughing:

EthnTuttle: Yes, I had the idea from review #2518.

>  If we have yarn available, do we still need enforce nix shell with isNix check?! At present, we only use nix to provide yarn as a standard dev tool, so we don't require ui devs to install yarn globally before they can start the UIs.

`isNix` does not enforce anything. It checks for the ENV VAR and outputs a message if it's not present. I'm happy to remove it if that is conflicting but there is no enforcement of nix with this.

> If we still want to enforce nix shell, we should be careful to do so in dev only, but not in prod scenarios

Similar to previous, but this leads to another question around how we envision these web apps being deployed. I don't believe we would want to run npm commands to deploy? Usually, you run [build commands to create a static site and then serve from a web server](https://create-react-app.dev/docs/deployment/). `yarn build` is unaffected by `isNix`.

elsirion: > `isNix` does not enforce anything.

That sounds like exactly what we want then!

elsirion:
Accepting patches in the GitHub UI doesn't work very well because it often leads to rustfmt complaining. It's probably best to pull the branch, run rustfmt, `git commit --amend` and then force push. Otherwise the PR looks good to go after clearing up the `isNix` confusion :)

justinmoon (approved):

PR #2535, opened 2023-05-16
fix(client-ng): ignore modules of unknown kind

elsirion:
For a client it is ok to not understand all modules supported by a server, it can still interact with the ones it knows.


justinmoon (approved):

justinmoon:
Needs rebase

PR #2536, opened 2023-05-16
Cache client operation outcomes

elsirion:
Subscribing to operation updates is expensive, if we just want to know what the outcome of a finished operation was we can now read it from the operation log entry.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -73,6 +73,7 @@ use std::io::{Error, Read, Write};
 use std::sync::Arc;
 
 use anyhow::{anyhow, bail};
+use async_stream::stream;
 use fedimint_core::api::{DynFederationApi, IFederationApi, WsFederationApi};
 use fedimint_core::config::{ClientConfig, FederationId, ModuleGenRegistry};
 use fedimint_core::core::{DynInput, DynOutput, IInput, IOutput, ModuleInstanceId, ModuleKind};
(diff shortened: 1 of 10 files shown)


fedimint-client/src/lib.rs
```diff
@@ -624,12 +638,50 @@ impl Client {
...
+        let outcome_json = serde_json::to_value(outcome).expect("Outcome is not serializable");
+
+        let mut dbtx = db.begin_transaction().await;
+        let mut operation = Client::get_operation_inner(&mut dbtx, operation_id)
             .await
+            .expect("Operation exists");
+        operation.outcome = Some(outcome_json);
+        dbtx.insert_entry(&OperationLogKey { operation_id }, &operation)
+            .await;
+        dbtx.commit_tx_result().await.expect("commit failed");
```
m1sterc001guy, line 668: Can you use `commit_tx` instead of `commit_tx_result`? Can remove the expect that way

fedimint-client/src/lib.rs
```diff
@@ -624,12 +638,50 @@ impl Client {
...
+
+        Ok(())
+    }
+
+    /// Tries to set the outcome of an operation, but only logs an error if it
+    /// fails and does not return it. Since the outcome can always be recomputed
+    /// from na update stream failing to save it isn't a problem in cases where
```
m1sterc001guy, line 675: is na a typo?

fedimint-client/src/lib.rs
```diff
@@ -624,12 +638,50 @@ impl Client {
...
+    /// we do this merely for caching.
+    pub async fn opportunistically_set_operation_outcome(
```
m1sterc001guy, line 677: "optimistically" might be a better description

fedimint-client/src/lib.rs
elsirion: yes, should be `… an …`

m1sterc001guy (approved):

PR #2537, opened 2023-05-16
Primary module balance fetch&subscribe

elsirion:


jkitman (dismissed):
Nice making this part of the primary module

fedimint-client/src/lib.rs
```diff
@@ -768,6 +768,33 @@ impl Client {
     pub async fn get_config(&self) -> &ClientConfig {
         &self.inner.config
     }
+
+    /// Balance available to the client for spending
+    pub async fn get_balance(&self) -> Amount {
```
jkitman, line 773: If this is available on the `Client` can we also remove `MintClientExt::total_amount`?

fedimint-cli/src/ng.rs
```diff
@@ -134,7 +134,7 @@ pub async fn handle_ng_command(
         }
         ClientNg::Spend { amount } => {
             let (operation, notes) = client
-                .spend_notes(amount, Duration::from_secs(30), ())
+                .spend_notes(amount, Duration::from_secs(3600), ())
```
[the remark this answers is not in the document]
elsirion: It doesn't, but since we never reissue these notes I wanted to make sure that they don't get re-claimed in the middle of some other test.

fedimint-client/src/lib.rs
elsirion: Thx, fixed :)

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -694,6 +678,31 @@ impl PrimaryClientModule for MintClientModule {
...
+                    match state {
+                        MintClientStateMachines::Output(MintOutputStateMachine {
+                            state: MintOutputStates::Succeeded(_),
+                            ..
+                        }) => Some(()),
+                        MintClientStateMachines::Input(MintInputStateMachine {
+                            state: MintInputStates::Created(_),
+                            ..
+                        }) => Some(()),
+                        _ => None,
```
elsirion, line 701: TODO: Trigger on OOB spend

PR #2542, opened 2023-05-16
 feat: simple cli test for ng client backup&restore 

dpc:


modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -180,6 +193,12 @@ impl MintRestoreInProgressState {
                             }
                         }
 
+                        debug!(
+                            target: LOG_CLIENT_RECOVERY_MINT,
+                            len = finalized.unconfirmed_notes.len(),
+                            "Restoring unconfigured notes state machins"
```
douglaz, line 199:
```suggestion
                            "Restoring unconfigured notes state machines"
```

elsirion:
Tests still failing, I'd also recommend putting an `info!` log before every section so one roughly knows where things go wrong. The error looked like an e-cash reissue failure at first glance because that's not the case yet.

Screenshot 2023-05-25 at 21-15-55 feat simple cli test for ng client backup restore · fedimint_fedimint@dcdb527

dpc:
@elsirion Changing to a fixed operation id for mint restoration broke subsequent restoration. First one works. Following ones don't (or maybe they will, but waiting for them finishes prematurely).

Looking at it, I don't think hardcoding the ID is a right approach. `ClientModule::restore` could return `Vec<OperationId>`, and then the restore cli could wait for all of them, or something like that. Special casing mint restoration in the general client doesn't feel right.

dpc:
@elsirion I wanted something like 3cc89e3fa6c607e4f4a87016857009985937c050 , but I don't know if we have any way to get notifications for subscriptions for all the modules?

elsirion:
> @elsirion Changing to a fixed operation id for mint restoration broke subsequent restoration. First one works. Following ones don't (or maybe they will, but waiting for them finishes prematurely).

Doesn't that just mean that we should probably completely clear our DB before restoring? Restoring with old state still being around sounds like a bad idea anyway.

elsirion:
Completely wiping the DB sounds reasonable, we don't really want to get into conflict resolution between old and new state. How to best do that is tbd imo (add a clear db command to API, let the user figure it out, …)

dpc:
Moved to #2655 2655

PR #2547, opened 2023-05-17
fix(ui): update guardian api types

okjodom:
- Updates api transalation in guardian setup ui after #2540

### DIFF
--- a/fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
+++ b/fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
@@ -33,8 +33,7 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
       numPeers: stateNumPeers,
     },
     api,
-    submitFollowerConfiguration,
-    submitHostConfiguration,
+    submitConfiguration,
   } = useGuardianContext();
   const theme = useTheme();
   const isHost = role === GuardianRole.Host;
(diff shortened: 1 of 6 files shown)


okjodom:
@jkitman, after these updates, the follower guardian never gets consensus config gen params, even though the leader successfully sets and retrieves these configs

image

fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -199,44 +244,59 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
...
+            <>
+              <FormControl>
+                <FormLabel>Block confirmations</FormLabel>
+                <Input
+                  type='number'
+                  min={1}
+                  value={blockConfirmations}
+                  onChange={(ev) => {
+                    const value = ev.currentTarget.value;
+                    isValidNumber(value) && setBlockConfirmations(value);
```
EthnTuttle, line 260: Should `setBlockConfirmations()` run only if `isValidNumber()` is true? The `&&` here is just a boolean AND and doesn't function like in bash script.

fedimint-ui/apps/guardian-ui/src/types.tsx
```diff
@@ -47,31 +47,58 @@ export interface Peer {
...
+}
+
 export type PeerHashMap = Record<string, string>;
 
-export type LnFedimintModule = ['ln', null];
-export type MintFedimintModule = ['mint', { mint_amounts: number[] }];
+export type LnFedimintModule = [
+  'ln',
+  {
+    consensus?: object;
```
EthnTuttle, line 60: Do we have plans to make these more concrete types? I imagine this could be done when the modules are no longer hard coded, etc.

EthnTuttle (approved):
Was able to get the federation up and running again. Only non-blocking questions.
Nice job, thnx for repairing the feature.

PR #2554, opened 2023-05-18
Break the dependency of LN on Wallet module

### LINKED ISSUE #1043: Allow caller-provided DB tx in `ModuleInterconnect`
Currently interconnect calls are not guaranteed to be consistent. Fortunately their only usage (checking consensus block height from the LN module) is not affected by that shortcoming. For more general use cases this might pose a consistency problem.

I did some initial hacking (1437981362b191ddd2dec9ee94e7ed79aa871107), but it appears we need to fundamentally fix our lifetimes in the `ServerModulePlugin` trait before this can progress further.

### LINKED ISSUE #1801: Make interconnect available in `await_consensus_proposal` and possibly other such fns

### LINKED ISSUE #2492: Remove ModuleInterconnect
We only use it for LN getting block height, but it could have it's own version of the consensus item since they don't need to be in sync.

jkitman:
Break the dependency of LN on Wallet module

Fixes #2492

Also we should close #1801 and close  #1043


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -306,16 +329,24 @@ impl ServerModule for Lightning {
         consensus_items: Vec<(PeerId, LightningConsensusItem)>,
         _consensus_peers: &BTreeSet<PeerId>,
     ) -> Vec<PeerId> {
-        for (peer, decryption_share) in consensus_items.into_iter() {
+        let mut peer_heights = dbtx.get_value(&BlockHeightKey).await.unwrap_or_default();
```
EthnTuttle, line 332: What are the implications of peers having a different BlockHeight? I would imagine this would be a short lived situation?

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1167,6 +1201,7 @@ mod fedimint_migration_tests {
                             "validate_migrations was not able to read any ProposeDecryptionShares"
                         );
                         }
+                        DbKeyPrefix::BlockHeight => {}
```
EthnTuttle, line 1204: Should there be a test here?

EthnTuttle (approved):
This doesn't break the UI so LGTM from that perspective.

modules/fedimint-ln-server/src/lib.rs
jkitman: We use the median of block height between peers which works if enough honest peers have contributed.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -277,27 +292,35 @@ impl ServerModule for Lightning {
     }
 
     async fn await_consensus_proposal(&self, dbtx: &mut ModuleDatabaseTransaction<'_>) {
-        if !self.consensus_proposal(dbtx).await.forces_new_epoch() {
-            std::future::pending().await
+        while !self.consensus_proposal(dbtx).await.forces_new_epoch() {
+            sleep(Duration::from_millis(1000)).await;
```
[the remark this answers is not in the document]
jkitman: We do the same in the wallet module, it's because new items can emerge spontaneously from bitcoin rpc's height changing.  Eventually we will eliminate this kind of logic with the new consensus algo.

modules/fedimint-ln-common/src/config.rs
```diff
@@ -7,10 +8,31 @@ use threshold_crypto::serde_impl::SerdeSecret;
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct LightningGenParams {
-    pub local: EmptyGenParams,
-    pub consensus: EmptyGenParams,
+    pub local: LightningGenParamsLocal,
+    pub consensus: LightningGenParamsConsensus,
+}
+
+impl LightningGenParams {
+    pub fn regtest(bitcoin_rpc: BitcoinRpcConfig) -> Self {
```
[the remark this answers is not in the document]
jkitman: Just for testing.  In the UI we can pass in whatever network the user configures.

PR #2557, opened 2023-05-19
feature: split out global and module api traits

### LINKED ISSUE #2526: Create separate types for the FederationApi and ModuleApi
In order to prevent mistakenly calling an module-scoped API when you want the federation-scoped API (or vice-versa) we should make these separate types.

jkitman:
Fixes #2526

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -29,7 +29,7 @@ use fedimint_client_legacy::{Client, UserClientConfig};
 use fedimint_core::admin_client::WsAdminClient;
 use fedimint_core::api::{
     ClientConfigDownloadToken, FederationApiExt, FederationError, GlobalFederationApi,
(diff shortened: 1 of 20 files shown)


douglaz (approved):
LGTM

okjodom (approved):
Is there any chance we'd have API name and signature collisions between the global vs module scoped apis? Before this change, how would I mistakenly call one instead of the other

jkitman:
> Is there any chance we'd have API name and signature collisions between the global vs module scoped apis? Before this change, how would I mistakenly call one instead of the other

Yeah before this both would be `IFederationApi` which we attached both global and module methods to through traits.

elsirion (commented):
While it's a nicely simple fix it leaves some room for improvement: whenever both module and global API calls are to be made *two* API structs now need to be present. That's kinda unnecessary, we could instead have the ModuleAPI also expose the GlobalAPI (just not the other way around) instead.

PR #2558, opened 2023-05-19
better DKG failures if modules are missing 

### LINKED ISSUE #2538: Fail if params for module config gen are missing
Currently the module is skipped, which is unexpected.

jkitman:
Fixes #2538

### DIFF
--- a/fedimint-server/src/config/mod.rs
+++ b/fedimint-server/src/config/mod.rs
@@ -485,6 +485,7 @@ impl ServerConfig {
         let hbbft_keys = keys[&KeyType::Hbbft].threshold_crypto();
         let epoch_keys = keys[&KeyType::Epoch].threshold_crypto();
 
+        let mut registered_modules = registry.kinds();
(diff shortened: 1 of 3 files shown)


fedimint-server/src/config/mod.rs
```diff
@@ -485,23 +485,28 @@ impl ServerConfig {
...
+                    None => Err(DkgError::ModuleNotFound(kind.clone())),
+                    Some(gen) => gen.distributed_gen(&dkg, module_params).await,
+                };
                 (module_instance_id, result)
             }
         });
         for (module_instance_id, config) in join_all(modules_runner).await {
-            module_cfgs.insert(module_instance_id, config?);
+            let config = config?;
+            registered_modules.remove(config.consensus_json.kind());
```
okjodom, line 505: Took me a minute to understand why we remove each kind from known registered modules

okjodom (approved):
Neat

okjodom (approved):

PR #2560, opened 2023-05-22
feat(ui): use number input for block confirmations and guardian number

EthnTuttle:
Uses the NumberInput for some form fields. Adds error text from when things are out of validity.

Delete the number or type in non-numeric strings to get an error printed.

### DIFF
--- a/fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
+++ b/fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
@@ -10,6 +10,12 @@ import {
   Button,
   Text,
   useTheme,
+  FormErrorMessage,
+  NumberInput,
+  NumberInputField,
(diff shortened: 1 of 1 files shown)


fedimint-ui/apps/guardian-ui/src/components/SetConfiguration.tsx
```diff
@@ -141,7 +147,7 @@ export const SetConfiguration: React.FC<Props> = ({ next }: Props) => {
                 'ln',
                 {
                   consensus: { network: network as Network },
-                  local: { bitcoin_rpc: bitcoinRpc }
+                  local: { bitcoin_rpc: bitcoinRpc },
```
EthnTuttle, line 150: Prettier formatting. Should probably add a pre-commit hook.

okjodom (approved):

PR #2568, opened 2023-05-23
Gateway test suite: 1/n

okjodom:
Updates existing gateway tests, and start filling in test cases in gateway integration test suite
- [x] existing auth tests use the test framework
- [x] test gateway can register with multiple feds
- [x] test gateway info covers all connected federations


fedimint-testing/src/fixtures.rs
```diff
@@ -101,20 +101,30 @@ impl Fixtures {
...
             decoders,
             ClientModuleGenRegistry::from(self.clients.clone()),
         )
-        .await;
+        .await
+    }
+
+    /// Starts a new gateway connected to a fed
+    pub async fn new_connected_gateway(&self, fed: &FederationTest) -> GatewayTest {
+        // TODO: Make construction easier
```
m1sterc001guy, line 125: Does this TODO still apply? Construction here looks pretty easy

m1sterc001guy:
This looks good to me. We've been making lots of changes to the test suites lately so I just want to understand the scope of our tests and their intended coverage. This is my understanding:

1. `fedimint-ln-tests`: for testing the fedimint lightning client. Ideally would only depend on the Lightning module (and not the mint module), similar to how other module client tests are written.
2. Gateway client tests: currently being worked on, but would ideally be structured similarly to `fedimint-ln-tests` and not depend on any other modules.
3. `Gatewayd` tests that directly exercise the web server API. That's what this PR is adding.
4. CLI tests: Full E2E tests that spawn a real `Gatewayd` and interact with it via the command line.

Is that an accurate description?

justinmoon (approved):

PR #2569, opened 2023-05-23
gateway test suite: 2/n

okjodom:
test gateway balance

### DIFF
--- a/gateway/ln-gateway/tests/integration_tests.rs
+++ b/gateway/ln-gateway/tests/integration_tests.rs
@@ -6,7 +6,7 @@ mod fixtures;
 
 use fedimint_testing::federation::FederationTest;
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
-use ln_gateway::rpc::ConnectFedPayload;
+use ln_gateway::rpc::{BalancePayload, ConnectFedPayload};
 
 #[tokio::test(flavor = "multi_thread")]
 async fn gatewayd_supports_connecting_multiple_federations() {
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -59,17 +59,65 @@ async fn gatewayd_shows_info_about_all_connected_federations() {
...
+    let (_, rpc, fed1, fed2, _) = fixtures::fixtures(None).await;
 
-    Ok(())
+    let id1 = fed1.connection_code().id;
+    let id2 = fed2.connection_code().id;
+
+    connect_federations(&rpc, &[fed1, fed2]).await.unwrap();
+
+    let bal1 = rpc
+        .get_balance(BalancePayload { federation_id: id1 })
```
okjodom, line 74: @jkitman, can we support balance queries and assertions with just the dummy module?
this gateway test fails with an error that signals the lack of mint module

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -59,17 +59,65 @@ async fn gatewayd_shows_info_about_all_connected_federations() {
...
+async fn gatewayd_allows_deposit_to_any_connected_federation() {
+    let (_, rpc, fed, _, bitcoin) = fixtures::fixtures(None).await;
 
-    Ok(())
+    let id = fed.connection_code().id;
+
+    connect_federations(&rpc, &[fed]).await.unwrap();
+
+    let addr = rpc
+        .get_deposit_address(DepositAddressPayload { federation_id: id })
```
okjodom, line 100: same here, except now we need the wallet module

gateway/ln-gateway/tests/integration_tests.rs
jkitman: The gateway needs to be converted to use the NG client.  I suggest focusing on this before we can get these tests to all pass.

gateway/ln-gateway/tests/integration_tests.rs
jkitman: We will need to use the wallet module for testing deposits, but again better to wait for the NG client.

gateway/ln-gateway/tests/integration_tests.rs
okjodom: ack.

gateway/ln-gateway/tests/integration_tests.rs
okjodom: ack.

okjodom:
closed in favor of a new test suite coming in with gateway #2600

PR #2571, opened 2023-05-23
fix: broken devshell on aarch64

Maan2003:


oleonardolima (approved):
I've tested on macos, and now it works :shipit:

justinmoon (approved):

PR #2574, opened 2023-05-23
chore: protect legacy client high-level note-related operations from concurrent use

dpc:
Just protect them with a mutex, bolted a bit ad-hoc in there.

In seems that in the gateway we already handle incoming payments on a separate task, without any protection, and in the Fedi Alpha deployement we've noticed some strage behavior where some outputs/notes seems to be missing etc.

This should also enable some concurrent usage of the `Client`, serializing only what we know is not safe.


m1sterc001guy (approved):
New client fixes this :)

dpc:
Landing to avoid merge conflict. Feel free to point out any change requests.

PR #2576, opened 2023-05-24
Client-NG: on-chain withdrawals

elsirion:


modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -291,11 +470,64 @@ impl WalletClientModule {
...
     }
+
+    pub async fn get_withdraw_fees(
+        &self,
+        address: bitcoin::Address,
+        amount: bitcoin::Amount,
+    ) -> anyhow::Result<PegOutFees> {
+        check_address(&address, self.cfg.network)?;
+
+        self.module_api
```
m1sterc001guy, line 481: Directly calling the module api here to me just seems like a shortcut to spawning a state machine which can access the api. 

I think its fine here since spawning a state machine is probably overkill for one API call. For future module developers though, do you think there's a worry here that APIs can be called without spawning a state machine? With this couldn't you implement logic that is very similar to the legacy client?

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -127,61 +172,177 @@ impl WalletClientExt for Client {
...
-            }
-            yield DepositState::Claimed;
-        }))
+        let operation_id = OperationId(thread_rng().gen());
+
+        let withdraw_output = wallet_client
+            .create_withdraw_output(operation_id, address.clone(), amount, fee.clone())
+            .await?;
+        let tx_builder =
+            TransactionBuilder::new().with_output(withdraw_output.into_dyn(instance.id));
```
m1sterc001guy, line 253: Is it possible for this transaction to create change?

modules/fedimint-wallet-client/src/lib.rs
elsirion: This is very intentional: we need a way to get permission from the user to pay these fees, meaning if we'd fetch them inside the state machine we'd need to concoct a way to get user feedback in there.

Due to this having the fetching inside a state machine would be of little to no use since any sensible integrator will block on user feedback anyway and fetching fees does not give us any guarantees or requires us to take action later in regards to the federation, making all the possible benefits of doing so in a state machine moot.

modules/fedimint-wallet-client/src/lib.rs
elsirion: Yes, it is, you think we should wait for that to arrive before considering the operation finished for easier testing?

modules/fedimint-wallet-client/src/lib.rs
m1sterc001guy: Yep I agree with all of that. My main concern was for new module developers who might not know when to use a state machine and when not to. With the module api directly exposed they now have the option of directly implementing their logic in the ClientModule. 

Agreed that this is the correct approach.

modules/fedimint-wallet-client/src/lib.rs
m1sterc001guy: Ideally I think the outer state machine should have a state that awaits the change, although I see that the devimint test doesn't directly verify the change in ecash notes so maybe its not necessary

elsirion: I added rudimentary change awaiting, not sure how useful this is for real apps, but ok with me for now since it allowed me to also test the balance change.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -127,61 +172,185 @@ impl WalletClientExt for Client {
...
 async fn next_deposit_state<S>(stream: &mut S) -> Option<DepositStates>
 where
     S: Stream<Item = WalletClientStates> + Unpin,
 {
-    match stream.next().await? {
-        WalletClientStates::Deposit(ds) => Some(ds.state),
+    loop {
+        if let WalletClientStates::Deposit(ds) = stream.next().await? {
+            return Some(ds.state);
+        }
```
douglaz, line 342: I like to use [yield_now](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) inside loops on async functions. It avoids starving other threads:
```suggestion
        }
        tokio::task::yield_now().await;
```

It saved me a couple of times.

elsirion: TIL, added it, but want to get rid of these functions anyway, not entirely sure how yet.

justinmoon (approved):

PR #2577, opened 2023-05-24
Remove legacy DKG

jkitman:
- Switched devimint to our trusted config gen since it's far simpler/faster and we already have good coverage of DKG and the config gen api
- Cleaned up a ton of the old DKG related code that we no longer need.


fedimint-cli/src/lib.rs
```diff
@@ -578,6 +584,38 @@ enum Command {
     /// Show the status according to the `status` endpoint
     Status,
 
+    ConfigDecrypt {
```
elsirion, line 587: This will be slightly annoying to use imo since `fedimint-cli` always wants a client datadir afaik. I think we might want to reorganize the commands a bit.

elsirion (approved):
LGTM :tada:

PR #2588, opened 2023-05-26
fix(ln-server): save output outcome for outgoing contract cancellations

elsirion:


m1sterc001guy (approved):
Verified on my WIP branch that is implementing Gateway cancellation. Thanks!

PR #2591, opened 2023-05-26
fix: actually test wasm in CI

Maan2003:
https://github.com/fedimint/fedimint/pull/2579#discussion_r1207095544

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -92,6 +92,26 @@ jobs:
         if: (github.ref == 'refs/heads/master' || matrix.build-in-pr) && matrix.run-tests
         run: nix build -L .#ci.cli-test.all
 
+  wasm-tests:
+    name: "Wasm tests"
+    runs-on: buildjet-8vcpu-ubuntu-2004
+    timeout-minutes: 20
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: cachix/install-nix-action@v21
+        with:
+          nix_path: nixpkgs=channel:nixos-22.05
+      - uses: cachix/cachix-action@v12
+        with:
+          name: fedimint
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+        continue-on-error: true
+
+      - name: Tests
+        run: nix develop .#crossWasm -c ./scripts/wasm-tests.sh
+
   audit:
     name: "Audit"
     runs-on: buildjet-2vcpu-ubuntu-2004
@@ -298,7 +318,7 @@ jobs:
 
   notifications:
     # if: github.ref == 'refs/heads/master'
-    if: always()                              
+    if: always()
     name: "Notifications"
     timeout-minutes: 1
     runs-on: ubuntu-22.04
@@ -307,7 +327,7 @@ jobs:
     needs: [ build, cross, ccov, containers, pkgs ]
 
     steps:
-    - name: Discord notifications on failure                                                         
+    - name: Discord notifications on failure
       # https://stackoverflow.com/a/74562058/134409
       if: ${{ always() && contains(needs.*.result, 'failure') }}
       # https://github.com/marketplace/actions/actions-status-discord
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -676,6 +676,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "console_error_panic_hook"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


nix/craneBuild.nix
```diff
@@ -279,5 +279,23 @@ craneLib.overrideScope' (self: prev: {
...
+    cargoToml = self.commonArgs.src + "/fedimint-wasm-tests/Cargo.toml";
+    cargoLock = self.commonArgs.src + "/fedimint-wasm-tests/Cargo.lock";
+    buildPhaseCargoCommand = "cargo build --tests --manifest-path fedimint-wasm-tests/Cargo.toml";
+    nativeBuildInputs = self.commonArgs.nativeBuildInputs ++ [ pkgs.breakpointHook ];
+    cargoArtifacts = null;
+    doInstallCargoArtifacts = false;
+    installPhaseCommand = ''
+      mv fedimint-wasm-tests/target/wasm32-unknown-unknown/debug/deps/fedimint*.wasm $out
+    '';
+    preBuild = ''
```
dpc, line 295: There's a variable in `nix/toolchain.nix` that already exports env vars for wasm.

nix/craneCommon.nix
```diff
@@ -49,7 +49,7 @@ craneLib.overrideScope' (self: prev: {
   filterWorkspaceDepsBuildFiles = src: self.filterSrcWithRegexes [ "Cargo.lock" "Cargo.toml" ".cargo" ".cargo/.*" ".*/Cargo.toml" ".*/proto/.*" ] src;
 
   # Filter only files relevant to building the workspace
-  filterWorkspaceFiles = src: self.filterSrcWithRegexes [ "Cargo.lock" "Cargo.toml" ".cargo" ".cargo/.*" ".*/Cargo.toml" ".*\.rs" ".*\.html" ".*/proto/.*" "db/migrations/.*" ] src;
+  filterWorkspaceFiles = src: self.filterSrcWithRegexes [ "Cargo.lock" "Cargo.toml" ".cargo" ".cargo/.*" ".*/Cargo.toml" ".*/Cargo.lock" ".*\.rs" ".*\.html" ".*/proto/.*" "db/migrations/.*" ] src;
```
dpc, line 52: We have more than root `Cargo.lock`?

nix/craneBuild.nix
```diff
@@ -279,5 +279,23 @@ craneLib.overrideScope' (self: prev: {
...
+    pname = "fedimint-wasm-tests";
+    CARGO_BUILD_TARGET = "wasm32-unknown-unknown";
+    cargoToml = self.commonArgs.src + "/fedimint-wasm-tests/Cargo.toml";
+    cargoLock = self.commonArgs.src + "/fedimint-wasm-tests/Cargo.lock";
+    buildPhaseCargoCommand = "cargo build --tests --manifest-path fedimint-wasm-tests/Cargo.toml";
+    nativeBuildInputs = self.commonArgs.nativeBuildInputs ++ [ pkgs.breakpointHook ];
+    cargoArtifacts = null;
+    doInstallCargoArtifacts = false;
+    installPhaseCommand = ''
+      mv fedimint-wasm-tests/target/wasm32-unknown-unknown/debug/deps/fedimint*.wasm $out
```
dpc, line 293: A test derivation shouldn't need to put anything into `$out`. The result of it mostly "success" or "failure".

Maan2003:
really annoying, ring is trying to write build.rs outputs in its source tree which is allowed in dev shell (~/.cargo/git) but not in nix build (/nix/store)

.github/workflows/ci-nix.yml
```diff
@@ -92,6 +92,26 @@ jobs:
...
+        with:
+          nix_path: nixpkgs=channel:nixos-22.05
+      - uses: cachix/cachix-action@v12
+        with:
+          name: fedimint
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+        continue-on-error: true
+
+      - name: Tests
+        run: nix develop .#crossWasm -c ./scripts/wasm-tests.sh
```
dpc, line 113: I need to verify, but since this is using `workspaceBuild`, we might want to run it in the normal job, after existing tests, so that we are not doing whole workspace build twice.

dpc (commented):
Oh, actually I see that you didn't go with the Nix route. That will have many downsides.

Basically there will be no caching of any kind: the whole workspace + wasm build + tests will be running every time.

If we reused `workspaceBuild`, then things would go much faster.

Maan2003:
I think no caching is fine. Because this is a new job, it will not create a bottleneck.
maybe we can get devimint from `workspaceBuild`. But that can be done in future PRs.

dpc:
Fixed in #2656

PR #2592, opened 2023-05-26
Build Gateway UI with gateway API

okjodom:
- Cleans up a bunch of mocks in gateway UI
- Simplifies offering of gateway UI mvp (_and we'll look for design guidance on the experience to build from here on_)
- Simplifies API definition and restructures app to follow `guardian-ui` source organization
- Builds and uses gateway api `/info`
- Builds and uses gateway api `/address`

Replaces ##2323, initial but shelved attempt at connecting gateway UI to the API


okjodom:
migrated to UI repo: https://github.com/fedimint/ui/pull/25

PR #2593, opened 2023-05-28
feat: deprecate ui code in favor of new repo

EthnTuttle:
This remove the `fedimint-ui` folder from the repository so we can migrate over to `fedimint/ui` as the main repo for the UI code.

Please see https://github.com/EthnTuttle/fedimint-ui for what I think to be a good starting point for the UI repo.

Marked Draft because we should probably wait to confirm deletion of the UI, and there's probably more UI code that can be removed from this repo (nix, ci, etc.)


okjodom:
We need to first prep https://github.com/fedimint/ui then we can pull off a migration without breaking continuity in development.

Maybe we should
 - [ ] push a fully configured docker compose configuration with instructions on how to run it
 - [ ] make a copy of `fedimint-ui/` workspace in new repo
 - [ ] make this change (deletion) while adding instructions that guide devs to the new repo
 - [ ] migrate any UI issues over to the new repo
 
 @EthnTuttle, let's collab on this

EthnTuttle:
Closing to revisit later.

PR #2596, opened 2023-05-29
refactor: remove LnRpcAdapter and simplify lightning test clients

m1sterc001guy:
This PR attempts to simplify the lightning test clients that are used for the integration tests. Before, there was an `LnRpcAdapter` wrapper struct and fake and real implementations of lightning RPCs, which in my opinion became a bit difficult to reason about. This PR makes the following changes:

- Removes `LnRpcAdapter`
- `LightningTest` is now an extension trait of `ILnRpcClient`. Logically, this means it is a lightning client that the gateway can use, but it also defines some extra functions that are useful for tests (ex: `FakeLightningTest`, `ClnLightningTest`, `LndLightningTest`.
- `RealLightningTest` has been split into CLN and LND. The previous implementation kept connections to both Lightning nodes, but the Fake implementation didn't have this same notion, which caused the fake and real implementations to be different (this was the actual requirement for `LnRpcAdapter`). Splitting these implementations into two simplifies the code and allows for the removal of `LnRpcAdapter`

### DIFF
--- a/fedimint-testing/src/ln/mock.rs
+++ b/fedimint-testing/src/ln/mock.rs
@@ -1,8 +1,6 @@
-use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use anyhow::anyhow;
 use async_trait::async_trait;
 use bitcoin::hashes::{sha256, Hash};
 use bitcoin::secp256k1::{PublicKey, SecretKey};
@@ -20,8 +18,6 @@ use ln_gateway::gatewaylnrpc::{
 use ln_gateway::lnrpc_client::{ILnRpcClient, RouteHtlcStream};
 use ln_gateway::GatewayError;
 use rand::rngs::OsRng;
-use tokio::sync;
-use tokio::sync::RwLock;
 use tokio_stream::wrappers::ReceiverStream;
 
 use super::LightningTest;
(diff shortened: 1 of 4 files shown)


fedimint-testing/src/ln/mock.rs
```diff
@@ -134,70 +130,3 @@ impl ILnRpcClient for FakeLightningTest {
...
-
-    /// Register `invoice` to fail `times` before (attempt) succeeding. The
-    /// invoice will be dropped from the HashMap after succeeding
-    #[allow(dead_code)]
-    pub async fn fail_invoice(&self, invoice: PayInvoiceRequest, times: u8) {
-        self.fail_invoices
-            .lock()
-            .await
-            .insert(invoice.invoice, times + 1);
-    }
```
justinmoon, line 167: So we had a mechanism for failing invoices but we never used it anywhere? cc @okjodom

m1sterc001guy: Yeah I don't think it is used. If we need that functionality I don't think we need another adapter class, we can use add it to `LightningTest` instead

fedimint-testing/src/ln/real.rs
```diff
@@ -204,7 +175,137 @@ impl RealLightningTest {
...
+
+#[async_trait]
+impl LightningTest for LndLightningTest {
+    async fn invoice(
+        &self,
+        amount: Amount,
+        expiry_time: Option<u64>,
+    ) -> ln_gateway::Result<Invoice> {
+        info!("fetching invoice from lnd");
+        let mut lnd_rpc = self.rpc_lnd.lock().await;
```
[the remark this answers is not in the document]
m1sterc001guy: I tried to abstract each implementation to be a connection to a single node. If you need an invoice from a different node, just call `invoice` from a different instance (i.e `ClnLightningTest`) then call pay on `LndLightningTest`. Does that make sense?

elsirion (approved):
@justinmoon please coordinate on merging :)

justinmoon (approved):

Issue #2514, opened 2023-05-13
Cient NG TODO list


elsirion:
- [x] #2506
- [x] Operation terminal state cache https://github.com/fedimint/fedimint/pull/2536
- [x] Notifications on balance update #2537
- [x] Abstracted balance fetching #2537
- [x] Pagination for operations https://github.com/fedimint/fedimint/pull/2594
- [x] Ignore unknown modules in config #2535
- [x] Add sent e-cash notes to out-of-band send operation meta https://github.com/fedimint/fedimint/pull/2860
- [x] Deduplicate denomination in derivation path  #2539
- [x] Clean CLI shutdown https://github.com/fedimint/fedimint/pull/2821
- [x] Derive deposit address tweaks from root secret https://github.com/fedimint/fedimint/pull/2852
- [x] Check version compatibility with server #2922
- [x] Update #2347 -style metadata to state machines after they've been created.
- [x] Swallows panics https://github.com/fedimint/fedimint/issues/2585
- [x] Reissuing own oob-spent e-cash notes #2597
- [x] Test withdraw refund case https://github.com/fedimint/fedimint/pull/2921
- [x] https://github.com/fedimint/fedimint/issues/2585
- [x] Add default esplora server to client config https://github.com/fedimint/fedimint/pull/2813
- [x] Debug https://github.com/fedimint/fedimint/issues/2861

elsirion:
> Do we need to support multiple modules having assets? Definitely complicates creating inputs for payments.

Not for now imo, that would be equivalent to supporting multiple primary modules, right?

jkitman:
> Not for now imo, that would be equivalent to supporting multiple primary modules, right?

Yes, could be possible if we mix account-based and ecash modules.  Or maybe when we do multi-asset.

douglaz:
I think some kind of remint is missing. I mean, how is a client supposed to recreate some lower denomination notes after a spend?

jkitman:
> I think some kind of remint is missing.

Txs can have multiple outputs, the change output contains notes of any denomination.

justinmoon:
I'm pretty sure client ng can't call `reissue` on it's own notes. This is hard to do now because the serialization of the `ng spend` command isn't clipboard-friendly. But it seems both the `spend` and the `reissue` of the same set of notes will cause a `"We already reissued these notes"` error just because the client has seen this not before, but not because they've been reissued.

justinmoon:
[...]

PR #2597, opened 2023-05-29
fix(wallet-client): don't fail on self-reissue

elsirion:
Fixes the issue described in https://github.com/fedimint/fedimint/issues/2514#issuecomment-1555368893


douglaz (approved):
LGTM

justinmoon (approved):

PR #2598, opened 2023-05-29
feat: gateway outgoing payments

m1sterc001guy:
State machines and client code for the lightning gateway's outgoing payments

### DIFF
--- a/gateway/ln-gateway/src/ng/mod.rs
+++ b/gateway/ln-gateway/src/ng/mod.rs
@@ -40,7 +40,7 @@ use crate::lnrpc_client::ILnRpcClient;
 const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
 
 /// The high-level state of a reissue operation started with
-/// [`LightningClientExt::pay_bolt11_invoice`].
+/// [`GatewayClientExt::gateway_pay_bolt11_invoice`].
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 pub enum GatewayExtPayStates {
     Created,
--- a/gateway/ln-gateway/src/ng/mod.rs
+++ b/gateway/ln-gateway/src/ng/mod.rs
@@ -22,6 +22,7 @@ use fedimint_core::{apply, async_trait_maybe_send, Amount, OutPoint, Transaction
 use fedimint_ln_client::api::LnFederationApi;
 use fedimint_ln_client::contracts::ContractId;
 use fedimint_ln_common::config::LightningClientConfig;
+use fedimint_ln_common::contracts::Preimage;
 use fedimint_ln_common::route_hints::RouteHint;
 use fedimint_ln_common::{
     ln_operation, LightningCommonGen, LightningGateway, LightningModuleTypes, LightningOutput, KIND,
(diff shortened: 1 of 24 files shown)


justinmoon:
looks like there's some issues with rustdoc in ci

m1sterc001guy:
> looks like there's some issues with rustdoc in ci

Yeah had some issues with the aquamarine diagram. Should be fixed now.

gateway/ln-gateway/src/ng/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+
+const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
+
+/// The high-level state of a reissue operation started with
+/// [`GatewayClientExt::gateway_pay_bolt11_invoice`].
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum GatewayExtPayStates {
+    Created,
+    Preimage,
+    Success,
```
justinmoon, line 48: Return `Preimage` here, too. I think this is what the `/pay_invoice` HTTP method will actually be waiting for.

modules/fedimint-ln-common/src/api.rs
```diff
@@ -44,7 +44,7 @@ where
     }
 
     async fn fetch_contract(&self, contract: ContractId) -> FederationResult<ContractAccount> {
-        self.request_current_consensus("wait_account".to_string(), ApiRequestErased::new(contract))
+        self.request_current_consensus("account".to_string(), ApiRequestErased::new(contract))
```
justinmoon, line 47: I'm not sure about this change.

justinmoon: Also, the difference between a "contract" and a "contract account" has always been confusing. Not much to do here, but just making note.

justinmoon: This actually seems like a good change. `fetch_contract` should probably just get the contract and fail if it's not there. I'm not aware of any case where the gateway will go looking for a contract before it has been created and need to wait for it to show up.

gateway/ln-gateway/src/ng/pay.rs
```diff
@@ -0,0 +1,439 @@
...
+                    common,
+                    state: GatewayPayStates::CancelContract(Box::new(GatewayPayCancelContract {
+                        contract,
+                        error: e,
+                    })),
+                },
+                OutgoingPaymentError::OutgoingContractDoesNotExist { contract_id } => {
+                    GatewayPayStateMachine {
+                        common,
+                        state: GatewayPayStates::Canceled(None, contract_id),
```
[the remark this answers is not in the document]
m1sterc001guy: `OfferDoesNotExist` sounds better. Will make that change.

modules/fedimint-ln-common/src/api.rs
m1sterc001guy: The gateway might not, but does the lightning client? This is shared between both. I don't think this is necessary, I would probably prefer to revert unless others have strong opinions.

gateway/ln-gateway/src/ng/tests.rs
```diff
@@ -110,8 +110,19 @@ async fn test_gateway_client_pay_valid_invoice() -> anyhow::Result<()> {
...
+                _ => {
+                    panic!("Unexpected state occurred, expected Preimage");
+                }
+            }
+            match gw_pay_sub.ok().await? {
+                GatewayExtPayStates::Success { preimage: _ } => {}
+                _ => {
+                    panic!("Unexpected state occurred, expected Success");
+                }
+            }
```
justinmoon, line 124: I think you can use something like `assert_matches!(gw_pay_sub.ok().await?, GatewayExtPayStates::Success { .. });`

PR #2602, opened 2023-05-30
clients can exec internal payments within fed

okjodom:
defines  a state machine used by clients to make internal payments within federation, without involving gateways


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -125,6 +152,28 @@ pub enum LnReceiveState {
...
+    let check_first_hop_to_current_fed = invoice
+        .route_hints()
+        .first()
+        .and_then(|rh| rh.0.last())
+        .map(|hop| (hop.src_node_id, hop.short_channel_id))
+        == Some((gateway.node_pub_key, gateway.mint_channel_id));
+
+    Ok(
+        check_first_hop_to_current_fed
+            && instance.api.offer_exists(*invoice.payment_hash()).await?,
```
justinmoon, line 173: Not sure if we want to check whether the offer exists right here. If the route hint checks out, but no offer exists, then I think it will currently try to pay via the gateway. But the first step of the internal payment state machine could check this and fail the payment without needing to connect to the gateway. 

I think either one would probably work, just thinking out loud.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -125,6 +154,24 @@ pub enum LnReceiveState {
     Claimed,
 }
 
+async fn check_internal_payment(
```
justinmoon, line 157: I think we should check whether the route hint matches **any** gateway, not just the active one.

okjodom: ack

fedimint-cli/src/ng.rs
```diff
@@ -196,29 +198,52 @@ pub async fn handle_ng_command(
...
+
+            match pay_type {
+                PayType::Internal(operation_id) => {
+                    let mut updates = client
+                        .subscribe_internal_pay(operation_id)
+                        .await?
+                        .into_stream();
+
+                    while let Some(update) = updates.next().await {
+                        if let InternalPayState::Preimage(preimage) = update {
```
m1sterc001guy, line 211: Is the success case the only state that we want the CLI to wait on? What happens if there's a refund?

gateway/ln-gateway/src/ng/mod.rs
```diff
@@ -488,7 +467,7 @@ impl GatewayClientModule {
 #[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
 pub enum GatewayClientStateMachines {
     Pay(GatewayPayStateMachine),
-    Receive(GatewayReceiveStateMachine),
+    Receive(InternalTxStateMachine),
```
m1sterc001guy, line 470: I think `InternalTxStateMachine` is still a bit confusing. Because this will also process external lightning payments into the federation. Can we call it `IncomingStateMachine` or something like that? That covers both the internal tx and external lightning payments.

gateway/ln-gateway/src/ng/mod.rs
```diff
@@ -531,3 +510,42 @@ impl State for GatewayClientStateMachines {
...
     }
 }
+
+#[derive(Error, Debug, Serialize, Deserialize, Encodable, Decodable, Clone, Eq, PartialEq)]
+pub enum ReceiveError {
+    #[error("Route htlc error")]
+    RouteHtlcError,
+}
+#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
+pub struct Htlc {
```
m1sterc001guy, line 520: Did this code just move?

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -95,8 +105,27 @@ pub trait LightningClientExt {
     ) -> anyhow::Result<UpdateStreamOrOutcome<'_, LnReceiveState>>;
 }
 
-/// The high-level state of a reissue operation started with
-/// [`LightningClientExt::pay_bolt11_invoice`].
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum PayType {
+    // Payment from this client to another user within the federation
+    Internal(OperationId),
+    // Payment from this client to another user, facilitated by a gateway
```
m1sterc001guy, line 112: This enum is also used in the receive case though too right? Not just Outgoing

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -128,6 +154,19 @@ pub enum LnReceiveState {
...
+async fn check_internal_payment(invoice: &Invoice, gateways: Vec<LightningGateway>) -> bool {
+    // Asserts that the first hop of the invoice route hints
+    // leads through a registered gateway into the current federation
+    gateways.into_iter().any(|gateway| {
+        invoice
+            .route_hints()
+            .first()
+            .and_then(|rh| rh.0.last())
+            .map(|hop| (hop.src_node_id, hop.short_channel_id))
+            == Some((gateway.node_pub_key, gateway.mint_channel_id))
```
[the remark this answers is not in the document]
m1sterc001guy: We had some offline discussion about this last week. Not something to pursue for this PR but could potentially be improved in the future.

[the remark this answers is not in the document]
m1sterc001guy: https://github.com/fedimint/fedimint/issues/2618

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -171,24 +213,31 @@ impl LightningClientExt for Client {
...
-                operation_id,
-                instance.api,
-                invoice.clone(),
-                active_gateway,
-                self.get_config().await.federation_id,
-                rand::rngs::OsRng,
-            )
-            .await?;
+        let gateways = self.fetch_registered_gateways().await?;
+        let (pay_type, output, contract_id) = if check_internal_payment(&invoice, gateways).await {
```
m1sterc001guy, line 222: nit: I prefer `is_internal_payment` so its obvious what the function is returning

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -404,30 +486,30 @@ impl ClientModuleGen for LightningClientGen {
...
-        _module_api: DynModuleApi,
+        module_api: DynModuleApi,
     ) -> anyhow::Result<Self::Module> {
         Ok(LightningClientModule {
             cfg,
             notifier,
+            redeem_key: module_root_secret
+                .child_key(ChildId(0))
+                .to_secp_key(&Secp256k1::new()),
             secp: secp256k1_zkp::Secp256k1::new(),
```
m1sterc001guy, line 500: It's kind of weird that we instantiate two secp contexts here? Can we use the same one? I forget why we need the zkp context here, maybe @elsirion knows

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -468,19 +551,19 @@ impl ClientModule for LightningClientModule {
 }
 
 #[derive(Error, Debug, Serialize, Deserialize)]
-pub enum LightningPayError {
+pub enum PayError {
     #[error("Lightning payment was canceled")]
```
m1sterc001guy, line 555: Should we remove "Lightning" from all of these error messages or is `Failed` the only one that is re-used in the internal payment case?

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -580,6 +663,50 @@ impl LightningClientModule {
...
+
+        let (incoming_output, contract_id) = create_incoming_contract_output(
+            &self.module_api,
+            *payment_hash,
+            invoice_amount,
+            self.redeem_key,
+        )
+        .await?;
+
+        let client_output = ClientOutput::<LightningOutput, LightningClientStateMachines> {
```
m1sterc001guy, line 692: Hmmm why was this part not included in `create_incoming_contract_output`? Its the same for the gateway and the client right?

m1sterc001guy: What's the difference between this function and `create_funding_incoming_contract_output` in the gateway?

modules/fedimint-ln-common/src/internal.rs
```diff
@@ -1,50 +1,40 @@
+//! # Internal Transaction State Machine
```
m1sterc001guy, line 1: As mentioned previously, I think we should rename this

modules/fedimint-ln-common/src/internal.rs
```diff
@@ -1,50 +1,40 @@
...
+};
+use crate::contracts::{Contract, ContractId, DecryptedPreimage, IdentifiableContract, Preimage};
+use crate::{
+    ContractOutput, LightningClientContext, LightningInput, LightningOutput, LightningOutputOutcome,
+};
 
 #[cfg_attr(doc, aquamarine::aquamarine)]
-/// State machine that executes the Lightning payment on behalf of
-/// the fedimint user that requested an invoice to be paid.
+/// State machine that executes a transaction between two users
```
m1sterc001guy, line 35: This is not necessarily true. It could be for an external lightning payment when it is used for the gateway

modules/fedimint-ln-common/src/internal.rs
```diff
@@ -329,3 +306,45 @@ pub struct PreimageState {
 pub struct RefundSuccessState {
     refund_txid: TransactionId,
 }
+
+async fn fetch_and_validate_offer(
```
m1sterc001guy, line 310: I don't think this function and `create_incoming_contract_output` should be in the state machine code. I think it should be moved to lib.rs of ln-common

modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -80,13 +82,16 @@ async fn makes_internal_payments_within_gateway() -> anyhow::Result<()> {
     assert_eq!(sub1.ok().await?, LnReceiveState::Created);
     assert_matches!(sub1.ok().await?, LnReceiveState::WaitingForPayment { .. });
 
-    let (op, _) = client2.pay_bolt11_invoice(invoice).await?;
-    let mut sub2 = client2.subscribe_ln_pay(op).await?.into_stream();
-    assert_eq!(sub2.ok().await?, LnPayState::Created);
-    // TODO: Finish after gw moves from legacy client
-    // assert_eq!(next(sub2).await, LnPayState::Funded);
-    // assert_matches!(next(sub2).await, LnPayState::Success{..});
-    // assert_eq!(next(sub1).await, LnReceiveState::Funded);
```
m1sterc001guy, line 89: yay :)

m1sterc001guy (changes requested):
Looks like it works to me, I do think we should do the renames though

gateway/ln-gateway/src/ng/mod.rs
okjodom: Yes. from `ln-gateway/src/ng/incoming.rs` which no longer exists

modules/fedimint-ln-client/src/lib.rs
okjodom: No. `PayType` enum is used only to differentiate the two cases of `pay_bolt11_invoice` action by the fedimint client

modules/fedimint-ln-client/src/lib.rs
okjodom: makes sense. Updated

modules/fedimint-ln-client/src/lib.rs
okjodom: No, they differ on the typed state machine wrap within the client output. These differ in type only, where here we return a `LightningClientStateMachines::InternalPay`, the gateway scenario returns a `GatewayClientStateMachines::Receive`.

In my opinion, these state machines each make sense for their respective scenarios, so there is no need to find some universal name for a state machine we can return in both scenarios, from a shared utility fn

modules/fedimint-ln-common/src/internal.rs
okjodom: Well we should consider the gateway as just another (specialized) user within the fed, no?
Occured to me that this is why we can share the same SM both in `gateway-to-client`, and `client-to-client` scenarios. In the end it's a tx between two users of a federation, but for different reasons

modules/fedimint-ln-client/src/lib.rs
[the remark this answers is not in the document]
okjodom: ack. makes for a good follow-up as improvement on this change

modules/fedimint-ln-common/src/internal.rs
okjodom: ack. moved

modules/fedimint-ln-tests/tests/tests.rs
okjodom: :)

gateway/ln-gateway/src/ng/mod.rs
okjodom: `IncomingStateMachine` makes sense to me from the perspective of the contract being created and funded, but does not seem satisfactory, especially considering all the other SMs are not named per the underlying contracts. Adopting anyways

fedimint-cli/src/ng.rs
okjodom: cli now reports refund, and also breaks the loop on error case

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -373,6 +422,39 @@ impl LightningClientExt for Client {
...
+        let (lightning, _instance) = self.get_first_module::<LightningClientModule>(&KIND);
+
+        let operation = ln_operation(self, operation_id).await?;
+        Ok(operation.outcome_or_updates(self.db(), operation_id, || {
+            stream! {
+                yield InternalPayState::Funding;
+
+                let mut stream = lightning.notifier.subscribe(operation_id).await;
+                let state = loop {
+                    if let Some(LightningClientStateMachines::InternalPay(state)) = stream.next().await {
```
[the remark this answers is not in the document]
okjodom: oh yeah

gateway/ln-gateway/src/ng/mod.rs
m1sterc001guy: > especially considering all the other SMs are not named per the underlying contracts

This is probably an oversight on my part. In my opinion we should probably rename the `pay` state machine as well.

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Ah ok I was confused, thanks!

m1sterc001guy (dismissed):

modules/fedimint-ln-client/src/lib.rs
okjodom: Reverted the string change in `Failed` error variant. `PayError` is no-longer shared between the two, and I think we should refactor to get rid of `await_lightning_payment_success` and `await_refund` functions. There map error cases into `PayError`, but all of these errors are then swallowed when yielding extension state, for example in L398.

modules/fedimint-ln-client/src/lib.rs
okjodom: shared it. lmk if there's some bad in sharing

m1sterc001guy (approved):

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -172,24 +214,31 @@ impl LightningClientExt for Client {
...
-            .create_outgoing_output(
-                operation_id,
-                instance.api,
-                invoice.clone(),
-                active_gateway,
-                self.get_config().await.federation_id,
-                rand::rngs::OsRng,
-            )
-            .await?;
+        let gateways = self.fetch_registered_gateways().await?;
```
justinmoon, line 222: Note: this adds an extra network round trip to a threshold of guardians every time we pay an invoice.

okjodom: we could cache known gateways. adopting [this proposal](https://github.com/fedimint/fedimint/issues/2618#issuecomment-1587369593) would completely save us the network trip if we never have to look through registered gateways list to identify internal payments

Issue #2605, opened 2023-05-31
Audit functionality extremely slow


dpc:
On alpha.fedi.xyz federation, we've noticed very long time to process epoch. [I did some performance debugging.](https://github.com/fedimint/fedimint/pull/2603):

```
May 30 23:36:59 alpha-mutinynet-01 fedimintd-start[1499533]: 2023-05-30T23:36:59.724463Z  INFO consensus: Processing items from missing epoch 71498                                                                  
May 30 23:36:59 alpha-mutinynet-01 fedimintd-start[1499533]: 2023-05-30T23:36:59.724684Z DEBUG process_consensus_outcome{epoch=71498}: timing: Operation timing name=process_module_consensus_items duration_ms=0    
May 30 23:36:59 alpha-mutinynet-01 fedimintd-start[1499533]: 2023-05-30T23:36:59.724701Z DEBUG process_consensus_outcome{epoch=71498}: timing: Operation timing name=build_verification_caches duration_ms=0         
May 30 23:36:59 alpha-mutinynet-01 fedimintd-start[1499533]: 2023-05-30T23:36:59.724707Z DEBUG process_consensus_outcome{epoch=71498}: timing: Operation timing name=process_transactions duration_ms=0              
May 30 23:36:59 alpha-mutinynet-01 fedimintd-start[1499533]: 2023-05-30T23:36:59.791507Z DEBUG process_consensus_outcome{epoch=71498}: timing: Operation timing name=add sig to prev duration_ms=58                  
May 30 23:36:59 alpha-mutinynet-01 fedimintd-start[1499533]: 2023-05-30T23:36:59.798441Z DEBUG process_consensus_outcome{epoch=71498}: timing: Operation timing name=finalize_process_epoch duration_ms=73           
May 30 23:37:08 alpha-mutinynet-01 fedimintd-start[1499533]: 2023-05-30T23:37:08.349480Z DEBUG process_consensus_outcome{epoch=71498}: timing: Operation timing name=process_consensus_outcome duration_ms=8624
```

and after scratch head and adding more timing info (already squashed in the #2603) I've found the culprit:

```
May 31 01:56:11 alpha-mutinynet-01 fedimintd-start[1507456]: 2023-05-31T01:56:11.628170Z DEBUG process_consensus_outcome{epoch=71676}: timing: Operation timing name=audit duration_ms=8255
```

[...]

PR #2606, opened 2023-05-31
fix: don't accumulate empty epochs on rejoin requests

dpc:
In Fedi Alpha deployement, probably caused by #2605 , we've noticed Federations accumulating hundreds of empty epochs to be created, due to how currently they are being accumulated (`+=`) on every request.

Change the code to cap the pending epochs to a max value (10).

While at it, do some renaming.


elsirion (approved):
I think this is the right thing to do since we saw rejoin requests pile up, pinging @jkitman just in case.

jkitman:
Seems reasonable, although I'm not sure what would be causing it...will be a relief once we switch away from this consensus mechanism.

PR #2608, opened 2023-05-31
feat: fund client ng from devimint dev-fed

### LINKED ISSUE #2507: `devimint dev-fed` should fund new cli client
Probably best to wait for https://github.com/fedimint/fedimint/pull/2416 and just do a peg-in

jp1ac4:
This is to resolve #2507.

I've created a new method `pegin_ng` based on the changes from #2416. This is called as part of `devimint dev-fed` to fund the new client and also as part of the CLI tests where I took the original code from.

I've updated some CLI tests to reflect the non-zero initial balance of the new client.

I've also reverted the changes from #2509.

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -884,18 +884,7 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         .as_u64()
         .unwrap();
 
-    let deposit = cmd!(fed, "ng", "deposit-address").out_json().await?;
-    let deposit_address = deposit["address"].as_str().unwrap();
-    let deposit_operation_id = deposit["operation_id"].as_str().unwrap();
-
-    bitcoind
-        .send_to(deposit_address.to_owned(), 100_000) // deposit in sats
-        .await?;
-    bitcoind.mine_blocks(100).await?;
-
-    cmd!(fed, "ng", "await-deposit", deposit_operation_id)
-        .run()
-        .await?;
+    fed.pegin_ng(100_000).await?; // deposit in sats
 
     let post_deposit_walletng_balance = cmd!(fed, "ng", "info").out_json().await?["total_msat"]
         .as_u64()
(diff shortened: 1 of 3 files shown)


devimint/src/main.rs
```diff
@@ -553,13 +553,14 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
     info!("Testing Client-NG");
     // ## reissue e-cash
     info!("Testing reissuing e-cash");
+    const CLIENT_NG_INITIAL_BALANCE: u64 = 10_000_000; // from initial pegin
```
Maan2003, line 556: tests shouldn't have to deal with auto pegin.

should do pegin in dev-fed subcommand handler, not in `dev_fed` function

jp1ac4: OK that makes sense thanks.

I made a naive attempt as follows but then it breaks the old client pegin when running `just mprocs`:
```diff
@@ -1176,8 +1165,9 @@ async fn main() -> Result<()> {
         }
         Cmd::DevFed => {
             let (process_mgr, task_group) = setup(args.common).await?;
-            let _daemons =
+            let daemons =
                 write_ready_file(&process_mgr.globals, dev_fed(&process_mgr).await).await?;
+            daemons.fed.pegin_ng(10_000).await?;
             task_group.make_handle().make_shutdown_rx().await.await?;
         }

```
I'll be travelling a bit for the next few days so will try to resolve soon.

Maan2003: do pegin before calling write_ready_file

jp1ac4: Ah yes thanks, that fixed it :)

jp1ac4:
I've moved the pegin to the `DevFed` subcommand handler (and squashed and rebased commits).

This is ready for review again.

Maan2003 (approved):
Thanks!
Would be nice to move the rest of mprocs startup to devimint.

PR #2594, opened 2023-05-28
feat(client-ng): operation log paging

elsirion:

PR #2609, opened 2023-05-31
feat(cli-ng): list operations

elsirion:
Allows inspecting the operation log using the CLI client. Builds on #2594.

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -1,5 +1,5 @@
 use std::str::FromStr;
-use std::time::Duration;
+use std::time::{Duration, UNIX_EPOCH};
 
 use anyhow::{anyhow, bail};
 use bitcoin::secp256k1;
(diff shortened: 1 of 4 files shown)


fedimint-cli/src/client.rs
```diff
@@ -118,6 +120,10 @@ pub enum ClientCmd {
     },
     /// Print the secret key of the client
     PrintSecret,
+    ListOperations {
```
dpc, line 123: Is this for currently active / all current and past? Would it make sense to add a some filters right away?

fedimint-cli/src/client.rs
```diff
@@ -349,6 +355,51 @@ pub async fn handle_ng_command(
...
+            const ISO8601_CONFIG: iso8601::EncodedConfig = iso8601::Config::DEFAULT
+                .set_formatted_components(iso8601::FormattedComponents::DateTime)
+                .encode();
+            let operations = client
+                .operation_log()
+                .list_operations(limit, None)
+                .await
+                .into_iter()
+                .map(|(k, v)| {
+                    //
```
dpc, line 378: `//` intentional (for formatting maybe?) or leftover?

dpc (dismissed):

fedimint-cli/src/client.rs
elsirion: I'd rather compose queries via `jq`. This is not a user-facing tool.

fedimint-cli/src/client.rs
elsirion: fixed

dpc (approved):

PR #2612, opened 2023-06-01
Use cookie file for auth when username does not exist

VeronicaTee:
To use the cookie file for auth when no username is present, as discussed in #1990


VeronicaTee:
I see some tests failing, I will work on them. I'm also currently working on a test for this, and will put up a comment as soon as it is done. Kindly hold off on reviews, while I ensure the checks pass.

dpc:
Will try to wrap it up in https://github.com/fedimint/fedimint/pull/2793

PR #2614, opened 2023-06-01
fix: dummy client uses wrong amount when receiving funds

m1sterc001guy:
While working on the gateway receive state machines, I found two bugs with the dummy client.

1. `receive_money` adds the current balance from the `DummyOutputOutcome` onto the current balance, instead of the difference.
2. Both the input state and the output state use `Done` has their sync state, so `Done` will be emitted twice which makes it difficult to await on the output of a transaction with both an input and output from the dummy module.

This PR changes:
1. `receive_money` just takes the new balance from the `DummyOutputOutcome`
2. Adds `InputDone` and `OutputDone` so `await_primary_module_output` can just wait for `OutputDone` to be reached.

### DIFF
--- a/modules/fedimint-dummy-client/src/lib.rs
+++ b/modules/fedimint-dummy-client/src/lib.rs
@@ -84,6 +84,17 @@ impl DummyClientExt for Client {
             .finalize_and_submit_transaction(op_id, KIND.as_str(), outpoint, tx)
             .await?;
 
+        // Wait for the output of the primary module
+        if self
+            .await_primary_module_output(op_id, OutPoint { txid, out_idx: 0 })
+            .await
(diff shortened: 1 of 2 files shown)


justinmoon (approved):

modules/fedimint-dummy-client/src/lib.rs
```diff
@@ -284,7 +300,7 @@ impl PrimaryClientModule for DummyClientModule {
                 .filter_map(|state| async move {
                     match state {
                         // Since Done also happens for inputs we will fire too often, but that's ok
```
elsirion, line 302: This comment could be removed now, thx for the refactor!

PR #2616, opened 2023-06-02
feat(gateway-cli): show balance for each federation

### LINKED ISSUE #2508: `gateway-cli info` should show the balance for each federation

jp1ac4:
This is to resolve #2508.

I've added a `total_msat` field to the `FederationInfo` struct, which is the same name used in the output of `fedimint-cli ng info`.

It was suggested in https://github.com/fedimint/fedimint/issues/2508#issuecomment-1546471869 to remove the `balance` command, which I could do here or separately as required.


jp1ac4:
This is still in progress. I need to investigate the failing test.

PR #2617, opened 2023-06-04
Randomly select active gateway

justinmoon:
Without something like this, you can't really run fedimint without multiple gateways. At any given time, all clients will select the same "active gateway".

Also, fixes bug where select_active_gateway would always return an expired gateway if one was returned by server and no active gateway. Luckily, the server also checks that gateways haven't expired.


justinmoon:
> Eventually will be replaced by some fee estimation logic right?

Yes. Something more sophisticated than random.

dpc:
Do we have any client side fee limits? Can anyone join a federation now, charging exuberant fees, and get random traffic?

PR #2621, opened 2023-06-06
Initial Implementation of the atomic broadcast abstraction based on Aleph BFT

joschisan:
# Motivation
Just like the bitcoin network a federation has to agree on a total order of transactions before they can be processed in order to decide which transaction is valid in case of conflicts between transactions. An algorithm which achieves this ordering in a byzantine fault tolerant (BFT) setting, meaning when up to f out of 3f + 1 total nodes are malicious, is called an atomic broadcast. Since its inception, Fedimint used the so called Honey Badger BFT (HBBFT) as its atomic broadcast. However, the algorithm and its implementation have a some critical problems:

* HBBFT trades of transaction latency for throughput by design. Furthermore the transaction latency is very sensitive towards bad network conditions, faulty guardians and number of guardians and all those four factors  compound. @elsirions performance measurements showed that with HBBFT we could only support at most seven guardians in good network condition with at most a thousand users. This was the initial reason that motivated my research into alternative atomic broadcast algorithms and has already been discussed in https://github.com/fedimint/fedimint/discussions/2086 
* If more then f out of 3f + 1 guardians crash a the same time HBBFT may not be able to restart at all.
* HBBFT assumes a reliable network layer, which means that fedimint has to guarantee the delivery of transactions. In case of fellow guardians being offline for a prolonged period this leads to growing in memory message queues.
* HBBFT requires threshold cryptography which relies on a somewhat brittle distributed key generation before the broadcast  can start.

In addition, there is another  problems relating to the broadcast within Fedimint as well. Currently all transactions ever ordered are stored on disk, regardless of them being valid transaction or not. We have to store them in order to catch up crashed nodes or for a guardians recovering from a corruption of its database. This is not only very inefficient as transactions may be submitted by multiple guardians but also opens up an critical DOS vector: any client can sent n pairwise conflicting transactions, such as spending the same cash note, to a guardian each. As all of those transactions are ordered they need to be stored on disk for the entire lifetime of the federation, however, the attacker only pays transaction costs on a single note.

# Design
We will address all these problems at once by switching from using HBBFT directly to our own custom atomic broadcast abstraction based on Aleph BFT, which is an alternative atomic broadcast algorithm. 

The broadcast abstraction orders serialized items in the form of byte vectors. Though the broadcast depends on fedimint_core for PeerId, Encodable and Database it implements no consensus logic specific to Fedimint, to which we will refer as Fedimint Consensus going forward. To the broadcast a consensus item is merely a vector of bytes without any further structure. 

The Broadcast is able to recover from a crash at any time via a backup that it maintains in the servers Database. Furthermore, while we are at it, we will move the recovery for long term offline nodes into this abstraction. The current mechanism is interwoven with the rest of Fedimint, making it difficult to identify which parts of the code are critical for the broadcast to make progress. The broadcast stores the history of items in the form of threshold signed blocks in the database in order to allow for recovery of fellow guardians which have been offline for a prolonged period of time. As our new abstraction now maintains its history on disk we mitigate the described DOS vector by storing an ordered item only on disk long-term if it changes the consensus state. 

This design should solve all problems described previously. Please render the crates documentation for details on the API.

# Context of this pull request
Switching from HBBFT to our abstraction is no trivial matter as Fedimint depends on informal properties of HBBFTs implementation which our new abstraction does not share. The most important example of this the concept of epochs in HBBFT which is for example used to identify and ban malicious peers preventing creation of a threshold signature.

Therefore this pull request is only intended to be the first in a sequence of pull requests, eventually realising the complete switch. As such it is intended to create a first implementation and the abstractions interface and guarantees may still need to be altered. However, it is supposed to be a high-quality solution according to my current understanding of the problem and should be reviewed as such.

### DIFF
--- a/fedimint-atomic-broadcast/Cargo.toml
+++ b/fedimint-atomic-broadcast/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "fedimint-atomic-broadcast"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+aleph-bft = { version = "0.20.5", default-features = false }
+anyhow = "1.0.71"
+async-channel = "1.8.0"
+async-trait = "0.1.68"
+bitcoin = "0.30.0"
+bitcoin_hashes = "0.12.0"
+fedimint-core = { path = "../fedimint-core" }
+futures = "0.3.28"
+parity-scale-codec = "3.5.0"
+secp256k1 = {version = "0.27.0", features=["rand-std","bitcoin-hashes"]}
+tokio = "1.28.1"
+tracing = "0.1.37"
+tracing-subscriber = "0.3.17"
--- a/fedimint-atomic-broadcast/src/broadcast.rs
+++ b/fedimint-atomic-broadcast/src/broadcast.rs
@@ -1,0 +1,155 @@
+use std::time::Duration;
+
+use fedimint_core::db::Database;
+use fedimint_core::task::sleep;
+use fedimint_core::PeerId;
+use tokio::sync::{mpsc, oneshot, watch};
+
+use crate::keychain::Keychain;
+use crate::{db, session, Decision, Message, OrderedItem, Recipient, Shutdown, SignedBlock};
+
+async fn relay_messages(
+    db: Database,
+    incoming_message_receiver: async_channel::Receiver<(Message, PeerId)>,
+    outgoing_message_sender: async_channel::Sender<(Message, Recipient)>,
+    network_data_sender: async_channel::Sender<Vec<u8>>,
+    signed_block_sender: async_channel::Sender<SignedBlock>,
+) {
+    while let Ok((message, peer_id)) = incoming_message_receiver.recv().await {
+        match message {
+            Message::NetworkData(network_data) => {
+                network_data_sender.send(network_data).await.ok();
+            }
+            Message::BlockRequest(index) => {
+                if let Some(signed_block) = db::load_block(&db, index).await {
+                    outgoing_message_sender
+                        .send((Message::Block(signed_block), Recipient::Peer(peer_id)))
+                        .await
+                        .ok();
+
+                    tracing::info!("Served the block with index {} to peer {}", index, peer_id);
+                }
+            }
+            Message::Block(signed_block) => {
+                signed_block_sender.send(signed_block).await.ok();
+            }
+        }
+    }
+
+    std::future::pending().await
+}
+
+async fn process_signed_block(
+    signed_block: SignedBlock,
+    ordered_item_sender: mpsc::Sender<(OrderedItem, u64, oneshot::Sender<Decision>)>,
+) -> anyhow::Result<()> {
+    let mut decision_receivers = vec![];
+    for ordered_item in signed_block.block.items.into_iter() {
+        let (decision_sender, decision_receiver) = oneshot::channel();
+
+        ordered_item_sender
+            .send((ordered_item, signed_block.block.index, decision_sender))
+            .await?;
+
+        decision_receivers.push(decision_receiver);
+    }
+
+    for decision_receiver in decision_receivers {
+        // The threshold signed blocks items have to be accepted by Fedimint Consensus.
+        assert!(decision_receiver.await? == Decision::Accept);
+    }
+
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
+pub async fn run(
+    keychain: Keychain,
+    db: Database,
+    mut index: u64,
+    mempool_item_receiver: async_channel::Receiver<Vec<u8>>,
+    incoming_message_receiver: async_channel::Receiver<(Message, PeerId)>,
+    outgoing_message_sender: async_channel::Sender<(Message, Recipient)>,
+    ordered_item_sender: mpsc::Sender<(OrderedItem, u64, oneshot::Sender<Decision>)>,
+    clean_shutdown_receiver: watch::Receiver<Option<(u64, Duration)>>,
+) -> Shutdown {
+    let (network_data_sender, network_data_receiver) = async_channel::bounded(256);
+    let (signed_block_sender, signed_block_receiver) = async_channel::bounded(16);
+
+    let relay_handle = tokio::spawn(relay_messages(
+        db.clone(),
+        incoming_message_receiver,
+        outgoing_message_sender.clone(),
+        network_data_sender,
+        signed_block_sender,
+    ));
+
+    loop {
+        if let Some(signed_block) = db::load_block(&db, index).await {
+            tracing::info!("Loaded block with index {}", index);
+            if process_signed_block(signed_block, ordered_item_sender.clone())
+                .await
+                .is_err()
+            {
+                relay_handle.abort();
+                relay_handle.await.ok();
+
+                return Shutdown::MidSession(index);
+            }
+        } else {
+            tracing::info!("Run session with index {}", index);
+            let (backup_loader, backup_saver) = db::open_session(db.clone(), index).await;
+
+            let session_result = session::run(
+                index,
+                keychain.clone(),
+                backup_loader,
+                backup_saver,
+                mempool_item_receiver.clone(),
+                network_data_receiver.clone(),
+                outgoing_message_sender.clone(),
+                ordered_item_sender.clone(),
+                signed_block_receiver.clone(),
+            );
+
+            match session_result.await {
+                Ok(signed_block) => {
+                    tracing::info!("Completed session with index {}", index);
+                    db::complete_session(&db, index, signed_block.clone()).await;
+
+                    outgoing_message_sender
+                        .send((Message::Block(signed_block), Recipient::Everyone))
+                        .await
+                        .ok();
+                }
+                Err(..) => {
+                    relay_handle.abort();
+                    relay_handle.await.ok();
+
+                    return Shutdown::MidSession(index);
+                }
+            };
+        }
+
+        let clean_shutdown = clean_shutdown_receiver.borrow().to_owned();
+        if let Some((shutdown_index, shutdown_delay)) = clean_shutdown {
+            if index == shutdown_index {
+                tracing::info!("Initiate clean shutdown after index {}", index);
+
+                // prevents the relay loop from hanging if  the channels to become full
+                network_data_receiver.close();
+                signed_block_receiver.close();
+
+                // we delay the shutdown to allow lagging nodes to complete the session as well
+                sleep(shutdown_delay).await;
+
+                relay_handle.abort();
+                relay_handle.await.ok();
+
+                return Shutdown::Clean(index);
+            }
+        }
+
+        index += 1;
+    }
+}
--- a/fedimint-atomic-broadcast/src/data_provider.rs
+++ b/fedimint-atomic-broadcast/src/data_provider.rs
@@ -1,0 +1,106 @@
+use aleph_bft::Keychain as KeychainTrait;
+use bitcoin_hashes::Hash;
+use tokio::sync::watch;
+
+use crate::consensus_hash_sha256;
+use crate::keychain::Keychain;
+
+type ConsensusItem = Vec<u8>;
+
+// This limits the RAM consumption of a Unit to roughly 12kB
+const ITEM_LIMIT: usize = 100;
+const BYTE_LIMIT: usize = 10_000;
+
+#[derive(
+    Clone, Debug, PartialEq, Eq, Hash, parity_scale_codec::Encode, parity_scale_codec::Decode,
+)]
+pub enum UnitData {
+    Batch(Vec<ConsensusItem>, [u8; 64], aleph_bft::NodeIndex),
+    Signature([u8; 64], aleph_bft::NodeIndex),
+}
+
+impl UnitData {
+    // in order to bound the RAM consumption of a session we have to bound an
+    // individual units size, hence the size of its attached unit data in memory
+    pub fn is_valid(&self) -> bool {
+        match self {
+            UnitData::Signature(..) => true,
+            UnitData::Batch(items, ..) => {
+                // the lazy evaluation prevents overflow when summing
+                items.len() <= ITEM_LIMIT
+                    && items.iter().all(|item| item.len() <= BYTE_LIMIT)
+                    && items.iter().map(Vec::len).sum::<usize>() <= BYTE_LIMIT
+            }
+        }
+    }
+}
+
+pub struct DataProvider {
+    keychain: Keychain,
+    mempool_item_receiver: async_channel::Receiver<ConsensusItem>,
+    signature_receiver: watch::Receiver<Option<[u8; 64]>>,
+    leftover_item: Option<Vec<u8>>,
+}
+
+impl DataProvider {
+    pub fn new(
+        keychain: Keychain,
+        mempool_item_receiver: async_channel::Receiver<ConsensusItem>,
+        signature_receiver: watch::Receiver<Option<[u8; 64]>>,
+    ) -> Self {
+        Self {
+            keychain,
+            mempool_item_receiver,
+            signature_receiver,
+            leftover_item: None,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl aleph_bft::DataProvider<UnitData> for DataProvider {
+    async fn get_data(&mut self) -> Option<UnitData> {
+        // we only attach our signature as no more items can be ordered in this session
+        if let Some(signature) = *self.signature_receiver.borrow() {
+            return Some(UnitData::Signature(
+                signature,
+                self.keychain.peer_id().to_usize().into(),
+            ));
+        }
+
+        let mut items = vec![];
+
+        if let Some(item) = self.leftover_item.take() {
+            if item.len() <= BYTE_LIMIT {
+                items.push(item);
+            } else {
+                tracing::error!("Consensus item length is over BYTE_LIMIT");
+            }
+        }
+
+        // if the channel is empty we want to return the batch immediatly in order to
+        // not delay the creation of our next unit, even if the batch is empty
+        while let Ok(item) = self.mempool_item_receiver.try_recv() {
+            let n_bytes = items.iter().map(Vec::len).sum::<usize>();
+
+            #[allow(clippy::int_plus_one)]
+            if items.len() + 1 <= ITEM_LIMIT && n_bytes + item.len() <= BYTE_LIMIT {
+                items.push(item);
+            } else {
+                self.leftover_item = Some(item);
+                break;
+            }
+        }
+
+        // enables us to verify which peer has submitted a item
+        let hash = consensus_hash_sha256(&items);
+        let signature = self.keychain.sign(hash.as_byte_array()).await;
+
+        let unit_data =
+            UnitData::Batch(items, signature, self.keychain.peer_id().to_usize().into());
+
+        assert!(unit_data.is_valid());
+
+        Some(unit_data)
+    }
+}
--- a/fedimint-atomic-broadcast/src/db.rs
+++ b/fedimint-atomic-broadcast/src/db.rs
@@ -1,0 +1,131 @@
+use std::io::Cursor;
+
+use fedimint_core::db::Database;
+use fedimint_core::encoding::{Decodable, Encodable};
+use fedimint_core::impl_db_record;
+
+use crate::SignedBlock;
+
+#[derive(Debug, Encodable, Decodable)]
+struct SignedBlockKey(u64);
+
+impl_db_record!(
+    key = SignedBlockKey,
+    value = SignedBlock,
+    db_prefix = 0x00,
+    notify_on_modify = false,
+);
+
+#[derive(Debug, Encodable, Decodable)]
+struct UnitsKey(u64, u64);
+
+impl_db_record!(
+    key = UnitsKey,
+    value = Vec<u8>,
+    db_prefix = 0x01,
+    notify_on_modify = false,
+);
+
+pub async fn load_block(db: &Database, index: u64) -> Option<SignedBlock> {
+    db.begin_transaction()
+        .await
+        .get_value(&SignedBlockKey(index))
+        .await
+}
+
+/// This function loads the aleph bft backup from disk and creates a UnitSaver
+/// instance which allows aleph bft to append further bytes to the existing
+/// backup
+pub async fn open_session(db: Database, session_index: u64) -> (Cursor<Vec<u8>>, UnitSaver) {
+    let mut buffer = vec![];
+    let mut units_index = 0;
+    let mut dbtx = db.begin_transaction().await;
+
+    while let Some(bytes) = dbtx.get_value(&UnitsKey(session_index, units_index)).await {
+        buffer.extend(bytes);
+        units_index += 1;
+    }
+
+    std::mem::drop(dbtx);
+
+    tracing::info!("Loaded aleph buffer with {} bytes", buffer.len());
+
+    // the cursor enables aleph bft to read the units via std::io::Read
+    let unit_loader = Cursor::new(buffer);
+
+    // we pass the first free unit index to the UnitSaver as an offset
+    let unit_saver = UnitSaver::new(db, session_index, units_index);
+
+    (unit_loader, unit_saver)
+}
+
+/// The UnitSaver enables aleph bft to store its local directed acyclic graph of
+/// units on disk in order to recover from a mid session crash. By implementing
+/// std::io::Write we allow aleph bft to append bytes to its existing backup
+/// similar to a open file in append mode.
+pub struct UnitSaver {
+    db: Database,
+    session_index: u64,
+    units_index: u64,
+    buffer: Vec<u8>,
+}
+
+impl UnitSaver {
+    fn new(db: Database, session_index: u64, units_index: u64) -> Self {
+        Self {
+            db,
+            session_index,
+            units_index,
+            buffer: vec![],
+        }
+    }
+}
+
+impl std::io::Write for UnitSaver {
+    fn write(&mut self, buffer: &[u8]) -> std::io::Result<usize> {
+        self.buffer.extend_from_slice(buffer);
+        Ok(buffer.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        let units_key = UnitsKey(self.session_index, self.units_index);
+
+        futures::executor::block_on(async {
+            let mut dbtx = self.db.begin_transaction().await;
+
+            dbtx.insert_new_entry(&units_key, &self.buffer).await;
+
+            dbtx.commit_tx_result()
+                .await
+                .expect("This is the only place where we write to this key");
+        });
+
+        self.buffer.clear();
+        self.units_index += 1;
+
+        Ok(())
+    }
+}
+
+/// The function removes the units stored by aleph bft and stores the signed
+/// block instead
+pub async fn complete_session(db: &Database, index: u64, signed_block: SignedBlock) {
+    let mut dbtx = db.begin_transaction().await;
+
+    dbtx.insert_new_entry(&SignedBlockKey(index), &signed_block)
+        .await;
+
+    let mut units_index = 0;
+
+    while dbtx
+        .remove_entry(&UnitsKey(index, units_index))
+        .await
+        .is_some()
+    {
+        units_index += 1;
+    }
+
+    dbtx.commit_tx_result()
+        .await
+        .expect("The function is only called after we have terminated the aleph session");
+}
--- a/fedimint-atomic-broadcast/src/keychain.rs
+++ b/fedimint-atomic-broadcast/src/keychain.rs
@@ -1,0 +1,108 @@
+use std::collections::BTreeMap;
+
+use aleph_bft::Keychain as KeychainTrait;
+use fedimint_core::PeerId;
+use secp256k1::hashes::sha256;
+use secp256k1::{schnorr, Message};
+
+use crate::conversion;
+
+#[derive(Clone, Debug)]
+pub struct Keychain {
+    peer_id: PeerId,
+    public_keys: BTreeMap<PeerId, secp256k1::XOnlyPublicKey>,
+    keypair: secp256k1::KeyPair,
+    secp: secp256k1::Secp256k1<secp256k1::All>,
+}
+
+impl Keychain {
+    pub fn new(
+        peer_id: PeerId,
+        public_keys: BTreeMap<PeerId, secp256k1::XOnlyPublicKey>,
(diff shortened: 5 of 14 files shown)


fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -0,0 +1,148 @@
...
+    db: Database,
+    incoming_message_receiver: async_channel::Receiver<(Message, PeerId)>,
+    outgoing_message_sender: async_channel::Sender<(Message, Recipient)>,
+    network_data_sender: async_channel::Sender<Vec<u8>>,
+    signed_block_sender: async_channel::Sender<SignedBlock>,
+) {
+    while let Ok((message, peer_id)) = incoming_message_receiver.recv().await {
+        match message {
+            Message::NetworkData(network_data) => {
+                network_data_sender.send(network_data).await.ok();
```
jkitman, line 22: instead of `ok()` should we `expect()` since the channel should never close?

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -0,0 +1,148 @@
...
+        incoming_message_receiver.clone(),
+        outgoing_message_sender.clone(),
+        network_data_sender.clone(),
+        signed_block_sender.clone(),
+    ));
+
+    loop {
+        if let Some(signed_block) = db::load_block(&db, index).await {
+            tracing::info!("Loaded block with index {}", index);
+            if process_signed_block(signed_block, ordered_item_sender.clone())
```
jkitman, line 87: How can this error?

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -0,0 +1,148 @@
...
+            };
+        }
+
+        let clean_shutdown = clean_shutdown_receiver.borrow().to_owned();
+        if let Some((shutdown_index, shutdown_delay)) = clean_shutdown {
+            if index == shutdown_index {
+                tracing::info!("Initiate clean shutdown after index {}", index);
+
+                // we delay the shutdown to allow lagging nodes to complete the session as well
+                tokio::time::sleep(shutdown_delay).await;
```
jkitman, line 137: Could this result in inconsistent saved state?

fedimint-atomic-broadcast/src/data_provider.rs
```diff
@@ -0,0 +1,91 @@
...
+        if let Some(signature) = self.signature_receiver.borrow().clone() {
+            return Some(UnitData::Signature(signature, self.keychain.peer_id.into()));
+        }
+
+        let mut n_items = 0;
+        let mut batch_size = 0;
+        let mut items = vec![];
+
+        if let Some(bytes) = self.item.take() {
+            // if the stored item is larger then the BYTE_LIMIT we discard it
```
jkitman, line 58: Should probably log this as an error.

fedimint-atomic-broadcast/src/data_provider.rs
```diff
@@ -0,0 +1,91 @@
...
+            if bytes.len() <= BYTE_LIMIT {
+                n_items += 1;
+                batch_size += bytes.len();
+                items.push(bytes);
+            }
+        }
+
+        while let Ok(item) = self.mempool_item_receiver.try_recv() {
```
jkitman, line 66: What does item_limit control?

Also, why `try_recv`?

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: It is correct that the network data channel and signed block channel actually should never be closed. However, I did generally avoid to panic if there is no irreparable corruption of the systems state. If we use expect and refactor at some point we might overlook it and could get a race condition.

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: process_signed_block errors if the ordered_item_receiver or decision_sender are dropped - we use them as a signal to trigger a mid session shutdown.

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: Do you mean complete_session? In that case I don't think so as we remove the aleph backup buffer and write the signed block to the db in a single transaction. If we have a signed block in the database we do not open the session again if that is what you are concerned by...?

fedimint-atomic-broadcast/src/data_provider.rs
joschisan: I have changed this accordingly.

fedimint-atomic-broadcast/src/data_provider.rs
joschisan: As a batch is a Vec<Vec<u8>> we need to limit the outer Vectors length as well as the sum of the inner Vectors lengths in order to limit the size of the batch in memory. As every node produces 4 batches a second with this limit a federation of four can still process 1600 items a second.

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub fn header(&self) -> [u8; 40] {
+        let mut header = [0; 40];
+
+        header[..8].copy_from_slice(&self.index.to_be_bytes());
+
+        let leaf_hashes = self.items.iter().map(|item| {
+            let mut engine = sha256::HashEngine::default();
+            item.consensus_encode(&mut engine)
+                .expect("Writing to HashEngine cannot fail");
+            sha256::Hash::from_engine(engine)
```
elsirion, line 216: Why not simply call `consensus_hash`? That does the same internally.

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+
+impl Block {
+    /// A blocks header consists of 40 bytes formed by its index in big endian bytes
+    /// concatenated with the merkle root build from the consensus hashes of its [AcceptedItem]s 
+    /// or 32 zero bytes if the block is empty. The use of a merkle tree allows for 
+    /// efficient inclusion proofs of accepted consensus items for clients.
+    pub fn header(&self) -> [u8; 40] {
+        let mut header = [0; 40];
+
+        header[..8].copy_from_slice(&self.index.to_be_bytes());
```
elsirion, line 210: A different way to go about writing to that mutable array would be creating a `Cursor` that operates on it. [Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0edb2a213afe40520ffad87a5d80404d)

fedimint-atomic-broadcast/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "fedimint-atomic-broadcast"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+aleph-bft = { version = "0.20.5", default-features = false }
+anyhow = "1.0.71"
+async-channel = "1.8.0"
+async-trait = "0.1.68"
+bitcoin = "0.30.0"
```
elsirion, line 11: TODO: upgrade the rest of fedimint to rust-bitcoin 0.30

fedimint-atomic-broadcast/src/spawner.rs
```diff
@@ -0,0 +1,29 @@
...
+        let (res_tx, res_rx) = futures::channel::oneshot::channel();
+
+        tokio::spawn(async move {
+            task.await;
+            res_tx.send(()).expect("We own the rx.");
+        });
+
+        Box::pin(async move { res_rx.await.map_err(|_| ()) })
+    }
+}
```
elsirion, line 29: More than adequate for a MVP since aleph seems to do proper task management itself, eventually we might want to integrate with task groups though (ability to monitor what's going on etc.). FYI that we are not using task groups here for now @dpc.

fedimint-atomic-broadcast/src/session.rs
```diff
@@ -0,0 +1,234 @@
...
+    AcceptedItem, Block, Decision, Message, OrderedItem, Recipient, SignedBlock,
+};
+
+// this function completes a session with the following steps:
+// - set up the aleph bft session
+// - periodically request the corresponding signed block from peers
+// - combine the accepted items into a block until we reach a preset
+//   number of ordered batches or a signed block arrives
+// - collect signatures until we reach the threshold or a signed block arrives
+pub async fn run(
```
elsirion: Is it intentional that this isn't a docstring but a normal comment?

elsirion (commented):
Only looked at the first commit and not as deeply as I'd like to, but looks amazing! I don't think any other part of the code is as well-documented :heart_eyes:

fedimint-atomic-broadcast/src/lib.rs
joschisan: Because we need the sha256::Hash for bitcoin_hashes version 0.12.0 instead of 0.11.0 in order to create the merkle root with bitcoin version 0.30.0. When we bump bitcoin_hashes in core to 0.12.0 we can use the consensus hash.

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -0,0 +1,154 @@
...
+                    tracing::info!("Served the block with index {} to peer {}", index, peer_id);
+                }
+            }
+            Message::Block(signed_block) => {
+                signed_block_sender.send(signed_block).await.ok();
+            }
+        }
+    }
+
+    std::future::pending().await
```
dpc, line 41: Why not return from function instead?

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        ordered_item_sender.try_send(OrderedItem {
+            item: item.clone(),
+            block_index: signed_block.block.index,
+            peer_id: *peer_id,
+            decision_sender,
+        })?;
+
+        // The threshold signed blocks items have to be accepted by Fedimint Consensus.
+        assert!(decision_receiver.await? == Decision::Accept);
```
dpc, line 59: Is there a reason not to send all items first, collect the receivers, and only then assert responses? That would eliminate the ping-pong-latency here, afaict.

Also - it looks a bit odd altogether, as I don't understand why would items be sent one by one, instead of once in some vector. (but I'm scanning whole PR top to bottom, so maybe will understand it later).

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -0,0 +1,154 @@
...
+        if let Some((shutdown_index, shutdown_delay)) = clean_shutdown {
+            if index == shutdown_index {
+                tracing::info!("Initiate clean shutdown after index {}", index);
+
+                // prevents the relay loop from hanging if  the channels to become full
+                network_data_receiver.close();
+                signed_block_receiver.close();
+
+                // we delay the shutdown to allow lagging nodes to complete the session as well
+                sleep(shutdown_delay).await;
```
dpc, line 143: Fixed shutdown like this are almost never correct. In practice if the delay is long enough, it might not matter, but just saying.

fedimint-atomic-broadcast/src/data_provider.rs
```diff
@@ -0,0 +1,93 @@
...
+            if n_items < ITEM_LIMIT && batch_size + item.len() <= BYTE_LIMIT {
+                n_items += 1;
+                batch_size += item.len();
+                items.push(item);
+            } else {
+                self.item = Some(item);
```
dpc, line 74: This confused me. Looks like this is a `leftover_item` or something like that, from a previous run. `item` name is too generic, IMO.

fedimint-atomic-broadcast/src/db.rs
```diff
@@ -0,0 +1,133 @@
...
+    db.begin_transaction()
+        .await
+        .get_value(&SignedBlockKey(index))
+        .await
+}
+
+pub async fn open_session(db: Database, index: u64) -> (Cursor<Vec<u8>>, UnitSaver) {
+    let mut buffer = vec![];
+    let mut buffer_index = 0;
+    let mut transaction = db.begin_transaction().await;
```
dpc, line 39: Style: we tend to name these `dbtx` to avoid confusion with a "consensus transaction" etc.

fedimint-atomic-broadcast/src/db.rs
```diff
@@ -0,0 +1,133 @@
...
+    // previously read via the loader - hence we pass the first free buffer
+    // index to the UnitSaver as an offset
+    let saver = UnitSaver::new(db, index, buffer_index);
+
+    (loader, saver)
+}
+
+pub struct UnitSaver {
+    db: Database,
+    index: u64,
```
dpc, line 67: It might be because I don't understand the context, but `index`, `buffer` and `buffer_index` don't seem like great names. When there's an index what's interesting is what is an index of, similar with "buffer". "buffer_index" between a "index" and "buffer" is very confusing. :D

fedimint-atomic-broadcast/src/db.rs
```diff
@@ -0,0 +1,133 @@
...
+            index,
+            buffer_index,
+            buffer: vec![],
+        }
+    }
+}
+
+impl std::io::Write for UnitSaver {
+    fn write(&mut self, buffer: &[u8]) -> std::io::Result<usize> {
+        self.buffer.extend_from_slice(buffer);
```
dpc, line 85: I have no idea what's going on, but I would expect some `flush()` if a certain threshold is crossed.

fedimint-atomic-broadcast/src/keychain.rs
```diff
@@ -0,0 +1,103 @@
...
+use secp256k1::{schnorr, Message};
+
+use crate::conversion;
+
+#[derive(Clone, Debug)]
+pub struct Keychain {
+    pub peer_id: usize,
+    pub peer_count: usize,
+    pub threshold: usize,
+    public_keys: BTreeMap<PeerId, secp256k1::XOnlyPublicKey>,
```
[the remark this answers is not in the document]
dpc: Yeap. So turning this field into public method, might be a good idea.

fedimint-atomic-broadcast/src/keychain.rs
```diff
@@ -0,0 +1,103 @@
...
+            peer_count,
+            threshold,
+            public_keys,
+            keypair,
+            secp,
+        }
+    }
+}
+
+impl aleph_bft::Index for Keychain {
```
dpc, line 43: This seems weird. Maybe there are some reason for it, but "keychain being an index" is weird. Seems like "peer_id is an index" which makes sense, and the fact that keychain has an index is a implementation detail and IMO some abstraction layer is violated here.

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+//! ));
+//! ```
+//!
+//! # The journey of a ConsensusItem
+//!
+//! Let us sketch the journey of an [fedimint_core::epoch::ConsensusItem] into a
+//! signed block.
+//!
+//! * The node which wants to order the item calls consensus_encode to serialize
+//!   it and sends the resulting serialization it to its running atomic
```
dpc, line 43: The second part of this sentence is hard to decipher.

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+//!   broadcast instance via the mempool item sender.
+//! * Every 250ms the broadcasts currently running session instance creates a
+//!   new batch from its mempool and attaches it to a unit in the form of a
+//!   UnitData::Batch. The size of a batch and therefore the size of a
+//!   serialization is limited to 10kB.
```
dpc, line 48: Isn't 10kB kind of ... very little? Every 250ms, it's 40KB/s, hardcoded at this very level.

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+//! * The unit is then included in a [Message] and send to the server via the
```
dpc, line 49: `node` vs `server` is unclear to me here. Isn't everything here p2p, so all things are "peers"?

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+mod spawner;
+
+use bitcoin::merkle_tree;
+use bitcoin_hashes::{sha256, Hash};
+/// This function runs the broadcast until a shutdown is intiated either via the
+/// shutdown sender or the [OrderedItem] receiver is dropped.
+pub use broadcast::run;
+use fedimint_core::encoding::{Decodable, Encodable};
+use fedimint_core::PeerId;
+/// This keychain implements naive threshold schnorr signatures over secp256k1.
```
dpc, line 137: Probably better to move these comments to the item being imported themselves.

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+/// If two correct nodes obtain two ordered items from the broadcast they
+/// are guaranteed to be in the same order. However, an ordered items is
+/// only guaranteed to be seen by all correct nodes if a correct node decides to
+/// accept it.
+#[derive(Debug)]
+pub struct OrderedItem {
+    pub item: Vec<u8>,
+    pub block_index: u64,
+    pub peer_id: PeerId,
+    pub decision_sender: oneshot::Sender<Decision>,
```
dpc, line 179: This field does not seem to belong here. All other fields seem like pure data regarding item and its ordering, and this one is a handle to a logical resource (channel sender).

fedimint-atomic-broadcast/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/// A signed block combines a block with the naive threshold secp schnorr
+/// signature for its header created by the federation. The signed blocks allow
+/// clients and recovering guardians to verify the federations consensus
+/// history. After a signed block has been created it is stored in the database.
+#[derive(Clone, Debug, Encodable, Decodable)]
+pub struct SignedBlock {
+    pub block: Block,
+    pub signature: std::collections::BTreeMap<PeerId, [u8; 64]>,
```
dpc, line 234: signatureS maybe? Maybe `peer_signatures` would be most informative.

fedimint-atomic-broadcast/src/network.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+pub type NetworkData = aleph_bft::NetworkData<
+    Hasher,
+    UnitData,
+    <Keychain as aleph_bft::Keychain>::Signature,
+    <Keychain as aleph_bft::MultiKeychain>::PartialMultisignature,
+>;
+
+pub struct Network {
```
dpc, line 35: Going through this PR I'm generally confused about thread&channel architecture of the whole thing. I think it would be worth drawing via mermaid chart in the main `lib.rs`

Anyway - `Network` does not seem like the right name for two inter-task/thread communication endpoints. Not sure how exactly to name it.

dpc: Oh dear. Now I see it was probably named this way due to `aleph_bft::Network` :shrug:

fedimint-atomic-broadcast/src/network.rs
```diff
@@ -0,0 +1,98 @@
...
+        // parity_scale_codec::Encode to serialize it such that Message can
+        // implement Encodable
+        self.outgoing_message_sender
+            .try_send((Message::NetworkData(network_data.encode()), recipient))
+            .ok();
+    }
+
+    async fn next_event(&mut self) -> Option<NetworkData> {
+        // This limits the RAM consumption of a Unit to roughly 12kB
+        const ITEM_LIMIT: usize = 100;
```
dpc, line 71: Didn't I see these constants before? Should they be one and the same constant?

fedimint-atomic-broadcast/src/network.rs
```diff
@@ -0,0 +1,98 @@
...
+            if let Ok(network_data) = NetworkData::decode(&mut IoReader(&*network_data)) {
+                // in order to bound the RAM consumption of a session we have to bound an
+                // individual units size, hence the size of its attached unitdata in memory
+                if network_data.included_data().iter().all(|unit_data| {
+                    match unit_data {
+                        UnitData::Signature(..) => true,
+                        UnitData::Batch(items, ..) => {
+                            // the lazy evaluation prevents overflow when summing over the item
+                            // sizes
+                            items.len() <= ITEM_LIMIT
```
dpc, line 84: Looks like a candidate for a method somewhere. `fn is_valid(&self) -> bool` or something.

fedimint-atomic-broadcast/src/session.rs
```diff
@@ -0,0 +1,242 @@
...
+
+use crate::conversion::{to_node_index, to_peer_id};
+use crate::data_provider::{DataProvider, UnitData};
+use crate::finalization_handler::FinalizationHandler;
+use crate::keychain::Keychain;
+use crate::network::Network;
+use crate::spawner::Spawner;
+use crate::{db, AcceptedItem, Block, Decision, Message, OrderedItem, Recipient, SignedBlock};
+
+// this function completes a session with the following steps:
```
dpc, line 19: `///` is better than `//` :D

fedimint-atomic-broadcast/src/session.rs
```diff
@@ -0,0 +1,242 @@
...
+        index,
+    );
+
+    // In order to bound a sessions RAM consumption we need to bound its number of
+    // units and therefore its number of rounds. Since we use a session to
+    // create a threshold signature for the corresponding block we have to
+    // guarantee that an attacker cannot exhaust our memory by preventing the
+    // creation of a threshold signature, thereby keeping the session open
+    // indefinitely. Hence we increase the delay between rounds exponentially
+    // such that MAX_ROUND would only be reached after roughly 350 years.
```
dpc, line 55: This seems a bit worrying. Does it mean that one malicious peer could stop the consensus indefinitely?

fedimint-atomic-broadcast/src/session.rs
```diff
@@ -0,0 +1,242 @@
...
+    // In case of such an attack the broadcast stops ordering any items until the
+    // attack subsides.
+    config.max_round = MAX_ROUND;
+    config.delay_config.unit_creation_delay = std::sync::Arc::new(|round_index| {
+        let delay = if round_index == 0 {
+            0.0
+        } else if round_index < EXPONETIAL_SLOWDOWN_OFFSET {
```
dpc, line 62: Wouldn't this branch disappear if `round_index.saturating_sub(EXPONETIAL_SLOWDOWN_OFFSET)` or something like that be used below?

fedimint-atomic-broadcast/src/session.rs
```diff
@@ -0,0 +1,242 @@
...
+    // n_batches_per_block blocks or a signed block arrives from our peers
+    while n_batches < n_batches_per_block {
+        tokio::select! {
+            unit_data = unit_data_receiver.recv() => {
+                if let UnitData::Batch(batch_items, signature, node_index) = unit_data? {
+                    let mut engine = sha256::HashEngine::default();
+                    batch_items
+                        .consensus_encode(&mut engine)
+                        .expect("Writing to HashEngine cannot fail");
+                    let hash = sha256::Hash::from_engine(engine);
```
dpc, line 126: You could throw that hash calculation in some standalone function somewhere. I've seen Eric asked about it and there's some bitcoin crate version issue, but one way or another it should be extractable to a function.

dpc (commented):
Very rudimentary and shallow review. Generally I'd add more smaller comments on most important `struct`s (as opposed to one big on top of the crate), but both are even better.

The biggest gap in my understanding RN is understanding threads/task and communication between them. Tracking all these senders and receivers and spawns manually is quite hard. I wish software industry had a better developed APIs, patterns and language around structured concurrency so it's easier to describe these in words. I hit the same problem in m own projects which usually are heavily parallelized.

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: I fully agree. However, this should down method would only be needed if someone would want to upgrade a federation with a coordinated switch of the fedimint binary after a session. However, since it seems currently unclear if we even want to support this upgrade method I have opted for this simple solution mainly for testing purposes or if one were to shutdown a single guardian at the end of a session without intention to upgrade.

fedimint-atomic-broadcast/src/data_provider.rs
```diff
@@ -0,0 +1,93 @@
...
+                n_items += 1;
+                batch_size += bytes.len();
+                items.push(bytes);
+            } else {
+                tracing::error!("Consensus item length is over BYTE_LIMIT");
+            }
+        }
+
+        while let Ok(item) = self.mempool_item_receiver.try_recv() {
+            if n_items < ITEM_LIMIT && batch_size + item.len() <= BYTE_LIMIT {
```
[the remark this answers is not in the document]
joschisan: Agree. Done.

fedimint-atomic-broadcast/src/data_provider.rs
joschisan: Agree. Done.

fedimint-atomic-broadcast/src/db.rs
joschisan: Done

fedimint-atomic-broadcast/src/network.rs
joschisan: Exactly

fedimint-atomic-broadcast/src/session.rs
joschisan: Only when collaborating with a network adversary powerful enough to completely block f correct nodes.

fedimint-atomic-broadcast/src/keychain.rs
joschisan: The aleph bft API requires its keychain to implement this trait.

fedimint-atomic-broadcast/src/lib.rs
joschisan: I put it there such that all public documentation is in one file

fedimint-atomic-broadcast/src/lib.rs
joschisan: server meant the network layer. Fixed.

fedimint-atomic-broadcast/src/db.rs
```diff
@@ -0,0 +1,133 @@
...
+        .get_value(&SignedBlockKey(index))
+        .await
+}
+
+pub async fn open_session(db: Database, index: u64) -> (Cursor<Vec<u8>>, UnitSaver) {
+    let mut buffer = vec![];
+    let mut buffer_index = 0;
+    let mut transaction = db.begin_transaction().await;
+
+    while let Some(bytes) = transaction
```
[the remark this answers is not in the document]
dpc: Oh. Async. Right.

fedimint-atomic-broadcast/src/network.rs
dpc: It is *some* reason, though I don't think we necessarily need to follow it. I think it's OK if "(our) NetworkComm implements alephf Network", etc.

fedimint-atomic-broadcast/src/session.rs
joschisan: Done

fedimint-atomic-broadcast/src/session.rs
joschisan: Done

fedimint-atomic-broadcast/src/lib.rs
joschisan: Done

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: Well this function models the behaviour of a running session sending ordered items; in the latter case it is a little simpler to do this ping pong style. However, I changed it to use vectors as well

fedimint-atomic-broadcast/src/db.rs
joschisan: I have improved the naming and added some comments.

fedimint-atomic-broadcast/src/db.rs
joschisan: The caller calls flush often enough, about 4 times a second.

fedimint-atomic-broadcast/src/db.rs
```diff
@@ -0,0 +1,133 @@
...
+impl std::io::Write for UnitSaver {
+    fn write(&mut self, buffer: &[u8]) -> std::io::Result<usize> {
+        self.buffer.extend_from_slice(buffer);
+        Ok(buffer.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        let buffer_key = AlephBufferKey(self.index, self.buffer_index);
+
+        futures::executor::block_on(async {
```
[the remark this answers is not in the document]
joschisan: Unfortunately Aleph Bft requires us to implement std::io::Write like this for its backup. In their examples they backup into files were it makes more sense.

fedimint-atomic-broadcast/src/network.rs
joschisan: Yes in data_provider.rs. However, as they only occur in two different places I opted for duplication instead of an import.

fedimint-atomic-broadcast/src/network.rs
joschisan: Done

joschisan: Done

fedimint-atomic-broadcast/src/lib.rs
joschisan: its 40KB/s per guardian per second. So just under 10mb per minute in a federation of four. All of which needs to be stored on disk forever if all items are valid. We could probably set this higher, this is conservative for sure, however we would probably be limited by disk space anyways.

fedimint-atomic-broadcast/src/network.rs
```diff
@@ -0,0 +1,98 @@
...
+            .ok();
+    }
+
+    async fn next_event(&mut self) -> Option<NetworkData> {
+        // This limits the RAM consumption of a Unit to roughly 12kB
+        const ITEM_LIMIT: usize = 100;
+        const BYTE_LIMIT: usize = 10_000;
+
+        while let Ok(network_data) = self.network_data_receiver.recv().await {
+            if let Ok(network_data) = NetworkData::decode(&mut IoReader(&*network_data)) {
```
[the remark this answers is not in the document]
joschisan: That would be an error by parity_scale_codec which should not happen in practice.

PR #2622, opened 2023-06-06
refactor: one route_htlcs instead of per-actor

m1sterc001guy:
We are moving away from `GatewayActor` and instead `Gatewayd` will keep a map of federation clients. To prepare for this change, this PR refactors a bunch of logic away from `GatewayActor` but does not instantiate the new client yet.

Changes:
 - Remove Lightning re-connection logic (can be added back in once we have the new client) https://github.com/fedimint/fedimint/issues/2625
 - `ILnRpcClient` has changed from `Arc<RwLock<dyn ILnRpcClient>>` to `Arc<dyn ILnRpcClient>`. The new client accepts the later.
 - `route_htlcs` has moved from `GatewayActor` to `Gatewayd`. This means we will only have one gRPC channel between the lightning node and `Gatewayd` instead of per-actor.
 - Subscribing to HTLCs has been removed. All HTLCs are sent from the lightning implementations to `gatewayd`. https://github.com/fedimint/fedimint/issues/2626
 - Streaming logic has been simplified. The "server" will send `InterceptHtlcRequest` when an HTLC is intercepted and the "client" (i.e `Gatewayd`) will send `CompleteHtlcResponse` in response to the intercepted HTLC. No other messages are allowed currently.
 - All `TaskGroup`s have been changed to be passed by mutable reference (except the ones that will eventually go away with the new client).

### DIFF
--- a/gateway/ln-gateway/proto/gatewaylnrpc.proto
+++ b/gateway/ln-gateway/proto/gatewaylnrpc.proto
@@ -24,10 +24,14 @@ service GatewayLightning {
    * GatewayLightning implementations should respond with a channel stream
    * over which intercepted HTLCs are continually sent to the client.
    */
-  rpc RouteHtlcs(stream RouteHtlcRequest) returns (stream RouteHtlcResponse) {}
+  rpc RouteHtlcs(stream CompleteHtlcResponse) returns (stream InterceptHtlcRequest) {}
+
+  rpc SubscribeMintHtlcs(SubscribeInterceptHtlcsRequest) returns (EmptyResponse) {}
 }
 
 message EmptyRequest {}
+
+message EmptyResponse {}
 
 message GetNodeInfoResponse {
   // The public key of the associated lightning node
@@ -50,22 +54,6 @@ message PayInvoiceResponse {
   bytes preimage = 1;
 }
 
-// Request to begin intercepting HTLCs or to complete
-// a specific HTLC.
-message RouteHtlcRequest {
-  oneof action {
-    SubscribeInterceptHtlcsRequest subscribe_request = 1;
-    CompleteHtlcsRequest complete_request = 2;
-  }
-}
-
-message RouteHtlcResponse {
-  oneof action {
-    SubscribeInterceptHtlcsResponse subscribe_response = 1;
-    CompleteHtlcsResponse complete_response = 2;
-  }
-}
-
 // Request to subscribe to HTLCs with a specific short channel id
 //
 // Send this request when the gateway just assigned a new channel id to a
@@ -76,7 +64,7 @@ message SubscribeInterceptHtlcsRequest {
   uint64 short_channel_id = 1;
 }
 
-message SubscribeInterceptHtlcsResponse {
+message InterceptHtlcRequest {
   // The HTLC payment hash.
   // Value is not guaranteed to be unique per intercepted HTLC
   bytes payment_hash = 1;
@@ -110,7 +98,7 @@ message SubscribeInterceptHtlcsResponse {
   uint64 htlc_id = 13;
 }
 
-message CompleteHtlcsRequest {
+message CompleteHtlcResponse {
   message Settle {
     // The preimage for settling an intercepted HTLC
     bytes preimage = 1;
@@ -120,6 +108,8 @@ message CompleteHtlcsRequest {
     // The reason for the cancellation of an intercepted HTLC
     string reason = 1;
   }
+
+  message Forward {}
 
   oneof action {
     // Request to complete an intercepted HTLC with success result after
@@ -137,6 +127,9 @@ message CompleteHtlcsRequest {
     // intercepted HTLC. GatewayLightning will fail/cancel the intercepted HTLC
     // with reason provided.
     Cancel cancel = 2;
+
+    // Request to just forward the HTLC without failing or settling it.
+    Forward forward = 3;
   }
 
   // The id of the incoming channel
@@ -145,8 +138,6 @@ message CompleteHtlcsRequest {
   // The index of the incoming htlc in the incoming channel
   uint64 htlc_id = 5;
 }
-
-message CompleteHtlcsResponse {}
 
 message GetRouteHintsResponse {
   message RouteHintHop {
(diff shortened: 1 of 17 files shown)


gateway/ln-gateway/proto/gatewaylnrpc.proto
```diff
@@ -110,7 +98,7 @@ message SubscribeInterceptHtlcsResponse {
   uint64 htlc_id = 13;
 }
 
-message CompleteHtlcsRequest {
+message CompleteHtlcResponse {
```
okjodom, line 101: maybe we should rename to `InterceptHtlcResponse` given it now includes `Forward` outcome

m1sterc001guy: I like that, will change

gateway/ln-gateway/src/lib.rs
```diff
@@ -227,16 +223,93 @@ impl Gateway {
...
 
         Ok(lnrpc)
     }
 
+    pub async fn route_htlcs(&mut self, task_group: &mut TaskGroup) -> Result<()> {
+        // Create a stream used to communicate with the Lightning implementation
+        let (sender, ln_receiver) = mpsc::channel::<InterceptHtlcResponse>(100);
+
+        let mut lnrpc: Box<dyn ILnRpcClient> = match &self.lightning_mode {
+            Some(LightningMode::Cln { cln_extension_addr }) => {
```
okjodom, line 238: - For CLN would we have to make `subscribe_mint_htlcs` rpc before we open the stream with `route_htlcs` rpc? Maybe the CLN extension as currently implemented would just filter out all htlcs without forwarding any to gatewayd

 - If above is true, it's very strange that we need these RPCs to be called in a certain order, and that `subscribe_mint_htlcs` affects behavior of `route_htlcs`.
 
 - Second, we'd need a way of reversing a filter registered by `subscribe_mint_htlcs`

gateway/ln-gateway/src/lib.rs
```diff
@@ -204,19 +205,14 @@ impl Gateway {
         Ok(gw)
```
m1sterc001guy, line 205: need to call `route_htlcs` here too

gateway/ln-gateway/proto/gatewaylnrpc.proto
```diff
@@ -24,11 +24,15 @@ service GatewayLightning {
    * GatewayLightning implementations should respond with a channel stream
    * over which intercepted HTLCs are continually sent to the client.
    */
-  rpc RouteHtlcs(stream RouteHtlcRequest) returns (stream RouteHtlcResponse) {}
+  rpc RouteHtlcs(stream InterceptHtlcResponse) returns (stream InterceptHtlcRequest) {}
+
+  rpc SubscribeMintHtlcs(SubscribeInterceptHtlcsRequest) returns (EmptyResponse) {}
```
m1sterc001guy, line 29: Add comment here

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Offline call with @okjodom. Going to remove the functionality for subscribing to HTLCs for now. This PR doesn't support the multiple `gatewayd` case and implementing that feature with the new client architecture likely requires more careful/scoped implementation in a separate PR.

okjodom: More generally, the paired behavior of `subscribe_mint_htlcs` and `route_htlcs` makes `route_htlcs` rpc unpredictable, and `subscribe_mint_htlcs` does not set clear API expectations to whoever is implementing the rpc.

After my discussion with @m1sterc001guy today, we resolved to remove `subscribe_mint_htlcs` rpc and consider re-introducing filtering on the edge at a later point. See #2626

gateway/ln-gateway/proto/gatewaylnrpc.proto
m1sterc001guy: Removed and modified the comment above

okjodom (approved):
Looks good to go and iterate upon

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -287,18 +288,7 @@ impl ClnRpcService {
...
-                            action: Some(route_htlc_response::Action::CompleteResponse(
-                                CompleteHtlcsResponse {},
-                            )),
-                        }))
-                        .await
-                        .map_err(|e| {
-                            error!("Failed to send CompleteResponse to gatewayd: {:?}", e);
-                        });
-                }
+                Ok(_) => {}
```
justinmoon, line 291: This might look a little nicer with `if let Err(e) = ...`

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -482,51 +472,50 @@ impl GatewayLightning for ClnRpcService {
...
-                        None => {
-                            error!("No action was sent as part of RouteHtlcRequest");
-                        }
-                    }
+                if let Ok(complete_request) = res {
+                    let _ = Self::complete_htlc(complete_request, interceptors.clone())
+                        .await
+                        .map_err(|e| {
+                            error!("CLN extension failed to complete HTLC: {:?}", e);
+                        });
```
justinmoon, line 499: I don't really like this `let _ = ...` pattern. Probably better to also do `if let Err(e) = ...` ?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -358,63 +314,65 @@ impl ILnRpcClient for GatewayLndClient {
...
+                            .map_err(|e| {
+                                error!("Failed to forward HTLC: {:?}", e);
+                            });
                     }
                     None => {
-                        error!("No action was sent as part of RouteHtlcRequest");
+                        error!("No action specified for intercepted htlc. This should not happen. ChanId: {} HTLC ID: {}", incoming_chan_id, htlc_id);
+                        let _ = Self::cancel_htlc(CircuitKey { chan_id: incoming_chan_id, htlc_id }, lnd_sender.clone()).await.map_err(|e| {
+                            error!("Failed to cancel HTLC: {:?}", e);
+                        });
```
justinmoon, line 362: Another one where `if let Err(e) = ...` probably better than `let _  = ...map_err()`

PR #2623, opened 2023-06-06
feat: Gatewayd database

m1sterc001guy:
This PR changes gatewayd's persistence. Instead of persisting `GatewayClientConfig` for each client, it now stores this metadata in a "gatewayd" database. There is still a database per federation client, but now there is an additional database for gatewayd to store metadata.

Right now, gatewayd stores just enough to reconstruct `GatewayClientConfig`, but the downside is that when it re-creates the actors it needs to contact each federation to get the configuration. Once we migrate off of the old client, we can optimize this and store configuration for the new client. This is just a backwards compatible workaround, but allows for easy changing once the NG client is in place.


okjodom (approved):
🚀 

Also opens path to #2624

PR #2628, opened 2023-06-08
feat: register gateway state machine

m1sterc001guy:
Adds a state machine for registering the lightning gateway with the federation.

### DIFF
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -1,6 +1,7 @@
 use fedimint_core::config::FederationId;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::{impl_db_lookup, impl_db_record};
+use fedimint_ln_common::LightningGateway;
 use lightning::routing::gossip::RoutingFees;
 use secp256k1::KeyPair;
 
@@ -8,6 +9,7 @@ use secp256k1::KeyPair;
 #[derive(Clone, Debug)]
 pub enum DbKeyPrefix {
     FederationConfig = 0x04,
+    FederationRegistration = 0x05,
 }
 
 #[derive(Debug, Clone, Encodable, Decodable, Eq, PartialEq, Hash)]
@@ -34,3 +36,14 @@ impl_db_record!(
 );
 
 impl_db_lookup!(key = FederationIdKey, query_prefix = FederationIdKeyPrefix);
+
+#[derive(Debug, Clone, Encodable, Decodable, Eq, PartialEq, Hash)]
+pub struct FederationRegistrationKey {
+    pub id: FederationId,
+}
+
+impl_db_record!(
+    key = FederationRegistrationKey,
+    value = LightningGateway,
+    db_prefix = DbKeyPrefix::FederationRegistration,
+);
--- a/gateway/ln-gateway/src/ng/mod.rs
+++ b/gateway/ln-gateway/src/ng/mod.rs
@@ -39,12 +39,15 @@ use thiserror::Error;
 use url::Url;
 
 use self::pay::{GatewayPayCommon, GatewayPayInvoice, GatewayPayStateMachine, GatewayPayStates};
-use self::register::GatewayRegisterStateMachine;
+use self::register::RegisterWithFederationStateMachine;
 use crate::gatewaylnrpc::{GetNodeInfoResponse, InterceptHtlcRequest};
 use crate::lnrpc_client::ILnRpcClient;
-use crate::ng::register::{GatewayRegisterCommon, GatewayRegisterStates, RegisterGateway};
+use crate::ng::register::{
+    RegisterWithFederation, RegisterWithFederationCommon, RegisterWithFederationStates,
+};
 
-const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
+pub const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
+pub const INITIAL_REGISTER_BACKOFF_DURATION: Duration = Duration::from_secs(15);
 
 /// The high-level state of a reissue operation started with
 /// [`GatewayClientExt::gateway_pay_bolt11_invoice`].
@@ -75,6 +78,7 @@ pub enum GatewayExtReceiveStates {
 pub enum GatewayExtRegisterStates {
     Registering,
     Success,
+    Done,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -103,6 +107,7 @@ pub trait GatewayClientExt {
         &self,
         gateway_api: Url,
         route_hints: Vec<RouteHint>,
+        time_to_live: Duration,
     ) -> anyhow::Result<OperationId>;
 
     /// Attempt fulfill HTLC by buying preimage from the federation
@@ -223,26 +228,32 @@ impl GatewayClientExt for Client {
         &self,
         gateway_api: Url,
         route_hints: Vec<RouteHint>,
+        time_to_live: Duration,
     ) -> anyhow::Result<OperationId> {
         let (gateway, instance) = self.get_first_module::<GatewayClientModule>(&KIND);
         let registration_info =
-            gateway.to_gateway_registration_info(route_hints, GW_ANNOUNCEMENT_TTL, gateway_api);
+            gateway.to_gateway_registration_info(route_hints, time_to_live, gateway_api);
 
         self.db()
             .autocommit(
                 |dbtx| {
                     Box::pin(async {
                         let registration = registration_info.clone();
-                        let operation_id = OperationId(registration.mint_pub_key.serialize());
+                        let operation_id = OperationId(rand::random());
 
                         let state_machines = vec![GatewayClientStateMachines::Register(
-                            GatewayRegisterStateMachine {
-                                common: GatewayRegisterCommon {
+                            RegisterWithFederationStateMachine {
+                                common: RegisterWithFederationCommon {
                                     operation_id,
-                                    time_to_live: GW_ANNOUNCEMENT_TTL,
+                                    time_to_live,
                                     registration_info: registration,
+                                    federation_id: self.get_config().await.federation_id,
                                 },
-                                state: GatewayRegisterStates::Register(RegisterGateway {}),
+                                state: RegisterWithFederationStates::Register(
+                                    RegisterWithFederation {
+                                        backoff_duration: INITIAL_REGISTER_BACKOFF_DURATION,
+                                    },
+                                ),
                             },
                         )];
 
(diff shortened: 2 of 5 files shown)


gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,149 @@
...
+        common: GatewayRegisterCommon,
+    ) -> Vec<StateTransition<GatewayRegisterStateMachine>> {
+        let ttl = common.time_to_live;
+        vec![StateTransition::new(sleep(ttl), move |_, _, _| {
+            Box::pin(Self::transition_wait_for_ttl(common.clone()))
+        })]
+    }
+
+    async fn transition_wait_for_ttl(common: GatewayRegisterCommon) -> GatewayRegisterStateMachine {
+        GatewayRegisterStateMachine {
```
m1sterc001guy, line 144: Instead of immediately transitioning to re-register, we should probably check the database to see if we should re-register or not

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,149 @@
...
+#[cfg_attr(doc, aquamarine::aquamarine)]
+/// State machine that registers the lightning gateway with
+/// the federation.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    Register -- register lightning gateway succeeded --> WaitForTTL
+///    Register -- register lightning gateway failed --> Register
```
okjodom, line 25: Would we ever need to exit this transition if we have continued `Register --> failed --> Register` loop?
The federation might have permanently gone offline (rugged), should we then transition to some recovery / archival states for this federation?

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,149 @@
...
+///    WaitForTTL -- wait for time to live to expire --> Register
+/// ```
+#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
+pub enum GatewayRegisterStates {
```
okjodom, line 29: IMO, `RegisterFederation` and `RegisterFederationStates` might be more inntuitive names for this SM

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,149 @@
...
+    Register(RegisterGateway),
+    WaitForTTL(WaitForTimeToLive),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
+pub struct GatewayRegisterCommon {
```
okjodom, line 35: `RegisterFederationCommon` as a name suggestion change

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,149 @@
...
+/// State machine that registers the lightning gateway with
+/// the federation.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    Register -- register lightning gateway succeeded --> WaitForTTL
+///    Register -- register lightning gateway failed --> Register
+///    WaitForTTL -- wait for time to live to expire --> Register
```
okjodom, line 26: Re-registration after TTL seem like a good opportunity to update the gateway info registered with the Fed. for example gateway might want to update it's node pubkey.

- How would an existing SM know of this info update so it can forward it on next registration?
- Might an existing Register SM be canceled and replaced by a newer one, even before TTL expires?

gateway/ln-gateway/src/ng/register.rs
m1sterc001guy: Now that we have a metadata database for gatewayd, I think we can just check for the presence of the `FederationId` in the db. Un-registering the federation would just delete that key, then when this workflow wakes up it will see that its not there anymore and transition to `Done` so that it terminates.

gateway/ln-gateway/src/ng/register.rs
m1sterc001guy: Open to suggestions on how to handle this. Ideally we wouldn't keep trying over and over again like the PR does now, but we do want to try again in case the federation does come back online. Perhaps exponential backoff would be a good fit here.

gateway/ln-gateway/src/ng/mod.rs
```diff
@@ -197,12 +219,60 @@ impl GatewayClientExt for Client {
...
-        Ok(())
+        let registration_info =
+            gateway.to_gateway_registration_info(route_hints, GW_ANNOUNCEMENT_TTL, gateway_api);
+
+        self.db()
+            .autocommit(
+                |dbtx| {
+                    Box::pin(async {
+                        let registration = registration_info.clone();
+                        let operation_id = OperationId(registration.mint_pub_key.serialize());
```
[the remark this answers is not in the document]
m1sterc001guy: Changed to just use a random OperationId

gateway/ln-gateway/src/ng/register.rs
m1sterc001guy: We now have three cases:

 - `LightningGateway` in the database has been updated with new value -> Transition to `Done`. This means that a new state machine updated the `LightningGateway` and we should just let that one take care of re-registration.
 - `LightningGateway` in the database no longer exists. This means the gateway has been unregistered. Transition to `Done`
 - `LightningGateway` in the database is the same as the previous registration. Transition to `Registering` so that it can be re-registered.

gateway/ln-gateway/src/ng/register.rs
m1sterc001guy: I added a `FailureBackoff` state so that sleeping on failure does not block other state machines. I also implemented exponential backoff (starting at 15 seconds and doubling) with a maximum of one day so that if the federation ever does come back, the gateway can be registered but it won't spam it with requests.

gateway/ln-gateway/src/ng/register.rs
m1sterc001guy: The new implementation of `transition_wait_for_ttl` should make these situations easy. See this comment https://github.com/fedimint/fedimint/pull/2628#discussion_r1224746331

The high level idea is that if the `LightningGateway` registration info is updated in the local db, this state machine will stop, but that means a new one was created which will take care of the registration. The old one just waits for the TTL to expire and then it exits. 

Un-registering a gateway hasn't been implemented yet but it should be trivial. Just delete the `FederationRegistrationKey` from the database.

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+#[derive(Error, Debug, Serialize, Deserialize, Encodable, Decodable, Clone, Eq, PartialEq)]
+pub enum GatewayRegistrationError {
+    #[error("Error registering the Lightning Gateway. Gateway API: {api:?}")]
+    RegistrationError { api: Url },
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
+pub struct RegisterGateway {
```
okjodom, line 92: nit: rename to `RegisterFederation`

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    fn operation_id(&self) -> fedimint_client::sm::OperationId {
+        self.common.operation_id
+    }
+}
+
+#[derive(Error, Debug, Serialize, Deserialize, Encodable, Decodable, Clone, Eq, PartialEq)]
+pub enum GatewayRegistrationError {
```
okjodom, line 86: nit: rename to `RegisterFederationError`

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,260 @@
...
+
+#[cfg_attr(doc, aquamarine::aquamarine)]
+/// State machine that registers the lightning gateway with
+/// the federation.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///    Register -- register lightning gateway succeeded --> WaitForTTL
```
okjodom, line 27: nit:
- /// Register -- register gateway with federation succeeded --> WaitForTTL
- /// Register -- register gateway with federation failed --> FailureBackoff

gateway/ln-gateway/src/ng/register.rs
```diff
@@ -0,0 +1,260 @@
...
+        dbtx: &mut ClientSMDatabaseTransaction<'_, '_>,
+    ) -> RegisterFederationStateMachine {
+        let mut dbtx = dbtx.module_tx();
+        let registration_info = dbtx
+            .get_value(&FederationRegistrationKey {
+                id: common.federation_id,
+            })
+            .await;
+
+        if let Some(registration_info) = registration_info {
```
okjodom, line 233: nice :)

gateway/ln-gateway/src/db.rs
```diff
@@ -1,13 +1,15 @@
...
 use fedimint_core::{impl_db_lookup, impl_db_record};
+use fedimint_ln_common::LightningGateway;
 use lightning::routing::gossip::RoutingFees;
 use secp256k1::KeyPair;
 
 #[repr(u8)]
 #[derive(Clone, Debug)]
 pub enum DbKeyPrefix {
     FederationConfig = 0x04,
+    Registration = 0x05,
```
okjodom, line 12: nit: `RegisterFederation` or `FederationRegister`? Just looking for a prefix which reminds me of  registration scenario

okjodom (commented):
looks good! I only have nits and naming suggestions

gateway/ln-gateway/src/ng/register.rs
m1sterc001guy: Does `RegisterFederation` make more sense? This operation is actually registering the gateway _with_ the federation. This seems to imply that it's registering the federation somehow.

m1sterc001guy: I think I might just change it to `RegisterWithFederation`

gateway/ln-gateway/src/db.rs
m1sterc001guy: I opted for `FederationRegistration`

gateway/ln-gateway/src/ng/register.rs
okjodom: > Un-registering a gateway hasn't been implemented yet but it should be trivial. Just delete the FederationRegistrationKey from the database.

In which case, a registered gateway would continue to serve until TTL expires right?

okjodom (approved):

elsirion:
The current state machine framework currently logs every intermediate state of a state machine forever and doesn't allow pruning yet. This is problematic for long-running state machines with many intermediate steps like the re-registration one here. 

In the case of re-registration there is a simple fix: instead of modeling the state machine the re-registration loop can run in a trigger function that never terminates. The only thing state transitions do is saving progress, but that's not something we care too much about here. If the gateway is restarted and re-registers slightly too often due to that that's no biggie.

Another fix would be adding a property to state machines specifying if past states should be logged (generally turning this off isn't a good option).

justinmoon:
What are the benefits of modeling this as a state machine? Why not just add a method to the Ext trait which spawns a task which tries re-registering in a loop?

elsirion:
Good question :laughing: having one less random dangling task is nice, but you are right, it's not a great fit for the SM executor.

justinmoon:
The state machine abstractions are complex and come with logs of boilerplate so I think we should only use them when they're really needed.

elsirion:
It's also a question of responsibility distribution between gateway and gateway client. I'm coming more and more to the view that registration is a gateway responsibility.

PR #2629, opened 2023-06-09
correct ln-invoice command

vnprc:
This command in the dev-running doc page appears to be out of date. It now requires the `--amount` and `--description` flags before each command line argument.


douglaz (approved):
LGTM

m1sterc001guy (approved):
Thanks! Client is currently undergoing some changes so it's likely that some of the docs are out of date.

PR #2632, opened 2023-06-11
Splits up clients into separate features for wasm-compatibility

jkitman:
- Modifies API to conform to client-side API
- Splits up clients into separate features for wasm-compatibility (should be fixed later)

Some context copied from the comment below:

Currently `fedimint-wallet-client` uses an esplora client directly and not through an interface, which makes mocking it out impossible.

Originally I wanted to create a separate bitcoin client interface for `fedimint-wallet-client` than used in `fedimint-wallet-server`, but @jkitman convinced me that it's not too bad to use the same one, even though not every implementor of `IBitcoindRpc` will support both client and server side usage.

To be able to select which implementations of `IBitcoindRpc` should be included when depending on the bitcoin RPC crate (not all impls support WASM) this PR separates them into different modules and adds feature flags.


fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -314,27 +207,26 @@ impl IBitcoindRpc for ElectrumClient {
...
-            // if transaction is confirmed, we're going to find the confirmation event in
-            // the history of ifs first output
-            Ok(self
-                .0
-                .script_get_history(&output.script_pubkey)?
-                .iter()
-                .any(|history_item| {
-                    (history_item.height as u64) == height && history_item.tx_hash == txid
-                }))
-        })
```
elsirion, line 334: Removing `block_in_place` here could lead to starving other tasks for no good reason since `script_get_history` might block.

elsirion (commented):
Great PR in general, I just don't get why you remove the `block_in_place`s in the first commit. If anything, we should check that they are added everywhere where we have to interact with non-async network/io code.

fedimint-bitcoind/src/bitcoincore.rs
```diff
@@ -0,0 +1,104 @@
...
+use fedimint_core::{apply, async_trait_maybe_send, Feerate};
+use tracing::info;
+use url::Url;
+
+use crate::{DynBitcoindRpc, IBitcoindRpc, IBitcoindRpcFactory, RetryClient};
+
+#[derive(Debug)]
+pub struct BitcoindFactory;
+
+impl IBitcoindRpcFactory for BitcoindFactory {
```
dpc, line 15: Oh dear. Do we really need a whole factory?

fedimint-bitcoind/src/lib.rs
```diff
@@ -45,7 +59,7 @@ pub fn register_bitcoind(kind: String, factory: DynBitcoindRpcFactory) {
     registry.insert(kind, factory);
 }
 
-pub trait IBitcoindRpcFactory: Debug {
+pub trait IBitcoindRpcFactory: Debug + Send + Sync {
```
dpc, line 62: Unclear what `fn create` does and when it can fail. Does it immediately connect?

Interfaces are by their nature the most important to document well.

fedimint-bitcoind/src/bitcoincore_rpc.rs
```diff
@@ -121,17 +121,13 @@ impl IBitcoindRpc for BitcoinClient {
...
-        height: u64,
-    ) -> Result<bool> {
-        let block_hash = self.get_block_hash(height).await?;
-        let block = self.0.get_block(&block_hash)?;
-        Ok(block
-            .txdata
-            .iter()
-            .any(|tx| tx.txid() == transaction.txid()))
+    async fn get_tx_block_height(&self, txid: &Txid) -> Result<Option<u64>> {
+        let info = self.0.get_raw_transaction_info(txid, None).ok();
```
dpc: Are we silently swallowing errors here? Bad for debugging.

dpc:
> Removes blocking and let retry client log errors

Is this part of the description stale now?

I'm confused what is going on in this PR (or rather why). "Add client functionality" - which kind of client? Fedimint client libraries? Why do they need anything to do with bitcoind (or alternative node backends)?

Anyway - left some comments.

elsirion:
@dpc Yes the PR is missing some context. Currently `fedimint-wallet-client` uses an esplora client directly and not through an interface, which makes mocking it out impossible.

Originally I wanted to create a separate bitcoin client interface for `fedimint-wallet-client` than used in `fedimint-wallet-server`, but @jkitman convinced me that it's not too bad to use the same one, even though not every implementor of `IBitcoindRpc` will support both client and server side usage.

To be able to select which implementations of `IBitcoindRpc` should be included when depending on the bitcoin RPC crate (not all impls support WASM) this PR separates them into different modules and adds feature flags.

There's a bunch of unrelated stuff going on, which I'm less sure about.

elsirion:
Also, could you reorder and squash the first and last commit please @jkitman? Undoing changes from an earlier commit in a later one is annoying to review and bad style imo.

fedimint-bitcoind/src/bitcoincore.rs
elsirion: That has been there for a while now (just got moved around), and I think it's needed for dynamic BTC backend choice at runtime.

fedimint-bitcoind/src/lib.rs
elsirion: Conceptually all the `*ModuleGen`s are also factories, might make sense to unify naming.

fedimint-bitcoind/src/bitcoincore.rs
jkitman: Yeah since we need to create multiple clients in places.  This didn't really change but moved, I'll add a doc.

fedimint-bitcoind/src/bitcoincore_rpc.rs
jkitman: Yes because this happens when the tx is not found in the mempool which we treat as non-existing.

I opened this issue for ensuring that consensus agrees on UTXOs https://github.com/fedimint/fedimint/issues/2635

dpc:
Oh. I wasn't aware the wallet's client side needs rpc to bitcoind. Does it mean I won't need to manually fetch tx inclusion proofs etc. now(/soon)?

jkitman:
> Does it mean I won't need to manually fetch tx inclusion proofs etc. now(/soon)?

Yep, I'll be adding that to the interface soon, so we'll have a standard way of communicating with bitcoin RPCs (as well as the ability to mock them out entirely).

elsirion:
> Oh. I wasn't aware the wallet's client side needs rpc to bitcoind. Does it mean I won't need to manually fetch tx inclusion proofs etc. now(/soon)?

Yes, that's how client-NG deposits already work, this PR moves towards making this functionality more testable.

elsirion (approved):

Issue #2625, opened 2023-06-07
Implement Lightning reconnection


okjodom:
#2622 regresses gatewayd functionality by removing Lightning re-connection logic from gatewayd. After successful migration Gateway NG, #2600 , we should  the re-implement LN reconnection logic

justinmoon:
Would we be able to handle this internally within structs with impl `ILnRpcClient`? 

Would be nice if it was an internal implementation detail of node-specific clients and not part of the gateway itself.

PR #2636, opened 2023-06-12
feat: Gatewayd NG Client

m1sterc001guy:
Removes the legacy client from `gatewayd` and uses client NG for payments and receiving.

TODO as follow ups:
- HTLCs are not processed in parallel https://github.com/fedimint/fedimint/issues/2649
- REST requests are not processed in parallel https://github.com/fedimint/fedimint/issues/2637
- Backup/Restore is not implemented https://github.com/fedimint/fedimint/issues/2652
- Lightning Reconnection https://github.com/fedimint/fedimint/issues/2625
- Add module gens the same way that `fedimintd` does https://github.com/fedimint/fedimint/issues/2650
- Fix timestamp bug in registrations that are returned as part of `info` https://github.com/fedimint/fedimint/issues/2651

Test coverage
- `drop_peers_who_dont_contribute_decryption_shares` -> cannot implement with new test framework yet
- `lightning_gateway_pays_internal_invoice` -> not applicable as we removed `gatewayd`'s ability to pay an internal invoice and refactored it to the client
- `lightning_gateway_pays_outgoing_invoice` -> `test_gateway_client_pay_valid_invoice`
- `lightning_gateway_claims_refund_for_internal_invoice` -> not applicable as we removed `gatewayd`'s ability to pay an internal invoice and refactored it to the client
- `set_lightning_invoice_expiry` -> `test_set_lightning_invoice_expiry`
- `receive_lightning_payment_valid_preimage` -> `test_gateway_client_intercept_valid_htlc`
- `receive_lightning_payment_invalid_preimage` -> `test_gateway_client_intercept_htlc_invalid_offer`
- `lightning_gateway_cannot_claim_invalid_preimage` -> `test_gateway_cannot_claim_invalid_preimage`
- `lightning_gateway_can_abort_payment_to_return_user_funds` -> `test_gateway_client_pay_invalid_invoice`


fedimint-testing/src/gateway.rs
```diff
@@ -63,27 +61,26 @@ impl GatewayTest {
...
         .unwrap();
 
         gateway
             .spawn_webserver(listen, password.clone(), &mut task)
             .await;
-        task.spawn("gatewayd", move |handle| async {
-            if let Err(err) = gateway.run(handle).await {
+        let mut subgroup = task.make_subgroup().await;
+        task.spawn("gatewayd", |_| async move {
+            if let Err(err) = gateway.run(&mut subgroup).await {
```
m1sterc001guy, line 83: `run` is deprecated and going to go away since we do not need that event loop. Will remove in a subsequent PR.

gateway/cli/src/main.rs
```diff
@@ -41,17 +39,6 @@ pub enum Commands {
         #[clap(long)]
         federation_id: FederationId,
     },
-    /// Deposit funds into a gateway federation
-    Deposit {
```
m1sterc001guy, line 45: `Deposit` is no longer needed since `deposit_address` will spawn a state machine that will take care of getting the txoutproof and the raw transaction.

gateway/ln-gateway/Cargo.toml
```diff
@@ -79,6 +79,8 @@ fedimint-dummy-common = { path = "../../modules/fedimint-dummy-common" }
 fedimint-ln-client = { path = "../../modules/fedimint-ln-client" }
 fedimint-ln-server = { path = "../../modules/fedimint-ln-server" }
 fedimint-ln-common = { path = "../../modules/fedimint-ln-common" }
+fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
+fedimint-wallet-client = { path = "../../modules/fedimint-wallet-client" }
```
m1sterc001guy, line 83: Ideally we should refactor the tests into a separate crate from `gatewayd`. That would allow us to remove these dependencies for the test crate.

gateway/ln-gateway/src/lib.rs
```diff
@@ -24,49 +23,56 @@ use std::time::Duration;
...
 use axum::response::{IntoResponse, Response};
-use bitcoin::Address;
+use bitcoin::{Address, Txid};
 use bitcoin_hashes::hex::ToHex;
 use clap::Subcommand;
-use fedimint_client::module::gen::ClientModuleGenRegistry;
-use fedimint_client_legacy::ln::PayInvoicePayload;
-use fedimint_client_legacy::modules::ln::route_hints::RouteHint;
-use fedimint_client_legacy::{ClientError, GatewayClient};
+use fedimint_client_legacy::ClientError;
```
m1sterc001guy, line 29: Still need to completely remove the dependency on `fedimint_client_legacy`

gateway/ln-gateway/src/lib.rs
```diff
@@ -275,44 +274,98 @@ impl Gateway {
...
+                        if let Some(federation_id) = federation_id {
+                            let clients = clients.read().await;
+                            let client = clients.get(federation_id);
+                            // Just forward the HTLC if we do not have a client that
+                            // corresponds to the federation id
+                            if let Some(client) = client {
+                                let htlc: Result<Htlc> = htlc_request.clone().try_into().map_err(|_| GatewayError::ClientNgError);
+                                if let Ok(htlc) = htlc {
+                                    let intercept_op = client.gateway_handle_intercepted_htlc(htlc).await;
+                                    // TODO: Refactor this into the state machine so we don't need to wait here
```
m1sterc001guy, line 302: This code is ugly and has lots of indentation. Ideally, I think the response to the lightning implementation should be inside the state machine, so that this loop does not need to wait for a response before handling the next HTLC. However, we can't do this without adding a separate `complete_htlc` RPC instead of using the bidirectional stream, since the bidirectional stream does not survive restarts. Will do this in a separate PR.

gateway/ln-gateway/src/lib.rs
```diff
@@ -523,66 +584,100 @@ impl Gateway {
...
-            .await?
-            .deposit(txout_proof, transaction)
-            .await
+            .get_deposit_address(now() + Duration::from_secs(86400 * 365))
+            .await?;
+        Ok(address)
     }
 
-    async fn handle_withdraw_msg(&self, payload: WithdrawPayload) -> Result<TransactionId> {
+    async fn handle_withdraw_msg(&self, payload: WithdrawPayload) -> Result<Txid> {
```
m1sterc001guy, line 629: Should this return the fedimint `TransactionId` or the bitcoin `Txid`?

gateway/ln-gateway/src/lib.rs
```diff
@@ -523,66 +584,100 @@ impl Gateway {
...
+        unimplemented!("Backup is not currently supported");
     }
 
     async fn handle_restore_msg(
         &self,
-        RestorePayload { federation_id }: RestorePayload,
+        RestorePayload { federation_id: _ }: RestorePayload,
     ) -> Result<()> {
-        self.select_actor(federation_id).await?.restore().await
+        unimplemented!("Restore is not currently supported");
```
m1sterc001guy, line 680: Has backup/restore been implemented for client NG?

integrationtests/tests/tests.rs
```diff
@@ -405,6 +405,7 @@ async fn drop_peers_who_dont_contribute_peg_out_psbts() -> Result<()> {
     .await
 }
 
+/*
```
m1sterc001guy, line 408: Tests are broken, so they are commented out for now

fedimint-testing/src/gateway.rs
m1sterc001guy: https://github.com/fedimint/fedimint/issues/2637

gateway/ln-gateway/Cargo.toml
justinmoon: Let's do this in a follow-up

gateway/ln-gateway/src/client.rs
```diff
@@ -112,47 +61,59 @@ dyn_newtype_define! {
...
-            db_factory,
-            gateway_api,
-        }
+    pub fn new(work_dir: PathBuf) -> Self {
+        Self { work_dir }
     }
 }
 
 #[async_trait]
 impl IGatewayClientBuilder for StandardGatewayClientBuilder {
```
justinmoon, line 73: Is it worth having a trait with only 1 impl?

gateway/ln-gateway/src/client.rs
```diff
@@ -196,26 +147,14 @@ impl IGatewayClientBuilder for StandardGatewayClientBuilder {
...
-                .await?;
-            configs.push(gateway_config);
-        }
-
-        Ok(configs)
+            .collect::<BTreeMap<FederationIdKey, FederationConfig>>()
+            .await
+            .values()
+            .cloned()
+            .collect::<Vec<_>>())
```
justinmoon, line 158: Big improvement

gateway/ln-gateway/src/db.rs
```diff
@@ -37,6 +41,26 @@ impl_db_record!(
...
+        node_pub_key: PublicKey,
+        api: Url,
+    ) -> LightningGateway {
+        LightningGateway {
+            mint_channel_id: self.mint_channel_id,
+            mint_pub_key: self.redeem_key.x_only_public_key().0,
+            node_pub_key,
+            api,
+            route_hints,
+            valid_until: fedimint_core::time::now() + time_to_live,
```
justinmoon, line 58: The way we have an implicit `fedimint_core::time::now()` seems like a foot-gun to me. Let's leave it like this for now, but should fix this somehow ...

gateway/ln-gateway/src/lib.rs
```diff
@@ -163,22 +166,19 @@ impl Gateway {
...
             lightning_mode: Some(lightning_mode),
             fees,
             gatewayd_db,
+            api,
         };
 
-        let (route_hints, node_pub_key) = gw.fetch_route_hints_and_node_pub_key().await?;
-        gw.load_actors(decoders, module_gens, route_hints, node_pub_key)
-            .await?;
+        gw.load_clients(task_group).await?;
```
justinmoon, line 181: Might be nice to make the `clients` a newtype with a contructor that does this automatically.

gateway/ln-gateway/src/client.rs
```diff
@@ -112,47 +61,59 @@ dyn_newtype_define! {
...
+        tg: &mut TaskGroup,
+    ) -> Result<fedimint_client::Client> {
+        let federation_id = config.config.federation_id;
+
+        let db_path = self.work_dir.join(format!("{federation_id}.db"));
+
+        let db =
+            fedimint_rocksdb::RocksDb::open(db_path).map_err(|_| GatewayError::DatabaseError)?;
+
+        // TODO: Should this come from outside?
```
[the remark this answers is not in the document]
m1sterc001guy: Yeah it should probably be structured similarly to `fedimintd`. Also can be done in a follow up.

gateway/ln-gateway/src/db.rs
m1sterc001guy: Yes agreed

gateway/ln-gateway/src/client.rs
m1sterc001guy: I can remove

gateway/ln-gateway/src/lib.rs
```diff
@@ -438,81 +495,93 @@ impl Gateway {
...
+        let (route_hints, node_pub_key, alias) = self.fetch_lightning_route_info().await?;
+        for (federation_id, client) in federation_clients {
+            // TODO: We're reconstructing these registrations, which could have changed in
+            // the meantime, which might break some tests if they're expecting
+            // the same values as the previous registration
+            let (gateway, _) = client.get_first_module::<GatewayClientModule>(&KIND);
+            let registration = gateway.to_gateway_registration_info(
+                route_hints.clone(),
+                GW_ANNOUNCEMENT_TTL,
+                self.api.clone(),
```
justinmoon, line 571: The `valid_until` value here won't be the same as what the registration with the federation since internally this calls `fedimint_core::time::now()`

gateway/ln-gateway/src/lib.rs
justinmoon: Bitcoin transaction ID is far more useful. When withdrawing I'd pop this into mempool.space or something to watch for confirmations.

Maybe it would be good to return both? But let's refine that once we get rid of that `run` loop and all the oneshot sender stuff.

gateway/ln-gateway/src/lib.rs
justinmoon: Yes: https://github.com/fedimint/fedimint/blob/772c23c4dccdf2fa96fb508a1336a1725e9d98c0/fedimint-cli/src/ng.rs#L315C3-L325

Restore currently needs to be called when [initializing a client](https://github.com/fedimint/fedimint/blob/772c23c4dccdf2fa96fb508a1336a1725e9d98c0/fedimint-cli/src/lib.rs#L1086-L1106). We'd probably want to do roughly the same thing. Once it finishes, then just update the `clients` map to contain this new client as the value.

We'll have to be mindful about having 2 clients with same root secret for a federation.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Yes this is a bug. Ultimately I don't think `info` should return the registrations. The only source of truth of that is the federation, so we could just query the federation here, but ideally I think this function shouldn't be dependent on the federation. So I think this should be revisited and the amount of data returned to the client should be slimmed down.

gateway/ln-gateway/Cargo.toml
m1sterc001guy: Unfortunately also needed to add the dummy client here as well. Refactoring to allow `gatewayd` to take in module gens from the outside is too much for this PR.

gateway/ln-gateway/src/lib.rs
```diff
@@ -438,81 +495,93 @@ impl Gateway {
...
+        // time. Just retry after a small delay
+        let gw_client_cfg = loop {
+            match self
+                .client_builder
+                .create_config(connect.clone(), channel_id, fees)
+                .await
+            {
+                Ok(gw_client_cfg) => break gw_client_cfg,
+                Err(_) => {
+                    let random_delay = rand::thread_rng().gen_range(1..=3);
```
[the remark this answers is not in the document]
m1sterc001guy: Changed to be between 0 and 1 second

gateway/ln-gateway/Cargo.toml
m1sterc001guy: Related issue that should ideally be fixed at the same time: https://github.com/fedimint/fedimint/issues/2650

gateway/ln-gateway/src/lib.rs
m1sterc001guy: https://github.com/fedimint/fedimint/issues/2652

gateway/ln-gateway/src/db.rs
m1sterc001guy: https://github.com/fedimint/fedimint/issues/2651

gateway/ln-gateway/src/lib.rs
m1sterc001guy: https://github.com/fedimint/fedimint/issues/2651

okjodom (approved):
lgtm and looking forward to the follow ups planned

PR #2640, opened 2023-06-14
Internal Payment With Markers

### LINKED ISSUE #2618: Internal fedimint transaction still requires the presense of a lightning gateway
While reviewing https://github.com/fedimint/fedimint/pull/2602 I noticed that we still use the lightning gateway's public key to determine if the payment is to another user within the federation by looking at the route hints in the lightning invoice. This works, but it implies that if a user just wants to do an internal transfer within the federation, there still needs to be a lightning gateway even though the gateway is not being used (https://github.com/fedimint/fedimint/pull/2602 refactors the internal logic to be in the client, rather than the gateway).

The main scenario that this would address would be a mint that does not have any lightning gateway connected, but still wants to produce invoices to users to do internal transfers.

Had some offline discussion with @okjodom and @justinmoon. One suggestion was to transform the federation id into an elliptic curve point, then put that in the route hint of the invoice. The client could then directly determine which federation the invoice belonged to, without needing a lightning gateway present.

okjodom:
Implements a [proposal allowing clients to create invoices when there are no gateways connected](https://github.com/fedimint/fedimint/issues/2618).

In this scenario, invoices have a node pubkey and short channel id markers which make them payable within the current federation only, but not from an external payer

closes #2618

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -168,6 +168,24 @@ impl FederationId {
     fn try_from_bytes(bytes: [u8; 48]) -> Option<Self> {
         Some(Self(threshold_crypto::PublicKey::from_bytes(bytes).ok()?))
     }
+
+    /// Converts a federation id to a public key to which we know but discard
+    /// the private key.
+    ///
+    /// Clients MUST never use this private key for any signing operations!
+    ///
+    /// That is ok because we only use the public key for adding a route
+    /// hint to LN invoices that tells fedimint clients that the invoice can
+    /// only be paid internally. Since no LN node with that pub key can exist
+    /// other LN senders will know that they cannot pay the invoice.
+    pub fn to_fake_ln_pub_key(
+        &self,
+        secp: &secp256k1::Secp256k1<secp256k1_zkp::All>,
+    ) -> anyhow::Result<secp256k1::PublicKey> {
+        let bytes = <Sha256 as bitcoin_hashes::Hash>::hash(&self.0.to_bytes()[..]);
+        let sk = secp256k1::SecretKey::from_slice(&bytes)?;
+        Ok(secp256k1::PublicKey::from_secret_key(secp, &sk))
+    }
 }
 
 impl FromStr for FederationId {
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -265,7 +265,18 @@ impl LightningClientExt for Client {
         expiry_time: Option<u64>,
     ) -> anyhow::Result<(OperationId, Invoice)> {
         let (lightning, instance) = self.get_first_module::<LightningClientModule>(&KIND);
-        let active_gateway = self.select_active_gateway().await?;
+        let (src_node_id, short_channel_id, route_hints) = match self.select_active_gateway().await
+        {
+            Ok(active_gateway) => (
+                active_gateway.node_pub_key,
+                active_gateway.mint_channel_id,
+                active_gateway.route_hints,
+            ),
+            Err(_) => {
+                let markers = self.get_internal_payment_markers()?;
+                (markers.0, markers.1, vec![])
+            }
+        };
 
         let (operation_id, invoice, output) = lightning
             .create_lightning_receive_output(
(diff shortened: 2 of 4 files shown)


fedimint-core/src/config.rs
```diff
@@ -168,6 +168,10 @@ impl FederationId {
     fn try_from_bytes(bytes: [u8; 48]) -> Option<Self> {
         Some(Self(threshold_crypto::PublicKey::from_bytes(bytes).ok()?))
     }
+
+    pub fn to_pub_key(&self) -> Result<secp256k1::PublicKey, anyhow::Error> {
```
okjodom, line 172: @jkitman, @elsirion, would you have pointers on how would to derive a valid `secp256k1::PublicKey` from the threshold pubkey used in federation id?

elsirion: I'd do it like this:

```rust
    /// Converts a federation id to a public key to which we don't know the
    /// private key
    ///
    /// That is ok because we only use it for adding a route
    /// hint to LN invoices that tells fedimint clients that the invoice can
    /// only be paid internally. Since no LN node with that pb key can exist
    /// other LN senders will know that they cannot pay the invoice.
    fn to_fake_ln_pub_key(&self) -> secp256k1::PublicKey {
        let mut bytes = [0u8; 33];
        bytes[0] = 0x02;
        bytes[1..].copy_from_slice(&<bitcoin_hashes::sha256::Hash as bitcoin_hashes::Hash>::hash(&self.0.to_bytes()[..])[..]);
        secp256k1::PublicKey::from_slice(&bytes).unwrap()
    }
```

okjodom: Asked for help in [bitcoin stack exchange](https://bitcoin.stackexchange.com/questions/118646/derive-secp256k1-curve-public-key-starting-off-from-known-32-bytes-string?noredirect=1#comment135292_118646)

okjodom: Resolved from Pieter's suggestion in that post

fedimint-core/src/config.rs
```diff
@@ -168,6 +168,22 @@ impl FederationId {
...
     }
+
+    /// Converts a federation id to a public key to which we know but discard
+    /// the private key.
+    ///
+    /// Clients MUST never use this private key for any signing operations!
+    ///
+    /// That is ok because we only use the public key for adding a route
+    /// hint to LN invoices that tells fedimint clients that the invoice can
+    /// only be paid internally. Since no LN node with that pb key can exist
```
m1sterc001guy, line 179: nit: typo on `pb`

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -157,9 +157,24 @@ pub enum LnReceiveState {
...
-    // leads through a registered gateway into the current federation
+async fn invoice_has_internal_payment_markers(
+    invoice: &Invoice,
+    markers: (secp256k1::PublicKey, u64),
+) -> bool {
+    // Asserts that the invoice src_node_id and short_channel_id match known
+    // values used as internal payment markers
+    invoice
+        .route_hints()
+        .first()
```
m1sterc001guy, line 168: Is it guaranteed that the payment markers will be in the first route hint in the vector?

m1sterc001guy (approved):
Looks good to me, just left some nits and questions

fedimint-core/src/config.rs
```diff
@@ -168,6 +168,24 @@ impl FederationId {
...
+    ///
+    /// Clients MUST never use this private key for any signing operations!
+    ///
+    /// That is ok because we only use the public key for adding a route
+    /// hint to LN invoices that tells fedimint clients that the invoice can
+    /// only be paid internally. Since no LN node with that pub key can exist
+    /// other LN senders will know that they cannot pay the invoice.
+    pub fn to_fake_ln_pub_key(
+        &self,
+        secp: secp256k1::Secp256k1<secp256k1_zkp::All>,
```
m1sterc001guy, line 183: Can this be borrowed?

okjodom: ah def!

m1sterc001guy (approved):

PR #2646, opened 2023-06-14
experiment: tests fail when asserting balance sheet sums to zero

justinmoon:
I don't intend to merge this, just sharing some behavior that seems sus to me.

Our federation audits currently assert that liabilities never exceed assets so that federations have the ability to charge fees which could make assets exceed liabilities. I was wondering what would happen if I asserted that assets and liabilities are strictly equal and ran the tests.

`ecash_in_wallet_can_sent_through_a_tx` fails locally, which doesn't seem to do anything with federation fees ...


justinmoon:
Nevermind, `mine_spendable_utxo` just basically hacks the database putting it into this state where assets don't equal liabilities

Issue #2634, opened 2023-06-12
Limit Lightning route hints


elsirion:
Currently invoices will contain route hints for all channels of the gateway's LN node. This leads to big invoices and may even lead to errors since the BOLT11 spec defines a maximum number of hints. When selecting route hints to include we should probably consider incoming liquidity as the main metric.

Issue #2038, opened 2023-03-27
Make extra LN gateway route hints optional


elsirion:
While testing I found out that CLN apparently has problems with seeing itself as the first hop of a route hint and won't route (complains that it can't find *itself*). This sounds similar to the limitation that CLN can't pay itself.

To work around that in small testing networks with only two Lightning nodes we should add a config option or command line flag to the gateway that makes it start as if no extra route hints were available.

A separate issue is why did this not occur in CI so far? The CLN version in the test setup was `v0.12.1`. Or do we start the gateway so quickly that the channel isn't open yet, so not returned as a route hint?

m1sterc001guy:
This was fixed with https://github.com/fedimint/fedimint/pull/2956

PR #2647, opened 2023-06-14
Lightning gateway registers at most 2 route hints

justinmoon:
Chooses the 2 channels with the most incoming capacity

Closes https://github.com/fedimint/fedimint/issues/2634

### DIFF
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -1,5 +1,5 @@
 use std::array::TryFromSliceError;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


justinmoon:
let's also have a bare 1 hop route hint

elsirion:
I'm wondering if the filtering should happen in the gateway instead of in the extension. That way we'd only need to code it once. Otoh we'd need to increase how much data we give to the GW. Either way works, just came to my mind.

justinmoon:
Note: we should also have the client choose a gateway at random if we don't already ...

justinmoon:
> Is this just blocked on rebasing + cleanup? The overall approach looks ok to me (I can take this over, just wondering if there's any outstanding issues)

Yes

m1sterc001guy:
In order to test this properly, I would like to spin up a few additional lightning nodes, create channels, then verify that the route hints are returned correctly. This should be possible after I refactor this PoC to spawn an LDK node https://github.com/fedimint/fedimint/pull/2869

elsirion:
If we made the number of route hints configurable we could fix https://github.com/fedimint/fedimint/issues/2038 at the same time.

justinmoon:
Closing in favor of https://github.com/fedimint/fedimint/pull/2956

PR #2653, opened 2023-06-14
feature: decouple wallet client from depending on esplora

jkitman:
This PR decouples the wallet-client from esplora, so we can run with other clients (currently only bitcoincore), and more importantly write the first wallet tests.


fedimint-bitcoind/src/electrum.rs
```diff
@@ -97,4 +98,22 @@ impl IBitcoindRpc for ElectrumClient {
             }
         }
     }
+
+    async fn watch_script_history(
+        &self,
+        script: &Script,
+    ) -> anyhow::Result<Vec<bitcoin::Transaction>> {
+        let mut results = vec![];
+        let transactions = block_in_place(|| self.0.script_get_history(script))?;
```
dpc, line 107: There's so many `self.0` here that I wonder if we should just `impl Deref`

dpc (approved):

Issue #2637, opened 2023-06-12
Remove gateway's `run` event loop


m1sterc001guy:
In `gatewayd` today we have an event loop that is spawned with `run`. When the webserver receives a request, it currently just sends a message to this event loop which processes it and sends a response back that is eventually passed back to the client that made the initial web request.

This is not necessary. The web server handle for each request can just directly handle the request. This should simplify the code.

Issue #2293, opened 2023-04-21
Switch Gateway APIs to use Websockets


jkitman:
Our federation API uses JSON over WS, but the gateway API uses JSON over HTTPS.  To make it easier on those writing the clients / UI we should standardize on JSON over WS.

@okjodom

elsirion:
I agree. We had this discussion before but working with protobuf/grpc has been a pita due to the poor support for properly typed message structs. With JSON RPC we can just use serde, wich has been working excellently (maybe except for some inefficiencies when dealing with binary data).

okjodom:
Ack. Standardization of the rest webserver makes sense.

> We had this discussion before but working with protobuf/grpc has been a pita due to the poor support for properly typed message structs


Standardization would start from Gateway webserver, before we get to a refactor of the gateway grpc spec for lightning interfaces.

When we get to the grpc interfaces, an additional consideration would be if it's easier to ship JSON rpc spec for external parties to implement lightning extension services to the gateway.

elsirion:
I wonder if that's still worth it or if we maybe want to look into Iroh for the GW API too instead. @m1sterc001guy @joschisan

m1sterc001guy:
Personally I would prefer to stick with axum. I don't see much benefit in switching to websockets. Would that make iroh integration easier?

PR #2654, opened 2023-06-14
feat: remove run loop to allow payments to be parallel

m1sterc001guy:
Fixes https://github.com/fedimint/fedimint/issues/2637

Removes the `run` loop in `gatewayd` to allow each REST request to be processed in parallel. Some changes were needed to make the `Gateway` struct `Clone` so that this is possible.


okjodom (approved):
LGTM. Sets stage for #2293. Will look to pick that up, @m1sterc001guy

PR #2659, opened 2023-06-16
refactor(load-test-tool): use ng cli by default

douglaz:
Plus minor refactorings

### DIFF
--- a/fedimint-load-test-tool/src/common.rs
+++ b/fedimint-load-test-tool/src/common.rs
@@ -3,10 +3,9 @@ use std::str::FromStr;
 use std::time::Duration;
 use std::vec;
 
-use anyhow::{anyhow, bail, Result};
+use anyhow::{bail, Context, Result};
 use bitcoin::secp256k1;
 use devimint::cmd;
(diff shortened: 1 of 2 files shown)


fedimint-load-test-tool/src/common.rs
```diff
@@ -302,92 +298,50 @@ pub async fn remint_denomination(
     Ok(())
 }
 
-pub struct FedimintCli;
-impl ToCmdExt for FedimintCli {
-    type Fut = std::future::Ready<devimint::util::Command>;
+fn get_command_for_alias(alias: &str, default: &str) -> devimint::util::Command {
```
dpc, line 301: That's no really an "alias", but an env var?

dpc (approved):

fedimint-load-test-tool/src/common.rs
douglaz: Yeah. I uses alias because for each of theses variables there is a corresponding alias, like: `alias fedimint-cli='$FM_MINT_CLIENT'`

PR #2660, opened 2023-06-16
test: add wallet tests against real bitcoind

jkitman:
Start testing the new wallet tests against real bitcoind.

Will need more changes to get real LN tests working.

### DIFF
--- a/crypto/aead/src/lib.rs
+++ b/crypto/aead/src/lib.rs
@@ -3,8 +3,8 @@ use std::io::Write;
 use std::path::PathBuf;
 
 use anyhow::{bail, format_err, Result};
-use argon2::password_hash::{Salt, SaltString};
-use argon2::{Argon2, Params, PasswordHasher};
+use argon2::password_hash::SaltString;
+use argon2::{Argon2, Params};
 use rand::rngs::OsRng;
 use rand::Rng;
 use ring::aead::Nonce;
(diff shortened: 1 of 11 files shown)


modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -217,7 +217,7 @@ async fn await_btc_transaction_confirmed(
         };
 
         if !confirmation_height
-            .map(|confirmation_height| consensus_height > confirmation_height)
+            .map(|confirmation_height| consensus_height >= confirmation_height)
```
jkitman, line 220: Found a small off-by-one bug here.

crypto/aead/src/lib.rs
```diff
@@ -98,19 +98,6 @@ pub fn get_encryption_key(password: &str, salt: &str) -> Result<LessSafeKey> {
     Ok(LessSafeKey::new(key))
 }
 
-/// Memory-hard Argon2 key stretching for password-based authentication
-///
-/// * `password` - Strong user-created password
-/// * `salt` - B64 encoded nonce between 4 and 64 bytes
```
jkitman, line 104: Removing some dead code because it will cause tests to fail if we run all the tests with real fixtures eventually.

elsirion (approved):

PR #2666, opened 2023-06-19
wip: don't look

dpc:
just so I don't forget


fedimint-cli/src/ng.rs
```diff
@@ -101,6 +101,10 @@ pub enum ClientNg {
         #[clap(long)]
         force: bool,
     },
+    /// Discover the common api version to use to communicate with the
+    /// federation
+    #[clap(hide = true)]
```
dpc, line 106: Maybe we should move some of these into one hidden `dev` sub-command.

fedimint-client/src/lib.rs
```diff
@@ -693,6 +700,45 @@ impl Client {
             }
         })
     }
+
+    pub async fn discover_common_api_version(&self) -> anyhow::Result<ApiVersionSet> {
```
dpc, line 704: I'm still not sure if this is the best way to name it.

jkitman:
So this allows clients to confirm whether the server supports one of their versions for every module?

And servers can support multiple API versions per module?

fedimint-core/src/module/mod.rs
```diff
@@ -584,29 +597,220 @@ pub struct ApiVersion {
...
+    ) -> result::Result<Self, ApiVersion> {
+        Result::from_iter(iter)
+    }
+
+    /// Insert `version` to the list of supported APIs
+    ///
+    /// Returns `Ok` if no existing element with the same `major` version was
+    /// found and new `version` was successfully inserted. Returns `Err` if
+    /// an existing element with the same `major` version was found, to allow
+    /// modifing its `minor` number. This is useful when merging required /
```
douglaz, line 644:
```suggestion
    /// modifying its `minor` number. This is useful when merging required /
```

fedimint-core/src/query.rs
```diff
@@ -252,6 +256,277 @@ impl<R: Eq + Clone + Debug> QueryStrategy<R> for CurrentConsensus<R> {
...
+        peer: PeerId,
+        result: api::MemberResult<SupportedApiVersionsSummary>,
+    ) -> QueryStep<ApiVersionSet> {
+        match self.inner.process(peer, result) {
+            QueryStep::Success(o) => {
+                match discover_common_api_versions_set(&self.client_versions, o) {
+                    Ok(o) => QueryStep::Success(o),
+                    // TODO: do we need QueryStep::HighLevelFailureOfSomeKind?
+                    Err(e) => QueryStep::Failure(BTreeMap::from([(
+                        PeerId::from(0),
```
douglaz, line 517:
```suggestion
                       peer,
```
?

fedimint-core/src/query.rs
```diff
@@ -252,6 +256,277 @@ impl<R: Eq + Clone + Debug> QueryStrategy<R> for CurrentConsensus<R> {
...
+                if self.responses.len() == self.num_peers {
+                    return QueryStep::Success(mem::take(&mut self.responses));
+                }
+                QueryStep::Continue
+            }
+            // we rely on retries and timeouts to detect a deadline passing
+            Err(_e) => QueryStep::RetryMembers(BTreeSet::from([peer_id])),
+        };
+
+        if Instant::now() <= self.deadline {
```
douglaz, line 296: Not super familiar with this query process, but this `if` seems inverted.

```suggestion
        if Instant::now() > self.deadline {
```

fedimint-core/src/query.rs
```diff
@@ -252,6 +256,277 @@ impl<R: Eq + Clone + Debug> QueryStrategy<R> for CurrentConsensus<R> {
...
+            return QueryStep::Success(mem::take(&mut self.responses));
+        }
+
+        step
+    }
+}
+
+/// Query for supported api versions from all the guardians (with a deadline)
+/// and calculate the best versions to use for each componenet (core + modules).
```
douglaz, line 305:
```suggestion
/// and calculate the best versions to use for each component (core + modules).
```

fedimint-core/src/query.rs
dpc: The failure here is logical and global, and `0` is used only because we have no error case like that yet, yet we want to put some informative error somewhere.

dpc:
@jkitman 

> So this allows clients to confirm whether the server supports one of their versions for every module?

Yes.

> And servers can support multiple API versions per module?

Both client and server components (core + each module) can potentially support multiple "major" api versions. For each component client will figure out a best (most supported & compatible with the client) api version to use.

fedimint-core/src/query.rs
dpc: You're right. Thanks!

fedimint-core/src/query.rs
```diff
@@ -252,6 +256,277 @@ impl<R: Eq + Clone + Debug> QueryStrategy<R> for CurrentConsensus<R> {
...
+                    PeerId(1),
+                    SupportedCoreApiVersions {
+                        core_consensus,
+                        api: MultiApiVersion::try_from_iter([ApiVersion { major: 3, minor: 1 }])
+                            .unwrap(),
+                    }
+                )
+            ])
+        ),
+        Some(ApiVersion { major: 3, minor: 1 })
```
douglaz, line 429: This is a bit confusing, shouldn't we return the major version which is the most supported by the peers, `2` in this case?

fedimint-client/src/module/mod.rs
```diff
@@ -47,6 +47,10 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
 
     fn context(&self) -> Self::ModuleStateMachineContext;
 
+    /// Api versions of the corresponding serverr side module's API
```
m1sterc001guy, line 50: typo: server

fedimint-core/src/module/mod.rs
```diff
@@ -584,29 +597,220 @@ pub struct ApiVersion {
     pub minor: u32,
 }
 
+/// Multiple, disjoin, minimum required or maximum supported, api versions.
```
m1sterc001guy, line 600: nit: disjoint?

fedimint-core/src/module/mod.rs
```diff
@@ -584,29 +597,220 @@ pub struct ApiVersion {
...
+///
+/// If a given component can (potentially) support multiple different (distinct
+/// major number), of an API, this type is used to express it.
+///
+/// Each element must have a distinct major api number, and means
+/// either minimum required API version of this major number, or maximum
+/// supported version of this major number.
```
m1sterc001guy, line 607: Who does the interpretation of this struct? The client?

fedimint-core/src/query.rs
elsirion: A 2.x server would need to at least support minor version 3, so the 2.x servers present here are all out. I hope we'll never have to deal with that grade of incompatibility in one federation though, that would weaken censorship resistance since we essentially only talk to 1 guardian server in this case here.

elsirion (approved):

fedimint-core/src/api.rs
```diff
@@ -153,6 +154,15 @@ pub trait IFederationApi: Debug + MaybeSend + MaybeSync {
     ) -> result::Result<Value, jsonrpsee_core::Error>;
 }
 
+/// Set of api versions for each component (core + modules)
+///
+/// E.g. result of federated common api versions discovery.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ApiVersionSet {
+    pub core: ApiVersion,
+    pub modules: BTreeMap<ModuleInstanceId, ApiVersion>,
```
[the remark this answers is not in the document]
dpc: Yes, thought they do all reset when core consensus version is incremented.

fedimint-core/src/module/mod.rs
dpc: Whatever code using it. But the client common api versions discovery does happen on a client side.

justinmoon:
I think we should also one of these functions before we instantiate the client inside `fedimint-cli` to check we can actually build a compatible client.

dpc:
> I think we should also call one of these functions before we instantiate the client inside `fedimint-cli` to check we can actually build a compatible client.

Yeap. That's the plan: https://github.com/fedimint/fedimint/pull/2710

PR #2669, opened 2023-06-19
fix: declare max fee percent for lnd pay

### LINKED ISSUE #2663: LND gateway implementation doesn't specify max fee
We need to set a fee limit when we pay invoices: https://lightning.engineering/api-docs/api/lnd/lightning/send-payment-sync/index.html#lnrpcfeelimit

okjodom:
closes #2663


gateway/ln-gateway/src/lnd.rs
```diff
@@ -286,12 +289,21 @@ impl ILnRpcClient for GatewayLndClient {
...
+        } = request;
+
         let mut client = self.client.clone();
         let send_response = client
             .lightning()
             .send_payment_sync(SendRequest {
-                payment_request: invoice.invoice.to_string(),
+                payment_request: invoice.to_string(),
+                fee_limit: Some(FeeLimit {
+                    limit: Some(Limit::Percent(max_fee_percent.round() as i64)),
```
okjodom, line 305: - need to validate there's no data loss and safety in casting `max_fee_percent`
- need to validate `max_fee_percent` is respected by LND pay

elsirion: How would this *not* lead to loss of information. Can we just change this limit to an absolute value? I don't know what made me use % in the first pace.

okjodom: LND gRPC as proxied by `tonic_lnd` demands an i64 which seems odd because why would a scenario want to set negative percentages as fee limit? I think this should be a u64 similar to what CLN pay api expects.

m1sterc001guy: It might make sense to cast and then explicitly check if the value is < 0 (and throw error if it is), just to be sure

m1sterc001guy:
What's the plan for validating/testing this?

gateway/ln-gateway/src/lnd.rs
okjodom: Ack. What do you think of pre-computing a fixed value based on the % and fee amount, and rounding that?

elsirion: I'm baffled that the API seems to actually represent a percentage as an integer without any multiplier (e.g. making it a parts per million value instead). We really need to switch to absolute values, integer % granularity is horrible.

https://github.com/lightningnetwork/lnd/blob/0d8d0fd261402ac7b00deec4b01d5f6183302858/lnrpc/lightning.proto#L774C1-L775

justinmoon:
> What's the plan for validating/testing this?

In order to validate this we'd need a new lightning channel which charges more fees than this. Try to pay across it and CLN / LND should reject it.

I wonder how other lightning projects deal with this scenario where a test might want a certain configuration of lightning nodes. Spawning *another* daemon and waiting for a channel open will be kinda slow, especially if we end up with many of these kinds of tests.

gateway/ln-gateway/src/lnd.rs
okjodom: Migrated to absolute values

okjodom: Still, LNDs use of unsigned values might trip us in case we ever need to set very large fee limits `fee_limit_msat > i64::MAX`, but I think this should be a super rare case. Should it happen, this PR proposes we fall back to [LNDs default fee limits as specified here](https://github.com/lightningnetwork/lnd/blob/0d8d0fd261402ac7b00deec4b01d5f6183302858/lnrpc/lightning.proto#L835:L838)

elsirion: `i64::MAX` msat = 92M BTC if I didn't make a mistake, so likely not a problem :laughing:

m1sterc001guy (approved):
LGTM, looks like there is a test failure though that needs investigating

gateway/ln-gateway/src/lnd.rs
```diff
@@ -318,17 +321,38 @@ impl ILnRpcClient for GatewayLndClient {
...
+        let fee_limit = match fee_cap {
+            Ok(fee_cap) => Some(FeeLimit {
+                limit: Some(Limit::FixedMsat(fee_cap)),
+            }),
+            // LND API allows fee limits in the `i64` range, but we use `u64` for max_fee_msat.
+            // This means we can only set an enforceable fee limit between 0 and i64::MAX
+            //
+            // In case the value we try to set is too large for the range allowed by LND API
+            // Avoid setting a fee limit and let LND use its defaults
+            Err(_) => None,
```
elsirion, line 348: The safe thing to do would be to abort the payment instead of proceeding without a limit.

elsirion (changes requested):

okjodom:
Notes from manual testing of this feature:

**Setup 1**:
- Network of two nodes (CLN and LND), each can be a gateway in turn
- **Trigger**: Pay through a gateway node to the peer node
- **Goal**: Validate a `max_fee` is respected by the gateway node when paying invoices on behalf of FM client

Observations
 - **Direct payments between the peers charges zero fees**
 - LND node fails to find path when invoice amount is < fee  set by peer on channel
 - CLN always finds path and makes payment  even when invoice amnt < fee set by peer on channel

elsirion (approved):
Looks good to me

elsirion:
@justinmoon do you think we can merge and figure out testing later or do you want to wait? I'd rather have a manually tested fix for that bug than none.

okjodom:
> LGTM, looks like there is a test failure though that needs investigating

failure was a flaky test

PR #2671, opened 2023-06-19
feat: lightning reconnect

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2625

Allows `gatewayd` to reconnect to the lightning node if the connection drops.

Each `ILnRpcClient` implementation now only holds the connection information and not the actual connect. A connection is created when an RPC is requested. Each implementation will try 10 times to reconnect and fail if it can't connect (so that RPCs don't block indefinitely). 

Since the connections are no longer cached by the `ILnRpcClient` implementation, this might have a small perf overhead.

### DIFF
--- a/devimint/src/lib.rs
+++ b/devimint/src/lib.rs
@@ -47,7 +47,7 @@ pub struct DevFed {
 #[derive(Clone)]
 pub struct Gatewayd {
     _process: ProcessHandle,
-    ln: Option<LightningNode>,
+    pub ln: Option<LightningNode>,
 }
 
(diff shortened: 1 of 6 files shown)


okjodom (approved):
LGTM. Should we add a test for reconnect?

m1sterc001guy:
> LGTM. Should we add a test for reconnect?

There's already a test in devimint. I modified it slightly to do a payment after the disconnect to verify that the HTLC stream is re-created.

PR #2677, opened 2023-06-22
Remove the concept of epochs and dropping of peers from the dummy module

joschisan:


modules/fedimint-dummy-server/src/db.rs
```diff
@@ -62,47 +77,51 @@ pub async fn migrate_to_v1(dbtx: &mut DatabaseTransaction<'_>) -> Result<(), any
...
-impl_db_lookup!(key = DummyFundsKeyV1, query_prefix = DummyFundsKeyV1Prefix);
 
 /// Lookup signature requests by key or prefix
 #[derive(Debug, Clone, Encodable, Decodable, Eq, PartialEq, Hash, Serialize)]
-pub struct DummySignKeyV1(pub String);
+pub struct DummySignatureKey(pub String);
 
 #[derive(Debug, Encodable, Decodable)]
-pub struct DummySignV1Prefix;
+pub struct DummySignaturePrefix;
```
elsirion, line 118: Why the rename? I'm not too hung up about it, but pinging @jkitman. Generally renaming is nicer as a separate commit/PR since it can be done fully automatically and just adds noise.

elsirion (approved):
I'm ok with this, some renames that I'm not sure about but looks good otherwise.

PR #2678, opened 2023-06-22
Remove the concept of epochs and dropping of peers from the mint module

joschisan:


modules/fedimint-mint-server/src/lib.rs
```diff
@@ -363,15 +361,143 @@ impl ServerModule for Mint {
...
+                continue;
+            }
+
+            // check if we already have a valid signature share for this peer
+            if dbtx
+                .get_value(&ReceivedPartialSignatureKey(out_point, peer_id))
+                .await
+                .is_some()
+            {
+                continue;
```
elsirion, line 380: nit: we might want to log this?

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -363,15 +361,143 @@ impl ServerModule for Mint {
...
+            }
+
+            // check if are currently collecting signature shares for this out_point
+            let our_contribution = match dbtx
+                .get_value(&ProposedPartialSignatureKey(out_point))
+                .await
+            {
+                Some(contribution) => contribution,
+                None => {
+                    warn!("Received signature share for non-existent out point");
```
elsirion, line 390: `warn` is correct because we already checked if there is a finalized signature already, otherwise this warning would occur every time all n guardians submit shares since the last `n-t` shares wouldn't be needed.

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -363,15 +361,143 @@ impl ServerModule for Mint {
...
+
+            // check if the received signatures are valid for the reference_messages
+            if !signatures.0.iter_items().zip(reference_messages).all(
+                // the key used for the signature is different for every peer and amount
+                |((amount, (.., sig)), ref_msg)| match self.pub_key_shares[&peer_id].tier(&amount) {
+                    Ok(amount_key) => verify_blind_share(*ref_msg, *sig, *amount_key),
+                    Err(_) => false,
+                },
+            ) {
+                continue;
```
elsirion, line 414: nit: this is probably also a warning

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -363,15 +361,143 @@ impl ServerModule for Mint {
...
+
+            for key in remove_audit_keys {
+                dbtx.remove_entry(&key).await;
+            }
+
+            dbtx.insert_entry(&MintAuditItemKey::IssuanceTotal, &issuances)
+                .await;
+            dbtx.insert_entry(&MintAuditItemKey::RedemptionTotal, &redemptions)
+                .await;
         }
```
elsirion, line 499: I think we need to find a better way to handle audits, doing them on every consensus item but *not* when transactions are processed seems too random. Previously this was done after every epoch and mechanical migrations of the code got us here. @jkitman do you have enough context to think about this? I could we move the DB compaction into the audit function of the module?

elsirion (approved):
Looks good, only minor nits

modules/fedimint-mint-server/src/lib.rs
joschisan: if all nodes are correct and contribute we would log this f times for every created blind signature though.

modules/fedimint-mint-server/src/lib.rs
joschisan: Done

modules/fedimint-mint-server/src/lib.rs
joschisan: Done

modules/fedimint-mint-server/src/lib.rs
joschisan: To me it seems like it does not matter when the DB compaction runs (could even be different for every peer). Sensible options going forward seem to be either before every audit, hence after every transaction, or after every session - right now adding a function like on_session_switch to the ServerModule trait seems to make sense for other reasons anyways. However, for this pr we can keep it where it is and I can make a note to move it to a more sensible option later...

elsirion (approved):

PR #2679, opened 2023-06-22
Remove the concept of epochs and dropping of peers from the wallet module

joschisan:


elsirion:
You can delete `db/migrations/wallet-v0` and run the `prepare_migration_snapshots` test in `fedimint-wallet-server` to fix the migration test. This means you acknowledge breaking backwards compatibility, but since there isn't a release yet that's ok.

modules/fedimint-wallet-common/src/lib.rs
```diff
@@ -37,15 +37,24 @@ pub type PegInDescriptor = Descriptor<CompressedPublicKey>;
     Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, UnzipConsensus, Encodable, Decodable,
 )]
 pub enum WalletConsensusItem {
-    RoundConsensus(RoundConsensusItem),
+    BlockHeight(u32), /* FIXME: use block hash instead, but needs more complicated
+                       * * verification logic */
```
elsirion, line 41: First time we used multi-line comments like this afaik, not a problem but a bit odd.

recoverytool/src/main.rs
```diff
@@ -262,28 +262,8 @@ fn input_tweaks_output_present(
...
-                .downcast_ref::<WalletConsensusItem>()
-                .expect("Instance id mapping incorrect");
-            match wci {
-                WalletConsensusItem::RoundConsensus(rci) => Some(rci.randomness),
-                WalletConsensusItem::PegOutSignature(_) => None,
-            }
-        })
-        .fold([0; 32], xor)
+fn round_tweak(_module_cis: impl Iterator<Item = DynModuleConsensusItem>) -> [u8; 32] {
+    unimplemented!()
```
elsirion, line 266: Could refer to the issue in a TODO

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -329,51 +321,100 @@ impl ServerModule for Wallet {
...
+
+        if block_height_proposal != self.consensus_block_height(dbtx).await {
+            items.push(WalletConsensusItem::BlockHeight(block_height_proposal));
         }
+
+        let fee_rate_proposal = self.fee_rate().await;
+
+        if fee_rate_proposal != self.consensus_fee_rate(dbtx).await {
+            items.push(WalletConsensusItem::Feerate(fee_rate_proposal));
+        }
```
elsirion, line 340: The fee rate changing doesn't trigger an epoch, but that's not a problem (and indeed the old behavior too) but we'll have to think about how to propose consensus items in an AlephBFT world and how that would change the module interface.

E.g. `consensus_proposal` could return a stream and the Fedimint consensus just combines all these streams and sends their output to AlephBFT. But maybe you already had something else in mind and I forgot.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -329,51 +321,100 @@ impl ServerModule for Wallet {
...
     ) -> Vec<PeerId> {
         trace!(?consensus_items, "Received consensus proposals");
 
-        // Separate round consensus items from signatures for peg-out tx. While
-        // signatures can be processed separately, all round consensus items
-        // need to be available at once.
-        let UnzipWalletConsensusItem {
-            peg_out_signature: peg_out_signatures,
-            round_consensus: round_items,
-        } = consensus_items.into_iter().unzip_wallet_consensus_item();
```
elsirion, line 359: I wonder where this is still used now, would be nice if we could remove it (iirc the impl is generated by a proc macro)

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -329,51 +321,100 @@ impl ServerModule for Wallet {
...
+                                self.sync_up_to_consensus_height(
+                                    dbtx,
+                                    old_consensus_height,
+                                    new_consensus_height,
+                                )
+                                .await;
+                            }
+                        }
+                        Ordering::Less => {
+                            warn!("Peer block height vote decreased");
```
elsirion, line 381: We might want to log the peer id here.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -329,51 +321,100 @@ impl ServerModule for Wallet {
...
+                        }
+                        Ordering::Equal => {}
```
elsirion, line 383: Can this happen with correct peers?

elsirion (approved):
Looks good, some nits, lmk if you want me to merge+follow up or want to amend this PR.

elsirion:
Nevermind, the only real nit was "log peer id", merged.

modules/fedimint-wallet-server/src/lib.rs
joschisan: Yes I do have something in mind: consensus_proposal can be simplified to always return all possible consensus items on every call; we call consensus_proposal every second or so (as only the transactions are latency critical) and send all of them to the broadcast. The broadcast then filters the redundant items within a session based on their consensus_hash. So in this case we would simply send the current feerate to the broadcast which would submit it only if it changed.

modules/fedimint-wallet-server/src/lib.rs
joschisan: Less could not happen. If we filter redundant items within a session equal may happen if submitted in two different sessions by a valid peer.

PR #2680, opened 2023-06-22
Remove the concept of epochs and dropping of peers from the lightning module

joschisan:


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -855,16 +840,29 @@ impl Lightning {
...
+
+        let mut heights = vec![];
+
+        for index in 0..peer_count {
+            heights.push(
+                dbtx.get_value(&BlockHeightVoteKey(PeerId::from(index as u16)))
+                    .await
+                    .unwrap_or(0),
+            )
         }
```
elsirion, line 861: Could this be a single call to `dbtx.find_by_prefix` instead?

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -855,16 +840,29 @@ impl Lightning {
         Ok(Lightning { cfg, btc_rpc })
     }
 
-    /// Returns the median block height if enough peers have contributed
-    pub async fn consensus_height(&self, dbtx: &mut ModuleDatabaseTransaction<'_>) -> Option<u64> {
-        let peer_heights = dbtx.get_value(&BlockHeightKey).await.unwrap_or_default();
-        if peer_heights.len() < self.cfg.consensus.threshold() {
-            return None;
+    pub async fn block_height(&self) -> u64 {
```
elsirion, line 843: This function seems a bit dangerous, accidental use in consensus logic would break consensus.

elsirion (changes requested):
Some minor nits, one DoS vector I'd like to discuss/solve before merging.

modules/fedimint-ln-common/src/db.rs
```diff
@@ -125,11 +127,18 @@ impl_db_lookup!(
 );
 
 #[derive(Debug, Encodable, Decodable, Serialize)]
-pub struct BlockHeightKey;
+pub struct BlockHeightVoteKey(pub PeerId);
```
m1sterc001guy, line 130: Do we want this to be backwards compatible? This will require a database migration

elsirion: Probably not until we do a release.

modules/fedimint-ln-server/src/lib.rs
joschisan: Done

modules/fedimint-ln-server/src/lib.rs
joschisan: We could rename it to local_block_height or something?

elsirion (approved):
I'm ok with this version, maybe you can squeeze in the rename in the next PR as a separate commit :)

PR #2681, opened 2023-06-22
feat: fedimint-metrics same process

douglaz:
Alternative to https://github.com/fedimint/fedimint/pull/2673 with metrics being collected directly from the internal fedimintd events.


fedimintd/src/fedimintd.rs
```diff
@@ -257,5 +260,24 @@ async fn run(
...
+    api_result?;
+    metrics_api_result?;
+    Ok(())
+}
+
+async fn spawn_metrics_server(
+    bind_address: &SocketAddr,
+    mut task_group: TaskGroup,
+) -> anyhow::Result<()> {
+    let rx = fedimint_metrics::run_api_server(bind_address, &mut task_group).await?;
```
dpc, line 276: We don't have to spawn anything (other than maybe prom http request handling thread on another port if we want one) and run any periodic metrics update loop. You can see how people usually handle it in some existing projects. I looked for dependents of `prometheus` and `electrs` is one:

Seems like they pass `Metrics` handle to the each major component like Mempool](https://github.com/romanz/electrs/blob/c473556401fd6e56f3411a668a5da655b4298bac/src/mempool.rs#L50), then [update each metric to the right value somewhere next to where the corresponding logic handles it](https://github.com/romanz/electrs/blob/c473556401fd6e56f3411a668a5da655b4298bac/src/mempool.rs#L140).

That's also how I would do it, except maybe I would make all metrics static globals because I'm lazy, and by definition one will have one metrics per the whole process anyway and it's not a library, but an application.

douglaz: This server is to serve the metric results, not to run the metrics update loop

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -542,8 +542,24 @@ impl ServerModule for Lightning {
                         amount: amount.amount,
                         contract: contract.contract.clone().to_funded(out_point),
                     });
-                dbtx.insert_entry(&contract_db_key, &updated_contract_account)
+                let previous = dbtx
+                    .insert_entry(&contract_db_key, &updated_contract_account)
                     .await;
+                let inserted = previous.is_none();
+                if inserted {
```
m1sterc001guy, line 549: I assume if this value is overwritten (i.e previous is not none) that's unexpected behavior? Should we explicitly check for that?

m1sterc001guy:
This approach makes sense to me. I'm hoping to get some time to try it soon. A screenshot of an example graph would be great!

douglaz:
> This approach makes sense to me. I'm hoping to get some time to try it soon. A screenshot of an example graph would be great!

One easy way to see the metrics is just to do some payments then run something like:
```bash
❯ curl -sq http://localhost:3000/metrics   
# HELP ln_funded_contract_incoming contracts::FundedContract::Incoming
# TYPE ln_funded_contract_incoming counter
ln_funded_contract_incoming 1
# HELP ln_funded_contract_incoming_account_amounts_sats contracts::FundedContract::Incoming account amount in sats
# TYPE ln_funded_contract_incoming_account_amounts_sats histogram
ln_funded_contract_incoming_account_amounts_sats_bucket{le="0"} 0
ln_funded_contract_incoming_account_amounts_sats_bucket{le="0.5"} 0
ln_funded_contract_incoming_account_amounts_sats_bucket{le="1"} 1
ln_funded_contract_incoming_account_amounts_sats_bucket{le="1000"} 1
ln_funded_contract_incoming_account_amounts_sats_bucket{le="+Inf"} 1
ln_funded_contract_incoming_account_amounts_sats_sum 1
ln_funded_contract_incoming_account_amounts_sats_count 1
# HELP ln_funded_contract_outgoing contracts::FundedContract::Outgoing
# TYPE ln_funded_contract_outgoing counter
ln_funded_contract_outgoing 7
# HELP ln_funded_contract_outgoing_account_amounts_sats contracts::FundedContract::Outgoing account amounts in sats
# TYPE ln_funded_contract_outgoing_account_amounts_sats histogram
ln_funded_contract_outgoing_account_amounts_sats_bucket{le="0"} 0
ln_funded_contract_outgoing_account_amounts_sats_bucket{le="0.5"} 0
ln_funded_contract_outgoing_account_amounts_sats_bucket{le="1"} 0
ln_funded_contract_outgoing_account_amounts_sats_bucket{le="1000"} 7
ln_funded_contract_outgoing_account_amounts_sats_bucket{le="+Inf"} 7
ln_funded_contract_outgoing_account_amounts_sats_sum 7.069999999999999
ln_funded_contract_outgoing_account_amounts_sats_count 7
# HELP ln_incoming_offer contracts::IncomingContractOffer
# TYPE ln_incoming_offer counter
ln_incoming_offer 1
# HELP ln_output_outcome_cancel_outgoing_contract LightningOutputOutcome::CancelOutgoingContract
# TYPE ln_output_outcome_cancel_outgoing_contract counter
ln_output_outcome_cancel_outgoing_contract 0
```

So this is the result after doing 7 outgoing payments and 1 incoming payment.

Here is a example graph with two periods of tests. The first graph is the absolute counter value (which can be resetted if the server is restarted) and the second is the rate of change (payments per second) using
https://prometheus.io/docs/prometheus/latest/querying/functions/#rate over the above metric:

image

elsirion (commented):
Both this PR and #2673 have the problem that they only work for default modules, but afaik these modules being absent shouldn't be a problem. I wonder if there is a more dynamic way of registering metrics.

I also had a misconception that in-process metrics would require persistent data storage, but it appears that on restart we are ok with just resetting counters etc. Given that I'm not that opposed to in-process metrics anymore, the only thing that would probably be useful is making it optional, i.e. not spawning metrics server if the env var isn't provided.

dpc:
> I also had a misconception that in-process metrics would require persistent data storage, but it appears that on restart we are ok with just resetting counters etc. 

Prometheus detects counters being decremented after reset (to near 0), and handles it correctly. It's not 100% accurate, but in practice good enough.

douglaz:
Addressed both concerns:

> Both this PR and #2673 have the problem that they only work for default modules, but afaik these modules being absent shouldn't be a problem. I wonder if there is a more dynamic way of registering metrics.

It is actually dynamic because we are using the global registry of the prometheus crate. To make it more clear I moved the metric definitions from the fedimint-metrics crate to the module itself. As you can see, any custom module could define its own metrics this way.

> I also had a misconception that in-process metrics would require persistent data storage, but it appears that on restart we are ok with just resetting counters etc. Given that I'm not that opposed to in-process metrics anymore, the only thing that would probably be useful is making it optional, i.e. not spawning metrics server if the env var isn't provided.

Now the metrics api server won't spawn by default unless you give `FM_BIND_METRICS_API` or `--bind-metrics-api`.

douglaz:
> > I also had a misconception that in-process metrics would require persistent data storage, but it appears that on restart we are ok with just resetting counters etc.
> 
> Prometheus detects counters being decremented after reset (to near 0), and handles it correctly. It's not 100% accurate, but in practice good enough.

The grafana agent (which is supposed to be a drop-in replacement to prometheus) doesn't seems to handle that so gracefully, as you can see in the graph I posted above. But I think it's still useful as a relative metric (using for instance the `rate` function).

dpc:
> doesn't seems to handle that so gracefully, as you can see in the graph I posted above.

This might be just an artifact of not aggregating across multiple instances of the same metric with different attributes.

Usually it's something like `pod=myservice-q8hskjdfkjd` vs `pod=myservice-jkhhsflsdf` in k8s. But it could be something like `pid=1234` vs `pid=2345` locally.

If you do `sum(counter)`, possibly with some `by`/`without` it should get fixed.

You want to make sure to calculate `rate(...)` of the above as well.

The only imprecision in prometheus counters on reset is that it can miss the last few counts. If prometheus took a snapshot with count=10, then event happened and count=11, and service restarted, prom will query it after restart and count=0 (or 1,  3, whatever), that one count will never be accounted for. But that barely every matters.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -43,13 +43,64 @@ use fedimint_ln_common::{
...
+    Histogram, IntCounter,
+};
 use fedimint_server::config::distributedgen::PeerHandleOps;
 use futures::StreamExt;
 use rand::rngs::OsRng;
 use serde::{Deserialize, Serialize};
 use strum::IntoEnumIterator;
 use tracing::{debug, error, info_span, instrument, trace, warn};
 
+lazy_static! {
```
elsirion, line 57: Defining these here really made it click for me!

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -43,13 +43,64 @@ use fedimint_ln_common::{
...
+    pub static ref LN_INCOMING_OFFER: IntCounter = register_int_counter!(opts!(
+        "ln_incoming_offer",
+        "contracts::IncomingContractOffer"
+    ))
```
elsirion, line 61: These don't seem to be module instance aware, but I really don't mind for now.

elsirion (approved):
Looks gerat! I'd go with this one, less deployment complexity and no non-consensus DB writes which were my initial concern with in-process.

PR #2682, opened 2023-06-22
Version hash test

vnprc:
Add a devimint version hash test that compares the result of the `version-hash` command of each binary against the git HEAD commit hash.

This change includes two fixes:
- devimint set_code_version() was failing to set the correct commit hash on a few binaries due to a reliance on a hardcoded relative filepath. The fix was to remove the check for the HEAD file and just run the `git rev-parse HEAD` command directly, writing errors to stderr.
- The gateway-cli binary `version-hash` command returned a different format than all other binaries. The fix was to remove the extraneous "version: " text and only return the hash.

### DIFF
--- a/fedimint-build/src/lib.rs
+++ b/fedimint-build/src/lib.rs
@@ -1,21 +1,31 @@
 use std::env;
-use std::fs::File;
+use std::io::{self, Write};
 use std::process::Command;
+
+const FAKE_COMMIT_HASH: &str = "01234569afbe457afa1d2683a099c7af48a523c1";
 
 pub fn set_code_version() {
     if env::var_os("CODE_VERSION").is_none() {
-        let git_hash = if File::open("./../../.git/HEAD").is_ok() {
-            let output = Command::new("git")
-                .args(["rev-parse", "HEAD"])
-                .output()
-                .unwrap();
-            String::from_utf8(output.stdout).unwrap()
-        } else {
-            // we set a fake hash here and make it easily recognizable
-            // by giving it a prefix (0x0123456)
-            String::from("01234569afbe457afa1d2683a099c7af48a523c1")
-        };
-        println!("cargo:rustc-env=CODE_VERSION={git_hash}");
+        let output = Command::new("git").args(["rev-parse", "HEAD"]).output();
+
+        match output {
+            Ok(output) => {
+                if output.status.success() {
+                    let git_hash = String::from_utf8_lossy(&output.stdout).trim().to_string();
+                    println!("cargo:rustc-env=CODE_VERSION={}", git_hash);
+                } else {
+                    let error_message = String::from_utf8_lossy(&output.stderr);
+                    io::stderr().write_all(error_message.as_bytes()).unwrap();
+                    println!("cargo:rustc-env=CODE_VERSION={}", FAKE_COMMIT_HASH);
+                }
+            }
+            Err(error) => {
+                io::stderr()
+                    .write_fmt(format_args!("Failed to execute git command: {}", error))
+                    .unwrap();
+                println!("cargo:rustc-env=CODE_VERSION={}", FAKE_COMMIT_HASH);
+            }
+        }
     }
     println!("cargo:rerun-if-env-changed=CODE_VERSION");
 }
--- a/fedimint-build/src/lib.rs
+++ b/fedimint-build/src/lib.rs
@@ -1,31 +1,42 @@
 use std::env;
-use std::io::{self, Write};
-use std::process::Command;
+use std::process::{Command, Stdio};
 
 const FAKE_COMMIT_HASH: &str = "01234569afbe457afa1d2683a099c7af48a523c1";
 
 pub fn set_code_version() {
     if env::var_os("CODE_VERSION").is_none() {
-        let output = Command::new("git").args(["rev-parse", "HEAD"]).output();
+        let output = match Command::new("git")
+            .args(["rev-parse", "HEAD"])
+            .stdout(Stdio::piped())
+            .stderr(Stdio::inherit())
+            .output()
+        {
+            Ok(o) => o,
+            Err(e) => match env::var("FEDIMINT_BUILD_ALLOW_GIT_FAIL") {
+                Ok(_) => {
+                    eprintln!("Failed to execute git command.");
+                    println!("cargo:rustc-env=CODE_VERSION={FAKE_COMMIT_HASH}");
+                    return;
+                }
+                Err(_) => {
+                    panic!("Failed to execute git command: {e}");
+                }
+            },
+        };
 
-        match output {
-            Ok(output) => {
-                if output.status.success() {
-                    let git_hash = String::from_utf8_lossy(&output.stdout).trim().to_string();
-                    println!("cargo:rustc-env=CODE_VERSION={}", git_hash);
-                } else {
-                    let error_message = String::from_utf8_lossy(&output.stderr);
-                    io::stderr().write_all(error_message.as_bytes()).unwrap();
-                    println!("cargo:rustc-env=CODE_VERSION={}", FAKE_COMMIT_HASH);
+        let git_hash = if output.status.success() {
+            match String::from_utf8(output.stdout) {
+                Ok(hash) => hash.trim().to_string(),
+                Err(_) => {
+                    eprintln!("Invalid UTF-8 sequence detected.");
+                    FAKE_COMMIT_HASH.to_string()
                 }
             }
-            Err(error) => {
-                io::stderr()
-                    .write_fmt(format_args!("Failed to execute git command: {}", error))
-                    .unwrap();
-                println!("cargo:rustc-env=CODE_VERSION={}", FAKE_COMMIT_HASH);
-            }
-        }
+        } else {
+            FAKE_COMMIT_HASH.to_string()
+        };
+
+        println!("cargo:rustc-env=CODE_VERSION={git_hash}");
     }
     println!("cargo:rerun-if-env-changed=CODE_VERSION");
 }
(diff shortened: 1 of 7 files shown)


fedimint-build/src/lib.rs
```diff
@@ -1,21 +1,31 @@
...
+        let output = Command::new("git").args(["rev-parse", "HEAD"]).output();
+
+        match output {
+            Ok(output) => {
+                if output.status.success() {
+                    let git_hash = String::from_utf8_lossy(&output.stdout).trim().to_string();
+                    println!("cargo:rustc-env=CODE_VERSION={}", git_hash);
+                } else {
+                    let error_message = String::from_utf8_lossy(&output.stderr);
+                    io::stderr().write_all(error_message.as_bytes()).unwrap();
```
Maan2003, line 18: I would use Command::spawn with stderr set to inherit and stdout set to pipe

fedimint-build/src/lib.rs
```diff
@@ -1,21 +1,31 @@
...
-            // by giving it a prefix (0x0123456)
-            String::from("01234569afbe457afa1d2683a099c7af48a523c1")
-        };
-        println!("cargo:rustc-env=CODE_VERSION={git_hash}");
+        let output = Command::new("git").args(["rev-parse", "HEAD"]).output();
+
+        match output {
+            Ok(output) => {
+                if output.status.success() {
+                    let git_hash = String::from_utf8_lossy(&output.stdout).trim().to_string();
```
Maan2003, line 14: String::from_utf8 not lossy

fedimint-build/src/lib.rs
```diff
@@ -1,21 +1,31 @@
...
+                    println!("cargo:rustc-env=CODE_VERSION={}", git_hash);
+                } else {
+                    let error_message = String::from_utf8_lossy(&output.stderr);
+                    io::stderr().write_all(error_message.as_bytes()).unwrap();
+                    println!("cargo:rustc-env=CODE_VERSION={}", FAKE_COMMIT_HASH);
+                }
+            }
+            Err(error) => {
+                io::stderr()
```
Maan2003: use eprintln!

nassersaazi:
would love to test this! any leads @vnprc ?

fedimint-build/src/lib.rs
vnprc: I'm not sure what value command::spawn adds here since you immediately need to wait for the child process to resolve. I was unable to find a construction I liked better than command::new. Happy to revisit if you disagree.

vnprc:
@nassersaazi Thank you! I will gladly accept your offer. =D

There are a lot of error conditions in this code. I'm not sure we're handling them in the best way but I tried not to alter the behavior of `set_code_version()` in order to contain the scope of this PR.

Here's a list of (perhaps insignificant) error conditions:
- git command fails: should print "Failed to execute git command." to stderr
- git command returns an error code: the git command will print to stderr outside of the scope of this program (is this the behavior we want?)
- git command returns a non-UTF-8 string: should print "Invalid UTF-8 sequence detected." to stderr

In all three cases the `CODE_VERSION` env var will get set to the fake commit hash value.

The success case is easy to test: run `devimint version-hash-tests` and it should print the following output:
```
$ devimint version-hash-tests
git hash:                       37b5ede0b6b27cda2ac67644ed0a7f90cd3f1aa7
fedimint-cli hash:              37b5ede0b6b27cda2ac67644ed0a7f90cd3f1aa7
fedimintd hash:                 37b5ede0b6b27cda2ac67644ed0a7f90cd3f1aa7
gateway-cli hash:               37b5ede0b6b27cda2ac67644ed0a7f90cd3f1aa7
gateway-cln-extension hash:     37b5ede0b6b27cda2ac67644ed0a7f90cd3f1aa7
gatewayd hash:                  37b5ede0b6b27cda2ac67644ed0a7f90cd3f1aa7
```

fedimint-build/src/lib.rs
```diff
@@ -1,20 +1,36 @@
...
-                .unwrap();
-            String::from_utf8(output.stdout).unwrap()
+        let output = match Command::new("git")
+            .args(["rev-parse", "HEAD"])
+            .stdout(Stdio::piped())
+            .stderr(Stdio::inherit())
+            .output()
+        {
+            Ok(o) => o,
+            Err(_) => {
```
dpc, line 15: IMO this branch should be forbidden (fail the build), unless certain environment variable is set to allow it. We would set it in the CI that uses Nix and expect it, but otherwise we always expect the `git rev-parse HEAD` to just work. We might reconsider in the future when we start publishing these crates.

fedimint-build/src/lib.rs
```diff
@@ -1,20 +1,41 @@
...
-                .unwrap();
-            String::from_utf8(output.stdout).unwrap()
+        let output = match Command::new("git")
+            .args(["rev-parse", "HEAD"])
+            .stdout(Stdio::piped())
+            .stderr(Stdio::inherit())
+            .output()
+        {
+            Ok(o) => o,
+            Err(e) => match env::var("ALLOW_GIT_FAIL") {
```
Maan2003, line 15: No, why can't the builder set CODE_VERSION manually instead of setting this variable

dpc: In CI Nix will actually not use it, because we don't want to rebuild everything from scratch every time, even if nothing changed, just because git hash is different.

But yeah. `FEDIMINT_VERSION_HASH=...` should make the build not run `git ...` at all, and just take and use it. And if it wasn't set that means we *require* `git` to succeed.

elsirion:
Trying to trigger CI

vnprc:
Yeah, sorry. I broke the CI build somehow and had to stop soon after. I need to figure out how to test this locally to get the new env var working correctly. Then I will clean up my commit history. Hopefully, then it will be ready to merge.

nix/craneCommon.nix
```diff
@@ -72,6 +72,7 @@ craneLib.overrideScope' (self: prev: {
     PROTOC = "${pkgs.protobuf}/bin/protoc";
     PROTOC_INCLUDE = "${pkgs.protobuf}/include";
     CARGO_PROFILE = self.commonProfile;
+    ALLOW_GIT_FAIL = "true";
```
dpc, line 75: A convention is to prefix your env variable name with a name of the project they belong to. `CARGO_`, `PROTOC_`, etc. For us it would be `FEDIMINT_`. Maybe `FEDIMINT_BUILD_` even?

elsirion:
Looks good, but do we actually run the test yet? Can't find the change in this PR.

vnprc:
lol oops

nix/craneCommon.nix
```diff
@@ -72,6 +72,7 @@ craneLib.overrideScope' (self: prev: {
     PROTOC = "${pkgs.protobuf}/bin/protoc";
     PROTOC_INCLUDE = "${pkgs.protobuf}/include";
     CARGO_PROFILE = self.commonProfile;
+    FEDIMINT_BUILD_ALLOW_GIT_FAIL = "true";
```
dpc, line 75: I think we will need to refactor it and move it to a new nix variable, because this one gets sourced into our dev shell:

```
> nix develop
> echo $FEDIMINT_BUILD_ALLOW_GIT_FAIL 
true
```

which is probably not what we want.

But let's land and we can fix later.

dpc (dismissed):

nix/craneCommon.nix
vnprc: It still fails in CI. I think we need to install git in the CI environment to get it to pass.

scripts/test-ci-all.sh
```diff
@@ -27,6 +27,16 @@ cargo build ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} --all --all-targets
 >&2 echo "Pre-building tests..."
 cargo test --no-run ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} -p fedimint-tests
 
+function cli_test_bin_version_hash() {
+  set -eo pipefail # pipefail must be set manually again
+  trap 'echo "## FAILED: ${FUNCNAME[0]}"' ERR
+
+  echo "## START: ${FUNCNAME[0]}"
+  unshare -rn bash -c "ip link set lo up && exec unshare --user ./scripts/version-hash-tests.sh" 2>&1 | ts -s
```
dpc, line 35: Ah, right. This runs in a nix environment which does not have `git` or `.git` available, that's why we set this env var to allow the failure. I don't think there's anything we can do about it, LOL. I mean - we could, but that would require rebuilding everything all the time, which just doesn't seem worth it.



We could skip this test when running in CI, by detecting `FEDIMINT_BUILD_...` var I guess, then people would be testing it when running it locally...

But I'm not sure if it's worth it. All in all - it's not an end of the world.

In CI we could run `nix run .#fedimint-cli version-hash` and verify that it was set etc.  but that tests something completely different (nix substituting version, no the build scripts).

vnprc:
I also agree that it would be good to merge what works in the dev environment and continue working on getting it to pass in CI. But I'd like to rebase out some of the (failed) experimental commits first. No need to pollute the git history in the master branch.

scripts/test-ci-all.sh
vnprc: > that would require rebuilding everything all the time
I don't follow your logic here. The CI build has to check out the code, right? Can we not grab the git hash at that point and pass it through the system somehow? If the git hash is not passed in when the test is executed, fall back to calling git directly, which should take care of the dev env use case.

dpc:
BTW. https://github.com/fedimint/fedimint/pull/2704

vnprc:
@dpc I won't be able to get back to this work until late next week. Do you think it would be good to merge this as it is and create a new issue to incorporate this test into the CI build?

elsirion:
What's the CI issue about? With failing CI we can't merge and I restarted a few times already, so it's unlikely to be a flaky test.

dpc:
Moved to https://github.com/fedimint/fedimint/pull/2726 . Dropped the tests, as we test in the CI now the release binaries (built by nix).

PR #2686, opened 2023-06-23
refactor: make a primary module a runtime concept, not type system one

dpc:
The fact that primary modules implement a separate trait generate a lot of friction, for rather small gains.

In particular the fact that primary module is not present in the client's `modules` makes functionality that need to something on "all module instances" error prone, and require extra ceremony.

image

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -53,8 +53,8 @@
 //! module and its smart contracts are only used to incentivize LN payments, not
 //! to hold money. The mint module on the other hand holds e-cash note and can
 //! thus be used to fund transactions and to absorb change. Module clients with
-//! this ability should implement the [`module::PrimaryClientModule`] trait so
-//! they can be chosen as the primary module for a [`Client`].
+//! this ability should implement [`ClientModule::supports_being_primary`] and
+//! related methods.
 //!
 //! For a example of a client module see [the mint client](https://github.com/fedimint/fedimint/blob/master/modules/fedimint-mint-client/src/lib.rs).
 //!
@@ -91,6 +91,7 @@ use fedimint_core::{
 pub use fedimint_derive_secret as derivable_secret;
 use fedimint_derive_secret::DerivableSecret;
 use futures::StreamExt;
+use module::DynClientModule;
 use rand::thread_rng;
 use secp256k1_zkp::{PublicKey, Secp256k1};
 use secret::DeriveableSecretClientExt;
@@ -102,9 +103,7 @@ use crate::db::ClientSecretKey;
 use crate::module::gen::{
     ClientModuleGen, ClientModuleGenRegistry, DynClientModuleGen, IClientModuleGen,
 };
-use crate::module::{
-    ClientModule, ClientModuleRegistry, DynPrimaryClientModule, IClientModule, StateGenerator,
-};
+use crate::module::{ClientModule, ClientModuleRegistry, IClientModule, StateGenerator};
 use crate::oplog::OperationLog;
 use crate::secret::RootSecretStrategy;
 use crate::sm::executor::{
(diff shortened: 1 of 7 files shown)


fedimint-client/src/lib.rs
```diff
@@ -667,10 +666,17 @@ impl Client {
         &self.inner.config
     }
 
+    /// Get the primary module
+    pub fn primary_module(&self) -> &DynClientModule {
+        self.inner
+            .modules
+            .get(self.inner.primary_module_instance)
+            .expect("primary module must be prevent")
```
douglaz, line 674:
```suggestion
            .expect("primary module must be present")
```

douglaz (approved):
LGTM

dpc:
Needs an ack from @elsirion

fedimint-client/src/module/mod.rs
```diff
@@ -102,6 +102,87 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    /// * [`Self::create_sufficient_input`]
+    /// * [`Self::create_exact_output`]
+    /// * [`Self::await_primary_module_output`]
+    /// * [`Self::get_balance`]
+    /// * [`Self::subscribe_balance_changes`]
+    fn supports_being_primary(&self) -> bool {
+        false
+    }
+
+    // TODO: unclear if we should return a vec of inputs
```
elsirion, line 119: I discussed something related with @joschisan: if we ever want to make e-cash notes separate inputs/outputs and remove `TieredMulti` that would be useful.

fedimint-client/src/module/mod.rs
```diff
@@ -102,6 +102,87 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    /// * A set of private keys belonging to the input for signing the
+    ///   transaction
+    /// * The input of **at least** `min_amount`, the actual amount might be
+    ///   larger, the caller has to handle this case and possibly generate
+    ///   change using `create_change_output`.
+    /// * A closure that generates states belonging to the input. This closure
+    ///   takes the transaction id of the transaction in which the input was
+    ///   used and the input index as input since these cannot be known at time
+    ///   of calling `create_funding_input` and have to be injected later.
+    ///
```
elsirion, line 133: nit unrelated to this PR: We might want to move that documentation to `ClientInput`. Originally these items were returned as a tuple, that's why the documentation is still here.

fedimint-client/src/module/mod.rs
```diff
@@ -102,6 +102,87 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+
+    /// Creates an output of **exactly** `amount` that will pay into the
+    /// holdings managed by the module.
+    ///
+    /// It returns:
+    /// * The output of **exactly** `amount`.
+    /// * A closure that generates states belonging to the output. This closure
+    ///   takes the transaction id of the transaction in which the output was
+    ///   used and the output index as input since these cannot be known at time
+    ///   of calling `create_change_output` and have to be injected later.
```
elsirion, line 153: Same here

fedimint-client/src/lib.rs
```diff
@@ -711,7 +717,6 @@ struct ClientInner {
     db: Database,
     federation_id: FederationId,
     federation_meta: BTreeMap<String, String>,
-    primary_module: DynPrimaryClientModule,
     primary_module_instance: ModuleInstanceId,
     primary_module_kind: ModuleKind,
```
elsirion, line 721: The primary module kind isn't needed anymore. The only reason we kept it around was because the kind wasn't saved in the module registry like for all other modules:

https://github.com/fedimint/fedimint/blob/cc179f060e1029139a1eab69f98033be7189b14e/fedimint-client/src/lib.rs#L631-L641

elsirion (commented):
I think it would be good to remove the `primary_module_kind`. It could be a follow-up, but imo that change would fit in here perfectly fine.

douglaz (approved):

PR #2688, opened 2023-06-23
feature: real LN tests

### LINKED ISSUE #2650: Attach Modules using builder pattern in `gatewayd`
`gatewayd` currently attaches the modules when each client is built, which results in us needing to attach all modules (including the dummy module) even when some modules are not needed. We should switch `gatewayd` to use a builder pattern, similar to fedimintd, so that we can attach exactly the modules that are needed.

jkitman:
Runs our LN tests with the real nodes.

Refactors the LN tests to leverage the gateways rather than construct clients directly.  This will allow us to test gateway-federation interactions better.

Subtask  of #2315.
Fixes #2650, moving client registration into the `bin`.


gateway/ln-gateway/src/client.rs
```diff
@@ -60,7 +53,8 @@ impl StandardGatewayClientBuilder {
 
         let mut client_builder = ClientBuilder::default();
         client_builder.with_module_gens(registry);
-        client_builder.with_primary_module(1);
```
jkitman, line 63: @m1sterc001guy the original code seems to set the dummy module as the primary module, no?

gateway/ln-gateway/src/ng/tests.rs
```diff
@@ -124,7 +74,8 @@ pub fn sha256(data: &[u8]) -> sha256::Hash {
 #[tokio::test(flavor = "multi_thread")]
 async fn test_gateway_client_pay_valid_invoice() -> anyhow::Result<()> {
     gateway_test(
-        |_, other_lightning_client, _, user_client, gateway| async move {
+        |gateway, other_lightning_client, fed, user_client| async move {
+            let gateway = gateway.remove_client(&fed).await;
```
jkitman, line 78: @m1sterc001guy we remove the client from the gateway so you can manually manipulate the gateway.  However, we could also not do this and let this be more of an E2E test that will test the gateway RPC logic.

scripts/rust-tests.sh
```diff
@@ -32,6 +32,8 @@ export FM_TEST_USE_REAL_DAEMONS=1
 
 if [ -z "${FM_TEST_ONLY:-}" ] || [ "${FM_TEST_ONLY:-}" = "bitcoind" ]; then
   >&2 echo "### Testing against bitcoind"
+  env RUST_BACKTRACE=1 cargo test -p ln-gateway -- --test-threads=$(($(nproc) * 2)) "$@"
+  env RUST_BACKTRACE=1 cargo test -p fedimint-ln-tests can_switch_active_gateway -- --test-threads=$(($(nproc) * 2)) "$@"
```
jkitman, line 36: Adds all of our LN tests to run with real daemons.

fedimint-testing/src/fixtures.rs
```diff
@@ -134,29 +126,44 @@ impl Fixtures {
         .await
     }
 
-    /// Starts a new gateway
-    pub async fn new_gateway(&self, password: Option<String>) -> GatewayTest {
+    /// Starts a new gateway with a given lightning node
+    pub async fn new_gateway(&self, ln: Arc<dyn LightningTest>) -> GatewayTest {
         // TODO: Make construction easier
```
m1sterc001guy, line 131: This probably isn't a TODO anymore right? Not sure exactly what its referring to now

fedimint-testing/src/gateway.rs
```diff
@@ -29,56 +37,58 @@ impl GatewayTest {
...
-        fed.last().expect("No feds registered").registration.clone()
+            .unwrap()
     }
 
     pub(crate) async fn new(
         base_port: u16,
         password: String,
-        lightning: impl ILnRpcClient + 'static,
+        lightning: Arc<dyn LightningTest>,
         decoders: ModuleDecoderRegistry,
```
m1sterc001guy, line 58: Can we get the decoders from the registry now that we're passing that in?

gateway/ln-gateway/src/bin/gatewayd.rs
```diff
@@ -84,7 +85,14 @@ async fn main() -> Result<(), anyhow::Error> {
     );
 
     // Create federation client builder
-    let client_builder = StandardGatewayClientBuilder::new(data_dir.clone());
+    let mut registry = ClientModuleGenRegistry::new();
+    registry.attach(MintClientGen);
+    registry.attach(WalletClientGen::default());
```
m1sterc001guy, line 90: Step closer to this https://github.com/fedimint/fedimint/issues/2650

gateway/ln-gateway/src/ng/tests.rs
```diff
@@ -26,93 +25,44 @@ use fedimint_ln_common::contracts::{EncryptedPreimage, FundedContract, Preimage}
...
-        },
-        timelock_delta: 10,
-        mint_channel_id: 1,
-    });
-    let gateway = fed.new_gateway_client(registry).await;
-    {
-        let fake_api = Url::from_str("http://127.0.0.1:8175").unwrap();
-        let fake_route_hints = Vec::new();
-        let register_op = gateway
-            .register_with_federation(fake_api, fake_route_hints, GW_ANNOUNCEMENT_TTL)
```
m1sterc001guy, line 70: When does registration happen now?

m1sterc001guy: Found it, nevermind :)

gateway/ln-gateway/src/ng/tests.rs
m1sterc001guy: ah ok I think I understand. So `GatewayTest` here will do actual RPCs to the webserver. I think that works, maybe we could just rename `gateway` here to be `gateway_client` to make that more clear

gateway/ln-gateway/tests/fixtures.rs
```diff
@@ -1,39 +1,35 @@
...
-    let mut fixtures =
-        Fixtures::new_primary(1, DummyClientGen, DummyGen, DummyGenParams::default());
+    let mut fixtures = Fixtures::new_primary(DummyClientGen, DummyGen, DummyGenParams::default());
     let ln_params = LightningGenParams::regtest(fixtures.bitcoin_server());
-    fixtures = fixtures.with_module(0, LightningClientGen, LightningGen, ln_params);
+    fixtures = fixtures.with_module(LightningClientGen, LightningGen, ln_params);
 
-    let gateway = fixtures.new_gateway(password).await;
+    let lnd = fixtures.lnd().await;
+    let gateway = fixtures.new_gateway(lnd).await;
```
m1sterc001guy, line 28: I forget when this is used. Is `lnd` always the gateway when this is called?

m1sterc001guy (approved):

gateway/ln-gateway/tests/fixtures.rs
jkitman: So `lnd()` and `cln()` will give you the two kinds of test LN nodes

You can then pass them into as many gateways as you wish.  Then connect gateways to as many feds as you wish.

fedimint-testing/src/gateway.rs
justinmoon: Seems like a good idea

PR #2690, opened 2023-06-24
fix: update some commands on docs

douglaz:
There are no default notes on the old client, making the current guide very confusing. Also the gateway commands changed.


nassersaazi:
this is really helpful!

dpc (approved):

Issue #2649, opened 2023-06-14
Process HTLCs in parallel


m1sterc001guy:
Today, the `route_htlcs` loop waits for a response to send back to the lightning node (`Forward`, `Settle`, or `Cancel`) before processing the next HTLC. If we can move the sending of the response into the state machine, this would allow the `route_htlcs` thread to process HTLCs in parallel without waiting for the previous HTLC to finish. 

This would also help `gatewayd` survive restarts if it crashes before it sends the response back to the lightning node.

PR #2692, opened 2023-06-27
feat: parallelize htlc processing

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2649

Today, HTLC processing in `gatewayd` in sequential - `gatewayd` will never process another HTLC until it has sent the response to the previous one. This hurts parallelism if there are a lot of payments flowing through the lightning node.

This PR parallelizes HTLC handling. Now, when an HTLC is intercepted, `gatewayd` will spawn two state machines - one to handle the incoming contract and another to complete the HTLC. The `complete` state machine will wait on the `preimage` from the `incoming` state machine, since it needs that to complete the payment. This allows the incoming state machine to be separate from the completion of the HTLC, since that workflow is also used for internal payments.

This is also a reliability fix - if `gatewayd` crashes before it has a chance to complete the HTLC, it can now recover since the completion of the HTLC is now part of the state machine workflow.

Update: This PR also changes the lifetime of the webserver and how/when the `Gateway` is created. This is because when the connection to the node dies, the existing clients will not have the bi-directional stream setup. To solve this, we re-create the gateway and the webserver if the connection to the lightning node breaks.

### DIFF
--- a/gateway/ln-gateway/Cargo.toml
+++ b/gateway/ln-gateway/Cargo.toml
@@ -54,7 +54,6 @@ fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
 fedimint-dummy-client = { path = "../../modules/fedimint-dummy-client" }
 fedimint-wallet-client = { path = "../../modules/fedimint-wallet-client" }
 futures = "0.3.24"
-lazy_static = "1.4.0"
 lightning = "0.0.113"
 lightning-invoice = "0.21.0"
 prost = "0.11"
--- a/gateway/ln-gateway/Cargo.toml
+++ b/gateway/ln-gateway/Cargo.toml
@@ -54,6 +54,7 @@ fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
 fedimint-dummy-client = { path = "../../modules/fedimint-dummy-client" }
 fedimint-wallet-client = { path = "../../modules/fedimint-wallet-client" }
 futures = "0.3.24"
+lazy_static = "1.4.0"
 lightning = "0.0.113"
 lightning-invoice = "0.21.0"
 prost = "0.11"
--- a/gateway/ln-gateway/proto/gatewaylnrpc.proto
+++ b/gateway/ln-gateway/proto/gatewaylnrpc.proto
@@ -25,10 +25,14 @@ service GatewayLightning {
    * to response with `InterceptHtlcResponse` after the HTLC has been processed with
    * the appropriate action (Settle, Forward, Cancel).
    */
-  rpc RouteHtlcs(stream InterceptHtlcResponse) returns (stream InterceptHtlcRequest) {}
+  rpc RouteHtlcs(EmptyRequest) returns (stream InterceptHtlcRequest) {}
+
+  rpc CompleteHtlc(InterceptHtlcResponse) returns (EmptyResponse) {}
 }
 
 message EmptyRequest {}
+
+message EmptyResponse {}
 
 message GetNodeInfoResponse {
   // The public key of the associated lightning node
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -14,14 +14,13 @@ use cln_rpc::model;
 use cln_rpc::primitives::ShortChannelId;
 use fedimint_core::task::TaskGroup;
 use fedimint_core::Amount;
-use futures::stream::StreamExt;
 use ln_gateway::gatewaylnrpc::gateway_lightning_server::{
     GatewayLightning, GatewayLightningServer,
 };
 use ln_gateway::gatewaylnrpc::get_route_hints_response::{RouteHint, RouteHintHop};
 use ln_gateway::gatewaylnrpc::intercept_htlc_response::{Action, Cancel, Forward, Settle};
 use ln_gateway::gatewaylnrpc::{
-    EmptyRequest, GetNodeInfoResponse, GetRouteHintsResponse, InterceptHtlcRequest,
+    EmptyRequest, EmptyResponse, GetNodeInfoResponse, GetRouteHintsResponse, InterceptHtlcRequest,
     InterceptHtlcResponse, PayInvoiceRequest, PayInvoiceResponse,
 };
 use secp256k1::PublicKey;
@@ -477,10 +476,8 @@ impl GatewayLightning for ClnRpcService {
 
     async fn route_htlcs(
         &self,
-        request: tonic::Request<tonic::Streaming<InterceptHtlcResponse>>,
+        _: tonic::Request<EmptyRequest>,
     ) -> Result<tonic::Response<Self::RouteHtlcsStream>, Status> {
-        let mut stream = request.into_inner();
-
         // First create new channel that we will use to send responses back to gatewayd
         let (gatewayd_sender, gatewayd_receiver) =
             mpsc::channel::<Result<InterceptHtlcRequest, Status>>(100);
@@ -488,21 +485,16 @@ impl GatewayLightning for ClnRpcService {
         let mut sender = self.interceptor.sender.lock().await;
         *sender = Some(gatewayd_sender.clone());
 
-        // Spawn new thread that listens for events from the input stream
-        let interceptors = self.interceptor.clone();
-        tokio::spawn(async move {
-            while let Some(res) = stream.next().await {
-                if let Ok(complete_request) = res {
-                    let _ = Self::complete_htlc(complete_request, interceptors.clone())
-                        .await
-                        .map_err(|e| {
-                            error!("CLN extension failed to complete HTLC: {:?}", e);
-                        });
-                }
-            }
-        });
-
         Ok(tonic::Response::new(ReceiverStream::new(gatewayd_receiver)))
+    }
+
+    async fn complete_htlc(
+        &self,
+        intercept_response: tonic::Request<InterceptHtlcResponse>,
+    ) -> Result<tonic::Response<EmptyResponse>, Status> {
+        let complete_htlc = intercept_response.into_inner();
+        Self::complete_htlc(complete_htlc, self.interceptor.clone()).await?;
+        Ok(tonic::Response::new(EmptyResponse {}))
     }
 }
 
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -235,79 +235,6 @@ impl ClnRpcService {
             })
             .map_err(ClnExtensionError::RpcError)?
     }
-
-    async fn complete_htlc(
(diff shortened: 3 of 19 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -248,22 +248,19 @@ impl Gateway {
                             break;
                         }
 
-                        // Create a stream used to communicate with the Lightning implementation
-                        let (sender, ln_receiver) = mpsc::channel::<InterceptHtlcResponse>(100);
-
                         if let Some(ln_mode) = ln_mode.clone() {
-                            let mut lnrpc = Self::create_boxxed_lightning_client(ln_mode).await;
+                            let mut lnrpc_route = Self::create_boxxed_lightning_client(ln_mode).await;
```
douglaz, line 252: I know it's unrelated but could fix typo:
```suggestion
                            let mut lnrpc_route = Self::create_boxed_lightning_client(ln_mode).await;
```

gateway/ln-gateway/Cargo.toml
```diff
@@ -54,6 +54,7 @@ fedimint-mint-client = { path = "../../modules/fedimint-mint-client" }
 fedimint-dummy-client = { path = "../../modules/fedimint-dummy-client" }
 fedimint-wallet-client = { path = "../../modules/fedimint-wallet-client" }
 futures = "0.3.24"
+lazy_static = "1.4.0"
```
dpc, line 57: Not enthusiastic about it.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -476,33 +475,26 @@ impl GatewayLightning for ClnRpcService {
...
-
         // First create new channel that we will use to send responses back to gatewayd
         let (gatewayd_sender, gatewayd_receiver) =
             mpsc::channel::<Result<InterceptHtlcRequest, Status>>(100);
 
         let mut sender = self.interceptor.sender.lock().await;
         *sender = Some(gatewayd_sender.clone());
 
-        // Spawn new thread that listens for events from the input stream
-        let interceptors = self.interceptor.clone();
```
dpc, line 491: I'm super confused why you can just delete it here, and leave this function as is. :D

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -476,33 +475,26 @@ impl GatewayLightning for ClnRpcService {
...
-                            error!("CLN extension failed to complete HTLC: {:?}", e);
-                        });
-                }
-            }
-        });
-
         Ok(tonic::Response::new(ReceiverStream::new(gatewayd_receiver)))
     }
+
+    async fn complete_htlc(
```
dpc, line 490: The code you deleted was already calling `Self::complete_htlc`, so ... how can you add this method here without conflicting with an existing one ... . I don't see deleting an existing one ...

I'm so confused.

douglaz: Here we are implementing `GatewayLightning for ClnRpcService`. The existing method is directly on `ClnRpcService`.

Rust traits are fun :smile:

gateway/ln-gateway/src/lnd.rs
```diff
@@ -18,14 +18,18 @@ use tracing::{error, info, trace};
...
-    PayInvoiceRequest, PayInvoiceResponse,
+    EmptyResponse, GetNodeInfoResponse, GetRouteHintsResponse, InterceptHtlcRequest,
+    InterceptHtlcResponse, PayInvoiceRequest, PayInvoiceResponse,
 };
 use crate::lnrpc_client::{ILnRpcClient, RouteHtlcStream, MAX_LIGHTNING_RETRIES};
 use crate::GatewayError;
 
 type HtlcSubscriptionSender = mpsc::Sender<Result<InterceptHtlcRequest, Status>>;
 
+lazy_static::lazy_static! {
```
dpc, line 29: Any reason why this needs to be global? Couldn't we embedded it into some struct?

gateway/ln-gateway/Cargo.toml
elsirion: We already depend on it transitively in a bunch of places and it's a well-known crate, so not too big of a deal imo.

gateway/ln-gateway/proto/gatewaylnrpc.proto
```diff
@@ -25,11 +25,15 @@ service GatewayLightning {
    * to response with `InterceptHtlcResponse` after the HTLC has been processed with
    * the appropriate action (Settle, Forward, Cancel).
    */
-  rpc RouteHtlcs(stream InterceptHtlcResponse) returns (stream InterceptHtlcRequest) {}
+  rpc RouteHtlcs(EmptyRequest) returns (stream InterceptHtlcRequest) {}
+
+  rpc CompleteHtlc(InterceptHtlcResponse) returns (EmptyResponse) {}
```
elsirion, line 30: This seems like a much better API design to me.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -476,33 +475,26 @@ impl GatewayLightning for ClnRpcService {
...
+        &self,
+        intercept_response: tonic::Request<InterceptHtlcResponse>,
+    ) -> Result<tonic::Response<EmptyResponse>, Status> {
+        let complete_htlc = intercept_response.into_inner();
+        Self::complete_htlc(complete_htlc, self.interceptor.clone()).await?;
```
elsirion, line 495: Not sure if `<Self as …>` notation would work here, given the overloading it would make things easier to read imo. I kinda expected this to be recursive on first glance.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -476,33 +475,26 @@ impl GatewayLightning for ClnRpcService {
...
+        _: tonic::Request<EmptyRequest>,
     ) -> Result<tonic::Response<Self::RouteHtlcsStream>, Status> {
-        let mut stream = request.into_inner();
-
         // First create new channel that we will use to send responses back to gatewayd
         let (gatewayd_sender, gatewayd_receiver) =
             mpsc::channel::<Result<InterceptHtlcRequest, Status>>(100);
 
         let mut sender = self.interceptor.sender.lock().await;
         *sender = Some(gatewayd_sender.clone());
```
elsirion, line 485: Logging if a sender is replaced might be nice for debugging (should only happen on GW restart afaik).

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -476,33 +475,26 @@ impl GatewayLightning for ClnRpcService {
...
-            while let Some(res) = stream.next().await {
-                if let Ok(complete_request) = res {
-                    let _ = Self::complete_htlc(complete_request, interceptors.clone())
-                        .await
-                        .map_err(|e| {
-                            error!("CLN extension failed to complete HTLC: {:?}", e);
-                        });
-                }
-            }
-        });
```
elsirion, line 502: I assume that deleted code was concerned with processing the HTLC completion messages streamed to us by the client?

gateway/ln-gateway/src/lnd.rs
elsirion: Same question on my side, mutable globals are a big code smell. Given its usage I'd put it into `GatewayLndClient`.

elsirion (changes requested):
I've been out of the loop with the LN GW for a bit, but looks reasonable to me. The only thing that should be changed imo is the reliance on a mutable global variable.

gateway/ln-gateway/proto/gatewaylnrpc.proto
m1sterc001guy: We actually originally had it like this, but then thought streaming would simplify the code a bit. Learned in the process that there's some issues doing that :)

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: Yeah perhaps I should just move the existing `complete_htlc` functionality into here to eliminate the confusion

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: Will refactor to make it more readable

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: Open to suggestions on how to solve this, I'm not a fan of having this be global either.

The reason it cannot be embedded in `GatewayLndClient` right now is because we use multiple lightning clients in different spots. We use one lightning client for handling HTLC interceptions and we use a different lightning client when "completing" those HTLCs.

We could refactor so that each of these operations uses the same client, but that essentially restricts us to a single `GatewayLndClient` that can be used, which functionally is similar to a global variable. Perhaps its cleaner code though.

m1sterc001guy: Actually, using `Settleinvoice` and `CancelInvoice` will probably solve this problem

okjodom: @m1sterc001guy, check out #2755 which avails these `invoicesrpc` features you need

gateway/ln-gateway/Cargo.toml
m1sterc001guy: Removed

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: This has been removed.

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: This code was processing a stream of `InterceptHtlcResponse` from gatewayd. This PR changes the way gatewayd sends responses back, it now uses a separate RPC (`complete_htlc`). So this code is not necessary anymore.

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: Yep, exactly. A little more context here: https://github.com/fedimint/fedimint/pull/2692/files#r1262890249

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: Moved this logic so that there's only one `complete_htlc`

m1sterc001guy:
> This will allow us to use SettleInvoice and CancelInvoice which allows us to remove the ugly global variable that is currently required for LND.

I misinterpreted the LND docs, the approach above will not work. I refactored the logic inside the Gateway though to eliminate the use of the ugly global variable, so this is no longer a problem.

PR #2693, opened 2023-06-27
refactor: use gateway public key instead of node public key

m1sterc001guy:
`switch_gateway` currently uses `node_pub_key` as the key for determining which gateway the user desires. This is broken because it is possible multiple gateways could use the same lightning node.

We have `gateway_pub_key` which the gateway uses the redeem contracts from the federations its connected to. This PR changes `switch_gateway` to use that key for switching gateways.

### DIFF
--- a/fedimint-load-test-tool/src/main.rs
+++ b/fedimint-load-test-tool/src/main.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use std::vec;
 
-use anyhow::{anyhow, bail, Context};
+use anyhow::{bail, Context};
 use clap::{Args, Parser, Subcommand, ValueEnum};
 use common::{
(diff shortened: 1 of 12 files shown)


fedimint-load-test-tool/src/main.rs
```diff
@@ -774,11 +774,11 @@ async fn get_gateway_public_key(
             cmd!(GatewayClnCli, "info").out_json().await
         }
     }?;
-    let node_pub_key = gateway_json["federations"][0]["registration"]["node_pub_key"]
+    let gateway_pub_key = gateway_json["federations"][0]["registration"]["gateway_pub_key"]
         .as_str()
         .ok_or_else(|| anyhow!("Missing node_pub_key field"))?;
```
douglaz, line 779:
```suggestion
        .context("Missing gateway_pub_key field")?;
```
And change the import to `use anyhow::{bail, Context};`

douglaz (approved):
LGTM

Issue #2674, opened 2023-06-20
LND Gateway error: payment is in transition


douglaz:
Generating invoices with CLN and paying with LND using two or more concurrent users, like:
```bash
fedimint-load-test-tool --users 2 load-test --generate-invoice-with cln-lightning-cli 
```
Is generating an error on the gateway-lnd:
```log
2023-06-20T00:42:49.993522Z  INFO ln_gateway::lnd: send response SendResponse { payment_error: "payment is in transition", payment_preimage: [], payment_route: None, payment_hash: [155, 97, 160, 66, 247, 123, 110, 44, 161, 58, 215, 43, 140, 200, 231, 215, 242, 61, 170, 76,  162, 40, 64, 23, 86, 206, 137, 225, 38, 175, 128, 11] }                                                                                 
2023-06-20T00:42:50.250968Z ERROR pay_invoice: ln_gateway::rpc::rpc_server: error=Other: Outgoing contract canceled                      
                                                                                                                                         
Stack backtrace:                                                                                                                         
   0: anyhow::__private::format_err                                                                                                      
             at /home/master/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.71/src/lib.rs:669:13                             
   1: {async_fn#0}                                                                                                                       
             at ./gateway/ln-gateway/src/lib.rs:625:52                                                                                   
   2: {async_block#0}                                                                                                                    
             at ./gateway/ln-gateway/src/rpc/rpc_server.rs:127:59                                                                        
   3: {async_block#0}
             at ./gateway/ln-gateway/src/rpc/rpc_server.rs:122:1
   4: poll<ln_gateway::rpc::rpc_server::pay_invoice::{async_fn#0}::{async_block_env#0}>
             at /home/master/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-0.1.37/src/instrument.rs:272:9
   5: {async_fn#0}
             at ./gateway/ln-gateway/src/rpc/rpc_server.rs:122:1
   6: {async_block#0}<fn(axum::extension::Extension<ln_gateway::Gateway>, axum::json::Json<fedimint_ln_client::pay::PayInvoicePayload>) -
> ln_gateway::rpc::rpc_server::pay_invoice::{async_fn_env#0}, ln_gateway::rpc::rpc_server::pay_invoice::{async_fn_env#0}, (), hyper::body
[...]

Issue #2496, opened 2023-05-11
`pay_invoice` is not idempotent on the gateway side


m1sterc001guy:
All APIs that the gateway exposes should be idempotent. `pay_invoice` is not idempotent, which causes issues for the Lightning client. This required us to put the API request in the transition function in this PR https://github.com/fedimint/fedimint/pull/2486, when it should be in the trigger function.

justinmoon:
What exactly isn't idempotent?

m1sterc001guy:
When the gateway retrieves the `OutgoingContract` from the federation, if it has already been paid, it will try and retrieve the `IncomingContract` again. This contract will have a value of 0msats, which causes the gateway to error.

PR #2695, opened 2023-06-28
feat: outgoing idempotent payments

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2496

Also fixes: https://github.com/fedimint/fedimint/issues/2674

Payments are occasionally getting stuck on LND with this fix, but it does not appear to be the same issue.

Uses a map on the gateway to de-duplicate payment requests and only executes a new payment if it has not been executed before. 

On the client side, this allows us to move the `pay_invoice` request back into the trigger function, where it belongs.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -27,6 +27,7 @@ use bitcoin::{Address, Txid};
 use bitcoin_hashes::hex::ToHex;
 use clap::Subcommand;
 use client::StandardGatewayClientBuilder;
+use fedimint_client::sm::OperationId;
 use fedimint_core::api::{FederationError, WsClientConnectInfo};
 use fedimint_core::config::FederationId;
 use fedimint_core::db::Database;
@@ -36,6 +37,7 @@ use fedimint_core::util::NextOrPending;
 use fedimint_core::Amount;
 use fedimint_ln_client::contracts::Preimage;
 use fedimint_ln_client::pay::PayInvoicePayload;
+use fedimint_ln_common::contracts::ContractId;
 use fedimint_ln_common::route_hints::RouteHint;
 use fedimint_ln_common::KIND;
 use fedimint_wallet_client::{WalletClientExt, WithdrawState};
@@ -150,6 +152,7 @@ pub struct Gateway {
     gatewayd_db: Database,
     api: Url,
     task_group: TaskGroup,
+    contracts: Arc<Mutex<BTreeMap<ContractId, OperationId>>>,
 }
 
 impl Gateway {
@@ -174,6 +177,7 @@ impl Gateway {
             gatewayd_db,
             api,
             task_group: TaskGroup::new(),
+            contracts: Arc::new(Mutex::new(BTreeMap::new())),
         };
 
         gw.load_clients().await?;
@@ -201,6 +205,7 @@ impl Gateway {
             gatewayd_db,
             api,
             task_group: TaskGroup::new(),
+            contracts: Arc::new(Mutex::new(BTreeMap::new())),
         };
 
         gw.load_clients().await?;
@@ -613,14 +618,25 @@ impl Gateway {
         })
     }
 
-    async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<Preimage> {
+    async fn handle_pay_invoice_msg(&mut self, payload: PayInvoicePayload) -> Result<Preimage> {
         let PayInvoicePayload {
             federation_id,
             contract_id,
         } = payload;
 
         let client = self.select_client(federation_id).await?;
-        let operation_id = client.gateway_pay_bolt11_invoice(contract_id).await?;
+
+        let operation_id = {
+            let mut contracts_guard = self.contracts.lock().await;
+            if let Some(operation_id) = contracts_guard.get(&contract_id) {
+                operation_id.to_owned()
+            } else {
+                let operation_id = client.gateway_pay_bolt11_invoice(contract_id).await?;
+                contracts_guard.insert(contract_id, operation_id.clone());
+                operation_id
+            }
+        };
+
         let mut updates = client
             .gateway_subscribe_ln_pay(operation_id)
             .await?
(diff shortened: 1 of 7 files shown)


gateway/ln-gateway/src/lnd.rs
```diff
@@ -203,6 +205,30 @@ impl GatewayLndClient {
...
+        client: &mut LndClient,
+    ) -> Result<Option<Payment>, GatewayError> {
+        let payments = client
+            .lightning()
+            .list_payments(ListPaymentsRequest {
+                // TODO: Returning ALL payments here is bad, because there could be a lot of
+                // payments We can't filter by timestamp until we fix our tonic_lnd
+                // dependency
+                include_incomplete: true,
+                ..Default::default()
```
m1sterc001guy, line 222: The only way I could figure out to get the status of the current payment is to list all payments. This is not good if the node has done a lot of payments in the past, since it will return all of them. I'd like to filter based on the invoice's creation timestamp, but our version of tonic_lnd is not up to date enough.

It looks like we currently maintain our own fork of `tonic_lnd`. @okjodom do you remember why we are maintaining our own fork?

Regardless, it doesn't look like the upstream repo even has the most up to date proto file https://github.com/Kixunil/tonic_lnd/blob/master/vendor/lightning.proto

It seems to me that we should work to eliminate our dependency on tonic_lnd, and especially not maintain our own fork. Updating these dependencies to take advantage of the latest LND features seems really cumbersome.

m1sterc001guy: Is the main advantage of using `tonic_lnd` that it has all of the legwork for generating the rust code from the proto files?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -325,27 +351,51 @@ impl ILnRpcClient for GatewayLndClient {
...
-            return Err(GatewayError::LnRpcError(tonic::Status::new(
-                tonic::Code::Internal,
-                "LND did not return a preimage",
-            )));
+        // If the payment exists, that means we've already tried to pay the invoice
+        let preimage = if let Some(mut payment) = self
+            .lookup_payment(invoice.invoice.clone(), invoice.timestamp, &mut client)
+            .await?
+        {
+            while payment.status == 1 {
```
m1sterc001guy, line 360: This loop sometimes gets stuck for payments. LND will always return an `IN_FLIGHT` status. Haven't figured out why....

gateway/ln-gateway/src/lnd.rs
douglaz: In the past I've used this fork: https://github.com/yzernik/tonic_openssl_lnd/tree/master

It had all the features at the time.

It seems we're maintaining our own version because upstream is unwilling to upgrade stuff: https://github.com/Kixunil/tonic_lnd/pull/31

douglaz: @m1sterc001guy but why not use https://lightning.engineering/api-docs/api/lnd/lightning/lookup-invoice here?

douglaz: @m1sterc001guy I think the solution here is to use `.router().track_payment_v2` (https://lightning.engineering/api-docs/api/lnd/router/track-payment-v2)

gateway/ln-gateway/src/lnd.rs
justinmoon: Would be nice to have a descriptive variable instead of `1` ...

gateway/ln-gateway/src/lnd.rs
okjodom: > It looks like we currently maintain our own fork of tonic_lnd. @okjodom do you remember why we are maintaining our own fork?

So we can make faster updates to the proto file if we need to. Should I bump the `lightning.proto` file to match LND version we use in dev?

okjodom: Also, I added features to `tonic_lnd` allowing us to have a slim dependency rather than all pulling in rpcs we don't need. These are available in out fork but have not landed in upstream : https://github.com/Kixunil/tonic_lnd/pull/29

justinmoon:
One somewhat pathological case where it might not be idempotent:

- user makes request to pay invoice
- gateway tells lnd to pay invoice, it succeeds, returns success respones
- gateway restarts
- lnd crashes
- user client requests to pay same invoice

First user client will get successful response, second one will get failed response  (because node is down and the `contractid -> operationid` map was lost when gateway restarted)

gateway/ln-gateway/src/lib.rs
```diff
@@ -613,14 +618,25 @@ impl Gateway {
         })
     }
 
-    async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<Preimage> {
+    async fn handle_pay_invoice_msg(&mut self, payload: PayInvoicePayload) -> Result<Preimage> {
```
justinmoon, line 621: Does this need to be `mut`?

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: Removed

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Nope, removed

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: Changed to use `TrackPaymentV2` which solves the above problems

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -265,43 +276,24 @@ impl LightningPayFunded {
...
-                    Ok(contract) => contract.contract.timelock,
-                    Err(_) => {
-                        return LightningPayStateMachine {
-                            common: old_state.common,
-                            state: LightningPayStates::Failure(
-                                "Failed to retrieve OutgoingContract".to_string(),
-                            ),
-                        }
-                    }
-                };
```
elsirion, line 294: Great idea to just keep the timelock around!

elsirion (approved):

elsirion:
> the `contractid -> operationid` map was lost when gateway restarted

I don't think this can happen if the state machine was constructed correctly. The gateway should always persist the contract first and then try to pay. Caveat: I only to a quick glance at the GW side of the new client.

PR #2698, opened 2023-06-29
feat: a "general" federation api error

dpc:
In a common api versions discovery PR, I needed to express an error where not a single memeber was actually at fault, and I had to do it using member errors field and a fake, fixed `peer-0`.

This change addrs a field for `general` member error.

### DIFF
--- a/fedimint-core/src/query.rs
+++ b/fedimint-core/src/query.rs
@@ -249,7 +249,10 @@ impl<R: Eq + Clone + Debug> QueryStrategy<R> for CurrentConsensus<R> {
         }
 
         if self.errors.len() >= self.required {
-            return QueryStep::Failure(mem::take(&mut self.errors));
+            return QueryStep::Failure {
+                general: None,
+                members: mem::take(&mut self.errors),
+            };
         }
 
         QueryStep::Continue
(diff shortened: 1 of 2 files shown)


fedimint-core/src/query.rs
```diff
@@ -512,17 +515,16 @@ impl QueryStrategy<SupportedApiVersionsSummary, ApiVersionSet> for DiscoverApiVe
...
+                        general: Some(e),
+                        members: BTreeMap::new(),
+                    },
                 }
             }
             QueryStep::RetryMembers(v) => QueryStep::RetryMembers(v),
             QueryStep::FailMembers(v) => QueryStep::FailMembers(v),
             QueryStep::Continue => QueryStep::Continue,
-            QueryStep::Failure(v) => QueryStep::Failure(v),
+            QueryStep::Failure { general, members } => QueryStep::Failure { general, members },
```
dpc, line 527: Interesting fact: Since these two are different types, one can't just use `failure @ ... => failure`.

fedimint-core/src/query.rs
```diff
@@ -547,5 +549,10 @@ pub enum QueryStep<R> {
     /// Return the successful result
     Success(R),
     /// Fail the whole request and remember errors from given members
-    Failure(BTreeMap<PeerId, MemberError>),
+    /// Note: member errors are to be added to any errors previously returned
+    /// with `FailMembers`
+    Failure {
```
dpc, line 554: Somewhat surprisingly, this is a different type from `FederationError` because it has very different semantics (`members` is meant to be added to previous member errrors).

douglaz (approved):
LGTM

elsirion (approved):

PR #2703, opened 2023-07-01
fix(client): remove redundant `await_tx_rejected` fn

elsirion:
Removes one of the many client-ng TODOs.


m1sterc001guy (approved):
LGTM looks like there just might be a flaky test

douglaz (approved):
LGTM

Issue #2696, opened 2023-06-28
Waiting a transaction to complete uses 100% of cpu on client


douglaz:
For instance if we call `await_tx_accepted` on a transaction, it will be a busy loop:

```
2023-06-28T20:53:05.316487Z DEBUG find_by_prefix{key=ActiveModuleOperationStateKeyPrefix { operation_id: OperationId(1f3a22eec8118166655e4ca3168e28928ab2505702f8ee095c0767213d91de6a), module_instance: 65535, _pd: PhantomData<fedimint_client::DynGlobalClientContext>[80/1947]
mint_core::db: find by prefix                                                                                                                                                                                                                                                     
2023-06-28T20:53:05.323885Z DEBUG find_by_prefix{key=ActiveStateKeyPrefix(PhantomData<fedimint_client::DynGlobalClientContext>)}: fedimint_core::db: find by prefix                                                                                                               
2023-06-28T20:53:05.323960Z DEBUG fedimint_client::sm::executor: No state transitions available, waiting before re-trying                                                                                                                                                         
2023-06-28T20:53:05.355741Z DEBUG find_by_prefix{key=InactiveModuleOperationStateKeyPrefix { operation_id: OperationId(1f3a22eec8118166655e4ca3168e28928ab2505702f8ee095c0767213d91de6a), module_instance: 65535, _pd: PhantomData<fedimint_client::DynGlobalClientContext> }}: fe
dimint_core::db: find by prefix                                                                                                                                                                                                                                                   
2023-06-28T20:53:05.356644Z DEBUG find_by_prefix{key=ActiveModuleOperationStateKeyPrefix { operation_id: OperationId(1f3a22eec8118166655e4ca3168e28928ab2505702f8ee095c0767213d91de6a), module_instance: 65535, _pd: PhantomData<fedimint_client::DynGlobalClientContext> }}: fedi
mint_core::db: find by prefix                                                                                                                                                                                                                                                     
2023-06-28T20:53:05.394827Z DEBUG find_by_prefix{key=InactiveModuleOperationStateKeyPrefix { operation_id: OperationId(1f3a22eec8118166655e4ca3168e28928ab2505702f8ee095c0767213d91de6a), module_instance: 65535, _pd: PhantomData<fedimint_client::DynGlobalClientContext> }}: fe
dimint_core::db: find by prefix                                                                                                                                                                                                                                                   
[...]

PR #2709, opened 2023-07-03
fix: change state machines execution from quadratic to linear time

douglaz:
Closes https://github.com/fedimint/fedimint/issues/2696

This reduces the execution time of a transaction with 250 outputs from 1 hour to just 30 seconds.

The idea is to not throw away the trigger/transitions futures we have created after some state transition happens. We just add the next transitions to the list and keep polling them until completion.

### DIFF
--- a/fedimint-client/src/sm/executor.rs
+++ b/fedimint-client/src/sm/executor.rs
@@ -12,6 +12,7 @@ use fedimint_core::encoding::{Decodable, DecodeError, Encodable};
 use fedimint_core::fmt_utils::AbbreviateJson;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::task::TaskGroup;
+use fedimint_core::util::BoxFuture;
 use fedimint_core::{maybe_add_send_sync, task};
 use futures::future::select_all;
 use futures::stream::StreamExt;
@@ -19,6 +20,7 @@ use tokio::select;
 use tokio::sync::Mutex;
 use tracing::{debug, error, info, trace, warn};
 
+use super::state::StateTransitionFunction;
 use crate::sm::notifier::Notifier;
 use crate::sm::state::{DynContext, DynState};
 use crate::sm::{ClientSMDatabaseTransaction, GlobalContext, OperationId, State, StateTransition};
@@ -267,6 +269,12 @@ where
     }
 }
 
+type TransitionForActiveState<GC> = (
+    serde_json::Value,
+    DynState<GC>,
+    StateTransitionFunction<DynState<GC>>,
+    ActiveState,
+);
 impl<GC> ExecutorInner<GC>
 where
     GC: GlobalContext,
@@ -275,7 +283,7 @@ where
         info!("Starting state machine executor task");
         loop {
             if let Err(err) = self
-                .execute_next_state_transition(&global_context_gen)
+                .execute_next_state_transitions(&global_context_gen)
                 .await
             {
                 warn!(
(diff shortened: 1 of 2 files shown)


fedimint-client/src/sm/executor.rs
```diff
@@ -312,115 +349,133 @@ where
...
-                            dbtx.insert_entry(
-                                &InactiveStateKey::from_state(new_state.clone()),
-                                &ActiveState::new().into_inactive(),
+            let num_states = active_state_count;
+            let num_transitions = transitions.len();
+            debug!(
+                num_states,
+                num_transitions, "Awaiting any state transition to become ready"
+            );
+            let new_state_added = async move {
```
douglaz, line 370: We still break the loop if new states are added and all existing futures are pending. Perhaps we could query the active states and then figure out what states are new in some way. Then we could just add them to the list of transitions.

@elsirion @m1sterc001guy  ideas here?

fedimint-client/src/sm/state.rs
```diff
@@ -89,8 +89,9 @@ where
 }
 
 type TriggerFuture = Pin<Box<maybe_add_send!(dyn Future<Output = serde_json::Value> + 'static)>>;
+
 // TODO: remove Arc, maybe make it a fn pointer?
-type StateTransitionFunction<S> = Arc<
+pub(super) type StateTransitionFunction<S> = Arc<
```
dpc, line 94: TIL

fedimint-client/src/sm/executor.rs
```diff
@@ -775,6 +830,16 @@ where
     type Record = InactiveStateKey<GC>;
 }
 
+enum ActiveOrInactiveState<GC> {
```
dpc, line 833: Kind of a weird name. `TriggerFunctionState` (or something? I'm not sure yet what it is)

fedimint-client/src/sm/executor.rs
```diff
@@ -267,6 +269,12 @@ where
     }
 }
 
+type TransitionForActiveState<GC> = (
```
dpc, line 272: With 4 fields, it looks like each field should get a name.

fedimint-client/src/sm/executor.rs
elsirion: Refactoring this into a channel would be good, I think I wanted to avoid forgetting a place where states are created when I built this and didn't have the time for a full refactor.

In the channel version we'd only load from DB on startup and then only add states via the channel. We'd still save the state to DB before sending it over the channel for crash safety though.

fedimint-client/src/sm/executor.rs
```diff
@@ -312,115 +349,133 @@ where
...
+                                let v = ActiveState::new();
+                                dbtx.insert_entry(&k, &v).await;
+                                Ok(ActiveOrInactiveState::Active {
+                                    dyn_state: new_state,
+                                    active_state: v,
+                                })
+                            }
+                        })
+                    },
+                    Some(100),
```
elsirion, line 449: Complaint to self: needs a const/name

fedimint-client/src/sm/executor.rs
elsirion: `StateTransitionOutcome` maybe? Although I don't have a problem with the current name.

elsirion (approved):
Definitely a step into the right direction, next would be figuring out how to not poll the DB for new states.

Not fully related but maybe helpful, I'm also a fan of DPCs idea mentioned some months ago of having state machine IDs. I could imagine the identity by a state to be defined as a triple of operation ID, state machine ID and state number (indicating how many transitions that particular SM already had).

PR #2710, opened 2023-07-04
refactor: move common version discovery to module gen (from module)

dpc:
This will allow passing the api version to use to the `init` function, so the client module implementation can use it to communicate using right APIs.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -724,40 +724,57 @@ impl Client {
         })
     }
 
-    /// Query the federation for API version support and then calculate
-    /// the best API version to use (supported by most guardians).
     pub async fn discover_common_api_version(&self) -> anyhow::Result<ApiVersionSet> {
         Ok(self
             .api()
-            .discover_api_version_set(&self.supported_api_versions_summary().await)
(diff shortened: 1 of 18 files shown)


fedimint-client/src/lib.rs
```diff
@@ -686,7 +686,7 @@ impl Client {
     }
 
     /// Returns the config with which the client was initialized.
-    pub async fn get_config(&self) -> &ClientConfig {
+    pub fn get_config(&self) -> &ClientConfig {
```
elsirion, line 689: Probably ok to do this change, would be a problem if we ever decide to only keep the config in the DB and load on demand (unlikely imo).

fedimint-client/src/module/gen.rs
```diff
@@ -28,10 +28,12 @@ pub trait ClientModuleGen: ExtendsCommonModuleGen + Sized {
     fn supported_api_versions(&self) -> MultiApiVersion;
 
     /// Initialize a [`ClientModule`] instance from its config
+    #[allow(clippy::too_many_arguments)]
```
elsirion, line 31: :laughing:

elsirion (approved):

PR #961, opened 2022-11-21
Use database transaction for input/output validation to get rid of `ConflictFilter`

elsirion:
Previously we needed to check if an epoch contained conflicting transactions since we couldn't *read* from a database batch. When using DB transactions, from which we can read, that's not necessary anymore. This helps with modularization.

PR #2715, opened 2023-07-05
fix: prevent double spend of ecash note within the same MintInput

joschisan:


elsirion (approved):
Great find! This regression was introduced by removing the `ConflictFilter` after migrating to true database transactions in #961. 

I was under the wrongful assumption that if conflicts between inputs could now be detected, we wouldn't need any additional filtering anymore. I overlooked conflicts between e-cash notes inside *one single* input though.

dpc:
This suggests we should have a test for it...

Issue #2665, opened 2023-06-19
Documentation for docker-based deployment of gatewayd and fedimintd


justinmoon:

nassersaazi:
@justinmoon i could write this up. any hints on where to look? 👀

nassersaazi:
i have successfully run the docker-compose in the ui repo and i have also looked at scripts/start-containers.sh in the fedimint repo, though here i got some problems with certs generation. 
 i have some questions:

- how different are these two in terms of functionality and containers run?
- is it the right approach to just transfer and document the docker-compose set up from the ui repo and put it into the fedimint repo or are there some additional steps required.
- am also curious about why fedimint opted for scripts instead of setting up a docker-compose directly as it's done in the UI repo?

m1sterc001guy:
`scripts/start-containers.sh` is mostly legacy and was built when we only could do DKG from the cli rather than through the new UI. Now that the new UI is in place, it makes it much easier to create a single docker-compose file rather than use the functions in `scripts/start-containers.sh`.

Using the docker-compose file is the correct way to go and we should probably delete `scripts/start-containers.sh` now that that's in place.

justinmoon:
This is done

PR #2725, opened 2023-07-06
feat: minimum docker-compose for fedimintd + ui

douglaz:
Partially addresses https://github.com/fedimint/fedimint/issues/2665


dpc (dismissed):

m1sterc001guy:
Maybe remove `scripts/start-containers.sh` too? That script shouldn't be necessary anymore with this docker-compose file

docker/fedimintd/docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - "0.0.0.0:8173:8173"
+      - "0.0.0.0:8174:8174"
+    environment:
+      - FM_BITCOIN_RPC_KIND=esplora
+      - FM_BITCOIN_RPC_URL=https://mutinynet.com/api/
+      - FM_BITCOIN_NETWORK=signet
+      - FM_BIND_P2P=0.0.0.0:8173
+      - FM_P2P_URL=fedimint://127.0.0.1:8173
+      - FM_BIND_API=0.0.0.0:8174
+      - FM_API_URL=ws://127.0.0.1:8174
```
elsirion, line 17: If someone was to use this in production they should really put a https reverse proxy in front of fedimint (making this a `wss://…` URL). Maybe we should add a warning to that effect? I don't think documenting "how to run NGINX" is really worth it rn.

elsirion (approved):
Looks good, one nit, feel free to merge+follow up though.

PR #2726, opened 2023-07-06
fix: improve version hash detection and handling

dpc:
Don't silently ignore failures.

### DIFF
--- a/fedimint-build/src/lib.rs
+++ b/fedimint-build/src/lib.rs
@@ -1,14 +1,19 @@
 use std::env;
 use std::process::Command;
 
+/// Env variable to set to force git hash during build process
 const FORCE_GIT_HASH_ENV: &str = "FEDIMINT_BUILD_FORCE_GIT_HASH";
+
+/// Env variable the cargo will set during crate build to pass the detected git
+/// hash to the binary itself.
+const GIT_HASH_ENV: &str = "FEDIMINT_BUILD_CODE_VERSION";
 
 fn set_code_version_inner() -> Result<(), String> {
     println!("cargo:rerun-if-env-changed={FORCE_GIT_HASH_ENV}");
 
     if let Ok(hash) = env::var(FORCE_GIT_HASH_ENV) {
         eprintln!("Forced hash via {FORCE_GIT_HASH_ENV} to {hash}");
-        println!("cargo:rustc-env=CODE_VERSION={hash}");
+        println!("cargo:rustc-env={GIT_HASH_ENV}={hash}");
         return Ok(());
     }
 
@@ -34,7 +39,7 @@ fn set_code_version_inner() -> Result<(), String> {
         }
     };
 
-    println!("cargo:rustc-env=CODE_VERSION={hash}");
+    println!("cargo:rustc-env={GIT_HASH_ENV}={hash}");
 
     Ok(())
 }
--- a/fedimint-build/src/lib.rs
+++ b/fedimint-build/src/lib.rs
@@ -1,21 +1,49 @@
 use std::env;
-use std::fs::File;
 use std::process::Command;
 
-pub fn set_code_version() {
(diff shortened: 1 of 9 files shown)


fedimint-build/src/lib.rs
```diff
@@ -1,21 +1,59 @@
...
+    if let Ok(hash) = env::var(FORCE_GIT_HASH_ENV) {
+        eprintln!("Forced hash via {FORCE_GIT_HASH_ENV} to {hash}");
+        println!("cargo:rustc-env={GIT_HASH_ENV}={hash}");
+        return Ok(());
+    }
+
+    let output = match Command::new("git")
+        .args(["rev-parse", "HEAD"])
+        .stdout(Stdio::piped())
+        .stderr(Stdio::inherit())
```
Maan2003, line 23:
```suggestion
```

fedimint-build/src/lib.rs
```diff
@@ -1,21 +1,59 @@
...
+        Err(e) => {
+            return Err(format!("Failed to execute `git` command: {e}"));
+        }
+    };
+
+    if !output.status.success() {
+        return Err(format!(
+            "`git` command failed: stderr: {}; stdout: {}",
+            String::from_utf8_lossy(&output.stderr),
+            String::from_utf8_lossy(&output.stdout)
```
Maan2003, line 36: >         .stderr(Stdio::inherit())

this will be empty?

fedimint-build/src/lib.rs
```diff
@@ -1,21 +1,54 @@
...
+fn set_code_version_inner() -> Result<(), String> {
+    println!("cargo:rerun-if-env-changed={FORCE_GIT_HASH_ENV}");
+
+    if let Ok(hash) = env::var(FORCE_GIT_HASH_ENV) {
+        eprintln!("Forced hash via {FORCE_GIT_HASH_ENV} to {hash}");
+        println!("cargo:rustc-env={GIT_HASH_ENV}={hash}");
+        return Ok(());
+    }
+
+    let output = match Command::new("git").args(["rev-parse", "HEAD"]).output() {
```
Maan2003, line 20: there is no `rerun-if` for this, means this will never get rerun

so 
```
cargo build
... commit stuff ...
cargo build
```

will not update the hash.

dpc: @Maan2003 Oh. You're right. But is there anything we can do about it? Do we just need to force it to run on every build?

Maan2003: rerunning on every build is sad, it will remove all intecremental compilation :/

maybe we can just afford this being incorrect?

dpc: I'm not sure if we really care. E.g. right now we don't record anywhere a dirty hash, so just because the local binary has a given hash, doesn't mean it wasn't built with something far, far detached from it.

Nix-built binaries is what really matters.

Maan2003: and in nix we are not even setting the hash :smile:

dpc: @Maan2003 

```
    // Note: best effort approach to force a re-run when the git hash in
    // the local repo changes without wrecking the incremental compilation
    // completely.
    if Path::new(".git/HEAD").exists() {
        println!("cargo:rerun-if-changed=.git/HEAD");
    }
    // Common(?) `git workdir` setup
    if Path::new("../HEAD").exists() {
        println!("cargo:rerun-if-changed=../HEAD");
    }
```

LGTY?

Maan2003: I don't know about workdir. but what is .git/HEAD relative to?

dpc: Ah, damn it. Right. @Maan2003 How about now?

Maan2003 (approved):
LGTM

PR #2727, opened 2023-07-06
refactor: move versioning into own module and add some docs

dpc:


fedimint-core/src/module/version.rs
```diff
@@ -0,0 +1,372 @@
+//! Fedimint consensus and API versioning.
```
dpc, line 1: Some higher level summary that I've added.

elsirion (approved):

PR #2728, opened 2023-07-06
fix: don't config gen in run-ui

Maan2003:


justinmoon:
Pushed a commit setting the metrics port. Also changed it to start from 3510 instead of 3000 (which was used by fedimint/ui).

elsirion:
Is there any blocker apart from CI?

justinmoon:
> Is there any blocker apart from CI?

I don't think so. Might want to rename this command. It used to actually run the UI server, but that doesn't exist anymore. Might rename to something like `dev-fed-no-dkg` or something.

Maan2003:
what about `devimint fed-setup`

justinmoon (approved):

PR #2732, opened 2023-07-07
fix: remove FM_PID_FILE

justinmoon:
devimint can cleanup the the child processes it spawns, so we don't need a text fill containing these process IDs. We can just kill devimint.

Closes https://github.com/fedimint/fedimint/issues/2236

### DIFF
--- a/integrationtests/README.md
+++ b/integrationtests/README.md
@@ -69,5 +69,5 @@ cargo test -p fedimint-tests -- --test-threads=1
 
 If you wish to clean-up the services run:
 ```shell
-kill_fedimint_processes
+pkill devimint
 ```
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -17,13 +17,11 @@ else
   FM_TEST_DIR="${2-"$(mktemp --tmpdir -d XXXXX)"}"
 fi
 export FM_TEST_DIR
-export FM_PID_FILE="$FM_TEST_DIR/.pid"
 export FM_LOGS_DIR="$FM_TEST_DIR/logs"
 
 echo "Setting up env variables in $FM_TEST_DIR"
 
 mkdir -p "$FM_TEST_DIR"
-touch "$FM_PID_FILE"
 
 # Symlink $FM_TEST_DIR to local gitignored target/ directory so they're easier to find
 rm -f target/devimint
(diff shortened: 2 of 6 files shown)


scripts/build.sh
```diff
@@ -36,16 +32,9 @@ if [ -z "${SKIP_CARGO_BUILD:-}" ]; then
...
-  echo "Killing fedimint processes"
-  PIDS=$(cat $FM_PID_FILE | sed '1!G;h;$!d') # sed reverses order
-  if [ -n "$PIDS" ]
-  then
-    kill $PIDS 2>/dev/null
-  fi
-  rm -f $FM_PID_FILE
+function kill_devimint {
+  echo "Killing devimint and child processes"
+  pkill devimint
```
m1sterc001guy, line 37: Does this need -9?

integrationtests/README.md
```diff
@@ -66,8 +66,3 @@ You can now run the integration tests against real instances of Bitcoin and Ligh
 export FM_TEST_USE_REAL_DAEMONS=1
 cargo test -p fedimint-tests -- --test-threads=1
 ```
-
-If you wish to clean-up the services run:
-```shell
-kill_fedimint_processes
```
m1sterc001guy, line 72: Maybe add the equivalent documentation for `devimint`?

m1sterc001guy (approved):
So much cleaner :)

scripts/build.sh
Maan2003: no I think, SIGTERM is better. this will allow devimint to kill its children.

dpc, line 37: What if I'm running multiple devimints at the same time? :P

elsirion: Maybe it should just be a pid file for devimint?

justinmoon: I don't think devimint supports multiple concurrent sessions. 

If we're going to have a PID file, I suggest devimint manages it and not shell scripts. Currently you need to copy a bunch of shell scripts in order to use devimint in external repository which sucks.

dpc: Fedimint should manage it, yes.

douglaz (dismissed):
LGTM

douglaz:
That's awkward, many tests are failing :thinking:

elsirion:
There seems to be some fundamental problem that restarting CI doesn't fix. I couldn't pinpoint it yet, can you think of anything @justinmoon?

justinmoon:
I don't know why tests are failing here. For external integrations it would be nice to not have to worry about this file. Going to leave as draft for now.

integrationtests/README.md
```diff
@@ -69,5 +69,5 @@ cargo test -p fedimint-tests -- --test-threads=1
 
 If you wish to clean-up the services run:
 ```shell
-kill_fedimint_processes
+pkill devimint
```
dpc, line 72: It's probably this. It's just very meh way to do stuff.

scripts/build.sh
dpc: It's probably this (causing tests to fail). It's just very meh way to do stuff.

dpc:
Go with #3007

PR #2736, opened 2023-07-07
 feat(client): cache common api versions to use in the DB 

dpc:


fedimint-core/src/api.rs
```diff
@@ -180,10 +180,28 @@ pub trait FederationApiExt: IFederationApi {
...
+        params: ApiRequestErased,
+    ) -> FederationResult<FedRet> {
+        self.request_with_strategy_with_timeout(strategy, method, params, None)
+            .await
+    }
+
+    /// Make an aggregate request to federation, using `strategy` to logically
+    /// merge the responses.
+    ///
+    /// Optionally put a timeout on every request.
```
elsirion, line 195: The implicit result might be that the request might run forever, right? Because different peers may time out and then work again, never leading to complete failure but also not to success.

fedimint-client/src/lib.rs
```diff
@@ -782,6 +783,64 @@ impl Client {
...
+            let module_gens = module_gens.clone();
+            let api = api.clone();
+            let db = db.clone();
+            tokio::task::spawn(async move {
+                if let Err(e) =
+                    Self::refresh_common_api_version_static(&config, &module_gens, &api, &db).await
+                {
+                    warn!("Failed to discover common api versions: {e}");
+                }
+            });
```
elsirion, line 814: This isn't in a task group because it's short-lived/potentially spawned many times? Also, didn't we have some WASM-compat `spawn` method?

elsirion (approved):
Looks good, just some nits :)

fedimint-core/src/api.rs
dpc: Yes.

Now that mention it, I think maybe the timeout should be provided by a strategy, and not a separate argument. Because it only makes sense in the context of a specific strategy that relies on it...

fedimint-client/src/lib.rs
dpc: Yeah, I don't think we need to wait for it. It will either succeed or not. It's kind of a best effort side quest.

And yeah, I think I should use fedimint_core::task instead.

Maan2003:
`ClientBuilder::build` will always use a old api versions.
So after federation upgrade, client will always try to connect assuming old versions at least for one session. 

Is this correct?

dpc:
> So after federation upgrade, client will always try to connect assuming old versions at least for one session.

Yes.

We might tweak with it in the future in various ways (eg. federation with a scheduled consensus change would pass a flag to change this behavior, or cached entries expiring after a time, etc.)

Maan2003 (approved):

PR #2739, opened 2023-07-09
fix: remove the concept of epochs and dropping of peers from FedimintConsensus

joschisan:


fedimint-server/src/consensus/mod.rs
```diff
@@ -142,30 +133,34 @@ impl FedimintConsensus {
...
+                            .flat_map(|(peer_id, items)| items.into_iter().map(move |item| (peer_id, item)));
+
+                        for (peer_id, consensus_item) in items {
+                            dbtx.set_tx_savepoint()
+                                .await
+                                .expect("Setting transaction savepoint failed");
+
+                            let result = self.process_consensus_item(dbtx, consensus_item.clone(), peer_id).await;
+
+                            if result.is_err(){
```
douglaz, line 152: Perhaps log this error?

fedimint-server/src/consensus/mod.rs
```diff
@@ -142,30 +133,34 @@ impl FedimintConsensus {
...
+                                dbtx.rollback_tx_to_savepoint()
+                                    .await
+                                    .expect("Rolling back transaction to savepoint failed");
+
+                                if let ConsensusItem::Transaction(transaction) = consensus_item.clone() {
+                                    if self.process_consensus_item(dbtx, consensus_item, peer_id).await.is_err() {
```
douglaz, line 158: And/or this one?

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -342,79 +341,83 @@ impl ServerModule for Wallet {
...
+                let unsigned_key = UnsignedTransactionKey(txid);
+
+                let mut unsigned = match dbtx.get_value(&unsigned_key).await {
+                    Some(unsigned) => unsigned,
+                    None => bail!("Unsigned trasnaction for this peg out signature does not exist"),
+                };
+
+                if self
+                    .sign_peg_out_psbt(&mut unsigned.psbt, &peer_id, &peg_out_signature)
+                    .is_err()
```
douglaz, line 399:
```suggestion
                self
                    .sign_peg_out_psbt(&mut unsigned.psbt, &peer_id, &peg_out_signature)
                    .context("Peg out signature is invalid")?;
```

fedimint-server/src/consensus/mod.rs
```diff
@@ -190,205 +186,162 @@ impl FedimintConsensus {
...
+                if signature_shares.len() <= pks.threshold() {
+                    return Ok(());
+                }
 
-        for peer in drop_peers {
-            dbtx.insert_entry(&DropPeerKey(peer), &()).await;
-        }
+                let threshold_signature = pks
+                    .combine_signatures(signature_shares.iter().map(|(peer, share)| (peer, share)))
+                    .expect("All signature shares are valid");
```
douglaz, line 316: Do we really need to abort the process?
```suggestion
                    .context("All signature shares are valid")?;
```

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -342,79 +341,83 @@ impl ServerModule for Wallet {
...
+                {
+                    bail!("Peg out signature is invalid");
+                }
```
douglaz, line 402:
```suggestion
                self
                    .sign_peg_out_psbt(&mut unsigned.psbt, &peer_id, &peg_out_signature)
                    .context("Peg out signature is invalid")?;
```

fedimint-server/src/consensus/mod.rs
joschisan: This result will later be used to decide wether to accept the consensus item into the history of the federation or to discard it. An error is therefore also returned in case of items that may be valid but are redundant - we only need a threshold of 2f+ 1 sig shares for an ecash note so the last f shares will be discarded; if an item is actually invalid we not only return an error but also warn! directly inside the functions. Therefore I do not log an error here...

fedimint-server/src/consensus/mod.rs
joschisan: This call actually have not been there in the first place.

fedimint-server/src/consensus/mod.rs
joschisan: We have checked the signature shares before we saved them to disk, so if the validation fails either our cryptography library has a bug and contradicts itself or our database  has been corrupted so I think we should panic.

modules/fedimint-wallet-server/src/lib.rs
joschisan: This is a good suggestion and has me actually questioning wether we should differentiate between invalid items we should log a warning for and merely redundant items in the returned result so we can log the warning at a higher level and use context as you described. converted the pr back to a draft to work this out...

joschisan: I have update the code to only return an error if we want it to be logged, otherwise we return an Ok(ConsensusDecision::Discard). Then the caller can log the error.

fedimint-server/src/consensus/mod.rs
```diff
@@ -142,35 +134,43 @@ impl FedimintConsensus {
...
+                                Err(error) => {
+                                    warn!(target: "consensus", "Invalid consensus item from {peer_id}: {error}");
+                                    ConsensusDecision::Discard
+                                }
+                            };
+
+                            if decision == ConsensusDecision::Discard {
+                                dbtx.rollback_tx_to_savepoint()
+                                    .await
+                                    .expect("Rolling back transaction to savepoint failed");
```
elsirion, line 162: Is this just an added safety or is it expected that modules may have changed the DB even if the item is discarded? Both are ok, just for my understanding.

fedimint-server/src/consensus/mod.rs
```diff
@@ -190,205 +191,165 @@ impl FedimintConsensus {
...
+        match consensus_item {
+            ConsensusItem::Module(module_item) => {
+                let moduletx = &mut dbtx.with_module_prefix(module_item.module_instance_id());
+
+                self.modules
+                    .get_expect(module_item.module_instance_id())
+                    .process_consensus_item(moduletx, module_item, peer_id)
+                    .await
+            }
+            ConsensusItem::Transaction(transaction) => {
```
elsirion, line 217: Maybe all these match arms should be separate functions?

fedimint-server/src/db.rs
```diff
@@ -2,26 +2,24 @@ use std::collections::BTreeSet;
...
 #[repr(u8)]
 #[derive(Clone, EnumIter, Debug)]
 pub enum DbKeyPrefix {
     AcceptedTransaction = 0x02,
-    DropPeer = 0x03,
-    RejectedTransaction = 0x04,
     EpochHistory = 0x05,
     LastEpoch = 0x06,
     ClientConfigSignature = 0x07,
+    ClientConfigSignatureShare = 0x3,
```
elsirion, line 22: Please add `0` for uniformity

fedimint-server/src/net/api.rs
```diff
@@ -141,95 +135,65 @@ impl ConsensusApi {
...
-            }
-        }
+    pub async fn wait_transaction_status(&self, txid: TransactionId) -> TransactionStatus {
+        let (outputs, mut dbtx) = self
+            .db
+            .wait_key_check(&AcceptedTransactionKey(txid), std::convert::identity)
+            .await;
+
+        self.accepted_transaction_status(txid, outputs, &mut dbtx)
+            .await
```
elsirion, line 170: This is a huge semantic change, we need to make sure the client still works in error cases. It's likely some state machines need to be changed.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -377,167 +377,157 @@ impl ServerModule for Lightning {
...
+                    .find_by_prefix(&AgreedDecryptionShareContractIdPrefix(contract_id))
+                    .await
+                    .map(|(key, decryption_share)| (key.1, decryption_share))
+                    .collect::<Vec<_>>()
+                    .await;
 
-                    // we save the first ordered valid decryption share for every peer
-                    dbtx.insert_new_entry(&AgreedDecryptionShareKey(contract_id, peer_id), &share)
-                        .await;
+                if decryption_shares.len() < self.cfg.consensus.threshold() {
```
elsirion, line 432: There appear to be some changes in here that are hard to review since they are mixed with reformatting. I think we can go ahead, but maybe these should have been separate commits doing these changes before the for loop unroll.

fedimint-server/src/consensus/mod.rs
```diff
@@ -142,35 +134,43 @@ impl FedimintConsensus {
...
-                        let epoch = consensus_outcome.epoch;
-                        let outcome = consensus_outcome.clone();
-
-                        let UnzipConsensusItem {
-                            epoch_outcome_signature_share: _epoch_outcome_signature_share_cis,
-                            client_config_signature_share: _client_config_signature_share_cis,
-                            transaction: transaction_cis,
-                            consensus_upgrade: consensus_upgrade_cis,
-                            module: module_cis,
-                        } = consensus_outcome
```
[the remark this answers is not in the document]
elsirion: Call: still used by recoverytool

fedimint-server/src/consensus/mod.rs
elsirion: Call: tried it, seemed less readable

fedimint-server/src/consensus/mod.rs
elsirion: Call: for current modules this is just an additional safety net. But it's easier to always rollback and not just for transactions, so we will make it part of the module interface contract that modules *may* change the DB.

PR #2742, opened 2023-07-10
 chore: encode length of dyn types in their encodings 

dpc:
This allows clients missing corresponding module implementations
to skip these, without losing the ability to decode the structures
containing them.

The encoding is:

```
  instance_id | length | data
```


fedimint-core/src/core.rs
```diff
@@ -112,11 +125,17 @@ macro_rules! module_dyn_newtype_impl_encode_decode {
                 modules: &$crate::module::registry::ModuleDecoderRegistry,
             ) -> Result<Self, fedimint_core::encoding::DecodeError> {
                 let key = fedimint_core::core::ModuleInstanceId::consensus_decode(reader, modules)?;
-                modules.get_expect(key).decode(reader, key, modules)
+                let len = u32::consensus_decode(reader, modules)?;
+                let mut buf = vec![0; len as usize];
```
dpc, line 129: DoS

fedimint-core/src/core.rs
```diff
@@ -111,12 +124,28 @@ macro_rules! module_dyn_newtype_impl_encode_decode {
                 reader: &mut R,
                 modules: &$crate::module::registry::ModuleDecoderRegistry,
             ) -> Result<Self, fedimint_core::encoding::DecodeError> {
-                let key = fedimint_core::core::ModuleInstanceId::consensus_decode(reader, modules)?;
-                modules.get_expect(key).decode(reader, key, modules)
+                let module_instance_id =
+                    fedimint_core::core::ModuleInstanceId::consensus_decode(reader, modules)?;
+                let total_len = u32::consensus_decode(reader, modules)? as usize;
+                let mut buf = Vec::with_capacity(512);
```
douglaz, line 130: Why not something like:
```suggestion
                let mut buf = vec![0u8; total_len as usize];
                reader
                    .read_exact(&mut buf)
                    .map_err(|e| fedimint_core::encoding::DecodeError::new_custom(e.into()))?;
```

dpc: @douglaz `total_len` is potentially untrusted and could be set to `0xfffffffff...` by someone malicious.

dpc: I actually had it like that in first version before I realized it's no good.

douglaz: That's a good point, but with this algorithm in the end you will still have a buffer of `total_len`, right?

Issue #2747, opened 2023-07-10
LND Stuck Payments


m1sterc001guy:
When paying many invoices in parallel using LND gateway, sometimes payments get stuck. It is not deterministic but it is fairly easy to reproduce.

Repro:

`fedimint-load-test-tool --users 10 load-test --generate-invoice-with cln-lightning-cli`

Doesn't finish:
![image](https://github.com/fedimint/fedimint/assets/1859166/ff8be5db-39ec-4169-a6d8-89eeec09b7b9)

Payments are still IN_FLIGHT according to LND:
![image](https://github.com/fedimint/fedimint/assets/1859166/249f1d5c-c08d-4bdc-991e-10792beccd8c)

m1sterc001guy:
So I implemented this https://github.com/fedimint/fedimint/issues/2301 hoping it would solve the issue.

`SendPaymentV2` has the same problem though. I suspect `SendPaymentV2` is not atomic. I see this in the LND logs when I replicate the issue:

![image](https://github.com/fedimint/fedimint/assets/1859166/e7d9490d-8bc5-413e-8ce0-98073be9bcb5)

Presumably, `SendPaymentV2` is being pre-empted due to our state machine framework, which is causing the payment to become stuck

Issue #2301, opened 2023-04-23
lnd: Switch to asynchronous SendPaymentV2 gRPC method


justinmoon:
We are currently using the [lightningrpc.SendPaymentSync](https://lightning.engineering/api-docs/api/lnd/lightning/send-payment-sync#grpc) gRPC API. According to the docs, "This RPC is intended to be consumed by clients of the REST proxy" which is not what we're doing.

Should we move to the [routerrpc.SendPaymentV2](https://lightning.engineering/api-docs/api/lnd/router/send-payment-v2) gRPC method which returns a stream?

elsirion:
Closed by #2750

PR #2750, opened 2023-07-10
fix: LND Stuck Payments

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2747
Also addresses: https://github.com/fedimint/fedimint/issues/2301

The crux of the stuck payments issue for LND is that `SendPaymentV2` and `SendPaymentSync` are not atomic. This means that if they're pre-empted while executing, it is possible that the payment becomes stuck and will not succeed. 

To fix this, I have moved `buy_preimage_over_lightning` into the transition function so that it can leverage the state machine framework's atomicity when moving between states. Since this function is already idempotent, I believe this is a safe refactor.

I'm now able to get 20+ concurrent payments with LND gateway (maybe more, haven't tried higher)

`fedimint-load-test-tool --users 20 load-test --generate-invoice-with cln-lightning-cli`

image

### DIFF
--- a/gateway/ln-gateway/src/lnd.rs
+++ b/gateway/ln-gateway/src/lnd.rs
@@ -2,7 +2,7 @@ use std::fmt;
 use std::str::FromStr;
 use std::time::Duration;
 
-use anyhow::anyhow;
+use anyhow::{anyhow, Context};
 use async_trait::async_trait;
 use fedimint_core::task::{sleep, TaskGroup};
 use secp256k1::PublicKey;
@@ -236,7 +236,14 @@ impl GatewayLndClient {
                         .await
                         .map_err(|_| GatewayError::ClientNgError)?
                     {
-                        return Ok(Some(payment.payment_preimage));
+                        if payment.status() == PaymentStatus::Succeeded {
+                            return Ok(Some(payment.payment_preimage));
+                        }
+
+                        let failure_reason = payment.failure_reason();
+                        return Err(GatewayError::Other(anyhow!(
+                            "LND payment failed. Failure Reason: {failure_reason:?}"
+                        )));
                     }
                 }
                 Err(e) => {
@@ -416,18 +423,18 @@ impl ILnRpcClient for GatewayLndClient {
                 .into_inner()
                 .message()
                 .await
-                .map_err(|_| GatewayError::ClientNgError)?
+                .context("Failed to get payment status")?
             {
                 Some(payment) if payment.status() == PaymentStatus::Succeeded => {
                     bitcoin_hashes::hex::FromHex::from_hex(payment.payment_preimage.as_str())
-                        .map_err(|_| anyhow::anyhow!("Failed to convert preimage"))?
+                        .context("Failed to convert preimage")?
                 }
                 Some(payment) => {
                     return Err(GatewayError::Other(anyhow!(
                         "LND failed to complete payment: {payment:?}"
                     )));
                 }
-                _ => {
+                None => {
                     return Err(GatewayError::Other(anyhow!(
                         "Failed to get payment status for payment_hash {payment_hash:?}"
                     )));
--- a/gateway/ln-gateway/src/lnd.rs
+++ b/gateway/ln-gateway/src/lnd.rs
@@ -10,12 +10,11 @@ use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::Status;
 use tonic_lnd::lnrpc::failure::FailureCode;
-use tonic_lnd::lnrpc::fee_limit::Limit;
-use tonic_lnd::lnrpc::{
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/lnd.rs
```diff
@@ -394,31 +393,46 @@ impl ILnRpcClient for GatewayLndClient {
...
-            send_response.payment_preimage
+            match payments
+                .into_inner()
+                .message()
+                .await
+                .map_err(|_| GatewayError::ClientNgError)?
+            {
+                Some(payment) if payment.status() == PaymentStatus::Succeeded => {
+                    bitcoin_hashes::hex::FromHex::from_hex(payment.payment_preimage.as_str())
+                        .map_err(|_| anyhow::anyhow!("Failed to convert preimage"))?
```
douglaz, line 423:
```suggestion
                        .context("Failed to convert preimage")?
```
or
```suggestion
                        .with_context(|| anyhow::anyhow!("Failed to convert preimage: {}", payment.payment_preimage))?
```

gateway/ln-gateway/src/lnd.rs
```diff
@@ -394,31 +393,46 @@ impl ILnRpcClient for GatewayLndClient {
...
-                    tonic::Code::Internal,
-                    "LND did not return a preimage",
-                )));
-            };
+                .map_err(|e| {
+                    GatewayError::Other(anyhow!("Failed to make outgoing payment: {e:?}"))
+                })?;
 
-            send_response.payment_preimage
+            match payments
```
douglaz, line 415: payments is a stream, I think you should loop calling .message() until you get None

gateway/ln-gateway/src/lnd.rs
```diff
@@ -394,31 +393,46 @@ impl ILnRpcClient for GatewayLndClient {
...
+                .into_inner()
+                .message()
+                .await
+                .map_err(|_| GatewayError::ClientNgError)?
```
douglaz, line 419:
```suggestion
                .map_err(|e| GatewayError::Other(anyhow!("Failed to get payment status: {e:?}")))?
```

gateway/ln-gateway/src/lnd.rs
```diff
@@ -242,16 +243,14 @@ impl GatewayLndClient {
                     // Break if we got a response back from the LND node that indicates the payment
                     // hash was not found.
                     if e.code() == Code::NotFound {
-                        break;
+                        return Ok(None);
```
douglaz, line 246: In the code above I think you should check `payment.failure_reason` before returning the preimage, if there is an error the preimage will be invalid

douglaz: Also check `.status` field to see if equals to `SUCCEEDED`

gateway/ln-gateway/src/lnd.rs
douglaz: I think here you should loop getting .message() until `.status` is `SUCCEEDED`

gateway/ln-gateway/src/lnd.rs
```diff
@@ -394,31 +393,46 @@ impl ILnRpcClient for GatewayLndClient {
...
+                .message()
+                .await
+                .map_err(|_| GatewayError::ClientNgError)?
+            {
+                Some(payment) if payment.status() == PaymentStatus::Succeeded => {
+                    bitcoin_hashes::hex::FromHex::from_hex(payment.payment_preimage.as_str())
+                        .map_err(|_| anyhow::anyhow!("Failed to convert preimage"))?
+                }
+                Some(payment) => {
+                    return Err(GatewayError::Other(anyhow!(
```
douglaz, line 426: this isn't a failure, it can be inflight, right? you should keep looping the stream

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: Probably sufficient to just check the `status` field, then return the `failure_reason` in the error message?

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: > I think here you should loop getting .message() until .status is SUCCEEDED or you get none

As mentioned offline, I don't think that is necessary since we're using the `no_inflight_updates` flag. We should only receive one update - if the payment succeeds or fails.

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: See this comment https://github.com/fedimint/fedimint/pull/2750/files#r1260345253

gateway/ln-gateway/src/lnd.rs
m1sterc001guy: Changed the above to check the `status` field

douglaz (approved):
LGTM

PR #2752, opened 2023-07-11
chore: switch to variable len. encoding for numbers (using `BigSize`)

dpc:
Note: Unclear do we want to do it at all, if so what encoding should we use.


dpc:
Blocked on https://github.com/fedimint/fedimint/pull/2742 as it will need to recreate migration files as well.

fedimint-core/src/encoding/mod.rs
```diff
@@ -653,16 +735,13 @@ mod tests {
...
-                vec![0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 42],
-            ),
-            (TestEnum::Foo(None), vec![0, 0, 0, 0, 0, 0, 0, 0, 0]),
+            (TestEnum::Foo(Some(42)), vec![0, 1, 42]),
+            (TestEnum::Foo(None), vec![0, 0]),
             (
                 TestEnum::Bar {
                     bazz: vec![1, 2, 3],
                 },
-                vec![0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 3, 1, 2, 3],
```
dpc, line 665: Longer

fedimint-core/src/encoding/mod.rs
```diff
@@ -653,16 +735,13 @@ mod tests {
...
+                vec![1, 3, 1, 2, 3],
```
dpc, line 744: Shorter

fedimint-core/src/encoding/mod.rs
```diff
@@ -143,7 +144,27 @@ impl DecodeError {
...
+    }
+}
+
+impl Decodable for BigSize {
+    fn consensus_decode<R: std::io::Read>(
+        r: &mut R,
+        _modules: &ModuleDecoderRegistry,
+    ) -> Result<Self, DecodeError> {
+        BigSize::read(r)
+            .map_err(|e| DecodeError::new_custom(anyhow::anyhow!("BigSize decoding error: {e}")))
```
douglaz, line 163: Always use `{:?}` for errors to preserve the backtrace
```suggestion
            .map_err(|e| DecodeError::new_custom(anyhow::anyhow!("BigSize decoding error: {e:?}")))
```

fedimint-core/src/encoding/mod.rs
```diff
@@ -575,6 +617,46 @@ impl Decodable for Cow<'static, str> {
     }
 }
 
+/// A writer counting number of writes written to it
+///
+/// Copy&pasted from https://github.com/SOF3/count-write which
```
douglaz, line 622: Clippy will warn you of:
```suggestion
/// Copy&pasted from <https://github.com/SOF3/count-write> which
```

fedimint-core/src/encoding/mod.rs
dpc: > Always use {:?} for errors to preserve the backtrace

Ohh... I didn't know that. Is this documented anywhere? Could we have a lint for it?

dpc:
> Any encoding that permits two valid encodings of the same integer CAN NOT be used by us as we use the the encoding for consensus_hash.

is not really a problem, as "non-minimal" encodings are detected and rejected at decoding as invalid. Not a pretty design, I don't particularly enjoy it, kind of crude, but technically works, sorts lexicographicaly just fine, AFAICT.

> I have come away with the impression that a consensus critical variable length integer encoding is a potential footgun and would therefore advise against it as the space savings seem negligible anyways, or not?

The savings depend on what is being encoded exactly, but generally... there seem to be a lot of numbers in consensus.

I don't share a fear of variable length encoding. Either it's correct, or not. We have bunch of other variable len. encoded stuff like collections, `Dyn*` types etc.

fedimint-core/src/encoding/mod.rs
douglaz: Not sure if it's documented, but it's a fact that I've learned after a lot of pain. I don't think clippy supports that warning and manually implementing such linter is tricky.

douglaz:
> > Any encoding that permits two valid encodings of the same integer CAN NOT be used by us as we use the the encoding for consensus_hash.
> 
> is not really a problem, as "non-minimal" encodings are detected and rejected at decoding as invalid. Kind of crude, but technically works, sorts lexicographicaly just fine, AFAICT.

I saw that comment on LDK and my conclusion is exactly the same: it's just fine

joschisan:
> Note: Unclear do we want to do it at all, if so what encoding should we use.

Right now I do not think a varint encoding is worth our time, at least if our motive is saving space, as there is so much way worse waste of space everywhere in consensus right which we should address instead, for example:

We write every accepted transaction to disk twice but only ever read the module_instace_ids and nothing else. (This is fixed in my next pr)

I believe we also write redundant transactions submitted by different peers within on epoch to disk. (This is gonna be fixed with aleph bft)

Furthermore, for every generated ecash note we put 2f+1 signature shares on the broadcast and therefore also on the disk forever. Every share is 64 bytes and since ecash reissue is probably our most common transaction, each transaction containing several notes, signature shares may easily end up constituting the majority of our disk usage. However, we don’t even not need to order them, the client can easily collect the shares from the guardians and combine the shares by itself. Would also save a the guardians a ton of compute as every signature share also needs to be verified which is actually the most expensive operation in threshold crypto as it requires two pairings.

We use Amount for ecash notes even though we only have 30 tiers which could be encoded in a single byte. With encoding the amounts with variant will only encode them with several bytes as it will encode the actual msats for every tier which may easily require 3-4 bytes.

All of the aforementioned changes either do not add complexity or even reduce it. Therefore my argument is that right now is not the time to worry about varint encodings to save space.

dpc:
It's subjective, but to me varint is not a lot of complexity in grand scheme of things, and a win is a win: every number, enum tag, length will get space savings. We use consensus encoding in more places than only consensus: non-consensus db stuff gets smaller, QR codes get smaller, less bandwidth here and there.

I agree that the optimizations listed by @joschisan are more impactful and would love to see all of them, but they do require more work so we will have to wait longer for them, even if they make things simpler overall, and at least some do add some small complexity (e.g. having to have a federation config to lookup note amount; still totally worth it, IMO!).

elsirion:
I read the code of `BigSize` and they reject non-minimal representations, so I agree with @dpc and @douglaz that it's not a big deal. It will be slightly harder to read hex-encoded data structures when debugging, but I have the feeling that only few people do that anyway. Not having to worry as much about the size of some integer we add here or there is worth it imo.

elsirion (dismissed):
Can you regenerate the migration test DB, then this looks good to go imo.

fedimint-core/src/encoding/mod.rs
```diff
@@ -143,7 +144,27 @@ impl DecodeError {
     }
 }
 
-macro_rules! impl_encode_decode_num {
+pub use lightning::util::ser::BigSize;
```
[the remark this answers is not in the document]
dpc: We could just copy&paste. No shame in that, and `Encodable` is Fedimint-specific anyway. I was just lazy to see if it works, and then it's unclear if we are going to do it at all anyway.

elsirion (approved):

Issue #2651, opened 2023-06-14
`Gatewayd` should not re-construct the `LightningGateway` registrations


m1sterc001guy:
Currently, when `info` is called on `gatewayd` it will return all of the federations that this gateway is registered with. However, it builds these registrations itself, rather than querying each federation. This is problematic because when it re-creates a `LightningGateway`, it now has the wrong timestamp since we're using `now() + TTL` as the time that the registration is valid for.

Ideally I don't think that `info` should return the exact `LightningGateway` structure. Some of this data is repeated across registrations and makes the output difficult to parse.

justinmoon:
Potentially we could just save this to db and return with CLI

PR #2758, opened 2023-07-12
refactor: Gateway ID Refactor

m1sterc001guy:
We currently use `gateway_redeem_key` when the client determines which gateway to use. This works, but once a gateway is registered with > 1 federation, it gets confusing since each of these registrations will have a different `gateway_redeem_key` (since this key is generated according to the root module secret for each client), even though they're the same gateway.

This PR tries to simplify this by renaming `gateway_pub_key` to `gateway_redeem_key` which more accurately describes what it is used for. A new key, `gateway_public_key` is introduced to be used as an identifier for the entire gateway, which will be stable across federations. This key is generated when the gateway first boots and saves it to the database.

This will unblock https://github.com/fedimint/fedimint/issues/2651 and allow us to clean-up the redundant information in `gateway-cli info`


gateway/ln-gateway/src/db.rs
```diff
@@ -45,3 +46,12 @@ impl_db_record!(
     value = LightningGateway,
     db_prefix = DbKeyPrefix::FederationRegistration,
 );
+
+#[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable)]
+pub struct GatewaydPublicKey;
```
douglaz, line 51: `Gateway` or `Gatewayd`?

m1sterc001guy: They're essentially the same thing right now, but I'll change it to `Gateway` just to avoid confusion

okjodom (dismissed):
As a general nit, could we rename `gateway_public_key` to `GatewayId`, for consistency with `FederationId`?

okjodom (approved):

PR #2759, opened 2023-07-12
feat: helpers to install fedimintd/gatewayd with docker

douglaz:


docker/download-mutinynet.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
wbobeirne, line 1: We should document usage of this shell script in the README, otherwise it probably won't be obvious to people that this convenient thing exists.

However, I'm a little confused as to what this is for. If you have this shell script from this repo, you also have the `docker-compose.yml` files right here. Why are we downloading them? Is this meant to be copy and pasted and run outside of this repo? If so, this feels like something that should be in its own quickstart / docker repo.

docker/fedimintd-mutinynet/docker-compose.yaml
```diff
@@ -18,15 +20,19 @@ services:
...
     image: fedimintui/guardian-ui:latest
     ports:
       - "0.0.0.0:3000:3000"
     environment:
+      - PORT=3000
       - REACT_APP_FM_CONFIG_API=ws://127.0.0.1:8174
     depends_on:
       - fedimintd
     restart: always
+    platform: linux/amd64
```
wbobeirne, line 35: These can be removed now that `version: '3'` is in, but no harm in leaving it either.

docker/gateway-mutinynet/docker-compose.yaml
```diff
@@ -45,16 +62,110 @@ services:
...
+            network: signet
+        EOF
+        tail -f /dev/null
+    volumes:
+      - thunderhub_datadir:/thconfig
+    depends_on:
+      - lnd
+    restart: always
+
+  thunderhub:
```
wbobeirne, line 86: I think we need a `BASE_PATH` config, otherwise you have to just know that you have to visit it at `/thub`: https://docs.thunderhub.io/setup#running-on-different-base-path

Screenshot 2023-07-13 at 3 40 37 PM

wbobeirne (commented):
Had some notes in here, everything more or less worked great for me though so I'm happy to approve if my notes are non-issues.

justinmoon (approved):
Looks good. Will test once it's on master.

docker/download-mutinynet.sh
douglaz: The goal is to put this in the documentation:
`curl -sSL https://raw.githubusercontent.com/fedimint/fedimint/master/docker/download-mutinynet.sh | bash`

Then people will just copy and paste this line.

docker/gateway-mutinynet/docker-compose.yaml
douglaz: good catch. thunderhub is currently broken due to https://github.com/apotdevin/thunderhub/pull/545, so it's hard to test

PR #2760, opened 2023-07-12
fix: sleep a little before htlc reconnection

douglaz:
Sometimes this failure may happen a lot of times flooding the log


douglaz:
At least one test starts to act weird with this change. Will investigate more.

douglaz:
Test issue seems to be gone

douglaz:
Yes, it looks much better. Will close it for now

PR #2765, opened 2023-07-14
fix: add default esplora signet

douglaz:


justinmoon (approved):

justinmoon:
At some point we need to be able to distinguish between mutinynet and the default signet. But since we're doing experimentation with mutinynet right now I think this is a good intermediate improvement.

justinmoon:
@dpc Is there a way to pass the bitcoin rpc configuration to wallet module of fedimint-client? I looked briefly but didn't see anything ...

elsirion:
Not yet, I thought I added a way to propagate the server but probably only discussed the options. Added to #2514.

Issue #2763, opened 2023-07-13
`impl Decodable for Vec<T>` is :sus:


dpc:
Hint: size hints.

dpc:
Maybe not...

PR #2766, opened 2023-07-14
fix: panic on malicious len when decoding `Vec`

dpc:
Re #2763

I'm sure that at some point in the past in the CI I've seen panics due to trying to allocate block of unsanitized length in decoding, but I was never able to figure out what caused it.

This is so far my best guess, but I wasn't able to reproduce the problem for some reason, even thought tracing the code I'm quite sure it's there. stdlib is using non trivial generic trait specializations, so I might have missed something. One way or another, at least in theory I think the current problem is not safe.

### DIFF
--- a/fedimint-core/src/encoding/mod.rs
+++ b/fedimint-core/src/encoding/mod.rs
@@ -14,6 +14,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{Debug, Formatter};
 use std::io::{Error, Read, Write};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::{cmp, mem};
 
 use anyhow::format_err;
 use bitcoin_hashes::hex::{FromHex, ToHex};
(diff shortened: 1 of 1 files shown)


Maan2003:
How is Vec::push() better than preallocating?

dpc:
> How is Vec::push() better than preallocating?

It isn't. Not sure what you mean. :)

Maan2003:
I don't understand how this change is an improvement.

dpc:
> I don't understand how this change is an improvement.

Before:

```rust
        (0..len).map(|_| T::consensus_decode(d, modules)).collect()
```

if the message encoded `len` `0xfff..ff` (or some other large value) would make it attempt to allocate tons of memory and either panic the whole process, or have it swap/slow down/OOM.

Now:

```rust
        (0..cap_len).map(|_| T::consensus_decode(d, modules)).collect()
```

the same message will make it allocate up to a capped size, and attempt to append extra elements one by one, hitting end of a message faster than exhausting memory.

We still need to make sure we enforce some limits on the size of messages, but that usually happens quite naturally, and I'll probably just back-port the same mechanism as rust-bitcoin is using: https://docs.rs/bitcoin/latest/bitcoin/consensus/encode/trait.Decodable.html#method.consensus_decode_from_finite_reader

Maan2003:
Thanks!

I finally understand. It is likely that T::consensus_decode stopping us from allocating more memory.

fedimint-core/src/encoding/mod.rs
```diff
@@ -228,10 +229,45 @@ where
...
         let len = u64::consensus_decode(d, modules)?;
-        (0..len).map(|_| T::consensus_decode(d, modules)).collect()
+
+        // `collect` under the hood uses `FromIter::from_iter`, which can potentially be
+        // backed by code like:
+        // <https://github.com/rust-lang/rust/blob/fe03b46ee4688a99d7155b4f9dcd875b6903952d/library/alloc/src/vec/spec_from_iter_nested.rs#L31>
+        // This can take `size_hint` from input iterator and pre-allocate memory
+        // upfront with `Vec::with_capacity`. Because of that untrusted `len`
+        // should not be used directly.
+        let cap_len = cmp::min(8_000, len);
```
Maan2003, line 239: limit should also depend on `mem::size_of::<T>()`

dpc:
> I finally understand. It is likely that T::consensus_decode stopping us from allocating more memory.

Now I'm confused. :D

fedimint-core/src/encoding/mod.rs
dpc: Right.

Maan2003 (approved):

Maan2003:
> Now I'm confused. :D

edited to make it more clear

dpc:
Ah. Yes. The fact that the input can't possibly go on into gigabytes is what guarantees termination before using too much memory.

elsirion:
Should we also limit what we can encode to catch errors early if we ever hit that limit with a legitimate use case?

Issue #2768, opened 2023-07-14
Publish linux binaries


justinmoon:
https://github.com/fedimint/fedimint/issues/1008 added support for building binaries which can run on any linux distribution. As we move closer to a release, we should add a CI job to publish these as a GitHub release.

dpc:
Need to verify if https://github.com/fedimint/fedimint/pull/2885 did it.

elsirion:
Technically they are being published :stuck_out_tongue_winking_eye:, but there [seems to be something wrong](https://github.com/fedimint/fedimint/pull/2885#issuecomment-1664083922), would be good to reproduce on a non-NixOS machine.

justinmoon:
Call: maybe updating nixpkgs fixed?

dpc:
I tried `fedimintd` and `fedimint-cli` on Debian in a VM now and both worked (tried `--help`).

PR #2771, opened 2023-07-14
chore: bundle linux binaries

EthnTuttle:


EthnTuttle:
#2768
Just waiting to see what happens when it runs

.github/workflows/ci-nix.yml
```diff
@@ -332,3 +330,39 @@ jobs:
         webhook: ${{ secrets.DISCORD_WEBHOOK }}
         # current job is a success, but that's not what we're interested in
         status: failure
+
+  deb-binary:
```
dpc, line 334: We should probably do this in `Release packages` workflow, as both will be building release builds of stuff, so possibly can re-use most of the work.

EthnTuttle: I've added the `Release packages` to the artifact upload.
Is that what you meant?

.github/workflows/ci-nix.yml
```diff
@@ -290,7 +288,7 @@ jobs:
           - pkg: gateway-pkgs
             bin: gateway-cli
 
-    runs-on: buildjet-4vcpu-ubuntu-2004
+    runs-on: ubuntu-20.04
```
EthnTuttle, line 291: I will change this before marking this PR ready, but I don't have this image in my repo's CI

.github/workflows/ci-nix.yml
dpc: Nope.

There is already a job "Release packages" https://github.com/fedimint/fedimint/blob/8864e239e539c716b4f970397f508598dbb1c383/.github/workflows/ci-nix.yml#L281 and I think under the hood is already has to build the same thing (release versions of packages). Bundling into deb/rpm is just an extra step.

So by doing all the stuff here in that job, we could avoid doing the same thing twice.

.github/workflows/ci-nix.yml
```diff
@@ -332,3 +336,39 @@ jobs:
...
         # current job is a success, but that's not what we're interested in
         status: failure
+
+  deb-binary:
+    name: "Bundling for DEB and RPM"
+    runs-on: ubuntu-20.04 
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+        if: github.ref == 'refs/heads/master'
```
elsirion, line 346: The whole job should be conditional. Making single steps conditional makes later steps fail right now and adding the same if condition to all of them to fix that seems suboptimal.

dpc:
Took a stab at wrapping it up in:  #2859

justinmoon:
https://github.com/fedimint/fedimint/pull/2859 is merged, closing this

PR #2776, opened 2023-07-17
fix: lightning_gw_reconnect_test not flaky anymore

douglaz:
It may take a while for the gateway to be able to connect to the ln node.

It seems that previously the test would consider that if `gateway-cli info` returns successful then it means that this reconnection have already happened. But this isn't the case.

So we try to pay the invoice at most 10 times, waiting 1 second between each retry, and only fail after that.


m1sterc001guy (approved):
Ah good catch, thanks for finding this

PR #2782, opened 2023-07-19
feat: wait for leader to signal ready for DKG

okjodom:
Followers wait for guardian server to get to `ReadyForConfigGen`, thus signalling all guardians are ready for DKG

depends on #2790

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -2,6 +2,7 @@ use std::collections::BTreeMap;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::sync::{Mutex, MutexGuard};
+use std::time::Duration;
 
 use async_trait::async_trait;
 use bitcoin_hashes::sha256::HashEngine;
@@ -21,7 +22,7 @@ use fedimint_core::encoding::Encodable;
 use fedimint_core::module::{
     api_endpoint, ApiAuth, ApiEndpoint, ApiEndpointContext, ApiError, ApiRequestErased,
 };
-use fedimint_core::task::TaskGroup;
+use fedimint_core::task::{sleep, TaskGroup};
 use fedimint_core::util::write_new;
 use fedimint_core::PeerId;
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


justinmoon (approved):
LGTM. Tested this in UI yesterday and it fixed https://github.com/fedimint/ui/issues/83.

justinmoon (approved):

fedimint-server/src/config/api.rs
```diff
@@ -185,21 +186,46 @@ impl ConfigGenApi {
...
         };
+
         self.update_leader().await?;
 
+        // Followers wait for leader to signal readiness for DKG
+        if let Some(client) = leader {
+            loop {
+                let status = client.status().await.map_err(|_| {
+                    ApiError::not_found("Unable to connect to the leader".to_string())
+                })?;
```
elsirion, line 208: This should probably become more robust in the future, so a short internet disconnection doesn't kill the whole setup process.

fedimint-server/src/config/api.rs
```diff
@@ -185,21 +186,46 @@ impl ConfigGenApi {
...
+        let leader = {
             let mut state = self.require_status(ServerStatus::SharingConfigGenParams)?;
+            // Update our state
             state.status = ServerStatus::ReadyForConfigGen;
-            (
-                state.get_config_gen_params(&request, response.consensus)?,
-                state.settings.registry.clone(),
-            )
+            // Create a WSClient for the leader
+            state.local.clone().and_then(|local| {
```
elsirion, line 194: Afaik `state.local` can never be none in this state, should we `expect` here instead of `and_then`? Or under which conditions can it be `None`?

fedimint-server/src/config/api.rs
```diff
@@ -185,21 +186,46 @@ impl ConfigGenApi {
...
+                sleep(Duration::from_millis(100)).await;
+            }
+        };
+
+        // Get params and registry
+        let request = self.get_requested_params()?;
+        let response = self.get_consensus_config_gen_params(&request).await?;
+        let (params, registry) = {
+            let state: MutexGuard<'_, ConfigGenState> =
+                self.require_status(ServerStatus::ReadyForConfigGen)?;
```
elsirion, line 221: We set that state in L192, so could this be an assertion/panic instead (error=can reasonably occur due to faulty input, panic=broken invariant)? I.e. can we ever go backwards from `ReadyForConfigGen`?

PR #2790, opened 2023-07-20
Safe Auth

okjodom:
Safer auth in guardian webserver by explicitly requiring password on `WsClientAdmin` methods. This prevents accidental password leak when making api calls between guardians

### DIFF
--- a/fedimint-core/src/admin_client.rs
+++ b/fedimint-core/src/admin_client.rs
@@ -22,22 +22,20 @@ use crate::PeerId;
 // TODO: Maybe should have it's own CLI client so it doesn't need to be in core
 pub struct WsAdminClient {
     inner: DynGlobalApi,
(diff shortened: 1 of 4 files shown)


douglaz (approved):
Certainly looks better

fedimint-core/src/admin_client.rs
```diff
@@ -186,8 +186,7 @@ impl WsAdminClient {
 
     /// Returns the status of the server
     pub async fn status(&self) -> FederationResult<StatusResponse> {
-        self.request_auth("status", ApiRequestErased::default())
-            .await
+        self.request("status", ApiRequestErased::default()).await
```
elsirion, line 189: Good catch!

elsirion (approved):
Looks like an overall improvement.

PR #2793, opened 2023-07-20
Use cookie file for auth when username does not exist

dpc:
This is a rework of #2612

When `FM_BITCOIND_COOKIE_FILE` env is set, use it as path to
the cookie file.

AFAIK, auth is not required for JsonRPC, so we I don't think we
can do any automatic cookie file detection/fallback if auth
information is missing.

### DIFF
--- a/fedimint-bitcoind/src/bitcoincore.rs
+++ b/fedimint-bitcoind/src/bitcoincore.rs
@@ -1,9 +1,12 @@
+use std::env;
 use std::io::Cursor;
+use std::path::PathBuf;
 
-use anyhow::anyhow as format_err;
+use anyhow::{anyhow as format_err, bail};
 use bitcoin::{BlockHash, Network, Script, Transaction, Txid};
 use bitcoincore_rpc::bitcoincore_rpc_json::EstimateMode;
 use bitcoincore_rpc::{Auth, RpcApi};
+use fedimint_core::bitcoinrpc::FM_BITCOIND_COOKIE_FILE_VAR_NAME;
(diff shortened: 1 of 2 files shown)


fedimint-bitcoind/src/bitcoincore.rs
```diff
@@ -125,7 +144,11 @@ pub fn from_url_to_url_auth(url: &Url) -> anyhow::Result<(String, Auth)> {
             )
         }),
         if url.username().is_empty() {
-            Auth::None
+            let auth = match read_cookie_file() {
+                Ok(cookie) => Some(Auth::UserPass(url.username().to_owned(), cookie)),
+                Err(_) => None,
+            };
+            auth.unwrap_or(Auth::None)
```
elsirion, line 151: 1. This implementation seems faulty. The cookie file can contain both a user name and a password. [At leas `bitcoincore-rpc` implements it that way.](https://docs.rs/bitcoincore-rpc/0.17.0/src/bitcoincore_rpc/client.rs.html#203-219)
2. Why not just use the `bitcoincore-rpc` implementation in the first place?

dpc: Oh. I didn't know about `Auth::Cookie`.

dpc:
The whole thing feels rather clumsy. But I have the same feeling about cookie files in the first place, so :shrug: .

elsirion (approved):

douglaz (approved):

elsirion:
@dpc might need to rebase to get rid of some of the CI flakyness

PR #2797, opened 2023-07-20
fix: dyn-types encoding incorrect

dpc:


fedimint-core/src/core.rs
```diff
@@ -107,13 +107,7 @@ macro_rules! module_dyn_newtype_impl_encode_decode {
...
                 assert_eq!(buf.len(), buf_written);
 
-                let buf_len = u64::try_from(buf.len())
-                    .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
-
-                written += buf_len.consensus_encode(writer)?;
-
-                writer.write_all(buf.as_slice())?;
-                written += buf.len();
+                written += buf.consensus_encode(writer)?;
```
dpc, line 110: This is just a simplification. `Vec` is encoded `len, data`, so exactly the same.

fedimint-core/src/encoding/mod.rs
```diff
@@ -787,7 +787,17 @@ where
             fedimint_core::core::ModuleInstanceId::consensus_decode(reader, decoders)?;
         Ok(match decoders.get(module_instance_id) {
             Some(decoder) => {
-                DynRawFallback::Decoded(decoder.decode(reader, module_instance_id, decoders)?)
```
dpc, line 790: This was the mistake. `decoder.decode` does not encode `module_instance_id` and `len` and apparently this path was not covered. But now it is, after a change landed in parallel.

m1sterc001guy (approved):
LGTM

PR #2802, opened 2023-07-20
fix: LND get info before intercepting HTLCs

m1sterc001guy:
PR contains small fixes for the lightning reconnect logic:

1) We no longer re-try when the gateway fails to be created. This is because the gateway creation should always succeed, if it doesn't, it likely indicates that something is wrong with the db, which will not be fixed by retrying. The retry logic has been replaced with an `expect`.

2) For LND, we try to call `get_info` before attempting to intercept HTLCs. This is because there is a small period of time where the client can connect to LND, but RPCs aren't successful. During this time, we don't want to spawn the new thread to intercept HTLCs. So instead we first wait for `get_info` to succeed, which will increase the probability of `route_htlcs` succeeding.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -257,28 +257,19 @@ impl Gateway {
                                     scid_to_federation.clone(),
                                     tg.clone(),
                                 )
-                                .await;
+                                .await.expect("Failed to created Gateway");
 
-                                match gateway {
-                                    Ok(gateway) => {
-                                        info!("Successfully created Gateway");
+                                info!("Successfully created Gateway");
 
-                                        let tx = run_webserver(password.clone(), listen, gateway)
-                                            .await
-                                            .expect("Failed to start webserver");
-                                        info!("Successfully started webserver");
+                                let tx = run_webserver(password.clone(), listen, gateway)
+                                    .await
+                                    .expect("Failed to start webserver");
+                                info!("Successfully started webserver");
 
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -257,28 +257,19 @@ impl Gateway {
...
+
+                                info!("Successfully created Gateway");
+
+                                let tx = run_webserver(password.clone(), listen, gateway)
+                                    .await
+                                    .expect("Failed to start webserver");
+                                info!("Successfully started webserver");
+
+                                Self::handle_htlc_stream(stream, ln_client, handle.clone(), scid_to_federation.clone(), clients.clone()).await;
+                                warn!("HTLC Stream Lightning connection broken. Stopping webserver...");
```
justinmoon, line 270: This log is a little confusing. We're shutting down the whole gateway at this point, right?

m1sterc001guy: We shut down the webserver and drop the `Gateway` struct, but `Gatewayd` doesn't stop. Maybe I should change the message to be a little clearer?

justinmoon: I guess this is in a loop so webserver should get restarted again?

justinmoon (approved):

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Once it successfully connects to the lightning node, yes the webserver will come back up.

elsirion:
Could this have been the cause of some of our test flakiness?

m1sterc001guy:
Maybe, not sure yet. I mainly ran into this when debugging another issue.

PR #2808, opened 2023-07-20
fix: allow get_verify_config_hash after verification is done

okjodom:
After a guardian has already verified configs from other peers and proceeded to `VerifiedConfigs` state added in #2807, they should still have the ability to fetch and share their own config hash. Helps fix https://github.com/fedimint/ui/issues/14.

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -261,7 +261,17 @@ impl ConfigGenApi {
     /// Returns the consensus config hash, tweaked by our TLS cert, to be shared
     /// with other peers
     pub fn get_verify_config_hash(&self) -> ApiResult<BTreeMap<PeerId, sha256::Hash>> {
-        let state = self.require_status(ServerStatus::VerifyingConfigs)?;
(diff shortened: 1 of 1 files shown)


fedimint-server/src/config/api.rs
```diff
@@ -261,7 +261,17 @@ impl ConfigGenApi {
     /// Returns the consensus config hash, tweaked by our TLS cert, to be shared
     /// with other peers
     pub fn get_verify_config_hash(&self) -> ApiResult<BTreeMap<PeerId, sha256::Hash>> {
-        let state = self.require_status(ServerStatus::VerifyingConfigs)?;
+        let state = self
+            .require_status(ServerStatus::VerifyingConfigs)
```
justinmoon, line 265: Might be good to have a `require_states(Vec)` in the future

justinmoon (approved):

fedimint-server/src/config/api.rs
okjodom: makes sense

PR #2810, opened 2023-07-21
refactor: move user messages to end of script

EthnTuttle:
When running this script, if both fedimintd and gatewayd are selected, the message about connecting to the fedimintd dashboard gets lost in the terminal. This ensures all messages for connecting are echoed at the end


docker/download-mutinynet.sh
```diff
@@ -59,17 +59,23 @@ if [[ ${fedimintd_install[*]} =~ ^[Yy]?$ ]]; then
...
   echo "Optionally run 'docker-compose -f $GATEWAY_DIR/docker-compose.yaml logs -f' to see the logs"
+  echo
+fi
+
+if [[ ${fedimintd_install[*]} =~ ^[Yy]?$ ]]; then
+  echo "You can access the fedimintd dashboard at http://$external_ip:3000"
+  echo "Note: by default you should open ports 8173 and 8174 for external access on your router/firewall, plus the ports mentioned above"
+fi
+
+if [[ ${gateway_install[*]} =~ ^[Yy]?$ ]]; then
```
elsirion, line 77: Since this variable is compared repeatedly, maybe we should set a variable that's easier to test, although the lack of support for boolean variable in bash could reduce the benefit.

elsirion (approved):

PR #2813, opened 2023-07-21
Make bitcoin blockchain API used by the client configurable

elsirion:
This should allow using custom signets/regtests networks and setting alternative defaults for mainnet.

If we want to expose the parameter in the UI I can open an issue there.

### DIFF
--- a/fedimintd/src/lib.rs
+++ b/fedimintd/src/lib.rs
@@ -17,6 +17,7 @@ use fedimint_wallet_server::common::config::{
     WalletGenParams, WalletGenParamsConsensus, WalletGenParamsLocal,
 };
 use fedimint_wallet_server::WalletGen;
+use url::Url;
 
 /// Module for creating `fedimintd` binary with custom modules
 pub mod fedimintd;
@@ -42,6 +43,7 @@ pub fn attach_default_module_gen_params(
                     // TODO this is not very elegant, but I'm planning to get rid of it in a next
                     // commit anyway
                     finality_delay,
+                    client_default_bitcoin_rpc: default_esplora_server(network),
                 },
             },
         )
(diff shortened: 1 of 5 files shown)


fedimintd/src/lib.rs
```diff
@@ -67,3 +69,21 @@ pub fn attach_default_module_gen_params(
...
+pub fn default_esplora_server(network: Network) -> BitcoinRpcConfig {
+    let url = match network {
+        Network::Bitcoin => Url::parse("https://blockstream.info/api/")
+            .expect("Failed to parse default esplora server"),
+        Network::Testnet => Url::parse("https://blockstream.info/testnet/api/")
+            .expect("Failed to parse default esplora server"),
+        Network::Regtest => {
+            Url::parse("http://127.0.0.1:50002/").expect("Failed to parse default esplora server")
+        }
+        Network::Signet => Url::parse("https://mutinynet.com/api/")
```
m1sterc001guy, line 82: Just curious, is our plan to mainly only support mutiny net (as opposed to the non-forked "other" signet) ?

m1sterc001guy (approved):

fedimintd/src/lib.rs
elsirion: No, but it's also just a default for the config gen parameter, i.e. this should be changeable via the setup UI eventually and is already changeable in the consensus config after this PR (with some annoying manual steps).

justinmoon: I wish rust-bitcoin had a way to specify a custom signet here

elsirion: I thought about opening a PR, but iirc there wouldn't be a good way to inject all the network-specific constants, so this would mean a major API and functionality change :/

PR #2821, opened 2023-07-22
fix(fedimint-cli): clean shutdown

elsirion:
Previously tokio would complain that it's awaiting futures while the runtime is shut down.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -71,6 +71,7 @@ use std::collections::{BTreeMap, HashSet};
 use std::fmt::{Debug, Formatter};
 use std::io::{Error, Read, Write};
 use std::sync::Arc;
+use std::time::Duration;
 
 use anyhow::{anyhow, bail};
 use async_stream::stream;
@@ -104,7 +105,8 @@ use rand::thread_rng;
 use secp256k1_zkp::{PublicKey, Secp256k1};
 use secret::DeriveableSecretClientExt;
 use serde::Serialize;
-use tracing::{debug, info, warn};
+use tokio::sync::Mutex;
+use tracing::{debug, error, info, warn};
 
 use crate::backup::Metadata;
 use crate::db::ClientSecretKey;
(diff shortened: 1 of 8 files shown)


m1sterc001guy (approved):

elsirion:
New attempt that should not have the problem that the client gets dropped before the task group is shut down, which also upsets it.

m1sterc001guy (approved):
Tested using tmuxinator and generating and invoice using `ln-invoice`. Works for me

fedimint-client/src/lib.rs
```diff
@@ -447,11 +452,47 @@ fn states_add_instance(
...
+        // Not sure if Ordering::SeqCst is strictly needed here, but better safe than
+        // sorry.
+        let client_count = self
+            .inner
+            .client_count
+            .fetch_sub(1, std::sync::atomic::Ordering::SeqCst);
+
+        // `fetch_sub` returns previous value, so if it is 1, it means this is the last
+        // client reference
+        if client_count == 1 {
```
dpc, line 481: Why not do it in `ClientInner::drop` then?

fedimint-client/src/lib.rs
```diff
@@ -860,6 +903,17 @@ struct ClientInner {
     root_secret: DerivableSecret,
     operation_log: OperationLog,
     secp_ctx: Secp256k1<secp256k1_zkp::All>,
+    /// Task group used by this client and shut down when the client is dropped.
+    /// If a subgroup is used, dropping the parent task group stops the client's
+    /// background tasks.
+    tg: Mutex<Option<TaskGroup>>,
+    /// Number of [`Client`] instances using this `ClientInner`.
+    ///
+    /// The `ClientInner` struct is both used for the client itself as well as
```
dpc, line 912: Hmmmm... OK

dpc (approved):

fedimint-client/src/lib.rs
elsirion: I also tried to make all the context uses of `ClientInner` weak, but that was very inelegant (how to propagate up the error case when upgrading?) and lead to weird errors that I gave up debugging since I had wasted too much time on this already.

PR #2822, opened 2023-07-22
chore: remove unused peer id argument from admin client constructor

### LINKED ISSUE #2801: WsAdminClient constructor peer id
The constructor for `WsAdminClient` takes a `PeerId` variable confusingly named `our_id` which is always hard-coded to `PeerId(0)` AFAICT. 

There is also [this note](https://github.com/fedimint/fedimint/blob/0c02f9885a11ff9334e9ea47aeb14f2d3c09a599/fedimint-server/src/config/api.rs#L107-L108) that they aren't used and should maybe removed. I would be in favor of removing because they're confusing and not used.

image

elsirion:
Fixes #2801

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -272,7 +272,7 @@ impl Opts {
             .expect("Endpoint exists")
             .url
             .clone();
-        Ok(WsAdminClient::new(url, *our_id))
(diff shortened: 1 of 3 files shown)


justinmoon (approved):

justinmoon:
I think https://github.com/fedimint/fedimint/pull/2822 is the same as this

PR #2823, opened 2023-07-22
feat: add consensus gen auth endpoint for validating password

okjodom:
replicates #2820 for the consensus api

### DIFF
--- a/fedimint-server/src/lib.rs
+++ b/fedimint-server/src/lib.rs
@@ -293,3 +293,13 @@ impl FedimintApiHandler {
         self.handle.stopped().await;
     }
 }
+
+pub type ApiResult<T> = std::result::Result<T, ApiError>;
+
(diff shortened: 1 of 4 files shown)


fedimint-server/src/lib.rs
```diff
@@ -293,3 +293,13 @@ impl FedimintApiHandler {
         self.handle.stopped().await;
     }
 }
+
+pub type ApiResult<T> = std::result::Result<T, ApiError>;
+
+pub fn check_auth(context: &mut ApiEndpointContext) -> ApiResult<()> {
```
elsirion, line 299: I wonder if this could be a method of `ApiEndpointContext`.

elsirion (approved):

fedimint-server/src/lib.rs
EthnTuttle: Or would it fit as an additional macro? `api_endpoint!` and then `api_endpoint_with_auth!`

PR #2829, opened 2023-07-24
refactor: Add Gatewayd struct with ClientModuleGenRegsitry

m1sterc001guy:
Adds a new struct named `Gatewayd` which is responsible for the main "blocking" loop of the daemon. This loop creates a `Gateway` instance and spawns the webserver.

Fixes: https://github.com/fedimint/fedimint/issues/2650

This refactor aligns the gateway with how `Fedimintd` boots as well (attaching modules to a struct), which should make it easier to modularize in the future (if that's the route we want to go).

Added benefit is that the `main` for Gatewayd is really clean now:

image


gateway/ln-gateway/src/lib.rs
```diff
@@ -84,6 +93,221 @@ pub const DEFAULT_FEES: RoutingFees = RoutingFees {
...
+        );
+
+        let mut tg = TaskGroup::new();
+        let rx = self.start_gateway(&mut tg, client_builder, db).await?;
+        rx.await?;
+        info!("Gatewayd exiting...");
+        Ok(())
+    }
+
+    pub async fn start_gateway(
```
elsirion, line 216: Shouldn't this be private since without it the modules defined in `Gatewayd` will be discarded, which is counterintuitive?

elsirion (approved):
Looks good

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Ah yes, this was overlooked when copy/pasted. It should be private

gateway/ln-gateway/src/lib.rs
```diff
@@ -84,6 +93,221 @@ pub const DEFAULT_FEES: RoutingFees = RoutingFees {
...
+    pub fn with_module<T>(mut self, gen: T) -> Self
+    where
+        T: ClientModuleGen + 'static + Send + Sync,
+    {
+        self.registry.attach(gen);
+        self
+    }
+
+    pub fn with_default_modules(self) -> Self {
+        // Gateway module will be attached when the federation clients are created.
```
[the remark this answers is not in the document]
m1sterc001guy: Correct. Will update comment

elsirion (approved):

Issue #2627, opened 2023-06-07
Gateway doesn't seem to check invoice expiry before trying to pay


justinmoon:
I just saw this log message in a running federation:

```
Jun 07 12:20:10 env[283225]:     LND error: Status { code: Unknown, message: "invoice expired. Valid until 2023-05-15 11:47:23 +0000 UTC", metadata: MetadataMap { headers: {"content-type": "application/grpc"} },>
```

Perhaps the gateway should check invoice expiry before attempting to contact the lightning node?

PR #2830, opened 2023-07-24
fix: Check invoice expiry before paying

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2627

### DIFF
--- a/gateway/ln-gateway/src/ng/pay.rs
+++ b/gateway/ln-gateway/src/ng/pay.rs
@@ -1,4 +1,5 @@
 use std::sync::Arc;
+use std::time::Duration;
 
 use fedimint_client::sm::{ClientSMDatabaseTransaction, OperationId, State, StateTransition};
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/ng/pay.rs
```diff
@@ -319,6 +322,10 @@ impl GatewayPayInvoice {
             return Err(OutgoingContractError::TimeoutTooClose);
         }
 
+        if invoice.is_expired() {
+            return Err(OutgoingContractError::InvoiceExpired(invoice.expiry_time()));
+        }
```
elsirion, line 327: We might want the same on the client side. Maybe not that fn since it probably uses `SystemTime::now` which isn't WASM compatible, but the earlier we catch the problem the less hassle there is with refunding etc.

elsirion (approved):

PR #2831, opened 2023-07-24
chore: minor improvements on scripts ux

douglaz:


docker/tls-download-mutinynet.sh
```diff
@@ -40,11 +32,21 @@ if ! [ -x "$(command -v docker-compose)" ]; then
   fi
 fi
 
+COMMANDS="awk curl sed tr wc jq"
```
douglaz, line 35: if we install docker compose first it should install most of these commands

justinmoon (approved):

docker/download-mutinynet.sh
```diff
@@ -3,15 +3,37 @@
 # This file downloads the mutinynet docker-compose files for the LN gateway and fedimintd
 # You can download and run it with: curl -sSL https://raw.githubusercontent.com/fedimint/fedimint/master/docker/download-mutinynet.sh | bash
 
+if ! [ -x "$(command -v docker-compose)" ]; then
+  # check if we are running as root
+  if [ "$EUID" -ne 0 ]; then
+    echo 'Error: docker-compose is not installed and we can not install it for you.' >&2
+    exit 1
+  fi
```
elsirion, line 11: Should we encourage running our scripts a root? I'd much rather just print the command that is likely to work on the machine but leave it to the user to execute it as root.

douglaz: Good point. But I don't think it's a big encouragement. It isn't like we are saying "please try again running this script as root so it can install the dependencies"

PR #2832, opened 2023-07-24
chore: update 'check-wasm' command to check fedimint-client

justinmoon:
Locally I get the following error, @douglaz said he gets the same on linux:

```
error[E0432]: unresolved import `super::sysrand_chunk`
   --> /Users/justin/.cargo/git/checkouts/ring-236810bb234b0c6b/52a1294/src/rand.rs:307:16
    |
307 |     use super::sysrand_chunk::chunk;
    |                ^^^^^^^^^^^^^ could not find `sysrand_chunk` in `super`
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1289,6 +1289,7 @@ dependencies = [
  "futures",
  "itertools",
  "rand",
+ "ring",
  "secp256k1-zkp",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


justinmoon:
Pushed another commit that makes `fedimint-client` use a WASM-compatible `ring` (like `fedimint-client-legacy` does) and now `just check-wasm` works for me.

douglaz (approved):
LGTM

elsirion (approved):

PR #2833, opened 2023-07-24
chore: only install firefox on linux

justinmoon:


elsirion (approved):
Maybe we should even split off the WASM integration test env into a separate shell. I'm also not too keen to install some random version of firefox just because I want to test wasm compilation :laughing:

PR #2834, opened 2023-07-24
test: will ci fail?

justinmoon:
I am skeptical that the wasm tests are actually running in CI (https://github.com/fedimint/fedimint/actions/runs/5650404463/job/15306686903?pr=2833#step:10:1). 

This PR tests if they are.


justinmoon:
CI failed 🎉

PR #2835, opened 2023-07-25
chore: ln-gateway integration balance test

zacheryasc:


gateway/ln-gateway/tests/fixtures.rs
```diff
@@ -24,6 +27,10 @@ pub async fn fixtures() -> (
     let ln_params = LightningGenParams::regtest(fixtures.bitcoin_server());
     fixtures = fixtures.with_module(LightningClientGen, LightningGen, ln_params);
 
+    let wallet_params = WalletGenParams::regtest(fixtures.bitcoin_server());
+    let wallet_client = WalletClientGen::new(fixtures.bitcoin_client());
+    fixtures = fixtures.with_module(wallet_client, WalletGen, wallet_params);
```
m1sterc001guy, line 32: It's probably a bit confusingly named, but you don't need the Wallet module here. The wallet module handles peg-ins and peg-outs for on-chain bitcoin. The test you're implementing is just validating if the `info` command returns the balance of the federation (which in the case of tests, is the dummy module. In a real Fedimint it will be the Mint module).

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -60,10 +61,60 @@ async fn gatewayd_shows_info_about_all_connected_federations() {
...
+
+    let adr2 = rpc
+        .get_deposit_address(DepositAddressPayload { federation_id: id2 })
+        .await
+        .unwrap();
+
+    bitcoin.prepare_funding_wallet().await;
+    bitcoin
+        .send_and_mine_block(&adr1, bitcoin::Amount::from_sat(5_000_000))
+        .await;
```
m1sterc001guy, line 101: As mentioned above, you don't need to do this. Just print the money using the dummy module then verify that `info` returns the expected value.

zacheryasc: I was using the fed client directly to do this, and still had the issue. I just pushed two commits to make it explicit how the test fails, appreciate a look.

m1sterc001guy:
Test looks good to me, I just restarted CI. Let me know if you're still hitting issues.

m1sterc001guy (dismissed):

m1sterc001guy (approved):

okjodom (approved):

PR #2844, opened 2023-07-26
fix(consensus): processing same tx submitted multiple times 

### LINKED ISSUE #2842: Recovery without backup not working
If I comment out [this line](https://github.com/fedimint/fedimint/blob/0403bdbacda41ef6e372f4661f8fab6c4d151400/devimint/src/main.rs#L768) and run `./scripts/cli-test.sh` I get the following:

```
2023-07-25T21:47:59.845015Z  INFO devimint: Testing backup&restore
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `420000`,
 right: `0`', devimint/src/main.rs:784:9
```

dpc:
See commit messages.

Fix #2842

### DIFF
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -137,7 +137,9 @@ impl FedimintConsensus {
 
                     Box::pin(async move {
                         let mut rejected_txs = BTreeSet::new();
-                        let mut accepted_txes = BTreeSet::new();
+                        // Since multiple peers can submit the same tx within epoch, track them
+                        // and handle only once.
+                        let mut processed_txes = BTreeSet::new();
 
                         let items = consensus_outcome.clone()
                             .contributions
@@ -150,16 +152,13 @@ impl FedimintConsensus {
                                 .expect("Setting transaction savepoint failed");
 
                             if let  ConsensusItem::Transaction(ref transaction) = consensus_item {
-                                if accepted_txes.contains(&transaction.tx_hash()) {
+                                if !processed_txes.insert(transaction.tx_hash()) {
                                     continue;
                                 }
                             }
 
                             let decision = match self.process_consensus_item(dbtx, consensus_item.clone(), peer_id).await {
                                 Ok(decision) => {
-                                    if let  ConsensusItem::Transaction(ref transaction) = consensus_item {
-                                        accepted_txes.insert(transaction.tx_hash());
-                                    }
                                     decision
                                 },
                                 Err(error) => {
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -137,6 +137,7 @@ impl FedimintConsensus {
 
                     Box::pin(async move {
                         let mut rejected_txs = BTreeSet::new();
+                        let mut accepted_txes = BTreeSet::new();
 
                         let items = consensus_outcome.clone()
                             .contributions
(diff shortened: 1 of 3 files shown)


joschisan:
Okay first problem I see is that a transactions is added to the accepted transactions even if the consensus decision for it equals discard. Just because the result from process_consensus_item is Ok does not mean the item has been accepted. 

However, the more fundamental problem is this. Filtering out redundant transactions within an epoch will not be possible anymore when there are no epochs anymore. It actually used to filter out redundant transactions just as you proposed and I took it out in the last pr to prepare for Aleph Bft. When we took it out it broke the test ecash_can_be_recovered in the integration_tests, thats why we took out the test. Kitman wanted to take care of it when porting the tests to the new framework.

So we can now simply but it back in but have to fix the recovery to work without deduplicated transactions.

fedimint-server/src/consensus/mod.rs
```diff
@@ -148,14 +149,26 @@ impl FedimintConsensus {
                                 .await
                                 .expect("Setting transaction savepoint failed");
 
+                            if let  ConsensusItem::Transaction(ref transaction) = consensus_item {
+                                if accepted_txes.contains(&transaction.tx_hash()) {
+                                    continue;
+                                }
+                            }
+
```
elsirion, line 157: Just skipping the transaction without making a decision will be a bug once we migrate to AlephBFT (correct me if I'm wrong @joschisan). So ideally we'd instead just make the insert operation into `rejected_txs` dependent on if the rejection happened due to invalidity or double-inlusion. The concept of rejected transactions will disappear with AlephBFT anyway since only accepted ones will be persisted in the on-disk log.

https://github.com/fedimint/fedimint/blob/6603c8bc2c6e9e616d89f26984af6560b4e75c31/fedimint-server/src/consensus/mod.rs#L177-L179

elsirion (changes requested):
I think this fix is in conflict with @joschisan's ongoing AlephBFT project. I outlined a possible solution. We should also introduce a test so that such regressions get caught next time.

fedimint-server/src/consensus/mod.rs
```diff
@@ -148,14 +149,26 @@ impl FedimintConsensus {
...
                             let decision = match self.process_consensus_item(dbtx, consensus_item.clone(), peer_id).await {
-                                Ok(decision) => decision,
+                                Ok(decision) => {
+                                    if let  ConsensusItem::Transaction(ref transaction) = consensus_item {
+                                        accepted_txes.insert(transaction.tx_hash());
+                                    }
```
elsirion, line 162: Only add transactions that were actually accepted.

elsirion:
I talked to @joschisan, we can basically merge this as-is, with the exception that only actually accepted transactions should be inserted into the accepted set (that could lead to bugs in certain edge cases rn).

My concerns around not `continue`ing are unfounded since @joschisan will remove the entire concept of rejected transactions in the AlephBFT PR. It won't be needed anymore because the consensus log of AlephBFT only contains accepted transactions, so the client won't have to filter anymore.

dpc:
@elsirion @joschisan If I rename `accepted_txs` to `processed_txs`, what are the corner cases I'm missing? I just want to process each tx once.

elsirion (approved):
I think this works as well and is simpler.

justinmoon:
Call: we now don't have any test coverage of recovery with backup. Would be good to have coverage there.

Issue #2827, opened 2023-07-24
LND can't pay 1 millisat invoice in mprocs


justinmoon:
```
# switch to using the LND gateway:
$ fedimint-cli switch-gateway  "0385d6b533d9e8175a12180b42f56220e13ba4714f750181df45e9edfd332b7c1d"
2023-07-24T02:45:58.587858Z  INFO fedimint_client::sm::executor: Starting state machine executor task
{
  "active": true,
  "api": "http://127.0.0.1:28175/",  # LND gateway runs on 28175
  ...
}

# generate invoice with CLN (pbcopy just copies to clipboard)
$ lightning-cli invoice 1 sdfdsdf sdfsdf | jq -r '.bolt11' | pbcopy

# paying fails
$ fedimint-cli ln-pay <invoice>

2023-07-24T02:46:13.817959Z  INFO fedimint_client::sm::executor: Starting state machine executor task
2023-07-24T02:46:13.833949Z  INFO fedimint_cli::client: Update: Created
2023-07-24T02:46:14.904253Z  INFO fedimint_cli::client: Update: Funded
2023-07-24T02:46:16.477583Z  INFO fedimint_cli::client: Update: WaitingForRefund { block_height: 763, gateway_error: GatewayInternalError { error_code: Some(500), error_message: "500 Internal Server Error" } }
2023-07-24T02:46:18.575915Z  INFO fedimint_cli::client: Lightning Gateway failed to pay invoice. ErrorCode: Some(500) ErrorMessage: 500 Internal Server Error
```

<details>

<summary>CLN gateway shows the following:</summary>

```
2023-07-24T02:44:51.374927Z ERROR pay_invoice: ln_gateway::rpc::rpc_server: error=Other: Outgoing contract canceled

Stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/justin/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.67/src/backtrace/libunwind.rs:93:5
      backtrace::backtrace::trace_unsynchronized
             at /Users/justin/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.67/src/backtrace/mod.rs:66:5
   1: backtrace::backtrace::trace
             at /Users/justin/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.67/src/backtrace/mod.rs:53:14
   2: anyhow::backtrace::capture::Backtrace::create
[...]

PR #2846, opened 2023-07-26
chore: improve gatewayd error messages

m1sterc001guy:
Error messages in the gateway are not structured correctly right now, we frequently swallow the underlying error and just output a vague string that something went wrong. This makes debugging hard.

This PR improves the error handling logic to attach errors in different states in the gateway. Fixes: https://github.com/fedimint/fedimint/issues/2827

CLI output after PR:
image

Gateway output after PR:
image

Gateway output contains the underlying error message now (e.g `FailureReasonNoRoute`). Client is returned a user friendly message that gives a bit of a description on what went wrong.

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -202,7 +202,9 @@ pub async fn handle_ng_command(
                 info!("Update: {:?}", update);
             }
 
-            return Err(anyhow::anyhow!("Lightning receive failed"));
+            return Err(anyhow::anyhow!(
+                "Unexpected end of update stream. Lightning receive failed"
+            ));
         }
         ClientCmd::LnPay { bolt11 } => {
             client.select_active_gateway().await?;
@@ -226,13 +228,13 @@ pub async fn handle_ng_command(
                                 })
                                 .unwrap());
                             }
-                            InternalPayState::RefundSuccess(outpoint) => {
+                            InternalPayState::RefundSuccess { outpoint, error } => {
                                 let e = format!(
-                                    "Internal payment failed. A refund was issued to {outpoint}"
+                                    "Internal payment failed. A refund was issued to {outpoint} Error: {error}"
                                 );
                                 return Err(anyhow!(e));
                             }
-                            InternalPayState::Error(e) => {
+                            InternalPayState::UnexpectedError(e) => {
                                 return Err(anyhow!(e));
                             }
                             _ => {}
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -20,7 +20,6 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::anyhow;
 use axum::http::StatusCode;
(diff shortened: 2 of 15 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -109,31 +109,41 @@ pub enum LightningMode {
...
+                ),
+                StatusCode::CONFLICT,
+            ),
+            _ => (
+                format!("An internal gateway error occurred"),
+                StatusCode::INTERNAL_SERVER_ERROR,
+            ),
+        };
+        let mut err = Cow::<'static, str>::Owned(error_message).into_response();
+        *err.status_mut() = status_code;
```
m1sterc001guy, line 146: @wbobeirne do you have any opinions on what status code should be returned here. Right now, I'm just returning 409 "conflict" when there's an error paying an invoice and 500 for everything else that might go wrong. But I'm happy to change this is there's something the makes more sense on the frontend

wbobeirne: Nah anything 4XX is fine if we're being intentionally vague. My only thought would be that just a straight 400 is the most generic whereas someone might interpret a 409 to mean that there's some stateful reason that the request doesn't work now, but may have worked in the past or work in the future. 400 tells the client not to try again with that request.

m1sterc001guy: Done. Switched to 400

fedimint-cli/src/client.rs
```diff
@@ -226,13 +228,13 @@ pub async fn handle_ng_command(
                                 })
                                 .unwrap());
                             }
-                            InternalPayState::RefundSuccess(outpoint) => {
+                            InternalPayState::RefundSuccess { outpoint, error } => {
                                 let e = format!(
-                                    "Internal payment failed. A refund was issued to {outpoint}"
+                                    "Internal payment failed. A refund was issued to {outpoint} Error: {error}"
```
dpc, line 233: I was told that by using `{x:?}` a stacktrace will not be lost.

dpc (dismissed):

fedimint-cli/src/client.rs
m1sterc001guy: I actually don't think we want a stack trace here. The error in this case occurred on the gateway side, a stack trace of the CLI won't be that useful

okjodom (approved):
✅

gateway/ln-gateway/src/lib.rs
```diff
@@ -109,31 +109,40 @@ pub enum LightningMode {
...
-        *err.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
+        // For privacy reasons, we do not return too many details about the failure of
+        // the request back to the client to prevent malicious clients from
+        // deducing state about the gateway/lightning node.
+        let (error_message, status_code) = match self {
+            GatewayError::OutgoingPaymentError(_) => (
+                "Error while paying lightning invoice. Outgoing contract will be refunded."
+                    .to_string(),
+                StatusCode::BAD_REQUEST,
+            ),
```
justinmoon, line 138: Fewer 500 errors 🎉

PR #2848, opened 2023-07-26
refactor(devimint): run DKG using the apis

douglaz:
devimint is supposed to test binaries, but currently DKG is using fedimint as a library, making it harder to run tests with custom fedimintd binaries.

This change drops this "library" usage and makes DKG runs as api calls to the underlining fedimintd process.

One difficulty of implementing this is that we don't know before DKG what will be the peer ids. But we need to start everything with the conventional names and configurations.

So it's tricky to set up different passwords for different peers (we don't know exactly which fedimintd will be which peer). To make things simple, we are now using the same password for all the fedimintds.

### DIFF
--- a/devimint/src/bin/faucet.rs
+++ b/devimint/src/bin/faucet.rs
@@ -1,5 +1,7 @@
+use std::path::PathBuf;
 use std::sync::Arc;
 
+use anyhow::Context;
 use axum::extract::State;
 use axum::http::StatusCode;
 use axum::routing::{get, post};
@@ -20,7 +22,7 @@ struct Cmd {
     #[clap(long, env = "FM_CLN_SOCKET")]
     cln_socket: String,
     #[clap(long, env = "FM_CONNECT_STRING")]
-    connect_string: String,
+    connect_string: Option<String>,
 }
 
 #[derive(Clone)]
@@ -35,7 +37,11 @@ impl Faucet {
         let url = cmd.bitcoind_rpc.parse()?;
         let (host, auth) = fedimint_bitcoind::bitcoincore::from_url_to_url_auth(&url)?;
         let bitcoin = Arc::new(bitcoincore_rpc::Client::new(&host, auth)?);
-        let ln_rpc = Arc::new(Mutex::new(ClnRpc::new(&cmd.cln_socket).await?));
+        let ln_rpc = Arc::new(Mutex::new(
+            ClnRpc::new(&cmd.cln_socket)
+                .await
+                .with_context(|| format!("couldn't open CLN socket {}", &cmd.cln_socket))?,
+        ));
         Ok(Faucet { bitcoin, ln_rpc })
     }
 
@@ -89,6 +95,18 @@ impl Faucet {
     }
 }
 
+fn get_connect_string(connect_string: Option<String>) -> anyhow::Result<String> {
+    match connect_string {
+        Some(s) => Ok(s),
+        None => {
+            let data_dir = std::env::var("FM_DATA_DIR")?;
+            Ok(std::fs::read_to_string(
+                PathBuf::from(data_dir).join("client-connect"),
+            )?)
+        }
+    }
+}
+
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     TracingSetup::default().init()?;
@@ -97,7 +115,10 @@ async fn main() -> anyhow::Result<()> {
     let router = Router::new()
         .route(
             "/connect-string",
-            get(|| async move { cmd.connect_string.clone() }),
+            get(|| async move {
+                get_connect_string(cmd.connect_string)
+                    .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, format!("{e:?}")))
+            }),
         )
         .route(
             "/pay",
@@ -105,7 +126,7 @@ async fn main() -> anyhow::Result<()> {
                 faucet
                     .pay_invoice(invoice)
                     .await
-                    .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))
+                    .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, format!("{e:?}")))
             }),
         )
         .route(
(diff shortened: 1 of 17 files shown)


devimint/src/federation.rs
```diff
@@ -30,16 +32,59 @@ impl Federation {
...
         bitcoind: Bitcoind,
-        vars: BTreeMap<usize, vars::Fedimintd>,
+        servers: usize,
+        // vars: BTreeMap<usize, vars::Fedimintd>,
     ) -> Result<Self> {
         let mut members = BTreeMap::new();
-        for (peer, var) in &vars {
+        let mut vars = BTreeMap::new();
+
+        // FIXME: this is a hack to pass fed.server_gen_params in
```
douglaz, line 41: @justinmoon do you think this is still something to be fixed? what would be the proper way of doing it?

devimint/src/federation.rs
```diff
@@ -30,16 +32,59 @@ impl Federation {
...
+        // local_config_gen_params call below
+        let fed = FedimintBuilder::new()?.with_default_modules();
+        let peers: Vec<_> = (0..servers).map(|id| PeerId::from(id as u16)).collect();
+        let params: HashMap<PeerId, ConfigGenParams> =
+            local_config_gen_params(&peers, BASE_PORT, fed.server_gen_params.clone())?;
+
+        let mut admin_clients: BTreeMap<PeerId, WsAdminClient> = BTreeMap::new();
+        for (peer, peer_params) in &params {
+            // FIXME: we should do this inside
```
douglaz, line 50: @justinmoon should we also drop the fixme?

justinmoon:
> One difficulty of implementing this is that we don't know before DKG what will be the peer ids

Because DKG sorts TLS certs to produce peer IDs

devimint/src/bin/faucet.rs
```diff
@@ -97,7 +103,13 @@ async fn main() -> anyhow::Result<()> {
     let router = Router::new()
         .route(
             "/connect-string",
-            get(|| async move { cmd.connect_string.clone() }),
+            get(|| async move {
+                cmd.connect_string.unwrap_or_else(|| {
+                    // return error if not set
+                    let data_dir = std::env::var("FM_DATA_DIR").unwrap();
+                    std::fs::read_to_string(PathBuf::from(data_dir).join("client-connect")).unwrap()
```
justinmoon, line 110: would be nice to not unwrap here

devimint/src/federation.rs
```diff
@@ -30,16 +31,56 @@ impl Federation {
     pub async fn new(
         process_mgr: &ProcessManager,
         bitcoind: Bitcoind,
-        vars: BTreeMap<usize, vars::Fedimintd>,
+        servers: usize,
+        // vars: BTreeMap<usize, vars::Fedimintd>,
```
justinmoon, line 35: comments

devimint/src/federation.rs
justinmoon: not sure what i meant here, just delete comment

devimint/src/main.rs
```diff
@@ -152,7 +152,7 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         "--in-file={data_dir}/server-0/private.encrypt",
         "--out-file={data_dir}/server-0/config-plaintext.json"
     )
-    .env("FM_PASSWORD", "pass0")
+    .env("FM_PASSWORD", "pass")
```
justinmoon, line 155: Note: we now are using the same password for every server in the devimint environment

devimint/src/main.rs
```diff
@@ -1175,6 +1174,18 @@ async fn main() -> Result<()> {
...
+    let ready_file = PathBuf::from(env::var("FM_TEST_DIR")?).join("ready");
+    match handle_command().await {
+        Ok(r) => Ok(r),
+        Err(e) => {
+            write_overwrite_async(ready_file, "ERROR").await?;
+            Err(e)
+        }
+    }
+}
+
```
justinmoon, line 1188: If devimint returned an error, I think it wasn't writing error status to that "ready file" to inform mprocs

misc/mprocs.yaml
```diff
@@ -23,5 +23,7 @@ procs:
     shell: tail -n +0 -F $FM_LOGS_DIR/bitcoind.log
   devimint:
     shell: tail -n +0 -F $FM_LOGS_DIR/devimint.log
+  devimint-outer:
+    shell: tail -n +0 -F $FM_LOGS_DIR/devimint-outer.log
```
justinmoon, line 27: I don't think we need both. Maybe just have `devimint` tab point at the output of calling `devimint` on the command line (which we call `devimint-outer.log` here). This file has strictly more information in it.

scripts/mprocs.sh
```diff
@@ -12,7 +12,8 @@ export FM_VERBOSE_OUTPUT=0
 
 source scripts/build.sh
 
-devimint dev-fed 2>/dev/null &
+mkdir -p $FM_LOGS_DIR
+devimint dev-fed 2>$FM_LOGS_DIR/devimint-outer.log &
```
justinmoon, line 16: tmuxinator, too?

justinmoon (approved):

douglaz:
@elsirion added a `poll_max_retries` which matches more closely the desired behavior

devimint/src/util.rs
```diff
@@ -236,6 +236,39 @@ macro_rules! cmd {
...
+                Ok(true) => return Ok(()),
+                other => {
+                    i += 1;
+                    debug!("polling {name} failed with: {other:?}, will retry... ({i}/{retries})");
+                    task::sleep(Duration::from_secs(1)).await;
+                }
+            }
+        }
+    }
+}
```
elsirion, line 270: Feels like `poll_value` could now be implemented as a special case of this fn, but good enough.

elsirion (approved):

scripts/mprocs.sh
justinmoon: I wonder why this shows error output than `devimint.log` does? It's kind of confusing that we have 2 different log files. Ideally we could just get the errors to show up in `devimint.log` ...

devimint/src/util.rs
douglaz: I think we could create a `poll_max_retries_value`. But the behavior regarding errors on the `pool` ones is a bit different. They fail immediately, so there is no "retry" on errors, it's just a simple "polling". I didn't analyze the current calls enough to see if it makes sense to change that. @Maan2003 do you have opinions here?

elsirion: I was playing around with a few solutions but seems too complicated indeed, nevermind.

scripts/mprocs.sh
justinmoon: Problem with having devimint write to `devimint.log` internally is that it doesn't catch panics -- which are are pretty imporant!

justinmoon (approved):

devimint/src/util.rs
Maan2003: We can make it explicitly using `ControlFlow<Result<T>>` instead of `Result<Option<T>>`

Issue #2838, opened 2023-07-25
Devimint + mprocs doesn't always cleanly shut down


justinmoon:
Sometimes if I start mprocs with `just mprocs` and the close it with `ctrl-a q y`, some processes like `fedimintd` or `lnd` continue to run indefinitely.

I think there's 2 separate problems here:

* More reliably stop child processes
* Communicate to user when devimint fails to stop child processes. Currently, you have scrutinize logs for port conflicts to diagnose this problem.

Maan2003:
There should be `ProcessManager::wait_all_dead` that should wait for all processes to die. (store all pids in ProcessMangaer)
and this function should be called at end of main() in devimint

justinmoon:
> There should be `ProcessManager::wait_all_dead` that should wait for all processes to die. (store all pids in ProcessMangaer) and this function should be called at end of main() in devimint

I made a little prototype of this: https://github.com/justinmoon/fedimint/tree/devimint-shutdown

I used a crate called [sys_info](https://crates.io/crates/sysinfo) to see if process with some `pid` were still alive. This crate seems unreliable, though, because sometimes during shutdown it will report that the number of processes has increased. Might be better to just have `ProcessManager` store child processes.

I think part of the problem here is that we redirect the output of `devimint dev-fed` to `/dev/null` so when you leave mprocs, you don't see any output. We could have `mprocs.sh` call something like `devimint wait-shutdown` like we do on startup and use a file to communicate progress.

douglaz:
one test is exiting mprocs without waiting the normal startup after running:
```bash
$ just mprocs && for i in $(seq 0 100); do killall -0 devimint  bitcoind electrs esplora lnd lightningd fedimintd gatewayd fedimint-cli faucet; echo $i;sleep 1;done
./scripts/mprocs.sh
Run with 'source ./scripts/build.sh [fed_size] [dir]
Setting up env variables in /tmp/nix-shell.7rPUBK/fm-Mb3j
    Finished dev [unoptimized + debuginfo] target(s) in 0.33s
Killing fedimint processes
gatewayd: no process found
fedimint-cli: no process found
faucet: no process found
0
gatewayd: no process found
fedimint-cli: no process found
faucet: no process found
1
fedimint-cli: no process found
2
fedimint-cli: no process found
3
fedimint-cli: no process found
4
fedimint-cli: no process found
5
fedimint-cli: no process found
6
fedimint-cli: no process found
7
fedimint-cli: no process found
8
9
[...]

PR #2853, opened 2023-07-28
fix: make devfed immediately kill processes when exiting

douglaz:
Improves a lot https://github.com/fedimint/fedimint/issues/2838

Will make devimint always shut down the subprocesses when devfed exits (e.g when tmuxinator/mprocs is closed).

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -1091,7 +1091,7 @@ use std::str::FromStr;
 
 use fedimint_core::encoding::Decodable;
 
-async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
+async fn setup(arg: CommonArgs) -> Result<ProcessManager> {
     let globals = vars::Global::new(&arg.test_dir, arg.fed_size).await?;
     let log_file = fs::OpenOptions::new()
         .write(true)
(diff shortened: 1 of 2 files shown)


elsirion (approved):

devimint/src/main.rs
```diff
@@ -1114,59 +1114,93 @@ async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
...
 }
 
 #[tokio::main]
 async fn main() -> Result<()> {
     let args = Args::parse();
     match args.command {
         Cmd::ExternalDaemons => {
-            let (process_mgr, task_group) = setup(args.common).await?;
+            let process_mgr = setup(args.common).await?;
+            let task_group = TaskGroup::new();
```
okjodom, line 1126: @douglaz, I'm trying to understand the implication of this refactor for creating TG and installing kill handler. Does moving these two lines from within setup have a specific effect in killing processes later?

okjodom: oh maybe it's because tg is only used in `DevFed` and `ExternalDaemons` scenario?

okjodom (approved):
I had a curiosity question, otherwise, lgtm.
Landing to get the fresh benefits in `fedimint/ui`

devimint/src/main.rs
```diff
@@ -1114,59 +1114,93 @@ async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
...
         }
         Cmd::DevFed => {
-            let (process_mgr, task_group) = setup(args.common).await?;
-            let dev_fed = dev_fed(&process_mgr).await?;
-            dev_fed.fed.pegin(10_000).await?;
-            dev_fed.fed.pegin_gateway(20_000, &dev_fed.gw_cln).await?;
-            dev_fed.fed.pegin_gateway(20_000, &dev_fed.gw_lnd).await?;
-            let _daemons = write_ready_file(&process_mgr.globals, Ok(dev_fed)).await?;
-            task_group.make_handle().make_shutdown_rx().await.await?;
+            let process_mgr = setup(args.common).await?;
```
justinmoon, line 1134: The PR generally looks good, but this code block is a little hard to follow.

devimint/src/main.rs
```diff
@@ -1114,59 +1114,93 @@ async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
...
+            let task_group = TaskGroup::new();
```
dpc, line 1135: Maybe we should have `TaskGroup::new_with_kill_handler()` to saves some typing. :D

devimint/src/util.rs
```diff
@@ -21,19 +22,16 @@ fn kill(child: &Child) {
...
-        let mut child = std::mem::take(&mut process_handle_inner.child).unwrap();
-        info!(LOG_DEVIMINT, "killing {}", process_handle_inner.name);
-        kill(&child);
-        child.wait().await?;
+    pub async fn kill(&self) -> Result<()> {
+        let mut inner = self.0.lock().await;
+        if let Some(mut child) = inner.child.take() {
+            info!(LOG_DEVIMINT, "killing {}", inner.name);
+            kill(&child);
+            child.wait().await?;
```
dpc, line 33: We should ignore the error here. If we're killing it, chances are high it will fail, and we don't really care anymore.

devimint/src/main.rs
douglaz: Yes, it's not used everywhere and I think being explicit is better if you are reading the code for the first time

devimint/src/util.rs
dpc: We probably shouldn't even wait here in the first place. We've sent it a kill signal, we're done.

devimint/src/util.rs
```diff
@@ -79,10 +82,19 @@ impl ProcessManager {
...
             child: Some(child),
-        })))
+        })));
+        self.handles.lock().await.push(handle.clone());
+        Ok(handle)
+    }
+
+    pub async fn kill_all_children(&self) -> Result<()> {
+        for handle in self.handles.lock().await.iter() {
+            handle.kill().await?;
```
dpc, line 95: This terminating early could prevent us from killing other children. Biggest issue I see.

devimint/src/util.rs
douglaz: No, we must be sure they are dead :wink:

devimint/src/main.rs
douglaz: To understand what is happening read the log messages

douglaz: Also added to `run-ui` command.

devimint/src/main.rs
douglaz: For me it's fine the way it is.

devimint/src/util.rs
douglaz: Now wil kill everything in parallel and just log the resulting errors

douglaz:
Addressed most of comments

devimint/src/util.rs
dpc: But the wait here doesn't not check if killed them right. It checks the return status of the command itself, which will be an error ... because we killed it. Unless it was already dead. No?

douglaz: It may or may be not an error. We are sending a SIGTERM, so the process can gracefully shutdown.

douglaz: If the process just ignores SIGTERM, then this won't return

douglaz: we could implement a timeout then retry will SIGKILL, but our processes are well behaved, so I don't think we need this for now

dpc: Oh. You're right. I confused myself and thought we're sending SIGKILL already.

But I still think that most programs will terminate with an non-zero code on any signal:

image

(red `>` means non-zero exit value, Ctrl+C sends a SIGINT).

So we should wait, but is there a point in making in an `Err`?

devimint/src/util.rs
```diff
@@ -52,13 +50,18 @@ impl Drop for ProcessHandleInner {
     }
 }
 
+#[derive(Clone)]
 pub struct ProcessManager {
-    pub globals: vars::Global,
+    pub globals: Arc<vars::Global>,
+    handles: Arc<Mutex<Vec<ProcessHandle>>>,
```
Maan2003, line 56: incorrect: keeping a reference to ProcessHandle will prevent it from stopping on drop.

douglaz: So what? Shouldn't we always explicit kill them?

dpc: Should we make it a `Vec<WeakProcessHandle>` (with `Weak` instead of `Arc`?)

Maan2003: I liked the invariant that Bitcoind value existing means bitcoind is running, maybe it is not important?

Maan2003:
Sorry misclick on disabling auto merge

douglaz:
WASM tests are failing :cry:

devimint/src/main.rs
```diff
@@ -1114,58 +1114,106 @@ async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
...
+async fn cleanup_on_exit(
+    main_process: impl futures::Future<Output = Result<()>>,
+    task_group: TaskGroup,
+    process_mgr: ProcessManager,
+) -> anyhow::Result<()> {
+    // This select makes it possible to exit earlier if a signal is received before
+    // the main process is finished
+    tokio::select! {
+        _ = task_group.make_handle().make_shutdown_rx().await => {
+            info!("Received shutdown signal before finishing setup");
```
Maan2003, line 1129: not sure about "Setup". Should be "main process"?
same for rest of logs in this function.

Maan2003 (commented):
LGTM, minor nits

douglaz:
Closed in favor of https://github.com/fedimint/fedimint/pull/2873

PR #2854, opened 2023-07-28
fix: our ring fork not compiling c code for wasm

### LINKED ISSUE #2843: Backup/recovery breaks WASM
We are currently using a [custom patched version of `ring`](https://github.com/Maan2003/ring) for WASM compatibility. Unfortunately there seems to be a bug in there that leads to runtime failures whenever ChaCha20 is used and thus not removed via LTO, which is the case whenever backup/recovery is invoked.

The problem is that certain functions that WASM expects are missing, which leads to the generated JS trying to load them from some `env` thing, which leads to errors. When removing the offending code the main problem, the missing functions, appears, namely:

* `GFp_poly1305_init`
* `GFp_poly1305_update`
* `GFp_poly1305_finish`
* `GFp_memcmp`

All of them seem related to ChaCha20. 

## Reproducing

The problem can be reproduced in our WASM tests suit using the following patch:

```patch
diff --git a/fedimint-wasm-tests/src/lib.rs b/fedimint-wasm-tests/src/lib.rs
index f6c9f21398..430d9ae073 100644
--- a/fedimint-wasm-tests/src/lib.rs
+++ b/fedimint-wasm-tests/src/lib.rs
@@ -59,6 +59,7 @@ mod tests {
     use fedimint_ln_client::{LightningClientExt, LnPayState, LnReceiveState, PayType};
     use futures::StreamExt;
     use wasm_bindgen_test::wasm_bindgen_test;
+    use fedimint_client::derivable_secret::DerivableSecret;
 
     use super::*;
 
@@ -98,6 +99,17 @@ mod tests {
         Err(anyhow::anyhow!("Lightning receive failed"))
     }
 
+
+    #[wasm_bindgen_test]
+    async fn derive_chacha_key() -> Result<()> {
+        let root_secret = DerivableSecret::new_root(&[0x42; 32], &[0x2a; 32]);
+        let key = root_secret.to_chacha20_poly1305_key();
+        // Prevent optimization
+        println!("key: {key:?}");
+
+        Ok(())
+    }
+
     #[wasm_bindgen_test]
     async fn receive_and_pay() -> Result<()> {
         let client = client(&faucet::connect_string().await?.parse()?).await?;
```

After applying it using `git apply` run `nix build -L .#ci.wasm-tests`, which should error and output the following line (amongst a lot of gibberish):

```
wasm-tests>     JavaScript error: http://127.0.0.1:42379/wasm-bindgen-test, line 1: TypeError: The specifier “env” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.
```

## Debugging

To further debug without running the WASM tests each time one can create a HTML file that loads the generated JS/WASM and open it in a browser. Maybe @wbobeirne or @justinmoon have a minimal example. In the browser, open the JS console, which should display a similar error line. Then you can manipulate the JS file to get to the main errors.

To debug WASM there is a useful tool `wasm2wat` (with nix just run `nix-shell -p wabt` to get a shell with it installed) which can convert binary WASM to human readable WAT.

dpc:
The actual fix:

https://github.com/dpc/ring/commit/5493e7e76d0d8fb1d3cbb0be9c4944700741b802

The reason the poly stuff was missing was:

https://github.com/dpc/ring/blob/5493e7e76d0d8fb1d3cbb0be9c4944700741b802/build.rs#L326

which took me embarrassingly long to figure out.

I just re-introduced `wasm32_c`, to keep the changes minimal.

Fix #2843

I confirmed that the "import from env" is gone. Though the repro in it now fails with:

```
Error: non-200 response code: 500
{"value":{"error":"unknown error","message":"Failed to decode response from marionette","stacktrace":""}}
error: test failed, to rerun pass `--lib`
```

but that seems something entirely different.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,4 +90,4 @@ debug = 1
 
 [patch.crates-io]
 secp256k1-zkp = { git = "https://github.com/dpc/rust-secp256k1-zkp/", branch = "sanket-pr" }
-ring = { git = "https://github.com/Maan2003/ring", rev = "52a1294a8fc066cc0656911ec22b8358fb763ef1" }
+ring = { git = "https://github.com/dpc/ring", rev = "5493e7e76d0d8fb1d3cbb0be9c4944700741b802" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4231,7 +4231,7 @@ dependencies = [
 [[package]]
 name = "ring"
 version = "0.16.20"
-source = "git+https://github.com/Maan2003/ring?rev=52a1294a8fc066cc0656911ec22b8358fb763ef1#52a1294a8fc066cc0656911ec22b8358fb763ef1"
+source = "git+https://github.com/dpc/ring?rev=5493e7e76d0d8fb1d3cbb0be9c4944700741b802#5493e7e76d0d8fb1d3cbb0be9c4944700741b802"
 dependencies = [
  "cc",
  "libc",
--- a/fedimint-client-legacy/Cargo.toml
+++ b/fedimint-client-legacy/Cargo.toml
@@ -57,7 +57,7 @@ jsonrpsee-ws-client = { version = "0.18.0", features = ["webpki-tls"], default-f
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 jsonrpsee-wasm-client = { version = "0.18.0" }
-ring = { version = "0.16.20", features = ["wasm32_unknown_unknown_js"] }
+ring = { version = "0.16.20", features = ["wasm32_unknown_unknown_js", "wasm32_c"] }
 
 [dev-dependencies]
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -85,9 +85,9 @@ threshold_crypto = { opt-level = 3 }
 
 [profile.ci]
 inherits = "dev"
-split-debuginfo = "packed"
-debug = 1
+# split-debuginfo = "packed"
+# debug = 1
```
Maan2003, line 89: unrelated

Maan2003 (approved):

Cargo.toml
dpc: Missed these leftovers.

Maan2003 (approved):

justinmoon (approved):

PR #2734, opened 2023-07-07
Client NG Join Federation

okjodom:
Adds ability for client builder to download config from Federation when building clients. The builder saves any successfully downloaded configs in client database. It skips re-downloads when building a client with a previously-used db, and prefers loading the config from the db.

Next:
- [x] Migrate `fedimint-cli` to `with_invite_code`
- [ ] Migrate gateway client builder to `with_invite_code` : https://github.com/fedimint/fedimint/pull/2855
- [ ] #2769 after moving all tests fixtures from `with_config` to `with_invite_code`

PR #2855, opened 2023-07-28
Refactor: Gateway Client Builder

okjodom:
Refactor gateway client builder to persist only the connection info, channel id, fees and timelock in gateway db.
The gateway client builder delegates registration with the federation to the common `ClientBuilder`, to which it provides a connection info the first time a client is registered

depends on #2734

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -74,7 +74,7 @@ use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::{anyhow, bail};
+use anyhow::{anyhow, bail, ensure};
 use async_stream::stream;
 use db::{CachedApiVersionSet, CachedApiVersionSetKey, ClientConfigKey, ClientConfigKeyPrefix};
 use fedimint_core::api::{
@@ -1256,10 +1256,17 @@ impl ClientBuilder {
         self.module_inits.attach(module_init);
     }
 
-    /// Uses this invite code to connect to the federation
+    /// Instructs the client builder to use this `InviteCode` in building a
+    /// client. Use this configuration method as an alternative to
+    /// `with_config` builder option.
     ///
-    /// ## Panics
-    /// If there was a config or invite code added previously
+    /// ## Failure Conditions
+    ///
+    /// If a `ConfigSource` already exists in the builder, calling this method
+    /// fails. If a `ClientConfig` already exists in the database, calling
+    /// `ClientBuilder::build` fails.
+    /// - You can use `get_config_from_db` to check if a config exists in the
+    ///   database.
     pub fn with_invite_code(&mut self, invite_code: InviteCode) {
         let was_replaced = self
             .config_source
@@ -1271,12 +1278,17 @@ impl ClientBuilder {
         )
     }
 
-    /// FIXME: <https://github.com/fedimint/fedimint/issues/2769>
+    /// Instructs the client builder to use this `ClientConfig` in building a
+    /// client. Use this configuration method as an alternative to
+    /// `with_invite_code` builder option.
     ///
-    /// Uses this config to initialize the client
+    /// ## Failure Conditions
     ///
-    /// ## Panics
-    /// If there is a invite code or config added previously
+    /// If a `ConfigSource` already exists in the builder, calling this method
+    /// fails. If a `ClientConfig` already exists in the database, calling
+    /// `ClientBuilder::build` fails.
+    /// - You can use `get_config_from_db` to check if a config exists in the
+    ///   database.
     pub fn with_config(&mut self, config: ClientConfig) {
         let was_replaced = self
             .config_source
@@ -1525,19 +1537,12 @@ async fn get_config(
     db: &Database,
     config_source: Option<ConfigSource>,
 ) -> anyhow::Result<ClientConfig> {
-    let mut dbtx = db.begin_transaction().await;
-    let config_res = match dbtx
-        .find_by_prefix(&ClientConfigKeyPrefix)
-        .await
-        .next()
-        .await
-    {
-        Some((_, config)) => {
-            // TODO: Enable after <https://github.com/fedimint/fedimint/pull/2855>
-            // assert!(
-            //     config_source.is_none(),
-            //     "Alternative config source provided but config was found in DB"
-            // );
+    let config_res = match get_config_from_db(db).await {
+        Some(config) => {
+            ensure!(
+                config_source.is_none(),
+                "Alternative config source provided but config was found in DB"
+            );
             Ok(config)
         }
         None => {
@@ -1569,6 +1574,18 @@ async fn get_config(
     // some borrowck lifetime limitation thing
     #[allow(clippy::let_and_return)]
     config_res
+}
+
+pub async fn get_config_from_db(db: &Database) -> Option<ClientConfig> {
+    let mut dbtx = db.begin_transaction().await;
+    #[allow(clippy::let_and_return)]
+    let config = dbtx
+        .find_by_prefix(&ClientConfigKeyPrefix)
+        .await
+        .next()
+        .await
+        .map(|(_, config)| config);
+    config
 }
 
 /// Tries to download the client config from the federation,
--- a/gateway/ln-gateway/src/client.rs
+++ b/gateway/ln-gateway/src/client.rs
@@ -5,12 +5,11 @@ use std::sync::Arc;
 
 use fedimint_client::module::init::ClientModuleInitRegistry;
 use fedimint_client::secret::PlainRootSecretStrategy;
-use fedimint_client::ClientBuilder;
-use fedimint_core::api::{DynGlobalApi, GlobalFederationApi, InviteCode, WsFederationApi};
+use fedimint_client::{get_config_from_db, ClientBuilder};
 use fedimint_core::core::ModuleInstanceId;
-use fedimint_core::db::DatabaseTransaction;
+use fedimint_core::db::{Database, DatabaseTransaction};
+use fedimint_core::module::registry::ModuleDecoderRegistry;
 use futures::StreamExt;
-use lightning::routing::gossip::RoutingFees;
 
 use crate::db::{FederationConfig, FederationIdKey, FederationIdKeyPrefix};
 use crate::lnrpc_client::ILnRpcClient;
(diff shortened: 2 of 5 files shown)


gateway/ln-gateway/src/client.rs
```diff
@@ -48,21 +47,31 @@ impl StandardGatewayClientBuilder {
...
-        let federation_id = config.config.federation_id;
+        let FederationConfig {
+            connection,
+            mint_channel_id,
+            timelock_delta,
+            fees,
+        } = config;
+        let federation_id = connection.id;
+
+        // TODO: remove after <https://github.com/fedimint/fedimint/issues/2506>
```
m1sterc001guy, line 58: What is the plan after this issue is resolved?

gateway/ln-gateway/src/client.rs
```diff
@@ -48,21 +47,31 @@ impl StandardGatewayClientBuilder {
...
+        let api: DynGlobalApi = WsFederationApi::from_connect_info(&[connection.clone()]).into();
+        let client_config = api.download_client_config(&connection).await?;
```
m1sterc001guy, line 60: If this fails, will `gatewayd` fail to start, or does it just skip initializing this client for the federation?

m1sterc001guy:
> we can always download FM client config and recreate gateway config

This might actually be a good thing? i.e if we can't download the config from the federation, the federation is likely not available, in which case it's not super useful to construct the client anyway.

gateway/ln-gateway/src/client.rs
okjodom: gateway config builder does not have to download configs. Instead, the client builder will take over that responsibility, plus persisting downloaded configs

gateway/ln-gateway/src/client.rs
okjodom: Skips initializing client for the federation. So a `/connect-fed` request fails

m1sterc001guy: I believe these clients are initialized when the gateway first boots up though too. So if federation X is offline when `gatewayd` boots up, does that prevent `gatewayd` from starting correctly?

m1sterc001guy: I think ideally if this fails we should just catch the failure and emit a warning message or something that we failed to load the client for that federation (and at some point make that visible in the UI)

okjodom:
> > we can always download FM client config and recreate gateway config
> 
> This might actually be a good thing? i.e if we can't download the config from the federation, the federation is likely not available, in which case it's not super useful to construct the client anyway.

Like liveness check on the federation :)

gateway/ln-gateway/src/client.rs
okjodom: > I believe these clients are initialized when the gateway first boots up though too. So if federation X is offline when gatewayd boots up, does that prevent gatewayd from starting correctly

I hadn't considered the case of re-initializing clients on gatewayd boot. It's true failure to load any single client results in failure to load gateway.

With this change where we always download config from federation, we have higher chances of failing to load gateway!

> I think ideally if this fails we should just catch the failure and emit a warning message or something that we failed to load the client for that federation (and at some point make that visible in the UI)

Agreed. Tracking here https://github.com/fedimint/fedimint/issues/2857

elsirion:
Blocked on #2734

fedimint-client/src/lib.rs
```diff
@@ -1469,19 +1469,12 @@ async fn get_config(
...
-            // assert!(
-            //     config_source.is_none(),
-            //     "Alternative config source provided but config was found in DB"
-            // );
+    let config_res = match get_config_from_db(db).await {
+        Some(config) => {
+            assert!(
+                config_source.is_none(),
+                "Alternative config source provided but config was found in DB"
+            );
```
elsirion, line 1477: Can we make this an [`ensure!`](https://docs.rs/anyhow/1.0.73/anyhow/macro.ensure.html) please? It feels way too easy to get wrong and we don't want to needlessly blow up the whole application. It's a close call here tbf.

fedimint-client/src/lib.rs
```diff
@@ -1513,6 +1506,17 @@ async fn get_config(
     config_res
 }
 
+pub async fn get_config_from_db(db: &Database) -> Option<ClientConfig> {
```
elsirion, line 1509: The docs for [`with_invite_code`](https://docs.fedimint.org/fedimint_client/struct.ClientBuilder.html#method.with_invite_code) should refer to this function as it can be used to check if a config already exists. 

While multi-federation wallets will typically have a way to tell which federation the user joined or not single-federation wallets might just want to check if there is any config in the DB.

gateway/ln-gateway/src/client.rs
```diff
@@ -69,6 +71,7 @@ impl StandardGatewayClientBuilder {
                 GatewayError::DatabaseError(anyhow::anyhow!("Error opening rocksdb: {e:?}"))
             })?;
             client_builder.with_database(db);
+            client_builder.with_invite_code(invite_code);
```
elsirion, line 74: The `StandardGatewayClientBuilder::build` function is called when a gateway is being restarted and is creating clients for already joined federations. Wouldn't this be a problem here since there'd already be a config in the DB?

elsirion (commented):
Only one place I don't understand why it works, looks good otherwise.

I don't have a full understanding of the gateway's internal mechanics, so would refer to @m1sterc001guy for a final review.

gateway/ln-gateway/src/client.rs
okjodom: The restart path provides an `old_client` reference, in which case we don't provide an invite code to the client builder

elsirion: It seems to me that `start_gateway` creates an empty `client` map, passes it to `Gateway::new`, which initialized a `Gateway` struct with the empty client map and then calls `Gateway::load_clients` which tried to get an old client from that map, but won't find one and thus passes `None`:

https://github.com/fedimint/fedimint/blob/41e8ce310193408359adda152081ad79a46ee8dc/gateway/ln-gateway/src/lib.rs#L564-L568

I'm very suspicious that tests are running though, so I likely have an error in my thinking somewhere, but that also means the control/data flow is hard to understand. I'll leave the final merge decision to @m1sterc001guy since he has more context.

elsirion (approved):

fedimint-client/src/lib.rs
```diff
@@ -1226,10 +1226,17 @@ impl ClientBuilder {
         self.module_gens.attach(module_gen);
     }
 
-    /// Uses this invite code to connect to the federation
+    /// Instructs the client builder to use this `InviteCode` in building a
+    /// client Use this configuration method as an alternative to
```
m1sterc001guy, line 1230: Nit: missing .

gateway/ln-gateway/src/client.rs
m1sterc001guy: > and then calls Gateway::load_clients which tried to get an old client from that map

`Gateway::load_clients` actually reads from `Gatewayd`'s database, which is why it will actually find a `FederationConfig` there and won't pass `None`. The Lightning reconnect test should be explicitly testing this case, since this is exercised when the connection to the lightning node drops which prompts `Gatewayd` to re-create the `Gateway` struct and re-build clients. So if that test is passing I think we're good. 

Maybe we should add a comment to `load_clients` or rename the `clients` map? That might make this a little easier to understand.

fedimint-client/src/lib.rs
```diff
@@ -1494,19 +1506,12 @@ async fn get_config(
...
-        .next()
-        .await
-    {
-        Some((_, config)) => {
-            // TODO: Enable after <https://github.com/fedimint/fedimint/pull/2855>
-            // assert!(
-            //     config_source.is_none(),
-            //     "Alternative config source provided but config was found in DB"
-            // );
+    let config_res = match get_config_from_db(db).await {
```
m1sterc001guy, line 1509: Where is this function used? Does the gateway ever use `get_config`?

fedimint-client/src/lib.rs
okjodom: fixed :)

fedimint-client/src/lib.rs
okjodom: The static methods `get_config` and `get_config_from_db` are used in `ClientBuilder`. Gateway composes the client builder. See `build_stopped` method of the client builder

gateway/ln-gateway/src/client.rs
okjodom: > It seems to me that `start_gateway` creates an empty `client` map, passes it to `Gateway::new`, which initialized a `Gateway` struct with the empty client map and then calls `Gateway::load_clients` which tried to get an old client from that map, but won't find one and thus passes `None`:
> 
> https://github.com/fedimint/fedimint/blob/41e8ce310193408359adda152081ad79a46ee8dc/gateway/ln-gateway/src/lib.rs#L564-L568
> 
> I'm very suspicious that tests are running though, so I likely have an error in my thinking somewhere, but that also means the control/data flow is hard to understand. I'll leave the final merge decision to @m1sterc001guy since he has more context.

utACK.

> The Lightning reconnect test should be explicitly testing this case, since this is exercised when the connection to the lightning node drops which prompts Gatewayd to re-create the Gateway struct and re-build clients. So if that test is passing I think we're good.

To recreate the fail scenario @elsirion described above, we need to [rerun Gatewayd with the same db path](https://github.com/fedimint/fedimint/blob/e29d7aac010cefef2ea817bc0467def1556dbca5/gateway/ln-gateway/src/lib.rs#L192C1-L208C11). Lightning tests don't presently cover this scenario

okjodom: > To recreate the fail scenario @elsirion described above, we need to [rerun Gatewayd with the same db path](https://github.com/fedimint/fedimint/blob/e29d7aac010cefef2ea817bc0467def1556dbca5/gateway/ln-gateway/src/lib.rs#L192C1-L208C11). Lightning reconnect tests don't presently cover this scenario 

Not necessary. The bug and untested scenario occurs because we can rebuild gateway fed clients with the same db, but `old_client` is not a sufficient flag to tell that we are in a rebuild scenario, where db possibly has a client config already.

Fix is to explicitly check if config is in client database, using `get_config_from_db`

fedimint-client/src/lib.rs
okjodom: gateway client builder now uses `get_config_from_db` to detect client rebuild scenario

gateway/ln-gateway/src/client.rs
m1sterc001guy: > The bug and untested scenario occurs because we can rebuild gateway fed clients with the same db, but old_client is not a sufficient flag to tell that we are in a rebuild scenario, where db possibly has a client config already.

Is this because we construct `old_client` based on whether or not there's a `FederationConfig` in the database, when we should actually be looking for `ClientConfig`? I'm a bit confused what the gap is here.

okjodom: > I haven't fully tested this, so marking PR as draft

PR #3241 adds test coverage for gateway reboot scenario, which helps us maintain some sanity around gateway client-building on gatewayd boot.

gateway/ln-gateway/src/client.rs
```diff
@@ -47,30 +46,47 @@ impl StandardGatewayClientBuilder {
...
+                })?;
+
+                // Initialize a client database to check if a config was previously saved in it
+                let db = Database::new(rocksdb, ModuleDecoderRegistry::default());
+                if (get_config_from_db(&db).await).is_none() {
+                    client_builder.with_invite_code(invite_code);
+                }
+            }
+
+            let rocksdb = fedimint_rocksdb::RocksDb::open(db_path.clone()).map_err(|e| {
```
m1sterc001guy, line 86: Nit: Do we need to open a second handle to `rocksdb` or is it possible to just open it once, check if the config exists, then add the db using `with_database`?

gateway/ln-gateway/src/db.rs
```diff
@@ -23,10 +24,10 @@ pub struct FederationIdKeyPrefix;
 
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable)]
 pub struct FederationConfig {
+    pub invite_code: InviteCode,
```
m1sterc001guy, line 27: This will be a breaking change for anyone running a gateway? I assume we're ok with that.

m1sterc001guy (dismissed):
Looks good to me

gateway/ln-gateway/src/db.rs
okjodom: Yeah. Okay with breaking change. But now that we have releases, should we start tracking in changelog for the next release?

gateway/ln-gateway/src/client.rs
okjodom: I couldn't figure how to keep that first rocksdb handle after consuming it and checking if config is in db
```rs
    let db = Database::new(rocksdb, ModuleDecoderRegistry::default());
    if (get_config_from_db(&db).await).is_none() {
        client_builder.with_invite_code(invite_code);
    }
```

okjodom: Any suggestions on how to do this?

m1sterc001guy: I don't think rocksdb is cloneable so this might not be easy to do, I think its fine as is then

gateway/ln-gateway/src/db.rs
m1sterc001guy: Maybe @justinmoon has thoughts on how to track breaking changes for the next release, otherwise I'm good to merge this

okjodom: dev call: we can add an identifying tag to PRs  to track breaking changes

m1sterc001guy (approved):

PR #2863, opened 2023-07-31
cleanup: simplify logging

joschisan:


elsirion (approved):
Much simpler imo. I still think we should re-add `warn` level logging for the unexpected cases again that would indicate malicious peers (or more likely some bug).

PR #2864, opened 2023-07-31
chore(tests): add WASM test for ring chacha20 being available

### LINKED ISSUE #2843: Backup/recovery breaks WASM

elsirion:
Fixes #2843

### DIFF
--- a/fedimint-wasm-tests/src/lib.rs
+++ b/fedimint-wasm-tests/src/lib.rs
@@ -54,6 +54,7 @@ mod faucet {
 
 wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
 mod tests {
+    use fedimint_client::derivable_secret::DerivableSecret;
     use fedimint_core::task::TaskGroup;
     use fedimint_core::Amount;
     use fedimint_ln_client::{LightningClientExt, LnPayState, LnReceiveState, PayType};
@@ -96,6 +97,18 @@ mod tests {
             }
         }
         Err(anyhow::anyhow!("Lightning receive failed"))
+    }
+
+    // Tests that ChaCha20 crypto functions used for backup and recovery are
+    // available in WASM at runtime. Related issue: https://github.com/fedimint/fedimint/issues/2843
+    #[wasm_bindgen_test]
+    async fn derive_chacha_key() {
+        let root_secret = DerivableSecret::new_root(&[0x42; 32], &[0x2a; 32]);
+        let key = root_secret.to_chacha20_poly1305_key();
+
+        // Prevent optimization
+        // FIXME: replace with `std::hint::black_box` once stabilized
+        assert!(format!("key: {key:?}").len() > 8);
     }
 
     #[wasm_bindgen_test]


fedimint-wasm-tests/src/lib.rs
```diff
@@ -98,6 +99,16 @@ mod tests {
         Err(anyhow::anyhow!("Lightning receive failed"))
     }
 
+    #[wasm_bindgen_test]
+    async fn derive_chacha_key() -> Result<()> {
+        let root_secret = DerivableSecret::new_root(&[0x42; 32], &[0x2a; 32]);
+        let key = root_secret.to_chacha20_poly1305_key();
+        // Prevent optimization
+        println!("key: {key:?}");
```
Maan2003, line 107: [`std::hint::black_box`](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html)

because println might be noop on wasm.

dpc: While at it, a comment explaining why we have it might be handy.

elsirion: @Maan2003 thx for the pointer, unfortunately it's only available on nightly :(

Maan2003: I think our cargo is too old.

Maan2003: black box has been stable since 1.66

Maan2003 (approved):

fedimint-wasm-tests/src/lib.rs
justinmoon: My cargo version inside `nix develop` is `cargo 1.67.0 (8ecd4f20a 2023-01-10)`

Maan2003: https://doc.rust-lang.org/stable/src/core/hint.rs.html#290 says 1.66 though :thinking:

dpc (approved):

PR #2865, opened 2023-07-31
fix: improve killing of devimint

Maan2003:
fixes devimint doesn't die correctly if setup fails.


dpc:
What about https://github.com/fedimint/fedimint/pull/2853 ?

douglaz:
If we shutdown here we may not have enough time to do the killing of subprocesses

Maan2003:
Closed in favour of #2853

PR #2868, opened 2023-07-31
Update to thunderhub v0.13.20 with signet support

Schnitzel:
thunderhub before v0.13.20 does not have signet support and therefore fails. This upgrades to v0.13.20.


douglaz (approved):
Thanks. This was a known problem. Have you tested the setup with this version? Perhaps we will need further tweaking on docker-compose.

Anyway it's a step in the right direction, so LGTM

PR #2869, opened 2023-07-31
WIP: Fedimint pays LDK Node

m1sterc001guy:
Don't merge this.

This is a Proof of Concept implementation for spawning an LDK node in our tests, creating an invoice, and paying that invoice using Fedimint.


justinmoon:
Removing sqlite might be better as a separate PR?

gateway/ln-gateway/src/ng/tests.rs
```diff
@@ -74,10 +78,83 @@ pub fn sha256(data: &[u8]) -> sha256::Hash {
...
+
+        Ok(())
+    })
+    .await
+}
+
 #[tokio::test(flavor = "multi_thread")]
 async fn test_gateway_client_pay_valid_invoice() -> anyhow::Result<()> {
     gateway_test(
-        |gateway, other_lightning_client, fed, user_client| async move {
```
dpc, line 80: Seems to me like these would be better of in:

```rust
#[non_exhaustive]
Fixtures {
  gateway: Gateway,
  client: Client,
}
```

and then it could be:

```rust
gateway_test(|Fixtures { gateway, client, .. }| /* body */);
```

kind of thing.

elsirion: I agree, but could be a separate refactor imo.

fedimint-testing/src/ln/mod.rs
```diff
@@ -55,9 +55,8 @@ pub trait LightningTest: ILnRpcClient {
             .unwrap())
     }
 
-    /// Returns the amount that the gateway LN node has sent
-    async fn amount_sent(&self) -> Amount;
-
```
elsirion, line 60: Don't we use this at all in tests? :open_mouth:

elsirion (approved):
Looks like a good fist step to me :)

fedimint-testing/src/ln/mod.rs
m1sterc001guy: I was surprised to learn that we don't use this at all. Perhaps leftover from a previous refactor

PR #2870, opened 2023-07-31
chore: test backup&restore both with and without an existing backup

dpc:


devimint/src/federation.rs
```diff
@@ -27,6 +27,52 @@ pub struct Federation {
...
+    /// fork each other as they please.
+    async fn new_forked(name: &str) -> Result<Client> {
+        let workdir: PathBuf = env::var("FM_DATA_DIR")?.parse()?;
+        let client_dir = workdir.join("clients").join(name);
+
+        std::fs::create_dir_all(&client_dir)?;
+
+        cmd!(
+            "cp",
+            "-R",
```
justinmoon, line 49: Don't need the `-R`

PR #2872, opened 2023-07-31
fix: purge "block height", use "block count"

dpc:
It has been established more than a half-century ago that in computer science and engineering when operating on indexes, ranges and counts it's best to start numbering from 0, and use half-open ranges:

https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html

Mathematicians sometimes disagree, but what do they know. And there's Lua, which is a neat PL, yet most people using it complain precisely about the 1-based indexing (unless they are mathematicians).

Given that block height are 0-based and genesis block is has a height of 0:

https://mempool.space/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f

it is most ergonomic to operate on "block count".

Bitcoin Core did the right thing and has a `getblockcount` call.

It's also worth noting that "block height" (as opposed to a
 "block count") is a misnomer. Which block? "Blockchain height" could
maybe work, so would "last/current/tip height", but bare "block height" is just confusing.

There are situations in which "give me the *content* (or a hash) of the last block" is a useful API to have, but even then a combination of "get block count" and "get content/hash at height N" is a much better API, unless that extra-call latency is really intolerable (which is never).

I guess people "on the streets" (Twitters?), like to talk about "we're at block height X" to meassure blockchain growth, but they also count from 1, so yeah...

### DIFF
--- a/fedimint-client-legacy/src/lib.rs
+++ b/fedimint-client-legacy/src/lib.rs
@@ -727,20 +727,20 @@ impl<T: AsRef<ClientConfig> + Clone + MaybeSend> Client<T> {
         self.reissue(notes_to_reissue, rng).await
     }
 
-    pub async fn await_consensus_block_height(
+    pub async fn await_consensus_block_count(
         &self,
-        block_height: u64,
+        block_count: u64,
     ) -> std::result::Result<u64, task::Elapsed> {
         task::timeout(Duration::from_secs(30), async {
-            self.await_consensus_block_height_inner(block_height).await
+            self.await_consensus_block_count_inner(block_count).await
         })
         .await
     }
 
-    async fn await_consensus_block_height_inner(&self, block_height: u64) -> u64 {
+    async fn await_consensus_block_count_inner(&self, requested_count: u64) -> u64 {
         loop {
-            match self.context.api.fetch_consensus_block_height().await {
-                Ok(height) if height >= block_height => return height,
+            match self.context.api.fetch_consensus_block_count().await {
+                Ok(current_count) if requested_count <= current_count => return current_count,
                 _ => sleep(Duration::from_millis(100)).await,
             }
         }
@@ -860,8 +860,8 @@ impl Client<UserClientConfig> {
         let mut dbtx = self.context.db.begin_transaction().await;
         let mut tx = TransactionBuilder::default();
 
-        let consensus_height = self.context.api.fetch_consensus_block_height().await?;
-        let absolute_timelock = consensus_height + OUTGOING_LN_CONTRACT_TIMELOCK;
+        let consensus_count = self.context.api.fetch_consensus_block_count().await?;
+        let absolute_timelock = consensus_count + OUTGOING_LN_CONTRACT_TIMELOCK - 1;
 
         let contract = self
             .ln_client()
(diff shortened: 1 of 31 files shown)


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -829,7 +830,14 @@ impl ServerModule for Lightning {
             api_endpoint! {
                 "block_height",
```
dpc, line 831: Probably should just get rid of these...

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -593,7 +592,13 @@ impl ServerModule for Wallet {
             api_endpoint! {
                 "block_height",
```
dpc, line 593: And this one.

dpc:
Need to think about it.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -321,13 +321,10 @@ impl ServerModule for Wallet {
             .collect::<Vec<WalletConsensusItem>>()
             .await;
 
-        let block_height_proposal = self
-            .block_height()
-            .await
-            .saturating_sub(self.cfg.consensus.finality_delay);
```
dpc, line 327: swallowed it?

fedimint-client-legacy/src/lib.rs
```diff
@@ -727,20 +727,20 @@ impl<T: AsRef<ClientConfig> + Clone + MaybeSend> Client<T> {
...
         .await
     }
 
-    async fn await_consensus_block_height_inner(&self, block_height: u64) -> u64 {
+    async fn await_consensus_block_count_inner(&self, requested_count: u64) -> u64 {
         loop {
-            match self.context.api.fetch_consensus_block_height().await {
-                Ok(height) if height >= block_height => return height,
+            match self.context.api.fetch_consensus_block_count().await {
+                Ok(current_count) if requested_count <= current_count => return current_count,
```
douglaz, line 743: I hope there is not a off by one error here

douglaz (dismissed):
minor stuff, LGTM

fedimint-client-legacy/src/lib.rs
dpc: Double check me, but I've swapped current and requested as I like everything `a < b`, and they were renamed as they were poorly named.

dpc:
I think I broke something w.r.t LN contracts timing, but had no time to debug it yet.

fedimint-client-legacy/src/lib.rs
douglaz: Perhaps it's better to keep the previous behavior?
```suggestion
                Ok(current_count) if requested_count < current_count => return current_count,
```

dpc: Previous one was `>=` (now `<=` because I flipped them), no?

douglaz: Don't listen to me, I'm crazy

dpc: Me too! :D

douglaz:
rebase?

elsirion (approved):
Lgtm, I think we should check the LN module for off-by-ones anyway at some point, it always makes me feel uneasy.

douglaz (dismissed):
L G T :fire:

dpc:
I accidentally forced pushed something completely else here. :/
I restored previous code.

douglaz (approved):

elsirion:
> I accidentally forced pushed something completely else here. :/
> I restored previous code.

monty-python

PR #2873, opened 2023-08-01
fix(devimint): better handling of kill/terminate on subprocesses

douglaz:
Based on https://github.com/fedimint/fedimint/pull/2853

This PR introduces SIGKILL on drop and will leave SIGTERM for when graceful/controlled termination is required.

It will also finish earlier when exiting tmux/mprocs, sending a SIGKILL to subprocesses (through drop of `DevFed`/daemons).

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -183,8 +183,8 @@ impl Lightningd {
             .to_string())
     }
 
-    pub async fn kill(self) -> Result<()> {
-        self.process.kill().await
+    pub async fn terminate(self) -> Result<()> {
+        self.process.terminate().await
     }
 }
 
@@ -272,8 +272,8 @@ impl Lnd {
         Ok(())
     }
 
-    pub async fn kill(self) -> Result<()> {
-        self.process.kill().await
+    pub async fn terminate(self) -> Result<()> {
+        self.process.terminate().await
     }
 }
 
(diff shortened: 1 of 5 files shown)


Maan2003:
LGTM at conceptual level.

justinmoon:
LGTM as well. Would be nice if `mprocs.sh` and `tmuxinator.sh` would actually wait until devimint finished shutdown before exiting. If we could wait for shutdown [at the end of these files](https://thefedi.slack.com/archives/C03QSGEEYF5/p1690826351398209) that would really help the UX of using devimint -- so it's harder to re-launch it and hit port conflicts before the old instance shut down.

justinmoon (dismissed):

douglaz:
Investigating the tests failures

douglaz:
The problem seems to be the good old flaky tests like `cli_test_rust_tests_bitcoind`, but most of time I can't reproduce locally

dpc (approved):

dpc:
Rebasing might help.

justinmoon:
Attempt `#4` seems disk space related https://github.com/fedimint/fedimint/actions/runs/5732521842/job/15555554741?pr=2873#step:9:2287

douglaz:
Rebased to use https://github.com/fedimint/fedimint/pull/2893 which improves things a lot

justinmoon (approved):

PR #2875, opened 2023-08-01
fix: remove autocommit

joschisan:
We remove the autocommit here since there should never be write conflicts for consensus state. If there is a conflict we would rather crash instead of continuing with inconsistent state.

### DIFF
--- a/fedimint-server/src/consensus/mod.rs
+++ b/fedimint-server/src/consensus/mod.rs
@@ -128,73 +128,74 @@ impl FedimintConsensus {
         reference_rejected_txs: Option<BTreeSet<TransactionId>>,
     ) -> SignedEpochOutcome {
         let _timing /* logs on drop */ = timing::TimeReporter::new("process_consensus_outcome");
+
+        let mut rejected_txs = BTreeSet::new();
+        // Since multiple peers can submit the same tx within epoch, track them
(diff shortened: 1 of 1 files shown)


elsirion (approved):
What do you think @dpc @justinmoon @m1sterc001guy? The most likely cause for such a conflict is writing to consensus-relevant DB tables from API handlers, which is typically a very bad idea. So I agree we want to surface these.

We can always re-introduce this as a patch if we see such a problem in the wild and can't fix the underlying cause quickly enough. Another idea to make this easier would be setting `max_attempts` to 1.

m1sterc001guy:
Fully agreed with the above. I think this is a good change.

> Another idea to make this easier would be setting max_attempts to 1
IMO it is easier to read the code without the autocommit closure, but its essentially the same thing.

> If there is a conflict we would rather crash instead of continuing with inconsistent state.
Small nit on this, because we use snapshot isolation in RocksDb, the Fedimint database will never have an inconsistent state (in the database sense, perhaps at the application level). In this case, a the fedimint transaction would fail, since the db transaction commit would return an error. This would be unexpected behavior, and I agree the proper response is for `fedimintd` to crash.

m1sterc001guy (approved):

PR #2876, opened 2023-08-01
chore: devimint sets federation_name meta field

justinmoon:
Small fix from https://github.com/fedimint/fedimint/pull/2848. So far we've always set `federation_name` in the meta fields. Let's continue to do this.


douglaz (approved):
LGTM

PR #2877, opened 2023-08-01
fix: Symlink FM_TEST_DIR so they're easier to find

justinmoon:
Before this PR, if you wanted to inspect say `config.json` while running `just mprocs` you would need to enter user shell, and run something like `echo $FM_TEST_DIR` to find the path containing configs and logs.

Now you can just run `ls target/devimint/`:

```
$ ls target/devimint/
bitcoin/  cfg/  cln/  electrs/  env  esplora/  gw-cln/  gw-lnd/  lnd/  logs/
```

Should we add a new gitignored path for this? Or just re-use `target/` like I do here?

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -25,6 +25,10 @@ echo "Setting up env variables in $FM_TEST_DIR"
 mkdir -p "$FM_TEST_DIR"
 touch "$FM_PID_FILE"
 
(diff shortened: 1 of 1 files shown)


dpc (approved):

m1sterc001guy (approved):
Looks good to me, I'm ok with reusing `target/`

Issue #1477, opened 2023-01-22
Derive all client secrets from `Client`'s root secret


justinmoon:
If a client fails or is lost for whatever reason while waiting for confirmations to execute a peg-in, then you won't be able to complete the peg-in unless you saved a backup of your client db which contains the "tweak" used to create the deposit address from federation's public descriptor.

Now that the client has a [root secret](https://github.com/fedimint/fedimint/blob/5aefd0b57b6e38a109d8a6428ee44524a73a9cd2/client/client-lib/src/lib.rs#L150) we should consider deriving tweaks from this secret.

One question would be how to reset the tweak derivation index when recovering a client.

elsirion:
> One question would be how to reset the tweak derivation index when recovering a client.

Wouldn't it be similar to the e-cash module? You have a gap limit and scan that many addresses beyond the last one that had some activity. Once you found the last generated address that way you can set the next derivation index to that plus one.

justinmoon:
> Wouldn't it be similar to the e-cash module? 

Yes. But we can query the federation, right? Don't want to query bitcoin blockchain ...

justinmoon:
While we're at it, we should do `OutgoingContract::user_key` and lightning invoice preimages, too.

elsirion:
> Yes. But we can query the federation, right? Don't want to query bitcoin blockchain ...

I don't see a way around that, but it wouldn't be via `bitcoind`. The electrum protocol has the right API to retrieve the status of a lot of addresses efficiently.

justinmoon:
I guess this doesn't have to be a trusted electrum server, right? If it lies to you then the worst that happens is basically DOS?

elsirion:
I opened #1481 with a higher level description of the problem.

justinmoon:
Changed the title of this issue to cover just deriving client keys from the root secret. https://github.com/fedimint/fedimint/issues/1481 can be about the backup side of things.

justinmoon:
Remaining
- [ ] `payment_keypair` in `LightningClientModule::create_lightning_receive_output` ([link](https://github.com/fedimint/fedimint/blob/02e5d0e7c9f13a7be543ad9477fc6f53e364b8f7/modules/fedimint-ln-client/src/lib.rs#L865))
[...]

Issue #1481, opened 2023-01-22
Generalized module client backup API


elsirion:
Currently we only support backup/restore for the e-cash module, while it is the most at-risk for loss of funds, other modules also hold money for users for shorter amounts of time.

* **Wallet module client:** Deposits are managed by the wallet client till they are redeemed for e-cash. By deriving the tweaks used to generate peg-in addresses deterministically we could scan the BTC blockchain for activity on theses addresses (e.g. via the electrum protocol) and recover lost in-flight deposits that way.
* **LN module**: While paying or receiving money via LN funds are locked into smart contracts. For backupability the claim keys for these contracts need to be derived deterministically. For outgoing contracts the epoch history can be searched for ones with a claim key derived from the root secret (user+gw). For incoming ones the gw can do the same while the user has to wait on any contracts with a pending preimage since the preimage contains the claim key.

Ideally we can build a generalized API that allows the client to query module clients for data they want to back up regularly and also to restore their DBs from such backups and epoch histroy (see [backup_and_recovery.md](https://github.com/fedimint/fedimint/blob/master/docs/backup_and_recovery.md) for context ). This is related to client modularization #1447.

elsirion:
Sup of #2977

PR #2878, opened 2023-08-01
feat(lightning-client): deterministically derive lightning keys

m1sterc001guy:
Related PR: https://github.com/fedimint/fedimint/pull/2852

Issue: https://github.com/fedimint/fedimint/issues/1477

This PR deterministically derives the keys used on the client side for the lightning module.

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -323,8 +323,12 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
     // This test can't tolerate "spend", but not "reissue"d coins currently,
     // and there's a no clean way to do `reissue` on `spend` output ATM
     // so just putting it here for time being.
-    cli_tests_backup_and_restore(&fed).await?;
+    //cli_tests_backup_and_restore(&fed).await?;
 
+    fed.use_gateway(&gw_cln).await?;
+    cli_tests_lightning_backup_and_restore(&fed, &lnd).await?;
+
+    /*
     // # Spend from client
     info!("Testing spending from client");
     let notes = cmd!(fed, "spend", CLIENT_SPEND_AMOUNT)
@@ -656,6 +660,7 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
     let expected_wallet_balance = initial_walletng_balance - 5_000_000 - (fees_sat * 1000);
 
     assert_eq!(post_withdraw_walletng_balance, expected_wallet_balance);
+    */
 
     Ok(())
 }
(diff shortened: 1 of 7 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -857,16 +889,22 @@ impl LightningClientModule {
...
+            .module_root_secret
+            .child_key(LIGHTNING_INCOMING_CHILD_ID)
+            .child_key(Self::get_next_incoming_contract_child_id(dbtx).await)
+            .to_secp_key(&self.secp);
+        // Use the public key of the `payment_keypair` as the preimage for the invoice.
         let preimage: [u8; 32] = payment_keypair.x_only_public_key().0.serialize();
         let payment_hash = bitcoin::secp256k1::hashes::sha256::Hash::hash(&preimage);
 
-        // Temporary lightning node pubkey
+        // Temporary lightning node pubkey - does not need to be deterministic
```
m1sterc001guy, line 907: Since this key just signs the invoice and doesn't handle anything with funds, I don't think this needs to be deterministic. LMK if I misinterpreted this.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -846,6 +870,14 @@ impl LightningClientModule {
         }
     }
 
+    async fn get_next_incoming_contract_child_id(
```
m1sterc001guy, line 873: Similar to the wallet tweak key, I decided to keep an index for the `payment_keypair` since the public key of this is used as the preimage to the invoice, but this is also used to spend from the contract. We don't want to use the same key here since we want a unique preimage for each contract.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -550,9 +571,12 @@ impl ClientModuleGen for LightningClientGen {
         Ok(LightningClientModule {
             cfg,
             notifier,
-            redeem_key: module_root_secret.child_key(ChildId(0)).to_secp_key(&secp),
+            outgoing_redeem_key: module_root_secret
+                .child_key(LIGHTNING_OUTGOING_CHILD_ID)
+                .to_secp_key(&secp),
```
elsirion, line 576: That's a gateway-only thing, right? Unrelated to this PR but should be documented in the struct.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -664,7 +688,7 @@ impl LightningClientModule {
         let contract_amount_msat = invoice_amount_msat + base_fee + margin_fee;
         let contract_amount = Amount::from_msats(contract_amount_msat);
 
-        let user_sk = bitcoin::KeyPair::new(&self.secp, &mut rng);
+        let user_sk = self.outgoing_redeem_key;
```
elsirion, line 691: This is a huge privacy regression. A new key should be derived each time.

elsirion (changes requested):
The privacy regression needs a fix, likely a second key index or by adding a "key type" field to the current DB key.

justinmoon:
Note: there will also be a privacy regression whenever users recovery until we are able to recover derivation indices during recovery: https://github.com/fedimint/fedimint/issues/1481. I think that's fine as long as we can fix the recovery side of this pretty soon.

elsirion:
@justinmoon Indeed, even worse it might burn money in some cases :grimacing: I guess implementing recovery just became a high prio, good observation!

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -312,41 +314,60 @@ impl LightningClientExt for Client {
...
+                        // invoice will not be able to be paid
+                        self.transaction_updates(operation_id)
+                            .await
+                            .await_tx_accepted(txid)
+                            .await
+                            .map_err(|e| {
+                                anyhow::anyhow!("Offer transaction was not accepted: {e:?}")
+                            })?;
+
+                        Ok((operation_id, invoice))
```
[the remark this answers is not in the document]
m1sterc001guy: Moved outside of `autocommit`

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Which part is gateway only?

This key has been removed to fix the privacy regression. A new redeem key is generated each time an outgoing contract is created, so this isn't needed anymore.

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: This should be fixed now. Similar to incoming contracts, each time an outgoing contract is created, we increment an index that is saved to the database so that we use a different key per outgoing contract.

m1sterc001guy:
@dpc do we have backup/recovery tests for the lightning module?

justinmoon:
> @dpc do we have backup/recovery tests for the lightning module?

I think we can add to [this test](https://github.com/fedimint/fedimint/blob/b1b8d3bcc3ea45259e795a8c09871a22a65a82c8/devimint/src/main.rs#L723)

We could:
1. generate an invoice before wiping
2. wipe the client, 
3. pay the invoice
4. recover the client
5. assert we received the funds from payment of (1)

I'm guessing the `restore()` function should run the receive state machine automatically in (4)?

m1sterc001guy:
Making draft until backup/restore is implemented

devimint/src/main.rs
```diff
@@ -323,8 +323,12 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
     // This test can't tolerate "spend", but not "reissue"d coins currently,
     // and there's a no clean way to do `reissue` on `spend` output ATM
     // so just putting it here for time being.
-    cli_tests_backup_and_restore(&fed).await?;
+    //cli_tests_backup_and_restore(&fed).await?;
```
dpc, line 326: You should be able to use client forking to run both/all.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -602,6 +653,86 @@ impl ClientModule for LightningClientModule {
...
+                &NextIncomingContractIndexKey,
+                &snapshot.next_incoming_contract_index,
+            );
+            module_dbtx.insert_new_entry(
+                &NextOutgoingContractIndexKey,
+                &snapshot.next_outgoing_contract_index,
+            );
+        }
+
+        Ok(())
```
dpc, line 722: This doesn't not attempt to "rewind the history" from the point of snapshot to the present. Is it planned&possible?

m1sterc001guy: In progress :)

m1sterc001guy:
Any opposition to closing this? This work is dependent on the generalized backup/recovery mechanism and I suspect once that's in place, not much of this PR will be relevant anymore anyway.

justinmoon:
Sounds good. We can always cherry-pick commits or re-open if needed.

PR #2880, opened 2023-08-01
feat: secp public key setup for atomic broadcast

joschisan:


fedimint-server/src/config/mod.rs
```diff
@@ -366,6 +376,11 @@ impl ServerConfig {
         let private = self.private.clone();
         let id = identity.to_usize();
 
+        let my_public_key = private.broadcast_secret_key.public_key(&Secp256k1::new());
```
elsirion, line 379: Creating a new secp context every time is a bit of an anti-pattern. Not sure if using the [global context](https://docs.rs/secp256k1/0.27.0/secp256k1/static.SECP256K1.html) would be worse though since we are dealing with secret data here.

fedimint-server/src/config/mod.rs
```diff
@@ -403,6 +418,14 @@ impl ServerConfig {
         let mut rng = OsRng;
         let peer0 = &params[&PeerId::from(0)];
 
+        let mut broadcast_pks = BTreeMap::new();
+        let mut broadcast_sks = BTreeMap::new();
+        for peer_id in peer0.peer_ids() {
+            let (broadcast_sk, broadcast_pk) = secp256k1_zkp::generate_keypair(&mut OsRng);
+            broadcast_pks.insert(peer_id, broadcast_pk);
+            broadcast_sks.insert(peer_id, broadcast_sk);
+        }
```
elsirion, line 427: This code is perfectly fine, just FYI that this could also be achieved with less mutable state (didn't compile it, just a sketch):

```rust
let (broadcast_pks, broadcast_sks) = peer0.peer_ids()
    .into_iter()
    .map(|peer_id| {
        let (broadcast_pk, broadcast_sk) = secp256k1_zkp::generate_keypair(&mut OsRng);
        ((peer_id, broadcast_pk), (peer_id, broadcast_sk))
    })
    .unzip::<BTreeMap<_, _>, BTreeMap<_, _>, _, _>();
```

PR #2881, opened 2023-08-01
refine the atomic broadcast before switch

joschisan:


fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -1,155 +1,175 @@
...
+                        // running a session
+                        signed_block_sender.try_send(signed_block).ok();
+                    }
                 }
             }
-            Message::Block(signed_block) => {
-                signed_block_sender.send(signed_block).await.ok();
-            }
+
+            std::future::pending().await
```
elsirion, line 66: nit: why not return instead of blocking forever? We might also want to log that we exited the relay loop.

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -1,155 +1,175 @@
...
+    /// all items ordered in this session and a supersequence of the accepted
+    /// items. The end of a session is signaled by the return of Some(None)
+    /// while the return of None directly signals that the session has been
+    /// interrupted, either by a call to shutdown or by dropping a
+    /// decision_sender without sending a decision.
+    pub async fn run_session(
+        &self,
+        index: u64,
+    ) -> mpsc::Receiver<Option<(OrderedItem, oneshot::Sender<Decision>)>> {
+        let (ordered_item_sender, ordered_item_receiver) = mpsc::channel(256);
```
elsirion, line 90: 1. What happens if this channel fills up? Is there any risk of getting stuck, thus the high limit?
2. This seems to allow AlephBFT to be 256 items ahead of Fedimint consensus, which might lead to a lot of outdated contributions (for the ones that depend on previous consensus items).

This channel size seems like something we want to expose for Fedimint to configure, much like [`max_future_epochs` in HBBFT](https://docs.rs/hbbft/0.1.1/hbbft/honey_badger/struct.Params.html#structfield.max_future_epochs).

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -1,155 +1,175 @@
...
-
-                return Shutdown::MidSession(index);
-            }
+                for decision_receiver in decision_receivers {
+                    // The items in a threshold signed block have to be accepted
+                    if let Ok(decision) = decision_receiver.await {
+                        assert!(decision == Decision::Accept)
+                    }
+                }
+            });
```
elsirion, line 118: How is this shut down when fedimint is shut down ([`TaskGroup`s](https://docs.fedimint.org/fedimint_core/task/struct.TaskGroup.html) would help here)?

fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -1,155 +1,175 @@
...
-            }
-            Message::BlockRequest(index) => {
-                if let Some(signed_block) = db::load_block(&db, index).await {
-                    outgoing_message_sender
-                        .send((Message::Block(signed_block), Recipient::Peer(peer_id)))
-                        .await
-                        .ok();
+    network_data_receiver: async_channel::Receiver<Vec<u8>>,
+    signed_block_receiver: async_channel::Receiver<SignedBlock>,
+    relay_handle: JoinHandle<()>,
```
elsirion, line 16: Task groups might help with managing this task too

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: If we would return here the current aleph session would shut down as the senders get dropped. We only want to shut it down though if the ordered_item_receiver is dropped.

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: No there is no risk of getting stuck. The reason this channel is not of capacity 1 is to prevent switching tasks for every item under load. So this value is somewhat random, we could make it smaller. Any specific suggestions?

fedimint-atomic-broadcast/src/broadcast.rs
joschisan: Probably, for the first broadcast implementation we once decided to go without them though.

fedimint-atomic-broadcast/src/broadcast.rs
elsirion: Makes sense. How many items do we expect at most per second right now? That would sound like a reasonable number (task switching cost is relatively low compared to threads and even threads do ok with time slices on the order of a few 100ms).

Not urgent, can be fixed later.

elsirion (approved):

PR #2882, opened 2023-08-01
chore: rename WsClientConenctionInfo and related params, methods, comments, etc

okjodom:
rename clarifies what the code is and what it's used for

### DIFF
--- a/devimint/src/bin/faucet.rs
+++ b/devimint/src/bin/faucet.rs
@@ -21,8 +21,8 @@ struct Cmd {
     bitcoind_rpc: String,
     #[clap(long, env = "FM_CLN_SOCKET")]
     cln_socket: String,
-    #[clap(long, env = "FM_CONNECT_STRING")]
-    connect_string: Option<String>,
+    #[clap(long, env = "FM_INVITE_CODE")]
+    invite_code: Option<String>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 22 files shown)


devimint/src/bin/faucet.rs
```diff
@@ -22,7 +22,7 @@ struct Cmd {
     #[clap(long, env = "FM_CLN_SOCKET")]
     cln_socket: String,
     #[clap(long, env = "FM_CONNECT_STRING")]
-    connect_string: Option<String>,
+    invite_string: Option<String>,
```
justinmoon, line 25: Can we always "invite code"? Having both "invite code" and "invite string" just invites confusion. Also, can you update the `FM_CONNECT_STRING` to `FM_INVITE_CODE`? Also, we have a `client-connect` file that save the invite code in devimint. Can we call that `invite-code`? Let's make sure we change it everywhere.

fedimint-cli/src/lib.rs
```diff
@@ -361,7 +361,7 @@ enum Command {
 
     /// Join a federation using it's ConnectInfo
     JoinFederation {
-        connect: String,
+        invite: String,
```
justinmoon, line 364: Could we call this `invite_code`?

devimint/src/bin/faucet.rs
wbobeirne: Fedimint only ever calls this `connect_string`, I don't see any results for `connect_code` or `invite_code`
Screenshot 2023-08-01 at 4 28 53 PM

okjodom:
think I got em all

justinmoon:
Seems like connection string encoding might be broken? Re-running CI in case but this sounds like it might not be a flake. https://github.com/fedimint/fedimint/actions/runs/5732408213/job/15542810909?pr=2882#step:9:245

devimint/src/bin/faucet.rs
```diff
@@ -116,7 +116,7 @@ async fn main() -> anyhow::Result<()> {
         .route(
             "/connect-string",
```
okjodom, line 117: TODO: rename faucet route to `/invite-code`?

justinmoon: Rename it everywhere!

justinmoon:
Would be good to fix that other one in follow-on pr.

justinmoon (approved):

PR #2890, opened 2023-08-02
chore: Minimize disk usage in CI to avoid running out of space

douglaz:
This reduces ~20GB of disk usage during build/test execution

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,8 +85,12 @@ threshold_crypto = { opt-level = 3 }
 
 [profile.ci]
 inherits = "dev"
-split-debuginfo = "packed"
-debug = 1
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -85,8 +85,12 @@ threshold_crypto = { opt-level = 3 }
 
 [profile.ci]
 inherits = "dev"
-split-debuginfo = "packed"
-debug = 1
+# Minimize disk usage in CI to avoid running out of space
+incremental = false
+debug = 0
```
dpc, line 90: But without it, if something panics/errors we don't get stacktraces, which is very handy when you're investigating some random flaky test.

douglaz:
Closed in favor of https://github.com/fedimint/fedimint/pull/2893

PR #2891, opened 2023-08-02
fix(client): task management issues

dpc:
See commit messages.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -480,14 +480,16 @@ impl Drop for Client {
         // client reference
         if client_count == 1 {
             info!("Last client reference dropped, shutting down client task group");
-            futures::executor::block_on(async {
-                let Some(tg) = self.inner.tg.lock().await.take() else {
-                    return;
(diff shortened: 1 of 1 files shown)


fedimint-client/src/lib.rs
```diff
@@ -508,7 +508,8 @@ impl Client {
         ClientBuilder::default()
     }
 
-    pub async fn start_executor(&self, mut tg: TaskGroup) {
+    pub async fn start_executor(&self, tg: TaskGroup) {
```
elsirion, line 513: `tg` can be a reference if we only create a subgroup.

elsirion (approved):

elsirion:
WASM CI seems very unhappy, apparently `block_in_place` is only available in multi-threaded executors (which kinda makes sense …). Maybe stopping a client needs to be an explicit function and cannot be done on `Drop`?

Screenshot 2023-08-03 at 08-29-58 block_in_place in tokio task - Rust

elsirion:
#2898 got merged

PR #2898, opened 2023-08-03
fix(client): make client independent on task groups

elsirion:
Possible alternative to #2891.


fedimint-client/src/sm/executor.rs
```diff
@@ -501,6 +548,28 @@ where
     }
 }
 
+impl<GC> ExecutorInner<GC> {
+    /// See [`Executor::stop_executor`].
+    fn stop_executor(&self) -> Option<oneshot::Receiver<()>> {
+        let Some(shutdown_sender) = self.shutdown_executor
+            .try_lock()
```
elsirion, line 555: I could make it blocking but I'd rather detect possible deadlocks.

dpc (approved):

PR #2900, opened 2023-08-03
fix: we should connect without delay

douglaz:


elsirion (approved):
Right, we can do this now since only one side tries to connect.

PR #2901, opened 2023-08-03
chore: workaround test flakiness until a proper fix

douglaz:

PR #2903, opened 2023-08-03
fix(core): fix mock network

elsirion:
Our network stack assumes that broken connections

 a) are reliable or break completely
 b) broken connections are noticed by both sides eventually

 So far the mock network has not conformed to these requirements, causing flaky tests.

Alternative to #2901

### DIFF
--- a/fedimint-server/src/net/connect.rs
+++ b/fedimint-server/src/net/connect.rs
@@ -256,7 +256,7 @@ pub mod mock {
     use std::sync::Arc;
     use std::time::Duration;
 
-    use anyhow::Error;
+    use anyhow::{anyhow, Error};
     use fedimint_core::{task, PeerId};
     use futures::{pin_mut, FutureExt, SinkExt, Stream, StreamExt};
     use rand::Rng;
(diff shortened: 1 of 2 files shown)


douglaz (approved):
LGTM

dpc (dismissed):

elsirion:
Apparently not a full fix :sob:  https://github.com/fedimint/fedimint/actions/runs/5760106649/job/15615415492

dpc:
It seems to me that without heartbeats/retries it all might be broken even on a real setup.

When you complete `write` on a socket, it only means that the bytes were successfully buffered for sending out. Even if kernel detects the connection is no longer working and ACK packets are not being delivered, the app on neither side will know about it, until they try `write` again and get some error return value.

So a test that sends out stuff and then just waits, might never trigger it.

Seems like the network stack should have a timeout that it resets on every sent message (5s or so) and if it passes, it should re-send the first unconfirmed message (if any) to give itself chance to detect broken socket.

@elsirion @douglaz

elsirion:
I just polished some code I had already written while debugging myself. It adds ping messages every 10s if no other messages were sent. Besides fixing the detection of broken connections it also prevents the re-sending of a lot of messages on reconnect if only one side was sending (so the other never got to ACK).

elsirion:
I did 35 successful runs without any failures of `test_connect` now, seems good. Let's see if CI likes it.

dpc (approved):

PR #2904, opened 2023-08-03
fix: skip test_gateway_client_pay_invalid_invoice without mocks

dpc:
This test relies on mock LND node that treats invoices with `INVALID` description as invalid.

How is it even working right now is unclear.

image


okjodom (approved):
🤦🏿

elsirion:
I agree that the documentation is misleading, but this invoice is doubly not payable by a real LN node:

1. The (recipient) node id is a randomly generated key, so there will be no route to that node.
2. The payment hash/preimage is hard-coded and not known to any of the real LN nodes

The `INVALID_INVOICE_DESCRIPTION` hack was only necessary because the mock LN implementation cannot detect the first two problems with it, so we helped it a bit (thinking about it, we could also have matched the payment hash instead).

So this test should be able to run in both mock and real scenarios. What was the reason to deactivate it? If it's flakiness I'm pretty sure the problem is only masked by deactivating the test.

Issue #2906, opened 2023-08-04
Reactivate `test_gateway_client_pay_invalid_invoice` for real LN tests


elsirion:
> I agree that the documentation is misleading, but this invoice is doubly not payable by a real LN node:
>
> 1. The (recipient) node id is a randomly generated key, so there will be no route to that node.
> 2. The payment hash/preimage is hard-coded and not known to any of the real LN nodes
>
> The `INVALID_INVOICE_DESCRIPTION` hack was only necessary because the mock LN implementation cannot detect the first two problems with it, so we helped it a bit (thinking about it, we could also have matched the payment hash instead).
> 
> So this test should be able to run in both mock and real scenarios. What was the reason to deactivate it? If it's flakiness I'm pretty sure the problem is only masked by deactivating the test.

_Originally posted by @elsirion in https://github.com/fedimint/fedimint/issues/2904#issuecomment-1665166880_

justinmoon:
It's a confusing test because there is nothing "invalid" about the invoice. It just can't be paid. Should rename test and add a comment about what's going on.

PR #2907, opened 2023-08-04
Reactivate test_gateway_client_pay_invalid_invoice and improve documentation

justinmoon:
Closes https://github.com/fedimint/fedimint/issues/2906

### DIFF
--- a/fedimint-testing/src/ln/mod.rs
+++ b/fedimint-testing/src/ln/mod.rs
@@ -26,7 +26,10 @@ pub trait LightningTest: ILnRpcClient {
     ) -> ln_gateway::Result<Invoice>;
 
     /// Creates an invoice that is not payable
-    fn invalid_invoice(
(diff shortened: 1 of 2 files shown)


fedimint-testing/src/ln/mod.rs
```diff
@@ -38,6 +41,7 @@ pub trait LightningTest: ILnRpcClient {
         // `FakeLightningTest` will fail to pay any invoice with
         // `INVALID_INVOICE_DESCRIPTION` in the description of the invoice.
         Ok(InvoiceBuilder::new(Currency::Regtest)
+            .payee_pub_key(kp.public_key())
```
justinmoon, line 44: This will make it a little more obvious that invalid node pubkey is used. I know that's already the case, but this makes it a little more clear.

elsirion (approved):

PR #2910, opened 2023-08-07
chore: improve error message when refund transaction fails to be accepted

m1sterc001guy:
I have noticed `test_gateway_cannot_pay_expired_invoice` fail sometimes because the refund transaction is not accepted by the federation: https://github.com/fedimint/fedimint/actions/runs/5777285019/job/15657290448#step:9:6985

This PR just modifies the error message a bit to help with debugging.

@elsirion @joschisan any ideas on what might cause a refund transaction to not be accepted?


dpc (approved):

elsirion:
I guess this turned out to be related to #2912?

PR #2913, opened 2023-08-07
chore: Make MintMetaVariants public

justinmoon:
When integrating with the client, it is helpful for this to be public


m1sterc001guy (dismissed):

justinmoon:
Pushed update to make some `MintMeta` fields pub as well

m1sterc001guy (approved):

PR #2914, opened 2023-08-07
chore: make fedimint-ln-client::db public

justinmoon:
When integrating with the client, it's nice to have access to db structs.


m1sterc001guy:
> When integrating with the client, it's nice to have access to db structs.

For reading purposes right?

m1sterc001guy (approved):

PR #2916, opened 2023-08-07
chore: don't use `Task::on_shutdown` for `Task::make_shutdown_rx`

dpc:
It makes it easy to leak memory.

Instead, use `tokio::sync::watch` to update tasks that might be waiting for shutdown signal.

### DIFF
--- a/fedimint-core/src/task.rs
+++ b/fedimint-core/src/task.rs
@@ -3,8 +3,6 @@
 use std::collections::VecDeque;
 use std::future::Future;
 use std::pin::Pin;
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::Ordering::SeqCst;
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime};
 
@@ -12,12 +10,12 @@ use fedimint_core::time::now;
 use fedimint_logging::LOG_TASK;
 #[cfg(target_family = "wasm")]
 use futures::channel::oneshot;
-use futures::future::BoxFuture;
 use futures::lock::Mutex;
 pub use imp::*;
 use thiserror::Error;
 #[cfg(not(target_family = "wasm"))]
 use tokio::sync::oneshot;
+use tokio::sync::watch;
 #[cfg(not(target_family = "wasm"))]
 use tokio::task::{JoinError, JoinHandle};
 use tracing::{error, info, warn};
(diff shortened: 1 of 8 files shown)


douglaz:
Tests are more flaky than normal?

fedimint-core/src/task.rs
```diff
@@ -379,16 +398,34 @@ impl TaskHandle {
...
+                }
+
+                if *self.0.borrow() {
+                    break;
+                }
+            }
+        };
 
-        shutdown_rx
+        Box::pin(inner).as_mut().poll(cx)
```
elsirion, line 428: We can save one heap allocation here:

```suggestion
        futures::pin_mut!(inner);
        inner.poll(cx)
```

fedimint-core/src/task.rs
```diff
@@ -31,23 +32,41 @@ type JoinError = anyhow::Error;
...
-#[derive(Debug, Default)]
+#[derive(Debug)]
 struct TaskGroupInner {
     /// Was the shutdown requested, either externally or due to any task
     /// failure?
     is_shutting_down: AtomicBool,
     #[allow(clippy::type_complexity)]
     on_shutdown: Mutex<Vec<Box<dyn FnOnce() -> BoxFuture<'static, ()> + Send + 'static>>>,
+    on_shutdown_tx: watch::Sender<bool>,
+    on_shutdown_rx: watch::Receiver<bool>,
```
elsirion, line 43: Why keep the `rx` side around? It can be re-created using [`Sender::subscribe`](https://docs.rs/tokio/1.29.1/tokio/sync/watch/struct.Sender.html#method.subscribe) at any time.

fedimint-core/src/task.rs
```diff
@@ -31,23 +32,41 @@ type JoinError = anyhow::Error;
 #[error("deadline has elapsed")]
 pub struct Elapsed;
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 struct TaskGroupInner {
     /// Was the shutdown requested, either externally or due to any task
     /// failure?
     is_shutting_down: AtomicBool,
```
elsirion, line 39: `is_shutting_down` is redundant now. We can use `*on_shutdown_tx.borrow()` instead.

fedimint-core/src/task.rs
```diff
@@ -379,16 +398,34 @@ impl TaskHandle {
...
+            loop {
+                if self.0.changed().await.is_err() {
+                    // if disconnected, shutdown
+                    break;
+                }
+
+                if *self.0.borrow() {
+                    break;
+                }
+            }
```
elsirion, line 425: Isn't that pretty much what [`Receiver::wait_for`](https://docs.rs/tokio/1.29.1/tokio/sync/watch/struct.Receiver.html#method.wait_for) would do if given the identity function as its predicate?

elsirion (commented):
Great improvement, just a few nits :)

fedimint-core/src/task.rs
dpc: The whole thing is buggy. I can't be recreating that inner future on every call. Do you think there's still a way to avoid Box?

douglaz (approved):
Much better than previous version. LGTM

PR #2918, opened 2023-08-08
2836 improve tutorial in our docs

EthnTuttle:


docs/dev-running.md
```diff
@@ -16,7 +16,7 @@ This uses a tool called [mprocs](https://github.com/pvolok/mprocs) to spawn work
 
 ![screenshot of the federation running in mprocs](mprocs.png)
 
-### Using the client
+### Using the `fedimint-cli client`
```
m1sterc001guy, line 19: Did you forget a backtick here?

docs/dev-running.md
```diff
@@ -83,9 +83,69 @@ $ fedimint-cli reissue BgAAAAAAAAAgAAAAAAAAAAEAAAAAAAAAwdt...
...
+    "64": 2,
+    "65536": 2,
+    "8": 2,
+    "8192": 2
+  },
+  "total_msat": 10000000
+}
+```
+
+To deposit into a Fedimint, you must generate a new deposit address, funds sent to it can later be claimed
```
m1sterc001guy, line 124: Not sure how to properly communicate this, but funds sent to an onchain address are automatically claimed by the client state machine framework in the background. The user doesn't need to do anything

docs/dev-running.md
```diff
@@ -83,9 +83,69 @@ $ fedimint-cli reissue BgAAAAAAAAAgAAAAAAAAAAEAAAAAAAAAwdt...
...
+```
+
+Save the `operation_id` somewhere accessible, we'll need that later to claim our ecash notes. The `operation_id` is a key that associates a client with the relevant [transaction](`https://github.com/fedimint/fedimint/blob/0f6096b3bb5961a1697e4591b4d2526629b13c2d/fedimint-core/src/transaction.rs#L19`).
+Use the `"address"` field with whatever on-chain capable wallet you use. When using the development environment, you can use `bitcoin-cli sendtoaddress bcrt1qzwpac0pfk8vvdye6anyhtvm359485d3hezpvwfhheqdz4yv9yx6qsn2u06 <amount>` to send bitcoin. Then `bitcoin-cli -regtest -generate 10` to mine some blocks.
+
+Now use `await-deposit` to claim your ecash. Your ecash balance should be increased by the amount sent. Use `fedmint-cli info` to check the balance has increased by `<amount>`.
+
+```shell
+$ fedimint-cli await-deposit d156b6c420353113c9ad1f72652c0975e18dc8a589d5d026ea5c9caa673df693
+null
```
m1sterc001guy, line 141: null?

scripts/kill-ports.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Check for required commands
+for cmd in lsof rg awk xargs; do
```
m1sterc001guy, line 4: Do we install all of these tools using nix?

scripts/kill-ports.sh
```diff
@@ -0,0 +1,30 @@
...
+  exit 0
+fi
+
+# Confirm the action
+echo "The following processes will be killed:"
+echo "$processes"
+read -p "Are you sure you want to continue? (y/N) " choice
+
+if [[ $choice =~ ^[Yy]$ ]]; then
+  echo "$processes" | xargs kill -9
```
m1sterc001guy, line 26: If you use pkill instead of kill then you don't need to list the processes.

scripts/kill-ports.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo "Processes killed."
+else
+  echo "Operation canceled."
+fi
```
m1sterc001guy, line 30: Nit: Would probably be best to separate this script into a new PR from the documentation improvements

scripts/kill-ports.sh
EthnTuttle: Oops. this script wasn't supposed to be in here. Will rebase. devimint has been improved to fix this.
https://github.com/fedimint/fedimint/pull/2873

scripts/kill-ports.sh
EthnTuttle: 100%, this was just bad local branching. 😅

EthnTuttle:
Believe this to be OBE.

PR #2919, opened 2023-08-08
cleanup: simplify flagging of peers

joschisan:


elsirion:
Looks ok from my side, some high level comments:

* A lot of tests got removed, are you sure they didn't contain any interesting test cases?
* Removing timing information can make debugging slightly harder (e.g. have to find epoch in logs to know which wall clock time an epoch relates to)
* This might need coordination with the admin UI team to update what is shown there. Do you know if this is the case @EthnTuttle, @okjodom?

joschisan:
> * A lot of tests got removed, are you sure they didn't contain any interesting test cases?

The tests were mostly testing the flagging of peers which has been simplified so far now that testing it seemed unnecessary, removing the tests also enabled me to remove the indirection where we had to calculate the consensus state in a separate function.
> * Removing timing information can make debugging slightly harder
I removed the timing because it is subjective to a peer. A peer that has been offline for a while will perceive a contribution as younger than it actually is. When debugging the consensus it tells us much more in which epoch / session and item has been submitted.

fedimint-server/src/consensus/server.rs
```diff
@@ -607,22 +605,12 @@ impl ConsensusServer {
...
-            value: epoch + 1,
-            time: fedimint_core::time::now(),
-        };
+
         self.latest_contribution_by_peer
             .write()
             .await
-            .entry(peer)
-            .and_modify(|c| {
-                // only update if epoch count is higher
```
douglaz, line 620: I would keep this as a sanity check

fedimint-server/src/net/api.rs
```diff
@@ -264,21 +264,55 @@ impl ConsensusApi {
...
-            peers_connection_status,
-            MAX_DURATION_FOR_RECENT_CONTRIBUTION,
-        ))
+        let epoch_count = self.get_epoch_count().await;
+
+        let status_by_peer = peers_connection_status
+            .into_iter()
+            .map(|(peer, connection_status)| {
+                let last_contribution = latest_contribution_by_peer.get(&peer).cloned();
+                let flagged = last_contribution.unwrap_or(0) + 100 < epoch_count;
```
douglaz, line 275: 100 is a lot. it doesn't make sense for HBBFT. Does it make sense for Aleph?

fedimint-core/src/api.rs
```diff
@@ -968,10 +968,7 @@ impl<C: JsonRpcClient> WsFederationApi<C> {}
 /// The status of a server, including how it views its peers
 #[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ConsensusStatus {
-    /// The last contribution that this server has made to the consensus, it's
-    /// equivalent to [`PeerConsensusStatus::last_contribution`] and
-    /// [`ConsensusContribution::value`]
-    pub last_contribution: u64,
+    pub epoch_count: u64,
```
douglaz, line 971: `contribution` was used as a generic term to keep the api backwards-compatible even if we removed "epochs". Will Aleph call it `epoch_count`?

fedimint-server/src/net/api.rs
```diff
@@ -613,125 +577,11 @@ impl<T: Clone> ExpiringCache<T> {
...
 mod tests {
+    use std::time::Duration;
 
-    use fedimint_core::api::ConsensusContribution;
     use fedimint_core::task;
-    use fedimint_core::time::now;
-
-    use super::*;
-    #[test]
-    fn test_server_status_all_ok() {
```
douglaz, line 623: These tests still make sense, with minor modifications

fedimint-server/src/net/api.rs
joschisan: No for aleph it would be one as a session is already 5 min. What number do you prefer?

fedimint-core/src/api.rs
joschisan: I will rename it to block_count then.

fedimint-server/src/net/api.rs
joschisan: I switched it to 10, does that work?

fedimint-server/src/consensus/server.rs
joschisan: Done.

fedimint-server/src/net/api.rs
joschisan: Yes but without the flagging grace time the calculation of the ConsensusStatus is so simple now that tests seemed unnecessary and this allows me to remove the indirection of having code that is only called in one place in a seperate function for testing purposes only.

fedimint-server/src/net/api.rs
douglaz: On previous version this was zero. But I think one is okay.

fedimint-server/src/net/api.rs
douglaz: I don't see removing a function as a big win compared to the reduced test coverage. Anyway we can add the tests back later.

douglaz (approved):
LGTM

PR #2920, opened 2023-08-08
fix(server): logs don't claim an epoch was missing if it wasn't

elsirion:
This has been annoying me for a while, probably introduced by a refactor a few months ago.


m1sterc001guy (approved):
This has been confusing me for a bit too :)

justinmoon:
Thanks. I've been bothered by this as well.

Issue #2849, opened 2023-07-27
Make tools compatible with unknown modules


elsirion:
Now that #2299 is solved we should make tooling (dbtool, recoverytool, maybe fedimint-cli, …) compatible with unknown modules.

dpc:
Clarification: The #2299 enables parsing consensus history (well... consensus items) without having corresponding modules. Normal client operation without certain modules should have been possible for a while now:

https://github.com/fedimint/fedimint/blob/becd7cd9300a663f6ebba407dc2f43cf486bb83d/fedimint-client/src/lib.rs#L1362C10-L1362C11

So it should be mostly about db/recovery tools.

dpc:
Looking into it.

dpc:
I think it's been handled in most tools already, and core code changed to make graceful handling the default and natural way. Closing. We'll have to keep it in mind for time being, eventually add some tests (with time, as things mature).

justinmoon:
I haven't seen any issues with this recently 👍

PR #2927, opened 2023-08-09
 feat: dbtool handling of unknown modules

dpc:
Step towards #2849

Rather rough change, but I think I'm going to spend more time with that part of the code and send follow-ups.

```
just mprocs
# then ...
fedimint-dbtool --no-modules --database target/devimint/cfg/server-0/database dump --password pass --cfg-dir target/devimint/cfg/server-0 2>&1 | less
```


fedimint-core/src/core.rs
```diff
@@ -120,29 +120,50 @@ macro_rules! module_dyn_newtype_impl_encode_decode {
...
+                        v
+                    }
+                    None => $name::from_typed(
+                        module_instance_id,
+                        $crate::core::DynUnknown(Vec::<u8>::consensus_decode(
+                            reader,
+                            &Default::default(),
+                        )?),
+                    ),
+                };
```
elsirion: This change seems dangerous, we should double-check consensus for reliance on never encountering e.g. consensus items with unknown module instance IDs. Previously these would have been filtered out at the network level, now we'll have to do that later I assume.

Pinging @joschisan

dpc: Since we rely on `Decodable` on both client and server, I don't think we can avoid this type. Client needs to be able to decode dyn-newtype instances that it doesn't have.

On server this should never happen, but it's a good thing to verify - what attacks a malicious peer could do by injecting unknown dyn-newtypes here.

Notably this type fully validates and preserves information. There can be no ambiguity on encoding level. Every dyn-newtype is `module_instance_id, Vec<u8>`, no matter if it known or not.

elsirion: Sure, I'd just want to avoid suddenly having a way to trigger a panic in our consensus loop.

E.g. this code will panic due to `get_expected`:

https://github.com/fedimint/fedimint/blob/05a03c8f02a31ad5c72c434e5087ac2c0fd98219/fedimint-server/src/consensus/mod.rs#L241-L254

elsirion (changes requested):
As mentioned in https://github.com/fedimint/fedimint/pull/2927#discussion_r1290026107 this will create a DoS vulnerability triggerable by any guardian.

fedimint-core/src/core.rs
dpc: @elsirion Fixed.

elsirion (approved):

PR #2931, opened 2023-08-09
fix(client): notify about state machines being added

### LINKED ISSUE #2929: `LightningClientExt::pay_bolt11_invoice` doesn't trigger balance update stream

elsirion:
Fixes #2929.

We were not triggering notifications on *adding* state machines so far, only on successful state transition.

For most use cases that was sufficient since one would only subscribe to updates once one had started the state machine. But when adding balance subscriptions I didn't think about it because for spending the balance updates are triggered by the *first* state, which is added instead of reached through a transition.


fedimint-core/src/db/mod.rs
```diff
@@ -1030,6 +1030,7 @@ pub struct DatabaseTransaction<'a> {
     tx: Box<dyn ISingleUseDatabaseTransaction<'a>>,
     decoders: ModuleDecoderRegistry,
     commit_tracker: CommitTracker,
+    on_commit_hooks: Vec<Box<maybe_add_send!(dyn FnOnce())>>,
```
m1sterc001guy, line 1033: Would we want these commit hooks to return a `Result`?

m1sterc001guy (approved):

fedimint-core/src/db/mod.rs
elsirion: What would we do with the result? When the hooks get called we already committed the dbtx and if we were to run them before then they'd fire without knowing if the dbtx actually succeeded.

Issue #2926, opened 2023-08-09
fedimint-core/src/db/mod.rs looking sus


dpc:
See if it can be improved.

PR #2933, opened 2023-08-09
 chore: unify common db ops under one trait 

dpc:
Looking at #2926 . Some initial cleanup.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -83,26 +83,6 @@ pub trait DatabaseValue: Sized + Debug {
 
 pub type PrefixStream<'a> = Pin<Box<maybe_add_send!(dyn Stream<Item = (Vec<u8>, Vec<u8>)> + 'a)>>;
 
-#[apply(async_trait_maybe_send!)]
-pub trait IDatabase: Debug + MaybeSend + MaybeSync + 'static {
-    async fn begin_transaction<'a>(&'a self) -> Box<dyn ISingleUseDatabaseTransaction<'a>>;
-}
-
-#[derive(Clone, Debug)]
-pub struct Database {
-    inner_db: Arc<DatabaseInner<dyn IDatabase>>,
-    module_instance_id: Option<ModuleInstanceId>,
-}
-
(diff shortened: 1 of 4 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -339,6 +339,40 @@ impl Database {
     }
 }
 
+#[apply(async_trait_maybe_send!)]
+pub trait IDatabaseTransactionOps<'a>: 'a + MaybeSend {
```
dpc, line 343: Basically I extracted this, so this methods can't diverge between multiple traits.

fedimint-rocksdb/src/lib.rs
```diff
@@ -133,6 +135,26 @@ impl<'a> IDatabaseTransaction<'a> for RocksDbTransaction<'a> {
         }))
     }
 
+    async fn raw_remove_by_prefix(&mut self, key_prefix: &[u8]) -> anyhow::Result<()> {
+        fedimint_core::task::block_in_place(|| {
```
dpc, line 139: @m1sterc001guy Notably, I was forced to implement it.

elsirion (dismissed):
Great PR! I remember being annoyed about the duplication and not having the time to fix it.

elsirion:
Looking at CI there seems to be a bug though … :(

dpc:
@elsirion was missing range ending conditions in my implementation of delete_by_prefix for rocksdb.

fedimint-rocksdb/src/lib.rs
m1sterc001guy: We were relying on the default implementation before right?

m1sterc001guy: This looks good to me, as long as `verify_remove_by_prefix` is passing I think we're good

m1sterc001guy (approved):
LGTM

fedimint-rocksdb/src/lib.rs
dpc: Yes, but that default impl was meh (loading everything into memory at once), and now it doesn't work due to object-safety anyway.

PR #2934, opened 2023-08-09
fix: remove consensus from naming

joschisan:


elsirion (approved):

elsirion:
Seems we broke the UI repo, mea culpa, I wasn't aware or I would have raised that. https://github.com/fedimint/ui/pull/137

PR #2936, opened 2023-08-10
Fix race conditions in Query Strategies

joschisan:


m1sterc001guy:
@EthnTuttle this is the fix for the bug we were discussing earlier

fedimint-core/src/query.rs
```diff
@@ -80,16 +79,19 @@ pub struct UnionResponses<R> {
...
     existing_results: Vec<R>,
     current: CurrentConsensus<Vec<R>>,
-    required: usize,
+    threshold: usize,
 }
 
 impl<R> UnionResponses<R> {
-    pub fn new(required: usize) -> Self {
+    pub fn new(total_peers: usize) -> Self {
+        let max_evil = (total_peers - 1) / 3;
```
m1sterc001guy, line 87: Let's make this a simple helper function since you're using it below too.

fedimint-core/src/query.rs
```diff
@@ -80,16 +79,19 @@ pub struct UnionResponses<R> {
...
+        let threshold = total_peers - max_evil;
```
m1sterc001guy, line 88: So if I understand correctly, the reason we're getting rid of `EventuallyConsistent` and not using `one_honest` anymore is because `one_honest` is not safe and can lead to race conditions. Therefore, we actually always need to wait for `total - max_evil` to truly know if a transaction has been included in consensus or not. Correct?

fedimint-core/src/query.rs
```diff
@@ -163,100 +168,67 @@ impl<R: Debug + Eq + Clone> QueryStrategy<R, Vec<R>> for UnionResponsesSingle<R>
...
             }
         }
 
-        if self.errors.len() >= self.required {
-            return QueryStep::Failure {
-                general: None,
-                members: mem::take(&mut self.errors),
-            };
+        if self.responded_peers.len() >= self.threshold {
+            QueryStep::RetryMembers(mem::take(&mut self.responded_peers))
```
m1sterc001guy, line 228: I'm not super familiar with this part of the code. What will `RetryMembers` do? Is there some sort of backoff?

m1sterc001guy:
I tested this by running `test_gateway_cannot_pay_expired_invoice` 10 times in a row. Before this change, the test would consistently fail at least once in those 10 runs. With this fix I haven't seen it fail, so looks good to me.

Unclear to me if the `reconnect` test failure in CI is related to your changes though

EthnTuttle:
Is this current consensus matching surfaceable using the client API? I would imagine a client might want this information because staying online a bit longer, showing a loading spinner, before showing "done" would make a user pause before shutting down/locking their phone/browser.

fedimint-core/src/query.rs
joschisan: Yes, using one_honest instead of threshold leads to a race condition. Furthermore, the old CurrentConsensus did not implement retries if no quorum was reached which lead to a different race condition, hence we always use retries now just as EventuallyConsistent used to which is why we removed EventuallyConsistent as retries were the only difference to CurrentConsensus.

fedimint-core/src/query.rs
joschisan: If we return RetryMembers the query will call the endpoint again and call process on the result. this allows peers to update their value in order for us to achieve a quorum.

joschisan:
> Unclear to me if the `reconnect` test failure in CI is related to your changes though

It fails consistently though so it probably is.

fedimint-core/src/query.rs
m1sterc001guy: Is there a backoff and timeout?

fedimint-core/src/query.rs
```diff
@@ -163,100 +168,77 @@ impl<R: Debug + Eq + Clone> QueryStrategy<R, Vec<R>> for UnionResponsesSingle<R>
...
+    threshold: usize,
+    max_evil: usize,
 }
 
-impl<R: Eq + Clone + Debug> QueryStrategy<R> for EventuallyConsistent<R> {
-    fn process(&mut self, peer: PeerId, result: api::MemberResult<R>) -> QueryStep<R> {
-        self.responses.insert(peer);
+impl<R> CurrentConsensus<R> {
+    pub fn new(total_peers: usize) -> Self {
+        let max_evil = (total_peers - 1) / 3;
```
elsirion, line 183: Having an object implementing `NumPeers` in here would also avoid this code duplication, there'd just be a `max_evil` function to call.

fedimint-core/src/query.rs
```diff
@@ -163,100 +168,77 @@ impl<R: Debug + Eq + Clone> QueryStrategy<R, Vec<R>> for UnionResponsesSingle<R>
...
             }
         }
 
-        for (result, peers) in &self.existing_results {
-            if peers.len() >= self.required {
-                return QueryStep::Success(result.clone());
+        if let Some(response) = self
+            .responses
+            .values()
+            .max_by_key(|response| self.responses.values().filter(|r| r == response).count())
```
elsirion, line 228: Quadratic complexity, but for small numbers still reasonable.

fedimint-cli/src/lib.rs
```diff
@@ -631,7 +631,7 @@ impl FedimintCli {
                         .map_err_cli_general()?,
                     None => ws_api
                         .request_with_strategy(
-                            EventuallyConsistent::new(ws_api.peers().len()),
+                            CurrentConsensus::full_participation(ws_api.peers().len()),
```
elsirion, line 634: I think the problem lies here: the old eventually consistent afaik retried on errors as long as some good responses came back. This test has been flaky before, it's likely that we relied on some implicit retry behavior. Will try to replicate.

elsirion: Yep, looks if if that was the cause. I pushed the fix.

elsirion: Ok, *now* it should work.

fedimint-core/src/query.rs
joschisan: Yes there is a initial delay of 10ms that doubles on every retry to the same peer.

elsirion (approved):

PR #2940, opened 2023-08-11
fix: rocksdb still running blocking operations in async runtime

### LINKED ISSUE #2607: I think we're still running blocking IO in async code
```
    async fn raw_find_by_prefix_sorted_descending(
        &mut self,
        key_prefix: &[u8],
    ) -> Result<PrefixStream<'_>> {
        let prefix = key_prefix.to_vec();
        let next_prefix = next_prefix(&prefix);
        let iterator_mode = if let Some(next_prefix) = &next_prefix {
            rocksdb::IteratorMode::From(next_prefix, rocksdb::Direction::Reverse)
        } else {
            rocksdb::IteratorMode::End
        };
        Ok(fedimint_core::task::block_in_place(|| {
            let mut options = rocksdb::ReadOptions::default();
            options.set_iterate_range(rocksdb::PrefixRange(prefix.clone()));
            let iter = self.0.snapshot().iterator_opt(iterator_mode, options);
            let rocksdb_iter = iter
                .map_while(move |res| {
                    let (key_bytes, value_bytes) = res.expect("Error reading from RocksDb");
                    key_bytes
                        .starts_with(&prefix)
                        .then_some((key_bytes, value_bytes))
                })
                .map(|(key_bytes, value_bytes)| (key_bytes.to_vec(), value_bytes.to_vec()));
            Box::pin(stream::iter(rocksdb_iter))
        }))
    }
```

In this code a `block_in_place` is used in the hope of isolating blocking IO from async runtime, but what actually happens is that `rocksdb_iter` is being returned and that `rocksdb_iter` when polled later is calling blocking IO operations. What would be a point of an db iterator if it didn't fetch more data when ask for (instead of upfront).

This might lead to weird slowdown and latency issues. Noticed it when debugging why e.g. selecting notes for transaction input seems to be taking the gateway around 1s.

Ping @m1sterc001guy  :)

dpc:
Fix #2607

### DIFF
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -115,6 +115,21 @@ impl<'a> IDatabaseTransactionOps<'a> for RocksDbTransaction<'a> {
     }
 
     async fn raw_find_by_prefix(&mut self, key_prefix: &[u8]) -> Result<PrefixStream<'_>> {
+        // turn an `iter` into a `Stream` where every `next` is ran inside
+        // `block_in_place` to offload the blocking calls
+        fn convert_to_async_stream<'i, I>(iter: I) -> impl futures::Stream<Item = I::Item>
+        where
+            I: Iterator + Send + 'i,
+            I::Item: Send,
+        {
+            stream::unfold(iter, |mut iter| async move {
+                tokio::task::block_in_place(move || {
+                    let item = iter.next();
+                    item.map(move |item| (item, iter))
+                })
+            })
+        }
+
         Ok(fedimint_core::task::block_in_place(|| {
             let prefix = key_prefix.to_vec();
             let mut options = rocksdb::ReadOptions::default();
@@ -129,7 +144,7 @@ impl<'a> IDatabaseTransactionOps<'a> for RocksDbTransaction<'a> {
                     .starts_with(&prefix)
                     .then_some((key_bytes.to_vec(), value_bytes.to_vec()))
             });
-            Box::pin(stream::iter(rocksdb_iter))
+            Box::pin(convert_to_async_stream(rocksdb_iter))
         }))
     }
 


m1sterc001guy (approved):

fedimint-rocksdb/src/lib.rs
```diff
@@ -115,6 +115,21 @@ impl<'a> IDatabaseTransactionOps<'a> for RocksDbTransaction<'a> {
...
+        // `block_in_place` to offload the blocking calls
+        fn convert_to_async_stream<'i, I>(iter: I) -> impl futures::Stream<Item = I::Item>
+        where
+            I: Iterator + Send + 'i,
+            I::Item: Send,
+        {
+            stream::unfold(iter, |mut iter| async move {
+                tokio::task::block_in_place(move || {
+                    let item = iter.next();
+                    item.map(|item| (item, iter))
```
elsirion, line 128: Why doesn't this need to be a `move` closure? How is `iter` getting in there by value or isn't it? :open_mouth:

elsirion (dismissed):

fedimint-rocksdb/src/lib.rs
dpc: Yeah. I haven't noticed and I don't understand. Must be some auto-convenience thing, or something being `Copy` here?

Anyway, changed to `move` as it feels more ... proper.

justinmoon (approved):

fedimint-rocksdb/src/lib.rs
elsirion: TIL: apparently using something by-value in a closure moves it automatically and the `move` keyword is only useful when variables are used by-ref but one still wants them to move into the closure for lifetime issues.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d81e7e6dc74c49ddaeef283a63d20d01

PR #2943, opened 2023-08-12
chore: speed up commit check in CI

dpc:


elsirion:
Are you sure? It seems the Lint step still takes ~3min: https://github.com/fedimint/fedimint/actions/runs/5837950292/job/15834277032

dpc:
Yeah, I've noticed that the time it did help locally was probably about something else.

PR #2944, opened 2023-08-12
refactor: unify query strategies

joschisan:


fedimint-core/src/query.rs
```diff
@@ -62,11 +62,38 @@ impl<R: Debug + Eq + Clone> QueryStrategy<R> for VerifiableResponse<R> {
...
+        Self {
+            errors: BTreeMap::new(),
+            threshold,
+        }
+    }
+
+    pub fn process<R>(&mut self, peer: PeerId, error: MemberError) -> QueryStep<R> {
+        assert!(self.errors.insert(peer, error).is_none());
+
+        if self.errors.len() == self.threshold {
```
elsirion, line 109: If `threshold == 0` this could result in an endless loop, should the constructor just assert `threshold > 0` since 0 is nonsensical anyway?

elsirion (dismissed):
One nit

elsirion (approved):

fedimint-core/src/query.rs
joschisan: Yes

Issue #1397, opened 2023-01-18
Contract offers malleability


elsirion:
Currently we do not check if an offer for the same preimage already exists when a new offer is created. This can be exploited by the gateway by creating a second offer, copying encrypted preimage and payment hash but setting a lower amount required to buy it.

If we can assume that an encrypted preimage cannot be altered by an adversary to make its serialization different while still decrypting to useful data (big if) then we'd only need to check if the same encrypted preimage was posted in an offer previously (just add another DB entry indexing these encrypted preimages). 

Long term I'd like to explore more elegant ways of binding the amount to the encrypted preimage (possibly two-step encryption …).

justinmoon:
@elsirion Could you try describing this? I'm going to do some work on gateway next week or two so I'm shopping for issues :)

elsirion:
Done.

justinmoon:
Call: 
- Deploy version number and minimum viable fix

m1sterc001guy:
@joschisan can we close this since LNv2 fixes this?

joschisan:
@m1sterc001guy yes

PR #2947, opened 2023-08-13
fix(ln): don't allow same preimage for sale twice

elsirion:
Mitigation for #1397.

TODO:
- [x] Unit test


modules/fedimint-ln-common/src/lib.rs
```diff
@@ -365,6 +365,8 @@ pub enum LightningError {
     ZeroOutput,
     #[error("Offer contains invalid threshold-encrypted data")]
     InvalidEncryptedPreimage,
+    #[error("Offer contains a ciphertext that has already been usec")]
```
m1sterc001guy, line 368: Nit: typo

m1sterc001guy (approved):
Looks good to me. Could you explain why this is a mitigation and not a full fix? Is it because a gateway could collude with the guardians to post the encrypted preimage twice?

elsirion:
> why this is a mitigation and not a full fix

We use the threshold encryption scheme as a blackbox and assume some minimal set of guarantees like keeping the plaintext secret from an adversary. Something we don't know about the scheme is if there is a way for an adversary, given a ciphertext `c=Enc(p)`, to construct a `c'` that decrypts to a `p'=Dec(c')` such that the adversary can infer `p` from `p'`.

@joschisan already has a candidate scheme to fully fix the problem and also simplify the LN logic quite a bit at the same time (get rid of offers). But it still needs to be proven secure and implemented, which will probably take a bit.

justinmoon:
> @joschisan already has a candidate scheme

Looking forward to this!!!

PR #2950, opened 2023-08-14
chore: update `miniscript` to a released version

dpc:


modules/fedimint-wallet-common/src/txoproof.rs
```diff
@@ -130,14 +130,7 @@ impl Tweakable for Descriptor<CompressedPublicKey> {
...
 
-            fn pkh(
-                &mut self,
-                pkh: &CompressedPublicKey,
-            ) -> Result<CompressedPublicKey, Infallible> {
-                Ok(CompressedPublicKey::new(
-                    pkh.key.tweak(self.tweak, self.secp),
-                ))
-            }
+            translate_hash_clone!(CompressedPublicKey, bitcoin::PublicKey, Infallible);
```
dpc, line 133: @elsirion I'm confused about this one.

It feels like this trait wants me to be able to translate from a hash. But how am I supposed to recover key from the hash ... ? Or is the key the hash already and I just need to cast?

https://docs.rs/miniscript/latest/miniscript/trait.Translator.html

elsirion: No (not necessarily at least, but the API is ridiculously flexible …), you are supposed to be able to translate from one hash type to another. This can be useful if you were to work with two incompatible versions of `bitcoin_hashes` or third party hash libararies.

We are using the trait for something else though: to tweak keys instead of only changing their type/representation. So using `translate_hash_clone` isn't the right macro here afaik (and the old code would be buggy in this edge case too because I wasn't aware of the `impl Translator for T
where T: PkTranslator` default impl :frowning:). 

The right choice would be to fail imo, so [`translate_hash_fail!`](https://docs.rs/miniscript/latest/miniscript/macro.translate_hash_fail.html) since we don't have a defined semantic for what to do with hashes.

PR #2951, opened 2023-08-14
chore: Rust  toolchain update

dpc:
I was looking into some slowness in the CI, and it would be improved with sparse index support which was recently made stable

https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html#cargos-sparse-protocol

(+ made default in 1.70).

There's a separate PR for miniscript update, which seems to be required here as well: https://github.com/fedimint/fedimint/pull/2950

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,12 +28,18 @@
     flake-utils.lib.eachDefaultSystem
       (system:
         let
-          pkgs = import nixpkgs {
-            inherit system;
-          };
 
           pkgs-unstable = import nixpkgs-unstable {
             inherit system;
+          };
+
+          pkgs = import nixpkgs {
+            inherit system;
+            overlays = [
+              (final: prev: {
+                cargo-udeps = pkgs-unstable.cargo-udeps;
+              })
+            ];
           };
 
           pkgs-kitman = import nixpkgs-kitman {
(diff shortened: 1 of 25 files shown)


justinmoon (approved):

flake.nix
```diff
@@ -28,12 +28,18 @@
...
           };
 
-          pkgs-unstable = import nixpkgs-unstable {
+          pkgs = import nixpkgs {
             inherit system;
+            overlays = [
+              (final: prev: {
+                cargo-udeps = pkgs-unstable.cargo-udeps;
+              })
+            ];
```
elsirion, line 42: What's the benefit of using the unstable udeps version this way instead of directly?

devimint/src/lib.rs
```diff
@@ -141,7 +141,7 @@ impl Gatewayd {
 
     pub async fn connect_fed(&self, fed: &Federation) -> Result<()> {
         let invite_code = fed.invite_code()?;
-        poll_max_retries("gateway connect-fed", 10, || async {
+        poll_max_retries("gateway connect-fed", 60, || async {
```
elsirion, line 144: Is this related to #2971?

elsirion (approved):

flake.nix
dpc: Can't forget about using the right once. :D

devimint/src/lib.rs
dpc: Yes. Removing it as it was moved to another PR.

Issue #2134, opened 2023-04-05
Organize files under `./scripts`


dpc:
Everything goes into `./scripts`. We could at least use some sub-directories splitting things between:

* tmuxinator
* tests
* other stuff

PR #2952, opened 2023-08-14
organize scripts folder

justinmoon:
Closes https://github.com/fedimint/fedimint/issues/2134


scripts/final-checks.sh
justinmoon, line 1: I removed this file, doesn't seem like anyone uses it

scripts/gw-reload.sh
justinmoon, line 1: I removed this file, don't think anyone uses it now that the cln extension is much simpler #fightingentropy

dpc (dismissed):
I'm sure it will break something, but there's no other way really. :D

dpc:
Probably missed adjusting flake source filters @justinmoon

scripts/gw-reload.sh
m1sterc001guy: Now that the extension reliably dies with CLN, its easier to just kill/restart CLN itself.

justinmoon:
@dpc passes ci now

elsirion (approved):

PR #2956, opened 2023-08-14
feat: Lightning Gateway registers configurable number of route hints

m1sterc001guy:
Same PR as https://github.com/fedimint/fedimint/pull/2647 except this makes the number of route hints configurable from a `gatewayd` CLI parameter and it adds a test that verifies the correct number of route hints are returned.

Fixes: https://github.com/fedimint/fedimint/issues/2634


elsirion:
Test for this feature is failing.

m1sterc001guy:
> Test for this feature is failing.

I believe there is a race when setting up the channels and registering the gateway. Need to refactor a bit

m1sterc001guy:
Taking me a bit longer to debug than I would have liked. I think there is currently a bug or race condition when getting CLN's route hints. Sometimes it is returning 1 route hint when it should be returning 2.

m1sterc001guy:
LDK Node was causing the test I was writing to be pretty flaky. The channel would sometimes close between CLN and LDK Node which caused the wrong number of route hints to be returned. 

Given that this PR is not dependent on LDK node, I just took that part out for now and will return to the LDK Node issues later. The test should be reliable now.

gateway/ln-gateway/src/lib.rs
```diff
@@ -496,6 +509,7 @@ impl Gateway {
                         continue;
                     }
                 };
+
             if !route_hints.is_empty() || num_retries == ROUTE_HINT_RETRIES {
```
elsirion, line 532: Can we either:

1. Add a condition for `num_route_hints == 0` here to also exit in that case because then we expect the `route_hints` to be empty?
2. Add an early return path before the loop if `num_route_hints == 0` because we don't even need to fetch the routes in that case (although that might mask RPC connectivity problems and we should test that separately).

gateway/ln-gateway/src/lnd.rs
```diff
@@ -309,10 +312,15 @@ impl ILnRpcClient for GatewayLndClient {
             .map_err(|status| LightningRpcError::FailedToGetRouteHints {
                 failure_reason: format!("Failed to list channels {status:?}"),
             })?
-            .into_inner();
+            .into_inner()
+            .channels;
+
+        // Take the channels with the largest incoming capacity
+        channels.sort_by(|a, b| b.remote_balance.cmp(&a.remote_balance));
+        channels.truncate(num_route_hints);
```
elsirion, line 320: Some redundancy in logic here, but probably less complexity than communicating channel balances to `gatewayd`.

elsirion (commented):
Looks good, needs rebase.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: This should be handled at line 510:

```
if num_route_hints == 0 {
            return Ok(vec![]);
        }
```

elsirion: Sorry, I'm an idiot and didn't realize I was in per-commit view (and just looked at the first one apparently).

gateway/ln-gateway/src/ng/tests.rs
```diff
@@ -583,3 +585,94 @@ async fn test_gateway_cannot_pay_expired_invoice() -> anyhow::Result<()> {
...
+                    );
+                    let route_hint_pub_key = route_hint.0.get(0).unwrap().src_node_id;
+                    assert_eq!(
+                        route_hint_pub_key, public_key,
+                        "Public key of route hint hop did not match expected public key"
+                    );
+                }
+                _ => {
+                    // If there's more than one route hint, we're expecting the invoice to contain
+                    // `num_route_hints`. Each should have 2 hops.
```
elsirion, line 655: We might want to change that to also including the one-hop route hint even when including the 2-hop ones. But that can be a separate PR.

elsirion (approved):

elsirion (approved):

PR #2959, opened 2023-08-15
fix: avoid rebuilds

dpc:
From the code itself:

```sh
# 'cargo test' does not have a possibility of building whole workspace (to avoid any rebuilds), yet
# running just a subset of tests from a given package. To overcome it we parse the output of this
# command and run test binaries directly. Not elegant, but works.```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,7 +3,7 @@ rustflags = ["--cfg", "tokio_unstable"]
 
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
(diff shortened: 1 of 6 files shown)


dpc:
There's just no good way to run a subset of tests from a package, without causing rustc to rebuild the project due to selecting a subset of workspace. :/

dpc:
TODO: Use `$(cargo test --no-run  --workspace --all-targets 2>&1 | sed -n 's/.*(\([^)]*\)).*/\1/p' | grep ln_server)` to pick the test binaries package by package.

elsirion (dismissed):

elsirion (approved):

PR #2962, opened 2023-08-16
Make client subscription streams static

### LINKED ISSUE #2928: Update streams returned by client should be `'static`
Operation update streams and the balance stream have the lifetime of the client, which makes them hard to work with in an async environment.

elsirion:
If update streams have non-static lifetimes they cannot be moved between tasks/captured easily, making them less useful, in particular they cannot be given to front end code that typically doesn't have a lifetime guaranteed to be shorter than that of the client.

To achieve static lifetimes I had to clone the `Client(Arc<ClientInner>)` and move the reference into the stream. This means as long as any of these streams is still around the client will not terminate. Is this a problem @justinmoon? It does mean that, if we want to drop the client and its database (e.g. when wanting to leave and re-join a federation, all UI elements/other objects/tasks referencing these streams need to be dropped.

Fixes #2928

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -756,17 +756,20 @@ impl Client {
 
     /// Returns a stream that yields the current client balance every time it
     /// changes.
-    pub async fn subscribe_balance_changes(&self) -> BoxStream<'_, Amount> {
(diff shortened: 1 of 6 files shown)


douglaz (approved):
LGTM

PR #2968, opened 2023-08-16
fix: `just clippy` + unify some `justfile` commands

dpc:


douglaz (approved):
I was going to create a similar PR

PR #2969, opened 2023-08-16
cleanup: remove outdated test

douglaz:
On the new client there is a explicit state for unconfirmed transactions and this is already being tested: https://github.com/fedimint/fedimint/blob/3ed7fe0f4504617d250136e4150900df0a799ff1/modules/fedimint-wallet-tests/tests/tests.rs#L49

So we don't need this test anymore


dpc (approved):

elsirion (commented):
This test isn't doing the same. The test you refer to tests that the client will not prematurely submit the deposit claim transaction to the federation. We still need to test that the federation would not accept such a transaction.

PR #2971, opened 2023-08-17
fix: random ln gateway failures on start

dpc:
I think I've seen one cli_test_lightning_reconnect failing, locally due to lnd taking its sweet time after restart, while gateway giving up on getting the hint.

The other time, I've seen it failing on parsing some weird (temporary) response.

BTW. We should rewrite this part so gateway continuously checks for hints or something. There should be no such thing as panicking for external reasons.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -80,7 +80,7 @@ const INITIAL_SCID: u64 = 1;
 /// How long a gateway announcement stays valid
 pub const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
 
-const ROUTE_HINT_RETRIES: usize = 10;
+const ROUTE_HINT_RETRIES: usize = 30;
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -464,35 +464,51 @@ impl Gateway {
...
-
+        for num_retries in 0.. {
+            let (route_hints, node_pub_key, alias) =
+                match Self::fetch_lightning_route_info_try(lnrpc.as_ref()).await {
+                    Ok(res) => res,
+                    Err(e) => {
+                        if num_retries == ROUTE_HINT_RETRIES {
+                            return Err(e);
+                        }
+                        warn!("Could not fetch route hints: {e}");
```
elsirion, line 494: nit: I would have probably included the fact that it will be retried in `ROUTE_HINT_RETRY_SLEEP`

elsirion (approved):

PR #2976, opened 2023-08-17
fix(wallet-server): inconsistent DB state on startup

### LINKED ISSUE #2974: Some race condition in peg-in tests
```
 fedimint-test-all> 00:04:43 ---- on_chain_peg_in_and_peg_out_happy_case stdout ----
fedimint-test-all> 00:04:43 thread 'on_chain_peg_in_and_peg_out_happy_case' panicked at 'assertion failed: `(left == right)`
fedimint-test-all> 00:04:43   left: `Failed("Failed to claim: SubmitRejected(\"Federation rpc error {general => Received errors from 1 peers), 1 => Rpc error: ErrorObject { code: ServerError(400), message: \\\"Unknown block hash in peg-in proof: 38ce3ae24e621977181279f2913675693962340450deb35a5d0835ecb675bab7\\\", data: None }), }\")")`,
```

https://github.com/fedimint/fedimint/actions/runs/5887796866/job/15967798258?pr=2951

elsirion:
Fixes #2974.

Previous to 5f5e9de8106de060c4f9ac7d5fbad6c2878ae0cd the wallet server would not actually sync block hashes after the first block height consensus was reached but the consensus block height was still updated, making the database inconsistent since blocks that were below the consensus height would not be recognized.

This lead to the client rightfully assuming that it could claim the deposit but the server, not knowing about the block in which the transaction was mined yet, rejecting the claim transaction.

To reproduce you can run `on_chain_peg_in_and_peg_out_happy_case` in a single-threaded runtime (just change `#[tokio::test(flavor = "multi_thread")]` to `#[tokio::test]`), this is also the reason to include 303c7e3bd29c18a3654de48e420527086d59870f in this PR.


elsirion:
I also cherry-picked #2987 so we don't get problems in the merge queue.

justinmoon (approved):

PR #2978, opened 2023-08-17
fix: more retries in devimint's connect-fed 

dpc:
I've seen it failing, and I think it just happened to get a bit slow.


douglaz (approved):
LGTM

PR #2981, opened 2023-08-17
Derivable secret and query strategy docs

Kodylow:
Added notes from deep dive today on derivable secrets and query strategies to rust documentation.

### DIFF
--- a/crypto/derive-secret/src/lib.rs
+++ b/crypto/derive-secret/src/lib.rs
@@ -97,8 +97,8 @@ impl DerivableSecret {
         }
     }
 
-    // secp256k1 keys are used for bitcoin deposit addresses, redeem keys and
-    // contract keys for lightning.
+    /// secp256k1 keys are used for bitcoin deposit addresses, redeem keys and
+    /// contract keys for lightning.
     pub fn to_secp_key<C: Signing>(self, ctx: &Secp256k1<C>) -> KeyPair {
         for key_try in 0u64.. {
             let secret = self
@@ -114,8 +114,8 @@ impl DerivableSecret {
         unreachable!("If key generation fails this often something else has to be wrong.")
     }
 
-    // bls12-381 keys are used for the guardians' threshold signature scheme,
-    // and most importantly for its use for the blinding keys for e-cash notes.
+    /// bls12-381 keys are used for the guardians' threshold signature scheme,
+    /// and most importantly for its use for the blinding keys for e-cash notes.
     pub fn to_bls12_381_key(&self) -> Scalar {
         Scalar::from_bytes_wide(&self.kdf.derive(&tagged_derive(BLS12_381_TAG, ChildId(0))))
     }
--- a/crypto/derive-secret/src/lib.rs
+++ b/crypto/derive-secret/src/lib.rs
@@ -19,7 +19,51 @@ const RAW_BYTES: &[u8; 8] = b"rawbytes";
 #[derive(Debug, Copy, Clone, Encodable, Decodable)]
 pub struct ChildId(pub u64);
 
-/// Secret key that allows deriving child secret keys
+/// `DerivableSecret` represents a secret key that can be used to derive child
+/// secret keys. Fedimint on init generates a root key, and then derives child
+/// keys from it for each of the modules. Each module's child key can then be
+/// used to derive child keys for its specific use cases. This derivable secret
+/// pattern is especially useful for creating backups for both client and server
+/// side operations.
+///
+/// The `DerivableSecret` struct in this implementation is only used for
+/// deriving secret keys, not public keys. This allows us to support multiple
+/// crypto schemes for the different cryptographic operations used across the
+/// different modules:
+/// - secp256k1 for bitcoin deposit addresses, redeem keys and contract keys for
+///   lightning,
+/// - bls12-381 for the guardians' threshold signature scheme,
+/// - chacha20-poly1305 for symmetric encryption used for backups.
+///
+/// The `DerivableSecret` struct contains two fields:
+/// - `level`: This represents the derivation level of the secret key and acts
+///   as an index. The root key starts at level 0, and each derived child key
+///   increments this level.
+/// - `kdf`: This is an instance of the HKDF (Hash-based Key Derivation
+///   Function) with SHA-512 as the underlying hash function. It is used to
+///   derive child keys.
+///
+/// The `salt` used in the `new_root` function is just random data that is used
+/// as an additional input to the HKDF.
+///
+/// # Example
+/// In the context of e-cash, the `DerivableSecret` can be used to derive
+/// blinding and spend keys for e-cash notes. Here's an example from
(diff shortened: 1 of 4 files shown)


crypto/derive-secret/src/lib.rs
```diff
@@ -19,7 +19,51 @@ const RAW_BYTES: &[u8; 8] = b"rawbytes";
...
+/// `DerivableSecret` represents a secret key that can be used to derive child
+/// secret keys. Fedimint on init generates a root key, and then derives child
+/// keys from it for each of the modules. Each module's child key can then be
+/// used to derive child keys for its specific use cases. This derivable secret
+/// pattern is especially useful for creating backups for both client and server
+/// side operations.
+///
+/// The `DerivableSecret` struct in this implementation is only used for
+/// deriving secret keys, not public keys. This allows us to support multiple
+/// crypto schemes for the different cryptographic operations used across the
```
elsirion, line 31: I'd probably add "in contrast to BIP32", otherwise it might be hard to understand because you can do `xPriv -- derive --> xPriv -- convert --> Public` and just not `xPriv -- convert --> xPub -- derive --> xPub/Public`.

crypto/derive-secret/src/lib.rs
```diff
@@ -19,7 +19,51 @@ const RAW_BYTES: &[u8; 8] = b"rawbytes";
...
+/// The `DerivableSecret` struct contains two fields:
+/// - `level`: This represents the derivation level of the secret key and acts
+///   as an index. The root key starts at level 0, and each derived child key
+///   increments this level.
+/// - `kdf`: This is an instance of the HKDF (Hash-based Key Derivation
+///   Function) with SHA-512 as the underlying hash function. It is used to
+///   derive child keys.
+///
+/// The `salt` used in the `new_root` function is just random data that is used
+/// as an additional input to the HKDF.
```
elsirion, line 47: I'd move that to the function itself which doesn't have docs rn

crypto/derive-secret/src/lib.rs
```diff
@@ -19,7 +19,51 @@ const RAW_BYTES: &[u8; 8] = b"rawbytes";
...
+///
+/// # Example
+/// In the context of e-cash, the `DerivableSecret` can be used to derive
+/// blinding and spend keys for e-cash notes. Here's an example from
+/// modules/fedimint-mint-client/src/output.rs
```
elsirion, line 52: Can we make this an actual link? Or is it already (I doubt it)?

crypto/derive-secret/src/lib.rs
```diff
@@ -19,7 +19,51 @@ const RAW_BYTES: &[u8; 8] = b"rawbytes";
...
+/// let blinding_key = BlindingKey(secret.child_key(BLINDING_KEY_CHILD_ID).to_bls12_381_key());
+/// let blinded_nonce = blind_message(nonce.to_message(), blinding_key);
+///
+/// let cr = NoteIssuanceRequest {
+///     spend_key,
+///     blinding_key,
+/// };
+///
+/// (cr, BlindNonce(blinded_nonce))
+/// ```
```
elsirion, line 66: This doctest fails because it misses some setup code that generates `secret` and imports a bunch of structs. [If the setup becomes too big you can hide parts of it by prefixing with `#`.](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example)

crypto/derive-secret/src/lib.rs
```diff
@@ -19,7 +19,51 @@ const RAW_BYTES: &[u8; 8] = b"rawbytes";
 #[derive(Debug, Copy, Clone, Encodable, Decodable)]
 pub struct ChildId(pub u64);
 
-/// Secret key that allows deriving child secret keys
+/// `DerivableSecret` represents a secret key that can be used to derive child
```
elsirion, line 22: Some of this might also be good module-level documentation? At least the module-level docs should link here imo since it's the main struct.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -472,6 +472,40 @@ impl ClientModuleGen for MintClientGen {
...
+/// only the owner of the e-cash note can spend it.
+///
+/// # E-Cash Note Creation
+///
+/// When creating an e-cash note, the `MintClientModule` first derives the
+/// blinding and spend keys from the `DerivableSecret`. It then creates a
+/// `NoteIssuanceRequest` containing the blinded spend key and sends it to the
+/// mint server. The mint server signs the blinded spend key and returns it to
+/// the client. The client can then unblind the signed spend key to obtain the
+/// e-cash note, which can be spent using the spend key.
```
elsirion, line 508: I'd either move this to the field `secret` itself or maybe to the function [`new_note_secret_static`](https://github.com/fedimint/fedimint/blob/3b13fc285e863d5777a3a85482bfdebc50fb0d9e/modules/fedimint-mint-client/src/lib.rs#L1060). Or it might be best to split out the relevant parts to each place/cross-reference.

fedimint-core/src/query.rs
```diff
@@ -13,6 +13,17 @@ use crate::module::{
...
+/// we only trust that a threshold of t/n guardians are honest. Where f is the
+/// number of malicious guardians, we require 3f + 1 total nodes n and t = n-f
+/// for the honest threshold. This means that we need to query multiple
+/// guardians for the same information, and then combine the results. The
+/// QueryStrategy generally retries until we have enough responses to meet the
+/// threshold, and handles when a guardian doesn't respond. There are nuanced
+/// cases where instead of arriving at a single t/n result the client will want
+/// a union e.g. guardians 1 and 2 respond (A,B) and guardian 3 responds (C) so
+/// return (A,B,C). This is handled by the specfic QueryStrategy
+/// implementations.
```
elsirion, line 26: When mentioning specific strategies there should be a link to them. I think the whole "trusting only a threshold" problem could be described more succinctly and the concrete strategies could go into more detail.

modules/fedimint-ln-common/src/api.rs
```diff
@@ -59,6 +59,9 @@ where
         .await
     }
 
+    /// There is no consensus within Fedimint on the gateways, each guardian
+    /// might be aware of different ones, so we just return the union of all
+    /// responses and allow client selection.
```
elsirion, line 64: Could be more succinct, only describing what it does, i.e. "returns union of gateways registered with guardian servers". Why that's the case is less relevant for the function's documentation (that's the caller's problem) and should thus only be mentioned further down if at all.

crypto/derive-secret/src/lib.rs
justinmoon: "Fedimint on init generates a root key, and then derives child keys from it for each of the modules. " is a very ambiguous sentence. Is this client or server? In this context I think we just care about what it does and not where / when other crates use it.

dpc:
Finished #3257

PR #2982, opened 2023-08-17
chore: increase peg-in timeout in test

dpc:
A shot in a dark until someone debugs it better.


dpc:
Pfff. I guess it's not it.

elsirion:
See #2976.

dpc:
@elsirion Is there a reason to not keep the timeout longer, BTW? Just ... in case?

PR #2991, opened 2023-08-18
chore: minor tweaks after debugging

dpc:


devimint/src/util.rs
```diff
@@ -251,30 +252,23 @@ pub async fn poll_max_retries<Fut>(name: &str, retries: usize, f: impl Fn() -> F
...
-                    debug!("polling {name} failed with: {other:?}, will retry... ({i}/{retries})");
-                    task::sleep(Duration::from_secs(1)).await;
-                }
+            Err(e) => {
+                bail!("{name} failed after {retries} retries with: {e:?}");
             }
         }
     }
+
+    unreachable!();
```
elsirion, line 271: I'm pretty sure I saw that change already, did you split another PR I had already reviewed?

elsirion (approved):

devimint/src/util.rs
dpc: I did similar change twice in two places, and also I think this commit was somewhere else already.

modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -27,7 +27,7 @@ fn bsats(satoshi: u64) -> bitcoin::Amount {
 
 const PEG_IN_AMOUNT_SATS: u64 = 5000;
 const PEG_OUT_AMOUNT_SATS: u64 = 1000;
-const PEG_IN_TIMEOUT: Duration = Duration::from_secs(30);
+const PEG_IN_TIMEOUT: Duration = Duration::from_secs(60);
```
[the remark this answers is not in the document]
dpc: Nope. :D . But unless we ever expect to wait for all this time anywhere, I guess it doesn't hurt to have these be a bit longer, in case CI machine is extra slow.

PR #2992, opened 2023-08-18
fix: compress debug symbols directive not working

dpc:
before:

```
-rwxr-xr-x 2 dpc dpc 46M Aug 18 08:38 ./target/release/fedimintd
```

after:

```
-rwxr-xr-x 2 dpc dpc 44M Aug 18 08:46 ./target/release/fedimintd
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,7 +3,7 @@ rustflags = ["--cfg", "tokio_unstable"]
 
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
-rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "link-arg=-Wl,--compress-debug-sections=zlib", "--cfg", "tokio_unstable"]
+rustflags = ["-C", "link-arg=-Wl,--compress-debug-sections=zlib,-fuse-ld=lld", "--cfg", "tokio_unstable"]
 
 # mold is currently broken on MacOS
 # [target.aarch64-apple-darwin]


elsirion (approved):

elsirion:
Calling strip on it brings it down to 33M btw, not sure what the drawbacks of that would be.

dpc:
> Calling strip on it brings it down to 33M btw, not sure what the drawbacks of that would be.

No line:column info on panics and errors, etc. , IIRC.

.cargo/config.toml
```diff
@@ -3,7 +3,7 @@ rustflags = ["--cfg", "tokio_unstable"]
 
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
-rustflags = ["-C", "link-arg=-fuse-ld=lld", "-C", "link-arg=-Wl,--compress-debug-sections=zlib", "--cfg", "tokio_unstable"]
+rustflags = ["-C", "link-arg=-Wl,--compress-debug-sections=zlib,-fuse-ld=lld", "--cfg", "tokio_unstable"]
```
bradleystachurski, line 6: After rebasing, I'm getting the following error when building

```
clang-11: error: linker command failed with exit code 137 (use -v to see invocation)
```

Reverting this change allows the build to succeed. I've created an [issue](https://github.com/fedimint/fedimint/issues/3000) (3000, woo!) so we can find a solution that compresses and works with the linker.

elsirion: Might that be a Linux vs. Mac problem again? What's your platform?

PR #2994, opened 2023-08-18
chore: make it easier to identify test's tmp dir in log dump

dpc:
image


elsirion (approved):
That's a creative solution!

PR #2996, opened 2023-08-18
fix(devimint): always await after SIGKILL

douglaz:
(This is related to https://github.com/fedimint/fedimint/pull/2853 and https://github.com/fedimint/fedimint/pull/2873)

Today we rely on drop to sigkill the processes and we suppose they are killed almost immediately. But this isn't always the case, it's very possible for devimint to exit and then we start a new one and the new daemons will fail because the older ones are not quite dead and using the same listen address.

(if this happen inside the context of tests, the tests may keep retrying for many minutes before some timeout or retry limit happens)

So now devimint will always await for the processes to exit. To accomplish this we can't rely on drop anymore and must call explicitly a kill method.

Also note that because we don't want to rely on drop, we can't use something like weak refs on `ProcessManager`. We need absolute control of the termination of the subprocesses. So either they are terminated explicitly with `terminate()` or they are killed before devimint exits with `kill_all()`

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -1140,7 +1140,8 @@ async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
     Ok((process_mgr, task_group))
 }
 
-async fn cleanup_on_exit<T>(
+async fn wait_to_exit<T>(
+    process_mgr: ProcessManager,
     main_process: impl futures::Future<Output = Result<T>>,
     task_group: TaskGroup,
 ) -> anyhow::Result<()> {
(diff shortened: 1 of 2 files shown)


dpc:
>  To accomplish this we can't rely on drop anymore and must call explicitly a kill method.

Can't we do `block_in_place` + `block_on`?

douglaz:
> > To accomplish this we can't rely on drop anymore and must call explicitly a kill method.
> 
> Can't we do `block_in_place` + `block_on`?

I don't like the idea very much, but it should work and I see the Drop on Client is already doing something very similar to that, so let's try it.

douglaz:
Closed in favor of https://github.com/fedimint/fedimint/pull/2998

PR #3005, opened 2023-08-21
fix: make wallet tests wait consensus

douglaz:
Also don't return zero when there is no block count consensus.

Build on https://github.com/fedimint/fedimint/pull/2976 and implements the fix for https://github.com/fedimint/fedimint/issues/2974 in a way that fixes the tests but don't introduce regressions on initial blockchain scanning.

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -180,6 +180,8 @@ impl Federation {
 
     pub async fn pegin(&self, amount: u64) -> Result<()> {
         info!(amount, "Peg-in");
+        // Some blocks must always be mined first before pegin
+        self.await_minimum_blocks().await?;
         let deposit = cmd!(self, "deposit-address").out_json().await?;
         let deposit_address = deposit["address"].as_str().unwrap();
         let deposit_operation_id = deposit["operation_id"].as_str().unwrap();
@@ -197,6 +199,8 @@ impl Federation {
 
     pub async fn pegin_gateway(&self, amount: u64, gw: &Gatewayd) -> Result<()> {
         info!(amount, "Pegging-in gateway funds");
+        // Some blocks must always be mined first before pegin
+        self.await_minimum_blocks().await?;
         let fed_id = self.federation_id().await;
         let pegin_addr = cmd!(gw, "address", "--federation-id={fed_id}")
             .out_json()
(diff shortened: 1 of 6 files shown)


douglaz:
Not sure why this broke random tests :confused:

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -574,7 +577,8 @@ impl ServerModule for Wallet {
             api_endpoint! {
                 "block_count",
                 async |module: &Wallet, context, _params: ()| -> u32 {
-                    Ok(module.consensus_block_count(&mut context.dbtx()).await)
+                    // TODO: should we change this to Option?
+                    Ok(module.consensus_block_count(&mut context.dbtx()).await.unwrap_or_default())
```
elsirion, line 581: It would be the cleanest way to do it, but it shouldn't lead to any problems to leave it as-is imo. The client generally waits for non-zero consensus block height due to the finality delay.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -370,11 +370,14 @@ impl ServerModule for Wallet {
                 let new_consensus_block_count = self.consensus_block_count(dbtx).await;
 
                 // only sync from the first non-default consensus block count
```
elsirion, line 372: "non-default" isn't the right term anymore in this comment imo

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -370,11 +370,14 @@ impl ServerModule for Wallet {
...
+                if old_consensus_block_count.is_some()
+                    && new_consensus_block_count.is_some()
+                    && new_consensus_block_count.unwrap() > old_consensus_block_count.unwrap()
+                {
                     self.sync_up_to_consensus_height(
                         dbtx,
-                        old_consensus_block_count,
-                        new_consensus_block_count,
+                        old_consensus_block_count.unwrap(),
+                        new_consensus_block_count.unwrap(),
```
elsirion, line 380: I'm not a big fan of the unwrapping here. A nested if-let+if or a match statement would be nicer:

```rust
match (old_consensus_block_count, new_consensus_block_count) {
    (Some(old), Some(new)) if new > old => {
         self.sync_up_to_consensus_height(dbtx, old, new).await;
    }
    _ => {}
}
```

douglaz:
Looks good enough

modules/fedimint-wallet-server/src/lib.rs
dpc: Ascending order better, IMO. `old < new` 


```rust
    if let (Some(old), Some(new))  = (old_consensus_block_count, new_consensus_block_count){
        if old < new {
            println!("yes");
        }
    }
```

Unfortunately `&& old < new` is not yet stable.

douglaz: Let's agree to disagree :smile:

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -369,14 +381,21 @@ impl ServerModule for Wallet {
...
+                    ?new_consensus_block_count,
+                    "Received block count vote"
+                );
+
+                // only sync when we have a consensus block count
+                match (old_consensus_block_count, new_consensus_block_count) {
+                    (Some(old), Some(new)) if new > old => {
+                        self.sync_up_to_consensus_height(dbtx, old, new).await;
+                    }
+                    _ => {}
```
elsirion, line 398: We could assert a few things that should never happen (like going backwards) here. But I'll discuss that with @joschisan. Out of scope for this fix.

devimint/src/federation.rs
```diff
@@ -281,8 +296,9 @@ impl Federation {
     }
 
     pub async fn generate_epochs(&self, epochs: usize) -> Result<()> {
+        let finality_delay = self.get_finality_delay().await?;
         for _ in 0..epochs {
-            self.bitcoind.mine_blocks(10).await?;
+            self.bitcoind.mine_blocks(finality_delay.into()).await?;
```
elsirion, line 301: Mining finality delay many block isn't necessary each time to generate an epoch, I guess it helps with the first time that wouldn't trigger if height-delay<0.

elsirion (approved):

PR #3009, opened 2023-08-22
feat(client): expose secret derived from seed for external use

### LINKED ISSUE #2979: Client should not expose the root secret
Instead we should have a derivation path not conflicting with any modules etc. whose corresponding secret can be used for client-external features.

elsirion:
Fixes #2979


justinmoon (approved):
Looks good to me

PR #3014, opened 2023-08-22
chore: clean up naming of tests a bit

dpc:
* Get rid of that "cli" everywhere. It means nothing(?).
* Singular is better than plural.


justinmoon:
These tests were originally called "cli tests" because they were written in bash and exercised `fedimint-cli`. Agree it's good to get rid of.

elsirion:
The `wasm-tests.sh` -> `wasm-test.sh` broke CI.

elsirion (approved):

Issue #3017, opened 2023-08-22
`just mprocs` errors on start


dpc:
```
2023-08-22T22:47:49.145684Z  WARN consensus: ERROR FederationError { general: Some(Received errors from 2 peers

Stack backtrace:
   0: fedimint_core::query::ErrorStrategy::process
             at ./fedimint-core/src/query.rs:106:31
   1: <fedimint_core::query::ThresholdConsensus<R> as fedimint_core::query::QueryStrategy<R>>::process
             at ./fedimint-core/src/query.rs:169:27
   2: fedimint_core::api::FederationApiExt::request_with_strategy::{{closure}}
             at ./fedimint-core/src/api.rs:236:41
   3: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/future/future.rs:125:9
   4: fedimint_core::api::FederationApiExt::request_current_consensus::{{closure}}
             at ./fedimint-core/src/api.rs:300:10
   5: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/future/future.rs:125:9
   6: <T as fedimint_core::api::GlobalFederationApi>::consensus_config_hash::{{closure}}
             at ./fedimint-core/src/api.rs:560:14
   7: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/future/future.rs:125:9
   8: fedimint_server::consensus::server::ConsensusServer::run_consensus::{{closure}}
             at ./fedimint-server/src/consensus/server.rs:263:52
   9: fedimint_server::FedimintServer::run::{{closure}}
             at ./fedimint-server/src/lib.rs:100:56
  10: <futures_util::future::maybe_done::MaybeDone<Fut> as core::future::future::Future>::poll
             at /home/dpc/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.28/src/future/maybe_done.rs:95:38
  11: fedimintd::fedimintd::run::{{closure}}::{{closure}}
             at /home/dpc/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.28/src/async_await/join_mod.rs:95:13
  12: <futures_util::future::poll_fn::PollFn

...

  56: main
[...]

PR #3018, opened 2023-08-22
chore: improve output on start of `just mprocs`

dpc:
Re #3017

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -12,6 +12,7 @@ use fedimint_core::encoding::DecodeError;
 use fedimint_core::epoch::{
     ConsensusItem, EpochOutcome, EpochVerifyError, SerdeConsensusItem, SignedEpochOutcome,
 };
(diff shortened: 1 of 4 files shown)


fedimint-server/src/consensus/server.rs
```diff
@@ -264,7 +265,7 @@ impl ConsensusServer {
                 Ok(consensus_hash) if consensus_hash == our_hash => break,
                 Ok(_) => bail!("Our consensus config doesn't match peers!"),
                 Err(e) => {
-                    warn!(target: LOG_CONSENSUS, "ERROR {:?}", e)
+                    warn!(target: LOG_CONSENSUS, "Could not check consensus config hash: {}", OptStracktrace(e))
```
dpc, line 268: @douglaz You showed me the `{:?}` printing stacktrace, but it seems to me now that oftentimes this creates tons of output for errors that are expected, so I created this `OptStacktrace` wrapper to have ability to get the stacktrace if really needed without being too noisy. We'll have to find some balance, case by case, I guess.

elsirion (approved):

PR #3019, opened 2023-08-23
feat(shell): wait for background processes to exit

douglaz:
It's inconvenient when a script exits but still leaves stuff running (even if these processes have just received a signal and will hopefully die soon). This can cause failures if some other script starts right away and uses the same resources.


douglaz:
I have a better idea

PR #3020, opened 2023-08-23
chore(ci): increase linux build timeout

elsirion:
We are [seeing a lot of tests time out](https://github.com/fedimint/fedimint/actions/runs/5953944182/job/16149367352?pr=3008) while fetching from our nix caches, there might be temporary network congestion, but that shouldn't make our tests fail.


dpc (approved):

elsirion:
I doubt it will help after chatting with @dpc about it :(

elsirion:
Maybe I was wrong and whatever was stuck unstuck itself after an ungodly amount of time?!

PR #3022, opened 2023-08-23
fix(devimint): proper shutdown lightningd

douglaz:
The cln extension may continue running for a while if we don't shutdown it explicitly. This may cause errors on the next execution


m1sterc001guy:
I thought killing CLN would cause the extension to stop too? We had issues with this in the past but I thought we had fixed those problems

douglaz:
> I thought killing CLN would cause the extension to stop too? We had issues with this in the past but I thought we had fixed those problems

It doesn't seem to receive a kill at the same time. It keeps running for many seconds then it seems to realize lightningd is dead then it exits.

douglaz:
A command to test is inside mprocs:
```bash
killall lightningd;for i in {0..1000}; do echo $i;killall -0 gateway-cln-extension;sleep 0.1;done
```

On my machine it takes >12s for gateway-cln-extension to disappear

You can also see that if you send a SIGKILL the time is the same:

```bash
killall -9 lightningd;for i in {0..1000}; do echo $i;killall -0 gateway-cln-extension;sleep 0.1;done
```
So lightningd doesn't try to kill the extension even on SIGTERM.

dpc (approved):

douglaz:
Drop should be in a inner object protected by `Arc`, otherwise we may keep try to terminate when a cloned object goes out of context.

elsirion (approved):

PR #3024, opened 2023-08-24
Prefix federation ID to e-cash notes when spending/reissuing/validating

### LINKED ISSUE #2783: Prefix serialized e-cash with federation ID
When serializing e-cash we currently rely on the user to remember to which federation it belongs. To be able to give better error messages we should include the federation id so that incompatible federations can be detected. Otherwise the federation would just complain that the signatures are invalid.

shaurya947:
We create a new struct called `OOBNotes` that encapsulates the federation ID + notes. Modules aren't aware of the federation ID so we also have to inject that.

Should close #2783

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -14,14 +14,12 @@ use fedimint_client::Client;
 use fedimint_core::config::{ClientConfig, FederationId};
 use fedimint_core::core::{ModuleInstanceId, ModuleKind};
 use fedimint_core::time::now;
-use fedimint_core::{Amount, ParseAmountError, TieredMulti, TieredSummary};
+use fedimint_core::{Amount, ParseAmountError, TieredSummary};
 use fedimint_ln_client::contracts::ContractId;
 use fedimint_ln_client::{
     InternalPayState, LightningClientExt, LnPayState, LnReceiveState, PayType,
 };
-use fedimint_mint_client::{
-    parse_ecash, serialize_ecash, MintClientExt, MintClientModule, SpendableNote,
-};
+use fedimint_mint_client::{MintClientExt, MintClientModule, OOBNotes};
 use fedimint_wallet_client::{WalletClientExt, WalletClientModule, WithdrawState};
 use futures::StreamExt;
 use serde::{Deserialize, Serialize};
@@ -55,10 +53,7 @@ pub enum ClientCmd {
     /// Display wallet info (holdings, tiers)
     Info,
     /// Reissue notes received from a third party to avoid double spends
-    Reissue {
-        #[clap(value_parser = parse_ecash)]
-        notes: TieredMulti<SpendableNote>,
-    },
+    Reissue { oob_notes: OOBNotes },
     /// Prepare notes to send to a third party as a payment
     Spend {
         #[clap(value_parser = parse_fedimint_amount)]
@@ -66,10 +61,7 @@ pub enum ClientCmd {
     },
     /// Verifies the signatures of e-cash notes, but *not* if they have been
     /// spent already
-    Validate {
-        #[clap(value_parser = parse_ecash)]
-        notes: TieredMulti<SpendableNote>,
-    },
+    Validate { oob_notes: OOBNotes },
     /// Create a lightning invoice to receive payment via gateway
     LnInvoice {
         #[clap(long, value_parser = parse_fedimint_amount)]
@@ -148,10 +140,10 @@ pub async fn handle_ng_command(
 ) -> anyhow::Result<serde_json::Value> {
     match command {
         ClientCmd::Info => get_note_summary(&client).await,
-        ClientCmd::Reissue { notes } => {
-            let amount = notes.total_amount();
+        ClientCmd::Reissue { oob_notes } => {
+            let amount = oob_notes.total_amount();
 
-            let operation_id = client.reissue_external_notes(notes, ()).await?;
+            let operation_id = client.reissue_external_notes(oob_notes, ()).await?;
             let mut updates = client
                 .subscribe_reissue_external_notes(operation_id)
                 .await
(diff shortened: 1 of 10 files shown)


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -74,14 +74,27 @@ const MINT_BACKUP_RESTORE_OPERATION_ID: OperationId = OperationId([0x01; 32]);
 
 pub const LOG_TARGET: &str = "client::module::mint";
 
+#[derive(Clone, Debug, Decodable, Encodable, Deserialize, Serialize)]
+pub struct OOBNotes {
+    federation_id: FederationId,
+    pub notes: TieredMulti<SpendableNote>,
+}
```
elsirion, line 81: Why is one `pub` but not the other?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -74,14 +74,27 @@ const MINT_BACKUP_RESTORE_OPERATION_ID: OperationId = OperationId([0x01; 32]);
 
 pub const LOG_TARGET: &str = "client::module::mint";
 
+#[derive(Clone, Debug, Decodable, Encodable, Deserialize, Serialize)]
```
elsirion, line 77: The `Serialize`/`Deserialize` impl does not what we want here, the struct should be encoded as one string if the encoding format is human readable. If you aren't familiar with custom implementations of `Serialize`/`Deserialize` you could just keep the `serde_ecash` module around for now.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1274,49 +1307,22 @@ pub struct SpendableNote {
...
+pub fn serialize_ecash(ecash: &OOBNotes) -> String {
     let mut bytes = Vec::new();
     Encodable::consensus_encode(ecash, &mut bytes).expect("encodes correctly");
     base64::encode(&bytes)
 }
 
 /// Decode a set of e-cash notes from a base64 string. See also
 /// [`serialize_ecash`].
-pub fn parse_ecash(s: &str) -> anyhow::Result<TieredMulti<SpendableNote>> {
+pub fn parse_ecash(s: &str) -> anyhow::Result<OOBNotes> {
```
elsirion, line 1318: These functions should be transformed into `FromStr` and `Display` implementations now that there is a local type for serializable e-cash.

fedimint-cli/src/client.rs
```diff
@@ -57,7 +57,7 @@ pub enum ClientCmd {
     /// Reissue notes received from a third party to avoid double spends
     Reissue {
         #[clap(value_parser = parse_ecash)]
```
elsirion, line 59: Once `OOBNotes` implements `FromStr` this line won't be necessary anymore.

elsirion (changes requested):

elsirion:
Also I forgot to mention: you can squash the first two and the last two commits. That way there is one changing the module API and one making use of it. Currently it's a bit too granular/the commits only make sense in each others context.

modules/fedimint-mint-client/src/lib.rs
shaurya947: I initially exposed `notes` for `totalAmount()` but then decided to make a pass-through function on `OOBNotes` itself. I suppose they could both be non-`pub` and there could be getters for both fields? Or they could both be made `pub`? Probably the latter is good enough?

modules/fedimint-mint-client/src/lib.rs
shaurya947: Ah I think I follow. We'd still like for the serialization of the notes (`OOBNotes`) to be one string containing a base-64 byte encoding of the struct. I will write a custom serializer.

dpc: There's a `serde_as_encodable_hex` for exactly this already.

modules/fedimint-mint-client/src/lib.rs
dpc: I'd make both `pub`. This struct is just a no-logic aggregator of two datums.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -437,6 +438,7 @@ impl ClientModuleGen for WalletClientGen {
 
     async fn init(
         &self,
+        _federation_id: FederationId,
```
dpc, line 441: Seems a bit one-off.

We keep tweaking the arguments to this function...

shaurya947:
I think I've addressed all the feedback in the latest push. `rust-tests` passes  but `load-test-tool-test` fails with this error:

image

To debug, I modified the below code to print the note:

image

I can see the notes being printed:

image

I tried the value in an online validator and indeed it's not padded (even though it's valid). I may be missing something small but how do I ensure that the encoding is padded as necessary?

shaurya947:
I wonder if I'm doing this wrong by using `serde_as_encodable_hex` because that's hexifying the bytes after encoding them, and that's not the kind of string input we expect for `OOBNotes`. Going to try with custom ser/de impl to verify.

shaurya947:
> I wonder if I'm doing this wrong by using `serde_as_encodable_hex` because that's hexifying the bytes after encoding them, and that's not the kind of string input we expect for `OOBNotes`. Going to try with custom ser/de impl to verify.

Ok, yes, that was it. I feel silly, hah! Should be all good now—I added a custom ser/de impl.

On second thought I suppose we _could_ use hex (and change the `FromStr` impl to account for it). Let me know if that's preferable and I can change it back.

elsirion:
Hex encoding would be significantly longer, that's why I chose base64 in the first place.

elsirion (approved):
Looks good, just needs a rebase :)

elsirion (approved):

dpc:
Could have made `serde_as_encodable_base64` macro. :D

shaurya947:
Thanks for the guidance folks, I enjoyed working on this enhancement :raised_hands:

Issue #3026, opened 2023-08-24
Re-enable flaky peg_outs_support_rbf test


justinmoon:
Disabled in https://github.com/fedimint/fedimint/pull/3023

PR #3027, opened 2023-08-24
fix(wallet-tests): rbf test isn't flaky anymore

douglaz:
Add tests back and now just accept that we can't control which transaction will be mined (the original or the RBF one)

Closes https://github.com/fedimint/fedimint/issues/3026

### DIFF
--- a/modules/fedimint-wallet-tests/tests/tests.rs
+++ b/modules/fedimint-wallet-tests/tests/tests.rs
@@ -161,12 +161,7 @@ async fn peg_out_fail_refund() -> anyhow::Result<()> {
     Ok(())
 }
 
-// FIXME: This test is flaky, it fails with:
-// left: `Amount { msats: 3270000 }`,
-// right: `Amount { msats: 2540000 }`',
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -221,10 +216,17 @@ async fn peg_outs_support_rbf() -> anyhow::Result<()> {
...
+    let current_balance = client.get_balance().await;
+    assert_eq!(balance_sub.ok().await?, current_balance);
+    // So we don't know which transaction will get mined first, it could be any one
+    // of the two, so we accept both
+    if current_balance != balance_after_rbf_peg_out
+        && current_balance != balance_after_normal_peg_out
+    {
+        bail!(
+            "Balance is {current_balance}, expected {balance_after_rbf_peg_out} or {balance_after_normal_peg_out}")
+    }
```
elsirion, line 230: I wonder if we could control which tx gets included … but this seems good enough.

elsirion (approved):

PR #3032, opened 2023-08-25
Client operation log ergonomic improvements

### LINKED ISSUE #3028: Make operation meta types more discoverable
I'd probably be good to make it a convention to have the meta type be a associated type of client extension traits.

### LINKED ISSUE #3029: Rename `operation_type` to `operation_module_kind` in `OperationLogEntry`
The only way the property is used right now is to store the module kind.

### LINKED ISSUE #3030: Add paid fee to lightning send operation meta struct

elsirion:
Fixes #3028
Fixes #3029 
Fixes #3030

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -6,7 +6,7 @@ use std::iter::once;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
-use anyhow::{bail, ensure, format_err};
+use anyhow::{bail, ensure, format_err, Context};
 use async_stream::stream;
 use bitcoin::{KeyPair, Network};
 use bitcoin_hashes::Hash;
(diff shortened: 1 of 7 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -274,10 +274,28 @@ impl LightningClientExt for Client {
...
+            LightningOutput::Contract(contract) => {
+                let fee_msat = contract
+                    .amount
+                    .msats
+                    .checked_sub(
+                        invoice
+                            .amount_milli_satoshis()
+                            .ok_or(anyhow::anyhow!("MissingInvoiceAmount"))?,
+                    )
+                    .expect("Contract amount should be greater or equal than invoice amount");
```
douglaz, line 289: panic in a client? :thinking:

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -274,10 +274,28 @@ impl LightningClientExt for Client {
...
+        // it/bounds for it
+        let fee = match &output.output {
+            LightningOutput::Contract(contract) => {
+                let fee_msat = contract
+                    .amount
+                    .msats
+                    .checked_sub(
+                        invoice
+                            .amount_milli_satoshis()
+                            .ok_or(anyhow::anyhow!("MissingInvoiceAmount"))?,
```
douglaz, line 287:
```suggestion
                            .context("MissingInvoiceAmount")?,
```

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Maybe better to just return an `anyhow::anyhow!` error and prevent the client from paying? The client probably should do some reasonable fee checking before attempting the payment

elsirion: This happening would indicate a bug in the client logic. Underfunding an outgoing contract doesn't make any sense, the gateway will just reject. So this really cannot happen unless we make a mistake (e.g. forgetting to change this should we ever implement negative fees).

modules/fedimint-ln-client/src/lib.rs
elsirion: TIL: one can use `context` on `Option`s :open_mouth:

dpc (approved):

PR #3033, opened 2023-08-25
chore: pin ldk-node to git hash

justinmoon:
The only thing pinning `ldk-node` to a specific version is our `Cargo.lock`. But if anyone uses `fedimint-testing` as a library, our `Cargo.lock` has no effect and they'll just pull `ldk-node` from master, which will be prone to breaking.


m1sterc001guy (approved):

elsirion (approved):
Good catch!

PR #3035, opened 2023-08-26
fix: allow LND to pay < 1 sat invoices

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2827

image

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -15,7 +15,8 @@ use fedimint_logging::LOG_DEVIMINT;
 use futures::executor::block_on;
 use tokio::fs;
 use tokio::sync::{MappedMutexGuard, Mutex, MutexGuard};
-use tonic_lnd::lnrpc::GetInfoRequest;
+use tonic_lnd::lnrpc::policy_update_request::Scope;
+use tonic_lnd::lnrpc::{ChanInfoRequest, GetInfoRequest, ListChannelsRequest, PolicyUpdateRequest};
 use tonic_lnd::Client as LndClient;
 use tracing::{info, warn};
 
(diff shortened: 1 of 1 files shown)


devimint/src/external.rs
```diff
@@ -373,6 +374,60 @@ pub async fn open_channel(bitcoind: &Bitcoind, cln: &Lightningd, lnd: &Lnd) -> R
...
+            .expect("No channels returned")
+            .chan_id;
+        let chan_info = lnd_client
+            .lightning()
+            .get_chan_info(ChanInfoRequest { chan_id })
+            .await;
+
+        if chan_info.is_err() {
+            return Ok(false);
+        }
```
elsirion, line 405: Is this expected to happen? If not we might want to log it?

devimint/src/external.rs
```diff
@@ -373,6 +374,60 @@ pub async fn open_channel(bitcoind: &Bitcoind, cln: &Lightningd, lnd: &Lnd) -> R
...
+            .await?
+            .into_inner();
+
+        if channels.channels.is_empty() {
+            return Ok(false);
+        }
+
+        let chan_id = channels
+            .channels
+            .get(0)
```
elsirion, line 395: Not waiting for all channels could lead to flakiness if we ever open more because sometimes the other channels might be ready and sometimes not, leading to `update_channel_policy` taking effect on them or not.

devimint/src/external.rs
m1sterc001guy: It is expected to happen, since we try to update the policy immediately after the channel is open. Rather than add an artbitrary sleep, I figured it was better to just try and then re-try if LND doesn't see the channel yet.

devimint/src/external.rs
m1sterc001guy: I reworked this a bit, now it explicitly waits for a channel with the `remote_public_key` that corresponds to CLN. That should help prevent flakiness in the future

elsirion (approved):

PR #3037, opened 2023-08-27
 chore: optimize pre-commit check

dpc:
Feels slow. Bothers me.

### DIFF
--- a/justfile
+++ b/justfile
@@ -70,7 +70,8 @@ typos:
   set -eo pipefail
 
   git_ls_files="$(git ls-files)"
-  git_ls_nonbinary_files="$(echo "$git_ls_files" | xargs file --mime | grep -v "; charset=binary" | cut -d: -f1)"
+  git_ls_nonbinary_files="$(echo "$git_ls_files" |  grep -v -E "^db/|\.png\$|\.ods\$")"
+
 
   if ! echo "$git_ls_nonbinary_files" | parallel typos {} ; then
     >&2 echo "Typos found: Valid new words can be added to '_typos.toml'"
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -70,7 +70,8 @@ typos:
   set -eo pipefail
 
   git_ls_files="$(git ls-files)"
-  git_ls_nonbinary_files="$(echo "$git_ls_files" | xargs file --mime | grep -v "; charset=binary" | cut -d: -f1)"
+  git_ls_nonbinary_files="$(echo "$git_ls_files" |  grep -v -E "^db/|\.png\$|\.ods\$")"
```
elsirion, line 73: This regex will be slightly annoying to maintain, does anything bad happen if there are binary files in that list?

misc/git-hooks/pre-commit
```diff
@@ -199,14 +199,16 @@ function check_typos() {
 }
 export -f check_typos
 
-parallel ::: \
+parallel \
+  --jobs 16 \
```
elsirion, line 203: I take it that these tasks are mostly io bound and so using more processes than cores is ok?

elsirion (approved):
Just some nits, feel free to merge.

justfile
dpc: Some lints like `typos` will complain that they see strings they don't like, pointing us to update the list.

I wanted to avoid manual list, but `xargs file --mime | grep -v "; charset=binary"` costs around a second on every `git commit` on my machine.

misc/git-hooks/pre-commit
dpc: I think so, yes.

PR #3038, opened 2023-08-27
chore: use faster `typos` versions

dpc:
In combination with #3037 it makes commit check even faster, especially on cold cache.

Before (`master` branch):

```
Executed in    4.86 secs    fish           external
   usr time   25.39 secs    0.00 micros   25.39 secs
   sys time   10.23 secs  466.00 micros   10.23 secs
```

After (#3037 + #3038):

```
Executed in    1.33 secs    fish           external
   usr time    8.23 secs   20.29 millis    8.21 secs
   sys time    1.54 secs    4.15 millis    1.54 secs
```

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,2 +1,4 @@
 [default.extend-words]
 substituters = "substituters"
+# trips on some base64 string in the test
+wel = "wel"
--- a/devimint/src/lib.rs
+++ b/devimint/src/lib.rs
@@ -201,6 +201,7 @@ pub async fn dev_fed(process_mgr: &ProcessManager) -> Result<DevFed> {
     cmd!(fed, "join-federation", fed.invite_code()?)
         .run()
         .await?;
+    info!(LOG_DEVIMINT, "await gateways registered");
     fed.await_gateways_registered().await?;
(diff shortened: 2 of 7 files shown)


devimint/src/lib.rs
```diff
@@ -201,6 +201,7 @@ pub async fn dev_fed(process_mgr: &ProcessManager) -> Result<DevFed> {
     cmd!(fed, "join-federation", fed.invite_code()?)
         .run()
         .await?;
+    info!(LOG_DEVIMINT, "await gateways registered");
```
elsirion, line 204: unrelated, but ok

_typos.toml
```diff
@@ -1,2 +1,4 @@
 [default.extend-words]
 substituters = "substituters"
+# trips on some base64 string in the test
+wel = "wel"
```
elsirion, line 4: Wel, so be it … :rofl:

elsirion (approved):

devimint/src/lib.rs
dpc: I have no idea how did this got here really... . It doesn't hurt, but ... weird.

PR #3040, opened 2023-08-28
fix(nix): lint shell broken as semgrep not supported on MacOS

dpc:


elsirion (approved):
*Over the (at least) last 5 months nobody ran lint on Mac :see_no_evil:*

PR #3041, opened 2023-08-28
feat: tag docker with release tag

douglaz:
Let docker-compose files use a tagged release.

### DIFF
--- a/docker/full-tls-mutinynet/docker-compose.yaml
+++ b/docker/full-tls-mutinynet/docker-compose.yaml
@@ -32,7 +32,7 @@ services:
       - "/var/run/docker.sock:/var/run/docker.sock:ro"
 
   fedimintd:
-    image: fedimint/fedimintd:master
+    image: fedimint/fedimintd:v0.1.0-rc3
     volumes:
       - fedimintd_data:/data
     ports:
@@ -71,7 +71,7 @@ services:
       - "traefik.http.routers.guardian-ui.tls.certresolver=myresolver"
 
   gatewayd:
-    image: fedimint/gatewayd:master
+    image: fedimint/gatewayd:v0.1.0-rc3
     command: gatewayd lnd
     environment:
       # Path to folder containing gateway config and data files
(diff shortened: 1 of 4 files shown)


elsirion:
Maybe let's do this after #3043?

douglaz:
Using the new tag

dpc (approved):

docker/full-tls-mutinynet/docker-compose.yaml
```diff
@@ -32,7 +32,7 @@ services:
       - "/var/run/docker.sock:/var/run/docker.sock:ro"
 
   fedimintd:
-    image: fedimint/fedimintd:master
+    image: fedimint/fedimintd:v0.1.0-rc3
```
elsirion, line 35: This doesn't look sustainable. I should have thought of that when ACKing #3044. I still think an `v0.1-latest` tag would be better. Sure, as @dpc pointed out the image will get cached, but if you really want the latest version you should probably just manually pull it again.

scripts/dev/change-docker-tags.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [ -z "$files" ]; then
+  echo "No docker-compose files found in $path"
+  exit 1
+fi
+
+for file in $files; do
+  sed -ri "s/image: fedimint\/(fedimintd|gatewayd):.*/image: fedimint\/\1:$new_tag/g" "$file"
+done
+
```
elsirion, line 16: Waiting for someone to build a workflow that runs this script+opens a PR whenever a tag is created on a release branch :laughing:

elsirion (approved):
Begrudgingly ACKing till we have a better idea …

docker/full-tls-mutinynet/docker-compose.yaml
douglaz: I think any upgrade to a federation should be done carefully and manually. So here we should specify a very specific tag/hash.

PR #3046, opened 2023-08-28
chore: build with lld on more systems

dpc:


elsirion (approved):
Fine by me, if it doesn't work we can always revert. Why didn't we do so in the first place?

elsirion:
Test seems to fail repeatedly.

dpc:
> Fine by me, if it doesn't work we can always revert. Why didn't we do so in the first place?

> Test seems to fail repeatedly.

dpc:
Such a waste of time. :D

PR #3047, opened 2023-08-28
chore: *do* strip symbols on MacOS

dpc:
It was reported that it trips something somewhere on Macs which leads to segfault.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -119,7 +119,7 @@
               inherit name;
 
               dontUnpack = true;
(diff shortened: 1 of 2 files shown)


justinmoon:
It's not a segfault, the OS kills the binary when you try to execute it. Similar to a problem we saw last summer with bitcoind https://github.com/NixOS/nixpkgs/issues/179474.

justinmoon (approved):

Issue #2786, opened 2023-07-19
Gateway password stored in database


justinmoon:
Currently the gateway password is set as an environment variable. This prevents user of the gateway UI from being able to set a password via the UI.

Let's add an API to create a password for the gateway. Store this password in the database. Only allow creating password if none exists in database. Leave the environment variable as an option but the password set in the API has priority.

justinmoon:
Call: this is part of broader project to be able to update configs like fees "fees" at runtime

PR #3048, opened 2023-08-29
refactor: Decouple Web Server from Lightning Connection

m1sterc001guy:
Working towards: https://github.com/fedimint/fedimint/issues/2786

Currently, the gateway's webserver is tied to the connection status of the lightning node. This is because there are some operations that the webserver needs a lightning connection to do (like fetching route hints). However, this makes it difficult to allow for "configuration" APIs that might be needed before the lightning connection is even established.

This PR decouples the webserver from the lightning connection. The webserver is now long-lived, but has a `GatewayState` struct that allows it to be in different states. Some APIs will not be accessible when the Gateway is in certain states (i.e `GatewayState::Disconnected`). In the future, we can add states like `Configuring` to allow for a configuration phase before connecting to the lightning node. This also allows us to cache Lightning connection data structures in the `Running` phase.

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -11,7 +11,7 @@ use cln_rpc::ClnRpc;
 use fedimint_core::encoding::Encodable;
 use fedimint_core::task::{block_in_place, sleep};
 use fedimint_logging::LOG_DEVIMINT;
-use fedimint_testing::gateway::LightningNodeName;
+use fedimint_testing::gateway::LightningNodeType;
 use futures::executor::block_on;
 use tokio::fs;
 use tokio::sync::{MappedMutexGuard, Mutex, MutexGuard};
@@ -444,10 +444,10 @@ pub enum LightningNode {
 }
 
 impl LightningNode {
-    pub fn name(&self) -> LightningNodeName {
+    pub fn name(&self) -> LightningNodeType {
         match self {
-            LightningNode::Cln(_) => LightningNodeName::Cln,
-            LightningNode::Lnd(_) => LightningNodeName::Lnd,
+            LightningNode::Cln(_) => LightningNodeType::Cln,
+            LightningNode::Lnd(_) => LightningNodeType::Lnd,
         }
     }
 }
--- a/fedimint-testing/src/gateway.rs
+++ b/fedimint-testing/src/gateway.rs
@@ -1,29 +1,28 @@
-use std::collections::BTreeMap;
 use std::fmt::{Display, Formatter};
 use std::net::SocketAddr;
 use std::sync::Arc;
 
+use async_trait::async_trait;
 use fedimint_client::module::init::ClientModuleInitRegistry;
 use fedimint_client::Client;
-use fedimint_client_legacy::modules::ln::config::GatewayFee;
 use fedimint_core::config::FederationId;
 use fedimint_core::db::mem_impl::MemDatabase;
 use fedimint_core::db::Database;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
-use fedimint_core::task::TaskGroup;
 use lightning::routing::gossip::RoutingFees;
 use ln_gateway::client::StandardGatewayClientBuilder;
+use ln_gateway::lnrpc_client::{ILnRpcClient, LightningBuilder};
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
-use ln_gateway::rpc::rpc_server::run_webserver;
 use ln_gateway::rpc::{ConnectFedPayload, FederationInfo};
 use ln_gateway::Gateway;
 use secp256k1::PublicKey;
 use tempfile::TempDir;
-use tokio::sync::RwLock;
 use url::Url;
 
 use crate::federation::FederationTest;
-use crate::fixtures::test_dir;
+use crate::fixtures::{test_dir, Fixtures};
+use crate::ln::mock::FakeLightningTest;
+use crate::ln::real::{ClnLightningTest, LndLightningTest};
 use crate::ln::LightningTest;
 
 /// Fixture for creating a gateway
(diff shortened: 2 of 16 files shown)


fedimint-testing/src/gateway.rs
```diff
@@ -86,46 +87,59 @@ impl GatewayTest {
...
 
-        let listening_addr = lightning.listening_address();
-        let info = lightning.info().await.unwrap();
-
-        let mut tg = TaskGroup::new();
-        // Create the stream to route HTLCs. We cannot create the Gateway until the
-        // stream to the lightning node has been setup.
-        let (stream, ln_client) = lightning.route_htlcs(&mut tg).await.unwrap();
+        let lightning_builder: Arc<dyn LightningBuilder + Send + Sync> =
+            match Fixtures::is_real_test() {
```
douglaz, line 91: if/else is much better than match true/false

fedimint-testing/src/gateway.rs
```diff
@@ -86,46 +87,59 @@ impl GatewayTest {
...
+            {
+                break;
+            }
+
+            if gateway_state_iterations > 9 {
+                panic!("Gateway did not start running after 10 attempts");
+            }
+
+            gateway_state_iterations += 1;
+            sleep(Duration::from_millis(100)).await;
```
douglaz, line 138: This is a very low waiting period, it will fail on CI

gateway/ln-gateway/src/lib.rs
```diff
@@ -251,30 +310,22 @@ impl Gatewayd {
...
-                                let task_group = run_webserver(self.password.clone(), self.listen, gateway)
-                                    .await
-                                    .expect("Failed to start webserver");
-                                info!("Successfully started webserver");
-
-                                Gateway::handle_htlc_stream(stream, ln_client, handle.clone(), scid_to_federation.clone(), clients.clone()).await;
-                                warn!("HTLC Stream Lightning connection broken. Stopping webserver...");
-                                task_group.shutdown();
+                                let (lightning_public_key, lightning_alias) =
+                                Self::fetch_lightning_node_info(ln_client.clone()).await.expect("Failed to retrieve Lightning info after establishing a connection");
```
douglaz, line 314: Perhaps retry this and other operations below? But this could be another PR

gateway/ln-gateway/src/lib.rs
```diff
@@ -860,6 +492,140 @@ impl Gateway {
...
+        payload: ConnectFedPayload,
+    ) -> Result<FederationInfo> {
+        if let GatewayState::Running {
+            lnrpc,
+            lightning_public_key,
+            lightning_alias: _,
+        } = self.state.read().await.clone()
+        {
+            let invite_code = InviteCode::from_str(&payload.invite_code).map_err(|e| {
+                GatewayError::InvalidMetadata(format!("Invalid federation member string {e}"))
```
douglaz, line 553: Use the debug format so the stack trace will be included
```suggestion
                GatewayError::InvalidMetadata(format!("Invalid federation member string {e:?}"))
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -873,4 +639,250 @@ impl Gateway {
...
+                                        .await
+                                        .is_err()
+                                    {
+                                        error!("Error registering federation {federation_id}");
+                                    }
+                                }
+                            }
+                            Err(_) => {
+                                error!(
+                                    "Could not retrieve route hints, gateway will not be registered."
```
douglaz, line 756: include the error?

gateway/ln-gateway/src/lib.rs
```diff
@@ -873,4 +639,250 @@ impl Gateway {
...
+                                        .register_with_federation(
+                                            api.clone(),
+                                            route_hints.clone(),
+                                            GW_ANNOUNCEMENT_TTL,
+                                            gateway_id,
+                                        )
+                                        .await
+                                        .is_err()
+                                    {
+                                        error!("Error registering federation {federation_id}");
```
douglaz, line 750: Include the error?

fedimint-testing/src/gateway.rs
m1sterc001guy: I increased it to 10 seconds, waiting a second each time.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: The likelihood of this failing is hopefully small since just above it established the intercepted HTLC stream, but I added the retry anyway.

gateway/ln-gateway/src/lib.rs
```diff
@@ -251,249 +310,82 @@ impl Gatewayd {
...
-
-                                let task_group = run_webserver(self.password.clone(), self.listen, gateway)
-                                    .await
-                                    .expect("Failed to start webserver");
-                                info!("Successfully started webserver");
-
-                                Gateway::handle_htlc_stream(stream, ln_client, handle.clone(), scid_to_federation.clone(), clients.clone()).await;
-                                warn!("HTLC Stream Lightning connection broken. Stopping webserver...");
-                                task_group.shutdown();
+                                if let Ok((lightning_public_key, lightning_alias)) =
```
douglaz, line 313: `match` here and log the `Err(e)`. Avoid swallowing errors

gateway/ln-gateway/src/lib.rs
```diff
@@ -860,6 +497,140 @@ impl Gateway {
         ))
     }
 
+    async fn handle_pay_invoice_msg(&self, payload: PayInvoicePayload) -> Result<Preimage> {
+        if let GatewayState::Running {
+            lnrpc: _,
+            lightning_public_key: _,
+            lightning_alias: _,
+        } = self.state.read().await.clone()
```
elsirion, line 505: Could be a let-else statement to avoid nesting.

elsirion (approved):

devimint/src/external.rs
```diff
@@ -444,10 +444,10 @@ pub enum LightningNode {
 }
 
 impl LightningNode {
-    pub fn name(&self) -> LightningNodeName {
+    pub fn name(&self) -> LightningNodeType {
```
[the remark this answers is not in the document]
Jjajsja: https://github.com/fedimint/fedimint/issues/2786#issue-1812243842

PR #3053, opened 2023-08-29
chore: set consensus version on master to `u32::MAX`

elsirion:
This will prevent accidentally using master with deployed federations.

Related to #3052

PR #3054, opened 2023-08-29
chore: set consensus version to 1

elsirion:
This will prevent accidentally mixing up the first release with federations previously deployed off random master commits.

Related to #3053

### DIFF
--- a/fedimint-server/src/config/mod.rs
+++ b/fedimint-server/src/config/mod.rs
@@ -209,7 +209,7 @@ impl ServerConfigConsensus {
     }
 }
 
-pub const CORE_CONSENSUS_VERSION: CoreConsensusVersion = CoreConsensusVersion(0);
+pub const CORE_CONSENSUS_VERSION: CoreConsensusVersion = CoreConsensusVersion(1);
 
 impl ServerConfig {
     /// Api versions supported by this server
--- a/modules/fedimint-dummy-server/src/lib.rs
+++ b/modules/fedimint-dummy-server/src/lib.rs
@@ -63,7 +63,7 @@ impl ServerModuleInit for DummyGen {
     }
 
     fn supported_api_versions(&self) -> SupportedModuleApiVersions {
-        SupportedModuleApiVersions::from_raw(0, 0, &[(0, 0)])
+        SupportedModuleApiVersions::from_raw(1, 0, &[(0, 0)])
     }
 
     /// Initialize the module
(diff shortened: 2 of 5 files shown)


justinmoon:
Should we also update module versions?

elsirion:
Debugging the test failure:
* The client seems to assume all its modules are incompatible, so tries to initialize without a primary module, which fails
* Using `websocat` I get the following version info:
```json
{"jsonrpc": "2.0", "method": "version", "id":"0", "params": [{"params": null}]}
{"jsonrpc":"2.0","result":{"core":{"core_consensus":1,"api":[{"major":0,"minor":0}]},"modules":{"0":{"core_consensus":0,"module_consensus":0,"api":[{"major":0,"minor":0}]},"1":{"core_consensus":0,"module_consensus":0,"api":[{"major":0,"minor":0}]},"2":{"core_consensus":0,"module_consensus":0,"api":[{"major":0,"minor":0}]}}},"id":"0"}
```
<details>
<summary>Formatted result</summary>

```json
{
  "core": {
    "core_consensus": 1,
    "api": [
      {
        "major": 0,
        "minor": 0
      }
    ]
  },
  "modules": {
    "0": {
      "core_consensus": 0,
      "module_consensus": 0,
      "api": [
        {
          "major": 0,
          "minor": 0
        }
      ]
    },
    "1": {
      "core_consensus": 0,
      "module_consensus": 0,
      "api": [
        {
          "major": 0,
          "minor": 0
        }
      ]
    },
    "2": {
      "core_consensus": 0,
      "module_consensus": 0,
      "api": [
        {
          "major": 0,
          "minor": 0
        }
      ]
    }
  }
}
```
</details>
* I guess the per-module `core_consensus` value is wrong? Why is it needed anyway? Is it ever expected to be not in line with the main `core_consensus`? Should such a `fedimintd` even start (or compile)?

dpc:
[This](https://github.com/fedimint/fedimint/blob/d4ae51df9149dd5e7f3fef16f7e5f04c04ca117d/fedimint-client/src/lib.rs#L840) and [this](https://github.com/fedimint/fedimint/blob/d4ae51df9149dd5e7f3fef16f7e5f04c04ca117d/fedimint-client/src/lib.rs#L854) looks wrong. We should be taking these from the constant of the build we're building.

dpc (approved):

dpc (approved):

PR #3064, opened 2023-08-30
chore: surface needless rebuilds

dpc:
Problem causing these is described in https://github.com/ipetkov/crane/issues/370 , but I actually just need to to show up so I remember about it.


elsirion (approved):
Happy hunting! :lady_beetle:

PR #3065, opened 2023-08-30
chore: add backport bot

elsirion:
This workflow is intended to automate the backport process in cases where cherry-picking succeeds without merge conflicts. For more details see [the action's README](https://github.com/korthout/backport-action#how-it-works).

## How it works

I created a tag `backport releases/v0.1`. If this is added before merging it should work according to the docs. I also changed the reference workflow slightly so that it should also work if the tag is added after merging.


dpc (approved):
Cool

PR #3066, opened 2023-08-30
[WIP] chore: adding more logs and trying to remove flakiness

douglaz:


devimint/src/external.rs
```diff
@@ -77,16 +77,18 @@ impl Bitcoind {
...
 
     pub async fn mine_blocks(&self, block_num: u64) -> Result<()> {
-        info!(block_num, "Mining bitcoin blocks");
+        info!(?block_num, "Mining bitcoin blocks");
         let client = self.client();
         let addr = client.get_new_address(None, None)?;
         let initial_block_count = client.get_block_count()?;
         client.generate_to_address(block_num, &addr)?;
         while client.get_block_count()? < initial_block_count + block_num {
-            trace!(LOG_DEVIMINT, "Waiting for blocks to be mined");
```
elsirion, line 86: Please don't remove `LOG_DEVIMINT`, add the required `taret: ` instead. 

https://github.com/fedimint/fedimint/blob/9a14f7b8379c13546821ae91cf0067e27e0f4426/devimint/src/external.rs#L510

@dpc had a good reason for targets iirc.

dpc: In devimint it would allow setting one `RUST_LOG` that has filter for devimint and stuff it starts and calls.

devimint/src/external.rs
```diff
@@ -77,16 +77,18 @@ impl Bitcoind {
     }
 
     pub async fn mine_blocks(&self, block_num: u64) -> Result<()> {
-        info!(block_num, "Mining bitcoin blocks");
+        info!(?block_num, "Mining bitcoin blocks");
```
dpc, line 80: Is there a difference?

modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -58,14 +58,15 @@ async fn peg_in<'a>(
...
+    info!(?address, "Peg-in address generated");
     let (_proof, tx) = bitcoin
         .send_and_mine_block(&address, bsats(PEG_IN_AMOUNT_SATS))
         .await;
     let height = dyn_bitcoin_rpc
         .get_tx_block_height(&tx.txid())
         .await?
         .context("expected tx to be mined")?;
-    info!(?height, "Peg-in transaction mined");
+    info!(?height, ?tx, "Peg-in transaction mined");
```
dpc, line 69: Is this going to be ... longish? Should we use one of the abbreviation wrappers?

dpc (commented):
Some comments comments, but generally solid improvement.

devimint/src/external.rs
douglaz: > The ? sigil is shorthand that specifies a field should be recorded using its [fmt::Debug](https://doc.rust-lang.org/nightly/core/fmt/trait.Debug.html) implementation

But in this case it doesn't make any difference

modules/fedimint-wallet-tests/tests/tests.rs
elsirion: This isn't as easy as it appears, maybe the `Debug` impl of `Transaction` should do that automatically? Or at least impl `AbbreviatedDebug`?

devimint/src/external.rs
elsirion: I took the liberty to resolve this

elsirion (approved):

PR #3070, opened 2023-08-31
Add LightningAuditItemKeys and db compaction

### LINKED ISSUE #2605: Audit functionality extremely slow

bradleystachurski:
Fixes #2605 

The associated issue shows calculating an audit takes a long time on alpha.fedi.xyz (> 8s!). This PR improves the time it takes to run an audit by
- Adding a separate key `LightningAuditItemKey` that associates an Incoming/Outgoing contract to an outstanding Amount
- Compacting (removing) the audit key once the associated contract reaches a terminal state

---
## Benchmarks

The benchmarks use devimint and run ~1 hour worth of iterations where an iteration consists of completing incoming and outgoing payments using both the CLN and LND gateways.

Benchmarks were calculated in the branch https://github.com/bradleystachurski/fedimint/tree/audit-benchmark and can be replicated by checking out the appropriate commit (baseline or including changes) and running

```bash
./scripts/tests/audit-benchmark.sh
```

### Baseline
Benchmark - Baseline

We can clearly observe the time it takes to complete an audit grows linearly as we increase the amount of contracts we need to query while running an audit.

### Using Audit Keys and Compaction
Benchmark - Including Audit Keys _ Compaction

Running the benchmark using the changes in this PR, we observe constant time to run an audit regardless of the growth in total contracts.


m1sterc001guy:
What does DB compaction mean in this context? LSM trees/RocksDB does DB compaction but I suspect that's not what you're referring to...

bradleystachurski:
> What does DB compaction mean in this context? LSM trees/RocksDB does DB compaction but I suspect that's not what you're referring to...

Good question! Your hunch is correct, I'm using the term slightly different. Compaction in this context refers to removing keys that are no longer relevant for calculating an audit. When an incoming/outgoing contract reaches a terminal state, we no longer need to query the db for the associated amount, so this PR adds a new key specific to ln audits and removes the unnecessary keys.

Let me know if you think there's a better term to use that doesn't clash with existing concepts. I was leaning on the wikipedia [definition](https://en.wikipedia.org/wiki/Data_compaction)

> In [telecommunication](https://en.wikipedia.org/wiki/Telecommunication), data compaction is the reduction of the number of [data elements](https://en.wikipedia.org/wiki/Data_element), [bandwidth](https://en.wikipedia.org/wiki/Bandwidth_(computing)), cost, and time for the generation, [transmission](https://en.wikipedia.org/wiki/Transmission_(telecommunications)), and [storage of data](https://en.wikipedia.org/wiki/Computer_storage) without loss of [information](https://en.wikipedia.org/wiki/Information) by eliminating unnecessary [redundancy](https://en.wikipedia.org/wiki/Redundancy_(information_theory)), removing irrelevancy, or using special coding.

I'll have an updated PR description a bit later today once I make some benchmark graphs 😄

modules/fedimint-ln-common/src/db.rs
```diff
@@ -57,6 +58,41 @@ impl_db_lookup!(
...
 );
 
+/// We keep a separate mapping of incoming and outgoing ContractIds to Amounts,
+/// which allows us to quickly audit the total liabilities in the Lightning
+/// module
+#[derive(Debug, Clone, Encodable, Decodable, Serialize, PartialEq)]
+pub enum LightningAuditItemKey {
+    Incoming(ContractId),
+    Outgoing(ContractId),
+}
```
bradleystachurski, line 68: The reviewer may notice this differs from `MintAuditItemKey`s, since it doesn't include an aggregate `*Total` key. The motivation for not including the aggregate key is how the Amount associated to the contract [mutates](https://github.com/bradleystachurski/fedimint/blob/4b28e27831ab9c914eb4df68b994e717a5b0d3e1/modules/fedimint-ln-server/src/lib.rs#L624) in the LN module. When a contract reaches a terminal state, the associated amount updates to 0. The additional complexity to update both the individual incoming/outgoing contract audit keys along with the aggregate audit key didn't seem necessary.

In contrast to the mint module, the total number of LN audit keys with a non-zero amount will not grow linearly, so querying the LN audit keys with a non-zero amount should remain quick. I ran benchmark (not included in the PR description) that only uses the audit keys without compaction, which showed the time to run an audit also remained constant. As an aside, that also indicates the root cause for slow audit times is slow de/serialization in the `ContractKey` -> `ContractAccount` db records.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -885,11 +919,60 @@ impl ServerModule for Lightning {
...
+                Self::compact_audit_keys_task(db, &handle).await;
+            })
+            .await;
         Ok(Lightning { cfg, btc_rpc })
     }
 
+    async fn compact_audit_keys_task(db: Database, tg_handle: &TaskHandle) {
+        let interval = Duration::from_secs(60 * 5);
+        while !tg_handle.is_shutting_down() {
+            let mut dbtx = db.begin_transaction().await;
```
dpc, line 939: Can this possibly interfere with rest of the module and cause `commit_tx` failures elsewhere?

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -885,11 +919,60 @@ impl ServerModule for Lightning {
...
+    ///
+    /// When a contract reaches a terminal state, the associated amount will be
+    /// updated to 0. At this point, the contract no longer needs to be tracked
+    /// for auditing liabilities, so we can safely remove the audit key.
+    async fn compact_audit_keys(dbtx: &mut DatabaseTransaction<'_>) {
+        let _timing /* logs on drop */ = TimeReporter::new("compact_audit_keys");
+        let remove_audit_keys = dbtx
+            .find_by_prefix(&LightningAuditItemKeyPrefix)
+            .await
+            .filter_map(|(key, amount)| async move {
```
dpc, line 961: nit: this could be just `filter(|(_, amount) amount == 0)`? It doesn't seem to transform what it's filtering.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -885,11 +919,60 @@ impl ServerModule for Lightning {
...
+                if amount.msats == 0u64 {
+                    Some(key)
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+            .await;
+
+        for key in remove_audit_keys {
```
dpc, line 971: Could this collection ever get big enough to use a stream and `.for_each(/* remove */)` instead of `collect`?

modules/fedimint-ln-common/src/db.rs
dpc: Consider putting this comment in the code. :)

dpc: Consider putting this comment in the code. :)

dpc (commented):
LGTM + comments/questions.

modules/fedimint-ln-server/src/lib.rs
elsirion: Yes, this is a problem, we don't handle commit failures because we decided that consensus DB entries should not be written to concurrently. Could we instead just check after each `process_input` if a contract has become 0-value?

I'd even go a step further and require eventually that all funds be withdrawn at once.

m1sterc001guy:
LGTM as well, agree with Eric that the compaction shouldn't happen on a separate thread since the commit could trigger transaction failures due to the write-write conflict.

Is the audit benchmark something that we can add to the tests and run during CI? That would ensure we don't regress this logic in the future (love the graphs).

bradleystachurski:
Thanks everyone for the quick reviews! I should be able to address feedback this weekend. Wanted to prioritize some of the issues with the audit RPC ([PR](https://github.com/fedimint/fedimint/pull/3090)).

modules/fedimint-ln-server/src/lib.rs
bradleystachurski: Oof, great catch! I've moved the logic to remove 0-value audit keys as the last step in `process_input`.

>I'd even go a step further and require eventually that all funds be withdrawn at once.

Eric, does this comment mean if there's a commit failure for consensus DB entries it would trigger a force withdrawal of all funds from the federation?

modules/fedimint-ln-server/src/lib.rs
bradleystachurski: I wanted to return only the key instead of a tuple of `(key, amount)` so `filter_map` was used here to transform the return type. That said, it may be clearer to use only `filter` and handle the tuple downstream, but this logic has been removed with the latest update.

modules/fedimint-ln-server/src/lib.rs
bradleystachurski: That's a great point and led to an interesting puzzle. The compaction background task has been removed so it's no longer relevant for this PR, but for future reference do you know of a way to reference the dbtx inside of a stream that begins from a dbtx?

I didn't continue investigating, but perhaps it's possible to create an additional `DatabaseTransaction` method that's similar to `find_by_prefix` but also returns the dbtx as part of the stream?

This is the implementation and compiler error.

```rust
        dbtx.find_by_prefix(&LightningAuditItemKeyPrefix)
            .await
            .filter_map(|(key, amount)| async move {
                if amount.msats == 0u64 {
                    Some(key)
                } else {
                    None
                }
            })
            .for_each(|key| async move {
                dbtx.remove_entry(&key);
            })
            .await;
```

```
error[E0507]: cannot move out of `dbtx`, a captured variable in an `FnMut` closure
    --> modules/fedimint-ln-server/src/lib.rs:1007:29
     |
956  |       async fn compact_audit_keys(dbtx: &mut DatabaseTransaction<'_>) {
     |                                   ---- captured outer variable
...
1007 |               .for_each(|key| async move {
     |  _______________________-----_^
     | |                       |
     | |                       captured by this `FnMut` closure
1008 | |                 dbtx.remove_entry(&key);
     | |                 ----
     | |                 |
     | |                 variable moved due to use in generator
     | |                 move occurs because `dbtx` has type `&mut DatabaseTransaction<'_>`, which does not implement the `Copy` trait
1009 | |             })
     | |_____________^ `dbtx` is moved here
```

Perhaps something similar to `.fold` may work if we added a method to `DatabaseTransaction` that returns the results of `find_by_prefix` along with the original dbtx?

```rust
        dbtx.find_by_prefix(&LightningAuditItemKeyPrefix)
            .await
            .filter_map(|(key, amount)| async move {
                if amount.msats == 0u64 {
                    Some(key)
                } else {
                    None
                }
            })
            .fold(dbtx, |dbtx, key| async move {
                dbtx.remove_entry(&key);
                dbtx
            })
            .await;
```

```
error[E0505]: cannot move out of `dbtx` because it is borrowed
    --> modules/fedimint-ln-server/src/lib.rs:1011:19
     |
956  |     async fn compact_audit_keys(dbtx: &mut DatabaseTransaction<'_>) {
     |                                 ---- binding `dbtx` declared here
...
998  |         dbtx.find_by_prefix(&LightningAuditItemKeyPrefix)
     |         ------------------------------------------------- borrow of `*dbtx` occurs here
...
1011 |             .fold(dbtx, |dbtx, key| async move {
     |              ---- ^^^^ move out of `dbtx` occurs here
     |              |
     |              borrow later used by call
```

bradleystachurski:
> Is the audit benchmark something that we can add to the tests and run during CI? That would ensure we don't regress this logic in the future

I've been chewing on this and it's not immediately obvious how to add a benchmark test to CI that will complete fast enough (the benchmarks took ~1 hour locally), have enough data points for a statistically significant regression, and doesn't trigger false failures due to CI flakiness (https://github.com/fedimint/fedimint/issues/3107 amongst others).

I think it might be best to focus efforts on improving the performance of de/serialization then adding benchmark tests to ensure that remains a fast operation. We should be able to de/serialize a large set of contracts (~10,000?) fairly fast in CI and make sure the time of the round trip stays under a limit.

I'm open to suggestions if anyone has thoughts on other approaches.

modules/fedimint-ln-common/src/db.rs
```diff
@@ -57,6 +58,51 @@ impl_db_lookup!(
...
+///
+/// This differs from MintAuditItemKeys, since it doesn't include an aggregate
+/// *Total key. The motivation for not including the aggregate key is how the
+/// Amount associated to the contract mutates in the LN module. When a contract
+/// reaches a terminal state, the associated amount updates to 0. The additional
+/// complexity to update both the individual incoming/outgoing contract audit
+/// keys along with the aggregate audit key is not necessary.
+//
+// In contrast to the mint module, the total number of LN audit keys with a non-zero amount will not
+// grow linearly, so querying the LN audit keys with a non-zero amount should remain quick.
```
elsirion, line 73: Is the missing `/` intentional, i.e. you don't want this to be displayed in the docs?

elsirion (approved):
Looks good, thx for the great PR! I'll wait for if you want to fix the one nit.

modules/fedimint-ln-common/src/db.rs
bradleystachurski: Good eye! That was unintentional :smiley:

elsirion (approved):

PR #3072, opened 2023-08-31
feat: module_json RPC command

justinmoon:
This will help @wbobeirne finish the guardian MVP implementation. See [discussion on discord](https://discord.com/channels/990354215060795454/990354215878688860/1146567059610808330).

### DIFF
--- a/fedimint-server/src/net/api.rs
+++ b/fedimint-server/src/net/api.rs
@@ -10,7 +10,7 @@ use fedimint_core::api::{
     FederationStatus, InviteCode, PeerConnectionStatus, PeerStatus, ServerStatus, StatusResponse,
 };
 use fedimint_core::backup::ClientBackupKey;
-use fedimint_core::config::{ClientConfig, ClientConfigResponse};
+use fedimint_core::config::{ClientConfig, ClientConfigResponse, JsonWithKind};
 use fedimint_core::core::backup::SignedBackupRequest;
 use fedimint_core::core::ModuleInstanceId;
 use fedimint_core::db::{Database, DatabaseTransaction, ModuleDatabaseTransaction};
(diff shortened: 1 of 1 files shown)


fedimint-server/src/net/api.rs
```diff
@@ -577,11 +577,17 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
...
         api_endpoint! {
             "auth",
-            async |_config: &ConsensusApi, context, _v: ()| -> () {
+            async |_fedimint: &ConsensusApi, context, _v: ()| -> () {
                 check_auth(context)?;
                 Ok(())
             }
         },
+        api_endpoint! {
+            "modules_json",
```
dpc, line 586: The name seems off. Probably should have a "config" word in it. Should this be maybe `config_json` (to match `config`) and maybe return whole config, but with module configs as json or something? Just thinking aloud.

justinmoon: My thinking was that the contents of the client config are already available, so putting them in here would be redundant. Further, it's kind of nice that this can't be accidentally used as a client config which might mess up config signatures somehow?

Also, choose not to just stick all this in the config itself because that is some extra IO that every client would need to do on startup to load the config. Would also be a breaking change that couldn't be applied to release branch backwards-compatibly.

justinmoon: `config_json` doesn't make sense because the config is already json. It's just that the `modules` section has consensus hex-encoded values in it.

dpc: `modules_config_json` then? :D

justinmoon: 🤝 updated

dpc:
Due to crane/cargo issue I've been hunting down code coverage is doing some more rebuilds which makes it timeout sometimes. You might want to increase it to 45 seconds (I did in another PR, but it is waiting for review).

dpc:
https://github.com/fedimint/fedimint/pull/3075

dpc (approved):

PR #3077, opened 2023-08-31
Start externals with `devimint run-ui`

wbobeirne:
It's handy to have things like esplora and the lightning nodes in our development testing, so this starts them with `run-ui` just like `dev-fed` does.

Also added a bit of extra logging to when LND fails to connect, since that dragged me down for ~30m trying to figure that out since it fails silently.

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -271,13 +271,16 @@ impl Lnd {
         .await?;
 
         poll("lnd_connect", || async {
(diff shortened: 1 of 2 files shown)


justinmoon (approved):
This is useful for developing guardian UI. It allows us to actually use the federation via `fedimint-cli` once it's setup.

dpc (approved):

PR #3079, opened 2023-09-01
Expose JSON encoded client config

elsirion:
Large parts of the client config are binary-encoded, even when transported via JSON RPC, so it is easier to deterministically apply cryptographic operations to them without re-encoding (hashing, signing). Unfortunately this leads to difficulties when debugging or integrating with software that doesn't have access to Fedimint's encoding framework.

This PR exposes a JSON version of the client config both in the CLI as well as via the API:
* CLI: `fedimint-cli config`: fixes #3074 
* ~API: `fedimint-cli dev api config_json`: fixes #3067~ Already done in #3072. Removed commits:
  * 23935ccb5531d7e14e15db500fd625daf9437471: expose config via API
  * 9f29fddce08a3083751822a059f27fcad2f9d6ce: test that API and CLI command return the same config

Issue #3074, opened 2023-08-31
fedimint-cli needs a way to display client config


justinmoon:
Since https://github.com/fedimint/fedimint/pull/2734, `fedimint-cli` saves client config to rocksdb and not `client.json` on filesystem like it used to. For debugging purposes, it can be helpful to be able to print out the config. We can no longer just look in the `--data-dir`. Would be nice to have a `fedimint-cli dev config` command or something like that. Would also be nice to jsonify the module section of the config (kinda like https://github.com/fedimint/fedimint/pull/3072 did for API).

justinmoon:
Perhaps `fedimintd` should have a command as well? https://github.com/fedimint/fedimint/issues/3080#issuecomment-1713864002

PR #3081, opened 2023-09-01
feat(cli): Add command for outputting config

cbergqvist:
This is an unofficial stab at issue #3074 based on @justinmoon suggesting I take a stab at it, I just saw PR #3079 for a more extensive solution, which is necessary to get nicer output formatting.

Without significantly changing the types declared outside of this single file, this is the least bad thing I could come up with - using `format!` + Display to convert module configs to `String`s. Using `Debug` gives cleaner output but I'm not sure that's okay in committed code. I cannot use `serde_json::to_value()` as `DynClientConfig` doesn't implement the `Serializable` trait and I didn't want the code in fedimint-cli to `match` on the module kind and explicitly cast to concrete types (like `LightningClientConfig` which is `Serializable`).

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -26,6 +26,7 @@ use fedimint_core::api::{
     IFederationApi, InviteCode, WsFederationApi,
 };
 use fedimint_core::config::{ClientConfig, FederationId};
+use fedimint_core::core::ModuleInstanceId;
 use fedimint_core::db::DatabaseValue;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::epoch::{SerdeEpochHistory, SignedEpochOutcome};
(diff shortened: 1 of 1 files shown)


elsirion:
Thx for the PR, but this might have been a bit much for a first issue, I think #3079 is the technically better solution. But if there wasn't #3079 already I'd see this as an ok enough short-term fix.

Just as feedback for how to structure PRs: small fixups like "make clippy happy" can just be added to the previous commit using `git commit --amend` followed by a force push `git push -f`. If there already are two commits these can be [squashed](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) (combined).

cbergqvist:
Thanks for the feedback @elsirion! I'm not super-used to GitHub review etiquette so I held off squashing the clippy commit and force-pushing, thinking reviewers could rely on the "Files changed" tab or equivalent. But I probably should have gone ahead as you say, especially before anyone has had time to look at it.

elsirion:
Yeah, you are right, for huge PRs that is much nicer. But we don't do it consistenly as it would add an additional round-trip at the end, which especially for small changes just isn't worth it.

If you are interested in the *correct* workflow for such fixes [look into `--fixup` and `--autosquash`](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html).

Issue #2857, opened 2023-07-28
Gateway: Boot even when a previously known client fails to (re)initialize


okjodom:
Gatewayd [re-initializes all previously known clients](https://github.com/fedimint/fedimint/blob/ad196fcbeb64cf26801186caa3838822d09dcfb0/gateway/ln-gateway/src/lib.rs#L415) on boot. If initializing any single client fails for reasons like failure to download config from that specific federation, the whole gateway boot fails. Ideally we should just catch the failure and emit a warning message or something that we failed to load the client for that federation (and at some point make that visible in the UI)

PR #3082, opened 2023-09-01
fix: boot when a client fails to initialize

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2857

Instead of failing the whole gateway init, now we catch the error and emit a warning. Gateway operator can check which federations are currently connected by doing `gateway-cli info`


elsirion (approved):
I'm ok with merging because a federation being connected is likely covered implicitly by existing tests, but generally making a hard failure a warning likely requires more testing that said warning didn't occur.

PR #3083, opened 2023-09-01
feat(module/wallet): add `block_count_local` api

### LINKED ISSUE #3069: RPC to get current bitcoin block height from wallet module
The admin UI would like to show the current block height that the wallet module's bitcoin node is at, so we'll need an endpoint for getting that. It might be nice to just get the entirety of `getblockchaininfo` in case anything else is needed (`chain`, `bestblockhash` etc.), but I also don't want to ask for more than we need.

dpc:
Fix #3069

```
> fedimint-cli dev api module_2_block_count
2023-09-01T18:58:46.559652Z  INFO fedimint_client::sm::executor: Starting state machine executor task
2023-09-01T18:58:46.559655Z  INFO fedimint_client: Last client reference dropped, shutting down client task group
2023-09-01T18:58:46.560027Z  INFO fedimint_client::sm::executor: Shutting down state machine executor runner due to shutdown signal
{
  "value": 284
}

> fedimint-cli dev api module_2_block_count_local
2023-09-01T19:00:01.080251Z  INFO fedimint_client: Last client reference dropped, shutting down client task group
2023-09-01T19:00:01.080371Z  INFO fedimint_client::sm::executor: Starting state machine executor task
2023-09-01T19:00:01.080948Z  INFO fedimint_client::sm::executor: Shutting down state machine executor runner due to shutdown signal
{
  "value": 294
}
```

### DIFF
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -321,7 +321,8 @@ impl ServerModule for Wallet {
             .collect::<Vec<WalletConsensusItem>>()
             .await;
 
-        let block_count = self.block_count().await;
+        // TODO: We should not be panicking
+        let block_count = self.get_block_count().await.expect("bitcoind rpc failed");
         let block_count_proposal = block_count.saturating_sub(self.cfg.consensus.finality_delay);
         let consensus_block_count = self.consensus_block_count(dbtx).await;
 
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -659,6 +667,8 @@ pub struct Wallet {
     cfg: WalletConfig,
     secp: Secp256k1<All>,
     btc_rpc: DynBitcoindRpc,
+    /// The result of last successful get_block_count
+    block_count_local: std::sync::Mutex<Option<u32>>,
```
elsirion, line 671: Why not [`tokio::sync::watch::Receiver`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html)?

dpc: Overkill for updating/reading an u32. I was going to go with `AtomicU32`, but then I need to use a magic number for none, and I'm superstitious.

dpc:
@wbobeirne We still want it, right?

elsirion (approved):

Issue #3068, opened 2023-08-30
Add node alias to `list_gateways` response


wbobeirne:
For the admin & gateway UIs, we'd like to show the node's alias when it's been configured, as a human-readable name for the gateway. It'd be nice if this was attached as an optional string to the `list_gateways` response.

PR #3085, opened 2023-09-01
fix: add lightning alias to LightningGateway

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3068

image


gateway/ln-gateway/src/lib.rs
```diff
@@ -677,7 +677,7 @@ impl Gateway {
         if let GatewayState::Running {
             lnrpc,
             lightning_public_key,
-            lightning_alias: _,
```
elsirion, line 680: I was quite confused where the alias was coming from without change to the LND and CLN modules :laughing: it was there all along!

elsirion (approved):

gateway/ln-gateway/src/lib.rs
m1sterc001guy: We had it for `gateway-cli info` but never put it on the `LightningGateway` registration

PR #3090, opened 2023-09-01
Refactor Audit RPC (module instance IDs and defaults)

### LINKED ISSUE #3071: Audit RPC endpoint should be instance-id aware
Our module system is built to allow multiple instances of the same module running concurrently. As a result, we can't just use "module types" as keys to lookup information regarding modules because there may be 2 different instances of these modules. We should modify `audit` RPC method introduced [here](https://github.com/fedimint/fedimint/pull/2999) to use module instance IDs as keys instead of module types.

### LINKED ISSUE #3076: Audit RPC should return default values for modules without transactions
The audit RPC currently returns summaries only for modules that have associated `AuditItem`s (see [`generate_module_summaries`](https://github.com/fedimint/fedimint/blob/0d0b5144616ceb3652fd34ce9a3a17aedbdad6a8/fedimint-core/src/module/audit.rs#L99)). This causes confusing behavior when consuming the `/audit` endpoint (relevant Discord [thread](https://discord.com/channels/990354215060795454/990354215878688860/1146913872981278854)).

We should return a summary for all modules, which includes defaults values for modules that don't yet have `AuditItem`s.

bradleystachurski:
Fixes #3076, fixes #3071 

- update module summaries from module name -> module summary to module instance id -> module summary
  - includes the module `kind` in the summary so consumers can identify the module without additional network round trips
  - this update is necessary since multiple instances of a single module kind can exist (i.e. module kind is not guaranteed to be a singleton)
- add a default placeholder for modules without `AuditItem`s

---
### Updated response payload

Starting devimint without peg-ins

```bash
bash-5.1$ fedimint-cli --our-id 0 --password pass admin audit
{
  "net_assets": 0,
  "module_summaries": {
    "0": {
      "net_assets": 0,
      "kind": "ln"
    },
    "1": {
      "net_assets": 0,
      "kind": "mint"
    },
    "2": {
      "net_assets": 0,
      "kind": "wallet"
    }
  }
}
```

### DIFF
--- a/fedimint-core/src/module/audit.rs
+++ b/fedimint-core/src/module/audit.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 
+use fedimint_core::core::ModuleInstanceId;
 use futures::StreamExt;
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
@@ -17,14 +18,14 @@ impl Audit {
         AuditItem {
             name: "Net assets (sats)".to_string(),
             milli_sat: calculate_net_assets(self.items.iter()),
-            module_name: "".to_string(),
+            module_instance_id: None,
         }
     }
 
     pub async fn add_items<KP, F>(
         &mut self,
         dbtx: &mut ModuleDatabaseTransaction<'_>,
-        module_name: &str,
+        module_instance_id: ModuleInstanceId,
         key_prefix: &KP,
         to_milli_sat: F,
     ) where
@@ -41,7 +42,7 @@ impl Audit {
                 AuditItem {
                     name,
                     milli_sat,
-                    module_name: module_name.to_string(),
+                    module_instance_id: Some(module_instance_id),
                 }
             })
             .collect::<Vec<AuditItem>>()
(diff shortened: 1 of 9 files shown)


wbobeirne (commented):
Since this is a breaking API change, shall we get into the habit of bumping API versions? Not sure what our convention around that is yet.

bradleystachurski:
> Since this is a breaking API change, shall we get into the habit of bumping API versions? Not sure what our convention around that is yet.

Good point! Since we're getting close to cutting a release, I think that's a great idea.

bradleystachurski:
☝️ empty diff - force push to kick off another CI run. The code coverage step timed out after 45m

elsirion:
> Since this is a breaking API change, shall we get into the habit of bumping API versions? Not sure what our convention around that is yet.

We need to discuss our policy. My intuition was master is unversioned and if we were to backport this change we'd bump the API version. Each consensus bump resets the API version anyway.

fedimint-core/src/module/audit.rs
```diff
@@ -63,50 +68,86 @@ impl Display for Audit {
...
+    fn module_kind_or(&self, default: String) -> String {
+        self.module_metadata
+            .as_ref()
+            .map(|meta| meta.module_kind.to_string())
+            .unwrap_or(default)
+    }
+}
+
+pub struct ModuleMetadata {
+    pub module_kind: String,
```
elsirion, line 88: Saving the kind here every time seems redundant. Don't we have the instance->kind mapping available when querying the audit items? Injecting the kind after aggregating the audit items would be most elegant imo.

fedimint-core/src/module/mod.rs
```diff
@@ -874,7 +874,12 @@ pub trait ServerModule: Debug + Sized {
     ///
     /// Summing over all modules, if liabilities > assets then an error has
     /// occurred in the database and consensus should halt.
-    async fn audit(&self, dbtx: &mut ModuleDatabaseTransaction<'_>, audit: &mut Audit);
+    async fn audit(
+        &self,
+        dbtx: &mut ModuleDatabaseTransaction<'_>,
+        audit: &mut Audit,
+        module_instance_id: ModuleInstanceId,
+    );
```
elsirion, line 882: We'll have to re-think this API, ideally modules aren't aware of their own instance id. We might want to remove the DB as a communication mechanism and let modules do a tally internally and directly return the audit items? Thoughts @dpc?

dpc: Agreed.

fedimint-core/src/module/audit.rs
elsirion: I think this one would be nice to fix @bradleystachurski before merging. The other comment needs more thought and is out of scope.

bradleystachurski: Great call, that's a much better approach 👍

fedimint-server/src/net/api.rs
```diff
@@ -342,12 +342,21 @@ impl ConsensusApi {
...
+                    &mut audit,
+                    module_instance_id,
+                )
                 .await
         }
-        Ok(AuditSummary::from_audit(&audit))
+        Ok(AuditSummary::from_audit(
+            &audit,
+            &module_instance_id_to_kind,
+        ))
```
elsirion, line 359: It's slightly annoying that we have to mutate two variables here, but I see how the pre-existing code makes this the most concise option to build the instance->kind map. I guess by inlining `from_audit` we could remove the need for it, but then this fn might be doing too much.

elsirion (approved):

elsirion:
Since it is an API break I wouldn't want to backport it to v0.1, is it good enough if it's available in v0.2 @wbobeirne?

justinmoon:
> Since it is an API break I wouldn't want to backport it to v0.1, is it good enough if it's available in v0.2 @wbobeirne?

Yea I think that's fine.

PR #3091, opened 2023-09-01
 chore(devimint): speed up fed initialization 

dpc:


vincenzopalazzo (approved):

devimint/src/external.rs
```diff
@@ -81,8 +81,10 @@ impl Bitcoind {
         let client = self.client();
         let addr = client.get_new_address(None, None)?;
         let initial_block_count = client.get_block_count()?;
-        client.generate_to_address(block_num, &addr)?;
-        while client.get_block_count()? < initial_block_count + block_num {
+        tokio::task::block_in_place(|| client.generate_to_address(block_num, &addr))?;
+        while tokio::task::block_in_place(|| client.get_block_count())?
```
elsirion, line 85: Should these wrappers be inside the functions since they are most likely always called from an async context?

dpc: The `client` here is `bitcoincore_rpc::Client` (external create). We could wrap it in a newtype and make an async API, but right now it seems not used often enough to bother.

elsirion: Ah, totally missed that

elsirion (approved):

PR #3093, opened 2023-09-01
chore: put a timeout on an api request for config

dpc:


dpc:
I'm not sure if this is a good idea. The call will be retried anyway, and I think I've seen it hanging at least once for some reasons for 30 seconds (tcp/internal jsonrpc timeout?), because ... something (panic, or servers was not ready), but maybe it was something else.

fedimint-core/src/api.rs
```diff
@@ -544,7 +544,9 @@ where
             },
             false,
             self.all_peers().total(),
-        );
+        )
+        // downloading a config shouldn't take too long
+        .with_request_timeout(Duration::from_secs(5));
```
elsirion, line 549: This limits each request, not the total time fetching config might take, right? Is this the intention?

elsirion (approved):

fedimint-core/src/api.rs
dpc: Yes. Each single request. Higher level timeouts are for the strategy itself to figure out.

PR #3094, opened 2023-09-01
feature(gateway): add ClientConfig to FederationInfo for /info api endpoint

### LINKED ISSUE #3089: Provide more federation info to gateway from `/info` endpoint
Currently the federation info we get from the gateway `/info` endpoint is the bare minimum:

https://github.com/fedimint/fedimint/blob/946944531c58600447a2e35e5583bcaeb7eab80b/gateway/ln-gateway/src/rpc/mod.rs#L58-L64

For the Gateway UI, it'd be nice if we got much more information about the federation, probably just `client.get_config()` but not sure if we'd want anything more.

wbobeirne:
Closes #3089.

Adds `{ config: ClientConfig; }` to the `federations: FederationInfo[]` array we get from `/info`. Unfortunately this includes the hex encoded module configs, so it might be nice to add a `modules_config_json` in the future ala #3072, but for now it's not necessary for anything we've got in the designs for the gateway UI.

Could be nice to backport this to 0.1.0 so that gateway config can incorporate this into the gateway admin for release too, but the merge is a little messy due to some refactoring that's been done here.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -29,6 +29,7 @@ use clap::{Parser, Subcommand};
 use client::StandardGatewayClientBuilder;
 use db::{FederationRegistrationKey, GatewayPublicKey};
 use fedimint_client::module::init::ClientModuleInitRegistry;
+use fedimint_client::Client;
 use fedimint_core::api::{FederationError, InviteCode};
 use fedimint_core::config::FederationId;
 use fedimint_core::core::{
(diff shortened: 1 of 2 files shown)


m1sterc001guy (approved):
Looks like the `lightning_reconnect_test` is failing, but I don't think it ever even got to killing the lightning processes. I'll re-run it

elsirion (approved):
Good enough for now if it helps the GW UI. v0.2 should get a better solution where the module configs are JSON. See #3079 for how it could be done nicely without the redundant `modules_config_json` hack.

PR #3106, opened 2023-09-05
fix: missing commit_tx call 

dpc:
At very least the wallet module does writes to it.


elsirion (commented):
ACK to the logging improvements, NACK to just adding a commit where we don't really expect writes. @joschisan do you have any idea where the write could be coming from? Apparently just undoing it wasn't a problem so far?

EDIT: @dpc maybe I'm blind, but I followed `dbtx` through the `Wallet::consensus_proposal` function and can't find the write.

dpc:
> EDIT: @dpc maybe I'm blind, but I followed `dbtx` through the `Wallet::consensus_proposal` function and can't find the write.


I can't find it anymore. I think I scrolled too far and seen it in `process_consensus_item`. But the uncommitted write warning was there. I've found the one from #3111 couple hours later.

Anyway #3111 already lands the logging change, so closing this one, and if it happens again, will look for it then. Maybe I was hallucinating or something. :shrug: :rofl:

PR #3111, opened 2023-09-05
 fix: warning about uncommited writes 

dpc:


fedimint-core/src/db/mod.rs
```diff
@@ -1099,6 +1100,11 @@ impl<'parent> DatabaseTransaction<'parent> {
         }
     }
 
+    /// Cancel the tx to avoid debugging warnings about uncommitted writes
+    pub fn cancel(mut self) {
+        self.commit_tracker.has_writes = false;
+    }
```
bradleystachurski, line 1106: Instead of needing to call `dbtx.cancel()` everywhere we'll use a read-only transaction, what are your thoughts on adding a new method (`new_read_only_tx`?) along with an additional field in `CommitTracker` that checks if the tx is read only when we drop?

elsirion: We don't need to call it everywhere, most transactions don't have writes and don't produce warnings on drop. It's only the ones with writes that we need to explicitly cancel. Read-only tx would be nice, but ideally we'd enforce that on the type level and that's a lot of work. So this is a good fix for now.

elsirion (approved):

PR #3115, opened 2023-09-06
Abort database transaction in `submit_transaction`

elsirion:
The database transaction is used as scratch space to evaluate the validity of the fedimint transaction and we *want* to drop it. Doing so implicitly leads to a warning though, which this PR fixes by making the abort explicit.


elsirion:
I should really stop doing things myself before finishing my GH inbox. Dup of #3111.

PR #3118, opened 2023-09-06
fix(client): reintroduce module commands

elsirion:


shaurya947 (approved):

devimint/src/main.rs
```diff
@@ -388,6 +388,19 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         .unwrap();
     assert_eq!(client_ng_reissue_amt, reissue_amount);
 
+    // Ensure that client can reissue via module commands
+    info!("Testing reissuing e-cash via module commands");
+    let reissue_notes = cmd!(fed, "spend", reissue_amount).out_json().await?["notes"]
+        .as_str()
+        .map(|s| s.to_owned())
+        .unwrap();
+    let client_ng_reissue_amt = cmd!(fed, "module", "--module", "mint", "reissue", reissue_notes)
```
dpc, line 397: If we are picking module by "kind", shouldn't it be `--kind mint` (vs `--instance 1`)?

dpc: Oh... it does auto-detection... I might have even implemented it... :rofl: . Nvmd me.

dpc (approved):

devimint/src/main.rs
elsirion: I was wondering the same but too lazy to change the behavior since the existing one was good enough.

dpc:
Seems backward compatible.

PR #3126, opened 2023-09-07
feat: pass extra dkg meta to fedimintd

dpc:
```
> env FM_EXTRA_DKG_META=foo=bar just mprocs
```

gives:

image

### DIFF
--- a/devimint/src/util.rs
+++ b/devimint/src/util.rs
@@ -1,7 +1,8 @@
+use std::collections::BTreeMap;
 use std::ffi::OsStr;
 use std::unreachable;
 
-use anyhow::bail;
+use anyhow::{bail, format_err};
 use fedimint_core::task::{self, block_in_place};
 use fedimint_core::time::now;
 use futures::executor::block_on;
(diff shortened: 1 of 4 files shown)


fedimintd/src/fedimintd.rs
```diff
@@ -32,6 +33,8 @@ use crate::attach_default_module_init_params;
 /// Time we will wait before forcefully shutting down tasks
 const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
 
+pub const FM_EXTRA_DKG_META_VAR: &str = "FM_EXTRA_DKG_META";
```
dpc, line 36: I'm not sure about the name.

devimint/src/util.rs
```diff
@@ -13,6 +14,24 @@ use tracing::{debug, warn};
...
+        let parts: Vec<&str> = pair.split('=').collect();
+        if parts.len() == 2 {
+            map.insert(parts[0].to_string(), parts[1].to_string());
+        } else {
+            return Err(format_err!("Invalid pair in map: {}", pair));
+        }
+    }
+    Ok(map)
+}
+
```
elsirion, line 34: Can't you import this duplicated function from a common dependency of `fedimintd` and `devimint`?

dpc: Me and me were debating it while implementing and reviewing it yesterday.

I don't know if we should bloat public interface of `fedimint-core` with every utility function that we happen to use in multiple places.

We probably ... want some `fedimint-internal` or `fedimint-utils-internal` that we could specifically keep low key, so external users don't use it. But even then we will have deal with it's semver, possibly by just incrementing major release like there's no tomorrow.

elsirion (approved):

fedimint-backports:
Successfully created backport PR for `releases/v0.1`:
- #3330

PR #3128, opened 2023-09-08
fix: add missing migration test for `BlockCountVote`

bradleystachurski:
We missed adding a migration test and updating the v0 migration file for `BlockCountVote`. Additional context in [Discord](https://discord.com/channels/990354215060795454/990354215878688860/1146823883543019551).


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1578,11 +1582,22 @@ mod fedimint_migration_tests {
                                 .await;
                             let num_shares = proposed_decryption_shares.len();
                             ensure!(
-                            num_shares > 0,
-                            "validate_migrations was not able to read any ProposeDecryptionShares"
-                        );
```
bradleystachurski, line 1583: The `ensure!` macro seems to give `cargo fmt` issues that go unnoticed. Has anyone bumped into other issues with `cargo fmt` silently failing?

elsirion: A lot of macros have that problem (e.g. `stream!` and `select!`).

elsirion (approved):
Good find, thx for the fix!

Issue #3129, opened 2023-09-08
Prevent GH workflows on forks that access secrets


bradleystachurski:
Several GH workflows require access to secrets, which causes a spam of errors reported on forks. We should check the repository is `fedimint/fedimint` for workflows that require access to secrets.

Additional context: https://github.com/fedimint/fedimint/pull/3114

PR #3130, opened 2023-09-08
chore: prevent gh workflows with secrets on forks

bradleystachurski:
Fixes https://github.com/fedimint/fedimint/issues/3129

Followup to https://github.com/fedimint/fedimint/pull/3114#pullrequestreview-1612985083

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -49,6 +49,7 @@ jobs:
         run: nix build -L .#ci.workspaceCargoUdeps
 
   build:
+    if: github.repository == 'fedimint/fedimint'
     strategy:
       matrix:
         host:
@@ -125,13 +126,14 @@ jobs:
       - name: Upload failed test build dirs
         if: always()
         uses: actions/upload-artifact@v3
-        with: 
+        with:
           name: "fedimint-failed-test-logs-${{ github.run_number }}-${{ matrix.host}}"
           path: |
             /tmp/nix-build-*/
             !/tmp/nix-build-*/source/
 
   audit:
+    if: github.repository == 'fedimint/fedimint'
     name: "Audit"
     runs-on: buildjet-2vcpu-ubuntu-2004
     timeout-minutes: 10
@@ -157,6 +159,7 @@ jobs:
   # Which means we can not reuse much from `build` job. Might as well run it as another
   # build in parallel
   ccov:
+    if: github.repository == 'fedimint/fedimint'
     name: "Code coverage"
     runs-on: buildjet-8vcpu-ubuntu-2004
     timeout-minutes: 45
@@ -246,6 +249,7 @@ jobs:
         run: nix build -L .#cross.${{ matrix.target }}.client-pkgs
 
   containers:
+    if: github.repository == 'fedimint/fedimint'
     name: "Containers"
     runs-on: buildjet-8vcpu-ubuntu-2004
     timeout-minutes: 30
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-nix.yml
```diff
@@ -20,6 +20,7 @@ on:
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   lint:
+    if: github.repository == 'fedimint/fedimint'
```
elsirion, line 23: I think this should work on forks if we only set the `authToken` if it's available and gives at least rudimentary feedback

elsirion (commented):
There are two reasons for why a build might not work:
* Secrets being unavailable
  * Often this should be fixable by running the cachix action read-only
* Runners being unavailable because we use BuildJet and the fork isn't
  * We could supply alternative config, but GH runners are so slow that it's probably best to just run CI locally (easily possible due to nix) or by just opening a draft PR.

.github/workflows/ci-nix.yml
dpc: What doesn't work about lints on forks? cachix uses `continue-on-error: true`

dpc (commented):
Seems like `cachix` should not be a blocker for forks. They can still potentially fetch stuff from public fedimint's cachix, and not being able to push is ignored.

.github/workflows/ci-nix.yml
bradleystachurski: Verified `continue-on-error: true` will work on forks 👍 

This is a successful run on my fork's master branch: https://github.com/bradleystachurski/fedimint/actions/runs/6128318965

.github/workflows/daily.yml
```diff
@@ -16,15 +16,14 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
   notifications:
-    # if: github.ref == 'refs/heads/master'
-    if: always()                              
+    if: always() && github.repository == 'fedimint/fedimint'
     name: "Notifications"
     timeout-minutes: 1
     runs-on: ubuntu-22.04
-    needs: [ check-unused-dependencies, audit ]
```
bradleystachurski, line 24: We removed the `check-unused-dependencies` job in https://github.com/fedimint/fedimint/commit/b7a6e368192114376c1818f774f13dcde9803442.

.github/workflows/docs.yml
```diff
@@ -21,7 +21,8 @@ jobs:
       - name: Build docs
         run: nix build -L .#debug.workspaceDocExport
 
-      - name: Deploy docs
+      - if: github.repository == 'fedimint/fedimint'
```
bradleystachurski, line 24: Since this job `runs-on: ubuntu-22.04` instead of buildjet, we can allow it to run until the deploy step on forks.

bradleystachurski:
> There are two reasons for why a build might not work:
> 
> * Secrets being unavailable
>   
>   * Often this should be fixable by running the cachix action read-only
> * Runners being unavailable because we use BuildJet and the fork isn't
>   
>   * We could supply alternative config, but GH runners are so slow that it's probably best to just run CI locally (easily possible due to nix) or by just opening a draft PR.

Thanks for the context!

The latest changes will prevent jobs on forks that
- Are configured to run with buildjet
  - Since we don't define an `idle_timeout`, the job will hang waiting for an available runner that will never be available on the fork (assuming no buildjet setup)
- Attempt to access secrets that don't fail gracefully (everything except cachix)

The goal for these changes is to prevent spam for failed CI runs on forks that have GH actions enabled. The draft PR and local `just test-ci-all` approaches make the alternate fork configs unnecessary.

bradleystachurski:
> Since we don't define an idle_timeout, the job will hang waiting for an available runner that will never be available on the fork (assuming no buildjet setup)

Scratch that, looks like it times out after 1 day 😃 
https://github.com/bradleystachurski/fedimint/actions/runs/6122825276

elsirion (approved):

PR #3134, opened 2023-09-09
refactor: add endpoints for single outpoint outcomes

joschisan:


fedimint-server/src/net/api.rs
```diff
@@ -276,6 +277,29 @@ impl ConsensusApi {
...
+        let mut dbtx = self.await_transaction_acceptance(outpoint.txid).await;
+
+        for (module_id, .., module) in self.modules.iter_modules() {
+            if let Some(outcome) = module
+                .output_status(&mut dbtx.with_module_prefix(module_id), outpoint, module_id)
+                .await
+            {
+                return Ok((&outcome).into());
+            }
+        }
```
elsirion, line 290: This scales horribly, we should just make the module instance id an argument too.

fedimint-server/src/net/api.rs
```diff
@@ -276,6 +277,29 @@ impl ConsensusApi {
         VerificationCaches { caches }
     }
 
+    pub async fn await_output_outcome(&self, outpoint: OutPoint) -> Result<SerdeOutputOutcome> {
+        let mut dbtx = self.await_transaction_acceptance(outpoint.txid).await;
+
+        for (module_id, .., module) in self.modules.iter_modules() {
+            if let Some(outcome) = module
+                .output_status(&mut dbtx.with_module_prefix(module_id), outpoint, module_id)
+                .await
```
elsirion, line 286: The transaction being accepted does not mean the outcome is ready/final, so this API endpoint would still result in polling.

To discuss: how to better model output outcomes so that there is a way to await the state one is interested in (currently there can be multiple states that change).

elsirion (approved):

Issue #3107, opened 2023-09-05
Intermittent `switch-gateway` failures


bradleystachurski:
After rebasing on master and running [bechmarks](https://github.com/bradleystachurski/fedimint/tree/audit-benchmark), I'm repeatedly observing the following error that shuts down devimint. This occurs ~10 minutes after previous successful calls to `switch-gateway`.

```
Error: command: fedimint-cli --data-dir=/tmp/nix-shell.ClwUdE/fm-d408/cfg switch-gateway 02b619f6bc1fa2ac04f817dc333e657627ed0518cf2f2cac5df54e1203734f13a2

Caused by:
    exit status: 1
    stdout:

    stderr:
    2023-09-05T20:49:39.433207Z  INFO fedimint_client::sm::executor: Starting state machine executor task
    2023-09-05T20:49:39.435544Z  INFO fedimint_client: Last client reference dropped, shutting down client task group
    2023-09-05T20:49:39.435616Z  INFO fedimint_client::sm::executor: Shutting down state machine executor runner due to shutdown signal
    {
      "error": "CliError",
      "kind": "GeneralFailure",
      "message": "failure",
      "raw_error": "Could not find any gateways"
    }
```

PR #3135, opened 2023-09-09
fix: set gateway state to running prior to register clients timer

bradleystachurski:
Fixes https://github.com/fedimint/fedimint/issues/3107

If `GatewayState` is not set to `Running` prior to calling `register_clients_timer`, no task will be scheduled to re-register the gateway with the federation prior to gateway TTL expiration.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -132,22 +132,23 @@ pub struct GatewayOpts {
     pub num_route_hints: Option<usize>,
 }
 
+#[cfg_attr(doc, aquamarine::aquamarine)]
 /// ```mermaid
 /// graph LR
 /// classDef virtual fill:#fff,stroke-dasharray: 5 5
 ///
-/// graph LR
-/// classDef virtual fill:#fff,stroke-dasharray: 5 5
-///    Initializing -- begin intercepting HTLCs --> Running
+///    Initializing -- begin intercepting HTLCs --> Connected
 ///    Initializing -- gateway needs config --> Configuring
-///    Configuring -- configuration set --> Running
+///    Configuring -- configuration set --> Connected
+///    Connected -- load federation clients --> Running
 ///    Running -- disconnected from lightning node --> Disconnected
-///    Disconnected -- re-established lightning connection --> Running
+///    Disconnected -- re-established lightning connection --> Connected
 /// ```
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub enum GatewayState {
     Initializing,
     Configuring,
+    Connected,
     Running {
         lnrpc: Arc<dyn ILnRpcClient>,
         lightning_public_key: PublicKey,
@@ -350,7 +351,7 @@ impl Gateway {
     }
 
     async fn start_gateway(mut self, task_group: &mut TaskGroup) -> Result<()> {
-        let mut tg = task_group.make_subgroup().await;
+        let tg = task_group.clone();
         task_group
             .spawn(
                 "Subscribe to intercepted HTLCs in stream",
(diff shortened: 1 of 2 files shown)


m1sterc001guy (changes requested):
Ah nice catch, this is an issue. There's a small race condition here though, where now we change the `GatewayState` to `Running` before we've actually loaded the federation clients.

Can we change it so `register_clients_timer` starts the background thread regardless of the `GatewayState`, but only registers the federations if we're in the `Running` state? I think that would fix the race condition and this bug as well.

bradleystachurski:
Thanks for the quick review!

> There's a small race condition here though, where now we change the `GatewayState` to `Running` before we've actually loaded the federation clients.

I'm having a hard time spotting the race condition. Can you elaborate on where you see it?

A couple other questions:
- Should we modify `load_clients()` to handle setting the gateway state to running only after the clients have been loaded in memory?
- If the HTLC stream breaks and we attempt to reconnect, is it intentional to have multiple background tasks from `register_clients_timer`?

m1sterc001guy:
Ah good questions. I think we have another bug here. Ideally, we only want to service some API requests (e.g pay_invoice) after the gateway has connected to the lightning node AND the federation clients have loaded. Right now, we only have 1 state for `Running` and since the federation clients require a lightning connection, we need to wait until after we're connected to the lightning node to load the federation clients. This introduces a race condition since it might take some time for the federation clients to load.

It should probably be something like this:
 - After connecting to the lightning node, set `state` to `GatewayState::Connected` (this is a new state)
 - After loading the federation clients, set `state` to `GatewayState::Running`
 - APIs that require both lightning and the federation clients loaded should verify that we're in the `Running` state.

> If the HTLC stream breaks and we attempt to reconnect, is it intentional to have multiple background tasks from register_clients_timer?

Nope, not intentional. This loop should break if it becomes disconnected IMO.

bradleystachurski:
Quick update for visibility: I should have time to focus on this Wednesday, 2023-09-13

gateway/ln-gateway/src/lib.rs
```diff
@@ -772,102 +788,106 @@ impl Gateway {
...
                                     }
                                 }
-                            }
-                            Err(e) => {
-                                error!(
-                                    "Could not retrieve route hints, gateway will not be registered: {e:?}"
-                                );
+                                Err(e) => {
+                                    error!(
+                                        "Could not retrieve route hints, gateway will not be registered: {e:?}"
```
m1sterc001guy, line 875: Probably good to log the gateway's state here

gateway/ln-gateway/src/lib.rs
```diff
@@ -772,102 +788,106 @@ impl Gateway {
...
-                    while !handle.is_shutting_down() {
+        let clients = self.clients.clone();
+        let api = self.api_addr.clone();
+        let gateway_id = self.gateway_id;
+        let num_route_hints = self.num_route_hints;
+        let state = self.state.clone();
+        task_group
+            .spawn("register clients", move |handle| async move {
+                let registration_loop = async {
+                    loop {
```
m1sterc001guy, line 850: How does this loop break? We should probably add a `is_shutting_down` check right?

gateway/ln-gateway/src/lnd.rs
```diff
@@ -127,12 +127,27 @@ impl GatewayLndClient {
...
-                        error!("Error received over HTLC stream: {:?}", e);
+                // To gracefully handle shutdown signals, we need to be able to receive signals
+                // while waiting for the next message from the HTLC stream.
+                //
+                // If we're in the middle of processing a message from the stream, we need to
+                // finish before stopping the spawned task. Checking if the task group is
+                // shutting down at the start of each iteration will cause shutdown signals to
+                // not process until another message arrives from the HTLC stream, which may
+                // take a long time, or never.
+                while let Some(htlc) = tokio::select! {
```
m1sterc001guy, line 138: Ah thank you, nice catch

gateway/ln-gateway/src/lib.rs
bradleystachurski: Good question! I added a comment above `tokio::select!`, but let me know if that isn't clear. I was looking for a way to receive a shutdown signal without needing to wait for the next iteration of the loop, which will include a long registration delay. Using `tokio::select` seemed like the cleanest approach, but let me know if you think there's a better way to short circuit the loop on a shutdown signal.

I didn't see any harm in allowing the task to cancel while in the middle of registering, but please let me know if you see an issue if only a subset of the clients finish registering with the federation when the task terminates.

bradleystachurski: If there is an issue with only a subset of the clients registering before the task terminates, we can tweak this to `tokio::select` on either the shutdown signal or the registration delay, while ensuring the logic to register clients with the federation finishes before the task terminates.

m1sterc001guy: No, I don't think that's an issue. I see the gateway registration as "best effort" anyway, since its possible that a gateway registers then immediately goes offline. 

`tokio::select` should be fine too, I'm mainly wondering if the `registration_loop` future in `tokio::select` is even necessary, since it doesn't look like that loop ever breaks.

bradleystachurski: If we don't use the registration loop in the select and instead add a check for `is_shutting_down` at the start of the loop, shutting down the task group gracefully can take up to 85% of the TTL (8.5 mins).

The existing behavior is an infinite loop to re-register clients, so this just modifies it to handle a shutdown signal gracefully without needing to finish sleeping and registering clients.

m1sterc001guy (approved):

Issue #3142, opened 2023-09-11
Remove "ng" terminology from client


justinmoon:
A couple months ago we re-wrote our client. We used the acronym "ng" to stand for "next generation" because @elsirion is a Star Trek nerd 😛. At this point we should remove this terminology from the code because it's not informative.

I just noticed that `fedimint-cli` has a `handle_ng_command` function. There are probably other instances.

a-moreira:
hey @justinmoon I'd like to tackle this, if possible. if it's just about removing all instances of `ng`, I will start right now 😃

a-moreira:
@justinmoon https://github.com/fedimint/fedimint/pull/3149
I picked the name "common.rs" for the file that holds the code previously in "mod.rs", but I can change that, or re-organize it another way.

PR #3149, opened 2023-09-11
chore: remove 'ng' references

a-moreira:
closes https://github.com/fedimint/fedimint/issues/3142

I picked the name "common.rs" for the file that holds the code previously in "mod.rs", but I can change that, or even use another approach.

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -890,6 +890,79 @@ async fn lightning_gw_reconnect_test(dev_fed: DevFed, process_mgr: &ProcessManag
     Ok(())
 }
 
+async fn gw_reboot_test(dev_fed: DevFed, process_mgr: &ProcessManager) -> Result<()> {
+    #[allow(unused_variables)]
+    let DevFed {
+        bitcoind,
+        cln,
+        lnd,
+        fed,
+        gw_cln,
+        gw_lnd,
+        electrs,
+        esplora,
+        faucet,
+    } = dev_fed;
+
+    // Query current gateway infos
+    let mut cln_cmd = cmd!(gw_cln, "info");
+    let mut lnd_cmd = cmd!(gw_lnd, "info");
+    let (cln_value, lnd_value) = tokio::try_join!(cln_cmd.out_json(), lnd_cmd.out_json())?;
+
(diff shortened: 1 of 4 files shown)


m1sterc001guy:
I think its probably worth it to still keep a module for the state machines for the gateway. Perhaps instead of deleting `ng` folder, you can rename it to `sm`?

gateway/ln-gateway/src/common.rs
```diff
@@ -373,12 +371,12 @@ impl ClientModuleInit for GatewayClientGen {
 
 #[derive(Debug, Clone)]
 pub struct GatewayClientContext {
-    lnrpc: Arc<dyn ILnRpcClient>,
-    redeem_key: bitcoin::KeyPair,
-    timelock_delta: u64,
-    secp: secp256k1_zkp::Secp256k1<secp256k1_zkp::All>,
+    pub lnrpc: Arc<dyn ILnRpcClient>,
```
m1sterc001guy, line 374: Why do these need to be made public if all we're doing is renaming?

devimint/src/main.rs
```diff
@@ -381,31 +381,31 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         .as_str()
         .map(|s| s.to_owned())
         .unwrap();
-    let client_ng_reissue_amt = cmd!(fed, "reissue", reissue_notes)
+    let client_reissue_amt = cmd!(fed, "reissue", reissue_notes)
```
m1sterc001guy, line 384: Did we delete the legacy client code in this test? I remember explicitly naming these with `ng` because it was really easy to use the same name as the legacy client, which resulted in weird issues (either errors or silent successes)

elsirion: The legacy CLI tests have been gone for a while. The only legacy client tests left are pure rust tests.

m1sterc001guy:
Why was this closed? We still need this refactor :)

a-moreira:
@m1sterc001guy Github closed it automatically when I reverted my commits and force pushed, sorry lol

already re-created it here: https://github.com/fedimint/fedimint/pull/3251 :-)

PR #3150, opened 2023-09-11
chore: migrated wallet test peg_outs_must_wait_for_available_utxos

douglaz:


modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -102,6 +102,7 @@ async fn await_consensus_to_catch_up(client: &Client, block_count: u64) -> anyho
 }
 
 #[tokio::test(flavor = "multi_thread")]
+//#[ignore]
```
dpc, line 105: Leftover or on purpose?

douglaz: on purpose, being able to switch this on is very handy for development purposes

dpc: I don't get why you move one test, and then disable bunch of others. Could you move and ... keep other ones as they were?

douglaz: Let's just say sometimes I'm this cat:
https://twitter.com/buitengebieden/status/1698416516457181500

elsirion: I still don't understand the ignores, seems like there has to be a better way.

elsirion (dismissed):

modules/fedimint-wallet-tests/tests/tests.rs
dpc: :confused: Can anyone explain for the slower members of the community such as myself why do we disable these innocent looking tests? :D

douglaz: If I disable some tests the whole suite will run faster. But then why not just run only the test I want? Because if the test run alone, I can't detect some types of flakiness. And this isn't hypothetical, right now these tests are flaky only if some of them run together.

So I disable most of tests and run a subset to be a fast but yet realistic scenario.

douglaz:
When code runs with coverage on it starts creating infinite epochs. :confused:

elsirion:
Not sure that's your test's fault @douglaz, I've seen that kind of flakiness before :(

elsirion (dismissed):

elsirion:
Can you rebase? Apparently clippy on master is a bit more nitpicky and fails in the merge queue.

justinmoon (approved):

PR #3151, opened 2023-09-11
 fix: DBTransaction::cancel doesn't handle errors well

dpc:
Instead of trying to cancel dbtx at the end (which can be skipped
by an early return like `?`), mark uncommited writes as expected
upfront.


elsirion (approved):
Good catch!

PR #3152, opened 2023-09-11
chore: update fenix flake input (rust toolchain)

dpc:
We've detected some ios crashes when built on 1.71 debug mode, while 1.72 works just fine.


dpc:
@justinmoon I guess doesn't hurt to do it early?

elsirion:
Clippy is unhappy.

PR #3154, opened 2023-09-12
feat: add gateway config api

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/2786

This PR adds a `Configuring` step before the gateway becomes running. This will allow the UI (or other clients) to set the configuration via `/set_configuration` before the gateway is running. Once the configuration is set, the gateway continues the rest of the loading process. 

Right now, we only support setting the password via this API, but we will move other gateway specific configurations into this API.

To allow this to be backwards compatible, we still allow the password to be set from the command line or environment variable.

TODO
 - Move all gateway specific configurations to use this API https://github.com/fedimint/fedimint/issues/3167
 - Expose Gateway state via API https://github.com/fedimint/fedimint/issues/3158
 - Cleanup gateway tests https://github.com/fedimint/fedimint/issues/3155


m1sterc001guy:
@wbobeirne take a look at this if you get some time, this will be the main API used for configuring the gateway

wbobeirne (commented):
If I'm reading this correctly, I have no way of knowing if gatewayd is in the `Configuring` state, so I wouldn't know when to present the user with the set password screen vs the enter password screen.

m1sterc001guy:
> If I'm reading this correctly, I have no way of knowing if gatewayd is in the Configuring state, so I wouldn't know when to present the user with the set password screen vs the enter password screen.

Will add this in a follow up

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -144,3 +180,27 @@ async fn restore(
...
+    Extension(sender): Extension<mpsc::Sender<GatewayConfiguration>>,
+    Json(payload): Json<SetConfigurationPayload>,
+) -> Result<impl IntoResponse, GatewayError> {
+    let gateway_configuration = GatewayConfiguration {
+        password: payload.password.clone(),
+    };
+
+    sender.send(gateway_configuration).await.map_err(|e| {
+        GatewayError::UnexpectedState(format!("Gateway is not in Configuring state: {e:?}"))
+    })?;
```
wbobeirne, line 195: Have we considered if there's any configuration we would want to change after initial setup? If you want to change something while your gateway is already running, how would we do that?

m1sterc001guy: This should be supported. There's a different `set_configuration` endpoint on the webserver that runs after the config is set as well, which writes the config to the database.

The tricky part is applying the change after it is set in the database. Other operations (like fees) should be able to just query the db to pickup the new change. For the gateway's password though, we need to restart the webserver, which seemed like too much to implement for this PR.

gateway/ln-gateway/src/lib.rs
```diff
@@ -638,6 +651,76 @@ impl Gateway {
...
+        }
+
+        self.set_gateway_state(GatewayState::Configuring).await;
+
+        let mut task_group = TaskGroup::new();
+        let mut rx = run_config_webserver(self.listen, &mut task_group).await;
+
+        // Wait for the `/set_configuration` API to send a `GatewayConfiguration`
+        info!("Waiting for GatewayConfiguration to be set");
+        match rx.recv().await {
```
elsirion, line 704: Do we actually need that receiver or could we [wait for the DB key](https://docs.fedimint.org/fedimint_core/db/struct.Database.html#method.wait_key_exists) and do the insert in the API handler?

elsirion (approved):
Feels a little bit convoluted, maybe we can get rid of some of the channel from the API. But if you think it's a step in the right direction go ahead.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Forgot about that, I'll use that instead

m1sterc001guy:
> Feels a little bit convoluted, maybe we can get rid of some of the channel from the API. But if you think it's a step in the right direction go ahead.

I figured out a better way to do this. Channel is removed and we rely on `wait_key_exists` to know when to progress. 

Also reworked this to refresh the webserver when the password is changed. So no need to restart `gatewayd` anymore

fedimint-testing/src/gateway.rs
```diff
@@ -113,44 +116,50 @@ impl GatewayTest {
...
+    pub async fn wait_for_gateway_state(gateway: Gateway, func: impl Fn(GatewayState) -> bool) {
+        let mut gateway_state_iterations = 0;
+        loop {
+            let gw_state = gateway.state.read().await.clone();
+            if func(gw_state) {
+                break;
+            }
+
+            if gateway_state_iterations >= 30 {
+                panic!("Gateway did not start running after 10 attempts");
```
elsirion, line 156: Is it always the "started" state that is awaited? (just because the fn is now parametrized with a predicate)

gateway/ln-gateway/src/lib.rs
```diff
@@ -275,20 +280,75 @@ impl Gateway {
...
+                                if let Some(old_config) = gateway_config.clone() {
+                                    old_config.password != cfg.clone().password
+                                } else {
+                                    true
+                                }
+                            })
+                        })
+                        .await;
+
+                    info!("GatewayConfiguration has been updated, restarting webserver...");
```
elsirion, line 332: That you have to do that is kidna weird, but ok, let's iterate.

elsirion (dismissed):
Looks like a good start, just needs a `rustfmt` fix.

fedimint-testing/src/gateway.rs
m1sterc001guy: This is used in two spots. One spot waits for `Configuring` or `Running` (to test the case where the password has not been set at all) and the other case only waits for `Running` since it has already set the password.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Yeah, couldn't figure out a way to overwrite the bearer token without restarting the webserver. If we find a better way let's definitely change it!

elsirion (approved):

justinmoon:
@m1sterc001guy Will we be able to cherry-pick this to release branch? I guess state machines change here and that's backwards incompatible?

m1sterc001guy:
> Will we be able to cherry-pick this to release branch? I guess state machines change here and that's backwards incompatible?

I think it would be difficult, we would also need to backport https://github.com/fedimint/fedimint/pull/3048, which unfortunately contains a small bug with gateway registration. That bug will be fixed with https://github.com/fedimint/fedimint/pull/3135

PR #3174, opened 2023-09-13
chore: check in updated Cargo.lock

elsirion:
Somehow master has an out-of-date `Cargo.lock`, I thought we had a test against that @dpc? Or is something weird with my local env? I already updated the nix shell I'm using.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1088,7 +1088,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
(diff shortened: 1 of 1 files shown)


elsirion:
So close and debug locally?

dpc:
We do `--locked` couple of times in the CI. I also tried `cargo b` and ` Cargo.lock` seems untouched.

Try `direnv reload` maybe? I've seen some cases when I needed to run it manually. @elsirion

Issue #3158, opened 2023-09-12
Expose Gateway State


m1sterc001guy:
The gateway can now be in a few different "states" depending on what is happening with the gateway. For example, it will be useful for the UI to know if the lightning connection is down or if the gateway still needs to be configured.

okjodom:
Would this API also show state for the underlying FM client(s)?

m1sterc001guy:
No, this only exposes the gateway's state. This is mainly to communicate whether or not we're connected to the lightning node, gateway needs configuring, etc.

PR #3179, opened 2023-09-13
feat: expose gateway state

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3158

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -516,7 +516,7 @@ impl Gateway {
                 fees: self.fees,
                 route_hints,
                 gateway_id: self.gateway_id,
-                gateway_state: self.state.read().await.clone().to_string(),
+                gateway_state: self.state.read().await.to_string(),
             });
         }
 
@@ -528,7 +528,7 @@ impl Gateway {
             fees: self.fees,
             route_hints: vec![],
             gateway_id: self.gateway_id,
-            gateway_state: self.state.read().await.clone().to_string(),
+            gateway_state: self.state.read().await.to_string(),
         })
     }
 
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -158,6 +158,17 @@ pub enum GatewayState {
     Disconnected,
 }
 
+impl From<GatewayState> for String {
+    fn from(gw_state: GatewayState) -> Self {
+        match gw_state {
+            GatewayState::Initializing => "Initializing".to_string(),
+            GatewayState::Configuring => "Configuring".to_string(),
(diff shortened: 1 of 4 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -156,6 +156,17 @@ pub enum GatewayState {
     Disconnected,
 }
 
+impl From<GatewayState> for String {
+    fn from(gw_state: GatewayState) -> Self {
+        match gw_state {
+            GatewayState::Initializing => "Initializing".to_string(),
+            GatewayState::Configuring => "Configuring".to_string(),
+            GatewayState::Running { .. } => "Running".to_string(),
```
okjodom, line 164: nit: Maybe `GatewayState::LnConnected` and `GatewayState::LnDisconnected` instead of `GatewayState::Running` and `GatewayState::Disconnected`

m1sterc001guy: See https://github.com/fedimint/fedimint/pull/3135

I think the state that @bradleystachurski is adding is more appropriate to be called `LightningConnected` rather than this state

okjodom: makes sense

gateway/ln-gateway/src/lib.rs
```diff
@@ -156,6 +156,17 @@ pub enum GatewayState {
     Disconnected,
 }
 
+impl From<GatewayState> for String {
```
douglaz, line 159: There is a already a trait for that: `impl fmt::Display for GatewayState` which will also implement the `ToString` trait giving you a `.to_string()`
See https://doc.rust-lang.org/std/fmt/trait.Display.html

m1sterc001guy: Fixed

gateway/ln-gateway/src/lib.rs
```diff
@@ -503,6 +516,7 @@ impl Gateway {
                 fees: self.fees,
                 route_hints,
                 gateway_id: self.gateway_id,
+                gateway_state: self.state.read().await.clone().to_string(),
```
douglaz, line 519:
```suggestion
                gateway_state: self.state.read().await.to_string(),
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -514,6 +528,7 @@ impl Gateway {
             fees: self.fees,
             route_hints: vec![],
             gateway_id: self.gateway_id,
+            gateway_state: self.state.read().await.clone().to_string(),
```
douglaz, line 531:
```suggestion
            gateway_state: self.state.read().await.to_string(),
```

douglaz (approved):

Issue #3171, opened 2023-09-12
Use updated CLN RPC to get channels


m1sterc001guy:
https://github.com/fedimint/fedimint/pull/3132/files#diff-11314b558aa2437b3f9376371d29c660caf580b98f1e52fe9f7f548d7cc851d0R281

PR #3181, opened 2023-09-13
refactor: use new ListPeersChannel RPC for CLN extension

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3171

This will probably require CLN users to update their CLN version

### DIFF
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -259,49 +259,28 @@ impl GatewayLightning for ClnRpcService {
             .await
             .map_err(|err| tonic::Status::internal(err.to_string()))?;
 
-        let peers_response = client
-            .call(cln_rpc::Request::ListPeers(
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -259,49 +259,28 @@ impl GatewayLightning for ClnRpcService {
...
-                    .into_iter()
-                    .flatten()
-                    .map(move |chan| (peer.id, chan))
-            })
-            .filter_map(|(peer_id, chan)| {
-                // TODO: upstream eq derive
-                if !matches!(
-                    chan.state,
-                    model::responses::ListpeersPeersChannelsState::CHANNELD_NORMAL
-                ) {
```
elsirion, line 293: We probably still want to filter out inactive channels.

elsirion (changes requested):

elsirion (approved):

PR #3183, opened 2023-09-13
feat: set task names

douglaz:
This helps debugging by giving tasks some meaningful name, which will show by tokio on trace debugging level.

It doesn't fully acomplish my original goal that was setting a name for threads, but it's a improvement nonetheless.

### DIFF
--- a/fedimint-atomic-broadcast/src/broadcast.rs
+++ b/fedimint-atomic-broadcast/src/broadcast.rs
@@ -1,4 +1,5 @@
 use fedimint_core::db::Database;
+use fedimint_core::task::spawn;
 use fedimint_core::PeerId;
 use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinHandle;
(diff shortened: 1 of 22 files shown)


fedimint-atomic-broadcast/src/broadcast.rs
```diff
@@ -92,7 +94,7 @@ impl AtomicBroadcast {
         if let Some(signed_block) = db::load_block(&self.db, index).await {
             tracing::info!("Loaded block with index {}", index);
 
-            tokio::spawn(async move {
+            spawn("atomic run session", async move {
```
elsirion, line 97: Might it be useful to distinguish the two scenarios? Maybe call this one `atomic run session db`

fedimint-atomic-broadcast/src/session.rs
```diff
@@ -72,24 +72,28 @@ pub async fn run(
...
+            Spawner::new(),
+            aleph_bft::Terminator::create_root(terminator_receiver, "Terminator"),
         ),
-        Network::new(network_data_receiver, outgoing_message_sender.clone()),
-        keychain.clone(),
-        Spawner::new(),
-        aleph_bft::Terminator::create_root(terminator_receiver, "Terminator"),
-    ));
+    )
+    .expect("some handle");
```
elsirion, line 91: The reason why there will always be a handle is that this only runs on non-WASM platforms. Maybe we want to mention that in the expect?

elsirion (dismissed):
Looks good, some nits

dpc (approved):
Why not.

Issue #3137, opened 2023-09-11
backport #3130


justinmoon:
https://github.com/fedimint/fedimint/pull/3130

justinmoon:
Call: if backport bot fails, open issue automatically

bradleystachurski:
Closed by the two referenced PRs

PR #3185, opened 2023-09-14
chore: open issue when backport bot fails

bradleystachurski:
Addresses what was [discussed](https://github.com/fedimint/fedimint/issues/3137#issuecomment-1714229910) on the dev call to open an issue automatically when the backport bot fails.

I'm not sure how to test this prior to merging. Perhaps a sane strategy is to carefully review then quickly revert if we notice unwanted behavior?

Useful for reviewing
- `create-an-issue` GH [action](https://github.com/marketplace/actions/create-an-issue)
- backport action [outputs](https://github.com/korthout/backport-action#outputs) (specifically `was_successful`)
- Consuming outputs from previous jobs [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs)
  - outputs are unicode strings
- Referencing PR number
  - GH [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow)
  - [SO](https://stackoverflow.com/a/62436027)

### DIFF
--- a/.github/templates/failed-backport-issue.md
+++ b/.github/templates/failed-backport-issue.md
@@ -1,0 +1,4 @@
+---
+title: Backport #{{ env.PR_NUMBER }}
+---
+The backport bot was unable to open a pull request to backport changes from PR #{{ env.PR_NUMBER }}. Please open a PR to backport these changes.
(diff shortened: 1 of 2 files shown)


dpc (approved):

elsirion (approved):

elsirion:
Agreed, we'll see if it doesn't work as expected.

Issue #3155, opened 2023-09-12
Cleanup Gateway tests


m1sterc001guy:
We have a bunch of different test suite files in the gateway and the distinction between each file has become less and less clear. I think we should just consolidate all tests into a single integration tests file.

PR #3188, opened 2023-09-14
refactor: single integration test file for the gateway

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3155

Moves all gateway integration tests into a single test suite


elsirion (approved):

m1sterc001guy:
Flaky test failed, going to fix that next but it will be easier after this is merged since the fix will conflict

okjodom (approved):

elsirion:
I did investigate the failure and on the last run it seemed like only `on_chain_peg_in_and_peg_out_happy_case` got really stuck. I downloaded the log as `failure.log` and then ran:

```bash
sed -n 's/.*test \([^ ]*\).* \.\.\. ok/\1/p' failure.log | sort > success.log
sed -n 's/.*test \(.*\) has been.*/\1/p' failure.log | sort > too_long.log
comm -23 too_long.log success.log # checks which entries are only in too_long but not success
```

I couldn't reproduce locally so far.

PR #3191, opened 2023-09-14
fix: mprocs doesn't work when nushell is a login shell

dpc:
As a new Nushell user, I'll now be fixing all sorts of software just to have stuff that used to work, work again. Everybody makes bad life choices sometimes, right? Right? :shrug:

Before approving, please try `just mprocs` in your shell. BTW. Add to your `~/.gitconfig` this convenient alias:

```
[alias]
pr = "!f() { git fetch -fu ${2:-$(git remote |grep '^upstream$' || echo origin)} refs/pull/$1/head:pr/$1 && git checkout pr/$1; }; f"
```

So you can `git pr 3191` to do it.

### DIFF
--- a/misc/mprocs.yaml
+++ b/misc/mprocs.yaml
@@ -1,7 +1,7 @@
 server: 127.0.0.1:4050
 procs:
   user:
-    shell: bash --init-file scripts/dev/mprocs/user-shell.sh
+    shell: bash -c scripts/dev/mprocs/user-shell.sh
     stop: SIGKILL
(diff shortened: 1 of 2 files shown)


dpc:
I reverted from nushell. 

My understanding here is as follows:

There's no way to start an interactive bash shell (that could accept user commands) without triggering all the system scripts, which destroy environment sometimes/somehow. For some reason when nushell was my logic shell, it would break what we're doing currently.

Issue #3025, opened 2023-08-24
Future proof arguments to `ClientModuleInit::init`


dpc:
I noticed we keep tweaking these, adding, removing refactoring arguments there, forcing changes in downstream implementations.

Maybe we should pass one:

```rust 
struct ModuleInitArgs {
  // all fields private
}
```

than then has:

```rust
impl ModuleInitArgs {
  pub db(&self) -> &Database {
    // ...
  }

  pub fn api_version(&self) -> ApiVersion  {
    // ...
  }

  // other methods
}
```

then we can add new fields, deprecate old ones etc. without breaking the Api.

We probably should do it in some other module API traits methods as well, but need to be careful not to go overboard with it. Things that have multiple arguments that historically changed.

justinmoon:
Was this closed by https://github.com/fedimint/fedimint/pull/3254 and https://github.com/fedimint/fedimint/pull/3255?

dpc:
Yes

PR #3205, opened 2023-09-15
fix: wallet module spaming epochs

### LINKED ISSUE #3202: Wallet module spamming epochs on zero height or something
The reason for https://github.com/fedimint/fedimint/issues/3102 is that wallet modules is spamming with epochs (generating them endlessly) for some reason. If I can get   https://github.com/fedimint/fedimint/pull/3201 to fail i'll be able to debug. :D

dpc:
Fix #3202

### DIFF
--- a/modules/fedimint-wallet-common/src/config.rs
+++ b/modules/fedimint-wallet-common/src/config.rs
@@ -61,6 +61,8 @@ pub struct WalletConfig {
 pub struct WalletConfigLocal {
     /// Configures which bitcoin RPC to use
     pub bitcoin_rpc: BitcoinRpcConfig,
+    /// Our own peer_id
+    pub our_peer_id: PeerId,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
@@ -140,6 +142,7 @@ impl WalletConfig {
     pub fn new(
         pubkeys: BTreeMap<PeerId, CompressedPublicKey>,
         sk: SecretKey,
+        our_peer_id: PeerId,
         threshold: usize,
         network: Network,
         finality_delay: u32,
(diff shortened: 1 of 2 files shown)


modules/fedimint-wallet-common/src/config.rs
```diff
@@ -61,6 +61,8 @@ pub struct WalletConfig {
 pub struct WalletConfigLocal {
     /// Configures which bitcoin RPC to use
     pub bitcoin_rpc: BitcoinRpcConfig,
+    /// Our own peer_id
+    pub our_peer_id: PeerId,
```
dpc, line 65: I don't know if we should put it here, or is there's a better way.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -325,16 +327,19 @@ impl ServerModule for Wallet {
         // TODO: We should not be panicking
         let block_count = self.get_block_count().await.expect("bitcoind rpc failed");
         let block_count_proposal = block_count.saturating_sub(self.cfg.consensus.finality_delay);
-        let consensus_block_count = self.consensus_block_count(dbtx).await;
```
dpc, line 328: Observation: we don't care what's the consensus here. Our job is to submit our vote, and our previous vote is all that concerns us.

douglaz (approved):
I don't know if there is a better way, but LGTM

modules/fedimint-wallet-common/src/config.rs
elsirion: Maybe this should be a parameter of `ServerModuleInit::init` instead?

https://github.com/fedimint/fedimint/blob/6c15b0ea133b5239085884c14c334623287d555b/fedimint-core/src/module/mod.rs#L583-L588

dpc: Maybe. Would be easier if Re https://github.com/fedimint/fedimint/issues/3025 was ready. :D

dpc: Why not both? :shrug: :D

PR #3206, opened 2023-09-15
 chore: run backend tests with cargo nextest 

dpc:


fedimint-testing/src/btc/real.rs
```diff
@@ -175,14 +179,26 @@ impl RealBitcoinTest {
...
         trace!("Trying to acquire global bitcoin lock");
-        let _guard = REAL_BITCOIN_LOCK.lock().await;
+        let _guard = tokio::task::block_in_place(|| {
+            let lock_file_path = std::env::temp_dir().join("fm-test-bitcoind-lock");
+            fs_lock::FileLock::new_exclusive(
+                std::fs::OpenOptions::new()
+                    .write(true)
+                    .create(true)
+                    .truncate(true)
+                    .open(&lock_file_path)
```
Maan2003, line 196: Use std::fs::create, it does the same under the hood

dpc: We have a semgrep lint for it, to force people to think about truncation behavior, as it was used incorrectly in the past. I don't mind longer version.

Maan2003 (approved):

PR #3210, opened 2023-09-16
chore: try patched version of crane to fix weird crate rebuilds

dpc:


dpc:
Nope. `workspaceBuild` still rebuilds needlessly some crates.

PR #3212, opened 2023-09-16
fix: needless rebuilds

dpc:
Re: https://github.com/ipetkov/crane/issues/370
Re: https://github.com/rust-lang/cargo/issues/12457

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,13 @@ license-file = "LICENSE"
 keywords = ["bitcoin", "lightning", "chaumian", "e-cash", "federated"]
 
 
+# Workaround: https://github.com/rust-lang/cargo/issues/12457 which causes
+#             https://github.com/ipetkov/crane/issues/370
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -61,6 +61,13 @@ license-file = "LICENSE"
 keywords = ["bitcoin", "lightning", "chaumian", "e-cash", "federated"]
 
 
+# Workaround: https://github.com/rust-lang/cargo/issues/12457 which causes
```
tvolk131, line 64: I assume this issue will be addressed by the Rust team at some point, would it be worth adding a TODO comment to remove this eventually?

tvolk131 (approved):

Cargo.toml
dpc: If we forget to remove it, we just won't get debug symbols in build script binaries, which is never going to be a problem.

Maan2003 (approved):

PR #3215, opened 2023-09-16
refactor: TieredMulti::map() doesn't self-consume

tvolk131:
This allows us to sign a collection of blind nonces without consuming them

### DIFF
--- a/fedimint-core/src/tiered_multi.rs
+++ b/fedimint-core/src/tiered_multi.rs
@@ -62,26 +62,6 @@ impl<T> TieredMulti<T> {
         self.count_items() == 0
     }
 
-    /// Applies the given closure to every `(Amount, T)` pair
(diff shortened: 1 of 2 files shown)


fedimint-core/src/tiered_multi.rs
```diff
@@ -62,20 +62,20 @@ impl<T> TieredMulti<T> {
         self.count_items() == 0
     }
 
-    /// Applies the given closure to every `(Amount, T)` pair
-    pub fn map<F, N, E>(self, f: F) -> Result<TieredMulti<N>, E>
```
dpc, line 66: Thought: Why do we even need it, if we have `iter()` and `IntoIterator`?

tvolk131: Good point! Done, we're now calling `TieredMulti::iter_items()` and I removed `TieredMulti::map()`.

dpc (approved):
Grug like less code.

PR #3217, opened 2023-09-16
fix: cargo nextest triggering macos SIP

dpc:
Re:  https://github.com/nextest-rs/nextest/issues/962


dpc:
Worked here: https://github.com/fedimint/fedimint/actions/runs/6210193331/job/16858139303

Maan2003 (approved):
testing using `./scripts/backend-tests.sh`

PR #3218, opened 2023-09-17
fix: run_dkg in background task instead of blocking

### LINKED ISSUE #2815: Config API: Split `ReadyForConfigGen` to 2 states
The `run_dkg` config API RPC method transitions state to `ReadyForConfigGen` and blocks until it succeeds or fails. Let's split these state into 2: `ReadyForConfigGen` and `RunningConfigGen`. 

`run_dkg` will transition to `RunningForConfigGen`, then spawn a task which actually runs DKG and transitions state according to success / failure as it does currently. The big change is doing this in a background task instead of blocking. Then the UI can just poll server status instead of waiting for a blocking command which might timeout or get disconnected.

`RunningConfigGen` state will hold a handle to that task which runs DKG. This gets dropped when we move to another state causing the DKG task to terminate.

bradleystachurski:
Closes #2815 

These changes update the `run_dkg` API to set the server status to `ReadyForConfigGen`, then return after spawning a background task to handle the rest of coordinating DKG.

Previously, calls to `run_dkg` would block until DKG completed amongst all peers. This is troublesome for clients since there was no way to gracefully recover if a client experienced connectivity issues or elapsed a timeout.

### UI

These changes are not guaranteed to be backwards compatible for all clients, however testing against the UI repo (https://github.com/fedimint/ui/tree/c0cf511902f9e33cd66bb6438e4a56c95a395991), I verified these changes fix the issue without required changes.

#### Before

_Introduces a 100s sleep in `distributed_gen`, simulating poor network behavior which causes clients to timeout_

Screenshot 2023-09-16 at 6 06 56 PM

#### After

_Same 100s sleep as before, along with closing tabs for all followers and reconnecting_

https://github.com/fedimint/fedimint/assets/11034691/cb8417b7-a1e3-4526-b92d-2a75ef0818bc

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 use std::net::SocketAddr;
 use std::path::PathBuf;
-use std::sync::{Mutex, MutexGuard};
+use std::sync::{Arc, Mutex, MutexGuard};
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -37,11 +37,12 @@ use crate::net::peers::DelayCalculator;
 use crate::{check_auth, ApiResult, HasApiContext};
 
 /// Serves the config gen API endpoints
+#[derive(Clone)]
 pub struct ConfigGenApi {
     /// Directory the configs will be created in
     data_dir: PathBuf,
     /// In-memory state machine
-    state: Mutex<ConfigGenState>,
+    state: Arc<Mutex<ConfigGenState>>,
     /// DB not really used
     db: Database,
     /// Tracks when the config is generated
@@ -60,7 +61,7 @@ impl ConfigGenApi {
     ) -> Self {
         Self {
             data_dir,
-            state: Mutex::new(ConfigGenState::new(settings)),
+            state: Arc::new(Mutex::new(ConfigGenState::new(settings))),
             db,
             config_generated_tx,
             task_group: task_group.clone(),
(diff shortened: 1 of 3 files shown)


wbobeirne (commented):
This worked great for me with the UI. The lack of additional status has the quirk of returning the user to the approve screen since there's no way to tell that the user has already approved, but otherwise everything seems to behave OK even if guardians come and go in the middle of setup. I think that's worth the quirk to maintain backwards compatibility and simplify the change, but I'll let someone weigh in with a proper review of the rust code.

dpc (approved):

fedimint-server/src/config/api.rs
```diff
@@ -180,8 +178,13 @@ impl ConfigGenApi {
...
 
-    /// Once configs are generated, starts DKG and await its
-    /// completion, calling a second time will return an error
+    /// Once configs are generated, updates status to ReadyForConfigGen and
+    /// spawns a task to coordinate DKG, then returns. Coordinating DKG in a
+    /// separate thread allows clients to poll the server status instead of
+    /// blocking until completion, which can be fragile due to timeouts, poor
+    /// network connections, etc.
+    ///
+    /// Calling a second time will return an error.
```
okjodom, line 187: checking for `SharingConfigGenParams` or `ReadyForConfigGen` status would allow success returns instead of error when api is called multiple times

okjodom (approved):

fedimint-backports:
Successfully created backport PR for `releases/v0.1`:
- #3236

PR #3219, opened 2023-09-17
chore: use a new incremental zstd crane mode

dpc:
This should greatly decrease the sizes of workspace artifacts that crane produces.

See https://github.com/ipetkov/crane/pull/387


dpc:
Even though it's still a PR to crane, the author said it will land, and we can safely use it, as it makes the CI much more lightweight.

elsirion (approved):

PR #3225, opened 2023-09-18
chore(ci): skip some steps on MacOS

dpc:
The MacOS build is already struggling. Disabling these two steps should save around 15 minutes of build time for it, and they are both unlikely to fail, and even if they somehow did - it shouldn't matter.

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -106,11 +106,11 @@ jobs:
         run: nix build -L .#ci.workspaceClippy
 
       - name: Run cargo doc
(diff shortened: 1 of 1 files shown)


Maan2003:
we can also skip wasm tests. linux or mac doesn't matter for it.

Maan2003:
we can also skip clippy, we don't have any cfg(macos) code AFAICT

dpc:
I think I remember some conditional compilation etc that was failing on MacOS, so let's keep it. We'll see how it fares.

Maan2003 (approved):

Maan2003 (approved):

PR #3227, opened 2023-09-18
refactor: remove TieredMulti from MintInput and MintOutput, use single ecash notes instead

joschisan:
The goal of this refactor is to simplify the mint client and mint consensus. Furthermore, this refactor is necessary to prepare a follow up pr https://github.com/fedimint/fedimint/pull/3419 in which we will move the aggregation of ecash signature shares from the federations mint module in the clients, to save disk space for the broadcast and cryptographic compute. This pr proceeds in the following steps:
* We enable the primary output to create multiple inputs and outputs when finalising a transaction, so that the mint module can only create inputs and outputs with certain denominations.
* We only build transactions with a single ecash notes for every input and output.
* We change the state machines for inputs, outputs, and out-of-band transmission to work with single ecash notes instead of TieredMulti
* We change the ecash recovery to compute a utxo set first and then use the mint output state machine to retrieve the blind signatures for every utxo. Though this has been previously calculated from the mint consensus items, we now ignore those in preparation for the change to the client side ecash aggregation which will remove the mint consensus items. This enables a simple follow pr as well as overall simpler recovery state machine.
* Finally, we can simply remove the TieredMulti abstraction as the entire system has been switched to single notes.


fedimint-client/src/lib.rs
```diff
@@ -53,7 +53,7 @@
 //! module and its smart contracts are only used to incentivize LN payments, not
 //! to hold money. The mint module on the other hand holds e-cash note and can
 //! thus be used to fund transactions and to absorb change. Module clients with
-//! this ability should implement [`ClientModule::supports_being_primary`] and
+//! this ability should implement [`ClientModule::  supports_being_primary`] and
```
dpc, line 56: nit

dpc:
I don't understand what exactly is going on. (I know still a draft, just saying).

fedimint-client/src/transaction/builder.rs
```diff
@@ -76,6 +76,22 @@ impl TransactionBuilder {
         self
     }
 
+    pub fn with_inputs(mut self, inputs: Vec<ClientInput>) -> Self {
+        for input in inputs {
+            self.inputs.push(input);
+        }
```
elsirion, line 82: Either use `Vec::append` or make `inputs` an iterator and use `Vec::extend`.

fedimint-client/src/lib.rs
```diff
@@ -1098,14 +1118,18 @@ impl ClientInner {
                     missing_amount,
                 )
                 .await?;
-            partial_transaction.inputs.push(input);
+            partial_transaction.inputs.extend(inputs);
         }
 
-        let mut change_idx: Option<u64> = None;
+        let mut change_range = Range {
```
dpc, line 1124: **suggestion:** This being mutable, just to update it later is confusing to the reader. Probably better to just save the `range_start = ...` here, and construct the range at the end.

fedimint-client/src/lib.rs
```diff
@@ -1128,20 +1153,23 @@ impl ClientInner {
 
         let (tx, states) = partial_transaction.build(&self.secp_ctx, thread_rng());
 
-        Ok((tx, states, change_idx))
+        Ok((tx, states, change_range))
```
dpc, line 1156: **suggestion:** Please add a comment explaining the range, as it's even more non-obvious than it was before now. :D

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -306,38 +270,20 @@ impl MintRestoreInProgressState {
     /// errors via `sender` itself.
     ///
     /// TODO: could be internal to recovery_loop?
-    fn fetch_block_stream<'a>(
+    async fn await_block<'a>(
         api: DynGlobalApi,
         decoders: ModuleDecoderRegistry,
-        epoch_range: Range<u64>,
```
dpc, line 312: **issue (blocking):** What's the reason to remove it? It was there so we can process the blocks/epochs while this stream is downloading them in parallel (with some buffer) as a stream, making things go faster.

modules/fedimint-mint-client/src/backup/recovery/tests.rs
```diff
@@ -1,717 +0,0 @@
-use std::collections::{BTreeMap, HashMap};
```
dpc, line 1: **question:** Did this got deleted or moved somewhere?

dpc (commented):
LGTM in a big picture, pointed out some issues (and there were some existing ones too).

modules/fedimint-mint-client/src/backup/recovery.rs
joschisan: The previous setting was to process up to 500 epochs in one stream, which is roughly one block so I removed it to just use blocks directly. Saves on complexity for now.

modules/fedimint-mint-client/src/backup/recovery/tests.rs
joschisan: The tests got deleted. The differences between the old and new design where big and the new design seemed simple enough that the devimint integration tests for recovery seemed sufficient.

fedimint-client/src/lib.rs
joschisan: I tried to avoid mutable state here but in this case it actually seemed like the the cleanest way to do it imo.

fedimint-client/src/lib.rs
joschisan: Done.

modules/fedimint-mint-client/src/backup/recovery.rs
dpc: While the code was processing the epoch/block next one could have been downloading already, which is a significant improvement.

Oh well.

fedimint-client/src/lib.rs
dpc: @joschisan

dpc (approved):

dpc (approved):

modules/fedimint-mint-client/src/backup/recovery.rs
elsirion: Should we track this somewhere? I remember recovery performance was a problem in the past.

joschisan: The performance of the new recovery may have changed significantly. We now only download, verify and aggregate the blind signature shares for our unspent nonces; therefore we save a lot of bandwidth and possibly noticeable amount of compute. This might help with previous performance issues already. So a lot of variables are changed now.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -280,14 +285,19 @@ impl MintClientExt for Client {
...
+                MintCommonGen::KIND.as_str(),
+                operation_meta_gen,
+                tx,
+            )
+            .await
+            .context("We already reissued these notes")?
+            .1;
+
+        self.await_primary_module_outputs(operation_id, change)
+            .await?;
```
maan2003, line 300: this made `reissue_external_notes` block forever when offline, is this expected behavior?

cc @elsirion @joschisan

joschisan: No it is not, the await_primary_module_outputs call should be in subscribe_reissue_external_notes. We need to return the change variable with the operation id and make it an argument in subscribe_reissue_external_notes.

maan2003: I thought it was just a copy paste bug, created an issue to discuss further
https://github.com/fedimint/fedimint/issues/4355

PR #3241, opened 2023-09-19
feat: test gateway reboot

okjodom:
Reboots a running gateway instance and asserts its state progression up to `Running` state.
Feels like this should be part of gateway package, though it should still use devimint framework. hmm

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -889,6 +889,79 @@ async fn lightning_gw_reconnect_test(dev_fed: DevFed, process_mgr: &ProcessManag
     Ok(())
 }
 
+async fn gw_reboot_test(dev_fed: DevFed, process_mgr: &ProcessManager) -> Result<()> {
+    #[allow(unused_variables)]
+    let DevFed {
+        bitcoind,
+        cln,
+        lnd,
+        fed,
+        gw_cln,
+        gw_lnd,
+        electrs,
+        esplora,
+        faucet,
+    } = dev_fed;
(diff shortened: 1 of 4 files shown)


devimint/src/main.rs
```diff
@@ -889,6 +889,79 @@ async fn lightning_gw_reconnect_test(dev_fed: DevFed, process_mgr: &ProcessManag
...
+    } = dev_fed;
+
+    // Query current gateway infos
+    let mut cln_cmd = cmd!(gw_cln, "info");
+    let mut lnd_cmd = cmd!(gw_lnd, "info");
+    let (cln_value, lnd_value) = tokio::try_join!(cln_cmd.out_json(), lnd_cmd.out_json())?;
+
+    // Drop references to cln and lnd gateways so the test can kill them
+    drop(gw_cln);
+    drop(gw_lnd);
```
m1sterc001guy, line 913: Does this explicitly kill the lnd gateway process? For the lightning reconnect test, we actually need to call `stop_lightning_node` to kill the process

okjodom: this only kills gatewayd but preserves the lightning node processes. it then respawns gatewayd and provides handles to lightning

m1sterc001guy: So `drop` kills the gateway process then?

okjodom: true. thats the intent here

m1sterc001guy (approved):

PR #3242, opened 2023-09-19
Revert "Revert "chore: use a new incremental zstd crane mode""

dpc:
Our build deps being the worst case scenario, and MacOS always a snowflake, we've hit some unforeseen problems after we've landed.

Has been quickly reverted and now I fixed it, so we unrevert it.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -47,17 +47,17 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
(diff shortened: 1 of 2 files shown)


elsirion (approved):
Here goes nothing!

dpc:
:crossed_fingers: 

We now test MacOs in the merge queue, so I'm more confident about it.

Issue #3204, opened 2023-09-15
(client-ng): support dbdump with the new client


joschisan:

m1sterc001guy:
I don't think we completed this right? Maybe the issue just needs to be renamed

m1sterc001guy:
@elsirion the new client uses the db to save data for the state machines. Do you think this is worth exposing via `dbtool`? I think this data might be kind of hard to interpret, since its contextual to the state machine.

justinmoon:
> 
I don't think we completed this right?

Oops

elsirion:
> I think this data might be kind of hard to interpret, since its contextual to the state machine.

We need a decoder registry I guess so we can decode module-specific data.

We kinda already do that for the server side, but not very elegantly/extensible. So ideally the `dbtool` would become something one initializes with a custom set of modules when forking Fedimint like the main `fedimintd` and `fedimint-cli`. It would also need a config file or at least instance-kind mapping.

PR #3243, opened 2023-09-19
feat: modularize client dbdump

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3204

Also adds support for dumping gatewayd's database.

Moved `dump_database` to a common location for both the server modules and the client modules to implement. `dbtool` still just attaches the default modules (mint, ln, wallet). If more modules are needed it should be straightforward to fork and add the modules, with little code change.


fedimint-rocksdb/src/lib.rs
```diff
@@ -31,10 +32,10 @@ impl RocksDb {
 }
 
 impl RocksDbReadOnly {
-    pub fn open_read_only(db_path: impl AsRef<Path>) -> Result<RocksDbReadOnly, rocksdb::Error> {
+    pub fn open_read_only(db_path: String) -> Result<RocksDbReadOnly, rocksdb::Error> {
```
elsirion, line 35: Why not `PathBuf`? I'm also not a fan of hiding this change in this commit, could use some explanation in a commit message.

elsirion (approved):
Looks ok, not a big fan of the read only DB change, would like to know the reasoning.

fedimint-rocksdb/src/lib.rs
m1sterc001guy: Let me see if I can fix this before merging

m1sterc001guy: Sorry I was being dumb, these changes aren't necessary. Reverted all changes in `fedimint-rocksdb`

m1sterc001guy:
> Looks ok, not a big fan of the read only DB change, would like to know the reasoning.

DB change was not necessary and has been removed

elsirion (approved):

PR #3245, opened 2023-09-20
 chore: better errors on federation id or peer id invite code mismatch 

### LINKED ISSUE #3195: Verify `InviteCode::id` on the server side
> It seems to me like the server side is not checking `InviteCode::id` at all, which probably is not a best way to handle it. We should return `incorrect federation id`, possibly preventing some weird confusion attacks etc.

dpc:
Fix #3195


elsirion (dismissed):
The current config fetching is a mess, we should support multiple URLs, but till we do this doesn't make anything worse I guess …

dpc:
Good CI. Fixed, hopefully.

elsirion (approved):

Issue #2309, opened 2023-04-24
Manual recovery option for failed federations


elsirion:
Currently, when more than `f=n-t` guardians are shut down in an uncoordinated manner (without calling the `upgrade` API endpoint) it can happen that some are shut down *while* processing an epoch while others are already finished, resulting in some guardians being on different epochs than others.

This poses a problem when restarting since restarts from such a failed state (>`f` offline = failed/stalled federation) requires a quorum of `t` guardians to agree on the epoch the federation is in. If the guardians can agree out of band that the most up-to-date epoch any guardian observed actually happened they should be able to recover manually though.

For this to be practical we need an API endpoint that allows manual approval/application of epoch outcomes, either by specifying the trusted epoch hash and letting `fedimintd` fetch the epoch from its peers or by giving it whole epoch outcomes. I'm not sure what the optimal design is, either would work imo, the former being slightly more convenient.

CC: @jkitman @dpc

jkitman:
Seems okay...but we should never start from an older epoch or we will get an inconsistent database.

PR #3250, opened 2023-09-20
Consensus code cleanup

tvolk131:


fedimint-server/src/consensus/mod.rs
```diff
@@ -424,7 +424,9 @@ impl FedimintConsensus {
         current
     }
 
+    /// Wait for any module to send us a proposal
```
elsirion, line 427: @joschisan will remove this method soon. Might be worth dropping the commit to save him some rebase conflicts.

fedimint-server/src/consensus/server.rs
```diff
@@ -442,7 +441,7 @@ impl ConsensusServer {
                 _ => break vec![],
             };
         };
-        let proposal = self.process_events_then_propose(override_proposal).await;
```
elsirion, line 461: On mobile, so can't test, but why doesn't this break the functionality that we can force a guardian to accept a specific epoch? See #2309

elsirion (commented):
Nicely structured PR!

fedimint-server/src/consensus/mod.rs
tvolk131: Done.

fedimint-server/src/consensus/server.rs
tvolk131: I'm not familiar enough with the consensus code to have much input on that. I just noticed that the only place `override_proposal` was used, `None` was passed in. If there are plans to eventually use this for something, I can drop the commit!

dpc: I think it's a leftover of forcing an epoch being moved to be admin api endpoint.

```
    // Save any API events we have in the channel then create a proposal
    async fn process_events_then_propose(
        &mut self,
        override_proposal: Option<ConsensusProposal>,
    ) -> ConsensusProposal {
        while let Some(Some(event)) = self.api_receiver.next().now_or_never() {
            match event {
                ApiEvent::ForceProcessOutcome(outcome) => self.force_process_epoch(outcome).await,
                event => {
                    self.consensus.api_event_cache.insert(event);
                }
            };
        }
        let consensus_proposal = self.consensus.get_consensus_proposal().await;
        self.consensus.api_event_cache.clear();
        override_proposal.unwrap_or(consensus_proposal)
    }
```

We now force an epoch as an event sent through `api_endpoint`, while the `override_proposal: Option<ConsensusProposal>,` never actually gets set to any `Some` on the server side. @elsirion

dpc (approved):

elsirion (approved):
Some good improvements!

PR #3252, opened 2023-09-20
chore: implement unbalanced_transactions_get_rejected

douglaz:
Replaces the old test:

https://github.com/fedimint/fedimint/blob/1132d742c2357fd70a0c3691636f80637edc0adb/integrationtests/tests/tests.rs#L167C10-L167C46

### DIFF
--- a/modules/fedimint-dummy-tests/tests/tests.rs
+++ b/modules/fedimint-dummy-tests/tests/tests.rs
@@ -1,11 +1,19 @@
+use std::sync::Arc;
+
+use anyhow::bail;
+use fedimint_client::transaction::{ClientOutput, TransactionBuilder};
+use fedimint_core::api::GlobalFederationApi;
 use fedimint_core::config::ClientModuleConfig;
-use fedimint_core::core::ModuleKind;
+use fedimint_core::core::{IntoDynInstance, ModuleKind};
 use fedimint_core::module::ModuleConsensusVersion;
 use fedimint_core::{sats, Amount};
-use fedimint_dummy_client::{DummyClientExt, DummyClientGen};
+use fedimint_dummy_client::states::DummyStateMachine;
+use fedimint_dummy_client::{DummyClientExt, DummyClientGen, DummyClientModule};
 use fedimint_dummy_common::config::{DummyClientConfig, DummyGenParams};
+use fedimint_dummy_common::DummyOutput;
 use fedimint_dummy_server::DummyGen;
 use fedimint_testing::fixtures::Fixtures;
+use secp256k1::Secp256k1;
 
 fn fixtures() -> Fixtures {
     Fixtures::new_primary(DummyClientGen, DummyGen, DummyGenParams::default())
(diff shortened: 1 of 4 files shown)


modules/fedimint-dummy-tests/tests/tests.rs
```diff
@@ -70,3 +78,29 @@ async fn federation_should_abort_if_balance_sheet_is_negative() -> anyhow::Resul
...
+    let account = client.account();
+
+    let (_dummy, instance) =
+        client.get_first_module::<DummyClientModule>(&fedimint_dummy_common::KIND);
+    let output = ClientOutput {
+        output: DummyOutput { amount, account },
+        state_machines: Arc::new(move |_, _| Vec::<DummyStateMachine>::new()),
+    };
+    let tx = TransactionBuilder::new().with_output(output.into_dyn(instance.id));
+    let secp = Secp256k1::new();
```
dpc, line 97: Style: would inline these two (`secp` and `rng`).

modules/fedimint-dummy-tests/tests/tests.rs
```diff
@@ -70,3 +78,29 @@ async fn federation_should_abort_if_balance_sheet_is_negative() -> anyhow::Resul
 
     Ok(())
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn unbalanced_transactions_get_rejected() -> anyhow::Result<()> {
+    let fed = fixtures().new_fed().await;
+    let client = fed.new_client().await;
+
+    let amount = sats(1000);
```
dpc, line 87: I'm only nitpicky because I'm trying to understand what is this test doing.

This `amount` is kind of vague. Would inline and rename to `output_value` or something.

modules/fedimint-dummy-tests/tests/tests.rs
```diff
@@ -70,3 +78,29 @@ async fn federation_should_abort_if_balance_sheet_is_negative() -> anyhow::Resul
 
     Ok(())
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn unbalanced_transactions_get_rejected() -> anyhow::Result<()> {
```
dpc, line 83: `/// Make sure that a tx with more output value than inputs get rejected`?

But why is this test in a dummy module in the first place?

elsirion: You need some module to generate an unbalanced tx, the dummy module is the most straight-forward. We could have some global integration testing crate again, but it wouldn't do anything differently imo, so we can just put it there. 

For the other test that checks the audit we actually need the dummy module because there we can patch in an accounting bug.

elsirion (approved):

modules/fedimint-dummy-tests/tests/tests.rs
douglaz: Added a minimal commentary

dpc (approved):

PR #3254, opened 2023-09-20
refactor: Future proof arguments to ClientModuleInit::init

dpc:
Re #3025

### DIFF
--- a/fedimint-client/src/module/init.rs
+++ b/fedimint-client/src/module/init.rs
@@ -88,7 +88,6 @@ pub trait ClientModuleInit: ExtendsCommonModuleInit + Sized {
     fn supported_api_versions(&self) -> MultiApiVersion;
 
     /// Initialize a [`ClientModule`] instance from its config
-    #[allow(clippy::too_many_arguments)]
     async fn init(&self, args: &ClientModuleInitArgs<Self>) -> anyhow::Result<Self::Module>;
 }
 
--- a/fedimint-client/src/module/init.rs
+++ b/fedimint-client/src/module/init.rs
@@ -18,6 +18,67 @@ use crate::DynGlobalClientContext;
 
 pub type ClientModuleInitRegistry = ModuleInitRegistry<DynClientModuleInit>;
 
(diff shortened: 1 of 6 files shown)


elsirion (dismissed):
:tada:

fedimint-client/src/module/init.rs
```diff
@@ -102,16 +153,16 @@ where
     ) -> anyhow::Result<DynClientModule> {
         let typed_cfg: &<<T as fedimint_core::module::ExtendsCommonModuleInit>::Common as CommonModuleInit>::ClientConfig = cfg.cast()?;
         Ok(self
-            .init(
+            .init(&ClientModuleInitArgs {
```
elsirion, line 156: One thing I remembered when reviewing the server version of this PR: wouldn't we want the same struct as a dyn version? I guess it's less necssary since we view the dyn stuff as internal API, not to be touched by module authors?

elsirion: I removed the PR from the merge queue for now, just merge if that was your reasoning and not an oversight.

fedimint-client/src/module/init.rs
```diff
@@ -28,17 +89,7 @@ pub trait ClientModuleInit: ExtendsCommonModuleInit + Sized {
 
     /// Initialize a [`ClientModule`] instance from its config
     #[allow(clippy::too_many_arguments)]
```
m1sterc001guy, line 91: nit: probably can get rid of this `allow` now

fedimint-client/src/module/init.rs
dpc: Crossed my mind, but as it is an internal signature it doesn't seem necessary. It has only one generic implementation, so it doesn't seem to gain anthing, other than maybe keeping it more 1:1 to the typed version. I mean - it would be like taking any internal function with N arguments and making it a struct. 

@elsirion .

justinmoon (approved):
Reviewed lightly. Most re-approving based on @elsirion's previous approval.

PR #3257, opened 2023-09-21
chore: Derivable secret and query strategy docs

dpc:
Finish https://github.com/fedimint/fedimint/pull/2981

Drop tricky parts, move some stuff.  Let's just get in what we can so it doesn't go to waste.

### DIFF
--- a/crypto/derive-secret/src/lib.rs
+++ b/crypto/derive-secret/src/lib.rs
@@ -97,8 +97,8 @@ impl DerivableSecret {
         }
     }
 
-    // secp256k1 keys are used for bitcoin deposit addresses, redeem keys and
-    // contract keys for lightning.
+    /// secp256k1 keys are used for bitcoin deposit addresses, redeem keys and
+    /// contract keys for lightning.
     pub fn to_secp_key<C: Signing>(self, ctx: &Secp256k1<C>) -> KeyPair {
         for key_try in 0u64.. {
             let secret = self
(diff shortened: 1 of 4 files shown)


crypto/derive-secret/src/lib.rs
```diff
@@ -19,16 +32,23 @@ const RAW_BYTES: &[u8; 8] = b"rawbytes";
...
+    ///   Function) with SHA-512 as the underlying hash function. It is used to
+    ///   derive child keys.
     // TODO: wrap in some secret protecting wrappers maybe?
     kdf: Hkdf<Sha512>,
 }
 
 impl DerivableSecret {
+    /// Derive root secret key from a secret material and salt.
+    ///
+    /// The `salt` is just additional data t used
```
justinmoon, line 50: `t` can be removed

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1163,6 +1188,16 @@ impl MintClientModule {
     ///
     /// Static to help re-use in other places, that don't have a whole [`Self`]
     /// available
+    ///
+    /// # E-Cash Note Creation
+    ///
+    /// When creating an e-cash note, the `MintClientModule` first derives the
+    /// blinding and spend keys from the `DerivableSecret`. It then creates a
+    /// `NoteIssuanceRequest` containing the blinded spend key and sends it to
+    /// the mint server. The mint server signs the blinded spend key and
```
justinmoon, line 1197: `containing the blinded spend key` should probably be `containing the blinded spend public key`.

justinmoon (approved):
2 little suggestions

PR #3260, opened 2023-09-21
feat: switch to Aleph BFT

joschisan:


fedimint-atomic-broadcast/src/db.rs
```diff
@@ -2,27 +2,31 @@ use std::io::Cursor;
...
+pub struct SignedBlockKey(pub u64);
+
+#[derive(Debug, Encodable, Decodable)]
+pub struct SignedBlockPrefix;
 
 impl_db_record!(
     key = SignedBlockKey,
     value = SignedBlock,
-    db_prefix = 0x00,
+    db_prefix = 0x04,
```
justinmoon, line 18: elsewhere in the codebase we use enums for these constants

joschisan: We will switch to the whims before we merge, will require to move the fedimint-atomic-broadcast crate in a module in fedimint-server to avoid a cyclical dependency between the crates

Issue #3167, opened 2023-09-12
Allow updating of Gatewayd's parameters


m1sterc001guy:
After https://github.com/fedimint/fedimint/pull/3154 is merged, we need to move the other gateway configurations to use this API.

Configs:
 - Password (DONE)
 - Gateway fees
 - Number of Route Hints

PR #3273, opened 2023-09-21
feat: Gateway configure RoutingFees and number of route hints

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3167

Allows the gateway operator to configure their routing fees and the number of route hints at run time.


gateway/ln-gateway/src/lib.rs
```diff
@@ -134,6 +134,26 @@ pub struct GatewayOpts {
...
+impl GatewayOpts {
+    fn to_gateway_parameters(&self) -> GatewayParameters {
+        // TODO: Consume self
+        GatewayParameters {
+            listen: self.listen,
+            password: self.password.clone()
+        }
+    }
+}
+*/
```
elsirion: Can we remove this?

elsirion (dismissed):
Looks reasonable, maybe someone with more context about the GW wants to do a more thorough review.

gateway/ln-gateway/src/lib.rs
```diff
@@ -134,6 +134,26 @@ pub struct GatewayOpts {
     pub num_route_hints: Option<usize>,
 }
 
+/*
```
okjodom: nit, commented code?

m1sterc001guy: Fixed in a later commit , perhaps I should squash those two together

gateway/ln-gateway/src/lib.rs
```diff
@@ -500,10 +513,15 @@ impl Gateway {
             lightning_alias,
         } = self.state.read().await.clone()
         {
+            let gateway_config = self
+                .get_gateway_configuration()
+                .await
+                .expect("Gateway configuration should be set");
```
okjodom, line 533: should we gracefully exit the api call instead?

m1sterc001guy: It should be guaranteed to exist at this point. `get_gateway_configuration` should only return `None` if we're in the `Configuring` state. I'll add a comment

okjodom (dismissed):

okjodom (approved):

Issue #3259, opened 2023-09-21
`just test-ci-all` fails locally on linux x86-64


bradleystachurski:
Running `just test-ci-all` locally consistently fails for `backend_test_*` on [master](https://github.com/fedimint/fedimint/tree/20c85475d522d447c481120d9065025590d5bc84).

```
Some tests failed. Full job log:
Seq     Host    Starttime       JobRuntime      Send    Receive Exitval Signal  Command
1       :       1695303761.978       0.029      0       63      0       0       always_success_test
7       :       1695303793.687      43.157      0       12842   0       0       lightning_reconnect_test
9       :       1695303804.610      43.655      0       12246   0       0       load_test_tool_test
8       :       1695303799.537      74.386      0       14958   0       0       devimint_cli_test
5       :       1695303782.622     117.121      0       10977   0       0       latency_test
6       :       1695303787.740     115.105      0       15324   0       0       reconnect_test
4       :       1695303777.588     249.151      0       5774    100     0       backend_test_esplora
3       :       1695303772.046     260.084      0       63010   100     0       backend_test_electrs
2       :       1695303767.002     284.655      0       93649   100     0       backend_test_bitcoind
```

CI is working for `Build on linux`
https://github.com/fedimint/fedimint/actions/runs/6260474673/job/16998577685

Mac users aren't reporting any issues, however other users are observing failed tests locally (Discord [thread](https://discord.com/channels/990354215060795454/990354215878688860/1154223019603005551)).

dpc:
@fedimint/infra

douglaz:
I think this is a different issue

dpc:
Too many issues. :D

dpc:
Probably: https://github.com/fedimint/fedimint/pull/3269

justinmoon:
Call: was consistently failing for @bradleystachurski but not only 10% of time ... decreasing parallelism in `cargo nextest` can help fix this because it's fighting over bitcoind lock

bradleystachurski:
`just test-ci-all` is consistently passing locally, thanks to https://github.com/fedimint/fedimint/pull/3589 and likely several other improvements since this issue was created.

We can open a separate issue if we start observing failures again.

Issue #3209, opened 2023-09-16
Improve backend tests (especially ln-gateway) scalability


dpc:
devimint starts only one lnd gateway. LND gateway can have only one HTLC interceptor at the time. All ln gateway tests want to be an interceptor. Tests try to run in parallel. Effectively ln-gateway only work because they keep trying until they register.

In addition some tests will deadlock if run in parallel in certain combinations because of this.

We also have some tests that use [`lock_exclusive`](https://github.com/fedimint/fedimint/blob/19606001b703dd44a39bebdb2ceb623115528e83/fedimint-testing/src/btc/mod.rs#L13) to get an exclusive lock on bitcoin node.

This is all not very scalable.

Some approaches we could take (possibly in combination) sorted by amount of work:

#### Add LND lock

Implement something like `lock_exclusive` for LND instance, to at least eliminate deadlocks.

### Start multiple LND instances

Combined with the locking should be relatively simple and unblock some level of parallelism.

### Have devimint start and manage a pool of LND nodes

Just have each test request an LND instance to use, then get it locked, all managed by devimint. Would require some layer of rpc from test to devimint, but can be very elastic.

Some way to cap the maximum amount of these might be needed.

### Have a devimint manage a whole pool of environments.

Like "pool of LND nodes" but for the whole environment. This could eliminate bitcoind lock_exclusive. If a test needs an exclusive access to bitcoind, just give it a whole deviming instance. So we could have one devimin binary spawn multiple "devimint env" instances.

Some way to cap the maximum amount of these might be needed.

dpc:
@fedimint/infra

m1sterc001guy:
@bradleystachurski I assigned this to you, let me know if that's ok or if you need any pointers.

bradleystachurski:
@m1sterc001guy Thanks! Definitely ok, I'll holler if I get stuck and need some pointers.

bradleystachurski:
Dev call: split into separate issue

elsirion:
@dpc @bradleystachurski still relevant?

m1sterc001guy:
IMO this can be closed. `ln-gateway` tests were refactored to only use mocks, we rely on devimint tests now for E2E which can be parallelized much easier.

PR #3278, opened 2023-09-22
 fix: properly shutdown the gateway

douglaz:
Fix a infinite loop on `test_gateway_filters_route_hints_by_inbound`  where a `GatewayTest` will run, then another `GatewayTest` will start but all tasks and HTLC interceptors from the previous one are still running.

This is specially problematic for LND because you can't have two or more HTLC interceptors (both will fail).

So the solution is to properly shutdown the `GatewayTest` on `Drop`. This required to review all "blocking" calls to also select `handle.make_shutdown_rx()`.

I also added new debugging logs that helped find and solve the issue.

This improves https://github.com/fedimint/fedimint/issues/3259

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -341,11 +341,9 @@ impl Gateway {
         Box::new(gateway_items.into_iter())
     }
 
-    pub async fn run(mut self) -> anyhow::Result<TaskShutdownToken> {
-        let mut tg = TaskGroup::new();
-
-        self.start_webserver(&mut tg).await;
-        self.start_gateway(&mut tg).await?;
+    pub async fn run(mut self, tg: &mut TaskGroup) -> anyhow::Result<TaskShutdownToken> {
+        self.start_webserver(tg).await;
+        self.start_gateway(tg).await?;
         let handle = tg.make_handle();
         let shutdown_receiver = handle.make_shutdown_rx().await;
         Ok(shutdown_receiver)
@@ -377,22 +375,17 @@ impl Gateway {
                     .await
                     .expect("Failed to start webserver");
                     info!("Successfully started webserver");
-
-                    gateway_db
-                        .wait_key_check(&GatewayConfigurationKey, |v| {
-                            v.filter(|cfg| {
-                                if let Some(old_config) = gateway_config.clone() {
-                                    old_config.password != cfg.clone().password
-                                } else {
-                                    true
-                                }
-                            })
-                        })
-                        .await;
-
-                    info!("GatewayConfiguration has been updated, restarting webserver...");
-                    if let Err(e) = webserver_group.shutdown_join_all(None).await {
-                        panic!("Error shutting down server: {e:?}");
+                    tokio::select! {
+                        _ = wait_for_new_password(&gateway_db, gateway_config) => {
+                            info!("GatewayConfiguration has been updated, restarting webserver...");
+                            if let Err(e) = webserver_group.shutdown_join_all(None).await {
+                                panic!("Error shutting down server: {e:?}");
+                            }
+                        },
+                        _ = handle.make_shutdown_rx().await => {
+                            info!("Received shutdown signal, exiting....");
+                            break;
+                        }
                     }
                 }
             })
@@ -418,6 +411,7 @@ impl Gateway {
                 move |handle| async move {
                     loop {
                         if handle.is_shutting_down() {
+                            info!("Gateway HTLC handler loop is shutting down");
                             break;
                         }
 
@@ -452,9 +446,17 @@ impl Gateway {
                                             lightning_alias,
                                         }).await;
 
-                                        // Blocks until the connection to the lightning node breaks
-                                        self.handle_htlc_stream(stream, handle.clone()).await;
-                                        warn!("HTLC Stream Lightning connection broken. Gateway is disconnected");
+                                        // Blocks until the connection to the lightning node breaks or we receive the shutdown signal
+                                        tokio::select! {
+                                            _ = self.handle_htlc_stream(stream, handle.clone()) => {
+                                                warn!("HTLC Stream Lightning connection broken. Gateway is disconnected");
+                                            },
+                                            _ = handle.make_shutdown_rx().await => {
+                                                info!("Received shutdown signal");
+                                                self.handle_disconnect(htlc_task_group).await;
+                                                break;
+                                            }
+                                        }
                                     }
                                     Err(e) => {
                                         error!("Failed to retrieve Lightning info: {e:?}");
@@ -466,10 +468,7 @@ impl Gateway {
                             }
                         }
 
-                        self.set_gateway_state(GatewayState::Disconnected).await;
-                        if let Err(e) = htlc_task_group.shutdown_join_all(None).await {
-                            error!("HTLC task group shutdown errors: {}", e);
-                        }
+                        self.handle_disconnect(htlc_task_group).await;
 
                         error!("Disconnected from Lightning Node. Waiting 5 seconds and trying again");
                         sleep(Duration::from_secs(5)).await;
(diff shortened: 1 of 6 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -406,9 +400,16 @@ impl Gateway {
...
-                                        self.handle_htlc_stream(stream, handle.clone()).await;
-                                        warn!("HTLC Stream Lightning connection broken. Gateway is disconnected");
+                                        // Blocks until the connection to the lightning node breaks or we receive the shutdown signal
+                                        tokio::select! {
+                                            _ = self.handle_htlc_stream(stream, handle.clone()) => {
+                                                warn!("HTLC Stream Lightning connection broken. Gateway is disconnected");
+                                            },
+                                            _ = handle.make_shutdown_rx().await => {
+                                                info!("Received shutdown signal");
+                                                break;
```
bradleystachurski, line 410: Ahh nice, thanks for adding a check for a shutdown signal on the HTLC stream 👍 

If we break the loop here, we won't hit the codepath to set the gateway state to disconnected and to shutdown the htlc task group.

https://github.com/douglaz/fedimint/blob/fix_tests/gateway/ln-gateway/src/lib.rs#L424-L427

Perhaps we should create a `handle_disconnect` function that's called before the break and also at the end of the loop?

bradleystachurski (commented):
Thanks for digging into this!

Running `just test-ci-all` on this branch, I'm still getting errors with `backend_test_bitcoind`. Are you also getting errors on your end?

```
Seq     Host    Starttime       JobRuntime      Send    Receive Exitval Signal  Command
1       :       1695391419.341       0.030      0       63      0       0       always_success_test
8       :       1695391454.988      43.088      0       10541   0       0       gateway_reboot_test
7       :       1695391449.867      50.742      0       13626   0       0       lightning_reconnect_test
10      :       1695391465.452      45.032      0       11978   0       0       load_test_tool_test
9       :       1695391460.131      75.519      0       13360   0       0       devimint_cli_test
5       :       1695391439.689     151.527      0       10977   0       0       latency_test
6       :       1695391444.743     165.076      0       14753   0       0       reconnect_test
3       :       1695391429.642     410.755      0       6512    0       0       backend_test_electrs
4       :       1695391434.660     412.970      0       7400    0       0       backend_test_esplora
2       :       1695391424.381     601.398      0       8271    -1      15      backend_test_bitcoind
```

gateway/ln-gateway/src/lib.rs
m1sterc001guy: +1. I'd be in favor of creating more helper functions here, maybe everything inside of `Ok((stream, ln_client))` can be a function. The diff is just getting hard to read

gateway/ln-gateway/src/lib.rs
```diff
@@ -442,40 +443,59 @@ impl Gateway {
...
+                        }
+                    }
+                    other => {
+                        info!("Got {other:?} while handling HTLC stream, exiting from loop...");
+                        break;
+                    }
                 }
             }
+        } else {
+            info!("Gateway isn't in a running state")
```
m1sterc001guy, line 498: IMO this should be an error, we shouldn't be intercepting HTLCs if we're not in the `Running` state

gateway/ln-gateway/src/lib.rs
```diff
@@ -442,40 +443,59 @@ impl Gateway {
...
+                        let scid_to_feds = self.scid_to_federation.read().await;
+                        let federation_id = scid_to_feds.get(&htlc_request.short_channel_id);
+                        // Just forward the HTLC if we do not have a federation that
+                        // corresponds to the short channel id
+                        if let Some(federation_id) = federation_id {
+                            let clients = self.clients.read().await;
+                            let client = clients.get(federation_id);
+                            // Just forward the HTLC if we do not have a client that
+                            // corresponds to the federation id
+                            if let Some(client) = client {
```
m1sterc001guy, line 462: Maybe we can split some of these `if let Some` patterns into helper functions to help with indentation here too

gateway/ln-gateway/src/lib.rs
```diff
@@ -372,6 +365,7 @@ impl Gateway {
                 move |handle| async move {
                     loop {
                         if handle.is_shutting_down() {
+                            info!("We are shutting down, exiting from loop...");
```
m1sterc001guy, line 368: Nit: Maybe a more description message like "Gateway HTLC handler loop is shutting down"

gateway/ln-gateway/src/lnd.rs
```diff
@@ -112,21 +112,30 @@ impl GatewayLndClient {
...
+                                error!("Failed to establish htlc stream");
+                                let e = LightningRpcError::FailedToGetRouteHints {
+                                    failure_reason: format!("Failed to subscribe to LND htlc stream {e:?}"),
+                                };
+                                debug!("Error: {e}");
+                                return;
+                            }
+                        }
+                    },
+                    _ = handle.make_shutdown_rx().await => {
```
m1sterc001guy, line 132: Ugh. This is a little weird that you have to do this. We already wait on the shutdown future below when we select over `message()` and `make_shutdown_rx`. But I think we've seen the `htlc_interceptor` function block before the first HTLC comes in. Is that why this is necessary?

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -859,55 +859,64 @@ async fn test_gateway_configuration() -> anyhow::Result<()> {
...
+
+            let invite2 = fed2.invite_code();
+            let info = rpc
+                .connect_federation(ConnectFedPayload {
+                    invite_code: invite2.to_string(),
+                })
+                .await
+                .unwrap();
+            assert_eq!(info.federation_id, invite2.id);
+            drop(gateway); // keep until the end to avoid the gateway shutting down too early
```
m1sterc001guy, line 886: Wont `drop` be called automatically after it goes out of scope?

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -859,55 +859,64 @@ async fn test_gateway_configuration() -> anyhow::Result<()> {
...
+            assert_eq!(info.federations.len(), 2);
+            assert!(info
+                .federations
+                .iter()
+                .any(|info| info.federation_id == id1 && info.balance_msat == Amount::ZERO));
+            assert!(info
+                .federations
+                .iter()
+                .any(|info| info.federation_id == id2 && info.balance_msat == Amount::ZERO));
+            drop(gateway); // keep until the end to avoid the gateway shutting down too early
```
m1sterc001guy, line 916: Same question as above

m1sterc001guy:
Do you know why CI hasn't been hitting this issue? Some sort of race when shutting down the previous gateway and starting the new one?

gateway/ln-gateway/tests/integration_tests.rs
douglaz: Yes, but if we don't use this variable it will drop before the test even starts

douglaz:
> Do you know why CI hasn't been hitting this issue? Some sort of race when shutting down the previous gateway and starting the new one?

Probably, but I'm not sure about that because running locally I can reproduce the issue 100% of time.

dpc (commented):
This is good, this is fresh. :+1: 

Ping me after addressing (or pushing back) on some of the comments, happy to help land.

douglaz:
> Thanks for digging into this!
> 
> Running `just test-ci-all` on this branch, I'm still getting errors with `backend_test_bitcoind`. Are you also getting errors on your end?
> 
> ```
> Seq     Host    Starttime       JobRuntime      Send    Receive Exitval Signal  Command
> 1       :       1695391419.341       0.030      0       63      0       0       always_success_test
> 8       :       1695391454.988      43.088      0       10541   0       0       gateway_reboot_test
> 7       :       1695391449.867      50.742      0       13626   0       0       lightning_reconnect_test
> 10      :       1695391465.452      45.032      0       11978   0       0       load_test_tool_test
> 9       :       1695391460.131      75.519      0       13360   0       0       devimint_cli_test
> 5       :       1695391439.689     151.527      0       10977   0       0       latency_test
> 6       :       1695391444.743     165.076      0       14753   0       0       reconnect_test
> 3       :       1695391429.642     410.755      0       6512    0       0       backend_test_electrs
> 4       :       1695391434.660     412.970      0       7400    0       0       backend_test_esplora
> 2       :       1695391424.381     601.398      0       8271    -1      15      backend_test_bitcoind
> ```

This fixed for me:
```diff
diff --git a/scripts/tests/test-ci-all.sh b/scripts/tests/test-ci-all.sh
index d2509f5324..bbd1e982db 100755
--- a/scripts/tests/test-ci-all.sh
+++ b/scripts/tests/test-ci-all.sh
@@ -112,7 +112,7 @@ PATH="$(pwd)/scripts/dev/run-isolated-test/:$PATH"
 if parallel \
   --halt-on-error 1 \
   --joblog "$joblog" \
-  --timeout 600 \
+  --timeout 1200 \
   --load 150% \
   --delay 5 \
   --jobs "$parallel_jobs" \
```

@bradleystachurski can you test on your side?

gateway/ln-gateway/src/lib.rs
douglaz: Added a `handle_disconnect`

gateway/ln-gateway/src/lib.rs
douglaz: Not feeling very creative today. Although the diff is shitty, the final code looks fine.

dpc:
20 minutes to finish a test? Even if they do run serially, that seems like kind of a lot.

dpc:
BTW. Check https://github.com/fedimint/fedimint/issues/3209 . Might be easier to build some devimint facilities to let these tests run somewhat in parallel than try to make them run reasonably fast serially.

douglaz:
> 20 minutes to finish a test? Even if they do run serially, that seems like kind of a lot.

What's funny here is that it seems the faster the machine the slower it runs

dpc:
> What's funny here is that it seems the faster the machine the slower it runs

We've seen it before when one "thing" (task, test, etc.) somehow interrupts/slows down another one, and they fight over what they need faster.

bradleystachurski (approved):
> @bradleystachurski can you test on your side?

Looks good!

```
## COMPLETE: backend_test_bitcoind
All tests successful
```

Once this is merged I'm happy to help tackle some followups (#3209, refactoring to cleanup indentation).

m1sterc001guy (approved):
LGTM let's merge and iterate from here. We can fix the indentation issues next and also improve the test so that it doesn't require a restart of the gateway

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Let's do this in a follow up

PR #3281, opened 2023-09-22
fix(mint-client): reject zero-value out-of-band spends

### LINKED ISSUE #3182: Make sure we can handle serialized `TieredMulti<SpendableNote>` with no notes inside it
Somewhere we might assume that this isn't empty -- for instance, in the creation of operation ids.

elsirion:
These could lead to weird error messages since all zero-value OOB spends have the same identity. All but the first would fail.

Fixes #3182

### DIFF
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -93,10 +93,14 @@ impl FromStr for OOBNotes {
     /// Decode a set of out-of-band e-cash notes from a base64 string.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let bytes = base64::decode(s)?;
-        Ok(Decodable::consensus_decode(
+        let oob_notes: OOBNotes = Decodable::consensus_decode(
             &mut std::io::Cursor::new(bytes),
             &ModuleDecoderRegistry::default(),
-        )?)
+        )?;
+
+        ensure!(!oob_notes.notes.is_empty(), "OOBNotes cannot be empty");
+
+        Ok(oob_notes)
     }
 }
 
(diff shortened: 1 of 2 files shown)


modules/fedimint-mint-tests/tests/tests.rs
```diff
@@ -40,3 +40,39 @@ async fn sends_ecash_out_of_band() -> anyhow::Result<()> {
...
+    let res = client1
+        .reissue_external_notes(
+            OOBNotes {
+                federation_id: client1.federation_id(),
+                notes: Default::default(),
+            },
+            (),
+        )
+        .await;
+    assert!(res.is_err(), "Zero-amount receives should be forbidden");
```
elsirion, line 76: I should check the error message.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -248,6 +248,11 @@ impl MintClientExt for Client {
             federation_id,
             notes,
         } = oob_notes;
+        ensure!(
```
dpc, line 251: Should we reject these on deserialization level?

elsirion: Maybe? It just seemed more natural to do it here because at least generating them has to be prevented in the client impl. I could add a second line of defense on the decoding level I guess.

elsirion: Addressed in [48a8125](https://github.com/fedimint/fedimint/pull/3281/commits/48a8125d2b26600cbe0903f5d9e1e393c936026b)

dpc (approved):

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -93,10 +93,14 @@ impl FromStr for OOBNotes {
     /// Decode a set of out-of-band e-cash notes from a base64 string.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let bytes = base64::decode(s)?;
-        Ok(Decodable::consensus_decode(
+        let oob_notes: OOBNotes = Decodable::consensus_decode(
```
dpc, line 96: Oh, I thought we would do it by making a custom implementation of `Decodable` & `Encodable`. :)

justinmoon (approved):

dpc:
> Oh, I thought we would do it by making a custom implementation of Decodable & Encodable. :)

Boo, landed. I think it would be much more solid if done there. @elsirion

elsirion:
Right, we could make it an invariant, but then we need a constructor that checks it too. That would make sense *instead* of the current approach, but I understood your comment as adding an earlier failure point in addition. Can always change it again.

PR #3284, opened 2023-09-24
docs: add module flowchart

EthnTuttle:
This flow chart came from session @justinmoon and I had, many thanks for the help. Repeated it back to @m1sterc001guy with good feedback and wanted to share with a wider audience.

I tried to capture the basic and generalized flow of a module creating something for consensus. The Client side is simplified in favor of focusing on the server side flow.

I have included a PNG and draw.io file in this PR. Please let me know if there's a better format for sharing. 

I hope for a lot of feedback on this item so it can be iterated on and expanded to other parts of the codebase that would be useful for external devs interested in integrating with Fedimint to know about.


EthnTuttle:
There also may be a better place to store/maintain this?

docs/fedimint-module.drawio.png
shaurya947, line 1: `await_consensus_proposal` could theoretically use any trigger—not just the existence of DB entries. In the Stability Pool module, for example, the trigger is the system clock.

Based on what I've learned, consensus is expensive and should be regarded as such. A module developer should think carefully as to what's the bare minimum that needs to be part of consensus, and what could be deterministically calculated by each guardian independently, even if they were catching up after being offline for a period of time. The BFT system guarantees that honest guardians see consensus items in the same order.

I think using fedimint transactions where possible (and where it makes sense) is probably better? That way all the regular affordances are there for you already, such as an idiomatic way to charge fees if you desire to prevent DOS-like attack vectors. Clients bypassing regular transactions to directly trigger consensus items is something I have actually come to see as a _bypass_ unless some async coordination needs to happen between guardians.

So to give an anecdotal example, initially the Stability Pool module had several consensus items: for the system to advance its cycles, as well as for clients to indicate interest in participating in different capacities. The former has to be done with a consensus item between the guardians since the trigger is time-based, but the latter could be accomplished just as easily with regular fedimint transactions initiated by the client. That simplifies the architecture quite a bit as we're leveraging the transaction-related tools already provided by the core code, instead of having to reinvent the wheel and worry about security, attacks, consensus, threshold etc. etc. from scratch on our own.

shaurya947: > A module developer should think carefully as to what's the bare minimum that needs to be part of consensus, and what could be deterministically calculated by each guardian independently,

To elaborate on this further w.r.t. to the Stability Pool, as long as the current time and current price of Bitcoin is part of the consensus item, _everything else_ can be calculated deterministically by each guardian: the exact contracts for each cycle, as well as even the index of the current cycle (because the ordering is guaranteed by the BFT algorithm). Drawing extensively on this example to show how much slimmer a module's contribution to consensus can be compared to what one might think initially. Really for the entire machine of the Stability Pool, all that guardians need to communicate is a time and a price every so often. Everything else can be deterministically calculated by each guardian: the long positions, the short positions, the payouts—everything!

elsirion:
Pinging @joschisan so we can maybe already disseminate the knowledge about the changes that AlephBFT will require. Afaik the only outstanding change is the removal of await_consensus_proposal.

dpc:
@EthnTuttle What do we do about it? Seems to be getting moldy. ;)

dpc:
Closing just because PR backlog keeps growing. Feel free to reopen if you think it should stay open.

Issue #2573, opened 2023-05-23
Make lightning gateway registration TTL relative


elsirion:
Currently it uses an absolute timestamp for timeouts, which requires well-synchronized clocks (especially questionable for clients). With relative TTLs akin to DNS this won't be a requirement anymore.

Probably best to start from here

https://github.com/fedimint/fedimint/blob/e6d3133609745761bed52acf0ebb959c1599d4fc/modules/fedimint-ln-common/src/lib.rs#L196-L197

changing the type to a TTL type. From there on the compiler will tell what else needs to be changed.

The TTL type would probably be a combination of an `Instant` to track time since entry was added and a `Duration` signifying the remaining lifetime when the entry was added.

When an entry is submitted via the network only the remaining lifetime (original lifetime - time since entry was created) is submitted and the recipient tracks the receive time themselves. Once `time_since_receive > remaining_lifetime_on_receipt` the entry times out and is removed.

a-moreira:
hey @elsirion, I'd like to tackle this task, if possible.
I tried implementing this using this struct (naming still temporary):
```
#[derive(Debug, Clone, Serialize, Deserialize, Hash, Decodable, Eq, PartialEq)]
pub struct GatewayLifetime {
    pub start_time: Instant,
    pub time_to_live: Duration,
}
```
then
```
...
 pub valid_until: GatewayLifetime, 
```

but first issue is that `Instant` is not `Serialize`able with serde. I'm gonna think of a workaround. What do you think?

elsirion:
Do you even need to serialize it? The thing that should be transmitted over the wire should always be the remaining TTL, i.e. `time_to_live - (now - start_time)`. So maybe a custom serde impl that only serializes this remaining TTL would work, making the inner fields a private implementation detail?

This assumes that the struct is only serialized for being sent over the wire and never saved for longer. If that isn't true we'd need [two different serde implementations for the parent struct (see `#[serde(with = "…")]`)](https://serde.rs/field-attrs.html#with).

If you ever need to actually serialize a time, look into `SystemTime`. [It's good to be aware of the difference between `Instant` and `SystemTime` anyway](https://doc.rust-lang.org/std/time/struct.SystemTime.html).

a-moreira:
This works for the Serde part:
```
use std::time::Instant;
use serde::{Deserializer, Serializer};

#[derive(Debug, Clone, Hash, PartialEq, Eq)]
pub struct GatewayValidity {
    start_time: Instant,
[...]

PR #3288, opened 2023-09-25
feat: Lightning Gateway Registration DB Garbage Collection

tvolk131:
Previously we've been indefinitely storing all Lightning Gateway registrations and filtering any expired registrations when calling `list_gateways()`. Now, we delete expired registrations on every write of a new registration. We're deleting on write instead of on read for two reasons:

1. To optimize for read speed (since reads will likely be much higher qps).
2. To avoid potential write conflicts during a read-only API call. If `delete_expired_gateways()` is called twice at the same time (where each call opens a separate dbtx), the call that commits second will fail since they're both attempting to delete the same keys. A read-only API call that sometimes returns a DB write error would be confusing.

### DIFF
--- a/modules/fedimint-ln-common/src/lib.rs
+++ b/modules/fedimint-ln-common/src/lib.rs
@@ -202,6 +202,12 @@ pub struct LightningGateway {
     pub gateway_id: secp256k1::PublicKey,
 }
 
+impl LightningGateway {
+    pub fn is_expired(&self) -> bool {
+        self.valid_until < fedimint_core::time::now()
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Encodable, Decodable, Serialize, Deserialize)]
 pub enum LightningConsensusItem {
     DecryptPreimage(ContractId, PreimageDecryptionShare),
--- a/modules/fedimint-ln-server/src/lib.rs
+++ b/modules/fedimint-ln-server/src/lib.rs
@@ -1021,7 +1021,7 @@ impl Lightning {
         stream
             .filter_map(|(_, gw)| async {
                 // FIXME: actually remove from DB
-                if gw.valid_until > fedimint_core::time::now() {
+                if !gw.is_expired() {
                     Some(gw)
                 } else {
                     None
--- a/modules/fedimint-ln-server/src/lib.rs
+++ b/modules/fedimint-ln-server/src/lib.rs
@@ -1020,7 +1020,6 @@ impl Lightning {
         let stream = dbtx.find_by_prefix(&LightningGatewayKeyPrefix).await;
         stream
             .filter_map(|(_, gw)| async {
-                // FIXME: actually remove from DB
                 if !gw.is_expired() {
                     Some(gw)
                 } else {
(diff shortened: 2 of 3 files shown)


modules/fedimint-ln-common/src/lib.rs
```diff
@@ -202,6 +202,12 @@ pub struct LightningGateway {
     pub gateway_id: secp256k1::PublicKey,
 }
 
+impl LightningGateway {
+    pub fn is_expired(&self) -> bool {
+        self.valid_until < fedimint_core::time::now()
```
m1sterc001guy, line 207: Not an issue for this PR, but until this https://github.com/fedimint/fedimint/issues/2573 is solved, this function could potentially return different results if called on the server or client side

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1036,9 +1035,31 @@ impl Lightning {
         dbtx: &mut ModuleDatabaseTransaction<'_>,
         gateway: LightningGateway,
     ) {
+        // Garbage collect expired gateways (since we're already writing to the DB)
```
m1sterc001guy, line 1038: Nit: This might make guardian registration fail more often, since now if two gateways try to register at the same time, they both might try to delete the same key and the database transaction will fail. I don't think we can do this anywhere else though. Clients are already expected to retry

m1sterc001guy (dismissed):
Nit: I was initially confused by the PR title, since the gateway also has a database. But this PR is actually cleaning up the gateway registrations on the server side.

Otherwise LGTM

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1036,9 +1035,31 @@ impl Lightning {
...
+        self.delete_expired_gateways(dbtx).await;
+
         dbtx.insert_entry(&LightningGatewayKey(gateway.node_pub_key), &gateway)
```
m1sterc001guy, line 1041: Let's make a issue for this. We shouldn't be storing gateways by their `node_pub_key`. It is possible that two gateways use the same lightning node.

modules/fedimint-ln-server/src/lib.rs
tvolk131: Someone in the call today mentioned the idea of adding a lock. What do you think about adding an in-memory deletion lock? If the lock is free, grab it and delete expired gateways. If the lock is taken, skip deleting expired gateways.

tvolk131:
> I was initially confused by the PR title, since the gateway also has a database. But this PR is actually cleaning up the gateway registrations on the server side

My bad! I renamed the PR to clarify what is being done.

modules/fedimint-ln-server/src/lib.rs
m1sterc001guy: IMO not worth it right now, we can add later if we see failures

modules/fedimint-ln-common/src/lib.rs
tvolk131: I have a draft PR out for that (#3289)

modules/fedimint-ln-server/src/lib.rs
tvolk131: Sounds good! I added a comment to the code here about this "gotcha".

dpc (approved):

PR #3289, opened 2023-09-25
feat: LightningGateway TTL

### LINKED ISSUE #2573: Make lightning gateway registration TTL relative

tvolk131:
Fixes #2573

Open question: Should all DB record values be of type `LightningGatewayRegistration` even if they don't currently use the `valid_until`?

Edit: I talked with @m1sterc001guy offline and the short answer is yes.


gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -471,35 +472,37 @@ impl GatewayClientModule {
...
-            node_pub_key: self.node_pub_key,
-            lightning_alias: self.lightning_alias.clone(),
-            api,
-            route_hints,
-            valid_until: fedimint_core::time::now() + time_to_live,
-            fees: self.fees,
-            gateway_id,
+    ) -> LightningGatewayRegistrationAnnouncement {
+        LightningGatewayRegistrationAnnouncement {
+            info: LightningGatewayInfo {
```
m1sterc001guy, line 480: Nit: `LightningGatewayInfo` -> `LightningGateway`

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -176,9 +176,63 @@ impl std::fmt::Display for LightningOutputOutcome {
...
+        self.valid_until < fedimint_core::time::now()
+    }
+}
+
+/// Information about a gateway that is shared with other federations and
+/// expires based on a TTL to allow for sharing between nodes with
+/// unsynchronized clocks which can each anchor the announcement to their local
+/// system time.
+#[derive(Debug, Clone, Serialize, Deserialize, Encodable, Decodable, PartialEq, Eq, Hash)]
+pub struct LightningGatewayRegistrationAnnouncement {
```
m1sterc001guy, line 214: Nit: `LightningGatewayRegistrationAnnouncement` -> `LightningGatewayAnnouncement`

m1sterc001guy:
> Open question: Should all DB record values be of type LightningGatewayRegistration even if they don't currently use the valid_until?

I think we answered this offline today, we decided that they should (we might use the ttl in the future)

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -214,12 +215,12 @@ impl LightningClientExt for Client {
         let (_lightning, instance) = self.get_first_module::<LightningClientModule>(&KIND);
         let mut dbtx = instance.db.begin_transaction().await;
         match dbtx.get_value(&LightningGatewayKey).await {
-            Some(active_gateway) => Ok(active_gateway),
+            Some(active_gateway) => Ok(active_gateway.info),
             None => self
                 .fetch_registered_gateways()
                 .await?
                 .into_iter()
-                .filter(|gw| !gw.is_expired())
```
m1sterc001guy, line 222: Not filtering on `is_expired` anymore?

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -176,7 +176,61 @@ impl std::fmt::Display for LightningOutputOutcome {
...
 }
 
-/// Information a gateway registers with a fed
+/// Information about a gateway that is stored locally and expires based on
+/// local system time
+#[derive(Debug, Clone, Serialize, Deserialize, Encodable, Decodable, PartialEq, Eq, Hash)]
+pub struct LightningGatewayRegistration {
+    pub info: LightningGateway,
+    /// Limits the validity of the announcement to allow updates
+    /// Anchored to local system time
```
m1sterc001guy, line 185: Nit: Anchored does need to be caps

m1sterc001guy:
Overall looks good to me but I think some of the failing tests need debugging

gateway/ln-gateway/src/state_machine/mod.rs
tvolk131: Done.

modules/fedimint-ln-common/src/lib.rs
tvolk131: Done.

modules/fedimint-ln-client/src/lib.rs
tvolk131: `fetch_registered_gateways()` now returns a list of `LightningGatewayAnnouncement`s which contain a `Duration`-based TTL, so checking for expired ones doesn't make sense since `Duration`s can't be negative.

modules/fedimint-ln-common/src/lib.rs
tvolk131: Done.

tvolk131:
> I think some of the failing tests need debugging

Fixed! The issue was that prior to this PR, `LightningGateway`s stored by each guardian were exactly the same since they each had identical `SystemTime`s for expiration, allowing a client to get all guardian's registered gateways and filter out duplicates by checking for exact equality. This invariant was broken by using `LightningGatewayAnnouncement` for sharing over the network, since the _unanchor -> transmit -> reanchor_ process allows for each guardian's re-anchored expiration time to be different. I added `filter_duplicate_gateways()` to address this.

modules/fedimint-ln-common/src/api.rs
```diff
@@ -127,3 +141,52 @@ where
...
+            });
+
+    // For each gateway, we may have multiple announcements with different settings
+    // and/or TTLs. We want to filter out duplicates in a way that doesn't allow a
+    // malicious guardian to override the caller's view of the gateways by
+    // returning a gateway with a shorter TTL. Instead, if we receive multiple
+    // announcements for the same gateway ID, we only filter out announcements
+    // that have the same settings, keeping the one with the longest TTL.
+    gateways_by_gateway_id
+        .into_values()
```
m1sterc001guy, line 169: A bit confused. Why do we create a hashmap above if we just throw away the keys and only use the values?

modules/fedimint-ln-common/src/api.rs
```diff
@@ -127,3 +141,52 @@ where
...
+                    .entry(gateway)
+                    .and_modify(|t| {
+                        if ttl > *t {
+                            *t = ttl;
+                        }
+                    })
+                    .or_insert(ttl);
+            }
+
+            gateways.into_iter().map(|a| LightningGatewayAnnouncement {
```
m1sterc001guy, line 186: Nit: use `announcement` instead of just `a`

modules/fedimint-ln-common/src/api.rs
```diff
@@ -127,3 +141,52 @@ where
...
+) -> Vec<LightningGatewayAnnouncement> {
+    let gateways_by_gateway_id =
+        gateways
+            .into_iter()
+            .fold(std::collections::HashMap::new(), |mut acc, gateway| {
+                acc.entry(gateway.info.gateway_id)
+                    .or_insert_with(Vec::new)
+                    .push(gateway);
+                acc
+            });
```
elsirion, line 160: Could be replaced with [`Itertools::into_group_map`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.into_group_map).

elsirion (approved):
This PR looks great! Very well documented and a good balance of functional and imperative style. Sorry that review took so long, too many conferences :see_no_evil:

elsirion:
Oh, needs a rebase :(

modules/fedimint-ln-common/src/api.rs
elsirion: The gateways are still grouped by the keys (values are `Vec`s of GWs with the same id), even though they aren't used directly.

modules/fedimint-ln-common/src/api.rs
tvolk131: Very cool! Done.

modules/fedimint-ln-common/src/api.rs
tvolk131: Done.

tvolk131:
> Oh, needs a rebase :(

Done. This should be ready to submit!

dpc (approved):

PR #3297, opened 2023-09-27
feat: gateway swap between federations

okjodom:
When a gateway is connected to any two federations, it can facilitate a direct e-cash swap between these federations, without routing over lightning.

## Specifics:

`GatewayPayInvoice` state machine checks if a proposed payment can be settled via direct swap. Specifically an LN payment that can be swapped will _have the gateway node pubkey_ and _a short channel id assigned by this gateway_ to one of it's connected federations in the route hint last hop. If these markers are present, the gateway instructs the client for the receiving federation to buy the preimage, and in turn, settles the payment in the originating federation client

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -456,6 +456,7 @@ fn states_add_instance(
     })
 }
 
+#[derive(Debug)]
 pub struct Client {
     inner: Arc<ClientInner>,
 }
--- a/gateway/ln-gateway/src/state_machine/mod.rs
+++ b/gateway/ln-gateway/src/state_machine/mod.rs
@@ -53,6 +53,7 @@ use crate::lnrpc_client::ILnRpcClient;
 use crate::state_machine::complete::{
     GatewayCompleteCommon, GatewayCompleteStates, WaitForPreimageState,
 };
+use crate::{FederationToClientMap, ScidToFederationMap};
 
 pub const GW_ANNOUNCEMENT_TTL: Duration = Duration::from_secs(600);
 pub const INITIAL_REGISTER_BACKOFF_DURATION: Duration = Duration::from_secs(15);
@@ -322,6 +323,8 @@ impl GatewayClientExt for Client {
 #[derive(Debug, Clone)]
 pub struct GatewayClientGen {
     pub lnrpc: Arc<dyn ILnRpcClient>,
+    pub all_clients: FederationToClientMap,
+    pub all_scids: ScidToFederationMap,
     pub node_pub_key: secp256k1::PublicKey,
     pub lightning_alias: String,
     pub timelock_delta: u64,
@@ -354,6 +357,8 @@ impl ClientModuleInit for GatewayClientGen {
     async fn init(&self, args: &ClientModuleInitArgs<Self>) -> anyhow::Result<Self::Module> {
         Ok(GatewayClientModule {
             lnrpc: self.lnrpc.clone(),
+            all_clients: self.all_clients.clone(),
+            all_scids: self.all_scids.clone(),
             cfg: args.cfg().clone(),
             notifier: args.notifier().clone(),
             redeem_key: args
@@ -373,6 +378,8 @@ impl ClientModuleInit for GatewayClientGen {
 #[derive(Debug, Clone)]
 pub struct GatewayClientContext {
     lnrpc: Arc<dyn ILnRpcClient>,
+    all_clients: FederationToClientMap,
+    all_scids: ScidToFederationMap,
     redeem_key: bitcoin::KeyPair,
     timelock_delta: u64,
     secp: secp256k1_zkp::Secp256k1<secp256k1_zkp::All>,
@@ -395,6 +402,8 @@ impl From<&GatewayClientContext> for LightningClientContext {
 pub struct GatewayClientModule {
     lnrpc: Arc<dyn ILnRpcClient>,
     cfg: LightningClientConfig,
+    all_clients: FederationToClientMap,
+    all_scids: ScidToFederationMap,
     pub notifier: ModuleNotifier<DynGlobalClientContext, GatewayClientStateMachines>,
     pub redeem_key: KeyPair,
     node_pub_key: PublicKey,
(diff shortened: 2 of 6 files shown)


fedimint-client/src/lib.rs
```diff
@@ -455,6 +455,7 @@ fn states_add_instance(
     })
 }
 
+#[derive(Debug)]
```
m1sterc001guy, line 458: Is this necessary in this PR? Seems unrelated

gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -132,7 +133,13 @@ pub trait GatewayClientExt {
     /// Attempt fulfill HTLC by buying preimage from the federation
     async fn gateway_handle_intercepted_htlc(&self, htlc: Htlc) -> anyhow::Result<OperationId>;
 
-    /// Subscribe to updates when the gateway is handling an intercepted HTLC
+    async fn gateway_handle_direct_swap(
```
m1sterc001guy, line 136: Nit: add comment on this function too

gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -627,3 +693,22 @@ impl TryFrom<InterceptHtlcRequest> for Htlc {
...
+    payment_hash: sha256::Hash,
+    amount_msat: Amount,
+}
+
+impl TryFrom<lightning_invoice::Bolt11Invoice> for SwapParameters {
+    type Error = anyhow::Error;
+
+    fn try_from(s: lightning_invoice::Bolt11Invoice) -> Result<Self, Self::Error> {
+        let payment_hash = *s.payment_hash();
+        let amount_msat = s.amount_milli_satoshis().map(Amount::from_msats).unwrap();
```
m1sterc001guy, line 708: Return this `Err` instead of unwrap

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -136,6 +140,11 @@ pub enum OutgoingPaymentError {
         error: OutgoingContractError,
         contract: OutgoingContractAccount,
     },
+    #[error("An error occurred while attempting direct swap.")]
```
m1sterc001guy, line 143: Nit: "direct swap between federations"

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -347,6 +435,26 @@ impl GatewayPayInvoice {
             invoice,
         })
     }
+
+    async fn check_swap_to_federation(
```
m1sterc001guy, line 439: Nit: lets add some comments here

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -242,20 +251,89 @@ impl GatewayPayInvoice {
...
+        let mut preimage_sub = client
+            .gateway_subscribe_ln_receive(operation_id)
+            .await
+            .map_err(|e| OutgoingPaymentError::SwapFailed {
+                contract: contract.clone(),
+                swap_error: format!("Failed to subscribe to direct swap: {}", e),
+            })?
+            .into_stream();
+
+        while let Some(state) = preimage_sub.next().await {
```
m1sterc001guy, line 281: I don't think we want to wait on the output of the receive state machine in the transition function. If we detect that this should be an internal swap, I think we should spawn the state machine, but transition to a new state that explicitly waits on the state of the child sm in the trigger function. We do this in `complete.rs` as well.

We don't want to block in the transition function ideally since we're holding a lock and no other state machines can make progress. The only reason we do `pay` for external lightning payments in the transition function is because `pay` is not idempotent.

m1sterc001guy (changes requested):
Mostly nits, but one issue regarding when we do the direct swap IMO

fedimint-client/src/lib.rs
okjodom: necessary because we reference `FederationToClientMap` within `GatewayClientGen` which is debug

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -962,6 +962,77 @@ async fn test_gateway_shows_balance_for_any_connected_federation() -> anyhow::Re
...
+                    assert_matches!(waiting_funds, LnReceiveState::AwaitingFunds { .. });
+                    let claimed = receive_sub.ok().await?;
+                    assert_matches!(claimed, LnReceiveState::Claimed);
+                    assert_eq!(client2.get_balance().await, msats(2_500));
+                }
+                _ => panic!("Expected Lightning payment!"),
+            }
+
+            // Check gateway balances after facilitating direct swap between federations
+            let post_balances = get_balances(&rpc, &[id1, id2]).await;
```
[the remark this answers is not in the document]
okjodom: added check below

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -242,21 +264,75 @@ impl GatewayPayInvoice {
...
+                                        contract,
+                                        federation_id,
+                                        operation_id,
+                                    },
+                                )),
+                            };
+                        }
+                        Err(e) => {
+                            tracing::debug!("Failed to buy preimage via direct swap: {}", e)
+                            // Fall back to lightning payments
```
m1sterc001guy, line 324: Hmm do we want to fall back here? If the node in the route hint matches the gateway node, that means that the lightning node is always going to try and do a self payment, which we know won't work. Maybe we should just fail?

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -409,6 +528,105 @@ impl GatewayPayClaimOutgoingContract {
...
+pub struct GatewayPayWaitForSwapPreimage {
+    contract: OutgoingContractAccount,
+    federation_id: FederationId,
+    operation_id: OperationId,
+}
+
+impl GatewayPayWaitForSwapPreimage {
+    fn transitions(
+        &self,
+        _global_context: DynGlobalClientContext,
```
m1sterc001guy, line 541: You can just remove this parameter

gateway/ln-gateway/src/state_machine/pay.rs
okjodom: I was undecided on the desired behavior. I agree failing here makes more sense

m1sterc001guy (approved):

m1sterc001guy (approved):

Issue #2955, opened 2023-08-14
Gateway will use the default rpc client


douglaz:
It seems the gateway will try to use the bitcoind defined on the wallet module consensus configuration, like:
```json
    "2": {
      "kind": "wallet",
      "client_default_bitcoin_rpc": {
        "kind": "esplora",
        "url": "http://127.0.0.1:50002/"
      },
```
Ignoring all local configuration like FM_BITCOIN_* variables

(Perhaps this problem isn't gateway specific and will happen with any client?)

elsirion:
Yes, the client doesn't take env vars into account. The problem is that there is a wide variety of clients and only a subset of them supports env vars, so it didn't occur to me.

I think it would be interesting to support as long as there is no better way of overriding this default (which will likely be stateful and more complex).

justinmoon:
Gatewayd could read from env [here](https://github.com/fedimint/fedimint/blob/2bb733ad6ac818f89cb2f00f352c0ce9896997ed/gateway/ln-gateway/src/lib.rs#L185-L190) and set the client bitcoin rpc inside `WalletClientGen` instead of just using `WalletClientGen::default()`. This seems cleaner than having client library ready from environment.

m1sterc001guy:
This seems like a configuration that would be good to add to the gateway's config API: https://github.com/fedimint/fedimint/pull/3154

I think most gateway operators will only want to configure their bitcoind instance once, and not per-client.

m1sterc001guy:
This is blocked on updating `bitcoincoire-rpc` to 0.17.0. The client can't currently use bitcoind for RPCs because it calls `import_address_script`, which throws an error when called on descriptor wallets. I believe this needs to be upgraded to `import_descriptors` which is available in `bitcoincore-rpc` 0.17.0

elsirion:
https://github.com/fedimint/fedimint/issues/7235 should address this too.

PR #2667, opened 2023-06-19
wip: update to bitcoin 0.30.0

dpc:
Currently blocked on 

- [ ] https://github.com/ElementsProject/lightning/issues/6349
- [x] https://github.com/bitcoindevkit/rust-esplora-client/pull/51
- [ ] possibly something else?

PR #3301, opened 2023-09-27
chore(wallet-client): use btc rpc client from env var

m1sterc001guy:
Partially addresses: https://github.com/fedimint/fedimint/issues/2955

This allows the bitcoin rpc method to be overridden using environment variables. 

Set `FM_BITCOIN_RPC_KIND` to `esplora`, or `electrum` (`bitcoind` is not supported yet, see below)
Set `FM_BITCOIN_RPC_URL` to the URL of the esplora or electrum server.

`bitcoind` is not supported yet because we need to update our `bitcoin`, `bitcoin_hashes`, `secp*` crates. We are currently blocked on some downstream dependencies not supporting `bitcoin 0.30.0` yet. https://github.com/fedimint/fedimint/pull/2667

### DIFF
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -468,7 +468,7 @@ impl ClientModuleInit for WalletClientGen {
         let rpc_config = self
             .0
             .clone()
-            .unwrap_or(args.cfg().default_bitcoin_rpc.clone());
+            .unwrap_or(WalletClientModule::get_rpc_config(args.cfg()));
         Ok(WalletClientModule {
             cfg: args.cfg().clone(),
             module_root_secret: args.module_root_secret().clone(),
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -557,6 +557,20 @@ pub struct WalletClientContext {
...
 impl WalletClientModule {
+    fn get_rpc_config(cfg: &WalletClientConfig) -> BitcoinRpcConfig {
+        if let Ok(rpc_config) = BitcoinRpcConfig::from_env_vars() {
+            // TODO: Wallet client cannot support bitcoind RPC until the bitcoin dep is
+            // updated to 0.30
+            if rpc_config.kind != "bitcoind" {
+                rpc_config
+            } else {
+                cfg.default_bitcoin_rpc.clone()
+            }
```
elsirion, line 568: Not sure we should do this, can't we just fail and tell the user that what they intended to do didn't work?

m1sterc001guy: Yes, unfortunately it breaks our developer environment since we use the same environment variable on the server side, which is set to `bitcoind` in devimint. I could use different environment variables for the client?

elsirion: never mind, we'll hopefully be upgrading bitcoin soon anyway

PR #3318, opened 2023-10-02
refactor: move incoming state machine into lightning client crate

m1sterc001guy:
`ln-common` is for sharing code across client/server boundary. `Incoming` state machine is never used on the server side, it is just shared between the client and the gateway. The gateway already depends on the client, so it is fine to just move the state machine into `ln-client`


okjodom (approved):
lgtm

PR #3319, opened 2023-10-02
refactor: rename fetch_contract -> wait_contract

m1sterc001guy:
`fetch_contract` is currently confusingly named, since it actually blocks and waits for a contract to exist. It is useful to have both `wait_contract` and `fetch_contract` since we also will want to validate when a contract does not exist.

### DIFF
--- a/modules/fedimint-ln-common/src/api.rs
+++ b/modules/fedimint-ln-common/src/api.rs
@@ -1,8 +1,8 @@
 use bitcoin_hashes::sha256::Hash as Sha256Hash;
 use fedimint_core::api::{FederationApiExt, FederationResult, IModuleFederationApi};
 use fedimint_core::endpoint_constants::{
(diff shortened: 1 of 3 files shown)


dpc (approved):

modules/fedimint-ln-common/src/api.rs
```diff
@@ -51,6 +52,14 @@ where
     }
 
     async fn fetch_contract(&self, contract: ContractId) -> FederationResult<ContractAccount> {
```
okjodom, line 54: are we keeping the old api for back compat? should we somehow mark as deprecated?

okjodom:
sorry @m1sterc001guy , just spotted one question after you added this to the merge queue

modules/fedimint-ln-common/src/api.rs
m1sterc001guy: Old API changed to `wait_contract`. I added `fetch_contract`, but it currently isn't used anywhere. But it will be in https://github.com/fedimint/fedimint/pull/3322

okjodom: ack. looks good

Issue #3294, opened 2023-09-26
Remove polling from lightning client


justinmoon:
I count 4 places where we poll but could probably switch to a websocket subscription:

* https://github.com/fedimint/fedimint/blob/769bd72e5ca34b770f2b37ebe3f4f3d694886930/modules/fedimint-ln-client/src/pay.rs#L386
* https://github.com/fedimint/fedimint/blob/769bd72e5ca34b770f2b37ebe3f4f3d694886930/modules/fedimint-ln-client/src/pay.rs#L403
* https://github.com/fedimint/fedimint/blob/769bd72e5ca34b770f2b37ebe3f4f3d694886930/modules/fedimint-ln-client/src/receive.rs#L195
* https://github.com/fedimint/fedimint/blob/769bd72e5ca34b770f2b37ebe3f4f3d694886930/modules/fedimint-ln-common/src/incoming.rs#L227

PR #3328, opened 2023-10-04
fix: Remove polling from Lightning Client

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3294

We still poll if the API request returns a `FederationError` (likely meaning the federation is offline)


modules/fedimint-ln-client/src/incoming.rs
```diff
@@ -225,32 +225,24 @@ impl DecryptingPreimageState {
         global_context: DynGlobalClientContext,
         contract_id: ContractId,
     ) -> Result<Preimage, IncomingSmError> {
-        // TODO: Get rid of polling
-        let preimage = loop {
-            let contract = global_context
+        loop {
```
okjodom, line 228: Thought it's this loop we need to rid of? Same comment for the other polls below

m1sterc001guy: So, theses loops are harder to get rid of, because the API call can still return a `FederationError` (i.e the federation is offline). If the federation is offline, we don't necessarily want to transition to a failed state, because the federation could come back later and the decryption could be successful. This means that we do want to retry here until we get a clear signal if the decryption was successful.

The improvement here is that we don't poll unless we receive a `FederationError`. `wait_preimage_decryption` will block until there is a clear failure or if the preimage is successfully decrypted. Previously we were polling even if the decryption was in progress.

Same goes for the other API calls.

okjodom: ack.
Pinging @justinmoon, @dpc for further reviews

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -383,38 +383,34 @@ impl LightningPayRefundable {
...
-                .await;
-            if let Ok(contract) = contract {
-                if contract.contract.cancelled {
-                    return;
-                }
+                .wait_outgoing_contract_cancelled(contract_id)
+                .await
+                .is_ok()
+            {
+                return;
```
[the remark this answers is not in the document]
m1sterc001guy: Added

elsirion (approved):

PR #3339, opened 2023-10-09
feat: return gateway fee to client when paying lightning invoice

m1sterc001guy:


okjodom (approved):

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -258,7 +258,7 @@ impl LightningClientExt for Client {
     async fn pay_bolt11_invoice(
         &self,
         invoice: Bolt11Invoice,
-    ) -> anyhow::Result<(PayType, ContractId)> {
+    ) -> anyhow::Result<(PayType, ContractId, Amount)> {
```
elsirion, line 261: TODO: get a quote first and then use that quote to make the payment.

This is how withdraws work rn: the user asks the federation how much fees they'll need to pay and then uses that to create the actual withdraw tx.

This has two benefits:
1. The UI can show how much the user will be paying
2. Later we could upgrade our fee estimation by asking the GW for a guess for that particular destination and then using that instead of a static %age as we do rn.

elsirion (approved):

Issue #3345, opened 2023-10-10
AlephBFT panic


dpc:
`just mprocs` fails for me:

```
        assert!(bytes.len() <= BYTE_LIMIT);
```

```
thread 'tokio-runtime-worker' panicked at 'assertion failed: bytes.len() <= BYTE_LIMIT', fedimint-server/src/atomic_broadcast/data_provider.rs:105:9
stack backtrace:
```

I "fixed" it with:

```
> git diff HEAD
diff --git a/fedimint-server/src/atomic_broadcast/data_provider.rs b/fedimint-server/src/atomic_broadcast/data_provider.rs
index 29b1a167b7..27008ff086 100644
--- a/fedimint-server/src/atomic_broadcast/data_provider.rs
+++ b/fedimint-server/src/atomic_broadcast/data_provider.rs
@@ -9,7 +9,7 @@ use tokio::sync::watch;
 use crate::LOG_CONSENSUS;
 
 // This limits the RAM consumption of a Unit to roughly 10kB
-const BYTE_LIMIT: usize = 10_000;
+const BYTE_LIMIT: usize = 100_000;
 
 #[derive(
     Clone, Debug, PartialEq, Eq, Hash, parity_scale_codec::Encode, parity_scale_codec::Decode,
```

but it suggests some deeper issues that can result in a DoS.

joschisan:
So the issue is that line 91 in data provider needs to be copied and pasted between 73 and 74. On the same topic, what is the maximum number of bytes of  a u64 integer with bigsize? More then 8 bytes? Then line 63 needs to be updated accordingly.

joschisan:
What is happening here is that we want to create a vector of ConsensusItems that encodes within at most BYTE_LIMIT bytes. This is our submission every single round so every 250ms. We assert that we have bound the length of the encoding correctly which I obviously did not do.

dpc:
> On the same topic, what is the maximum number of bytes of a u64 integer with bigsize

```
			_ => {
				0xFFu8.write(writer)?;
				(self.0 as u64).write(writer)
			},
```

9 bytes.

PR #3350, opened 2023-10-11
fix: calculate the maximum length of the encoding correctly

joschisan:
Should fix https://github.com/fedimint/fedimint/issues/3345

### DIFF
--- a/fedimint-server/src/atomic_broadcast/data_provider.rs
+++ b/fedimint-server/src/atomic_broadcast/data_provider.rs
@@ -59,8 +59,8 @@ impl aleph_bft::DataProvider<UnitData> for DataProvider {
             return Some(UnitData::Signature(signature));
         }
 
-        // an empty vector is encoded in at most 8 bytes
-        let mut n_bytes = 8;
+        // the length of a vector is encoded in at most 9 bytes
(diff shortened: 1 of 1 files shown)


fedimint-server/src/atomic_broadcast/data_provider.rs
```diff
@@ -59,8 +59,8 @@ impl aleph_bft::DataProvider<UnitData> for DataProvider {
             return Some(UnitData::Signature(signature));
         }
 
-        // an empty vector is encoded in at most 8 bytes
-        let mut n_bytes = 8;
+        // the length of a vector is encoded in at most 9 bytes
+        let mut n_bytes = 9;
```
elsirion, line 63: This probably didn't cause the problem, right, but rather the other places in which the size wasn't added up correctly?

elsirion (approved):

fedimint-server/src/atomic_broadcast/data_provider.rs
joschisan: I assume so

dpc:
Before I was able to reliably hit the issue on every `just mprocs`, now I tried and it seems to work so LGTM.

PR #3351, opened 2023-10-11
fix: only propose new fee rate when our vote changes

### LINKED ISSUE #3337: Wallet module continuously proposes fee level
Even if the local bitcoin fee rate of a guardian did not change they will keep proposing a fee rate if theirs differs from the consensus fee rate.

elsirion:
Fixes #3337

### DIFF
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -342,10 +342,14 @@ impl ServerModule for Wallet {
             items.push(WalletConsensusItem::BlockCount(block_count_proposal));
         }
 
+        let current_fee_rate_vote = dbtx
+            .get_value(&FeeRateVoteKey(self.cfg.local.our_peer_id))
+            .await
+            .unwrap_or(self.cfg.consensus.default_fee);
         // TODO: We should not be panicking
         let fee_rate_proposal = self.get_fee_rate().await.expect("bitcoind rpc failed");
 
-        if fee_rate_proposal != self.consensus_fee_rate(dbtx).await {
+        if fee_rate_proposal != current_fee_rate_vote {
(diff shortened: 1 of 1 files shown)


dpc (approved):

fedimint-backports:

Please cherry-pick the changes locally.
```bash
git worktree add -d .worktree/backport-3351-to-releases/v0.1 origin/releases/v0.1
cd .worktree/backport-3351-to-releases/v0.1
git checkout -b backport-3351-to-releases/v0.1
ancref=$(git merge-base a8c6dddede7ca4fe600286536eb228311f42d416 1572a437e2db0aff01624f249ef758945aa9b6e6)
git cherry-pick -x $ancref..1572a437e2db0aff01624f249ef758945aa9b6e6
```

elsirion:
Backporting this will be trickier than expected, the problem is that we don't have `our_peer_id` in the module config on 0.1 since it was introduced in #3205 after branching off (similar fix like this one).

The least-breaking way I can imagine backporting this now:

I'll do a refactor on master first that removes `our_peer_id` from config and introduces it in the signature of `ServerModuleInit::init` instead. This can then be manually backported since it only slightly changes the module interface but no on-disk data that was generated by old federation setups. Based on that I can backport both #3351 and #3205.

If we would have had #3255 on 0.1 this would not be a breaking change at all.

Thoughts @dpc?

dpc:
Maybe we should backport https://github.com/fedimint/fedimint/pull/3255 on @elsirion . It breaks the Rust API, but our 0.1 is more about the external compatibility, less about external modules?

BTW. That's a good illustration that we need different version for Rust API, and different for "general use" (or whatever one would call it).

elsirion:
If it's not too much work it might be worth also backporting #3255, will certainly make the others apply more cleanly.

Issue #3356, opened 2023-10-11
Backport some rust-API-breaking commits to v0.1


dpc:
https://github.com/fedimint/fedimint/pull/3351#issuecomment-1758435425

* [x] backport https://github.com/fedimint/fedimint/pull/3254
* [x] backport https://github.com/fedimint/fedimint/pull/3255
* [x] backport https://github.com/fedimint/fedimint/pull/3205
* [x] backport https://github.com/fedimint/fedimint/pull/3351

elsirion:
Finished in  #3373

PR #3359, opened 2023-10-11
fix: wallet module spaming epochs (don't land yet)

dpc:
Backport #3205 

Context #3356

On top of #3255

See commit message for original commit messages.

### DIFF
--- a/modules/fedimint-wallet-common/src/config.rs
+++ b/modules/fedimint-wallet-common/src/config.rs
@@ -61,6 +61,8 @@ pub struct WalletConfig {
 pub struct WalletConfigLocal {
     /// Configures which bitcoin RPC to use
     pub bitcoin_rpc: BitcoinRpcConfig,
+    /// Our own peer_id
+    pub our_peer_id: PeerId,
 }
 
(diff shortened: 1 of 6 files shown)


modules/fedimint-wallet-common/src/config.rs
```diff
@@ -151,7 +155,10 @@ impl WalletConfig {
         );
 
         Self {
-            local: WalletConfigLocal { bitcoin_rpc },
+            local: WalletConfigLocal {
+                bitcoin_rpc,
```
dpc, line 159: **question (blocking):** @elsirion it's unclear to me if this is OK. Are we sure we are not storing `WalletConfigLocal` anywhere?

Maybe `out_peer_id` should be stored in a separate field so we avoid storing it twice (and breaking compatibility)? If so, maybe we should do it on `master` too?

elsirion: See  #3367

dpc:
This is replaced by https://github.com/fedimint/fedimint/pull/3373

PR #3360, opened 2023-10-12
feat: Add snake-casing for all the serde serialized structs [Fixes #3332]

### LINKED ISSUE #3332: JSON-serialization of fedimint-client states in snake_case
When client operation log is exposed to something (e.g. via gateway HTTP API) we will have CapitalCase keys and values as you can see in the screenshot below. [With serde we can annotate these structs to make them serialize as snake_case ](https://serde.rs/attr-rename.html).

image

EshaanAgg:
Fixes #3332 

I have added the `#[serde(rename_all(serialize = "snake_case"))]` attribute to all the structs and enums  in the `fedimint-cli` package, even if they have all their attributes in snake_case (to ensure that if in future if some attributes are added to them, they do not break this stylistic guideline).

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -374,6 +374,7 @@ pub async fn handle_command(
         }
         ClientCmd::ListOperations { limit } => {
             #[derive(Serialize)]
+            #[serde(rename_all(serialize = "snake_case"))]
             struct OperationOutput {
                 id: OperationId,
                 creation_time: String,
@@ -491,6 +492,7 @@ async fn get_note_summary(client: &Client) -> anyhow::Result<serde_json::Value> 
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all(serialize = "snake_case"))]
 struct InfoResponse {
     federation_id: FederationId,
     network: Network,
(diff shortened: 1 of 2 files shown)


dpc (approved):

dpc:
Nice. Thanks.

EshaanAgg:
@dpc The PR seems to be removed from the merge-queue. Is there anything I can do to help fix this?

dpc:
Re-merged. Looks like something weird happened with one of the CI runners?

elsirion:
Does this fix the problem in the issue mentioned? The only annotation that does anything here is the one on the error enum. Structs are snake case already. I think the original issue was about the operation log meta enums, e.g.

https://github.com/fedimint/fedimint/blob/9516c6c9ac6077f1230c7368c38e793127d8a20b/modules/fedimint-ln-client/src/lib.rs#L557-L570

EshaanAgg:
Hey @elsirion! Thanks for pointing this out. As I'm getting started with fedimint, it's difficult for me to keep track of how the huge codebase works. I'll try to look into the files pointed out by you and open a corrective PR. 
Sorry for the oversight and thanks for the help!

justinmoon:
dev call: can we just make this a global default?

PR #3371, opened 2023-10-12
Add snake_case attribute to fedimint-In-client for JSON serilaization [Fixes @3332]

EshaanAgg:
Hopefully fixes @3332

As pointed out by @elsirion, the merged PR #3360 did not completely solve the scope of the issue. I have added the rename to snake_case attribute to all the relevant enums and structs in the `fedimint-In-client` module. 

As I am pretty new to `fedimint`, I would greatly appreciate it if you could help me with these queries along with reviewing this PR:
- Are there any commands I can run to verify the changed nature of the JSON output as shown in the original issue description?
- Is there any documentation of which libraries/modules perform what? This would help me understand the codebase better. 

Thanks!

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -374,7 +374,7 @@ pub async fn handle_command(
         }
         ClientCmd::ListOperations { limit } => {
             #[derive(Serialize)]
-            #[serde(rename_all(serialize = "snake_case"))]
+            #[serde(rename_all = "snake_case")]
             struct OperationOutput {
                 id: OperationId,
                 creation_time: String,
@@ -492,7 +492,7 @@ async fn get_note_summary(client: &Client) -> anyhow::Result<serde_json::Value> 
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(rename_all(serialize = "snake_case"))]
+#[serde(rename_all = "snake_case")]
 struct InfoResponse {
     federation_id: FederationId,
     network: Network,
@@ -513,7 +513,7 @@ pub fn parse_fedimint_amount(s: &str) -> Result<fedimint_core::Amount, ParseAmou
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(rename_all(serialize = "snake_case"))]
+#[serde(rename_all = "snake_case")]
 struct PayInvoiceResponse {
     operation_id: OperationId,
     contract_id: ContractId,
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -46,7 +46,7 @@ use crate::client::ClientCmd;
 
 /// Type of output the cli produces
 #[derive(Serialize)]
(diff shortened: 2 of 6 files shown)


EshaanAgg:
Update: The tests fail due to a JSON deserialization error of the `Pay` enum type. I can't figure out where the error is. I would appreciate any help if possible.

dpc:
Probably somewhere in the tests we have a hardcoded a string with old encoding?

EshaanAgg:
Even I thought so but couldn't find anything similar

dpc:
I was looking at it yesterday and also couldn't find anything.

Please stuff the code surrounding the panic with `dbg!`s to capture the exact serialized text and see if you can use it to find a source that create it.. Run tests locally for faster iteration time.

It could be:

* some serde bug/misfeature that does not handle enum variant name renaming in certain case
* somewhere we share this definition of struct redundantly and also need an update
* somewhere we have a fixture that is using (could it be a schema of an external server even?)

dpc:
This is easily repro. Right.

```
> cargo test -- makes_internal_payments_within_federation
...
[fedimint-client/src/oplog.rs:193] self.meta.clone() = Object {
    "receive": Object {
        "out_point": Object {
            "txid": String("ddd3d73e0e213d2eae398ad033c48fe8d90709e66a993d00af1c5956f8debc99"),
            "out_idx": Number(0),
        },
        "invoice": String("lnbcrt2500n1pjjj7yhdq5wa5hg6pdd4shy6m9wfespp5ypfv4m6nswfur947nk6sc3ajns20m9ujueau7ax27gdfajsjhxwssp5nm9zu7ynr04g7ucplkg067587h0pp0f5rnqrcvyzt7j86e3unnlq9qrsgqcqpjnp4qvte3dkgsa645usn4xvl3ftu8uf3k6yhk0kdflwu9j397lacvjvawxqrrssrzjqv4d6v9k6f805ftm47vkrj0esk4kern8z7hr05s2mmx7u3d4v5gjvqqqqqqqqqqqqqqqqqqqqqqqqqqqrcs94h25mf96ev9ynn59sxxz3egl2h3cm5dek88lwxv605hu0kugtnn2c496htxvuduw7dhlrwv8zcqmus0ewskc6vj6cu88ta0gfp2fsqvkh5he"),
        "extra_meta": Null,
    },
}
```

So it's not even serialization. It's the deserialization itself.

dpc:
Ooooooohhhhh... :rofl: 

```
#[serde(rename_all(serialize = "snake_case"))]
```

Well, of course, if you rename it only during serialization, then no wonder it fails at deserialization. @EshaanAgg

EshaanAgg:
> Ooooooohhhhh... 🤣
> 
> ```
> #[serde(rename_all(serialize = "snake_case"))]
> ```
> 
> Well, of course, if you rename it only during serialization, then no wonder it fails at deserialization. @EshaanAgg
> 
> IIRC just `#[serde(rename_all = "snake_case")]` works is what you want.

Ahhh!!! I feel so stupid. Thanks for catching it. I will make the changes.

dpc (approved):

PR #3386, opened 2023-10-12
feat: `IClientConfig::to_json`

dpc:
For client side functionality it's useful to be able to a validate client module confing into a serializable form in a generic way.


elsirion (changes requested):
This is creating redundancy with existing functionality. Your approach is the better one imo, but could you please remove the old code?

https://github.com/fedimint/fedimint/blob/5f533fefc82355a42ac0b0c3c8cdc6da47700754/fedimint-client/src/module/mod.rs#L275-L276

justinmoon:
superseded by https://github.com/fedimint/fedimint/pull/3394

PR #3389, opened 2023-10-13
Gateway registrations are stored on servers by `gateway_id`

### LINKED ISSUE #3290: Gateway registrations should use `gateway_id` instead of `node_pub_key` on server side
See https://github.com/fedimint/fedimint/pull/3288/files#r1336124666

tvolk131:
Fixes #3290

### DIFF
--- a/modules/fedimint-ln-server/src/lib.rs
+++ b/modules/fedimint-ln-server/src/lib.rs
@@ -9,7 +9,9 @@ use fedimint_core::config::{
     TypedServerModuleConfig, TypedServerModuleConsensusConfig,
 };
 use fedimint_core::core::ModuleInstanceId;
-use fedimint_core::db::{DatabaseVersion, ModuleDatabaseTransaction};
+use fedimint_core::db::{
+    DatabaseTransaction, DatabaseVersion, MigrationMap, ModuleDatabaseTransaction,
+};
 use fedimint_core::encoding::Encodable;
 use fedimint_core::endpoint_constants::{
     ACCOUNT_ENDPOINT, BLOCK_COUNT_ENDPOINT, LIST_GATEWAYS_ENDPOINT, OFFER_ENDPOINT,
@@ -55,7 +57,7 @@ use fedimint_metrics::{
     Histogram, IntCounter,
 };
 use fedimint_server::config::distributedgen::PeerHandleOps;
-use futures::StreamExt;
+use futures::{FutureExt, StreamExt};
 use rand::rngs::OsRng;
 use strum::IntoEnumIterator;
 use tracing::{debug, error, info_span, trace};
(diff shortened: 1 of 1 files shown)


m1sterc001guy:
I know we discuss on the dev call awhile back that this might be a good candidate to do a database migration for. Any thoughts?

m1sterc001guy:
Some tests likely need to change as well

tvolk131:
> I know we discuss on the dev call awhile back that this might be a good candidate to do a database migration for. Any thoughts?

Ahh yes, I forgot about that! Thanks for the reminder, I'll give that a shot.

tvolk131:
This is ready for review, except I'm not sure what strategy we want to use to test the migration. The V0 data uses the same pubkey for `node_pub_key` and `gateway_id`:

https://github.com/fedimint/fedimint/blob/b76f0a9459e6387f1be1f7dec3115e7a285c9493/modules/fedimint-ln-server/src/lib.rs#L1435-L1439

...

https://github.com/fedimint/fedimint/blob/b76f0a9459e6387f1be1f7dec3115e7a285c9493/modules/fedimint-ln-server/src/lib.rs#L1512-L1528

So asserting that gateways are keyed by `gateway_id` in the migration test passes even if the migration doesn't run. I think it would be worth rewriting `create_db_with_v0_data()` to generate more realistic data. Does this seem reasonable?

m1sterc001guy:
> So asserting that gateways are keyed by gateway_id in the migration test passes even if the migration doesn't run. I think it would be worth rewriting create_db_with_v0_data() to generate more realistic data. Does this seem reasonable?

When adding a migration, you should create a new test, called `create_db_with_v1_data` that tests the migration from a "v1 backup", where the V1 backup contains a db backup of the new data.

That said, this is a bit weird, since the database change here is actually backwards compatible, since they're both the same kind of key. Usually the database change would cause the v0 db backup to be invalid, which is not the case for this change. So I would agree that we should change the v0 test to use different keys for `node_pub_key` and `gateway_id`.

elsirion:
Do you want to backport to 0.1? Otherwise don't worry about DB migrations.

elsirion:
This PR currently only consists of DB migration code it seems, what's the status here?

* Is the main functionality still being worked on?
* Do we want migrations? This is only needed if we intend to backport to the `v0.1` branch

tvolk131:
I think we decided we don't want to backport, so we're going to skip the migration. Apologies for the delay on this one - I've been busy and simply haven't gotten around to removing the migration.

m1sterc001guy:
@tvolk131 What is the status of this?

elsirion:
Since we stabilized Fedimint since then we'll need migrations now :see_no_evil:

tvolk131:
> @tvolk131 What is the status of this?

No update, haven't touched this in a while but I can circle back to it this week!

> Since we stabilized Fedimint since then we'll need migrations now 🙈

Ope! Sorry I didn't get around to this sooner! but as Justin mentioned, could be a great test for a simple migration.

m1sterc001guy:
Fixed in https://github.com/fedimint/fedimint/pull/4114

PR #3401, opened 2023-10-15
Server mint consensus config stores denomination base instead of all denominations

### LINKED ISSUE #3383: Simplify E-Cash Mint Server Module Consensus Settings
The consensus settings for this module contains a field `mint_amounts` which allows for arbitrary denomination amounts to be issued:

https://github.com/fedimint/fedimint/blob/9d871a719056415503d92e7d73e05c53d4b145f0/modules/fedimint-mint-common/src/config.rs#L20-L22

Currently we only ever use base-2 amounts, and we do this by generating the amounts and setting `mint_amounts` to the generated values:

https://github.com/fedimint/fedimint/blob/9d871a719056415503d92e7d73e05c53d4b145f0/fedimintd/src/lib.rs#L55-L60

https://github.com/fedimint/fedimint/blob/9d871a719056415503d92e7d73e05c53d4b145f0/modules/fedimint-mint-common/src/config.rs#L29-L34

https://github.com/fedimint/fedimint/blob/9d871a719056415503d92e7d73e05c53d4b145f0/modules/fedimint-mint-server/src/lib.rs#L789-L791

In the first two examples above, we can also see that the values are being generated using a specified maximum denomination value.

We want to retain the ability to use other bases for denominations, but we don't need to be able to issue arbitrary note amounts with no shared denomination base. Let's replace the `mint_amounts` field with a field `denomination_base` which would be used to generate the actual denomination amounts. Let's also put a cap on denomination size at 1 million coins. We can hardcode this value and remove the existing parameter since there isn't much to be gained in tweaking it. Here is the existing parameter:

https://github.com/fedimint/fedimint/blob/662799742c6a13c3342b0b3655bf1d82f6ca1e02/fedimintd/src/fedimintd.rs#L66-L69

Let's add a method to `MintGenParamsConsensus` to generate the denomination amounts (i.e. consolidate the calls to `Tiered::gen_denominations()` in the links above).

_Originally discussed in https://github.com/fedimint/fedimint/discussions/3327_

tvolk131:
Fixes #3383

### DIFF
--- a/fedimint-core/src/lib.rs
+++ b/fedimint-core/src/lib.rs
@@ -109,12 +109,16 @@ pub struct Amount {
 impl Amount {
     pub const ZERO: Self = Self { msats: 0 };
 
-    pub const fn from_msats(msat: u64) -> Amount {
-        Amount { msats: msat }
+    pub const fn from_msats(msats: u64) -> Amount {
+        Amount { msats }
     }
 
-    pub const fn from_sats(sat: u64) -> Amount {
-        Amount { msats: sat * 1000 }
+    pub const fn from_sats(sats: u64) -> Amount {
+        Amount::from_msats(sats * 1000)
+    }
+
+    pub const fn from_coins(coins: u64) -> Amount {
+        Amount::from_sats(coins * 100_000_000)
     }
 
     pub fn from_str_in(s: &str, denom: Denomination) -> Result<Amount, ParseAmountError> {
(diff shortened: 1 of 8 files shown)


joschisan (approved):

fedimint-core/src/lib.rs
```diff
@@ -109,12 +109,16 @@ pub struct Amount {
...
+        Amount { msats }
     }
 
-    pub const fn from_sats(sat: u64) -> Amount {
-        Amount { msats: sat * 1000 }
+    pub const fn from_sats(sats: u64) -> Amount {
+        Amount::from_msats(sats * 1000)
+    }
+
+    pub const fn from_coins(coins: u64) -> Amount {
```
dpc, line 120: **suggestion:** No idea what "from_coins" supposed to mean, and only after looks at the impl I see it's "from_btc"

tvolk131: I can throw together a PR that replaces this line:

https://github.com/fedimint/fedimint/blob/479bebe32648841c203993eed925462caebd65ce/modules/fedimint-mint-common/src/config.rs#L27

with

`const MAX_DENOMINATION_SIZE: Amount = Amount::from_sats(1_000_000 * 100_000_000);`

and remove `from_coins()`

PR #3403, opened 2023-10-16
refactor: remove AtomicBroadcast and FedimintConsensus

joschisan:
The AtomicBroadcast and FedimintConsensus struct were previously fields of ConsensusServer. All three struct shared fields and had overlapping functionality which is why they were merged.


fedimint-server/src/atomic_broadcast/mod.rs
```diff
@@ -112,3 +108,13 @@ pub enum Recipient {
...
+
+pub fn to_peer_id(node_index: NodeIndex) -> PeerId {
+    u16::try_from(usize::from(node_index))
+        .expect("The node index corresponds to a valid PeerId")
+        .into()
+}
+
+pub fn to_node_index(peer_id: PeerId) -> NodeIndex {
+    usize::from(u16::from(peer_id)).into()
+}
```
elsirion, line 120: nit: Could that be part of an extension trait so one could call `node.to_peer_id()`?

elsirion (approved):
Looks reasonable, a bit hard to follow the moved code, but the consensus is quite in flux rn anyway.

PR #3412, opened 2023-10-17
feat(client): expose BitcoinTransactionData in onchain deposit

### LINKED ISSUE #3324: fedimint-cli/gateway-cli should show how much was pegged-in
Currently there is no easy easy to figure out how much a client has pegged-in the past. This makes it difficult, specially for the gateway, to do accounting of its funds in a given federation.

It would also be good to account the peg-outs.

Maan2003:
fixes #3324

### DIFF
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -96,13 +96,21 @@ pub trait WalletClientExt {
     ) -> anyhow::Result<UpdateStreamOrOutcome<WithdrawState>>;
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct BitcoinTransactionData {
+    /// The bitcoin transaction is saved as soon as we see it so the transaction
+    /// can be re-transmitted if it's evicted from the mempool.
+    pub btc_transaction: bitcoin::Transaction,
+    /// Index of the deposit output
+    pub out_idx: u32,
+}
+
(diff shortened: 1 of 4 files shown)


justinmoon:
This would be a great one to backport. But it's backwards-incompatible. Can we build a database migration for it?

m1sterc001guy:
We don't currently have the framework setup for client side db migrations. Most of the work is done, we just need to add an equivalent call to `apply_migrations` on the client side. 

Are we sure we want db migrations for states within the state machines? It's not clear to me if these will always be backwards compatible (i.e if the execution semantics of a state have changed, do we want to restart the client in this state without having executed the previous states?)

Perhaps we should make a separate issue to discuss

justinmoon:
@m1sterc001guy https://github.com/fedimint/fedimint/issues/3417

@Maan2003 needs rebase to fix flake self-check

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -96,13 +96,22 @@ pub trait WalletClientExt {
...
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub enum DepositState {
     WaitingForTransaction,
-    WaitingForConfirmation,
-    Confirmed,
+    WaitingForConfirmation(BitcoinTransactionData),
+    Confirmed(BitcoinTransactionData),
     // TODO: add amount
```
elsirion, line 113: I think this TODO can be removed now

elsirion (approved):

Issue #3405, opened 2023-10-16
Lightning client should retry payments in refund case


justinmoon:
If the gateway is offline when the lightning client attempts an outgoing payment, the user is put into a refund state where it waits for N block confirmations and then it gets a refund from the federation.

But if the Gateway comes back online, the client could retry the payment but currently doesn't.

m1sterc001guy:
Hmm I've been thinking about this more and I'm not sure if its desirable. If we retry the payment until the lightning gateway comes back online, this effectively locks the user's funds in the escrow contract until either the gateway comes back online, or the timeout occurs. Retrying the payment over and over again inside the state machine might result in unexpected behavior, like the payment being successful minutes or hours later.

Perhaps the lowest hanging fruit here is to just ping the gateway to confirm it is available before locking funds in the outgoing contract? That way the only way funds should get stuck is if the gateway goes offline between this "availability check" and the funding of the outgoing contract.

m1sterc001guy:
Or I suppose we could do both. Add an availability check to confirm the gateway is online, at which point we've "committed" to using that gateway. Then we retry the payment until either it is successful, the contract is canceled, or the timeout has occurred.

elsirion:
If we implemented a fix for https://github.com/fedimint/fedimint/issues/2746 that associates a public key with every outgoing contract (before the preimage is decrypte) we could add an authenticated API endpoint to the GW that lets users request a cancellation of the contract by the GW.

m1sterc001guy:
IMO the client should retry for some timeout, at which point it should request a cancellation (with some retries + backoff). The thinking is that, after a certain amount of time, the user isn't going to expect the payment to succeed and it might result in a weird user experience.

With the availability check that we recently added though, I don't think this is a high priority.

m1sterc001guy:
Not worth investigating with LNv2 shipping in v0.6 by default. Closing.

PR #3415, opened 2023-10-17
feat: add availability check before paying lightning invoice

m1sterc001guy:
A common footgun is to try and pay an invoice when the "active" lightning gateway is unavailable. This results in funds being locked until the timeout, which is not a good experience. 

An easy mitigation to this is just to ping the gateway before paying the invoice, to verify that it is available. That way, funds should only be stuck if the gateway goes down between the availability check and the funding transaction. 

Partially addresses: https://github.com/fedimint/fedimint/issues/3405

Next is adding retries to paying the invoice after the funding transaction is successful.

### DIFF
--- a/gateway/ln-gateway/src/rpc/rpc_server.rs
+++ b/gateway/ln-gateway/src/rpc/rpc_server.rs
@@ -27,7 +27,9 @@ pub async fn run_webserver(
 ) -> axum::response::Result<()> {
     let (routes, admin_routes) = if let Some(gateway_config) = config {
         // Public routes on gateway webserver
-        let routes = Router::new().route("/pay_invoice", post(pay_invoice));
+        let routes = Router::new()
+            .route("/pay_invoice", post(pay_invoice))
+            .route("/is_available", post(is_available));
 
         // Authenticated, public routes used for gateway administration
         let admin_routes = Router::new()
@@ -161,5 +163,13 @@ async fn set_configuration(
     Json(payload): Json<SetConfigurationPayload>,
 ) -> Result<impl IntoResponse, GatewayError> {
     gateway.handle_set_configuration_msg(payload).await?;
+    Ok(Json(json!(())))
+}
+
+#[instrument(skip_all, err)]
+async fn is_available(
+    Extension(_): Extension<Gateway>,
+    Json(_): Json<()>,
+) -> Result<impl IntoResponse, GatewayError> {
     Ok(Json(json!(())))
 }
--- a/gateway/ln-gateway/src/rpc/rpc_server.rs
+++ b/gateway/ln-gateway/src/rpc/rpc_server.rs
@@ -1,7 +1,7 @@
 use std::net::SocketAddr;
 
 use axum::response::IntoResponse;
-use axum::routing::post;
+use axum::routing::{get, post};
 use axum::{Extension, Json, Router};
 use axum_macros::debug_handler;
 use bitcoin_hashes::hex::ToHex;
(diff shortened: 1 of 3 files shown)


justinmoon:
Big concept ACK. This will really help end-user UX.

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -163,3 +165,11 @@ async fn set_configuration(
...
     Ok(Json(json!(())))
 }
+
+#[instrument(skip_all, err)]
+async fn is_available(
+    Extension(_): Extension<Gateway>,
+    Json(_): Json<()>,
+) -> Result<impl IntoResponse, GatewayError> {
+    Ok(Json(json!(())))
+}
```
elsirion, line 175: nit: We could probably make this an `info` endpoint that tells us a bit more about the GW, so that we e.g. know we are actually talking to the correct one.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -702,6 +702,31 @@ enum PayError {
 }
 
 impl LightningClientModule {
+    // Ping gateway endpoint to verify that it is available before locking funds in
+    // OutgoingContract
+    async fn verify_gateway_availability(&self, gateway: &LightningGateway) -> anyhow::Result<()> {
+        let response = reqwest::Client::new()
+            .post(
```
elsirion, line 709: Why is this a POST endpoint?

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -702,6 +702,31 @@ enum PayError {
...
+                gateway
+                    .api
+                    .join("is_available")
+                    .expect("is_available contains no invalid characters for a URL")
+                    .as_str(),
+            )
+            .json(&())
+            .send()
+            .await
+            .map_err(|e| anyhow::anyhow!("Gateway is not available: {e}"))?;
```
elsirion, line 719: nit: I think the more idiomatic way would be using `.context("Gateway is not available")?` here.

elsirion (approved):
Only nits, feel free to change or merge.

elsirion:
Actually, how would we go about backporting this? Could we abuse `/pay_invoice` in some way so new clients can still talk to old gateways?

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Changed to GET

gateway/ln-gateway/src/rpc/rpc_server.rs
m1sterc001guy: Changed this endpoint to `id`

elsirion (approved):

elsirion:
The second commit is more of a fixup to be fair ;)

justinmoon:
Wondering if there is any way to do an availability check without adding another round of communication? Perhaps something like user client sends escrow transaction to gateway (probably have to encrypt it so ecash inputs aren't visible?) then gateway would send this to the mint. Once it's confirmed, then it tries to pay lightning invoice.

This way the client would only need to make 1 API call to the gateway and 0 to the "mint". Currently it makes 2 API calls to gateway and 1 to the mint. I imaging that gateways will have faster internet connections, too, so offloading the responsibility for broadcasting the initial escrow transaction to the gateway might speed things up a little there as well.

elsirion:
> Perhaps something like user client sends escrow transaction to gateway (probably have to encrypt it so ecash inputs aren't visible?) then gateway would send this to the mint.

This might already be the plan for LN 2.0 @joschisan?

joschisan:
In the current design it should be only one API call from the client to the gateway to submit the decryption contract.

joschisan:
If I understand correctly the user can send the transaction funding the htlc to the gateway which then submits it, obviously only if the gateway is only. So its 1 API call to the gateway and 0 to the "mint".

PR #3419, opened 2023-10-18
Client-side ecash aggregation

joschisan:
Instead of collecting, verifying and aggregating the ecash blind signature shares via the guardians over the broadcast we move this functionality in the client. This significantly simplifies the mint server module aus it simply returns its own signature share for every accepted mint output and is done. Most importantly though it reduces the compute and disk consumption of the mint module - and therefore the federation as whole - to a fraction. Furthermore, the compute and disk consumption is now constant instead of linear in the number of guardians.

### DIFF
--- a/modules/fedimint-mint-client/src/output.rs
+++ b/modules/fedimint-mint-client/src/output.rs
@@ -1,17 +1,25 @@
+use std::collections::BTreeMap;
 use std::time::Duration;
 
+use anyhow::{anyhow, bail};
 use fedimint_client::sm::{ClientSMDatabaseTransaction, State, StateTransition};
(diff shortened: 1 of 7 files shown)


modules/fedimint-mint-client/src/output.rs
```diff
@@ -105,6 +113,29 @@ pub struct MintOutputStatesCreated {
     pub(crate) issuance_request: NoteIssuanceRequest,
 }
 
+pub fn verify_map(
```
dpc, line 116: **suggestion:** Comments, comments. I know it takes time to write, but as a reviewer i'm having much harder time reviewing without them and having to figure out what is what.

dpc (approved):

PR #3422, opened 2023-10-18
feat(mint-client): oob_notes_operation_id

Maan2003:
useful to cancel oob spend only using ecash notes without knowing the operation_id.

### DIFF
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -1101,11 +1101,7 @@ impl MintClientModule {
 
         let spendable_selected_notes = Self::select_notes(dbtx, min_amount).await?;
 
-        let operation_id = OperationId(
-            spendable_selected_notes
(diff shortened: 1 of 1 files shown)


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1279,6 +1275,16 @@ impl MintClientModule {
     }
 }
 
+pub fn oob_spend_operation_id(
```
elsirion, line 1278: Couldn't you even make this a member fn of [`OOBNotes`](https://docs.fedimint.org/fedimint_mint_client/struct.OOBNotes.html)? That's defined in the same crate afaik.

elsirion (dismissed):
Just one nit, feel free to merge :)

modules/fedimint-mint-client/src/lib.rs
Maan2003: tried this, `OOBNotes` has `FederationId` this function only. also changed the function name now.

elsirion (approved):

Issue #2746, opened 2023-07-10
2nd fedimint user to pay lightning invoice via a gateway receives "successful payment" response


justinmoon:
In [making gateway payments idempotent](https://github.com/fedimint/fedimint/pull/2695), we might have made them a little too idempotent! If 2 users pay the same invoice through the same gateway, the 2nd user will receive a "payment successful" response even though they didn't pay the invoice.

To solve this I think the `/pay_invoice` API endpoint might need a secret which the gateway can use to determine if calls to pay a previously paid invoice are from them same client that paid it (return "invoice paid"), or a different client (return error).

This is low priority.

elsirion:
There are two sub-problems imo:

1. How to prevent users from locking up funds in contracts for invoices that were already paid (federation-local)
2. How to prevent the described problem even for gateways serving multiple federations

If the gateway detects double-payment it could refund the second contract out of goodwill with an error code indicating double payment (it could also just take the money though … so this is more of a UX fix and not a fundamental security fix).

m1sterc001guy:
> To solve this I think the /pay_invoice API endpoint might need a secret which the gateway can use to determine if calls to pay a previously paid invoice are from them same client that paid it (return "invoice paid"), or a different client (return error)

For this, could we generate an ephemeral keypair and send the public key/signature to the gateway for each `pay_invoice` invocation? For the first `pay_invoice` request that the gateway receives, it could validate the signature and save the public key in the database, indexed by the contract id. Then for subsequent requests to `pay_invoice`, the gateway can enforce that the client provides a signature to the already saved public key (that way other users can't get the preimage for an already paid invoice).

I don't think this solves all of the above problems that @elsirion mentioned, but it should prevent leaking the preimage to users that did not pay the invoice. It would introduce a signature verification though which might add latency.

m1sterc001guy:
@joschisan in case he has some ideas on how to solve this

elsirion:
> For this, could we generate an ephemeral keypair and send the public key/signature to the gateway for each `pay_invoice` invocation?

Even more pragmatically we could generate some secret that is just sent with the API request and associated with that instance of the invoice payment. This secret could be the hash of a static secret derived from the module root secret and the payment hash, meaning it would be unique but deterministic for each user-invoice pair. If anyone else tries to do anything with an invoice that has stuff going on under a different secret the GW cancels the contract (if any) or just returns an error.

PR #3425, opened 2023-10-18
feat: Add Preimage Authentication to Gateway

m1sterc001guy:
This PR adds a "preimage authentication" as part of the payload that is sent to the gateway. The gateway will save the first secret that it sees. If a subsequent request does not send this same preimage authentication, the gateway will throw an error. This prevents clients that did not pay the invoice from retrieving the preimage.

Fixes: https://github.com/fedimint/fedimint/issues/2746

### DIFF
--- a/gateway/ln-gateway/src/state_machine/pay.rs
+++ b/gateway/ln-gateway/src/state_machine/pay.rs
@@ -2,6 +2,7 @@ use std::fmt::Display;
 use std::sync::Arc;
 use std::time::Duration;
 
+use bitcoin_hashes::sha256;
 use fedimint_client::sm::{ClientSMDatabaseTransaction, State, StateTransition};
 use fedimint_client::transaction::{ClientInput, ClientOutput};
 use fedimint_client::{ClientArc, DynGlobalClientContext};
@@ -9,6 +10,7 @@ use fedimint_core::config::FederationId;
 use fedimint_core::core::OperationId;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::{Amount, OutPoint, TransactionId};
+use fedimint_ln_client::pay::PayInvoicePayload;
 use fedimint_ln_common::api::LnFederationApi;
 use fedimint_ln_common::contracts::outgoing::OutgoingContractAccount;
 use fedimint_ln_common::contracts::{ContractId, FundedContract, IdentifiableContract, Preimage};
@@ -22,6 +24,7 @@ use tokio_stream::StreamExt;
 use super::{
     GatewayClientContext, GatewayClientExt, GatewayClientStateMachines, GatewayExtReceiveStates,
 };
+use crate::db::PreimageAuthentication;
 use crate::fetch_lightning_node_info;
 use crate::gateway_lnrpc::{PayInvoiceRequest, PayInvoiceResponse};
 use crate::lnrpc_client::LightningRpcError;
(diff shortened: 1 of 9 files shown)


gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -164,7 +168,7 @@ impl Display for OutgoingPaymentError {
 
 #[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
 pub struct GatewayPayInvoice {
-    pub contract_id: ContractId,
+    pub pay_invoice_payload: PayInvoicePayload,
```
okjodom, line 171: thought: it's slightly odd to see the payload abstraction being passed through to state machines

okjodom (commented):
Strong concept ACK :)
- Would it be good practice for clients to randomize the payment secret with every pay request? Gateway could profile clients based on the payment secret
- This is a first breaking change to `/pay_invoice`, might be a good idea to start versioning the API

elsirion:
Please don't call it payment secret, that's very confusing since it already has a meaning in LN.

m1sterc001guy:
> Would it be good practice for clients to randomize the payment secret with every pay request? Gateway could profile clients based on the payment secret

We can't do this, subsequent requests must use the same preimage auth each request, otherwise the payment we will fail the payment. We do, however, use a different preimage auth for every payment, so this kind of fingerprinting isn't possible across payments.

> This is a first breaking change to /pay_invoice, might be a good idea to start versioning the API

Any recommendations on how to do this? Do we want to support the API versioning negotiation that @dpc built for Fedimint already? I'm not sure how much of that is re-useable.

gateway/ln-gateway/src/state_machine/pay.rs
m1sterc001guy: It seemed to me to be the most convenient way to pass around this data. I didn't think making a new struct was worth it, but I'm open to suggestions

m1sterc001guy:
> Please don't call it payment secret, that's very confusing since it already has a meaning in LN.

Changed name to `PreimageAuthentication`

dpc:
> Any recommendations on how to do this? Do we want to support the API versioning negotiation that @dpc built for Fedimint already? I'm not sure how much of that is re-useable.

API version negotiation is a complication due to federated nature of Fedimint.

IIUC this is about call from client to a single gateway, so it's a standard thing. I don't have strong opinions, but there are plenty of articles about [best practices](https://www.xmatters.com/blog/best-practices-for-api-versioning).

One can use a header, new name for a new endpoint (especially useful for some one-off additions like here).

m1sterc001guy:
On second thought, I don't think we should support a backwards compatible version of this API. This is fixing a privacy leak, if we make it backwards compatible, then the gateway will still leak the preimage.

elsirion:
I don't think we need backwards compatibility, we are of the final push for 0.2 and don't really need to backport this to 0.1 imo.

Something more annoying that occurred to me going through all the issues: to implement the most complete solution to #3405 we will need public key auth anyway :confused:, this is relevant for the case:

- User wants to pay via LN, but GW goes offline after contract is locked
- Gateway comes back online a day after
- User doesn't want to send payment anymore and wants to ask the GW to cancel, but has never contacted the GW before

This is not super critical since we now check GW availability beforehand, but still good to keep in mind.

elsirion (dismissed):

okjodom (dismissed):

m1sterc001guy:
Rebased, should be ready to go

PR #3426, opened 2023-10-18
chore: migrate to using Flakebox

dpc:
[Flakebox](https://github.com/rustshop/flakebox) was explicitly created based on our dev env in Fedimint, generalizing to make it usable and customizable in any Rust project.

Moving all our building scripts, Nix cross-compilation system etc. out into a dedicate project has multiple benefits:

Fedimint module authors will have easier time re-using our setup. Possibly other Bitcoin-related projects as well.

Projects not related to Fedimint can use it, so it will get better coverage, features and testing.

The outside repo is smaller and allows focusing on core features of Nix dev shell and Rust support. E.g. in Flakebox there are Nix unit-tests making sure cross-compilation works well. This would be harder and confusing to build inside Fedimint repo itself.

Some of the things in Flakebox are already better than what we were doing in Fedimint. There's a chance for some minor things being broken in this migration, but it should be easy to fix with time.

Overall diff is mostly red color. It would look even more red if the lines of things that are auto-generated where not being counted in.

We return to the source.

[Return to the sauce](https://www.youtube.com/watch?v=S5LvhKbsHWU)


dpc:
@elsirion @justinmoon This changes the naming of the toolchains from the full triplet to the shorter "systems" strings, so will need admin override. Sorry for trouble.

justinmoon (dismissed):

justinmoon (dismissed):

justinmoon (approved):

Issue #2556, opened 2023-05-19
`connect-info` after a `join-federation` should work


douglaz:
From an UX standpoint, it's surprising that `fedimint-cli connect-info` will fail:
```
{
  "error": "CliError",
  "kind": "GeneralFederationError",
  "message": "cannot read connect string",
  "raw_error": "No such file or directory (os error 2)"
}
```
If I've joined through `fedimint-cli join-federation`

justinmoon:
That `[connect-info` ](https://github.com/fedimint/fedimint/blob/62dd3992b05c5d85eee4bea50929ba18c52fd1d6/fedimint-cli/src/lib.rs#L888-L901)command is completely incorrect. It reads a file that our test environment writes.

Right now it's impossible to generate a connection string from a client config. To implement this command we would either need to:
- Save the connection string we used to join
- Add a way to derive a connection string from client config
- Add an API for fetching a connection string

justinmoon:
Call: consider making the download token an `Option`. Then connection string would be derivable from config

okjodom:
use `fedimint-cli invite-code`

PR #3432, opened 2023-10-19
fix: fedimint-cli invite-code

### LINKED ISSUE #3178: `fedimint-cli dev invite-code` broken in mprocs
and I don't see how the current impl could ever work since the client is restarted for every command, resetting the internal state that's not persisted to DB. Maybe we should just save the invite code to the DB so this functionality can be implemented properly.

okjodom:
Removes `invite-code` command from `fedimint-cli dev` and re-adds it as a non-dev command. The invite code is sourced from client database where it was stored the first time a client is created

fixes #3178

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -77,7 +77,10 @@ use std::time::Duration;
 
 use anyhow::{anyhow, bail, ensure};
 use async_stream::stream;
-use db::{CachedApiVersionSet, CachedApiVersionSetKey, ClientConfigKey, ClientConfigKeyPrefix};
+use db::{
+    CachedApiVersionSet, CachedApiVersionSetKey, ClientConfigKey, ClientConfigKeyPrefix,
+    ClientInviteCodeKey, ClientInviteCodeKeyPrefix,
+};
 use fedimint_core::api::{
     ApiVersionSet, DynGlobalApi, DynModuleApi, GlobalFederationApi, IGlobalFederationApi,
     InviteCode, WsFederationApi,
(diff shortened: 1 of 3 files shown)


fedimint-client/src/lib.rs
```diff
@@ -1605,6 +1607,18 @@ pub async fn get_config_from_db(db: &Database) -> Option<ClientConfig> {
     config
 }
 
+pub async fn get_invite_code_from_db(db: &Database) -> Option<InviteCode> {
+    let mut dbtx = db.begin_transaction().await;
+    #[allow(clippy::let_and_return)]
+    let invite = dbtx
+        .find_by_prefix(&ClientInviteCodeKeyPrefix)
```
dpc, line 1614: **thought:** We should just have `DatabaseTransaction:find_singleton(key) -> Option<T>` with a default impl to make this pattern more usable.

fedimint-client/src/lib.rs
```diff
@@ -1605,6 +1607,18 @@ pub async fn get_config_from_db(db: &Database) -> Option<ClientConfig> {
     config
 }
 
+pub async fn get_invite_code_from_db(db: &Database) -> Option<InviteCode> {
+    let mut dbtx = db.begin_transaction().await;
+    #[allow(clippy::let_and_return)]
```
dpc, line 1612: **suggestion:** I would just apply the suggested clippy fix. Makes sense to me.

dpc (approved):
I tried and it works. Some suggestions, but LGTM.

fedimint-client/src/lib.rs
justinmoon: Thoughts @m1sterc001guy?

fedimint-client/src/lib.rs
okjodom: we need to drop dbtx before other local vars which forces us to use let and return

dpc:
```
pub async fn get_config_from_db(db: &Database) -> Option<ClientConfig> {
    db.begin_transaction()
        .await
        .find_by_prefix(&ClientConfigKeyPrefix)
        .await
        .next()
        .await
        .map(|(_, config)| config)
}
```

Works.

justinmoon:
@okjodom Does this close https://github.com/fedimint/fedimint/issues/2556?

okjodom:
Yes, but also the issue is outdated. command is now `fedimint-cli invite-code`

PR #3434, opened 2023-10-20
feat: propose vetted gateways meta field

okjodom:
While we should allow any gateway to connect to a federation, we need additional mechanisms by which federation members can vet such gateways. This PR proposes meta field structure for vetted gateways


elsirion (commented):
Please also add a link to this new meta field here: https://github.com/fedimint/fedimint/blob/master/docs/meta_fields/README.md

Also, I think writing the standard is a good first step and can land as an independent PR.

elsirion (approved):

PR #3435, opened 2023-10-20
bug: test for ecash oob cancel failure

Maan2003:
this test fails, 
run with 
```bash
cargo nextest run -E 'package(fedimint-mint-tests)' -- sends_ecash_out_of_band_cancel
```

### DIFF
--- a/modules/fedimint-mint-tests/tests/tests.rs
+++ b/modules/fedimint-mint-tests/tests/tests.rs
@@ -42,6 +42,27 @@ async fn sends_ecash_out_of_band() -> anyhow::Result<()> {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn sends_ecash_out_of_band_cancel() -> anyhow::Result<()> {
+    // Print notes for client1
+    let fed = fixtures().new_fed().await;
+    let client = fed.new_client().await;
+    let (op, outpoint) = client.print_money(sats(1000)).await?;
+    client.await_primary_module_output(op, outpoint).await?;
+
+    // Spend from client1 to client2
+    let (op, _) = client.spend_notes(sats(750), TIMEOUT, ()).await?;
+    let sub1 = &mut client.subscribe_spend_notes(op).await?.into_stream();
+    assert_eq!(sub1.ok().await?, SpendOOBState::Created);
+
+    client.try_cancel_spend_notes(op).await;
+    assert_eq!(sub1.ok().await?, SpendOOBState::UserCanceledProcessing);
+    assert_eq!(sub1.ok().await?, SpendOOBState::UserCanceledSuccess);
+
+    assert_eq!(client.get_balance().await, sats(1000));
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
 async fn error_zero_value_oob_spend() -> anyhow::Result<()> {
     // Print notes for client1
     let fed = fixtures().new_fed().await;


Maan2003:
figured out why this happens

https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Sender.html#method.subscribe

> new [Receiver](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html) handle that will receive values sent **after** this call to subscribe.

so if we call `try_cancel_ecash_spend` too fast, the message will get missed by state machine.

dpc:
Sending via channel here is not going to fly at all. We shouldn't really rely on the receiver being there at all.

Well, it is already even in the comment there:

```
    async fn try_cancel_spend_notes(&self, operation_id: OperationId) {
        let (mint, _instance) = self.get_first_module::<MintClientModule>(&KIND);

        // TODO: make robust by writing to the DB, this can fail
        let _ = mint.cancel_oob_payment_bc.send(operation_id);
    }
```

dpc:
You know what? Just returning this error to the caller would be an easy improvement already. And `try_*` suggests this being a best effort anyway.

dpc:
Clarification: I don't think this can only happen "if we're fast enough". IIRC there might be cases where we cancel and rebuild the state futures, and during that time there will be no `Receiver` either.

elsirion:
Also, trigger futures can be aborted after they read from the receiver but before the state transition happened. In practice  that should not be too much of a problem, but we might want to wait for the cancelled state to be reached and retry otherwise in this fn?

Alternatively we could have a set of to-be-canceled operations and only remove operations from that set when cancellation happened.

Both seemed overkill at the time, but some solution is clearly needed (if only for testability).

dpc:
Cancellations are PITA. See: async Rust. :D

If we are to have a general purpose cancellation, then every state machine implementation should take into account a possibility of being canceled, etc. Seems like a bad idea.

Which make think - maybe we don't want cancellation, but just some API to send programmatic events to state machines. In theory state machines could just do it off-band, but that seems laborious and error-prone.

So somewhere in the context or global_context:

```
    fn transitions(
        &self,
        context: &Self::ModuleContext,
        global_context: &DynGlobalClientContext,
```

we expose an API, to block waiting for the next event (typed and persisted).

But then I think... geeee... that's sure a lot of complexity.

Maybe it is better to just let the state machine explicity go through the database and wait for some key to appear to means whatever it wants.

So the state machine would `select` on external conditions and a key in the database, and have an explicit API method to write that key. All persisted, so should be OK.

elsirion:
> If we are to have a general purpose cancellation, then every state machine implementation should take into account a possibility of being canceled, etc. Seems like a bad idea.

They already had to afaik? In the initial design all non-selected trigger futures were cancelled. Later @douglaz optimized this, but I don't see "trigger futures never get cancelled" as part of the contract.

Generally these trigger futures should never write to persistent storage, they can output a value that is then given to the state transition function, where it can be persisted.

> Which make think - maybe we don't want cancellation, but just some API to send programmatic events to state machines. In theory state machines could just do it off-band, but that seems laborious and error-prone.

Yes, that's what I tried to hack together here :grimacing: (note to self: rather not implement something than doing it in a way that supposedly mostly works)

> Maybe it is better to just let the state machine explicity go through the database and wait for some key to appear to means whatever it wants.

We do have DB notifications for specific keys, that sounds reasonable.

dpc:
> They already had to afaik? In the initial design all non-selected trigger futures were cancelled.

The Rust Futures that state machines use to switch stat are cancellable on the language level. But the state machines itself are driven to completion. I don't think there's some implicit "Canceled" state that one could make a State Machine not execute anymore, and I think it's probably undesirable.

elsirion:
Then I fully agree with you @dpc.

elsirion:
Closed by #3900

PR #3437, opened 2023-10-20
backport: feat(client): expose BitcoinTransactionData in onchain deposit

Maan2003:
backport #3412

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -331,7 +331,7 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         .unwrap()
         .to_owned();
     assert_eq!(operation["operation_kind"].as_str().unwrap(), "wallet");
-    assert_eq!(operation["outcome"].as_str().unwrap(), "Claimed");
+    assert!(operation["outcome"]["Claimed2"].as_object().is_some());
 
     info!("Testing backup&restore");
     // TODO: make sure there are no in-progress operations involved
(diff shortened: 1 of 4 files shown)


devimint/src/main.rs
```diff
@@ -331,7 +331,7 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         .unwrap()
         .to_owned();
     assert_eq!(operation["operation_kind"].as_str().unwrap(), "wallet");
-    assert_eq!(operation["outcome"].as_str().unwrap(), "Claimed");
+    assert!(operation["outcome"]["Claimed"].as_object().is_some());
```
dpc, line 334: **suggestion:** `assert_eq!(operation["outcome"].as_str(), Some("claimed"));`

But yeah, it's a backport so whateva.

dpc (dismissed):

Maan2003:
`Claimed(Option<BitcoinTransactionData>)` doesn't work.
old serialization was: `"Claimed"`
new is: `{ "Claimed2": {...} }`
note: old is string, new is object

If I use Option<BitcoinTransaction>,
Claimed(None) will be `{ "Claimed": null }`

devimint/src/main.rs
Maan2003: this is incorrect, default serialization of enums is very weird in serde.

see other comment

justinmoon (approved):

PR #3443, opened 2023-10-23
refactor: cleanup OutgoingPaymentError in gateway

m1sterc001guy:
Error handling in the gateway's pay state machine is a bit cumbersome, adding another error type requires handling code in too many places. Additionally, each `OutgoingPaymentError` holds the contract, but this can be shared across all of the error types which simplifies the error handling.

Should simplify some of the logic in: https://github.com/fedimint/fedimint/pull/3425


okjodom (approved):
lgtm :)

Issue #3272, opened 2023-09-21
Leave Federation


justinmoon:
Conceptually clients can "leave a federation". AFAIK fedimint doesn't have any code to implement this anywhere, but I'm sure many of us have deleted client data directories created by `fedimint-cli` once they are no longer needed.

We should formalize what exactly it means to "leave a federation". In particular, we may want to add a method to `ClientModule` to cleanup any pending contracts which could cause the users to lose funds if they "left" the federation immediately.

For example:
* Cancel lightning offers so nobody pays the corresponding invoice funding an account which wouldn't be accessible once recipient leaves the federation.
* Sweep any lightning sends stuck in the refund case, but this requires waiting.
* External modules like the "stability pool" will need to withdraw from price-lock contracts into ecash.

justinmoon:
Quick fix idea: just check for active operations and don't allow "leaving" if there are any

tvolk131:
For my own understanding, your quick fix idea would involve:

1. Checking that no client funds are in-flight into/out of a federation
2. Checking that no client funds are at rest in a federation
3. Deleting client-side federation data

Would we also want to tell the federation that it can wipe any backup data for the given user?

justinmoon:
@tvolk131 Yes. We could say that (3) could be responsibility of whatever developer is integrating the client. (2) is pretty trivial. (3) is what needs work. Our trait system for the client needs a way to report "no client funds are in-flight into/out of a federation".

tvolk131:
> (3) is what needs work.

Typo? I think you mean (1).

justinmoon:
> Typo? I think you mean (1).

Yes. Fixed.

PR #3446, opened 2023-10-23
feat: ClientModule::leave 

dpc:
First step towards #3272

### DIFF
--- a/fedimint-client/src/module/mod.rs
+++ b/fedimint-client/src/module/mod.rs
@@ -3,6 +3,7 @@ use std::ffi;
 use std::fmt::Debug;
 use std::sync::Arc;
 
+use anyhow::bail;
 use fedimint_core::api::DynGlobalApi;
 use fedimint_core::core::{
     Decoder, DynInput, DynOutput, IntoDynInstance, ModuleInstanceId, OperationId,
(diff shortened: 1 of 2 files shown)


fedimint-client/src/module/mod.rs
```diff
@@ -183,6 +184,57 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    ///
+    /// This call indicates the desire of the caller client code
+    /// to orderly and safely leave the Federation by this module instance.
+    ///
+    /// The client module implementation should do whatever necessary and
+    /// possible to serve this request. E.g. persist this
+    /// state, preventing new interactions with the Federation. Notably
+    /// modules should not disable interaction that might be necessary
+    /// for the user (possibly through other modules) to leave. In particular
+    /// a Mint module should retain ability to create new modules,
```
shaurya947, line 202: `create new modules` or `create new ecash`?

fedimint-client/src/module/mod.rs
```diff
@@ -183,6 +184,57 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    /// MUST return a `Err`.
+    ///
+    /// This function should return `Ok` only if from the perspective
+    /// of this module instance, it is safe to delete its data and
+    /// stop using it.
+    ///
+    /// Calling code must NOT assume that a module that once returned `Ok`,
+    /// will not return `Err` at later point. E.g. a Mint module might have
+    /// no outstanding balance at first, but other modules winding down
+    /// might "cash-out" to Ecash.
```
shaurya947, line 221: Nice callout. Will be the case more often than not.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -909,6 +909,29 @@ impl ClientModule for MintClientModule {
...
+        }
+
+        if executor
+            .get_active_states()
+            .await
+            .into_iter()
+            .any(|s| s.0.module_instance_id() == module_instance_id)
+        {
+            bail!("Pending operations")
+        }
```
shaurya947, line 932: This code looks to be module-agnostic. Can we provide it to modules for free without having to write it by hand?

fedimint-client/src/module/mod.rs
```diff
@@ -183,6 +184,57 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    ///
+    /// Before leaving the Federation and deleting any state the calling code
+    /// must collect a full round of `Err`.
+    ///
+    /// Calling code should allow the user to override and ignore any
+    /// outstanding errors, after sufficient amount of warnings. Ideally,
+    /// this should be done on per-module basis, to avoid mistakes.
+    async fn leave(
```
shaurya947, line 229: Is it correct to say that this function shouldn't have any side effects. Meaning it should simply observe the current state and report back whether its safe to leave or not depending on that state (without changing any state)?
- If the above is correct, is there a way to enforce that through code? Perhaps somehow making the DB handle read-only?
- If the above is correct, how/where do we envision doing the necessary state changes to actually be able to leave? And what if those state changes take several minutes?

dpc: I think it should have side-effects. Like closing down pending contracts, etc.

shaurya947: Doc is vague in that case:

```
    /// The client module implementation should do whatever necessary and
    /// possible to serve this request. E.g. persist this
    /// state, preventing new interactions with the Federation.
```

^^ this makes it seem like module should do whatever is necessary _within the leave() function_.

```
    /// It should avoid running indefinitely trying to complete any cleanup
    /// actions necessary to reach a clean state, preferring spawning new
    /// state machines and returning an informative error.
```

^^ this makes it seem like _actually maybe we should think twice about doing whatever is necessary_. We might think doing a contract clean up or freeing up some locked balance should only take 30s but what if it could take 10+ minutes in some cases? As a module developer reading this documentation leaves me a bit confused.

It would be really nice to resolve ambiguity. Something more opinionated but specific would be nice, even if it was half the amount of documentation. Maybe even including an example in the documentation, like

```
    // Let's say there is a pending lightning contract which cannot be
    // cancelled and we have to wait X blocks for it to expire or go through.
    // In the case this function should do X, Y, and finally Z.
```

modules/fedimint-mint-client/src/lib.rs
dpc: Possibly sometehing to do along https://github.com/fedimint/fedimint/issues/3447

fedimint-client/src/module/mod.rs
```diff
@@ -184,6 +185,71 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    ///
+    /// The client module implementation should handle the request
+    /// and return as fast as possible avoiding blocking for longer than
+    /// necessary. This would usually involve some combination of:
+    ///
+    /// * recording the state of being in process of leaving the Federation to
+    ///   prevent initiating new conditions that could delay its completion;
+    /// * performing any fast to complete cleanup/exit logic;
+    /// * initiating any time-consuming logic (e.g. canceling outstanding
+    ///   contracts), as background jobs, tasks machines, etc.
```
elsirion, line 208: nit: did you mean "state machines"?

fedimint-client/src/module/mod.rs
```diff
@@ -184,6 +185,71 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    ///
+    /// Calling code should allow the user to override and ignore any
+    /// outstanding errors, after sufficient amount of warnings. Ideally,
+    /// this should be done on per-module basis, to avoid mistakes.
+    async fn leave(
+        &self,
+        _dbtx: &mut ModuleDatabaseTransaction<'_>,
+        _module_instance_id: ModuleInstanceId,
+        _executor: Executor<DynGlobalClientContext>,
+        _api: DynGlobalApi,
```
elsirion, line 249: Instead of supplying an executor reference+API, could we just supply a `DynGlobalClientContext`?

elsirion (approved):
Looking good, only one real nit.

fedimint-client/src/module/mod.rs
dpc: Copied it from the method above. I'd love to refactor altogether.

PR #3453, opened 2023-10-24
feat: add a xmpp server to full tls docker-compose

douglaz:
As fedimint uses ecash, which works very well through chat, it may be useful for test federations to have a standardized chat server like xmpp.

Uses prosody, which is very stable and lightweight. Tested on one machine with 2GB of RAM and there is no increase in memory or cpu requirements compared to current version.

### DIFF
--- a/docker/full-tls-mutinynet/docker-compose.yaml
+++ b/docker/full-tls-mutinynet/docker-compose.yaml
@@ -7,6 +7,7 @@
 # - gatewayd.fedimint.my-super-host.com
 # - gateway-ui.fedimint.my-super-host.com
 # - rtl.fedimint.my-super-host.com
+# - xmpp.fedimint.my-super-host.com
 
 version: "3.3"
 
@@ -21,8 +22,6 @@ services:
       - "--providers.docker.exposedbydefault=false"
       - "--entrypoints.websecure.address=:443"
       - "--certificatesresolvers.myresolver.acme.tlschallenge=true"
-      # - "--certificatesresolvers.myresolver.acme.caserver=https://acme-staging-v02.api.letsencrypt.org/directory"
-      # - "--certificatesresolvers.myresolver.acme.email=postmaster@example.com"
       - "--certificatesresolvers.myresolver.acme.storage=/letsencrypt/acme.json"
     ports:
       - "443:443"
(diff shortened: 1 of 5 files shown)


docker/full-tls-mutinynet/docker-compose.yaml
```diff
@@ -241,6 +241,172 @@ services:
     restart: always
     platform: linux/amd64
 
+  prosody_config_writer:
```
otech47, line 244: cant all instances of `fedimint.my-super-host.com` be written in as an environment variable or no?

douglaz: Not sure :thinking:

otech47 (commented):
I just tried testing this with a [tweaked commit](https://github.com/setlife-network/fedimint/commit/36a29121a0a2b487cce725d90aaeadb0962cba39) using my own DNS and get a failure to download the image

Screenshot 2023-10-25 at 11 10 19 AM

I guess because this PR needs to be merged first for the image to be deployed?

on a broader note I remember the install script used to prompt you to install the gateway `y/n` but i don't think i got that prompt anymore... i was going to suggest that this xmpp server also be an optional install. what do you think @douglaz? or does the same reasoning for removing the gateway optional install apply here as well?

douglaz:
> concept ACK
> 
> I just tried testing this with a [tweaked commit](https://github.com/setlife-network/fedimint/commit/36a29121a0a2b487cce725d90aaeadb0962cba39) using my own DNS and get a failure to download the image
> 
> Screenshot 2023-10-25 at 11 10 19 AM
> 
> I guess because this PR needs to be merged first for the image to be deployed?
> 
> on a broader note I remember the install script used to prompt you to install the gateway `y/n` but i don't think i got that prompt anymore... i was going to suggest that this xmpp server also be an optional install. what do you think @douglaz? or does the same reasoning for removing the gateway optional install apply here as well?

Yeah, it needs to be merged first. The prompt is defined only for the non-tls version.

justinmoon (approved):

docker/full-tls-mutinynet/docker-compose.yaml
```diff
@@ -44,6 +43,7 @@ services:
       - FM_P2P_URL=fedimint://fedimint.my-super-host.com:8173
       - FM_BIND_API=0.0.0.0:80
       - FM_API_URL=wss://fedimintd.fedimint.my-super-host.com
+      - FM_EXTRA_DKG_META=chat_server_domain=xmpp.fedimint.my-super-host.com
```
elsirion, line 46: The chat server meta field needs to be standardized by adding a description in [`docs/meta_fields`](https://github.com/fedimint/fedimint/tree/master/docs/meta_fields) or a similar external repository (in that case please prefix with the organization name).

PR #3463, opened 2023-10-26
dev: expose FM_INVITE_CODE in user shells

### LINKED ISSUE #3316: Devimint should set a `FM_INVITE_CODE` env var
I was just writing a little script that needed the invite code and had to read it from `invite-code` file on-disk. Would be nice if an environment variable was exported for this.

okjodom:
mprocs and tmux user shells get a ready to use `$FM_INVITE_CODE`

fix #3316

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -1440,6 +1440,20 @@ async fn rpc_command(rpc: RpcCmd, common: CommonArgs) -> Result<()> {
             .await?;
             let env = fs::read_to_string(&ready_file).await?;
             print!("{env}");
+
+            // Append invite code to devimint env
+            let env_file = common.test_dir.join("env");
+            let invite_file = common.test_dir.join("cfg/invite-code");
+            if fs::try_exists(&env_file).await.ok().unwrap_or(false)
+                && fs::try_exists(&invite_file).await.ok().unwrap_or(false)
+            {
+                let invite = fs::read_to_string(&invite_file).await?;
+                let mut env_string = fs::read_to_string(&env_file).await?;
+                writeln!(env_string, r#"export FM_INVITE_CODE="{invite}""#)?;
+                std::env::set_var("FM_INVITE_CODE", invite);
+                write_overwrite_async(common.test_dir.join("env"), env_string).await?;
+            }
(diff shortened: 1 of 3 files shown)


dpc:
Shouldn't this be happening in `devimint env`? Sorry, I'm out of context.

okjodom:
> Shouldn't this be happening in `devimint env`? Sorry, I'm out of context.

expected as much, but found out the devimint subprocess that has this info cannot mutate the parent process env var. Or, I haven't found out how to... https://github.com/fedimint/fedimint/issues/3316#issuecomment-1771791788

justinmoon:
This is an improvement, but only in the fedimint repo. Agree with @dpc that ideally devimint would handle this so that external repos using devimint (e.g. 3rd party modules) would have access to this environment variable. @Maan2003 Is this feasible?

Maan2003:
devimint can only set env variables during startup, not after after dkg. maybe `devimint wait` and `devimint env` can be combined.

Maan2003:
maybe `devimint wait` should also return some environment.

justinmoon (approved):

Maan2003 (approved):

devimint/src/main.rs
```diff
@@ -1440,6 +1440,20 @@ async fn rpc_command(rpc: RpcCmd, common: CommonArgs) -> Result<()> {
...
+            let invite_file = common.test_dir.join("cfg/invite-code");
+            if fs::try_exists(&env_file).await.ok().unwrap_or(false)
+                && fs::try_exists(&invite_file).await.ok().unwrap_or(false)
+            {
+                let invite = fs::read_to_string(&invite_file).await?;
+                let mut env_string = fs::read_to_string(&env_file).await?;
+                writeln!(env_string, r#"export FM_INVITE_CODE="{invite}""#)?;
+                std::env::set_var("FM_INVITE_CODE", invite);
+                write_overwrite_async(common.test_dir.join("env"), env_string).await?;
+            }
```
elsirion, line 1455: Nit: should we warn about failure here?

elsirion (approved):
LGTM

PR #3464, opened 2023-10-26
Refactor ClientBuilder to take in DerivableSecret

### LINKED ISSUE #3451: Allow specifying BIP32 `Xpriv` for root secret when building `fedimint-client`
Currently, `fedimint-client` is built using a random root secret as observed here:

https://github.com/fedimint/fedimint/blob/14f44c0f7890694bfb52e945c7a1d8c1f91c39f3/fedimint-client/src/lib.rs#L1681

The consumer who is building the client can choose between either a [`PlainRootSecretStrategy`](https://github.com/fedimint/fedimint/blob/14f44c0f7890694bfb52e945c7a1d8c1f91c39f3/fedimint-client/src/secret.rs#L64) (which is just random 64 bytes) or [`Bip39RootSecretStrategy`](https://github.com/fedimint/fedimint/blob/14f44c0f7890694bfb52e945c7a1d8c1f91c39f3/fedimint-bip39/src/lib.rs#L12) (which uses [rust-bip39 crate](https://github.com/rust-bitcoin/rust-bip39) to generate a random 12-word mnemonic and convert it to bytes).

Since `fedimint-client` is designed to be composable by other apps, front-ends, CLIs, etc. there arises a very valid use case whereby the _thing_ that's composing `fedimint-client` would like to have control over the root secret for the `fedimint-client` instance. For example, a CLI app might run several instances of `fedimint-client` on behalf of a user who is a part of several federations of varying versions. In such a case, the CLI app might want to have its own root BIP-39 seed that the user backs up once, and be able to initialize `fedimint-client` instances using `m/44'/0'/i'` xpriv derivations where `i` would the index that the CLI app maintains internally for each `fedimint-client` instance.

To address such use cases, we propose:
- adding a new `RootSecretStrategy` that utilizes the [`Xpriv`](https://github.com/rust-bitcoin/rust-bitcoin/blob/1eb6e0c7d85fbc5e3f265de069b91be2cb722359/bitcoin/src/bip32.rs#L69) type as the `Encoding` associated type.
- enhancing the builder API for `fedimint-client` to allow specifying an `Xpriv` from which to obtain the root secret.

shaurya947:
Closes #3451 

There are certain limitations with how root secret management works in `fedimint-client` today:
- A root secret cannot be provided externally. This makes `fedimint-client` difficult to compose in another application that wishes to manage the secret.
- Root secret management occurs implicitly within the `build()` call, with very little transparency to the caller. Here is the sequence of events:
  - Caller simply specifies the `RootSecretStrategy` ZST flavor to use as a generic argument on the `build()` call. This is typically one of the two pre-defined types: `Plain<>` or `Bip39<>`, or some custom external flavor.
  -  Using this strategy, root secret is generated by either (1) loading the secret material from the `fedimint-client` DB or (2) producing a random secret material as per the strategy's "encoding" and storing it in the DB—and then using the secret material to output a `DerivableSecret` which becomes the root secret.

This PR aims to decouple the process of root secret generation from the `build()` operation as follows:
- `build()` now takes as input a `DerivableSecret`. No secret storing/loading happens within `build()`. This has the following benefits:
  - A composer of `fedimint-client` can manage and transparently provide the root secret
  - `build()` no longer needs a generic strategy parameter
- There are separate methods exposed on `ClientBuilder` to store/load secret material as `Encodable`/`Decodable` generic types. This makes `fedimint-client`'s secret management service opt-in, further increasing flexibility to consumers.

`RootSecretStrategy` and its variants are still left as-is for convenience for `fedimint-cli`, `ln-gateway`, and some testing packages. It could be removed/simplified in a follow-up PR if desired, since now we don't have to play within the constraints of ZSTs.

**Note that we don't lose anything by moving away from the strategy pattern towards `Encodable`/`Decodable` types for storing/loading secret material.** This is because a consumer of `fedimint-client` may very well provide the wrong strategy type for loading a secret in the old implementation, just as they might provide the wrong `Decodable` type for loading a secret in the new implementation.

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -535,12 +535,30 @@ impl FedimintCli {
                     .build_client_builder(&self.module_inits, None)
                     .await
                     .map_err_cli_general()?;
-                client_builder
-                    .store_encodable_client_secret(secret)
+                let client_secret = match client_builder
+                    .load_decodable_client_secret::<[u8; 64]>()
                     .await
-                    .map_err_cli_general()?;
+                {
+                    Ok(existing_secret) if existing_secret == secret => secret,
+                    Ok(_) => {
+                        return Err(anyhow::anyhow!(
+                            "Provided secret does not match existing secret"
+                        ))
+                        .map_err_cli_general()
+                    }
+                    Err(_) => {
+                        let new_secret = PlainRootSecretStrategy::random(&mut thread_rng());
+                        client_builder
+                            .store_encodable_client_secret(new_secret)
+                            .await
+                            .map_err_cli_general()?;
+                        new_secret
+                    }
+                };
                 let client = client_builder
-                    .build_restoring_from_backup(PlainRootSecretStrategy::to_root_secret(&secret))
+                    .build_restoring_from_backup(PlainRootSecretStrategy::to_root_secret(
+                        &client_secret,
+                    ))
                     .await
                     .map_err_cli_msg(CliErrorKind::GeneralFailure, "failure")?
                     .0;
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -15,8 +15,8 @@ use std::{fs, result};
 use clap::{CommandFactory, Parser, Subcommand};
 use fedimint_aead::{encrypted_read, encrypted_write, get_encryption_key};
 use fedimint_client::module::init::{ClientModuleInit, ClientModuleInitRegistry};
-use fedimint_client::secret::PlainRootSecretStrategy;
-use fedimint_client::{get_invite_code_from_db, ClientBuilder, ClientSecret};
+use fedimint_client::secret::{PlainRootSecretStrategy, RootSecretStrategy};
+use fedimint_client::{get_invite_code_from_db, ClientBuilder};
 use fedimint_core::admin_client::WsAdminClient;
 use fedimint_core::api::{
     ClientConfigDownloadToken, FederationApiExt, FederationError, GlobalFederationApi,
@@ -36,6 +36,7 @@ use fedimint_mint_client::{MintClientExt, MintClientGen, SpendableNote};
 use fedimint_server::config::io::SALT_FILE;
 use fedimint_wallet_client::api::WalletFederationApi;
 use fedimint_wallet_client::{WalletClientGen, WalletClientModule};
+use rand::thread_rng;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 use thiserror::Error;
@@ -284,9 +285,23 @@ impl Opts {
         module_inits: &ClientModuleInitRegistry,
         invite_code: Option<InviteCode>,
     ) -> CliResult<fedimint_client::Client> {
-        let client_builder = self.build_client_builder(module_inits, invite_code).await?;
+        let mut client_builder = self.build_client_builder(module_inits, invite_code).await?;
+        let client_secret = match client_builder
+            .load_decodable_client_secret::<[u8; 64]>()
+            .await
+        {
+            Ok(secret) => secret,
+            Err(_) => {
+                let secret = PlainRootSecretStrategy::random(&mut thread_rng());
+                client_builder
+                    .store_encodable_client_secret(secret)
+                    .await
+                    .map_err_cli_general()?;
+                secret
+            }
+        };
         client_builder
-            .build::<PlainRootSecretStrategy>()
+            .build(PlainRootSecretStrategy::to_root_secret(&client_secret))
             .await
             .map_err_cli_general()
     }
@@ -516,13 +531,16 @@ impl FedimintCli {
                 hash: env!("FEDIMINT_BUILD_CODE_VERSION").to_string(),
             }),
             Command::Client(ClientCmd::Restore { secret }) => {
-                let client = cli
+                let mut client_builder = cli
                     .build_client_builder(&self.module_inits, None)
                     .await
-                    .map_err_cli_msg(CliErrorKind::GeneralFailure, "failure")?
-                    .build_restoring_from_backup(ClientSecret::<PlainRootSecretStrategy>::new(
-                        secret,
-                    ))
+                    .map_err_cli_general()?;
+                client_builder
+                    .store_encodable_client_secret(secret)
+                    .await
+                    .map_err_cli_general()?;
+                let client = client_builder
+                    .build_restoring_from_backup(PlainRootSecretStrategy::to_root_secret(&secret))
                     .await
                     .map_err_cli_msg(CliErrorKind::GeneralFailure, "failure")?
                     .0;
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -292,6 +292,7 @@ impl Opts {
         {
             Ok(secret) => secret,
             Err(_) => {
+                info!("Generating secret and writing to client storage");
                 let secret = PlainRootSecretStrategy::random(&mut thread_rng());
                 client_builder
                     .store_encodable_client_secret(secret)
@@ -547,6 +548,7 @@ impl FedimintCli {
                         .map_err_cli_general()
                     }
                     Err(_) => {
+                        info!("Generating secret and writing to client storage");
                         let new_secret = PlainRootSecretStrategy::random(&mut thread_rng());
                         client_builder
                             .store_encodable_client_secret(new_secret)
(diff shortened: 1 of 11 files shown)


fedimint-cli/src/lib.rs
```diff
@@ -284,9 +285,23 @@ impl Opts {
...
     ) -> CliResult<fedimint_client::Client> {
-        let client_builder = self.build_client_builder(module_inits, invite_code).await?;
+        let mut client_builder = self.build_client_builder(module_inits, invite_code).await?;
+        let client_secret = match client_builder
+            .load_decodable_client_secret::<[u8; 64]>()
+            .await
+        {
+            Ok(secret) => secret,
+            Err(_) => {
+                let secret = PlainRootSecretStrategy::random(&mut thread_rng());
```
shaurya947, line 295: Print an info log whenever we generate it.

fedimint-cli/src/lib.rs
```diff
@@ -516,13 +531,16 @@ impl FedimintCli {
...
-                let client = cli
+                let mut client_builder = cli
                     .build_client_builder(&self.module_inits, None)
                     .await
-                    .map_err_cli_msg(CliErrorKind::GeneralFailure, "failure")?
-                    .build_restoring_from_backup(ClientSecret::<PlainRootSecretStrategy>::new(
-                        secret,
-                    ))
+                    .map_err_cli_general()?;
+                client_builder
```
shaurya947, line 538: `devimint` backup-and-restore test fails because `wipe` command only deletes module state, not the secret. So here we have a few options:
- do what we're doing everywhere else, that is, try loading it first and write it only if absent
- load it, expect it to be there, and compare it with the provided secret, and proceed only if the two match
- `wipe` could also be made to delete the global client state

We might also not need `wipe` now that we have devimint and can just fork clients?

So for now we just go with:
- load secret
- if it exists, compare, and return err if different
- if it doesn't exist, store it
- continue with restoration

shaurya947: Done

fedimint-cli/src/lib.rs
```diff
@@ -516,12 +532,34 @@ impl FedimintCli {
...
+                {
+                    Ok(existing_secret) if existing_secret == secret => secret,
+                    Ok(_) => {
+                        return Err(anyhow::anyhow!(
+                            "Provided secret does not match existing secret"
+                        ))
+                        .map_err_cli_general()
+                    }
+                    Err(_) => {
+                        info!("Generating secret and writing to client storage");
```
dpc, line 551: **suggestion:** As this repeats verbatim with lines 295-301, I would make a small free function for it `populate_with_random_secret` or something.

fedimint-client/src/db.rs
```diff
@@ -1,23 +1,18 @@
...
 use crate::oplog::OperationLogEntry;
-use crate::secret::RootSecretStrategy;
-use crate::ClientSecret;
 
 #[repr(u8)]
 #[derive(Clone, EnumIter, Debug)]
 pub enum DbKeyPrefix {
-    ClientSecret = 0x29,
+    EncodedClientSecret = 0x28,
+    ClientSecret = 0x29, // Unused
```
dpc, line 15: **question:** If unused, why not delete it? If we want to make sure we won't re-use it, maybe we should change the comment `// tombstone - used in the past` or something.

gateway/ln-gateway/src/client.rs
```diff
@@ -94,9 +96,24 @@ impl GatewayClientBuilder {
             client_builder.with_database(rocksdb);
         }
 
+        let client_secret = match client_builder
```
dpc, line 99: **suggestion:** Looking at this makes me realize that possibly secret should be a part of the builder.

So the whole thing turns into:

```
client_builder
  .secret_new_if_missing(|| PlainRootSecretStrategy::random(...))
  .build()
  .await
  .map_err(GatewayError::ClientStateMachineError)?
```

or something like that.

gateway/ln-gateway/src/client.rs
```diff
@@ -94,9 +96,24 @@ impl GatewayClientBuilder {
...
+            .load_decodable_client_secret::<[u8; 64]>()
+            .await
+        {
+            Ok(secret) => secret,
+            Err(_) => {
+                info!("Generating secret and writing to client storage");
+                let secret = PlainRootSecretStrategy::random(&mut thread_rng());
```
dpc, line 106: **thought:** That argument to `random` is so needless (or rather: useful for 1% of the cases). We should probably have `random()` and `random_with(rng)`, where `random` just forwards to `random_with(&mut thread_rng())`

gateway/ln-gateway/src/client.rs
dpc: Oh, and if we go something like `.secret_new_if_missing(|| lambda)`, there's nothing preventing us from turning this confusing ZST "strategy` into `trait EncodedSecret` or something.

fedimint-cli/src/lib.rs
shaurya947: It's not quite verbatim.. over here we have a provided secret to compare against.

Would you like a `populate_random_secret_if_missing(validation_secret: Option<[u8; 64]>)`? I personally don't like it.

fedimint-client/src/db.rs
shaurya947: I wasn't sure if there's a need to ensure that it doesn't get re-used. I can just remove it.

gateway/ln-gateway/src/client.rs
shaurya947: What I didn't like about the strategy pattern was that it was implicit and hidden away whether a secret is being loaded and re-used, or whether a secret is being generated and stored. A function like `.secret_new_if_missing(|| PlainRootSecretStrategy::random(...))` takes us back to that downside in my mind. I like the caller providing explicit instructions and handling both cases externally. What do you think?

gateway/ln-gateway/src/client.rs
shaurya947: At this point, the `RootSecretStrategy` trait is just for convenience and we even talked about possibly removing it or moving it to the `fedimint-cli` package. I don't wish to further iterate on that trait within this PR if that's ok with you.

fedimint-client/src/db.rs
dpc: I'm also not sure. Probably safer to leave with a good comment and remove in the future.

gateway/ln-gateway/src/client.rs
dpc: Fair.

gateway/ln-gateway/src/client.rs
dpc: I guess I just like the superficial natural flow of the builder pattern. When the chain gets broken, especially with a larger block, it becomes less clear what is going on.

With the builder pattern the advantage is that we can just use lambdas and we don't need to deal with traits. 

A

```
builder
  .stmh()
  .smth();

let secret = builder.bar.bar + 12 some calls;

builder.build(secret);
```

Just structurally looks worse than:

```
builder
  .stmh()
  .smth();
  .secret(|prev/builder/somth| smth.bar.bar + 12 some calls)
  .build();
```
  
as the details of dealing with the secret are structurally delegated (extra indentation).

In both cases we did move out dealing with the secrets from the internals of the `build`, giving user more flexibility. It's just looks a bit more readable, IMO.

`secret_new_if_missing` is kind of clunky, but maybe some `secret_with(|existing : Option<_>| ...)` would be handier. Or something.

Not a blocker, we can iterate later.

dpc (approved):

fedimint-client/src/db.rs
elsirion: Why not keep the same key? From a first glance you don't seem to change the encoding?

gateway/ln-gateway/src/client.rs
elsirion: The real question seems to be who is responsible for storing the secret. This PR makes it kinda an external responsibility, but the builder has some helper functions (because it has a DB reference I guess), which feels a bit unclean. In the future there could be a `SecretStore` object/trait that is independent of the client builder to separate concerns fully.

@dpc's approach would keep it a client responsibility with a simple way to override instead.

I think in the long run having an external secret store abstraction is the more flexible solution and we can iterate on developer UX as we go.

gateway/ln-gateway/src/client.rs
elsirion: Yes, `RootSecretStrategy` feels like accidental complexity because I didn't think of just storing a `Vec<u8>` in the DB instead of `ClientSecret`, which then needed a way of specifying the encoding. Should likely be removed and the encoding/strategy can be baked into whatever manages the secret (not the client anymore).

elsirion (approved):
A step into the right direction, I think we'll need some iterations to dial in developer experience vs. flexibility.

Issue #3447, opened 2023-10-23
refactor: Replace `ClientModuleExt` and tons of arguments with a `ClientModuleCtx`.


dpc:
If you look at `ClientModuleInit::init` (`ClientModuleInitArgs`), and then all the methods of `ClientModule` taking all these arguments to executor, and then at all the `WalletClientExt` and alikes, it's clear to see that all these combinations are due to client module not having access to the outside executor and "other context" of the client it is running it. This causes us to pass these bits over and over, or extra functionality to work on the `Client` and module from the outside.

It seems to me that it would be much better to just give each module some "handle" to the `Client` and other "context". I call it `ClientModuleCtx` here, but naming in particular is open for debate.

```rust
struct ClientModuleCtx<C> where C: ClientModuleInit {
  module_instance_id: ModuleInstanceId,
  client: Arc<sync::Weak.html>, // `Weak` to prevent ownership cycles
  // possibly other stuff, possibly typed with `C`
}
```

This is basically a generalization of `struct ClientModuleInitArgs<C>`. The client module would be expected to store it and use it, instead of all these arguments.

We can use `ClientModuleCtx` to mediate the access between module logic and the outer-context. How exactly would it like will remain to be designed once build around it, but I think it will largely (though maybe not completely) eliminate any need of implementing things in `WalletClientExt`-like way.

dpc:
@elsirion thoughts?

dpc:
I done.

PR #3466, opened 2023-10-26
refactor: ClientWeak and move logic to ClientInner

dpc:
An initial step for #3447

Rationale:

I'd like a `ClientWeak` to be a thing. Easy.

But then I'd also want a way for APIs to use `ClientWeak`, upgrade
it to `Client` (strong Arc), yet not allow the user of such APIs
to store them (to prevent introducing harmful ownership cycles).

For this to not require duplicating forwarding methods manually,
the inner `ClientInner` should just do as much as possible.
So in thie change I'm moving all methods to `ClientInner`,
making it more usable.

Also, now `Client` will just `Deref` to `ClientInner` to make the
whole thing mostly invisible from the user perspective. And
I think that's totally OK and even preferable. After all `Client`
is just a newtype-ed smart-pointer to `ClientInner`.

We can control the visibility of methods of `ClientInner` with
`pub`, `pub(crate)` etc. Importantly, this ends the confusing
and annoying split: some methods being implemented on `Client`,
some on `ClientInner`.

The only thing that bothers me is the naming now.

### DIFF
--- a/fedimint-client/src/backup.rs
+++ b/fedimint-client/src/backup.rs
@@ -14,7 +14,7 @@ use secp256k1_zkp::{KeyPair, Secp256k1};
 use serde::{Deserialize, Serialize};
 use tracing::{debug, info, warn};
 
-use super::Client;
+use super::ClientArc;
 use crate::get_decoded_client_secret;
 use crate::secret::DeriveableSecretClientExt;
 
@@ -135,7 +135,7 @@ impl EncryptedClientBackup {
     }
 }
 
-impl Client {
+impl ClientArc {
     /// Create a backup, include provided `metadata`
     pub async fn create_backup(&self, metadata: Metadata) -> anyhow::Result<ClientBackup> {
         let fedimint_block_count = self.inner.api.fetch_block_count().await?;
--- a/fedimint-client/src/backup.rs
+++ b/fedimint-client/src/backup.rs
@@ -14,7 +14,7 @@ use secp256k1_zkp::{KeyPair, Secp256k1};
 use serde::{Deserialize, Serialize};
 use tracing::{debug, info, warn};
 
-use super::ClientArc;
+use super::Client;
 use crate::get_decoded_client_secret;
 use crate::secret::DeriveableSecretClientExt;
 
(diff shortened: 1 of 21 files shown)


fedimint-client/src/backup.rs
```diff
@@ -135,7 +135,7 @@ impl EncryptedClientBackup {
     }
 }
 
-impl Client {
+impl ClientArc {
```
shaurya947, line 138: In the PR description, it says

> For this to not require duplicating forwarding methods manually,
the inner ClientInner should just do as much as possible.
So in thie change I'm moving all methods to ClientInner,
making it more usable.

Shouldn't we also move all these backup-related methods on to `ClientInner`?

fedimint-client/src/lib.rs
```diff
@@ -362,7 +362,7 @@ impl Debug for ClientInner {
 /// aware of their instance id etc.
 #[derive(Clone, Debug)]
 struct ModuleGlobalClientContext {
-    client: Arc<ClientInner>,
+    client: Arc<Client>,
```
shaurya947, line 365: Should this just be a `ClientArc`?

fedimint-client/src/lib.rs
```diff
@@ -533,50 +565,143 @@ const SUPPORTED_CORE_API_VERSIONS: &[fedimint_core::module::ApiVersion] =
...
+pub struct ClientModuleInstance {
+    /// Instance id of the module
+    pub id: ModuleInstanceId,
+    /// Module-specific DB
+    pub db: Database,
+    /// Module-specific API
+    pub api: DynModuleApi,
+}
+
+pub struct Client {
```
shaurya947, line 578: `ClientInner` was not pub, but `Client` is now pub.. how come?

fedimint-client/src/lib.rs
dpc: Yes. :)

fedimint-client/src/lib.rs
dpc: Now we need to be able to call methods on it directly, as `&Client` supposed to be usable, where before `Client` (now `ClientArc`) had all the public methods, usually just delegating to the inner one.

We can achieve the same control of method visibility with `pub`, `pub(crate)` and private methods.

fedimint-client/src/backup.rs
dpc: Which methods exactly? Did I overlooked anything? Generally yes, everything should be on a new `Client`.

elsirion (approved):
Looks cleaner, thx!

Issue #3467, opened 2023-10-27
Next Generation Lightning Integration


joschisan:
The current plan is to do complete rewrite of the lightning module code by creating a new module instead of updating the legacy code.
- [x] Implement threshold point encryption based on BLS cryptography, update distributed key generation to support the scheme, implement a new ServerModule.
- [x] Implement a new ClientModule
- [x] Update the Gateway

m1sterc001guy:
I don't think we need to actually implement a new module on the client side, since the outgoing payment contract structure isn't changing. I think we just need to implement a new receive state machine on the client side and a new incoming state machine on the gateway side. That should save us some work.

elsirion:
How would we select which one to use? If the new LN module was just a completely new module kind that would be simple. If we keep the old client lib then maybe we could use versioning? @dpc

dpc:
@elsirion 

IIRC we could use versioning inside one codebase as well. The code can do different things depending on the module consensus version.

Which one will be easier is a hard call to make. Separate module will have a problem that both modules will be "available", and we need somehow make new Federations generate only the new one, but also handle the existing one. Doable, but worth mentioning.

With the single module we could actually try out module consensus versioning. Theoretically everything should be there to handle it already. `ServerModuleInit::versions` will return both versions, so will `ServerModuleInit::api_versions`, then `ServerModuleInit::init` will get the config (with the version to use) in `&ServerModuleInitArgs`, and then the code will be sprinkled with conditionals (shouldn't be too bad, literally an `if` in every trait function that forwards to one or the other implementation).

m1sterc001guy:
Remaining follow up items in LNv2 integration

- [ ] Document `crypto/tpe` crate
- [ ] Gateways and lightning nodes are assumed to be 1-1, which doesn't allow for custodial LN wallets to be liquidity providers. This means the gateway takes a fee even for internal payments within the federation. Currently, internal payments do no bypass the gateway.
- [ ] CLN `create_invoice` isn't implemented yet.
- [ ] Errors do no contain any debugging information except a message
- [ ] Gateway fees are not configurable
- [ ] Gateway tests need to be re-implemented for LNv2
- [ ] `PaymentResult` that prevents the LN client from paying an invoice twice needs to be implemented
- [ ] More informative debugging when things go wrong

elsirion:
Call: we can close this

PR #3468, opened 2023-10-27
Implement Next Generation Lightning ServerModule

joschisan:


modules/fedimint-ln-client-ng/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+mod internal_send;
+mod lightning_send;
+mod receive;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LightningOperationMeta;
+
+const RETRY_DELAY: Duration = Duration::from_secs(1);
+
+/// The high-level state of an payment operation over lightning,
```
m1sterc001guy, line 49: Just adding this comment so I don't forget (I know its a nit), but we'll need to add more metadata to these high level states

modules/fedimint-ln-client-ng/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+                LightningCommonInit::KIND.as_str(),
+                |_, _| LightningOperationMeta,
+                transaction,
+            )
+            .await?;
+
+        Ok(operation_id)
+    }
+
+    pub async fn await_incoming_contract(
```
m1sterc001guy, line 210: Doesn't look like this function is used currently

modules/fedimint-ln-client-ng/src/lightning_send.rs
```diff
@@ -0,0 +1,466 @@
...
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///  Created -- funding tx is rejected --> FundingRejected
+///  Created -- funding tx is accepted --> Funded    
+///  Funded -- post invoice returns preimage  --> Success
+///  Funded -- post invoice returns forfeit tx --> Refunding
+///  Funded -- await_preimage returns preimage --> Success
+///  Funded -- timelock expires --> Refunding
+///  Funded -- post invoice returns invalid response --> GatewayError
```
m1sterc001guy, line 43: Do we want to sit in this temporary `GatewayError` state when the gateway does something unexpected? The scenario I'm thinking of is this: say there's a bug on the gateway side that causes an error to be returned and the client to transition to `GatewayError`. We fix the bug, then the gateway operator updates their gateway code and now the payment should succeed. In this case the client will never try again and will wait until the timelock....but they could try again and since they gateway has been updated, it should succeed.

modules/fedimint-ln-client-ng/src/lightning_send.rs
```diff
@@ -0,0 +1,466 @@
...
+    }
+
+    async fn try_gateway_pay_invoice(
+        gateway_api: SafeUrl,
+        payload: PayInvoicePayload,
+    ) -> anyhow::Result<GatewayResponse> {
+        Ok(reqwest::Client::new()
+            .post(
+                gateway_api
+                    .join("pay_invoice")
```
m1sterc001guy, line 223: Note: We'll need to use a new endpoint here

modules/fedimint-ln-client-ng/src/receive.rs
```diff
@@ -0,0 +1,168 @@
...
+const RETRY_DELAY: Duration = Duration::from_secs(1);
+
+#[cfg_attr(doc, aquamarine::aquamarine)]
+/// State machine that waits on the receipt of a Lightning payment.
+///
+/// ```mermaid
+/// graph LR
+/// classDef virtual fill:#fff,stroke-dasharray: 5 5
+///
+///     Created -- incoming contract is confirmed --> Claiming
```
m1sterc001guy, line 30: TODO: Add Pending state here?

modules/fedimint-ln-client-ng/src/lightning_send.rs
joschisan: The alternative here would be to stay in the Funded state and retry like we would do in case of a network error. I have considered this and am actually still torn about it so good catch! It would make the state machine simpler. The small downside would be continuous retries - however we could do an exponential backoff here? Its a somewhat theoretical state since the gateway would have to return a invalid signature or the wrong preimage to trigger the transition to GatewayError.

joschisan: So I think it makes sense to just retry and remove the gateway error state. Just pushed a commit.

m1sterc001guy: +1 agreed with that. The retry delay in this scenario can be much longer than the network error retry

benthecarman:
Is there a writeup anywhere on the goals with this and how it differs from the old model?

justinmoon:
> Is there a writeup anywhere on the goals with this and how it differs from the old model?

Issue is https://github.com/fedimint/fedimint/issues/3467, but it's missing a lot of detail I imagine.

Cargo.toml
```diff
@@ -42,6 +43,10 @@ members = [
     "modules/fedimint-ln-client",
     "modules/fedimint-ln-server",
     "modules/fedimint-ln-tests",
+    "modules/fedimint-ln-common-ng",
```
dpc, line 46: Note: `ng` does not get old well. `lnv2` maybe? So in two years we don't need to have `ln-ng-ng` or something. ;)

elsirion: It's funny when you decommission hardware that was once named "NG", I've seen that a few times :laughing:

m1sterc001guy: +1 to `lnv2`

fedimint-testing/src/ln/mock.rs
```diff
@@ -163,4 +163,28 @@ impl ILnRpcClient for FakeLightningTest {
...
+        let invoice = InvoiceBuilder::new(Currency::Regtest)
+            .description("".to_string())
+            .payment_hash(payment_hash)
+            .current_timestamp()
+            .min_final_cltv_expiry_delta(0)
+            .payment_secret(PaymentSecret([0; 32]))
+            .amount_milli_satoshis(create_invoice_request.amount_msat)
+            .expiry_time(Duration::from_secs(create_invoice_request.expiry as u64))
+            .build_signed(|m| ctx.sign_ecdsa_recoverable(m, &self.gateway_node_sec_key))
+            .unwrap();
```
m1sterc001guy, line 184: Nit: use `expect` instead of `unwrap`

gateway/ln-gateway/src/db.rs
```diff
@@ -151,6 +153,15 @@ async fn migrate_to_v1(dbtx: &mut DatabaseTransaction<'_>) -> Result<(), anyhow:
     Ok(())
 }
 
+#[derive(Debug, Encodable, Decodable)]
+pub struct CreateInvoicePayloadKey(pub [u8; 32]);
```
m1sterc001guy, line 157: This is the `payment_hash` correct? Lets make this a named parameter

modules/fedimint-ln-common-ng/src/db.rs
```diff
@@ -0,0 +1,127 @@
+use bitcoin_hashes::sha256;
```
m1sterc001guy, line 1: This file should be moved to `fedimint-ln-server-ng`, we shouldn't share db structures between client and server (recently fixed this in the legacy implementation as well)

modules/fedimint-ln-common-ng/src/lib.rs
```diff
@@ -0,0 +1,173 @@
+//! # Lightning Module
+//!
+//! This module allows to atomically and trustlessly (in the federated trust
+//! model) interact with the Lightning network through a Lightning gateway.
+//!
+//! ## Attention: only one operation per contract and round
+//! If this module is active the consensus' conflict filter must ensure that at
+//! most one operation (spend, funding) happens per contract per round
```
m1sterc001guy, line 8: Could you clarify this comment? The consensus logic already provides this assurance right? Or is this something that the module needs to ensure?

scripts/tests/backend-test.sh
```diff
@@ -48,6 +48,10 @@ function run_tests() {
   if [ -z "${FM_TEST_ONLY:-}" ] || [ "${FM_TEST_ONLY:-}" = "bitcoind-ln-gateway" ]; then
     >&2 echo "### Testing against bitcoind for ln-gateway"
 
+    cargo nextest run --locked --workspace --all-targets \
+          ${CARGO_PROFILE:+--cargo-profile ${CARGO_PROFILE}} ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} \
+          ${TEST_ARGS_SERIALIZED} \
+          -E 'package(fedimint-ln-tests-ng)'
```
m1sterc001guy, line 54: nit: formatting is slightly off?

fedimint-testing/src/ln/mock.rs
```diff
@@ -163,4 +163,28 @@ impl ILnRpcClient for FakeLightningTest {
...
+        let ctx = bitcoin::secp256k1::Secp256k1::new();
+
+        let payment_hash = sha256::Hash::from_slice(&create_invoice_request.payment_hash)
+            .expect("Failed to lookup FederationId");
+        let invoice = InvoiceBuilder::new(Currency::Regtest)
+            .description("".to_string())
+            .payment_hash(payment_hash)
+            .current_timestamp()
+            .min_final_cltv_expiry_delta(0)
+            .payment_secret(PaymentSecret([0; 32]))
```
m1sterc001guy, line 180: Note: going forward, if we want to support MPP payments I think we'll need to set this correctly. Obviously not blocking for now.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -496,6 +497,13 @@ impl GatewayLightning for ClnRpcService {
         }
         Ok(tonic::Response::new(EmptyResponse {}))
     }
+
+    async fn create_invoice(
+        &self,
+        _create_invoice_request: tonic::Request<CreateInvoiceRequest>,
+    ) -> Result<tonic::Response<CreateInvoiceResponse>, Status> {
+        todo!()
```
m1sterc001guy, line 505: nit: return error for now instead of todo?

fedimint-client/src/module/mod.rs
```diff
@@ -452,6 +452,38 @@ where
...
+        operation_id: OperationId,
+        op_type: &str,
+        operation_meta: impl serde::Serialize + Debug,
+        sms: Vec<DynState>,
+    ) -> anyhow::Result<()> {
+        let db = self.client.get().db().clone();
+        let mut dbtx = db.begin_transaction().await;
+
+        if Client::operation_exists_dbtx(&mut dbtx.to_ref_nc(), operation_id).await {
+            bail!("Operation already exists");
```
m1sterc001guy, line 467: Add which `OperationId` this was for debugging purposes

fedimint-client/src/module/mod.rs
```diff
@@ -452,6 +452,38 @@ where
...
     }
+
+    pub async fn manual_operation_start(
+        &self,
+        operation_id: OperationId,
+        op_type: &str,
+        operation_meta: impl serde::Serialize + Debug,
+        sms: Vec<DynState>,
+    ) -> anyhow::Result<()> {
+        let db = self.client.get().db().clone();
```
m1sterc001guy, line 463: Do you need to clone the db here?

fedimint-client/src/module/mod.rs
```diff
@@ -452,6 +452,38 @@ where
...
+            .await;
+
+        self.client
+            .get()
+            .executor
+            .add_state_machines_dbtx(&mut dbtx.to_ref_nc(), sms)
+            .await
+            .unwrap();
+
+        dbtx.commit_tx_result().await.unwrap();
```
m1sterc001guy, line 483: Can we use `module_autocommit_v2`?

fedimint-client/src/module/mod.rs
```diff
@@ -452,6 +452,38 @@ where
     pub fn get_internal_payment_markers(&self) -> anyhow::Result<(PublicKey, u64)> {
         self.client.get().get_internal_payment_markers()
     }
+
+    pub async fn manual_operation_start(
```
m1sterc001guy, line 456: Doc comment here would be great. This is a nice helper method!

gateway/ln-gateway/src/gateway_module_ng/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn operation_id(&self) -> OperationId {
+        match self {
+            GatewayClientStateMachinesNG::Receive(state) => state.operation_id(),
+            GatewayClientStateMachinesNG::Send(state) => state.operation_id(),
+        }
+    }
+}
+
+impl GatewayClientModuleNG {
+    pub async fn start_send_state_machine(
```
m1sterc001guy, line 189: nit: bit of a technical name for the high level function that pays the invoice. Any objections to `pay_bolt11_invoice_v2`?

gateway/ln-gateway/src/gateway_module_ng/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+                vec![self.client_ctx.make_dyn_state(send_sm)],
+            )
+            .await
+    }
+
+    pub async fn subscribe_send(
+        &self,
+        operation_id: OperationId,
+        contract: OutgoingContract,
+    ) -> Result<[u8; 32], Signature> {
```
m1sterc001guy, line 222: I think its probably worth making a new type for `Preimage`, I've seen it in a few places in the code

gateway/ln-gateway/src/gateway_module_ng/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+                GatewayOperationMetaNG,
+                vec![self.client_ctx.make_dyn_state(send_sm)],
+            )
+            .await
+    }
+
+    pub async fn subscribe_send(
+        &self,
+        operation_id: OperationId,
+        contract: OutgoingContract,
```
m1sterc001guy, line 221: Since `OutgoingContract` is a parameter here, does this mean that in order for the gateway to subscribe to a contract, it needs to keep track of the `OutgoingContract` in addition to the `OperationId`? Can we put the `OutgoingContract` in the operation log so that we only need to keep track of the `OperationId`?

gateway/ln-gateway/src/gateway_module_ng/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    pub async fn start_receive_state_machine(
+        &self,
+        operation_id: OperationId,
+        contract: IncomingContract,
```
m1sterc001guy, line 241: Same question as above for `OutgoingContract`

gateway/ln-gateway/src/gateway_module_ng/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    pub async fn subscribe_receive(&self, operation_id: OperationId) -> Option<[u8; 32]> {
+        let mut stream = self.notifier.subscribe(operation_id).await;
+
+        loop {
+            if let Some(GatewayClientStateMachinesNG::Receive(state)) = stream.next().await {
+                match state.state {
+                    ReceiveSMState::Funding(..) => {}
+                    ReceiveSMState::Rejected(..) => return None,
```
m1sterc001guy, line 279: This function should probably return the `UpdateStreamOrOutcome` similar to how we do it in legacy LN rather than just a preimage, so that we know what went wrong (rejected vs failure, etc)

gateway/ln-gateway/src/gateway_module_ng/receive_sm.rs
```diff
@@ -0,0 +1,243 @@
...
+pub struct ReceiveStateMachine {
+    pub operation_id: OperationId,
+    pub contract: IncomingContract,
+    pub out_point: OutPoint,
+    pub state: ReceiveSMState,
+}
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+pub enum ReceiveSMState {
```
m1sterc001guy, line 39: These state machines should be much simpler than before, but can we still add the mermaid diagram? I found those helpful

gateway/ln-gateway/src/gateway_module_ng/receive_sm.rs
```diff
@@ -0,0 +1,243 @@
...
+    fn transitions(
+        &self,
+        context: &Self::ModuleContext,
+        global_context: &DynGlobalClientContext,
+    ) -> Vec<StateTransition<Self>> {
+        let gc = global_context.clone();
+        let tpe_agg_pk = context.tpe_agg_pk;
+
+        match &self.state {
+            ReceiveSMState::Funding(..) => {
```
m1sterc001guy, line 59: I love this. So much simpler!

gateway/ln-gateway/src/gateway_module_ng/receive_sm.rs
```diff
@@ -0,0 +1,243 @@
...
+        out_point: OutPoint,
+        decryption_contract: IncomingContract,
+    ) -> BTreeMap<PeerId, DecryptionKeyShare> {
+        let verify_decryption_share = move |peer, outcome: SerdeOutputOutcome| {
+            let outcome = deserialize_outcome::<LightningOutputOutcome>(outcome, &module_decoder)?;
+
+            match outcome {
+                LightningOutputOutcome::Incoming(share) => {
+                    if !decryption_contract.verify_decryption_share(
+                        tpe_pks.get(&peer).ok_or(anyhow!("Unknown peer pk"))?,
```
m1sterc001guy, line 143: We probably will want to make these error messages (here and below) slightly more descriptive in case this ever happens in a production deployment

gateway/ln-gateway/src/gateway_module_ng/receive_sm.rs
```diff
@@ -0,0 +1,243 @@
...
+            .map(|(peer, share)| (peer.to_usize() as u64 + 1, share))
+            .collect();
+
+        let agg_decryption_key = aggregate_decryption_shares(&decryption_shares);
+
+        if !old_state
+            .contract
+            .verify_agg_decryption_key(&tpe_agg_pk, &agg_decryption_key)
+        {
+            // This implies that the lightning client config's public keys are inconsistent
```
m1sterc001guy, line 206: We need a loud error message here if this happens

gateway/ln-gateway/src/gateway_module_ng/receive_sm.rs
```diff
@@ -0,0 +1,243 @@
...
+            return ReceiveStateMachine {
+                operation_id: old_state.operation_id,
+                contract: old_state.contract,
+                out_point: old_state.out_point,
+                state: ReceiveSMState::Failure,
```
m1sterc001guy, line 211: I think the error message should also be included in the state

gateway/ln-gateway/src/gateway_module_ng/receive_sm.rs
```diff
@@ -0,0 +1,243 @@
...
+                contract: old_state.contract,
+                out_point: old_state.out_point,
+                state: ReceiveSMState::Success(preimage),
+            };
+        }
+
+        let client_input = ClientInput::<LightningInput, LightningClientStateMachines> {
+            input: LightningInput {
+                amount: old_state.contract.commitment.amount,
+                witness: Witness::Incoming(old_state.contract.contract_key(), agg_decryption_key),
```
m1sterc001guy, line 227: Definitely worth adding a comment here describing what this is doing.

gateway/ln-gateway/src/gateway_module_ng/send_sm.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+pub struct SendStateMachine {
+    pub operation_id: OperationId,
+    pub contract: OutgoingContract,
+    pub state: SendSMState,
+}
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+pub enum SendSMState {
```
m1sterc001guy, line 29: Same comment as above to add Mermaid diagram

gateway/ln-gateway/src/gateway_module_ng/send_sm.rs
```diff
@@ -0,0 +1,183 @@
...
+    Sending(Sending),
+    Claiming([u8; 32], Vec<OutPoint>),
```
m1sterc001guy, line 31: Let's use named fields

gateway/ln-gateway/src/gateway_module_ng/send_sm.rs
```diff
@@ -0,0 +1,183 @@
...
+        context: GatewayClientContextNG,
+        max_delay: u64,
+        max_fee_msat: u64,
+        invoice: Bolt11Invoice,
+    ) -> Result<[u8; 32], ()> {
+        let lightning_context = loop {
+            match context.gateway.get_lightning_context().await {
+                Ok(lightning_context) => break lightning_context,
+                Err(_) => {
+                    sleep(std::time::Duration::from_secs(1)).await;
```
m1sterc001guy, line 97: I don't think we want to sleep forever here. We're sleeping just in case there's some sort of transient error in connecting to the lightning node right? I think it might just be simpler to fail and request that the client tries the payment again.

gateway/ln-gateway/src/gateway_module_ng/send_sm.rs
```diff
@@ -0,0 +1,183 @@
...
+                }
+            }
+        };
+
+        if lightning_context.lightning_public_key == invoice.recover_payee_pub_key() {
```
m1sterc001guy, line 102: Ok so this is checking if the gateway's lightning node is the same as the lightning node that generated the invoice, then we can do an internal swap and just start the receive state machine. A comment here would help.

Will this still work if there are > 1 gateways connected to the same lightning node? (we don't have explicit support for this today, but we might want to in the future for custodial payment providers for instance).

gateway/ln-gateway/src/gateway_module_ng/send_sm.rs
```diff
@@ -0,0 +1,183 @@
...
+                Ok(preimage) => Ok(preimage),
+                Err(e) => {
+                    error!("Failed to route internal payment {:?}", e);
+
+                    Err(())
+                }
+            };
+        }
+
+        match lightning_context
```
m1sterc001guy, line 124: In the legacy payment state machine, we check a bunch of parameters before actually making the payment in `await_get_payment_parameters` and `validate_outgoing_account`. We will want to add similar checks here.

gateway/ln-gateway/src/gateway_module_ng/send_sm.rs
```diff
@@ -0,0 +1,183 @@
...
+                    invoice.payment_hash().into_inner(),
+                    Amount::from_msats(invoice_msats),
+                )
+                .await
+            {
+                Ok(preimage) => Ok(preimage),
+                Err(e) => {
+                    error!("Failed to route internal payment {:?}", e);
+
+                    Err(())
```
m1sterc001guy, line 119: We should add a real error type here and propagate up

gateway/ln-gateway/src/gateway_module_ng/send_sm.rs
```diff
@@ -0,0 +1,183 @@
...
+                }
+            };
+        }
+
+        match lightning_context
+            .lnrpc
+            .pay(PayInvoiceRequest {
```
m1sterc001guy, line 126: We will need to support `pay_private` as well

Cargo.toml
joschisan: So just to be clear, which option exactly?
fedimint-lnv2-server
fedimint-ln-v2-server
fedimint-ln-server-v2

I would assume the first as its the most consistent?

dpc: First seems good.

modules/fedimint-lnv2-tests/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "fedimint-lnv2-tests"
+version = "0.3.0-alpha"
+authors = ["The Fedimint Developers"]
+edition = "2021"
+description = "fedimint-ln-tests contains integration tests for the lightning module"
```
m1sterc001guy, line 6: nit: crate name in description isn't updated

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+
+        Ok(server.to_erased())
+    }
+
+    fn validate_config(
+        &self,
+        _identity: &PeerId,
+        _config: ServerModuleConfig,
+    ) -> anyhow::Result<()> {
+        Ok(())
```
m1sterc001guy, line 242: Should this verify the `SecretKeyShare`?

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+            network: config.network,
+        })
+    }
+}
+
+fn dealer_keygen(
+    threshold: usize,
+    keys: usize,
+) -> (AggregatePublicKey, Vec<PublicKeyShare>, Vec<SecretKeyShare>) {
+    let mut rng = OsRng; // FIXME: pass rng
```
m1sterc001guy, line 263: This should be an easy fix

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+
+    let pks = sks
+        .iter()
+        .map(|sk| PublicKeyShare((G1Projective::generator() * sk.0).to_affine()))
+        .collect();
+
+    (AggregatePublicKey(apk), pks, sks)
+}
+
+fn eval_polynomial(coefficients: &[Scalar], x: &Scalar) -> Scalar {
```
m1sterc001guy, line 280: Do we want this and `dealer_keygen` to be in the lightning server module crate? Maybe the `tpe` crate makes more sense?

modules/fedimint-lnv2-common/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    LightningInput,
+    LightningOutput,
+    LightningOutputOutcome,
+    LightningConsensusItem,
+    LightningInputError,
+    LightningOutputError
+);
+
+#[derive(Debug, Clone)]
+pub struct LightningClientContext {
```
m1sterc001guy, line 160: This should be moved into the client crate

modules/fedimint-lnv2-common/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    #[error("The aggregate decryption key is invalid")]
+    InvalidDecryptionKey,
+    #[error("The forfeit signature is invalid")]
+    InvalidForfeitSignature,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Error, Encodable, Decodable)]
+pub enum LightningOutputError {
+    #[error("The contract is invalid")]
+    InvalidContract,
```
m1sterc001guy, line 113: Can we add the contract id to each of these errors for debugging purposes?

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+        out_point: OutPoint,
+    ) -> Result<TransactionItemAmount, LightningOutputError> {
+        let outcome = match output {
+            LightningOutput::Outgoing(contract) => {
+                if dbtx
+                    .insert_entry(&OutgoingContractKey(contract.contract_id()), contract)
+                    .await
+                    .is_some()
+                {
+                    return Err(LightningOutputError::ContractAlreadyExists);
```
m1sterc001guy, line 429: As mentioned above, lets add the `contract_id` for debugging purposes here and below

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+                let contract = dbtx
+                    .remove_entry(&IncomingContractKey(*contract_id))
+                    .await
+                    .ok_or(LightningInputError::UnknownContract)?;
+
+                if !contract
+                    .verify_agg_decryption_key(&self.cfg.consensus.tpe_agg_pk, agg_decryption_key)
+                {
+                    return Err(LightningInputError::InvalidDecryptionKey);
+                }
```
m1sterc001guy, line 398: Do we need to check contract expiry here too? I only see `consensus_unix_time` being checked in `process_output`, just trying to understand.

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+
+    async fn await_incoming_contract(
+        &self,
+        db: Database,
+        contract_id: ContractId,
+        expiration: u64,
+    ) -> Option<ContractId> {
+        loop {
+            timeout(
+                Duration::from_secs(60 * 10),
```
m1sterc001guy, line 607: Is the client expected to try again if this timeout hits?

m1sterc001guy: Oh nevermind, I see it will loop

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    async fn await_incoming_contract(
+        &self,
+        db: Database,
+        contract_id: ContractId,
+        expiration: u64,
+    ) -> Option<ContractId> {
+        loop {
+            timeout(
```
m1sterc001guy, line 606: We should have a log statement here in case this gets stuck for a long time.

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    async fn await_preimage(
+        &self,
+        db: Database,
+        contract_id: ContractId,
+        expiration: u64,
+    ) -> Option<[u8; 32]> {
+        loop {
+            timeout(
```
m1sterc001guy, line 634: Same comment as above for the log statement.

modules/fedimint-lnv2-common/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+impl std::fmt::Display for LightningOutputOutcome {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "LightningOutputOutcome")
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Error, Encodable, Decodable)]
+pub enum LightningInputError {
+    #[error("No contract found for given ContractId")]
+    UnknownContract,
```
m1sterc001guy, line 97: Add contract id here too

modules/fedimint-lnv2-common/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    #[error("The preimage is invalid")]
+    InvalidPreimage,
+    #[error("The contracts locktime has passed")]
+    Expired,
+    #[error("The contracts locktime has not yet passed")]
+    NotExpired,
+    #[error("The aggregate decryption key is invalid")]
+    InvalidDecryptionKey,
```
m1sterc001guy, line 105: Printing out the hex of the key would probably be helpful

modules/fedimint-lnv2-common/src/contracts.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub expiration: u64,
+    pub claim_pk: PublicKey,
+    pub refund_pk: PublicKey,
+    pub ephemeral_pk: PublicKey,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
+pub struct IncomingContract {
+    pub commitment: Commitment,
+    pub ciphertext: CipherText,
```
m1sterc001guy, line 28: Should we rename this to `encrypted_preimage`?

modules/fedimint-lnv2-common/src/contracts.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+impl IncomingContract {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        agg_pk: AggregatePublicKey,
+        encryption_seed: [u8; 32],
+        preimage: [u8; 32],
+        amount: Amount,
+        expiration: u64,
```
m1sterc001guy, line 38: Is this in blocks or unix timestamp? Just changing the name would be helpful

modules/fedimint-lnv2-common/src/contracts.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub invoice_hash: sha256::Hash,
+}
+
+impl OutgoingContract {
+    pub fn contract_id(&self) -> ContractId {
+        ContractId(self.consensus_hash())
+    }
+
+    pub fn forfeit_message(&self) -> Message {
+        Message::from(self.contract_id().0)
```
m1sterc001guy, line 137: Can we use a named field here? I forget what this is

m1sterc001guy: Oh nevermind, I see its just the hash inside the `ContractId`

modules/fedimint-lnv2-common/src/contracts.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    pub fn verify_preimage(&self, preimage: &[u8; 32]) -> bool {
+        preimage.consensus_hash::<sha256::Hash>() == self.payment_hash
+    }
+
+    pub fn verify_forfeit_signature(&self, signature: &Signature) -> bool {
```
m1sterc001guy, line 144: nit: you could return `Result` here and just to ? when calling instead of checking a boolean

modules/fedimint-lnv2-common/src/config.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct LightningConfigConsensus {
+    pub tpe_agg_pk: AggregatePublicKey,
+    pub tpe_pks: BTreeMap<PeerId, PublicKeyShare>,
+    pub fee_consensus: FeeConsensus,
+    pub network: Network,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LightningConfigPrivate {
+    // TODO: propose serde(with = "…") based protection upstream instead
```
m1sterc001guy, line 63: I don't understand this comment

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+
+/// Default expiration time for lightning invoices
+const INVOICE_EXPIRATION_SECONDS_DEFAULT: u32 = 3600;
+
+/// The high-level state of an payment operation over lightning
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SendState {
+    Funding,
+    Funded,
```
m1sterc001guy, line 59: I think one of these states should emit the `block_height` at which the outgoing contract is redeemable. I believe wallets use that today to show in the UI

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+
+    pub fn incoming_contract_amount(&self, msats: u64) -> Amount {
+        Amount::from_msats(msats.saturating_sub(self.receive.fee(msats)))
+    }
+}
+
+#[derive(
+    Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Decodable, Encodable,
+)]
+pub struct PaymentFee {
```
m1sterc001guy, line 118: Is there a specific reason not to use `RoutingFees` from the `lightning` crate?

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+        .consensus_hash::<sha256::Hash>()
+        .into_inner();
+
+    (ephemeral_tweak, ephemeral_keypair.public_key())
+}
+
+impl LightningClientModule {
+    pub async fn payment_info(&self, gateway_api: SafeUrl) -> anyhow::Result<PaymentInfo> {
+        reqwest::Client::new()
+            .post(
```
m1sterc001guy, line 249: This can be a `GET`

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+                    .join("payment_info")
+                    .expect("'payment_info' contains no invalid characters for a URL")
+                    .as_str(),
+            )
+            .json(&self.federation_id)
+            .send()
+            .await?
+            .json::<Option<PaymentInfo>>()
+            .await?
+            .ok_or(anyhow!("The gateway does not support our federation"))
```
m1sterc001guy, line 260: nit: `The Gateway does not support LightningV2`

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+            max_fee >= payment_info.payment_fees.send,
+            "The gateway send fee is above {max_fee:?}"
+        );
+
+        // we double it to account for cltv expiry deltas along the lightning route
+        let min_outgoing_cltv_delta = 2 * payment_info.outgoing_cltv_delta;
+
+        ensure!(
+            min_outgoing_cltv_delta <= expiration_delta,
+            "The minimum possible expiration delta is {min_outgoing_cltv_delta} blocks"
```
m1sterc001guy, line 305: Do we need to fail here? Couldn't we just use `expiration_delta` if we deem `min_outgoing_cltv_delta` is too low? Would also help to modify the error message to indicate this is a gateway LN node config issue, since clients could hit this and wonder why their payments are not working.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+            payment_hash: *invoice.payment_hash(),
+            amount: contract_amount,
+            expiration: consensus_block_count + expiration_delta,
+            claim_pk: payment_info.public_key,
+            refund_pk: refund_keypair.public_key(),
+            ephemeral_pk,
+            invoice_hash: invoice.consensus_hash(),
+        };
+
+        let operation_id = OperationId(contract.consensus_hash::<sha256::Hash>().into_inner());
```
m1sterc001guy, line 332: In the legacy client we store a `PaymentResult` that keeps track of successfully paid invoices, so that we don't try to pay the same invoice twice. I think we will need something like that here too.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+                loop {
+                    if let Some(LightningClientStateMachines::Send(state)) = stream.next().await {
+                        match state.state {
+                            SendSMState::Funding(..) => yield SendState::Funding,
+                            SendSMState::Funded(..) => yield SendState::Funded,
+                            SendSMState::Success(preimage) => {
+                                // the preimage has been verified by the state machine previously
+                                assert!(state.contract.verify_preimage(&preimage));
+
+                                yield SendState::Success(preimage);
```
m1sterc001guy, line 385: In legacy send, we wait for change since we might be paying an invoice with ecash notes that we can't make an exact amount of. Do we need to do that here too?

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+                                            &state.contract.contract_id(),
+                                            0
+                                        ).await {
+                                            if state.contract.verify_preimage(&preimage) {
+                                                yield SendState::Success(preimage);
+                                                return;
+                                            }
+                                        }
+
+                                        yield SendState::Failure;
```
m1sterc001guy, line 410: We should probably include an error message on this too. This should never happen right?

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+    ) -> Option<TransactionItemAmount> {
+        Some(TransactionItemAmount {
+            amount: output.amount(),
+            fee: self.cfg.fee_consensus.contract_output,
+        })
+    }
+}
+
+fn generate_ephemeral_tweak(static_pk: PublicKey) -> ([u8; 32], PublicKey) {
+    let ephemeral_keypair = KeyPair::new(secp256k1::SECP256K1, &mut rand::thread_rng());
```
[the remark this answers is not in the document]
m1sterc001guy: Ah I see that the ephemeral public key is stored in the contract, that should be sufficient to recover

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+
+        let operation_id = self
+            .receive_external_contract(contract)
+            .await
+            .expect("The contract has been generated with our public key");
+
+        Ok((invoice, operation_id))
+    }
+
+    pub async fn create_contract_and_fetch_invoice(
```
m1sterc001guy, line 468: Can this function and the one below be made private or do you anticipate higher layers calling this function?

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+            operation_id,
+            contract,
+            state: ReceiveSMState::Pending(Pending {
+                claim_keypair,
+                agg_decryption_key,
+            }),
+        });
+
+        // this may only fail if the operation id is already in use, in which case we
+        // ignore the error such that the method is idempotent
```
m1sterc001guy, line 590: Instead of doing this, maybe we should just check if the operation id is already in use before recovering the incoming contract keys

modules/fedimint-lnv2-common/src/api.rs
```diff
@@ -0,0 +1,87 @@
+use std::time::Duration;
```
m1sterc001guy, line 1: This file can move to the client crate

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: Saves us some crypto operations

modules/fedimint-lnv2-client/src/send_sm.rs
```diff
@@ -0,0 +1,338 @@
...
+                    .as_str(),
+            )
+            .json(&SendPaymentPayload {
+                federation_id,
+                contract,
+                invoice,
+            })
+            .send()
+            .await?
+            .json::<Result<Result<[u8; 32], Signature>, String>>()
```
m1sterc001guy, line 225: Maybe we can newtype `Result<[u8; 32], Signature>`. Seeing this return a Result of a Result is a little weird

m1sterc001guy: (but it makes sense knowing the the forfeit signature is returned on failure)

modules/fedimint-lnv2-client/src/send_sm.rs
```diff
@@ -0,0 +1,338 @@
...
+                        } else {
+                            error!("Invalid gateway response: {gateway_response:?}");
+                        }
+                    }
+                    Err(error) => {
+                        error!("Gateway returned error: {error}")
+                    }
+                },
+                Err(error) => {
+                    error!("Error while trying to reach gateway: {error}");
```
m1sterc001guy, line 197: Just a thought: we might want to retry faster if we receive a 404 error code from the gateway vs a different error code (we do this in the legacy pay in this PR https://github.com/fedimint/fedimint/pull/4024/files#diff-7e39423fc03d68d7d9b2c7f7799175ee71df89d5043d053b77f7d8067228bf42)

modules/fedimint-lnv2-client/src/send_sm.rs
```diff
@@ -0,0 +1,338 @@
...
+                state: SendSMState::Success(preimage),
+            };
+        }
+
+        let funded = match old_state.state {
+            SendSMState::Funded(funded) => funded,
+            _ => panic!("Unexpected prior state"),
+        };
+
+        let client_input = ClientInput::<LightningInput, LightningClientStateMachines> {
```
m1sterc001guy, line 317: Perhaps make this and the logic below into a function, I see it is used above when the gateway returns a forfeit signature as well.

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -311,3 +317,30 @@ async fn get_gateway_id(
 ) -> Result<impl IntoResponse, GatewayError> {
     Ok(Json(json!(gateway.gateway_id)))
 }
+
+async fn payment_info_ng(
```
m1sterc001guy, line 321: nit: remove ng

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -137,7 +139,11 @@ fn v1_routes(gateway: Gateway) -> Router {
     // Public routes on gateway webserver
     let public_routes = Router::new()
         .route("/pay_invoice", post(pay_invoice))
-        .route("/id", get(get_gateway_id));
+        .route("/id", get(get_gateway_id))
+        // These routes are for next generation lightning
+        .route("/payment_info", post(payment_info_ng))
+        .route("/send_payment", post(send_payment_ng))
```
m1sterc001guy, line 145: This is logically the same operation as `pay_invoice`. Should we use `pay_invoice_v2` to be consistent?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1617,3 +1860,57 @@ impl Display for PrettyInterceptHtlcRequest<'_> {
         )
     }
 }
+
+#[derive(Error, Debug, Clone, Eq, PartialEq)]
+pub enum SendPaymentError {
+    #[error("The federation id is unknown")]
+    UnknownFederationId,
```
m1sterc001guy, line 1867: We need to include additional debugging information on all of these errors. i.e federation id, contract, id, invoice hash, timeouts, etc

gateway/ln-gateway/src/lib.rs
```diff
@@ -113,6 +125,8 @@ pub const DEFAULT_FEES: RoutingFees = RoutingFees {
     proportional_millionths: 10000,
 };
 
+const OUTGOING_CLTV_DELTA_NG: u64 = 144;
```
m1sterc001guy, line 128: nit: remove NG

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,234 @@ pub(crate) async fn fetch_lightning_node_info(
     Ok((node_pub_key, alias, network))
 }
 
+impl Gateway {
+    async fn public_key(&self, federation_id: &FederationId) -> Option<PublicKey> {
```
m1sterc001guy, line 1564: Can we rename this to `gateway_client_public_key`? Or `gateway_redeem_key`? Right now the gateway also has a public key for the lightning node and a public key for the `gateway_id`. Lots of public keys to keep track of!

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: Ok sorry I think I got confused earlier, we need to fail here since its not safe to use a cltv_delta that is lower than the gateway's. A better error message would be helpful here

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,234 @@ pub(crate) async fn fetch_lightning_node_info(
...
+
+        if client.operation_exists(operation_id).await {
+            return Ok(module.subscribe_send(operation_id, payload.contract).await);
+        }
+
+        if payload.contract.claim_pk != module.keypair.public_key() {
+            return Err(SendPaymentError::NotOurKey);
+        }
+
+        if payload.invoice.consensus_hash::<sha256::Hash>() != payload.contract.invoice_hash {
```
m1sterc001guy, line 1614: Invoice should already have the payment hash, we shouldn't need to hash it again

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,234 @@ pub(crate) async fn fetch_lightning_node_info(
...
+            .amount_milli_satoshis()
+            .ok_or(SendPaymentError::InvoiceMissingAmount)?;
+
+        let contract_amount = self.payment_fees().outgoing_contract_amount(invoice_msats);
+
+        if contract_amount != payload.contract.amount {
+            return Err(SendPaymentError::Unbalanced);
+        }
+
+        let max_fee_msat = (contract_amount.msats - invoice_msats) / 2;
```
m1sterc001guy, line 1641: Why are we dividing by 2?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,234 @@ pub(crate) async fn fetch_lightning_node_info(
...
+
+    async fn create_invoice(
+        &self,
+        payload: CreateInvoicePayload,
+    ) -> std::result::Result<Bolt11Invoice, CreateInvoiceError> {
+        if !payload.contract.verify() {
+            return Err(CreateInvoiceError::InvalidContract);
+        }
+
+        loop {
```
m1sterc001guy, line 1665: We can use `autocommit` here instead of manually looping

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,234 @@ pub(crate) async fn fetch_lightning_node_info(
...
+                payload.invoice,
+                payload.contract.clone(),
+            )
+            .await
+            .ok();
+
+        Ok(module.subscribe_send(operation_id, payload.contract).await)
+    }
+
+    async fn create_invoice(
```
m1sterc001guy, line 1657: nit: I'd prefer `register_invoice` here since this is just saving it to the db. Then `create_invoice_via_lnrpc` can be changed to `create_invoice`

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,234 @@ pub(crate) async fn fetch_lightning_node_info(
...
+                expiry: expiry_time,
+                description,
+            })
+            .await
+            .map_err(|e| e.to_string())?;
+
+        Bolt11Invoice::from_str(&response.invoice).map_err(|e| e.to_string())
+    }
+
+    pub async fn receive(
```
m1sterc001guy, line 1744: Is this called anywhere outside of tests yet?

fedimint-client/src/module/mod.rs
```diff
@@ -452,6 +452,42 @@ where
     pub fn get_internal_payment_markers(&self) -> anyhow::Result<(PublicKey, u64)> {
         self.client.get().get_internal_payment_markers()
     }
+
+    /// This method starts n state machines with given operation id without a
```
m1sterc001guy, line 456: typo: "n"

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
```diff
@@ -0,0 +1,202 @@
...
+        context: GatewayClientContextV2,
+        max_delay: u64,
+        max_fee_msat: u64,
+        invoice: Bolt11Invoice,
+    ) -> Result<[u8; 32], ()> {
+        let lightning_context = loop {
+            match context.gateway.get_lightning_context().await {
+                Ok(lightning_context) => break lightning_context,
+                Err(_) => {
+                    sleep(std::time::Duration::from_secs(1)).await;
```
m1sterc001guy, line 113: We should log here so that we know if a payment is stuck because the LN node is down

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
```diff
@@ -0,0 +1,202 @@
...
+                }
+            }
+        };
+
+        if lightning_context.lightning_public_key == invoice.recover_payee_pub_key() {
```
m1sterc001guy, line 118: I think this check assumes that a lightning node and a gateway are 1 to 1. This might not be true in the future.

gateway/ln-gateway/src/gateway_module_v2/receive_sm.rs
```diff
@@ -0,0 +1,255 @@
...
+        let verify_decryption_share = move |peer, outcome: SerdeOutputOutcome| {
+            let outcome = deserialize_outcome::<LightningOutputOutcome>(outcome, &module_decoder)?;
+
+            match outcome {
+                LightningOutputOutcome::Incoming(share) => {
+                    if !decryption_contract.verify_decryption_share(
+                        tpe_pks.get(&peer).ok_or(anyhow!("Unknown peer pk"))?,
+                        &share,
+                    ) {
+                        bail!("Invalid decryption share");
```
m1sterc001guy, line 158: Should we log the share here in case we need to debug this?

gateway/ln-gateway/src/gateway_module_v2/receive_sm.rs
```diff
@@ -0,0 +1,255 @@
...
+            .map(|(peer, share)| (peer.to_usize() as u64 + 1, share))
+            .collect();
+
+        let agg_decryption_key = aggregate_decryption_shares(&decryption_shares);
+
+        if !old_state
+            .contract
+            .verify_agg_decryption_key(&tpe_agg_pk, &agg_decryption_key)
+        {
+            // This implies that the lightning client config's public keys are inconsistent
```
m1sterc001guy, line 218: Need a error! log here in case this happens (and a error message on `Failure`)

modules/fedimint-lnv2-tests/tests/tests.rs
```diff
@@ -0,0 +1,267 @@
...
+    assert_eq!(client.get_balance().await, Amount::from_bitcoins(1));
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn pay_external_invoice() -> anyhow::Result<()> {
+    let fixtures = fixtures();
+    let fed = fixtures.new_fed().await;
+    let gw = gateway(&fixtures, &fed).await;
+
+    let cln = fixtures.cln().await;
```
m1sterc001guy, line 70: We will eventually want a loop here that tests both gateways

m1sterc001guy:
Ok I think I made it through all of the code, nice work! :) 

I believe most of my comments are nits or around naming, or debugging suggestions as I tried to read through and understand things. IMO here are the outstanding bigger TODOs that I saw, I'm curious what other's opinions are on what we should try to shoot for before merging. I think I'd prefer to handle most of the nits and then tackle these bigger items in follow ups.

- Test coverage of gateway state machines and other tests that the legacy has (maybe as follow up we should copy the relevant tests over, similar to how we did for aleph).
- Internal payments require the presence of the gateway. This makes it so even for internal payments, the gateway can charge a fee. I believe this also requires a gateway and a lightning node to be 1-1, since we check for an internal payment by looking at the lightning node pubkey. This means that it will be hard to use custodial services (alby, zbd, coinos, etc) as liquidity providers for gateways.
- Create Invoice/Integration with HTLC interception. Haven't implemented CLN yet, the plugin will need modifying. I've been looking into this but I'm behind :)

crypto/tpe/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "fedimint-tpe"
+version = "0.3.0-alpha"
```
m1sterc001guy, line 3: So we don't forget: this needs to change to 0.4.0-alpha

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,236 @@ pub(crate) async fn fetch_lightning_node_info(
...
+    pub fn payment_fees(&self) -> PaymentFees {
+        PaymentFees {
+            // we take a fee of one percent for outgoing contracts
+            send: PaymentFee::default(),
+            // we take a fee of one percent for incoming contracts
+            receive: PaymentFee::default(),
+        }
+    }
+
+    async fn send_payment(
```
elsirion, line 1591: Should be named differently to avoid confusion. We won't be using LNv2 by default for a while due to upgradability limitations.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,236 @@ pub(crate) async fn fetch_lightning_node_info(
...
+                payload.invoice,
+                payload.contract.clone(),
+            )
+            .await
+            .ok();
+
+        Ok(module.subscribe_send(operation_id, payload.contract).await)
+    }
+
+    async fn create_invoice(
```
elsirion, line 1658: Same naming concerns here. Could these methos maybe even be in an extension trait that isn't in scope by default?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -137,7 +139,11 @@ fn v1_routes(gateway: Gateway) -> Router {
     // Public routes on gateway webserver
     let public_routes = Router::new()
         .route("/pay_invoice", post(pay_invoice))
-        .route("/id", get(get_gateway_id));
+        .route("/id", get(get_gateway_id))
+        // These routes are for next generation lightning
```
elsirion, line 143: I assume that means we don't provide any stability guarantees for these, so don't need to bump the API version.

elsirion (approved):
I only reviewed the parts that touch existing code. I think this shouldn't break anything existing (and if it does we now have some time till 0.4 to iron it out).

For the LNv2 code I'll just go by @m1sterc001guy's and @joschisan's judgement. It'll have to prove itself over the next 1-2 releases before it will become the default imo.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -0,0 +1,744 @@
...
+        funding_change_outpoints: Vec<OutPoint>,
+        gateway_api: SafeUrl,
+        contract: OutgoingContract,
+        invoice: Bolt11Invoice,
+    },
+    Receive {
+        operation_id: OperationId,
+        contract: IncomingContract,
+    },
+}
```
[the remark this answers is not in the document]
maan2003: we use it, but there is lot of time before we use lnv2

elsirion (approved):

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
m1sterc001guy: Can we add a comment here?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1545,6 +1560,236 @@ pub(crate) async fn fetch_lightning_node_info(
     Ok((node_pub_key, alias, network))
 }
 
+impl Gateway {
+    async fn public_key_v2(&self, federation_id: &FederationId) -> Option<PublicKey> {
```
m1sterc001guy, line 1564: I would prefer `gateway_claim_public_key` or something. There's so many public keys, this is going to confuse me

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Also, instead of looping, you can just return `HashAlreadyRegistered` if `commit_tx_result` returns an error. On the second loop its going to return that anyway, it will be nice to get rid of the loop here.

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: @joschisan would be good to address this before merging

m1sterc001guy (approved):
IMO looks good, we should merge to avoid rebase conflicts.

Documented follow ups here: https://github.com/fedimint/fedimint/issues/3467#issuecomment-2043530651

PR #3470, opened 2023-10-27
fix: don't limit max number of concurrent requests

douglaz:


elsirion:
Can you PR to master first and then backport please (or if done already I haven't caught up yet, but the original PR should be linked here)? It's too easy to lose track otherwise.

douglaz:
> Can you PR to master first and then backport please (or if done already I haven't caught up yet, but the original PR should be linked here)? It's too easy to lose track otherwise.

This was on purpose, I didn't want to include this on master/0.1 without testing it first

justinmoon:
Call: let's target master with this PR, first

elsirion:
Since #3497 is on master now, let's also merge this one.

elsirion (approved):

PR #3455, opened 2023-10-25
feat: gateway checks network on connect federation

okjodom:
Adds Network configuration to gatewayd.

Speccd' and implemented behavior:
- gatewayd starts off without a network configuration
- on `set_configuration`, gatewayd assumes network configuration if provides, or defaults to regtest
- `set_configuration` can be used to change the network config when gatewayd is in running state, provided a federation has not been connected. This prevents changing the configured network to a different setting from the connected federations

Gatewayd the only allows (re)connecting a federations whose lightning module is on the same network

Tasks:
- [x] Gatewayd should support configuring of a network
- [x] Gateway should expose it's configured network via info api
- [x] Gatewayd should ensure all connected federations are on the same configured network
- [x] Gatewayd should ensure connected lightning node / service is on the same configured network : #3488

PR #3488, opened 2023-10-30
feat: match node network to the gateway configuration

okjodom:
Follow up to #3455


m1sterc001guy:
Maybe I should just wait until https://github.com/fedimint/fedimint/pull/3455 lands, but what is new in this PR?

justinmoon:
Generally looks good to me. Didn't review too closely.

okjodom:
> Maybe I should just wait until #3455 lands, but what is new in this PR?

@m1sterc001guy , this ensures the connected lightning node is on the same network as gatewayd and the federations. Complements the network configuration that is introduced by #3455

gateway/ln-gateway/src/lib.rs
```diff
@@ -842,13 +850,24 @@ impl Gateway {
...
+                lightning_network, ..
+            } => {
+                if network.is_some() && network != Some(lightning_network) {
+                    return Err(GatewayError::GatewayConfigurationError(
+                        "Cannot change network while connected to a lightning node".to_string(),
+                    ));
+                }
+                lightning_network
+            }
+            _ => DEFAULT_NETWORK,
```
m1sterc001guy, line 868: Let's add a comment here that explains that, if the gateway is not in the `Running` state, the network will be set later once the connection to the lightning node is established.

m1sterc001guy (dismissed):
Some small comments :)

gateway/ln-gateway/src/lib.rs
okjodom: 💯

m1sterc001guy (approved):

Issue #3478, opened 2023-10-28
Gateway fails to run state machines until it's restarted


justinmoon:
We are running a federation with `v0.1.3` release. It is in a region with pretty poor internet connectivity in general. Everything was working smoothly for a few days. Then all-of-a-sudden lightning payments started to fail. If you generated an invoice using `fedimint-cli`, and called `fedimint-cli wait-invoice` it would get to the `WaitingForPayment` state and get stuck there.

From the gateway logs ([which are pretty sparse](https://github.com/fedimint/fedimint/issues/3477)) it seemed like the gateway's `handle_htlc_stream` was seeing the HTLC, and [this database commit](https://github.com/fedimint/fedimint/blob/master/fedimint-client/src/lib.rs#L632-L639) was being hit.

But the gateway's fedimint transaction buying the preimage from the federation was never being confirmed by the federation. In fact, it seemed like the `submit_transaction` RPC method wasn't getting called because [this log](https://github.com/fedimint/fedimint/blob/4943c209c1ba02f8e5b92d2db7f636373fe24b48/fedimint-server/src/net/api.rs#L209) wasn't showing up. But if we restarted the gateway, both that log and the transaction itself would show up. The `fedimint-cli wait-invoice...` command which was previously stuck on `WaitingForPayment` also promptly completed.

We thought the root cause was jsonrpsee slots getting exhausted so we tried https://github.com/fedimint/fedimint/pull/3471 and https://github.com/fedimint/fedimint/pull/3470 but neither worked.

We decided to spin this gateway down and setup another to get things working again. In the process of withdrawing funds, the `withdrawal` command from `gateway-cli` also failed to complete until the gateway was restarted which suggests the problem is somewhere in the state machine executor.

justinmoon:
Call: we need better tooling to inspect state machines when debugging things like this

douglaz:
One of the problems is due to payment taking too long. Currently we have this:
```rust
impl GatewayPayInvoice {
    fn transitions(
        &self,
        global_context: DynGlobalClientContext,
        context: GatewayClientContext,
        common: GatewayPayCommon,
    ) -> Vec<StateTransition<GatewayPayStateMachine>> {
        vec![StateTransition::new(
            Self::await_get_payment_parameters(global_context, self.contract_id, context.clone()),
            move |_dbtx, result, _old_state| {
                Box::pin(Self::transition_buy_preimage(
                    context.clone(),
                    result,
                    common.clone(),
                ))
            },
[...]

PR #3492, opened 2023-10-31
DO NOT MERGE: feat(dbtool): dump state machines

elsirion:
Adds a tool to debug state machines from gateway DB dumps (to debug #3478):

```
Lists state machine states from a client database in the format: module instance | active | creation time | state debug print

Usage: fedimint-dbtool --database <DATABASE> list-states [OPTIONS]

Options:
      --active                 List active states
      --inactive               List inactive states
      --pretty                 Print the state debug output on multiple lines
      --operation <OPERATION>  Only show states belonging to operation
      --instance <INSTANCE>    Only show states belonging to this module instance
  -h, --help                   Print help
```

E.g. use in combination with `less -S` to get the following, filtered output:

```
fedimint-dbtool --database <path-to-db> list-states --active --inactive --operation fd64d9c190f180f13eead2bb3361f5bbc618bd727f84c7cd90843dced962c570 | less -S
```

2023-10-31-193552_5120x1441_scrot

TODO:
- [ ] Port to master
- [ ] Add mode for user clients
- [ ] Maybe also add `Serialize` to `DynState` so that the output can be JSON for further filtering

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,7 +70,7 @@ dependencies = [
  "derivative",
  "futures",
  "futures-timer",
- "itertools",
+ "itertools 0.10.5",
  "log",
  "parity-scale-codec",
  "parking_lot 0.12.1",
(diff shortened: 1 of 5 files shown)


elsirion:
I want to change the DB format for SMs anyway, allowing for a better version of this tool.

PR #3495, opened 2023-10-31
fix: let load-test-tool properly load the wallet from archive dir

douglaz:
If both an invite code and an archive dir is given, the tool will fail the second time with
```
Error: Alternative config source provided but config was found in DB
```

So if the persisted wallet exists, we don't need to provide the invite code to the client builder.

Plus provide context on some file opening.


elsirion (approved):
We should re-think the client builder API if it causes such pain :/

PR #3497, opened 2023-11-01
fix: increase limit of max number of concurrent requests

elsirion:
Originally by @douglaz as #3470 to 0.1, but we want to bring it to master first.


douglaz (approved):
LGTM

PR #3498, opened 2023-11-01
chore(docker): use 0.1 docker tags for ui images to always use latest 0.1.x release

wbobeirne:
Updates all the docker configs to use `0.1` instead of `0.1.x` so that when people run these, they're always running the latest version of the 0.1 UI. The one downside here is that if you've pulled an old image (0.1.3) and we push a new one (0.1.4) I _think_ by default Docker will just use what it has around, you'd need to run `docker-compose pull` to grab the upgraded image.

I'll let @douglaz decide which he'd prefer, otherwise I can just update all of them to use 0.1.4.

Also, fixed some that were still trying to use non-existent `v` prefixed tags.

### DIFF
--- a/docker/fedimintd-mutinynet/docker-compose.yaml
+++ b/docker/fedimintd-mutinynet/docker-compose.yaml
@@ -23,7 +23,7 @@ services:
     platform: linux/amd64
 
   guardian-ui:
-    image: fedimintui/guardian-ui:v0.1.1
+    image: fedimintui/guardian-ui:v0.1
     ports:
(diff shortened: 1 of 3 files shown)


elsirion (approved):
I think this is the right approach, but the update procedures should be documented.

elsirion:
Maybe add a README here: https://github.com/fedimint/fedimint/tree/c942b085167a429d10cc37f8cfa6ec62c1c3e805/docker/fedimintd-mutinynet

douglaz (approved):

Issue #3333, opened 2023-10-05
Re-enable federations of fewer than 4 guardians


justinmoon:
https://github.com/fedimint/fedimint/pull/3313 introduced initial support for AlephBFT. One simplification we made to just complete the switch was removing the ability for federations of fewer than 4 guardians.

This issue is to add back this functionality.

justinmoon:
My understanding is that we won't support 2/2 or 3/3 federations @joschisan @elsirion?

joschisan:
Correct

PR #3505, opened 2023-11-02
feat: enable single guardians

joschisan:
Closes https://github.com/fedimint/fedimint/issues/3333


elsirion (approved):
We still need tests, see #3517, but will merge to not produce conflicts.

PR #3512, opened 2023-11-03
fix: remove panics from wallet consensus_proposal

### LINKED ISSUE #3266: Wallet module server shouldn't panic if bitcoin RPC fails
* https://github.com/fedimint/fedimint/blob/1a5e1f0a30ac975433e26152098a8bc37518d797/modules/fedimint-wallet-server/src/lib.rs#L328
* https://github.com/fedimint/fedimint/blob/1a5e1f0a30ac975433e26152098a8bc37518d797/modules/fedimint-wallet-server/src/lib.rs#L347

bradleystachurski:
Fixes #3266 

- Gracefully handle btc_rpc errors instead of panicking
- Log the associated error and return a sane fallback

### DIFF
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -69,7 +69,7 @@ use miniscript::{translate_hash_fail, Descriptor, TranslatePk};
 use rand::rngs::OsRng;
 use secp256k1::{Message, Scalar};
 use strum::IntoEnumIterator;
-use tracing::{debug, info, instrument, trace, warn};
+use tracing::{debug, error, info, instrument, trace, warn};
 
 #[derive(Debug, Clone)]
 pub struct WalletGen;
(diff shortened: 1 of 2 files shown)


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -320,8 +320,14 @@ impl ServerModule for Wallet {
...
-        // TODO: We should not be panicking
-        let block_count = self.get_block_count().await.expect("bitcoind rpc failed");
+        // If we are unable to get a block count from the node, we default to 0 to skip
+        // adding a block count vote to consensus items.
+        //
+        // The potential impact of not including the latest block count from our peer's
+        // node is delayed processing of change outputs for the federation, which is
+        // an acceptable risk since subsequent rounds of consensus will reattempt to
+        // fetch the latest block count.
+        let block_count = self.get_block_count().await.unwrap_or(0);
```
elsirion, line 330: Can't we propose our `current_vote` again (see L339)? That seems safer.

elsirion (commented):
Thx for the PR, looking good!

We have similar logic for determining block height in the LN module, could you please check if the same fixes are applicable there too?

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -861,11 +867,16 @@ impl Wallet {
...
     }
 
-    pub async fn get_fee_rate(&self) -> anyhow::Result<Feerate> {
-        Ok(self
-            .get_fee_rate_opt()
-            .await?
-            .unwrap_or(self.cfg.consensus.default_fee))
+    /// Returns the fee rate from Bitcoind or the default if the node is
+    /// unreachable or isn't able to provide an estimate
+    pub async fn get_fee_rate(&self) -> Feerate {
```
dpc, line 872: IMO, we should return `Option` here, and on error a `None`, which would make consensus code just not change the vote at all (use previous one). Otherwise we might get into a lot of noise and choppy voting - with fee rate bouncing on every intermittent error etc.

modules/fedimint-wallet-server/src/lib.rs
bradleystachurski: I've updated so we use our `current_vote` instead of 0 if we're unable to get a block count from the node, but this will still have the same effect of skipping including the block count vote to consensus items.

If we propose our `current_vote` again, we'll cause `process_consensus_item` to [`bail!`](https://github.com/fedimint/fedimint/blob/5f533fefc82355a42ac0b0c3c8cdc6da47700754/modules/fedimint-wallet-server/src/lib.rs#L382).

Calls to `process_consensus_item` appear to be idempotent when constructing a block in `complete_signed_block` using the [`unit_data_receiver`](https://github.com/fedimint/fedimint/blob/de9889bfd524f9bcb52c6e2e249701f42366a494/fedimint-server/src/consensus/server.rs#L370), however I want to flag that we will likely get unnecessary error logs since `consensus_proposal` may be called multiple times before observing a new block.

Let me know if you'd still like to propose our `current_vote` again, i.e. change

```rust
        if current_vote < block_count_proposal {
```

to

```rust
        if current_vote <= block_count_proposal {
```

Also, let me know if I'm misinterpreting your question 😃

modules/fedimint-wallet-server/src/lib.rs
bradleystachurski: Good call. I removed the existing function since it's only used in `consensus_proposal` and the abstraction was a bit awkward (`get_fee_rate_or_default_or_none_on_error`?). This seems much easier to read.

bradleystachurski:
> We have similar logic for determining block height in the LN module, could you please check if the same fixes are applicable there too?

@elsirion  Ahh, thanks for highlighting! I made a similar change so we don't panic in the LN module.

modules/fedimint-wallet-server/src/lib.rs
elsirion: > If we propose our current_vote again, we'll cause process_consensus_item to [bail!](https://github.com/fedimint/fedimint/blob/5f533fefc82355a42ac0b0c3c8cdc6da47700754/modules/fedimint-wallet-server/src/lib.rs#L382).

But we aren't doing that because of the if condition further down as you noticed. I guess it would be cleaner to work with an `Option` instead? Both your original and my current solutions cause us to just not make a proposal :laughing: 

I wasn't thinking about actually on initial review, so making it explicit that we just not propose in this case would be good imo.

dpc (approved):

elsirion (approved):

PR #3515, opened 2023-11-03
feat: make federation info available before joining

### LINKED ISSUE #3514: Separate config fetching from building the client
There should be a struct that just contains the federation config and can be created from the join code, without initializing a full client. This can be used to:

* Preview Federation info
* Create a API client that talks to all the guardian servers
* Give it to a client builder to instantiate a cient

elsirion:
Fixes #3514

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -74,7 +74,7 @@ use std::sync::atomic::AtomicUsize;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
 
-use anyhow::{anyhow, bail, ensure};
+use anyhow::{anyhow, bail, ensure, Context};
 use async_stream::stream;
 use db::{
     CachedApiVersionSet, CachedApiVersionSetKey, ClientConfigKey, ClientConfigKeyPrefix,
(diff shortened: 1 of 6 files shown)


fedimint-client/src/lib.rs
```diff
@@ -1258,21 +1258,77 @@ impl TransactionUpdates {
...
+        Ok(FederationInfo {
+            config,
+            invite_code: None,
+        })
+    }
+
+    /// Returns the federations configuration
+    pub fn config(&self) -> &ClientConfig {
+        &self.config
+    }
```
elsirion, line 1294: Do we actually want to make that available or should there be functions for all the relevant fields? @wbobeirne @justinmoon

fedimint-client/src/lib.rs
```diff
@@ -1258,21 +1258,77 @@ impl TransactionUpdates {
...
+pub struct FederationInfo {
+    config: ClientConfig,
+    // TODO: make non-optional or remove
+    invite_code: Option<InviteCode>,
+}
+
+impl FederationInfo {
+    /// Download federation info using invitation code
+    pub async fn from_invite_code(invite: InviteCode) -> anyhow::Result<FederationInfo> {
+        let config = try_download_config(invite.clone(), 10).await?;
```
elsirion, line 1273: I just kept the old semantics, but maybe we should make the retry-behavior configurable?

fedimint-client/src/lib.rs
```diff
@@ -1258,21 +1258,77 @@ impl TransactionUpdates {
...
+            invite_code: None,
+        })
+    }
+
+    /// Returns the federations configuration
+    pub fn config(&self) -> &ClientConfig {
+        &self.config
+    }
+
+    /// If the federation info was created from
```
okjodom, line 1296: nit: dangling comment?

fedimint-client/src/lib.rs
okjodom: I previously thought the client builder would take config for retries. Makes sense for me

justinmoon: https://github.com/fedimint/fedimint/issues/3519

fedimint-client/src/lib.rs
justinmoon: I think just spitting up the whole config is best for now

justinmoon (approved):
LGTM

okjodom (dismissed):

fedimint-client/src/lib.rs
elsirion: Completed my comment

elsirion:
Oh, I didn't see that you already queued it :laughing:

Issue #3249, opened 2023-09-20
Lightning module client should re-attempt refund transaction when transaction submission fails


m1sterc001guy:
Today, it is possible a refund transaction fails to be submitted/accepted if the federation is unavailable. This transitions the state machine into a failed state and it will never attempt to re-submit the refund transaction, which can result in a loss of funds.

Ideally, the state machine should periodically retry (perhaps with exponential backup + some maximum) so that when the federation does come back online, the refund transaction is re-submitted.

If we don't want to support a potential infinite loop in the state machine framework, we could not support this and instead instruct users to recover their wallet, which should detect the escrowed funds and submit a refund transaction. This is worse user experience though, since it requires an extra manual step for users to get their money back.

elsirion:
Instance of #3173.

justinmoon:
Dev call: @m1sterc001guy will review this

justinmoon:
@m1sterc001guy will review this again :)

m1sterc001guy:
I believe this has been fixed.

PR #3516, opened 2023-11-03
Review and fix retry behavior of all state machines

### LINKED ISSUE #3173: Client sometimes fails successful transactions when offline
The new client has the goal to gracefully handle temporary failures like network interruptions, but I discovered at least one place where it can fail when going offline at the wrong time

Let's take the example of the state machine responsible for issuing new e-cash:

https://github.com/fedimint/fedimint/blob/ff3ad0658d4d16c2fa5ed30d91611f019374ab82/modules/fedimint-mint-client/src/output.rs#L162-L182

If *anything*, including network failure, goes wrong here the state machine goes into a failed state. What should happen instead is something similar to what the transaction submission state machine does: if the error is retryable wait for a bit and then try again.

https://github.com/fedimint/fedimint/blob/ff3ad0658d4d16c2fa5ed30d91611f019374ab82/fedimint-client/src/transaction/sm.rs#L196-L203

We need to check all state machines for similar mistakes to make sure everything works even with flaky internet connections.

elsirion:
Fixes #3173

Please double-check!

### DIFF
--- a/modules/fedimint-dummy-client/src/states.rs
+++ b/modules/fedimint-dummy-client/src/states.rs
@@ -2,17 +2,21 @@ use std::time::Duration;
 
 use fedimint_client::sm::{DynState, State, StateTransition};
 use fedimint_client::DynGlobalClientContext;
-use fedimint_core::api::GlobalFederationApi;
+use fedimint_core::api::{GlobalFederationApi, OutputOutcomeError};
 use fedimint_core::core::{Decoder, IntoDynInstance, ModuleInstanceId, OperationId};
 use fedimint_core::db::ModuleDatabaseTransaction;
 use fedimint_core::encoding::{Decodable, Encodable};
+use fedimint_core::task::sleep;
 use fedimint_core::{Amount, OutPoint, TransactionId};
 use fedimint_dummy_common::DummyOutputOutcome;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use tracing::debug;
 
 use crate::db::DummyClientFundsKeyV0;
 use crate::{get_funds, DummyClientContext};
+
+const RETRY_DELAY: Duration = Duration::from_secs(1);
 
 /// Tracks a transaction
 #[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
@@ -98,16 +102,31 @@ async fn await_dummy_output_outcome(
     outpoint: OutPoint,
     module_decoder: Decoder,
 ) -> Result<(), DummyError> {
-    global_context
-        .api()
-        .await_output_outcome::<DummyOutputOutcome>(
-            outpoint,
-            Duration::from_millis(i32::MAX as u64),
-            &module_decoder,
-        )
-        .await
-        .map_err(|_| DummyError::DummyInternalError)?;
-    Ok(())
+    loop {
+        match global_context
+            .api()
+            .await_output_outcome::<DummyOutputOutcome>(
+                outpoint,
+                Duration::from_millis(i32::MAX as u64),
+                &module_decoder,
+            )
+            .await
+        {
+            Ok(_) => {
+                return Ok(());
+            }
+            Err(OutputOutcomeError::Federation(e)) if e.is_retryable() => {
+                debug!(
+                    "Awaiting output outcome failed, retrying in {}s",
+                    RETRY_DELAY.as_secs_f64()
+                );
+                sleep(RETRY_DELAY).await;
+            }
+            Err(_) => {
+                return Err(DummyError::DummyInternalError);
+            }
+        }
+    }
 }
 
 // TODO: Boiler-plate
--- a/modules/fedimint-ln-client/src/pay.rs
+++ b/modules/fedimint-ln-client/src/pay.rs
@@ -5,7 +5,7 @@ use bitcoin_hashes::sha256;
 use fedimint_client::sm::{ClientSMDatabaseTransaction, State, StateTransition};
 use fedimint_client::transaction::ClientInput;
 use fedimint_client::DynGlobalClientContext;
-use fedimint_core::api::GlobalFederationApi;
+use fedimint_core::api::{GlobalFederationApi, OutputOutcomeError};
 use fedimint_core::config::FederationId;
 use fedimint_core::core::{Decoder, OperationId};
 use fedimint_core::encoding::{Decodable, Encodable};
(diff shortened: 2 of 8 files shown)


justinmoon:
Does it fix https://github.com/fedimint/fedimint/issues/3249 too?

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -258,6 +283,28 @@ impl LightningPayFunded {
     async fn gateway_pay_invoice(
         gateway: LightningGateway,
         payload: PayInvoicePayload,
+    ) -> Result<String, GatewayPayError> {
+        let mut last_error = None;
+
+        // If we don't succeed to reach the gateway in 30 seconds we'd rather abort the
+        // payment so the user isn't surprised by a heavily delayed payment
+        for _ in 0..30 {
```
justinmoon, line 291: One a slow network this could take a lot longer then 30 seconds. Would it be better to just have ` loop` wraped in a `timeout()`?

elsirion: Good idea!

modules/fedimint-dummy-client/src/states.rs
```diff
@@ -14,6 +15,8 @@ use thiserror::Error;
 use crate::db::DummyClientFundsKeyV0;
 use crate::{get_funds, DummyClientContext};
 
+const RESTRY_DELAY: Duration = Duration::from_secs(1);
```
douglaz, line 18:
```suggestion
const RETRY_DELAY: Duration = Duration::from_secs(1);
```

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -258,6 +287,40 @@ impl LightningPayFunded {
     async fn gateway_pay_invoice(
         gateway: LightningGateway,
         payload: PayInvoicePayload,
+    ) -> Result<String, GatewayPayError> {
+        // If we don't succeed to reach the gateway in 30 seconds we'd rather abort the
+        // payment so the user isn't surprised by a heavily delayed payment
```
douglaz, line 292: ChatGPT suggests to rewrite this as:
```suggestion
        // Abort the payment if we can't reach the gateway within 30 seconds
        // to prevent unexpected delays for the user.

```

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -258,6 +287,40 @@ impl LightningPayFunded {
...
+                    match response {
+                        Ok(preimage) => return Ok(preimage),
+                        Err(e) => {
+                            warn!("Error while trying to reach gateway: {}", e);
+                            last_error = Some(e);
+                            sleep(RETRY_DELAY).await;
+                        }
+                    }
+                },
+                () = sleep(duration_til_deadline) => {
```
douglaz, line 311: I'm not sure this abortion is correct, because we only abort on client side. It seems it's possible that the payment succeeds and we still returns a timeout

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -205,6 +207,8 @@ impl LightningReceiveConfirmedInvoice {
                     Some(_) => return Ok(incoming_contract_account),
                     None => return Err(LightningReceiveError::InvalidPreimage),
                 },
+                // FIXME: should we filter for retryable errors here to not swallow implementation
+                // bugs? (there exist more places like this)
                 Err(error) => {
                     error!("External LN payment error waiting for preimage decryption: {error:?}");
                 }
```
douglaz, line 214: Use a const below for better organization?
```rust
sleep(Duration::from_secs(RETRY_DELAY)).await;
```

modules/fedimint-ln-client/src/pay.rs
elsirion: Good catch!

modules/fedimint-dummy-client/src/states.rs
```diff
@@ -98,16 +101,27 @@ async fn await_dummy_output_outcome(
...
+                outpoint,
+                Duration::from_millis(i32::MAX as u64),
+                &module_decoder,
+            )
+            .await
+        {
+            Ok(_) => {
+                return Ok(());
+            }
+            Err(OutputOutcomeError::Federation(e)) if e.is_retryable() => {
```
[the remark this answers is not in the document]
elsirion: I did so for all newly introduced retry loops. #3530 will allow a more uniform logging strategy.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -258,6 +294,29 @@ impl LightningPayFunded {
...
         gateway: LightningGateway,
         payload: PayInvoicePayload,
+    ) -> Result<String, GatewayPayError> {
+        // Abort the payment if we can't reach the gateway within 30 seconds
+        // to prevent unexpected delays for the user.
+        let deadline = fedimint_core::time::now() + GATEWAY_API_TIMEOUT;
+
+        let mut last_error = None;
+        while fedimint_core::time::now() < deadline {
+            match Self::try_gateway_pay_invoice(gateway.clone(), payload.clone()).await {
```
m1sterc001guy, line 304: Hmm it's probably fine for now, but it looks like this will retry even if the gateway returns a 5XX error. This means that the gateway is reachable, but something went wrong when processing the payment. I don't think we should retry in that case.

m1sterc001guy (dismissed):

modules/fedimint-ln-client/src/pay.rs
douglaz: I think it's reasonable to retry even on 5xx errors. For instance you may be getting 502 Bad Gateway because the server is down or a 500 Internal Server Error may happen because the LN node is behaving in a weird way. But of course in the long term we should try to get more granular on these different errors.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -227,6 +261,8 @@ pub enum GatewayPayError {
     },
     #[error("OutgoingContract was not created in the federation")]
     OutgoingContractError,
+    #[error("Gateway did not respond in time")]
+    Timeout,
```
douglaz, line 265: Is this being used?

elsirion: Good catch! I introduced this error variant to handle the "API call took too long but didn't return error" case that you rightfully argued against being an error case.

douglaz (approved):
LGTM

dpc (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-3516-to-releases/v0.1 origin/releases/v0.1
cd .worktree/backport-3516-to-releases/v0.1
git switch --create backport-3516-to-releases/v0.1
git cherry-pick -x 5e5bed4c1fc24be5ffe46c19af0f28582da943f8 2dfd1586ced92f7f125e9d932b2a1ad2100cc9c3 89422f871628ff95cd411b2c50013f2471c1ef84 86c38a5f2cfbcf875e442bbf9646ebc72991a289 f0b89a1eecfe3976a7bcf78856d46d8af326263c
```

justinmoon:
> Does it fix #3249 too?

@elsirion @m1sterc001guy

elsirion:
@justinmoon I'm not sure how the linked issue happened in the first place, the tx submission state machine looked good. What was the version this happened on?

PR #3517, opened 2023-11-03
DON'T MERGE: single guardian tests

elsirion:


elsirion:
We want to test the single guardian consensus. My idea was to run in single guardian mode whenever the tests run against mocks anyway, but I didn't see the mocked tests running anywhere so I added that to the backend tests. Does this seem correct @dpc?

justinmoon:
Dev call: this could use some review for whether he's duplicating any work

fedimint-server/src/consensus/server.rs
```diff
@@ -214,7 +211,76 @@ impl ConsensusServer {
...
+                        item,
+                        self.cfg.local.identity,
+                    )
+                    .await
+                    .is_ok()
+                {
+                    item_index += 1;
+                }
+
+                // we rely on the module consensus items to notice the timeout
```
dpc, line 253: Does it ever matter if we the network is quiet and this doesn't gets triggered for an extend period of time with items already in a current session?

dpc (dismissed):

fedimint-server/src/consensus/server.rs
elsirion: Possibly for recovery as one currently can't recover transactions that happened in an unfinished session. @joschisan

dpc: Probably not a biggy, but why not wrap the `self.submission_receiver.recv()` in a timeout with a deadline to `sessions_end_time` and just do the right thing?

elsirion:
We are running the CLI tests (not the latency tests) in single guardian mode *in addition* to the normal 3-of-4 federation mode.

https://github.com/fedimint/fedimint/blob/19127bd2ef23c9b671be0ef16858109db734fde5/scripts/tests/test-ci-all.sh#L133-L134

PR #3520, opened 2023-11-03
refactor: simplify database traits/handling

dpc:
During work on passing client context to client modules, I've hit some problems/limitations w.r.t. the database traits, and I started digging, finding little things that I didn't like or thought were wrong, and decided to take a stab and see if it can be improved. Took forever and required great attention to details of generics and lifetimes, but here is the result.

There are some other things that I might bring up for debate (e.g. do we even need `ModuleDatabaseTransaction` at all now?), but I wanted to limit the changes at the first point where changes are coherent and  actually work.

#### Notable changes:

"single use" and "notifying" database transaction structs are gone. It was all merged into a single `BaseDatabaseTransaction`.

The database implementations are only concerned with producing `IRawDatabase`, then it gets wrapped with `BaseDatabase` to make it `impl IDatabase`, then the whole `IDatabase` is the composable interface that we can nest. We wrap it in `Database` for UX.

Everything about handling prefixes is handled by `PrefixDatabase`, and nesting prefixes is unlimited now. The special cases of `module_instance_id` on a database level was weird.

I moved typed dbtx queries to a trait that gets automatically implemented for any type that implements raw queries. Upside: we get typed queries everywhere where we get raw queries. Downside: we need to import a trait.

Some update docs: https://github.com/dpc/fedimint/blob/refactor-db/docs/database.md#base-implementations

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -91,7 +91,9 @@ use fedimint_core::config::{
 use fedimint_core::core::{
     DynInput, DynOutput, IInput, IOutput, ModuleInstanceId, ModuleKind, OperationId,
 };
-use fedimint_core::db::{AutocommitError, Database, DatabaseTransaction, IDatabase};
+use fedimint_core::db::{
+    AutocommitError, Database, DatabaseTransaction, IDatabaseTransactionOpsCoreTyped, IRawDatabase,
+};
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::module::{
@@ -1330,7 +1332,7 @@ pub struct ClientBuilder {
 }
 
 pub enum DatabaseSource {
-    Fresh(Box<dyn IDatabase>),
+    Fresh(Database),
     Reuse(ClientArc),
 }
 
@@ -1385,13 +1387,13 @@ impl ClientBuilder {
     // TODO: impl config from federation
 
     /// Uses this database to store the client state
-    pub fn with_database<D: IDatabase + 'static>(&mut self, db: D) {
-        self.with_dyn_database(Box::new(db));
+    pub fn with_raw_database<D: IRawDatabase + 'static>(&mut self, db: D) {
+        self.with_database(Database::new(db, Default::default()));
     }
 
-    /// Uses this database to store the client state, allowing for flexibility
-    /// on the caller side by accepting a type-erased trait object.
-    pub fn with_dyn_database(&mut self, db: Box<dyn IDatabase>) {
+    // /// Uses this database to store the client state, allowing for flexibility
+    // /// on the caller side by accepting a type-erased trait object.
+    pub fn with_database(&mut self, db: Database) {
         let was_replaced = self.db.replace(DatabaseSource::Fresh(db)).is_some();
         assert!(
             !was_replaced,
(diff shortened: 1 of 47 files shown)


dpc:
@m1sterc001guy Let me know what you think. I might have missed some reasons behind stuff etc. Always happy to jump on a call to chat about it.

fedimint-client/src/lib.rs
```diff
@@ -1442,14 +1439,9 @@ impl ClientBuilder {
...
-        let fake_notifications = Default::default();
         let mut dbtx = match self.db.as_ref().ok_or(anyhow!("No database provided"))? {
-            DatabaseSource::Fresh(db) => DatabaseTransaction::new(
-                db.begin_transaction().await,
-                Default::default(),
-                &fake_notifications,
-            ),
-            DatabaseSource::Reuse(db) => db.db().begin_transaction().await,
+            DatabaseSource::Fresh(db) => db.begin_transaction().await,
+            DatabaseSource::Reuse(client) => client.db().begin_transaction().await,
```
elsirion, line 1444: Unrelated note: We likely want to remove this again. It was a workaround (451c97296bd0556f9516d5ffe84bd681cb1b02ac) for not being able to open the same DB again for recovery.

fedimint-core/src/db/mod.rs
```diff
@@ -354,674 +391,762 @@ impl Database {
...
+/// A database that wraps an `inner` one and adds a prefix to all operations,
+/// effectively creating an isolated partition.
+#[derive(Clone, Debug)]
+pub struct PrefixDatabase<Inner>
+where
+    Inner: Debug,
+{
+    prefix: Vec<u8>,
+    inner: Inner,
+}
```
elsirion, line 420: This is great! :pray:

elsirion (commented):
I really like the refactoring and couldn't find anything wrong with it but will wait for @m1sterc001guy to weight in :)

fedimint-core/src/db/mod.rs
```diff
@@ -354,674 +400,762 @@ impl Database {
...
 }
 
-/// Fedimint requires that the database implementation implement Snapshot
-/// Isolation. Snapshot Isolation is a database isolation level that guarantees
-/// consistent reads from the time that the snapshot was created (at transaction
-/// creation time). Transactions with Snapshot Isolation level will only commit
-/// if there has been no write to the modified keys since the snapshot (i.e.
-/// write-write conflicts are prevented).
-///
-/// Specifically, Fedimint expects the database implementation to prevent the
```
m1sterc001guy, line 398: Did this comment just get moved? This might still be helpful for understanding the concurrency semantics that Fedimint expects

fedimint-core/src/db/mod.rs
```diff
@@ -354,674 +400,762 @@ impl Database {
...
-            .add_notification_key(key)
+    async fn raw_get_bytes(&mut self, key: &[u8]) -> Result<Option<Vec<u8>>> {
+        (**self).raw_get_bytes(key).await
     }
-}
 
-// TODO: use macro again
-#[doc = " A handle to a type-erased database implementation"]
-#[derive(Clone)]
-pub struct CommitTracker {
```
m1sterc001guy, line 551: Do we still log when there's a write but no commit?

fedimint-core/src/db/mod.rs
m1sterc001guy: Ah I see it now

fedimint-rocksdb/src/lib.rs
```diff
@@ -433,17 +451,6 @@ mod fedimint_rocksdb_tests {
         .await;
     }
 
-    #[tokio::test(flavor = "multi_thread")]
-    #[should_panic(expected = "Cannot isolate and already isolated database.")]
-    async fn test_cannot_isolate_already_isolated_db() {
```
m1sterc001guy, line 438: Yay :)

m1sterc001guy:
> I really like the refactoring and couldn't find anything wrong with it

+1 to this as well. It seems to simplify the abstractions that we had earlier. I think it would help me understand more if we can update the diagram on this doc: https://github.com/fedimint/fedimint/blob/master/docs/database.md

fedimint-core/src/db/mod.rs
dpc: Still there. Just moved and diff shows it as removed + added ?

douglaz (approved):
Great improvement, LGTM

PR #3523, opened 2023-11-04
Only give gatewayd read access to lnd files

benthecarman:


elsirion (approved):
Good observation, thx!

PR #3524, opened 2023-11-04
fix: make sure we consider a transaction rejected only if it is actually invalid

joschisan:


fedimint-client/src/lib.rs
```diff
@@ -232,9 +231,9 @@ impl DynGlobalClientContext {
         update_stream
             .filter_map(|tx_update| {
                 std::future::ready(match tx_update.state {
-                    TxSubmissionStates::Accepted { txid } if txid == query_txid => Some(Ok(())),
-                    TxSubmissionStates::Rejected { txid, error } if txid == query_txid => {
-                        Some(Err(error))
```
elsirion, line 237: Please don't just throw away error messages, these are vital for debugging.

fedimint-client/src/transaction/sm.rs
```diff
@@ -39,44 +34,28 @@ impl IntoDynInstance for TxSubmissionContext {
...
 pub enum TxSubmissionStates {
     /// The transaction has been created and potentially already been submitted,
     /// but no rejection or acceptance happened so far
-    Created {
-        // TODO: put into wrapper struct
-        /// We need this filed to be able to submit multiple transactions in one
-        /// operation.
-        txid: TransactionId,
-        tx: Transaction,
-        next_submission: SystemTime,
```
elsirion, line 65: Removing this is good, explicitly modeling a state transition from created to created after some time was a mistake (spams the inactive state log).

fedimint-client/src/transaction/sm.rs
```diff
@@ -154,44 +107,37 @@ impl IntoDynInstance for TxSubmissionStates {
...
-                sleep(RESUBMISSION_INTERVAL).await;
+            Ok(submission_result) => {
+                if let SubmissionResult::Rejected(..) = submission_result {
+                    return;
+                }
+            }
+            Err(error) => {
+                if !error.is_retryable() {
+                    warn!(target: LOG_TARGET, ?error, "Federation returned non-retryable error");
+                }
```
elsirion, line 121: We need an upper bound on re-trying (or at least back off exponentially, although after a restart that would need to be preserved to be effective) as otherwise state machines might accumulate and crash the entire client (maximum slots for requests, subscriptions aren't super reliable under high load, …).

fedimint-server/src/net/api.rs
```diff
@@ -500,18 +503,21 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
...
-
-                Ok(tx_id)
+                // we want to return a rejection if and only if the submitted transaction is
+                // invalid and will be rejected if we were to submit it to consensus
+                match fedimint.submit_transaction(transaction).await{
+                    Ok(()) => Ok(SubmissionResult::Submitted(txid)),
+                    // we do not return the error as any difference in serialization
+                    // between peers may cause the client's api call to fail
+                    Err(..) => Ok(SubmissionResult::Rejected(txid))
+                }
```
elsirion, line 520: This is good

fedimint-client/src/lib.rs
joschisan: I had to throw it away since we do not return it via the api endpoint anymore, @elsirion already commented on this

justinmoon:
Dev call: Related to https://github.com/fedimint/fedimint/issues/3173

fedimint-client/src/lib.rs
elsirion: We talked about this on a call and it makes sense, what do you think is the best way to get some insight into failures back though? Not having any client-side debug info will be a pain once there are independent federations where we can't see the logs. A custom query strategy that is a mix of union (for errors) and threshold (for acceptance) could work, right?

fedimint-server/src/net/api.rs
```diff
@@ -119,46 +124,14 @@ impl ConsensusApi {
...
-        }
-
-        funding_verifier.verify_funding()?;
+        process_transaction_with_dbtx(self.modules.clone(), &mut dbtx, transaction.clone()).await?;
 
         self.submission_sender
             .send(ConsensusItem::Transaction(transaction))
-            .await?;
+            .await
+            .ok();
```
elsirion, line 133: Should we log if that ever fails? Or can this normally happen during shutdown?

fedimint-client/src/transaction/sm.rs
```diff
@@ -154,44 +124,47 @@ impl IntoDynInstance for TxSubmissionStates {
...
-            Err(e) if e.is_retryable() => {
-                debug!("Got {e} while submitting transaction, will sleep for {RESUBMISSION_INTERVAL:?}");
-                sleep(RESUBMISSION_INTERVAL).await;
+            Ok(submission_result) => {
+                if let Err(submission_error) = submission_result {
+                    return Ok(submission_error.to_string());
+                }
+            }
+            Err(error) => {
+                if !error.is_retryable() {
```
elsirion, line 139: Any reason to not use `Err(e) if e.is_retryable()` anymore?

elsirion (approved):
Looks good, two nits :)

fedimint-server/src/net/api.rs
joschisan: May happen during shutdown I believe.

fedimint-client/src/transaction/sm.rs
joschisan: I find match guards in this case less intuitive and early return here more idiomatic.

fedimint-backports:

```bash
git worktree add -d .worktree/backport-3524-to-releases/v0.1 origin/releases/v0.1
cd .worktree/backport-3524-to-releases/v0.1
git switch --create backport-3524-to-releases/v0.1
git cherry-pick -x 1b95df4e605b6506a29b2c4c864dbfb4cfbe8751
```

PR #3533, opened 2023-11-06
feat: clients use vetted gateway meta field in selecting gateway

### LINKED ISSUE #3420: Whitelist gateways
Currently a poorly-run gateway can join a federation and cause a lot of payment failures because the client isn't smart about ignoring such a gateway. For this reason, we should have an optional mechanism to whitelist gateways enabling guardians to prevent such unknown gateways from registering. 

Simplest way to do this is a `FM_GATEWAY_WHITELIST` parameter that you can set to a list of `gateway_id`s. But this might be something that should be under consensus and configurable via guardian dashboard in the future.

### LINKED ISSUE #3649: cache meta overrides sourced from meta_override_url
Clients should [cache meta override json](https://github.com/fedimint/fedimint/issues/3420#issuecomment-1817898112) sourced from some configured `meta_override_url`

okjodom:
Fedimint clients select gateways based on [`vetted_gateways` meta field](https://github.com/fedimint/fedimint/blob/master/docs/meta_fields/vetted_gateways.md)
as configured for a federation via some `meta_override_url`

**Testing:**
- Configure FM  with`meta_override_url`. For example in devimint, you can add the following to [federation.rs](https://github.com/fedimint/fedimint/blob/c5ce0e99cb6ee18e3e975578891e9b0f752b327e/devimint/src/federation.rs#L501) and then `just mprocs`

```rs
let mut meta = BTreeMap::from([
    ("federation_name".to_string(), "testfed".to_string()),
    (
        "meta_override_url".to_string(),
        serde_json::json!("https://getln.xyz/fedimint/meta").to_string(),
    ),
])
```
- Get the `federation-id` and one or more `gateway-id`s then update `vetted_gateways` served by meta override url
```
curl -X POST https://getln.xyz/fedimint/meta -H "Content-Type: application/json" -d '{
    "federation-id": {
        "vetted_gateways": ["gateway-id"]
    }
}'
```
This federation id and vetted gateways should show up in `curl -X POST https://getln.xyz/fedimint/meta` json

- Run `fedimint-cli list-gateways` and confirm the vetted gateways are visible to the client
- Remove one or all gateways from the vetted list
```
curl -X POST https://getln.xyz/fedimint/meta -H "Content-Type: application/json" -d '{
    "federation-id": {
        "vetted_gateways": []
    }
}'
```

Meta configs are **cached up to 10 minutes** by the client

closes #3420, closes #3649


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -808,11 +827,94 @@ impl LightningClientModule {
...
+            let vetted_gids = match meta
+                .get(META_VETTED_GATEWAYS_KEY)
+                .and_then(|vetted| serde_json::from_value::<Vec<String>>(vetted.clone()).ok())
+            {
+                Some(vetted) => {
+                    debug!("Found the following vetted gateways: {:?}", vetted);
+                    vetted
+                        .into_iter()
+                        .map(|pk| PublicKey::from_str(&pk).map_err(anyhow::Error::from))
+                        .collect::<Result<Vec<_>, _>>()?
```
elsirion, line 906: nit: one faulty pub key doesn't necessarily need to invalidate all the others. Not a big deal though imo.

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -197,13 +200,15 @@ impl LightningGatewayRegistration {
...
-/// Information about a gateway that is shared with other federations and
+/// Information about a gateway that is shared with other federation members and
 /// expires based on a TTL to allow for sharing between nodes with
 /// unsynchronized clocks which can each anchor the announcement to their local
 /// system time.
 #[derive(Debug, Clone, Serialize, Deserialize, Encodable, Decodable, PartialEq, Eq, Hash)]
 pub struct LightningGatewayAnnouncement {
     pub info: LightningGateway,
+    /// Indicates if this announcement has been vetted by the federation
+    pub vetted: bool,
```
elsirion, line 211: Not sure if this belongs into the `LightningGatewayAnnouncement` and `LightningGatewayRegistration`, feels like this could easily lead to bugs where gateways can announce themselves as "vetted". As far as I can tell the only thing preventing this right now is that `filter_duplicate_gateways` always sets the field to `false`.

elsirion (approved):
Looks really good, just two nits but I think this is merge-ready. Just giving @m1sterc001guy a chance to have a look.

elsirion:
Ah, you need to regenerate the migration test snapshot. See this `just` command:

https://github.com/fedimint/fedimint/blob/c5ce0e99cb6ee18e3e975578891e9b0f752b327e/justfile.fedimint.just#L39-L45

Last time it didn't work for me and I had to tweak it a bit, removing some argument iirc., but I think the command will point you in the right direction.

okjodom:
> Last time it didn't work for me and I had to tweak it a bit, removing some argument iirc., but I think the command will point you in the right direction.

Seemed to work out of box except for the trailing whitespaces in the migration logs. Previous migrations have these whitespaces, so how do they pass pre-commit and lint checks?!

modules/fedimint-ln-common/src/lib.rs
okjodom: Makes sense. Is there a better fix than replacing `LightningGatewayAnnouncement` in `register_gateway` with a different struct? didn't want to add yet another struct for the gateway registration circuit

elsirion:
> Seemed to work out of box

Then something is wrong with my setup :grimacing: 

I always use `--no-verify` on those and somehow CI is happy while git hooks aren't.

modules/fedimint-ln-common/src/lib.rs
elsirion: It's ok for now, I don't have a much better idea. Is it really needed for both of the structs? Feels like only the one saved to/loaded from the DB needs it but not the one transmitted over the network. Then the field could default to `false` whenever one is converted to the other,

elsirion (approved):

modules/fedimint-ln-common/src/lib.rs
okjodom: could remove it for `LightningGatewayRegistration` transmitted over the network

okjodom: though when we have dynamic meta, then we'd apply vetted field to the registration struct, and not announcement

PR #3544, opened 2023-11-07
cleanup: remove threshold signing of and download tokens for client config

joschisan:
Signing the client config is unnecessary, adds complexity to the consensus and does not allow for updates of the client config. Instead of a public key the FederationId commits to the guardians api endpoint set; then we can download the full set from a single peer via the one url provided in the invite code and then request the full client config via request_current_consensus. Furthermore, it relies on the unmaintained threshold_crypto crate that we want to get rid of. We will follow up this pr with https://github.com/fedimint/fedimint/pull/3545 that removes the threshold_crypto crate from the dummy module so that the only place where it remains is the legacy lightning module which will be replaced by https://github.com/fedimint/fedimint/pull/3468 such that we can finally remove threshold_crypto.

Since we now have to download the ClientConfig from a threshold of peers instead of only one we would have to provide a download token for every guardian, not just one, in the InviteCode. Instead we opt to remove the download tokens entirely as it cannot prevent the ClientConfig to be shared by a client anyways. The inherent anonymity of fedimint makes it impossible for the guardians to restrict their users to a certain group. We can, however, set a meta field boolean flag that indicates if the client is supposed to share the client config with new users.


dpc:
Cool beans.

Do we have any future escape hatch for membership-set changes (if ever decided to pursue it)? With public key & threshold signing of the config, the idea was to maintain a list of signed membership set changes in a chain from the original id.

Maybe we could commit to endpoints + some public keys which initially don't need to sign anything, just in case. In the absence of a threshold_crypto, we could just gather a naive multisig of `threshold` members, and form a chain from the original ID. Since these should be rare, the extra size doesn't really matter.

fedimint-core/src/config.rs
```diff
@@ -264,8 +251,7 @@ impl FederationId {
         &self,
         secp: &secp256k1::Secp256k1<secp256k1_zkp::All>,
     ) -> anyhow::Result<secp256k1::PublicKey> {
-        let bytes = <Sha256 as bitcoin_hashes::Hash>::hash(&self.0.to_bytes()[..]);
-        let sk = secp256k1::SecretKey::from_slice(&bytes)?;
+        let sk = secp256k1::SecretKey::from_slice(&self.0)?;
```
elsirion, line 258: I wonder if this secret key being essentially public is a problem in any way?

fedimint-server/src/config/mod.rs
```diff
@@ -102,15 +100,6 @@ pub struct ServerConfigPrivate {
...
     pub broadcast_secret_key: SecretKey,
-    /// Secret key for contributing to threshold auth key
-    #[serde(with = "serde_binary_human_readable")]
-    pub auth_sks: SerdeSecret<hbbft::crypto::SecretKeyShare>,
-    /// Secret key for contributing to HBBFT consensus
-    #[serde(with = "serde_binary_human_readable")]
-    pub hbbft_sks: SerdeSecret<hbbft::crypto::SecretKeyShare>,
-    /// Secret key for signing consensus epochs
-    #[serde(with = "serde_binary_human_readable")]
-    pub epoch_sks: SerdeSecret<hbbft::crypto::SecretKeyShare>,
```
elsirion, line 113: :tada:

elsirion (approved):
Looks good to me, I'm very happy to get rid of `threshold_crypt` (at least in the core fedimint code, LN module to follow). One thing I'm concerned about is making it harder to change the API endpoints (e.g. because a domain expired), but that's not implemented right now anyway. As long as we version our join codes before 0.2 (#3556) this seems ok though.

PR #3558, opened 2023-11-09
fix: DatabaseTransactionRef should share commit tracker with parent

dpc:
This fixes annoying warnings about uncommited writes when `DatabaseTransactionRef` (which is effectively just a reference to the real Tx) is dropped.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -1164,43 +1164,48 @@ impl Drop for CommitTracker {
 /// This is used in APIs that require logic to be a part of a larger
 /// database transaction, given that `commit_tx` takes `&mut self` (
 /// not `self`, due to trait/Sized and support for `&mut T` issues).
-pub struct DatabaseTransactionRef<'a>(DatabaseTransaction<'a>);
+pub struct DatabaseTransactionRef<'tx> {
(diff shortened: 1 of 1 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -1286,14 +1291,12 @@ impl<'tx> DatabaseTransaction<'tx> {
         's: 'a,
     {
         let decoders = self.decoders.clone();
-        let commit_tracker = self.commit_tracker.clone();
 
-        DatabaseTransactionRef(DatabaseTransaction {
+        DatabaseTransactionRef {
             tx: Box::new(&mut self.tx),
```
elsirion, line 1296: Do we actually need this `Box<dyn IDatabaseTransaction + 'tx>` or could this be `&dyn IDatabaseTransaction + 'tx`?

elsirion (approved):
Looks good, one minor nit/question.

fedimint-core/src/db/mod.rs
dpc: It would work in this case, but for other case we need to put a `PrefixDatabaseTransaction` which forces us to box it. &mut PrefixDatabaseTransaction` would cause a reference to a local object.

We could make it a `Cow` or something, but probably not worthwhile.

PR #3560, opened 2023-11-09
cleanup: remove threshold signing of messages from dummy module

joschisan:
The threshold signing functionality was implemented with two anti-patterns in module design, namely api calls writing directly to the database and collecting signature shares via the broadcast and combining them server side. Since we merged https://github.com/fedimint/fedimint/pull/3419 the mint module demonstrates the proper way to sign messages like nonces by signing a message if it is the output of a accepted transaction and collecting and combining the signature shares client side. Since this is pretty simple in the mint module already there is no need to demonstrate this functionality again in the dummy module.


elsirion (approved):
Lol, I didn't even know it *did* implement message signing! Might be worth to compile a "best practices" document when it comes to consensus design in modules.

joschisan:
> Lol, I didn't even know it _did_ implement message signing! Might be worth to compile a "best practices" document when it comes to consensus design in modules.

It would be something like this:

* avoid non-final / stateful outputs
* avoid collecting signature / decryption shares over the broadcast, make the combined result verifiable and push it out to the client
* if a consensus item is not necessary to process transactions it probably should not be a CI in the first place
* do not write to the db in the api, all state changes should come from transactions
* stick to the transaction model as closely as possible, so inputs should always consume BTC of previous outputs. Ideally force a transactions input to fully consume the output, avoid partially consumed outputs.

The Mint module is now an example of this practices, the next generation ln module will be as well.

PR #3563, opened 2023-11-10
refactor: get rid of DummyClientExt

dpc:
Now that `ClientModule` has access to its context (`Client`), it can use it instead of exposing `*Ext` trait.

I picked dummy module as a first one to refactor, because I thought it will be the easiest. Didn't realize it's used all over the tests...

Due to lifetimes issues in async context my original idea of exposing only weak pointer didn't work, but I also realized that we probably don't want to expose the whole `Client` anyway, and we should be more selective about what we allow, so decided to just delegate `Client` methods I need on the `ClientContext` for now.

While working on it I noticed lots of possible improvements, and don't want to waste time splitting things into smaller commits, etc. so just did what I need immediate to make things simpler. Will follow through with some more changes.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -466,7 +466,24 @@ fn states_add_instance(
 /// methods live.
 #[derive(Debug)]
 pub struct ClientArc {
+    // Use [`ClientArc::new`] instead
     inner: Arc<Client>,
+
+    __use_constructor_to_create: (),
+}
+
+impl ClientArc {
+    /// Create
+    fn new(inner: Arc<Client>) -> Self {
+        inner
+            .client_count
+            .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        Self {
+            inner,
+            // this is the constructor
+            __use_constructor_to_create: (),
+        }
+    }
 }
 
 impl ops::Deref for ClientArc {
@@ -487,12 +504,7 @@ impl ClientArc {
 
 impl Clone for ClientArc {
     fn clone(&self) -> Self {
-        self.inner
-            .client_count
-            .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
-        ClientArc {
-            inner: self.inner.clone(),
-        }
+        ClientArc::new(self.inner.clone())
     }
 }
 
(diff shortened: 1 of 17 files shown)


fedimint-client/src/lib.rs
```diff
@@ -564,13 +564,26 @@ const SUPPORTED_CORE_API_VERSIONS: &[fedimint_core::module::ApiVersion] =
...
+impl<'m, M> ops::Deref for ClientModuleInstance<'m, M>
+where
+    M: ClientModule,
+{
+    type Target = M;
+
+    fn deref(&self) -> &Self::Target {
+        self.module
+    }
 }
```
elsirion, line 587: Feels a bit dirty but I like it xD

fedimint-client/src/lib.rs
```diff
@@ -564,13 +564,26 @@ const SUPPORTED_CORE_API_VERSIONS: &[fedimint_core::module::ApiVersion] =
...
 
 /// Resources particular to a module instance
-pub struct ClientModuleInstance {
+pub struct ClientModuleInstance<'m, M: ClientModule> {
     /// Instance id of the module
     pub id: ModuleInstanceId,
     /// Module-specific DB
     pub db: Database,
     /// Module-specific API
     pub api: DynModuleApi,
```
elsirion, line 573: Separate issue: Should these be accessible via functions so we can change the way they are stored?

elsirion (approved):
Great PR! Once this is done for all modules `fedimint-cli` should probably also be changed to only allow modularized operations.

elsirion:
I think you actually broke some tests.

fedimint-client/src/lib.rs
dpc: We could just make it a method or a field, and it wouldn't be that much of a downgrade.

... but if we squint our eyes, we can consider it a smart pointer over module, so I guess it's OK?

fedimint-client/src/lib.rs
dpc: Was thinking the same. Probably yes, so the owner can't assign at least?

elsirion: I could also imagine just having a normal dbtx reference in there and creating the module dbtx on demand etc.

justinmoon (approved):

Issue #2785, opened 2023-07-19
Gateway option to "withdraw everything"


justinmoon:
Currently it's a little error-prone because you have to guess the fees that will be charged.

justinmoon:
Call: the lightning and wallet clients should be able to do this in general

justinmoon:
Was just discussing this with @Kodylow. I think we could modify `WithdrawPayload` used by the gateway webserver to allow users to sweep their whole wallet. We could just say `withdrawal_amount = balance_amount - fee_amount` if the user requested a sweep.

okjodom:
> if the user requested a sweep

Sounds good but I'm not sure I understand how the API change would capture user intent to sweep

Kodylow:
Add a param for `sweep: bool`

justinmoon:
@okjodom raises a good point: we could have the client fetch the fee and calculate the amount required to sweep which wouldn't require a change to the withdraw RPC.

The [peg_out_fees](https://github.com/fedimint/fedimint/blob/23ee7cb6e96fce89bb024fbc1fcfccbfb3dc968b/modules/fedimint-wallet-server/src/lib.rs#L634-L662) RPC requires user to provide address and amount. Amount is tricky because that creates circular dependency -- we're calling `peg_out_fees` because we're trying to calculate the amount. Would we just send the user's entire balance as the amount parameter to `peg_out_fees`? Then there's a small chance we'd over-pay our fee if removing the fee amount causes us `varint` serialization of the transaction to be smaller. But that will basically never happen.

elsirion:
> Add a param for `sweep: bool`

I think the CLN way is a bit more elegant: allow the string `all` as a valid parameter and make the amount an enum over `all` or a concrete amount.

@justinmoon we could just loop and double-check that the estimated fee is still the same after subtracting the fee from the withdraw amount and break if either the fee stays the same or we hit some iteration limit.

PR #3573, opened 2023-11-10
feat: withdraw_fees gateway rpc command

Kodylow:
Adds a gateway rpc command to just get the withdraw fees to use when calculating a sweep transaction.

Partial fix to #2785

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -59,7 +59,7 @@ use gateway_lnrpc::{GetNodeInfoResponse, InterceptHtlcResponse};
 use lightning_invoice::RoutingFees;
 use lnrpc_client::{ILnRpcClient, LightningBuilder, LightningRpcError, RouteHtlcStream};
 use rand::rngs::OsRng;
-use rpc::{AmountOrAll, FederationInfo, SetConfigurationPayload};
+use rpc::{BitcoinAmountOrAll, FederationInfo, SetConfigurationPayload};
 use secp256k1::PublicKey;
 use serde::{Deserialize, Serialize};
 use state_machine::pay::OutgoingPaymentError;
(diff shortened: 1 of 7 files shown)


elsirion:
Why does `gateway-cli` need to know that fee? Can't we tell `gatewayd` to send `all` and it then figures out internally what that means?

elsirion:
The problem is (in rare occasions) the fee might change after substracting the fee from the withdraw amount, so it's not something you really want to do manually.

Kodylow:
That makes sense, will update.

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -50,11 +51,31 @@ pub struct DepositAddressPayload {
...
     pub federation_id: FederationId,
-    #[serde(with = "bitcoin::util::amount::serde::as_sat")]
-    pub amount: bitcoin::Amount,
+    pub amount: AmountOrAll,
     pub address: Address,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum AmountOrAll {
```
justinmoon, line 60: Maybe call it `BitcoinAmountOrAll`? Just looking at the struct name I would assume this was a fedimint amount. Also, could you add a doc string to this struct https://x.com/dpc_pw/status/1722457120350589305?s=20

gateway/ln-gateway/src/lib.rs
```diff
@@ -670,8 +672,13 @@ impl Gateway {
             address,
             federation_id,
         } = payload;
-
         let client = self.select_client(federation_id).await?;
+
+        let amount: bitcoin::Amount = match amount {
+            AmountOrAll::All => bitcoin::Amount::from_sat(client.get_balance().await.msats * 1000),
```
justinmoon, line 678: I think this should probably use like `client.get_balance() - client.get_peg_out_fees()` and make sure this is valid amount

justinmoon:
Might be good to also modify `fedimint-cli` to accept `all` as an amount?

Kodylow:
@justinmoon see #3574 for add to fedimint-cli

gateway/ln-gateway/src/lib.rs
```diff
@@ -670,12 +672,24 @@ impl Gateway {
...
 
-        let operation_id = client.withdraw(address, amount, fees).await?;
+        // TODO: Fees should probably be passed in as a parameter
+        let (amount, fee) = match amount {
+            // If the amount is "all", then we need to subtract the fees from
+            // the amount we are withdrawing
+            BitcoinAmountOrAll::All => {
+                let balance = bitcoin::Amount::from_sat(client.get_balance().await.msats * 1000);
+                let fees = client.get_withdraw_fee(address.clone(), balance).await?;
+                (balance - fees.amount(), fees)
```
elsirion, line 684: You should do this in a loop imo to check that the fee you have to pay stayed the same. There is no guarantee that the fee cannot rise from withdrawing a different amount (maybe there is currently due to our UTXO selection approach, but I wouldn't count on it).

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -50,11 +51,32 @@ pub struct DepositAddressPayload {
...
     pub address: Address,
 }
 
+/// Amount of bitcoin to send, or "all" to send all available funds
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum BitcoinAmountOrAll {
+    Amount(#[serde(with = "bitcoin::util::amount::serde::as_sat")] bitcoin::Amount),
+    All,
+}
```
elsirion, line 64: Do we want the serialization of `All` to be `null`? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=50dfc340d46f6d2652902b19a56ff6c1

gateway/ln-gateway/src/lib.rs
dpc: Oh, if the fee is supposed to be both of Fedimint tx and on chain fees, then we should use msats for `balance` (and the `fee`) BTW, just as I thought though.

Also - the loop termination condition should probably be "not increasing" or something. I suspect "stayed the same" can lead to hangs.

Kodylow:
closing in favor of #3574

PR #3574, opened 2023-11-10
feat: add BitcoinAmountOrAll to fedimint-cli 

Kodylow:
builds on #3573 BitcoinAmountOrAll, moves struct into fedimint_core and adds 'all' option to fedimint-cli clap args

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -13,7 +13,7 @@ use fedimint_client::ClientArc;
 use fedimint_core::config::{ClientConfig, FederationId};
 use fedimint_core::core::{ModuleInstanceId, ModuleKind, OperationId};
 use fedimint_core::time::now;
-use fedimint_core::{Amount, ParseAmountError, TieredSummary};
+use fedimint_core::{Amount, BitcoinAmountOrAll, ParseAmountError, TieredSummary};
 use fedimint_ln_client::{
     InternalPayState, LightningClientModule, LnPayState, LnReceiveState, OutgoingLightningPayment,
     PayType,
@@ -91,7 +91,7 @@ pub enum ClientCmd {
     /// Withdraw funds from the federation
     Withdraw {
         #[clap(long)]
-        amount: bitcoin::Amount,
+        amount: BitcoinAmountOrAll,
         #[clap(long)]
         address: bitcoin::Address,
     },
@@ -439,9 +439,24 @@ pub async fn handle_command(
         }
         ClientCmd::Withdraw { amount, address } => {
             let wallet_module = client.get_first_module::<WalletClientModule>();
-            let fees = wallet_module
-                .get_withdraw_fees(address.clone(), amount)
-                .await?;
+            let (amount, fees) = match amount {
+                // If the amount is "all", then we need to subtract the fees from
+                // the amount we are withdrawing
+                BitcoinAmountOrAll::All => {
+                    let balance =
+                        bitcoin::Amount::from_sat(client.get_balance().await.msats * 1000);
+                    let fees = wallet_module
+                        .get_withdraw_fees(address.clone(), balance)
+                        .await?;
+                    (balance - fees.amount(), fees)
+                }
+                BitcoinAmountOrAll::Amount(amount) => (
+                    amount,
+                    wallet_module
+                        .get_withdraw_fees(address.clone(), amount)
+                        .await?,
+                ),
+            };
             let absolute_fees = fees.amount();
 
             info!("Attempting withdraw with fees: {fees:?}");
--- a/fedimint-core/src/lib.rs
+++ b/fedimint-core/src/lib.rs
@@ -146,6 +146,28 @@ pub fn sats(amount: u64) -> Amount {
     Amount::from_sats(amount)
 }
 
+/// Amount of bitcoin to send, or "all" to send all available funds
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum BitcoinAmountOrAll {
+    Amount(#[serde(with = "bitcoin::util::amount::serde::as_sat")] bitcoin::Amount),
+    All,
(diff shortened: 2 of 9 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -670,10 +670,22 @@ impl Gateway {
...
         let client = self.select_client(federation_id).await?;
-        // TODO: This should probably be passed in as a parameter
-        let fees = client.get_withdraw_fee(address.clone(), amount).await?;
+
+        // TODO: Fees should probably be passed in as a parameter
+        let (amount, fees) = match amount {
+            // If the amount is "all", then we need to subtract the fees from
+            // the amount we are withdrawing
+            BitcoinAmountOrAll::All => {
+                let balance = bitcoin::Amount::from_sat(client.get_balance().await.msats * 1000);
```
dpc, line 680: I don't think there's any reason to round it down here. It should all work on msats, no?

dpc: Yeah, `fedimint_core::Amount` has a higher precision, and if user has `1000.005sats`, fee is `1000sats`, they should be able to widraw `0.005msat` just fine, no matter how uneconomical it is. :D

Kodylow: For an onchain pegout transaction you can't get down to msats, it has to be nearest sat?

dpc: :brain::hammer: 

Right. :D

Thought I'd still keep it in highest precision available&needed, and round down at the very last place where it is necessary as a best practice. Otherwise the code e.g. ever changes it might lead to a mistake.

But not a blocker.

dpc (dismissed):

gateway/ln-gateway/src/lib.rs
Kodylow: amount isn't used anywhere past withdraw, should we rewrite withdraw to take a FedimintAmount instead of a BitcoinAmount then?

dpc: Nah. Whateva.

fedimint-cli/src/client.rs
```diff
@@ -91,7 +91,7 @@ pub enum ClientCmd {
     /// Withdraw funds from the federation
     Withdraw {
         #[clap(long)]
-        amount: bitcoin::Amount,
+        amount: BitcoinAmountOrAll,
```
Maan2003, line 94: maybe a type `AllOr<T>`?

Maan2003: depends on if we want more of these `AllOr` types

fedimint-cli/src/client.rs
```diff
@@ -421,10 +421,23 @@ pub async fn handle_command(
...
+                        bitcoin::Amount::from_sat(client.get_balance().await.msats * 1000);
+                    let fees = client.get_withdraw_fee(address.clone(), balance).await?;
+                    (balance - fees.amount(), fees)
+                }
+                BitcoinAmountOrAll::Amount(amount) => (
+                    amount,
+                    client.get_withdraw_fee(address.clone(), amount).await?,
+                ),
+            };
+            let absolute_fee = fees.amount();
```
elsirion, line 438: Feels like this should be implemented on the wallet module so we don't duplicate the code in the gateway. #3573

fedimint-core/src/lib.rs
```diff
@@ -146,6 +146,28 @@ pub fn sats(amount: u64) -> Amount {
...
+        if s == "all" {
+            Ok(BitcoinAmountOrAll::All)
+        } else {
+            bitcoin::Amount::from_str(s)
+                .map(BitcoinAmountOrAll::Amount)
+                .map_err(|_| bitcoin::util::amount::ParseAmountError::InvalidFormat)
+        }
+    }
+}
+
```
elsirion, line 170: We could make this type generic over the amount type and keep it in fedimint-core, but as it is it feels more like something wallet-specific.

elsirion (commented):
I'm a bit confused, can we close #3573 in favor of this one? If so, please still see the comments there.

fedimint-cli/src/client.rs
Kodylow: Moved into fedimint-wallet-common. Then if we want to refactor to have the generalized `AllOr<T>` we can do that in fedimint-core but for now wallet sounds good.

fedimint-core/src/lib.rs
Kodylow: Agreed, moved into wallet, see prev comment.

Kodylow:
@elsirion see https://github.com/fedimint/fedimint/pull/3574/commits/3a2a50990f27aeeb8736e0af272853b014b21b84 for fix for 'all' serialization from other PR's comments.

fedimint-cli/src/client.rs
Kodylow: gateway doesn't have fedimint-wallet as a dependency, cool to add it and move to fedimint-wallet or should I keep it in fedimint core and make it generic `AllOr<T>` like @Maan2003 suggests?

elsirion: I see, then let's leave it here

elsirion:
Clippy is currently unhappy https://github.com/fedimint/fedimint/actions/runs/6858954816/job/18650509987?pr=3574#step:6:74

dpc:
> Clippy is currently unhappy https://github.com/fedimint/fedimint/actions/runs/6858954816/job/18650509987?pr=3574#step:6:74

Clippy is generally grumpy and uncooperative, so don't take it personally. Just hard character to work with for everyone.

elsirion:
You should be able to run clippy locally: `just clippy`. Much faster than waiting for CI every time.

dpc:
And `just clippy-fix` :)

dpc (approved):

PR #3577, opened 2023-11-11
chore: enable cargo-deny check in the CI 

dpc:
Currently really only used to enforce approved licenses.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,277 @@
+# This template contains all of the possible sections and their default values
+
+# Note that all fields that take a lint level have these possible values:
+# * deny - An error will be produced and the check will fail
+# * warn - A warning will be produced, but the check will not fail
+# * allow - No warning or error will be produced, though in some cases a note
+# will be
+
+# The values provided in this template are the default values that will be used
+# when any section or field is not specified in your own configuration
+
+# Root options
+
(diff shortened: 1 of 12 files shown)


deny.toml
```diff
@@ -0,0 +1,277 @@
...
+# of features from the crate(s) to all of the graph roots can be far too verbose.
+# This option can be overridden via `--feature-depth` on the cmd line
+feature-depth = 1
+
+# This section is considered when running `cargo deny check advisories`
+# More documentation for the advisories section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html
+[advisories]
+# The path where the advisory database is cloned/fetched into
+db-path = "~/.cargo/advisory-db"
```
elsirion, line 61: Aren't we already checking advisories in another way? Is that redundant now?

elsirion (approved):

deny.toml
dpc: We don't use advisories part of `cargo-deny` as it's redundant, and requires scanning whole source code, which makes caching here worse that the existing solution.

dpc:
I'm debugging an unrelated crane/flakebox problem that is blocking this PR.

elsirion (approved):

PR #3582, opened 2023-11-12
cleanup: associated error types for server modules

joschisan:
The associated error types allow us to achieve consensus encode the error that may be returned on submission of a transaction. Previously we had to turn the error into a string before returning it, hence querying the submission endpoint with request_current_consensus may fail if the error get serialised differently. We do not create a  third error type for consensus items since they are not returned via the API.

Please note that I have removed a test since the transaction submission state machine would require the testing context to implement decoders now but most importantly testing the happy path of transaction submission as a unit test seemed redundant here as its tested over and over in the integration tests.


elsirion (approved):
Great improvement!

elsirion (approved):

PR #3584, opened 2023-11-13
chore: remove client.json from client data dir

Maan2003:


elsirion (approved):
I think we can get rid of `client.json` altogether, but this is a good start.

PR #3589, opened 2023-11-14
fix: a typo that lead to all backend tests being skipped

dpc:
Upstream issue to figure out why this has been failing silently, as I still don't understand: https://github.com/nextest-rs/nextest/issues/1108

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -5,4 +5,4 @@ slow-timeout = { period = "60s", terminate-after = 8 }
 [profile.ci]
 fail-fast = true
 failure-output = "immediate"
-slow-timeout = { period = "60s", terminate-after = 4 }
+slow-timeout = { period = "60s", terminate-after = 8 }
--- a/fedimint-bitcoind/src/lib.rs
+++ b/fedimint-bitcoind/src/lib.rs
@@ -9,6 +9,7 @@ use anyhow::Context;
 pub use anyhow::Result;
 use bitcoin::{BlockHash, Network, Script, Transaction, Txid};
 use fedimint_core::bitcoinrpc::BitcoinRpcConfig;
+use fedimint_core::fmt_utils::OptStacktrace;
 use fedimint_core::task::TaskHandle;
 use fedimint_core::txoproof::TxOutProof;
 use fedimint_core::util::SafeUrl;
@@ -175,7 +176,7 @@ impl<C> RetryClient<C> {
                         return Err(e);
                     }
 
-                    info!(target: LOG_BLOCKCHAIN, "Bitcoind error {:?}, retrying", e);
+                    info!(target: LOG_BLOCKCHAIN, "Bitcoind error {}, retrying", OptStacktrace(e));
                     std::thread::sleep(retry_time);
                     retry_time = min(RETRY_SLEEP_MAX_MS, retry_time * 2);
                 }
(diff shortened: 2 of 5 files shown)


dpc:
Fails correctly. :upside_down_face:

bradleystachurski (approved):
CI failures match `just test-ci-all` failures on my machine 👍 

```
fedimint-test-all-ci> 00:00:27 called `Result::unwrap()` on an `Err` value: Connection refused (os error 111)
```

We're unable to connect to the nodes (I'm investigating, no obvious fix yet).

bradleystachurski:
What's a sane approach for merging this? If we merge as-is, all subsequent PRs will fail CI and we'll grind to a merge halt until we fix the root issue. Would it be better to hold off merging until we resolve the root issue and continue allowing other PRs to skip?

elsirion:
After thinking about the nextest execution model it appears to me that it cannot work with our way of testing: we rely on locking access to stateful services (e.g. `bitcoind`) in certain scenarios and that doesn't work anymore if the tests suddenly run in different processes (or at least would require file system based locking or some other form of ipc).

elsirion:
After using the old `cargo test` based version of running these tests only some gateway tests fail:

```
00:01:02     can_switch_active_gateway
00:01:02     cannot_pay_same_external_invoice_twice
00:01:02     makes_internal_payments_within_federation
00:01:02     rejects_wrong_network_invoice
```

All fail with the same error:

```
00:01:02 thread 'can_switch_active_gateway' panicked at /build/source/fedimint-testing/src/gateway.rs:151:10:
00:01:02 Gateway failed to start: Gateway did not reach desired state within 30 seconds
```

This indicates we always get stuck here:

https://github.com/fedimint/fedimint/blob/256bdfada2cfe7716f58e5e119ab248418158a18/fedimint-testing/src/gateway.rs#L145-L151

Did we maybe change the GW setup flow in a way that silently breaks devimint and just makes the GW not come online @okjodom, @m1sterc001guy?

bradleystachurski:
@elsirion Thanks for applying the port fixes!

We hardcode the esplora port in the fixtures, which was also causing connection errors. When you have a chance, can you please also apply the diff from [fedimint-testing/src/fixtures.rs](https://github.com/fedimint/fedimint/pull/3590/files#diff-10190eebcc8d9a344d6dfd2cfb64b437a1200997f418994b64f5d823cb7ffd5b)?

elsirion:
Hey, sorry for not seeing your PR earlier @bradleystachurski, I thought this would be a simple fix to this PR, but it got messier and I don't want to keep force-pushing to @dpc's branch. I created my own version of the PR #3594 that includes the patch you linked to.

dpc:
> After thinking about the nextest execution model it appears to me that it cannot work with our way of testing: we rely on locking access to stateful services (e.g. `bitcoind`) in certain scenarios and that doesn't work anymore if the tests suddenly run in different processes (or at least would require file system based locking or some other form of ipc).

@elsirion https://github.com/fedimint/fedimint/blob/d0fabfb79182dcc3f487f340179c60885c9caa6b/fedimint-testing/src/btc/real.rs#L191 it's implemented

bradleystachurski (commented):
Pulled the latest (https://github.com/fedimint/fedimint/pull/3589/commits/d4c933bc140cda3967176d8bf9cf3689bd621ad2) and everything passes locally 🎉 

Likely flakiness causing the timeout of `fedimint-wallet-tests::fedimint_wallet_tests peg_ins_that_are_unconfirmed_are_rejected`. Perhaps kickoff a rerun of CI and we can merge soon?

elsirion (approved):

PR #3590, opened 2023-11-14
[WIP] fix: CI backend test connections

bradleystachurski:
WIP
- kicking off CI
- fixes connection issues, still others to resolve
- builds off https://github.com/fedimint/fedimint/pull/3589

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -33,6 +33,7 @@ pub struct Bitcoind {
 
 impl Bitcoind {
     pub async fn new(processmgr: &ProcessManager) -> Result<Self> {
+        println!("inside Bitcoind::new");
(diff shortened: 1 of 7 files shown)


elsirion:
I think the extra logging you introduced could be valuable, but should happen via `tracing::{warn,info,debug,trace}`.

bradleystachurski:
> I think the extra logging you introduced could be valuable, but should happen via `tracing::{warn,info,debug,trace}`.

Oh interesting, I can reopen. I'm hesitant to keep the url/auth logs for bitcoind, but I can keep the others.

bradleystachurski:
@elsirion I thought it was cleaner to open a separate PR (https://github.com/fedimint/fedimint/pull/3611) instead of reopening this one

PR #3594, opened 2023-11-14
Reactivate tests

elsirion:
Alternative to  #3589, relevant context for TODOs:

> After using the old `cargo test` based version of running these tests only some gateway tests fail:
> 
> ```
> 00:01:02     can_switch_active_gateway
> 00:01:02     cannot_pay_same_external_invoice_twice
> 00:01:02     makes_internal_payments_within_federation
> 00:01:02     rejects_wrong_network_invoice
> ```
> 
> All fail with the same error:
> 
> ```
> 00:01:02 thread 'can_switch_active_gateway' panicked at /build/source/fedimint-testing/src/gateway.rs:151:10:
> 00:01:02 Gateway failed to start: Gateway did not reach desired state within 30 seconds
> ```
> 
> This indicates we always get stuck here:
> 
> https://github.com/fedimint/fedimint/blob/256bdfada2cfe7716f58e5e119ab248418158a18/fedimint-testing/src/gateway.rs#L145-L151
> 
> Did we maybe change the GW setup flow in a way that silently breaks devimint and just makes the GW not come online @okjodom, @m1sterc001guy?

### DIFF
--- a/fedimint-bitcoind/src/lib.rs
+++ b/fedimint-bitcoind/src/lib.rs
@@ -9,6 +9,7 @@ use anyhow::Context;
 pub use anyhow::Result;
 use bitcoin::{BlockHash, Network, Script, Transaction, Txid};
 use fedimint_core::bitcoinrpc::BitcoinRpcConfig;
+use fedimint_core::fmt_utils::OptStacktrace;
 use fedimint_core::task::TaskHandle;
 use fedimint_core::txoproof::TxOutProof;
 use fedimint_core::util::SafeUrl;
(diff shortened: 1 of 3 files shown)


okjodom:
Oh these tests are racing to establish htlc connection with the same LND instance from devimint and we don't know how to do that. Only one of the tests will manage a successful connection, and the rest are guaranteed to fail!

dpc:
This is not new and know issue: https://github.com/fedimint/fedimint/issues/3209

My understanding was these tests worked because once a gateway gets to be an interceptor it effectively locks out all other gateways from doing so.

> All ln gateway tests want to be an interceptor. Tests try to run in parallel. Effectively ln-gateway only work because they keep trying until they register.

See more in the issue.

okjodom:
Proposed a patch that uses CLN gateway as the basis of the tests: https://github.com/elsirion/fedimint/pull/1
If we can't to keep using LND gateway before a fix to #3596 , we can consider the following patches:
- Serially run all `fedimint-ln-tests`
- Increase timeout for gateway state to change from `30s` to about `100s`

elsirion:
https://github.com/fedimint/fedimint/pull/3589 got merged

PR #3601, opened 2023-11-14
feat(load-test-tool): LN circular load-test

douglaz:
This load test is focused on the lightning payments and is intended to be run for long periods of time:
```rust
    /// Run a load test where many users in parallel will receive then send a
    /// payment through lightning.
    /// It's 'circular' because the funds always come back to the same user then
    /// we can keep making the payments in a loop
```

It comes in three flavors (strategies) that may help stressing different parts of the gateway:

```rust
enum LnCircularStrategy {
    /// The user will pay its own invoice
    SelfPayment,
    /// One gateway will pay/receive to/from the other, then they will swap
    /// places
    TwoGateways,
    /// Two clients will pay to each other using the same gateway
    PartnerPingPong,
}
```
This load test was fundamental to reproduce the issues on https://github.com/fedimint/fedimint/issues/3478

### DIFF
--- a/devimint/src/vars.rs
+++ b/devimint/src/vars.rs
@@ -125,6 +125,7 @@ declare_vars! {
 
         FM_GATEWAY_API_ADDR: String = f!("http://127.0.0.1:{FM_PORT_GW_CLN}");
         FM_GATEWAY_PASSWORD: String = "theresnosecondbest";
+        FM_GATEWAY_FEES: String = "0,0"; // Enable to us to make an unbounded number of payments
 
         FM_CLN_EXTENSION_LISTEN_ADDRESS: String = f!("0.0.0.0:{FM_PORT_CLN_EXTENSION}");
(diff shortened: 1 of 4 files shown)


devimint/src/vars.rs
```diff
@@ -125,6 +125,7 @@ declare_vars! {
 
         FM_GATEWAY_API_ADDR: String = f!("http://127.0.0.1:{FM_PORT_GW_CLN}");
         FM_GATEWAY_PASSWORD: String = "theresnosecondbest";
+        FM_GATEWAY_FEES: String = "0,0"; // Enable to us to make an unbounded number of payments
```
elsirion, line 128: Not ideal to not test fees anymore, but ok.

elsirion (approved):

devimint/src/vars.rs
douglaz: It makes sense for automated tests but it doesn't make sense for manual tests

Issue #1298, opened 2023-01-12
Use `SafeUrl` all over the code


dpc:
To help with accidentally exposing auth data in urls we use accross the code. [Original discussion](https://github.com/fedimint/fedimint/pull/1288#discussion_r1068517727)

dpc:
Yeah, so 

```rust
use url::Url;

fn main() {
    let url = Url::parse("ftp://foo:bar@localhost").unwrap();
    println!("{url}, {url:?}")
}
```

Prints:

```
ftp://foo:bar@localhost/, Url { scheme: "ftp", cannot_be_a_base: false, username: "foo", password: Some("bar"), host: Some(Domain("localhost")), port: None, path: "/", query: None, fragment: None }
```

so we need a wrapper.

dpc:
Fixed by #1419

dpc:
We should use the type from #1419 everywhere. This might require some extra `impl`s to make it work in all sorts of places, but should be relatively easy to do.

Aankirz:
Hey @dpc , I want to work on this issue can you pls assign this to me ?, 

thanks

justinmoon:
Give it a try @Aankirz. Try to just find one place where this helps and make a PR for that. To make sure you're on the right track.

cbergqvist:
Hope you don't mind me grabbing this. Felt like one of the more straightforward issues and no apparent progress for half a year.

shaurya947:
It's a bit inconvenient that there isn't an `Into` or `From` to create a `SafeUrl` type from a `Url` type. Leaving this here so we can add that before closing this issue.

dpc:
It should be possible to turn `Url` into `SafeUrl` with `From`/`Into`, but the other way around it should require some `to_unsafe` (or maybe something else as `unsafe` has specific meaning ... thought I guess this ship has already sailed with `Safe` in the `SafeUrl`...)

cbergqvist:
See I'm still assigned to this. Can do some follow-up changes to #3133 in the suggested directions in the following days.

cbergqvist:
Let me know if there's anything more you think should be done before closing this issue.

PR #3602, opened 2023-11-15
refactor: Rename `SafeUrl::reap_guts()` -> `SafeUrl::to_unsafe()`

cbergqvist:
Also add one-way `From`-trait implementation.

Per discussions in #1298.

### DIFF
--- a/fedimint-core/src/util/mod.rs
+++ b/fedimint-core/src/util/mod.rs
@@ -85,7 +85,7 @@ impl SafeUrl {
 
     /// Warning: This removes the safety.
     // nosemgrep: ban-raw-url
-    pub fn reap_guts(self) -> Url {
+    pub fn to_unsafe(self) -> Url {
         self.0
     }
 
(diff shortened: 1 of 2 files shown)


fedimint-core/src/util/mod.rs
```diff
@@ -85,7 +85,7 @@ impl SafeUrl {
 
     /// Warning: This removes the safety.
     // nosemgrep: ban-raw-url
```
dpc, line 87: This suggest there's a semgrep rule that should be adjusted as well.

cbergqvist: It's still something that should be discouraged, but not be made impossible. Could you please elaborate what you envision - making the rule also trigger on uses of `to_unsafe()` or what?

elsirion (approved):
But the colorful language!

fedimint-core/src/util/mod.rs
dpc:
```
- id: ban-raw-url
  languages:
    - rust
  message: "`url::Url` may expose confidential username + passwords to logs etc. Use `fedimint_core::SafeUrl` instead."
  pattern: url::Url
  severity: WARNING
```

OK, so I guess the rule protects from using `url::Url` only, so it doesn't need to change.

Nvmd then. Let's keep as is for now.

Issue #3477, opened 2023-10-27
Add a lot of debug logging to the gateway


justinmoon:
Today @douglaz, @okjodom and I were trying to debug an issue with the gateway and we noticed that logging is very insufficient. This is understandable because we completely re-wrote the gateway losing some useful logging that accumulated in the old gateway, and we haven't run it in production much yet.

m1sterc001guy:
Any suggestions on what to add? We don't currently log on successful payments iirc, we should probably add that.

This is probably a good first issue.

douglaz:
I would consider some things:
- Track of payments, both success and failure with easy to read logs. Today we have some debug statements that shows part of the state machine where we can search for a given amount for instance, but it's very hard to read. It would be good to be able to grep the whole process of the payment since the beginning until its success/failure. Perhaps we should attach some temporary id to every logging statement
- Related to above, perhaps we could have some pretty print of state machines in general
- If there is a loop, we must know if the loop is stuck or not (more important for loop {...} statements than for ... in ...)
- If it makes a I/O or RPC call, we must know if it is attempting to make that call and if it failed or not. Also we should consider adding timeouts here to detect and proper log "blocking" or slow calls

All note that the logging shouldn't be excessively verbose (i.e it must be compatible with DEBUG level, we can't rely on TRACE level for production purposes)

PR #3605, opened 2023-11-15
feat: more debugging logs on gateway

douglaz:
Closes https://github.com/fedimint/fedimint/issues/3477

It's certainly not perfect, but we can add/remove stuff on future iterations.

### DIFF
--- a/gateway/ln-gateway/src/state_machine/pay.rs
+++ b/gateway/ln-gateway/src/state_machine/pay.rs
@@ -21,6 +21,7 @@ use lightning_invoice::Bolt11Invoice;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio_stream::StreamExt;
+use tracing::{debug, error, info, warn};
 
 use super::{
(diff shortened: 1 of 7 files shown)


gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -404,36 +422,38 @@ impl GatewayPayInvoice {
...
+                    Some(contract) => GatewayPayStateMachine {
+                        common,
+                        state: GatewayPayStates::CancelContract(Box::new(
+                            GatewayPayCancelContract { contract, error: e },
+                        )),
+                    },
+                    None => GatewayPayStateMachine {
+                        common,
+                        state: GatewayPayStates::OfferDoesNotExist(e.contract_id),
+                    },
```
elsirion, line 454: Change in behavior: We now allow cancellation of contracts if an internal payment fails (which it never should), seems ok.

elsirion (approved):
Thx! This should make debugging much easier!

PR #3609, opened 2023-11-15
feat(mint): make `OOBNotes` forwards-compatible

### LINKED ISSUE #3568: Version `OOBNotes` serialization

elsirion:
Same idea as #3595, see that for an explanation of the design.

Fixes #3568 (implicit versioning, but even better: extensibility without breaking old clients)

### DIFF
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -13,6 +13,7 @@ use std::cmp::Ordering;
 use std::collections::BTreeMap;
 use std::ffi;
 use std::fmt::{Display, Formatter};
+use std::io::Read;
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -80,10 +81,84 @@ pub const LOG_TARGET: &str = "client::module::mint";
 /// An encapsulation of [`FederationId`] and e-cash notes in the form of
 /// [`TieredMulti<SpendableNote>`] for the purpose of spending e-cash
 /// out-of-band. Also used for validating and reissuing such out-of-band notes.
+///
+/// ## Invariants
```
dpc, line 85: Would be nice if we had it in code. 

Make a `fs check_invariants`, `impl Decodable` manually to use it, etc.?

dpc (approved):

Issue #3571, opened 2023-11-10
Add minor consensus versions


elsirion:

justinmoon:
call: if we add functionality in backwards-compatible way, we need a way for the federation to signal whether this is supported or not

PR #3610, opened 2023-11-15
Make mint module associated types enums for extensibility

### LINKED ISSUE #3570: Make all module associated types enums for extensibility
Depends on #3569

elsirion:
Part of #3570

The strategy implemented here is:
* Make input/output/output outcome/consensus item types enums with only one main `V0` variant for now and a `Default` variant that is used when encountering an unknown variant on deserialization.
* **Server:** throw error when encountering `Default` variants as we don't understand the semantics
* **Client:**
  * **Recovery:** Ignore `Default` variants and keep recovering with the ones we understand
  * **Other places:** We only deal with self-generated transactions, panic on `Default`
* If we ever need to change functionality we can add a new `V1` variant in a new minor consensus version (#3571) without breaking old clients

Should I proceed with this approach and introduce it to all modules? @dpc @douglaz @joschisan


dpc (approved):
Makes sense to me.

douglaz:
Looks reasonable

dpc (approved):
:shipit:

elsirion:
@dpc I have two more modules in the pipeline, but if you want we can already merge this first part, I won't get to all the modules today.

dpc:
The faster we land, the merge conflict resolution we need to do.

dpc (approved):

PR #3615, opened 2023-11-16
Let us properly pay HOLD invoices

douglaz:
- Moves the ln payment from the state transition to the trigger, avoiding all the state machines to get stuck while paying HOLD invoices
Closes https://github.com/fedimint/fedimint/issues/3478
- Adds a devimint test where we pay HOLD invoices and test the above scenario
- Adds `--finish-in-background` flag to `fedimint-cli ln-pay` to let us pay HOLD invoices without blocking the terminal
- Adds a `fedimint-cli await-ln-pay` to let us get the result of a ln payment in case it was interrupted or called using the above flag
- Fixes an error on the gateway when we try to resume a ln payment
Closes https://github.com/fedimint/fedimint/issues/3639
- Minor code cleanups

### DIFF
--- a/gateway/ln-gateway/src/state_machine/pay.rs
+++ b/gateway/ln-gateway/src/state_machine/pay.rs
@@ -49,7 +49,7 @@ use crate::state_machine::GatewayClientModule;
 ///    CancelContract -- cancel tx submission successful --> Canceled
 ///    CancelContract -- cancel tx submission unsuccessful --> Failed
 /// ```
-#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable)]
+#[derive(Debug, Clone, Eq, PartialEq, Decodable, Encodable, Serialize, Deserialize)]
 pub enum GatewayPayStates {
     PayInvoice(GatewayPayInvoice),
     CancelContract(Box<GatewayPayCancelContract>),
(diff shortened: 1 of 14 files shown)


gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -183,20 +183,104 @@ impl GatewayPayInvoice {
...
+            Self::fetch_parameters_and_pay(
                 global_context,
-                self.pay_invoice_payload.contract_id,
+                payload.clone(),
                 context.clone(),
+                common.clone(),
             ),
-            move |_dbtx, result, _old_state| {
-                Box::pin(Self::transition_buy_preimage(
+            move |_dbtx, result, _old_state| Box::pin(futures::future::ready(result)),
```
elsirion, line 192: The more canonical way of implementing trigger future and state transition would be to only give the result to the state transition function and then constructing the new state there. Was there any particular reason not to do it that way?

douglaz: Yeah, it's a bit unorthodox, but this was the natural way of doing it after the refactoring. The problem is that the output of this function is very complex due to the different outcomes and the best representation are the state structures  themselves.

But are there any issues doing this way? I couldn't find.

elsirion: There aren't any direct problems with it, just a bit confusing.

dpc (approved):

dpc (approved):

douglaz:
@elsirion @m1sterc001guy @okjodom Merging now to avoid more code conflicts, but you're welcome to do more rounds of reviews. I can address any comments or issues in future PRs

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -182,20 +182,104 @@ impl GatewayPayInvoice {
...
+            .await
+        } else {
+            Self::buy_preimage_over_lightning(
+                context,
+                payment_parameters,
+                contract.clone(),
+                common.clone(),
+            )
+            .await
+        }
```
elsirion, line 282: Buying the preimage in the trigger future is the correct way to go imo, but we need to consider cancel safety: are we sure that if the trigger future gets interrupted and re-run no errors are thrown that would make the state machine fail/refund while the payment was actually sent?

douglaz: I'll try to add a test to cover this scenario

douglaz: I've tested by changing the code to call  `fetch_parameters_and_pay` twice instead of once. And all the tests and load tests works. I'll create an issue to try to convert this scenario into a proper automated test.

PR #3616, opened 2023-11-16
feat(fedimint-cli): lock the database with advisory fs locks

dpc:
This makes semi-concurrent use possible, as `fedimint-cli` will just wait for a different process to finish.

Can be tested in `just mprocs` by running a couple of `for i in $(seq 100); do fedimint-cli info ;done &` at the same time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1216,6 +1216,7 @@ name = "fedimint-cli"
 version = "0.2.0-alpha"
 dependencies = [
  "anyhow",
+ "async-trait",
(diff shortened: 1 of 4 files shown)


elsirion (approved):
Really cool! I tried it out and works as expected. I wonder if there should be a timeout though?

dpc:
@elsirion After https://github.com/cargo-bins/cargo-binstall/pull/1496 is released we could improve and add options to fail instead of waiting and/or waiting a certain time.

PR #3617, opened 2023-11-16
feat(devimint): dev-fed --exec <cmd> 

dpc:
This should simplify running scripts against federation.

Setup and cleanup happens automatically, no need to run things in the background, call `env` etc.

Longer scripts can be executed by making them a separate file or using exported bash functions, simulated below:

```bash
#!/usr/bin/env bash

function fake_devimimint_exec() {
  >&2 echo "Starting deviming with: ${*}"
  export FM_DIR=some
  "$@"
}


echo "Starting our test suite"

export EXTERNAL_VARIABLE=yo
function some_test_block() {
  echo "TEST is running with FM_DIR=$FM_DIR EXTERNAL_VARIABLE=$EXTERNAL_VARIABLE"
}
export -f some_test_block

export EXTERNAL_VARIABLE=ye
fake_devimimint_exec bash -c some_test_block```

```> bash test-long.sh
Starting our test suite
Starting deviming with: bash -c some_test_block
TEST is running with FM_DIR=some EXTERNAL_VARIABLE=ye```

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -1,7 +1,7 @@
-use std::env;
 use std::ops::ControlFlow;
 use std::path::PathBuf;
 use std::time::{Duration, Instant};
+use std::{env, ffi};
 
 use anyhow::{anyhow, Context, Result};
 use bitcoincore_rpc::bitcoin;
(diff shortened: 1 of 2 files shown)


elsirion (approved):

Maan2003 (commented):
:tada:

PR #3625, opened 2023-11-16
chore: get rid of some noisy error level log messages 

dpc:


fedimint-logging/src/lib.rs
```diff
@@ -68,8 +68,15 @@ impl TracingSetup {
...
         use tracing_subscriber::fmt::writer::{BoxMakeWriter, Tee};
-        let filter_layer =
-            EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
+
+        let var = env::var(tracing_subscriber::EnvFilter::DEFAULT_ENV).unwrap_or_default();
+        let filter_layer = EnvFilter::builder().parse(format!(
+            // We prefix everything with a default general log level and
+            // good per-module specific default. User provided RUST_LOG
+            // can override one or both
+            "info,jsonrpsee_core::client::async_client=off,{}",
```
elsirion: Should we tune down AlephBFT the same way?

elsirion (approved):

fedimint-logging/src/lib.rs
dpc: Done already. :)

elsirion: I did not expect you to still be online :see_no_evil:

PR #3629, opened 2023-11-17
cleanup: remove ability to return multiple pks per input

joschisan:
We previously had to return multiple keys per input since we had multiple ecash notes within the same input. However, we switched to single ecash note per input to reduce complexity - furthermore it seems like partially spendable inputs are generally not a good idea which is why we want to enforce it by only allowing a single pub key per input now.


elsirion (approved):
Thx! So much cleaner!

PR #3634, opened 2023-11-17
 feat: make lightning associated types extensible 

elsirion:
Another step towards #3570


elsirion:
We don't seem to have migration tests for any of these structs? At least I didn't need to regenerate the migration snapshot DB.

dpc (approved):

PR #3636, opened 2023-11-17
Gateway leave federation

Kodylow:
Adds RPC command to leave federation.

Leave-fed removes federation key from gateway-db and session-db, if you reconnect to the federation while still having ecash the ecash is still there.

Tested with Eric leaving federation, restarting gateway, and rejoining.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -28,7 +28,9 @@ use bitcoin::{Address, Network, Txid};
 use bitcoin_hashes::hex::ToHex;
 use clap::{Parser, Subcommand};
 use client::GatewayClientBuilder;
-use db::{DbKeyPrefix, GatewayConfiguration, GatewayConfigurationKey, GatewayPublicKey};
+use db::{
+    DbKeyPrefix, FederationIdKey, GatewayConfiguration, GatewayConfigurationKey, GatewayPublicKey,
+};
 use fedimint_client::module::init::ClientModuleInitRegistry;
 use fedimint_client::ClientArc;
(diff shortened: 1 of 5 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -840,6 +842,18 @@ impl Gateway {
...
+    async fn handle_leave_federation(&mut self, payload: LeaveFedPayload) -> Result<()> {
+        self.remove_client(payload.federation_id).await?;
+        let mut dbtx = self.gateway_db.begin_transaction().await;
+        dbtx.remove_entry(&FederationIdKey {
+            id: payload.federation_id,
+        })
+        .await;
+        dbtx.commit_tx_result()
+            .await
+            .map_err(GatewayError::DatabaseError)
```
elsirion, line 854: We explicitly don't delete the client DB here since that could lead to loss of funds. This means joining the same federation again will get you all the old state back. If that's not what we want we could just rename the DB folder into something like `<fed-id>.bak[.<index>]`.

elsirion (approved):
I did some pair-programming with @Kodylow and think this is good, would still appreciate a second pair of eyes :)

gateway/ln-gateway/src/lib.rs
okjodom: > We explicitly don't delete the client DB here since that could lead to loss of funds. This means joining the same federation again will get you all the old state back.

Makes sense to me

> ... we could just rename the DB folder into something like `<fed-id>.bak[.<index>]`.

Seems like this could be under backup/restore apis, and `leave-federation` would do a backup

okjodom (approved):
🚀

PR #3641, opened 2023-11-17
fix: calculate tx fees using vbytes

### LINKED ISSUE #3534: Investigate potential overpaying fees by ~4x

bradleystachurski:
Fixes #3534 

- A user reported observing 4x the fees for an on-chain withdrawal compared to what should have been paid using `bitcoin-cli estimatesmartfee`
- We've been calculating the tx fee as `sats_per_vbyte * tx weight` instead of `sats_per_vbyte * tx_vbytes`
- `tx_weight ~== 4 * tx_vbytes`, hence we're overpaying fees 4x


elsirion (approved):
Thx for investigating and the high-quality fix!

PR #3642, opened 2023-11-17
 refactor: remove (some) need for manual module id handling 

dpc:

PR #3644, opened 2023-11-18
experiment: Eliminated `DatabaseTransacationRef` by (ab-)using lifetimes

dpc:
On top of #3642

Having a separate type for commitable vs non-commitable is a bit of a pain that I've hit during refactoring, as there's no good way to generalize over the two, and adding another layer of traits seems like a nightmare.

It occurred to me that by keeping an extra lifetime (basically  "to parent tx", where `'static` means "no parent") it would be possible to have `commit(&mut self)` only on `DatabaseTransaction<'static, 'tx>`.

After a lot of grinding I was able to make it work. From the user perspective it's not even a bad UI (an extra `'_` on the type).

Unfortunately when dealing with certain non-trivial core logic around commiting (mostly internal), it gets extremely hard to work with, as it requires dealing with things like: https://users.rust-lang.org/t/argument-requires-that-is-borrowed-for-static/66503/2?u=yandros

I need to sleep on it, and think a bit if this complication is worth it. I also consider just getting rid of `DatabaseTransactionRef` and making the whole thing a `bool`, so trying to `commit` on a reference dbtx just panics. At certain point one needs to side with simplicity/

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -864,7 +864,7 @@ impl Client {
         let autocommit_res = self
             .db
             .autocommit(
-                |dbtx, _| {
+                move |dbtx, _| {
                     let operation_type = operation_type.clone();
                     let tx_builder = tx_builder.clone();
                     let operation_meta = operation_meta.clone();
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -757,7 +757,7 @@ impl Client {
 
     pub async fn add_state_machines(
         &self,
-        dbtx: &mut DatabaseTransaction<'_>,
+        dbtx: &mut DatabaseTransaction<'_, '_>,
         states: Vec<DynState<DynGlobalClientContext>>,
     ) -> AddStateMachinesResult {
         self.executor.add_state_machines_dbtx(dbtx, states).await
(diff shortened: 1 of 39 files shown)


dpc:
@elsirion @m1sterc001guy Let me know what you think.


It just occurred to me that session types might work better than a lifetime here. Basically `commit` would be available only on `DatabaseTransaction<'tx, Commitable>`, operations that "create a non-commitable reference" would return `DatabaseTransaction<'tx, NonCommitable>`, the type would default to `DatabaseTransaction<'tx, T = NonCommitable>`, and only code that needs Commitable capabilities would need to change `DatabaseTransaction<'tx>` to `DatabaseTransaction<'tx, Commitable>` or do generics over T.

So 4 choices:

* keep separate `DatabaseTransactionRef<'tx>`
* lifetimes (this PR)
* runtime check and panic
* session types

elsirion:
I'm a big fan of session types and actually saw your lifetime solution as a special case of that concept. Personally I'd like to avoid having to mess with complicated lifetime bounds, especially since you need the `PhantomBound` hack that becomes part of our public API.

dpc:
It's a shame that I realized that sessions types will be better only after I did this and writing a late PR description for this.

But on a positive note, I practiced my lieftime-fu ( I feel like I still know nothing).

dpc:
#3647 seems indeed better.

PR #3647, opened 2023-11-19
chore: get rid of `DatabaseTransactionRef`

dpc:
A (hopefully) better approach than #3644.

Having a separate type for commitable vs non-commitable is a bit of a pain that I've hit during refactoring, as there's no good way to generalize over the two, and adding another layer of traits seems like a nightmare.

This change unifies `DatabaseTransactionRef`
with `DatabseTransaction` using session types to express committability.

This makes generalizing overr `DatabaseTransaction` easier, though it's not exactly perfect either. But I believe it's a bit better.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -354,7 +354,7 @@ impl Database {
         Ok(())
     }
 
-    /// Begin a database transaction
+    /// Begin a new committable database transaction
     pub async fn begin_transaction<'s, 'tx>(&'s self) -> DatabaseTransaction<'tx, Committable>
     where
         's: 'tx,
@@ -363,6 +363,14 @@ impl Database {
             self.inner.begin_transaction().await,
             self.module_decoders.clone(),
         )
+    }
+
+    /// Begin a new non-committable database transaction
+    pub async fn begin_transaction_nc<'s, 'tx>(&'s self) -> DatabaseTransaction<'tx, NonCommittable>
+    where
+        's: 'tx,
+    {
+        self.begin_transaction().await.into_nc()
     }
 
     /// Runs a closure with a reference to a database transaction and tries to
@@ -418,7 +426,7 @@ impl Database {
 
             let mut dbtx = self.begin_transaction().await;
 
-            let val = tx_fn(&mut dbtx.to_ref_non_committable(), PhantomData)
+            let val = tx_fn(&mut dbtx.to_ref_nc(), PhantomData)
                 .await
                 .map_err(|err| AutocommitError::ClosureError {
                     attempts: curr_attempts,
@@ -1290,7 +1298,7 @@ fn decode_value<V: DatabaseValue>(
 
 impl<'tx, Cap> DatabaseTransaction<'tx, Cap> {
     /// Convert into a non-committeable version
-    pub fn into_non_committable(self) -> DatabaseTransaction<'tx, NonCommittable> {
+    pub fn into_nc(self) -> DatabaseTransaction<'tx, NonCommittable> {
         DatabaseTransaction {
             tx: self.tx,
             decoders: self.decoders,
@@ -1301,11 +1309,11 @@ impl<'tx, Cap> DatabaseTransaction<'tx, Cap> {
     }
 
     /// Get a reference to a non-committeable version
-    pub fn to_ref_non_committable<'s, 'a>(&'s mut self) -> DatabaseTransaction<'a, NonCommittable>
+    pub fn to_ref_nc<'s, 'a>(&'s mut self) -> DatabaseTransaction<'a, NonCommittable>
     where
         's: 'a,
     {
-        self.to_ref().into_non_committable()
+        self.to_ref().into_nc()
     }
 
     /// Get [`DatabaseTransaction`] isolated to a `prefix`
@@ -1817,7 +1825,7 @@ pub async fn apply_migrations(
 
         while current_db_version < target_db_version {
             if let Some(migration) = migrations.get(&current_db_version) {
-                migration(&mut dbtx.to_ref_non_committable()).await?;
+                migration(&mut dbtx.to_ref_nc()).await?;
             } else {
                 panic!("Missing migration for version {current_db_version}");
             }
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -37,6 +37,8 @@
 use std::collections::BTreeMap;
 use std::error::Error;
 use std::fmt::{self, Debug};
+use std::marker::{self, PhantomData};
+use std::ops::{self, DerefMut};
 use std::pin::Pin;
 use std::sync::Arc;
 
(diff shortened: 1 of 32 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -385,13 +392,17 @@ impl Database {
...
         &'s self,
         tx_fn: F,
         max_attempts: Option<usize>,
     ) -> Result<T, AutocommitError<E>>
     where
-        for<'a> F: Fn(&'a mut DatabaseTransaction<'dt>) -> BoxFuture<'a, Result<T, E>>,
+        's: 'dbtx,
+        for<'r, 'o> F: Fn(
+            &'r mut DatabaseTransaction<'o>,
+            PhantomBound<'dbtx, 'o>,
```
dpc, line 404: I can not get rid of it, but I'm not certain it's necessary.

dpc: I think the reason why `DatabaseTransactionRef` worked before is that the `into_non_commitable` is introducing an extra lifetime which introduces the problem described in the article linked in `PhantomBound` doc comment, but it's a bit hard to reason about.

dpc: After we land I might just ask on user forums and maybe someone can confirm / fix.

modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -508,7 +512,9 @@ async fn peg_ins_that_are_unconfirmed_are_rejected() -> anyhow::Result<()> {
         .await;
     let block_count = dyn_bitcoin_rpc.get_block_count().await?;
     sync_wallet_to_block(
-        &mut dbtx.dbtx_ref_with_prefix_module_id(module_instance_id),
+        &mut dbtx
+            .to_ref_with_prefix_module_id(module_instance_id)
+            .into_non_committable(),
```
dpc, line 517: Comparing to lifetime version, this is a bit worse. Lifetimes figured it out automatically, while here we need to convert the type somehow. We could `DerefMut` from commitable to non-commitable maybe.

fedimint-core/src/db/mod.rs
elsirion: Could we have a `BorrowMut` impl instead? Since we always seem to pass mut references `DerefMut` would be inconvenient if I'm not mistaken as it requires dereferencing first before creating a reference again?

dpc: @elsirion I don't understand why. `Deref` and `DerefMut` are the only two operators that are implicit, which makes it tempting to use it for all sorts of automatic conversion. The callee could already take `dbtx: DatabaseTransaction<Cap>`, which would make the call work for both `Committable` and `Uncommittable` and save the caller the conversion from `Commitable` to `Uncommittable`, but that requires extra work and remembering about.


One thing I can immediately improve is introduce `beging_transaction_ro` that returns `DatabaseTransaction<Uncommittable` since RO transactions don't ever need `commit_tx`. That should cut down number of conversion by maybe 75% in the user code, since lots of txes are RO.

dpc:
@m1sterc001guy ping just in case.

fedimint-core/src/db/mod.rs
dpc: @elsirion I started introducing `.begin_transaction_ro` and while it works, and does eliminate need for `.into_non_commitable` it's a bit weird that it' still possible to call `insert` etc. on them, even if they technically can't modify anything because the transaction will never commit...

Not sure what to think about it.

dpc: Maybe it should be `begin_transaction_non_committable` or something like that instead.

dpc: Also what would be nice is if "committable" wasn't 100 characters long with m and t repeating 5 times. Maybe we could abbreviate everywhere to `nc` (kind of like `ro` for ReadOnly)?

fedimint-server/src/db.rs
```diff
@@ -184,8 +184,6 @@ mod fedimint_migration_tests {
             .await;
 
         let _consensus_items = vec![ConsensusItem::Transaction(transaction)];
-
-        dbtx.commit_tx().await;
```
dpc, line 188: This is now responsibility of a caller. It just makes more sense.

fedimint-core/src/db/mod.rs
elsirion: I stand corrected, `Deref` magic is stronger than I thought, played around with it a bit and seems like the sensible thing to do indeed. The only annoyance is that you can only deref to exactly one target type, but it seems unlikely that we will want more target types in this case.

It might still be useful to have explicit functions to get committable or non-commitable versions in case we implement generic logic where the type marker cannot be inferred.

elsirion: Actually, does `Deref` even work in this case without some unsafe `transmute`ing? The function sig is `Deref::deref(&self) -> &Self::Target`, so we get a reference and want to change the type of that reference, but cannot create a new struct with the correct type either because that will go out of scope on return. So the only option I see is transmuting the reference? Or am I missing something?

fedimint-core/src/module/mod.rs
```diff
@@ -155,7 +155,7 @@ impl ApiError {
 /// State made available to all API endpoints for handling a request
 pub struct ApiEndpointContext<'dbtx> {
     db: Database,
-    dbtx: DatabaseTransaction<'dbtx>,
+    dbtx: DatabaseTransaction<'dbtx, Committable>,
     has_auth: bool,
     request_auth: Option<ApiAuth>,
 }
```
elsirion, line 161: A longer term goal is getting rid of writable DBTXns in API endpoints to avoid writing non-consensus data to consensus-relevant DB columns. One way of implementing this would be supplying a non-committable consensus DBTX and a committable non-consensus DBTX that uses another prefix.

fedimint-server/src/db.rs
elsirion: Not committing felt wrong when I saw it because we drop the DBTX at the end of the scope and it's not obvious that it's just a pointer. I think in most other place we pass around mut refs anyway, so it isn't a problem. How do others think about this?

modules/fedimint-dummy-client/src/lib.rs
```diff
@@ -161,7 +161,7 @@ impl ClientModule for DummyClientModule {
         stream.next_or_pending().await
     }
 
-    async fn get_balance(&self, dbtc: &mut DatabaseTransactionRef<'_>) -> Amount {
+    async fn get_balance(&self, dbtc: &mut DatabaseTransaction<'_>) -> Amount {
```
elsirion, line 164: preexistent typo: `s/dbtc/dbtx/`

elsirion (approved):
Looks good, I will take some time to get fully used to the new API, especially with the implicit/default `NonCommittable` type parameter. But seems like an improvement overall.

fedimint-core/src/db/mod.rs
dpc: > So the only option I see is transmuting the reference? Or am I missing something?

Oh, right. We can't do `Deref` without `transmute`, and I'm too scared to use `transmute`

dpc:
> Looks good, I will take some time to get fully used to the new API, especially with the implicit/default `NonCommittable` type parameter. But seems like an improvement overall.

I'm open to adjustement. Seemed like most of the code needs only `NonCommittable` so a good default. Also we could have type aliases for both versions if it feels handy. @elsirion

elsirion:
Let's just merge it, especially the type alias can be done later if we feel it's needed.

PR #3651, opened 2023-11-20
 refactor: stop ClientModule from accessing own module_id (mostly)

dpc:
On top of #3650

PR #3655, opened 2023-11-20
Introduce module associated type for backups

elsirion:
Fixes #3654

* Makes it easier to add versioning to module backups by making them enums
* Probably enables removing `supports_backup` since there's now the `NoBackup` placeholder type (does this sound correct @dpc?)

PR #3659, opened 2023-11-20
 refactor: ClientModule::backup to use ClientContext 

dpc:
On top of #3651


elsirion (approved):
Will have to rebase #3655, but got some test debugging to do anyway.

PR #3669, opened 2023-11-21
feat: restart federation setup

### LINKED ISSUE #3535: RPC method to restart setup process
Sometimes setups fail in a way where they could still succeed if the process is retried. An RPC method to restart the setup process would help UI implement such a restart flow https://github.com/fedimint/ui/issues/253.

okjodom:
After setting config gen params and before starting consensus, any peer can initiate a restart of the federation setup using `restart_federation_setup` rpc.

> **server-side behavior**

- This API sets the peer server in `SetupRestarted` status.
- The Leader server waits for all peers to progress to `SetupRestarted`.
- Leader then automatically makes progress to `AwaitingPassword`.
- Followers watch leader status and when they see `AwaitingPassword`, they also make progress to the same state

All peers can then begin the setup process

> **suggested user experience**

While the server is in any of these states `SharingConfigGenParams` | `ReadyForConfigGen` | `ConfigGenFailed` |  `VerifyingConfigs` | `VerifiedConfigs`, the setup UI should watch peer information in consensus status. Should any peer be in `SetupRestarted`, the setup UI should immediately prompt the user to initiate restart by make a call to `restart_federation_setup`. Suggestion here it to make this some blocking but interactive UX

closes #3535

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::fs;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex, MutexGuard};
@@ -19,10 +20,10 @@ use fedimint_core::core::ModuleInstanceId;
 use fedimint_core::db::Database;
 use fedimint_core::endpoint_constants::{
     ADD_CONFIG_GEN_PEER_ENDPOINT, AUTH_ENDPOINT, CONFIG_GEN_PEERS_ENDPOINT,
-    CONSENSUS_CONFIG_GEN_PARAMS_ENDPOINT, DEFAULT_CONFIG_GEN_PARAMS_ENDPOINT, RUN_DKG_ENDPOINT,
-    SET_CONFIG_GEN_CONNECTIONS_ENDPOINT, SET_CONFIG_GEN_PARAMS_ENDPOINT, SET_PASSWORD_ENDPOINT,
-    START_CONSENSUS_ENDPOINT, STATUS_ENDPOINT, VERIFIED_CONFIGS_ENDPOINT,
-    VERIFY_CONFIG_HASH_ENDPOINT,
+    CONSENSUS_CONFIG_GEN_PARAMS_ENDPOINT, DEFAULT_CONFIG_GEN_PARAMS_ENDPOINT,
+    RESTART_FEDERATION_SETUP_ENDPOINT, RUN_DKG_ENDPOINT, SET_CONFIG_GEN_CONNECTIONS_ENDPOINT,
+    SET_CONFIG_GEN_PARAMS_ENDPOINT, SET_PASSWORD_ENDPOINT, START_CONSENSUS_ENDPOINT,
+    STATUS_ENDPOINT, VERIFIED_CONFIGS_ENDPOINT, VERIFY_CONFIG_HASH_ENDPOINT,
 };
 use fedimint_core::module::{
     api_endpoint, ApiAuth, ApiEndpoint, ApiEndpointContext, ApiError, ApiRequestErased,
@@ -35,7 +36,9 @@ use tokio::sync::mpsc::Sender;
 use tokio_rustls::rustls;
 use tracing::{error, info};
 
-use crate::config::io::{read_server_config, write_server_config, PLAINTEXT_PASSWORD, SALT_FILE};
+use crate::config::io::{
+    read_server_config, write_server_config, CONFIG_STAGING_DIR, PLAINTEXT_PASSWORD, SALT_FILE,
+};
 use crate::config::{gen_cert_and_key, ConfigGenParams, ServerConfig};
 use crate::net::peers::DelayCalculator;
 use crate::{check_auth, get_verification_hashes, ApiResult, HasApiContext};
@@ -281,7 +284,7 @@ impl ConfigGenApi {
                     let mut state = self_clone.state.lock().expect("lock poisoned");
                     match config {
                         Ok(config) => {
-                            self_clone.write_configs(&config, &state)?;
+                            self_clone.stage_configs(&config, &state)?;
                             state.status = ServerStatus::VerifyingConfigs;
                             state.config = Some(config);
                             info!(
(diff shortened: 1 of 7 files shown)


fedimint-server/src/config/api.rs
```diff
@@ -94,6 +94,17 @@ impl ConfigGenApi {
         Ok(state)
     }
 
+    fn require_status_maybe(
```
elsirion, line 100: nit: why call it `_maybe`? Isn't this more of a `require_any_status` fn?

fedimint-server/src/config/io.rs
```diff
@@ -123,3 +123,19 @@ fn encrypted_json_write<T: Serialize + DeserializeOwned>(
...
+    }
+    Ok(())
+}
+
+pub fn remove_server_config(path: PathBuf) -> anyhow::Result<()> {
+    remove_existing_file(path.join(LOCAL_CONFIG).with_extension(JSON_EXT))?;
+    remove_existing_file(path.join(CONSENSUS_CONFIG).with_extension(JSON_EXT))?;
+    remove_existing_file(path.join(CLIENT_INVITE_CODE_FILE))?;
+    remove_existing_file(path.join(CLIENT_CONFIG).with_extension(JSON_EXT))?;
+    remove_existing_file(path.join(PRIVATE_CONFIG).with_extension(ENCRYPTED_EXT))?;
```
elsirion, line 139: This is very brittle if we ever add any files. In another PR we should introduce a list of files that FM potentially creates and have a test that checks that data dirs contain only these and nothing more.

fedimint-server/src/config/api.rs
```diff
@@ -388,6 +399,86 @@ impl ConfigGenApi {
...
+        let self_clone = self.clone();
+        let mut sub_group = self.task_group.make_subgroup().await;
+        sub_group
+            .spawn("restart", move |_handle| async move {
+                if let Some(client) = leader {
+                    // Followers wait for leader to signal that all peers have restarted setup
+                    loop {
+                        let status = client.status().await.map_err(|_| {
+                            ApiError::not_found("Unable to connect to the leader".to_string())
+                        })?;
```
elsirion, line 445: Shouldn't we keep re-trying if there is a network error?

fedimint-server/src/config/api.rs
```diff
@@ -388,6 +399,86 @@ impl ConfigGenApi {
...
+                        if status.server == ServerStatus::AwaitingPassword {
+                            break;
+                        }
```
elsirion, line 448: Let's hope the leader isn't a speed-typer xD

elsirion (dismissed):
Looks good, does anyone more familiar with the setup take a look? @justinmoon

fedimint-server/src/config/api.rs
```diff
@@ -295,17 +309,16 @@ impl ConfigGenApi {
...
-                    ServerStatus::VerifyingConfigs,
-                    ServerStatus::VerifiedConfigs
-                ))
-            })?;
-
+        let expected_status = [
+            ServerStatus::VerifyingConfigs,
+            ServerStatus::VerifiedConfigs,
+        ];
+        let state = self.require_status_maybe(&expected_status).map_err(|_| {
```
justinmoon, line 316: nice refactor 👍

fedimint-server/src/config/api.rs
```diff
@@ -378,6 +399,86 @@ impl ConfigGenApi {
...
+                ))
+            })?;
+
+            self.remove_configs()?;
+            state.status = ServerStatus::SetupRestarted;
+            info!(
+                target: fedimint_logging::LOG_NET_PEER_DKG,
+                "Update config gen status to 'Setup restarted'"
+            );
+            // Create a WSClient for the leader
```
justinmoon, line 425: Why not just set the state to like `AwaitingPassword` or whatever the initial state is and return right here?

fedimint-server/src/config/io.rs
okjodom: tracking here #3692

fedimint-server/src/config/api.rs
okjodom: Makes it much harder for UX to determine clear intent to restart setup. We'd need to add an additional info to `AwaitingPassword` to differentiate first time setup vs restart of setup. Alternatively, `SetupRestarted` is a clear state to signal request to restart by one of the peers, and we can watch it to see which peers have accepted the restart

fedimint-server/src/config/api.rs
okjodom: now checking for `AwaitingPassword` | `SharingConfigGenParams` just in case

fedimint-server/src/config/api.rs
okjodom: renamed

fedimint-server/src/config/api.rs
okjodom: allowed for 3 retries

elsirion:
CI is failing and it seems related to the PR

okjodom:
@elsirion , now green on CI

fedimint-server/src/config/api.rs
```diff
@@ -378,6 +389,93 @@ impl ConfigGenApi {
...
+                    let mut state = self_clone.state.lock().expect("lock poisoned");
+                    state.status = ServerStatus::AwaitingPassword;
+                    info!(
+                        target: fedimint_logging::LOG_NET_PEER_DKG,
+                        "Update config gen status to 'Awaiting password'"
+                    );
+                }
+                self_clone.update_leader().await
+            })
+            .await;
```
elsirion, line 475: nit(readability): would be easier to comprehend if split into functions

elsirion (approved):
Looks good to me, but I'll leave @justinmoon a chance to take another look

elsirion:
One concern is that we could accidentally delete important key material if we ever called these delete fns by accident somewhere. To reduce that problem I'd propose to `chmod -w` config files right before starting the consensus, so that deletes are prevented.

dpc:
> One concern is that we could accidentally delete important key material if we ever called these delete fns by accident somewhere. To reduce that problem I'd propose to `chmod -w` config files right before starting the consensus, so that deletes are prevented.

That's a good point, but probably best practice to keep the in-progress things in a temp directory (randomized) and `rename` it atomically on success.

okjodom:
> > One concern is that we could accidentally delete important key material if we ever called these delete fns by accident somewhere. To reduce that problem I'd propose to `chmod -w` config files right before starting the consensus, so that deletes are prevented.
> 
> That's a good point, but probably best practice to keep the in-progress things in a temp directory (randomized) and `rename` it atomically on success.

marked as draft while applying this

elsirion:
Are you on this @okjodom? @justinmoon should this get into 0.2?

okjodom:
> Are you on this @okjodom? @justinmoon should this get into 0.2?

Trying to debug the failing reconnect test now. Been up and down

I think an api changed?

okjodom:
Ps, this was originally marked for 0.2.1
I bumped it after first round of reviews because thinking I'd be able to land it fast

okjodom:
@elsirion, @justinmoon  I'm tracking two issues as follow-ups to this. Take a look

okjodom:
fixes #3788

fedimint-server/src/config/api.rs
```diff
@@ -378,6 +428,103 @@ impl ConfigGenApi {
...
+            .spawn("restart", move |_handle| async move {
+                if let Some(client) = leader {
+                    self_clone.await_leader_restart(client).await?;
+                } else {
+                    self_clone.await_peer_restart().await;
+                }
+                // Progress status to AwaitingPassword
+                {
+                    let mut state = self_clone.state.lock().expect("lock poisoned");
+                    state.status = ServerStatus::AwaitingPassword;
```
elsirion, line 474: Shouldn't we completely reset the `ConfigGenState` here? Otherwise, wouldn't a leader still use the same set of peers that had registered previously?

fedimint-server/src/config/api.rs
```diff
@@ -1026,6 +1190,99 @@ mod tests {
...
+        // peer, and any peer could trigger a restart before the leader does.
+        for peer in &followers {
+            peer.client
+                .restart_federation_setup(peer.auth.clone())
+                .await
+                .ok();
+        }
+
+        leader.wait_status(ServerStatus::SetupRestarted).await;
+        leader.wait_status(ServerStatus::AwaitingPassword).await;
```
elsirion, line 1286: Can we test if a second try of the setup will actually run through? E.g. also test what happens if one peer had a misconfigured API/p2p address with which they registered with the leader and need to correct it.

okjodom: now rerunning setup steps after restart

fedimint-server/src/config/api.rs
```diff
@@ -384,6 +434,99 @@ impl ConfigGenApi {
...
+            info!(
+                target: fedimint_logging::LOG_NET_PEER_DKG,
+                "Update config gen status to 'Setup restarted'"
+            );
+            // Create a WSClient for the leader
+            state
+                .local
+                .clone()
+                .and_then(|local| local.leader_api_url.map(WsAdminClient::new))
+        };
```
elsirion, line 462: What's the reason to put this into a separate block?

fedimint-server/src/config/api.rs
```diff
@@ -571,6 +714,20 @@ impl ConfigGenState {
...
+        self.auth = None;
+        self.requested_params = None;
+        self.status = ServerStatus::AwaitingPassword;
+        self.local = None;
+
+        info!(
+            target: fedimint_logging::LOG_NET_PEER_DKG,
+            "Reset config gen state"
+        );
+    }
```
elsirion, line 730: Do we need to reset any on-disk state? I think the `PASSWORD` file has been written already at that point?

elsirion (approved):
I don't think this is stability-relevant, we don't need to guarantee stability of the setup API, just need to release the admin UI in lockstep (but any reasonable deployment should use the admin UI that fits the Fedimint version).

I also didn't find any Rust API breakage, so in principle this could be backported to 0.2 if we saw the need (I'd like to avoid that though).

fedimint-server/src/config/api.rs
okjodom: this allows us to acquire lock on `self.state`, use it and and releasing lock, before we call `update_leader`. update leader creates a separate lock on state

fedimint-server/src/config/api.rs
okjodom: at this point, `PASSWORD` or any other config file is already written to a file in a temporary dir. `remove_staged_config()` resets disk state by deleting the whole dir, and this function finally resets server state

justinmoon (approved):
LGTM

fedimint-server/src/config/api.rs
elsirion: I didn't remember that `require_any_status` returned a mutex guard, makes sense!

elsirion:
Thx for bearing with us, the release took a lot of bandwidth and this PR a few iterations, but I think we arrived at a good result at the end!

PR #3671, opened 2023-11-21
chore: remove musig2 signature compression for transaction signatures

joschisan:
The signature compression via musig relied on unstable code. We therefore switch to naive multisignatures for 0.2 and will make the signature an extendible enum to maybe enable a compressed variant in the future. Furthermore this allows us to use PublicKeys instead of XOnlyPublicKeys.

### DIFF
--- a/fedimint-core/src/transaction.rs
+++ b/fedimint-core/src/transaction.rs
@@ -1,12 +1,10 @@
 use bitcoin::hashes::Hash as BitcoinHash;
 use bitcoin::XOnlyPublicKey;
-use bitcoin_hashes::hex::ToHex;
 use fedimint_core::core::{DynInput, DynOutput};
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::module::SerdeModuleEncoding;
 use fedimint_core::{Amount, TransactionId};
-use rand::Rng;
-use secp256k1_zkp::{schnorr, Secp256k1, Signing, Verification};
+use secp256k1_zkp::schnorr;
 use thiserror::Error;
 
 use crate::core::{DynInputError, DynOutputError};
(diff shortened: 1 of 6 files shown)


elsirion:
Can you add the reasoning from our call as description please?

fedimint-core/src/transaction.rs
```diff
@@ -30,8 +28,8 @@ pub struct Transaction {
     /// In the future the nonce can be used for grinding a tx hash that fulfills
     /// certain PoW requirements.
     pub nonce: [u8; 8],
-    /// Aggregated MuSig2 signature over all the public keys of the inputs
-    pub signature: Option<schnorr::Signature>,
+    /// signatures for all the public keys of the inputs
+    pub signatures: Vec<schnorr::Signature>,
```
elsirion, line 32: TODO: open issue to make extensible

fedimint-core/src/transaction.rs
```diff
@@ -65,116 +63,37 @@ impl Transaction {
...
+                .is_err()
+            {
+                return Err(TransactionError::InvalidSignature {
+                    tx: self.consensus_encode_to_hex().expect("Can't fail"),
+                    hash: self
+                        .tx_hash()
+                        .consensus_encode_to_hex()
+                        .expect("Can't fail"),
+                    sig: signature.consensus_encode_to_hex().expect("Can't fail"),
+                    key: pk.consensus_encode_to_hex().expect("Can't fail"),
```
elsirion, line 90: Unrelated to this PR: Lol, why are these `String`s?!

elsirion (approved):

fedimint-core/src/transaction.rs
joschisan: Found this confusing as well

PR #3674, opened 2023-11-21
chore(devimint): increase bitcoin rpc timeout on tests

douglaz:
Due to this timeout some tests were failing on my local machine. There is no reason for it to be so low.


dpc (approved):

dpc:
LGTM, but I kind of expect the actual problem here is we're hitting default limit of rpc thread on bitcoind.

We should increase it to something like 64, and maybe also add some retries.

douglaz:
> LGTM, but I kind of expect the actual problem here is we're hitting default limit of rpc thread on bitcoind (AFAIR: 4).
> 
> We should increase it to something like 64, and maybe also add some retries.

I'm not sure this is the issue, but increased `rpcworkqueue` and `rpcthreads` anyway

dpc (approved):

PR #3675, opened 2023-11-21
patch: partially disable intermittently failing gateway test

okjodom:
In [test_gateway_executes_swaps_between_connected_federations](https://github.com/fedimint/fedimint/blob/fdf7edd27162b139b98bdca137fbb38ab2d27818/gateway/ln-gateway/tests/integration_tests.rs#L1109C47-L1109C47), claiming the outgoing contract seems takes too long(?) and asserting gateway balances after a swap intermittently fails.

this change disables the assertion, while we investigate why the balances sometimes fail to update as expected

### DIFF
--- a/gateway/ln-gateway/tests/integration_tests.rs
+++ b/gateway/ln-gateway/tests/integration_tests.rs
@@ -1184,13 +1184,17 @@ async fn test_gateway_executes_swaps_between_connected_federations() -> anyhow::
             assert_matches!(claimed, LnReceiveState::Claimed);
             assert_eq!(client2.get_balance().await, invoice_amt);
 
+            // FIXME: https://github.com/fedimint/fedimint/issues/3672
+            // This assertion intermittently fails and is a blocker on the CI pipeline
+            // we need to investigate further why the gateway balance may not be updating
+            //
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1185,12 +1185,12 @@ async fn test_gateway_executes_swaps_between_connected_federations() -> anyhow::
             assert_eq!(client2.get_balance().await, invoice_amt);
 
             // Check gateway balances after facilitating direct swap between federations
-            let post_balances = get_balances(&rpc, &[id1, id2]).await;
-            assert_eq!(
-                post_balances[0],
-                pre_balances[0] + (invoice_amt + fee).msats
-            );
-            assert_eq!(post_balances[1], pre_balances[1] - invoice_amt.msats);
+            // let post_balances = get_balances(&rpc, &[id1, id2]).await;
```
dpc, line 1188: Is there an issue + comment would be great.

okjodom: added comment

dpc (approved):

PR #3678, opened 2023-11-21
fix: debug logging for fetch gateways without meta overrides

okjodom:
when client is fetching gateway list and attempting to apply vetted gateway meta, we should debug log cases where there is no meta override url, or if there's no configured meta override for the federation

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -868,11 +868,17 @@ impl LightningClientModule {
 
     async fn fetch_meta_overrides(&self, config: ClientConfig) -> anyhow::Result<String> {
         let federation_id = config.global.federation_id();
-        let override_src = FederationInfo::from_config(config)
+        let override_src = match FederationInfo::from_config(config)
             .await?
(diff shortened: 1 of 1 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -909,12 +915,16 @@ impl LightningClientModule {
...
 
-                m.get(&federation_id.to_string())
-                    .ok_or(anyhow::anyhow!(
-                        "No meta overrides for federation id: {}",
-                        federation_id
-                    ))?
-                    .to_string()
+                match m.get(&federation_id.to_string()) {
+                    Some(meta) => {
+                        debug!("Found meta overrides for federation: {}", federation_id);
```
douglaz, line 920: nit:
```suggestion
                        debug!("Found meta overrides for federation: {federation_id}");
```

douglaz (approved):
LGTM

modules/fedimint-ln-client/src/lib.rs
okjodom: fixed nit

douglaz (approved):

PR #3684, opened 2023-11-22
feat: provide more control in restore_from_backup to clients

Maan2003:
made restore_from_backup public
made it accept backup as argument as instead of downloading itself.


elsirion (approved):
Very useful! I actually had to do something similar for debugging my modularized backup PR but didn't have the energy left to PR.

PR #3688, opened 2023-11-22
fix(client): only list operation ids with associated operations as active

### LINKED ISSUE #3687: Remove workaround for `get_active_operations` returns oeprations IDs not associated with operations
In particular caused by the special case of recovery state machines. Thx @Maan2003 for reporting :)

elsirion:
Fixes #3687. Tl;dr: `get_active_operations` returned the recovery operation id for which no operation log entry exists.

@dpc since you want to replace the state machines for recovery anyway and @Maan2003 has a workaround for now, do we want this fix? It likely makes the function much slower, but also prevents this from happening again in the future if SMs get used for other tasks.


dpc:
We can land and remove and improve later.

Issue #2977, opened 2023-08-17
Rearchitect modularized backup recovery


elsirion:
# **Edit**: TODO checklist:

* [x] [main refactoring change unblocking other improvements](https://github.com/fedimint/fedimint/pull/4035)
* [x] [make it go brrrr](https://github.com/fedimint/fedimint/pull/4042) (stream history)
* [x] [make it reuse history](https://github.com/fedimint/fedimint/pull/4046) (downloaded epochs) so multiple modules don't duplicate work
* [x] [do something about weird `add_state_machines_inactive_dbtx`](https://github.com/fedimint/fedimint/pull/4035#discussion_r1455020761)

# Previous discussion

In the design described below a badly implemented module could stall the recovery of all other modules and we decided against it. I leave it here for historic context. For the actual proposal see https://github.com/fedimint/fedimint/issues/2977#issuecomment-1819626967

**OUTDATED:**

The current recovery interface was built in a hurry to bring the new client library to feature-parity with the old one so that a switch could happen. Unfortunately this lead to a design with multiple shortcomings mostly stemming from the fact that recovery is modeled as a module state machine and the different modules cannot easily interact.

* **Speed**: The state machine executor is not built for heavy loads and computation cannot be easily parallelized. https://github.com/fedimint/fedimint/issues/2784
* **Storage Requirements**: to mitigate loss of funds and allow for recovery in case of client bugs the state machine executor keeps a log of all intermediate states of a state machine. The large recovery state bloats that log and would need special handling, making the executor more complex.
* **No Shared Epoch Download**: All modules need to scan the epoch history since the last backup for items belonging to the module/user. Currently each module would need to download the entire history itself to look for its item, resulting in a lot of unnecessary traffic.
* **Unclear Client State**: While the recovery process is running the `Client` object already exists, but it is a programmer error to call any module methods that could trigger DB writes and thus interfere with recovery. Instead a recovering client should be a separate type.

## Proposed Architecture

I already talked to @dpc about the following new architecture:

* The `build_restoring_from_backup` returns a `RecoveringClient` struct
  * That struct has very limited external functionality:
    * `RecoveringClient::subscribe_progress() -> watch::Receiver<Progress>`: A way to track recovery progress
    * `impl Future<Output=Result<Client>>`: the `RecoveringClient` is a future that can be awaited and returns the client on successful completion of recovery
  * Internally, it consists of:
[...]

PR #3689, opened 2023-11-22
fix(wallet-client): deactivate secret derivation

elsirion:
Deriving keys from a root key without implementing recovery from backup is dangerous. Keys might be re-used since the last used key index isn't recovered. To avoid such problems we reactivate key derivation till recovery is implemented for the wallet module.

This will take #2977+implementing module recovery for the wallet off the critical path for 0.2.0.

### DIFF
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -155,9 +155,16 @@ impl ClientModuleInit for WalletClientInit {
             .0
             .clone()
             .unwrap_or(WalletClientModule::get_rpc_config(args.cfg()));
+
+        // FIXME: reactivate key derivation once we implement recovery
+        let random_root_secret = {
+            let (key, salt): ([u8; 32], [u8; 32]) = thread_rng().gen();
+            DerivableSecret::new_root(&key, &salt)
+        };
+
         Ok(WalletClientModule {
             cfg: args.cfg().clone(),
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -155,9 +155,16 @@ impl ClientModuleInit for WalletClientInit {
...
+        // FIXME: reactivate key derivation once we implement recovery
+        let random_root_secret = {
+            let (key, salt): ([u8; 32], [u8; 32]) = thread_rng().gen();
+            DerivableSecret::new_root(&key, &salt)
+        };
+
         Ok(WalletClientModule {
             cfg: args.cfg().clone(),
-            module_root_secret: args.module_root_secret().clone(),
+            module_root_secret: random_root_secret,
```
dpc, line 167: Just to double check - we're fine with this being random *on every start*?

dpc (approved):

modules/fedimint-wallet-client/src/lib.rs
elsirion: The only usage is here:

https://github.com/fedimint/fedimint/blob/d5ade3ee36ba748268686cd4235cb358d3a639b3/modules/fedimint-wallet-client/src/lib.rs#L284-L288

Previously this was always a random key. It being derived helped @douglaz with recovering from the timeout bug, but we could have extracted the key from the state machine log as well.

PR #3694, opened 2023-11-22
fix: make consensus fast again

### LINKED ISSUE #3682: Make AlephBFT parameters configurable and find more performant defaults
In manual testing AlephBFT turned out to be slower than HBBFT on low-latency networks. This is likely related to the value chosen for `ROUND_DELAY`.

Both `ROUND_DELAY` and `EXPECTED_ROUNDS_PER_SESSION` should probably be configurable, so that guardians can tweak the performance to their typical network conditions.

elsirion:
Payment latency on current master (4a6a580f4eeccc7f95baa8c693c2f5ae52900a24):

```
18 gateway_pay_invoice_success: avg 5.68456783s, median 5.652043845s, max 11.529719549s, min 4.11351511s
18 gateway_create_invoice: avg 1.168869878s, median 2.123909154s, max 3.395704957s, min 19.291994ms
9 gateway_payment_received_started: avg 4.464089591s, median 4.046802124s, max 9.151851322s, min 3.32866744s
2 reissue_notes: avg 2.494038092s, median 2.630916954s, max 2.630916954s, min 2.35715923s
```

Payment latency with 95deafc7b529836785059617ace5ffc49a9ea64c:

```
2 reissue_notes: avg 1.481310515s, median 1.600800859s, max 1.600800859s, min 1.361820172s
17 gateway_payment_received_started: avg 1.958627338s, median 1.884378943s, max 2.91291775s, min 1.761244498s
34 gateway_create_invoice: avg 636.523164ms, median 1.181066008s, max 1.281812029s, min 18.758735ms
34 gateway_pay_invoice_success: avg 2.957889402s, median 3.209708471s, max 4.322514205s, min 2.326591537s
```

Fixes #3682

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -326,19 +326,17 @@ impl ConsensusServer {
     }
 
     pub async fn run_session(&self, session_index: u64) -> anyhow::Result<()> {
-        // if all nodes are correct the session will take 45 to 60 seconds. The
-        // more nodes go offline the longer the session will take to complete.
-        const EXPECTED_ROUNDS_PER_SESSION: usize = 45 * 20;
         // this constant needs to be 3000 or less to guarantee that the session
         // can never reach MAX_ROUNDs.
-        const EXPONENTIAL_SLOWDOWN_OFFSET: usize = 3 * EXPECTED_ROUNDS_PER_SESSION;
-        const MAX_ROUND: u16 = 5_000;
-        const ROUND_DELAY: f64 = 50.0;
+        let exponential_slowdown_offset: usize =
+            3 * self.cfg.consensus.broadcast_expected_rounds_per_session as usize;
+        let max_round = self.cfg.consensus.broadcast_max_rounds_per_session;
+        let round_delay = self.cfg.local.broadcast_round_delay_ms as f64;
         const BASE: f64 = 1.01;
 
         // this is the minimum number of unit data that will be ordered before we reach
         // the EXPONENTIAL_SLOWDOWN_OFFSET even if f peers do not attach unit data
-        let batches_per_session = EXPECTED_ROUNDS_PER_SESSION * self.keychain.peer_count();
+        let batches_per_session = exponential_slowdown_offset * self.keychain.peer_count();
 
         // In order to bound a sessions RAM consumption we need to bound its number of
         // units and therefore its number of rounds. Since we use a session to
@@ -350,12 +348,12 @@ impl ConsensusServer {
         // In case of such an attack the broadcast stops ordering any items until the
         // attack subsides as not items are ordered while the signatures are collected.
         let mut delay_config = aleph_bft::default_delay_config();
-        delay_config.unit_creation_delay = std::sync::Arc::new(|round_index| {
+        delay_config.unit_creation_delay = std::sync::Arc::new(move |round_index| {
             let delay = if round_index == 0 {
                 0.0
             } else {
-                ROUND_DELAY
-                    * BASE.powf(round_index.saturating_sub(EXPONENTIAL_SLOWDOWN_OFFSET) as f64)
+                round_delay
+                    * BASE.powf(round_index.saturating_sub(exponential_slowdown_offset) as f64)
             };
 
             Duration::from_millis(delay.round() as u64)
@@ -365,7 +363,7 @@ impl ConsensusServer {
             self.keychain.peer_count().into(),
             self.keychain.peer_id().to_usize().into(),
             session_index,
-            MAX_ROUND,
+            max_round,
             delay_config,
             Duration::from_secs(100 * 365 * 24 * 60 * 60),
         )
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -328,12 +328,12 @@ impl ConsensusServer {
     pub async fn run_session(&self, session_index: u64) -> anyhow::Result<()> {
         // if all nodes are correct the session will take 45 to 60 seconds. The
         // more nodes go offline the longer the session will take to complete.
-        const EXPECTED_ROUNDS_PER_SESSION: usize = 45 * 4;
+        const EXPECTED_ROUNDS_PER_SESSION: usize = 45 * 20;
         // this constant needs to be 3000 or less to guarantee that the session
         // can never reach MAX_ROUNDs.
         const EXPONENTIAL_SLOWDOWN_OFFSET: usize = 3 * EXPECTED_ROUNDS_PER_SESSION;
-        const MAX_ROUND: u16 = 5000;
-        const ROUND_DELAY: f64 = 250.0;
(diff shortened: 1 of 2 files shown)


fedimint-server/src/consensus/server.rs
```diff
@@ -328,12 +328,12 @@ impl ConsensusServer {
     pub async fn run_session(&self, session_index: u64) -> anyhow::Result<()> {
         // if all nodes are correct the session will take 45 to 60 seconds. The
         // more nodes go offline the longer the session will take to complete.
-        const EXPECTED_ROUNDS_PER_SESSION: usize = 45 * 4;
+        const EXPECTED_ROUNDS_PER_SESSION: usize = 45 * 20;
         // this constant needs to be 3000 or less to guarantee that the session
         // can never reach MAX_ROUNDs.
         const EXPONENTIAL_SLOWDOWN_OFFSET: usize = 3 * EXPECTED_ROUNDS_PER_SESSION;
```
elsirion, line 334: We are getting close to this limit, is this a concern @joschisan? I tried to up `MAX_ROUND` to 10k to have more safety margin again, but that lead to unclear errors.

fedimint-server/src/consensus/server.rs
```diff
@@ -328,12 +328,12 @@ impl ConsensusServer {
...
-        const MAX_ROUND: u16 = 5000;
-        const ROUND_DELAY: f64 = 250.0;
+        const MAX_ROUND: u16 = 5_000;
+        const ROUND_DELAY: f64 = 50.0;
```
elsirion, line 336: I wonder if we could calculate all the values from a given round delay and a target length of one session. That would make changing these values easier.

douglaz:
This is much better

elsirion:
Would be good to see how it performs in an actual deployment, but if I understand these parameters correctly the worst that could happen is that the slowest guardian with >50ms latency to its peers will never get any CIs confirmed.

fedimint-server/src/consensus/server.rs
joschisan: Getting close to the limit is not a concern. No safety margin necessary here.

fedimint-server/src/consensus/server.rs
joschisan: The thing is that MAX_ROUND, ROUND_DELAY, EXPECTED_ROUNDS the number of peers and the maximum size of a unit all play into the maximal throughput of kb of txs per second, maximal ram consumption per session, session time bounds and latency. Besides ROUND_DELAY we need consensus on all constants so they can not be adapted after the federation is running. So tuning this is quite tricky here and can not be changed afterwards, therefore I opted to set a sensible default instead of exposing this complexity to the guardians. I think it will probably make sense to make this configurable in the future, especially when can simplify this interplay by upstreaming a small change to Aleph BFT.

joschisan:
> Would be good to see how it performs in an actual deployment, but if I understand these parameters correctly the worst that could happen is that the slowest guardian with >50ms latency to its peers will never get any CIs confirmed (shouldn't be a problem with non-byzantine faults, we need to add back banning so we are resistant against byzantine ones in that case).

Yes that is correct. Bringing back automatic banning via consensus like before would be very tricky with aleph, we would have to upstream changes to do it I believe.

joschisan (approved):

douglaz:
Unfortunely it still happens:
```
36 gateway_pay_invoice_success: avg 3.011156321s, median 3.13429771s, max 7.595922224s, min 2.496601472s
```

elsirion:
Under how much load? Just the normal latency test? How is this so different from my results? Maybe differences in the thread scheduler? How often did you have to run it to get this?

douglaz:
> Under how much load? Just the normal latency test? How is this so different from my results? Maybe differences in the thread scheduler? How often did you have to run it to get this?

I ran like 5 times and got this result 4 times. I've increased the test time to 200s:
```
fedimint-load-test-tool --users 1 ln-circular-load-test --test-duration-secs 200 --strategy two-gateways
```

elsirion:
I ran the same command 5 times (reordered the output lines for easier comparison):

Round 1
```
58 gateway_pay_invoice_success: avg 2.900430064s, median 3.191206754s, max 3.71039477s, min 2.22849282s
58 gateway_create_invoice: avg 636.233331ms, median 1.190670509s, max 1.277283363s, min 18.549844ms
29 gateway_payment_received_started: avg 1.91697157s, median 1.887153069s, max 2.325132243s, min 1.769363503s
2 reissue_notes: avg 1.490606604s, median 1.591824943s, max 1.591824943s, min 1.389388266s
```

Round 2
```
58 gateway_pay_invoice_success: avg 2.878919148s, median 3.124184415s, max 3.959212133s, min 2.225206058s
58 gateway_create_invoice: avg 639.651312ms, median 1.189635651s, max 1.270131503s, min 19.075474ms
29 gateway_payment_received_started: avg 1.900186942s, median 1.880769644s, max 2.53698865s, min 1.722318664s
2 reissue_notes: avg 1.494801728s, median 1.630906615s, max 1.630906615s, min 1.358696841s
```

Round 3
```
56 gateway_pay_invoice_success: avg 2.87683378s, median 3.176470078s, max 3.65178892s, min 2.294883115s
56 gateway_create_invoice: avg 727.634487ms, median 1.194285317s, max 6.212039184s, min 18.645532ms
28 gateway_payment_received_started: avg 1.901550745s, median 1.888899789s, max 2.222559379s, min 1.754060976s
2 reissue_notes: avg 1.507199002s, median 1.636300947s, max 1.636300947s, min 1.378097058s
```

Round 4
```
56 gateway_pay_invoice_success: avg 2.945463356s, median 3.176235666s, max 7.503406603s, min 2.277732146s
56 gateway_create_invoice: avg 634.789697ms, median 1.168618746s, max 1.274456762s, min 18.984688ms
28 gateway_payment_received_started: avg 1.888088496s, median 1.894764102s, max 2.11020239s, min 1.761036663s
2 reissue_notes: avg 1.501823348s, median 1.616952068s, max 1.616952068s, min 1.386694628s
```

Round 5
```
56 gateway_pay_invoice_success: avg 2.977994837s, median 3.16884684s, max 7.778382406s, min 2.276854256s
56 gateway_create_invoice: avg 623.751514ms, median 1.182066618s, max 1.263180005s, min 18.625988ms
28 gateway_payment_received_started: avg 1.9142979s, median 1.926075169s, max 2.005058872s, min 1.77389033s
2 reissue_notes: avg 1.50978187s, median 1.599912936s, max 1.599912936s, min 1.419650804s
```

So I see this behavior too, just much less frequently. Might have something to do with how long session switches take. @joschisan how long do you think could we reasonably make sessions while keeping the latency as low as right now?

joschisan:
Iirc we could enable 3-4 times longer sessions with aleph bft right now. If we can upstream a small change we get another multiple of 3 without increasing maximal RAM consumption.

dpc:
This seems to cause #3782

PR #3697, opened 2023-11-22
test: re-enable gateway swap balance assertion

### LINKED ISSUE #3679: Reactivate partially disabled intermittently failing gateway test
See #3675

okjodom:
asserting gateway balance after swap seems sensitive to consensus +  gateway state
 machine performance. this change adds a 5s sleep, hopefully giving the gateway a chance to claim
outgoing contract and update balances before we assert new balances

fixes #3679 
drafting this as a fix but we'd need correlation with latency test data from #3696

### DIFF
--- a/gateway/ln-gateway/tests/integration_tests.rs
+++ b/gateway/ln-gateway/tests/integration_tests.rs
@@ -1184,17 +1184,28 @@ async fn test_gateway_executes_swaps_between_connected_federations() -> anyhow::
             assert_matches!(claimed, LnReceiveState::Claimed);
             assert_eq!(client2.get_balance().await, invoice_amt);
 
-            // FIXME: https://github.com/fedimint/fedimint/issues/3672
-            // This assertion intermittently fails and is a blocker on the CI pipeline
-            // we need to investigate further why the gateway balance may not be updating
-            //
             // Check gateway balances after facilitating direct swap between federations
-            // let post_balances = get_balances(&rpc, &[id1, id2]).await;
-            // assert_eq!(
-            //     post_balances[0],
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1184,17 +1184,16 @@ async fn test_gateway_executes_swaps_between_connected_federations() -> anyhow::
             assert_matches!(claimed, LnReceiveState::Claimed);
             assert_eq!(client2.get_balance().await, invoice_amt);
 
-            // FIXME: https://github.com/fedimint/fedimint/issues/3672
-            // This assertion intermittently fails and is a blocker on the CI pipeline
-            // we need to investigate further why the gateway balance may not be updating
-            //
+            // Give the gateway a chance to update it's balances
+            sleep(Duration::from_secs(5)).await;
```
justinmoon, line 1188: You could speed this up by polling the balance for 5 seconds, and fail the test if it doesn't update in 5 seconds, but continue if it does.

justinmoon (dismissed):

gateway/ln-gateway/tests/integration_tests.rs
okjodom: updated

found it non-intuitive that we have to do
```rs
if post_balances[0] == pre_balances[0] && post_balances[1] == pre_balances[1] { // error balances not updated }
```
and not
```rs
if post_balances == pre_balances { // error balances not updated }
```
because we are watching perf and latencies in two distinct federations

justinmoon (approved):

elsirion (approved):

PR #3704, opened 2023-11-23
fix: rederive root_secret using federation ID

### LINKED ISSUE #3508: Derive root secret that is actually used using federation id
Using the same actual root secret in two federation could have catastrophic consequences (leaking the secrets), depending on cryptography used underneath.

It would be safest if we eliminated the possibility completely, but always "hashing-in" the federation id inside `Client`, so the caller providing the `root_secret` can't forget about it.

shaurya947:
Closes #3508

Also removes `Client::external_secret` because we can now supply a derivable secret `root->child_client_0` to the client while we use `root->other_use_case` for our external needs. When the client still managed the secret this function was the only way to use the same secret as the client.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1589,6 +1589,14 @@ impl ClientBuilder {
 
         let final_client = FinalClient::default();
 
+        // Re-derive client's root_secret using raw bytes from the provided root_secret
+        // and salt from the federation ID. This eliminates the possibility of having
+        // the same client root_secret across multiple federations.
+        let root_secret = DerivableSecret::new_root(
+            &root_secret.to_random_bytes::<32>(),
+            &config.global.federation_id().0,
+        );
+
         let modules = {
             let mut modules = ClientModuleRegistry::default();
             for (module_instance, module_config) in config.modules.clone() {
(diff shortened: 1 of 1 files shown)


fedimint-client/src/lib.rs
```diff
@@ -1589,6 +1589,14 @@ impl ClientBuilder {
...
         let final_client = FinalClient::default();
 
+        // Re-derive client's root_secret using raw bytes from the provided root_secret
+        // and salt from the federation ID. This eliminates the possibility of having
+        // the same client root_secret across multiple federations.
+        let root_secret = DerivableSecret::new_root(
+            &root_secret.to_chacha20_poly1305_key_raw(),
+            &config.global.federation_id().0,
+        );
+
```
shaurya947, line 1599: 1. Is this the correct approach—to get deterministic bytes from the provided `root_secret` and salt it with the byte representation of federation ID?
2. Is it fine to reuse `DerivableSecret::to_chacha20_poly1305_key_raw()` for this purpose? I also see `DerivableSecret::to_random_bytes()` but the "random" in the name made me a bit nervous about it being deterministic (even though it looks like it is?)

elsirion: `to_random_bytes` would be better imo, less confusing. The randomness there is deterministic, i.e. it's using a PRNG.

shaurya947: Updated

elsirion (approved):

PR #3716, opened 2023-11-24
WIP: chore: update consensus versions

elsirion:
Release TODO list:
- [x] Do we want to stay on master with the release? It seems to make sense if we want to keep backwards compatibility anyway? Or do we still want to fork off to be able to deploy smaller patches without having to go through testing for a bunch of unrelated changes?
  * Yes, let's go with a release branch, I'll re-open this PR to the new `releases/v0.2` branch tomorrow
- [x] Increment consensus versions so old clients don't try to talk to new federations
  * Do we need to increment module consensus version numbers? Probably not since core makes them incompatible anyway, but it doesn't hut to stay in sync for now imo.
- [ ] Set rust crate version to `0.2.0`
  * Right after releasing we probably want to push a change to `0.3.0-alpha` to distinguish the master version from the release
  * Rust versions will diverge from consensus versions due to SemVer and rust interfaces changing independently from consensus.
  * `cargo release` can hopefully automate most of this
- [ ] Following the release: adapt tutorial/docker files with newest tags

### DIFF
--- a/modules/fedimint-ln-server/src/lib.rs
+++ b/modules/fedimint-ln-server/src/lib.rs
@@ -235,7 +235,7 @@ impl ServerModuleInit for LightningInit {
     }
 
     fn supported_api_versions(&self) -> SupportedModuleApiVersions {
-        SupportedModuleApiVersions::from_raw((u32::MAX, 0), (0, 0), &[(0, 0)])
+        SupportedModuleApiVersions::from_raw((1, 0), (1, 0), &[(0, 0)])
     }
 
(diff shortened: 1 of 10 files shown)


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -235,7 +235,7 @@ impl ServerModuleInit for LightningInit {
     }
 
     fn supported_api_versions(&self) -> SupportedModuleApiVersions {
-        SupportedModuleApiVersions::from_raw((u32::MAX, 0), (0, 0), &[(0, 0)])
```
shaurya947, line 238: L233 might need updating as well?

PR #3717, opened 2023-11-24
feat: improved load-test-metrics

douglaz:
- Segregate `gateway_payment_received_success` and `gateway_pay_invoice_success` to detect asymmetries between receiving and sending payments
- Change `gateway_payment_received_started_{method}` to `gateway_{gateway_name}_payment_received_started` to make sure all the metrics use the same convention of referring to the gateway that is receiving/sending the payment
- Sort metrics output based on metric name

### DIFF
--- a/fedimint-load-test-tool/src/main.rs
+++ b/fedimint-load-test-tool/src/main.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


fedimint-load-test-tool/src/main.rs
```diff
@@ -1191,18 +1195,15 @@ async fn handle_metrics_summary(
                 .await?,
         ))
     }
-
-    let mut results = HashMap::new();
+    let mut results = BTreeMap::new();
```
elsirion, line 1198: Thx! The ordering being inconsistent was quite annoying xD

elsirion (approved):

PR #3740, opened 2023-11-27
fedimint-client secret derivation path in fedimint-CLI

shaurya947:
See included markdown document.


shaurya947:
@elsirion looks like module secrets currently have an assertion requiring their parent to have level 0. How best to proceed? Should we alter the derivation scheme, or remove the assertion?
image

fedimint-client/src/lib.rs
```diff
@@ -1494,6 +1494,24 @@ impl ClientBuilder {
         }
     }
 
+    pub async fn get_federation_id(&self) -> anyhow::Result<FederationId> {
+        if let Some(db) = self.db.as_ref() {
+            let db = match db {
```
dpc, line 1499: We should have a method on `DatabaseSource` for it.

dpc (approved):
LGTM, but @elsirion should probably take a look.

docs/secret_derivation.md
```diff
@@ -0,0 +1,24 @@
+# Secret Derivation
+
+`fedimint-client` needs a `DerivableSecret`. Since `fedimint-client` is intended to be used as a library, the consumer of `fedimint-client` is free to arrive at this `DerivableSecret` in a number of ways. Here we describe a recommended approach to encourage interoperability.
+
+Let's say that the consuming application of `fedimint-client` has some `global_root_secret` that it may use for deriving all sorts of secrets deterministically. We then arrive at `fedimint-client`'s `DerivableSecret` using the derivation path:
+
+```
+global_root_secret/<key-type=per-federation=0>/<federation-id>/<wallet-number=0>/<key-type=fedimint-client=0>
```
elsirion, line 8: I'm not sure if my original syntax was optimal, but we can iterate on this.

docs/secret_derivation.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+For convenience, the provided `Bip39RootSecretStrategy` can be used to:
+- generate a BIP-39 mnemonic, and then
+- generate the `global_root_secret` from the mnemonic by calling `Bip39RootSecretStrategy::to_root_secret(mnemonic)`
+
+Then starting with the `global_root_secret`, the following derivation steps are performed:
+
+1. Call `global_root_secret.child_key(0)`. The 0 here indicates the first segment of the derivation path, `<key-type=per-federation=0>`. The returned value is `multi_federation_root_secret` of type `DerivableSecret`.
```
elsirion, line 17: Explaining remaining child ids might be useful. Do we want to reserve some namespace for internal usage and some for external (same for aux per-federation keys)?

fedimint-cli/src/lib.rs
```diff
@@ -781,6 +794,19 @@ fn salt_from_file_path(file_path: &Path) -> PathBuf {
...
+fn get_default_client_secret(
+    global_root_secret: &DerivableSecret,
+    federation_id: &FederationId,
+) -> DerivableSecret {
+    let multi_federation_root_secret = global_root_secret.child_key(ChildId(0));
+    let federation_root_secret = multi_federation_root_secret.federation_key(federation_id);
+    let federation_wallet_root_secret = federation_root_secret.child_key(ChildId(0)); // wallet-number=0
+    federation_wallet_root_secret.child_key(ChildId(0)) // key-type=fedimint-client=0
+}
+
```
elsirion, line 809: Can this become part of `fedimint-client` so other integrators can use it too?

crypto/derive-secret/src/lib.rs
```diff
@@ -79,9 +79,12 @@ impl DerivableSecret {
     ///
     /// This is useful to ensure that the same root secret is not reused
     /// across multiple `fedimint-client` instances for different federations.
+    ///
+    /// We reset the level to 0 here since `fedimint-client` expects its root
+    /// secret to be at that level.
```
elsirion, line 84: I see that it's hard to assert anything in the client otherwise, still seems quite inconsistent :/ 

I think this is something we could change later on if it's a problem, so good enough.

fedimint-client/src/lib.rs
elsirion: We should just get rid of these DB source shenanigans next release. They are only useful if one can't re-open the same DB because the client leaked a DB ref somewhere. I think it was related to recovery where the DB is closed and reopened without restarting the process.

elsirion (approved):

elsirion (approved):

crypto/derive-secret/src/lib.rs
dpc: Why do you find it inconsistent?

fedimint-backports:

```bash
git worktree add -d .worktree/backport-3740-to-releases/v0.2 origin/releases/v0.2
cd .worktree/backport-3740-to-releases/v0.2
git switch --create backport-3740-to-releases/v0.2
git cherry-pick -x 8ec2e02e4a6c4ce068f1615f6aacb9be3222b882 4f1cf2e355f1d7fafb8d4b601e6a504d1445f4c9 64ff17e96c4b0fe016add37cce154fbb2d68c451
```

crypto/derive-secret/src/lib.rs
elsirion: Now the level gets reset by some derivation fn and incremented by others.

dpc: @elsirion The way I think about it is that "level" has a meaning in a context it is being used. The outer application can use (or not) `level` to track and prevent bugs at it's level, and do as many derivations levels at it wants, but the moment it's passed to `Client`, the `Client` changes the context to its internal one, as the secret and its derivations becomes its internal details.

elsirion: Sure, but even in the outer application doing derivations using the federation id will reset the level right now.

dpc: @elsirion Oh, so if it's about this particular function doing the resetting vs the caller, then it's a different story.

If we expect some other callers might want to use it, we should just not change level and move resetting the level to some other explicit call.

Or we can consider this function to always just produce 0-level federation root key, and if the caller wants something else, they can write their own function.

Both are reasonable, IMO. I already wanted `pub fn federation_key` to be a `fedimint-client` function, in which case keeping it private and internal would make resettting more OK thing to do. I doubt anyone should need it anyway.

@shaurya947

justinmoon:
@okjodom @m1sterc001guy @elsirion Does the gateway need to be updated at all for this?

elsirion:
Does the GW implement backup+recovery rn?

justinmoon:
> Does the GW implement backup+recovery rn?

Yes https://github.com/fedimint/fedimint/blob/361a16f8ed03d0f931e801a1216fcc9be8c4d112/gateway/ln-gateway/src/rpc/rpc_server.rs#L42-L43

justinmoon:
Update: these RPC calls are noops apparently. So I guess it isn't implemented. https://discord.com/channels/990354215060795454/1127969523203579944/1179519499574190101

Issue #3664, opened 2023-11-20
Add automated tests for recovery tool


bradleystachurski:
We don't have any tests in CI for the recovery tool. As a starting point, we can add a devimint test using steps similar to https://github.com/fedimint/fedimint/pull/3663#issuecomment-1819875676.

bradleystachurski:
Closed by https://github.com/fedimint/fedimint/pull/3777

PR #3746, opened 2023-11-28
fix: use 33 byte tweaks in recovery tool parser

bradleystachurski:
Followup to https://github.com/fedimint/fedimint/pull/3709 (removing x only public keys makes tweaks 33 bytes - see [BIP-137](https://github.com/bitcoin/bips/blob/master/bip-0137.mediawiki#background-on-ecdsa-signatures))

I would normally wait to include this fix with https://github.com/fedimint/fedimint/issues/3664, however it would be nice to get this in the next release candidate for 0.2.0 sooner than later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1944,6 +1944,7 @@ dependencies = [
  "fedimint-wallet-server",
  "futures",
  "miniscript",
+ "rand",
  "secp256k1 0.24.3",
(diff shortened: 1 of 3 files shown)


elsirion (approved):
Good catch!

fedimint-backports:
- #3764

Issue #3754, opened 2023-11-28
`fedimint-cli withdraw --amount all` assumes user has balance


justinmoon:
If user has balance of zero [this line](https://github.com/fedimint/fedimint/blob/62787696e01173520524e4b2a779b19941de3e8e/fedimint-cli/src/client.rs#L432C3-L432C3) will blow up

Issue #3752, opened 2023-11-28
fedimint-cli can't withdraw "all"


justinmoon:
From `just mprocs`:

```
$  fedimint-cli withdraw --amount all --address bcrt1quluutn45cmz76sqy85ahmrx3lzpfzqe2yq2yja
2023-11-28T04:18:49.629733Z  INFO fedimint_cli::db_locked: Acquiring database lock
2023-11-28T04:18:49.810847Z  INFO fedimint_client::sm::executor: Starting state machine executor task
2023-11-28T04:18:49.829455Z  INFO fedimint_client: Last client reference dropped, shutting down client task group
2023-11-28T04:18:49.829508Z  INFO fedimint_client::sm::executor: Shutting down state machine executor runner due to shutdown signal
{
  "error": "CliError",
  "kind": "general_failure",
  "message": "failure",
  "raw_error": "Federation didn't return peg-out fees"
}
```

Server logs:

```
WARN fedimint_wallet_server: Error returning peg-out fees Not enough SpendableUTXO
```

We really need to add tests before merging things like this going forward.

elsirion:
That's an interesting one! What I think happens is that `GW balance == federation balance`. When the federation is asked for a fee estimate it internally tries to construct a tx that sends this amount, but fails because including the fee it's not possible.

What it likely should do is just assume that all UTXOs are spent completely and return the fee for that, then the client would subtract that fee from the total amount and withdraw the total federation balance, consuming all UTXOs, instead of failing.

Idk how probable this case is irl as some e-cash will always be lost.

justinmoon:
@elsirion https://github.com/fedimint/fedimint/pull/3755 closes this as well. The CLI was multiplying instead of dividing when doing msat -> sat conversion 😬.

PR #3755, opened 2023-11-28
fix 'withdraw all' feature

justinmoon:
- Closes https://github.com/fedimint/fedimint/issues/3754
- Closes https://github.com/fedimint/fedimint/issues/3752, which was caused by multiplying by 1000 instead of dividing when doing msats -> sats conversions
- Trying but currently failing to fix `BitcoinAmountOrAll::All` de/serialization. Currently it can't deserialize `all`. Demands `All`. Can't figure out how to serialize this lowercase 😭.

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -425,11 +425,15 @@ pub async fn handle_command(
                 // the amount we are withdrawing
                 BitcoinAmountOrAll::All => {
                     let balance =
-                        bitcoin::Amount::from_sat(client.get_balance().await.msats * 1000);
+                        bitcoin::Amount::from_sat(client.get_balance().await.msats / 1000);
                     let fees = wallet_module
                         .get_withdraw_fees(address.clone(), balance)
                         .await?;
-                    (balance - fees.amount(), fees)
+                    let amount = balance.checked_sub(fees.amount());
+                    if amount.is_none() {
+                        bail!("Not enough funds to pay fees");
+                    }
+                    (amount.unwrap(), fees)
                 }
                 BitcoinAmountOrAll::Amount(amount) => (
                     amount,
--- a/fedimint-core/src/lib.rs
+++ b/fedimint-core/src/lib.rs
@@ -147,11 +147,12 @@ pub fn sats(amount: u64) -> Amount {
 }
 
 /// Amount of bitcoin to send, or "all" to send all available funds
-#[derive(Debug, Serialize, Deserialize, Clone)]
-#[serde(untagged)]
+#[derive(Debug, Eq, PartialEq, Copy, Hash, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
 pub enum BitcoinAmountOrAll {
-    Amount(#[serde(with = "bitcoin::util::amount::serde::as_sat")] bitcoin::Amount),
     All,
+    #[serde(untagged)]
+    Amount(#[serde(with = "bitcoin::util::amount::serde::as_sat")] bitcoin::Amount),
 }
 
 impl FromStr for BitcoinAmountOrAll {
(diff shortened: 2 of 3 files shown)


fedimint-core/src/lib.rs
```diff
@@ -151,6 +151,7 @@ pub fn sats(amount: u64) -> Amount {
 #[serde(untagged)]
 pub enum BitcoinAmountOrAll {
     Amount(#[serde(with = "bitcoin::util::amount::serde::as_sat")] bitcoin::Amount),
+    #[serde(rename = "all")]
```
justinmoon, line 154: Can't deserialize `all`

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -52,7 +52,7 @@ pub struct DepositAddressPayload {
     pub federation_id: FederationId,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "snake_case")]
```
justinmoon, line 55: Can't deserialize `all` as `BitcoinAmountOrAll`

Maan2003:
serde help: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=04c2a56c36cbb48f2dda7b8387574053

gateway/ln-gateway/src/rpc/mod.rs
dpc: It must be on `BitcoinAmountOrAll`

fedimint-cli/src/client.rs
```diff
@@ -425,11 +425,15 @@ pub async fn handle_command(
...
+                        bitcoin::Amount::from_sat(client.get_balance().await.msats / 1000);
                     let fees = wallet_module
                         .get_withdraw_fees(address.clone(), balance)
                         .await?;
-                    (balance - fees.amount(), fees)
+                    let amount = balance.checked_sub(fees.amount());
+                    if amount.is_none() {
+                        bail!("Not enough funds to pay fees");
+                    }
+                    (amount.unwrap(), fees)
```
benthecarman, line 436: is balance is equal to fees this would result in `Some(0)` and pass. Probably would be better if it was instead

```rust 
let amount = balance.saturating_sub(fees.amount());
if amount == 0 {
	bail!("Not enough funds to pay fees");
}
(amount, fees)
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -688,11 +688,15 @@ impl Gateway {
...
+                let balance = bitcoin::Amount::from_sat(client.get_balance().await.msats / 1000);
                 let fees = wallet_module
                     .get_withdraw_fees(address.clone(), balance)
                     .await?;
-                (balance - fees.amount(), fees)
+                let withdraw_amount = balance.checked_sub(fees.amount());
+                if withdraw_amount.is_none() {
+                    return Err(GatewayError::InsufficientFunds);
+                }
+                (withdraw_amount.unwrap(), fees)
```
benthecarman, line 699: same here

fedimint-cli/src/client.rs
justinmoon: Good point. This case should throw an error when trying to broadcast the transaction, though. Somewhere we should also be checking that we aren't creating dust and IIRC that is currently handled by whatever bitcoin node we're using to broadcast. But don't recall.

benthecarman: yeah, maybe just doing is enough

```
if amount <= 1_000 {
	bail!("Not enough funds to pay fees");
}
```

fedimint-core/src/lib.rs
dpc: `rename_all` on the parent?

fedimint-cli/src/client.rs
elsirion: Let's optimize this in a follow-up, there are a few cases to consider here (fees, dust, …). Currently the tx will fail+refund, which seems ok (just a bit slower feedback).

fedimint-backports:
Git push to origin failed for releases/v0.2 with exitcode 1

fedimint-backports:
- #3803

PR #3757, opened 2023-11-28
feat: add a heartbeat for peer status reporting

### LINKED ISSUE #3744: `status` RPC always returns `last_contribution: 0`
```
$ just mprocs
...
$ fedimint-cli dev api status --peer-id 0
{
  "value": {
    "server": "ConsensusRunning",
    "federation": {
      "session_count": 0,
      "status_by_peer": {
        "3": {
          "last_contribution": 0,
          "flagged": false
        },
        "2": {
          "last_contribution": 0,
          "flagged": false
        },
        "1": {
          "last_contribution": 0,
          "flagged": false
        }
      },
      "peers_online": 3,
      "peers_offline": 0,
      "peers_flagged": 0
    }
  }
}
```

elsirion:
Fixes #3744 (well, it kinda breaks the API but adds the functionality back in)

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -1,6 +1,6 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant, SystemTime};
 
 use aleph_bft::Keychain as KeychainTrait;
 use anyhow::{anyhow, bail};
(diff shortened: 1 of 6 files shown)


fedimint-server/src/consensus/server.rs
```diff
@@ -668,6 +673,18 @@ impl ConsensusServer {
...
+                let now = now();
+                let latency = now.duration_since(send_time).unwrap_or_default();
+
+                debug!(target: LOG_CONSENSUS, "Received heratbeat from {} with latency {} ms sent at {:?}", peer_id, latency.as_millis(), send_time);
+                self.latest_heartbeat_by_peer
+                    .write()
+                    .await
+                    .insert(peer_id, (session_index, now, latency));
+
+                Ok(())
```
elsirion, line 686: Should be handled in `process_consensus_item` instead so we can drop the item and not save it to disk.

PR #3775, opened 2023-11-28
fix: avoid warnings on recovery from mid session crash

joschisan:


elsirion (approved):
I think this approach works well for this particular case but should otherwise be discouraged. What do you think @dpc? The alternative would be disabling the warning also on the other early return, which would be more standard.

elsirion:
Waiting for DPC ;)

dpc:
LGTM. The less mutation, the better, but whatever is simpler works, and it's all just diagnostics anyway.

fedimint-backports:
- #3778

PR #3777, opened 2023-11-28
feat: added recoverytool tests

douglaz:
Added tests for `utxos`and `epochs` method. Reviewed the README

### DIFF
--- a/recoverytool/src/main.rs
+++ b/recoverytool/src/main.rs
@@ -2,7 +2,7 @@ use std::cmp::Ordering;
 use std::collections::BTreeSet;
 use std::fmt::{Display, Formatter};
 use std::hash::Hasher;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use anyhow::anyhow;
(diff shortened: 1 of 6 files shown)


douglaz:
@elsirion Not sure the best way to test the `direct` method. How is one supposed to get the tweaks?

elsirion (approved):
Thx a lot! One thing we might want to add in the future is doing a peg-out before, so we can test that change also gets recovered :)

justinmoon:
Very nice

recoverytool/src/main.rs
```diff
@@ -67,6 +67,10 @@ struct RecoveryTool {
     /// Network to operate on, has to be specified if --cfg isn't present
     #[arg(long, default_value = "bitcoin", requires = "descriptor")]
     network: Network,
+    /// Open the database in read-only mode, useful for debugging, should not be
+    /// used in production
+    #[arg(long)]
+    readonly: bool,
```
bradleystachurski, line 73: Useful 👍

PR #3781, opened 2023-11-29
chore: get rid of `DatabaseSource` workaround

dpc:
It was needed when we were leaking database referencess, which should no longer be the case.

In LN gateway code we should be able to just use an existin `Client` instead of building new one.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1366,12 +1366,7 @@ pub struct ClientBuilder {
     module_inits: ClientModuleInitRegistry,
     primary_module_instance: Option<ModuleInstanceId>,
     config: Option<FederationInfo>,
-    db: Option<DatabaseSource>,
-}
-
-pub enum DatabaseSource {
-    Fresh(Database),
-    Reuse(ClientArc),
+    db: Option<Database>,
 }
 
 impl ClientBuilder {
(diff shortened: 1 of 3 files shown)


elsirion:
CI is failing

dpc:
Will debug when I find some energy.

dpc:
@okjodom @m1sterc001guy so it looks like this fails because something hangs in lnd gateawy, 99% because of the `ArcClient` re-use.

I tried to debug it but I have trouble understanding the code around it (I created #3807 motivated by it), and I don't have a good pre-existing understanding of the gateway architecture.

I don't even undersatnd why the gateway would need multiple client to the same federation, while the code (and my `dbg!` debugging statements) indicate that this can actually be the case? This is most probably fixable in some way, but hard to tell why without undrestanding the requirements.

justinmoon:
dev call: dpc is blocked on this one because some lightning stuff is failing cc @okjodom @m1sterc001guy

m1sterc001guy:
I think I know the crux of the issue here. Today, we have a number of properties on `GatewayClientModule` and these properties are setup when the clients are created. One of these properties is a connection to the lightning node (`Arc<dyn ILnRpcClient>`). So when the connection to the lightning node goes now, we someone need to re-create this connection. Today we naively re-create all of the clients. I remember running into an issue that indicated we couldn't reopen an already existing connection to the database when re-creating the connection, hence the workaround to have `DatabaseSource::Reuse` so that we would just re-use the already opened db.

Ideally, I don't think we should be re-creating the clients at all. We should just create the federation clients when the gateway boots up and never again. So instead of passing the lightning node connection `Arc<dyn ILnRpcClient>` to `GatewayClientModule`, we should just pass a copy of the `Gateway` struct. `Gateway` is already clonable and has convenient methods for checking if the lightning node is online and accessing it. This would allow us to completely delete `DatabaseSource::Reuse` since we'd only ever open the database once, because we only create the clients once.

dpc:
Was done in another PR

PR #3749, opened 2023-11-28
feat: can pass --auth flag to `fedimint-cli dev api`

justinmoon:
I couldn't figure out how to make authenticated RPC requests using `fedimint-cli dev api` and formatting the `params` argument correctly. Tried many variations on `'[{"auth":"pass","params":null}]'` and gave up!

Now you can invoke like:

```
fedimint-cli dev api --peer-id 0 --auth pass audit
```

PR #3785, opened 2023-11-29
fix: only send authentication to one peer

elsirion:
Builds on #3749, I didn't just want to add this restriction there without discussion in case it breaks an intended use case, but in general we probably only want to send authentication to a single guardian server (ours). In the future we might even want to hash the password together with the peer id client-side, so that it cannot accidentally leak to other guardians.


fedimint-cli/src/lib.rs
```diff
@@ -395,6 +395,10 @@ Examples:
         /// Which server to send request to
         #[clap(long = "peer-id")]
         peer_id: Option<u16>,
+        /// Guardian password in case authenticated API endpoints are being
+        /// called. Only use together with --peer-id.
+        #[clap(long = "auth", requires = "peer_id")]
```
dpc, line 400: The format of auth is not clear to me. And example would be great to have, possibly on the whole command (we already have some on `DevCmd`).

dpc: Is it `user:pass`?

elsirion: There is no user, so I thought it would be self-explanatory (Guardian password=what you set on federation setup). But I actually noticed we already have an optional password flag at the top level that we could use here.

dpc: Oh, OK. BTw. Made me realize that in real settings people will probably quickly want user + password and some permission system so they can more granularily work in their organizations etc.

dpc (approved):

PR #3790, opened 2023-11-29
chore: lower the parameters on latency tests

douglaz:


devimint/src/main.rs
```diff
@@ -182,9 +182,9 @@ pub async fn latency_tests(dev_fed: DevFed) -> Result<()> {
               LN RECV: {ln_receives_stats}"
     );
     // FIXME: should be smaller
-    assert!(reissue_stats.median < Duration::from_secs(4));
-    assert!(ln_sends_stats.median < Duration::from_secs(6));
-    assert!(ln_receives_stats.median < Duration::from_secs(6));
+    assert!(reissue_stats.median < Duration::from_secs(5));
```
elsirion, line 185: The PR says lower, but what you are doing is increasing, is that intentional?

douglaz: Yeah, perhaps tuning would be a better word. But I'll close this PR. I think the current values are okay.

PR #3800, opened 2023-11-30
fix: dont require muc.xmpp as a traefik router

### LINKED ISSUE #3799: traefik fails to generate certificate for xmpp domain during tls-mutinynet setup
When i run `curl -sSL https://raw.githubusercontent.com/fedimint/fedimint/master/docker/tls-download-mutinynet.sh | bash` to setup a federation, I noticed I am not able to connect via websocket to the `xmpp.host` URL

I checked `docker-compose logs traefik` and saw that the `muc.xmpp.host` is failing because the setup script never prompts to add a CNAME record for that domain. so that causes `xmpp.host` to also fail

Screenshot 2023-11-30 at 6 27 33 PM

2 ways to fix this I think:
1. don't require muc.xmpp as DNS (see https://prosody.im/doc/chatrooms#dns)
2. make sure `muc.xmpp` is added to SERVICES in the set up script so the user knows to add the CNAME record

otech47:
fixes #3799

in XMPP, the multi-user chat (MUC) component does not strictly require DNS config on the address, only xmpp.example.com is needed since it is a virtual host

more details: https://prosody.im/doc/chatrooms#dns

### DIFF
--- a/docker/full-tls-mutinynet/docker-compose.yaml
+++ b/docker/full-tls-mutinynet/docker-compose.yaml
@@ -403,7 +403,7 @@ services:
     labels:
       - "traefik.enable=true"
       - "traefik.http.services.xmpp.loadbalancer.server.port=5280"
-      - "traefik.http.routers.xmpp.rule=Host(`xmpp.fedimint.my-super-host.com`) || Host(`muc.xmpp.fedimint.my-super-host.com`)"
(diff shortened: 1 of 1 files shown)


douglaz (approved):

fedimint-backports:
- #3987

PR #3801, opened 2023-11-30
chore: lower cpu use when consensus is running locally in dev profile

dpc:


douglaz (approved):
Not sure if it fixes Global Warming, but LGTM

elsirion (approved):

PR #3811, opened 2023-12-01
fix: allow `mainnet` as a valid network

douglaz:


justinmoon:
Perhaps we should eventually build a more holistic way to represent `network` from external software like LND. But this seems like a good fix hotfix.

justinmoon (approved):

dpc (approved):

fedimint-backports:
- #3813

Issue #278, opened 2022-07-16
Do not give full invoice to LN Gateway when sending lightning payments


elsirion:
Instead only give payment hash, destination and amount to LN Gateway.

Stretch goal: implement some custom (optional?) routing based on LDK that uses some probing server @moneyball described in the chat. The goal would be to outperform cln's native pay plugin. (this might become a separate issue once the main part is solved)

tonyimpervious:
What's wrong with giving the full invoice? If no routing hints, then paying through unannounced channels won't be possible and that might be the only way to reach some nodes. 

Also, with route blinding (either fetching through bolt12 invoices or included in bolt11 invoices), you may need the rest of the invoice information.

TonyGiorgio:
Is this just to minimize the data being passed around? If there's routing hints necessary to make the payment (if the end destination is private and/or using an LSP like muun or breez), then you wouldn't be able to pay them.

Also with route blinding around the corner, that may be necessary too. Either route blinding with bolt11 or with bolt12 (which should also be smaller if the goal is minimizing data).

justinmoon:
> Is this just to minimize the data being passed around?

The goal would be for the gateway to learn as little as possible about how federation users are spending their money.

Route blinding sounds interesting. With any of these approaches our client would need to get smarter. Right now the client is extremely simple. For sending it (1) escrows tokens in federation payable to anyone who find preimage and (2) tells gateway about it. [This is what the escrow contract actually looks like](https://github.com/fedimint/minimint/blob/ebb32e7137eece50d7b4ef539d8ccb15284fe4bb/modules/minimint-ln/src/contracts/outgoing.rs#L6-L24). Notice the entire invoice is included. For it to do route blinding the client would need to learn to route which it can't do currently ...

moneyball:
This project is relevant to this: https://github.com/lightningdevkit/rust-lightning/tree/main/lightning-rapid-gossip-sync

It would allow for performant client-side pathfinding. My understanding is that LDK should be modular enough to allow for just the pieces you want to include on the client (invoice management, pathfinding, key management(?)) while keeping other components on the LSP/server if so desired.

Remind me again of the design...are private keys for the LN channels stored on the LSP or the Fedimint user client device? ie is the LN aspect to this custodial or non-custodial?

justinmoon:
> lightning-rapid-gossip-sync

[...]

PR #3816, opened 2023-12-02
Make LN payments more private

elsirion:
When I woke up today I remembered #278 and thought we should fix it before 0.2, turns out a later fix would be annoyingly inefficient because we'd need to put dummy invoices into the outgoing contracts.

**Since the PR contains breaking changes it has to be backported to 0.2 while there isn't a release. Backport PR: #3817.**


gateway/ln-gateway/src/lnd.rs
```diff
@@ -510,6 +513,147 @@ impl ILnRpcClient for GatewayLndClient {
...
+            let payments = client
+                .router()
+                .send_payment_v2(SendPaymentRequest {
+                    amt_msat,
+                    dest: invoice.destination.serialize().to_vec(),
+                    payment_hash: invoice.payment_hash.to_vec(),
+                    route_hints: route_hints_to_lnd(&invoice.route_hints),
+                    final_cltv_delta,
+                    cltv_limit,
+                    allow_self_payment: true,
```
benthecarman, line 584: Idk if you want this on, this is normally only for when trying to do channel rebalances

elsirion: Idk, I just copied that from the other pay function that argument was already there and seemed to work. Later on, when @m1sterc001guy is back, we need to do some refactoring and go over this code again. I haven't touched the gateway in a while but thought fixing this issue in 0.2 is important enough to try it myself anyway.

elsirion:
@justinmoon do you think this is reasonable to backport to 0.2? If not in full I'd like to at least get the first commit in, otherwise we'll need dummy invoices in the contracts going forward.

I could also add functionality for gateway operators to turn off private payments in case there is a bug with it (e.g. some stuff from the invoice we aren't passing on).

gateway/ln-gateway/src/lnd.rs
elsirion: Do you think it could be harmful?

benthecarman: The worst case is that you pay a high fee to route to yourself when it could have been 0

elsirion:
Seems to apply cleanly to 0.2 (#3817), let's see if CI runs through too.

gateway/ln-gateway/src/lnrpc_client.rs
```diff
@@ -62,6 +64,23 @@ pub trait ILnRpcClient: Debug + Send + Sync {
         invoice: PayInvoiceRequest,
     ) -> Result<PayInvoiceResponse, LightningRpcError>;
 
+    async fn pay_private(
```
justinmoon, line 67: Doc string please

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -579,3 +603,15 @@ pub async fn ln_operation(
 
     Ok(operation)
 }
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Decodable, Encodable)]
+pub struct PrunedInvoice {
```
justinmoon, line 608: doc string please

justinmoon (dismissed):

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -563,8 +564,10 @@ impl LightningPayRefund {
 pub struct PayInvoicePayload {
     pub federation_id: FederationId,
     pub contract_id: ContractId,
+    // FIXME: use pruned, privacy friendly version without description etc.
+    /// Invoice containing metadata on how to obtain the preimage
+    pub invoice: lightning_invoice::Bolt11Invoice,
```
elsirion: This breaks decoding, this means this change either needs to get both on `master` and `0.2` or cannot be merged.

gateway/ln-gateway/src/lnd.rs
```diff
@@ -510,6 +513,147 @@ impl ILnRpcClient for GatewayLndClient {
         Ok(PayInvoiceResponse { preimage })
     }
 
+    async fn pay_private(
```
elsirion, line 515: This fn is mostly a copy of `pay`, needs deduplication.

gateway/ln-gateway/src/lnd.rs
```diff
@@ -510,6 +513,147 @@ impl ILnRpcClient for GatewayLndClient {
...
+            }
+        };
+        Ok(PayInvoiceResponse { preimage })
+    }
+
+    /// Returns true if the lightning backend supports payments without full
+    /// invoices
+    fn supports_private_payments(&self) -> bool {
+        true
+    }
```
elsirion, line 653: @justinmoon We could make this configurable in case there is a problem with the private pay implementation.

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -409,12 +428,12 @@ impl GatewayPayInvoice {
 
     async fn buy_preimage_via_direct_swap(
         client: ClientArc,
-        invoice: Bolt11Invoice,
+        paymet_data: PaymentData,
```
douglaz, line 431:
```suggestion
        payment_data: PaymentData,
```

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -245,6 +252,31 @@ pub struct LightningGatewayRegistration {
     pub valid_until: SystemTime,
 }
 
+impl Encodable for LightningGatewayRegistration {
+    fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
+        let json_repr = serde_json::to_string(self).map_err(|e| {
+            Error::new(
+                ErrorKind::Other,
+                format!("Failed to serialize LightningGatewayRegistration: {}", e),
```
douglaz, line 260:
```suggestion
                format!("Failed to serialize LightningGatewayRegistration: {e}"),
```

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -245,6 +252,31 @@ pub struct LightningGatewayRegistration {
...
+}
+
+impl Decodable for LightningGatewayRegistration {
+    fn consensus_decode<R: Read>(
+        r: &mut R,
+        modules: &ModuleDecoderRegistry,
+    ) -> Result<Self, DecodeError> {
+        let json_repr = String::consensus_decode(r, modules)?;
+        serde_json::from_str(&json_repr).map_err(|_| {
+            DecodeError::from_str("Failed to deserialize LightningGatewayRegistration")
```
douglaz, line 275: Include the error?

elsirion: Neither `from_str` or `from_err` would work if I wanted to add context, but I just found out about `new_custom`! That way it's possible.

gateway/ln-gateway/src/lnd.rs
elsirion: done

TonyGiorgio (commented):
Concept ACK, quick scan through. Would be nice for some helpful conversions here.

douglaz (dismissed):
LGTM

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -654,3 +654,27 @@ pub struct PrunedInvoice {
     pub timestamp: SystemTime,
     pub expiry: Duration,
 }
+
+impl TryFrom<Bolt11Invoice> for PrunedInvoice {
+    type Error = anyhow::Error;
+
+    fn try_from(invoice: Bolt11Invoice) -> Result<Self, Self::Error> {
+        Ok(PrunedInvoice {
```
benthecarman, line 662: if you're just wrapping in an `Ok()` can just use do `Impl From<...`

elsirion: Since this becomes a public interface I didn't want to assume the invoice having an amount, which this conversion requires. 

https://github.com/fedimint/fedimint/blob/54006ee0b89244529c1943dbfc4237b418a6c3b3/modules/fedimint-ln-common/src/lib.rs#L664-L666

douglaz (approved):

modules/fedimint-ln-common/src/lib.rs
benthecarman: Ah missed that

PR #3819, opened 2023-12-03
chore: remove unneeded network_to_currency

benthecarman:
LDK supports converting from network to currency and vice versa, can remove this function


elsirion (approved):
Good catch!

Issue #3417, opened 2023-10-18
Backwards-compatible client state machine changes


justinmoon:
We are currently unable to alter the inner state machine enums on release branches because these enums are stored in the database and modifying them would be backwards-incompatible. https://github.com/fedimint/fedimint/pull/3412, for example. This inability to iterate on the state is a pretty big limitation because we often only notice things are missing when sometime tries to integrate the client in an app.

How can we fix this? Can we migrate state machines?

elsirion:
> Are we sure we want db migrations for states within the state machines? It's not clear to me if these will always be backwards compatible (i.e if the execution semantics of a state have changed, do we want to restart the client in this state without having executed the previous states?)
@m1sterc001guy in https://github.com/fedimint/fedimint/pull/3412#issuecomment-1766949428

I think we do want a way to migrate states. If there are differences in semantics that's exactly where we'd want to "translate" old states to new ones (or of course if the encoding changed).

m1sterc001guy:
Here is the order I plan on implementing this:

- [ ]  Client DB Migrations and tests for all modules, but only for non-state machine related data.
- [ ]  Client DB Migrations for state machine data.
- [ ]  Gateway DB migrations for state machine data
- [ ]  Gateway DB migrations for non-state machine related data.

m1sterc001guy:
Completed! Time to start writing migrations :)

PR #3820, opened 2023-12-03
feat: Support receiving LN for other users

benthecarman:
The goal here is to be able to create a lightning invoice on behalf of another user and when it is paid the funds are locked to their key rather than our own. 

Any feedback would be much appreciated!


elsirion:
I discussed a similar scheme with @Kodylow recently for a trusted but federated-custody LNURL server:

* User registers a public key `P` with LNURL server, receives LNURL `U`
* Each time an invoice  is requested for `U`:
  * A public key `P_i=tweak(P, i)` is derived by tweaking `P` with a unique, continuous index `i`
  * The LNURL server posts an `IncomingContractOffer` where the preimage `pre=preimage(P_i)` is derived from `P_i` and the amount is the one from the LNURL request (this is where we need to trust the server to be honest)
  * An invoice is created with payment hash `hash(pre)` and the respective amount and returned to the requester
* When a user comes back online and wants to check for received money:
  * They ask the LNURL server for the last used index `last`
  * They know the index up to which they scanned last time
  * They scan for incoming contracts belonging to keys `P_prev,...,P_last` and start claim state machines for all non-empty, non-expired offers.

This way the LNURL server, while being trusted, never holds any money.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -222,21 +222,34 @@ impl LightningReceiveConfirmedInvoice {
...
+                        }
+                    }
+                    ReceivingKey::External(_) => {
+                        // Claim successful
+                        // fixme is this okay?
+                        LightningReceiveStateMachine {
+                            operation_id: old_state.operation_id,
+                            state: LightningReceiveStates::Success(vec![]),
+                        }
+                    }
```
elsirion, line 252: Should these both cases even share one state machine? I think there's an easier way:

* Only spawn a tx submission SM for submitting the offer
* Don't spawn any lightning module-specific SM as there is nothing for the module to *do* beyond that point, we only want to *observe* payments
* In `subscribe_ln_receive_external` we first await acceptance of the offer submission tx and then wait for the contract being funded. (the two tiered state management is a bit complex, see e.g. how `subscribe_ln_receive` relates to the LN module receive state machine, happy to go into more detail if you want)

elsirion:
> I wonder if just doing an xpub would be simpler

Conceptually not really, you basically do the same tweaking, just with more complexity involved. Maybe it's preferable because it's a bit more standardized, but not by much imo.

benthecarman:
Added the tweaking, just copy-pasted some of the code that you linked since it was in the wallet module

Kodylow:
@elsirion tagging for visibility on support for receive on behalf of other user by pubkey

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -88,7 +88,7 @@ impl State for LightningReceiveStateMachine {
 pub struct LightningReceiveSubmittedOffer {
     pub offer_txid: TransactionId,
     pub invoice: Bolt11Invoice,
-    pub payment_keypair: KeyPair,
+    pub receiving_key: ReceivingKey,
```
elsirion, line 91: This isn't backwards compatible, so either relies on state machine migrations #3417 or needs to be implemented as a new, separate state machine.

elsirion:
See 684551aa36920ff183d0d3976869e733916eee2b for a very hacky version of external invoice generation that needs a lot of refactoring to make it:
* reuse as much logic as possible
* make it backwards compatible

elsirion:
I think #4282 unblocked this PR.

benthecarman:
Rebased the current changes, not really sure where to take it from here tbh

m1sterc001guy:
> Rebased the current changes, not really sure where to take it from here tbh

To make this backwards compatible, you essentially need to map the old `LightningReceiveSubmittedOffer` to the new one. 

1. Make a copy of `LightningReceiveSubmittedOffer` with no changes (can be called v0 or something)
2. Make a new struct that is identical, but contains `ReceivingKey` instead of `KeyPair`
3. Inside `get_database_migrations` of `fedimint-ln-client`, add a new closure into the `BTreeMap` where the closure reads the active/inactive states and checks if any of them are type `LightningReceiveSubmittedOfferV0` (i.e with `KeyPair`). From these "old" states, create the new `LightningReceiveSubmittedOffer` and return it from the closure. There's an example migration in `dummy-client` that does something similar. 
4. Bump the database version in `fedimint-ln-client`.

This will essentially map the old state struct to the new one that you're adding. Hopefully that makes sense. We're doing a deep dive on this next week but feel free to ping me in the meantime.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1562,3 +1626,17 @@ async fn set_payment_result(
             .await;
     }
 }
+
+fn tweak_user_key<Ctx: Verification + Signing>(
```
m1sterc001guy, line 1630: Let's add a docstring here too. Would probably be helpful to explain why we need the key tweak

benthecarman:
I think I did this correctly, not sure how to test. Added as a separate commit for now for easier review

modules/fedimint-ln-client/src/db.rs
```diff
@@ -80,3 +87,69 @@ impl_db_record!(
 );
 
 impl_db_lookup!(key = MetaOverridesKey, query_prefix = MetaOverridesPrefix);
+
+/// DB migration from version 0 to version 1
+pub(crate) async fn migrate_to_v1(
```
m1sterc001guy, line 92: Looks good to me!

m1sterc001guy:
> not sure how to test

Best way to test is probably to follow the dummy client example again.

1. Add function `create_client_states` in `fedimint-ln-tests` and call it from `snapshot_client-db_migrations`. Create the old state that you expect to be migrated.
2. Add some code in `test_client_db_migrations` that verifies that all of the old states are now the new states.
3. Run `just snapshot-client-db-migrations fedimint-ln-tests`. This will re-generate the database backup with your new state written to the db and run the migration test, which will verify that the migration was successful.
4. If all of that passes, check in everything to your branch, including the database files.

benthecarman:
Okay got migration tests passing, actually caught a bug so that feels good

benthecarman:
It changed some commit log files, not sure if those are supposed to be included or not

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -234,21 +234,34 @@ impl LightningReceiveConfirmedInvoice {
...
+                            }),
+                        }
+                    }
+                    ReceivingKey::External(_) => {
+                        // Claim successful
+                        // fixme is this okay?
+                        LightningReceiveStateMachine {
+                            operation_id: old_state.operation_id,
+                            state: LightningReceiveStates::Success(vec![]),
+                        }
```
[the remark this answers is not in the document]
elsirion: I looked at `subscribe_ln_receive` and am fairly certain that it would get stuck (but *should* panic imo since it's an internal programming inconsistency) for external receives since it doesn't expect `Success` before `Funded`.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -38,6 +38,7 @@ const RETRY_DELAY: Duration = Duration::from_secs(1);
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub enum LightningReceiveStates {
+    SubmittedOfferV0(LightningReceiveSubmittedOfferV0),
```
elsirion, line 41: You cannot prepend anything to this enum without breaking decoding for all following variants due to their index changing.

You could add the *new* `SubmitOffer` variant to the end, but I've described a better way in the next comment.

modules/fedimint-ln-client/src/db.rs
```diff
@@ -80,3 +87,75 @@ impl_db_record!(
...
+    module_instance_id: ModuleInstanceId,
+    active_states: Vec<(Vec<u8>, OperationId)>,
+    inactive_states: Vec<(Vec<u8>, OperationId)>,
+    decoders: ModuleDecoderRegistry,
+) -> anyhow::Result<Option<(Vec<DynState>, Vec<DynState>)>> {
+    let mut new_active_states = Vec::with_capacity(active_states.len());
+    for (active_state, _) in active_states {
+        // Try to decode the bytes as a `DynState`
+        let dynstate = DynState::from_bytes(active_state.as_slice(), &decoders)?;
```
elsirion, line 101: To not require a complete, old copy of `LightningClientStateMachines` I'd implement some manual decoding here.

Since we are only interested in the old `SubmittedOffer` state the byte sequence I'd be expecting here is (with concrete values we can match added after `=`):

```
<module_instance_id:varint> <ln_sm_variant:varint=2> <ln_sm_len:varint> <operation_id:32b> <receive_sm_variant:varint=0> <receive_sm_len:varint> <LightningReceiveSubmittedOfferV0>
```

If you parse all SMs that match this pattern using the old `LightningReceiveSubmittedOfferV0` struct while parsing everything else directly as a `DynState` you can implement the migration logic without keeping many redundant structs around while also avoiding polluting production enums with no-longer-used variants.

benthecarman:
Responded to all review besides the fancier version of the db migrations, still figuring out how to properly do that

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1316,6 +1337,49 @@ impl LightningClientModule {
...
+
+    /// Receive over LN with a new invoice for another user
+    pub async fn create_bolt11_invoice_for_user<M: Serialize + Send + Sync>(
+        &self,
+        amount: Amount,
+        description: String,
+        expiry_time: Option<u64>,
+        extra_meta: M,
+        user_key: PublicKey,
+        index: u64,
```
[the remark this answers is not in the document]
benthecarman: made a separate function, one that tweaks and one that just takes a key

modules/fedimint-ln-client/src/receive.rs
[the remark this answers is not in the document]
benthecarman: think i fixed this by just checking if it is an external receiving and yielding the correct state

benthecarman:
Alright think I got the fancy migrations working :rocket:

modules/fedimint-ln-client/src/db.rs
benthecarman: Have this working, however haven't removed the enum variant because otherwise I am unsure on how to do construct the tests. `create_client_states` needs to return a `Vec<LightningClientStateMachines>`, not sure how we'd do that without keeping the `SubmittedOfferV0` variant.

douglaz:
Failing tests?

benthecarman:
> Failing tests?

Needed to re generate, fixed

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -60,6 +61,16 @@ impl State for LightningReceiveStateMachine {
...
         match &self.state {
+            LightningReceiveStates::SubmittedOfferV0(offer) => {
+                warn!("Got old SubmittedOfferV0 state!");
+                // can just translate into the newer version
+                let new_offer = LightningReceiveSubmittedOffer {
+                    offer_txid: offer.offer_txid,
+                    invoice: offer.invoice.clone(),
+                    receiving_key: ReceivingKey::Personal(offer.payment_keypair),
+                };
+                new_offer.transitions(global_context)
```
m1sterc001guy, line 72: Since we're keeping the old state to not break the decoding, maybe we don't even need the db migration? I think with the db migration, this should essentially never happen.

Alternatively we could remove the `SubmittedOfferV0` enum (but keep `LightningReceiveSubmittedOfferV0`), then use the db migration to migrate from `LightningReceiveSubmittedOfferV0` to `SubmittedOffer(LightningReceiveSubmittedOffer)`. That's perhaps a tad cleaner since it doesn't require adding a new enum variant, but I'm good with this too.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -617,15 +652,17 @@ impl LightningClientModule {
...
         operation_id: OperationId,
-    ) -> Result<(), LightningReceiveError> {
+    ) -> Result<bool, LightningReceiveError> {
         let mut stream = self.notifier.subscribe(operation_id).await;
         loop {
             match stream.next().await {
                 Some(LightningClientStateMachines::Receive(state)) => match state.state {
-                    LightningReceiveStates::Funded(_) => return Ok(()),
+                    LightningReceiveStates::Funded(_) => return Ok(false),
+                    LightningReceiveStates::Success(_) => return Ok(true),
```
m1sterc001guy, line 665: I believe `LightningReceiveStates::Success` can happen when the payment is an internal receive as well, but when its an internal receive `Funded` should happen first. I think there might be a small race condition where if you call `await_receive_success` after `Funded` has been emitted for an internal payment, this function will return true even if its an internal payment.

modules/fedimint-ln-client/src/db.rs
```diff
@@ -80,3 +88,94 @@ impl_db_record!(
 );
 
 impl_db_lookup!(key = MetaOverridesKey, query_prefix = MetaOverridesPrefix);
+
+fn get_v1_migrated_state(
```
m1sterc001guy, line 92: Nice!

modules/fedimint-ln-client/src/receive.rs
m1sterc001guy: Like you said above though, having the enum variant around is convenient for the tests. I think if we wanted to delete it we'd need to use the raw db methods to write in the serialized bytes. We should probably modify our migration tests so this is easier, I don't think this is a blocker for this PR

m1sterc001guy:
I don't want to block on this too much more since I think it's looking good and would like to get it in, but we should be able to add a test for this right? Maybe that can be done in a follow up.

Kodylow (commented):
Looks good, just general question maybe to @elsirion are there any other changes that need to be made on the ecash claiming side for the client? would this affect recovery if it needs the tweaks for all the ecash received via external LN ? I don't see where it reissues that received ecash

modules/fedimint-ln-client/src/db.rs
```diff
@@ -80,3 +88,94 @@ impl_db_record!(
...
+    let mut cursor = std::io::Cursor::new(bytes);
+    let key = fedimint_core::core::ModuleInstanceId::consensus_decode(&mut cursor, decoders)?;
+    debug_assert_eq!(key, module_instance_id, "Unexpected module instance ID");
+
+    let ln_sm_variant = u16::consensus_decode(&mut cursor, decoders)?;
+
+    // If the state machine is not a receive state machine, return None
+    if ln_sm_variant != 2 {
+        return Ok(None);
+    }
```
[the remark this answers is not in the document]
benthecarman: its just the ordering of it in the enum, not sure the best way to clarify it without comments

modules/fedimint-ln-client/src/lib.rs
benthecarman: changed to return `outpoints.is_empty()` this should only be true for external receives

modules/fedimint-ln-client/src/receive.rs
benthecarman: Yeah not totally sure what the best move is here, imo the current impl probably makes the most sense for now until there is a way to do tests where we manually write serialized bytes

benthecarman:
> I don't see where it reissues that received ecash

Trying to work on adding that, getting the tests to work for me has been a struggle, still figuring out what I'm doing wrong

benthecarman:
got the beginnings of a test working, need to add functionality to claim the output now

benthecarman:
> need to add functionality to claim the output now

maybe this is better for a follow up, I am pretty lost on how to do this. Might be easier for someone else to do it so I am not blocking anything

modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -452,6 +454,100 @@ async fn makes_internal_payments_within_federation() -> anyhow::Result<()> {
...
+        }
+        _ => panic!("Expected internal payment!"),
+    }
+
+    // TEST internal payment when there is a registered gateway
+    gateway(&fixtures, &fed).await;
+
+    let (op, invoice, _) = client1
+        .get_first_module::<LightningClientModule>()
+        .create_bolt11_invoice(sats(250), "with-gateway-hint".to_string(), None, ())
```
kernelkind, line 515: Should this be `create_bolt11_invoice_for_user`?

benthecarman: woops, fixed. Didn't realize it did the same thing twice, with and without the gateway

benthecarman:
rebased

modules/fedimint-ln-client/src/api.rs
```diff
@@ -120,6 +123,14 @@ where
         .await
     }
 
+    async fn get_unspent_incoming_contracts(&self) -> FederationResult<Vec<ContractId>> {
+        self.request_current_consensus(
```
m1sterc001guy, line 127: From @dpc  https://github.com/fedimint/fedimint/pull/4542#discussion_r1523857020

We need to add an API check here to make sure the server supports this API

m1sterc001guy: Added version check

Kodylow (dismissed):

modules/fedimint-ln-client/src/db.rs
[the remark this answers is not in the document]
elsirion: Our derive macros will need to become smarter in the future, maybe we'll enforce manual indexing such that reordering doesn't accidentally lead to consensus breaking changes :grimacing:

modules/fedimint-ln-client/src/db.rs
```diff
@@ -80,3 +88,94 @@ impl_db_record!(
...
+
+    let _ln_sm_len = u16::consensus_decode(&mut cursor, decoders)?;
+    let operation_id = OperationId::consensus_decode(&mut cursor, decoders)?;
+    let receive_sm_variant = u16::consensus_decode(&mut cursor, decoders)?;
+
+    // If the receive state machine is not a SubmittedOffer variant, return None
+    if receive_sm_variant != 0 {
+        return Ok(None);
+    }
```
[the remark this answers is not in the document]
elsirion: Normally we don't have to think about this too much, we could also make a deep copy of the LN state machine structure definition tree, but that would be a lot of copied code. So instead we just manually implement the relevant parts of the parsing code that would otherwise be derived.

Maybe we need documentation for our encoding scheme, but documenting each version of a every struct is a waste of time imo. You can read the struct/enum definitions and infer the serialized format if encoding/decoding is derived or you can read the manual impls if not.

modules/fedimint-ln-client/src/db.rs
```diff
@@ -83,3 +91,94 @@ impl_db_record!(
...
+                let state_machine = dynstate
+                    .as_any()
+                    .downcast_ref::<LightningClientStateMachines>()
+                    .expect("Unexpected DynState supplied to migration function");
+                state_machine.clone().into_dyn(module_instance_id)
+            }
+        };
+
+        new_active_states.push(state);
+    }
```
elsirion, line 162: This looks really good now! Thx @m1sterc001guy and @benthecarman!

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -38,11 +39,12 @@ const RETRY_DELAY: Duration = Duration::from_secs(1);
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub enum LightningReceiveStates {
-    SubmittedOffer(LightningReceiveSubmittedOffer),
+    SubmittedOfferV0(LightningReceiveSubmittedOfferV0),
```
elsirion, line 42: I think with the new migration code we can now just replace that v0 variant with the updated one?

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1326,6 +1360,111 @@ impl LightningClientModule {
...
+            input,
+            keys: vec![key_pair],
+            state_machines: Arc::new(|_, _| vec![]),
+        };
+
+        let tx =
+            TransactionBuilder::new().with_input(self.client_ctx.make_client_input(client_input));
+        let operation_meta_gen = |_, _| LightningOperationMeta {
+            variant: LightningOperationMetaVariant::Claim,
+            extra_meta: ().into(),
```
elsirion, line 1451: Should be exposed to integrators so they can add custom operation metadata.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -881,6 +882,13 @@ impl ServerModule for Lightning {
                         .await)
                 }
             },
+            api_endpoint! {
+                LIST_UNSPENT_INCOMING_CONTRACTS_ENDPOINT,
+                ApiVersion::new(0, 2),
+                async |module: &Lightning, context, _v: ()| -> Vec<ContractId> {
+                    Ok(module.list_unspent_incoming_contracts(&mut context.dbtx().into_nc()).await)
+                }
+            },
```
elsirion, line 891: Listing all unclaimed contracts isn't scalable at all. You can just use the account endpoint to query if a contract with a given contract id (which is the payment hash) exists. You just do that as long as you find contracts plus a gap limit.

There is a problem here that if more than gap limit many unpaid invoices were generated then you won't find later paid ones. This can be solved by introducing an API endpoint that lets you query incoming offers. Every time an invoice is generated an offer is created, so that doesn't depend on payment status.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1063,6 +1071,25 @@ impl Lightning {
...
+                if let LightningAuditItemKey::Incoming(incoming) = contract {
+                    Some(incoming)
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+            .await
+    }
+
```
elsirion, line 1092: Using the audit API for this makes the audit API harder to change later on, I'd like to avoid that.

elsirion:
> Looks good, just general question maybe to @elsirion are there any other changes that need to be made on the ecash claiming side for the client? would this affect recovery if it needs the tweaks for all the ecash received via external LN ? I don't see where it reissues that received ecash

@Kodylow modules don't have to know of each other, so anything we do in the LN module can't affect the e-cash module. All the LN module sees is an API that lets it dump money into a primary module that can keep it somehow. This primary module could be the e-cash module or some other hypothetical, maybe account-based, module.

modules/fedimint-ln-client/src/receive.rs
m1sterc001guy: Ideally, yes, but unfortunately it makes testing harder. I made this as a follow up, which would allow us to delete states https://github.com/fedimint/fedimint/issues/4536

modules/fedimint-ln-server/src/lib.rs
m1sterc001guy: Hmm is a gap limit necessary? I thought the LNURL server would be able to keep track of the tweak indices that have been paid and be able to provide them to the client. Worst case scenario (if LNURL server went away) a wallet implementation could keep track of a gap limit and "last scanned index" themselves, but I thought it would be best to leave this up to higher level consumers of the client. That's why I added `indices: Vec<u64>`, to try to make the scanning a bit more efficient by only scanning the indices the LNURL server is tracking.

I agree that listing all unclaimed contracts isn't scalable, that was the main motivation with using the `LightningAuditItemKeyPrefix`, since I believe this should only query _funded_ unclaimed contracts. Unless there's a lot of concurrent lightning receives, I don't think this list should be too big. And these records are also cleaned up (unlike `ContractKey`), so the query should be performant. If you think its better to fetch each contract individually, we can change it to that.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1326,6 +1360,111 @@ impl LightningClientModule {
...
 
+    /// Checks the negotiated module API version vs the version that introduced
+    /// claiming incoming unspent contracts. Throws an error if the server
+    /// does not support claiming unspent contracts.
+    fn supports_claim_unspent_incoming_contract(&self) -> anyhow::Result<()> {
+        const VERSION_THAT_INTRODUCED_CLAIM_UNSPENT_INCOMING_CONTRACT: ApiVersion =
+            ApiVersion { major: 0, minor: 2 };
+
+        if self.module_api_version < VERSION_THAT_INTRODUCED_CLAIM_UNSPENT_INCOMING_CONTRACT {
+            return Err(anyhow::anyhow!(
```
douglaz, line 1371:
```suggestion
            anyhow::bail!(
```

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1326,6 +1360,111 @@ impl LightningClientModule {
...
+        let contract_hashes = self.module_api.get_unspent_incoming_contracts().await?;
+        self.scan_receive_internal(key_pair, contract_hashes).await
+    }
+
+    /// Scan unspent incoming contracts for a payment hash that matches a public
+    /// key and claim the incoming contract
+    async fn scan_receive_internal(
+        &self,
+        key_pair: KeyPair,
+        contract_hashes: Vec<ContractId>,
```
douglaz, line 1415:
```suggestion
        contract_hashes: &HashSet<ContractId>,
```

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1326,6 +1360,111 @@ impl LightningClientModule {
...
+        key_pair: KeyPair,
+        indices: Vec<u64>,
+    ) -> anyhow::Result<Amount> {
+        self.supports_claim_unspent_incoming_contract()?;
+        let contract_hashes = self.module_api.get_unspent_incoming_contracts().await?;
+
+        for i in indices {
+            let key_pair_tweaked = tweak_user_secret_key(&self.secp, key_pair, i);
+            let amount = self
+                .scan_receive_internal(key_pair_tweaked, contract_hashes.clone())
```
douglaz, line 1392:
```suggestion
                .scan_receive_internal(key_pair_tweaked, &contract_hashes)
```

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1326,6 +1360,111 @@ impl LightningClientModule {
...
+
+    /// Scan unspent incoming contracts for a payment hash that matches a
+    /// tweaked keys in the `indices` vector
+    pub async fn scan_receive_for_user_tweaked(
+        &self,
+        key_pair: KeyPair,
+        indices: Vec<u64>,
+    ) -> anyhow::Result<Amount> {
+        self.supports_claim_unspent_incoming_contract()?;
+        let contract_hashes = self.module_api.get_unspent_incoming_contracts().await?;
```
douglaz, line 1387:
```suggestion
        let contract_hashes: HashSet<_> = self.module_api.get_unspent_incoming_contracts().await?.into_iter().collect();
```

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1326,6 +1360,111 @@ impl LightningClientModule {
...
+    /// Claim the funded, unspent incoming contract by submitting a transaction
+    /// to the federation and awaiting the primary module's outputs
+    async fn claim_funded_incoming_contract(
+        &self,
+        key_pair: KeyPair,
+        contract_id: ContractId,
+    ) -> anyhow::Result<Amount> {
+        let incoming_contract_account = get_incoming_contract(self.module_api.clone(), contract_id)
+            .await?
+            .ok_or(anyhow::anyhow!("No contract account found"))?;
```
douglaz, line 1438:
```suggestion
            .with_context(|| format!("No contract account found for {contract_id:?}"))?;
```

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1423,7 +1633,13 @@ impl LightningClientModule {
...
                 match self_ref.await_receive_success(operation_id).await {
-                    Ok(()) => {
+                    Ok(is_external) => {
+                        // If the payment was external, we can consider it claimed
+                        if is_external {
+                            yield LnReceiveState::Claimed;
+                            return;
+                        }
+
                         yield LnReceiveState::Funded;
```
douglaz, line 1643:
```suggestion
                    Ok(is_external) if is_external => {
                        // If the payment was external, we can consider it claimed
                        yield LnReceiveState::Claimed;
                        return;
                    }
                    Ok(_is_external) => {
                        yield LnReceiveState::Funded;
```

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -60,6 +62,16 @@ impl State for LightningReceiveStateMachine {
         global_context: &DynGlobalClientContext,
     ) -> Vec<StateTransition<Self>> {
         match &self.state {
+            LightningReceiveStates::SubmittedOfferV0(offer) => {
+                warn!("Got old SubmittedOfferV0 state!");
```
douglaz, line 66:
```suggestion
                warn!("Got old SubmittedOfferV0 state for {offer:?}");
```

modules/fedimint-ln-server/src/lib.rs
elsirion: > Hmm is a gap limit necessary?

You are right, not necessary if you assume the LNURL server is reachable. A hybrid approach is probably best.

modules/fedimint-ln-client/src/receive.rs
elsirion: Ok, then we should at least open an issue/leave a FIXME that this has to be cleaned up as soon as we can test it.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1357,6 +1358,79 @@ impl LightningClientModule {
         }))
     }
 
+    /// Scan unspent incoming contracts for a payment hash that matches a
+    /// tweaked keys in the `indices` vector
+    pub async fn scan_receive_for_user_tweaked(
+        &self,
+        key_pair: KeyPair,
+        indices: Vec<u64>,
+    ) -> anyhow::Result<Amount> {
```
elsirion, line 1367: Currently, this function can only be called only once the contract is funded, relying on the LNURL server to tell us about funded contracts. Later on we should migrate to a model where we scan for incoming offers and then start our normal state machines looking for contracts belonging to these offers.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1357,6 +1358,79 @@ impl LightningClientModule {
...
+        for i in indices {
+            let key_pair_tweaked = tweak_user_secret_key(&self.secp, key_pair, i);
+            if let Ok(amount) = self.scan_receive_for_user(key_pair_tweaked).await {
+                return Ok(amount);
```
elsirion, line 1371: Is returning early really what we want to do here? Won't we be missing contracts or is this function meant to be called multiple times with the same list of indices?

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1357,6 +1358,79 @@ impl LightningClientModule {
...
+    /// Scan unspent incoming contracts for a payment hash that matches a public
+    /// key and claim the incoming contract
+    pub async fn scan_receive_for_user(&self, key_pair: KeyPair) -> anyhow::Result<Amount> {
+        let preimage_key: [u8; 33] = key_pair.public_key().serialize();
+        let preimage = sha256::Hash::hash(&preimage_key);
+        let contract_id = ContractId::from_hash(sha256::Hash::hash(&preimage));
+        self.claim_funded_incoming_contract(
+            key_pair,
+            contract_id,
+            "claiming unspent funded incoming contract",
```
elsirion, line 1387: `extra_meta` should be exposed to integrators to add custom metadata, not filled with redundant data.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1357,6 +1358,79 @@ impl LightningClientModule {
...
+            .finalize_and_submit_transaction(
+                operation_id,
+                LightningCommonInit::KIND.as_str(),
+                operation_meta_gen,
+                tx,
+            )
+            .await?;
+        self.client_ctx
+            .await_primary_module_outputs(operation_id, out_points)
+            .await
```
elsirion, line 1431: Waiting here makes the function blocking, why do that? This fn should return an operation id and there should be a `subscribe_*_updates` fn instead that can be used to follow the progress of the claim.

I see the update stream was changed to accomodate such claims, so should already work.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1704,3 +1778,22 @@ fn tweak_user_key<Ctx: Verification + Signing>(
...
+    let mut hasher = HmacEngine::<sha256::Hash>::new(&public_key.serialize()[..]);
+    hasher.input(&index.to_be_bytes());
+    let tweak = Hmac::from_engine(hasher).into_inner();
+
+    let secret_key = key_pair.secret_key();
+    let sk_tweaked = secret_key
+        .add_tweak(&Scalar::from_be_bytes(tweak).expect("Cant fail"))
+        .expect("Cant fail");
+    KeyPair::from_secret_key(secp, &sk_tweaked)
+}
```
elsirion, line 1799: Might be good to expose the pub key equivalent so that LNURL server builders have an easier time

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Ah yeah, makes more sense to keep a total and return at the end

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: I figured this function would already be called in the background, since the wallet will need to listen (or poll) the LNURL server when it receives. Returning the `OperationId` works for me too, I will change it to that

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: I'll change this to return `Vec<OperationId>`

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Even if we scan for offers instead of contracts, aren't we still relying on the LNURL server to tell us the tweak that it used for the public key? If it doesn't reveal the tweak and it is hard to guess by scanning (i.e not within some gap limit), I don't think the client will be able to create the `SecretKey` needed to claim the contract

m1sterc001guy: I misinterpreted this comment, discussed offline. Added `subscribe_ln_claim` in the short term to address this

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Changed to non-blocking

elsirion (approved):
Good enough to get it in and iterate on, I didn't see anything that would cause us too much long term trouble. 

Some TODOs for follow-ups:
* Remove v0 receive state
* Define derivation path of external user base key that the LNURL server uses to derive the actual keys
* Let the LN module do the querying for new offers/contracts, currently it's the job of integrators

Thx @m1sterc001guy and @benthecarman for pushing it over the finishing line!

joschisan (approved):

douglaz (approved):

PR #3821, opened 2023-12-03
feat(wallet): special-case descriptor for single-guardian instances

elsirion:
Using p2wpkh is more efficient than a 1-of-1 multisig


benthecarman:
Could also do taproot

Are no changes needed for the spending? Does miniscript handle it all here?

justinmoon (approved):
LGTM

justinmoon:
Does recovery tool need to be updated to support this?

bradleystachurski:
> Does recovery tool need to be updated to support this?

Capturing what we discussed in the dev call: We don't need to update the recovery tool to support, however it doesn't hurt to add a test case.

https://github.com/fedimint/fedimint/issues/3916

PR #3823, opened 2023-12-04
Get rid of needless nixpkgs

dpc:
Every evaluation of nixpkgs requires downloading extra data and even more so: lots of time and memory when evaluating nixpkgs expressions (because it has tens of thousands of packages).


justinmoon:
dev call: something about core lightning is broken

dpc:
clightning not starting properly, repro with: `env FM_TEST_ONLY=bitcoind bash -x ./scripts/tests/backend-test.sh`

Maan2003 (approved):
🚀

flake.nix
```diff
@@ -1,42 +1,56 @@
...
+                    url = "https://github.com/ElementsProject/lightning/releases/download/v${version}/clightning-v${version}.zip";
+                    sha256 = "sha256-Tj5ybVaxpk5wmOw85LkeU4pgM9NYl6SnmDG2gyXrTHw=";
+                  };
+                  makeFlags = [ "VERSION=v${version}" ];
                   configureFlags = [ "--enable-developer" "--disable-valgrind" ];
                 } // pkgs.lib.optionalAttrs (!pkgs.stdenv.isDarwin) {
                   NIX_CFLAGS_COMPILE = "-Wno-stringop-truncation -w";
                 });
 
                 # Note: we are using cargo-nextest from pkgs-unstable because it has some fixes we need
```
douglaz, line 51: Is this comment still correct?

dpc: No. :)

.config/flakebox/.gitignore
```diff
@@ -0,0 +1 @@
+tmp/
```
elsirion, line 1: Why do we need that? Will this directory be created now or is it something local that could live in `.git/info/exclude`?

flake.nix
```diff
@@ -1,42 +1,71 @@
...
+                    };
+                  });
+
+                rocksdb_6_23 = prev.rocksdb_6_23.overrideAttrs (oldAttrs:
+                  pkgs.lib.optionalAttrs pkgs.stdenv.isDarwin {
+                    # C++ and its damn super-fragie compilation
+                    env = oldAttrs.env // {
+                      NIX_CFLAGS_COMPILE = oldAttrs.env.NIX_CFLAGS_COMPILE + " -Wno-error=unused-but-set-variable";
+                    };
+                  });
```
elsirion, line 45: ~Why do we need two versions of rocksdb :open_mouth:~

I see, one for Fedimint and one for elecctrs.

elsirion (approved):

.config/flakebox/.gitignore
elsirion: Oh, that's in flakebox, nevermind

dpc:
Still testing build on aarch64-darwin machine. :(

elsirion:
I'll also mark #3824 as draft then.

dpc:
Passing on aarch64-darwin now :crossed_fingers:

douglaz (dismissed):

flake.nix
dpc: Nope, that would be too easy. One for electrs, one for esplora-electrs, and normal one for us. :D

dpc:
Took > 12h of debugging and learning, but I think I've reached a combination that works everywhere. The biggest issue was that that newest stdenv provided by NixOS 23.11 (clang 16) fails to link rocksdb with Rust. I had to do bunch of refactoring in Flakebox to allow injecting custom stdenv. :shrug:

dpc:
@elsirion :play_or_pause_button: :pray:

flake.nix
```diff
@@ -1,7 +1,8 @@
 {
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.05";
-    nixpkgs-unstable.url = "github:NixOS/nixpkgs/nixos-unstable";
+    # nixpkgs - We use nixpkgs as input of `flakebox`, as it locks things like toolchains,
+    #           in versions that are actually tested in flakebox's CI to cross-compile things
+    #           well. This also saves us download and Nix evaluation time.
```
elsirion: This comment is a bit confusing, by reading the code I get what it means though. We are using the `nixpkgs` used by `flakebox, so we don't need it listed here (kind of the other way around than specifying that falkebox follows our `nixpkgs`).

nix/esplora-electrs.nix
```diff
@@ -0,0 +1,55 @@
...
+, rustPlatform
+}:
+
+let
+  rocksdb = rocksdb_6_23;
+in
+rustPlatform.buildRustPackage {
+  pname = "esplora";
+  # last tagged version is far behind master
+  version = "20230218";
```
elsirion, line 18: I wonder if they'd accept a flake in their repo as a PR.

flake.nix
```diff
@@ -1,8 +1,13 @@
...
-    #           in versions that are actually tested in flakebox's CI to cross-compile things
-    #           well. This also saves us download and Nix evaluation time.
+    nixpkgs = {
+      url = "github:nixpkgs/nixos-23.11";
+      # We use nixpkgs as input of `flakebox`, as it locks things like
+      # toolchains, in versions that are actually tested in flakebox's CI to
+      # cross-compile things well. This also saves us download and Nix
+      # evaluation time.
+      follows = "flakebox/nixpkgs";
+    };
```
elsirion: What's the difference between using the input from flakebox directly and specifying our own input but specifying that it follows flakebox?

flake.nix
```diff
@@ -32,7 +32,17 @@
...
+                # https://bitcoin.stackexchange.com/questions/84765/how-can-channel-policy-be-missing
+                # note that config-time `--enable-developer` turns into run-time `--developer` at some
+                # point
+                clightning = prev.clightning.overrideAttrs (oldAttrs: rec {
+                  version = "23.05.2";
+                  src = prev.fetchurl {
+                    url = "https://github.com/ElementsProject/lightning/releases/download/v${version}/clightning-v${version}.zip";
+                    sha256 = "sha256-Tj5ybVaxpk5wmOw85LkeU4pgM9NYl6SnmDG2gyXrTHw=";
+                  };
+                  makeFlags = [ "VERSION=v${version}" ];
```
elsirion, line 66: Do you know if there's an issue in the lightning repo yet? I cannot find anything but also lack a bit of context. Is this about slower channel sync, i.e. that it takes longer for all the channel policies to be announced?

elsirion (dismissed):
I don't claim to understand everything you did, but looks good and should reduce the amount of stuff we have to download, which is a big win!

flake.nix
dpc: The github actions checking fresheness like it better. It might have some flake.lock effects too, but I'm unclear on it.

flake.nix
dpc: I asked ton Discord #support channel and I don't think I got any answer. Will need to follow-up.

Issue #3822, opened 2023-12-04
Potential fedimint-cli improvements/problems


dpc:
While working on https://twitter.com/dpc_pw/status/1731468650111221878 it came to my attention

* [ ] `fedimint-cli` should create a data dir if doesn't exit
* [ ] #3798 changed `FM_DATA_DIR` to `FM_CLIENT_DIR` as an env for `--data-dir`. This wasn't initially my intention (I kind of wanted just `devimint` to use `FM_CLIENT_DIR`, somehow). But maybe it's OK how it is now? In this case maybe we should rename `--data-dir` to `--client-dir`? . Ping @mayrf 

* [ ] We should get rid of `WORKDIR` term in `--help` output. It should just match the argument name. So either `CLIENT_DIR` or `DATA_DIR`.

![image](https://github.com/fedimint/fedimint/assets/9209/23242d6d-466f-486d-8e8a-603c64210d11)

* [ ] `fedimint-cli` is too chatty by default. We should probably have a `-v` flag that enables the info log levels we're seeing ATM (logs of highest-level client steps)
* [ ] `fedimint-cli info --total-balance` to just get the total balance without using `jq`
* [ ] Could `await-invoice` work with the invoice itself instead of `operation-id`? How hard would it be for the client to just look it up? If that was the case then `ln-invoice` could output just the invoice instead of a json.
* [ ] Generally I'm not a fan of dealing with json. Would be nice if at least some commands could have `--json` vs `--text` argument to change how things are being shown.
* [ ] `fedimint-cli spend` takes a positional argument (which I'm almost never a fan of), while `fedimint-cli ln-invoice` takes `--amount`. Would turn positional into named argument.

justinmoon:
dev call: `await-invoice` might be able to use `invoice` because the operation id is derived from the invoice.

douglaz:
Related PR: https://github.com/fedimint/fedimint/pull/3848

15IITian:
@dpc  I think , `fedimint-cli info` has no  `--total-balance`  argument in it.

![image](https://github.com/fedimint/fedimint/assets/133089376/af36cbe5-967c-43c0-8f08-38f59036395a)

elsirion:
TODO: split into sub-issues and close.

PR #3830, opened 2023-12-04
fix: make client less verbose

elsirion:
Part of https://github.com/fedimint/fedimint/issues/3822

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -227,7 +227,7 @@ impl fmt::Display for CliError {
     }
 }
 
-#[derive(Parser)]
+#[derive(Parser, Clone)]
 #[command(version)]
 struct Opts {
     /// The working directory of the client containing the config and db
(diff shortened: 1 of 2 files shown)


fedimint-cli/src/lib.rs
```diff
@@ -242,6 +242,11 @@ struct Opts {
     #[arg(long, env = "FM_PASSWORD")]
     password: Option<String>,
 
+    /// Activate more verbose logging, for full control use the RUST_LOG env
+    /// variable
+    #[arg(short = 'v', long)]
```
dpc, line 247: **note:** It's possible to make clap count number of times this was passed and go even more verbose. Not sure if we need it yet.

dpc (approved):

fedimint-cli/src/lib.rs
elsirion: But then we'd need to think about activating the always-muted modules again at some point. It seems much easier to just refer to `RUST_LOG`.

Which logs to put into which verbosity level is a discussion I didn't want to get into.

Issue #3806, opened 2023-11-30
Gateway Failed to establish htlc stream (v0.2rc2)


TonyGiorgio:
Running through the mutinynet-tls script that sets up all of the docker containers, I'm running into an issue on first launch. This happened to me consistently 3 times in a row on v0.2

```
docker-gatewayd-1               | 2023-11-30T17:39:31.472216Z  INFO ln_gateway::lnd: Gateway configured to connect to LND LnRpcClient at 
docker-gatewayd-1               |  address: https://lnd:10009/,
docker-gatewayd-1               |  tls cert path: /lnd_data/tls.cert,
docker-gatewayd-1               |  macaroon path: /lnd_data/data/chain/bitcoin/signet/admin.macaroon 
docker-gatewayd-1               | 2023-11-30T17:39:31.477478Z  INFO ln_gateway::lnd: Gateway configured to connect to LND LnRpcClient at 
docker-gatewayd-1               |  address: https://lnd:10009/,
docker-gatewayd-1               |  tls cert path: /lnd_data/tls.cert,
docker-gatewayd-1               |  macaroon path: /lnd_data/data/chain/bitcoin/signet/admin.macaroon 
docker-gatewayd-1               | 2023-11-30T17:39:31.477497Z  INFO ln_gateway: Established HTLC stream
docker-lnd-1                    | 2023-11-30 17:39:31.478 [ERR] RPCS: [/routerrpc.Router/HtlcInterceptor]: interceptor already exists
docker-gatewayd-1               | 2023-11-30T17:39:31.479172Z ERROR ln_gateway::lnd: Failed to establish htlc stream
docker-gatewayd-1               | 2023-11-30T17:39:31.525121Z  INFO ln_gateway: Successfully loaded Gateway clients.
docker-gatewayd-1               | 2023-11-30T17:39:31.525139Z  INFO ln_gateway: Got None while handling HTLC stream, exiting from loop...
docker-gatewayd-1               | 2023-11-30T17:39:31.525144Z  WARN ln_gateway: HTLC Stream Lightning connection broken. Gateway is disconnected
docker-gatewayd-1               | 2023-11-30T17:39:31.525151Z  INFO task: Waiting for task to finish task=LND HTLC Subscription
docker-gatewayd-1               | 2023-11-30T17:39:31.525158Z  INFO task: Task finished task=LND HTLC Subscription
[...]

PR #3833, opened 2023-12-04
fix(gateway): disconnect before changing the network

douglaz:
Closes https://github.com/fedimint/fedimint/issues/3806


okjodom (approved):

douglaz:
It would be good a test for that scenario, but let's just solve the problem for now.

fedimint-backports:
- #3835

PR #3836, opened 2023-12-05
chore: bundle per-binary debs and rpms

dpc:
Seems like bundling outputs with multiple
binaries does not work, and the bundle assumes
the name of the binary is the same as the name
of the output.

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -345,9 +345,6 @@ jobs:
     strategy:
       matrix:
         build:
-          - flake-output: client-pkgs
-            bins: ""
-            deb: fedimint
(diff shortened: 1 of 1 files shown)


elsirion (dismissed):

dpc:
@elsirion Tried on an debian VM and seems to me like it's working. The only problem is that `--force-overwrite` is needed as two debs come with overlapping files.

elsirion:
I didn't notice the `--force-overwrite` thing and I *think* I tried both `fedimintd` and `fedimint-cli` :confused:  so one has to add that argument to the binary when running? Can that be fixed in the future?

elsirion (approved):

Issue #3838, opened 2023-12-05
Remove `lightning-invoice` `std` feature from client


elsirion:
Remove `lightning-invoice` `std` feature from client and test that it compiles independently (without the workspace pulling the feature in again). This might be more involved, so I split it out into a separate issue.

@TonyGiorgio anything else we can do that would help you guys with WASM compatibility?

TonyGiorgio:
You guys have really good wasm32 practice already, so not much more. I think the weirdness mostly comes when using a library like LDK that pulls in `rust-bitcoin` `std`, or trying to go `no-std` and uses `rust-bitcoin` `no-std` and then other things break. 

I hear this recent merge will help LDK work around it: https://github.com/rust-bitcoin/rust-bitcoin/pull/2066

It might be required unless you also want to move to `rust-bitcoin` `no-std` as well. 

We've tried to get a couple things into LDK like swapping the `now()` calls with wasm32 capable calls or removing them from `wasm32` but they are all rejected: https://github.com/lightningdevkit/rust-lightning/pull/2560 which is why we run with this patch: https://github.com/MutinyWallet/rust-lightning/commit/a9925126ea3b27b91e0e804e2e2dd8ae29815218

though as I noted in https://github.com/fedimint/fedimint/pull/3839#issuecomment-1841085709 we might just have to loosen up a couple of the `not(wasm32)` flags we added so we can attempt to use some of the `lightning-invoice` calls you guys are using. 

In the end, LDK's `std` flag is not safe to run in wasm32 and I think their preferred solution is to only allow that use case in `no-std` with the ability to eventually allow `rust-bitcoin` `std` as an option. 

Maybe @thebluematt can chime in here. Another option is to just rip out `lightning-invoice` if you just need invoice parsing since it's heavily reliant upon the `rust-lightning` and their practices.

elsirion:
We are successfully using WASM builds, nothing actionable here afaik.

PR #3839, opened 2023-12-05
fix: don't depend on std features from lightning-invoice

### LINKED ISSUE #3837: Client should not rely on `std` features of `lightning-invoice`
This is a problem for Munitnywallet: https://discord.com/channels/990354215060795454/990354215878688860/1181358717606834246

2023-12-05-110832_5120x1441_scrot

elsirion:
Fixes #3837. Would be awesome if you could test this today @TonyGiorgio so we can get it into 0.2.0.

### DIFF
--- a/gateway/ln-gateway/src/state_machine/pay.rs
+++ b/gateway/ln-gateway/src/state_machine/pay.rs
@@ -1,6 +1,5 @@
 use std::fmt::Display;
 use std::sync::Arc;
-use std::time::Duration;
 
 use bitcoin_hashes::sha256;
 use fedimint_client::sm::{ClientSMDatabaseTransaction, State, StateTransition};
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
No, not working. Still uses `is_expired()` which is also being a `std` flag. We'll probably just edit our fork of LDK to allow some of the `std` flags in wasm32 that don't explicitly use `SystemTime::now()`.

elsirion:
Ahhh, damn it. I removed all the *added* calls to non-std functions, `is_expired` existed before, but I didn't consider that I moved it from the gateway to the client crate. So you should be able to just add that function as `unimplemented!` to your fork, let's figure out a way to test that we don't use std there longer term (#3838).

TonyGiorgio:
Thanks for your help. Added a comment there. I'll try a few things out in our patch and get it working again. Thank you!

TonyGiorgio:
Actually, you may want to pull out `is_expired()`, it calls `SystemTime::now()`. I can add `unimplemented` or `false` on our end but you'll run into this as well running the fedimint client in any wasm32 environment

```
	#[cfg(all(feature = "std", not(target_arch = "wasm32")))]
		Self::is_expired_from_epoch(&self.timestamp(), self.expiry_time())
	}

	/// Returns whether the expiry time from the given epoch has passed.
	#[cfg(all(feature = "std", not(target_arch = "wasm32")))]
	pub(crate) fn is_expired_from_epoch(epoch: &SystemTime, expiry_time: Duration) -> bool {
		match epoch.elapsed() {
			Ok(elapsed) => elapsed > expiry_time,
			Err(_) => false,
		}
	}
```

```
    #[stable(feature = "time2", since = "1.8.0")]
    pub fn elapsed(&self) -> Result<Duration, SystemTimeError> {
        SystemTime::now().duration_since(*self)
    }
```

elsirion:
Currently this `is_expired` fn is only called on the gateway and not the user client. But I'll fix, just not sure if worth backporting.

PR #3847, opened 2023-12-05
chore: change info endpoint http method to get

### LINKED ISSUE #3809: Gateway API's `info` endpoint should be GET request
Idiomatically this should be a GET request.

Currently it's really hard to exercise it because you need to pass `null` as HTTP payload: `curl -v -X POST -H "Content-Type: application/json" --data null http://localhost:8175/info`

mayrf:
Fixes #3809


douglaz (approved):
LGTM

PR #3848, opened 2023-12-06
feat(fedimint-cli): added support to pay a lnurl

douglaz:
Added support to paying lnurls. Very common nowadays.

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -23,11 +23,12 @@ use fedimint_ln_common::contracts::ContractId;
 use fedimint_mint_client::{MintClientModule, OOBNotes};
 use fedimint_wallet_client::{WalletClientModule, WithdrawState};
 use futures::StreamExt;
+use lightning_invoice::Bolt11Invoice;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use time::format_description::well_known::iso8601;
 use time::OffsetDateTime;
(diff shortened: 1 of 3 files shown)


fedimint-cli/src/client.rs
```diff
@@ -489,6 +497,51 @@ pub async fn handle_command(
...
 
+async fn get_invoice(info: &str, amount: Option<Amount>) -> anyhow::Result<Bolt11Invoice> {
+    let info = info.trim();
+    match lightning_invoice::Bolt11Invoice::from_str(info) {
+        Ok(invoice) => {
+            debug!("Parsed parameter as bolt11 invoice: {invoice}");
+            match (invoice.amount_milli_satoshis(), amount) {
+                (Some(_), Some(_)) => {
+                    bail!("Amount specified in both invoice and command line")
+                }
```
benthecarman, line 508: if these are equal it should be fine

elsirion: If the invoice doesn't have an amount the recipient won't enforce it, so we'd trust the gateway to actually send the whole amount. One day the client might do the routing, then this wouldn't be a problem anymore.

elsirion: Ah, sorry, totally missed that, good point!

elsirion (approved):

fedimint-backports:
- #3854

Issue #3852, opened 2023-12-06
[RC5] "The gateway is disconnected" on new federation setups


TonyGiorgio:
I'm using rc5 for the fedimint docker images and rc4 for the fedimintui docker images. 

Whenever I go through a new federation setup (using the mutiny tls scripts), I run into something that I believe might be related to this recent issue (https://github.com/fedimint/fedimint/issues/3806) & PR (https://github.com/fedimint/fedimint/pull/3833)

I was using RC2 before when I had that previous issue with a gateway that wouldn't connect to LND upon init, but restarting the docker images would fix the problem. However now a `docker compose restart` doesn't fix the problem. Here are the logs after a restart. 

```
fedimint-user@localhost:~/fedimint/docker$ docker compose logs -f gatewayd
docker-gatewayd-1  | 2023-12-06T04:57:20.009511Z  INFO ln_gateway: Starting gatewayd (version: 24113a64b270bf2ecb49b8f4554ba975025711db)
docker-gatewayd-1  | 2023-12-06T04:57:20.016193Z  INFO ln_gateway::lnd: Gateway configured to connect to LND LnRpcClient at 
docker-gatewayd-1  |  address: https://lnd:10009,
docker-gatewayd-1  |  tls cert path: /lnd_data/tls.cert,
docker-gatewayd-1  |  macaroon path: /lnd_data/data/chain/bitcoin/signet/admin.macaroon 
docker-gatewayd-1  | 2023-12-06T04:57:20.018245Z  INFO ln_gateway: Successfully started webserver
docker-gatewayd-1  | 2023-12-06T04:57:20.025080Z ERROR ln_gateway: Disconnected from Lightning Node. Waiting 5 seconds and trying again
docker-gatewayd-1  | 2023-12-06T04:57:25.026398Z  INFO ln_gateway::lnd: Gateway configured to connect to LND LnRpcClient at 
docker-gatewayd-1  |  address: https://lnd:10009,
docker-gatewayd-1  |  tls cert path: /lnd_data/tls.cert,
docker-gatewayd-1  |  macaroon path: /lnd_data/data/chain/bitcoin/signet/admin.macaroon 
docker-gatewayd-1  | 2023-12-06T04:57:25.086815Z ERROR ln_gateway: Disconnected from Lightning Node. Waiting 5 seconds and trying again
[...]

PR #3855, opened 2023-12-06
fix(gateway): allow setting configuration when disconnected

douglaz:
Looks reasonable to me. 

Closes https://github.com/fedimint/fedimint/issues/3852 and is a follow up of https://github.com/fedimint/fedimint/pull/3833

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -914,10 +914,6 @@ impl Gateway {
         }: SetConfigurationPayload,
     ) -> Result<()> {
         let gw_state = self.state.read().await.clone();
-        if matches!(gw_state, GatewayState::Disconnected) {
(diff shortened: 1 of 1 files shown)


Kodylow:
ACK, works when I run it too. @TonyGiorgio if you bump to the newer 0.17.3-beta for LND then it doesn't do the restarts

elsirion:
@okjodom @m1sterc001guy could you take a look at this later please?

elsirion (approved):

Kodylow:
@okjodom @m1sterc001guy tagging for visibility

okjodom:
I haven't tested new behavior, but change  makes sense to me

douglaz:
Although there are some errors, it works ¯\_(ツ)_/¯

justinmoon:
dev call: what is the benefit of setting the network on the gateway in the first place if this will not respect it?

PR #3858, opened 2023-12-06
 feat: ability to customize version string when building custom bins 

dpc:


fedimint-server/src/config/mod.rs
```diff
@@ -241,7 +241,7 @@ impl ServerConfig {
             modules: Default::default(),
         };
         let consensus = ServerConfigConsensus {
-            code_version: CODE_VERSION.to_string(),
+            code_version: version_hash,
```
dpc, line 244: **note:** Importantly before two different custom versions of Fedimint would not fail the version check during dkg.

elsirion:
I'm a bit confused what the use case is here.

elsirion (approved):

elsirion:
Doctests failing

dpc:
> So maybe I get it now: if I build a custom fedimintd, currently the version hash reported would be the one from the main fedimint dependency and not the custom repo? This allows overriding it?

Exactly.

PR #3861, opened 2023-12-07
feat: `just backport-pr`

dpc:
Using https://github.com/sqren/backport

Requires creating an access token and putting it in a local configuration file.

I think it uses the same code our backport bot is using, but will execute it locally using `npx`.


elsirion:
Certainly looks fancy, let's see how useful it actually is.

elsirion (approved):

PR #3872, opened 2023-12-07
chore: use larger amounts in tests

### LINKED ISSUE #3866: Increase amounts in some of our tests
We are noticing some issues that only happen with large amount transactions containing many ecash notes. Our tests would have caught this if they didn't always use tiny amounts.

shaurya947:
Closes #3866

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -87,7 +87,7 @@ pub async fn latency_tests(dev_fed: DevFed) -> Result<()> {
     let iterations = 30;
     let mut reissues = Vec::with_capacity(iterations);
     for _ in 0..iterations {
-        let notes = cmd!(fed, "spend", "50000").out_json().await?["notes"]
+        let notes = cmd!(fed, "spend", "1000000").out_json().await?["notes"]
             .as_str()
             .context("note must be a string")?
             .to_owned();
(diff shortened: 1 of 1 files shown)


devimint/src/main.rs
```diff
@@ -87,7 +87,7 @@ pub async fn latency_tests(dev_fed: DevFed) -> Result<()> {
     let iterations = 30;
     let mut reissues = Vec::with_capacity(iterations);
     for _ in 0..iterations {
-        let notes = cmd!(fed, "spend", "50000").out_json().await?["notes"]
+        let notes = cmd!(fed, "spend", "1000000").out_json().await?["notes"]
```
elsirion, line 90: This still isn't a whole lot :laughing: (1k sat)

dpc: It has so many zeros already... :D

Maybe we need `ksats(100).to_string()`

douglaz (approved):
LGTM

devimint/src/main.rs
```diff
@@ -373,8 +373,8 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
     info!("Testing Client");
     // ## reissue e-cash
     info!("Testing reissuing e-cash");
-    const CLIENT_START_AMOUNT: u64 = 420000;
-    const CLIENT_SPEND_AMOUNT: u64 = 42;
+    const CLIENT_START_AMOUNT: u64 = 5_000_000_000;
```
elsirion, line 376: This is the only change that makes the test test any amount of significance.

elsirion (approved):
Marginally better, but we should have a "reissue 10 BTC when the wallet had only 10 BTC in total" (spending everything is the worst case, up to 4 notes per denomination) test case to give us more confidence.

PR #3873, opened 2023-12-07
chore: change warning to a info

douglaz:
This happened during our federation setup on deep dive and it was confusing


justinmoon (dismissed):

gateway/ln-gateway/src/lib.rs
```diff
@@ -1130,7 +1130,7 @@ impl Gateway {
                                     }
                                 }
                             } else {
-                                warn!("GatewayState must be Running to register with federation. Current state: {gateway_state:?}");
+                                info!("Will not register federation yet because gateway still not in Running state. Current state: {gateway_state:?}. Will keep waiting...");
```
elsirion, line 1133: I think this is the reason why on startup the gateway doesn't register. At least I also observed this warning after restart. I'm ok with this patch, but would like to see if we can do something to retry sooner in that case.

elsirion (approved):

gateway/ln-gateway/src/lib.rs
douglaz: Nice catch. I didn't connect the two problems. Changed to retry with a smaller delay.

elsirion (approved):

fedimint-backports:
- #3878

PR #3877, opened 2023-12-08
fix: wasm-bindgen 0.2.88 was yanked

dpc:
https://crates.io/crates/wasm-bindgen/versions

This causes audits to fail:

https://github.com/fedimint/fedimint/actions/runs/7137792394/job/19438296750

### DIFF
--- a/fedimint-wasm-tests/Cargo.toml
+++ b/fedimint-wasm-tests/Cargo.toml
@@ -21,7 +21,7 @@ fedimint-mint-common = { path = "../modules/fedimint-mint-common" }
 fedimint-ln-client = { path = "../modules/fedimint-ln-client" }
 fedimint-wallet-client = { path = "../modules/fedimint-wallet-client" }
 js-sys = "0.3.61"
-wasm-bindgen = "=0.2.88" # must match the nix provided wasm-bindgen-cli version
+wasm-bindgen = "=0.2.89" # must match the nix provided wasm-bindgen-cli version
 wasm-bindgen-futures = "0.4.34"
 wasm-bindgen-test = "0.3.34"
 gloo-net = "0.2.6"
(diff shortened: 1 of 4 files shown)


fedimint-wasm-tests/Cargo.toml
```diff
@@ -21,7 +21,7 @@ fedimint-mint-common = { path = "../modules/fedimint-mint-common" }
 fedimint-ln-client = { path = "../modules/fedimint-ln-client" }
 fedimint-wallet-client = { path = "../modules/fedimint-wallet-client" }
 js-sys = "0.3.61"
-wasm-bindgen = "=0.2.88" # must match the nix provided wasm-bindgen-cli version
+wasm-bindgen = "=0.2.89" # must match the nix provided wasm-bindgen-cli version
```
elsirion, line 24: Tests fail because the nix-provided version diverged as warned against in the comment.

flake.nix
```diff
@@ -28,6 +28,26 @@
...
+                  };
+
+                  nativeBuildInputs = [ final.pkg-config ];
+
+                  buildInputs = [ final.openssl ] ++ lib.optionals stdenv.isDarwin [ final.curl final.Security ];
+
+                  nativeCheckInputs = [ final.nodejs ];
+
+                  # tests require it to be ran in the wasm-bindgen monorepo
+                  doCheck = false;
```
elsirion, line 48: `nativeCheckInputs` feels redundant, why isn't it? (just trying to improve my Nix understanding)

elsirion (approved):

flake.nix
dpc: Huh. I guess it makes sense to have inputs that are only used in the check phase (which here is disabled anyway, but I guess it was added for completeness?). I copied the whole thing from the nixpkgs.

elsirion (approved):

PR #3880, opened 2023-12-08
chore: separate fedimint config info from gateway info endpoint

### LINKED ISSUE #3815: `gateway-cli info` displays far too much information
It includes the entire client config. This is a problem because when you run `gateway-cli info` you can't see at a glance how many federations you're in.

mayrf:
Fixes #3815. I am not sure it's completely done and would appreciate feedback.

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -6,8 +6,8 @@ use fedimint_core::BitcoinAmountOrAll;
 use fedimint_logging::TracingSetup;
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
 use ln_gateway::rpc::{
-    BackupPayload, BalancePayload, ConnectFedPayload, DepositAddressPayload, LeaveFedPayload,
-    RestorePayload, SetConfigurationPayload, WithdrawPayload,
+    BackupPayload, BalancePayload, ConfigPayload, ConnectFedPayload, DepositAddressPayload,
+    LeaveFedPayload, RestorePayload, SetConfigurationPayload, WithdrawPayload,
 };
 use serde::Serialize;
 
@@ -30,6 +30,11 @@ pub enum Commands {
     VersionHash,
     /// Display high-level information about the Gateway
     Info,
+    /// Display config information about the Gateways federation
+    Config {
+        #[clap(long)]
+        federation_id: Option<FederationId>,
+    },
     /// Check gateway balance
     Balance {
         #[clap(long)]
@@ -102,6 +107,12 @@ async fn main() -> anyhow::Result<()> {
         }
         Commands::Info => {
             let response = client().get_info().await?;
+
+            print_response(response).await;
+        }
+
+        Commands::Config { federation_id } => {
+            let response = client().get_config(ConfigPayload { federation_id }).await?;
 
             print_response(response).await;
         }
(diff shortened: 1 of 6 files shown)


gateway/cli/src/main.rs
```diff
@@ -30,6 +30,8 @@ pub enum Commands {
     VersionHash,
     /// Display high-level information about the Gateway
     Info,
+    /// Display config information about the Gateways federation
+    Config,
```
elsirion, line 34: Imo this should be a per-federation command, i.e. takes the federation id as argument.

elsirion (commented):
Looks good, but I think we can simplify it a bit more.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1217,24 +1233,37 @@ impl Gateway {
...
+        federation_id: FederationId,
+    ) -> FederationConfigInfo {
+        let balance_msat = client.get_balance().await;
+        let config = client.get_config().clone();
+
+        FederationConfigInfo {
+            federation_id,
+            balance_msat,
+            config,
+        }
```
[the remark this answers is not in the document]
dpc: If we remove "balance", ... should it be just a map? id -> config ?

gateway/cli/src/main.rs
dpc: I wonder which is better. This or `--federation-id` acting as a filter, and lack of it meaning "all".

elsirion: Why do we need the "all" version of the command?

dpc: I don't know. If you're in one federation only you can get the config without specifying the id. Or if you're writing a script you could loop over all in one go (one `| jq ...`). Not very strong reasons, but also doesn't change anything for someone that just wants one config.

gateway/ln-gateway/src/lib.rs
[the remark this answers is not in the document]
dpc: I guess it's a follow up to do we allow listing all or not. If we do , will be easier to parse/filter, if not it's kind of needless.

mayrf:
Sorry for the delay. I went for dpcs suggestion and implemented an --federation-id option, returning either the config of the federation specified or those of all federations, in case no federation-id is specified.

gateway/ln-gateway/src/rpc/rpc_client.rs
```diff
@@ -94,7 +99,7 @@ impl GatewayRpcClient {
             .base_url
             .join("/set_configuration")
             .expect("invalid base url");
-        self.call(Method::POST, url, Some(payload)).await
+        self.call_post(url, payload).await
```
elsirion, line 102: Do we want to have composite calls in fedimint-cli? Can be fixed in another PR.

elsirion (approved):

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -82,6 +92,12 @@ pub struct GatewayInfo {
     pub network: Option<Network>,
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct GatewayFedConfig {
+    pub federations: BTreeMap<FederationId, JsonClientConfig>,
+    // pub federations: BTreeMap<FederationId, ClientConfig>,
```
okjodom, line 98: nit: let's clean up commented code

okjodom (approved):
left nit comments, yet this looks and works great!

tACK https://github.com/fedimint/fedimint/pull/3880/commits/1bc5bfd33cd65d9975775d11e56623ee93193b25

okjodom (approved):

elsirion (approved):

PR #3882, opened 2023-12-08
feat(ln-client): make LN gateway known upfront

elsirion:
Otherwise, fees that will be paid cannot be shown to users upfront.

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -254,11 +254,14 @@ pub async fn handle_command(
             let lightning_module = client.get_first_module::<LightningClientModule>();
             lightning_module.select_active_gateway().await?;
 
+            let gateway = lightning_module.select_active_gateway_opt().await;
             let OutgoingLightningPayment {
                 payment_type,
                 contract_id,
                 fee,
-            } = lightning_module.pay_bolt11_invoice(bolt11, ()).await?;
+            } = lightning_module
+                .pay_bolt11_invoice(gateway, bolt11, ())
+                .await?;
             let operation_id = payment_type.operation_id();
             info!("Gateway fee: {fee}, payment operation id: {operation_id}");
             if finish_in_background {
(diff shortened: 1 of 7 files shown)


justinmoon (approved):
Looks good, but seems like a lightning test is failing

Should we do the same thing to `create_bolt11_invoice`?

elsirion:
Turns out we have tests for being able to pay internal invoices even if no LN GW is present (which is probably good). I'm trying a new, minimal approach, but longer term we need to think about how to best pull that pay fn apart so that users can be better informed.

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -142,12 +150,13 @@ async fn pay_valid_invoice(
...
     // User client pays test invoice
     let user_lightning_module = &user_client.get_first_module::<LightningClientModule>();
+    let gateway = user_lightning_module.select_active_gateway().await.ok();
     let OutgoingLightningPayment {
         payment_type,
         contract_id,
         fee: _,
     } = user_lightning_module
-        .pay_bolt11_invoice(invoice.clone(), ())
+        .pay_bolt11_invoice(gateway, invoice.clone(), ())
```
douglaz, line 159: why not call `pay_invoice` here?

fedimint-cli/src/client.rs
```diff
@@ -254,11 +254,14 @@ pub async fn handle_command(
             let lightning_module = client.get_first_module::<LightningClientModule>();
             lightning_module.select_active_gateway().await?;
 
+            let gateway = lightning_module.select_active_gateway().await.ok();
```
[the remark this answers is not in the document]
elsirion: I was a bit unsure about this, the reason why it's an option in the first place is that we currently have support for internal payments without a gateway being present. So it wasn't entirely clear to me if thats something we'd want to warn about till it's an actual problem. But you are right that it makes debugging the root cause harder, so will add it.

gateway/ln-gateway/tests/integration_tests.rs
elsirion: I think I got tired of repeating the same code after a while, introduced the fn but forgot that one.

douglaz (approved):

okjodom (approved):

PR #3889, opened 2023-12-08
refactor: amount improvements

douglaz:
The initial motivation for this PR was making `gateway-cli withdraw --amount` (which was using `bitcoin::Amount` through `BitcoinAmountOrAll`) accept a similar syntax to other commands that accept an `amount` (most use `fedimint_core::Amount` or `parse_fedimint_amount`).

Now we upgrade `fedimint_core::Amount` and let its `from_str` accept what `parse_fedimint_amount` was able and what `bitcoin::Amount::from_str` was able to do.

We also make `fedimint_core::Amount` serializable through  `#[serde(with = "amount::serde::as_sat")` which  is analogous to `#[serde(with = "bitcoin::util::amount::serde::as_sat")]`.

The final result is that every command line is backwards compatible but also accepts the new syntax.

### DIFF
--- a/fedimint-core/src/lib.rs
+++ b/fedimint-core/src/lib.rs
@@ -7,6 +7,7 @@ use std::io::Error;
 use std::num::ParseIntError;
 use std::str::FromStr;
 
+use anyhow::bail;
 use bitcoin::Denomination;
 use bitcoin_hashes::hash_newtype;
 use bitcoin_hashes::sha256::Hash as Sha256;
@@ -85,6 +86,8 @@ impl FromStr for PeerId {
     }
 }
 
+pub const SATS_PER_BITCOIN: u64 = 100_000_000;
+
 /// Represents an amount of BTC inside the system. The base denomination is
 /// milli satoshi for now, this is also why the amount type from rust-bitcoin
 /// isn't used instead.
@@ -118,9 +121,13 @@ impl Amount {
         Amount::from_msats(sats * 1000)
     }
 
+    pub const fn from_bitcoins(bitcoins: u64) -> Amount {
+        Amount::from_sats(bitcoins * SATS_PER_BITCOIN)
+    }
+
     pub fn from_str_in(s: &str, denom: Denomination) -> Result<Amount, ParseAmountError> {
         if let Denomination::MilliSatoshi = denom {
-            return Self::from_str(s);
+            return Ok(Self::from_msats(s.parse()?));
         }
         let btc_amt = bitcoin::util::amount::Amount::from_str_in(s, denom)?;
         Ok(Self::from(btc_amt))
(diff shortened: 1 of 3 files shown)


fedimint-core/src/lib.rs
```diff
@@ -131,6 +137,10 @@ impl Amount {
             msats: self.msats.saturating_sub(other.msats),
         }
     }
+
+    pub const fn sats(&self) -> u64 {
+        self.msats / 1000
+    }
```
elsirion, line 143: Should at least have a warning about potentially losing msats.

fedimint-core/src/lib.rs
```diff
@@ -146,25 +156,43 @@ pub fn sats(amount: u64) -> Amount {
...
+            use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+            pub fn serialize<S: Serializer>(a: &crate::Amount, s: S) -> Result<S::Ok, S::Error> {
+                u64::serialize(&a.sats(), s)
+            }
+
+            pub fn deserialize<'d, D: Deserializer<'d>>(d: D) -> Result<crate::Amount, D::Error> {
+                Ok(crate::Amount::from_sats(u64::deserialize(d)?))
+            }
+        }
```
elsirion, line 175: I don't think we should make it easy to accidentally throw away msat precision like that. E.g. in any JSON API call amounts will be rounded down that way.

fedimint-core/src/lib.rs
```diff
@@ -146,25 +156,43 @@ pub fn sats(amount: u64) -> Amount {
...
+    }
+}
+
 /// Amount of bitcoin to send, or "all" to send all available funds
 #[derive(Debug, Eq, PartialEq, Copy, Hash, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub enum BitcoinAmountOrAll {
```
elsirion, line 182: If we want to use `AmountOrAll` for amounts with different precisions we should make it generic over the amount type. Converting back and forth between different precisions can lead to all kinds of problems:

* APIs dealing with sat precision now silently throw away the msat part of the amount instead of returning an error
* APIs dealing with msat precision might accidentally lose it due to how serialization is implemented in this PR

elsirion (changes requested):

fedimint-core/src/lib.rs
```diff
@@ -131,6 +137,10 @@ impl Amount {
             msats: self.msats.saturating_sub(other.msats),
         }
     }
+
+    pub const fn sats(&self) -> u64 {
```
dpc, line 141: **thought:**  `sats_round_down()`

dpc (approved):

elsirion (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-3889-to-releases/v0.2 origin/releases/v0.2
cd .worktree/backport-3889-to-releases/v0.2
git switch --create backport-3889-to-releases/v0.2
git cherry-pick -x 3dd3eb8fbe42e32cb101f9a64313be241b266408
```

PR #3894, opened 2023-12-09
fix: electrs new port causing bind conflicts

dpc:


elsirion (approved):
ACKing because we have this problem now, but I think in the future we should just deactivate the nixpkgs up-to-dateness check on release branches and not change the test/build env.

dpc:
@elsirion This PR does not update nixpkgs. It did on the master version, but here seemed unnecessary. The problem with the ports on master was actually not caused by the nixpkgs update. It was already there and just happened to surface more often after update (something (electrs/esplora-electrs) got faster/slower to start I guess.).

PR #3896, opened 2023-12-09
fix: supply LN invoice features to LN gateway node

### LINKED ISSUE #3895: Inter-Federation payments are broken for private payments
Apparently LND requires the destination node features for non-public recipient nodes, e.g. other fedimint users from different federations. We currently don't pass on feature flags in `PrunedInvoice`.

Why didn't we notice that before? We only tested with "public" nodes, i.e. ones that are known to the sender from gossip, and in that case LND will just happily look up the supported features from the node announcement.

elsirion:
Without this, private payments to non-public nodes are failing. Fixes #3895

## Completeness
Let's go through the field list of [`SendPaymentV2`](https://lightning.engineering/api-docs/api/lnd/router/send-payment-v2/index.html) to avoid missing any other required fields (I originally went by the docs *"If no payment request is specified, the following fields are required: dest, amt and payment_hash."* but that obviously didn't go that well).

* `dest`: we supply :white_check_mark:
* `amt`: redundant with the next one :+1: 
* `amt_msat`: we supply :white_check_mark: 
* `payment_hash`: we supply :white_check_mark:
* `final_cltv_delta`: we supply :white_check_mark:
* `payment_addr`: we supply, turns out to be the payment secret :white_check_mark:
* `payment_request`: not needed, we supply all necessary arguments manually :+1: 
* `timeout_seconds`: we supply :white_check_mark:
* `fee_limit_sat`: redundant with the next one :+1: 
* `fee_limit_msat`: we supply :white_check_mark:
* `outgoing_chan_id`: unnecessary, we don't want to control the outgoing channel :+1: 
* `outgoing_chan_ids`: unnecessary, we don't want to control the outgoing channel :+1: 
* `last_hop_pubkey`: unnecessary, we don't want to control the last hop :+1: 
* `cltv_limit`: we supply :white_check_mark:
* `route_hints`: we supply :white_check_mark:
* `dest_custom_records`: we aren't interested in passing additional TLV records :+1: 
* `allow_self_payment`: defaults to false how we  want it :+1: 
* `dest_features`: that's what this PR is about :white_check_mark:
* `max_parts`: I think the defaults will do, this is mostly about performance optimization afaik :+1: 
* `no_inflight_updates`: set to false :white_check_mark: 
* `max_shard_size_msat`: we don't want to influence MPP splitting behavior :+1: 
* `amp`: `false` seems like a good default :+1: 
* `time_pref`: we should make this configurable, but good enough for now given that `pay` also doesn't set it :+1: 

## Compatibility

On the API level this change is made backwards-compatible by using the empty feature set as default (as was the case so far). Outgoing LN state machines in the state `Funding` will cause a panic when trying to decode them. If we don't backport to the unreleased 0.2 (#3897) this would require a state machine migration.

## Alternatives

Deactivating private payments for now.

### DIFF
--- a/gateway/ln-gateway/src/lnd.rs
+++ b/gateway/ln-gateway/src/lnd.rs
@@ -3,6 +3,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
+use anyhow::ensure;
 use async_trait::async_trait;
 use bitcoin_hashes::hex::ToHex;
 use fedimint_core::task::{sleep, TaskGroup};
@@ -571,12 +572,18 @@ impl ILnRpcClient for GatewayLndClient {
                         failure_reason: format!("max delay exceeds valid LND range {error:?}"),
                     })?;
 
+            let dest_features = wire_features_to_lnd_feature_vec(&invoice.destination_features)
+                .map_err(|e| LightningRpcError::FailedPayment {
+                    failure_reason: e.to_string(),
+                })?;
+
             debug!("LND payment does not exist for invoice {invoice:?}, will attempt to pay");
             let payments = client
                 .router()
                 .send_payment_v2(SendPaymentRequest {
                     amt_msat,
                     dest: invoice.destination.serialize().to_vec(),
+                    dest_features,
                     payment_hash: invoice.payment_hash.to_vec(),
                     payment_addr: invoice.payment_secret.to_vec(),
                     route_hints: route_hints_to_lnd(&invoice.route_hints),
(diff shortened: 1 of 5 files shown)


gateway/ln-gateway/src/lnd.rs
```diff
@@ -749,3 +756,30 @@ fn route_hints_to_lnd(
...
+    ensure!(
+        features_wire_encoded.len() <= 1_000,
+        "Will not process feature bit vectors larger than 1000 byte"
+    );
+
+    let lnd_features = features_wire_encoded
+        .iter()
+        .rev()
+        .enumerate()
+        .flat_map(|(byne_idx, &feature_byte)| {
```
bradleystachurski, line 770: [nit] If you need to do another force push, `s/byne/byte`

gateway/ln-gateway/src/lnd.rs
```diff
@@ -749,3 +756,30 @@ fn route_hints_to_lnd(
         })
         .collect()
 }
+
+fn wire_features_to_lnd_feature_vec(features_wire_encoded: &[u8]) -> anyhow::Result<Vec<i32>> {
```
bradleystachurski, line 760: I don't think it's a blocker for merging, but should we make a followup that adds some test vectors to ensure we're correctly translating?

elsirion: Good point, added a test.

bradleystachurski (commented):
Looks good! Thanks for adding a test and updating the description with the completeness checklist. I don't see any issues with this approach, but I'll hold off approving until other folks with more context can review.

> `time_pref`: we should make this configurable, but good enough for now given that pay also doesn't set it 👍

Is there an existing issue to track?

justinmoon (approved):

PR #3893, opened 2023-12-09
fix(devimint): off-by-one error for mempool transactions

bradleystachurski:
Fixes #3884

- Add helper to mine enough blocks to finalize mempool transactions
- Remove `generate_epochs` since epochs are no longer part of consensus

PR #3898, opened 2023-12-09
refactor(devimint): add get_block_count to bitcoind

bradleystachurski:
Followup for https://github.com/fedimint/fedimint/pull/3893#discussion_r1421246877

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -120,6 +120,15 @@ impl Bitcoind {
         self.client.clone()
     }
 
+    /// Returns the total number of blocks in the chain.
+    ///
+    /// Fedimint's IBitcoindRpc considers block count the total number of
+    /// blocks, where bitcoind's rpc returns the height. Since the genesis
+    /// block has height 0, we need to add 1 to get the total block count.
+    pub fn get_block_count(&self) -> Result<u64> {
+        Ok(self.client().get_block_count()? + 1)
+    }
+
     pub async fn mine_blocks(&self, block_num: u64) -> Result<()> {
         info!(target: LOG_DEVIMINT, ?block_num, "Mining bitcoin blocks");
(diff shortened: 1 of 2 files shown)


devimint/src/external.rs
```diff
@@ -120,6 +120,15 @@ impl Bitcoind {
         self.client.clone()
     }
 
+    /// Returns the total number of blocks in the chain.
+    ///
+    /// Fedimint's IBitcoindRpc considers block count the total number of
+    /// blocks, where bitcoind's rpc returns the height. Since the genesis
+    /// block has height 0, we need to add 1 to get the total block count.
+    pub fn get_block_count(&self) -> Result<u64> {
+        Ok(self.client().get_block_count()? + 1)
```
elsirion, line 129: We already account for that afaik:

https://github.com/fedimint/fedimint/blob/747585f226488da67fbe0cd58894c789e131ef89/fedimint-bitcoind/src/bitcoincore.rs#L56-L61

Maybe something else is wrong?

bradleystachurski: I also found this a bit confusing. The client for `Bitcoind` in devimint external uses `bitcoincore_rpc::Client` instead of `IBitcoindRpc`, so we're essentially duplicating impl.

Should we consider changing the devimint external client to `IBitcoindRpc`? If so, I can create an issue.

bradleystachurski: Hmm, on further thought I'm not sure it's a good idea to switch to `IBitcoindRpc` in devimint external. We rely on accessing several methods from bitcoind's rpc that would make unifying with `IBitcoindRpc` pretty messy.

elsirion: Oh, then nevermind.

elsirion (approved):

PR #3899, opened 2023-12-10
feat: Support receiving LN to multiple outputs

benthecarman:
If possible would be nice to get this in before the 0.2 release since adding this would be backwards incompatible. Feel free to take this over for deadlines if needed

This attempts to add the option to receive to multiple pubkeys for a single lighting receive, this is so you do things split payments but still only pay a single invoice, useful for things like donations addresses, tips, or taking a service fee.

I am pretty sure what I had to change for `process_input` is susceptible to someone just continually replaying with their amount until it is drained, not sure how that function is supposed to work exactly.


elsirion:
Can't we already do that client-side via MuSig in the cooperative setting? I'm not very comfortable with changing a) consensus logic b) the already overly complex LN logic so close to final cutoff.

Also, it's not impossible to patch in later, just more annoying if we want it in the current LN module. If it's not urgent I'd rather work with @joschisan on LNv2 which is already under development and is a far more optimal design.

PR #3902, opened 2023-12-11
Make CLI spend e-cash more configurable

### LINKED ISSUE #3834: `fedimint-cli spend` should have a flag to give exact change or error if it can't
Currently `fedimint-cli spend <amount>` will give us `<amount>` if it has those notes on hand, and give more than we ask for if it doesn't.

I was just writing a test and this behavior caused my test to be flaky. It would be nice for this reason (and probably in general) to be able to request exact amounts from `fedimint-cli spend`.

elsirion:
Fixes #3834

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -61,9 +61,16 @@ pub enum ClientCmd {
     Reissue { oob_notes: OOBNotes },
     /// Prepare notes to send to a third party as a payment
     Spend {
+        /// The amount of e-cash to spend
         amount: Amount,
+        /// If the exact amount cannot be represented, return e-cash of a higher
+        /// value instead of failing
         #[clap(long)]
         allow_overpay: bool,
+        /// After how many seconds we will try to reclaim the e-cash if it
+        /// hasn't been redeemed by the recipient. Defaults to one week.
+        #[clap(long, default_value_t = 60 * 60 * 24 * 7)]
+        timeout: u64,
     },
     /// Verifies the signatures of e-cash notes, but *not* if they have been
     /// spent already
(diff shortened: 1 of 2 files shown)


fedimint-cli/src/client.rs
```diff
@@ -57,8 +59,17 @@ pub enum ClientCmd {
...
+        /// The amount of e-cash to spend
         #[clap(value_parser = parse_fedimint_amount)]
         amount: Amount,
+        /// If the exact amount cannot be represented, return e-cash of a higher
+        /// value instead of failing
+        #[clap(long)]
+        allow_overpay: bool,
+        /// After how many seconds we will try to reclaim the e-cash if it
+        /// hasn't been redeemed by the recipient
+        #[clap(long, default_value_t = 3600)]
```
dpc, line 71: **thought:**  Is this automatic behavior? If so, isn't this a bit low for a default for something that might be passed in a DM or even piece of paper?

dpc (approved):

fedimint-cli/src/client.rs
elsirion: Good point, what would you propose? A week?

dpc: Sounds reasonable.

BTW. Maybe we should also `eprintln!` the deadline to warn/educate the user?

elsirion: could that be a `warn` level log? These are still shown and should be printed to stderr anyway (if not we should fix that).

dpc: The way I think about it is: logging is for reporting what the program is doing, while `(e)println` for the actual output for the user. Not feeling too strongly about it one way or another.

elsirion: We need a third stream, stdlog! :laughing:

justinmoon: In the CLI what would trigger this refund?

elsirion: Yep, basically anything that runs the SM executor for long enough. What I like to do to trigger this manually is create a LN invoice I don't intend to pay and then await it. We might want a more general purpose "run executor" command.

douglaz (approved):
Requires rebase, but LGTM

Maan2003 (approved):

PR #3906, opened 2023-12-11
fix: downstream consumer broken (0.1)

dpc:
See commit messages.


elsirion (approved):
Can we please EOL 0.1? :laughing: 

On master/0.2 this isn't a problem anymore as we published our own fork `fedimint-cln-rpc`.

PR #3927, opened 2023-12-12
feat: better invoice expiration

douglaz:
Anyone that has used lightning in the real world has probably suffered with the usual 1 hour default expiration time on invoices. Let's help a bit the users by giving them a better default.

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -56,7 +56,6 @@ use futures::StreamExt;
 use incoming::IncomingSmError;
 use lightning_invoice::{
     Bolt11Invoice, Currency, InvoiceBuilder, PaymentSecret, RouteHint, RouteHintHop, RoutingFees,
-    DEFAULT_EXPIRY_TIME,
 };
 use rand::seq::IteratorRandom;
 use rand::{CryptoRng, Rng, RngCore};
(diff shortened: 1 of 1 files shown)


elsirion:
Should it just be configurable instead? Let the integrator worry about that: they can choose a default or give the user a way to configure it.

douglaz:
> Should it just be configurable instead? Let the integrator worry about that: they can choose a default or give the user a way to configure it.

It's already configurable: just use `Some(expiry_time)` instead of `None`. This default is only for clueless wallets and integrators.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -770,7 +773,7 @@ impl LightningClientModule {
             .min_final_cltv_expiry_delta(18)
             .payee_pub_key(node_public_key)
             .expiry_time(Duration::from_secs(
-                expiry_time.unwrap_or(DEFAULT_EXPIRY_TIME),
+                expiry_time.unwrap_or(DEFAULT_INVOICE_EXPIRY_TIME.as_secs()),
```
dpc, line 776: **nitpick:** That conversion back and forth from/to `Duration` looks weird. Would `expire_time.map(Duration::from_secs).unwrap_or(DEFAULT_INVOICE_EXPIRY_TIME)` work? I would just do somewhere before:

```
let expire_time = expire_type.map(Duration::from_secs);
```

to get it of the way, then `expire_time.unwrap_or(DEFAULT_INVOICE_EXPIRY_TIME)` would work just perfectly.

dpc (approved):
No idea if this is a good idea, but technically PR LGTM.

justinmoon:
Default expiration of 24 hours sounds good to me

Issue #3881, opened 2023-12-08
Lightning client payment receive state machines have issues


douglaz:
There are at least two issues

- It will timeout even if all the errors are connection errors (i.e the system is offline)
- It won't timeout if the receiving machine is frequently interrupted

Relevant state transitions: https://github.com/fedimint/fedimint/blob/b891544bb779a30e8764a06a10434065410ec616/modules/fedimint-ln-client/src/receive.rs#L169

PR #3929, opened 2023-12-12
fix: more robust lightning receive

douglaz:
Closes https://github.com/fedimint/fedimint/issues/3881

Timeout now will happen at the right time: when we're sure the preimage decryption is pending and the expiry time relative to the invoice timestamp has passed.

There are no risks anymore of the timeout not happening because the state machines have been interrupted or it incorrectly happening while the client is dealing with network issues.

### DIFF
--- a/fedimint-core/src/time.rs
+++ b/fedimint-core/src/time.rs
@@ -12,3 +12,10 @@ pub fn now() -> SystemTime {
     SystemTime::UNIX_EPOCH
         + std::time::Duration::from_secs_f64(js_sys::Date::new_0().get_time() / 1000.)
 }
+
+/// Returns the duration since the Unix epoch
+pub fn duration_since_epoch() -> std::time::Duration {
+    now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("time to work")
+}
--- a/modules/fedimint-ln-client/src/pay.rs
+++ b/modules/fedimint-ln-client/src/pay.rs
@@ -1,5 +1,5 @@
 use std::sync::Arc;
-use std::time::{Duration, UNIX_EPOCH};
+use std::time::Duration;
 
 use bitcoin_hashes::sha256;
 use fedimint_client::sm::{ClientSMDatabaseTransaction, State, StateTransition};
@@ -10,7 +10,7 @@ use fedimint_core::config::FederationId;
 use fedimint_core::core::{Decoder, OperationId};
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::task::sleep;
-use fedimint_core::time::now;
+use fedimint_core::time::duration_since_epoch;
 use fedimint_core::{Amount, OutPoint, TransactionId};
 use fedimint_ln_common::api::LnFederationApi;
 use fedimint_ln_common::contracts::outgoing::OutgoingContractData;
@@ -643,13 +643,7 @@ impl PaymentData {
     }
 
     pub fn is_expired(&self) -> bool {
-        let Some(now) = now().duration_since(UNIX_EPOCH).map(|t| t.as_secs()).ok() else {
-            // Something is very wrong (time out of bounds), we'll not attempt too pay the
-            // invoice
-            return true;
-        };
-
-        self.expiry_timestamp() < now
+        self.expiry_timestamp() < duration_since_epoch().as_secs()
     }
 
     /// Returns the expiry timestamp in seconds since the UNIX epoch
--- a/modules/fedimint-ln-client/src/receive.rs
+++ b/modules/fedimint-ln-client/src/receive.rs
@@ -1,4 +1,3 @@
-use std::ops::Add;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -12,11 +11,12 @@ use fedimint_core::task::sleep;
 use fedimint_core::{OutPoint, TransactionId};
 use fedimint_ln_common::api::LnFederationApi;
 use fedimint_ln_common::contracts::incoming::IncomingContractAccount;
+use fedimint_ln_common::contracts::DecryptedPreimageStatus;
 use fedimint_ln_common::{LightningClientContext, LightningInput};
 use lightning_invoice::Bolt11Invoice;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
-use tracing::{error, info, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::LightningClientStateMachines;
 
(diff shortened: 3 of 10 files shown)


modules/fedimint-ln-client/src/receive.rs
```diff
@@ -266,23 +278,6 @@ impl LightningReceiveConfirmedInvoice {
 
         global_context.claim_input(dbtx, client_input).await
     }
-
-    async fn await_payment_timeout(timeout: Duration) {
-        // Add 10% of the invoice expiry_time as a buffer before we stop awaiting the
-        // payment
-        let timeout_buffer = timeout.as_secs_f64() * 0.1;
```
elsirion, line 273: I think having a certain percentage of grace period (previously 10%) is good since clocks aren't perfectly synchronized. Also the contract might get funded right before timeout and we'd just ignore it.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -202,12 +195,32 @@ impl LightningReceiveConfirmedInvoice {
...
             {
-                Ok((incoming_contract_account, preimage)) => match preimage {
-                    Some(_) => return Ok(incoming_contract_account),
-                    None => return Err(LightningReceiveError::InvalidPreimage),
+                Ok((incoming_contract_account, status)) => match status {
+                    DecryptedPreimageStatus::Pending => {
+                        // only when we are sure that the invoice is still pending that we can check
+                        // for a timeout
+                        let now = fedimint_core::time::now();
+                        match invoice.timestamp().checked_add(invoice.expiry_time()) {
```
elsirion, line 206: [`timestamp()` is only available with `std`](https://docs.rs/lightning-invoice/latest/lightning_invoice/struct.Bolt11Invoice.html#method.timestamp), this might break MutinyWallet again. @benthecarman @TonyGiorgio There are alternative, more clunky methods to get the timestamp, e.g. [`duration_since_epoch`](https://docs.rs/lightning-invoice/latest/lightning_invoice/struct.Bolt11Invoice.html#method.duration_since_epoch)

benthecarman: Yeah that will break

TonyGiorgio: Yeah if you can help it, `duration_since_epoch` is better, though in this case we could alter our fork to allow that call, since I followed the chain and it doesn't appear to be calling `SystemTime::now()`.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -643,13 +643,7 @@ impl PaymentData {
...
 
     pub fn is_expired(&self) -> bool {
-        let Some(now) = now().duration_since(UNIX_EPOCH).map(|t| t.as_secs()).ok() else {
-            // Something is very wrong (time out of bounds), we'll not attempt too pay the
-            // invoice
-            return true;
-        };
-
-        self.expiry_timestamp() < now
+        self.expiry_timestamp() < duration_since_epoch().as_secs()
```
dpc, line 646: **thought:** Not a problem of this PR, but would rename `expiry_timestamp` to `expiry_timestamp_secs` or something.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -200,14 +193,30 @@ impl LightningReceiveConfirmedInvoice {
...
             {
-                Ok((incoming_contract_account, preimage)) => match preimage {
-                    Some(_) => return Ok(incoming_contract_account),
-                    None => return Err(LightningReceiveError::InvalidPreimage),
+                Ok((incoming_contract_account, status)) => match status {
+                    DecryptedPreimageStatus::Pending => {
+                        // only when we are sure that the invoice is still pending that we can check
+                        // for a timeout
+                        const BUFFER: f64 = 1.10; // 10% buffer to account for clock skew
+                        let invoice_since_epoch_secs =
```
dpc, line 208: **issue:** 

```
> date +%s
1702448004
```

Isn't 10% of `1702448004`s ... 5 years? Do we really need such a buffer? Seems like 60s should be enough for everyone?

BTW. "buffer" seems like a wrong word. Would just inline it to not have to come up with a better name.

elsirion: The previous implementation applied it on top of the relative timeout. So 6min per hour.

fedimint-core/src/time.rs
```diff
@@ -12,3 +12,10 @@ pub fn now() -> SystemTime {
     SystemTime::UNIX_EPOCH
         + std::time::Duration::from_secs_f64(js_sys::Date::new_0().get_time() / 1000.)
 }
+
+/// Returns the duration since the Unix epoch
+pub fn duration_since_epoch() -> std::time::Duration {
+    now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("time to work")
```
elsirion, line 20: nit: I avoided such expects more recently since a horribly misconfigured system might have a time from the past.

modules/fedimint-ln-client/src/receive.rs
elsirion: @douglaz the factor should be applied on the relative expiry time, not the absolute one.

elsirion (changes requested):

modules/fedimint-ln-client/src/receive.rs
douglaz: I'm sorry, my commit didn't include the fix. Look now.

fedimint-core/src/time.rs
douglaz: If your system is so misconfigured you should expect things to blow up

modules/fedimint-ln-client/src/receive.rs
dpc: I think fixed offset between computers is much more likely than clocks going 10% slower/faster (which would right away caused fixed offset anyway). Unless I'm missing something.

BTW. Makes me wonder if federation should have a consensus on time as well. :D

fedimint-core/src/time.rs
dpc: There's some room here for maicious input causing panics, I guess.

dpc: Actually - for this to panic, the `now()` would have to return a date before unix epoch... ? Which is probably impossible at least on Linux, because the representation of time start with unix epoch?

modules/fedimint-ln-client/src/receive.rs
elsirion: Also good to always allow a fixed grace period as long as there is one.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -200,14 +193,29 @@ impl LightningReceiveConfirmedInvoice {
...
                 .await
             {
-                Ok((incoming_contract_account, preimage)) => match preimage {
-                    Some(_) => return Ok(incoming_contract_account),
-                    None => return Err(LightningReceiveError::InvalidPreimage),
+                Ok((incoming_contract_account, status)) => match status {
+                    DecryptedPreimageStatus::Pending => {
+                        // only when we are sure that the invoice is still pending that we can check
+                        // for a timeout
+                        const TOLERANCE: Duration = Duration::from_secs(60); // tolerate some clock skew
```
dpc, line 207: **praise:** I like TOLERANCE much more. Sounds very inclusive, while BUFFER sounded exclusive, so :+1: :smile: :+1:

dpc (approved):

fedimint-backports:
- #3939

elsirion:
On closer inspection this looks more like a quick fix. Since the underlying timeout issue was hard to debug we re-introduce polling again for the one API call that can take longer than some built-in timeout. It's the pragmatic thing to do, but we should track fixing this issue properly.

Issue #3665, opened 2023-11-20
Investigate why devimint data dir takes 1.5G after start


dpc:
```
> du -cksh $FM_DATA_DIR/..
1.5G    /tmp/devimint-3301602/cfg/..
1.5G    total
> du -cksh $FM_DATA_DIR/../esplora/
858M    /tmp/devimint-3301602/cfg/../esplora/
858M    total
> bitcoin-cli -conf=$FM_BTC_DIR/bitcoin.conf getblockcount
194
```

dpc:
How does esplora turn 194 empty regtest blocks into 858M? Some pre-allocation maybe?

dpc:
https://github.com/Blockstream/esplora/issues/471

dpc:
@m1sterc001guy Our own database acts exactly the same, but with smallerr value (70M == 64M?), while esplora is doing (260 per table = 256M?)

I suspect this is due to pre-allocation of rocksdb. Notably the size reported by `du` and `ls` are different (because this is using some pre-allocation API):

```
16:05:58 …regtest/newindex/cache
> du -cksh *
282M    000005.log
4.0K    CURRENT
4.0K    IDENTITY
0       LOCK
128K    LOG
4.0M    MANIFEST-000004
8.0K    OPTIONS-000007
286M    total
```

Unclear to me if having 100s of these directories does use / reserve my `/tmp/` disk space, but I think it does in the `df -h` at least.

We also should have a way to do decrease it for tests.

dpc:
https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_write_buffer_size ?

dpc:
https://github.com/Blockstream/electrs/blob/fd35014283c7d3a7a85c77b9fd647c9f09de12c9/src/new_index/db.rs#L92 ya.

PR #3932, opened 2023-12-13
Make devimint tmp dir much smaller

dpc:
Fix https://github.com/fedimint/fedimint/issues/3665

This shrinks each devimint tmp directory from 1.3GB to 83MB. This was caused be excessive default rocksdb pre-allocation, especially in esplora-electrs, which is not useful for tiny testing setups.

### DIFF
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -1,7 +1,8 @@
 #![allow(where_clauses_object_safety)] // https://github.com/dtolnay/async-trait/issues/228
 use std::path::Path;
+use std::str::FromStr;
 
-use anyhow::Result;
+use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
 use fedimint_core::db::{
     IDatabaseTransactionOps, IDatabaseTransactionOpsCore, IRawDatabase, IRawDatabaseTransaction,
@@ -10,6 +11,7 @@ use fedimint_core::db::{
 use futures::stream;
 pub use rocksdb;
 use rocksdb::{OptimisticTransactionDB, OptimisticTransactionOptions, WriteOptions};
+use tracing::debug;
 
 #[derive(Debug)]
 pub struct RocksDb(rocksdb::OptimisticTransactionDB);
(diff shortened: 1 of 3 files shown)


elsirion:
I think you linked the wrong issue xD

justinmoon (approved):

justinmoon:
> I think you linked the wrong issue xD

Edited to point at correct issue

bradleystachurski (approved):
Beautiful!

```
nix@lenovo-m75q:/tmp/nix-shell.JXa0PP/devimint-1234649$ du -sh
928M    .

nix@lenovo-m75q:/tmp/nix-shell.M81P2Y/devimint-1247514$ du -sh
87M     .
```

fedimint-rocksdb/src/lib.rs
```diff
@@ -28,14 +31,46 @@ impl RocksDb {
     }
 }
 
+fn is_power_of_two(num: usize) -> bool {
+    num != 0 && (num & (num - 1)) == 0
+}
```
elsirion, line 36: I think this would be easier to reason about if it used [`count_ones`](https://doc.rust-lang.org/stable/std/primitive.usize.html#method.count_ones):

```rust
num.count_ones() == 1
```

nix/esplora-electrs.nix
```diff
@@ -18,10 +18,17 @@ rustPlatform.buildRustPackage {
...
+    # owner = "Blockstream";
+    # repo = "electrs";
+    # rev = "adedee15f1fe460398a7045b292604df2161adc0";
+    # hash = "sha256-KnN5C7wFtDF10yxf+1dqIMUb8Q+UuCz4CMQrUFAChuA=";
+
+    # pre-allocation size patch:
+    owner = "dpc";
+    repo = "esplora-electrs";
+    rev = "8186331b7ca33668d838dab91e2dc52c388ac689";
+    hash = "sha256-D+ZdtZ57RoQsqebW0f2KsWz5/Di4Joy6walqGvGm/4o=";
```
elsirion, line 31: Once merged please open an issue to switch back once upstream releases a version with the patch.

elsirion (commented):
I really don't like that power of 2 check :grimacing: made me think way too hard about its correctness. Otherwise looking good.

fedimint-rocksdb/src/lib.rs
bradleystachurski: Agreed, it was a bit of a puzzle. The current impl seems to be a common bit twiddle hack, so if we don't modify to use `count_ones`, perhaps a comment with a reference would help our future selves.

https://stackoverflow.com/a/600306

dpc: I was rushing to wrap it and just googled / asked chatgpt and slapped some tests on top. :D

dpc:
https://github.com/romanz/electrs/pull/975

https://github.com/Blockstream/electrs/pull/64

elsirion (approved):

PR #3942, opened 2023-12-14
fix(mint-client): invalidate all spent notes on recovery

### LINKED ISSUE #3941: Recovery results in spents e-cash notes in wallet
This leads to transactions failing that should otherwise succeed.

Cause: recovery only invalidates notes as spent that it recovered itself, not ones that were in the backup file retrieved from the federation.

elsirion:
> Cause: recovery only invalidates notes as spent that it recovered itself, not ones that were in the backup file retrieved from the federation.

Fix #3941


okjodom (commented):
Lgtm

okjodom (approved):

PR #3953, opened 2023-12-15
fix: client detects too big transactions and rejects them

### LINKED ISSUE #3868: `BYTE_LIMIT` limits number of e-cash notes per transaction
After #3671 spending one e-cash note consumes approximately 156bytes (~92b input + 64b input signature) of transaction space. At the same time our instantiation of AlephBFT limits transactions to less than 10kb, which means even in the best case we cannot have more than 64 e-cash inputs, which severely limits the range of amounts we can encode with full precision.

As a quick fix #3867 increases said limit to 50kb, allowing us to include over 300 e-cash notes, exponentially increasing the range of amounts we can represent at full precision, making the limit a non-issue at the cost of a higher worst-case memory consumption.

Longer term we should reduce that limit again and make the client more intelligent to deal with such cases.

elsirion:
Fixes #3868. While the increase of `BYTE_LIMIT` prevents oversized transactions from being generated through standard usage patterns, there is still a risk of third party modules or hand crafted transactions exceeding the limit. This makes the client throw an error in these cases.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -74,7 +74,7 @@ use std::sync::atomic::AtomicUsize;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
 
-use anyhow::{anyhow, bail, Context};
+use anyhow::{anyhow, bail, ensure, Context};
 use async_stream::stream;
 use db::{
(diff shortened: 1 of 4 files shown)


dpc (approved):

douglaz (approved):
LGTM

PR #3955, opened 2023-12-15
fix(client): missing expiry time

douglaz:


dpc (approved):
Wooops.

dpc (commented):
Maybe I need more coffee but since it had a bug and I still has trouble reading maybe we should either add a test, or figure out some way to make this code obviously correct.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -205,8 +205,9 @@ impl LightningReceiveConfirmedInvoice {
                         // only when we are sure that the invoice is still pending that we can check
                         // for a timeout
                         const TOLERANCE: Duration = Duration::from_secs(60); // tolerate some clock skew
-                        let invoice_since_epoch = invoice.duration_since_epoch() + TOLERANCE;
-                        if now_since_epoch > invoice_since_epoch {
+                        let invoice_expiration_epoch =
+                            invoice.duration_since_epoch() + invoice.expiry_time() + TOLERANCE;
```
dpc, line 209: **thought:** So here we add "duration" + "time" + "tolerance (duration)", so what's confusing is that (I guess, `expiry_time` is "duration", while `duration_since_epoch` is actually "(absolute) time"). :D.

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -205,8 +205,9 @@ impl LightningReceiveConfirmedInvoice {
...
+                        if now_since_epoch > invoice_expiration_epoch {
```
dpc, line 210: **thought:** And now we call on3e thin `_since_epoch` and the other `_epoch` which also seems meh. :D

dpc:
I guess we'll want a quick backport and release of this, as is makes all ln payments have 60s deadlines?

elsirion (approved):

modules/fedimint-ln-client/src/receive.rs
dpc, line 209: **thought:** BTW. If this 3 lines were in a method, than I think we could have a really tiny unit test that crafts a fake invoice and then checks against values against and after the timeout.

fedimint-backports:
- #3957

PR #3956, opened 2023-12-15
feat: latency test for restore

Maan2003:
restore has become really slow, we need to track these.

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -205,18 +205,40 @@ pub async fn latency_tests(dev_fed: DevFed) -> Result<()> {
     let ln_sends_stats = stats_for(ln_sends);
     let ln_receives_stats = stats_for(ln_receives);
     let fm_pay_stats = stats_for(fm_internal_pay);
+
+    info!("Testing latency of restore");
+    let backup_secret = cmd!(fed, "print-secret").out_json().await?["secret"]
+        .as_str()
+        .map(ToOwned::to_owned)
+        .unwrap();
+    let restore_client = Client::create("restore").await?;
+    let start_time = Instant::now();
+    cmd!(
+        restore_client,
+        "restore",
+        "--mnemonic",
+        &backup_secret,
+        "--invite-code",
+        fed.invite_code()?
+    )
+    .run()
+    .await?;
+    let restore_time = start_time.elapsed();
+
     println!(
         "================= RESULTS ==================\n\
               REISSUE: {reissue_stats}\n\
               LN SEND: {ln_sends_stats}\n\
               LN RECV: {ln_receives_stats}\n\
-              FM PAY: {fm_pay_stats}"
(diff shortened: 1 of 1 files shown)


devimint/src/main.rs
```diff
@@ -175,16 +178,29 @@ pub async fn latency_tests(dev_fed: DevFed) -> Result<()> {
...
               LN SEND: {ln_sends_stats}\n\
-              LN RECV: {ln_receives_stats}"
+              LN RECV: {ln_receives_stats}\n\
+              RESTORE: {restore_time:?}"
     );
     // FIXME: should be smaller
     assert!(reissue_stats.median < Duration::from_secs(4));
     assert!(ln_sends_stats.median < Duration::from_secs(6));
     assert!(ln_receives_stats.median < Duration::from_secs(6));
+    assert!(restore_time < Duration::from_secs(160));
```
Maan2003, line 203: It was 83 on my machine, I multiplied with factor of 2 to account for CI slow down.

devimint/src/main.rs
```diff
@@ -175,16 +178,29 @@ pub async fn latency_tests(dev_fed: DevFed) -> Result<()> {
     let reissue_stats = stats_for(reissues);
     let ln_sends_stats = stats_for(ln_sends);
     let ln_receives_stats = stats_for(ln_receives);
+
+    info!("Testing latency of restore");
+    let backup_secret = cmd!(backup_client, "print-secret").out_json().await?["secret"]
+        .as_str()
+        .map(ToOwned::to_owned)
+        .unwrap();
+    cmd!(backup_client, "wipe", "--force",).run().await?;
```
dpc, line 187: **thought:** I though I removed this command already.

dpc:
During Alepbft refactor the prefetching of epochs was removed "for simplicity" and at the time I was suspecting it will lead to poor performance, so that would be my first bet.

Maan2003:
> @Maan2003 Are you on latest master?

I was testing on 0.2 branch. but now rebased to master.

elsirion:
There's also a random, constant slowdown: we wait for the session we started recovery in to end so we get all transactions we may have missed (since currently we don't have an API for fetching transactions from a running epoch)

elsirion:
@Maan2003 the restore invocation uses named arguments, that's why CI is failing.

devimint/src/main.rs
elsirion: This may or may not work as you could have been at an arbitrary point inside the current session. If you start restoring at the end of a session it will take shorter, if you start at the beginning it will take longer. https://github.com/fedimint/fedimint/pull/3956#issuecomment-1860148765

But let's just see how it does on master and narrow down the allowed time as we optimize restoring.

elsirion (approved):

elsirion:
Needs rebase :(

justinmoon (approved):

okjodom (approved):

PR #3959, opened 2023-12-15
fix: retry funding offer

Maan2003:


modules/fedimint-ln-client/src/incoming.rs
```diff
@@ -159,6 +159,7 @@ impl FundingOfferState {
     ) -> Result<(), IncomingSmError> {
         debug!("Awaiting funding success for outpoint: {out_point:?}");
         for retry in 0.. {
+            let sleep = retry * 15;
```
douglaz, line 162: Perhaps enforce a maximum value here?
Something like:
```suggestion
            let sleep = max(retry * 15, 90);
```

elsirion (approved):

justinmoon (approved):

PR #3963, opened 2023-12-16
fix(lightning): proper handle timeouts while receiving a payment

douglaz:
https://github.com/fedimint/fedimint/pull/3929 was a mess:

1) It required https://github.com/fedimint/fedimint/pull/3955

2) It still doesn't fully solve the timeout handling as it calls `context.wait_key_exists(ContractKey(contract_id))` which will block until a payment has been received.
In this situation the client will never timeout because a timeout will be considered an error and it will keep retrying.

3)  It still considers that a `DecryptedPreimage::Pending` is a good moment to timeout. But in fact if we got `DecryptedPreimage::Pending` the payment may still succeed, so we can't timeout here.

This PR tries to fix these issues. It also introduces tests for the expiration time check.

Note: this PR should be fully compatible with previous releases. We are just properly using an existing endpoint to solve the issue on the client-side.

### DIFF
--- a/modules/fedimint-ln-common/src/api.rs
+++ b/modules/fedimint-ln-common/src/api.rs
@@ -25,7 +25,10 @@ use crate::{ContractAccount, LightningGateway, LightningGatewayAnnouncement};
 #[apply(async_trait_maybe_send!)]
 pub trait LnFederationApi {
     async fn fetch_consensus_block_count(&self) -> FederationResult<Option<u64>>;
-    async fn fetch_contract(&self, contract: ContractId) -> FederationResult<ContractAccount>;
+    async fn fetch_contract(
+        &self,
+        contract: ContractId,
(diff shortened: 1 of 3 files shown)


modules/fedimint-ln-common/src/api.rs
```diff
@@ -75,7 +78,10 @@ where
         .await
     }
 
-    async fn fetch_contract(&self, contract: ContractId) -> FederationResult<ContractAccount> {
+    async fn fetch_contract(
+        &self,
+        contract: ContractId,
+    ) -> FederationResult<Option<ContractAccount>> {
```
elsirion, line 84: So previously decoding would fail here when `None` was returned?

elsirion (approved):
Looks correct, how important is backporting this? @douglaz

okjodom (approved):

fedimint-backports:
- #3965

PR #3698, opened 2023-11-23
refactor: remove in-built client in devimint fed

okjodom:
Removes in-built client from devimint `Federation`, and replaces it with means to create one or more clients connected to the devimint fed instance. This allows us to write devimint tests that need more than one client.

> motivation: need for #3695 and #3696

PR #3964, opened 2023-12-17
test: latency of internal payments within fed

### LINKED ISSUE #3695: Add latency test for internal swap
[latency tests](https://github.com/fedimint/fedimint/blob/4a6a580f4eeccc7f95baa8c693c2f5ae52900a24/devimint/src/main.rs#L27) should have coverage for internal swap between two clients in the same federation

okjodom:
Adds latency test coverage to transactions between two fedimint clients in a federation.
From initial tests, turns out these internal transactions are currently much slower than LN sends / receives!
A follow up here is to investigate and improve internal payments perf

```rs
REISSUE: min: 1.7s, avg: 1.8s, median: 1.8s, p90: 1.9s, max: 2.0s, sum: 54.1s
LN SEND: min: 2.9s, avg: 3.1s, median: 3.1s, p90: 3.2s, max: 4.4s, sum: 62.8s
LN RECV: min: 1.9s, avg: 2.2s, median: 2.2s, p90: 3.0s, max: 3.1s, sum: 44.1s

FM PAY: min: 4.2s, avg: 4.7s, median: 4.4s, p90: 5.0s, max: 9.4s, sum: 93.5s
```

closes #3695

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -227,16 +227,21 @@ impl Federation {
 
     pub async fn pegin(&self, amount_sats: u64) -> Result<()> {
         info!(amount_sats, "Peg-in");
-        let deposit = cmd!(self, "deposit-address").out_json().await?;
+        self.pegin_client(amount_sats, &self.client).await
+    }
+
+    pub async fn pegin_client(&self, amount: u64, client: &Client) -> Result<()> {
+        info!(amount, "Pegging-in client funds");
+        let deposit = cmd!(client, "deposit-address").out_json().await?;
         let deposit_address = deposit["address"].as_str().unwrap();
         let deposit_operation_id = deposit["operation_id"].as_str().unwrap();
 
         self.bitcoind
-            .send_to(deposit_address.to_owned(), amount_sats)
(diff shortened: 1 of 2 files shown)


devimint/src/federation.rs
```diff
@@ -242,6 +242,23 @@ impl Federation {
...
+        self.bitcoind
+            .send_to(deposit_address.to_owned(), amount)
+            .await?;
+        self.bitcoind.mine_blocks(21).await?;
+
+        cmd!(client, "await-deposit", deposit_operation_id)
+            .run()
+            .await?;
+        Ok(())
+    }
```
elsirion, line 260: The `pegin` function should be deduplicate with this one.

okjodom: Made `pegin` delegate to `pegin_client`. I should fully remove pegin in a rewrite of #3698 which is now quite stale

elsirion (approved):
Looks good :)

justinmoon:
Nice. @okjodom could you make an issue for the investigation?

justinmoon (approved):

PR #3969, opened 2023-12-18
fix: prevent spamming blockchain.info with requests

dpc:
In some circumstances, possibly triggered by a bug or unrelated failure it seems possible
for us to keep calling blockstream.info (or other electrs/esplora backend) from the autocommit loop.

I identified two orthogonal improvements:

* watch needs to be called only once, so calling it again on retries is a waste of time
* watch & get are two separate operations so should be treated as such

Each of them separately is effectively fixing the problem we witnessed, but both make sense, so I guess makes sense to land both.

### DIFF
--- a/modules/fedimint-wallet-client/src/deposit.rs
+++ b/modules/fedimint-wallet-client/src/deposit.rs
@@ -107,6 +107,13 @@ async fn await_created_btc_transaction_submitted(
         .script_pubkey();
     loop {
         match context.rpc.watch_script_history(&script).await {
+            Ok(_) => break,
+            Err(e) => warn!("Error while awaiting btc tx submitting: {e}"),
+        }
+        sleep(TRANSACTION_STATUS_FETCH_INTERVAL).await;
+    }
+    loop {
+        match context.rpc.get_script_history(&script).await {
             Ok(received) => {
                 // TODO: fix
                 if received.len() > 1 {
--- a/modules/fedimint-wallet-client/src/deposit.rs
+++ b/modules/fedimint-wallet-client/src/deposit.rs
@@ -1,3 +1,4 @@
+use std::cmp;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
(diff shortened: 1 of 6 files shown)


modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -398,10 +406,19 @@ impl WalletClientModule {
...
+                            .expect("locking failed")
+                            .contains(script)
+                        {
+                            // Begin watching the script address
+                            self.rpc.watch_script_history(script).await?;
+                            already_watched
+                                .lock()
+                                .expect("locking failed")
+                                .insert(script.clone());
+                        }
```
elsirion, line 421: How likely is this retry to happen? Should only occur if the `get_deposit_address` fn is called highly concurrently. So is the additional code complexity worth it?

elsirion (commented):
The second commit is an improvement, the first idk, a bit of complexity for not a whole lot of gain imo.

dpc:
I think what we've observed was actually caused be accumulating multiple deposit addresses. Created a separate issue in: https://github.com/fedimint/fedimint/issues/3970

modules/fedimint-wallet-client/src/lib.rs
dpc: I think I was wrong and it is very unlikely. Will drop it.

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -106,7 +107,19 @@ async fn await_created_btc_transaction_submitted(
         .tweak(&tweak.public_key(), &context.secp)
         .script_pubkey();
     loop {
-        match context.rpc.watch_script_history(&script).await {
+        if context.rpc.watch_script_history(&script).await.is_ok() {
+            break;
+        }
```
douglaz, line 112: log the error. it may be as debug!

douglaz (approved):

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -143,9 +150,9 @@ async fn await_created_btc_transaction_submitted(
                 warn!("Error fetching transaction history for {script:?}: {e}");
             }
         }
-
-        sleep(TRANSACTION_STATUS_FETCH_INTERVAL).await;
     }
+
+    unreachable!()
```
elsirion, line 155: Will be hit after about 100k years, sgtm.

elsirion (approved):

PR #3971, opened 2023-12-19
feat: added comment/description for lnurl invoices

douglaz:


douglaz:
@elsirion what are your thoughts on backporting small features like this one?

fedimint-cli/src/client.rs
```diff
@@ -101,6 +101,9 @@ pub enum ClientCmd {
         /// Amount to pay, used for lnurl
         #[clap(long)]
         amount: Option<Amount>,
+        /// Invoice comment/description, used on lnurl
+        #[clap(long)]
+        comment: Option<String>,
```
justinmoon, line 106: would it make sense to call this `lnurl_comment`? just to make it clearer that this is an lnurl-specific thing?

douglaz: Not sure it would cause confusion, but yeah, we can change that.

elsirion (approved):

justinmoon (approved):

PR #3975, opened 2023-12-19
feat: more server metrics

douglaz:


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -85,7 +85,19 @@ lazy_static! {
         "contracts::FundedContract::Outgoing"
     ))
     .unwrap();
-    static ref AMOUNTS_BUCKETS_SATS: Vec<f64> = vec![0.0, 0.5, 1.0, 1000.0];
+    static ref AMOUNTS_BUCKETS_SATS: Vec<f64> = vec![
+        0.0,
```
dpc, line 89: Isn't 0.0 useless?

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -85,7 +85,19 @@ lazy_static! {
...
+        0.5,
```
dpc, line 90: 0.1 for consistency? :D

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -505,12 +565,13 @@ impl ServerModule for Wallet {
         {
             return Err(WalletInputError::PegInAlreadyClaimed);
         }
-
+        let amount = fedimint_core::Amount::from_sats(input.tx_output().value);
+        let fee = self.cfg.consensus.fee_consensus.peg_in_abs;
+        WALLET_PEGIN_AMOUNTS_SATS.observe(amount.sats_f64());
+        WALLET_PEGIN_FEES_AMOUNTS_SATS.observe(fee.sats_f64());
+        WALLET_PEGIN.inc();
```
dpc, line 572: AFAIR histograms already track total count, so this one is not necessary.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -71,6 +75,58 @@ use secp256k1::{Message, Scalar};
 use strum::IntoEnumIterator;
 use tracing::{debug, error, info, instrument, trace, warn};
 
+lazy_static! {
+    static ref WALLET_PEGIN: IntCounter =
+        register_int_counter!(opts!("wallet_pegin", "Wallet peg-in")).unwrap();
```
dpc, line 80: + `_total`?

https://prometheus.io/docs/practices/naming/

modules/fedimint-ln-server/src/lib.rs
douglaz: It's one way to track what's zero.

modules/fedimint-ln-server/src/lib.rs
douglaz: nops

modules/fedimint-wallet-server/src/lib.rs
douglaz: image

modules/fedimint-wallet-server/src/lib.rs
douglaz: I'll just remove this

modules/fedimint-ln-server/src/lib.rs
douglaz: One example where the zero makes it clear:
```
mint_redeemed_ecash_sats_bucket{le="0"} 0
mint_redeemed_ecash_sats_bucket{le="0.5"} 678
mint_redeemed_ecash_sats_bucket{le="1"} 722
```

```
mint_redeemed_ecash_fees_sats_bucket{le="0"} 756
mint_redeemed_ecash_fees_sats_bucket{le="0.5"} 756
mint_redeemed_ecash_fees_sats_bucket{le="1"} 756
```

dpc: I don't get it. We don't even have fees for ecash txes, do we? And peg-in/peg-outs must have non-zero fees?

modules/fedimint-ln-server/src/lib.rs
dpc: I don't get it. Is there something specially meaningful about 0.5 (500msat)? :D

BTW. I would consider `0.1 0.5 1 5 10 50` anyway. 10x granularity is kind of a big step. Typically I'd see powers of 2, or `1, 2, 5, 10` for stuff like that.

modules/fedimint-ln-server/src/lib.rs
douglaz: It seems we do have fees for ecash transactions.

modules/fedimint-ln-server/src/lib.rs
douglaz: I chose 0.5 a long time ago, but you're right, 0.1 is also fine for these purposes

modules/fedimint-ln-server/src/lib.rs
dpc: > It seems we do have fees for ecash transactions

WAT? How? Why I don't know about it.

douglaz: https://github.com/fedimint/fedimint/blob/b013367ba68474c89e213d439a8c747d31acddf5/modules/fedimint-mint-server/src/lib.rs#L351

dpc: TIL.

dpc (dismissed):

modules/fedimint-ln-server/src/lib.rs
elsirion: Yep, every module *could* charge fees in theory. In practice we don't do it since it would massively complicate the client (especially for e-cash). It would be easy to get scenarios where coin selection runs into loops where adding/removing e-cash makes adding/removing more e-cash necessary because of fees.

Note: these fees are a separate concept from withdraw or LN fees, they'd be going to guardians (well, technically the federation would just become overcollateralized since there is no way to withdraw them).

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -47,6 +48,52 @@ use tbs::{
...
+        "Value of redeemed e-cash notes in sats",
+        AMOUNTS_BUCKETS_SATS.clone()
+    ))
+    .unwrap();
+    static ref MINT_REDEEMED_ECASH_FEES_SATS: Histogram = register_histogram!(histogram_opts!(
+        "mint_redeemed_ecash_fees_sats",
+        "Value of e-cash fees during reissue in sats",
+        AMOUNTS_BUCKETS_SATS.clone()
+    ))
+    .unwrap();
```
elsirion, line 76: I guess this is nice to have for future fee support, but not really useful rn.

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -375,11 +427,11 @@ impl ServerModule for Mint {
...
-
-        Ok(TransactionItemAmount {
-            amount: output.amount,
-            fee: self.cfg.consensus.fee_consensus.note_issuance_abs,
-        })
+        let amount = output.amount;
+        let fee = self.cfg.consensus.fee_consensus.note_issuance_abs;
+        MINT_ISSUED_ECASH_SATS.observe(amount.sats_f64());
+        MINT_ISSUED_ECASH_FEES_SATS.observe(fee.sats_f64());
+        Ok(TransactionItemAmount { amount, fee })
```
elsirion, line 434: I'm not super happy with having this code in here between the consensus logic. Maybe it could be a separate function `fn observe_mint_input(TransactionItemAmount)`.

elsirion (changes requested):
It just occurred to me that `process_input/output` is called twice for every input/output: once to verify the tx on submission and another time to process after consensus has been achieved. This would lead to double-counting of most metrics in this PR.

A potential solution for that might be registering [`on_commit` hooks](https://docs.rs/fedimint-core/latest/fedimint_core/db/struct.DatabaseTransaction.html#method.on_commit) with the database transaction which only run if the fedimint tx is processed successfully.

Do we already have metrics for accepted vs. rejected transactions btw?

dpc:
> It just occurred to me that `process_input/output` is called twice for every input/output: once to verify the tx on submission and another time to process after consensus has been achieved.

Would it make sense to add another method to the trait for "this thing was already accepted". We could make it have a default no-op impl, so it wouldn't even break anything.

modules/fedimint-ln-server/src/lib.rs
dpc: > It would be easy to get scenarios where coin selection runs into loops where adding/removing e-cash makes adding/removing more e-cash necessary because of fees.

I don't think hitting this would be common. It is exactly like in Bitcoin - not really a problem people complain about. A bit of looping, and tiny overpaying fees should do.

IMO, the bigger issue the whole new aspect of "guardians making money". We have no affordances for it, no means for them to do anything useful with it (withdraw or e.g. contribute to peg-outs, or something), etc.

elsirion:
> Would it make sense to add another method to the trait for "this thing was already accepted"

That would be a lot of API surface only for metrics :confused: this would need to be implemented for inputs+outputs in `ServerModule` + `IServerModule`.

elsirion (approved):
Looks good to me now, but we should get into the habit of testing our PRs. The double-counting would likely have been caught by a test.

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1163,6 +1164,36 @@ impl Lightning {
     }
 }
 
+fn calculate_funded_contract_metrics(
+    updated_contract_account: ContractAccount,
+    dbtx: &mut DatabaseTransaction<'_>,
+) {
+    dbtx.on_commit(move || match updated_contract_account.contract {
```
dpc, line 1171: Oh, I forgot we have this. So good. Very smart. Much :brain: .

dpc (approved):

dpc:
> he double-counting would likely have been caught by a test.

Bugs in metrics are kind of whateva, :D IMO, and they are a PITA to test because they use global counter.

fedimint-backports:

```bash
git worktree add -d .worktree/backport-3975-to-releases/v0.2 origin/releases/v0.2
cd .worktree/backport-3975-to-releases/v0.2
git switch --create backport-3975-to-releases/v0.2
git cherry-pick -x 62521575788147153abdd8248de417de74d5ee7a
```

Issue #3962, opened 2023-12-16
Devimint backwards-compatibility tests


bradleystachurski:
We're gonna need perhaps a devimint feature where you can replace the client or the server or any component with a different version (older or newer) then run all the tests and see if they pass.

_Originally posted by @douglaz in https://github.com/fedimint/fedimint/discussions/3950#discussioncomment-7866528_

bradleystachurski:
Might be useful to reference bitcoin core's approach for backwards compatibility tests.

https://github.com/bitcoin/bitcoin/blob/3695ecbf680a66b718f97d504308578d001eec49/test/functional/wallet_backwards_compatibility.py

bradleystachurski:
> run all the tests and see if they pass

What does "all" tests mean in this context? Is this all devimint tests or everything run by `just test-ci-all`? If this includes all the backend tests, we'll also need to consider how we'll use dynamic versions for clients and servers in `modules/fedimint-*-tests` instead of just devimint.

bradleystachurski:
I haven't fully considered the implications, but one potential approach if we need to introduce clients/servers with previous versions is to reference the git tag.

```
fedimint-client-old = { git = "https://github.com/fedimint/fedimint", tag = "v0.2.0", package = "fedimint-client" }
```

This is hacky and doesn't scale well if we need to test against multiple versions, so there's likely a better solution.

elsirion:
I think devimint should be able to instantiate different versions of federations. The binaries (`fedimintd`, `fedimint-cli`, …) should come from nix imo since then we can use caching more effectively. @dpc probably knows best how to do this.

bradleystachurski:
Now that the initial approach in CI is merged and all broken tests are resolved, are we comfortable closing? I think we can track future improvements in separate issues.

cc: @elsirion @dpc

Issue #3977, opened 2023-12-20
Master has broken the compatibility (gateway `/info` now is a GET instead of a POST)


elsirion:
@elsirion master has already broken the compatibility (gateway `/info` now is a GET instead of a POST)

_Originally posted by @douglaz in https://github.com/fedimint/fedimint/discussions/3950#discussioncomment-7892026_

PR #3982, opened 2023-12-21
feat: overridable binaries and cross test

douglaz:
Let all binaries from devimint be replaceable through environment variables.

Added a "cross test" as example where we run the master tests against `gatewayd` [v0.2](https://github.com/fedimint/fedimint/tree/releases/v0.2) reproducing https://github.com/fedimint/fedimint/issues/3977 (so the cross test will correctly fail right now)

What we do is just override the `gatewayd` executable with:
```bash
export FM_GATEWAYD_BASE_EXECUTABLE="nix run git+https://github.com/fedimint/fedimint.git?ref=releases/v0.2#gatewayd --"
```

Then if we run `./scripts/tests/test-ci-all.sh` the `gatewayd` binary will come from `releases/v0.2` and the other binaries will come normally from the current repository compilation. Any mix and match is possible.

Also this approach (or something very similar to it) makes it possible to test migrations (for instance start an old fedimintd version, stop it, then upgrade to current master during a test and see if everything is still working)

Will eventually close https://github.com/fedimint/fedimint/issues/3962

Opening now as draft for feedback
@elsirion @dpc @bradleystachurski

Note: CI isn't working right now but you can test locally with `./scripts/tests/cross-test-ci-all.sh`

### DIFF
--- a/nix/flakebox.nix
+++ b/nix/flakebox.nix
@@ -376,6 +376,15 @@ rec {
     '';
   };
 
+  # ciCrossTestAll = craneLibTests.mkCargoDerivation {
+  #   pname = "${commonCliTestArgs.pname}-all-cross";
+  #   cargoArtifacts = workspaceBuild;
+  #   buildPhaseCargoCommand = ''
+  #     patchShebangs ./scripts
+  #     ./scripts/tests/cross-test-ci-all.sh || exit 1
+  #   '';
+  # };
+
   alwaysFailTest = craneLibTests.mkCargoDerivation {
     pname = "${commonCliTestArgs.pname}-always-fail";
     cargoArtifacts = workspaceBuild;
--- a/scripts/tests/cross-test-ci-all.sh
+++ b/scripts/tests/cross-test-ci-all.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+set -x
+
+# Test current versions against gatewayd v0.2
+export FM_GATEWAYD_BASE_EXECUTABLE="nix run git+https://github.com/fedimint/fedimint.git?ref=releases/v0.2#gatewayd --"
+# shellcheck disable=SC2260
+$FM_GATEWAYD_BASE_EXECUTABLE --help || true
+./scripts/tests/test-ci-all.sh
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -126,6 +126,12 @@ jobs:
         if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests && (matrix.host != 'macos')
         run: nix build -L .#wasm32-unknown.ci.wasmTest --keep-failed
 
+
+      - name: Cross Tests
+        if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests
+        run: |
+          ./scripts/tests/cross-test-ci-all.sh
+
       - name: Prepare failed test build dirs
         if: always()
         run:  |
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -43,7 +43,10 @@ impl Bitcoind {
         );
         write_overwrite_async(processmgr.globals.FM_BTC_DIR.join("bitcoin.conf"), conf).await?;
         let process = processmgr
-            .spawn_daemon("bitcoind", cmd!("bitcoind", "-datadir={btc_dir}"))
+            .spawn_daemon(
+                "bitcoind",
+                cmd!(crate::util::Bitcoind, "-datadir={btc_dir}"),
+            )
             .await?;
 
         let url = processmgr.globals.FM_BITCOIN_RPC_URL.parse()?;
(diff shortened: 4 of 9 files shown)


dpc:
I don't think we should be shelling-out and parse these commands as arrays. It's extra complexity for little benefit, and might lead to usage problems and confusion. If someone wants to call a shell-script, they can just write it themselves.

scripts/tests/cross-test-ci-all.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+set -x
+
+# Test current versions against gatewayd v0.2
+export FM_GATEWAYD_BASE_EXECUTABLE="nix run git+https://github.com/fedimint/fedimint.git?ref=releases/v0.2#gatewayd --"
```
dpc, line 7: There are couple of problems with this:

* some extra stuff gets printed on stderr which can confuse some tools and pollute the logs
* the slowdown happens somewhere deep inside so it might be hard to tell if the whole thing hanged or just building

A:

```
>&2 echo "Compiling old binaries..."
nix build git+https://github.com/fedimint/fedimint.git?ref=releases/v0.2#gatewayd
PATH="$(readlink result)/bin/:$PATH"
```

should achieve the same thing without the downsides, and also doesn't require parsing shell commands as arguments in devimint.

dpc: @bradleystachurski We've discussed adding old releases to inputs, but I guess the above approach has a nice property that it's on-demand, not requiring building old binaries upfront, so no need for extra flake inputs and/or separate dev shells.

dpc: BTW `nix build 'github:fedimint/fedimint?branch=releases/0.2#gatewayd'` is shorter

bradleystachurski: Since the `releases/v0.2` branch merges backport PRs, it's a bit misleading. Right now `relesaes/v0.2` (https://github.com/fedimint/fedimint/commit/51d09c76b8a15615114d4f2c2663c738d8ff3135) is using [v0.2.1-rc2](https://github.com/fedimint/fedimint/releases/tag/v0.2.1-rc2) as the base.

I think it's more clear to use the explicit tag we want as a base and also allows for a shorter nix command
```
nix build 'github:fedimint/fedimint/v0.2.0#gatewayd'
```

justinmoon: Seems to me we should just override binaries by writing to the PATH

bradleystachurski: @dpc re: https://github.com/fedimint/fedimint/pull/3982/files#r1433538187

How would we handle the binaries with a same name and different version using this approach?

Would it make sense to export the path for the specific binary instead of prepending to the PATH?

```
nix build 'github:fedimint/fedimint/v0.2.0#gatewayd'
export FM_GATEWAYD_BASE_EXECUTABLE="$(readlink result)/bin/gatewayd"

File: nix/flakebox.nix [hunk 2]
```diff
@@ -376,6 +376,15 @@ rec {
     '';
   };
 
+  # ciCrossTestAll = craneLibTests.mkCargoDerivation {
+  #   pname = "${commonCliTestArgs.pname}-all-cross";
```
bradleystachurski, line 380: Taking a wild guess since this is commented out.

If you're bumping into issues running `nix build .#ci.ciCrossTestAll -L` due to cln socket errors, it's because the socket path > 108 chars when running locally.

https://unix.stackexchange.com/a/367012

I haven't thought of an elegant way to handle. I'll create an issue to track.

scripts/tests/cross-test-ci-all.sh
elsirion: Instead of `readlink` we could use `nix build .#… --print-out-paths`

dpc: You all all right! :)

scripts/tests/cross-test-ci-all.sh
```diff
@@ -0,0 +1,10 @@
...
+# shellcheck disable=SC2260
+$FM_GATEWAYD_BASE_EXECUTABLE --help || true
+./scripts/tests/test-ci-all.sh
```
bradleystachurski, line 10: You mentioned in the PR description this script is just an example, but worth highlighting if we use `test-ci-all.sh` in a backwards compatibility run the backend tests will still use clients/servers on the same version.

elsirion:
This is probably the most important PR unless severe bugs are found, otherwise we won't be able to keep developing with confidence. We should discuss what kind of test matrix we want, my best guess is:

* `{old, new} federation x {old, new} client2 x {old, new} gateway` = 8 cases
* `new federation x old client x new client` = 1 case
* `new federation x old gateway x new gateway` = 1 case

I don't know how long this will run, we might want to gate it with a label so we only run the full test suite once we think a PR is merge-ready.

bradleystachurski:
dev call: @douglaz since you'll be afk for the next few weeks I'll pickup this PR. Thanks for putting up a great initial approach!

bradleystachurski:
Closing since this is handled in https://github.com/fedimint/fedimint/pull/4010

Issue #3988, opened 2023-12-22
Client is less parallelism-safe than expected


elsirion:
Reproducible using this test in `modules/fedimint-mint-tests/tests/tests.rs`:

```rust
#[tokio::test(flavor = "multi_thread")]
async fn sends_ecash_oob_highly_parallel() -> anyhow::Result<()> {
    // Print notes for client1
    let fed = fixtures().new_fed().await;
    let (client1, client2) = fed.two_clients().await;
    let client1_dummy_module = client1.get_first_module::<DummyClientModule>();
    let (op, outpoint) = client1_dummy_module.print_money(sats(1000)).await?;
    client1.await_primary_module_output(op, outpoint).await?;

    let num_par = 2;

    // Spend from client1 to client2 25 times in parallel
    let mut spend_tasks = vec![];
    for num_spend in 0..num_par {
        let task_client1 = client1.clone();
        spend_tasks.push(tokio::spawn(async move {
            info!("Starting spend {num_spend}");
            let client1_mint = task_client1.get_first_module::<MintClientModule>();
            let (op, notes) = client1_mint
                .spend_notes(sats(30), TIMEOUT, ())
                .await
                .unwrap();
            let sub1 = &mut client1_mint
                .subscribe_spend_notes(op)
                .await
                .unwrap()
                .into_stream();
            assert_eq!(sub1.ok().await.unwrap(), SpendOOBState::Created);
            notes
        }));
    }

    let note_bags = futures::stream::iter(spend_tasks)
        .then(|handle| async move { handle.await.unwrap() })
        .collect::<Vec<_>>()
        .await;

    let mut reissue_tasks = vec![];
    for (num_reissue, notes) in note_bags.into_iter().enumerate() {
        let task_client2 = client2.clone();
        reissue_tasks.push(tokio::spawn(async move {
            info!("Starting reissue {num_reissue}");
            let client2_mint = task_client2.get_first_module::<MintClientModule>();
            let op = client2_mint
                .reissue_external_notes(notes, ())
                .await
                .unwrap();
            let sub2 = client2_mint
                .subscribe_reissue_external_notes(op)
                .await
                .unwrap();
            let mut sub2 = sub2.into_stream();
            assert_eq!(sub2.ok().await.unwrap(), ReissueExternalNotesState::Created);
            assert_eq!(sub2.ok().await.unwrap(), ReissueExternalNotesState::Issuing);
[...]

PR #3989, opened 2023-12-22
fix: handle write write conflicts in mem db

Maan2003:
related https://github.com/fedimint/fedimint/issues/3988

### DIFF
--- a/fedimint-core/src/db/mem_impl.rs
+++ b/fedimint-core/src/db/mem_impl.rs
@@ -1,11 +1,11 @@
 use std::collections::BTreeMap;
 use std::fmt::Debug;
-use std::sync::Mutex;
 
 use anyhow::Result;
 use bitcoin_hashes::hex::ToHex;
 use futures::{stream, StreamExt};
 use macro_rules_attribute::apply;
+use tokio::sync::Mutex;
 
 use super::{
     IDatabaseTransactionOps, IDatabaseTransactionOpsCore, IRawDatabase, IRawDatabaseTransaction,
@@ -17,11 +17,13 @@ use crate::db::PrefixStream;
 pub struct DatabaseInsertOperation {
     pub key: Vec<u8>,
     pub value: Vec<u8>,
+    pub old_value: Option<Vec<u8>>,
 }
 
 #[derive(Debug, Default)]
 pub struct DatabaseDeleteOperation {
     pub key: Vec<u8>,
+    pub old_value: Option<Vec<u8>>,
 }
 
 #[derive(Debug)]
@@ -53,8 +55,8 @@ impl MemDatabase {
         Default::default()
     }
 
-    pub fn dump_db(&self) {
-        let data = self.data.lock().unwrap();
+    pub async fn dump_db(&self) {
+        let data = self.data.lock().await;
         let data_iter = data.iter();
         for (key, value) in data_iter {
             println!("{}: {}", key.to_hex(), value.to_hex());
@@ -66,7 +68,7 @@ impl MemDatabase {
 impl IRawDatabase for MemDatabase {
     type Transaction<'a> = MemTransaction<'a>;
     async fn begin_transaction<'a>(&'a self) -> MemTransaction<'a> {
-        let db_copy = self.data.lock().unwrap().clone();
+        let db_copy = self.data.lock().await.clone();
         let mut memtx = MemTransaction {
             operations: Vec::new(),
             tx_data: db_copy.clone(),
@@ -86,16 +88,16 @@ impl IRawDatabase for MemDatabase {
 #[apply(async_trait_maybe_send!)]
 impl<'a> IDatabaseTransactionOpsCore for MemTransaction<'a> {
     async fn raw_insert_bytes(&mut self, key: &[u8], value: &[u8]) -> Result<Option<Vec<u8>>> {
-        let val = self.raw_get_bytes(key).await;
         // Insert data from copy so we can read our own writes
-        self.tx_data.insert(key.to_vec(), value.to_owned());
+        let old_value = self.tx_data.insert(key.to_vec(), value.to_owned());
         self.operations
             .push(DatabaseOperation::Insert(DatabaseInsertOperation {
                 key: key.to_vec(),
                 value: value.to_owned(),
+                old_value: old_value.clone(),
             }));
         self.num_pending_operations += 1;
-        val
+        Ok(old_value)
     }
 
     async fn raw_get_bytes(&mut self, key: &[u8]) -> Result<Option<Vec<u8>>> {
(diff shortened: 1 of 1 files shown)


Maan2003:
still broken :/

test deadlocks now

fedimint-core/src/db/mem_impl.rs
```diff
@@ -176,21 +180,25 @@ impl<'a> IDatabaseTransactionOps for MemTransaction<'a> {
 #[apply(async_trait_maybe_send!)]
 impl<'a> IRawDatabaseTransaction for MemTransaction<'a> {
     async fn commit_tx(self) -> Result<()> {
+        let mut data = self.db.data.lock().await;
```
dpc, line 183: If you move the lock all the way to the top, what's the point of comparing old values. Nothing could have changed them. It's the read-write conflicts that might break snapshot level isolation.

Maan2003: imagine 
tx a: delete(x)
tx b: delete(x)

if we don't compare old values, then commit both transactions will succeed.

fedimint-core/src/db/mem_impl.rs
```diff
@@ -66,7 +68,7 @@ impl MemDatabase {
 impl IRawDatabase for MemDatabase {
     type Transaction<'a> = MemTransaction<'a>;
     async fn begin_transaction<'a>(&'a self) -> MemTransaction<'a> {
-        let db_copy = self.data.lock().unwrap().clone();
+        let db_copy = self.data.lock().await.clone();
```
dpc, line 71: OK, I see I don't understand how things work.

m1sterc001guy:
Without looking too much at the details here, are we sure we want to implement this into our memdb? We're essentially implementing MVCC snapshot isolation into a database that is maintained only for Fedimint. Just wondering if its more implementation and maintenance effort than it is worth.

Maan2003:
> Without looking too much at the details here, are we sure we want to implement this into our memdb? We're essentially implementing MVCC snapshot isolation into a database that is maintained only for Fedimint. Just wondering if its more implementation and maintenance effort than it is worth.

memdb is pretty much only thing we can use from wasm

douglaz:
I was playing with this PR in the last couple of days. I still don't know what's the issue but I was able to reproduce it with a very small PR: https://github.com/fedimint/fedimint/pull/3995

elsirion:
I think we can land this PR with a different test suite that tests write-write conflicts. @m1sterc001guy already wrote these tests, e.g. `fedimint_core::db::expect_write_conflict`, so that should be quick. The deadlock issue can be debugged separately.

dpc (approved):
LGTM, but we should just use a rocksdb db with a random path, instead of spending time on home-made MVCC. :P

Maan2003:
we still have to use memdb partially on wasm

elsirion:
I restarted CI, either the correct impl triggers some flakiness more effectively or the PR has a subtle bug.

fedimint-core/src/db/mem_impl.rs
```diff
@@ -265,6 +273,11 @@ mod tests {
         fedimint_core::db::verify_remove_by_prefix(database()).await;
     }
 
+    #[test_log::test(tokio::test)]
+    async fn test_expect_write_conflict() {
+        fedimint_core::db::expect_write_conflict(database()).await;
+    }
+
```
elsirion, line 280: Is this the only test we can add? We should compare to the rocksdb impl that has the full test suite.

Maan2003:
checking the old values makes commit_tx() slow enough, now we are able to reproduce same bug as https://github.com/fedimint/fedimint/pull/3995

same test `sends_ecash_out_of_band` is failing

it just gets stuck on request timeouts in the end.

dpc:
It probably can't be done because of something that I'm not aware, but I wish we just wrap whole db into mutex when using memdb instead of trying to grow and maintain our database implementation. The difference in performance is kind of irrelevant.

Maan2003:
someone holding to database transaction and not commit will block others forever

Maan2003:
rebased after #4116

dpc (approved):
Oh well.

fedimint-core/src/db/mem_impl.rs
```diff
@@ -86,16 +88,17 @@ impl IRawDatabase for MemDatabase {
 #[apply(async_trait_maybe_send!)]
 impl<'a> IDatabaseTransactionOpsCore for MemTransaction<'a> {
     async fn raw_insert_bytes(&mut self, key: &[u8], value: &[u8]) -> Result<Option<Vec<u8>>> {
-        let val = self.raw_get_bytes(key).await;
+        let old_value = self.tx_data.get(key).cloned();
         // Insert data from copy so we can read our own writes
         self.tx_data.insert(key.to_vec(), value.to_owned());
```
douglaz, line 93:
```suggestion
        let old_value = self.tx_data.insert(key.to_vec(), value.to_owned());
```

dpc: :brain:

fedimint-core/src/db/mem_impl.rs
```diff
@@ -176,21 +180,25 @@ impl<'a> IDatabaseTransactionOps for MemTransaction<'a> {
...
         for op in self.operations {
             match op {
                 DatabaseOperation::Insert(insert_op) => {
-                    self.db
-                        .data
-                        .lock()
-                        .unwrap()
-                        .insert(insert_op.key, insert_op.value);
+                    anyhow::ensure!(
+                        data_copy.insert(insert_op.key, insert_op.value) == insert_op.old_value,
```
elsirion, line 189: It just occurred to me that this check is not bullet-proof, let's imaging an entry `x=0` at the beginning, we now run the following operations:

* Start dbtx `A`, where `x=0`
* `A` writes `x=1`
* Start dbtx `B` with `x=0`
* `B` writest `x=2`
* `B` commits successfully, `x=2` now
* Start dbtx `C` with `x=2`
* `C` writes `x=0`
* `C` commits successfully, `x=0` now
* `A` commits, should fail due to a write/write conflict, but succeeds with this code

I don't think this is too problematic for how we use the DB, but it might come to bite us in the ass eventually. I think to do this properly entries need version numbers as I described here https://github.com/elsirion/webimint-rs/issues/31. Thoughts @m1sterc001guy?

fedimint-core/src/db/mem_impl.rs
elsirion: Still relevant ^

elsirion (commented):
I think we could merge this as an incremental improvement with @douglaz change applied.

fedimint-core/src/db/mem_impl.rs
```diff
@@ -176,21 +180,25 @@ impl<'a> IDatabaseTransactionOps for MemTransaction<'a> {
 #[apply(async_trait_maybe_send!)]
 impl<'a> IRawDatabaseTransaction for MemTransaction<'a> {
     async fn commit_tx(self) -> Result<()> {
+        let mut data = self.db.data.lock().await;
+        let mut data_copy = data.clone();
```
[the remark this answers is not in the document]
Maan2003: how? we have to rollback in case check fails bellow.

fedimint-core/src/db/mem_impl.rs
Maan2003: yes, we have same tests as rocksdb now

except testing_desecending order but that method is not implemented by memdb

fedimint-core/src/db/mem_impl.rs
dpc: BTW. I recommend https://www.amazon.com/Database-Internals-deep-dive-distributed-systems/dp/1492040347/ . Very good book to read, since we're already implementing our own databases. :D

Thought even in it they don't dive to deep into implementing MVCC...

elsirion (approved):

dpc (approved):

PR #3995, opened 2023-12-29
Reproduce deadlock

douglaz:
This is an attempt to reproduce the deadlock issue arised on https://github.com/fedimint/fedimint/pull/3989

We can see it's possible to reproduce the issue with 2 small changes from master:
- Changing the test DB from `MemDatabase` to `RocksDb`
- Adding a `task::yield_now()` to `commit_tx()`

I also increased the timeout of the connection requests just to avoid noise in the logs due to the low 60s default timeout.

One way to see a test deadlocking for debug purposes is to run:
```bash
RUST_LOG=trace,hyper=info,soketto=info,fedimint_server::net::framed=debug cargo test --package fedimint-mint-tests --test fedimint_mint_tests -- sends_ecash_out_of_band --exact --nocapture
```

The goal of this PR is just to show that there is some issue on Fedimint db layer that isn't exclusive to `MemDatabase` and that can be reproduced without adding new tests. Created an issue to track it: https://github.com/fedimint/fedimint/pull/3995

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -804,6 +804,7 @@ impl JsonRpcClient for WsClient {
         return WsClientBuilder::default()
             .use_webpki_rustls()
             .max_concurrent_requests(u16::MAX as usize)
+            .request_timeout(Duration::from_secs(3600))
             .build(url.as_str())
             .await;
 
--- a/fedimint-testing/src/federation.rs
+++ b/fedimint-testing/src/federation.rs
@@ -10,7 +10,6 @@ use fedimint_core::config::{
     META_FEDERATION_NAME_KEY,
 };
 use fedimint_core::core::ModuleInstanceId;
-use fedimint_core::db::mem_impl::MemDatabase;
 use fedimint_core::db::Database;
 use fedimint_core::module::ApiAuth;
 use fedimint_core::task::TaskGroup;
@@ -54,7 +53,9 @@ impl FederationTest {
     pub async fn new_client_with_config(&self, client_config: ClientConfig) -> ClientArc {
         info!(target: LOG_TEST, "Setting new client with config");
         let mut client_builder = Client::builder(fedimint_client::DatabaseSource::Fresh(
-            MemDatabase::new().into(),
+            fedimint_rocksdb::RocksDb::open(tempfile::tempdir().unwrap())
+                .unwrap()
+                .into(),
         ));
         client_builder.with_module_inits(self.client_init.clone());
         client_builder.with_primary_module(self.primary_client);
@@ -109,7 +110,10 @@ impl FederationTest {
 
             let instances = config.consensus.iter_module_instances();
             let decoders = server_init.available_decoders(instances).unwrap();
-            let db = Database::new(MemDatabase::new(), decoders);
+            let db = Database::new(
+                fedimint_rocksdb::RocksDb::open(tempfile::tempdir().unwrap()).unwrap(),
+                decoders,
+            );
 
             let (consensus_server, consensus_api) = ConsensusServer::new_with(
                 config.clone(),
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -269,6 +269,7 @@ impl<'a> IDatabaseTransactionOps for RocksDbTransaction<'a> {
 #[async_trait]
 impl<'a> IRawDatabaseTransaction for RocksDbTransaction<'a> {
     async fn commit_tx(self) -> Result<()> {
+        tokio::task::yield_now().await;
         fedimint_core::task::block_in_place(|| {
             self.0.commit()?;
             Ok(())
--- a/fedimint-server/src/lib.rs
+++ b/fedimint-server/src/lib.rs
@@ -53,7 +53,7 @@ pub mod config;
 pub mod multiplexed;
 
 /// How long to wait before timing out client connections
-const API_ENDPOINT_TIMEOUT: Duration = Duration::from_secs(60);
+const API_ENDPOINT_TIMEOUT: Duration = Duration::from_secs(3600);
 
 /// Has the context necessary for serving API endpoints
 ///


fedimint-core/src/api.rs
```diff
@@ -804,6 +804,7 @@ impl JsonRpcClient for WsClient {
         return WsClientBuilder::default()
             .use_webpki_rustls()
             .max_concurrent_requests(u16::MAX as usize)
+            .request_timeout(Duration::from_secs(3600))
```
Maan2003, line 807: maybe this is just making it really slow?

fedimint-testing/src/federation.rs
```diff
@@ -109,7 +110,10 @@ impl FederationTest {
 
             let instances = config.consensus.iter_module_instances();
             let decoders = server_init.available_decoders(instances).unwrap();
-            let db = Database::new(MemDatabase::new(), decoders);
+            let db = Database::new(
+                fedimint_rocksdb::RocksDb::open(tempfile::tempdir().unwrap()).unwrap(),
+                decoders,
+            );
```
Maan2003, line 116: can't repro with only client db changed (with server memdb)
and I thought this was client side deadlock, maybe it is server side?

Maan2003:
this could be separate deadlock not same as #3989

dpc:
This looks like `MintClientModule::await_output_finalized` subscribing to notifications, filtering out some `Created` and never getting any updates. Possibly a timing/race condition issue somewhere, not a db problem.

dpc:
I've pin-pointed the problem the that AlephBFT hanging:

```
impl std::io::Write for UnitSaver {
    fn write(&mut self, buffer: &[u8]) -> std::io::Result<usize> {
        self.buffer.extend_from_slice(buffer);
        Ok(buffer.len())
    }

    fn flush(&mut self) -> std::io::Result<()> {
        info!(units_index = %self.units_index, "##### UNIT FLUSH START");
        block_on(async {
                let mut dbtx = self.db.begin_transaction().await;

                dbtx.insert_new_entry(&AlephUnitsKey(self.units_index), &self.buffer)
                    .await;

                dbtx.commit_tx_result()
                    .await
        })

        info!(units_index = %self.units_index, "##### UNIT FLUSH END");

        self.buffer.clear();
        self.units_index += 1;

        Ok(())
    }
}
```

It is using `futures::executor::block_on;` which in my understanding corresponds to some kind of loop polling future repeatedly on the current thread, and I guess this is somehow incompatible with `tokio::task::yield_now` which I guess expects to be running on tokio runtime direclty(?).

I'm pretty sure this `block_on` is incorrect, but happens to work.  I tried couple of quick attempts to fix it using different `block_on`s and `block_in_place` but I only made things even worse, so I guess this requires some more debugging and careful analysis, but this made me remind myself that async Rust is a waste of everyone's time and ruined otherwise perfect Rust programming language and should be avoided.


Anyway - I think it makes sense, the only thing that this PR does is add `yield_now`, this blocks aleph, so the submitted transaction is never processed and the test hangs.

The original purpose of this PR was to show that there are some problems with the database that are not about wrong implementation of MVCC in MemDatabase, but I think the above goes to show that this is not the case (at least not here).

douglaz:
> `tokio::task::yield_now` which I guess expects to be running on tokio runtime direclty(?).

false. yield_now is a very simple future implementation that will be pending once then always ready after that. there is nothing tokio specific there. this issue isn't a incompatibility between async runtimes because there is only one runtime in this project: tokio

douglaz:
Another way to reproduce this deadlock is changing `spawn` to `spawn_blocking` on alephbft:

```diff
diff --git a/fedimint-server/src/consensus/server.rs b/fedimint-server/src/consensus/server.rs
index 9e58052de1..f8d24c1064 100644
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -383,24 +383,20 @@ impl ConsensusServer {
         let (terminator_sender, terminator_receiver) = futures::channel::oneshot::channel();
 
         let (loader, saver) = atomic_broadcast::backup::load_session(self.db.clone()).await;
-
-        let aleph_handle = spawn(
-            "aleph run session",
-            aleph_bft::run_session(
-                config,
-                aleph_bft::LocalIO::new(
-                    DataProvider::new(self.submission_receiver.clone(), signature_receiver),
-                    FinalizationHandler::new(unit_data_sender),
-                    saver,
-                    loader,
-                ),
-                Network::new(self.connections.clone()),
-                self.keychain.clone(),
-                Spawner::new(),
-                aleph_bft_types::Terminator::create_root(terminator_receiver, "Terminator"),
+        let f = aleph_bft::run_session(
+            config,
+            aleph_bft::LocalIO::new(
+                DataProvider::new(self.submission_receiver.clone(), signature_receiver),
+                FinalizationHandler::new(unit_data_sender),
+                saver,
+                loader,
             ),
-        )
-        .expect("some handle on non-wasm");
+            Network::new(self.connections.clone()),
+            self.keychain.clone(),
+            Spawner::new(),
+            aleph_bft_types::Terminator::create_root(terminator_receiver, "Terminator"),
+        );
+        let aleph_handle = tokio::task::spawn_blocking(|| f);
 
         let signed_session_outcome = self
             .complete_signed_session_outcome(

```

dpc:
> negative. yield_now is a very simple future implementation that will be pending once then always ready after that. there is nothing tokio specific there. this issue isn't a incompatibility between async runtimes because there is only one runtime in this project: tokio

The `tokio::task::yield_now` is OK, but I don't think `block_on` is, because it explicitly ["Runs a future to completion on the current thread."](https://docs.rs/futures/latest/futures/executor/fn.block_on.html) and the "current thread" is already tokio's executor which most likely does not like being blocked like that. So that `yield_on` is getting scheduled on a task queue of an executor thread that is no longer scheduling (because it blocked on an ablocking `block_on`). There's a reason Tokio runtime has it's own `block_on`(s), so probably that should be used instead in some form, but when I tried it wasn't working and I think @Maan2003 is hitting same issues in https://github.com/fedimint/fedimint/issues/4054 . Let's continue discussion there.

The log line before that `block_on` is printed, the one after is not, the alephbft executor gets stuck, transaction are being submitted but not processed, test gets stuck. Whatever it is, we should fix it, and we can continue reasoning about this PR.

elsirion:
>  I tried couple of quick attempts to fix it using different `block_on`s and `block_in_place` but I only made things even worse

@dpc so you tried this approach I assume?

https://github.com/fedimint/fedimint/blob/9c36777c4a0f21742bc99c29c54c12d3f74a4e52/devimint/src/external.rs#L208-L214

elsirion:
I see, @Maan2003 did and it helped? https://github.com/fedimint/fedimint/commit/28b036126d361325bd9c167b96b3ad9e7cae2a1c (or at least for that one PR)

justinmoon:
Still relevant?

elsirion:
No, this was related to the AlephBFT sync/async bug.

PR #3997, opened 2023-12-31
feat: Return preimage when creating ln invoice

benthecarman:
Currently it doesn't seem it is possible (at least non-obvious) for how to get the preimage for your own invoices. This is useful for verifying if people paid your invoices and for things like zaps.

Would be nice if this was backported to 0.2.1 as this isn't a breaking change

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -303,7 +303,7 @@ pub async fn handle_command(
             let lightning_module = client.get_first_module::<LightningClientModule>();
             lightning_module.select_active_gateway().await?;
 
-            let (operation_id, invoice) = lightning_module
+            let (operation_id, invoice, _) = lightning_module
                 .create_bolt11_invoice(amount, description, expiry_time, ())
                 .await?;
             Ok(serde_json::to_value(LnInvoiceResponse {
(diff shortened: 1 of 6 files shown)


m1sterc001guy (approved):

okjodom (approved):

justinmoon:
LGTM. I think this needs a rebase to get the flake self-check passing?

benthecarman:
> LGTM. I think this needs a rebase to get the flake self-check passing?

this is on latest master..

Kodylow:
LGTM

justinmoon:
Rebasing to include https://github.com/fedimint/fedimint/pull/4005 once it's merged should fix CI

benthecarman:
rebased

elsirion:
This PR cannot be backported to the 0.2 branch as it is unfortunately. While it doesn't break Fedimint compatibility, it *does* break the rust API, so cannot be done in a minor version. 

If you want it backported I'd propose adding a new fn with the signature introduced here (but under a new name) that returns the preimage. Then the current fn can just call that extended fn and throw away the preimage.

elsirion:
Note on process: that would have been nice to consider before merging to master.

PR #4000, opened 2024-01-02
feat: versioned gateway api

### LINKED ISSUE #3904: Gateway API versioning
Add a gateway API version and make gateway clients aware of it. Would help with issue created when server is running https://github.com/fedimint/fedimint/pull/3847 but client is on older version.

okjodom:
Declares v1 for gateway apis with the following schema: `gateway-http-host/version/endpoint`

After this change, clients should prefer to call `gateway-http-host/v1/endpoint`. The non-versioned api with the format `gateway-http-host/endpoint` is still supported but will likely will deprecate this, or arbitrarily change the apis as necessary

## Consuming api from fedimint-ln-client

This PR proposes that the gateway should register a versioned api url as part of `LightningGateway` payload. The client can
- use the api as is, assuming compatibility. **This is the current client behaviour.**
- check the version before constructing compatible payloads
- ignore a gateway that doesn't have compatible api version

closes #3904

### DIFF
--- a/fedimint-testing/src/gateway.rs
+++ b/fedimint-testing/src/gateway.rs
@@ -36,7 +36,7 @@ pub const DEFAULT_GATEWAY_PASSWORD: &str = "thereisnosecondbest";
 /// Fixture for creating a gateway
 pub struct GatewayTest {
     /// URL for the RPC
-    api: SafeUrl,
+    versioned_api: SafeUrl,
     /// Handle of the running gateway
     pub gateway: Gateway,
     /// Temporary dir that stores the gateway config
@@ -52,7 +52,7 @@ pub struct GatewayTest {
 impl GatewayTest {
     /// RPC client for communicating with the gateway admin API
     pub async fn get_rpc(&self) -> GatewayRpcClient {
-        GatewayRpcClient::new(self.api.clone(), None)
+        GatewayRpcClient::new(self.versioned_api.clone(), None)
     }
 
     /// Removes a client from the gateway
@@ -91,6 +91,8 @@ impl GatewayTest {
     ) -> Self {
         let listen: SocketAddr = format!("127.0.0.1:{base_port}").parse().unwrap();
         let address: SafeUrl = format!("http://{listen}").parse().unwrap();
+        let versioned_api = address.join("v1").unwrap();
+
         let (path, _config_dir) = test_dir(&format!("gateway-{}", rand::random::<u64>()));
 
         // Create federation client builder for the gateway
@@ -138,7 +140,7 @@ impl GatewayTest {
             .await;
 
         // Wait for the gateway web server to be available
-        GatewayTest::wait_for_webserver(address.clone(), cli_password)
+        GatewayTest::wait_for_webserver(versioned_api.clone(), cli_password)
             .await
             .expect("Gateway web server failed to start");
 
@@ -154,7 +156,7 @@ impl GatewayTest {
         let info = lightning.info().await.unwrap();
 
         Self {
-            api: address,
+            versioned_api,
             _config_dir,
             gateway,
             node_pub_key: PublicKey::from_slice(info.pub_key.as_slice()).unwrap(),
@@ -163,8 +165,11 @@ impl GatewayTest {
         }
     }
 
-    pub async fn wait_for_webserver(api: SafeUrl, password: Option<String>) -> anyhow::Result<()> {
-        let rpc = GatewayRpcClient::new(api, password);
+    pub async fn wait_for_webserver(
+        versioned_api: SafeUrl,
+        password: Option<String>,
+    ) -> anyhow::Result<()> {
+        let rpc = GatewayRpcClient::new(versioned_api, password);
         for _ in 0..30 {
             let rpc_result = rpc.get_info().await;
             if rpc_result.is_ok() {
--- a/fedimint-testing/src/gateway.rs
+++ b/fedimint-testing/src/gateway.rs
@@ -19,7 +19,7 @@ use lightning_invoice::RoutingFees;
 use ln_gateway::client::GatewayClientBuilder;
 use ln_gateway::lnrpc_client::{ILnRpcClient, LightningBuilder};
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
-use ln_gateway::rpc::{ConnectFedPayload, FederationConnectionInfo};
+use ln_gateway::rpc::{ConnectFedPayload, FederationConnectionInfo, V1_API_ENDPOINT};
 use ln_gateway::{Gateway, GatewayState};
 use secp256k1::PublicKey;
 use tempfile::TempDir;
@@ -91,7 +91,7 @@ impl GatewayTest {
     ) -> Self {
         let listen: SocketAddr = format!("127.0.0.1:{base_port}").parse().unwrap();
         let address: SafeUrl = format!("http://{listen}").parse().unwrap();
-        let versioned_api = address.join("v1").unwrap();
+        let versioned_api = address.join(V1_API_ENDPOINT).unwrap();
 
         let (path, _config_dir) = test_dir(&format!("gateway-{}", rand::random::<u64>()));
 
@@ -165,6 +165,12 @@ impl GatewayTest {
         }
     }
 
+    /// Waits for the webserver to be ready.
+    ///
+    /// This function is used to ensure that the webserver is fully initialized
+    /// and ready to accept incoming requests. It is designed to be used in
+    /// a concurrent environment where the webserver might be initialized in a
+    /// separate thread or task.
     pub async fn wait_for_webserver(
         versioned_api: SafeUrl,
         password: Option<String>,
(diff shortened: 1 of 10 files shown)


justinmoon:
Will gateway clients will only support 1 version of the gateway API? This works slightly differently than the consensus API versioning which supports a range of versions IIRC but we're comparing REST and JSON RPC so might not be apples-to-apples.

Do we expect that gateway API to serve multiple versions? For instance, once we add v2 would we continue serving v1 (like you do here with old un-versioned routes)?

What happens if a `gateway-cli` running this version of the code encounters a v2 API? Will they see a generic 404 error or a more helpful message telling them to upgrade?

okjodom:
> Will gateway clients will only support 1 version of the gateway API?

Clients can consume multiple versions of the API, but we should start simple where the client assumes a single  / non-versioned gateway API

> This works slightly differently than the consensus API versioning which supports a range of versions IIRC but we're comparing REST and JSON RPC so might not be apples-to-apples

More reason to build a JSON RPC for gatewayd and overlay it with this REST API #2293 . Adding this to 0.4.0 or 0.5.0 milestones

> Do we expect that gateway API to serve multiple versions? For instance, once we add v2 would we continue serving v1 (like you do here with old un-versioned routes)?

Gatewayd will serve multiple api versions, but we should be hesitant / resistant to add new versions

> What happens if a gateway-cli running this version of the code encounters a v2 API? Will they see a generic 404 error or a more helpful message telling them to upgrade?

Not included in current PR but we should include checks for automatic fallback / error handling for incompatible api calls. _maybe we can use un-versioned routes to build compat between versions_

justinmoon:
Rebasing to include https://github.com/fedimint/fedimint/pull/4005 once it's merged should fix flake CI job

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -26,9 +26,34 @@ pub async fn run_webserver(
     gateway: Gateway,
     task_group: &mut TaskGroup,
 ) -> axum::response::Result<()> {
-    let (routes, admin_routes) = if let Some(gateway_config) = config {
+    let routes = v1_routes(config, gateway.clone());
+    let api_v1 = Router::new()
+        .nest("/v1", routes.clone())
+        // Backwards compatibility: Continue supporting gateway APIs without versioning
```
m1sterc001guy, line 32: In the client, perhaps we could add a warning message that is printed whenever the non-versioned API is used? The worry I have with supporting backwards compatibility is introducing unexpected changes to the non-versioned API

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -343,6 +343,8 @@ pub struct LightningGateway {
     pub gateway_redeem_key: secp256k1::PublicKey,
     pub node_pub_key: secp256k1::PublicKey,
     pub lightning_alias: String,
+    /// URL to the gateway's versioned public API
+    /// (e.g. <https://gateway.example.com/v1>)
     pub api: SafeUrl,
```
m1sterc001guy, line 348: nit: rename this to `versioned_api` too

m1sterc001guy (approved):
This seems reasonable to me, most of my comments are nits

modules/fedimint-ln-common/src/lib.rs
okjodom: > nit: rename this to `versioned_api` too

Idk, might be useful to keep this generic in case we need to pass non-versioned apis or non-RESTful api?

gateway/ln-gateway/src/rpc/rpc_server.rs
okjodom: A new gateway deployed will register it's versioned API so all clients in the fed never get to use it's non-versioned endpoints. Non-versioned api's are likely to be (mis)used by gateway UIs

gateway/ln-gateway/src/lib.rs
```diff
@@ -269,11 +275,14 @@ impl Gateway {
         num_route_hints: u32,
         gateway_db: Database,
     ) -> anyhow::Result<Gateway> {
+        let versioned_api = api_addr.join("v1").map_err(|e| {
+            anyhow::anyhow!("Failed to version gateway API address: {api_addr:?}, error: {e:?}")
```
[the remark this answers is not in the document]
okjodom: I'm afraid of expects :)

m1sterc001guy (approved):

okjodom:
@Kodylow , could you take a look at this PR? you've have good context on axum servers

fedimint-testing/src/gateway.rs
```diff
@@ -163,8 +165,11 @@ impl GatewayTest {
         }
     }
 
-    pub async fn wait_for_webserver(api: SafeUrl, password: Option<String>) -> anyhow::Result<()> {
-        let rpc = GatewayRpcClient::new(api, password);
+    pub async fn wait_for_webserver(
```
Kodylow, line 168: function docstring

gateway/cli/src/main.rs
```diff
@@ -99,7 +99,8 @@ async fn main() -> anyhow::Result<()> {
     TracingSetup::default().init()?;
 
     let cli = Cli::parse();
-    let client = || GatewayRpcClient::new(cli.address, cli.rpcpassword);
+    let versioned_api = cli.address.join("v1")?;
```
Kodylow, line 102: can we move this "v1" into a const or config maybe? it's hardcoded in like 5 places

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -26,9 +26,34 @@ pub async fn run_webserver(
     gateway: Gateway,
     task_group: &mut TaskGroup,
 ) -> axum::response::Result<()> {
-    let (routes, admin_routes) = if let Some(gateway_config) = config {
+    let routes = v1_routes(config, gateway.clone());
```
Kodylow, line 29: should make this let v1_routes

Kodylow (approved):
Looks good, minor nits

fedimint-testing/src/gateway.rs
```diff
@@ -163,8 +165,17 @@ impl GatewayTest {
         }
     }
 
-    pub async fn wait_for_webserver(api: SafeUrl, password: Option<String>) -> anyhow::Result<()> {
-        let rpc = GatewayRpcClient::new(api, password);
+    /// Asynchronously waits for the webserver to be ready.
```
dpc, line 168: I think it's best to skip "Asynchronously" here. The function is async, but it just "blocks" from the code's PoV.

dpc (dismissed):

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -26,9 +26,34 @@ pub async fn run_webserver(
     gateway: Gateway,
     task_group: &mut TaskGroup,
 ) -> axum::response::Result<()> {
-    let (routes, admin_routes) = if let Some(gateway_config) = config {
+    let v1_routes = v1_routes(config, gateway.clone());
+    let api_v1 = Router::new()
+        .nest(&format!("/{V1_API_ENDPOINT}"), v1_routes.clone())
+        // Backwards compatibility: Continue supporting gateway APIs without versioning
+        .merge(v1_routes);
```
elsirion, line 33: Should we add an API endpoint that tells us the min/max supported version?

dpc: Not something commonly seen. I guess it's because it falls out of versioned apis, rendering it weirdly special.

In web apis apps usually target a single api version and it's server side job to keep around older versions long enough. Unclear if the same strategy is OK here. But the app can just iterate and try couple of versions to discover which one is available if really needed anyway.

dpc (approved):

dpc:
Will go ahead and approve, extra call (as any other changes) can still be addded in a follow-up.

gateway/ln-gateway/src/rpc/rpc_server.rs
okjodom: following up with this endpoint

elsirion: The problem is we want newer clients to be compatible with older federations, that's not a problem centralized APIs typically have. I think we want backwards and forwards compatibility strictly speaking. For that having some discover mechanism (either iteration or an endpoint) is very useful.

okjodom: I want to read more on [OpenAPI v3.1.0](https://spec.openapis.org/oas/latest.html#openapi-object) spec which defines an info object with a version field. We can aim for spec compliance, serve the api docs at a path like `/` or `/openapi` or `/api-docs` and then expect clients to use this for discovering api capabilities

elsirion: I doubt adhering to a spec which to read and understand would take multiple days is worth the effort unless there is really good rust tooling already that removes this necessity.

PR #3994, opened 2023-12-29
chore: remove usage of bitcoin_hashes::hex::ToHex trait

tvolk131:
Replace usage of `bitcoin_hashes::hex::ToHex` trait with `hex::ToHex` since the former was [removed](https://github.com/rust-bitcoin/rust-bitcoin/pull/1531) in `bitcoin 0.30.0`. Also replace `bitcoin_hashes::hex::FromHex` with `hex::FromHex` where it is easy to do so, even though `bitcoin_hashes::hex::FromHex` still exists.

This is a small step towards updating to `bitcoin 0.30` (#2356).

PR #4005, opened 2024-01-03
chore: `nix flake update`

justinmoon:
Should fix CI for https://github.com/fedimint/fedimint/pull/3997 and https://github.com/fedimint/fedimint/pull/3994

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -190,11 +190,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1701156937,
-        "narHash": "sha256-jpMJOFvOTejx211D8z/gz0ErRtQPy6RXxgD2ZB86mso=",
+        "lastModified": 1703992652,
+        "narHash": "sha256-C0o8AUyu8xYgJ36kOxJfXIroy9if/G6aJbNOpA5W0+M=",
         "owner": "nixos",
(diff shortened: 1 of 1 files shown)


justinmoon:
Didn't work. Could use some help here @dpc.

justinmoon:
`nix flake lock --update-input flakebox/nixpkg` fixed it. Thanks @Maan2003!

elsirion (approved):

Maan2003 (approved):

okjodom (approved):

benthecarman:
Looks like merge for this failed

okjodom:
Re-added to the merge queue but this [failed run on macos](https://github.com/fedimint/fedimint/actions/runs/7400054843/job/20132861506) looks very suspect. Did it fail because of the skipped stages?

justinmoon:
> Looks like merge for this failed

IIRC the MacOS job isn't required because it can be slow. Is that right @dpc?

justinmoon:
Note: when I updated it re-compiled bitcoind when running `nix develop`. IIRC we didn't have to manually re-compile bitcoind in the past. Seems like we should be able to just get a bitcoind binary. Perhaps this only happens in MacOS.

PR #4006, opened 2024-01-03
refactor: remove `DatabaseSource` from client builder

okjodom:
attempting #3781


m1sterc001guy:
See this comment https://github.com/fedimint/fedimint/pull/3781#issuecomment-1885878214, I think there's a better refactor we can do to eliminate the need to re-create the clients at all.

okjodom:
> See this comment [#3781 (comment)](https://github.com/fedimint/fedimint/pull/3781#issuecomment-1885878214), I think there's a better refactor we can do to eliminate the need to re-create the clients at all.

ah makes sense!

PR #4009, opened 2024-01-04
refactor: tbs

joschisan:
We remove generics to make the code more explicit and adhere to a consistent functional style.


crypto/tbs/src/lib.rs
```diff
@@ -174,40 +175,49 @@ pub fn sign_blinded_msg(msg: BlindedMessage, sks: SecretKeyShare) -> BlindedSign
...
-            let x = Scalar::from((idx as u64) + 1);
-            let y = share.0.into();
-            (x, y)
-        })
-        .collect::<Vec<(Scalar, G1Projective)>>();
-    if points.len() < threshold {
-        panic!("Not enough signature shares");
-    }
+/// If the amount of shares supplied is less than two
+pub fn aggregate_signature_shares(
```
elsirion, line 183: Should likely not be pub as it's not safe to use for non-experts. This function assumes that:
* Supplied shares are valid
* The correct number of shares is supplied (unintuitively, supplying too many will also lead to wrong results)

crypto/tbs/src/lib.rs
```diff
@@ -174,40 +175,49 @@ pub fn sign_blinded_msg(msg: BlindedMessage, sks: SecretKeyShare) -> BlindedSign
...
+    shares: &BTreeMap<u64, BlindedSignatureShare>,
+) -> BlindedSignature {
+    BlindedSignature(
+        lagrange_multiplier(shares.keys().cloned().map(Scalar::from).collect())
+            .into_iter()
+            .zip(shares.values())
+            .map(|(lm, y)| lm * y.0)
+            .reduce(|a, b| a + b)
```
elsirion, line 191: Please use more descriptive names, this ain't a math paper :stuck_out_tongue_winking_eye:, e.g. `lagrange_m` instead of `lm` (whatever m stands for xD)

crypto/tbs/src/lib.rs
```diff
@@ -174,40 +175,49 @@ pub fn sign_blinded_msg(msg: BlindedMessage, sks: SecretKeyShare) -> BlindedSign
...
+            .expect("We have at least one share")
```
elsirion, line 192: How do we handle the single-sig case? Probably elsewhere?

crypto/tbs/src/lib.rs
```diff
@@ -174,40 +175,49 @@ pub fn sign_blinded_msg(msg: BlindedMessage, sks: SecretKeyShare) -> BlindedSign
...
-    }
+pub fn aggregate_public_key_shares(shares: &BTreeMap<u64, PublicKeyShare>) -> AggregatePublicKey {
+    AggregatePublicKey(
+        lagrange_multiplier(shares.keys().cloned().map(Scalar::from).collect())
+            .into_iter()
+            .zip(shares.values())
+            .map(|(lm, y)| lm * y.0)
+            .reduce(|a, b| a + b)
+            .expect("We have at least one share")
+            .to_affine(),
```
elsirion, line 205: Feels like parts of this could be refactored out and de-duplicated xD but I guess your goal was to simplify and accept redundancy.

crypto/tbs/src/lib.rs
```diff
@@ -174,40 +175,49 @@ pub fn sign_blinded_msg(msg: BlindedMessage, sks: SecretKeyShare) -> BlindedSign
...
+    )
+}
 
-    let bsig: G1Projective = poly::interpolate_zero(points.into_iter());
-    BlindedSignature(bsig.to_affine())
+fn lagrange_multiplier(scalars: Vec<Scalar>) -> Vec<Scalar> {
```
elsirion, line 209: nit: Do we need to collect or could this return `impl Iterator<Item=Scalar>`?

crypto/tbs/src/poly.rs
```diff
@@ -1,117 +1,28 @@
...
-            result.mul_assign(x);
-            result.add_assign(c);
-        }
-        result
-    }
-
-    pub fn coefficients(&self) -> Iter<G> {
-        self.coefficients.iter()
-    }
+pub fn random_scalar_coefficients(degree: usize, rng: &mut impl RngCore) -> Vec<Scalar> {
```
elsirion, line 7: Technically we probably also want to require `CryptoRng` to be implemented by the rng

crypto/tbs/src/poly.rs
```diff
@@ -1,117 +1,28 @@
...
-{
-    pub fn random(degree: usize, rng: &mut impl RngCore) -> Self {
-        assert_ne!(degree, usize::max_value());
-        let coefficients = (0..=degree).map(|_| G::from_random(rng)).collect();
-        Poly {
-            coefficients,
-            _pd: PhantomData,
-        }
-    }
+pub fn evaluate_polynomial_scalar(coefficients: &[Scalar], x: &Scalar) -> Scalar {
```
elsirion, line 11: Documentation needed: what are assumptions about inputs (at least one coeff, …)

elsirion (commented):
I think removing some of the type-level complexity that I introduced in an overzealous attempt at avoiding code duplication is good, but we should not expose functions that are easy to mis-use. The `tbs` crate should be kinda idiot-proof. We might not get around exposing some of the lower level stuff for the DKG, which we might not be able to move into the tbs crate (but would be worth a try), but the mint module should not need to know that it cannot accumulate more than a given number of notes.

crypto/tbs/src/lib.rs
joschisan: We need to collect.

joschisan:
I think there a two further conceptual issues with the tbs crate that relate to some of the comments here:

The aggregation of public key shares: after the dkg every peer obtains the same polynomial that it uses to calculate the public key shares for every peer. Then the shares are stored in the MintConsensusConfig and are later aggregated (in two different places even) into the public keys with the same Lagrange interpolation as the blind signatures which causes the code duplication @elsirion commented on. However the public key is just the polynomial obtained from the dkg evaluated at zero so the information is trivially available when creating the MintConsensusConfig. Therefore the aggregated public keys should really be stored in the MintConsensusConfig. Aggregating the pk shares is just weird. The main question here, it seems, is how to do it not if?

The amalgamation of key generation with tbs. Both tbs and the threshold point encryption for the ng ln module are quite simple, the design space is small. However, both of them require the same dkg while the dkg is comparatively complicated and has a larger design space as well. That's why I think it makes sense to split the dkg from tbs and tpe - then we would also not have to expose low level functions like evaluate_polynomial in the tbs crate. The dealer_keygen in tbs would be private for testing reasons only. Though this would lead to some code duplication of the helper functions, like dealer_keygen and evaluate_polynomial_scalar it would allow for clean conceptual separation between the crates. This could be done with this pr and the question is if we want to do it at all?

crypto/tbs/benches/tbs.rs
```diff
@@ -35,13 +37,12 @@ mod bench {
...
-            .map(|sk| sign_blinded_msg(bmsg, *sk))
-            .enumerate()
-            .collect::<Vec<_>>();
+        let shares: BTreeMap<u64, BlindedSignatureShare> = (1_u64..)
+            .zip(sks.iter().map(|sk| sign_blinded_msg(bmsg, *sk)))
+            .take(4)
+            .collect();
 
-        bencher.iter(move || combine_valid_shares(shares.clone(), 4));
+        bencher.iter(move || aggregate_signature_shares(&shares.clone()));
```
dpc, line 45: Both `&` and `.clone()` looks unnecessary, but I am probably missing something.

crypto/tbs/src/lib.rs
dpc: IMO if the scope of a symbol is tiny enough, it's kind of OK to go supa-short-names, as it helps readability.

Also [`reduce(Add::add)`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3e921efa3500d42945a5aff8b520a8f7) works and avoids naming anything. 

And isn't the whole thing just a `.sum()`?

elsirion:
> The aggregation of public key shares

Maybe we want to save the polynomial in the cfg instead of the key sahres then and calculate both aggregated pub key and key shares on startup? That should be much more straight forward (and is also more in line with the design of the old threshold crypto crate).

As long as we don't have redundancy in the config I'm ok with it.

elsirion:
> That's why I think it makes sense to split the dkg from tbs and tpe - then we would also not have to expose low level functions like evaluate_polynomial in the tbs crate.

I'm a bit confused: if the `tbs`+`tpe` crates don't expose lower-level functionality, how do you imple the DKG on top? Or would there be a common `poly` crate that is consumed by `dkg`, `tbs` and `tpe` and allows `dkg` to be used for both `tbs` and `tpe` as kind of a blackbox?

joschisan:
> > The aggregation of public key shares
> 
> Maybe we want to save the polynomial in the cfg instead of the key sahres then and calculate both aggregated pub key and key shares on startup? That should be much more straight forward (and is also more in line with the design of the old threshold crypto crate).
> 
> As long as we don't have redundancy in the config I'm ok with it.

Yes that would be a clean way to do it. However we would have to write a db migration that recovers the original polynomial coefficients from the evaluations. Though this is possible we would have to implement new crypto code to do it. Therefore I would suggest to just add the aggregated pk as we already have the code for the db migration this would require. Though this technically adds redundancy to the config please keep in mind that the config already has redundancy by saving the public key shares for every peer instead of just the polynomial.

crypto/tbs/src/lib.rs
joschisan: sum does not work because in case of an empty iterator it returns the zero type but we want to panic

joschisan:
> > That's why I think it makes sense to split the dkg from tbs and tpe - then we would also not have to expose low level functions like evaluate_polynomial in the tbs crate.
> 
> I'm a bit confused: if the `tbs`+`tpe` crates don't expose lower-level functionality, how do you imple the DKG on top? Or would there be a common `poly` crate that is consumed by `dkg`, `tbs` and `tpe` and allows `dkg` to be used for both `tbs` and `tpe` as kind of a blackbox?

I think a common poly crate is overkill as the shared code from polynomial evaluation are just a few lines. Right now I opted for duplication of the poly eval code and also duplicated the dealer_keygen function such that tbs actually only exposes threshold blind signing as the name implies. In the future we may create a keygen crypto crate for dealer_keygen and all cryptographic primitives for dkg such as Feldman or Pedersen commitments, polynomial evaluation. Then we could remove the duplicated code again. However as I do not consider the keygen code stable as tbs I would hold of on this.

fedimint-server/src/config/distributedgen.rs
```diff
@@ -379,6 +377,23 @@ where
...
+
+pub fn evaluate_polynomial_scalar(coefficients: &[Scalar], x: &Scalar) -> Scalar {
+    coefficients
+        .iter()
+        .cloned()
+        .rev()
+        .reduce(|acc, coefficient| acc * x + coefficient)
+        .expect("We have at least one coefficient")
+}
+
```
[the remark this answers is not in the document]
joschisan: I agree. Just wanted to get tbs in its final state already for external review.

crypto/tbs/src/lib.rs
```diff
@@ -174,40 +137,72 @@ pub fn sign_blinded_msg(msg: BlindedMessage, sks: SecretKeyShare) -> BlindedSign
...
+            .into_iter()
+            .zip(shares.values())
+            .map(|(lagrange_multiplier, share)| lagrange_multiplier * share.0)
+            .reduce(|a, b| a + b)
+            .expect("We have at least one share")
+            .to_affine(),
+    )
+}
+
+fn lagrange_multipliers(scalars: Vec<Scalar>) -> Vec<Scalar> {
```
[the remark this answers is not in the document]
joschisan: Where its used it can always consume the vector, does not make a difference.

elsirion (approved):

elsirion (approved):

elsirion:
Dev call: Maybe add a reference to "D. L. Vo, F. Zhang, and K. Kim, “A new threshold blind signature scheme from pairings,” 2003."

dpc (approved):
LGTM, but my expertise here is shallow.

PR #4010, opened 2024-01-04
Add backwards-compatible test matrix

bradleystachurski:
Building off of https://github.com/fedimint/fedimint/pull/3982
Progress on https://github.com/fedimint/fedimint/issues/3962

Adds backwards-compatibility testing for `fedimintd`, `gatewayd`, `fedimint-cli`, and `gateway-cli` defined as a Cartesian product of the binaries and versions defined in `scripts/tests/cross-test-ci-all.sh` (just `v0.2.1` for now). This approach won't scale well, however this is useful to begin identifying breaking changes in `master`, which already exist.

In CI, this step takes ~30 minutes so I've decided to run on each push instead of only in the merge queue. I think it's a good future improvement to parallelize the runs and to consider running only on merge.

Backwards-compatibility matrix is [logged](https://github.com/fedimint/fedimint/actions/runs/7434144574/job/20227985635?pr=4010#step:5:5554) at the end of the run.

```
Backwards-compatibility tests summary:
fed_version  client_version  gateway_version  exit_code
v0.2.1       v0.2.1          current          1
v0.2.1       current         v0.2.1           1
v0.2.1       current         current          1
current      v0.2.1          v0.2.1           1
current      v0.2.1          current          1
current      current         v0.2.1           1
```

If any combination fails, this step will fail CI.


scripts/tests/cross-test-ci-all.sh
```diff
@@ -0,0 +1,93 @@
...
+    export FM_GATEWAYD_BASE_EXECUTABLE
+  fi
+}
+
+test_results="fed_version,client_version,gateway_version,exit_code\n"
+has_failure=false
+versions+=("current")
+for fed_version in "${versions[@]}"; do
+  for client_version in "${versions[@]}"; do
+    for gateway_version in "${versions[@]}"; do
```
bradleystachurski, line 59: During the deep dive we discussed `fedimintd` and `gatewayd` will be on the same version since we should assume gateway operators are technically savvy. I'm hesitant to make the assumption that all gateway operators will coordinate upgrades with all federations they're connected to and think it would be good to know if we're introducing breaking changes.

I'm open to removing if others feel strongly.

scripts/tests/cross-test-ci-all.sh
```diff
@@ -0,0 +1,93 @@
...
+  if [[ "$version" == "current" ]]; then
+    unset FM_FEDIMINTD_BASE_EXECUTABLE
+  else
+    >&2 echo "Compiling fed binaries for version $version..."
+    FM_FEDIMINTD_BASE_EXECUTABLE="$(nix_build_binary_for_version 'fedimintd' "$version")"
+    export FM_FEDIMINTD_BASE_EXECUTABLE
+  fi
+}
+
+function use_client_binaries_for_version() {
```
bradleystachurski, line 29: I'm allowing `gateway-cli` and `gatewayd` to be on different versions since that's the only way automated tests would have caught https://github.com/fedimint/fedimint/issues/3977.

If we're assuming a gateway operator is updating the daemon at the same time as the client, do we need to ensure changes are backwards-compatible?

scripts/tests/cross-test-ci-all.sh
```diff
@@ -0,0 +1,93 @@
...
+      >&2 echo "client version: $client_version"
+      >&2 echo "gateway version: $gateway_version"
+    done
+  done
+done
+
+>&2 echo "Backwards-compatibility tests summary:"
+echo -e "$test_results" | >&2 column -t -s ','
+# CI requires explicitly exiting with an error or success code
+[[ "$has_failure" == "true" ]] && exit 1 || exit 0
```
bradleystachurski, line 93: Alternatively, we can temporarily remove `exit 1` so this step won't fail CI and we'll still be able to investigate/fix failures. Once master is in a good state, we can reintroduce `exit 1`.

scripts/tests/cross-test-ci-all.sh
elsirion: Happy to cover it in tests, more is better, but we'll have to decide what to run when. The full matrix will likely be too slow for running on every push to a PR.

.github/workflows/ci-nix.yml
```diff
@@ -126,6 +126,15 @@ jobs:
         if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests && (matrix.host != 'macos')
         run: nix build -L .#wasm32-unknown.ci.wasmTest --keep-failed
 
+
+      - name: Cross Tests
```
justinmoon, line 130: "cross" is slightly confusing because sometimes we use that cross-compilation

.github/workflows/ci-nix.yml
```diff
@@ -126,6 +126,15 @@ jobs:
...
+        if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests
+        run: |
+          # the default tmp dir is too long (/home/ubuntu/actions-runner/_work/_temp/)
+          mkdir /home/ubuntu/tmp
```
dpc, line 134: Ubuntu? :S

Any reason just `/tmp` doesn't work? Then at least `$HOME/tmp` maybe so I can pretend we are not running Ubuntu? :D

scripts/tests/cross-test-ci-all.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
```
dpc, line 1: `cross` not a good name for this file

scripts/tests/test-ci-all.sh
```diff
@@ -74,17 +74,26 @@ function load_test_tool_test() {
 export -f load_test_tool_test
 
 function backend_test_bitcoind() {
-  fm-run-test "${FUNCNAME[0]}" env FM_TEST_ONLY=bitcoind ./scripts/tests/backend-test.sh
+  # backend tests don't support different versions, so we skip for backwards-compatibility tests
```
dpc, line 77: I'm confused why we need to touch anything here.

scripts/tests/cross-test-ci-all.sh
dpc: Instead of one `fedimint-cli x fedimintd x gateways`, I'd rather do targeted ``fedimint-cli x fedimintd`, `fedimintd x gateways`, `fedimint-cli x gateways`, and each only with tests that make sense for it.

There might be some cases where "x can't communicate with y, but only if z is in a weird version", but seems significantly less likely and we could run a full matrix before releases.

bradleystachurski: I agree that sounds like a good improvement. I put up this PR with the following feedback in mind.

> This is probably the most important PR unless severe bugs are found, otherwise we won't be able to keep developing with confidence.

https://github.com/fedimint/fedimint/pull/3982#issuecomment-1873857758

Do you think it's okay to address in followup PRs or is this a blocker for merging? For context, I won't have much bandwidth this week to audit which tests make sense specifically for backwards compatibility, so this will delay merging any backwards compatibility tests for a while.

.github/workflows/ci-nix.yml
bradleystachurski: Running on ubunutu in buildjet gives us user:group as ubuntu:ubuntu. I was hoping to use /tmp, but that's owned by root. I updated to $HOME/tmp.

scripts/tests/cross-test-ci-all.sh
```diff
@@ -0,0 +1,93 @@
...
+      use_gateway_binaries_for_version "$gateway_version"
+
+      >&2 echo "========== Starting backwards-compatibility run ==========="
+      >&2 echo "fed version: $fed_version"
+      >&2 echo "client version: $client_version"
+      >&2 echo "gateway version: $gateway_version"
+
+      # continue running against other versions if there's a failure
+      set +e
+      (./scripts/tests/test-ci-all.sh)
```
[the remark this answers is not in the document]
bradleystachurski: Happy to address. Similar to https://github.com/fedimint/fedimint/pull/4010#discussion_r1446310554

scripts/tests/test-ci-all.sh
bradleystachurski: I think you answered your own question in the next comment 😄 
> Ohhhhh... So we're running all tests as a "backward compat test"?

In case it's still unclear, running `test-ci-all` inside the backwards-compatibility tests triggers the backend tests. This PR only adds different versioned binaries, so there isn't a way to use different versioned clients/daemons in our rust code. Running the backend tests will only test the current version, so I'm explicitly skipping since that's already tested in CI.

.github/workflows/ci-nix.yml
dpc:
```
> ls -alh / | grep tmp
drwxrwxrwt  73 root root  44K Jan  9 08:46 tmp
```

`/tmp` is owned by `root` on every Linux system. It uses sticky bit `t` to allow creation of entries by other users. There are probably some other reasons wny action runner would default to something else, probably invalid ones. :D

dpc (approved):

dpc:
I'm fine with landing, but I think the first most needed improvement would be running these tests in parallel.

Screenshot from 2024-01-10 09-31-28
Screenshot from 2024-01-10 09-31-19

scripts/tests/cross-test-ci-all.sh
bradleystachurski: @elsirion I see you triggered a [rerun](https://github.com/fedimint/fedimint/actions/runs/7466845389/job/20422326326?pr=4010) of CI. I'm explicitly failing CI if the backwards compatibility tests fail, so we'll need to either address the failures or allow CI to pass prior to merging.

dpc (approved):

elsirion:
Should we also test old/new `gateway-cli` with old/new gateway? I think we currently test some combinations by accident, at least I'm getting the following error, which is caused by #3880:

```
00:01:27 Error: polling gateway connect-fed after 60 retries
00:01:27 
00:01:27 Caused by:
00:01:27     0: command: /nix/store/k6x11hjg5frcc8frqhadylh66yirz6lk-gateway-cli/bin/gateway-cli --rpcpassword=theresnosecondbest -a ht
tp://127.0.0.1:11434 connect-fed fed11qgqpw9thwvaz7te3xgmjuvpwxqhrzw3j8qmrsvp0qqqjpvtxswx3mg6s3dfg8xsc0jjfjg86zs75nvtgxpzqvk4na2rpcv3kg
xhfny
00:01:27     1: exit status: 1
00:01:27        stdout:
00:01:27        
00:01:27        stderr:
00:01:27        Error: error decoding response body: missing field `balance_msat` at line 1 column 59549
00:01:27        
00:01:27        Caused by:
00:01:27            missing field `balance_msat` at line 1 column 59549
```

bradleystachurski:
> Should we also test old/new gateway-cli with old/new gateway? I think we currently test some combinations by accident, at least I'm getting the following error, which is caused by https://github.com/fedimint/fedimint/pull/3880

I agree it's useful to test old/new `gateway-cli` with old/new `gatewayd`. It's already intentionally included in the test matrix for this PR :smile:

https://github.com/fedimint/fedimint/pull/4010#discussion_r1443873703

I've previously reverted https://github.com/fedimint/fedimint/pull/3880 locally and verified the `balance_msat` error is resolved in the backwards compatibility tests.

douglaz:
Good stuff

elsirion:
> I agree it's useful to test old/new `gateway-cli` with old/new `gatewayd`. It's already intentionally included in the test matrix for this PR 😄

Sorry :see_no_evil: I went by [the table](https://github.com/fedimint/fedimint/pull/4010#issue-2066005232) and didn't check the code :sweat_smile:

bradleystachurski:
> Sorry 🙈 I went by https://github.com/fedimint/fedimint/pull/4010#issue-2066005232 and didn't check the code 😅

No worries!

Also, for visibility to folks following this PR (https://github.com/fedimint/fedimint/pull/4050), I'm currently debugging some issues in CI in a separate PR to avoid cluttering this one. The GH actions runner started running out of disk space before finishing the backwards-compatibility run.

.github/workflows/ci-nix.yml
```diff
@@ -126,6 +126,17 @@ jobs:
         if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests && (matrix.host != 'macos')
         run: nix build -L .#wasm32-unknown.ci.wasmTest --keep-failed
 
+      - name: Cleanup disk space
+        if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests
+        run: sudo rm -rf /usr/local/lib/android
```
bradleystachurski, line 131: The buildjet runner started running out of space during the backwards compatibility tests. It's not obvious to me why this just started happening after several successful CI runs previously without a rebase. In a separate branch I identified some low hanging fruit to cleanup, along with deleting the `devimint-*` directory, which was adding ~2GB per matrix iteration.

https://github.com/fedimint/fedimint/pull/4050

https://github.com/fedimint/fedimint/actions/runs/7554418498/job/20567202313?pr=4050#step:11:4336

```
Filesystem      Size  Used Avail Use% Mounted on
udev             14G     0   14G   0% /dev
tmpfs           2.8G  992K  2.8G   1% /run
/dev/sda1       114G  112G  1.6G  99% /
tmpfs            14G  4.0K   14G   1% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs            14G     0   14G   0% /sys/fs/cgroup
/dev/sda15      105M  6.1M   99M   6% /boot/efi
/dev/loop0       64M   64M     0 100% /snap/core20/1822
/dev/loop1       50M   50M     0 100% /snap/snapd/17950
/dev/loop2       92M   92M     0 100% /snap/lxd/24061
tmpfs           2.8G     0  2.8G   0% /run/user/0
/dev/loop3       64M   64M     0 100% /snap/core20/2105
Did not have permissions for all directories
6.8G       ┌── CodeQL          │███▒▒░                                    │   6%
 10G     ┌─┴ hostedtoolcache   │█████░                                    │  10%
 13G   ┌─┴ opt                 │██████                                    │  12%
 11G   │               ┌── deps│█████▓▓▓                                  │  10%
 17G   │             ┌─┴ debug │███████▓                                  │  17%
 17G   │           ┌─┴ target  │███████▓                                  │  17%
 17G   │         ┌─┴ fedimint  │███████▓                                  │  17%
 17G   │       ┌─┴ fedimint    │███████▓                                  │  17%
 17G   │     ┌─┴ _work         │███████▓                                  │  17%
 18G   │   ┌─┴ actions-runner  │████████                                  │  17%
 19G   │ ┌─┴ ubuntu            │████████                                  │  19%
 20G   ├─┴ home                │████████                                  │  19%
 31G   │ ┌── store             │█████████████                             │  30%
 31G   ├─┴ nix                 │█████████████                             │  30%
6.2G   │ ┌── lib               │███░░░░░░░░░░░                            │   6%
 12G   │ │     ┌── sdk         │█████▓▒▒▒░░░░░                            │  11%
 12G   │ │   ┌─┴ android       │█████▓▒▒▒░░░░░                            │  11%
 13G   │ │ ┌─┴ lib             │██████▒▒▒░░░░░                            │  13%
 22G   │ ├─┴ local             │█████████░░░░░                            │  21%
 36G   ├─┴ usr                 │██████████████                            │  34%
106G ┌─┴ /                     │█████████████████████████████████████████ │ 100%
```

scripts/tests/cross-test-ci-all.sh
bradleystachurski: @dpc and I discussed out-of-band that a sane strategy for getting this PR merged is to not return an `exit 1` if there's a backwards-compatibility failure, which allows CI to pass.

Once we investigate/fix failures we can reintroduce failing CI and preventing merging if there's a breaking change introduced.

douglaz (dismissed):
Perhaps there are some optimizations that could be done, but looks good enough for now

.github/workflows/ci-nix.yml
dpc: @bradleystachurski The problem is that `cargo build` etc. is very wasteful.

Nix makes it better because `crane` compresses `./target` directory and keeps it deduplicated between runs.

Now this change introduces a whole new build of `./target`, just like a developer would have locally, and that does not decompose with with anything else.

We could avoid it, by making these scripts a little smarter and use binaries built already via Nix, instead of running an extra raw `cargo build`.

This can be made conditional by the `CI` env var , that would skip all `cargo build`s, and for the "testing current version" use `nix build` to get the binaries of the current source code.

dpc: Eh... but this won't work for running tests ... Oh dear ...

In that case maybe running the whole thing on a separate GHA workflow, that runs *after* the main one passes is the way to go.

dpc: Fixing this might require twisting the solution inside out, so I'm not sure if there's a point landing it...

We might want to think a bit deeper about the whole thing. :/

bradleystachurski:
The latest change moves the backwards-compatibility tests to a separate github action workflow that isn't required to pass.
- Doesn't add any additional time to the required `Build on linux` workflow
- Gives a clear pass/fail of backwards-compatibility in CI if we wait for CI to fnish
- Doesn't block merging if backwards-compatibility doesn't finish or fails

bradleystachurski:
Rebasing to include the fix from https://github.com/fedimint/fedimint/pull/4104

.github/workflows/ci-backwards-compatibility.yml
```diff
@@ -0,0 +1,83 @@
...
+    branches: [ "main", "master", "devel", "releases/v*" ]
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+  pull_request:
+    branches: [ "main", "master", "devel", "releases/v*" ]
+  merge_group:
+    branches: [ "main", "master", "devel", "releases/v*" ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
dpc, line 18: I would like to avoid keeping it as a PR forever, but I'm worried that even non-required run doesn't buy us anything, will cost money to run, will still cause PRs to wait for it to finish slowing us down and will make the :heavy_check_mark: to turn into :negative_squared_cross_mark: in the UI.

That's why I was thinking about making it available as `just ...` command for now, and maybe we could make it a workflow that one can dispatch manually, and just not run it on PRs at all for now.

But I guess we can land first, see how it goes, and tweak later.

dpc (approved):

.github/workflows/ci-backwards-compatibility.yml
bradleystachurski: re: costs

Some back-of-the-envelope math

- 70 commits to `fedimint/fedimint` in the last month
  - https://github.com/fedimint/fedimint/pulse/monthly
  - Assuming 1 backwards-compatibility run per commit
- Assuming cost is accurately reflected in buildjet's [pricing](https://buildjet.com/for-github-actions/docs/about/pricing), we pay $0.016 / min
- Backwards-compatibility workflow takes ~30 minutes

`0.016 $/min * 30 min * 70 commits/month => ~$34 / month`

Ballpark, $34 / month seems reasonable.

elsirion: Runtime cost isn't our biggest concern imo, but number of blocked runners are (and upping our runner count would be $$$). I'm ok with turning it on and seeing how it goes.

elsirion (approved):

.github/workflows/ci-nix.yml
```diff
@@ -126,6 +126,12 @@ jobs:
         if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests && (matrix.host != 'macos')
         run: nix build -L .#wasm32-unknown.ci.wasmTest --keep-failed
 
+
+      - name: Cross Tests
+        if: (github.event_name != 'pull_request' || matrix.build-in-pr) && matrix.run-tests
+        run: |
+          ./scripts/tests/cross-test-ci-all.sh
+
```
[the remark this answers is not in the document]
elsirion: I forgot to submit an old review :see_no_evil:

nix/flakebox.nix
```diff
@@ -374,6 +374,15 @@ rec {
...
   };
 
+  # ciCrossTestAll = craneLibTests.mkCargoDerivation {
+  #   pname = "${commonCliTestArgs.pname}-all-cross";
+  #   cargoArtifacts = workspaceBuild;
+  #   buildPhaseCargoCommand = ''
+  #     patchShebangs ./scripts
+  #     ./scripts/tests/cross-test-ci-all.sh || exit 1
+  #   '';
+  # };
```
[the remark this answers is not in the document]
elsirion: I forgot to submit an old review :see_no_evil:

.github/workflows/ci-nix.yml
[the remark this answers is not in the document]
bradleystachurski: If the workflow becomes a burden, I think trying this or @dpc's suggestion of disabling the workflow and running locally are good choices.

PR #4011, opened 2024-01-04
refactor: rename server db migrations

m1sterc001guy:
In preparation for the client DB migrations, this PR refactors the server DB migration tests:

- Module DB migration tests are moved to their corresponding testing crates (e.g fedimint-ln-tests)
- Folders underneath `db/` have been updated to include `server` (e.g `mint-server-v0`)
- Server migration snapshots have been re-generated (but no breaking changes)
- Git pre-commit hook was updated to exclude DB migration files from whitespace check


modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -585,3 +585,349 @@ fn build_wallet_server_configs(
     )?;
     Ok((wallet_cfg.into_values().collect(), client_cfg))
 }
+
+#[cfg(test)]
```
elsirion, line 589: Is this `cfg` attribute still strictly needed?

elsirion (approved):
Good idea, since the tests are already in their own modules maybe it would also be nice to have them in separate files?

modules/fedimint-wallet-tests/tests/tests.rs
Maan2003: yes, we shouldn't need cfg(test) in -tests crate

Maan2003 (approved):

Issue #2626, opened 2023-06-07
HTLC filtering on the edge


okjodom:
#2622 moves htlc filtering into gatewayd by changing the gRPC api spec of `route_htlcs`. After successful migration Gateway NG, #2600 , we should  consider efficiency gains to lightning service providers by filtering htlcs within `LightningGateway` service

m1sterc001guy:
Ideally this should be designed to support multiple `gatewayd` processing subscribing to HTLCs from a service provider.

m1sterc001guy:
Core Lightning support will be removed post v0.5 https://github.com/fedimint/fedimint/issues/6218, which is the only issue this applies to. Closing

PR #4012, opened 2024-01-04
feat: cln selective htlc filtering (edge filtering aka predicate pushdown)

Kodylow:
addresses https://github.com/fedimint/fedimint/issues/2626

Moves htlc filtering logic to the lightning node side instead of forwarding all htlcs to gateway to filter.

Adds `update_fedimint_scids` rpc for gateway to call to update which federations to look out for. Works with `cln` extension by adding a `fedimint_scids` to the extension struct, but would require a daemon for lnd to run alongside the lnd node to maintain state.


Kodylow:
@m1sterc001guy tagging for vis. Should we just finish impling this for cln then leave lnd as todo until we write the lnd companion daemon? Think that companion daemon will pair with the gateway limits idea we were discussing on another issue with @elsirion

m1sterc001guy:
> Should we just finish impling this for cln then leave lnd as todo until we write the lnd companion daemon? 

I don't think that's necessary, we can still implement the logic for the LND implementation we have. Just need to send the short channel ids over to the thread that is doing the HTLC interception for LND, and implement the filtering logic there too.

okjodom:
should you also remove the [in-gateway filtering of intercepted htlcs ](https://github.com/fedimint/fedimint/blob/af1f99ef2b56ee6253b2e981dd7f72eaddc50d25/gateway/ln-gateway/src/lib.rs#L555C18-L555C36) as part of this change

elsirion:
> should you also remove the [in-gateway filtering of intercepted htlcs ](https://github.com/fedimint/fedimint/blob/af1f99ef2b56ee6253b2e981dd7f72eaddc50d25/gateway/ln-gateway/src/lib.rs#L555C18-L555C36)as part of this change

Maybe the filtering code can be reused there? If we want to continue with the gateway extension concept where third parties can build these I'd rather have a sanity-check inside the gateway to catch any HTLCs we can't handle in case someone builds a broken extension.

m1sterc001guy:
> If we want to continue with the gateway extension concept where third parties can build these I'd rather have a sanity-check inside the gateway to catch any HTLCs we can't handle in case someone builds a broken extension.

+1 to this. `Gatewayd` and the extensions operate as a distributed system, we should not be making any assumptions about how the extensions operate.

Also, we briefly chatted on Friday about the issue of integrating filtering into our LND implementation. Due to the complexities of how LND sets up the HTLC interception streams, we determined that it is probably best to not implement it in our current LND implementation, but to save it for the standalone LND companion daemon.

fedimint-testing/src/ln/real.rs
```diff
@@ -729,6 +743,13 @@ impl ILnRpcClient for LdkLightningTest {
     ) -> Result<EmptyResponse, LightningRpcError> {
         unimplemented!("Unsupported: we dont currently support HTLC interception for LDK Node");
     }
+
+    async fn update_fedimint_scids(
+        &self,
+        _scids: Vec<u64>,
+    ) -> Result<EmptyResponse, LightningRpcError> {
+        unimplemented!("Unsupported: we dont currently support updating scids for LDK Node");
```
okjodom, line 751: nit: should we return empty response for consistency between node impls where we don't support edge filtering?

okjodom (dismissed):

fedimint-testing/src/ln/real.rs
Kodylow: updated

fedimint-testing/src/ln/mock.rs
```diff
@@ -162,4 +162,11 @@ impl ILnRpcClient for FakeLightningTest {
     ) -> Result<EmptyResponse, LightningRpcError> {
         Ok(EmptyResponse {})
     }
+
+    async fn update_fedimint_scids(
```
m1sterc001guy, line 166: nit: I don't think the word fedimint does much for us here. I would prefer `update_target_scids`

gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -172,3 +178,8 @@ message GetRouteHintsResponse {
   // The route hints to the associated lightning node
   repeated RouteHint route_hints = 1;
 }
+
+message UpdateFedimintScidsRequest {
+  // The short channel ids of the associated lightning node
```
m1sterc001guy, line 183: nit: comment can be updated to mention that the short channel ids correspond to federations that are connected to a gateway

gateway/ln-gateway/src/lib.rs
```diff
@@ -503,6 +503,18 @@ impl Gateway {
...
 
+                                        let scids = self
+                                            .scid_to_federation
+                                            .read()
+                                            .await
+                                            .keys()
+                                            .cloned()
+                                            .collect();
+                                        if let Err(e) = ln_client.update_fedimint_scids(scids).await {
+                                            warn!("Failed to update fedimint scids: {e:?}");
```
m1sterc001guy, line 514: Let's update this comment to mention that all HTLCs will be sent to `gatewayd` for inspection

gateway/ln-gateway/src/lib.rs
```diff
@@ -898,6 +910,15 @@ impl Gateway {
...
                 .await?;
+            let scids: Vec<u64> = self
+                .scid_to_federation
+                .read()
+                .await
+                .keys()
+                .cloned()
+                .collect();
+
+            lnrpc.update_fedimint_scids(scids).await?;
```
m1sterc001guy, line 921: I think we should log and swallow the error here, instead of propagating it up. The reason being since if we fail to update the scids here, everything should still work, we just won't get the optimization right away. But propagating the error up will fail the registration, which I don't think we want

gateway/ln-gateway/src/lib.rs
```diff
@@ -914,7 +935,20 @@ impl Gateway {
...
+        if let GatewayState::Running { lnrpc, .. } = self.state.read().await.clone() {
+            let scids: Vec<u64> = self
+                .scid_to_federation
+                .read()
+                .await
+                .keys()
+                .cloned()
+                .collect();
+
+            lnrpc.update_fedimint_scids(scids).await?;
```
m1sterc001guy, line 948: Same comment as above, we shouldn't fail here if we fail to update the short channel ids.

gateway/ln-gateway/src/lnrpc_client.rs
```diff
@@ -220,6 +228,20 @@ impl ILnRpcClient for NetworkLnRpcClient {
...
+
+    async fn update_fedimint_scids(
+        &self,
+        scids: Vec<u64>,
+    ) -> Result<EmptyResponse, LightningRpcError> {
+        let mut client = Self::connect(self.connection_url.clone()).await?;
+        let res = client
+            .update_fedimint_scids(UpdateFedimintScidsRequest { scids })
+            .await
+            .map_err(|status| LightningRpcError::FailedToCompleteHtlc {
```
m1sterc001guy, line 240: Use a different error here rather than `FailedToCompleteHtlc`

gateway/ln-gateway/src/lib.rs
Kodylow: what comment? the warn error?

gateway/ln-gateway/src/lnrpc_client.rs
Kodylow: updated to `FailedToUpdateScids`

gateway/ln-gateway/src/lib.rs
Kodylow: updated to warn and swallow

gateway/ln-gateway/src/lib.rs
Kodylow: updated to warn and swallow

gateway/ln-gateway/proto/gateway_lnrpc.proto
Kodylow: updated

fedimint-testing/src/ln/mock.rs
Kodylow: I changed it across the board to just `update_scids`

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -543,13 +556,15 @@ type HtlcOutcomeSender = oneshot::Sender<serde_json::Value>;
 pub struct ClnHtlcInterceptor {
     pub outcomes: Arc<Mutex<BTreeMap<(u64, u64), HtlcOutcomeSender>>>,
     sender: Arc<Mutex<Option<HtlcInterceptionSender>>>,
+    pub fedimint_scids: Arc<Mutex<BTreeSet<u64>>>,
```
m1sterc001guy, line 559: nit: can remove fedimint here too

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Sorry, yes I meant the warn message. That might be useful for someone running the gateway to know

gateway/ln-gateway/src/lnrpc_client.rs
```diff
@@ -99,6 +102,8 @@ pub trait ILnRpcClient: Debug + Send + Sync {
         &self,
         htlc: InterceptHtlcResponse,
     ) -> Result<EmptyResponse, LightningRpcError>;
+
+    async fn update_scids(&self, scids: Vec<u64>) -> Result<EmptyResponse, LightningRpcError>;
```
m1sterc001guy, line 106: Let's add a doc comment here too, we probably should add one for `complete_htlc` as well :)

m1sterc001guy:
Looking pretty good imo, I only have a few more nit comments

gateway/ln-gateway/src/lib.rs
elsirion: The warning should only occur when running a new gateway+old extension, right?

gateway/ln-gateway/src/lib.rs
```diff
@@ -901,6 +913,17 @@ impl Gateway {
...
+                .scid_to_federation
+                .read()
+                .await
+                .keys()
+                .cloned()
+                .collect();
+
+            if let Err(e) = lnrpc.update_scids(scids).await {
+                warn!("Failed to update scids: {:?}", e);
+            }
```
elsirion, line 926: Code duplication, can this become a function?

gateway/ln-gateway/src/lib.rs
```diff
@@ -917,7 +940,22 @@ impl Gateway {
...
```
elsirion, line 955: even more duplication

elsirion (commented):
Looks mostly good, some code duplication nits

gateway/ln-gateway/src/lib.rs
Kodylow: added

gateway/ln-gateway/src/lib.rs
Kodylow: moved into function

gateway/ln-gateway/src/lib.rs
```diff
@@ -514,6 +514,10 @@ impl Gateway {
                                             }
                                         }
 
+                                        self.register_clients_timer(&mut htlc_task_group).await;
+                                        self.load_clients().await;
```
m1sterc001guy, line 518: This changed since your last rebase. We don't want to re-load the clients and re-register the timer loop on every connection to the lightning node anymore.

gateway/ln-gateway/src/lib.rs
```diff
@@ -559,6 +563,20 @@ impl Gateway {
         Ok(())
     }
 
+    async fn update_fed_scids(&self, ln_client: &Arc<dyn ILnRpcClient>) {
```
m1sterc001guy, line 566: Add function comment

gateway/ln-gateway/src/lib.rs
Kodylow: removed in latest commit

gateway/ln-gateway/src/lib.rs
Kodylow: added

Kodylow:
Had to do a giant rebase to get this working with the lightning dir refactor, shoulda landed that one after this but oh well haha

elsirion (approved):
I think this makes sense and we can iterate on the extension interface (I think it's safe to assume that extension and GW will be update together).

gateway/ln-gateway/src/lightning/cln.rs
```diff
@@ -139,4 +140,15 @@ impl ILnRpcClient for NetworkLnRpcClient {
...
         Ok(res.into_inner())
     }
+
+    async fn update_scids(&self, scids: Vec<u64>) -> Result<EmptyResponse, LightningRpcError> {
+        let mut client = Self::connect(self.connection_url.clone()).await?;
+        let res = client
+            .update_scids(UpdateScidsRequest { scids })
+            .await
+            .map_err(|status| LightningRpcError::FailedToUpdateScids {
+                failure_reason: status.message().to_string(),
```
douglaz, line 150: Perhaps this gives more info?
```suggestion
                failure_reason: format!("{status:?}"),
```

m1sterc001guy:
I realized we never merged this. Would it be possible to rebase this and get it in? I can take it over if needed.

Kodylow:
Just redid it to rebase @m1sterc001guy

m1sterc001guy:
@Kodylow with lnv2 almost ready to ship, I don't think this is worth it anymore. If we decide to keep CLN support, we may want to filter HTLCs based off payment hash instead of short channel ids. Are you ok if we close this?

Kodylow:
yup all good, can close

PR #4015, opened 2024-01-05
fix: add exponential backoff to autocommit with random delay

Maan2003:
this makes transactions less likely to conflict in next try

originally part of #3989

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -41,12 +41,14 @@ use std::marker::{self, PhantomData};
 use std::ops::{self, DerefMut};
 use std::pin::Pin;
 use std::sync::Arc;
+use std::time::Duration;
 
 use anyhow::{bail, Context, Result};
 use fedimint_core::util::BoxFuture;
 use fedimint_logging::LOG_DB;
 use futures::{Stream, StreamExt};
 use macro_rules_attribute::apply;
+use rand::Rng;
 use serde::Serialize;
 use strum_macros::EnumIter;
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -455,6 +457,9 @@ impl Database {
                     }
                 }
             }
+            let delay = (2u64.pow(curr_attempts.min(8) as u32) * 10).min(2000)
```
dpc, line 460: Not a big deal and I doubt it matters here, but adding `0..30` last is kind of meh. Ideally one wants to randomize for the wide range (as wide as waiting time) to break any possibilities of two things retrying in any kind of a lockstep. So instead of `exp + gen_range(0..30)`, something like `gen_range(exp..(exp * 2))` or `gen_range((exp / 2)..(exp + (exp / 2)))` to keep the expected time the same.

Maan2003: done

dpc (approved):

fedimint-core/src/db/mod.rs
```diff
@@ -455,6 +457,9 @@ impl Database {
                     }
                 }
             }
+            let delay = (2u64.pow(curr_attempts.min(7) as u32) * 10).min(1000);
```
elsirion, line 460: The second `min` seems a bit redundant given that the max delay would still be 1280ms without it.

fedimint-core/src/db/mod.rs
```diff
@@ -455,6 +457,9 @@ impl Database {
...
+            let delay = rand::thread_rng().gen_range(delay..(2 * delay));
```
elsirion, line 461: nit: `delay` doesn't have a semantic type, better alternatives would be:
* Converting to `Duration` as soon as possible (idk if we can generate a range of `Duration`, but worth a try)
* Give a more descriptive name that includes the unit (ms)

elsirion (approved):
Left some nits, but should improve performance in concurrent settings, so ACKing.

PR #4018, opened 2024-01-08
test: Run Dummy Migration Tests

m1sterc001guy:
The dummy module is the only module that currently has an example database migration. We previously did not have a test for this migration, so this PR adds the database migration tests for the dummy module.

Also includes some minor refactors to better support multiple versions within the migration tests.


elsirion:
Given what I encountered and fixed in edf8d46813780bf8603867a37cb47fb2d905b140 I'm not sure if the tests are actually testing migrations here or only read the checked-in DBs.

m1sterc001guy:
> Given what I encountered and fixed in https://github.com/fedimint/fedimint/commit/edf8d46813780bf8603867a37cb47fb2d905b140 I'm not sure if the tests are actually testing migrations here or only read the checked-in DBs.

Separate from that PR, this PR has some issues dealing with the db_version. I will fix.

m1sterc001guy:
@elsirion I simplified this PR, added your fix for writing the `DatabaseVersion` into the snapshot dbs, and added some logging to the tests to validate that the migrations are actually run.

dpc (approved):

modules/fedimint-dummy-tests/tests/tests.rs
```diff
@@ -140,3 +140,129 @@ async fn unbalanced_transactions_get_rejected() -> anyhow::Result<()> {
...
+                DUMMY_INSTANCE_ID,
+                DummyCommonInit::KIND,
+                <Dummy as ServerModule>::decoder(),
+            )]),
+        )
+        .await
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_migrations() -> anyhow::Result<()> {
```
Maan2003, line 210: follow up: would be nice to have a failing migration test

Maan2003 (approved):

PR #4019, opened 2024-01-08
chore: speed up recovery by batching and streaming

dpc:
This is effectively undoing some parts of
21da63533cf34722cff1e5afb9e8e06cd02a5eeb
hopefully to improve the real life recovery performance, which people are reporting as very slow.

The WIP recovery refactoring should render this irrelevant, but might be a good bandaid for 0.2.

### DIFF
--- a/modules/fedimint-mint-client/src/backup/recovery.rs
+++ b/modules/fedimint-mint-client/src/backup/recovery.rs
@@ -1,6 +1,6 @@
-use std::cmp::max;
+use std::cmp::{self, max};
 use std::collections::BTreeMap;
-use std::fmt;
+use std::{fmt, ops};
 
 use fedimint_client::sm::{State, StateTransition};
 use fedimint_client::DynGlobalClientContext;
@@ -16,6 +16,7 @@ use fedimint_core::{Amount, NumPeers, OutPoint, PeerId, Tiered, TieredMulti};
 use fedimint_derive_secret::DerivableSecret;
 use fedimint_logging::LOG_CLIENT_RECOVERY_MINT;
 use fedimint_mint_common::{MintInput, MintOutput, Nonce};
+use futures::StreamExt as _;
 use serde::{Deserialize, Serialize};
 use tbs::{AggregatePublicKey, BlindedMessage, PublicKeyShare};
 use threshold_crypto::G1Affine;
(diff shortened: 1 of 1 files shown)


modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -246,20 +247,32 @@ impl MintRestoreInProgressState {
         decoders: ModuleDecoderRegistry,
         secret: DerivableSecret,
     ) -> Self {
+        const PROGRESS_SNAPSHOT_BLOCKS: u64 = 50;
```
dpc, line 250: Hard to pick a good value here. On idle federations this will be quite small, on very busy ones it could be a lot.

Recovery refactoring should render this irrelevant, so I'm siding with 'go fast' for now.

dpc: This constant was picked by multiplying previous (implicit) `1` by a logarithm of "recovery takes hours" reported by some users.

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -246,20 +247,32 @@ impl MintRestoreInProgressState {
...
+
         assert_eq!(secret.level(), 2);
 
-        info!(target: LOG_CLIENT_RECOVERY_MINT, "Processing block {}", self.next_epoch);
+        let block_range = self.next_epoch
+            ..cmp::min(
+                self.next_epoch.wrapping_add(PROGRESS_SNAPSHOT_BLOCKS),
+                self.end_epoch,
+            );
```
elsirion, line 258: This is kinda ugly syntax :/

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -246,20 +247,32 @@ impl MintRestoreInProgressState {
...
     ) -> Self {
+        const PROGRESS_SNAPSHOT_BLOCKS: u64 = 50;
+
         assert_eq!(secret.level(), 2);
 
-        info!(target: LOG_CLIENT_RECOVERY_MINT, "Processing block {}", self.next_epoch);
+        let block_range = self.next_epoch
+            ..cmp::min(
+                self.next_epoch.wrapping_add(PROGRESS_SNAPSHOT_BLOCKS),
+                self.end_epoch,
```
elsirion, line 257: Not including `self.end_epoch` *looked* like an off-by-one to me initially, but looking into the current state machine code it appears to not change the existing behavior.

elsirion (approved):

modules/fedimint-mint-client/src/backup/recovery.rs
dpc: In what sense?

elsirion: A range that's defined over multiple lines seems a bit unusual, but it's only a nit.

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -270,20 +283,32 @@ impl MintRestoreInProgressState {
...
+                            Err(e) => {
+                                info!(e = %e, block_idx, "Error trying to fetch signed block");
+                            }
+                        }
+                    };
+
+                    (block_idx, block)
+                })
+            })
+            .buffered(8)
```
Maan2003, line 311: what is optimal value for this? maybe better to send more requests at same time

elsirion: Good catch, at least this should be a constant that has some docs about what it does. Maybe one day we'll even make it "configurable" (more like auto-detecting device capabilities and adapting it using that).

dpc:
I have this functionality ready in a new refactor, so I don't want to bother with merging this. Will backport manually.

dpc:
This (along with some requests for tweaks) is in https://github.com/fedimint/fedimint/pull/4042 (only better, because the stream doesn't need to get recreated).

Issue #4013, opened 2024-01-04
Re-register gateway after updating the configuration


m1sterc001guy:
As @TonyGiorgio pointed out [here](https://github.com/fedimint/fedimint/issues/3990#issuecomment-1868214852), when updating a gateway's routing fees, the operator either needs to wait for ~10 minutes for the gateway to re-register, or leave the federation and re-join for it to show up immediately. This is a bad experience and unnecessary, config changes should just trigger re-registration with all of the connected federations.

elsirion:
I'm not even sure if it's *only* a problem with registration. Last time I changed fee levels re-registering didn't help because the global config that can be changed via the CLI or env variables is duplicated into the per-federation client config, re-joining helped. I think the way we handle malleable gateway config has to change fundamentally.

m1sterc001guy:
> Last time I changed fee levels re-registering didn't help because the global config that can be changed via the CLI or env variables is duplicated into the per-federation client config

Good catch, this is just a bug imo, leftover from before we had a database for the gateway. Gateway fees are a gateway-level setting, there shouldn't be any need to replicate it into each individual client db.

I think the fix is pretty easy, unfortunately it will require a database migration (currently working on this) if we want to remove it from the client databases. 

> I think the way we handle malleable gateway config has to change fundamentally

Let me know if you have additional thoughts on this. My intention with the original design was to have a simple API that stored gateway-level configs in the gateway's db, which could just be re-read at runtime. It wasn't intended for per-federation client level configs.

elsirion:
I think long-term we will want some per-federation config options because gateway operators may charge higher fees for less-active or higher-risk federations.

m1sterc001guy:
Sure, I think we can still support that inside `gatewayd`'s database. I don't think those kind of settings need to be replicated into each individual client's db.

Issue #3961, opened 2023-12-15
Gateway can charge separate fee rates on different federations


justinmoon:
The fee charged by a gateway reflects the custodial risk the gateway operator associates with holding ecash at a given mint. Given that this risk can be very different with different mints, `gatewayd` should have a way to charge different fees for different federations.

elsirion:
Generally how to set configuration requires an overhaul (changing the fee rate for a already-connected federation is quite broken), that's a prerequisite for fixing this issue imo.

fparisotto:
I'll like to fix this issue, I'm building a [PR (draft for now)](https://github.com/fedimint/fedimint/pull/4679), but still missing tests (I'll add than in my next commits).
I'm 100% sure if I'm on the right track, any help is more than welcome

m1sterc001guy:
I left some feedback, thanks for tackling it! Let me know if you have any questions

justinmoon:
Can we close @m1sterc001guy ?

elsirion:
Closing since the PR got merged.

PR #4020, opened 2024-01-08
fix: re-register gateway when routing fees are updated

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/4013

This change updates the `FederationConfig` for each federation in the gateway and also re-registers the lightning gateway when the routing fees are updated. 

This PR also gets us most of the way to per-federation routing fees: https://github.com/fedimint/fedimint/issues/3961


Kodylow (approved):
Looks good, ran locally and test passes.

okjodom (approved):
tACK

PR #4023, opened 2024-01-09
chore(ci): upload release artifacts on tag builds

dpc:
Fix:

image

"upload artifacts" we already have is not for uploading release artifacts.

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -397,6 +397,12 @@ jobs:
           name: "${{ matrix.build.flake-output}}-linux-x86_64"
           path: "bins/**"
 
+      - name: Release Binaries
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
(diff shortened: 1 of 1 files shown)


elsirion:
I think this will create the GH release by itself? That way we don't have as much control over the release text etc. and might forget to fix it up later. Could we at least add a good automatic change log generator because this will prevent me from using GitHub's builtin one?

dpc:
> I think this will create the GH release by itself? That way we don't have as much control over the release text etc. and might forget to fix it up later. Could we at least add a good automatic change log generator because this will prevent me from using GitHub's builtin one?

I'm not sure about the exact behavior, but releases can be edited after the fact, even deleted , set as pre-release, etc. so I don't think it's a big deal.

elsirion (approved):

m1sterc001guy (approved):
I don't have much experience with this myself but looks good as far as I can tell

fedimint-backports:
Git push to origin failed for releases/v0.2 with exitcode 1

Issue #4014, opened 2024-01-05
Cancelled LN payment update stream doesn't yield `Cancelled` state


elsirion:
See #4001

Repro (gets stuck):

```rust
#[tokio::test(flavor = "multi_thread")]
async fn gateway_cancels_payment() -> anyhow::Result<()> {
    let fixtures = fixtures();
    let fed = fixtures.new_fed().await;
    let client = fed.new_client().await;
    let gw = gateway(&fixtures, &fed).await;
    let dummy_module = client.get_first_module::<DummyClientModule>();

    // Print money for client
    let (op, outpoint) = dummy_module.print_money(sats(1000)).await?;
    client.await_primary_module_output(op, outpoint).await?;

    let cln = fixtures.cln().await;
    let invoice = Bolt11Invoice::from_str("lnbcrt420n1pje07kcdqqpp5vjadfxk9sxar4030cdh3ws4n9rfc2r6gzsrhrrysnc2ax75zfa7ssp53jaxxultl94h4an04a2rgv76xlgt3pjgy4txnd8h8mg8pdhumtlq9qrsgqcqpjnp4qwljl52cc7vnn76dmaau8qptly6q2uh3rl8d0gdn047wxxerf799yxqyz5vqrzjqt5a8fp65nlccjnyzhtfj82dw20pshv9wxlwt2j0u69snyypgpqvcqqqqqqqqqqqqyqqqqqqqqqqqqqqrcr9yqd332tteyjmz53r7k9a3yr86gmfgwdzxrv05h8mhkf0tpwz6d7j4sqqqwqqqqqgqqyqqqqlgqqqqqqgq2qpwn5ay82j0lrz2vs2adxgaf4efuxzas4cmaed2flngkzvss9qypnqqqqqqqqqqqqqsqqqqqqqqqqqqqq0qtuj76w6vuuxka7wg4ve357r8pdgh0msxpj9f0kqypc4k7g03uyl9v6sceh4hsdrv40pw3z8l4vz8vvz70zy3uyn5xys96c6clw6ksaqq5eax9q").expect("Valid invoice");

    let OutgoingLightningPayment {
        payment_type,
        contract_id: _,
        fee: _,
    } = pay_invoice(&client, invoice.clone()).await?;
    match payment_type {
        PayType::Lightning(operation_id) => {
            let mut sub = client
                .get_first_module::<LightningClientModule>()
                .subscribe_ln_pay(operation_id)
                .await?
                .into_stream();

[...]

PR #4024, opened 2024-01-09
refactor: remove refundable state

m1sterc001guy:
See https://github.com/fedimint/fedimint/issues/4014 for motivation.

I believe the `Refundable` state is unnecessary in the lightning client state machine and can be removed. This eliminates a state transition, but also eliminates the arbitrary 30 second timeout on payments.


justinmoon:
What kind of migration would be required for this? If a client is in `Refundable`, can we just stick them in another state?

m1sterc001guy:
> What kind of migration would be required for this? If a client is in Refundable, can we just stick them in another state?

It should be possible to transition the old client back from `Refundable` to `Funded`, since `Funded` has a superset of transitions that `Refundable` did (i.e the transitions that `Refundable` used to have are now moved to `Funded`). 

Once client side migrations have been implemented of course :) 

I think the PR is a bit mature since client side migrations are not implemented yet, but wanted to give an example of what I was thinking to fix this problem https://github.com/fedimint/fedimint/issues/4014. I think this should help with some of the problems the Mutiny guys are hitting.

elsirion (commented):
Now I'm wondering how we'll determine/show that a payment will likely be refunded. Getting an error from the GW but no cancellation doesn't tell us anything, could still be processed, but it's very unlikely. Maybe we can add a "likely to be refunded" flag to the `Funded` state and set it when the GW is doing things it shouldn't do? It wouldn't influence the SM, but could be used to give users more clarity.

m1sterc001guy:
> Maybe we can add a "likely to be refunded" flag to the Funded state and set it when the GW is doing things it shouldn't do? It wouldn't influence the SM, but could be used to give users more clarity.

Yep, good point. Perhaps the refund height can go here too. Like you said, I think ultimately the `cancelled` flag on the contract should be the source of truth, but we can perhaps give the end user some additional info on what is happening. 

I'm going to put this aside for the moment and try to finish state machine db migrations so that this can be a backwards-compatible change.

m1sterc001guy:
Note: @joschisan's NG lightning implementation does the same thing as this PR https://github.com/fedimint/fedimint/pull/3468

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -146,20 +145,11 @@ pub enum LnPayState {
     Created,
     Canceled,
     Funded,
-    WaitingForRefund {
-        block_height: u32,
```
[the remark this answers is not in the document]
m1sterc001guy: It is available on the `Funded` state now.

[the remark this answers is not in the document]
m1sterc001guy: This is available in the `Funded` state now

m1sterc001guy:
This should be ready for review. I think this makes our lightning payments simpler.

I tried to do this in the most backwards compatible way possible, so I left some states in enums that can be cleaned up later. I wrote a db migration to add an `error_reason` to the `Refund` state, but if we won't want that, the db migration isn't required.

Writing a test for the db migration is currently kind of difficult since we need to keep old structs around. https://github.com/fedimint/fedimint/issues/4536

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -143,21 +142,12 @@ pub enum InternalPayState {
 pub enum LnPayState {
     Created,
     Canceled,
-    Funded,
-    WaitingForRefund {
-        block_height: u32,
-        gateway_error: GatewayPayError,
-    },
+    Funded { block_height: u32 },
```
elsirion, line 145: I think this change also needs a DB migration since old `Funded` states will be 0 bytes in length while the new parsing code will expect 4 bytes.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1279,49 +1221,65 @@ impl LightningClientModule {
...
-                                    return;
+                let mut stream = self_ref.notifier.subscribe(operation_id).await;
+
+                loop {
+                    match stream.next().await {
+                        Some(LightningClientStateMachines::LightningPay(state)) => match state.state {
+                            LightningPayStates::Canceled => {
+                                yield LnPayState::Canceled;
+                                return;
+                            }
```
elsirion, line 1232: The canceled terminology here is very confusing: it means "funding tx rejected by federation" while I first assumed it was "gateway refunded us" (we use the term "cancel contract" there too). My fault, not yours.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1279,49 +1221,65 @@ impl LightningClientModule {
...
-                        if let Ok(out_points) = self_ref.await_refund(operation_id).await {
-                            // need to await primary module to get refund
-                            if client_ctx.await_primary_module_outputs(operation_id, out_points).await.is_ok() {
-                                yield LnPayState::Refunded { gateway_error: error };
                                 return;
                             }
+                            _ => {}
                         }
+                        Some(_) => {}
+                        None => {}
```
elsirion, line 1279: nit: should debug-log that stream closed when receiving `None`. This isn't necessarily unexpected on shutdown, but if it happens during normal operation it could indicate a problem.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1279,49 +1221,65 @@ impl LightningClientModule {
...
 
-                        if let Ok(out_points) = self_ref.await_refund(operation_id).await {
-                            // need to await primary module to get refund
-                            if client_ctx.await_primary_module_outputs(operation_id, out_points).await.is_ok() {
-                                yield LnPayState::Refunded { gateway_error: error };
                                 return;
                             }
+                            _ => {}
                         }
+                        Some(_) => {}
```
elsirion, line 1278: Can we ever receive a non-pay state under the same operation id here? If not let's panic.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1279,49 +1221,65 @@ impl LightningClientModule {
...
+                                    }
+                                }
 
-                        if let Ok(out_points) = self_ref.await_refund(operation_id).await {
-                            // need to await primary module to get refund
-                            if client_ctx.await_primary_module_outputs(operation_id, out_points).await.is_ok() {
-                                yield LnPayState::Refunded { gateway_error: error };
                                 return;
                             }
+                            _ => {}
```
elsirion, line 1276: nit: make the remaining variants explicit, so when adding variants there is a compile error here.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1279,49 +1221,65 @@ impl LightningClientModule {
...
-                            match client_ctx.await_primary_module_outputs(operation_id, change).await {
-                                Ok(_) => {}
-                                Err(_) => {
-                                    yield LnPayState::UnexpectedError { error_message: "Error occurred while waiting for the primary module's output".to_string() };
-                                    return;
+                let mut stream = self_ref.notifier.subscribe(operation_id).await;
+
+                loop {
+                    match stream.next().await {
+                        Some(LightningClientStateMachines::LightningPay(state)) => match state.state {
```
elsirion, line 1228: This being a `loop` makes me slightly nervous. This means we don't enforce the correct order of events anymore and might miss implementation bugs that way. I think fixing the bug that prompted this PR is more important and I'm thinking about an abstraction that lets us model these high-level state graphs with less boilerplate while enforcing all invariants.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -52,8 +49,11 @@ pub enum LightningPayStates {
     Canceled,
     Funded(LightningPayFunded),
     Success(String),
+    // Deprecated: SM skips over this state now and will retry payments until cancellation or
+    // timeout
     Refundable(LightningPayRefundable),
     Refund(LightningPayRefund),
+    // Deprecated: No need to wait for the refund tx to be accepted
```
elsirion, line 56: Can we annotate with `#[deprecated]` here?

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -268,49 +270,91 @@ impl LightningPayFunded {
...
     ) -> Result<String, GatewayPayError> {
-        // Abort the payment if we can't reach the gateway within 30 seconds
-        // to prevent unexpected delays for the user.
-        let deadline = fedimint_core::time::now() + GATEWAY_API_TIMEOUT;
-
-        let mut last_error = None;
-        while fedimint_core::time::now() < deadline {
+        // Retry until the payment is successful. This future will be cancelled if
+        // the contract is cancelled.
+        loop {
```
elsirion, line 330: We don't necessarily want to retry for the entire timeout period. Chances are that the user figured out another way to pay by then.

If the timeout is a few days then it would be quite unexpected for the user that a transaction still goes through a day later or so. This can, in principle, happen anyway if the GW goes offline at the exactly wrong time, but will be very rare, this change makes it more likely imo.

elsirion (commented):
One missing migration bug, looks mostly good otherwise :)

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Is this ever stored in the database? It's not `Encodable`, I thought it's just yielded during `subscribe_ln_receive`

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: We could change the name. Maybe `FundingRejected`?

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: Right now we re-try every 30 seconds. Maybe we should expose a timeout parameter? It seems reasonable to me to wait a few minutes but I'm not sure if every wallet will want the same value. 

After the timeout crosses we could sleep until past the block timeout.

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: I also don't like the loop here. I re-wrote this without a loop that enforces ordering. Hopefully this is more intuitive.

modules/fedimint-ln-client/src/pay.rs
elsirion: The timeout path seems to be a separate state transition, so we could just do

```rust
std::future::pending().await;
```

here, which blocks forever.

m1sterc001guy: I added a timeout parameter of 180 seconds and changed the re-try interval to 10 seconds between retries. So if the gateway is failing unexpectedly, the client will try 18 times for a total of 3 minutes before giving up and blocking until the block timeout. Let me know if you think these guesses are unreasonable, I can change them.

joschisan (approved):
Resulting state machine looks good, its now similar to the current implementation of lightning ng.

elsirion (approved):

elsirion:
This will require cleanup after the fact to get rid of no longer used states after #4536 lands.

elsirion:
Also, I vote against backporting this. It fixes a minor bug but is a rather complex change that could massively delay 0.3.

m1sterc001guy:
If we're not going to backport, we will need to wait to merge until after we fix https://github.com/fedimint/fedimint/issues/4700. Otherwise it will make fixing that issue painful

elsirion:
@m1sterc001guy want to merge it now? The #4700 fix doesn't seem to have resulted in any conflicts.

justinmoon:
dev call: don't put it into 0.3.0. Try to backport to 0.3.1. @m1sterc001guy going to remove need to duplicate enums before merging.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -247,7 +256,7 @@ impl LightningPayCreatedOutgoingLnContract {
 pub struct LightningPayFunded {
     payload: PayInvoicePayload,
     gateway: LightningGateway,
-    timelock: u32,
+    pub timelock: u32,
 }
```
elsirion, line 260: Since we are doing another iteration on this we should include the creation/funding time here. Both are close enough to each other normally, so I'd go with the funding time (avoids changing the created state).

That way the trigger future trying to contact the gateway stays inactive after the first n minutes of trying. Currently the code retries to reach the gateway for `TIMEOUT_DURATION` after each restart.

m1sterc001guy:
Going to work on https://github.com/fedimint/fedimint/issues/4823, since this will make the migrations easier to write.

modules/fedimint-ln-client/src/pay.rs
m1sterc001guy: Fixed. Added the `funding_time` to the funded state

m1sterc001guy:
This is finally ready for another round of review.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -170,22 +169,13 @@ pub enum InternalPayState {
...
-    },
-    Refunded {
-        gateway_error: GatewayPayError,
-    },
-    UnexpectedError {
-        error_message: String,
-    },
+    Success { preimage: String },
+    Refunded { error_reason: String },
+    UnexpectedError { error_message: String },
```
elsirion, line 178: We had this discussion before, and you are right that we never save this state in the DB consensus encoded, but we do save *final outcomes* in the operation log as JSON.

So I think we need to introduce an alias for `FundingRejected` since it might be saved as `canceled` in the OpLog and give the `Refunded` variant a default error message since the old one has the `GatewayPayError`, which we no longer parse. Eventually we'll want to add migrations for this too.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1185,55 +1128,97 @@ impl LightningClientModule {
...
+                                };
                             }
                         }
                     }
-                    _ => {}
+                    Some(state) => {
+                        yield LnPayState::UnexpectedError { error_message: format!("Found unexpected state during lightning payment: {state:?}") };
+                    }
+                    None => {
+                        error!("Unexpected end of lightning pay state machine");
```
elsirion, line 1219: Should this panic or return an `UnexpetedError`? Otherwise we'll save the last returned state as the final outcome.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -85,19 +92,19 @@ impl State for LightningPayStateMachine {
...
+                funded.transitions(self.common.clone(), context.clone(), global_context.clone())
             }
             LightningPayStates::Success(_) => {
                 vec![]
             }
             LightningPayStates::Refundable(refundable) => {
                 refundable.transitions(self.common.clone(), global_context.clone())
             }
-            LightningPayStates::Refund(refund) => refund.transitions(global_context),
+            LightningPayStates::Refund(_) => vec![],
```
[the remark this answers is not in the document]
m1sterc001guy: The only thing the `Refund` state did previously was transition to `Refunded` once the transaction was accepted. So it was really just a indicator to `subscribe_ln_pay` that the refund was accepted. We have the same logic now in `subscribe_ln_pay` when `await_primary_module_outputs` is called. 

I believe this is safe. If a user is in this state while upgrading, they will no longer transition to `Refunded`, but `subscribe_ln_pay` still has logic to know when the refund tx has been accepted.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -48,11 +47,19 @@ const RETRY_DELAY: Duration = Duration::from_secs(1);
...
 pub enum LightningPayStates {
     CreatedOutgoingLnContract(LightningPayCreatedOutgoingLnContract),
-    Canceled,
+    FundingRejected,
     Funded(LightningPayFunded),
     Success(String),
+    #[deprecated(
+        since = "0.4.0",
+        note = "Pay State Machine skips over this state and will retry payments until cancellation or timeout"
+    )]
```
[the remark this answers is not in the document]
m1sterc001guy: Yeah its to save migration code. `Refunded` can be deleted, but then we'd need to write a migration for the `Failure` state. `Refundable` can also be deleted, but then we'd need to add another migration, and since it has the same logic as `Funded` now, I think its easier just to let the state machine execute and take care of the transition.

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Ah I see. Should I change `FundingRejected` back to `Canceled`? That change is mostly just style, I thought it made more sense, but if it's going to cause issues it probably makes more sense to just keep it as `Canceled`.

How do we add migrations for these states? Would that involve reading the OpLog and re-writing the JSON with it encoded as the new states?

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: I think it should return `UnexpectedError`

modules/fedimint-ln-client/src/lib.rs
elsirion: > How do we add migrations for these states? Would that involve reading the OpLog and re-writing the JSON with it encoded as the new states?

Yes, I fear so :( but in this case we can probably work around it by having a default error message (and JSON already ignores superfluous fields).

m1sterc001guy: Hmm yeah this is more change than I think I want to do with this PR. I'll see if I can change this back so that it is backwards compatible.

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Fixed

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -385,160 +450,117 @@ impl LightningPayRefundable {
...
-                .module_api()
-                .wait_outgoing_contract_cancelled(contract_id)
-                .await
-            {
-                Ok(_) => return,
-                Err(error) => {
-                    error!("Error waiting for outgoing contract to be cancelled: {error:?}");
-                }
+/// Waits for a contract with `contract_id` to be cancelled by the gateway.
+async fn await_contract_cancellable(
```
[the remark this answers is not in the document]
m1sterc001guy: Fixed

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: I changed `FundingRejected` back to `Cancelled` and `Refunded` to contain `GatewayPayError` again. Now only `Funded` and `WaitingForRefund` have new fields.

For my understanding, what is the impact of breaking the JSON in the operation log? Will that impact downstream clients that parse it?

elsirion (approved):
I *think* these fixes make this change good, but would love some more people to look at it with a very critical eye towards upgradability/compatability.

PR #4025, opened 2024-01-09
refactor: 0.2.1, and 0.1 docker scripts

Kodylow:
Splits docker organization into 2 folder: 

0.1: all the old stuff
0.2: all the new stuff, including fedimint-thunderhub and bumped version of lnd

Updates install scripts to ask which version the user wants before downloading, 0.1 or 0.2


elsirion (commented):
We should heavily discourage v0.1 and not offer it in an easy to run script. With future versions that will be different due to stability guarantees, but 0.1 will be incompatible with any state of the art Fedimint wallet.

Kodylow:
kk i was tracking to maintain 0.1, I'll adjust it and make separate legacy scripts

Kodylow:
Updated to always install 0.2.1 from the scripts, and renamed 0.1 to "legacy"

elsirion (approved):
I played around with the mutiny install script and it seems to work, lgtm. Thx for updating it!

dpc (approved):

PR #4026, opened 2024-01-11
refactor: Only load federation clients when gateway boots

m1sterc001guy:
See https://github.com/fedimint/fedimint/pull/3781#issuecomment-1885878214 for context.

- Removes `DatabaseSource`
- Adds check when connecting a federation if the federation has already been connected
- Wraps every usage of `lnrpc` behind a wrapper function that first checks if the lightning node is connected to the gateway.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -56,6 +56,7 @@ use fedimint_wallet_client::{
     WalletClientInit, WalletClientModule, WalletCommonInit, WithdrawState,
 };
 use futures::stream::StreamExt;
+use futures::Future;
 use gateway_lnrpc::intercept_htlc_response::Action;
 use gateway_lnrpc::{GetNodeInfoResponse, InterceptHtlcResponse};
 use lightning_invoice::RoutingFees;
@@ -256,6 +257,21 @@ pub struct Gateway {
     channel_id_generator: Arc<Mutex<AtomicU64>>,
 }
 
+impl std::fmt::Debug for Gateway {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Gateway")
+            .field("gateway_parameters", &self.gateway_parameters)
+            .field("state", &self.state)
+            .field("client_builder", &self.client_builder)
+            .field("gateway_db", &self.gateway_db)
+            .field("clients", &self.clients)
+            .field("scid_to_federation", &self.scid_to_federation)
+            .field("gateway_id", &self.gateway_id)
+            .field("channel_id_generator", &self.channel_id_generator)
+            .finish()
+    }
+}
+
 impl Gateway {
     #[allow(clippy::too_many_arguments)]
     pub async fn new_with_custom_registry(
@@ -399,6 +415,8 @@ impl Gateway {
 
     pub async fn run(mut self, tg: &mut TaskGroup) -> anyhow::Result<TaskShutdownToken> {
         self.start_webserver(tg).await;
+        self.register_clients_timer(tg).await;
+        self.load_clients().await;
         self.start_gateway(tg).await?;
         let handle = tg.make_handle();
         let shutdown_receiver = handle.make_shutdown_rx().await;
@@ -493,15 +511,6 @@ impl Gateway {
                                                 continue;
                                             }
                                         }
-
-                                        self.register_clients_timer(&mut htlc_task_group).await;
-                                        self.load_clients(
-                                            ln_client.clone(),
-                                            lightning_public_key,
-                                            lightning_alias.clone()
-                                        )
-                                        .await
-                                        .expect("Failed to load gateway clients");
 
                                         info!("Successfully loaded Gateway clients.");
                                         self.set_gateway_state(GatewayState::Running {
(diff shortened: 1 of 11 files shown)


dpc (commented):
LGTM, but someone that understands lngateway should look.

okjodom (approved):
reviewed and tested https://github.com/fedimint/fedimint/pull/4026/commits/1588620d87328153201d2862dd5917e40c4208af looks good to me

gateway/ln-gateway/src/lib.rs
```diff
@@ -256,6 +257,21 @@ pub struct Gateway {
     channel_id_generator: Arc<Mutex<AtomicU64>>,
```
elsirion, line 257: A `Mutex` around an atomic type seems quite useless.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1154,60 +1165,38 @@ impl Gateway {
...
-                        all_clients,
-                        all_scids,
-                        old_client,
-                        self.gateway_db.clone(),
-                    )
+    async fn load_clients(&mut self) {
+        let dbtx = self.gateway_db.begin_transaction().await;
+        let configs = self.client_builder.load_configs(dbtx.into_nc()).await;
+        let channel_id_generator = self.channel_id_generator.lock().await;
+        let mut next_channel_id = channel_id_generator.load(Ordering::SeqCst);
```
elsirion, line 1179: nit: locking+load/store is redundant, should only lock since the operation is too complex to be expressed in one atomic operation.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1349,6 +1338,32 @@ impl Gateway {
...
+                    lightning_alias,
+                    lightning_network,
+                )
+                .await
+            }
+            _ => {
+                anyhow::bail!("Lightning node is not connected");
+            }
+        }
+    }
```
elsirion, line 1373: This seems overly complicated, why not have a function that returns a `Result<RunningContext, anyhow::Error>` where `RunningContext` is the 4 fn arguments? Then that could be `?`-ed and used and result in the same semantics: early return with error if not running.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Removed `AtomicU64` and replaced with `u64`. Also added overflow logic when registering a federation and removed the long standing/annoying TODO.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Removed `AtomicU64`. We still lock here.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Replaced `execute_with_lightning_connection` with `get_lightning_context`

okjodom (approved):

gateway/ln-gateway/src/lib.rs
```diff
@@ -253,7 +256,22 @@ pub struct Gateway {
 
     // ID generator that atomically increments. Used for creation of new short channel ids that
     // represent federations.
-    channel_id_generator: Arc<Mutex<AtomicU64>>,
+    channel_id_generator: Arc<Mutex<u64>>,
```
douglaz, line 259: Perhaps
```suggestion
    channel_id_generator: Arc<AtomicU64>,
```

gateway/ln-gateway/src/state_machine/complete.rs
```diff
@@ -174,28 +177,40 @@ impl CompleteHtlcState {
...
+
+                    lightning_context
+                        .lnrpc
+                        .complete_htlc(htlc)
+                        .await
+                        .map_err(|_| CompleteHtlcError::FailedToCompleteHtlc)?;
+                    return Ok(());
+                }
+                Err(_) => {
+                    warn!("Trying to complete HTLC but the lightning node is not connected");
```
douglaz, line 209: In fact you shouldn't make the assumption the error is about disconnection. Just make it generic:
```suggestion
                Err(e) => {
                    warn!("Trying to complete HTLC but got {e}, will keep retrying...");
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -1349,6 +1335,18 @@ impl Gateway {
...
     }
+
+    /// Checks the Gateway's current state and returns the proper
+    /// `LightningContext` if it is available. Sometimes the lightning node
+    /// will not be connected and this will return an error.
+    pub async fn get_lightning_context(&self) -> anyhow::Result<LightningContext> {
+        match self.state.read().await.clone() {
+            GatewayState::Running { lightning_context } => Ok(lightning_context),
+            _ => {
+                anyhow::bail!("Lightning node is not connected");
```
douglaz, line 1346: perhaps return `LightningRpcError::FailedToConnect` here?

gateway/ln-gateway/src/state_machine/complete.rs
m1sterc001guy: Done

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Changed

elsirion (approved):

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -371,9 +371,21 @@ impl GatewayPayInvoice {
                 .amount()
                 .expect("We already checked that an amount was supplied");
 
+        let lightning_context = match context
+            .gateway
+            .get_lightning_context()
+            .await
+            .map_err(|_| LightningRpcError::FailedToConnect)
```
douglaz, line 378: Not necessary anymore

douglaz (approved):
LGTM

dpc (approved):

Issue #3924, opened 2023-12-12
Add license and README link to each `Cargo.toml`


elsirion:
The workspace-wide one doesn't seem to make it into the released crates:

![Screenshot 2023-12-12 at 17-47-31 fedimint-aead - crates io Rust Package Registry](https://github.com/fedimint/fedimint/assets/84478054/0d749ded-4b5c-4003-9243-5d64055d76cd)

elsirion:
Need to check if relative paths outside the crate but inside the workspace work.

tvolk131:
Anything else that needs to be done for this? Looks complete.

PR #4030, opened 2024-01-11
chore: add Readme.md links in each Cargo.toml

15IITian:
Added Readme.md links in those Crate.toml where there exists a README.md file in that individual crate 
#3924

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ keywords = ["bitcoin", "lightning", "chaumian", "e-cash", "federated"]
 threshold_crypto = { version = "0.1", package = "fedimint-threshold-crypto" }
 tonic_lnd = { version = "0.1.3", package="fedimint-tonic-lnd", features = ["lightningrpc", "routerrpc"] }
 cln-rpc = { package = "fedimint-cln-rpc", version = "0.4.0" }
+fedimint-aead = { version = "0.0.1" }
 
 # Workaround: https://github.com/rust-lang/cargo/issues/12457 which causes
 #             https://github.com/ipetkov/crane/issues/370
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -61,6 +61,7 @@ keywords = ["bitcoin", "lightning", "chaumian", "e-cash", "federated"]
 threshold_crypto = { version = "0.1", package = "fedimint-threshold-crypto" }
 tonic_lnd = { version = "0.1.3", package="fedimint-tonic-lnd", features = ["lightningrpc", "routerrpc"] }
 cln-rpc = { package = "fedimint-cln-rpc", version = "0.4.0" }
+fedimint-aead = { version = "0.0.1" }
```
dpc, line 64: ?

15IITian: By Mistake, I thought that I have to include this line in root Cargo.toml file as per the screenshot :)

15IITian:
image

Not getting why these links are dead as I have not made any changes in it & they are working fine in the Readme.md file .

elsirion (approved):
Doesn't fully solve https://github.com/fedimint/fedimint/issues/3924 since crates without README still don't have any linked, but it's an improvement.

15IITian:
> Doesn't fully solve #3924 since crates without README still don't have any linked, but it's an improvement.

For the crates not having their own README -> I have to give link of the root Readme file?

elsirion:
> For the crates not having their own README -> I have to give link of the root Readme file?

I think that's a complicated problem since the root README isn't included when publishing sub-crates by default (see my comment in the issue). So this requires some Cargo wizardry, which is probably not worth it rn.

15IITian:
> > For the crates not having their own README -> I have to give link of the root Readme file?
> 
> I think that's a complicated problem since the root README isn't included when publishing sub-crates by default (see my comment in the issue). So this requires some Cargo wizardry, which is probably not worth it rn.

Okk,
I think that's the same problem in the License field case in each Cargo.toml?

15IITian:
@elsirion
 For testing,  I have published one sub_crate  named `another_trial`  from  a workspace , which do not has its own README file ->  thus has relative path of the root README file-> and that's working fine as required ->
 
image

image

dpc (approved):

elsirion:
Interesting, then let's try this in another PR!

PR #4035, opened 2024-01-13
chore: client module recovery refactor

dpc:
Re #2977

This is a initial step towards client module recovery system described/agreed on in #2977.

### Most notable points of the new design:

Module recoveries work as just standalone function (future): `ClientModuleInit::recover` called instead of usual `ClientModuleInit::init`. This makes module job easier (in a sense at least), as it is not constantly being interrupted (will help with efficient streaming through blocks). It's up to the implementation to periodically save state and send progress updates, but a follow up changes will introduce helper functions to set the module implementation on the right track. It also avoids saving tons of in-progress data in the state machine execuation log. Conceptually module recovery is not a client state machine.

Module recovery state is persisted and tracked and modules become available as they complete their recovery (on the next client start). A slow module will not block recovery of other modules.

### Follow-up work

* [ ] Mint recovery should be refactored into a framework-function that will take care of saving progress and sending updates (a method on `ClientmoduleRecoverArgs`)
* [x] Add streaming of blocks like in #4019, but this time for the whole range all at once (should lead to much better performance): https://github.com/fedimint/fedimint/pull/4042
* [x] Caching of `api.await_block` in a transparent LRU: https://github.com/fedimint/fedimint/pull/4080

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -838,16 +838,6 @@ impl Client {
         self.executor.add_state_machines_dbtx(dbtx, states).await
     }
 
-    pub async fn add_state_machines_inactive(
-        &self,
-        dbtx: &mut DatabaseTransaction<'_>,
-        states: Vec<DynState<DynGlobalClientContext>>,
-    ) -> AddStateMachinesResult {
-        self.executor
-            .add_state_machines_inactive_dbtx(dbtx, states)
-            .await
-    }
-
     // TODO: implement as part of [`OperationLog`]
     pub async fn get_active_operations(&self) -> HashSet<OperationId> {
         let active_states = self.executor.get_active_states().await;
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -70,15 +70,18 @@ use std::cmp::Ordering;
 use std::collections::{BTreeMap, HashSet};
 use std::fmt::{Debug, Formatter};
 use std::ops::{self, Range};
+use std::pin::Pin;
 use std::sync::atomic::AtomicUsize;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
 
 use anyhow::{anyhow, bail, ensure, Context};
 use async_stream::stream;
+use backup::ClientBackup;
 use db::{
     CachedApiVersionSet, CachedApiVersionSetKey, ClientConfigKey, ClientConfigKeyPrefix,
-    ClientInviteCodeKey, ClientInviteCodeKeyPrefix, EncodedClientSecretKey,
+    ClientInitStateKey, ClientInviteCodeKey, ClientInviteCodeKeyPrefix, ClientModuleRecovery,
+    EncodedClientSecretKey, InitMode,
 };
 use fedimint_core::api::{
     ApiVersionSet, DynGlobalApi, DynModuleApi, GlobalFederationApi, IGlobalFederationApi,
@@ -104,13 +107,14 @@ use fedimint_core::task::{sleep, MaybeSend, MaybeSync, TaskGroup};
 use fedimint_core::transaction::Transaction;
 use fedimint_core::util::{BoxStream, NextOrPending};
 use fedimint_core::{
-    apply, async_trait_maybe_send, dyn_newtype_define, maybe_add_send_sync, Amount, OutPoint,
-    TransactionId,
+    apply, async_trait_maybe_send, dyn_newtype_define, maybe_add_send, maybe_add_send_sync, Amount,
+    OutPoint, TransactionId,
 };
 pub use fedimint_derive_secret as derivable_secret;
 use fedimint_derive_secret::DerivableSecret;
 use fedimint_logging::LOG_CLIENT;
-use futures::StreamExt;
+use futures::{Future, StreamExt};
+use module::recovery::RecoveryProgress;
 use module::{DynClientModule, FinalClient};
 use rand::thread_rng;
 use secp256k1_zkp::{PublicKey, Secp256k1};
@@ -118,10 +122,11 @@ use secret::{DeriveableSecretClientExt, PlainRootSecretStrategy, RootSecretStrat
 use thiserror::Error;
 #[cfg(not(target_family = "wasm"))]
 use tokio::runtime::{Handle as RuntimeHandle, RuntimeFlavor};
+use tokio::sync::watch;
 use tracing::{debug, error, info, warn};
 
 use crate::backup::Metadata;
-use crate::db::OperationLogKey;
+use crate::db::{ClientMetadataKey, ClientModuleRecoveryState, InitState, OperationLogKey};
 use crate::module::init::{
     ClientModuleInit, ClientModuleInitRegistry, DynClientModuleInit, IClientModuleInit,
 };
@@ -620,6 +625,8 @@ pub struct Client {
     root_secret: DerivableSecret,
     operation_log: OperationLog,
     secp_ctx: Secp256k1<secp256k1_zkp::All>,
+
+    task_group: TaskGroup,
     /// Number of [`ClientArc`] instances using this `Client`.
     ///
     /// The `Client` struct is both used for the client itself as well as
@@ -627,6 +634,10 @@ pub struct Client {
     /// cannot rely on the reference count of the `Arc<Client>` to
     /// determine if the client should shut down.
     client_count: AtomicUsize,
+
+    /// Updates about client recovery progress
+    client_recovery_progress_receiver:
+        watch::Receiver<BTreeMap<ModuleInstanceId, RecoveryProgress>>,
 }
 
 impl Client {
@@ -1120,12 +1131,11 @@ impl Client {
     /// primary module will always be returned before any other modules (which
     /// themselves are ordered by their instance ID).
     pub fn get_first_instance(&self, module_kind: &ModuleKind) -> Option<ModuleInstanceId> {
-        if &self
+        if self
             .modules
             .get_with_kind(self.primary_module_instance)
-            .expect("must have primary module")
-            .0
-            == module_kind
+            .map(|(kind, _)| kind == module_kind)
+            .unwrap_or(false)
         {
             return Some(self.primary_module_instance);
         }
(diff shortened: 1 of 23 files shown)


fedimint-client/src/module/mod.rs
```diff
@@ -142,6 +146,10 @@ where
             .await
     }
 
+    pub fn decoders(&self) -> ModuleDecoderRegistry {
+        self.client.client.get().decoders().clone()
+    }
```
elsirion, line 162: Are the actually cheap to clone and do we need them to be owned? Iirc this is a `BTreeMap<_, Arc<_>>`, so not great/not terrible.

fedimint-client/src/module/mod.rs
```diff
@@ -279,6 +290,30 @@ where
             .await
     }
 
+    // TODO: rename `module_autocommit` to something else and make this a default
+    pub async fn module_autocommit_2<'s, 'dbtx, F, T>(
+        &'s self,
+        tx_fn: F,
+        max_attempts: Option<usize>,
+    ) -> anyhow::Result<T>
```
elsirion, line 309: Why not allow arbitrary errors to be returned from the closure?

fedimint-client/src/module/recovery.rs
```diff
@@ -102,3 +102,51 @@ impl IntoDynInstance for NoModuleBackup {
...
+
+/// Progress of the recovery
+///
+/// This includes "magic" value: if `total` is `0` the progress is "not started
+/// yet"/"empty"/"none"
+#[derive(Debug, Copy, Clone, Encodable, Decodable)]
+pub struct RecoveryProgress {
+    pub complete: u32,
+    pub total: u32,
+}
```
elsirion, line 114: What's the unit? Does it matter? I guess not, but still good to know. Unless we actively don't commit to any particular meaning and adopt a notion of "work units".

fedimint-client/src/module/recovery.rs
```diff
@@ -102,3 +102,51 @@ impl IntoDynInstance for NoModuleBackup {
...
+
+impl RecoveryProgress {
```
elsirion, line 116: Maybe a `fn to_percent(&self) -> f32` would also be useful.

modules/fedimint-mint-client/src/client_db.rs
```diff
@@ -13,6 +14,7 @@ pub enum DbKeyPrefix {
     Note = 0x20,
     NextECashNoteIndex = 0x2a,
     CancelledOOBSpend = 0x2b,
+    RestoreState = 0x2c,
```
elsirion, line 17: It had to happen eventually (in fact, it already did), still sad that we are now overloading the DB key prefixes. It's not a problem due to DB isolation, but reading raw hex encoded data will be a bit more ambiguous (e.g. if one doesn't know the instance<>kind mapping).

elsirion (commented):
First part of review, a lot to follow, but better to get it out than to forget about it.

fedimint-client/src/module/init.rs
```diff
@@ -150,6 +272,55 @@ where
...
+        api: DynGlobalApi,
+        snapshot: Option<&DynModuleBackup>,
+        progress_tx: watch::Sender<RecoveryProgress>,
+    ) -> anyhow::Result<()> {
+        let typed_cfg: &<<T as fedimint_core::module::ModuleInit>::Common as CommonModuleInit>::ClientConfig = cfg.cast()?;
+        let snapshot: Option<&<<Self as ClientModuleInit>::Module as ClientModule>::Backup> =
+            snapshot
+                .map(|s| {
+                    s.as_any().downcast_ref().ok_or_else(|| {
+                        anyhow::anyhow!("can't convert client module backup to desired type")
```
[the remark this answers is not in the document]
dpc: Probably should. I copied it from some existing function, so we do it in more places it seems.

fedimint-client/src/module/mod.rs
dpc: IIRC I had an infalliable code and type inference wanted an error type, which just annoyed me, combined with having to handle failed commit. I don't think strong about it, but the plan was to have a "easy to use" default `module_autocommit` using anyhow (since we use anyhow almost for everything anyway), and another `module_autocommit_something` for cases when someone really wants to handle custom errors / commit fail.

modules/fedimint-mint-client/src/client_db.rs
dpc: Yeah, we'll need dbdump do a good job. eg. even for keys that are not implemented print `<kind> <prefix>` or something.

fedimint-client/src/lib.rs
```diff
@@ -1356,6 +1376,201 @@ impl Client {
...
+                    recovery_sender,
+                    module_recoveries,
+                    module_recovery_progress_receivers,
+                )
+                .await
+            })
+            .await;
+    }
+
+    async fn run_module_recoveries_task(
```
Maan2003, line 1499: shouldn't there be a client.modules.register_module call after module's recovery is complete?
or we are meant to restart client after recovery is complete.

fedimint-client/src/module/recovery.rs
Maan2003, line 114: nice, apps can measure speed of recovery using updates and provide some ETA.

Maan2003: units would depend on the module

Maan2003 (commented):
LGTM, just small nits

fedimint-client/src/lib.rs
dpc: The client is already `Arc`ed, potentially working etc. It would introduce a lot of locking and complications that might be a source of mistakes (hard to reason about things when module set changes underneath), so I think the limitation that recovered modules start being usable only after next client start is a good tradeoff.

fedimint-client/src/lib.rs
```diff
@@ -1356,6 +1376,201 @@ impl Client {
...
+            .wait_for(|in_progress| {
+                !in_progress
+                    .iter()
+                    .filter(|(module_instance_id, _progress)| {
+                        self.config.modules[module_instance_id].kind == module_kind
+                    })
+                    .any(|(_id, progress)| !progress.is_done())
+            })
+            .await
+            .context("Recovery task completed, but some modules could not recover")?;
```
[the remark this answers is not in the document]
dpc: No, but I reworded the context to make it more clear.

dpc:
> units would depend on the module

Yeah. I didn't want to handle the unit explicitily as then it gets complicated (translate the units, plural/singular, etc.) so I just let the module pick the numerator and denominator are hopefully it's will be good enough UX.

fedimint-client/src/module/recovery.rs
dpc: > units would depend on the module

Yeah. I didn't want to handle the unit explicitly as then it gets complicated (translate the units, plural/singular, etc.) so I just let the module pick the numerator and denominator are hopefully it's will be good enough UX.

Can be improved later.

fedimint-client/src/module/mod.rs
dpc: Cheap enough unless proved otherwise. :)

fedimint-client/src/module/init.rs
```diff
@@ -113,6 +218,23 @@ pub trait IClientModuleInit: IDynCommonModuleInit + Debug + MaybeSend + MaybeSyn
...
 
+    #[allow(clippy::too_many_arguments)]
+    async fn recover(
+        &self,
+        final_client: FinalClient,
+        federation_id: FederationId,
+        cfg: ClientModuleConfig,
+        db: Database,
+        // FIXME: don't make modules aware of their instance id
+        instance_id: ModuleInstanceId,
```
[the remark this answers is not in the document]
dpc: Stale comment. `ClientContext` has largely (if not completely) hid the id from the module impl.

fedimint-client/src/module/recovery.rs
dpc: numerator + denominator seem nice and informative, and the end application can convert as they please.

fedimint-client/src/lib.rs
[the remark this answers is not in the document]
Maan2003: right, end of stream always means something failed if module recovery didn't complete.

Maan2003 (approved):

fedimint-client/src/lib.rs
```diff
@@ -1552,18 +1761,42 @@ impl ClientBuilder {
         Ok(client)
     }
 
+    pub async fn join(
```
elsirion, line 1784: Needs documentation on when to use imo (with some big warnings about risk of fund loss if re-joining a federation with a previously used key without recovering). Even if the user never provided the key and it was randomly generated, but e.g. an app allowed leaving+joining federations that will be a problem.

Maybe we should always call recover and only attempt recovery if there is a backup (since otherwise it will take forever for real federations anyway)?

fedimint-client/src/sm/executor.rs
```diff
@@ -180,6 +182,48 @@ where
...
+
+            dbtx.insert_entry(
+                &ActiveStateKey::from_state(state.clone()),
+                &ActiveState::new(),
+            )
+            .await;
+        }
+
+        Ok(())
+    }
```
elsirion, line 226: The naming of this fn seems weird. It adds a state as active to the executor without checking if it actually is active. This also means afaik that we *have to* only call it with active states, otherwise inactive ones will get stuck in the active table forever, as they don't have any state transitions by definition.

Either we make the `is_terminal` check not require the actual module context (that seems to be the problem here) or we have to add a filter in the executor that will remove inactive states from the active states table if they end up there by accident.

dpc: > Either we make the `is_terminal` check not require the actual module context (that seems to be the problem here) or we have to add a filter in the executor that will remove inactive states from the active states table if they end up there by accident.

Detecting terminal ones and handling them correctly seems to simplify the caller and be more robust.

fedimint-client/src/lib.rs
dpc: Oh, I actually wanted to have a top level secret key to be an `enum`:

```rust
enum RootSecret {
  Fresh(DeriveableSecret),
  FromUser(DeriveableSecret),
}
```

so that `Fresh` is only returned when the key is randomly generated, and `FromUser` when entered by the user. This way type system would steer the implementation to handle it correctly, and it would be a good place to put docs explaining why it's needed.

dpc: Ehhh... I wrote some docstrings, but it made me realize that without some reliable way to tell if a given `root_secret` ever joined Federations, **most** apps will have to pessimize and call `recover` a lot which is particularly slow when a given Federation was not used (as it causes a whole history scan, because there can't be any previous backup).

> Maybe we should always call recover and only attempt recovery if there is a backup (since otherwise it will take forever for real federations anyway)?

And I'm not sure if relying on a backup is fundamentally sound. I wouldn't be surprised if backups will be a source of DoS and popular Federations will need to disable it altogether, or prune very old ones etc.

Maan2003 (dismissed):

fedimint-client/src/sm/executor.rs
elsirion: > Detecting terminal ones and handling them correctly seems to simplify the caller and generally more robust.

Can we do this in this PR please? I want to avoid forgetting about it after this one is used.

dpc: 🙄 Ok, ok. See the last commit. I wasn't all that confident about how I handled it there, but I hope it's OK.

Maan2003 (approved):

fedimint-client/src/sm/executor.rs
```diff
@@ -416,6 +376,25 @@ where
...
+            warn!(module_id = module_instance, "A terminal state where only active states are expected. Please report this bug upstream.");
+            self.db
+                .autocommit::<_, _, anyhow::Error>(
+                    |dbtx, _| {
+                        Box::pin(async {
+                            let k = InactiveStateKey::from_state(state.clone());
+                            let v = ActiveState::new().into_inactive();
+                            dbtx.insert_entry(&k, &v).await;
+                            Ok(())
+                        })
```
elsirion, line 391: Don't you also have to remove the active state?

dpc: Oh, right. I copied it from the place that was rejecting the new SM, which didn't have to do that. :facepalm:

elsirion (approved):
Thx for adding the last commit!

Maan2003 (approved):

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1526,7 +1600,6 @@ pub enum MintClientStateMachines {
     Output(MintOutputStateMachine),
     Input(MintInputStateMachine),
     OOB(MintOOBStateMachine),
-    Restore(MintRestoreStateMachine),
```
elsirion, line 1529: I should have caught this, running recoveries or logs of past recoveries could crash the client. We can:
1. Revert this change and put a dummy struct in there (since enums are length-encoded we don't have to preserver `MintRestoreStateMachine` but can just ignore whatever bytes are found)
2. Write a client DB migration that removes this particular SM. This would be super hacky since it would have to be done on the client and not module level.
3. Wait for SM migrations to land. What's the progress looking there @m1sterc001guy?

PR #4036, opened 2024-01-14
fix: incorrect URL for Launch Lightning Gateway in Mutinynet setup guide

wqxoxo:
The previous URL incorrectly pointed to 'lightning-gateway', whereas it should direct users to 'fedimint-gateway'. This change ensures users are guided to the correct page for setting up the Lightning Gateway in the context of Fedimint on Mutinynet.


elsirion (approved):
Thx!

Kodylow (approved):
LGTM thx!

PR #4042, opened 2024-01-15
feat: stream blocks in new refactoring of mint module recovery

dpc:
This makes recovery blazingly fast.


Maan2003:
might be interesting 
https://docs.rs/jsonrpsee/latest/jsonrpsee/async_client/struct.Client.html#method.batch_request

dpc:
> might be interesting
> https://docs.rs/jsonrpsee/latest/jsonrpsee/async_client/struct.Client.html#method.batch_request

I was never a fan of batch requests in jsonrpc. The overhead of just making multiple calls is not that large, and one can just open multiple connections to have them run in parallel. I guess saving number of connections could be a reason to use it sometimes... maybe?

Anyway - too much of a complication for the APIs (caching gets harder, timeouts, partial responses, etc) vs minimal (if any) benefit, IMO.

elsirion:
Added one comment in another PR https://github.com/fedimint/fedimint/pull/4046#discussion_r1455094494

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -533,6 +538,43 @@ impl MintClientInit {
...
+
+                    let block = loop {
+                        info!(target: LOG_CLIENT_RECOVERY_MINT, block_idx, "Awaiting signed block");
+                        match api.await_block(block_idx, &decoders).await {
+                            Ok(block) => break block,
+                            Err(e) => {
+                                info!(e = %e, block_idx, "Error trying to fetch signed block");
+                            }
+                        }
+                    };
```
elsirion, line 571: This could run tens of requests per second on failure iiuc, might want to add some exp backoff/delay.

elsirion (commented):
I think this is nearly good to go once the main refactor lands

modules/fedimint-mint-client/src/lib.rs
dpc: Good point. Will do.

dpc:
Added exponential backoff.

joschisan (approved):

Maan2003 (approved):

PR #4044, opened 2024-01-15
chore: speed up recovery by batching and streaming (v0.2 non-backport)

dpc:
This is effectively undoing some parts of
21da63533cf34722cff1e5afb9e8e06cd02a5eeb
hopefully to improve the real life recovery performance, which people are reporting as very slow.

The WIP recovery refactoring should render this irrelevant, but might be a good bandaid for 0.2.

Backport of #4019 which won't get merged because #4042 implements the same thing post recovery refactoring.

### DIFF
--- a/modules/fedimint-mint-client/src/backup/recovery.rs
+++ b/modules/fedimint-mint-client/src/backup/recovery.rs
@@ -1,6 +1,6 @@
-use std::cmp::max;
+use std::cmp::{self, max};
 use std::collections::BTreeMap;
-use std::fmt;
+use std::{fmt, ops};
 
 use fedimint_client::sm::{State, StateTransition};
 use fedimint_client::DynGlobalClientContext;
@@ -16,6 +16,7 @@ use fedimint_core::{Amount, NumPeers, OutPoint, PeerId, Tiered, TieredMulti};
 use fedimint_derive_secret::DerivableSecret;
 use fedimint_logging::LOG_CLIENT_RECOVERY_MINT;
 use fedimint_mint_common::{MintInput, MintOutput, Nonce};
+use futures::StreamExt as _;
 use serde::{Deserialize, Serialize};
 use tbs::{AggregatePublicKey, BlindedMessage, PublicKeyShare};
 use threshold_crypto::G1Affine;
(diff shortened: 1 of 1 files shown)


modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -271,20 +284,32 @@ impl MintRestoreInProgressState {
...
+                            Err(e) => {
+                                info!(e = %e, block_idx, "Error trying to fetch signed block");
+                            }
+                        }
+                    };
+
+                    (block_idx, block)
+                })
+            })
+            .buffered(8)
```
Maan2003, line 312: maybe bigger number

elsirion: 8 isn't super bad, I think it should be the number of downloads that can complete in one round trip time to not waste any time waiting. For active federations with medium client/server latency this would actually be a ok number, but since we currently mostly have inactive ones we could increase it by an order of magnitude.

elsirion (approved):
Looks like a good band-aid.

Maan2003 (approved):

modules/fedimint-mint-client/src/backup/recovery.rs
dpc: It's important to remember from the queue processing theory that increasing queue times is usually harmful: wastes resources and increases latencies. 

On top of my head an example: https://en.wikipedia.org/wiki/Bufferbloat

The role of a buffer is mostly to amortize the variations in the processing/job arrival times (burstiness, etc) on both ends, and allow both sides to work in parallel.

[`buffered` also polls buffered futures in parallel](https://docs.rs/futures-util/0.3.30/src/futures_util/stream/stream/buffered.rs.html#63) effectively increasing maximum parallelism level - I guess I could imagine implementation that polls only the first one, while buffering up to N resolved ones or something. Anyway, parallism is welcomed speedup here as well, but increasing it too much might not yield much improvement and considerably increase the load on the Federation, which is not great either.

So my feeling is that an 8 is a solid choice. :D

@Maan2003 @elsirion

PR #4046, opened 2024-01-16
 feat: await_block LRU cache 

dpc:
On top of #4035 

I feel a bit icky about global LRU, but since it's small and targeted, it seems much simpler than refactoring bunch of api traits trying to fit a cache in there somewhere.

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -412,6 +412,7 @@ where
     })
 }
 
+#[cfg(not(target_family = "wasm"))]
 lazy_static! {
     /// Small LRU used as `GlobalFederatinApi::await_block` cache.
     ///
@@ -445,6 +446,7 @@ where
         .await
     }
 
+    #[cfg(not(target_family = "wasm"))]
     async fn await_block(
         &self,
         block_index: u64,
@@ -480,6 +482,21 @@ where
         *block_lock = Some(block.clone());
 
         Ok(block)
+    }
+
+    #[cfg(target_family = "wasm")]
+    async fn await_block(
+        &self,
+        block_index: u64,
+        decoders: &ModuleDecoderRegistry,
+    ) -> anyhow::Result<SessionOutcome> {
+        self.request_current_consensus::<SerdeModuleEncoding<SessionOutcome>>(
+            AWAIT_SESSION_OUTCOME_ENDPOINT.to_string(),
+            ApiRequestErased::new(block_index),
+        )
+        .await?
+        .try_into_inner(decoders)
+        .map_err(|e| anyhow!(e.to_string()))
     }
 
     async fn session_count(&self) -> FederationResult<u64> {
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::fmt::{self, Debug, Display, Formatter};
 use std::io::{Cursor, Read};
+use std::num::NonZeroUsize;
 use std::ops::Add;
 use std::pin::Pin;
 use std::str::FromStr;
@@ -36,6 +37,7 @@ use jsonrpsee_core::client::{ClientT, Error as JsonRpcClientError};
 use jsonrpsee_wasm_client::{Client as WsClient, WasmClientBuilder as WsClientBuilder};
 #[cfg(not(target_family = "wasm"))]
 use jsonrpsee_ws_client::{WsClient, WsClientBuilder};
+use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use thiserror::Error;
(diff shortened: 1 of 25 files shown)


Maan2003:
you can use `thread_local` on wasm, because wasm doesn't have threads

fedimint-core/src/api.rs
```diff
@@ -432,13 +450,36 @@ where
...
+        let block = self
+            .request_current_consensus::<SerdeModuleEncoding<SessionOutcome>>(
+                AWAIT_SESSION_OUTCOME_ENDPOINT.to_string(),
+                ApiRequestErased::new(block_index),
+            )
+            .await?
+            .try_into_inner(decoders)
+            .map_err(|e| anyhow!(e.to_string()))?;
+
+        *block_lock = Some(block.clone());
```
Maan2003, line 482: fetching from network blocks other read from cache operations.
maybe use RwLock

dpc:
> you can use `thread_local` on wasm, because wasm doesn't have threads

Anything I can use instead?

dpc:
Wait'a minute. I'm not using thread local, but only `lazy_static`.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -533,6 +538,43 @@ impl MintClientInit {
...
+        futures::stream::iter(epoch_range)
+            .map(move |block_idx| {
+                let api = api.clone();
+                let decoders = decoders.clone();
+                Box::pin(async move {
+                    info!(block_idx, "Fetching epoch");
+
+                    let block = loop {
+                        info!(target: LOG_CLIENT_RECOVERY_MINT, block_idx, "Awaiting signed block");
+                        match api.await_block(block_idx, &decoders).await {
```
elsirion, line 565: Should we have retry behavior here? I guess not since we lose 7 already-fetched blocks in the worst case and hopefully have retry behavior further up?

fedimint-core/src/api.rs
```diff
@@ -410,6 +412,22 @@ where
...
+    /// The LRU can be be fairly small, as if the modules are (near-)bottlenecked
+    /// on fetching blocks they will naturally synchronize, or split into
+    /// a handful of groups. And if they are not, no LRU here is going to help
+    /// them.
+    static ref API_AWAIT_BLOCK_LRU: tokio::sync::Mutex<lru::LruCache<u64, Arc<tokio::sync::Mutex<Option<SessionOutcome>>>>> =
+        tokio::sync::Mutex::new(lru::LruCache::new(
+            NonZeroUsize::new(32).expect("is non-zero")
+        ));
+}
+
```
elsirion, line 431: If this *has to be* global then the federation id has to be part of the cache key, otherwise two clients running in the same process would get blocks from each other's federations.

Also, didn't we rename blocks to session outcomes? @joschisan

joschisan: yes we did in the api endpoints and the struct itself, just not everywhere in the code.

Maan2003:
> Wait'a minute. I'm not using thread local, but only lazy_static.

lazy static uses global `static`, which needs data to be sync

we can use `thread_local` on wasm

fedimint-core/src/api.rs
dpc: > If this _has to be_ global then the federation id has to be part of the cache key, otherwise two clients running in the same process would get blocks from each other's federations.

:facepalm: Right, good point.



> Also, didn't we rename blocks to session outcomes?

So I'm supposed to call it "sessions" everywhere from now on instead of blocks?

modules/fedimint-mint-client/src/lib.rs
dpc: I don't understand. The end state is that there's one big stream for a given recovery, this is a loop fetching one particular block/session in order and retrying until it works. We don't loose anything.

dpc:
> I don't think a global cache is a good idea anyway, the only benefit would be being able to set a limit on cache size that applies to all fedimint clients running in the process in total.

I'm not excited about the global cache, but if it's not global, it has to be somewhere (created during Client building, I guess) and needs to passed somehow to this place via bunch of traits/extension traits complicating everything in its path (especially blanket impls for `GlobalApiExt` etc. now will have to be impl for some kind of tuple-struct of `(T, cache)` etc. :roll_eyes: 


Ehhh... I guess wasm not being able to do `lazy_static` is another argument to avoid the global cache, and I'll have to do it.

fedimint-core/src/api.rs
elsirion: > So I'm supposed to call it "sessions" everywhere from now on instead of blocks?

Yes, sorry for the confusion. We renamed it (again) after you (and me too) complained about the ambiguity.

elsirion:
We could use the global cache for now if it makes things *much* simpler, but at least it has to be federation-keyed. I didn't fully consider API implications on blanket impls. But maybe there is an easy way out, like `IFederationApi` having a `get_session_cache` function which blanket impls on `T: IFederationApi` can use.

dpc:
I will attempt non-global version, mostly because wasm.

dpc:
https://github.com/fedimint/fedimint/pull/4080 instead

modules/fedimint-mint-client/src/lib.rs
elsirion: I was just blind

PR #4050, opened 2024-01-16
ci sandbox

bradleystachurski:
Testing behavior of running backwards compatibility as a separate ci workflow. Using a separate PR to avoid cluttering https://github.com/fedimint/fedimint/pull/4010


bradleystachurski:
This PR served its purpose 🙂

PR #4051, opened 2024-01-16
fix: more peer errors should be retriable

dpc:
Most unexpected weird stuff should be retried. Otherwise bussines logic fails things that it shouldn't because of weird lower level issues.

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -81,18 +81,23 @@ pub enum PeerError {
 impl PeerError {
     pub fn is_retryable(&self) -> bool {
         match self {
-            PeerError::ResponseDeserialization(_) => false,
+            PeerError::ResponseDeserialization(_) => true,
             PeerError::InvalidPeerId { peer_id: _ } => false,
             PeerError::Rpc(rpc_e) => match rpc_e {
-                // TODO: Does this cover all retryable cases?
                 JsonRpcClientError::Transport(_) => true,
                 JsonRpcClientError::MaxSlotsExceeded => true,
                 JsonRpcClientError::RequestTimeout => true,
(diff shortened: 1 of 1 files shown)


shaurya947 (approved):

fedimint-core/src/api.rs
```diff
@@ -81,18 +81,23 @@ pub enum PeerError {
...
             PeerError::Rpc(rpc_e) => match rpc_e {
-                // TODO: Does this cover all retryable cases?
                 JsonRpcClientError::Transport(_) => true,
                 JsonRpcClientError::MaxSlotsExceeded => true,
                 JsonRpcClientError::RequestTimeout => true,
                 JsonRpcClientError::RestartNeeded(_) => true,
-                JsonRpcClientError::Call(e) => e.code() == 404,
-                _ => false,
+                JsonRpcClientError::Call(_) => true,
+                JsonRpcClientError::ParseError(_) => true,
```
elsirion, line 92: Can we at least add a very loud warning/error in these cases? The problem is that state machines get stuck and will run ~forever if there is something truly unrecoverable.

dpc: It's generally up to the caller to print errors on retries, as a general good practice. Printing them here would just make them be printed twice.

elsirion: The problem is that the caller cannot determine the severity of the retry. E.g. `ParseError` is never supposed to happen, but we'd rather try to carry on to avoid losing money, but we want to log it loudly. In contrast a transport error (e.g. client went offline) is quite common and expected and should not be logged too loudly.

dpc:
Maybe https://github.com/fedimint/fedimint/pull/4064 instead?

PR #4055, opened 2024-01-18
chore: add module dkg message

m1sterc001guy:
For the Nostr federated client module that @EthnTuttle and I are building, I need to be able to exchange some messages during DKG to setup the Frost key. There might be a better way to implement FROST directly into Fedimint, but I think it also might be helpful to allow modules to exchange arbitrary messages during DKG. This PR adds `DkgPeerMsg::Module(Vec<u8>)` that allows modules to exchange any type of message during DKG.

Note: `exchange_public_keys` could probably be replaced with `exchange_with_peers<PublicKey>` but I'm not sure if that's backwards compatible, so I left the existing function.

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -747,11 +747,11 @@ pub trait TypedServerModuleConfig: DeserializeOwned + Serialize {
 }
 
 /// Things that a `distributed_gen` config can send between peers
-// TODO: Needs to be modularized in case modules want to send new message types for DKG
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum DkgPeerMsg {
     PublicKey(secp256k1::PublicKey),
     DistributedGen(SupportedDkgMessage),
+    Module(Vec<u8>),
     // Dkg completed on our side
     Done,
 }
@@ -772,6 +772,8 @@ pub enum DkgError {
     ModuleNotFound(ModuleKind),
     #[error("Params for modules were not found {0:?}")]
     ParamsNotFound(BTreeSet<ModuleKind>),
+    #[error("Failed to decode module message {0:?}")]
+    ModuleDecodeError(ModuleKind),
 }
 
 /// Supported (by Fedimint's code) `DkgMessage<T>` types
--- a/fedimint-server/src/config/distributedgen.rs
+++ b/fedimint-server/src/config/distributedgen.rs
@@ -7,8 +7,12 @@ use anyhow::{ensure, format_err};
 use async_trait::async_trait;
 use bitcoin::secp256k1;
 use bitcoin_hashes::sha256::{Hash as Sha256, HashEngine};
-use fedimint_core::config::{DkgGroup, DkgMessage, DkgPeerMsg, DkgResult, ISupportedDkgMessage};
-use fedimint_core::core::ModuleInstanceId;
+use fedimint_core::config::{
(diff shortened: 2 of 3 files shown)


fedimint-core/src/config.rs
```diff
@@ -747,11 +747,11 @@ pub trait TypedServerModuleConfig: DeserializeOwned + Serialize {
 }
 
 /// Things that a `distributed_gen` config can send between peers
-// TODO: Needs to be modularized in case modules want to send new message types for DKG
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum DkgPeerMsg {
     PublicKey(secp256k1::PublicKey),
     DistributedGen(SupportedDkgMessage),
+    Module(Vec<u8>),
```
elsirion, line 754: Eventually I envision this to be used for all module DKGs, even the ones that have a more specific message type right now. If we properly abstract the DKGs for each key type we should be able to make these messages more about the usage/instance of the key than its type.

That would also mean making it a proper `Dyn*` type eventually, but we got more important things to do right now, this unblocks you and we don't give stability guarantees for DKG for now as it's only run on startup and we can assume everyone to be on the same version.

fedimint-server/src/config/distributedgen.rs
```diff
@@ -506,6 +510,14 @@ pub trait PeerHandleOps {
         dkg_key: String,
         key: secp256k1::PublicKey,
     ) -> DkgResult<BTreeMap<PeerId, secp256k1::PublicKey>>;
+
+    async fn exchange_with_peers<T: Encodable + Decodable + Send + Sync>(
```
elsirion, line 514: Please document the semantics of the function. Does it send to everyone, what about itself?

elsirion (approved):

fedimint-core/src/config.rs
m1sterc001guy: Yep, sounds good to me.

elsirion (approved):

elsirion:
We probably need other primitives too, basically an API to implement arbitrary DKGs against, but this is good enough for now.

@joschisan which primitives would we eventually expose to modules, my best guess is:
* Encrypted+authenticated unicast

fedimint-server/src/config/distributedgen.rs
Maan2003: IMO docs should be on function declaration in trait, not on implementation because goto definition takes you to trait definition.

m1sterc001guy: good call, moved to the trait. Also added one for `exchange_pubkeys`

dpc (approved):

fedimint-server/src/config/distributedgen.rs
```diff
@@ -562,4 +581,41 @@ impl<'a> PeerHandleOps for PeerHandle<'a> {
...
+        while peer_data.len() < self.peers.len() {
+            match self
+                .connections
+                .receive((self.module_instance_id, dkg_key.clone()))
+                .await?
+            {
+                (peer, DkgPeerMsg::Module(bytes)) => {
+                    let received_data: T = T::consensus_decode_vec(bytes, &modules)
+                        .map_err(|_| DkgError::ModuleDecodeError(kind.clone()))?;
+                    peer_data.insert(peer, received_data);
```
elsirion, line 611: nit: we should probably warn/reject if a peer overwrites their message.

elsirion (approved):
One new minor nit I found this round, feel free to address or merge.

PR #4062, opened 2024-01-18
Add doc strings modules/fedimint-ln-common/src/api.rs

justinmoon:
- First commit adds doc strings
- Second commit groups related method names together
- Third commit renames some method names to use "fetch" instead of "get" prefix. THIS WILL BREAK RUST API SO MIGHT NOT WANT TO DO IT. But should be backwards-compatible federation-wise.  I didn't modify `get_decrypted_preimage_status` because the `get` verb is in the RPC method name itself which would be difficult to update backwards-compatibly, so I left it as-is.

### DIFF
--- a/modules/fedimint-ln-common/src/api.rs
+++ b/modules/fedimint-ln-common/src/api.rs
@@ -39,7 +39,7 @@ pub trait LnFederationApi {
     /// Await fedration consensus on a lightning module [`ContractAccount`]
     async fn wait_contract(&self, contract: ContractId) -> FederationResult<ContractAccount>;
     /// Fetch [`OutgoingContractAccount`] for a given contract ID
-    async fn get_outgoing_contract(
+    async fn fetch_outgoing_contract(
         &self,
         id: ContractId,
     ) -> FederationResult<OutgoingContractAccount>;
@@ -68,7 +68,7 @@ pub trait LnFederationApi {
     /// Check whether [`IncomingContractOffer`] exists
     async fn offer_exists(&self, payment_hash: Sha256Hash) -> FederationResult<bool>;
     /// Fetch [`IncomingContractAccount`] for a given contract ID
-    async fn get_incoming_contract(
+    async fn fetch_incoming_contract(
         &self,
         id: ContractId,
     ) -> FederationResult<IncomingContractAccount>;
@@ -204,7 +204,7 @@ where
             .is_some())
     }
 
-    async fn get_incoming_contract(
+    async fn fetch_incoming_contract(
         &self,
         id: ContractId,
     ) -> FederationResult<IncomingContractAccount> {
@@ -220,7 +220,7 @@ where
         }
     }
 
-    async fn get_outgoing_contract(
+    async fn fetch_outgoing_contract(
         &self,
         id: ContractId,
     ) -> FederationResult<OutgoingContractAccount> {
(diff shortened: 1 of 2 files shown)


modules/fedimint-ln-common/src/api.rs
```diff
@@ -22,47 +23,63 @@ use crate::contracts::outgoing::OutgoingContractAccount;
...
     async fn get_decrypted_preimage_status(
         &self,
         contract: ContractId,
     ) -> FederationResult<(IncomingContractAccount, DecryptedPreimageStatus)>;
+    /// Waits for preimage decryption to complete
     async fn wait_preimage_decrypted(
         &self,
         contract: ContractId,
     ) -> FederationResult<(IncomingContractAccount, Option<Preimage>)>;
+    /// Fetch an [IncomingContractOffer]
```
bradleystachurski, line 63: I haven't verified, but I think markdown link check is failing since this doesn't have backticks.

modules/fedimint-ln-common/src/api.rs
```diff
@@ -22,47 +23,63 @@ use crate::contracts::outgoing::OutgoingContractAccount;
...
     ) -> FederationResult<Option<ContractAccount>>;
+    /// Await fedration consensus on a lightning module [`ContractAccount`]
     async fn wait_contract(&self, contract: ContractId) -> FederationResult<ContractAccount>;
-    async fn wait_block_height(&self, block_height: u64) -> FederationResult<()>;
+    /// Fetch [`OutgoingContractAccount`] for a given contract ID
+    async fn fetch_outgoing_contract(
+        &self,
+        id: ContractId,
+    ) -> FederationResult<OutgoingContractAccount>;
+    /// Wait for the lightning gateway to successfully user client escrow
```
dpc, line 46: No empty line between items looks crowded and makes skimming through code harder, IMO.

dpc (approved):

modules/fedimint-ln-common/src/api.rs
```diff
@@ -22,43 +23,59 @@ use crate::contracts::outgoing::OutgoingContractAccount;
...
+    /// Fetch a lightning module [`ContractAccount`] by ID
     async fn fetch_contract(
         &self,
         contract: ContractId,
     ) -> FederationResult<Option<ContractAccount>>;
+    /// Await fedration consensus on a lightning module [`ContractAccount`]
     async fn wait_contract(&self, contract: ContractId) -> FederationResult<ContractAccount>;
+    /// Wait for a consensus block height to be reached
     async fn wait_block_height(&self, block_height: u64) -> FederationResult<()>;
+    /// Wait for the lightning gateway to successfully user client escrow
```
elsirion, line 46: Is a verb missing here?

modules/fedimint-ln-common/src/api.rs
```diff
@@ -39,7 +39,7 @@ pub trait LnFederationApi {
     /// Await fedration consensus on a lightning module [`ContractAccount`]
     async fn wait_contract(&self, contract: ContractId) -> FederationResult<ContractAccount>;
     /// Fetch [`OutgoingContractAccount`] for a given contract ID
-    async fn get_outgoing_contract(
+    async fn fetch_outgoing_contract(
```
elsirion, line 42: These functions should generally not be used by external software, so changing the verb sounds ok. We don't intend to backport this to 0.2 anyway for lack of functional changes.

Issue #4070, opened 2024-01-19
Discription field missing in Cargo.toml of `devimint` and `fedimint-wasm-test`


15IITian:

elsirion:
@dpc is there a case to be made for publishing `devimint` to crates.io? I didn't think so, which would make `Cargo.toml` metadata completeness much less of a concern.

dpc:
Not a strong case, but I guess if someone wants to use devimint for something and want to install it via `cargo install` ... why not?

PR #4071, opened 2024-01-19
chore: add discription in Cargo.toml

15IITian:
#4070

### DIFF
--- a/fedimint-wasm-tests/Cargo.toml
+++ b/fedimint-wasm-tests/Cargo.toml
@@ -3,6 +3,7 @@ name = "fedimint-wasm-tests"
 version = "0.0.0"
 edition = "2021"
 license = "MIT"
+description = "Wasm tests for the fedimint."
 publish = false
 
 [lib]


fedimint-wasm-tests/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "fedimint-wasm-tests"
 version = "0.0.0"
 edition = "2021"
 license = "MIT"
+description = "Wasm tests for the fedimint."
 publish = false
```
elsirion, line 7: This crate isn't being published, I don't think this is needed.

15IITian: > This crate isn't being published, I don't think this is needed.

@elsirion , It was my  first thought but all other unpublishable crates have `discription` field :

image

elsirion: I mean ok, now that it's already open, but in general this isn't worth the team's bandwidth it consumes in review.

elsirion (approved):

fedimint-wasm-tests/Cargo.toml
15IITian: > I mean ok, now that it's already open, but in general this isn't worth the team's bandwidth it consumes in review.

MB, I understands -> will  try to avoid it .

Maan2003 (approved):

PR #4080, opened 2024-01-19
 feat: await_block LRU cache (non-global) 

dpc:
On top of #4035 and #4042.

This is a version #4046 without global statics.

Wasm couldn't handle `lazy_static`, and globals are icky. The price is wrangling with an already overly complicated set of traits in this area.

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -40,6 +40,7 @@ use jsonrpsee_ws_client::{WsClient, WsClientBuilder};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use thiserror::Error;
+use tokio::sync::OnceCell;
 use tracing::{debug, error, instrument, trace, warn};
 
 use crate::backup::ClientBackupSnapshot;
@@ -510,9 +511,7 @@ struct GlobalFederationApiWithCache<T> {
     /// synchronize, or split into a handful of groups. And if they are not,
     /// no LRU here is going to help them.
     #[allow(clippy::type_complexity)]
-    await_block_lru: Arc<
-        tokio::sync::Mutex<lru::LruCache<u64, Arc<tokio::sync::Mutex<Option<SessionOutcome>>>>>,
-    >,
+    await_block_lru: Arc<tokio::sync::Mutex<lru::LruCache<u64, Arc<OnceCell<SessionOutcome>>>>>,
 }
 
 impl<T> GlobalFederationApiWithCache<T> {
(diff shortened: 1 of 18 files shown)


fedimint-core/src/api.rs
```diff
@@ -360,15 +385,6 @@ pub trait GlobalFederationApi {
 
     async fn await_transaction(&self, txid: TransactionId) -> FederationResult<TransactionId>;
 
-    async fn await_output_outcome<R>(
```
dpc, line 363: Not object safe so had be kicked out (to `impl DynGlobalApi`).

fedimint-core/src/api.rs
```diff
@@ -440,6 +474,78 @@ where
...
+        let mut block_lock = block_entry_arc.lock().await;
+
+        if let Some(cached_res) = block_lock.clone() {
+            debug!(block_index, "Using a cached await_block response");
+            return Ok(cached_res);
+        }
+
+        let block = self.await_block_raw(block_index, decoders).await?;
+
+        *block_lock = Some(block.clone());
```
elsirion, line 545: Should we also debug log wen we are fetching the block or is that already done by `await_block_raw`?

elsirion (commented):
Looks good, but blocked on two other PRs

fedimint-core/src/api.rs
dpc: Will check when I rebase, it should also be easier to review in full.

dpc: Unclear if it's really needed.

dpc:
@elsirion :heavy_check_mark: :question: :pray: :rofl:

fedimint-core/src/api.rs
elsirion: I mean, you do have the log when the block/session was cached. But fine by me :man_shrugging: more logging more good though

elsirion (approved):

fedimint-core/src/api.rs
```diff
@@ -497,6 +544,78 @@ where
...
+        let block_entry_arc = lru_lock
+            .get_or_insert(block_index, || Arc::new(tokio::sync::Mutex::new(None)))
+            .clone();
+
+        // we drop the lru lock so requests for other `block_index` can work in parallel
+        drop(lru_lock);
+
+        // but multiple request for the same `block_index` will serialize on the
+        // per-index lock, avoiding doing the same work
+        let mut block_lock = block_entry_arc.lock().await;
```
Maan2003, line 594: block_entry can be a `OnceCell`

https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html

Maan2003 (approved):

fedimint-core/src/api.rs
Maan2003: there is already log in await_block_raw

Maan2003 (approved):

fedimint-core/src/api.rs
Maan2003: i pushed https://github.com/fedimint/fedimint/pull/4080/commits/4efb0f22a261e007ed2d42d71a2fca60b9a32d16 to reduce review cycles

Maan2003 (approved):

elsirion (approved):
Good catch, makes the code a lot simpler!

fedimint-core/src/api.rs
dpc: Interesting.

PR #4081, opened 2024-01-19
refactor: Lightning Dir for Gateway

Kodylow:
This PR refactors the Gateway's lightning implementations into a `lighting` directory.

Started as a hack to get the gateway to work with the alby api, but halfway through realized this would be a good refactor for adding additional implementations and apis as lightning options for the gateway in the future.

### DIFF
--- a/gateway/ln-gateway/src/lightning/cln.rs
+++ b/gateway/ln-gateway/src/lightning/cln.rs
@@ -3,103 +3,22 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use async_trait::async_trait;
-use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::task::{sleep, TaskGroup};
 use fedimint_core::util::SafeUrl;
-use fedimint_core::Amount;
-use fedimint_ln_common::PrunedInvoice;
 use futures::stream::BoxStream;
-use serde::{Deserialize, Serialize};
-use thiserror::Error;
 use tonic::transport::{Channel, Endpoint};
 use tonic::Request;
 use tracing::info;
 
+use super::{ILnRpcClient, LightningRpcError};
 use crate::gateway_lnrpc::gateway_lightning_client::GatewayLightningClient;
 use crate::gateway_lnrpc::{
     EmptyRequest, EmptyResponse, GetNodeInfoResponse, GetRouteHintsRequest, GetRouteHintsResponse,
     InterceptHtlcRequest, InterceptHtlcResponse, PayInvoiceRequest, PayInvoiceResponse,
 };
-use crate::lnd::GatewayLndClient;
-use crate::LightningMode;
+use crate::lightning::MAX_LIGHTNING_RETRIES;
 pub type HtlcResult = std::result::Result<InterceptHtlcRequest, tonic::Status>;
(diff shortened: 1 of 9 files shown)


okjodom:
sweet! Just resolved conflicts and pushed

okjodom (approved):
ship it

gateway/ln-gateway/src/lightning/cln.rs
```diff
@@ -3,104 +3,23 @@ use std::sync::Arc;
...
 use crate::gateway_lnrpc::gateway_lightning_client::GatewayLightningClient;
 use crate::gateway_lnrpc::{
     EmptyRequest, EmptyResponse, GetNodeInfoResponse, GetRouteHintsRequest, GetRouteHintsResponse,
     InterceptHtlcRequest, InterceptHtlcResponse, PayInvoiceRequest, PayInvoiceResponse,
 };
-use crate::lnd::GatewayLndClient;
-use crate::LightningMode;
+use crate::lightning::MAX_LIGHTNING_RETRIES;
 pub type HtlcResult = std::result::Result<InterceptHtlcRequest, tonic::Status>;
 pub type RouteHtlcStream<'a> = BoxStream<'a, HtlcResult>;
```
tvolk131, line 21: Should we move these to `gateway/ln-gateway-src-lightning/mod.rs`? They don't look CLN-specific.

tvolk131 (changes requested):

gateway/ln-gateway/src/lightning/cln.rs
okjodom: `cln.rs` is a misnomer, but these types relate to client implementation that uses grpc to connect to lightning service over a network hop.

This is not meant to be CLN specific

Kodylow: Yeah that's a good point, it's a generic we use with cln but could be reapplied elsewhere, @okjodom what's a better naming scheme here?

okjodom: my alt name suggestion was `network.rs` but that has potential to be more confusing :)

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    // called once per client. A stream of intercepted HTLCs and a `Arc<dyn
+    // ILnRpcClient> are returned to the caller. The caller can use this new
+    // client to interact with the lightning node, but since it is an `Arc` is
+    // cannot call `route_htlcs` again.
+    async fn route_htlcs<'a>(
+        self: Box<Self>,
+        task_group: &mut TaskGroup,
+    ) -> Result<(RouteHtlcStream<'a>, Arc<dyn ILnRpcClient>), LightningRpcError>;
+
+    async fn complete_htlc(
```
m1sterc001guy, line 94: Nit: I would love to add a doc comment here, we keep forgetting to do so!

Kodylow: can you suggest a good one, I'm a little confused myself about what this part does and doesn't do as the generic haha

m1sterc001guy: I can follow up with a separate PR, we don't need to do it here

PR #4083, opened 2024-01-20
Alby Lightning Gateway Integration using Webhook and Alby API

Kodylow:
Built this as a proof of concept for how we might be able to use a Lightning Service Provider API with the Gateway.

This fully implements the pay side of the gateway, receive side requires some changes on their end but working with @m1sterc001guy and @okjodom we think we found a pretty good solution:

Most custodial lightning APIs don't expose the htlc interceptor and it's a big ask to expose that, but they DO have the ability to register webhooks for updates. 

This impl adds a rpc_webhook_server that starts a webhook endpoint for the Alby API to forward htlcs marked for the gateway to. We use a btreemap of oneshot channels so that same endpoint can forward the htlc into the normal gateway htlc stream and listen for the response of Fail or Success(preimage), then return the preimage as a response to the webhook call.

So the entire ask for getting a lightning api provider to integrate with us turns into:

We need you to implement a single endpoint, where you POST an htlc that you would normally fail and the response will contain the preimage for that htlc or Error.

Which we were thinking is a way lighter lift than getting them to actually run gateways or our daemons/extensions themselves.

This is a WIP and was more of an exercise to see if it might work, would love to hear feedback/thoughts on the approach.

### DIFF
--- a/fedimint-testing/src/gateway.rs
+++ b/fedimint-testing/src/gateway.rs
@@ -17,7 +17,7 @@ use fedimint_logging::LOG_TEST;
 use futures::executor::block_on;
 use lightning_invoice::RoutingFees;
 use ln_gateway::client::GatewayClientBuilder;
-use ln_gateway::lnrpc_client::{ILnRpcClient, LightningBuilder};
+use ln_gateway::lightning::{ILnRpcClient, LightningBuilder};
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
 use ln_gateway::rpc::{ConnectFedPayload, FederationConnectionInfo};
 use ln_gateway::{Gateway, GatewayState};
--- a/fedimint-testing/src/ln/mock.rs
+++ b/fedimint-testing/src/ln/mock.rs
@@ -18,7 +18,8 @@ use ln_gateway::gateway_lnrpc::{
     self, EmptyResponse, GetNodeInfoResponse, GetRouteHintsResponse, InterceptHtlcResponse,
     PayInvoiceRequest, PayInvoiceResponse,
 };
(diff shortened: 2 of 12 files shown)


asoltys:
I'd be willing to add an endpoint to https://github.com/coinos/coinos-server to play with this. It's a nodejs API server and can talk to our CLN node over RPC.

Kodylow:
Was planning to implement a bunch of these tomorrow morning Pacific time and coinos just jumped to the top of the list hahaha awesome!

Shoot me a discord message if you need any help or would like to pair, I'll be in the fedimint general voice channel all day https://discord.com/invite/exjN3NAF I'm @kodylow

Kodylow:
just fixed the imports issue, you should be able to run this as is. I'll work on a `coinos.rs` in the morning (basically mirror alby.rs but using your api for info and pay methods on the trait) , the receive methods should be exactly the same for coinos
as they are in this alby impl

m1sterc001guy:
As mentioned on the dev call, the biggest concern is figuring out how to test something like this. One option is to build out our own "test-custodial" API that could serve as an example to service providers on what APIs are needed to be exposed in order to support a Fedimint LN Gateway. Otherwise this might need to live outside of the Fedimint repo

Kodylow:
agreed, let's maintain this just as a draft so we can provide reference to people looking to implement the webhook side though, they've already found it useful

dpc:
Being opened it kind of gets in a way and looks like unaddressed work. I would prefer this was closed if no one minds. Maybe link it from some documentation or something. Otherwise I doubt anyone will find it here gathering dust anyway.

Kodylow:
Yup closing, will attack more generically this was a good exercise though

PR #4084, opened 2024-01-20
Coinos Api Impl for Lightning Gateway

Kodylow:
@asoltys Here's most of an implementation using the Coinos API. Pay and info are complete, we'd have to work with you on how best to register the webhook. It looks like you've already got a webhook impl for notifications when a payment is complete, this would be a webhook that sends whenever the onion contains a specific scid.

Let's chat on monday to see what's most effective on the coinos-server side but this looks really promising.


Kodylow:
Working with @asoltys one issue we might run into is that the scid we're using isn't unique, we're just using 0 indexed counter. We'd need to be able to pass a specific gateway id if we're using a node with many sub accounts.

image

dpc:
Looks still salvagable, so will just post a comment to make it dissapear from the "longest time with no progress" list.

PR #4086, opened 2024-01-21
chore: migrate to miniscript v10

tvolk131:
Migrate from [miniscript v9.0.2](https://crates.io/crates/miniscript/9.0.2) to [miniscript v10.0.0](https://crates.io/crates/miniscript/10.0.0) everywhere except one callsite where bitcoin v0.29's `PartiallySignedTransaction` uses miniscript v9's `PsbtExt` trait and provided implementation.

Miniscript v9 relies on bitcoin v0.29, whereas miniscript v10 relies on bitcoin v0.30, which is why we're adding `fedimint-core/src/bitcoin_migration.rs` which contains functions that can convert necessary types between these two versions of the bitcoin crate.

### DIFF
--- a/fedimint-core/src/bitcoin_migration.rs
+++ b/fedimint-core/src/bitcoin_migration.rs
@@ -1,0 +1,199 @@
+use std::str::FromStr;
+
+use bitcoin30::hashes::Hash as Bitcoin30Hash;
+use bitcoin_hashes::Hash as Bitcoin29Hash;
+
+pub fn bitcoin29_to_bitcoin30_public_key(pk: bitcoin::PublicKey) -> bitcoin30::PublicKey {
+    bitcoin30::PublicKey::from_slice(&pk.to_bytes())
+        .expect("Failed to convert bitcoin v29 public key to bitcoin v30 public key")
+}
+
+pub fn bitcoin29_to_bitcoin30_network(network: bitcoin::Network) -> bitcoin30::Network {
+    match network {
+        bitcoin::Network::Bitcoin => bitcoin30::Network::Bitcoin,
+        bitcoin::Network::Testnet => bitcoin30::Network::Testnet,
+        bitcoin::Network::Signet => bitcoin30::Network::Signet,
+        bitcoin::Network::Regtest => bitcoin30::Network::Regtest,
+    }
+}
+
+pub fn bitcoin30_to_bitcoin29_script(script: bitcoin30::ScriptBuf) -> bitcoin::Script {
+    script.to_bytes().into()
+}
+
+pub fn bitcoin30_to_bitcoin29_address(address: bitcoin30::Address) -> bitcoin::Address {
+    bitcoin::Address::from_str(&address.to_string())
+        .expect("Failed to convert bitcoin v30 address to bitcoin v29 address")
+}
+
+pub fn bitcoin29_to_bitcoin30_ripemd160_hash(
+    hash: bitcoin::hashes::ripemd160::Hash,
+) -> bitcoin30::hashes::ripemd160::Hash {
+    bitcoin30::hashes::ripemd160::Hash::from_byte_array(hash.into_inner())
+}
+
+pub fn bitcoin29_to_bitcoin30_hash160_hash(
+    hash: bitcoin::hashes::hash160::Hash,
+) -> bitcoin30::hashes::hash160::Hash {
+    bitcoin30::hashes::hash160::Hash::from_byte_array(hash.into_inner())
+}
+
+pub fn bitcoin29_to_bitcoin30_sha256_hash(
+    hash: bitcoin::hashes::sha256::Hash,
+) -> bitcoin30::hashes::sha256::Hash {
+    bitcoin30::hashes::sha256::Hash::from_byte_array(hash.into_inner())
+}
+
+#[cfg(test)]
+mod tests {
+    use bitcoin_hashes::hex::FromHex;
+    use rand::thread_rng;
+
+    use super::*;
+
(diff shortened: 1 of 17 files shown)


fedimint-core/src/bitcoin_migration.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+pub fn bitcoin30_to_bitcoin29_address(address: bitcoin30::Address) -> bitcoin::Address {
+    bitcoin::Address::from_str(&address.to_string()).unwrap()
+}
+
+pub fn bitcoin29_to_bitcoin30_ripemd160_hash(
+    hash: bitcoin::hashes::ripemd160::Hash,
+) -> bitcoin30::hashes::ripemd160::Hash {
+    bitcoin30::hashes::ripemd160::Hash::from_str(&hash.to_string()).unwrap()
```
elsirion, line 29: I got distracted yesterday trying to figure it out myself, so just leaving the question: could we just go `Hash -> &[u8] -> Hash` instead? A typical pattern in `bitcoin_hashes` was to use `&hash[..]` to get the byte representation as `&[u8]` (idk if there's an even better way now). But serializing and deserializing looks very wasteful.

tvolk131: Good point! Done.

fedimint-core/src/bitcoin_migration.rs
```diff
@@ -0,0 +1,196 @@
...
+    match network {
+        bitcoin::Network::Bitcoin => bitcoin30::Network::Bitcoin,
+        bitcoin::Network::Testnet => bitcoin30::Network::Testnet,
+        bitcoin::Network::Signet => bitcoin30::Network::Signet,
+        bitcoin::Network::Regtest => bitcoin30::Network::Regtest,
+    }
+}
+
+pub fn bitcoin30_to_bitcoin29_script(script: bitcoin30::ScriptBuf) -> bitcoin::Script {
+    bitcoin::Script::from_hex(&script.to_hex_string()).unwrap()
```
m1sterc001guy, line 21: We shouldn't use `unwrap` here, `expect` with a descriptive error message is better. 

I'm not sure how we should handle errors here though, should we assume that every `bitcoin30::ScriptBuf` is a valid `bitcoin::Script`?

fedimint-core/src/bitcoin_migration.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+pub fn bitcoin30_to_bitcoin29_address(address: bitcoin30::Address) -> bitcoin::Address {
+    bitcoin::Address::from_str(&address.to_string()).unwrap()
```
m1sterc001guy, line 25: Same comment as above

modules/fedimint-wallet-common/Cargo.toml
```diff
@@ -22,7 +22,8 @@ bitcoin = { version = "0.29.2", features = [ "rand", "serde"] }
 erased-serde = "0.3"
 fedimint-core = { version = "0.3.0-alpha", path = "../../fedimint-core" }
 futures = "0.3"
-miniscript = { version = "9.0.2", features = [ "compiler", "serde" ] }
+miniscript = { version = "10.0.0", features = [ "compiler", "serde" ] }
+miniscript9 = { package = "miniscript", version = "9.0.2" }
```
m1sterc001guy, line 26: Why does `wallet-common` need both versions of `miniscript`?

fedimint-core/src/bitcoin_migration.rs
elsirion: `Script`/`ScriptBuf` doesn't do any semantic validation. The error comes from the hex decoding, which isn't necessary since we can use the [`from/to_bytes`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/script/struct.ScriptBuf.html#method.from_bytes) methods instead

fedimint-core/src/bitcoin_migration.rs
elsirion: This seems a bit more interesting, need to check if the old version parsed taproot addresses correctly already.

elsirion (commented):
Looks good apart from the `unwrap`s.

fedimint-core/src/bitcoin_migration.rs
elsirion: I checked and 0.29 already has taproot code, so I assume the both can always be converted between each other.

tvolk131: Thanks for looking into that!

modules/fedimint-wallet-common/Cargo.toml
tvolk131: Because this piece of code returns a miniscript v9 error:

https://github.com/fedimint/fedimint/blob/28ecd67ca0c7279c9d2d5d00f34fe20fc77e3062/modules/fedimint-wallet-server/src/lib.rs#L909-L911

We could add a function to `bitcoin_migration.rs` to convert this, but the error type has [40 variants](https://docs.rs/miniscript/9.0.1/miniscript/enum.Error.html) and I don't know that we really use it for much anyway. I figure it's easier to wait until we upgrade to bitcoin v30 to swap this out.

fedimint-core/src/bitcoin_migration.rs
m1sterc001guy: Can you use `from/to_bytes` like @elsirion mentioned here?

m1sterc001guy (approved):

fedimint-core/src/bitcoin_migration.rs
tvolk131: Done.

m1sterc001guy (approved):

elsirion (approved):

PR #4088, opened 2024-01-22
chore(fedimint-cli): convert invalid JSON data to JSON string

### LINKED ISSUE #3194: `fedimint-cli dev api` usability is meh.
Took more forever to get config from the server for debugging purposes.

First #3192 , so I had to get the invite code from `$FM_DATA_DIR`.

But then I couldn't figure out what is the syntax for aguments, and Justin had to tell me it should be:

```
fedimint-cli dev api config '"fed114znk7uk7ppugdjuytr8venqf2tkywd65cqvg3u93um64tu5cw4yr0n3fvn7qmwvm4g48cpndgnm4gqq4waen5te0xyerwt3s9cczuvf6xyurzde597s7crdvsk2vmyarjw9gwyqjdzj"'
```

but that fails with:

```
> fedimint-cli dev api config '"fed114znk7uk7ppugdjuytr8venqf2tkywd65cqvg3u93um64tu5cw4yr0n3fvn7qmwvm4g48cpndgnm4gqq4waen5te0xyerwt3s9cczuvf6xyurzde597s7crdvsk2vmyarjw9gwyqjdzj"'
2023-09-14T21:32:32.288091Z  INFO fedimint_client: Last client reference dropped, shutting down client task group
2023-09-14T21:32:32.288097Z  INFO fedimint_client::sm::executor: Starting state machine executor task
2023-09-14T21:32:32.288633Z  INFO fedimint_client::sm::executor: Shutting down state machine executor runner due to shutdown signal
{
  "error": "CliError",
  "kind": "GeneralFailure",
  "message": "Federation rpc error {general => Received errors from 1 peers: peer-1: Rpc error: ErrorObject { code: ServerError(400), message: \"Download token not found\", data: None }), 1 => Rpc error: ErrorObject { code: ServerError(400), message: \"Download token not found\", data: None }), }",
  "raw_error": "Federation rpc error {general => Received errors from 1 peers: peer-1: Rpc error: ErrorObject { code: ServerError(400), message: \"Download token not found\", data: None }), 1 => Rpc error: ErrorObject { code: ServerError(400), message: \"Download token not found\", data: None }), }"
}
```

which Justin explained to me is because invite-code is per-peer. So the right invocation is actually:

```
fedimint-cli dev api --peer-id 0 config '"fed114znk7uk7ppugdjuytr8venqf2tkywd65cqvg3u93um64tu5cw4yr0n3fvn7qmwvm4g48cpndgnm4gqq4waen5te0xyerwt3s9cczuvf6xyurzde597s7crdvsk2vmyarjw9gwyqjdzj"'
```

At very least we should add some examples to `fedimint-cli dev api --help`

In addition, if the invite-codes are per-guardian, then maybe we should check it somewhere for better usability.

E.g. we could embed the `PeerId` in `InviteCode` as a field, or `url` could be a map/list. Just so we can return a better error ("this invite code is for a different peer).

BTW. It seems to me like the server side is not checking `InviteCode::id` at all, which probably is not a best way to handle it. We should return `incorrect federation id`, possibly preventing some weird confusion attacks etc. Created #3195 for that.

15IITian:
If `params` fails to get parsed into JSON -> then convert it into valid `JSON String` wrapped in `Value::String`

fix #3194

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -699,8 +699,14 @@ impl FedimintCli {
             }) => {
                 //Parse params to JSON.
                 //If fails, convert to JSON string.
-                let params = serde_json::from_str::<Value>(&params)
-                    .unwrap_or_else(|_| serde_json::Value::String(params));
+                let params = serde_json::from_str::<Value>(&params).unwrap_or_else(|err| {
+                    debug!(
+                        "Failed to serialize params:{}. Converting it to JSON string",
+                        err
+                    );
+
+                    serde_json::Value::String(params)
+                });
 
                 let mut params = ApiRequestErased::new(params);
                 if let Some(auth) = auth {
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -325,7 +325,7 @@ async fn load_or_generate_mnemonic(db: &Database) -> Result<Mnemonic, CliError> 
 #[derive(Subcommand, Clone)]
 #[allow(clippy::large_enum_variant)]
 enum Command {
-    /// Print the latest git commit hash this bin. was build with
+    /// Print the latest Git commit hash this bin. was built with.
     VersionHash,
 
     #[clap(flatten)]
@@ -697,8 +697,11 @@ impl FedimintCli {
                 peer_id,
                 password: auth,
             }) => {
-                let params: Value = serde_json::from_str(&params)
-                    .map_err_cli_msg(CliErrorKind::InvalidValue, "Invalid JSON-RPC parameters")?;
+                //Parse params to JSON.
+                //If fails, convert to JSON string.
+                let params = serde_json::from_str::<Value>(&params)
+                    .unwrap_or_else(|_| serde_json::Value::String(params));
+
                 let mut params = ApiRequestErased::new(params);
                 if let Some(auth) = auth {
                     params = params.with_auth(ApiAuth(auth))


fedimint-cli/src/lib.rs
```diff
@@ -696,8 +696,13 @@ impl FedimintCli {
                 peer_id,
                 password: auth,
             }) => {
-                let params: Value = serde_json::from_str(&params)
-                    .map_err_cli_msg(CliErrorKind::InvalidValue, "Invalid JSON-RPC parameters")?;
+                let params = serde_json::from_str::<Value>(&params).unwrap_or_else(|_| {
+                    serde_json::Value::String(
+                        // If parsing as JSON fails, treat it as a JSON string
+                        serde_json::to_string(&params).expect("Failed to convert to JSON String"),
+                    )
```
Maan2003, line 703: incorrect?
`foo` will be converted to `Value::String("\"foo\"")`

15IITian: Actually ,It gives error when I try to return `Json_string` because of `mismatched types` .
I am not getting how to solve this problem ,
can you please help me out.

Maan2003:
```suggestion
                    serde_json::Value::String(params)
```

15IITian: Do I have to handle error related  `serde_json::Value::String(params)`?
if Yes, then should I return `Value::Null` in case of error ?

fedimint-cli/src/lib.rs
```diff
@@ -325,7 +325,7 @@ async fn load_or_generate_mnemonic(db: &Database) -> Result<Mnemonic, CliError>
 #[derive(Subcommand, Clone)]
 #[allow(clippy::large_enum_variant)]
 enum Command {
-    /// Print the latest git commit hash this bin. was build with
+    /// Print the latest git commit hash ,this bin was build with.
```
GetPsyched, line 328: The comma is not needed.
```suggestion
    /// Print the latest Git commit hash this bin was built with.
```
Also, the `.` you removed likely symbolised `bin` being an abbreviation; so, it should remain or the abbreviation be expanded.

GetPsyched:
Please mark each review as resolved after you've added the suggested changes; it clutters the PR view for future reviews.

elsirion (approved):
Looks useful, some nits

fedimint-cli/src/lib.rs
```diff
@@ -325,7 +325,7 @@ async fn load_or_generate_mnemonic(db: &Database) -> Result<Mnemonic, CliError>
 #[derive(Subcommand, Clone)]
 #[allow(clippy::large_enum_variant)]
 enum Command {
-    /// Print the latest git commit hash this bin. was build with
+    /// Print the latest Git commit hash this bin. was built with.
```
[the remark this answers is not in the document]
15IITian: Ok , will keep in mind.
and Thanks for your suggestions :+1:

fedimint-cli/src/lib.rs
```diff
@@ -696,8 +696,11 @@ impl FedimintCli {
                 peer_id,
                 password: auth,
             }) => {
-                let params: Value = serde_json::from_str(&params)
-                    .map_err_cli_msg(CliErrorKind::InvalidValue, "Invalid JSON-RPC parameters")?;
+                //Parse params to JSON.
+                //If fails, convert to JSON string.
+                let params = serde_json::from_str::<Value>(&params)
+                    .unwrap_or_else(|_| serde_json::Value::String(params));
```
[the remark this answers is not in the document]
15IITian: is it OK ?

image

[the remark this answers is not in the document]
elsirion: It doesn't need to be two statements, can you make it one please?

elsirion (approved):
There might be some funny edge cases, like `null` being valid JSON, but overall this will be a quality of life improvement for this debug tool.

elsirion:
I don't think it's solvable without creating other edge-cases.

15IITian:
Ok

dpc (approved):

PR #4103, opened 2024-01-23
chore: Client Database Migrations (without state machine migrations)

m1sterc001guy:
Step 1 of https://github.com/fedimint/fedimint/issues/3417

This PR adds the equivalent tests and functions for client db migrations that the server side has.

Best to review this PR per commit, each module is a separate commit.

State machine db migrations will come next.


fedimint-cli/src/lib.rs
```diff
@@ -518,6 +518,7 @@ impl FedimintCli {
         let mut client_builder = Client::builder(db);
         client_builder.with_module_inits(self.module_inits.clone());
         client_builder.with_primary_module(1);
+        client_builder.migrate_database().await.map_err_cli_io()?;
```
dpc, line 521: Probably not the right place to hook it. Seems to me that that `ClientBuilder::make_client_builder` would be better. Why? Because at this stage we are not attempting to build the client yet, so why would it start doing anything.

BTW. I have a recovery PR that refactors bunch of things here so there will be trouble (merge conflicts). But not a big deal.

fedimint-client/src/lib.rs
```diff
@@ -1487,6 +1488,31 @@ impl ClientBuilder {
...
     }
 
+    pub async fn migrate_database(&self) -> anyhow::Result<()> {
+        // Only apply the client database migrations if the database has been
+        // initialized.
+        if let Ok(client_config) = self.load_existing_config().await {
+            for (module_id, module_cfg) in client_config.modules {
+                let kind = module_cfg.kind.clone();
+                let Some(init) = self.module_inits.get(&kind) else {
+                    bail!("Detected configuration for unsupported module id: {module_id}, kind: {kind}")
```
dpc, line 1498: We're not supposed to fail on modules we don't support, but just ignore them. Unless I'm too tired to be reviewing.

gateway/ln-gateway/src/client.rs
```diff
@@ -71,6 +71,7 @@ impl GatewayClientBuilder {
         let mut client_builder = Client::builder(db);
         client_builder.with_module_inits(registry);
         client_builder.with_primary_module(self.primary_module);
+        client_builder.migrate_database().await?;
```
dpc, line 74: Shouldn't `ClientBuilder` call that already for the ln-gateway's client? :sus:

dpc: Hmmm... maybe gateway has it's own builder ... which seems odd. Shouldn't be it be using the standard client builder under the hood anyway?

modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -653,6 +665,69 @@ mod fedimint_migration_tests {
...
 
+    async fn create_client_db_with_v0_data(mut dbtx: DatabaseTransaction<'_>) {
+        dbtx.insert_new_entry(&DatabaseVersionKey, &DatabaseVersion(0))
+            .await;
+
+        // Generate fake private/public key
+        let (_, pk) = secp256k1::generate_keypair(&mut OsRng);
+        dbtx.insert_new_entry(
+            &fedimint_ln_client::db::LightningGatewayKey,
+            &LightningGatewayRegistration {
```
dpc, line 676: style: Kind of a long and ugly fixture. Maybe we could move it out of the way (into a constant or some util function).

dpc (commented):
Generally LGTM, with some minor comments.

One thing that it's not doing is touching active executor states. Do you think it's not neccessary (because meta on it is serialized in json, so it can be somewhat gradually handled?)

m1sterc001guy:
Thanks for the review!

> One thing that it's not doing is touching active executor states. Do you think it's not neccessary (because meta on it is serialized in json, so it can be somewhat gradually handled?)

Nope, I think this does need to be handled. I just wanted to do it in a separate PR since this one was rather big already.

fedimint-cli/src/lib.rs
m1sterc001guy: This is called in `make_client_builder` right? Maybe I'm just not understanding your comment.

I put it here because this seemed to be the earliest point at which the database was loaded, and I wanted to minimize the chance that we try to read anything from the database before the db can be migrated (even though we still actually read the config from the db before migration currently).

Alternatively I could move this to be a function on `Client` instead of `ClientBuilder`. Then it would be the external library's responsibility to call `migrate_database` before doing anything with the database. But then we'd need to be careful to essentially never touch the database instead of `ClientBuilder` because it could break.

This is a small addition, so I'm ok with handling the merge conflicts :)

fedimint-client/src/lib.rs
m1sterc001guy: Ack, will fix

gateway/ln-gateway/src/client.rs
m1sterc001guy: It uses the same `ClientBuilder`, we just create that struct ourself instead of using `make_client_builder`. I think there might have been a refactor in the cli that we didn't do in the gateway.

I can try to change this to use `make_client_builder` if that's preferrable

fedimint-cli/src/lib.rs
dpc: > Seems to me that `ClientBuilder::make_client_builder` would be better. 

My mistake. I was to write `ClientBuilder::build_stopped`. This is where the `Client` is actually built, recovery/init functions are called. We should do migrations right before that.

m1sterc001guy: Cool, works for me.

elsirion: Could there be any reason not to migrate the DB? In that case we could add a boolean to the builder that controls if migration will run on construction.

dpc: New code will expect new datastructures, it's possibly not even safe to call anything in the module without migrations up to date.

dpc:
Big recovery refactor landed, good time to rebase and land.

fedimint-cli/src/lib.rs
m1sterc001guy: Moved to `build_stopped`

fedimint-client/src/lib.rs
m1sterc001guy: Fixed

gateway/ln-gateway/src/client.rs
m1sterc001guy: Removed, since this is now called in `build_stopped`

m1sterc001guy:
I added the `MintRecoveryState` to the mint migration test and did my best to populate as much of it as possible. I had to make some things public that were previously only within the crate or private, hopefully this is an acceptable tradeoff for having migration coverage

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-mod db;
+pub mod db;
```
elsirion, line 1: I'm ok with making this pub, but is there a reason for migration tests not to live in their respective crate (client or server)? We mainly have the tests crates for integration-testing client and server together.

modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -937,6 +938,7 @@ mod fedimint_migration_tests {
                                 num_gateways > 0,
                                 "validate_migrations was not able to read any LightningGateways"
                             );
+                            info!("Validated LightningGateways");
```
elsirion, line 948: Wrong commit (ln test changes in wallet db migration commit)

crypto/tbs/src/lib.rs
```diff
@@ -194,50 +194,44 @@ pub fn verify(msg: Message, sig: Signature, pk: AggregatePublicKey) -> bool {
     pairing(&msg.0, &pk.0) == pairing(&sig.0, &G2Affine::generator())
 }
 
-#[cfg(test)]
-mod tests {
-    use std::collections::BTreeMap;
+pub fn dealer_keygen(
```
elsirion, line 197: Did you want to touch that crypto code here or is this a rebase conflict that got resolved the wrong way? @joschisan refactored `tbs` recently: #3468.

modules/fedimint-mint-tests/tests/tests.rs
```diff
@@ -235,6 +249,53 @@ mod fedimint_migration_tests {
...
+                spend_key: keypair,
+            },
+        )
+        .await;
+
+        dbtx.insert_new_entry(&NextECashNoteIndexKey(Amount::from_sats(1000)), &3)
+            .await;
+
+        dbtx.insert_new_entry(&CancelledOOBSpendKey(OperationId(BYTE_32)), &())
+            .await;
```
elsirion, line 302: Restore state is likely being added later?

elsirion (approved):
Not the cleanest commit history, but an important step towards client version migrations :tada:

crypto/tbs/src/lib.rs
elsirion: Joscha made a good observation: you don't need a valid signature for this test, so you can pick a random point and don't need `dealer_keygen`. E.g.:

```rust
Signature(G1Affine::generator())
```

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: I originally did it just so all of the migration tests were in the same spot, but now I'm somewhat regretting it as it made me change all of these things to pub. Would you prefer if I moved them back? Probably best done in a separate PR

crypto/tbs/src/lib.rs
m1sterc001guy: Just a rebase conflict, I'll fix with joshican's suggestion

modules/fedimint-mint-tests/tests/tests.rs
m1sterc001guy: Yes, added in commit after rebase

modules/fedimint-ln-client/src/lib.rs
elsirion: I agree it's not worth uprooting this PR. Let's try to tidy up public interfaces in a separate PR. Not super high priority.

fedimint-client/src/module/init.rs
```diff
@@ -182,6 +182,12 @@ where
 pub trait ClientModuleInit: ModuleInit + Sized {
     type Module: ClientModule;
 
+    const DATABASE_VERSION: DatabaseVersion;
+
+    fn get_database_migrations(&self) -> MigrationMap {
```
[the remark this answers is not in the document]
m1sterc001guy: Added docs

modules/fedimint-ln-tests/tests/tests.rs
m1sterc001guy: Fixed this and did my best to cleanup the rest of the commit history

crypto/tbs/src/lib.rs
m1sterc001guy: Fixed, no more `crypto` changes

modules/fedimint-ln-client/src/lib.rs
m1sterc001guy: Will leave this for now and handle in separate PR

Maan2003 (approved):

elsirion (approved):

PR #4104, opened 2024-01-23
fix: make single guardian devimint cli test backwards-compatible

### LINKED ISSUE #4017: `devimint_cli_test_single` fails backwards-compatibility test
https://github.com/fedimint/fedimint/actions/runs/7434410505/job/20228534969?pr=4010#step:11:2879

```
00:00:17 thread 'main' panicked at devimint/src/main.rs:459:9:
00:00:17 assertion failed: descriptor.contains(\"wpkh(\")
```

bradleystachurski:
Fixes #4017 

I verified locally against the full matrix of backwards-compatibility tests that `devimint_cli_test_single` succeeds (see: https://github.com/fedimint/fedimint/pull/4010).

```
Backwards-compatibility tests summary:
fed_version  client_version  gateway_version  exit_code
v0.2.1       v0.2.1          current          0
v0.2.1       current         v0.2.1           0
v0.2.1       current         current          0
current      v0.2.1          v0.2.1           0
current      v0.2.1          current          0
current      current         v0.2.1           0
```

### DIFF
--- a/devimint/src/main.rs
+++ b/devimint/src/main.rs
@@ -416,19 +416,27 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         .state();
     anyhow::ensure!(invoice_status == tonic_lnd::lnrpc::invoice::InvoiceState::Settled);
 
-    // # Test the correct descriptor is used
-    let config = cmd!(client, "config").out_json().await?;
-    let guardian_count = config["global"]["api_endpoints"].as_object().unwrap().len();
-    let descriptor = config["modules"]["2"]["peg_in_descriptor"]
-        .as_str()
-        .unwrap()
-        .to_owned();
+    // fedimintd introduced wpkh for single guardian federations in v0.3.0 (9e35bdb)
+    // The code path is backwards-compatible, however this test will fail if we
+    // check against earlier fedimintd versions.
+    let fedimintd_version_res = cmd!(devimint::util::FedimintdCmd, "--version")
+        .out_string()
+        .await;
+    if devimint::util::is_min_version(0, 3, 0, fedimintd_version_res) {
+        // # Test the correct descriptor is used
+        let config = cmd!(client, "config").out_json().await?;
+        let guardian_count = config["global"]["api_endpoints"].as_object().unwrap().len();
+        let descriptor = config["modules"]["2"]["peg_in_descriptor"]
+            .as_str()
+            .unwrap()
+            .to_owned();
 
-    info!("Testing generated descriptor for {guardian_count} guardian federation");
-    if guardian_count == 1 {
-        assert!(descriptor.contains("wpkh("));
-    } else {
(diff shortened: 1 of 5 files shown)


elsirion (dismissed):
:tada:

devimint/src/main.rs
```diff
@@ -416,19 +416,27 @@ async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
-        .as_str()
-        .unwrap()
-        .to_owned();
+    // fedimintd introduced wpkh for single guardian federations in v0.3.0 (9e35bdb)
+    // The code path is backwards-compatible, however this test will fail if we
+    // check against earlier fedimintd versions.
+    let fedimintd_version_res = cmd!(devimint::util::FedimintdCmd, "--version")
+        .out_string()
+        .await;
+    if devimint::util::is_min_version(0, 3, 0, fedimintd_version_res) {
```
dpc, line 425: `is_min_version` confuses me. Would expect `is_version_at_least` or `is_version_gt` kind of things.

Also `0, 3, 0` ...  looks weird.

devimint/src/util.rs
```diff
@@ -419,3 +420,153 @@ impl GatewayLndCli {
...
+        minor: captured["minor"].parse()?,
+        patch: captured["patch"].parse()?,
+    })
+}
+
+/// Returns true if the provided version string is greater than or equal to the
+/// min version provided. If there's an error parsing the version, returns
+/// false.
+pub fn is_min_version(
+    min_major: u32,
```
dpc, line 480: If we have `CrateVersion` why not use it here?

devimint/src/util.rs
```diff
@@ -419,3 +420,153 @@ impl GatewayLndCli {
...
+    min_minor: u32,
+    min_patch: u32,
+    version_res: Result<String, anyhow::Error>,
```
dpc, line 483: Taking `Result<String, anyhow::Error>` here is odd and confusing.

If for some reason older versions didn't have `--version` or something like that, then the version parsing function should default to some magic number (0.2.0? 0.1.0?) to unify all the cases.

dpc: Without these issues, we can use some more natural version comparison (I'm not sure if `cmp::Ordering` and operators like `<=` could be used for semver, need more coffee).

BTW. https://docs.rs/semver/latest/semver/ exists and [is not a heavy crate](https://crates.io/crates/semver/1.0.21/dependencies), maybe we should just use it?

bradleystachurski: Good call 👍 

If we're comfortable introducing the semver crate, it simplifies this PR.

https://github.com/fedimint/fedimint/pull/4104/commits/2fa38846037d28a13348254f95eb4d0a12139257

dpc (approved):
:shipit:

elsirion (approved):

PR #4106, opened 2024-01-23
chore: better (hopefully) lock file handling

dpc:
I've just read:

https://blog.aspect.dev/easier-merges-on-lockfiles

and it occured to me that it would be very easy to implement in a Nix shell.

Never tried it, so I'm not sure if it's a good idea, but I'm curious. I always resolve lock file conflicts this way anyway (just manually).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -327,6 +327,10 @@
                   if [ "$(ulimit -Sn)" -lt "1024" ]; then
                       >&2 echo "⚠️  ulimit too small. Run 'ulimit -Sn 1024' to avoid problems running tests"
                   fi
+
+                  if [ -z "$(git config --global merge.ours.driver)" ]; then
+                      >&2 echo "⚠️  Recommended to run 'git config --global merge.ours.driver true' to enable better lock file handling. See https://blog.aspect.dev/easier-merges-on-lockfiles for more info"
+                  fi
                 '';
               };
             in
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.lock merge=ours


flake.nix
```diff
@@ -327,6 +327,10 @@
                   if [ "$(ulimit -Sn)" -lt "1024" ]; then
                       >&2 echo "⚠️  ulimit too small. Run 'ulimit -Sn 1024' to avoid problems running tests"
                   fi
+
+                  if [ -z "$(git config --global merge.ours.driver)" ]; then
+                      >&2 echo "⚠️  Recommended to run 'git config --global merge.ours.driver true' to enable better lock file handling. See https://blog.aspect.dev/easier-merges-on-lockfiles for more info"
+                  fi
```
Maan2003, line 333: can we set this in repo local config

dpc: Possibly, but changes like that don't propagate with `git clone` (because `true` here is an arbitrary command) so I don't think it changes much. And if we/you/someone e.g. want to use it in multiple projects, it's easier to do it once in global user's config than everywhere.

Maan2003 (approved):

elsirion (approved):
This avoids merge conflicts, but if one doesn't run some cargo command before continuing the rebase the lock file will be outdated, right? Or maybe some git hook already checks that the lock file is up to date?

dpc:
> This avoids merge conflicts, but if one doesn't run some cargo command before continuing the rebase the lock file will be outdated, right? Or maybe some git hook already checks that the lock file is up to date?

Possibly yes, but I we do have checks for it in the CI and I think in the git hook, so all in all I think it will be more convenient.

Issue #4093, opened 2024-01-22
Don't allow LN gateways to join with 0PPM


TonyGiorgio:
It doesn't appear that the current LN gateways with 0PPM will do anything since it can't route payments with a max fee of 0. Disallow this in some way or warn against this since it is not transparent about what the issues are with the gateway when trying to make payments. See my confusion on this previous issue: https://github.com/fedimint/fedimint/issues/3990

okjodom:
I'm surprised gateways default to 0PPM when [global default fee](https://github.com/fedimint/fedimint/blob/bc384160febd44152eca93a5a04f6164bc7431b8/gateway/ln-gateway/src/lib.rs#L97) has 1000PPM, 0msat base fee. Investigating

justinmoon:
dev call: @elsirion would be better to implement this filtering on client rather than server

TonyGiorgio:
I've implemented this on my client.

PR #4108, opened 2024-01-23
test gateway fees and revise docker defaults

okjodom:
- updates `v0.2.*` docker compose files to package gateway with **0 base msat, 10000 PPM** default fees
- updates tests to include assertions for variable part of globally configured gateway fees

partially addresses #4093 by updating fee defaults in docker packaged gatewayd

### DIFF
--- a/modules/fedimint-ln-common/src/config.rs
+++ b/modules/fedimint-ln-common/src/config.rs
@@ -4,7 +4,7 @@ pub use bitcoin::Network;
 use fedimint_core::bitcoinrpc::BitcoinRpcConfig;
 use fedimint_core::core::ModuleKind;
 use fedimint_core::encoding::{Decodable, Encodable};
-use fedimint_core::plugin_types_trait_impl_config;
+use fedimint_core::{msats, plugin_types_trait_impl_config, Amount};
 use lightning_invoice::RoutingFees;
 use serde::{Deserialize, Serialize};
 use threshold_crypto::serde_impl::SerdeSecret;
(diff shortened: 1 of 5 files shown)


fedimint-testing/src/gateway.rs
```diff
@@ -115,10 +114,7 @@ impl GatewayTest {
             address.clone(),
             cli_password.clone(),
             None, // Use default Network which is "regtest"
-            RoutingFees {
-                base_msat: 0,
-                proportional_millionths: 0,
-            },
+            DEFAULT_FEES,
```
m1sterc001guy, line 117: Does this break any of the outgoing payment tests? I thought we had a test that assumed no fees were taken out. Either way, this is changing the fees so I'm confused why more of the tests didn't require changes

okjodom: One test is broken, because I didn't break it. Actually revising this so I don't change the default fees, but since my intent is to have better coverage for fees configured, I added ppm part to `test_can_change_routing_fees`

modules/fedimint-ln-common/src/config.rs
```diff
@@ -146,3 +146,27 @@ impl FromStr for GatewayFee {
         }))
     }
 }
+
+pub trait ToAmount {
```
m1sterc001guy, line 150: nit: add trait doc

m1sterc001guy (dismissed):

modules/fedimint-ln-common/src/config.rs
okjodom: ah yes

elsirion (approved):
Looks reasonable. Last two commits looks like they should be squashed, the last removes code that got duplicated in the second-to-last.

dpc (approved):

Issue #4027, opened 2024-01-11
`LeaveFederation` gateway RPC should optimistically try to de-register itself from the federation


m1sterc001guy:
Rather than just rely on the TTL of a registered gateway with a federation, if the gateway operator explicitly calls `LeaveFederation`, we should try to contact the federation to remove the registration record.

elsirion:
Deregistering should also be triggered on shutdown if the gateway is shutdown cooperatively using SIGINT. Idk what we use in other places rn, but a quick search showed [`tokio-shutdown`](https://docs.rs/tokio-shutdown/latest/tokio_shutdown/), which should probably work.

m1sterc001guy:
The API to remove the gateway's registrations needs to be authenticated, so a rogue user can't just make API requests to the federation and delete all of the gateway registrations. One way to do this is to require the gateway to send a signature that the federation can verify. Since the `gatewayd_id` is already a `PublicKey`, the gateway can sign with this key. The main issue IMO is figuring out what to sign. If the message being signed is deterministic, the same signature can be presented to the federation multiple times, which doesn't really solve anything.

One solution is to generate some random bytes for each gateway registration, then require the gateway to sign these if it wants to remove that registration record. This seems like the most doable solution, but unfortunately requires a database migration :/

justinmoon:
It could be challenge:response. Hit the federation to ask for some message (could be random) and gateway signs it. This message isn't a consensus thing so I think this would work.

PR #4114, opened 2024-01-24
chore: store Lightning Gateway Registrations by gateway id

m1sterc001guy:
Replaces: https://github.com/fedimint/fedimint/pull/3389 @tvolk131 for visibility

Working towards https://github.com/fedimint/fedimint/issues/4027 I'd like to de-register the gateway when it receives SIGINT or it leaves the federation, but we can't do that until we store gateways by `gateway_id`, otherwise we risk deleting multiple gateways if they share the same `node_pub_key` (should only happen if using a custodial API)

I believe this is a rare care where a change to a database structure does not require a migration. The structure of the key stays the same (it is still a `PublicKey`), the value is just different. Additionally, on the next registration of the gateway, a new record in the db will be created, and the old one will just be cleaned up once it expires. Because of this, I do not believe a database migration is required.

### DIFF
--- a/modules/fedimint-ln-server/src/lib.rs
+++ b/modules/fedimint-ln-server/src/lib.rs
@@ -1138,7 +1138,7 @@ impl Lightning {
         self.delete_expired_gateways(dbtx).await;
 
         dbtx.insert_entry(
-            &LightningGatewayKey(gateway.info.node_pub_key),
+            &LightningGatewayKey(gateway.info.gateway_id),
             &gateway.anchor(),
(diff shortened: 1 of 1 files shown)


okjodom (approved):

okjodom:
such a happy coincidence in id types!

justinmoon:
> I believe this is a rare care where a change to a database structure does not require a migration

This makes sense to me.

justinmoon (approved):

PR #4116, opened 2024-01-24
fix: use fedimint-aleph-bft

### LINKED ISSUE #4054: `UnitSaver::flush` is using wrong `block_on`?
[While investigating something entirely different](https://github.com/fedimint/fedimint/pull/3995#issuecomment-1897513085) I've noticed that `UnitSaver::flush` is using `futures::executor::block_on` which I don't think is entirely right thing to do when running in tokio Runtime (and thus everything hanged when `tokio::task::yield_now` was added to the mix). Unfortunately these kinds of issues require at least 200 IQ points to solve so I could use some help. :D

Maan2003:
fixes #4054

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,31 +67,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "aleph-bft"
-version = "0.30.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb050890181b868ad4d601661a5e7106aca6191273c99c092d6d472119f46e56"
-dependencies = [
- "aleph-bft-rmc",
- "aleph-bft-types",
- "anyhow",
- "async-trait",
- "derivative",
- "futures",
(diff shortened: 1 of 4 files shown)


Maan2003:
I didn't use async `BackupReader` because we use `Cursor<Vec<u8>>` which is fine to read in async code.

joschisan:
> I didn't use async `BackupReader` because we use `Cursor<Vec<u8>>` which is fine to read in async code.

I think using BackupReader would allow for a cleaner implementation with a second UnitReader struct and two separate constructors instead of the load_session method that now returns both, but we can also leave it until we upstreamed the change.

joschisan (dismissed):

Maan2003:
> two separate constructors instead of the load_session method

I attempted this, we still need to find initial `units_index` for `UnitSaver`, for which we have to check the entire old units

elsirion (approved):

PR #4119, opened 2024-01-24
backport: feat(ln-client): make LN gateway known upfront

Maan2003:
backport #3882

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -65,7 +65,7 @@ use secp256k1_zkp::{All, Secp256k1};
 use serde::{Deserialize, Serialize};
 use strum::IntoEnumIterator;
 use thiserror::Error;
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 use crate::db::{
     LightningGatewayKeyPrefix, MetaOverrides, MetaOverridesKey, MetaOverridesPrefix,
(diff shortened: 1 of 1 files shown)


Maan2003:
this breaks rust api, but never promised rust api stability in our releases

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1002,9 +1012,15 @@ impl LightningClientModule {
...
+    /// supplied only internal payments are possible.
+    ///
+    /// The `gateway` can be acquired by calling
+    /// [`LightningClientModule::select_active_gateway`].
     pub async fn pay_bolt11_invoice<M: Serialize + MaybeSend + MaybeSync>(
         &self,
+        maybe_gateway: Option<LightningGateway>,
         invoice: Bolt11Invoice,
         extra_meta: M,
     ) -> anyhow::Result<OutgoingLightningPayment> {
```
elsirion, line 1026: Backporting takes some more work here: the old fn signature has to be maintained and a new fn added with the updated signature. Otherwise we break SemVer.

Maan2003: do we commit to semver?

Maan2003: it made backport smaller, so fine for me

Maan2003: oh right we publish on crates io

Issue #4016, opened 2024-01-07
`fedimint-cli --mnemonic` backwards incompatible change


bradleystachurski:
`fedimint-cli --mnemonic` (long) is introduced in https://github.com/fedimint/fedimint/commit/3746d51dc6d060ed8a10662a2ccdd065a501160e#diff-c10cb3590cef71888e923fecbce68e80ab9628e12e69e2ec5a3afa7aa335810cR155

Failing backwards-compatibility test logs: https://github.com/fedimint/fedimint/actions/runs/7434410505/job/20228534969?pr=4010#step:11:3809

```
00:00:24 Error: command: /nix/store/fdbrlr3n4n94gs5a1lgnzskhgmv8pimq-fedimint-cli/bin/fedimint-cli --data-dir=/home/ubuntu/tmp/nix-shell.aNMRXi/devimint-299624/clients/restore-without-backup restore --mnemonic panel␣swear␣leave␣traffic␣luxury␣gallery␣embody␣squeeze␣unable␣parent␣clip␣shield --invite-code fed11qgqpw9thwvaz7te3xgmjuvpwxqhrzw33x5uryd30qqqjqu7mqtxs7tqqc75lcc7s6zm3daw7w8myrq8yx6gzht7dwkxuf9evqrandu
00:00:24 
00:00:24 Caused by:
00:00:24     exit status: 2
00:00:24     stdout:
00:00:24     
00:00:24     stderr:
00:00:24     error: unexpected argument '--mnemonic' found
00:00:24     
00:00:24       tip: to pass '--mnemonic' as a value, use '-- --mnemonic'
00:00:24     
00:00:24     Usage: fedimint-cli restore <MNEMONIC>
00:00:24     
00:00:24     For more information, try '--help'.
00:00:24     
```

```
nix@lenovo-m75q:~/fedimint$ git tag --contains 3746d51dc6d060ed8a10662a2ccdd065a501160e
nix@lenovo-m75q:~/fedimint$ git branch --contains 3746d51dc6d060ed8a10662a2ccdd065a501160e
* master
```

elsirion:
Should we adapt the test script instead here? At least in my mind we didn't commit to CLI stability (maybe we should? @justinmoon)

PR #4120, opened 2024-01-24
feat(spend): convert `amount` postional to named arg

15IITian:
#3822

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -62,6 +62,7 @@ pub enum ClientCmd {
     /// Prepare notes to send to a third party as a payment
     Spend {
         /// The amount of e-cash to spend
+        #[clap(long)]
         amount: Amount,
         /// If the exact amount cannot be represented, return e-cash of a higher
         /// value instead of failing


elsirion:
Not worth it imo as it would require a branch in our test suite to pass backwards compatibility testing. @bradleystachurski would you agree? @dpc thoughts as the biggest fan of named arguments I know? :P

bradleystachurski:
Agreed, we should wait for backwards-compatibility in CI 👍 

I'd expect this will break compatibility similar to `--mnemonic` (https://github.com/fedimint/fedimint/issues/4016).

I do think it'd be good to establish a pattern for updating our CLI in a backwards-compatible way, but I'd expect that'll wait until we fix `mnemonic`.

15IITian:
> Not worth it imo as it would require a branch in our test suite to pass backwards compatibility testing. @bradleystachurski would you agree? @dpc thoughts as the biggest fan of named arguments I know? :P

If there is such a case -> then please filter out the required work to do.
#3822

dpc:
:shrug:

elsirion:
I'll close this for now. Feel free to open issues for things you want to work on first so we can discuss the design/if it makes sense to do for the project right now.

15IITian:
@elsirion  
can I work on #3822 second last task?

Issue #4126, opened 2024-01-25
Go back to `aleph-bft` v0.30


elsirion:
In #4116 we upgraded AlephBFT from 0.30 to 0.33 without considering that when a session is still running it has persisted data in the database. Since they changed their on-disk format and we didn't want this fix to be a consensus upgrade we need to apply @Maan2003's patch on 0.30 instead and downgrade the version we depend on again.

In the future, when we intend to upgrade AlephBFT, this will mean we need to wait for the current session to end and probably bump the minor consensus version.

Maan2003:
backport for aleph-bft https://github.com/Maan2003/AlephBFT/commits/backport-blocking-fixes-0.30/

PR #4134, opened 2024-01-25
0.2.2-rc7

Maan2003:
related #4126

backport #4133 #4135

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -224,7 +224,7 @@ impl ConsensusServer {
         while !task_handle.is_shutting_down() {
             let session_index = self.get_finished_session_count().await;
 
-            let mut item_index = self.build_session_outcome().await.items.len() as u64;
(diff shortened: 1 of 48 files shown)


elsirion (approved):
Please wait for master to be merged

fedimint-server/src/consensus/server.rs
```diff
@@ -508,18 +514,15 @@ impl ConsensusServer {
         self.modules.decoder_registry()
     }
 
-    pub async fn build_session_outcome(&self) -> SessionOutcome {
-        let items = self
-            .db
+    pub async fn pending_accepted_items(&self) -> Vec<AcceptedItem> {
```
elsirion, line 517: Technically this breaks SemVer, but nobody is supposed to directly rely on this (we should make it private on master).

elsirion (commented):
I think you forgot to update the nix version?

elsirion (approved):

PR #4137, opened 2024-01-25
refactor: generic framework for from-history module recoveries

dpc:


joschisan (dismissed):

fedimint-client/src/module/init/recovery.rs
```diff
@@ -0,0 +1,417 @@
...
+        client_ctx
+            .clone()
+            .module_autocommit_2(
+                move |dbtx, _| {
+                    let state = state.clone();
+                    {
+                        Box::pin(async move {
+                            state.delete_dbtx(&mut dbtx.module_dbtx()).await;
+                            state.finalize_dbtx(dbtx).await?;
+                            Recovery::store_finalized(&mut dbtx.module_dbtx(), true).await;
```
elsirion, line 405: Further down you write that finalization can happen without recovery being terminated/removed from DB? I can't see how.

fedimint-client/src/module/init/recovery.rs
```diff
@@ -0,0 +1,417 @@
...
+        {
+            // In rare circumstances, the finalization could complete, yet the completion
+            // of `recover` function not yet persisted in the database. So
+            // it's possible that `recovery` would be called again on an
+            // already finalized state. Because of this we need to store some
+            // finalization marker in the same dbtx as the finalization itself, detect this
+            // here and exit early.
+            warn!("Previously finalized, exiting");
+            return Ok(());
+        }
```
elsirion, line 342: I don't see how this could happen, should we maybe do more stuff in the same dbtx?

fedimint-client/src/module/init/recovery.rs
```diff
@@ -0,0 +1,417 @@
...
+        let current_session_count = client_ctx.global_api().session_count().await?;
+
+        let (mut state, mut common_state) =
+            // TODO: if load fails, we could ... start from scratch? at least being able to force this behavior might be useful
```
elsirion, line 346: I don't think I understand that comment

fedimint-client/src/module/init/recovery.rs
```diff
@@ -0,0 +1,417 @@
...
+    async fn handle_session(
+        &mut self,
+        client_ctx: &ClientContext<<Self::Init as ClientModuleInit>::Module>,
+        _session_idx: u64,
+        session: &SessionOutcome,
+    ) -> anyhow::Result<()> {
+        for accepted_item in &session.items {
+            if let ConsensusItem::Transaction(ref transaction) = accepted_item.item {
+                self.handle_transaction(client_ctx, transaction).await?;
+            }
```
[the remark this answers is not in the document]
dpc: We can add more methods (`handle_consensus_item` with blanket no-op implementations) and forward from `handle_session` at any point (backward-compatible), plus client can always override the whole `handle_session`, even if we didn't.

fedimint-client/src/module/init/recovery.rs
dpc: I missed this problem in all previous implementations, as it's easy to overlook.

It can't happen anymore because of this `state.finalize_dbtx` and the check where this comment was explaining why it's necessary.

Example sequence how this happens (if `finalize_dbtx` didn't exist):

0. module progress is complete and saved to the db
1. `dbtx` with finalization commits, completing recovery on the client module side
2. client crashes/gets terminated
3. client starts again
4. client never observed/persisted that the module finished recovery, so calls module recovery again
5. module loads progress and sees it's complete, moves on to finalization again
6. module runs finalization again and probably fails because it's actually not idempotent and doesn't expect the already existing state.

fedimint-client/src/module/init/recovery.rs
dpc: If e.g. we didn't have db migration code to migrate existing recovery progress value in the db, and we failed to decode ... we could delete the existing progress and start recovery from scratch...

elsirion: That makes sense!

dpc:
@elsirion Added more comments on that finalization corner case, fixed type mistakes.

fedimint-client/src/module/init/recovery.rs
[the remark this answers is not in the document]
dpc: Without having the actual implementation that needs these I don't think it's worth adding yet. AFAICT it's entirely backward-compat to add later.

elsirion (approved):

dpc:
Tiny merge conflict. Rebased.

justinmoon (approved):

Maan2003 (approved):

PR #4138, opened 2024-01-26
fix: recognise vetted gateways configured in base meta

okjodom:
Fedimint clients would only recognise [vetted gateways meta] field if configured via meta override url. This change allows clients to also read the same meta field on the base meta config. vetted gateways meta value from `meta_override_url` is preferred over value supplied from the base meta.

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -190,6 +190,19 @@ impl ClientConfig {
     pub fn federation_id(&self) -> FederationId {
         self.global.federation_id()
     }
+
+    /// Get the value of a given meta field
+    pub fn meta<V: serde::de::DeserializeOwned>(
+        &self,
+        key: &str,
+    ) -> Result<Option<V>, anyhow::Error> {
+        let Some(str_value) = self.global.meta.get(key) else {
+            return Ok(None);
+        };
+        serde_json::from_str(str_value)
+            .map(Some)
+            .context(format!("Decoding meta field '{key}' failed"))
+    }
 }
 
 /// The federation id is a copy of the authentication threshold public key of
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -22,11 +22,9 @@ use fedimint_client::oplog::UpdateStreamOrOutcome;
 use fedimint_client::sm::util::MapStateTransitions;
 use fedimint_client::sm::{DynState, ModuleNotifier, State, StateTransition};
 use fedimint_client::transaction::{ClientOutput, TransactionBuilder};
-use fedimint_client::{sm_enum_variant_translation, DynGlobalClientContext, FederationInfo};
+use fedimint_client::{sm_enum_variant_translation, DynGlobalClientContext};
 use fedimint_core::api::DynModuleApi;
-use fedimint_core::config::{
-    ClientConfig, FederationId, META_OVERRIDE_URL_KEY, META_VETTED_GATEWAYS_KEY,
-};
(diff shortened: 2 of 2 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -975,29 +966,37 @@ impl LightningClientModule {
...
+                None => {
+                    debug!("No meta override source configured");
+                    global_meta
                 }
             };
 
+            if meta.is_empty() {
+                debug!("No meta overrides found");
+                return Ok(gateways);
+            }
```
elsirion, line 994: This return path seems redundant?

elsirion (approved):

modules/fedimint-ln-client/src/lib.rs
okjodom: yes. need to refactor this a bit more, and there's a bug with the json parsing to fix

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -966,53 +979,75 @@ impl LightningClientModule {
...
-                Ok(meta) => {
-                    if meta.is_empty() {
-                        debug!("No meta overrides found");
-                        return Ok(gateways);
+            let global_meta: BTreeMap<String, String> = config.global.meta.clone();
+            let federation_id = config.global.federation_id();
+
+            let meta = match FederationInfo::from_config(config)
+                .await?
+                .meta::<String>(META_OVERRIDE_URL_KEY)?
```
elsirion, line 996: This looks like it might do more work than necessary, I assume you use this construct for the `meta` helper decoding fn? We could instead add a similar helper to the global config struct, which you already access a few lines above.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -966,53 +979,75 @@ impl LightningClientModule {
...
+            {
+                Some(override_src) => {
+                    match self.fetch_meta_overrides(federation_id, override_src).await {
+                        Ok(override_meta) => {
+                            if override_meta.is_empty() {
```
elsirion, line 1001: `override_meta` could be non-empty but not contain the trusted GW key, how is this handled?

okjodom: we need to fall back to global meta, yes

fedimint-core/src/config.rs
```diff
@@ -190,6 +190,20 @@ impl ClientConfig {
     pub fn federation_id(&self) -> FederationId {
         self.global.federation_id()
     }
+
+    /// Get the value of a given meta field
+    pub fn meta<V: serde::de::DeserializeOwned>(
+        &self,
+        key: &str,
+    ) -> Result<Option<V>, crate::encoding::DecodeError> {
```
elsirion, line 198: Why are you using `DecodeError` here? This isn't doing any binary encoding, where this error belongs to. By now it's an anyhow error anyway internally, so might as well just go with that.

elsirion (approved):
Looks ok, one nit

fedimint-core/src/config.rs
okjodom: ah yes. fixed!

elsirion (approved):

justinmoon (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-4138-to-releases/v0.2 origin/releases/v0.2
cd .worktree/backport-4138-to-releases/v0.2
git switch --create backport-4138-to-releases/v0.2
git cherry-pick -x 13a54346c3ebb5d63df4f735316a4be70b0afc64
```

PR #4139, opened 2024-01-26
chore: disable discord alerts for backwards-compatibility

bradleystachurski:
The discord `alerts` channel is too noisy with these alerts, especially while we know there are outstanding fixes.

### DIFF
--- a/.github/workflows/ci-backwards-compatibility.yml
+++ b/.github/workflows/ci-backwards-compatibility.yml
@@ -61,23 +61,3 @@ jobs:
           path: |
             /tmp/nix-build-*/
             !/tmp/nix-build-*/source/
-
-  notifications:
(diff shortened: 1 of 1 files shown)


dpc (approved):

Maan2003 (approved):

elsirion (approved):

justinmoon:
Do we plan to re-enable them in the future?

bradleystachurski:
> Do we plan to re-enable them in the future?

Yep 👍 

I think a sane approach is:
1. re-enable once all backwards-compatibility tests are fixed (only outstanding [PR](https://github.com/fedimint/fedimint/pull/4166))
2. once CI runs fast enough, make the workflow required to pass before merging

Does that sound reasonable?

Issue #4111, opened 2024-01-23
observation: rust-bitcoin does not impl `Encodable` for `Vec<T>` because they prefer `Vec<u8>` specialization for perf. reasons.


dpc:
https://github.com/rust-bitcoin/rust-bitcoin/issues/2390

You can't have both `impl Decodable for Vec<T : Decodable>` and `impl Decoable for Vec<u8>`. Implementing `Vec<u8>` in terms of reading things byte by byte is kind of slow. Not having `Vec<T>` inconvenient...

Something tells me we actually don't use `Vec<u8>` very often, but maybe I'm wrong. Something to remember, maybe we could add some workarounds in places where this matters.

Maan2003:
we can always replace `Vec<u8>` with a custom type `Bytes` if it is a big performance issue.

dpc:
Need to see if https://github.com/rust-bitcoin/rust-bitcoin/issues/2390#issuecomment-1906713853 works.

elsirion:
Wait, might that be the reason for our rather poor decode performance? :grimacing: It might not have been a problem at the beginning, we mostly avoided writing variable length byte strings, but now with a lot of TLV encoding it might actually become a problem.

dpc:
Possibly.

dpc:
https://github.com/fedimint/fedimint/pull/4145

PR #4145, opened 2024-01-26
chore: specialize `Encodable`/`Decodable` for bytes

dpc:
Re #4111

Re https://github.com/rust-bitcoin/rust-bitcoin/issues/2390

Due to Rust limitations, we can't have impls for both `Vec<T>` and `Vec<u8>`. Handling byte case in a generic way is potentially *very* inefficient, and we do use it in some places, thought it's hard to find all of them (and then prevent even more from being introduced).

The most appealing approach seems to be specializing that case using `TypeId` check and `unsafe`.

Cons:

* `TypeId` requires `'static` bound
* unsafe (but small and easy to reason about)

Pros:

* zero cost (at runtime)
* works automatically everywhere
* no loss in ergonomics

**Edit**:

In idle `just mprocs` I can't tell if there's any performance difference.

### DIFF
--- a/fedimint-core/src/encoding/mod.rs
+++ b/fedimint-core/src/encoding/mod.rs
@@ -10,6 +10,7 @@ mod tbs;
 #[cfg(not(target_family = "wasm"))]
 mod tls;
 
+use std::any::TypeId;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{Debug, Formatter};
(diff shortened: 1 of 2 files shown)


fedimint-core/src/encoding/mod.rs
```diff
@@ -248,17 +249,103 @@ macro_rules! impl_encode_decode_tuple {
...
+    writer.write_all(bytes)?;
+    Ok(bytes.len())
+}
+
+struct ReadBytesFromFiniteReaderOpts {
+    len: usize,
+    chunk_size: usize,
+}
+
+/// Read `opts.len` bytes from reader, where `opts.len` could potentially be
```
dpc, line 278: This reminds me I need to port https://github.com/rust-bitcoin/rust-bitcoin/blob/e2b9555070d9357fb552e56085fb6fb3f0274560/bitcoin/src/consensus/encode.rs#L304 to Fedimint ...

elsirion (approved):
In a not very scientific test this made running `fedimint-dbtool dump` 18% faster.

fedimint-core/src/encoding/mod.rs
```diff
@@ -248,17 +249,99 @@ macro_rules! impl_encode_decode_tuple {
...
+
+    Ok(ret)
+}
+
+/// Specialized version of Decodable for bytes
+pub fn consensus_decode_bytes<D: std::io::Read>(r: &mut D) -> Result<Vec<u8>, DecodeError> {
+    let len = u64::consensus_decode(r, &Default::default())?;
+
+    let len: usize =
+        usize::try_from(len).map_err(|_| DecodeError::from_str("size exceeds memory"))?;
```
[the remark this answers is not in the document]
dpc: This errors if trying to decode byte array larger than total theoretical virtual memory available for the process (which we wouldn't be able to allocate anyway).

dpc:
> In a not very scientific test this made running `fedimint-dbtool dump` 18% faster.

I'm happy to hear you were able to confirm this improves something. :D

tcharding:
> `TypeId` requires `'static` bound

That is a pretty big con, right? In a real life program when would one decode static things? (Excuse my ignorance but the only `'static` decodable things are static strings in tests, are they not?)

dpc:
> That is a pretty big con, right? In a real life program when would one decode static things? (Excuse my ignorance but the only `'static` decodable things are static strings in tests, are they not?)

AFAIR `T: 'static` means "any references held by this type need to be of static lifetime. Doesn't seem like a big con. It would interefere with zero-copy de/serialization, I guess.

fedimint-core/src/encoding/mod.rs
```diff
@@ -336,14 +431,29 @@ where
     }
 }
 
+// From <https://github.com/rust-lang/rust/issues/61956>
+unsafe fn horribe_array_transmute_workaround<const N: usize, A, B>(mut arr: [A; N]) -> [B; N] {
```
Maan2003, line 435: `#[inline(always)]` to make array doesn't have to be copied

fedimint-core/src/encoding/mod.rs
```diff
@@ -277,12 +360,16 @@ where
...
-    T: Decodable,
+    T: Decodable + 'static,
 {
     fn consensus_decode<D: std::io::Read>(
         d: &mut D,
         modules: &ModuleDecoderRegistry,
     ) -> Result<Self, DecodeError> {
+        if TypeId::of::<T>() == TypeId::of::<u8>() {
+            // unsafe: we've just checked that T is `u8` so the transmute here is a no-op
+            return Ok(unsafe { mem::transmute::<Vec<u8>, Vec<T>>(consensus_decode_bytes(d)?) });
```
Maan2003, line 371: wait you can't transmute `Vec` you have to go through `{to, from}_raw` methods

Maan2003: see https://doc.rust-lang.org/nomicon/transmutes.html?highlight=Vec%20transmute#transmutes

Maan2003: actually it is correct, because T is u8

Maan2003:
I think we can remove unsafe from this, using `dyn Any` api

Maan2003:
yes, I removed a lot of unsafe in 3779ff79f932fc3eac66458d852d5058e88c8dd8

but for &[T], rust complains about 'static

fedimint-core/src/encoding/mod.rs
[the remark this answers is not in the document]
elsirion: Right, but for all reasonable platforms these days `u64==usize`

fedimint-core/src/encoding/mod.rs
```diff
@@ -248,17 +249,99 @@ macro_rules! impl_encode_decode_tuple {
...
+    r.read_exact(bytes.as_mut_slice())
+        .map_err(DecodeError::from_err)?;
+    Ok(bytes)
+}
+
 impl_encode_decode_tuple!(T1, T2);
 impl_encode_decode_tuple!(T1, T2, T3);
 impl_encode_decode_tuple!(T1, T2, T3, T4);
 
 impl<T> Encodable for &[T]
```
Maan2003, line 332: we should impl for `[T]` instead of `&[T]`

Maan2003: but using Any still doesn't work
because you can't do .downcast_ref::<[u8]>(), because downcast_ref needs `Sized`

Maan2003: we could impl for Vec<u8> without unsafe, but not for `Box<[u8]>` or Arc, Rc

Kixunil:
To shed more light on why `'static` is required: Rust doesn't want to allow non-`'static` casting because it's not clear how type IDs should interact with lifetimes. Also while apparently there could be a safe downcasting function it's not expressible with current syntax (at least I think so):

```rust
fn downcast_ref<'a, T, U>(value:: &'a T) -> Option<&'a U> { /* ... */ }
```

There needs to be a bound requiring that `T` outlives `U` (assuming `T` and `U` are covariant) to ban this code:

```rust
fn do_stuff<T>(value: &T, s: &mut &'static str) {
    *s = *downcast_ref(value).unwrap();
}
```

But such bound is currently inexpressible.

This PR *wold be* sound even without the `'static` bound because instead of a generic `U` it uses a compile time known type `u8`, which never has any references (this is not the same as `'static`!) and thus always outlives itself. (So yes, having an imagined `U: 'always_static` bound above would work too but it'd still be more restrictive than necessary.) However there's no way to check if `T` is `u8` because you can't obtain `TypeId` for non-`'static` `T`.

dpc:
@Kixunil 100% matches my understanding.

dpc:
@Maan2003 I don't see a reason to worry about wrapping this `unsafe` into anything. These are the only places where it seems worthwhile and it's already small, isolated and easy to check that it's correct.

Maan2003 (approved):
Agreed

Maan2003:
wow github doesn't like this pr, maybe we broke something

dpc:
> wow github doesn't like this pr, maybe we broke something

It already passed before the CI. It must be some flakiness on `master`.

PR #4146, opened 2024-01-26
fix: ignore TracingSetup error in migration tests

m1sterc001guy:
Running `cargo test` causes an error with the db migration tests since the tracing subscriber is already setup. Simple fix is just to ignore the error.


dpc (approved):

elsirion (approved):
Sounds reasonable, did this lead to flakiness?

dpc:
> Sounds reasonable, did this lead to flakiness?

@elsirion Makes `cargo test` fail. CI is fine because `cargo nextest` we're using spawns every test in a different process (that's the secret sauce of it).

PR #4148, opened 2024-01-27
 fix: recovery waiting for the last session to close (mostly in tests) 

### LINKED ISSUE #4043: Waiting for recovery to finish needs to wait for the current block to end which is slow.
The recovery code needs to pesimize and assume even the current epoch (at the time the recovery started) can contain important changes. Because of that it needs to wait for the current sessions/block to finish, which can take considerable time (couple minutes). This is clearly visible in tests (run with `env RUST_LOG=debug` to see it timing out and waiting for the last block).

This could be greatly improved if we had a call that would return the current items in the current block.

dpc:
Use https://github.com/fedimint/fedimint/pull/4150 to avoid waiting for the final session to finish.

Fix #4043

### DIFF
--- a/fedimint-client/src/module/init/recovery.rs
+++ b/fedimint-client/src/module/init/recovery.rs
@@ -6,8 +6,8 @@ use fedimint_core::db::DatabaseTransaction;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::epoch::ConsensusItem;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
-use fedimint_core::module::ModuleCommon;
-use fedimint_core::session_outcome::SessionOutcome;
+use fedimint_core::module::{ApiVersion, ModuleCommon};
+use fedimint_core::session_outcome::{AcceptedItem, SessionStatus};
 use fedimint_core::task::{MaybeSend, MaybeSync};
 use fedimint_core::transaction::Transaction;
 use fedimint_core::{apply, async_trait_maybe_send, OutPoint};
@@ -100,9 +100,9 @@ pub trait RecoveryFromHistory: std::fmt::Debug + MaybeSend + MaybeSync + Clone {
         &mut self,
         client_ctx: &ClientContext<<Self::Init as ClientModuleInit>::Module>,
         _session_idx: u64,
-        session: &SessionOutcome,
+        session_items: &Vec<AcceptedItem>,
     ) -> anyhow::Result<()> {
-        for accepted_item in &session.items {
+        for accepted_item in session_items {
             if let ConsensusItem::Transaction(ref transaction) = accepted_item.item {
                 self.handle_transaction(client_ctx, transaction).await?;
             }
@@ -245,26 +245,40 @@ where
         /// errors via `sender` itself.
         fn fetch_block_stream<'a>(
             api: DynGlobalApi,
+            api_version: ApiVersion,
             decoders: ModuleDecoderRegistry,
             epoch_range: ops::Range<u64>,
-        ) -> impl futures::Stream<Item = (u64, SessionOutcome)> + 'a {
+        ) -> impl futures::Stream<Item = (u64, Vec<AcceptedItem>)> + 'a {
             // How many request for blocks to run in parallel (streaming).
             const PARALLISM_LEVEL: usize = 8;
+            const VERSION_THAT_INTRODUCED_GET_SESSION_STATUS: ApiVersion =
+                ApiVersion { major: 0, minor: 1 };
 
-            futures::stream::iter(epoch_range)
-                .map(move |block_idx| {
+            futures::stream::iter(epoch_range.clone())
+                .map(move |session_idx| {
                     let api = api.clone();
                     let decoders = decoders.clone();
                     Box::pin(async move {
-                        info!(block_idx, "Fetching epoch");
+                        info!(session_idx, "Fetching epoch");
 
                         let mut retry_sleep = Duration::from_millis(10);
                         let block = loop {
-                            info!(target: LOG_CLIENT_RECOVERY, block_idx, "Awaiting signed block");
-                            match api.await_block(block_idx, &decoders).await {
+                            info!(target: LOG_CLIENT_RECOVERY, session_idx, "Awaiting signed block");
+
+                            let items_res = if api_version <= VERSION_THAT_INTRODUCED_GET_SESSION_STATUS {
+                                api.await_block(session_idx, &decoders).await.map(|s| s.items)
+                            } else {
+                                api.get_session_status(session_idx, &decoders).await.map(|s| match s {
+                                    SessionStatus::Initial => panic!("Federation missing session that existed when we started recovery"),
+                                    SessionStatus::Pending(items) => items,
+                                    SessionStatus::Complete(s) => s.items,
+                                })
+                            };
+
+                            match items_res {
                                 Ok(block) => break block,
                                 Err(e) => {
-                                    info!(e = %e, block_idx, "Error trying to fetch signed block");
+                                    info!(e = %e, session_idx, "Error trying to fetch signed block");
                                     // We don't want PARALLISM_LEVEL tasks hammering Federation
                                     // with requests, so max sleep is significant
                                     const MAX_SLEEP: Duration = Duration::from_secs(120);
@@ -278,7 +292,7 @@ where
                             }
                         };
 
-                        (block_idx, block)
+                        (session_idx, block)
                     })
                 })
                 .buffered(PARALLISM_LEVEL)
@@ -290,7 +304,7 @@ where
             client_ctx: &ClientContext<<Init as ClientModuleInit>::Module>,
             common_state: &mut RecoveryFromHistoryCommon,
             state: &mut Recovery,
-            block_stream: &mut (impl Stream<Item = (u64, SessionOutcome)> + Unpin),
+            block_stream: &mut (impl Stream<Item = (u64, Vec<AcceptedItem>)> + Unpin),
         ) -> anyhow::Result<()>
         where
             Init: ClientModuleInit,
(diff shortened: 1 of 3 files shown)


dpc:
@elsirion @joschisan will keep it as a draft, so pinging you as you might not see it otherwise

elsirion:
Since old federations don't have this API endpoint this seems like a minor API version change? We should detect this, only use it if available and mark the code path in case it's unavailable as deprecated for future removal.

elsirion:
> If `get_session_outcome_or_pending_items` returned a tuple (current_session_count, pending_items), there would be no need to do a double session-count call before and after.

Good idea, we can mess around with the API on master till we cut a release (we should bump the API version anyway, I forgot that in the original PR).

joschisan:
The problem is that under load pending_accepted_items may never return via request_current_consensus. I would propose the following for recovery. instead of the two endpoints get_session_outcome and pending_accepted_items I would create an endpoint called current_accepted_items that takes a session index as an argument and returns a Vec<AcceptedItems> for the given session_index. The endpoint does the following: check if we have an outcome, if so return its items as a vector; then check if the index matches the current session index, if so return the pending accepted items as a vector. Otherwise, the requested index is higher than the current sessions index (which we assert!)at which point we either return and Error, None or an empty vector. 

We can now use this endpoint for recovery by checking the current session index first and then using current_accepted_items to obtain the history. For indices smaller then the current session index we will obtain the same result as await_session_outcome. In the current session we will either obtain the outcome directly if it has been completed while we downloaded the history, or we try to obtain the pending accepted items via request_current_consensus - however if we are under load we the query strategy will retry the current_accepted_items endpoint until the session completes and we can establish consensus.

joschisan:
I was just thinking maybe we should postpone this to 0.4? Since sessions are at most a minute and will stay that way for 0.3 this I only a little annoying in testing if I understand correctly @dpc? Seems to only really be a problem if sessions get longer… I was considering longer sessions for 0.4 with a synchronised update then we could still improve this. Would give us more time to think about this since endpoints are a little annoying to update.

dpc:
> Please find the endpoints implementation here: https://github.com/fedimint/fedimint/pull/4150

Commented, but generally :+1: approach.

> Since old federations don't have this API endpoint this seems like a minor API version change?

That's a good point.

> Seems to only really be a problem if sessions get longer… I was considering longer sessions for 0.4 with a synchronised update then we could still improve this.

I don't want to dwell too long and postpone things. We have gazillion other things to worry about, so IMO we should do the best we can think of, get it out of the way and move on. #4150 with that "closed vs not" flag would be perfect from all angles I can see. 

I will add the code to make that code conditional based on negotiated API version (if easy enough), or I'll make just make the code fall back to `await_session_outcome` if the `current_accepted_items` fails for any reason, which will make the test go brrrr and work even on old Federations.

dpc:
@elsirion @joschisan 

Depending on how we go about https://github.com/fedimint/fedimint/pull/4150 it I'll have to change this PR in different ways.

I'd like to go with #4150, but make it return `(Vec<AcceptedItem>, bool)` to tell the client if the session being returned is closed/open, to make caching possible. Then I can just use that for everything, and get rid of bunch of complexity and workarounds I added here, without loosing anything.

joschisan:
> I'd like to go with #4150, but make it return `(Vec<AcceptedItem>, bool)` to tell the client if the session being returned is closed/open, to make caching possible.

How about we use a SessionStatus enum to return three variants, Complete, Pending and Initial? That would be more expressive. With the bool we can not cover the third case..

dpc:
What's an Initial? :D

joschisan:
> What's an Initial? :D

Means the session is not yet started.

dpc:
@joschisan  Oh. In that case it's a bit of a confusing name. A `None`, `Missing`, `NotFound` or something? But a detail anyway, can be tweaked later.

Now that I thought about it... even more of a detail... maybe enum variants should be reversed? :D So in the encoding the session status beings with `0` (not opened yet), moves to `1` (in progress), `2` (sealed). :D

fedimint-core/src/api.rs
```diff
@@ -576,24 +611,63 @@ where
...
+        enum NoCacheErr {
+            Initial,
+            Pending(Vec<AcceptedItem>),
+            Err(anyhow::Error),
+        }
+        match entry_arc
+            .get_or_try_init(|| async {
+                match self.get_session_status_raw(session_idx, decoders).await {
+                    Err(e) => Err(NoCacheErr::Err(e)),
+                    Ok(SessionStatus::Initial) => Err(NoCacheErr::Initial),
```
dpc, line 655: Not in love in this inside-out-thingy, but it will do.

dpc:
So much better now. :)

fedimint-client/src/module/init/recovery.rs
```diff
@@ -247,24 +247,32 @@ where
...
+                        info!(session_idx, "Fetching epoch");
 
                         let mut retry_sleep = Duration::from_millis(10);
                         let block = loop {
-                            info!(target: LOG_CLIENT_RECOVERY, block_idx, "Awaiting signed block");
-                            match api.await_block(block_idx, &decoders).await {
+                            info!(target: LOG_CLIENT_RECOVERY, session_idx, "Awaiting signed block");
+
+                            let items_res = api.get_session_status(session_idx, &decoders).await.map(|s| match s {
+                                    SessionStatus::Initial => panic!("Federation missing a block that existed when we started recovery"),
```
elsirion, line 266: Is this possibly racy @joschisan? If we get the current session height and we are in between session the value returned could reasonably be `Initial` afaik? Or is the session switch atomic and initial will only be returned for strictly future sessions?

I don't think this matters too much because the session switch time is likely on the order of milli seconds and recovery will take much longer, just want to validate my mental model.

fedimint-client/src/module/init/recovery.rs
```diff
@@ -247,24 +247,32 @@ where
...
                     let api = api.clone();
                     let decoders = decoders.clone();
                     Box::pin(async move {
-                        info!(block_idx, "Fetching epoch");
+                        info!(session_idx, "Fetching epoch");
 
                         let mut retry_sleep = Duration::from_millis(10);
                         let block = loop {
-                            info!(target: LOG_CLIENT_RECOVERY, block_idx, "Awaiting signed block");
-                            match api.await_block(block_idx, &decoders).await {
```
elsirion, line 264: We cannot switch away from that API *in general*. I'd propose using `await_block` for API `0.0` (causing recovery to take longer) while using the new API for `0.1` so we can deprecate the `await_block` endpoint longer-term (should add a FIXME annotation).

For API version determination we can likely add a cached API call to `DynGlobalAPI`.

Great to see the compatibility tests catching this @bradleystachurski!

fedimint-core/src/api.rs
```diff
@@ -511,14 +517,28 @@ struct GlobalFederationApiWithCache<T> {
     /// synchronize, or split into a handful of groups. And if they are not,
     /// no LRU here is going to help them.
     #[allow(clippy::type_complexity)]
-    await_block_lru: Arc<tokio::sync::Mutex<lru::LruCache<u64, Arc<OnceCell<SessionOutcome>>>>>,
+    await_session_lru: Arc<tokio::sync::Mutex<lru::LruCache<u64, Arc<OnceCell<SessionOutcome>>>>>,
```
elsirion, line 520: This cache hasn't been released yet, right? Could we just remove it now that it's unused? We could also wait for the deprecation of the `await_block` API.

elsirion (changes requested):

fedimint-client/src/module/init/recovery.rs
dpc: Oh, right. Forgot about it. Again.

fedimint-core/src/api.rs
dpc: Oh, are we planning to remove `await_block`? Seemed like a useful API to have as it's safer to use (can't ever mishandle partially filed session) and the only way to actually cleanly block waiting for session to close without retrying. @elsirion

fedimint-client/src/module/init/recovery.rs
dpc: One benefit of https://github.com/fedimint/fedimint/pull/4159 is that session count and session status come from the same source in the database, so there should be any room for races anymore (thought I'm not 100% of the db isolation level thingies).

fedimint-client/src/module/init/recovery.rs
```diff
@@ -245,26 +245,40 @@ where
...
+                            info!(target: LOG_CLIENT_RECOVERY, session_idx, "Awaiting signed block");
+
+                            let items_res = if api_version <= VERSION_THAT_INTRODUCED_GET_SESSION_STATUS {
```
dpc, line 268: Look ma, I'm using api version auto-negotation!

fedimint-core/src/api.rs
elsirion: I mean, do we still use it anywhere after this PR?

I'd be ok with keeping it if you only wanted to fetch the last session via the new API, in which case that wouldn't really need caching unless we expect to have 100s of modules running in a client.

But if we fully switch to the new API as done in this PR afaik, then there is no need to keep the old one around.

dpc: Just today we were writing tests that require waiting for a session to close, so it seems useful as a general purpose API. @justinmoon was writing a test checking if with one peer offline sessions close.

dpc: Anyway, I think we should land, and decide to delete (or not) this endpoint elsewhere.

This cache is probably like 64B until something actually uses it, so no point in increasing the scope of the change anymore.

elsirion (approved):

justinmoon:
dev call: merge conflict and we need another review

joschisan (dismissed):

justinmoon (dismissed):

dpc:
Merge conflict, rebased.

elsirion (approved):

PR #4150, opened 2024-01-27
fix: remove racy endpoint PENDING_ACCEPTED_ITEMS_ENDPOINT for SESSION_STATUS_ENDPOINT

joschisan:


dpc (commented):
Generally fine with me and better than before, but one problem with this is that the result can't be cached.

If the response included some flag if this session is closed or not the it could (when closed).

dpc (dismissed):
:heart:

elsirion:
Do we actually need to cache this? I'd imagine to only use this endpoint for the last session we want to fetch. I think that's lower complexity overall than trying to adapt the cache to potentially non-final responses.

elsirion (dismissed):
LGTM

elsirion:
Also needs backporting since the previous version was backported too #4134 and we don't want two different APIs to live on in the next release.

@dpc don't we also have to bump the minor API version here?

dpc (approved):

dpc:
> @dpc don't we also have to bump the minor API version here?

Right.

dpc (approved):

fedimint-backports:
- #4180

elsirion:
I get the feeling we should also document when an API was introduced and enforce that via CI somehow. Otherwise it will be fairly hard to figure out later on.

dpc:
> I get the feeling we should also document when an API was introduced and enforce that via CI somehow.

Semgrep rule to put a specially formatted comment on every api_endpoint!?

Or maybe just an extra argument to the macro...

PR #4154, opened 2024-01-28
refactor: remove unused error paths, use expect when possible

joschisan:


modules/fedimint-mint-client/src/output.rs
```diff
@@ -220,68 +212,66 @@ impl MintOutputStatesCreated {
...
+                .into_iter()
+                .map(|(peer, share)| (peer.to_usize() as u64 + 1, share))
+                .collect(),
+        );
+
+        let amount_key = tbs_pks
+            .tier(&created.amount)
+            .expect("We obtained this amount from tbs_pks when we created the output");
+
+        // this implies that the mint client config's public keys are inconsistent
```
elsirion, line 238: So maybe in the future we can just check config consistency when joining the federation and then expect here?

elsirion (approved):
Great refactor!

modules/fedimint-mint-client/src/output.rs
joschisan: Yes we could

crypto/tbs/src/lib.rs
```diff
@@ -185,6 +185,14 @@ fn lagrange_multipliers(scalars: Vec<Scalar>) -> Vec<Scalar> {
         .collect()
 }
 
+pub fn verify_blinded_signature(
```
m1sterc001guy, line 188: Some docs might be useful here for everything that isnt familiar with the pairing crypto

modules/fedimint-mint-client/src/output.rs
```diff
@@ -220,68 +212,66 @@ impl MintOutputStatesCreated {
...
-                MintOutputStateMachine {
-                    common: old_state.common,
-                    state: MintOutputStates::Succeeded(MintOutputStatesSucceeded { amount }),
-                }
-            }
-            Err(error) => MintOutputStateMachine {
+        let agg_blind_signature = aggregate_signature_shares(
+            &blinded_signature_shares
+                .into_iter()
+                .map(|(peer, share)| (peer.to_usize() as u64 + 1, share))
```
m1sterc001guy, line 230: +1 here is just to prevent a peer with id 0?

joschisan: We need the +1 here so the cryptography checks out.

m1sterc001guy (approved):
I'm not the most well-versed in the mint module but this looks like a good refactor to me

modules/fedimint-mint-client/src/output.rs
elsirion: I knew what it was for but also raised an eye brow that there is no fn to convert the peer ID. But that's how I wrote it an eternity ago and Joscha just kept it that way.

Issue #3119, opened 2023-09-06
Expose helpful info about available modules from CLI


shaurya947:
Once #3118 gets merged we'll have the ability to call arbitrary commands that are specific to custom, non-default modules. As such it would be helpful to have a command that exposes all the available modules (both instance IDs + kinds) so it's easy to figure out which modules can be interacted with in a session.

Furthermore, it might also be worth investigating ways to make the `module` subcommand smarter such that it can use all the available modules to return more helpful error messages or even have something akin to an autocomplete etc.

15IITian:
I understood some part that we require a new command that list all the modules  available ( default and non-default ) .
If yes, please guide how to start with it.

dpc:
> I think a `fedimint module list` command would be good, then I'd use `fedimint module <module> <command> <args…>` for custom module commands. What do you think @dpc, I know you wanted shorter commands, but I think if we keep the shortcuts while allowing arbitrary commands to be called this way that would be ok.

Generally sounds good to me, short is good, but not most important.

> There should also be two optional, mutually exclusive flags `--federation` and `--client` that display the theoretically supported list of modules of the client and federation respectively, while the normal command only shown the modules supported by both.

I'm no sure if it's necessary. Client knows about modules it doesn't support, so could still list them with some info.

```
> fedimint-cli module list
{
   "0": {
     "kind": "bazinga",
     "status": "unsupported",
    },
   "1": {
      "kind": "mint"
      "status": "recovery",
      "progress": {
         "complete": 41,
         "total": 666
       }
       /// possibly some other most important fields
    },
    "2": {
       "kind": "wallet",
       "status": "active",
       // possibly more
    }
}
```

elsirion:
> I'm no sure if it's necessary. Client knows about modules it doesn't support, so could still list them with some info.

Good point! I was thinking about making the CLI output human readable by default in the future, but we can make this a future problem. We use JSON everywhere anyway for now.

15IITian:
After deep dive into the code , I have deduced some conclusions , Please do tell what I am missing out -> 
 * There is `ServerModuleRegistry` struct which contains all the server-related modules ( I think server== federation)
* same for `ClientModuleRegistry` struct containing client-related modules things.
* There are functions defined on them in `fedimint-core/registry.rs` file which can be used to iterate over the modules data.

15IITian:
> I'm no sure if it's necessary. Client knows about modules it doesn't support, so could still list them with some info.
[...]

PR #4158, opened 2024-01-28
feat(fedimint-cli): create fedimint-cli module list subcommand

15IITian:
* Created `ModuleStatus` struct containing status of a module, `ModuleSubcommand `enum containing `List` variant
 * Created `Data` struct which contains module related data
 * On running `fedimint-cli module list` command -> will store all federation supported modules in a Vector of `Data` Struct.
 
 #3119

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -10,7 +10,7 @@ use bitcoin_hashes::hex::ToHex;
 use clap::Subcommand;
 use fedimint_client::backup::Metadata;
 use fedimint_client::ClientArc;
-use fedimint_core::config::FederationId;
+use fedimint_core::config::{ClientModuleConfig, FederationId};
 use fedimint_core::core::{ModuleInstanceId, ModuleKind, OperationId};
 use fedimint_core::encoding::Encodable;
 use fedimint_core::time::now;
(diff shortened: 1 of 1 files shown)


15IITian:
image

To access client-supported modules -> I try to take `modules` field from the `client` variable -> as  `modules: ClientModuleRegistry` which has a function `iter-modules`  to  get all modules.

But `Client` struct fields are `private` and I  found no such function in its implementation to iterate over modules .

15IITian:
as per https://github.com/fedimint/fedimint/issues/3119#issuecomment-1913091136 -> I have not included `recovering`  for now.

fedimint-cli/src/client.rs
```diff
@@ -161,15 +161,25 @@ pub enum ClientCmd {
     },
     /// Call a module subcommand
     Module {
+        #[clap(subcommand)]
+        command: ModuleSubcommand,
         /// Module selector (either module id or module kind)
-        #[clap(long)]
-        module: ModuleSelector,
+        module: Option<ModuleSelector>,
         args: Vec<ffi::OsString>,
```
Maan2003, line 168: likely subcommand and args are incompatible

15IITian: is it because of making it a postional argument ?

Maan2003: no, just incompatible
args should be part of subcommand

15IITian: @Maan2003 what can be the problem now ?

15IITian: I think,  backward incompatibility test fails due to this error  ->

image


Please tell how to remove it.

justinmoon:
No progress in 3 weeks. Closing.

PR #4159, opened 2024-01-29
cleanup: remove redundant session counter from db

joschisan:


elsirion:
From our call: make revert of 62b6df4 more explicit and split the PR into two commits.

fedimint-server/src/consensus/server.rs
```diff
@@ -744,8 +734,11 @@ impl ConsensusServer {
 }
 
 pub(crate) async fn get_finished_session_count_static(dbtx: &mut DatabaseTransaction<'_>) -> u64 {
-    dbtx.get_value(&SignedSessionOutcomeCountKey)
+    dbtx.find_by_prefix_sorted_descending(&SignedSessionOutcomePrefix)
+        .await
+        .next()
         .await
+        .map(|entry| (entry.0 .0) + 1)
```
Maan2003, line 741:
```suggestion
        .map(|entry| (entry.0.0) + 1)
```
that space really messed up my brain

joschisan: Yes, but cargo fmt insists on it.

fedimint-server/src/consensus/server.rs
```diff
@@ -744,8 +734,11 @@ impl ConsensusServer {
 }
 
 pub(crate) async fn get_finished_session_count_static(dbtx: &mut DatabaseTransaction<'_>) -> u64 {
-    dbtx.get_value(&SignedSessionOutcomeCountKey)
+    dbtx.find_by_prefix_sorted_descending(&SignedSessionOutcomePrefix)
```
Maan2003, line 737: make sure to measure it

fedimint-server/src/db.rs
```diff
@@ -105,25 +89,7 @@ impl_db_record!(
...
-
-    let session_count = dbtx
-        .find_by_prefix(&SignedSessionOutcomePrefix)
-        .await
-        .count()
-        .await as u64;
-    dbtx.insert_new_entry(&SignedSessionOutcomeCountKey, &session_count)
-        .await;
-    Ok(())
+    MigrationMap::new()
```
m1sterc001guy, line 92: Why are we removing the migration? Wasn't this just recently added?

joschisan: Yes but the added db entry was redundant and we found a way to make it unnecessary.

fedimint-server/src/consensus/server.rs
dpc: While better than counting, I think the underlying implementation will race ahead and start reading more than just one key, making this call actually generate a heavier IO load than expected.

fedimint-server/src/consensus/server.rs
dpc: Even if it responds here fast, it might have a higher underlying cost which will be hard to measure here. (see other comment).

fedimint-server/src/db.rs
elsirion: Since this wasn't part of any official release yet we can remove it.

fedimint-server/src/consensus/server.rs
m1sterc001guy: Rocksdb should be caching pages that have been recently written to. This entry is written every session (~5mins) right? Hard to know the actual IO cost without benchmarking it at all but if its significant IO that would be surprising to me

m1sterc001guy (approved):

fedimint-server/src/consensus/server.rs
dpc: I would slap in-memory cache on top of the whole thing, so this code doesn't even need to do any IO except a first time each session.

dpc (approved):

fedimint-server/src/consensus/server.rs
elsirion: Let's see how it behaves.

justinmoon:
@elsirion We need to figure out how to backport this

elsirion:
> @elsirion We need to figure out how to backport this

Opened #4203

Issue #3844, opened 2023-12-05
Remove "active gateway" concept


justinmoon:
### Update (March 1)

We decided to remove the "active gateway" concept so edited the title

### Old issue description

In the past, the client would try to cache one of the "gateway registration" locally and always try to use that gateway. Now it doesn't (what I was observing in https://github.com/fedimint/fedimint/issues/3479, which I'm closing and broadening the discussion here).

A benefit of this approach was that we didn't have to do a bunch of network requests to the federation while paying a lightning payment (https://github.com/fedimint/fedimint/issues/3842). It also provided a (somewhat primitive) hook for client consumers to choose which gateway the client would use.

Right now on `master` the `fedimint-cli list-gateways` doesn't even return an active gateway in mprocs:

```
$ just mprocs
$ fedimint-cli list-gateways
[
  {
    ...
    "active": false
  },
  {
    ..
    "active": false
  }
]
```

`fedimint-cli ln-pay ...` still [makes a call to select_active_gateway](https://github.com/fedimint/fedimint/blob/af69ced9118ee9a997f9ea444edb0a6b5e6beb07/fedimint-cli/src/client.rs#L242). This used to hydrate the  "active gateway cache" mentioned in the first paragraph, but now it accomplishes nothing because it doesn't write to the active gateway key so now `fedimint-cli ln-pay` makes 3 requests to fetch list of gateway (this one plus the two mentioned in https://github.com/fedimint/fedimint/issues/3842). That's 12 total websocket requests in a 3/4 federation because `list_gateways` employs the `UnionResponses` strategy.

Lastly, right now it's unclear which gateway will be used when paying or receiving lightning payment. As a result, it's unclear how a `fedimint-client` consumer should display fees that will be charged for the payment. It is possible, however, by calling `set_active_gateway`. A current downside of this is that `set_active_gateway` calls `fetch_registered_gateways` so consumers will be doing a lot of network requests just to make sure this registration is up-to-date. What happens if the registration stored in `LightningGatewayKey` get out-of-date? The `LightningGateway` struct that is stored in this key doesn't contain any registration information so I'm not even sure to tell whether it is expired or not?

To summarize:
* "active gateway" doesn't have consistent behavior
* Way too many network requests. 
  * If we want to check whether a gateway is still there, a request to the gateway itself requires less bandwidth and likely less latency than a request to the federation.
[...]

PR #4162, opened 2024-01-29
fix: make set_active_gateway concurrency safe

### LINKED ISSUE #4110: set_active_gateway is not concurrency safe
https://github.com/fedimint/fedimint/blob/11339763400401b5394e726b05bf6decc031283e/modules/fedimint-ln-client/src/lib.rs#L866-L888

if called concurrently, second one would panic on database commit

Maan2003:
fixes #4110 

~~just put a lock around it for now~~

return error on concurrent calls

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -883,7 +883,7 @@ impl LightningClientModule {
 
         dbtx.insert_entry(&LightningGatewayKey, &gateway.anchor())
             .await;
-        dbtx.commit_tx().await;
+        dbtx.commit_tx_result().await?;
         Ok(())
     }
 


Maan2003:
i am conflicted with return Error and using lock

error might help you find concurrent calls with set_active_gateway, but lock feels more correct semantically

elsirion:
Why not get rid of `set_active_gateway` altogether?

elsirion:
> i am conflicted with return Error and using lock

Why not auto-commit, if you want to avoid the error case? I don't think it makes sense though, just as the lock version. 

In order of preference:

* Remove the active GW concept
* Fail on concurrent writes
* Autocommit
* Locking

Maan2003:
we still have to keep active gateway in fedimint-cli because of devimint

elsirion (approved):

justinmoon:
Relevant discussion in https://github.com/fedimint/fedimint/issues/3844

m1sterc001guy (approved):
Propagating the error up is IMO the best short term solution

PR #4166, opened 2024-01-29
fix: make devimint backup & restore test backwards-compatible

bradleystachurski:
Fixes [#4106](https://github.com/fedimint/fedimint/issues/4016)

For visibility, resurfacing this comment from the associated issue:

> Should we adapt the test script instead here? At least in my mind we didn't commit to CLI stability (maybe we should? @justinmoon)

https://github.com/fedimint/fedimint/issues/4016#issuecomment-1893662992

This approach resolves by adapting the test script and doesn't change the CLI.

PR #4167, opened 2024-01-29
fix: back-compat ci tests are marking everything as :red-x:

dpc:
And it's just annoying. :shrug:

image

I look at my PRs and it's not clear to me if something failed in them or not. On the other hand when looking at PRs I could review, I can't tell if they are still being fixed or passing.

### DIFF
--- a/.github/workflows/ci-backwards-compatibility.yml
+++ b/.github/workflows/ci-backwards-compatibility.yml
@@ -4,15 +4,16 @@ name: "Backwards-Compatibility"
 
 # Controls when the workflow will run
 on:
-  # Triggers the workflow on push or pull request events but only for the "main" branch and release tags
-  push:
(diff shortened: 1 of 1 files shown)


bradleystachurski:
This fixes the last broken test. Would you still prefer to disable all backwards-compatibility tests?

https://github.com/fedimint/fedimint/pull/4166

dpc:
> This fixes the last broken test. Would you still prefer to disable all backwards-compatibility tests?
> 
> #4166

Let's wait then. :)

justinmoon (approved):

bradleystachurski (approved):
Might be worth keeping this PR around and approved if we want to disable CI quickly, even though we're going to see how things go after all the backwards-compatibility fixes are merged

Maan2003 (approved):

dpc:
Will let it slide... this time. ;)

Issue #1938, opened 2023-03-15
Verify that MPP / AMP payments don't cause problems for the gateway


justinmoon:
We think that the our invoices set flags to signal that we DON't accept AMP / MPP payments. We should verify this. Perhaps have a tests that try to do MPP / AMP and assert that the node will refuse to pay it because our invoice doesn't signal support.

justinmoon:
I think we could create an automated test for this. 

Spawn 2 lightning nodes between the payer and recipient gateway. Paying node and gateway have direct channels to both of these nodes, but not to eachother. Then try to send payments greater than individual channel balances but less than sum of the 2 channel balances forcing mpp / amp. Perhaps there's also some more manual way of telling the node to use AMP / MPP ...

dpc:
@m1sterc001guy Please reopen if still a problem.

m1sterc001guy:
During invoice creation we indicate that MPP/AMP is not supported, so this is fine to close.

We are considering supporting this in LNv2 but need to carefully consider some edge cases.

PR #4169, opened 2024-01-29
feat: set mpp for generated invoices

benthecarman:
Noticed this was missing from when building the invoice, MPP should be enabled so it can be paid with MPP. Should in general improve reliability with receives


justinmoon:
Relevant issue: https://github.com/fedimint/fedimint/issues/1938

benthecarman:
> Relevant issue: #1938

Ah yeah, don't merge this until it's been verified that the gateway handles MPP properly.

elsirion:
Don't you need to decrypt the last-hop onion to get some secret for MPP to work? We currently just throw that away because it's encrypted with a key only known to the user who is potentially offline at receive time.

Issue #4171, opened 2024-01-30
Ability to run full test suite (rust and devimint) where all 4-guardian federations have 1 offline


justinmoon:
As discussed in the dev call today, we can figure out when exactly our CI should run these checks. But we should at least have a simple way to run this locally.

Issue #4130, opened 2024-01-25
Client becomes slow if one guardian is down


elsirion:
Possibly related to #4064

justinmoon:
dev call: we need more testing on degraded federations. For example, 3/4 guardians online.

dpc:
Seems to me like we've enabled tests with degraded peers, and that's probably enough to spot any huge performance degradation, so closing this one.

PR #4173, opened 2024-01-30
feat: latency tests on a degraded federation

maan2003:
not completely implmenting #4171

this increases latency significantly

related https://github.com/fedimint/fedimint/issues/4130

healthy federation:

```
REISSUE: min: 1.8s, avg: 1.9s, median: 1.9s, p90: 2.0s, max: 2.0s, sum: 56.8s
LN SEND: min: 2.9s, avg: 3.1s, median: 3.0s, p90: 4.0s, max: 4.1s, sum: 62.3s
LN RECV: min: 1.9s, avg: 2.2s, median: 2.1s, p90: 2.6s, max: 3.1s, sum: 43.8s
FM PAY : min: 4.3s, avg: 4.4s, median: 4.4s, p90: 4.5s, max: 4.6s, sum: 87.7s
RESTORE: 60.401258381s
```

degraded federation: still running after 30 minutes, i gave up

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -1750,7 +1750,10 @@ pub async fn recoverytool_test(dev_fed: DevFed) -> Result<()> {
 pub enum TestCmd {
     /// `devfed` then checks the average latency of reissuing ecash, LN receive,
     /// and LN send
-    LatencyTests,
+    LatencyTests {
+        #[arg(long)]
+        degraded: bool,
+    },
     /// `devfed` then kills and restarts most of the Guardian nodes in a 4 node
     /// fedimint
     ReconnectTest,
@@ -1813,9 +1816,13 @@ pub async fn handle_command(cmd: TestCmd, common_args: CommonArgs) -> Result<()>
             };
             cleanup_on_exit(main, task_group).await?;
         }
-        TestCmd::LatencyTests => {
+        TestCmd::LatencyTests { degraded } => {
+            let fed_size = common_args.fed_size;
             let (process_mgr, _) = setup(common_args).await?;
-            let dev_fed = dev_fed(&process_mgr).await?;
+            let mut dev_fed = dev_fed(&process_mgr).await?;
+            if degraded {
+                dev_fed.fed.terminate_server(fed_size - 1).await?;
+            }
             latency_tests(dev_fed).await?;
         }
         TestCmd::ReconnectTest => {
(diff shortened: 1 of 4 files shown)


Maan2003:
if i reduce iteration count, it gets stuck on recovery
cc @dpc

Maan2003:
gateways were failing

justinmoon:
@Maan2003 Seems like CI doesn't invoke this test correctly https://github.com/fedimint/fedimint/actions/runs/7707693732/job/21005321860#step:9:64

justinmoon:
Added a commit awaiting session completion in the latency test

dpc (approved):

justinmoon:
@Maan2003 this needs a rebase

devimint/src/tests.rs
```diff
@@ -234,6 +234,27 @@ pub async fn latency_tests(dev_fed: DevFed) -> Result<()> {
...
+            .run()
+            .await
+        {
+            Err(e) => Err(ControlFlow::Continue(e)),
+            Ok(_) => Ok(()),
+        }
+    })
+    .await?;
+    let session_found_in = start.elapsed();
+
```
elsirion, line 257: I don't think this measures what you expect? If you are *somewhere* within a session it could take any time between 0s-3m to finish it. If you want to measure session completion time you have to start measuring at the start of a session.

Maan2003: it is not measuring session completion, it is just an assertion that a session completed

dpc: If you display the time in the output it does kinda looks like you were measuring it... :D

justinmoon: This test could have a concurrent task that just measures session durations in parallel with the rest of the test.

Maan2003: i have removed this commit for now, lets do it in a separate PR

dpc (approved):

justinmoon (approved):

Maan2003:
oh backwards compat tests won't work because 0.2.1 has broken degraded federation

elsirion:
Can we just deactivate the backwards compat tests for it for now?

bradleystachurski:
> Can we just deactivate the backwards compat tests for it for now?

Yep, we can wrap in a check similar to: https://github.com/fedimint/fedimint/pull/4173/files#diff-e5c7cc37f7292b713d50077a8badf27398452732e2042c0bb9a57b4d8346fa81R57-R60

bradleystachurski:
dev call: Let's go with https://github.com/fedimint/fedimint/pull/4247 since it runs all tests degraded instead of only latency.

devimint/src/tests.rs
bradleystachurski: Tracked in https://github.com/fedimint/fedimint/issues/4208

PR #4182, opened 2024-01-30
debug: consensus issues [don't merge]

dpc:
image

This sleep there never completes, as the alternative branch receives every second a batch of 0 bytes, even when no user does anything. Is this expected?

Anyway, when running:

```
env RUST_LOG=client::net::api=trace,consensus=debug,info just mprocs
```

and then `ps fax | grep fedimintd` and `kill <pid>` one of the peers, we no longer get new sessions, which seems to repro some consensus issue.

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -732,7 +732,9 @@ impl ConsensusServer {
 
         loop {
             // we wait until we have stalled
+            info!(target: LOG_CONSENSUS, "### SLEEP START");
             sleep(Duration::from_secs(5)).await;
+            info!(target: LOG_CONSENSUS, "### SLEEP END");
 
             let result = federation_api
                 .request_with_strategy(
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -428,16 +428,31 @@ impl ConsensusServer {
         while num_batches < batches_per_session_outcome {
             tokio::select! {
                 unit_data = unit_data_receiver.recv() => {
-                    if let (UnitData::Batch(bytes), peer) = unit_data? {
-                        if let Ok(items) = Vec::<ConsensusItem>::consensus_decode(&mut bytes.as_slice(), &self.decoders()){
+                    let (unit, peer) = match unit_data {
+                        Ok((unit, peer)) => (unit, peer),
+                        Err(err) => {
+                            warn!(target: LOG_CONSENSUS, item_index, "Unit receiving error");
+                            return Err(err.into());
+                        }
+                    };
+                    debug!(target: LOG_CONSENSUS, item_index, %peer, "Received data unit");
+                    if let UnitData::Batch(bytes) = unit {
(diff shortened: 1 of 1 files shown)


dpc:
@elsirion @joschisan Please halp.

joschisan:
> This sleep there never completes, as the alternative branch receives every second a batch of 0 bytes, even when no user does anything. Is this expected?

Yes this is the intended behavior. The sleep is only supposed to complete if we have fallen so far behind our peers that the already moved on to the next session or more. In that case, we will not finalise more batches, the sleep times out and we download the signed session outcome from our peers.

joschisan:
So @elsirion and I found a way to fix this, probably without increasing the minor consensus version. The issue was that we wrongly calculated the the cutoff for the session switch and therefore the session runs into exponential slowdown, where it orders items slower and slower and never ends. We can fix this by increasing the exponential slowdown offset by the factor 3.

elsirion:
The fix I pushed still needs to be validated that it can be applied without a synchronized shutdown:

* Start federation with old code
* Generate some CIs in session
* Shut down all `fedimintd`s
* Compile version with fix
* Restart `fedimintd`s with fix
* See that the session finishes and the federatio keeps running

joschisan:
> The fix I pushed still needs to be validated that it can be applied without a synchronized shutdown:

Here's a clean implementation with updated comments: https://github.com/fedimint/fedimint/pull/4188

Maan2003:
for validating the fix

https://github.com/Maan2003/fedimint/commits/degraded-fixes/

dpc:
Great to see it reprod and fix pending, but I've got to say that:

> We can fix this by increasing the exponential slowdown offset by the factor 3.

sounds like wonky fix / wonky failure mode. :D . Depending on time is just... brittle. Hosts can get paused by hypervisor, or share a machine with some heavy load and get slower, sleep longer then requested, etc. Sorry for not understanding things better and not being able to offer anything more constructive.

One way or the other would be good to have some kind of e2e devimint test to make peers do wonky stuff (submit things faster, submit them slower, stop submitting for time being etc.) and make sure things are working in all circumstance. Could be achieved with a couple of internal env var flags to simulate some things malfunctioning.

joschisan:
> > We can fix this by increasing the exponential slowdown offset by the factor 3.
> 
> sounds like wonky fix / wonky failure mode. :D . Depending on time is just... brittle. 

The exponential slowdown offset is not a time but a number of rounds. It will now work as it was intended before, but sessions wil stilll be three times as long as originally intended - hence about 2:15 to 3:00 minutes instead of 45-60s. Therefore a sessions maximal theoretical ram consumption is also three times higher at 750mb for every peer, though this would only be reached in an adversarial setting. However, that still can pose a problem in federations with more than 4 guardians and I would like to lower the number at some point. However, we choose this fix for now as it’s the only option that does not require a coordinated shutdown but can simply be released with 0.2.2. I will come back to this as soon as we do coordinated upgrades.

dpc:
I guess this is OK to close. Feel free to reopen if you need it open longer.

PR #4184, opened 2024-01-31
fix(justfile): don't use `cargo -p`

dpc:
It causes different features/deps being selected, which causes rebuilds of almost whole project, which can be painful.

Instead of `cargo test -p pkg` it's better to use
`cargo nextest run -E 'package(pkg)'`.

### DIFF
--- a/fedimint-server/src/db.rs
+++ b/fedimint-server/src/db.rs
@@ -114,7 +114,7 @@ mod fedimint_migration_tests {
     use fedimint_dummy_common::{DummyCommonInit, DummyInput, DummyOutput};
     use fedimint_dummy_server::Dummy;
     use fedimint_logging::TracingSetup;
-    use fedimint_testing::db::{prepare_db_migration_snapshot, validate_migrations, BYTE_32};
+    use fedimint_testing::db::{snapshot_db_migrations, validate_migrations, BYTE_32};
     use futures::StreamExt;
     use rand::rngs::OsRng;
     use secp256k1_zkp::Message;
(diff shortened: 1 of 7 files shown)


dpc:
@m1sterc001guy fyi

dpc:
@m1sterc001guy Idea. Could generating migration aferwards run `just ....` to test the generated migration (which should be a separate command).

dpc:
Did a bunch of renaming to unify stuff. I'm open for pushback.

fedimint-server/src/db.rs
```diff
@@ -114,7 +114,7 @@ mod fedimint_migration_tests {
     use fedimint_dummy_common::{DummyCommonInit, DummyInput, DummyOutput};
     use fedimint_dummy_server::Dummy;
     use fedimint_logging::TracingSetup;
-    use fedimint_testing::db::{prepare_db_migration_snapshot, validate_migrations, BYTE_32};
+    use fedimint_testing::db::{snapshot_db_migrations, validate_migrations, BYTE_32};
```
dpc, line 117: I decided to use `snapshot` as a verb to make it more uniform and shorter.

elsirion (approved):

m1sterc001guy (approved):
Ack. Good improvement

m1sterc001guy:
> I generated new migration, now I'd like to see if it worked and need to do it manually...

For development it has been convenient for me to generate the snapshot first, just to confirm that the test was written correctly, then run the validation. While developing its just slightly more confusing to run both at the same time since it takes a tad more work to track down where the problem is.

This is just a development workflow though so I can temporarily change the justfile to support that for myself. Your change here is more convenient for everyone else.

PR #4187, opened 2024-01-31
refactor(gateway): make `gateway-cli ` commands names similar to `fedimint-cli`

### LINKED ISSUE #1561: Unify `gateway-cli` and `fedimint-cli` commands
These two should behave generally very similarly.

* `connect-fed` vs `join-federation,`
* `--work-dir` vs federationid (positional, meh)
* `connect-info` vs `info`
* lot of other commands

15IITian:
fix #1561

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -41,7 +41,7 @@ pub enum Commands {
         federation_id: FederationId,
     },
     /// Generate a new peg-in address, funds sent to it can later be claimed
-    Address {
+    DepositAddress {
         #[clap(long)]
         federation_id: FederationId,
     },
@@ -57,7 +57,7 @@ pub enum Commands {
         address: Address,
     },
     /// Register federation with the gateway
-    ConnectFed {
+    JoinFederation {
         /// InviteCode code to connect to the federation
         invite_code: String,
     },
@@ -130,7 +130,7 @@ async fn main() -> anyhow::Result<()> {
 
             print_response(response).await;
         }
-        Commands::Address { federation_id } => {
+        Commands::DepositAddress { federation_id } => {
             let response = client()
                 .get_deposit_address(DepositAddressPayload { federation_id })
                 .await?;
@@ -152,7 +152,7 @@ async fn main() -> anyhow::Result<()> {
 
             print_response(response).await;
         }
-        Commands::ConnectFed { invite_code } => {
+        Commands::JoinFederation { invite_code } => {
             let response = client()
                 .connect_federation(ConnectFedPayload { invite_code })
                 .await?;


justinmoon:
@bradleystachurski will this break backwards compatibility tests?

bradleystachurski (commented):
> will this break backwards compatibility tests?

Yep, this breaks compatibility with the CLI. It's still an open question what our philosophy is for CLI stability, but I think we should avoid incompatible changes unless there's a material benefit.

> Should we adapt the test script instead here? At least in my mind we didn't commit to CLI stability (maybe we should? @justinmoon)

https://github.com/fedimint/fedimint/issues/4016#issuecomment-1893662992

For visibility, this is the failure from CI

```
00:01:12     0: command: gateway-cli --rpcpassword=theresnosecondbest -a http://127.0.0.1:19421/v1 connect-fed fed11qgqpw9thwvaz7te3xgmjuvpwxqhrzw33xgurxd30qqqjqsljxgwp2q73rzk58n0jzdw04fvlduekzdnfuuaxcnvj5n4zfajpydcezc
00:01:12     1: exit status: 2
00:01:12        stdout:
00:01:12        
00:01:12        stderr:
00:01:12        error: unrecognized subcommand 'connect-fed'
00:01:12        
00:01:12        Usage: gateway-cli [OPTIONS] <COMMAND>
00:01:12        
00:01:12        For more information, try '--help'.
```

https://github.com/fedimint/fedimint/actions/runs/7728708478/job/21070150985?pr=4187#step:5:4878

If CI was able to progress, we'd also bump into issues with `address` changing to `deposit-address`.

bradleystachurski:
> I think we should avoid incompatible changes unless there's a material benefit.

As a followup, I don't have a strong opinion on unifying these names with `fedimint-cli`. If we think it's worth introducing a breaking change we can handle versions specifically in devimint similar to https://github.com/fedimint/fedimint/blob/3a020c09a01046e6d6c0c98d8e77a67fa3425d02/devimint/src/main.rs#L454

15IITian:
@bradleystachurski 
I hope this  question isn't  silly   ,I am curious if there is a way to anticipate potential risk of backward incompatibility test to fail before making any changes?

elsirion:
> I hope this question isn't silly ,I am curious if there is a way to anticipate potential risk of backward incompatibility test to fail before making any changes?

Not a silly Q, but you'd need to know our test suite for that. In general changing APIs (CLI parsing, RPCs, …) should always break our test suite, otherwise we are missing test coverage. Now that we want to provide more compatibility guarantees this makes certain minor renames rather unattractive or at least a lot of work.

bradleystachurski:
Yep, I agree with what Eric said. Also, if you're curious if your changes break backwards-compatibility prior to opening a PR, you can run the tests locally with

```
./scripts/tests/backwards-compatibility-test.sh
```

15IITian:
> ./scripts/tests/backwards-compatibility-test.sh

Thanks for it.

justinmoon:
No progress in 3 weeks. Closing.

PR #4188, opened 2024-01-31
fix: calculate the batches per sessions correctly to achieve fault tolerance

joschisan:


dpc (approved):
Probably @elsirion should review before merging, but might as well tick for the second reviewer.

Maan2003:
you can validate it using https://github.com/Maan2003/fedimint/commits/consensus-test-old

./scripts/tests/consensus-fix.sh /path/to/old-fedimintd

justinmoon:
How are we going to test this change works as intended? I would like to see a test where a 3/4 federation with 1 guardian offline produces a session. https://github.com/fedimint/fedimint/pull/4198 may help with this.

fedimint-server/src/consensus/server.rs
```diff
@@ -323,29 +323,30 @@ impl ConsensusServer {
...
-
         // In order to bound a sessions RAM consumption we need to bound its number of
         // units and therefore its number of rounds. Since we use a session to
         // create a threshold signature for the corresponding session outcome we have to
         // guarantee that an attacker cannot exhaust our memory by preventing the
         // creation of a threshold signature, thereby keeping the session open
-        // indefinitely. Hence we increase the delay between rounds exponentially
-        // such that MAX_ROUND would only be reached after roughly 350 years.
+        // indefinitely. Hence, we increase the delay between rounds exponentially
+        // such that max_round would only be reached after a minimum of 100 years.
```
elsirion, line 332: That is enforced by AlephBFT on startup, right?

elsirion (approved):
Looks equivalent to 051cb734391686f45db75455411ca2371fea5a44, which made sense to me.

elsirion:
We really need to fix CI :sob:

fedimint-server/src/consensus/server.rs
joschisan: Yes

fedimint-backports:
- #4212

PR #4189, opened 2024-01-31
fix: remove non-fault tolerant query strategy

joschisan:
Requiring full participation in this one place is unexpected behaviour in a bft system. All query strategies should be bft.

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -30,7 +30,6 @@ use fedimint_core::config::{ClientConfig, FederationId};
 use fedimint_core::core::OperationId;
 use fedimint_core::db::{Database, DatabaseValue};
 use fedimint_core::module::{ApiAuth, ApiRequestErased};
-use fedimint_core::query::ThresholdConsensus;
 use fedimint_core::util::SafeUrl;
 use fedimint_core::{task, PeerId, TieredMulti};
 use fedimint_ln_client::LightningClientInit;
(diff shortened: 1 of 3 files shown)


fedimint-core/src/query.rs
```diff
@@ -226,9 +226,7 @@ pub struct ThresholdConsensus<R> {
 }
 
 impl<R> ThresholdConsensus<R> {
-    // Require that enough participants return the same message to ensure that we
-    // overcome the threshold for malicious nodes
-    pub fn overcome_evil(total_peers: usize) -> Self {
```
dpc, line 231: LOL, that was a very heroic name.

dpc (dismissed):

justinmoon (dismissed):

dpc:
```
        >   --> fedimint-cli/src/lib.rs:33:5
       >    |
       > 33 | use fedimint_core::query::ThresholdConsensus;
       >    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       >    |
       >    = note: `-D unused-imports` implied by `-D warnings`
```

Clippy demands a sacrifice.

justinmoon (approved):

fedimint-cli/src/lib.rs
```diff
@@ -720,11 +719,7 @@ impl FedimintCli {
                         .await
                         .map_err_cli_general()?,
                     None => ws_api
-                        .request_with_strategy(
-                            ThresholdConsensus::full_participation(ws_api.peers().len()),
-                            method,
-                            params,
-                        )
+                        .request_current_consensus(method, params)
```
elsirion, line 722: Was this actually a good idea? It's only a dev command. It didn't occur to me when I looked at this PR, but now I'm reviewing #4220 and I'm wondering if we sometimes actually want to wait for all the guardians in tests? @joschisan @dpc

dpc: I don't recall any case like that, and if they do/will need it, they can loop over each peer_id etc.

joschisan: Problem was that it was not marked as such and caused confusion when we took one peer offline. Just using loops for now if needed at all seems fine.

Issue #3481, opened 2023-10-28
Modules can't write to `0x50` database key


toyota-corolla0:
When ```PeersProposedTimestamp``` is set to ```0x50``` as shown below I get the error below. Changing it to other values fixes the issue. I am using a prefix to get all entries in ```PeersProposedTimestamp```. My project is forked from [fedimint-custom-module-example](https://github.com/fedimint/fedimint-custom-modules-example).

Some other values that I have set ```PeersProposedTimestamp``` to include ```0x51```, ```0x55```, ```0x80```, and ```0xa0```. All of these other values work fine.

### DBKeyPrefix:
```rust
/// Namespaces DB keys for this module
#[repr(u8)]
#[derive(Clone, EnumIter, Debug)]
pub enum DbKeyPrefix {
    /// ----- 00-1f reserved for struct storage -----

    /// [PredictionMarketsOutput] [OutPoint] to [PredictionMarketsOutputOutcome]
    Outcome = 0x00,

    /// [NewMarket] [OutPoint] to [Market]
    Market = 0x01,

    /// Owner's [XOnlyPublicKey] to [Order]
    Order = 0x02,

    /// ----- 20-3f reserved for market operation -----

    /// Used to produce time priority for new orders
    ///
    /// Market's [OutPoint] to [TimeOrdering]
    NextOrderTimeOrdering = 0x20,

    /// Used for payouts
    ///
    /// (Market's [OutPoint], Order's [OutPoint]) to ()
    OrdersByMarket = 0x21,

    /// Used to implement orderbook. Only holds active orders.
    ///
    /// Amount is (contract_price - price of order) for buys
    /// Amount is (price of order) for sells
    ///
    /// (Market's [OutPoint], [OutcomeSize], [Side], [Amount], [TimeOrdering]) to (Order's [XOnlyPublicKey])
    OrderPriceTimePriority = 0x22,

    /// ----- 40-4f reserved for api lookup indexes -----

    /// Indexes outcome control keys to the markets they belong to
    /// Used by client for data recovery in case of data loss
    ///
    /// ([XOnlyPublicKey], Market's [OutPoint]) to ()
    OutcomeControlMarkets = 0x40,

    /// ----- 50-5f reserved for consensus items -----
    
    /// Used for creating an agreed upon timestamp for order
    /// matching information (candlestick data)
    /// 
    /// [PeerId] to [UnixTimestamp] 
    PeersProposedTimestamp = 0x50,
}
```

### db record
```rust
[...]

PR #4192, opened 2024-01-31
chore: Migrate `DatabaseVersion` to global namespace

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/3481

In developing the state machine migrations, I have found it annoying to work with `DatabaseVersion` that is within the module's namespace (since states for state machines are not kept in the module's namespace). Given that this is an existing issue, I decided to fix this first.

- Adds database migration for `DatabaseVersion`. Since this is a migration that needs to be applied to clients and server, and is migrating keys from modules to the global namespace, I wrote some custom migration code for this. IMO this should be the only time this is necessary, since it should not be common to move keys from modules to the global namespace.
- This PR also modifies the migration tests to use isolated databases, which matches what the client and server will do in production. Because of this, all of the database backups needed to be re-generated.
- Has a simple test in the dummy client that verifies 0x50 prefix cannot be read after the migration, which should verify that the key collision issue is fixed.


modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -903,6 +916,7 @@ mod fedimint_migration_tests {
     async fn snapshot_client_db_migrations() -> anyhow::Result<()> {
         snapshot_db_migrations(
             "lightning-client-v0",
+            Some(LEGACY_HARDCODED_INSTANCE_ID_LN),
```
dpc, line 919: We should not be adding more places using `LEGACY_HARDCODED_`.

fedimint-core/src/db/mod.rs
```diff
@@ -1841,28 +1861,127 @@ pub async fn apply_migrations(
...
+    db: &Database,
+    target_db_version: DatabaseVersion,
+    module_instance_id: Option<ModuleInstanceId>,
+    kind: String,
+) -> Result<(), anyhow::Error> {
+    db.ensure_global()?;
+    let mut global_dbtx = db.begin_transaction().await;
+
+    // Use 0xff for fedimint-server and the `module_instance_id` for each module
+    let key_module_instance_id = if let Some(module_instance_id) = module_instance_id {
```
dpc, line 1911: Isn't this just `module_instance_id_or_global` ?

fedimint-core/src/db/mod.rs
```diff
@@ -1841,28 +1861,127 @@ pub async fn apply_migrations(
...
+    // `DatabaseVersionKey` If `DatabaseVersion` already exists, there is
+    // nothing to do.
+    if global_dbtx
+        .get_value(&DatabaseVersionKey(key_module_instance_id))
+        .await
+        .is_none()
+    {
+        // Read and remove the previous `DatabaseVersion`, which is in the module's
+        // isolated namespace (but not for fedimint-server)
+        let current_version_in_module = if let Some(module_instance_id) = module_instance_id {
```
dpc, line 1927: Sanity check: verify that the value has a length we expect, panic otherwise?

fedimint-testing/src/db.rs
```diff
@@ -42,6 +43,7 @@ pub fn get_project_root() -> io::Result<PathBuf> {
 /// this function will do nothing.
 pub async fn snapshot_db_migrations<F>(
     snapshot_name: &str,
+    module_instance_id: Option<ModuleInstanceId>,
```
dpc, line 46: `snapshot_db_migrations` can just hardcode an arbitrary module_instance_id, no?

fedimint-core/src/db/mod.rs
```diff
@@ -1841,28 +1861,127 @@ pub async fn apply_migrations(
...
+            remove_current_db_version_if_exists(
+                &mut global_dbtx.to_ref().into_nc(),
+                target_db_version,
+            )
+            .await
+        };
+
+        // Write the previous `DatabaseVersion` to the new `DatabaseVersionKey`
+        info!(target: LOG_DB, "Migrating DatabaseVersionKeyV0 of {kind} to DatabaseVersionKey. Version: {current_version_in_module}");
+        global_dbtx
```
dpc, line 1945: We should make sure to write db versions for each module on the first run (set to `0`, even if no migrations for them is defined), to make sure this insert happens, and this migration no longer can happen, freeing the key for the module from the start.

modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -977,11 +991,15 @@ mod fedimint_migration_tests {
                     WalletCommonInit::KIND.to_string(),
```
dpc, line 991: The `apply_migrations` 2 lines above here, should be different `apply_migrations`, one that would hardcode the module id for the purpose of testing, while the "real" `apply_migrations` will take it as an argument (from the global code).

dpc (commented):
:rocket:

fedimint-core/src/db/mod.rs
m1sterc001guy: I wrote the function then forgot to update this :)

modules/fedimint-wallet-tests/tests/tests.rs
m1sterc001guy: Are you suggesting to change hardcode the module instance id here? You're right that we don't need to use the `LEGACY` constants, the number just needs to be consistent across the snapshot test and the validate test

dpc: Module developer writing migration tests should not need to specify any module ID or even be aware there's such a thing, yes. @m1sterc001guy

fedimint-core/src/db/mod.rs
m1sterc001guy: Yes I believe this is already the case, even if no migration is defined. `migrate_database_version` only does anything interesting if there is no `DatabaseVersionKey` in the database, otherwise it is a no-op. If the old `DatabaseVersionKeyV0` also does not exist, it simply writes `target_db_version` into the database, which will prevent this migration from happening the next time the client is constructed.

m1sterc001guy: Only difference from what you said is that I think it needs to write `target_db_version` instead of `0` since that is the current version of the structs that the client is aware of.

modules/fedimint-ln-tests/tests/tests.rs
m1sterc001guy: Remove all instances of `LEGACY` from the db migration module tests

modules/fedimint-wallet-tests/tests/tests.rs
m1sterc001guy: Removed all instances of module ids from the module db migration tests

fedimint-core/src/db/mod.rs
m1sterc001guy: `DatabaseVersion` is just a `u64`. Are we doing variable length encoding of integers? Is it possible to this to ever not be 8 bytes?

dpc: I'm thinking: is it possible someone could have a key `0x50` and didn't ever do a migration, so that migration code would mistake their value for the version? Unlikely, but is it impossible?

> Are we doing variable length encoding of integers? 

Yes. Thought I don't think anyone will have > 253 or so migration versions to start getting into 3 byte encodings. :)

modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -784,54 +781,43 @@ mod fedimint_migration_tests {
...
     #[tokio::test(flavor = "multi_thread")]
     async fn snapshot_server_db_migrations() -> anyhow::Result<()> {
-        snapshot_db_migrations(
-            "wallet-server-v0",
-            |dbtx| {
-                Box::pin(async move {
-                    create_server_db_with_v0_data(dbtx).await;
-                })
-            },
-            ModuleDecoderRegistry::from_iter([(
```
dpc, line 805: Oh, no need for the other extra stuff either now? Nice!

fedimint-core/src/db/mod.rs
m1sterc001guy: Hmm yes I suppose this is possible. A module would need to write a key/value using the 0x50 prefix and then execute this migration.

Although I believe in that case the decoding of the `DatabaseVersion` should fail (unless it also coincidentally happens to be a `u64`, in which case I don't think there's much we can do) and there is already code that asserts on decode failure. So I think this length check is already implicitly covered.

dpc: > Although I believe in that case the decoding of the `DatabaseVersion` should fail (unless it also coincidentally happens to be a `u64`, in which case I don't think there's much we can do)

You're right. In that case it's good enough.

dpc (approved):

fedimint-core/src/db/mod.rs
```diff
@@ -1841,28 +1877,117 @@ pub async fn apply_migrations(
...
+    let key_module_instance_id = module_instance_id_or_global(module_instance_id);
+
+    // First check if the module has a `DatabaseVersion` written to
+    // `DatabaseVersionKey` If `DatabaseVersion` already exists, there is
+    // nothing to do.
+    if global_dbtx
+        .get_value(&DatabaseVersionKey(key_module_instance_id))
+        .await
+        .is_none()
+    {
```
elsirion, line 1932: nit: early return might be slightly more readable

fedimint-core/src/db/mod.rs
```diff
@@ -1841,28 +1877,117 @@ pub async fn apply_migrations(
...
+                target_db_version,
+            )
+            .await
+        } else {
+            remove_current_db_version_if_exists(
+                &mut global_dbtx.to_ref().into_nc(),
+                target_db_version,
+            )
+            .await
+        };
```
elsirion, line 1949: nit: The actual fn call could be factored out, only the DB argument needs to change depending on the instance id being set or not.

elsirion (approved):
Looks good, just some nits

fedimint-core/src/module/mod.rs
```diff
@@ -410,6 +417,21 @@ pub trait IDynCommonModuleInit: Debug {
 pub trait ModuleInit: Debug + Clone + Send + Sync + 'static {
     type Common: CommonModuleInit;
 
+    /// This represents the module's database version that the current code is
+    /// compatible with. It is important to increment this value whenever a
+    /// key or a value that is persisted to the database within the module
+    /// changes. It is also important to add the corresponding
+    /// migration function in `get_database_migrations` which should define how
+    /// to move from the previous database version to the current version.
+    const DATABASE_VERSION: DatabaseVersion;
```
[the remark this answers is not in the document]
m1sterc001guy: Yep!

PR #4193, opened 2024-01-31
feat: allow multiple devimint clients with the same name

maan2003:
allow you to run latency tests twice on same test dir for example

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -22,6 +22,7 @@ use fedimint_testing::federation::local_config_gen_params;
 use fedimint_wallet_client::config::WalletClientConfig;
 use fedimintd::attach_default_module_init_params;
 use fedimintd::fedimintd::FM_EXTRA_DKG_META_VAR;
+use fs_lock::FileLock;
 use futures::future::join_all;
 use rand::Rng;
 use tracing::{debug, info};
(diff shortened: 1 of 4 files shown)


dpc:
> allow you to run latency tests twice on same test dir for example

Is this fundamentally sound? Two tests using same client at the same time seems like asking for trouble.

Maan2003:
it is not same client though, second one uses a different client dir. (foo-1)

Maan2003:
it lets you do

```
latency_test(dev_fed).await;
latency_test(dev_fed).await;
```

currently it error that client dir already exists

elsirion (dismissed):
This is clever!

devimint/src/federation.rs
```diff
@@ -48,36 +48,52 @@ pub struct Client {
...
+            }
+            unreachable!()
+        })
     }
 
     /// Open or create a [`Client`] that starts with a fresh state.
     pub async fn open_or_create(name: &str) -> Result<Client> {
         let client = Self {
-            name: name.to_owned(),
+            name: format!("{name}-0"),
```
dpc, line 96: There's no need for this function now, as `create` work like it: it can't fail if the dir already exists. Also - since it doesn't go through the lock, it would be racy with `crate`.

dpc:
OK. Makese sense. :)

devimint/src/federation.rs
Maan2003: now removed

dpc (approved):

elsirion (dismissed):

justinmoon:
What's status here?

maan2003:
now CI is passing

dpc (approved):

justinmoon (approved):

PR #4181, opened 2024-01-30
[0.2] devimint latency tests with degraded federations

Maan2003:
backports #4173

PR #4198, opened 2024-02-01
Run devimint and rust tests with 1/4 guardians offline

justinmoon:
* Extended https://github.com/fedimint/fedimint/pull/4181 allowing all devimint tests to run with with 1 guardian offline. This condition is now triggered by `FM_DEVIMINT_DEGRADED` environment variable.
* Modified the `FederationTest` fixture to start with 1 guardian offline
* In the future we should parameterize devimint and `FederationTests` so we can try any valid federation size and number of online / offline guardians. But I didn't attempt this here.

### DIFF
--- a/fedimint-testing/src/federation.rs
+++ b/fedimint-testing/src/federation.rs
@@ -84,7 +84,9 @@ impl FederationTest {
             .federation_id()
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub(crate) async fn new(
+        num_online: u16,
         num_peers: u16,
         base_port: u16,
         params: ServerModuleConfigGenParamsRegistry,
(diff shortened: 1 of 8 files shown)


scripts/tests/test-ci-all.sh
```diff
@@ -5,6 +5,9 @@ set -euo pipefail
 # prevent locale settings messing with some setups
 export LANG=C
 
+# run with 1 guardian offline
+export FM_DEVIMINT_DEGRADED=1
```
justinmoon, line 9: This just sets a default to run all these tests with 1 guardians offline. I'm not sure what's best to do here.

fedimint-testing/src/federation.rs
```diff
@@ -102,6 +104,9 @@ impl FederationTest {
 
         let mut task = TaskGroup::new();
         for (peer_id, config) in configs.clone() {
+            if (u16::from(peer_id)) >= num_online {
```
douglaz, line 107:
```suggestion
            if u16::from(peer_id) >= num_online {
```

elsirion:
I think @bradleystachurski took this over in  #4247

PR #4210, opened 2024-02-01
Better logs for gateway [0.2]

Maan2003:
backports #4175 

FYI 5bf8f219c102d37711cd9496308458b107504496 didn't backport trivial

### DIFF
--- a/fedimint-core/src/util/mod.rs
+++ b/fedimint-core/src/util/mod.rs
@@ -1,6 +1,7 @@
 /// Copied from `tokio_stream` 0.1.12 to use our optional Send bounds
 pub mod broadcaststream;
 
+use std::convert::Infallible;
 use std::fmt::{Debug, Display, Formatter};
 use std::future::Future;
 use std::hash::Hash;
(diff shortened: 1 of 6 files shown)


fedimint-core/src/util/mod.rs
```diff
@@ -217,6 +218,81 @@ pub async fn write_new_async<P: AsRef<Path>, C: AsRef<[u8]>>(
         .await
 }
 
+#[derive(Debug, Clone)]
+pub struct Spanned<T> {
```
dpc, line 222: Comment?

dpc: I'm confused what is this trying to accomplish.

Maan2003: i don't want to this backport from master.

it is utility struct to enter a span whenever you want to use the wrapped object

dpc (approved):

fedimint-core/src/util/mod.rs
dpc: Missed opportunity to call it "spanner". :D

PR #4221, opened 2024-02-02
cleanup: remove unused error paths and consistently retry all errors

joschisan:


dpc (dismissed):

dpc:
We probably want to backport?

fedimint-client/src/transaction/sm.rs
```diff
@@ -54,7 +56,7 @@ pub enum TxSubmissionStates {
     ///
     /// **This state is final**
     Rejected(TransactionId, String),
-    /// this should never happen with a honest federation and bug-free code
+    /// this state is currently never assumed as all errors are retried
```
elsirion, line 59: Maybe [mark it as deprecated](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) so we get warning if it's being used?

fedimint-client/src/transaction/sm.rs
```diff
@@ -114,48 +102,47 @@ impl State for TxSubmissionStates {
...
+                        }
+                    }
+                    Err(decode_error) => {
+                        warn!(target: LOG_CLIENT_NET_API, error = %decode_error, "Failed to decode SerdeModuleEncoding")
+                    }
+                },
+                Err(error) => {
+                    error.report_if_important();
                 }
             }
```
elsirion, line 123: This is very hard to read, could we maybe use early-return and then document why in some cases we retry while in others we return?

elsirion (commented):
I think this change is overall good, but could be improved further if we are already touching the code.

fedimint-client/src/transaction/sm.rs
joschisan: done

fedimint-client/src/transaction/sm.rs
joschisan: I do not see how to factor this better tbh; I have change the variable name submission_result to serve_result and now the formatting is better at least.

elsirion: Tracking which level of error handling we are in is a bit confusing here, maybe it's just the diff view but at first I confused the inner/outer error.

dpc (approved):

elsirion (approved):

PR #4226, opened 2024-02-03
test: assert that clients join degraded federations promptly

justinmoon:
This test currently fails because joining the federation takes ~60 seconds if 1/4 guardians is offline:

```
Error: client join took too long: 60.644507708s
```

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -32,6 +32,7 @@ use super::vars::utf8;
 use crate::util::{poll, FedimintdCmd};
 use crate::{poll_eq, vars};
 
+#[derive(Clone)]
 pub struct Federation {
     // client is only for internal use, use cli commands instead
     members: BTreeMap<usize, Fedimintd>,
@@ -43,6 +44,7 @@ pub struct Federation {
 }
 
 /// `fedimint-cli` instance (basically path with client state: config + db)
+#[derive(Clone)]
 pub struct Client {
     name: String,
 }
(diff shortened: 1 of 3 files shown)


joschisan:
Do I assume correctly that joining a federation involves downloading the client config? If so consider to test removing the timeout in line 559 of fedimint-core/src/api.rs; its the only place where we use a turnout like this in a query strategy - I took it over from the previous implementation when I rewrote the download, do not really now why its there..? To me it Dows not make sense to timeout here.

bradleystachurski:
Ahh, @joschisan thanks for the clue. I think the culprit is discovering the api version set, which uses a query strategy of `AllOrDeadline`, where we give a 60 second timeout.

https://github.com/fedimint/fedimint/blob/09fe6aa/fedimint-core/src/api.rs#L699

Raises the question of how do we want to handle discovering the api version set with a degraded number of guardians?

elsirion:
I think we could just make the timeout shorter here. We do ideally want to get API versions for all guardians since if we only have the version of e.g. 3/4 as other strategies would give us even though all are online then if one actually went offline we might be down to 2/4 guardians.

If we set the timeout to e.g. 5s the worst case when joining is waiting 5s too long. Ever consecutive startup of the client should be fast (TODO: test) as the versions are updated asynchronously in the background (we can generally assume that old API versions stay around for long enough to not require an immediate switch).

bradleystachurski:
@justinmoon Thanks for flagging the issue in the PR. We can close since we're handling here: https://github.com/fedimint/fedimint/pull/4247#discussion_r1495638960

PR #4228, opened 2024-02-03
fix: justfile uses import instead of \!include

elsirion:
With the latest nixpkgs update we apparently got a new just version that errors with our justfile:

```bash
$ just mprocs
error: The `!include` directive has been stabilized as `import`
 ——▶ justfile:2:1
  │
2 │ !include justfile.fedimint.just
  │ ^
```


dpc (approved):

dpc:
Good workaround. I'll need to fix it in flakebox.

Maan2003 (approved):

PR #4229, opened 2024-02-04
chore(ln-client): improve logging when choosing gateway

### LINKED ISSUE #4194: Log should tell which gateway was chosen
This message could be improved by giving some details about the gateway that was selected:
https://github.com/fedimint/fedimint/blob/58717cb4a08ec89dc4631251a82e96d06a7b9a9d/modules/fedimint-ln-client/src/lib.rs#L855

elsirion:
Fixes #4194

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -838,6 +838,7 @@ impl LightningClientModule {
             Some(active_gateway) => Ok(active_gateway.info),
             None => {
                 let gateways = self.fetch_registered_gateways().await?;
+                let num_gateways = gateways.len();
 
                 let vetted = gateways
                     .clone()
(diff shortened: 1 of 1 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -838,26 +838,29 @@ impl LightningClientModule {
...
                 if !vetted.is_empty() {
-                    debug!("Choosing a vetted gateway");
-                    vetted
+                    let chosen_gateway = vetted
                         .into_iter()
                         .map(|gw| gw.info)
                         .choose(&mut rand::thread_rng())
-                        .ok_or(anyhow::anyhow!("Could not choose a vetted gateway"))
+                        .expect("vetted is non-empty");
+                    debug!(gateway_id = %chosen_gateway.gateway_id, %num_gateways, "Choosing a vetted gateway");
```
Maan2003, line 854: you don't use display for primitive types - https://docs.rs/tracing/latest/tracing/trait.Value.html#foreign-impls
```suggestion
                    debug!(gateway_id = %chosen_gateway.gateway_id, num_gateways, "Choosing a vetted gateway");
```

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -838,26 +838,29 @@ impl LightningClientModule {
...
                     .collect::<Vec<_>>();
                 if !vetted.is_empty() {
-                    debug!("Choosing a vetted gateway");
-                    vetted
+                    let chosen_gateway = vetted
                         .into_iter()
                         .map(|gw| gw.info)
                         .choose(&mut rand::thread_rng())
-                        .ok_or(anyhow::anyhow!("Could not choose a vetted gateway"))
+                        .expect("vetted is non-empty");
```
15IITian, line 853: But this error message  will  log  when there is panic due to  some bug or unexpected condition -> so IMO , this should not be the message.

Maan2003: feels correct to me, given we just check vetted.is_empty just above

Maan2003 (approved):

modules/fedimint-ln-client/src/lib.rs
elsirion: Thx, updated!

justinmoon (approved):

dpc (approved):

PR #4227, opened 2024-02-03
chore: remove Garbage Collection ;)

dpc:
I'm a simple developer. I see a "GC" I find the way to get rid of it.

https://www.reddit.com/r/rustjerk/comments/n8b09i/ron_swanson_on_garbage_collection/

PR #4230, opened 2024-02-04
refactor: improve client executor loop

dpc:
On top of #4227 

Make client side state machines go brrrr:

* no busy looping/polling anything (good for phone bateries)
* no canceling anything
* parallel state transitions

 No state machine was harmed during the rewrite.

### DIFF
--- a/fedimint-client/src/sm/executor.rs
+++ b/fedimint-client/src/sm/executor.rs
@@ -576,8 +576,6 @@ impl ExecutorInner {
                                                     state.operation_id(),
                                                 );
                                                 if new_state.is_terminal(context, &global_context) {
-                                                    // TODO: log state machine id or something
-                                                    debug!("State machine reached terminal state");
                                                     let k = InactiveStateKey::from_state(
                                                         new_state.clone(),
                                                     );
@@ -607,6 +605,7 @@ impl ExecutorInner {
                                 debug!(
                                     target: LOG_CLIENT_REACTOR,
                                     operation_id = %state.operation_id(),
+                                    terminal = !outcome.is_active(),
                                     ?outcome,
                                     "Finished executing state transition",
                                 );
--- a/fedimint-client/src/sm/executor.rs
+++ b/fedimint-client/src/sm/executor.rs
@@ -1,8 +1,9 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::convert::Infallible;
 use std::fmt::{Debug, Formatter};
 use std::io::{Error, Read, Write};
 use std::sync::Arc;
-use std::time::{Duration, SystemTime};
+use std::time::SystemTime;
 
 use anyhow::anyhow;
 use fedimint_core::core::{IntoDynInstance, ModuleInstanceId, OperationId};
@@ -12,14 +13,15 @@ use fedimint_core::db::{
 };
 use fedimint_core::encoding::{Decodable, DecodeError, Encodable};
 use fedimint_core::fmt_utils::AbbreviateJson;
+use fedimint_core::maybe_add_send_sync;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::task::spawn;
 use fedimint_core::util::BoxFuture;
-use fedimint_core::{maybe_add_send_sync, task};
+use fedimint_logging::LOG_CLIENT_REACTOR;
 use futures::future::{self, select_all};
-use futures::stream::StreamExt;
+use futures::stream::{FuturesUnordered, StreamExt};
 use tokio::select;
-use tokio::sync::{oneshot, Mutex};
+use tokio::sync::{mpsc, oneshot, Mutex};
 use tracing::{debug, error, info, trace, warn, Instrument};
 
 use super::state::StateTransitionFunction;
@@ -30,10 +32,6 @@ use crate::{AddStateMachinesError, AddStateMachinesResult, DynGlobalClientContex
 
 /// After how many attempts a DB transaction is aborted with an error
 const MAX_DB_ATTEMPTS: Option<usize> = Some(100);
-
-/// Wait time till checking the DB for new state machines when there are no
-/// active ones
-const EXECUTOR_POLL_INTERVAL: Duration = Duration::from_secs(1);
 
 pub type ContextGen =
     Arc<maybe_add_send_sync!(dyn Fn(ModuleInstanceId, OperationId) -> DynGlobalClientContext)>;
@@ -65,6 +63,10 @@ struct ExecutorInner {
     valid_module_ids: BTreeSet<ModuleInstanceId>,
     notifier: Notifier,
     shutdown_executor: Mutex<Option<oneshot::Sender<()>>>,
+    /// Any time executor should notice state machine update (e.g. because it
+    /// was created), it's must be sent through this channel for it to notice.
+    sm_update_tx: mpsc::UnboundedSender<DynState>,
+    sm_update_rx: Mutex<Option<mpsc::UnboundedReceiver<DynState>>>,
 }
 
 /// Builder to which module clients can be attached and used to build an
@@ -174,7 +176,11 @@ impl Executor {
             )
             .await;
             let notify_sender = self.inner.notifier.sender();
-            dbtx.on_commit(move || notify_sender.notify(state));
+            let sm_updates_tx = self.inner.sm_update_tx.clone();
+            dbtx.on_commit(move || {
+                notify_sender.notify(state.clone());
+                let _ = sm_updates_tx.send(state);
+            });
         }
 
         Ok(())
@@ -249,6 +255,13 @@ impl Executor {
             !replaced_old_context_gen,
             "start_executor was called previously"
         );
+        let sm_update_rx = self
+            .inner
+            .sm_update_rx
+            .lock()
+            .await
+            .take()
+            .expect("start_executor was called previously: no sm_update_rx available");
 
         let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel::<()>();
 
@@ -265,8 +278,8 @@ impl Executor {
         );
 
         let task_runner_inner = self.inner.clone();
-        let _handle = spawn("client state machine", async move {
-            let executor_runner = task_runner_inner.run(context_gen);
+        let _handle = spawn("state machine executor", async move {
+            let executor_runner = task_runner_inner.run(context_gen, sm_update_rx);
             select! {
                 shutdown_happened_sender = shutdown_receiver => {
                     match shutdown_happened_sender {
@@ -315,35 +328,37 @@ impl Drop for ExecutorInner {
     }
 }
 
-type TransitionForActiveState = (
-    serde_json::Value,
-    DynState,
-    StateTransitionFunction<DynState>,
-    ActiveState,
-);
+struct TransitionForActiveState {
+    outcome: serde_json::Value,
+    state: DynState,
+    meta: ActiveState,
+    transition_fn: StateTransitionFunction<DynState>,
+}
 
 impl ExecutorInner {
-    async fn run(&self, global_context_gen: ContextGen) {
+    async fn run(
+        &self,
+        global_context_gen: ContextGen,
+        sm_update_rx: tokio::sync::mpsc::UnboundedReceiver<DynState>,
+    ) {
         info!("Starting state machine executor task");
-        loop {
-            if let Err(err) = self
-                .execute_next_state_transitions(&global_context_gen)
-                .await
-            {
-                warn!(
-                    %err,
-                    "An unexpected error occurred during a state transition"
-                );
-            }
+        if let Err(err) = self
+            .run_state_machines_executor_inner(global_context_gen, sm_update_rx)
+            .await
+        {
+            warn!(
+                %err,
+                "An unexpected error occurred during a state transition"
+            );
         }
     }
 
     async fn get_transition_for(
         &self,
-        state: DynState,
+        state: &DynState,
         meta: ActiveState,
         global_context_gen: &ContextGen,
-    ) -> Option<BoxFuture<TransitionForActiveState>> {
+    ) -> Vec<BoxFuture<'static, TransitionForActiveState>> {
         let module_instance = state.module_instance_id();
         let context = &self
             .module_contexts
@@ -362,7 +377,12 @@ impl ExecutorInner {
                         trigger,
                         transition,
                     } = transition;
-                    (trigger.await, state, transition, meta)
+                    TransitionForActiveState {
+                        outcome: trigger.await,
+                        state,
+                        transition_fn: transition,
+                        meta,
+                    }
                 });
                 f
             })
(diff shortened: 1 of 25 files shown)


dpc:
I think it increases flakiness of one/two tests.

fedimint-client/src/sm/executor.rs
```diff
@@ -64,6 +61,9 @@ struct ExecutorInner {
     module_contexts: BTreeMap<ModuleInstanceId, DynContext>,
     valid_module_ids: BTreeSet<ModuleInstanceId>,
     notifier: Notifier,
+    /// Any time executor should notice state machine update (e.g. because it
+    /// was created), it's [`OperationId`] is sent through this channel.
```
dpc, line 65: Leftover from previous version

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,250 @@ impl ExecutorInner {
...
-                .get_transition_for(state, meta, global_context_gen)
-                .await
-            {
-                None => {}
-                Some(t) => transitions.push(t),
-            }
+        /// All futures in the executor resolve to this type, so the handling
+        /// code can tell them apart.
+        enum ExecutorLoopEvent {
+            /// New active `DynState` showed up in the db and needs to be
```
dpc, line 448: Fix: At this point we actually don't know if it's true, and only received a notification.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,250 @@ impl ExecutorInner {
...
+            /// progressed
+            New {
+                state: DynState,
+                sm_update_rx: mpsc::UnboundedReceiver<DynState>,
+            },
+            /// One of trigger function of a state machine finished and returned
+            /// transition function to run
+            Triggered(TransitionForActiveState),
+            /// Transition function and all the accounting around are done
```
dpc, line 457: it

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,250 @@ impl ExecutorInner {
...
+                        sm_update_rx,
+                    }
+                } else {
+                    ExecutorLoopEvent::Disconnected
+                }
+            }));
+        }
+
         let active_states = self.get_active_states().await;
-        // TODO: use DB prefix subscription instead of polling
```
dpc, line 415: Note: world doesn't end on db subscription, and when possible it's more efficient to use in-memory channels.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,250 @@ impl ExecutorInner {
...
+                                        last_error,
+                                        attempts,
+                                    } => last_error.context(format!(
+                                        "Failed to commit after {attempts} attempts"
+                                    )),
+                                    AutocommitError::ClosureError { error, .. } => error,
+                                }).expect("autocommit of transition fn should never fail");
+
+                            // TODO: add a INFO version without all the details
+                            debug!(
```
dpc, line 617: Dedup with another logging I added in `Complete` handler.

fedimint-client/src/sm/executor.rs
```diff
@@ -701,6 +807,7 @@ impl ExecutorBuilder {
 pub struct ActiveStateKey {
     // TODO: remove redundant operation id from state trait
     pub operation_id: OperationId,
+    // TODO: state being a key... seems ... risky?
```
dpc, line 810: TODO: delete . I was surprised that we can't key `DynState` on something ... shorter. It forces me to pass `DynState` around a lot ... which isn't a big deal, I guess.

fedimint-client/src/sm/executor.rs
```diff
@@ -932,7 +1039,9 @@ impl ::fedimint_core::db::DatabaseLookup for InactiveStateKeyPrefix {
 enum ActiveOrInactiveState {
     Active {
         dyn_state: DynState,
-        active_state: ActiveState,
+        // TODO: remove?
```
dpc, line 1042: I think it was used to log something, and that's about it. Remove?

fedimint-client/src/sm/executor.rs
dpc: @elsirion While intuitive I'd think that every module's sm would be unique ... if someone comes up with a use-case where it isn't ... it will be confusing for them what is happening. :D

elsirion: I agree that it's not optimal, longer term I'd like to key SMs the following way: `<operation_id> <sm_idx> <step_idx>`. I also want to get rid of the duplication between active and inactive state keys and instead keep a list of `<operation_id> <sm_idx>` prefixes for still-running SMs.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,258 @@ impl ExecutorInner {
...
+
+        // main reactor loop: wait for next thing that completed, react (possibly adding
+        // more things to `futures`)
+        while let Some(event) = futures.next().await {
+            match event {
+                ExecutorLoopEvent::New {
+                    state,
+                    sm_update_rx,
+                } => {
+                    re_add_sm_update_receiver(&mut futures, sm_update_rx);
```
elsirion, line 483: Instead of running the `sm_update_rx.recv()` future inside `futures`, why not do a `select!` over both `futures.next()` and `sm_update_rx.recv()`? That way you wouldn't need to awkwardly re-add that receive future all the time.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,258 @@ impl ExecutorInner {
...
+
+                    if currently_running_sms.contains(&state) {
+                        warn!(target: LOG_CLIENT_REACTOR, operation_id = %state.operation_id(), "Received a state machine that is already running. Ignoring");
+                        continue;
                     }
+                    let Some(meta) = self.get_active_state(&state).await else {
+                        warn!(target: LOG_CLIENT_REACTOR, operation_id = %state.operation_id(), "Couldn't look up received state machine. Ignoring.");
+                        continue;
+                    };
```
elsirion, line 492: This is to enforce that the SM gets written to the DB first?

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,258 @@ impl ExecutorInner {
...
+
+                    let transitions = self
+                        .get_transition_for(&state, meta, &global_context_gen)
+                        .await;
+                    if transitions.is_empty() {
+                        warn!(target: LOG_CLIENT_REACTOR, operation_id = %state.operation_id(), "Received an active state that doesn't produce any transitions. Ignoring.");
```
elsirion, line 498: Shouldn't the active state be removed and made an inactive one? Otherwise this would happen on every restart.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,258 @@ impl ExecutorInner {
...
-                                        active_state: v,
-                                    })
+                                .await
+                                .map_err(|e| match e {
+                                    AutocommitError::CommitFailed {
+                                        last_error,
+                                        attempts,
+                                    } => last_error.context(format!(
+                                        "Failed to commit after {attempts} attempts"
+                                    )),
```
elsirion, line 615: You use `None` as the number of retries, that means it will run forever, so this could be an `unreachable!`?

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,258 @@ impl ExecutorInner {
...
+                }) => {
+                    let span = tracing::info_span!(
+                        "state_machine_transition",
+                        operation_id = %state.operation_id()
+                    );
+                    // Perform the transition as another future, so transitions can happen in
+                    // parallel.
+                    // Database write conflicts might be happening quite often here,
+                    // but transaction functions are supposed to be idempotent anyway,
+                    // so it seems like a good stress-test in the worst case.
```
elsirion, line 525: Depending on how heavy operations are run here that could be quite inefficient, imagine multiple SMs trying to create an e-cash output for the same index, all hashing to the BLS curve, which takes a few ms. Let's see how it goes in practice.

fedimint-client/src/sm/executor.rs
elsirion: Why it is a key: originally I assumed the DB had prefix notifications and I wanted to build the notification system around the serialized SMs in the DB.

fedimint-client/src/sm/executor.rs
```diff
@@ -1076,11 +1202,14 @@ mod tests {
...
-        let executor = executor_builder
+        let (executor, sm_update_rx) = executor_builder
             .build(db.clone(), Notifier::new(db.clone()))
             .await;
         executor
-            .start_executor(Arc::new(|_, _| DynGlobalClientContext::new_fake()))
+            .start_executor(
+                Arc::new(|_, _| DynGlobalClientContext::new_fake()),
+                sm_update_rx,
+            )
```
elsirion, line 1212: If we just return `sm_update_rx` here to give it back, why not put that `Mutex<Option<…>>` into `Executor` to avoid this?

elsirion (commented):
Some comments, not awake enough to do a proper review

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +405,258 @@ impl ExecutorInner {
...
+            /// returned transition function to run
+            Triggered(TransitionForActiveState),
+            /// Transition function and all the accounting around it are done
+            Completed {
+                state: DynState,
+                outcome: ActiveOrInactiveState,
+            },
+            /// New job receiver disconnected, that can only mean termination
+            Disconnected,
         }
```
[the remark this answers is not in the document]
dpc: It's only used internally, isn't useful outside.

fedimint-client/src/sm/executor.rs
dpc: Yes. Initially I thought that I'll be getting some "id of a sm" and thus I do need to actually read the thing from the db. Only then I noticed that "sm itself is the key", but I still kept the original flow - the sm needs to be in a db already. Enforcing correctness on the sender is kind of a side effect.

fedimint-client/src/sm/executor.rs
dpc: True. I wanted to keep everything going through these `FuturesUnurdered` that I missed that I can still compose it.

fedimint-client/src/sm/executor.rs
dpc: I think so. BTW. I see `Some(100)` etc. all over the code, and I wonder if we have preferences.

fedimint-client/src/sm/executor.rs
dpc: Felt like unnecessary mutability. :D

But I guess it can be done.

fedimint-client/src/sm/executor.rs
dpc: It still is, in `get_transition_for` (I kept it there, and only removed composing the selected here, as it seems to belong to this layer of abstraction (polling things)).

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
-                num_transitions, "Awaiting any state transition to become ready"
-            );
-            let new_state_added = async move {
-                loop {
-                    // Prioritize existing active states over new states
-                    fedimint_core::task::sleep(EXECUTOR_POLL_INTERVAL).await;
-                    let new_active_states_count = self.get_active_states().await.len();
-                    if new_active_states_count > active_state_count {
-                        return;
+            let event = tokio::select! {
```
dpc, line 457: I thought that my dislike for `select` was irrational, but then I  couldn't `return`,`break` and `continue` from the select's arm, and I felt a little better about myself. Also no formatting.

But I guess the way here is OK, short side-effect free select, then normal code.

fedimint-client/src/sm/executor.rs
elsirion: Eventually we should only be passing the id, so this is a good design.

fedimint-client/src/sm/executor.rs
elsirion: `None` is more correct, but potentially harder to notice issues with. I often chose `Some(100)` as a canary that will tell us if something is being retried excessively.

fedimint-client/src/sm/executor.rs
elsirion: Didn't you do exactly this on the client level instead? Will give the PR a complete review again, then I'll see if it was about the same oject.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
-        if active_states.is_empty() {
-            // FIXME: what to do in this case? Probably best to subscribe to DB eventually
-            trace!("No state transitions available, waiting before re-trying");
-            task::sleep(EXECUTOR_POLL_INTERVAL).await;
-            return Ok(());
+        trace!(target: LOG_CLIENT_REACTOR, "Starting active states: {:?}", active_states);
+        for (state, _meta) in active_states {
+            self.sm_update_tx
+                .send(state)
+                .expect("Must be able to send operation_id to own opened channel");
```
elsirion, line 428: nit: not only the operation ID is being sent, slight inaccuracy

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
-                .get_transition_for(state, meta, global_context_gen)
-                .await
-            {
-                None => {}
-                Some(t) => transitions.push(t),
-            }
+        /// All futures in the executor resolve to this type, so the handling
+        /// code can tell them apart.
+        enum ExecutorLoopEvent {
+            /// Notivication about `DynState` arrived and should be handled,
```
elsirion, line 434: `s/Notivication/Notification/`

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
+            /// usually added to the list of pending futures.
+            New { state: DynState },
+            /// One of trigger functions of a state machine finished and
```
elsirion, line 437: nit: `s/trigger functions/trigger futures/`

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
+                    // Perform the transition as another future, so transitions can happen in
+                    // parallel.
+                    // Database write conflicts might be happening quite often here,
+                    // but transaction functions are supposed to be idempotent anyway,
+                    // so it seems like a good stress-test in the worst case.
+                    futures.push({
+                        let sm_update_tx = self.sm_update_tx.clone();
+                        let db = self.db.clone();
+                        let notifier = self.notifier.clone();
+                        let module_contexts = self
```
elsirion, line 520: The code creating this future is quite long and nested, did you think about extracting it into its own fn? Main challenge seems making it return a `'static` future while taking `&'a self`.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
+                                                Ok(ActiveOrInactiveState::Inactive {
+                                                    dyn_state: new_state,
+                                                })
+                                            } else {
+                                                let k =
+                                                    ActiveStateKey::from_state(new_state.clone());
+                                                let v = ActiveState::new();
+                                                dbtx.insert_entry(&k, &v).await;
+                                                dbtx.on_commit(move || {
+                                                    let _ = sm_update_tx.send(state);
```
elsirion, line 587: Are we sending this message twice now? First here …

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
-                                        dyn_state: new_state,
-                                        active_state: v,
-                                    })
+
+                            match &outcome {
+                                ActiveOrInactiveState::Active {
+                                    dyn_state,
+                                    meta: _,
+                                } => {
+                                    sm_update_tx.send(dyn_state.clone()).expect("can't fail: we are the receiving end");
```
elsirion, line 607: … second time here?

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
+                                    notifier.notify(dyn_state.clone());
```
elsirion, line 608: It's a bit annoying that we now have two notification systems: one for active states only and one for all state transitions. Ideally these could be combined eventually imo.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
+                    info!(target: LOG_CLIENT_REACTOR, "Disconnected. Terminating.");
+                    break;
                 }
-                anyhow::Ok(())
             }
-            .instrument(span)
-            .await?
         }
+
+        trace!(target: LOG_CLIENT_REACTOR, "Terminated");
```
elsirion, line 640: What's the reason for having both these logs at these two levels?

fedimint-client/src/sm/executor.rs
elsirion: Having some timing info is good, it's currently our only way of ordering the state machine log in the DB :see_no_evil: I know that's not bullet-proof.

modules/fedimint-mint-client/src/output.rs
```diff
@@ -333,6 +334,13 @@ pub struct NoteIssuanceRequest {
     blinding_key: BlindingKey,
 }
 
+impl hash::Hash for NoteIssuanceRequest {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.spend_key.hash(state);
+        // ignore `blinding_key` as it doesn't impl Hash; `spend_key` has enough
+        // entropy anyway
+    }
+}
```
elsirion, line 343: Wondering why clippy isn't complaining that `Hash` is implemented manually while `PartialEq` is being derived. I think technically what you are doing is correct, could just lead to more collisions if someone was to use this data structure in a nonsensical way (same spend key, different blinding keys).

elsirion (commented):
Great work!

fedimint-client/src/sm/executor.rs
dpc: Not a big issue, and not clear if it merging them will be better/possible. If there are N notifications needed then there we send it N times. At least the executor channel is mpsc.

fedimint-client/src/sm/executor.rs
dpc: Sloppiness.

elsirion (approved):

fedimint-client/src/sm/executor.rs
elsirion: The diff view (or any other width-limited view) would be easier to read :P

fedimint-client/src/sm/executor.rs
dpc: It also has a semantic difference. If there are two things that require notifications and sender is responsible for sending notifications, then two sends is the right thing to do.

I'm not a fan of the whole db subscription thing in particular because by definition it couples notifications with database implementation details. Receiver needs to know exact keys (and thus data model) of the sender to communicate. Refactoring of the notifying party can introduce bugs all the same as forgetting to send, especially that we often leave the old keys/variants around for backward compat reasons. The communication channel is implicit: for the reader of the "sender" code it's much harder to figure out the answer to "well, now what, what's going to happen?".

With channels the communication is explicit. "To make the executor notice new state machine, you need to send it a notification, here is that `send`". In practice it's hard to "forget" about it, because it shouldn't be sprinkled all over the code, but in one(few) function doing it, so the error-proneness should not be any worse.

> and required introducing on_commit to even work, which I did in the end

DB notifications are just a special cased `dbtx.on_commit(|| implicit_channel.send(MAGIC_KEY));` with extra steps . :shrug:  *insert change my mind meme here*

dpc:
@elsirion I'm afraid until https://github.com/fedimint/fedimint/pull/4250 is figured out, this is making the CI too flaky to land.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,230 @@ impl ExecutorInner {
...
+                    // parallel.
+                    // Database write conflicts might be happening quite often here,
+                    // but transaction functions are supposed to be idempotent anyway,
+                    // so it seems like a good stress-test in the worst case.
+                    futures.push({
+                        let sm_update_tx = self.sm_update_tx.clone();
+                        let db = self.db.clone();
+                        let notifier = self.notifier.clone();
+                        let module_contexts = self
+                                                .module_contexts.clone();
```
douglaz, line 521: remove this new line here

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,230 @@ impl ExecutorInner {
...
+                        let global_context_gen = global_context_gen.clone();
+                        Box::pin(
+                        async move {
+                            debug!(target: LOG_CLIENT_REACTOR, operation_id = %state.operation_id() , "Executing state transition");
+                            debug!(
+                                ?state,
+                                outcome = ?AbbreviateJson(&outcome),
```
douglaz, line 528: no message here?

fedimint-client/src/sm/executor.rs
dpc: Why is rustfmt not formatting it ??

dpc: WTH. Why do I have to do code formatting manually like a caveman? :D

fedimint-client/src/sm/executor.rs
dpc: Copied it from previous code.

dpc: Merged both debug lines now.

maan2003: i believe first one used to `info` when split it

fedimint-client/src/sm/executor.rs
douglaz: > Why is rustfmt not formatting it ??

rustfmt sometimes won't remove some excessive new lines...

dpc: Nothing in here (much larger block) gets formatted for some reason, AFAICT.

fedimint-client/src/sm/executor.rs
dpc: Splited it again into shorter info + detailed debug.

douglaz:
> Make client side state machines go brrrr:
> 
> * no busy looping/polling anything (good for phone bateries)

> * parallel state transitions

The parallel state transitions may lead to duplicated processing. So it may or may not improve phone batteries.

Perhaps we should keep them serial for now? I'm not sure we understand all the possible side effects and new issues that may happen.

dpc:
> The parallel state transitions may lead to duplicated processing. So it may or may not improve phone batteries.

Not polling on empty executor is the main benefit. The second one is not canceling all pending futures. They key to saving battery is not waking up CPUs unnecessarily, so they can sit in low power mode for as long as possible.

> Perhaps we should keep them serial for now? I'm not sure we understand all the possible side effects and new issues that may happen.


Processing state change in parallel or not is a minor thing, IMO, one or the other. My bet is conflicts here will not be happening a lot, and even if they do, coupe of retries with a backoff and everything should complete anyway in a burst of activity, after which the CPUs can go back to sleep.

I don't want to pesimise right away. If it looks like db conflicts are common, I'll start by introducing a map of `operation_id -> mutex` or `module_id -> mutex` to serialize things touching same parts.

And I think And some conflicts from time to time will be beneficial as they will exercise idempotency of state transition functions.

douglaz:
@dpc the question is when we are including this in a stable release. If we want to include it in the foreseen future, we should be more conservative, incremental and pessimistic about the changes.

dpc:
> @dpc the question is when we are including this in a stable release. If we want to include it in the foreseen future, we should be more conservative, incremental and pessimistic about the changes.

Not planing to backport it. If we do backport it, I guess I can wrap in a mutex there.

maan2003:
oh formatting is totally messed up, rustfmt is giving up

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
+                        target: LOG_CLIENT_REACTOR,
+                        operation_id = %state.operation_id(),
+                        outcome_active = outcome.is_active(),
+                        total = futures.len(),
+                        "State transition complete");
+                    trace!(
+                        target: LOG_CLIENT_REACTOR,
+                        ?outcome,
+                        operation_id = %state.operation_id(), total = futures.len(),
+                        "State transition complete");
```
maan2003, line 632: lisp vibes, closing bracket on same line

dpc:
> oh formatting is totally messed up, rustfmt is giving up

The arm of that `match` for `ExecutorLoopEvent::Triggered` is the only part affected, it seems. Beats me why, and it's hard to tweak it too much without breaking it.

I don't think I care too much, unless you see some particular eyesores.

dpc:
BTW. We do have a big red `warn!` on autocommit fail, so we'll see quickly if parallel state machine transfer functions are conflicting.

fedimint-client/src/sm/executor.rs
```diff
@@ -390,181 +410,235 @@ impl ExecutorInner {
...
+                                                    meta: v,
+                                                })
+                                            }
+                                        })
+                                    },
+                                    None,
                                 )
-                                .await;
+                                .await
+                                .expect("autocommit should keep trying to commit (max_attempt: None) and body doesn't return errors");
```
maan2003, line 600: running this expect fixes rustfmt :shrug: 

I pushed to 6e7946ac4a, feel free to use that

dpc: "running"? I don't understand. But thanks a lot!

maan2003: edited, I can't even blame auto correct on laptop :laughing:

dpc:
Amended with Manmeet's fix. I still don't understand what he did there. :D

fedimint-client/Cargo.toml
```diff
@@ -31,6 +31,7 @@ fedimint-derive-secret = { version = "0.3.0-alpha", path = "../crypto/derive-sec
 fedimint-aead = { version = "0.3.0-alpha", path = "../crypto/aead" }
 fedimint-logging = { version = "0.3.0-alpha", path = "../fedimint-logging" }
 futures = "0.3.26"
+futures-util = "0.3.30"
```
maan2003, line 34: is this used anywhere, futures crate reexports all of futures-util

dpc: Good catch.

maan2003:
I created #4291 to make sure this doesn't break wasm

elsirion (approved):

dpc:
Now that flakiness is fixed and random failures dissabled, I rebased and this should be good to go.

douglaz:
Why backwards-compatibility tests are failing?

bradleystachurski:
Looks like a few of the rust unit tests are timing out (flake?).

```
TIMEOUT [ 480.039s] fedimint-ln-gateway::gatewayd-integration-tests
TIMEOUT [ 480.092s] fedimint-wallet-tests::fedimint_wallet_tests peg_out_fail_refund
```

I think we can disable `rust_unit_tests` in backwards-compatibility tests since it's not using binaries from previous versions.

dpc:
Something smell wrong about this change. Will be debugging.

dpc:
Hopefully #4329 will finally make it green.

douglaz (approved):

dpc:
Seems like this change improves latency tests 2 times.

```
before 1st try:
REISSUE: min: 1.5s, avg: 1.6s, median: 1.6s, p90: 1.6s, max: 1.6s, sum: 47.0s
LN SEND: min: 2.9s, avg: 3.0s, median: 2.9s, p90: 3.0s, max: 3.8s, sum: 59.0s
LN RECV: min: 1.4s, avg: 1.8s, median: 1.5s, p90: 2.4s, max: 2.4s, sum: 35.5s
FM PAY: min: 3.5s, avg: 3.6s, median: 3.6s, p90: 3.6s, max: 3.6s, sum: 71.3s


before, 2nd try:
REISSUE: min: 1.5s, avg: 1.6s, median: 1.6s, p90: 1.6s, max: 1.6s, sum: 47.0s
LN SEND: min: 2.8s, avg: 2.9s, median: 2.9s, p90: 2.9s, max: 3.7s, sum: 58.7s
LN RECV: min: 1.4s, avg: 1.8s, median: 1.5s, p90: 2.4s, max: 2.4s, sum: 35.0s
FM PAY: min: 3.5s, avg: 3.8s, median: 3.6s, p90: 3.6s, max: 8.6s, sum: 76.3s


after, 1st try:
REISSUE: min: 0.5s, avg: 0.6s, median: 0.6s, p90: 0.6s, max: 0.6s, sum: 16.6s
LN SEND: min: 1.3s, avg: 1.5s, median: 1.5s, p90: 1.7s, max: 2.0s, sum: 30.0s
LN RECV: min: 1.3s, avg: 1.6s, median: 1.6s, p90: 1.8s, max: 1.9s, sum: 32.4s
FM PAY: min: 1.5s, avg: 1.6s, median: 1.6s, p90: 1.6s, max: 1.8s, sum: 32.1s


after, 2nd try:
REISSUE: min: 0.5s, avg: 0.6s, median: 0.5s, p90: 0.6s, max: 0.6s, sum: 16.6s
LN SEND: min: 1.3s, avg: 1.5s, median: 1.4s, p90: 1.7s, max: 2.1s, sum: 29.9s
LN RECV: min: 1.3s, avg: 1.6s, median: 1.6s, p90: 1.6s, max: 1.6s, sum: 31.9s
FM PAY: min: 1.5s, avg: 1.6s, median: 1.6s, p90: 1.6s, max: 1.7s, sum: 32.2s
```

justinmoon (approved):

PR #4231, opened 2024-02-04
feat(mint-client): include federation join info in `OOBNotes`

### LINKED ISSUE #4048: E-cash should optionally contain invite code
That way receiving e-cash from unknown federations becomes possible. While this isn't recommended there seems to be strong user demand.

elsirion:
Fixes #4048

### DIFF
--- a/fedimint-client/src/module/mod.rs
+++ b/fedimint-client/src/module/mod.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use std::{ffi, marker, ops};
 
 use anyhow::bail;
-use fedimint_core::api::DynGlobalApi;
+use fedimint_core::api::{DynGlobalApi, InviteCode};
 use fedimint_core::config::ClientConfig;
 use fedimint_core::core::{
     Decoder, DynInput, DynOutput, IntoDynInstance, ModuleInstanceId, ModuleKind, OperationId,
@@ -432,6 +432,22 @@ where
 
     pub fn get_config(&self) -> ClientConfig {
         self.client.get().get_config().clone()
+    }
+
+    /// Returns an invite code for the federation that points to an arbitrary
+    /// guardian server for fetching the config
+    pub fn get_invite_code(&self) -> InviteCode {
+        let cfg = self.get_config().global;
+        let (any_guardian_id, any_guardian_url) = cfg
+            .api_endpoints
+            .iter()
+            .next()
+            .expect("A federation always has at least one guardian");
+        InviteCode::new(
+            any_guardian_url.url.clone(),
+            *any_guardian_id,
+            cfg.federation_id(),
+        )
     }
 
     pub fn get_internal_payment_markers(&self) -> anyhow::Result<(PublicKey, u64)> {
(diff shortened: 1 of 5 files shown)


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -110,11 +119,25 @@ impl OOBNotes {
         ])
     }
 
+    pub fn new_with_invite(notes: TieredMulti<SpendableNote>, invite: InviteCode) -> Self {
+        Self(vec![
+            // FIXME: once we can break compatibility with 0.2 we can remove the prefix in case an
```
Maan2003, line 124: maybe `FIXME(0.2)` so we can grep for it later

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -92,6 +93,14 @@ pub struct OOBNotes(Vec<OOBNotesData>);
 enum OOBNotesData {
```
Maan2003, line 93: OOBNotesData was a good idea!

fedimint-client/src/module/mod.rs
```diff
@@ -434,6 +434,22 @@ where
...
+        let (any_guardian_id, any_guardian_url) = cfg
+            .api_endpoints
+            .iter()
+            .next()
+            .expect("A federation always has at least one guardian");
+        InviteCode::new(
+            any_guardian_url.url.clone(),
+            *any_guardian_id,
+            cfg.federation_id(),
+        )
```
Maan2003, line 450: would be nice to also have this function on config type

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -92,6 +93,14 @@ pub struct OOBNotes(Vec<OOBNotesData>);
...
     Notes(TieredMulti<SpendableNote>),
     FederationIdPrefix(FederationIdPrefix),
+    /// Invite code to join the federation by which the e-cash was issued
+    ///
+    /// Introduce in 0.3.0
+    Invite {
```
dpc, line 99: If this is present the `FederationIdPrefix` is redundant, though we might start to omit it in some future. Though we might already change code that uses `FederationIdPrefix` to use see if `Invite` is available.

elsirion: Is this a documentation request?

dpc: No. Code request, unless I'm missing something.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -110,11 +119,25 @@ impl OOBNotes {
...
+            },
+        ])
+    }
+
     pub fn federation_id_prefix(&self) -> FederationIdPrefix {
         self.0
             .iter()
             .find_map(|data| match data {
                 OOBNotesData::FederationIdPrefix(prefix) => Some(*prefix),
+                OOBNotesData::Invite { federation_id, .. } => Some(federation_id.to_prefix()),
```
elsirion, line 140: @dpc feels like this is what you were asking for in your comment?

modules/fedimint-mint-client/src/lib.rs
elsirion: https://github.com/fedimint/fedimint/pull/4231#discussion_r1477646687

modules/fedimint-mint-client/src/lib.rs
dpc: :+1:

dpc (approved):

Maan2003 (approved):

PR #4237, opened 2024-02-05
chore: improve devimint api

Maan2003:


devimint/src/federation.rs
```diff
@@ -32,17 +32,19 @@ use super::vars::utf8;
 use crate::util::{poll, FedimintdCmd};
 use crate::{poll_eq, vars};
 
+#[derive(Clone)]
 pub struct Federation {
```
dpc, line 36: Does it make sense to clone this? Isn't it more like a ... resource?

Maan2003: all daemon structs are clone in devimint

dpc: We don't do things like "kill on `Drop`" etc. on them?

dpc (approved):

devimint/src/federation.rs
Maan2003: that happens in ProcessHandle, which is Arc<ProcessHandleInner>

elsirion (approved):

PR #4240, opened 2024-02-05
chore: make debug statement more informative

### LINKED ISSUE #4196: Log output outcome failure
Without telling more about the error and what the client was trying to do this debug log is useless:

https://github.com/fedimint/fedimint/blob/58717cb4a08ec89dc4631251a82e96d06a7b9a9d/modules/fedimint-ln-client/src/pay.rs#L174

15IITian:
*Tells which error is causing failure.
*Gives information of the Contract by giving its `TransactionId` and `ContractId`

fix #4196

### DIFF
--- a/modules/fedimint-ln-client/src/pay.rs
+++ b/modules/fedimint-ln-client/src/pay.rs
@@ -169,8 +169,12 @@ impl LightningPayCreatedOutgoingLnContract {
                 }
                 Err(e) => {
                     e.report_if_important();
+
                     debug!(
-                        "Awaiting output outcome failed, retrying in {}s",
+                        error = e.to_string(),
+                        transaction_id = txid.to_string(),
+                        contract_id = contract_id.to_string(),
+                        "Retrying in {}s",
                         RETRY_DELAY.as_secs_f64()
                     );
                     sleep(RETRY_DELAY).await;
--- a/modules/fedimint-wallet-client/src/withdraw.rs
+++ b/modules/fedimint-wallet-client/src/withdraw.rs
@@ -92,7 +92,9 @@ async fn await_withdraw_processed(
 
                 e.report_if_important();
                 debug!(
-                    "Awaiting output outcome failed, retrying in {}s",
+                    error = e.to_string(),
+                    operation_id = operation_id.to_string(),
+                    "Retrying in {}s",
                     RETRY_DELAY.as_secs_f64()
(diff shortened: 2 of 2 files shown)


modules/fedimint-ln-client/src/pay.rs
```diff
@@ -170,8 +170,13 @@ impl LightningPayCreatedOutgoingLnContract {
                 }
                 Err(e) => {
                     e.report_if_important();
+
                     debug!(
-                        "Awaiting output outcome failed, retrying in {}s",
```
dpc, line 174: We should prefer structured logging https://docs.rs/tracing/latest/tracing/#recording-fields over string interpolation.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -169,8 +169,12 @@ impl LightningPayCreatedOutgoingLnContract {
...
+                        Error = e.to_string(),
```
elsirion, line 174: Let's not capitalize the key here, we don't do this anywhere else.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -169,8 +169,12 @@ impl LightningPayCreatedOutgoingLnContract {
...
+                        transactionId = txid.to_string(),
+                        contractId = contract_id.to_string(),
```
elsirion, line 176: snake_case please

elsirion (approved):
Consistency is key in these simple changes

15IITian:
> Consistency is key in these simple changes

I'll remember it.

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -169,8 +169,12 @@ impl LightningPayCreatedOutgoingLnContract {
                 }
                 Err(e) => {
                     e.report_if_important();
+
                     debug!(
-                        "Awaiting output outcome failed, retrying in {}s",
+                        error = e.to_string(),
+                        transaction_id = txid.to_string(),
+                        contract_id = contract_id.to_string(),
```
elsirion, line 176: nit: instead of `transaction_id = txid.to_string()` you can write `transaction_id = %txid`

elsirion (approved):
Good enough

dpc (approved):

Issue #4243, opened 2024-02-05
Build and push mutinynet-bitcoind CI job fails


bradleystachurski:
```
buildx failed with: ERROR: failed to solve: failed to read dockerfile: open docker/mutinynet-bitcoind-docker/Dockerfile: no such file or directory
```
https://github.com/fedimint/fedimint/actions/runs/7783254396/job/21221336034#step:6:207

PR #4253, opened 2024-02-06
fix: update docker paths for CI build and push jobs

bradleystachurski:
Fixes https://github.com/fedimint/fedimint/issues/4243

### DIFF
--- a/.github/workflows/extra-docker-images.yml
+++ b/.github/workflows/extra-docker-images.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Build and push mutinynet-bitcoind
         uses: docker/build-push-action@v5
         with:
-          file: docker/mutinynet-bitcoind-docker/Dockerfile
+          file: docker/0.2.1/mutinynet-bitcoind-docker/Dockerfile
           push: true
           tags: ${{ steps.meta-mutinynet-bitcoind.outputs.tags }}
           labels: ${{ steps.meta-mutinynet-bitcoind.outputs.labels }}
@@ -55,7 +55,7 @@ jobs:
       - name: Build and push xmpp-prosody-chat-server
         uses: docker/build-push-action@v5
         with:
-          context: "{{defaultContext}}:docker/full-tls-mutinynet/prosody"
+          context: "{{defaultContext}}:docker/0.2.1/full-tls-mutinynet/prosody"
           push: true
           tags: ${{ steps.meta-xmpp-prosody.outputs.tags }}
           labels: ${{ steps.meta-xmpp-prosody.outputs.labels }}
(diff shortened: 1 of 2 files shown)


.github/workflows/extra-docker-images.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Build and push xmpp-prosody-chat-server
         uses: docker/build-push-action@v5
         with:
-          context: "{{defaultContext}}:docker/full-tls-mutinynet/prosody"
+          context: "{{defaultContext}}:docker/0.2.1/full-tls-mutinynet/prosody"
```
dpc, line 58: Should we ... get it from a current tag or something?

bradleystachurski: We had a `latest` dir, but that was removed in https://github.com/fedimint/fedimint/commit/0f3a65268c06e9a592d87a6c2e8c145a53139735

@Kodylow (tagging since you have context)

Do you have a pattern you'd like to use so we can reference Dockerfiles similar to a current tag?

Kodylow: I'm happy doing latest if those are stable, I tagged to 0.2.1 as the stable release. If we do latest will it pull 0.2.2 or will it be pulling master updates?

bradleystachurski: Since we're statically defining the path in the repo, we shouldn't be pulling (please correct me if I'm misinterpreting).

For context, this is the build and push job in CI, which is failing since the previous directory structure was refactored. We just need to point to the correct path.

```
failed to read dockerfile: open docker/mutinynet-bitcoind-docker/Dockerfile: no such file or directory
```

The job fails at the previous step for bitcoind, but we'll also fail for the prosody path.

dpc: What I meant was: ... do only on tag builds and use some `${{ something.tag-name }}` kind of thing to get the version automaticaly?

bradleystachurski: I believe we're already getting the version automatically from the `docker/metadata-action@v5` steps.

Just to make sure I understand, are you suggesting we remove `branches` from the workflow and only run the job on tag events?

https://github.com/fedimint/fedimint/blob/master/.github/workflows/extra-docker-images.yml#L5-L7

Also, we still need to give the job a path to the Dockeriles/contexts. For now, the common place appears to be `fedimint/docker/0.2.1/...`. I'm curious if there's a pattern where the common artifacts (i.e. bitcoind and prosody) should live that isn't 0.2.1.

elsirion: I opened #4262 to discuss

dpc:
A month without an update. Still viable? Our PR backlog is growing so I'm making a round over oldest PRs to remind people about them or close non-viable ones. No stress, just trying to keep things moving. @bradleystachurski

bradleystachurski:
Closing in favor of https://github.com/fedimint/fedimint/pull/4490

PR #4255, opened 2024-02-06
fix: raw bytes Debug formatting

dpc:
Currently it's a (potentially) very long and unhelpful `[0, 52, 192, ...`


elsirion (approved):

bradleystachurski (approved):
Would it make sense to instead impl `AbbreviatedDebug`?

https://github.com/fedimint/fedimint/blob/bb850e6/fedimint-core/src/fmt_utils.rs#L208

PR #4257, opened 2024-02-07
chore: do not model write latency, read one is enough

dpc:


elsirion (approved):
I think this is ok since we are buffering sends anyway, but otherwise this could simulate a full TCP socket that takes time to send stuff out?

dpc:
> but otherwise this could simulate a full TCP socket that takes time to send stuff out?

The kernel is buffering IO to sucket as well, so unless the code is piping a lot of data saturating the link the kernel too is going to hide that latency in practice.

dpc:
> The code now is very agnostic regarding the real world. Of course we can change it, but is this going to help anything?

Not really. :)

I was suspecting this to interfere with alephbft ability to send enough data during debugging. Since we send messages one by one serially, an artificial ubuffered latency basically limits the throughput of outgoing messages in a way that wouldn't be a case on a real link. I thought it might lead to timeouts (slowdown long enough to make the test not complete on time), even if rest of alephbft is correct - just needs to exchange more messages and thus more time. However after more investigation, I don't think it's actually the case.

Issue #4040, opened 2024-01-15
OOB Notes are hard to copy / paste


justinmoon:
For example, this is an OOB note:

```
AgEEQS0qkwBTAYABo7NUVBpvj2096FtkaZxPh4l/WeilQSlm8KCkZOUzjumE65xMiH4rtcqF2Z3LYom1sP0K8AVmrXj9rOT/3ym5F/5+kp4wySg1F5NdLOrzabM=
```

The `/`, `+`, `=`, ... make it difficult to copy and paste

elsirion:
We added decoding support for URL-safe a long time ago, we should switch to it being the default encoding now with non-url-safe being the alternative encoding to give integrators an opt-out.

PR #4259, opened 2024-02-07
feat: decoding of url-safe base64 encoding for oobnotes

dpc:
To help the transition, we first implement decoding of url-safe base64 version (accepting both that and old version) and after a while we'll switch the encoding to url-safe one.

Re #4040


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -227,12 +228,21 @@ impl Decodable for OOBNotes {
...
     type Err = anyhow::Error;
 
     /// Decode a set of out-of-band e-cash notes from a base64 string.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let bytes = base64::decode(s)?;
+        let bytes = if let Ok(bytes) = BASE64_URL_SAFE.decode(s) {
+            bytes
+        } else {
+            base64::engine::general_purpose::STANDARD.decode(s)?
+        };
```
elsirion, line 245: Looked at the standard and there seems to be no case where there is a valid non-URL-safe base64 string that is also a valid url-safe base64 string encoding different data.

elsirion (approved):
Great idea!

elsirion:
Let's backport so we can make the window between introducing decoding and switching encoding longer/allow earlier switching without breaking backwards compat guarantees.

elsirion:
I see you were already a step ahead :laughing: #4260

maan2003 (approved):

PR #4260, opened 2024-02-07
feat: decoding of url-safe base64 encoding for oobnotes (v0.2)

dpc:
Backport of #4259

To help the transition, we first implement decoding of url-safe base64 version (accepting both that and old version) and after a while we'll switch the encoding to url-safe one.

Re #4040


elsirion (approved):
@justinmoon want to hold off merging so we stay maximally agile for hotfixes? This one seems benign enough to just backport tbh.

justinmoon:
Sorry I was slow to respond. Near GitHub inbox bankruptcy.

PR #4261, opened 2024-02-07
fix: remove random timeout from download of client config

joschisan:
We remove the timeout since this is the only place were we use such a timeout and therefore its unexpected behaviour, especially in a generally async system. If we want to timeout the download of the client config we should do so explicitly by wrapping either try_download_client_config or FederationInfo::from_invite_code with a timeout instead of somewhere in the middle.

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -628,7 +628,7 @@ where
     ) -> FederationResult<ClientConfig> {
         // we have to download the api endpoints first
         let id = invite_code.federation_id();
-        let qs = FilterMap::new(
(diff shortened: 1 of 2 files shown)


elsirion (approved):
How can we test this? Wasn't there another draft PR that replicated the slowness? Maybe that replication can be cherry-picked here now that it's being fixed.

justinmoon:
> How can we test this? Wasn't there another draft PR that replicated the slowness? Maybe that replication can be cherry-picked here now that it's being fixed.

https://github.com/fedimint/fedimint/pull/4226 ?

justinmoon (approved):

PR #4266, opened 2024-02-07
chore: gateway integration tests use select_client, not remove_client

justinmoon:
`remove_client()` is confusing. I spent a while wondering why cilents were being removed from the gateway. But I don't think we actually mean to remove this client from the gateway at all -- just to use it.

### DIFF
--- a/gateway/ln-gateway/tests/integration_tests.rs
+++ b/gateway/ln-gateway/tests/integration_tests.rs
@@ -219,7 +219,7 @@ async fn test_gateway_can_pay_ldk_node() -> anyhow::Result<()> {
         )
         .await?;
 
-        let gateway = gateway.remove_client(&fed).await;
+        let gateway = gateway.select_client(fed.id()).await;
         // Print money for user_client
         let dummy_module = user_client.get_first_module::<DummyClientModule>();
         let (_, outpoint) = dummy_module.print_money(sats(1000)).await?;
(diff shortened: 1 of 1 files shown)


m1sterc001guy:
So we originally used `remove_client` as a testing hack. The reason we needed to do this was because we wanted to test the gateway state machines and verify their state transitions explicitly. If we initiate a payment from the client side, the client will hit the gateway's webserver and initiate the send state machines, but the test wouldn't have a handle to verify the state transitions.

So to get around this, we actually removed the client from the gateway so that the client's web request would result in the client not being found, then the test could explicitly start the state machines. I'm ok with changing `remove_client` but we probably need to figure out a different way to fix this issue so the tests pass

justinmoon:
"build on linux" passes but code coverage fails on gateway integration tests https://github.com/fedimint/fedimint/actions/runs/7818904056/job/21330143397?pr=4266#step:6:2910

Going to re-run just in case, but yea it does seem like we can't do this.
\

justinmoon:
Not a flake 🏳️

okjodom:
> If we initiate a payment from the client side, the client will hit the gateway's webserver and initiate the send state machines, but the test wouldn't have a handle to verify the state transitions.

Looks like we can achieve the same by stopping the test gateway webserver after `select_client`, but we need to [disable this client policy](https://github.com/fedimint/fedimint/blob/c8f9611df18641982a19a9c9758e60bf913710b9/modules/fedimint-ln-client/src/lib.rs#L498) for each test

PR #4267, opened 2024-02-07
chore: document version an endpoint was introduced in ...

### LINKED ISSUE #4190: `api_endpoint!` invocation should require a version number in which an api was introduced
This argument will be throw away normally, but it will enforce documenting introduction of every single API endpoint, so we don't drown in a mess of our own making. [Original discussion](https://github.com/fedimint/fedimint/pull/4150#issuecomment-1918605620)

There might be some ways to employ these versions (e.g. an env var that would make `fedimintd` ignore endpoints newer than X, for backward compat purposes). But that's just speculation, and the goal here RN is documentation only.

dpc:
... via api_endpont macro argument.

Fix #4190

### DIFF
--- a/fedimint-server/src/net/api.rs
+++ b/fedimint-server/src/net/api.rs
@@ -29,8 +29,8 @@ use fedimint_core::epoch::ConsensusItem;
 use fedimint_core::module::audit::{Audit, AuditSummary};
 use fedimint_core::module::registry::ServerModuleRegistry;
 use fedimint_core::module::{
-    api_endpoint, ApiEndpoint, ApiEndpointContext, ApiError, ApiRequestErased, SerdeModuleEncoding,
-    SupportedApiVersionsSummary,
+    api_endpoint, ApiEndpoint, ApiEndpointContext, ApiError, ApiRequestErased, ApiVersion,
+    SerdeModuleEncoding, SupportedApiVersionsSummary,
 };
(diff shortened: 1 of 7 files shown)


fedimint-server/src/net/api.rs
```diff
@@ -432,44 +440,51 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
...
         api_endpoint! {
             AWAIT_SIGNED_SESSION_OUTCOME_ENDPOINT,
+            ApiVersion::new(0, 0),
             async |fedimint: &ConsensusApi, _context, index: u64| -> SerdeModuleEncoding<SignedSessionOutcome> {
                 Ok((&fedimint.await_signed_session_outcome(index).await).into())
             }
         },
         api_endpoint! {
             SESSION_STATUS_ENDPOINT,
+            ApiVersion::new(0, 1),
```
dpc, line 464: This one is different.

elsirion (dismissed):

justinmoon (dismissed):

justinmoon:
Seems like CI is failing due to this PR https://github.com/fedimint/fedimint/actions/runs/7820733924/job/21336093216?pr=4267

bradleystachurski (approved):

justinmoon (approved):

PR #4268, opened 2024-02-08
chore: dont fail server db migration on unknown module

m1sterc001guy:


fedimint-server/src/consensus/server.rs
```diff
@@ -116,7 +117,8 @@ impl ConsensusServer {
         for (module_id, module_cfg) in &cfg.consensus.modules {
             let kind = module_cfg.kind.clone();
             let Some(init) = module_inits.get(&kind) else {
-                bail!("Detected configuration for unsupported module id: {module_id}, kind: {kind}")
```
dpc, line 119: Why is this change needed? We can't be running fedimint server if we don't support any of the modules used for consensus.

elsirion: Afaik this will just crash at a later point, so not very useful.

m1sterc001guy: ah I missed that the modules were from consensus. Closing

PR #4269, opened 2024-02-08
v0.3.0 mainnet scripts

TonyGiorgio:
Working on some mainnet scripts, which is also turning into making the scripts a bit more flexible regarding how a federation member might want to set up, including without a gateway. Keeping as a draft, doesn't have to be merged but could be helpful for other people. 

The main part is in the `tls-download-mainnet.sh` script: 

```
curl -sSL https://raw.githubusercontent.com/tonygiorgio/fedimint/mainnet-deploy/docker/tls-download-mainnet.sh | bash
```

---

Some TBD instructions, not sure if it would be useful to put in an `.md` somewhere:

### ssh into your server

`ssh -i ~/.ssh/YOUR_ID root@<serverip>`

### update your system

```
sudo apt update
```

### install deps

```
sudo apt install docker-compose
sudo apt-get install jq
```

You might have to add your linux user to the docker group:

```
sudo usermod -aG docker YOUR_USER_HERE
```

log out and back in after you do that. You may not have to do this step, but if you get docker permission errors at the end, you'll have to do this and then log out and run the script again.

### open up the correct ports

```
sudo ufw allow 443
sudo ufw allow 8173
sudo ufw allow 9735
sudo ufw enable
sudo systemctl enable ufw
```

### run the script

```
curl -sSL https://raw.githubusercontent.com/tonygiorgio/fedimint/mainnet-deploy/docker/tls-download-mainnet.sh | bash
```

### follow the script instructions

1. say no when it asks you if you want to run an LND gateway

### after it's done

if you need to restart your server you can start up fedimint again by running
```
docker-compose up -d
```

to see the status 
```
docker-compose ps
```

to see some logs
```
docker-compose logs -f
```

### Guardian Setup

Now go to your guardian UI and start the setup flow

It should be something like: guardian-ui.fedimint.yourdomain.com

If you are just creating a one person test, go to "solo", if you are doing this with multiple people, then one person will start with "leader" and start the process. They will create the federation name and general federation info. Once the leader has an invite URL, then the rest should go through the "follower" flow. The followers will create their member info and then there's a confirmation stage where everyone confirms their session IDs together. Then the federation is set up.

### DIFF
--- a/docker/0.2.2/fedimintd-mutinynet/docker-compose.yaml
+++ b/docker/0.2.2/fedimintd-mutinynet/docker-compose.yaml
@@ -1,0 +1,43 @@
+# Important: 127.0.0.1 should be replaced by the external ip address of the machine
+version: "3"
+
+services:
+  fedimintd:
+    image: fedimint/fedimintd:v0.2.1
+    volumes:
+      - fedimintd_data:/data
+    ports:
+      - "0.0.0.0:8173:8173"
+      - "0.0.0.0:8174:8174"
+    environment:
+      - FM_BITCOIN_RPC_KIND=esplora
+      - FM_BITCOIN_RPC_URL=https://mutinynet.com/api/
+      - FM_BITCOIN_NETWORK=signet
+      - FM_BIND_P2P=0.0.0.0:8173
+      - FM_P2P_URL=fedimint://127.0.0.1:8173
+      - FM_BIND_API=0.0.0.0:8174
+      # note: for production environments it's recommended to use a https reverse proxy,
+      # so it will be prefixed by wss:// instead of ws://
+      - FM_API_URL=ws://127.0.0.1:8174
+    restart: always
+    platform: linux/amd64
+
+  guardian-ui:
+    image: fedimintui/guardian-ui:0.2.1
+    ports:
+      - "0.0.0.0:3000:3000"
+    environment:
+      - PORT=3000
+      - REACT_APP_FM_CONFIG_API=ws://127.0.0.1:8174
+    depends_on:
+      - fedimintd
+    restart: always
+    platform: linux/amd64
+
+volumes:
+  fedimintd_data:
+    driver: local
+
+# Example of running it without docker compose:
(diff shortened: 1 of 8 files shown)


dpc:
@TonyGiorgio What do we do about it? Is it done? Regrettably it looks like no one touched it for a while.

Personally I have no bandwidth to be reviewing and maintaining deployment configs, docker-compose setups, etc. It would be a shame if it went to waste though. Possibly you want to discuss what to do about it in Deep Dive call or just on Discord.

Maybe other developers want to look at it. Maybe we could have some donno... `contrib` repo/directory. Or maybe we can just mention it somewhere in the docs, etc so Google can find it if someone is looking for it.

justinmoon:
dev call: tony will update these for 0.3 ... these have some useful improvements that tony wanted to upstream

TonyGiorgio:
I have updated this to also include 0.3 mainnet scripts.

elsirion:
@Kodylow I think you were working on something similar, maybe you can coordinate. What I'd like to see is a version where the gateway is separated out (it got nothing to do with guardians) and where the server running `fedimintd` only needs one domain for both admin UI and WS server (https://github.com/fedimint/fedimint/pull/4872#pullrequestreview-1989401146).

Kodylow:
yes this is exactly what I was working on, will be adding in so it has mainnet / mutinynet in decision tree but otherwise the same

TonyGiorgio:
Will maintain this elsewhere.

markskram:
The old version of `docker-compose` was giving me problems on Digital Ocean. I upgraded to v2 and used `docker compose up --detatch` to get it working. https://github.com/TonyGiorgio/fedimint/pull/1

TonyGiorgio:
I have moved this here: https://github.com/MutinyWallet/fedimint-deployment-docs

oleonardolima:
> The old version of `docker-compose` was giving me problems on Digital Ocean and Ubuntu 24.04 LTS. I upgraded to v2 and used `docker compose up --detatch` to get it working. [TonyGiorgio#1](https://github.com/TonyGiorgio/fedimint/pull/1)

btw I faced the same problem with the current scripts on master, it only worked on 22.04 lts (without the docker-compose changes).

PR #4270, opened 2024-02-08
fix: document consensus

joschisan:


fedimint-server/src/consensus/server.rs
```diff
@@ -454,7 +458,11 @@ impl ConsensusServer {
...
                     let pending_accepted_items = self.pending_accepted_items().await;
 
-                    let (processed, unprocessed) = signed_session_outcome.session_outcome.items.split_at(pending_accepted_items.len());
+                    // this panics if we have more accepted items than the signed session outcome
+                    let (processed, unprocessed) = signed_session_outcome
+                        .session_outcome
+                        .items
+                        .split_at(pending_accepted_items.len());
 
                     assert!(processed.iter().eq(pending_accepted_items.iter()));
```
elsirion, line 467: I think a message in that assert (that gets shown in the panic if it fails) would also be nice

elsirion (approved):

justinmoon:
I took one note in our call today that `complete_signed_session_outcome` could use a doc string

dpc (approved):

Issue #3985, opened 2023-12-21
Devimint supports multiple feds with shared daemons


okjodom:
We might want to run dev federations while sharing daemons like bitcoind, electrs, esplora, lightning nodes and gateways. This would be useful in a scenario like #3696

justinmoon:
Related to https://github.com/fedimint/fedimint/issues/3209

elsirion:
@m1sterc001guy didn't you implement this recently?

m1sterc001guy:
Ah yes I did! Can be closed.

PR #4275, opened 2024-02-08
test: gateway short channel id assignments

okjodom:
- adds coverage for short channel id assignments when federations are connected to gateway
- shows  each scid assignment info for each connected federation
- shows all scid assignments in gateway info. This includes scids assigned to federations that have since been removed
- sets const `INITIAL_SCID = 0` because we always increment the scid value before assignment.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -1229,6 +1229,12 @@ impl Gateway {
             .into_value();
 
         client.wait_until_fully_dropped().await;
+
+        // Remove previously assigned scid from `scid_to_federation` map
+        self.scid_to_federation
+            .write()
+            .await
+            .retain(|_, fid| *fid != federation_id);
         Ok(())
     }
 
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -964,7 +964,16 @@ impl Gateway {
         Err(GatewayError::Disconnected)
     }
 
-    pub async fn handle_leave_federation(&mut self, payload: LeaveFedPayload) -> Result<()> {
+    pub async fn handle_leave_federation(
+        &mut self,
+        payload: LeaveFedPayload,
+    ) -> Result<FederationInfo> {
+        let federation_info = {
+            let client = self.select_client(payload.federation_id).await?;
+            self.make_federation_info(client.value(), payload.federation_id)
+                .await
+        };
+
         // TODO: This should optimistically try to contact the federation to remove the
         // registration record
         let _client_joining_lock = self.client_joining_lock.lock().await;
@@ -977,7 +986,8 @@ impl Gateway {
         .await;
         dbtx.commit_tx_result()
             .await
-            .map_err(GatewayError::DatabaseError)
+            .map_err(GatewayError::DatabaseError)?;
+        Ok(federation_info)
     }
 
     pub async fn handle_backup_msg(
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -272,9 +272,9 @@ pub struct Gateway {
     // A public key representing the identity of the gateway. Private key is not used.
     pub gateway_id: secp256k1::PublicKey,
 
-    // ID generator that atomically increments. Used for creation of new short channel ids that
-    // represent federations.
-    channel_id_generator: Arc<Mutex<u64>>,
+    // Tracker for short channel ID assignments. When connecting a new federation,
+    // this value is incremented and assigned to the federation as the `mint_channel_id`
+    max_used_scid: Arc<Mutex<u64>>,
 }
 
 impl std::fmt::Debug for Gateway {
@@ -287,7 +287,7 @@ impl std::fmt::Debug for Gateway {
             .field("clients", &self.clients)
             .field("scid_to_federation", &self.scid_to_federation)
             .field("gateway_id", &self.gateway_id)
-            .field("channel_id_generator", &self.channel_id_generator)
+            .field("max_used_scid", &self.max_used_scid)
             .finish()
     }
 }
@@ -325,7 +325,7 @@ impl Gateway {
             client_joining_lock: Arc::new(Mutex::new(ClientsJoinLock)),
             scid_to_federation: Arc::new(RwLock::new(BTreeMap::new())),
             gateway_id: Gateway::get_gateway_id(gateway_db).await,
-            channel_id_generator: Arc::new(Mutex::new(INITIAL_SCID)),
+            max_used_scid: Arc::new(Mutex::new(INITIAL_SCID)),
         })
     }
 
@@ -360,7 +360,7 @@ impl Gateway {
             lightning_builder: Arc::new(GatewayLightningBuilder {
                 lightning_mode: opts.mode.clone(),
             }),
-            channel_id_generator: Arc::new(Mutex::new(INITIAL_SCID)),
+            max_used_scid: Arc::new(Mutex::new(INITIAL_SCID)),
             gateway_parameters: opts.to_gateway_parameters()?,
             state: Arc::new(RwLock::new(GatewayState::Initializing)),
             client_builder,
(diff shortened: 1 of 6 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -911,7 +909,14 @@ impl Gateway {
                 .build(gw_client_cfg.clone(), self.clone())
                 .await?;
 
-            let federation_info = self.make_federation_info(&client, federation_id).await;
+            // Instead of using `make_federation_info`, we manually create federation info
+            // here because short channel id is not yet persisted
```
m1sterc001guy, line 913: I'm not sure I understand this comment. Doesn't `make_federation_info` do the same thing?

gateway/ln-gateway/src/lib.rs
```diff
@@ -938,6 +943,7 @@ impl Gateway {
                 .write()
                 .await
                 .insert(mint_channel_id, federation_id);
+            info!("SCID {mint_channel_id} assigned to {federation_id}");
```
m1sterc001guy, line 946: Let's make this a little more informative. Can we add some statement that says a new federation was connected? And perhaps we want to move this to below `save_config`. It might be kind of confusing to see this message and then have `save_config` fail for whatever reason

gateway/ln-gateway/src/lib.rs
```diff
@@ -950,18 +956,34 @@ impl Gateway {
...
+
+        let federation_info = self
+            .make_federation_info(client.value(), payload.federation_id)
+            .await;
+
+        // Remove previously assigned scid from `scid_to_federation` map
+        self.scid_to_federation
+            .write()
+            .await
+            .retain(|_, fid| *fid != payload.federation_id);
```
m1sterc001guy, line 975: I'm confused by this. Why are we retaining it? I think we should ideally remove it from `scid_to_federation` when we leave the federation, I'm surprised we didn't do this already in `remove_client`.

Is this so we can track short channel ids from federations that we've left?

okjodom: the predicate in retain has `!=` to make sure we remove this scid. I know. Looks confusing at fist glance

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Oh I see `make_federation_info` has changed

gateway/ln-gateway/src/lib.rs
m1sterc001guy: I don't think I like the idea of using `scid_to_federation` has the way of keeping around old short channel ids. If I'm reading this correctly, when the gateway reboots, it re-reads the short channel ids from the configs that are persisted in the database. In this case, since previously the gateway just left a federation, the config will no longer be there, and the `scid_to_federation` will not contain this federation either. Is this the intended behavior?

gateway/ln-gateway/src/lib.rs
okjodom: Updated. I've also dropped log level to debug because this handler returns an informative result immediately after the log point

gateway/ln-gateway/src/lib.rs
okjodom: I think the function name `retain` is confusing, but the predicate logic actually removes the scid that was previously assigned to the federation client we are removing.
> I think we should ideally remove it from scid_to_federation when we leave the federation

It results in this behavior you described, i.e. `scid_to_federation` only retains scids that are assigned to currently connected federations, even beyond gateway reboot

m1sterc001guy: Got it. Can we move this into `remove_client`? Seems to be the more natural spot to me

okjodom: Agreed that is more natural. It was challenging because we need to `make_federation_info` with the client we just removed. I did it here as a convenience so I can still use the make federation helper, then it seemed like a reasonable fit because we also do other irreversible cleanup operations here like db updates

m1sterc001guy: Can we call `make_federation_info` first then remove the client? I think this should be in `remove_client`, we call `remove_client` in our tests in addition to here and it seems wrong to me that that function doesn't remove it from `scid_to_federation` map

okjodom: Naa, we should revive #4266  and delete `remove_client`. I'll make an attempt of it, picking off from @justinmoon 's investigation

okjodom: @m1sterc001guy, should we go in this direction?  https://github.com/fedimint/fedimint/pull/4266#issuecomment-1936778256

m1sterc001guy: I'm open to a workaround for our tests so that we can use `select_client` instead of `remove_client`, but I think that will take some time. I think we should move this logic into `remove_client` for now and if we need we can move it back out when we have a fix for our tests that allows us to delete `remove_client`

okjodom: refactored :)

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1167,6 +1167,82 @@ async fn test_gateway_shows_info_about_all_connected_federations() -> anyhow::Re
     .await
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_gateway_can_leave_connected_federations() -> anyhow::Result<()> {
```
m1sterc001guy, line 1171: Should we also add a devimint test that reboots the gateway in between to verify that the short channel ids are correct after the restart (or maybe the existing devimint test can be modified to cover this)?

m1sterc001guy (dismissed):
LGTM

gateway/ln-gateway/src/lib.rs
```diff
@@ -1240,8 +1240,8 @@ impl Gateway {
                 warn!("Failed to load client for federation: {federation_id}");
             }
 
-            if config.mint_channel_id > next_channel_id {
-                next_channel_id = config.mint_channel_id + 1;
+            if scid > next_channel_id {
+                next_channel_id = scid + 1;
```
elsirion, line 1268: Doing the "max used channel id" calculation in this loop isn't very readable. I'd much prefer iterating over all configs a second time to determine the next channel id to be given out:

```rust
let next_channel_id = configs.iter()
    .map(|cfg| cfg.mint_channel_id)
    .max() + 1;
```

After writing this I also think the current code is subtly broken: if `scid == next_channel_id` then `next_channel_id` won't be incremented even though the SCID is already in use.

gateway/ln-gateway/tests/integration_tests.rs
justinmoon: Yea the bug we observed *only* happened after restarting the gateway.

douglaz: The duplicated id was certainly after restarting. But we also had issues (same? others?) before restarting.

gateway/ln-gateway/src/lib.rs
okjodom: ah yes. Fixed this by track just the max used scid. we only need increment the value when connecting new federations

gateway/ln-gateway/tests/integration_tests.rs
okjodom: Adding these tests highlighted several bugs / slight logical with scid assignments. If we backported this to 0.2.1 and updated a live gateway, future scid assignments when connecting / disconnecting feds would be predictable as defined by the new test spec.

> Tricky thing is that we may need at least 2 federations for this devimint test?

Yes. We have a [gateway reboot test](https://github.com/fedimint/fedimint/blob/412aaf965805263f397364242618df5a27cc7b58/devimint/src/tests.rs#L1342), but we can't improve it's coverage to scid assignments because devimint is limited. Relevant to #3985

> We can't really tests a gatewayd restart in this test because it uses memdb?

Well this also is a limitation for testing persisted states

elsirion:
Needs rebase

elsirion (approved):

justinmoon (approved):

PR #4277, opened 2024-02-08
refactor(nix): split workspaceCov into separate build and test steps

dpc:


justinmoon (approved):
Don't fully understand it, but I'm trusting you do!

bradleystachurski (approved):
This commit has already proved useful in https://github.com/fedimint/fedimint/pull/4278

👍

PR #4282, opened 2024-02-09
chore: client state machine migrations

m1sterc001guy:
Next step towards addressing https://github.com/fedimint/fedimint/issues/3417

Adds Client State Machine Migrations.

This PR adds the `ClientMigrationMap`, which is analogous to `ServerMigrationMap`, but adds `active_states` and `inactive_states` as additional parameters to the migration function, and expected `new_active_states` and `new_inactive_states` in return. These new states represent the new set of states that should be written to the database.

Clients can now define their migrations using the database transaction or just by reading the states provided by the migration function.

Add client db migration has been added to the dummy module, just by adding a state that can never be reached, and migrating it to the `Input` state.


modules/fedimint-dummy-client/src/db.rs
```diff
@@ -57,5 +63,41 @@ pub async fn migrate_to_v1(dbtx: &mut DatabaseTransaction<'_>) -> anyhow::Result
...
     }
 
-    Ok(())
+    Ok(None)
+}
+
+pub async fn migrate_to_v2(
+    module_instance_id: ModuleInstanceId,
+    active_states: Vec<DynState>,
+    inactive_states: Vec<DynState>,
```
m1sterc001guy, line 72: I don't like that I needed to pass `DynState` here instead of something like `T::States`. But I couldn't figure out how to add a generic to the `Fn` in the `ClientMigrationMap`. @dpc or @Maan2003 if you guys have any rust pointers here :)

fedimint-testing/src/db.rs
```diff
@@ -82,22 +91,27 @@ where
...
             bail!("{FM_PREPARE_DB_MIGRATION_SNAPSHOTS_ENV} set, but {} already exists already exists. Set to 'force' to overwrite.", snapshot_dir.display());
         }
         (Some(_), false) => {
+            println!("OPENING SNAPSHOT DB...");
             debug!(dir = %snapshot_dir.display(), "Snapshot dir does not exist. Creating.");
             let db = open_snapshot_db(decoders, snapshot_dir, is_isolated).await?;
             prepare_fn(db).await;
         }
         (None, true) => {
+            println!("SNAPSHOT ALREADY EXISTS, NOTHING TODO");
```
m1sterc001guy, line 110: Will remove

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,155 @@ impl_db_lookup!(
...
+    dbtx.find_by_prefix(&InactiveStateKeyPrefix::new())
+        .await
+        .map(|(state, _)| state.state)
+        .collect::<Vec<_>>()
+        .await
+}
+
+pub async fn persist_active_states(
+    dbtx: &mut DatabaseTransaction<'_>,
+    new_active_states: Vec<(ActiveStateKey, ActiveState)>,
```
m1sterc001guy, line 408: This should probably just take the vector as `Vec<DynState>` and pass the `module_instance_id` as a parameter and do the mapping inside this function

modules/fedimint-dummy-client/src/db.rs
m1sterc001guy: Maybe this is sufficient for now

modules/fedimint-dummy-tests/tests/tests.rs
```diff
@@ -204,10 +210,15 @@ mod fedimint_migration_tests {
         dbtx.commit_tx().await;
     }
 
+    fn create_client_states() -> (Vec<DummyStateMachine>, Vec<DummyStateMachine>) {
```
m1sterc001guy, line 213: This should add a few states that won't be migrated so that we verify that we don't accidentally delete any states

modules/fedimint-dummy-client/src/db.rs
maan2003: you will have to use trait instead of `Fn` in that case, but this is fine being untyped

m1sterc001guy: I see. Right now that seems like a good improvement but maybe in a separate PR.

One unfortunate side affect of this being untyped though is that the higher layer does not know which state is applicable to which module. So all of the states are passed into each module, and each module is responsible for handling states that it doesn't know about.

Since the migration function expects the entire set of states to be returned back, a malicious module could actually drop states that belong to other modules. This is definitely not ideal and I think the proper fix is to make this typed.

dpc: At very least filtering by `dyn_state.module_instance_id()` should be trivial.

Generic down and up-casting is definitely doable as well, but could be done as a next step.

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,173 @@ impl_db_lookup!(
...
+                .await?
+            } else {
+                warn!("Missing client db migration for version {current_db_version}");
+                None
+            };
+
+            // If the client migration returned new states, a state machine migration has
+            // occurred, and the new states need to be persisted to the database.
+            if let Some((new_active_states, new_inactive_states)) = new_states {
+                persist_active_states(&mut global_dbtx.to_ref_nc(), new_active_states).await;
```
dpc, line 362: Where do we delete the old states?

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,173 @@ impl_db_lookup!(
...
+        target_db_version
+    };
+
+    global_dbtx.commit_tx_result().await?;
+    info!(target: LOG_DB, "{} module db version: {}", kind, db_version);
+    Ok(())
+}
+
+/// Reads all active states from the database and returns `Vec<DynState>`.
+pub async fn get_active_states(dbtx: &mut DatabaseTransaction<'_>) -> Vec<DynState> {
```
dpc, line 383: The fact that we can't list only states of a given module is pretty bad. Can you add a comment to `ActiveStateKey: `TODO: we would like to have a module_instance_id prefix to be able to query only states of a givem module id`?

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,173 @@ impl_db_lookup!(
...
+
+/// Persists new active states by first removing all current active states, and
+/// re-writing with the new set of active states. `new_active_states` is
+/// expected to contain all active states, not just the newly created states.
+pub async fn persist_active_states(
+    dbtx: &mut DatabaseTransaction<'_>,
+    new_active_states: Vec<DynState>,
+) {
+    // Remove all existing active states
+    dbtx.remove_by_prefix(&ActiveStateKeyPrefix::new()).await;
```
dpc, line 408: With filtering by module_instance_id we should be able to delete only module's own states.

fedimint-client/src/sm/executor.rs
```diff
@@ -170,7 +170,7 @@ impl Executor {
 
             dbtx.insert_entry(
                 &ActiveStateKey::from_state(state.clone()),
-                &ActiveState::new(),
```
dpc, line 173: @elsirion  Should we rename `ActiveState` to something like `ActiveStateInfo`, `ActiveStateData`, `ActiveStateMeta`, etc.? The way things are RN it only contains `created_at`, the actual data is in the key, and we call it `meta: ActiveState` in a bunch of places, which was confusing me a lot.

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,173 @@ impl_db_lookup!(
     key = ClientModuleRecovery,
     query_prefix = ClientModuleRecoveryPrefix
 );
+
+/// ClientMigrationMap is a BTreeMap that maps DatabaseVersions to async
+/// functions. These functions are expected to "migrate" the database from the
+/// keyed DatabaseVersion to DatabaseVersion + 1.
+pub type ClientMigrationMap = BTreeMap<
```
dpc, line 283: I don't like the name, but need more coffee to rationalize why. :D

Would consider wrapping it in a newtype.

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,173 @@ impl_db_lookup!(
...
+    DatabaseVersion,
+    for<'r, 'tx> fn(
```
dpc, line 285: `type MigrationFn = ...` or something it.

dpc: Nice type with `for ...`, two lifetimes and pining and futures. Proper nice async Rust. :D

fedimint-client/src/db.rs
m1sterc001guy: In this function :) Perhaps this is a bad name, will change

fedimint-client/src/db.rs
m1sterc001guy: Sure, I can add the todo. This would unfortunately require a somewhat nasty database migration since it is not module specific, but it can be done similar to how we just did the database version.

fedimint-client/src/sm/executor.rs
m1sterc001guy: +1. This confused me as well when first looking at it. Should I change now or leave or separate PR?

fedimint-client/src/db.rs
m1sterc001guy: Made an issue, assigned myself https://github.com/fedimint/fedimint/issues/4287

dpc (commented):
I think the general mechanism works, there might be some improvements to be made.

The dyn to static conversion will be the biggest one.

Another one would be introducing laziness point, so we can load these state machines if the migration actually plans to do something with them, instead of eagerly.

fedimint-client/src/db.rs
m1sterc001guy: `BTreeMap<DatabaseVersion, ClientMigrationFn>` is probably more legible.

fedimint-client/src/sm/executor.rs
m1sterc001guy: Not as bad of a change as I initially thought, changed these to `ActiveStateMeta` and `InactiveStateMeta` in a separate commit

fedimint-client/src/db.rs
m1sterc001guy: Done

fedimint-client/src/db.rs
m1sterc001guy: TODO added

m1sterc001guy:
> Another one would be introducing laziness point, so we can load these state machines if the migration actually plans to do something with them, instead of eagerly.

Could you expand on this? Not sure I follow. Agreed on the other points.

dpc:
> Another one would be introducing laziness point, so we can load these state machines if the migration actually plans to do something with them, instead of eagerly.

@m1sterc001guy RN, you need to read the existing state machines eagerly before you call the migration fn. Most of the time the migration will just throw them away.

If instead the migration fn could somehow "request" the state machines when it actually needs them, it would be lazy. E.g. migration function could take an argument function `migrate_state_machines: FnOnce((Vec<DynState>, Vec<DynState>)) -> ((Vec<DynState>, Vec<DynState>))` and then call it (or not) to do the migration there. It will have to be async, and there will probably be some lifetime issues and so on, so it will be a bit gnarly but possible. The good part is, the migration function itself can stop returning `Option<(Vec<DynState>, Vec<DynState>)>`

Then to help with all gnarliness we could then change all these arguments passed to migration into one `struct ClientMigrationArgs`, add generaics, and as a side effect, make adding other arguments in the future not be a API breaking change.

m1sterc001guy:
Got it, makes sense, thank you. I agree that's a good improvement, probably best done after https://github.com/fedimint/fedimint/issues/4287

modules/fedimint-dummy-client/src/db.rs
m1sterc001guy: I added filtering logic by `module_instance_id` in `get_active_states` and the other functions, so this isn't as much as issue. Modules should only receive states from their state machines now. I'll leave making this typed for a different PR

fedimint-client/src/db.rs
```diff
@@ -441,9 +479,25 @@ pub async fn remove_old_and_persist_new_active_states(
 pub async fn remove_old_and_persist_new_inactive_states(
     dbtx: &mut DatabaseTransaction<'_>,
     new_inactive_states: Vec<DynState>,
+    module_instance_id: ModuleInstanceId,
 ) {
     // Remove all existing active states
-    dbtx.remove_by_prefix(&InactiveStateKeyPrefix::new()).await;
+    let states_to_remove = dbtx
```
dpc, line 485: Don't you already have this one from the time you queried it?

m1sterc001guy: ah good point, fixed

elsirion (commented):
I'm not sure if it's an intentional limitation for now or an oversight, but what do we do if we want to change the encoding of states? This PR only allows semantic changes, where one state is changed to (a potentially newly introduced) other, but not changing the on-disk format used for states.

m1sterc001guy:
> I'm not sure if it's an intentional limitation for now or an oversight, but what do we do if we want to change the encoding of states?

Hmm let me know if you have another use case in mind, but I believe this can be addressed by introducing a "new" state that is functionally equivalent to the old one, but has a different encoding of the data. Then the migration could just map the old states to this new state.

I think if we want to change the encoding for _all_ states, we would need to do a migration similar to what we did for the database version, where migration is written at the higher layer.

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,221 @@ impl_db_lookup!(
...
     query_prefix = ClientModuleRecoveryPrefix
 );
+
+/// `ClientMigrationFn` is a function that modules can implement to "migrate"
+/// the database to the next database version.
+pub type ClientMigrationFn = for<'r, 'tx> fn(
+    &'r mut DatabaseTransaction<'tx>,
+    ModuleInstanceId,
+    Vec<DynState>, // active states
+    Vec<DynState>, // inactive states
```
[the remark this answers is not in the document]
m1sterc001guy: In order for the module to do be able to determine which state the raw bytes are, the decoder still needs to be registered with the db right? If we can't decode the old format I don't think the migration can do anything interesting (maybe I'm missing a use case here)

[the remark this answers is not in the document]
m1sterc001guy: dev call: using bytes here allows the migration to change the fields of a struct. Old struct needs to still be kept around so it can be decoded, but it can be removed from the state machine enum

fedimint-client/src/db.rs
```diff
@@ -262,3 +276,221 @@ impl_db_lookup!(
...
+
+        if current_db_version > target_db_version {
+            return Err(anyhow::anyhow!(format!(
+                "On disk database version for module {kind} was higher than the code database version."
+            )));
+        }
+
+        while current_db_version < target_db_version {
+            let active_states =
+                get_active_states(&mut global_dbtx.to_ref_nc(), module_instance_id).await;
```
m1sterc001guy, line 337: Call: this can be made more efficient by querying for the states first, then filtering per-module

m1sterc001guy: Fixed. Now we read the states at the beginning of the migration process and pass the new states to the next version migration.

fedimint-client/src/db.rs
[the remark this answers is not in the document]
m1sterc001guy: Unless I have a misunderstanding, I think passing `DynState` here is more convenient for module developers, since they can always call `to_bytes` on the state, decode it using `consensus_decode`, and do the migration with the resulting struct (which doesn't need to be a state listed in the module's state machine).

m1sterc001guy:
Tests are failing since querying the states before they are registered with the `ModuleDecoderRegistry` does not work. I'll need to figure out a different way to query the states

fedimint-client/src/db.rs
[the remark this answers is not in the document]
m1sterc001guy: Ah nevermind, states that have been deleting can no longer be encoded as `DynState` so this will not work. `Vec<u8>` is the only option, working on a test for this now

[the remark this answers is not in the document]
m1sterc001guy: Fixed. Changed this to `Vec<u8>` and changed the test to be as close to a state deletion as possible.

m1sterc001guy:
> Tests are failing since querying the states before they are registered with the ModuleDecoderRegistry does not work. I'll need to figure out a different way to query the states

Fixed, this is ready for more review

fedimint-client/src/db.rs
```diff
@@ -262,3 +278,250 @@ impl_db_lookup!(
...
 );
+
+/// `ClientMigrationFn` is a function that modules can implement to "migrate"
+/// the database to the next database version.
+pub type ClientMigrationFn = for<'r, 'tx> fn(
+    &'r mut DatabaseTransaction<'tx>,
+    ModuleInstanceId,
+    Vec<(Vec<u8>, OperationId)>, // active states
+    Vec<(Vec<u8>, OperationId)>, // inactive states
+    ModuleDecoderRegistry,
```
elsirion, line 289: Do we actually need this? The module should be able to decode its own SMs and they shouldn't contain any data specific to other modules? Maybe with the exception of the transaction submission SM, which doesn't belong to a module anyway though. Let's keep it for flexibility though for now.

fedimint-client/src/db.rs
```diff
@@ -262,3 +278,250 @@ impl_db_lookup!(
...
+                        .into_nc(),
+                    module_instance_id,
+                    active_states.clone(),
+                    inactive_states.clone(),
+                    decoders.clone(),
+                )
+                .await?
+            } else {
+                warn!("Missing client db migration for version {current_db_version}");
+                None
```
elsirion, line 364: Should this be a hard error in case we ever remove wayyyy too old migrations?

elsirion (approved):
Looks good :)

fedimint-client/src/db.rs
m1sterc001guy: I don't think its strictly necessary but I included it because I wanted to use `DynState::from_bytes` to convert to the typed states. I'm ok with keeping it for flexibility.

fedimint-client/src/db.rs
m1sterc001guy: Ah I was actually thinking of a different use case, where if a module developer accidentally increments the `DATABASE_VERSION` too high, in which case there would be intermediate versions with no migrations. I didn't want to explicitly fail in that case. 

I think I'll keep it like this for now, if we do make the decision to cleanup old migrations in the future, we should probably add additional sanity checks and change this to an explicit error.

fedimint-client/src/sm/executor.rs
```diff
@@ -843,6 +843,45 @@ impl Decodable for ActiveStateKey {
     }
 }
 
+#[derive(Debug)]
+pub struct ActiveStateKeyBytes {
```
dpc, line 847: Comment missing, I'm confused what's this for. :D

dpc (approved):

dpc:
Flaked on https://github.com/fedimint/fedimint/pull/4381

m1sterc001guy:
I suspect this actually makes `test_can_change_routing_fees` a little more flaky, so I'm going to debug that next

PR #4294, opened 2024-02-11
fix: test_restart_setup doesn't allocate enough ports

dpc:


dpc:
Funny how fix for one test flakes on another flaky test... :D

fedimint-server/src/config/api.rs
```diff
@@ -1128,8 +1130,8 @@ mod tests {
 
         apis.push(api);
 
-        for i in 1..4 {
-            let port = base_port + (i * 2);
+        for i in 1..PEER_NUM {
+            let port = base_port + (i * PORTS_PER_PEER);
```
elsirion, line 1134: Why not just call `port_alloc(PORTS_PER_PEER)` every time instead of doing these manual calculations?

dpc: :brain:

PR #4296, opened 2024-02-11
cleanup: remove operation id argument from await_tx_accepted

joschisan:
This is another step in the direction of making client state machines unaware of their operation id.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -224,10 +224,7 @@ pub trait IGlobalClientContext: Debug + MaybeSend + MaybeSync + 'static {
         sm: Box<maybe_add_send_sync!(dyn IState)>,
     ) -> AddStateMachinesResult;
 
-    async fn transaction_update_stream(
-        &self,
-        operation_id: OperationId,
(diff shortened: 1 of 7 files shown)


elsirion:
I don't think this can work: `Notifier::subscribe` needs the operation ID since it loads all past state transitions from disk to avoid missing any in case the client got restarted. Loading state transitions from disk is only feasible if we can limit the scope, this is done using the `operation_id`.

It's ok to load tens of unrelated state transitions from disk and filter them out, not doing so with potentially thousands (without the operation scoping).

joschisan:
@elsirion I thing there is a misunderstanding here, Notifier::subscribe still takes the operation id as an argument. Its just that ModuleGlobalClientContext already has the operation id as an attribute and can therefore call self.client.transaction_update_stream(self.operation). Works the same with fund_output_dyn and claim_input_dyn.

elsirion:
Makes a lot of sense, I was wondering why you didn't have to change the notify impl, but then saw compile errors in CI and thought maybe it was still WIP. Sorry, should have looked deeper.

elsirion (approved):

m1sterc001guy (approved):
Makes sense to me

PR #4297, opened 2024-02-11
Remove FederationInfo

joschisan:
Many methods were never used, the invite_code field was never read. Every time this struct was created from an invite code the caller just wanted to obtain the client config.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -103,7 +103,7 @@ use fedimint_core::module::{
     ApiVersion, MultiApiVersion, SupportedApiVersionsSummary, SupportedCoreApiVersions,
     SupportedModuleApiVersions,
 };
-use fedimint_core::task::{sleep, MaybeSend, MaybeSync, TaskGroup};
+use fedimint_core::task::{MaybeSend, MaybeSync, TaskGroup};
 use fedimint_core::transaction::Transaction;
 use fedimint_core::util::{BoxStream, NextOrPending};
 use fedimint_core::{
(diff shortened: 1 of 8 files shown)


elsirion (changes requested):
The whole point of this struct is to allow 3rd party integrators to gather some basic information about the federation to show to users before joining. I guess we could try to merge its functionality with the `ClientConfig` type now that we can generate `InviteCode`s from config again (that wasn't possible when this type was created).

joschisan:
I think porting the functionality to the client config makes sense as FederationInfo is practically only a wrapper around the config since we can generate invite codes from the client config. What functionality would you like to see ported?

elsirion:
The whole public interface except `from_config` I guess, since there was a reason to add ever fn (not inside our code base, but really useful when building a client app).

fedimint-client/src/lib.rs
```diff
@@ -2214,29 +2159,66 @@ pub async fn get_invite_code_from_db(db: &Database) -> Option<InviteCode> {
 
 /// Tries to download the client config from the federation,
 /// attempts up to `retries` number times
```
elsirion, line 2161: `retries` isn't an argument anymore.

fedimint-client/src/lib.rs
```diff
@@ -2214,29 +2159,66 @@ pub async fn get_invite_code_from_db(db: &Database) -> Option<InviteCode> {
...
-async fn try_download_config(
-    invite_code: InviteCode,
-    max_retries: usize,
-) -> anyhow::Result<ClientConfig> {
+pub async fn download_client_config(invite_code: &InviteCode) -> anyhow::Result<ClientConfig> {
```
elsirion, line 2162: Can we make this a constructor/member of `ClientConfig` to replicate the `FederationInfo` interface. I.e. the user should be able to call `ClientConfig::download_with_invite(invite)` or something similar.

fedimint-core/src/config.rs
```diff
@@ -190,6 +191,26 @@ impl ClientConfig {
...
+        };
+        serde_json::from_str(str_value).context(format!("Decoding meta field '{key}' failed"))
+    }
+
+    /// Creates an API client for the federation
+    pub fn api(&self) -> DynGlobalApi {
+        DynGlobalApi::from_config(self)
+    }
+
+    pub fn invite_code(&self, peer: &PeerId) -> Option<InviteCode> {
```
elsirion, line 208: Documentation needed: what does the `peer` argument do?

joschisan: Done

fedimint-client/src/lib.rs
joschisan: Done

fedimint-client/src/lib.rs
joschisan: Done

fedimint-core/src/config.rs
```diff
@@ -190,6 +196,94 @@ impl ClientConfig {
...
+    /// be used to download this client config
+    pub fn invite_code(&self, peer: &PeerId) -> Option<InviteCode> {
+        self.global
+            .api_endpoints
+            .get(peer)
+            .map(|peer_url| InviteCode::new(peer_url.url.clone(), *peer, self.federation_id()))
+    }
+
+    /// Tries to download the client config from the federation,
+    /// attempts to retry teb times before giving up.
```
elsirion, line 223: Nit: `s/teb/ten/`?

elsirion (approved):

fedimint-core/src/config.rs
elsirion: still relevant, but whatever …

elsirion (approved):

elsirion:
Might be worth incorporating #4371 into this PR too or a follow-up.

justinmoon (approved):

PR #4303, opened 2024-02-12
fix(flake): don't install some irrelevant things

dpc:


flake.nix
```diff
@@ -293,9 +293,6 @@
                   pkgs.nodePackages.bash-language-server
                 ] ++ lib.optionals (!stdenv.isAarch64 && !stdenv.isDarwin) [
                   pkgs.semgrep
-                ] ++ lib.optionals (stdenv.isLinux) [
-                  pkgs.xclip
-                  pkgs.wl-clipboard
```
maan2003, line 298: these are clipboard for in mprocs, but user should install this globally.

@elsirion was having issues

justinmoon: Yea IIRC those can be useful. Perhaps we should have comments explaining what deps are used for?

dpc: I started looking into it because availability on `xclip` make helix not generate OSC 52 codes. And as a wayland user, I don't want  xclip on my desktop boxes anyway.

PR #4304, opened 2024-02-12
feat: wasm ecash tests

maan2003:
hopefully this will catch more wasm bugs on master

### DIFF
--- a/fedimint-wasm-tests/src/lib.rs
+++ b/fedimint-wasm-tests/src/lib.rs
@@ -88,11 +88,15 @@ mod faucet {
 
 wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
 mod tests {
+    use std::time::Duration;
+
+    use anyhow::anyhow;
     use fedimint_client::derivable_secret::DerivableSecret;
     use fedimint_core::Amount;
     use fedimint_ln_client::{
         LightningClientModule, LnPayState, LnReceiveState, OutgoingLightningPayment, PayType,
     };
+    use fedimint_mint_client::{MintClientModule, ReissueExternalNotesState};
     use futures::StreamExt;
     use wasm_bindgen_test::wasm_bindgen_test;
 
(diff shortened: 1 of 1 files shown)


fedimint-wasm-tests/src/lib.rs
```diff
@@ -208,4 +212,43 @@ mod tests {
...
+            .subscribe_reissue_external_notes(operation_id)
+            .await?
+            .into_stream();
+        loop {
+            match updates.next().await {
+                Some(ReissueExternalNotesState::Done) => {
+                    break;
+                }
+                Some(ReissueExternalNotesState::Failed(error)) => {
+                    return Err(anyhow::anyhow!("reissue failed {error}"));
```
bradleystachurski, line 234: [nit] 
```suggestion
                    return Err(anyhow!("reissue failed {error}"));
```

fedimint-wasm-tests/src/lib.rs
```diff
@@ -208,4 +212,43 @@ mod tests {
 
         Ok(())
     }
+
+    async fn send_and_recv_ecash_once(
+        client: fedimint_client::ClientArc,
+    ) -> Result<(), anyhow::Error> {
+        let mint = client.get_first_module::<MintClientModule>();
+        let (_, notes) = mint
+            .spend_notes(Amount::from_sats(11), Duration::from_secs(10000), false, ())
```
bradleystachurski, line 221: I've noticed 10 seconds for `try_cancel_after` in other mint tests. Would it make sense to reduce here?

bradleystachurski (commented):
Thanks for the additional coverage! Curious about the `try_cancel_after` value, but otherwise looks good.

fedimint-wasm-tests/src/lib.rs
dpc: This is for cancellation that shouldn't be needed here and isn't waited for. I'd be a bit worried that it's too little and under heavy load possibly it could fail due to it?

dpc (dismissed):

bradleystachurski (approved):

fedimint-wasm-tests/src/lib.rs
maan2003: no, we never want client to cancel the ecash here, because are reissuing it in very next step.

PR #4258, opened 2024-02-07
chore: check out if nix-fast-build would work for us

dpc:

PR #4314, opened 2024-02-13
fix: switch back to `nix build` for release builds on master

dpc:
For these we do want to produce the final binary every time, and their runtime is not as important.

Re #4258

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -412,7 +412,7 @@ jobs:
             deb: fedimint-gateway
 
     runs-on: buildjet-4vcpu-ubuntu-2004
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v4
       - uses: cachix/install-nix-action@v25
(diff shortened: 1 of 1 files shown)


oleonardolima (approved):
ACK

.github/workflows/ci-nix.yml
```diff
@@ -444,7 +452,7 @@ jobs:
         run: |
           bins="${{ matrix.build.bins }}"
           for bin in ${bins//,/ } ; do
-            nix run "github:Mic92/nix-fast-build?rev=4376b8a33b217ee2f78ba3dcff01a3e464d13a46" -- --skip-cached --no-nom --flake .#legacyPackages.$(nix eval --raw --impure --expr builtins.currentSystem).$bin && sha256sum "./result/bin/$bin"
+            nix build .#$bin && sha256sum "./result/bin/$bin"
```
bradleystachurski, line 455: Is this step meant to have a test similar to the check versions step above? I believe this only prints the sha256 hash without comparing.

```
bin=fedimint-cli
nix build .#$bin && sha256sum "./result/bin/$bin"
20e9140fd21367f8b53f6a26f2a75bcf7191b5124058722c3f991e3e4e1b8e58  ./result/bin/fedimint-cli
```

bradleystachurski (approved):

.github/workflows/ci-nix.yml
dpc: It's only for printing, so one can take it from CI output and compare with a local build.

justinmoon (approved):

PR #4317, opened 2024-02-13
fix(ci): client-pkgs doesn't contain any binaries (v0.2)

dpc:
This is fixed already on `master`, and produces an empty deb&rpm release artifacts.

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -345,9 +345,6 @@ jobs:
     strategy:
       matrix:
         build:
-          - flake-output: client-pkgs
-            bins: ""
-            deb: fedimint
           - flake-output: fedimint-pkgs
             bins: fedimintd,fedimint-cli,fedimint-dbtool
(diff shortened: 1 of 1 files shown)


justinmoon (approved):

bradleystachurski:
Do we only require one approval to merge to `releases/v0.2`? If yes, do we want to bump to 2?

dpc:
> Do we only require one approval to merge to `releases/v0.2`? If yes, do we want to bump to 2?

It's only one. The idea is that v0.2 is mostly backports (already at least architecturally approved) and minor tweaks, so it doesn't need all that much attention.

bradleystachurski:
> > Do we only require one approval to merge to `releases/v0.2`? If yes, do we want to bump to 2?
> 
> It's only one. The idea is that v0.2 is mostly backports (already at least architecturally approved) and minor tweaks, so it doesn't need all that much attention.

Gotcha, thanks for the context 👍

Issue #4225, opened 2024-02-03
Devimint should save one invite code per guardian


justinmoon:
Currently devimint just save 1 invite code for the whole federation. But since invite codes currently only encode a single guardian url, but commit to the others, this invite code isn't usable if the guardian who issued this invite code is offline. This was just causing a test with one guardian offline to randomly fail for me.

https://github.com/fedimint/fedimint/blob/7fe0a203af3bd9683d4f856baa9c517ec9bad2f6/devimint/src/federation.rs#L171-L179

kernelkind:
Hi, this is my first issue so I'm going to describe my thought processes for how I plan on solving it.

It looks like this issue occurs because only one of the guardian's urls is saved in the invite code. Instead, all of the guardian's urls should be saved to the `InviteCode` struct. 

https://github.com/fedimint/fedimint/blob/06dcc978fa48d73ad61cbc48c6dafb42325149e7/fedimint-core/src/api.rs#L771C1-L776C6
should be modified to be:
```
new(api_endpoints: Vec<(PeerId, SafeUrl)>, federation_id: FederationId) -> Self
```
Then, each entry in `api_endpoints` can be saved as an `InviteCodeData::Api`.

 https://github.com/fedimint/fedimint/blob/06dcc978fa48d73ad61cbc48c6dafb42325149e7/fedimint-core/src/api.rs#L779C5-L779C33
 
becomes `InviteCode::urls(&self) -> Vec<SafeUrl>`

and

https://github.com/fedimint/fedimint/blob/06dcc978fa48d73ad61cbc48c6dafb42325149e7/fedimint-core/src/api.rs#L791C5-L791C33

becomes `InviteCode:peers(&self) -> Vec<PeerId>`

All of the usages of `InviteCode::new` would of course be modified to accommodate this new constructor, including
https://github.com/fedimint/fedimint/blob/06dcc978fa48d73ad61cbc48c6dafb42325149e7/fedimint-server/src/config/mod.rs#L262C1-L270C6

Instead of only including the current api endpoint, include all of them (all of the guardians). This method is used to write the invite code to the `invite-code` file.

Then, in `federation.rs`, since the invite code will contain the urls for all of the guardians, it's fine to just get the invite code from the first guardian since they will all contain the urls for all of the guardians.

elsirion:
@kernelkind this would make the invite codes quite large. I think what @justinmoon meant was saving all possible invited codes, such that a different one can be tried in tests.

[...]

PR #4318, opened 2024-02-13
feat: save invite code for each guardian

kernelkind:
Closes: https://github.com/fedimint/fedimint/issues/4225

The code compiles, but I didn't write any tests for it because I'm not sure where to put them. All of the tests in `tests.rs` seem like big integration tests and testing the small `Federation::invite_code_for` method along side the other big integration tests doesn't seem good.

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -147,7 +147,7 @@ impl Federation {
         servers: usize,
     ) -> Result<Self> {
         let mut members = BTreeMap::new();
-        let mut vars = BTreeMap::new();
+        let mut peer_to_env_vars_map = BTreeMap::new();
 
         let peers: Vec<_> = (0..servers).map(|id| PeerId::from(id as u16)).collect();
         let params: HashMap<PeerId, ConfigGenParams> = local_config_gen_params(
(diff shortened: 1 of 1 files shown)


devimint/src/federation.rs
```diff
@@ -170,18 +170,28 @@ impl Federation {
...
-        info!("moved client configs");
+        let cfg_dir: &str = utf8(cfg_dir);
+        let invite_code_filename_original = String::from("invite-code");
+        for (index, var) in &vars {
+            let out_dir: &PathBuf = &var.FM_DATA_DIR;
+            let out_dir = utf8(out_dir);
+            // move configs to config directory
+
+            let invite_code_filename_indexed = format!("invite-code-{}", index);
+            let invite_code_filename_new: String = if index == &0 {
```
m1sterc001guy, line 182: Is there a reason why we're special casing index 0? Seems to me the code here and below would be simpler if it just always expected the files to be named `invite-code-X`, even for the 0th peer

kernelkind: I definitely agree it would be simpler. My thought process was that it would avoid any backwards compatibility issues that could arise, and there are a few other places the `invite-code` file was used so I didn't want to cause any issues:
- `fedimint-server/src/config/io.rs`
- `devimint/src/bin/faucet.rs`
- `devimint/src/cli.rs`

If these are irrelevant, then I can happily make the change to simplify the code.

dpc: Make a copy in both places first, then as a follow-up we could get rid of places that use the old path.

justinmoon (approved):

devimint/src/federation.rs
```diff
@@ -170,17 +170,32 @@ impl Federation {
...
-        .context("moving invite code file")?;
+        .context("copying invite-code file")?;
+
+        // copy over each guardian's invite-code file to the client's directory
+        // appending the guardian id to the end
+        for (index, var) in &vars {
+            let out_dir = utf8(&var.FM_DATA_DIR);
+
+            let invite_code_filename_indexed = format!("{invite_code_filename_original}-{}", index);
+            tokio::fs::rename(
```
dpc, line 192: Comment says "copy", here is `rename`. I'm confused. :D

devimint/src/federation.rs
```diff
@@ -170,17 +170,32 @@ impl Federation {
...
 
-        let out_dir = &vars[&0].FM_DATA_DIR;
-        let cfg_dir = &process_mgr.globals.FM_CLIENT_DIR;
-        let out_dir = utf8(out_dir);
-        let cfg_dir = utf8(cfg_dir);
         // move configs to config directory
-        tokio::fs::rename(
-            format!("{out_dir}/invite-code"),
-            format!("{cfg_dir}/invite-code"),
+        let cfg_dir = utf8(&process_mgr.globals.FM_CLIENT_DIR);
```
dpc, line 174: We should stop using `cfg` for "client dir".

devimint/src/federation.rs
```diff
@@ -170,17 +170,32 @@ impl Federation {
...
+        let invite_code_filename_original = "invite-code";
+
+        // copy over invite-code file to client directory
+        let out_dir = utf8(&vars[&0].FM_DATA_DIR);
```
dpc, line 178: It's not something this change introduces but `vars` is confusing and not informative. :D

devimint/src/federation.rs
```diff
@@ -203,6 +218,13 @@ impl Federation {
         Ok(invite_code)
     }
 
+    pub fn invite_code_for(&self, peer_id: u16) -> Result<String> {
```
dpc, line 221: `self` not used?

kernelkind: Should `Federation::invite_code(&self) -> Result<String>` be changed to `Federation::invite_code() -> Result<String>` as well? This method is already being used and all of the places it is used would need to be changed since it would become an associated function instead of a method without `self`

kernelkind:
I've been amending the commit & force pushing, is that the desired workflow? Or should I make additional commits when there are comments to address?

m1sterc001guy (dismissed):

devimint/src/federation.rs
```diff
@@ -203,6 +225,13 @@ impl Federation {
         Ok(invite_code)
     }
 
+    pub fn invite_code_for(peer_id: u16) -> Result<String> {
```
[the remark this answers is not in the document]
m1sterc001guy: Sure, push a new commit and I'll re-approve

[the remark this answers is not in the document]
kernelkind: There was no need to call `into` as `PeerId` implements `Display`

m1sterc001guy (approved):

devimint/src/federation.rs
```diff
@@ -158,34 +158,56 @@ impl Federation {
...
-            format!("{out_dir}/invite-code"),
-            format!("{cfg_dir}/invite-code"),
+        let client_dir = utf8(&process_mgr.globals.FM_CLIENT_DIR);
+        let invite_code_filename_original = "invite-code";
+
+        // copy over invite-code file to client directory
+        let peer_data_dir = utf8(&peer_to_env_vars_map[&0].FM_DATA_DIR);
+        tokio::fs::copy(
+            format!("{peer_data_dir}/{invite_code_filename_original}"),
+            format!("{client_dir}/{invite_code_filename_original}"),
```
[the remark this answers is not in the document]
kernelkind: Yes, the `invite-code` file is identical to `invite-code-0`, and yeah, it was left so it didn't break code elsewhere which relies on the presence of `invite-code`

PR #4326, opened 2024-02-14
cleanup: remove invite code from client db

joschisan:
We can generate invite codes from the client config, the separate invite code in the db is a remnant from a time when we had to request an invite code from the guardian.

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -21,7 +21,7 @@ use fedimint_bip39::Bip39RootSecretStrategy;
 use fedimint_client::module::init::{ClientModuleInit, ClientModuleInitRegistry};
 use fedimint_client::module::ClientModule as _;
 use fedimint_client::secret::{get_default_client_secret, RootSecretStrategy};
-use fedimint_client::{get_invite_code_from_db, Client, ClientBuilder, ClientHandle};
+use fedimint_client::{Client, ClientBuilder, ClientHandle};
 use fedimint_core::admin_client::WsAdminClient;
 use fedimint_core::api::{
     FederationApiExt, FederationError, IRawFederationApi, InviteCode, WsFederationApi,
@@ -335,7 +335,9 @@ enum Command {
     Dev(DevCmd),
 
     /// Config enabling client to establish websocket connection to federation
-    InviteCode,
+    InviteCode {
+        peer: PeerId,
+    },
 
     /// Join a federation using it's InviteCode
     JoinFederation {
@@ -539,7 +541,6 @@ impl FedimintCli {
                     &client_config.global.calculate_federation_id(),
                 ),
                 client_config.clone(),
-                invite_code,
             )
             .await
             .map_err_cli_general()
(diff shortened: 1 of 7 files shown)


justinmoon:
Does this require a client migration? We'll be leaving behind a key in the database that may represent something else in the future.

justinmoon:
Now that https://github.com/fedimint/fedimint/issues/3417 is closed, we should be able to write this migration.

fedimint-client/src/db.rs
```diff
@@ -19,7 +19,7 @@ pub enum DbKeyPrefix {
     ChronologicalOperationLog = 0x2d,
     CommonApiVersionCache = 0x2e,
     ClientConfig = 0x2f,
-    ClientInviteCode = 0x30,
+    ClientInviteCode = 0x30, // Unused, we can get an invite code from the client config
```
elsirion, line 22: As long as this stays here future developers will know to clean out remnant data before re-using, but ideally we'd add a migration that removes all entries with that prefix, then we can also remove this enum variant.

fedimint-cli/src/lib.rs
```diff
@@ -602,19 +603,28 @@ impl FedimintCli {
...
+                    .ok_or_cli_msg(CliErrorKind::GeneralFailure, "client config code not found")?;
+
+                let invite_code = client_config
+                    .global
+                    .api_endpoints
+                    .get(&peer)
+                    .map(|peer_url| {
+                        InviteCode::new(peer_url.url.clone(), peer, client_config.federation_id())
+                    })
+                    .ok_or_cli_msg(CliErrorKind::GeneralFailure, "peer not found")?;
```
elsirion, line 626: If we merged #4297 first we wouldn't have to manually construct the invite here but could rather use `ClientConfig::invite_code(&Self, PeerId)`.

elsirion (commented):
Two nts, what do you think about merge order @joschisan?

fedimint-cli/src/lib.rs
joschisan: I always intended to rebase on https://github.com/fedimint/fedimint/pull/4297, that's why this is still a draft.

fedimint-client/src/db.rs
elsirion: @joschisan what are your thoughts about this?

joschisan: I added a comment about cleaning out remnants before reusing and would not do it for now as it would need a new migration hook for non-module db migration and maybe we get a chance in the future to toss it out with less effort, should we ever make a breaking change to the client.

fedimint-client/src/db.rs
```diff
@@ -35,7 +35,7 @@ pub enum DbKeyPrefix {
     ChronologicalOperationLog = 0x2d,
     CommonApiVersionCache = 0x2e,
     ClientConfig = 0x2f,
-    ClientInviteCode = 0x30,
+    ClientInviteCode = 0x30, // Unused; clean out remnant data before re-using!
```
dpc, line 38: We don't need a migration. We can just always delete this key on start, and over time it will get deleted from all clients and rotated out of existance.

dpc (approved):
LGTM, but @elsirion should take a look, also consider just deleting unconditionally to avoid any remnants whatsoever.

fedimint-client/src/db.rs
joschisan: I considered this, but we could never be sure could we?

m1sterc001guy: > We don't need a migration. We can just always delete this key on module start, and over time it will get deleted from all clients and rotated out of existence.

This PR has an example of how to do this https://github.com/fedimint/fedimint/pull/4427/files#diff-d754f9b1f77285e47d11bdecd4e3df652f846e5cedd12ded095ae84816a64cc2

dpc: > I considered this, but we could never be sure could we?

I mean - we would still keep `ClientInviteCode` variant in the code with annotation not to use it, but instead of not to the future self: "clean out remnant data before re-using!" we would just do it eagerly.


After long enough time for all practical purposes we'd be sure no supported client can possiblity still have it there.

I doubt it is going to matter in practice, but conceptually it's strictly better approach.

maan2003: `#[deprecated]` ?

joschisan: > I doubt it is going to matter in practice, but conceptually it's strictly better approach.

I agree with both, would prefer to get this merged now though.

elsirion: Deleting unconditionally is a hack imo and DB migrations are basically free to add.

If we eventually remove old migrations the client can refuse to start if a too old DB is encountered, but that only works if all DB migrations are modeled properly.

elsirion (approved):
Not a fan of the missing DB migration, but let's get this in.

PR #4327, opened 2024-02-14
Gatewayd migrations

m1sterc001guy:
Builds on https://github.com/fedimint/fedimint/pull/4282

Adds database snapshots for gatewayd and database migrations.


gateway/ln-gateway/src/db.rs
```diff
@@ -1,19 +1,23 @@
...
 use serde::{Deserialize, Serialize};
 use strum_macros::EnumIter;
 
+pub const GATEWAYD_DATABASE_VERSION: DatabaseVersion = DatabaseVersion(0);
+
 #[repr(u8)]
 #[derive(Clone, EnumIter, Debug)]
 pub enum DbKeyPrefix {
     FederationConfig = 0x04,
-    FederationRegistration = 0x05,
```
m1sterc001guy, line 16: This isn't used anywhere in the codebase and hasn't for sometime. I think it should be safe to remove.

elsirion: Was the key present in 0.2? If so I'd want a migration that deletes old data with that key prefix. I wouldn't bother to check probably and just add a migration, should be easy enough in this case.

elsirion (approved):

dpc (approved):

gateway/ln-gateway/src/db.rs
benthecarman: We're running 0.2.2, thanks for the call out though

elsirion: Great! Then I just misremembered :)

PR #4329, opened 2024-02-15
fix(memdb): waaay too slow, at least use rwlock

dpc:
After a lot of debuging I discovered that some flakes are due to all the slowness of `MemDatabase`.

On an even ligthly filled test database, coping all the data start taking considerable amount of time (single digit milliseconds), which combined with number of database transactions being opened and commited, can lead to quite long delays in processing in random places (as ordering of locking is somewhat arbitrary).

It's possible that something is also running db transactions in a tight loop somwhere.

For time being using an `RwLock` helps considerably, as all db transactions at least need to start, and we do two copied of the database content there, so being able to do it in parallel, significantly improves the throughput.

Having said that, current implementation is not practical for any production use, I'm afraid.

### DIFF
--- a/fedimint-core/src/db/mem_impl.rs
+++ b/fedimint-core/src/db/mem_impl.rs
@@ -5,7 +5,6 @@ use anyhow::Result;
 use bitcoin_hashes::hex::ToHex;
 use futures::{stream, StreamExt};
 use macro_rules_attribute::apply;
-use tokio::sync::Mutex;
 
 use super::{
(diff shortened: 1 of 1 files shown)


dpc:
#4335 even better, I hope.

joschisan (approved):

maan2003 (approved):

m1sterc001guy (approved):

elsirion:
Longer-term we might want a copy-on-write map type?

dpc:
#4335 @elsirion

PR #4330, opened 2024-02-15
chore: Notifier::notify improve logging

dpc:
Error on send here is normal and expected.

### DIFF
--- a/fedimint-client/src/sm/notifier.rs
+++ b/fedimint-client/src/sm/notifier.rs
@@ -42,9 +42,11 @@ impl Notifier {
         trace!(?state, %queue_len, "Sending notification about state transition");
         // FIXME: use more robust notification mechanism
         if let Err(e) = self.broadcast.send(state) {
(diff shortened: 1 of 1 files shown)


fedimint-client/src/sm/notifier.rs
```diff
@@ -42,9 +42,11 @@ impl Notifier {
         trace!(?state, %queue_len, "Sending notification about state transition");
         // FIXME: use more robust notification mechanism
         if let Err(e) = self.broadcast.send(state) {
-            error!(
+            debug!(
                 ?e,
-                "Could not send state transition notification, the client may misbehave"
+                %queue_len,
+                receivers=self.broadcast.receiver_count(),
```
dpc, line 48: Just to double-check. It's not atomic, so might actually be off, but :shrug:

m1sterc001guy (approved):

bradleystachurski (approved):

PR #4332, opened 2024-02-15
chore: better consensus item formatting

dpc:
Display txid, which is very useful for debuging.

Be lazy-evaluated which should give better perf when logging is not enabled and avoid extra allocation anyway.

### DIFF
--- a/fedimint-server/src/consensus/debug.rs
+++ b/fedimint-server/src/consensus/debug.rs
@@ -1,30 +1,41 @@
-use std::fmt::Write;
-
-use fedimint_core::transaction::Transaction;
+use std::fmt;
 
 use crate::ConsensusItem;
 
-pub fn item_message(item: &ConsensusItem) -> String {
-    match item {
-        // TODO: make this nice again
-        ConsensusItem::Module(mci) => {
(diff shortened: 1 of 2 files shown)


fedimint-server/src/consensus/debug.rs
```diff
@@ -1,30 +1,41 @@
...
-                write!(tx_debug, "\n    Input: {input}").unwrap();
+/// A newtype for a nice [`fmt::Debug`] of a [`ConsensusItem`]
+pub struct FmtDbgConsensusItem<'ci>(pub &'ci ConsensusItem);
+
+impl<'ci> fmt::Debug for FmtDbgConsensusItem<'ci> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.0 {
+            ConsensusItem::Module(mci) => {
+                f.write_fmt(format_args!(
+                    "Module CI: module={} ci={}",
```
m1sterc001guy, line 13: Maybe this could include the kind?

m1sterc001guy (approved):

justinmoon (approved):

PR #4333, opened 2024-02-15
chore: add highest-level logs to sends_ecash_out_of_band

dpc:
IMO we should be doing it in every test, so I'm going to show it around.

When you are debugging a tricky problem by analyzing thousands of logs lines of a failing test, having some reference points telling you where the highest level points of test execution happened is very, very useful.

Prefixing them with `###`, it makes jumping around and finding them easier.


m1sterc001guy (approved):
Ack, I'll start adding these to gateway test logs too

bradleystachurski (approved):
`## FAILED` has been super useful for searching, so I think this is great to highlight 👍

justinmoon:
I like this idea. It can be hard to tell what failed in CI runs. This should help.

PR #4338, opened 2024-02-15
Gateway has built-in LDK lightning node (and it generates invoices rather than the client)

tvolk131:
VERY broken, do not merge!


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -699,16 +731,74 @@ impl LightningClientModule {
...
+            bail!("Gateway created invoice with incorrect description");
+        }
+        if invoice.expiry_time() != expiry_time {
+            bail!(
+                "Gateway created invoice with incorrect expiry time: expected={}, got={}",
+                expiry_time.as_secs(),
+                invoice.expiry_time().as_secs()
+            );
+        }
+
```
benthecarman, line 790: need to verify correct payment hash

benthecarman, line 790: Need to verify same payment hash

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -699,16 +731,74 @@ impl LightningClientModule {
...
+            .await
+            .context("Failed to register payment hash with gateway")?;
+        if !response.status().is_success() {
+            bail!(
+                "Failed to register payment hash with gateway. Returned error code: {}",
+                response.status()
+            );
+        }
+
+        let invoice: Bolt11Invoice = response.json().await?;
```
benthecarman, line 767: probably makes sense to have this as a json object so it can be more extensible in the future

fedimint-testing/Cargo.toml
```diff
@@ -37,6 +37,7 @@ ln-gateway = { version = "0.3.0-alpha", package = "fedimint-ln-gateway", path =
 ldk-node = { package = "fedimint-ldk-node", version = "0.1.0" }
 futures = "0.3"
 lightning-invoice = "0.26.0"
+lightning-invoice29 = { package = "lightning-invoice", version = "0.29.0" }
```
benthecarman, line 40: ldk-node re-exports lightning invoice. Can probably just use that

tvolk131: Requires https://github.com/fedimint/fedimint/pull/4354 since we're currently using [`fedimint-ldk-node` v0.1.0](https://crates.io/crates/fedimint-ldk-node) which re-exports `lightning-invoice` v0.23.0 instead of v0.29.0. Leaving this open until that PR is merged.

modules/fedimint-ln-client/src/lib.rs
tvolk131: Whoops! Good catch, thanks!

tvolk131:
Closing since we merged #5554 and #5728 instead

PR #4342, opened 2024-02-16
demo: ldk node upgrade with tests against real nodes

bradleystachurski:


bradleystachurski:
Closing since the reference has been closed

https://github.com/fedimint/fedimint/pull/4354#issuecomment-1989472534

PR #4346, opened 2024-02-16
feat: Impl Eq & ParialEq for OOBNotes

benthecarman:


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -86,10 +86,10 @@ pub const LOG_TARGET: &str = "client::module::mint";
 /// ## Invariants
 /// * Has to contain at least one `Notes` item
 /// * Has to contain at least one `FederationIdPrefix` item
-#[derive(Clone, Debug, Encodable)]
+#[derive(Clone, Debug, Encodable, PartialEq, Eq)]
```
maan2003, line 89: shouldn't Eq be smarter?
i would only compare OOBNotesData::Notes

dpc: Why not compare other variants?

FWICT, `Invite::peer_apis` would be better if it had canonical ordering, but in practice I don't think it matters.

maan2003 (approved):

modules/fedimint-mint-client/src/lib.rs
elsirion: I think `Eq` should ignore ordering indeed, possibly by just sorting the entries if `Ord` is easy to derive. Semantically `[Notes, FederationIdPrefix]` and `[FederationIdPrefix, Notes]` are identical.

dpc: I'm not sure if we need semantic Eq, especially that it would require extra sort on every comparison, which I guess could be surprising.

elsirion: Ok, I just hope this never comes to bite us … (sort should be quite quick for such small data structures). Maybe it's not too late to introduce canonical encoding …

dpc: I guess sorting on serialization is a good first step, and after a while we can reject unsorted encodings.

... though looking at it RN, I'm not sure why we even have `PeerId` there in the first place. Aren't peer ids always starting from 0 and contiguous?

elsirion: Oh, I was talking about all the fields, not just the invite, we are dealing with a `Vec<OOBNotesData>` and that was my concern. Internally in `OOBNotesData::Invite` I see the problem too now. Why include the id: we likely never want to include all APIs and want to have the ability to skip a long-term offline "middle" guardian.

PR #4354, opened 2024-02-17
Upgrade to ldk-node 0.2.1

tvolk131:
Requires #4086

This cannot be merged because it adds `openssl` as a dependency. This is a dependency of `vss-client` which is added as a dependency to `ldk-node` in 0.2.0.

Perhaps we should make a branch in our [fork of ldk node](https://github.com/fedimint/ldk-node) which disables vss and use that instead of the official crate.


m1sterc001guy:
Is it possible to use the cfg flag that was introduced in ldk-node to disable vss? The whole point of upgrading for us is to eliminate the dependency on our forked repo

elsirion:
Opened https://github.com/lightningdevkit/ldk-node/issues/254 in `ldk-node`

tvolk131:
Thanks @elsirion and @tnull for ironing this out! Just rebased and bumped the version of `vss-client`. This PR is now just waiting on #4086, since this PR uses one of the helper functions added there (`bitcoin30_to_bitcoin29_public_key()`). If we don't expect that PR to be merged soon, I can add it separately in this PR so we can get this merged.

tvolk131:
#4086 is merged! This PR should be safe to submit.

fedimint-core/src/bitcoin_migration.rs
```diff
@@ -8,6 +8,11 @@ pub fn bitcoin29_to_bitcoin30_public_key(pk: bitcoin::PublicKey) -> bitcoin30::P
         .expect("Failed to convert bitcoin v29 public key to bitcoin v30 public key")
 }
 
+pub fn bitcoin30_to_bitcoin29_public_key(pk: bitcoin30::PublicKey) -> bitcoin::PublicKey {
+    bitcoin::PublicKey::from_slice(&pk.to_bytes())
+        .expect("Failed to convert bitcoin v30 public key to bitcoin v29 public key")
+}
+
```
elsirion, line 15: nit: ideally this code move wouldn't be here, likely a rebase artifact.

elsirion (approved):

fedimint-core/src/bitcoin_migration.rs
tvolk131: Not a rebase artifact - we're moving this out of the `tests` module now that we actually need it in prod code. Previously we were just calling it in a few tests.

fedimint-testing/src/ln/real.rs
```diff
@@ -415,12 +420,14 @@ impl LdkLightningTest {
...
-        );
+        builder
+            .set_listening_addresses(vec![SocketAddress::from_str(
+                format!("0.0.0.0:{unused_port}").as_str(),
+            )
+            .expect("Couldnt parse listening address")])
+            .expect("Couldnt set listening address");
         builder.set_storage_dir_path(db_path.to_string_lossy().to_string());
         let esplora_port = std::env::var("FM_PORT_ESPLORA").unwrap_or(String::from("50002"));
         builder.set_esplora_server(format!("http://127.0.0.1:{esplora_port}"));
```
bradleystachurski, line 433: Odd - when running the real tests locally for bitcoind, which now covers fedimint-ln-gateway tests, this passes. Out of scope for this PR, but perhaps we should run the ln-gateway backend tests against esplora instead of bitcoind?

bradleystachurski (commented):
I was kicking the tires locally and it looks like we aren't running real tests for `fedimint-ln-gateway`. These changes look good, but I think we should rebase after merging https://github.com/fedimint/fedimint/pull/4388 so these changes are tested in CI.

Locally, this diff looks good including the fix to run the real tests.

```
    Starting 1 test across 49 binaries (20 skipped)
        SLOW [> 60.000s] fedimint-ln-gateway::gatewayd-integration-tests test_gateway_can_pay_ldk_node
        PASS [  70.124s] fedimint-ln-gateway::gatewayd-integration-tests test_gateway_can_pay_ldk_node
------------
     Summary [  70.125s] 1 test run: 1 passed (1 slow), 20 skipped
```

elsirion:
I don't think we need to rebase, the updated CI should be used in the merge queue.

bradleystachurski:
> I don't think we need to rebase, the updated CI should be used in the merge queue.

Ah yep, good call 👍 

We should remove the `#[ignore]` in `test_gateway_can_pay_ldk_node`
https://github.com/fedimint/fedimint/pull/4388/files#diff-b1fe1eec690ca593c767be467f8b226e8335b51733d649b91e2fa23e4ba774d5R204

@tvolk131 I heard you have a busy week with limited availability. @elsirion do you know if there's a way to add a commit to this PR that removes the `#[ignore]`? If not, I can open a separate PR that keeps Tommy's commit's along with a commit that removes the ignore.

elsirion:
You may need 3c2491f6a9456575d11cbb10b9e6e406476c46a9, seems like @bradleystachurski ran into a timeout.

elsirion (dismissed):

bradleystachurski:
> I just rebased and removed the ignore!

Beautiful, thanks!

> You may need [3c2491f](https://github.com/fedimint/fedimint/commit/3c2491f6a9456575d11cbb10b9e6e406476c46a9), seems like @bradleystachurski ran into a timeout.

Yep, I can tackle running `test_gateway_can_pay_ldk_node` in parallel in a followup to keep CI running quickly.

m1sterc001guy:
On second thought, I'd prefer to just work towards supporting `ldk-node` in the gateway, rather than also supporting it in `fedimint-testing`. I opened a PR to remove it https://github.com/fedimint/fedimint/pull/4521

If we're cool with that, I think we can close this.

m1sterc001guy:
Closing in favor of https://github.com/fedimint/fedimint/pull/4521

PR #4356, opened 2024-02-18
fix: remove use of system time in ln client

benthecarman:
Calling `.elapsed` will call `SystemTime::now()` under the hood. The semgrep did not catch this sadly.

Looks like `elapsed` is used other places but just for tests and server side stuff

This is present in 0.2.2 so should be backported as well

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -32,6 +32,7 @@ use fedimint_core::module::{
     ApiVersion, CommonModuleInit, ModuleCommon, ModuleInit, MultiApiVersion, TransactionItemAmount,
 };
 use fedimint_core::task::{timeout, MaybeSend, MaybeSync};
+use fedimint_core::time::now;
 use fedimint_core::{
     apply, async_trait_maybe_send, push_db_pair_items, Amount, OutPoint, TransactionId,
 };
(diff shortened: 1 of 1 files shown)


maan2003 (approved):

benthecarman:
fixed rebase conflicts

maan2003 (approved):

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -908,7 +909,8 @@ impl LightningClientModule {
             .get_value(&MetaOverridesKey {})
             .await
         {
-            if cache.fetched_at.elapsed().unwrap() < META_OVERRIDE_CACHE_DURATION {
+            let elapsed = now().duration_since(cache.fetched_at).unwrap();
```
bradleystachurski, line 912: [nit] Since this function returns a `Result`, might as well convert the `unwrap`

```suggestion
            let elapsed = now().duration_since(cache.fetched_at)?;
```

bradleystachurski (approved):

justinmoon:
Why didn’t semgrap catch this?

dpc:
If we don't want this to repeat we should add semgrep rule for `elapsed` and annotate existing usages that are OK to ignore them by semgrep.

benthecarman:
looks like the merge failed

justinmoon:
retrying

PR #4357, opened 2024-02-18
chore: remove unneeded usage of `pub` throughout gateway code

tvolk131:


gateway/ln-gateway/src/lib.rs
```diff
@@ -1165,7 +1165,7 @@ impl Gateway {
     /// - `GatewayConfiguration` is read from the database.
     /// - All cli or environment variables are set such that we can create a
     ///   `GatewayConfiguration`
-    pub async fn get_gateway_configuration(&self) -> Option<GatewayConfiguration> {
+    async fn get_gateway_configuration(&self) -> Option<GatewayConfiguration> {
```
elsirion, line 1168: Is this something you might want to call if you instantiate the gateway yourself as part of a custom binary?

gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -240,7 +239,7 @@ impl ClientModule for GatewayClientModule {
 }
 
 impl GatewayClientModule {
-    pub fn to_gateway_registration_info(
+    fn to_gateway_registration_info(
```
elsirion, line 242: Maybe useful for integrators

elsirion (approved):
I think in this case it's ok to reduce visibility, but @douglaz made a good point yesterday that making things pub again when needed is a lengthy process. So maybe for now we only want to make things private where it's necessary for guaranteeing invariants.

elsirion:
@douglaz do you think this PR is worth it? If so it needs a rebase.

tvolk131:
Rebased!

elsirion (approved):

gateway/ln-gateway/src/lib.rs
m1sterc001guy: I'd say yes, but right now I don't see us supporting custom binaries for gateways

m1sterc001guy (approved):
Send it

PR #4361, opened 2024-02-19
chore: add single peer request strategy

m1sterc001guy:
For the module I am working on, I have the need to submit an authenticated API request to a single peer. In this context, the client is authenticated with a single guardian, but should not have authenticated access to other guardians. For my module, this instructs the guardian to submit a consensus item.

Right now I don't think this is possible to do since `request_with_strategy` by default contacts all guardians.

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -250,6 +250,30 @@ pub struct ApiVersionSet {
 /// [`IRawFederationApi`].
 #[apply(async_trait_maybe_send!)]
 pub trait FederationApiExt: IRawFederationApi {
+    /// Make a request to a single peer in the federation with an optional
+    /// timeout.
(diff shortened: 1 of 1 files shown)


m1sterc001guy:
> Could we replace [WsAdminClient](https://github.com/fedimint/fedimint/blob/f58e5f96827871db74461c96c3b0d56890805a35/fedimint-core/src/admin_client.rs#L25) with this eventually?

Hmm it appears that `WsAdminClient` uses `request_current_consensus` underneath, which I'm a little confused about, since I thought authenticated APIs should only request a single guardian. I must be missing something about how `WsAdminClient` is used, maybe someone else can comment

elsirion:
I think it's always initialized with exactly one peer/API endpoint :see_no_evil:

dpc (approved):

PR #4366, opened 2024-02-19
feat: remove gateway from federation

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/4027

Adds a challenge-response protocol for removing gateway registrations from federations.

Challenge is the sha256 hash of <gateway-id><valid-until> bytes. This ensures that the challenge changes every time the gateway is re-registered. It also ensures that each guardian has its own challenge.

Gateways since with their private key (public key is the `gateway_id`).


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -943,14 +953,39 @@ impl ServerModule for Lightning {
                     Ok(())
                 }
             },
+            api_endpoint! {
+                REMOVE_GATEWAY_CHALLENGE_ENDPOINT,
+                ApiVersion::new(0, 0),
```
elsirion, line 958: This isn't true, will only be available on next minor API version (also need to bump it)

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1173,6 +1208,65 @@ impl Lightning {
...
+    ) -> Option<sha256::Hash> {
+        if let Some(gateway) = dbtx.get_value(&LightningGatewayKey(gateway_id)).await {
+            let mut valid_until_bytes = gateway.valid_until.to_bytes();
+            let mut challenge_bytes = gateway_id.to_bytes();
+            challenge_bytes.append(&mut valid_until_bytes);
+            Some(sha256::Hash::hash(&challenge_bytes))
+        } else {
+            None
+        }
+    }
```
elsirion, line 1231: Very elegant replay prevetion!

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1173,6 +1208,65 @@ impl Lightning {
...
+        let gateway_id = remove_gateway_request.gateway_id;
+        let our_peer_id = self.our_peer_id;
+        let signature = remove_gateway_request.signatures.get(&our_peer_id);
+
+        if signature.is_none() {
+            warn!("No signature provided for gateway: {gateway_id}");
+            return Err(anyhow::anyhow!(
+                "No signature provided for gateway {gateway_id}"
+            ));
+        }
```
elsirion, line 1250: This could be an [`.ok_or_else(|| {…})`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else):

```rust
let signature = remove_gateway_request.signatures.get(&our_peer_id)
    .ok_or_else(|| {
        warn!("No signature provided for gateway: {gateway_id}");
        anyhow::anyhow!(
            "No signature provided for gateway {gateway_id}"
        );
    })?;
```

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1173,6 +1208,65 @@ impl Lightning {
...
+
+        let signature = signature.expect("Already checked for none");
```
elsirion, line 1252: With `ok_or_else` and `?` this wouldn't be necessary anymore

modules/fedimint-ln-server/src/lib.rs
```diff
@@ -1173,6 +1208,65 @@ impl Lightning {
...
+            signature.verify(&msg, &gateway_id.x_only_public_key().0)?;
+
+            dbtx.remove_entry(&LightningGatewayKey(gateway_id)).await;
+            info!("Successfully removed gateway: {gateway_id}");
+            return Ok(());
+        }
+
+        Err(anyhow::anyhow!(
+            "Gateway {gateway_id} is not registered with peer {our_peer_id}"
+        ))
```
elsirion, line 1268: The same `ok_or_else` trick would allow returning early here and might make the code slightly easier to follow.

modules/fedimint-ln-common/src/api.rs
```diff
@@ -220,6 +290,16 @@ where
...
 }
 
+/// Request sent to the federation that requests the removal of a gateway
+/// registration. Each peer is expected to check the `signatures` map for the
+/// signature that validates the gateway authorized the removal of this
+/// registration.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RemoveGatewayRequest {
+    pub gateway_id: PublicKey,
+    pub signatures: BTreeMap<PeerId, Signature>,
```
elsirion, line 300: Not ideal that we have to send all the signatures to all guardians, but not something worth spending time on optimizing.

gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -381,6 +383,52 @@ impl GatewayClientModule {
...
+        let gateway_id = gateway_keypair.public_key();
+        let challenges = self
+            .module_api
+            .get_remove_gateway_challenge(gateway_id)
+            .await?;
+
+        let mut signatures: BTreeMap<PeerId, Signature> = BTreeMap::new();
+        for (peer_id, challenge) in challenges {
+            if let Some(challenge) = challenge {
+                let msg = Message::from_slice(&challenge)?;
```
elsirion, line 414: Security bug: We should not use the guardian-provided challenge directly, this would allow malicious guardians to make us sign anything. This isn't super bad since the gateway key isn't used for anything but registrations right now, but at least malicious guardians could de-register us from other federations.

We should build our actual signed message this way:

```
tag = "remove_request";
m = hash(tag + federation_id + guardian_id + challenge)
```

This way the message is only valid for the specific guardian and the use case of de-registering.

elsirion (changes requested):
Some nits, one security concern

modules/fedimint-ln-server/src/lib.rs
m1sterc001guy: Bumped the `ApiVersion` here and above in `supported_api_version`

modules/fedimint-ln-common/src/api.rs
m1sterc001guy: Agreed :/

gateway/ln-gateway/src/state_machine/mod.rs
m1sterc001guy: Good catch! Fixed. I ended up using the Federation's public key instead of the federation id though, since it was easier to get inside the server module.

gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -381,6 +383,53 @@ impl GatewayClientModule {
...
+
+        let fed_public_key = self.cfg.threshold_pub_key;
+        let mut signatures: BTreeMap<PeerId, Signature> = BTreeMap::new();
+        for (peer_id, challenge) in challenges {
+            if let Some(challenge) = challenge {
+                let msg = create_gateway_remove_message(fed_public_key, peer_id, challenge);
+                let signature = gateway_keypair.sign_schnorr(msg);
+                signatures.insert(peer_id, signature);
+            }
+        }
```
elsirion, line 419: nit: could be a iterator:

```rust
let signatures = challenges.into_iter()
    .filter_map(|(peer_id, challange)| {
        let msg = create_gateway_remove_message(fed_public_key, peer_id, challenge?);
        let signature = gateway_keypair.sign_schnorr(msg);
        Some((peer_id, signature))
    })
    .collect::<BTreeMap<_, _>>()
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -1431,6 +1440,21 @@ impl Gateway {
             _ => Err(LightningRpcError::FailedToConnect),
         }
     }
+
+    /// Iterates through all of the federations the gateway is registered with
+    /// and requests to remove the registration record.
+    pub async fn leave_all_federations(&self) {
+        let mut dbtx = self.gateway_db.begin_transaction_nc().await;
+        if let Some(keypair) = dbtx.get_value(&GatewayPublicKey).await {
```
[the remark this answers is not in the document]
m1sterc001guy: Yes this can be changed to an `expect`

elsirion (approved):

gateway/ln-gateway/src/lib.rs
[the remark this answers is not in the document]
m1sterc001guy: Apologies, I thought I fixed this. Should be done now.

dpc (approved):

elsirion (approved):

PR #4368, opened 2024-02-20
fix: make sure blinding key is not zero or one

benthecarman:
This should be basically impossible but important to check. If the key is 0 it'll cause a panic during execution and if it is one it won't actually blind the message.

Maybe these checks should be done in `blind_message` as well


benthecarman:
> Do we *really* need to guard against a 1/2^256 chance event? This is as likely to occur as someone randomly guessing Bitcoin private keys …

I had same thought after I opened this, fine to close, not a real threat

PR #4369, opened 2024-02-20
Add semgrep rule for `.elapsed()`

bradleystachurski:
Followup to prevent further bugs similar to https://github.com/fedimint/fedimint/pull/4356

This rule catches usages of method calls to `.elapsed()`. I did an audit of all public functions for `std::time::SystemTime` and this is the only one that calls `std::time::SystemTime::now`.

Semgrep supports typed metavariables ([docs](https://semgrep.dev/docs/writing-rules/pattern-syntax/#using-typed-metavariables)), which would be much better than catching any usage of `.elapsed()`, however after testing locally it doesn't catch the usage in the associated PR. Rust support is beta and I believe it's limited in its ability to parse types from struct fields. We're currently using semgrep v1.37.0, so I installed the latest version and that doesn't work either.

Semgrep PR introducing typed metavariables for rust: https://github.com/semgrep/semgrep/pull/8201

### DIFF
--- a/.config/semgrep.yaml
+++ b/.config/semgrep.yaml
@@ -8,6 +8,21 @@ rules:
 
   severity: WARNING
 
+- id: ban-system-time-elapsed
+  languages:
+    - rust
+  message: Use fedimint_core::time::now and std::time::duration_since to compare an elapsed time for better wasm compatibility.
+  pattern: $SYSTEM_TIME.elapsed()
+  paths:
+    exclude:
+      # server only
+      - fedimint-server/
+      # doesn't run in wasm
+      - fedimint-load-test-tool/
+      # doesn't run in wasm
+      - devimint/
+  severity: WARNING
+
 - id: ban-instant-now
   languages:
     - rust
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -909,7 +909,7 @@ impl LightningClientModule {
             .get_value(&MetaOverridesKey {})
             .await
         {
-            let elapsed = now().duration_since(cache.fetched_at).unwrap();
+            let elapsed = now().duration_since(cache.fetched_at).unwrap_or_default();
             if elapsed < META_OVERRIDE_CACHE_DURATION {
                 debug!("Using cached meta overrides");
                 match serde_json::from_str(&cache.value) {


.config/semgrep.yaml
```diff
@@ -8,6 +8,21 @@ rules:
 
   severity: WARNING
 
+- id: ban-system-time-elapsed
+  languages:
+    - rust
+  message: Use fedimint_core::time::now and std::time::duration_since to compare an elapsed time for better wasm compatibility.
+  pattern: $SYSTEM_TIME.elapsed()
```
dpc, line 15: I was hoping:

```
- id: ban-system-time-elapsed
  languages:
    - rust
  message: TBD
  pattern-either:
    - pattern: std::time::SystemTime::elapsed
    - pattern: std::time::Instant::elapsed

  severity: WARNING
```

would work, but it seems it doesn't.

dpc (approved):

.config/semgrep.yaml
bradleystachurski: Yeah, I was hoping the same and came to the same conclusion for method calls via `$foo.bar`.

It's also worth noting a footgun:

```
pattern: '($SYSTEM_TIME: std::time::SystemTime).elapsed()'
```

This is valid syntax for typed metavariables and implies it will only match method calls to `SystemTime.elapsed()`, however semgrep is unable to parse the type from struct fields so we could accidentally sneak in additional calls to `.elapsed()`.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -909,7 +909,7 @@ impl LightningClientModule {
             .get_value(&MetaOverridesKey {})
             .await
         {
-            let elapsed = now().duration_since(cache.fetched_at).unwrap();
+            let elapsed = now().duration_since(cache.fetched_at)?;
```
elsirion, line 912: Can we make this an `unwrap_or_default` to stay in sync with 0.2.3? See #4370

elsirion (approved):

justinmoon:
What about just banning use of `std::time` in all our client module crates? Anything time-related that you want to use needs to come from `fedimint-core` which should always have WASM-compatible implementations?

elsirion:
> What about just banning use of `std::time` in all our client module crates? Anything time-related that you want to use needs to come from `fedimint-core` which should always have WASM-compatible implementations?

I think that's a great idea!

bradleystachurski:
> What about just banning use of `std::time` in all our client module crates?

Shoot, I like the idea but semgrep doesn't catch the import. This is a draft PR with a global ban of `std::time` but semgrep passes. Looks like there are known limitations for `use` declarations (https://github.com/semgrep/semgrep/issues/6593).

https://github.com/fedimint/fedimint/pull/4386

dpc (approved):

PR #4370, opened 2024-02-20
fix: remove use of system time in ln client

### LINKED ISSUE #4363: Backport PR `#4356`
The backport bot was unable to open a pull request to backport changes from PR #4356. Please open a PR to backport these changes.

benthecarman:
Closes #4363

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -34,6 +34,7 @@ use fedimint_core::module::{
     ApiVersion, CommonModuleInit, ModuleCommon, ModuleInit, MultiApiVersion, TransactionItemAmount,
 };
 use fedimint_core::task::{timeout, MaybeSend, MaybeSync};
+use fedimint_core::time::now;
 use fedimint_core::{
     apply, async_trait_maybe_send, push_db_pair_items, Amount, OutPoint, TransactionId,
 };
@@ -903,7 +904,8 @@ impl LightningClientModule {
             .get_value(&MetaOverridesKey {})
             .await
         {
-            if meta.fetched_at.elapsed().unwrap() < META_OVERRIDE_CACHE_DURATION {
+            let elapsed = now().duration_since(meta.fetched_at).unwrap_or_default();
+            if elapsed < META_OVERRIDE_CACHE_DURATION {
                 debug!("Using cached meta overrides");
                 return Ok(meta.value);
             }


dpc (dismissed):

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -903,7 +904,8 @@ impl LightningClientModule {
             .get_value(&MetaOverridesKey {})
             .await
         {
-            if meta.fetched_at.elapsed().unwrap() < META_OVERRIDE_CACHE_DURATION {
+            let elapsed = now().duration_since(meta.fetched_at).unwrap();
```
dpc, line 907: BTW. We should have probably write a either a free-standing `elapsed` function in `fedimint_core` (just like `now`), or even use an extension trait.

m1sterc001guy, line 907: Let's remove the `unwrap` here too and replace with `expect`

benthecarman: done

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -903,7 +904,10 @@ impl LightningClientModule {
             .get_value(&MetaOverridesKey {})
             .await
         {
-            if meta.fetched_at.elapsed().unwrap() < META_OVERRIDE_CACHE_DURATION {
+            let elapsed = now()
+                .duration_since(meta.fetched_at)
+                .expect("time is always increasing");
```
elsirion, line 909: That is not technically correct for `SystemTime`:

Screenshot 2024-02-20 at 20-05-21 SystemTime in std time - Rust

Maybe an `unwrap_or` would be better?

benthecarman: Did `unwrap_or(0)` so it'll just refetch

dpc (approved):

modules/fedimint-ln-client/src/lib.rs
bradleystachurski: Good point 👍 

Tracking in https://github.com/fedimint/fedimint/issues/4383

modules/fedimint-ln-client/src/lib.rs
bradleystachurski: I believe `unwrap_or(0)` (or `unwrap_or_default` in merged diff) will return from the cache instead of refetching. I think either works, but want to highlight incase you specifically wanted the behavior to refetch.

Issue #4306, opened 2024-02-12
Invite codes should be able to encode arbitrary number of guardian URLs


justinmoon:
Previously it encoded enough to work if the federation was operational. Currently it just encodes 1. It should allow developer to choose either option.

PR #4371, opened 2024-02-20
feat: generate essential num guardians InviteCode

kernelkind:
The option is now available to generate an InviteCode with max_evil + 1 number of guardian urls in case some guardians go offline.

Closes: https://github.com/fedimint/fedimint/issues/4306

### DIFF
--- a/fedimint-core/src/api.rs
+++ b/fedimint-core/src/api.rs
@@ -870,6 +870,26 @@ impl InviteCode {
         ])
     }
 
+    /// Constructs an [`InviteCode`] which contains as many guardian URLs as
+    /// needed to always be able to join a working federation
+    pub fn new_with_essential_num_guardians(
+        peer_to_url_map: &BTreeMap<PeerId, SafeUrl>,
+        federation_id: FederationId,
+    ) -> Self {
+        let max_size = peer_to_url_map.max_evil() + 1;
+        let mut code_vec: Vec<InviteCodeData> = peer_to_url_map
+            .iter()
+            .take(max_size)
+            .map(|(peer, url)| InviteCodeData::Api {
+                url: url.clone(),
+                peer: *peer,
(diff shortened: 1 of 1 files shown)


kernelkind:
I don't think the build pipeline fail is valid, can we re-run it?

fedimint-core/src/api.rs
```diff
@@ -772,6 +772,21 @@ impl InviteCode {
         ])
     }
 
+    pub fn new_max_guardians_code(
```
elsirion, line 775: 1. Please document what the fn does
2. Maybe there's a better name, we aren't including the ma number of guardian URLs, but rather the minimum required to always be able to join a working federation

fedimint-core/src/api.rs
```diff
@@ -772,6 +772,21 @@ impl InviteCode {
...
+        peer_to_url_map: BTreeMap<PeerId, SafeUrl>,
```
elsirion, line 776: Maybe we can make this a reference? If we don't consume all entries we don't need to clone all of them, but could rather call `clone` inside the `map` statement.

fedimint-core/src/api.rs
```diff
@@ -870,6 +870,26 @@ impl InviteCode {
         ])
     }
 
+    // Constructs an InviteCode which contains as many guardian URLs as needed to
```
elsirion, line 873: Docstrings use three slashes, otherwise they don't render in docs.rs. It's also nice to use code formatting for struct names in docs.

elsirion (commented):
Looks good, just the docstring needs fixing

elsirion (approved):

m1sterc001guy (approved):

justinmoon:
Nice PR @kernelkind! Thanks for contribution.

PR #4375, opened 2024-02-20
refactor(nix): split-out overlays into own files

dpc:


elsirion (approved):
I don't fully understand it, but looks like way more idiomatic usage of flakes.

maan2003 (approved):

PR #4376, opened 2024-02-20
chore: use sleep_in_test over sleep in tests

kernelkind:
The function fedimint_core::task::sleep_in_test should be used over fedimint_core::task::sleep in tests so the duration of the sleep and a comment describing the reason for sleeping can be logged.

Closes: https://github.com/fedimint/fedimint/issues/1792

### DIFF
--- a/fedimint-core/src/task.rs
+++ b/fedimint-core/src/task.rs
@@ -8,7 +8,7 @@ use std::time::{Duration, Instant, SystemTime};
 
 use anyhow::bail;
 use fedimint_core::time::now;
-use fedimint_logging::LOG_TASK;
+use fedimint_logging::{LOG_TASK, LOG_TEST};
 #[cfg(target_family = "wasm")]
 use futures::channel::oneshot;
 use futures::lock::Mutex;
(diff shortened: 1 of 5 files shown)


fedimint-core/src/task.rs
```diff
@@ -617,6 +617,17 @@ impl<T: Sync> MaybeSync for T {}
 #[cfg(target_family = "wasm")]
 impl<T> MaybeSync for T {}
 
+pub async fn sleep_in_test(comment: impl AsRef<str>, duration: Duration) {
```
dpc, line 620: Comment explaining what do we use it for, why, etc.

fedimint-core/src/task.rs
```diff
@@ -617,6 +617,19 @@ impl<T: Sync> MaybeSync for T {}
 #[cfg(target_family = "wasm")]
 impl<T> MaybeSync for T {}
 
+// Used in tests when sleep functionality is desired so it can be logged.
+// Must include comment describing the reason for sleeping.
+pub async fn sleep_in_test(comment: impl AsRef<str>, duration: Duration) {
```
maan2003, line 622: might be to use `tracing::Span` instead of str comment

dpc: Shouldn't `comment` be `reason`, and just another value with a fixed `"Sleep"` message? Or are there any other benefits? I'm not even sure how to do `Span` :D

maan2003: Span just lets you express variables better instead of using format!. also spans have parent-child relation.

something like this 
```
async move {
    info!(target: LOG_TEST, ?duration, "sleeping");
    sleep(duration).await;
}.instrument(span).await
```

maan2003: caller is expected to use `info_span!` or `debug_span!` to create spans

elsirion: Is it useful to create a new span just for a sleep? I guess if you already have one it's nice, but what would one gain from adding it here otherwise?

elsirion (approved):

PR #4377, opened 2024-02-20
refactor: remove register_with_federation_inner

m1sterc001guy:
Since we removed the client extension traits we no longer need to use the inner pattern here.


gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -366,19 +352,14 @@ impl GatewayClientModule {
...
+        let gateway_id = registration_info.info.gateway_id;
 
-            let federation_id = self.client_ctx.get_config().global.federation_id();
-            self.register_with_federation_inner(federation_id, registration_info)
-                .await?;
-            Ok(())
-        }
+        let federation_id = self.client_ctx.get_config().global.federation_id();
+        self.module_api.register_gateway(&registration_info).await?;
+        debug!("Successfully registered gateway {gateway_id} with federation {federation_id}");
```
elsirion, line 361: nit: in the future we can use such refactors to move away from string interpolation to … whatever we call this syntax:

```rust
debug!(%gateway_id, %federation_id, "Successfully registered gateway with federation");
```

elsirion (approved):

justinmoon (approved):

PR #4381, opened 2024-02-21
fix: flaky latency test

dpc:
The failing assertion compare max latency vs p90. But after improvements to client sm executor all the averages, mins, median and p90 went down.

I don't see why some rare max timing would neccessarily need to be within any range compared to p90. Especially given how many tests we're running in parallel on a VM, etc.


bradleystachurski (approved):

maan2003 (approved):

bradleystachurski:
Our old friend 😄 

```
TIMEOUT [ 480.038s] fedimint-mint-tests::fedimint_mint_tests sends_ecash_out_of_band_cancel
```

PR #4386, opened 2024-02-21
demo: semgrep fails to identify std::time import

bradleystachurski:
Semgrep has limitations identifying certain imports and `use std::time` isn't identified.

https://github.com/semgrep/semgrep/issues/6593


bradleystachurski:
This was intended as a demo and the associated PR was merged.

PR #4389, opened 2024-02-22
 chore(backward-compat-test): test each version of every component once 

dpc:
We only really need to test that current version works with every other component in every past version.

While it's possible there are bugs that only happen in certain combinations of old versions on multiple components, it's extremely unlikely, and worth the time it takes to test it.

So:

```
v0.2.1       v0.2.1          current          0
v0.2.1       current         v0.2.1           0
v0.2.1       current         current          0
current      v0.2.1          v0.2.1           0
current      v0.2.1          current          0
current      current         v0.2.1           0
```

becomes:

```
v0.2.1       current         current          0
current      v0.2.1          current          0
current      current         v0.2.1           0
```

and with even more versions the complexity changes from `num_versions^3` to `num_versions*3`

### DIFF
--- a/scripts/tests/backwards-compatibility-test.sh
+++ b/scripts/tests/backwards-compatibility-test.sh
@@ -103,7 +103,7 @@ for fed_version in "${versions[@]}"; do
       fi
 
       # cleair=$(dirname "$(mktemp -u)")
-      rm -rf "$tmpdir"/devimint-*
+      rm -rf "${TMPDIR:-/tmp}"/devimint-*
 
       >&2 echo "========== Finished backwards-compatibility run ==========="
       >&2 echo "fed version: $fed_version"
--- a/scripts/tests/backwards-compatibility-test.sh
+++ b/scripts/tests/backwards-compatibility-test.sh
@@ -57,12 +57,30 @@ function use_gateway_binaries_for_version() {
 test_results="fed_version,client_version,gateway_version,exit_code\n"
 has_failure=false
 versions+=("current")
+
+# count number of times the first argument appears in rest of the arguments
+function filter_count() {
+  item="$1"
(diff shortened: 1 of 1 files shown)


maan2003 (approved):
smart, this should catch most of the incompats

scripts/tests/backwards-compatibility-test.sh
```diff
@@ -84,9 +102,8 @@ for fed_version in "${versions[@]}"; do
         has_failure=true
       fi
 
-      # cleanup devimint dir
-      tmpdir=$(dirname "$(mktemp -u)")
-      rm -rf "$tmpdir"/devimint-*
+      # cleair=$(dirname "$(mktemp -u)")
```
bradleystachurski, line 105: [nit] Comment got a bit mangled

bradleystachurski (approved):
> While it's possible there are bugs that only happen in certain combinations of old versions on multiple components, it's extremely unlikely

Agreed, I'm comfortable with this tradeoff.

Should we consider running the full matrix prior to cutting a release?

scripts/tests/backwards-compatibility-test.sh
dpc: Oh. I messed up. We can fix it in post. :D

dpc:
> Should we consider running the full matrix prior to cutting a release?


Yes, I'd like to add some flag like that later down the line. Eg. instead of `== 2` we could do `1` or `2` etc. to increase number of combinations.

dpc:
Flaked in the MQ on something weird. https://github.com/fedimint/fedimint/issues/4393

elsirion:
> Should we consider running the full matrix prior to cutting a release?

I'd like to be able to do this

PR #4391, opened 2024-02-22
chore(devimint): print out stderr of commands we're running

dpc:
I find logs of commands used by devimint as useful (if not more) than the logs of devimint itself.


EthnTuttle (approved):

dpc:
@bradleystachurski We should not need to parse stderr of commands, at least commonly. If you really do need it, you could just not use the macro and the method under the hood, and configure `tokio::process::Command` ad-hoc. Or possibly we could have another macro and/or function for it.

bradleystachurski:
@dpc Sounds good, I should have a PR up soon with an approach that lets us eat our stderr and have it too.

Issue #3548, opened 2023-11-07
`fedimint-cli config` crashes if there are unknown modules


justinmoon:
This is a connection string to a mutinynet federation running 0.1 release containing 2 custom modules: `fed115fjxjp793mdynuq78ju724wxelmhj336pqw845d8s4a5jfm75a3d8hc8hwmg4nmskmrpkzw5cqk6yqpfwaehxw309askcurgvyhx6at5d9h8jmn9wsknqv3wv3jhvtnxv4jxjcn5vvhxxmmd97a5ttq64kahlwtahn07zr64k2c`

If you run `fedimint-cli config` after joining this federation, you get the following:

```
2023-11-07T20:59:30.109474Z  INFO fedimint_client: Detected configuration for unsupported module id: 3, kind: fedi-social
2023-11-07T20:59:30.109554Z  INFO fedimint_client: Detected configuration for unsupported module id: 4, kind: stability_pool
2023-11-07T20:59:30.162592Z  WARN fedimint_client: Module kind fedi-social of instance 3 not found in module gens, skipping
2023-11-07T20:59:30.162613Z  WARN fedimint_client: Module kind stability_pool of instance 4 not found in module gens, skipping
2023-11-07T20:59:30.163068Z  INFO fedimint_client::sm::executor: Starting state machine executor task
thread 'main' panicked at /Users/justin/work/fedimint/fedimint-core/src/encoding/mod.rs:734:17:
Expected decoded value. Possibly `redecode_raw` call is missing.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2023-11-07T20:59:30.164639Z  INFO fedimint_client: Last client reference dropped, shutting down client task group
2023-11-07T20:59:30.164728Z  INFO fedimint_client::sm::executor: Shutting down state machine executor runner due to shutdown signal
```

justinmoon:
Right after observing this, I tried to spend some ecash and observed this:

```
$ cargo run --bin fedimint-cli -- --data-dir federations/mutinynet-02 spend 1
    Finished dev [unoptimized + debuginfo] target(s) in 0.44s
     Running `target/debug/fedimint-cli --data-dir federations/mutinynet-02 spend 1`
2023-11-07T21:03:15.559629Z  INFO fedimint_client: Detected configuration for unsupported module id: 3, kind: fedi-social
2023-11-07T21:03:15.559700Z  INFO fedimint_client: Detected configuration for unsupported module id: 4, kind: stability_pool
[...]

PR #4399, opened 2024-02-22
feat: `unknown` module

dpc:
We would like to make sure all client code can gracefully handle Federations with modules that the client doesn't support.

The laziest way to do it is to add a new dummy module that has only server side implementation and then instantiate it liberally in tests. This should exercise code paths handling such case without any :crossed_fingers: downsides.

Easy to test with `just mprocs`:

image


fedimintd/src/fedimintd.rs
```diff
@@ -182,9 +184,16 @@ impl Fedimintd {
...
+            .with_module(LightningInit)
             .with_module(MintInit)
-            .with_module(WalletInit)
+            .with_module(WalletInit);
+
+        if std::env::var_os(FM_USE_UNKNOWN_MODULE_ENV).is_some() {
+            s.with_module(UnknownInit)
+        } else {
+            s
+        }
```
elsirion, line 196: I'm not a fan of branching like this in production code just to make tests work. Aren't rust integration tests enough? If we really want devimint tests we can build a separate testing `fedimintd` with this extra module.

dpc: Having separate binaries seem cumbersome for integrations and alike. There's really nothing hacky or insecure about this module, it's a proper module that... is not very useful in practice. If you don't like branching, I could just make `fedimintd` produce this module unconditionally. :D

I bet eventually default Fedimint will ship with some optional modules anyway, or even the current modules will become opitonal, and users will be choosing them at DKG. It's just we don't have this API surface yet, so an env var is used.

scripts/tests/backwards-compatibility-test.sh
```diff
@@ -11,6 +11,9 @@ versions=( "${@:-${default_versions[@]}}" )
 
 # signal to downstream test scripts
 export FM_BACKWARDS_COMPATIBILITY_TEST=1
+# previous versions did not have unknown module so can't support it,
```
dpc, line 14: @bradleystachurski Food for thought: we will need to deal with all sorts of incompatibilities between versions.

dpc:
Hey, it detect one legitimate bug right away.

justinmoon:
https://github.com/fedimint/fedimint/issues/3548 will be a good one to test with this

dpc:
> #3548 will be a good one to test with this

That was the one it found, I just pushed it as a separate PR, as it's not clear when we'll land this and how yet.

fedimint-client/src/lib.rs
```diff
@@ -1254,7 +1254,11 @@ impl Client {
                         *instance_id,
                         JsonWithKind::new(
                             kind.clone(),
-                            config.clone().expect_decoded().to_json().into(),
+                            config
+                                .clone()
+                                .decoded()
+                                .map(|decoded| decoded.to_json().into())
+                                .unwrap_or(serde_json::Value::Null),
```
elsirion, line 1261: Eventually we might want to display the raw binary instead of null?

elsirion (dismissed):

scripts/tests/backwards-compatibility-test.sh
bradleystachurski: Yeah, I'd anticipate we'll have a lot of conditional logic in devimint based on versions similar to [`18f00a8` (#4247)](https://github.com/fedimint/fedimint/pull/4247/commits/18f00a86e731a1c037f2345b7f1e3aa9890022c0#diff-75cef4315751be4fa6703860684a38e9c382d522c05f6df275846e94a373562dR102)

elsirion (approved):

fedimint-core/src/core.rs
```diff
@@ -140,6 +140,7 @@ pub const MODULE_INSTANCE_ID_GLOBAL: u16 = u16::MAX;
 pub const LEGACY_HARDCODED_INSTANCE_ID_LN: ModuleInstanceId = 0;
 pub const LEGACY_HARDCODED_INSTANCE_ID_MINT: ModuleInstanceId = 1;
 pub const LEGACY_HARDCODED_INSTANCE_ID_WALLET: ModuleInstanceId = 2;
+pub const LEGACY_HARDCODED_INSTANCE_ID_UNKNOWN: ModuleInstanceId = 3;
```
bradleystachurski, line 143: Is the comment above still relevant?

```
// Note: needs to be in alphabetical order of ModuleKind of each module,
// as this is the ordering we currently hardcoded.
```

fedimint-core/src/encoding/mod.rs
```diff
@@ -891,6 +891,13 @@ impl<T> DynRawFallback<T>
 where
     T: Decodable + 'static,
 {
+    pub fn decoded(self) -> Option<T> {
```
bradleystachurski, line 894: [nit] New pub fn without docs

fedimint-core/src/envs.rs
```diff
@@ -2,6 +2,15 @@
 // introducing extra dependencies between core and build modules.
 pub const FEDIMINT_BUILD_CODE_VERSION_ENV: &str = "FEDIMINT_BUILD_CODE_VERSION";
 
+/// In tests we want to routinely enable an extra unknown module to ensure
+/// all client code handles correct modules that client doesn't know about.
+pub const FM_USE_UNKNOWN_MODULE_ENV: &str = "FM_USE_UNKNOWN_MODULE";
+
+/// Check if env variable is set and not equal `0` or `false` which are common
+/// ways to disable something.
+pub fn is_env_var_set(var: &str) -> bool {
```
bradleystachurski, line 11: If you make changes, might be able to sneak this in https://github.com/fedimint/fedimint/pull/4247#discussion_r1500582466

No biggie, I can also address in a followup.

fedimint-testing/src/fixtures.rs
```diff
@@ -113,6 +113,19 @@ impl Fixtures {
         self
     }
 
+    pub fn with_server_only_module(
```
bradleystachurski, line 116: Would this make sense to use for other fixtures instead of only ln-gateway?

modules/fedimint-unknown-server/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "fedimint-unknown-server"
+version = "0.3.0-alpha"
+authors = ["The Fedimint Developers"]
+edition = "2021"
+description = "fedimint-dummy is a test fedimint module that doesn't have any client side implementation."
```
bradleystachurski, line 6: [nit] s/fedimint-dummy/fedimint-unknown

modules/fedimint-unknown-server/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+        Vec::new()
+    }
+
+    async fn process_consensus_item<'a, 'b>(
+        &'a self,
+        _dbtx: &mut DatabaseTransaction<'b>,
+        _consensus_item: UnknownConsensusItem,
+        _peer_id: PeerId,
+    ) -> anyhow::Result<()> {
+        bail!("The dummy module does not use consensus items");
```
bradleystachurski, line 155: [nit] s/dummy/unknown

bradleystachurski (approved):
Looks good!

CI is complaining but I think you can resolve by building locally and committing `Cargo.lock`.

scripts/tests/backwards-compatibility-test.sh
bradleystachurski: It might be useful to come up with a system that'll allow us to easily cleanup conditional logic that's no longer needed in devimint.

e.g. `// FIXME: deprecated v0.2.1` so we can grep for a version when it's reached EOL

fedimint-testing/src/fixtures.rs
dpc: Yes, but with very diminishing returns.

fedimint-core/src/core.rs
dpc: I just got rid of it.

scripts/tests/backwards-compatibility-test.sh
dpc: :brain: But we should make it an issue and brainstorm a little.

bradleystachurski: 👍 https://github.com/fedimint/fedimint/issues/4424

bradleystachurski (approved):

elsirion (approved):

PR #4405, opened 2024-02-23
feat: replit support

Kodylow:
With Replit's new Nix Persistence feature enabled, this PR make Fedimint's developer environment work in Replit using the nix flake and mprocs. 

Fork Fedimint into a Repl, it automatically start you in the nix flake devShell, hitting "Run" start the mprocs dev environment

Special thanks to @ryantm !

image


.envrc
```diff
@@ -1 +1,4 @@
+if ! has nix_direnv_version || ! nix_direnv_version 3.0.4; then
```
dpc, line 1: Please add a comment explaining why is this needed.

.replit
```diff
@@ -1,47 +1,15 @@
...
-# # Problem: `nix develop` too large
-#
-# There's too many dependencies in default shell, so I created a minimal `.#replit` dev shell.
-#
-# With this `bidgen` works.
-run = "nix --extra-experimental-features nix-command --extra-experimental-features flakes develop .#replit -c cargo check"
+# Replit support 
+run = "direnv allow && nix develop --command just mprocs"
 hidden = ["target"]
+entrypoint = "src/main.rs"
```
dpc, line 4: We don't have a main `src/main.rs` file.

elsirion (approved):

elsirion:
After this PR things will likely be less broken than before, but if you have the time documenting what options do/why they are set would be great. Also, is this replit integration useful enough that we should have a badge for it? https://docs.replit.com/programming-ide/using-git-on-replit/running-github-repositories-replit#adding-a-run-on-replit-badge

.envrc
Kodylow: Replit needs that check for direnv to get picked up when you reconnect, if it's not there when you reconnect in browser then it kicks you out of the nix develop shell and you have to run `nix develop` again, not a big deal if I remove

.replit
Kodylow: Entrypoint is just the first file shown when you open the repl, changing it to our README

.envrc
elsirion: I think just adding this comment in the code would go a long way re maintainability.

Kodylow:
We could add the run with replit badge after nix persistence gets generally launched, for now it requires a special flag to use nix flakes. The normal replit.nix works now but once we have the flake option then when someone forks fedimint on replit it will start with the built nix artifacts for the flake on master so getting into the develop devShell takes like 10 seconds which will be super useful.

.envrc
dpc: Yeah, I don't think it hurts to have it here, but we just need a comment why it is there, otherwise it looks odd, unnecessary and even suspicious.

.config/bashrc
```diff
@@ -0,0 +1 @@
+eval "$(direnv hook bash)"
```
dpc, line 1: While at it, please add a comment here as well on top, that the whole thing is for replit support.

.replit
```diff
@@ -53,11 +21,15 @@ packageSearch = true
...
 # this must be kept in sync with `nixpkgs` channel in `flake.nix`
-channel = "stable-22_11"
+channel = "stable-23_11"
 
 [gitHubImport]
 requiredFiles = [".replit", "replit.nix"]
+
+[[ports]]
+localPort = 8333
+externalPort = 80
```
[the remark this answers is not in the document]
Kodylow: they're for deployments but we can't deploy currently since there's no way to persist rocksdb, removing.

dpc (approved):

.gitignore
```diff
@@ -8,3 +8,7 @@
 .direnv
 .vscode
 .DS_Store
+
+.cargo/registry
+.cargo/.package-cache
+.cargo/.package-cache-mutate
```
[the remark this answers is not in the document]
Kodylow: added it, our git hooks throw weird errors in replit so have to --no-verify, I logged the git issues with them

dpc (approved):

.gitignore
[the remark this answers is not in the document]
dpc: Well, I'm more thinking about long term solution so replit contributors get the text editor to do the right thing automatically. Possibly replit supports `.editconfig` or something?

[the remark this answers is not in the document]
dpc: https://editorconfig.org/

replit.nix
```diff
@@ -1,5 +1,11 @@
...
+    env = {
+        OPENSSL_DIR = "${pkgs.openssl.dev}";
+        LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
+      };
   deps = [
+    pkgs.direnv
+    pkgs.just
     pkgs.gcc
     pkgs.mold
     pkgs.rustc
```
elsirion, line 11: This looks rather redundant, why can't we use our normal shell in replit?

elsirion (approved):

replit.nix
Kodylow: This lets you build it outside nix develop without throwing an error when the nix persistence flag is enabled. If you have nix persistence enabled it ignores the replit.nix and just uses the flake

PR #4409, opened 2024-02-23
chore(client): document joining a federation

### LINKED ISSUE #4397: Add network to fedimint invite code
It is non-obvious how to check for the network when joining a federation. The current way requires creating a wallet client and asking what the network is after you've connected to the federation. Would be nice if you check up font without needing to connect.

I imagine this was never done because the network is really at the module level rather than the federation level but would be nice to be able to have some indication.

elsirion:
Hopefully fixes #4397 (mostly a documentation issue imo). @benthecarman would this have helped?

Screenshot 2024-02-23 at 18-16-44 ClientBuilder in fedimint_client - Rust

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1792,6 +1792,11 @@ impl ClientBuilder {
 
     /// Join a new Federation
     ///
+    /// When a user wants to connect to a new federation this function fetches
+    /// the federation config and initializes the client database. If a user
+    /// already joined the federation in the past and has a preexisting database
+    /// use [`ClientBuilder::open`] instead.
+    ///
     /// **Warning**: Calling `join` with a `root_secret` key that was used
     /// previous to `join` a Federation will lead to all sorts of malfunctions
     /// including likely loss of funds.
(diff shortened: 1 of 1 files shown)


benthecarman:
That's look good one! One of thing we ran into was downloading the config every start up, I think these docs would make people think they need to do a similar thing.

elsirion:
I think  #4326 will land before this, then I'll have to rebase/fix the docs anyway and can add something about `open`.

elsirion:
I added the comment about `open`.

fedimint-client/src/lib.rs
```diff
@@ -1792,6 +1792,11 @@ impl ClientBuilder {
 
     /// Join a new Federation
     ///
+    /// When a user wants to connet to a new federation this function fetches
```
bradleystachurski, line 1795: [nit] s/connet/connect

bradleystachurski (approved):
Looks good!

fedimint-client/src/lib.rs
elsirion: Thx, fixed :)

bradleystachurski (approved):

dpc (approved):

Issue #4350, opened 2024-02-17
test_can_change_routing_fees flaky


dpc:
```
fedimint-test-all-ci> 00:00:09 thread 'test_can_change_routing_fees' panicked at gateway/ln-gateway/tests/integration_tests.rs:300:13:
fedimint-test-all-ci> 00:00:09 assertion `left == right` failed
fedimint-test-all-ci> 00:00:09   left: Amount { msats: 750000 }
fedimint-test-all-ci> 00:00:09  right: Amount { msats: 747490 }
```

https://github.com/fedimint/fedimint/actions/runs/7934025448/job/21664069124

Will disable.

dpc:
Another (different?): https://github.com/fedimint/fedimint/actions/runs/8008441209/job/21874886159

PR #4410, opened 2024-02-23
fix: flaky test_can_change_routing_fees

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/4350#event-11895482425

Lightning client will choose a random gateway when it doesn't have an "active" gateway. In the routing fees test, we update each gateway's routing fees independently, so it is possible the gateway with the old fees is chosen, which results in the test failing since the incorrect amount is paid.

Side note....we should remove the notion of an active gateway :)


gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -290,14 +291,19 @@ async fn test_can_change_routing_fees() -> anyhow::Result<()> {
...
+                .get_first_module::<LightningClientModule>()
+                .set_active_gateway(&gateway.gateway.gateway_id)
+                .await?;
+
             // Create test invoice
             let invoice_amount = sats(250);
             let invoice = other_lightning_client.invoice(invoice_amount, None).await?;
 
             let gateway = gateway.remove_client_hack(&fed).await;
             pay_valid_invoice(invoice, &user_client, &gateway).await?;
```
elsirion, line 304: Could we, in preparation for deprecating the active gateway concept, add a gateway id argument to that function here? I think internally it already calls a function that accepts a gateway as argument.

m1sterc001guy: Good call. Done

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -145,13 +145,19 @@ pub fn sha256(data: &[u8]) -> sha256::Hash {
...
     invoice: Bolt11Invoice,
     user_client: &ClientArc,
-    client: &ClientArc,
+    gateway_client: &ClientArc,
+    gateway_id: &PublicKey,
 ) -> anyhow::Result<()> {
-    // User client pays test invoice
+    // Set the active gateway to be the gateway with id `gateway_id`
     let user_lightning_module = &user_client.get_first_module::<LightningClientModule>();
+    user_lightning_module.set_active_gateway(gateway_id).await?;
```
elsirion, line 158: Sorry, should have been more clear: a few lines down you call `pay_invoice`:

https://github.com/fedimint/fedimint/blob/f26de71c1ed23565cfcb34934183961eed58898e/gateway/ln-gateway/tests/integration_tests.rs#L59-L65

Internally that already just fetches a gateway (I don't even know if it still cares about the active gateway concept, i.e. if this PR would improve anything). If we don't fetch a random gateway but can supply one then we don't have to call `set_active_gateway` here anymore and get one step closer to removing it altogether.

justinmoon (dismissed):

gateway/ln-gateway/tests/integration_tests.rs
m1sterc001guy: I add the function `select_gateway` which should be helpful moving forward in removing the active gateway concept. This PR fixes the flaky test, I'm going to work on removing the active gateway concept next.

elsirion (dismissed):

dpc (dismissed):

dpc:
Need'a rebase @m1sterc001guy

dpc (approved):

elsirion (approved):

PR #4412, opened 2024-02-23
fix: Gateway Enforce Routing Fees

m1sterc001guy:
In debugging https://github.com/fedimint/fedimint/issues/4350 I realized that the gateway is not independently verifying that the client attached routing fees to the contract. We check if the contract is underfunded, but only against the invoice amount, not invoice amount + gateway fee. 

This PR adds gateway enforcement of fees and adds a test that verifies clients cannot cheat and not pay fees.

### DIFF
--- a/gateway/ln-gateway/tests/integration_tests.rs
+++ b/gateway/ln-gateway/tests/integration_tests.rs
@@ -48,6 +48,9 @@ use ln_gateway::rpc::rpc_client::{GatewayRpcClient, GatewayRpcError, GatewayRpcR
 use ln_gateway::rpc::{
     BalancePayload, ConnectFedPayload, LeaveFedPayload, SetConfigurationPayload,
 };
+use ln_gateway::state_machine::pay::{
+    OutgoingContractError, OutgoingPaymentError, OutgoingPaymentErrorType,
+};
 use ln_gateway::state_machine::{
     GatewayClientModule, GatewayClientStateMachines, GatewayExtPayStates, GatewayExtReceiveStates,
     GatewayMeta, Htlc,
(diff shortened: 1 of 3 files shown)


elsirion:
Iirc we don't check that fees were paid but only allow the LN node to pay as much in fees as the user. We don't have a guarantee of making money anyway rn (could introduce a fee margin that's guaranteed to go to the gateway to fix this).

m1sterc001guy:
Right now the only check we have is if a contract is "underfunded", which essentially just compares the amount in the contract to the amount that the client tells the gateway to pay. The issue is that, if the gateway is charging a routing fee, we do not currently independently compute that fee. So malicious clients could create a contract that pays exactly the amount that the invoice is, and the gateway won't enforce that their fee is included.

This is hard to exploit currently since it requires a malicious client to not attach fees for the gateway. In practice though this probably can occur if a gateway changes their advertised fees, since it might not re-register with the federation for another ~10mins.

m1sterc001guy:
@elsirion let me know if you think this is still necessary, I think we're on the same page. This essentially just turns the under-funding of a contract (without fees) into an explicit error, whereas before the gateway would rely on routing on the LN to make a profit

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -311,6 +314,11 @@ async fn test_can_change_routing_fees() -> anyhow::Result<()> {
             })
             .await;
 
+            user_client
```
m1sterc001guy, line 317: Remove

elsirion (approved):
I don't think it's super useful rn, but it's well-tested and might become useful if we ever want to enforce profit for gateways.

justinmoon (approved):

PR #4413, opened 2024-02-24
refactor: run backwards-compatibility tests in parallel

### LINKED ISSUE #4206: Run backwards-compatibility tests in parallel
This will hopefully reduce the total time required for the workflow to finish in CI.

Related: https://github.com/fedimint/fedimint/pull/4010#issuecomment-1885300207

bradleystachurski:
Closes #4206 

Runs backwards-compatibility tests in parallel using `parallel`. This reduces the time to run backwards-compatibility in CI from ~21 minutes down to ~15 minutes and I'd expect the time savings to be material for the full backwards-compatibility matrix.

This requires refactoring how backwards-compatibility tests call `test-ci-all.sh`, where we now pass as arguments the versions to run each bash test function. The existing behavior of `test-ci-all.sh` is preserved by running as `current`.

For further improvements, we can consider merging the backwards-compatibility CI workflow with the build on linux test step. I've tested merging locally, which adds ~5 minutes to the build on linux tests step, so I'd prefer to keep backwards-compatibility separate for now.


justfile.fedimint.just
```diff
@@ -23,7 +23,7 @@ test-count:
   ./scripts/tests/test-cov.sh
 
 test-compatibility *VERSIONS:
-  ./scripts/tests/backwards-compatibility-test.sh {{VERSIONS}}
+  ./scripts/tests/test-ci-all.sh {{VERSIONS}}
```
bradleystachurski, line 26: For those using `just test-compatibility`, this change removes the default so you now need to explicitly pass which versions to test (e.g. `just test-compatibility v0.2.1`)

scripts/tests/test-ci-all.sh
```diff
@@ -151,6 +146,64 @@ function always_success_test() {
...
 
+tagged_versions=("$@")
+num_versions="$#"
+versions=( "current" "${tagged_versions[@]}" )
+if [[ "$num_versions" == "0" ]]; then
+  current_binaries_count=3
+else
+  # precompile binaries
+  binaries=( "fedimintd" "fedimint-cli" "gateway-cli" "gatewayd" )
+  parallel nix_build_binary_for_version "{1}" "{2}" ::: "${binaries[@]}" ::: "${tagged_versions[@]}"
```
elsirion, line 157: Is running the build in parallel actually useful? At which point will resource contention slow down things again? Significant parts of the build process run at 100% CPU for me. @dpc

elsirion (approved):
Looks good, always learning some new arcane bash syntax when reviewing these build system PRs :laughing:

scripts/tests/test-ci-all.sh
bradleystachurski: I'm not sure if running this with `parallel` will have any performance gains relative to nested for loops, but I was a fan of how concise parallel allows us to write nested for loops.

Worth noting that precompiling here does have a benefit later since calling `run_test_for_versions` in parallel will kick off multiple nix builds for the same binary and version (https://github.com/fedimint/fedimint/pull/4413/files#diff-6f67d3a8330a3c924847c4e3e44fa893ed4178fee92b9bfeba68fd6fdbaff5ecR43-R45).

dpc (approved):

scripts/tests/test-ci-all.sh
dpc: > Significant parts of the build process run at 100% CPU for me.

Are you running at least `CARGO_PROFILE=ci` build profile?

dpc:
https://github.com/fedimint/fedimint/pull/4418 some follow-ups.

Issue #4280, opened 2024-02-09
Gateway should retry if registration fails


justinmoon:
The following code is run on a timer to re-register gateways with federations: https://github.com/fedimint/fedimint/blob/1738332d8e6366cfe4c44561c11189337f8f4e27/gateway/ln-gateway/src/lib.rs#L1098-L1144

The behavior is that if a federation is unreachable at the moment this timer goes off (once per 10 minutes), then your gateway won't be registered to the federation for the next 10 minutes until the timer goes off again. I think it would be better if it tried retrying, say, every 10 seconds while registration is failing.

kernelkind:
https://github.com/fedimint/fedimint/blob/9f4926cb9cba50bb016df59a4989e62f657d01a6/gateway/ln-gateway/src/lib.rs#L1322-L1338

Is this not already happening on line 1334?

justinmoon:
It seems like line 1334 is triggered when the gateway isn't in a "running" state. IIRC this might mean that it hasn't connected to the lightning node yet, for example. 

If `register_all_federations()` call fails, we'll sleep `GW_ANNOUNCEMENT_TTL.mul_f32(0.85)` (line 1346) which is 8.5 minutes.

PR #4414, opened 2024-02-24
chore: decrease sleep time when registration fails

kernelkind:
If a federation is unreachable, the delay for retrying is reduced to 10 seconds instead of a multiple of GW_ANNOUNCEMENT_TTL.

Closes: https://github.com/fedimint/fedimint/issues/4280

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -1152,7 +1152,9 @@ impl Gateway {
 
             dbtx.commit_tx().await;
 
-            Self::register_all_federations(self, gateway_config).await;
+            if let Err(e) = Self::register_all_federations(self, gateway_config).await {
+                warn!("{e:?}")
+            };
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -1182,6 +1185,7 @@ impl Gateway {
                             .await
                             {
                                 warn!("Error registering federation {federation_id}: {e:?}");
+                                Err(GatewayError::FederationError(FederationError::general(e)))?
```
elsirion, line 1188: Might it make sense to not log here now that we return errors and just let the caller decide at which level to log? That way one of the call-sites wouldn't have to ignore the result entirely.

kernelkind: I could see it making sense if there are callers that don't care to log some things, but it could start getting messy if we replaced all logs in this method with errors. For each error they desire to log for, each caller would have to match the error and log

elsirion (approved):

gateway/ln-gateway/src/lib.rs
m1sterc001guy: + 1 to Eric's comment. Returning errors would actually make this code a bit more concise. The warnings below could just be returned as errors and the caller can log them

elsirion (approved):

m1sterc001guy (approved):

PR #4415, opened 2024-02-24
Allow downloading guardian config

### LINKED ISSUE #4039: Make guardian config+keys backupable via the admin UI
* Tar up the server config files
* Encode and send via Admin API
* UI has to decode and allow user to download

elsirion:
Fixes #4039 on the `fedimintd` side, also needs a UI feature. 

@EthnTuttle, @Kodylow how hard would that be to integrate on the frontend side? [It seems possible to download a file "from memory", although in the typical web tech hacky way :see_no_evil: ](https://stackoverflow.com/questions/3665115/how-to-create-a-file-in-memory-for-user-to-download-but-not-through-server)

For manual testing you can run

```bash
fedimint-cli --our-id 0 --password pass admin guardian-config-backup | jq -r .tar_archive_bytes | xxd -r -p  > backup.tar
```

it decodes the returned hex and writes it to `backup.tar`. Comparing the files they look reasonably similar (some differences due to non-pretty encoding):

```
$ ls -la target/devimint/fedimintd-0/
total 200
drwxr-xr-x  3 user users   4096 Feb 25 21:19 .
drwxr-xr-x 13 user users   4096 Feb 25 21:19 ..
-rw-r--r--  1 user users  59805 Feb 25 21:19 client.json
-rw-r--r--  1 user users 103889 Feb 25 21:19 consensus.json
drwxr-xr-x  2 user users   4096 Feb 25 21:42 database
-rw-r--r--  1 user users    898 Feb 25 21:19 local.json
-rw-r--r--  1 user users      4 Feb 25 21:19 password.private
-rw-r--r--  1 user users  10454 Feb 25 21:19 private.encrypt
-rw-r--r--  1 user users     22 Feb 25 21:19 private.salt

$ tar -tvf backup.tar 
---------- 0/0             628 1970-01-01 01:00 local.json
---------- 0/0          100143 1970-01-01 01:00 consensus.json
---------- 0/0              22 1970-01-01 01:00 private.salt
---------- 0/0           10454 1970-01-01 01:00 private.encrypt
```

TODO here: testing. I'm thinking of deleting the config of a guardian, replacing it with the tar archive versions and restarting it to see if it can start. I hope devimint supports such shenanigans.

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::io::Write;
 use std::ops::ControlFlow;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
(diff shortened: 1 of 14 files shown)


fedimint-server/src/net/api.rs
```diff
@@ -277,6 +279,42 @@ impl ConsensusApi {
...
+        append(
+            "consensus.json",
+            &serde_json::to_vec(&self.cfg.consensus).expect("Error encoding consensus config"),
+        );
+
+        // TODO: encrypt
+        append(
+            "private.json",
+            &serde_json::to_vec(&self.cfg.private).expect("Error encoding private config"),
+        );
```
elsirion, line 309: Encrypt using password given for auth. This means the salt/encrypted data of the downloaded backup will differ from the on-disk version, but the password used will still be the guardian password.

EthnTuttle:
> fedimint-cli --our-id 0 --password pass admin guardian-config-backup | jq -r .tar_archive_bytes | xxd -r -p  > backup.tar

Yikes! bytes embedded in a JSON file. Not exactly how I'd serve a binary file but I don't see what this can't work in the UI. I assume it was easy to integrate with the existing websockets API instead of spinning up and entire web server?

elsirion:
> I assume it was easy to integrate with the existing websockets API instead of spinning up and entire web server?

Yes, exactly that :sweat_smile: :sob:

scripts/tests/test-ci-all.sh
```diff
@@ -183,6 +189,7 @@ tests_to_run_in_parallel=(
   "devimint_cli_test_single"
   "load_test_tool_test"
   "recoverytool_tests"
+  "guardian_backup"
```
elsirion, line 192: @bradleystachurski /@dpc am I holding it correctly?

scripts/tests/test-ci-all.sh
```diff
@@ -89,6 +89,12 @@ function latency_test_restore() {
 }
 export -f latency_test_restore
 
+function guardian_backup() {
+  # guardian-backup-test runs a degraded federation, so we need to override FM_OFFLINE_NODES
```
bradleystachurski, line 93: 👍

scripts/tests/test-ci-all.sh
bradleystachurski: Yep, this is perfect. Do you know if this test takes a while? If so, we might want to move it higher up so `parallel` starts it sooner.

elsirion:
@elsirion: bump API version!!!

dpc (approved):

scripts/tests/test-ci-all.sh
elsirion: Nah, running it locally was under 30s I'd say

devimint/src/tests.rs
```diff
@@ -1812,6 +1813,142 @@ pub async fn recoverytool_test(dev_fed: DevFed) -> Result<()> {
     Ok(())
 }
 
+pub async fn guardian_backup_test(dev_fed: DevFed, process_mgr: &ProcessManager) -> Result<()> {
```
justinmoon, line 1816: Great test

maan2003 (commented):
ack devimint changes

devimint/src/tests.rs
```diff
@@ -1812,6 +1813,142 @@ pub async fn recoverytool_test(dev_fed: DevFed) -> Result<()> {
...
+    use fedimint_core::core::LEGACY_HARDCODED_INSTANCE_ID_WALLET;
+
+    log_binary_versions().await?;
+
+    let fedimint_cli_version = crate::util::FedimintCli::version_or_default().await;
+    let fedimintd_version = crate::util::FedimintdCmd::version_or_default().await;
+    if VersionReq::parse("<0.3.0-alpha")?.matches(&fedimint_cli_version)
+        || VersionReq::parse("<0.3.0-alpha")?.matches(&fedimintd_version)
```
bradleystachurski, line 1824: Good candidate to tag with

```
// TODO(support:v0.2)
```
https://github.com/fedimint/fedimint/issues/4424

devimint/src/tests.rs
```diff
@@ -1812,6 +1813,142 @@ pub async fn recoverytool_test(dev_fed: DevFed) -> Result<()> {
...
+            .write(true)
+            .create(true)
+            .open(data_dir.join(name))
+            .expect("could not open file");
+        file.write_all(data).expect("could not write file");
+        file.flush().expect("could not flush file");
+    };
+
+    write_file("backup.tar", &backup_tar);
+    write_file("password.private", "pass".as_bytes());
```
bradleystachurski, line 1903: [nit] It might be more obvious how this filename is special using the pub const

```
write_file(
    fedimint_server::config::io::PLAINTEXT_PASSWORD,
    "pass".as_bytes(),
);
```

nix/flakebox.nix
```diff
@@ -357,6 +357,12 @@ rec {
     buildPhaseCargoCommand = "patchShebangs ./scripts ; ./scripts/tests/latency-test.sh";
   };
 
+  guardianBackupTest = craneLibTests.mkCargoDerivation {
```
bradleystachurski, line 360: For devimint tests should we continue adding these? I believe we're only using `ciTestAllBase` which will handle calling each individual test.

cc: @dpc

bradleystachurski (approved):
Looks good!

devimint/src/tests.rs
elsirion: Will do a follow-up for your nits :)

fedimint-server/src/config/mod.rs
```diff
@@ -208,7 +208,7 @@ impl ServerConfig {
     pub fn supported_api_versions() -> SupportedCoreApiVersions {
         SupportedCoreApiVersions {
             core_consensus: CORE_CONSENSUS_VERSION,
-            api: MultiApiVersion::try_from_iter([ApiVersion { major: 0, minor: 1 }])
+            api: MultiApiVersion::try_from_iter([ApiVersion { major: 0, minor: 2 }])
```
bradleystachurski, line 211: Do we also need to bump the client core api version?

https://github.com/fedimint/fedimint/blob/427f09b/fedimint-client/src/lib.rs#L659

elsirion: I don't think so, we'd do that once we drop support for 0.0 afaik. @dpc

PR #4417, opened 2024-02-26
chore: add blank .semgrepignore & fix findings

kernelkind:
If .semgrepignore is not present, semgrep uses a default .semgrepignore file. The defaults excluded some tests which is not the desired outcome.

Add a blank .semgrepignore file to indicate that nothing additional to .gitignore should be ignored by semgrep.

Also, add a second pattern to ban-direct-sleep-in-tests rule to find the sleep function when the fully qualified function name is present.

### DIFF
--- a/.config/semgrep.yaml
+++ b/.config/semgrep.yaml
@@ -105,13 +105,14 @@ rules:
   languages:
     - rust
   patterns:
-    - pattern: sleep($X)
+    - pattern-either:
+      - pattern: sleep(...)
+      - pattern: fedimint_core::task::sleep(...)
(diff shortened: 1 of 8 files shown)


kernelkind:
I removed `fedimint-load-test-tool/` because the code inside the directory seem to not be tests themselves, but tools to run tests. I can re-add it and add the `sleep_in_test` methods if it actually is needed though

dpc (approved):

elsirion (approved):

elsirion:
Good catch, thx!

PR #4418, opened 2024-02-26
Parallel backward compat. tests improvements

dpc:
Follow up to #4206

### DIFF
--- a/.github/workflows/ci-backwards-compatibility.yml
+++ b/.github/workflows/ci-backwards-compatibility.yml
@@ -43,6 +43,13 @@ jobs:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         continue-on-error: true
 
+      # caching ~/.cargo to avoid downloading all (especially git) deps
+      - name: Cache ~/.cargo
+        uses: actions/cache@v4
+        with:
+          path: ~/.cargo
+          key: ${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
+
       - name: Backwards-compatibility tests
         run: |
           # the default tmp dir is too long (/home/ubuntu/actions-runner/_work/_temp/)
--- a/.github/workflows/ci-backwards-compatibility.yml
+++ b/.github/workflows/ci-backwards-compatibility.yml
@@ -47,7 +47,7 @@ jobs:
         run: |
           # the default tmp dir is too long (/home/ubuntu/actions-runner/_work/_temp/)
           # we need to use `nix develop -c` to be able to use `nix build` inside of backwards-compatibility-test
-          env TMPDIR=/tmp nix develop -c ./scripts/tests/test-ci-all.sh v0.2.1
+          env TMPDIR=/tmp CARGO_PROFILE=ci nix develop -c ./scripts/tests/test-ci-all.sh v0.2.1
 
       - name: Prepare failed test build dirs
         if: always()
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -55,6 +55,15 @@ function run_test_for_versions() {
     export FM_OFFLINE_NODES=1
   fi
 
+  if \
+    [ "$fed_version" != "current" ] ||
+    [ "$client_version" != "current" ] ||
+    [ "$gateway_version" != "current" ] ; then
+    export FM_RUN_TEST_VERSIONS="$fed_version $client_version $gateway_version"
+  else
+    unset FM_RUN_TEST_VERSIONS
+  fi
+
   $fn_name
 }
 export -f run_test_for_versions
--- a/scripts/tests/test-ci-all.sh
+++ b/scripts/tests/test-ci-all.sh
@@ -221,7 +221,7 @@ echo "$parsed_test_commands" | if parallel \
   --halt-on-error 1 \
   --joblog "$joblog" \
   --timeout 600 \
-  --load 150% \
+  --load 1000% \
   --delay 5 \
   --jobs "$parallel_jobs" \
   --memfree 1G \
--- a/scripts/tests/test-ci-all.sh
+++ b/scripts/tests/test-ci-all.sh
@@ -196,8 +196,8 @@ tests_to_run_in_parallel=(
 )
 
 tests_with_versions=()
-for version_combo in "${version_matrix[@]}"; do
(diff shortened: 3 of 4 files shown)


scripts/_common.sh
```diff
@@ -55,6 +55,15 @@ function run_test_for_versions() {
     export FM_OFFLINE_NODES=1
   fi
 
+  if \
+    [ "$fed_version" != "current" ] ||
+    [ "$client_version" != "current" ] ||
+    [ "$gateway_version" != "current" ] ; then
+    export FM_RUN_TEST_VERSIONS="$fed_version $client_version $gateway_version"
+  else
+    unset FM_RUN_TEST_VERSIONS
```
elsirion, line 64: Was this env var set before elsewhere or are you just unsetting it to avoid confusion if it's ever set by accident?

scripts/tests/test-ci-all.sh
```diff
@@ -221,7 +221,7 @@ echo "$parsed_test_commands" | if parallel \
   --halt-on-error 1 \
   --joblog "$joblog" \
   --timeout 600 \
-  --load 150% \
+  --load 1000% \
```
elsirion, line 224: @dpc wants to see our CPUs burn :laughing:

Q4tDgE7

elsirion (approved):

scripts/_common.sh
bradleystachurski: With the current setup, I don't believe this env var will ever be set since each call to `run_test_for_versions` is in `parallel`, so each job should be setting `FM_RUN_TEST_VERSIONS` for the first time.

I'm a fan of keeping though since it removes a potential footgun.

maan2003 (approved):

scripts/tests/test-ci-all.sh
bradleystachurski: Let it rip!

I played with tweaking this in the previous refactor (https://github.com/fedimint/fedimint/pull/4413) and noticed flakiness with the latency tests. Worth keeping an eye on.

- Should we also relax `--nice 15`?
- Would it make sense to make `--load` profile dependent? Aggressive in CI but slower locally?

scripts/tests/test-ci-all.sh
```diff
@@ -196,8 +196,8 @@ tests_to_run_in_parallel=(
 )
 
 tests_with_versions=()
-for version_combo in "${version_matrix[@]}"; do
-  for test in "${tests_to_run_in_parallel[@]}"; do
+for test in "${tests_to_run_in_parallel[@]}"; do
+  for version_combo in "${version_matrix[@]}"; do
```
bradleystachurski, line 200: Good call 😄

.github/workflows/ci-backwards-compatibility.yml
```diff
@@ -43,11 +43,18 @@ jobs:
...
+        with:
+          path: ~/.cargo
+          key: ${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
+
       - name: Backwards-compatibility tests
         run: |
           # the default tmp dir is too long (/home/ubuntu/actions-runner/_work/_temp/)
           # we need to use `nix develop -c` to be able to use `nix build` inside of backwards-compatibility-test
-          env TMPDIR=/tmp nix develop -c ./scripts/tests/test-ci-all.sh v0.2.1
+          env TMPDIR=/tmp CARGO_PROFILE=ci nix develop -c ./scripts/tests/test-ci-all.sh v0.2.1
```
bradleystachurski, line 57: Out of scope for this PR - I've been playing with using `nix build .#ci.workspaceBuild` and extracting using zstd (referencing the logic in [inheritCargoArtifactsHook](https://github.com/ipetkov/crane/blob/7d5b46c17d857ee9ddb2e8d88185729a3e5637b6/lib/setupHooks/inheritCargoArtifactsHook.sh#L27)). That works well, but I can't get rustc to respect `FM_CARGO_DENY_COMPILATION`. Perhaps you know why?

I can open a separate PR to discuss.

bradleystachurski (approved):

.github/workflows/ci-backwards-compatibility.yml
bradleystachurski: Also, back-compat is running fast enough where this is no longer the bottleneck in CI. Perhaps it's not worth the effort to solve the intermediate zstd extraction.

scripts/_common.sh
dpc: I think it's not needed, but explit here doesn't hurt, and I'm paranoid.

scripts/tests/test-ci-all.sh
dpc: On modern systems `nice` doesn't really do much (it's relative only to the current terminal session, or something like that, and not global), so tweaking nice doesn't really change much one way or the other.

If I run it locally I want it to be done ASAP as well. I was thinking about making it `10000` an locally I tested 30000.

If anything gets flaky that's great, maybe we've uncovered a bug. Unless it's about something getting too slow, then I guess we'll back off a bit. :D

.github/workflows/ci-backwards-compatibility.yml
dpc: > Perhaps you know why?

@bradleystachurski Need to make sure that timestamps, file ownership, location, etc. are all preserved. See [crane's version](https://github.com/ipetkov/crane/blob/2c94ff9a6fbeb9f3ea0107f28688edbe9c81deaa/lib/setupHooks/inheritCargoArtifactsHook.sh#L28)

> Perhaps it's not worth the effort to solve the intermediate zstd extraction.

Yeah, I think it's good enough right now.

scripts/tests/test-ci-all.sh
dpc: One Linux you want to be using `chrt` and `ionice`. Here is a cargo alias I use locally:

```
exec ionice -c 3 nice -n 20 chrt -i 0 cargo "$@"
```

and on Nix, you want to set your builder scheduling cvlasses to:

```
  nix = {
    extraOptions = ''
      experimental-features = nix-command flakes
    '';
    settings = {
      max-jobs = "auto";
      auto-optimise-store = true;
    };
    daemonIOSchedClass = "idle";
    daemonIOSchedPriority = 7;
    daemonCPUSchedPolicy = "idle";
  };
```

PR #4423, opened 2024-02-26
fix: check if in repl before running replit direnv check

Kodylow:
Gets rid of this warning I introduced when running direnv on local from the replit support PR, sorry @dpc :

```
> cd fedimint
nix-direnv: error current version v2.3.0 is older than the desired version v3.0.4
:bulb: Run 'just' for a list of available 'just ...' helper recipes
```


dpc (approved):

dpc:
Verified locally that the warning is gone.

elsirion (approved):

Issue #4421, opened 2024-02-26
Add ability to run full backwards-compatibility matrix


bradleystachurski:
CI is currently running backwards-compatibility tests where all binary versions are current except for one. We should add the ability to test the full matrix of versions. This is particularly useful as a sanity check prior to cutting a release.

PR #4425, opened 2024-02-27
chore: add full matrix option to back-compat

bradleystachurski:
Closes https://github.com/fedimint/fedimint/issues/4421


scripts/_common.sh
```diff
@@ -130,3 +126,60 @@ function use_gateway_binaries_for_version() {
...
+  versions=("$@")
+  version_matrix=()
+  for fed_version in "${versions[@]}"; do
+    for client_version in "${versions[@]}"; do
+      for gateway_version in "${versions[@]}"; do
+        if "$filter_fn" "$fed_version" "$client_version" "$gateway_version"; then
+          continue
+        fi
+
+        version_matrix+=("$fed_version $client_version $gateway_version")
```
dpc, line 142: Nit: Seems like we could echo it here right away.

scripts/_common.sh
```diff
@@ -130,3 +126,60 @@ function use_gateway_binaries_for_version() {
...
+      done
+    done
+  done
+
+  for version in "${version_matrix[@]}"; do
+    echo "$version"
+  done
+}
+
+function full_matrix_filter() {
```
dpc, line 152: Not clear what and why it does. Comment would be helpful. Also should it be called `is_...` ?

scripts/_common.sh
```diff
@@ -130,3 +126,60 @@ function use_gateway_binaries_for_version() {
 }
 export -f use_gateway_binaries_for_version
 
+function generate_matrix() {
```
dpc, line 129: Comment. :)

scripts/_common.sh
bradleystachurski: Good call 👍 

I also changed the filter logic to be inclusive instead of a `filter_not`, since that's a bit of a mind twister.

scripts/_common.sh
bradleystachurski: Let me know if the refactors for an inclusive filter plus the comments/docs clears this up (or if it's too verbose). Happy to keep iterating.

dpc (approved):

scripts/_common.sh
dpc: I got more comments than I could dream for. :)

elsirion (approved):
Inb4: rewriting Fedimint in bash :zany_face:

dpc:
> Inb4: rewriting Fedimint in bash 🤪

Maybe we should rewrite this script in https://rune-rs.github.io/ :thinking:

PR #4426, opened 2024-02-27
feat: enable modules to return confidential amounts 

joschisan:


joschisan:
I will reopen this pr later.

PR #4427, opened 2024-02-27
feat: Remove active gateway

m1sterc001guy:
Removes the "active gateway" concept from the client. Makes our tests more deterministic since every lightning operation explicitly accepts a gateway as a parameter. If a gateway is not supplied, it is assumed to be an internal payment.

 - Adds a gateway cache
 - Higher level apps like the CLI select a gateway from the cache to use when paying or creating an invoice.
 - Cache can be updated using `update_gateway_cache`. Higher level apps can call this periodically to stay in sync with gateway registrations.
 - Adds `UpdateGatewayCache` CLI command
 - Changes devimint tests to be backwards compatible.
 - Adds DB migration to remove active gateway entry

Closes https://github.com/fedimint/fedimint/issues/3844


justinmoon:
I think this closes https://github.com/fedimint/fedimint/issues/3844

m1sterc001guy:
Two problems make this a tad difficult:

1. Devimint tests need to be made backwards compatible.
2. Removing "active gateway" effectively removes caching of the Lightning Gateways, which regresses the latency tests. Perhaps a better approach is to change "active gateway" into a gateway cache, but still force the cli to pass a gateway id.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -826,76 +825,6 @@ impl LightningClientModule {
...
-            None => {
-                let gateways = self.fetch_registered_gateways().await?;
-                let num_gateways = gateways.len();
-
-                let vetted = gateways
-                    .clone()
-                    .into_iter()
-                    .filter(|g| g.vetted)
-                    .collect::<Vec<_>>();
-                if !vetted.is_empty() {
```
m1sterc001guy, line 853: Another question that arises is how we want to handle the "vetted" field. Previously we were filtering non-vetted gateways when selecting the active gateway. Perhaps we should just leave it up to the integrating client.

elsirion: I think vetted gateways should be prioritized if provided and online, if not we should try another one. That would mean the "ping the GW before making a payment" logic would move into the selection function so we can skip offline vetted gateways.

This might lead to the "gateway cache" becoming its own service in the background that autonomously fetches available gateways and tries to connect to them to figure out which one to use when asked

m1sterc001guy: I can take this another direction if you like, but I opted to keep the API at the `fedimint-client` level as simple as possible. Lightning operations always expect a gateway to be passed in and callers can acquire this gateway either using `select_gateway` or implementing their own logic in combination with `list_gateways` (they can select vetted gateways this way).

The goal of this PR was to remove non-determinism from our tests and I think implementing a fallback gateway causes some issues (it was the source of https://github.com/fedimint/fedimint/issues/4350#issue-2139759549), since the client might unexpectedly choose a different gateway.

I think higher level apps can implement this fallback or select vetted gateways if they want, but the way it is now allows our tests to be more straightforward

devimint/src/federation.rs
```diff
@@ -139,15 +140,20 @@ impl Client {
             .unwrap())
     }
 
+    // TODO(support:v0.2): remove
```
dpc, line 143: nit: I've seen it before and forgot to mention it, but "remove" here is not helpful at all. It would be more informative to say something 'in 0.3 we changed this and that' etc.

m1sterc001guy: Fixed, added more description to this comment

devimint/src/devfed.rs
```diff
@@ -52,6 +52,10 @@ pub async fn dev_fed(process_mgr: &ProcessManager) -> Result<DevFed> {
 
     info!(target: LOG_DEVIMINT, "federation and gateways started");
 
+    std::env::set_var("GWID_CLN", gw_cln.gateway_id().await?);
+    std::env::set_var("GWID_LND", gw_lnd.gateway_id().await?);
```
justinmoon, line 56: Was going to suggest prefixing with `FM_` but then realized that the only reason we did this in the past was to be able to copy all relevant envirment variables into the mprocs shell IIRC. But devimint fixes this for us now. So it would probably make more sense to remove the `FM_` prefixes!

fedimint-load-test-tool/src/common.rs
```diff
@@ -217,16 +218,17 @@ pub async fn gateway_pay_invoice(
     client: &ClientHandle,
     invoice: Bolt11Invoice,
     event_sender: &mpsc::UnboundedSender<MetricEvent>,
+    ln_gateway: Option<LightningGateway>,
 ) -> anyhow::Result<()> {
     let m = fedimint_core::time::now();
     let lightning_module = &client.get_first_module::<LightningClientModule>();
-    let gateway = lightning_module.select_active_gateway_opt().await;
+    //let gateway = lightning_module.select_active_gateway_opt().await;
```
justinmoon, line 225: dead code

fedimint-load-test-tool/src/main.rs
```diff
@@ -531,9 +531,9 @@ async fn get_user_client(
         invite_code.clone()
     };
     let (client, invite_code) = build_client(user_invite_code, user_db.as_ref()).await?;
-    if let Some(gateway_id) = gateway_id {
-        switch_default_gateway(&client, gateway_id).await?;
-    }
+    //if let Some(gateway_id) = gateway_id {
+    //    switch_default_gateway(&client, gateway_id).await?;
+    //}
```
justinmoon, line 536: dead code?

fedimint-load-test-tool/src/main.rs
```diff
@@ -786,13 +821,15 @@ async fn do_ln_circular_test_user_task(
             };
             while still_ontime().await {
                 let gateway_id = get_gateway_id(invoice_generation).await?;
-                switch_default_gateway(&client, &gateway_id).await?;
+                let ln_gateway = get_lightning_gateway(&client, Some(gateway_id)).await;
+                //switch_default_gateway(&client, &gateway_id).await?;
```
justinmoon, line 825: dead code?

fedimint-load-test-tool/src/common.rs
```diff
@@ -323,14 +325,16 @@ pub async fn cln_wait_invoice_payment(label: &str) -> anyhow::Result<()> {
     }
 }
 
-pub async fn switch_default_gateway(client: &ClientHandle, gateway_id: &str) -> anyhow::Result<()> {
+/*
+pub async fn switch_default_gateway(client: &ClientArc, gateway_id: &str) -> anyhow::Result<()> {
```
justinmoon, line 329: dead code?

justinmoon (dismissed):
Reviewed pretty thoroughly and it looks good. Only question is if we should delete the active gateway db key from code since we removed from the db in a migration.

justinmoon (approved):

dpc (approved):

devimint/src/devfed.rs
dpc: It's good to have project-specfic prefixes in long living env variables, as it makes chances of accidental conflicts lower. So we should keep `LM_`. @m1sterc001guy @justinmoon

elsirion (commented):
Apparently I never submitted these comments :see_no_evil:

modules/fedimint-ln-client/src/db.rs
```diff
@@ -12,9 +11,10 @@ use crate::OutgoingLightningPayment;
 #[repr(u8)]
 #[derive(Clone, EnumIter, Debug)]
 pub enum DbKeyPrefix {
-    LightningGateway = 0x28,
```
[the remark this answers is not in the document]
m1sterc001guy: Migration was added here https://github.com/fedimint/fedimint/pull/4427/commits/c66c4b108aa03c519575a194432b097af68f47f9

modules/fedimint-ln-client/src/db.rs
```diff
@@ -12,9 +11,10 @@ use crate::OutgoingLightningPayment;
...
     PaymentResult = 0x29,
     MetaOverrides = 0x30,
+    // Lightning Gateway cache records are stored the same way as the server, using 0x45
+    // [`fedimint_ln_common::db::DbKeyPrefix::LightningGateway`].
```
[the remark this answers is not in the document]
m1sterc001guy: It was mainly to re-use the existing code, since we already had a struct definition that we needed. I understand the concern though, we should probably move `db.rs` in `fedimint-ln-common` into the server and only use the db records from within the respective crates.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -812,17 +809,53 @@ impl LightningClientModule {
...
+    }
+
+    /// Returns all gateways that are currently in the gateway cache.
+    pub async fn list_gateways(&self) -> Vec<LightningGatewayAnnouncement> {
+        let mut dbtx = self.client_ctx.module_db().begin_transaction_nc().await;
+        dbtx.find_by_prefix(&LightningGatewayKeyPrefix)
+            .await
+            .map(|(_, gw)| gw.unanchor())
+            .collect::<Vec<_>>()
             .await
```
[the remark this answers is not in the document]
m1sterc001guy: I wanted to keep the API as simple as possible, with no unexpected network activity. I think higher level apps can make this decision themselves if they see that the cache is empty, they can call `update_gateway_cache`.

modules/fedimint-ln-client/src/db.rs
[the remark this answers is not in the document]
m1sterc001guy: Fixed here: https://github.com/fedimint/fedimint/pull/4567

Issue #3935, opened 2023-12-13
When gateway leaves a federation, we should ensure they have no funds remaining


justinmoon:
https://github.com/fedimint/fedimint/pull/3636 makes no such checks

kernelkind:
> Another idea: what happens if gateway only has dust? I think we should still be able to leave, otherwise they'd need to deposit and withdraw to get back to zero balance?

In this situation, the gateway would run the command `gateway-cli leave-fed --federation_id X --forced`

m1sterc001guy:
I don't think this is relevant anymore, it is useful to be able to leave a federation with funds still pegged into the federation, then re-join, to force a re-registration. Closing.

PR #4433, opened 2024-02-28
feat: handle gateway having funds when leaving fed

kernelkind:
When a gateway sends the command to leave a federation without the
forced flag, it will receive an error `GatewayError::UnexpectedState`.
When a gateway sends the command to leave the federation with the forced
flag, it will be able to leave and no error will be returned.

Closes: https://github.com/fedimint/fedimint/issues/3935

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -1530,6 +1530,95 @@ pub async fn gw_reboot_test(dev_fed: DevFed, process_mgr: &ProcessManager) -> Re
     Ok(())
 }
 
+async fn gw_leave_fed_test_one_gw(fed: &Federation, gw: &Gatewayd, gw_name: &str) -> Result<()> {
+    let fed_id = fed.calculate_federation_id().await;
+
+    let gateway_balance = cmd!(gw, "balance", "--federation-id={fed_id}")
+        .out_json()
+        .await?
+        .as_u64()
+        .unwrap();
+
+    assert_eq!(
+        gateway_balance, 0u64,
+        "Balance should have been zero for {gw_name} gateway."
+    );
+
+    info!(
+        LOG_DEVIMINT,
+        "{gw_name} attempting to leave federation unforced while having a zero balance."
+    );
+    cmd!(gw, "leave-fed", "--federation-id={fed_id}")
+        .run()
+        .await?;
+    info!(LOG_DEVIMINT, "{gw_name} leave successful.");
+
+    info!("Reconnecting {gw_name} gateway to federation.");
+    gw.connect_fed(fed).await?;
+
+    info!("Adding payment to {gw_name} gateway.");
+    fed.pegin_gateway(99_999, gw).await?;
+
+    let gateway_balance = cmd!(gw, "balance", "--federation-id={fed_id}")
+        .out_json()
+        .await?
+        .as_u64()
+        .unwrap();
+
+    assert_ne!(
+        gateway_balance, 0u64,
+        "{gw_name} balance should have been non-zero."
+    );
+    info!(LOG_DEVIMINT, "{gw_name} balance is non-zero.");
+
+    info!(
+        LOG_DEVIMINT,
+        "{gw_name} attempting to leave federation unforced while having a non-zero balance."
+    );
+    cmd!(gw, "leave-fed", "--federation-id={fed_id}")
+        .run()
+        .await
+        .expect_err("{gw_name} should have errored, gateway is trying to leave with a non-zero balance without the --forced flag.");
+    info!(
+        LOG_DEVIMINT,
+        "{gw_name} gateway did not leave federation since it has a non-zero balance and the --forced flag was not used."
+    );
+
+    info!(
+        LOG_DEVIMINT,
+        "{gw_name} attempting to leave federation forced while having a non-zero balance."
+    );
+    cmd!(gw, "leave-fed", "--federation-id={fed_id}", "--forced=true")
+        .run()
+        .await?;
+    info!(
+        LOG_DEVIMINT,
+        "{gw_name} successfully left federation forced while having a non-zero balance."
+    );
+
+    Ok(())
+}
+
+pub async fn gw_leave_fed_test(dev_fed: DevFed) -> Result<()> {
+    #[allow(unused_variables)]
+    let DevFed {
+        bitcoind,
+        cln,
+        lnd,
+        fed,
+        gw_cln,
+        gw_lnd,
+        electrs,
+        esplora,
+    } = dev_fed;
+
+    info!(LOG_DEVIMINT, "starting gw_leave_fed_test");
+    gw_leave_fed_test_one_gw(&fed, &gw_cln, "CLN").await?;
+    gw_leave_fed_test_one_gw(&fed, &gw_lnd, "LND").await?;
+    info!(LOG_DEVIMINT, "gw_leave_fed_test: success");
+    Ok(())
+}
+
 pub async fn do_try_create_and_pay_invoice(
     gw: &Gatewayd,
     client: &Client,
@@ -2175,6 +2264,8 @@ pub enum TestCmd {
     /// `devfed` then reboot gateway daemon for both CLN and LND. Test
     /// afterward.
     GatewayRebootTest,
+    /// `devfed` then attempts to leave with funds
+    GatewayLeaveFedTest,
     /// `devfed` then tests if the recovery tool is able to do a basic recovery
     RecoverytoolTests,
     /// `devfed` then spawns faucet for wasm tests
@@ -2256,6 +2347,14 @@ pub async fn handle_command(cmd: TestCmd, common_args: CommonArgs) -> Result<()>
             let (process_mgr, _) = setup(common_args).await?;
             let dev_fed = dev_fed(&process_mgr).await?;
             gw_reboot_test(dev_fed, &process_mgr).await?;
+        }
+        TestCmd::GatewayLeaveFedTest => {
+            let fedimint_gatewayd_version = crate::util::Gatewayd::version_or_default().await;
+            if VersionReq::parse(">=0.3.0-alpha")?.matches(&fedimint_gatewayd_version) {
+                let (process_mgr, _) = setup(common_args).await?;
+                let dev_fed = dev_fed(&process_mgr).await?;
+                gw_leave_fed_test(dev_fed).await?;
+            }
         }
         TestCmd::RecoverytoolTests => {
             let (process_mgr, _) = setup(common_args).await?;
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -965,6 +965,9 @@ impl Gateway {
         Err(GatewayError::Disconnected)
     }
 
+    /// Attempt to leave the federation. If the gateway balance is non-zero and
+    /// the `LeaveFedPayload.force_leave` flag is not set, a
+    /// `GatewayError::NonZeroBalanceLeaveError` error will be returned.
     pub async fn handle_leave_federation(
         &mut self,
         payload: LeaveFedPayload,
@@ -977,6 +980,10 @@ impl Gateway {
             let federation_info = self
                 .make_federation_info(client.value(), payload.federation_id)
                 .await;
+
+            if payload.force_leave != Some(true) && federation_info.balance_msat.msats != 0u64 {
+                return Err(GatewayError::NonZeroBalanceLeaveError);
+            }
 
             let keypair = dbtx
                 .get_value(&GatewayPublicKey)
@@ -1543,6 +1550,8 @@ pub enum GatewayError {
     InsufficientFunds,
     #[error("Federation already connected")]
     FederationAlreadyConnected,
+    #[error("Gateway has non-zero balance")]
+    NonZeroBalanceLeaveError,
 }
 
 impl IntoResponse for GatewayError {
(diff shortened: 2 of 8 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -1017,6 +1017,18 @@ impl Gateway {
             federation_info
         };
 
+        let balance = self
+            .handle_balance_msg(BalancePayload {
```
m1sterc001guy, line 1021: Instead of reconstructing `BalancePayload`, you can just call `get_balance` on the client at line 1003

elsirion:
We should be able to override this. We sometimes needed to leave+rejoin as a quick fix to some bug. Since we keep the client directory around there is little risk for loss of funds.

justinmoon:
Before merging this we should make sure that `gateway-cli leave-federation` is updated to support this change. Might be good to have coverage for this command in devimint tests.

justinmoon:
Another idea: what happens if gateway only has dust? I think we should still be able to leave, otherwise they'd need to deposit and withdraw to get back to zero balance?

kernelkind:
I've got a bunch of changes to add for this MR including a devimint test and backend test, but I've got a question before I push any changes.

https://github.com/fedimint/fedimint/blob/c1cb0e6207779dd8441f4c3798caa6471e2b15ec/gateway/ln-gateway/src/rpc/rpc_server.rs#L182-L189

I see that if I were to throw an error in `handle_leave_federation`, the client has no way of receiving it, since the router panics on error. I need to somehow let the client know that the request to leave the federation was denied (because the client didn't use the `--forced` flag & has remaining funds). 

There are a couple ways I'm thinking of making changes. Here's the first:
```
async fn leave_fed(
    Extension(mut gateway): Extension<Gateway>,
    Json(payload): Json<LeaveFedPayload>,
) -> Result<impl IntoResponse, GatewayError> {
    let fed = gateway.handle_leave_federation(payload).await;
    Ok(Json(json!(fed.ok())))
}
```

Return an `Option<FederationInfo>` instead of returning `FederationInfo`. This way, if the client sees that the result is `None`, it knows the federation didn't leave successfully. This isn't a great implementation in my opinion, the client doesn't see what error was returned.

Another, could be:
In `gateway/ln-gateway/src/rpc/rpc_server.rs`:
```
async fn leave_fed(
    Extension(mut gateway): Extension<Gateway>,
    Json(payload): Json<LeaveFedPayload>,
) -> Result<impl IntoResponse, GatewayError> {
    let fed = gateway.handle_leave_federation(payload).await;
    let res = match fed {
        Ok(fed) => FederationLeaveResult { info: Some(fed), error: None },
        Err(e) => FederationLeaveResult { info: None, error: Some(e.to_string()) },
    };
    Ok(Json(json!(res)))
}
```
and in `gateway/ln-gateway/src/rpc/mod.rs`:
```
#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
pub struct FederationLeaveResult {
    pub info: Option<crate::FederationInfo>,
    pub error: Option<String>,
}
```

Let me know your thoughts on how the server can inform the client that leaving was unsuccessful.

m1sterc001guy:
> I see that if I were to throw an error in handle_leave_federation, the client has no way of receiving it, since the router panics on error.

I'm confused, where are you seeing this? The gateway should not panic on any error, if it does, that's a bug.

`GatewayError` is mapped to a `StatusCode` in `IntoResponse`

```
impl IntoResponse for GatewayError {
    fn into_response(self) -> Response {
        // For privacy reasons, we do not return too many details about the failure of
        // the request back to the client to prevent malicious clients from
        // deducing state about the gateway/lightning node.
        let (error_message, status_code) = match self {
            GatewayError::OutgoingPaymentError(_) => (
                "Error while paying lightning invoice. Outgoing contract will be refunded."
                    .to_string(),
                StatusCode::BAD_REQUEST,
            ),
            GatewayError::Disconnected => (
                "The gateway is disconnected from the Lightning Node".to_string(),
                StatusCode::NOT_FOUND,
            ),
            _ => (
                "An internal gateway error occurred".to_string(),
                StatusCode::INTERNAL_SERVER_ERROR,
            ),
        };
        let mut err = Cow::<'static, str>::Owned(error_message).into_response();
        *err.status_mut() = status_code;
        err
    }
}
```

You should be able to just return a `GatewayError` and check for that on the client side.

kernelkind:
> I'm confused, where are you seeing this? The gateway should not panic on any error, if it does, that's a bug.

Ah I was mistaken. I saw this line:
https://github.com/fedimint/fedimint/blob/348c4d3a7a56c2e89236383664d609e8f3ef3399/gateway/ln-gateway/src/rpc/rpc_server.rs#L70

And the implementation of the route method:
```
    pub fn route(mut self, path: &str, method_router: MethodRouter<S, B>) -> Self {
        panic_on_err!(self.path_router.route(path, method_router));
        self
    }
```

and assumed that was causing the panicking in my tests. I see now that's not the case

kernelkind:
I can remove the `scripts/tests/backend-leave-fed-test.sh` and `scripts/tests/gateway-leave-fed-test.sh`, I just used them for easier testing

kernelkind:
> Another idea: what happens if gateway only has dust? I think we should still be able to leave, otherwise they'd need to deposit and withdraw to get back to zero balance?

In this situation, the gateway would run the command `gateway-cli leave-fed --federation_id X --forced`

elsirion (dismissed):
Great PR, very nice testing! :tada:

devimint/src/tests.rs
```diff
@@ -1504,6 +1504,95 @@ pub async fn gw_reboot_test(dev_fed: DevFed, process_mgr: &ProcessManager) -> Re
...
+        LOG_DEVIMINT,
+        "{gw_name} attempting to leave federation unforced while having a non-zero balance."
+    );
+    cmd!(gw, "leave-fed", "--federation-id={fed_id}")
+        .run()
+        .await
+        .expect_err("{gw_name} should have errored, gateway is trying to leave with a non-zero balance without the --forced flag.");
+    info!(
+        LOG_DEVIMINT,
+        "{gw_name} unsuccessfully left federation unforced while having a non-zero balance."
```
m1sterc001guy, line 1558: Nit: this statement is a bit confusing to me, I would expect "unsuccessfully" to indicate an error

gateway/ln-gateway/src/lib.rs
```diff
@@ -1610,6 +1619,7 @@ impl IntoResponse for GatewayError {
                 "The gateway is disconnected from the Lightning Node".to_string(),
                 StatusCode::NOT_FOUND,
             ),
+            GatewayError::UnexpectedState(msg) => (msg, StatusCode::EXPECTATION_FAILED),
```
m1sterc001guy, line 1622: Can we make a new `GatewayError` instead of re-using `UnexpectedState`? I think `UnexpectedState` should still be a 500 error code, since that was originally made to indicate that a state machine did not execute correctly.

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1393,6 +1400,71 @@ async fn test_gateway_can_leave_connected_federations() -> anyhow::Result<()> {
...
+                .connect_federation(ConnectFedPayload {
+                    invite_code: invite2.to_string(),
+                })
+                .await
+                .unwrap();
+
+            let info = rpc.get_info().await.unwrap();
+            assert_eq!(info.federations.len(), 2);
+            assert_eq!(
+                info.channels.unwrap().keys().cloned().collect::<Vec<u64>>(),
```
m1sterc001guy, line 1464: Nit: we're trying to reduce the number of `unwrap` in the code, lets change this to an `expect`

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1393,6 +1400,71 @@ async fn test_gateway_can_leave_connected_federations() -> anyhow::Result<()> {
                 vec![3, 4]
             );
 
+            // send sats to first federation
```
m1sterc001guy, line 1403: Great test!

scripts/tests/backend-leave-fed-test.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
m1sterc001guy, line 1: I don't think this script is necessary, the gateway leaving test should be the same regardless of whether the lightning node is mocked or not

scripts/tests/gateway-leave-fed-test.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
m1sterc001guy, line 1: We can keep this one, should be convenient for running the devimint test :)

gateway/ln-gateway/tests/integration_tests.rs
kernelkind: I used `.expect_or_else` in some places because clippy didn't like putting `format` inside the `expect` modifier.

gateway/ln-gateway/src/lib.rs
kernelkind: I made `GatewayError::NonZeroBalanceLeaveError` but still used `StatusCode::EXPECTATION_FAILED`, not sure if there's a better status code to use for this

scripts/tests/gateway-leave-fed-test.sh
dpc, line 1: Will need to plug it to test-ci-all.sh

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Works for me!

m1sterc001guy:
I think you need to make the `forced` flag optional, otherwise the backwards compatibility tests will fail. Otherwise its looking good to me.

kernelkind:
I made it optional, but the backwards compatibility test still fails. It's weird, why would it fail before supplying the `--forced=true` flag? I would understand if it failed because the test was supplying the forced flag to an older version of the gateway-cli that didn't support it

m1sterc001guy:
Looks like the endpoint is returning null instead of `FederationInfo` like expected.

```
00:00:11 Error: error decoding response body: invalid type: null, expected struct FederationInfo at line 1 column 4
00:00:11 
00:00:11 Caused by:
00:00:11     invalid type: null, expected struct FederationInfo at line 1 column 4
```

The logs in CI only show the output of the devimint test, but not the logs of each daemon (like the gateway). I'd recommend running the test locally or trying to repro locally. Something is probably wrong in `handle_leave_federation`

kernelkind:
A less elegant solution that I think wouldn't break the backwards compatibility tests is to revert the `LeaveFed` command and create a new one called `LeaveFedForced`. And `LeaveFedPayload` would be reverted and a new payload `LeaveFedForcedPayload` would also be created. Would this be acceptable?

m1sterc001guy:
Can you try to make `force_leave` in `LeaveFedPayload` an `Option<bool>` instead of just `bool`? I believe `Deserialize` will default to None if no value is present in the serialized struct, which should fix the backwards compatibility issues. A value of `None` can just be interpreted as false.

kernelkind:
This method in `v0.2.1` returns `Result<()>` instead of `Result<FederationInfo>`:
https://github.com/fedimint/fedimint/blob/a8422b84102ab5fc768307215d5b20d807143f27/gateway/ln-gateway/src/lib.rs#L874-L885


 so in the test `gateway_leave_fed_test FM: current CLI: current GW: v0.2.1`,  the `v0.2.1` rpc_server for gateway sends `Ok(())`
https://github.com/fedimint/fedimint/blob/a8422b84102ab5fc768307215d5b20d807143f27/gateway/ln-gateway/src/rpc/rpc_server.rs#L143-L149


But current `rpc_client` `leave_federation` expects a `GatewayRpcResult<FederationInfo>`, not `GatewayRpcResult<()>`:

https://github.com/fedimint/fedimint/blob/580a95b1f24b248c3f75429c992416f082a79098/gateway/ln-gateway/src/rpc/rpc_client.rs#L83-L89

So now the error makes sense:
`00:00:15     invalid type: null, expected struct FederationInfo at line 1 column 4`

How should I proceed?

m1sterc001guy:
@okjodom added versioning for the gateway API awhile back. @okjodom could you advise on how to modify `leave_federation` so it is backwards compatible?

okjodom:
> @okjodom added versioning for the gateway API awhile back. @okjodom could you advise on how to modify `leave_federation` so it is backwards compatible?

I suppose we can cherry pick https://github.com/fedimint/fedimint/pull/4275/commits/4aa64957d9faf0fe204f7d36cd3f44a0d4151284 to `v0.2.1` so the whole test  matrix is compatible. Otherwise we should only consider `v0.2.2` and later, in the tests covering this api

kernelkind:
If we wanted to use `v0.2.2` it looks like 4aa6495 would need to be chery-picked as well since it doesn't have the `Result<FederationInfo>` return type for `handle_leave_federation` etc. It doesn't matter to me which we use, but it would be nice to get this PR across the finish line at some point

elsirion:
> If we wanted to use `v0.2.2` it looks like [4aa6495](https://github.com/fedimint/fedimint/commit/4aa64957d9faf0fe204f7d36cd3f44a0d4151284) would need to be chery-picked here as well

Not worth it, let's only run this for current and future versions, it's a new feature.

elsirion:
@kernelkind I don't think the test failure is due to your code, we just didn't test leaving the federation at all earlier and apparently broke compatibility in #4275: before the PR the API returned `()` while after the PR it returned `FederationInfo`. Luckily, just [decoding it as `Option<T>`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f31dc2e7cc8131af9615580a5fd46023) should solve the problem.

elsirion:
Feel free to re-open or others could take over, closing for now.

PR #4437, opened 2024-02-29
fix: support fees for primary module inputs

joschisan:


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1485,10 +1501,16 @@ async fn select_notes_from_stream<Note>(
                 "notes are not sorted in descending order"
             );
             previous_amount = Some(note_amount);
-            match note_amount.cmp(&pending_amount) {
+
+            if note_amount < note_fee {
```
elsirion, line 1505: Should this be a less or equal?

modules/fedimint-mint-tests/tests/tests.rs
```diff
@@ -196,7 +216,7 @@ async fn sends_ecash_out_of_band_cancel() -> anyhow::Result<()> {
     // FIXME: UserCanceledSuccess should mean the money is in our wallet
     for _ in 0..200 {
         sleep_in_test("sats not in wallet yet", Duration::from_millis(100)).await;
-        if client.get_balance().await == sats(1000) {
+        if client.get_balance().await >= sats(1000) - EXPECTED_MAXIMUM_FEE {
```
elsirion, line 219: This approximation ain't great, but probably the best we can do for now. We need APIs to get the fees paid for operations/transactions.

elsirion (approved):
Currently, we will be generating lots of dust as change. I think that's ok for now. I can imagine us introducing a load-based fee system which would allow clients to consolidate dust during low load times. Otherwise we should allow overfunded transactions and just overpay fees instead of generating dust change.

dpc (approved):

Issue #4434, opened 2024-02-29
vetted_gateways and meta_override_url use different encodings


douglaz:
Most of [meta fields](https://github.com/fedimint/fedimint/blob/bf2763463e8157204e9d8a4464c58b8ebddfd60f/docs/meta_fields/README.md) are used only by third party clients, but we have two that are interpreted by fedimint-client:

1. vetted_gateways: https://github.com/fedimint/fedimint/blob/bf2763463e8157204e9d8a4464c58b8ebddfd60f/modules/fedimint-ln-client/src/lib.rs#L1034
2. meta_override_url: https://github.com/fedimint/fedimint/blob/bf2763463e8157204e9d8a4464c58b8ebddfd60f/modules/fedimint-ln-client/src/lib.rs#L1006

Unfortunately they use conflicting encodings. The only way they work now is by using something like:
```json
"meta": {
    "vetted_gateways": ["02db27c3adaa8dfbd53b3bbe3ef79758e23e974e6a6a168f9726bc1722166dd04a","03c79f33bbae76c058fc4951658a1dfd1f400531456d1a2718c9814c8bf684a770"],
    "meta_override_url": "\"https://meta.dev.fedibtc.com/meta.json\""
  },
```

But this mixes json and json-encoded strings. 

So it should be either:

```json
"meta": {
    "vetted_gateways": "[\"02db27c3adaa8dfbd53b3bbe3ef79758e23e974e6a6a168f9726bc1722166dd04a\",\"03c79f33bbae76c058fc4951658a1dfd1f400531456d1a2718c9814c8bf684a770\"]",
    "meta_override_url": "\"https://meta.dev.fedibtc.com/meta.json\""
  },
```
Or:
```json
"meta": {
    "vetted_gateways": ["02db27c3adaa8dfbd53b3bbe3ef79758e23e974e6a6a168f9726bc1722166dd04a","03c79f33bbae76c058fc4951658a1dfd1f400531456d1a2718c9814c8bf684a770"],
    "meta_override_url": "https://meta.dev.fedibtc.com/meta.json"
  },
```

A related issue is https://github.com/fedimint/fedimint/issues/4432

@elsirion What's the correct convention?

elsirion:
The correct one is:

```json
"meta": {
[...]

Issue #4432, opened 2024-02-28
Inconsistent encoding of federation_name


douglaz:
According to the [docs](https://github.com/fedimint/fedimint/blob/master/docs/meta_fields/README.md):
https://github.com/fedimint/fedimint/blob/bf2763463e8157204e9d8a4464c58b8ebddfd60f/docs/meta_fields/README.md?plain=1#L23

But on a new federation we usually see:
```json
  "meta": {
    "federation_name": "myfederation",
    "meta_override_url": "\"https://mymeta.com/meta.json\""
  }
```

Instead of:
```json
  "meta": {
    "federation_name": "\"myfederation\"",
    "meta_override_url": "\"https://mymeta.com/meta.json\""
  }
```

Other examples from documentation/code:
https://github.com/fedimint/fedimint/blob/bf2763463e8157204e9d8a4464c58b8ebddfd60f/docs/tutorial.md?plain=1#L41
https://github.com/fedimint/fedimint/blob/bf2763463e8157204e9d8a4464c58b8ebddfd60f/devimint/src/federation.rs#L640

elsirion:
We made our parsing lenient to allow bare strings as a special case.

Issue #3648, opened 2023-11-19
`meta_override_url` is unnecessarily wrapped in quotes


elsirion:
![Screenshot 2023-11-19 at 16-47-32 feat clients use vetted gateway meta field in selecting gateway by okjodom · Pull Request #3533 · fedimint_fedimint](https://github.com/fedimint/fedimint/assets/84478054/d2348b59-3d2a-40a9-aee3-000f8434753f)

See how we have to call `json!(…).to_string()` instead of writing the plain string: https://github.com/fedimint/fedimint/pull/3533

okjodom:
We need the url to be a valid json string when we parse it later.

`"\"https://getln.xyz/fedimint/meta\"".to_string()` works just as well, but plain `"https://getln.xyz/fedimint/meta"` will fail json parse with `serde_json::from_str()`

elsirion:
I see, so currently there are two functions `meta` and `meta_raw`, the latter returns the raw string slice. But I see that this API is confusing. The problem currently is that some string meta fields use the raw format (e.g. federation name) and others the JSON format (e.g. meta_override_url). It feels like there should be a standard?

okjodom:
Could we standardise all meta fields as json strings?

elsirion:
> Could we standardise all meta fields as json strings?

@wbobeirne @otech47 would you be ok with that? I guess you already need to impl JSON decoding anyway, so if e.g. the federation name was also awkwardly wrapped this wouldn't be too much of a problem from the code side (just not aesthetically pleasing when looking at the config).

wbobeirne:
The logic I've been going with with meta is something like:
```ts
try {
  // Try to parse it as JSON, could be string | number | boolean | object
  const value = JSON.parse(meta[key])
  return value
} catch {
  // If parsing fails, return it as a plain old string | undefined
  return meta[key]
}
```
So in that case, something being extra string wrapped or just a plain string would work fine. Can't guarantee that all implementations out there will be like this, but it's definitely the safest way to do it.

PR #4438, opened 2024-02-29
fix: make `String`s in meta backwards compatible

elsirion:
For a long time we allowed plain strings in meta fields, i.e. without some ugly additional JSON quotation, but removed that bug/feature. Turns out that's a backwards incompatible change.

Related to #3648, #3699, #4432 and #4434.

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -198,16 +198,30 @@ impl ClientConfig {
     }
 
     /// Get the value of a given meta field
-    pub fn meta<V: serde::de::DeserializeOwned>(
+    pub fn meta<V: serde::de::DeserializeOwned + 'static>(
         &self,
         key: &str,
     ) -> Result<Option<V>, anyhow::Error> {
         let Some(str_value) = self.global.meta.get(key) else {
             return Ok(None);
         };
-        serde_json::from_str(str_value)
+        let res = serde_json::from_str(str_value)
             .map(Some)
-            .context(format!("Decoding meta field '{key}' failed"))
+            .context(format!("Decoding meta field '{key}' failed"));
+
+        // In the past we encoded some string fields as "just a string" without quotes,
+        // this code ensures that old meta values still parse since config is hard to
+        // change
+        if res.is_err() && std::any::TypeId::of::<V>() == std::any::TypeId::of::<String>() {
+            let string_ret = Box::new(str_value.clone());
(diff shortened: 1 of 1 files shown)


fedimint-core/src/config.rs
```diff
@@ -198,16 +198,30 @@ impl ClientConfig {
...
+        // In the past we encoded some string fields as "just a string" without quotes,
+        // this code ensures that old meta values still parse since config is hard to
+        // change
+        if res.is_err() && std::any::TypeId::of::<V>() == std::any::TypeId::of::<String>() {
+            let string_ret = Box::new(str_value.clone());
+            let ret: Box<V> = unsafe {
+                // We can transmute a String to V because we know that V==String
+                std::mem::transmute(string_ret)
+            };
+            Ok(Some(*ret))
```
elsirion, line 221: I wasn't able to do this with a plain `V`, the compiler complained about unknown size, this seemed like the easiest fix.

elsirion: (@dpc used a similar unsafe trick elsewhere iirc, it's not new)

douglaz (approved):
Ugly but works

fedimint-core/src/config.rs
bradleystachurski: > (@dpc used a similar unsafe trick elsewhere iirc, it's not new)

Yep, there's some insightful discussion related to this pattern in the PR that introduced it https://github.com/fedimint/fedimint/pull/4145

This looks good to me, but I'll defer approval to folks with more context.

fedimint-core/src/config.rs
```diff
@@ -198,16 +198,30 @@ impl ClientConfig {
...
+            .context(format!("Decoding meta field '{key}' failed"));
+
+        // In the past we encoded some string fields as "just a string" without quotes,
+        // this code ensures that old meta values still parse since config is hard to
+        // change
+        if res.is_err() && std::any::TypeId::of::<V>() == std::any::TypeId::of::<String>() {
+            let string_ret = Box::new(str_value.clone());
+            let ret: Box<V> = unsafe {
+                // We can transmute a String to V because we know that V==String
+                std::mem::transmute(string_ret)
```
dpc, line 219: Putting bounds here explicitly would make paranoid people feel a bit better about it.

dpc (approved):

PR #4439, opened 2024-02-29
chore: include binary name in test version string

bradleystachurski:
It's not obvious which versions map to which binaries when looking through the devimint logs.

```
## FAILED: latency_test_ln_send (current v0.2.1 current)
```

This will make it obvious at a glance what version each binary maps to.

```
## FAILED: latency_test_ln_send (FM: current, CLI: v0.2.1, GW: current)
```

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -53,7 +53,7 @@ function run_test_for_versions() {
     export FM_BACKWARDS_COMPATIBILITY_TEST=1
     # run back-compat tests with 4/4 setup
     export FM_OFFLINE_NODES=0
-    export FM_RUN_TEST_VERSIONS="$fed_version $client_version $gateway_version"
+    export FM_RUN_TEST_VERSIONS="FM: $fed_version, CLI: $client_version, GW: $gateway_version"
   else
     # default to run current tests in 3/4 setup
     export FM_OFFLINE_NODES=1


scripts/_common.sh
```diff
@@ -53,7 +53,7 @@ function run_test_for_versions() {
     export FM_BACKWARDS_COMPATIBILITY_TEST=1
     # run back-compat tests with 4/4 setup
     export FM_OFFLINE_NODES=0
-    export FM_RUN_TEST_VERSIONS="$fed_version $client_version $gateway_version"
+    export FM_RUN_TEST_VERSIONS="fed_version: $fed_version, client_version: $client_version, gateway_version: $gateway_version"
```
elsirion, line 56: nit: let's make it shorter: `FM`, `CLI`, `GW`

elsirion (approved):

scripts/_common.sh
bradleystachurski: Ooh, good call 👍

elsirion (approved):

dpc (approved):

dpc:
@bradleystachurski You reminded me that I wanted to figure out how to make backwardcompt `parallel` print the output only of the failing runs, because the output there is getting massive and painful to download and look through.

bradleystachurski:
@dpc Yeah that's quite annoying, that's for calling that out. Tracking in https://github.com/fedimint/fedimint/issues/4443

PR #4442, opened 2024-02-29
fix: persist xmpp message archives

Kodylow:
The data_path where xmpp archives messages in our prosody config writer is writing to /var/lib/prosody instead of the prosody_datadir volume that gets accessed by the xmpp service


elsirion (approved):

elsirion:
Fixin since it's broken, that XMPP stuff should ideally move out of this repo though :)

justinmoon (approved):

Issue #4444, opened 2024-02-29
Allocate client databases prefix for app use and other purposes


dpc:
Users of `fedimint_client::Client` should not be allowed to just write random stuff, at least not without some Rust-API enforced warning that they should not be doing it.

Instead we should designate a db prefix under which they can store whatever stuff they desire, and build APIs to returned an `IsolatedDatabase` that locks them into that prefix.

We should also allocate range of prefixes for this type of use.

See #4445 for initial commit.

PR #4445, opened 2024-02-29
chore: allocate client db prefix range for external use

dpc:
Re #4444

I mentioned it before, but it was never followed through.

Sometimes applications using client will want to store some data, in client database. Potentially because it's already there and Fedimint provides all the infrastructure to use it, but also due to atomic transaction requirements.

We should wrap it up in a nicer interface in the future, but for now, just officially allocate it and document it in the code.

Not going to lie - while the exact range is somewhat arbitrary, and probably a smaller one would be enough, Fedi app has already started using this range, so it will make our life easier if we reserve it, and we will internally migrate to use only designated prefix when Fedimint has proper support for it, and we have some time to implement a migration.

While I think a single prefix would be enough for this exact use case, we (the Fedimint project) might come up with a need for other external use scenarios (you can't predict that stuff ahead of time), so might as well reserve a range.

New users (app integrators) are encouraged to just use the single designated user data prefix. If needs arise, we can discuss and allocate new ones in the reserved range.

### DIFF
--- a/fedimint-client/src/db.rs
+++ b/fedimint-client/src/db.rs
@@ -38,6 +38,26 @@ pub enum DbKeyPrefix {
     ClientInviteCode = 0x30,
     ClientInitState = 0x31,
     ClientMetadata = 0x32,
+    /// Arbitrary data of the applications integrating Fedimint client and
+    /// wanting to store some Federation-specific data in Fedimint client
+    /// database.
+    ///
+    /// New users are encouraged to use this single prefix only.
+    //
+    // TODO: https://github.com/fedimint/fedimint/issues/4444
+    //       in the future, we should make all global access to the db private
+    //       and only expose a getter returning isolated database.
+    UserData = 0xb0,
+    /// Prefixes between 0xb1..=0xcf shall all be considered allocated for
+    /// historical and future external use
+    ExternalReservedStart = 0xb1,
+    /// Prefixes between 0xb1..=0xcf shall all be considered allocated for
+    /// historical and future external use
+    ExternalReservedEnd = 0xcf,
+    /// 0xd0.. reserved for Fedimint internal use
+    InternalReservedStart = 0xd0,
+    /// Per-module instance data
+    ModuleGlobalPrefix = 0xff,
 }
 
 impl std::fmt::Display for DbKeyPrefix {


dpc:
@benthecarman I hope this is fine with Mutiny Fedimint integration. If you already use different keys etc. please let me know and we can find something that works for everyone.

I'm not aware of any other integration that might be affected, I kind of doubt there are any. After this PR at least we'll have official and code-documented guidance on the matter.

TonyGiorgio:
I'm not sure what this means or how this would affect us. Could you clarify?

dpc:
> I'm not sure what this means or how this would affect us. Could you clarify?

I don't if Mutiny app stores anything custom in the Fedimint client rocksdb database. If you don't and don't plan to, it doesn't affect you at all. If you do (before or in the future), you should probably use the reserved prefix from this PR.

TonyGiorgio:
I see, no we don't. We don't use the rocksdb yet and have a simple indexeddb wrapper over the fedimint db interfaces and only use that for the normal fedimint data.

fedimint-client/src/db.rs
```diff
@@ -38,6 +38,22 @@ pub enum DbKeyPrefix {
...
+    // TODO: https://github.com/fedimint/fedimint/issues/4444
+    //       in the future, we should make all global access to the db private
+    //       and only expose a getter returning isolated database.
+    UserData = 0xb0,
+    /// Prefixes between 0xb1-0xff shall all be considered allocated for
+    /// historical and future external use
+    ExternalReservedStart = 0xb1,
+    /// Prefixes between 0xb1-0xff shall all be considered allocated for
+    /// historical and future external use
+    ExternalReservedEnd = 0xff,
```
m1sterc001guy, line 56: This is exclusive or inclusive? iirc we use 0xff to denote a global database entry (i.e not in a module)

m1sterc001guy (approved):

fedimint-client/src/db.rs
dpc: I missed that. Ha.

OK. So maybe `0xb0..=0xcf` for external and `0xd0..=0xff` for internal use.

fedimint-client/src/db.rs
```diff
@@ -38,6 +38,26 @@ pub enum DbKeyPrefix {
...
+    //       and only expose a getter returning isolated database.
+    UserData = 0xb0,
+    /// Prefixes between 0xb1..=0xcf shall all be considered allocated for
+    /// historical and future external use
+    ExternalReservedStart = 0xb1,
+    /// Prefixes between 0xb1..=0xcf shall all be considered allocated for
+    /// historical and future external use
+    ExternalReservedEnd = 0xcf,
+    /// 0xd0.. reserved for Fedimint internal use
+    InternalReservedStart = 0xd0,
```
elsirion, line 58: Why would we have an internal reserve? I'd say unless something is marked as "External" integrators don't have any business writing in our DB.

dpc: Yeah. It's not necessary, but let's book it so even we are not adding anything there so we have a block of contiguous prefixes for special purposes in the future etc.

I basically started by booking the whole range up to 0xff for users, but 0xff is already used for special stuff (modules) so did it this way.

elsirion (approved):

m1sterc001guy (approved):

fedimint-client/src/db.rs
m1sterc001guy: Good with me

PR #4447, opened 2024-03-01
WIP: generate small e-cash bags for rough amounts

elsirion:


elsirion:
Idea from call: since it's too hacky it shouldn't live in the main module code base, we could put it into a separate crate or feature-flag it. @dpc

PR #4448, opened 2024-03-01
chore: update log level for mprocs

bradleystachurski:
Using the unknown module in mprocs will cause warning logs which can be unsettling for users. This change updates to a `debug!` log, which suppresses these warning.

See: [discord](https://discord.com/channels/990354215060795454/990354215878688860/1213174844275826728)

```
bash-5.2$ fedimint-cli info
2024-03-01T20:31:51.184211Z  WARN fedimint_client: Detected configuration for unsupported module id: 3, kind: unknown
2024-03-01T20:31:51.186137Z  WARN fedimint_client: Module kind unknown of instance 3 not found in module gens, skipping
{
...
```

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1714,7 +1714,7 @@ impl ClientBuilder {
             for (module_id, module_cfg) in client_config.modules {
                 let kind = module_cfg.kind.clone();
                 let Some(init) = self.module_inits.get(&kind) else {
-                    warn!("Detected configuration for unsupported module id: {module_id}, kind: {kind}");
+                    debug!("Detected configuration for unsupported module id: {module_id}, kind: {kind}");
                     continue;
                 };
 
@@ -1979,7 +1979,7 @@ impl ClientBuilder {
             for (module_instance_id, module_config) in config.modules.clone() {
                 let kind = module_config.kind().clone();
                 let Some(module_init) = self.module_inits.get(&kind).cloned() else {
-                    warn!("Module kind {kind} of instance {module_instance_id} not found in module gens, skipping");
+                    debug!("Module kind {kind} of instance {module_instance_id} not found in module gens, skipping");
                     continue;
                 };
 


dpc:
Maybe we should change the logging level then. Same unsettling logs will be visible to normal users.

dpc:
Part of they point of having it in mprocs is making sure the ux is right as well.

dpc:
Maybe instead of this we should info! a summary of supported, unsupported etc.

bradleystachurski:
> Maybe we should change the logging level then. Same unsettling logs will be visible to normal users.

I'm interpreting normal users to mean users in production instead of mprocs. In that env, I think keeping the warning logs is the correct UX.

> Maybe instead of this we should info! a summary of supported, unsupported etc.

Each time I invoke a `fedimint-cli` command, I would be annoyed with a summary of supported and unsupported modules, but I would like to know a warning that the federation I'm connected to has a module my client doesn't support.

I'm not sure if we anticipate it's common for clients to connect to federations that have unsupported modules. If so, we should probably chew on a better way to surface that to users.

dpc:
Alternatively we could just change RUST_LOG in mprocs. But Federations that have unsupported modules might not be something special and unusual in real use. If what user is doing is unrelated is there really a point earning them about anything?

bradleystachurski:
I like the approach of updating RUST_LOG. WARN seems like the correct log level but there's no need for the user to see that as part of the output when interacting with mprocs.

douglaz:
It should be a debug log and we could show this as warning on `fedimint-cli info`

dpc:
So I guess we're not doing it and went with a different approach (change offending line to `debug!`).

elsirion:
> So I guess we're not doing it and went with a different approach (change offending line to `debug!`).

I think that's the right way to go, but it's still a TODO, right? At least I found a bunch of places that use the exact wording "Detected configuration for unsupported module".

elsirion (approved):

dpc (approved):

PR #4457, opened 2024-03-04
feat: helper for canceling a future on task group shutdown

maan2003:


maan2003:
how do you feel about clippy error, to me feels like we should disable that lint

dpc:
> how do you feel about clippy error, to me feels like we should disable that lint

It seems to me it has a point, because when matching that clippy error with the diff, I wasn't sure where is the `if` even ending. :D

dpc (dismissed):
LGTM, but clippy doesn't lie. :D

elsirion (approved):
Nice cleanup!

justinmoon (approved):

PR #4462, opened 2024-03-05
fix: Client::has_pending_recoveries

maan2003:
earlier it returned true if client_recovery_progress_receiver returned empty btreemap

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1470,7 +1470,7 @@ impl Client {
             .client_recovery_progress_receiver
             .borrow()
             .iter()
-            .any(|(_id, progress)| !progress.is_done())
+            .all(|(_id, progress)| progress.is_done())
     }
 
     /// Wait for all module recoveries to finish


bradleystachurski (approved):
Nice catch! Out of curiosity, how'd you find this? I don't see any usages of `has_pending_recoveries` in this repo.

maan2003:
> Nice catch! Out of curiosity, how'd you find this? I don't see any usages of has_pending_recoveries in this repo.

we are using it in fedi

dpc (approved):

fedimint-client/src/lib.rs
```diff
@@ -1470,7 +1470,7 @@ impl Client {
             .client_recovery_progress_receiver
             .borrow()
             .iter()
-            .any(|(_id, progress)| !progress.is_done())
+            .all(|(_id, progress)| progress.is_done())
```
elsirion, line 1473: The problem with this fn was that it returned the opposite of what its name suggests, you could have just flipped the negation on line 1469 :sweat_smile: 

```
!any(!done) == all(done) // What we had
any(!done) == !all(done) // What we want
```

Someone likely just confused themselves with all these negations when writing this initially.

dpc: The any vs all behavior on empty is tricky. Maybe this whole methods should have been `are_all_recoveries_done` indeed. I kind of wrote it just to show it's now possible to have APIs like that (including subscribing to updates), and the code is not used anywhere which means that 99% it doesn't work. :D

@maan2003 As you are working on integrating this in Fedi app, don't be afraid to completely change anything there. IMO, the most useful part of TDD is that only when you actually write the user code, you have a good feel for what good API should look like. And when you don't, you get buggy code and bad APIs. (like I did here) :D

PR #4470, opened 2024-03-05
chore: move dev shell target to `./target-nix`

dpc:
People quite often report problems with constant rebuilds when using VSCode and other IDEs started outside of the dev shell.

These dev shells are using different rust-analyzer and Rust toolchain versions that mututally invalidate each other with dev shell setup.

While there are ways to configure such IDEs to co-exist with dev shell, they require customizing every text editor for every user.

We can't easily control IDEs of everyone, but we have a perfect control of the dev shell. It seems quite easy to just get out of the way, not use the default `./target` and problem mostly solved:

Dev shell users won't really notice the change, mostly.

Shell scripts need to use `CARGO_BUILD_TARGET_DIR`.

IDE users will have a separate ./target and ./target-nix, but it seems much better then suffering constant rebuilds of everything.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -249,6 +249,8 @@
                 ];
 
                 shellHook = ''
+                  root="$(git rev-parse --show-toplevel)"
+
                   # workaround https://github.com/rust-lang/cargo/issues/11020
                   cargo_cmd_bins=( $(ls $HOME/.cargo/bin/cargo-{clippy,udeps,llvm-cov} 2>/dev/null) )
                   if (( ''${#cargo_cmd_bins[@]} != 0 )); then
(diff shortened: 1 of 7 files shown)


flake.nix
```diff
@@ -279,6 +281,7 @@
                   fi
 
                   export RUSTC_WRAPPER=${pkgs.sccache}/bin/sccache
+                  export CARGO_BUILD_TARGET_DIR="$root/target-nix"
```
maan2003, line 284: should respect user's CARGO_BUILD_TARGET_DIR if it is already set

justinmoon (approved):

elsirion:
Excited to try this out, I had a bunch of weirdness happen to me in RustRover lately :sweat_smile:

PR #4472, opened 2024-03-06
chore(ci): in PR and MQ, build ci version of containers

dpc:


bradleystachurski (approved):

.github/workflows/ci-nix.yml
```diff
@@ -367,27 +367,47 @@ jobs:
 
       - name: Build fedimintd container
         run: |
-          nix build -L .#container.fedimintd
+          if [ "$GITHUB_EVENT_NAME" == "pull_request" ] || [ "$GITHUB_EVENT_NAME" == "merge_group" ]; then
+            nix build -L .#ci.container.fedimintd
```
elsirion, line 371: I'm having a very hard time even finding where the `ci` namespace is defined, I assume it's related to cargo profiles so this will lead to faster compile times/more re-use of existing artifacts? A little explanation for the non-Nix gods would be great :laughing:

elsirion (approved):

.github/workflows/ci-nix.yml
dpc: `craneMultiBuild` internally in Flakebox is preparing a matrix of things including toolchains & cargo build profiles. They are all exposed as `legacyPackages` output, thought due to being nested it's hard to discover them without knowing they are there. @elsirion

dpc: A bit documented https://github.com/rustshop/flakebox/blob/master/docs/flakeboxLib.md#flakeboxlibcranemultibuild-

but best documentation is always the code itself: https://github.com/rustshop/flakebox/blob/54a6daa911a1f38eda1e30b924bf36d2fae41dcf/lib/craneMultiBuild.nix#L16

PR #4449, opened 2024-03-02
feat: parse-json for oobNotes and cli-command

Kodylow:
Adds a helper function to OOBNotes and a cli command to parse OOBNotes to JSON, been running into a bunch of situations where I'd like to parse a notes string and see what's in it.

Happy for any feedback on how to do it better, mostly want it for the fedimint-ts client so I don't have to try binary parsing notes in typescript.

```
$ fedimint-cli parse-json AgEEGhuHwAD1AgIBpf2ztoZ3YI9DX1iN06H+WqaQlgw1cG2s5plFzyTonpKQ6B9lANXwvD0H2TuH8vdANLzSLklv4WEPnNtknsStaY4DC4YUZp7tHsmKZ5uCHUoIArQxr5S/SzxUq5vsKXd1GKAAvMIAUXzbqMGUXgO61vUq0aOutjXDamJU+YJm0CxRBz7TW9CdTZwE/gax7DnA2XU5XuEzdBd7sJqr9LGY75lSjodKEWoYVBRy348b+YNufcLXXlAzVVOSw/CgcsFC2xAg7pMI+4IPbEuGc9i3nlFWgCParnCKCNXQGWMTTT+12h3ZiEcvmXkD0iwFSOTYvYU=

{
  "federation_id_prefix": [
    26,
    27,
    135,
    192
  ],
  "notes": {
    "2": [
      {
        "signature": "a5fdb3b68677608f435f588dd3a1fe5aa690960c35706dace69945cf24e89e9290e81f6500d5f0bc3d07d93b87f2f740",
        "spend_key": "34bcd22e496fe1610f9cdb649ec4ad698e030b8614669eed1ec98a679b821d4a"
      }
    ],
    "8": [
      {
        "signature": "b431af94bf4b3c54ab9bec29777518a000bcc200517cdba8c1945e03bad6f52ad1a3aeb635c36a6254f98266d02c5107",
        "spend_key": "3ed35bd09d4d9c04fe06b1ec39c0d975395ee13374177bb09aabf4b198ef9952"
      },
      {
        "signature": "8e874a116a18541472df8f1bf9836e7dc2d75e5033555392c3f0a072c142db1020ee9308fb820f6c4b8673d8b79e5156",
        "spend_key": "8023daae708a08d5d01963134d3fb5da1dd988472f997903d22c0548e4d8bd85"
      }
    ]
  }
}
Kodys-MacBook-Pro-2:fedimint kody$ 
```

PR #4473, opened 2024-03-06
feat: dev commands to decode and encode notes and invite_code to and from json

Kodylow:
replaces https://github.com/fedimint/fedimint/pull/4449

Adds dev codes for fedimint-cli to encode/decode to/from json for notes and invite_code

Would it be out of scope to add cashu encoding here? It'd just rename the keys and put the amount in each of the notes then spit out the json

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -352,7 +352,6 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
 }
 
 pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
-    let fedimint_cli_version = crate::util::FedimintCli::version_or_default().await;
     log_binary_versions().await?;
     let data_dir = env::var("FM_DATA_DIR")?;
 
@@ -419,10 +418,29 @@ pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
 
     let fed_id = fed.calculate_federation_id().await;
     let invite = fed.invite_code()?;
-    let invite_code = cmd!(client, "dev", "decode-invite-code", invite.clone())
-        .out_json()
-        .await?;
-    anyhow::ensure!(
+
+    let fedimint_cli_version = crate::util::FedimintCli::version_or_default().await;
+    let version_req = VersionReq::parse(">=0.3.0-alpha")?;
+
+    let invite_code = if version_req.matches(&fedimint_cli_version) {
+        cmd!(client, "dev", "decode", "invite-code", invite.clone())
+    } else {
+        cmd!(client, "dev", "decode-invite-code", invite.clone())
+    }
+    .out_json()
+    .await?;
+
+    let encode_invite_output = if version_req.matches(&fedimint_cli_version) {
+        cmd!(
+            client,
+            "dev",
+            "encode",
+            "invite-code",
+            format!("--url={}", invite_code["url"].as_str().unwrap()),
+            "--federation_id={fed_id}",
+            "--peer=0"
+        )
+    } else {
         cmd!(
             client,
             "dev",
@@ -431,10 +449,14 @@ pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
             "--federation_id={fed_id}",
             "--peer=0"
         )
-        .out_json()
-        .await?["invite_code"]
+    }
+    .out_json()
+    .await?;
+
+    anyhow::ensure!(
+        encode_invite_output["invite_code"]
             .as_str()
-            .unwrap()
+            .expect("invite_code must be a string")
             == invite,
         "failed to decode and encode the client invite code",
     );
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -755,7 +755,7 @@ impl FedimintCli {
                 client
                     .wait_for_all_active_state_machines()
                     .await
-                    .map_err_cli()?;
+                    .map_err_cli_msg("failed to wait for all active state machines")?;
                 Ok(CliOutput::Raw(serde_json::Value::Null))
             }
             Command::Dev(DevCmd::Decode { decode_type }) => match decode_type {
@@ -764,7 +764,9 @@ impl FedimintCli {
                     federation_id: invite_code.federation_id(),
                 }),
                 DecodeType::Notes { notes } => {
-                    let notes_json = notes.notes_json().map_err_cli_general()?;
+                    let notes_json = notes
+                        .notes_json()
+                        .map_err_cli_msg("failed to decode notes")?;
                     Ok(CliOutput::Raw(notes_json))
                 }
             },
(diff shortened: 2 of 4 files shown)


fedimint-cli/src/lib.rs
```diff
@@ -761,19 +794,37 @@ impl FedimintCli {
...
-                url,
-                federation_id,
-                peer,
-            }) => Ok(CliOutput::InviteCode {
-                invite_code: InviteCode::new(url, peer, federation_id),
-            }),
+                }),
+                DecodeType::Notes { notes } => Ok(CliOutput::DecodeNotes {
+                    federation_id_prefix: notes.federation_id_prefix().to_string(),
+                    notes: notes.notes().clone(),
```
elsirion, line 804: I actually liked the previous version of your JSON encoding code which was able to display all `OOBNoteData` variants. Now we are missing invite codes when provided.

fedimint-cli/src/lib.rs
```diff
@@ -359,6 +364,35 @@ enum AdminCmd {
...
+        #[clap(long = "peer")]
+        peer: PeerId,
+    },
+
+    /// Encode a JSON string of notes to an ecash string
+    Notes {
+        #[clap(long = "federation_id_prefix")]
+        federation_id_prefix: String,
+        #[clap(long = "notes")]
+        notes: String,
```
elsirion, line 392: ~Shouldn't this be a `Vec<_>`? Ideally a `Vec<SpendableNote>`, but due to https://github.com/clap-rs/clap/issues/1704 likely rather a `Vec<String>`.~ Maybe supplying the whole vec as JSON is actually more convenient.

fedimint-cli/src/lib.rs
```diff
@@ -359,6 +364,35 @@ enum AdminCmd {
...
+    /// Decode a string of ecash notes into a JSON representation
+    Notes { notes: OOBNotes },
+}
+
+#[derive(Debug, Clone, Subcommand)]
+enum EncodeType {
+    /// Encode connection info from its constituent parts
+    InviteCode {
+        #[clap(long = "url")]
+        url: SafeUrl,
```
elsirion, line 380: I think if the field name is the same as the long argument name you can just annotate it as `#[clap(long)]`

fedimint-cli/src/lib.rs
Kodylow: I originally had it as a vec but it's super tedious to input that way, doing it as Json string you can just wrap the JSON output from the dev decode command with single quotes like `'{"notes": stuff}'`

fedimint-cli/src/lib.rs
Kodylow: updated to use it on OOB notes through .notes_json

fedimint-cli/src/lib.rs
Kodylow: updated

elsirion:
Breaks backwards compat tests, I had to deal with that in #4415, I bet you also have to do some version switching for the command to use for decoding since you changed the CLI.

devimint/src/tests.rs
```diff
@@ -346,6 +346,8 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
 }
 
 pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
+    let fedimint_cli_version = crate::util::FedimintCli::version_or_default().await;
+    let version_req = VersionReq::parse(">=0.3.0-alpha")?;
```
dpc, line 350: If I see correctly, this variable is used only once, and having it here just makes it far from where it's used.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -154,6 +155,44 @@ impl OOBNotes {
             .expect("Invariant violated: OOBNotes does not contain any notes")
     }
 
+    pub fn notes_json(&self) -> Result<serde_json::Value, serde_json::Error> {
+        let mut map = serde_json::Map::new();
```
dpc, line 159: Nit: `map` is a meh name for a variable - it doesn't describe the meaning of it in it.

dpc (approved):

devimint/src/tests.rs
Kodylow: I can move it down, was thinking that eventually all the tests will require version matching so define up top but nbd

dpc: Fair enough. :+1: 

But... in that case it might need a longer name.

dpc (dismissed):

dpc:
I'd just squash it all at this point, no biggy.

dpc (commented):
LGTM, but really could use a squash.

dpc:
```
fix: backwards compatibility tests

fix: fix

fix: test

fix: hold

fix: tests fix

fix: encoding ecash

fix: decode on OOBNotesData

feat: encode dev command for notes and invite-code

feat: decode notes to json
```

not a good commit message :P

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -154,6 +155,44 @@ impl OOBNotes {
             .expect("Invariant violated: OOBNotes does not contain any notes")
     }
 
+    pub fn notes_json(&self) -> Result<serde_json::Value, serde_json::Error> {
+        let mut notes_map = serde_json::Map::new();
+        for notes in self.0.iter() {
+            match notes {
+                OOBNotesData::Notes(notes) => {
+                    let notes_json = serde_json::to_value(notes).expect("serialization can't fail");
```
m1sterc001guy, line 163: Nit: the other `to_value` return `Result` rather than `expect`, you could do the same here

devimint/src/tests.rs
```diff
@@ -425,11 +443,12 @@ pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
...
-        .await?["invite_code"]
-            .as_str()
-            .unwrap()
-            == invite,
+    }
+    .out_json()
+    .await?;
+
+    anyhow::ensure!(
+        encode_invite_output["invite_code"].as_str().unwrap() == invite,
```
m1sterc001guy, line 451: Nit: use `expect` instead of `unwrap`

m1sterc001guy (dismissed):

m1sterc001guy (dismissed):

dpc (dismissed):

dpc (approved):
But linters not happy

m1sterc001guy (approved):

dpc (approved):

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -154,6 +155,44 @@ impl OOBNotes {
             .expect("Invariant violated: OOBNotes does not contain any notes")
     }
 
+    pub fn notes_json(&self) -> Result<serde_json::Value, serde_json::Error> {
+        let mut notes_map = serde_json::Map::new();
+        for notes in self.0.iter() {
+            match notes {
+                OOBNotesData::Notes(notes) => {
+                    let notes_json = serde_json::to_value(notes)?;
+                    notes_map.insert("notes".to_string(), notes_json);
```
elsirion, line 164: We should probably assert that insert returns `None` to avoid confusing ourselves should we ever allow multiple instances of some data type.

PR #4474, opened 2024-03-06
include wasm-test.sh in test-ci-all (almost)

dpc:
This required:

* reachitecting flakebox target vs toolchain handling
* fixing wasm-test.sh to stop invalidating ./target
* bunch of nix glue to stich native vs wasm32-unknown ./targets together
* continuously ignoring the question is it even worth the time

Unfortunately after tackling all the scripts & Nix issues, it turned out that wasm-tests (headless firefox?) just doesn't work when being run in GNU `parallel` or `xargs -P`, and debugging it seems like lots of work.

So we'll still run wasm-test separately, but at least it will now re-use build artifacts with the main build, so it should still be faster. And in the future we can try again.

### DIFF
--- a/nix/flakebox.nix
+++ b/nix/flakebox.nix
@@ -219,6 +219,17 @@ rec {
     buildPhaseCargoCommand = "cargoWithProfile doc --locked ; cargoWithProfile check --all-targets --locked ; cargoWithProfile build --locked --all-targets";
   };
 
+  workspaceDepsWasmTest = craneLib.buildWorkspaceDepsOnly {
+    pname = "${commonArgs.pname}-wasm-test";
+    buildPhaseCargoCommand = "cargoWithProfile build --locked --tests -p fedimint-wasm-tests";
+  };
+
+  workspaceBuildWasmTest = craneLib.buildWorkspace {
+    pnameSuffix = "-workspace-wasm-test";
+    cargoArtifacts = workspaceDepsWasmTest;
+    buildPhaseCargoCommand = "cargoWithProfile build --locked --tests -p fedimint-wasm-tests";
+  };
+
   workspaceTest = craneLib.cargoNextest {
     cargoArtifacts = workspaceBuild;
     cargoExtraArgs = "--workspace --all-targets --locked";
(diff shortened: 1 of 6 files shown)


dpc:
@maan2003 

Any idea why this test might be failing now?

It seems like it hangs and timeout kills it. The same thing happens locally on my machine, so I doubt it's running out of resources.

```
nix build -L .#wasm32-unknown.ci.wasmTest
```

uses the same method and works perfectly.

Seems like two rust tests in that hanged wasm-test are actualy passing so everything looks like it's working.

maan2003:
very weird I have no idea why it would hang

nix/flakebox.nix
```diff
@@ -389,24 +400,43 @@ rec {
...
+        # default to building for native; running test for cross-compilation targets
+        # here doesn't make any sense, and `wasm32-unknown-unknown` toolchain is used
+        # mostly to opt-in into wasm tests
+        unset CARGO_BUILD_TARGET
+
+        # TODO: hack, remove
+        echo "target"
+        ls ./target/pkgs
+        echo "target/wa..."
+        ls ./target/pkgs/fedimint-wasm-tests
```
elsirion: I guess you wanted to remove that before merging?

fedimint-client/src/envs.rs
```diff
@@ -0,0 +1,26 @@
+use std::str::FromStr;
+use std::time::Duration;
+
+use tracing::warn;
+
+pub const FM_DISCOVER_API_VERSION_TIMEOUT_ENV: &str = "FM_DISCOVER_API_VERSION_TIMEOUT";
+
+#[cfg(not(target_family = "wasm"))]
+pub fn get_discover_api_version_timeout() -> Duration {
```
elsirion, line 9: This commit was in a different PR already, I guess it will be dropped by GH on merge? If you do another round would be good to drop manually.

nix/flakebox.nix
```diff
@@ -418,12 +418,6 @@ rec {
         # here doesn't make any sense, and `wasm32-unknown-unknown` toolchain is used
         # mostly to opt-in into wasm tests
         unset CARGO_BUILD_TARGET
-
-        # TODO: hack, remove
-        echo "target"
-        ls ./target/pkgs
-        echo "target/wa..."
-        ls ./target/pkgs/fedimint-wasm-tests
```
elsirion: I see you did this later

elsirion (commented):
The last commit undoes a lot of the second to last one. I think you can squash them to just keep the changes that actually worked.

maan2003 (approved):

Kodylow (approved):
ACK, issue created for TODO in test-ci-all

Issue #4398, opened 2024-02-22
Create `network_timeout` variable in the API client


bradleystachurski:
How about defining a `network_timeout` variable in the API client that can be set via the client builder? That way the tests could use a lower timeout.

_Originally posted by @elsirion in https://github.com/fedimint/fedimint/pull/4247#discussion_r1495638960_

dpc:
I believe  initial version discovery is the only place where this matters, so we don't really need to go all generic "network timeout", but possibly we could just make the initial (blocking discovery be much shorter), or if that doesn't work well condtionally shorten it in test via some env var.

dpc:
TODO: move env var from fedimint-client to fedimint-cli https://github.com/fedimint/fedimint/pull/4476#issuecomment-1981476810

PR #4476, opened 2024-03-06
chore: timeouts on version discovery

dpc:
When working locally I've noticed test failing due to too short api discovery timeout.

Parametrize timeout of api version discovery using env variable. Set it to low value only when testing offline nodes.

### DIFF
--- a/fedimint-client/src/envs.rs
+++ b/fedimint-client/src/envs.rs
@@ -1,0 +1,26 @@
+use std::str::FromStr;
+use std::time::Duration;
+
+use tracing::warn;
+
+pub const FM_DISCOVER_API_VERSION_TIMEOUT_ENV: &str = "FM_DISCOVER_API_VERSION_TIMEOUT";
+
+#[cfg(not(target_family = "wasm"))]
+pub fn get_discover_api_version_timeout() -> Duration {
+    if let Ok(s) = std::env::var(FM_DISCOVER_API_VERSION_TIMEOUT_ENV) {
+        match FromStr::from_str(&s) {
+            Ok(secs) => return Duration::from_secs(secs),
+            Err(err) => warn!(
+                %err,
+                var = FM_DISCOVER_API_VERSION_TIMEOUT_ENV,
+                "Could not parse env variable"
+            ),
+        }
+    }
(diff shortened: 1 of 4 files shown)


elsirion (approved):

maan2003 (commented):
fedimint-client should not use environment variables.
libraries using env variables is super weird. 
we should move env variables checking to fedimint-cli

elsirion:
The problem with that is getting the setting deep into the client. Eventually we'll need a general solution for things like this to also override the default esplora server etc., but for now this seems like the only reasonable-effort way to adjust the timeout in tests without breaking clients on slow connections.

dpc:
> fedimint-client should not use environment variables.
> libraries using env variables is super weird.
> we should move env variables checking to fedimint-cli

I agree, but it isn't a big crime and that can be a follow-up. We'll have some builder pattern on `Client` where we can set this value, and read it from variable. But this PR is better than hardcoding a super-short timeout, which will break production cases, and all in all it isn't too bad. @maan2003

bradleystachurski (approved):
@dpc Thanks for tackling this. I think we can close https://github.com/fedimint/fedimint/issues/4398, but feel free to keep open if you want to refactor the misdemeanor in a followup.

@maan2003 I share your discomfort with using env vars to trigger different logic in prod vs tests. I'm approving since I prefer the tradeoff to fix the broken behavior in prod first, which can be refactored in a fast follow.

dpc:
```
copying path '/nix/store/mx1mwi6bfzjsblhw1m95qs1zwb0l382c-ndk-bundle-23.0.7344513-rc4' from 'https://fedimint.cachix.org/'...
Error: The operation was canceled.
```

@bradleystachurski Seems like the stale timeout also didn't help? :(

bradleystachurski:
@dpc Should we try temporarily bumping the cross compile timeout just?

dpc:
> @dpc Should we try temporarily bumping the cross compile timeout?

Worth giving a shot, I guess. Can you make a PR?

bradleystachurski:
https://github.com/fedimint/fedimint/pull/4487

PR #4477, opened 2024-03-06
refactor: create_final_inputs_and_outputs

joschisan:
This refactor gives modules more freedom when finalising a transaction as inputs and outputs don't have to be created separately anymore. This change is in preparation for ecash-ng.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -67,7 +67,6 @@
 //!
 //! For a hacky instantiation of a complete client see the [`ng` subcommand of `fedimint-cli`](https://github.com/fedimint/fedimint/blob/55f9d88e17d914b92a7018de677d16e57ed42bf6/fedimint-cli/src/ng.rs#L56-L73).
 
-use std::cmp::Ordering;
 use std::collections::{BTreeMap, HashSet};
 use std::fmt::{Debug, Formatter};
 use std::ops::{self, Range};
@@ -141,9 +140,8 @@ use crate::sm::{
     ClientSMDatabaseTransaction, DynState, Executor, IState, Notifier, OperationState, State,
 };
 use crate::transaction::{
-    tx_submission_sm_decoder, ClientInput, ClientOutput, TransactionBuilder,
-    TransactionBuilderBalance, TxSubmissionContext, TxSubmissionStates,
-    TRANSACTION_SUBMISSION_MODULE_INSTANCE,
+    tx_submission_sm_decoder, ClientInput, ClientOutput, TransactionBuilder, TxSubmissionContext,
+    TxSubmissionStates, TRANSACTION_SUBMISSION_MODULE_INSTANCE,
 };
 
 /// Client backup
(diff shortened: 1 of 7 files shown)


dpc:
@elsirion probably best for you to look at, I don't remember if anyone is familiar with note selection.

elsirion:
This is related to a long-term project (e-cash NG) and just had a very low prio. The solution sounded good, but wasn't worth the bandwidth rn. Let's revisit later.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -722,35 +732,57 @@ impl ClientModule for MintClientModule {
...
+            .await;
+        }
+
+        let inputs = self
+            .create_input_from_notes(operation_id, selected_notes)
+            .await?;
+
+        assert!(!inputs.is_empty());
+
+        Ok((inputs, selected_amount))
```
elsirion, line 769: Instead of returning the selected amount, why not just sum the input amount where needed? Returning this tuple was very confusing to me at the call-site and is also redundant and thus a potential source of bugs.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -650,23 +650,33 @@ impl ClientModule for MintClientModule {
...
-    ) -> anyhow::Result<Vec<ClientInput<MintInput, MintClientStateMachines>>> {
-        self.create_input(dbtx, operation_id, min_amount).await
-    }
+        input: Amount,
+        output: Amount,
+    ) -> anyhow::Result<(
+        Vec<ClientInput<MintInput, MintClientStateMachines>>,
+        Vec<ClientOutput<MintOutput, MintClientStateMachines>>,
+    )> {
+        let (inputs, selected_input) = self
```
elsirion, line 663: `selected_input` is a very bad name for the `selected_input_amount`, as mentioned before this should just be calculated here for readability.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -650,23 +650,33 @@ impl ClientModule for MintClientModule {
...
-        operation_id: OperationId,
-        amount: Amount,
-    ) -> Vec<ClientOutput<MintOutput, MintClientStateMachines>> {
-        // FIXME: don't hardcode notes per denomination
-        self.create_output(dbtx, operation_id, 2, amount).await
+        let selected_input_fee = self
+            .cfg
+            .fee_consensus
+            .note_spend_abs
+            .mul_u64(inputs.len() as u64);
```
elsirion, line 671: The fee is already contained in the `ClientInput`, isn't it?

modules/fedimint-mint-client/src/lib.rs
joschisan: Done.

modules/fedimint-mint-client/src/lib.rs
joschisan: Done.

fedimint-client/src/lib.rs
```diff
@@ -912,10 +910,7 @@ impl Client {
     /// # Panics
     /// If any of the input or output versions in the transaction builder are
     /// unknown by the respective module.
-    fn transaction_builder_balance(
-        &self,
-        builder: &TransactionBuilder,
-    ) -> TransactionBuilderBalance {
+    fn transaction_builder_balance(&self, builder: &TransactionBuilder) -> (Amount, Amount) {
```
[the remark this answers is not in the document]
joschisan: Seems cumbersome for a private method only called twice in one other function, updated the docs though.

modules/fedimint-mint-client/src/lib.rs
elsirion: Ah, right, we discussed moving it there but didn't do so …

elsirion (approved):

PR #4478, opened 2024-03-06
fix: race in client shutdown

maan2003:
previous implementation feels racy if Weak was upgraded after checking the client count

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -520,9 +520,6 @@ pub struct ClientArc {
 impl ClientArc {
     /// Create
     fn new(inner: Arc<Client>) -> Self {
-        inner
-            .client_count
-            .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         Self {
             inner: Some(inner),
             // this is the constructor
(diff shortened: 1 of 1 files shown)


fedimint-client/src/lib.rs
```diff
@@ -699,13 +687,6 @@ pub struct Client {
     secp_ctx: Secp256k1<secp256k1_zkp::All>,
 
     task_group: TaskGroup,
-    /// Number of [`ClientArc`] instances using this `Client`.
-    ///
-    /// The `Client` struct is both used for the client itself as well as
-    /// for the global context used in the state machine executor. This means we
-    /// cannot rely on the reference count of the `Arc<Client>` to
-    /// determine if the client should shut down.
```
maan2003, line 707: I don't understand this comment at all

elsirion: That's basically saying we have cyclic `Arc`s, so we introduce a counter that only counts the ones that we see as "root" and once all of these are dropped makes sure the remaining `Arc`s are also dropped to break the cycle. Is that roughly correct @dpc?

dpc: I wasn't the one that come up with it. (I thought @maan2003 was :D), but here is how I understand it:

The `inner: Option<Arc<Client>>`, can be cloned somewhere internally (the `Arc<Client>`) in modules and executors. So we can not rely on it to know when the last reference to the `ClientArc` was dropped.

Therefore we track the number of `ClientArc`s in `Client::client_count` instead and decrement/increment only on `ClientArc` clones/drops.

When ``Client::client_count` goes to zero we know that all `ClientArc`s are gone, and in the last `drop` we send a shutdown signal to state machine executor and recovery executor, then we block waiting for them to finish. This should drop all the module code, and free up the inner `Arc<Client>`. Then we can unblock the `ClientArc` drop to initialize it.

As is this PR (#4478) seems incorrect, and it will just often not stop SM executor. So it might look like things are working better, but that's because we have no attempted to shut down anything. This will cause database reuse errors.

Now that I look at it, I think it would better to just use two layers of `Arc`.

Rename `ClientArc` to `ClientArc`

```
struct ClientHandle(Arc<SharedClientHandle>);
struct SharedClientHandle(Option<Arc<Client>>);
```


and then get rid of `Client::client_count`. Then stopping executor can be done in `SharedClientHandle`, by sending stop signal, `take()`ing the `Option` to drop it, and blocking waiting for database to be released. But this is just nicer and more automatic version of the current code, AFAICT, and by itself wouldn't explain/fix any bugs.

dpc: @maan2003 @elsirion Happy to jump on a call to align the vision and come up with an exact strategy to handle this.

maan2003:
closed in favor of #4484

PR #4482, opened 2024-03-06
chore: client shutdown on handle drop incorrect

dpc:
Fundamentally we can't rely on database count to detect when client executors (state machines and recoveries) finished, because external user might hold some references too.

Also, we must not be dropping `inner : Arc<Client>` before we ensure executors stopped, because something running in them might still require that `Arc<Client>` is alive and weak references to them can be upgraded.

Instead use the old and (somewhat) trusted TaskGroup to do the job.

### DIFF
--- a/fedimint-client/src/module/init/recovery.rs
+++ b/fedimint-client/src/module/init/recovery.rs
@@ -405,7 +405,7 @@ where
             self.api().clone(),
             *self.core_api_version(),
             client_ctx.decoders(),
-            common_state.next_session..common_state.end_session,
+            block_stream_session_range,
         );
         let client_ctx = self.context();
 


fedimint-client/src/lib.rs
```diff
@@ -623,29 +623,35 @@ impl Drop for ClientArc {
...
+                        if let Err(err) = inner
+                            .task_group
+                            .clone()
+                            .shutdown_join_all(Some(Duration::from_secs(30)))
+                            .await
+                        {
+                            warn!(target: LOG_CLIENT, %err, "Error waiting for client task group to shut down");
+                        }
+                    });
+                });
```
elsirion, line 640: This might lead to a warning in the drop impl of `ExecutorInner` now, that calls `stop_executor`:

https://github.com/fedimint/fedimint/blob/3599ce3b8ab32667243a291271967fc090f11a87/fedimint-client/src/sm/executor.rs#L724-L726

fedimint-client/src/lib.rs
```diff
@@ -623,29 +623,35 @@ impl Drop for ClientArc {
...
-                                    %federation_id,
-                                    strong_count,
-                                    "Waiting for client database to stop being used"
-                                );
-                            }
-                            fedimint_core::task::sleep(Duration::from_millis(10)).await;
-                        });
-                    });
+                if client_strong_count != 1 {
+                    debug!(target: LOG_CLIENT, count = client_strong_count - 1, LOG_CLIENT, "External Client references remaining after last handle dropped");
```
elsirion, line 649: Should this be a warning?

elsirion (approved):

fedimint-client/src/lib.rs
dpc: AFAIU, at least currently it seems possible to get ahold of `Arc<Client>` that would not be a `ClientHandle` from the user code and maybe there's a legitimate reason to do it. E.g. not everyone will need to be restarting the `Client`.

It's even more the case with the `Database` references, which will generally always be possible to hold longer than the `Client` itself.

So for now I just made these debug diagnostics.

fedimint-client/src/lib.rs
dpc: We can downgrade to `debug!` or rethink the approach if it becomes a problem, I guess.

fedimint-client/src/lib.rs
```diff
@@ -623,29 +623,35 @@ impl Drop for ClientArc {
                 }
 
                 let db = inner.db.clone();
-                let federation_id = inner.federation_id();
 
+                tokio::task::block_in_place(|| {
+                    tokio::runtime::Handle::current().block_on(async {
```
maan2003, line 628: need explicit method of shutdown

fedimint-client/src/module/init/recovery.rs
```diff
@@ -405,7 +405,7 @@ where
             self.api().clone(),
             *self.core_api_version(),
             client_ctx.decoders(),
-            block_stream_session_range,
+            common_state.next_session..common_state.end_session,
```
maan2003, line 408: unrelated change snuck in

fedimint-client/src/lib.rs
maan2003: impl in https://github.com/fedimint/fedimint/pull/4492

fedimint-client/src/module/init/recovery.rs
dpc: Oh, I accidentally inlined `block_stream_session_range` here. No biggy.

maan2003 (approved):

dpc:
Just a tiny fix to the issue Manmeet pointed out (accidentally inlined a var)

bradleystachurski (approved):

PR #4483, opened 2024-03-06
fix: wrong git version hash calculated if source dirty

dpc:
Mathz are hard.

Hey, at least it's not off by 1.

### DIFF
--- a/fedimint-build/src/lib.rs
+++ b/fedimint-build/src/lib.rs
@@ -66,7 +66,7 @@ fn set_code_version_inner() -> Result<(), String> {
         // state by replacing the middle with 0s. This should
         // be noticeable enough, while letting find out the
         // root commit anyway.
-        format!("{}00000000{}", &hash[0..16], &hash[26..40])
+        format!("{}00000000{}", &hash[0..16], &hash[(40 - 16)..40])
     } else {
         hash
     };


fedimint-build/src/lib.rs
```diff
@@ -66,7 +66,7 @@ fn set_code_version_inner() -> Result<(), String> {
         // state by replacing the middle with 0s. This should
```
bradleystachurski, line 66: I'm a bit confused by this comment. Are we supposed to keep the hash 40 hex chars long? If so, wouldn't we want to skip the 8 chars of "0" in the middle? i.e. `[(40 - 16 + 8)..40]`?

bradleystachurski: Ahh scratch that.

bradleystachurski (approved):

fedimint-build/src/lib.rs
```diff
@@ -66,7 +66,7 @@ fn set_code_version_inner() -> Result<(), String> {
...
         // be noticeable enough, while letting find out the
         // root commit anyway.
-        format!("{}00000000{}", &hash[0..16], &hash[26..40])
+        format!("{}00000000{}", &hash[0..16], &hash[(40 - 16)..40])
```
maan2003, line 69: maybe `16+8` instead of 40

bradleystachurski: Yeah, I agree this would be easier to reason about at a glance if it were

```
[0..16], [(16+8)..40]
```

or even

```
let zeros_count = 8;
format!(
    "{}{}{}",
    &hash[0..16],
    "0".repeat(zeros_count),
    &hash[(16 + zeros_count)..40]
)
```

dpc: The damage is done, no need to overthink it. When it was as a constant (incorrect at at) no that it isn't, there's 10 ways to express it, and bikesheding starts. :D

elsirion (approved):
button-press-the-button

PR #4484, opened 2024-03-06
 refactor(client): remove manual client arc counting 

dpc:
On top of #4482 

Use two levels of `Arc` instead. While at it, rename and document things
better.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -113,7 +113,7 @@ use fedimint_core::{
 };
 pub use fedimint_derive_secret as derivable_secret;
 use fedimint_derive_secret::DerivableSecret;
-use fedimint_logging::LOG_CLIENT;
+use fedimint_logging::{LOG_CLIENT, LOG_CLIENT_RECOVERY};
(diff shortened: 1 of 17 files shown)


fedimint-client/src/lib.rs
```diff
@@ -594,8 +608,8 @@ pub(crate) struct ClientWeak {
 }
 
 impl ClientWeak {
-    pub fn upgrade(&self) -> Option<ClientHandle> {
-        Weak::upgrade(&self.inner).map(ClientHandle::new)
+    pub fn upgrade(&self) -> Option<ClientStrong> {
```
elsirion, line 611: Ah, I think you fixed that problem here

elsirion (approved):
I think in total the PR is good, just one intermediately broken commit.

dpc:
> I think in total the PR is good, just one intermediately broken commit.

Fixed at first, but then decided to squash the 3 last commits, as it doesn't help anything to have a commit with a problem.

elsirion (approved):

Kodylow (approved):
tACK

PR #4490, opened 2024-03-07
Fix CI on master

elsirion:
- [x] Extra docker containers
- [x] Docs


elsirion:
I can't figure out how to make crane use a nightly toolchain without adding an entirely new version of `craneLib` that overwrites the toolchain, maybe @dpc can shed some light here: 5f865aaa1c6ff135b2e18cccb565de34829dd5e6

elsirion:
Let's at least attempt to fix docker

elsirion:
Thx @dpc for showing me how to build things with nightly in flakebox!

dpc (approved):

dpc:
BTW. If it was sending notifications to Discord, like `ci-nix.yml` does we would have noticed faster.

bradleystachurski (approved):
I think we'll also need to update the path for prosody, but we can tackle in a followup.

Closes https://github.com/fedimint/fedimint/issues/4243

elsirion:
> BTW. If it was sending notifications to Discord, like `ci-nix.yml` does we would have noticed faster.

I don't think it's about noticing, it's about getting annoyed enough to fix it :rofl:

dpc:
> I don't think it's about noticing, it's about getting annoyed enough to fix it 🤣

Flakiness has been good enough recently, that I turned on notification on every message in #allerts recently. @elsirion

Issue #4364, opened 2024-02-19
Don't allow guardians to choose a password but auto-generate it


elsirion:
Humans are really bad at choosing high-entropy passwords, we should instead generate a password, require the guardian to write it down/save it and the request it again at the next step. 16 alphanumeric, non-case sensitive characters should be more than enough (83bit of security).

elsirion:
Let's close in favor of https://github.com/fedimint/ui/issues/393

PR #4491, opened 2024-03-07
Add functionality to generate random password

isaack-njama:
This PR addresses #4364, generating a random password for now if one is not provided to encrypt sensitive config files

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -38,7 +38,9 @@ use fedimint_mint_client::{MintClientInit, MintClientModule, SpendableNote};
 use fedimint_server::config::io::SALT_FILE;
 use fedimint_wallet_client::api::WalletFederationApi;
 use fedimint_wallet_client::{WalletClientInit, WalletClientModule};
-use rand::thread_rng;
+use rand::distributions::Alphanumeric;
+use rand::seq::IteratorRandom;
+use rand::{thread_rng, Rng};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


fedimintd/src/fedimintd.rs
```diff
@@ -300,9 +304,7 @@ async fn run(
     // TODO: Fedimintd should use the config gen API
     // on each run we want to pass the currently passed password, so we need to
     // overwrite
-    if let Some(password) = opts.password {
```
dpc, line 303: Do we have any protection against overwriting old password? Is it going to get overwritten every time now?

isaack-njama: No protection to overwrite old password. Will need to check if a password exists before setting a new one.

> Do we have any protection against overwriting old password? Is it going to get overwritten every time now with a new random one?

fedimint-cli/src/lib.rs
```diff
@@ -320,6 +321,15 @@ async fn load_or_generate_mnemonic(db: &Database) -> Result<Mnemonic, CliError>
     )
 }
 
+fn generate_random_password() -> String {
+    const PASSWORD_LENGTH: usize = 16;
+    thread_rng()
+        .sample_iter(&Alphanumeric)
```
elsirion, line 327: We should choose one case (upper or lower) and remove characters that look similar (`1`, `I`, `l`, …).

fedimintd/src/fedimintd.rs
```diff
@@ -344,3 +353,12 @@ async fn spawn_metrics_server(
...
     Ok(())
 }
+
+fn generate_random_password() -> String {
+    const PASSWORD_LENGTH: usize = 16;
+    rand::thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(PASSWORD_LENGTH)
+        .map(char::from)
+        .collect()
```
elsirion, line 363: We should not duplicate this code

elsirion (changes requested):
I don't think this is the right approach. How does the guardian learn of the password if it's generated by the server?

I think the much easier solution is to generate a password in the admin UI and just not letting users set it there. If a technical user wants to set a custom password via `fedimint-cli` or the env var I'm ok with that.

elsirion:
I think we should close this PR and just build it in the UI repo. See my reasoning above. Thoughts? @justinmoon @Kodylow

Kodylow:
Agree

PR #4492, opened 2024-03-07
feat: manual shutdown of client

maan2003:
on top #4484


dpc (approved):
:+1:

dpc (approved):

dpc:
Followup: #4535

fedimint-client/src/lib.rs
```diff
@@ -508,92 +507,137 @@ fn states_add_instance(
...
-                break;
-            }
-            // we want to retry fast, give feedback, but not spam
-            if attempt % 100 == 0 {
-                info!("Waiting for ArcClient to stop being used");
-            }
-            fedimint_core::task::sleep(Duration::from_millis(10)).await;
+    /// Shutdown the client.
+    /// Returns false if there are other clones of [`ClientHandle`].
+    pub async fn shutdown(self) -> bool {
```
dpc, line 552: I feel like this should return a result...

maan2003: maybe Result<(), ClientHandle>

dpc: Yeap.

BTW. That `Client` with stopped reactor is an extra flag and state, that I'm not sure if it's useful for anything ever. Maybe we should cut it. It's easier if "handle exist == it's running" always holds. @maan2003

dpc (approved):

justinmoon (approved):

Issue #4395, opened 2024-02-22
Deserialization DoS protection tracking


elsirion:
* [ ] Fuzz deserialization
* [ ] Use limited readers
* [ ] ~~Prevent leftover input https://github.com/fedimint/fedimint/pull/4494/files#r1516713082~~ this requires knowing the reader's size upfront, which can't be done on a generic reader.

PR #4494, opened 2024-03-07
feat: fuzzing

dpc:
How to use: see ./docs/fuzzing.md

### DIFF
--- a/docs/fuzzing.md
+++ b/docs/fuzzing.md
@@ -1,0 +1,44 @@
+### How
+
+Enter dev shell with fuzzing support:
+
+```
+nix develop .#fuzz
+```
+
+Select fuzzing target:
+
+```
+just fuzz-target
+```
+
+Starting fuzzing:
+
+```
+just fuzz-target <target>
+```
+
+If you have found a crash:
+
+```
+just fuzz-target-debug <target>
+```
+
+When `lldb` shows up:
+
+List the frames in the stack-trace:
+
+```
+bt
+```
+
+Select the one with a crash/panic/assert:
+
+```
+frame 8
+```
+
+Inspect the variables and use other debugger facilities, check lldb/gdb tutorial
+for more info on that.
+
+See corresponding `just` rules for details.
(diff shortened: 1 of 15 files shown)


fuzz/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+pub fn test_decodable<T>(data: &[u8])
+where
+    T: Decodable + Encodable,
+{
+    if let Ok(v) = T::consensus_decode(&mut &data[..], &Default::default()) {
+        let encoded_vec = v.consensus_encode_to_vec();
+        // helps debugging to have it standalone
+        let encoded = encoded_vec.as_slice();
+
+        assert_eq!(&data[..encoded.len()], encoded);
```
dpc, line 12: @elsirion As things are `consensus_decode` will not fail on extra leftover input... . Which is fine when they are used internally, but on e2e case it's probably not what the user wants and could lead to some serious bugs.

Will add to #4395

docs/fuzzing.md
```diff
@@ -0,0 +1,44 @@
...
+just fuzz-target <target>
+```
+
+If you have found a crash:
+
+```
+just fuzz-target-debug <target>
+```
+
+When `lldb` shows up:
```
maan2003, line 27: gdb has better rust support right?

dpc: Not sure. I tried both and couldn't tell a difference.

It defaults to `rust-lldb` which lowers to lldb. https://github.com/rust-fuzz/honggfuzz-rs?tab=readme-ov-file#hfuzz_debugger

Kodylow (approved):
tACK

justinmoon (approved):

fuzz/src/lib.rs
elsirion: I did add checks against that in some places I remember, but surely forgot it in others.

elsirion:
Awesome! :tada:

fuzz/src/lib.rs
dpc: @elsirion I looked into it, and it's not possible to do it on `Encodable` trait level, because `Read` does not have a fixed length known in advance. So it always need to be done on the level that holds `&[u8]` or something like that.

elsirion: Yes, does it make sense otherwise? I was thinking of the case where we have a `Cursor<&[u8]>` and can make sure that the cursor was read till the end (at the callsite of decode).

PR #4496, opened 2024-03-07
fix: wait_for_all_recoveries

maan2003:


fedimint-client/src/lib.rs
```diff
@@ -1491,7 +1491,7 @@ impl Client {
             .wait_for(|in_progress| {
                 !in_progress
```
maan2003, line 1492: this is hard, my head hurts
```suggestion
                in_progress
```

dpc (approved):

dpc:
There's only so many combinations, we'll get it right eventually.

bradleystachurski (approved):

fedimint-client/src/lib.rs
```diff
@@ -1489,9 +1489,9 @@ impl Client {
         let mut recovery_receiver = self.client_recovery_progress_receiver.clone();
         recovery_receiver
             .wait_for(|in_progress| {
-                !in_progress
+                in_progress
                     .iter()
-                    .any(|(_id, progress)| !progress.is_done())
+                    .all(|(_id, progress)| progress.is_done())
```
elsirion, line 1494: This only improves readability, there is no functional change here.

elsirion (approved):

fedimint-client/src/lib.rs
maan2003: >There's only so many combinations, we'll get it right eventually.

lol we are repeating combinations now

dpc: Me and Manmeet playing with `!`, `any` and `all`.

image

justinmoon: lmaoooo

Issue #287, opened 2022-07-21
Add fuzz testing


elsirion:
Add fuzz testing for:

* Parsing logic receiving untrusted input
* Network stack
* maybe some consensus logic/modules

[See rust-bitcoin for examples.](https://github.com/rust-bitcoin/rust-bitcoin/tree/master/fuzz/fuzz_targets)

NicolaLS:
After doing some research I think that we can use [hongfuzz](https://github.com/rust-fuzz/honggfuzz-rs) this is also the fuzzer rust-bitcoin uses in their CI (they also have `afl` and `cargo-fuzz` available but I'd just go with `hong_fuzz`
Fuzzing e. g. `client-lib utils functions` will be easy, but I wanted to start fuzzing the API endpoints of the `fedimintd` server.
I haven't really found a good approach of doing this..My last thought was to expose the handlers (the closures) in some way (maybe in a `HashMap<'static str, FnOnce>` or something like that (It feels like a dumb take while writing this (also that is not correct syntax I think) but the idea comes across. Any ideas @elsirion @justinmoon @jkitman ?

justinmoon:
I don't know anything about fuzzing. @m1sterc001guy any thoughts here?

m1sterc001guy:
Haven't had time to sufficiently look into this to have an opinion. Why can't the fedimintd endpoint be directly fuzzed instead of the handlers? @NicolaLS let me know if there's something I can help test.

wilfredallyn:
I saw a blogpost about [fuzzing in BDK](https://afilini.com/blog/fuzz-your-mother-if-you-want-fuzz/) that could be helpful.

I also saw that [LDK](https://github.com/lightningdevkit/rust-lightning/tree/main/fuzz) had a helpful readme about their fuzzing

elsirion:
> Why can't the fedimintd endpoint be directly fuzzed instead of the handlers?

Wouldn't that increase latency while possibly reducing the capability of the fuzzer to learn because now there's all that HTTP stuff in between it and the thing we want to test?

NicolaLS:
> I saw a blogpost about [fuzzing in BDK](https://afilini.com/blog/fuzz-your-mother-if-you-want-fuzz/) that could be helpful.
> 
> I also saw that [LDK](https://github.com/lightningdevkit/rust-lightning/tree/main/fuzz) had a helpful readme about their fuzzing

ty, good information

NicolaLS:
[...]

PR #4501, opened 2024-03-08
chore: add DoS protection against infinite reader on decoding 

### LINKED ISSUE #4395: Deserialization DoS protection tracking

dpc:
This is based on the same API/traits from rust-bitcoin:

https://docs.rs/bitcoin/latest/bitcoin/consensus/encode/trait.Decodable.html#method.consensus_decode_from_finite_reader

Fix #4395

On top of #287 .

A problem was incidentally discovered when fuzzing the whole thing and the fix already has a merge conflict, so is tagging along.

### DIFF
--- a/fedimint-core/src/encoding/mod.rs
+++ b/fedimint-core/src/encoding/mod.rs
@@ -113,13 +113,74 @@ pub trait Encodable {
     }
 }
 
+/// Maximum size, in bytes, of data we are allowed to ever decode
+/// for a single value.
+pub const MAX_DECODE_SIZE: usize = 16_000_000;
+
 /// Data which can be encoded in a consensus-consistent way
 pub trait Decodable: Sized {
-    /// Decode an object with a well-defined format
+    /// Decode `Self` from a size-limited reader.
+    ///
+    /// Like `consensus_decode` but relies on the reader being limited in the
+    /// amount of data it returns, e.g. by being wrapped in
+    /// [`std::io::Take`].
+    ///
+    /// Failing to abide to this requirement might lead to memory exhaustion
+    /// caused by malicious inputs.
+    ///
+    /// Users should default to `consensus_decode`, but when data to be decoded
+    /// is already in a byte vector of a limited size, calling this function
+    /// directly might be marginally faster (due to avoiding extra checks).
+    ///
+    /// ### Rules for trait implementations
+    ///
+    /// * Simple types that that have a fixed size (own and member fields),
+    ///   don't have to overwrite this method, or be concern with it, should
+    ///   only impl `consensus_decode`.
+    /// * Types that deserialize based on decoded untrusted length should
+    ///   implement `consensus_decode_from_finite_reader` only:
+    ///   * Default implementation of `consensus_decode` will forward to
+    ///     `consensus_decode_bytes_from_finite_reader` with the reader wrapped
+    ///     by `Take`, protecting from readers that keep returning data.
+    ///   * Implementation must make sure to put a cap on things like
+    ///     `Vec::with_capacity` and other allocations to avoid oversized
+    ///     allocations, and rely on the reader being finite and running out of
+    ///     data, and collections reallocating on a legitimately oversized input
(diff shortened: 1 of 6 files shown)


fedimint-core/src/encoding/mod.rs
```diff
@@ -113,13 +113,74 @@ pub trait Encodable {
...
 }
 
+/// Maximum size, in bytes, of data we are allowed to ever decode
+/// for a single value.
+pub const MAX_DECODE_SIZE: usize = 16_000_000;
+
 /// Data which can be encoded in a consensus-consistent way
 pub trait Decodable: Sized {
-    /// Decode an object with a well-defined format
+    /// Decode `Self` from a size-limited reader.
```
elsirion, line 122: I have a bad feeling about this new API, it took me a few minutes to understand how to use it, I can imagine this to be the cause for some confusion in the future, especially with the circular default impls.

At the same time I don't have a great idea hoe to make it better, so might just go with it and see how bad it actually is.

I think a trait-level comment stating "You typically implement `consensus_decode_from_finite_reader` and call `consensus_decode_from_finite_reader` in these implementations. For normal decoding outside `Decodable` implementations you likely want to use `consensus_decode`" would go a long way.

fedimint-core/src/encoding/mod.rs
```diff
@@ -759,19 +838,28 @@ where
...
-            let v = V::consensus_decode(d, modules)?;
-            if res.insert(amt, v).is_some() {
+            let k = K::consensus_decode_from_finite_reader(d, modules)?;
+            if prev_key
+                .as_ref()
+                .map(|prev_key| &k <= prev_key)
+                .unwrap_or_default()
+            {
+                return Err(DecodeError::from_str("Non-canonical encoding"));
+            }
```
elsirion, line 859: Great catch!

fedimint-core/src/encoding/mod.rs
```diff
@@ -795,16 +883,25 @@ where
...
-            let k = K::consensus_decode(d, modules)?;
+            let k = K::consensus_decode_from_finite_reader(d, modules)?;
+            if prev_key
+                .as_ref()
+                .map(|prev_key| &k <= prev_key)
+                .unwrap_or_default()
+            {
+                return Err(DecodeError::from_str("Non-canonical encoding"));
+            }
+            prev_key = Some(k.clone());
```
elsirion, line 904: I'm wondering if we should imple this as `BTreeMap<K, ()>::consensus_decode_from_finite_reader(…)`.

fedimint-core/src/encoding/mod.rs
```diff
@@ -966,15 +1063,17 @@ impl<T> Decodable for DynRawFallback<T>
...
-            fedimint_core::core::ModuleInstanceId::consensus_decode(reader, decoders)?;
+            fedimint_core::core::ModuleInstanceId::consensus_decode_from_finite_reader(
+                reader, decoders,
+            )?;
         Ok(match decoders.get(module_instance_id) {
             Some(decoder) => {
-                let total_len_u64 = u64::consensus_decode(reader, decoders)?;
+                let total_len_u64 = u64::consensus_decode_from_finite_reader(reader, decoders)?;
                 let mut reader = reader.take(total_len_u64);
                 let v: T = decoder.decode(&mut reader, module_instance_id, decoders)?;
```
elsirion, line 1078: This value being ridiculously high won't be a problem because the underlying reader is already limited.

elsirion (approved):

fedimint-core/src/encoding/mod.rs
dpc: @elsirion FWIW this exact system has been implemented in rust-bitcoin for more than a year now I think and no one complained yet. :D

maan2003 (approved):

PR #4502, opened 2024-03-08
chore(ci): increase timeout on code coverage run in CI

dpc:
We're seeing failures due to these being slow, like:

https://github.com/fedimint/fedimint/actions/runs/8198894922/job/22423145415

Not sure if got slower recently, or something else is wrong, but it's blocking the CI.


dpc:
Actually, it was me.

bradleystachurski (dismissed):
@dpc I think we can still merge this to unblock CI while we find a different approach for the discover api timeout in tests.

dpc:
I think we should https://github.com/fedimint/fedimint/pull/4503 instead.

PR #4503, opened 2024-03-08
fix: decrease api discovery timeouts in dev shell and tests

dpc:
We started getting slow test and timeouts in the CI:

https://github.com/fedimint/fedimint/actions/runs/8198894922/job/22423145415

caused by #4476.

Turns out there are other tests just degraded federation tests that run with degraded federations.

Generally in dev shell, tests and CI, we can just default to 10 seconds timeout, should be enough even under heavy load.


bradleystachurski (approved):

dpc:
@elsirion

maan2003 (approved):

PR #4504, opened 2024-03-08
chore(client): on first start, get api version from half of clients

dpc:
This blocking of client on a first start is slowing everything down and creating a lot of problems. A simple solution would be to go fast and optimistic on the first start.

Subsequent runs will happen in the background, so then we can do slow, but more precise negotiation.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -179,6 +179,13 @@ pub enum AddStateMachinesError {
     Other(#[from] anyhow::Error),
 }
 
+pub enum DiscoverCommonApiVersionMode {
+    /// Get the response from only a few peers, or until a timeout
+    Fast,
+    /// Try to get a reasponse from all peers, or until a timeout
+    Full,
+}
+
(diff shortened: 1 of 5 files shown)


dpc:
:scratching-head: , somehow it sometimes panicks, and I don't understand why. Will do more debugging tomorrow. The idea seems good, but there are probably some `!`, `any` or `all` I need to tweak here. ;)

fedimint-client/src/lib.rs
```diff
@@ -179,6 +179,13 @@ pub enum AddStateMachinesError {
     Other(#[from] anyhow::Error),
 }
 
+pub enum DiscoverCommonApiVersionMode {
+    /// Get the response from only a few peers, or until a timeout
+    Fast,
+    /// Try to get a reasponse from all peers, or until a timeout
```
Kodylow, line 185: typo

Kodylow (approved):
just a typo, ACK, reviewed live with DPC

PR #4506, opened 2024-03-08
feat: Rename federation_id() to calculate_federation_id()

benthecarman:
The name federation_id() makes it sound like it is just retrieving the id, however it is actually calculating a hash and can take a mild amount of cpu power. Renaming this to calculate_federation_id() implies that there is some work being done and the user should instantiate it as a variable to not duplicate that work.

Found a place where this was happening in the codebase already. We did this in several places in Mutiny too.


dpc (approved):

Kodylow (approved):
ACK

PR #4511, opened 2024-03-08
chore: add `fedimint-empty` module

dpc:
Initiaially the goal of `fedimint-dummy` module was to be a good template to start a new module.

But with time, it was hijacked to be become `fedimint-`"print me money out of thin air" module (aka `fedimint-federal-reserve` module), and it now actually contains quite a bit of code.

So when trying to use it as a starting point, one needs to go a rename, delete and fix more code than neccessary.

This module is a clone of `fedimint-dummy`, but with almost everything removed.

It will get compiled, but will not be used for anything anywhere.


m1sterc001guy (approved):
Maybe this should be in the custom module repo?

dpc:
> Maybe this should be in the custom module repo?

Then it will be broken all the time. Maybe. Not sure if we're stable enough yet for this not to be the case.

m1sterc001guy:
> Then it will be broken all the time

Yeah I get that. It's just hard to ask external module developers to do the same if we can't maintain the empty module

dpc:
> Yeah I get that. It's just hard to ask external module developers to do the same if we can't maintain the empty module

Module developers could just `cp -a upstream-fedimint ...` etc. Or we can do that from time to time, when we update the example repo. Having this module in the tree makes sure it's *always* working, not just when we update other repo.

maan2003 (approved):

PR #4512, opened 2024-03-09
Refactor such that we can derive Encodable and Decodable for crypto crates APIs

joschisan:
Previously we derived encodable and decodable by hand for all the public types of tpe. However, with new crypto also built on bls12_381 like lightning-ng and ecash-ng that has more complex public structs this becomes increasingly cumbersome and repetitious. Hence, we refactor the code to be able to derive Encodable and Decodable for all future crypto crates built on bls12_381.

In a next step we want to do the same with serde Serialize and Deserialize by upstreaming an implementation to bls12_381 such that we can derive the traits as well. Notice that we duplicate the serde implementation for Scalar, G1, G2 as we copy it from tbs to core were we need it as well. We leave it in tbs though as the copy in core is supposed to be removed after we upstreamed it.


fedimint-core/src/encoding/threshold_crypto.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    #[test_log::test]
+    fn test_ciphertext() {
+        let sks = threshold_crypto::SecretKeySet::random(1, &mut rand::thread_rng());
+        let pks = sks.public_keys();
+        let pk = pks.public_key();
+
+        let message = b"Hello world!";
+        let ciphertext = pk.encrypt(message);
+        let decryption_share = sks.secret_key_share(0).decrypt_share(&ciphertext).unwrap();
```
m1sterc001guy, line 116: Nit: use `expect` instead of `unwrap`

fedimint-core/src/encoding/bls12_381.rs
```diff
@@ -0,0 +1,88 @@
...
+    ) -> Result<Self, crate::encoding::DecodeError> {
+        let mut bytes = [0u8; 48];
+
+        d.read_exact(&mut bytes)
+            .map_err(crate::encoding::DecodeError::from_err)?;
+
+        let point = G1Affine::from_compressed(&bytes);
+
+        if point.is_some().unwrap_u8() == 1 {
+            Ok(point.unwrap())
```
m1sterc001guy, line 53: Nit: use `expect`

fedimint-core/src/bls12_381_serde.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        let mut byte_array = [0u8; 32];
+
+        byte_array.copy_from_slice(&bytes);
+
+        let scalar = Scalar::from_bytes(&byte_array);
+        // FIXME: probably safe with public data, but doesn't look nice
+        if scalar.is_some().into() {
+            Ok(scalar.unwrap())
```
m1sterc001guy, line 36: Nit: `expect`

m1sterc001guy (approved):
Someone with more context than me should also probably look, but LGTM

dpc:
@elsirion

dpc (commented):
On a lower level (mechanicaly what's goin on) it look OK, but on the higher level I'm lost.

The legacy vs new encoding seems like a complication, but for a good reason.

I also don't understand what's the status of "upstream serde support" or what exactly it means here.

@elsirion is probably best person to review

elsirion:
Need to take a look, it's less of a priority than other things we might want to get into 0.3 since it's mostly useful for LN and e-cash NG, which are both longer-term research projects.

fedimint-core/src/encoding/bls12_381.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl Decodable for Scalar {
+    fn consensus_decode<D: std::io::Read>(
+        d: &mut D,
+        _modules: &ModuleDecoderRegistry,
+    ) -> Result<Self, crate::encoding::DecodeError> {
+        let mut bytes = [0u8; 32];
+
+        d.read_exact(&mut bytes)
```
elsirion, line 19: Why use a different function for decoding than encoding? You could use `<[u8; 32] as Decodable>::consensus_decode(…)` instead (or a variation thereof).

The problem here is that you make assumptions about the inner workings of `<[u8; 32] as Encodable>::consensus_encode` by re-implementing the corresponding decoding functionality manually.

fedimint-core/src/encoding/bls12_381.rs
```diff
@@ -0,0 +1,88 @@
...
+            .map_err(crate::encoding::DecodeError::from_err)?;
+
+        let scalar = Scalar::from_bytes(&bytes);
+
+        if scalar.is_some().unwrap_u8() == 1 {
```
elsirion, line 24: Same improvement is possible as mentioned above.

fedimint-core/src/encoding/bls12_381.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+impl Encodable for G1Affine {
+    fn consensus_encode<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, std::io::Error> {
+        self.to_compressed().consensus_encode(writer)
+    }
+}
+
+impl Decodable for G1Affine {
```
elsirion, line 40: Same comments as for `Scalar` apply here too.

fedimint-core/src/encoding/bls12_381.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+impl Encodable for G2Affine {
+    fn consensus_encode<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, std::io::Error> {
+        self.to_compressed().consensus_encode(writer)
+    }
+}
+
+impl Decodable for G2Affine {
```
elsirion, line 68: Same comments as for `Scalar` apply here too.

fedimint-core/src/encoding/threshold_crypto.rs
```diff
@@ -0,0 +1,121 @@
+use std::io::{Error, Read, Write};
+
+use threshold_crypto::group::Curve;
+use threshold_crypto::{G1Affine, G1Projective};
+
+use crate::encoding::{Decodable, DecodeError, Encodable};
+use crate::module::registry::ModuleDecoderRegistry;
+
+impl Encodable for threshold_crypto::PublicKeySet {
```
elsirion, line 9: Can't we just derive all this now? Care would need to be taken not to break anything, but worth looking into imo (follow-up, not this PR).

fedimint-core/src/encoding/threshold_crypto.rs
```diff
@@ -0,0 +1,121 @@
+use std::io::{Error, Read, Write};
```
elsirion, line 1: I take it you just copied this whole file over.

dpc (dismissed):

fedimint-core/src/bls12_381_serde.rs
```diff
@@ -0,0 +1,90 @@
...
+        } else {
+            Err(D::Error::custom("Could not decode compressed G1Affine"))
+        }
+    }
+}
+
+macro_rules! impl_serde_g {
```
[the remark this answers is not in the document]
joschisan: Done

fedimint-core/src/encoding/bls12_381.rs
joschisan: Done

fedimint-core/src/encoding/bls12_381.rs
joschisan: Done

fedimint-core/src/encoding/bls12_381.rs
joschisan: Done

fedimint-core/src/encoding/bls12_381.rs
joschisan: Done

fedimint-core/src/encoding/threshold_crypto.rs
joschisan: Yes

dpc (approved):

fedimint-core/src/bls12_381_serde.rs
```diff
@@ -0,0 +1,90 @@
...
+            s.serialize_bytes(&bytes)
+        }
+    }
+
+    pub fn deserialize<'d, D: Deserializer<'d>>(d: D) -> Result<Scalar, D::Error> {
+        let bytes: Vec<u8> = if d.is_human_readable() {
+            let deser: String = Deserialize::deserialize(d)?;
+            Vec::<u8>::from_hex(&deser).map_err(serde::de::Error::custom)?
+        } else {
+            Deserialize::deserialize(d)?
```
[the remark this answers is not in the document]
joschisan: This is just copy and paste the old code from tbs for now.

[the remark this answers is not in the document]
elsirion: I was confused, `serialize_bytes` always prepends the size, that's why deserializing a `Vec<u8>` works here.

elsirion (approved):

PR #4513, opened 2024-03-09
 chore: `meta` module

dpc:
## Overview

Meta module is meant to allow guardians to collectively setup and change over time a set of values for various auxiliary purposes, e.g.: establish customization of Federation like welcome messages, ToS and other legal documents, references to community information, chat servers, etc. This way the end user Fedimint client applications can discover these settings and customize the user experience per-federation. 

It acts as a simple key-value store, where each guardian can submit a proposed value for a given key, and if `threshold` number of peers submit the same value, it becomes an official consensus.

This module is intentionally simple and limited in scope.

The keys are effectively a single byte, so there can only ever be 256 of them. Server side module does not interpret these values in any way and treats them as opaque binary blobs.

It is up to the community of Fedimint end user applications to establish the meaning and format of meta module values. The initial intention is to use a single key with the value of a json string. The structure of that JSON document is open-ended and can be estabilished and evolved over time, and applications can gracefully handle it in backward and forward compatible way.

This is all to move the more social consensus around Federation management outside / on top-of the hard to evolve Fedimint consensus.

The extra 255 keys are an escape hatch - to avoid dealing with module consensus changes if a single value ever turns out to not be enough.

We should follow up with some partitioning and allocation of these keys, and the Json schema for ones deemed to be used as Json (default one 0 at least).

## Testing

`just mprocs` and then try:

```
fedimint-cli module meta help
fedimint-cli --our-id 1 --password asdf module meta submit '{ "a": "b"}'
fedimint-cli --our-id 1 --password asdf module meta get-submissions
```

etc.

See if you can figure it out. :)

### References

* https://github.com/fedimint/fedimint/tree/master/docs/meta_fields

### Status

I think everything is there. Now needs some feedback, tests, fixes and merging.

For reviewers - there are some nifty details and improvements that I accumulated along the way. Some of them pushed as separate PRs already, some integral to the change.

TODO:

* [ ] write some (devimint?) tests
* [ ] fix back-compat - probably something about new module


dpc:
Status: server side is kind of ready, but needs some client side functions to make this usable and at least some basic tests.

Feel free to already review.

dpc:
Seems like back compat is failing because v0.2 can't handle unknown modules.

Should be fixed with a quick backport: https://github.com/fedimint/fedimint/pull/4533 , but we might need a new patch release and switch our backcompat tests to use it?

m1sterc001guy:
Needs rebase

dpc:
@m1sterc001guy rebased

fedimint-client/src/module/init.rs
```diff
@@ -107,13 +117,15 @@ where
     C: ClientModuleInit,
 {
     federation_id: FederationId,
+    num_peers: NumPeers,
```
elsirion, line 120: future work: Now that we have this argument here we should check if we can remove the `NumPeersExt` impl for key sets etc.

modules/fedimint-meta-server/src/lib.rs
```diff
@@ -0,0 +1,509 @@
...
+                    salt,
+                })
+            {
+                to_delete.push(key);
+            } else if Self::get_consensus(dbtx, key).await.as_ref() == Some(&value) {
+                // When a value is new and equal to current consensus, it should both: be
+                // submitted so it can clear/cancel previous submissions, and
+                // deleted (afterwards).
+                to_delete.push(key);
+                to_submit.push(MetaConsensusItem { key, value, salt });
```
elsirion, line 236: Unclear why we *need* to submit again, but ok I guess? Others will likely filter it out if it's a dup?

modules/fedimint-meta-server/src/db.rs
```diff
@@ -0,0 +1,96 @@
...
+
+#[derive(Debug, Encodable, Decodable)]
+pub struct MetaDesiredKeyPrefix;
+
+#[derive(Debug, Encodable, Decodable, Serialize, PartialEq, Eq)]
+pub struct MetaDesiredValue {
+    // Since AlephBft will merge and not re-submit the exact same item twice within one session,
+    // changing submitted item in sequence `a -> b -> a` will simply ignore the second `a`.
+    // To avoid this behavior, an otherwise meaningless `salt` field is used.
+    pub salt: u64,
```
elsirion, line 39: @joschisan something we might want to change since it's quite a foot gun

modules/fedimint-meta-common/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+/// A [`MetaValue`] in a consensus (which means it has a revision number)
+#[derive(Debug, Clone, Encodable, Decodable, Serialize, Deserialize, PartialEq, Eq)]
+pub struct MetaConsensusValue {
+    pub revision: u64,
+    pub value: MetaValue,
+}
+
+/// Input for a fedimint transaction
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
+pub struct MetaInput;
```
elsirion, line 178: Should this maybe be a variantless enum?

modules/fedimint-meta-client/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+use fedimint_core::{apply, async_trait_maybe_send, Amount, PeerId};
+pub use fedimint_meta_common as common;
+use fedimint_meta_common::{MetaCommonInit, MetaModuleTypes};
+use states::MetaStateMachine;
+use strum::IntoEnumIterator;
+
+#[derive(Debug)]
+pub struct MetaClientModule {
+    module_api: DynModuleApi,
+    admin_auth: Option<ApiAuth>,
```
elsirion, line 32: future: Should this be given via some context struct given to `handle_cli_command`?

fedimintd/src/fedimintd.rs
```diff
@@ -188,8 +189,13 @@ impl Fedimintd {
...
             .with_module(MintInit)
-            .with_module(WalletInit)
-            .with_module(MetaInit);
+            .with_module(WalletInit);
+
+        let s = if !is_env_var_set(FM_DISABLE_META_MODULE_ENV) {
+            s.with_module(MetaInit)
+        } else {
+            s
+        };
```
elsirion, line 198: Not great, but ok for now. Maybe our tests should just specify which modules they want during setup?

elsirion (dismissed):
LGTM, let's get this in asap so we can cut 0.3-rc.0.

We should probably provide some helper function which scans all the possible meta sources with a given precedence.

modules/fedimint-meta-client/src/lib.rs
dpc: It is given via `init` or `recovery`, so no need pass it via `handle_cli_command`, IMO.

modules/fedimint-meta-server/src/lib.rs
dpc: Yeah. If it's a dup, it will be ignored. If it isn't it will clear guardian's previous submission to match the consensus.

fedimint-client/src/module/init.rs
dpc: :+1:

Was never a fan of NumPeersExt on collections etc. Very surprising when reading code without LSP.

modules/fedimint-meta-server/src/db.rs
dpc: Seems like some stuff already already relies on this behavior (including this PR), so ... not sure if we can change it now.

Does not seem like a big problem, but it is a footgun, so we should document it loudly in many places.

fedimintd/src/fedimintd.rs
dpc: The end goal, IMO, is DKG/tests etc. specifying what module instances they want, without anything being enabled "by default" (other than maybe in the DKG UI level).

But even then we are going to need conditional logic in some compatibility tests to not enable certain modules (or skip test altogether if it requires it), if previous version simply can't handle given module. It will however remove a need for disabling some modules via env vars, and make it as an explicit step.

modules/fedimint-meta-client/src/lib.rs
elsirion: But should it? That way every module will have to keep its own copy. Not a big deal though either way.

fedimint-client/src/module/init.rs
elsirion: I think the reason for it was that we didn't have the `Args` struct and didn't want to add yet another argument but reuse existing ones. Since that's not a concern anymore let's get rid of it.

modules/fedimint-meta-client/src/lib.rs
dpc: Only modules that need it will keep a copy...

But I got confused before and I thought we're talking about the `module_api` and `global_api`, which are now different (have only one peer in them) when the client is "guardian client (`--our-id` passed)".

I guess is now an architectural question now:

Is the whole `Client` started in a "guardian mode", or is it just a single command? Because in `fedimint-cli` we build a whole client for a single command every time, so these are currently conflated and hard to tell.

E.g. would it make sense to have "guardian only State Machines"? So guardians could kickstart longer running things on their clients that would make guardian-only API calls?

@elsirion 

I kind of thing it makes sense to support it eventually, but didn't made any drastic moves yet.

Also there's still the core question from https://github.com/fedimint/fedimint/issues/4519 : if you're a `Client` started by the guardian, should we even bother calling other guardians if you can fully trust your own? Maybe for redundancy ... ? In that case we'll need to have `all_peers` + `guardian_peer` and on the traits, and use one or the other on appropriate calls.

elsirion: > E.g. would it make sense to have "guardian only State Machines"? So guardians could kickstart longer running things on their clients that would make guardian-only API calls?

Could be useful eventually, but low priority imo.

> Also there's still the core question from https://github.com/fedimint/fedimint/issues/4519 : if you're a Client started by the guardian, should we even bother calling other guardians if you can fully trust your own? Maybe for redundancy ... ? In that case we'll need to have all_peers + guardian_peer and on the traits, and use one or the other on appropriate calls.

As it stands right now we have to only use the guardian peer if authentication is present because we leak our secret otherwise? Ideally we could tell the admin functions to only call the guardian API while all others act normally. There might be cases where a guardian does maintenaince but still wants to be able to transact.

dpc: > As it stands right now we have to only use the guardian peer if authentication is present because we leak our secret otherwise?

We only send auth with rpcs that require it, and these use separate methods.

>  There might be cases where a guardian does maintenaince but still wants to be able to transact.


Yeah, so we'll need the trait to have separate `all_peers` for normal calls and `our_peer_id() -> Option<PeerId>` for own guardian APIs. Small refactor, I believe.

dpc: > Could be useful eventually, but low priority imo.

We don't have to do it. Module authors probabl already can because they get everything in `init`&`recovery` args.

modules/fedimint-meta-client/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct MetaClientInit;
+
+// TODO: Boilerplate-code
+#[apply(async_trait_maybe_send!)]
+impl ModuleInit for MetaClientInit {
+    type Common = MetaCommonInit;
+    const DATABASE_VERSION: DatabaseVersion = DatabaseVersion(2);
```
bradleystachurski, line 151: I don't think this breaks anything, but should we start at 0?

modules/fedimint-meta-common/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+impl From<&[u8]> for MetaValue {
+    fn from(value: &[u8]) -> Self {
+        Self(value.to_vec())
+    }
+}
+
+impl MetaValue {
+    /// Maximum size of a [`MetaValue`]
+    /// More than 1MB would lead to problems.
+    pub const MAX_LEN_BYTES: usize = 1024 * 1024 * 1024;
```
bradleystachurski, line 91: 👍

modules/fedimint-meta-server/src/lib.rs
```diff
@@ -0,0 +1,509 @@
...
+
+/// Generates the module
+#[derive(Debug, Clone)]
+pub struct MetaInit;
+
+// TODO: Boilerplate-code
+#[async_trait]
+impl ModuleInit for MetaInit {
+    type Common = MetaCommonInit;
+    const DATABASE_VERSION: DatabaseVersion = DatabaseVersion(1);
```
bradleystachurski, line 61: Similar question, should we start at 0?

bradleystachurski (approved):
Looks good! No blocking comments. A devimint test would be excellent but can be addressed in a followup.

modules/fedimint-meta-client/src/lib.rs
dpc: Yes. This is what I get for copying it from existing module. Good catch. Will follow-up.

fedimint-client/src/lib.rs
```diff
@@ -1998,6 +2009,17 @@ impl ClientBuilder {
         Ok(client)
     }
 
+    fn admin_api_from_id(id: PeerId, cfg: &ClientConfig) -> anyhow::Result<DynGlobalApi> {
+        let url = cfg
+            .global
+            .api_endpoints
+            .get(&id)
+            .ok_or_else(|| anyhow::format_err!("Invalid admin peer_id"))?
```
douglaz, line 2017:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("Invalid admin peer_id: {id:?}"))?
```

modules/fedimint-meta-client/src/api.rs
```diff
@@ -0,0 +1,73 @@
...
+use fedimint_core::{apply, async_trait_maybe_send};
+use fedimint_meta_common::endpoint::{
+    GetConsensusRequest, GetSubmissionResponse, GetSubmissionsRequest, SubmitRequest,
+    GET_CONSENSUS_ENDPOINT, GET_CONSENSUS_REV_ENDPOINT, GET_SUBMISSIONS_ENDPOINT, SUBMIT_ENDPOINT,
+};
+use fedimint_meta_common::{MetaConsensusValue, MetaKey, MetaValue};
+
+#[apply(async_trait_maybe_send!)]
+pub trait MetaFederationApi {
+    async fn get_consensus(&self, key: MetaKey) -> FederationResult<Option<MetaConsensusValue>>;
```
m1sterc001guy, line 13: Nit: would be good to get docs on these

m1sterc001guy (approved):

modules/fedimint-meta-client/src/api.rs
m1sterc001guy: I think they're documented in `endpoint.rs` which works for now but was initially unintuitive for me

fedimint-core/src/lib.rs
```diff
@@ -250,38 +250,58 @@ pub enum ParseAmountError {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct NumPeers(usize);
+
+impl From<usize> for NumPeers {
+    fn from(value: usize) -> Self {
+        Self(value)
+    }
+}
+impl NumPeers {
+    pub fn as_usize(self) -> usize {
```
douglaz, line 268:
```suggestion
    pub fn as_usize(&self) -> usize {
```

dpc: `NumPeers` is `Copy`, so no need to `&`

bradleystachurski:
🤔 
```
> perhaps a crate was updated and forgotten to be re-vendored?
```

https://github.com/fedimint/fedimint/actions/runs/8300886514/job/22719760154#step:5:1625

fedimint-core/src/lib.rs
douglaz: yeah, but it's ugly an `as_*` method taking `self`

bradleystachurski (dismissed):

fedimint-core/src/lib.rs
dpc: Good point.

bradleystachurski (approved):

modules/fedimint-meta-server/src/db.rs
```diff
@@ -0,0 +1,96 @@
...
+
+/// Namespaces DB keys for this module
+#[repr(u8)]
+#[derive(Clone, EnumIter, Debug)]
+pub enum DbKeyPrefix {
+    /// How we want to vote
+    ///
+    /// Private, not part of the consensus, but only local state.
+    Desired = 0x00,
+    /// Current consensuson
```
douglaz, line 15:
```suggestion
    /// Current consensus
```

Kodylow:
Just finished testing locally and got a branch of guardian UI to print the updated meta, lgtm!

modules/fedimint-meta-server/src/lib.rs
```diff
@@ -0,0 +1,509 @@
...
+        info!(target: LOG_SERVER_MODULE_META,
+             %peer_id,
+             %key,
+            value_len = %new_value.value.as_slice().len(),
+             matching = %matching_submissions.len(),
+            %threshold, "Peer submitted a value");
+
+        // if threshold or more, change the consensus value
+        if threshold <= matching_submissions.len() {
+            Self::change_consensus(dbtx, key, new_value.value, matching_submissions).await;
```
douglaz, line 358:
```suggestion
        // if threshold or more, change the consensus value
        let will_change_consensus = threshold <= matching_submissions.len() ;
        info!(target: LOG_SERVER_MODULE_META,
             %peer_id,
             %key,
             %will_change_consensus,
            value_len = %new_value.value.as_slice().len(),
             matching = %matching_submissions.len(),
            %threshold, "Peer submitted a value");

        if will_change_consensus {
            Self::change_consensus(dbtx, key, new_value.value, matching_submissions).await;
```

justinmoon (approved):
haven't reviewed thoroughly. re-approving on behalf of previous approvals.

dpc:
Known flake failed.

PR #4514, opened 2024-03-09
feat: add liquidity management API to ILnRpcClient

tvolk131:
Requires #4959

We can use this API to consolidate testing behavior which will help pave the way towards supporting an LDK-based gateway with a lightning node that lives in the gateway process

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -26,7 +26,8 @@ use tracing::{debug, info, trace, warn};
 
 use crate::util::{poll, ClnLightningCli, ProcessHandle, ProcessManager};
 use crate::vars::utf8;
-use crate::{cmd, poll_eq};
+use crate::version_constants::VERSION_0_4_0_ALPHA;
+use crate::{cmd, poll_eq, Gatewayd};
 
 #[derive(Clone)]
 pub struct Bitcoind {
@@ -377,6 +378,8 @@ impl Lightningd {
         Ok(rpc.call_typed(request).await?)
     }
 
+    // TODO(tvolk131): Remove this method and instead use
+    // `Gatewayd.wait_for_chain_sync()` once 0.4.0 is released
     pub async fn await_block_processing(&self) -> Result<()> {
         poll("lightningd block processing", || async {
             let btc_height = self
@@ -514,6 +517,8 @@ impl Lnd {
             .identity_pubkey)
     }
 
+    // TODO(tvolk131): Remove this method and instead use
+    // `Gatewayd.wait_for_chain_sync()` once 0.4.0 is released
     pub async fn await_block_processing(&self) -> Result<()> {
         poll("lnd block processing", || async {
             let synced = self
@@ -541,6 +546,8 @@ impl Lnd {
     }
 }
 
+// TODO(tvolk131): Remove this method and instead use
+// `open_channel_between_gateways()` below once 0.4.0 is released
 pub async fn open_channel(
     process_mgr: &ProcessManager,
     bitcoind: &Bitcoind,
@@ -610,6 +617,111 @@ pub async fn open_channel(
         .map_err(ControlFlow::Continue)
     })
     .await?;
+
+    bitcoind.mine_blocks(10).await?;
+
+    poll("Wait for channel update", || async {
+        let mut lnd_client = lnd.client.lock().await;
+        let channels = lnd_client
+            .lightning()
+            .list_channels(ListChannelsRequest {
+                active_only: true,
+                ..Default::default()
+            })
+            .await
+            .context("lnd list channels")
+            .map_err(ControlFlow::Break)?
+            .into_inner();
+
+        if let Some(channel) = channels
+            .channels
+            .iter()
+            .find(|channel| channel.remote_pubkey == cln_pubkey)
+        {
+            let chan_info = lnd_client
+                .lightning()
+                .get_chan_info(ChanInfoRequest {
+                    chan_id: channel.chan_id,
+                })
+                .await;
+
+            match chan_info {
+                Ok(info) => {
+                    let edge = info.into_inner();
+                    if edge.node1_policy.is_some() {
+                        return Ok(());
+                    } else {
+                        debug!(?edge, "Empty chan info");
+                    }
+                }
+                Err(e) => {
+                    debug!(%e, "Getting chan info failed")
+                }
+            }
+        }
+
+        Err(ControlFlow::Continue(anyhow!("channel not found")))
+    })
+    .await?;
+
+    Ok(())
+}
+
+// TODO(tvolk131): Remove the `cln` and `lnd` arguments and instead use `gw_cln`
+// and `gw_lnd` once version 0.4.0 is released
+pub async fn open_channel_between_gateways(
+    process_mgr: &ProcessManager,
+    bitcoind: &Bitcoind,
+    cln: &Lightningd,
+    gw_cln: &Gatewayd,
+    lnd: &Lnd,
+    gw_lnd: &Gatewayd,
+) -> Result<()> {
+    let gateway_cli_version = crate::util::GatewayCli::version_or_default().await;
+    let gatewayd_version = crate::util::Gatewayd::version_or_default().await;
+    if gateway_cli_version < *VERSION_0_4_0_ALPHA || gatewayd_version < *VERSION_0_4_0_ALPHA {
+        return open_channel(process_mgr, bitcoind, cln, lnd).await;
+    }
+
+    lnd.client
+        .lock()
+        .await
+        .lightning()
+        .update_channel_policy(PolicyUpdateRequest {
+            min_htlc_msat: 1,
+            scope: Some(Scope::Global(true)),
+            time_lock_delta: 80,
+            base_fee_msat: 0,
+            fee_rate: 0.0,
+            fee_rate_ppm: 0,
+            max_htlc_msat: 10000000000,
+            min_htlc_msat_specified: true,
+        })
+        .await?;
+
+    tokio::try_join!(
+        gw_cln.wait_for_chain_sync(bitcoind),
+        gw_lnd.wait_for_chain_sync(bitcoind)
+    )?;
+    info!(target: LOG_DEVIMINT, "block sync done");
+    let cln_addr = gw_cln.get_funding_address().await?;
+
+    bitcoind.send_to(cln_addr, 100_000_000).await?;
+    bitcoind.mine_blocks(10).await?;
+
+    let lnd_pubkey = gw_lnd.lightning_pubkey().await?;
+    let cln_pubkey = gw_cln.lightning_pubkey().await?;
+
+    gw_cln
+        .connect_to_peer(
+            lnd_pubkey.parse()?,
+            format!("{}:{}", "127.0.0.1", process_mgr.globals.FM_PORT_LND_LISTEN),
+        )
+        .await?;
+
+    gw_cln
+        .open_channel(lnd_pubkey.clone(), 10_000_000, Some(5_000_000))
+        .await?;
 
     bitcoind.mine_blocks(10).await?;
 
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -548,7 +548,7 @@ impl Lnd {
 
 // TODO(tvolk131): Remove this method and instead use
 // `open_channel_between_gateways()` below once 0.4.0 is released
-pub async fn open_channel(
+async fn open_channel(
     process_mgr: &ProcessManager,
     bitcoind: &Bitcoind,
     cln: &Lightningd,
--- a/devimint/src/gatewayd.rs
+++ b/devimint/src/gatewayd.rs
@@ -8,7 +8,7 @@ use tracing::info;
 
 use crate::cmd;
 use crate::envs::{FM_GATEWAY_API_ADDR_ENV, FM_GATEWAY_DATA_DIR_ENV, FM_GATEWAY_LISTEN_ADDR_ENV};
-use crate::external::LightningNode;
+use crate::external::{Bitcoind, LightningNode};
 use crate::federation::Federation;
 use crate::util::{poll, Command, ProcessHandle, ProcessManager};
 use crate::vars::utf8;
@@ -114,6 +114,15 @@ impl Gatewayd {
         Ok(gateway_id)
     }
 
+    pub async fn lightning_pubkey(&self) -> Result<String> {
+        let info = cmd!(self, "info").out_json().await?;
+        let gateway_id = info["lightning_pub_key"]
+            .as_str()
+            .context("lightning_pub_key must be a string")?
+            .to_owned();
+        Ok(gateway_id)
+    }
+
     pub async fn connect_fed(&self, fed: &Federation) -> Result<()> {
         let invite_code = fed.invite_code()?;
         poll("gateway connect-fed", || async {
@@ -134,5 +143,72 @@ impl Gatewayd {
             .as_str()
             .context("address must be a string")?
             .to_owned())
+    }
+
+    pub async fn connect_to_peer(&self, pubkey: String, host: String) -> Result<()> {
+        cmd!(
+            self,
+            "lightning",
+            "connect-to-peer",
+            "--pubkey",
+            pubkey,
+            "--host",
+            host
+        )
+        .run()
+        .await?;
+        Ok(())
+    }
+
+    pub async fn get_funding_address(&self) -> Result<String> {
+        let address = cmd!(self, "lightning", "get-funding-address")
+            .out_string()
+            .await?;
+        Ok(address)
+    }
+
+    pub async fn open_channel(
+        &self,
+        pubkey: String,
+        channel_size_sats: u64,
+        push_amount_sats: Option<u64>,
+    ) -> Result<()> {
+        cmd!(
+            self,
+            "lightning",
+            "open-channel",
+            "--pubkey",
+            pubkey,
+            "--channel-size-sats",
+            channel_size_sats,
+            "--push-amount-sats",
+            push_amount_sats.unwrap_or(0)
+        )
+        .run()
+        .await?;
+        Ok(())
+    }
+
+    pub async fn wait_for_chain_sync(&self, bitcoind: &Bitcoind) -> Result<()> {
+        poll("lightning node block processing", || async {
+            let block_height = bitcoind
+                .get_block_count()
+                .await
+                .map_err(ControlFlow::Continue)?
+                - 1;
+            cmd!(
+                self,
+                "lightning",
+                "wait-for-chain-sync",
+                "--block-height",
+                block_height
+            )
+            .run()
+            .await
+            .map_err(ControlFlow::Continue)?;
+            Ok(())
+        })
+        .await?;
+        Ok(())
     }
 }
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -28,7 +28,7 @@ use crate::cli::{cleanup_on_exit, exec_user_command, setup, write_ready_file, Co
 use crate::envs::{FM_DATA_DIR_ENV, FM_DEVIMINT_RUN_DEPRECATED_TESTS_ENV, FM_PASSWORD_ENV};
 use crate::federation::{self, Client, Federation};
 use crate::util::{poll, poll_with_timeout, LoadTestTool, ProcessManager};
-use crate::version_constants::{VERSION_0_3_0, VERSION_0_3_0_ALPHA};
+use crate::version_constants::{VERSION_0_3_0, VERSION_0_3_0_ALPHA, VERSION_0_4_0_ALPHA};
 use crate::{cmd, dev_fed, poll_eq, DevFed, Gatewayd, LightningNode, Lightningd, Lnd};
 
 pub struct Stats {
@@ -1014,7 +1014,16 @@ pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         })
         .await?
         .bolt11;
-    try_join!(cln.await_block_processing(), lnd.await_block_processing())?;
+    let gateway_cli_version = crate::util::GatewayCli::version_or_default().await;
+    let gatewayd_version = crate::util::Gatewayd::version_or_default().await;
+    if gateway_cli_version >= *VERSION_0_4_0_ALPHA && gatewayd_version >= *VERSION_0_4_0_ALPHA {
+        try_join!(
+            gw_cln.wait_for_chain_sync(&bitcoind),
+            gw_lnd.wait_for_chain_sync(&bitcoind)
+        )?;
+    } else {
+        try_join!(cln.await_block_processing(), lnd.await_block_processing())?;
+    }
     ln_pay(&client, invoice.clone(), lnd_gw_id.clone(), false).await?;
     let fed_id = fed.calculate_federation_id().await;
 
--- a/fedimint-testing/src/ln/mock.rs
+++ b/fedimint-testing/src/ln/mock.rs
@@ -19,8 +19,9 @@ use lightning_invoice::{
     SignedRawBolt11Invoice, DEFAULT_EXPIRY_TIME,
 };
 use ln_gateway::gateway_lnrpc::{
-    self, CreateInvoiceRequest, CreateInvoiceResponse, EmptyResponse, GetNodeInfoResponse,
-    GetRouteHintsResponse, InterceptHtlcResponse, PayInvoiceRequest, PayInvoiceResponse,
+    self, CreateInvoiceRequest, CreateInvoiceResponse, EmptyResponse, GetFundingAddressResponse,
+    GetNodeInfoResponse, GetRouteHintsResponse, InterceptHtlcResponse, PayInvoiceRequest,
+    PayInvoiceResponse,
 };
 use ln_gateway::lightning::cln::{HtlcResult, RouteHtlcStream};
 use ln_gateway::lightning::{ILnRpcClient, LightningRpcError};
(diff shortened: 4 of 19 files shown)


m1sterc001guy:
From an offline discussion with @tvolk131, we thought it would be a good idea to add a very simple liquidity management API to the gateway, essentially mirroring a similar API exposure to LDK-node. 

While I don't think we want the gateway to become a de-factor lightning management tool, it would be nice to expose very simple tasks like connecting to peers and opening channels. This should make the LDK integration easier and also simplify some of the `devimint` logic.

fedimint-testing/src/ln/mock.rs
```diff
@@ -163,4 +163,34 @@ impl ILnRpcClient for FakeLightningTest {
     ) -> Result<EmptyResponse, LightningRpcError> {
         Ok(EmptyResponse {})
     }
+
+    async fn connect_to_peer(
+        &self,
+        _pubkey: String,
+        _host: String,
+    ) -> Result<EmptyResponse, LightningRpcError> {
+        Ok(EmptyResponse {})
```
m1sterc001guy, line 172: `unimplemented!` is probably better for these functions, we shouldn't ever use these in the mocks

fedimint-testing/src/ln/real.rs
```diff
@@ -728,6 +788,38 @@ impl ILnRpcClient for LdkLightningTest {
     ) -> Result<EmptyResponse, LightningRpcError> {
         unimplemented!("Unsupported: we dont currently support HTLC interception for LDK Node");
     }
+
+    async fn connect_to_peer(
```
m1sterc001guy, line 792: Removed in prior PR, needs rebase

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -108,6 +114,26 @@ pub trait ILnRpcClient: Debug + Send + Sync {
         &self,
         htlc: InterceptHtlcResponse,
     ) -> Result<EmptyResponse, LightningRpcError>;
+
+    async fn connect_to_peer(
```
m1sterc001guy, line 118: Add docstring to this function and the others

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -108,6 +114,26 @@ pub trait ILnRpcClient: Debug + Send + Sync {
...
+        &self,
+        pubkey: String,
```
m1sterc001guy, line 120: Can this be a `secp256k1::PublicKey`? Why String?

gateway/ln-gateway/src/rpc/rpc_client.rs
```diff
@@ -109,6 +110,41 @@ impl GatewayRpcClient {
         self.call_post(url, payload).await
     }
 
+    pub async fn connect_to_peer(&self, payload: ConnectToPeerPayload) -> GatewayRpcResult<()> {
+        let url = self
+            .base_url
+            .join("/connect_to_peer")
```
m1sterc001guy, line 116: Nit: we should probably make constants out of `connect_to_peer` similar to how we've done for the federation API. Perhaps could be a follow up

gateway/cli/src/main.rs
```diff
@@ -92,6 +93,40 @@ pub enum Commands {
...
+    OpenChannel {
+        /// The public key of the node to open a channel with
+        #[clap(long)]
+        pubkey: String,
+        /// The amount to fund the channel with
+        #[clap(long)]
+        channel_size_sats: u64,
+        /// The amount to push to the other side of the channel
+        #[clap(long)]
+        push_amount_sats: u64,
```
m1sterc001guy, line 122: Can we make this optional and default to 0?

gateway/cli/src/main.rs
```diff
@@ -92,6 +93,40 @@ pub enum Commands {
         #[clap(long)]
         network: Option<bitcoin::Network>,
     },
+    #[command(subcommand)]
+    Lightning(LightningCommands),
+}
+
+#[derive(Subcommand)]
+pub enum LightningCommands {
```
m1sterc001guy, line 101: Maybe add a comment here that this API should be intentionally kept very thin

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -733,6 +739,143 @@ impl ILnRpcClient for GatewayLndClient {
...
+                sat_per_vbyte: 0,
+                node_pubkey: hex::decode(pubkey).map_err(|e| {
+                    LightningRpcError::FailedToOpenChannel {
+                        failure_reason: format!("Failed to decode pubkey {e:?}"),
+                    }
+                })?,
+                node_pubkey_string: String::new(),
+                local_funding_amount: channel_size_sats.try_into().expect("u64 -> i64"),
+                push_sat: push_amount_sats.try_into().expect("u64 -> i64"),
+                target_conf: 0,
```
m1sterc001guy, line 822: Nit: I think you can use `..Default::default()` syntax if you want to make this a tad less verbose

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -733,6 +739,143 @@ impl ILnRpcClient for GatewayLndClient {
...
+                .map_err(|e| LightningRpcError::FailedToWaitForChainSync {
+                    failure_reason: format!("Failed to get node info {e:?}"),
+                })?
+                .into_inner();
+
+            if response.block_height >= block_height && response.synced_to_chain {
+                return Ok(EmptyResponse {});
+            }
+
+            sleep(Duration::from_secs(5)).await;
```
m1sterc001guy, line 872: Maybe make this a constant too

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -495,6 +499,139 @@ impl GatewayLightning for ClnRpcService {
...
+                    }) => Ok(blockheight),
+                    _ => Err(ClnExtensionError::RpcWrongResponse),
+                })
+                .map_err(|e| {
+                    error!("cln getinfo rpc returned error {:?}", e);
+                    tonic::Status::internal(e.to_string())
+                })?
+                .map_err(|e| tonic::Status::internal(e.to_string()))?;
+
+            if blockheight >= target_block_height {
```
m1sterc001guy, line 626: Do you need to check if the chain is synced here like you did for LND?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -495,6 +499,139 @@ impl GatewayLightning for ClnRpcService {
...
+                return Ok(tonic::Response::new(EmptyResponse {}));
+            }
+
+            sleep(Duration::from_secs(5)).await;
```
m1sterc001guy, line 630: Maybe the total deadline for waiting for the chain tip should be passed in by the client. Here it is 5 seconds * 10 retries, which is only 50 seconds. I'm not sure yet where we plan on calling this but that seems low to me.

m1sterc001guy:
Overall the code looks good but I think we will need tests before merging. Maybe we could replace the devimint logic for opening channels with this? It would allow us to simplify the devimint code.

Also, I'm not sure if the API that is being added is enough. `devimint` has "wait for channel update" inside of `open_channel`, which waits until the channel gossip has updated in LND after CLN opens the channel. I don't think `wait_for_chain_sync` will do the same thing.

fedimint-testing/src/ln/mock.rs
tvolk131: Done.

fedimint-testing/src/ln/real.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning/mod.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning/mod.rs
tvolk131: Good point! Done.

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -93,3 +93,21 @@ pub struct SetConfigurationPayload {
     pub routing_fees: Option<String>,
     pub network: Option<Network>,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ConnectToPeerPayload {
+    pub pubkey: String,
```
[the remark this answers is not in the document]
tvolk131: Done.

gateway/cli/src/main.rs
tvolk131: Done. I only did this for the CLI argument (i.e. we're performing `.unwrap_or(0)` here in the CLI code). I did this since I think a plain `u64` is more appropriate for `ILnRpcClient` since `None` represents the same behavior as `Some(0)`. I'm assuming this was your intention too but just wanted to lay out the reasoning here.

gateway/ln-gateway/src/lightning/lnd.rs
tvolk131: Oh snap, I wasn't familiar with this syntax. Much cleaner!

gateway/ln-gateway/src/rpc/rpc_client.rs
tvolk131: Conforming to existing pattern for now. Created #4545 to track this.

gateway/cli/src/main.rs
tvolk131: Done.

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: I didn't see a way to do this in CLN but after looking closer, it appears you're right! We're now also waiting for `warning_bitcoind_sync` to be `None`.

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: Done. I kept the defaults the same for the CLI though - lmk if you think we should change them.

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -629,6 +633,132 @@ impl ILnRpcClient for GatewayLndClient {
...
+        &self,
+        pubkey: secp256k1::PublicKey,
+        host: String,
+    ) -> Result<EmptyResponse, LightningRpcError> {
+        let mut client = Self::connect(
+            self.address.clone(),
+            self.tls_cert.clone(),
+            self.macaroon.clone(),
+        )
+        .await?;
```
elsirion, line 647: This is repeated a lot, maybe worth to create a helper fn?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -496,6 +499,153 @@ impl GatewayLightning for ClnRpcService {
...
+            if let Some(retry_delay) = retry_delay.clone() {
+                sleep(retry_delay.try_into().map_err(|e: DurationError| {
+                    Status::internal(format!(
+                        "Failed to convert prost Duration to std Duration: {}",
+                        e
+                    ))
+                })?)
+                .await;
+            }
+        }
```
elsirion, line 645: We do have the [`retry`](https://docs.fedimint.org/fedimint_core/util/fn.retry.html) utility, maybe it can help here too?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -148,6 +149,10 @@ fn v1_routes(gateway: Gateway) -> Router {
         .route("/leave-fed", post(leave_fed))
         .route("/backup", post(backup))
         .route("/restore", post(restore))
+        .route("/connect_to_peer", post(connect_to_peer))
+        .route("/get_funding_address", get(get_funding_address))
```
elsirion, line 153: `get_funding_address` should be a `POST` request too since it causes mutation of the service's internal state and should not be cached. One could argue that it always returns *some* address belonging to the node, so is kinda idempotent and can be `GET`, but `POST` feels better.

elsirion:
Mostly looking good, but needs a rebase anyway, so maybe some improvements can also make it in.

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -305,6 +310,41 @@ async fn set_configuration(
...
+#[instrument(skip_all, err, fields(?payload))]
+async fn open_channel(
+    Extension(gateway): Extension<Gateway>,
+    Json(payload): Json<OpenChannelPayload>,
+) -> Result<impl IntoResponse, GatewayError> {
+    gateway.handle_open_channel_msg(payload).await?;
+    Ok(Json(json!(())))
+}
+
+#[instrument(skip_all, err)]
```
m1sterc001guy, line 339: nit: `#[instrument(skip_all, err, fields(?payload))]`

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -93,3 +94,23 @@ pub struct SetConfigurationPayload {
...
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ConnectToPeerPayload {
+    pub pubkey: secp256k1::PublicKey,
+    pub host: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct OpenChannelPayload {
+    pub pubkey: String,
```
m1sterc001guy, line 106: Can this be a `secp256k1::PublicKey`?

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -111,6 +118,35 @@ pub trait ILnRpcClient: Debug + Send + Sync {
...
     ) -> Result<EmptyResponse, LightningRpcError>;
+
+    /// Connect to a peer lightning node from the gateway's lightning node.
+    async fn connect_to_peer(
+        &self,
+        pubkey: secp256k1::PublicKey,
+        host: String,
+    ) -> Result<EmptyResponse, LightningRpcError>;
+
+    /// Get the funding address belonging to the gateway's lightning node
```
m1sterc001guy, line 129: nit: "Get the funding address" -> "Get a funding address"

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -111,6 +118,35 @@ pub trait ILnRpcClient: Debug + Send + Sync {
...
+    /// wallet.
+    async fn get_funding_address(&self) -> Result<GetFundingAddressResponse, LightningRpcError>;
+
+    /// Open a channel with a peer lightning node from the gateway's lightning
+    /// node.
+    async fn open_channel(
+        &self,
+        pubkey: String,
```
m1sterc001guy, line 137: Can this be a `secp256k1::PublicKey` too?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1571,6 +1620,8 @@ pub enum GatewayError {
     InsufficientFunds,
     #[error("Federation already connected")]
     FederationAlreadyConnected,
+    #[error("Error parsing response: {}", OptStacktrace(.0))]
+    ResponseParseError(anyhow::Error),
```
m1sterc001guy, line 1624: nit: `LightningResponseParseError`? My first thought was "response from what"?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1123,6 +1123,55 @@ impl Gateway {
...
+        WaitForChainSyncPayload {
+            block_height,
+            max_retries,
+            retry_delay,
+        }: WaitForChainSyncPayload,
+    ) -> Result<()> {
+        let context = self.get_lightning_context().await?;
+        context
+            .lnrpc
+            .wait_for_chain_sync(block_height, max_retries, retry_delay)
```
m1sterc001guy, line 1170: Polling for block height can be implemented here instead of polling in each impl

gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -176,3 +190,37 @@ message GetRouteHintsResponse {
...
+  // The amount of sats that should be pushed to the
+  // counterparty once the channel is opened.
+  uint64 push_amount_sats = 3;
+}
+
+message WaitForChainSyncRequest {
+  // The height that the lightning node should have fully processed before returning.
+  uint32 block_height = 1;
+
+  uint32 max_retries = 2;
```
m1sterc001guy, line 223: Can remove this if we refactor to just get the block height

devimint/src/tests.rs
```diff
@@ -805,7 +806,18 @@ pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
         })
         .await?
         .bolt11;
-    tokio::try_join!(cln.await_block_processing(), lnd.await_block_processing())?;
+    let gateway_cli_version = crate::util::GatewayCli::version_or_default().await;
+    let gatewayd_version = crate::util::Gatewayd::version_or_default().await;
+    if gateway_cli_version >= Version::parse("0.4.0-alpha")?
```
m1sterc001guy, line 811: Let's make 0.4.0-alpha a constant

devimint/src/gatewayd.rs
```diff
@@ -112,4 +121,71 @@ impl Gatewayd {
...
+    }
+
+    pub async fn get_funding_address(&self) -> Result<String> {
+        let address = cmd!(self, "lightning", "get-funding-address")
+            .out_string()
+            .await?;
+        Ok(address)
+    }
+
+    pub async fn open_channel(
```
m1sterc001guy, line 147: nit: are `connect_to_peer` and `open_channel` in devimint both necessary? Maybe `open_channel` could always call `connect_to_peer` first

devimint/src/external.rs
```diff
@@ -661,6 +668,131 @@ pub async fn open_channel(
     Ok(())
 }
 
+// TODO(tvolk131): Remove the `cln` and `lnd` arguments and instead use `gw_cln`
```
m1sterc001guy, line 671: Let's include versions in these TODOs so we know when its safe to remove them

devimint/src/external.rs
```diff
@@ -661,6 +668,131 @@ pub async fn open_channel(
...
+// and `gw_lnd`
+pub async fn open_channel_between_gateways(
+    process_mgr: &ProcessManager,
+    bitcoind: &Bitcoind,
+    cln: &Lightningd,
+    gw_cln: &Gatewayd,
+    lnd: &Lnd,
+    gw_lnd: &Gatewayd,
```
m1sterc001guy, line 679: Not a blocker right now but we'll need to think about how to restructure this for LDK

devimint/src/external.rs
```diff
@@ -661,6 +668,131 @@ pub async fn open_channel(
...
+        Err(ControlFlow::Continue(anyhow!("channel not found")))
+    })
+    .await?;
+
+    lnd.client
+        .lock()
+        .await
+        .lightning()
+        .update_channel_policy(PolicyUpdateRequest {
+            min_htlc_msat: 1,
```
m1sterc001guy, line 782: This is necessary for allowing < 1 sat payments right?

gateway/cli/src/main.rs
```diff
@@ -96,6 +102,51 @@ pub enum Commands {
...
+        #[clap(long)]
+        host: String,
+    },
+    /// Get a Bitcoin address to fund the gateway
+    GetFundingAddress,
+    /// Open a channel with another lightning node
+    OpenChannel {
+        /// The public key of the node to open a channel with
+        #[clap(long)]
+        pubkey: String,
```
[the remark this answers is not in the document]
tvolk131: Nope, that was simply an oversight on my part. Fixed!

gateway/ln-gateway/src/rpc/rpc_server.rs
tvolk131: Done.

gateway/ln-gateway/src/rpc/mod.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning/mod.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning/mod.rs
tvolk131: Done.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

devimint/src/external.rs
tvolk131: Yep! My goal is to remove the `cln` and `lnd` arguments, then rename `gw_cln` and `gw_lnd` to something more generic, like `gw1` and `gw2`. Then an LDK gateway will just be another instance of `Gatewayd`.

devimint/src/external.rs
tvolk131: Yep! Linking #2827

devimint/src/tests.rs
tvolk131: Done.

gateway/ln-gateway/src/rpc/rpc_server.rs
tvolk131: Good point! Done.

gateway/ln-gateway/src/lib.rs
tvolk131: Makes sense. Before I do that, what are your thoughts about removing `wait_for_chain_sync()`, adding a `synced_block_height` field to `GatewayInfo`, then having the `wait-for-chain-sync` CLI command simply poll the gateway CLI's `info` command until the value of the new `synced_block_height` field is >= the target block height?

devimint/src/gatewayd.rs
tvolk131: Good point! It looks like ldk-node's [`connect_open_channel`](https://docs.rs/ldk-node/0.2.1/ldk_node/struct.Node.html#method.connect_open_channel) method would support this and wouldn't run into issues if the peer is already connected (see [here](https://docs.rs/ldk-node/0.2.1/src/ldk_node/lib.rs.html#981)). However, it looks like this is not the case for LND and CLN. In order for this to work well, we'd need to be able to differentiate already-connected errors from internal or network errors. Thoughts?

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -629,6 +633,132 @@ impl ILnRpcClient for GatewayLndClient {
...
+                return Ok(EmptyResponse {});
+            }
+
+            sleep(retry_delay).await;
+        }
+
+        Err(LightningRpcError::FailedToWaitForChainSync {
+            failure_reason: "Failed to wait for chain sync".to_string(),
+        })
+    }
```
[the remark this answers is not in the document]
tvolk131: Do we want an endpoint specifically for fetching the block count, or would it make more sense to add a block count field to `GatewayInfo` and then poll `gateway-cli info`?

https://github.com/fedimint/fedimint/blob/a3b9990ebcc5690f67b6a8e8895d8a9024bab77c/gateway/ln-gateway/src/rpc/mod.rs#L72-L85

gateway/cli/src/main.rs
```diff
@@ -96,6 +102,51 @@ pub enum Commands {
...
+        #[clap(long)]
+        block_height: u32,
+
+        /// The maximum number of retries
+        #[clap(long)]
+        max_retries: Option<u32>,
+
+        /// The delay between retries
+        #[clap(long)]
+        retry_delay_seconds: Option<u64>,
```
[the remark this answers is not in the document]
tvolk131: Leaving this for now while we figure out if we're even going to keep `WaitForChainSync`

gateway/ln-gateway/src/lightning/lnd.rs
tvolk131: Done. Created #4924 and rebased this PR on top of it.

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: Leaving this for now while we figure out if we're even going to keep `WaitForChainSync`

gateway/ln-gateway/src/lightning/lnd.rs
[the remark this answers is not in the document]
elsirion: Maybe I misunderstand the use case for the wait block height API? What is it used for? Only tests? Even if so, I think having it as part of `GatewayInfo` would be more generally useful (although I think there's a different trait for extra test fixture APIs?)

elsirion (approved):

elsirion:
Something broke backwards compatibility.

gateway/cli/src/main.rs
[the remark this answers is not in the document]
tvolk131: Update: As discussed elsewhere, the under-the-hood behavior has been refactored so that rather than having implementation-specific wait-for-chain-sync functions that each have their own polling logic, the CLI command polls `gateway info` and checks the returned block count until it matches or exceeds the requested height.

elsirion (dismissed):

elsirion:
This PR was in rebase hell for quite a while. Looked good enough, but please take another look @m1sterc001guy :)

tvolk131:
Pushed one more commit that makes the same change as #4991 but in the new channel open codepath. This should fix the flakiness we ran into when attempting to merge earlier. Can we toss this in the merge queue again?

elsirion (approved):

dpc:
Looks like legitimate wasm problem somewhere.

tvolk131:
> Looks like legitimate wasm problem somewhere.

Another rebase seemed to fix it ¯\_(ツ)_/¯.

bradleystachurski (approved):
CI is more stable so let's give this another try in the merge queue

tvolk131:
Oooh looks like this merge attempt was so close! Can we try again?

Issue #4519, opened 2024-03-10
RFC: Getting rid of `WsAdminClient`


dpc:
I want to have a client module that can call module endpoint that is admin-auth protected.

I'm looking at:

```rust
pub struct WsAdminClient {
    inner: DynGlobalApi,
    pub url: SafeUrl,
}
```

and looks like this `url` is only used during construction:

```rust
impl WsAdminClient {
    pub fn new(url: SafeUrl) -> Self {
        // The peer ids given to the federation API are only useful when connected to
        // multiple peers so errors can be attributed. The admin client has no use for
        // them.
        Self {
            inner: DynGlobalApi::from_endpoints(vec![(PeerId(0), url.clone())]),
            url,
        }
    }
}
```

to make a `DynGlobalApi` that looks like a single-peer federation. Otherwise it's unused so the `WsAdminClient` could be just ...`DynGlobalApi`.

Could we just add to the `Client` and `ClientBuilder` add ability to set "admin auth" (basically peer_id + pass), and that would make all the `DynGlobalApi` and `DynModuleApi` become single-peer/endpoint under the hood *for everything* and save the `auth: Option<Auth>` as a field?

The benefit is that everything perfectly unifies and simplifies. There is no need for extra types, extra wrappers, etc. Auth and non-auth endpoints can use the same `DynGlobalApi`, and we don't need to special case anything auth.

BTW. This made me wonder: if you are a guardian, why would you ever call any other guardian? Wouldn't you want to everything through your own instance?

TODO:

* [x] refactor the new methods https://github.com/fedimint/fedimint/pull/4520#discussion_r1525393351
* [x] possibly should have `all_peers` and `guardian_self_peer_id -> Option<PeerId>`

justinmoon:
During the setup ceremony, guardians need to connect to each other. For example, the "leader" sets the configuration parameters and "followers" need to download these parameters from the leader and approve or deny them. I believe all of this uses `WsAdminClient` (could be wrong).

justinmoon:
Did https://github.com/fedimint/fedimint/pull/4520 close this?

dpc:
> Did #4520 close this?

Still some follow-up TODOs. See first post.

PR #4520, opened 2024-03-10
chore: get rid of WsAdminClient

dpc:
Re #4519

I wrote it just to see if it's possible and everything still works.

Now I think we might tweak `IRawFederationApi` a bit and retain ability to do both. I.e. split `IRawFederationApi::all_peers` into `all_peers` + `admin_peer`, so that  `FederationApiExt::request_admin` can call only the admin, while other requests  - all guardians.

But one way or another I think getting rid of `WsAdminClient` is easy and the right thing to do.


fedimint-core/src/api.rs
```diff
@@ -421,6 +468,16 @@ impl AsRef<dyn IGlobalFederationApi + 'static> for DynGlobalApi {
...
+    pub fn from_pre_peer_id_endpoint(url: SafeUrl) -> Self {
+        // PeerIds are used only for informational purposes, but just in case, make a
+        // big number so it stands out
+        GlobalFederationApiWithCache::new(WsFederationApi::new(vec![(PeerId::from(1024), url)]))
+            .into()
+    }
+
+    pub fn from_single_endpoint(peer: PeerId, url: SafeUrl) -> Self {
+        GlobalFederationApiWithCache::new(WsFederationApi::new(vec![(peer, url)])).into()
+    }
```
elsirion, line 480: Is it really worth having two of these functions?

fedimint-core/src/api.rs
```diff
@@ -500,6 +557,98 @@ pub trait IGlobalFederationApi: IRawFederationApi {
         timeout: Duration,
         num_responses_required: Option<usize>,
     ) -> FederationResult<ApiVersionSet>;
+
+    /// Sets the password used to decrypt the configs and authenticate
+    ///
+    /// Must be called first before any other calls to the API
+    async fn set_password(&self, auth: ApiAuth) -> FederationResult<()>;
```
elsirion, line 564: > Must be called first before any other calls to the API

This doesn't make a lot of sense if read in the context of the normal client API. Maybe the admin API should be defined as an extension trait just like module APIs are? I think that would make it easier to understand the separate concepts of "normal API" and "admin API"

elsirion (approved):
I think this makes sense, having two clients annoyed me too.

As a next step towards the goal of a "guardian client" I could imagine being able to save `our_id` and the password to the client DB, which would lead to guardian functionality becoming available. To allow modules to access guardian APIs we should add something like `ClientContext::guardian_auth(&self) -> Option<(PeerId, ApiAuth)>` that can be used to determine if the user is a guardian and if so what the authentication token is.

elsirion:
Needs rebase :(

fedimint-core/src/api.rs
dpc: Doesn't matter too much, just wanted a simplest patch possible. We can clean up with time.

fedimint-core/src/api.rs
dpc: I just moved it.

dpc:
Rebased.

elsirion (approved):

fedimint-core/src/api.rs
```diff
@@ -396,6 +406,43 @@ pub trait FederationApiExt: IRawFederationApi {
...
+    where
+        Ret: serde::de::DeserializeOwned + Eq + Debug + Clone + MaybeSend,
+    {
+        // We would never want to accidentally send our password to everyone
+        assert_eq!(
+            self.all_peers().len(),
+            1,
+            "attempted to broadcast admin password?!"
+        );
+        self.request_current_consensus(method.into(), params.with_auth(auth))
```
m1sterc001guy, line 425: Could we use `request_single_peer` here instead of `request_current_consensus`?

fedimint-core/src/api.rs
```diff
@@ -396,6 +406,43 @@ pub trait FederationApiExt: IRawFederationApi {
...
+        Ret: serde::de::DeserializeOwned + Eq + Debug + Clone + MaybeSend,
+    {
+        // There is no auth involved, but still - it should only ever be called on a
+        // single endpoint
+        assert_eq!(
+            self.all_peers().len(),
+            1,
+            "attempted to broadcast an admin request?!"
+        );
+        self.request_current_consensus(method.into(), params).await
```
m1sterc001guy, line 444: Same comment as above

fedimint-core/src/api.rs
dpc: It makes sense, but it changes the type of the result, which will lead further changes, and I have other PRs on top of this one, and just wanted to move these methods 1:1 to a new place to keep it small. We'll follow up with improvements. Added a TODO to #4519 4519

m1sterc001guy: Works for me

m1sterc001guy (approved):

okjodom:
> As a next step towards the goal of a "guardian client"

Could you elaborate more on this goal. What functionalities would this guardian client have, differentiated from the current Federation/User client, and from the direct API calls to a Guardian app makes to ConfigGen / Consensus APIs

dpc:
> Could you elaborate more on this goal. What functionalities would this guardian client have, differentiated from the current Federation/User client, and from the direct API calls to a Guardian app makes to ConfigGen / Consensus APIs

I'm not sure about the original meaning but, the AFAICT next goals are:

* Turn [`IRawFederationApi::all_peers`](https://github.com/fedimint/fedimint/blob/ad21594a614446a994768f1431138966e7abfad4/fedimint-core/src/api.rs#L245) into two calls: `all_peers` and `our_id() -> Option<PeerdID>`
* Make unauth calls use one, the admin calls the others
* Make the client initializing implementations of `IRawFederationApi` with correct settings so they can return them from the DB
* Maybe make the `Client` remember peer_id + auth in the database.

PR #4522, opened 2024-03-10
chore: make `fedimnt-cli module <module>` be positional

dpc:
I know that I'm usually advocating for long/named arguments but:

```
fedimint-cli module --module meta get-consensus
```

is just silly. Instead if we:

```
fedimint-cli module meta get-consensu
```

along with some other improvements I'm working on will make module-specific first class citizens of fedimint-cli.


bradleystachurski (commented):
We'll need to conditionally handle for backwards compatibility.

```
38665   │ 2024-03-10T20:02:24.4435443Z 00:00:33 error: unexpected argument '--module' found
38666   │ 2024-03-10T20:02:24.4435510Z 00:00:33
38667   │ 2024-03-10T20:02:24.4435674Z 00:00:33   tip: to pass '--module' as a value, use '-- --module'
```

bradleystachurski (approved):

m1sterc001guy (approved):

PR #4524, opened 2024-03-10
 chore: make the test-ci-all work harder 

dpc:


bradleystachurski (dismissed):
Verified `just test-ci-all` works locally, but I'm seeing errors on back-compat.

```
22867   │ 00:00:25 Error: polling lightningd after 10 retries
22868   │ 00:00:25
22869   │ 00:00:25 Caused by:
22870   │ 00:00:25     0: connect to lightningd
22871   │ 00:00:25     1: No such file or directory (os error 2)
22872   │ ## FAILED: devimint_cli_test (FM: current, CLI: current, GW: v0.2.1)
```

I'll continue tweaking params on my end. Approving since I think it's safe to merge and followup.

@m1sterc001guy If you have a chance I'm curious if this resolves the issues you had locally.

dpc:
> Verified `just test-ci-all` works locally, but I'm seeing errors on back-compat.

For this one we should just increase the time. 10s is not a lot. 30s seems more reasonable.

bradleystachurski:
I bumped to 30 and now seeing networking errors with the bitcoind client

```
00:00:21 2024-03-11T23:32:23.403138Z  WARN devimint: Failed to create wallet ... retrying JSON-RPC error: transport error: Couldn't connect to host: Connection refused (os error 111)
00:00:24 2024-03-11T23:32:26.068629Z  INFO devimint::external: Beginning to mine 101 blocks to address bcrt1qj06wd2nptduccl jvp52k84ynlkv9s2ehmdn30a
00:01:09 2024-03-11T23:33:11.942712Z  INFO devimint: sending SIGKILL to bitcoind and waiting for it to exit
00:01:09 Error: Failed to generate blocks
00:01:09
00:01:09 Caused by:
00:01:09     JSON-RPC error: transport error: Didn't receive response data in time, timed out.
## FAILED: devimint_cli_test (FM: current, CLI: v0.2.1, GW: current)
```

dpc:
@bradleystachurski All, right. I've lowered the default load again, and added ability to customize it locally:

```
> cat .shrc.local
export FM_TEST_CI_ALL_MAX_LOAD=40000
export FM_TEST_CI_ALL_DELAY=.5
```

drop a file in there like that, tune to your liking, and it should do the job.

bradleystachurski (approved):
`.shrc.local` is a solid solution. Making tweaks locally allowed `just test-compatibility` to pass again 👍

justinmoon (approved):
Didn’t test

devimint/src/tests.rs
```diff
@@ -288,7 +288,7 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
             let fm_pay_stats = stats_for(fm_internal_pay);
 
             println!("### LATENCY FM PAY: {fm_pay_stats}");
-            assert!(fm_pay_stats.median < Duration::from_secs(12));
+            assert!(fm_pay_stats.median < Duration::from_secs(15));
```
elsirion, line 291: Maybe latency tests should not run in parallel on the same machine? This hack may hide latency introduced from round trips that don't require much compute?

bradleystachurski: Good callout 👍 

Made an issue to make sure we keep track https://github.com/fedimint/fedimint/issues/4579.

PR #4526, opened 2024-03-11
chore(ci): cleanup output and print times in test-ci-all

dpc:


bradleystachurski (commented):
I think the ci test env will also need `time` now that it's called with `command`

```
 > 00:00:00 /build/source/scripts/dev/run-test/fm-run-test: line 32: time: command not found
```
https://github.com/fedimint/fedimint/actions/runs/8226271625/job/22492472456?pr=4526#step:9:2042

scripts/dev/run-test/fm-run-test
```diff
@@ -9,14 +11,25 @@ if [ -z "$test_name" ]; then
 fi
 
 version_str="${FM_RUN_TEST_VERSIONS:+ ($FM_RUN_TEST_VERSIONS)}"
+test_out_path="$(mktemp --tmpdir fm-XXXXX)"
+time_out_path="$(mktemp --tmpdir fm-XXXXX)"
+time_fmt='%e\t%U\t%S'
```
bradleystachurski, line 16: [nit] In the logs it wasn't obvious what each time column represented. Perhaps a short description would help?

```
## RUN:  backend_test_bitcoind
## TIME:   243.73  138.65   17.87
## DONE: backend_test_bitcoind
```

Turns into

```
## RUN:  backend_test_bitcoind
## TIME:   243.73 (elapsed)  138.65 (user)  17.87 (system)
## DONE: backend_test_bitcoind
```

Also on that note, do we care about the user and system times? Would elapsed be enough?

bradleystachurski (commented):
```
1931: All tests successful
```

All of back-compat logs in 1931 lines 😍 

This is excellent, thanks for tackling!

For other reviewers, prior to this change we would have to download the log file from ci, unzip, then look through logs locally since the output was too large for github's ui.

scripts/dev/run-test/fm-run-test
dpc: I changed it to:

<time>s <memory>B <waits>W <ctx-switches>c


I don't see a good way to document it in-line (s and B will self explanatory, but W and c are just formats from https://www.man7.org/linux/man-pages/man1/time.1.html ), but I also think not many people really need to understand them.

bradleystachurski: Ooh neat, I like it. Small followup nit, `%M` is in Kbytes on the man page, so the `B` unit is a _bit_ confusing.

bradleystachurski (approved):
Verified failures log as expected locally 👍

m1sterc001guy (approved):
LGTM

PR #4542, opened 2024-03-12
feat: API for listing incoming unspent contract hashes

m1sterc001guy:
In order to support claiming ecash that is committed to a contract via an external pubkey, we need some additional functionality:

1. Federation API to list current incoming unspent contracts.
2. Logic on the client side to iterate through all hashes, apply any necessary tweaks to a private key, and check if the hash matches, then spawn a state machine to claim the ecash.

This PR implements #1. It adds a very simple test to verify that the API is working, this test can be iterated on and expanded as we add the claim functionality.

@Kodylow @benthecarman FYI

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -1457,6 +1457,10 @@ impl LightningClientModule {
             }
         }))
     }
+
+    pub async fn get_unspent_incoming_contract_hashes(&self) -> anyhow::Result<Vec<sha256::Hash>> {
+        Ok(self.module_api.get_unspent_incoming_contracts().await?)
+    }
 }
 
 #[allow(clippy::large_enum_variant)]
--- a/fedimint-core/src/endpoint_constants.rs
+++ b/fedimint-core/src/endpoint_constants.rs
@@ -1,4 +1,5 @@
 pub const ACCOUNT_ENDPOINT: &str = "account";
+pub const LIST_UNSPENT_INCOMING_CONTRACTS_ENDPOINT: &str = "list_unspent_incoming_contracts";
 pub const ADD_CONFIG_GEN_PEER_ENDPOINT: &str = "add_config_gen_peer";
 pub const AUDIT_ENDPOINT: &str = "audit";
 pub const GUARDIAN_CONFIG_BACKUP_ENDPOINT: &str = "download_guardian_backup";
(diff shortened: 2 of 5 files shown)


benthecarman:
Is it possible to attach a timestamp to each hash? That way you don't need to re-attempt each old hash when scanning. Can only check the new ones since your last time.

m1sterc001guy:
> Is it possible to attach a timestamp to each hash? That way you don't need to re-attempt each old hash when scanning. Can only check the new ones since your last time.

`IncomingContract` unfortunately doesn't have a timestamp or anything, so it would require a database migration (and I think the timestamp would need to be agreed on in consensus), probably not worth it for this. Hopefully the number of in-progress receives is somewhat low, so performance won't be an issue. We could also implement similar logic on the client side, although that seems messier.

Maybe for Lightning NG we can add a timestamp @joschisan

joschisan:
@m1sterc001guy what you describe here is what I had in mind for lightning ng, and coincidentally I noticed the advantage of som kind of ordering yesterday as well. I just pushed a commit that allows to create incoming contracts for a foreign public key by tweaking the public key with a ephemeral secret key. This also allows for stateless recovery since the method that finds incoming contracts for ourself that were created by someone else automatically also finds incoming contracts for ourself that we created ourself.

m1sterc001guy:
Another thought: rather than changing the value back to 0, can we simply remove the `IncomingContract` from the database after it has been claimed? That would help with the performance on the server side of reading all contracts. I'm not sure if there's some auditing code or if there's another reason for keeping them around.

dpc:
> Another thought: rather than changing the value back to 0, can we simply remove the `IncomingContract` from the database after it has been claimed?

Probably must have, as these will add up quickly. If we can't delete, we could have two keys - one for all, one for quick look of active ones.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1457,6 +1457,10 @@ impl LightningClientModule {
             }
         }))
     }
+
+    pub async fn get_unspent_incoming_contract_hashes(&self) -> anyhow::Result<Vec<sha256::Hash>> {
```
dpc, line 1461: I'm unsure do we need to increment Api version requirement for client module now, since we are adding a method that server needs to have to support this version of a client.

On the server side we might have already increased the api version advertised for some recovery-related stuff that was unreleased, but we didn't need to do it for the client, as it was able to auto-detect that's the case and hadle it gracefully.

I guess if you dont not want to require newer server, you'd need to make this `get_unspent_incoming_contract_hashes_opt(..) -> Result<Option<...>>` and return `Ok(None)` if server doesn't support it?

See https://github.com/fedimint/fedimint/blob/02abf76eb111682b5e8d38ad7288a716934fe76a/fedimint-client/src/module/init/recovery.rs#L268

m1sterc001guy:
I'm going to close this PR in favor of https://github.com/fedimint/fedimint/pull/3820

I think it makes more sense to review the whole feature rather than just the new API.

PR #4544, opened 2024-03-12
chore(prometheus): prefix with fm_ and add lots of metrics

dpc:


modules/fedimint-ln-server/src/lib.rs
```diff
@@ -256,10 +251,6 @@ impl ServerModuleInit for LightningInit {
     }
 
     async fn init(&self, args: &ServerModuleInitArgs<Self>) -> anyhow::Result<DynServerModule> {
-        // Ensure all metrics are initialized
```
douglaz, line 259: Removing this will only show the metric if the value is greater than zero. I think it should show even if it's zero.

dpc: AFAIK Prometheus is fine with that, and for most purposes it just doesn't matter. Prometheus (or it's formulas) will treat missing values as zero, and will handle it just like it handles e.g. counter resets.

If it was necessary, it would be happening automatically and user libraries wouldn't require users to do second double accounting of all values.

Pretty much only very rarely updated gauges need manually initializing on start.

douglaz: > If it was necessary, it would be happening automatically and user libraries wouldn't require users to do second double accounting of all values.

It does happen automatically. The problem is that we are using lazy val here.

dpc: Good point, but it still works fine, and such lazy_static usage is mentioned all over the examples, etc. and no one worries.

Did you experience any e2e problems with it in queries, etc.?

douglaz: The problem is the discoverability of the metrics. If you are going to remove the initialization of metrics then we will need a documentation with all the metrics so the user can create alerts and dashboards using them.

If discoverability isn't important then I wonder why you are adding a prefix to all metrics :thinking:

dpc: These metrics will show up right away, typically after a couple of seconds the first time they are touched. As soon as prometheus sees them it will rememeber them. In a long running software like fedimintd it's completely not an issue. @douglaz

douglaz: This isn't true for something like LN_OUTPUT_OUTCOME_CANCEL_OUTGOING_CONTRACT. It may take weeks or months to happen, but when it happens you want to know right away because it may indicate an issue on the gateway.

dpc: @douglaz Then I rather eagerly initialize these handful ones that are like that than maintain the whole list of everything.

Gauges that change very rarely (.e.g blockchain block height, session count, etc.) need initializing like that with a proper value on start as well.

Any other one you think is like that?

dpc: > Any other existing ones you think are like that?

I guess all wallet ones are like that... :D

douglaz (approved):

douglaz:
It seems Cargo.lock needs update

dpc:
@douglaz done.

elsirion (approved):

elsirion:
I think some of the metrics around tx values that are more business intelligence related should probably be replaced with other analytics tools, but there's a lot of interesting ops information being exposed here that should allow us to detect problems! :tada:

maan2003 (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-4544-to-releases/v0.3 origin/releases/v0.3
cd .worktree/backport-4544-to-releases/v0.3
git switch --create backport-4544-to-releases/v0.3
git cherry-pick -x ee3734f16ee6d9084b948f0cffab37122ba19ff7 a1bce3dfa83d82583d8660b4fec8c3464be977ca fa7a2960e39c9e10bfda485db958eb92003acc31 6f974bf5c906788c04e700d2a637b9ef8cf094c4 cc107a8b996b2a0617a9c8f5b667dd9ae1bf007c ce7038f5ca464b249ef53a9bdeaa8e5a88b48455 b370d309f031f710438413aac07f1366a75bf289 c7fa6bd04a719dd72a5c0bce020f98f07ace8243 bfa148ab316c8b916461784f2a973e00ef80788a
```

PR #4546, opened 2024-03-13
chore: add defaults to `just test-compatibility`

dpc:


bradleystachurski (approved):

m1sterc001guy (approved):

justfile.fedimint.just
```diff
@@ -22,10 +22,10 @@ test-ci-all:
 test-count:
   ./scripts/tests/test-cov.sh
 
-test-compatibility *VERSIONS:
+test-compatibility *VERSIONS="v0.2.1":
```
elsirion, line 25: Is the idea here that this is our MSFMV for now?

dpc: I think so, but pinging @bradleystachurski .

bradleystachurski: The motivation for adding the default was to make it less confusing for folks when calling the just command. There was confusion reported in [Discord](https://discord.com/channels/990354215060795454/1217141774812713008/1217253961631924274).

Does MSFMV == Minimum Supported Fedimint Version?

dpc: > Does MSFMV == Minimum Supported Fedimint Version?

Certainly. :)

PR #4549, opened 2024-03-13
chore: update jsonrpsee to 0.22.1

dpc:
I need it for something.

### DIFF
--- a/fedimint-server/src/lib.rs
+++ b/fedimint-server/src/lib.rs
@@ -23,8 +23,7 @@ use fedimint_core::task::TaskGroup;
 use fedimint_core::PeerId;
 use fedimint_logging::{LOG_CONSENSUS, LOG_CORE, LOG_NET_API};
 use futures::FutureExt;
-use jsonrpsee::server::{ServerBuilder, ServerHandle};
-use jsonrpsee::types::error::CallError;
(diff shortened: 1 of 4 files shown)


justinmoon (approved):

fedimint-server/src/lib.rs
```diff
@@ -265,20 +263,15 @@ impl FedimintServer {
...
-                            None::<()>,
-                        )))
+                        ErrorObject::owned(500, "API handler panicked", None::<()>)
                     })?
                     .map_err(|tokio::time::error::Elapsed { .. }| {
-                        jsonrpsee::core::Error::RequestTimeout
+                        // TODO: find a better error for this, the error we used before:
+                        // jsonrpsee::core::Error::RequestTimeout
+                        // was moved to be client-side only
+                        ErrorObject::owned(-32000, "Request timeout", None::<()>)
```
maan2003, line 272: so this is the -32000 we were hitting before

maan2003 (approved):

fedimint-server/src/lib.rs
dpc: I actually don't know what's the right code for request timeout, but jsonrpc error codes are not http codes and are like this.

maan2003: yea we should define our own error codes, -32000 is "unknown error code"

Issue #3234, opened 2023-09-18
Port "legacy" integration tests


justinmoon:
https://github.com/fedimint/fedimint/pull/3213 removed a bunch of integration tests. On dev call we said that the most important ones had already been ported, and that we should port remaining useful tests "sometime". This issue is to make sure we don't forget!

---
Investigation 2024-03-06

Referencing https://github.com/fedimint/fedimint/pull/3216/files (`integrationtests/tests/tests.rs`) vs https://github.com/fedimint/fedimint/tree/7a36cafa295ce30528ac1917575cb82ce8965907 (latest master)

- Missing
  - ~ecash_cannot_double_spent_with_different_nodes~: not needed since coverage in `cannot_replay_transactions` is sufficient
  - ~limits_client_config_downloads~: not present anymore
  - ~rejoin_consensus_split_peers~: not needed anymore due to aleph bft
- Ported
  - cannot_replay_transactions
    - https://github.com/fedimint/fedimint/pull/4550
    - https://github.com/fedimint/fedimint/blob/4926eab064755dfab092fafa5e5e6f9e06e147e4/devimint/src/tests.rs#L2061
  - audit_negative_balance_sheet_panics
    - https://github.com/fedimint/fedimint/blob/7a36cafa295ce30528ac1917575cb82ce8965907/modules/fedimint-dummy-tests/tests/tests.rs#L65
  - can_have_federations_with_one_peer
    - https://github.com/fedimint/fedimint/blob/af73efdb64c0e2c614353764d7e92d8a4ef4aebe/scripts/tests/test-ci-all.sh#L108
  - unbalanced_transactions_get_rejected
    - https://github.com/fedimint/fedimint/blob/7a36cafa295ce30528ac1917575cb82ce8965907/modules/fedimint-dummy-tests/tests/tests.rs#L113
  - rejoin_consensus_single_peer
  - rejoin_consensus_threshold_peers
  - ecash_can_be_recovered
    - https://github.com/fedimint/fedimint/blob/4926eab064755dfab092fafa5e5e6f9e06e147e4/devimint/src/tests.rs#L1224
  - can_get_signed_epoch_history
    - implicitly tested in backup and recovery
    - https://github.com/fedimint/fedimint/blob/4926eab064755dfab092fafa5e5e6f9e06e147e4/devimint/src/tests.rs#L1224
  - ecash_backup_can_recover_metadata
    - https://github.com/fedimint/fedimint/pull/4617
[...]

PR #4550, opened 2024-03-13
Port replay protection test

bradleystachurski:
Working towards https://github.com/fedimint/fedimint/issues/3234

Ports `cannot_replay_transactions` from the legacy integration tests.

https://github.com/fedimint/fedimint/blob/1132d742c2357fd70a0c3691636f80637edc0adb/integrationtests/tests/tests.rs#L348

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -1952,6 +1952,86 @@ pub async fn guardian_backup_test(dev_fed: DevFed, process_mgr: &ProcessManager)
     Ok(())
 }
 
+pub async fn cannot_replay_tx_test(dev_fed: DevFed) -> Result<()> {
+    log_binary_versions().await?;
+
+    #[allow(unused_variables)]
+    let DevFed {
+        bitcoind,
+        cln,
+        lnd,
+        fed,
+        gw_cln,
+        gw_lnd,
+        electrs,
+        esplora,
+    } = dev_fed;
+
+    let client = fed.new_joined_client("cannot-replay-client").await?;
+
+    // Make the start and spend amount the same so we spend all ecash
+    const CLIENT_START_AMOUNT: u64 = 5_000_000_000;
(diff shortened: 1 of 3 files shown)


dpc (approved):

devimint/src/util.rs
```diff
@@ -193,39 +193,64 @@ impl Command {
...
-            .run_inner()
+            .run_inner(false)
+            .await
+            .with_context(|| format!("command: {}", self.command_debug()))?;
+        let output = String::from_utf8(output.stdout)?;
+        Ok(output.trim().to_owned())
+    }
+
+    /// Run the command and get its output string, capturing stderr.
+    pub async fn out_string_capture_stderr(&mut self) -> Result<String> {
```
bradleystachurski, line 207: @dpc re: https://github.com/fedimint/fedimint/pull/4391#issuecomment-1992895340

Let me know if you're okay with this approach. I agree we shouldn't need stderr too often and this seemed minimally invasive. I do like having the extra confidence the double spend is failing for the expected reason instead of an unrelated error.

dpc: Yeah, it's fine for now.

But I feel like we're kind of inconsistent in what we are doing.

If we have a tool (`fedimint-cli`) that outputs json ... shouldn't it output json on `stderr` with an error as well?

It could print both human readable thing on `stderr` and json output on stderr that includes `error` field. 

@maan2003 What do you think?

maan2003: vote for json output with error on stdout, and non zero exit code.
and human readable output on stderr

bradleystachurski: That's a good idea :+1: 

For now I've removed from this PR and will address in a followup https://github.com/fedimint/fedimint/issues/4558

dpc (approved):

m1sterc001guy (approved):

PR #4556, opened 2024-03-13
chore(test-ci-all): lower number of parallel jobs on dev machines

dpc:
Need to be more conservative here.

### DIFF
--- a/scripts/tests/test-ci-all.sh
+++ b/scripts/tests/test-ci-all.sh
@@ -215,14 +215,26 @@ done
 
 parsed_test_commands=$(printf "%s\n" "${tests_with_versions[@]}")
 
-delay="${FM_TEST_CI_ALL_DELAY:-$((64 / $(nproc) + 1))}"
-load="${FM_TEST_CI_ALL_MAX_LOAD:-1000}"
 parallel_args=()
-export parallel_jobs='0'
+export parallel_jobs='1'
 
 if [ -z "${CI:-}" ]; then
   parallel_args+=(--eta)
 fi
+
+if [ -n "${FM_TEST_CI_ALL_JOBS:-}" ]; then
+  # when specifically set, use the env var
+  parallel_args+=(--jobs "${FM_TEST_CI_ALL_JOBS}")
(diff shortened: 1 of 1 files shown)


maan2003 (approved):

scripts/tests/test-ci-all.sh
```diff
@@ -215,35 +215,42 @@ done
 
 parsed_test_commands=$(printf "%s\n" "${tests_with_versions[@]}")
 
-delay="${FM_TEST_CI_ALL_DELAY:-$((64 / $(nproc) + 1))}"
-load="${FM_TEST_CI_ALL_MAX_LOAD:-1000}"
 parallel_args=()
-export parallel_jobs='0'
+export parallel_jobs='1'
```
bradleystachurski, line 219: [nit] Do we still need this?

scripts/tests/test-ci-all.sh
```diff
@@ -215,35 +215,42 @@ done
...
 if [ -z "${CI:-}" ]; then
   parallel_args+=(--eta)
 fi
 
+if [ -n "${FM_TEST_CI_ALL_JOBS:-}" ]; then
+  # when specifically set, use the env var
+  parallel_args+=(--jobs "${FM_TEST_CI_ALL_JOBS}")
+elif [ -n "${CI:-}" ]; then
+  # in CI, we know number of cpus works OK
+  parallel_args+=(--jobs +0)
```
bradleystachurski, line 230: Did you want to keep this `0` instead of `+0` so we can run as many jobs as possible in CI, or did you specifically want to switch back to jobs == num_cpu?

bradleystachurski (approved):

scripts/tests/test-ci-all.sh
dpc: Nope. Will fix at some point.

scripts/tests/test-ci-all.sh
dpc: In CI, let's do jobs == num_cpu for now too.

PR #4559, opened 2024-03-14
feat: add federation_id endpoint

Kodylow:
We expect this in the guardian ui and I need it for guardian authentication there.


fedimint-core/src/config.rs
```diff
@@ -120,6 +120,8 @@ pub struct PeerUrl {
 /// This includes global settings and client-side module configs.
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Encodable, Decodable)]
 pub struct ClientConfig {
+    #[serde(default)]
```
dpc, line 123: We should add a better comment about requirements and constrains of evolving `ClientConfig` as from the top of my head, I don't remember them.

I think now that `ClientConfig` it's not signed a new field should be fine, and old clients will just ignore it, so LGTM?

fedimint-core/src/config.rs
```diff
@@ -120,6 +120,8 @@ pub struct PeerUrl {
...
+    pub federation_id: Option<FederationId>,
```
m1sterc001guy, line 124: Why `Option`? To make it backwards compatible?

m1sterc001guy (dismissed):

fedimint-core/src/config.rs
Kodylow: Yes, happy to change and add versioning but this seemed more straightforward

dpc: Add a `// support(v0.2): can be changed into non-option` so we don't forget

dpc (approved):

elsirion (commented):
Can we just add an API endpoint that returns the federation id instead? Adding redundant information in the client config sounds like a bad idea, especially when taking into account backwards/forwards compatibility.

Kodylow:
Just changed to be federation_id api endpoint per eric, renaming pr

PR #4560, opened 2024-03-14
chore(dev-env): `just devimint-env`

dpc:
Copy&paste in `mprocs` is always annoying. I started improving tmux working in nested tmux, but then it occured to me, that I actually don't want devimint starting either.

All I really want is a shell inside devimint environment. I can `tail` logs myself if I need to (I usually don't). And that's what `just devimint-env` is.

Please give it a try locally before approving. High chance there are some details that can fail on different shells/systems, even if I tried to keep it *really* simple.

### DIFF
--- a/scripts/dev/devimint-env.sh
+++ b/scripts/dev/devimint-env.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+source scripts/_common.sh
+
+ensure_in_dev_shell
+build_workspace
+add_target_dir_to_path
+
+export DEVIMINT_DIR="${CARGO_BUILD_TARGET_DIR:-target}/devimint"
+rm -f "$DEVIMINT_DIR/devimint"
+
+function devimint_env {
+  set -euo pipefail
+
+  export RUST_LOG=info
+  export RUST_BACKTRACE=1
+
+  if [ -n "${PS1:-}" ]; then
+    PS1="[devimint] $PS1"
+  else
+    PS1="[devimint]"
+  fi
+
+  >&2 echo "Devimint Env Shell"
+  "${SHELL}"
+}
+export -f devimint_env
+
+env RUST_LOG=warn,jsonrpsee-client=off \
+  devimint --link-test-dir "${CARGO_BUILD_TARGET_DIR:-$PWD/target}/devimint" "$@" dev-fed \
+    --exec bash -c devimint_env
--- a/justfile.fedimint.just
+++ b/justfile.fedimint.just
@@ -106,6 +106,15 @@ exit-mprocs:
 tmuxinator:
   ./scripts/dev/tmuxinator/run.sh
 
+devimint-env *PARAMS:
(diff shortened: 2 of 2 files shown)


justinmoon:
I ran it and it worked nicely

justinmoon (approved):

bradleystachurski:
`complete` and `bind` are the only two commands giving issues when I try locally.

```
bash: complete: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
bash: bind: command not found
\[\]nix@lenovo-m75q-sandbox:~/fedimint$
```

dpc:
@bradleystachurski Default shell, distro?

scripts/dev/devimint-env.sh
```diff
@@ -0,0 +1,31 @@
...
+  fi
+
+  >&2 echo "Devimint Env Shell"
+  "${SHELL}"
+}
+export -f devimint_env
+
+env RUST_LOG=warn,jsonrpsee-client=off \
+  devimint --link-test-dir "${CARGO_BUILD_TARGET_DIR:-$PWD/target}/devimint" "$@" dev-fed \
+    --exec bash -c devimint_env
```
maan2003, line 31: maybe we should run the user shell script?

dpc: @maan2003 The problem with sourcing is that it only works in bash, and RN it works in fish as well. I'd actually want to replace aliases with simple scripts we put in a directory and add to path. But it can be done over time.

bradleystachurski (approved):
Looks like the root issue on my end is sourcing `fzf`. Disabling fixes and I'll debug on my end when I need fzf in devimint-env.

dpc:
> Looks like the root issue on my end is sourcing `fzf`. Disabling fixes and I'll debug on my end when I need fzf in devimint-env.

It should work, in principle. Possibly it shouldn't need to re-source it because it was already sourced in the outter shell. Maybe you need to add some guards from not running the same thigs twice. Nested shell `.bashrc`s were always tricky. I think I usually had an env var I'd set when sourcing, and then quick early when running in a nested shell. @bradleystachurski

elsirion:
Interesting! Would be cool if one could "atach" to an already running one (I guess that just means setting a bunch of env vars/aliases). That way there would be even less need for tmux, just open another shell.

dpc:
> Interesting! Would be cool if one could "atach" to an already running one (I guess that just means setting a bunch of env vars/aliases). That way there would be even less need for tmux, just open another shell.

IMO: too complex to explain, and the whole problem too custom-solution-prone. As in - you can start `just devimint-env` and then do whatever you want. e.g. just start tmux server/zellij, to get more shells if really needed.

In practice I never needed more than one shell anyway, and I have some ideas how to make devimint even faster too start.

Issue #4558, opened 2024-03-13
Include fedimint-cli errors in json output for devimint


bradleystachurski:
vote for json output with error on stdout, and non zero exit code.
and human readable output on stderr

_Originally posted by @maan2003 in https://github.com/fedimint/fedimint/pull/4550#discussion_r1522808845_

PR #4563, opened 2024-03-14
chore(fedimint-cli): write json error to stdout

bradleystachurski:
Closes https://github.com/fedimint/fedimint/issues/4558

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -2032,6 +2032,7 @@ pub async fn guardian_backup_test(dev_fed: DevFed, process_mgr: &ProcessManager)
 
 pub async fn cannot_replay_tx_test(dev_fed: DevFed) -> Result<()> {
     log_binary_versions().await?;
+    let fedimint_cli_version = crate::util::FedimintCli::version_or_default().await;
 
     #[allow(unused_variables)]
     let DevFed {
@@ -2096,10 +2097,25 @@ pub async fn cannot_replay_tx_test(dev_fed: DevFed) -> Result<()> {
         CLIENT_START_AMOUNT - CLIENT_SPEND_AMOUNT
     );
 
-    cmd!(double_spend_client, "reissue", double_spend_notes)
-        .run()
-        .await
-        .expect_err("double spend must fail");
+    // TODO(support:v0.2): remove
+    if VersionReq::parse(">=0.3.0-alpha")?.matches(&fedimint_cli_version) {
+        let reissue_error = cmd!(double_spend_client, "reissue", double_spend_notes)
+            .expect_err_json()
+            .await?
+            .get("error")
+            .expect("json error contains error field")
+            .as_str()
+            .context("not a string")?
+            .to_owned();
+
+        assert!(reissue_error.contains("The transaction had an invalid input"));
+    } else {
+        // v0.2 clients don't write json errors to stdout, so we can't parse
+        cmd!(double_spend_client, "reissue", double_spend_notes)
+            .run()
+            .await
+            .expect_err("double spend must fail");
+    }
 
     let double_spend_client_post_spend_balance = double_spend_client.balance().await?;
     assert_eq!(
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -40,9 +40,9 @@ use fedimint_wallet_client::api::WalletFederationApi;
 use fedimint_wallet_client::{WalletClientInit, WalletClientModule};
 use rand::thread_rng;
 use serde::{Deserialize, Serialize};
-use serde_json::{json, Value};
+use serde_json::Value;
 use thiserror::Error;
-use tracing::{debug, info};
+use tracing::{debug, error, info};
 use utils::parse_peer_id;
 
 use crate::client::ClientCmd;
@@ -98,23 +98,6 @@ impl fmt::Display for CliOutput {
     }
 }
 
-/// Types of error the cli return
-#[derive(Debug, Serialize, Deserialize)]
-#[serde(rename_all = "snake_case")]
-enum CliErrorKind {
-    NetworkError,
-    IOError,
-    InvalidValue,
-    OSError,
-    GeneralFederationError,
-    AlreadySpent,
-    Timeout,
-    InsufficientBalance,
-    SerializationError,
-    GeneralFailure,
-    MissingAuth,
-}
-
 /// `Result` with `CliError` as `Error`
 type CliResult<E> = Result<E, CliError>;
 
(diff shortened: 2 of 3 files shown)


fedimint-cli/src/lib.rs
```diff
@@ -506,6 +506,7 @@ impl FedimintCli {
                 let _ = writeln!(std::io::stdout(), "{output}");
             }
             Err(err) => {
+                let _ = writeln!(std::io::stdout(), "{err}");
```
dpc, line 509: stdout should output only json, so we want to wrap the error in some kind of:

```json
{
  error: "Things failed"
}
```

and in the future evolve to:

```
```json
{
  error: "Things failed"
  code: 134,
  hint: "Try again on Monday"
}
```

bradleystachurski: Let me know if I'm misinterpreting, but `err` is a `CliError` (https://github.com/fedimint/fedimint/blob/cd9b5a2/fedimint-cli/src/lib.rs#L124-L133) so this should output only json unless I'm misinterpreting.

```rust
/// Cli error
#[derive(Serialize, Error)]
#[serde(tag = "error", rename_all(serialize = "snake_case"))]
struct CliError {
    kind: CliErrorKind,
    message: String,
    #[serde(skip_serializing)]
    #[source]
    raw_error: Option<anyhow::Error>,
}
```

dpc: Is that how it is being printed already? I assumed `{err}` will format it as a human readable string instread of json. I haven't checked. If you have a good way to trigger it in devimint env, please let me know so I can confirm when I get to it.

bradleystachurski: > Is that how it is being printed already?

Yep, `CliError` implements `fmt::Display` which converts to json (https://github.com/fedimint/fedimint/blob/cd9b5a2/fedimint-cli/src/lib.rs#L219-L227)

```rust
impl fmt::Display for CliError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let mut json = serde_json::to_value(self).unwrap();
        if let Some(err) = &self.raw_error {
            json["raw_error"] = json!(*err.to_string())
        }
        write!(f, "{}", serde_json::to_string_pretty(&json).unwrap())
    }
}
```

> If you have a good way to trigger it in devimint env, please let me know so I can confirm when I get to it.

Thanks for asking! This uncovered how awkward this PR makes the output when interacting with the cli. You can repro in devimint env using

```bash
> just mprocs
> fedimint-cli spend 10000000
(this will succeed)
...
> fedimint-cli spend 10000000
2024-03-15T03:10:26.559703Z  WARN fedimint_cli::client: The client will try to double-spend these notes after the duration specified by the --timeout option to recover any unclaimed e-cash.
{
  "error": "CliError",
  "kind": "general_failure",
  "message": "failure",
  "raw_error": "Insufficient balance: requested 10000000 msat but only 0 msat available"
}
{
  "error": "CliError",
  "kind": "general_failure",
  "message": "failure",
  "raw_error": "Insufficient balance: requested 10000000 msat but only 0 msat available"
}
```

The issue I was hoping to resolve is parsing the json error in devimint tests, which we can't do cleanly using stderr since it contains the logs.

You can verify by redirecting stdout to /dev/null

```bash
> fedimint-cli spend 10000000 > /dev/null
2024-03-15T03:14:22.166187Z  WARN fedimint_cli::client: The client will try to double-spend these notes after the duration specified by the --timeout option to recover any unclaimed e-cash.
{
  "error": "CliError",
  "kind": "general_failure",
  "message": "failure",
  "raw_error": "Insufficient balance: requested 10000000 msat but only 0 msat available"
}
```

I think a different approach that may work better is to set `RUST_LOG=off` specifically when calling `expect_err_json` in devimint tests, then parsing the error from stderr.

Rubber ducking at this point, I'll ping for another review when this is cleaner 😄

dpc: IO we should just log the error (only the `raw_error`  part) as `error!(...) on stderr, and not the whole json anymore, as it's now redundant. @bradleystachurski 

So it will look like:

```
2024-03-15T05:13:18.774024Z  INFO task: Task finished task=state machine executor
2024-03-15T05:13:18.774024Z ERROR fedimint-cli: Insufficient balance: requested 10000000 msat but only 0 msat available
{
  "error": "CliError",
  "kind": "general_failure",
  "message": "failure",
  "raw_error": "Insufficient balance: requested 10000000 msat but only 0 msat available"
}
```

dpc: BTW. This error json structure doesn't make sense to me.

The human readable error should go into `error`.

The `kind` seems suspiciously useless, and `message` worthless.

I don't know if we copied it from some existing project that had a better use for these, or something, but if this is to be parsed now by our tests, it's the last chance to improve it for free.

dpc: @elsirion If you're around.

elsirion: I could imagine `kind` being used to distinguish between errors originating in modules vs. `fedimint-cli`? But I think the way it's used rn is redundant. Probably best to start over and cut the error struct down to the minimum: `{"error": "Some serialized error …"}`.

dpc: True. Adding a field is backward-compat, but removing it always potentially breaks someone.

bradleystachurski: Updated so we have a minimal error struct printed on stdout and the error message logged as an error on stderr:

```bash
> just mprocs
> fedimint-cli spend 10000000
(this will succeed)
...
> fedimint-cli spend 10000000
2024-03-16T20:59:08.258889Z  WARN fedimint_cli::client: The client will try to double-spend these notes after the duration specified by the --timeout option to recover any unclaimed e-cash.
2024-03-16T20:59:08.275700Z ERROR fedimint_cli: Insufficient balance: requested 10000000 msat but only 0 msat available
{
  "error": "Insufficient balance: requested 10000000 msat but only 0 msat available"
}
```

dpc:
image

tACK

dpc (approved):

devimint/src/tests.rs
```diff
@@ -2096,10 +2097,25 @@ pub async fn cannot_replay_tx_test(dev_fed: DevFed) -> Result<()> {
...
+        let reissue_error = cmd!(double_spend_client, "reissue", double_spend_notes)
+            .expect_err_json()
+            .await?
+            .get("error")
+            .expect("json error contains error field")
+            .as_str()
+            .context("not a string")?
+            .to_owned();
+
+        assert!(reissue_error.contains("The transaction had an invalid input"));
```
elsirion, line 2111: Feels like this could be an `assert_error` function, much like `expect_err_json`, that takes a predicate closure. A specialization of it may just take a string that should be contained in the error message (the only thing we can check for now).

elsirion (approved):
Looks good :)

devimint/src/tests.rs
bradleystachurski: Good call 👍 

Opened a followup PR https://github.com/fedimint/fedimint/pull/4605

PR #4566, opened 2024-03-15
chore: remove repetitive words

soonsouth:


maan2003 (approved):

m1sterc001guy (approved):

dpc (approved):
I'm surprised `typos` is not catching them.

elsirion:
We could use `grep -P '\b(\w+)\s+\1\b'` on all files (won't find duplicates across lines, but for that we'd need comment syntax-aware tool imo.

PR #4567, opened 2024-03-15
refactor: Move db.rs into server crates

m1sterc001guy:
There is no reason for db.rs to be in the common crates, since the client and server databases are separate and don't store the same things.

This PR moves db.rs for each of the modules into the server crates from common.


dpc:
Why do we need ln snapshot regenerated? Seems :sus:

m1sterc001guy:
> Why do we need ln snapshot regenerated? Seems :sus:

Because the ln client previously re-used the `LightningGateway` key from the server code. I shouldn't have done it that way, but it was available and was convenient to re-use code. Now that the struct exists on the client side, this PR adds the record to the backup.

dpc:
> Because the ln client previously re-used the `LightningGateway` key from the server code. I shouldn't have done it that way, but it was available and was convenient to re-use code. Now that the struct exists on the client side, this PR adds the record to the backup.

Do we need a migration?

dpc (approved):
LGTM, but do we need a migration?

m1sterc001guy:
> Do we need a migration?

Nope, its the same prefix and the same struct.

modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -908,6 +914,19 @@ mod fedimint_migration_tests {
...
+                                .await
+                                .collect::<Vec<_>>()
+                                .await;
+                            let num_gateways = gateways.len();
+                            ensure!(
+                                num_gateways > 0,
+                                "validate_migrations was not able to read any LightningGateways"
+                            );
+                            info!("Validated LightningGateways");
+                        }
```
elsirion, line 929: I guess this is why the DB snapshot needed to be regenerated. Before the test was just missing.

elsirion (approved):

PR #4576, opened 2024-03-15
feat: federation_id guardian endpoint

Kodylow:
Closed https://github.com/fedimint/fedimint/pull/4559 in favor of this to clean force pushes / git history, just adds federation_id endpoint to not affect global client config backwards compat but still be able to get it from the guardian ui.

### DIFF
--- a/fedimint-server/src/net/api.rs
+++ b/fedimint-server/src/net/api.rs
@@ -486,11 +486,11 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
         },
         api_endpoint! {
             FEDERATION_ID_ENDPOINT,
-            ApiVersion::new(0, 0),
+            ApiVersion::new(0, 2),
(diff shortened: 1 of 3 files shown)


fedimint-server/src/net/api.rs
```diff
@@ -484,6 +484,13 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
                 Ok(fedimint.cfg.get_invite_code().to_string())
             }
         },
+        api_endpoint! {
+            FEDERATION_ID_ENDPOINT,
+            ApiVersion::new(0, 0),
```
dpc, line 489: We need to increment a minor API version on the core API module, and put it here, I believe.

Kodylow: bumped to 0,2 , that was the last api added (guardian backup), should I bump to 0,3?

elsirion (approved):

dpc (approved):

PR #4578, opened 2024-03-15
fix(ci): test-ci-all running `parallel` with `--eta`

dpc:
`[ -t 1 ]` is true when running in terminal + a flag to disable it from Nix script.

### DIFF
--- a/scripts/tests/test-ci-all.sh
+++ b/scripts/tests/test-ci-all.sh
@@ -224,7 +224,7 @@ parsed_test_commands=$(printf "%s\n" "${tests_with_versions[@]}")
 parallel_args=()
 export parallel_jobs='1'
 
-if [ -z "${CI:-}" ]; then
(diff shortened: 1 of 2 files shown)


scripts/tests/test-ci-all.sh
```diff
@@ -224,7 +224,7 @@ parsed_test_commands=$(printf "%s\n" "${tests_with_versions[@]}")
 parallel_args=()
 export parallel_jobs='1'
 
-if [ -z "${CI:-}" ]; then
+if [ -z "${CI:-}" ] && [[ -t 1 ]] && [ -z "${FM_TEST_CI_ALL_DISABLE_ETA:-}" ]; then
```
bradleystachurski, line 227: Can we remove the CI env var check now that there's the file descriptor 1 check?

bradleystachurski (approved):

scripts/tests/test-ci-all.sh
dpc: I don't know. It seems to make sense to disable it if CI is set... kinda... too lazy to to change something that I don't have a strong feeling for...

justinmoon (approved):

PR #4580, opened 2024-03-16
chore(client): speed up by merging "waiting for rejected" call

dpc:
Use a small lru, just like we did before for merging module recovery
calls for session outcomes, to merge multiple calls for the same
acceptance/rejection query from multiple mint output state machines.

During mint bags creation, all notes are a different state
machine and there's a lot of them. However, they are all outputs
of the same transaction, so the result for all is going to be
exactly the same. Making dozens of same queries for exactly same
thing is a huge waste.

The mint output SM has two transition function there: one for
rejection, another for decryption shares (happy path). Ideally
we could merge both groups of calls, not just the one for rejection.
I'm thinking - if instead of asking for shares on each output,
we had some calls that returns shares for all outputs, we could
use the same technique as well.


dpc:
IIRC we used to have bags as outputs, and now each note is an output. We did it for simplification, but the big downside is that now we have many more state machines and calls we need to make, wasting resources, unless we do something about it.

Merging via cache is a bit convoluted, but not too terrible, IMO. There are some alternatives. We could have a concept of a group of state machines or something - but that also seems complex.

It's late, so I'm just pushing what I have to get feedback. I was able to confirm via logging that this indeed merges all the calls, and it seems like speeds up everything, but I need more time get some better estimates.

fedimint-client/src/lib.rs
```diff
@@ -288,11 +288,38 @@ dyn_newtype_define! {
...
+            return self.await_tx_accepted(txid).await;
+        };
+
+        let mut lru_lock = cache.lock().await;
+
+        let entry_arc = lru_lock
+            .get_or_insert(txid, || Arc::new(OnceCell::new()))
+            .clone();
+
+        // we drop the lru lock so requests for other `session_idx` can work in parallel
```
maan2003, line 314:
```suggestion
        // we drop the lru lock so requests for other `txid` can work in parallel
```

maan2003 (approved):
:tada:

dpc:
I need to see if I can stash this cache into more global place, etc.

dpc:
Oh. Under the hood this subscribes to in memory notifier, so ... this doesn't really do much. :D

PR #4581, opened 2024-03-16
chore(devimint): faster and cleaner start

dpc:


devimint/src/util.rs
```diff
@@ -331,25 +332,37 @@ const DEFAULT_RETRIES: usize = 30;
 pub async fn poll<Fut, R>(
     name: &str,
     // TODO: this should be `Duration`, not number of retries --dpc
```
bradleystachurski, line 334: You did it! :smile:

devimint/src/util.rs
```diff
@@ -331,25 +332,37 @@ const DEFAULT_RETRIES: usize = 30;
...
-                debug!("polling {name} failed with: {err:?}, will retry... ({i}/{retries})");
-                task::sleep(Duration::from_secs(1)).await;
+            Err(ControlFlow::Continue(err))
+                if attempt_start
+                    .duration_since(start)
+                    .expect("time goes forward")
+                    < timeout =>
+            {
+                debug!(target: LOG_DEVIMINT, %attempt, %err, "Polling {name} failed, will retry...");
+                task::sleep(Duration::from_millis(attempt.min(1000))).await;
```
bradleystachurski, line 357: `attempt.min(1000)` is pretty aggressive. Should we multiply attempt by a factor (100?) or did you want to sleep only a few millis?

bradleystachurski (approved):

devimint/src/util.rs
dpc: 1s is not agressive, IMO. All these commands are supposed to wrap in a handful of seconds if everything is OK.

bradleystachurski: Perhaps I'm misreading, but `Duration::from_millis(attempt.min(1000))` will wait 1 milli on the first attempt, then 2 until 1000 attempts.

dpc: Exactly. :D

1 ms + time of running the command

justinmoon (approved):

PR #4582, opened 2024-03-16
chore(fedimint-cli): print locking message  only if lock busy

dpc:
With recent dependency updates, this is now possible. I know because I contributed it, just so I can eventually get it to work like this. :D

### DIFF
--- a/fedimint-cli/src/db_locked.rs
+++ b/fedimint-cli/src/db_locked.rs
@@ -35,11 +35,17 @@ impl LockedBuilder {
                 .open(lock_path)
                 .with_context(|| format!("Failed to open {}", lock_path.display()))?;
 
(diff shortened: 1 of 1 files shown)


fedimint-cli/src/db_locked.rs
```diff
@@ -37,9 +37,17 @@ impl LockedBuilder {
 
             // TODO: Use https://github.com/cargo-bins/cargo-binstall/pull/1496 to
             // give user feedback only when the initial `new_try_exclusive` failed.
```
bradleystachurski, line 39: [nit] Can remove now

bradleystachurski (approved):

bradleystachurski (approved):

justinmoon (approved):

PR #4587, opened 2024-03-16
chore(latency-tests): reissue more notes

dpc:
Reissuing is so fast now, that it's hard to bench any potential improvements. Issue more notes, to make it work harder.

In my tests, when using `ci` build profile:

Before:

LATENCY REISSUE: min: 0.6s, avg: 0.7s, median: 0.6s, p90: 0.7s, max: 0.7s, sum: 19.5s

Now:

LATENCY REISSUE: min: 1.3s, avg: 1.4s, median: 1.4s, p90: 1.5s, max: 1.5s, sum: 42.4s

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -118,7 +118,8 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
 
     let client = fed.new_joined_client("latency-tests-client").await?;
     client.use_gateway(&gw_cln).await?;
-    fed.pegin_client(10_000_000, &client).await?;
+    let initial_balance_sats = 100_000_000;
+    fed.pegin_client(initial_balance_sats, &client).await?;
 
     // LN operations take longer, we need less iterations
     let iterations = 20;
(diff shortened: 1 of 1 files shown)


maan2003 (approved):

devimint/src/tests.rs
```diff
@@ -132,8 +133,13 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
             // for enough time to catch up a session end
             let iterations = 30;
             let mut reissues = Vec::with_capacity(iterations);
+            let amount_per_iteration_msats =
+                // use a highest 2^-1 amount that fits, to try to use as many notes as possible
```
elsirion, line 137: nit: 2^x-1

devimint/src/tests.rs
```diff
@@ -132,8 +133,13 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
...
+                ((initial_balance_sats * 1000 / iterations as u64).next_power_of_two() >> 1) - 1;
```
elsirion, line 138: TIL about `next_power_of_two` :exploding_head: 

Note that this might blow up if we ever switch away from base-2 denominations as the default. The only commitment we made and assume everywhere is that denominations are b^x.

elsirion (approved):

devimint/src/tests.rs
elsirion: Actually, it should fail gracefully, just not issuing as many notes anymore.

PR #4588, opened 2024-03-16
docs: add section on cachix to dev-env docs

### LINKED ISSUE #2149: Docs explain what trusting fedimint's cachix during initial dev shell setup means
It can speed up nix command builds. At least explain what that means since you'll be asked.

emilioziniades:
Ran into this footgun trying to get a dev environment set up. Since it's in the bootstrap script it makes sense to include it here as well. Closes #2149.

### DIFF
--- a/docs/dev-env.md
+++ b/docs/dev-env.md
@@ -73,6 +73,25 @@ To use a different shell for `nix develop`, try `nix develop -c zsh`. You can al
 don't want to remember about it. That's the recommended way to use a different shell
 for `nix develop`.
 
+## Cachix binary cache
+
+Fedimint uses a [Cachix](https://www.cachix.org/) binary cache to cache builds.
+To benefit from this cache and avoid building everything from scratch, you must
+ensure that your user is a [trusted user](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-trusted-users).
+You can do this by modifying `/etc/nix/nix.conf`, adding the following line.
+
+```
+trusted-users = the_name_of_your_user
+```
+
+Alternatively, if you do not want to add your user to the list of trusted users, you can
+run the following command, which will add <https://fedimint.cachix.org> and its public key
+to your nix configuration.
+
+```
+nix develop .#bootstrap -c cachix use fedimint
+```
+
 ## Setting up `direnv` or `lorri`
 
 One of the biggest QoL improvements you can do when working with flake-enabled projects


docs/dev-env.md
```diff
@@ -86,6 +86,30 @@ project they will automatically set up the environment for you, without starting
...
+nix develop .#bootstrap -c cachix use fedimint
+```
+
+This invokes the Cachix CLI and updates your nix configuration. It adds the
+Cachix cache as a substituter and trusts its public key. See the nix
+documentation on [substituters](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-substituters)
+and [trusted public keys](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-trusted-public-keys)
+for more information. 
+
+Now you can run `nix develop` and benefit from the binary cache.
```
dpc, line 111: `nix develop` should ask on first start if you want to trust our cachix repo as well.

emilioziniades: Yeah agree. Not sure if that's possible. 

Did a little digging, and it's possible to "extend" nix.conf on a flake-by-flake basis using the `nixConfig` output.

I see this already exists:
https://github.com/fedimint/fedimint/blob/85f4e4944e582f9e6e4ca91bea287c5ef9dde26b/flake.nix#L377-L380

BUT, the catch is that this only has an effect when the user sets the [`accept-flake-config`](https://nixos.org/manual/nix/stable/command-ref/conf-file#conf-accept-flake-config) global setting. 

So, maybe the solution is to encourage the user to turn on "accept-flake-config" in the docs.

dpc: > Yeah agree. Not sure if that's possible.

@emilioziniades I mean - that's already happening, or is at least supposed to.

https://github.com/fedimint/fedimint/blob/8d88f30ab25f55e37cbe4d46572cba2dfdeda434/flake.nix#L377

is responsible for it, and it always worked for me (thought I already answered it and saved my choice so I don't see it anymore).

dpc: > BUT, the catch is that this only has an effect when the user sets the [`accept-flake-config`](https://nixos.org/manual/nix/stable/command-ref/conf-file#conf-accept-flake-config) global setting.

Nope. I don't have that set, and would not enable it globally. I personally don't want to trust any binary caches on my personal devices. I still get the prompt asking.

emilioziniades: > Nope. I don't have that set, and would not enable it globally. I personally don't want to trust any binary caches on my personal devices. I still get the prompt asking.

Makes complete sense.

> @emilioziniades I mean - that's already happening, or is at least supposed to.

Ah apologies, I misunderstood you.

When I run `nix develop` for the first time, I get this warning
```
warning: ignoring untrusted substituter 'https://fedimint.cachix.org', you are not a trusted user.
```
And then it proceeds to build everything from source.

This is what my nix.conf (using nix-darwin) looks like:

```
# WARNING: this file is generated from the nix.* options in
# your nix-darwin configuration. Do not edit it!
allowed-users = *
auto-optimise-store = false
build-users-group = nixbld
builders = 
cores = 0
experimental-features = nix-command flakes
extra-sandbox-paths = 
max-jobs = auto
require-sigs = true
sandbox = false
sandbox-fallback = false
trusted-substituters = 
trusted-users = root
substituters = https://cache.nixos.org https://cache.nixos.org/
trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
```

But when I add my user to trusted-users, then it uses cachix without any issues.

Have you added your user to `trusted-users`?

And, do you know if there are any security implications to doing so?

dpc: >  you are not a trusted user.


OOooooooh. I see. Nix build daemon will only trust users in `trusted-users` with substituting binary caches. That's why it won't even display the prompt for users that don't have their unix user added there.

I didn't think of that.

dpc: So yeah, I guess it's worth documenting that the user needs to be in a `trusted-users` config for binary caching to work.

emilioziniades: > So yeah, I guess it's worth documenting that the user needs to be in a `trusted-users` config for binary caching to work.

Ok cool, I've modified the docs change to reflect what we've discussed here - would you mind taking another look?

docs/dev-env.md
```diff
@@ -73,6 +73,25 @@ To use a different shell for `nix develop`, try `nix develop -c zsh`. You can al
...
+```
+trusted-users = the_name_of_your_user
+```
+
+Alternatively, if you do not want to add your user to the list of trusted users, you can
+run the following command, which will add <https://fedimint.cachix.org> and its public key
+to your nix configuration.
+
+```
+nix develop .#bootstrap -c cachix use fedimint
```
dpc, line 92: I don't think this actually adds user to trusted-users... . I don't see it anywhere in the code.

elsirion: That's not what the comment claims, the command edits `~/.config/nix/nix.conf` and adds the binary substituter there.

elsirion (approved):

docs/dev-env.md
dpc: Does it thought?

emilioziniades: To clarify, the section is saying "there are two ways to use the binary cache: either add yourself to trusted-users, or run this command (which happens to modify nix.conf in a different way)"

> Does it thought?

Yeah so for me, before adding my user to `trusted-users`, I ran `cachix use fedimint` and my `/etc/nix/nix.conf` was updated. It added <https://fedimint.cachix.org> to `trusted-substituters` and added the public key to `trusted-public-keys`

Looks like the fallback is `/etc/nix/nix.conf` for untrusted users: https://docs.cachix.org/faq#what-happens-when-i-run-cachix-use-both-immediately-and-any-stateful-effects-for-the-future. 

Doesn't make much sense to me- I haven't had to use a non-nixpkgs binary cache before so this is all new to me.

dpc (approved):

docs/dev-env.md
elsirion: > Does it thought?

For me it did, I removed cachix from the file and ran `watch cat ~/.config/nix/nix.conf` and saw it happen ^^

PR #4596, opened 2024-03-18
chore(devimint): polling maybe a bit too aggressive

dpc:
Brad was right, I was wrong.

I'm not worried about the performance, but the flood of logging messages is inconvenient.

### DIFF
--- a/devimint/src/util.rs
+++ b/devimint/src/util.rs
@@ -335,14 +335,14 @@ macro_rules! poll_eq {
 pub(crate) use cmd;
 
 const DEFAULT_POLL_TIMEOUT: Duration = Duration::from_secs(30);
-/// Will retry calling `f`.
+/// Retry until `f` succeeds or timeout is reached
+///
 /// - if `f` return Ok(val), this returns with Ok(val).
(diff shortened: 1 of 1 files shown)


dpc:
Need to use it a bit longer while debugging to see if it's OK.

devimint/src/util.rs
```diff
@@ -335,14 +335,14 @@ macro_rules! poll_eq {
...
 
 const DEFAULT_POLL_TIMEOUT: Duration = Duration::from_secs(30);
-/// Will retry calling `f`.
+/// Retry until `f` succeeds or timeout is reached
+///
 /// - if `f` return Ok(val), this returns with Ok(val).
 /// - if `f` return Err(Control::Break(err)), this returns Err(err)
-/// - if `f` return Err(ControlFlow::Continue(err)), retries a maximum of
-///   `retries` times.
+/// - if `f` return Err(ControlFlow::Continue(err)), retries untile timeout
```
bradleystachurski, line 342: [nit] s/untile/until

bradleystachurski (approved):

elsirion (approved):

PR #4598, opened 2024-03-18
fix: add_target_dir_to_path using wrong dir

dpc:


maan2003 (approved):

elsirion (approved):
Could we just symlink dev -> debug so we don't have to special-case this everywhere? :rofl:

PR #4601, opened 2024-03-18
chore: remove async from TaskGroup::spawn

maan2003:
fixes https://github.com/fedimint/fedimint/issues/4493

command used for replace:

	rust-analyzer ssr '$tg.spawn($n, $f).await ==>> $tg.spawn($n, $f)'

PR #4602, opened 2024-03-18
chore: use tokio::sync::oneshot even on wasm

maan2003:
futures::channel::oneshot has #[must_use] which incorrectly warns on wasm currently about receiver being unused


elsirion (approved):
We could just go straight for #4601 if you want

maan2003:
yes go for #4601, merge queue doesn't behave with stacked PR

PR #4603, opened 2024-03-18
chore: improve error of download config

maan2003:
now returns the correct error with added context


fedimint-core/src/config.rs
```diff
@@ -239,20 +239,23 @@ impl ClientConfig {
...
+                    }
 
                     sleep(Duration::from_millis(500)).await;
                 }
             }
+            attempt += 1;
         }
-
-        Err(anyhow!("Failed to download client config"))
     }
```
elsirion, line 259: Can we make this a utility function (or do we have one already)?

```rust
// Return Ok on first success or last Err after a given number of retries
async fn(usize, Fn() -> Future<Output=Result<T, E>>) -> Result<T, E>
```

That would at least encapsulate that mutability (`attempt`) in the loop a bit.

elsirion (approved):
Certainly an improvement, but candidate for further refactoring

fedimint-core/src/config.rs
maan2003: fixed

fedimint-core/src/util/mod.rs
```diff
@@ -307,13 +309,57 @@ pub fn handle_version_hash_command(version_hash: &str) {
...
+    let op_name = op_name.into();
+    assert_ne!(max_attempts, 0, "max_attempts must be greater than 0");
+    let mut attempts = 0;
+    loop {
+        attempts += 1;
+        match op_fn().await {
+            Ok(result) => return Ok(result),
+            Err(err) if attempts < max_attempts => {
+                // run closure op_fn again
+                info!(
```
elsirion, line 339: nit: `info` seems like a weird log level, given that we are still retrying I'd go with `debug` and only `warn` when actually failed.

elsirion (dismissed):

justinmoon (approved):

fedimint-core/src/util/mod.rs
```diff
@@ -307,13 +309,67 @@ pub fn handle_version_hash_command(version_hash: &str) {
...
+/// # Returns
+///
+/// - If the closure runs successfully, the result is immediately returned
+/// - If the closure did not run successfully for `max_attempts` times, the
+///   error of the closure is returned
+pub async fn retry<F, Fut, T>(
+    op_name: impl Into<String>,
+    op_fn: F,
+    interval: Duration,
+    max_attempts: u32,
```
dpc, line 324: max_attempts < timeout , just saying.

Isn't this `poll` from `devimint` where I just added deadlines and turned interval to be growing? Can we move it here from there and rename?

elsirion: That one uses `ControlFlow`, while here we want that to be implicitly implied by `Result`. `retry` could be implemented using `poll` though.

dpc: That control flow there is a bit of a PITA, BTW (because of type inference on two variants). Not sure how badly we need it. In testing if there's an error that should really terminate polling... maybe we should just panic, instead of complicating common case? Just thinking aloud.

But one way or the other - would prefer to see a timeout/deadline approach here.

elsirion: Will leave it to @maan2003, don't have a strong opinion here.

dpc:
Let's land it already, doesn't matter all that much in a great scheme of things, Manmeet is free to follow up (or not). :D

fedimint-core/src/util/mod.rs
maan2003: :shrug: devimint and this retry can just stay separate, attempt + interval is pretty close to deadline though

PR #4615, opened 2024-03-19
feat: Support description hashes for invoices

benthecarman:
This is needed along with #3820 for the lightning address server. Without being able to set the description hash some clients will reject the invoice as invalid and will make all resulting zaps invalid as well.


m1sterc001guy (approved):
LGTM

elsirion (approved):

Issue #4616, opened 2024-03-19
remove Database::ensure_global calls


maan2003:
I want to use same database(with_prefix) for different fedimint clients and fedimint code calls ensure_global and errors out.

cc @m1sterc001guy

m1sterc001guy:
The client side migrations today assume that the global database is the db namespace where the states are stored. For any use case that needs multi-level nesting of isolated databases (e.g multiple clients within the gateway), this check doesn't make sense anymore and can be removed.

PR #4620, opened 2024-03-19
fix: remove Database::ensure_global calls

maan2003:
> The client side migrations today assume that the global database is the db namespace where the states are stored. For any use case that needs multi-level nesting of isolated databases (e.g multiple clients within the gateway), this check doesn't make sense anymore and can be removed.

quote from https://github.com/fedimint/fedimint/issues/4616

fixes  https://github.com/fedimint/fedimint/issues/4616

### DIFF
--- a/fedimint-client/src/backup.rs
+++ b/fedimint-client/src/backup.rs
@@ -264,13 +264,11 @@ impl Client {
 
     async fn load_previous_backup(&self) -> Option<ClientBackup> {
         let mut dbtx = self.db.begin_transaction_nc().await;
-        dbtx.ensure_global().expect("global tx");
         dbtx.get_value(&LastBackupKey).await
     }
 
     async fn store_last_backup(&self, backup: &ClientBackup) {
         let mut dbtx = self.db.begin_transaction().await;
-        dbtx.ensure_global().expect("global tx");
         dbtx.insert_entry(&LastBackupKey, backup).await;
         dbtx.commit_tx().await;
     }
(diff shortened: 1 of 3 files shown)


m1sterc001guy (approved):

dpc (approved):
OK, but I wish we could replace it with something.

Eg. every `Database` carrying a `&'static str` instead of counter, so that when we isolate the database, we can set kind of a "context" that can be asserted in code using it.

The reason for all of this , is that during refactoring work it's very easy to pass a wrong (too or not enough isolated) `Database` somewhere, and the wrong kind of fun follows.

By replacing dumb counter with a context, we could preserve this protetion.

@maan2003 @m1sterc001guy WDYT?

m1sterc001guy:
I think a "namespace context" could work as a replacement for this check. The client state machine migrations need to check that they're not in the module context though, so we would need some way of registering a module context (e.g ln, mint, wallet, meta) so that we're able to do this check. Or maybe its good enough to just have one context string for all modules.

dpc:
> Or maybe its good enough to just have one context string for all modules.

One is enough. We really need two ATM: `"global"`, `"module"`. IIRC @maan2003 wants to add ability to run multiple Federations inside same database, and it will keep working, might set it to some 3rd value like `"multi"`.

m1sterc001guy:
>  to add ability to run multiple Federations inside same database

I would like to do this for the gateway as well.

Made https://github.com/fedimint/fedimint/issues/4629 to track, I think this is good to land to unblock @maan2003

fedimint-backports:
- #4632

Issue #4283, opened 2024-02-09
Make audit functionality more useful


elsirion:
Currently, we don't enforce correct implementation of the audit functionality for liability modules. If they just don't implement it we aren't accounting their liabilities, appearing overfunded, which is ok for Fedimint.

This was ok for our internal modules, but we need more sanity checks to avoid external module authors forgetting to implement it correctly. Fedimint consensus should independently record in/out flows of modules and check that against the self-reported audit by the modules. Thoughts @joschisan @bradleystachurski ?

bradleystachurski:
> Fedimint consensus should independently record in/out flows of modules and check that against the self-reported audit by the modules.

Conceptually I think that's a useful system.

Please feel free to redirect the conversation if you think it's too early to talk about approaches, but did you have a high level approach in mind? For each liability module we could track `<LiabilityModule>AuditItemKey::{Issuance,Redemption}Total` whenever consensus observes a valid input/output for the associated module, then compare the aggregate keys to whatever is reported by the module's audit.

How should we handle discrepancies?

elsirion:
We'll want to figure out what to do about modules that fail this test: flag and block or shut down federation?

PR #4621, opened 2024-03-19
feat: charge deposit fees by default to counter dust attacks

### LINKED ISSUE #4611: Default to non-zero peg-in fee
To prevent abuse, and (in future) move the cost of spending pegged-in UTXOs from the spender to the creator of UTXO, mints should always charge a peg-in fee.

https://github.com/fedimint/fedimint/blob/5070824f364ae40d35d148422dc3f73ffe68d6a2/modules/fedimint-wallet-common/src/config.rs#L109

Probably something like 1000sats per-peg-in should be OK.

We need to expose ability to customize it and discover it by the client applications, so it can be integrated in the UX.

elsirion:
Fixes #4611

### DIFF
--- a/modules/fedimint-wallet-common/src/config.rs
+++ b/modules/fedimint-wallet-common/src/config.rs
@@ -15,6 +15,10 @@ use crate::envs::FM_PORT_ESPLORA_ENV;
 use crate::keys::CompressedPublicKey;
 use crate::{PegInDescriptor, WalletCommonInit};
 
+/// Helps against dust attacks where an attacker deposits UTXOs that, with
+/// higher fee levels, cannot be spent profitably.
+const DEFAULT_DEPOSIT_FEE_SATS: u64 = 1000;
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct WalletGenParams {
     pub local: WalletGenParamsLocal,
@@ -36,6 +40,7 @@ impl WalletGenParams {
                     ))
                     .expect("Failed to parse default esplora server"),
                 },
+                fee_consensus: Default::default(),
             },
         }
     }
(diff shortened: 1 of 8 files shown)


joschisan:
Should we take a similar fee for outgoing payments while we are at it?

joschisan:
> No, outgoing payments already pay a (typically much higher) on-chain fee.

Could not the same argument be made against input fees? I am wondering if there might be a not yet known DOS vector for outgoing payments as well that is prevented by fees in a similar manner.

dpc (approved):

modules/fedimint-wallet-common/src/config.rs
```diff
@@ -52,6 +57,11 @@ pub struct WalletGenParamsConsensus {
     pub finality_delay: u32,
     /// See [`WalletConfigConsensus::client_default_bitcoin_rpc`].
     pub client_default_bitcoin_rpc: BitcoinRpcConfig,
+    /// Fees to be charged for deposits and withdraws _by the federation_ in
+    /// addition to any on-chain fees.
+    ///
+    /// Deposit fees in particular are a protection against dust attacks.
+    pub fee_consensus: FeeConsensus,
```
bradleystachurski, line 64: Out of scope, but it may be useful to dynamically update fees. It's an edge case, but if a federation observes the threshold is too low for a given fee environment it would be useful to update without restarting the federation with a new config.

bradleystachurski (approved):
Looks good, thanks for tackling!

Is there a way to fetch the total fee balance of the federation? If not, should we create an issue to track?

Edit: I suppose a hacky workaround is `audit.net_assets()` assuming the only way net assets increase is fees.

modules/fedimint-wallet-common/src/config.rs
elsirion: Tracked as part of #4751. I hope we can get generalized dynamic config somehow, but worst case we could special-case certain config parts. @joschisan is researching this problem, the general case is quite complex unfortunately.

elsirion:
> Edit: I suppose a hacky workaround is `audit.net_assets()` assuming the only way net assets increase is fees.

This is the way for now. With #4283 we could calculate it even more effortlessly and make it a public API (audits are private for now due to DoS concerns).

modules/fedimint-wallet-common/src/config.rs
bradleystachurski: Thanks for the link! Sounds good, certainly not in scope for this PR.

PR #4626, opened 2024-03-19
chore: use info tracing level for htlc stream open failure

douglaz:
I had to enable debug logging just to see this message.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -576,7 +576,7 @@ impl Gateway {
                                 }
                             }
                             Err(e) => {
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -576,7 +576,7 @@ impl Gateway {
                                 }
                             }
                             Err(e) => {
-                                debug!("Failed to open HTLC stream: {e:?}");
+                                info!("Failed to open HTLC stream: {e:?}");
```
m1sterc001guy, line 579: `error!` instead of `info!`?

m1sterc001guy (approved):

elsirion (approved):

Issue #3795, opened 2023-11-30
Allow setting up federation via `fedimint-cli admin …` commands


elsirion:
This seems easier than learning typescript.

PR #3796, opened 2023-11-30
WIP: allow setting up federation via `fedimint-cli`

elsirion:
Fixes #3795

PR #4628, opened 2024-03-19
feat: allow setting up federation via `fedimint-cli`

douglaz:
Port of https://github.com/fedimint/fedimint/pull/3796

Plus missing methods and minor improvements

Closes https://github.com/fedimint/fedimint/issues/3795

In future we may create another command that does the full setup using these subcommands and use them to setup devimint, perhaps replacing the current method: https://github.com/fedimint/fedimint/blob/0b8cdd855cc46e58a74fa04678849bddd8da3832/devimint/src/federation.rs#L501

Calling example:
```bash
fedimint-cli admin dkg --ws ws://127.0.0.1:24329/ ws-status
```

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -14,7 +14,7 @@ use std::time::Duration;
 use std::{fs, result};
 
 use bip39::Mnemonic;
-use clap::{CommandFactory, Parser, Subcommand};
+use clap::{Args, CommandFactory, Parser, Subcommand};
 use db_locked::LockedBuilder;
 use fedimint_aead::{encrypted_read, encrypted_write, get_encryption_key};
 use fedimint_bip39::Bip39RootSecretStrategy;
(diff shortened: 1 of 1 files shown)


fedimint-cli/src/lib.rs
```diff
@@ -694,6 +733,83 @@ impl FedimintCli {
                         .map_err_cli_msg("invalid response")?,
                 ))
             }
+            Command::Admin(AdminCmd::WsStatus) => {
```
dpc, line 736: Thought: We should split handling admin commands to a separate function. The whole thing is getting way too long.

dpc (approved):

fedimint-cli/src/lib.rs
douglaz: Ok, reorganized a bit now it's a subcommand and calls another function.

elsirion:
Should have a devimint test to see if it actually works :P

dpc (approved):

elsirion (approved):
Just be aware this is very likely to break (or not even work in the first place) without testing. But since it's just some dev tooling let's get it in and iterate.

fedimint-backports:
- #4713

PR #4630, opened 2024-03-19
chore(ci): every hour run tests and report failure to #alerts

dpc:
I think we got flakes under control, but we must remain vigilant.

This workflow will give us 24 run a day, to make sure any new flakes us quickly spotted.

### DIFF
--- a/.github/workflows/hourly-run.yml
+++ b/.github/workflows/hourly-run.yml
@@ -1,0 +1,107 @@
+on:
+  schedule:
+    - cron:  '30 * * * *' # every hour at `x:30`
+  workflow_dispatch:
+
+name: Hourly check
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  self-care:
+    name: Flake self-check
+    # no one is in the rush for this to funish, so use default (free) Ubuntu instances
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check Nix flake inputs
+        uses: DeterminateSystems/flake-checker-action@v5
+        with:
+          fail-mode: true
+          nixpkgs-keys: nixpkgs
+
+  test:
+    if: github.repository == 'fedimint/fedimint'
+    strategy:
+      matrix:
+        host:
+          - linux
+        include:
+          - host: linux
+            # no one is in the rush for this to funish, so use default (free) Ubuntu instances
+            runs-on: ubuntu-latest
+            timeout: 75
+            run-tests: true
+
+    name: "Test on ${{ matrix.host }}"
+    runs-on: ${{ matrix.runs-on }}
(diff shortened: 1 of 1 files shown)


dpc:
@elsirion It seems like a good idea, and the only potential downside is an extra buildjet cost.

.github/workflows/hourly-run.yml
```diff
@@ -0,0 +1,105 @@
...
+        with:
+          name: fedimint
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+        continue-on-error: true
+
+      - name: Tests (5 times more)
+        run: |
+          set -euo pipefail
+          # first one might do nothing
+          nix build -L .#wasm32-unknown.ci.ciTestAll5Times --keep-failed
```
bradleystachurski, line 57: Interesting, ciTestAll5Times will sometimes not run? Same for wasmTest? Or is this actually a ciTestAll10Times?

.github/workflows/hourly-run.yml
```diff
@@ -0,0 +1,105 @@
...
+
+  test:
+    if: github.repository == 'fedimint/fedimint'
+    strategy:
+      matrix:
+        host:
+          - linux
+        include:
+          - host: linux
+            runs-on: buildjet-8vcpu-ubuntu-2004
```
bradleystachurski, line 29: [nit] Should we consider bumping our buildjet runners from 2004 to 2204?

bradleystachurski (dismissed):
CI is in a healthy state and high signal, so I think this is a great idea 👍 

Back-of-the-envelope costs:
- ~50min for ciTestAll5times and wasmTest ([recent run](https://github.com/fedimint/fedimint/actions/runs/8351416363/job/22859711551))
- $0.016/min `buildjet-8vcpu-ubuntu-2204` ([pricing](https://buildjet.com/for-github-actions/docs/about/pricing))
- 24 runs/day
- ~$19.20/day in buildjet costs

elsirion:
Can we maybe run a bit less often, every 2-3h or so? This sounds like a great reason to switch to self-hosted build infra that is just there 24/7, then we could just run this whenever the load is low.

.github/workflows/hourly-run.yml
dpc: It's already in the Nix cache, so Nix will just not run it. :shrug: 

We just ensure it's there, then we can run with `--rebuild` which will run it.

.github/workflows/hourly-run.yml
dpc: Ubuntu 20.04 :D

dpc:
Switched it to run on free ubuntu boxes.

bradleystachurski (approved):

maan2003 (approved):
lets see how it goes, we can always revert

dpc:
> lets see how it goes, we can always revert

That's the right attitude! :)

PR #4631, opened 2024-03-20
chore: improve gateway webserver logging

m1sterc001guy:
Spoke too soon on the flakiness with `test_gateway_configuration` :/ https://github.com/fedimint/fedimint/actions/runs/8350129601/job/22855854147

It's currently difficult to tell why the webserver is not coming back up after we change the gateway's password. This PR adds some logging to make it easier to tell what's happening.

I suspect there might be a race condition that is causing the webserver to still bind to the port, causing the restart to fail.


m1sterc001guy:
Actually, I would prefer to refactor this so it doesn't require a restart of the webserver. Closing this.

PR #4627, opened 2024-03-19
 chore: `just sign-release v0.3.0-rc.0` by dpc 

dpc:

PR #4635, opened 2024-03-20
chore: `just sign-release v0.3.0-rc.0` by @elsirion

elsirion:
Builds on #4627


elsirion:
Too much other stuff going on, I'll check in some signatures once 0.3.0 is released, signing rc's is a nice exercise but adds too much noise rn.

PR #4639, opened 2024-03-20
chore: update consensus version

maan2003:
we forgot to update consensus version for rc.0


elsirion:
Afaik we didn't introduce any consensus-breaking changes? Ideally we'd bump on every such change, just like with the API version.

joschisan:
> Afaik we didn't introduce any consensus-breaking changes?

I did not introduce any

maan2003:
but u32::MAX needs to be replaced for sure

elsirion:
Oh right, we should probably also fix this on master now that *normally* everything should be compatible (I wonder why version negotiation even works in compatibility tests rn …).

I think we should just set it all to 2.0, I think any updates we will make in the near future will be "soft forks", so would only bump the minor consensus version. E.g. if we introduced new consensus items that would be 2.1, 2.2, …. Only if we removed stuff that would be breaking and lead to a 3.0 bump.

dpc:
> Oh right, we should probably also fix this on master now that _normally_ everything should be compatible

Oh, I thought that normally we would keep it at INT_MAX on master, just to mark it as incompatible with anything (potentially).

elsirion:
> Oh, I thought that normally we would keep it at INT_MAX on master, just to mark it as incompatible with anything (potentially).

That was the reasoning when I didn't think we'd aim for compatibility anytime soon :sweat_smile: I think by now we are good enough at reviewing for breaking changes that

> I kind of dislike INT_MAX […] especially the fact that we need to remember about changing it

I'm coming to the same conclusion :laughing:

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -45,7 +45,7 @@ use crate::contracts::{Contract, ContractId, ContractOutcome, Preimage, Preimage
 use crate::route_hints::RouteHint;
 
 pub const KIND: ModuleKind = ModuleKind::from_static_str("ln");
-const CONSENSUS_VERSION: ModuleConsensusVersion = ModuleConsensusVersion::new(0, 0);
```
elsirion, line 48: Interesting, looks like we forget to set this to MAX on master.

elsirion (approved):

PR #4642, opened 2024-03-20
feat: Gateway Authentication using middleware

m1sterc001guy:
Previously we had a naively implementation for Gateway authentication. We were using `ValidateRequestHeaderLayer::bearer`, which required a password to be specified during the setup of the webserver. Therefore, in order to support changing the password via the `set_configuration` API, we needed to restart the webserver. This is not ideal because it leads to longer wait times, failed requests, and made our tests flaky https://github.com/fedimint/fedimint/actions/runs/8350129601/job/22855854147

These PR removes `ValidateRequestHeaderLayer::bearer` and introduces middleware functions that intercept the requests to the gateway before they are processed and validate them again the current password that is stored within the gateway. This way, when the password is updated, it is not required to restart the webserver and subsequent requests will not fail.

### DIFF
--- a/gateway/ln-gateway/src/rpc/rpc_server.rs
+++ b/gateway/ln-gateway/src/rpc/rpc_server.rs
@@ -1,5 +1,6 @@
-use std::net::SocketAddr;
-
+use axum::extract::Request;
+use axum::http::{header, StatusCode};
+use axum::middleware::{self, Next};
 use axum::response::IntoResponse;
 use axum::routing::{get, post};
 use axum::{Extension, Json, Router};
@@ -10,24 +11,18 @@ use fedimint_ln_client::pay::PayInvoicePayload;
 use serde_json::json;
 use tokio::net::TcpListener;
 use tower_http::cors::CorsLayer;
-use tower_http::validate_request::ValidateRequestHeaderLayer;
-use tracing::{error, instrument};
+use tracing::{error, info, instrument};
 
 use super::{
     BackupPayload, BalancePayload, ConnectFedPayload, DepositAddressPayload, InfoPayload,
     LeaveFedPayload, RestorePayload, SetConfigurationPayload, WithdrawPayload, V1_API_ENDPOINT,
 };
-use crate::db::GatewayConfiguration;
 use crate::rpc::ConfigPayload;
 use crate::{Gateway, GatewayError};
 
-pub async fn run_webserver(
-    config: Option<GatewayConfiguration>,
-    bind_addr: SocketAddr,
-    gateway: Gateway,
-    task_group: &mut TaskGroup,
-) -> anyhow::Result<()> {
-    let v1_routes = v1_routes(config, gateway.clone());
+/// Creates the webserver's routes and spawns the webserver in a separate task.
+pub async fn run_webserver(gateway: Gateway, task_group: &mut TaskGroup) -> anyhow::Result<()> {
+    let v1_routes = v1_routes(gateway.clone());
     let api_v1 = Router::new()
         .nest(&format!("/{V1_API_ENDPOINT}"), v1_routes.clone())
         // Backwards compatibility: Continue supporting gateway APIs without versioning
@@ -35,7 +30,7 @@ pub async fn run_webserver(
 
     let handle = task_group.make_handle();
     let shutdown_rx = handle.make_shutdown_rx().await;
-    let listener = TcpListener::bind(&bind_addr).await?;
+    let listener = TcpListener::bind(&gateway.listen).await?;
     let serve = axum::serve(listener, api_v1.into_make_service());
     task_group.spawn("Gateway Webserver", move |_| async move {
         let graceful = serve.with_graceful_shutdown(async {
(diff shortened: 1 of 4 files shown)


justinmoon:
@okjodom Are you able to take a look? I think you're most familiar with this code.

okjodom:
> @okjodom Are you able to take a look? I think you're most familiar with this code.

yes. it's in my review bucket for the day 🔖

m1sterc001guy:
> This diff contains name suggestions and some DRY suggestions.

Can you post the comments to Github?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -44,49 +39,118 @@ pub async fn run_webserver(
...
+/// - Un-authenticated: anyone can request these routes. Used by fedimint
+///   clients.
+fn v1_routes(gateway: Gateway) -> Router {
+    // Public routes on gateway webserver
+    let public_routes = Router::new()
+        .route("/pay_invoice", post(pay_invoice))
+        .route("/id", get(get_gateway_id));
+
+    // Authenticated, public routes used for gateway administration
+    let always_admin_routes = Router::new()
```
okjodom, line 128: name suggestion: `always_authenticated_routes`

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -44,49 +39,118 @@ pub async fn run_webserver(
...
+        .route("/withdraw", post(withdraw))
+        .route("/connect-fed", post(connect_fed))
+        .route("/leave-fed", post(leave_fed))
+        .route("/backup", post(backup))
+        .route("/restore", post(restore))
+        .layer(middleware::from_fn(admin_authentication_middleware));
+
+    // Routes that are un-authenticated before gateway configuration, then become
+    // authenticated after a password has been set.
+    let admin_after_config_routes = Router::new()
```
okjodom, line 140: name suggestion: `authenticated_after_config_routes`

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -44,49 +39,118 @@ pub async fn run_webserver(
...
+    next: Next,
+) -> Result<impl IntoResponse, StatusCode> {
+    // If the gateway's config has not been set, allow the request to continue, so
+    // that the gateway can be configured
+    let gateway_config = gateway.gateway_config.read().await.clone();
+    if gateway_config.is_none() {
+        return Ok(next.run(request).await);
+    }
+
+    // Otherwise, validate that the Bearer token matches the gateway's password
```
okjodom, line 101: let's extract shared authentication logic and reuse across both middleware fns
```rs
///  Validate that the Bearer token matches the gateway's password
async fn authenticate(
    gateway_config: GatewayConfiguration,
    request: Request,
    next: Next,
) -> Result<axum::response::Response, StatusCode> {
    let gateway_password = gateway_config.password;
    let token = extract_bearer_token(&request)?;
    if gateway_password == token {
        return Ok(next.run(request).await);
    }

    Err(StatusCode::UNAUTHORIZED)
}
```

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -44,49 +39,118 @@ pub async fn run_webserver(
...
+    }
+
+    Err(StatusCode::UNAUTHORIZED)
+}
+
+/// Middleware to authenticate an incoming request. Routes that are
+/// authenticated with this middleware are un-authenticated if the gateway has
+/// not yet been configured. After the gateway is configured, this middleware
+/// enforces that a Bearer token must be supplied in the Authorization header.
+async fn admin_after_config_authentication_middleware(
```
okjodom, line 89: name suggestion: `auth_after_config_middleware_fn`

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -44,49 +39,118 @@ pub async fn run_webserver(
...
+        return Ok(token);
+    }
+
+    Err(StatusCode::UNAUTHORIZED)
+}
+
+/// Middleware to authenticate an incoming request. Routes that are
+/// authenticated with this middleware always require a Bearer token to be
+/// supplied in the Authorization header.
+async fn admin_authentication_middleware(
```
okjodom, line 69: name suggestion: `auth_middleware_fn`

gateway/ln-gateway/src/rpc/rpc_server.rs
m1sterc001guy: Made this change but dropped the `_fn` part, should be obvious that it is a function :)

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -44,49 +39,120 @@ pub async fn run_webserver(
...
+}
+
+/// Validate that the Bearer token matches the gateway's password
+async fn authenticate(
+    gateway_password: String,
+    request: Request,
+    next: Next,
+) -> Result<axum::response::Response, StatusCode> {
+    let token = extract_bearer_token(&request)?;
+    if gateway_password == token {
```
dpc, line 110: BTW. Do we need to use some constant time password comparison here, to prevent timing attacks?

Alternatively we should hash both sides (hash the set password in the first place and not store it anywhere in the plaintext). Can be a follow-up.

If so, can you think of any other places like this and create an issue?

dpc (approved):
LGTM, but we should really be hashing passwords.

gateway/ln-gateway/src/rpc/rpc_server.rs
m1sterc001guy: Made issue. I'm not aware of other spots where this is a problem. https://github.com/fedimint/fedimint/issues/4702

Main challenge will be doing it in a backwards compatible way.

okjodom (approved):
thanks for the auth improvement

PR #4643, opened 2024-03-20
refactor: how fedimintd exposes module configuration

dpc:
As things are currently the way modules and module instances are attached to `Fedimintd` in custom builds is confusing and error-prone.

Notable changes:

* Make more stuff in `Fedimintd` private
* Stop talking about config params in public api. Now you add either module kind of module instance with `with_module_kind` or `with_module_instance`, where module instance takes params as config. Seems more logical.

Making default module params private made me duplicate this setup in devimint, but I think it's OK, and in the future which modules devimint runs and which devimintd does by default will probably diverge more anyway, as we need to test more federation configurations.


maan2003 (commented):
going to try this in fedi before approving

dpc:
rebased

fedimintd/src/fedimintd.rs
```diff
@@ -119,25 +127,27 @@ fn parse_map(s: &str) -> anyhow::Result<BTreeMap<String, String>> {
...
 /// ```
 pub struct Fedimintd {
-    pub server_gens: ServerModuleInitRegistry,
-    pub server_gen_params: ServerModuleConfigGenParamsRegistry,
-    pub version_hash: String,
+    server_gens: ServerModuleInitRegistry,
+    server_gen_params: ServerModuleConfigGenParamsRegistry,
+    version_hash: String,
+    opts: ServerOpts,
+    bitcoind_rpc: BitcoinRpcConfig,
```
elsirion, line 150: `bitcoind_rpc` is only used by modules, why have it at the top level?

fedimintd/src/fedimintd.rs
```diff
@@ -160,66 +170,116 @@ impl Fedimintd {
...
+            .with_module_instance(
+                MintInit::kind(),
+                MintGenParams {
+                    local: Default::default(),
+                    consensus: MintGenParamsConsensus::new(
+                        2,
+                        fedimint_mint_server::common::config::FeeConsensus::default(),
+                    ),
+                },
+            )
```
elsirion, line 248: Longer term I think I'd rather leave it to the setup tool (admin UI, CLI, …) to determine which modules to instantiate "by default". These parameters are only defaults anyway afaik.

elsirion (approved):

elsirion:
needs rebase :(

dpc:
Landed as a part of https://github.com/fedimint/fedimint/pull/4724 already

PR #4652, opened 2024-03-21
feat: add gateway id in lightning operation meta

maan2003:
I want to adjust priority of gateway when a payment fails or succeeds.

Imagine you have two vetted gateways, you can switch to second one when the
payment fails for first gateway.

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -232,6 +232,7 @@ pub struct LightningOperationMetaPay {
     pub change: Vec<OutPoint>,
     pub is_internal_payment: bool,
     pub contract_id: ContractId,
+    pub gateway_id: Option<secp256k1::PublicKey>,
 }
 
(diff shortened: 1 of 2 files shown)


dpc (approved):
I'm not deep into LN, but looks reasonable.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -232,6 +232,7 @@ pub struct LightningOperationMetaPay {
     pub change: Vec<OutPoint>,
     pub is_internal_payment: bool,
     pub contract_id: ContractId,
+    pub gateway_id: Option<secp256k1::PublicKey>,
```
elsirion, line 235: TIL: [you don't need `#[serde(default)]` annotation for `Option<T>` fields :exploding_head: ](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9739232131409ded8e1d6e6f2199ea88)

elsirion (approved):

fedimint-backports:
- #4654

PR #4653, opened 2024-03-21
feat(nix): expose fedimintd NixOS module

dpc:
Note: currently it does not include the UI.

Note: It will be easier to test and improve after it lands.

### DIFF
--- a/nix/modules/fedimintd.nix
+++ b/nix/modules/fedimintd.nix
@@ -1,0 +1,233 @@
+{ config, lib, pkgs, ... }:
+with lib;
+
+let
+  eachFedimintd = filterAttrs (fedimintdName: cfg: cfg.enable) config.services.fedimintd;
+  fedimintdOpts = { config, lib, name, ... }: {
+
+    options = {
+
+      enable = mkEnableOption (lib.mdDoc "fedimint");
+
+      extraEnvironment = mkOption {
+        type = types.attrsOf types.str;
+        description = lib.mdDoc "Extra Environment variables to pass to the fedimintd.";
+        default = { };
+        example = { RUST_BACKTRACE = "1"; };
+      };
+
+      package = mkOption {
+        type = types.nullOr types.package;
+        default = pkgs.fedimintd or null;
+        defaultText = lib.literalExpression "pkgs.fedimint (after available)";
+        description = lib.mdDoc ''
+          Package of the fedimintd to use.
+        '';
+      };
+
+      p2p = {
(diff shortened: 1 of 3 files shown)


nix/modules/fedimintd.nix
```diff
@@ -0,0 +1,233 @@
...
+      extraEnvironment = mkOption {
+        type = types.attrsOf types.str;
+        description = lib.mdDoc "Extra Environment variables to pass to the fedimintd.";
+        default = { };
+        example = { RUST_BACKTRACE = "1"; };
+      };
+
+      package = mkOption {
+        type = types.nullOr types.package;
+        default = pkgs.fedimintd or null;
```
elsirion, line 21: Should we default to the falke's own package definition for now? This reads to me as if we were using a non-existent nixpkgs version? Moreso, it's a bit confusing if the module from a specific branch/version of the repo will depend on a random other version of Fedimint.

nix/modules/fedimintd.nix
```diff
@@ -0,0 +1,233 @@
...
+        };
+        rpc = {
+          address = mkOption {
+            type = types.str;
+            default = "http://[::1]:38332";
+            example = "signet";
+            description = lib.mdDoc "Bitcoin node (bitcoind/electrum/esplora) address to connect to";
+          };
+          kind = mkOption {
+            type = types.str;
```
elsirion, line 89: We could make this an enum later on for better documentation of the available options.

elsirion (approved):
:heart_eyes:

nix/modules/fedimintd.nix
maan2003: we should have an overlay that add `fedimintd` to pkgs

nix/modules/fedimintd.nix
```diff
@@ -0,0 +1,233 @@
...
+              ];
+              StateDirectory = "fedimintd";
+              StateDirectoryMode = "0700";
+              ExecStart = "${cfg.package}/bin/fedimintd";
+              Restart = "always";
+              RestartSec = 10;
+              StartLimitBurst = 5;
+              UMask = "077";
+              LimitNOFILE = "100000";
+            };
```
maan2003, line 228: did you try running this, I wouldn't be surprised something got broken by systemd sandbox

maan2003 (commented):
looks good code wise, but nix is not easy to review without running it.

maan2003:
we should have nixos vm test for this

PR #4665, opened 2024-03-21
fix: version in nix derivations and bundles

dpc:


Cargo.toml
```diff
@@ -53,6 +53,10 @@ members = [
 ]
 resolver = "2"
 
+[workspace.package]
```
dpc, line 56: @elsirion It's unclear to me which is the right place to keep an overal project version, `[workspace.metadata]` or `[workspace.package]`. Seems like `metadata` is for external tools, while `package` for inheriting via `version = { workspace = true }`.

Anyway - when releasing please pay attention to this one, as crane will use it for setting all Nix/deb/rpm versions.

elsirion (approved):

justinmoon (approved):

PR #4669, opened 2024-03-21
fix: hourly run

dpc:
Fix for what caused #4657 

Tested in a different draft PRs (#4668).  We can land, then I should be able manually trigger some runs, see if they look OK, and then we can revert #4657


bradleystachurski (approved):
Solid plan 👍

Kodylow (approved):
LGTM

PR #4672, opened 2024-03-22
chore: cleanup devimint logs

maan2003:
remove jsonrpc error level logs during startup

### DIFF
--- a/devimint/src/cli.rs
+++ b/devimint/src/cli.rs
@@ -113,6 +113,8 @@ pub async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
 
     fedimint_logging::TracingSetup::default()
         .with_file(Some(log_file))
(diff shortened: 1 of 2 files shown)


devimint/src/cli.rs
```diff
@@ -113,6 +113,8 @@ pub async fn setup(arg: CommonArgs) -> Result<(ProcessManager, TaskGroup)> {
 
     fedimint_logging::TracingSetup::default()
         .with_file(Some(log_file))
+        // jsonrpsee is expected to fail during startup
```
dpc, line 116: Should we just do it in `fedimint_logging`?

dpc (approved):

elsirion (approved):

devimint/src/cli.rs
dpc: @maan2003

PR #4676, opened 2024-03-22
chore: use workspace dependencies instead of repeating them

### LINKED ISSUE #4517: Refactor: use workspace dependencies way more
We keep repeating the same deps, including our local ones with all the `versions`, `path`, `features` etc, where we could do `{ workspace = true }`.

harsh-ps-2003:
Fixes #4517 

Refactoring the `cargo.toml` files to use workspace dependencies as much as possible.

On running `cargo build`, no errors are seen!


elsirion (approved):
Nice, even deduplicated two dependencies!

dpc (approved):

dpc:
BTW. I think we should be able to do it for internal dependencies, to avoid having to type `path`s over and over.

PR #4678, opened 2024-03-22
feat: add commands to check gatewayd and gateway-cli versions

tvolk131:
I've added the same code in #4514 but @kernelkind is blocked on this so I pulled this functionality into a separate PR

### DIFF
--- a/devimint/src/util.rs
+++ b/devimint/src/util.rs
@@ -596,6 +596,14 @@ impl Gatewayd {
             &[GATEWAYD_FALLBACK],
         ))
     }
+
+    /// Returns the gatewayd version from clap or default min version
+    pub async fn version_or_default() -> Version {
+        match cmd!(Gatewayd, "--version").out_string().await {
+            Ok(version) => parse_clap_version(&version),
+            Err(_) => DEFAULT_VERSION,
+        }
+    }
 }
 
 pub struct FedimintCli;
@@ -639,6 +647,14 @@ impl GatewayCli {
                 .map(String::as_str)
                 .collect::<Vec<_>>(),
         ))
+    }
+
+    /// Returns the gateway-cli version from clap or default min version
+    pub async fn version_or_default() -> Version {
+        match cmd!(GatewayCli, "--version").out_string().await {
+            Ok(version) => parse_clap_version(&version),
+            Err(_) => DEFAULT_VERSION,
+        }
     }
 }
 


bradleystachurski (approved):

fedimint-backports:
- #4694

devimint/src/util.rs
```diff
@@ -640,6 +648,14 @@ impl GatewayCli {
                 .collect::<Vec<_>>(),
         ))
     }
+
+    /// Returns the gateway-cli version from clap or default min version
+    pub async fn version_or_default() -> Version {
+        match cmd!(GatewayCli, "--version").out_string().await {
```
m1sterc001guy, line 654: It looks like `gateway-cli`'s version command is `--version-hash`. Will this just always return `DEFAULT_VERSION`?

devimint/src/util.rs
```diff
@@ -596,6 +596,14 @@ impl Gatewayd {
             &[GATEWAYD_FALLBACK],
         ))
     }
+
+    /// Returns the gatewayd version from clap or default min version
+    pub async fn version_or_default() -> Version {
+        match cmd!(Gatewayd, "--version").out_string().await {
```
m1sterc001guy, line 602: I don't think we have `gatewayd --version` currently....

m1sterc001guy: `version-hash`? 

`handle_version_hash_command(fedimint_build_code_version_env!());`

devimint/src/util.rs
tvolk131: The gateway CLI supports `gateway-cli version-hash` _and_ `gateway-cli --version` so the method above should work

devimint/src/util.rs
tvolk131: Ahh yes you're correct. Rather than changing `version_or_default()`, would it make sense to add a `--version` command to gatewayd?

m1sterc001guy: Yes we should add that

devimint/src/util.rs
bradleystachurski: Good question! Clap gives us the version flag with

https://github.com/fedimint/fedimint/blob/4aa64957d9faf0fe204f7d36cd3f44a0d4151284/gateway/cli/src/main.rs#L15

devimint/src/util.rs
bradleystachurski: @m1sterc001guy Good catch!

We're not currently parsing gatewayd versions for back-compat, but to be safe I added parsing to map version hash to version for the old binaries that won't have `--version`.

https://github.com/fedimint/fedimint/pull/4701

PR #4679, opened 2024-03-22
gateway can charge fees per federation

fparisotto:
Addresses this issue https://github.com/fedimint/fedimint/issues/3961

This PR encompass the following changes:
- ~~updates the `connect-fed` sub command in `gateway-cli`, by adding a routing fee parameter~~
- update the `set-configuration` sub command in `gateway-cli`, by adding routing fees by federation id
- refactor the update configuration command to write the configuration data of gateway and federations in the same transaction (this is a little bit out of scope, if does not make sense I can revert it)

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -1,3 +1,4 @@
+use anyhow::bail;
 use bitcoin::address::NetworkUnchecked;
 use bitcoin::Address;
 use clap::{CommandFactory, Parser, Subcommand};
@@ -11,7 +12,8 @@ use fedimint_logging::TracingSetup;
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
 use ln_gateway::rpc::{
     BackupPayload, BalancePayload, ConfigPayload, ConnectFedPayload, DepositAddressPayload,
-    LeaveFedPayload, RestorePayload, SetConfigurationPayload, WithdrawPayload, V1_API_ENDPOINT,
+    FederationRoutingFees, LeaveFedPayload, RestorePayload, SetConfigurationPayload,
+    WithdrawPayload, V1_API_ENDPOINT,
 };
 use serde::Serialize;
 
@@ -90,12 +92,46 @@ pub enum Commands {
         #[clap(long)]
         num_route_hints: Option<u32>,
 
+        /// Default routing fee for all new federations. Setting it won't affect
+        /// existing federations
         #[clap(long)]
         routing_fees: Option<String>,
 
         #[clap(long)]
         network: Option<bitcoin::Network>,
+
+        /// Format federation id,base msat,proportional to millionths part. Any
+        /// other federations not given here will keep their current fees.
+        #[clap(long)]
+        per_federation_routing_fees: Option<Vec<PerFederationRoutingFees>>,
     },
+}
+
+#[derive(Clone)]
+pub struct PerFederationRoutingFees {
+    pub federation_id: FederationId,
+    pub routing_fees: FederationRoutingFees,
+}
+
+impl std::str::FromStr for PerFederationRoutingFees {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Some((federation_id, rounting_fees)) = s.split_once(',') {
+            Ok(PerFederationRoutingFees {
+                federation_id: federation_id.parse()?,
+                routing_fees: rounting_fees.parse()?,
+            })
+        } else {
+            bail!("Wrong format, please provide: <federation id>,<base msat>,<proportional to millionths part>");
+        }
+    }
+}
+
+impl From<PerFederationRoutingFees> for (FederationId, FederationRoutingFees) {
+    fn from(val: PerFederationRoutingFees) -> Self {
+        (val.federation_id, val.routing_fees)
+    }
 }
 
 #[tokio::main]
@@ -188,13 +224,17 @@ async fn main() -> anyhow::Result<()> {
             num_route_hints,
             routing_fees,
             network,
+            per_federation_routing_fees,
         } => {
+            let per_federation_routing_fees = per_federation_routing_fees
+                .map(|input| input.into_iter().map(Into::into).collect());
             client()
                 .set_configuration(SetConfigurationPayload {
                     password,
                     num_route_hints,
-                    routing_fees,
                     network: network.map(bitcoin30_to_bitcoin29_network),
+                    routing_fees,
+                    per_federation_routing_fees,
                 })
                 .await?;
         }
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -525,6 +525,7 @@ impl Gateway {
                                         network: Some(bitcoin30_to_bitcoin29_network(lightning_network)),
                                         num_route_hints: None,
                                         routing_fees: None,
+                                        per_federation_routing_fees: None,
                                     }).await.expect("Failed to set gateway configuration");
                                     continue;
                                 }
@@ -913,8 +914,6 @@ impl Gateway {
                 invite_code,
                 mint_channel_id,
                 timelock_delta: 10,
-                // TODO: Today we use a global routing fees setting. When the gateway supports
-                // per-federation routing fees, this value will need to be updated.
                 fees: gateway_config.routing_fees,
             };
 
@@ -936,6 +935,7 @@ impl Gateway {
                 balance_msat: client.get_balance().await,
                 config: client.get_config().clone(),
                 channel_id: Some(mint_channel_id),
+                routing_fees: Some(gateway_config.routing_fees.into()),
             };
 
             self.check_federation_network(
@@ -1037,6 +1037,7 @@ impl Gateway {
             network,
             num_route_hints,
             routing_fees,
+            per_federation_routing_fees,
         }: SetConfigurationPayload,
     ) -> Result<()> {
         let gw_state = self.state.read().await.clone();
@@ -1083,10 +1084,8 @@ impl Gateway {
                 prev_config.num_route_hints = num_route_hints;
             }
 
-            // TODO: Today, the gateway only supports a single routing fee configuration.
-            // We will eventually support per-federation routing fees. This configuration
-            // will be deprecated when per-federation routing fees are
-            // supported.
+            // Using this routing fee config as a default for all federation that has none
+            // routing fees specified.
             if let Some(fees_str) = routing_fees.clone() {
                 let routing_fees = GatewayFee::from_str(fees_str.as_str())?.0;
                 prev_config.routing_fees = routing_fees;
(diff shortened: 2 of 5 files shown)


fparisotto:
I'll add tests in my next commits. Since this is my first contribution, I need some help to check if I'm in the right track or missing something, any help is more than welcome :)

gateway/cli/src/main.rs
```diff
@@ -91,9 +95,61 @@ pub enum Commands {
...
+}
+
+impl std::str::FromStr for FederationIdAndRoutingFeesParameter {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut parts = s.split(',');
+        let federation_id = parts
+            .next()
+            .ok_or_else(|| anyhow::anyhow!("missing federation id"))?
```
douglaz, line 119:
```suggestion
            .context("missing federation id")?
```

gateway/cli/src/main.rs
```diff
@@ -184,13 +246,17 @@ async fn main() -> anyhow::Result<()> {
             num_route_hints,
             routing_fees,
             network,
+            per_federation_routing_fees,
         } => {
+            let per_federation_routing_fees = per_federation_routing_fees
+                .map(|input| input.into_iter().map(|e| e.into()).collect());
```
douglaz, line 252:
```suggestion
                .map(|input| input.into_iter().map(Into::into).collect());
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -918,13 +920,16 @@ impl Gateway {
                     ))?;
             *max_used_scid = mint_channel_id;
 
+            let fees = payload
+                .routing_fees
+                .map(|e| e.into())
```
douglaz, line 925: you know

gateway/cli/src/main.rs
```diff
@@ -91,9 +96,54 @@ pub enum Commands {
 
         #[clap(long)]
         network: Option<bitcoin::Network>,
+
+        /// Format <federation id>,<base msat>,<proportional to millionths part>
+        #[clap(long)]
+        per_federation_routing_fees: Option<Vec<FederationIdAndRoutingFeesParameter>>,
```
m1sterc001guy, line 102: Do we even need to add this? Just for simplicity we could just use the default federation routing fees when connecting a new federation

gateway/cli/src/main.rs
```diff
@@ -91,9 +96,54 @@ pub enum Commands {
...
     },
 }
 
+#[derive(Clone)]
+pub struct FederationIdAndRoutingFeesParameter {
```
m1sterc001guy, line 107: See above comment....but if we elect to keep it I think `PerFederationRoutingFees` is a better name

gateway/ln-gateway/src/lib.rs
```diff
@@ -1082,10 +1089,8 @@ impl Gateway {
                 prev_config.num_route_hints = num_route_hints;
             }
 
-            // TODO: Today, the gateway only supports a single routing fee configuration.
-            // We will eventually support per-federation routing fees. This configuration
-            // will be deprecated when per-federation routing fees are
-            // supported.
+            // Using this routing fee config as a default for all federation that has none
+            // routing fees specified.
```
m1sterc001guy, line 1093: Clever, good backwards compatible way to make this work. Maybe we could change the name of `routing_fees` to `default_routing_fees` so its not confusing in the code which value is which

gateway/ln-gateway/src/lib.rs
```diff
@@ -1108,48 +1113,51 @@ impl Gateway {
...
-
-            if let Err(e) = Self::register_all_federations(self, gateway_config).await {
-                warn!("{e:?}")
-            };
+        let configs: Vec<(FederationIdKey, FederationConfig)> = dbtx
+            .find_by_prefix(&FederationIdKeyPrefix)
+            .await
+            .collect::<Vec<_>>()
+            .await;
+        for (id_key, mut fed_config) in configs {
```
m1sterc001guy, line 1154: You shouldn't need to iterate through all the federations here. You already have the `federation_id`, you can just directly update that federation's config

m1sterc001guy: Now that we don't need to iterate through all federations, this function is likely not necessary at all and the logic can be inlined in `handle_set_configuration_msg`

gateway/ln-gateway/src/lib.rs
```diff
@@ -1108,48 +1113,51 @@ impl Gateway {
             }
         };
 
+        let per_federation_routing_fees: HashMap<FederationId, RoutingFees> =
+            per_federation_routing_fees
+                .unwrap_or_default()
```
m1sterc001guy, line 1118: I don't think we want to use `unwrap_or_default` here. Ideally we should only update the federation's routing fees is routing fees were supplied in `SetPayloadConfiguration` (i.e `per_federation_routing_fees.is_some()`. If I'm reading this correctly, if a gateway operator only wants to update their route hints or another unrelated value, the fees are going to be overwritten with whatever the default routing fees are.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1108,48 +1113,51 @@ impl Gateway {
...
+            &mut dbtx,
+            &gateway_config,
+            &per_federation_routing_fees,
+        )
+        .await?;
         dbtx.commit_tx().await;
 
-        self.update_federation_routing_fees(routing_fees, &gateway_config)
-            .await?;
+        if let Err(e) = Self::register_all_federations(self, &gateway_config).await {
```
m1sterc001guy, line 1133: We no longer need to register all federations! One of the good parts of this change. We should only re-register the with the federation who's routing fees have changed.

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -86,10 +91,46 @@ pub struct GatewayFedConfig {
     pub federations: BTreeMap<FederationId, JsonClientConfig>,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct RoutingFeesWrapper {
```
m1sterc001guy, line 95: I kind of hate the name `Wrapper` :) Maybe just `FederationRoutingFees`?

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -86,10 +91,46 @@ pub struct GatewayFedConfig {
...
+            base_msat: val.base_msat,
+            proportional_millionths: val.proportional_millionths,
+        }
+    }
+}
+
+impl FromStr for RoutingFeesWrapper {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
m1sterc001guy, line 112: I think I've seen this logic before, lets make a separate function for it

fedimint-testing/src/gateway.rs
```diff
@@ -84,9 +84,32 @@ impl GatewayTest {
...
+        rpc.connect_federation(ConnectFedPayload {
+            invite_code,
+            routing_fees: None,
+        })
+        .await
+        .unwrap()
+    }
+
+    /// Connects to a new federation passing a routing fees, and stores the info
+    pub async fn connect_fed_with_routing_fees(
```
m1sterc001guy, line 96: I don't think this function is necessary, lets just keep `connect_fed` as is and allow them to change the fees later

fedimint-testing/src/gateway.rs
```diff
@@ -84,9 +84,32 @@ impl GatewayTest {
...
+            .get_rpc()
             .await
-            .unwrap()
+            .with_password(Some(DEFAULT_GATEWAY_PASSWORD.to_string()));
+        rpc.connect_federation(ConnectFedPayload {
+            invite_code,
+            routing_fees: Some(routing_fees),
+        })
+        .await
+        .unwrap()
```
m1sterc001guy, line 112: nit: use expect instead of unwrap

m1sterc001guy (commented):
Looking good so far, great start! A few high level comments:

Wherever we can, let's rename the existing routing fees parameter that applies to all federations to `default_routing_fees`. The one place we likely can't change it is in the CLI and environment variable, since that would break backwards compatibility. 

A test as a next step would be great! You can modify `test_gateway_configuration` to add this functionality. We should verify that when joining a new federation, the routing fees are set to the default and that we can change the routing fees for individual federations. Note that this test is changing slightly in this outstanding PR: https://github.com/fedimint/fedimint/pull/4642

gateway/cli/src/main.rs
fparisotto: Good one, less moving parts is better, I'll remove it

okjodom: let's remove. I think it's  cheap to connect federation and then configure fees and any other params once connected. This is also intuitive for non CLI ux people will build for gateway

gateway/ln-gateway/src/lib.rs
douglaz: I think there is some confusion about the semantics here. This commands currently sets the fees for all federations, so if some federation is missing, it will go to default. I think this is good because we need some way to set federations to default. But of course if we need to change just one federation, perhaps we're gonna need another command.

gateway/ln-gateway/src/lib.rs
douglaz: See comment above/below. We're are potentially changing all the federations fees to default.

gateway/ln-gateway/src/lib.rs
fparisotto: Good point @douglaz , I tried to preserve this behavior, so in my point of view, set-configuration will overwrite every state in federation config (at least this was the current behavior, correct me if I'm wrong). Maybe if we want to "patch" a fee for a existing federation, we need another command.

m1sterc001guy: This was not the previous behavior. The previous behavior only modified the configs that were `Some` in the `SetConfigurationPayload`. If the field was not set, it would use the previous set configuration. We should preserve these semantics imo

m1sterc001guy: I think we should leave it as is and the new default will be applicable only for new connected federations

gateway/cli/src/main.rs
```diff
@@ -90,14 +92,46 @@ pub enum Commands {
         #[clap(long)]
         num_route_hints: Option<u32>,
 
+        /// Default routing fee for all federations
```
douglaz, line 95:
```suggestion
        /// Default routing fee for all new federations. Setting it won't affect existing federations
```

gateway/cli/src/main.rs
```diff
@@ -90,14 +92,46 @@ pub enum Commands {
...
         #[clap(long)]
         routing_fees: Option<String>,
 
         #[clap(long)]
         network: Option<bitcoin::Network>,
+
+        /// Format federation id,base msat,proportional to millionths part
```
douglaz, line 102:
```suggestion
        /// Format federation id,base msat,proportional to millionths part. Any other federations not given here will keep their current fees.
```

gateway/ln-gateway/src/lib.rs
```diff
@@ -1090,52 +1091,98 @@ impl Gateway {
...
-            for (id_key, mut fed_config) in configs {
-                fed_config.fees = routing_fees;
-                dbtx.insert_entry(&id_key, &fed_config).await;
+        for (federation_id, routing_fees) in per_federation_routing_fees.iter() {
+            let federation_key = FederationIdKey { id: *federation_id };
+            if let Some(mut federation_config) = dbtx.get_value(&federation_key).await {
+                federation_config.fees = routing_fees.clone().into();
+                dbtx.insert_entry(&federation_key, &federation_config).await;
+                if let Err(e) = self
+                    .register_federation(gateway_config, federation_id, &federation_config)
```
douglaz, line 1127: Now we are re-registering only the federations with new fees. But what about other parameters like `num_route_hints`, will them affect other federations? Should we register again all federations? What do you think @m1sterc001guy ?

elsirion:
The linter is complaining about formatting. Running `just format` should fix it.

elsirion (dismissed):
The gateway code is notoriously hard to reason about, but looks correct and tests run, so let's give it a try :)

elsirion (approved):
As stated on the dev call, someone with more gateway context please do the second review :)

gateway/ln-gateway/src/lib.rs
```diff
@@ -1456,11 +1503,19 @@ impl Gateway {
                 }
             });
 
+        let mut dbtx = self.gateway_db.begin_transaction().await.into_nc();
```
m1sterc001guy, line 1506: nit: you can use `begin_transaction_nc` instead

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Yeah, agreed with @douglaz. For other parameters like `num_route_hints`, we should also re-register the gateway with all the federations. 

I think updating the `FederationConfig` and the re-register function can be split apart. Re-register can take a vector of `FederationId`, so we can call it when either `num_route_hints` is updated or when the federation fees are updated.

Eventually with lightning NG we'll be able to get rid of a lot of this complexity.

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1523,13 +1642,38 @@ async fn test_gateway_executes_swaps_between_connected_federations() -> anyhow::
...
     )
     .await
 }
 
+fn routing_fees_in_msats(routing_fees: &FederationRoutingFees, amount: &Amount) -> u64 {
+    ((amount.msats * routing_fees.proportional_millionths as u64) / 1_000_000)
+        + routing_fees.base_msat as u64
+}
+
+async fn reconnect_federation(rpc: &GatewayRpcClient, fed: &FederationTest) -> anyhow::Result<()> {
```
m1sterc001guy, line 1661: nit: I don't think this needs to return a `Result`

m1sterc001guy:
Looking very good, I think we just need to fix the re-registration issue when the `num_route_hints` are updated

gateway/ln-gateway/src/lib.rs
fparisotto: Thanks for catching that! I did the fix as suggested in my last commit.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1195,26 +1175,48 @@ impl Gateway {
...
     }
 
+    /// Iterates through all of the federation configurations and registers the
+    /// gateway with each federation.
+    async fn register_all_federations(
+        gateway: &Gateway,
+        gateway_config: &GatewayConfiguration,
+    ) -> Result<()> {
+        let mut all_federations: Vec<(FederationId, FederationConfig)> = Vec::new();
+        let mut dbtx = gateway.gateway_db.begin_transaction_nc().await;
```
m1sterc001guy, line 1205: A more compact way of doing the same thing is something like this:

```
dbtx.find_by_prefix(&FederationIdPrefix).await.collect::<(FederationId, FederationConfig)>().await
```

You don't need to lookup each config individually. Then you can delete this function since its only 1 line to collect all of the federations, and just call `register_federations` instead

gateway/ln-gateway/src/lib.rs
```diff
@@ -1108,77 +1107,58 @@ impl Gateway {
...
+                } else {
+                    warn!("Given federation {federation_id} not found for updating routing fees");
+                }
+            }
+        }
+
+        // If 'num_route_hints' is provided, all federations must be re-registered.
+        // Otherwise, only those affected by the new fees need to be re-registered.
+        if num_route_hints.is_some() {
+            Self::register_all_federations(self, &new_gateway_config).await?;
```
m1sterc001guy, line 1130: Would prefer to delete this function (see below)

m1sterc001guy:
Would be good to squash some of the commits together

m1sterc001guy (approved):

justinmoon (dismissed):

douglaz (approved):
LGTM

fedimint-backports:

```bash
git worktree add -d .worktree/backport-4679-to-releases/v0.3 origin/releases/v0.3
cd .worktree/backport-4679-to-releases/v0.3
git switch --create backport-4679-to-releases/v0.3
git cherry-pick -x 50a5407756286fe8ad3df62f0246cf6ca6d22c19
```

PR #4681, opened 2024-03-22
feat: Make claim_funded_incoming_contract public

benthecarman:
Missed this in #3820 but it'd be nice to have this publicly available so we don't need to scan everytime we try to claim. We can instead give the user the `contract_id`s they need to claim


benthecarman:
> Couldn't you call `scan_receive_for_user` instead? That's probably poorly named in retrospect, that function doesn't actually scan

Ah yeah, name confused me, maybe worth naming before 0.3.0 final is cut

elsirion (approved):

fedimint-backports:
- #4683

Issue #4561, opened 2024-03-14
Upgrade all (most) dependencies before releasing 0.3


maan2003:
deps left:
- [x] bech32
- [x] bitcoin
- [x] bitcoincore-rpc
- [x] bitcoin_hashes
- [x] bls12_381
- [x] cln-plugin
- [x] electrum-client
- [x] esplora-client
- [x] ff
- [x] gloo-net
- [x] group
- [x] lightning
- [x] lightning-invoice
- [x] miniscript
- [x] rcgen
- [x] secp256k1
- [x] secp256k1-zkp

~tokio-rustls~ edit: this was broken off into a separate issue due to its complexity: #5793

maan2003:
https://github.com/fedimint/fedimint/pull/4557 upgrade most of deps

justinmoon:
Is this done?

maan2003:
nope, all deps listed are still not upgraded

elsirion:
Unless there's a CVE being on the latest version isn't important for 0.3 imo. It's generally a good practice to avoid huge version jumps later on, but that can also happen on master later on. Only drawback is that any change to master will make backporting harder, but that's always the case no matter the change.

justinmoon:
dev call: let's split the alephbft ones into separate issues

tvolk131:
Assigning to myself since I already did most of these and can wrap up the rest soon

tvolk131:
Finished!

Issue #2356, opened 2023-04-29
Update rust-bitcoin


jkitman:
It's one of our main crypto libs we are a few versions behind.

edit (justinmoon): edited this to be "update rust-bitcoin" from "update secp256k1"

elsirion:
This will have to happen in lock step with rust-bitcoin, miniscript, all the BTC clients using rust-bitcoin structs (bitcoind, esplora, electrum) and possibly some other crates. We need to make sure that all required ones support the same secp version.

elsirion:
This will likely be part of #2667.

m1sterc001guy:
This is also blocked on the `lightning` and `lightning-invoice` crates. They do not support `bitcoin 0.30.1` or `secp256k1 0.27.0` yet.

benthecarman:
Latest `LDK` release is now on `0.30`

https://github.com/lightningdevkit/rust-lightning/releases/tag/v0.0.119

EthnTuttle:
> Latest `LDK` release is now on `0.30`
> 
> https://github.com/lightningdevkit/rust-lightning/releases/tag/v0.0.119

https://github.com/lightningdevkit/rust-lightning/pull/2799

justinmoon:
> [lightningdevkit/rust-lightning#2799](https://github.com/lightningdevkit/rust-lightning/pull/2799)

If this is the only regression, I think we can still integrate because the only part of LDK that currently runs in WASM is `lightning_invoice` and the regression this PR fixes is elsewhere. 

I propose we update on master but don't backport it.

tvolk131:
If no one's currently working on this, I can give it a shot!

m1sterc001guy:
I have a broken branch attempting to do this here: https://github.com/m1sterc001guy/fedimint/tree/update_rust_bitcoin

These updates touch a lot of files and some structs in `rust-bitcoin` removed their `Encodable` and `Decodable` traits.

justinmoon:
Relevant TODO https://github.com/fedimint/fedimint/blob/47eb296afc274c9526163993141edc42f8c8c9d1/fedimint-core/src/session_outcome.rs#L70-L78

tvolk131:
Unless anyone knows of more work to do for this issue, I think we can mark this as done!

m1sterc001guy:
@tvolk131 is going to make a separate issue for removing bitcoin 29 from the codebase.

benthecarman:
congrats!

PR #4686, opened 2024-03-22
chore: upgrade lightning and lightning-invoice deps

tvolk131:
Part of #4561 and #2356

### DIFF
--- a/fedimint-core/src/bitcoin_migration.rs
+++ b/fedimint-core/src/bitcoin_migration.rs
@@ -286,6 +286,34 @@ pub fn bitcoin30_to_bitcoin29_address(address: bitcoin30::Address) -> bitcoin::A
         .expect("Failed to convert bitcoin v30 address to bitcoin v29 address")
 }
 
+pub fn bitcoin29_to_bitcoin30_recoverable_signature(
+    signature: &bitcoin::secp256k1::ecdsa::RecoverableSignature,
+) -> bitcoin30::secp256k1::ecdsa::RecoverableSignature {
+    let (recid, data) = signature.serialize_compact();
+    bitcoin30::secp256k1::ecdsa::RecoverableSignature::from_compact(
+        &data,
+        bitcoin30::secp256k1::ecdsa::RecoveryId::from_i32(recid.to_i32())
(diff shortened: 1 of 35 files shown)


fedimint-core/src/bitcoin_migration.rs
```diff
@@ -155,6 +183,12 @@ pub fn bitcoin29_to_bitcoin30_sha256_hash(
     bitcoin30::hashes::sha256::Hash::from_byte_array(hash.into_inner())
 }
 
+pub fn bitcoin30_to_bitcoin29_sha256_hash(
+    hash: bitcoin30::hashes::sha256::Hash,
+) -> bitcoin::hashes::sha256::Hash {
+    bitcoin::hashes::sha256::Hash::from_str(&hash.to_string()).unwrap()
```
elsirion, line 189: Why always strings and not bytes :sob: there probably is some `into_inner`/`to_bytes` function.

fedimint-testing/src/ln/mock.rs
```diff
@@ -67,19 +70,26 @@ impl LightningTest for FakeLightningTest {
...
     ) -> ln_gateway::Result<Bolt11Invoice> {
-        let ctx = bitcoin::secp256k1::Secp256k1::new();
+        let ctx = bitcoin30::secp256k1::Secp256k1::new();
 
         Ok(InvoiceBuilder::new(Currency::Regtest)
             .description("".to_string())
-            .payment_hash(sha256::Hash::hash(&[0; 32]))
+            .payment_hash(bitcoin29_to_bitcoin30_sha256_hash(sha256::Hash::hash(
+                &[0; 32],
+            )))
```
elsirion, line 79: You can just construct the `bitcoin30` hash right away here afaik?

tvolk131: Done.

fedimint-core/src/bitcoin_migration.rs
tvolk131: Ope 🙊 my bad! Fixed.

tvolk131:
Closing now that we merged https://github.com/fedimint/fedimint/pull/5118 which also does this

PR #4688, opened 2024-03-23
refactor: better key&value decoding in db code

dpc:
And other improvements around it.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1738,7 +1738,7 @@ pub struct ClientBuilder {
     module_inits: ClientModuleInitRegistry,
     primary_module_instance: Option<ModuleInstanceId>,
     admin_creds: Option<AdminCreds>,
-    db: Database,
+    db_no_encoders: Database,
     stopped: bool,
 }
 
@@ -1748,7 +1748,7 @@ impl ClientBuilder {
             module_inits: Default::default(),
             primary_module_instance: Default::default(),
             admin_creds: None,
-            db,
+            db_no_encoders: db,
             stopped: false,
         }
     }
@@ -1758,7 +1758,7 @@ impl ClientBuilder {
             module_inits: client.module_inits.clone(),
             primary_module_instance: Some(client.primary_module_instance),
             admin_creds: None,
-            db: client.db.clone(),
+            db_no_encoders: client.db.with_decoders(Default::default()),
             stopped: false,
         }
     }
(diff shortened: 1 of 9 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -1293,6 +1298,15 @@ pub struct DatabaseTransaction<'tx, Cap = NonCommittable> {
...
 
+impl<'tx, Cap> fmt::Debug for DatabaseTransaction<'tx, Cap> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_fmt(format_args!(
+            "DatabaseTransaction {{ tx: {:?}, decoders={:?} }}",
+            self.tx, self.decoders
+        ))
+    }
+}
+
```
elsirion, line 1319: nit: why not use [`std::fmt::DebugStruct`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html)?

fedimint-rocksdb/src/lib.rs
```diff
@@ -35,6 +36,18 @@ fn is_power_of_two(num: usize) -> bool {
     num.count_ones() == 1
 }
 
+impl<'a> fmt::Debug for RocksDbReadOnlyTransaction<'a> {
```
elsirion, line 39: ~`MemTransaction` currenlty derives `Debug`, resulting in a lot of detail being printed. Maybe we want to thin it out a bit if DBTX are going to be logged a lot.~

I see you did so in the next commit :laughing:

elsirion (approved):

fedimint-client/src/lib.rs
```diff
@@ -1758,7 +1758,7 @@ impl ClientBuilder {
             module_inits: client.module_inits.clone(),
             primary_module_instance: Some(client.primary_module_instance),
             admin_creds: None,
-            db: client.db.clone(),
+            db_no_encoders: client.db.with_decoders(Default::default()),
```
m1sterc001guy, line 1761: This function is eventually going away right? Initially just confused me since this is called `db_no_encoders` and its created using `with_decoders`

fedimint-core/src/db/mod.rs
```diff
@@ -1324,6 +1335,37 @@ fn decode_value<V: DatabaseValue>(
     V::from_bytes(value_bytes, decoders)
 }
 
+fn decode_value_expect<V: DatabaseValue>(
```
m1sterc001guy, line 1338: nit: docstring

m1sterc001guy (approved):

fedimint-core/src/db/mod.rs
dpc: TIL;

But looking at it - it doesn't look like that much of an improvement. Will use next time I'm doing anything like that. If I remember. :D

fedimint-client/src/lib.rs
dpc: > `from_existing` is eventually going away right? 

No idea. If so, good riddance. :)

fedimint-core/src/db/mod.rs
elsirion: I think I learned about it reviewing a Fedimint PR :laughing: probably by @maan2003, he always knows about these cool things!

fedimint-client/src/lib.rs
elsirion: Yep, commented that in the other PR, we don't need it anymore now that the client can shut down cleanly, dropping all DB references.

PR #4689, opened 2024-03-23
chore: dyntypes decoding improvements

dpc:


fedimint-client/src/lib.rs
```diff
@@ -1758,7 +1758,7 @@ impl ClientBuilder {
             module_inits: client.module_inits.clone(),
             primary_module_instance: Some(client.primary_module_instance),
             admin_creds: None,
-            db: client.db.clone(),
+            db_no_encoders: client.db.with_decoders(Default::default()),
```
dpc, line 1761: Notice - in `ClientBuilder`, the `db` should not contain any built-in decoders. It's somewhat important, and in here e.g. we would unnecessarily inherit decoders from previous `Client`, which is harmless right now, but could lead to bugs. For this reason I decided to rename it to this somewhat odd looking name.

fedimint-core/src/core.rs
```diff
@@ -322,15 +343,53 @@ impl Decoder {
         DecoderBuilder::default()
     }
 
+    /// System Dyn-type, don't use.
+    #[doc(hidden)]
+    pub fn builder_system() -> DecoderBuilder {
```
dpc, line 348: A bit of a ... workaround... . `DynState` is not really a `DynType` in the same sense as `DynInput` etc. are - it doesn't belong to a real module. But it does work exactly the same, so it's convenient to just make up and ID etc. for it. But architecturally it's a hack, and it's going to lead to papercuts and workarounds like here.

fedimint-core/src/core.rs
```diff
@@ -322,15 +343,53 @@ impl Decoder {
...
+        &self,
+        reader: &mut dyn Read,
+        total_len: u64,
+        module_id: ModuleInstanceId,
+        decoders: &ModuleDecoderRegistry,
+    ) -> Result<DynType, DecodeError> {
+        let mut reader = reader.take(total_len);
+
+        let val = self.decode_partial(&mut reader, module_id, decoders)?;
+        let left = reader.limit();
```
dpc, line 369: Originally with @maan2003 when debugging, we were printing the bytes here as well, but I decided against it as it opens the code to problems: e.g. what if reading leftover bytes leads to IO errors? Which error we return, which ignore. Also - what if the reader blocks, etc. Seems not worth it.

maan2003: just first print error without bytes, then read bytes and print bytes in seperate message. if all this work, return the error to caller

fedimint-core/src/core.rs
elsirion: I think we could get rid of nested dyn types by serializing the transaction before putting it into the tx state machine. Over the wire it's sent as hex-encoded bytes anyway (using `SerdeModuleEncoding<Transaction>`).

fedimint-client/src/lib.rs
elsirion: Now that the client can shut down cleanly we should get rid of `from_existing`, that was our old workaround to not have to re-open the DB.

fedimint-client/src/lib.rs
```diff
@@ -1822,12 +1822,12 @@ impl ClientBuilder {
         Ok(())
     }
 
-    pub fn db(&self) -> &Database {
-        &self.db
+    pub fn db_no_encoders(&self) -> &Database {
```
elsirion, line 1825: Should be called `db_no_decoders` imo, encoding doesn't require any registry, only decoding.

fedimint-core/src/core.rs
dpc: Might require migration, possibly extra work -> outside of scope of this change. :)

fedimint-core/src/core.rs
dpc: I still don't like it. :D

fedimint-client/src/lib.rs
dpc: Oops. brainfart.

elsirion (dismissed):

elsirion (approved):
I trust you didn't make major changes since last review, at least it didn't look like it.

maan2003 (approved):

Issue #4699, opened 2024-03-23
Client DB migrations not being run


elsirion:
Old v0.2 clients did not have a concept of database versions, so they did not write versioning information to their DB. The current DB migration code, particularly `remove_current_db_version_if_exists`, assumes that "no versioning info" means we have a fresh DB without any contents, thus can skip migrations and just set the current version to the target version.

https://github.com/fedimint/fedimint/blob/64468b6957f6424791a71094f4c183661be6dcf4/fedimint-core/src/db/mod.rs#L1980-L1994

PR #4703, opened 2024-03-23
fix: database migrations not run

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/4699

The database version migration code previously assumed the either the old DB version key exists or that the new one exists. For old clients, neither of them exist, which causes the migration code to assume that it is a new database. This is not a safe assumption. Instead of setting the database version to `target_db_version`, it needs to be set to `DatabaseVersion(0)`.

Dummy client db migration test has been modified to cover this scenario.


dpc (approved):

elsirion (approved):

fedimint-backports:
- #4708

Issue #4536, opened 2024-03-12
Allow client migration tests to write arbitrary bytes to database snapshots


m1sterc001guy:
It's currently difficult to test client state machine migrations because we want to write the old state into the database, which requires keeping extra structs and enums around to be able to do this.

If the tests could write raw bytes directly into the database, we could serialize the old bytes into a string and write that into the db, which would allow us to delete the old structs and the old enums.

elsirion:
PRs that need cleanup once this lands:
- [ ] #4024
- [ ] #3820

Keeping these old enum variants around is very bad for maintainability.

PR #4707, opened 2024-03-23
Add missing state machine migration

### LINKED ISSUE #4700: Missing client DB migration in #3820
We forgot to add a migration for `LightningReceiveConfirmedInvoice`, which also had `KeyPair` changed to `ReceivingKey`:

https://github.com/fedimint/fedimint/blob/64468b6957f6424791a71094f4c183661be6dcf4/modules/fedimint-ln-client/src/receive.rs#L188-L191

Introduced in #3820

elsirion:
I was considering to also rebuild the other, existing migration in the way I built the new one, eliminating the need for the `LightningReceiveStates::SubmittedOfferV0` variant, but that seems like too big of a change to backport and will follow in a separate PR to master only.

Fixes #4700

### DIFF
--- a/fedimint-client/src/db.rs
+++ b/fedimint-client/src/db.rs
@@ -348,6 +348,14 @@ pub async fn apply_migrations_client(
         .get_value(&DatabaseVersionKey(module_instance_id))
         .await;
 
+    info!(
+        ?disk_version,
+        ?target_db_version,
+        module_instance_id,
+        kind,
+        "Migrating client module database"
+    );
+
     let db_version = if let Some(disk_version) = disk_version {
         let mut current_db_version = disk_version;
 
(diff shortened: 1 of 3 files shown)


m1sterc001guy:
> eliminating the need for the LightningReceiveStates::SubmittedOfferV0 variant

This means we will need to keep `LightningReceiveStates::SubmittedOffer` (the new one) around forever or write a migration to delete it right?

m1sterc001guy (approved):
I like your approach better for this issue too https://github.com/fedimint/fedimint/issues/4536 since it keeps some knowledge of the old structure around. I was thinking of just serializing/deserializing the whole state as hex.

justinmoon (approved):

dpc (approved):

fedimint-backports:
- #4709

PR #4715, opened 2024-03-25
feat(cli): add list-modules command

techsavvyash:
## Description

This PR tries to close https://github.com/fedimint/fedimint/issues/3119.

- [x] make the `module` in `fedimint-cli module <module>` optional with the behaviour for listing modules when <module> is not provided.

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -10,7 +10,7 @@ use bitcoin_hashes::hex::ToHex;
 use clap::Subcommand;
 use fedimint_client::backup::Metadata;
 use fedimint_client::ClientHandleArc;
-use fedimint_core::config::FederationId;
+use fedimint_core::config::{ClientModuleConfig, FederationId};
 use fedimint_core::core::{ModuleInstanceId, ModuleKind, OperationId};
 use fedimint_core::encoding::Encodable;
 use fedimint_core::time::now;
@@ -41,6 +41,19 @@ use crate::{metadata_from_clap_cli, LnInvoiceResponse};
 pub enum ModuleSelector {
     Id(ModuleInstanceId),
     Kind(ModuleKind),
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub enum ModuleStatus {
+    Active,
+    UnsupportedByClient,
+}
+
+#[derive(Serialize)]
+struct ModuleInfo {
+    kind: ModuleKind,
+    id: u16,
+    status: ModuleStatus,
 }
 
 impl FromStr for ModuleSelector {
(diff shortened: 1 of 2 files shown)


fedimint-cli/src/client.rs
```diff
@@ -601,6 +609,32 @@ pub async fn handle_command(
...
+                    id: *id,
+                    status: if let Some(_value) = client.get_config().modules.get(id) {
+                        ModuleStatus::Active
+                    } else {
+                        ModuleStatus::UnsupportedByClient
+                    },
+                })
+                .collect();
+            return Ok(json!({
+                "List": module_list,
```
dpc, line 635: In no other output we use capital letter keys.

techsavvyash: changes to `"list"`. Sample output looks like:
```json
{
  "List": [
    {
      "kind": "ln",
      "id": 0,
      "status": "Active"
    },
    {
      "kind": "mint",
      "id": 1,
      "status": "Active"
    },
    {
      "kind": "wallet",
      "id": 2,
      "status": "Active"
    },
    {
      "kind": "meta",
      "id": 3,
      "status": "Active"
    },
    {
      "kind": "unknown",
      "id": 4,
      "status": "UnsupportedByClient"
    }
  ]
}
```

techsavvyash:
> I think we should make the `module` in `fedimint module <module>` optional, and when `None` just print the list of available modules (both kind names, deduped, and ids). I think that was the goal.
> 
> https://github.com/fedimint/fedimint/blob/d7942cb039715707a2cfb5a7e7d7763596c100c7/fedimint-cli/src/client.rs#L174
> 
> , I'm not entirely sure if `None` can be used there.
> A `--list` flag would be OK there too.
> 
> We already have `discover-versions` that is kind of similar to `list-modules`... Not sure what to do about it, and o we need `list-modules` as well.

Made changes to make the `module` optional, now the following behaviour occurs:

```
Yashs-Mac-mini:fedimint techsavvyash$ fedimint-cli module
{
  "list": [
    {
      "kind": "ln",
      "id": 0,
      "status": "Active"
    },
    {
      "kind": "mint",
      "id": 1,
      "status": "Active"
    },
    {
      "kind": "wallet",
      "id": 2,
      "status": "Active"
    },
    {
      "kind": "meta",
      "id": 3,
      "status": "Active"
    },
    {
      "kind": "unknown",
      "id": 4,
      "status": "UnsupportedByClient"
    }
  ]
}
Yashs-Mac-mini:fedimint techsavvyash$ fedimint-cli module meta
Usage: meta <COMMAND>

Commands:
  get              Get current consensus value
  get-rev          Get current consensus value revision
  get-submissions  Get value change submissions
  submit           Submit value change proposal
  help             Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
Yashs-Mac-mini:fedimint techsavvyash$ fedimint-cli module wallet
2024-03-25T17:50:36.081795Z ERROR fedimint_cli: This module does not implement cli commands
{
  "error": "This module does not implement cli commands"
}
Yashs-Mac-mini:fedimint techsavvyash$ fedimint-cli module yash
2024-03-25T17:50:42.864549Z ERROR fedimint_cli: No module with this kind found
{
  "error": "No module with this kind found"
}

```

maan2003 (approved):

PR #4719, opened 2024-03-25
test: Allow Client DB Migration tests to write arbitrary bytes

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/4536

In retrospect, forcing the client db migration tests to use `DynState` actually turned out to be a mistake, since we can't easily create states that we want to delete. This PR fixes this by just returning `Vec<u8>` and refactoring the existing tests to create their states by encoding them into vectors.


elsirion (approved):

modules/fedimint-ln-client/src/receive.rs
```diff
@@ -39,12 +39,11 @@ const RETRY_DELAY: Duration = Duration::from_secs(1);
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub enum LightningReceiveStates {
-    SubmittedOfferV0(LightningReceiveSubmittedOfferV0),
+    SubmittedOffer(LightningReceiveSubmittedOffer),
```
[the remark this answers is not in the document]
elsirion: This cleanup requires another migration since currently the `SubmittedOffer` variant has index 5 when encoded, while it is switched back to 0 in this PR, which was the state in the previous DB version.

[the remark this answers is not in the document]
m1sterc001guy: I originally had it like then but ended up squashing since there were some refactors in the later commits that didn't fit. I'll try to preserve the history next time

m1sterc001guy:
review call: need to add another migration to delete the newly added state

modules/fedimint-ln-client/src/receive.rs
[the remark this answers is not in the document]
m1sterc001guy: Fixed the commit history! Hopefully it is easier to review now.

I also added the new migration and a test for it.

fedimint-client/src/db.rs
```diff
@@ -557,3 +557,58 @@ pub async fn remove_old_and_persist_new_inactive_states(
...
+pub async fn migrate_state<S: State + IntoDynInstance<DynType = DynState>>(
+    module_instance_id: ModuleInstanceId,
+    active_states: Vec<(Vec<u8>, OperationId)>,
+    inactive_states: Vec<(Vec<u8>, OperationId)>,
+    decoders: ModuleDecoderRegistry,
+    migrate: fn(
+        &[u8],
+        ModuleInstanceId,
+        &ModuleDecoderRegistry,
+    ) -> anyhow::Result<Option<DynState>>,
```
elsirion, line 573: * Why return a `Result`, what do we do if anything fails? Imo the developer should decide what to do in that case (delete state?)
* TODO(follow-up): not use `DynState` and just return `Vec<u8>`, then we can also get rid of `ModuleInstanceId`

fedimint-client/src/db.rs
```diff
@@ -557,3 +557,58 @@ pub async fn remove_old_and_persist_new_inactive_states(
...
 }
+
+/// Migrates a particular state by looping over all active and inactive states.
+/// If the `migrate` closure returns `None`, this state was not migrated and
+/// should be added to the new state machine vectors.
+pub async fn migrate_state<S: State + IntoDynInstance<DynType = DynState>>(
+    module_instance_id: ModuleInstanceId,
+    active_states: Vec<(Vec<u8>, OperationId)>,
+    inactive_states: Vec<(Vec<u8>, OperationId)>,
+    decoders: ModuleDecoderRegistry,
```
elsirion, line 568: Do we need decoders?

elsirion (approved):
All fixable later, let's go!

fedimint-client/src/db.rs
dpc: Returning an error might let us panic nicer (with more information about which module, etc.) in a higher level.

modules/fedimint-dummy-tests/tests/tests.rs
```diff
@@ -207,19 +208,66 @@ mod fedimint_migration_tests {
...
+            Amount::from_sats(1000)
+                .consensus_encode(&mut input_state)
+                .expect("Amount is encodable");
+            TransactionId::from_slice(&BYTE_32)
+                .expect("Couldn't create TransactionId")
+                .consensus_encode(&mut input_state)
+                .expect("TransactionId is encodable");
+            input_operation_id
+                .consensus_encode(&mut input_state)
+                .expect("OperationId is encodable");
```
maan2003, line 225: maybe?
```rust
(Amount::from_sats(1000), TransactionId::from_slice(&BYTE_32),  input_operation_id).consensus_encode_to_vec()
```

maan2003 (approved):
LGTM, we can fix the nit in follow up

fedimint-client/src/db.rs
m1sterc001guy: Decoders are convenient for how we've structured the db migrations right now, but once we switch to returning a `Vec<u8>` they won't be necessary

fedimint-client/src/db.rs
m1sterc001guy: Follow up: https://github.com/fedimint/fedimint/issues/4823

PR #4724, opened 2024-03-25
fix(fedimint-server): build module decoder registry from config

dpc:
`FedimintServer` needs to be running with decoders matchings
the config. That means that:

On first start, run without any decoders (as there are no module
instances to decode anything in the first place).

After DKG, or on subsequent starts, configure decoders from the
config.

On top of: #4643


maan2003 (approved):

douglaz (approved):
Looks good enough

dpc:
image


Everything passed, yet GHA removed it from the queue...

Issue #4702, opened 2024-03-23
Gateway stores password in plaintext


m1sterc001guy:
Today the gateway stores the administrator password in plaintext in the database. We should instead store a hash and comparing hashes while authenticating. https://github.com/fedimint/fedimint/pull/4642#discussion_r1536659840

Main challenge will be doing this in a backwards compatible way.

m1sterc001guy:
> Main challenge will be doing this in a backwards compatible way.

I supposed the gateway could compare the stored hash against the plaintext password and the plaintext password hashed, to provide backwards compatibility.

dpc:
> Main challenge will be doing this in a backwards compatible way.

Make fields options, rewrite old field  on access.

BTW. Make sure to add some random generated salt.

elsirion:
The same needs to be done for the guardian password (although there it might be simpler, just delete the file and require password on startup and keep it in memory from there on).

PR #4727, opened 2024-03-25
fix: store gateway password hash

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/4702

Uses a database migration to migrate the old password to storing a hash instead.


dpc (approved):

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1118,6 +1119,20 @@ async fn test_gateway_configuration() -> anyhow::Result<()> {
     })
     .await;
 
+    // Verify that the gateway's password is set and its the hash of the plaintext
```
tvolk131, line 1122: Nit: I think this comment is somewhat unclear. What about:

"Verify that the gateway's password is stored correctly (i.e. the stored hash and salt match the password)"

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -157,6 +168,18 @@ fn v1_routes(gateway: Gateway) -> Router {
         .layer(CorsLayer::permissive())
 }
 
+/// Creates a password hash by appending a 4 byte salt to the plaintext
```
tvolk131, line 171: Nit: I know 32 bits of salt already makes rainbow tables extremely expensive, but would it be worth bumping the salt to 64 bits since it's only an extra few bytes of storage?

tvolk131 (approved):

gateway/ln-gateway/src/rpc/rpc_server.rs
dpc: I completely missed it's a number. Make that stuff a 16B array of randomness, don't think twice. :)

m1sterc001guy: Changed to 16 bytes

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -157,6 +168,18 @@ fn v1_routes(gateway: Gateway) -> Router {
...
+/// password.
+pub fn hash_password(plaintext_password: String, salt: [u8; 16]) -> sha256::Hash {
```
dpc, line 173: Would newtype it.

dpc (approved):

okjodom (approved):
🔥

gateway/ln-gateway/src/db.rs
```diff
@@ -63,18 +69,37 @@ impl_db_record!(
...
 
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable, Serialize, Deserialize)]
 pub struct GatewayConfiguration {
-    pub password: String,
+    pub hashed_password: sha256::Hash,
     pub num_route_hints: u32,
     #[serde(with = "serde_routing_fees")]
     pub routing_fees: RoutingFees,
     pub network: Network,
+    pub password_salt: [u8; 16],
```
elsirion, line 94: nit: the two fields could have been next to each other

gateway/ln-gateway/src/rpc/rpc_server.rs
elsirion, line 171: The docstring never got changed after increasing the salt length

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -157,6 +168,18 @@ fn v1_routes(gateway: Gateway) -> Router {
...
+    let mut bytes = Vec::<u8>::new();
+    plaintext_password
+        .consensus_encode(&mut bytes)
+        .expect("Password is encodable");
+    salt.consensus_encode(&mut bytes)
+        .expect("Salt is encodable");
+    sha256::Hash::hash(&bytes)
+}
```
elsirion, line 181: Probably fine since the salt is drawn randomly, but normally one would use an HMAC imo or even specialized password hashing algorithms.

PR #4733, opened 2024-03-26
chore: document `PayInvoiceRequest` fields

tvolk131:


gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -50,12 +50,16 @@ message GetNodeInfoResponse {
...
+  // Bolt11 invoice
   string invoice = 1;
 
+  // The maximum delay, in blocks, that the payment route can be locked for
   uint64 max_delay = 2;
 
+  // The maximum fee, in millisats, that will be paid as a lightning routing fee
   uint64 max_fee_msat = 3;
 
+  // The hash of the payment
```
m1sterc001guy, line 62: Can you add that this is needed for `pay_private`?

elsirion (dismissed):

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: Done. Leaving unresolved since I don't have a good answer to the question above.

gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -59,7 +59,7 @@ message PayInvoiceRequest {
   // The maximum fee, in millisats, that will be paid as a lightning routing fee
   uint64 max_fee_msat = 3;
 
-  // The hash of the payment
+  // The hash of the payment, used for private payments (see `ILnRpcClient::pay_private()`)
```
elsirion, line 62: Not sure if this is very helpful, @m1sterc001guy does this API support private payments? I'd just create a new endpoint, otherwise you'll end up making the invoice optional, but adding a whole bunch of other optional fields (destination pub key, payment secret, …).

m1sterc001guy: `payment_hash` isn't used for CLN, its only used for LND in `pay_private`. 

We use the same serialization of requests (protobuf) for both CLN and LND, even though for LND the request never goes over the wire. I think its fine to keep it, can be fixed later if we refactor so that CLN is the only implementation that uses the protobuf generated code.

m1sterc001guy (approved):

gateway/ln-gateway/proto/gateway_lnrpc.proto
elsirion: This `PayInvoiceRequest` cannot be used to call `pay_private` though, it just lacks the necessary information.

I checked the places where `payment_hash` from this struct is used and there are exactly two:
1. In `buy_preimage_over_lightning` where it is extracted from the invoice (which is also sent in the same request)
https://github.com/fedimint/fedimint/blob/5918169d58d10d5f2febcb48cab1ddf36946a210/gateway/ln-gateway/src/state_machine/pay.rs#L414
2. In `ClnRpcService`, where it's thrown away:
https://github.com/fedimint/fedimint/blob/5918169d58d10d5f2febcb48cab1ddf36946a210/gateway/ln-gateway/src/bin/cln_extension.rs#L365-L370

There once was a third use in the LND implementation of `pay`, but you recently removed that. `pay_private` doesn't use the RPC structs at all because a) it was too annoying to impl for CLN and b) gRPC in Rust sucks (way too much boilerplate needed, I'd much rather use our usual JSON RPC, but that's a refactoring for another time :smile:).

m1sterc001guy: Ah ok sorry I was looking at the wrong pay haha. Yes you are right, @tvolk131 lets remove the `pay_private` comment.

tvolk131: Done.

m1sterc001guy (approved):

elsirion (approved):

PR #4743, opened 2024-03-26
chore: db migrations logging and shortcircuit improvements

m1sterc001guy:
This PR makes some small improvements to the logging during the database migrations and also re-introduces a short-circuit mechanism to prevent running the database migrations on newly created databases.

Some of the messages were confusing to deduce what happened based on the output. Also the comments have been changed to clarify certain aspects.

The main improvement is to return `target_db_migration` if we detect that the database is a newly created database. Right now newly created database's always set their database version to 0, which causes them to run all migrations needlessly. Using `target_db_version` is a short-circuit mechanism to make the migrations faster and make the logs more intuitive.

### DIFF
--- a/fedimint-client/src/db.rs
+++ b/fedimint-client/src/db.rs
@@ -4,8 +4,9 @@ use fedimint_core::api::ApiVersionSet;
 use fedimint_core::config::{ClientConfig, FederationId};
 use fedimint_core::core::{ModuleInstanceId, OperationId};
 use fedimint_core::db::{
-    migrate_database_version, Database, DatabaseTransaction, DatabaseValue, DatabaseVersion,
-    DatabaseVersionKey, IDatabaseTransactionOpsCoreTyped,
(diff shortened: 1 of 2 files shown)


fedimint-client/src/db.rs
```diff
@@ -324,24 +324,14 @@ pub async fn apply_migrations_client(
     module_instance_id: ModuleInstanceId,
     decoders: ModuleDecoderRegistry,
 ) -> Result<(), anyhow::Error> {
-    // TODO(support:v0.3):
```
m1sterc001guy, line 327: On further inspection I realized this comment isn't exactly right, since we're relying on this function to create the database version for newly created databases too.

That's what prompted the change from `migrate_database_version` to `create_database_version`. So we can't remove this after 0.3, but I think that's ok since for newly created database, all this will do is write the database version to disk.

elsirion (approved):
Looks reasonable enough, probably not worth the complexity now but will become worth it in the future when we have more migrations to run through.

dpc (approved):

PR #4745, opened 2024-03-26
chore: don't buil `cargo-workspace` on macos

dpc:
Seems broken:

https://github.com/fedimint/fedimint/actions/runs/8442492153/job/23124016592


maan2003 (approved):
makes sense, you need a real machine to cut releases

bradleystachurski (approved):

PR #4469, opened 2024-03-05
fix(ci): stop notifing about failing merge queue builds

dpc:

PR #4748, opened 2024-03-27
chore: notify for back-compat failures in merge group

bradleystachurski:
We recently started skipping notifications for failures in the merge queue in https://github.com/fedimint/fedimint/pull/4469. That makes sense for `ci-nix` since the workflows are required to pass, but back-compat isn't required and we may fail without notifications (e.g. [failed run](https://github.com/fedimint/fedimint/actions/runs/8445095915/job/23131758594)).

Alternatively, and perhaps preferably, we can make back-compat required in the merge queue.

### DIFF
--- a/.github/workflows/ci-backwards-compatibility.yml
+++ b/.github/workflows/ci-backwards-compatibility.yml
@@ -78,7 +78,7 @@ jobs:
             !/tmp/nix-build-*/source/
 
   notifications:
-    if: always() && github.repository == 'fedimint/fedimint' && github.event_name != 'merge_group'
+    if: always() && github.repository == 'fedimint/fedimint'
     name: "Notifications"
     timeout-minutes: 1
     runs-on: ubuntu-22.04


dpc:
The Notification workflow doesn't do anything for Required or not. It is meant pure for sending notifications and not block (or not block) anything.

Backward Compat should just be Required in the CI, which means it will be required in the MQ as well. @bradleystachurski 

Did you notice anything passing through MQ with a failing Backward Compat?

bradleystachurski:
Apologies if the PR description is confusing. I agree, I think back-compat should be required to pass in the MQ :+1:

If we didn't want that constraint, this PR is an alternative since we had a failed back-compat run in the MQ that merged and didn't make noise in `#alerts` (which motivated https://github.com/fedimint/fedimint/pull/4749).

https://github.com/fedimint/fedimint/actions/runs/8445095915/job/23131758594

@dpc, okay if I close in favor of making back-compat required in MQ?

dpc:
> @dpc, okay if I close in favor of making back-compat required in MQ?

I was under impression that whatever "is Required" is also "Required in MQ". So we just make Backward-Compat Required, yes. Now I see it's not, which seems like we forgot about it, since it's running smoothly for a while now and times are good too.

bradleystachurski:
Closing in favor of making back-compat required

PR #4754, opened 2024-03-28
test: do we use the non-human readbable serde encoding for tbs public types?

joschisan:


elsirion:
Can be closed after #4755 afaik

PR #4755, opened 2024-03-28
fix: do not encode the known length for bls12_381 public types

joschisan:
This pr unifies both implementations of serde for the bls12_381 public types Scalar, G1Affine, G2Affine into one. 

Furthermore we fix the non-human readable encoding to not encode the length of the array since this is known at deserialisation. This is not a breaking change since we only use this encoding for the DKG and tbs public types. In the DKG the encodings are only used for messages between peers and the tpe public types are only ever encoded in human readable format as we have tested here: https://github.com/fedimint/fedimint/pull/4754

The new serde implementation is identical to the one we are trying to upstream here: https://github.com/zkcrypto/bls12_381/pull/125


dpc (dismissed):
But this should get @elsirion approval before landing.

fedimint-core/src/bls12_381_serde.rs
```diff
@@ -1,81 +1,74 @@
...
-        byte_array.copy_from_slice(&bytes);
+        serdect::array::deserialize_hex_or_bin(&mut byte_array, d)?;
 
         Option::from(Scalar::from_bytes(&byte_array))
-            .ok_or_else(|| Error::custom("Could not decode compressed G1Affine"))
+            .ok_or(Error::custom("Could not decode scalar"))
     }
 }
 
-macro_rules! impl_serde_g {
```
elsirion, line 38: It's so compact now that the macro was more overhead than worth I guess?

fedimint-core/src/bls12_381_serde.rs
```diff
@@ -1,81 +1,74 @@
...
+        D: Deserializer<'d>,
+    {
+        let mut byte_array = [0; 32];
 
-        byte_array.copy_from_slice(&bytes);
+        serdect::array::deserialize_hex_or_bin(&mut byte_array, d)?;
 
         Option::from(Scalar::from_bytes(&byte_array))
-            .ok_or_else(|| Error::custom("Could not decode compressed G1Affine"))
+            .ok_or(Error::custom("Could not decode scalar"))
```
elsirion, line 22: `Error::custom` is a heavy-ish operation (it allocates), should be only run on-demand by keeping `ok_or_else`.

Screenshot 2024-03-29 at 21-18-49 value rs - source

elsirion (approved):
Looks good, one nit.

dpc (approved):

elsirion (approved):

PR #4756, opened 2024-03-28
Implement meta field fallback logic and caching

elsirion:
The new meta module (#4513) allows guardians to change meta fields by voting on them. But without providing an interface for actually fetching meta fields it's cumbersome to use. Furthermore, old federations still use a meta override file or the federation's static config. 

This PR implements background fetching and querying in order of precedence of all these sources.

TODO:
- [ ] Test: likely manual? Did we ever set up an override meta server in devimint?
- [ ] Expose via CLI for testing


elsirion (commented):
* Allow force updates
* Run less frequently
* Notifications?

justinmoon:
superceded by https://github.com/fedimint/fedimint/pull/4768

PR #4760, opened 2024-03-28
feat: backoff in retry helper

maan2003:
I really like 0.2, 0.2, 0.4, 0.6, 1.0, 1.6, 2.6, 4.6, 5, 5...

We should export some statergy in fedimint-core but naming is hard.

### DIFF
--- a/fedimint-core/Cargo.toml
+++ b/fedimint-core/Cargo.toml
@@ -53,6 +53,7 @@ macro_rules_attribute = "0.2.0"
 bitvec = "1.0.1"
 parity-scale-codec = { version = "3.6.9", features = ["derive"] }
 imbl = "2.0.3"
+backon = "=0.4.3" # don't upgrade unless really needed
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 jsonrpsee-ws-client = { version = "0.22.2", features = ["webpki-tls"], default-features = false }
--- a/fedimint-core/src/util/mod.rs
+++ b/fedimint-core/src/util/mod.rs
@@ -9,7 +9,6 @@ use std::io::Write;
 use std::path::Path;
 use std::pin::Pin;
 use std::str::FromStr;
-use std::time::Duration;
 use std::{fs, io};
 
 use anyhow::format_err;
@@ -309,6 +308,11 @@ pub fn handle_version_hash_command(version_hash: &str) {
     }
 }
 
+pub use backon::{
+    BackoffBuilder, ConstantBuilder as ConstantBackoff, ExponentialBuilder as ExponentialBackoff,
+    FibonacciBuilder as FibonacciBackoff,
+};
+
 /// Run the supplied closure `op_fn` up to `max_attempts` times. Wait for the
 /// supplied `Duration` `interval` between attempts
 ///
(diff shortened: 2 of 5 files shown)


fedimint-core/Cargo.toml
```diff
@@ -53,6 +53,7 @@ macro_rules_attribute = "0.2.0"
 bitvec = "1.0.1"
 parity-scale-codec = { version = "3.6.9", features = ["derive"] }
 imbl = "2.0.3"
+backon = "0.4.3"
```
dpc, line 56: Every new dep worries me, but seems small and very popular ... https://crates.io/crates/backon...

fedimint-core/src/util/mod.rs
```diff
@@ -320,41 +324,41 @@ pub fn handle_version_hash_command(version_hash: &str) {
 pub async fn retry<F, Fut, T>(
```
dpc, line 324: Unclear if we even need it anymore. Seems like any future is retriable now? https://docs.rs/backon/latest/backon/trait.Retryable.html#tymethod.retry

fedimint-core/Cargo.toml
maan2003: maybe we just should fork it, just maths so should be pretty stable

fedimint-core/src/util/mod.rs
maan2003: we don't get logging, also that uses `tokio::time::sleep` (read hates wasm)

fedimint-core/Cargo.toml
dpc: Nah. We pin it, it's good. It's not a big crate.

fedimint-core/src/util/mod.rs
maan2003: also (|| async { }).retry(...) feels very awkward

fedimint-core/Cargo.toml
maan2003: pinned and added comment so someone doesn't approve dependabot :smile:

maan2003:
adding backport label, it would be nice to have in fedi.

fedimint-core/src/util/mod.rs
dpc: True.

fedimint-core/src/util/mod.rs
```diff
@@ -456,7 +466,15 @@ mod tests {
...
 
-        let _ = retry("Run 3 times".to_string(), closure, Duration::ZERO, 3).await;
+        let _ = retry(
+            "Run 3 once",
+            closure,
+            ConstantBackoff::default()
+                .with_delay(Duration::ZERO)
+                // retry two error
+                .with_max_times(2),
+        )
```
maan2003, line 476: NOTE: slight change in semantics

fedimint-core/src/util/mod.rs
```diff
@@ -320,41 +324,41 @@ pub fn handle_version_hash_command(version_hash: &str) {
 pub async fn retry<F, Fut, T>(
     op_name: impl Into<String>,
     op_fn: F,
-    interval: Duration,
-    max_attempts: u32,
+    statergy: impl backon::BackoffBuilder,
```
elsirion, line 327: Could we use this opportunity to switch the argument order? It's awkward to have the closure in the middle.

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -1486,8 +1486,10 @@ async fn test_gateway_executes_swaps_between_connected_federations() -> anyhow::
                     };
                     Ok(post_balances)
                 },
-                Duration::from_secs(1),
-                15,
+                FibonacciBackoff::default()
+                    .with_min_delay(Duration::from_millis(200))
+                    .with_max_delay(Duration::from_secs(3))
+                    .with_max_times(10),
```
elsirion, line 1492: New max delay: 15.6s, pretty similar to before.

```python
In [1]: def fib(start, n):
   ...:     prev = 0
   ...:     current = start
   ...:     for i in range(0,n):
   ...:         yield current
   ...:         next_current = prev + current
   ...:         prev = current
   ...:         current = next_current
   ...: 

In [2]: sum([min(3000, x) for x in fib(200, 10)])
Out[2]: 15600
```

elsirion (approved):

fedimint-core/src/util/mod.rs
maan2003: updated

elsirion (approved):

fedimint-core/src/util/mod.rs
```diff
@@ -319,42 +323,42 @@ pub fn handle_version_hash_command(version_hash: &str) {
 ///   error of the closure is returned
 pub async fn retry<F, Fut, T>(
     op_name: impl Into<String>,
+    statergy: impl backon::BackoffBuilder,
```
joschisan, line 326: typo

joschisan (approved):

fedimint-core/src/util/mod.rs
maan2003: we can live with typo :)
much better than reacquiring approvals

fedimint-backports:
- #4769

Issue #4486, opened 2024-03-07
Add ability for devimint to test upgraded binaries


bradleystachurski:
Capturing an idea from @justinmoon in an out-of-band chat.

Having the ability to test the behavior of upgrading binaries in devimint would be useful. For example:

- Start devimint with fedimintd v0.2.2
- Run specific devimint test (e.g. latency tests)
- Using the same dev_fed/datadirs, shutdown all peers, then start using fedimintd v0.3.0
- Run same devimint test

Backwards-compatibility tests will verify the behavior between different versions, but will not capture the behavior of a repeated test with an upgraded binary.

PR #4675, opened 2024-03-22
feat: upgrade test

maan2003:
use `./scripts/upgrade-test.sh path/to/0.2/fedimintd` to run.

PR #4762, opened 2024-03-29
feat(devimint): add upgrade tests for all binaries

bradleystachurski:
Inspired by https://github.com/fedimint/fedimint/pull/4675 and closes https://github.com/fedimint/fedimint/issues/4486.

Use with `just test-upgrades` (defaults to linear upgrade paths for v0.2.1 -> v0.2.2 -> current).

`fedimintd` upgrades fail against `master` but work on `releases/v0.3`, so use the backport PR (https://github.com/fedimint/fedimint/pull/4763) if you verify locally.

Tests upgrades for `fedimintd`, `fedimint-cli`, and `gatewayd`:
- Runs latency tests
- Upgrades path for single binary
- Reruns latency tests
- Continues for all versions passed to script

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -17,9 +17,9 @@ use fedimint_logging::LOG_DEVIMINT;
 use hex::ToHex;
 use ln_gateway::rpc::GatewayInfo;
 use serde_json::json;
-use tokio::fs;
 use tokio::net::TcpStream;
 use tokio::time::timeout;
+use tokio::{fs, try_join};
 use tracing::{debug, info};
 
 use crate::cli::{cleanup_on_exit, exec_user_command, setup, write_ready_file, CommonArgs};
@@ -98,7 +98,11 @@ pub async fn log_binary_versions() -> Result<()> {
     Ok(())
 }
 
-pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
+pub async fn latency_tests(
+    dev_fed: DevFed,
+    r#type: LatencyTest,
+    upgrade_clients: Option<&UpgradeClients>,
+) -> Result<()> {
     log_binary_versions().await?;
 
     #[allow(unused_variables)]
@@ -116,7 +120,17 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
     let max_p90_factor = 5.0;
     let p90_median_factor = 7;
 
-    let client = fed.new_joined_client("latency-tests-client").await?;
+    let client = match upgrade_clients {
+        Some(c) => match r#type {
+            LatencyTest::Reissue => c.reissue_client.clone(),
+            LatencyTest::LnSend => c.ln_send_client.clone(),
+            LatencyTest::LnReceive => c.ln_receive_client.clone(),
+            LatencyTest::FmPay => c.fm_pay_client.clone(),
+            LatencyTest::Restore => bail!("no reusable upgrade client for restore"),
+        },
+        None => fed.new_joined_client("latency-tests-client").await?,
+    };
+
     client.use_gateway(&gw_cln).await?;
     let initial_balance_sats = 100_000_000;
     fed.pegin_client(initial_balance_sats, &client).await?;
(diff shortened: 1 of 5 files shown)


devimint/src/tests.rs
```diff
@@ -350,6 +350,151 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
     Ok(())
 }
 
+pub async fn upgrade_tests(process_mgr: &ProcessManager, binary: UpgradeTest) -> Result<()> {
+    match binary {
+        UpgradeTest::Fedimintd { mut paths } => {
+            anyhow::ensure!(!paths.is_empty(), "Must provide at least 1 binary path");
+            let oldest_fedimintd = paths.first().expect("at least 1 binary was provided");
```
maan2003, line 357: let .. else maybe

devimint/src/tests.rs
```diff
@@ -350,6 +350,151 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
...
+                fedimintd_version
+            );
+            let mut dev_fed = dev_fed(process_mgr).await?;
+            let client = dev_fed.fed.new_joined_client("test-client").await?;
+            try_join!(stress_test_fed(&dev_fed), wait_session(&client))?;
+
+            // add the current fedimintd binary from PATH
+            paths.push("fedimintd".into());
+            for path in paths.iter().skip(1) {
+                dev_fed.fed.restart_all_with_bin(process_mgr, path).await?;
```
maan2003, line 372: this doesn't have the staggered shutdown like I had in initial upgrade tests

it is like to have some activity on some peers when upgrading

devimint/src/tests.rs
```diff
@@ -350,6 +350,151 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
...
+            info!(
+                "running first stress test for fedimint-cli version: {}",
+                fedimint_cli_version
+            );
+            let dev_fed = dev_fed(process_mgr).await?;
+            let client = dev_fed.fed.new_joined_client("test-client").await?;
+            try_join!(stress_test_fed(&dev_fed), wait_session(&client))?;
+
+            // add the current fedimint-cli binary from PATH
+            paths.push("fedimint-cli".into());
```
maan2003, line 407: ditto

devimint/src/tests.rs
```diff
@@ -350,6 +350,151 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
...
+            for path in paths.iter().skip(1) {
+                update_fedimint_cli_path(path);
+                let fedimint_cli_version = crate::util::FedimintCli::version_or_default().await;
+                info!("upgraded fedimint-cli to version: {}", fedimint_cli_version);
+                try_join!(stress_test_fed(&dev_fed), wait_session(&client))?;
```
maan2003, line 412: incorrect: this creates a new client everytime, you don't use use the old use old datadir.

this tests nothing

devimint/src/tests.rs
```diff
@@ -350,6 +350,151 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
...
+            try_join!(stress_test_fed(&dev_fed), wait_session(&client))?;
+
+            // add the current gatewayd binary from PATH
+            paths.push("gatewayd".into());
+            for path in paths.iter().skip(1) {
+                try_join!(
+                    dev_fed.gw_cln.restart_with_bin(process_mgr, path),
+                    dev_fed.gw_lnd.restart_with_bin(process_mgr, path),
+                )?;
+                try_join!(stress_test_fed(&dev_fed), wait_session(&client))?;
```
maan2003, line 485: what restarted the gateways while the stress tests were running, 

something like
```rust
try_join!(stress_test_fed(&dev_fed), async { sleep(30s).await; upgrade_gws(); })
```

justinmoon:
I had a slightly different vision for how this would work (which I could have explained better).

Basically, you'd need to specify a vector of versions somehow. Maybe you just expect all the binaries for a given version of fedimint to be in a `bin` folder. So let's say you provide `v0.1_bin_folder`, `v0.2_bin_folder` and `v0.3_bin_folder` to the test.

The test would run latency tests with all binaries from `v0.1_bin_folder`, then restart with binaries from `v0.2_bin_folder`, then restart with binaries from `v0.3_bin_folder`. Basically it seems like your version creates specific tests for each binary upgrade. Whereas my version just tests upgrading all binaries at the same time.

elsirion:
> your version creates specific tests for each binary upgrade. Whereas my version just tests upgrading all binaries at the same time

That version would be a bit more efficient indeed, and is probably worthwhile (@maan2003 also had some good comments to iterate on and it needs a rebase anyway).

But I prefer the "just specify version/commit an let nix do all the compiling" approach instead of just having some folders as you describe.

devimint/src/tests.rs
bradleystachurski: I updated to stagger the shutdown/restart of all peers since that's likely a closer approximation for upgrades in the wild. Let me know if there's a reason to prefer only staggering the last peer.

devimint/src/tests.rs
bradleystachurski: Good catch! Let me know if you see a more elegant solution than my latest approach.

devimint/src/tests.rs
bradleystachurski: Capturing our chat on the dev call: this is a bit tricky to restart while running latency tests and can be addressed later.

devimint/src/tests.rs
```diff
@@ -350,6 +350,151 @@ pub async fn latency_tests(dev_fed: DevFed, r#type: LatencyTest) -> Result<()> {
...
+                info!("upgraded fedimint-cli to version: {}", fedimint_cli_version);
+                try_join!(stress_test_fed(&dev_fed), wait_session(&client))?;
+                info!(
+                    "### fedimint-cli passed stress test for version {}",
+                    fedimint_cli_version
+                );
+            }
+            info!("## fedimint-cli upgraded all binaries successfully")
+        }
+        UpgradeTest::GatewayCli { mut paths } => {
```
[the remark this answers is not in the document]
bradleystachurski: Context for other reviewers: `gateway-cli` doesn't use a persistent data dir, so testing upgrades doesn't add additional coverage relative to existing back-compat tests.

bradleystachurski:
@maan2003 would you prefer handling the backport in https://github.com/fedimint/fedimint/pull/4856?

I can close my separate PR (https://github.com/fedimint/fedimint/pull/4763) for backporting to `releases/v0.3`, which is necessary for testing locally due to the issue with discovering api versions on master.

maan2003:
it is not client related, we can backport this separately

maan2003 (approved):

devimint/src/tests.rs
maan2003: I delayed last peer so other peers could complete a session without

elsirion:
Do we even need to backport this? It seems to me as if we could just run any upgrade tests off master, specifying the versions we want to test.

elsirion (approved):

bradleystachurski:
> Do we even need to backport this? It seems to me as if we could just run any upgrade tests off master, specifying the versions we want to test.

@elsirion good question! Testing against master breaks since we fail to discover a common api version, but if we don't test the target dir binaries we're able to test upgrades vs master.

https://github.com/fedimint/fedimint/pull/4947

PR #4765, opened 2024-03-29
feat: merge concurrent calls to update_gateway_cache

maan2003:
this is much simpler after https://github.com/fedimint/fedimint/pull/4785

this is also fixes a panic because we were not using autocommit.

## Summary by CodeRabbit

- **Refactor**
	- Updated the `update_gateway_cache` function for better control flow.
	- Improved error handling and logging throughout the system.
	- Renamed and removed specific configurations for clarity and future-proofing.
	- Enhanced continuous update capabilities for gateway cache.

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -66,6 +66,7 @@ use secp256k1_zkp::{All, Secp256k1};
 use serde::{Deserialize, Serialize};
 use strum::IntoEnumIterator;
 use thiserror::Error;
+use tokio::sync::Mutex;
 use tracing::error;
 
 use crate::db::PaymentResultPrefix;
@@ -381,7 +382,11 @@ pub struct LightningClientModule {
     module_api: DynModuleApi,
     preimage_auth: KeyPair,
     client_ctx: ClientContext<Self>,
+    gateway_cache_update_mutex: Mutex<Option<LastError>>,
 }
+
+#[derive(Debug, Clone)]
+struct LastError(String);
 
 impl ClientModule for LightningClientModule {
     type Init = LightningClientInit;
@@ -463,6 +468,7 @@ impl LightningClientModule {
                 .to_secp_key(&secp),
             secp,
             client_ctx: args.context(),
+            gateway_cache_update_mutex: Mutex::new(None),
         };
 
         // Only initialize the gateway cache if it is empty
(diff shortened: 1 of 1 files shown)


elsirion (dismissed):
Feels quite complex for something that should happen rather seldomly and then doesn't have a whole lot of downsides (apart from wasted bandwidth)?

Side note: Feels like the GW selection functionality could be encapsulated in its own private struct to separate it from the rest of the module. Not urgent, but many structs grow over time and become unwieldy and we should try to separate functionality wherever possible.

maan2003:
I think it can be quite common after https://github.com/fedimint/fedimint/pull/4766

imagine background is fetching and you do force update at same time (you notice no gateways are avaiable when paying), you also save bandwidth and time on slow networks. time is more important.

joschisan:
In my opinion https://github.com/fedimint/fedimint/pull/4766 is already quite complex; this pr is not worth the complexity imo, the mutex logic is hard to follow.

maan2003:
revived on top of  https://github.com/fedimint/fedimint/pull/4785.
this is much simpler now. 
cc @joschisan

maan2003:
marked as draft because its parent is draft

elsirion (dismissed):

m1sterc001guy (approved):
Yay, so much simpler. I think it just needs a rebase to pickup the nix flake update.

maan2003:
updated

maan2003:
@m1sterc001guy could you also review its parent #4785

elsirion (dismissed):

elsirion (approved):

maan2003:
I hate github, rebased

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -910,7 +916,20 @@ impl LightningClientModule {
     ///
     /// See also [`Self::update_gateway_cache_continuously`].
     pub async fn update_gateway_cache(&self) -> anyhow::Result<()> {
-        let gateways = self.module_api.fetch_gateways().await?;
+        let Ok(mut guard) = self.gateway_cache_update_mutex.try_lock() else {
```
dpc, line 919: I am not sure if this is correct, and it rings a spider-sense, that it might be unnecessarily non-trivial.

As it is RN, e.g. I don't see any `.take()` or error being overwritten with `None` so where is the error ever cleared? 

The chances that this has contention seems rather small, so why not just lock , and not merge?

dpc: Alternatively, if you really, relaly want to merge, then I would do it with an extra `AtomicU64` counter that would get incremented on every success. `load` counter before locking, after locking check if it increased, if it did return `Ok(())`, otherwise let the function try again, as the previous one must have failed. Avoids `e.to_string()` shenigenans which I in particuarlly don't like as they potentially break error chains, etc. which might lead to inconsitent, non-deterministric operation.

m1sterc001guy: Yeah I missed this last time I looked. IMO the gateway cache is just best effort, we shouldn't try to exactly synchronize with what is registered with the federation. I'm in favoring of just locking and avoiding merging.

maan2003:
superseded by #4838

PR #4766, opened 2024-03-29
feat: update ln gateway cache in background

maan2003:
on top of #4760 

## Summary by CodeRabbit

- **New Features**
	- Introduced continuous update functionality for the gateway cache to enhance reliability and performance.
- **Documentation**
	- Added comprehensive documentation for the newly implemented method to update the gateway cache.

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -36,8 +36,9 @@ use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::module::{
     ApiVersion, CommonModuleInit, ModuleCommon, ModuleInit, MultiApiVersion, TransactionItemAmount,
 };
-use fedimint_core::task::{timeout, MaybeSend, MaybeSync};
+use fedimint_core::task::{self, timeout, MaybeSend, MaybeSync};
 use fedimint_core::time::now;
+use fedimint_core::util::{retry, FibonacciBackoff};
 use fedimint_core::{
     apply, async_trait_maybe_send, push_db_pair_items, Amount, OutPoint, TransactionId,
 };
@@ -896,6 +897,8 @@ impl LightningClientModule {
 
     /// Updates the gateway cache by fetching the latest registered gateways
     /// from the federation.
+    ///
+    /// See also [`Self::update_gateway_cache_continuously`].
     pub async fn update_gateway_cache(&self, apply_meta: bool) -> anyhow::Result<()> {
         let gateways = self.fetch_registered_gateways(apply_meta).await?;
         let mut dbtx = self.client_ctx.module_db().begin_transaction().await;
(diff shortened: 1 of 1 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -916,6 +919,55 @@ impl LightningClientModule {
...
+                .into_iter()
+                .map(|x| x.ttl.saturating_sub(ABOUT_TO_EXPIRE))
+                .min();
+
+            let sleep_time = match next_expiry {
+                Some(next_expiry) => {
+                    // found non empty list => reset the retries iterator
+                    // try after 2 seconds first time
+                    empty_gws_retries = std::iter::once(Duration::from_secs(2))
+                        .chain(empty_gws_retry_builder.build());
```
elsirion, line 948: I guess we need this in case a gateway restarts, since it's likely to be back quickly, while some federations could be without one for longer and we don't want to keep polling at a high rate. Still don't like the mutability and complexity here, but don't see a more elegant way either :/

elsirion (approved):

elsirion (approved):

modules/fedimint-ln-client/src/lib.rs
joschisan: I find it hard to figure out the exact strategy from the code itself, which is a problem for something that should be relatively simple. I would very much prefer something simpler here, which might be less optimal, at least for now and see if actually becomes a problem.

joschisan: We can still change this anytime so I want to avoid premature optimisation here.

maan2003: what would simple approach look like?

maan2003: in simple version, i would remove special case for empty gateways. then we will keep polling federation again and again which is horrible on mobile.

other way would be to have constant sleep, and constant sleep < 5-10 minutes is horrible on mobile.

maan2003: made it constant 10'

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -916,6 +919,43 @@ impl LightningClientModule {
...
+    ///
+    /// Client integrators are expected to call this function in a spawned task.
+    pub async fn update_gateway_cache_continuously(&self, apply_meta: bool) -> ! {
+        let mut first_time = true;
+
+        const ABOUT_TO_EXPIRE: Duration = Duration::from_secs(30);
+        const EMPTY_GATEWAY_SLEEP: Duration = Duration::from_secs(10 * 60);
+        loop {
+            let gateways = self.list_gateways().await;
+            // FIXME: filter gateways by vetted
```
maan2003, line 932: will do in follow up, this is fine for now

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -916,6 +919,43 @@ impl LightningClientModule {
...
+                    EMPTY_GATEWAY_SLEEP
+                });
+            task::sleep(sleep_time).await;
+
+            // should never fail with usize::MAX attempts.
+            let _ = retry(
+                "update_gateway_cache",
+                FibonacciBackoff::default()
+                    .with_min_delay(Duration::from_secs(1))
+                    .with_max_delay(Duration::from_secs(2 * 60))
```
maan2003, line 950: what why is this max delay so low, what if federation is down or something

maan2003: changed to 10 minutes max delay, min is still the same

modules/fedimint-ln-client/src/lib.rs
maan2003: changed to TODO

elsirion (approved):
Much simpler :)

dpc (approved):

fedimint-backports:
- #4790

PR #4767, opened 2024-03-29
feat: concurrency in fetch_registered_gateways

maan2003:


maan2003:
I am debating about complexity, but I think it is worth saving one network call worth of latency.
and select logic is easy to understand IMO

maan2003:
closing because we will get meta caching anyways

PR #4768, opened 2024-03-29
feat: meta caching

maan2003:
implement meta caching and updating

left for follow-up:
- implement meta source for the meta module

on top of #4783 

## Summary by CodeRabbit

- **New Features**
	- Enhanced metadata management capabilities, including storing, fetching, and updating metadata with prioritization and legacy support.
	- Implemented a new waiting mechanism for asynchronous task handling.

- **Enhancements**
	- Integrated a new external dependency to improve network request features.

- **Refactor**
	- Updated structures and methods to support metadata management and waiting features.

### DIFF
--- a/fedimint-client/Cargo.toml
+++ b/fedimint-client/Cargo.toml
@@ -42,6 +42,7 @@ thiserror = { workspace = true }
 tokio = { version = "1.36.0", features = [ "time", "macros" ] }
 tokio-stream = { version = "0.1.14", features = [ "time", "sync" ] }
 tracing = { workspace = true }
+reqwest = { version = "0.12.2", features = ["json", "rustls-tls"], default-features = false }
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 ring = { version = "0.17.8", features = ["wasm32_unknown_unknown_js"] }
--- a/fedimint-client/src/db.rs
+++ b/fedimint-client/src/db.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::time::SystemTime;
 
 use fedimint_core::api::ApiVersionSet;
 use fedimint_core::config::{ClientConfig, FederationId};
@@ -39,6 +40,8 @@ pub enum DbKeyPrefix {
     ClientInitState = 0x31,
     ClientMetadata = 0x32,
     ClientLastBackup = 0x33,
+    ClientMetaField = 0x34,
+    ClientMetaServiceInfo = 0x35,
     /// Arbitrary data of the applications integrating Fedimint client and
     /// wanting to store some Federation-specific data in Fedimint client
     /// database.
@@ -295,6 +298,38 @@ impl_db_record!(
     value = ClientBackup,
     db_prefix = DbKeyPrefix::ClientLastBackup
 );
+
+#[derive(Encodable, Decodable, Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct MetaFieldKey(pub String);
+
+#[derive(Encodable, Decodable, Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct MetaFieldPrefix;
+
+#[derive(Encodable, Decodable, Debug, Clone)]
+pub struct MetaFieldValue(pub String);
+
+#[derive(Encodable, Decodable, Debug)]
+pub struct MetaServiceInfoKey;
+
+#[derive(Encodable, Decodable, Debug)]
+pub struct MetaServiceInfo {
+    pub last_updated: SystemTime,
+    pub revision: u64,
+}
+
+impl_db_record!(
+    key = MetaFieldKey,
+    value = MetaFieldValue,
+    db_prefix = DbKeyPrefix::ClientMetaField
+);
+
+impl_db_record!(
+    key = MetaServiceInfoKey,
+    value = MetaServiceInfo,
+    db_prefix = DbKeyPrefix::ClientMetaServiceInfo
+);
+
+impl_db_lookup!(key = MetaFieldKey, query_prefix = MetaFieldPrefix);
 
 /// `ClientMigrationFn` is a function that modules can implement to "migrate"
 /// the database to the next database version.
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -112,6 +112,7 @@ pub use fedimint_derive_secret as derivable_secret;
 use fedimint_derive_secret::DerivableSecret;
 use fedimint_logging::{LOG_CLIENT, LOG_CLIENT_RECOVERY};
 use futures::{Future, Stream, StreamExt};
+use meta::{LegacyMetaSource, MetaService};
 use module::recovery::RecoveryProgress;
 use module::{DynClientModule, FinalClient};
 use rand::thread_rng;
@@ -159,6 +160,9 @@ pub mod secret;
 pub mod sm;
 /// Structs and interfaces to construct Fedimint transactions
 pub mod transaction;
+
+/// Management of meta fields
+pub mod meta;
 
 pub type InstancelessDynClientInput = ClientInput<
     Box<maybe_add_send_sync!(dyn IInput + 'static)>,
@@ -742,6 +746,7 @@ pub struct Client {
     root_secret: DerivableSecret,
     operation_log: OperationLog,
     secp_ctx: Secp256k1<secp256k1_zkp::All>,
+    meta_service: Arc<MetaService>,
 
     task_group: TaskGroup,
 
@@ -977,6 +982,11 @@ impl Client {
 
     pub fn operation_log(&self) -> &OperationLog {
         &self.operation_log
+    }
+
+    /// Get the meta manager to read meta fields.
+    pub fn meta_service(&self) -> &Arc<MetaService> {
+        &self.meta_service
     }
 
     /// Adds funding to a transaction or removes over-funding via change.
@@ -1757,17 +1767,20 @@ pub struct ClientBuilder {
     primary_module_instance: Option<ModuleInstanceId>,
     admin_creds: Option<AdminCreds>,
     db_no_decoders: Database,
+    meta_service: Arc<MetaService>,
     stopped: bool,
 }
 
 impl ClientBuilder {
     fn new(db: Database) -> Self {
+        let meta_service = MetaService::new(LegacyMetaSource::default());
         ClientBuilder {
             module_inits: Default::default(),
             primary_module_instance: Default::default(),
             admin_creds: None,
             db_no_decoders: db,
             stopped: false,
+            meta_service,
         }
     }
 
@@ -1778,8 +1791,11 @@ impl ClientBuilder {
             admin_creds: None,
             db_no_decoders: client.db.with_decoders(Default::default()),
             stopped: false,
+            // non unique
+            meta_service: client.meta_service.clone(),
         }
     }
+
     /// Replace module generator registry entirely
     pub fn with_module_inits(&mut self, module_inits: ClientModuleInitRegistry) {
         self.module_inits = module_inits;
@@ -1808,6 +1824,10 @@ impl ClientBuilder {
             !was_replaced,
             "Only one primary module can be given to the builder."
         )
+    }
+
+    pub fn with_meta_service(&mut self, meta_service: Arc<MetaService>) {
+        self.meta_service = meta_service;
     }
 
     async fn migrate_database(
@@ -2300,7 +2320,19 @@ impl ClientBuilder {
             task_group,
             operation_log: OperationLog::new(db),
             client_recovery_progress_receiver,
+            meta_service: self.meta_service,
         });
+        client_inner
+            .task_group
+            .spawn_cancellable("MetaService::update_continuously", {
+                let client_inner = client_inner.clone();
+                async move {
+                    client_inner
+                        .meta_service
+                        .update_continuously(&client_inner)
+                        .await
+                }
+            });
 
         let client_arc = ClientHandle::new(client_inner);
 
--- a/fedimint-client/src/meta.rs
+++ b/fedimint-client/src/meta.rs
@@ -1,0 +1,286 @@
+use std::collections::BTreeMap;
+use std::sync::Arc;
+use std::time::{Duration, SystemTime};
+
+use anyhow::{bail, Context as _};
+use fedimint_core::db::{Database, DatabaseTransaction, IDatabaseTransactionOpsCoreTyped};
+use fedimint_core::task::waiter::Waiter;
+use fedimint_core::task::{MaybeSend, MaybeSync};
+use fedimint_core::util::{retry, FibonacciBackoff};
+use fedimint_core::{apply, async_trait_maybe_send};
+use serde::de::DeserializeOwned;
+use tracing::{debug, instrument, warn};
+
+use crate::db::{
+    MetaFieldKey, MetaFieldPrefix, MetaFieldValue, MetaServiceInfo, MetaServiceInfoKey,
+};
+use crate::Client;
+
+#[apply(async_trait_maybe_send!)]
+pub trait MetaSource: MaybeSend + MaybeSync + 'static {
+    /// Wait for next change in this source.
+    async fn wait_for_update(&self);
+    async fn fetch(
+        &self,
+        client: &Client,
+        fetch_kind: FetchKind,
+        last_revision: Option<u64>,
+    ) -> anyhow::Result<MetaValues>;
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FetchKind {
+    /// Meta source should return fast, retry less.
+    /// This blocks getting any meta values.
+    Initial,
+    /// Meta source can retry infinitely.
+    Background,
+}
+
+#[derive(Debug, Clone)]
+pub struct MetaValues {
+    values: BTreeMap<MetaFieldKey, MetaFieldValue>,
+    revision: u64,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct MetaValue<T> {
+    pub fetch_time: SystemTime,
+    pub value: Option<T>,
+}
+
+/// Service for managing the caching of meta fields.
+// a fancy DST to save one allocation.
+pub struct MetaService<S: ?Sized = dyn MetaSource> {
+    initial_fetch_waiter: Waiter,
+    source: S,
+}
+
+impl<S: MetaSource + ?Sized> MetaService<S> {
+    pub fn new(source: S) -> Arc<MetaService>
+    where
+        S: Sized,
+    {
+        // implicit cast `Arc<MetaService<S>>` to `Arc<MetaService<dyn MetaSource>>`
+        Arc::new(MetaService {
+            initial_fetch_waiter: Waiter::new(),
+            source,
+        })
+    }
+
+    /// Get the value for the meta field.
+    ///
+    /// This may wait for significant time on first run.
+    pub async fn get_field<V: DeserializeOwned + 'static>(
+        &self,
+        db: &Database,
+        field: &str,
+    ) -> Option<MetaValue<V>> {
+        if let Some(value) = self.get_field_from_db(db, field).await {
+            // might be from in old cache.
+            // TODO: maybe old cache should have a ttl?
+            Some(value)
+        } else {
+            // wait for initial value
+            self.initial_fetch_waiter.wait().await;
+            self.get_field_from_db(db, field).await
+        }
+    }
+
+    async fn get_field_from_db<V: DeserializeOwned + 'static>(
+        &self,
+        db: &Database,
+        field: &str,
+    ) -> Option<MetaValue<V>> {
+        let dbtx = &mut db.begin_transaction_nc().await;
+        let info = dbtx.get_value(&MetaServiceInfoKey).await?;
+        let value = dbtx
+            .get_value(&MetaFieldKey(field.to_string()))
+            .await
+            .and_then(|value| parse_meta_value_static::<V>(&value.0).ok());
+        Some(MetaValue {
+            fetch_time: info.last_updated,
+            value,
+        })
+    }
+
+    async fn current_revision(&self, dbtx: &mut DatabaseTransaction<'_>) -> Option<u64> {
+        dbtx.get_value(&MetaServiceInfoKey)
+            .await
+            .map(|x| x.revision)
+    }
+
+    /// Update all source in background.
+    ///
+    /// Caller should run this method in a task.
+    pub(crate) async fn update_continuously(&self, client: &Client) -> ! {
+        let mut current_revision = self
+            .current_revision(&mut client.db().begin_transaction_nc().await)
+            .await;
+        let meta_values = self
+            .source
+            .fetch(client, FetchKind::Initial, current_revision)
+            .await;
+        let failed_initial = meta_values.is_err();
+        match meta_values {
+            Ok(meta_values) => Self::save_meta_values(client, &meta_values).await,
+            Err(error) => warn!(%error, "failed to fetch source"),
+        };
+        self.initial_fetch_waiter.done();
+
+        // don't wait if we failed first item
+        if !failed_initial {
+            self.source.wait_for_update().await;
+        }
+
+        // now keep updating slowly
+        loop {
+            if let Ok(meta_values) = self
+                .source
+                .fetch(client, FetchKind::Background, current_revision)
+                .await
+            {
+                current_revision = Some(meta_values.revision);
+                Self::save_meta_values(client, &meta_values).await;
+            }
+            self.source.wait_for_update().await;
+        }
+    }
+
+    async fn save_meta_values(client: &Client, meta_values: &MetaValues) {
+        let mut dbtx = client.db().begin_transaction().await;
+        dbtx.remove_by_prefix(&MetaFieldPrefix).await;
+        dbtx.insert_entry(
+            &MetaServiceInfoKey,
+            &MetaServiceInfo {
+                last_updated: fedimint_core::time::now(),
+                revision: meta_values.revision,
+            },
+        )
+        .await;
+        for (key, value) in &meta_values.values {
+            dbtx.insert_entry(key, value).await;
+        }
+        dbtx.commit_tx().await;
+    }
+}
+
+/// Legacy non-meta module config source uses client config meta and
+/// meta_override_url meta field.
+#[derive(Clone, Debug, Default)]
+#[non_exhaustive]
+pub struct LegacyMetaSource {
+    reqwest: reqwest::Client,
+}
+
+#[apply(async_trait_maybe_send!)]
+impl MetaSource for LegacyMetaSource {
+    async fn wait_for_update(&self) {
+        fedimint_core::task::sleep(Duration::from_secs(10 * 60)).await
+    }
+
+    async fn fetch(
+        &self,
+        client: &Client,
+        fetch_kind: FetchKind,
+        last_revision: Option<u64>,
+    ) -> anyhow::Result<MetaValues> {
+        let config_iter = client
+            .config()
+            .global
+            .meta
+            .iter()
+            .map(|(key, value)| (MetaFieldKey(key.to_owned()), MetaFieldValue(value.clone())));
+        let backoff = match fetch_kind {
+            // need to be fast the first time.
+            FetchKind::Initial => FibonacciBackoff::default()
+                .with_min_delay(Duration::from_millis(300))
+                .with_max_delay(Duration::from_secs(3))
+                .with_max_times(10),
+            FetchKind::Background => FibonacciBackoff::default()
+                .with_min_delay(Duration::from_secs(10))
+                .with_max_delay(Duration::from_secs(10 * 60))
+                .with_max_times(usize::MAX),
+        };
+        let overrides = retry("fetch_meta_overrides", backoff, || {
+            fetch_meta_overrides(&self.reqwest, client, "meta_override_url")
+        })
+        .await?;
+        Ok(MetaValues {
+            values: config_iter.chain(overrides).collect(),
+            revision: last_revision.map_or(0, |r| r + 1),
+        })
+    }
+}
+
+pub async fn fetch_meta_overrides(
+    reqwest: &reqwest::Client,
+    client: &Client,
+    field_name: &str,
+) -> anyhow::Result<BTreeMap<MetaFieldKey, MetaFieldValue>> {
+    let Some(url) = client.config().meta::<String>(field_name)? else {
+        return Ok(BTreeMap::new());
+    };
+    let response = reqwest
+        .get(&url)
+        .send()
+        .await
+        .context("Meta override source could not be fetched")?;
+
+    debug!("Meta override source returned status: {response:?}");
+
+    if response.status() != reqwest::StatusCode::OK {
+        bail!(
+            "Meta override request returned non-OK status code: {}",
+            response.status()
+        );
+    }
+
+    let mut federation_map = response
+        .json::<BTreeMap<String, BTreeMap<String, serde_json::Value>>>()
+        .await
+        .context("Meta override could not be parsed as JSON")?;
+
+    let federation_id = client.federation_id().to_string();
+    let meta_fields = federation_map
+        .remove(&federation_id)
+        .with_context(|| anyhow::format_err!("No entry for federation {federation_id} in {url}"))?
+        .into_iter()
+        .filter_map(|(key, value)| match value {
+            serde_json::Value::String(value_str) => {
+                Some((MetaFieldKey(key), MetaFieldValue(value_str)))
+            }
+            _ => {
+                warn!("Meta override map contained non-string key: {key}, ignoring");
+                None
+            }
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    Ok(meta_fields)
+}
+
+/// Tries to parse `str_value` as JSON. In the special case that `V` is `String`
+/// we return the raw `str_value` if JSON parsing fails. This necessary since
+/// the spec wasn't clear enough in the beginning.
+#[instrument(err)] // log on every failure
+pub fn parse_meta_value_static<V: DeserializeOwned + 'static>(
+    str_value: &str,
+) -> anyhow::Result<V> {
+    let res = serde_json::from_str(str_value)
+        .with_context(|| format!("Decoding meta field value '{str_value}' failed"));
+
+    // In the past we encoded some string fields as "just a string" without quotes,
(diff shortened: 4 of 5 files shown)


fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+#[apply(async_trait_maybe_send!)]
+pub trait MetaSource: MaybeSend + MaybeSync {
+    /// Higher value overwrites sources with lower priority value.
+    fn priority(&self) -> u64;
+    /// Wait for next change in this source.
+    async fn wait_for_update(&self);
+    async fn fetch(&self, client: &Client) -> anyhow::Result<BTreeMap<String, String>>;
```
elsirion, line 54: Handing in the client is very smart! That way there can be a source using a module mounted before the client is constructed :eyes:

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub async fn get_field<V: DeserializeOwned + 'static>(
+        &self,
+        db: &Database,
+        field: &str,
+    ) -> Option<V> {
+        // TODO: maybe wait for initial sources
```
elsirion, line 72: :+1:

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,244 @@
...
+            .await
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .find_map(|(_, value)| parse_meta_value_static::<V>(&value).ok())
+    }
+
+    pub(crate) fn add_source(&mut self, source: impl MetaSource + 'static) {
+        Arc::get_mut(&mut self.sources)
+            .expect("add_source on non unique MetaManager")
```
elsirion, line 85: This cannot happen because `add_source` is only used before the client is constructed, thus before one could access the `MetaManager` from outside?

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,244 @@
...
+                )
+                .await
+                .context("timeout while fetch source"),
+            )
+        }))
+        .await;
+        // intially write all once
+        let mut dbtx = client.db().begin_transaction().await;
+        for (source, meta) in metas {
+            if let Ok(meta) = meta.and_then(|x| x) {
```
elsirion, line 115: TIL: `Result::flatten` is nightly-only

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,244 @@
...
+        1
+    }
+
+    async fn wait_for_update(&self) {
+        // update every 10 minutes
+        fedimint_core::task::sleep(Duration::from_secs(10 * 60)).await
+    }
+
+    async fn fetch(&self, client: &Client) -> anyhow::Result<BTreeMap<String, String>> {
+        let Some(url) = client.config().meta::<String>("TODO")? else {
```
elsirion, line 207: Let's make the field name a parameter stored in the struct. The default constructor would use the standardized `meta_override_url`, but for compatibility reasons the pre-standardization variants could also be used.

fedimint-client/src/meta.rs
elsirion: Could we use DB notifications here? I guess looking for *any priority* is a problem here.

fedimint-client/src/meta.rs
maan2003: moved this expect to client builder to make logic more local

fedimint-client/src/lib.rs
```diff
@@ -1810,6 +1824,18 @@ impl ClientBuilder {
...
 
+    pub fn with_meta_source(&mut self, meta_source: impl MetaSource + 'static) {
+        Arc::get_mut(&mut self.meta_manager)
+            // Self::from_existing is private, so should never happen
+            .expect("with_meta_source when meta manager is non unique")
+            .add_source(meta_source);
+    }
+
+    /// NOTE: [`Self::with_meta_source`] may fail if this meta_manager is not unique.
+    pub fn with_meta_manager(&mut self, meta_manager: Arc<MetaManager>) {
```
maan2003, line 1835: this method allows you to even more default meta sources.

For example you can wrap the default config source to modify some meta values.

fedimint-client/Cargo.toml
```diff
@@ -42,6 +42,7 @@ thiserror = { workspace = true }
 tokio = { version = "1.36.0", features = [ "time", "macros" ] }
 tokio-stream = { version = "0.1.14", features = [ "time", "sync" ] }
 tracing = { workspace = true }
+reqwest = { version = "0.12.2", features = ["json", "rustls-tls"] }
```
elsirion, line 45: You'll have to deactivate default features here so that OpenSSL isn't included.

maan2003: oh right, I have become so lazy with `cargo add` now

fedimint-client/src/meta.rs
maan2003: we don't have notifications for database prefixes :/

maan2003: added using Waiter from #4783

fedimint-client/src/lib.rs
```diff
@@ -2300,7 +2327,19 @@ impl ClientBuilder {
             task_group,
             operation_log: OperationLog::new(db),
             client_recovery_progress_receiver,
+            meta_manager: self.meta_manager,
         });
+        client_inner
```
maan2003, line 2332: should this happen automatically?

fedimint-client/src/lib.rs
```diff
@@ -2300,7 +2327,19 @@ impl ClientBuilder {
...
+            .task_group
+            .spawn_cancellable("MetaManager::update_continuously", {
+                let client_inner = client_inner.clone();
+                async move {
+                    client_inner
+                        .meta_manager
+                        .update_continuously(&client_inner)
+                        .await
+                }
+            });
```
maan2003, line 2342: should we do this automatically?

fedimint-client/src/lib.rs
```diff
@@ -112,6 +112,7 @@ pub use fedimint_derive_secret as derivable_secret;
 use fedimint_derive_secret::DerivableSecret;
 use fedimint_logging::{LOG_CLIENT, LOG_CLIENT_RECOVERY};
 use futures::{Future, Stream, StreamExt};
+use meta::{LegacyClientConfigMetaSource, LegacyMetaOverrideUrlSource, MetaManager, MetaSource};
```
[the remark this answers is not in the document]
maan2003: you should write the docs instead of complaining :smile:

fedimint-client/Cargo.toml
```diff
@@ -42,6 +42,7 @@ thiserror = { workspace = true }
 tokio = { version = "1.36.0", features = [ "time", "macros" ] }
 tokio-stream = { version = "0.1.14", features = [ "time", "sync" ] }
 tracing = { workspace = true }
+reqwest = { version = "0.12.2", features = ["json", "rustls-tls"], default-features = false }
```
dpc, line 45: :/ I wonder how heavy it is.

fedimint-client/src/lib.rs
```diff
@@ -742,6 +746,7 @@ pub struct Client {
     root_secret: DerivableSecret,
     operation_log: OperationLog,
     secp_ctx: Secp256k1<secp256k1_zkp::All>,
+    meta_manager: Arc<MetaManager>,
```
dpc, line 749: I'm not a fan of managers. :D . Would name with something more concrete.

fedimint-client/src/lib.rs
```diff
@@ -2145,32 +2177,34 @@ impl ClientBuilder {
...
-                                        .await
-                                        .map_err(|err| {
-                                            warn!(
-                                                module_id = module_instance_id, %kind, %err, "Module failed to recover"
-                                            );
-                                            err
-                                        })
+                                        snapshot.as_ref().and_then(|s| s.modules.get(&module_instance_id).to_owned()),
+                                        progress_tx,
+                                        meta_manager,
```
dpc, line 2199: "meta" was supposed to be extra functionality for end apps, why would models now need to depend on it?

dpc (commented):
With meta module I wanted to simplify meta handling, not hire an extra manager... :/

maan2003:
in 0.5 maybe we will fire the manager

fedimint-client/src/lib.rs
maan2003: @coderabbitai please suggest a better name

fedimint-client/src/lib.rs
dpc: That whole LN depending on Meta seems like a potential mistake. I thought we already moved active gateway out of client to the user, should vetted gateways be kind of the same?

Also - if we do want to reuse meta for it, maybe we should use a key `1` for "LN-meta" as opposed to default `0`, so they can be voted, changed and fetched independently?

dpc:
> in 0.5 maybe we will fire the manager

In my experience it's very hard to fire a manager. They tend to fire you instead, and then multiply.

The moment you introduce a first manager, you set in motion a fault-tolerant system that is impossible to stop, and you get Manager Directory and VP really quick.

Just saying...

fedimint-client/src/db.rs
```diff
@@ -296,6 +297,23 @@ impl_db_record!(
...
+    pub name: MetaFieldName,
+    pub priority: u64,
+}
+
+#[derive(Encodable, Decodable, Debug)]
+pub struct MetaFieldName(pub String);
+
+impl_db_record!(
+    key = MetaFieldKey,
+    value = String,
```
dpc, line 311: I don't see `revision` being used anywhere. I specifically introduced monotonically incrementing revisions to make checking for updates more lightweight. Would be good to use it.

fedimint-client/src/lib.rs
maan2003: fixing it slightly in https://github.com/fedimint/fedimint/pull/4785

now with meta manager, it should be much easier for client impls.

fedimint-client/Cargo.toml
maan2003: we already use reqwest on client side, through jsonrpsee

fedimint-client/src/lib.rs
maan2003: removed accesss to meta manager from module args

fedimint-client/src/db.rs
maan2003: will do in a follow-up, along with subscription for changes in meta. this is blocking some PR.

we don't promise database compat on master, so we won't need any migration

maan2003:
what is worse meta managers or Meta services? 

fedimint now has Meta services integrated for telemetry :smile:

fedimint-client/src/db.rs
maan2003: doing in https://github.com/fedimint/fedimint/pull/4787

maan2003: added revision in this PR, because this doesn't have enough approvals yet :smile:

maan2003:
changes: priority is not for source, but for values now

maan2003:
found a bug, sources can never remove values from meta

fedimint-client/Cargo.toml
elsirion: It's also already used by the LN module to do the same

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    pub async fn get_field<V: DeserializeOwned + 'static>(
+        &self,
+        db: &Database,
+        field: &str,
+    ) -> Option<TypedMetaValue<V>> {
+        if let Some(value) = self.get_field_inner(db, field).await {
+            // might be from in old cache.
+            // TODO: maybe old cache should have a ttl?
```
elsirion, line 86: I think it's better to return an outdated value than none here.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,351 @@
...
+    pub revision: Option<u64>,
+    pub priority: u64,
+}
+
+impl MetaService {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub async fn get_field<V: DeserializeOwned + 'static>(
```
elsirion, line 79: We should document that this can block for a significant amount of time on first run (currently for around 60s).

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,351 @@
...
+            .global
+            .meta
+            .iter()
+            .map(|(key, value)| {
+                (
+                    MetaFieldName(key.to_owned()),
+                    MetaFieldValue {
+                        value: value.clone(),
+                        revision: Some(0), // config is constant
+                        priority: LEGACY_CLIENT_CONFIG_PRIORITY,
```
elsirion, line 248: Giving each value a priority chosen by the source seems like over-engineering given that we have quite a clear precedence that is dependent on the source (2 legacy, one current). I don't think we'll ever have more than one non-legacy source.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,351 @@
...
+pub struct LegacyClientConfigMetaSource {}
+
+pub const LEGACY_CLIENT_CONFIG_PRIORITY: u64 = 0;
+pub const LEGACY_CLIENT_CONFIG_SOURCE_ID: u64 = 0;
+
+#[apply(async_trait_maybe_send!)]
+impl MetaSource for LegacyClientConfigMetaSource {
+    fn source_id(&self) -> u64 {
+        LEGACY_CLIENT_CONFIG_SOURCE_ID
+    }
```
elsirion, line 227: The `source_id` not just being a counter means we need to coordinate the values when creating new sources, which is slightly annoying (but that won't happen too often I hope).

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,351 @@
...
+                        revision: None,
+                        priority: self.priority,
+                    },
+                )),
+                None => {
+                    warn!("Meta override map contained non-string key: {key}, ignoring");
+                    None
+                }
+            })
+            .collect::<BTreeMap<_, _>>();
```
elsirion, line 347: I think you forgot to take into account here that there can be multiple federation entries in a `meta.json`, thus we have to first get the entry corresponding to our federation (might be my mistake if you copied it, ran into it with `fedimint-observer` too).

fedimint-client/src/meta.rs
maan2003: but we can also wait for initial fetch like in None case

fedimint-client/src/meta.rs
maan2003: fixed it

fedimint-client/src/meta.rs
maan2003: we can say the order of adding source determines the source id and priority, you have to keep it consistent and you can't remove any source in future

fedimint-client/src/meta.rs
maan2003: fixed

fedimint-client/src/meta.rs
elsirion: Well, that's true, integrators will have to deal with this fn potentially blocking anyway …

fedimint-client/src/meta.rs
maan2003: fixed

fedimint-client/src/meta.rs
elsirion: Good point, or these old DB entries will never be invalidated …

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,371 @@
...
+            self.sources
+                .iter()
+                .map(|source| self.update_source_continuously(client, &**source)),
+        )
+        .await;
+        // to get ! type
+        values[0]
+    }
+
+    async fn update_source_continuously(&self, client: &Client, source: &dyn MetaSource) -> ! {
```
elsirion, line 171: Since this is *always* started now running it twice doesn't sound too useful, thoughts about making the method private?

elsirion (approved):

fedimint-client/src/meta.rs
maan2003: it is private :thinking: , you mean the other one

maan2003: fixed

fedimint-client/src/lib.rs
```diff
@@ -979,6 +984,11 @@ impl Client {
         &self.operation_log
     }
 
+    /// Get the meta manager to read meta fields.
+    pub fn meta_service(&self) -> Arc<MetaService> {
+        self.meta_service.clone()
+    }
+
```
elsirion, line 991: Nit: do we need to always clone?

fedimint-client/src/meta.rs
elsirion: Oh yeah, but fixed I see :sweat_smile:

elsirion (approved):

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    pub fn add_source(&mut self, source: impl MetaSource + 'static) {
+        self.sources.push(Box::new(source))
+    }
+
+    /// Update all source in background.
+    ///
+    /// Caller should run this method in a task.
+    pub(crate) async fn update_continuously(&self, client: &Client) -> ! {
+        // need to be fast the first time.
```
dpc, line 130: Even if we have them already in the database and they are reasonably fresh?

fedimint-client/src/lib.rs
```diff
@@ -1757,17 +1767,23 @@ pub struct ClientBuilder {
     primary_module_instance: Option<ModuleInstanceId>,
     admin_creds: Option<AdminCreds>,
     db_no_decoders: Database,
+    // unique Arc unless from_existing is called.
+    meta_service: Arc<MetaService>,
```
dpc, line 1771: Why would a builder have meta service, when it's unknown what the config is going to be an Federation. Seems to me like this should be create during actually `build`, when the config is known etc.

fedimint-client/src/lib.rs
```diff
@@ -1810,6 +1829,19 @@ impl ClientBuilder {
...
     }
 
+    pub fn with_meta_source(&mut self, meta_source: impl MetaSource + 'static) {
+        Arc::get_mut(&mut self.meta_service)
+            // Self::from_existing is private, so should never happen
+            .expect("with_meta_source when meta manager is non unique")
+            .add_source(meta_source);
+    }
+
+    /// NOTE: [`Self::with_meta_source`] may fail if this meta_service is not
```
dpc, line 1839: It doesn't look like it returns `Result` or panics, so I don't understand.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    /// Get the value for the meta field.
+    ///
+    /// This may wait for significant time on first run.
+    pub async fn get_field<V: DeserializeOwned + 'static>(
+        &self,
+        db: &Database,
+        field: &str,
+    ) -> Option<TypedMetaValue<V>> {
+        if let Some(value) = self.get_field_inner(db, field).await {
```
dpc, line 89: _inner in this context is not informative.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+            // TODO: maybe old cache should have a ttl?
+            Some(value)
+        } else {
+            // wait for initial value
+            self.initial_fetch_waiter.wait().await;
+            self.get_field_inner(db, field).await
+        }
+    }
+
+    async fn get_field_inner<V: DeserializeOwned + 'static>(
```
dpc, line 100: `get_stored_fiel_value`? I guess it's about getting it from the copy saved in the the database?

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+            .find_map(|(key, value)| {
+                Some(TypedMetaValue {
+                    revision: value.revision,
+                    value: parse_meta_value_static::<V>(&value.value).ok()?,
+                    priority: key.priority,
+                })
+            })
+    }
+
+    pub fn add_source(&mut self, source: impl MetaSource + 'static) {
```
dpc, line 122: IMO, mutability here is asking for troubles. The whole thing should start with all sources known and then not require/support modifications. Possibly a builder can be used to assemble the sources, then convert to a ready service.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+                Self::update_meta_values(&mut dbtx, source.priority(), source.source_id(), &meta)
+                    .await;
+                // TODO: what should have happen in case of conflicts?
+                // should we ignore the conflict or auto commit
+                let _ = dbtx.commit_tx_result().await;
+            }
+        }
+    }
+
+    async fn update_meta_values(
```
dpc, line 195: `save_`? `store_`? "update" here is vague. "update" as in, fetch + store / fetch, etc.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+        dbtx: &mut DatabaseTransaction<'_, Committable>,
+        priority: u64,
+        source_id: &'static str,
+        meta: &BTreeMap<MetaFieldName, MetaFieldValue>,
```
dpc, line 199: I don't know if this "meta is a map of fieldname => fielvalue" model is a good idea. Intuitively I would just fetch, cache, store, etc the whole thing as `serde_json::Value` and not mess with it much.

IMO. If `meta` module is used, the other methods of meta should be just ignored. So why do we split them by fields, and keep updating everything all the time, etc? New Federations should just set meta in the module.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+    ) {
+        let stream = dbtx.find_by_prefix(&MetaFieldPrefix).await;
+
+        let to_remove = stream
```
dpc, line 203: E.g. this (filtering by source_id) would be unnecessary if we just stored the whole thing.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,375 @@
...
+            }
+        }
+        dbtx.commit_tx().await;
+        self.initial_fetch_waiter.done();
+
+        // now keep updating all
+        let values = join_all(
+            self.sources
+                .iter()
+                .map(|source| self.update_source_continuously(client, &**source)),
```
dpc, line 164: If meta module is used, what bother continuously updating other sources?

dpc:
In my mind if the Federation config has a module with `meta` kind, the client can just ignore other metas. Merging anything should be avoided.

I bet the reason for all complexity here is because people are really set on reusing meta module for this gateway vetting, which doesn't really fit here. LN module should keep track of it's vetted gateway on its own.

It creates dependencies between modules, will have a different lifecycle, etc.

Eg. LN module can just have an admin endpoint for guardians to vet/ban gateways, and return such info on a dedicated LN-module endpoint. Clients can fetch and cache such list when they need it. So if you don't make LN payements, you don't need to check anything every 10 minutes. All sorts of smarter and more agile things can be done then - e.g. count number of guardians vetting things, require different amount of thresholds, etc. While meta module is specifically for slow-moving meta information. All without any inter-module dependencies.

@elsirion @maan2003

fedimint-client/src/lib.rs
maan2003: so you can add source before building, we could have `MetaServiceBuilder` but that doesn't work with `ClientBuilder::from_existing`

fedimint-client/src/meta.rs
maan2003: given `update_continuously` is `(&self) -> !` it forces the manager to be immutable

maan2003:
> LN module should keep track of it's vetted gateway on its own.

too late :/
we already cut a 0.3 with vetted gateways in meta

maan2003:
wasn't meta module meant as storage for other modules

we even thought of storing registered gateways in meta module

fedimint-client/src/meta.rs
maan2003: maybe database entry should last fetch time

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,263 @@
...
+    Initial,
+    /// Meta source can retry infinitely.
+    Background,
+}
+
+/// Service for managing the caching of meta fields.
+// a fancy DST to save one allocation.
+pub struct MetaService<S: ?Sized = dyn MetaSource> {
+    initial_fetch_waiter: Waiter,
+    source: S,
```
maan2003, line 42: I am satisfied as a rust programmer :smile:

elsirion:
> In my mind if the Federation config has a module with `meta` kind, the client can just ignore other metas. Merging anything should be avoided.

That's a good point actually. Initially my idea was that we want a smooth transition, but that's not actually the case before 0.4 the earliest: new federations will use the meta module while old ones have to continue using the override file.

If we were to automatically add the meta module to existing federations having the merge behavior would be important imo to not accidentally remove all the meta fields from a user's pov.

> I bet the reason for all complexity here is because people are really set on reusing meta module for this gateway vetting

I don't think that has much to do with it, just the thought of requiring two client restarts to get the latest meta values is unappealing (first: fetch values on start, but still return old ones to not block, second: actually return new ones). Then the natural conclusion is we need an update in the background that can communicate to the frontend once new values have arrived. At that point, why not just let the updater run continuously for slightly more timely changes on the client side (it's basically free compared to the other shenanigans we do in state machines).

elsirion:
That observation that vetted gateways don't need consensus is correct imo, could just be a per-guardian endpoint and then the client counts which GW comes out on top.

dpc:
> we already cut a 0.3 with vetted gateways in meta

Aaaa! :D

> wasn't meta module meant as storage for other modules

No... As a name of the module suggest, it is supposed to be a replacement for "metadata" about Federation that guardians can update over time, without using centralized place.

We probably have put LN vetting gateways in external meta, as a hack "because it can be easily updated".

Now hacks get married, have kids, the kids get ID, voting rights...

> we even thought of storing registered gateways in meta module

I don't even...

> new federations will use the meta module while old ones have to continue using the override file.

Note: even if existing Federations could add meta module, they could just put existing meta into meta module and thus switch right into it atomically. Then after some time (Clients get to see it) just delete the whole external module (can't delete the one in config, I guess, but otherwise also could be done).

> If we were to automatically add the meta module to existing federations having the merge behavior would be important imo to not accidentally remove all the meta fields from a user's pov.

Meta module keys start with uninitialized value, so even right after being addded (if it was possible), the client can tell that nothing was set yet and fallaback to previous sources.

> I don't think that has much to do with it, just the thought of requiring two client restarts to get the latest meta values is unappealing (first: fetch values on start, but still return old ones to not block, second: actually return new ones). 

I was just thinking yesterday, that we could add to `ClientModuleInit` an ability to flag that they don't ever need any recover and thus can be initialized immediately, avoiding the restart after the recovery.

Either a `fn ClientModuleInit::needs_recovery() -> bool` or just `fn ClientModuleINit::recover(...) -> Option<impl Future<...>>` where `None` means, I'm done, no need to wait.

> it's basically free compared to the other shenanigans we do in state machines

Most state machines should be chatty for a couple of minutes of activity and wrap it up, or go to a long, long sleep. Otherwise people will complain about batter life, and that might matter a lot in rural areas on second hand old phones.

> That observation that vetted gateways don't need consensus is correct imo, could just be a per-guardian endpoint and then the client counts which GW comes out on top.

If e.g. a gateway missbehaves it might take a really inconvenient time to gather threshold amount of votes to change the meta keys, while

elsirion (approved):
~I think this PR is sane enough, I mainly want to get some standardized way of fetching meta into 0.3.1 so we don't see everyone doing something randomly different. Most of the internal workings can be adapted if need be.~

Scrap that, apparently these comments weren't outdated but I got interrupted by some call :sob:

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,263 @@
...
+            Err(error) => warn!(%error, "failed to fetch source"),
+        };
+        self.initial_fetch_waiter.done();
+
+        // now keep updating slowly
+        loop {
+            // don't wait if we failed first item
+            if !failed_initial {
+                MetaSource::wait_for_update(&self.source).await;
+            }
```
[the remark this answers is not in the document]
dpc: Could just keep `previous_value : Option<...>`, or if we're worried about it wasting space `fetched_ok_before: bool`, avoiding negation.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,262 @@
...
+    reqwest: reqwest::Client,
+}
+
+#[apply(async_trait_maybe_send!)]
+impl MetaSource for LegacyMetaSource {
+    async fn wait_for_update(&self) {
+        fedimint_core::task::sleep(Duration::from_secs(10 * 60)).await
+    }
+
+    async fn fetch(
```
dpc, line 148: fetch, should take `existing_revision: Option<u64>`, so in case of meta module, it can issue `check-revision` and 99.99% avoid downloading everything again.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,262 @@
...
+use tracing::{debug, instrument, warn};
+
+use crate::db::{MetaFieldKey, MetaFieldPrefix, MetaFieldValue};
+use crate::Client;
+
+#[apply(async_trait_maybe_send!)]
+pub trait MetaSource: MaybeSend + MaybeSync + 'static {
+    /// Wait for next change in this source.
+    async fn wait_for_update(&self);
+    async fn fetch(
```
dpc, line 21: `fetch`, should take `existing_revision: Option<u64>`, so in case of meta module, it can issue `check-revision` and 99.99% avoid downloading everything again.

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,262 @@
...
+        &self,
+        client: &Client,
+        fetch_kind: FetchKind,
+    ) -> anyhow::Result<BTreeMap<MetaFieldKey, MetaFieldValue>>;
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FetchKind {
```
dpc, line 29: Why make it an implementation concern? Can't caller of `fetch` deal with retrying?

dpc (commented):
The implementation is smaller and cleaner now. Or at least that's my impression. Some minor, but not unimportant details, rest LGTM, for now.

fedimint-client/src/meta.rs
maan2003: revision is per field right?

fedimint-client/src/meta.rs
elsirion: :+1: but can be done in 0.4 if it's a big concern. Your typical meta for one federation is rather small. To really implement it I'd like to add the previous revision to `GET_CONSENSUS_ENDPOINT`, so we don't need to make two calls but the server can decide if it should send us anything.

fedimint-client/src/meta.rs
maan2003: if you have two network calls in fetch, one succeeded you don't want to retry that one. it is consequence of having just a single meta source

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,239 @@
...
+            FetchKind::Background => FibonacciBackoff::default()
+                .with_min_delay(Duration::from_secs(10))
+                .with_max_delay(Duration::from_secs(10 * 60))
+                .with_max_times(usize::MAX),
+        };
+        let overrides = retry("fetch_meta_overrides", backoff, || {
+            fetch_meta_overrides(&self.reqwest, client, "meta_override_url")
+        })
+        .await?;
+        config_meta.extend(overrides);
```
elsirion, line 164: nit: avoid mutability:

```rust
config_meta.into_iter()
    .chain(overrides)
    .collect()
```

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        // now keep updating slowly
+        loop {
+            if let Ok(meta) = self.source.fetch(client, FetchKind::Background).await {
+                Self::save_meta_values(client, &meta).await;
+            }
+            self.source.wait_for_update().await;
+        }
+    }
```
elsirion, line 111: Isn't there still the problem that `failed_initial` never gets reset and a `fetch` impl that doesn't have some exp backoff internally will just run this loop at max speed?

elsirion (approved):

maan2003:
significantly changed revision see: https://github.com/fedimint/fedimint/compare/c2079950dba1f86145e1b5554b4bf421bb0b3818..d61300fb2b05c49be84fa9b9e0ac093e85d90b25

fedimint-client/src/meta.rs
maan2003: [fixed](https://github.com/fedimint/fedimint/compare/d61300fb2b05c49be84fa9b9e0ac093e85d90b25..6697b722b063fd9a25664c55f9e2dead6774a9a7)

fedimint-client/src/meta.rs
maan2003: done

fedimint-client/src/meta.rs
```diff
@@ -0,0 +1,286 @@
...
+                .with_max_delay(Duration::from_secs(10 * 60))
+                .with_max_times(usize::MAX),
+        };
+        let overrides = retry("fetch_meta_overrides", backoff, || {
+            fetch_meta_overrides(&self.reqwest, client, "meta_override_url")
+        })
+        .await?;
+        Ok(MetaValues {
+            values: config_iter.chain(overrides).collect(),
+            revision: last_revision.map_or(0, |r| r + 1),
```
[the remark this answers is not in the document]
maan2003: we can't store old values anywhere. MetaSources are meant to be mostly stateless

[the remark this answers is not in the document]
elsirion: Right, from inside there we don't have access to the current values.

fedimint-client/src/meta.rs
dpc: > Your typical meta for one federation is rather small.

The limit is 1MB.

Did you see the legal disclaimer about joining the pop-up federation? :D . That thin along makes me think maybe I should have made it 16MB.

I also expect people will want some kind of images sometimes.

And since you really want to continuously poll it, and it is supposed to work in rural African communities, probably worth it already, at least for some people.

dpc (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-4768-to-releases/v0.3 origin/releases/v0.3
cd .worktree/backport-4768-to-releases/v0.3
git switch --create backport-4768-to-releases/v0.3
git cherry-pick -x 6697b722b063fd9a25664c55f9e2dead6774a9a7
```

Issue #4777, opened 2024-03-30
"nix develop" command is incomplete to enter into the developer environment to locally run fedimint


star-gazer111:
`nix develop` command as stated [here](https://github.com/fedimint/fedimint?tab=readme-ov-file#local-development) doesn't takes into the developer environment throwing the following error: 

`error: experimental Nix feature 'nix-command' is disabled; add '--extra-experimental-features nix-command' to enable it`

This error is not a build error in the fedimint rather a customisation issue while building through Nix. The reason for this “weirdness” I guess is that the official command is still nix-build, which will be replaced in the future by nix build, but this command is not considered stable yet. 

Following the Nix manual & the error message thrown by Nix itself, the command that comes out to work is this:
`nix --extra-experimental-features nix-command develop`

but rather this will pop the following error:
`error: experimental Nix feature 'flakes' is disabled; add '--extra-experimental-features flakes' to enable it`

The experimental-features config option is mentioned in the nix manual about the nix.conf:

[https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-experimental-features 1.5k](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-experimental-features)

The nixos option in the option search:

[https://search.nixos.org/options?channel=22.11&show=nix.settings&from=0&size=50&sort=relevance&type=packages&query=nix.settings 282](https://search.nixos.org/options?channel=22.11&show=nix.settings&from=0&size=50&sort=relevance&type=packages&query=nix.settings)

Which exact values to add or which experimental features are available is not documented in either. The most commonly used are nix-command and flakes, though there are also ca-derivations (content addressed derivations), impure-derivations and recursive-nix (allows nix being used in a derivation).

So rather, the command that will work perfectly and directly while building on almost any system is this: 

`nix --extra-experimental-features nix-command --extra-experimental-features flakes develop`

Just added this so a newcomer doesn't have to hussle with the NIx customisations and rather directly focus on local development & building.

star-gazer111:
The option to enable flakes is actually mentioned [here](https://github.com/fedimint/fedimint/blob/master/docs/dev-env.md#enable-nix-flakes) but I think the same can be directly reduced to a single command and mentioned in the main README.md .

star-gazer111:
Fix here[4778](https://github.com/fedimint/fedimint/pull/4778)

PR #4778, opened 2024-03-30
Add flakes to the final command to directly enter into the developer …

star-gazer111:
…environment while building locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ git clone https://github.com/your-username/fedimint.git
 Then enter the nix developer environment.
 
 ```bash
-nix develop
+nix --extra-experimental-features nix-command --extra-experimental-features flakes develop
 ```
 
 and start a local regtest network with 4 guardians and a lightning gateway.


star-gazer111:
[Fix](https://github.com/fedimint/fedimint/issues/4777)

README.md
```diff
@@ -77,7 +77,7 @@ git clone https://github.com/your-username/fedimint.git
 Then enter the nix developer environment.
 
 ```bash
-nix develop
+nix --extra-experimental-features nix-command --extra-experimental-features flakes develop
```
dpc, line 80: We should change the installer recommendation to one from https://github.com/DeterminateSystems/nix-installer , so the users don't need to worry about enabling flakes. It is generally considered much better than the official one, and

star-gazer111: Hmm, that's even better I suppose. So should I make the appropriate changes in the Nix Installation section?

dpc: Go for it. Thanks!

PR #4780, opened 2024-03-30
fix: base64 url safe for display

Kodylow:
For decode we're handling both cases but for encode we're writing stand b64, should write as url safe

### DIFF
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -320,11 +320,20 @@ impl FromStr for OOBNotes {
 }
 
 impl Display for OOBNotes {
-    /// Base64UrlSafe encode a set of e-cash notes for out-of-band spending.
+    /// Base64 encode a set of e-cash notes for out-of-band spending.
+    ///
+    /// Defaults to standard base64 for backwards compatibility.
+    /// For URL-safe base64 as alternative display use:
+    /// `format!("{:#}", oob_notes)`
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         let mut bytes = Vec::new();
         Encodable::consensus_encode(self, &mut bytes).expect("encodes correctly");
-        f.write_str(&BASE64_URL_SAFE.encode(&bytes))
+
+        if f.alternate() {
+            f.write_str(&BASE64_URL_SAFE.encode(&bytes))
+        } else {
+            f.write_str(&base64::engine::general_purpose::STANDARD.encode(&bytes))
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


dpc:
> For decode we're handling both cases but for encode we're writing stand b64, should write as url safe

IIRC I did it this way to let clients in the wild get support for new encoding before we start to actually create it.

I think I'm fine with doing it on master now, but we should not backport it to 0.3.

dpc (approved):

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -295,11 +295,11 @@ impl FromStr for OOBNotes {
 }
 
 impl Display for OOBNotes {
-    /// Base64 encode a set of e-cash notes for out-of-band spending.
+    /// Base64UrlSafe encode a set of e-cash notes for out-of-band spending.
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         let mut bytes = Vec::new();
         Encodable::consensus_encode(self, &mut bytes).expect("encodes correctly");
-        f.write_str(&base64::engine::general_purpose::STANDARD.encode(&bytes))
+        f.write_str(&BASE64_URL_SAFE.encode(&bytes))
```
elsirion, line 302: I think we should still allow alternate encoding the legacy format https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.alternate

justinmoon:
dev call: when did we add ability to parse url safe base 64? we can't deprecate non-url safe encodings if they are out in the wild.

dpc:
> when did we add ability to parse url safe base 64?

IIRC it will really be out in 0.3, as it was added post 0.2

> we can't deprecate non-url safe encodings if they are out in the wild.

We are not deprecating anything yet, and we don't really need to anytime soon. 

This PR will change the encoding of invitations etc. produced by this version.

elsirion:
@dpc that would mean 0.4 e-cash notes and invites could not be parsed by 0.2 clients. There should be at least a way to allow integrators to easily keep compatibility in cases where URL-safety isn't needed anyway.

dpc:
> @dpc that would mean 0.4 e-cash notes and invites could not be parsed by 0.2 clients.

Yes.

> There should be at least a way to allow integrators to easily keep compatibility in cases where URL-safety isn't needed anyway.

I don't understand.

But if we want 0.2 clients to keep parsing stuff OK, we should just keep the default serialization to be as it is longer. It's a cosmetic/UX thing only anyway. All new clients roll with an ability to deserialize both new and old format.


Decoding for url-safe notes landed in:

```
commit fc0a7c6c9331c0a1553419b03390f092c4f5a4d0 (o/24-02-06-oob-base64, 24-02-06-oob-base64)
Author: Dawid Ciężarkiewicz <dpc@dpc.pw>
Date:   Tue Feb 6 23:28:23 2024 -0800
```

so not that long ago.

elsirion:
Let's do this:
* Let's stick with base64 as default
* Add an alt formatting that uses base64urlsafe, that can be used by whoever wants and doesn't care about compatibility

elsirion (approved):

PR #4781, opened 2024-03-30
fix: typos

Kodylow:
Typos wasn't catching these inside nix develop, when I run `typos` in normal shell it catches these.


dpc:
You might have a newer `typos` than our dev shell. We'll bump it eventually.

elsirion (approved):

PR #4783, opened 2024-03-31
feat: add Waiter helper in fedimint-core

maan2003:
see other dependent PRs for usage 

used by #4784 and #4768

PR #4784, opened 2024-03-31
chore: simplify JitCore impl

maan2003:
using the Waiter from previous commit

on top of https://github.com/fedimint/fedimint/pull/4783


fedimint-core/src/task/jit.rs
```diff
@@ -69,15 +69,14 @@ where
...
-            val: self.val.clone(),
+            shared: self.shared.clone(),
         }
     }
 }
 impl<T, E> JitCore<T, E>
 where
-    T: MaybeSend + 'static,
-    E: MaybeSend + 'static + fmt::Display,
+    T: MaybeSend + MaybeSync + 'static,
```
dpc, line 78: New bond not great as it potentially constrain the use.

fedimint-core/src/task/jit.rs
```diff
@@ -88,24 +87,26 @@ where
...
+        let shared = Arc::new(JitCoreShared {
+            val: std::sync::OnceLock::new(),
+            initialized_waiter: Waiter::new(),
+        });
+        let shared2 = shared.clone();
         super::imp::spawn(
             &format!("JitTry {} value", std::any::type_name::<T>()),
             async move {
                 select! {
+                    biased;
```
dpc, line 100: I doubt it matters, but if anything - I would rather get the value than cancelation if both finished at the same time.

fedimint-core/src/task/jit.rs
```diff
@@ -88,24 +87,26 @@ where
         Fut: Future<Output = std::result::Result<T, E>> + 'static + MaybeSend,
     {
         let (cancel_tx, mut cancel_rx) = tokio::sync::mpsc::channel(1);
-        let (val_tx, val_rx) = oneshot::channel();
         let type_name = std::any::type_name::<T>();
+        let shared = Arc::new(JitCoreShared {
+            val: std::sync::OnceLock::new(),
+            initialized_waiter: Waiter::new(),
```
dpc, line 93: `val_ready` maybe?

dpc (commented):
It is simpler, but that extra `+ MaybeSync` is a downgrade. I don't even understand why it is needed. Maybe it isn't?

fedimint-core/src/task/jit.rs
maan2003: there is no point in getting the value when there is no one left to use it :)

maan2003:
it is needed because OnceLock is accessed from two threads now, I think `Sync` is fine, most values should be Sync especially when we are in async land.

fedimint-core/src/task/jit.rs
maan2003: also non biased uses `rand` to decide which one to poll

dpc:
> it is needed because OnceLock is accessed from two threads now, I think `Sync` is fine, most values should be Sync especially when we are in async land.


I don't get, but I guess I'd have to remove it and try. Before it worked, so it seems it should still work. These kind of limitations tend to bite later. I try to do something else that is slightly non-trivial, and I have to invent extra workarounds.

fedimint-core/src/task/jit.rs
```diff
@@ -53,13 +53,13 @@ where
...
-    #[allow(clippy::type_complexity)]
-    val_rx:
-        Arc<std::sync::Mutex<Option<tokio::sync::oneshot::Receiver<std::result::Result<T, E>>>>>,
-    val: Arc<sync::OnceCell<T>>,
+    shared: Arc<JitCoreShared<T, E>>,
+}
+
+#[derive(Debug)]
+struct JitCoreShared<T, E> {
+    val: std::sync::OnceLock<Result<T, StdMutex<Option<E>>>>,
```
elsirion, line 61: @maan2003 showing me all the new goodies in `std` :grin:

fedimint-core/src/task/jit.rs
```diff
@@ -114,70 +115,30 @@ where
...
-    pub async fn get_try(&self) -> std::result::Result<&T, OneTimeError<E>> {
-        /// Temporarily taken data out of `arc`, that will be put back on drop,
-        /// unless canceled
-        ///
-        /// Used to achieve cancelation safety.
-        struct PutBack<'a, T> {
-            val: Option<T>,
-            arc: &'a Arc<std::sync::Mutex<Option<T>>>,
-        }
-
```
elsirion, line 133: Avoiding the cancel safety complexity here seems like a good trade-off even though we get more restrictions on `T`.

elsirion (approved):

elsirion (dismissed):

elsirion:
@dpc the `PutBack` mechanic is a bit convoluted, at the same time: why touch working code when it leads to more restrictions? If I had my way it would be just a `watch` channel and require anything in there to be `Clone` (for errors/results there is a solution now, everything else can be wrapped in an `Arc` if need be). But changing it doesn't feel like a great use of time.

maan2003:
lets close this for now then, we can try this in case we find a bug in jit core in future

PR #4791, opened 2024-04-01
chore(consensus): log dropped consensus proposals

dpc:


elsirion:
Why close? Looks useful? @joschisan

dpc:
I think it only can error during disconnects, which would be less useful?

PR #4792, opened 2024-04-01
chore(consensus): increase polling rate for consensus proposals

dpc:


dpc:
@joschisan Does this value make any difference?

elsirion:
There are two philosophies to this:
* me: Ideally we'd make this a async fn or channel so that modules can send CIs whenever deemed necessary, with the lowest latency.
* @joschisan: CIs shouldn't be latency critical. Once the old LN module is phased out the polling could be done every 10s or so even.

What's the use case @dpc? My position is based on the idea that there *might* be a module type where it's the best solution (e.g. trading engine) to do something latency critical via consensus that doesn't need to be a transaction.

But @joschisan did great work refactoring all existing modules such that this isn't the case anymore, mostly by removing signature aggregation from the server. So maybe there's a simple fix here too.

joschisan:
The only latency critical items we had / still have are signature shares, which is a strongly discouraged pattern since it’s very wasteful with respect to disk and compute consumption and it’s not even necessary to order the shares, so we have a unnecessary ordering latency added.

Therefore, I suspected that latency critical CIs may be an indicator of a misled module design and wanted to support them when we have a module that actually does require them.

dpc:
@elsirion @joschisan 

If I have a test that sets a meta value on peers and then waits to see if it was set, IIUC, it might be x10 speedup for such test. A win is a win.

And a delay here (1s) dwarfs the AlephBfts internal tick time, IIUC, which is like 60ms. Decreasing the exact number here is not a complexity problem, and I doubt the load will increase anyway.

IIUC the delay here is adding latency for anything submitted through the API to get into consensus. And it's not a problem for other modules, because we basically special-cased transactions.

BTW. A relatively simple solution would be to turn that sleep into a select over `(sleep, some_notification_channel)`, and let modules send over that channel when they did something that could have benefit from instant consensus proposal check. Which is effectively the same thing we're doing for Transactions, I think. The API can stay the same, `consensus_proposal` can be dumb, and we can get optimal performance.

dpc:
I guess after  #4811 , the default time here could be even increased if we desired, so closing.

PR #4794, opened 2024-04-01
fix(fedimintd): fix doc comment for finality_delay option

emilioziniades:
Noticed that the docstring was a copy-paste of the option above it. Fixed it.


dpc (approved):

m1sterc001guy (approved):

fedimintd/src/fedimintd.rs
```diff
@@ -83,7 +83,8 @@ pub struct ServerOpts {
     /// The bitcoin network that fedimint will be running on
     #[arg(long, env = FM_BITCOIN_NETWORK_ENV, default_value = "regtest")]
     network: bitcoin::network::constants::Network,
-    /// The bitcoin network that fedimint will be running on
+    /// The number of blocks mined after which an on-chain transaction is
+    /// considered finalized
     #[arg(long, env = FM_FINALITY_DELAY_ENV, default_value = "10")]
     finality_delay: u32,
```
elsirion, line 89: Certainly better than the previous (wrong) comment, but with a finality delay of 10 you actually need 11 confirmations. Finality delay is by how much the federation stays behind the blockhain tip, but a tx also needs to be included in a block first.

dpc: BTW. Why didn't we just go with `confirmations_required` and `-1` (`.saturating_sub(1)` to be precise) in this 3 places of code that use it?

elsirion: I was never thinking in terms of confirmations required till we began actually using Fedimint outside CI :see_no_evil: now we are stuck with it in the config. We can write a utility fn that returns the required confirmations, but re-writing the config doesn't seem worth it for now (I still have PTSD from the `block_count`/`block_height` change 😩 *of course* Bitcoin Core doesn't count the genesis block 😖).

emilioziniades: > Certainly better than the previous (wrong) comment, but with a finality delay of 10 you actually need 11 confirmations. Finality delay is by how much the federation stays behind the blockhain tip, but a tx also needs to be included in a block first.
> 
> (I confused myself often enough with this)

Oh I see! I was thinking about "finality" for the layer-one blockchain, like when you withdraw BTC from coinbase or wherever and it requires 6 confirmation blocks above the block that has the withdrawal tx for the withdrawal to be "confirmed". I guess it's an overloaded term. But this makes sense in terms of fedimint.

I'll amend the commit

elsirion (approved):

dpc (approved):

PR #4798, opened 2024-04-02
fix: await_block_sync() polls the underlying command

tvolk131:
Without this fix, I will get the following error about 80% of the time when running `just mprocs` on my M1 Pro Macbook:

```
Tommys-MacBook-Pro-14:fedimint tommyvolk$ just mprocs
./scripts/dev/mprocs/run.sh 
warning: /Users/tommyvolk/Documents/GitHub/fedimint/Cargo.toml: unused manifest key: workspace.package.name
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
2024-04-01T12:54:13.681421Z  INFO devimint: Setting up test dir path=/tmp/nix-shell.Xf3wjs/devimint-96819-163
2024-04-01T12:54:13.682836Z  INFO devimint::devfed: Starting dev federation
2024-04-01T12:54:13.683420Z  INFO devimint: Peging in client and gateways
2024-04-01T12:54:21.554939Z  INFO devimint: Running DKG
2024-04-01T12:54:25.144207Z  INFO devimint: Starting consensus
2024-04-01T12:54:29.269692Z ERROR fedimint_cli: Encoded client secret not present in DB
2024-04-01T12:54:36.991989Z  INFO devimint: Pegins completed elapsed_ms=23308
2024-04-01T12:54:37.059300Z  INFO devimint: Setup gateway environment variables
2024-04-01T12:54:37.315101Z  WARN devimint: Main process failed, will shutdown err=command: fedimint-cli --data-dir=/tmp/nix-shell.Xf3wjs/devimint-96819-163/clients/default-0 dev wait-block-count 122
Error: command: fedimint-cli --data-dir=/tmp/nix-shell.Xf3wjs/devimint-96819-163/clients/default-0 dev wait-block-count 122
error: Recipe `mprocs` failed on line 104 with exit code 1
```

Where this line in particular will cause the dev federation setup to crash:

`
2024-04-01T12:54:29.269692Z ERROR fedimint_cli: Encoded client secret not present in DB`

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -371,9 +371,14 @@ impl Federation {
         let finality_delay = self.get_finality_delay().await?;
         let block_count = self.bitcoind.get_block_count().await?;
         let expected = block_count.saturating_sub(finality_delay.into());
-        cmd!(self.client, "dev", "wait-block-count", expected)
-            .run()
-            .await?;
+        poll("await-block-sync", || async {
+            cmd!(self.client, "dev", "wait-block-count", expected)
+                .run()
+                .await
+                .map_err(ControlFlow::Continue)?;
+            Ok(())
+        })
+        .await?;
         Ok(expected)
     }
 


devimint/src/federation.rs
```diff
@@ -371,9 +371,14 @@ impl Federation {
         let finality_delay = self.get_finality_delay().await?;
         let block_count = self.bitcoind.get_block_count().await?;
         let expected = block_count.saturating_sub(finality_delay.into());
-        cmd!(self.client, "dev", "wait-block-count", expected)
-            .run()
-            .await?;
+        poll("await-block-sync", || async {
```
dpc, line 374: This looks OK, but would great to know what exactly is failing there. IIRC `stderr` of commands ran by devimint is printed out.

Possibly we should make the `fedimint-cli dev wait-block-count` retry etc. as it is used in other places, etc. We might be assuming it can just keep waiting.

maan2003: https://github.com/fedimint/fedimint/pull/4800

tvolk131: I tried that branch out locally, but my computer still has the same issue :(

tvolk131: This appears to be where the error is being thrown:

https://github.com/fedimint/fedimint/blob/747392ecd73bcf5304a19c2cb034911345ebece8/fedimint-client/src/lib.rs#L804-L825

`client_secret` is `None`, so `load_decodable_client_secret_opt()` returns `Ok(None)` which causes `load_decodable_client_secret()` to bail.

justinmoon:
On current master I ran `just mprocs` twice, got the error @tvolk131 mentioned in PR description the first time and then it worked the second time.

devimint/src/federation.rs
dpc: @maan2003 That 60s timeout is too large to make a difference in practice.

Oh, I see it now

```
            Command::Dev(DevCmd::WaitBlockCount { count: target }) => {
                task::timeout(Duration::from_secs(60), async move {
                    let client = self.client_open(&cli).await?;
                    loop {
                        let wallet = client.get_first_module::<WalletClientModule>();
```

That `client_open` should be *inside* the loop. And also - all errors should lead to retry, not finishing early. We basically have no retry on error there, only retry on success.

dpc: Let's discuss in #4800 and depending what are going to do, we should land or close this one.

@tvolk131 This is a great find one way or the other, and addressing it will improve robustness of all our scripts, so :+1: even if we don't land this PR.

dpc:
Let's discuss in #4800 and depending what are going to do, we should land or close this one.

@tvolk131 This is a great find one way or the other, and addressing it will improve robustness of all our scripts, so :+1: even if we don't land this PR.

dpc:
Alternatively: https://github.com/fedimint/fedimint/pull/4801

elsirion:
Is this still relevant after #4801?

tvolk131:
> Is this still relevant after #4801?

Nope! Closing.

PR #4800, opened 2024-04-02
fix: retry in wait-block-count

maan2003:
why was timeout added here?

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -14,6 +14,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use std::{fs, result};
 
+use anyhow::format_err;
 use bip39::Mnemonic;
 use clap::{Args, CommandFactory, Parser, Subcommand};
 use db_locked::LockedBuilder;
@@ -33,8 +34,8 @@ use fedimint_core::config::{
 use fedimint_core::core::OperationId;
 use fedimint_core::db::{Database, DatabaseValue};
 use fedimint_core::module::{ApiAuth, ApiRequestErased};
-use fedimint_core::util::{handle_version_hash_command, SafeUrl};
-use fedimint_core::{fedimint_build_code_version_env, task, PeerId, TieredMulti};
+use fedimint_core::util::{handle_version_hash_command, retry, ConstantBackoff, SafeUrl};
+use fedimint_core::{fedimint_build_code_version_env, PeerId, TieredMulti};
 use fedimint_ln_client::LightningClientInit;
 use fedimint_logging::{TracingSetup, LOG_CLIENT};
 use fedimint_meta_client::MetaClientInit;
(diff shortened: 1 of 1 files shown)


fedimint-cli/src/lib.rs
```diff
@@ -793,23 +793,19 @@ impl FedimintCli {
...
+                        .with_module(wallet.id)
+                        .fetch_consensus_block_count()
+                        .await?;
+                    if count >= target {
+                        break Ok(CliOutput::WaitBlockCount { reached: count });
                     }
-                })
-                .await
-                .map_err_cli_msg("reached timeout")?
+                    task::sleep(Duration::from_millis(100)).await;
```
dpc, line 807: We should log something on each failure, just to let user know what's going on and that it didn't just hang.

dpc:
See: https://github.com/fedimint/fedimint/pull/4798#discussion_r1548428237

We currently have no retries for errors (`?` returning early). Should we? Normally we don't do it, but given that this command is often used for scripting, maybe we should here?

fedimint-cli/src/lib.rs
elsirion: That will actually lead to a log every 60s with current server-side timeouts.

elsirion (dismissed):
Idk why there was a timeout in the first place, feel free to add some logging and re-request review :)

dpc (dismissed):

fedimint-cli/src/lib.rs
dpc, line 807: BTW. I wonder if we should close the client when sleeping to unlock the client for other things to run, potentially.

maan2003: updated

dpc (approved):
LGTM, but I think the linter doesn't like something. **Edit**: doesn't compile. `FnMut` + `async` fun times.

maan2003:
rust needs async closures, closure + async block sucks

maan2003:
I also gave up after 10 minutes, should have trusted you.
made `open` accept `&self`

dpc:
> I also gave up after 10 minutes, should have trusted you.
> made `open` accept `&self`

I couldn't even make that one work ... and now I don't understand why your version works. But I accept you're just better at Rust. :D

dpc (dismissed):

fedimint-cli/src/lib.rs
```diff
@@ -792,25 +793,29 @@ impl FedimintCli {
...
+                    let wallet = client.get_first_module::<WalletClientModule>();
+                    let count = client
+                        .api()
+                        .with_module(wallet.id)
+                        .fetch_consensus_block_count()
+                        .await?;
+                    if count >= target {
+                        Ok(CliOutput::WaitBlockCount { reached: count })
+                    } else {
+                        Err(format_err!("target not reached"))
```
dpc, line 812: Thought: should we `info!(...)` something here (e.g. target, current, "Block count not reached").

maan2003: done

elsirion (approved):

PR #4772, opened 2024-03-29
 chore: sanity test meta module via standalone devimint test

dpc:
I thought it would be good to have basic test for meta module, and I discovered a problem.

To reproduce:

```
env RUST_LOG=server::module::meta=trace,info ./scripts/tests/meta-module-test.sh
```

Sometimes (quite often on my system) it will fail timing out polling for reaching a consensus for a meta value.

You can look up the `/tmp/devimint...` dir and look at the peer logs (e.g. `less /tmp/devimint-3273235-123/logs/fedimintd-3.log`)

This is from a fedimintd-3, that seemed to fail to submit:

```
2024-03-29T18:12:41.445613Z DEBUG server::module::meta: Received a submitted a value peer_id=1 key=0 value=7b22666f6f223a22626172227d salt=1
3548918736504005055
2024-03-29T18:12:41.445673Z  INFO server::module::meta: Peer submitted a value peer_id=1 key=0 value_len=13 matching=1 threshold=3
2024-03-29T18:12:41.771767Z  INFO server::module::meta: Our own guardian submitted a value key=0 value_len=13
2024-03-29T18:12:42.096532Z DEBUG server::module::meta: Received a submitted a value peer_id=2 key=0 value=7b22666f6f223a22626172227d salt=1
196510382307997764
2024-03-29T18:12:42.096603Z  INFO server::module::meta: Peer submitted a value peer_id=2 key=0 value_len=13 matching=2 threshold=3
2024-03-29T18:12:42.748009Z DEBUG server::module::meta: Received a submitted a value peer_id=0 key=0 value=7b22626172223a2262617a227d salt=1
5463300193985426068
2024-03-29T18:12:42.748090Z  INFO server::module::meta: Peer submitted a value peer_id=0 key=0 value_len=13 matching=1 threshold=3
```

The `Our own guardian submitted a value` is followed by receiving that submission in `process_consensus_item` with `peer_id=2` instead of `3`!

I was under impression that `peer_id` will always mean originator of the message, but it doesn't look like that's the case. If indeed it is not guaranteed, then we should not be passing it to `process_consensus_item` as it doesn't really mean anything reliable on this level of abstraction, and it will just lead to bugs like this one. And I'll need to add some signing etc.

Or maybe it should be reliable, and it's a bug somewhere.

Please sanity check my findings and let me know.

PR #4803, opened 2024-04-02
fix(meta): potential (and mostly harmless) race condition

dpc:
on clearing the outstanding minority submissions.

See https://github.com/fedimint/fedimint/pull/4772#issuecomment-2033170196

### DIFF
--- a/modules/fedimint-meta-server/src/lib.rs
+++ b/modules/fedimint-meta-server/src/lib.rs
@@ -218,24 +218,45 @@ impl Meta {
         let mut to_delete: Vec<MetaKey> = vec![];
         let mut to_submit = vec![];
 
-        for (key, MetaDesiredValue { value, salt }) in desired {
-            if Self::get_submission(dbtx, key, self.our_peer_id)
-                .await
-                .as_ref()
+        for (
+            key,
+            MetaDesiredValue {
+                value: desired_value,
+                salt,
+            },
+        ) in desired
+        {
+            let consensus_value = &Self::get_consensus(dbtx, key).await;
+            let consensus_submission_value =
+                Self::get_submission(dbtx, key, self.our_peer_id).await;
+            if consensus_submission_value.as_ref()
                 == Some(&MetaSubmissionValue {
-                    value: value.clone(),
+                    value: desired_value.clone(),
                     salt,
                 })
             {
                 to_delete.push(key);
-            } else if Self::get_consensus(dbtx, key).await.as_ref() == Some(&value) {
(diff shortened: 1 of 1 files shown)


modules/fedimint-meta-server/src/lib.rs
```diff
@@ -218,24 +218,45 @@ impl Meta {
...
-                // When a value is new and equal to current consensus, it should both: be
-                // submitted so it can clear/cancel previous submissions, and
-                // deleted (afterwards).
-                to_delete.push(key);
-                to_submit.push(MetaConsensusItem { key, value, salt });
+            } else if consensus_value.as_ref() == Some(&desired_value) {
+                if consensus_submission_value.is_none() {
+                    // our desired value cleared the submission (as it is equal the consensus)
+                    // so we can stop proposing it
+                    to_delete.push(key);
```
elsirion, line 243: There is a delay between us proposing something and it being accepted, so just because `consensus_submission_value` is `None` now doesn't mean it won't become `Some(_)` at a later point without us sending another proposal. I'm not sure if this is mitigated by the guarantee that earlier proposals will always precede later proposals in consensus output. 

Isn't there any versioning in the votes, so that old votes don't matter anymore anyway? Sounds like a much easier solution. What is that `salt` about? Maybe it can be (ab)used for that?

dpc: > There is a delay between us proposing something and it being accepted, so just because consensus_submission_value is None now doesn't mean it won't become Some(_) at a later point without us sending another proposal.

Hmmmmm...

For this to happen, the current consensus would had to be some `C`, our current submission `None` (meaning `= C`) and then desired value requested to be set to `X != C` and submitted. While it's propagating, desired value set to `C` again,  and then the user would need to set some other key to something else, because only then we actually act on that `to_delete` and delete it.


Technically still a race that could lead to a request fail to clear that `X`. In practice it's even less probable than it was before, and just as before - probably doesn't matter.

> What is that salt about? Maybe it can be (ab)used for that?

The problem is that the submission equal to consensus is just deleted, which includes the salt, so there's nothing to compare with.


Ehh... I think, since I can really make it technically correct, and don't want to deal with changing the parts that are actually consensus-sentive, I'm just going to stop deleting own desired key at all. Just not worth it, tryint to be too clever by half about it.

dpc: @elsirion will try https://github.com/fedimint/fedimint/pull/4836 instead

PR #4623, opened 2024-03-19
chore: bump version to 0.4.0-alpha

elsirion:
Also adds missing version to metrics dependency in fedimint-server.

PR #4804, opened 2024-04-02
fix(devimint): use Version instead of VersionReq for comparisons

bradleystachurski:
`VersionReq` follows the semver range spec, which has unexpected behavior.

```rust
#[test]
fn version_req_behavior() -> Result<()> {
    use semver::{Version, VersionReq};
    let fedimint_cli_version = Version::parse("0.4.0-alpha")?;
    let req_version = VersionReq::parse(">=0.3.0-alpha")?;

    // 0.4.0-alpha is NOT >= 0.3.0-alpha since it contains a prerelease
    assert!(!req_version.matches(&fedimint_cli_version));

    // comparing using `Version`s gives us the desired behavior
    assert!(Version::parse("0.4.0-alpha")? >= Version::parse("0.3.0-alpha")?);
    Ok(())
}
```

This breaks multiple tests now that we're trying to bump to 0.4.0-alpha (noticed in https://github.com/fedimint/fedimint/pull/4623). We can instead compare using `Version`s.

For context why semver chose this, refer to section 5 from https://github.com/semver/semver/blob/efcff2c838c9945f79bfd21c1df0073271bcc29c/ranges.md. There's also additional context in this `semver` crate [issue](https://github.com/dtolnay/semver/issues/172).

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -24,7 +24,7 @@ use fedimintd::envs::FM_EXTRA_DKG_META_ENV;
 use fs_lock::FileLock;
 use futures::future::join_all;
 use rand::Rng;
-use semver::VersionReq;
+use semver::Version;
 use tokio::time::Instant;
 use tracing::{debug, info};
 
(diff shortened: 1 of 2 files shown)


devimint/src/federation.rs
```diff
@@ -399,7 +399,7 @@ impl Federation {
         let start_time = Instant::now();
         debug!(target: LOG_DEVIMINT, "Awaiting LN gateways registration");
         let fedimint_cli_version = crate::util::FedimintCli::version_or_default().await;
-        let command = if VersionReq::parse("<0.3.0-alpha")?.matches(&fedimint_cli_version) {
+        let command = if fedimint_cli_version < Version::parse("0.3.0-alpha")? {
```
m1sterc001guy, line 402: Just a thought: do we want to put all of our version constants in a central place (maybe a new file?). Might help with reasoning about which versions have changes between them.

m1sterc001guy (approved):

maan2003 (approved):

maan2003:
lets just merge child https://github.com/fedimint/fedimint/pull/4805

devimint/src/federation.rs
bradleystachurski: https://github.com/fedimint/fedimint/pull/4828

Issue #4132, opened 2024-01-25
Remove `async-trait` crate


tvolk131:
[Async traits are stable](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html)! The two downsides of using them are mentioned [here](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#should-i-still-use-the-async_trait-macro). Assuming they aren't problems, let's remove our usage of the `async-trait` crate wherever we can.

elsirion:
We want dynamic dispatch for nearly every async trait afaik.

elsirion:
> We want dynamic dispatch for nearly every async trait afaik.

Not actionable due to limitations.

PR #4806, opened 2024-04-03
chore: remove async trait

maan2003:
lets try with simple trait first

the impl looks simple, the definition is slightly complicated.

### DIFF
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -430,7 +430,6 @@ pub trait IDynCommonModuleInit: Debug {
 }
 
 /// Trait implemented by every `*ModuleInit` (server or client side)
-#[apply(async_trait_maybe_send!)]
 pub trait ModuleInit: Debug + Clone + Send + Sync + 'static {
     type Common: CommonModuleInit;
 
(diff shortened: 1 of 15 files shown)


maan2003:
this reduces one allocation per call to async function (not requiring Box::pin)

elsirion (approved):
Saving one allocation in a function only called by a debug tool … :rofl: 

I see this more as a test if we could do the same for other traits. Probably still not really worth it unless benchmarks show we lose a lot due to boxing.

fedimint-core/src/module/mod.rs
```diff
@@ -442,11 +441,17 @@ pub trait ModuleInit: Debug + Clone + Send + Sync + 'static {
...
     const DATABASE_VERSION: DatabaseVersion;
 
-    async fn dump_database(
+    fn dump_database(
         &self,
         dbtx: &mut DatabaseTransaction<'_>,
         prefix_names: Vec<String>,
-    ) -> Box<dyn Iterator<Item = (String, Box<dyn erased_serde::Serialize + Send>)> + '_>;
+    ) -> maybe_add_send!(
+        impl Future<
```
dpc, line 449: Yyyy... maybe not this one? :D

maan2003: more macros!

`maybe_send_future![T]`

dpc: I heard that async in trait was still limited and something about it was unstable and recommeded to use only in private traits, etc. Is this no longer true?

I think we don't need to eliminate `#[async_trait]` altogether, we can be lazy and opportunistic about it.

tvolk131:
Related: #4132

dpc (approved):

fedimint-core/src/module/mod.rs
dpc: Hmmm... I think I was thinking about https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#is-it-okay-to-use---impl-trait-in-traits

> For private traits you can use -> impl Trait freely. For public traits, it's best to avoid them

PR #4809, opened 2024-04-03
chore: TaskGroup::make_subgroup non async

maan2003:


m1sterc001guy (approved):

dpc (approved):

tvolk131:
There's a clippy rule that checks for this (`cargo clippy -- -D clippy::unused_async`) and for me it found a few other places where we could remove `async`. Do we want to enforce this in CI?

dpc:
> `clippy::unused_async`) and for me it found a few other places where we could remove `async`. Do we want to enforce this in CI?

It's not enabled by default? If so, give it a try, :+1: .

There might few places where it might make sense to have "useless" `async`, eg. to support future implementation that could require async without breaking the callers. But we can `#[allow(...)]` such places with a comment.

tvolk131:
> give it a try

Created #4907 to track this, as I don't have time to work on this right now

PR #4811, opened 2024-04-03
chore: modules to trigger `consensus_proposal` manualy, use in meta

dpc:
Re https://github.com/fedimint/fedimint/pull/4792#issuecomment-2035027975

Some modules will inevitably be bound in some places by the latency of `consensus_proposal` polling, and this is a clean and easy to use method to let them trigger consensus item processing.

This mimics how submitted transaction are being handled already - a side-channel. And saves module authors dealing with complicated calling conventions and wiring similar channels manually. The `consensus_proposal` can stay relatively dumb.

### DIFF
--- a/fedimint-core/Cargo.toml
+++ b/fedimint-core/Cargo.toml
@@ -33,7 +33,7 @@ serde_json = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 hex = { version = "0.4.3", features = ["serde"] }
-tokio = { version = "1.36.0", features = ["sync", "io-util"] }
+tokio = { version = "1.37.0", features = ["sync", "io-util"] }
 thiserror = { workspace = true }
 tracing = { workspace = true }
 threshold_crypto = { workspace = true }
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -581,7 +581,7 @@ pub trait CommonModuleInit: Debug + Sized {
 }
 
 /// Can be used to send a notification to consensus server,
-/// to trigger immediate [`Server::consensus_proposal`] call.
+/// to trigger immediate [`ServerModule::consensus_proposal`] call.
 ///
 /// Useful for modules to optionally decrease latency of consensus
 /// item handling.
@@ -840,6 +840,27 @@ pub trait ServerModule: Debug + Sized {
     }
 
     /// This module's contribution to the next consensus proposal
+    ///
+    /// This method is polled by the server periodically to gather
+    /// consensus items modules would like included in the consensus.
+    ///
+    /// Note that returning an item does not guarantee its processing,
+    /// and modules should keep proposing the difference between
+    /// the desired state and current consensus view.
+    ///
+    /// Warning: The consensus algorithm will deduplicate and ignore
+    /// previously submitted items. This can lead to weird behavior
+    /// when identically deserialized items can be generated multiple times.
+    /// Use some random-like data to make such requests unique to avoid
+    /// this problem.
+    ///
+    ///
+    /// Modules should be architected to avoid making the consensus
+    /// processing latency a bottleneck, so the rate at which this call
+    /// is made is not guaranteed and relatively infrequent. If a module
+    /// desires immediate processing, it can use [`ConsensusProposalTrigger`]
+    /// that it can obtain via
+    /// [`ServerModuleInitArgs::consensus_proposal_trigger`].
     async fn consensus_proposal<'a>(
         &'a self,
         dbtx: &mut DatabaseTransaction<'_>,
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -24,6 +24,7 @@ use fedimint_logging::LOG_NET_API;
 use futures::Future;
 use jsonrpsee_core::JsonValue;
 use serde::{Deserialize, Serialize};
+use tokio::sync::watch;
 use tracing::instrument;
 
 // TODO: Make this module public and remove the wildcard `pub use` below
@@ -507,6 +508,7 @@ pub trait IServerModuleInit: IDynCommonModuleInit {
         db: Database,
         task_group: &TaskGroup,
         our_peer_id: PeerId,
+        consensus_proposal_trigger: ConsensusProposalTrigger,
     ) -> anyhow::Result<DynServerModule>;
 
     fn validate_params(&self, params: &ConfigGenModuleParams) -> anyhow::Result<()>;
@@ -578,6 +580,26 @@ pub trait CommonModuleInit: Debug + Sized {
     fn decoder() -> Decoder;
 }
 
+/// Can be used to send a notification to consensus server,
+/// to trigger immediate [`Server::consensus_proposal`] call.
+///
+/// Useful for modules to optionally decrease latency of consensus
+/// item handling.
+#[derive(Debug, Clone)]
+pub struct ConsensusProposalTrigger(watch::Sender<u64>);
+
+impl From<watch::Sender<u64>> for ConsensusProposalTrigger {
+    fn from(value: watch::Sender<u64>) -> Self {
+        Self(value)
+    }
+}
+
+impl ConsensusProposalTrigger {
+    pub fn send(&self) {
+        self.0.send_modify(|v| *v = v.wrapping_add(1))
+    }
+}
+
 pub struct ServerModuleInitArgs<S>
 where
     S: ServerModuleInit,
@@ -587,6 +609,7 @@ where
     task_group: TaskGroup,
     our_peer_id: PeerId,
     num_peers: NumPeers,
+    consensus_proposal_trigger: ConsensusProposalTrigger,
     // ClientModuleInitArgs needs a bound because sometimes we need
     // to pass associated-types data, so let's just put it here right away
     _marker: marker::PhantomData<S>,
@@ -614,6 +637,12 @@ where
 
     pub fn our_peer_id(&self) -> PeerId {
         self.our_peer_id
+    }
+
+    /// Get [`ConsensusProposalTrigger`] useful for triggering immediate
+    /// [`ServerModule::consensus_proposal`]
+    pub fn consensus_proposal_trigger(&self) -> &ConsensusProposalTrigger {
+        &self.consensus_proposal_trigger
     }
 }
 /// Module Generation trait with associated types
@@ -701,6 +730,7 @@ where
         db: Database,
         task_group: &TaskGroup,
         our_peer_id: PeerId,
+        consensus_proposal_trigger: ConsensusProposalTrigger,
     ) -> anyhow::Result<DynServerModule> {
         <Self as ServerModuleInit>::init(
             self,
@@ -711,6 +741,7 @@ where
                 task_group: task_group.clone(),
                 our_peer_id,
                 _marker: Default::default(),
+                consensus_proposal_trigger,
             },
         )
         .await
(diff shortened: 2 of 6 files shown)


joschisan:
Do I understand correctly that you are trying to achieve consensus on the meta values faster? Why is that so important?

dpc:
> Do I understand correctly that you are trying to achieve consensus on the meta values faster? Why is that so important?
> 
> This pr subverts the direction of for proposing consensus items I have been working on for a year now.

@joschisan 

It's not "that important", but it also isn't some architectural gymnastics or unbearable complexity cost to make it run in an optimal way.

I don't see how making that one sleep a bit smarter "subverts the direction of for proposing consensus" unless the "direction" involves intentionally making everything except some whitelisted path artificially slower to artificially punish certain use-cases.

When a module needs to submit a CI, ideally we want it processed ASAP, even if just for the tests to run as fast as possible. Yes, consensus latency should be avoided as a bottleneck in protocols designed for Fedimint. It doesn't mean that we should not optimize consensus latency all the same, just like we did before.

To be exaggerate: Just because `fedimintd` starts once and then runs for hours doesn't mean we should add `sleep(Duration::from_secs(30))` at the start "because it's not important", and in no way does not having it "subvert the goal of using fedimintd as a long running process".

Now, if there's something this change makes worse that I just don't see, or if there's a better approach, or a complexity is deemed too high (which I don't think is the case, but it's a judgment call), then fine and I'm open for better suggestions or will even admit nothing case be done, and I'll have to live with it.

douglaz:
This looks fine to me. There is a lot of room for optimization on fedimint and if it's a relatively simple change then we must do it.

Just double check the docs, something is failing there.

fedimint-core/Cargo.toml
```diff
@@ -33,7 +33,7 @@ serde_json = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 hex = { version = "0.4.3", features = [ "serde"] }
-tokio = { version = "1.36.0", features = ["sync", "io-util"] }
+tokio = { version = "1.37.0", features = ["sync", "io-util"] }
```
elsirion, line 36: Why the version bump *here* and not just `Cargo.lock`? Any particular feature that was introduced?

fedimint-core/src/module/mod.rs
```diff
@@ -578,6 +580,26 @@ pub trait CommonModuleInit: Debug + Sized {
     fn decoder() -> Decoder;
 }
 
+/// Can be used to send a notification to consensus server,
+/// to trigger immediate [`ServerModule::consensus_proposal`] call.
+///
+/// Useful for modules to optionally decrease latency of consensus
+/// item handling.
+#[derive(Debug, Clone)]
+pub struct ConsensusProposalTrigger(watch::Sender<u64>);
```
elsirion, line 589: @maan2003 didn't you build something similar recently? I think it was called `Wait`. Could we maybe unify the underlying mechanism? I like this approach.

fedimint-core/src/module/mod.rs
```diff
@@ -590,6 +612,7 @@ where
     // ClientModuleInitArgs needs a bound because sometimes we need
     // to pass associated-types data, so let's just put it here right away
     _marker: marker::PhantomData<S>,
+    consensus_proposal_trigger: ConsensusProposalTrigger,
```
elsirion, line 615: nit: move in front of `_marker`

fedimint-core/src/module/mod.rs
```diff
@@ -809,6 +840,27 @@ pub trait ServerModule: Debug + Sized {
...
+    ///
+    /// Note that returning an item does not guarantee its processing,
+    /// and modules should keep proposing the difference between
+    /// the desired state and current consensus view.
+    ///
+    /// Warning: The consensus algorithm will deduplicate and ignore
+    /// previously submitted items. This can lead to weird behavior
+    /// when identically deserialized items can be generated multiple times.
+    /// Use some random-like data to make such requests unique to avoid
+    /// this problem.
```
elsirion, line 855: Is this still the case? I think you mentioned on one of our calls that this was fixed @joschisan?

elsirion (approved):
Very elegant solution

elsirion:
Needs a docs fix though for CI https://github.com/fedimint/fedimint/actions/runs/8550335981/job/23427144737?pr=4811#step:7:136

fedimint-core/src/module/mod.rs
dpc: The mechanism is unified already. It's `tokio::sync::watch`. :D

Here I'm just wrapping it in a newtype, so the API is more 🌈*type-driven*🌈.

fedimint-core/src/module/mod.rs
dpc: Oh, really? That would be important to know. Please link to a PR if that's the case. I'm curious how it is done.

fedimint-core/src/module/mod.rs
elsirion: I misremembered, the other abstraction ([`Waiter`](https://docs.fedimint.org/fedimint_core/task/waiter/struct.Waiter.html)) had slightly different semantics so probably not worth switching to this approach.

elsirion:
CI is quite red

fedimint-core/src/module/mod.rs
```diff
@@ -578,6 +580,26 @@ pub trait CommonModuleInit: Debug + Sized {
...
+
+impl From<watch::Sender<u64>> for ConsensusProposalTrigger {
+    fn from(value: watch::Sender<u64>) -> Self {
+        Self(value)
+    }
+}
+
+impl ConsensusProposalTrigger {
+    pub fn send(&self) {
+        self.0.send_modify(|v| *v = v.wrapping_add(1))
```
elsirion, line 599: *Technically*, there is a race condition that, after 2^64 sends, a recipient that did not get to execute in the meantime will not notice the notification :zany_face:

elsirion (approved):

fedimint-core/src/module/mod.rs
maan2003, line 589: can this be `watch::Sender<()>`, Sender and Receiver don't require T: PartialEq so it can't check if value in `Receiver::changed()`.

fedimint-core/src/module/mod.rs
dpc: The polling is best effort and even then consensus doesn't guarantee anything, so *technically* still ok. 🤣

fedimint-core/src/module/mod.rs
dpc: Maybe but I'm spooked about the "changed" aspect of the unit so whatever. 🤣

douglaz (approved):
LGTM

fedimint-server/src/consensus/server.rs
```diff
@@ -838,7 +845,13 @@ async fn submit_module_consensus_items(
                     }
                 }
 
-                sleep(Duration::from_secs(1)).await;
+                drop(dbtx);
+
+                tokio::select! {
+                    biased;
+                    _ = sleep(Duration::from_secs(1)) => {},
```
elsirion, line 852: nit: Should likely use a tokio [`Interval`](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html) so this doesn't reset.

elsirion (changes requested):

elsirion:
This doesn't seem too urgent, @joschisan had some good ideas to improve the PR.

fedimint-server/src/consensus/server.rs
maan2003: resetting after submitting consensus proposal feels like correct behavior?

joschisan:
The mechanism as currently implemented cannot provide reliable low latency since proposed consensus items are not guaranteed to be ordered. Though within every aleph round we order at least consensus items from 2f + 1 peers, at session switch we loose a few rounds of items. Therefore, for a module that actually requires low latency, like lightning legacy, we would have to poll all modules again at the beginning of every session if we were to increase the regular polling time above the current one second. 

Furthermore we should not be polling all modules for items but only the one triggered the low latency poll. Furthermore, we cannot reset the poll timer every time such that other modules are still polled at their intended frequency.

fedimint-server/src/consensus/server.rs
```diff
@@ -838,7 +845,13 @@ async fn submit_module_consensus_items(
...
+                    _ = consensus_proposal_trigger.changed() => {}
+                }
```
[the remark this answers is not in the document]
dpc: Could be done, but does it matter in practice?  I would expect these to be called infrequently, so just grabbing all seems like a simplification without downsides.

dpc:
> The mechanism as currently implemented cannot provide reliable low latency since proposed consensus items are not guaranteed to be ordered.

Not sure if that's a problem. The consensus can't *guarantee* low latency, and that's OK. If it can go slightly faster 99.9% of the time, that's probably good enough. Protocols that do need something better than best effort, should do the effort to avoid Fedimint consensus on the critical path.

The goal here was to do something with a good "value / complexity" ratio.

One code-wise simple improvement could be that after "trigger", the sleep is temporarily reduced for a couple of seconds in addition to just breaking the sleep. That requires only keeping some kind of a counter and resetting it when a trigger was sent. 

> we would have to poll all modules again at the beginning of every session if we were to increase the regular polling time above the current one second

So for these cases increasing the 1s now to 30s could be detrimental and frequent enough that we should just keep the current 1s?

> Furthermore we should not be polling all modules for items but only the one triggered the low latency poll. Furthermore, we cannot reset the poll timer every time such that other modules are still polled at their intended frequency.

What's the the downside of just polling them all? I would expect that it would be infrequent enough to not matter, and I was under impression that 1s here is somewhat arbitrary. Polling a couple of times more often would just temporarily increase the load a bit.

@joschisan

fedimint-server/src/consensus/server.rs
elsirion: Only if all are polled, otherwise one module might starve others.

elsirion:
> Protocols that do need something better than best effort, should do the effort to avoid Fedimint's consensus on the critical path.

Then this change isn't really worth it imo because it's basically just a "make testing faster" thing. I kinda hoped to get a general solution for "modules need lower latency CIs". It might still be a step into that direction, but a rather unsatisfying one :(

joschisan:
> Then this change isn't really worth it imo because it's basically just a "make testing faster" thing. I kinda hoped to get a general solution for "modules need lower latency CIs". It might still be a step into that direction, but a rather unsatisfying one 

I agree with Eric here, a general reliable solution is more involved than the current pr and imo only justified if we have an actual use case that has yet to arrive. However, I can improve the docs, explain the design choice and add that if a module designer still thinks they do need it the should talk to us.

dpc:
> I kinda hoped to get a general solution for "modules need lower latency CIs".

Modules do get lower latency CIs.

If session has around 600s (10minutes), each second has (1000ms / 66ms) 15 epochs, and you loose 3 epochs on the switch, that's 99.9997% CIs getting their latency improved.

If you do the counter that makes a handful of subsequent polls faster, you pretty much getting 100% guarantee that the item will get into consensus ASAP.

> I agree with Eric here, a general reliable solution is more involved than the current pr and imo only justified if we have an actual use case that has yet to arrive.

Once something better is around, you can remove this thing thing.


I'm tired of this PR. I'm asking question (does it cost us anything to poll more frequently), I never get an answer. It's a trivial change that improves vast majority of outcomes, can be removed later. I don't know why I'm still wasting my time here.

Fine lets just have 0-1s extra delay to every CI item submitted via API.

PR #4815, opened 2024-04-03
fix: migrations keep printing on start

dpc:
image

Right now, when trying #4800:

```
fedimint devimint> fedimint-cli dev wait-block-count 500
2024-04-03T18:43:16.364397Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(3)) target_db_vers
ion=DatabaseVersion(3) module_instance_id=0 kind="ln"
2024-04-03T18:43:16.364440Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(0)) target_db_vers
ion=DatabaseVersion(0) module_instance_id=1 kind="mint"
2024-04-03T18:43:16.364461Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(0)) target_db_vers
ion=DatabaseVersion(0) module_instance_id=2 kind="wallet"
2024-04-03T18:43:16.364479Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(0)) target_db_vers
ion=DatabaseVersion(0) module_instance_id=3 kind="meta"
2024-04-03T18:43:16.563345Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(3)) target_db_vers
ion=DatabaseVersion(3) module_instance_id=0 kind="ln"
2024-04-03T18:43:16.563379Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(0)) target_db_vers
ion=DatabaseVersion(0) module_instance_id=1 kind="mint"
2024-04-03T18:43:16.563399Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(0)) target_db_vers
ion=DatabaseVersion(0) module_instance_id=2 kind="wallet"
2024-04-03T18:43:16.563418Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(0)) target_db_vers
ion=DatabaseVersion(0) module_instance_id=3 kind="meta"
2024-04-03T18:43:16.763470Z  INFO fedimint_client::db: Migrating client module database disk_version=Some(DatabaseVersion(3)) target_db_vers
ion=DatabaseVersion(3) module_instance_id=0 kind="ln"
... over and over...
```

I basically moved `info!` but while tweaking I couldn't help but to shorten/simplify the names.

PR #4816, opened 2024-04-03
 chore: clarify client vs server/core side logging 

dpc:
On top of #4815 (inspired by some changes there).

### DIFF
--- a/.config/semgrep.yaml
+++ b/.config/semgrep.yaml
@@ -94,6 +94,22 @@ rules:
       - fedimint-server/
       - fedimint-core/
 
+- id: use-client-log-targets
+  languages:
+    - rust
+  message: Log statements in clients should have static target prefixed with `LOG_CLIENT_` (`client::`)
+  patterns:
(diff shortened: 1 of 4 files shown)


.config/semgrep.yaml
```diff
@@ -94,6 +94,27 @@ rules:
...
+  paths:
+    include:
+      - fedimint-client/
+      - fedimint-cli/
+      - modules/fedimint-ln-client/
+      - modules/fedimint-mint-client/
+      - modules/fedimint-wallet-client/
+      - modules/fedimint-dummy-client/
+      - modules/fedimint-unknown-client/
+      - modules/fedimint-meta-client/
```
maan2003, line 116: can we do glob? no one will add new module here otherwise

dpc: Seems like we can. Lemme try.

elsirion:
More logging, more good. The parent PR is merged now, needs a rebase.

maan2003 (approved):

Issue #4545, opened 2024-03-13
Add string constants for gateway route names


tvolk131:
The route names here should be extracted to constants:

https://github.com/fedimint/fedimint/blob/043f6c3a46ddc6833621daf775bbf59fe1b75b83/gateway/ln-gateway/src/rpc/rpc_server.rs#L53-L92

And those new constants should also be used where these routes are consumed in [rpc_client.rs](https://github.com/fedimint/fedimint/blob/043f6c3a46ddc6833621daf775bbf59fe1b75b83/gateway/ln-gateway/src/rpc/rpc_client.rs).

See [here](https://github.com/fedimint/fedimint/blob/043f6c3a46ddc6833621daf775bbf59fe1b75b83/fedimint-core/src/endpoint_constants.rs) for an example where we do something similar.

m1sterc001guy:
Fixed in the linked PRs

PR #4819, opened 2024-04-03
Define string constants for gateway endpoints

isaacknjama:
Addresses #4545


gateway/ln-gateway/src/endpoint_constants.rs
```diff
@@ -0,0 +1,13 @@
+pub const PAY_INVOICE: &str = "pay_invoice";
```
m1sterc001guy, line 1: nit: would prefer `ENDPOINT` on the end of each of these, like we do for fedimint server endpoints

gateway/ln-gateway/src/rpc/rpc_client.rs
```diff
@@ -36,65 +40,65 @@ impl GatewayRpcClient {
...
+        let url = self
+            .base_url
+            .join(HANDLE_POST_INFO)
+            .expect("invalid base url");
         self.call_post(url, ()).await
     }
 
     pub async fn get_config(&self, payload: ConfigPayload) -> GatewayRpcResult<GatewayFedConfig> {
-        let url = self.base_url.join("/config").expect("invalid base url");
+        let url = self.base_url.join(CONFIGURATION).expect("invalid base url");
```
m1sterc001guy, line 57: You need to change these on the server side too, not just the gateway. You'll also need to change them inside the lightning client. Do a `grep` for each of these strings to find where they're currently being used, then run the tests to make sure nothing is broken and the client can still reach the server's endpoints.

isaack-njama: Done. Would it be necessary to replace all occurrences of the strings in the project?

m1sterc001guy: You have some failing tests like `test_can_change_routing_fees`. Try to debug those tests and find the locations where you need to update the strings.

modules/fedimint-ln-client/src/endpoint_constants.rs
```diff
@@ -0,0 +1,2 @@
+pub const GET_GATEWAY_ID_ENDPOINT: &str = "id";
+pub const PAY_INVOICE_ENDPOINT: &str = "pay_invoice";
```
elsirion, line 2: If this was in `fedimint-ln-common` it would likely be more useful because we could use it for the API client too.

elsirion:
Converted to draft while tests are failing. Please mark as ready for review once they are green and all comments have been addressed. You might also want to squash commits that only addresses comments before doing so.

gateway/ln-gateway/src/rpc/rpc_client.rs
isaack-njama: Done

dpc (approved):
Have someone from @fedimint/lightning approve before landing, but LGTM.

modules/fedimint-ln-client/src/endpoint_constants.rs
m1sterc001guy: Going to echo what Eric said here, we should not have to re-define `GET_GATEWAY_ID_ENDPOINT`.  Your `endpoint_constants.rs` file inside the gateway should go into `fedimint-ln-common` so it can be referenced by both the gateway and the client.

isaack-njama: Done.

modules/fedimint-ln-common/src/endpoint_constants.rs
```diff
@@ -1,13 +1,29 @@
+pub const ADDRESS_ENDPOINT: &str = "/address";
 pub const ACCOUNT_ENDPOINT: &str = "account";
```
m1sterc001guy, line 2: Can we make a new file and separate these? Now its really difficult to tell which endpoints are for the gateway and which are for the federation

isaacknjama: Split the endpoints accordingly between `federation_endpoint_constants` and `gateway_endpoint_constants`

modules/fedimint-ln-common/src/gateway_endpoint_constants.rs
```diff
@@ -0,0 +1,16 @@
+pub const ADDRESS_ENDPOINT: &str = "/address";
+pub const BACKUP_ENDPOINT: &str = "/backup";
+pub const BALANCE_ENDPOINT: &str = "/balance";
+pub const CONFIGURATION_ENDPOINT: &str = "/config";
+pub const CONNECT_FED_ENDPOINT: &str = "/connect_fed";
+pub const CREATE_INVOICE_V2_ENDPOINT: &str = "/create_invoice";
+pub const GATEWAY_INFO_ENDPOINT: &str = "/info";
+pub const GET_GATEWAY_ID_ENDPOINT: &str = "/id";
+pub const HANDLE_POST_INFO_ENDPOINT: &str = "/info";
```
m1sterc001guy, line 9: `HANDLE` is weird. Can this be named `GATEWAY_INFO_POST_ENDPOINT`?

isaacknjama: Done

m1sterc001guy:
LGTM, running CI

m1sterc001guy:
Lint is failing

m1sterc001guy:
I think this needs a rebase to pickup some recent changes

dpc (approved):

dpc (dismissed):

elsirion (approved):

m1sterc001guy:
Can you try rebasing and then running the backwards compatibility tests locally? They're failing CI

dpc (approved):

dpc:
Typo somewhere, that causes all previous versions not be able to communicate and back compat tests fail?

isaacknjama:
> Can you try rebasing and then running the backwards compatibility tests locally? They're failing CI

Running backward compatibility tests:

Command:
CARGO_BUILD_JOBS=1 env TMPDIR=/tmp CARGO_PROFILE=ci nix develop -c   nix shell github:rustshop/fs-dir-cache -c   scripts/ci/run-in-fs-dir-cache.sh backward-compat   env -u RUSTC_WRAPPER   ./scripts/tests/test-ci-all-backcompat.sh v0.2.1   > backward-compatibility-tests.txt

Results:
[backward-compatibility-tests.txt](https://github.com/fedimint/fedimint/files/15212577/backward-compatibility-tests.txt)

dpc:
Commands will print stuff like `--data-dir=/tmp/nix-shell.vapQS2/latency_test_ln_receive-IZa6/devimint-272590-882/`. This is the dir that contains all the stuff like logs of all services. You can inspect `logs/` of `gateway-lnd and `gateway-cln` and look for hints what is wrong.

modules/fedimint-ln-common/src/gateway_endpoint_constants.rs
```diff
@@ -0,0 +1,16 @@
+pub const ADDRESS_ENDPOINT: &str = "/address";
+pub const BACKUP_ENDPOINT: &str = "/backup";
+pub const BALANCE_ENDPOINT: &str = "/balance";
+pub const CONFIGURATION_ENDPOINT: &str = "/config";
+pub const CONNECT_FED_ENDPOINT: &str = "/connect_fed";
```
tvolk131, line 5: This should be `/connect-fed`. I believe this is what's breaking backwards compatibility.

elsirion (approved):

PR #4822, opened 2024-04-04
 chore: standalone devimint tests PoC

dpc:
The fact that `devimint` binary carries all the tests with it is bothering me. I did some experiments how else could we achive it, and it seems to me that compiling devimint tests as standalone binaries is most reasonable.

Test with:

```
> just b && env PATH="$(pwd)/target-nix/debug:$PATH" ./target-nix/debug/meta-dummy-test
```

It's not an Rust integration  test (`./crate/src/tests`) or a unit-test (`#[test]`), as it can't really run without binaries in the `PATH` making it "something else" that requires setup `cargo` will never know about.

It has to be a separate crate, so it doesn't carry `devimint` as a dependency, as we can't use `dev-dependencies`.

### DIFF
--- a/devimint/src/cli.rs
+++ b/devimint/src/cli.rs
@@ -2,6 +2,7 @@ use std::ffi;
 use std::fmt::Write;
 use std::ops::ControlFlow;
 use std::path::{Path, PathBuf};
+use std::time::Duration;
 
 use anyhow::{anyhow, Context, Result};
 use clap::{Parser, Subcommand};
@@ -33,7 +34,7 @@ fn random_test_dir_suffix() -> String {
         .collect::<String>()
 }
 
-#[derive(Parser, Clone)]
+#[derive(Parser, Clone, Default)]
 pub struct CommonArgs {
     #[clap(short = 'd', long, env = FM_TEST_DIR_ENV)]
     pub test_dir: Option<PathBuf>,
@@ -181,10 +182,13 @@ pub async fn cleanup_on_exit<T>(
             Ok(())
         }
         Ok(Ok(v)) => {
-            debug!(target: LOG_DEVIMINT, "Main process finished successfully, will wait for shutdown signal");
-            task_group.make_handle().make_shutdown_rx().await.await;
-            debug!(target: LOG_DEVIMINT, "Received shutdown signal, shutting down");
-            drop(v); // execute destructors
+            debug!(target: LOG_DEVIMINT, "Main process finished successfully, shutting down task group");
+            task_group
+                .shutdown_join_all(Duration::from_secs(30))
+                .await?;
+
+            // drop v here, not before the shutdown
+            drop(v);
             Ok(())
         }
         Ok(Err(err)) => {
--- a/devimint/src/envs.rs
+++ b/devimint/src/envs.rs
@@ -137,3 +137,6 @@ pub const FM_GWCLI_CLN_ENV: &str = "FM_GWCLI_CLN";
 
 // Env variable to define command for the LND client
 pub const FM_GWCLI_LND_ENV: &str = "FM_GWCLI_LND";
+
+/// Make `devimint` print stderr of called commands directly on its own stderr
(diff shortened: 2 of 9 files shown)


devimint/src/cli.rs
```diff
@@ -180,11 +180,12 @@ pub async fn cleanup_on_exit<T>(
...
             Ok(())
         }
-        Ok(Ok(v)) => {
+        Ok(Ok(_v)) => {
             debug!(target: LOG_DEVIMINT, "Main process finished successfully, will wait for shutdown signal");
-            task_group.make_handle().make_shutdown_rx().await.await;
-            debug!(target: LOG_DEVIMINT, "Received shutdown signal, shutting down");
-            drop(v); // execute destructors
+            // TODO: why do we do this?
+            // task_group.make_handle().make_shutdown_rx().await.await;
```
dpc, line 186: Oh. I guess we want to kill all the other stuff that might have gotten spawned.

It hangs for me.

devimint/src/lib.rs
```diff
@@ -14,3 +20,15 @@ pub mod gatewayd;
 pub mod tests;
 pub mod util;
 pub mod vars;
+
+pub async fn run_test<F, FF>(f: F) -> anyhow::Result<()>
+where
+    F: FnOnce(ProcessManager) -> FF,
+    FF: Future<Output = anyhow::Result<()>>,
+{
+    let args = cli::CommonArgs::parse_from::<_, ffi::OsString>(vec![]);
```
dpc, line 29: Actually, we could probably parse command line ATM. I did it this way because I was trying unit tests first, and then test runner args were confusing it.

modules/fedimint-meta-tests/src/bin/meta-dummy-test.rs
```diff
@@ -0,0 +1,12 @@
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    devimint::run_test(|process_mgr| async move {
```
dpc, line 3: Eventually I'd like this to be a builder:

```
devimint::new()
   .with_ln_gateway()
   .with_pegged_in_client()
   .with_dev
   .run(|fed| async {
      // test goes here
    })
```

devimint/src/cli.rs
dpc: Oh, but this is not really correct then. We should send a shutdown and wait on `join_all` kind of thing, not expect someone to shut us down.

modules/fedimint-meta-tests/src/bin/meta-dummy-test.rs
```diff
@@ -0,0 +1,12 @@
...
+        let dev_fed = devimint::devfed::DevJitFed::new(&process_mgr)?;
+
+        dev_fed.finalize(&process_mgr).await?;
+        let client = dev_fed.internal_client().await?;
+        client.balance().await?;
+        Ok(())
+    })
+    .await
+}
```
elsirion, line 12: Can we actually test something about the meta module here? Currently it's more of a test framework test.

devimint/src/envs.rs
```diff
@@ -137,3 +137,6 @@ pub const FM_GWCLI_CLN_ENV: &str = "FM_GWCLI_CLN";
 
 // Env variable to define command for the LND client
 pub const FM_GWCLI_LND_ENV: &str = "FM_GWCLI_LND";
+
+/// Make `devimint` print stderr of called commands directly on its own stderr
+pub const FM_DEVIMINT_CMD_INHERIT_STDERR_ENV: &str = "FM_DEVIMINT_CMD_INHERIT_STDERR";
```
elsirion, line 142: Where would which of both options be appropriate?

elsirion (dismissed):
I like this approach of using devimint for more module-specific tests in a more modularized way!

modules/fedimint-meta-tests/src/bin/meta-dummy-test.rs
dpc: Now that I know we're fine, I've converted #4772 to this approach.

dpc:
Will make #4772 use this approach.

PR #4825, opened 2024-04-04
fix(devimint): shutdown on success sequence looks weird

dpc:
And hangs for me.


bradleystachurski (approved):

justinmoon (approved):

dpc:
@maan2003 even if it lands, would be good if you took a look. We can always revert.

maan2003 (commented):
Approved, makes sense

PR #4828, opened 2024-04-04
refactor(devimint): define version constants

bradleystachurski:
@m1sterc001guy had some good feedback in https://github.com/fedimint/fedimint/pull/4804#discussion_r1548781274

>Just a thought: do we want to put all of our version constants in a central place (maybe a new file?). Might help with reasoning about which versions have changes between them.

Let's give it a shot.

### DIFF
--- a/devimint/Cargo.toml
+++ b/devimint/Cargo.toml
@@ -34,6 +34,7 @@ fedimint-server = { path = "../fedimint-server" }
 fedimint-testing = { path = "../fedimint-testing" }
 futures = { workspace = true }
 hex = { workspace = true }
+lazy_static = "1.4.0"
 ln-gateway = { package = "fedimint-ln-gateway", path = "../gateway/ln-gateway" }
 nix = { version = "0.28.0", features = ["signal"] }
 rand = { workspace = true }
--- a/devimint/src/version_constants.rs
+++ b/devimint/src/version_constants.rs
@@ -1,0 +1,7 @@
+use lazy_static::lazy_static;
+use semver::Version;
+
+lazy_static! {
+    pub static ref VERSION_0_3_0_ALPHA: Version =
+        Version::parse("0.3.0-alpha").expect("version is parsable");
+}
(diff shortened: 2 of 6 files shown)


m1sterc001guy (approved):
Saves some parsing!

devimint/Cargo.toml
```diff
@@ -34,6 +34,7 @@ fedimint-server = { path = "../fedimint-server" }
 fedimint-testing = { path = "../fedimint-testing" }
 futures = { workspace = true }
 hex = { workspace = true }
+lazy_static = "1.4.0"
```
dpc, line 37: BTW. We should stop using lazy_static

https://morestina.net/blog/2055/rust-global-variables-two-years-on

dpc (approved):
If passes CI, good enough for me.

devimint/Cargo.toml
tvolk131: Should we create an issue to remove existing usage and add a semgrep to block it going forward?

dpc: @tvolk131 :+1:

devimint/src/version_constants.rs
```diff
@@ -0,0 +1,7 @@
+use lazy_static::lazy_static;
+use semver::Version;
+
+lazy_static! {
+    pub static ref VERSION_0_3_0_ALPHA: Version =
```
dpc, line 5: Since `Version` fields are public, you could just make it a const, and avoid any locks etc. https://docs.rs/semver/latest/semver/struct.Version.html#

dpc: Oh, actually there might be no const ways to parse the build & pre. Oh well.

bradleystachurski: Yeah, at some point we'll need to parse and handle a result because of the prerelease.

devimint/Cargo.toml
bradleystachurski: Neat!

Issue #4826, opened 2024-04-04
Always add `context` to decoding errors being propagated


elsirion:
This should make debugging decoding failures much easier.

elsirion:
Fixed in [#4829](https://github.com/fedimint/fedimint/pull/4829)

PR #4829, opened 2024-04-04
chore: add better error reporting to derived `Decodable` trait impls

elsirion:
Working towards #4826.

### DIFF
--- a/fedimint-derive/src/lib.rs
+++ b/fedimint-derive/src/lib.rs
@@ -200,6 +200,7 @@ pub fn derive_decodable(input: TokenStream) -> TokenStream {
     let output = quote! {
         impl ::fedimint_core::encoding::Decodable for #ident {
             fn consensus_decode_from_finite_reader<D: std::io::Read>(d: &mut D, modules: &::fedimint_core::module::registry::ModuleDecoderRegistry) -> std::result::Result<Self, ::fedimint_core::encoding::DecodeError> {
+                use ::fedimint_core:: anyhow::Context;
                 #decode_inner
             }
         }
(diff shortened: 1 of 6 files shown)


fedimint-derive/src/lib.rs
```diff
@@ -248,7 +248,15 @@ fn derive_enum_decode(ident: &Ident, variants: &Punctuated<Variant, Comma>) -> T
             // FIXME: make sure we read all bytes
             quote! {
                 #variant_idx => {
-                    let bytes: Vec<u8> = ::fedimint_core::encoding::Decodable::consensus_decode_from_finite_reader(d, modules)?;
+                    // FIXME: feels like there's a way more elegant way to do this with limited readers
+                    let bytes_res = ::fedimint_core::encoding::Decodable::consensus_decode_from_finite_reader(d, modules);
+                    let bytes: Vec<u8> = ::anyhow::Context::context(
```
dpc, line 253: I don't understand why this isn't just `.context(....)` on the same line.

dpc: Is putting `use anyhow::Context as _;` right here an issue?

dpc (approved):

fedimint-derive/src/lib.rs
elsirion: You are right, I could put that into the top-level derive fn that defines the decode fn. I got a bit hung up on not being able to `use` it where I need it since that would import it multiple times leading to errors.

elsirion: fixed

fedimint-derive/src/lib.rs
```diff
@@ -257,7 +263,12 @@ fn derive_enum_decode(ident: &Ident, variants: &Punctuated<Variant, Comma>) -> T
...
                     if read_bytes != total_bytes {
                         return Err(::fedimint_core::encoding::DecodeError::new_custom(anyhow::anyhow!(
-                            "Partial read: got {total_bytes} bytes but only read {read_bytes}"
+                            concat!(
+                                "Partial read: got {total_bytes} bytes but only read {read_bytes} when decoding ",
+                                stringify!(#ident),
+                                "::",
+                                stringify!(#variant)
+                            )
                         )));
```
maan2003, line 272: doesn't work as expected, anyhow! will not format the string

concat!(..) will be treated as `expr` not `literal`

elsirion: will fix tomorrow

maan2003 (approved):

dpc (approved):

PR #4832, opened 2024-04-04
chore(devimint): add semgrep rule for VersionReq

bradleystachurski:
Prevents reintroducing `VersionReq`, which doesn't work with how we compare versions in devimint (see: https://github.com/fedimint/fedimint/pull/4804).


tvolk131 (approved):
LGTM! My only thought is that it might be helpful to include a reference in the error description to the relevant issue/conversation thread where it was initially discovered that `VersionReq` doesn't work

elsirion (approved):

maan2003 (approved):

PR #4836, opened 2024-04-04
fix(meta): (mostly harmless) race condition (attempt 2)

dpc:
on clearing the outstanding minority submissions.

See https://github.com/fedimint/fedimint/pull/4772#issuecomment-2033170196

See https://github.com/fedimint/fedimint/pull/4803/files#r1552459723

Seems like never deleting desired values is the simplest correct solution, even if it can potentially waste some storage.

### DIFF
--- a/modules/fedimint-meta-server/src/lib.rs
+++ b/modules/fedimint-meta-server/src/lib.rs
@@ -204,44 +204,6 @@ pub struct Meta {
 }
 
 impl Meta {
-    /// Check the difference between what's desired to be submitted and what's
-    /// already submitted or consensus value.
-    ///
-    /// Returns:
-    /// `(items_to_submit, desired_keys_to_delete)`
-    async fn desired_submitted_diff(
-        &self,
-        dbtx: &mut DatabaseTransaction<'_, NonCommittable>,
-    ) -> (Vec<MetaConsensusItem>, Vec<MetaKey>) {
-        let desired: Vec<_> = Self::get_desired(dbtx).await;
-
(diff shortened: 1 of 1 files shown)


modules/fedimint-meta-server/src/lib.rs
```diff
@@ -204,42 +204,10 @@ pub struct Meta {
 }
 
 impl Meta {
-    /// Check the difference between what's desired to be submitted and what's
-    /// already submitted or consensus value.
-    ///
-    /// Returns:
-    /// `(items_to_submit, desired_keys_to_delete)`
     async fn desired_submitted_diff(
```
dpc, line 207: Damn it.

elsirion (approved):
Nice simplification at the cost of at most `f` more CIs being sent and the DB having up to 256 more entries.

joschisan (approved):

modules/fedimint-meta-server/src/lib.rs
```diff
@@ -298,15 +260,60 @@ impl ServerModule for Meta {
...
+                } else {
+                    // we want to submit the same value as the current consensus, usually
+                    // to clear the previous submission that did not became the consensus (we were
+                    // outvoted)
+                    to_submit.push(MetaConsensusItem {
+                        key,
+                        value: desired_value,
+                        salt,
+                    });
+                }
```
elsirion, line 307: When looking at it with @joschisan we noticed that this could be further simplified, but getting a correct version of the code was more important to me than reducing the complexity to its minimum. We can do that later since it's only the submission logic.

PR #4838, opened 2024-04-05
fix: panic in update_gateway_cache if called concurrently

maan2003:


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -910,21 +910,29 @@ impl LightningClientModule {
...
-        let mut dbtx = self.client_ctx.module_db().begin_transaction().await;
-
-        // Remove all previous gateway entries
-        dbtx.remove_by_prefix(&LightningGatewayKeyPrefix).await;
-
-        for gw in gateways.iter() {
-            dbtx.insert_entry(
-                &LightningGatewayKey(gw.info.gateway_id),
-                &gw.clone().anchor(),
+        self.client_ctx
```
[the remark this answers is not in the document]
elsirion: Commits are atomic, right? Can it even happen that *neither* of the two succeed?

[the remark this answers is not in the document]
dpc: @elsirion 

They are atomic and both will eventually succeed, but not necessarily in the same order they started.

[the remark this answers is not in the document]
elsirion: I see now, yeah, not ideal but honestly also likely not a big deal at all :sweat_smile:

I'm open to an alternative lock-based version, but we should likely build an abstraction for that by now, we have a bunch of places where we use `Mutex<()>`/`Mutex<Option<Error>>` to block synchronize some cache update. @maan2003

elsirion:
CI failing :(

modules/fedimint-ln-client/src/lib.rs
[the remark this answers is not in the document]
dpc: It's not a big deal for sure. Purely theoretical, AFAICT.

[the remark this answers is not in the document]
maan2003: is it a race?
it is just (one of) expected behavior of calling update_gateway_cache concurrently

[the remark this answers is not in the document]
maan2003: the user introduced race by calling update_gateway_cache concurrently, it will also exist in the lock version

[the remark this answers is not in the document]
dpc: Without auto-retry there is no possibility of race. There's a possibility of panic though. :D

maan2003:
changed to mutex, maybe that will fix ci

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -910,6 +913,8 @@ impl LightningClientModule {
     /// See also [`Self::update_gateway_cache_continuously`].
     pub async fn update_gateway_cache(&self) -> anyhow::Result<()> {
         let gateways = self.module_api.fetch_gateways().await?;
+        // held to prevent transaction conflict
+        let _g = self.update_gateway_cache_lock.lock().await;
```
elsirion, line 917: Can't we do a try_lock before and if that failed return immediately after actually acquiring the lock and instantly releasing it since some else already updated the cache?

maan2003: baited me to properly implementing merging :smile:

dpc: If the other thread failed, that would tell the caller things were updated while they weren't.

maan2003: now it is implemented in new struct, it is much better now

fedimint-core/src/util/update_merge.rs
```diff
@@ -0,0 +1,41 @@
...
+        UpdateMerge {
+            last_failed: Mutex::new(false),
+        }
+    }
+}
+impl UpdateMerge {
+    // if two update are called concurrently then calls are merge.
+    // if first call failed then the second call is run again.
+    // `fut` is never ran concurrently.
+    pub async fn merge<E>(&self, fut: impl Future<Output = Result<(), E>>) -> Result<(), E> {
```
maan2003, line 20: @coderabbitai could you please write docs :)

fedimint-core/src/util/update_merge.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// The future `fut` is never executed concurrently.
+    pub async fn merge<E>(&self, fut: impl Future<Output = Result<(), E>>) -> Result<(), E> {
+        let mut guard = if let Ok(gaurd) = self.last_failed.try_lock() {
+            // not running => run now
+            gaurd
+        } else {
+            // already running concurrently
+            // wait for other call to return
+            let guard = self.last_failed.lock().await;
+            match *guard {
```
dpc, line 31: `match` on a bool? Hmmm...

maan2003: match >> if

fedimint-core/src/util/update_merge.rs
```diff
@@ -0,0 +1,47 @@
...
+                // last call completed successfully
+                // => merge the call
+                false => return Ok(()),
+                // last call failed
+                // => run again
+                true => guard,
+            }
+        };
+        // future may panic, use mark as failed initially
+        *guard = true;
```
dpc, line 41: So dilligent. Very correct. Much :+1: .

dpc (approved):
I don't know why... but I think it we went through all failure modes and this works. :D

fedimint-core/src/util/update_merge.rs
dpc: You think it's Match better?

maan2003:
typo fix, it is so hard to spell guard

fedimint-core/src/util/update_merge.rs
maan2003: it reads better with comments here

dpc (approved):

elsirion (approved):

elsirion:
Beautiful abstraction!

PR #4846, opened 2024-04-05
fix: break lockstep between peers in a round by adding jitter

dpc:
AlephBFT each round will basically take the first N responses from peers, and ignore other ones.

What can happen in a devimint run is that each `fedimintd` runs the exactly same loop but offseted by some tiny, tiny time, because we start peers in a `for _ in 0..N`.

So a last peer can get ignored consistently for a long while.

Adding any jitter breaks the lockstep immediately, but AFAIK in distributed systems it's a good practice to just use +/- 50% to immediately break any corelations, so we went with `0.5..=1.5`.


elsirion (approved):
Fascinating!

joschisan (approved):

PR #4850, opened 2024-04-05
feat: lets break tokio rt on wasm

maan2003:


dpc:
@maan2003 If the tokio's `block_on` fails on wasm, we should have a conditional `fedimint_core::task::block_on` and enforce using that with `semgrep`?

maan2003:
makes sense, we may want to just avoid block_on in client code

maan2003:
nice, finally failing

elsirion:
#4854 is a reaction to this, right?

maan2003:
yep

elsirion:
Can this be closed then?

PR #4858, opened 2024-04-06
feat: gateway filter for update_gateway_cache_continuously

maan2003:
allows impls to ignore non vetted gateways for example

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -56,7 +56,7 @@ use fedimint_ln_common::{
     LightningGatewayAnnouncement, LightningGatewayRegistration, LightningInput,
     LightningModuleTypes, LightningOutput, LightningOutputV0,
 };
-use futures::{FutureExt, StreamExt};
+use futures::{Future, FutureExt, StreamExt};
 use incoming::IncomingSmError;
 use lightning_invoice::{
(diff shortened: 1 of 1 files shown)


elsirion:
Do you generally want to ignore them? If e.g. none were available anymore we'd probably want to fall back to unvetted ones. In that case that fn would need to hold some `Receiver<_>`/`Arc<Mutex_>>` to be told about that. Not sure this abstraction will stand the test of time.

maan2003:
change gateway filter from `Fn(LightningGateway) -> bool` to `Fn(Vec<LightningGatewayAnnouncement>) -> Vec<LightningGatewayAnnouncement>`

the caller can detect if there are no vetted gateways

maan2003:
this is just a filter what gateways we check for ttl, to prevent too many wakeups.

this doesn't change gateway selection

m1sterc001guy:
Ah I see, still early for me :)

m1sterc001guy (approved):

elsirion:
> this doesn't change gateway selection

The problem is that if you *might* select an unvetted GW eventually you can't tell upfront to filter out all unvetted gateways since then the selected one might expire and be unvailable. I think this can be fixed using channels, but I wonder if there's a more elegant abstraction.

elsirion (approved):

maan2003:
i think gateway_filters = 
```
|gws| if gws.iter(),any(vetted) { gws.into_iter().filter(vetted).collect() } else { gws }
```

will work pretty well.

PR #4854, opened 2024-04-05
chore: unify tokio runtime primitives in a new module & fix some wasm issues

dpc:
And add some improvements while at it.

PR #4864, opened 2024-04-07
feat: simplify JitCore (take 2)

maan2003:
&mut Future impls Future

we don't need to model None case, oneshot::Receiver already does that in case Future completed once.


maan2003:
rebased on top https://github.com/fedimint/fedimint/pull/4854

and using join handles now

fedimint-core/src/task/jit.rs
```diff
@@ -87,89 +86,42 @@ where
...
+                        // return ()
                     })
-                };
-
-                recv.cancel();
-
-                val_res.map_err(|err| OneTimeError(Some(err)))
             })
-            .await
+            .await;
```
dpc, line 120: I had some doubts if *technically* it's cancleation safe and doesn't lose the error, but I think it's OK.

No idea why the you don't like the `Drop` thing so much. It is lenghtly, because it uses a lot of boilerplate. If it was a `scopeguard` it could have been a oneliner-ish, but didn't want to bring a new crate for that.

fedimint-core/src/task/jit.rs
```diff
@@ -51,15 +49,13 @@ where
...
-    val_rx:
-        Arc<std::sync::Mutex<Option<tokio::sync::oneshot::Receiver<std::result::Result<T, E>>>>>,
-    val: Arc<sync::OnceCell<T>>,
+    inner: Arc<JitInner<T, E>>,
+}
+
+#[derive(Debug)]
+struct JitInner<T, E> {
+    handle: sync::Mutex<JoinHandle<Result<T, E>>>,
+    val: sync::OnceCell<Result<T, ()>>,
```
dpc, line 58: These parts seem like an improvement.

fedimint-core/src/task/jit.rs
```diff
@@ -87,89 +86,42 @@ where
...
-                    };
-                    recv.await.unwrap_or_else(|_| {
-                        panic!("Jit value {} panicked", std::any::type_name::<T>())
+    pub async fn get_try(&self) -> Result<&T, OneTimeError<E>> {
+        let mut init_error = None;
+        let value = self
+            .inner
+            .val
+            .get_or_init(|| async {
+                let handle: &mut _ = &mut *self.inner.handle.lock().await;
```
dpc, line 111: Actually ... no. It moves the handle out here, and there's a `await` 2 lines later where the handle could be dropped.

AFAIK `Drop` is the only reliable tool against cancellation safety.

maan2003: are you sure, it doesn't move out it

handle: &mut JoinHandle

and tokio says `&mut JoinHandle` is cancellation safe 
https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#cancel-safety

maan2003: if &mut JoinHandle poll becomes Ready, we don't have any more .await points, it will always preserve the error.

dpc: Ah. Nevermind. I got confused, seen things that weren't there.

maan2003: the old drop guard code is round-about way to get `&mut Fut`

dpc (dismissed):
I ran out of things to complain about.

fedimint-core/src/task/jit.rs
dpc: Now that I look at it, I don't understand how is it possible to await the result of the future on `&mut JoinHandle`, without consuming it. What happens if you call it again?  :D

maan2003: > Now that I look at it, I don't understand how is it possible to await the result of the future on &mut JoinHandle, without consuming it. What happens if you call it again? :D

`Future::poll` only takes `Pin<&mut Self>`.

from https://doc.rust-lang.org/stable/std/future/trait.Future.html

> Once a future has completed (returned Ready from poll), calling its poll method again may panic, block forever, or cause other kinds of problems; the Future trait places no requirements on the effects of such a call. However, as the poll method is not marked unsafe, Rust’s usual rules apply: calls must never cause undefined behavior (memory corruption, incorrect use of unsafe functions, or the like), regardless of the future’s state.

In practice, good libraries always have FusedFutures, i.e. they return `Pending` after they become `Ready` once.

maan2003:
just rebased on parent

dpc (approved):

elsirion (approved):
Very cool solution :exploding_head:

elsirion:
Let's get [#4854](https://github.com/fedimint/fedimint/pull/4854) in first.

maan2003:
rebased https://github.com/fedimint/fedimint/pull/4854

dpc (approved):

bradleystachurski (approved):

PR #4865, opened 2024-04-07
fix: only deduplicate transactions, not consensus items

joschisan:


dpc (approved):
Oh, this will make so that consensus items (non-transactions) are delivered every time?

Is that OK? I thought that deduplication was there on purpose, so that modules can keep returning the same item for simplicity.

Anyway - just double checking. I guess I'll have to document somewhere that the "salt" I aded to meta module CIs is no longer neccessary.

joschisan:
Yes consensus items are now ordered every time, this gets inefficient if consensus_proposal is polled to often. I originally only wanted to remove this deduplication when we can increase the polling time after removal of lightning legacy.

justinmoon:
dev call: is this a consensus-breaking change in middle of a session? some questions about rollout.

joschisan:
> dev call: is this a consensus-breaking change in middle of a session? some questions about rollout.

No. It’s just about proposing items and has nothing to do with ordered items.

PR #4872, opened 2024-04-08
chore: update docker files to 0.3.0

douglaz:


Kodylow:
tACK both guardian (solo + 2 federations) and gateway

Kodylow (approved):
tACK

docker/0.3.0/full-tls-mutinynet/docker-compose.yaml
```diff
@@ -0,0 +1,359 @@
...
+
+        VirtualHost "xmpp.fedimint.my-super-host.com"
+          enabled = true;
+        EOF
+        tail -f /dev/null
+    volumes:
+      - prosody_datadir:/prosody_data
+    restart: always
+
+  xmpp:
```
elsirion, line 334: I think that shouldn't be in here

elsirion (approved):
I think it would be nice to have a version that is just a guardian server with TLS proxy and admin UI on the same domain. That's probably the most common use case.

How we can achieve single-domain:
* `mydomain.com`: admin UI
* `mydomain.com/ws`: websocket API of fedimintd

I've used this with a Nix-based setup in the past and it works well.

dpc (approved):

PR #4879, opened 2024-04-08
chore(client): report module_id on too large module backup

dpc:


fedimint-client/src/backup.rs
```diff
@@ -142,13 +142,17 @@ impl ClientBackup {
                     let size_previous = last_module_backup.consensus_encode_to_len();
                     warn!(
                         size,
-                        size_previous, limit, "Module backup too large, will use previous version"
+                        limit,
+                        %module_id,
```
maan2003, line 146:
```suggestion
                        module_id,
```

maan2003: actually nvm, it is not number

maan2003: oh it is u16

maan2003 (approved):

m1sterc001guy (approved):

Issue #4584, opened 2024-03-16
Versions support tracking


dpc:
I think we need:

* `fedimint-cli dev` subcommands that dump:
   * All server side supported API versions (coree + all discovered modules).
   * All client side support API versions.
   * Client-side db schema versions.
   * Server-side db schema versions.
* On every release run these and dump the output to some `./releases/api-versions.md` file so we can refer to it during development.

Rationale:

* As things are right now it's hard to answer the questions like "did we already increase the minor API version on this new module we are adding new api endpoint to?" or "In the next release what is the minimum API version we can expect to have?"

Updating this file should become a a part of automated release process.

dpc:
IIRC Someone already implemented dumping currently negotiated and used API version set. This new one would be somewhat like that.

dpc:
@Kodylow for awareness in case you are looking for good issues for new contributors.

emilioziniades:
Hi @dpc, I'm new around here but interested in having a go at this!

dpc:
@emilioziniades Go for it. Feel free to ask for help on Discord.

0xSaksham:
> IIRC Someone already implemented dumping currently negotiated and used API version set. 

Hey @dpc, where can I find it? For reference.

dpc:
@0xSaksham 

```
devimint> fedimint-cli discover-version
```

emilioziniades:
@dpc I think it makes sense to break up this work into three separate PRs:
- [ ] server-side "list versions" subcommand (PR: #4882)
- [ ] client-side "list versions" subcommand
- [ ] CI work to capture the output of these commands on every release.

What do you think?

dpc:
@emilioziniades Sounds great.

PR #4882, opened 2024-04-08
`fedimintd dev list-versions` subcommand

emilioziniades:
The server-side work for #4584.

Creates an optional subcommand `dev` for fedimintd, similar to fedimint-cli. 

~~Adds a single subcommand to `dev`: `list-versions`.This dumps all API and database versions and exits early.~~ 
Adds two subcommands to `dev`: `list-api-versions` and `list-db-versions`. These dump API and database versions respectively and exit early.

e.g.
```
FM_BITCOIN_RPC_KIND="bitcoind" FM_BITCOIN_RPC_URL="http://localhost:5002" fedimintd dev list-versions
```

<details>
<summary>And here's a sample of the output</summary>

```
{
  "core_api_versions": [
    {
      "major": 0,
      "minor": 2
    }
  ],
  "modules": [
    {
      "kind": "ln",
      "api_versions": [
        {
          "major": 0,
          "minor": 1
        }
      ],
      "database_version": 0
    },
    {
      "kind": "meta",
      "api_versions": [
        {
          "major": 0,
          "minor": 0
        }
      ],
      "database_version": 0
    },
    {
      "kind": "mint",
      "api_versions": [
        {
          "major": 0,
          "minor": 0
        }
      ],
      "database_version": 0
    },
    {
      "kind": "wallet",
      "api_versions": [
        {
          "major": 0,
          "minor": 0
        }
      ],
      "database_version": 0
    }
  ]
}
```

</details>

Unfortunately, the FM_BITCOIN_RPC_KIND and FM_BITCOIN_RPC_URL environment variables still need to be provided because otherwise fedimintd throws an error in `Fedimintd::new`.

It seems weird that they aren't part of the explicit CLI options but rather grabbed from the environment later:
https://github.com/fedimint/fedimint/blob/11c24a949840b317d52f85e282c1d43cabb2bafe/fedimintd/src/fedimintd.rs#L182

I did consider refactoring this into the `ServerOpts` struct (clap already supports reading from environment variables), but this doesn't resolve the issue that they are still required fields.

Basically I'd love to be able to do `fedimint dev list-versions` without any additional args (cli or env-vars), but wasn't sure how to get there. And I thought I'd rather get some feedback on what I've done before proceeding.

### DIFF
--- a/fedimint-core/src/module/version.rs
+++ b/fedimint-core/src/module/version.rs
@@ -58,7 +58,8 @@ use std::{cmp, result};
 
 use serde::{Deserialize, Serialize};
 
-use crate::core::ModuleInstanceId;
+use crate::core::{ModuleInstanceId, ModuleKind};
+use crate::db::DatabaseVersion;
 use crate::encoding::{Decodable, Encodable};
 
 /// Consensus version of a core server
@@ -441,4 +442,17 @@ impl SupportedModuleApiVersions {
 pub struct SupportedApiVersionsSummary {
     pub core: SupportedCoreApiVersions,
     pub modules: BTreeMap<ModuleInstanceId, SupportedModuleApiVersions>,
+}
+
+/// A summary of server API versions for core and all registered modules.
+#[derive(Serialize)]
+pub struct ServerApiVersionsSummary {
+    pub core: MultiApiVersion,
+    pub modules: BTreeMap<ModuleKind, MultiApiVersion>,
+}
+
+/// A summary of server database versions for all registered modules.
+#[derive(Serialize)]
+pub struct ServerDbVersionsSummary {
+    pub modules: BTreeMap<ModuleKind, DatabaseVersion>,
 }
--- a/fedimintd/src/fedimintd.rs
+++ b/fedimintd/src/fedimintd.rs
@@ -56,7 +56,7 @@ const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
 pub struct ServerOpts {
     /// Path to folder containing federation config files
     #[arg(long = "data-dir", env = FM_DATA_DIR_ENV)]
-    pub data_dir: PathBuf,
+    pub data_dir: Option<PathBuf>,
     /// Password to encrypt sensitive config files
     // TODO: should probably never send password to the server directly, rather send the hash via
     // the API
@@ -441,6 +441,8 @@ async fn run(
         });
     }
 
+    let data_dir = opts.data_dir.context("data-dir option is not present")?;
+
     // TODO: Fedimintd should use the config gen API
     // on each run we want to pass the currently passed password, so we need to
     // overwrite
@@ -464,11 +466,11 @@ async fn run(
     };
 
     let db = Database::new(
-        fedimint_rocksdb::RocksDb::open(opts.data_dir.join(DB_FILE))?,
+        fedimint_rocksdb::RocksDb::open(data_dir.join(DB_FILE))?,
         Default::default(),
     );
 
-    FedimintServer::new(opts.data_dir, settings, db, version_hash)
+    FedimintServer::new(data_dir, settings, db, version_hash)
         .run(&module_inits, task_group.clone())
         .await?;
 
--- a/fedimintd/src/fedimintd.rs
+++ b/fedimintd/src/fedimintd.rs
@@ -5,8 +5,8 @@ use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::time::Duration;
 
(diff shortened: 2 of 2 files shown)


elsirion (dismissed):
tACK, idk if we really want to make it a `fedimintd` subcommand, but I remember @dpc discussing the direction to be taken in the issue, I'll leave it to you.

```
$ FM_BITCOIN_RPC_KIND="bitcoind" FM_BITCOIN_RPC_URL="http://127.0.0.1" cargo run --bin fedimintd -- dev list-versions
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target-nix/debug/fedimintd dev list-versions`
{
  "core_api_versions": [
    {
      "major": 0,
      "minor": 2
    }
  ],
  "modules": [
    {
      "kind": "ln",
      "api_versions": [
        {
          "major": 0,
          "minor": 1
        }
      ],
      "database_version": 0
    },
    …
  ]
}
```

An alternative I could imagine: add the DB version (even though it's a `fedimintd` internal) to the `version` API endpoint.

```
$ fedimint-cli -- --data-dir cfg/… dev api version
{
  "value": {
    "core": {
      "core_consensus": {
        "major": 2,
        "minor": 0
      },
      "api": [
        {
          "major": 0,
          "minor": 1
        }
      ]
    },
    "modules": {
      "0": {
        "core_consensus": {
          "major": 2,
          "minor": 0
        },
        "module_consensus": {
          "major": 2,
          "minor": 0
        },
        "api": [
          {
            "major": 0,
            "minor": 0
          }
        ]
      },
      …
    }  
  }
}
```

dpc:
Try to keep the format match the existing:

```
fedimint devimint> fedimint-cli discover-version
{
  "versions": {
    "core": {
      "major": 0,
      "minor": 2
    },
    "modules": {
      "0": {
        "major": 0,
        "minor": 1
      },
      "1": {
        "major": 0,
        "minor": 0
      },
      "2": {
        "major": 0,
        "minor": 0
      },
      "3": {
        "major": 0,
        "minor": 0
      }
    }
  }
}
```

in particular: modules should be a map (keyed on the `kind` like you do already, not an id, since we talk about static-support, not the config).

I'm guessing two sub-commands, one for db, the other for api will be the easiest to use for the consuming scripts.

> Unfortunately, the FM_BITCOIN_RPC_KIND and FM_BITCOIN_RPC_URL environment variables still need to be provided 

The pattern we use in `fedimint-cli` for same problem is making such fields `Option`, private, and add a `pub $name(&self) -> Result<$type>` getter that returns a descriptive error (`"$name needs to be provided"`). This way we only error out if the field is really needed.

dpc (approved):
LGTM, but please get one more opinion before landing.

emilioziniades:
> > Unfortunately, the FM_BITCOIN_RPC_KIND and FM_BITCOIN_RPC_URL environment variables still need to be provided
> 
> The pattern we use in `fedimint-cli` for same problem is making such fields `Option`, private, and add a `pub $name(&self) -> Result<$type>` getter that returns a descriptive error (`"$name needs to be provided"`). This way we only error out if the field is really needed.

I considered this, but the problem is that the fields are required in `Fedimintd::new`, so they are needed early on in the initialization process.

https://github.com/fedimint/fedimint/blob/96b93f011815a22d1c022e4751167c2e87302834/fedimintd/src/fedimintd.rs#L183

Would sensible default values make sense here? e.g. FM_BITCOIN_RPC_KIND defaults to "bitcoind" and FM_BITCOIN_RPC_URL defaults to "http://localhost:8332"

emilioziniades:
This is how the output looks for each subcommand:
```
$ FM_BITCOIN_RPC_KIND="bitcoind" FM_BITCOIN_RPC_URL="http://localhost:8332" cargo run --quiet --bin fedimintd -- dev list-api-versions
{
  "core": [
    {
      "major": 0,
      "minor": 2
    }
  ],
  "modules": {
    "ln": [
      {
        "major": 0,
        "minor": 1
      }
    ],
    "meta": [
      {
        "major": 0,
        "minor": 0
      }
    ],
    "mint": [
      {
        "major": 0,
        "minor": 0
      }
    ],
    "wallet": [
      {
        "major": 0,
        "minor": 0
      }
    ]
  }
}

$ FM_BITCOIN_RPC_KIND="bitcoind" FM_BITCOIN_RPC_URL="http://localhost:8332" cargo run --quiet --bin fedimintd -- dev list-db-versions
{
  "modules": {
    "ln": 0,
    "meta": 0,
    "mint": 0,
    "wallet": 0
  }
}
```

elsirion (approved):

PR #4888, opened 2024-04-09
fix: dont pay an invoice twice

joschisan:


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -375,6 +375,32 @@ impl LightningClientModule {
...
+            if self.client_ctx.has_active_states(operation_id).await {
+                bail!("Previous payment attempt still in progress. Previous Operation Id: {operation_id}");
+            } else if self.client_ctx.has_inactive_states(operation_id).await {
+                let mut stream = self
+                    .subscribe_send(operation_id)
+                    .await
+                    .expect("operation_id has inactive states")
+                    .into_stream();
+
+                while let Ok(state) = stream.ok().await {
```
elsirion, line 391: If all you do is matching `Ok(_)` then you could just use `next` here and match `Some(_)`.

elsirion (approved):
Looks correct, just not very pretty/performant.

joschisan:
> Looks correct, just not very pretty/performant.

On the performance has_active_states / hash_inactive_states is implemented very inefficiently right now, but could be changed into a efficient db lookup so the happy path is quick. So we would not query all the operation states to check if any exist anymore.

subscribe_send is only called on second attempts, so inefficiency would not that big of a problem or still?

elsirion:
How do you feel about a separate DB entry to track the number of retries per payment hash? Not super important after thinking about it more (retrying should be rare as you say).

joschisan:
> How do you feel about a separate DB entry to track the number of retries per payment hash? Not super important after thinking about it more (retrying should be rare as you say).

What would you want to achieve with that?

elsirion:
I just realized: we can even make this version more efficient (original response below):
* Only check if the operation exists each round, find the highest index `n` for which an operation exists (lookups in the operation look should be more efficient since they only fetch a single entry).
  * You know that you won't have created a `i+1`-st operation without the `i`-th one having failed
* Now only check the operation state for operation `n` by subscribing to the update stream (you probably want a way to avoid waiting in case the operation isn't finished yet and return an error?)
* Create operation id `n+1`

Original response:

You save `n-1` checks for failed active/inactive states and operation outcomes on the `n`-th call, making it less of a [Schlemiel the painter algorithm](https://en.wikichip.org/wiki/schlemiel_the_painter%27s_algorithm) and don't have to check active/inactive states explicitly but just check the operation outcome for the `n`-th one.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -375,6 +375,34 @@ impl LightningClientModule {
         Ok(operation_id)
     }
 
+    async fn get_operation_id(&self, invoice: &Bolt11Invoice) -> anyhow::Result<OperationId> {
```
elsirion, line 378: Something I totally missed but you reminded me of the other day: you probably want to do all that in a single database transaction (together with spawning the payment state machines) so that concurrent calls to the pay function don't cause havoc (can easily happen due to UI bugs that such a function is called concurrently).

joschisan: If we get a collision of operation ids do to concurrent calls with the same invoice finalize_and_submit_transaction will return the appropriate error which we return so its safe. If we do it in one dbtx we would get the more appropriate error of "Previous payment attempt still in progress.", but achieving this is seems much more involved.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -375,6 +375,34 @@ impl LightningClientModule {
...
+                .subscribe_send(operation_id)
+                .await
+                .expect("operation_id exists")
+                .into_stream();
+
+            while let Some(state) = stream.next().await {
+                if let SendState::Success(..) = state {
+                    bail!("Previous payment attempt was successful. Previous Operation Id: {operation_id}");
+                }
+            }
```
elsirion, line 400: nit: maybe leave a comment here why this won't potentially block for a long time. We can only assume that because we checked for active states and there were none, so by definition a final state has to have been reached.

elsirion (approved):

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -380,6 +380,39 @@ impl LightningClientModule {
         Ok(operation_id)
     }
 
+    async fn get_next_operation_id(
```
dpc, line 383: Nit: could use a comment. I'm a bit confused what is it for, without looking at the context and/or documentation first.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -380,6 +380,39 @@ impl LightningClientModule {
...
+        &self,
+        invoice: &Bolt11Invoice,
+    ) -> Result<OperationId, SendPaymentError> {
+        for payment_attempt in 0..u64::MAX {
```
dpc, line 387: Very nit: `0..` is shorter and suggest to me that "it will never reach the end" better.

dpc (approved):

PR #4889, opened 2024-04-09
fix: update ServerModule documentation on how to propose and process …

joschisan:
…consensus items


elsirion (approved):
Great clarifications on `consensus_proposal` and background for returning errors for valid but redundant CIs!

dpc (approved):

PR #4900, opened 2024-04-09
chore(jit-values): return string-copy of the original error

dpc:
Only the first caller to `Jit::get_try` can get the original error, but we can actually return a string-copy of it to other callers, which in most cases should be good enough.

The whole complication here allows retaining the original error (including source chain, backtrace, etc.) which might be useful when using in more concrete use-cases than just starting services in Devimint.

### DIFF
--- a/fedimint-core/src/task/jit.rs
+++ b/fedimint-core/src/task/jit.rs
@@ -17,7 +17,10 @@ pub type JitTryAnyhow<T> = JitCore<T, anyhow::Error>;
 /// Newtype over `Option<E>` that allows better user (error conversion mostly)
 /// experience
 #[derive(Debug)]
-pub struct OneTimeError<E>(Option<E>);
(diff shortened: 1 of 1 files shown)


maan2003 (approved):

maan2003:
what if return `Arc<anyhow::Error>`, we won't lose anything then

maan2003:
looks std has impl, so it must be a sensible impl
https://doc.rust-lang.org/stable/std/error/trait.Error.html#impl-Error-for-Arc%3CT%3E

dpc:
> what if return `Arc<anyhow::Error>`, we won't lose anything then

Is this even legal?

Try it, if it doesn't require some weird bounds, why not?

maan2003:
:facepalm: anyhow::Error doesn't impl From<Arc<anyhow::Error>>

elsirion (approved):

PR #4904, opened 2024-04-10
chore: use fm:: prefix for log targets

maan2003:
client may match some other client, we need to specify it is fedimint


elsirion (approved):
@TonyGiorgio @justinmoon if we merge this integrators should likely add these new targets as additional logging directives *now* so it doesn't get forgotten once 0.4 comes around.

maan2003:
we could just backport it

elsirion:
Same problem, nobody will remember when 0.3.1 is actually out and integrated because nobody cares about logging till it's needed :sweat_smile:

maan2003:
I think `client=debug` will match `fm::client`

elsirion:
I highly doubt that, afaik it operates on prefixes and not infixes.

maan2003:
yep it is prefix

maan2003:
also https://docs.rs/tracing-subscriber/latest/src/tracing_subscriber/filter/env/directive.rs.html#224

TonyGiorgio:
> @TonyGiorgio @justinmoon if we merge this integrators should likely add these new targets as additional logging directives *now* so it doesn't get forgotten once 0.4 comes around.

Really? Changing the prefix won't show the logs unless we do something different? I don't think we've had to specify any specific logging changes so far with fedimint, it has just worked.

elsirion:
> Changing the prefix won't show the logs unless we do something different?

Well, if you only use a global log level that gives you enough info you don't have to do anything. But if over time you discovered the components you are more interested in, then you might need to adapt your filter directives.

For example I'd recommend logging `client::reactor` with `trace` or at least `debug` level in case there's ever a weird state machine error one has to debug. For other components that's likely over the top.

TonyGiorgio:
Gotcha. Yeah since we don't do anything specific now, happy with these changes. Thanks for the heads up.

TonyGiorgio (approved):

dpc (approved):

PR #4911, opened 2024-04-10
chore: add a timeout to polling for consensus items

dpc:
Make it relatively large, just to detect and unstuck modules that really got stuck.

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -788,7 +788,7 @@ impl ConsensusServer {
             match result {
                 Ok(signed_session_outcome) => return signed_session_outcome,
                 Err(error) => {
-                    tracing::error!("Error while requesting signed session outcome: {}", error)
+                    tracing::error!(target: LOG_CONSENSUS, "Error while requesting signed session outcome: {}", error)
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


joschisan (approved):
Looks good, would prefer to refactor it to use a match statement though instead the use of continue as the latter makes it less readable imo and we can avoid to let assignments as well.

douglaz (approved):
LGTM

fedimint-server/src/consensus/server.rs
```diff
@@ -822,13 +824,25 @@ async fn submit_module_consensus_items(
...
-                        .await;
+                for (module_id, kind, module) in modules.iter_modules() {
+                    let module_consensus_items = tokio::time::timeout(
+                        CONSENSUS_PROPOSAL_TIMEOUT,
+                        module.consensus_proposal(
+                            &mut dbtx.to_ref_with_prefix_module_id(module_id).into_nc(),
+                            module_id,
+                        ),
+                    )
+                    .await;
```
elsirion, line 835: There are some theoretical concerns around being able to make guardians not propose by third parties somehow making `consensus_proposal` take longer than the timeout. But the practical concerns far outweigh this.

dpc: This can always be solved by doing slow operations on background thread or something like that, exactly like in #4916 .

And since modules are polled for consensus proposal serially RN, a slow modules is making consensus proposal slower for all other modules. @elsirion 

We should probably spawn a separate loop for every module instance. @joschisan what do you think?

PR #4912, opened 2024-04-10
fix: LSP getting spooked

dpc:
I don't digged why, but without these extra annotations sometimes LSP thinks it's a recursive function and displays fake compilation errors.


maan2003 (approved):
yep I have seen this before

joschisan (approved):

PR #4916, opened 2024-04-10
chore(wallet-module): Move block count and fee rate update to bg task

dpc:
Doing rpc from consensus method is meh, and in extreme cases can lead to hanging whole consensus.


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -385,19 +390,10 @@ impl ServerModule for Wallet {
...
-
-                dbtx.get_value(&FeeRateVoteKey(self.our_peer_id))
-                    .await
-                    .unwrap_or(self.cfg.consensus.default_fee)
-            }
-        };
+        let fee_rate_proposal = self
+            .get_fee_rate_opt()
+            .await
+            .unwrap_or(self.cfg.consensus.default_fee);
```
joschisan, line 396: I think defaulting to self.cfg.consensus.default_fee makes sense here.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -385,19 +390,10 @@ impl ServerModule for Wallet {
         // If there's an error getting the fee rate from the node we default to the most
         // recent fee rate vote. Using an alternative fee rate may cause unwanted
         // jitter.
```
joschisan, line 392: This comment does not make sense anymore.

joschisan (dismissed):
Please remove outdated comment, looks good otherwise.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1206,6 +1184,84 @@ impl Wallet {
...
+            }
+        });
+    }
+
+    fn spawn_bitcoin_update_task(
+        task_group: &mut TaskGroup,
+        bitcoind: &DynBitcoindRpc,
+    ) -> (Arc<AtomicU32>, Arc<RwLock<Option<Feerate>>>) {
+        let block_count_bg = Arc::new(AtomicU32::new(0));
+        let fee_rate_bg = Arc::new(RwLock::new(None));
```
elsirion, line 1207: Why not initialize with default from config instead of making it an `Option` if we `unwrap_or` anyway later?

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -760,6 +753,11 @@ pub struct Wallet {
     /// The result of last successful get_block_count
     block_count_local: std::sync::Mutex<Option<u32>>,
     our_peer_id: PeerId,
+    /// Block count updated periodically by a background task
+    /// Magic value `0` is used to denote it was not initialized yet.
+    block_count_bg: Arc<AtomicU32>,
+    /// Fee rate updated periodically by a background task
+    fee_rate_bg: Arc<RwLock<Option<Feerate>>>,
```
elsirion, line 760: You know my unreasonable fondness for `watch` channels @dpc :smile:, this could be:

```rust
fee_rate_bg: tokio::sync::watch::Receiver<Feerate>
```

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -927,25 +910,20 @@ impl Wallet {
...
+        let block_count = self
+            .block_count_bg
+            .load(std::sync::atomic::Ordering::Relaxed);
 
-        match res {
-            Ok(count) => *self.block_count_local.lock().expect("Failed to lock") = Some(count),
-            Err(ref err) => error!("Error while calling get_block_count: {:?}", err),
+        if block_count == 0 {
+            bail!("Block count not available yet");
         }
```
elsirion, line 920: nit: an explicit `Option` would be slightly cleaner

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1206,6 +1184,84 @@ impl Wallet {
...
+                            warn!(target: LOG_MODULE_WALLET, %err, "Unable to get fee rate from the node");
+                        }
+                    }
+                };
+
+                let desired_frequency = Duration::from_secs(1);
+                loop {
+                    let start = now();
+                    update_block_count().await;
+                    update_fee_rate().await;
```
elsirion, line 1253: Is this not in a `join!` on purpose to not stress `bitcoind` too much?

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1206,6 +1184,84 @@ impl Wallet {
...
+                        Ok(None) => {
+                            debug!(target: LOG_MODULE_WALLET, "Bitcoin node did not return a fee rate");
+                        }
+                        Err(err) => {
+                            warn!(target: LOG_MODULE_WALLET, %err, "Unable to get fee rate from the node");
+                        }
+                    }
+                };
+
+                let desired_frequency = Duration::from_secs(1);
```
elsirion, line 1249: Feels excessive, thinking of setups where multiple guardians poll the same `bitcoind`.

elsirion (approved):
Feel free to merge if none of my concerns seem big enough.

modules/fedimint-wallet-server/src/lib.rs
dpc: That's the current rate of consensus proposal loop. Also 1/s is much little, so slow, very nothing.

Increasing it makes devfed starts slower, as Federation notices block height updates slower, so I was thinking about ... decreasing it... :smiling_imp:

modules/fedimint-wallet-server/src/lib.rs
dpc: Nah, I just didn't think is necessary.

modules/fedimint-wallet-server/src/lib.rs
dpc: Actually... 

Maybe we could make it poll less often, but it needs some workaround for tests / dev env.

elsirion (dismissed):

joschisan (dismissed):

PR #4921, opened 2024-04-10
chore: add a WARNING for `process_consensus_items`

dpc:
In modules that are expected to commonly be used as a template for 3rd party modules.

I attempted to do it using return type changes, but indeeded without https://doc.rust-lang.org/core/ops/trait.Try.html stabilized, it leads to DX degradation and I have no better ideas.


joschisan:
I would specifically mention that this is the case for redundant items like redundant votes on a certain value.

joschisan (approved):

dpc:
I added pointers to just read the docs. Normally people don't read the docs, but if you warn them, some might.

joschisan (approved):

elsirion (approved):

Issue #4913, opened 2024-04-10
Client should consolidate e-cash


elsirion:
Currently, the client only makes sure that it has a minimum number of e-cash for each denomination. We should add logic that also consolidates excessive notes of the same denomination. This could easily be done for both receives and sends in `finalize_and_submit_transaction` by just adding excessive denomination notes as inputs to any transaction that we make.

douglaz:
This should be optional because some clients may need specific notes and don't want consolidation (e.g load-test-tool, printed ecash clients)

douglaz:
I think this issue is misguided, I'll copy my [comment](https://github.com/fedimint/fedimint/pull/4922#issuecomment-2051993692) here:

> I don't think "excess notes" was or will ever be an issue. The issue we had and will continue to have is a matter of a lack of notes (in that specific incident, a lack of mid-sized ones). So we should be talking about shortage of notes and filling these gaps. While filling these gaps we may or may not consume "excess notes". Probably we want spend high-tier notes to fill mid-tier ones.

Also see the [followup](https://github.com/fedimint/fedimint/pull/4922#issuecomment-2052309157) 

So consolidating ecash may be a partial solution to note gaps/shortage, but it will not be enough or always the best solution.

elsirion:
Fixed in #4922. We could be more sophisticated, but good enough for now.

PR #4922, opened 2024-04-10
feat(client): consolidate excessive amount of notes in a given tier

dpc:
In Fedi's LN gateway deployement, we've hit a condition where LN gateway accumulated tons of tiny notes and at some point is trying to spend all of them at once, creating tx larger than consensus limit.

Fundamentally, when an client like LN gateway is handling a lot of tiny payements, it will keep dealing with and accumulating tiny note denominations. The change outputs note selection algorithm tries to create outputs so that user has at least certain amount of notes available at every tier, but does nothing to get rid of excessive ones.

At some point it has to consolidate them.

This PR hooks up into the tx finalization, to potentially add some excessive notes to each transaction.

There are multiple ways to approach this problem, some might be more elegant or smarter, but this solution is simple and should deal with the issue without changing normal use-cases.

### Testing

Some automated test would be better, but RN I tested by manually lowering the tresholds and in devimint env running:

```
export RUST_LOG=fm::client::module::mint=debug,info
fedimint-cli info
fedimint-cli ln-invoice --amount 4msat
etc.
```

when the treshold of given denomination is reached:

```
2024-04-11T01:37:21.106950Z DEBUG fm::client::module::mint: Will consolidate excessive notes note_num=16 denominations_msats=[2, 2, 128, 128, 512, 512, 1024, 10
24, 4096, 4096, 32768, 32768, 524288, 524288, 1048576, 1048576]
```

Testing this is tricky, because it requires creating unbalanced notes, while the change selector keeps recreating notes to fill up the tiers.

Re #4913


dpc:
After thinking about it more, it seems to me that making note selector spend excessive notes as a priority would be much better approach, but that seems like much more involved change, and I created #4923 on how I'd approach it.

dpc:
I just can't help but keep refactoring...

fedimint-client/src/module/mod.rs
```diff
@@ -581,6 +581,16 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
...
+    /// Gives module a chance to consolidate some inputs, e.g. when accumulating
+    /// a lot of notes of the same denomination.
+    async fn consolidate_inputs(
+        &self,
+        _dbtx: &mut DatabaseTransaction<'_>,
+        _operation_id: OperationId,
+    ) -> anyhow::Result<Vec<ClientInput<<Self::Common as ModuleCommon>::Input, Self::States>>> {
+        Ok(vec![])
+    }
+
```
elsirion, line 593: That's a  very e-cash specific API extension. Couldn't this just be part of `create_sufficient_input`? That one doesn't guarantee any particular amount anyway, it could be arbitrarily larger.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -843,6 +851,70 @@ impl MintClientModule {
...
+            .await
+            .map(|(key, note)| (key.amount, note))
+            .fold(
+                BTreeMap::<Amount, Vec<SpendableNote>>::new(),
+                |mut acc, (amount, note)| async move {
+                    acc.entry(amount).or_default().push(note);
+                    acc
+                },
+            )
+            .await;
```
elsirion: This loads all the notes to memory, we tried to avoid that for selection, I don't remember if that was premature optimization or if we ran into a particular problem.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -843,6 +851,70 @@ impl MintClientModule {
...
+
+        /// At how many notes of the same denomination should we try to
+        /// consolidate
+        const MAX_NOTES_PER_TIER: usize = 10;
+        /// Number of notes per tier to leave after threshold was crossed
+        const MIN_NOTES_PER_TIER: usize = 4;
```
elsirion, line 889: The reasoning for this hysteresis would be interesting.

Imo it makes sense to use a power (of the base used for tier generations) difference between these values, so all the notes get consolidated into a higher tier. But the delta is 6 here and we should also not hard-code the assumption that we deal with base 2 denominations imo.

fedimint-client/src/lib.rs
```diff
@@ -998,6 +998,12 @@ impl Client {
         operation_id: OperationId,
         mut partial_transaction: TransactionBuilder,
     ) -> anyhow::Result<(Transaction, Vec<DynState>, Range<u64>)> {
+        partial_transaction.inputs.extend(
+            self.primary_module()
+                .consolidate_inputs(self.primary_module_instance, dbtx, operation_id)
+                .await?,
+        );
+
```
elsirion, line 1006: Good call to add these before `create_sufficient_input`, but imo we could do the same inside `create_sufficient_input`: first select consolidation inputs and then select the remaining amount (if any) using the normal algorithm.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -758,6 +758,28 @@ impl MintClientModule {
             .await
     }
 
+    pub async fn get_notes_by_tier_counts(
```
elsirion, line 761: Docs would be nice, seems to return *up to* the number of notes for each denomination specified in `counts`?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -861,7 +883,7 @@ impl MintClientModule {
         operation_id: OperationId,
     ) -> anyhow::Result<Vec<ClientInput<MintInput, MintClientStateMachines>>> {
         let notes_by_denomination: BTreeMap<Amount, Vec<SpendableNote>> = dbtx
-            .find_by_prefix_sorted_descending(&NoteKeyPrefix)
+            .find_by_prefix(&NoteKeyPrefix)
```
elsirion: Does the order matter at all here?

douglaz:
I'm inclined to CONCEPT NACK this. I don't think "excess notes" was or will ever be an issue. The issue we had and will continue to have is a matter of a _lack_ of notes (in that specific incident, a lack of mid-sized notes). So we should be talking about _shortage_ of notes and filling these gaps. While filling these gaps we may or may not consume "excess notes". Probably we want spend high-tier notes to fill mid-tier ones.

modules/fedimint-mint-client/src/lib.rs
dpc: Nope, but the natural order might be just easier for the db to do.

dpc:
> I'm inclined to NACK this. I don't think "excess notes" was or will ever be an issue. The issue we had and will continue to have is a matter of a _lack_ of notes (in that specific incident, a lack of mid-sized ones). So we should be talking about _shortage_ of notes and filling these gaps. While filling these gaps we may or may not consume "excess notes". Probably we want spend high-tier notes to fill mid-tier ones.

"Lack of high-tier notes" is the same thing as "having too many low-tier notes". If we have 100s of low-tier notes, and consolidate them, we'll get the extra higher-tier notes we were lacking.

> Probably we want spend high-tier notes to fill mid-tier ones.

If we lack a mid-tier note we can always spend a high-tier one and get a change. Also low-tier notes first consolidate into mid-tier, mid-tier into high-tier.

The goal is to generally have a flatter distribution altogether, so it's always possible to spend everything at once.

Though again - this is a quick provisional solution.

modules/fedimint-mint-client/src/lib.rs
dpc: I was hopping to minimize churn here. We definitely don't want to just take one note from a tier that happened to go over the limit, because it might just get remitted as a same tier.

Or if tier X gets consolidated to X+1, then in next TX tier X+1 to X+2, etc. that's a lot of churn, while it would be better to go straight to X+4 or higher.

Though I now think I should first see if any tier crossed HIGH, and then take all notes above LOW from *any* tier, up to LIMIT of total notes. Basically two passes: check for overflow, start grabbing notes.

I didn't do it the first time around, because I didn't have convenient APIs, which I added later down the line.
@elsirion

fedimint-client/src/module/mod.rs
dpc: Initially I was a bit afraid where should be added and if it will not break anything, so I made it a separate function. But now I think it will be OK as a part of `create_sufficient_input`.

dpc: @elsirion OK, so I attempted it and one imediately difference is that now consolidations would only happen if the tx being built in under-funded, because only then we call `create_sufficient_input`.

I think it's OK, but worth noting if case I'm wrong.

douglaz:
> > I'm inclined to NACK this. I don't think "excess notes" was or will ever be an issue. The issue we had and will continue to have is a matter of a _lack_ of notes (in that specific incident, a lack of mid-sized ones). So we should be talking about _shortage_ of notes and filling these gaps. While filling these gaps we may or may not consume "excess notes". Probably we want spend high-tier notes to fill mid-tier ones.
> 
> "Lack of high-tier notes" is the same thing as "having too many low-tier notes". If we have 100s of low-tier notes, and consolidate them, we'll get the extra higher-tier notes we were lacking.

lower tier notes won't give too much yield because of their small value. the way of solving lack of high tier notes spending even higher notes.

> 
> > Probably we want spend high-tier notes to fill mid-tier ones.
> 
> If we lack a mid-tier note we can always spend a high-tier one and get a change. Also low-tier notes first consolidate into mid-tier, mid-tier into high-tier.

Lower tier notes are more "useful" (i.e they can always be used). Proactively spending them may be harmful. I think any consolidation algorithm should focus on spending higher tiers first.


> 
> The goal is to generally have a flatter distribution altogether, so it's always possible to spend everything at once.
A pyramid shape  is perfectly ok supposing we don't have missing tiers on the middle.

> 
> Though again - this is a quick provisional solution.
> 
> @douglaz 👍 ❓

Ok, this may help in some cases, so it's a valid quick-solution

My guess is that it would not be enough to avoid that particular incident and will create a slightly worse note distribution in some extraordinary circumstances.

Anyway it's probably better than we have today.

douglaz (approved):
It's an improvement, but not a solution to the type of incident that generate this issue.

dpc:
> Lower tier notes won't give too much yield because of their small value. the way of solving the lack of high tier notes is by spending even higher notes.

The change consolidates all notes all the same. Notes will continuously consolidate into higher ones, higher into even higher.

> Lower tier notes are more "useful" (i.e they can always be used if within the transaction limit). Proactively spending them may be harmful. I think any consolidation algorithm should focus on spending higher tiers first.

I don't understand. As implemented RN, consolidation only ever spends notes that you already have plenty of.


Maybe we can discuss the whole thing in a Deep Dive. Added to https://github.com/fedimint/fedimint/wiki/Deep-Dive-Topics

dpc:
Commits with overlapping changes were pilling up, so just decided to squash it all.

Notably I changed the algo. so it first detects if any tier is above a MAX_TRIGGER and then consolidates all notes above MIN per tier. This should lead to less churn.

I need to test it again manually, and maybe think about a good test of some kinid, so will not merge just yet.

dpc:
OK, seems to work.

I'd really love a test for that. Need to find some good way to test though.

elsirion (commented):
LGTM, but as you said: let's test it :)

douglaz:
Provide some initial funds to a client, then create transactions generating uneven notes (see how fedimint-load-test-tool does that) then check if a normal spending/reissue will rebalance the notes.

maan2003:
I would love a explicit "rebalance" function that I can run whenever client is online, so in future selection is better when offline

justinmoon:
dev call: needs testing

dpc:
Test should be ready.

devimint/src/util.rs
```diff
@@ -209,6 +209,13 @@ impl Command {
         self
     }
 
+    pub fn args<T: ToString>(mut self, args: impl IntoIterator<Item = T>) -> Self {
```
dpc, line 212: Kept it from a different approach. It's useful if you want to pass an array of args.

fedimint-cli/src/client.rs
```diff
@@ -74,7 +74,11 @@ pub enum ClientCmd {
     /// Display wallet info (holdings, tiers)
     Info,
     /// Reissue notes received from a third party to avoid double spends
-    Reissue { oob_notes: OOBNotes },
+    Reissue {
+        oob_notes: OOBNotes,
+        #[arg(long = "no-wait", action = clap::ArgAction::SetFalse)]
```
bradleystachurski, line 79: We'll need to conditionally handle this for back-compat

```
00:00:58 Error: command: /nix/store/fdbrlr3n4n94gs5a1lgnzskhgmv8pimq-fedimint-cli/bin/fedimint-cli --data-dir=/tmp/nix-shell.cTg2BC/mint_client_sanity-AneI/devimint-1686873-266/clients/receiver-0 reissue --no-wait AgEELaftFQBTAQEBsFhhsRLPTo/0Ad9oJmWF8RN3Qgx5x9k9JcI8wh6Md8H7gryAC6XrHjlx80dOG9BlWzWp9lJaw/qI+VGtcofUIP5jHzsmgEAX7OCNTy6cY/8=
00:00:58 
00:00:58 Caused by:
00:00:58     exit status: 2
00:00:58     stdout:
00:00:58     
00:00:58     stderr:
00:00:58     2024-05-01T09:03:57.056540Z DEBUG fedimint_cli: Starting fedimint-cli (version: a8422b84102ab5fc768307215d5b20d807143f27)
00:00:58     error: unexpected argument '--no-wait' found
```

https://github.com/fedimint/fedimint/actions/runs/8907459104/job/24461360848?pr=4922#step:5:1327

dpc: Right. Will take care of it.

modules/fedimint-mint-tests/src/bin/mint-client-sanity.rs
```diff
@@ -0,0 +1,80 @@
...
+        .args(&all_notes[all_notes.len() - 1..])
+        .run()
+        .await?;
+
+    let info = cmd!(receiver, "info").out_json().await?;
+    info!(%info, "receiver info");
+    // receiver has the balance
+    assert_eq!(info["total_amount_msat"].as_i64().unwrap(), 20);
+    // without the consolidation, this would be 20 1msat notes
+    assert!(info["denominations_msat"]["1"].as_i64().unwrap() < 20);
```
elsirion, line 77: That's not the case afaik, without it it would likely be 17-18 1sat notes because we'd first fill up lower tiers. Did you test that the test fails without consolidation?

dpc: @elsirion Lower than 1msat?

I did, but let me verify one more time, as I was tweaking the exact method to speed it up, after I verified, I think.

dpc:
```
thread 'main' panicked at modules/fedimint-mint-tests/src/bin/mint-client-sanity.rs:77:5:
assertion failed: info["denominations_msat"]["1"].as_i64().unwrap() < 20
```

```
> g diff HEAD
diff --git a/modules/fedimint-mint-client/src/lib.rs b/modules/fedimint-mint-client/src/lib.rs
index 5ecd9a33d5..6320040534 100644
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -939,6 +939,7 @@ impl MintClientModule {
         operation_id: OperationId,
     ) -> anyhow::Result<(Vec<ClientInput<MintInput, MintClientStateMachines>>, Amount)> {
+        return Ok((vec![], Amount::ZERO));
         /// At how many notes of the same denomination should we try to
         /// consolidate
         const MAX_NOTES_PER_TIER_TRIGGER: usize = 8;
```

@elsirion

elsirion: oh, sorry, I read sat :see_no_evil:

elsirion (approved):

dpc:
```
 00:00:41 assertion failed: epochs_descriptors.contains(*utxo_descriptor)
```

Known problem, re-added to MQ.

fedimint-backports:

```bash
git worktree add -d .worktree/backport-4922-to-releases/v0.3 origin/releases/v0.3
cd .worktree/backport-4922-to-releases/v0.3
git switch --create backport-4922-to-releases/v0.3
git cherry-pick -x 966b8c4e79e871705e7f1da229d5b78cfa48088a c13c65cf22e7a91df9d70155497d0759837704a1 1b76381186ba595a1f3b4416948a78093ad7166e
```

Issue #4841, opened 2024-04-05
Fix docs.rs build


elsirion:
https://docs.rs/crate/fedimint-core/0.3.0/builds/1175722

elsirion:
Broken:
* https://docs.rs/crate/fedimint-gw-client/latest
* https://docs.rs/crate/fedimint-gwv2-client/latest
* https://docs.rs/crate/fedimint-testing/latest
* https://docs.rs/crate/fedimint-recoverytool/latest
* https://docs.rs/crate/fedimint-server-tests/latest

PR #4908, opened 2024-04-10
chore(deps): bump reqwest to 0.12.5

tvolk131:
~~Bump everywhere except for `fedimint-cli` since it currently uses `lnurl-rs` v0.4.1 which uses `reqwest` v0.11 and causes an error here if we bump it, due to mismatching versions:~~ https://github.com/fedimint/fedimint/blob/a3b9990ebcc5690f67b6a8e8895d8a9024bab77c/fedimint-cli/src/client.rs#L679

~~Once a new version of `lnurl-rs` is published that uses `reqwest` v0.12, we should be able to bump them both together for `fedimint-cli` without issue~~

Edit: a new version of `lnurl-rs` has been published and we're now bumping `reqwest` to v0.12.3 across the entire codebase along with bumping `lnurl-rs` from v0.4.1 to v0.5.0

PR #4927, opened 2024-04-11
fix: docs.rs compilation error

dpc:
Re #4841

AFAIK this line is not neccessary, and `fedimint-core` always enables `sync` on `tokio`, which this conditional seems to be related to.

### DIFF
--- a/fedimint-core/src/util/broadcaststream.rs
+++ b/fedimint-core/src/util/broadcaststream.rs
@@ -15,7 +15,6 @@ use crate::util::BoxFuture;
 ///
 /// [`tokio::sync::broadcast::Receiver`]: struct@tokio::sync::broadcast::Receiver
 /// [`Stream`]: trait@futures::Stream
-#[cfg_attr(docsrs, doc(cfg(feature = "sync")))]
 pub struct BroadcastStream<T> {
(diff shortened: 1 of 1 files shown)


tvolk131:
Do we know where the discrepancy is? Is it worth filing a bug to add a CI check to prevent regressions?

dpc:
> Do we know where the discrepancy is? Is it worth filing a bug to add a CI check to prevent regressions?

Not really. It's just conditional compilation was hidding a directive that works only on nightly rust, etc. unlikely it will happen again, so I wouldn't overreact.

justinmoon (approved):

elsirion (approved):

dpc:
Flake. Back to the merge queue.

tvolk131:
> Flake.

Looks like a timeout - would it be worth bumping the threshold again like #3269 did? (But also for CI)

tvolk131:
FWIW I ran into a timeout while submitting #4908 earlier today as well: https://github.com/fedimint/fedimint/pull/4908#event-12453913496

dpc:
> Looks like a timeout - would it be worth bumping the threshold again like #3269 did? (But also for CI)

Looks like test going wrong somewhere and getting stuck, I don't think bumping any timeouts will help. We need to diagnose and fix.

PR #4930, opened 2024-04-11
feat: allow guardians to manually maintain gateway set for LN V2

joschisan:
This pr proposes a very different approach to gateway registration than lightning legacy. The design is in no way final and is intended to initiate an open discussion. 

Lightning V2 has been designed such that the only information that a client requires to route payments via a gateway is the gateways SafeUrl. When the client wants to route via a gateway it requests all necessary information to do so from the gateway right before, which doubles as availability check which we would want to do anyways. This, for example, allows a lightning address server to set up an incoming contract for a federation that the address server does not have a client config for, only the federation id is required.

Since the only information maintained by the guardians for any given gateway - the url - is now static, guardians can maintain the set of registered gateways by hand.  A gateways url is simply added or removed by the guardians via the guardians UI.

One argument for this approach, besides its simplicity and absence of DOS vectors, is that maintaining a set of gateways for the optimal user experience is better solved as a social problem than a technical one. Factors like uptime, fees and routing reliability need to be weighed differently for every federation. Furthermore, though all three factors could be measured automatically, at least the latter may be distorted by malicious actors. In that case, the relationship between guardians and gateway operator becomes the deciding factor in recommending the gateway to users or not.


fedimint-core/src/query.rs
```diff
@@ -261,28 +261,37 @@ impl<R> UnionResponses<R> {
...
+                if self.responses.len() == self.threshold {
+                    let mut union = Vec::new();
+
+                    for response in self.responses.values() {
+                        for r in response {
+                            if !union.contains(r) {
+                                union.push(r.clone());
+                            }
+                        }
                     }
```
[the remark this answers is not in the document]
joschisan: Done.

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -544,6 +557,37 @@ impl ServerModule for Lightning {
                     Ok(module.outgoing_contract_expiration(db, contract_id).await)
                 }
             },
+            api_endpoint! {
+                ADD_GATEWAY_ENDPOINT,
```
dpc, line 561: Should we generally name more stuff here with "vetted" word in it? endpoints, some constants, db fields. I guess that's the meaning and it greatly clarifies the context here.

modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -544,6 +557,37 @@ impl ServerModule for Lightning {
...
+                    check_auth(context)?;
+
+                    let db = context.db();
+
+                    Ok(Lightning::add_gateway(db, gateway).await)
+                }
+            },
+            api_endpoint! {
+                REMOVE_GATEWAY_ENDPOINT_V2,
+                ApiVersion::new(0, 0),
```
dpc, line 573: Should we increment minor api version of this module so clients can detect new calls, and then `ApiVersion` here needs to be set to the new version.

modules/fedimint-lnv2-client/src/api.rs
```diff
@@ -24,6 +26,8 @@ pub trait LnFederationApi {
         &self,
         contract_id: &ContractId,
     ) -> FederationResult<Option<u64>>;
+
+    async fn fetch_gateways(&self) -> FederationResult<Vec<SafeUrl>>;
```
dpc, line 30: I don't see client calls to add and remove. 

Consider following up with:

* these client calls
* command line handling so that `fedimint-cli module lnv2 vetted add <url>`, `fedimint-cli module lnv2 vetted list` can be already used.
* sanity test

All of these are done in meta module, so you might copy&paste&modify if it saves you any time.

modules/fedimint-lnv2-server/src/lib.rs
joschisan: I am open to it.

modules/fedimint-lnv2-server/src/lib.rs
joschisan: I don't follow, do you want to change something here?

modules/fedimint-lnv2-client/src/api.rs
joschisan: I just pushed the first two changes a a commit, however I am unsure how to factor it. Initially I extended the LnFederationApi with the add and remove calls, however so far they are currently only used in the cli which is only compiled conditionally, similar situation with the admin_auth method on the client module itself like in the meta module but here clippy actually errored. I am wondering, how will the UI call these methods and If there is a better way to factor this. I have to little context here to decide this.

Tests can be done In a follow up.

modules/fedimint-lnv2-client/src/cli.rs
```diff
@@ -0,0 +1,77 @@
...
+    Remove { gateway: SafeUrl },
+    /// List our own vetted gateways
+    List,
+}
+
+pub(crate) async fn handle_cli_command(
+    lightning: &LightningClientModule,
+    args: &[ffi::OsString],
+) -> anyhow::Result<serde_json::Value> {
+    let opts = Opts::parse_from(iter::once(&ffi::OsString::from("lnv2")).chain(args.iter()));
```
elsirion, line 29: Why is the `lnv2` prepended here?

fedimint-core/src/query.rs
```diff
@@ -261,28 +261,37 @@ impl<R> UnionResponses<R> {
...
-                }
 
-                assert!(self.responses.insert(peer));
+                    union.sort_by_cached_key(|r| {
+                        self.responses
+                            .values()
+                            .filter(|response| !response.contains(r))
+                            .count()
+                    });
```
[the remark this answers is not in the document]
elsirion: The docs disagree actually:

> The sorted iterator, if directly collected to a Vec, is converted without any extra copying or allocation cost.

It's a bit of a rabbit hole to understand how this works https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3CT%3E-for-Vec%3CT%3E

modules/fedimint-lnv2-server/src/lib.rs
elsirion: @dpc there are no stability guarantees around LNv2 for now, so I don't think that's necessary.

elsirion:
Also needs rebase.

modules/fedimint-lnv2-client/src/cli.rs
joschisan: I copied this part from the meta module as @dpc suggested.

elsirion: I have a suspicion: `clap` might skip the first arg because it's the binary name when parsing command line args directly. Here we are only parsing a subset of the command line args, so need to add a dummy first one.

elsirion (approved):

modules/fedimint-lnv2-server/src/lib.rs
dpc: Oh, right.

modules/fedimint-lnv2-client/src/cli.rs
dpc: @elsirion That's right.

We could consider passing from the higher level the name of the module itself to make it unnecessary, and maybe even allow modules to detect their own kind.

PR #4942, opened 2024-04-12
chore: add more context `api::FederationError`

elsirion:
When we log these errors I'd like to see why we made the failing API call. Alternatively we could also attach it as context to the anyhow error now that I think about it. But maybe having the data in a more structured form is also nice.


dpc (approved):

dpc:
Adding back to MQ, seems like something weird failed on GHA/cachix action level.

PR #4947, opened 2024-04-12
chore: conditionally handle current binaries in upgrade tests

bradleystachurski:
Followup to https://github.com/fedimint/fedimint/pull/4762#issuecomment-2047483947

> Do we even need to backport this? It seems to me as if we could just run any upgrade tests off master, specifying the versions we want to test.

Previously upgrade tests would fail to discover a common api version against master. Instead, we can skip testing against the binaries in the target dir and use only the versions passed to `just test-upgrades`.

### DIFF
--- a/scripts/tests/upgrade-test.sh
+++ b/scripts/tests/upgrade-test.sh
@@ -21,15 +21,17 @@ fedimintd_paths=()
 fedimint_cli_paths=()
 gatewayd_paths=()
 for version in "${versions[@]}"; do
-  fedimintd_paths+=("$(nix_build_binary_for_version 'fedimintd' "$version")")
-  fedimint_cli_paths+=("$(nix_build_binary_for_version 'fedimint-cli' "$version")")
-  gatewayd_paths+=("$(nix_build_binary_for_version 'gatewayd' "$version")")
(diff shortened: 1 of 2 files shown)


scripts/tests/upgrade-test.sh
```diff
@@ -26,11 +26,6 @@ for version in "${versions[@]}"; do
   gatewayd_paths+=("$(nix_build_binary_for_version 'gatewayd' "$version")")
 done
 
-# Add current binaries from PATH
-fedimintd_paths+=("fedimintd")
-fedimint_cli_paths+=("fedimint-cli")
-gatewayd_paths+=("gatewayd")
```
elsirion, line 32: I think we want the ability to test upgrades to current master (which I assume these lines are about). Maybe there could be a special-cased version `current` that turns these on/off as the last step?

bradleystachurski: Good call 👍 

I've created an issue for fixing the error with discovering common core api versions when upgrading to `master`.

https://github.com/fedimint/fedimint/issues/4960

dpc (approved):

elsirion (approved):

PR #4949, opened 2024-04-12
chore: further simplify meta module CI proposal generation

### LINKED ISSUE #4895: Simplify meta module CI proposal logic
> When looking at it with @joschisan we noticed that this could be further simplified, but getting a correct version of the code was more important to me than reducing the complexity to its minimum. We can do that later since it's only the submission logic.

_Originally posted by @elsirion in https://github.com/fedimint/fedimint/pull/4836#discussion_r1558180646_

elsirion:
Fixes #4895 before I forget what it was about.

### DIFF
--- a/modules/fedimint-meta-server/src/lib.rs
+++ b/modules/fedimint-meta-server/src/lib.rs
@@ -279,32 +279,18 @@ impl ServerModule for Meta {
             },
         ) in desired
         {
-            let consensus_value = &Self::get_consensus(dbtx, key).await;
             let consensus_submission_value =
                 Self::get_submission(dbtx, key, self.our_peer_id).await;
+
             if consensus_submission_value.as_ref()
-                == Some(&MetaSubmissionValue {
+                != Some(&MetaSubmissionValue {
                     value: desired_value.clone(),
                     salt,
                 })
             {
-                // our submission is already registered, nothing to do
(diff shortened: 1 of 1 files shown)


modules/fedimint-meta-server/src/lib.rs
```diff
@@ -279,32 +279,18 @@ impl ServerModule for Meta {
...
+
             if consensus_submission_value.as_ref()
-                == Some(&MetaSubmissionValue {
+                != Some(&MetaSubmissionValue {
                     value: desired_value.clone(),
                     salt,
                 })
             {
-                // our submission is already registered, nothing to do
-            } else if consensus_value.as_ref() == Some(&desired_value) {
```
dpc, line 292: This ignores the current consensus value. After submitted value becomes consensus, the submission becomes `None`. So here, it will no longer `==` making this keep returning the value as a submission consensus item. But the `process_consensus_proposal` will detect that the submission is equal to consensus and ignore it. But since meta module fails to return `Err`... it will keep spamming and making the consensus history grow larger.

Not sure how to even write a test for it, because it's invisible on the API level.

elsirion: Nah, we weren't aware of that cleanup logic. I thought it was an easy win, but it isn't, so let's leave it be for now since it seems to work.

PR #4950, opened 2024-04-12
fix(meta): redundant consensus item handling in new Federations

dpc:
This introduces a config-based errata, so that new Federations will use a new behavior, while old ones stay supported.

```
> cat /tmp/devimint-2070662-004/fedimintd-0/local.json | grep errata
      "errata_redundant_err": true
```

### DIFF
--- a/modules/fedimint-meta-common/src/config.rs
+++ b/modules/fedimint-meta-common/src/config.rs
@@ -43,7 +43,16 @@ pub struct MetaClientConfig;
 
 /// Locally unencrypted config unique to each member
 #[derive(Clone, Debug, Serialize, Deserialize, Decodable, Encodable)]
-pub struct MetaConfigLocal;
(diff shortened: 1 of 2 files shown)


dpc:
@elsirion @joschisan This is a PoC of fixing consensus behavior in a backward compatible way, that relies on all peers using the same version of Fedimint to do DKG. It seems relatively straightforward to employ, and we could possibly use it elsewhere in the future.


As per https://github.com/fedimint/fedimint/pull/4949#discussion_r1563260845 - I'm inclined not to change the clearing submission behavior, as IMO it's just one extra condition on both sides (propose + process) and it gets rid of needless values sticking around for no reason (storage + APIs returning them), but if you were both vehemently against keeping it, I guess it could be done in the same errata). I guess I'll just have to make whole `process_consensus_item` conditional to keep it sane.

PR #4951, opened 2024-04-13
feat: ability to start devimint from existing state

dpc:
I need a way to verify certain behaviors manually between versions, etc. Similiar to upgrade tests etc.

It occurred to me that it should be possible to just start devimint skipping all the DKGs and other setups, and it should be relatively easy. So I implemeted it.

One blocker that I've hit was inability to tell fedimintd to use certain URL to connect to bitcoind, instead of whatever was added to the config. I made new env variables that can be used to force bitcoind kind&url. The way existing env vars are working is confusing (they are only used as a default for DKG, so I renamed them while at it (with backward-compat.  of course).

Another blocker was that on killing devimint, we send `kill -9` right away, which makes `bitcoind` not get any chance to flush database to disk, and thus lose block history. All other daemons get really upset about it. So I changed it to `kill`, 3s wait, `kill -9` sequence. A bit slower (`fedimimtd` takes some time to shutdown), but works, and seem better. We can fix & improve with time.

The implementation is a bit meh, but can be improved with time. I would also like to modularize the setup so e.g. test that doesn't need LND, or electrs, etc. does not need to start them, and can configure it via builder.

## Testing

Start devimint e.g. like this:

```
> just devimint-env
...
2024-04-13T04:29:23.013121Z  INFO fm::devimint: Devimint data dir path=/tmp/devimint-3321804-158
```

Stop it, and then start it again, but skip the setup.

```
env FM_TEST_DIR=/tmp/devimint-3321804-158 FM_SKIP_SETUP=true RUST_LOG=fm::devimint=debug just devimint-env
```

It takes it sweet time as LND Gateway blocks on LND starting to return route hints, but works.

### DIFF
--- a/devimint/src/cli.rs
+++ b/devimint/src/cli.rs
@@ -4,7 +4,7 @@ use std::ops::ControlFlow;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, ensure, Context, Result};
 use clap::{Parser, Subcommand};
 use fedimint_core::task::TaskGroup;
 use fedimint_core::util::write_overwrite_async;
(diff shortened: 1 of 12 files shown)


elsirion:
Lots of consensus item processing going on in the failed CI run 🤔

dpc:
Need to debug what did I break exactly.

dpc:
```
2024-04-15T22:22:19.728654Z  INFO address{payload=DepositAddressPayload { federation_id: FederationId(2aa93c5cc9fd7efe19d223482f78491078937c9659cb8f258776423ccf1cd2e7) }}: fm::net::blockchain: Bit
coind error JSON-RPC error: RPC error response: RpcError { code: -4, message: "Only legacy wallets are supported by this command", data: None }, retrying
```

Seems like something about this change made bitcoind return `Only legacy wallets are supported by this command` on gw trying to peg-in and create deposit address.

dpc:
Switched to `jj`, pushed this (with fixes :crossed_fingers:) as  #4971

Issue #4923, opened 2024-04-11
Refactor mint note selection and make note selector spend excessive notes.


dpc:
I don't know if this is going to be worth doing, since I've seen some denomination-less ecash design work, but while working on #4922 (which is more of a bandaid), it occurred to me that `NotesSelector::select_notes` has a weird interface.

It currently works taking "stream of notes" `stream: impl futures::Stream<Item = (Amount, Note)> + Send` as an input, but that's completely unnecessary and actually detrimental to trying to implement better note selection.

First - Ecash notes are fungible, so we don't need a list of notes. We just need to know how many notes in each given tier we have (aka `pub struct TieredSummary(Tiered<usize>)`), and produce how many notes of each tier we need to use (again, `TieredSummary`). This would make note selection a simple (from API perspective) function, easy to test and decoupled from database and even notes details.

Dealing with a stream might seem like an optimization, but in practice - how many notes does a client supposed to have. < 500, all sitting next to each other in sibling BTreeNodes of the database. Reading them all in is not an issue, and even if it was, it would be better to just maintain a in-memory `TieredSummary` copy and avoid reading from the db altogether.

After this is implemented, it should be easier to implement note selection that would try to spend notes from tiers with excessive amounts first, which again - requires knowing all notes upfront anyway.

dpc:
From @elsirion 's comment

> I think to remember that the slow thing about querying notes was decoding BLS curve points (and we'd only do that for the notes we actually want ideally).

We are doing some heavy CPU processing while reading stuff from the DB? That's no great.

What if instead of complicating usage to avoid it, we just have a parallel type that decodes note details as raw bytes. We can use the real type only when we intend to need the details.

elsirion:
Needs benchmarking. When I wrote my thesis about Fedimint ages ago I had to keep the wallet in memory for latency benchmarks. Later this was supposedly fixed by streaming notes, but I don't remember the details anymore. A lot of 1ms operations also add up.

Another clue might be that reading and decoding an entire federation DB of a few MB for analysis took a few minutes when I last did it with Fedimint 0.1, maybe we improved that though.

dpc:
TODO:

* [ ] write deserialization of notes bench
* [ ] if indeed to slow (1ms * 100 or even 300 is too slow), make a type that skips decoding.

@elsirion Are you OK on the idea to just read counts of notes using a different type, and then just pick only the nodes we need from the DB?

dpc:
See #4952 for the benching.

[...]

PR #4952, opened 2024-04-13
 chore: `SpendableNoteUndecoded` + bench SpendableNote decoding  (PoC)

dpc:
As a part of #4923 I wanted to see if decoding `SpendableNote` is indeed slow, and how much could we save if we skipped decoding of the signature using an alternative parallel type.

```
     Running benches/divan.rs (target-nix/release/deps/divan-2d8a606864c20c66)
Timer precision: 20 ns
divan                               fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ spendable_note_decode            93.5 µs       │ 188 µs        │ 94.32 µs      │ 96.49 µs      │ 100     │ 100
╰─ spendable_note_undecoded_decode  14.79 µs      │ 24.49 µs      │ 14.88 µs      │ 15.11 µs      │ 100     │ 100
```

Seems like decoding seem to be orders of magnitude less than what was suspected, in the range of 100us, which even with hundreds of notes should not be a huge problem.

Using an alternative type is indeed significantly faster. It wouldn't be a lot of effort to use it tactically e.g. when reading all notes for the purposes of using `TieredCounts` instead of "streaming notes" in #4923.

But given how relatively small numbers we see here, and how infrequent note selection is in practice, I'm not sure if it's worthwile.

You can try it yourself by running `cargo bench`, let me know what you think.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -799,6 +799,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim 0.11.0",
+ "terminal_size",
 ]
 
 [[package]]
@@ -860,6 +861,12 @@ checksum = "d16048cd947b08fa32c24458a22f5dc5e835264f689f4f5653210c69fd107363"
 dependencies = [
  "crossbeam-utils",
 ]
+
+[[package]]
+name = "condtype"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baf0a07a401f374238ab8e2f11a104d2851bf9ce711ec69804834de8af45c7af"
 
 [[package]]
 name = "console-api"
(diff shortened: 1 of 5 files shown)


dpc:
@elsirion @douglaz , as we have been discussing note consolidation PR #4922

elsirion (approved):
How do you like using divan? I played around with `criterion` in #4953 a bit, also works. I think we should get the benchmarks even if we end up not doing the delayed decoding.

dpc:
> How do you like using divan? I played around with `criterion` in #4953 a bit, also works. I think we should get the benchmarks even if we end up not doing the delayed decoding.

I usually go with criterion, but heard about about divan at some point and wanted to try it out. Criterion does bunch of extra stuff like html graphs, and storing previous runs under ./target/critterion which is useful... sometimes. Divan has a nicer derive macro and they say that they have better support for stuff like generics. By default it's also much faster, while criterion wants to run everything for many seconds (there are options to make it go faster). I don't have a strong preference either way.

dpc:
https://github.com/nvzqz/divan/issues/43 ehh, `divan` doesn't seem compatible with `cargo nextest` ...

dpc:
@elsirion Converted to criterrion and cherry-picked your decoding benches in  #4955 . Another commit, because I'm switching to `jj` and I'm still lame using it.

nvzqz:
@dpc @elsirion heads up that Divan now supports `cargo-nextest` as of v0.1.18, in case you're interested in giving it another try 😄

dpc:
@nvzqz Great. Thanks for letting us know. I'll have to try it out again sometime.

PR #4953, opened 2024-04-13
Fix tbs benchmarks and add one for loading e-cash from DB

elsirion:
Some insights for #4923: the time loading e-cash from DB takes is neither great nor terrible.

```
load 1000 e-cash notes from DB
                        time:   [104.47 ms 104.99 ms 105.58 ms]

signature decoding      time:   [83.402 µs 83.944 µs 84.520 µs]
```


elsirion:
LOL, should have read #4923 to the end and not just the one initial comment listing benchmarking as a TODO :see_no_evil: I guess at least this one also has the "read from DB" benchmark.

PR #4956, opened 2024-04-15
refactor: GlobalApi should make non-admin calls to all peers

dpc:
An-"admin-enabled" `GlobalApi` should behave for non-admin calls just like non-"admin-enabled" one. For this to happen self-peer-id needs to be tracked separately instead of just narrowing `all_peers` to one (self) peer_id.

In the future this should enable storing self-peer-id in the Client's db, and using such client for both admin and non-admin stuff all the same.

Re #4519

### DIFF
--- a/fedimint-api-client/src/api.rs
+++ b/fedimint-api-client/src/api.rs
@@ -156,6 +156,13 @@ impl FederationError {
         }
     }
 
+    pub fn new_one_peer(peer_id: PeerId, error: PeerError) -> Self {
+        Self {
+            general: None,
+            peers: [(peer_id, error)].into_iter().collect(),
+        }
+    }
+
     /// Report any errors
     pub fn report_if_important(&self) {
         if let Some(error) = self.general.as_ref() {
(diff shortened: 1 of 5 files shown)


fedimint-api-client/src/api.rs
```diff
@@ -287,6 +300,28 @@ pub trait FederationApiExt: IRawFederationApi {
         }
     }
 
+    /// Like [`Self::request_single_peer`], but API more like
+    /// [`Self::request_with_strategy`].
+    async fn request_single_peer_federation<FedRet>(
+        &self,
+        timeout: Option<Duration>,
```
elsirion, line 307: Why have an optional timeout here? If we want timeouts these should likely be added on externally. We could implement an extension trait that would let us do:

```rust
some_future
    .timeout(Duration::from_secs(1))
    .await
    .expect("timed out");
```

fedimint-api-client/src/api.rs
```diff
@@ -1278,9 +1344,10 @@ where
     }
 
     /// Creates a new API client
-    pub fn new_with_client(peers: Vec<(PeerId, SafeUrl)>) -> Self {
+    pub fn new_with_client(peers: Vec<(PeerId, SafeUrl)>, self_peer_id: Option<PeerId>) -> Self {
```
elsirion, line 1347: What does the `with_client` stand for here? That we already initialize the inner clients?

fedimint-api-client/src/api.rs
```diff
@@ -1257,6 +1316,13 @@ impl WsFederationApi<WsClient> {
...
+    pub fn with_self_peer_id(self, self_peer_id: PeerId) -> Self {
+        Self {
+            self_peer_id: Some(self_peer_id),
+            ..self
+        }
+    }
+
     /// Creates a new API client from a invite code, assumes they are in peer
     /// id order
     pub fn from_invite_code(info: &[InviteCode]) -> Self {
```
elsirion, line 1328: unrelated to this PR: this fn is such a foot-gun. Could be written much more elegantly now that single invite codes can contain multiple peer URLs.

elsirion (approved):
Could use a second pair of eyes @m1sterc001guy / @maan2003

elsirion:
Let's try this tagging approach :)

oleonardolima (commented):
ACK

PR #4958, opened 2024-04-15
fix: enable cancellation for checks with inherent timing assumptions

joschisan:
This pr makes the design choice that errors returned by send_payment and create_invoice remain strings instead of typed error since they are not intended to be shown to the user but only serve debugging purposes. 

The errors are either due to programming errors or only occurs in theoretically possible but unlikely timing scenarios like

* The user checks an invoice for expiration and its not yet expired but when the gateway checks it a second later it is.
* The user fetches the fees from the gateway and the consensus block count from the federation to create an outgoing contract but then one of he two is unreachable for the client for at least 144 blocks such that the contract is cancelled once the gateway and/or federation come back online and the remaining timelock delta is not sufficient.
* The user fetches the consensus block count from the federation for an outgoing lightning payment, but a second later the federation is offline when the gateway tries to confirm the contract. Then the payment gets stuck until the federation comes back online.

None of the additional information provided by the returned errors suggests any other behaviour by the user then just retrying the payment, probably with a different gateway if available, which is the same behaviour if the lightning payment failed or got stuck. Furthermore, these errors should be a negligible  percentage of errors experienced by the user as the most common error should be that the lightning payment fails or gets stuck; otherwise, what are we doing if the gateways are less reliable then lightning?

### DIFF
--- a/gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
+++ b/gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
@@ -11,6 +11,7 @@ use fedimint_lnv2_client::LightningClientStateMachines;
 use fedimint_lnv2_common::contracts::OutgoingContract;
 use fedimint_lnv2_common::{LightningInput, OutgoingWitness, Witness};
 use lightning_invoice::Bolt11Invoice;
+use serde::{Deserialize, Serialize};
 
 use crate::gateway_lnrpc::PayInvoiceRequest;
 use crate::gateway_module_v2::GatewayClientContextV2;
@@ -45,13 +46,22 @@ pub struct SendSMCommon {
 pub enum SendSMState {
     Sending,
     Claiming(Claiming),
-    Cancelled(String),
+    Cancelled(Cancelled),
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub struct Claiming {
     pub preimage: [u8; 32],
     pub outpoints: Vec<OutPoint>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable, Serialize, Deserialize)]
+pub enum Cancelled {
+    InvoiceExpired,
+    TimeoutTooClose,
+    Underfunded,
+    LightningRpcError(String),
+    ReceiveError(String),
 }
 
 #[cfg_attr(doc, aquamarine::aquamarine)]
@@ -82,6 +92,7 @@ impl State for SendStateMachine {
                         self.common.max_delay,
                         self.common.max_fee_msat,
                         self.common.invoice.clone(),
+                        self.common.contract.amount,
                     ),
                     move |dbtx, result, old_state| {
                         Box::pin(Self::transition_send_payment(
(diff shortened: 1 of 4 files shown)


gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
```diff
@@ -111,14 +122,39 @@ impl SendStateMachine {
...
+            .payment_fees_v2()
+            .send
+            .add_fee(invoice_msats);
+
+        if contract_amount < min_contract_amount {
+            return Err(Cancelled::Underfunded);
+        }
+
+        let excess_fee = contract_amount - min_contract_amount;
+        let max_fee_msat = excess_fee.msats + (min_contract_amount.msats - invoice_msats) / 2;
```
elsirion, line 151: Why divide by 2 here? Is there some inherent assumption that half of the fee budget will be guaranteed profit? If so that's a bad approach imo: just specify the min profit and then use the remaining fee paid for routing if necessary.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1590,61 +1591,58 @@ impl Gateway {
...
 
         if payload.invoice.consensus_hash::<sha256::Hash>() != payload.contract.invoice_hash {
-            return Err(SendPaymentErrorV2::InvalidInvoiceHash);
+            bail!("The invoices hash does not match the contract");
         }
 
+        let invoice_msats = payload
+            .invoice
+            .amount_milli_satoshis()
+            .ok_or(anyhow!("Invoice is missing amount"))?;
```
elsirion, line 1629: Here you make the assumption that the client will check that before and can display an error based on that preliminary check, which is valid imo.

elsirion (commented):
Given that you already used an error enum, what's the reason to rip it out? API flexibility for adding new error variants? I don't have a problem with it because the reasoning is sound, just normally we use `anyhow` due to laziness/time constraints because it's faster and not put extra effort into making things less structured :laughing: 

Thoughts @dpc @jus

gateway/ln-gateway/src/lib.rs
joschisan: Yes, the client checks that and has a structured Error variant for it.

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
joschisan: I am open to other fee strategies though.

elsirion: Just feels like a convoluted way of thinking about it, the gateway could just suggest a total fee and enforce a min fee that always goes to the gateway. Then you just subtract the min fee from the total fee and that's your routing budget.

joschisan:
> Given that you already used an error enum, what's the reason to rip it out?

Having a typed error is confusing when you always turn it into a string, especially when you turn it into a string on not just out of laziness but on purpose. Furthermore, its unnecessary indirection and about 50 locs.

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
joschisan: Well that is what we are doing here just that the total fee is twice the min fee and total fee is also enforced as a minimum. We then just let the client know the total fee instead of both / the exact split.

If we want to do a simplification here, I would suggest to get rid of overpayment that we enable now; instead we do it the same way as for receiving where there is one exact fee you have to pay and that's it. So both send and receive fee concepts function identical. Then the gateway is responsible for choosing a total fee that has the right tradeoff between reliability and fee for its users.

joschisan: https://github.com/fedimint/fedimint/pull/4975

elsirion (approved):

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
elsirion: I fundamentally disagree with taking agency away from the user regarding the fee they are willing to pay. Some UIs may do this because they optimize for average users, but I still want to be able to build power-user tools that let you choose.

The fee suggestion by the gateway is a hack that will never be optimal, so at least we should retain some flexibility around it. Eventually I'd like to build an API where you can instruct the gateway to probe to more correctly guess the appropriate fee upfront.

joschisan: Could you please check the pr above for this, and repost it there if applicable. I kept the gateway flexible to allow for  additional fees but set the default client behaviour to no additional fees.

PR #4971, opened 2024-04-16
Push uxnplqqnvozw

dpc:
I need a way to verify certain behaviors manually between versions, etc. Similiar to upgrade tests etc.

It occurred to me that it should be possible to just start devimint skipping all the DKGs and other setups, and it should be relatively easy. So I implemeted it.

One blocker that I've hit was inability to tell fedimintd to use certain URL to connect to bitcoind, instead of whatever was added to the config. I made new env variables that can be used to force bitcoind kind&url. The way existing env vars are working is confusing (they are only used as a default for DKG, so I renamed them while at it (with backward-compat.  of course).

Another blocker was that on killing devimint, we send `kill -9` right away, which makes `bitcoind` not get any chance to flush database to disk, and thus lose block history. All other daemons get really upset about it. So I changed it to `kill`, 3s wait, `kill -9` sequence. A bit slower (`fedimimtd` takes some time to shutdown), but works, and seem better. We can fix & improve with time.

The implementation is a bit meh, but can be improved with time. I would also like to modularize the setup so e.g. test that doesn't need LND, or electrs, etc. does not need to start them, and can configure it via builder.

## Testing

Start devimint e.g. like this:

```
> just devimint-env
...
2024-04-13T04:29:23.013121Z  INFO fm::devimint: Devimint data dir path=/tmp/devimint-3321804-158
```

Stop it, and then start it again, but skip the setup.

```
env FM_TEST_DIR=/tmp/devimint-3321804-158 FM_SKIP_SETUP=true RUST_LOG=fm::devimint=debug just devimint-env
```

It takes it sweet time as LND Gateway blocks on LND starting to return route hints, but works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,12 +3,14 @@
 /hfuzz_target/
 /hfuzz_workspace/
 /target-comp-bench/
+/tmp/
 # in case it's a symlink
 /target
 /target-nix
 /hfuzz_target
 /hfuzz_workspace
 /target-comp-bench
+/tmp
 
 /result
 /.shrc.local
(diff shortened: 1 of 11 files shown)


dpc:
The right PR is in another castle.  #4972

Bare with me. Still learning.

PR #4975, opened 2024-04-16
fix: use default total send fee provided by the gateway

joschisan:
The reasoning here is that the gateway is supposed to choose a fee for sending an outgoing payment that is sufficient for high lightning reliability; retrying with higher fees in case of cancellation is unnecessary. 

Furthermore, the fee behaviour is now identical to receiving an incoming payment.

### DIFF
--- a/gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
+++ b/gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
@@ -36,7 +36,7 @@ pub struct SendSMCommon {
     pub operation_id: OperationId,
     pub contract: OutgoingContract,
     pub max_delay: u64,
-    pub max_fee_msat: u64,
+    pub min_contract_amount: Amount,
     pub invoice: Bolt11Invoice,
     pub claim_keypair: KeyPair,
 }
@@ -90,9 +90,9 @@ impl State for SendStateMachine {
                     Self::send_payment(
                         context.clone(),
                         self.common.max_delay,
-                        self.common.max_fee_msat,
+                        self.common.min_contract_amount,
                         self.common.invoice.clone(),
-                        self.common.contract.amount,
+                        self.common.contract.clone(),
                     ),
                     move |dbtx, result, old_state| {
                         Box::pin(Self::transition_send_payment(
@@ -122,9 +122,9 @@ impl SendStateMachine {
     async fn send_payment(
         context: GatewayClientContextV2,
         max_delay: u64,
-        invoice_msats: u64,
+        min_contract_amount: Amount,
         invoice: Bolt11Invoice,
-        contract_amount: Amount,
+        contract: OutgoingContract,
     ) -> Result<[u8; 32], Cancelled> {
         // The following three checks may fail in edge cases since they have inherent
         // timing assumptions. Therefore, they may only be checked after we have created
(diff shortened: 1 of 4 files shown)


elsirion:
Why unnecessarily limit options? This will just be a headache compatibility-wise once we want to introduce a more dynamic protocol for whatever reason.

joschisan:
> Why unnecessarily limit options? This will just be a headache compatibility-wise once we want to introduce a more dynamic protocol for whatever reason.

If you look into the pr, you will find that this is only a client side change.

The Idea here for the send fee (and possibly also the time lock expiration delta) is the following. 

Successfully routing a payment depends on many factors, two of which are maximum fee and maximum cults expiry. If the gateway only sets a minimum fee for itself and a minimum expiration delta required for the gateway to claim the payment, it is left to the user to guess how much additional fee and expiration delta is required to route a payment successfully. Furthermore, appropriate settings for both depend on the gateways connectivity and may vary between gateways. Hence the user either chooses a high default for both (which we do right now) or the user may has to do multiple cancelled attempts with different fees and expiration deltas.

Alternatively, if we assume that reliability, meaning payments are successfully routed on the first try, is our priority, we can let the gateway set higher minimums for send fee and expiration delta, that includes a fee and time lock delta for lightning such that, if the payment still cannot be routed, it can be assumed by the user that the failure was very likely not caused by insufficient fees or expiration delta. Retrying with higher fees and expiration delta is very likely to fail again, the user should try other gateways if available or consider the invoice unpayable, receiver may be offline for example.

So the statement the gateway makes with the returned fee and expiration delta is 

"this is the minimum fee and expiration delta I require, please balance reliability, fees and timeout on your own" 

but instead 

"I can provide a reliable service for this fee and expiration delta"

which seems to me to be more appropriate to the less technical end user we generally assume.

Thoughts?

joschisan:
@elsirion and I decided to client-side rely on a default fee returned by the gateway, however we still leave the option for a custom fee such that a fee-sensitive user may probe for a specific invoice before setting a send fee.

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
```diff
@@ -137,18 +137,11 @@ impl SendStateMachine {
...
-            .add_fee(invoice_msats);
-
-        if contract_amount < min_contract_amount {
+        if contract.amount < min_contract_amount {
             return Err(Cancelled::Underfunded);
         }
 
-        let excess_fee = contract_amount - min_contract_amount;
-        let max_fee_msat = excess_fee.msats + (min_contract_amount.msats - invoice_msats) / 2;
+        let max_fee_msat = (contract.amount - min_contract_amount).msats;
```
elsirion, line 144: nit: `checked_sub` could do both in one

gateway/ln-gateway/src/lib.rs
```diff
@@ -1631,6 +1622,13 @@ impl Gateway {
             .amount_milli_satoshis()
             .ok_or(anyhow!("Invoice is missing amount"))?;
 
+        let min_contract_amount = self
+            .payment_info_v2(&payload.federation_id)
+            .await
+            .ok_or(anyhow!("Payment Info not available"))?
+            .send_fee_minimum
+            .add_fee(invoice_msats);
```
elsirion, line 1630: Not sure why you moved that here, but you probably had a good reason

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -134,16 +134,12 @@ pub struct SendPaymentPayload {
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Decodable, Encodable)]
 pub struct PaymentInfo {
     pub public_key: PublicKey,
-    pub payment_fees: PaymentFees,
+    pub send_fee_minimum: PaymentFee,
+    pub send_fee_default: PaymentFee,
+    pub receive_fee: PaymentFee,
```
elsirion, line 139: :+1: could use some docs on what these two send fees mean I guess

elsirion (approved):

PR #4982, opened 2024-04-16
feat: fedimint-cli module ln pay/invoice

dpc:
I wanted to see if architecturally moving cli handling into corresponding client modules works OK, so I decided to move `ln-pay` and `ln-invoice` first.

Old commands are still there for backward-compat, but hidden and will print warning to use new ones.

However I had to change the `ln-pay` to not display a breakdown of  notes anymore, because that comes from another module, that might not even be there, so kind of a bad architecture anyway. We'll see if any test complains.

While at it, I changed `ln pay` to take amount as a positinal arg. See commit description.


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1590,6 +1592,108 @@ impl LightningClientModule {
...
         }
     }
+
+    pub async fn wait_for_ln_payment(
+        &self,
+        payment_type: PayType,
+        contract_id: ContractId,
+        return_on_funding: bool,
+    ) -> anyhow::Result<Option<serde_json::Value>> {
+        // let lightning_module = client.get_first_module::<LightningClientModule>();
```
elsirion, line 1613: nit: commented out code

Cargo.lock
```diff
@@ -1906,6 +1906,7 @@ dependencies = [
  "fedimint-client",
  "fedimint-core",
  "fedimint-ln-common",
+ "fedimint-logging",
```
elsirion, line 1910: How did this happen?! There was no `Cargo.toml` change :eyes:

fedimint-cli/src/client.rs
```diff
@@ -99,6 +100,7 @@ pub enum ClientCmd {
         oob_notes: Vec<OOBNotes>,
     },
     /// Create a lightning invoice to receive payment via gateway
+    #[clap(hide = true)]
```
elsirion, line 103: Clever, so it keeps working in tests while we teach people to use the new CLI commands …

modules/fedimint-ln-client/src/cli.rs
```diff
@@ -14,7 +14,6 @@ use crate::OutgoingLightningPayment;
 enum Opts {
     /// Create a lightning invoice to receive payment via gateway
     Invoice {
-        #[clap(long)]
```
elsirion: :pray:

elsirion (approved):

fedimint-cli/src/client.rs
elsirion: We should still try to, over time, migrate tests so we actually test the real thing.

Cargo.lock
dpc: It might be in a different commit. I did change Cargo.tomls and there was a rebase conflict at some point on `Cargo.lock`.

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -526,9 +528,7 @@ impl LightningClientModule {
         let text_gateway_id = response.text().await?;
         let gateway_id = PublicKey::from_str(&text_gateway_id[1..text_gateway_id.len() - 1])?;
         if gateway_id != gateway.gateway_id {
-            return Err(anyhow::anyhow!(
-                "Unexpected gateway id returned: {gateway_id}"
-            ));
+            return Err(anyhow!("Unexpected gateway id returned: {gateway_id}"));
```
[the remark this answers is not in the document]
dpc: Agreed. `anyhow` ties to much to the specific library.

fedimint-cli/Cargo.toml
```diff
@@ -39,7 +39,7 @@ fedimint-api-client = { workspace = true }
 fedimint-rocksdb = { version = "=0.4.0-alpha", path = "../fedimint-rocksdb" }
 fedimint-mint-client = { version = "=0.4.0-alpha", path = "../modules/fedimint-mint-client" }
 fedimint-mint-common = { version = "=0.4.0-alpha", path = "../modules/fedimint-mint-common" }
-fedimint-ln-client = { version = "=0.4.0-alpha", path = "../modules/fedimint-ln-client" }
+fedimint-ln-client = { version = "=0.4.0-alpha", path = "../modules/fedimint-ln-client", features = [ "cli" ] }
```
bradleystachurski, line 42: Should we instead do `workspace = true`?

bradleystachurski (approved):

fedimint-cli/Cargo.toml
dpc: Always. But here I just did a quick amend and forgot.

PR #5003, opened 2024-04-18
chore: don't assume that a inputs / outputs know their amounts,  track amounts in ClientInput / ClientOutput instead

joschisan:
Currently we assume that a module inputs / outputs amount can be determined from the input / output itself. However, this is not necessarily the case of lnv2 inputs as those contain just the contract id of an unspent contract in the guardians database and a witness to claim the contract.

Furthermore, for blinded amounts neither the inputs nor outputs contain the amounts.

We can track the amounts in ClientInput / ClientOutput instead, and change the input_amount / output_amount method of the ClientModule trait to input_fee / output_fee. We maintain the ability to return different fees for different inputs / outputs since ln legacy needs this for outputs.


dpc (approved):
LGTM. Would be good for @elsirion to ack.

elsirion:
> Currently we assume that a module inputs / outputs amount can be determined from the input / output itself.

That's not entirely correct (and the docs are a bit confusing tbf). If the input/output always knew its amount we'd just impl a `amount()` for `DynInput`/`DynOutput`. What the current API gives us is the ability to use module-internal state to determine these amounts.

That being said, I can see how that could complicate things and the high-level of the change looks good. I'll take a closer look now.

elsirion (approved):

PR #5011, opened 2024-04-18
chore: make fedimintd and devimint shutdown stuff fast

dpc:
See commits.

### Testing

Start devimint with:

```
env RUST_LOG=fm=debug just devimint-env
```

`exit` in it, and observe the shutdown time. `tail -f target-nix/devimint/logs/fedimintd-0.log` if you want to see what's happening in `fedimintd`. Another way is to `killall fedimintd` from the outside with `tail -f`.

### DIFF
--- a/devimint/src/util.rs
+++ b/devimint/src/util.rs
@@ -86,14 +86,7 @@ pub struct ProcessHandle(Arc<Mutex<ProcessHandleInner>>);
 impl ProcessHandle {
     pub async fn terminate(&self) -> Result<()> {
         let mut inner = self.0.lock().await;
-        if let Some(mut child) = inner.child.take() {
-            debug!(
-                target: LOG_DEVIMINT,
-                "sending SIGTERM to {} and waiting for it to exit", inner.name
-            );
-            send_sigterm(&child);
-            child.wait().await?;
-        }
+        inner.terminate().await?;
         Ok(())
     }
     pub async fn is_running(&self) -> bool {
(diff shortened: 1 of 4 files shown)


fedimintd/src/fedimintd.rs
```diff
@@ -331,30 +330,24 @@ impl Fedimintd {
 
         let timing_total_runtime = timing::TimeReporter::new("total-runtime").info();
 
-        // DO NOT REMOVE, or spawn_local tasks won't run anymore
-        let local_task_set = tokio::task::LocalSet::new();
```
dpc, line 335: I guess this is not needed anymore, probably after alephbft refactoring.

devimint/src/util.rs
```diff
@@ -107,31 +100,60 @@ pub struct ProcessHandleInner {
...
+                target: LOG_DEVIMINT,
+                name=%self.name,
+                signal="SIGTERM",
+                "sending signal to terminate child process"
+            );
+
+            send_sigterm(child);
+
+            if (fedimint_core::runtime::timeout(Duration::from_secs(2), child.wait()).await)
+                .is_err()
```
maan2003, line 116: given fedimintd sleeps for 10s on shutdown, this will always send SIGKILL?

dpc: I fixed that. Now it shutdowns in <1s.

fedimintd/src/fedimintd.rs
```diff
@@ -367,14 +360,8 @@ impl Fedimintd {
                     sleep(SHUTDOWN_TIMEOUT).await;
```
maan2003, line 360: aren't we always sleeping for 10s herre

dpc: I don't understand what happened. I'm quite sure I was not seeing it.

But anyway - I got rid of this now, and also nested api runtime shenanigans that were causing a panic.

maan2003 (approved):

dpc:
Seems like this saves 2 mins on test-ci-all in the CI, from around 13:20 to 11:20.

maan2003:
needs further review for 1419130

Issue #4926, opened 2024-04-11
Slowest unit-tests


dpc:
```
        PASS [  34.959s] fedimint-ln-tests::fedimint_ln_tests gateway_protects_preimage_for_payment
        PASS [  40.010s] fedimint-server net::peers::tests::test_connect
        SLOW [> 60.000s] fedimint-ln-gateway::gatewayd-integration-tests test_gateway_cannot_pay_expired_invoice
        SLOW [> 60.000s] fedimint-ln-gateway::gatewayd-integration-tests test_gateway_client_pay_unpayable_invoice
        PASS [  66.788s] fedimint-ln-gateway::gatewayd-integration-tests test_gateway_client_pay_unpayable_invoice
        PASS [  68.614s] fedimint-ln-gateway::gatewayd-integration-tests test_gateway_cannot_pay_expired_invoice
```

dpc:
If you look at `env RUST_BACKTRACE=1 c nextest run  -E 'test(test_gateway_cannot_pay_expired_invoice)' --nocapture`

It gets stuck for a long time on:

```
2024-04-11T04:11:43.630878Z  WARN task{name="state machine executor"}: fedimint_ln_client::pay: Error while trying to reach gateway: Lightning Gateway failed to
 pay invoice. ErrorCode: Some(500) ErrorMessage: An internal gateway error occurred
2024-04-11T04:11:44.633183Z ERROR pay_invoice{payload=PayInvoicePayload { federation_id: FederationId(3559d055a1ca14eacb19b49af071beb92cbbaf607cb3a3bf6fe5a9eb03
19dbdb), contract_id: c7061d6728bf61b3948340818a9d8b190c637a71a57090bf5c484f021cca0a74, payment_data: Invoice(Bolt11Invoice { signed_invoice: SignedRawBolt11Inv
oice { raw_invoice: RawBolt11Invoice { hrp: RawHrp { currency: Regtest, raw_amount: Some(10), si_prefix: Some(Micro) }, data: RawDataPart { timestamp: PositiveT
imestamp(1712808688s), tagged_fields: [KnownSemantics(Description(Description(""))), KnownSemantics(PaymentHash(Sha256(66687aadf862bd776c8fc18b8e9f8e20089714856
[...]

PR #5015, opened 2024-04-18
 chore: disable meh ln-gateway integration tests 

dpc:
I just investigated #4926 a bit more, and discovered that two offending tests basically test expiration of the GW client state machine, and due to hacks they are using, they were not even succeeding (by failing) due to right error. They removed the internal client and then they were getting an error that the gateway did not joined the the corresponding federation.

Combined with the fact that they can't distinguish errors and need to wait for the full expiration time (30s), and we run it twice for each gw type, it makes them run the whole minute... to basically assert that they are broken.

There really has to be a more robust way to go about everything here.

Either we should make the error handling solid to detect the error and terminate state machine immediately after the irrecoverable condition is returned in the error, or we should maybe set up the gateway and mock responses to it, or just do something better.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,6 +1,6 @@
 [profile.default]
 # no matter the profile, we want to kill tests that hanged on something
-slow-timeout = { period = "60s", terminate-after = 8 }
+slow-timeout = { period = "30s", terminate-after = 3 }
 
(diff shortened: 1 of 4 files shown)


dpc:
OK, maybe we can just make them waste less time in #5018

PR #5016, opened 2024-04-19
chore(ci): speed up test_connect

dpc:
Turn a 40s test into 2s test.

### DIFF
--- a/fedimint-server/src/net/peers.rs
+++ b/fedimint-server/src/net/peers.rs
@@ -735,13 +735,15 @@ mod tests {
     use std::collections::HashMap;
     use std::time::Duration;
 
-    use fedimint_core::task::{sleep, TaskGroup};
+    use anyhow::{ensure, Context as _};
+    use fedimint_core::task::TaskGroup;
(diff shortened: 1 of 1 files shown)


fedimint-server/src/net/peers.rs
```diff
@@ -788,24 +790,32 @@ mod tests {
...
-            let status = peer_status_client_a.get_all_status().await;
-            assert_eq!(status.len(), 2);
-            assert!(status.values().all(|s| s.is_ok()));
+            async fn wait_for_connection(name: &str, peer_status: &PeerStatusChannels) {
+                retry(
+                    format!("wait for client {}", name),
+                    fedimint_core::util::FibonacciBackoff::default()
+                        .with_min_delay(Duration::from_millis(200))
+                        .with_max_delay(Duration::from_secs(5))
+                        .with_max_times(10),
```
maan2003, line 799: max delay = 20.8s, so should preserve old behavior in worst case

maan2003 (approved):

PR #5020, opened 2024-04-19
chore: preparations for bolt12 support

joschisan:


gateway/ln-gateway/src/db.rs
```diff
@@ -154,12 +154,19 @@ async fn migrate_to_v1(dbtx: &mut DatabaseTransaction<'_>) -> Result<(), anyhow:
 }
 
 #[derive(Debug, Encodable, Decodable)]
-pub struct CreateInvoicePayloadKey(pub [u8; 32]);
+pub struct RegisteredIncomingContractKey(pub [u8; 32]);
+
+#[derive(Debug, Encodable, Decodable)]
+pub struct RegisteredIncomingContract {
+    pub federation_id: FederationId,
+    pub incoming_amount: u64,
```
m1sterc001guy, line 162: `IncomingContract` has the amount in the `Commitment`, is this just for convenience?

modules/fedimint-lnv2-client/src/send_sm.rs
```diff
@@ -43,10 +43,23 @@ pub struct SendSMCommon {
     pub funding_txid: TransactionId,
     pub gateway_api: SafeUrl,
     pub contract: OutgoingContract,
-    pub invoice: Bolt11Invoice,
+    pub invoice: Invoice,
     pub refund_keypair: KeyPair,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+pub enum Invoice {
```
m1sterc001guy, line 51: Maybe this should go in the common crate?

gateway/ln-gateway/src/db.rs
m1sterc001guy: Or does this value include fees?

joschisan: The two amounts differ by the fee. incoming_amount is lower and is the amount in the invoice.

m1sterc001guy (approved):

modules/fedimint-lnv2-client/src/send_sm.rs
joschisan: I would not share this for cleaner separation.

gateway/ln-gateway/src/db.rs
m1sterc001guy: Can address in the future but a different name would help IMO

joschisan:
Please don't merge yet I am still thinking about it.

gateway/ln-gateway/src/db.rs
joschisan: incoming_htlc_amount?

modules/fedimint-lnv2-client/src/send_sm.rs
m1sterc001guy: Why is this cleaner? I see repeated code here and in the gateway client

m1sterc001guy (approved):
LGTM, only comment is that it looks like to me that the `Invoice` enum could be shared between the gateway client and the user client

Kodylow:
Note from call: moving things into enums like this is good practice for avoiding state machine migrations, @Kodylow open an issue for this

PR #5021, opened 2024-04-19
cleanup: dont use ln legacy to get the gateway api anymore

joschisan:


elsirion:
Looks simple enough and all non-trivial changes are in LNv2 anyway.

PR #5022, opened 2024-04-19
Removing validator Dependency

Minimega12121:
## Changes
Manually implemented validate function to replace previously implemented validate_peg_in_proof

```
pub fn validate(&self) -> Result<(), PegInProofError> {
        if !self.txout_proof.contains_tx(self.transaction.txid()) {
            return Err(PegInProofError::TransactionNotInProof);
        }

        if self.transaction.output.len() > u32::MAX as usize {
            return Err(PegInProofError::TooManyTransactionOutputs);
        }

        match self.transaction.output.get(self.output_idx as usize) {
            Some(txo) => {
                if txo.value > 2_100_000_000_000_000 {
                    return Err(PegInProofError::TxOutAmountOutOfRange);
                }
            }
            None => {
                return Err(PegInProofError::OutputIndexOutOfRange(
                    self.output_idx as u64,
                    self.transaction.output.len() as u64,
                ));
            }
        }

        Ok(());
}
```

## Why
To remove 'Validator' dependency which was previously only used once


elsirion:
@Minimega12121 have you looked into what the validator dependency does? We want to replicate exactly that behavior, just having the dependency isn't worth it when only used in one place.

elsirion:
See #5025 for what I roughly meant. Originally I thought we could just manually impl the serde logic, but turns out supporting JSON-like formats like that is a lot of work, so I decided to duplicate the struct to derive `Deserialize` and then do the check afterwards.

PR #5035, opened 2024-04-19
fix(ci): dependabot still not grouping patch updates

dpc:
Looking at

https://github.com/dependabot/dependabot-core/blob/main/.github/dependabot.yml

I think this part is off.

Would be nice to have a validator for this stuff...


bradleystachurski (approved):

maan2003:
one day we will write all config in nickel and we will have types for all of it

PR #5042, opened 2024-04-19
feat: minimal mainnet fedimintd docker setup

elsirion:
Addressing my own comment https://github.com/fedimint/fedimint/pull/4872#pullrequestreview-1989401146. FYI @Kodylow @TonyGiorgio to avoid duplicate work.

### DIFF
--- a/docker/0.3.0/fedimintd-mainnet-tls/.env
+++ b/docker/0.3.0/fedimintd-mainnet-tls/.env
@@ -1,0 +1,10 @@
+# This domain should point to the machine fedimintd is being deployed to
+FM_DOMAIN=my-super-host.com
+
+# Where bitcoind is reachable
+FM_BITCOIN_RPC_KIND=bitcoind
+FM_BITCOIN_RPC_URL=http://user@password@127.0.0.1:8332
+
+# For testing one can also use esplora
+# FM_BITCOIN_RPC_KIND=esplora
+# FM_BITCOIN_RPC_URL=https://blockstream.info/api/
--- a/docker/0.3.0/fedimintd-mainnet-tls/meta.json
+++ b/docker/0.3.0/fedimintd-mainnet-tls/meta.json
@@ -1,0 +1,5 @@
+{
+  "<federation_id_hex>": {
+    "federation_name": "Override Federation Name"
+  }
+}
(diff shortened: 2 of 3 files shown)


docker/0.3.0/fedimintd-mainnet-tls/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+  "<federation_id_hex>": {
+    "federation_name": "Override Federation Name"
+  }
+}
```
elsirion, line 5: TODO: generate that file automatically once the federation is set up

Kodylow:
Sorry thought I opened up the PR for it, this is covered in my new docker installer

Kodylow:
kk, if you mark it ready we can merge and I'll update docker deployer off it

docker/0.3.0/fedimintd-mainnet-tls/.env
```diff
@@ -0,0 +1,10 @@
+# This domain should point to the machine fedimintd is being deployed to
```
TonyGiorgio, line 1: Does docker automatically pull in the `.env` file? 

I saw something that said it might need to be something like this: 

```
docker-compose -f docker-compose.yml --env-file .env up
```

If that's the case, might be useful to put at the top of this file as a quick note.

elsirion: I think it does? It did for me iirc.

docker/0.3.0/fedimintd-mainnet-tls/meta.json
Kodylow: Do you want to land this change before merge or add a TODO for a follow up PR? @elsirion

docker/0.3.0/fedimintd-mainnet-tls/.env
elsirion: I was using `docker compose up` and not `docker-compose up` btw, if that makes a difference?

justinmoon:
@Kodylow @TonyGiorgio Could one of you review?

TonyGiorgio (approved):
This works, though a few different competing docker PR's at this point.

I'm more than likely going to close mine (https://github.com/fedimint/fedimint/pull/4269) and maintain it elsewhere. I have a preference towards it and it has worked flawlessly on many occasions. I don't know what the long term plan for making deployments easier, but I like that a minimum example is here.

docker/0.3.0/fedimintd-mainnet-tls/meta.json
elsirion: After thinking about it a bit more: it's not bad enough UX to worry about given that we are phasing out meta override files anyway

PR #5049, opened 2024-04-20
feat: relay incoming lightning payments for lnv2

joschisan:


m1sterc001guy (approved):
LGTM very similar to existing complete SM

gateway/ln-gateway/src/gateway_module_v2/mod.rs
```diff
@@ -279,10 +358,12 @@ impl GatewayClientModuleV2 {
             )
             .await?;
 
-        Ok(())
+        self.subscribe_receive(operation_id)
```
m1sterc001guy, line 361: Is this necessary to have here? Caller can always call this after calling `relay_direct_swap` and decide if they want to wait for a particular state or not.

m1sterc001guy (approved):
nit: I don't really like the "relay" verb, it sounds to me like its passing and HTLC to another LN node, when its actually revealing the preimage to the gateway's lightning node

PR #5050, opened 2024-04-21
docs: updating dev-env.md regarding nix installation

Minimega12121:
# Why
One of the main problems with Nix installation is that the official version requires you to change some settings in order to use `nix develop`. On the other hand, if you make a mistake during installation, uninstalling Nix can be challenging (as also mentioned as a con here: [https://nixos.org/download/](https://nixos.org/download/)). I faced this myself, so I'm recommending some of the uninstallation methods that worked for me.
# How
So, I have added an `Uninstallation` section in the `dev-setup.md` so that if someone installs Nix incorrectly or wants to uninstall it, they can do so using:
1. Official method
2. Brute-force method

Additionally, I have updated the recommended method to use the `Determinate Nix Installer` so that setting up the development environment goes smoothly for anyone attempting it for the first time (discussed with @Kodylow).

### DIFF
--- a/docs/dev-env.md
+++ b/docs/dev-env.md
@@ -31,10 +31,9 @@ it ensures consistent and reproducible environment for all developers.
 ### Install Nix
 
 You have 2 options to install nix:
-* [The official installer](https://nixos.org/download.html)
-* The [Determinate Nix Installer](https://zero-to-nix.com/start/install) which is maintained by a 3rd party, but is a little more user-friendly.
 
-If one doesn't work for you, consider trying the other. The end result is having a working `nix` command in your shell.
+* **RECOMMENDED:** The [Determinate Nix Installer](https://github.com/DeterminateSystems/nix-installer)
+* [The official installer](https://nixos.org/download.html)
 
 Example:
 


docs/dev-env.md
```diff
@@ -45,6 +44,22 @@ nix (Nix) 2.9.1
...
+
+Refer to the official Nix documentation for the recommended uninstallation process : [Official Docs](https://nixos.org/manual/nix/stable/installation/uninstall)
+
+#### Option 2: Brute Force Method
+
+If you encounter difficulties with the standard uninstallation or need to remove Nix forcefully, you can use the following steps:
+1. Use the command `which nix` to locate the Nix binaries.For example, this will give the path to the binaries like /nix/var/nix/profiles/default/bin/nix. For the next step, only consider /nix/var/nix/profiles/default/bin/, i.e., remove 'nix'.
+2. Open the folder containing the Nix files using `xdg-open`.Then search for nix.
+3. Many nix binaries will come for example `nix`, `nix-env`. Use the command `sudo rm nix` and similar commands to delete all files and directories related to Nix (ie that have nix in them).
+4. Once you've removed the Nix files, you can proceed to reinstall Nix using the Determinate Nix Installer mentioned above.
```
maan2003, line 61: looks scary and possibly incorrect, you are never removing /nix/store for example.

where did you get these instructions?

Minimega12121: My mistakes if I explained it incorrectly. Basically the problem I was facing was that even after the official uninstallation it would still show that nix is installed. So I only removed the executables of nix not any data to allow for reinstallation. 
Seems like there is a typo from my part the command is 'xdg-open .', This is for linux use only. And the executables I deleted are below:

Minimega12121: nix/store gets removed by the officail uninstaller maybe but the actual executables does'nt get removed by following it. So I think  brute force step zero would be to execute the official uninstallation and then come to this as a last resort.

maan2003: will defer to @dpc, but I don't like recommending these instructions.

dpc:
We should just recommend using https://github.com/DeterminateSystems/nix-installer everywhere.

Minimega12121:
> We should just recommend using https://github.com/DeterminateSystems/nix-installer everywhere.
Okay, I'll change the doc accordingly

docs/dev-env.md
```diff
@@ -31,10 +31,9 @@ it ensures consistent and reproducible environment for all developers.
 ### Install Nix
 
 You have 2 options to install nix:
-* [The official installer](https://nixos.org/download.html)
-* The [Determinate Nix Installer](https://zero-to-nix.com/start/install) which is maintained by a 3rd party, but is a little more user-friendly.
 
-If one doesn't work for you, consider trying the other. The end result is having a working `nix` command in your shell.
+* [RECOMMENDED] The [Determinate Nix Installer](https://github.com/DeterminateSystems/nix-installer)
```
elsirion, line 35: nit: would just make this bold instead of using []

m1sterc001guy (approved):

dpc (approved):

PR #5051, opened 2024-04-21
fix: remove tokio_unstable

Kodylow:
Removes tokio_unstable `Builder`, turns the thread name into a tracing span, still enforces using fedimint_core's spawn and spawn_local.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
 [build]
-rustflags = ["--cfg", "tokio_unstable"]
+# rustflags = ["--cfg", "tokio_unstable"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,0 @@
-[build]
-# rustflags = ["--cfg", "tokio_unstable"]
--- a/.config/semgrep.yaml
+++ b/.config/semgrep.yaml
@@ -64,6 +64,9 @@ rules:
   message: "`tokio::spawn` won't set task name by default. use `fedimint_core::runtime::spawn` instead"
   pattern: tokio::spawn
   severity: WARNING
+  paths:
+    exclude:
+      - fedimint-core/src/runtime.rs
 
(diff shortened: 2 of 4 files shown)


.cargo/config.toml
```diff
@@ -1,2 +1,2 @@
 [build]
-rustflags = ["--cfg", "tokio_unstable"]
+# rustflags = ["--cfg", "tokio_unstable"]
```
dpc, line 2: Don't be shy. Just delete it. :D

.config/semgrep.yaml
```diff
@@ -64,6 +64,9 @@ rules:
   message: "`tokio::spawn` won't set task name by default. use `fedimint_core::runtime::spawn` instead"
   pattern: tokio::spawn
   severity: WARNING
+  paths:
```
dpc, line 67: Why?

flake.nix
```diff
@@ -104,7 +104,7 @@
           };
 
           toolchainArgs = {
-            extraRustFlags = "--cfg tokio_unstable";
+            # extraRustFlags = "--cfg tokio_unstable";
```
dpc, line 107: Same here. :)

dpc (dismissed):
Fine with me. Not worth the downstream trouble.

dpc (commented):
But maybe clear the nits why we wait for a second opinion.

.config/semgrep.yaml
dpc: Oh.

You can use a `// nosemgrep: <rule-name>` to just exclude specific places that need to break the rule.

Kodylow: done

dpc (approved):

justinmoon (approved):
LGTM. Seemed like everyone was in favor this morning in the dev call.

fedimint-backports:
- #5213

PR #5068, opened 2024-04-23
chore: dont broadcast request_admin_no_auth

m1sterc001guy:
I noticed `request_admin_no_auth` still broadcasted it's requests using `request_current_consensus`, instead of `request_single_peer_federation`, which is how `request_admin` works.


dpc (approved):

dpc:
Good catch.

Issue #4365, opened 2024-02-19
Improve `gateway-cli info` command


justinmoon:
- Should report the SCID being used for each federation
- Shouldn't show `config` by default since it's huge and makes the output difficult to read

jrakibi:
Hi @justinmoon, I'll start working on this issue

PsychoPunkSage:
Hi @justinmoon 
I see one PR for this issue and it is open for quite some time...
I would like to work on it if no one else is working on it...

tvolk131:
I believe Justin is busy for the next few weeks. Go for it!

m1sterc001guy:
Fixed with https://github.com/fedimint/fedimint/pull/5803

PR #5070, opened 2024-04-23
fix: exclude config from default output in gateway-cli info rpc

jrakibi:
This PR addresses the feedback from issue #4365 by modifying the `gateway-cli info` command to exclude the `config` section from the default output.

Changes:
* The config section is now hidden by default to prevent information overload and enhance the command's usability.
* Users can still view the config details by using an `--include-config`

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -11,13 +11,15 @@ use fedimint_logging::TracingSetup;
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
 use ln_gateway::rpc::{
     BackupPayload, BalancePayload, ConfigPayload, ConnectFedPayload, ConnectToPeerPayload,
-    DepositAddressPayload, FederationRoutingFees, GetFundingAddressPayload, LeaveFedPayload,
-    OpenChannelPayload, RestorePayload, SetConfigurationPayload, WithdrawPayload, V1_API_ENDPOINT,
+    DepositAddressPayload, FederationRoutingFees, GatewayInfo, GetFundingAddressPayload,
+    LeaveFedPayload, OpenChannelPayload, RestorePayload, SetConfigurationPayload, WithdrawPayload,
+    V1_API_ENDPOINT,
 };
 use serde::Serialize;
 
 const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRIES: u32 = 12;
 const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRY_DELAY_SECONDS: u64 = 10;
+const DEFAULT_EXCLUDE_CONFIG: bool = true;
 
 #[derive(Parser)]
 #[command(version)]
@@ -37,7 +39,10 @@ pub enum Commands {
     /// Display CLI version hash
     VersionHash,
     /// Display high-level information about the Gateway
-    Info,
+    Info {
+        #[clap(long)]
+        exclude_config: Option<bool>,
+    },
     /// Display config information about the Gateways federation
     Config {
         #[clap(long)]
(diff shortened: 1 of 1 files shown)


m1sterc001guy:
This will unfortunately break backwards compatibility. To maintain backwards compatibility, you will need to change the `include_config` flag to be `exclude_config` so that by default the config is included. This way, old clients can continue to make the RPC as they did before, but new clients can add the `exclude_config` flag to have a more readable output.

The devimint tests will need to be updated. Then after a major release, we can invert the flag (or remove it entirely) so that its excluded by default.

jrakibi:
@m1sterc001guy, done. I updated the flag from `include_config` to `exclude_config`, but it seems that the job for backwards-compatibility is still failing

dpc (dismissed):

m1sterc001guy:
We need to get to the bottom of the backwards compatibility failures. Is there a default value for `exclude_config`? If there's no default value, I would assume that old CLI versions will fail since they do no understand the `exclude_config` flag.

dpc:
BTW. Probably good idea to make this controlled via env var (optionally), so in scritpts and for backward compat it can be set once for all invocations.

elsirion:
> old CLI versions will fail since they do no understand the `exclude_config` flag.

Old clients won't send the exclude flag, so new servers need a default. New clients may send the exclude flag, but old servers should ignore it due to JSON.

jrakibi:
The **exclude_config** parameter now defaults to **true** and is **optional**. However, the backward compatibility job is still failing. I'm not sure if this is related or not

m1sterc001guy:
Try rebasing and let's re-run the tests and see if it is still broken

jrakibi:
> Try rebasing and let's re-run the tests and see if it is still broken

Done

gateway/cli/src/main.rs
```diff
@@ -155,6 +163,9 @@ pub enum LightningCommands {
         /// The delay between retries
         #[clap(long)]
         retry_delay_seconds: Option<u64>,
+
+        #[clap(long)]
```
m1sterc001guy, line 167: nit: you can use `default_value` in clap

gateway/cli/src/main.rs
```diff
@@ -324,7 +350,9 @@ async fn main() -> anyhow::Result<()> {
                 block_height,
                 max_retries,
                 retry_delay_seconds,
+                exclude_config,
```
m1sterc001guy, line 353: You're just throwing away the config below anyway since the only field that is used is `block_height`, so this doesn't need to be a parameter. Just set it to true in the actual `get_info` RPC

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -65,7 +68,8 @@ pub struct WithdrawPayload {
 pub struct FederationInfo {
     pub federation_id: FederationId,
     pub balance_msat: Amount,
-    pub config: ClientConfig,
+    #[serde(skip_serializing_if = "Option::is_none")] // Skip serializing if config is None
+    pub config: Option<ClientConfig>,
```
m1sterc001guy, line 72: Hmm so the backwards compatibility tests might be failing because this changed from `ClientConfig` to `Option<ClientConfig>` (havent confirmed myself yet).

I don't think changing this is actually necessary. The main issue with the `ClientConfig` is just that it clutters up the CLI response. I think we should still return `ClientConfig` on the `get_info` requests, but on the CLI just filter out the config according to the `exclude_config` flag. This will make it so the CLI is more readable but also keep the change simple on the API/server side.

jrakibi: Oh perfect, sounds like a straightforward solution then. We will just update one file. 

I've made the changes and updated the PR. Can you please review it again and run the jobs to check if the backward compatibility issue still occurs?

elsirion:
Somewhere in the tests we assume that the config field is present, needs to be changed for this PR. https://github.com/fedimint/fedimint/actions/runs/9142510136/job/25229423380?pr=5070#step:9:250

m1sterc001guy:
Superceding with https://github.com/fedimint/fedimint/pull/5803

PR #5076, opened 2024-04-23
 fix(ci): not running enough tests in parallel 

dpc:

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -103,6 +103,7 @@ pub async fn latency_tests(
     dev_fed: DevFed,
     r#type: LatencyTest,
     upgrade_clients: Option<&UpgradeClients>,
+    iterations: usize,
 ) -> Result<()> {
     log_binary_versions().await?;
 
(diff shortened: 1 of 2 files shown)


devimint/src/tests.rs
```diff
@@ -136,17 +137,11 @@ pub async fn latency_tests(
...
 
-    // LN operations take longer, we need less iterations
-    let iterations = 20;
-
     let cln_gw_id = gw_cln.gateway_id().await?;
 
     match r#type {
         LatencyTest::Reissue => {
             info!("Testing latency of reissue");
-            // On AlephBFT session times may lead to high latencies, so we need to run it
```
dpc, line 147: Not worth waiting for, plus probably requires like hundreds of iterations to really hit it.

scripts/tests/test-ci-all.sh
```diff
@@ -237,7 +237,7 @@ fi
 if [ -n "${FM_TEST_CI_ALL_JOBS:-}" ]; then
   # when specifically set, use the env var
   parallel_args+=(--jobs "${FM_TEST_CI_ALL_JOBS}")
-elif [ -n "${CI:-}" ]; then
+elif [ -n "${CI:-}" ] || [ "${CARGO_PROFILE:-}" == "ci" ]; then
```
dpc, line 240: This is the main reason. We were actually running CI with 3 jobs parallel max, while we wanted 8 (num cpus).

bradleystachurski: Interesting, I would have expected `CI=true` from the GHA docs

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

bradleystachurski (approved):

scripts/tests/test-ci-all.sh
dpc: That was the mistake. It runs in nix derivation so CI envvar is not set! 😂

dpc: I probably assumed the same thing and didn't notice.

bradleystachurski: Ah! Nice find 😄

PR #5077, opened 2024-04-23
chore: switch to self-hosted runners

dpc:

### DIFF
--- a/scripts/ci/run-in-fs-dir-cache.sh
+++ b/scripts/ci/run-in-fs-dir-cache.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+job_name="$1"
+shift 1
+
+if [ -z "$job_name" ]; then
+    >&2 "error: no job name"
+    exit 1
+fi
+
+export FS_DIR_CACHE_ROOT="$HOME/.cache/fs-dir-cache" # directory to hold all cache (sub)directories
+export FS_DIR_CACHE_LOCK_ID="pid-$$-rnd-$RANDOM"     # acquire lock based on the current pid and something random (just in case pid gets reused)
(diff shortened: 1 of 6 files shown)


scripts/ci/run-in-fs-dir-cache.sh
```diff
@@ -0,0 +1,43 @@
...
+export FS_DIR_CACHE_LOCK_ID="pid-$$-rnd-$RANDOM"     # acquire lock based on the current pid and something random (just in case pid gets reused)
+export FS_DIR_CACHE_KEY_NAME="$job_name"             # the base name of our key
+export FS_DIR_CACHE_LOCK_TIMEOUT_SECS="$((60 * 30))" # unlock after timeout in case our job fails misereably
+
+log_file="$FS_DIR_CACHE_ROOT/log"
+
+fs-dir-cache gc unused --seconds "$((5 * 24 * 60 * 60))" # delete caches not used in more than a 5 days
+
+# create/reuse cache (sub-directory) and lock it (wait if already locked)
+cache_dir=$(fs-dir-cache lock --key-file Cargo.lock --key-str "${CARGO_PROFILE-:dev}" --key-file flake.lock)
```
elsirion, line 23: Won't this lead to higher latency when backwards compat tests have to wait for each other? I guess that's ok because most of the time this caching will save 90% of the compile time?

elsirion (approved):
Let's try this! :tada:

scripts/ci/run-in-fs-dir-cache.sh
dpc: It would, and I think it still would have been a fair tradoff, but actually the setup on NixOS self-hosted runners is that that each instance get its own `HOME`, so they can sometimes reuse their own previous run, and use some more space, but never wait for each other. @elsirion

PR #5079, opened 2024-04-24
wip: feat: more generic primary module

maan2003:
replace primary module with balance provider.

this is more flexible, you can use inputs from multiple modules

example: you have mint-v2 and mint modules, you can inputs from both mint and mint-v2, but only use output of mint-v2.

### DIFF
--- a/fedimint-client/src/balance_provider.rs
+++ b/fedimint-client/src/balance_provider.rs
@@ -1,0 +1,152 @@
+use fedimint_core::{
+    apply, async_trait_maybe_send,
+    core::{ModuleInstanceId, OperationId},
+    db::DatabaseTransaction,
+    task::{MaybeSend, MaybeSync},
+    util::BoxStream,
+    Amount, OutPoint,
+};
+
+use crate::{
+    module::DynClientModule,
+    transaction::{ClientInput, ClientOutput},
+    Client,
+};
+
+#[apply(async_trait_maybe_send!)]
+pub trait BalanceProvider: 'static + MaybeSend + MaybeSync {
+    /// Creates an input of **at least** a given `min_amount` from the holdings
+    /// managed by the module.
+    ///
+    /// If successful it returns:
+    /// * A set of private keys belonging to the input for signing the
+    ///   transaction
+    /// * The input of **at least** `min_amount`, the actual amount might be
+    ///   larger, the caller has to handle this case and possibly generate
+    ///   change using `create_change_output`.
+    /// * A closure that generates states belonging to the input. This closure
+    ///   takes the transaction id of the transaction in which the input was
+    ///   used and the input index as input since these cannot be known at time
+    ///   of calling `create_funding_input` and have to be injected later.
+    ///
+    /// The function returns an error if the client's funds are not sufficient
+    /// to create the requested input.
+    async fn create_sufficient_input(
+        &self,
+        client: &Client,
+        dbtx: &mut DatabaseTransaction<'_>,
+        operation_id: OperationId,
+        min_amount: Amount,
+    ) -> anyhow::Result<Vec<ClientInput>>;
+
+    /// Creates an output of **exactly** `amount` that will pay into the
+    /// holdings managed by the module.
+    ///
+    /// It returns:
+    /// * The output of **exactly** `amount`.
+    /// * A closure that generates states belonging to the output. This closure
+    ///   takes the transaction id of the transaction in which the output was
+    ///   used and the output index as input since these cannot be known at time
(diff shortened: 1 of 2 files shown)


fedimint-client/src/module/mod.rs
```diff
@@ -725,38 +645,69 @@ pub trait IClientModule: Debug {
 
     async fn backup(&self, module_instance_id: ModuleInstanceId)
         -> anyhow::Result<DynModuleBackup>;
+}
 
-    fn supports_being_primary(&self) -> bool;
-
+#[apply(async_trait_maybe_send!)]
+pub trait BalanceProvider: 'static + MaybeSend + MaybeSync {
```
maan2003, line 651: gist of PR: move methods of primary module to a trait that can be implemented by anything, not just a module

elsirion: I think directly implementing this for modules will be challenging. Primary modules had a separate trait in the beginning (much like proposed here, just with a different name) and it was a huge PITA. 

What could work is a mix of both approaches: the module trait has a `fn as_balance_provider(&self) -> Option<DynBalanceProvider>` which can return an independent struct or a clone of the module, just put into a different dyn box.

When specifying a balance provider we have to keep in mind that modules aren't initialized yet when we call builder functions. I'd propose giving the builder a function that, given the set of all initialized modules, creates a `DynBalanceProvider`. This function could either select one module or construct a more complicated struct that contains multiple balance providers and selects the appropriate one depending on context. This fn could default to selecting the first module returning `Some`.

maan2003: `Module` don't implement `BalanceProvider`, `ModuleBalanceProvider`.
trick is to make `BalanceProvider` state free, it takes a `Client` as argument in all functions. 

we can do
```
builder.with_balance_provider(mint_client::MintBalanceProvider::default())
```

fedimint-client/src/module/mod.rs
```diff
@@ -725,38 +645,69 @@ pub trait IClientModule: Debug {
...
+    ///   takes the transaction id of the transaction in which the input was
+    ///   used and the input index as input since these cannot be known at time
+    ///   of calling `create_funding_input` and have to be injected later.
+    ///
+    /// The function returns an error if the client's funds are not sufficient
+    /// to create the requested input.
     async fn create_sufficient_input(
         &self,
-        module_instance: ModuleInstanceId,
+        client: &Client,
```
dpc, line 670: Modules already have access to `Client` via `ClientCtx` passed in `args` in `init`.

fedimint-client/src/module/mod.rs
dpc: What's the benefit of moving to another trait? Balance provider could work with `DynClientModule` all the same, and filter out ones that don't support it.

It seems really like an aestetic benefit only. @maan2003

fedimint-client/src/module/mod.rs
maan2003: this is not meant to be implemented for Module type
https://github.com/fedimint/fedimint/pull/5079#discussion_r1578481967

fedimint-client/src/module/mod.rs
elsirion: Thx for correcting my understanding @maan2003. I tend to agree with @dpc now (I think). 

The module trait can stay the same and the `BalanceProvider` trait can be seen as a tuple of two closures that, given all client modules, will create an input/output respectively (just like primary modules do).

A `DefaultBalanceProvider` could e.g. scan through all modules, looking for the first primary one and using that. If we ever had multiple e-cash modules and wanted to migrate from one to the other we'd make it more sophisticated and use all modules for creating inputs (with priority on the old one) and only create outputs for the new one, so over time all funds are migrated.

I don't know how to achieve that yet, but I could also imagine using different balance providers for different transactions eventually:

```rust
client.get_first_module::<LightningClientModule>()
    .with_balance_provider(MyBalanceProvider)
    .pay_bolt11_invoice(…)
    .await?;
```

maan2003: oh per transaction is interesting

maan2003: we can just make `pay_bolt11_invoice` have a balance provider argument, call site is always more flexible than storing  stuff

elsirion:
Should be coordinated with #4477. @joschisan thoughts on this proposal to make the primary module concept more flexible?

justinmoon:
added "needs further review" b/c might be good tuesday topic

bradleystachurski:
review club: we discussed on 2024-05-06 and we can add back the "needs further review" tag once this moves out of draft

PR #5081, opened 2024-04-24
fix(ci): self hosted runner dir running out of space

dpc:

### DIFF
--- a/scripts/ci/run-in-fs-dir-cache.sh
+++ b/scripts/ci/run-in-fs-dir-cache.sh
@@ -10,7 +10,13 @@ if [ -z "$job_name" ]; then
     exit 1
 fi
 
-export FS_DIR_CACHE_ROOT="$HOME/.cache/fs-dir-cache" # directory to hold all cache (sub)directories
+USER_ROOT="$HOME"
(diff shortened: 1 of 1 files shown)


scripts/ci/run-in-fs-dir-cache.sh
```diff
@@ -10,7 +10,13 @@ if [ -z "$job_name" ]; then
     exit 1
 fi
 
-export FS_DIR_CACHE_ROOT="$HOME/.cache/fs-dir-cache" # directory to hold all cache (sub)directories
+USER_ROOT="$HOME"
+if [ -n "${RUNNER_ROOT:-}" ]; then
```
bradleystachurski, line 14: Is `RUNNER_ROOT` defined from here?

https://github.com/NixOS/nixpkgs/blob/5bc03b163ad1fbd5339481e2d9827e9f8a658e60/nixos/modules/services/continuous-integration/github-runner.nix#L209-L218

bradleystachurski (approved):

scripts/ci/run-in-fs-dir-cache.sh
dpc: You are absolute correct. :)

PR #5092, opened 2024-04-25
fix: backwards compatibility tests are started with too much parallel…

elsirion:
…ism for our self-hosted CI

Since we don't use VMs and the CPU is a shared resource between all runners we shouldn't use all available cores for parallelism in one job.


dpc (approved):

bradleystachurski (approved):
Arrived at a similar conclusion in https://github.com/fedimint/fedimint/pull/5095

👍

PR #4360, opened 2024-02-18
feat(flake): use `sccache` in the dev shell

dpc:
This should improve compilation times when switching between branches etc.

You can view the sccache with `sccache --show-stats`:

```
> sccache --show-stats
Compile requests                   2093
Compile requests executed          1501
Cache hits                         1150
Cache hits (Rust)                  1150
Cache misses                        339
Cache misses (Rust)                 339
Cache timeouts                        0
Cache read errors                     0
Forced recaches                       0
Cache write errors                    0
Compilation failures                 11
Cache errors                          0
Non-cacheable compilations            0
Non-cacheable calls                 592
Non-compilation calls                 0
Unsupported compiler calls            0
Average cache write               0.008 s
Average compiler                  0.564 s
Average cache read hit            0.004 s
Failed distributed compilations       0

Non-cacheable reasons:
crate-type                          456
incremental                         125
-                                    11

Cache location                  Local disk: "/home/dpc/.cache/sccache"
Use direct/preprocessor mode?   yes
Version (client)                0.7.4
Cache size                          937 MiB
Max cache size                       10 GiB
```

PR #5089, opened 2024-04-25
fix(ci): disable sccache wrapper in backward-compat test

dpc:

PR #5093, opened 2024-04-25
fix(ci): skip setting RUSTC_WRAPPER to sccache in ci

bradleystachurski:
The self hosted runner has permission issues due to `sccache`. The original intent of adding `sccache` was to make compilation faster on local dev environments when switching branches (see: https://github.com/fedimint/fedimint/pull/4360), so we can skip using in CI.

```
sccache: error: Failed to open file for hashing: "/var/lib/github-runner/runner-01-aa/.cache/fs-dir-cache/backward-compat-dbd8d46d51c79e46551ed37757b6f2894385816d10ab85bd32a4888003d03915/target/ci/deps/libaleph_bft_types-4504822894dd6e72.rmeta"
```

This is only an issue for back-compat in CI since the tests need to run with `nix develop -c`, which uses the `shellHook` unlike the rest of CI.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -285,7 +285,10 @@
                     fi
                   fi
 
-                  export RUSTC_WRAPPER=${pkgs.sccache}/bin/sccache
+
+                  if [ "''${CARGO_PROFILE:-}" != "ci" ]; then
+                    export RUSTC_WRAPPER=${pkgs.sccache}/bin/sccache
+                  fi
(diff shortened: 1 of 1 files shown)


bradleystachurski:
Oh jeez, @dpc beat me to the punch 😄 

Closing in favor of https://github.com/fedimint/fedimint/pull/5089

PR #5098, opened 2024-04-25
test: FederationTestBuilder for building federations in rust tests

m1sterc001guy:
Adds `FederationTestBuilder` which allows us to customize how we build federations in our tests.

### DIFF
--- a/fedimint-testing/src/federation.rs
+++ b/fedimint-testing/src/federation.rs
@@ -14,7 +14,7 @@ use fedimint_core::db::mem_impl::MemDatabase;
 use fedimint_core::db::Database;
 use fedimint_core::invite_code::InviteCode;
 use fedimint_core::module::ApiAuth;
-use fedimint_core::task::TaskGroup;
+use fedimint_core::task::{block_in_place, TaskGroup};
 use fedimint_core::PeerId;
 use fedimint_logging::LOG_TEST;
 use fedimint_rocksdb::RocksDb;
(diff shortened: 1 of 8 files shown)


dpc:
> In our rusts tests, we by default run with 1 peer offline to simulate a down peer. However, `FederationTest` allows us to customize how many peers are offline.

Oh. I think we misunderstood each other in Discord.

I wanted to stop injecting extra unreliability when having any peers offline, because it then it affect the performance severely.


Some modules randomly not delivering their CIs is a inherent feature of alephbft. Every module needs to keep proposing the same CI, until it sees it in it's `process_consensus_item`. In a nuthsell. You can check `meta` module for how it keeps track of `desired`.



@m1sterc001guy

m1sterc001guy:
Ok, that's fine, I can refactor it so we don't inject unreliability when there's a peer offline. I would still like the ability to create a reliable federation with all peers online though. I think I'll just make a new function for that.

dpc:
> I would still like the ability to create a reliable federation with all peers online though. I think I'll just make a new function for that.

The whole should be a builder, and test could be able to add `.force_offline_peers(0)` to disable defaults, or something like that.

m1sterc001guy:
Refactored and added `FederationTestBuilder` which allows us to customize how we create test federations. By default, we run with 4 guardians and 1 is offline, with reliable connections. Number of offline peers and reliability of the connections can be customized but none of our tests currently use it.

dpc (approved):

PR #5100, opened 2024-04-25
fix: skip very slow test on old Fedimint

dpc:

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -777,12 +777,19 @@ pub async fn cli_tests(dev_fed: DevFed) -> Result<()> {
     assert_eq!(operation["operation_kind"].as_str().unwrap(), "wallet");
     assert!(operation["outcome"]["Claimed"].as_object().is_some());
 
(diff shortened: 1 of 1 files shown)


bradleystachurski (approved):
I'm a bit uncomfortable losing coverage here for back-compat. Would you be okay if we merge now to speed up CI and I'll take a stab at splitting backup&restore into a separate devimint test?

dpc:
> I'm a bit uncomfortable losing coverage here for back-compat. Would you be okay if we merge now to speed up CI and I'll take a stab at splitting backup&restore into a separate devimint test?

This test is slow because old version had to wait for session to close. We never going to fix it, and on new versions it's much faster.

This test doesn't even interact between versions, unless we would create backup on one version and restore on another, which I don't think backward compat test tests.

bradleystachurski:
That's fair, we're only losing coverage of backup/restore for old fedimintd <> old client, which we should be fairly certain is unchanging and we've thoroughly tested for months without issue.

PR #5101, opened 2024-04-25
chore: no need to cache `.cargo` anymore

dpc:
Workers will cache it around, and downloading from github is kind of slow anyway.


bradleystachurski (approved):
Should we also remove from `ci-nix.yml`?

https://github.com/fedimint/fedimint/blob/375a7ce9dedf00b6d9daa1396799ca6ce1d298d8/.github/workflows/ci-nix.yml#L56

dpc:
@bradleystachurski Need to verify this one, and then sure, why not.

PR #5105, opened 2024-04-25
fix(ci): use `--start-load`, not `--load`

dpc:
`--load` means `--start-load` + `--run-load`, and `--run-load` actually slows down tests if the load of the system is too high, which is exactly what we want to avoid!


dpc:
Nvmd, I've read docs for `niceload` by accident and got spooked.

PR #5111, opened 2024-04-26
chore: lower scheduling prio of most heavy compilation items

dpc:
On Linux `chrt` and `iotop` can be used to set the scheduling priority for a process.

We can use it to tactically make all heavy compilation processes have "idle" priority which means they will only execute if nothing else is trying to. AFAIK, this is a global setting and should be respected between cgroups etc.

Since we can now schedule more jobs on a single CI box, such change makes it less likely for time sensitive tasks (tests) to miss some deadline, while still utilizing resources when possible.

### DIFF
--- a/nix/flakebox.nix
+++ b/nix/flakebox.nix
@@ -227,10 +227,35 @@ let
   });
 
 
+  # copied and modified from flakebox, to add `runLowPrio`, due to mistake in flakebox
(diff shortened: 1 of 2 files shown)


elsirion (dismissed):

dpc:
For macos I used `writeShellScript` instead of `writeShellScriptBin`. Fixed.

bradleystachurski (approved):

elsirion (approved):

PR #5112, opened 2024-04-26
 chore(devimint): fix LN channel on starting from previous state 

dpc:

### DIFF
--- a/devimint/src/cfg/lnd.conf
+++ b/devimint/src/cfg/lnd.conf
@@ -5,6 +5,7 @@ rpclisten=0.0.0.0:{rpc_port}
 restlisten=0.0.0.0:{rest_port}
 noseedbackup=1
 wtclient.active=false
+sync-freelist=true
 
 [Bitcoin]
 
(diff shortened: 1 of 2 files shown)


devimint/src/devfed.rs
```diff
@@ -171,9 +171,10 @@ impl DevJitFed {
                 let bitcoind = bitcoind.get_try().await?.deref().clone();
                 let lnd = lnd.get_try().await?.deref().clone();
                 let cln = cln.get_try().await?.deref().clone();
-                if !skip_setup {
-                    open_channel(&process_mgr, &bitcoind, &cln, &lnd).await?;
-                }
+                // Note: We open new channel even if starting from existing state
+                // as ports change on every start, and without this nodes will not find each
+                // other.
+                open_channel(&process_mgr, &bitcoind, &cln, &lnd).await?;
```
elsirion, line 177: Could we just connect them manually? They should recognize their pub keys and update the connect info.

elsirion (approved):
One nit, but landing

devimint/src/cfg/lnd.conf
```diff
@@ -4,6 +4,10 @@ listen=0.0.0.0:{listen_port}
 rpclisten=0.0.0.0:{rpc_port}
 restlisten=0.0.0.0:{rest_port}
 noseedbackup=1
+wtclient.active=false
+sync-freelist=true
```
bradleystachurski, line 8: These look good. I was also playing with setting `reset-wallet-transactions=false` for restarts, but not necessary if we see fast startups.

https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf#L346

PR #5113, opened 2024-04-26
fix(wallet-client): ignores rpc env var settings

dpc:
At least in devimint wallet-client will ignore env vars because it's `bitcoind`, only to switch to defaults from config ... which are also `bitcoind`.

### DIFF
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -255,13 +255,7 @@ impl Context for WalletClientContext {}
 impl WalletClientModule {
     fn get_rpc_config(cfg: &WalletClientConfig) -> BitcoinRpcConfig {
         if let Ok(rpc_config) = BitcoinRpcConfig::get_defaults_from_env_vars() {
-            // TODO: Wallet client cannot support bitcoind RPC until the bitcoin dep is
-            // updated to 0.30
-            if rpc_config.kind != "bitcoind" {
-                rpc_config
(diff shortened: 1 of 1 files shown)


dpc:
Oh. I see. We set in devimint the Federation to contain:

```
        .attach_config_gen_params(
            WalletInit::kind(),
            WalletGenParams {
                local: WalletGenParamsLocal {
                    bitcoin_rpc: bitcoin_rpc.clone(),
                },
                consensus: WalletGenParamsConsensus {
                    network,
                    // TODO this is not very elegant, but I'm planning to get rid of it in a next
                    // commit anyway
                    finality_delay,
                    client_default_bitcoin_rpc: default_esplora_server(
                        bitcoin29_to_bitcoin30_network(network),
                    ),
                    fee_consensus: Default::default(),
                },
            },
        );
```

Because `bitcoind` is borked.

OMG, why do we need so many hacks?

AFAIK, we could fix that bitcoind backend by just importing address descriptor for every individual address, or something like that.

But also - is `client_default_bitcoin_rpc` supposed to be a thing, or is it just a hack? Is Federation supposed to tell the client what backend to use by default?

elsirion:
> But also - is `client_default_bitcoin_rpc` supposed to be a thing, or is it just a hack? Is Federation supposed to tell the client what backend to use by default?

It's a hack to avoid needing modularized, user-provided config for the client :sob: I'd say ideally integrators should just choose a sensible default or make it configurable.

PR #5116, opened 2024-04-26
refactor gateways send_payment_v2

joschisan:


m1sterc001guy:
I don't really have any context on what this is refactoring. Is it just moving logic that was previously in the gateway into the gateway's client module?

joschisan:
@m1sterc001guy yes. This ist just to factor it in a similar way as receiving payments is factored now.

elsirion (approved):

Issue #5119, opened 2024-04-26
tests: missing unit test in fedimint/fedimint-core/src/time.rs


sivasathyaseeelan:
### Description

Currently, unit test are missing in fedimint/fedimint-core/src/time.rs

### Use case

The proposed solution is to address this by adding the necessary tests for the files in the bytes package, which would help in detecting bugs early, supporting CI/CD, reducing debugging time, providing safety for this codebase etc...

PR #5120, opened 2024-04-26
tests: added unit test in fedimint/fedimint-core/src/time.rs

sivasathyaseeelan:
### Fix
This PR adds unit test fin fedimint/fedimint-core/src/time.rs and partially solves #5119


dpc (approved):

fedimint-core/src/time.rs
```diff
@@ -19,3 +19,35 @@ pub fn duration_since_epoch() -> std::time::Duration {
...
+        assert!(system_time.elapsed().is_ok(), "SystemTime should be valid");
+    }
+
+    #[test]
+    #[cfg(target_family = "wasm")]
+    fn test_now_on_wasm() {
+        let system_time = now();
+        assert_eq!(
+            system_time,
+            SystemTime::UNIX_EPOCH,
```
dpc, line 49: I don't think this is correct, so I approved the run already, but ...

dpc (changes requested):
I don't really see a point in these tests. These functions are called so many times in every test that something was wrong there, it would be immediately obvious.

I like unit tests that test something complex + fragile + important in isolation. Unit-tests are great, when possible to use it. But asserting that duration since epoch is positive ... is just not far from `assert_eq!(1 + 1, 2);` . Sure it better be true, but is there any chance it will ever actually catch this not being the case?

Unless there are some existing issues about missing tests, I think your time (and time of anyone willing to improve our testing posture), would be to go through higher level tests and look for some glaring gaps.

E.g. our passphrase based backup & recovery was very basic IIRC.

Maybe check the past issues and PRs for bugfixes and see if you were to revert the fix, would any test catch the problem and if not - add the missing test.

Another way - try introducing random mistake in some code (add +1, turn `<` into `>` or `<=`, etc.) and check if any tests catches it and if not think about why is it and what kind of test would do it.

sivasathyaseeelan:
> I don't really see a point in these tests. These functions are called so many times in every test that something was wrong there, it would be immediately obvious.
> 
> I like unit tests that test something complex + fragile + important in isolation. Unit-tests are great, when possible to use it. But asserting that duration since epoch is positive ... is just not far from `assert_eq!(1 + 1, 2);` . Sure it better be true, but is there any chance it will ever actually catch this not being the case?
> 
> Unless there are some existing issues about missing tests, I think your time (and time of anyone willing to improve our testing posture), would be to go through higher level tests and look for some glaring gaps.
> 
> E.g. our passphrase based backup & recovery was very basic IIRC.
> 
> Maybe check the past issues and PRs for bugfixes and see if you were to revert the fix, would any test catch the problem and if not - add the missing test.
> 
> Another way - try introducing random mistake in some code (add +1, turn `<` into `>` or `<=`, etc.) and check if any tests catches it and if not think about why is it and what kind of test would do it.

Thank you @dpc for your detailed feedback on the unit tests. I understand your point about the current tests not being as effective in catching potential issues, especially considering their simplicity.

I'll take your advice and shift my focus to higher-level tests, as you suggested. I'll look into areas like passphrase-based backup and recovery, as well as review past issues and PRs to identify any gaps that need addressing.

fedimint-core/src/time.rs
elsirion: Yep, this conflicts with the first test imo. Only one of both can succeed and for any sane implementation the first one should and this one should fail.

dpc:
I don't think any of these tests is worth bothering with. :D

PR #5127, opened 2024-04-26
tests: missing unit test in fedimint-core/src/encoding/tls.rs

### LINKED ISSUE #5125: tests: missing unit test in fedimint/fedimint-core/src/encoding/tls.rs
### Description

Currently, unit test are missing in fedimint/fedimint-core/src/encoding/tls.rs

### Use case

The proposed solution is to address this by adding the necessary tests for the files in the bytes package, which would help in detecting bugs early, supporting CI/CD, reducing debugging time, providing safety for this codebase etc...

sivasathyaseeelan:
### Fix
This PR adds unit test fin fedimint/fedimint-core/src/encoding/tls.rs and closes #5125

### DIFF
--- a/fedimint-core/src/encoding/tls.rs
+++ b/fedimint-core/src/encoding/tls.rs
@@ -30,6 +30,8 @@ mod tests {
         }
     }
 
+    /// Test the `consensus_encode` method for `rustls::Certificate`.
+    /// This method encodes a certificate into a consensus format.
     #[test]
     fn test_consensus_encode() {
         let test_cases = [
--- a/fedimint-core/src/encoding/tls.rs
+++ b/fedimint-core/src/encoding/tls.rs
@@ -10,3 +10,47 @@ impl Encodable for rustls::Certificate {
         self.0.encode_hex::<String>().consensus_encode(writer)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Cursor;
+
+    struct MockCertificate(Vec<u8>);
+
+    impl MockCertificate {
+        fn new(data: Vec<u8>) -> Self {
+            MockCertificate(data)
+        }
+    }
+
+    impl Encodable for MockCertificate {
+        fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
+            self.0.encode_hex::<String>().consensus_encode(writer)
+        }
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


fedimint-core/src/encoding/tls.rs
```diff
@@ -10,3 +10,47 @@ impl Encodable for rustls::Certificate {
...
+    }
+
+    impl Encodable for MockCertificate {
+        fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
+            self.0.encode_hex::<String>().consensus_encode(writer)
+        }
+    }
+
+    #[test]
+    fn test_consensus_encode() {
```
dpc, line 34: What exactly are we testing here? (Please add in a docstring comment as well)

Seems like testing encoding crate in an essence.

sivasathyaseeelan: hi @dpc, I am testing the implementation of the consensus_encode method for the rustls::Certificate type, which is part of the encoding crate.

m1sterc001guy: It would make more sense to me to test this by encoding it and then decoding it back to the struct and comparing it against the original struct.

sivasathyaseeelan: Thanks for your input @m1sterc001guy, that makes more sense. I'll definitely implement that approach to further validate the encoding and decoding processes!

fedimint-core/src/encoding/tls.rs
```diff
@@ -10,3 +10,49 @@ impl Encodable for rustls::Certificate {
...
+        fn new(data: Vec<u8>) -> Self {
+            MockCertificate(data)
+        }
+    }
+
+    impl Encodable for MockCertificate {
+        fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
+            self.0.encode_hex::<String>().consensus_encode(writer)
+        }
+    }
```
elsirion, line 32: 1. This encoding is horribly broken (first encodes to hex, then as a byte vector)
2. It doesn't test any actual code from the module it is in, you just implement some mock that is never used elsewhere and test that (it *kinda* tests some of the core encoding functions, but such a test should live elsewhere and wouldn't require the `MockCertificate`)

elsirion (changes requested):
I'd just close this PR, it's not adding anything of value as it is.

PR #5141, opened 2024-04-27
tests: added unit test in fedimint-core/src/runtime.rs

### LINKED ISSUE #5140: tests: missing unit test in fedimint/fedimint-core/src/runtime.rs
### Description

Currently, unit test are missing in fedimint/fedimint-core/src/runtime.rs

### Use case

The proposed solution is to address this by adding the necessary tests for the files in the bytes package, which would help in detecting bugs early, supporting CI/CD, reducing debugging time, providing safety for this codebase etc...

sivasathyaseeelan:
### Fix
This PR adds unit test fin fedimint/fedimint-core/src/runtime.rs and closes #5140

### DIFF
--- a/fedimint-core/src/runtime.rs
+++ b/fedimint-core/src/runtime.rs
@@ -219,12 +219,21 @@ mod tests {
     #[tokio::test]
     async fn test_timeout_completed() {
         let duration = Duration::from_millis(1);
-        let future = async {
+
+        let (tx, rx) = oneshot::channel();
+
+        let future = async move {
             tokio::time::sleep(duration / 2).await;
-            "done"
+            let _ = tx.send("done");
         };
 
-        let result = timeout(duration, future).await;
-        assert!(result.is_ok());
+        tokio::select! {
+            result = timeout(duration, future) => {
+                assert!(result.is_ok());
+            }
+            _ = rx => {
+                assert!(false, "Timeout occurred before future completed");
+            }
+        }
     }
 }
--- a/fedimint-core/src/runtime.rs
+++ b/fedimint-core/src/runtime.rs
@@ -163,3 +163,66 @@ mod r#impl {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tokio::sync::oneshot;
+
+    #[tokio::test]
+    async fn test_spawn_and_await() {
+        let name = "test_spawn_and_await";
+        let (tx, _rx) = oneshot::channel();
+
+        let future = async move {
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            let _ = tx.send("done");
+        };
+
+        let _handle = spawn(name, future).await;
+    }
+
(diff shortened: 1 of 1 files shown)


dpc (changes requested):
Please don't use `sleep`, or at least not with 1 second values. CI time is a scarce resource. Where you really need to guarantee sleep, maybe use 1 millisecond. where you need coordination between threads use https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html to unblock things.

sivasathyaseeelan:
> Please don't use `sleep`, or at least not with 1 second values. CI time is a scarce resource. Where you really need to guarantee sleep, maybe use 1 millisecond. where you need coordination between threads use https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html to unblock things.

Hi @dpc, I have changed from 1 second to 1 millisecond. Can you please review this pr again?

fedimint-core/src/runtime.rs
```diff
@@ -216,7 +218,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_timeout_completed() {
-        let duration = Duration::from_secs(1);
+        let duration = Duration::from_millis(1);
```
dpc, line 221: This will be flaky. You can not rely on one thing finishing before the other. With 1s duration, it would probably work reliably, but technically still be incorrect.

If you need to guarantee that one thing finished / or not before the other blocking on a receiver of a oneshot channel, and sending message to unblock it from the other side is the way to go.

fedimint-core/src/runtime.rs
```diff
@@ -163,3 +163,77 @@ mod r#impl {
...
+        };
+
+        let _handle = spawn(name, future).await;
+    }
+
+    #[tokio::test]
+    async fn test_sleep() {
+        let duration = Duration::from_millis(1);
+        let start = Instant::now();
+        sleep(duration).await;
```
dpc, line 191: Meh. This might be a test in `tokio`, it doesn't test our code really.

fedimint-core/src/runtime.rs
```diff
@@ -163,3 +163,77 @@ mod r#impl {
...
+
+    #[tokio::test]
+    async fn test_spawn_and_await() {
+        let name = "test_spawn_and_await";
+        let (tx, _rx) = oneshot::channel();
+
+        let future = async move {
+            // Simulating some work
+            tokio::time::sleep(Duration::from_millis(1)).await;
+            let _ = tx.send("done");
```
dpc, line 181: I don't understand what this test is trying to test.

But probably the `future` should be only `rx.recv()`;

Then `spawn` that future as a task, then  `tx.send()` to unblock it, then `await` on the result.

maan2003 (changes requested):
native code for runtime module is mostly uninteresting.

wasm code has some logic, that might be good to test.

dpc:
Closing. Feel free to reopen with improvements or just create a different PR.

Issue #5139, opened 2024-04-27
CI flake: empty chan info


dpc:
I was able to repro the issue by running:

```
> retry-success.sh ./scripts/tests/test-ci-all-backcompat.sh v0.2.1
> cat (which retry-success.sh)
#!/usr/bin/env bash

while true : ; do "$@" || exit 0 ; sleep 1 ; done
```

3 times in parallel, and waiting some time.

The issue seems to be:

```
00:04:26 2024-04-27T17:36:22.874154Z DEBUG spawn{task_name="JitTry alloc::sync::Arc<()> value"}: devimint::external: Empty chan info edge=ChannelEdge { channel_id: 123145302376449, chan_point: "33ded66d14
f9b81bf2d1bea723aa0b4592b8718d6da7f605de625e67c07ae4b4:1", last_update: 1714239342, node1_pub: "032374ba841a546b8f02d9f937a1670a11f16e3cc49ed245274f41fb306048d4ea", node2_pub: "0338859add9c9381cb710423c74
2ae083b517722c93023a64a425da24be08414c9", capacity: 10000000, node1_policy: None, node2_policy: Some(RoutingPolicy { time_lock_delta: 80, min_htlc: 1000, fee_base_msat: 1000, fee_rate_milli_msat: 1, disab
led: false, max_htlc_msat: 10000000000, last_update: 1714239342 }) }
```

which we've seen before.

@fedimint/lightning what can we do about, I remember this issue before, and seems still to be there?

justinmoon:
dev call: looks like lnd bug. perhaps we should bring it up with them?

elsirion:
We probably have a workaround now judging by the issue history and no further comments.

PR #5142, opened 2024-04-28
fix(devimint): reset chan policy on retry

dpc:
Maybe a fix to #5139 .


bradleystachurski:
review club: moved chan policy to lnd.conf

PR #5144, opened 2024-04-28
fix(test-ci-all): compile nix binaries of versions outside of test

dpc:
`nix build` can potentially hang for a long time, even if the binary was previous built, e.g. because Nix daemon is busy with other stuff and its database is locked.

This could previous potentially cause the test to spend a large part of it's time trying to build binaries.

Instead, prebuild all binaries once before `parallel` is started, and store the resulting path into matching env variable. During tests just take the path out of variable.

### DIFF
--- a/scripts/tests/test-ci-all.sh
+++ b/scripts/tests/test-ci-all.sh
@@ -182,7 +182,16 @@ if [[ "$num_versions" == "0" ]]; then
 else
   # precompile binaries
   binaries=( "fedimintd" "fedimint-cli" "gateway-cli" "gatewayd" )
-  parallel nix_build_binary_for_version "{1}" "{2}" ::: "${binaries[@]}" ::: "${tagged_versions[@]}"
+  for version in "${versions[@]}" ; do
+    if [ "$version" == "current" ] ; then
+      continue
+    fi
+    for binary in "${binaries[@]}" ; do
+      var_name=$(nix_binary_version_var_name "$binary" "$version")
+      export "${var_name}=$(nix_build_binary_for_version "$binary" "$version")"
+    done
+  done
+
   if [ -n "${FM_FULL_VERSION_MATRIX:-}" ]; then
     mapfile -t version_matrix < <(generate_full_matrix "${versions[@]}")
   else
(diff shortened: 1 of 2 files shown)


scripts/tests/test-ci-all.sh
```diff
@@ -182,7 +182,16 @@ if [[ "$num_versions" == "0" ]]; then
 else
   # precompile binaries
   binaries=( "fedimintd" "fedimint-cli" "gateway-cli" "gatewayd" )
-  parallel nix_build_binary_for_version "{1}" "{2}" ::: "${binaries[@]}" ::: "${tagged_versions[@]}"
+  for version in "${versions[@]}" ; do
+    if [ "$version" == "current" ] ; then
```
bradleystachurski, line 186: [nit] Can alternatively use `tagged_versions` and skip this check

scripts/tests/test-ci-all.sh
```diff
@@ -182,7 +182,16 @@ if [[ "$num_versions" == "0" ]]; then
 else
   # precompile binaries
   binaries=( "fedimintd" "fedimint-cli" "gateway-cli" "gatewayd" )
-  parallel nix_build_binary_for_version "{1}" "{2}" ::: "${binaries[@]}" ::: "${tagged_versions[@]}"
```
bradleystachurski, line 185: Out of curiosity, did you compare the performance of removing `parallel`? It might be faster in sequence due to resource contention (originally flagged by @elsirion in https://github.com/fedimint/fedimint/pull/4413#discussion_r1501815988).

bradleystachurski (approved):

scripts/tests/test-ci-all.sh
dpc: If `nix build` is not contended all of this is rather instant.

bradleystachurski: Yep, that makes sense. I was curious about the initial build that we'll do for all versions (https://github.com/fedimint/fedimint/pull/5144/files#diff-e5c7cc37f7292b713d50077a8badf27398452732e2042c0bb9a57b4d8346fa81R191)

PR #5146, opened 2024-04-28
fix(client): current_consensus strategy should require threshold not all

dpc:
Or am i just confused?


justinmoon:
@joschisan

joschisan:
> Or am i just confused?

You are confused, the constructor takes the total value and calculates the threshold internally.

PR #5148, opened 2024-04-28
fix(ln-gateway): don't hang on `info` if there's no channels

dpc:
`info` is used in `devimint` to check if the gateway is ready to respond to API calls, at which point there's absolutely no chance the channel was created yet.


dpc:
Landed as a part of a different PR.

Issue #4875, opened 2024-04-08
RPC is very slow when we are connected to a LND without channels


douglaz:
If we have stuff like:
```
INFO info: ln_gateway: LN node returned no route hints, trying again in 2s num_retries=23
```
Because LND is still not completely up (zero channels)

Then any RPC call, like `gateway-cli info` will take at least 30s to complete.

m1sterc001guy:
What do you think the recommended behavior should be? If there's no channels, the gateway isn't really usable, for outgoing or incoming payments. Maybe we should wait until a channel is setup before setting the gateway's state to `Running`

douglaz:
The response time should not change. If the operation is invalid under that circumstance then it should answer with a proper error.

elsirion:
I don't know if it should crash just because LND isn't ready yet, that seems to be a philosophical question (and @dpc tweeted about that recently iirc). But what we can certainly do is not block if we don't want to give extra route hints anyway. In that case we don't need to fetch them.

elsirion:
Empty hints are very bad for unknown LN nodes/with only private channels. The GW won't be able to receive. That's why we retry for 30s or so before giving up.

dpc:
If it's so important, by the time some calls `fedimint-cli info` GW should already have that info then in a cached value or something, I guess.

elsirion:
> Are you sure it's "any call", not just `info`? Because `fetch_lightning_route_hints` that seems to be what's blocking is called only in `handle_get_info` and `handle_connect_federation` making this somewhat less severe.

Oh, I didn't see your reply when posting mine (race conditions …) I don't think `info` needs to fetch this, ideally it would only return what the GW already knows/used for the last registration (idk how hard that would be to impl). `handle_connect_federation` needs it though imo for the first registration (unless we want to make that async).

dpc:
ARGGHGHGHGHGH!!!

This is absolutely terrible and source of so many bugs...

m1sterc001guy:
Fixed with https://github.com/fedimint/fedimint/pull/5166

PR #5149, opened 2024-04-28
chore(devimint): fix most of flakiness in setting up LN gateways

dpc:
There's a complicated web of initialization dependencies in areas of LN that leads to flakes.

Most importantly `devimint` expects `gateway-cli info` to tell if gatewayd started. This doesn't currently work well, because it will hang for 60s or until gateway gets a channel opened. But then in a "new way" of opening channel doing so can't happen until gateway is ready to respond to API calls.

And then 60s of timeout is as much as other steps in `devimint` are actually polling some stuff. Some then sometimes it works, sometimes it doesn't.

Then `fedimint-cli connect-fed` also has the same behavior of waiting for a route hint from the channel. I don't change that, because it would probably spook many places in the tests that expect the hint to be there.

But generally currently it's all messed up to the point where it's hard to keep track of write a reliable code that would setup up devimint with gateways.

#4875 needs to get fixed, it's a significant architectural issue.

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -689,10 +689,17 @@ pub async fn open_channel_between_gateways(
 ) -> Result<()> {
     let gateway_cli_version = crate::util::GatewayCli::version_or_default().await;
     let gatewayd_version = crate::util::Gatewayd::version_or_default().await;
-    if gateway_cli_version < *VERSION_0_4_0_ALPHA || gatewayd_version < *VERSION_0_4_0_ALPHA {
+    #[allow(clippy::overly_complex_bool_expr)]
+    if gateway_cli_version < *VERSION_0_4_0_ALPHA || gatewayd_version < *VERSION_0_4_0_ALPHA
+        // FIXME: The new way is actually slower and require more things to initialize, and
+        // due to more dependency chains lead to flaky CI pipeline
+        || true
+    {
+        debug!(target: LOG_DEVIMINT, "Opening channel between gateways (the old way)");
         return open_channel(process_mgr, bitcoind, cln, lnd).await;
     }
 
+    debug!(target: LOG_DEVIMINT, "Opening channel between gateways (the new way)");
     lnd.client
         .lock()
         .await
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -554,6 +554,22 @@ async fn open_channel(
     cln: &Lightningd,
     lnd: &Lnd,
 ) -> Result<()> {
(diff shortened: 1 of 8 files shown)


dpc:
@fedimint/lightning for awarness

dpc:
Before:

image


After:

image

fedimint-api-client/src/api.rs
```diff
@@ -1181,7 +1181,7 @@ impl FederationPeerClientShared {
     /// Wait (if needed) + update reconnection stats
     async fn wait_and_inc_reconnect(&mut self) {
         self.wait().await;
-        self.connection_attempts += 1;
+        self.connection_attempts += self.connection_attempts.saturating_add(1);
```
elsirion, line 1184: bug: you are adding twice now, this likely goes exponential?

dpc: Good catch. Thank you.

elsirion (approved):

devimint/src/external.rs
```diff
@@ -554,6 +554,22 @@ async fn open_channel(
     cln: &Lightningd,
     lnd: &Lnd,
 ) -> Result<()> {
+    debug!(target: LOG_DEVIMINT, "Await block ln nodes block processing");
```
m1sterc001guy, line 557: Was this refactor needed (moving this block from below) to make things faster or is it just ergonomic?

dpc: This exact like is just a casual shaving-off couple of milliseconds only because it was easy.

devimint/src/external.rs
```diff
@@ -683,16 +684,23 @@ pub async fn open_channel_between_gateways(
...
+    gw_lnd: &JitTryAnyhow<Arc<Gatewayd>>,
 ) -> Result<()> {
     let gateway_cli_version = crate::util::GatewayCli::version_or_default().await;
     let gatewayd_version = crate::util::Gatewayd::version_or_default().await;
-    if gateway_cli_version < *VERSION_0_4_0_ALPHA || gatewayd_version < *VERSION_0_4_0_ALPHA {
+    #[allow(clippy::overly_complex_bool_expr)]
+    if gateway_cli_version < *VERSION_0_4_0_ALPHA || gatewayd_version < *VERSION_0_4_0_ALPHA
+        // FIXME: The new way is actually slower and require more things to initialize, and
+        // due to more dependency chains lead to flaky CI pipeline
+        || true
```
m1sterc001guy, line 697: @tvolk131 FYI

PR #5157, opened 2024-04-29
cleanup wallet sync logic

joschisan:
The consensus logic is not actually changed by this pr, as None and Some(0) returned by consensus_block_count do not trigger a sync, so we might as well just replace any None with a 0 to begin with.

### DIFF
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -5,7 +5,7 @@ use std::convert::Infallible;
 #[cfg(not(target_family = "wasm"))]
 use std::time::Duration;
 
-use anyhow::{bail, format_err, Context};
+use anyhow::{bail, ensure, format_err, Context};
 use bitcoin::hashes::{sha256, Hash as BitcoinHash, HashEngine, Hmac, HmacEngine};
 use bitcoin::policy::DEFAULT_MIN_RELAY_TX_FEE;
 use bitcoin::secp256k1::{All, Secp256k1, Verification};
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -986,7 +981,7 @@ impl Wallet {
             "New consensus height, syncing up",
         );
 
-        for height in (old_height + 1)..=(new_height) {
```
dpc, line 989: Bleh. :D

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -986,7 +981,7 @@ impl Wallet {
...
+        for height in old_height..new_height {
```
dpc, line 984: :+1:

dpc (approved):
I'm a bit scared of consensus refactoring, but seems OK to me  and existing version is really meh.

modules/fedimint-wallet-server/src/lib.rs
dpc: I think initially this code was operating on "block count" like `bitcoind`s `getblockcount` rpc, which is actually not a block count, but the height of the highest block (so inclusive range), which is probably the dumbest part of `bitcoind`, right before reversting txids.

Then I went on mission to change everything to use real `block_count`, but this code might have got transformed mechanically to match previous behavior. Wouldn't surprise me if I did this it. :D

modules/fedimint-wallet-server/src/lib.rs
dpc: Though actually... I would rename `s/height/count` everywhere here.

elsirion (approved):
Agree with @dpc's height->count rename suggestion, but structurally the PR is already a big improvement.

PR #5158, opened 2024-04-29
feat: federated lightning address

joschisan:


justinmoon:
dev call: similar to hermes, but receiver needs to know what contracts to look for. bad performance characteristic??? performance of feature is usually independent of usage of the feature ...

PR #5164, opened 2024-04-29
chore: split backend test for bitcoind

dpc:
It has a bit long runtime, and each module can just run in parallel to bring it down.

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -55,7 +55,7 @@ function ensure_in_dev_shell() {
 function make_fm_test_marker() {
   if [ -n "${FM_TEST_NAME:-}" ]; then
     # make it easy to identify which tmp dir belongs to which test
-    touch "${TMP:-/tmp}-$(echo "$FM_TEST_NAME" |tr -cd '[:alnum:]-_')" || true
+    touch "${TMP:-/tmp}/$(echo "$FM_TEST_NAME" |tr -cd '[:alnum:]-_')" || true
   fi
 }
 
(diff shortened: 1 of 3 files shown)


dpc:
@m1sterc001guy Seems like `env FM_TEST_ONLY=bitcoind-ln-gateway` is actually the slowest one there. If we could figure out a way to split it in 2 or 3, that would be a win.

dpc:
@m1sterc001guy i've split it in half

bradleystachurski:
@dpc Did you have a chance to run locally? Bumping into a familiar error starting lightningd since the path is too long.

```
00:00:00 2024-04-30T05:50:36.867177Z  INFO fm::devimint: Devimint data dir path=/tmp/nix-shell.qfYw9L/backend_test_ln_gateway_not_client-9AF7/devimint-2801155-576
00:01:01 Error: Polling lightningd failed after 111 retries (timeout: 60s)
00:01:01
00:01:01 Caused by:
00:01:01     0: connect to lightningd
00:01:01     1: path must be shorter than libc::sockaddr_un.sun_path
```

dpc:
@bradleystachurski I don't think that's the case anymore in the latest version I pushed 3h ago.

bradleystachurski:
I'm getting that error against the latest commit

```
nix@lenovo-m75q-ubuntu:~/fedimint$ git rev-parse HEAD
773ff57051aa0b38e5b89699ce8b92e58090ee03
```

Using a shorter test name for `backend_test_ln_gateway_not_client` passes locally.

bradleystachurski (approved):

Issue #5126, opened 2024-04-26
peg_outs_must_wait_for_available_utxos flake `Unexpected state WaitingForConfirmations`


dpc:
```
2024-04-26T21:02:11.020873Z  INFO fedimint_wallet_tests: Peg-in transaction mined height=11 tx=Transaction { version: 0, lock_time: PackedLockTime(0), input: [], output: [TxOut { value: 6000, script_pubkey: Script(OP_0 OP_PUSHBYTES_32 fd4e0c4c7ea6e4185036bea5f45313c6b2f485fac75111adb33e0e4d0e47866f) }] }
2024-04-26T21:02:11.022853Z DEBUG task{name="state machine executor"}:state_machine_transition{operation_id=89769494695e2d60bbf8247117896c6ba2260b6d0741c6877722eb9aa1e9b356}: timing: Operation timing name=autocommit - commit_tx duration_ms=0
thread 'peg_outs_must_wait_for_available_utxos' panicked at /build/source/modules/fedimint-wallet-client/src/lib.rs:502:29:
Unexpected state WaitingForConfirmations(WaitingForConfirmationsDepositState { tweak_key: KeyPair(#78eaae26423b7463), btc_transaction: Transaction { version: 0, lock_time: PackedLockTime(0), input: [], output: [TxOut { value: 6000, script_pubkey: Script(OP_0 OP_PUSHBYTES_32 fd4e0c4c7ea6e4185036bea5f45313c6b2f485fac75111adb33e0e4d0e47866f) }] }, out_idx: 0 })
stack backtrace:
   0: rust_begin_unwind
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/panicking.rs:72:14
   2: {async_block#0}
             at /build/source/modules/fedimint-wallet-client/src/lib.rs:502:29
   3: <async_stream::async_stream::AsyncStream<fedimint_wallet_client::DepositState, <fedimint_wallet_client::WalletClientModule>::subscribe_deposit_updates::{closure#0}::{closure#1}::{closure#0}> as futures_core::stream::Stream>::poll_next
[...]

PR #5166, opened 2024-04-29
chore: never wait for fetch_lightning_route_hints

m1sterc001guy:
Follow up to https://github.com/fedimint/fedimint/pull/5149

`fetch_lightning_route_hints` currently loops and waits for the route hints to exist. This is not necessary, the client is already adding a route hint for the gateway's lightning node.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -71,7 +71,7 @@ use fedimint_wallet_client::{
 };
 use futures::stream::StreamExt;
 use gateway_lnrpc::intercept_htlc_response::Action;
-use gateway_lnrpc::{GetNodeInfoResponse, InterceptHtlcResponse};
+use gateway_lnrpc::{GetNodeInfoResponse, GetRouteHintsResponse, InterceptHtlcResponse};
 use hex::ToHex;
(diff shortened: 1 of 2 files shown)


m1sterc001guy:
Disabling the test that I think should be refactored to a devimint test https://github.com/fedimint/fedimint/issues/5168

gateway/ln-gateway/src/lib.rs
```diff
@@ -993,7 +985,8 @@ impl Gateway {
             client
                 .get_first_module::<GatewayClientModule>()
                 .register_with_federation(
-                    route_hints,
+                    // Route hints will be updated in the background
+                    Vec::new(),
```
dpc, line 989: Good enough, but if this overwrites the existing registration, would be better to not overwrite with empty value, even if just for a moment, as it might lead to flakiness (mostly worried about prod).

m1sterc001guy: I think that would be difficult to do, since we check if the federation is already connected or not in the code above.

dpc:
Unrelated flake #5126 fail in MQ. Adding back.

gateway/ln-gateway/src/lib.rs
dpc: Oh, OK. Nvmd then.

PR #5169, opened 2024-04-29
chore: don't set up channel policy

dpc:
There's some LND bug that makes it flacky, and turns out we are only doing it to make 1msat payement work, which we don't even use in our test suite anyway.


m1sterc001guy (dismissed):
If it passes the tests we have, LGTM

I think we can fix the 1 msat payments by changing this in the `lnd.conf` file https://github.com/lightningnetwork/lnd/blob/3d56ffd0e486fc43a14da4a1600ece86846ff81e/sample-lnd.conf#L631

m1sterc001guy (approved):

PR #5175, opened 2024-04-30
chore: put tighther deadlines on test runtime

dpc:
Will rebase after #5176 lands

This is to prevent silent regressions.

I wish we could go lower. Most devimint-based tests are already below 40s, and unit-test like tests should be OK under 20s, IMO.

We'll get there.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,6 +1,6 @@
 [profile.default]
 # no matter the profile, we want to kill tests that hanged on something
-slow-timeout = { period = "60s", terminate-after = 8 }
+slow-timeout = { period = "30s", terminate-after = 4 }
 
 # define `dev` to allow running cargo and nextest with the same profile name
 # inherits default profile config
@@ -9,4 +9,4 @@ slow-timeout = { period = "60s", terminate-after = 8 }
 [profile.ci]
 fail-fast = true
 failure-output = "immediate"
-slow-timeout = { period = "60s", terminate-after = 8 }
+slow-timeout = { period = "30s", terminate-after = 2 }
--- a/scripts/dev/run-test/fm-run-test
+++ b/scripts/dev/run-test/fm-run-test
@@ -37,7 +37,7 @@ on_exit() {
 trap on_error ERR
 trap on_exit EXIT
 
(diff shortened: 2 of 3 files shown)


bradleystachurski (approved):

bradleystachurski:
```
TIMEOUT [  60.007s] fedimint-wallet-tests::fedimint_wallet_tests peg_out_fail_refund
```

Perhaps a bit too tight?

bradleystachurski:
Timing out now with the updated `parallel` 130s timeout

```
       > 36    :       1714530498.663     130.344      0       28      -1      15      run_test_for_versions bckn_gw_not_client FM: current CLI: current GW: current
       >
       > 63      :       1714530587.481     130.960      0       28      -1      15      run_test_for_versions bckn_gw_not_client FM: current CLI: current GW: current
```

Perhaps we can resolve that one by breaking up the gateway tests into smaller chunks?

dpc:
> Perhaps we can resolve that one by breaking up the gateway tests into smaller chunks?

I already did and split it in two.

It's OK. I'll increase to 150.

bradleystachurski (dismissed):

maan2003:
IMO time limit should be defined per test

dpc:
> IMO time limit should be defined per test

Could be done as an additional one, but seems like somewhat manual thing to maintain.

dpc:
:crossed_fingers: This change is cursed and keeps triggering new unseen yet flakes.

bradleystachurski (dismissed):

bradleystachurski:
> This change is cursed

I'm becoming concerned that if we land this, subsequent PRs will have to get lucky to pass the merge queue. Should we hold off?

dpc:
> I'm becoming concerned that if we land this, subsequent PRs will have to get lucky to pass the merge queue. Should we hold off?

If we're hitting tests legitimately taking more time the deadlines we should raise them. But if it's unrelated flakes (which seems to have happened a lot here), holding off doesn't help.

dpc:
Last failure might be fixed by #5179 so will re-add after it lands.

bradleystachurski (approved):

Issue #5168, opened 2024-04-29
Lightning Test Improvements


m1sterc001guy:
With the development of the gateway (and devimint in parallel) our lightning tests have grown and morphed to a point where they are now difficult to work with. This issue describes the current issues and has a proposal for how to fix these tests going forward.

## Issues
 - There are multiple different test suites: `fedimint-ln-tests`, the gateway's `integration_tests.rs`, and tests that exist in `devimint`. When adding new tests its unclear where they should go.
 - LN Tests have a "fake" and a "real" mode. In the "real" mode, they connect to external daemons that are spawned by devimint. This is difficult to debug, since it requires setting an environment variable that new contributors might not know. It also hampers parallelism, since all rusts tests that run in parallel use the same backing daemon.
 - The gateway tests use an ugly hack for verifying the state machine transitions. The test first removes the gateway's federation client so that REST requests to the gateway fail, but the necessary data (i.e the contract) is created correctly. The results in confusing error messages in the logs, hacky code, etc.
 - The tests iterate over each gateway sequentially. This makes the tests much slower than normal tests, and because of the above issue, they can't easily be parallelized.

## Proposal
At the time most of this test suite was created, we did not have devimint tests like we do today. I believe most of the LN tests should be refactored into standalone devimint tests. These can run in parallel much easier since they use their own daemons.

`fedimint-ln-tests`: should only contain tests verifying state machine transitions of the LN client. All necessary dependencies should be mocked (i.e gateway)
`integration_tests.rs`: should only contain tests verifying state machine transitions of the gateway client. All necessary dependencies should be mocked (i.e LN node)

All other tests should be refactored to devimint tests.

`GatewayTest` should be deleted. `fedimint-testing/src/ln/real.rs` should be deleted.

## Test Breakdown

`fedimint-ln-tests`
- [x] `text_can_attach_extra_meta_to_receive_operation` - good
- [x] `cannot_pay_same_internal_invoice_twice` - good
- [x] `gateway_protects_preimage_for_payment` - should be devimint test
- [x] `cannot_pay_same_external_invoice_twice` - gateway needs to be mocked
- [x] `makes_internal_payments_within_federation` - gateway needs to be mocked
- [x] `can_receive_for_other_user` - gateway needs to be mocked
- [x] `can_receive_for_other_user_tweaked` - gateway needs to be mocked
- [x] `rejects_wrong_network_invoice` - gateway needs to be mocked

`integration_tests.rs`
[...]

PR #5179, opened 2024-05-01
chore: Mock RPCs between Client -> Gateway

m1sterc001guy:
Working towards: https://github.com/fedimint/fedimint/issues/5168

Gets rid of the ugly hack we used of removing the gateway's clients during the tests.

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -527,7 +527,7 @@ impl FedimintCli {
     }
 
     pub fn with_default_modules(self) -> Self {
-        self.with_module(LightningClientInit)
+        self.with_module(LightningClientInit::default())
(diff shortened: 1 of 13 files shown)


m1sterc001guy:
@dpc fixes one of the sores of the lightning tests, next step is to convert the other tests to devimint tests and then rip out the real LN part

dpc (approved):

dpc:
I think I've just seen a flake caused by the problem this fixes, so hitting the land button and adding the label to get more reviews later.

dpc:
Unrelated flake we are already working on a fix in #5182

elsirion:
Great work! We should strive to be able to mock out all network connection so that we can test state machines in module-local tests that don't need a whole devimint setup!

PR #5187, opened 2024-05-01
fix(ci): increase the poll time in sends_ecash_out_of_band_cancel

dpc:
I've seen a weird fail that might, or might have not been a timeout here. Seems we standardize on 60s in other places, so lets increase it here as well.


bradleystachurski (approved):

bradleystachurski:
Was about to make a similar change 😄 

Original failure for context

```
fedimint-workspace-lcov-ci> thread 'sends_ecash_out_of_band_cancel' panicked at modules/fedimint-mint-tests/tests/tests.rs:262:5:
```

https://github.com/fedimint/fedimint/actions/runs/8912020657/job/24474569251#step:8:1855

PR #5188, opened 2024-05-01
chore(nix-module): improvements

dpc:
* Ability to source the bitcoin rpc secret from a file (useful if you're using nix-bitcoin, or agenix, etc.)
* Ability to set user & group for the daemon.
* Cleanup.

Corresponding configuration.nix change (when using `nix-bitcoin`):

```
+  users.extraUsers.fedimintd-somefed.extraGroups = [ "bitcoinrpc-public" ];

   services.fedimintd."somefed" = {
     enable = true;
     package = pkgs.fedimintd;
+    extraEnvironment = {
+      "RUST_LOG" = "info";
+      "RUST_BACKTRACE" = "1";
+    };
     api = {
       address = "wss://${fmApiFqdn}/ws/";
       bind = "127.0.0.1";
@@ -133,7 +126,8 @@ in
     bitcoin = {
       network = "bitcoin";
       rpc = {
-        address = "http://bitcoin:${btcPass}@127.0.0.1:8332";
+        address = "http://bitcoin@127.0.0.1:8332";
+        secretFile = "/etc/nix-bitcoin-secrets/bitcoin-rpcpassword-public";
       };
     };
   };
```

### DIFF
--- a/nix/modules/fedimintd.nix
+++ b/nix/modules/fedimintd.nix
@@ -240,17 +240,20 @@ in
                 User = cfg.user;
                 Group = cfg.group;
 
+                Restart = "always";
+                RestartSec = 10;
+                StartLimitBurst = 5;
+                UMask = "077";
+                LimitNOFILE = "100000";
+
                 LockPersonality = true;
-                MemoryDenyWriteExecute = true;
                 ProtectClock = true;
                 ProtectControlGroups = true;
                 ProtectHostname = true;
                 ProtectKernelLogs = true;
                 ProtectKernelModules = true;
                 ProtectKernelTunables = true;
-                PrivateDevices = true;
                 PrivateMounts = true;
-                PrivateUsers = true;
                 RestrictAddressFamilies = [ "AF_INET" "AF_INET6" ];
                 RestrictNamespaces = true;
                 RestrictRealtime = true;
@@ -262,11 +265,14 @@ in
                 StateDirectory = "fedimintd";
                 StateDirectoryMode = "0700";
                 ExecStart = startScript;
-                Restart = "always";
-                RestartSec = 10;
-                StartLimitBurst = 5;
-                UMask = "077";
-                LimitNOFILE = "100000";
+
+
+                # Hardening measures
+                PrivateTmp = "true";
+                ProtectSystem = "full";
+                NoNewPrivileges = "true";
+                PrivateDevices = "true";
+                MemoryDenyWriteExecute = "true";
               };
             }
           )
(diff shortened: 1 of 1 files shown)


nix/modules/fedimintd.nix
```diff
@@ -110,11 +141,6 @@ let
           and is set to be read only.
         '';
       };
-      rustLogEnv = mkOption {
-        type = types.str;
-        default = "info,fedimint_server::request=debug,fedimint_client::request=debug";
-        description = "Value to set RUST_LOG to";
-      };
```
elsirion, line 117: Should we make this a default env var?

nix/modules/fedimintd.nix
```diff
@@ -173,60 +199,102 @@ in
...
+      mapAttrs'
+        (fedimintdName: cfg: (
+          nameValuePair "fedimintd-${fedimintdName}" (
+            let
+              startScript = pkgs.writeShellScript "fedimintd-start" (
+                (if cfg.bitcoin.rpc.secretFile != null then
+                  ''
+                    secret=$(${pkgs.coreutils}/bin/head -n 1 "${cfg.bitcoin.rpc.secretFile}")
+                    prefix="''${FM_BITCOIN_RPC_URL%*@*}"  # Everything before the last '@'
+                    suffix="''${FM_BITCOIN_RPC_URL##*@}"  # Everything after the last '@'
```
elsirion, line 212: TIL

nix/modules/fedimintd.nix
```diff
@@ -173,60 +199,102 @@ in
...
-    systemd.services = mapAttrs'
+    systemd.services =
+      mapAttrs'
+        (fedimintdName: cfg: (
+          nameValuePair "fedimintd-${fedimintdName}" (
+            let
+              startScript = pkgs.writeShellScript "fedimintd-start" (
+                (if cfg.bitcoin.rpc.secretFile != null then
+                  ''
+                    secret=$(${pkgs.coreutils}/bin/head -n 1 "${cfg.bitcoin.rpc.secretFile}")
```
elsirion, line 210: Which user does this start script run as? Do we need to worry about permissions? Would be nice to document if so.

elsirion (approved):
Looks good, let's use it in practice and iterate :)

elsirion:
Although, since this is to be backported I'd like to avoid churn. @dpc do you want to change anything?

nix/modules/fedimintd.nix
dpc: I kind of think that default should be the default (which is `info`). Might be an opportunity to tweak the logging statements itself to be better at default level.

nix/modules/fedimintd.nix
dpc: It runs as the unit's user and group which is configurable and default to `fedimint-<name>`

dpc:
I hope no changes will be needed (some changes probably will needed anyway :D).

PR #5189, opened 2024-05-01
fix(server): panic when there are no peers

dpc:

### DIFF
--- a/fedimint-server/src/net/peers.rs
+++ b/fedimint-server/src/net/peers.rs
@@ -341,7 +341,12 @@ where
     }
 
     async fn receive(&mut self) -> Cancellable<(PeerId, T)> {
-        // TODO: optimize, don't throw away remaining futures
+        // if all peers banned (or just solo-federation), just hang here as there's
+        // never going to be any message. This avoids panic on `select_all` with
+        // no futures.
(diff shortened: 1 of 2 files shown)


bradleystachurski (commented):
Should we also add here?

https://github.com/fedimint/fedimint/blob/5bd29c2d8c49cee33a5c3b9ff4bddcedb1d7eff4/fedimint-server/src/net/peers.rs#L343

dpc:
> Should we also add here?
> 
> https://github.com/fedimint/fedimint/blob/5bd29c2d8c49cee33a5c3b9ff4bddcedb1d7eff4/fedimint-server/src/net/peers.rs#L343

Most certainly. @bradleystachurski . Done.

But ... why do we have such a copypasta here ... ?

bradleystachurski:
> But ... why do we have such a copypasta here ... ?

I was wondering the same thing, but was hoping you had context :smile:

dpc:
> I was wondering the same thing, but was hoping you had context 😄

I'm just the janitor, sir.

fedimint-backports:
- #5191

bradleystachurski:
review club: requires more investigation to know if we can unify the different networking paths

PR #5194, opened 2024-05-02
feat: enable shutdown after certain session index for a coordinated consensus upgrade

joschisan:
The intended shutdown procedure for this is that all guardians select a session to shutdown, let’s say via a group chat, that is sufficiently long in the future for all of them to set the index to shutdown at, so let’s say something like 15minutes. Then, when the guardians have shut down, every guardian needs to check that it has shut down at the at the intended index and confirm with each other via chat; this may not be the case of a guardian is lagging behind sufficiently. If not, retry. Only if successful for every guardian we can initialise the upgrade.

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -75,8 +75,8 @@ pub struct ConsensusServer {
     api_endpoints: Vec<(PeerId, SafeUrl)>,
     cfg: ServerConfig,
     submission_receiver: Receiver<ConsensusItem>,
+    shutdown_receiver: watch::Receiver<Option<u64>>,
     latest_contribution_by_peer: Arc<RwLock<LatestContributionByPeer>>,
-
     /// Just a string version of `cfg.local.identity` for performance
     self_id_str: String,
     /// Just a string version of peer ids for performance
@@ -176,6 +176,7 @@ impl ConsensusServer {
         );
 
         let (submission_sender, submission_receiver) = async_channel::bounded(TRANSACTION_BUFFER);
+        let (shutdown_sender, shutdown_receiver) = watch::channel(None);
 
         // Build P2P connections for the atomic broadcast
         let (connections, peer_status_channels) = ReconnectPeerConnections::new(
@@ -195,6 +196,7 @@ impl ConsensusServer {
             modules: modules.clone(),
             client_cfg: cfg.consensus.to_client_config(&module_inits)?,
             submission_sender: submission_sender.clone(),
+            shutdown_sender,
             supported_api_versions: ServerConfig::supported_api_versions_summary(
                 &cfg.consensus.modules,
                 &module_inits,
@@ -235,6 +237,7 @@ impl ConsensusServer {
                 .collect(),
             cfg: cfg.clone(),
             submission_receiver,
+            shutdown_receiver,
             latest_contribution_by_peer,
             modules,
         };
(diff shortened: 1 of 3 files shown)


elsirion:
Would be great to have a devimint test that checks that
1. The signed session count in the DB is as expected for all peers
2. That the current-session AlephBFT DB table is empty

fedimint-server/src/consensus/server.rs
```diff
@@ -301,8 +305,7 @@ impl ConsensusServer {
 
             info!(target: LOG_CONSENSUS, "Session {session_index} completed");
 
-            // if the submission channel is closed we are shutting down
-            if self.submission_receiver.is_closed() {
```
dpc, line 305: We should probably retain the shutdown on submission channel closure as well, no?

fedimint-server/src/consensus/server.rs
```diff
@@ -75,8 +75,8 @@ pub struct ConsensusServer {
     api_endpoints: Vec<(PeerId, SafeUrl)>,
     cfg: ServerConfig,
     submission_receiver: Receiver<ConsensusItem>,
+    shutdown_receiver: watch::Receiver<Option<u64>>,
```
dpc, line 78: The fact that this is not persisted makes it a bit clunky in practice. Someone's fedimintd might get restarted and the whole scheduled upgrade fails because their instance ran ahead.

fedimint-server/src/net/api.rs
```diff
@@ -551,6 +557,15 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
                 Ok((&fedimint.session_status(index).await).into())
             }
         },
+        api_endpoint! {
+            SHUTDOWN_ENDPOINT,
+            ApiVersion::new(0, 0),
+            async |fedimint: &ConsensusApi, context, index: Option<u64>| -> () {
+                check_auth(context)?;
+                fedimint.shutdown(index);
```
dpc, line 565: Ideally here we should check against current session count and error out if it's too close or past it. This is to prevent races. We should require that the shutdown is at least 1 or maybe 2 sessions in the future, so that delay between here and consensus thread can't make it miss the shutdown. Also persist.

fedimint-server/src/consensus/server.rs
dpc: Hmmm... I guess if it happened only for one member it would be fine, as other's will shut down, so maybe not a problem?

fedimint-server/src/consensus/server.rs
joschisan: I cannot remember why I put it in there at the time, it’s also only in the single guardian case and not in the consensus part and seems unnecessary to me so I took it out for consistency.

fedimint-server/src/consensus/server.rs
joschisan: The idea is that the index is only set a few sessions in the future and the guardians monitor the server the entire time between setting the index and the system actually shutting down to be sure it happened at the right index.

joschisan: If one member misses the session by setting the index to late the guardians would have to reorganise via chat and restart the process

dpc (approved):

justinmoon:
dev call: we should have test coverage for this. we could implement by dumping the db after all fedimintd's have shutdown ...

fedimint-backports:

```bash
git worktree add -d .worktree/backport-5194-to-releases/v0.3 origin/releases/v0.3
cd .worktree/backport-5194-to-releases/v0.3
git switch --create backport-5194-to-releases/v0.3
git cherry-pick -x 5d934d601a2c5d05eee5c9a5bd024ddc527eaf52
```

PR #5197, opened 2024-05-03
chore(deps): use secp256k1 version exported by bitcoin

tvolk131:
The bitcoin crate exports the version of the secp256k1 crate that it uses. Why not just rely on that?

Also rely on secp256k1 instead of secp256k1-zkp.


tvolk131:
Closing in favor of #5219 based on a devcall discussion

PR #5211, opened 2024-05-06
feat: enable migration of ln config to lnv2 config

joschisan:
This is safe because encrypted preimage can only be decrypted in their respective module as the bls signatures internal to the ciphertext which authenticate the encryption commit to different messages hashes.

### DIFF
--- a/modules/fedimint-lnv2-common/src/config.rs
+++ b/modules/fedimint-lnv2-common/src/config.rs
@@ -5,6 +5,7 @@ use fedimint_core::core::ModuleKind;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::envs::BitcoinRpcConfig;
 use fedimint_core::{plugin_types_trait_impl_config, Amount, PeerId};
+use group::Curve;
 use serde::{Deserialize, Serialize};
 use tpe::{AggregatePublicKey, PublicKeyShare, SecretKeyShare};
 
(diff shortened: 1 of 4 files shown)


justinmoon:
dev call: ideally we could add arbitrary new modules to running federation, but that's long-term project. these changes help us to add just the lnv2 module to a running federation that has the lnv1 module.

modules/fedimint-lnv2-common/src/config.rs
```diff
@@ -107,3 +108,51 @@ impl Default for FeeConsensus {
...
 }
+
+#[allow(dead_code)]
+fn migrate_config_consensus(
+    config: fedimint_ln_common::config::LightningConfigConsensus,
+    peer_count: u16,
+) -> LightningConfigConsensus {
+    LightningConfigConsensus {
+        tpe_agg_pk: AggregatePublicKey(config.threshold_pub_keys.public_key().0.to_affine()),
+        tpe_pks: (0..peer_count)
```
elsirion, line 119: Didn't we talk about not wanting this redundancy anymore? You proposed saving the public polynomial instead iirc and deriving both shares and the aggregated key from that on startup.

elsirion (approved):
Looks reasonable.

m1sterc001guy (dismissed):

dpc (approved):

Issue #5195, opened 2024-05-02
mint_client_sanity flake: losing 1msat balance


dpc:
```
 fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:43 2024-05-02T20:26:22.963236Z  INFO mint_client_sanity: receiver info info={"federation_id":"7267eec085ffc1b2c3289a043741bc39e8036930679e71c387d9a1011b6eb151","network":"regtest","meta":{"federation_name":"testfed"},"total_amount_msat":19,"total_num_notes":10,"denominations_msat":{"1":5,"2":3,"4":2}}
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:43 thread 'main' panicked at modules/fedimint-mint-tests/src/bin/mint-client-sanity.rs:75:5:
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:43 assertion `left == right` failed
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:43   left: 19
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:43  right: 20
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:43 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fedimint-test-all-wasm32-unknown-unknown-ci> ## FAIL: mint_client_sanity
```

https://github.com/fedimint/fedimint/actions/runs/8929917822/job/24528829288

It could be related to another flake we've seen where we are sometimes losing a note.

dpc:
Ping @elsirion for awareness.

dpc:
```
> retry-success.sh ./scripts/tests/mint-client-sanity.sh
```

reproduces on my system after a long while

dpc:
Interesting observation: if you restore the client from passphrase it has correct balance. So it's definitely the client that lost something.

maan2003:
there is `Cmd::run_with_logging`, you can't use `out_string` then

maan2003:
oh there is also `FM_DEVIMINT_CMD_INHERIT_STDERR` which will have stderr inherit

dpc:
Yeah, I run with `retry-success.sh env RUST_LOG=fm=debug,info FM_DEVIMINT_CMD_INHERIT_STDERR=1 ./scripts/tests/mint-client-sanity.sh`, and also wrote a test in which a client remints over and over `1msat` and it's the fastest way to trigger. Eventually the note disappears.

[...]

PR #5212, opened 2024-05-06
fix: make RocksDB maximally safe

elsirion:
While investigating #5195 I came across the idea that RocksDB may be losing data in case of unclean shutdowns. While I don't think there's a high probability for that causing #5195, it still seems worth fixing. @m1sterc001guy lmk if I misunderstood something.

See WAL docs: https://github.com/facebook/rocksdb/wiki/WAL-Performance

<p align="center">
  Screenshot 2024-05-06 at 15-53-36 PayPerQ
  &nbsp; &nbsp; &nbsp; &nbsp; 
  Screenshot 2024-05-06 at 15-53-22 PayPerQ
</p>

### DIFF
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -14,7 +14,9 @@ use fedimint_core::db::{
 };
 use futures::stream;
 pub use rocksdb;
-use rocksdb::{OptimisticTransactionDB, OptimisticTransactionOptions, WriteOptions};
+use rocksdb::{
+    DBRecoveryMode, OptimisticTransactionDB, OptimisticTransactionOptions, WriteOptions,
+};
 use tracing::debug;
 
 use crate::envs::FM_ROCKSDB_WRITE_BUFFER_SIZE_ENV;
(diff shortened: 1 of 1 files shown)


dpc:
Should we do it only on the client side? On the server side peer redundancy renders this less of a problem, and shutdowns happen less often. On the client side performance matters less and client gets abruptly terminated all the time.

m1sterc001guy (approved):
Hm RocksDb continues to amaze me. I would have thought `wal_sync=true` would have been the default option. This config seems to break the whole notion of a transaction, I'm surprised its even configurable.

m1sterc001guy:
Perf might be impacted by this, but likely only under highly concurrently scenarios. We should see if our latency tests get any worse

dpc:
> Perf might be impacted by this, but likely only under highly concurrently scenarios.

It's not only high concurrency scenarios. Any write to WAL will end with `fsync` which is often quite slow and will add some fixed amount of delay before it returns. On something like nvme it might be 10-100 us, on ssd 1-10ms, on hdd 10-100ms. If the system is busy it might be even higher, up to seconds.

@m1sterc001guy @elsirion

elsirion:
I don't think perf is a concern for now. We can always optimize, but reliability should be the highest priority.

PR #5217, opened 2024-05-06
chore: gateway list_active_channels command includes funding outpoint

tvolk131:
To close a channel, both LDK and LND require the channel's funding TXID and output index, so we're adding them to the `gateway lightning list-active-channels` command output to ensure all necessary data is present to close channels

### DIFF
--- a/gateway/ln-gateway/src/lightning/lnd.rs
+++ b/gateway/ln-gateway/src/lightning/lnd.rs
@@ -787,12 +787,25 @@ impl ILnRpcClient for GatewayLndClient {
                             0
                         };
 
+                    let (channel_point_txid, channel_point_output_index) = channel
+                        .channel_point
+                        .split_once(':')
+                        .map(|(txid, output_index)| {
+                            (
+                                txid.to_string(),
+                                output_index.parse::<u32>().unwrap_or_default(),
+                            )
+                        })
+                        .unwrap_or_default();
+
                     ChannelInfo {
                         remote_pubkey: channel.remote_pubkey,
                         channel_size_sats,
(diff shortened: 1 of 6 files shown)


gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -787,12 +787,25 @@ impl ILnRpcClient for GatewayLndClient {
...
+                    let (channel_point_txid, channel_point_output_index) = channel
+                        .channel_point
+                        .split_once(':')
+                        .map(|(txid, output_index)| {
+                            (
+                                txid.to_string(),
+                                output_index.parse::<u32>().unwrap_or_default(),
+                            )
+                        })
+                        .unwrap_or_default();
```
elsirion, line 799: We could actually save that custom parsing logic by just using [`OutPoit::from_str`](https://docs.rs/bitcoin/latest/bitcoin/struct.OutPoint.html#impl-FromStr-for-OutPoint)

m1sterc001guy:
See https://github.com/fedimint/fedimint/pull/5218#issuecomment-2100838239

We could save needing to do this if we just want to use the `remote_pubkey` as the key for indicating which channel(s) to close.

tvolk131:
> We could save needing to do this if we just want to use the `remote_pubkey` as the key for indicating which channel(s) to close.

Good point! Converting this to a draft for now, will probably close later.

justinmoon:
dev call: might not need this anymore ...

tvolk131:
Closing now that #5218 is merged, which allows closing channels by peer ID rather than by funding outpoint

PR #5218, opened 2024-05-07
feat: add close_channels_with_peer command to gateway liquidity API

tvolk131:
This will be an important command to have once we ship an LDK-powered gateway since the liquidity API will be the only way to manage channels

~~Only implementing for LND for now since LND and LDK support closing channels by funding TXID + output index, while CLN instead requires funding tx block height + transaction index + output index~~

Changed from closing by TXID + output index to closing all channels with a specific peer

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -10,9 +10,10 @@ use fedimint_core::{fedimint_build_code_version_env, BitcoinAmountOrAll};
 use fedimint_logging::TracingSetup;
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
 use ln_gateway::rpc::{
-    BackupPayload, BalancePayload, ConfigPayload, ConnectFedPayload, ConnectToPeerPayload,
-    DepositAddressPayload, FederationRoutingFees, GetFundingAddressPayload, LeaveFedPayload,
-    OpenChannelPayload, RestorePayload, SetConfigurationPayload, WithdrawPayload, V1_API_ENDPOINT,
+    BackupPayload, BalancePayload, CloseChannelsWithPeerPayload, ConfigPayload, ConnectFedPayload,
+    ConnectToPeerPayload, DepositAddressPayload, FederationRoutingFees, GetFundingAddressPayload,
+    LeaveFedPayload, OpenChannelPayload, RestorePayload, SetConfigurationPayload, WithdrawPayload,
+    V1_API_ENDPOINT,
 };
 use serde::Serialize;
 
@@ -139,6 +140,13 @@ pub enum LightningCommands {
         /// The amount to push to the other side of the channel
         #[clap(long)]
         push_amount_sats: Option<u64>,
+    },
+    /// Close all channels with a peer, claiming the funds to the lightning
+    /// node's on-chain wallet
+    CloseChannelsWithPeer {
+        // The public key of the node to close channels with
+        #[clap(long)]
+        pubkey: bitcoin::secp256k1::PublicKey,
     },
     /// List active channels
     ListActiveChannels,
@@ -315,6 +323,12 @@ async fn main() -> anyhow::Result<()> {
                         push_amount_sats: push_amount_sats.unwrap_or(0),
                     })
                     .await?;
+            }
+            LightningCommands::CloseChannelsWithPeer { pubkey } => {
+                let num_channels_closed = client()
+                    .close_channels_with_peer(CloseChannelsWithPeerPayload { pubkey })
+                    .await?;
+                println!("Closed {num_channels_closed} channel(s)");
             }
             LightningCommands::ListActiveChannels => {
                 let response = client().list_active_channels().await?;
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -325,10 +325,10 @@ async fn main() -> anyhow::Result<()> {
                     .await?;
             }
             LightningCommands::CloseChannelsWithPeer { pubkey } => {
-                let num_channels_closed = client()
+                let response = client()
                     .close_channels_with_peer(CloseChannelsWithPeerPayload { pubkey })
                     .await?;
-                println!("Closed {num_channels_closed} channel(s)");
+                print_response(response);
             }
             LightningCommands::ListActiveChannels => {
                 let response = client().list_active_channels().await?;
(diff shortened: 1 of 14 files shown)


m1sterc001guy:
> Only implementing for LND for now since LND and LDK support closing channels by funding TXID + output index, while CLN instead requires funding tx block height + transaction index + output index

CLN supports closing a channel by peer id: https://docs.corelightning.org/reference/lightning-close

For LND/LDK, you could lookup the TXID + output index using something like `ListChannels` https://github.com/lightningnetwork/lnd/blob/3e36df42e590ef44c8f8f99afc5657024dc462c7/lnrpc/lightning.proto#L1595

Maybe it makes more sense to have the API accept a `PublicKey` as an indicate of the channel(s) to close?

tvolk131:
> CLN supports closing a channel by peer id: https://docs.corelightning.org/reference/lightning-close

That works, but if there are multiple channels open with the same peer then CLN refuses to close any channels:

`Failed to close channel: Error code -1: Peer has multiple channels: use channel_id or short_channel_id`

What do you think about adding a `CloseChannelsWithPeer` command instead of `CloseChannel` and under the hood for LND CLN and LDK we list channels, filtering by peer ID, then close all matching channels?

gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -245,6 +248,11 @@ message OpenChannelRequest {
   uint64 push_amount_sats = 3;
 }
 
+message CloseChannelsWithPeerRequest {
+  // The public key of the node we're closing channels with.
+  string pubkey = 1;
```
m1sterc001guy, line 253: Can we use bytes instead of string?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -351,6 +357,15 @@ async fn open_channel(
     Ok(Json(json!(())))
 }
 
+#[instrument(skip_all, err, fields(?payload))]
+async fn close_channels_with_peer(
+    Extension(gateway): Extension<Gateway>,
+    Json(payload): Json<CloseChannelsWithPeerPayload>,
+) -> Result<impl IntoResponse, GatewayError> {
+    gateway.handle_close_channels_with_peer_msg(payload).await?;
+    Ok(Json(json!(())))
```
m1sterc001guy, line 366: Maybe we want to return how many channels were closed?

m1sterc001guy:
Overall I think its good enough to merge, I would prefer to fix the String vs bytes issue though.

Perhaps we should start adding devimint tests that exercise these endpoints?

gateway/ln-gateway/src/rpc/rpc_server.rs
tvolk131: Done.

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: Done. The other liquidity API calls already use a string instead of bytes, but I'll fix this in a separate PR.

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -732,6 +733,60 @@ impl ILnRpcClient for GatewayLndClient {
...
+            let channel_point =
+                bitcoin::OutPoint::from_str(&channel.channel_point).map_err(|e| {
+                    LightningRpcError::FailedToCloseChannelsWithPeer {
+                        failure_reason: format!("Failed to parse channel point {e:?}"),
+                    }
+                })?;
+
+            client
+                .lightning()
+                .close_channel(CloseChannelRequest {
```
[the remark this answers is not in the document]
tvolk131: I talked with tnull and it sounds like `ldk-node` v0.3 will provide a much better API for tracking channel lifecycle. It sounds like v0.3 will be released very soon, so I'll probably wait for that and then put more thought into how we want to expose channel state tracking behavior.

tvolk131:
> Perhaps we should start adding devimint tests that exercise these endpoints?

Agreed! First I'd like to remove `connect-to-peer` and merge the behavior into `open-channel`

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -732,6 +732,57 @@ impl ILnRpcClient for GatewayLndClient {
...
+                peer: pubkey.serialize().to_vec(),
+            })
+            .await
+            .map_err(|e| LightningRpcError::FailedToCloseChannelsWithPeer {
+                failure_reason: format!("Failed to list channels {e:?}"),
+            })?
+            .into_inner()
+            .channels;
+
+        for channel in &channels_with_peer {
```
elsirion, line 754: A failure in this loop would be annoying since the first channels are already closed while later ones stay open. Seems like a potential source of bugs. I'd much prefer the "single channel closing" semantics eventually.

elsirion (approved):
Feel free to merge

gateway/cli/src/main.rs
```diff
@@ -316,6 +324,12 @@ async fn main() -> anyhow::Result<()> {
                     })
                     .await?;
             }
+            LightningCommands::CloseChannelsWithPeer { pubkey } => {
+                let num_channels_closed = client()
+                    .close_channels_with_peer(CloseChannelsWithPeerPayload { pubkey })
+                    .await?;
+                println!("Closed {num_channels_closed} channel(s)");
```
m1sterc001guy, line 331: Ick. Can we change this to return JSON with `num_channels_closed` as one of the fields? That way tests can parse it and we can use `print_response` here instead

gateway/ln-gateway/src/lightning/lnd.rs
m1sterc001guy: This function doesn't wait for the channel to close anyway, we'll eventually need a way to track that as well. Rather than failing and returning here (and above in the CLN implementation), I think emitting a warning and trying to close the next channel would be sufficient

gateway/cli/src/main.rs
elsirion: Totally missed that :grima

gateway/ln-gateway/build.rs
```diff
@@ -9,6 +9,7 @@ fn main() {
         .build_server(true)
         .build_client(true)
         .protoc_arg("--experimental_allow_proto3_optional")
+        .type_attribute(".", "#[derive(serde::Serialize, serde::Deserialize)]")
```
tvolk131, line 12: We're adding this so we can encode `CloseChannelsWithPeerResponse` as JSON

elsirion (approved):

PR #5224, opened 2024-05-07
cleanup: simplify verification hashes

joschisan:
The verification hashes are used in the API to confirm the result of the DKG. Though all hashes are just derived from the config, and we could simply just compare the consensus config hash, the verification hashes are purposefully different for each peer such that their manual verification by the guardians via the UI is more robust.

After the dkg is done the verification hashes are not required anymore, the server checks config equality on startup via the SERVER_CONFIG_CONSENSUS_HASH_ENDPOINT.

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -18,6 +18,7 @@ use fedimint_core::config::{
 };
 use fedimint_core::core::ModuleInstanceId;
 use fedimint_core::db::Database;
+use fedimint_core::encoding::Encodable;
 use fedimint_core::endpoint_constants::{
     ADD_CONFIG_GEN_PEER_ENDPOINT, AUTH_ENDPOINT, CONFIG_GEN_PEERS_ENDPOINT,
     CONSENSUS_CONFIG_GEN_PARAMS_ENDPOINT, DEFAULT_CONFIG_GEN_PARAMS_ENDPOINT,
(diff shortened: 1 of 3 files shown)


fedimint-server/src/net/api.rs
```diff
@@ -575,14 +574,6 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
...
             }
         },
-        api_endpoint! {
-            VERIFY_CONFIG_HASH_ENDPOINT,
-            ApiVersion::new(0, 0),
-            async |fedimint: &ConsensusApi, context, _v: ()| -> BTreeMap<PeerId, sha256::Hash> {
-                check_auth(context)?;
-                Ok(get_verification_hashes(&fedimint.cfg))
-            }
-        },
```
elsirion, line 585: @okjodom @Kodylow Why did we add that endpoint to the running server API originally? Was it to prevent timing issues when some guardians started consensus before others? I can't see the point because I assume this API is only used by the guardian running this instance, so isn't needed after consensus is running, but want to be sure.

elsirion (dismissed):
Looks ok, but would like input from the UI side

fedimint-server/src/config/api.rs
```diff
@@ -327,13 +328,24 @@ impl ConfigGenApi {
...
         ];
+
         let state = self.require_any_status(&expected_status).await?;
+
         let config = state
             .config
             .clone()
             .ok_or(ApiError::bad_request("Missing config".to_string()))?;
 
-        Ok(get_verification_hashes(&config))
```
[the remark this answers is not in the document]
joschisan: Done

fedimint-server/src/net/api.rs
joschisan: The endpoint requires auth, hence can only be requested by the corresponding guardian via CLI or UI; however this guardian already explicitly started the consensus so why would he check the verification hashes now?

This is in contrast to the SERVER_CONFIG_CONSENSUS_ENDPOINT which also returns a config hash when running, however this is supposed to be requested by the other guardians to check against their own.

okjodom (approved):

PR #5227, opened 2024-05-07
chore: more informative error messages around accepting transactions

dpc:

### DIFF
--- a/fedimint-core/src/transaction.rs
+++ b/fedimint-core/src/transaction.rs
@@ -80,16 +80,20 @@ impl Transaction {
     pub fn validate_signatures(
         &self,
         pub_keys: Vec<secp256k1_zkp::PublicKey>,
+        txid: TransactionId,
     ) -> Result<(), TransactionError> {
         let signatures = match &self.signatures {
             TransactionSignature::NaiveMultisig(sigs) => sigs,
             TransactionSignature::Default { variant, .. } => {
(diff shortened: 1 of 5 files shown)


fedimint-core/src/transaction.rs
```diff
@@ -125,25 +129,34 @@ pub enum TransactionSignature {
...
+        error: DynInputError,
+        input_idx: u64,
+        txid: TransactionId,
+    },
+    #[error("The transaction {txid} had an invalid output at index {}: {}",  .output_idx, .error)]
+    Output {
+        error: DynOutputError,
+        output_idx: u64,
+        txid: TransactionId,
+    },
```
elsirion, line 161: Feels like this should be split into a struct containing the txid and an error variant that contains the information currently available. This feels like a useful quick-fix, but not really great code in general.

elsirion (approved):
I'm on the edge with this one, is there a good reason to repeat the txid over all variants other than "slicghtly easiert to impl"? Feel free to merge to not block #5195 debugging, but I'd rather see it refactored.

dpc:
> I'm on the edge with this one, is there a good reason to repeat the txid over all variants other than "slicghtly easiert to impl"?

If the message is to print the txid (which is handy), then trying to merge it going to be a bit involved. And a bit of very localized redundancy is a minor in-elegancy, not a big deal.

dpc:
This is broken because of #5238

dpc:
Replaced by #5244

PR #5228, opened 2024-05-07
chore(cli): no need to log what we are already printing out

dpc:


bradleystachurski (approved):
Looks good 👍 

For future reference, related thread regarding logging the error and writing the error json to stderr: https://github.com/fedimint/fedimint/pull/4563#discussion_r1525646521

PR #5234, opened 2024-05-08
fix: use async mutex

joschisan:


okjodom (approved):
good catch

elsirion:
I prefer it this way, but unless the lock was held across an await point a normal std `Mutex` is fine. The problem imo is that it's way too easy to accidentally break this invariant and it won't show immediately, but rather randomly deadlock in some cases.

Maybe you found such an await point crossing (I didn't check too closely)? Writing a PR description would help explain your reasoning.

joschisan:
This pr would not have worked with a sync mutex: https://github.com/fedimint/fedimint/pull/5235

PR #5235, opened 2024-05-08
simplify: do not stage config

joschisan:
Not writing to disk in the ConfigGenApi  makes it easier to reason about the behaviour of the ConfigGenApi struct as it is conceptually encapsulated. Now FedimintServers behaviour only depends on files that it has written to disk itself while ConfigGenApi has no side effects. Furthermore, a crashed config gen starts from a clean slate and does not have to deal with potential remnants.

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -1,13 +1,10 @@
 use std::collections::BTreeMap;
-use std::fs;
 use std::net::SocketAddr;
-use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
 use async_trait::async_trait;
 use bitcoin_hashes::sha256;
-use fedimint_aead::random_salt;
 use fedimint_api_client::api::{DynGlobalApi, StatusResponse};
 use fedimint_core::admin_client::{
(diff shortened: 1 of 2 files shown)


elsirion (changes requested):
ConceptNACK: staging was added on purpose to avoid a potential bug  from accidentally deleting the live guardian config. CC @justinmoon who originally proposed this.

joschisan:
@elsirion Could you please elaborate on the issue and why staging avoids it?

elsirion:
We only have code that can delete files in staging, not the final config directory, so if that function is called by accident nothing too had should happen.

joschisan:
@elsirion This pr does not introduce any code to delete the config files either; the config is only written to disk after the start_consensus method has been called.

elsirion:
After taking a deeper look at the code what you want to do is postponing writing of the config till DKG is finished and configs have been verified. That way we don't have to delete anything anymore when resetting the cofiggen process. One drawback of this is that you cannot restart `fedimintd` mid-setup anymore, which iirc was the reason for going with this approach, but we never really tested it, so I think we could just do away with this feature and this PR looks ok at a high level.

elsirion (dismissed):

elsirion:
Needs rebase :(

tvolk131 (approved):

dpc (approved):

PR #5236, opened 2024-05-08
fix: typed error for reissue external notes

maan2003:

### DIFF
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -743,6 +743,14 @@ impl ClientModule for MintClientModule {
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum ReissueExternalNotesError {
+    #[error("Federation ID does not match")]
+    WrongFederationId,
+    #[error("We already reissued these notes")]
+    AlreadyReissued,
+}
+
 impl MintClientModule {
     async fn create_sufficient_input(
         &self,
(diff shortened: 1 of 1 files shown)


dpc (approved):

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -743,6 +743,14 @@ impl ClientModule for MintClientModule {
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum ReissueExternalNotesError {
+    #[error("Federation ID does not match")]
+    DifferentFederationId,
```
dpc, line 749: "wrong" would be a better word than "different", imo

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -743,6 +743,14 @@ impl ClientModule for MintClientModule {
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum ReissueExternalNotesError {
```
tvolk131, line 747: What if we added a third error type?

```
    #[error("Untyped error")]
    Other(anyhow::Error)
```

Then we could change the return type of `reissue_external_notes()` to `Result<OperationId, ReissueExternalNotesError>` and remove this line:

```/// Can return error of type [`ReissueExternalNotesError`]```

maan2003: I think .downcast is fine, Other cause error chain issues

dpc (approved):

fedimint-backports:
- #5245

Issue #5238, opened 2024-05-08
Replace submit_transaction endpoint with less hopeless one :D


dpc:
```
        api_endpoint! {
            SUBMIT_TRANSACTION_ENDPOINT,
            async |fedimint: &ConsensusApi, _context, transaction: SerdeTransaction| -> SerdeModuleEncoding<Result<TransactionId, TransactionError>> {
                let transaction = transaction
                    .try_into_inner(&fedimint.modules.decoder_registry())
                    .map_err(|e| ApiError::bad_request(e.to_string()))?;

                // we return an inner error if and only if the submitted transaction is
                // invalid and will be rejected if we were to submit it to consensus
                Ok((&fedimint.submit_transaction(transaction).await).into())
            }
        },

```

`SerdeModuleEncoding<Result<TransactionId, TransactionError>>`. That's terrible. Basically we encode the result using consensus encoding, making it impossible to evolve.

The top-level error is:

```
pub struct ApiError {
    pub code: i32,
    pub message: String,
}
```

so we should just `impl From<TransactionError> for ApiError`, and convert every case to a specific code, and throw the `to_string()` of the error into a `message`.

This will break the backward compat, so we should just make a new call for it.

Generally the way we handle errors in api is terrible, AFAICT.

BTW. If `ApiError::code` is the jsonrpc error code, then we should use values between -32000 to -32099, as per https://www.jsonrpc.org/specification , not http codes (see `impls` for `ApiError`).

elsirion:
I remember lengthy discussions about this, is it still relevant @dpc?

dpc:
Not a priority, I guess.

PR #5244, opened 2024-05-09
Client API error handling overheal in a new submission call

dpc:
### Introduce a new tx submission call

See #5238 

To achieve that some more changes were needed.

### Overhaul error handling

Use jsonrpc error codes and `ApiError` as a carrier for api errors. This is a big change, that requires some thinking through.

JsonRPC standard says that all non-server-defied error codes are for application use, which jsonrpsee calls "Call"-errors.

It would generally make our errors much more idiomatic, which should make 3rd-party clients and tools job easier. Over time we should stop relying on error messages and use error codes, which now become first-class citizen.

The errors will now not be serialized, but JsonRPC error also has an `data` that can encode free form extra data. If needed even consensus-encoded fields can be passed in it (like I do for `DynInputError` etc. here).

### Adapt strategy error handling

The "call" errors need to get consensus on, in a similar way that normal values do. This requires some tweak to error strategy.


Cargo.toml
```diff
@@ -90,6 +90,8 @@ serdect = "0.2.0"
 bls12_381 = "0.8.0"
 group = "0.13.0"
 itertools = "0.12.1"
+jsonrpsee-core = "0.22.5"
```
EthnTuttle, line 93: image
Do any of these other Cargo.toml's need to be updated?

EthnTuttle: I only see the fedimint-api-client one changing.

dpc: Will need to `workspace = true` in more places. Not touching it RN in this PR. :D

EthnTuttle: https://github.com/fedimint/fedimint/issues/5248

fedimint-client/src/transaction/sm.rs
```diff
@@ -107,19 +108,27 @@ impl State for TxSubmissionStates {
...
         loop {
-            match context.api().submit_transaction(tx.clone()).await {
-                Ok(serde_result) => match serde_result.try_into_inner(context.decoders()) {
-                    Ok(result) => {
-                        if let Err(transaction_error) = result {
-                            return transaction_error.to_string();
-                        }
+            if SUBMIT_ENDPOINT_API_VERSION <= context.core_api_version() {
+                if let Err(e) = context.api().submit_transaction(tx.clone()).await {
+                    if let Some(call_error) = e.try_to_call_error(context.num_peers()) {
```
elsirion, line 113: Switching to `try_to_call_error` changes the semantics here and relaxes consensus requirements. Previously peers had to agree on the error data, now it can be only the error code in the worst case. Do we want that? Have we thought of all the potential side-effects?

fedimint-api-client/src/query.rs
```diff
@@ -80,11 +85,28 @@ impl ErrorStrategy {
...
+            // For a call-errors to surface as a federation-level, it needs get a threshold
+            // of responses being call-errors.
             QueryStep::Failure {
                 general: Some(anyhow!(
-                    "Received errors from {} peers: {}",
-                    self.threshold,
+                    "Received {} out of {} call errors from peers: {}",
+                    self.num_peers.threshold(),
+                    self.num_peers,
                     self.format_errors()
```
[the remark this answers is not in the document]
joschisan: We need the different threshold here since a query strategy should fail on f + 1 rpc errors, however on call errors (like TransactionError) we need a threshold of 2f + 1 to achieve consensus; if we use f+1 for call errors. Imo rpc errors and TransactionErrors are two very different concepts and should not be squashed into the same struct just because they are both called "Errors".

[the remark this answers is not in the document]
joschisan: Squashing these errors requires us to always keep in mind to call is_call_err to differentiate when handling them in the client; for example only retry when is_call_err returns false.

joschisan:
Concept NACK: Rpc errors and transaction errors are two completely different concepts and should not be squashed into one; For example Rpc errors can be retried while transaction errors are final and usually imply a critical failure like a faulty federation. Squashing this into one requires us to always differentiate via the is_call_err method to check and apply different logic, like different thresholds in the query strategy for example. The additional complexity introduced in code as central as the query strategies is imo not worth it; I feel very strongly about this as query strategies are generally very easy to mess up.

fedimint-api-client/src/api.rs
```diff
@@ -131,6 +140,81 @@ pub struct FederationError {
...
+        // If data doesn't match, that's OK, just ignore it
+        for ((code, message), count) in &count_by_no_data {
+            if num_peers.threshold() <= *count {
+                return Some(ApiError {
+                    code: *code,
+                    message: message.clone(),
+                    data: None,
+                });
+            }
+        }
```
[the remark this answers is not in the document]
dpc: The reality is that there can be consensus on the code, but not on the details. It's up to the caller to figure it out what to do about it. I would expect most client calls would be OK with just a code. Snooping in the data seems brittle anyway, and I would expect it to be useful mostly for debugging.

dpc:
> * Sticking to standards is nice, but given that there likely won't be a second client impl it probably isn't something we should expand too much energy on. Do error codes actually help us long term?


We need some way to enumerate potential failure cases. Using `Decodable` `enum` was kinda trying to do that, but is just too rigid on every other account.

Error codes are robust way to enumerate and distinguish errors.

And since client side needs to get a consensus on the error itself, getting consensus on error code itself, seems most natural. Otherwise if consensus on the application error is to be established on the entirety of the error (like it's encoding or serialization), it's impossible to even fix a typo in an error.

I'm not strongly set on **jsonrpc** error codes ... but it does have nice properties, like not requiring another level of wrapping error in a success on the wire.



> * Touching the query strategy makes me feel uneasy, huge potential to break stuff and cause us a lot of debugging.


There are two different aspects here:

* How is the application-level (aka transaction) error represented on the wire.
* How is it implemented in the client code.


Conceptually one way or the other the result of an api call to a peer has 3 outcomes:

```
enum ApiResult {
  TransportSuccess(CallOutcome)
  TransportError
}

enum CallOutcome {
  ApplicationError(...)
  ApplicationSuccess
}
```

The code in this PR. is clumsy w.r.t. query strategy, because it doesn't convert the underlying jsonrpc error to something like `ApiResult`. I think with a bit better refactoring, query strategy logic could largely remain intact.

> As you can see the behaviour of those strategies is quite subtle and I would avoid further complexity here at all costs to keep this maintainable.

Instead of being scared of it, maybe we should add some tests.

Fear is not a strategy for having maintainable code.

dpc:
To sum up:

* I agree that changes to the error handling query strategies here are clumsy, but I think it can be done better. It didn't occurred to me at the time, and I wanted to wrap it up so there's something to talk about.
* When client needs to establish some form of consensus on the non-success application-level outcome, what should it do it on exactly. Codes, error messages, some serialization of the whole thing? It seems to me that some form of "codes" is the way to go. "message" and "data" in my mind should be used for debugging purposes only. And I feel (much less strongly) that if it will be "codes", we might as well flatten it into "jsonrpc codes".
* If client code would "unflatten" the "flatten" jsonrpc error codes early, so that the query strategy logic can remain the same, would using jsonrpc error codes would be OK with you? If not - what else would you prefer and why?

elsirion:
First of all, I think @joschisan, @dpc and I should get on a call about this. You both have valid points and trying to get to a solution that everyone can agree on is very hard asynchronously.

> Using Decodable enum was kinda trying to do that, but is just too rigid

Could we just add a default variant? We need to model possible variants very rigidly here anyway since adding an unknown error that old clients can't interpret could lead to backwards incompatibility (e.g. they wouldn't know if it's retryable).

> Otherwise if consensus on the application error is to be established on the entirety of the error (like it's encoding or serialization), it's impossible to even fix a typo in an error.

The encoding of the error in question doesn't use strings, but enums (and thus indices).

> Query strategies are (or at least should be) very convenient to test. We call them with responses, they give back outputs that we can assert.

Well, I remember us having some very nasty bugs resulting from subtle errors in query strategies that @joschisan debugged for days back then. These were leading to very hard to debug, timing and online-ness dependent payment failures and loss of funds. Not messing them up again is a high priority. 

I second that they should be tested, but a comprehensive test suite would need to simulate different response orders, missing responses, etc. Fuzzing that would be great imo or possibly even exhaustive testing if the combinatorial complexity isn't too bad.

dpc:
> Could we just add a default variant? 

We could, but if the query strategy is just comparing whole encoding, then any new variant will not be considered the same error, even if it supposed to be.

> The encoding of the error in question doesn't use strings, but enums (and thus indices).

Right, my bad. So let's say - adding a new field, so it's possible to print the extra information. Adding a new field requires a new variant, even if that new variant means exactly same thing as the old variant, just carries more data.

The variants are an enumeration of possible ... encoded states. Error codes are an enumeration of possible "not OK results". Coupling them is the issue.

> We need to model possible variants very rigidly here anyway since adding an unknown error that old clients can't interpret could lead to backwards incompatibility (e.g. they wouldn't know if it's retryable).

Good point. We might not be able to add any new error codes to the existing rpc calls altogether. If you need more error codes, it needs to be a new api call, on a new api version.

BTW. That also made me realize that in a `submit`, possibly multiple inputs/outputs can have issues, so why would we return only the first one? That could possibly spook some logic somewhere.

Also - right now we completely ignore the transaction details. If enough peers return transaction error, it is just ignored. So possibly submit should just return a single application error code "transaction invalid" and that's about it. Everything else would be just "extra metdata" for debugging purposes.

I know that @joschisan was describing submit errors being useful for lnv2, but maybe that should be facilitated by another module-specific call? After lnv2-client-module tx gets "transaction invalid" from "submit", it can call some extra endpoint to figure out what to do about. This way it's actually possible to evolve the whole thing.

Module specific errors being returned from "submit" seems just wrong anyway.

> I second that they should be tested, but a comprehensive test suite would need to simulate different response orders, missing responses, etc. 


The timing doesn't matter, only the order of responses. So nothing fancy is really needed:

```rust
let mut query_strategy = SomeQueryStragy::new();
assert_eq!(query_strategy.process(PeerId(0), SomeFakeResponse { a: 1}), QueryStep::Continue);
assert_eq!(query_strategy.process(PeerId(1), SomeFakeResponse { a: 1}), QueryStep::Continue);
assert_eq!(query_strategy.process(PeerId(2), SomeFakeResponse { a: 1}), QueryStep::Success);
```

Sure, one could write a property test (fuzzing seems too low level, property testing fits better here) to test a lot of combinations, but in practice there are only a handful of boundary conditions worth testing for accidental malfunction.

elsirion:
> BTW. That also made me realize that in a `submit`, possibly multiple inputs/outputs can have issues, so why would we return only the first one?

Why waste compute on validating a tx that will not be accepted?

> I know that joschisan was describing submit errors being useful for lnv2, but maybe that should be facilitated by another module-specific call? After lnv2-client-module tx gets "transaction invalid" from "submit", it can call some extra endpoint to figure out what to do about. This way it's actually possible to evolve the whole thing.

Then we'd need to save information about rejected transactions, which iirc @joschisan specifically removed when migrating to AlephBFT (likely because DoS risk).

dpc:
@elsirion 

> Why waste compute on validating a tx that will not be accepted?

I don't mean we should accept it. We should reject it, but why would we return only details of the first error we spot. What if there are other errors? If client logic would ever depend on the details of the rejection, we can't guarantee that such details would be available, because how can we be certain that some irrelevant error wasn't spotted and returned first.

> Then we'd need to save information about rejected transactions

No we wouldn't. On rejected transaction the client, could send that whole transaction to e.g. LNv2 module API, so that module can validate LNv2 parts and respond to the client with whatever details the client needs to do what it needs to do.

dpc:
Stale, and I don't think it will get any action anytime soon.

PR #5247, opened 2024-05-09
Small improvements to Fedimint Server

joschisan:
This pr bundles a few small improvements in a larger effort to clean up and refactor Fedimint Server.

### DIFF
--- a/fedimint-server/src/atomic_broadcast/spawner.rs
+++ b/fedimint-server/src/atomic_broadcast/spawner.rs
@@ -1,24 +1,21 @@
+use fedimint_core::task::TaskGroup;
 use fedimint_logging::LOG_CONSENSUS;
 use tracing::warn;
 
 #[derive(Clone)]
-pub struct Spawner;
-
-impl Spawner {
-    pub fn new() -> Self {
-        Self {}
-    }
+pub struct Spawner {
+    task_group: TaskGroup,
 }
 
-impl Default for Spawner {
-    fn default() -> Self {
-        Self::new()
+impl Spawner {
+    pub fn new(task_group: TaskGroup) -> Self {
+        Self { task_group }
     }
 }
 
 impl aleph_bft::SpawnHandle for Spawner {
     fn spawn(&self, name: &str, task: impl futures::Future<Output = ()> + Send + 'static) {
-        fedimint_core::runtime::spawn(name, task);
+        self.task_group.spawn(name, move |_| task);
     }
 
     fn spawn_essential(
@@ -26,15 +23,16 @@ impl aleph_bft::SpawnHandle for Spawner {
         name: &str,
         task: impl futures::Future<Output = ()> + Send + 'static,
     ) -> aleph_bft::TaskHandle {
-        let (res_tx, res_rx) = futures::channel::oneshot::channel();
+        let (sender, receiver) = futures::channel::oneshot::channel();
 
-        fedimint_core::runtime::spawn(name, async move {
+        self.task_group.spawn(name, move |_| async {
             task.await;
-            if let Err(_err) = res_tx.send(()) {
+
+            if sender.send(()).is_err() {
                 warn!(target: LOG_CONSENSUS, "Unable to send essential spawned task completion. Are we shutting down?");
             }
         });
 
-        Box::pin(async move { res_rx.await.map_err(|_| ()) })
+        Box::pin(async move { receiver.await.map_err(|_| ()) })
     }
 }
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -1,11 +1,12 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
+use std::default::Default;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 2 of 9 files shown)


dpc:
> The server side migration is removed for now to avoid any confusion

I don't get it. Why?

fedimint-server/src/net/peers.rs
```diff
@@ -219,18 +181,16 @@ where
         delay_calculator: DelayCalculator,
         connect: PeerConnector<T>,
         task_group: &TaskGroup,
-    ) -> (Self, PeerStatusChannels) {
+        status_channels: Arc<RwLock<BTreeMap<PeerId, PeerConnectionStatus>>>,
```
dpc, line 184: I think I liked it returning multiple values instead of taking extra arg more.

But even more so, having a shorter named newtype was nicer.

dpc: BTW. Can't we just put `status_channels` in `self`?

dpc: Though if I see correctly, you're getting rid of these channels there, and replacing everything with just shared variables, with is :+1: and the original channels there in a map seemed clunky indeed.

fedimint-server/src/atomic_broadcast/spawner.rs
```diff
@@ -28,9 +25,10 @@ impl aleph_bft::SpawnHandle for Spawner {
     ) -> aleph_bft::TaskHandle {
         let (res_tx, res_rx) = futures::channel::oneshot::channel();
 
-        fedimint_core::runtime::spawn(name, async move {
+        self.task_group.spawn(name, move |_| async {
```
dpc, line 28: :+1:

fedimint-server/src/consensus/server.rs
```diff
@@ -77,6 +78,8 @@ pub struct ConsensusServer {
     self_id_str: String,
     /// Just a string version of peer ids for performance
     peer_id_str: Vec<String>,
+    connection_status_channels: Arc<RwLock<BTreeMap<PeerId, PeerConnectionStatus>>>,
```
dpc, line 78: Oh, I see you did put it into self, OK. But I'd at least make a `type ConnectionStatuses = Arc<RwLock<BTreeMap<PeerId, PeerConnectionStatus>>>`.

dpc: BTW. I guess there's no `_channels` anywhere here anymore.

joschisan:
> > The server side migration is removed for now to avoid any confusion
> 
> I don't get it. Why?

Having code here that we do not intend to use in the foreseeable future if ever is confusing imo. For every line of code there should be a reason why it exists, otherwise it clutters our project overview.

fedimint-server/src/consensus/server.rs
```diff
@@ -73,7 +71,7 @@ pub struct ConsensusServer {
     cfg: ServerConfig,
     submission_receiver: Receiver<ConsensusItem>,
     shutdown_receiver: watch::Receiver<Option<u64>>,
-    latest_contribution_by_peer: Arc<RwLock<LatestContributionByPeer>>,
+    last_ci_by_peer: Arc<RwLock<BTreeMap<PeerId, u64>>>,
```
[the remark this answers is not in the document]
joschisan: ci is common in consensus code but citem is also good.

fedimint-server/src/net/peers.rs
joschisan: It is actually not primarily about turning the return value into an argument but instead of being able to create the status_channels before we create the p2p network layer such that we can create the network layer after we build the consensus api which also contains a clone of status_channels. 

This allows us to more consistently factor fedimint server. Right now everything inside ConsensusServer::run is factored very consistently, in particular values are created as late as possible, right before we need them, so understanding a code section requires less context.

dpc:
> Having code here that we do not intend to use in the foreseeable future if ever is confusing imo. For every line of code there should be a reason why it exists, otherwise it clutters our project overview.

We don't have any server side migrations yet? Interesting. Seems like `dummy` has a migration just to test it.

Anyway - this code does have a reason to exists. One day we'll need to add an index to speed up something, etc. and we'll need it. Reverting it back and forth seems like a waste of time, and the fact that some migrations are not actually being run would be super confusing, especially given the module traits have existing methods on them, etc. I don't want to be testing and re-building infrastructure for it when we actually going to need it, after we did all that work. And if it's not there, somewhat tested and maintained, it will just be broken. 

"clutters project overview" is not a strong argument, IMO. Code is not a poetry. It primarily has to work well. Looking well is a secondary concern. And code that can't migrate db schema works less well, IMO.

fedimint-server/src/net/peers.rs
dpc: I see. Makes sense.

fedimint-server/src/consensus/server.rs
joschisan: I don't like new types since its more context to keep in my head.

joschisan:
@dpc took out the removal of server-side migration.

dpc (approved):

dpc:
Added a label for further review, probably best if someone else looks at it at least once.

dpc (approved):

PR #5251, opened 2024-05-10
chore: better tx submissin handling logging

dpc:

### DIFF
--- a/fedimint-server/src/consensus/server.rs
+++ b/fedimint-server/src/consensus/server.rs
@@ -587,6 +587,7 @@ impl ConsensusServer {
             .expect("This is the only place where we write to this key");
     }
 
+    #[instrument(target = "fm::consensus", skip(self, item), level = "info")]
     pub async fn process_consensus_item(
         &self,
         session_index: u64,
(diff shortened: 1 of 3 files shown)


fedimint-server/src/consensus/server.rs
```diff
@@ -587,6 +587,7 @@ impl ConsensusServer {
             .expect("This is the only place where we write to this key");
     }
 
+    #[instrument(target = "fm::consensus", skip(self, item), level = "info")]
```
dpc, line 590: I wish I could do constants we use, but it needs https://github.com/tokio-rs/tracing/pull/2941

fedimint-server/src/net/api.rs
```diff
@@ -111,7 +111,7 @@ impl ConsensusApi {
     ) -> Result<TransactionId, TransactionError> {
         let txid = transaction.tx_hash();
 
-        debug!(%txid, "Received mint transaction");
```
dpc, line 114: "mint" here is confusing (not for "mint" module).

joschisan (approved):

PR #5257, opened 2024-05-10
chore: better request handling logging

dpc:
Mostly about adding an ID, to help filter interleaved logging messages. But also I've noticed there was another span already there, just one that was not using `LOG_` constants.

### DIFF
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -18,13 +18,14 @@ use std::collections::BTreeMap;
 use std::fmt::{self, Debug, Formatter};
 use std::marker::{self, PhantomData};
 use std::pin::Pin;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 
 use fedimint_logging::LOG_NET_API;
 use futures::Future;
 use jsonrpsee_core::JsonValue;
 use serde::{Deserialize, Serialize};
-use tracing::instrument;
+use tracing::Instrument;
 
 // TODO: Make this module public and remove the wildcard `pub use` below
 mod version;
@@ -312,7 +313,6 @@ macro_rules! __api_endpoint {
             type Param = $param_ty;
             type Response = $resp_ty;
 
-            #[::tracing::instrument(skip_all, target="fm::net::api", level = "info", name="request_handler", fields(api=$path))]
             async fn handle<'state, 'context, 'dbtx>(
                 $state: &'state Self::State,
                 $context: &'context mut $crate::module::ApiEndpointContext<'dbtx>,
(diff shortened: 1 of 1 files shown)


fedimint-core/src/module/mod.rs
```diff
@@ -364,13 +367,6 @@ impl ApiEndpoint<()> {
         E::Param: Debug,
         E::Response: Debug,
     {
-        #[instrument(
-            target = "fedimint_server::request",
-            level = "trace",
-            skip_all,
-            fields(method = E::PATH),
-            ret,
-        )]
```
maan2003, line 373: you can do `fields(id = REQ_ID.fetch_add(1, Ordering::SeqCst))` btw

maan2003: or is it instrument not supporting CONST target

fedimint-core/src/module/mod.rs
```diff
@@ -356,6 +356,9 @@ pub struct ApiEndpoint<M> {
     pub handler: HandlerFn<M>,
 }
 
+/// Global request ID used for logging
+pub static REQ_ID: AtomicU64 = AtomicU64::new(0);
```
maan2003, line 360: why pub

fedimint-core/src/module/mod.rs
dpc: It was actually the `ID`, thanks, but while at it, I can also use const target, so let's just keep it this way.

maan2003: you can even do like 

```
fields(id = {
  static REQ_ID: AtomicU64 = AtomicU64::new(0); 
  REQ_ID.fetch_add(1, Ordering::SeqCst)) 
})

File: fedimint-core/src/module/mod.rs (hunk 2 above)
Comment (dpc) [reply fedimint-core/src/module/mod.rs]: Just general sloppiness.

**Review by maan2003 — APPROVED** [review]

File: fedimint-core/src/module/mod.rs [hunk 3]
```diff
@@ -400,7 +396,15 @@ impl ApiEndpoint<()> {
                         .to_typed()
                         .map_err(|e| ApiError::bad_request(e.to_string()))?;
 
-                    let ret = handle_request::<E>(m, &mut context, request).await?;
+                    let span = tracing::info_span!(
+                        target: LOG_NET_API,
+                        "api_req",
+                        id = REQ_ID.fetch_add(1, Ordering::SeqCst),
```
elsirion, line 402: Does JSON RPC allow sending some connection identifier? Would be great if we could track these requests across the gateway/client <> federation boundary (I think that's how distributed tracing is supposed to work?).

dpc: The client picks its own request ID and server responds with it.

But I'm not aware of any place to put a server-side ID in the response.

PR #5264, opened 2024-05-11
Refactor server

joschisan:
Just moving code around.


dpc (approved):

elsirion (dismissed):

elsirion (approved):

bradleystachurski:
review club: @joschisan what ended up being the solution for the problem you saw in CI?

Also, it would help in the review club if there was a description.

bradleystachurski:
review club: dpc relayed that the problem was resolved by removing the diff that caused it. We should followup with an investigation to see what we were getting the `No modules found of kind dummy` error for the mint tests.

bradleystachurski:
https://github.com/fedimint/fedimint/issues/5286

PR #5285, opened 2024-05-14
Refactor: remove unnecessary FedimintServer struct

joschisan:
Refactor: remove FedimintServer struct, turn run method in standalone function

### DIFF
--- a/fedimint-server/src/lib.rs
+++ b/fedimint-server/src/lib.rs
@@ -68,217 +68,193 @@ pub trait HasApiContext<State> {
     ) -> (&State, ApiEndpointContext<'_>);
 }
 
-/// Main server for running Fedimint consensus and APIs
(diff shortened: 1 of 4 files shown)


dpc (approved):
Seems like a good example of "if a class is just a constructor and one method, then it should be a function".

fedimint-server/src/lib.rs
```diff
@@ -68,217 +68,193 @@ pub trait HasApiContext<State> {
...
-        settings: ConfigGenSettings,
-        db: Database,
-        version_hash: String,
-    ) -> Self {
-        Self {
-            data_dir,
-            settings,
-            db,
-            version_hash,
+pub async fn run(
```
elsirion, line 71: nit: now that it's not a member anymore the fn should have a more descriptive name

PR #5288, opened 2024-05-14
fix(jit-values): they can actually get canceled

dpc:
Leading to panics on shutdown sometimes.

### DIFF
--- a/fedimint-core/src/task/jit.rs
+++ b/fedimint-core/src/task/jit.rs
@@ -110,27 +110,28 @@ where
             .val
             .get_or_init(|| async {
                 let handle: &mut _ = &mut *self.inner.handle.lock().await;
-                handle
-                    .await
-                    .unwrap_or_else(|err| {
+                match handle.await {
+                        Ok(Ok(o)) => Ok(o),
+                        Ok(Err(err)) => {
+                            let err_str = err.to_string();
+                            init_error = Some(err);
+                            Err(err_str)
+                        },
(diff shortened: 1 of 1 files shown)


fedimint-core/src/task/jit.rs
```diff
@@ -110,27 +110,28 @@ where
...
-                        if err.is_panic() {
-                            warn!(target: LOG_TASK, %err, type_name = %std::any::type_name::<T>(), "Jit value panicked");
-                            // Resume the panic on the main task
-                            panic::resume_unwind(err.into_panic());
-                        }
-                        #[cfg(not(target_family = "wasm"))]
-                        if err.is_cancelled() {
-                            warn!(target: LOG_TASK, %err, type_name = %std::any::type_name::<T>(), "Jit value task canceled:");
-                        }
-                        panic!("Jit value {} failed unexpectedly with: {}", std::any::type_name::<T>(), err);
```
elsirion, line 127: Took me way too long to find that diff :laughing: would have been nice if the pure refactoring was a separate PR, or at least if there was a comment pointing to the actual change :see_no_evil:

elsirion (approved):

fedimint-core/src/task/jit.rs
dpc: Oh. The whole outside thing had to change due needing to return some value. So there's no smaller way that would compile (that I could come up with).

Issue #5286, opened 2024-05-14
Investigate `No modules found of kind dummy` CI


bradleystachurski:
review club: dpc relayed that the problem was resolved by removing the diff that caused it. We should followup with an investigation to see what we were getting the `No modules found of kind dummy` error for the mint tests.

_Originally posted by @bradleystachurski in https://github.com/fedimint/fedimint/issues/5264#issuecomment-2110640119_

dpc:
"what we were getting"?

I'm kind of confused, but assuming this means "why"...

The test is very artificial and hacked together from different pieces. Client and federation have only one module.

The refactoring made is so federation is still down (all peers have API not started), while the test starts already executing.

In the test the Client is new, but it doesn't do a normal initialization, and gets config passed to it. So the first thing that actually runs in the initial API version discovery. That API version discovery fails, because connections to all peers fail. So it does what it seems reasonable - returns that it's unable to find version to use for all modules.

A big WARN logs are printed "Unable to discover blabhablah", "Disabling module dummy", etc. so it was immediately clear (to me) what is going on, but it does need to be immediately clear to everyone.

The test moves on, tries to get client dummy module to print some fake money, and panics because the module is not there.

The root cause is the APIs are down. Normally this would lead to client failing to get a config, but the test is hacked to skip this part. So version discovery fails loudly but doesn't really panic.

PR #5290, opened 2024-05-14
chore(client): retry api version discovery for at least 30 seconds

dpc:
During debugging with joschisan, we've spotted a case where in a test the Federation was not yet fully up (peer api servers were down), yet the test Client was already attempting to join with a pre-prepared config.

Normally in this case fetching the config would fail, but that was passed straight from the server fixture, so api discovery call was actually the first call. It queried all peers, all were down, the request logic would re-attempt connection but only once, which instantly failed as well, failing the whole initial api version discovery and disabling all client modules.

While in this case it seems like a problem of test failing to setup Federation before attempting to run client code, it makes me think that in real life there might be cases where the client is joining new Federation, gets a config, and then e.g. switches from wifi to GSM, etc. and has a brief period of being offline.

Again - this is only ever a problem on a first time client is joining. But to prevent such a hiccups, in this change, I'm making it so the client will attempt at least for a certain period to discover the version before giving it up, to lower the chances of such an unfortunate UX.


elsirion (approved):

elsirion:
Will this fix  #5286?

dpc:
> Will this fix https://github.com/fedimint/fedimint/issues/5286?

I believe something else already fixed it in @joschisan's PR. But if it had been there, it would probably also prevented it.

joschisan:
@dpc I did not fix this in general, just prevented this by awaiting the api  to come online.

dpc:
> @dpc I did not fix this in general, just prevented this by awaiting the api to come online.

Yeah. Waiting for api to come online was the reason of the problem. The client can't join federation if it's not online yet.

fedimint-client/src/lib.rs
```diff
@@ -1393,6 +1395,33 @@ impl Client {
...
+            if res.is_ok() {
+                return res;
+            }
+
+            if deadline < now() {
+                return res;
+            }
+            debug!(target: LOG_CLIENT, "Retrying getting api version from Federation");
+            sleep(Duration::from_secs(1)).await;
+        }
```
elsirion, line 1422: nit: #5323

elsirion (approved):

PR #5302, opened 2024-05-15
chore: update cargo-llvm-cov

dpc:
We were using one built ourselves to have newer version, but since then nixpkgs has an even newer version.

Before:

```
> cargo-llvm-cov llvm-cov -V
cargo-llvm-cov 0.5.31
```

After:

```
> cargo-llvm-cov llvm-cov -V
cargo-llvm-cov 0.6.9
```


elsirion (approved):

dpc:
Added a fix for macos.

elsirion (approved):

PR #5304, opened 2024-05-16
feat: variant index annotations for encodable enums

### LINKED ISSUE #5296: Manual index annotation for `Encodable` enums
As we add and remove enum variants of non-consensus enums over time it will often be easier to accept "holes" in the numbering scheme than writing DB migrations that increment/decrement every variant index by one.

For example:

```rust
#[derive(Encodable)]
enum MyEnum {
    Foo, // idx=0
    Bar, // idx=1
    Baz, // idx=2
}
```

is changed to

```rust
#[derive(Encodable)]
enum MyEnum {
    Foo, // idx=0
    Baz, // idx=1
}
```

now we'd have to write a DB migration that decodes the old `Baz` values with variant index 2 and re-encodes them with index 1. Instead we could allow the following annotation:

```rust
#[derive(Encodable)]
enum MyEnum {
    #[encodable_index = 0]
    Foo, // idx=0
    // note how idx=1 is missing
    #[encodable_index = 2]
    Baz, // idx=2
}
```

elsirion:
Fixes #5296. When removing the first variant of an enum we'd normally need to migrate all other variants to change their variant index. This is unnecessary work and we can just leave that index vacant by annotating the others with explicit indices.

### DIFF
--- a/fedimint-core/src/encoding/mod.rs
+++ b/fedimint-core/src/encoding/mod.rs
@@ -1335,6 +1335,45 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_custom_index_enum() {
+        #[derive(Debug, PartialEq, Eq, Encodable, Decodable)]
+        enum Old {
+            Foo,
+            Bar,
+            Baz,
+        }
+
+        #[derive(Debug, PartialEq, Eq, Encodable, Decodable)]
+        enum New {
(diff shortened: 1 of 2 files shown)


maan2003:
:tada: ACK for concept

dpc (approved):

PR #5305, opened 2024-05-16
feat: impl From for fedimint::Amount and bitcoin::Amount

benthecarman:
Found myself constantly converting between these types when implementing on-chain payments.


benthecarman:
nevermind looks like this already exists

Issue #5298, opened 2024-05-15
`NotEnoughSpendableUTXO`s error for mainnet federation with sufficient UTXOs


bradleystachurski:
A mainnet federation is getting a `NotEnoughSpendableUTXO` error when attempting to withdraw on-chain. I've spoken out-of-band with a few of the guardians and we've accumulated some clues:

- Federation on `v0.2.2`
- 3 successful peg-ins
- 2 successful withdrawals
- Subsequent withdrawal requests show `NotEnoughSpendableUTXO`
- Using `fedimint-observer`, we've identified the federation has 3 UTXOs with sufficient balance for subsequent withdrawal requests
- `fedimintd` logs show repeated `Broadcasting pending transactions`
- `bitcoind` backend running in pruned mode (550Mb)

Given this set of information, the current working theory is the wallet is unable to recognize change outputs from the previous two successful withdrawals. In the wallet server when we sync blocks up to consensus height, we call `get_tx_block_height` for all pending transactions each block.

https://github.com/fedimint/fedimint/blob/fa218de1641e2524b9307f8184d473adfc38067d/modules/fedimint-wallet-server/src/lib.rs#L999

If a peer is using `bitcoind` for a backend, this calls [getrawtransaction](https://developer.bitcoin.org/reference/rpc/getrawtransaction.html).

> By default this function only works for mempool transactions. ... When called without a blockhash argument, getrawtransaction will return the transaction if it is in the mempool, or if -txindex is enabled and the transaction is in a block in the blockchain.

This implies that if a peer is running `bitcoind` in pruned mode without enabling `txindex`, it will only query transactions stored in the mempool. Since the wallet syncs blocks up to consensus height once there's enough confirmations (10 blocks), the call to `get_tx_block_height` returns nothing since the transaction has been removed from the mempool for some time.

To test this theory, the guardians of this federation are going to attempt switching backends to esplora. If this is successful, this implies a material limitation to using a pruned bitcoind node for peers.

I'll update this issue as I learn more.

elsirion:
This looks like a potentially very severe bug. The `get_tx_block_height` approach sounds dangerous, I'd just get all transaction hashes of the block we are processing and then scan for inclusion of our change generating transactions. That would kill the electrum backend, but is anyone actually using it?

joschisan:
It seems to me that when the rpc call to get_tx_block_height happens to fail the change output is just not recognized.

bradleystachurski:
[...]

Issue #3513, opened 2023-11-03
Use `u64` instead of `u32` for block height


bradleystachurski:
Outstanding [TODO](https://github.com/fedimint/fedimint/blob/b9eeb4f807ccfe864057b9fa34237fe6555d9711/modules/fedimint-wallet/src/lib.rs#L1104) in codebase.

We have plenty of time until we run out of blocks using `u32` (approx. year 83667 if my back-of-the-envelope math is correct), however there are a few reasons why I agree it's worth cleaning up:
- Mixed usage of `u32` and `u64` results in multiple typecasts including an awkward [downcast](https://github.com/fedimint/fedimint/blob/68b3b774d48fd5e26251e051a58e30bfc6f56060/modules/fedimint-wallet-server/src/lib.rs#L856) in the consensus codepath
- The space savings in the db are negligible since we only store the most recent block count vote for each peer

dpc:
> The space savings in the db are negligible since we only store the most recent block count vote for each peer

Since we use varint encoding, there is not space savings. Both types encode to same values. And if you were to change them to `u64`, everything should still work, assuming no one will be running old/current versions in the year 84689.

PR #5307, opened 2024-05-17
fix: recognize change outputs for bitcoind backends without txindex

bradleystachurski:
Fixes https://github.com/fedimint/fedimint/issues/5298

Peers using bitcoind without txindex will not recognize change outputs. This PR introduces `is_tx_in_block` to `IBitcoindRpc`, which uses [`getblock`](https://developer.bitcoin.org/reference/rpc/getblock.html) for bitcoind that doesn't require txindex.

For pruned nodes, calls to `getblock` will return an error if the block data has been pruned (see: https://github.com/bitcoin/bitcoin/blob/v0.11.0/doc/release-notes.md#block-file-pruning). The minimum allowed for pruned nodes is 550MB, which is ~2 days/288 blocks assuming average block size ~1.9MB. If a peer attempts to sync blocks that have been pruned while there are outstanding pending transactions, the underlying call to `getblock` will fail.

---
This change will prevent further issues but does not resolve bad state for federations that have not recognized pending change transactions. I'm still investigating a simple fix along with ensuring we can use the recoverytool for impacted federations.

### DIFF
--- a/fedimint-bitcoind/src/bitcoincore.rs
+++ b/fedimint-bitcoind/src/bitcoincore.rs
@@ -100,6 +100,20 @@ impl IBitcoindRpc for BitcoinClient {
         Ok(height.map(|h| h as u64))
     }
 
+    async fn is_tx_in_block(
+        &self,
+        txid: &Txid,
+        block_hash: &BlockHash,
+        block_height: u64,
+    ) -> anyhow::Result<bool> {
+        let block_info = block_in_place(|| self.0.get_block_info(block_hash))?;
+        anyhow::ensure!(
+            block_info.height as u64 == block_height,
+            "Block height for block hash does not match expected height"
+        );
+        Ok(block_info.tx.contains(txid))
+    }
+
     async fn watch_script_history(&self, script: &ScriptBuf) -> anyhow::Result<()> {
         warn!(target: LOG_CORE, "Wallet operations are broken on bitcoind. Use different backend.");
         // start watching for this script in our wallet to avoid the need to rescan the
--- a/fedimint-bitcoind/src/electrum.rs
+++ b/fedimint-bitcoind/src/electrum.rs
@@ -108,6 +108,43 @@ impl IBitcoindRpc for ElectrumClient {
         }
     }
 
+    async fn is_tx_in_block(
+        &self,
+        txid: &Txid,
+        block_hash: &BlockHash,
+        block_height: u64,
+    ) -> anyhow::Result<bool> {
+        let tx = block_in_place(|| self.0.transaction_get(txid))
+            .map_err(|error| info!(?error, "Unable to get raw transaction"));
+
+        match tx.ok() {
+            None => Ok(false),
+            Some(tx) => {
+                let output = tx
+                    .output
+                    // use last since that's the change output we've constructed
+                    .last()
+                    .ok_or(format_err!("Transaction must contain at least one output"))?;
+
+                match block_in_place(|| self.0.script_get_history(&output.script_pubkey))?
+                    .iter()
+                    .find(|tx| tx.tx_hash == *txid && tx.height as u64 == block_height)
+                {
+                    Some(tx) => {
+                        let sanity_block_hash = self.get_block_hash(tx.height as u64).await?;
+                        anyhow::ensure!(
+                            *block_hash == sanity_block_hash,
+                            "Block height for block hash does not match expected height"
+                        );
+
+                        Ok(true)
+                    }
+                    None => Ok(false),
+                }
+            }
+        }
+    }
+
     async fn watch_script_history(&self, _: &ScriptBuf) -> anyhow::Result<()> {
         // no watching needed on electrs, has all the history already
         Ok(())
--- a/fedimint-testing/src/btc/mock.rs
+++ b/fedimint-testing/src/btc/mock.rs
@@ -346,6 +346,21 @@ impl IBitcoindRpc for FakeBitcoinTest {
         Ok(None)
     }
 
+    async fn is_tx_in_block(
+        &self,
+        txid: &Txid,
+        block_hash: &BlockHash,
+        block_height: u64,
+    ) -> BitcoinRpcResult<bool> {
+        let block = &self.inner.read().unwrap().blocks[block_height as usize];
+        assert!(
+            block.block_hash() == *block_hash,
+            "Block height for hash does not match expected height"
+        );
+
+        Ok(block.txdata.iter().any(|tx| tx.txid() == *txid))
+    }
+
     async fn watch_script_history(&self, _: &ScriptBuf) -> BitcoinRpcResult<()> {
         Ok(())
     }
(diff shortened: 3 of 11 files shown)


bradleystachurski:
@elsirion I'd like to do more investigation for how pruned nodes will behave before moving out of draft, but this is likely close if you want to take a quick look at the approach.

According to the docs, the min pruned value is 550MB, which is ~2 days/288 blocks assuming average block size ~1.9MB. If a peer attempts to sync blocks that have been pruned while there are outstanding pending transactions, the underlying call to `getblock` will fail.

bradleystachurski:
pr review club: out-of-scope for this pr, but it would be useful to not rely on network calls inside of process consensus item. For syncing blocks, this would mean creating an in-memory queue of blocks with txids, then when syncing a block after the finality delay we would read from the in-memory queue instead of relying on a network call.

A consideration for this approach is handling reorgs of the blocks in the in-memory queue. A naive implementation is to compare the block hash of the tip from bitcoind and the tip from the in-memory queue. If there's a divergence, simply reprocess the entire queue instead of trying to find the most recent ancestor.

fedimint-bitcoind/src/electrum.rs
```diff
@@ -108,6 +108,45 @@ impl IBitcoindRpc for ElectrumClient {
...
+                    .output
+                    // use last since that's the change output we've constructed
+                    .last()
+                    .ok_or(format_err!("Transaction must contain at least one output"))?;
+
+                let history = block_in_place(|| self.0.script_get_history(&output.script_pubkey))?;
+
+                // we check the earliest tx sent to our change address, since we can guarantee
+                // we constructed a withdrawal tx to our new change address but can't guarantee
+                // others have not sent subsequent transactions
```
dpc, line 133: I don't think we can guarantee anything about the ordering. Anyone could race with it with a higher fee.

dpc: Just `find` by `tx_hash` ? https://docs.rs/electrum-client/latest/electrum_client/struct.GetHistoryRes.html

fedimint-testing/src/btc/mod.rs
```diff
@@ -41,4 +41,12 @@ pub trait BitcoinTest {
 
     /// Waits till tx is found in mempool and returns the fees
     async fn get_mempool_tx_fee(&self, txid: &Txid) -> Amount;
+
+    /// Scans all transactions in all blocks
```
dpc, line 45: Wait, what?

dpc: I see. This comment might be just misplaced and should be on the impl?

fedimint-testing/src/btc/mock.rs
```diff
@@ -274,6 +275,17 @@ impl BitcoinTest for FakeBitcoinTest {
             return fee;
         }
     }
+
+    async fn get_tx_block_height(&self, txid: &Txid) -> Option<u64> {
+        self.inner
+            .read()
+            .expect("RwLock poisoned")
+            .blocks
+            .clone()
```
dpc, line 284: Why clone everything? That's going to take longer than the search itself.

dpc: Really, could add a new `tx_to_block_height: BTreeMap<TxId, BlockHeight>;`, maintaine it when adding blocks and be able to make a O(1) lookup here.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -974,11 +975,19 @@ impl Wallet {
...
-                .get_block_hash(height as u64)
+            let block_hash =
+                retry(
+                    "get_block_hash",
+                    FibonacciBackoff::default()
+                        .with_min_delay(Duration::from_secs(1))
+                        .with_max_delay(Duration::from_secs(10 * 60))
+                        .with_max_times(usize::MAX),
+                    || {
+                        self.btc_rpc.get_block_hash(height as u64) // TODO: use u64 for height everywhere
```
dpc, line 986: Why do we need u64? 4B blocks is going to get a while to get to?

dpc (commented):
Comments, but LGTM otherwise.

fedimint-bitcoind/src/electrum.rs
bradleystachurski: Ahh yes, that's a good point and I like your suggestion for using `find` to avoid the ordering issue.

fedimint-testing/src/btc/mod.rs
bradleystachurski: I see why that's confusing, thanks for pointing out. At the time it was true for all impls, but regardless doesn't belong in docstrings for a trait method.

fedimint-testing/src/btc/mock.rs
bradleystachurski: Good callout. I went with the `txid_to_block_height` approach and refactored the other method we're unnecessarily cloning all blocks (original inspiration that I didn't scrutinize).

modules/fedimint-wallet-server/src/lib.rs
bradleystachurski: Why so bearish on the longevity of the chain :smiley:

I'm preserving the original comment and created an issue a while ago that surfaces some of my reasoning why u64 makes sense. Feel free to push back on the issue.

https://github.com/fedimint/fedimint/issues/3513

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -974,11 +975,19 @@ impl Wallet {
 
             // TODO: use batching for mainnet syncing
             trace!(block = height, "Fetching block hash");
-            let block_hash = self
-                .btc_rpc
-                .get_block_hash(height as u64)
+            let block_hash =
+                retry(
```
bradleystachurski, line 979: Out-of-scope for this pr: should we consider making all calls to IBitcoindRpc retrying internally?

fedimint-bitcoind/src/bitcoincore.rs
```diff
@@ -100,6 +100,20 @@ impl IBitcoindRpc for BitcoinClient {
...
+        txid: &Txid,
+        block_hash: &BlockHash,
+        block_height: u64,
+    ) -> anyhow::Result<bool> {
+        let block_info = block_in_place(|| self.0.get_block_info(block_hash))?;
+        anyhow::ensure!(
+            block_info.height as u64 == block_height,
+            "Block height for block hash does not match expected height"
+        );
+        Ok(block_info.tx.contains(txid))
```
elsirion, line 114: This is slightly inefficient if we are waiting for multiple transactions since the block is being fetched multiple times. For now this isn't a problem and the alternative of just returning the transaction ids from the block wouldn't work with the electrum backend.

elsirion (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-5307-to-releases/v0.3 origin/releases/v0.3
cd .worktree/backport-5307-to-releases/v0.3
git switch --create backport-5307-to-releases/v0.3
git cherry-pick -x 6e8290f6cd6c035ef2c2b281086d9be2bdf99e27 b4090b789b09a2f9f7c581283d84f432aa2958ce d9c7e5a204c5f6d43764e895393e15ad733d3e4d
```

PR #5312, opened 2024-05-17
chore: reenable devimint gateway liquidity channel opening

tvolk131:
I audited the code for [`open_channel()`](https://github.com/fedimint/fedimint/blob/5bd3555a5bf445def3f44f8b1a9049299e7f956b/devimint/src/external.rs#L552) to ensure that [`open_channel_between_gateways()`](https://github.com/fedimint/fedimint/blob/5bd3555a5bf445def3f44f8b1a9049299e7f956b/devimint/src/external.rs#L654) matches it 1-to-1. I then re-enabled `open_channel_between_gateways()` for devimint (which was disabled [here](https://github.com/fedimint/fedimint/pull/5149) due to being slow and flaky). I measured the startup time for `just mprocs` 10+ times on my local machine before and after the change and with the changes in this PR it never failed and the time difference was within margin of error.

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -584,9 +584,12 @@ async fn open_channel(
     let cln_pubkey = cln.pub_key().await?;
 
     cln.request(cln_rpc::model::requests::ConnectRequest {
-        id: lnd_pubkey.parse()?,
-        host: Some("127.0.0.1".to_owned()),
-        port: Some(process_mgr.globals.FM_PORT_LND_LISTEN),
+        id: format!(
+            "{}@127.0.0.1:{}",
+            lnd_pubkey, process_mgr.globals.FM_PORT_LND_LISTEN
+        ),
+        host: None,
+        port: None,
     })
     .await
     .context("connect request")?;
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -19,8 +19,7 @@ use hex::ToHex;
 use tokio::fs;
 use tokio::sync::{MappedMutexGuard, Mutex, MutexGuard};
 use tokio::time::Instant;
-use tonic_lnd::lnrpc::policy_update_request::Scope;
-use tonic_lnd::lnrpc::{ChanInfoRequest, GetInfoRequest, ListChannelsRequest, PolicyUpdateRequest};
+use tonic_lnd::lnrpc::{ChanInfoRequest, GetInfoRequest, ListChannelsRequest};
 use tonic_lnd::Client as LndClient;
 use tracing::{debug, info, trace, warn};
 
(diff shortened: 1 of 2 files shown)


devimint/src/external.rs
```diff
@@ -711,16 +692,21 @@ pub async fn open_channel_between_gateways(
...
-    gw_cln
-        .open_channel(lnd_pubkey.clone(), 10_000_000, Some(5_000_000))
-        .await?;
+    poll("fund channel", || async {
+        gw_cln
+            .open_channel(lnd_pubkey.clone(), 10_000_000, Some(5_000_000))
+            .await
+            .map_err(ControlFlow::Continue)
+    })
+    .await?;
```
elsirion, line 701: What are we waiting for here? Why would it fail and later work? Maybe we want to wait for the node being up instead?

tvolk131: We're polling here because that's what the existing `open_channel()` devimint function does: https://github.com/fedimint/fedimint/blob/e694b7e4d18b4815d4dc27d37db5b6789fcd75b6/devimint/src/external.rs#L582-L606

I added a TODO comment to remove the polling in `open_channel_between_gateways()`. Do you mind if I address that in a follow-up?

dpc (approved):

dpc:
Will revert if it is causing issues. :shrug:  :rofl:

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -638,7 +638,7 @@ impl GatewayLightning for ClnRpcService {
             .await
             .map_err(|e| Status::internal(e.to_string()))?
             .call(cln_rpc::Request::NewAddr(model::requests::NewaddrRequest {
-                addresstype: Some(cln_rpc::model::requests::NewaddrAddresstype::BECH32),
+                addresstype: None,
```
dpc, line 641: This is the only change here that isn't dev-only. Is this OK? @tvolk131

elsirion:
MQ found some LN test problems, worth looking into before kicking it back into the MQ imo.

dpc:
image

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: I changed `addresstype` to `None` to match the existing behavior in `open_channel()`: https://github.com/fedimint/fedimint/blob/3617b966081793198b46e56443fa2f78f506788a/devimint/src/external.rs#L575

I don't think this will likely matter, but I wanted to make sure that the implementations of `open_channel()` and `open_channel_between_gateways()` match as closely as possible.

tvolk131:
Hmmm most of the CI errors are some form of the following out-of-disk-space:

fedimint-workspace-lcov-ci> 2024-05-22T22:56:47.489529Z ERROR connect_fed{payload=ConnectFedPayload { invite_code: "fed11qgqpw9thwvaz7te3xgmjuvpwxqhrzw33xymnsdf0qqqjqmlee0af0aap6fyvmslxhn94k9tsms6wtel7sq32z747qdhra0uuej7vxh" }}: ln_gateway::rpc::rpc_server: error=Failed to open the database: Error opening rocksdb: IO error: No space left on device: While appending to file: /build/test-14034218261564233250GBHOqh/gateway-16131950401517601773/6ff9cbfa97f7a1d248cdc3e6bccb5b1570dc34e5e7fe8022a17abe036e3ebf9c.db/000000.dbtmp: No space left on device

Whatever's causing this flakiness appears to be outside of the control of the code being changed in this PR.

PR #5313, opened 2024-05-17
chore: do not require identical git hash during DKG

dpc:
That is waay to constraining. Different peers can't even include non-code fixes for their local deployement and other problems.

Since we're kind of good at the release and versioning and stability things, change the requirement to a `code_version_str`, which consists of the cargo release version of Fedimint code and optional vendor suffix.

### DIFF
--- a/fedimintd/src/fedimintd.rs
+++ b/fedimintd/src/fedimintd.rs
@@ -151,7 +151,7 @@ fn parse_map(s: &str) -> anyhow::Result<BTreeMap<String, String>> {
 /// // Note: not called `main` to avoid rustdoc executing it
 /// // #[tokio::main]
 /// async fn main_() -> anyhow::Result<()> {
-///     Fedimintd::new(env!("FEDIMINT_BUILD_CODE_VERSION"))?
+///     Fedimintd::new(env!("FEDIMINT_BUILD_CODE_VERSION"), Some("vendor-xyz-1"))?
 ///         // use `.with_default_modules()` to avoid having
 ///         // to import these manually
 ///         .with_module_kind(WalletInit)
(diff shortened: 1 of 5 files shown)


douglaz:
Yeah, it makes sense to allow different git versions.

But now it's not clear if we support the following scenario:

So I create a new fedimintd with a new module and try to setup with the vanilla fedimintd (perhaps accepting the standard modules). Will this ceremony work?

dpc:
> So I create a new fedimintd with a new module and try to setup with the vanilla fedimintd (perhaps accepting the standard modules). Will this ceremony work?

Nope. As things are right now, we don't have any way for peers to vote/agree on a common set and composition of modules, so they have to be exactly the same everywhere.

Maybe one day thought. Then we can remove the whole requirement, and each module will be negotiated independently etc.

fedimintd/src/fedimintd.rs
```diff
@@ -205,7 +218,10 @@ impl Fedimintd {
             bitcoind_rpc,
             server_gens: ServerModuleInitRegistry::new(),
             server_gen_params: ServerModuleConfigGenParamsRegistry::default(),
-            version_hash: version_hash.to_owned(),
+            code_version_hash: code_version_hash.to_owned(),
+            code_version_str: code_version_vendor_suffix
```
elsirion, line 222: Instead of having this awkward `vendor_suffix`, why not serialize the available module kinds in ascending order? That should achieve 95% of what we want here and won't require an API change, so we could even backport.

dpc: @elsirion 

Hmm... That would probably work, but conceptually... is it only about module kinds?

If there's a vendor xyz, and it provides some tweaked `fedimintd`, wouldn't we want it to enforce that all guardians run the version from the same vendor?

The `vendor_suffix` is also useful in other places. If I run `fedimintd --version` etc. I'd like to see `fedimint (v0.4.5-fedi)` etc. So I don't think I'd want to get rid of "awkward vendor_suffix" anyway.

elsirion: I'm ok going with this version, but you kinda have to trust vendors to actually set this. Maybe we could combine both?

elsirion: Feel free to merge or squash in 5e85052b37dbfbe901e8ea05c21fd6fc04e64061

elsirion (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-5313-to-releases/v0.3 origin/releases/v0.3
cd .worktree/backport-5313-to-releases/v0.3
git switch --create backport-5313-to-releases/v0.3
git cherry-pick -x 0800e73e7641d488e14e02294574ff489e3e56e2
```

dpc:
I don't think we really need to backport it. People lived with it since now, so they can wait ~2 months for 0.4.

PR #5322, opened 2024-05-18
feat: private federations

dpc:
### Motivation

Running public federation has multiple risk, some of which are related to the sheer amount of usage it might bring.

As mints are perfectly private, there is no way to tell the users apart and control the usage.

However the mint can agree on a shared secret and circulate it among the intended users. Imagine a handful of families or a small community running federation for themselves and their friends.

The users and the guardians semi-know and trust each other, and would like to avoid strangers using their federation, even if just to limit resource usage. They would like to keep thing private between themselves though.

The scheme is as follows: guardians agree on a shared api secret. Api secret is enforced on HTTP level. From the outside perspective, public Federation endpoints appear as just inaccessible HTTP servers, and there's little to no sign that they are running Fedimint.

Guardian share the secret with intended users. Users can share it further if they want - there is no way for guardians to control it. But the expectation is that the users will have an incentive to keep it private.

On top if - guardians at any time (possibly periodically) can rotate the secret. E.g. if they detect unusally high activity, they can generate a new secret, circulate it with intended users again and thus shut-off the access for the non-intended users.

If the abuse persist, they can even rotate the secret and have the users get the secret one by one to reclaim their money, ultimately confiscating funds of people that were not supposed to be used.

Just this possibility alone should scare off strangers using private Federations.

Note: this does not ruin privacy of the intended users.

### Technical details

Secret are enforced on HTTP level using Basic http authorization scheme.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#basic

This makes private mints look mostly like normal http servers. Similiar mechanism can be done using Nginx config, but having it inside `fedimintd` allows better integration with Invite Codes, testing, etc.

While working on it it occured to me that I wish it was possible to authenticate guardian APIs on the HTTP level as well, but it seems that currently there's no way to pass any data from jsonrpsee http middleware to the jsonrpc middleware.

Each `fedimintd` can optionally set one main `api_secret`, and multiple `api_extra_secrets`. The extra secrets are intended for phasing out secrets and allowing users who did not receive the new secret yet, to continue using the existing one.

Invite codes now carry a new field that contains the secret.

OOB Notes carry the api secret as well. It's unclear if not doing it would be a better idea. Then sending OOB would not consist being an invitation.

After joining private federation `Client` stores the `api_secret` from the invitation in the database, just like the config. In the future we can add cli and methods to update it, e.g. when receiving newer invitiation. This will allow end-user app integrators to e.g. update secret by re-invitation.

The changes are mostly simple, but touch a lot of places, as the `api_secret` needs to be passed from the the cli options all the way to api handlers and clients.

### Testing

```
env FM_DEVIMINT_CMD_INHERIT_STDERR=1 RUST_LOG=fm=debug FM_API_SECRET="asfasdfasdfasdf" j devimint-env
```

will make everything (both fedimintd and fedimint-cli and other clients) use the api secret.

If you find in the logs api addresses used by one of the peers you can check that public access is not allowed with:

```
devimint> curl -i http://127.0.0.1:14172/
HTTP/1.1 401 Unauthorized
www-authenticate: Basic realm="Authentication needed"
content-length: 12
date: Sun, 19 May 2024 06:33:48 GMT

Unauthorized
```


fedimint-api-client/src/api.rs
```diff
@@ -1305,25 +1316,40 @@ impl<C: JsonRpcClient + Debug + 'static> IRawFederationApi for WsFederationApi<C
...
+            .max_concurrent_requests(u16::MAX as usize);
 
         #[cfg(target_family = "wasm")]
-        WsClientBuilder::default()
-            .max_concurrent_requests(u16::MAX as usize)
-            .build(url.as_str())
-            .await
+        let mut client = WsClientBuilder::default().max_concurrent_requests(u16::MAX as usize);
+
+        if let Some(api_secret) = api_secret {
```
dpc, line 1340: This is how client authorizes the request.

fedimint-core/src/invite_code.rs
```diff
@@ -94,6 +105,13 @@ impl InviteCode {
             .expect("Ensured by constructor")
     }
 
+    /// Returns the API URL of one of the guardians.
```
dpc, line 108: TODO: incorrect copy-pasted docstring

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -104,6 +104,7 @@ enum OOBNotesData {
         peer_apis: Vec<(PeerId, SafeUrl)>,
         federation_id: FederationId,
     },
+    ApiSecret(String),
```
dpc, line 107: TODO: should sending OOB notes pass the secret?

dpc:
```
       > error[E0599]: no method named `set_headers` found for struct `WasmClientBuilder` in the current scope
       >     --> fedimint-api-client/src/api.rs:1350:29
       >      |
       > 1350 |             client = client.set_headers(headers);
       >      |                             ^^^^^^^^^^^ method not found in `WasmClientBuilder`
```

Damn it wasm, why?

I think I can workaround it by just modifying the URL, which the underlying browser should translate to the same thing. Will look into it when I can.

maan2003:
ACK concept

should the federation invite contain the secret?
Then to update the secret on client side you ask the user to scan the invite qr code

dpc:
> I think I can workaround it by just modifying the URL, which the underlying browser should translate to the same thing.

AFAICT this should work, but I don't have a good way to test it RN.

elsirion:
> Each `fedimintd` can optionally set one main `api_secret`, and multiple `api_extra_secrets`

Why not just one list of secrets? 

> Invite codes now carry a new field that contains the secret.

Is that a good idea? I think it's better if the secret has to be communicated out of band, otherwise users might be surprised when it gets rotated. When entering the password apps can warn users of the risk of being locked out if they don't know someone in the inner circle.

fedimintd/src/fedimintd.rs
```diff
@@ -94,6 +95,24 @@ pub struct ServerOpts {
...
+    #[arg(env = FM_API_SECRET_ENV)]
+    api_secret: Option<String>,
+
+    /// Comma separated list of additional secrets that can be used to access
+    /// the Api
+    ///
+    /// Format: `secret1,secret2,...`
+    #[arg(env = FM_API_EXTRA_SECRETS_ENV, value_parser = parse_vec, default_value="")]
+    // Note: full path used due to https://github.com/clap-rs/clap/issues/4626#issuecomment-1593681582
+    api_extra_secrets: std::vec::Vec<String>,
```
elsirion, line 114: Why make this two options :sob:

elsirion (commented):
How do we communicate to the user/integrator when auth fails and needs to be renewed?

modules/fedimint-mint-client/src/lib.rs
dpc: @elsirion I think invites should. The apps should probably just display a warning etc. when sharing invite-only Federation.

Secrets should ideally be long and without it the UX will suffer.

And if someone really wants to share Federation publically, can just paste both the invite and secret so no point in being too bent on trying to prevent it.

fedimintd/src/fedimintd.rs
dpc: The peer uses an Api to talk with other peers. So there's a difference between "one I accept and use" and "one I just still accept". @elsirion

dpc:
> Is that a good idea? I think it's better if the secret has to be communicated out of band, otherwise users might be surprised when it gets rotated.

Prevents nothing, UX suffers.

modules/fedimint-mint-client/src/lib.rs
dpc: On the OOB notes I'm a bit torn. Seems like there are some fields on OOB notes to kind-of-contain an invite.

dpc:
> How do we communicate to the user/integrator when auth fails and needs to be renewed?

In case of bad auth the requests return 401 error code. It comes down to error handling and how to structure it to bubble up all the way to end app ux.

modules/fedimint-mint-client/src/lib.rs
elsirion: What would the UX look like for password changes? I imagined the password to be something easy to communicate `thereisnosecondbest`, `enjoyyourpizza`, … so when people meet they can update each other verbally. Realistically who'd spend the energy to brute-force their way into a federation? If it's a long/complicated password you'll have to go through the invite QR scanning procedure again I guess?

fedimintd/src/fedimintd.rs
elsirion: So would you want to communicate a warning to the client if they connect via an extra secret? Otherwise how does behavior differs between the two? If behavior doesn't differ, why not just leave it to the operators to put the active one first so they remember which ones to phase out?

dpc: > If behavior doesn't differ, why not just leave it to the operators to put the active one first so they remember which ones to phase out?

It could be done, but I thought having distinct fields will be self-describing and bring the meaning and function to the user.


Possibly the naming could be improved. Should it be `FM_ACTIVE_API_SECRET` and `FM_STALE_API_SECRETS`?

> So would you want to communicate a warning to the client if they connect via an extra secret?

I thought about it, but I don't see any easy way to do it, so I left it out for now. We could possibly add another API call that end user app could periodically make to verify and warn the user?

@elsirion

dpc:
Rebased.

@elsirion If there is not fundamental dissagrement about the idea, can we land it? This change is merge conflict magnet.

Things to argue about that we can still change:

* API secrets sharing UX: embedded in invites vs sharing passwords
* API secrets in OOB Notes

Things I think need to change anyway:

* we should probably store secrets in the DB, and switch to `DEFAULT_API_SECRET...` and `FORCE_API_...` env vars. Just like we now do bitcoind rpc. This way it's possible to manage secrets via DB, but still allow (via _FORCE_) to ignore/overwrite the DB value in tests and setups that don't rely on UI so much.

fedimint-server/src/net/api/http_auth.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn new(secrets: Vec<String>) -> Self {
+        if secrets.is_empty() {
+            info!(target: LOG_NET_AUTH, "Api available for public access");
+        } else {
+            info!(target: LOG_NET_AUTH, num_secrets = secrets.len(), "Api available for private access");
+        }
+        Self {
+            auth_base64: secrets
+                .iter()
+                .map(|p| STANDARD.encode(format!("fedimint:{p}")))
```
elsirion, line 34: The `fedimint:<secret>` format will allow us to eventually expand the supported auth schemes to blinded tokens etc.

elsirion (approved):
Should probably also migrate admin auth to this scheme. Could use `guardian:<secret>` to distinguish.

PR #5324, opened 2024-05-20
chore: define feature process

elsirion:
For discussion @dpc @bradleystachurski @m1sterc001guy @maan2003

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,24 @@ If you have any questions or need assistance, don't hesitate to reach out to us 
 
 Happy Hacking!
 
+# Feature Process
+
+Fedimint is a project with high inherent complexity, maintained by relatively few people and thus needs to avoid accidental complexity through unnecessary requirements. The following process aims to minimize feature creep:
+
+Any changes that cannot be easily reverted because they touch consensus, public APIs, etc. should go through a design phase. For that please open an issue describing:
+  * The minimum requirements and why they are necessary
+    * Possible additional requirements and why they are useful enough to justify the added complexity
+  * The design space you explored
+  * The design you consider ideal
+
+Two contributors familiar with the area should ACK before proceeding with an implementation. Feature creep should be avoided both in the design phase as well as when implementing.
+
+## Code/Requirement Ownership
+**Please consider yourself personally accountable for the requirements your implementation introduced - you should be able to explain and justify requirements on request.** Some of them might be implicit, especially when introduced by a bug fix and not through the feature process. Please avoid introducing requirements "just in case" and strive for a clean and straightforward design that is easy to maintain and may still be extended at a later point if it becomes necessary. Please try to keep track of your prior work as the project progresses since requirements may have become redundant.
+
+**Furthermore, question existing requirements and consider removing them if nobody can explain well why they exist;** if a removal needs to be reverted it does not mean that you made a mistake, at least now we are aware why it is necessary and can record this in a comment. If you never have to put something back in you are not removing enough code.
+
+Do not optimize code that should not exist in the first place. **Premature optimization is the root of all evil.**
 
 # Code Review Policy
 


CONTRIBUTING.md
```diff
@@ -53,6 +53,24 @@ If you have any questions or need assistance, don't hesitate to reach out to us
 
 Happy Hacking!
 
+# Feature Process
+
+Fedimint is a project with high inherent complexity, maintained by relatively few people and thus needs to avoid accidental complexity through unnecessary requirements. The following process aims to minimize feature creep:
+
+Any changes that cannot be easily reverted because they touch consensus, public APIs, etc. should go through a design phase. For that please open an issue describing:
```
dpc, line 60: Public APIs can be relatively easily fixed over time.

dpc:
The whole things seems a bit accusionary, like we did some crappy job so far, intentionally incremented scope, not designing well, and now we need to go through some hoops to make sure someone smarter than previous contributors make sure dummies don't make more mistakes.

I think some people when finding things about our codebase or designs in it that are less then ideal, might overlook the fact we've been scrambling to incrementally build enough of what was most severely lacking to be able to move on the next part that was lacking the most. A step at the time, discovering the problem space in the process, as some thing being completely wrong are only apparent when you have everything around them ready and exposing their imperfections. And that allowed us accomplish a lot with very little.


The reality of the situation is that we are trying to build very, very ambitious project with very few people with various experiences and skills - around 4-5 people working on the project at the time, many sharing other responsibilities (like e.g. building Fedi). And if we cut the scope too much it just not going to be very appealing, won't gain traction, there will be no one willing to invest resources into it, and will fizzle out.


So there just wasn't a lot of room for debating designs. Some stuff is implemented this or that way, because there was one person that had the time to do it, and they did it best as they could given the time, needs at the time, and what they knew.


If we were doing stuff with elaborated rounds of design then we would still be discussing TDD for the module system. Instead we had a rough idea what to aim for, trust in people who worked on it, some brief discussions and bursts of arguing about stuff, multiple rounds of incremental changes by different people, and eventually arrived in a happy place. I don't think it would have went any better, and definitely not faster, if we were debating longer, writing documents, etc.


Similarly - all the other parts of the project carried us well so far, even if they were often imperfect. An imperfect, but already working part is 100x better than a perfect design delayed into the future.

We had always enough foresight to always avoid painting ourselves into a corner, and leaving room for eventually moving on to better designs, architectures and implementations, precisely because we knew we can't get everything right or ready immediately.


Given all this, I think we've did a great job so far and I want to congratulate and appreciate everyone that were involved in this journey so far. We have a robust codebase, capable architecture, solid infra around it, lot of parts in a great shape, and vision for improvements in things that are lacking, with a strategy how to get there.


I don't think now is the time to fundamentally change our approach. Since there are so few of us, it's hard sometimes to get a basic PR approved. How are we going to find time to review and go back and forth on elaborated design documents, or gate stuff until limited amount of overworked people find the time to approve every design.


> Any changes that cannot be easily reverted because they touch consensus,

We rarely (never, except for alephbft transition) mess with core consensus. And modules can and will inevitably be re-implemented, in a same way lnv2 and mintv2 are shaping. And it's great. Early federations will get recreated, old implementations will be rotated out and deleted.

> public APIs

can be relatively easily rotated, and it's not even all that hard.

And for all of it we can have solid tests and safekeeping mechanism, and ideas for more of them.


At present time, I think our bottleneck are not imperfect designs, but still ability to implement all the stuff we need.


With time all designs will naturally mature, and we'll hopefully have more time to be more deliberate about everything. And making things actually better will require more discussions and design. Currently, I don't need to think much too hard e.g. about client wallet recovery, because there just isn't any. So basically any design that works would be better.

elsirion:
> The whole things seems a bit accusionary

I'm sorry that it came off in that way, I probably should have waited with posting till I had some more time to preface/give context and not in a hurry while @joschisan and I had a few hours after a conference.

I agree that we did a great job so far given the constraints, but also see how the code base has grown a lot, to a point that I can't keep it in my head anymore and reason about it holistically. At the same time we are reaching high-level feature-completeness: there are Federations out there that work (this assumption might be the root of our difference of opinion).

This doesn't mean we are done, there is tons of work left: testing, fixing bugs, better logging, incorporating feedback from the real world, DoS protection, cleaning up the code that got us so far but could be improved, …. Imo in many cases we do have the time now to take a step back and consider different approaches because things already mostly work.

We did have design/architecture discussion before implementation in the past, e.g. you kept me from going with a very rigid backup/recovery architecture, that would have had some limitations. I think we already follow a decent process most of the time, some would just like to codify it to have more of a ground to say "hold on a moment" when sometimes there's a bit of a rush to get things shipped.

I absolutely think our focus should shift from features to maintainability over time. That won't be a sudden shift and I don't feel strongly about codifying it, but the times of freely breaking APIs with every commit are (un)fortunately over (research projects are fun, but live systems change the world), so each breaking change requires a lot of migration code around it that we are stuck with for a considerable amount of time and that increases complexity. Given a limited and imo close-to-exhausted complexity budget that means being more deliberate with big changes.

I actually think we mostly agree on that reading your last paragraph again, just that we see the project at different stages of maturity.

m1sterc001guy:
As far as this PR goes, I don't have any problem with adding a disclaimer around feature designs that touch consensus or more public facing area of Fedimint. Core consensus changes are likely going to be rare going forward. 

I think where there is a bit of a grey area is around module design/implementation. Modules _use_ consensus but only interact with it through the exposed APIs. For module developers (especially external to the core fedimint repo), I definitely don't think they should need to go through a design process, get sign-off/advice from the Fedimint developers, just so we can verify that a module is not abusing the consensus APIs that are exposed, before implementing something. It's certainly possible to write a poor module, and if we see people doing that, we should take it as an opportunity to improve our own docs/APIs.

For changes internal to the Fedimint repo, I don't think merging this change does any harm, but I also agree with @dpc that introducing more formal process and structure around our design process isn't necessary right now. I've certainly worked on projects at big companies where the process for changing almost any minute detail in the code requires a functional doc, design doc, sign off from PMs, etc. A lot of the time when the process is all done we could have just written the code and seen if it worked or not. 

> I think some people when finding things about our codebase or designs in it that are less then ideal, might overlook the fact we've been scrambling to incrementally build enough of what was most severely lacking to be able to move on the next part that was lacking the most.

I'd like to second this point, it has certainly been my experience. With time comes increasing scope, and with increasing scope, sometimes the previous design is no longer ideal. I don't think its a bad thing, its a sign of an evolving project. Going forward, I don't think its possible to design for the full range of things that we might want to support, so this will always be an iterative process, and along the way we'll need to decide if its better to support in a backwards compatible way, or re-write a module like we are doing for LN and the mint (and probably soon the wallet :))

Issue #5334, opened 2024-05-21
in wild: overwritting a OperationLogKey


maan2003:
Database overwriting element when expecting insertion of new entry. 
Key: OperationLogKey { operation_id: OperationId(98fac40fb39916eea29bd8ad343ffbafccc76f8929a934d7568747920c49e055) }
Prev Value: OperationLogEntry { .... }

dpc:
We should probably make this debug-mode assertion/panic so we can catch all of these in the CI.

dpc:
I kind of doubt it's worth doing more about it. The problem might not be there already in 0.4 :crossed_fingers: . If it is, maybe we'll spot it in a failed CI run.

It's possible it was caused by snapshot level isolation bug that was recently fixed. https://github.com/fedimint/fedimint/pull/5284/files#diff-346ebc4a765d21fac9b2f913145a5c8e347bb943f3c4dbd34e524ab68efd846cR143

That would be my suspicion, as that operation (insert/insert_new) was affected.

So I'm closing.

PR #5336, opened 2024-05-21
chore: debug assert for overwrite existing values in db

maan2003:
ref #5334


elsirion (approved):
I wanted to complain that we should do the same for `remove_existing_entry`, but apparently it doesn't exist? Did we remove that at some point?

PR #5337, opened 2024-05-22
refactor: introduce TransactionSubmissionOutcome newtype

joschisan:
We introduce a newtype to to highlight the similarities to OutputOutcome.


elsirion (approved):
"Don't worry about this, API layer!"

PR #5338, opened 2024-05-22
chore: print shortened version of OperationId when logging

dpc:
`operation_id` is soo common and soo long in client side logging that it's just annoying.

Shorten it to `abcdefgh..abcdefgh` version.

To prevent unintentional mistakes, remove `Display` from `OperationId` and require going through `fmt_short` or `fmt_full` to declare intention.

Fix all the call-sites.

Before:

image

After:

image

### DIFF
--- a/fedimint-core/src/core.rs
+++ b/fedimint-core/src/core.rs
@@ -58,6 +58,9 @@ pub mod backup;
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Encodable, Decodable, PartialOrd, Ord)]
 pub struct OperationId(pub [u8; 32]);
 
+pub struct OperationIdFullFmt<'a>(&'a OperationId);
+pub struct OperationIdShortFmt<'a>(&'a OperationId);
+
 impl OperationId {
     /// Generate random [`OperationId`]
     pub fn new_random() -> Self {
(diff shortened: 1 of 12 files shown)


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -807,9 +807,9 @@ pub enum SendPaymentError {
     InvoiceMissingAmount,
     #[error("The invoice has expired")]
     InvoiceExpired,
-    #[error("A previous payment for the same invoice is still pending: {0}")]
+    #[error("A previous payment for the same invoice is still pending: {}", .0.fmt_full())]
     PendingPreviousPayment(OperationId),
-    #[error("A previous payment for the same invoice was successful: {0}")]
+    #[error("A previous payment for the same invoice was successful: {}", .0.fmt_full())]
     SuccessfulPreviousPayment(OperationId),
```
maan2003, line 813: why full here, but ...(see fqsjld)

dpc: This one looked like an error that could cross API-level boundary, while the other one like just some random ad hoc anyhow thing being yeeted on error.

No strong feelings. Can change the other one to long.

fedimint-core/src/core.rs
```diff
@@ -70,17 +73,33 @@ impl OperationId {
...
+    pub fn fmt_full(&self) -> OperationIdFullFmt {
+        OperationIdFullFmt(self)
+    }
+}
+
+impl<'a> Display for OperationIdShortFmt<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        bitcoin29::hashes::hex::format_hex(&self.0 .0[0..4], f)?;
+        f.write_str("_")?;
+        bitcoin29::hashes::hex::format_hex(&self.0 .0[28..], f)?;
```
maan2003, line 89: do we even care about the suffix?

dpc: Good question. I kind of wanted to make sure that the whole thing is longer and just got shortened, I guess, so "removing the middle" seemed to make sense. And if people are going to compare, they'll visually match beginning or the end. But no strong feelings about it. I guess `abcdefgh_` or even just `abcdefgh` would work.

maan2003: I think prefix is enough, we just want different ids for different operations

dpc: Are `OperationId`s always hashes thought? Could there be a case where lots of them have the same prefix for some reason? @maan2003

maan2003: I think they are always hashes now that recovery is not a state machine. But yes external module developer can implement it however they like

maan2003: then it is fine with prefix and suffix

maan2003 (approved):

PR #5341, opened 2024-05-22
chore: cleanup rust code (except for modules)

tvolk131:
I temporarily enabled clippy pedantic rules and fixed the issues it caught (except the ones that sacrifice code readability)


elsirion:
CI failure was unrelated, but needs rebase now.

tvolk131:
Markdown check failed because of a nix bug (see #5379) but everything else looks good!

fedimint-server/src/net/queue.rs
```diff
@@ -51,12 +51,7 @@ where
 
     pub fn ack(&mut self, msg_id: MessageId) {
         debug!("Received ACK for {:?}", msg_id);
-        while self
-            .queue
-            .front()
-            .map(|msg| msg.id <= msg_id)
-            .unwrap_or(false)
-        {
+        while self.queue.front().is_some_and(|msg| msg.id <= msg_id) {
```
elsirion, line 54: TIL

elsirion:
@dpc @joschisan How much value do you see in these kinds of cleanups? On the one hand, there are some really neat changes in there, otoh it's a lot of churn and not enforced in CI, so will deteriorate again over time.

elsirion:
Maybe we can enforce this in CI and just `#[ignore(clippy::…)]`?

dpc:
> Maybe we can enforce this in CI and just `#[allow(clippy::…)]`?

If there are some clippy lints that we find valuable that are not enabled be default, we should enable them, sure @tvolk131 @elsirion :+1:

tvolk131:
> Maybe we can enforce this in CI and just `#[allow(clippy::…)]`?

Done! See #5389

PR #5347, opened 2024-05-23
feat: add expert-user script for activating the meta module

elsirion:
Federations deployed prior to 0.3.0 do not have the meta module in their config, meaning it doesn't get deactivated even though >=0.3.0 fedimintds support it. We currently don't support adding modules after the fact, but in case of the meta module we can kinda hack it in since it doesn't need any config, which is what the added script does.

### DIFF
--- a/scripts/ops/add_meta_module.sh
+++ b/scripts/ops/add_meta_module.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+REQUIRED_BINS=(jq fedimint-cli)
+for REQUIRED_BIN in "${REQUIRED_BINS[@]}"; do
+    if ! which "$REQUIRED_BIN" &> /dev/null; then
+        echo "$REQUIRED_BIN is not installed."
+        exit 1
+    fi
+done
+
+if [ -z "${1:-}" ]; then
+    echo "Error: No directory specified."
+    echo "Usage: $0 <data_directory>"
+    exit 1
+fi
+DATA_DIR=$1
+
+echo "This script will add the meta module to federation config, which is technically"
+echo "a consensus-breaking change. Make sure that:"
+echo "  * All fedimintd instances are shut down"
+echo "  * You run this script on all of them"
+echo "  * Only restart once all are patched"
+echo "  * Let them run for a bit after restarting before voting on meta module"
+echo "    consensus proposals"
+echo ""
+echo -e "\e[1mIt's like open heart surgery on a federation, please only continue if you know\e[0m"
+echo -e "\e[1mwhat you are doing and can debug Fedimint problems yourself.\e[0m"
+echo ""
+echo -n "Do you wish to continue? (yes/no): "
+read response
+
+if [[ "$response" != "yes" ]]; then
+    echo "Operation aborted by the user."
+    exit 1
+fi
+
+exists=$(jq '.modules | to_entries | any(.value.kind == "meta")' "$DATA_DIR/consensus.json")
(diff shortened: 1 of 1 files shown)


scripts/ops/add_meta_module.sh
```diff
@@ -0,0 +1,55 @@
...
+
+if [[ "$response" != "yes" ]]; then
+    echo "Operation aborted by the user."
+    exit 1
+fi
+
+exists=$(jq '.modules | to_entries | any(.value.kind == "meta")' "$DATA_DIR/consensus.json")
+if [ "$exists" == "true" ]; then
+    echo "A module of kind 'meta' already exists."
+#    exit 1
```
bradleystachurski, line 34: I think we should still exit 😄

scripts/ops/add_meta_module.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ -z "$1" ]; then
+    echo "Error: No directory specified."
+    echo "Usage: $0 <data_directory>"
+    exit 1
+fi
+DATA_DIR=$1
+
+echo "This script will add the meta module to federation config, which is technically"
+echo "a consensus-breaking change. Make sure that:"
+echo "  * All fedimintd instances are shut down"
```
dpc, line 14: Do they need to be gracefully shut down at the session closure?

scripts/ops/add_meta_module.sh
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+NEXT_MOD_ID="$(jq '[.modules | keys[] | tonumber] | max + 1' "$DATA_DIR/consensus.json")"
+
+mv "$DATA_DIR/consensus.json" "$DATA_DIR/consensus.json.bak"
+mv "$DATA_DIR/local.json" "$DATA_DIR/local.json.bak"
+mv "$DATA_DIR/private.encrypt" "$DATA_DIR/private.encrypt.bak"
+
+PASSWORD="$(cat "$DATA_DIR/password.private")"
+fedimint-cli dev config-decrypt --in-file "$DATA_DIR/private.encrypt.bak" --out-file "$DATA_DIR/private.json.old" "$PASSWORD" > /dev/null
```
dpc, line 44: Nit: we're storing plaintext config on disk, which has security implications, but not that big of deal, I guess. We possibly could just read the content to memory.

dpc (commented):
Looks simple enough to not require rewriting in Rust. Yet.

scripts/ops/add_meta_module.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ -z "$1" ]; then
```
dpc, line 5: Though: would verify if `jq` and `fedimint-cli` are in the path, before starting to do anything.

scripts/ops/add_meta_module.sh
elsirion: Technically yes, but afaik the module doesn't send CIs on its own, right? So as long as you don't start voting on stuff right away and let a session pass first we should be good. @joschisan what do you think (except for "NOOOOOOOO!" :laughing: )?

scripts/ops/add_meta_module.sh
elsirion: I will worry about this once we stop storing the password right next to the encrypted config :P

scripts/ops/add_meta_module.sh
elsirion: done

PR #5351, opened 2024-05-23
chore: cut release candidate v0.3.2-rc.0

elsirion:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ resolver = "2"
 
 [workspace.package]
 name = "fedimint"
-version = "0.3.1-rc.1"
+version = "0.3.2-rc.0"
 
 [workspace.metadata]
 name = "fedimint"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,7 +1038,7 @@ dependencies = [
 
 [[package]]
 name = "devimint"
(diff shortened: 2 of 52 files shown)


Cargo.toml
```diff
@@ -55,7 +55,7 @@ resolver = "2"
 
 [workspace.package]
 name = "fedimint"
-version = "0.3.1-rc.1"
+version = "0.3.2-rc.0"
```
dpc, line 58: We really should `= { workspace = true}` everything here. Then this PR would be a one liner.

dpc (approved):

Cargo.toml
dpc: BTW. Why are we updating from `0.3.1-rc.1` instead of `0.3.1`?

elsirion: I branched off from the `v0.3.1-rc.1` tag to the `v0.3.1` branch since there were already more changes on top and we didn't want to go through another round of testing.

Untitled Diagram drawio(5)

elsirion: > We really should `= { workspace = true}` everything here. Then this PR would be a one-liner.

Let's do that on master (at least some crates were already switched to that scheme) so 0.4 will have it. These changes are 100% automated, so not a huge deal imo.

PR #5352, opened 2024-05-23
chore: unify & clean Cargo.toml versions and such

dpc:
Not everything, since I'm afraid to do it automatically, and I ran out patience.

### DIFF
--- a/crypto/aead/Cargo.toml
+++ b/crypto/aead/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fedimint-aead"
-version = "0.4.0-alpha"
+version = { workspace = true }
 authors = ["The Fedimint Developers"]
 edition = "2021"
 description = "aead utilities on top of ring"
--- a/devimint/Cargo.toml
+++ b/devimint/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "devimint"
-version = "0.4.0-alpha"
+version = { workspace = true }
 edition = "2021"
 license = "MIT"
 publish = false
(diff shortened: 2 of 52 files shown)


crypto/aead/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fedimint-aead"
-version = "0.4.0-alpha"
+version = { workspace = true }
```
dpc, line 3: All of these are now `{ workspace = true }`, at least.

devimint/Cargo.toml
```diff
@@ -52,4 +52,4 @@ fedimint-portalloc = { path = "../utils/portalloc" }
 fs-lock = "0.1.3"
 
 [build-dependencies]
-fedimint-build = { version = "=0.4.0-alpha", path = "../fedimint-build" }
+fedimint-build = { path = "../fedimint-build" }
```
elsirion, line 55: Won't that break publishing? I had to painstakingly add all these version numbers in addition to paths when first publishing to crates.io :sob:

elsirion (commented):
Can we dry-run publishing? Alternatively: how about just publishing this rather random alpha version :sweat_smile: it won't be pulled in automatically by tooling imo since it's a pre-release, so not too bad imo.

devimint/Cargo.toml
dpc: Oh. I thought I've seen plenty of ones that didn't have it ...

dpc:
My bad. Will redo.

dpc:
https://github.com/fedimint/fedimint/pull/5354

devimint/Cargo.toml
elsirion: Maybe some tooling will add it automagically or I was just holding it wrong?

PR #5358, opened 2024-05-23
chore: Cargo.toml cleanup

dpc:
Some trivia: in Helix, you can select bunch of deps, press `|`, `sort`, `Enter` and get all deps sorted.


elsirion (commented):
If you are doing any reformatting please enforce it in CI. I think it's a great idea, but having a big diff every few months after things get messy again isn't the way.

dpc:
> If you are doing any reformatting please enforce it in CI. I think it's a great idea, but having a big diff every few months after things get messy again isn't the way.

How did I not thought of looking for a tool to do it first!

https://github.com/fedimint/fedimint/pull/5364

bradleystachurski (approved):

PR #5363, opened 2024-05-24
chore(ci): run a round of fuzzing for every fuzz target

dpc:
Now that we self-hosted runners, that we don't pay for cycles used, it makes sense to run some fuzzing all the time, just to avoid accidental breakages and maybe even randomly discover something over time.

It will be running with the same `runLowPrio` we use for other CPU-heavy things, so it should not cause tests running in parallel to be affected too much.

Re #287

### DIFF
--- a/.github/workflows/ci-fuzz.yml
+++ b/.github/workflows/ci-fuzz.yml
@@ -1,0 +1,85 @@
+name: "Fuzzing"
+
+env:
+  PERFIT_SERVER: https://perfit.dev.fedimint.org
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the "main" branch and release tags
+  push:
+    branches: [ "main", "master", "devel", "releases/v*" ]
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+  pull_request:
+    branches: [ "main", "master", "devel", "releases/v*" ]
+  merge_group:
+    branches: [ "main", "master", "devel", "releases/v*" ]
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-fuzz.yml
```diff
@@ -0,0 +1,85 @@
...
+          # we need to use `nix develop -c` to be able to use `nix build` inside of backwards-compatibility-test
+          # Disable `sccache`, it seems incompatible with self-hosted runner sandbox for some reason, and
+          # it doesn't benefit us much here anyway.
+          env \
+            TMPDIR=/tmp \
+            CARGO_PROFILE=ci \
+            PERFIT_ACCESS_TOKEN="${{ secrets.PERFIT_ACCESS_TOKEN }}" \
+            nix develop .#fuzz -c \
+            nix shell github:rustshop/fs-dir-cache -c \
+            nix run 'github:rustshop/perfit?rev=a2ea3bae86b0e70d2ebdbca1fd16a843b7f0a3bd#perfit' -- \
```
elsirion, line 62: Why record runtime of a job that's always supposed to take 30s? :laughing:

elsirion (commented):
conceptACK

.github/workflows/ci-fuzz.yml
dpc: Maybe it lies. :D

Or maybe we'll change it at some point to do e.g. X iterations instead.

Or maybe I just want an extra pretty chart in the dashboard. :D

I don't think it hurts. :)

maan2003 (approved):

PR #5364, opened 2024-05-24
chore: introduce `cargo-sort` into our lives

dpc:
Currently not enforcing or modifing anything, as would be best to find a convenient time to minimize amount of merge conflicts.

After we get it in, we can run it, land the sorting PR, and then I'll add a check in the lint step or maybe right in the git check hook.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -223,6 +223,7 @@
                   pkgs.cargo-udeps
                   pkgs.cargo-audit
                   pkgs.cargo-deny
+                  pkgs.cargo-sort
                   pkgs.parallel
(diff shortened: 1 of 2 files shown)


dpc:
@bradleystachurski LOL. I don't know why, but if you run it 4 times in a row, then it works. :D

**Edit**:

OK. It seems that every time it does only one step of reformatting the file. Eg. it will just sort deps. If that was OK, it will just move the sections, etc.

I don't think this will be a problem after the initial run.

bradleystachurski (approved):

Issue #5366, opened 2024-05-24
Follow up with private federations changes


dpc:
https://github.com/fedimint/fedimint/pull/5322#issuecomment-2128376292

So I won't forget.

dpc:
Kind of done. I left the secrets in OOB notes. They can be removed (possibly optionally at runtime) later.

PR #5367, opened 2024-05-24
chore: future-proof api secrets in private federations

dpc:
Re #5366

In the future we might want to let guardians manage Api secrets of a Federation via UI, storing the Api Secrets configuration in the database.

This will make env var naming confusing, just like it was in the case of a `FM_BITCOIN_RPC_URL`, etc. which we've split into `FM_FORCE_` and `FM_DEFAULT_` recently.

Do the same split for `FM_FORCE_API_SECRETS`, so the meaning and behavior of it does not need to change in the future.

While at it merge existing two arguments into one, to avoid 2 * 2 env vars/flags, and unify everything in a newtype.

##### Testing

```
env RUST_LOG=fm=debug j devimint-env
```

Verify in the logs of fedimintd that it says near the top:

```
2024-05-24T21:50:36.733769Z  INFO spawn{task="main"}: fm::net::auth: Api available for public access
```

Then run:

```
env FM_FORCE_API_SECRETS=asdfadsf,asdfad,asdfadsfasdf RUST_LOG=fm=debug j devimint-env
```

Verify in the logs, that it now says:

```
2024-05-24T21:56:16.403660Z  INFO spawn{task="main"}: fm::net::auth: Api available for private access num_secrets=3
```

### DIFF
--- a/fedimint-server/src/net/api.rs
+++ b/fedimint-server/src/net/api.rs
@@ -1,14 +1,16 @@
 mod http_auth;
 
-use std::fmt::{Debug, Formatter};
+use std::fmt::{self, Debug, Formatter};
 use std::net::SocketAddr;
 use std::panic::AssertUnwindSafe;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::Context;
+use anyhow::{bail, Context};
 use async_trait::async_trait;
 use fedimint_core::core::ModuleInstanceId;
+use fedimint_core::encoding::{Decodable, Encodable};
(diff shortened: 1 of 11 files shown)


fedimint-server/src/net/api.rs
```diff
@@ -20,6 +22,60 @@ use tracing::{error, info};
...
+
+    /// Get "active" secret - one that should be used to call other peers
+    pub fn get_active(&self) -> Option<String> {
+        self.0.first().cloned()
+    }
+
+    /// Get all secrets
+    pub fn get_all(&self) -> Vec<String> {
+        self.0.clone()
+    }
```
elsirion, line 71: I was heard :D

elsirion (approved):

PR #5374, opened 2024-05-25
refactor(client): api version discovery no longer a query strategy

dpc:
Turn a clumsy query strategy into (hopefully) a bit less (but still quite a bit) clumsy normal code.

The improvement here now is that the discovery logic can store every peer's individual response, and in an absence of timely (or any) response in the future, use it again. This should make the whole thing much less fragile w.r.t. peer availability.

Other than that it should allow cleaning up "normal" query strategies APIs.


fedimint-api-client/src/api.rs
```diff
@@ -323,7 +319,26 @@ pub trait FederationApiExt: IRawFederationApi {
...
+    {
+        self.request_single_peer(timeout, method, params, peer_id)
+            .await
+            .map_err(PeerError::Rpc)
+            .and_then(|v| {
+                serde_json::from_value(v).map_err(|e| PeerError::ResponseDeserialization(e.into()))
+            })
+    }
 
+    // serde_json::from_value::<PeerRet>(o.0)
```
dpc, line 340: TODO: 

image

fedimint-api-client/src/query.rs
```diff
@@ -388,302 +385,3 @@ impl<R> QueryStrategy<R, BTreeMap<PeerId, R>> for ThresholdOrDeadline<R> {
         }
     }
 }
-
-/// Query for supported api versions from all the guardians (with a deadline)
-/// and calculate the best versions to use for each component (core + modules).
-pub struct DiscoverApiVersionSet {
```
dpc, line 394: The query strategy is gone.

fedimint-api-client/src/query.rs
```diff
@@ -388,302 +385,3 @@ impl<R> QueryStrategy<R, BTreeMap<PeerId, R>> for ThresholdOrDeadline<R> {
...
-                }
-            }
-            QueryStep::Retry(v) => QueryStep::Retry(v),
-            QueryStep::Continue => QueryStep::Continue,
-            QueryStep::Failure { general, peers } => QueryStep::Failure { general, peers },
-        }
-    }
-}
-
-fn discover_common_core_api_version(
```
dpc, line 444: This logic + tests for it were moved to a new file unmodified. They possibly suck and could use some thinking, but it has basic tests, so it shouldn't explode anytime soon.

fedimint-client/src/lib.rs
```diff
@@ -1491,50 +1518,93 @@ impl Client {
...
+        // task. This is all just to bound the time user can be waiting
+        // for the join operation to finish, at the risk of picking wrong version in
+        // very rare circumstances.
+        let _: Result<_, Elapsed> = runtime::timeout(
+            Duration::from_secs(15),
+            num_responses_receiver.wait_for(|num| num_peers.threshold() <= *num),
+        )
+        .await;
+
+        let mut peer_api_version_sets = BTreeMap::new();
```
dpc, line 1591: Make it a function.

fedimint-client/src/lib.rs
```diff
@@ -1491,50 +1518,93 @@ impl Client {
...
+        let mut dbtx = db.begin_transaction_nc().await;
+        for peer_id in num_peers.peer_ids() {
+            if let Some(v) = dbtx
+                .get_value(&PeerLastApiVersionsSummaryKey(peer_id))
+                .await
+            {
+                peer_api_version_sets.insert(peer_id, v.0);
+            }
+        }
+        drop(dbtx);
```
dpc, line 1602: Make it a function.

fedimint-core/src/module/version.rs
```diff
@@ -192,7 +192,7 @@ impl cmp::Ord for ApiVersion {
 /// Each element must have a distinct major api number, and means
 /// either minimum required API version of this major number (for the client),
 /// or maximum supported version of this major number (for the server).
-#[derive(Debug, Clone, Serialize, Default)]
+#[derive(Debug, Clone, Serialize, Default, Encodable, Decodable)]
```
dpc, line 195: This stuff here gets stored in a db now, so needs these.

fedimint-client/src/lib.rs
```diff
@@ -1471,14 +1487,25 @@ impl Client {
...
+
     /// Load the common api versions to use from cache and start a background
     /// process to refresh them.
     ///
     /// This is a compromise, so we not have to wait for version discovery to
     /// complete every time a [`Client`] is being built.
     async fn load_and_refresh_common_api_version_static(
         config: &ClientConfig,
-        module_inits: &ModuleInitRegistry<DynClientModuleInit>,
+        client_module_init: &ClientModuleInitRegistry,
```
dpc, line 1508: TODO: no need to rename, just an accident.

fedimint-client/src/lib.rs
```diff
@@ -1375,67 +1380,78 @@ impl Client {
...
-                },
+    pub async fn refresh_peers_api_versions(
+        num_peers: NumPeers,
+        api: DynGlobalApi,
+        db: Database,
+        num_responses_sender: watch::Sender<usize>,
+    ) {
+        let mut requests = FuturesUnordered::new();
+
+        // Make a single request to a peer, possibly for a delay
```
dpc, line 1393: s/for/after/

fedimint-api-client/src/api.rs
```diff
@@ -323,6 +319,23 @@ pub trait FederationApiExt: IRawFederationApi {
             request.await
         }
     }
+    async fn request_single_peer_typed<Ret>(
+        &self,
+        timeout: Option<Duration>,
```
elsirion, line 324: nit: Do you think that timeout option is useful here? It seems to me that generally handling timeouts at the highest level possible is the way (e.g. if there are two requests don't enforce 1s timeout for each but a total of 2s). Async code can be cancelled, so there is little need for passing timeouts down the call stack.

For consistency I'd try to remove the timeout option from `request_single_peer` later too.

fedimint-client/src/lib.rs
```diff
@@ -1375,67 +1380,78 @@ impl Client {
...
-        let deadline = now() + Duration::from_secs(30);
+        }
 
-        loop {
-            let res =
-                Self::discover_common_api_version_static_try(config, client_module_init, api, mode)
-                    .await;
+        for peer_id in num_peers.peer_ids() {
+            requests.push(make_request(Duration::ZERO, peer_id, &api))
+        }
```
elsirion, line 1420: nit: could be an iterator chain+collect here instead of mut access

fedimint-client/src/lib.rs
```diff
@@ -1375,67 +1380,78 @@ impl Client {
...
+                    let mut dbtx = db.begin_transaction().await;
+                    dbtx.insert_entry(
+                        &PeerLastApiVersionsSummaryKey(peer_id),
+                        &PeerLastApiVersionsSummary(o),
+                    )
+                    .await;
+                    dbtx.commit_tx().await;
+                    num_responses += 1;
+                    // ignore errors: we don't care if anyone is still listening
+                    let _ = num_responses_sender.send(num_responses);
```
elsirion, line 1452: nit: Instead of keeping `num_responses` around separately you could use [`send_modify`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_modify) directly on `num_responses_sender`

elsirion (commented):
Now that we save every peer's response do we actually still need `CachedApiVersionSetKey` in the DB? Calculating the supported version from peer responses sounds cheap enough and we could avoid some redundancy.

joschisan (approved):
Pr looks good. I wrote an alternative way to remove the ApiVersionDiscovery query strategy using FilterMapThreshold instead so I am linking it here for context: https://github.com/fedimint/fedimint/pull/5375 Everything stays the same but we always obtain 2f + 1 version summaries - I am not sure if if that is sufficient as right now we sometimes try to obtain n and sometimes n / 2 but do not seem to enforce any bound even below that? Feel free to merge whatever you prefer.

fedimint-api-client/src/api.rs
dpc: Copied it from the existing code, to not expand the change. I support removing it like in @joschisan PR.

dpc:
I will follow up with fixing the nits, pinky promise. I'm also open to discussing (or arguing, which makes more sense :D) the details of logic here and eventually improving it.

Right now the goal is to get this code out of the way to unblock things like removing the `timeout`.

dpc:
> Now that we save every peer's response do we actually still need CachedApiVersionSetKey in the DB? 

Yeah, crossed my mind and is probably true.

>  I am not sure if if that is sufficient as right now we sometimes try to obtain n and sometimes n / 2 but do not seem to enforce any threshold so might even calculate the api version from one response?

Since now we always have previous responses, even more than before it only matters during the first join.

So during the initial join the app is waiting and possibly can't display anything in the UI, until the `build_client` doesn't wrap it up, and to wrap it up it needs *some* version to use. So it is a shame to (probably mostly theoretically) ever get stuck here, on something that doesn't really have security properties.

PR #5380, opened 2024-05-27
Cleanup Server III

joschisan:


fedimint-server/src/atomic_broadcast/mod.rs
```diff
@@ -1,122 +0,0 @@
...
-//! Though the broadcast depends on [fedimint_core] for [fedimint_core::PeerId],
-//! [fedimint_core::encoding::Encodable] and [fedimint_core::db::Database]
-//! it implements no consensus logic specific to Fedimint, to which we will
-//! refer as Fedimint Consensus going forward. To the broadcast a consensus item
-//! is merely a vector of bytes without any further structure.
-//!
-//! # The journey of a ConsensusItem
-//!
-//! Let us sketch the journey of an [fedimint_core::epoch::ConsensusItem] into a
-//! signed session outcome.
```
elsirion, line 18: That sounds like an interesting writeup, why remove it? Is it too out of date?

joschisan: It is mostly out of date now.

elsirion (approved):

Issue #4903, opened 2024-04-10
Consensus state commitments


elsirion:
Currently, the consensus only forms agreement to state changes of the system. This leaves room for bugs to cause the consensus *state* that each guardian server holds in its DB to diverge. So far we haven't *seen* this with our main modules afaik, only experimental third party ones, but we cannot be sure.

What I'd like is:
* A function on the server module trait that can serialize the entire consensus state of that module
* A way to efficiently hash the entire consensus log (likely a hash chain? we had this with HBBFT iirc, afaik not anymore with AlephBFT)
* A way to hash any other global consensus state

At the end of each session the module state hashes and the consensus log hash (and any other global consensus state I might be missing) is sent as the first consensus item in the new session. At first we might only warn if they differ, so we can investigate, but later an automated shutdown should be triggered.

PR #5384, opened 2024-05-28
feat(server): diagnostic consensus checksum

dpc:
Inspired by the discussion during last Deep Dive, about verifing determinism of consensus between peers and shuting down when incosistency was detected, it occured to me that we could get 90% there by simply tracking a rolling checksum of what happened and displaying it as diagnostics that guardians could sometimes simply compare.

So here is an initial implementation.

The checksum is maintained separately over txes and each of the modules, so in case total checksum does not match it's possible to compare each module specific one and hopefully figure out which module is to blame.

The whole thing is purely diagnostic, though in the future we could expose it in the Admin UI somehow.

No attempt is made to synchronize/correct the checksum between peers. It seems perfectly fine that Federations that were created before this was introduced will simply have a potential mismatched.

### DIFF
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -1,4 +1,5 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
+use std::default::Default;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -7,7 +8,7 @@ use anyhow::{anyhow, bail};
 use async_channel::Receiver;
 use fedimint_api_client::api::{DynGlobalApi, FederationApiExt, PeerConnectionStatus};
 use fedimint_api_client::query::FilterMap;
-use fedimint_core::core::{DynOutput, MODULE_INSTANCE_ID_GLOBAL};
+use fedimint_core::core::{DynInput, DynOutput, MODULE_INSTANCE_ID_GLOBAL};
 use fedimint_core::db::{Database, DatabaseTransaction, IDatabaseTransactionOpsCoreTyped};
 use fedimint_core::encoding::Decodable;
 use fedimint_core::endpoint_constants::AWAIT_SIGNED_SESSION_OUTCOME_ENDPOINT;
@@ -28,6 +29,7 @@ use rand::Rng;
 use tokio::sync::{watch, RwLock};
 use tracing::{debug, info, instrument, warn, Level};
 
+use super::consensus_checksum::ConsensusChecksumTracker;
 use crate::config::ServerConfig;
 use crate::consensus::aleph_bft::backup::{BackupReader, BackupWriter};
 use crate::consensus::aleph_bft::data_provider::{DataProvider, UnitData};
@@ -68,11 +70,13 @@ pub struct ConsensusEngine {
     pub peer_id_str: Vec<String>,
     pub connection_status_channels: Arc<RwLock<BTreeMap<PeerId, PeerConnectionStatus>>>,
     pub task_group: TaskGroup,
+
+    pub consensus_checksum: ConsensusChecksumTracker,
 }
 
 impl ConsensusEngine {
     #[instrument(name = "run", skip_all, fields(id=%self.cfg.local.identity))]
-    pub async fn run(self) -> anyhow::Result<()> {
+    pub async fn run(mut self) -> anyhow::Result<()> {
         if self.cfg.consensus.broadcast_public_keys.len() == 1 {
             self.run_single_guardian(self.task_group.make_handle())
                 .await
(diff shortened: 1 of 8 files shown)


dpc:
@joschisan Does this pass a smell test?

Is there any other data that we could throw into the checksumming to make it more robust? E.g. `item_number` or `session_idx`?

fedimint-server/src/consensus/engine.rs
```diff
@@ -457,7 +462,7 @@ impl ConsensusEngine {
 
     #[instrument(target = "fm::consensus", skip(self, item), level = "info")]
     pub async fn process_consensus_item(
-        &self,
+        &mut self,
```
dpc, line 465: `&mut self` is needed on top level functions now to avoid locking/inner-mutability.

dpc:
Thought: currently the whole txid of an accepted tx is thrown it, but that does not keep any per-module accounting in case a tx id rejected.

We could keep track of tx in some more granular fashion (per-input/output) but the question is what exactly is the best way to do it? Another field that for tx that returned an error hashes-in  module_id of a corresponding input or output? This way it would relatively rarely need to do anything (low overhead). But then ... we don't commit anything on error, so hash changes wouldn't get committed either, and changing that feels like messing with the core logic way too much.

joschisan:
Currently the system already implicitly creates a checksum for every session in form of the header of a session outcome which is the Merkle root of all consensus items accepted within a session. At the end of a session this is checked by the peers as they start to collect each others signatures for the header, which implicitly verifies the signed messages are identical. Now if a consensus item is falsely accepted or rejected by a peer the Merkle root at the end of the session will differ. If there is still a sufficient number of correct peers they will finish the session and move on while the incorrect peer(s) would recognise they disagree with consensus and panic. If however, no threshold of peers agree on the header the system just hangs, continuously logging the warning "Received invalid signature from peer" which could me more informative here and should probably be an error.

So if I understand this pr correctly, if your final xor checksum behaves similar to the header and if it were to diverge peers would panic or the federations just hangs at the end of the session, however, we could determine which modules module consensus item caused the failure or if peers disagreed on a transaction.

Right now, if the header diverges, you can read all accepted consensus items for the pending session out of the db, manually or via an endpoint, which would allow you to actually obtain the item(s) of disagreement.

dpc:
> So if I understand this pr correctly, if your final xor checksum behaves similar to the header and if it were to diverge peers would panic or the federations just hangs at the end of the session, however, we could determine which modules module consensus item caused the failure or if peers disagreed on a transaction.

Oh, I wasn't aware of the checksum at the end of session. That's good to know.

In that case the biggest (only?) benefit of this PR, is that it could indeed possibly narrow it all down to a specific module by guardians just comparing checksums.

dpc:
Rebased, I've split the tx into separate hashes for inputs and outputs for each module_id.

@joschisan @elsirion 

Now the question is, is it worth the trouble on top of the existing check. I weakly think it doesn't cost much and could be helpful to debug any determinism bugs that are cough in the wild, especially in 3rd party modules. But I might be biased because I invested the time to write this code already. :D

elsirion:
When I was originally thinking about this problem in #4903 the initial idea was to hash the DB state of modules since that could easily diverge without leading to a consensus failure for a long time. I also mentioned the tx history, not remembering the signing at the end of sessions, but given that after each session we check consistency that's less of a concern. 

For debuggability https://github.com/fedimint/fedimint/pull/5386 (maybe with a few improvements) seems more pragmatic.

One thing your PR made me realize though is that we could have an API that maps the last 1000 or so item indices to their respective consensus hashes, so we could notice divergence even before the end of a session (accepting false negatives when a peer is too far behind and doesn't have the hashes cached anymore, we can't keep them forever imo).

dpc:
> DB state of modules since that could easily diverge without leading to a consensus failure for a long time. 

I was thinking about it, but it would rather heavy thing to track, and is a the wrong level of abstraction: Ultimately there is no obligation by any module to have exactly same state - e.g. meta module keeps different details about desired vote by the guardian. And different versions etc. could indefinitely use a slightly different state and yet still give same consensus results. S the acceptance or not of CIs and Txes sems like the correct place to verify determinism.

The fact that they can diverge without consensus failure isn't a big deal IMO. The consensus failure is when the peers gets a different result, and that's kind of what matters.

> For debuggability https://github.com/fedimint/fedimint/pull/5386 (maybe with a few improvements) seems more pragmatic.

That might be true.

dpc:
> > For debuggability #5386 (maybe with a few improvements) seems more pragmatic.
> 
> That might be true.

Hmmm... After looking at it for another moment...

I don't see a way to make it able to point easily to affected citem and/or modules that participated in it. You need to do some post-processing / use external tools to figure out which citem was the extra/missing one. The nice property of the current PR is that you can glance at the checksums and immediately tell. And being able to tell which module is possibly buggy is the most important thing, as the citem itself is highly likely innocent.

So I don't see these two PRs as alternatives, but rather complementary. Session verification might stop the peer, then guardians can verify logs to narrow down the suspects. Only then, if the problem is not already known/fixed, affects modules that someone cares about debugging etc. someone can fire up some tooling and/or collect database dumps to invest time in debugging it.

elsirion:
Regarding your concern about not being able to tell the module that rejected a transaction which another guardian accepted: that should be super rare during normal operation, we could log validation errors when they happen for transactions that came from consensus.

dpc:
Abandoning.

PR #5386, opened 2024-05-29
chore: improve logging on consensus failure

joschisan:
If such an error or panic occurs we need to read the pending accepted items of the current session from disk to obtain the item at which the consensus diverged.

### DIFF
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -26,7 +26,7 @@ use fedimint_core::{timing, NumPeers, PeerId};
 use futures::StreamExt;
 use rand::Rng;
 use tokio::sync::{watch, RwLock};
-use tracing::{debug, info, instrument, warn, Level};
+use tracing::{debug, error, info, instrument, warn, Level};
 
 use crate::config::ServerConfig;
 use crate::consensus::aleph_bft::backup::{BackupReader, BackupWriter};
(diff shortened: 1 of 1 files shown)


fedimint-server/src/consensus/engine.rs
```diff
@@ -393,13 +396,16 @@ impl ConsensusEngine {
                         if self.keychain.verify(&header, &signature, to_node_index(peer)){
                             signatures.insert(peer, signature);
                         } else {
-                            warn!(target: LOG_CONSENSUS, "Received invalid signature from peer {peer}");
+                            error!(target: LOG_CONSENSUS, "Consensus Failure: invalid header signature from {peer}");
```
elsirion, line 399: Could we also print all our items here? I suppose all guardians would get that error and then we could compare?

fedimint-server/src/consensus/engine.rs
```diff
@@ -350,7 +350,10 @@ impl ConsensusEngine {
                         .items
                         .split_at(pending_accepted_items.len());
 
-                    assert!(processed.iter().eq(pending_accepted_items.iter()));
+                    assert!(
+                        processed.iter().eq(pending_accepted_items.iter()),
+                        "Consensus Failure: pending accepted items disagree with federation consensus"
+                    );
```
elsirion, line 356: If this was an `assert_eq` (over a `Vec`, idk about iterators) we would see the difference.

fedimint-server/src/consensus/engine.rs
dpc: Hmm... Actually, right...

If all guardians print all items whenever they receive a wrong signature, all prefixed if their ID, then it should be relatively easy to binary search the outstanding item.

We should probably print some very compact representation of all items. The exact item is not all that useful, but the module_ids on it are.

dpc:
Follow-up: https://github.com/fedimint/fedimint/pull/5417

PR #5390, opened 2024-05-29
refactor(consensus): some pass-by-ref tweaks

dpc:
When working #5384 , which probably will get abandoned I needed/noticed some ways to avoid a clone and passed by reference stuff that didn't need to be moved.

The `&mut self` is not neccessary, but technically it does denote that the consensus code has an exclusive reference to `ConsensusEngine` which makese sense.

### DIFF
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -539,7 +539,7 @@ impl ConsensusEngine {
             .expect("Committing consensus epoch failed");
 
         CONSENSUS_ITEMS_PROCESSED_TOTAL
-            .with_label_values(&[peer_id_str])
+            .with_label_values(&[&self.peer_id_str[peer.to_usize()]])
             .inc();
         timing_prom.observe_duration();
 
@@ -559,11 +559,12 @@ impl ConsensusEngine {
         match consensus_item {
             ConsensusItem::Module(module_item) => {
                 let instance_id = module_item.module_instance_id();
+
                 let module_dbtx = &mut dbtx.to_ref_with_prefix_module_id(instance_id);
 
                 self.modules
                     .get_expect(instance_id)
-                    .process_consensus_item(module_dbtx, module_item, peer_id)
+                    .process_consensus_item(module_dbtx, &module_item, peer_id)
                     .await
             }
             ConsensusItem::Transaction(transaction) => {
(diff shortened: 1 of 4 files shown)


fedimint-server/src/consensus/engine.rs
```diff
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::default::Default;
```
bradleystachurski, line 2: [nit] What needed default?

bradleystachurski (dismissed):

fedimint-server/src/consensus/engine.rs
dpc: Yyyyy... I don't know, and I don't understand . :D

dpc:
Pinging @joschisan if he's fine with these.

fedimint-server/src/consensus/engine.rs
dpc: It could be some glitch w.r.t merging and merge base. I don't see anything that would need it, but then ... why is clippy to failing... ?

elsirion: Did you try removing it?

elsirion: `cargo check` was still running after removing it, so I pushed an updated version without this addition.

elsirion (dismissed):
Looks good to me, but let's wait for @joschisan.

joschisan:
lgtm!

dpc:
> @dpc does the mutable reference force us to clone the db?

I took the original PR that we decided to abandon, and removed all the stuff that we didn't need. I'll double-check.

> I used immutable references here to express that all attributes of ConsensusEngine are immutable since they somewhat global state.


So here is where Rust is confusing.

Before Rust 1.0 there was a big argument all over the community called "mutocalypse". The reason was that `&mut` is a misnomer, and large chunk of community wanted to rename it to `&uniq`, because `&mut` actually means "unique reference". The mutability part come from the fact that having an unique reference makes mutation safe.

So `&mut self` does kind of signify that "no other part of the code can be referencing `self` in any way, which is kind of a strong property to have.

> Besides that pr looks great! Do you have a strong opinion on this? 

I think I will just revert it. It doesn't matter.

elsirion:
Please squash (sorry if I messed up some workflow by force-pushing that lead to a second commit).

elsirion (approved):

PR #5391, opened 2024-05-29
feat: Support split partial invoice payments

### LINKED ISSUE #5103: Support split lightning payments
With the addition of `pay_private` it should be possible to make only a partial payment of an invoice and the rest of the payment can come from somewhere else (another mint/gateway, self custodial, etc). This should be pretty doable to add now, will basically need to add an amount [`LightningPayCommon`](https://github.com/fedimint/fedimint/blob/7998f2e95f588c63445cb30c8efb06e3c4476a30/modules/fedimint-ln-client/src/pay.rs#L62) that overrides the amount in the invoice and then bubble that down to where `pay_private` is called.

Last time I broke stuff with db migrations so I'll let someone else take a stab at it.

benthecarman:
Closes #5103

This should give functional support for doing split/partial payments of invoices for the v1 lightning module.

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -124,7 +124,7 @@ pub enum ClientCmd {
     LnPay {
         /// Lightning invoice or lnurl
         payment_info: String,
-        /// Amount to pay, used for lnurl
+        /// Amount to pay, used for lnurl or for a partial payment of an invoice
         #[clap(long)]
         amount: Option<Amount>,
         /// Invoice comment/description, used on lnurl
@@ -407,7 +407,7 @@ pub async fn handle_command(
                 contract_id,
                 fee,
             } = lightning_module
-                .pay_bolt11_invoice(ln_gateway, bolt11, ())
+                .pay_bolt11_invoice(ln_gateway, bolt11, amount, ())
                 .await?;
             let operation_id = payment_type.operation_id();
             info!(
--- a/gateway/ln-gateway/src/lightning/lnd.rs
+++ b/gateway/ln-gateway/src/lightning/lnd.rs
@@ -7,8 +7,10 @@ use anyhow::ensure;
 use async_trait::async_trait;
 use bitcoin_hashes::Hash;
 use fedimint_core::task::{sleep, TaskGroup};
+use fedimint_core::time::duration_since_epoch;
 use fedimint_core::{secp256k1, Amount};
 use fedimint_ln_common::PrunedInvoice;
+use futures::StreamExt;
 use hex::ToHex;
 use secp256k1::PublicKey;
 use tokio::sync::mpsc;
@@ -16,18 +18,21 @@ use tokio_stream::wrappers::ReceiverStream;
 use tonic::Status;
 use tonic_lnd::invoicesrpc::AddHoldInvoiceRequest;
 use tonic_lnd::lnrpc::failure::FailureCode;
+use tonic_lnd::lnrpc::fee_limit::Limit;
+use tonic_lnd::lnrpc::htlc_attempt::HtlcStatus;
 use tonic_lnd::lnrpc::payment::PaymentStatus;
 use tonic_lnd::lnrpc::{
-    ChanInfoRequest, ChannelPoint, CloseChannelRequest, ConnectPeerRequest, GetInfoRequest,
-    LightningAddress, ListChannelsRequest, OpenChannelRequest,
+    ChanInfoRequest, ChannelPoint, CloseChannelRequest, ConnectPeerRequest, FeeLimit,
+    GetInfoRequest, LightningAddress, ListChannelsRequest, MppRecord, OpenChannelRequest,
+    QueryRoutesRequest,
 };
 use tonic_lnd::routerrpc::{
     CircuitKey, ForwardHtlcInterceptResponse, ResolveHoldForwardAction, SendPaymentRequest,
-    TrackPaymentRequest,
+    SendToRouteRequest, TrackPaymentRequest,
 };
-use tonic_lnd::tonic::Code;
+use tonic_lnd::tonic::{Code, Response};
 use tonic_lnd::walletrpc::AddrRequest;
-use tonic_lnd::{connect, Client as LndClient};
+use tonic_lnd::{connect, Client as LndClient, Client};
 use tracing::{debug, error, info, trace, warn};
 
 use super::{ChannelInfo, ILnRpcClient, LightningRpcError, RouteHtlcStream, MAX_LIGHTNING_RETRIES};
@@ -42,7 +47,7 @@ use crate::gateway_lnrpc::{
 
 type HtlcSubscriptionSender = mpsc::Sender<Result<InterceptHtlcRequest, Status>>;
 
-const LND_PAYMENT_TIMEOUT_SECONDS: i32 = 180;
+const LND_PAYMENT_TIMEOUT_SECONDS: u64 = 180;
 
 pub struct GatewayLndClient {
     /// LND client
@@ -416,40 +421,74 @@ impl ILnRpcClient for GatewayLndClient {
         debug!("LND got client to pay invoice {invoice:?}, will check if payment already exists");
 
         // If the payment exists, that means we've already tried to pay the invoice
-        let preimage: Vec<u8> = if let Some(preimage) = self
+        if let Some(preimage) = self
             .lookup_payment(invoice.payment_hash.to_byte_array().to_vec(), &mut client)
             .await?
         {
             info!("LND payment already exists for invoice {invoice:?}");
-            hex::FromHex::from_hex(preimage.as_str()).map_err(|error| {
+            let preimage = hex::FromHex::from_hex(preimage.as_str()).map_err(|error| {
                 LightningRpcError::FailedPayment {
                     failure_reason: format!("Failed to convert preimage {error:?}"),
                 }
-            })?
-        } else {
-            // LND API allows fee limits in the `i64` range, but we use `u64` for
-            // max_fee_msat. This means we can only set an enforceable fee limit
-            // between 0 and i64::MAX
-            let fee_limit_msat: i64 =
-                max_fee
-                    .msats
-                    .try_into()
-                    .map_err(|error| LightningRpcError::FailedPayment {
-                        failure_reason: format!(
-                            "max_fee_msat exceeds valid LND fee limit ranges {error:?}"
-                        ),
-                    })?;
+            })?;
+            return Ok(PayInvoiceResponse { preimage });
+        }
 
-            let amt_msat = invoice.amount.msats.try_into().map_err(|error| {
+        // LND API allows fee limits in the `i64` range, but we use `u64` for
+        // max_fee_msat. This means we can only set an enforceable fee limit
+        // between 0 and i64::MAX
+        let fee_limit_msat: i64 =
+            max_fee
+                .msats
+                .try_into()
+                .map_err(|error| LightningRpcError::FailedPayment {
+                    failure_reason: format!(
+                        "max_fee_msat exceeds valid LND fee limit ranges {error:?}"
+                    ),
+                })?;
+
+        let send_amt_msat: i64 = invoice
+            .send_amount
+            .unwrap_or(invoice.amount)
+            .msats
+            .try_into()
+            .map_err(|error| LightningRpcError::FailedPayment {
+                failure_reason: format!("amount exceeds valid LND amount ranges {error:?}"),
+            })?;
+
+        let final_cltv_delta = invoice.min_final_cltv_delta.try_into().map_err(|error| {
+            LightningRpcError::FailedPayment {
+                failure_reason: format!("final cltv delta exceeds valid LND range {error:?}"),
+            }
+        })?;
+
+        let dest_features = wire_features_to_lnd_feature_vec(&invoice.destination_features)
+            .map_err(|e| LightningRpcError::FailedPayment {
+                failure_reason: e.to_string(),
+            })?;
+
+        debug!("LND payment does not exist for invoice {invoice:?}, will attempt to pay");
+
+        // if we're doing a partial payment, need to do custom routing, otherwise
+        // we can use the simpler send_payment_v2 rpc
+        if invoice.send_amount.is_some_and(|a| a != invoice.amount) {
+            pay_partial(
+                &mut client,
+                invoice,
+                max_delay,
+                send_amt_msat,
+                fee_limit_msat,
+                final_cltv_delta,
+                dest_features,
+            )
+            .await
+        } else {
+            let amt_msat: i64 = invoice.amount.msats.try_into().map_err(|error| {
                 LightningRpcError::FailedPayment {
                     failure_reason: format!("amount exceeds valid LND amount ranges {error:?}"),
                 }
             })?;
-            let final_cltv_delta = invoice.min_final_cltv_delta.try_into().map_err(|error| {
-                LightningRpcError::FailedPayment {
-                    failure_reason: format!("final cltv delta exceeds valid LND range {error:?}"),
-                }
-            })?;
+
             let cltv_limit =
                 max_delay
                     .try_into()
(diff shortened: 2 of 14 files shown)


fedimint-cli/src/client.rs
```diff
@@ -415,7 +415,7 @@ pub async fn handle_command(
                 contract_id,
                 fee,
             } = lightning_module
-                .pay_bolt11_invoice(ln_gateway, bolt11, ())
+                .pay_bolt11_invoice(ln_gateway, bolt11, amount, ())
```
elsirion, line 418: This behavior should be documented somewhere. Ideally the `clap` help and maybe also a comment here that for BOLT11 we allow split payments.

gateway/ln-gateway/tests/integration_tests.rs
```diff
@@ -150,7 +152,8 @@ fn sha256(data: &[u8]) -> sha256::Hash {
 fn get_payment_data(gateway: Option<LightningGateway>, invoice: Bolt11Invoice) -> PaymentData {
     match gateway {
         Some(g) if g.supports_private_payments => {
-            let pruned_invoice: PrunedInvoice = invoice.try_into().expect("Invoice has amount");
+            let pruned_invoice =
+                PrunedInvoice::try_from(&invoice, None).expect("Invoice has amount");
```
elsirion, line 156: I'd avoid calling the function the same as a widely-used trait. Maybe `(try_)from_invoice`?

modules/fedimint-ln-client/src/db.rs
```diff
@@ -282,6 +285,57 @@ pub(crate) fn get_v3_migrated_state(
...
+        pub gateway_fee: Amount,
+        pub preimage_auth: sha256::Hash,
+        pub invoice: Bolt11Invoice,
+    }
+
+    #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+    pub struct LightningPayStateMachineOld {
+        pub common: LightningPayCommonOld,
+        pub state: LightningPayStates,
+    }
```
elsirion, line 307: You didn't end up using this it appears? Can be removed.

modules/fedimint-ln-client/src/db.rs
```diff
@@ -282,6 +285,57 @@ pub(crate) fn get_v3_migrated_state(
     }
 }
 
+/// Adds the `amount` field to `LightningPayCommon`
+pub(crate) fn get_v4_migrated_state(
```
elsirion, line 289: Looks good to me, but @m1sterc001guy please take a look :)

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1006,15 +1032,15 @@ impl LightningClientModule {
         // it/bounds for it
         let fee = match &client_output.output {
             LightningOutputV0::Contract(contract) => {
+                let amt = amount
+                    .map(|a| a.msats)
+                    .or(invoice.amount_milli_satoshis())
+                    .ok_or(anyhow!("MissingInvoiceAmount"))?;
```
elsirion, line 1038: I'd like to assert that `invoice.amount_milli_satoshis()` is `Some` no matter what.

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
```diff
@@ -170,7 +170,8 @@ impl SendStateMachine {
             lightning_context
                 .lnrpc
                 .pay_private(
-                    PrunedInvoice::try_from(invoice).expect("Invoice has amount"),
+                    // todo support override amount
+                    PrunedInvoice::try_from(&invoice, None).expect("Invoice has amount"),
```
elsirion, line 174: I'm wondering why v2 does the conversion on the GW side, leaking the description to the gateway operator. @joschisan

modules/fedimint-ln-client/src/lib.rs
benthecarman: it was enforced implicity by either calling `create_incoming_output` or `create_outgoing_output` but will do so just to be sure

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -974,16 +990,26 @@ impl LightningClientModule {
         }
 
         let (pay_type, client_output, contract_id) = if is_internal_payment {
+            if amount.is_some() {
+                bail!("Cannot do internal payment with custom amount specified");
+            }
```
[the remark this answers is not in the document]
elsirion: What was the cause @benthecarman?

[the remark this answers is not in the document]
benthecarman: I needed to change it to `if amount.is_some_and(|a| a.msats != invoice_amount)` to protect against when the amount is set but it is the same as the invoice amount.

benthecarman:
Trying to play with this in harbor and the payment doesn't seem to be claimed. Still figuring out what is going wrong..

FWIW the db migrations seemed to work fine though!

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
joschisan: Its a hack until pruned invoices are supported on cln as well.

benthecarman:
Refactored on the gateway side for making the split payments, still need to test but this copies what they did on the cashu side 

https://github.com/cashubtc/nutshell/pull/492

benthecarman:
Converting to draft for now as I still need to test the functionality end-to-end

benthecarman:
Something I was thinking about for this in incorporating the amount into the operation id. Right now it is just payment_hash and attempt index.

benthecarman:
Successfully did a split lightning payment! On this, running two mutinynet federations with a gateway that supports this if people want to test:

`fed11qgqp78thwden5te0vejkg6tdd9h8gepwvchxxctjd4skutnvdakz7qqpyrkhcqkf44mltuzezxxgx2uj0d5xlylnme2dzmsh5aqfrxmzhc6zwnnvp3w`

`fed11qgqzvfrhwden5te0vejkg6tdd9h8gepwvccjumr0wejhg6r9vf6kcmrn9e3k7mf0qqqjp6yssfzp93ggzzutusy42anq72jpxv5xvu0x7uvyuaq06gx0jf7faudwfx`

benthecarman:
Ended up having to rewrite how `pay_private` implemented for partial payments, may make sense to bring back the old code and only use what I added for partial payments. Also need to make sure the new sending code there is secure against stuck htlcs. timeouts, etc

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -464,73 +473,48 @@ impl ILnRpcClient for GatewayLndClient {
...
-                        return Err(e);
-                    }
+            let sleep_fut = sleep(Duration::from_secs(LND_PAYMENT_TIMEOUT_SECONDS));
+            let payment_result_fut = attempt_private_payments(routes, invoice, &mut client);
+            pin_mut!(sleep_fut, payment_result_fut);
+
+            match futures::future::select(sleep_fut, payment_result_fut).await {
+                Either::Left(_) => {
+                    return Err(LightningRpcError::FailedPayment {
+                        failure_reason: "Timed out waiting for payment to complete".to_string(),
```
m1sterc001guy, line 508: More of a general question not specific to this PR: how should we handle stuck payments that time out here?

If I recall correctly, a timeout here will cause the gateway to cancel the payment and the user to get a refund. But that opens the door for the user getting refunded and the payment still going through, causing the gateway to lose money. Perhaps when this happens, the gateway should not cancel the payment and force the user to get refunded by waiting for the timeout :/

Core lightning has the same problem: https://github.com/fedimint/fedimint/pull/5449

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -847,6 +831,93 @@ impl ILnRpcClient for GatewayLndClient {
...
+        let last = route.hops.last_mut().expect("must have at least one hop");
+
+        last.mpp_record = Some(MppRecord {
+            payment_addr: invoice.payment_secret.to_vec(),
+            total_amt_msat,
+        });
+
+        trace!("Attempting to send payment to route {route:?}");
+
+        debug!("LND payment request sent for invoice {invoice:?}, waiting for payment status...");
```
m1sterc001guy, line 873: I think it would be good to log the whole invoice at the `debug` level, but also probably good to log just the `payment_hash` at the info level. It should be easy for gateway operators to find in the logs when payments are attempted

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -847,6 +831,93 @@ impl ILnRpcClient for GatewayLndClient {
...
+                info!("LND payment request failed for invoice {invoice:?} with {status:?}");
+                LightningRpcError::FailedPayment {
+                    failure_reason: format!("Failed to make outgoing payment {status:?}"),
+                }
+            });
+
+        match htlc_attempt.map(Response::into_inner) {
+            Ok(payment) => match payment.status() {
+                HtlcStatus::Failed => {
+                    info!("LND payment failed for invoice {invoice:?} with {payment:?}");
```
m1sterc001guy, line 893: nit: `error!(?invoice, ?payment, "LND payment failed for invoice");`

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -847,6 +831,93 @@ impl ILnRpcClient for GatewayLndClient {
...
+                    if let Some(failure) = payment.failure {
+                        last_failed_reason = Some(failure.code().as_str_name().to_string());
+                    }
+                }
+                HtlcStatus::InFlight => {
+                    debug!("LND payment is inflight");
+                    continue;
```
m1sterc001guy, line 900: Is each `route` in this for loop a separate complete route to the destination? I'm trying to understand why we want to continue here when a payment is inflight. This will cause `send_to_route_v2` RPC to be called again with a different route right?

benthecarman: ah yeah I thought I was processing a stream, need to modify here

benthecarman:
Fixed up to address the concerns about stuck payments and @m1sterc001guy's other comments. Working on re-testing but getting issues from the `tls` feature in some stuff

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -464,77 +474,163 @@ impl ILnRpcClient for GatewayLndClient {
...
+                        info!("LND payment request failed for invoice {invoice:?} with {status:?}");
+                        LightningRpcError::FailedPayment {
+                            failure_reason: format!("Failed to make outgoing payment {status:?}"),
+                        }
+                    });
+
+                match htlc_attempt.map(Response::into_inner) {
+                    Ok(payment) => match payment.status() {
+                        HtlcStatus::Failed => {
+                            info!("LND payment failed for invoice {invoice:?} with {payment:?}");
```
m1sterc001guy, line 555: Let's change this to `error!`

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -464,77 +474,163 @@ impl ILnRpcClient for GatewayLndClient {
...
+                            if let Some(failure) = payment.failure {
+                                last_failed_reason = Some(failure.code().as_str_name().to_string());
+                            }
+                        }
+                        HtlcStatus::InFlight => {
+                            debug!("LND payment is inflight, tracking...");
+                            let stream = client.router().track_payment_v2(TrackPaymentRequest {
+                                payment_hash: invoice.payment_hash.to_byte_array().to_vec(),
+                                no_inflight_updates: true, // just need payment success/failure
```
m1sterc001guy, line 564: We had this set to `true` before, might be helpful just to `debug` log the payment for inflight updates?

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -464,77 +474,163 @@ impl ILnRpcClient for GatewayLndClient {
...
+                            })?;
+
+                            let mut stream = stream.into_inner();
+                            while let Some(update) = stream.next().await {
+                                match update {
+                                    Ok(payment) => {
+                                        match payment.status() {
+                                            PaymentStatus::InFlight | PaymentStatus::Unknown => {} /* should not happen */
+                                            PaymentStatus::Failed => {
+                                                info!("LND payment failed for invoice {invoice:?} with {payment:?}");
```
m1sterc001guy, line 579: let's do `error!` here too

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -464,77 +474,163 @@ impl ILnRpcClient for GatewayLndClient {
...
+                let last = route.hops.last_mut().expect("must have at least one hop");
+
+                last.mpp_record = Some(MppRecord {
+                    payment_addr: invoice.payment_secret.to_vec(),
+                    total_amt_msat,
+                });
+
+                trace!("Attempting to send payment to route {route:?}");
+
+                debug!("LND payment request sent for invoice {invoice:?}, waiting for payment status...");
```
m1sterc001guy, line 535: Can we also log what route attempt we are on? I could see that being interesting

modules/fedimint-ln-client/src/db.rs
```diff
@@ -225,13 +243,17 @@ pub(crate) fn get_v3_migrated_state(
...
 
-            let new_pay = LightningPayStateMachine {
+            let new_pay = LightningPayStateMachineOld {
                 common,
                 state: LightningPayStates::Funded(v1),
             };
-            let new_sm = LightningClientStateMachines::LightningPay(new_pay);
-            let bytes = new_sm.consensus_encode_to_vec();
-            Ok(Some((bytes, operation_id)))
+            let pay_bytes = new_pay.consensus_encode_to_vec();
```
m1sterc001guy, line 250: nit: could make this a function

modules/fedimint-ln-client/src/db.rs
```diff
@@ -225,13 +243,17 @@ pub(crate) fn get_v3_migrated_state(
...
+            let mut sm_bytes: Vec<u8> = Vec::new();
+            1_u16.consensus_encode(&mut sm_bytes).unwrap();
+            let len = u16::try_from(pay_bytes.len()).unwrap();
+            len.consensus_encode(&mut sm_bytes).unwrap();
+            sm_bytes.extend_from_slice(&pay_bytes);
+            Ok(Some((sm_bytes, operation_id)))
```
m1sterc001guy, line 256: nit: could make this a function

modules/fedimint-ln-client/src/db.rs
```diff
@@ -247,18 +269,67 @@ pub(crate) fn get_v3_migrated_state(
...
+        preimage_auth: common.preimage_auth,
+        invoice: common.invoice,
+        amount: None,
+    };
+
+    let new_sm = LightningClientStateMachines::LightningPay(LightningPayStateMachine {
+        common: new_common,
+        state,
+    });
+    let bytes = new_sm.consensus_encode_to_vec();
```
m1sterc001guy, line 329: nit: Could use the logic from `get_v3_migrated_state` so that if this needs to be modified again we don't need to go back and fix `get_v4_migrated_state`

modules/fedimint-ln-client/src/db.rs
```diff
@@ -247,18 +269,67 @@ pub(crate) fn get_v3_migrated_state(
...
+            len.consensus_encode(&mut sm_bytes).unwrap();
+            sm_bytes.extend_from_slice(&pay_bytes);
+            Ok(Some((sm_bytes, operation_id)))
         }
         _ => Ok(None),
     }
 }
 
+/// Adds the `amount` field to `LightningPayCommon`
+pub(crate) fn get_v4_migrated_state(
```
m1sterc001guy, line 289: I know its cumbersome but would be great to have a test for this, for extra confidence

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -987,17 +1006,40 @@ impl LightningClientModule {
...
         }
 
+        // make sure the invoice amount is set
+        let invoice_amount = invoice
+            .amount_milli_satoshis()
+            .ok_or(anyhow!("MissingInvoiceAmount"))?;
+
         let (pay_type, client_output, contract_id) = if is_internal_payment {
+            if amount.is_some_and(|a| a.msats != invoice_amount) {
+                bail!("Cannot do internal payment with custom amount specified");
```
m1sterc001guy, line 1016: Can we make this (and below) `PayBolt11InvoiceError` values?

m1sterc001guy:
LGTM, comments are mostly about logging and nits. I will sign off after there's some test coverage for the db migration.

More incentive to get MPP payments working when receiving inside Fedimint so Fedimint's can do partial payments to other Fedimints!

benthecarman:
Responded to review and added tests for db migrations

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -417,124 +422,219 @@ impl ILnRpcClient for GatewayLndClient {
...
+                            preimage: payment.preimage,
                         });
                     }
-                    Err(e) => {
-                        info!("LND payment failed for invoice {invoice:?} with {e:?}");
-                        return Err(e);
-                    }
+                },
+                Err(e) => {
+                    info!("LND payment failed for invoice {invoice:?} with {e:?}");
```
m1sterc001guy, line 629: nit: `error!`

m1sterc001guy (dismissed):
A devimint test would be good to add too to validate the LND portion, but I am ok merging as is

benthecarman:
rebased and fixed the nit, don't know how to make a devimint test, happy to learn though

m1sterc001guy (approved):

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -417,124 +422,219 @@ impl ILnRpcClient for GatewayLndClient {
...
                 })?;
 
-            debug!("LND payment does not exist for invoice {invoice:?}, will attempt to pay");
-            let payments = client
+        let mut last_failed_reason: Option<String> = None;
+
+        let start = duration_since_epoch();
+        let mut attempts = 0;
+
+        for mut route in routes.routes {
```
elsirion, line 517: Do we know how this approach performs vs. the old payment code? I'd imagine iterating over a static set of routes without being able to take into account failures that happened between fetching the routes and trying any particular one would be a disadvantage over sendpay.

Generally I'd like to be very cautious here since this code is very central to the payment UX. We should likely try to get review from someone very familiar with LND's API (once we are happy with the code).

elsirion (commented):
The payment code has become a lot more complex, can we maybe factor it in a way that makes it easier to follow (extract functions etc.)? Reviewing the pay changes will already be hard, having one huge fn doesn't make it any simpler.

gateway/ln-gateway/src/lightning/lnd.rs
benthecarman: I believe it is functionally equivalent but not 100% sure. We could just use this when we aren't sending the full amount and use the old code for normal payments

benthecarman:
Refactored to bring back the old sending code when not doing a partial payment. Also split into smaller functions

elsirion:
@benthecarman @TonyGiorgio how important is this to you, do you want to get it into 0.4.0 (not ideal to block) or 0.4.1 (managable imo)?

benthecarman:
This can wait

benthecarman:
rebased

dpc:
Can you verify the "Implemented multi-mint LN payments." from https://github.com/fedimint/fedimint/releases/tag/v0.4.0-rc.0 ? The commit messages was there, but seeing this PR still opened, I'm wondering if I haven't misundertand something.

benthecarman:
@dpc it was done for lnv2 only I think

dpc:
@fedimint/lightning What's the status here? It's been a while.

m1sterc001guy:
I think we should close this, if we want to support it we can for LNv2

Issue #5388, opened 2024-05-29
Support multiple withdrawals per session in recoverytool epochs


bradleystachurski:
The recoverytool epochs method was originally created with the assumption that a maximum of one withdrawal is possible per session. This was a valid assumption for HBBFT, however it is no longer true with AlephBFT.

If more than one withdrawal happens in a session, we won't recover the change output.

https://github.com/fedimint/fedimint/blob/f22eb9decb9abf1ca17282376e394df823bad69a/recoverytool/src/main.rs#L234-L241

PR #5392, opened 2024-05-29
fix: recognize multiple peg-out tweaks in single session

bradleystachurski:
Fixes https://github.com/fedimint/fedimint/issues/5388

- Updates devimint recoverytool test to process multiple withdrawals in single session, which reveals the underlying issue with the epochs method
- Updates the change tweak index for each withdrawal, instead of only once per session outcome


dpc (approved):
LGTM, but

elsirion (approved):

PR #5400, opened 2024-05-31
feat: reduce RAM usage by factor of three, simplify switching trigger logic

joschisan:
This change requires a coordinated shutdown of all guardians at the end of a session such that all of them can restart with the 0.4.0 binary.

With the new defaults the maximum RAM consumption of a peers aleph bft instance is 230mb per peer in case of an attack while it is at most 180mb per peer under the assumption that all peers are correct. In a 3/4 federation we would have 920mb and 720mb respectively.

The new minimum session time will be less variable with a minimum of 3 minutes, and will not increase when a peer goes offline.

This is the last change to core consensus I have planned.

### DIFF
--- a/fedimint-server/src/config/mod.rs
+++ b/fedimint-server/src/config/mod.rs
@@ -46,19 +46,18 @@ pub mod io;
 
 /// The default maximum open connections the API can handle
 const DEFAULT_MAX_CLIENT_CONNECTIONS: u32 = 1000;
-// if all nodes are correct the session will take 45 to 60 seconds. The
-// more nodes go offline the longer the session will take to complete.
-const DEFAULT_BROADCAST_EXPECTED_ROUNDS_PER_SESSION: u16 = 45 * 20;
-const DEFAULT_BROADCAST_ROUND_DELAY_MS: u16 = 50;
-const DEFAULT_BROADCAST_MAX_ROUNDS_PER_SESSION: u16 = 5000;
 
-/// Set of consensus broadcast settings that results in around 10s session time,
-/// that is useful for testing purposes. Not that all the values here together
-/// need to satisfy bunch of constrains in alephbft, so tweaking them
-/// separately will most probably result in panics and other misbehavior.
-const DEFAULT_TEST_BROADCAST_EXPECTED_ROUNDS_PER_SESSION: u16 = 3 * 20;
+/// Consensus broadcast settings that result in 3 minutes session time
+const DEFAULT_BROADCAST_ROUND_DELAY_MS: u16 = 50;
+const DEFAULT_BROADCAST_ROUNDS_PER_SESSION: u16 = 3600;
+
+fn default_broadcast_rounds_per_session() -> u16 {
+    DEFAULT_BROADCAST_ROUNDS_PER_SESSION
+}
+
+/// Consensus broadcast settings that result in 10 seconds session time
 const DEFAULT_TEST_BROADCAST_ROUND_DELAY_MS: u16 = 50;
-const DEFAULT_TEST_BROADCAST_MAX_ROUNDS_PER_SESSION: u16 = 2700;
+const DEFAULT_TEST_BROADCAST_ROUNDS_PER_SESSION: u16 = 200;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 /// All the serializable configuration for the fedimint server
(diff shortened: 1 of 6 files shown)


elsirion (commented):
This will require:
* ~Changing the core consensus version (should happen in this PR imo so we don't forget)~ Not necessary because AlephBFT isn't a core consensus change, it just requires shutdown after a session.
* ~Editing the config to update the consensus version, we should likely automated that on startup (other PR):~
  1. ~Make sure all peers are running the same consensus version~
  2. ~Change our own config (maybe even squeeze in the adding of the ~LNv2 and~ meta modules if not present)~
  3. ~Start normally~

fedimint-server/src/consensus/aleph_bft/finalization_handler.rs
```diff
@@ -1,22 +1,37 @@
+use aleph_bft::{NodeIndex, Round};
 use fedimint_core::PeerId;
 
 use super::data_provider::UnitData;
 
+pub struct OrderedUnit {
```
dpc, line 6: Comments always great to have on structs. :)

dpc (approved):

dpc:
@elsirion @joschisan 

Makes me think we need to up our "release notes" game.

Maybe we can start by creating a CHANEGELOG.md with the most important changes affecting users (like this one)? I don't think we need to do it often enough that merge conflicts will be a problem yet.


BTW. What is going to happen if peers do not stop the Federation at the same time? I assume the whole thing will just get stuck, but there will not be irreparable damage

elsirion:
> BTW. What is going to happen if peers do not stop the Federation at the same time? I assume the whole thing will just get stuck, but there will not be irreparable damage

It really depends on the changes in AlephBFT. Most likely it will crash on startup because the ongoing AlephBFT session data can't be decoded anymore. Having a more robust check would be good though to mitigate the damage of guardians doing stupid things (which they will).

fedimint-server/src/consensus/aleph_bft/finalization_handler.rs
```diff
@@ -1,22 +1,37 @@
...
+    fn unit_finalized(&mut self, creator: NodeIndex, round: Round, data: Option<UnitData>) {
         // the channel is unbounded
         self.sender
-            .try_send((unit_data, super::to_peer_id(creator)))
+            .try_send(OrderedUnit {
+                creator: super::to_peer_id(creator),
+                round,
+                data,
+            })
             .ok();
```
elsirion, line 35: note: We need to use the less stable interface here so we can learn the creator id.

fedimint-server/src/consensus/engine.rs
```diff
@@ -251,11 +252,15 @@ impl ConsensusEngine {
...
             session_index,
-            max_round,
+            self.cfg
+                .consensus
+                .broadcast_rounds_per_session
+                .checked_add(EXP_SLOWDOWN_ROUNDS)
+                .expect("Rounds per session exceed maximum of u16::Max - EXP_SLOWDOWN_ROUNDS"),
             delay_config,
-            Duration::from_secs(100 * 365 * 24 * 60 * 60),
+            Duration::from_secs(10 * 365 * 24 * 60 * 60),
```
elsirion, line 261: nit: Should probably be a constant, denotes the time it would take for a session to reach max rounds, i.e. 10 years.

fedimint-server/src/consensus/engine.rs
```diff
@@ -218,29 +218,30 @@ impl ConsensusEngine {
...
+        const EXP_SLOWDOWN_ROUNDS: u16 = 1000;
+        const BASE: f64 = 1.02;
 
-        let expected_rounds = 3 * self.cfg.consensus.broadcast_expected_rounds_per_session as usize;
-        let max_round = 3 * self.cfg.consensus.broadcast_max_rounds_per_session;
+        let rounds_per_session = self.cfg.consensus.broadcast_rounds_per_session;
         let round_delay = f64::from(self.cfg.local.broadcast_round_delay_ms);
 
-        let exp_slowdown_offset = 3 * expected_rounds;
-
```
elsirion, line 235: Finally cleaning up the `* 3` debacle :laughing:

fedimint-server/src/consensus/engine.rs
```diff
@@ -391,12 +389,14 @@ impl ConsensusEngine {
...
         while signatures.len() < self.keychain.threshold() {
             tokio::select! {
-                unit_data = unit_data_receiver.recv() => {
-                    if let (UnitData::Signature(signature), peer) = unit_data? {
-                        if self.keychain.verify(&header, &signature, to_node_index(peer)){
-                            signatures.insert(peer, signature);
+                ordered_unit = ordered_unit_receiver.recv() => {
+                    let ordered_unit = ordered_unit?;
+
+                    if let Some(UnitData::Signature(signature)) = ordered_unit.data {
```
elsirion, line 395: nit: could use [`let-else` syntax](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) for early return to reduce indentation, e.g.

```rust
let Some(UnitData::Signature(signature)) = ordered_unit.data else { continue; };
```

elsirion (approved):
Some nits, also share some high-level concerns with @dpc, but feel free to merge @joschisan if you'd rather discuss them without blocking the PR.

fedimint-server/src/config/mod.rs
```diff
@@ -122,8 +117,10 @@ pub struct ServerConfigConsensus {
     pub version: CoreConsensusVersion,
     /// Public keys for the atomic broadcast to authenticate messages
     pub broadcast_public_keys: BTreeMap<PeerId, PublicKey>,
-    /// Determines how long a session is expected to run. Has to be less than
-    /// 1000.
+    #[deprecated(
+        since = "0.4.0",
+        note = "This value is deprecated and will be removed in the future"
+    )]
     pub broadcast_expected_rounds_per_session: u16,
```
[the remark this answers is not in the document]
elsirion: Lol, that was a remnant from an abandoned review :see_no_evil: we had a call and I mentioned it instead.

PR #5409, opened 2024-06-01
chore: api calls with call error codes

dpc:
This is a follow-up of the discussion in #5244 , though only the client side.

The motivation here is to allow API calls to return well defined "errors", that the client side can act on, possibly conditionally. We don't have many calls like this currently, because we were avoiding the matter altogether. The `submit_transaction` is the only call like that we have, though the approach taken there turned out to be too constraining, and frustrated by not being able to add a single field for better diagnostics, I created #5238 which started this whole effort.

#### Some rationale of the design

For the client logic to remain secure in the presence of potentially malicious peers, any application-level api errors need to be a subject of client side consensus (query strategies). Changing api response (including errors) is an API breaking change. This goes against the need to return rich data-structures and/or diagnostic information. Therefor error conditions should preferably be simple enumerations (error codes), with any diagnostics specifically not being part of client api result consensus logic.

JsonRpc already allows returning both application level error codes with extra information: https://www.jsonrpc.org/specification#error_object . It makes sense to just use it in an jsonrpc idiomatic way, as it fits our needs and and avoids inventing new serialization schemes for distinguishing successful responses from error conditions.

##### About the changes

The root change is elevating application level errors from being the same as any other errors (connection error, timeouts, server side jsonrpc error, etc.). This is so it can be used to return application level server side errors to the client as first class values.

In a nutshell, where before client api code (especially query strategies) worked on `T` as a server side api response, now it's a `Result<T, CallError>`. Query strategies consider the whole `Result<T, CallError>` as a succesful "response" and might collect the whole thing, gather a threshold of them and return as a consensus on the result to the call, etc.

`CallError` is a bit special. The `impl Eq for CallError` compares only the `code`. All other fields: `message` and `data` are considered diagnostics only. The reason for this is that if the client needs to agree on a consensus, anything it compares becomes effectively set in stone and immutable. If we are to compare `message`, it means it's no longer possible to fix a typo in it without breaking the client side. That's why server side api application errors need to be as small and precise as possible. Adding a new error code is effectively a API breaking change.

If client side ever needs more broader information about the details of error (e.g. submitted tx being rejected), it needs to facilitate it using separate follow-up api code. There are many reason for it. E.g. `submit_transaction` fail on a first input/output with a problem, and can't return all processing errors; or the fact that `submit_transaaction` can't start returning new error codes/data just because new module is available because it's a core-API endpoint. By introducing another call, it's possible to evolve per-case handling, without having to break the wholerror e core API to accommodate it.

As currently all except `submit_transaction` api calls don't use application-side level errors, and in the future most probably will not, a simple `.flatten()` method is provided for the call to turn `Result<Result<T, CallError>, PeerError>` into just `Result<T, PeerError>` like what was returned before. Logic that does want to handle specific error cases of an Api, will just not call `.flatten` and handle `CallError` directly.

Query strategy logic stays mostly intact. A `CallError` is just a part of `T`. Filter/map strategies might just want to call `?` to throw away `CallErrors`.

#####  Follow-up

If this is accepted, I'd like to rotate `submit_transaction` to a new API call, that uses call error codes for rejected transactions, somehow like in #5244, but with re-considered error codes.

Afterwards, we might considered changing `InputError`, `OutputError` and even `OutputOutcome` in the module system, as these might not require consensus encoding anymore.

### DIFF
--- a/fedimint-api-client/src/api.rs
+++ b/fedimint-api-client/src/api.rs
@@ -30,7 +30,7 @@ use fedimint_core::endpoint_constants::{
     STATUS_ENDPOINT, SUBMIT_TRANSACTION_ENDPOINT, VERIFIED_CONFIGS_ENDPOINT,
     VERIFY_CONFIG_HASH_ENDPOINT,
 };
-use fedimint_core::fmt_utils::{AbbreviateDebug, AbbreviateJson};
+use fedimint_core::fmt_utils::AbbreviateDebug;
 use fedimint_core::invite_code::InviteCode;
 use fedimint_core::module::audit::AuditSummary;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
@@ -52,7 +52,11 @@ use itertools::Itertools;
 use jsonrpsee_core::client::{ClientT, Error as JsonRpcClientError};
 use jsonrpsee_core::DeserializeOwned;
 #[cfg(target_family = "wasm")]
+use jsonrpsee_wasm_client::types::ErrorCode;
+#[cfg(target_family = "wasm")]
 use jsonrpsee_wasm_client::{Client as WsClient, WasmClientBuilder as WsClientBuilder};
+#[cfg(not(target_family = "wasm"))]
+use jsonrpsee_ws_client::types::ErrorCode;
 #[cfg(not(target_family = "wasm"))]
 use jsonrpsee_ws_client::{HeaderMap, HeaderValue};
 #[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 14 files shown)


fedimint-api-client/src/api.rs
```diff
@@ -320,16 +429,22 @@ pub trait FederationApiExt: IRawFederationApi {
...
-            .and_then(|v| {
-                serde_json::from_value(v).map_err(|e| PeerError::ResponseDeserialization(e.into()))
-            })
+        {
+            Ok(Ok(v)) => Ok(serde_json::from_value(v)
+                .map_err(|e| PeerError::ResponseDeserialization(e.into()))?),
+            Ok(Err(e)) => Ok(Err(e)),
+            Err(e) => Err(e),
+        }
+        // .and_then(|v| {
```
dpc, line 445: Trash

dpc:
Closing as it got stuck, and I doubt we'll work on it in near future.

PR #5412, opened 2024-06-03
chore: pipe payment hash from InterceptHtlcRequest to InterceptHtlcResponse

tvolk131:
# Background

On the lightning gateway, each incoming HTLC from the underlying lightning node fires out an `InterceptHtlcRequest` to the stream returned by `ILnRpcClient::route_htlcs()`. The gateway responds by calling `ILnRpcClient::complete_htlc()` and passing in an `InterceptHtlcResponse` which includes two pieces of data:

1. How the HTLC should be handled (`InterceptHtlcResponse.action`)
2. Data used to identify the HTLC (`InterceptHtlcResponse.incoming_chan_id` and `InterceptHtlcResponse.htlc_id`)

The `incoming_chan_id` and `htlc_id` fields combined act as a unique identifier for a given HTLC, essentially mapping an `InterceptHtlcResponse` to the `InterceptHtlcRequest` it is responding to. We use these two fields instead of simply using the payment hash because, while we don't currently support MPP, someone could attempt to pay using MPP, so the payment hash is not guaranteed to be unique to a single HTLC.

# What we're adding

In addition to `incoming_chan_id` and `htlc_id`, we're adding `payment_hash` as another field that can be used for uniquely mapping an `InterceptHtlcRequest` to an `InterceptHtlcResponse`.

# Why this is needed

This change will make it much easier to support an `ldk-node` lightning gateway. The folks who are working on `ldk-node` are planning on adding payment interception but _not_ HTLC interception. For intercepted payments, we will be given the payment hash but no HTLC IDs. However, the payment hash will be guaranteed to be unique since `ldk-node` will only trigger a `PaymentClaimable` event once the entire payment is received. Essentially we'll be pretending that incoming payments are single HTLCs even though they could be MPPs under the hood.

### DIFF
--- a/gateway/ln-gateway/proto/gateway_lnrpc.proto
+++ b/gateway/ln-gateway/proto/gateway_lnrpc.proto
@@ -91,7 +91,7 @@ message PayInvoiceResponse {
 
 message InterceptHtlcRequest {
   // The HTLC payment hash.
-  // Value is not guaranteed to be unique per intercepted HTLC
+  // HTLCs corresponding to the same payment will have the same payment hash.
   bytes payment_hash = 1;
 
   // The incoming HTLC amount in millisatoshi.
@@ -162,6 +162,10 @@ message InterceptHtlcResponse {
 
   // The index of the incoming htlc in the incoming channel
   uint64 htlc_id = 5;
+
+  // The HTLC payment hash.
+  // HTLCs corresponding to the same payment will have the same payment hash.
+  bytes payment_hash = 6;
 }
 
 message GetRouteHintsResponse {
--- a/gateway/ln-gateway/src/gateway_module_v2/complete_sm.rs
+++ b/gateway/ln-gateway/src/gateway_module_v2/complete_sm.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 
+use bitcoin_hashes::Hash;
 use fedimint_client::sm::{State, StateTransition};
 use fedimint_client::DynGlobalClientContext;
 use fedimint_core::core::OperationId;
@@ -44,6 +45,7 @@ impl CompleteStateMachine {
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub struct CompleteSMCommon {
     pub operation_id: OperationId,
+    pub payment_hash: bitcoin_hashes::sha256::Hash,
     pub incoming_chan_id: u64,
     pub htlc_id: u64,
 }
@@ -73,6 +75,7 @@ impl State for CompleteStateMachine {
             CompleteSMState::Completing(result) => vec![StateTransition::new(
                 Self::await_completion(
                     context.clone(),
+                    self.common.payment_hash,
                     self.common.incoming_chan_id,
                     self.common.htlc_id,
                     result.clone(),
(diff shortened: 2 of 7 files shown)


gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -157,6 +157,10 @@ message InterceptHtlcResponse {
     Forward forward = 3;
   }
 
+  // The HTLC payment hash.
+  // Value is not guaranteed to be unique per intercepted HTLC
+  bytes payment_hash = 6;
```
m1sterc001guy, line 162: Maybe put this at the bottom so that the numbers line up, weird to see 6 out of order

gateway/ln-gateway/src/gateway_module_v2/complete_sm.rs
```diff
@@ -44,6 +45,7 @@ impl CompleteStateMachine {
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub struct CompleteSMCommon {
     pub operation_id: OperationId,
+    pub payment_hash: bitcoin_hashes::sha256::Hash,
```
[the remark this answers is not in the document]
m1sterc001guy: Oh I suppose since this in LNv2 we can skip it

m1sterc001guy (approved):
Makes sense to me. Eventually, once we support MPP for all implementations, we probably should only use the payment hash. There's no need to send multiple `InterceptHtlcRequest` messages to the gateway, each LN implementation should just wait until it received all the HTLCs for the payment and then contact the gateway.

gateway/ln-gateway/src/lib.rs
```diff
@@ -662,25 +662,26 @@ impl Gateway {
                         break;
                     }
 
+                    let payment_hash =
+                        bitcoin_hashes::sha256::Hash::from_slice(&htlc_request.payment_hash)
```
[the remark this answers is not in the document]
tvolk131: I'm pretty sure that was already the case. In this snippet from the previous code:

```
                    if let Ok((payload, client)) = self
                        .get_payload_and_client_v2(
                            htlc_request
                                .payment_hash
                                .clone()
                                .try_into()
                                .expect("32 bytes"),
                            htlc_request.incoming_amount_msat,
                        )
                        .await
                    {
```

We're casting `htlc.payment_hash` into a `[u8; 32]`. In this PR we're instead attempting to turn it into a `bitcoin_hashes::sha256::Hash` which should also only fail if the field length != 32 bytes, so we're not changing any assumptions about this field.

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: Done.

gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -162,6 +162,10 @@ message InterceptHtlcResponse {
 
   // The index of the incoming htlc in the incoming channel
   uint64 htlc_id = 5;
+
+  // The HTLC payment hash.
+  // Value is not guaranteed to be unique per intercepted HTLC
```
m1sterc001guy, line 167: This comment confused me a little initially, since its phrased in the negative. I think something like "HTLCs corresponding to the same payment will have the same payment hash" is more clear.

m1sterc001guy (dismissed):

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: Done.

m1sterc001guy (approved):

PR #5417, opened 2024-06-03
chore: dump citems dump on consensus mismatch

dpc:
Follow-up to https://github.com/fedimint/fedimint/pull/5386/files#r1619463731

The goal here is to be able to do binary search using logs of peer nodes that have a mismatch and find the element(s) that are mismatched and at very least known which modules might have been involved, before attempting a full scale debugging with dbtool etc.

That's why the index and short-hash are printed first, and then it's some compact general info about txid, modules and/or total length.

If I'm reading it right, there is no guarantee that any peer with a majority-correct checksum/signature will notice invalid signature of another peer, but over 3 peers, the chance is high, and it's better than nothing.

Looks like this:

image

### DIFF
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -40,7 +40,7 @@ use crate::consensus::db::{
     AcceptedItemKey, AcceptedItemPrefix, AcceptedTransactionKey, AlephUnitsPrefix,
     SignedSessionOutcomeKey, SignedSessionOutcomePrefix,
 };
-use crate::consensus::debug::DebugConsensusItem;
+use crate::consensus::debug::{DebugConsensusItem, DebugConsensusItemCompact};
 use crate::consensus::transaction::process_transaction_with_dbtx;
 use crate::fedimint_core::encoding::Encodable;
 use crate::metrics::{
@@ -385,6 +385,8 @@ impl ConsensusEngine {
 
         let mut signatures = BTreeMap::new();
 
+        let items_dump = tokio::sync::OnceCell::new();
+
         // We collect the ordered signatures until we either obtain a threshold
         // signature or a signed session outcome arrives from our peers
         while signatures.len() < self.keychain.threshold() {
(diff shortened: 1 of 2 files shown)


fedimint-server/src/consensus/engine.rs
```diff
@@ -387,6 +389,8 @@ impl ConsensusEngine {
 
         let mut signatures = BTreeMap::new();
 
+        let items_dump = tokio::sync::OnceCell::new();
```
elsirion, line 392: Why `OnceCell` and not a simple mutable boolean? Does the `&mut bool` cause lifetime issues inside the `select!`?

dpc: No. I just though `OnceCell` expresses better that we want it to happen at max once. :D @elsirion

elsirion: So be it :man_shrugging: probably not hurting anything, just slightly unconventional I guess (and it saves a mut var!)

elsirion (approved):

dpc:
@joschisan :+1: / :-1: ?

joschisan:
> If I'm reading it right, there is no guarantee that any peer with a majority-correct checksum/signature will notice invalid signature of another peer, but over 3 peers, the chance is high, and it's better than nothing.

You are reading this right. If the peer with the incorrect checksum is lagging sufficiently behind the correct peers your logging will not be triggered for any of the peers, however, this is relatively unlikely I think.

We already have debugging for consensus items in debug.rs. Please move to debug.rs and consider unification with DebugConsensusItem.

joschisan (changes requested):
We already have debugging for consensus items in debug.rs. Please move ConsensusItemCompactDumpFmt to debug.rs and consider if it can be unified with DebugConsensusItem.

dpc:
> We already have debugging for consensus items in debug.rs.

@joschisan 

Right. Good point. I moved it there, but I didn't attempt to unify, because the other one is kind of long and multiline, and this one supposed to be super-compact. I added a comment explaining why a second one is needed.

Feel free to tweak them if you have any preferences. I just think that since this one might dump dozens or even hundreds of items it should be ideally very compact. I have also put a hex prefix first, so doing a binary search between e.g. two windows with log dumps should be convenient.

joschisan (approved):

elsirion (dismissed):

elsirion:
Oh no, merge conflict :cry:

joschisan (approved):

PR #5421, opened 2024-06-04
chore: enforce clippy nursery rules in fedimint-core

tvolk131:
Enforce most clippy nursery rules in fedimint-core. I disabled a few that seemed annoying. One rule that I'm on the fence about (that's currently enabled in this PR) is [missing_const_for_fn](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_const_for_fn) since removing `const` from a function later would break API compatibility. Lmk if anyone has thoughts on that. If people like the restrictions in this PR, I'll do a sweep across the rest of the rust codebase and enable it everywhere.

### DIFF
--- a/fedimint-core/src/config.rs
+++ b/fedimint-core/src/config.rs
@@ -382,20 +382,18 @@ impl ClientConfig {
     }
 
     pub fn get_module<T: Decodable + 'static>(&self, id: ModuleInstanceId) -> anyhow::Result<&T> {
-        if let Some(client_cfg) = self.modules.get(&id) {
-            client_cfg.cast()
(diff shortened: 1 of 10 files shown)


elsirion:
Let's not make everything const, const generics are still very limited and I don't really see a huge benefit of a lot of stuff being const. If we need it we can always add it in.

tvolk131:
> Let's not make everything const, const generics are still very limited and I don't really see a huge benefit of a lot of stuff being const. If we need it we can always add it in.

Fixed!

dpc (approved):

PR #5425, opened 2024-06-04
feat(client): consolidate excessive amount of notes in a given tier

dpc:
Backport of #4922 

We had some people reporting hitting tx too large problems, so decided to attempt a backport.

However, I'm not sure how I feel about landing it, as it feels like a big change to just add to a minor update to already matured 0.3.


elsirion:
I'd say let's not do anything major on 0.3 anymore.

PR #5426, opened 2024-06-04
chore: enforce clippy nursery rules

tvolk131:
Follw-up to #5421, but enable rules across the entire rust codebase


elsirion:
I think we are hitting a point of diminishing returns re stronger clippy rules.

tvolk131:
> I think we are hitting a point of diminishing returns re stronger clippy rules.

Sg!

PR #5427, opened 2024-06-06
Keychain Refactor

joschisan:

### DIFF
--- a/fedimint-server/src/consensus/aleph_bft/keychain.rs
+++ b/fedimint-server/src/consensus/aleph_bft/keychain.rs
@@ -2,9 +2,10 @@ use std::collections::BTreeMap;
 use std::io::Write;
 
 use aleph_bft::Keychain as KeychainTrait;
+use bitcoin_hashes::Hash;
 use fedimint_core::encoding::Encodable;
 use fedimint_core::session_outcome::SchnorrSignature;
-use fedimint_core::{secp256k1, BitcoinHash, NumPeersExt, PeerId};
+use fedimint_core::{secp256k1, NumPeersExt, PeerId};
 use secp256k1::hashes::sha256;
 use secp256k1::{schnorr, KeyPair, Message, PublicKey};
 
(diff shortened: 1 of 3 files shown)


fedimint-server/src/consensus/aleph_bft/keychain.rs
```diff
@@ -15,6 +16,7 @@ pub struct Keychain {
     peer_id: PeerId,
     public_keys: BTreeMap<PeerId, PublicKey>,
     keypair: KeyPair,
+    tag: sha256::Hash,
```
dpc, line 19: @joschisan I don't want to block your PRs, but consensus code is like the most sophisticated, important part of our codebase with least amount of understanding coverage among contributors, so while you're improving it, please, please add a comment to every struct and field you're touching. At least the non-trivial ones, or at least `struct`s. It  pays off big time when you're the reader trying to figure out something from the codebase.

Here, I'm opening a PR, I see a "tag", and I have no idea what I am looking at. Had the `Keychain` had a one sentence docstring, it would make understanding it much easier.

fedimint-server/src/consensus/aleph_bft/keychain.rs
```diff
@@ -26,28 +28,15 @@ impl Keychain {
                 .private
                 .broadcast_secret_key
                 .keypair(secp256k1::SECP256K1),
+            tag: cfg.consensus.broadcast_public_keys.consensus_hash(),
```
dpc, line 31: `consensus_hash` is more descriptive, `tag` is unclear to me. Might be just me, but just reporting my thoughts as I'm reading through it.

fedimint-server/src/consensus/aleph_bft/keychain.rs
joschisan: I agree, renamed and added a comment. I actually have no further changes to core consensus planned at the moment, so agree with you here.

dpc (approved):

PR #5429, opened 2024-06-06
refactor: make conversion to num peers explicit

dpc:
It always bothered me that we have e.g. `peer_urls: BTreeMap<PeerId, String>` and we can write code like:

```
peer_urls.max_evil();
```

and while reading code like that I couldn't help but to think, "how do we go from a map to some consensus security calculations like that ...".

So at some point I added the `struct NumPeers` which is supposed to be an explicit newtype that carries the meaning of "how many peers we have", which seems like a decent place for methods like `max_evil` and `threshold`.

I recently attempted to remove the `NumPeersExt` where the methods like `max_evil` are implemented for collections, but it turned out that this was too much of a inconvenience. It is nice to be able to smoothly go from the data that carries number of peers implicitly (like a map of peer urls) to `NumPeers`. It just shouldn't be so implicit.

So in this PR I implement the final solution - going through an explicit `NumPeersExt::to_num_peers()`.

So a:

```
peer_urls.max_evil();
```

turns into:

```
peer_urls.to_num_peers().max_evil();
```

and in a lot of places `peer_urls.to_num_peers()` can replace existing `NumPeers::new(peer_urls.to_num_peers().total())`.

It's a little thing, but recently I've seen in the comments @joschisan and @elsirion talking about then need to clean up these num peer stuff, so here it is.

I didn't finish it (converting all the places that now complain with `deprecated` warning). I only converted a subset, just to give some sense how would it look like in practice. I first want to get some feedback if this seems OK, or maybe there are other opinions and ideas.

### DIFF
--- a/fedimint-api-client/src/api.rs
+++ b/fedimint-api-client/src/api.rs
@@ -42,8 +42,8 @@ use fedimint_core::time::now;
 use fedimint_core::transaction::{SerdeTransaction, Transaction, TransactionSubmissionOutcome};
 use fedimint_core::util::SafeUrl;
 use fedimint_core::{
-    apply, async_trait_maybe_send, dyn_newtype_define, runtime, NumPeers, NumPeersExt, OutPoint,
-    PeerId, TransactionId,
+    apply, async_trait_maybe_send, dyn_newtype_define, runtime, NumPeersExt, OutPoint, PeerId,
+    TransactionId,
 };
 use fedimint_logging::LOG_CLIENT_NET_API;
 use futures::stream::FuturesUnordered;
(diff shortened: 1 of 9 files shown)


joschisan:
I like this refactor, would have done it the same. We should make it a convention then that functions, like the contractor of DkgRunner always take the NumPeer struct and call threshold, one_honest etc. inside the function next to the appropriate context. 

Furthermore, for NumPeers we could add a constructor that asserts that the u16 is neither two or three as those are not supported as federation sizes.

dpc:
> like the contractor of DkgRunner always take the NumPeer struct and call threshol

Attempted in #5430 but it needs some discussion, especially some unexpected things were discovered.

If there's an agreement on this one, let's land and I'll follow through with getting rid of deprecated messages.

joschisan (approved):

PR #5430, opened 2024-06-07
24 06 07 explicit num peers dkg runner

dpc:


fedimint-server/src/config/mod.rs
```diff
@@ -478,16 +477,6 @@ impl ServerConfig {
             "Peer {} running distributed key generation...", our_id
         );
 
-        // BFT uses a lower threshold of signing keys (f+1)
```
dpc, line 481: This appeared unused?

dpc:
Nah. Landed something different instead.

PR #5431, opened 2024-06-07
feat: add ldk gateway to mprocs

tvolk131:
Requires #5554


elsirion:
Call: Got merged as smaller PRs

PR #5432, opened 2024-06-07
feat: track ordering latency

joschisan:

### DIFF
--- a/fedimint-server/src/consensus/aleph_bft/data_provider.rs
+++ b/fedimint-server/src/consensus/aleph_bft/data_provider.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeSet;
+use std::time::Instant;
 
 use fedimint_core::config::ALEPH_BFT_UNIT_BYTE_LIMIT;
 use fedimint_core::encoding::Encodable;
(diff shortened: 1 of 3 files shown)


Kodylow:
@elsirion someone should look at this

elsirion:
The first session after a restart would lead to some bonkers latency values being reported. @douglaz @dpc is that something we ok with for prometheus metrics? For the most part they should be sane and we could probably filter out these spikes after restarts.

If we are not ok with wrong values being reported sometimes then maybe adding some additional events that get sent over the channel (e.g. "new session started") could just turn off latency reporting till the next fresh session starts.

dpc:
Oh. Why is this closed? I was looking forward to landing.

> @douglaz @dpc is that something we ok with for prometheus metrics? For the most part they should be sane and we could probably filter out these spikes after restarts.

I suggested we should send over the channel the length of encoded consensus item, so we can cheaply and relatively reliably discard incorrect items. After this, everything should be self-correcting, and very rare incorrect reports would to tolerable, IMO. 

@elsirion @joschisan

elsirion:
> relatively reliably discard incorrect items

*relatively* being the important word here :laughing: what's the probability of items having the same size? I guess it's rather high. I'd probably go with a CRC sum instead *if we want to fix the restart problem at al* (same data size and low computational overhead).

dpc:
> relatively being the important word here 

To be out of sync in the first place requires for the peer to restart right after sending a message but before receiving it. This already limits the problem to something like one in a >10000. So all that we need is to not persist in that state and self-heal in few messages. Sure messages can have the same size, but not over and over forever. Any two consecutive messages of a different size will .


CRC or some other checksum indeed even better, but I just doubt it being necessary. Might also send last length + 4 bytes etc. 

@elsirion

joschisan:
@elsirion @dpc this turned out not to be as simple as I anticipated and I closed it for now as I wanted to focus on other stuff. I can and probably will revisit this at a later point but in the current state I do not want go ahead.

Issue #4752, opened 2024-03-27
MVP Tor support in the client


elsirion:
Part of #2610, but without stream separation by operation. @oleonardolima already did most of the work for an MVP version, so let's get it in. The full version can follow later.

oleonardolima:
As I've mentioned in today's dev call, I still need to test and debug the connection to hidden services peers thoroughly, and  I think it's best to split the work done into two/three PRs:

I did update the description of #2610, extracting it from there, the current work in progress (step/milestone 2):

- **feat: add initial Tor support, without `OperationId` stream isolation:** https://github.com/fedimint/fedimint/issues/4752

    - [ ] feat: add initial Tor support, connection to `hostname` https://github.com/fedimint/fedimint/pull/5053 
    - [ ] feat: add initial Tor support, connection to `onion addresses` (hidden services)
    - [ ] feat: add initial Tor support, support for `onion addresses` into `InviteCode`

cc @elsirion (@benthecarman I'll focus the updates in this issue for now :))

oleonardolima:
update: FWIW, I've used harbor for testing alongside the fedimint-cli, building on top of the #5443,  here is harbor's branch/PR if anyone is curious about it https://github.com/MutinyWallet/harbor/pull/89

oleonardolima:
update: I think this initial mvp is almost there, I was able to do a test federation setup with hidden services, using the non-tls script / setup approach.

It only required to set up the hidden service beforehand, using the onion address during federation setup (the `SafeUrl`/`Url` handles it pretty well), and the onion addresses will be part of the config and returned in the api_endpoints responses, encoded in the federationId, and so on.

If some guardian wants to expose both through an onion address, it seems possible with some traefik settings, but the domain will still be exposed as it's part of the response in api_endpoints 🤔 

FYI @elsirion @benthecarman

Kodylow:
yaaaaaaaay

Issue #2610, opened 2023-05-31
Tor Support Tracking Issue


oleonardolima:
## Description
In the full-fledged version, fedimint clients should have full Tor support, being able to connect to federation peers through Tor network, either if they are hosted in clearnet (hostnames) or through hidden services (onion addresses).

It should also have `OperationId` stream isolation, to ensure that different `OperationId`'s rely on and have different connections, that said different Tor circuits, it's a great feature to avoid deanonymization due to correlations made by the connections and `OperationId`'s being requested by the client.

It can be split into some steps/milestones:

1. **refactor: jsonrpsee and add support to generic data streams and proxying:**
    - [x] https://github.com/paritytech/jsonrpsee/issues/1162
    - [x] https://github.com/paritytech/jsonrpsee/issues/870
    - [x] https://github.com/paritytech/jsonrpsee/pull/1168
    - [x] https://github.com/paritytech/jsonrpsee/issues/1241
    - [x] https://github.com/fedimint/fedimint/pull/3454
    - [x] https://github.com/fedimint/fedimint/pull/3934
    - [x] https://github.com/fedimint/fedimint/pull/3440

2. **feat: add initial Tor support, without `OperationId` stream isolation:** https://github.com/fedimint/fedimint/issues/4752
    - [x] feat: add initial Tor support, connection to `hostname` + `onion addresses (hidden services)` https://github.com/fedimint/fedimint/pull/5443
    - [ ] feat: add initial Tor support, support for `onion addresses` into `InviteCode`

4. **feat: add full-fledged Tor support, with `OperationId` stream isolation:**
    - [ ] refactor: make `WsFederationApi` based on `OperationId`, becoming`WsOperationIdFederationApi`
    - [ ] refactor: make `WsFederationApi` clients based on `Connector` variant of `WsOperationIdFederationApi`
    - [ ] ~~refactor: add new `Tor` as `Connector` variant, with `WsClient` over arti `DataStream`~~

--

### Chores:
- [x] chore: [`arti-hyper`](https://tpo.pages.torproject.net/core/doc/rust/arti_hyper/index.html) assessment integration and usage to support SOCKS5 proxy with jsonrpsee
- [x] chore: [`arti-client`](https://tpo.pages.torproject.net/core/doc/rust/arti_client/index.html#more-advanced-usage) assessment on json-rpc library support and integration

### References:
- https://github.com/fedimint/fedimint/issues/15
- https://github.com/fedimint/fedimint/issues/391
[...]

Issue #89, opened 2022-05-11
CI: Check that LN Gateway receives funds


elsirion:
The integration test should wait for the LN Gateway to actually receive the e-cash tokens for its LN payment. This probably requires creating a new (authenticated) API endpoint for LN Gateway operators to query its balance.

matthiasdebernardini:
You are referring to this integration test right?
https://github.com/fedimint/minimint/blob/0a2b472440e9226c15d7828f069f68c290a74506/scripts/integrationtest.sh

matthiasdebernardini:
https://github.com/fedimint/minimint/tree/master/integrationtests

because there is also this

matthiasdebernardini:
https://nigiri.vulpem.com I think we should make use of this, very nice way to run nodes

elsirion:
> You are referring to this integration test right? https://github.com/fedimint/minimint/blob/0a2b472440e9226c15d7828f069f68c290a74506/scripts/integrationtest.sh

Correct, the other one is really new, ideally both would test this.

> https://nigiri.vulpem.com/ I think we should make use of this, very nice way to run nodes

I'm not sure how well it works for automated testing, I kinda hope that @wiredhikari's nix-bitcoin integration project will eventually make other orchestration tools (like the current hacky bash script) obsolete for our CI.

justinmoon:
I think @elsirion is talking about adding an API endpoint [here](https://github.com/fedimint/minimint/blob/21be38357248c162ad3d940810fc73499a065d6d/ln-gateway/src/bin/ln_gateway.rs#L63)

After thinking about this a bit, it might be best to just add unauthenticated ` /balance` to the gateway http server for now. Then we can add authentication in the future once we define how a lightning node can enroll itself as a gateway for a federation. They'd get credentials here, provide an email or something for account recovery, and the federation would save all the public information about the users can find the node. But I don't think authenticating the API makes much sense until that point.

justinmoon:
Regarding the broader question of how to authenticate a gateway HTTP request, maybe we could just have they node pubkey sign the HTTP request or something. Then we don't have to worry about recovering credentials. (idea inspired by this recent conference talk https://twitter.com/skwp/status/1534265886005248006 by @ skwp).

matthiasdebernardini:
I am trying to get ln gateway configgen to run but it gets stuck in this function

[...]

PR #5443, opened 2024-06-11
feat(tor): add initial mvp tor support connection to hostnames and onion addresses

oleonardolima:
partially completes #4752 

### Description
This PR adds support to establish connections to federation peers over the Tor network, it uses the [`arti-client`](https://crates.io/crates/arti-client), an official Rust implementation of Tor.

It's a step towards the initial MVP Tor support, only adding support to connect to hostnames in this PR, connection to onion addresses (federation peers hosted as hidden services) will be addressed in another PR, please refer to #2610 and #4752 for detailed information. 

### How
In order to know through which transport layer the client wants to establish the connection to the federation peer, at this moment in time being either Tcp (clearnet) or the Tor network, this PR adds a new `Connector` enum, with `Connector::Tcp` and `Connector::Tor` variants. 

The `Connector` allows the `WsFederationApi`, and `FederationPeerClient` to build the `jsonrpsee` `WsClient` with the proper transport layer, using the specific`JsonRpcClient` method.

It also adds the required structures and auxiliary methods, such as `Target` and its methods, to establish a `Tls` connection, over the `Tor` anonymized `DataStream`, it's still required as the client can still be connecting to hosted peer using TLS, even though it's using the Tor network as the base transport. 

### Notes to the reviewers
I'm still not sure if adding new methods to `JsonRpcClient` trait is the best way to go, or if having another trait (e.g. `AnonymizedJsonRpcClient` would be better, but:

- How can we handle two different traits for the same `WsFederationApi` ?
- Should we then build the `WsFederationApi` on top of `Connector`, in different ways for each variant?
- A `Builder` pattern would improve usability and reduce "duplicated" code ? 

Looking for feedback and opinions on the approach.

### Testing Approach Suggestion

I was discussing with @Kodylow about some live testing (not unit/integration tests) approaches for this PR, I can suggest some testing approaches for this PR, at least the ones that I've been using in the development loop, please also let me know of any different approaches that you think are useful.

I usually use both: fedimint-cli and harbor (with the help of [PR #89](https://github.com/MutinyWallet/harbor/pull/89) ).

If you're looking to test it with fedimint-cli, you can:
1. checkout to this branch to get Tor support changes.
2. either compile the fedimint-cli to a binary, or use it through cargo run.
3. you can enhance the log level for both jsonrpsee, arti and tor crates, then you should be able to see the Tor circuit, connections being established.
4. you can also try to run it from inside a network-constrained environment, to check that the expected connections are going through Tor.

If you're looking to test it with harbor or any other application that relies on fedimint client (you can check the changes made on [PR #89](https://github.com/MutinyWallet/harbor/pull/89)), double-check that you're using the most up-to-date commit/branch (this one) in the project, and do check that you are still able to connect to your guardian, and do the operations, deposit, reissue, ... through Tor.


misc/git-hooks/pre-commit
```diff
@@ -75,8 +75,9 @@ function check_check_forbidden_dependencies() {
         return 1
       fi
     done
+    # FIXME: (@leonardo) temporally allow the `openssl`, it's currently being brought by tor-proto -...-> native-tls.
```
dpc, line 78: :fearful:

oleonardolima:
I'll squash some of the commits, but it should be ready for some scrutiny 😅 cc @elsirion @dpc

misc/git-hooks/pre-commit
dpc: Oh, BTW. This file is auto-generated...

The right way to change stuff here is to tweak flakebox config like I do in: https://github.com/fedimint/fedimint/pull/5473/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R83

dpc: Actually, the original file in nix/check-forbidden-deps.sh

fedimint-cli/src/envs.rs
```diff
@@ -5,6 +5,9 @@ pub const FM_CLIENT_DIR_ENV: &str = "FM_CLIENT_DIR";
 // Env variable to set the peer id of the guardian
 pub const FM_OUR_ID_ENV: &str = "FM_OUR_ID";
 
+// Env variable to set the connector choice (e.g Tcp / Tor)
+pub const FM_CONNECTOR_ENV: &str = "FM_CONNECTOR";
```
dpc, line 9: I don't know if the "connector" thing is a good way to go about it in the UI. Maybe we should just have `--use-tor` and `FM_USE_TOR`  or something?

dpc:
Thought: (in the future) we might want to put tor behind a feature flag, because I bet it brings tons of extra deps, while in some applications it doesn't make sense to use it.

fedimint-cli/src/envs.rs
oleonardolima: Sure, that looks like a better approach.

misc/git-hooks/pre-commit
oleonardolima: Great, I'll update that then.
I now noticed that the comment is outdated, it's due to a problem on the latest jsonrpsee release, which might be fixed by this issue: https://github.com/paritytech/jsonrpsee/issues/1407

maan2003: if add openssl please gate it behind a cargo feature flag, mobile apps build break otherwise

fedimint-cli/src/envs.rs
maan2003: env is nice to have

gateway/ln-gateway/src/client.rs
```diff
@@ -95,8 +95,9 @@ impl GatewayClientBuilder {
                 .open(root_secret)
                 .await
         } else {
-            let client_config =
-                fedimint_api_client::download_from_invite_code(&invite_code).await?;
+            let client_config = fedimint_api_client::api::net::Connector::default()
+                .download_from_invite_code(&invite_code)
```
maan2003, line 99: mention todo: don't use default connector here

fedimint-cli/src/lib.rs
```diff
@@ -241,7 +246,16 @@ impl Opts {
         api_secret: &Option<String>,
     ) -> CliResult<DynGlobalApi> {
         let our_id = self.our_id.ok_or_cli_msg("Admin client needs our-id set")?;
-        Ok(DynGlobalApi::from_config_admin(cfg, api_secret, our_id))
+
+        let connector = if self.use_tor {
+            Connector::Tor
+        } else {
+            Connector::default()
```
maan2003, line 253: maybe tor should be function as well `Connector::tor()`

fedimint-api-client/src/lib.rs
```diff
@@ -23,65 +24,74 @@ pub mod api;
...
-pub async fn download_from_invite_code(invite_code: &InviteCode) -> anyhow::Result<ClientConfig> {
-    debug!("Downloading client config from {:?}", invite_code);
+impl Connector {
+    /// Tries to download the [`ClientConfig`] from the federation with an
+    /// specified [`Connector`] variant, attempts to retry ten times before
+    /// giving up.
+    pub async fn download_from_invite_code(
+        &self,
+        invite_code: &InviteCode,
+    ) -> anyhow::Result<ClientConfig> {
```
maan2003, line 34: maybe this is fine as free function

```rust
pub async fn download_from_invite_code(invite_code: &InviteCode, connector: &Connector)
```

maan2003 (approved):
just some nits, otherwise looks fine

oleonardolima:
> just some nits, otherwise looks fine

Thanks, I'll address it

fedimint-api-client/src/api/mod.rs
```diff
@@ -533,42 +366,69 @@ impl AsRef<dyn IGlobalFederationApi + 'static> for DynGlobalApi {
...
         // PeerIds are used only for informational purposes, but just in case, make a
         // big number so it stands out
         let peer_id = PeerId::from(1024);
         GlobalFederationApiWithCache::new(
-            WsFederationApi::new(vec![(peer_id, url)], api_secret).with_self_peer_id(peer_id),
+            WsFederationApi::new(&Connector::default(), vec![(peer_id, url)], api_secret)
+                .with_self_peer_id(peer_id),
         )
         .into()
     }
```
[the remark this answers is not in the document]
oleonardolima: @maan2003 Do you you have any thoughts on this as well ?

gateway/ln-gateway/src/client.rs
oleonardolima: @maan2003 Do you think we should already address this in this PR, or just keep as TODO ?

fedimint-cli/src/lib.rs
oleonardolima: Sure, I'll use that instead

gateway/ln-gateway/src/client.rs
maan2003: just leave it as todo

misc/git-hooks/pre-commit
oleonardolima: It'll be fixed by https://github.com/fedimint/fedimint/pull/5544, depends on new jsonrpsee release with https://github.com/paritytech/jsonrpsee/pull/1419

oleonardolima: I fixed it upstream, and now with the latest jsonrpsee release, this is not a problem anymore.

oleonardolima:
As we now have fixes for `jsonrpsee` upstream, we can support both hostnames and onion addresses in a single PR, only 
 adding a few lines in a new commit. I'll rebase with API  announcement changes.

fedimint-core/src/util/mod.rs
```diff
@@ -159,6 +159,22 @@ impl SafeUrl {
...
+
+    pub fn is_onion_address(&self) -> bool {
+        let onion_suffix = ".onion";
+        let host = self
+            .host_str()
+            .expect("It should've asserted for `host` on construction");
+        let suffix_start = host
+            .len()
+            .checked_sub(onion_suffix.len())
+            .expect("It should not fail!");
```
elsirion, line 171: Why shouldn't this fail? E.g. `a.io` is a valid domain shorter than `.onion`. Could you just use [`str::ends_with`](https://doc.rust-lang.org/std/primitive.str.html#method.ends_with)?

oleonardolima: I was trying to follow what arti's team did, but ends_with will make it simpler for sure.

fedimint-api-client/Cargo.toml
```diff
@@ -38,6 +38,9 @@ tokio-rustls = { version = "0.26.0", default-features = false, features = ["logg
 webpki-roots = { version = "0.26.3" }
 rustls-pki-types = { version = "1.7.0" }
 arti-client = { version = "0.20.0", features = ["tokio", "rustls", "onion-service-client"], default-features = false }
+# We need to pin this arti's `curve25519-dalek` dependency, due to `https://rustsec.org/advisories/RUSTSEC-2024-0344` vulneraliblity
+# It's been updated by https://gitlab.torproject.org/tpo/core/arti/-/merge_requests/2211, should be removed in next release.
+curve25519-dalek = { version = ">=4.1.3" }
```
oleonardolima, line 43: @dpc any other specific way I should pin a dependency instead?

dpc: `= "4.1.3"`, is the same as `= ">=4.1.3"`, IIRC, so this is not really "pinning" anything.

Also I don't think one can pin indirect dependency. The line here **will most probably** make dependency unification ensure this or newer version. @oleonardolima

oleonardolima: > [...] The line here will most probably make dependency unification ensure this or newer version [...]

Yes, that's what I'm looking for. Asked because I wasn't sure if it required some other Nix-related change. All good, then.

fedimint-cli/src/envs.rs
oleonardolima: I updated it to expect the `--use-tor` and kept the `FM_USE_TOR` environment variable too.

fedimint-api-client/src/lib.rs
oleonardolima: I did this way to reduce the number of fns that needs to keep passing the `connector` parameter around.

I'm fine with reverting it to a free function though.

fedimint-core/src/util/mod.rs
oleonardolima: I've fixed it to use .ends_with() instead.

fedimint-core/src/util/mod.rs
```diff
@@ -159,6 +159,17 @@ impl SafeUrl {
...
         self.0.join(input).map(SafeUrl)
     }
+
+    // It can be removed to use `is_onion_address()` implementation,
+    // once https://gitlab.torproject.org/tpo/core/arti/-/merge_requests/2214 lands.
+    #[allow(clippy::case_sensitive_file_extension_comparisons)]
+    pub fn is_onion_address(&self) -> bool {
+        let host = self
+            .host_str()
+            .expect("It should've asserted for `host` on construction");
```
elsirion, line 169: How sure are you about that? Couldn't a maliciously crafted (i.e. relative URL) invite code trigger this?

elsirion (dismissed):
Looks great!

fedimint-core/src/util/mod.rs
oleonardolima: Oh, got it! You're correct, a manually crafted invite code could purposely trigger this.

I think returning an error and handling it, would be the proper way to go. Any other thoughts?

elsirion: Actually, let's just return false: if there is no host that's broken and it will fail later anyway.

oleonardolima: Sure, I did fix in 6df8232eb83e1ae338607c2ed627554bc861af7b

elsirion: https://github.com/fedimint/fedimint/commit/6df8232eb83e1ae338607c2ed627554bc861af7b isn't the cleanest solution (you `unwrap_or_default` an `Option<&str>` instead of an `Option<bool>`, so you get an empty string that you then check the suffix of) but good enough imo.

elsirion (approved):

fedimint-core/src/util/mod.rs
oleonardolima: Got it, I can probably make it cleaner in a future PR.

PR #5446, opened 2024-06-11
fix: fix LN deprecation warnings

m1sterc001guy:
https://github.com/fedimint/fedimint/pull/4024 added some deprecation warnings. This PR just suppresses them since we don't need to see them during build.

### DIFF
--- a/modules/fedimint-ln-client/src/pay.rs
+++ b/modules/fedimint-ln-client/src/pay.rs
@@ -21,47 +21,64 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::{debug, error, warn};
 
+pub use self::lightningpay::LightningPayStates;
 use crate::api::LnFederationApi;
 use crate::{set_payment_result, LightningClientContext, LightningClientStateMachines, PayType};
 
 const RETRY_DELAY: Duration = Duration::from_secs(1);
 
-#[cfg_attr(doc, aquamarine::aquamarine)]
-/// State machine that requests the lightning gateway to pay an invoice on
-/// behalf of a federation client.
-///
-/// ```mermaid
-/// graph LR
-/// classDef virtual fill:#fff,stroke-dasharray: 5 5
-///
-///  CreatedOutgoingLnContract -- await transaction failed --> Canceled
-///  CreatedOutgoingLnContract -- await transaction acceptance --> Funded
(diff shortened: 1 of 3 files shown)


modules/fedimint-ln-client/src/pay.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(deprecated)]
```
dpc, line 1: A crate level `allow` can let multiple new ones slip it, and also no one will remember to clean it up after it is no longer needed. IMO, we should put it on every line that is using the deprecated symbol.

m1sterc001guy: So I tried that initially, and I could only get the usage warning to go away, but for some reason when adding `#[allow(deprecated)]` on the declaration, the derive statement on the enum causes problems and the `#[allow(deprecated)]` macro is ignored. 

Maybe that's good enough?

dpc: Let's try what we for start.

dpc (approved):

modules/fedimint-ln-client/src/pay.rs
```diff
@@ -26,42 +27,53 @@ use crate::{set_payment_result, LightningClientContext, LightningClientStateMach
...
-    Refundable(LightningPayRefundable),
-    Refund(LightningPayRefund),
-    #[deprecated(
-        since = "0.4.0",
-        note = "Pay State Machine does not need to wait for the refund tx to be accepted"
-    )]
-    Refunded(Vec<OutPoint>),
-    Failure(String),
+#[allow(deprecated)]
+pub mod lightningpay {
```
dpc, line 31: `pub(crate)` or `pub(super)`.

dpc (changes requested):
actually, nits.

modules/fedimint-ln-client/src/pay.rs
dpc: And a comment why we need this module.

m1sterc001guy: done

m1sterc001guy: ah unfortunately needs to be `pub` since it is used in the testing crate

dpc (approved):

dpc (approved):

PR #5448, opened 2024-06-12
fix: add unicode 3.0 license to list of allowed licenses

tvolk131:
This is causing our CI dependency audit to fail

For details, you can run `cargo-deny check licenses` locally. The following deps currently use the Unicode-3.0 license:

`icu_collections`
`icu_locid`
`icu_locid_transform`
`icu_locid_transform_data`
`icu_normalizer`
`icu_normalizer_data`
`icu_properties`
`icu_properties_data`
`icu_provider`
`icu_provider_macros`
`litemap`
`tinystr`
`writeable`
`yoke`
`yoke-derive`
`zerofrom`
`zerofrom-derive`
`zerovec`
`zerovec-derive`

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -110,7 +110,8 @@ allow = [
     "ISC",
     "CC0-1.0",
     "Unlicense",
-    "MITNFA"
+    "MITNFA",
+    "Unicode-3.0",
 ]
 # List of explicitly disallowed licenses
 # See https://spdx.org/licenses/ for list of possible licenses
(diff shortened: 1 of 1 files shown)


m1sterc001guy:
I'm not familiar with the Unicode 3.0 license. Are there any implications for allowing code with this license?

tvolk131:
I read the license and it seems fine to me but another pair of eyes would be good. I dug more into where all of these deps came from, and the culprit is #5444 - specifically bumping `url` from v2.5.0 to v2.5.1 since that bumps its version of `idna` from v0.5.0 to v1.0.0 which introduces `icu_normalizer` and `icu_properties` which is where the unicode license cascade starts. We could just revert `url` back to v2.5.0.

dpc:
> The Unicode License is a permissive open source license approved by the Open Source Initiative (OSI).

dpc (approved):

PR #5449, opened 2024-06-12
feat: CLN Pruned Invoices

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/278

Adds the ability for CLN to pay pruned invoices.

I kept the old implementation of paying the full BOLT11 invoice, since the new payment implementation is new and might not be as robust as the old implementation. Eventually once we are satisfied it is robust, we can remove the legacy payment implementation.

### DIFF
--- a/gateway/ln-gateway/proto/gateway_lnrpc.proto
+++ b/gateway/ln-gateway/proto/gateway_lnrpc.proto
@@ -19,6 +19,11 @@ service GatewayLightning {
   rpc PayInvoice(PayInvoiceRequest) returns (PayInvoiceResponse) {}
 
   /*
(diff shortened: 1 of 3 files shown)


joschisan (approved):

elsirion (commented):
Am I missing something or is this impl only trying one route and fails if that doesn't work? To re-implement a naive `pay` we should at least loop over different routes (till we hit a time limit), keeping track of failed routes that can be given to the `get_route` RPC call as the `exclude` parameter so that we get  hopefully better one the next time.

m1sterc001guy:
Yes you are correct, I will be addressing in a follow up.

Issue #5450, opened 2024-06-12
Checkpoint database when completing an AlephBFT session


m1sterc001guy:
[Rocksdb has the ability to create database checkpoints in a separate directory](https://docs.rs/rocksdb/latest/rocksdb/checkpoint/struct.Checkpoint.html)

We could use this to create a checkpoint of the database when an AlephBFT session is completed. This way, if there is ever a consensus failure, guardians can recover by wiping their database and restarting from the last checkpoint where they were in consensus.

We'll want to limit the number of checkpoints retained to save on disk space. This should probably be an optional feature configurable by an environment variable in case the checkpoint operation is too heavy/expensive.

elsirion:
To be clear: rolling back to an old DB snapshot is a last-resort operation that can allow double-spends and mess up client state in some cases (depending on the module, e.g. suddenly missing LN contracts). So any such rollback should only be available to expert users.

PR #5451, opened 2024-06-12
feat: Checkpoint database when completing an AlephBFT session

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/5450

I think this might be useful for federations that potentially corrupt their database. The checkpoint retention  is configurable, so this feature can be turned off if guardian operators don't want the backups.

### DIFF
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -9,7 +9,7 @@ use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use async_trait::async_trait;
 use fedimint_core::db::{
     IDatabaseTransactionOps, IDatabaseTransactionOpsCore, IRawDatabase, IRawDatabaseTransaction,
@@ -157,6 +157,12 @@ impl IRawDatabase for RocksDb {
 
         rocksdb_tx
     }
+
+    fn checkpoint(&self, backup_path: &Path) -> Result<()> {
+        let checkpoint = rocksdb::checkpoint::Checkpoint::new(&self.0)?;
+        checkpoint.create_checkpoint(backup_path)?;
+        Ok(())
+    }
 }
 
 #[async_trait]
@@ -164,6 +170,10 @@ impl IRawDatabase for RocksDbReadOnly {
     type Transaction<'a> = RocksDbReadOnlyTransaction<'a>;
     async fn begin_transaction<'a>(&'a self) -> RocksDbReadOnlyTransaction<'a> {
         RocksDbReadOnlyTransaction(&self.0)
+    }
+
+    fn checkpoint(&self, _backup_path: &Path) -> Result<()> {
+        Err(anyhow!("Cannot checkpoint a read only database"))
     }
 }
 
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -9,7 +9,7 @@ use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
 
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
 use fedimint_core::db::{
     IDatabaseTransactionOps, IDatabaseTransactionOpsCore, IRawDatabase, IRawDatabaseTransaction,
@@ -172,8 +172,10 @@ impl IRawDatabase for RocksDbReadOnly {
         RocksDbReadOnlyTransaction(&self.0)
     }
 
-    fn checkpoint(&self, _backup_path: &Path) -> Result<()> {
-        Err(anyhow!("Cannot checkpoint a read only database"))
+    fn checkpoint(&self, backup_path: &Path) -> Result<()> {
+        let checkpoint = rocksdb::checkpoint::Checkpoint::new(&self.0)?;
+        checkpoint.create_checkpoint(backup_path)?;
+        Ok(())
     }
 }
 
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -1,8 +1,8 @@
 use std::collections::BTreeMap;
+use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
-use std::{env, fs};
 
 use aleph_bft::Keychain as KeychainTrait;
 use anyhow::{anyhow, bail};
@@ -44,7 +44,6 @@ use crate::consensus::db::{
 };
 use crate::consensus::debug::{DebugConsensusItem, DebugConsensusItemCompact};
 use crate::consensus::transaction::process_transaction_with_dbtx;
-use crate::envs::{FM_DB_CHECKPOINT_FREQ_ENV, FM_DB_CHECKPOINT_SESSION_DIFFERENCE_ENV};
 use crate::fedimint_core::encoding::Encodable;
 use crate::metrics::{
     CONSENSUS_ITEMS_PROCESSED_TOTAL, CONSENSUS_ITEM_PROCESSING_DURATION_SECONDS,
@@ -71,6 +70,7 @@ pub struct ConsensusEngine {
     pub connection_status_channels: Arc<RwLock<BTreeMap<PeerId, PeerConnectionStatus>>>,
     pub task_group: TaskGroup,
     pub data_dir: PathBuf,
+    pub checkpoint_retention: u64,
 }
 
 impl ConsensusEngine {
@@ -95,7 +95,7 @@ impl ConsensusEngine {
     pub async fn run_single_guardian(&self, task_handle: TaskHandle) -> anyhow::Result<()> {
         assert_eq!(self.num_peers(), NumPeers::from(1));
 
-        self.create_checkpoint_directory()?;
+        self.initialize_checkpoint_directory(self.get_finished_session_count().await)?;
 
         while !task_handle.is_shutting_down() {
             let session_index = self.get_finished_session_count().await;
@@ -168,7 +168,7 @@ impl ConsensusEngine {
         )
         .await;
 
-        self.create_checkpoint_directory()?;
+        self.initialize_checkpoint_directory(self.get_finished_session_count().await)?;
 
         while !task_handle.is_shutting_down() {
             let session_index = self.get_finished_session_count().await;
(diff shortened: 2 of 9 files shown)


fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +480,79 @@ impl ConsensusEngine {
...
+        let checkpoint_frequency: u64 = env::var(FM_DB_CHECKPOINT_FREQ_ENV)
+            .unwrap_or(10.to_string())
+            .parse()
+            .expect("FM_DB_CHECKPOINT_FREQ var is invalid");
+
+        if session_index % checkpoint_frequency == 0 {
+            let checkpoint_dir = self.data_dir.join("db_checkpoints");
+
+            if !checkpoint_dir.exists() {
+                if let Err(e) = fs::create_dir_all(&checkpoint_dir) {
```
dpc, line 498: Would make it once somewhere, unconditionally to avoid having to check over and over.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +480,79 @@ impl ConsensusEngine {
...
+                    warn!(target: LOG_CONSENSUS, ?checkpoint_dir, ?e, "Could not create checkpoint directory");
```
dpc, line 499: This should always be possible, we should just fail loudly if we can't do it.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +480,79 @@ impl ConsensusEngine {
...
+            .expect("FM_DB_CHECKPOINT_SESSION_DIFFERENCE var is invalid");
+        let session_threshold_index = session_index.saturating_sub(session_diff);
+
+        for entry in fs::read_dir(checkpoint_dir)? {
+            let entry = entry?;
+            let session_idx: u64 = entry
+                .file_name()
+                .into_string()
+                .map_err(|_| anyhow!("Could not convert OsString to String"))?
+                .parse()?;
```
dpc, line 547: Probably better to skip entries that don't parse to a number, just in case some external tool etc. left some trash in there.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +480,79 @@ impl ConsensusEngine {
...
+            if !checkpoint_dir.exists() {
+                if let Err(e) = fs::create_dir_all(&checkpoint_dir) {
+                    warn!(target: LOG_CONSENSUS, ?checkpoint_dir, ?e, "Could not create checkpoint directory");
+                    return;
+                }
+            }
+
+            let session_checkpoint_dir = checkpoint_dir.join(format!("{session_index}"));
+
+            match self.db.checkpoint(&session_checkpoint_dir) {
```
dpc, line 506: Wrap in a `TimeReporter` so we can check how long it takes. Based on that we can judge how much a slowdown it is.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +480,79 @@ impl ConsensusEngine {
...
+        checkpoint_dir: &Path,
+    ) -> anyhow::Result<u64> {
+        // By default, cleanup any checkpoint that is older than 50 sessions
+        let session_diff: u64 = env::var(FM_DB_CHECKPOINT_SESSION_DIFFERENCE_ENV)
+            .unwrap_or(50.to_string())
+            .parse()
+            .expect("FM_DB_CHECKPOINT_SESSION_DIFFERENCE var is invalid");
+        let session_threshold_index = session_index.saturating_sub(session_diff);
+
+        for entry in fs::read_dir(checkpoint_dir)? {
```
dpc, line 541: Instead of iterating every time, we could use some more adaptive strategy.

E.g. we could generate a random number between 0 and session different, and if it exist, delete it. This would prune most checkpoints, leaving out some snapshot that could be used to faster sync. Wrapping the whole in a TimeReporter probably a good idea too.

m1sterc001guy: Just realized that we don't need to iterate at all. Since we know the frequency and the current session, we should know the index of the session that needs to be deleted. Can just check if that exists, delete it, and be done.

m1sterc001guy:
Confirmed that checkpoints are incremental, since they take advantage of Rocksdb's internal structure to avoid doing a heavy copy each checkpoint: https://rocksdb.org/blog/2015/11/10/use-checkpoints-for-efficient-snapshots.html

TLDR: A hard link will be made for SSTables that are the same as the current database and a full copy is made for differing files. Manifest and log are copied each checkpoint.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +484,81 @@ impl ConsensusEngine {
...
+    }
+
+    /// Creates a backup of the database in the checkpoint directory. The
+    /// frequency of the checkpoints is controlled by
+    /// `FM_DB_CHECKPOINT_FREQ_ENV`. These checkpoints can be used to
+    /// restore the database in case the federation falls out of consensus.
+    fn checkpoint_database(&self, session_index: u64) {
+        // By default, checkpoint the database every 10 sessions
+        let checkpoint_frequency: u64 = env::var(FM_DB_CHECKPOINT_FREQ_ENV)
+            .unwrap_or(10.to_string())
```
dpc, line 506: I would default to `1`. Seems like a safest option.

`0` or `off` should turn it off completely, in case it turns out too slow.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +484,81 @@ impl ConsensusEngine {
...
+        Ok(())
+    }
+
+    /// Creates a backup of the database in the checkpoint directory. The
+    /// frequency of the checkpoints is controlled by
+    /// `FM_DB_CHECKPOINT_FREQ_ENV`. These checkpoints can be used to
+    /// restore the database in case the federation falls out of consensus.
+    fn checkpoint_database(&self, session_index: u64) {
+        // By default, checkpoint the database every 10 sessions
+        let checkpoint_frequency: u64 = env::var(FM_DB_CHECKPOINT_FREQ_ENV)
```
dpc, line 505: Moving reading this to a new function, and in some previous performance-sensitive code, I used `Once` to make sure it's as fast as possible.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +484,81 @@ impl ConsensusEngine {
...
+            .unwrap_or(10.to_string())
+            .parse()
+            .expect("FM_DB_CHECKPOINT_FREQ var is invalid");
+
+        if session_index % checkpoint_frequency == 0 {
+            let checkpoint_dir = self.data_dir.join("db_checkpoints");
+            let session_checkpoint_dir = checkpoint_dir.join(format!("{session_index}"));
+
+            {
```
dpc, line 514: Thought: it might be asking for trouble, but these two should be able to run in parallel.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +484,81 @@ impl ConsensusEngine {
...
+                if let Err(e) =
+                    Self::delete_old_database_checkpoints(session_index, &checkpoint_dir)
+                {
+                    warn!(target: LOG_CONSENSUS, ?e, "Could not delete old checkpoints");
+                }
+            }
+        }
+    }
+
+    /// Iterates through the db checkpoint directory and deletes the checkpoint
```
dpc, line 538: No longer iterates.

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +484,81 @@ impl ConsensusEngine {
...
+    /// equal to `session_index` - `session_diff`.
+    fn delete_old_database_checkpoints(
+        session_index: u64,
+        checkpoint_dir: &Path,
+    ) -> anyhow::Result<()> {
```
dpc, line 543: Thought: should we do the iteration one on start, just to avoid leftovers when changing retention settings?

fedimint-server/src/envs.rs
```diff
@@ -1,3 +1,10 @@
 /// The env var for maximum open connections the API can handle
 pub const FM_MAX_CLIENT_CONNECTIONS_ENV: &str = "FM_MAX_CLIENT_CONNECTIONS";
 pub const FM_PEER_ID_SORT_BY_URL_ENV: &str = "FM_PEER_ID_SORT_BY_URL";
+
+/// Environment variable for the session count determining when to cleanup old
+/// checkpoints.
+pub const FM_DB_CHECKPOINT_SESSION_DIFFERENCE_ENV: &str = "FM_DB_CHECKPOINT_SESSION_DIFFERENCE";
```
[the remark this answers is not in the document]
m1sterc001guy: Removed the frequency one and renamed the other to `FM_DB_CHECKPOINT_RETENTION`

fedimint-server/src/consensus/engine.rs
m1sterc001guy: Yes, good idea. Added to `initialize_checkpoint_directory`

fedimint-server/src/consensus/engine.rs
m1sterc001guy: Removed

fedimint-server/src/consensus/engine.rs
m1sterc001guy: Removed entirely, always checkpoint ever session now.

fedimint-server/src/consensus/engine.rs
m1sterc001guy: Moved to `run`

fedimint-server/src/consensus/engine.rs
m1sterc001guy: I'm leaving this for now, I think it will be fast enough without it.

fedimint-rocksdb/src/lib.rs
```diff
@@ -165,6 +171,10 @@ impl IRawDatabase for RocksDbReadOnly {
     async fn begin_transaction<'a>(&'a self) -> RocksDbReadOnlyTransaction<'a> {
         RocksDbReadOnlyTransaction(&self.0)
     }
+
+    fn checkpoint(&self, _backup_path: &Path) -> Result<()> {
+        Err(anyhow!("Cannot checkpoint a read only database"))
```
dpc, line 176: Thought: why not?

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +484,89 @@ impl ConsensusEngine {
...
+            debug!(
+                ?current_session,
+                "Removing database checkpoints up to `current_session`"
+            );
+
+            for checkpoint in fs::read_dir(checkpoint_dir)?.flatten() {
+                // Validate that the directory is a session index
+                if let Ok(file_name) = checkpoint.file_name().into_string() {
+                    if let Ok(session) = file_name.parse::<u64>() {
+                        if session < current_session - self.checkpoint_retention {
```
dpc, line 503: Possibility of  underflow in debug node?

fedimint-server/src/consensus/engine.rs
```diff
@@ -472,6 +484,89 @@ impl ConsensusEngine {
             .expect("This is the only place where we write to this key");
     }
 
+    /// Creates the directory within the data directory for storing the database
+    /// checkpoints or deletes checkpoints before `current_session` -
+    /// `checkpoint_retention`.
+    fn initialize_checkpoint_directory(&self, current_session: u64) -> anyhow::Result<()> {
+        let checkpoint_dir = self.data_dir.join("db_checkpoints");
```
dpc, line 491: Nit: would make a `fn db_checkpoints_dir() -> PathBuf` to avoid possibility of typos

dpc: And maybe `fn db_checkpoint_dir(session_num: u64) -> PathBuf` maybe. Seems like it repeats couple of times.

fedimint-server/src/consensus/mod.rs
```diff
@@ -131,6 +138,11 @@ pub async fn run(
         );
     }
 
+    let checkpoint_retention: u64 = env::var(FM_DB_CHECKPOINT_RETENTION_ENV)
+        .unwrap_or(FM_DB_CHECKPOINT_RETENTION_DEFAULT.to_string())
+        .parse()
+        .expect("FM_DB_CHECKPOINT_SESSION_DIFFERENCE var is invalid");
```
dpc, line 144: Previous name still here? Possibly a sign to use the const here too.

dpc (commented):
Landable, but enough nits and suggestions so not landing just yet.

fedimint-server/src/consensus/engine.rs
m1sterc001guy: Good catch, added check

fedimint-rocksdb/src/lib.rs
m1sterc001guy: Added :)

fedimint-server/src/consensus/engine.rs
m1sterc001guy: Done

fedimint-server/src/consensus/mod.rs
```diff
@@ -33,18 +35,23 @@ use tracing::log::warn;
...
 use crate::consensus::engine::ConsensusEngine;
+use crate::envs::FM_DB_CHECKPOINT_RETENTION_ENV;
 use crate::net;
 use crate::net::api::{ApiSecrets, RpcHandlerCtx};
 
 /// How many txs can be stored in memory before blocking the API
 const TRANSACTION_BUFFER: usize = 1000;
 
+// How many checkpoints from the current session should be retained on disk.
+const FM_DB_CHECKPOINT_RETENTION_DEFAULT: u64 = 1;
```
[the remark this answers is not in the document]
m1sterc001guy: Done

fedimint-server/src/consensus/mod.rs
m1sterc001guy: Fixed, added constant as well

dpc (approved):

dpc (approved):

PR #5454, opened 2024-06-13
refactor: move amount into issuance request

joschisan:

### DIFF
--- a/modules/fedimint-mint-client/src/backup.rs
+++ b/modules/fedimint-mint-client/src/backup.rs
@@ -2,8 +2,7 @@ use fedimint_client::module::recovery::{DynModuleBackup, ModuleBackup};
 use fedimint_client::module::ClientDbTxContext;
 use fedimint_core::core::{IntoDynInstance, ModuleInstanceId};
 use fedimint_core::encoding::{Decodable, Encodable};
-use fedimint_core::{Amount, OutPoint, Tiered, TieredMulti};
-use serde::{Deserialize, Serialize};
+use fedimint_core::{OutPoint, Tiered, TieredMulti};
 
 use super::MintClientModule;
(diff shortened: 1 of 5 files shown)


modules/fedimint-mint-client/src/backup.rs
```diff
@@ -41,10 +40,10 @@ impl EcashBackup {
 ///
 /// Used to speed up and improve privacy of ecash recovery,
 /// by avoiding scanning the whole history.
-#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug, Encodable, Decodable)]
+#[derive(Clone, PartialEq, Eq, Debug, Encodable, Decodable)]
 pub struct EcashBackupV0 {
     spendable_notes: TieredMulti<SpendableNote>,
-    pending_notes: Vec<(OutPoint, Amount, NoteIssuanceRequest)>,
+    pending_notes: Vec<(OutPoint, NoteIssuanceRequest)>,
```
m1sterc001guy, line 46: In order to maintain backwards compatibility, each place where we touch an `Encodable` struct, we will unfortunately need to write a db migration to remove the `Amount`. It's definitely needed here, it looks like there are some other structs too (e.g `MintRecoveryState`).

The lightning module has some examples on how to do this, [but they mainly migrate state machines](https://github.com/fedimint/fedimint/blob/master/modules/fedimint-ln-client/src/db.rs). This is just migrating a struct, so hopefully it will be easier. You just need to write a migration that reads the old struct from disk, then re-writes it as the new struct.

elsirion: It's even worse since this is a backup struct and not a SM: we'll have to support the old backup for a long time since we need to retain the ability to read old backups. Why remove that amount here anyway?

PR #5455, opened 2024-06-13
fix: devimint jit esplora waits for esplora server to be ready

tvolk131:
Necessary for ldk lightning gateway, since `ldk-node` panics if the esplora server it relies on is not able to respond to requests when the node starts up

### DIFF
--- a/devimint/Cargo.toml
+++ b/devimint/Cargo.toml
@@ -17,10 +17,7 @@ axum = { version = "0.7.5", features = ["tracing"] }
 bitcoincore-rpc = { workspace = true }
 clap = { version = "4.5.7", features = ["derive", "env", "std", "help", "usage", "error-context", "suggestions"], default-features = false }
 cln-rpc = { workspace = true }
-esplora-client = { version = "0.6.0", default-features = false, features = [
-    "async",
-    "async-https-rustls",
-] }
+esplora-client = { workspace = true }
 fedimint-aead = { path = "../crypto/aead" }
 fedimintd = { path = "../fedimintd" }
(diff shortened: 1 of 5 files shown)


devimint/Cargo.toml
```diff
@@ -17,6 +17,10 @@ axum = { version = "0.7.5", features = ["tracing"] }
 bitcoincore-rpc = { workspace = true }
 clap = { version = "4.5.7", features = ["derive", "env", "std", "help", "usage", "error-context", "suggestions"], default-features = false }
 cln-rpc = { workspace = true }
+esplora-client = { version = "0.6.0", default-features = false, features = [
```
dpc, line 20: If we are using it elsewhere already, can we make it a workspace dep?

devimint/src/devfed.rs
```diff
@@ -136,7 +136,26 @@ impl DevJitFed {
...
+
+                poll("esplora server ready", || async {
+                    client
+                        .get_fee_estimates()
+                        .await
+                        .map_err(|e| ControlFlow::Continue(anyhow::anyhow!(e)))?;
+
+                    Ok(())
+                })
+                .await?;
```
dpc, line 156: Nit: Could be a method on something or a standalone function.

dpc (commented):
LGTM, but have some suggestions. :D

devimint/Cargo.toml
tvolk131: Done.

devimint/src/devfed.rs
tvolk131: Done.

PR #5459, opened 2024-06-14
feat: url safe json encoded invite codes and ecash

joschisan:
The encoding unifies our cash with Cashu. Furthermore the url safe base64 encoding makes it easy to copy and paste a token from and into chats, email etc. Using json makes it extendable and human readable given any standard url safe decoder.

Furthermore, this will save us just short of 500 lines when we remove the old encoding and also allow us to rid ourselves of TieredMulti.

### DIFF
--- a/fedimint-core/src/invite_code.rs
+++ b/fedimint-core/src/invite_code.rs
@@ -74,6 +74,30 @@ impl InviteCode {
         s
     }
 
+    pub fn from_map(
+        peer_to_url_map: &BTreeMap<PeerId, SafeUrl>,
+        federation_id: FederationId,
+        api_secret: Option<String>,
+    ) -> Self {
+        let max_size = peer_to_url_map.to_num_peers().max_evil() + 1;
+        let mut code_vec: Vec<InviteCodePart> = peer_to_url_map
+            .iter()
+            .take(max_size)
+            .map(|(peer, url)| InviteCodePart::Api {
+                url: url.clone(),
+                peer: *peer,
+            })
+            .collect();
+
+        code_vec.push(InviteCodePart::FederationId(federation_id));
+
+        if let Some(api_secret) = api_secret {
+            code_vec.push(InviteCodePart::ApiSecret(api_secret));
+        }
+
+        InviteCode(code_vec)
+    }
+
     /// Constructs an [`InviteCode`] which contains as many guardian URLs as
     /// needed to always be able to join a working federation
     pub fn new_with_essential_num_guardians(
@@ -190,6 +214,10 @@ impl FromStr for InviteCode {
     type Err = anyhow::Error;
 
     fn from_str(encoded: &str) -> Result<Self, Self::Err> {
+        if let Ok(invite_code_v2) = InviteCodeV2::decode_base64(encoded) {
+            return invite_code_v2.into_v1();
+        }
+
         let (hrp, data) = bech32::decode(encoded)?;
 
         ensure!(hrp == BECH32_HRP, "Invalid HRP in bech32 encoding");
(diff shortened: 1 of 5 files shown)


joschisan:
Testing different encodings gives us the following sizes for five notes: 

JSON size 1513, CBOR size 1315, Binary size 648, RMP size 980

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -317,6 +317,11 @@ impl FromStr for OOBNotes {
     /// Decode a set of out-of-band e-cash notes from a base64 string.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let s: String = s.chars().filter(|&c| !c.is_whitespace()).collect();
+
+        if let Ok(notes_v2) = OOBNotesV2::decode_base64(&s) {
+            return notes_v2.into_v1();
+        }
+
```
elsirion, line 324: This breaks en/decoding round-trips, but that's ok I guess. We might want to separate the encoding structs from the semantic structs eventually (e.g. have v1 and v2 structs that can both be converted from/to a common notes struct).

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -377,6 +382,60 @@ impl OOBNotes {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Encodable, Decodable)]
+pub struct OOBNoteV2 {
+    pub amount: Amount,
+    pub sig: Signature,
+    pub key: KeyPair,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Encodable, Decodable)]
+pub struct OOBNotesV2 {
+    pub mint: InviteCodeV2,
```
elsirion, line 394: The `mint` field will be encoded as JSON and not as a base64 string, right?

elsirion:
The resulting encoding (from one of the test cases):

20240618_14h17m25s_grim

fedimint-core/src/invite_code.rs
```diff
@@ -261,4 +270,64 @@ mod tests {
...
+
+        InviteCode::from_str(&encoded).expect("Failed to decode to legacy");
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Encodable, Decodable)]
+pub struct InviteCodeV2 {
+    pub id: FederationId,
+    pub peers: BTreeMap<PeerId, SafeUrl>,
+    pub api_secret: Option<String>,
```
elsirion, line 298: Can you annotate this to use the default value (`None`) if not present and not serialize if not present? https://stackoverflow.com/a/53900684

elsirion (commented):
Looks mostly good, some nits

modules/fedimint-mint-client/src/lib.rs
joschisan: Yes.

fedimint-core/src/invite_code.rs
```diff
@@ -261,4 +270,64 @@ mod tests {
...
+        let json = &serde_json::to_string(self).expect("Encoding to JSON cannot fail");
+        let base_64 = base64_url::encode(json);
+
+        format!("fedimintA{base_64}")
+    }
+
+    pub fn decode_base64(s: &str) -> anyhow::Result<Self> {
+        ensure!(s.starts_with("fedimintA"), "Invalid Prefix");
+
+        let invite_code: Self = serde_json::from_slice(&base64_url::decode(&s[9..])?)?;
```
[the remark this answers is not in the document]
joschisan: Exactly

fedimint-core/src/invite_code.rs
```diff
@@ -261,4 +270,64 @@ mod tests {
...
+            .next()
+            .ok_or_else(|| anyhow!("Invite Code has no peer"))?;
+
+        Ok(InviteCode::new(
+            url.clone(),
+            *peer,
+            self.id,
+            self.api_secret,
+        ))
+    }
```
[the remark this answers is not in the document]
joschisan: fixed.

elsirion (approved):

elsirion:
Marking for further review since even minor bugs could delay deployment a lot when caught too late (we need to wait a bit after the first release with decoding support before switching to v2 as the default encoding).

Issue #5453, opened 2024-06-13
RBF withdrawal causes failed audit


bradleystachurski:
The existing test for rbf withdrawals succeeds, however if there is more than one UTXO the audit fails.

This is reproduced in https://github.com/bradleystachurski/fedimint/tree/sandbox/rbf-withdraw-investigation

Run with `just test-ci-all` (rbf withdraw test is the only test run)
```
00:00:18 thread 'tokio-runtime-worker' panicked at /home/nix/fedimint/fedimint-server/src/consensus/engine.rs:550:9:
00:00:18 Balance sheet of the fed has gone negative, this should never happen! - Balance Sheet -
00:00:18       -8564.000|DummyFundsKeyV1(PublicKey(4e014a3d82ed4f2a125a498a3834b54afee32a91bff694a3876b3b89530b9152120a64654fb4f9dc65
d85c14eb163280a6842eaabbe7135e90175b5f07be850f))
00:00:18        -218.000|UnsignedTransactionKey(0xba6c5dd4b6f44126a3010e7cd5fb9dcd7636341e03de0a9d0ad2c899bdc9b83d)
00:00:18       +4782.000|PendingTransactionKey(0x3c285f63d979eac5b491501d92d061d33a389f4267aff052c09a4d6ab4b8b300)
00:00:18       -4000.000|Net assets (sats)
```

https://github.com/fedimint/fedimint/blob/d995d9d51908230401ede563e385a23d31671393/modules/fedimint-wallet-tests/tests/tests.rs#L256

PR #5465, opened 2024-06-15
fix: use ancestor inputs first in RBF withdrawals

bradleystachurski:
Fixes https://github.com/fedimint/fedimint/issues/5453

The implications of this bug are severe, so please review with caution. Fixing the implementation for RBF withdrawals is one approach, however it's worth considering if disabling RBF withdrawals is a better approach. If we do go with this fix, it requires all peers to shutdown and upgrade at the same time, which is already planned for 0.4. This is due to a change in the coin selection algorithm which will cause a consensus failure if peers are on different versions.

## Context

For RBF transactions we need to first spend the inputs used in the original transaction, then consider additional UTXOs if the original inputs are insufficient to cover the new feerate. Our coin selection algorithm has a bug that constructs the transaction using new UTXOs for RBF transactions. This creates two separate on-chain transactions that can be mined since we are not spending any of the same inputs as the ancestor transaction. Thankfully, the audit fails preventing the user from double spending the ecash, however this will halt any federation that has multiple UTXOs where a user initiates an RBF withdrawal.

The existing RBF withdrawal test succeeds since we initiated only a single deposit, which causes the coin selection algorithm to consider the same UTXO. The tests have been upgraded to consider multiple UTXOs.

### DIFF
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -1423,7 +1423,7 @@ impl<'a> StatelessWallet<'a> {
         // Ensure deterministic ordering of UTXOs for all peers
         included_utxos.sort_by_key(|(_, utxo)| utxo.amount);
         remaining_utxos.sort_by_key(|(_, utxo)| utxo.amount);
-        included_utxos.extend(remaining_utxos);
+        remaining_utxos.extend(included_utxos);
 
         // Finally we initialize our accumulator for selected input amounts
         let mut total_selected_value = bitcoin::Amount::from_sat(0);
@@ -1431,7 +1431,7 @@ impl<'a> StatelessWallet<'a> {
         let mut fees = fee_rate.calculate_fee(total_weight);
 
         while total_selected_value < peg_out_amount + change_script.dust_value() + fees {
-            match included_utxos.pop() {
+            match remaining_utxos.pop() {
                 Some((utxo_key, utxo)) => {
                     total_selected_value += utxo.amount;
                     total_weight += max_input_weight;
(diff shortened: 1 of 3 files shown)


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1423,15 +1423,15 @@ impl<'a> StatelessWallet<'a> {
         // Ensure deterministic ordering of UTXOs for all peers
         included_utxos.sort_by_key(|(_, utxo)| utxo.amount);
         remaining_utxos.sort_by_key(|(_, utxo)| utxo.amount);
-        included_utxos.extend(remaining_utxos);
+        remaining_utxos.extend(included_utxos);
```
bradleystachurski, line 1426: Subtle bug :smile:

elsirion: Uff, easy to miss indeed :(

elsirion:
I think we agreed yesterday to remove RBF withdraw functionality, right (details tbd)? So let's not merge but it's interesting to look at to understand the bug.

dpc:
> perhaps using meta?

Reminder: meta is only for ... metadata about the Federation for external consumption. It was not designed as an input controlling consensus itself, which would be sort of a circular dependency and requires a bit more thought than 1 evening of hacking just because I was annoyed at clumsiness of editing external meta files.

joschisan:
> I think we agreed yesterday to remove RBF withdraw functionality, right (details tbd)? So let's not merge but it's interesting to look at to understand the bug.

@bradleystachurski and I talked about it today and he will open a pr to remove rbf. If I understand correctly this pr would then not be merged.

bradleystachurski:
> @bradleystachurski and I talked about it today and he will open a pr to remove rbf. If I understand correctly this pr would then not be merged.

Yep :+1:

I'll close this and open a separate PR that removes RBF withdrawals once it's ready.

PR #5471, opened 2024-06-17
chore: convert fee configuration tests to devimint tests

m1sterc001guy:
Part of https://github.com/fedimint/fedimint/issues/5168

Makes `test_can_change_default_routing_fees` and `test_can_change_federation_routing_fees` apart of the devimint tests.

### DIFF
--- a/gateway/integration_tests/src/main.rs
+++ b/gateway/integration_tests/src/main.rs
@@ -4,7 +4,7 @@ use clap::{Parser, Subcommand};
 use devimint::version_constants::VERSION_0_3_0;
 use devimint::{cmd, util};
 use fedimint_testing::gateway::LightningNodeType;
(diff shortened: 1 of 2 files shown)


elsirion:
@dpc I remember you adding support for module-specific devimint tests and it feels like this would be a good candidate for one (so we don't just put all our tests into this one file). @m1sterc001guy did you hit any problems preventing that?

m1sterc001guy:
If I understand correctly, this is using @dpc's module-specific devimint test framework, its just for the gateway module.

Some of the functionality might be able to be split into separate tests, but I wanted to amortize the cost of running a test, since setting up a federation takes some time.

elsirion (approved):
You are right, I got confused about the folder structure.

PR #5473, opened 2024-06-17
 refactor: whole client peg-in handling 

dpc:
Note: I didn't like this code when I was writing it. :D

Summary:

Instead of creating a state machine looking for a single peg-in tx each time user requests a new deposit address, run a single background task at all times that scans for deposits on (potentially) all derived deposit addresses and claim them as they become ready.

This allows gracefully handling rbf, and claiming funds even when user re-used peg-in address.

To prevent privacy degradation via querying all the new and past deposit addresses all the time, the loop does some accounting and custom logic to figure out which addresses should be checked and where, with optional ability to force checking (to be triggered by the user manually).

 
Known follow-ups needed:

* "recovery" - basically re-deriving pegin addresses a bit after last detected address used-on chain
* API to re-check an existing pegin addresss
* Better visibility into pending pegins.


fedimint-client/src/module/mod.rs
```diff
@@ -535,6 +587,13 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
 
     fn context(&self) -> Self::ModuleStateMachineContext;
 
+    /// Initialize client.
+    ///
+    /// Called by the core client code on start, after [`ClientContext`] is
+    /// fully initialized, so unlike during [`ClientModuleInit::init`],
+    /// access to global client is allowed.
+    async fn start(&self) {}
```
dpc, line 595: Starting monitor from `init` lead to race conditions.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -9,18 +9,23 @@
 pub mod api;
 
 pub mod client_db;
+/// Legacy, state-machine based peg-ins, replaced by `pegin_monitor`
+/// but retained for time being to ensure existing peg-ins complete.
 mod deposit;
```
dpc, line 14: At some point in the future, we should be able to clean up and remove, but for now, it can just do what it was doing before.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -162,24 +183,36 @@ impl ClientModuleInit for WalletClientInit {
...
+            cfg: args.cfg().clone(),
+            module_root_secret: args.module_root_secret().clone(),
+        };
+
         let rpc_config = self
             .0
             .clone()
             .unwrap_or(WalletClientModule::get_rpc_config(args.cfg()));
 
-        // FIXME: reactivate key derivation once we implement recovery
```
dpc, line 170: While the recovery is not implemented yet, re-creating same deposit addresses should work perfectly fine now, and we will just detect new deposits.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -401,45 +501,45 @@ impl WalletClientModule {
         })
     }
 
-    pub async fn get_deposit_address<M: Serialize + MaybeSend + MaybeSync>(
```
dpc, line 404: `get` just didn't seem right.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -453,89 +553,95 @@ impl WalletClientModule {
...
+                .await
+                .expect("Must have a corresponding claim record");
+            res.push((
+                outpoint,
+                claimed_peg_in_data.claim_txid,
+                claimed_peg_in_data.change,
+            ));
         }
 
-        let mut operation_stream = self.notifier.subscribe(operation_id).await;
```
dpc, line 482: While internally, only tests were consuming the intermediate states, currently nothing in the code is attempting to re-create the same level of visibility into intermediate states. It can be done as a follow-up.

benthecarman (commented):
Scanned through this, my understanding is that this flips the wallet module's relationship on its head, where now the client is responsible for notifying the mint when a deposit has come in. This sounds great because it fixes the address reuse problem and puts less strain on the guardians.

The one drawback is the client now has a requirement of needing bitcoind or esplora.

concept ack from me, seems great

is removing rbf going to be in a separate change?

dpc:
> Scanned through this, my understanding is that this flips the wallet module's relationship on its head, where now the client is responsible for notifying the mint when a deposit has come in. This sounds great because it fixes the address reuse problem and puts less strain on the guardians.

Correction:

I don't think the relationship changes much. The client side app/user just requests new deposit addresses, and the client mint module will keep observing them in a loop and claiming funds. Unlike the previous code there is no single state machine looking for a single deposit anymore. There's just a one big background task watching all addresesses all the time. This allows it to detect multiple deposits, not have any problems with rbf etc.

> The one drawback is the client now has a requirement of needing bitcoind or esplora.
> 
> concept ack from me, seems great

Client side already requires bitcoind/explora/etc. The client needed to obtain tx proof from somewhere already.


> is removing rbf going to be in a separate change?

Not sure what you mean. If you mean rbf on peg-outs, then it's a completely different work and @bradleystachurski is tracking it. This PR is only about peg-ins, and here rbf txes should work perfectly fine after this lands.

@benthecarman

dpc:
@benthecarman I have wrote a summary of PR. Should have done it before.

modules/fedimint-wallet-client/src/pegin_monitor.rs
```diff
@@ -0,0 +1,488 @@
...
+    data: &WalletClientModuleData,
+    tweak_idx: TweakIdx,
+    btc_rpc: &DynBitcoindRpc,
+    module_rpc: &DynModuleApi,
+    db: &Database,
+    client_ctx: &ClientContext<WalletClientModule>,
+) -> Result<Vec<CheckOutcome>, anyhow::Error> {
+    let current_consensus_block_count = module_rpc.fetch_consensus_block_count().await?;
+    let (script, address, tweak_key, operation_id) = data.derive_peg_in_script(tweak_idx);
+    btc_rpc.watch_script_history(&script).await?;
```
bradleystachurski, line 315: Probably fine to keep as-is since the client likely won't use bitcoind anytime soon, but we call `watch_script_history` when we allocate a deposit address, so I don't believe this is necessary and I'm also curious about the original intent behind this comment.

https://github.com/fedimint/fedimint/blob/449f030e78455b69e7422ac0fb71f46d4e4614c8/fedimint-bitcoind/src/lib.rs#L172-L173

modules/fedimint-wallet-client/src/pegin_monitor.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    let history = btc_rpc.get_script_history(&script).await?;
```
bradleystachurski, line 317: Not a blocker for this PR, but we may want to consider updating the esplora impl for paging since this will only return 25 items.

modules/fedimint-wallet-client/src/pegin_monitor.rs
```diff
@@ -0,0 +1,488 @@
...
+            });
+            debug!(target: LOG_CLIENT_MODULE_WALLET, %txid, %out_idx, "In the mempool");
+            continue;
+        };
+
+        // if a tx was confirmed in block N and N+1 is a current block count, it has 1
+        // confirmation, and 0 extra confirmations
+        let num_blocks_needed = tx_block_height
+            .saturating_add(1)
+            .saturating_sub(current_consensus_block_count);
```
bradleystachurski, line 360: [nit]

No need to change, but the difference between block height and block count might be easier to understand using an intermediate variable `tx_block_count`, e.g.

```rust
let tx_block_count = match btc_rpc.get_tx_block_height(&txid).await? {
    Some(tx_block_height) => tx_block_height.saturating_add(1),
    None => {
        outcomes.push(CheckOutcome::Pending {
            num_blocks_needed: finality_delay,
        });
        debug!(target: LOG_CLIENT_MODULE_WALLET, %txid, %out_idx, "In the mempool");
        continue;
    }
};
let num_blocks_needed = tx_block_count.saturating_sub(current_consensus_block_count);
```

modules/fedimint-wallet-client/src/pegin_monitor.rs
```diff
@@ -0,0 +1,488 @@
...
+};
+use crate::{WalletClientModule, WalletClientModuleData, WalletClientStates};
+
+/// A helper struct meant to combined data from all addresses/records
+/// into a single struct with all actionable data.
+#[derive(Debug, Clone)]
+struct NextActions {
+    /// Current time
+    now: SystemTime,
+    /// Index keys due to for a check
```
bradleystachurski, line 37: [nit] rm "to"

modules/fedimint-wallet-client/src/client_db.rs
```diff
@@ -14,11 +20,82 @@ impl std::fmt::Display for DbKeyPrefix {
     }
 }
 
+/// And index of a deposit address
```
bradleystachurski, line 23: [nit] s/And/An

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -272,60 +396,36 @@ impl WalletClientModule {
...
+        self.cfg().fee_consensus
     }
 
     pub async fn get_deposit_address_inner(
         &self,
-        valid_until: SystemTime,
         dbtx: &mut DatabaseTransaction<'_>,
-    ) -> (OperationId, WalletClientStates, Address) {
+    ) -> (OperationId, Address, TweakIdx) {
         let deposit_idx = get_next_peg_in_tweak_child_id(dbtx).await;
```
bradleystachurski, line 410: [nit] renaming `deposit_idx` to `tweak_idx` would make this consistent with other usages

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -401,45 +501,45 @@ impl WalletClientModule {
...
-                            .await?;
-
-                        dbtx.add_operation_log_entry(
-                            operation_id,
-                            WalletCommonInit::KIND.as_str(),
-                            WalletOperationMeta {
-                                variant: WalletOperationMetaVariant::Deposit {
-                                    address: checked_address_to_unchecked_address(&address),
-                                    expires_at: valid_until,
+                        dbtx.module_dbtx()
```
bradleystachurski, line 525: Why insert again here if we just inserted the new entry in `get_deposit_address_inner`? I think this is fine as-is, just curious if I'm missing something.

bradleystachurski (commented):
I'm still reviewing but want to provide some initial feedback. Thus far everything looks good and I don't see any issues with this approach :+1:

bradleystachurski:
@benthecarman 

> where now the client is responsible for notifying the mint when a deposit has come in

Thanks for bringing up this point! To expand on @dpc's comment/description:

One of the neat designs of fedimint is the flow for notifying the mint that a deposit is creditable. For centralized exchanges, this is typically done by creating a unique account id for a user then generating a new deposit address that maps to that account id. If the exchange observes any transactions paying to that address, it will credit the user. This design has significant privacy implications, since it requires introducing accounts.

The deposit flow in fedimint places the responsibility on the client to prove that a deposit is spendable by the mint. The client achieves this by generating a tweak applied to the mint's peg-in descriptor and sends funds to the tweaked address. Once the client observes enough confirmations, it will prove to the mint that the coins are spendable with the tweak and the merkle proof that the tx was included in a block. If the proof is valid, the mint will issue ecash for the deposit. With this flow, the mint won't be able to correlate user deposit activity if a new deposit address is used.

bradleystachurski:
Capturing a chat @dpc and I had out-of-band: If there are users with outstanding deposits that were RBF'ed and still pending, this approach won't recognize those deposits. We'll still need to individually help those users.

modules/fedimint-wallet-client/src/pegin_monitor.rs
dpc: > , so I don't believe this is necessary and I'm also curious about the original intent behind this comment.

What if the underlying bitcoind changed or something. I don't think there's a reason to not call it - I doubt the performance matters, and if the bitcoind already imported it, it will just immediately responde with "OK".

Maybe we should get rid of the whole call altogether.

modules/fedimint-wallet-client/src/pegin_monitor.rs
dpc: Yeah, we probably want to read all pages inside the impl of that function (with some very high limit maybe, to prevent abuse). We never call these with high performance requirements, and I doubt we'll want to handle pagination in the higher level loop.

modules/fedimint-wallet-client/src/lib.rs
dpc: Cuz I'm a distracted baboon. :disappointed_relieved:

modules/fedimint-wallet-client/src/pegin_monitor.rs
dpc: I like it.

modules/fedimint-wallet-client/src/pegin_monitor.rs
```diff
@@ -0,0 +1,488 @@
...
+    due_val: PegInTweakIndexData,
+    pengin_claimed_sender: &watch::Sender<()>,
+) -> Result<(), anyhow::Error> {
+    let now = time::now();
+    match check_idx_pegins(data, due_key.0, btc_rpc, module_api, db, client_ctx).await {
+        Ok(outcomes) => {
+            let next_check_time = CheckOutcome::retry_delay_vec(&outcomes, due_val.creation_time)
+                .map(|duration| now + duration);
+            client_ctx
+                        .module_autocommit_2(
```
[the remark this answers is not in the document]
dpc: rustfmt sometimes gives up without indication why. I fixed it manually a bit, but I kind of don't care, as long as it's not entirely out of whack.

modules/fedimint-wallet-client/src/pegin_monitor.rs
```diff
@@ -0,0 +1,488 @@
...
+        }
+
+        let now = time::now();
+        let next_wakeup = NextActions::from_db_state(&db).await.next.unwrap_or_else(||
+            /* for simplicity just wake up every hour, even when there's no need */
+             ( now + Duration::from_secs(60 * 60)));
+        let next_wakeup_duration = next_wakeup
+            .duration_since(now)
+            .unwrap_or_default()
+            .min(min_sleep);
```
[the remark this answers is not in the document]
dpc: Great save.

Kodylow:
Is there any performance issue here if the set off addresses we're watching explodes? I think most bitcoin impls handle this with gap limits around what they're actively looking for then adding a watch to an old addr as needed?

dpc:
I have verified that even without recovery code re-deriving addresses on a recovered client works OK.

modules/fedimint-wallet-client/src/client_db.rs
```diff
@@ -14,11 +20,82 @@ impl std::fmt::Display for DbKeyPrefix {
...
+}
+
+/// A counter tracking next index to use to derive a peg-in address
 #[derive(Clone, Debug, Encodable, Decodable, Serialize)]
 pub struct NextPegInTweakIndexKey;
 
 impl_db_record!(
     key = NextPegInTweakIndexKey,
-    value = u64,
+    value = TweakIdx,
```
bradleystachurski, line 49: Does this require a client db migration?

bradleystachurski (approved):
If introducing a client db migration isn't required for the `TweakIdx` wrapper, this looks good. I'm adding the "needs further review" tag so others can have a look, but I think feedback can be addressed in followups.

Capturing our chat: A test to verify rbf deposits would be useful but can be addressed in a followup.

modules/fedimint-wallet-client/src/client_db.rs
Kodylow: i believe "Encodable, Decodable" always means yes?

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -453,89 +539,176 @@ impl WalletClientModule {
...
+    }
 
-        if !matches!(
-            operation_meta.variant,
-            WalletOperationMetaVariant::Deposit { .. }
-        ) {
-            bail!("Operation is not a deposit operation");
-        }
+    /// Like [`Self::recheck_pegin_address`] but by `operation_id`
+    pub async fn recheck_pegin_address_by_op_id(
```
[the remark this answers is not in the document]
dpc: I'm a bit conflited on the "what should we key the peg-in checks on".

Before we *had* to use operation_id because checking was a state_machine. So to keep the CLI APIs as they were I needed functions that work on operatin_id. Now it's just an index of derived address (`TweakIdx`) and it seems easier. But we could also look up things by `address` to avoid users need to think about any indexes...

Anyway, I think it's OK to keep, as it's public and can be used by external apps. I don't think it hurts much.

bradleystachurski (approved):

modules/fedimint-wallet-client/src/lib.rs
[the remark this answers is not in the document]
Kodylow: If you key on address won't that fail reuse? And if someone does reuse a pegin address can't anyone watching just publish the inclusion proof + the info from the previous pegin to that address in order to receive the ecash?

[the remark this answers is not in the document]
dpc: There's one address for one operation_id with one tweak_idx. address and operation_id are both derived from the tweak_idx. Each of these can have multiple deposits though.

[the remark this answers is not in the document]
bradleystachurski: > And if someone does reuse a pegin address can't anyone watching just publish the inclusion proof + the info from the previous pegin to that address in order to receive the ecash?

I believe that's correct, if you know the tweak key and the deposit has enough confs you can construct a valid wallet input and claim the ecash.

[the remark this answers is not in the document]
bradleystachurski: Scratch that, it's not possible. An observer will know the public key used to tweak the peg-in descriptor, however they would also need to know the secret, which is used to sign the fedimint tx that claims the input.

https://github.com/fedimint/fedimint/blob/ef6d4b12241651c2faa842547cbc368ceabbe4e5/fedimint-client/src/transaction/builder.rs#L123

PR #5475, opened 2024-06-18
fedimint client rpc

maan2003:
hopefully we will get a fedimint typescript bindings out of this

### DIFF
--- a/fedimint-client/src/module/mod.rs
+++ b/fedimint-client/src/module/mod.rs
@@ -603,6 +603,16 @@ pub trait ClientModule: Debug + MaybeSend + MaybeSync + 'static {
         ))
     }
 
+    async fn handle_rpc(
+        &self,
+        _method: String,
+        _request: serde_json::Value,
+    ) -> BoxStream<'_, anyhow::Result<serde_json::Value>> {
+        Box::pin(futures::stream::once(std::future::ready(Err(
+            anyhow::format_err!("This module does not implement rpc"),
+        ))))
+    }
+
     /// Returns the fee the processing of this input requires.
     ///
     /// If the semantics of a given input aren't known this function returns
@@ -772,6 +782,12 @@ pub trait IClientModule: Debug {
     async fn handle_cli_command(&self, args: &[ffi::OsString])
         -> anyhow::Result<serde_json::Value>;
 
+    async fn handle_rpc(
+        &self,
+        method: String,
+        request: serde_json::Value,
+    ) -> BoxStream<'_, anyhow::Result<serde_json::Value>>;
+
     fn input_fee(&self, input: &DynInput) -> Option<Amount>;
 
     fn output_fee(&self, output: &DynOutput) -> Option<Amount>;
@@ -833,6 +849,14 @@ where
         args: &[ffi::OsString],
     ) -> anyhow::Result<serde_json::Value> {
         <T as ClientModule>::handle_cli_command(self, args).await
+    }
+
+    async fn handle_rpc(
+        &self,
+        method: String,
+        request: serde_json::Value,
+    ) -> BoxStream<'_, anyhow::Result<serde_json::Value>> {
+        <T as ClientModule>::handle_rpc(self, method, request).await
     }
 
     fn input_fee(&self, input: &DynInput) -> Option<Amount> {
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -31,7 +31,7 @@ use std::time::Duration;
 
 use anyhow::{anyhow, bail, ensure, format_err, Context};
 use api::LnFederationApi;
-use async_stream::stream;
+use async_stream::{stream, try_stream};
 use bitcoin::hashes::{sha256, Hash, HashEngine, Hmac, HmacEngine};
 use bitcoin::key::KeyPair;
 use bitcoin::Network;
@@ -58,7 +58,7 @@ use fedimint_core::module::{
 };
 use fedimint_core::task::{timeout, MaybeSend, MaybeSync};
 use fedimint_core::util::update_merge::UpdateMerge;
-use fedimint_core::util::{backon, retry};
+use fedimint_core::util::{backon, retry, BoxStream};
 use fedimint_core::{
     apply, async_trait_maybe_send, push_db_pair_items, runtime, Amount, OutPoint, TransactionId,
 };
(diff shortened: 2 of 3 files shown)


maan2003:
@dpc thoughts, this is similar to cli commands in modules

fedimint-client-wasm/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+anyhow.workspace = true
+fedimint-core.workspace = true
+fedimint-ln-client = { workspace = true, features = ["cli"] }
+fedimint-mint-client = { version = "=0.4.0-alpha", path = "../modules/fedimint-mint-client" }
+fedimint-client = { workspace = true }
+wasm-bindgen = "=0.2.92"                                                                      # must match the nix provided wasm-bindgen-cli version
```
dpc, line 15: That comment drifted...

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -2388,3 +2389,29 @@ mod tests {
         );
     }
 }
+
```
dpc, line 2392: Would move it  to some `rpc.rs` module from the get go.

fedimint-client/src/client_rpc.rs
```diff
@@ -0,0 +1,159 @@
...
+        } else {
+            Box::pin(futures::stream::once(async {
+                Err(anyhow::anyhow!("Handler not found"))
+            }))
+        }
+    }
+}
+
+#[apply(async_trait_maybe_send!)]
+pub trait ClientRpcHandler<Ctx> {
```
dpc, line 60: Do we need another trait? Couldn't this be a method on the existing `ClientModule`, just like cli is on it already?

fedimint-client/src/client_rpc.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::{collections::BTreeMap, pin::Pin};
+
+use anyhow::Context;
+use fedimint_core::task::{MaybeSend, MaybeSync};
+use fedimint_core::{apply, async_trait_maybe_send};
+use futures::{Stream, StreamExt as _};
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+struct RpcHandlerKey<'a> {
+    // Q: should we use instance id instead of module kind?
```
dpc, line 11: Yes

dpc (commented):
I would just add `fn handle_rpc(req: serde_json::Value) -> Result<serde_json::Value>` on `ClientModule` right next to existing `handle_cli_command`, and start from there.


Then request and response could become `struct Request` and `struct Response` with methods that helps casting them to the desired type for convenience.

Seems just simpler than all these extra traits. Also it could re-use `self` and any context on it. After all cli and rpc are very much alike.

So an rpc would just literally open `Client` and pass requests to `client.module.handle_rpc` and convert the response back to the caller.

We could even have `fedimint-cli start-rpc-server` or something, but of course one could wrap it in a http, jsonrpc or whatever else they desire.

maan2003:
I wanted more typed api and more fine grain control to integrator of rpc, eventually I want to use this rpc in fedi app.

the typed api means we can generates typescript definitions automatically.

@dpc

dpc:
> I wanted more typed api and more fine grain control to integrator of rpc, eventually I want to use this rpc in fedi app.
> 
> the typed api means we can generates typescript definitions automatically.
> 
> @dpc

You can still have a types with `#[derive(Deserialize)]`, make them public in some `xyz-module-rpc` and use them to generate bindings, and deserialize the body of the request using them so you know it stays correct, etc. The reason we use `Dyn*` magic in other places is so that we can have bunch of other methods on them, have a consensus binary encoding etc. With RPC, we know we are basically going to be sending json back and forth - `serde_json::Value` seems just right for the task.

Rpc and cli are fundamentally very much alike, so why would rpc needs all this extra machinery? In particular rpc not being must a method on the module instance (like cli handler is) seems to introduce a lot of extra complexity.

@maan2003

maan2003:
updated @dpc

fedimint-client-wasm/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+        let invite_code = InviteCode::from_str(&invite_code)?;
+        let config = fedimint_api_client::download_from_invite_code(&invite_code).await?;
+        let client = Arc::new(builder.join(root_secret, config, None).await?);
+        Ok(Self { client })
+    }
+
+    #[wasm_bindgen]
+    pub async fn rpc(&self, module: &str, body: String, cb: &js_sys::Function) {
+        let request = serde_json::from_str(&body).unwrap();
+        let mut stream = match module {
```
dpc, line 52: I'm very confused about everything in this file, but I trust you know what you're doing. :D

fedimint-client/src/lib.rs
```diff
@@ -722,7 +722,7 @@ pub struct ClientModuleInstance<'m, M: ClientModule> {
     /// Module-specific API
     pub api: DynModuleApi,
 
-    module: &'m M,
+    pub module: &'m M,
```
dpc, line 725: This type defers to `M` (right below), so I don't think we need this to be public.

fedimint-client/src/module/mod.rs
```diff
@@ -772,6 +781,11 @@ pub trait IClientModule: Debug {
     async fn handle_cli_command(&self, args: &[ffi::OsString])
         -> anyhow::Result<serde_json::Value>;
 
+    async fn handle_rpc(
+        &self,
+        request: serde_json::Value,
+    ) -> BoxStream<'_, anyhow::Result<serde_json::Value>>;
```
dpc, line 787: Why stream? Do we *reaaaly* need support for streaming answers?

Honestly I'm sour on the subscription etc. and here it's not even going through a net layer, so polling (possibly with timeouts for good latency) is practically overhead-free.

fedimint-client/src/module/mod.rs
```diff
@@ -772,6 +781,11 @@ pub trait IClientModule: Debug {
     async fn handle_cli_command(&self, args: &[ffi::OsString])
         -> anyhow::Result<serde_json::Value>;
 
+    async fn handle_rpc(
+        &self,
```
dpc, line 785: Where's the rpc name? Seems to me that name + payload is a more pragmatic approach, no?

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -433,6 +433,85 @@ impl ClientModule for LightningClientModule {
...
+                    let outgoing_payment = self.pay_bolt11_invoice(
+                        maybe_gateway,
+                        invoice,
+                        extra_meta,
+                    ).await?;
+                    yield serde_json::to_value(outgoing_payment)?;
+                },
+                LnClientRpcRequest::SubscribeLnPay { operation_id } => {
+                    for await state in self.subscribe_ln_pay(operation_id).await?.into_stream() {
+                        yield serde_json::to_value(state)?;
```
dpc, line 481: I see. Meh.

I regret any subscriptions in Fedimint code, including jsonrpc. E.g. this type of functionality can be handle better with a `GetNextLNPayEvent { cursor: Option<SomeEventId>, timeout: Option<Duration> }` kind of RPC.

What if someone wants to use these rpcs over http. Like what @Kodylow already has in `fedimint-http` or something. we should have `fedimint-http`-like thing, that is just a wrapper over these new rpcs. But now if we make things a stream here, what do we do? We either have to invent cursor-based schims to make it possible, or force jsonrpc and complexity on all downstream.

The opposite is actually great. If the inner-logic uses cursor-based API, then wrapping it around to get a stream is trivial.

Similar with all the client event logs operations, etc. Had they had cursor-based APIs, they would be easier to follow accross restarts, from external system. And also the complexity of handling them wouldn't suck you out of everything that touches them. :D

dpc:
For the record. We had a call with @maan2003 about this one. He punched me over the camera for being annoying. Just kidding, just kidding! We're going to make some tweaks, but nothing big and we seem to agree on everything.

fedimint-client/src/lib.rs
maan2003: defer doesn't have same lifetime :/

if I remove .module in wasm-client/lib.rs, it fails due to lifetime error. 
I can make it a getter.

maan2003:
Now this PR is much simpler, only containing rpc changes. 
wasm will be a separate PR.

dpc (approved):

PR #5477, opened 2024-06-18
chore: fibonacci backoff when connecting to guardians

### LINKED ISSUE #5460: Exponential back-off when connecting to guardians
I found this is where the back off happens, it seems it is linear with a max wait time of 5 seconds. Seems this can end up being pretty aggressive. In mutiny I can find if one of the guardians is down in a federation, my logs are filled with trying to connect to that guardian

https://github.com/fedimint/fedimint/blob/99b7fa570929429fd3f32dac5109092c109fdc72/fedimint-api-client/src/api.rs#L1139

tvolk131:
Fixes #5460

### DIFF
--- a/devimint/src/gatewayd.rs
+++ b/devimint/src/gatewayd.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 use std::time::Duration;
 
 use anyhow::{Context, Result};
-use fedimint_core::util::retry;
+use fedimint_core::util::{backon, retry};
 use ln_gateway::lightning::ChannelInfo;
 use ln_gateway::rpc::V1_API_ENDPOINT;
 use tracing::info;
@@ -121,7 +121,7 @@ impl Gatewayd {
     pub async fn get_info(&self) -> Result<serde_json::Value> {
         retry(
             "Getting {} gateway info via gateway-cli info",
-            fedimint_core::util::FibonacciBackoff::default()
+            backon::FibonacciBuilder::default()
                 .with_min_delay(Duration::from_millis(200))
                 .with_max_delay(Duration::from_secs(5))
                 .with_max_times(10),
--- a/fedimint-api-client/src/api.rs
+++ b/fedimint-api-client/src/api.rs
@@ -1174,7 +1174,7 @@ impl FederationPeerClientConnectionState {
 #[derive(Debug)]
 struct FederationPeerClient<C> {
     client: JitTryAnyhow<C>,
-    shared: Arc<tokio::sync::Mutex<FederationPeerClientConnectionState>>,
+    connection_state: Arc<tokio::sync::Mutex<FederationPeerClientConnectionState>>,
 }
 
 impl<C> FederationPeerClient<C>
@@ -1182,12 +1182,13 @@ where
     C: JsonRpcClient + 'static,
 {
     pub fn new(peer_id: PeerId, url: SafeUrl, api_secret: Option<String>) -> Self {
-        let shared: Arc<_> =
-            tokio::sync::Mutex::new(FederationPeerClientConnectionState::new()).into();
+        let connection_state = Arc::new(tokio::sync::Mutex::new(
+            FederationPeerClientConnectionState::new(),
+        ));
 
         Self {
-            client: Self::new_jit_client(peer_id, url, api_secret, shared.clone()),
-            shared,
+            client: Self::new_jit_client(peer_id, url, api_secret, connection_state.clone()),
+            connection_state,
         }
     }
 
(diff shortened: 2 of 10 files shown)


fedimint-api-client/src/api.rs
```diff
@@ -1117,29 +1118,35 @@ pub struct WsFederationApi<C = WsClient> {
...
+/// Connection state shared/preserved between [`FederationPeerClient`] and the
 /// Jit tasks it spawns.
 #[derive(Debug)]
-struct FederationPeerClientShared {
-    last_connection_attempt: SystemTime,
-    connection_attempts: u64,
+struct FederationPeerClientConnectionState {
+    /// Last time a connection attempt was made, or `None` if no attempt has
+    /// been made yet.
+    last_connection_attempt_or: Option<SystemTime>,
```
dpc, line 1127: Any benefit of not initializing with current time or something just to avoid handling `None` in other places?

fedimint-api-client/src/api.rs
```diff
@@ -1117,29 +1118,35 @@ pub struct WsFederationApi<C = WsClient> {
...
+    connection_backoff: backon::FibonacciBackoff,
 }
 
-impl FederationPeerClientShared {
+impl FederationPeerClientConnectionState {
+    const MIN_BACKOFF: Duration = Duration::from_millis(100);
+    const MAX_BACKOFF: Duration = Duration::from_secs(5);
```
dpc, line 1133: In #5460 :

> t seems it is linear with a max wait time of 5 seconds. 

Was the desire only to make it exponential or to increase the max as well? @benthecarman thoughts?

fedimint-api-client/src/api.rs
```diff
@@ -1168,42 +1177,57 @@ impl FederationPeerClientShared {
...
+                    "Connecting to peer...");
+                } else {
+                    debug!(
+                    target: LOG_CLIENT_NET_API,
+                    peer_id = %peer_id,
+                    url = %url,
+                    "Retrying connecting to peer...");
+                }
+
+                cs.wait().await;
```
dpc, line 1222: Would move this part into a function, as the current scope is getting long.

fedimint-api-client/src/api.rs
```diff
@@ -1168,42 +1177,57 @@ impl FederationPeerClientShared {
...
-                url = %url,
-                "Connecting to peer");
             let res = C::connect(&url, api_secret).await;
 
             match &res {
                 Ok(_) => {
-                    shared.lock().await.reset();
-                    debug!(
+                    connection_state.lock().await.reset();
+                    info!(
```
dpc, line 1230: I don't think connection and disconnection are interesting enough to print out by default (info is our default).

devimint/src/gatewayd.rs
```diff
@@ -121,7 +121,7 @@ impl Gatewayd {
     pub async fn get_info(&self) -> Result<serde_json::Value> {
         retry(
             "Getting {} gateway info via gateway-cli info",
-            fedimint_core::util::FibonacciBackoff::default()
+            backon::FibonacciBuilder::default()
```
dpc, line 124: Any reason we have a default / any reason we don't use it here?

fedimint-api-client/src/api.rs
dpc: I guess better logging is kind of nice.

tvolk131: The `None` state symbolizes that we haven't yet attempted to connect. This code below is the only reason we need that state:

```
                if cs.last_connection_attempt_or.is_none() {
                    info!(
                    target: LOG_CLIENT_NET_API,
                    peer_id = %peer_id,
                    url = %url,
                    "Connecting to peer...");
                } else {
                    debug!(
                    target: LOG_CLIENT_NET_API,
                    peer_id = %peer_id,
                    url = %url,
                    "Retrying connecting to peer...");
                }
```

We could keep `last_connection_attempt` as a non-`Option<T>`, but then we'd need a boolean value to keep track of whether we've attempted a connection already. I think the way it is now is more intuitive but I can change it if you disagree.

devimint/src/gatewayd.rs
tvolk131: The `Default` trait appears to be the only way to construct any of the builders in the `backon` crate.

dpc: Oh, so we only re-export the whole thing. Nvmd.

fedimint-api-client/src/api.rs
tvolk131: Done.

devimint/src/gatewayd.rs
tvolk131: I did that in this PR. Currently in master we're exporting only the `backon` builder structs, but we're renaming them. In this PR I removed the renamed export and am simply exporting the entire crate.

fedimint-api-client/src/api.rs
tvolk131: Reverted to debug

dpc (approved):
Ship it. If higher max retry duration is needed, we can do it in another change.

PR #5478, opened 2024-06-18
fix: multiple route attempts and exclude failed routes

m1sterc001guy:
As @elsirion mentioned here https://github.com/fedimint/fedimint/pull/5449#pullrequestreview-2125650786, CLN's `pay_private` implementation is a bit naive currently, it only tries one route and fails if that first route does not succeed.

This PR increases the number of route attempts to 10, and maintains a vector of `excluded_nodes` which is passed into each `getroute` request. This should hopefully increase the robustness of CLN's `pay_private`

### DIFF
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -47,6 +47,10 @@ use tonic::Status;
 use tracing::{debug, error, info, warn};
 
 const MAX_HTLC_PROCESSING_DURATION: Duration = Duration::MAX;
+// Attempt to get 10 different payment routes before returning an error
+const MAX_ROUTE_ATTEMPTS: u32 = 10;
+// Use a `riskfactor` of 10, which is the default for lightning-pay
+const ROUTE_RISK_FACTOR: u64 = 10;
 
 #[derive(Parser)]
 #[command(version)]
@@ -241,6 +245,7 @@ impl ClnRpcService {
         &self,
         pruned_invoice: PrunedInvoice,
         riskfactor: u64,
+        excluded_nodes: Vec<String>,
     ) -> Result<Vec<SendpayRoute>, ClnExtensionError> {
         let response = self
             .rpc_client()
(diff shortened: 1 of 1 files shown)


elsirion (commented):
Why a max number instead of a timeout? You could create an async block with an infinite loop and wrap it in a timeout like we do in other places. Trying ten rotues can take a few seconds to a few minutes depending on the nodes on the routes, so it doesn't tell us much about the required time.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -559,59 +564,117 @@ impl GatewayLightning for ClnRpcService {
...
-                fee.msat()
-            )));
-        }
+            loop {
+                let route = self
+                    .get_route(
+                        pruned_invoice.clone(),
+                        ROUTE_RISK_FACTOR,
+                        excluded_nodes.clone(),
+                    )
```
Kodylow, line 581: will give us routes we can't use since it doesn't take max_delay

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -559,59 +564,117 @@ impl GatewayLightning for ClnRpcService {
...
+                        let response = PayInvoiceResponse { preimage };
+                        return Ok(tonic::Response::new(response));
+                    }
+                    Err(e) => {
+                        error!(
+                            ?route_attempt,
+                            ?payment_hash,
+                            ?e,
+                            "Pruned invoice payment attempt failure"
+                        );
```
Kodylow, line 644: does this error contain any useful info about the node that failed? if you're a single channel to an LSP you don't want to ban your own lsp, should only exclude the specific along the route that failed.

Kodylow (dismissed):

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: Fixed. Now only returns the erring node.

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: Should we retry a new route in this case? Currently we just fail

elsirion: This isn't a strong enough reason to block anything, let's see if it's a problem in practice, I doubt it.

PR #5481, opened 2024-06-19
chore: split api client code into multiple files

tvolk131:
`fedimint-api-client/src/api.rs` is getting to be pretty big (1600+ lines) but has some chunks that can be broken out into separate files.


m1sterc001guy (approved):

dpc (approved):

dpc:
@tvolk131 These long file names are kind of unnecessary, IMO.

PR #5484, opened 2024-06-19
chore: simplify dev fed startup

tvolk131:
Remove duplicate calls to `wait_for_chain_sync()` and stop polling the underlying command since the command itself already performs polling here:

https://github.com/fedimint/fedimint/blob/9e15104e126bb850e792d148aed14aa26d1dad6b/gateway/cli/src/main.rs#L338-L357

Also fix a race condition in `gateway-cli lightning wait-for-chain-sync` and raise the polling rate to get slightly faster & more consistent mprocs startup times

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -21,8 +21,8 @@ use ln_gateway::rpc::{
 };
 use serde::Serialize;
 
-const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRIES: u32 = 12;
-const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRY_DELAY_SECONDS: u64 = 10;
+const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRIES: u32 = 60;
+const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRY_DELAY_SECONDS: u64 = 2;
 
 #[derive(Parser)]
 #[command(version)]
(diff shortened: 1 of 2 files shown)


gateway/cli/src/main.rs
```diff
@@ -21,8 +21,8 @@ use ln_gateway::rpc::{
 };
 use serde::Serialize;
 
-const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRIES: u32 = 12;
-const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRY_DELAY_SECONDS: u64 = 10;
+const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRIES: u32 = 60;
```
m1sterc001guy, line 24: Might be helpful for `wait_for_chain_sync` to have a `debug!` log or something to tell if it ever gets stuck

m1sterc001guy (approved):

devimint/src/gatewayd.rs
```diff
@@ -241,21 +241,15 @@ impl Gatewayd {
...
-            )
-            .run()
-            .await
-            .map_err(ControlFlow::Continue)?;
-            Ok(())
-        })
-        .await?;
-        Ok(())
+        let block_height = bitcoind.get_block_count()? - 1;
+        cmd!(
```
dpc, line 245: What if the call fails?

gateway/cli/src/main.rs
dpc: `poll` already had it - another reason not to remove(?)

dpc (approved):

PR #5485, opened 2024-06-20
feat(tor): add initial mvp tor support connection to onion addresses

oleonardolima:
partially completes https://github.com/fedimint/fedimint/issues/4752
depends on #5443


oleonardolima:
I added a new commit adding onion address support on #5443 instead, closing this one.

PR #5486, opened 2024-06-20
Fix::> Print consensus/module versions on startup of service

### LINKED ISSUE #3300: On start print consensus versions
* fedimintd
* gatewayd
* fedimint-cli (with RUST_LOG=debug)

should print all the consensus versions they have on start.

PsychoPunkSage:
Closes #3300 

Implements methods to expose the Consensus Version and Prints it on starting the services.


fedimintd/src/bin/main.rs
```diff
@@ -3,6 +3,7 @@ use fedimintd::Fedimintd;
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    println!("fedimintd:bin:main.rs -");
```
dpc, line 6: No `println`, only `info!` and similar from `tracing`.

PsychoPunkSage: This one is for my reference only....
I will use `info!` macro to print the version....
Though I have one question.... I have implemented the required line to get version But those lines don't provide me a function that will provide me the `Module/Consensus Version`
I was thinking may be I have to do something like this [LINK](https://github.com/fedimint/fedimint/blob/master/modules/fedimint-dummy-server/src/lib.rs#L102)
cause this will provide me the way to access version... Correct me If I'm wrong

elsirion:
Please do not merge master into your branch, instead [rebase on master](https://git-scm.com/docs/git-rebase). There are a lot of unrelated changes in this PR, it's a huge mess. Closing for now.

PR #5487, opened 2024-06-20
chore: replace locks with atomics

tvolk131:
Part of #1576

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -36,6 +36,7 @@ use std::net::SocketAddr;
 use std::ops::ControlFlow;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::atomic::AtomicU64;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -333,7 +334,7 @@ pub struct Gateway {
 
     // Tracker for short channel ID assignments. When connecting a new federation,
     // this value is incremented and assigned to the federation as the `mint_channel_id`
-    max_used_scid: Arc<Mutex<u64>>,
+    max_used_scid: Arc<AtomicU64>,
 
     // The Gateway's API URL.
     pub versioned_api: SafeUrl,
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -1025,16 +1026,7 @@ impl Gateway {
...
-            // connected.
-            let mut max_used_scid = self.max_used_scid.lock().await;
-            let mint_channel_id =
-                max_used_scid
-                    .checked_add(1)
-                    .ok_or(GatewayError::GatewayConfigurationError(
-                        "Too many connected federations".to_string(),
-                    ))?;
-            *max_used_scid = mint_channel_id;
+            let mint_channel_id = self.get_next_channel_id()?;
```
m1sterc001guy, line 1029: I don't think can be changed to be an `AtomicU64`. We explored this in the past. The issue is that we want to increase the `max_used_scid` atomically, but only if the entire creation of the client succeeds. With `AtomicU64` the client creation can fail later, causing the `max_used_scid` to be increased when it should not have.

TLDR: This entire operation needs to be atomic, not just the increasing of the `max_used_scid`, which is why a lock is used.

Perhaps we should add a comment explaining this.

dpc: Possibly this variable should be renamed to `scid_allocation_lock` or something.

dpc: BTW. It is usually better to track `next` then `last/max`.

tvolk131: > we want to increase the `max_used_scid` atomically, but only if the entire creation of the client succeeds

Digging into this more, it looks like `Gateway.max_used_scid` is only ever locked in methods that take a mutable reference to `self` (`&mut self`) which enforces synchronicity, so we shouldn't need any lock here at all, right?

dpc: Is there any reason to think on it for too much? Atomics are very constrained and worth it only when it's both simple case and performance gain is considerable.

I would consider upgrading Mutex to RwLock if there's a lot of codepaths that don't actually mutate.

PR #5491, opened 2024-06-20
wip: rm rbf withdrawals

bradleystachurski:
wip


bradleystachurski:
This is an approach that attempts to remove RBF withdrawal logic. After discussions with @dpc and @joschisan, I've decided to take a two step approach: 1. prevent, 2. remove logic/destructive migrations.

https://github.com/fedimint/fedimint/pull/5496

Issue #5121, opened 2024-04-26
Refactor Bitcoin::Network to be in global config of federation


m1sterc001guy:
See https://github.com/fedimint/fedimint/pull/5108#issuecomment-2079892814

Currently the bitcoin network (regest, testnet, signet, mainnet) is a config option inside the relevant modules (i.e wallet). The issue is to move this configuration to be a global property of the federation so it is not module specific.

PR #5492, opened 2024-06-20
chore: Add bitcoin::Network to ServerConfigConsensus

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/5121

I ended up keeping `network` in the individual module configs, but just passing the global value into it. To me this seemed like the easiest choice, but I'm open to other designs if there's suggestions.

Let me know if there's any backwards compatibility concerns.

I would like to get this in for 0.4 since I believe this is a consensus change.

### DIFF
--- a/devimint/src/federation/config.rs
+++ b/devimint/src/federation/config.rs
@@ -32,7 +32,7 @@ pub fn attach_default_module_init_params(
                 local: LightningGenParamsLocal {
                     bitcoin_rpc: bitcoin_rpc.clone(),
                 },
-                consensus: LightningGenParamsConsensus { network },
+                consensus: LightningGenParamsConsensus,
             },
         )
         .attach_config_gen_params(
@@ -49,7 +49,6 @@ pub fn attach_default_module_init_params(
                     bitcoin_rpc: bitcoin_rpc.clone(),
                 },
                 consensus: WalletGenParamsConsensus {
-                    network,
                     // TODO this is not very elegant, but I'm planning to get rid of it in a next
                     // commit anyway
                     finality_delay,
@@ -64,7 +63,7 @@ pub fn attach_default_module_init_params(
                 local: fedimint_lnv2_common::config::LightningGenParamsLocal {
                     bitcoin_rpc: bitcoin_rpc.clone(),
                 },
-                consensus: fedimint_lnv2_common::config::LightningGenParamsConsensus { network },
+                consensus: fedimint_lnv2_common::config::LightningGenParamsConsensus,
             },
         );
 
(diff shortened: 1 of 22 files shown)


joschisan:
When you add a new parameter to the config you either have to set a default value or we need to do it manually when migrating to 0.4. I don't now what @elsirion has planned here - we could set the default to main net maybe?

m1sterc001guy:
Where would the default be set? Is the issue with getting it from the CLI/env var that they could be different for each guardian? I'm not backfilling anything here so, just trying to understand how this is different from before

dpc:
@m1sterc001guy This might be tricky, because we need to able to read the existing config that does not have it written in. Otherwise fedimintd 0.4 will just crash when starting on existing federations.

I think the way to go, is to use a custom serde default value function that would somehow read it from the only existing place where it is: wallet module config. Possibly with some half-hacky ability to override via env var.

If reading the existing one  turns out impossible or just to hard, we could consider forcing guardians to just specify it via env var during 0.4 upgrade. If they do it incorrectly... well... oops.

dpc:
@m1sterc001guy IMO, the network should not exist in config of modules anymore. Such a redundancy is just asking for trouble. It should get passed to server-modules `init` via `args`, where modules that need it can just read it on every initialization, just like any other global settings.

m1sterc001guy:
Going to refactor this a bit, converting to draft

m1sterc001guy:
Ignoring backwards compatibility for a second, I think this is the correct design.

In order to make this backwards compatible:
 - `GlobalClientConfig` needs to be migrated to include `bitcoin::Network`. I think `bitcoin::Network::Bitcoin` should be sufficient as a default. This will need to migration mechanism in https://github.com/fedimint/fedimint/pull/5535
 - `LightningConfigConsensus` needs to be migrated to remove the `bitcoin::Network` field.
 - `WalletConfigConsensus` needs to be migrated to remove the `bitcoin::Network` field
 - Backwards compatibility CLI tests need to be modified to take the old generation JSON parameters when the federation is < 0.4.0.

Alternatively, we could deprecate the network value in the module configs if we don't want to do a migration.

elsirion:
I'd like to still be able to generate client config from server config and client config has to be kept backwards-compatible (i.e. keep the network). So either we inject the network into the module server -> module client cfg fn or we have to keep it in the module server cfg. 

I don't think it's acceptable even just breaking testnet clients, most people won't know that this was a conscious trade-off and see it as buggyness.

My preference is keeping the redundant field in module configs and asserting it being the same as the global config on startup.

m1sterc001guy:
> My preference is keeping the redundant field in module configs and asserting it being the same as the global config on startup.

I think this makes things easier, appreciate the feedback! I agree and will implement this.

m1sterc001guy:
I am going to close this, I don't think it is a priority and it is a decent amount of work to do correctly.

PR #5493, opened 2024-06-20
chore: cleanup TaskGroupInner

tvolk131:
Part of #1576

`TaskGroupInner` contains a few locks, but its data and functionality are intermixed with `TaskGroup`.

In this PR we move `TaskGroupInner` into a separate file, isolating its functionality from `TaskGroup` and making its locking behavior easier to grok.

Then we replace `TaskGroupInner`'s `VecDeque` of task handles with an unbounded channel, which removes the need for any locking when adding a new task. We still need to lock when accessing the task handles, but this is already done in an `async` function so we can wrap the channel reader in an async mutex rather than a blocking one.

### DIFF
--- a/fedimint-core/src/task.rs
+++ b/fedimint-core/src/task.rs
@@ -1,10 +1,11 @@
 #![cfg_attr(target_family = "wasm", allow(dead_code))]
 
+mod inner;
+
 /// Just-in-time initialization
 pub mod jit;
 pub mod waiter;
 
-use std::collections::VecDeque;
 use std::future::Future;
 use std::pin::{pin, Pin};
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


dpc (commented):
Can you explain the locking difference you mention in the description? Kind of hard to make sure I didn't miss something.

tvolk131:
> Can you explain the locking difference you mention in the description? Kind of hard to make sure I didn't miss something.

Sure! The change in locking behavior is entirely in the second commit of this PR. Previously, we stored all join handles for sub-tasks of a `TaskGroupInner` in this field:

`join: std::sync::Mutex<VecDeque<(String, JoinHandle<()>)>>`

We only use this field in two places - we call [`VecDeque::push_back()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.push_back) in `TaskGroupInner::add_join_handle()` and [`VecDeque::pop_front()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.pop_front) in `TaskGroupInner::join_all()`. The documentation for `VecDeque` mentions:

> The “default” usage of this type as a queue is to use [push_back](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.push_back) to add to the queue, and [pop_front](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.pop_front) to remove from the queue.

So this field is being used as an unbounded queue. But as-is, we're required to lock both on inserting to the queue and popping from it. In the second commit of this PR, we're replacing this unbounded queue with a [`tokio::sync::mpsc::unbounded_channel`](https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.unbounded_channel.html). The channel's receiver still needs to block in order to pop items, but the sender is lock-free and thread-safe and subsequently so is `TaskGroupInner::add_join_handle()`.

dpc (approved):

PR #5496, opened 2024-06-21
fix(wallet): prevent rbf withdrawals

bradleystachurski:
Closes https://github.com/fedimint/fedimint/issues/5453

This approach prevents peers from processing RBF withdrawal outputs in `process_output`.

These changes are meant to be as minimal as possible to prevent RBF withdrawals, with the intent of removing all RBF withdrawal logic along with destructive client and server db migrations once this version reaches End of Life support. I'm not sure if we've reached consensus on how long we want to support versions (see: https://github.com/fedimint/fedimint/discussions/3950#discussioncomment-7862690), but the linked chart and several out-of-band discussions suggest the minimum supported version is two major versions in the past, i.e. if this lands in 0.4.0, we can safely remove the remaining RBF logic in 0.6.0. If this approach is accepted, I'll create an issue to track the removal for 0.6.0.

Taking a two step process (1. prevent, 2. remove logic/destructive migrations) is necessary to support any private federations that successfully processed an RBF withdrawal. Any federation that had a single UTXO and a user with an RBF withdrawal would have been able to successfully process the transaction, so if any peer needed to resync history and we removed all RBF withdrawal logic, the peer would be unable to sync to tip. I've scanned all public federations using `fedimint-observer` and none of them have RBF withdrawals, however it's still possible a private federation does. This PR introduces an environment variable `FM_UNSAFE_ENABLE_RBF_WITHDRAWAL` to support this edge case.

An alternative approach to fix the underlying bug was attempted in https://github.com/fedimint/fedimint/pull/5465, however after further discussion (see: [20240617 - Fedimint Dev Call](https://bitcointv.com/w/kfpAKiETfPTa69NVDLrKAb)) we decided the complexity of maintaining RBF withdrawals wasn't worth the risk, especially since we can set an aggressive confirmation target for estimating fees (I'll create a followup PR to reduce conf target from 10 to 1 if this lands https://github.com/fedimint/fedimint/issues/5498).

### DIFF
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -584,6 +584,10 @@ impl WalletClientModule {
     /// replace by fee (RBF).
     /// This can prevent transactions from getting stuck
     /// in the mempool
+    #[deprecated(
+        since = "0.4.0",
+        note = "RBF withdrawals are rejected by the federation"
+    )]
     pub async fn rbf_withdraw<M: Serialize + MaybeSync + MaybeSend>(
         &self,
         rbf: Rbf,
--- a/modules/fedimint-wallet-common/src/lib.rs
+++ b/modules/fedimint-wallet-common/src/lib.rs
@@ -33,7 +33,7 @@ pub mod tweakable;
 pub mod txoproof;
 
 pub const KIND: ModuleKind = ModuleKind::from_static_str("wallet");
-pub const MODULE_CONSENSUS_VERSION: ModuleConsensusVersion = ModuleConsensusVersion::new(2, 0);
+pub const MODULE_CONSENSUS_VERSION: ModuleConsensusVersion = ModuleConsensusVersion::new(2, 1);
 
 pub const CONFIRMATION_TARGET: u16 = 10;
 
@@ -317,6 +317,12 @@ pub enum WalletOutputError {
     #[error("The wallet output version is not supported by this federation")]
     UnknownOutputVariant(#[from] UnknownWalletOutputVariantError),
 }
+
+// For backwards-compatibility with old clients, we use an UnknownOutputVariant
+// error when a client attempts a deprecated RBF withdrawal.
+// see: https://github.com/fedimint/fedimint/issues/5453
+pub const DEPRECATED_RBF_ERROR: WalletOutputError =
+    WalletOutputError::UnknownOutputVariant(UnknownWalletOutputVariantError { variant: 1 });
 
 #[derive(Debug, Error)]
 pub enum ProcessPegOutSigError {
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -30,7 +30,7 @@ use common::config::WalletConfigConsensus;
 use common::{
     proprietary_tweak_key, PegOutFees, PegOutSignatureItem, ProcessPegOutSigError, SpendableUTXO,
     WalletCommonInit, WalletConsensusItem, WalletCreationError, WalletInput, WalletModuleTypes,
-    WalletOutput, WalletOutputOutcome, CONFIRMATION_TARGET,
+    WalletOutput, WalletOutputOutcome, CONFIRMATION_TARGET, DEPRECATED_RBF_ERROR,
 };
 use fedimint_bitcoind::{create_bitcoind, DynBitcoindRpc};
 use fedimint_core::config::{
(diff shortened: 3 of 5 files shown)


modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -584,6 +584,10 @@ impl WalletClientModule {
     /// replace by fee (RBF).
     /// This can prevent transactions from getting stuck
     /// in the mempool
+    #[deprecated(
```
bradleystachurski, line 587: Deprecating this method has the advantage of allowing us to test:
- Clients will receive an error when attempting to RBF an existing withdrawal, but the original withdrawal still succeeds
- Clients that need to enable the environment variable `FM_UNSAFE_ENABLE_RBF_WITHDRAWAL` are able to successfully process RBF withdrawals

modules/fedimint-wallet-common/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub mod tweakable;
 pub mod txoproof;
 
 pub const KIND: ModuleKind = ModuleKind::from_static_str("wallet");
-pub const MODULE_CONSENSUS_VERSION: ModuleConsensusVersion = ModuleConsensusVersion::new(2, 0);
+pub const MODULE_CONSENSUS_VERSION: ModuleConsensusVersion = ModuleConsensusVersion::new(3, 0);
```
bradleystachurski, line 36: This is a breaking consensus change for the wallet module.

This diff will change the behavior of `process_output`, which causes a consensus failure if peers aren't on the same version.

dpc: I don't think we need to increment it yet. We will when we actually delete the old code.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -550,6 +550,22 @@ impl ServerModule for Wallet {
     ) -> Result<TransactionItemAmount, WalletOutputError> {
         let output = output.ensure_v0_ref()?;
 
+        // In 0.4.0 we began preventing RBF withdrawals. Once we reach EoL support
```
dpc, line 553: This method (`process_output`) is called both when checking submitted txes before accepting them into consensus, and when consensus being processed. Because of it I proposed we have a new method `check_output`, with a default impl that forwards the call to `process_output`. In tx submission code we'll call `check_output` instead of `process_output`. All modules except wallet will not implement new method (thus use a default fowarding to `process_output`), for wallet we will put this new "check" in it.

modules/fedimint-wallet-common/src/lib.rs
dpc: BTW. I'd like to use this opportunity to implement consensus version changing core code to be able to fix the whole issue properly in the near future. It should be actually very simple and we have all the parts for it in already.

modules/fedimint-wallet-server/src/lib.rs
bradleystachurski: Please let me know if I'm misunderstanding, but we only call `process_output` once in `process_transaction_with_dbtx`, which means we'd need to refactor so we have something like `process_transaction_with_dbtx` and `verify_transaction_with_dbtx` when we're processing the tx in consensus and when we're verifying the tx in submit tx.

To keep this PR's history easier to follow, I opened a separate draft PR to see if this approach is what you had in mind.

https://github.com/fedimint/fedimint/pull/5520

I prefer the implementation as-is with the minimal check in the implementation of `process_output` in the wallet module, but I don't have a strong opinion if you'd prefer to add the additional verify methods.

modules/fedimint-wallet-common/src/lib.rs
bradleystachurski: https://github.com/fedimint/fedimint/blob/f98c0b18c50eaba3391d2128763e2f48dbcc5904/fedimint-core/src/module/version.rs#L86-L89

Based off the docs, I think we should increment. If peers on older versions try to process an RBF withdrawal output while peers on this version return an error, we have a breaking change to the consensus rules.

Please let me know if I'm misinterpreting the intent of module consensus versions.

joschisan (approved):

modules/fedimint-wallet-server/src/lib.rs
dpc: Note after a call: It didn't occur to me that it's possible to restart syncing from scratch with the env var. It's not a best UX, but will do, and we generally don't expect anyone to hit it, so :+1:

dpc (approved):

bradleystachurski:
Oh shoot, removing from the merge queue so I can change the consensus version real quick :slightly_smiling_face:

joschisan (approved):

modules/fedimint-wallet-common/src/lib.rs
bradleystachurski: During the code review call we discussed that bumping the major version is an anti-pattern and we should instead bump the minor. Bumping the major would indicate to old clients that they're unable to communicate with servers running on that latest.

dpc (approved):

dpc:
@bradleystachurski We land?

PR #5500, opened 2024-06-21
chore: test lnv2 self payment with one gateway

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-client/src/cli.rs
+++ b/modules/fedimint-lnv2-client/src/cli.rs
@@ -1,15 +1,29 @@
 use std::{ffi, iter};
 
 use clap::Parser;
+use fedimint_core::core::OperationId;
 use fedimint_core::util::SafeUrl;
-use fedimint_core::PeerId;
+use fedimint_core::{Amount, PeerId};
+use lightning_invoice::Bolt11Invoice;
 use serde::Serialize;
+use serde_json::Value;
 
 use crate::api::LnFederationApi;
 use crate::LightningClientModule;
 
 #[derive(Parser, Serialize)]
 enum Opts {
+    /// Pay an invoice
+    Send {
+        gateway: SafeUrl,
+        invoice: Bolt11Invoice,
+    },
+    /// Await the final state of the send operation
+    AwaitSend { operation_id: OperationId },
+    /// Request an invoice
+    Receive { gateway: SafeUrl, amount: Amount },
+    /// Await the final state of the receive operation
+    AwaitReceive { operation_id: OperationId },
     /// Fetch vetted gateways
     Gateways { peer: Option<PeerId> },
     /// Add a vetted gateway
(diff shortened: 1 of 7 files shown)


modules/fedimint-lnv2-client/src/cli.rs
```diff
@@ -2,17 +2,31 @@ use std::{ffi, iter};
...
 
 #[derive(Parser, Serialize)]
 enum Opts {
+    /// Pay an invoice
+    Send {
+        gateway: SafeUrl,
+        invoice: Bolt11Invoice,
+    },
+    /// Await the final state of the send operation
+    AwaitSend { operation_id: OperationId },
```
dpc, line 25: Any way these two (AwaitSend and AwaitReceive) could be the same command?

modules/fedimint-lnv2-tests/src/bin/self_payment_with_two_gateways.rs
```diff
@@ -0,0 +1,84 @@
...
+        let gw_cln = devfed.gw_cln().addr;
+
+        let result: Result<(Bolt11Invoice, OperationId), FetchInvoiceError> =
+            serde_json::from_value(
+                cmd!(client, "module", "lnv2", "receive", gw_lnd, "100000")
+                    .out_json()
+                    .await?,
+            );
+
+        let Ok(invoice, receive_op) = result?;
```
dpc, line 44: Splitting these results into separate vars / operations is a bit ... odd. Is there a reason for it?

`serde_json::from_value::<(Bolt11Invoice, OperationId)>::(/*...*/)` would be more idiomatic, but then I noticed that you specify `FetchInvoiceError` and I wonder if it is even correct. It possibly might be working only because the error type is actually never there. `from_value` return deserialization result ...

I'm confused myself now ... ;D

joschisan: Fixed

modules/fedimint-lnv2-client/src/cli.rs
joschisan: I do not think so as we are awaiting different possible states here.

dpc (approved):

modules/fedimint-lnv2-client/src/cli.rs
dpc: @joschisan If we could figure out which one is it, we could just do the right thing.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -94,13 +94,21 @@ const INVOICE_EXPIRATION_SECONDS_DEFAULT: u32 = 3600;
 pub enum SendState {
     Funding,
     Funded,
-    Success([u8; 32]),
+    Success,
```
m1sterc001guy, line 97: Is there any reason we no longer want to return the preimage here? Wouldn't integrating apps potentially want this information?

modules/fedimint-lnv2-tests/src/bin/self_payment_with_one_gateway.rs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        if gatewayd_version < *VERSION_0_4_0_ALPHA {
+            info!(%gatewayd_version, "Version did not support lnv2 module, skipping");
+            return Ok(());
+        }
+
+        let fed = dev_fed.fed().await?;
+
+        let gw_lnd = dev_fed.gw_lnd().await?;
```
m1sterc001guy, line 33: Can we write these tests that is agnostic to the type of gateway? The v1 tests I'm doing do this, so that we can also test the CLN gateway, but this will also be useful once we have LDK.

dpc (approved):

PR #5508, opened 2024-06-22
chore: get rid of lazy_static

### LINKED ISSUE #4834: Remove `lazy_static` and prevent future usage
As @dpc mentioned [here](https://github.com/fedimint/fedimint/pull/4828#discussion_r1552223909), Rust's support for global variables has improved over the past few years to the point where we shouldn't need to use the [`lazy_static`](https://crates.io/crates/lazy_static) crate anymore. See here for details: https://morestina.net/blog/2055/rust-global-variables-two-years-on

Let's remove existing usage of this crate, then add a semgrep to prevent it from being used on accident in the future

dpc:
Fix #4834


tvolk131:
Are we able to use [`std::cell`](https://doc.rust-lang.org/std/cell/index.html) instead of the `once_cell` crate?

dpc:
> Are we able to use [`std::cell`](https://doc.rust-lang.org/std/cell/index.html) instead of the `once_cell` crate?

@tvolk131 

LazyCell and LazyLock not stable yet. https://doc.rust-lang.org/std/cell/struct.LazyCell.html . After they are we will be.

tvolk131 (approved):
> LazyCell and LazyLock not stable yet. https://doc.rust-lang.org/std/cell/struct.LazyCell.html . After they are we will be.

Sg and LGTM! Should we make a tracking issue to remove the `once_cell` crate once those are stabilized?

dpc:
@tvolk131 We'll get to it, no worries.

PR #5509, opened 2024-06-22
chore: rename CreateInvoiceRequest.expiry to expiry_secs

tvolk131:
A bit of a nit, but the rest of the fields in our proto definitions have unit suffixes where appropriate (`_msats` or `_sats`). Should be a simple change since I don't believe we've locked in backwards-compatibility for this field yet since it's used for LNv2.


dpc (approved):
Good change, as long as it doesn't break backward compat somwhere.

m1sterc001guy (approved):

PR #5513, opened 2024-06-23
fix: enable cli to fetch gateways without auth

joschisan:
The cli can now fetch the gateways from all peers or optionally supply a peer id to only fetch from one peer. The latter option is used by the guardian to check its own gateway set.

### DIFF
--- a/modules/fedimint-lnv2-client/src/cli.rs
+++ b/modules/fedimint-lnv2-client/src/cli.rs
@@ -5,9 +5,7 @@ use fedimint_api_client::api::FederationApiExt;
 use fedimint_core::module::ApiRequestErased;
 use fedimint_core::util::SafeUrl;
 use fedimint_core::PeerId;
-use fedimint_lnv2_common::endpoint_constants::{
-    ADD_GATEWAY_ENDPOINT, GATEWAYS_ENDPOINT, REMOVE_GATEWAY_ENDPOINT,
-};
+use fedimint_lnv2_common::endpoint_constants::{ADD_GATEWAY_ENDPOINT, REMOVE_GATEWAY_ENDPOINT};
 use serde::Serialize;
 
 use crate::api::LnFederationApi;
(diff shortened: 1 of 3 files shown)


modules/fedimint-lnv2-client/src/cli.rs
```diff
@@ -4,21 +4,23 @@ use clap::Parser;
...
 };
 use serde::Serialize;
 
+use crate::api::LnFederationApi;
 use crate::LightningClientModule;
 
 #[derive(Parser, Serialize)]
 enum Opts {
+    /// Fetch vetted gateways
+    Gateways { peer: Option<PeerId> },
```
dpc, line 19: `ListGateways` to match?

Or you could have a `Gateway` subcommand that has `Add`, `List`, `Remove` command.

modules/fedimint-lnv2-client/src/cli.rs
```diff
@@ -28,7 +30,25 @@ pub(crate) async fn handle_cli_command(
     let opts = Opts::parse_from(iter::once(&ffi::OsString::from("lnv2")).chain(args.iter()));
 
     match opts {
-        Opts::Add { gateway } => {
+        Opts::Gateways { peer } => {
+            let gateways = match peer {
+                Some(peer) => {
+                    lightning
+                        .module_api
+                        .request_single_peer_typed::<Vec<SafeUrl>>(
```
dpc, line 38: Seems to me we should avoid ad-hoc invocations of of API calls, and put all as logical API method on the extension trait.

modules/fedimint-lnv2-client/src/cli.rs
joschisan: I choose gateways, because its shorter, the only user command of the three and more canonical with how I named the endpoints; list has relatively little meaning like "fetch".

modules/fedimint-lnv2-client/src/cli.rs
joschisan: Makes sense - add and remove as well?

dpc (approved):

modules/fedimint-lnv2-client/src/cli.rs
dpc: @joschisan I don't know. Seems kind of inconsistent.

BTW. We could have aliases to make common commands more convenient to type.

dpc (approved):

Issue #5525, opened 2024-06-25
DNS Tracking Issue for Guardian API Endpoint Rotation


Kodylow:
Tracking Issue for roadmap to remove the DNS dependency created by including the guardian endpoints in the federation config, hashed into the federationId. 

We're going to remove that dependency and allow guardians to rotate their endpoints, announce the api rotation, other guardians start using the new endpoint, then when clients reconnect to an unchanged guardian they'll get a signed update from the changed guardian forwarded that they can authenticate and switch to for that guardian specifically.

Building this based off discussions across devs and @elsirion's response in https://github.com/fedimint/fedimint/issues/5482#issuecomment-2189671747 , feel free to modify and update as this progresses. Tasks below are just stubs, will fill in with implementation details as those get finalized.

- [ ] #5526
- [x] #5527
- [x] #5528
- [x] #5529
- [x] #5530
- [x] ~#5531~
- [x] #5532

csuwildcat:
I was wondering, could DIDs be of use here, given they provide dynamic routing capabilities and do not suffer from the same centralization risks DNS domains do?

Here's an example of a DID implementation, for reference: https://developer.tbd.website/blog/did-dht/

Kodylow:
I think potentially for clients to find and ack announcements, but for now not part of the core fedimint repo. This issue is for tracking specifically DNS rotation options as part of 0.4 which is the minimal near term target to address the issue of a single guardian needing to change its DNS endpoint. 

We've got PRs up adding TOR that are close to landing which are another near term option if people want to use/deploy with the associated tradeoffs. Then there's longer term (0.6) discussions about other solutions and protocols where arguments for DIDs might fit.

elsirion:
This was mostly implemented in #5570, the one outstanding issue is moving to `FederattionIdV2`, which will commit to the guardian keys instead of the API endpoints.

PR #5535, opened 2024-06-26
Broadcast keys in ClientConfig + Client DB Migration

Kodylow:
Step 1 for #5525  https://github.com/fedimint/fedimint/issues/5526

This changes the federationId to be a hash of the guardians' signing keys. Still need to implement a client side migration for the GlobalClientConfig, and can potentially keep the api endpoint hash fallback for new clients to still connect to old federations (new client scans old invite code, if config has public keys uses hash of that, if config does not have public keys uses hash of api endpoints.

Feel free to force push on here for changes @joschisan


Kodylow:
Added in handling for new client connecting to old federation when the client config it gets back doesn't have the guardians' public keys.

fedimint-api-client/src/lib.rs
```diff
@@ -49,8 +48,8 @@ pub async fn try_download_client_config(invite_code: &InviteCode) -> anyhow::Res
 
     let query_strategy = FilterMap::new(
         move |cfg: ClientConfig| {
-            if federation_id.0 != cfg.global.api_endpoints.consensus_hash() {
-                bail!("Guardian api endpoint map does not hash to FederationId")
+            if federation_id != cfg.global.calculate_federation_id() {
```
dpc, line 51: Why wouldn't want to try both here, so it works with old and new Federations all the same?

fedimint-core/src/config.rs
```diff
@@ -155,6 +156,8 @@ pub struct GlobalClientConfig {
     /// API endpoints for each federation member
     #[serde(deserialize_with = "de_int_key")]
     pub api_endpoints: BTreeMap<PeerId, PeerUrl>,
+    /// Signing session keys for each federation member
+    pub broadcast_public_keys: Option<BTreeMap<PeerId, PublicKey>>,
```
dpc, line 160: Please add comment explaining the `Option`

dpc (commented):
One nit, but I guess it works.

fedimint-core/src/config.rs
Kodylow: With the changes to calculate_federation_id it does test both

dpc (dismissed):

dpc:
@Kodylow Why can't we just check both old and new ID scheme so we can avoid any migrations?

Kodylow:
This requires a client database migration which we currently don't version (we only version the client's module DBs), so we'll introduce a clientdatabaseversion just like we do server side as part of this PR

fedimint-core/src/config.rs
dpc: No comment explaining that this field was added and might not be available on old clients.

fedimint-client/src/db.rs
```diff
@@ -418,8 +418,8 @@ pub fn get_core_client_database_migrations() -> BTreeMap<DatabaseVersion, CoreMi
                 };
 
                 let config_key = ClientConfigKey { id: id.id };
-                dbtx.insert_new_entry(&config_key, &config).await;
                 dbtx.remove_entry(&id).await;
+                dbtx.insert_new_entry(&config_key, &config).await;
```
dpc, line 422: We should assert that we only found 0 or 1 existing configs, in case something ever changes in the future (we store more than one?) and this migration needs a fix.

fedimint-client/src/lib.rs
```diff
@@ -1943,6 +1944,7 @@ impl ClientBuilder {
     async fn migrate_database(&self, db: &Database) -> anyhow::Result<()> {
         // Only apply the client database migrations if the database has been
         // initialized.
+        // This only works as long as you don't change the client config
```
m1sterc001guy, line 1947: Nit: this comment can be removed since we're applying the core client migrations before this

Kodylow:
closing, will merge as part of https://github.com/fedimint/fedimint/pull/5570

PR #5539, opened 2024-06-27
chore: split/restructure gateway

tvolk131:


gateway/ln-gateway/src/lib.rs
```diff
@@ -1277,7 +1270,13 @@ impl Gateway {
             }
 
             for (federation_id, federation_config) in federations {
-                if let Some(client) = self.clients.read().await.get(federation_id) {
+                if let Some(client) = self
+                    .federation_manager
+                    .clients
```
m1sterc001guy, line 1196: Love the idea of a `FederationManager` and a `LightningManager`. Can we abstract away some of these operations so that their internal fields are not exposed? Something like `self.federation_manager.get_client(federation_id)`? (not sure if you did this in a separate commit or are planning on a separate PR

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -61,4 +64,41 @@ impl FederationManager {
             next_scid: Arc::new(Mutex::new(INITIAL_SCID)),
         }
     }
+
+    pub async fn add_client(
+        &self,
+        scid: u64,
```
m1sterc001guy, line 71: this shouldnt need the scid as a parameter, the `FederationManager` should manage it itself right?

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -101,4 +102,18 @@ impl FederationManager {
...
+        &self,
+        short_channel_id: u64,
+    ) -> Option<Spanned<ClientHandleArc>> {
+        let scid_to_feds = self.scid_to_federation.read().await;
+        let clients = self.clients.read().await;
+
+        let federation_id = scid_to_feds.get(&short_channel_id)?;
+        // TODO(tvolk131): Cloning the client here could cause issues with client
+        // shutdown (see `remove_client` above). Perhaps this function should take a
+        // lambda and pass it into `client.with_sync`.
```
m1sterc001guy, line 116: Sorry, kind of tough to tell with everything moving around. Was this a problem before?

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -116,4 +116,20 @@ impl FederationManager {
         // lambda and pass it into `client.with_sync`.
         clients.get(federation_id).cloned()
     }
+
+    // TODO(tvolk131): Optimize this function by adding a reverse map from
+    // federation_id to scid.
```
m1sterc001guy, line 121: Might not be worth it, this map is likely to be small so I don't think performance is an issue here.

gateway/ln-gateway/src/lightning_manager.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+///    Initializing -- begin intercepting HTLCs --> Connected
+///    Initializing -- gateway needs config --> Configuring
+///    Configuring -- configuration set --> Connected
+///    Connected -- load federation clients --> Running
+///    Running -- disconnected from lightning node --> Disconnected
+///    Disconnected -- re-established lightning connection --> Connected
+/// ```
+#[derive(Clone, Debug)]
+pub enum GatewayState {
```
m1sterc001guy, line 117: Hmm I don't think I agree that this should be managed by `LightningManager`. The `Initializing` and `Configuring` steps have nothing to do with the Lightning node.

gateway/ln-gateway/src/lightning_manager.rs
```diff
@@ -25,6 +27,37 @@ impl std::fmt::Debug for LightningManager {
...
+        self.set_state(GatewayState::Connected).await;
+
+        Ok((stream, ln_client))
+    }
+
+    pub async fn get_state(&self) -> GatewayState {
+        self.state.read().await.clone()
+    }
+
+    pub async fn get_state_string(&self) -> String {
```
m1sterc001guy, line 95: eh, I dont think this is necessary, its just as ergonomic to call `to_string()` after `get_state`

gateway/ln-gateway/src/lightning_manager.rs
```diff
@@ -31,25 +36,46 @@ impl LightningManager {
...
         }
     }
 
     pub async fn connect_route_htlcs(
         &self,
-        task_group: &mut TaskGroup,
+        task_group: TaskGroup,
     ) -> Result<(RouteHtlcStream, Arc<dyn ILnRpcClient>), LightningRpcError> {
+        // If the gateway is already connected, stop the current HTLC interception.
+        self.stop_route_htlcs().await;
```
m1sterc001guy, line 48: I don't think we did this before. Did you find that this was necessary?

gateway/ln-gateway/src/lightning_manager.rs
```diff
@@ -31,25 +36,46 @@ impl LightningManager {
...
     }
 
+    /// Sets the gateway state to `Disconnected` and shuts down the task that is
+    /// listening for intercepted HTLCs if it is running.
+    pub async fn disconnect_stop_route_htlcs(&self) {
+        self.set_state(GatewayState::Disconnected).await;
+        self.stop_route_htlcs().await;
+    }
+
+    async fn stop_route_htlcs(&self) {
```
m1sterc001guy, line 71: Add function comment

gateway/ln-gateway/src/lightning_manager.rs
```diff
@@ -16,6 +17,10 @@ pub struct LightningManager {
 
     /// The current state of the Gateway.
     state: Arc<RwLock<GatewayState>>,
+
+    /// Task group for managing HTLC interception.
+    /// Is `Some` when the gateway is intercepting HTLCs, `None` otherwise.
+    htlc_task_group_or: Arc<RwLock<Option<TaskGroup>>>,
```
m1sterc001guy, line 23: The `Option` here is pretty ugly. Why is it necessary? We shouldn't need to recreate the `TaskGroup` anytime we disconnect from the Lightning node. Can we do something similar to what we do now, where we just pass in the `TaskGroup` as a parameter to a function, instead of storing it as a member variable here? e.g. `lightning_manager.listen_for_htlcs(task_group).await` ?

gateway/ln-gateway/src/lib.rs
tvolk131: 💯 agree, the data in these structs should ideally be silo'd from the rest of the gateway

Planning this in a follow-up PR!

gateway/ln-gateway/src/federation_manager.rs
tvolk131: The reason we need it here for now is because when we restart the gateway and load the clients from the DB, they are already assigned SCIDs. After some more untangling between `Gateway` and the new `LightningManager` + `FederationManager` structs, I think we should be able to ditch this. The main focus of this PR was to keep the structure of the gateway relatively unchanged but simply separate related data into their own structs.

gateway/ln-gateway/src/federation_manager.rs
tvolk131: Yes. On master, you can search for `.clients` in [`gateway/ln-gateway/src/lib.rs`](https://github.com/fedimint/fedimint/blob/master/gateway/ln-gateway/src/lib.rs) and find where `.clone()` or `.cloned()` are called on it. I'm guessing it hasn't surfaced any issues yet, but it strikes me as bad practice and could lead to the issue described in this comment. IMO ideally we should uphold the invariant that there is never more than one reference to a client by simply not using an `Arc` to hold it.

gateway/ln-gateway/src/lib.rs
```diff
@@ -606,49 +606,43 @@ impl Gateway {
...
+                                        info!("Got no HTLC result");
+                                    }
+                                    None
+                                })
+                                .await;
+                            if let Some(ControlFlow::Continue(())) = cf {
+                                continue;
                             }
+                        } else {
+                            info!("Got no client result");
```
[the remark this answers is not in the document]
tvolk131: Ahhh good catch! Moved to `FederationManager::get_client_for_scid` and made the error message more descriptive.

gateway/ln-gateway/src/federation_manager.rs
tvolk131: Removed.

gateway/ln-gateway/src/lightning_manager.rs
tvolk131: Should we just remove the `LightningManager` struct then? That just leaves it as a shell around the `LightningBuilder`

gateway/ln-gateway/src/lightning_manager.rs
tvolk131: It's not strictly necessary, but I wanted the `LightningManager` to fully own its task group while providing an interface that contains as few foot-guns/idiosyncrasies as possible. `LightningManager::stop_route_htlcs` is idempotent (if it's not routing any HTLCs, it just returns without doing anything), and calling that here makes `LightningManager::connect_route_htlcs` idempotent as well (since that way we're not just tossing the old task group into the wind).

gateway/ln-gateway/src/lightning_manager.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning_manager.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning_manager.rs
tvolk131: For my clarification, are you suggesting that we change

`htlc_task_group_or: Arc<RwLock<Option<TaskGroup>>>`

to

`htlc_task_group: Arc<RwLock<TaskGroup>>`

and then add a `TaskGroup` function parameter to `LightningManager::new_initializing`?

Or are you suggesting we no longer have `LightingManager` own the HTLC task group?

tvolk131:
Going to split this up into separate PRs for:

* `FederationManager` struct
* `LightningManager` struct
* Misc/cleanup

gateway/ln-gateway/src/federation_manager.rs
m1sterc001guy: Perhaps loading the clients from the db should also be apart of `FederationManager`

gateway/ln-gateway/src/lightning_manager.rs
m1sterc001guy: I think I still like `LightningManager` for the encapsulation of HTLC interception. It might get a little messy with the `FederationManager` though, since that needs to lookup clients. Let's do the `FederationManager` refactor first and see if it makes sense then.

gateway/ln-gateway/src/lightning_manager.rs
m1sterc001guy: I was hoping to not have `LightningManager` own the task group. Just pass it in as a parameter to the HTLC interception function.

gateway/ln-gateway/src/federation_manager.rs
m1sterc001guy: I think we had some sort of reference counting in `ClientArc` that solved this issue. Maybe @dpc or @maan2003 remembers. I think if we don't use an `Arc` it would prevent us from doing operations in parallel on the same client (maybe that's ok), e.g lookup the balance via the REST API and initiate a payment

PR #5540, opened 2024-06-27
chore: add recoverytool to fedimintd and devtools

m1sterc001guy:
`fedimintd` and `devtools` containers don't currently have the recoverytool. This PR adds it to both images.

### DIFF
--- a/nix/flakebox.nix
+++ b/nix/flakebox.nix
@@ -519,6 +519,7 @@ rec {
       "fedimintd"
       "fedimint-cli"
       "fedimint-dbtool"
+      "fedimint-recoverytool"
     ];
   };
 
@@ -583,6 +584,12 @@ rec {
     {
       pkg = gateway-pkgs;
       bin = "gateway-cln-extension";
+    };
+
+  fedimint-recoverytool = flakeboxLib.pickBinary
+    {
+      pkg = fedimint-pkgs;
+      bin = "recoverytool";
     };
 
   container =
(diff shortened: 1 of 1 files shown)


bradleystachurski (dismissed):

nix/flakebox.nix
```diff
@@ -585,6 +586,12 @@ rec {
       bin = "gateway-cln-extension";
     };
 
+  recoverytool = flakeboxLib.pickBinary
```
dpc, line 589: `fedimint-recoverytool` ?

Also, we should fix: https://github.com/NixOS/nixpkgs/blob/c8dec6df384ce6f24866adafbe11219ea5106be6/pkgs/by-name/fe/fedimint/package.nix#L57

m1sterc001guy: Oh, ugly. That's what it's called in `Cargo.toml`. I think if we change it to `recoverytool` there it should work

m1sterc001guy: ah sorry, misinterpreted. `fedimint-recoverytool` is probably better

dpc: Even in this file (flakebox.nix) there is `fedimint-dbtool` few lines above. But yeah - we should just make recoverytool look like `fedimint-dbtool` in all these aspects, from Cargo.toml to nix files.

m1sterc001guy: done

dpc (approved):

PR #5541, opened 2024-06-27
feat: Guardian Endpoint for signing client config with public broadcast key

Kodylow:
Builds on https://github.com/fedimint/fedimint/pull/5535 to address https://github.com/fedimint/fedimint/issues/5525

When client receives a V1 Client Config back from the federation (contains the guardians' broadcast keys and federation_id == hash(broadcast_keys) , client hits each endpoint with a challenge and each guardian responds back with a signed client config. When the client receives a threshold of these signed client configs that verify, client knows the client_config and its api_endpoints are correct.


fedimint-core/src/config.rs
```diff
@@ -155,6 +156,8 @@ pub struct GlobalClientConfig {
     /// API endpoints for each federation member
     #[serde(deserialize_with = "de_int_key")]
     pub api_endpoints: BTreeMap<PeerId, PeerUrl>,
+    /// Signing session keys for each federation member
+    pub broadcast_public_keys: BTreeMap<PeerId, PublicKey>,
```
elsirion: Likely needs to be [default](https://serde.rs/attr-default.html) if not present since old APIs won't return this field.

elsirion: never mind, appears to have become an `Option<_>` later on

dpc:
I'm under impression this was built on by Eric, and landed in #5570 , so closing. If I'm mistaken, please reopen.

Issue #5498, opened 2024-06-21
Reducde confirmation target from 10 blocks to 1


bradleystachurski:
An advantage of RBF withdrawals is giving the user the ability to increase the bump the feerate for a withdrawal if fees spike. Since RBF withdrawals will be removed due to a bug (see: https://github.com/fedimint/fedimint/issues/5453), we should use a more aggressive confirmation target.

I think we should reduce from 10 blocks to 1.

oleonardolima:
AFAIK using 10 blocks as a confirmation target brings safety against reorganization. How a reorg could be handled when using a smaller target (e.g. 1) ?

oleonardolima:
Oh, please ignore my previous comment if I misunderstood it as the deposit flow 🤔

benthecarman:
Yes please

bradleystachurski:
@oleonardolima No worries! Apologies if the description is unclear.

`CONFIRMATION_TARGET` is used to estimate the feerate that will incentivize miners to include a transaction within the confirmation target blocks.

https://github.com/fedimint/fedimint/blob/6815b57add1f58e39db6e943b4019741e3305eef/modules/fedimint-wallet-server/src/lib.rs#L1242

 It's currently hardcoded to 10, so peg-out transactions are expected to be mined with 10 blocks. Waiting 10 blocks for a peg-out to confirm ties up the wallet's liquidity and puts the transaction at risk of confirming much later if fees spike. Since we're deprecating RBF withdrawals, we should set a confirmation target that reduces the risk of peg-outs getting stuck in the mempool due to fee spikes.

PR #5543, opened 2024-06-28
chore(wallet): use next block conf target to estimate feerate

bradleystachurski:
Closes https://github.com/fedimint/fedimint/issues/5498

### DIFF
--- a/modules/fedimint-wallet-common/src/lib.rs
+++ b/modules/fedimint-wallet-common/src/lib.rs
@@ -35,7 +35,15 @@ pub mod txoproof;
 pub const KIND: ModuleKind = ModuleKind::from_static_str("wallet");
 pub const MODULE_CONSENSUS_VERSION: ModuleConsensusVersion = ModuleConsensusVersion::new(2, 1);
 
-pub const CONFIRMATION_TARGET: u16 = 10;
+/// Used for estimating a feerate that will confirm within a target number of
+/// blocks.
+///
+/// Since the wallet's UTXOs are a shared resource, we need to reduce the risk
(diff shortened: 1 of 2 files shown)


modules/fedimint-wallet-common/src/lib.rs
```diff
@@ -35,7 +35,7 @@ pub mod txoproof;
 pub const KIND: ModuleKind = ModuleKind::from_static_str("wallet");
 pub const MODULE_CONSENSUS_VERSION: ModuleConsensusVersion = ModuleConsensusVersion::new(2, 1);
 
-pub const CONFIRMATION_TARGET: u16 = 10;
+pub const CONFIRMATION_TARGET: u16 = 1;
```
tvolk131, line 38: Can we also update this?

https://github.com/fedimint/fedimint/blob/6dd24f54dbad66edf2c99529523f5e3d2110e175/docs/wallet_module.md?plain=1#L25

tvolk131: Oh also, can you add a comment above this constant mentioning why the target is 1? I think just a brief mention of lack of RBF support and federation UTXOs being a shared resource would be useful.

dpc (dismissed):

modules/fedimint-wallet-common/src/lib.rs
bradleystachurski: Good call 👍

bradleystachurski:
Looks like a blip with the runner while running back compat

> The self-hosted runner: runner-01-bb lost communication with the server

https://github.com/fedimint/fedimint/actions/runs/9761965930

PR #5545, opened 2024-06-28
chore: remove GatewayTest

m1sterc001guy:
Working towards: https://github.com/fedimint/fedimint/issues/5168

This should be one of the last cleanup PRs of the Lightning tests. This PR removes the `GatewayTest` struct.

`GatewayTest` was just a wrapper around `Gateway` that started the webserver. The webserver has been a source of issues in the past, we have to poll to wait for it to come up, and it causes flakiness. We do not even need it, everything is in process so we can just directly call functions on the gateway. Now that we have devimint tests, we already have coverage of the webserver too. This PR reduces the scope of the gateway tests to what they should be - module tests :)


m1sterc001guy:
Needs https://github.com/fedimint/fedimint/pull/5564

joschisan (approved):

PR #5546, opened 2024-06-28
feature(wallet-client): client wallet module backup & recovery

dpc:
During recovery scan the pegin addresses to find a best place to start deriving new
addresses.

Since we are already checking them and potentially leaking them to the node.

Notably adds module-cli `fedimint-cli module wallet ...` support.

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -142,6 +142,47 @@ impl Client {
         Ok(())
     }
 
+    /// Client to join a federation with a restore procedure
+    pub async fn restore_federation(&self, invite_code: String, mnemonic: String) -> Result<()> {
+        debug!(target: LOG_DEVIMINT, "Joining federation with restore procedure");
+        cmd!(
+            self,
+            "restore",
+            "--invite-code",
+            invite_code,
+            "--mnemonic",
+            mnemonic
+        )
+        .run()
+        .await?;
+
+        Ok(())
+    }
+
+    /// Client to join a federation
+    pub async fn new_restored(&self, name: &str, invite_code: String) -> Result<Self> {
+        let restored = Self::open_or_create(name)?;
+
+        let mnemonic = cmd!(self, "print-secret").out_json().await?["secret"]
+            .as_str()
+            .unwrap()
+            .to_owned();
+
+        debug!(target: LOG_DEVIMINT, name, "Restoring from mnemonic");
+        cmd!(
+            restored,
+            "restore",
+            "--invite-code",
+            invite_code,
+            "--mnemonic",
+            mnemonic
+        )
+        .run()
+        .await?;
+
+        Ok(restored)
+    }
+
     /// Create a [`Client`] that starts with a state that is a copy of
     /// of another one.
     pub async fn new_forked(&self, name: &str) -> Result<Client> {
@@ -210,6 +251,11 @@ impl Client {
         cmd!(self, "dev", "session-count").out_json().await?["count"]
             .as_u64()
             .context("count field wasn't a number")
+    }
+
+    /// Returns the current consensus session count
+    pub async fn wait_complete(&self) -> Result<()> {
+        cmd!(self, "dev", "wait-complete").run().await
     }
 }
 
@@ -522,7 +568,7 @@ impl Federation {
         Ok(())
     }
 
-    pub async fn pegin_client(&self, amount: u64, client: &Client) -> Result<()> {
+    pub async fn pegin_client_no_wait(&self, amount: u64, client: &Client) -> Result<String> {
         let deposit_fees_msat = self.deposit_fees()?.msats;
         assert_eq!(
             deposit_fees_msat % 1000,
@@ -538,6 +584,12 @@ impl Federation {
             .send_to(address, amount + deposit_fees)
             .await?;
         self.bitcoind.mine_blocks(21).await?;
+
+        Ok(operation_id)
+    }
+
+    pub async fn pegin_client(&self, amount: u64, client: &Client) -> Result<()> {
+        let operation_id = self.pegin_client_no_wait(amount, client).await?;
 
         client.await_deposit(&operation_id).await?;
         Ok(())
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -7,6 +7,10 @@
 #![allow(clippy::must_use_candidate)]
 
 pub mod api;
+#[cfg(feature = "cli")]
+mod cli;
+
+mod backup;
 
 pub mod client_db;
 /// Legacy, state-machine based peg-ins, replaced by `pegin_monitor`
@@ -16,21 +20,24 @@ mod deposit;
 mod pegin_monitor;
 mod withdraw;
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::future;
 use std::sync::Arc;
 use std::time::SystemTime;
 
 use anyhow::{anyhow, bail, ensure, Context as AnyhowContext};
 use async_stream::stream;
+use backup::WalletModuleBackup;
 use bitcoin::address::NetworkUnchecked;
 use bitcoin::{Address, Network, ScriptBuf};
 use client_db::{DbKeyPrefix, PegInTweakIndexKey, TweakIdx};
 use fedimint_api_client::api::DynModuleApi;
 use fedimint_bitcoind::{create_bitcoind, DynBitcoindRpc};
 use fedimint_client::derivable_secret::{ChildId, DerivableSecret};
-use fedimint_client::module::init::{ClientModuleInit, ClientModuleInitArgs};
-use fedimint_client::module::recovery::NoModuleBackup;
+use fedimint_client::module::init::{
+    ClientModuleInit, ClientModuleInitArgs, ClientModuleRecoverArgs,
+};
+use fedimint_client::module::recovery::RecoveryProgress;
 use fedimint_client::module::{ClientContext, ClientModule, IClientModule};
 use fedimint_client::oplog::UpdateStreamOrOutcome;
 use fedimint_client::sm::util::MapStateTransitions;
@@ -66,7 +73,7 @@ use tracing::{debug, instrument};
 use crate::api::WalletFederationApi;
 use crate::client_db::{
     ClaimedPegInData, ClaimedPegInKey, ClaimedPegInPrefix, NextPegInTweakIndexKey,
-    PegInTweakIndexData, PegInTweakIndexPrefix,
+    PegInTweakIndexData, PegInTweakIndexPrefix, RecoveryFinalizedKey,
 };
 use crate::deposit::DepositStateMachine;
 use crate::withdraw::{CreatedWithdrawState, WithdrawStateMachine, WithdrawStates};
@@ -166,6 +173,7 @@ impl ModuleInit for WalletClientInit {
                         "Claimed Peg-In"
                     );
                 }
+                DbKeyPrefix::RecoveryFinalized => {}
             }
         }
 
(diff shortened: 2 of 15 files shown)


modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -215,6 +223,132 @@ impl ClientModuleInit for WalletClientInit {
...
+            .get_value(&RecoveryFinalizedKey)
+            .await
+            .unwrap_or_default()
+        {
+            debug!(target: LOG_CLIENT_MODULE_WALLET, max_gap=MAX_GAP, "Recovery already complete before");
+            return Ok(());
+        }
+
+        let start_tweak_idx = match _snapshot {
+            Some(WalletModuleBackup::V0(backup)) => {
```
bradleystachurski, line 252: I believe we need to override `supports_backup` to return true in the wallet client, similar to the mint client. Without that, the client will skip including the wallet module backup when it creates and sends an encrypted backup to the federation.

https://github.com/fedimint/fedimint/blob/9c9b1bda637222993bdbe5865e592cd6a37e26f2/fedimint-client/src/module/mod.rs#L626

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -215,6 +223,132 @@ impl ClientModuleInit for WalletClientInit {
...
+
+        let new_start_tweak_idx = loop {
+            let gap_since_last_used = tweak_idx - last_used_idx.unwrap_or_default();
+
+            args.update_recovery_progress(RecoveryProgress {
+                complete: u32::try_from(tweak_idx.0).unwrap_or(u32::MAX),
+                total: u32::try_from(tweak_idx.0.saturating_add(MAX_GAP)).unwrap_or(u32::MAX),
+            });
+
+            if MAX_GAP < gap_since_last_used {
```
bradleystachurski, line 296: I think this causes an issue for backups with an index larger than the max gap.

ex:
  - `backup.next_tweak_idx == 100`
  - first iteration of loop, `last_used_idx == None`
  - `gap_since_last_used == 100`
  - loop breaks, `new_start_tweak_idx == 0 + 3`

Even though we know our next tweak index from the backup is 100, we'd only check the first 3 indexes and save the next peg-in tweak index as 3, causing us to reuse address and potentially miss claiming peg-ins that used an index greater than the backup.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -215,6 +223,132 @@ impl ClientModuleInit for WalletClientInit {
             task_group: args.task_group().clone(),
         })
     }
+
+    /// Wallet recovery
+    ///
+    /// Query bitcoin rpc for history of addresses from last known used
+    /// addresses (or index 0 one) until MAX_GAP unused ones.
```
bradleystachurski, line 230: [nit] rm "one"

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -215,6 +223,132 @@ impl ClientModuleInit for WalletClientInit {
...
+    ///
+    /// Notably does not persist the progress of addresses being querried,
```
bradleystachurski, line 232: [nit] s/querried/queried

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -215,6 +223,132 @@ impl ClientModuleInit for WalletClientInit {
...
+            .await
+            .get_value(&RecoveryFinalizedKey)
+            .await
+            .unwrap_or_default()
+        {
+            debug!(target: LOG_CLIENT_MODULE_WALLET, max_gap=MAX_GAP, "Recovery already complete before");
+            return Ok(());
+        }
+
+        let start_tweak_idx = match _snapshot {
```
bradleystachurski, line 251: [nit] since `_snapshot` is used, should we remove the `_`?

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -215,6 +223,132 @@ impl ClientModuleInit for WalletClientInit {
...
     }
+
+    /// Wallet recovery
+    ///
+    /// Query bitcoin rpc for history of addresses from last known used
+    /// addresses (or index 0 one) until MAX_GAP unused ones.
+    ///
+    /// Notably does not persist the progress of addresses being querried,
+    /// because it is not expected that it would take long enough to bother.
+    async fn recover(
```
bradleystachurski, line 234: [nit] We can update the fedimint cli docs for restore to also include the wallet module

https://github.com/fedimint/fedimint/blob/9c9b1bda637222993bdbe5865e592cd6a37e26f2/fedimint-cli/src/client.rs#L172-L173

devimint/src/federation.rs
```diff
@@ -142,6 +142,47 @@ impl Client {
         Ok(())
     }
 
+    /// Client to join a federation
+    pub async fn restore_federation(&self, invite_code: String, mnemonic: String) -> Result<()> {
```
bradleystachurski, line 146: [nit] I believe this is unused. If we want to keep it, it might worth updating the docs and the debug log.

modules/fedimint-wallet-tests/src/bin/wallet-recovery-test.rs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            let client = fed
+                .new_joined_client("wallet-client-recovery-origin")
+                .await?;
+            assert_eq!(0, client.balance().await?);
+
+            info!("Join and claim");
+            fed.pegin_client(peg_in_amount_sats, &client).await?;
+
+            info!("Backup a backup");
```
bradleystachurski, line 53: [nit] 😄

modules/fedimint-wallet-client/src/lib.rs
dpc: Oof.

modules/fedimint-wallet-client/src/lib.rs
dpc: Well spotted.

Changing:

```
            let gap_since_last_used = tweak_idx - last_used_idx.unwrap_or(start_tweak_idx);
```

bradleystachurski: Beautiful 👍 

With that approach, you'll also need to `unwrap_or` when we break. Alternatively, I think you can also init  `last_used_idx = start_tweak_idx`, something like

```rust
        let mut last_used_idx = start_tweak_idx;
        let mut current_tweak_idx = start_tweak_idx;
        let mut tweak_idxes_with_pegins = BTreeSet::new();

        debug!(target: LOG_CLIENT_MODULE_WALLET, max_gap=MAX_GAP, %start_tweak_idx, "Checking history");

        let new_start_tweak_idx = loop {
            let gap_since_last_used = current_tweak_idx - last_used_idx;

            args.update_recovery_progress(RecoveryProgress {
                complete: u32::try_from(current_tweak_idx.0).unwrap_or(u32::MAX),
                total: u32::try_from(current_tweak_idx.0.saturating_add(MAX_GAP))
                    .unwrap_or(u32::MAX),
            });

            if MAX_GAP < gap_since_last_used {
                break last_used_idx.advance(NUM_IDX_SKIP_AFTER_LAST_USED);
            }

            let (script, address, _tweak_key, _operation_id) =
                data.derive_peg_in_script(current_tweak_idx);
            btc_rpc.watch_script_history(&script).await?;
            let history = btc_rpc.get_script_history(&script).await?;

            debug!(target: LOG_CLIENT_MODULE_WALLET, %current_tweak_idx, %address, history_len=history.len(), "Checked address");
            if !history.is_empty() {
                last_used_idx = current_tweak_idx;
                tweak_idxes_with_pegins.insert(current_tweak_idx);
            }
            current_tweak_idx = current_tweak_idx.next();
        };
```

modules/fedimint-wallet-tests/src/bin/wallet-recovery-test.rs
dpc: :grin:

modules/fedimint-wallet-client/src/lib.rs
dpc: @bradleystachurski You're saving my butt. I'm tempted to extract the iteration logic to another function so I could easily unit-test it, but I'm afraid the additional boilerplate and abstraction is not worth it.

I'm going with `last_used_idx = start_tweak_idx;`. It makes more logical sense.

bradleystachurski: Ooh, I do love a pure function with some unit tests 🙂 

I think the tradeoff is worthwhile since the potential risks of a bug are address reuse and missing claims during a recovery, but I'll let you make the call.

dpc: I'll try it them and show it in another commit.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -215,6 +223,133 @@ impl ClientModuleInit for WalletClientInit {
...
+                data.derive_peg_in_script(tweak_idx);
+            dbtx.insert_new_entry(
+                &PegInTweakIndexKey(tweak_idx),
+                &PegInTweakIndexData {
+                    creation_time: now,
+                    next_check_time: if tweak_idxes_with_pegins.contains(&tweak_idx) {
+                        // The addresses that were already used before, or didn't seem to contain
+                        // anything don't need automatic peg-in attempt, and can be re-attempted
+                        // manually if needed.
+                        Some(now)
```
bradleystachurski, line 331: I'm thinking out loud and not sure if we need to change, but I believe this degrades privacy.

If a single client has only used Tor to claim peg-ins periodically in a mint with many Tor clients, the claims can't be correlated to a single client. If the client initiates recovery, the previously claimed peg-ins will become correlated to a single client.

Actually, now that I'm rereading the associated comment, perhaps the intention was to flip the logic? If there's already history for an address we don't schedule a next check time, but if there isn't activity we mark the tweak index due for a check now.

modules/fedimint-wallet-client/src/lib.rs
dpc: @bradleystachurski done

modules/fedimint-wallet-client/src/lib.rs
dpc: During recovery client goes through a gap of 10 sequentially derived addressess and ask for their history from blockchain nodes (esplora, etc.). That already potentially correlates them in the external source that is possibly not trusted.

In this code we trigger actual claiming attempt for ones that were detected to have history. This logic will fail (gracefully) if they were already claimed, or succeed if not. That claiming itself does not change anything for the blockchain node source, though it does correlates them in the Federation.

However I don't see a good way to avoid it entirely. We need to ask Federation to check if peg-in was claimed or not. Possibly we could add some randomized delays for stuff here and in the scanning itself, but I'm not sure if it helps much, and it does have UX downsides.


We could also make the gap smaller, and the padding higher. So e.g. we only scan looking for 5 unused addresses, and use the last a new next peg-in address. The problem here is that it could lead theoretically to situation where n-th recovery is missing things because the gap between peg-in addresses is actually higher then the gap we're scanning. Though it's a very small chance.

bradleystachurski (dismissed):
Looks good, thanks for adding unit tests!

bradleystachurski (approved):

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -921,3 +1062,183 @@ impl State for WalletClientStates {
...
+                    new_start_idx: TweakIdx(RECOVER_NUM_IDX_ADD_TO_LAST_USED),
+                    tweak_idxes_with_pegins: BTreeSet::from([])
+                }
+            );
+            assert!(last_checked.load(Ordering::SeqCst));
+        }
+
+        {
+            let last_checked = AtomicBool::new(false);
+            let last_checked = &last_checked;
```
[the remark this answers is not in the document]
dpc: Absoluted. Missed it.

PR #5548, opened 2024-06-28
chore: add gateway federation manager

tvolk131:


gateway/ln-gateway/src/lib.rs
```diff
@@ -1018,13 +1018,7 @@ impl Gateway {
...
         // The gateway deterministically assigns a channel id (u64) to each federation
         // connected.
-        let mut next_scid = self.federation_manager.next_scid.lock().await;
-        let mint_channel_id = *next_scid;
-        *next_scid = next_scid
-            .checked_add(1)
-            .ok_or(GatewayError::GatewayConfigurationError(
-                "Too many connected federations".to_string(),
-            ))?;
+        let mint_channel_id = self.federation_manager.pop_next_scid().await?;
```
tvolk131: Note for reviewers: I believe in the past `next_scid` was used to ensure synchronization when adding federations. However, above this we already claim `Gateway.client_joining_lock`. Since that lock is claimed before locking on `next_scid`, it should always be more restrictive, so any locking guarantees provided by `next_scid` shouldn't be needed.

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -127,4 +127,19 @@ impl FederationManager {
     pub async fn has_federation(&self, federation_id: FederationId) -> bool {
         self.clients.read().await.contains_key(&federation_id)
     }
+
+    pub async fn set_next_scid(&self, next_scid: u64) {
+        *self.next_scid.lock().await = next_scid;
+    }
```
elsirion: Would need more refactoring, but feels like that should be set by the constructor.

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -81,77 +69,62 @@ impl FederationManager {
...
+    pub fn has_federation(&self, federation_id: FederationId) -> bool {
+        self.clients.contains_key(&federation_id)
     }
 
-    // TODO(tvolk131): Sharing read lock guards could lead to write contention since
-    // we don't know how long callers will hold the guard for. Replace this function
-    // with more granular accessors that don't share lock guards.
-    pub async fn borrow_clients(
+    // TODO(tvolk131): Replace this function with more granular accessors.
+    pub fn borrow_clients(
```
elsirion, line 128: nit: just call it `clients()`?

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -73,6 +75,24 @@ impl FederationManager {
         Ok(())
     }
 
+    /// Iterates through all of the federations the gateway is registered with
+    /// and requests to remove the registration record.
+    pub async fn leave_all_federations(&self, gateway_keypair: KeyPair) {
```
elsirion, line 80: You just copied the name, but it's confusing imo. It's more like "un-announcing" ourselves, but not directly leaving yet (that would imply shutting down the client imo, which this fn can't do with an immutable self ref)

elsirion (commented):
Some nits, but looks good in general

gateway/ln-gateway/src/federation_manager.rs
m1sterc001guy: Kind of leaks the encapsulation. Why is this necessary currently?

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -0,0 +1,150 @@
...
+            .remove(&federation_id)
+            .ok_or(GatewayError::InvalidMetadata(format!(
+                "No federation with id {federation_id}"
+            )))?
+            .into_value();
+
+        if let Some(client) = Arc::into_inner(client) {
+            client.shutdown().await;
+        } else {
+            error!("client is not unique, failed to remove client");
```
m1sterc001guy, line 70: Should this return an error?

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -81,4 +82,23 @@ impl FederationManager {
...
+        let clients = self.clients.read().await;
+
+        let federation_id = scid_to_feds.get(&short_channel_id)?;
+        // TODO(tvolk131): Cloning the client here could cause issues with client
+        // shutdown (see `remove_client` above). Perhaps this function should take a
+        // lambda and pass it into `client.with_sync`.
+        if let Some(client) = clients.get(federation_id).cloned() {
+            Some(client)
+        } else {
+            error!("`FederationManager.scid_to_federation` is out of sync with `FederationManager.clients`! This is a bug.");
```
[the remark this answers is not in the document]
m1sterc001guy: +1 for adding panic here

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -41,4 +44,41 @@ impl FederationManager {
             next_scid: Arc::new(Mutex::new(INITIAL_SCID)),
         }
     }
+
+    pub async fn add_client(
+        &self,
+        scid: u64,
+        federation_id: FederationId,
+        client: Spanned<fedimint_client::ClientHandleArc>,
```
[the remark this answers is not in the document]
tvolk131: Good catch!

gateway/ln-gateway/src/lib.rs
```diff
@@ -721,48 +697,39 @@ impl Gateway {
...
+                            let htlc = htlc_request.clone().try_into();
+                            if let Ok(htlc) = htlc {
+                                match client
+                                    .get_first_module::<GatewayClientModule>()
+                                    .gateway_handle_intercepted_htlc(htlc)
+                                    .await
+                                {
+                                    Ok(_) => return Some(ControlFlow::<(), ()>::Continue(())),
+                                    Err(e) => {
+                                        info!("Got error intercepting HTLC: {e:?}, will retry...");
```
m1sterc001guy, line 722: Can we improve these errors while we're refactoring this? This should be `error!`, as well as below

gateway/ln-gateway/src/lib.rs
```diff
@@ -813,15 +780,22 @@ impl Gateway {
...
+                federations.push(
+                    client
+                        .borrow()
+                        .with(|client| self.make_federation_info(client, federation_id))
+                        .await,
+                );
+            }
+
+            (federations, federation_manager.clone_scid_map())
+        };
```
m1sterc001guy, line 798: Let's make this a function inside `FederationManager`

gateway/ln-gateway/src/lib.rs
```diff
@@ -860,8 +834,13 @@ impl Gateway {
                     client.borrow().with_sync(|client| client.get_config_json()),
                 );
             } else {
-                let federation_clients = self.clients.read().await.clone().into_iter();
-                for (federation_id, client) in federation_clients {
+                for (federation_id, client) in self
```
m1sterc001guy, line 837: I think almost this entire function `handle_get_federation_config` should move into `FederationManager`

gateway/ln-gateway/src/lib.rs
```diff
@@ -1351,7 +1328,13 @@ impl Gateway {
             }
 
             for (federation_id, federation_config) in federations {
-                if let Some(client) = self.clients.read().await.get(federation_id) {
+                if let Some(client) = self
+                    .federation_manager
+                    .read()
+                    .await
+                    .borrow_clients()
```
m1sterc001guy, line 1335: IMO there should be a `get_client_by_id` function rather than borrowing the entire clients map

gateway/ln-gateway/src/federation_manager.rs
tvolk131: Added a TODO comment

gateway/ln-gateway/src/lib.rs
m1sterc001guy: There's a few places where this would be useful

gateway/ln-gateway/src/federation_manager.rs
tvolk131: Renamed to `unannounce_from_all_federations`

gateway/ln-gateway/src/federation_manager.rs
tvolk131: Yeah probably. Done.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

gateway/ln-gateway/src/lib.rs
tvolk131: It's not quite that easy because `Gateway::make_federation_info` requires reading from the database. Surely there's a cleaner way to do this though - I dropped a TODO comment there to address this later.

gateway/ln-gateway/src/lib.rs
tvolk131: Done. I split it into `get_federation_config()` and `get_all_federation_configs()` within `FederationManager`.

elsirion:
Possible backport candidate if it doesn't get into 0.4.0-rc.0 since having a huge structural divergence between master and release will make emergency fix backports much harder.

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -0,0 +1,150 @@
...
+use crate::state_machine::GatewayClientModule;
+use crate::{GatewayError, Result};
+
+/// The first SCID that the gateway will assign to a federation.
+/// Note: This starts at 1 because an SCID of 0 is considered invalid by LND's
+/// HTLC interceptor.
+const INITIAL_SCID: u64 = 1;
+
+#[derive(Debug)]
+pub struct FederationManager {
```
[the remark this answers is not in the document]
tvolk131: Should we do this in a follow-up PR? I added a TODO comment.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

tvolk131:
> Possible backport candidate if it doesn't get into 0.4.0-rc.0

Last-minute submission before the release cut? 👀 Would love to not have to backport

elsirion (approved):

PR #5554, opened 2024-07-01
feat: add ldk gateway

tvolk131:
Add a third lightning node type to the gateway, a lightning node powered by [`ldk-node`](https://crates.io/crates/ldk-node) which runs in the same process as the gateway. The gateway can be run in ldk mode by setting the following environment variables:

`FM_LDK_ESPLORA_SERVER_URL`: Esplora server where blockchain data is accessed
`FM_LDK_NETWORK`: The bitcoin network being used (e.g. mainnet/testnet)
`FM_PORT_LDK`: The port that the lightning server is running on, where other lightning nodes can access it

The node's data is stored in a subdirectory of the gateway's data folder

This gateway mode should still be considered experimental!!! None of the behavior is tested or has locked-in backwards compatibility yet!

### DIFF
--- a/gateway/ln-gateway/src/lightning/ldk.rs
+++ b/gateway/ln-gateway/src/lightning/ldk.rs
@@ -1,0 +1,493 @@
+use std::path::Path;
+use std::str::FromStr;
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use bitcoin::{secp256k1, Network, OutPoint};
+use fedimint_core::runtime::spawn;
+use fedimint_core::task::TaskGroup;
+use fedimint_core::Amount;
+use ldk_node::lightning::ln::msgs::SocketAddress;
+use ldk_node::lightning::ln::PaymentHash;
+use ldk_node::lightning_invoice::Bolt11Invoice;
+use ldk_node::payment::{PaymentKind, PaymentStatus};
+use lightning::ln::PaymentPreimage;
+use lightning::util::scid_utils::scid_from_parts;
+use tokio::sync::mpsc::Sender;
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::Status;
+use tracing::{error, info};
+
+use super::{ChannelInfo, ILnRpcClient, LightningRpcError, RouteHtlcStream};
+use crate::gateway_lnrpc::create_invoice_request::Description;
+use crate::gateway_lnrpc::intercept_htlc_response::{Action, Settle};
+use crate::gateway_lnrpc::{
+    CloseChannelsWithPeerResponse, CreateInvoiceRequest, CreateInvoiceResponse, EmptyResponse,
+    GetFundingAddressResponse, GetNodeInfoResponse, GetRouteHintsResponse, InterceptHtlcRequest,
+    InterceptHtlcResponse, PayInvoiceResponse,
+};
+
+pub struct GatewayLdkClient {
+    /// The underlying lightning node.
+    node: Arc<ldk_node::Node>,
+
+    /// The client for querying data about the blockchain.
+    esplora_client: esplora_client::AsyncClient,
+
+    /// A handle to the task that processes incoming events from the lightning
+    /// node. Responsible for sending incoming HTLCs to the caller of
+    /// `route_htlcs`.
+    /// TODO: This should be a shutdown sender instead, and we can discard the
+    /// handle.
+    event_handler_task_handle: tokio::task::JoinHandle<()>,
+
+    /// The HTLC stream, until it is taken by calling
+    /// `ILnRpcClient::route_htlcs`.
+    htlc_stream_receiver_or:
+        Option<tokio::sync::mpsc::Receiver<Result<InterceptHtlcRequest, Status>>>,
+}
+
+impl std::fmt::Debug for GatewayLdkClient {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("GatewayLdkClient").finish_non_exhaustive()
+    }
+}
+
+impl GatewayLdkClient {
+    /// Creates a new `GatewayLdkClient` instance and starts the underlying
+    /// lightning node. All resources, including the lightning node, will be
+    /// cleaned up when the returned `GatewayLdkClient` instance is dropped.
+    /// There's no need to manually stop the node.
+    pub fn new(
+        data_dir: &Path,
+        esplora_server_url: &str,
+        network: Network,
+        lightning_port: u16,
+    ) -> anyhow::Result<Self> {
+        let mut node_builder = ldk_node::Builder::from_config(ldk_node::Config {
+            network,
+            listening_addresses: Some(vec![SocketAddress::TcpIpV4 {
+                addr: [0, 0, 0, 0],
+                port: lightning_port,
+            }]),
+            onchain_wallet_sync_interval_secs: 10,
+            wallet_sync_interval_secs: 10,
+            ..Default::default()
+        });
+        node_builder
+            .set_esplora_server(esplora_server_url.to_string())
+            .set_gossip_source_p2p();
+        let Some(data_dir_str) = data_dir.to_str() else {
+            return Err(anyhow::anyhow!("Invalid data dir path"));
+        };
+        node_builder.set_storage_dir_path(data_dir_str.to_string());
+
+        let node = Arc::new(node_builder.build()?);
+        node.start().map_err(|e| {
+            error!(?e, "Failed to start LDK Node");
+            LightningRpcError::FailedToConnect
+        })?;
+
+        let (htlc_stream_sender, htlc_stream_receiver) = tokio::sync::mpsc::channel(1024);
+
+        let node_clone = node.clone();
+        let event_handler_task_handle = spawn("ldk lightning node event handler", async move {
+            loop {
+                Self::handle_next_event(&node_clone, &htlc_stream_sender).await;
+            }
+        });
+
+        Ok(GatewayLdkClient {
+            node,
+            esplora_client: esplora_client::Builder::new(esplora_server_url).build_async()?,
+            event_handler_task_handle,
+            htlc_stream_receiver_or: Some(htlc_stream_receiver),
+        })
+    }
+
+    async fn handle_next_event(
+        node: &ldk_node::Node,
+        htlc_stream_sender: &Sender<Result<InterceptHtlcRequest, Status>>,
+    ) {
+        if let ldk_node::Event::PaymentClaimable {
+            payment_id: _,
+            payment_hash,
+            claimable_amount_msat,
+            claim_deadline,
+        } = node.next_event_async().await
+        {
+            if let Err(e) = htlc_stream_sender
+                .send(Ok(InterceptHtlcRequest {
+                    payment_hash: payment_hash.0.to_vec(),
+                    incoming_amount_msat: claimable_amount_msat,
+                    outgoing_amount_msat: 0,
+                    incoming_expiry: claim_deadline.unwrap_or_default(),
+                    short_channel_id: None,
+                    incoming_chan_id: 0,
+                    htlc_id: 0,
+                }))
+                .await
+            {
+                error!(?e, "Failed send InterceptHtlcRequest to stream");
+            }
+        }
+
+        // The `PaymentClaimable` event is the only event type that we are interested
+        // in. We can safely ignore all other events.
+        node.event_handled();
+    }
+
+    /// Converts a transaction outpoint to a short channel ID by querying the
+    /// blockchain.
+    async fn outpoint_to_scid(&self, funding_txo: OutPoint) -> anyhow::Result<u64> {
+        let block_height = self
+            .esplora_client
+            .get_merkle_proof(&funding_txo.txid)
+            .await?
+            .ok_or(anyhow::anyhow!("Failed to get merkle proof"))?
+            .block_height;
+
+        let block_hash = self.esplora_client.get_block_hash(block_height).await?;
+
+        let block = self
+            .esplora_client
+            .get_block_by_hash(&block_hash)
+            .await?
+            .ok_or(anyhow::anyhow!("Failed to get block"))?;
+
+        let tx_index = block
+            .txdata
+            .iter()
+            .enumerate()
+            .find(|(_, tx)| tx.txid() == funding_txo.txid)
+            .ok_or(anyhow::anyhow!("Failed to find transaction"))?
+            .0 as u32;
+
+        let output_index = funding_txo.vout;
+
+        scid_from_parts(
+            u64::from(block_height),
+            u64::from(tx_index),
+            u64::from(output_index),
+        )
+        .map_err(|e| anyhow::anyhow!("Failed to convert to short channel ID: {e:?}"))
+    }
+}
+
+impl Drop for GatewayLdkClient {
+    fn drop(&mut self) {
+        self.event_handler_task_handle.abort();
+
+        info!("Stopping LDK Node...");
+        if let Err(e) = self.node.stop() {
+            error!(?e, "Failed to stop LDK Node");
+        } else {
+            info!("LDK Node stopped.");
+        }
+    }
+}
+
+#[async_trait]
+impl ILnRpcClient for GatewayLdkClient {
+    async fn info(&self) -> Result<GetNodeInfoResponse, LightningRpcError> {
+        let node_status = self.node.status();
+
+        let Some(chain_tip_block_summary) = self
+            .esplora_client
+            .get_blocks(None)
+            .await
+            .map_err(|e| LightningRpcError::FailedToGetNodeInfo {
+                failure_reason: format!("Failed get chain tip block summary: {e:?}"),
+            })?
+            .into_iter()
+            .next()
+        else {
+            return Err(LightningRpcError::FailedToGetNodeInfo {
+                failure_reason:
+                    "Failed to get chain tip block summary (empty block list was returned)"
+                        .to_string(),
+            });
+        };
+
+        let esplora_chain_tip_timestamp = chain_tip_block_summary.time.timestamp;
+        let block_height: u32 = chain_tip_block_summary.time.height;
+
+        let synced_to_chain = node_status.latest_wallet_sync_timestamp.unwrap_or_default()
+            > esplora_chain_tip_timestamp
+            && node_status
+                .latest_onchain_wallet_sync_timestamp
+                .unwrap_or_default()
+                > esplora_chain_tip_timestamp;
+
+        Ok(GetNodeInfoResponse {
+            pub_key: self.node.node_id().serialize().to_vec(),
+            // TODO: This is a placeholder. We need to get the actual alias from the LDK node.
+            alias: format!("LDK Fedimint Gateway Node {}", self.node.node_id()),
+            network: match self.node.config().network {
+                Network::Bitcoin => "main",
+                Network::Testnet => "test",
+                Network::Signet => "signet",
+                Network::Regtest => "regtest",
+                _ => panic!("Unsupported network"),
+            }
+            .to_string(),
+            block_height,
+            synced_to_chain,
+        })
+    }
+
+    async fn routehints(
+        &self,
+        _num_route_hints: usize,
+    ) -> Result<GetRouteHintsResponse, LightningRpcError> {
+        // TODO: Return real route hints. Not strictly necessary but would be nice to
+        // have.
+        Ok(GetRouteHintsResponse {
+            route_hints: vec![],
+        })
+    }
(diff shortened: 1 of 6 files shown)


gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+
+    /// A handle to the task that seeds the HTLC stream.
+    /// TODO: This should be a shutdown sender instead, and we can discard the
+    /// handle.
+    htlc_stream_seeder_task_handle: tokio::task::JoinHandle<()>,
+
+    /// The HTLC stream, until it is taken by calling
+    /// `ILnRpcClient::route_htlcs`.
+    htlc_stream_receiver_or:
+        Option<tokio::sync::mpsc::Receiver<Result<InterceptHtlcRequest, Status>>>,
```
m1sterc001guy, line 45: This seems really awkward to me. Does `GatewayLdkClient` need to own this? Can it be passed in as an argument to a function?

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+            .set_esplora_server(esplora_server_url.to_string())
+            .set_gossip_source_p2p();
+        let Some(data_dir_str) = data_dir.to_str() else {
+            return Err(anyhow::anyhow!("Invalid data dir path"));
+        };
+        node_builder.set_storage_dir_path(data_dir_str.to_string());
+
+        let node = Arc::new(node_builder.build()?);
+        node.start().map_err(|e| {
+            error!("Failed to start LDK Node: {e:?}");
```
m1sterc001guy, line 85: nit: can avoid string interpolation `error!(?e, "Failed to start LDK Node");`

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        Ok(GatewayLdkClient {
+            node,
+            esplora_client: esplora_client::Builder::new(esplora_server_url).build_async()?,
+            htlc_stream_seeder_task_handle,
+            htlc_stream_receiver_or: Some(htlc_stream_receiver),
+        })
+    }
+
+    async fn seed_route_htlcs_stream(
```
m1sterc001guy, line 114: "seed" seems like a weird word for this? This is essentially a "payment interceptor" right? Similar to LND's HTLC interceptor

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        let (htlc_stream_sender, htlc_stream_receiver) = tokio::sync::mpsc::channel(1024);
+
+        let node_clone = node.clone();
+        let htlc_stream_seeder_task_handle = spawn(
+            "ldk lightning node incoming htlc stream seeder",
+            async move {
+                loop {
+                    if let Err(e) =
+                        Self::seed_route_htlcs_stream(&node_clone, &htlc_stream_sender).await
```
m1sterc001guy, line 97: `seed_route_htlcs_stream` only fails in one spot and all you do is log....you can push that log statement into the function and get rid of the error handling here

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+#[async_trait]
+impl ILnRpcClient for GatewayLdkClient {
+    async fn info(&self) -> Result<GetNodeInfoResponse, LightningRpcError> {
+        let node_status = self.node.status();
+
+        let chain_tip_block_summary = self
+            .esplora_client
+            .get_blocks(None)
+            .await
+            .unwrap()
```
m1sterc001guy, line 203: Return error instead of `unwrap`

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+            > esplora_chain_tip_timestamp
+            && node_status
+                .latest_onchain_wallet_sync_timestamp
+                .unwrap_or_default()
+                > esplora_chain_tip_timestamp;
+
+        Ok(GetNodeInfoResponse {
+            pub_key: self.node.node_id().serialize().to_vec(),
+            // TODO: This is a placeholder. We need to get the actual alias from the LDK node.
+            alias: "LDK Fedimint Gateway Node".to_string(),
```
m1sterc001guy, line 221: does ldk node not expose this yet?

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+            block_height,
+            synced_to_chain,
+        })
+    }
+
+    async fn routehints(
+        &self,
+        _num_route_hints: usize,
+    ) -> Result<GetRouteHintsResponse, LightningRpcError> {
+        Ok(GetRouteHintsResponse {
```
m1sterc001guy, line 239: Eventually will be good to implement this. You can create the route hints from the channels that the node is connected to. Not a blocker for this

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+                return Err(LightningRpcError::FailedPayment {
+                    failure_reason: format!("LDK payment failed to initialize: {e:?}"),
+                });
+            }
+        };
+
+        // TODO: Find a way to avoid looping/polling to find out when a payment is
+        // completed. `ldk-node` provides `PaymentSuccessful` and `PaymentFailed`
+        // events, but routing them to specific payment IDs isn't straightforward.
+        tokio::time::timeout(Duration::from_secs(30), async {
```
m1sterc001guy, line 267: nit: use `fedimint_core::runtime::timeout`

m1sterc001guy: Can the payment be successful after the 30 second timeout? If so, that could allow an attacker to get paid via lightning and refund their ecash in escrow, stealing money from the gateway

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+            Err(e) => {
+                return Err(LightningRpcError::FailedPayment {
+                    failure_reason: format!("LDK payment failed to initialize: {e:?}"),
+                });
+            }
+        };
+
+        // TODO: Find a way to avoid looping/polling to find out when a payment is
+        // completed. `ldk-node` provides `PaymentSuccessful` and `PaymentFailed`
+        // events, but routing them to specific payment IDs isn't straightforward.
```
m1sterc001guy, line 266: What does "routing them" mean here?

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+        htlc: InterceptHtlcResponse,
+    ) -> Result<EmptyResponse, LightningRpcError> {
+        let InterceptHtlcResponse {
+            action,
+            payment_hash,
+            incoming_chan_id: _,
+            htlc_id: _,
+        } = htlc;
+
+        let ph = PaymentHash(payment_hash.clone().try_into().unwrap());
```
m1sterc001guy, line 326: nit: remove `unwrap`

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+        let invoice = self
+            .node
+            .bolt11_payment()
+            .receive_for_hash(
+                create_invoice_request.amount_msat,
+                if let Some(Description::Direct(description)) = &create_invoice_request.description
+                {
+                    description
+                } else {
+                    ""
```
m1sterc001guy, line 385: does ldk node support setting the description hash in `receive_for_hash`?

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,501 @@
...
+                    None => 0,
+                },
+            });
+        }
+
+        Ok(channels)
+    }
+}
+
+fn create_short_channel_id(block_height: u32, transaction_index: u32, output_index: u32) -> u64 {
```
m1sterc001guy, line 495: We have a very similar function in the CLN extension called `scid_to_u64`. Might be worth to extract somewhere common

tvolk131: I learned that the `lightning` crate exposes an [`scid_from_parts`](https://docs.rs/lightning/0.0.123/lightning/util/scid_utils/fn.scid_from_parts.html) function that does exactly what this function here does, so I removed this one. As for `scid_to_u64`, cln_rpc's `ShortChannelId` struct is represented using a `u64` under the hood, so I made this PR to expose that so we won't need to do any conversion there:

https://github.com/ElementsProject/lightning/pull/7450

m1sterc001guy: Perfect

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: The gateway lightning client can be in two states - either it's inactive (not intercepting HTLCs) or active (intercepting HTLCs), and calling `ILnRpcClient::route_htlcs` spins up the HTLC interception, switching from the former state to the latter. Currently these two states are internally represented in the following way:

**Inactive**: `GatewayLdkClient.htlc_stream_seeder_task_handle` is already running and passing incoming HTLCs to the HTLC stream receiver. The HTLC stream receiver is held in `GatewayLdkClient.htlc_stream_receiver_or` until `route_htlcs` is called, at which point this receiver is taken and returned.

**Active**: `GatewayLdkClient.htlc_stream_seeder_task_handle` is still running. But the HTLC stream receiver has been passed to the caller of `route_htlcs` and so `GatewayLdkClient.htlc_stream_receiver_or` is `None`.

What if we changed `GatewayLdkClient.htlc_stream_seeder_task_handle` to be an `Option<tokio::task::JoinHandle<()>>`, removed `GatewayLdkClient.htlc_stream_receiver_or` altogether, and represented the states like this:

**Inactive**: `GatewayLdkClient.htlc_stream_seeder_task_handle_or` is `None` and `ldk-node` events are not being handled yet. The HTLC stream receiver doesn't exist yet.

**Active**: The HTLC stream sender and receiver are created when `route_htlcs` is called. The sender is passed into the htlc stream seeder task, which is started up and assigned to `GatewayLdkClient.htlc_stream_seeder_task_handle_or`. The receiver is returned by `route_htlcs` and never held by `GatewayLdkClient`.

WDYT?

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Good point... Should we just infinite-loop here and assume that `ldk-node` eventually moves the payment to a non-pending state? (I know this is a pretty gross way of waiting until a payment has been sent, but I believe we need to make some upstream changes to `ldk-node` in order to improve this)

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Yep, it's a payment interceptor. I renamed it to `handle_next_event()` and renamed `seed_route_htlcs_stream` to `event_handler_task_handle` since if we ever want to handle more `ldk-node` event types, we _must_ do that in this function since it skips over events it doesn't care about.

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Excellent point! Done.

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Not as far as I can see

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Sorry, that was a strange wording. What I mean is we should only be handling events from `ldk-node`'s event queue in one place to avoid having two threads stomping/skipping over events. Right now, event handling is done in `GatewayLdkClient::event_handler_task_handle`, so we'd need to add functionality there to handle `PaymentSuccessful` and `PaymentFailed` events and somehow be able to tell `GatewayLdkClient::pay` when a payment has succeeded or failed. This can definitely be done, but we'd need some sort of per-paymenthash queue or something.

I cleaned up the comment here

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: I don't think so. Drilling down into the function calls here, we end up at this:

https://github.com/lightningdevkit/rust-lightning/blob/f48a273211d5235e1ff335b2c25f8aacf3642f24/lightning-invoice/src/utils.rs#L431-L453

Where `_create_invoice_from_channelmanager_and_duration_since_epoch()` accepts a `Bolt11InvoiceDescription‎` (which can be a direct description or a hash) but is hardcoded to always be a direct description.

gateway/ln-gateway/src/lightning/ldk.rs
m1sterc001guy: Perhaps add a random identifier in the alias here.

gateway/ln-gateway/src/lightning/ldk.rs
m1sterc001guy: This would probably be good to make an issue upstream

gateway/ln-gateway/src/lightning/ldk.rs
m1sterc001guy: Yes I think so. I believe we have similar limitations in LND and CLN currently. I would prefer a payment to be stuck for a long time rather than allowing an attacker to steal funds.

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Done. I just appended the node's pubkey.

gateway/ln-gateway/src/lightning/ldk.rs

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Created https://github.com/lightningdevkit/ldk-node/issues/325

dpc (commented):
LGTM, but someone from @fedimint/lightning should approve.

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -238,6 +241,20 @@ pub enum LightningMode {
...
         cln_extension_addr: SafeUrl,
     },
+    #[clap(name = "ldk")]
+    Ldk {
+        /// LDK esplora server URL
+        #[arg(long = "ldk-esplora-server-url", env = FM_LDK_ESPLORA_SERVER_URL)]
+        esplora_server_url: String,
+
+        /// LDK network (defaults to regtest if not provided)
+        #[arg(long = "ldk-network", env = FM_LDK_NETWORK)]
```
m1sterc001guy, line 251: nit: I think you can specify a default using clap here

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,502 @@
...
+                return Err(LightningRpcError::FailedPayment {
+                    failure_reason: format!("LDK payment failed to initialize: {e:?}"),
+                });
+            }
+        };
+
+        // TODO: Find a way to avoid looping/polling to know when a payment is
+        // completed. `ldk-node` provides `PaymentSuccessful` and `PaymentFailed`
+        // events, but interacting with the node event queue here isn't
+        // straightforward.
```
m1sterc001guy, line 273: If you want to implement this, you could potentially use a channel to communicate from the event queue thread to here. Can be done in a follow up

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,502 @@
...
+        let route_htlc_stream = match self.htlc_stream_receiver_or.take() {
+            Some(stream) => Ok(Box::pin(ReceiverStream::new(stream))),
+            None => Err(LightningRpcError::FailedToRouteHtlcs {
+                failure_reason:
+                    "Stream does not exist. Likely was already taken by calling `route_htlcs()`."
+                        .to_string(),
+            }),
+        }?;
+
+        Ok((route_htlc_stream, Arc::new(*self)))
```
m1sterc001guy, line 313: This will move `self` rather than clone it, which is what we do in the other implementations. As far as I can tell this won't cause any issues (whereas with LND we need to clone it). I assume this was intentional?

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,502 @@
...
+            LightningRpcError::FailedToCompleteHtlc {
+                failure_reason: "Failed to parse payment hash".to_string(),
+            }
+        })?);
+
+        // TODO: Get the actual amount from the LDK node. Probably makes the
+        // most sense to pipe it through the `InterceptHtlcResponse` struct.
+        // This value is only used by `ldk-node` to ensure that the amount
+        // claimed isn't less than the amount expected, but we've already
+        // verified that the amount is correct when we intercepted the HTLC.
```
m1sterc001guy, line 337: nit: "intercepted the payment" not HTLC. Once LNv2 is merged we should refactor all HTLC names out of the codebase

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,502 @@
...
+                        claimable_amount_msat,
+                        PaymentPreimage(preimage.try_into().unwrap()),
+                    )
+                    .map_err(|_| LightningRpcError::FailedToCompleteHtlc {
+                        failure_reason: format!(
+                            "Failed to claim LDK payment with hash {ph_hex_str}"
+                        ),
+                    })?;
+            }
+            _ => {
```
m1sterc001guy, line 357: Log an `error!` here, need to have enough info to debug stuff :)

gateway/ln-gateway/src/lightning/ldk.rs
m1sterc001guy: Let's chat about this offline

Kodylow:
Blockers on this, waiting until after 0.4 to merge since we can't get this in this week and need to cut release with the DNS fix. This will be the next big merge.

- Liquidity API and other lightning functionality moved into gateway-cli and gateway api: The only way to interface with the underlying ldk-node is the gateway cli. Normally we use the node's api. For this, we need the gateway-cli to expose api functionality we'd normally just have the node use (open channel, close channel, sweep funds, etc) and add that into the gateway-cli. Instead of only doing this for the ldk node, we're going to implement this generically for all the supported node implementations across the gateway api, which means we can also do lightning side management from the gateway ui in https://github.com/fedimint/ui
  - We already support the liquidity ones for all node impls, we need to add the channel management ones in

- LNv2 requires multi-mint support, but we're dropping that as part of the bolt12 refactor: 0.4 will still use LNv1, but ship with LNv2 binaries that optionally run with an env variable, then for 0.5 both v1 and v2 will run side by side, then 0.6 we retire lnv1.

LNv2 requires an implementation of `pay_private`  for multi-mint payments which requires upstream changes to ldk-node (ldk-node doesn't let us do partial payments of an invoice). 
  - @joschisan : the way to go here is not to require multi-mint payments generally. We have pruned invoices which allow us to do partial payments, which are already supported by LND and CLN, so that made multi-mint payments easy. We intend to drop the pruned-invoices struct because they're causing a bunch of problems elsewhere in the codebase as we prepare for bolt12 support, which will require us to reimplement mutli-mint payments using a different mechanism. So we're going to remove that multi-mint payment as part of LNv2 for now, which removes this blocker.

- LNv2 testing: waiting on ldk to be added to this list of gateway tests):    https://github.com/fedimint/fedimint/pull/5651/files

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Yep! This was intentional. The reason that `ILnRpcClient::route_htlcs` consumes `self` and returns `Arc<Self>`  (IIUC) is so that the type checker can enforce that it cannot be called twice. That should still be enforced here.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -0,0 +1,493 @@
...
+            route_hints: vec![],
+        })
+    }
+
+    // TODO: Respect `max_delay` and `max_fee` parameters.
+    async fn pay(
+        &self,
+        invoice: Bolt11Invoice,
+        _max_delay: u64,
+        _max_fee: Amount,
```
joschisan, line 256: Respecting this parameters is security critical!

joschisan (approved):

PR #5557, opened 2024-07-01
fix: Fix call of std::thread::sleep

benthecarman:
caught this in mutiny, not sure why CI didn't catch this


dpc (approved):

dpc:
> caught this in mutiny, not sure why CI didn't catch this

`.config/semgrep.yaml` ,`ban-direct-sleep-in-tests` rule doesn't list this crate. Can you follow-up with that change as well?

dpc:
The blocking sleep here (async function) is quite bad for other reasons, and I would kind of expect clippy to complain about it...

PR #5559, opened 2024-07-01
chore: rename `recoverytool` dir to `fedimint-recoverytool`

dpc:
To be consisent e.g. with `fedimint-dbtool`


m1sterc001guy (dismissed):
Do you need to change here too? https://github.com/fedimint/fedimint/blob/3d0342eaf51bf18257f5f2ff54ce7bdce6b4fa8a/nix/flakebox.nix#L593

dpc:
> Do you need to change here too?
> 
> https://github.com/fedimint/fedimint/blob/3d0342eaf51bf18257f5f2ff54ce7bdce6b4fa8a/nix/flakebox.nix#L593

Done in a different PR that already landed.

fedimint-recoverytool/Cargo.toml
tvolk131, line 1: Nit: can we change the description here:

https://github.com/fedimint/fedimint/blob/070aa1ac5e6842e8c595d3cab41605875c7f9912/fedimint-recoverytool/Cargo.toml#L6

to something like: `tool for retrieving on-chain funds from a offline federation`

tvolk131 (changes requested):

tvolk131 (approved):

PR #5561, opened 2024-07-02
cleanup: remove redundant module testing

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-tests/tests/tests.rs
+++ b/modules/fedimint-lnv2-tests/tests/tests.rs
@@ -1,15 +1,11 @@
-use std::sync::Arc;
-
-use fedimint_client::ClientHandle;
 use fedimint_core::config::FederationId;
-use fedimint_core::core::OperationId;
 use fedimint_core::util::NextOrPending;
 use fedimint_core::{sats, Amount};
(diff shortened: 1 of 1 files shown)


modules/fedimint-lnv2-tests/tests/tests.rs
```diff
@@ -161,158 +136,3 @@ async fn refund_unpayable_invoice() -> anyhow::Result<()> {
...
-    let mut sub = client
-        .get_first_module::<LightningClientModule>()
-        .subscribe_send(send_op_refunded)
-        .await?
-        .into_stream();
-
-    assert_eq!(sub.ok().await?, SendState::Funding);
-    assert_eq!(sub.ok().await?, SendState::Funded);
-    assert_eq!(sub.ok().await?, SendState::Refunding);
-    assert_eq!(sub.ok().await?, SendState::Refunded);
```
m1sterc001guy, line 204: Do we need to delete these? These aren't verifying the same things as devimint tests. Devimint tests only verify E2E functionality, this actually verifies each state transition.

m1sterc001guy: Chatted offline, other tests cover these transitions already

m1sterc001guy (approved):

PR #5563, opened 2024-07-02
chore: test unilateral refund of outgoing contracts

joschisan:


modules/fedimint-lnv2-server/src/lib.rs
```diff
@@ -671,7 +671,7 @@ impl Lightning {
     ) -> Option<ContractId> {
         loop {
             timeout(
-                Duration::from_secs(60 * 10),
+                Duration::from_secs(10),
```
m1sterc001guy, line 674: Was this necessary or did you just want to increase the polling rate?

m1sterc001guy (approved):

PR #5565, opened 2024-07-02
chore: test payment with two gateways

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-tests/src/bin/self_payment.rs
+++ b/modules/fedimint-lnv2-tests/src/bin/self_payment.rs
@@ -30,6 +30,8 @@ async fn main() -> anyhow::Result<()> {
 
         let fed = dev_fed.fed().await?;
 
+        // test self payment with one gateway
+
         let gw_lnd = dev_fed.gw_lnd().await?;
 
         let client = fed.new_joined_client("lnv2-module-client").await?;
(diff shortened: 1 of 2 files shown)


modules/fedimint-lnv2-tests/src/bin/self_payment.rs
```diff
@@ -90,6 +92,64 @@ async fn main() -> anyhow::Result<()> {
...
+            .await?,
+        )?;
+
+        assert_eq!(
+            cmd!(
+                client,
+                "module",
+                "lnv2",
+                "await-send",
+                serde_json::to_string(&send_op)?.substring(1, 65)
```
m1sterc001guy, line 133: Why do you need to substring the `OperationId`?

modules/fedimint-lnv2-tests/src/bin/self_payment.rs
```diff
@@ -90,6 +92,64 @@ async fn main() -> anyhow::Result<()> {
             serde_json::to_value(FinalReceiveState::Claimed).expect("JSON serialization failed"),
         );
 
+        // test self payment with two gateways
+
+        let gw_cln = dev_fed.gw_cln().await?;
```
m1sterc001guy, line 97: Going forward we should consider writing these tests agnostic of the lightning node type. We'll need this for LDK as well

m1sterc001guy (approved):

modules/fedimint-lnv2-tests/src/bin/self_payment.rs
joschisan: serde json adds quotes we need to remove

PR #5573, opened 2024-07-05
chore: core cleanup

tvolk131:
Document some core structs, and split them into separate files

### DIFF
--- a/fedimint-core/src/amount.rs
+++ b/fedimint-core/src/amount.rs
@@ -1,0 +1,295 @@
+use std::num::ParseIntError;
+use std::str::FromStr;
+
+use anyhow::bail;
+use bitcoin::Denomination;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::encoding::{Decodable, Encodable};
+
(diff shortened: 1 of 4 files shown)


joschisan (approved):

fedimint-core/src/amount.rs
```diff
@@ -0,0 +1,295 @@
...
+    Hash,
+    Deserialize,
+    Serialize,
+    Encodable,
+    Decodable,
+)]
+#[serde(transparent)]
+pub struct Amount {
+    pub msats: u64,
+}
```
oleonardolima, line 42: food for thought: maybe in the future we could rely on [rust-bitcoin Amount](https://docs.rs/bitcoin/latest/bitcoin/struct.Amount.html) instead 🤔

oleonardolima (approved):
utACK 621ce6c3f5c053c6746750935dc791ca9231f96f

fedimint-core/src/amount.rs
oleonardolima: oh, I should've read the docstring first lol, forgot about the msats issue

PR #1412, opened 2023-01-19
Add our own hex::serde implementation

mxz42:
part of #1307 

the serialize and deserialize functions are specific to `Vec<u8>` but I think that works so far. 

thoughts? is it ok like this or should be more generic? 
@dpc @elsirion

PR #5574, opened 2024-07-05
chore: remove fedimint-core hex serde

tvolk131:
This was added in #1412 in an effort to remove the `hex` crate, but we now use this crate throughout the codebase, so this doesn't serve much purpose anymore.


oleonardolima (approved):
ACK ba159f81f2384489c06d6e211968a4642f779524

dpc (approved):

PR #5578, opened 2024-07-08
chore: portalloc cleanup

tvolk131:


utils/portalloc/src/data/dto.rs
```diff
@@ -32,14 +36,63 @@ impl Default for RootData {
...
+                            port, "Could not use a port. Will try a different range"
+                        );
+                        base_port = port + 1;
+                        continue 'retry;
+                    }
+                    Ok(l) => l,
+                };
+            }
+
+            const ALLOCATION_TIME_SECS: u64 = 120;
```
dpc, line 92: If you're refactoring it (which is a good idea), I would move the stuff from this line downwards to a separate function, so the logic here can be one larger loop-y thing on immutable `&self` and then mutations live outside. Unless I'm missing something here that makes it necessary to remain mutable.

utils/portalloc/src/data/dto.rs
```diff
@@ -37,6 +43,8 @@ impl Default for RootData {
 
 impl RootData {
     pub fn get_free_port_range(&mut self, range_size: u16) -> u16 {
+        self.reclaim(now_ts());
```
dpc: This now will not work with `&self`, but possibly can be moved out too.

dpc (commented):
Comment, but LGTM

dpc:
Would be good to wrap up and land. :) @tvolk131 DM for a snappy response when ready.

tvolk131:
Sorry, been swamped the past few days - will get to this tonight!

utils/portalloc/src/data/dto.rs
tvolk131: I moved this code into `RootData::insert`. But the loop here calls `RootData::reclaim` which requires `&mut self`.

utils/portalloc/src/data/dto.rs
tvolk131: I'm not sure what you mean here - can you clarify?

dpc (approved):

PR #5579, opened 2024-07-08
chore: use begin_transaction_nc where possible

tvolk131:
This makes it easier to see at a glance whether a database interaction is writing or just reading

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -797,7 +797,7 @@ impl Client {
     }
 
     pub async fn get_config_from_db(db: &Database) -> Option<ClientConfig> {
-        let mut dbtx = db.begin_transaction().await;
+        let mut dbtx = db.begin_transaction_nc().await;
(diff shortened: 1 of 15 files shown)


m1sterc001guy (approved):
Shouldn't compile if it broke anything, so LGTM :)

dpc:
@tvolk131 @m1sterc001guy For the record, technical correction: the "NC" (non-commitable) means that the code with such a reference can't commit the tx by itself. It still can call `insert_entry`, etc. but it just won't be able to commit by itself (because commit etc. are only impl for `<Commitable>` https://github.com/fedimint/fedimint/blob/93cc8598a264e8a55c814c923a328b9fc3883369/fedimint-core/src/db/mod.rs#L1543

m1sterc001guy:
Yeah, I meant if we had any db transactions that we changed to begin with `begin_transaction_nc` but actually did commit, that would cause a compilation error.

PR #5580, opened 2024-07-08
chore: FederationPeer::request retries with backoff

### LINKED ISSUE #5489: add backoff to FederationPeer::request
https://github.com/fedimint/fedimint/blob/22663ff31d80f0614bfea8ef5803508744ea2ae3/fedimint-api-client/src/api.rs#L1424

it is hammering the browser and crashed the app.

tvolk131:
Fixes #5489

### DIFF
--- a/fedimint-api-client/src/api/mod.rs
+++ b/fedimint-api-client/src/api/mod.rs
@@ -27,6 +27,7 @@ use fedimint_core::module::{ApiAuth, ApiRequestErased, ApiVersion, SerdeModuleEn
 use fedimint_core::session_outcome::{SessionOutcome, SessionStatus};
 use fedimint_core::task::{MaybeSend, MaybeSync};
 use fedimint_core::transaction::{Transaction, TransactionSubmissionOutcome};
+use fedimint_core::util::backon::{self, BackoffBuilder};
 use fedimint_core::util::SafeUrl;
 use fedimint_core::{
     apply, async_trait_maybe_send, dyn_newtype_define, runtime, NumPeersExt, OutPoint, PeerId,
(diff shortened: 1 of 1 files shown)


fedimint-api-client/src/api/mod.rs
```diff
@@ -940,7 +941,12 @@ where
...
     #[instrument(level = "trace", fields(peer = %self.peer_id, %method), skip_all)]
     pub async fn request(&self, method: &str, params: &[Value]) -> JsonRpcResult<Value> {
-        for attempts in 0.. {
+        for (attempts, retry_delay) in backon::FibonacciBuilder::default()
+            .with_min_delay(Duration::from_millis(200))
+            .with_max_delay(Duration::from_secs(5))
+            .build()
+            .enumerate()
+        {
             debug_assert!(attempts <= 1);
```
maan2003, line 950: no, this loop doesn't really loop

there should be some backoff in reconnect.

dpc: image

tvolk131: > at most it will do one more round to re-try after reconnecting once

In optimized builds, the `debug_assert` is removed. Just want to be sure that isn't what's causing this issue. @maan2003 the hammering issue occurs on a dev build, right?

Apologies if this is off-track, I'm still trying to understand what this code is doing lol

elsirion: @maan2003 @dpc I remember us also stumbling over this loop, we should think about a more understandable version of it

dpc: I actually want this loop to actually loop one day. BTW. IIRC we did have strategies that relied on it returning after first disconnect, but it seemed to me somewhat misguided and artificial. 

This loop looping or not has nothing to do with backoffs anyway. The backoff itself is faciliated by the actual underlying client being wrapped in `Jit` that has sleep in it, or something like that.

elsirion:
Call: we can likely close this

PR #5595, opened 2024-07-10
fedimint wasm client

maan2003:
on top #5822  

the actual wasm client


fedimint-client-wasm/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+async-stream = "0.3.5"
+fedimint-core.workspace = true
+fedimint-ln-client = { workspace = true, features = ["cli"] }
+fedimint-mint-client = { version = "=0.5.0-alpha", path = "../modules/fedimint-mint-client" }
+fedimint-client = { workspace = true }
+wasm-bindgen = "=0.2.92"                                                                      # must match the nix provided wasm-bindgen-cli version
+wasm-bindgen-futures = "0.4.42"
+wasm-bindgen-test = "0.3.42"
+fedimint-api-client.workspace = true
+serde_json.workspace = true
```
dpc, line 20: :exploding_head:

maan2003: you didn't know about this toml syntax?

maan2003: oh this broke deps sorting :smile:

dpc (approved):

dpc (dismissed):

fedimint-client-wasm/Cargo.toml
dpc: Just because I know about the syntax, doesn't mean I'm clever enough to succinctly use it every time. :D

dpc (approved):

fedimint-client-wasm/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+                    let mut stream = mint.handle_rpc(method.to_owned(), payload).await;
+                    while let Some(item) = stream.next().await {
+                        yield item?;
+                    }
+                }
+                _ => {
+                    Err(anyhow::format_err!("module not found: {module}"))?;
+                    unreachable!()
+                },
+            }
```
elsirion, line 131: Does this need to be matched explicitly? Shouldn't every module know its kind and we can just search for the `DynClientModule` with the correct kind and then call `handle_rpc` on it?

That way it would be much easier to make the `fedimint-client-wasm` crate instantiable with different module sets like all the other top-level crates.

maan2003: :+1: will do in follow up

PR #5596, opened 2024-07-10
refactor: refactor mint modular cli

joschisan:


modules/fedimint-mint-client/src/lib.rs
```diff
@@ -11,6 +11,8 @@
 
 // Backup and restore logic
 pub mod backup;
+/// Modularized Cli for sending and receiving out-of-band ecash
+mod cli;
```
dpc, line 15: We probably want to put it behind a feature flag, like in other modules. But can be done separately.

dpc (approved):

Issue #5581, opened 2024-07-08
HTLC Interception with LND doesn't work with hodl invoices


Kodylow:
per @m1sterc001guy testing v2 last week

PR #5597, opened 2024-07-10
feat: LNv2 LND Subscribe Invoices

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/5581

With LNv2 protocol, LND's HTLC interception mechanism no longer works. I believe this is because the HTLC interception only triggers for HTLCs that are being forwarded. In the LNv2 protocol, the gateway now generates the invoice, so the HTLCs are no longer needing to be forwarded, they are destined for the gateway's node. Therefore, in LND we need a new mechanism for detecting when a payment has been made and we can request the preimage from the federation.

To do this, we now use HOLD invoices and the `subscribe_invoices` and `subscribe_single_invoice` RPCs. Unfortunately, `subscribe_invoices` does not emit updates for HOLD invoices: https://github.com/lightningnetwork/lnd/issues/3120. To get around this limitation, we use `subscribe_invoices` to detect when a new invoice has been created, then we spawn a new task to actually monitor the status of the HOLD invoice using `subscribe_single_invoice`.

The devimint tests have also been refactored/cleaned up a bit to allow easier paying/creation of invoices.


scripts/tests/lnv2-module-test.sh
```diff
@@ -10,3 +10,5 @@ add_target_dir_to_path
 
 gateway_registration_sanity
 self_payment
+lightning_payment --gw-type lnd
```
m1sterc001guy, line 13: This can be split into a different script and run in parallel in CI if we think this will make the test too slow

m1sterc001guy: no longer necessary

gateway/ln-gateway/src/lib.rs
```diff
@@ -403,6 +403,7 @@ impl Gateway {
         Gateway::new(
             Arc::new(GatewayLightningBuilder {
                 lightning_mode: opts.mode,
+                gateway_db: gateway_db.clone(),
```
m1sterc001guy, line 406: @tvolk131 this unfortunately might break some of your proposed refactors/abstractions

joschisan (approved):

gateway/ln-gateway/src/lib.rs
tvolk131: So be it

PR #5600, opened 2024-07-11
fix: check that incoming payment is economical

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-client/src/lib.rs
+++ b/modules/fedimint-lnv2-client/src/lib.rs
@@ -669,6 +669,10 @@ impl LightningClientModule {
 
         let contract_amount = payment_info.receive_fee.subtract_fee(invoice_amount.msats);
 
+        if contract_amount < Amount::from_sats(100) {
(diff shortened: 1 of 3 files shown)


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -669,6 +669,10 @@ impl LightningClientModule {
 
         let contract_amount = payment_info.receive_fee.subtract_fee(invoice_amount.msats);
 
+        if contract_amount < Amount::from_sats(100) {
```
bradleystachurski, line 672: I don't have much context here, but how is 100 sats determined to be the economical threshold? If this is a magic number it might be useful to define as a const with docs.

bradleystachurski (approved):

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: +1 we should add a comment so we don't forget why 100 sats was chosen 3 years from now :)

PR #5605, opened 2024-07-11
chore: require acking v0.4 upgrade instructions via env var

### LINKED ISSUE #5603: Check that shutdown happened in lockstep
We could have a env var `FM_I_READ_THE_RELEASE_LOG` that can be set to `0.4.0` and we can check whenever we require lockstep upgrades.

dpc:
This will require all end users to acknowledge that they are aware of upgrading in the lockstep requirement.

Fix #5603

### DIFF
--- a/fedimintd/src/fedimintd.rs
+++ b/fedimintd/src/fedimintd.rs
@@ -5,7 +5,7 @@ use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::time::Duration;
 
-use anyhow::{format_err, Context};
+use anyhow::{bail, format_err, Context};
 use clap::{Parser, Subcommand};
 use fedimint_core::admin_client::ConfigGenParamsRequest;
 use fedimint_core::config::{
@@ -39,7 +39,7 @@ use fedimint_wallet_server::common::config::{
 };
 use fedimint_wallet_server::WalletInit;
 use futures::FutureExt;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 
 use crate::default_esplora_server;
 use crate::envs::{
@@ -215,6 +215,7 @@ impl Fedimintd {
         );
 
         handle_version_hash_command(code_version_hash);
+        check_release_notes_ack()?;
 
         let fedimint_version = env!("CARGO_PKG_VERSION");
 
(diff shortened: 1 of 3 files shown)


joschisan (approved):

dpc:
@douglaz Do you think this is OK w.r.t. to infra automation etc.? I think it shouldn't be a problem, but would love a second opinion.

douglaz:
Looks ok to me. But for people running on third party services like Clovyr it will be pretty much impossible to safely upgrade, I guess.

dpc:
Ping @Kodylow for opinion.

dpc:
> But for people running on third party services like Clovyr it will be pretty much impossible to safely upgrade, I guess.

@douglaz You mean due to the lockstep upgrade requirement, right? Or is setting env var an extra problem?

fedimintd/src/fedimintd.rs
```diff
@@ -464,6 +465,37 @@ impl Fedimintd {
...
+    // The actual code-path we want end users to take
+    if std::env::var(FM_REL_NOTES_ACK_ENV).unwrap_or_default() == FM_REL_NOTES_CUR_VAL {
+        return Ok(());
+    }
+
+    eprintln!(
+        "This version of fedimintd has some criticallly important requirements you should be aware of."
+    );
+    eprintln!("Not following them will likely result in a consensus failure or other problems.");
+    eprintln!("To ensure you have read them, you must set specific environment variable to a specific value.");
```
Kodylow, line 492: change to "Read release notes for instructions on how to perform the upgrade"

Kodylow:
ACK. yes this doesn't work with clovyr, no one is running mainnet clovyrs besides me though so that's fine.

fedimintd/src/fedimintd.rs
dpc: @Kodylow What's the reasoning behind the change.

One thing I like to highlight is setting this env (and thus this message) is displayed irrespective of the fact of setting up or upgrading the Federation. So "how to perform the upgrade" might be confusing for the user that is not doing any upgrade.

fedimintd/src/fedimintd.rs
```diff
@@ -464,6 +465,37 @@ impl Fedimintd {
     }
 }
 
+fn check_release_notes_ack() -> anyhow::Result<()> {
+    const VERSION: &str = "v0.4";
+    const FM_SKIP_REL_NOTES_ACK_ENV: &str = "FM_SKIP_REL_NOTES_ACK";
+    const FM_REL_NOTES_ACK_ENV: &str = "FM_REL_NOTES_ACK";
+    // the suffix here is to prevent people getting to smart and trying to
```
m1sterc001guy, line 472: nit: "too" :)

m1sterc001guy (approved):
LGTM

fedimint-backports:
- #5616

PR #5610, opened 2024-07-12
chore: split up htlc handling logic

tvolk131:


gateway/ln-gateway/src/lib.rs
```diff
@@ -198,6 +198,12 @@ pub struct LightningContext {
 // A marker struct, to distinguish lock over `Gateway::clients`.
 struct ClientsJoinLock;
 
+enum HandleStreamResult {
```
m1sterc001guy, line 201: Let's add a doc comment on this enum.

Also, I think we need a different name, I don't think its descriptive enough. `Result` is a bit weird here too, considering its not a normal `Result`. Perhaps `ReceivePaymentStreamAction`?

gateway/ln-gateway/src/lib.rs
```diff
@@ -419,113 +425,148 @@ impl Gateway {
...
-                            }
-                            Err(e) => {
-                                warn!("Failed to retrieve Lightning info: {e:?}");
-                            }
+                    debug!("Establishing HTLC stream...");
+                    let (stream, ln_client) = match lnrpc_route.route_htlcs(&htlc_task_group).await
+                    {
+                        Ok((stream, ln_client)) => (stream, ln_client),
+                        Err(e) => {
+                            warn!("Failed to open HTLC stream: {e:?}");
```
m1sterc001guy, line 450: nit: Can avoid string interpolation `warn!(?e, "Failed to open HTLC stream");`

gateway/ln-gateway/src/lib.rs
```diff
@@ -419,113 +425,148 @@ impl Gateway {
...
+                            continue
                         }
+                    };
+
+                    // Successful calls to route_htlcs establish a connection
+                    self_copy.set_gateway_state(GatewayState::Connected).await;
+                    info!("Established HTLC stream");
+
+                    let route_htlcs_response =
+                        self_copy.route_htlcs(&handle, stream, ln_client).await;
```
m1sterc001guy, line 460: Let's take this opportunity to improve on our naming to support LNv2. `route_htlcs` is a bad name since its the same as the function above. Can we call this `handle_payment_requests` and rename `route_htlcs` to `create_payment_stream`? (open to suggestions on naming, but in general we need to move away from using "HTLC")

gateway/ln-gateway/src/lib.rs
```diff
@@ -419,113 +425,148 @@ impl Gateway {
...
+                            sleep(Duration::from_secs(HTLC_STREAM_RETRY_SECONDS)).await;
+                        }
+                        HandleStreamResult::NoRetry => break,
                     }
                 }
+            },
+        );
+    }
+
+    async fn route_htlcs<'a>(
```
m1sterc001guy, line 480: Add function documentation

gateway/ln-gateway/src/lib.rs
```diff
@@ -419,113 +425,148 @@ impl Gateway {
...
+        }
+
+        let lightning_context = LightningContext {
+            lnrpc: ln_client,
+            lightning_public_key,
+            lightning_alias,
+            lightning_network,
+        };
+        self.set_gateway_state(GatewayState::Running { lightning_context })
+            .await;
```
m1sterc001guy, line 539: Can we add a `info!` here so we know when the gateway is running?

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Although, maybe we want to merge this first so the diff isn't impossible to read

gateway/ln-gateway/src/lib.rs
```diff
@@ -537,105 +578,149 @@ impl Gateway {
...
+        Self::forward_htlc(htlc_request, lightning_context).await;
+    }
+
+    /// Tries to handle an HTLC using the LNv2 protocol.
+    /// Returns `true` if the HTLC was handled, `false` otherwise.
+    async fn try_handle_htlc_lnv2(
+        &self,
+        htlc_request: &InterceptHtlcRequest,
+        lightning_context: &LightningContext,
+    ) -> bool {
```
m1sterc001guy, line 619: Can we use a `Result` instead of returning a boolean? I think that would be cleaner

gateway/ln-gateway/src/lib.rs
```diff
@@ -537,105 +578,149 @@ impl Gateway {
...
-                                    }
+            return true;
+        }
+
+        false
+    }
+
+    /// Tries to handle an HTLC using the legacy lightning protocol.
+    /// Returns `true` if the HTLC was handled, `false` otherwise.
+    async fn try_handle_htlc_ln_legacy(&self, htlc_request: &InterceptHtlcRequest) -> bool {
```
m1sterc001guy, line 669: Same comment as above, I think we can use a `Result` here

gateway/ln-gateway/src/lib.rs
```diff
@@ -537,105 +578,149 @@ impl Gateway {
...
+            .await
+        {
+            return;
+        }
+
+        if self.try_handle_htlc_ln_legacy(&htlc_request).await {
+            return;
+        }
+
+        Self::forward_htlc(htlc_request, lightning_context).await;
```
m1sterc001guy, line 610: I think `forward_htlc` can just be called from inside `try_handle_htlc_ln_legacy` if its not able to handle it

m1sterc001guy:
I like these changes, they make the gateway logic simpler and when we don't support v1 anymore, it will be easy to rip out the old code. Most suggestions are small naming changes or nits.

I think after this PR we should do a pass to remove "HTLC" remove a term that we use

gateway/ln-gateway/src/lib.rs
tvolk131: Yeah, I think it makes sense to handle this in a separate PR. Let's keep the "htlc" naming for now for consistency.

gateway/ln-gateway/src/lib.rs
tvolk131: I think that's a potential foot-gun. Ideally, the order of `try_handle_htlc_lnv2()` and `try_handle_htlc_ln_legacy()` could be swapped without any issue (since no payment should have a fake SCID last hop and _also_ have its payment hash registered with the gateway's lightning node). But calling `forward_htlc()` within `try_handle_htlc_ln_legacy()` would break that. Swapping the order of those functions would effectively shut off any attempts to handle incoming LNv2 payments, but that wouldn't be obvious to someone just looking at `Gateway::handle_htlc()`.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Fair enough, I'm good with keeping it.

When we remove LNv1 support, we should also remove the concept of forwarding, since it no longer makes sense in LNv2, which was my original thought for why they should be coupled.

m1sterc001guy (approved):

PR #5614, opened 2024-07-12
chore: consolidate usage of backoffs

tvolk131:
Limit the `backon` types that are exported from `fedimint-core` and export a few standard backoff functions:

`background_backoff()`, `aggressive_backoff()`, `immediate_backoff()`, `custom_backoff()`

### DIFF
--- a/devimint/src/gatewayd.rs
+++ b/devimint/src/gatewayd.rs
@@ -1,10 +1,9 @@
 use std::collections::HashMap;
 use std::ops::ControlFlow;
 use std::path::PathBuf;
-use std::time::Duration;
 
 use anyhow::{Context, Result};
-use fedimint_core::util::{backon, retry};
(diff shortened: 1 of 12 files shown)


fedimint-core/src/util/backon.rs
```diff
@@ -0,0 +1,44 @@
+use std::time::Duration;
```
dpc, line 1: I guess this is OK, but I fear this module having the exact same name as 3rd party crate is going to be a source of confusion.

Time will tell I guess.

dpc (dismissed):

fedimint-core/src/util/backon.rs
tvolk131: Renamed to `backoff_util`

dpc (approved):

PR #5620, opened 2024-07-12
fix(devimint): run_ui does not block

dpc:
The intention was for `devimint run-ui` to spawn stuff and block waiting for Ctrl+C. This got lost somewhere during refactorings of devimint and just now discovered.

I think the `clean_on_exit` is more versatile if it does not wait itself, and require logic that it wraps to block if that's the intention.


bradleystachurski (approved):

fedimint-backports:
- #5621

PR #5627, opened 2024-07-13
feat: prepare ptlc support

joschisan:


m1sterc001guy (approved):
Looks fine to me, maybe @elsirion has more thoughts

elsirion (dismissed):
LGTM, just not sure if including the current logic is even worth it. Having the enum in place certainly is though. Should it maybe even be extensible for future proofing?

elsirion (approved):

PR #5630, opened 2024-07-13
feat: shutdown after certain session index for a coordinated consensu…

elsirion:
…s upgrade

### DIFF
--- a/fedimint-server/src/net/api.rs
+++ b/fedimint-server/src/net/api.rs
@@ -27,7 +27,8 @@ use fedimint_core::endpoint_constants::{
     CLIENT_CONFIG_ENDPOINT, FEDERATION_ID_ENDPOINT, GUARDIAN_CONFIG_BACKUP_ENDPOINT,
     INVITE_CODE_ENDPOINT, MODULES_CONFIG_JSON_ENDPOINT, RECOVER_ENDPOINT,
     SERVER_CONFIG_CONSENSUS_HASH_ENDPOINT, SESSION_COUNT_ENDPOINT, SESSION_STATUS_ENDPOINT,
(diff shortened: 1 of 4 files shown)


fedimint-server/src/net/api.rs
```diff
@@ -547,6 +553,15 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
                 Ok((&fedimint.session_status(index).await).into())
             }
         },
+        api_endpoint! {
+            SHUTDOWN_ENDPOINT,
+            ApiVersion::new(0, 0),
```
dpc, line 558: This ideally should be a version the call was introduced in, and I think we're at `0.2`, and should increment to 0.3. https://github.com/fedimint/fedimint/blob/bffe4660774db857ab5672f654e6eae96522246b/fedimint-server/src/config/mod.rs#L211

dpc: Thought being a quick EoL fix for fedimint 0.3 branch, it all probably doesn't matter.

bradleystachurski: If we update here, should we also update master?

dpc: I think so, yes. For consistency at least. I doubt it makes a difference in practice, and mostly pointing it out so in the future we don't forget.

bradleystachurski:
Closing this PR so we can bump the api version while backporting in https://github.com/fedimint/fedimint/pull/5642

fedimint-server/src/net/api.rs
bradleystachurski: Updating master in https://github.com/fedimint/fedimint/pull/5640

PR #5551, opened 2024-06-30
chore: move gateway db functionality into new DbManager struct

tvolk131:
Requires #5548

PR #5633, opened 2024-07-14
chore: move gateway db code into extension trait

tvolk131:
Replaces #5551

Add a trait containing gateway db functionality, and implement it on the `Database` type. Ideally we could just add an `impl` block to `Database` that is only usable in the gateway codebase, but Rust doesn't allow `impl` blocks for structs defined in another crate.

Also de-indent `Gateway::register_federations`

### DIFF
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -1,17 +1,20 @@
 use std::collections::BTreeMap;
 
+use anyhow::{anyhow, bail};
 use bitcoin::Network;
-use bitcoin_hashes::sha256;
+use bitcoin_hashes::{sha256, Hash};
 use fedimint_core::config::FederationId;
 use fedimint_core::db::{
-    CoreMigrationFn, DatabaseTransaction, DatabaseVersion, IDatabaseTransactionOpsCoreTyped,
+    CoreMigrationFn, Database, DatabaseTransaction, DatabaseVersion,
+    IDatabaseTransactionOpsCoreTyped,
 };
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::invite_code::InviteCode;
 use fedimint_core::{impl_db_lookup, impl_db_record, secp256k1};
 use fedimint_ln_common::serde_routing_fees;
+use fedimint_lnv2_client::CreateBolt11InvoicePayload;
 use fedimint_lnv2_common::contracts::IncomingContract;
-use futures::FutureExt;
+use futures::{FutureExt, StreamExt};
 use lightning_invoice::RoutingFees;
 use rand::Rng;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


gateway/ln-gateway/src/db.rs
```diff
@@ -21,6 +24,88 @@ use crate::rpc::rpc_server::hash_password;
...
+        &self,
+        payload: CreateBolt11InvoicePayload,
+    ) -> anyhow::Result<()>;
+    async fn load_lnv2_incoming_contract(
+        &self,
+        payment_hash: [u8; 32],
+    ) -> anyhow::Result<RegisteredIncomingContract>;
+}
+
+impl GatewayDatabaseExt for Database {
```
dpc, line 41: I wonder if this wouldn't be better implemented on `DatabaseTransaction`, so that the caller is free to commit the way it wants.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1149,19 +1147,67 @@ impl Gateway {
     /// `per_federation_routing_fees` is changed, the Gateway will only
     /// re-register with the specified federation.
     pub async fn handle_set_configuration_msg(
+        &self,
+        payload: SetConfigurationPayload,
+    ) -> Result<()> {
+        let mut dbtx = self.gateway_db.begin_transaction().await;
+
+        let new_gateway_config = self.create_new_gateway_config(&payload).await?;
+        dbtx.insert_entry(&GatewayConfigurationKey, &new_gateway_config)
```
dpc, line 1156: Then this could be `dbtx.store_gateway_config` etc.

dpc (commented):
One thought, looks like an improvement to me, and maybe we should be using this pattern in more places. Quite often our functions have low level `insert_entry` etc. in between higher level logic, and this would make it cleaner.


One thing to pay attention during refactoring here is notion of atomicity and retries on write conflicts. However verifing it requires understanding of gw data model, which I'm not very versed on. In other word: "LGTM, but get a second opinion".

gateway/ln-gateway/src/db.rs
m1sterc001guy: +1 to this suggestion

gateway/ln-gateway/src/lib.rs
```diff
@@ -1220,47 +1267,8 @@ impl Gateway {
...
-        }
-
-        dbtx.commit_tx().await;
-
-        let mut curr_gateway_config = self.gateway_config.write().await;
-        *curr_gateway_config = Some(new_gateway_config.clone());
-
-        info!("Set GatewayConfiguration successfully.");
-
-        Ok(())
```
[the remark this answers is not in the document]
m1sterc001guy: Oh I see, it was moved above

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Needs to be the same `dbtx` though that gets committed at the bottom, so you previous comment is necessary for this

m1sterc001guy:
> However verifing it requires understanding of gw data model, which I'm not very versed on.

This will be easier to verify when the extension trait is on the database transaction

tvolk131:
Closing in favor of #5704

PR #5634, opened 2024-07-14
chore: split gateway cli

tvolk131:
With the addition of the lightning liquidity API, the current structure of the gateway CLI is becoming a bit unwieldy. In this PR, we're splitting the CLI commands into two groups: lightning commands (i.e. the liquidity API) and general commands. Each lives in its own file, along with its handling logic. The `General(GeneralCommands)` enum variant of `Commands` is annotated with `#[command(flatten)]` which prevents any behavior change (so, for example, `gateway-cli info` is not changed to `gateway-cli general info`).

### DIFF
--- a/gateway/cli/src/general_commands.rs
+++ b/gateway/cli/src/general_commands.rs
@@ -1,0 +1,224 @@
+use anyhow::bail;
+use bitcoin::address::NetworkUnchecked;
+use bitcoin::Address;
+use clap::Subcommand;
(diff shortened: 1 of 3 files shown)


m1sterc001guy:
Looks like this can be rebased, will be easier to review

tvolk131:
> Looks like this can be rebased, will be easier to review

Rebased!

dpc (approved):

PR #5635, opened 2024-07-15
fix: set base fee to ensure economical payments for the gateway

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-client/src/lib.rs
+++ b/modules/fedimint-lnv2-client/src/lib.rs
@@ -192,20 +192,18 @@ pub struct PaymentFee {
     pub parts_per_million: u64,
 }
 
-impl PaymentFee {
-    pub fn one_percent() -> Self {
-        PaymentFee {
-            base: Amount::ZERO,
-            parts_per_million: 10_000,
-        }
-    }
+const DUST_LIMIT: Amount = Amount::from_sats(50);
 
-    pub fn half_of_one_percent() -> Self {
-        PaymentFee {
-            base: Amount::ZERO,
-            parts_per_million: 5_000,
-        }
-    }
+impl PaymentFee {
+    pub const TEN_PROMILLE_PLUS_50_SATS: PaymentFee = PaymentFee {
+        base: DUST_LIMIT,
+        parts_per_million: 10_000,
+    };
+
+    pub const FIVE_PROMILLE_PLUS_50_SATS: PaymentFee = PaymentFee {
+        base: DUST_LIMIT,
+        parts_per_million: 5_000,
+    };
 
     pub fn add_fee(&self, msats: u64) -> Amount {
         Amount::from_msats(msats.saturating_add(self.fee(msats)))
@@ -354,7 +352,7 @@ impl LightningClientModule {
         self.send_internal(
             gateway_api,
             invoice,
-            PaymentFee::one_percent(),
+            PaymentFee::TEN_PROMILLE_PLUS_50_SATS,
             EXPIRATION_DELTA_LIMIT_DEFAULT,
             None,
         )
@@ -593,7 +591,7 @@ impl LightningClientModule {
             invoice_amount,
             INVOICE_EXPIRATION_SECONDS_DEFAULT,
             Bolt11InvoiceDescription::Direct(String::new()),
-            PaymentFee::one_percent(),
+            PaymentFee::TEN_PROMILLE_PLUS_50_SATS,
         )
         .await
     }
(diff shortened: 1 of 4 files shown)


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -195,14 +195,14 @@ pub struct PaymentFee {
 impl PaymentFee {
     pub fn one_percent() -> Self {
         PaymentFee {
-            base: Amount::ZERO,
+            base: Amount::from_sats(50),
```
m1sterc001guy, line 198: Should this be the same constant as in here? https://github.com/fedimint/fedimint/pull/5600/files

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -195,14 +195,14 @@ pub struct PaymentFee {
...
             parts_per_million: 10_000,
         }
     }
 
     pub fn half_of_one_percent() -> Self {
         PaymentFee {
-            base: Amount::ZERO,
```
m1sterc001guy, line 205: I think we should change the names of the above functions since they are no longer percentages

modules/fedimint-lnv2-client/src/lib.rs
joschisan: No

modules/fedimint-lnv2-client/src/lib.rs
joschisan: Sure

joschisan: Done

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: Why not? What is the difference between the two? I thought they were both a minimal for claiming an input

joschisan: The one you linked is only supposed to check that subtracting the fee via saturating_sub did in fact not saturate such that we got the full base fee subtracted.

m1sterc001guy: Sorry, I still don't quite understand the relationship between these numbers. i.e if this number was > 100 sats, would that cause problems?

modules/fedimint-lnv2-tests/tests/mock.rs
```diff
@@ -89,9 +89,9 @@ impl GatewayConnection for MockGatewayConnection {
     ) -> Result<Option<RoutingInfo>, GatewayError> {
         Ok(Some(RoutingInfo {
             public_key: self.keypair.public_key(),
-            send_fee_default: PaymentFee::one_percent(),
-            send_fee_minimum: PaymentFee::half_of_one_percent(),
-            receive_fee: PaymentFee::half_of_one_percent(),
+            send_fee_default: PaymentFee::ten_promille_plus_50_sats(),
```
dpc, line 92: Not necessary to change, but these probably could be `const` as `PaymentFee::TEN_PROMILLE_...` etc.

modules/fedimint-lnv2-client/src/lib.rs
dpc: If you two need some back and forth to explain and agree, it means me and other devs not all that familiar with it, will have no idea, about some suble but possibly important aspects of this. Please add some comments to give us a chance if we need it in the future.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -193,16 +193,16 @@ pub struct PaymentFee {
...
+            base: Amount::from_sats(50),
             parts_per_million: 10_000,
         }
     }
 
-    pub fn half_of_one_percent() -> Self {
+    pub fn five_promille_plus_50_sats() -> Self {
         PaymentFee {
-            base: Amount::ZERO,
+            base: Amount::from_sats(50),
```
benthecarman, line 205: 50 sats for a base fee is huge, very often lightning payments are less than this.

joschisan: This is a maximum fee that the client will allow on a payment by default, you can set any value you prefer as a limit though. The gateway right now actually takes this maximum as a fee, mostly for testing that the maximum fee will actually be accepted by the client. In the future we will make this configurable on the gateways side and the client will automatically use the lower base fee set by the gateway. How low the gateway can go here mostly depends on the fee we take per each note input / output as they are driving the transaction size.

Generally all such fees are set very conservatively right now as they have DOS implications but may be lowered at some point.

modules/fedimint-lnv2-tests/tests/mock.rs
joschisan: Done.

joschisan:
@dpc @m1sterc001guy added comments.

m1sterc001guy (approved):

PR #5637, opened 2024-07-15
fix(devimint): `devimint devfed` does not block

dpc:


fedimint-backports:
- #5641

PR #5638, opened 2024-07-15
chore: print redacted user&pass on SafeUrl fmt::Display

dpc:
Otherwise it's easy to overlook that it was used to serialize to string accidentally, dropping auth and leading to hard to debug issues.


Kodylow (approved):
lgtm

PR #5640, opened 2024-07-15
fix: set version SHUTDOWN_ENDPOINT was introduced to 0.3

bradleystachurski:
Re: https://github.com/fedimint/fedimint/pull/5630#discussion_r1677223016

### DIFF
--- a/fedimint-server/src/consensus/api.rs
+++ b/fedimint-server/src/consensus/api.rs
@@ -625,7 +625,7 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
         },
         api_endpoint! {
             SHUTDOWN_ENDPOINT,
-            ApiVersion::new(0, 0),
+            ApiVersion::new(0, 3),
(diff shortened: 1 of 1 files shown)


fedimint-server/src/consensus/api.rs
```diff
@@ -625,7 +625,7 @@ pub fn server_endpoints() -> Vec<ApiEndpoint<ConsensusApi>> {
         },
         api_endpoint! {
             SHUTDOWN_ENDPOINT,
-            ApiVersion::new(0, 0),
+            ApiVersion::new(0, 3),
```
dpc, line 628: @bradleystachurski Did we even increment currently supported to 0.3? Last time I checked it was 0.2.

bradleystachurski: Yep, we're on 0.3 since this is pointed to master 👍 

https://github.com/fedimint/fedimint/blob/6c888ae86cdabe52840f381dc241501b1229a052/fedimint-server/src/config/mod.rs#L207

I'm putting together the backport PR that also bumps releases/v0.3 from 0.2 to 0.3

Issue #5490, opened 2024-06-20
LNV2 Test Coverage


joschisan:
LNV2 is still missing the following tests. For the devimint tests to work we need to enable lnv2 in devimint.

- [x] (devimint) gateway registration sanity test: https://github.com/fedimint/fedimint/pull/5488
- [x] (devimint) self payment of a client with two one gateway (direct swap): https://github.com/fedimint/fedimint/pull/5500
- [x] (devimint) self payment of a client with two gateways (lightning swap) - @m1sterc001guy 
- [x] (modular) malicious incoming contract gets refunded: https://github.com/fedimint/fedimint/pull/5556
- [x] (modular) expired incoming contract cannot be decrypted and payment fails: https://github.com/fedimint/fedimint/pull/5556
- [x] (modular) expired outgoing contract is refunded unilaterally by the client - @joschisan 
- [x] (modular) incoming contract is verified - @joschisan

m1sterc001guy:
Some additional ideas. We have coverage for most of these in v1 already.

 - (devimint): Fedimint-LN payment to LND HOLD invoice
 - (devimint): Payment from LND/CLN into Fedimint via one gateway
 - (modular): Cannot pay expired invoice
 - (devimint): change gateway federation fees, outgoing/incoming payment and verify that gateway takes correct fees
 - (devimint): payment between two federations with two separate gateways
 - (devimint): payment between two federations with one gateway
 - (modular): Can receive Pay-To-Pubkey
 - (modular): cannot pay invoice from another network (e.g signet)
 - (modular): try to pay an invoice without attaching gateway fees, verify gateway rejects it.

joschisan:
@m1sterc001guy with Pay-To-Pubkey I assume you mean receiving to someone else's pub key? If so that does not need to be testes as every regular receive is just a pay to pub key with your own key.

m1sterc001guy:
Perhaps we could add a test that uses a tweaked public key, like we do for v1

Kodylow:
Closing this, @m1sterc001guy reopening issue with remaining ideas / tests

PR #5650, opened 2024-07-16
test: LNv2 add mint to mint test

m1sterc001guy:
Working towards https://github.com/fedimint/fedimint/issues/5490

Adds some test cases where a client in one federation can pay a client in a different federation.

This doesn't fundamentally add coverage to new code paths, since LNv2 treats this test case as the same, but I think its a cool test to have, and can serve as a nice check in the future to make sure this use case does not break.

### DIFF
--- a/modules/fedimint-lnv2-tests/src/bin/tests.rs
+++ b/modules/fedimint-lnv2-tests/src/bin/tests.rs
@@ -1,6 +1,7 @@
 use clap::Parser;
 use devimint::devfed::DevJitFed;
-use devimint::federation::Client;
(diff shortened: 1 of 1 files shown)


dpc (approved):

joschisan (changes requested):
This test should be written as a loop over all four possible combinations of gateways like the self payment test.

joschisan:
Maybe we should call this test_inter_federation_payment?

joschisan (approved):

dpc (approved):

PR #5658, opened 2024-07-16
Remove multi mint payments from lnv2 to enable ldk gateways

joschisan:
lnv2 multi mint payment support blocks the ldk gateway pr as discussed there: https://github.com/fedimint/fedimint/pull/5554


m1sterc001guy (dismissed):

m1sterc001guy (approved):
Discussed this offline, it unblocks LDK gateway right now and still preserves the ability to do multi-mint payments if we want in the future

PR #5663, opened 2024-07-17
chore: liquidity api uses bytes for pubkey

tvolk131:
> Can we use bytes instead of string?

_Originally posted by @m1sterc001guy in https://github.com/fedimint/fedimint/pull/5218#discussion_r1600141463_

When adding `close_channels_with_peer` command to the gateway liquidity API, we decided to use bytes instead of a string for the pubkey. But this was after we already added a few other commands that used strings for the pubkey. Here we're going back and changing those to also use bytes for the pubkey.

### DIFF
--- a/gateway/ln-gateway/src/lightning/lnd.rs
+++ b/gateway/ln-gateway/src/lightning/lnd.rs
@@ -1244,7 +1244,8 @@ impl ILnRpcClient for GatewayLndClient {
                         };
 
                     ChannelInfo {
-                        remote_pubkey: channel.remote_pubkey,
+                        remote_pubkey: PublicKey::from_str(&channel.remote_pubkey)
+                            .expect("Lightning node returned invalid remote channel pubkey"),
                         channel_size_sats,
                         outbound_liquidity_sats,
                         inbound_liquidity_sats,
--- a/gateway/ln-gateway/src/lightning/mod.rs
+++ b/gateway/ln-gateway/src/lightning/mod.rs
@@ -218,7 +218,7 @@ impl dyn ILnRpcClient {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct ChannelInfo {
-    pub remote_pubkey: String,
+    pub remote_pubkey: secp256k1::PublicKey,
     pub channel_size_sats: u64,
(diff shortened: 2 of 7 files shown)


dpc (dismissed):
Is this backward compat?

elsirion:
> Is this backward compat?

I think so far we treated the GW as a monolithic component without internal compatibility guarantees.

tvolk131:
> > Is this backward compat?
> 
> I think so far we treated the GW as a monolithic component without internal compatibility guarantees.

Sgtm - and no this is definitely not backwards compatible, but I don't believe we've enforced any API guarantees for the liquidity API yet

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -222,7 +222,7 @@ impl dyn ILnRpcClient {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct ChannelInfo {
-    pub remote_pubkey: String,
+    pub remote_pubkey: Option<secp256k1::PublicKey>,
```
elsirion, line 225: Why is it an option? I see that parsing might fail, but that could probably be handled with an `expect` since it means that the LN service is returning invalid data.

elsirion: @tvolk131

tvolk131: I'm fine removing the `Option`! Done.

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -1244,7 +1244,8 @@ impl ILnRpcClient for GatewayLndClient {
                         };
 
                     ChannelInfo {
-                        remote_pubkey: channel.remote_pubkey,
+                        remote_pubkey: PublicKey::from_str(&channel.remote_pubkey)
+                            .expect("Server returned invalid remote channel pubkey"),
```
elsirion, line 1248: Server means lightning node in this context, right?

tvolk131: Yep! I fixed the message to make that more clear.

elsirion (approved):

PR #5667, opened 2024-07-17
docs: update lock-step upgrade docs for 0.4

bradleystachurski:
In draft until steps for getting current session count are ready.

Updates documentation for upgrading to 0.4, which requires a lock-step upgrade and coordinated shutdown for all peers at a specific session count.


docs/RELEASE_NOTES-v0.4.md
```diff
@@ -10,18 +11,42 @@ versions of `fedimintd` requires stopping all peers at the exact same session co
...
+minutes implies 5 sessions after the current session count).
+
+To schedule stopping your `fedimintd` after a specific session count, use the
+`fedimint-cli` command:
+
+```
+fedimint-cli --password <password> --our-id <our_id> admin shutdown <session_count>
+```
+
+Alternatively, the current session count and scheduling a shutdown after session count is
```
bradleystachurski, line 35: @Kodylow I'm going off of memory from the UI screenshare, so please let me know if this isn't an accurate description.

docs/RELEASE_NOTES-v0.4.md
```diff
@@ -1,27 +1,54 @@
...
+shutdown at the expected session count. Once all guardians have verified, then you can proceed
+to upgrade to the latest version.
 
 ### Acknowledging
 
 To acknowledge that you are aware of requirements described above please set:
 
 ```
-FM_REL_NOTES_ACK=0_4_xyz
+FM_REL_NOTES_ACK=0_4_0
```
dpc, line 51: Nooo. It is supposed to be `xyz` to prevent people from trying to automate it. In 0.5 it might be `0_5_banana` or something.

fedimintd/src/fedimintd.rs
```diff
@@ -470,8 +470,9 @@ fn check_release_notes_ack() -> anyhow::Result<()> {
     const FM_SKIP_REL_NOTES_ACK_ENV: &str = "FM_SKIP_REL_NOTES_ACK";
     const FM_REL_NOTES_ACK_ENV: &str = "FM_REL_NOTES_ACK";
     // the suffix here is to prevent people getting to smart and trying to
-    // automate-it-out change it for every release
-    const FM_REL_NOTES_CUR_VAL: &str = "0_4_xyz";
+    // automate-it-out
+    // update to match the current release
+    const FM_REL_NOTES_CUR_VAL: &str = "0_4_0";
```
dpc, line 475: image

bradleystachurski: Oh interesting, I read that as "change it for every release"

I'm not sure I fully understand the solution here since you can still set `FM_SKIP_REL_NOTES_ACK_ENV` to automate the process of acking for each release.

Reverting

dpc: > you can still set FM_SKIP_REL_NOTES_ACK_ENV to automate the process of acking for each release.

:shushing_face: That's not documented anywhere. Or at least in official docs. :D

bradleystachurski: Since we'd only expect `FM_SKIP_REL_NOTES_ACK_ENV` to be enabled in test envs, could we instead use `is_running_in_test_env` to skip the ack step, that way users can't automate out the ack?

Also, you've likely already thought of this, but new federations starting from 0.4 will still need to set the upgrade ack even though it isn't necessary to have a coordinated shutdown.

dpc: >  that way users can't automate out the ack

If someone really, really, really need it, then I'd like an escape hatch. I just want to avoid telling people about it.

> will still need to set the upgrade ack even though it isn't necessary to have a coordinated shutdown.

Yes. I think it's OK. Maybe we should add a line about it in the docs. "If you are setting a new Federation with fedimintd 0.4, you don't need to do anything unusual, but you still need to acknowledge reading these release notes."

@bradleystachurski

bradleystachurski: If I'm understanding correctly, the logic of `check_release_notes_ack` means that for all versions going forward they'll need to set an ack var until we remove or update `check_release_notes_ack`.

Perhaps we can check the data dir from the server opts to see if this is a new fedimintd? I'm a bit concerned about the UX for first time users from all versions going forward.

Out-of-scope for this PR so I'll create a new issue to track.

bradleystachurski: https://github.com/fedimint/fedimint/issues/5674

dpc: I am hoping that by 0.5 we could get consensus version voting, and that should eliminate the need for the lock step upgrades for most changes going forward.

dpc (approved):

dpc:
Weird. I checked the workers, all have rather empty /tmp.

dpc:
@bradleystachurski This one looks like legit flake, created https://github.com/fedimint/fedimint/issues/5676

fedimint-backports:
- #5677

PR #5668, opened 2024-07-17
feat: add session count to fedimint-cli

bradleystachurski:
We only expose the session count using the dev commands. This adds session count to `fedimint-cli` so federations can easily get the current session count, which is useful for deciding on an upcoming session count to shutdown after for coordinated upgrades.

This requires backporting to `releases/v0.3` and `releases/v0.4`.

For improved UX, we should also surface the current session count in the admin ui. See: discord [thread](https://discord.com/channels/990354215060795454/990354215878688860/1263173477121130649).

---
```
bash-5.2$ fedimint-cli session-count
{
  "count": 1
}
```

### DIFF
--- a/fedimint-cli/src/client.rs
+++ b/fedimint-cli/src/client.rs
@@ -189,6 +189,8 @@ pub enum ClientCmd {
     },
     /// Returns the client config
     Config,
+    /// Gets the current fedimint AlephBFT session count
+    SessionCount,
 }
 
 pub async fn handle_command(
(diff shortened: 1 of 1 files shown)


dpc:
I don't get why would it be under `admin`. Should be either global or under `dev` IMO.

bradleystachurski:
`dev` seems awkward in a production environment, but perhaps that's the best approach for now since it already exists and would allow us to skip cutting a v0.3.3-rc.1.

@dpc If you don't have a strong opinion against `dev`, let's just go with that and I'll update the upgrade docs.

dpc:
@bradleystachurski I don't really see why it wouldn't be just global.

bradleystachurski:
@dpc Let me know if this is what you had in mind

dpc (approved):

fedimint-backports:
- #5670

fedimint-backports:
- #5671

PR #5673, opened 2024-07-17
chore(core): Encodable for `VecDeque`

dpc:

### DIFF
--- a/fedimint-core/src/encoding/mod.rs
+++ b/fedimint-core/src/encoding/mod.rs
@@ -17,7 +17,7 @@ mod tls;
 
 use std::any::TypeId;
 use std::borrow::Cow;
-use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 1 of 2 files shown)


fedimint-core/src/encoding/mod.rs
```diff
@@ -493,19 +493,76 @@ where
...
+            .collect::<Result<VecDeque<_>, DecodeError>>()?;
+
+        // Add any excess manually avoiding any surprises.
+        while (v.len() as u64) < len {
+            v.push_back(T::consensus_decode_from_finite_reader(d, modules)?);
+        }
+
+        assert_eq!(v.len() as u64, len);
+
+        Ok(v)
```
maan2003, line 551: can we just decode as Vec and do `VecDeque::from`?

dpc: You are absolutely correct. :brain:

maan2003 (approved):

PR #5692, opened 2024-07-18
chore: default to 0s when unable to detect git hash version during build

dpc:
During 0.4 release process we've hit a case where we are unable to have auto-detection working, potentially forcing all downstream users to set env vars when building their projects.

Seems like having that git hash version always is not worth the troubles like this for the end users, so let's just default to 0s when all else failed.

### DIFF
--- a/fedimint-build/src/lib.rs
+++ b/fedimint-build/src/lib.rs
@@ -132,7 +132,8 @@ pub fn set_code_version() {
     match set_code_version_inner() {
         Ok(()) => {}
         Err(e) => {
-            panic!("Failed to detect git hash version: {e}. Set {FORCE_GIT_HASH_ENV} to skip this check")
+            eprintln!("Failed to detect git hash version: {e}. Set {FORCE_GIT_HASH_ENV} to enforce the version and skip auto-detection.");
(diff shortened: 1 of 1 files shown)


fedimint-build/src/lib.rs
```diff
@@ -132,7 +132,8 @@ pub fn set_code_version() {
     match set_code_version_inner() {
         Ok(()) => {}
         Err(e) => {
-            panic!("Failed to detect git hash version: {e}. Set {FORCE_GIT_HASH_ENV} to skip this check")
+            eprintln!("Failed to detect git hash version: {e}. Set {FORCE_GIT_HASH_ENV} to enforce the version and skip auto-detection.");
+            println!("cargo:rustc-env={FEDIMINT_BUILD_CODE_VERSION_ENV}=0000000000000000000000000000000000000000");
```
bradleystachurski, line 136: 👍 40 chars

bradleystachurski (approved):

bradleystachurski:
Another issue with the self-hosted runner

```
The self-hosted runner: runner-02-cc lost communication with the server
```
https://github.com/fedimint/fedimint/actions/runs/9997756499?pr=5692

Restarted the failing jobs

fedimint-backports:
- #5696

PR #5697, opened 2024-07-19
chore: wallet client recovery privacy improvements

### LINKED ISSUE #5584: Wallet recovery: sync federation history to preserve privacy
Our current implementation of recovering the wallet client degrades privacy by batching requests to both the bitcoind backend (to check an address's history) and to the federation (to attempt claiming any potentially missed peg-ins). Since the queries are done for all tweaks controlled by the client until there's a gap of unused addresses, it's trivial for the federation to correlate the addresses to a single client even if the requests are made over separate Tor circuits.

Instead of iterating through all tweak indexes on the client until we find a tweaked address without history, the client can download the history of session outcomes of the federation, similar to fedimint-observer. The client can then reference the local copy of the full history of peg-ins to the federation without revealing to the federation the addresses the client tweaked. This also obviates the need to query a bitcoind backend to check if an address has history for already claimed peg-ins.

More context in this PR [thread](https://github.com/fedimint/fedimint/pull/5546#discussion_r1664721674)

dpc:
See comments in the code.

Fix #5584

### DIFF
--- a/fedimint-cli/src/lib.rs
+++ b/fedimint-cli/src/lib.rs
@@ -38,7 +38,6 @@ use fedimint_api_client::api::{
 use fedimint_bip39::Bip39RootSecretStrategy;
 use fedimint_client::meta::{FetchKind, MetaSource};
 use fedimint_client::module::init::{ClientModuleInit, ClientModuleInitRegistry};
-use fedimint_client::module::ClientModule as _;
 use fedimint_client::secret::{get_default_client_secret, RootSecretStrategy};
 use fedimint_client::{AdminCreds, Client, ClientBuilder, ClientHandleArc};
 use fedimint_core::admin_client::{ConfigGenConnectionsRequest, ConfigGenParamsRequest};
@@ -54,7 +53,7 @@ use fedimint_core::{fedimint_build_code_version_env, runtime, PeerId, TieredMult
 use fedimint_ln_client::LightningClientInit;
 use fedimint_logging::{TracingSetup, LOG_CLIENT};
 use fedimint_meta_client::{MetaClientInit, MetaModuleOrLegacyMetaSource};
-use fedimint_mint_client::{MintClientInit, MintClientModule, OOBNotes, SpendableNote};
+use fedimint_mint_client::{MintClientInit, OOBNotes, SpendableNote};
 use fedimint_wallet_client::api::WalletFederationApi;
 use fedimint_wallet_client::{WalletClientInit, WalletClientModule};
 use futures::future::pending;
@@ -794,10 +793,7 @@ impl FedimintCli {
                 // TODO: until we implement recovery for other modules we can't really wait
                 // for more than this one
                 debug!("Waiting for mint module recovery to finish");
-                client
-                    .wait_for_module_kind_recovery(MintClientModule::kind())
-                    .await
-                    .map_err_cli()?;
+                client.wait_for_all_recoveries().await.map_err_cli()?;
 
                 debug!("Recovery complete");
 
--- a/modules/fedimint-wallet-client/src/backup.rs
+++ b/modules/fedimint-wallet-client/src/backup.rs
@@ -443,35 +443,42 @@ where
     F: Fn(TweakIdx) -> FF,
     FF: Future<Output = anyhow::Result<Vec<T>>>,
 {
-    let mut last_used_idx = previous_next_unused_idx.prev();
+    // TODO: change fn arg to scan_from_idx instead of previous_next_unused_idx
+    let scan_from_idx = previous_next_unused_idx.prev().unwrap_or_default();
 
-    let mut cur_tweak_idx = previous_next_unused_idx;
+    let tweak_indexes_to_scan = (scan_from_idx.0..).map(TweakIdx).filter(|tweak_idx| {
+        let already_used = used_tweak_idxes.contains(tweak_idx);
+
+        if already_used {
+            debug!(target: LOG_CLIENT_MODULE_WALLET,
+                %tweak_idx,
+                "Skipping checking history of an address, as it was previously used"
+            );
+        }
+
+        !already_used
+    });
+
+    // Last tweak index which had on-chain activity, used to implement a gap limit,
+    // i.e. scanning a certain number of addresses past the last one that had
+    // activity.
+    let mut last_used_idx = None;
     let mut tweak_idxes_with_pegins = BTreeSet::new();
 
-    let new_start_idx = loop {
-        let gap_since_last_used = cur_tweak_idx - last_used_idx.unwrap_or_default();
-
-        if RECOVER_MAX_GAP <= gap_since_last_used {
-            break last_used_idx
-                .unwrap_or_default()
-                .advance(RECOVER_NUM_IDX_ADD_TO_LAST_USED);
+    for cur_tweak_idx in tweak_indexes_to_scan {
+        if RECOVER_MAX_GAP <= cur_tweak_idx - last_used_idx.unwrap_or_default() {
+            break;
         }
 
-        if used_tweak_idxes.contains(&cur_tweak_idx) {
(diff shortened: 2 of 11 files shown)


dpc:
ATM. This needs another round of self-review, but is probably in reviewable state.

fedimint-bitcoind/src/lib.rs
```diff
@@ -64,24 +61,16 @@ static BITCOIN_RPC_REGISTRY: Lazy<Mutex<BTreeMap<String, DynBitcoindRpcFactory>>
 pub fn create_bitcoind(config: &BitcoinRpcConfig, handle: TaskHandle) -> Result<DynBitcoindRpc> {
     let registry = BITCOIN_RPC_REGISTRY.lock().expect("lock poisoned");
 
-    let kind = env::var(FM_FORCE_BITCOIN_RPC_KIND_ENV)
```
dpc, line 67: Ooops. I didn't want to make this change.

elsirion (changes requested):
The algorithm likely skips scripts that could have received money

modules/fedimint-wallet-client/src/backup/recovery_history_tracker.rs
```diff
@@ -0,0 +1,128 @@
...
+/// are actually linked with each other.
+#[derive(Clone, Debug, Encodable, Decodable)]
+pub struct ConsensusPegInTweakIdxesUsedTracker {
+    pending_pubkey_scripts: BTreeMap<bitcoin::ScriptBuf, TweakIdx>,
+    next_pending_tweak_idx: TweakIdx,
+    next_unused_tweak_idx: TweakIdx,
+    decoys: VecDeque<bitcoin::ScriptBuf>,
+    // To avoid updating `decoys` for the whole recovery, which might be a lot of extra updates
+    // most of which will be thrown away, ignore script pubkeys from before this `session_idx`
+    decoy_session_threshold: u64,
```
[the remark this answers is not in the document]
dpc: The cost of a cloning is more of a problem, potentially. Also, we really only want some recently used addresses as a decoy, as we are going to check only few (potentially) last used addresses. An attacker might see through things if we query some much older addresses.

modules/fedimint-wallet-client/src/backup/recovery_history_tracker.rs
```diff
@@ -0,0 +1,128 @@
...
+    ) {
+        if let Some(tweak_idx) = self.pending_pubkey_scripts.remove(script) {
+            debug!(target: LOG_CLIENT_MODULE_WALLET, %session_idx, ?tweak_idx, "Found previously used tweak_idx in federation history");
+            self.generate_next_pending_tweak_idx(data);
+
+            let mut to_remove = self.next_unused_tweak_idx;
+            while to_remove <= tweak_idx {
+                self.remove_and_replenish_by_tweakidx(data, tweak_idx);
+                to_remove = to_remove.next();
+            }
```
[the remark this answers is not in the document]
dpc: Oh, good catch.

W.r.t using in order - since the request was to improve recovery privacy, my initial thought was that all I need is to find last claimed idx, and then query the blockchain from there to find pegged-in, but yet unclaimed addresses. Since we need to expose ability to recheck deposits to particular address in the apps, I thought that is good enough.

But indeed I can maintain a collection of all indexes that were already claimed instead, and then just skip querying blockchain for deposits on them. I don't really ever need to remove anything from the list of addresses I'm looking for - even if there's >10k of them, it shouldn't matter much anyway. And querying long ago unused addresses is not much of a privacy issue, as they will probably not get used in the future anyway.

modules/fedimint-wallet-client/src/backup/recovery_history_tracker.rs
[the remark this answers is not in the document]
dpc: Since I might need to query old addresses anyway, maybe I'll just replace it with some probabilistic thing to potentially keep some older addresses around...

modules/fedimint-wallet-client/src/backup.rs
```diff
@@ -37,3 +58,364 @@ pub struct WalletModuleBackupV0 {
...
+            cfg: args.cfg().clone(),
+            module_root_secret: args.module_root_secret().clone(),
+        };
+        Ok(dbtx.get_value(&RecoveryStateKey)
+            .await
+            .and_then(|(state, common)| {
+                if let WalletRecoveryState::V0(state) = state {
+                    Some((state, common))
+                } else {
+                    warn!(target: LOG_CLIENT_RECOVERY, "Found unknown version recovery state. Ignoring");
```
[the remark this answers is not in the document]
dpc: AFAICT, no matter what there's just no reason not to restart recovery from scratch.

modules/fedimint-wallet-client/src/backup.rs
```diff
@@ -411,11 +412,18 @@ where
...
+                "Skipping checking history of an address, as it was previously used");
             last_used_idx = Some(cur_tweak_idx);
-            tweak_idxes_with_pegins.insert(cur_tweak_idx);
+        } else {
+            let history = check_addr_history(cur_tweak_idx).await?;
+
+            if !history.is_empty() {
+                last_used_idx = Some(cur_tweak_idx);
+                tweak_idxes_with_pegins.insert(cur_tweak_idx);
+            }
```
elsirion, line 426: The amount of mutability here makes it quite hard to review, I was procrastinating on reviewing this PR for way too long due to it. Let's discuss during review club, I think there is a beautiful, small algorithm hidden in here that wants to get implemented with some iterator chaining :laughing:

modules/fedimint-wallet-client/src/backup.rs
```diff
@@ -40,3 +60,377 @@ pub struct WalletModuleBackupV0 {
...
+        // mutex
+        let tracker = &Arc::new(Mutex::new(self.state.tracker.clone()));
+
+        debug!(target: LOG_CLIENT_MODULE_WALLET,
+            next_unused_tweak_idx = ?self.state.next_unused_idx_from_backup,
+            "Scanning blockchain for used peg-in addresses");
+        let RecoverScanOutcome { last_used_idx: _, new_start_idx, tweak_idxes_with_pegins}
+            = recover_scan_idxes_for_activity(
+                self.state.next_unused_idx_from_backup,
+                self.state.tracker.used_tweak_idxes(),
```
elsirion, line 272: Eventually we should scan from index zero to not miss anything, but we need the already claimed list in the backup to do so privately.

modules/fedimint-wallet-client/src/backup.rs
elsirion: call: could loop over a range

dpc: @elsirion Hmm... now that I'm looking at it ... if I do `while` or `for`, I can't use `break value;` which will introduce a different mutable value.

elsirion (approved):
I think the final version we settled on is good, I'm trying to simplify that one fn nonetheless, see my take here: https://github.com/fedimint/fedimint/pull/5903/files/30ccddc406fe318d941d39350a7cd77d21414c8e..3bd09478781a718bf051c74d9725bd585d8fd4d3

elsirion (dismissed):

dpc:
Seems like a legitimate flake https://github.com/fedimint/fedimint/actions/runs/10528393578/job/29173781412 , will try to repro locally.

dpc:
@elsirion CI flakiness exposed a place I've missed.

fedimint-cli/src/lib.rs
```diff
@@ -794,10 +793,7 @@ impl FedimintCli {
                 // TODO: until we implement recovery for other modules we can't really wait
```
bradleystachurski, line 793: [nit] Can also remove this TODO

bradleystachurski (approved):

PR #5701, opened 2024-07-19
chore: extract ExecutorInner state fields into separate enum

tvolk131:
`ExecutorInner.{context, shutdown_executor, sm_update_rx}` all have specific values in each of three possible states of the `ExecutorInner`: unstarted, running, and stopped. These three states are somewhat loosely defined currently. This PR extracts the three struct fields into a state enum which contains those same values depending on the state, making it more clear and explicit what state the executor is in.

### DIFF
--- a/fedimint-client/src/sm/executor.rs
+++ b/fedimint-client/src/sm/executor.rs
@@ -2,6 +2,7 @@ use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::convert::Infallible;
 use std::fmt::{Debug, Formatter};
 use std::io::{Error, Read, Write};
+use std::mem;
 use std::sync::Arc;
 use std::time::SystemTime;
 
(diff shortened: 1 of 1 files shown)


fedimint-client/src/sm/executor.rs
```diff
@@ -156,12 +228,8 @@ impl Executor {
...
                 self.inner.module_contexts.get(&state.module_instance_id())
             {
-                let context = {
-                    let context_gen_guard = self.inner.context.lock().await;
-                    let context_gen = context_gen_guard
-                        .as_ref()
-                        .expect("should be initialized at this point");
-                    context_gen(state.module_instance_id(), state.operation_id())
+                let Some(context) = self.inner.state.lock().await.gen_context(&state) else {
+                    panic!("executor should be running at this point");
```
elsirion, line 232: Could be an `.expect(…)`

elsirion (approved):
Great refactor!

fedimint-client/src/sm/executor.rs
tvolk131: Done.

dpc (approved):

PR #5704, opened 2024-07-19
chore: move gateway db code into dbtx ext trait

tvolk131:
Replaces #5633

If people like this design pattern, it's probably worth applying it for all of our databases, not just the gateway.

    I wonder if this wouldn't be better implemented on `DatabaseTransaction`, so that the caller is free to commit the way it wants.

_Originally posted by @dpc in https://github.com/fedimint/fedimint/pull/5633#discussion_r1677212351_


dpc (commented):
LGTM!

dpc (approved):

PR #5707, opened 2024-07-19
chore: minor rustdoc improvements

dpc:
I'm making a presentation, and link to some places in the Technical Reference, and want to make it look a bit better.

### DIFF
--- a/fedimint-core/src/session_outcome.rs
+++ b/fedimint-core/src/session_outcome.rs
@@ -15,11 +15,16 @@ pub struct AcceptedItem {
     pub peer: PeerId,
 }
 
-/// All items ordered in a session that have been accepted by Fedimint Consensus
-/// are recorded in the corresponding block. A running Federation produces a
-/// [SessionOutcome] roughly every five minutes.  Therefore, just like in
-/// Bitcoin, a [SessionOutcome] might be empty if no items are ordered in that
-/// time or all ordered items are discarded by Fedimint Consensus.
+/// Items ordered in a single session that have been accepted by Fedimint
+/// consensus.
+///
+/// A running Federation produces a [`SessionOutcome`] every couple of minutes.
+/// Therefore, just like in Bitcoin, a [`SessionOutcome`] might be empty if no
+/// items are ordered in that time or all ordered items are discarded by
+/// Fedimint Consensus.
+///
+/// When session is closed it is signed over by the peers and produces a
+/// [`SignedSessionOutcome`].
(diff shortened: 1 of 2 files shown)


fedimint-core/src/session_outcome.rs
```diff
@@ -15,11 +15,16 @@ pub struct AcceptedItem {
...
+/// Items ordered in a single session that have been accepted by Fedimint
+/// consensus.
+///
+/// A running Federation produces a [SessionOutcome] roughly every five minutes.
+/// Therefore, just like in Bitcoin, a [SessionOutcome] might be empty if no
+/// items are ordered in that time or all ordered items are discarded by
+/// Fedimint Consensus.
+///
+/// When session is closed, signed over by the peers, producing
+/// [`SignedSessionOutcome`]
```
bradleystachurski, line 27: Perhaps this was meant to be combined with the docs added for `SignedSessionOutcome`?

fedimint-core/src/session_outcome.rs
```diff
@@ -15,11 +15,16 @@ pub struct AcceptedItem {
...
 
-/// All items ordered in a session that have been accepted by Fedimint Consensus
-/// are recorded in the corresponding block. A running Federation produces a
-/// [SessionOutcome] roughly every five minutes.  Therefore, just like in
-/// Bitcoin, a [SessionOutcome] might be empty if no items are ordered in that
-/// time or all ordered items are discarded by Fedimint Consensus.
+/// Items ordered in a single session that have been accepted by Fedimint
+/// consensus.
+///
+/// A running Federation produces a [SessionOutcome] roughly every five minutes.
```
bradleystachurski, line 21: [nit] Without high latency, I believe we should expect a session outcome every 180s. It might be worth updating to `three minutes`, but can also see how "roughly" is close enough and leaves wiggle room for high latency networks.

@joschisan please correct me if I'm misunderstanding

fedimint-core/src/session_outcome.rs
bradleystachurski: Just reread, scratch my previous comment.

It might be an improvement to update `producing` to `it produces a`.

bradleystachurski (dismissed):

fedimint-core/src/session_outcome.rs
tvolk131, line 21: Nit: "[SessionOutcome]" -> "[\`SessionOutcome\`]"

Same below

tvolk131 (changes requested):

bradleystachurski (approved):

PR #5714, opened 2024-07-20
chore: bump lnurl-rs

dpc:
New lnurl-rs pins transitivie dep to avoid breakage.

To make it work, we need to update reqwest as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1238,9 +1238,9 @@ dependencies = [
 
 [[package]]
 name = "email_address"
-version = "0.2.4"
+version = "0.2.5"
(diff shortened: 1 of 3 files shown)


tvolk131 (approved):
Should we backport #4908 and #5711 instead? If it doesn't really matter, then LGTM!

dpc:
This is the backport, no? I didn't try automatic backport, as we did a lot of sorting of deps on master, so I doubt it would work.

tvolk131:
> we did a lot of sorting of deps on master, so I doubt it would work

Sg! Send it!

PR #5716, opened 2024-07-20
fix(client): ignore user data in the new-db check

### LINKED ISSUE #5706: incorrect assumption about empty database
https://github.com/fedimint/fedimint/blob/50e895058b2e01b95df39a0439b19dd06dd6a0d7/fedimint-core/src/db/mod.rs#L1934-L1937

this is incorrect because external application (client integrators) may store data in the database ExternalReservedStart)

dpc:
Fix #5706

### DIFF
--- a/fedimint-client/src/db.rs
+++ b/fedimint-client/src/db.rs
@@ -432,7 +432,15 @@ pub async fn apply_migrations_core_client(
     target_version: DatabaseVersion,
     migrations: BTreeMap<DatabaseVersion, CoreMigrationFn>,
 ) -> Result<(), anyhow::Error> {
-    apply_migrations(db, kind, target_version, migrations, None).await
+    apply_migrations(
+        db,
+        kind,
+        target_version,
+        migrations,
+        None,
+        Some(DbKeyPrefix::UserData as u8),
+    )
+    .await
 }
 
 /// `apply_migrations_client` iterates from the on disk database version for the
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -1913,7 +1913,7 @@ pub async fn apply_migrations_server(
     target_db_version: DatabaseVersion,
     migrations: BTreeMap<DatabaseVersion, CoreMigrationFn>,
 ) -> Result<(), anyhow::Error> {
-    apply_migrations(db, kind, target_db_version, migrations, None).await
+    apply_migrations(db, kind, target_db_version, migrations, None, None).await
 }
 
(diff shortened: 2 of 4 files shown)


dpc:
@maan2003 @m1sterc001guy

maan2003 (commented):
we use all prefixes from `0xb1` to `0xbf` unfortunately in fedi.

dpc:
> we use all prefixes from `0xb1` to `0xbf` unfortunately in fedi.

I don not understand. In this PR I ignore everything from 0xb0 and above.

fedimint-core/src/db/mod.rs
```diff
@@ -1930,11 +1930,26 @@ pub async fn apply_migrations(
...
     // Newly created databases will not have any data since they have just been
     // instantiated.
     let mut dbtx = db.begin_transaction_nc().await;
-    let is_new_db = dbtx.raw_find_by_prefix(&[]).await?.next().await.is_none();
+    let is_new_db = dbtx
+        .raw_find_by_prefix(&[])
+        .await?
+        .filter(|(key, _v)| {
+            std::future::ready(
+                external_prefixes_above.map_or(true, |external_prefixes_above| {
```
maan2003, line 1945: nit: move this to if let outside the filter, we don't have to pay the checking cost on server side

i would be really suprised if compiler is able optimize through streams wrapper

maan2003 (approved):
oh smart, I didn't read code before

fedimint-core/src/db/mod.rs
dpc: True, but comparing to the IO this really doesn't matter. My guestimate is orders of magnitude (1ns vs 1us) differences. And this code runs ones on start anyway.

fedimint-client/src/db.rs
```diff
@@ -432,7 +432,15 @@ pub async fn apply_migrations_core_client(
...
     migrations: BTreeMap<DatabaseVersion, CoreMigrationFn>,
 ) -> Result<(), anyhow::Error> {
-    apply_migrations(db, kind, target_version, migrations, None).await
+    apply_migrations(
+        db,
+        kind,
+        target_version,
+        migrations,
+        None,
+        Some(DbKeyPrefix::UserData as u8),
```
bradleystachurski, line 441: This will also filter out `InternalReservedStart` and `ModuleGlobalPrefix`. Will that cause any issues for detecting if it's a new db?

fedimint-backports:
- #5719

PR #5722, opened 2024-07-22
refactor(server): remove bind addr's from local config

dpc:
They can be controlled by command line args / env vars directly, and having them redundantly in the config is just confusing.

### DIFF
--- a/fedimint-server/src/net/peers.rs
+++ b/fedimint-server/src/net/peers.rs
@@ -70,9 +70,6 @@ struct PeerConnection<T> {
 pub struct NetworkConfig {
     /// Our federation member's identity
     pub identity: PeerId,
-    /// Our listen address for incoming connections from other federation
-    /// members
-    pub bind_addr: SocketAddr,
     /// Map of all peers' connection information we want to be connected to
     pub peers: HashMap<PeerId, SafeUrl>,
 }
@@ -177,6 +174,7 @@ where
     /// `Connector`.
     #[instrument(skip_all)]
     pub(crate) async fn new(
+        p2p_bind_addr: SocketAddr,
         cfg: NetworkConfig,
         delay_calculator: DelayCalculator,
         connect: PeerConnector<T>,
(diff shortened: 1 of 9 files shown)


dpc:
Today, when testing with @Kodylow the API signing stuff in the UI, we had an example why. Kody was changing env vars in the docker compose file, and that was not taking effect, because the values used were sourced from local config file that was created during DKG.

IMO, bind addresses are entirely a part of the outside infrastructure control.

dpc:
Would be great to get @elsirion 's feedback before doing anything here.

Kodylow:
This is a super annoying blocker for actually having people do the DNS rotation in practice. They set their DNS configuration via environment variables, but then after the federation starts they have to change the DNS both in the env and in whatever stuff got set in the local config. We should use whatever gets passed through the env or command line instead of the local config if those do get set, I like how bitcoin core does it where you pass in the config file or data dir but then any command line arguments passed always override whatever was fed in through the file/dir.

Kodylow (dismissed):
tested local with UI and fix works, approving but wait on merge for @elsirion 's okay

fedimint-server/src/net/peers.rs
```diff
@@ -70,9 +70,6 @@ struct PeerConnection<T> {
 pub struct NetworkConfig {
     /// Our federation member's identity
     pub identity: PeerId,
-    /// Our listen address for incoming connections from other federation
-    /// members
-    pub bind_addr: SocketAddr,
```
elsirion, line 75: Not sure why the bind address is removed here, afaik this struct isn't ever (de-)serialized? This just leads to the bind address becoming another argument in all the functions we are passing `NetowrkConfig` to as far as I can tell.

elsirion (approved):
Makes sense to me, just one nit.

fedimint-server/src/net/peers.rs
dpc: Hmm... Had `Serialize`, so I assumed it's serialized. I'll investigate.

bradleystachurski (approved):

fedimint-backports:
- #5771

Kodylow:
Note: confirmed with @bradleystachurski that serde Deserialize ignores extra keys when you read in from an old local.json , so won't break if running off old

PR #5723, opened 2024-07-23
fix: choose minimal block count or unix time

joschisan:


dpc:
Why? :) (The answer would be best done as a comment in the code)

bradleystachurski:
One example for a 3/4 federation.
```
95 100 1000 1000
    |   |
    |   median
    |
    max_evil
```

dpc (changes requested):
Comments. :) "Why?"

joschisan:
@dpc @bradleystachurski please review the comments

dpc (approved):

elsirion:
This is a consensus-breaking change, are we committing to implementing non-additive consensus change support for 0.5?

PR #5731, opened 2024-07-23
chore(ci): add upgrade tests vs master

bradleystachurski:
Closes https://github.com/fedimint/fedimint/issues/5604

- Adds new GH workflow that runs upgrade tests vs master once daily during off hours
- Adds coordinated shutdown logic to upgrade tests in devimint

Upgrade tests currently take a while to run, so this is triggered as a daily cron and not part of the PR jobs.

### DIFF
--- a/.github/workflows/upgrade-tests.yml
+++ b/.github/workflows/upgrade-tests.yml
@@ -1,0 +1,72 @@
+# This is a basic workflow to help you get started with Actions
+
+name: "Upgrade Tests"
+
+env:
+  PERFIT_SERVER: https://perfit.dev.fedimint.org
+
+# Controls when the workflow will run
+on:
+  schedule:
+    # run daily during low usage hours
+    - cron:  '30 4 * * *'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  tests:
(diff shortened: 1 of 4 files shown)


dpc:
LGTM

.github/workflows/upgrade-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+            connect-timeout = 15
+            stalled-download-timeout = 15
+      - uses: cachix/cachix-action@v15
+        with:
+          name: fedimint
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+        continue-on-error: true
+
+      - name: Upgrade tests
+        run: |
```
bradleystachurski, line 42: @dpc I skipped adding perfit since it looks like each workflow has a unique `--metric` identifier. Let me know if I should add before merging.

https://github.com/fedimint/fedimint/blob/2e0208f49dcf1105fc4bc0c97c16ef65cf3ae9c5/.github/workflows/ci-backwards-compatibility.yml#L64-L68

dpc (approved):

.github/workflows/upgrade-tests.yml
dpc: Can be added after we migrate GH runners.

bradleystachurski: Sounds good, created an issue to track: https://github.com/fedimint/fedimint/issues/5736

PR #5734, opened 2024-07-24
feat(client): save and verify root secret hash when creating the client

### LINKED ISSUE #3504: Save root secret hash in DB
Prevents accidentally using a wrong secret with an existing client. Related to #3464.

dpc:
Fix #3504

This prevent user/application from accidentally using a different secret, which could lead to ... who knows what. Not good things, that's for sure.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -88,7 +88,7 @@ use std::pin::Pin;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
 
-use anyhow::{anyhow, bail, Context};
+use anyhow::{anyhow, bail, ensure, Context};
 use async_stream::stream;
 use backup::ClientBackup;
 use db::{
@@ -2069,7 +2069,7 @@ impl ClientBuilder {
 
     async fn init(
         self,
-        root_secret: DerivableSecret,
+        pre_root_secret: DerivableSecret,
         config: ClientConfig,
         api_secret: Option<String>,
         init_mode: InitMode,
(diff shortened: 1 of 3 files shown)


fedimint-client/src/lib.rs
```diff
@@ -2245,15 +2251,37 @@ impl ClientBuilder {
...
+        if let Some(secret_hash) = self
+            .db_no_decoders()
+            .begin_transaction_nc()
+            .await
+            .get_value(&ClientPreRootSecretHashKey)
+            .await
+        {
+            if pre_root_secret.derive_pre_root_secret_hash() != secret_hash {
+                bail!("Secret hash does not match. Incorrect secret")
+            }
```
elsirion, line 2268: nit: could be `ensure!` macro

fedimint-client/src/lib.rs
```diff
@@ -2245,15 +2251,37 @@ impl ClientBuilder {
...
+
+            let mut dbtx = self.db_no_decoders.begin_transaction().await;
+            dbtx.insert_entry(
+                &ClientPreRootSecretHashKey,
+                &pre_root_secret.derive_pre_root_secret_hash(),
+            )
+            .await;
+            dbtx.commit_tx().await;
```
elsirion, line 2276: This should be in an else branch, shouldn't it? If the hash is in the DB and matches we don't need to write it again.

fedimint-client/src/lib.rs
```diff
@@ -2245,15 +2251,37 @@ impl ClientBuilder {
...
+            .await
+            .get_value(&ClientPreRootSecretHashKey)
+            .await
+        {
+            if pre_root_secret.derive_pre_root_secret_hash() != secret_hash {
+                bail!("Secret hash does not match. Incorrect secret")
+            }
+
+            let mut dbtx = self.db_no_decoders.begin_transaction().await;
+            dbtx.insert_entry(
```
elsirion, line 2271: If we were to expect any parallelism reading the secret hash and writing it should happen in the same DBTX, but since we have ownership of self here this seems ok?

fedimint-client/src/lib.rs
dpc: Oops.

elsirion (approved):

PR #5738, opened 2024-07-25
Revert "[Backport releases/v0.4] chore: always include invite code in oob ecash"

elsirion:
I talked to @joschisan, not every federation will want users to leak the invite code. `OOBNotesV2` will need to have an optional invite code too instead of the other way around.


bradleystachurski:
Restarting failed CI jobs

```
The self-hosted runner: runner-01-aa lost communication with the server.
```

bradleystachurski (approved):

PR #5747, opened 2024-07-26
feat: track ordering latency

### LINKED ISSUE #3759: Measure consensus ordering latency
We could easily measure consensus latency by recording the time of our own submissions to consensus in a `VecDequeue` and comparing them in the order that *our* items are returned from consensus as they are guaranteed to be ordered identically. This doesn't require any new consensus items.

This came out of a discussion between @joschisan and me about #3757.

dpc:
This is resurrecting @joschisan 's #5432 with a fix I proposed for self-synchronization.

Fix #3759 - which I spotted during going through old issues and it reminded me about #5432.

This PR was&is a great idea, it has minimal implementation and I believe the fix I proposed is absolutely practical and will work just fine. There is really no point sitting on it: We could have been already gathering information from real deployments had it been a part of 0.4.

If you figure out a better way to do it, we can just replace it. It's not a consensus or any backward compat-encumbered code. Let's not let perfect be an enemy of good enough. Even if this metric would return wrong results 50% of the time, it still would be very useful.

### DIFF
--- a/fedimint-server/src/consensus/aleph_bft/data_provider.rs
+++ b/fedimint-server/src/consensus/aleph_bft/data_provider.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeSet;
+use std::time::Instant;
 
 use fedimint_core::config::ALEPH_BFT_UNIT_BYTE_LIMIT;
 use fedimint_core::encoding::Encodable;
@@ -33,18 +34,25 @@ pub struct DataProvider {
     signature_receiver: watch::Receiver<Option<SchnorrSignature>>,
     submitted_transactions: BTreeSet<TransactionId>,
     leftover_item: Option<ConsensusItem>,
+    // Since it's possible that `fedimintd` after restart will receive citems it
+    // sent before restart, we use the item's length in bytes, as a simple method
+    // to self-synchronize. See <https://github.com/fedimint/fedimint/pull/5432#issuecomment-2176860609>
+    // for discussion about it.
+    timestamp_sender: async_channel::Sender<(Instant, usize)>,
 }
 
 impl DataProvider {
     pub fn new(
         mempool_item_receiver: async_channel::Receiver<ConsensusItem>,
         signature_receiver: watch::Receiver<Option<SchnorrSignature>>,
+        timestamp_sender: async_channel::Sender<(Instant, usize)>,
     ) -> Self {
         Self {
             mempool_item_receiver,
             signature_receiver,
             submitted_transactions: BTreeSet::new(),
             leftover_item: None,
+            timestamp_sender,
         }
     }
 }
@@ -99,6 +107,11 @@ impl aleph_bft::DataProvider<UnitData> for DataProvider {
         let bytes = items.consensus_encode_to_vec();
 
         assert!(bytes.len() <= ALEPH_BFT_UNIT_BYTE_LIMIT);
+
+        self.timestamp_sender
+            .send((Instant::now(), bytes.len()))
+            .await
+            .ok();
 
         Some(UnitData::Batch(bytes))
     }
--- a/fedimint-server/src/consensus/aleph_bft/data_provider.rs
+++ b/fedimint-server/src/consensus/aleph_bft/data_provider.rs
@@ -1,6 +1,7 @@
 use std::collections::BTreeSet;
 use std::time::Instant;
 
+use bitcoin_hashes::Hash;
 use fedimint_core::config::ALEPH_BFT_UNIT_BYTE_LIMIT;
 use fedimint_core::encoding::Encodable;
 use fedimint_core::epoch::ConsensusItem;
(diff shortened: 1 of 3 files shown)


dpc:
```
fedimint devimint> curl -s http://127.0.0.1:11368/metrics | grep latency
# HELP fm_consensus_ordering_latency_seconds Duration of ordering a batch of consensus items
# TYPE fm_consensus_ordering_latency_seconds histogram
fm_consensus_ordering_latency_seconds_bucket{le="0.005"} 0
fm_consensus_ordering_latency_seconds_bucket{le="0.01"} 0
fm_consensus_ordering_latency_seconds_bucket{le="0.025"} 0
fm_consensus_ordering_latency_seconds_bucket{le="0.05"} 0
fm_consensus_ordering_latency_seconds_bucket{le="0.1"} 1
fm_consensus_ordering_latency_seconds_bucket{le="0.25"} 7264
fm_consensus_ordering_latency_seconds_bucket{le="0.5"} 9484
fm_consensus_ordering_latency_seconds_bucket{le="1"} 9494
fm_consensus_ordering_latency_seconds_bucket{le="2.5"} 9494
fm_consensus_ordering_latency_seconds_bucket{le="5"} 9494
fm_consensus_ordering_latency_seconds_bucket{le="10"} 9494
fm_consensus_ordering_latency_seconds_bucket{le="+Inf"} 9494
fm_consensus_ordering_latency_seconds_sum 2119.4367896849985
fm_consensus_ordering_latency_seconds_count 9494
```

elsirion (approved):
LGTM, I'd even backport to 0.4.1 once we cut 0.4.0

fedimint-server/src/consensus/aleph_bft/data_provider.rs
```diff
@@ -100,6 +108,11 @@ impl aleph_bft::DataProvider<UnitData> for DataProvider {
 
         assert!(bytes.len() <= ALEPH_BFT_UNIT_BYTE_LIMIT);
 
+        self.timestamp_sender
+            .send((Instant::now(), bytes.len()))
```
joschisan, line 112: Lets use a consensus hash here.

joschisan (changes requested):

fedimint-server/src/consensus/aleph_bft/data_provider.rs
elsirion: To be clear: send the hash instead of the length over the channel? Sounds good to me (or if we are stingy with bytes: take the first 8 bytes of the hash, should still be plenty collision-resistant and keeps the same size as length).

fedimint-server/src/consensus/engine.rs
```diff
@@ -341,6 +352,21 @@ impl ConsensusEngine {
...
 
                     if let Some(UnitData::Batch(bytes)) = ordered_unit.data {
+                        if ordered_unit.creator == self.identity() {
+                             match timestamp_receiver.try_recv() {
+                                Ok((timestamp, bytes_len)) => {
+                                    if bytes.len() == bytes_len {
+                                        CONSENSUS_ORDERING_LATENCY_SECONDS.observe(timestamp.elapsed().as_secs_f64());
+                                    } else {
+                                        info!(target: LOG_CONSENSUS, "Not reporting ordering latency on possibly out of sync item");
+                                    }
```
joschisan, line 362: This will not synchronise after a restart, the offset will remain.

dpc: Right. I need a loop. :D

fedimint-server/src/consensus/aleph_bft/data_provider.rs
dpc: Having to clone here `bytes.clone()` is where the performance cost is. Otherwise it doesn't matter.

I can probably do `len ^ bytes[..8]` kind of thing, which is cheap and use both len and encoding itself.

fedimint-server/src/consensus/aleph_bft/data_provider.rs
```diff
@@ -100,6 +108,20 @@ impl aleph_bft::DataProvider<UnitData> for DataProvider {
...
+            .send((Instant::now(), get_citem_bytes_chsum(&bytes)))
+            .await
+            .ok();
+
         Some(UnitData::Batch(bytes))
     }
 }
+
+/// Calculate a cheap chesum of an encoded citem, as a
+/// XOR of its len and first 8 bytes.
```
elsirion, line 121: Why only the first 8 bytes :sob: hashing isn't that expensive, even if it was we could use a CRC instead of SHA256.

dpc: Ehhh, the whole thing will be exercised once a year, max, because what are the chances that long running daemon like `fedimintd` gets restarted right after it broadcasted an item, and then after the restart but before receiving such item, it also broadcasted a different item already? And a accidental conflict here just means prom metrics will report a single or a handful of prom metric with some slightly off time. It really just doesn't matter at all, so why do we worry so much?

And the code here is exactly on the codepath we're trying to measure and keep well oiled. Pulling CRC deps or doing some extra hashing is IMO optimizing for the irrelevant.

Plus if I'm prove wrong, this can be changed extra easily.

dpc: Changed. Most important is to optimize for landing stuff.

joschisan (approved):

elsirion (approved):

dpc:
Following up in https://github.com/fedimint/fedimint/discussions/6521

PR #5750, opened 2024-07-26
chore: polish lnv2 tpe

joschisan:

### DIFF
--- a/crypto/tpe/src/lib.rs
+++ b/crypto/tpe/src/lib.rs
@@ -90,6 +90,7 @@ pub fn encrypt_preimage(
     }
 }
 
+/// The function asserts that the ciphertext is valid.
 pub fn verify_agg_decryption_key(
     agg_pk: &AggregatePublicKey,
     agg_dk: &AggregateDecryptionKey,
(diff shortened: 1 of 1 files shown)


crypto/tpe/src/lib.rs
```diff
@@ -130,14 +131,25 @@ fn hash_to_message(
...
+
+    engine
+        .write_all(&ephemeral_pk.to_compressed())
+        .expect("Writing to a hash engine cannot fail");
+
+    engine
+        .write_all(commitment.as_byte_array())
+        .expect("Writing to a hash engine cannot fail");
+
+    let seed = sha256::Hash::from_engine(engine).to_byte_array();
```
elsirion: Any chance we can get rid of the `fedimint-core` dep this way?

crypto/tpe/src/lib.rs
```diff
@@ -98,13 +99,24 @@ pub fn verify_agg_decryption_key(
 ) -> bool {
     let message = hash_to_message(&ct.encrypted_preimage, &ct.pk.0, commitment);
 
+    assert_eq!(
+        pairing(&G1Affine::generator(), &ct.signature.0),
+        pairing(&ct.pk.0, &message)
+    );
```
elsirion, line 144: Why is it ok to panic here if the ciphertext is invalid? Are users expected to check that before calling this fn? If so, should this be a `debug_assert!(…)` so that we don't run twice the pairing operations in production?

If it's not checked before, why can't it be exploited as a DoS vector? I assume `CipherText` may come from untrusted sources.

crypto/tpe/src/lib.rs
joschisan: No because we need it to derive the consensus encoding.

crypto/tpe/src/lib.rs
joschisan: Yes, users are expected to check the cipher text before calling this function.

joschisan: I would leave the assert for now as the compute cost is driven by the mint.

elsirion (approved):

PR #5754, opened 2024-07-26
fix(ci): cachix commands failing in builds flows

dpc:
Tested in https://github.com/fedimint/fedimint/actions/runs/10116210016/job/27978568323 as a part of a 0.4 backport of #5753


elsirion (approved):
Let's see if that works

PR #5766, opened 2024-07-27
feat(consensus): consensus version votes, improved (wip)

dpc:
Working on improving #5568

TODO:

* upgrade mechanism (manual, automatic, accelerated flag?)
* first upgrade mechanism (extra endpoint, manual?)
* enum vs Option<ModuleInstanceId>
* get rid of long lived cache

### DIFF
--- a/fedimint-server/src/consensus/db.rs
+++ b/fedimint-server/src/consensus/db.rs
@@ -243,6 +243,152 @@ fn get_consensus_from_votes(
     votes[0].map(|v| v.desired)
 }
 
+#[test]
+fn get_consensus_from_votes_sanity() {
+    for (raw_votes, res) in [
+        ([Some((0, 0, false))].as_slice(), Some((0, 0))),
+        (&[Some((1, 2, false)), None, None, None], None),
+        (
+            &[Some((1, 2, false)), Some((1, 2, false)), None, None],
+            None,
+        ),
+        (
+            &[
+                Some((1, 2, false)),
+                Some((1, 2, false)),
+                Some((1, 2, false)),
+                None,
+            ],
+            None,
+        ),
+        (
+            &[
+                Some((1, 2, false)),
+                Some((1, 2, false)),
+                Some((1, 2, false)),
+                Some((1, 2, false)),
+            ],
+            Some((1, 2)),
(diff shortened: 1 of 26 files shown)


fedimint-server/src/consensus/db.rs
```diff
@@ -95,6 +105,146 @@ pub fn get_global_database_migrations() -> BTreeMap<DatabaseVersion, CoreMigrati
...
+            .chain(stream::repeat(None))
+            .take(num_peers.total())
+            .collect::<Vec<_>>()
+            .await;
+
+        get_consensus_from_votes(&mut votes, num_peers)
+    }
+}
+
+fn get_consensus_from_votes(
```
dpc, line 217: TODO: write a lot of tests`!!!`

fedimint-server/src/consensus/engine.rs
```diff
@@ -804,3 +1019,44 @@ pub async fn get_finished_session_count_static(dbtx: &mut DatabaseTransaction<'_
...
+///
+/// Create by using [`ConsensusEngine::load_consensus_version_cache`]
+#[derive(Debug)]
+struct ConsensusVersionCache {
+    // Currently nothing is using it, but in the future we can pass these
+    // to modules.
+    #[allow(unused)]
+    versions: BTreeMap<Option<ModuleInstanceId>, ConsensusVersion>,
+    invalidated: bool,
+}
```
joschisan, line 1039: Caching this is way to risky.

dpc: @joschisan

I agree that caching here is a potential source of bugs, but I don't see any practical implementation without it.

Since we need to pass the consensus version for a given module to every `process_x` we can't be reading it from the db every time. That would possibly x10 amount of db reads in consensus processing. 


I implemented the caching carefully: there's only single place where the actual versions could change (processing of a vote), and the cache gets invalidated whenever that code runs.

dpc:
This is probably ready for a wider review. 

I'm still thinking about how to test it, and the only decent idea I have is writting a dev-only server module, that reports different supported versions triggered by different env variable values. But then this still require restarting `fedimintd`...

PR #5769, opened 2024-07-28
feat: port p2p networking layer to iroh net

joschisan:


elsirion:
Are we tracking the learnings and further steps that came from this experiment?

joschisan:
Not yet.

PR #5774, opened 2024-07-29
chore: remove api_version fns

tvolk131:


fedimint-client/src/module/init.rs
```diff
@@ -73,7 +73,8 @@ where
         &self.module_api_version
     }
 
-    // TODO: deprecate, use `module_api_version` instead
+    // TODO: Remove in 0.6.0
+    #[deprecated(since = "0.5.0", note = "Please use `module_api_version` instead")]
```
elsirion, line 77: Should we just remove the functions? Our rust API stability guarantees are much lower, every major release can be breaking imo (ideally not unnecessarily so of course). We don't have to keep redundant functionality around like with the client/server API.

tvolk131: Sgtm! Done.

elsirion (approved):

PR #5775, opened 2024-07-30
fix: increase maximum lightning routing fees to LDK default

joschisan:


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -355,7 +363,7 @@ impl LightningClientModule {
         self.send_internal(
             gateway_api,
             invoice,
-            PaymentFee::TEN_PROMILLE_PLUS_50_SATS,
+            PaymentFee::SEND_FEE_LIMIT_DEFAULT,
             EXPIRATION_DELTA_LIMIT_DEFAULT,
         )
         .await
```
joschisan, line 369: This is the important change here, the client accepts a higher fee to enable the gateway to route with higher fees.

PR #5777, opened 2024-07-30
refactor: cli crate should not depend on server crate

dpc:
All for one env variable ... ?


fedimint-cli/src/envs.rs
```diff
@@ -10,3 +10,6 @@ pub const FM_PASSWORD_ENV: &str = "FM_PASSWORD";
 
 // Api authentication secret
 pub const FM_API_SECRET_ENV: &str = "FM_API_SECRET";
+
+/// Salt backup for combining with the private key
+pub const SALT_FILE: &str = "private.salt";
```
dpc, line 15: Copied from `fedimint-server/src/envs.rs`.

Kodylow (approved):
lgtm

PR #5778, opened 2024-07-30
fix(docs): Search does not hide the static content

dpc:
If you go to https://docs.fedimint.org and type something in the search box, you'll see that the results are displayed under the index.html we added. It's annoying, so this PR fixes it.

### DIFF
--- a/justfile.fedimint.just
+++ b/justfile.fedimint.just
@@ -210,6 +210,10 @@ fuzz-target-debug TARGET="" CRASH="" *ARGS="--exit_upon_crash":
 build-docs:
   ./scripts/dev/build-docs.sh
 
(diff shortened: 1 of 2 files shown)


justfile.fedimint.just
```diff
@@ -210,6 +210,10 @@ fuzz-target-debug TARGET="" CRASH="" *ARGS="--exit_upon_crash":
 build-docs:
   ./scripts/dev/build-docs.sh
 
+# Build `cargo doc`-generated documentation just like on https://docs.fedimint.org
+build-docs-nix: build-docs
```
dpc, line 214: Use `just build-docs-nix` and `xdg-open result/share/doc/index.html` to test.

Kodylow (approved):
lgtm

PR #5786, opened 2024-08-02
chore: cleanup confusing loop

tvolk131:
This loop is confusing since it actually only loops once, but after looking it over it all seems necessary. Let's document it for future readers.

### DIFF
--- a/fedimint-api-client/src/api/mod.rs
+++ b/fedimint-api-client/src/api/mod.rs
@@ -963,23 +963,29 @@ where
 {
     #[instrument(level = "trace", fields(peer = %self.peer_id, %method), skip_all)]
     pub async fn request(&self, method: &str, params: &[Value]) -> JsonRpcResult<Value> {
+        // Strategies using timeouts often depend on failing requests returning quickly,
+        // so every request gets only one reconnection attempt.
+        const RETRIES: usize = 1;
+
         for attempts in 0.. {
-            debug_assert!(attempts <= 1);
+            // The `match` statement below should always return if `RETRIES <= attempts`, so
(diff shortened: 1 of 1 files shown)


fedimint-api-client/src/api/mod.rs
```diff
@@ -963,23 +963,29 @@ where
...
             let rclient = self.client.read().await;
             match rclient.client.get_try().await {
                 Ok(client) if client.is_connected() => {
                     return client.request::<_, _>(method, params).await;
                 }
                 Err(e) => {
-                    // Strategies using timeouts often depend on failing requests returning quickly,
-                    // so every request gets only one reconnection attempt.
-                    if 0 < attempts {
+                    if attempts >= RETRIES {
```
dpc, line 982: This is just an opinion, but I hope to eventually convenience everyone: https://thehardcorecoder.com/2014/03/05/always-use-less-than/

dpc (commented):
One opinion, but I agree that this is an improvement. :+1:

fedimint-api-client/src/api/mod.rs
tvolk131: Interesting, that makes sense to me! Done.

I also wonder how difficult this would be to enforce in pre-commit checks...

elsirion (approved):

fedimint-api-client/src/api/mod.rs
bradleystachurski: Ooh that is a useful convention :+1:

dpc: I have a mild attitude towards it. I think it's better, but not so much better to worry about enforcing it.

PR #5788, opened 2024-08-02
chore: remove Gateway.client_joining_lock

tvolk131:
Rather than having a lock for joining & leaving federations that carries only semantic meaning, we can grab the write lock for `Gateway.federation_manager`. Locking on the data we're modifying makes it more clear what's going on.


elsirion (approved):
Nice cleanup!

m1sterc001guy (approved):

PR #5794, opened 2024-08-04
chore: implement ILnRpcClient::routehints() for ldk gateway

tvolk131:


tvolk131:
Turns out we don't need this functionality. `ILnRpcClient::routehints()` is used by LNv1 for the final inbound lightning hop into the federation, since that happens over a fake lightning channel. For LNv2, we don't need this. However, since gateways will likely often only have one or two channels, we should make sure that LNv2 invoices (which are created by the gateway) have route hints. `ldk-node` v0.3.0 relies on `lightning-invoice` v0.31.0, which already adds route hints when creating invoices:

https://docs.rs/lightning-invoice/0.31.0/src/lightning_invoice/utils.rs.html#562-565

https://docs.rs/lightning-invoice/0.31.0/src/lightning_invoice/utils.rs.html#581-604

PR #5795, opened 2024-08-04
chore: remove IntoDynNever

tvolk131:


elsirion (approved):
Happy to remove unused code. I can't really tell where it was ever used @dpc. If it was meant for external modules please lmk, then we can revert.

PR #5796, opened 2024-08-04
chore: review our usage of footgun-y: `FuturesUnordered` and `BufferedStream`

dpc:
See commit messages.

### DIFF
--- a/fedimint-client/src/module/init/recovery.rs
+++ b/fedimint-client/src/module/init/recovery.rs
@@ -8,7 +8,7 @@ use fedimint_core::epoch::ConsensusItem;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::module::{ApiVersion, ModuleCommon};
 use fedimint_core::session_outcome::{AcceptedItem, SessionStatus};
-use fedimint_core::task::{MaybeSend, MaybeSync};
+use fedimint_core::task::{MaybeSend, MaybeSync, ShuttingDownError, TaskGroup};
 use fedimint_core::transaction::Transaction;
 use fedimint_core::{apply, async_trait_maybe_send, OutPoint};
 use fedimint_logging::LOG_CLIENT_RECOVERY;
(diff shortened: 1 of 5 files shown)


fedimint-client/src/module/init/recovery.rs
```diff
@@ -246,45 +248,53 @@ where
...
-                                    // We don't want PARALLISM_LEVEL tasks hammering Federation
-                                    // with requests, so max sleep is significant
-                                    if retry_sleep <= MAX_SLEEP {
-                                        retry_sleep = retry_sleep
-                                            + thread_rng().gen_range(Duration::ZERO..=retry_sleep);
+                    Box::pin(async move {
+                        // NOTE: Each block is fetched in a spawned task. This avoids a footgun
+                        // of stuff in streams not making any progress when the stream itself
+                        // is not being polled, and possibly can increase the fetching performance.
+                        task_group.spawn_cancellable("recovery fetch block", async move {
```
dpc, line 257: The logic was moved into `spawn_cancellable`.

fedimint-client/src/sm/executor.rs
```diff
@@ -504,29 +512,31 @@ impl ExecutorInner {
...
-                    let Some(meta) = self.get_active_state(&state).await else {
-                        warn!(target: LOG_CLIENT_REACTOR, operation_id = %state.operation_id().fmt_short(), "Couldn't look up received state machine. Ignoring.");
-                        continue;
-                    };
+                    currently_running_sms.insert(state.clone());
+                    let futures_len = futures.len();
+                    let global_context_gen = &global_context_gen;
+                    trace!(target: LOG_CLIENT_REACTOR, state = ?state, "Started new active state machine, details.");
+                    futures.push(Box::pin(async move {
+                        let Some(meta) = self.get_active_state(&state).await else {
```
dpc, line 520: This and and below check require `.await` during which other state machines are not being polled. It doesn't seem like much, but it is relatively easy to make them inside a new future, and if the checks fail, return a new type of a loop event. This way the main executor loop only ever polls all `futures`, and adds new futures to `futures`, avoiding any problems with `FuturesUnordered`.

fedimint-client/src/module/init/recovery.rs
m1sterc001guy: Is this related to the `FuturesUnordered` change or did you just want to move this into a cancellable task?

m1sterc001guy (approved):
Changes make sense to me, I'm not an expert on the executor though

fedimint-client/src/module/init/recovery.rs
dpc: Moving to a task is a solution to FuturesUnordered shortcommings.

maan2003:
Approved, I reviewed few days ago, just forgot to submit review.

fedimint-client/src/sm/executor.rs
elsirion: What's reason to not spawn a task for each contained future here? Performance?

Issue #5598, opened 2024-07-10
Cannot initially receive small lightning payments


m1sterc001guy:
In LNv2, there are fees charged when claiming an input, due to DOS reasons. Currently, if the client cannot claim an input due to insufficient balance, the client will crash:

```rust
async fn claim_input_dyn(
        &self,
        dbtx: &mut ClientSMDatabaseTransaction<'_, '_>,
        input: InstancelessDynClientInput,
    ) -> (TransactionId, Vec<OutPoint>) {
        let instance_input = ClientInput {
            input: DynInput::from_parts(self.module_instance_id, input.input),
            keys: input.keys,
            amount: input.amount,
            state_machines: states_add_instance(self.module_instance_id, input.state_machines),
        };

        self.client
            .finalize_and_submit_transaction_inner(
                &mut dbtx.global_tx().to_ref_nc(),
                self.operation,
                TransactionBuilder::new().with_input(instance_input),
            )
            .await
            .expect("Can only fail if additional funding is needed")
    }
```
```
thread 'tokio-runtime-worker' panicked at fedimint-client/src/lib.rs:477:14:
Can only fail if additional funding is needed: Insufficient balance: requested 5 msat but only 0 msat available

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /home/jumoell/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/backtrace.rs:27:14
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/result.rs:1964:27
   2: <fedimint_mint_client::SelectNotesWithAtleastAmount as fedimint_mint_client::NotesSelector<Note>>::select_notes::{{closure}}
             at ./modules/fedimint-mint-client/src/lib.rs:1725:12
   3: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/future/future.rs:123:9
   4: fedimint_mint_client::MintClientModule::select_notes::{{closure}}
             at ./modules/fedimint-mint-client/src/lib.rs:1244:14
   5: fedimint_mint_client::MintClientModule::create_sufficient_input::{{closure}}
[...]

PR #5809, opened 2024-08-06
chore: claim_input returns result

m1sterc001guy:
Partially Fixes: https://github.com/fedimint/fedimint/issues/5598

`claim_input` currently does not return a `Result`, but it can panic if there is not sufficient funds to claim the input. This is un-intuitive and a foot gun for module developers since it will crash if the client has no funds, which might be the default for newly initialized clients.

Instead, we should return a `Result` and force the module state machines to deal with this error. Currently, this PR just uses `expect` in each module state machine, so the behavior is the same, but in the future we could add looping to retrying claiming an input when funds are available.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -231,7 +231,7 @@ pub trait IGlobalClientContext: Debug + MaybeSend + MaybeSync + 'static {
         &self,
         dbtx: &mut ClientSMDatabaseTransaction<'_, '_>,
         input: InstancelessDynClientInput,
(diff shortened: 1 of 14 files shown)


dpc (approved):

m1sterc001guy:
Should we backport this?

dpc:
@m1sterc001guy If you are planning to backport a better fix, it makes sense. Otherwise ... I guess we can live with as is?

m1sterc001guy:
No not a severe problem, and not worthy of backporting on its own IMO, but since it touched a decent amount of files it might cause conflicts for others going forward.

Issue #5787, opened 2024-08-02
Jsonrpc authentication via http headers


dpc:
https://github.com/paritytech/jsonrpsee/issues/1370#issuecomment-2264694729

Follow-up and see it this works.

dpc:
It works. Working on a PR.

dpc:
It's possible, but requires bunch of work in various parts of the system, backward compatibility changes etc. so I'm going to drop it for now. I leave #5811 as a reference.

elsirion:
With JSON RPC on the way out this doesn't seem worth it anymore.

PR #5811, opened 2024-08-07
chore: pass auth headers from http request to jsonrpc request (WIP)

dpc:
Re #5787


dpc:
I don't think it's it's worthwhile anymore.

PR #5813, opened 2024-08-08
fix: esplora bitcoin backend returning reverse network hash / wrong network

dpc:
It seems to me that what happened is that during rust-bitcoin upgrade the internal representation of `BlockHash` was reversed and now naive `hex` crate encoding gives us wrong hash, that is interpreted as regtest network. Using `.to_string` on `BlockHash` does the right thing independent of underlying order of bytes.

This bug causes DKG to fail when esplora is used as a backend.

### DIFF
--- a/fedimint-bitcoind/src/esplora.rs
+++ b/fedimint-bitcoind/src/esplora.rs
@@ -6,7 +6,6 @@ use fedimint_core::task::TaskHandle;
 use fedimint_core::txoproof::TxOutProof;
 use fedimint_core::util::SafeUrl;
 use fedimint_core::{apply, async_trait_maybe_send, Feerate};
-use hex::ToHex;
 use tracing::{info, warn};
 
 use crate::{DynBitcoindRpc, IBitcoindRpc, IBitcoindRpcFactory, RetryClient};
@@ -44,7 +43,7 @@ impl IBitcoindRpc for EsploraClient {
         let genesis_height: u32 = 0;
         let genesis_hash = self.0.get_block_hash(genesis_height).await?;
 
-        let network = match genesis_hash.encode_hex::<String>().as_str() {
+        let network = match genesis_hash.to_string().as_str() {
             crate::MAINNET_GENESIS_BLOCK_HASH => Network::Bitcoin,
             crate::TESTNET_GENESIS_BLOCK_HASH => Network::Testnet,
             crate::SIGNET_GENESIS_BLOCK_HASH => Network::Signet,


Kodylow:
Found this when doing some new 0.4.0 deployments and confirmed fix with DPC live. ACK

Kodylow (approved):

bradleystachurski (approved):
> naive `hex` crate encoding gives us wrong hash

Agreed, we fixed the reversed bytes for `Txid` but `BlockHash` snuck through.

[`8be68a6` (#3994)](https://github.com/fedimint/fedimint/pull/3994/commits/8be68a6d2c70b78f56b5258e8261a128f3f4ba6a)

Doesn't look like the `hex` crate changes landed in v0.3.3 so no need for v0.3.4.

```
nix@lenovo-m75q-ubuntu:~/fedimint$ git tag --contains c34bf770493bce8cb991d24fd4494f383888a3fa
v0.4.0
v0.4.0-rc.0
v0.4.0-rc.1
v0.4.0-rc.2
v0.4.0-rc.3
v0.4.0-rc.4
v0.4.0-rc.5
v0.4.0-rc.6
v0.4.0-testing.0
```

---
Sanity check

```
nix@lenovo-m75q-ubuntu:~/sandbox/demo$ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `/home/nix/fedimint/target-nix/debug/demo`
genesis_hash: 0x000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
bug_hash_string: "6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000"
fixed_hash_string: "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f"
```

src/main.rs
```rust
use hex::ToHex;

#[tokio::main]
async fn main() {
    let builder = esplora_client::Builder::new("https://mempool.space/api");
    let esplora_client = builder.build_async().unwrap();

    let genesis_hash = esplora_client.get_block_hash(0u32).await.unwrap();
    println!("genesis_hash: {genesis_hash:?}");

    let bug_hash_string = genesis_hash.encode_hex::<String>();
    let bug_hash_string = bug_hash_string.as_str();
    println!("bug_hash_string: {bug_hash_string:?}");

    let fixed_hash_string = genesis_hash.to_string();
    let fixed_hash_string = fixed_hash_string.as_str();
    println!("fixed_hash_string: {fixed_hash_string:?}")
}
```

Cargo.toml
```
[package]
name = "demo"
version = "0.1.0"
edition = "2021"

[dependencies]
esplora-client = { version = "=0.6.0", default-features = false, features = [
    "async",
    "async-https-rustls",
] }
tokio = { version = "1", features = ["full"] }
hex = "0.4.3"
```

bradleystachurski:
> Using `.to_string` on `BlockHash` does the right thing independent of underlying order of bytes

This will still be a lurking footgun in the future. Perhaps out of scope for this quick fix, but it's worth considering a safer api so we don't need to know when `.to_string` is needed (e.g. `BlockHash`) and when `.encode_hex::<String>` is appropriate.

Kodylow:
@bradleystachurski anything pending to land this and cut hotfix? this is borking any new mutinynet deployments with esplora

bradleystachurski:
@Kodylow Nope, let's land. If other reviewers find an issue we can address with v0.4.1-rc.1

bradleystachurski:
I'd also expect any peers using esplora that upgraded to v0.4.0 to fail to start the wallet server

https://github.com/fedimint/fedimint/blob/fd074f9219ca6bd7403d7be935fbdac2076c411f/modules/fedimint-wallet-server/src/lib.rs#L798-L807

fedimint-backports:
- #5816

apoelstra:
The internal representation of `BlockHash` has always been a [u8; 32] and has never changed its ordering.

Our hex encoding of [u8; 32] has always been in the forward direction, as has that of our new hex-conservative and of the 3rd-party `hex` crate.

elsirion:
@apoelstra I looked into it and it was just a fuckup on our side that wasn't caught. I remember discussing serialization order in a different context, but that one slipped through review.

Regarding the availability of `hex_conservative`: I think at the time we were discussing unifying our many hex implementations it might not have been ready yet and I didn't check back when someone actually picked it up. There are some annoyances with it anyway (no streaming, always allocating), so we might end up switching again, either to our own fork or `hex_conservative` which looks promising.

@bradleystachurski It's something our test suite should have caught, are we still running tests against esplora?  Maybe we only ever did on the client side since running live federations against it isn't advisable anyway?

dpc:
We're running all sorts of tests, but all with regtest network. And all invalid networks turn into regtest.

elsirion:
> ? We have iterator and read/write support

I was referring to `hex` and this being another reason to possibly switch again

PR #5814, opened 2024-08-08
chore: parallelize mprocs lightning channel opening

tvolk131:
On my local machine (Apple M1 Pro, 32GB RAM) the mprocs startup times before/after are:

Before (master branch):
Attempt 1: 93.3s
Attempt 2: 95.5s
Attempt 3: 94.0s
Attempt 4: 93.8s
Attempt 5: 93.5s
Average: 94.0s

After (PR branch):
Attempt 1: 65.2s
Attempt 2: 64.0s
Attempt 3: 63.7s
Attempt 4: 65.7s
Attempt 5: 64.3s
Average: 64.6s

So this is roughly 30% faster and helps to undo some of the damage caused by merging the LDK gateway

### DIFF
--- a/devimint/src/devfed.rs
+++ b/devimint/src/devfed.rs
@@ -10,7 +10,7 @@ use tracing::debug;
 
 use crate::envs::{FM_GWID_CLN_ENV, FM_GWID_LDK_ENV, FM_GWID_LND_ENV};
 use crate::external::{
-    open_channel, open_channel_between_gateways, Bitcoind, Electrs, Esplora, Lightningd, Lnd,
+    open_channel, open_channels_between_gateways, Bitcoind, Electrs, Esplora, Lightningd, Lnd,
 };
 use crate::federation::{Client, Federation};
 use crate::gatewayd::Gatewayd;
@@ -273,12 +273,14 @@ impl DevJitFed {
                     let gw_cln = gw_cln.get_try().await?.deref();
                     let gw_lnd = gw_lnd.get_try().await?.deref();
 
-                    open_channel_between_gateways(&bitcoind, gw_cln, gw_lnd).await?;
+                    let gateways: &[(&Gatewayd, &str)] =
+                        if let Some(gw_ldk) = gw_ldk.get_try().await?.deref() {
+                            &[(gw_cln, "CLN"), (gw_lnd, "LND"), (gw_ldk, "LDK")]
+                        } else {
+                            &[(gw_cln, "CLN"), (gw_lnd, "LND")]
+                        };
 
-                    if let Some(gw_ldk) = gw_ldk.get_try().await?.deref() {
-                        open_channel_between_gateways(&bitcoind, gw_ldk, gw_cln).await?;
-                        open_channel_between_gateways(&bitcoind, gw_ldk, gw_lnd).await?;
-                    }
+                    open_channels_between_gateways(&bitcoind, gateways).await?;
                 }
 
                 Ok(Arc::new(()))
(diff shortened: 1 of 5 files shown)


devimint/src/devfed.rs
```diff
@@ -273,12 +273,14 @@ impl DevJitFed {
...
+                            &[(gw_cln, "CLN"), (gw_lnd, "LND"), (gw_ldk, "LDK")]
+                        } else {
+                            &[(gw_cln, "CLN"), (gw_lnd, "LND")]
+                        };
 
-                    if let Some(gw_ldk) = gw_ldk.get_try().await?.deref() {
-                        open_channel_between_gateways(&bitcoind, gw_ldk, gw_cln).await?;
-                        open_channel_between_gateways(&bitcoind, gw_ldk, gw_lnd).await?;
-                    }
+                    open_channels_between_gateways(&bitcoind, gateways).await?;
```
dpc, line 283: It would probably be better to parallize N calls to open_channels_between_gateways with a different gateway pair each. Parallizing whole thing, instead of each step in the thing separately allows smoothing out differences in each step times.

devimint/src/lib.rs
```diff
@@ -50,11 +50,11 @@ where
     let (process_mgr, task_group) = cli::setup(args).await?;
     log_binary_versions().await?;
     let dev_fed = devfed::DevJitFed::new(&process_mgr, false)?;
-    let res = cleanup_on_exit(f(dev_fed.clone(), process_mgr.clone()), task_group).await;
     // workaround https://github.com/tokio-rs/tokio/issues/6463
     // by waiting on all jits to complete, we make it less likely
     // that something is not finished yet and will block in `on_block`
     let _ = dev_fed.finalize(&process_mgr).await;
+    let res = cleanup_on_exit(f(dev_fed.clone(), process_mgr.clone()), task_group).await;
```
dpc, line 57: Why is this needed? IIUC this ruins parallelism by a lot.

dpc:
https://github.com/fedimint/fedimint/pull/5826 instead?

dpc:
Well, https://github.com/fedimint/fedimint/pull/5826 doesn't work, so whatevs. ;D

devimint/src/lib.rs
tvolk131: But without this change, we're running the test (`f()`) before waiting for the federation to be finalized. Isn't that a race condition?

To answer your question, the LNv2 tests fail without this change. I suspect this is because channel opens with LDK are slow (which is why mprocs is taking so long to start) and on `master` we open a channel between CLN and LND first, allowing for a quick time-to-first-channel:

https://github.com/fedimint/fedimint/blob/cae1f0c6a75262bcd3645b8ee77cc3cbc0c2fdb5/devimint/src/devfed.rs#L273-L281

Whereas on this branch, the time-to-all-channels is quicker but the time-to-first-channel is slower since we're waiting for all channels to reach a certain step before proceeding to the next.

elsirion (approved):

elsirion:
Looks good and makes devimint go brrrrr :grin:

Kodylow:
@dpc eric said on the call you might have a concern here, can you post if so otherwise can you land?

elsirion:
Ok, too much flakiness, we shouldn't merge yet :/

elsirion:
Talking to @tvolk131: likely unrelated, just unlucky PR

dpc:
@tvolk131 

> But without this change, we're running the test (f()) before waiting for the federation to be finalized. Isn't that a race condition?

The idea was that particular tests should wait for things it actually needs, so that we don't have to pessimize in ever test.

PR #5822, opened 2024-08-09
feat: global client rpc commands

maan2003:
prep for typescript client

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -88,7 +88,7 @@ use std::sync::{Arc, Weak};
 use std::time::Duration;
 
 use anyhow::{anyhow, bail, ensure, Context};
-use async_stream::stream;
+use async_stream::{stream, try_stream};
 use backup::ClientBackup;
 use db::{
     apply_migrations_client, apply_migrations_core_client, get_core_client_database_migrations,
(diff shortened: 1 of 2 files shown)


maan2003:
see https://github.com/fedimint/fedimint/pull/5475 as reference

dpc (approved):

fedimint-client/src/lib.rs
```diff
@@ -1919,6 +1920,70 @@ impl Client {
...
+                "get_balance" => {
+                    let balance = self.get_balance().await;
+                    yield serde_json::to_value(balance)?;
+                }
+                "subscribe_balance_changes" => {
+                    let mut stream = self.subscribe_balance_changes().await;
+                    while let Some(balance) = stream.next().await {
+                        yield serde_json::to_value(balance)?;
+                    }
+                }
```
elsirion, line 1940: Won't this just block the stream forever, not allowing any other commands to be processed? Or is the idea to open new streams for such commands anyway?

maan2003: yes streams will concurrently run.

https://github.com/fedimint/fedimint/pull/5595/files#diff-df844475f10253d85671819b33bad8d7ae1fcd5da5eb1f3c773ece59a874cd42R72

javascript can call rpc function even when other rpc is not finished yet.

Issue #5699, opened 2024-07-19
Bumping cargo version fails for dev dependencies


bradleystachurski:
```
error: failed to select a version for the requirement `fedimint-dummy-common = "=0.4.0-rc.1"`
candidate versions found which didn't match: 0.4.0-rc.2
location searched: /home/nix/fedimint/modules/fedimint-dummy-common
required by package `fedimint-server v0.4.0-rc.2 (/home/nix/fedimint/fedimint-server)`
if you are looking for the prerelease package it needs to be specified explicitly
    fedimint-dummy-common = { version = "0.4.0-rc.2" }
```

This only happens for `[dev-dependencies]` in the workspace. For now, I'm manually bumping the version, however we should find a better long term solution. A simple fix is switching to `sed`.

@dpc and I spent a chunk of time on this already and it appears to be a limitation of the `cargo-workspaces` crate (https://github.com/pksunkara/cargo-workspaces).

bradleystachurski:
Interesting note from the cargo-workspaces readme

https://github.com/pksunkara/cargo-workspaces#publish

> Note: dev-dependencies are not taken into account when building the dependency graph used to determine the proper publishing order. This is because dev-dependencies are ignored by cargo publish - as such, a dev-dependency on a local crate (with a path attribute), should not have a version field.

dpc:
I used `sed -i 's/=0.4.0-alpha/=0.5.0-alpha/g' **.toml` to bump the version. I don't see a reason to use anything more fancy.

The comment about dev dependencies not needing version seems completely wrong to me.

bradleystachurski:
Capturing chat with @dpc: we should consider making versions across all projects in the workspace use `{ workspace = true }`, so bumping only requires updating the root Cargo.toml

bradleystachurski:
`sed **.toml` didn't seem to find all the Cargo.toml files recursively. This worked for me locally

```
find . -name 'Cargo.toml' -exec sed -i 's/0.4.0-rc.6/0.4.1-rc.0/g' {} +
```

dpc:
> `sed **.toml` didn't seem to find all the Cargo.toml files recursively.

That works in fish, maybe in some other shells that support better globing, but not in bash.

PR #5824, opened 2024-08-09
chore: add just command to bump versions using sed

bradleystachurski:
Closes https://github.com/fedimint/fedimint/issues/5699

Replacing the cargo workspaces command since it fails to update `[dev-dependencies]`. I used this command to bump `v0.4.0 -> v0.4.1-rc.0` and `v0.4.1-rc.0 -> v0.4.1`

### DIFF
--- a/justfile.fedimint.just
+++ b/justfile.fedimint.just
@@ -236,11 +236,10 @@ docs: build-docs
     $open_cmd "$fallback_index_path" || true
   fi
 
-release-bump-version VERSION="prerelease":
-  cargo workspaces version --all --exact --no-git-commit --yes --force '*' --pre-id rc {{VERSION}}
-
-release-bump-version-custom VERSION:
-  cargo workspaces version --all --exact --no-git-commit --yes --force '*' custom {{VERSION}}
+# Bump all versions
+# ex: `just bump-version 0.5.0-alpha 0.5.0-rc.0`
+bump-version FROM_VERSION TO_VERSION:
+  find . -name 'Cargo.toml' -exec sed -i 's/{{FROM_VERSION}}/{{TO_VERSION}}/g' {} +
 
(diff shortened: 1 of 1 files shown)


justfile.fedimint.just
```diff
@@ -236,11 +236,10 @@ docs: build-docs
...
 
-release-bump-version VERSION="prerelease":
-  cargo workspaces version --all --exact --no-git-commit --yes --force '*' --pre-id rc {{VERSION}}
-
-release-bump-version-custom VERSION:
-  cargo workspaces version --all --exact --no-git-commit --yes --force '*' custom {{VERSION}}
+# Bump all versions
+# ex: `just bump-version 0.5.0-alpha 0.5.0-rc.0`
+bump-version FROM_VERSION TO_VERSION:
+  find . -name 'Cargo.toml' -exec sed -i 's/{{FROM_VERSION}}/{{TO_VERSION}}/g' {} +
```
maan2003, line 242: what if it changes version of some other dependency :/

dpc: The build will fail and we'll see it.

Since we're not using dependencies in versions `-alpha` or `-rc.x` a lot, it's theoretically a problem, but in practice good enough.

maan2003: right we now use -alpha.0 on master.

maan2003 (approved):

justfile.fedimint.just
bradleystachurski: > right we now use -alpha.0 on master

I think we're just `-alpha` on master

> what if it changes version of some other dependency :/

Yeah, it may still be a bit of a mess since we might need to manually inspect failures if the build fails (quick grep for 0.4.1 and 0.4.2 shows we'll bump into this soon).

The ideal solution is to update everything to workspace = true so we only need to update the root Cargo.toml when we do a version bump.

PR #5826, opened 2024-08-10
chore(devimint): open channels in parallel

dpc:


dpc:
Eeehhhh, why is it so flaky...

Issue #5069, opened 2024-04-23
Don't split e-cash when aborting OOB spends


elsirion:
After splitting e-cash inputs into single notes cancelling offline sends leads to each of the original notes to be split into smaller ones due to how our output generation algorithm works. This leads to slowdowns and enormous inefficiency.

bprice6:
Just one thing we've noticed - this could help the Fedi app by improving performance / reduce slowness any time someone wants to re-issue notes to themselves.

PR #5833, opened 2024-08-10
fix: don't split e-cash on refund

elsirion:
One possible solution for #5069.

## Problem
The main problem is that if I have e-cash in an OOB spend or funding any other transaction and need to re-claim it because the user cancelled or the transaction failed then a bunch of e-cash notes will be re-issued *in parallel*. 

Let's assume 10 notes are being re-issued into an empty wallet. The reissuance state machine of each note will notice the empty wallet and generate at least two e-cash notes per denomination, starting from 1msat (to make sure the wallet has e-cash of every denomination for OOB spends). Unfortunately, all 10 do the same and we end up with at least 20 notes per denomination.

## Solution Approach
This PR introduces a strategy enum for change generation that is used to control change generation in this scenario. 

**Benefits:**
- No state machine migrations necessary
- Conceptually simple/small change

**Drawbacks:**
- API  changes to a bunch of interfaces
- The change strategy is only really relevant for e-cash primary modules, so it breaks modularization a bit

## Alternative
We want to eventually use one SM for multiple e-cash notes in cases where they all result from the same operation. This will allow to reissue multiple notes in a single transaction except in the case that there was an invalid note among them (then each still needs to go into a separate transaction). When reissued in a single transaction the parallelism problem doesn't occur anymore.

Thoughts @dpc @joschisan?

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -162,8 +162,8 @@ use crate::sm::{
     ClientSMDatabaseTransaction, DynState, Executor, IState, Notifier, OperationState, State,
 };
 use crate::transaction::{
-    tx_submission_sm_decoder, ClientInput, ClientOutput, TransactionBuilder, TxSubmissionContext,
-    TxSubmissionStates, TRANSACTION_SUBMISSION_MODULE_INSTANCE,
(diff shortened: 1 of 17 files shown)


dpc (approved):

dpc:
#6176 alternative?

elsirion:
#6176 is better, I just didn't want to change APIs back then. But doesn't look all that bad.

PR #5834, opened 2024-08-11
chore: consolidate cln create_invoice()

tvolk131:
No functional change, just extracting duplicate code between two arms of a `match` statement


Kodylow (approved):
much nicer

PR #5835, opened 2024-08-11
feat: gateway can create normal invoices

tvolk131:


gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -797,8 +847,14 @@ impl GatewayLightning for ClnRpcService {
             description,
         } = create_invoice_request.into_inner();
 
-        let payment_hash = sha256::Hash::from_slice(&payment_hash)
-            .map_err(|e| tonic::Status::internal(e.to_string()))?;
+        let payment_hash = if payment_hash.is_empty() {
+            return self
+                .create_invoice_with_known_preimage(amount_msat, expiry_secs.into(), description)
```
Kodylow, line 852: Naming might be a little weird here: this means "I know the preimage so this is a payment to myself not on behalf of a fedimint client"

Kodylow: for_self or something obviosuly "this is mine" might be better

gateway/ln-gateway/src/lib.rs
```diff
@@ -920,6 +920,29 @@ impl Gateway {
...
+    ) -> Result<Bolt11Invoice> {
+        let GatewayState::Running { lightning_context } = self.get_state().await else {
+            return Err(GatewayError::Disconnected);
+        };
+
+        Bolt11Invoice::from_str(
+            &lightning_context
+                .lnrpc
+                .create_invoice(CreateInvoiceRequest {
+                    payment_hash: Vec::new(),
```
Kodylow, line 935: add a comment here

Kodylow (approved):
lgtm, nits: change the naming scheme to have "self" to clarify when it's the gateway doing things for itself vs fedimint client actions

modules/fedimint-ln-common/src/gateway_endpoint_constants.rs
```diff
@@ -17,6 +17,7 @@ pub const OPEN_CHANNEL_ENDPOINT: &str = "/open_channel";
 pub const CLOSE_CHANNELS_WITH_PEER_ENDPOINT: &str = "/close_channels_with_peer";
 pub const RECEIVE_ECASH_ENDPOINT: &str = "/receive_ecash";
 pub const ROUTING_INFO_V2_ENDPOINT: &str = "/routing_info";
+pub const CREATE_INVOICE_ENDPOINT: &str = "/create_invoice";
```
m1sterc001guy, line 20: Let's start migrating these to the lnv2 module

tvolk131: Good point! Done.

gateway/ln-gateway/src/lib.rs
tvolk131: Done.

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -363,33 +363,43 @@ impl ILnRpcClient for GatewayLdkClient {
...
+        // Currently `ldk-node` only supports direct descriptions.
+        // See https://github.com/lightningdevkit/ldk-node/issues/325.
+        // TODO: Once the above issue is resolved, we should support
+        // description hashes as well.
+        let description_str =
+            if let Some(Description::Direct(description)) = &create_invoice_request.description {
+                description
+            } else {
+                ""
+            };
```
elsirion, line 385: We should return an error on receiving a description hash. If someone does that they expect a specific behavior that we can't provide right now. It's better to fail.

gateway/ln-gateway/src/lib.rs
```diff
@@ -920,6 +920,30 @@ impl Gateway {
...
+    ) -> Result<Bolt11Invoice> {
+        let GatewayState::Running { lightning_context } = self.get_state().await else {
+            return Err(GatewayError::Disconnected);
+        };
+
+        Bolt11Invoice::from_str(
+            &lightning_context
+                .lnrpc
+                .create_invoice(CreateInvoiceRequest {
+                    payment_hash: Vec::new(), /* Empty payment hash indicates an invoice payable
```
elsirion, line 935: Can we make this more explicit? At least an `Option<Vec<u8>>`?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -383,6 +383,56 @@ impl ClnRpcService {
...
+        expiry_secs: u64,
+        description_or: Option<Description>,
+    ) -> Result<tonic::Response<CreateInvoiceResponse>, Status> {
+        let description = match description_or {
+            None => String::new(),
+            Some(Description::Direct(description)) => description,
+            Some(Description::Hash(_hash)) => {
+                // Currently CLN only supports direct descriptions.
+                String::new()
+            }
```
elsirion, line 401: Same here: better to fail

gateway/ln-gateway/src/lib.rs
tvolk131: Nope 🙁 `CreateInvoiceRequest` is a proto message, so the [default value](https://protobuf.dev/programming-guides/proto3/#default) for bytes is an empty list of bytes.

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: Done.

gateway/ln-gateway/src/lightning/ldk.rs

PR #5836, opened 2024-08-12
feat: parallelise signature checks

joschisan:
The verification of ecash note signatures is at the moment the biggest performance bottleneck of the system as we can only verify about 500 per second per core. Performance testing on my M2 MacBook Air with 8 cores showed a 5x speed up when parallelising the verification across 10 to 50 signatures.

This pr allows us to lower the input fee of one sat per ecash not to a more reasonable 100 msats for DOS protection.

### DIFF
--- a/fedimint-core/src/core/server.rs
+++ b/fedimint-core/src/core/server.rs
@@ -52,6 +52,15 @@ pub trait IServerModule: Debug {
         peer_id: PeerId,
     ) -> anyhow::Result<()>;
 
+    // Use this function to parallelise stateless cryptographic verification of
+    // inputs across a transaction. All inputs of a transaction are verified
+    // before any input is processed.
+    fn verify_input(
+        &self,
+        input: &DynInput,
+        module_instance_id: ModuleInstanceId,
+    ) -> Result<(), DynInputError>;
+
     /// Try to spend a transaction input. On success all necessary updates will
     /// be part of the database transaction. On failure (e.g. double spend)
     /// the database transaction is rolled back and the operation will take
@@ -164,6 +173,24 @@ where
             peer_id
         )
         .await
+    }
+
+    // Use this function to parallelise stateless cryptographic verification of
+    // inputs across a transaction. All inputs of a transaction are verified
+    // before any input is processed.
+    fn verify_input(
+        &self,
+        input: &DynInput,
+        module_instance_id: ModuleInstanceId,
+    ) -> Result<(), DynInputError> {
+        <Self as ServerModule>::verify_input(
+            self,
+            input
+                .as_any()
+                .downcast_ref::<<<Self as ServerModule>::Common as ModuleCommon>::Input>()
+                .expect("incorrect input type passed to module plugin"),
+        )
+        .map_err(|v| DynInputError::from_typed(module_instance_id, v))
     }
 
     /// Try to spend a transaction input. On success all necessary updates will
--- a/fedimint-core/src/module/mod.rs
+++ b/fedimint-core/src/module/mod.rs
@@ -851,6 +851,16 @@ pub trait ServerModule: Debug + Sized {
         peer_id: PeerId,
     ) -> anyhow::Result<()>;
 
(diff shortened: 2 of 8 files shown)


fedimint-core/src/core/server.rs
```diff
@@ -52,6 +52,12 @@ pub trait IServerModule: Debug {
         peer_id: PeerId,
     ) -> anyhow::Result<()>;
 
+    fn verify_input(
+        &self,
+        input: &DynInput,
+        module_instance_id: ModuleInstanceId,
+    ) -> Result<(), DynInputError>;
```
elsirion, line 59: Unrelated to the PR since `process_input` does the same: why do we have `module_instance_id` as an argument? `DynInput` already contains it:

```rust
#[doc = " An owned, immutable input to a [`Transaction`](fedimint_core::transaction::Transaction)"]
pub struct DynInput {
    module_instance_id: ::fedimint_core::core::ModuleInstanceId,
    inner: Box<dyn IInput + 'static + Send + Sync>,
}
```

I guess it's a sanity-check and supposed to be set to the *module's* module instance id?  In that case we should probably just use the same macro for `DynServerModule` as for `DynInput`, that includes the instance id in the dyn struct.

fedimint-core/src/module/mod.rs
```diff
@@ -851,6 +851,13 @@ pub trait ServerModule: Debug + Sized {
         peer_id: PeerId,
     ) -> anyhow::Result<()>;
 
+    fn verify_input(
+        &self,
+        _input: &<Self::Common as ModuleCommon>::Input,
+    ) -> Result<(), <Self::Common as ModuleCommon>::InputError> {
+        Ok(())
+    }
```
elsirion, line 859: This needs *a lot* more explanation. The semantics as I understand them:

* Anything that can be checked about an input as a pure function (i.e. *without* reading from the DB/any global state) and is computationally heavy enough to matter should be checked using `verify_input`.
  * It is guaranteed that `verify_input` is run before `process_input` and that `process_input` is not run if it fails.
  * For most modules `verify_input` can be ignored since their stateless checks are supposedly fast compared to their stateful check.
* Anything that interacts with the DB (e.g. spent-ness checks for e-cash, account balance checks for LN, …) still has to be checked by `process_input` since that runs serially inside the database transaction and can ensure there are no race conditions etc.

**Doing stateful checks inside `verify_input` will nearly always lead to inflation bugs, so should be avoided at all cost!**

fedimint-server/src/consensus/transaction.rs
```diff
@@ -19,6 +20,21 @@ pub async fn process_transaction_with_dbtx(
...
+    // error is not extensible at the moment we need to incorrectly return the
+    // InvalidWitnessLength variant.
+    transaction
+        .inputs
+        .clone()
+        .into_par_iter()
+        .try_for_each(|input| {
+            modules
+                .get_expect(input.module_instance_id())
+                .verify_input(&input, input.module_instance_id())
```
elsirion, line 34: The way we use the `module_instance_id` argument here is absolutely useless (could just call that fn inside `verify_input`).

elsirion (approved):
Great improvement! Just needs some docs so module authors don't shoot themselves in the foot by having a DB reference in their module and reading from it in `verify_input`.

joschisan:
@elsirion added docs and a test that verify_input is actually run.

elsirion (approved):

fedimint-server/src/consensus/transaction.rs
```diff
@@ -19,6 +20,21 @@ pub async fn process_transaction_with_dbtx(
         CONSENSUS_TX_PROCESSED_OUTPUTS.observe(out_count as f64);
     });
 
+    // We can not return the error here as errors are not returned in a specified
```
dpc, line 23: @joschisan I have a crate that solves the problem of ordering. Also https://github.com/rayon-rs/rayon/issues/210 is worth looking at.

https://crates.io/crates/pariter

PR #5837, opened 2024-08-12
fix: add transaction fee

joschisan:
A transaction fee prevents the malicious use of 1 input ecash transactions to prevent parallelisation of ecash signature checks within a transaction making it cheaper to compute DOS the federation with valid transaction. Therefore, it is sufficient to parallelise signature checks within a transaction as a regular tx has enough mint inputs for effective use of multiple cores. 

This is the pr that parallelises signature checks within a transaction: https://github.com/fedimint/fedimint/pull/5836 

Since we cannot set the fee to a non-zero value as long as we run ln legacy I tested a tx fee of 1 sat manually with lnv2.

### DIFF
--- a/fedimint-client/src/db.rs
+++ b/fedimint-client/src/db.rs
@@ -3,7 +3,10 @@ use std::io::Cursor;
 use std::time::SystemTime;
 
 use fedimint_api_client::api::ApiVersionSet;
-use fedimint_core::config::{ClientConfig, ClientConfigV0, FederationId, GlobalClientConfig};
+use fedimint_core::config::{
+    ClientConfig, ClientConfigV0, ClientConfigV1, FederationId, GlobalClientConfig,
+    GlobalClientConfigV1, TransactionFee,
+};
 use fedimint_core::core::{ModuleInstanceId, OperationId};
 use fedimint_core::db::{
     apply_migrations, create_database_version, CoreMigrationFn, Database, DatabaseTransaction,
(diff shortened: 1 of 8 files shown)


fedimint-client/src/db.rs
```diff
@@ -416,20 +428,45 @@ pub fn get_core_client_database_migrations() -> BTreeMap<DatabaseVersion, CoreMi
...
+                    consensus_version: config_v1.global.consensus_version,
+                    meta: config_v1.global.meta,
+                },
+                modules: config_v1.modules,
+            };
+
+            dbtx.insert_entry(&ClientConfigKey, &config)
+                .await
+                .expect("We removed the client config v1");
+
```
elsirion, line 469: I think in retrospect it was a mistake to patch in `broadcast_public_keys` into the client config. As discussed with @dpc we should add new config as API endpoints and sync it to the client. This has the additional benefit of allowing us to update values by re-fetching it from time to time.

elsirion (changes requested):
Imo the proper way of introducing this, as discussed on our call, would be:

* Create static `tx_fees` API endpoint, that always returns 0 (we can only turn it on in 2 versions)
* Implement syncing to client DB from that endpoint, ideally in a generic enough way that will allow us to treat module fees the same way in another PR
* Use the (currently 0) value in the client, make sure in tests that setting it to non-zero works with the current client
* Eventually implement the server side that allows setting a non-zero fee (likely through some consensus voting)

joschisan:
@elsirion I would prefer to merge this as is. I have no plans to make the transaction fee in the consensus dynamic as it will be negligible in regular use, especially compared to lightning fees, therefore this value is just as static as many other values currently in the client config. If you want to turn the client config into an endpoint at a later point be my guest, I am not contesting that, however, doing it with this pr is not the right place. This pr is a consensus change that needs to be supported by the client and therefore not the place to change client behaviour like how config values are fetched.

elsirion:
By going the API path a lot of the client changes become redundant, hence my hesitancy to merge as-is. If we make the fee a new DB entry we won't need the DB migration anymore, but do we want to remove it again once it was on master? We could, it's just a lot of back and forth that will mess up people's DBs in the process.

Issue #5805, opened 2024-08-06
Update codepaths for v0.3 MSFV (Minimum Supported Fedimint Version)


bradleystachurski:
Now that we've cut v0.4.0, v0.3.0 is the MSFV (see: https://github.com/fedimint/fedimint/discussions/3950#discussioncomment-7862690).

- [ ] remove all `TODO(support:v0.2*)` codepaths
- [ ] update back-compat to use v0.3.0 as MSFV

bradleystachurski:
Dev call:
- we may want to support ~1 year which is closer to 4 versions
- would also be nice to remove cruft from test suite

elsirion:
We did that up to 0.4 by now

PR #5838, opened 2024-08-12
feat: set non-zero input fee for mint module

joschisan:
The ability for the mint module to handle non-zero input fees in finalisation was shipped in 0.3.0. Therefore, we can set a non-zero default of 100 msats for a mint input for 0.5.0. Since we parallelised the signature checks in https://github.com/fedimint/fedimint/pull/5836 100 msats should be sufficient to account for the computational load.


elsirion (approved):
This will
a) break 0.2 clients (see https://github.com/fedimint/fedimint/issues/5805)
b) not be available to already-running federations, we should implement dynamic fees (https://github.com/fedimint/fedimint/issues/4751)

Good start to test it out though and get some DoS deterrence in for new federations in the meantime.

dpc (commented):
:+1:

elsirion:
We should look into more fine grained testing for relevant amount math before introducing fuzzy amount checks.

PR #5850, opened 2024-08-13
chore: `TieredMulti` refactor

tvolk131:


fedimint-core/src/tiered_multi.rs
```diff
@@ -204,68 +163,12 @@ where
...
-
-    fn next(&mut self) -> Option<Self::Item> {
-        let mut notes = Vec::with_capacity(self.iters.len());
-        let mut amount = None;
-        for iter in &mut self.iters {
-            match iter.next() {
-                Some((amt, note)) => {
-                    if let Some(amount) = amount {
-                        // This may fail if notes weren't tested for structural equality
-                        assert_eq!(amount, amt);
```
tvolk131, line 249: Note for reviewers: This assert is effectively removed in `new_aggregate_from_tiered_iter()` (which replaces `TieredMultiZip`). Does anyone know if this is a problem anywhere where `new_aggregate_from_tiered_iter()` is called?

elsirion: I think we will see weird failure modes if the supplied iterators over tiered aren't structurally equivalent, i.e. don't contain the same amount tiers. 

I like having a second line of defense, we could achieve something similar here by doing:

```rust
assert!(
    tiered_multi.summary()
        .iter()
        .map(|(_tier, count)| count)
        .all_equal(),
    "The supplied Tiered structs were not structurally equal"
);
```

([`all_equal` coming from `itertools`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.all_equal))

elsirion (commented):
Looks good, maybe worth adding the `assert` again

fedimint-core/src/tiered_multi.rs
tvolk131: Done. I added a TODO mentioning that this only asserts the shape of the _output_, not the input. Technically a regression in the strictness of this assertion, but I don't think it will likely matter.

fedimint-core/src/tiered_multi.rs
```diff
@@ -106,37 +88,6 @@ impl<T> TieredMulti<T> {
             .flat_map(|(amt, notes)| notes.into_iter().map(move |c| (amt, c)))
     }
 
-    /// Returns the length of the longest vector of all tiers, ignoring the
-    /// `except` tier
-    pub fn longest_tier_except(&self, except: &Amount) -> usize {
```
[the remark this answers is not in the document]
tvolk131: 😆 No idea

fedimint-core/src/tiered_multi.rs
```diff
@@ -112,15 +112,11 @@ impl<C> IntoIterator for TieredMulti<C>
...
+    type Item = (Amount, Vec<C>);
+    type IntoIter = Box<dyn Iterator<Item = (Amount, Vec<C>)> + Send>;
 
     fn into_iter(self) -> Self::IntoIter {
-        Box::new(
-            self.0
-                .into_iter()
-                .flat_map(|(amt, notes)| notes.into_iter().map(move |c| (amt, c))),
-        )
+        Box::new(self.0.into_iter())
```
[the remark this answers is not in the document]
tvolk131: Done.

fedimint-core/src/tiered_multi.rs
elsirion: Uff, you are right, I'm dumb :see_no_evil:

elsirion: At least now all or none of the operations following on the `TieredMulti` will fail, but this isn't what I had in mind :/ don't want to unnecessarily block this, it likely doesn't matter indeed.

elsirion (approved):

PR #5851, opened 2024-08-13
Lnv2 extensible outcome

joschisan:


elsirion (approved):
Breaking change to LNv2, but hasn't been rolled out yet, so that's ok.

Issue #4446, opened 2024-03-01
Lightning Gateway backupability


elsirion:
Do we have a common seed for all the clients in the gateway and can we export it/recover from it? If not we should.

justinmoon:
dev call: @joschisan reports that every client different seeds. Also, we use 64 bytes which can't be represented as BIP39 seed words.

kernelkind:
sounds related: https://github.com/fedimint/fedimint/issues/4170

justinmoon:
Also related to https://github.com/fedimint/fedimint/issues/2652. These 3 could probably be one issue.

Kodylow:
Working this rn with @m1sterc001guy , moving to bip39 mnemonic and reusing all dpc's fanciness in the fedimint-client with a manual migration: gateway operator moves their ecash out, leaves the federation, rejoins with the automatically created new client derived from the mnemonic

joschisan:
If you do a manual migration you can also move all dbs into one via prefixes.

m1sterc001guy:
> If you do a manual migration you can also move all client dbs into one via prefixes.

Yep, had the same idea, this is the plan.

Issue #5839, opened 2024-08-12
Lightning Module Backupability


Kodylow:
Separate from gatway root secret backups, we should be able to backup and recover the state for outstanding Lightning Contracts.

@elsirion

Kodylow:
little more complex, need to write additional code for v2 module

Kodylow:
stateless recovery yay

elsirion:
TODO: check if this already exists or not

PR #5852, opened 2024-08-13
feat: Gateway Mnemonic and Common Database

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/4446

Newly joined federations use clients that are created from a common mnemonic. This mnemonic is generated when the gateway boots. "Legacy" federations are still supported and can be upgraded by leaving, deleting the database, then re-joining. A test for this has been added and an automated endpoint for doing this can be added in a follow up.

This PR also cleans up the old standing issue of having separate databases per joined federation inside the gateway.

Note: This only provides backup-ability of the ecash of each joined federation. The recover API still needs to be added, as well as recovering lightning contracts as described  https://github.com/fedimint/fedimint/issues/5839


Kodylow:
it's so beautiful 😍

m1sterc001guy:
Blocked on https://github.com/fedimint/fedimint/pull/5871

devimint/src/devfed.rs
```diff
@@ -376,12 +375,6 @@ impl DevJitFed {
             "too many offline nodes ({offline_nodes}) to reach consensus"
         );
 
-        std::env::set_var(FM_GWID_CLN_ENV, self.gw_cln().await?.gateway_id().await?);
-        std::env::set_var(FM_GWID_LND_ENV, self.gw_lnd().await?.gateway_id().await?);
-        if let Some(gw_ldk) = self.gw_ldk().await? {
-            std::env::set_var(FM_GWID_LDK_ENV, gw_ldk.gateway_id().await?);
-        }
```
m1sterc001guy, line 383: For reviewers: these are not used anywhere meaningful and they trigger a polling loop by contacting each gateway. Just not necessary.

m1sterc001guy:
Had a quick chat with @bradleystachurski and he explained what I was doing wrong with the new mnemonic test. Now the mnemonic test runs as part of the upgrade tests. After the next release we can probably remove it, as we will have sufficient confidence that this upgrade path works.

Upgrade tests pass: https://github.com/fedimint/fedimint/actions/runs/10479242969/job/29024449941?pr=5852

gateway/ln-gateway/src/lib.rs
```diff
@@ -1433,10 +1462,39 @@ impl Gateway {
...
 
+    async fn load_or_generate_mnemonic(&self) -> Result<Mnemonic> {
+        Ok(
+            if let Ok(entropy) =
+                Client::load_decodable_client_secret::<Vec<u8>>(&self.gateway_db).await
+            {
+                Mnemonic::from_entropy(&entropy)
+                    .map_err(|e| GatewayError::ClientCreationError(e.to_string()))?
+            } else {
+                let mnemonic = if let Ok(words) = std::env::var(FM_GATEWAY_MNEMONIC_ENV) {
```
m1sterc001guy, line 1473: Currently the only way to recover from a mnemonic is to delete the database or recover using a new gateway with a different data directory. Is that ok from a user experience perspective? Thoughts?

gateway/ln-gateway/src/client.rs
```diff
@@ -57,49 +77,72 @@ impl GatewayClientBuilder {
...
-            let secret = PlainRootSecretStrategy::random(&mut thread_rng());
-            Client::store_encodable_client_secret(client_builder.db_no_decoders(), secret)
-                .await
-                .map_err(GatewayError::ClientStateMachineError)?;
-            secret
+            // Use the mint channel id as the prefix for the database because it is a short,
+            // unique identifier
+            let db = gateway
+                .gateway_db
+                .with_prefix(config.mint_channel_id.to_le_bytes().to_vec());
```
m1sterc001guy, line 115: Looking for some feedback on this too. I used the `mint_channel_id` here because it was smaller than `federation_id` and it is also unique, but ideally I'd like to remove this when we remove LNv1. Maybe `FederationIdPrefix` is better?

Kodylow: federationIdPrefix is definitely better

gateway/ln-gateway/src/lib.rs
Kodylow: This seems fine for now, recovery in separate PR

gateway/integration_tests/src/main.rs
```diff
@@ -13,24 +20,186 @@ use tracing::info;
...
+        gateway_type: LightningNodeType,
+        #[arg(long)]
+        old_gateway_cli_path: PathBuf,
+        #[arg(long)]
+        new_gateway_cli_path: PathBuf,
+        #[arg(long)]
+        old_gateway_cln_extension_path: PathBuf,
+        #[arg(long)]
+        new_gateway_cln_extension_path: PathBuf,
+    },
```
Kodylow, line 46: you said this is gonna be temporary until we phase out the legacy gateway multi dir structure? can we add a todo tagging it and we'll add to the tracking issue of things to phase out?

gateway/integration_tests/src/main.rs
```diff
@@ -13,24 +20,186 @@ use tracing::info;
...
+                old_gatewayd_path,
+                new_gatewayd_path,
+                gateway_type,
+                old_gateway_cli_path,
+                new_gateway_cli_path,
+                old_gateway_cln_extension_path,
+                new_gateway_cln_extension_path,
+            )
+            .await
+        }
```
Kodylow, line 73: same here, tag TODO for removing

Kodylow (commented):
lgtm can you add instructions for local testing? also think we'll need to surface the mnemonic behind a protected path on the rpc server so I can reveal in the gateway for when someone creates it locally and they want to back it up

m1sterc001guy:
> can you add instructions for local testing?

Sure. Running the upgrade tests should be sufficient to verify it works (`./scripts/tests/upgrade-test.sh v0.4.0`). Version here doesn't matter for the mnemonic test. You can also test it in mprocs (`gateway-lnd seed`) to see the seed phrase for the various gateways.

> also think we'll need to surface the mnemonic behind a protected path on the rpc server so I can reveal in the gateway for when someone creates it locally and they want to back it up

The `/mnemonic` path is already a protected path. You can verify in mprocs.

gateway/integration_tests/src/main.rs
m1sterc001guy: The test is really only needed until the next release, at which point we've essentially verified that this upgrade path works. We don't need to re-run a test with old code over and over again. Yes, I will add a comment todo tag

modules/fedimint-ln-common/src/gateway_endpoint_constants.rs
```diff
@@ -6,6 +6,7 @@ pub const BALANCE_ENDPOINT: &str = "/balance";
 pub const CONFIGURATION_ENDPOINT: &str = "/config";
 pub const CONNECT_FED_ENDPOINT: &str = "/connect-fed"; // uses `-` for backwards compatibility
 pub const GATEWAY_INFO_ENDPOINT: &str = "/info";
+pub const GATEWAY_MNEMONIC_ENDPOINT: &str = "/mnemonic";
```
Kodylow, line 9: oh sorry didn't see you'd added this already

gateway/integration_tests/src/main.rs
m1sterc001guy: Added TODO

gateway/ln-gateway/src/client.rs
m1sterc001guy: Changed to use `FederationIdPrefix`

devimint/src/gatewayd.rs
```diff
@@ -228,6 +229,25 @@ impl Gatewayd {
...
     }
 
+    pub async fn get_mnemonic(&self) -> Result<MnemonicResponse> {
+        let value = retry(
+            "Getting gateway mnemonic",
+            backoff_util::aggressive_backoff(),
+            || async { cmd!(self, "seed").out_json().await },
+        )
+        .await
+        .context("Getting gateway mnemonic")?;
```
tvolk131, line 239: Can you explain why this needs to be wrapped in a `retry()` call?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -187,6 +188,7 @@ fn v1_routes(gateway: Arc<Gateway>) -> Router {
         .route(LIST_ACTIVE_CHANNELS_ENDPOINT, get(list_active_channels))
         .route(GET_BALANCES_ENDPOINT, get(get_balances))
         .route(SPEND_ECASH_ENDPOINT, post(spend_ecash))
+        .route(GATEWAY_MNEMONIC_ENDPOINT, get(mnemonic))
```
tvolk131, line 191: Phew, good thing this is authenticated 😮‍💨

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -229,3 +229,9 @@ pub struct FederationBalanceInfo {
     pub federation_id: FederationId,
     pub ecash_balance_msats: Amount,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct MnemonicResponse {
+    pub mnemonic: Vec<String>,
+    pub legacy_federations: Vec<FederationId>,
```
tvolk131, line 236: Can you add a comment explaining what `legacy_federations` does and why it exists?

modules/fedimint-ln-common/src/gateway_endpoint_constants.rs
tvolk131, line 9: Nit: Is the `GATEWAY_` prefix necessary here? The file is prefixed with `gateway_` already.

fedimint-core/src/config.rs
```diff
@@ -370,6 +370,12 @@ impl FromStr for FederationIdPrefix {
     }
 }
 
+impl FederationIdPrefix {
+    pub fn as_bytes(&self) -> Vec<u8> {
```
dpc, line 374: Idiomatically, if it allocates, it should be `to_`. `as_` is only for near-zero-cost conversions.

dpc (commented):
I looked through it very lightly and LGTM.

gateway/ln-gateway/src/client.rs
```diff
@@ -57,49 +77,70 @@ impl GatewayClientBuilder {
...
+        let mut legacy_federations = Vec::new();
+        for federation_id in all_federations {
+            let db_path = self.work_dir.join(format!("{federation_id}.db"));
+            if db_path.exists() {
+                legacy_federations.push(federation_id);
+            }
+        }
+
+        legacy_federations
+    }
```
joschisan, line 145: Why not filter an iterator here?

devimint/src/gatewayd.rs
m1sterc001guy: In the test, this function is used right after restarting/upgrading the gateway. To avoid flakiness (i.e requesting the mnemonic when the webserver isnt up yet), it is easiest to just wrap in a retry.

gateway/ln-gateway/src/client.rs
m1sterc001guy: Done

fedimint-core/src/config.rs
m1sterc001guy: Done

modules/fedimint-ln-common/src/gateway_endpoint_constants.rs
m1sterc001guy: Done

gateway/ln-gateway/src/rpc/mod.rs
m1sterc001guy: Done

tvolk131 (dismissed):
LGTM! Should get another approval from someone with more familiarity before submitting

m1sterc001guy:
Ah shoot, newly added `gatewayd-mnemonic` test is failing. I will fix before trying to merge.

m1sterc001guy:
> Ah shoot, newly added gatewayd-mnemonic test is failing. I will fix before trying to merge.

Likely related to https://github.com/fedimint/fedimint/issues/5948

bradleystachurski:
> > Ah shoot, newly added gatewayd-mnemonic test is failing. I will fix before trying to merge.
> 
> Likely related to #5948

I still don't know the root cause, but reverting this commit fixed the upgrade tests locally. You can try a revert on your end to see if it's the same issue.

https://github.com/fedimint/fedimint/commit/11d61ad2046468acaf326dd63430e7b64906a62c

fedimint-core/src/util/mod.rs
```diff
@@ -440,8 +440,8 @@ where
                 } else {
                     warn!(
                         target: LOG_CORE,
-                        %error,
-                        %attempts,
+                        ?error,
+                        ?attempts,
```
elsirion, line 444: For `error` the `Debug` printing makes sense, less so for `attempts`.

gateway/ln-gateway/src/client.rs
```diff
@@ -36,19 +41,34 @@ impl GatewayClientBuilder {
...
+        let client_secret =
+            if let Ok(secret) = Client::load_decodable_client_secret::<[u8; 64]>(db).await {
+                secret
+            } else {
+                info!("Generating secret and writing to client storage");
+                let secret = PlainRootSecretStrategy::random(&mut thread_rng());
+                Client::store_encodable_client_secret(db, secret)
+                    .await
+                    .map_err(GatewayError::ClientStateMachineError)?;
+                secret
```
elsirion, line 54: Do we still need this branch in `client_plainrootsecret`? I'd imagine all new clients use the global secret, which is a BIP39 one and not plain?

gateway/ln-gateway/src/client.rs
```diff
@@ -57,49 +77,68 @@ impl GatewayClientBuilder {
...
-            let secret = PlainRootSecretStrategy::random(&mut thread_rng());
-            Client::store_encodable_client_secret(client_builder.db_no_decoders(), secret)
-                .await
-                .map_err(GatewayError::ClientStateMachineError)?;
-            secret
+            let db = gateway
+                .gateway_db
+                .with_prefix(federation_id.to_prefix().to_bytes());
+            let root_secret = Bip39RootSecretStrategy::<12>::to_root_secret(mnemonic);
+            (db, root_secret)
```
elsirion, line 115: We directly return the root secret here, but should be deriving a per-federation secret as described in the [secret derivation docs](https://github.com/fedimint/fedimint/blob/master/docs/secret_derivation.md).

The current version isn't *horribly* broken because out of abundance of caution clients internally derive a key based on the federation id, but to be compatible with other wallet tools using the standard derivation path would be nice.

gateway/ln-gateway/src/lib.rs
elsirion: In a follow-up PR it would be nice to make setting the mnemonic an API call.

gateway/ln-gateway/src/client.rs
m1sterc001guy: Hmm interesting, in early discussions with @dpc I thought the client was already deriving a per-federation secret using the federation id, and that adding it here would actually be doing the derivation twice unnecessarily. I actually originally had it like that, but ended up removing the derivation after our discussion. 

Is it recommended to derive a per-federation secret here?

gateway/ln-gateway/src/client.rs
m1sterc001guy: Correct, not necessary, good catch. Removed.

fedimint-core/src/util/mod.rs
m1sterc001guy: Changed attempts back

gateway/ln-gateway/src/client.rs
elsirion: We are doing the derivation twice if following the standard, so if an integrator doesn't follow the standard then they don't shoot themselves (or rather their users) in the foot. 

The reasoning was: 
1. It doesn't degrade security 
2. The time it takes is negligible
3. Integrators won't know about the second client-internal derivation, so it won't feel weird to them

m1sterc001guy: Fixed

gateway/integration_tests/src/main.rs
```diff
@@ -13,24 +20,189 @@ use tracing::info;
...
     }
 }
 
+/// TODO(v0.5.0): We do not need to run the `gatewayd-mnemonic` test from v0.4.0
```
bradleystachurski, line 77: [nit] It might be useful to duplicate this TODO in `upgrade-test.sh`

gateway/ln-gateway/src/client.rs
```diff
@@ -57,49 +64,83 @@ impl GatewayClientBuilder {
...
         } else {
-            info!("Generating secret and writing to client storage");
-            let secret = PlainRootSecretStrategy::random(&mut thread_rng());
-            Client::store_encodable_client_secret(client_builder.db_no_decoders(), secret)
-                .await
-                .map_err(GatewayError::ClientStateMachineError)?;
-            secret
+            let db = gateway
+                .gateway_db
+                .with_prefix(federation_id.to_prefix().to_bytes());
```
m1sterc001guy, line 100: Use `mint_channel_id` here, also add check to make sure we don't re-use a database for multiple clients

m1sterc001guy: Done. Also added a check that verifies the stored `ClientConfig`'s federation ID matches the federation ID we are creating. This should catch any accidental re-use of databases.

elsirion (approved):

fedimint-backports:

```bash
git worktree add -d .worktree/backport-5852-to-releases/v0.4 origin/releases/v0.4
cd .worktree/backport-5852-to-releases/v0.4
git switch --create backport-5852-to-releases/v0.4
git cherry-pick -x 13df2f0682cdf74073561085465d17734d1b5c69 532c66e8a5c74fdf6d7eb3eea967cb82024d6bfa 81f769f5761b2b2f78565078b673d11702e236d0 128d475986762c5c547f3f01fba5b182ef6fd6bd a0ddae713dc493171ad3b27705b224268fb6aa33
```

PR #5855, opened 2024-08-14
feat: enable modules to vote on their consensus version

joschisan:


fedimint-server/src/consensus/engine.rs
```diff
@@ -753,6 +757,63 @@ impl ConsensusEngine {
...
+                    ensure!(
+                        vote > current_vote,
+                        "Core consenus version vote is redundant"
+                    );
+
+                    dbtx.insert_entry(&CoreConsensusVersionVoteKey(peer_id), &vote)
+                        .await;
+
+                    assert!(
+                        self.consensus_core_consensus_version(dbtx).await <= CORE_CONSENSUS_VERSION,
```
bradleystachurski, line 776: Should we do this check before inserting the peer's vote in the db?

Kodylow:
from call: structure is in place and defined, just waiting on when to actually merge.

@dpc needs this to sleep at night though so hopefully soon.

PR #5857, opened 2024-08-14
feat(tor): put `tor` support behind a feature, use as default

oleonardolima:
related to #4752 

## Description

The usage of `arti-client` for adding Tor support (#5443) recently landed on master, but `arti-client` relies on a bunch of other tor related crates, which can lead to an increase on the binary size.

In order to allow users to have the opt-in/opt-out of such feature and the leading binary increase, this PR adds a new feature where required for Tor support.

## Note for Reviewers

I'm wondering if we should have it by default, as its removal from default-features becomes a SemVer incompatible change.

Any other thoughts on it?

### DIFF
--- a/fedimint-api-client/Cargo.toml
+++ b/fedimint-api-client/Cargo.toml
@@ -12,6 +12,7 @@ repository = "https://github.com/fedimint/fedimint"
 development = ["tokio-test"]
 
 [features]
+default = ["tor"]
 tor = ["dep:strum", "dep:curve25519-dalek", "arti-client/tokio", "arti-client/rustls", "arti-client/onion-service-client"]
 
 [lib]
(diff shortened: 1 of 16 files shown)


elsirion:
> I'm wondering if we should have it by default, as its removal from default-features becomes a SemVer incompatible change.

That's ok imo, we still have SemVer breaks regularly and opting out of Tor support should be something very explicit.

elsirion (approved):
I tested `cargo build -p fedimint-client --no-default-features` and that works. `cargo build --workspace --all-targets --no-default-features` fails to compile the gateway for some reason, but I doubt anyone is explicitly interested in a gateway without tor support.

oleonardolima:
> I tested `cargo build -p fedimint-client --no-default-features` and that works. `cargo build --workspace --all-targets --no-default-features` fails to compile the gateway for some reason, but I doubt anyone is explicitly interested in a gateway without tor support.

It's weird, I'll test it again and take a better look into it. 🤔

oleonardolima:
It fails on the `fedimint-gateway-cli` when running the `cargo build --workspace --all-targets --no-default-features` command, but it works fine when specifically running `cargo build -p fedimint-gateway-cli --no-default-features`.

PR #5823, opened 2024-08-09
feat: add gateway rpc + cli command to get balances

tvolk131:
This command lists the following:

* Total amount of on-chain bitcoin
* Local lightning liquidity
* eCash across all federations

The primary purpose of this command is to provide a liquidity overview in the gateway UI

PR #5858, opened 2024-08-14
chore: remove gateway balance command

tvolk131:
The `balance` command gets the e-cash balance for a given federation by its id. This is no longer needed since we added a more comprehensive `get-balances` command in #5823


m1sterc001guy:
> not implemented: FakeLightningTest does not support getting balances

just implement this and use zero or something for the on-chain and LN balances

m1sterc001guy:
I bundled this improvement into https://github.com/fedimint/fedimint/pull/6270

PR #5859, opened 2024-08-14
chore: FakeLightningTest doesn't panic for unimplemented fns

tvolk131:


elsirion (approved):

elsirion:
Not sure this was a problem, but will likely lead to better error reporting.

PR #5861, opened 2024-08-14
chore: make Gateway Debug impl more complete

tvolk131:


gateway/ln-gateway/src/db.rs
```diff
@@ -471,7 +471,6 @@ mod fedimint_migration_tests {
             federation_id,
             None,
         );
-        let connector = Connector::default();
```
elsirion, line 474: @oleonardolima this looks like an artifact from your Tor PR that was probably created by a rebase? Seems like we can just remove. I'm wondering why CI didn't catch it as unused code?

elsirion (approved):

PR #5862, opened 2024-08-14
feat: LDK gateway supports setting an LSP for LSPS2 (JIT Channels)

tvolk131:


tvolk131:
After talking with tnull, it sounds like LSPS2 (JIT Channels) isn't a good fit for the lightning gateway use-case. This is because one of the two [trust models of LSPS2](https://github.com/BitcoinAndLightningLayerSpecs/lsp/blob/main/LSPS2/README.md#trust-models) states:

> The LSP trusts the client to actually claim their payment once the channel is opened and the payment is forwarded.

Since this is in the hands of the fedimint user who initiated the inbound payment rather than the gateway itself, the gateway could involuntarily breach the trust of the LSP.

LSPS1 (Channel Request) appears to be more well-suited for the gateway. Channels can be requested from the LSP at any time, without needing an inbound payment.

'm hesitant to close this PR just yet. I'll keep it as a draft for now in case we might want LSPS2 support for the gateway operator to send funds from elsewhere to the gateway.

elsirion:
LSPS1 is what we actually want

PR #5873, opened 2024-08-16
fix: gateway has all underscore_separated endpoints

tvolk131:
~~Rename `CONNECT_FED_ENDPOINT` to `CONNECT_FED_OLD_ENDPOINT` and `LEAVE_FED_ENDPOINT` to `LEAVE_FED_OLD_ENDPOINT` (both for handler and client usage). Then add `CONNECT_FED_ENDPOINT` and `LEAVE_FED_ENDPOINT` which use underscores instead of dashes in their endpoint values. Set the gateway server to serve both the new (underscored) and old (dashed) endpoints using the same handlers. This way, once backwards compatibility is dropped in the future, we can switch the client to use the new endpoints.~~

Change `CONNECT_FED_ENDPOINT` and `LEAVE_FED_ENDPOINT` to use underscores instead of dashes, to match the rest of the endpoints


m1sterc001guy:
I don't think we care about backwards compatibility between `gateway-cli`, `gatewayd`, and `gateway-cln-extension`. All three of these binaries should be run by the same operator, if one of them is off they should realize this and upgrade.

https://github.com/fedimint/fedimint/pull/5871 is making the last bit of changes to devimint that should use the same version for all of these binaries across the backwards compatibility tests and the upgrade tests.

I'd say just fix the endpoints, don't keep the old ones around.

m1sterc001guy:
I would like to wait to merge until the upgrade tests are working again, just to be sure we don't introduce additional regressions

elsirion:
> I don't think we care about backwards compatibility between gateway-cli, gatewayd, and gateway-cln-extension. All three of these binaries should be run by the same operator, if one of them is off they should realize this and upgrade.

While we don't commit to stability yet, we should still consider cost/benefit. @Kodylow whyt do you think? Is it wort the changes to the Gateway UI?

m1sterc001guy:
Think this should be good to go unless @Kodylow has any objections

Kodylow:
Send it

elsirion (dismissed):

elsirion:
Something is weird here, how can such an easy change seemingly amplify flakiness?

m1sterc001guy:
> Something is weird here, how can such an easy change seemingly amplify flakiness?

```
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:05 2024-08-20T13:16:49.131172Z  INFO tests: Testing inter federation payment from LDK -> CLN
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:05 2024-08-20T13:16:49.951041Z  INFO fm::task: signal received, starting graceful shutdown
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06 2024-08-20T13:16:50.967509Z  INFO devimint::cli: Received shutdown signal before finishing main process, exiting early
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06 2024-08-20T13:16:50.969748Z  WARN fm::devimint: failed to terminate lightningd plugins: command: lightning-cli --network regtest --lightning-dir=/build/lnv2_module-3dkw/devimint-15741-419/cln plugin stop gateway-cln-extension
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06 Caused by:
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06     exit status: 2
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06     stdout:
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06     stderr:
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06     lightning-cli: Connecting to 'lightning-rpc': Connection refused
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06
fedimint-test-all-wasm32-unknown-unknown-ci> 00:04:06
fedimint-test-all-wasm32-unknown-unknown-ci> ## FAIL: lnv2_module
```

From the logs, it appears as though CLN is crashing during the inter-federation test of lnv2, since it fails to kill the extension. I doubt its related to this PR but its hard to tell why. I'm having trouble reproducing the problem locally.

@tvolk131 to help debug, we could add this change temporarily to your PR to get the logs from your CI runs
https://github.com/bradleystachurski/fedimint/commit/56f2abd1a4fa85e34079c9740045b61907ed2050

No guarantee it repros after that change though :/

elsirion (approved):

m1sterc001guy:
I believe it was electrum flakiness this time, retrying...

PR #5881, opened 2024-08-18
fix: omit ln fee for direct swaps

joschisan:

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -1434,20 +1434,29 @@ impl Gateway {
 
     /// Returns payment information that LNv2 clients can use to instruct this
     /// Gateway to pay an invoice or receive a payment.
-    pub async fn routing_info_v2(&self, federation_id: &FederationId) -> Option<RoutingInfo> {
-        Some(RoutingInfo {
-            public_key: self.public_key_v2(federation_id).await?,
-            send_fee_default: PaymentFee::SEND_FEE_LIMIT_DEFAULT,
-            // The base fee ensures that the gateway does not loose sats sending the payment due to
-            // fees paid on the transaction claiming the outgoing contract or subsequent
-            // transactions spending the newly issued ecash
-            send_fee_minimum: PaymentFee::SEND_FEE_MINIMUM,
-            // The base fee ensures that the gateway does not loose sats receiving the payment due
-            // to fees paid on the transaction funding the incoming contract
-            receive_fee: PaymentFee::RECEIVE_FEE_LIMIT_DEFAULT,
-            expiration_delta_default: 500,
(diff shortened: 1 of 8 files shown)


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -73,10 +73,13 @@ pub struct ReceiveOperationMeta {
     pub contract: IncomingContract,
 }
 
-/// Number of blocks until outgoing lightning contracts time out and user
+/// Number of blocks until outgoing lightning contracts times out and user
 /// client can refund it unilaterally
 pub const EXPIRATION_DELTA_LIMIT_DEFAULT: u64 = 500;
 
+/// A two hour buffer in case either the client or gateway go offline
+pub const CONTRACT_CONFIRMATION_BUFFER: u64 = 12;
```
elsirion, line 81: "The client" meaning the user? Is it a problem if they are offline? In LNv1 we explicitly support the user staying offline for extended periods of time while still being able to receive payments on previously generated addresses.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1434,20 +1434,29 @@ impl Gateway {
...
-            // The base fee ensures that the gateway does not loose sats receiving the payment due
-            // to fees paid on the transaction funding the incoming contract
-            receive_fee: PaymentFee::RECEIVE_FEE_LIMIT_DEFAULT,
-            expiration_delta_default: 500,
-            expiration_delta_minimum: EXPIRATION_DELTA_MINIMUM_V2,
-        })
+    pub async fn routing_info_v2(
+        &self,
+        federation_id: &FederationId,
+    ) -> anyhow::Result<Option<RoutingInfo>> {
```
elsirion, line 1440: Could this just be a `Result<RoutingInfo>`? Is the `None` return value ever being acted upon other than raising an error that the federation doesn't exist?

elsirion (approved):

gateway/ln-gateway/src/lib.rs
joschisan: No but that's the point - federation unsupported is a valid response not an error.

modules/fedimint-lnv2-client/src/lib.rs
joschisan: This buffer is only relevant for sending payments, to account for a possible delay between creating an outgoing contract and processing of the contract by the gateway. The expiration delta set by the client will still be accepted for a minimum of 2 hours after creation of the outgoing contract.

m1sterc001guy: Can we name this `CONTRACT_CONFIRMATION_BUFFER_HOURS`? IMO helps to have the unit denomination directly in the name

PR #5888, opened 2024-08-19
fix: bitcoin network naming mismatch

### LINKED ISSUE #5879: LDK node returns network as `main` instead of `bitcoin` on mainnet
This makes the `parsed_node_info` function fail. Testnet is likely also affected. I don't see the reason for this since LDK uses the same network enum as us internally afaik.

https://github.com/fedimint/fedimint/blob/1b3fca48499793196e322734bdc48c4e84a6f0b8/gateway/ln-gateway/src/lightning/mod.rs#L206-L221

tvolk131:
Fixes #5879

The [`Network`](https://docs.rs/bitcoin/0.30.2/bitcoin/network/constants/enum.Network.html) enum in the rust `bitcoin` crate is represented as the strings "bitcoin", "testnet", "regtest", and "signet" (see [here](https://docs.rs/bitcoin/0.30.2/src/bitcoin/network/constants.rs.html#569-578)). In `gateway_lnrpc.proto` we incorrectly documented `GetNodeInfoResponse.network` by saying that the acceptable values instead were "main", "test", "signet", and "regtest". Here we fix this documentation issue and change the LDK gateway implementation to stay true to that documentation change. No changes are needed for LND (we already convert "mainnet" to "bitcoin" [here](https://github.com/fedimint/fedimint/blob/df2073e53c496035fe37c999cb0821970f86dfcb/gateway/ln-gateway/src/lightning/lnd.rs#L671-L673)) and CLN (see the `network` field documentation [here](https://docs.corelightning.org/reference/lightning-getinfo)).

### DIFF
--- a/gateway/ln-gateway/src/lightning/ldk.rs
+++ b/gateway/ln-gateway/src/lightning/ldk.rs
@@ -224,14 +224,7 @@ impl ILnRpcClient for GatewayLdkClient {
             pub_key: self.node.node_id().serialize().to_vec(),
             // TODO: This is a placeholder. We need to get the actual alias from the LDK node.
             alias: format!("LDK Fedimint Gateway Node {}", self.node.node_id()),
-            network: match self.node.config().network {
-                Network::Bitcoin => "main",
-                Network::Testnet => "test",
-                Network::Signet => "signet",
-                Network::Regtest => "regtest",
-                _ => panic!("Unsupported network"),
-            }
-            .to_string(),
+            network: self.node.config().network.to_string(),
             block_height,
             synced_to_chain,
         })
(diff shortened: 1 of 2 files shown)


m1sterc001guy (dismissed):

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -225,8 +225,8 @@ impl ILnRpcClient for GatewayLdkClient {
             // TODO: This is a placeholder. We need to get the actual alias from the LDK node.
             alias: format!("LDK Fedimint Gateway Node {}", self.node.node_id()),
             network: match self.node.config().network {
-                Network::Bitcoin => "main",
-                Network::Testnet => "test",
+                Network::Bitcoin => "bitcoin",
```
elsirion, line 228: You could just use the `Display` impl, i.e. `.to_string()`

elsirion (approved):

PR #5889, opened 2024-08-20
chore: cln extension upgrade follow ups

m1sterc001guy:
Following up on some nit comments from https://github.com/fedimint/fedimint/pull/5871

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -116,6 +116,12 @@ export -f filter_count
 function nix_build_binary_for_version() {
   binary="$1"
   version="$2"
+
+  # nix flake did not build gateway-cln-extension prior to v0.4.0
+  if [[ ("$version" =~ ^v0\.[23]) && "$binary" == "gateway-cln-extension" ]] ; then
+    version="v0.4.0"
+  fi
+
   >&2 echo "Compiling ${binary} for version ${version} ..."
   echo "$(nix build 'github:fedimint/fedimint/'"$version"'#'"$binary" --no-link --print-out-paths)/bin/$binary"
 }
@@ -158,6 +164,7 @@ function use_gateway_binaries_for_version() {
   if [[ "$version" == "current" ]]; then
     unset FM_GATEWAYD_BASE_EXECUTABLE
     unset FM_GATEWAY_CLI_BASE_EXECUTABLE
+    unset FM_GATEWAY_CLN_EXTENSION_BASE_EXECUTABLE
   else
     var_name=$(nix_binary_version_var_name gatewayd "$version")
     FM_GATEWAYD_BASE_EXECUTABLE="${!var_name}"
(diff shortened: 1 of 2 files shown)


m1sterc001guy:
Upgrade tests pass: https://github.com/fedimint/fedimint/actions/runs/10476048832/job/29014106707?pr=5889

scripts/_common.sh
```diff
@@ -166,6 +167,17 @@ function use_gateway_binaries_for_version() {
     var_name=$(nix_binary_version_var_name gateway-cli "$version")
     FM_GATEWAY_CLI_BASE_EXECUTABLE="${!var_name}"
     export FM_GATEWAY_CLI_BASE_EXECUTABLE
+
+    # nix flake did not build gateway-cln-extension prior to v0.4.0
+    if [[ "$version" =~ ^v0\.[23] ]]; then
```
bradleystachurski, line 172: Hmm, I wouldn't expect this log in back-compat. Still investigating

```
Compiling gateway-cln-extension for version v0.2.1 ...
warning: ignoring untrusted flake configuration setting 'extra-substituters'.
Pass '--accept-flake-config' to trust it
warning: ignoring untrusted flake configuration setting 'extra-trusted-public-keys'.
Pass '--accept-flake-config' to trust it
error: flake 'github:fedimint/fedimint/v0.2.1' does not provide attribute 'packages.x86_64-linux.gateway-cln-extension', 'legacyPackages.x86_64-linux.gateway-cln-extension' or 'gateway-cln-extension'
```

https://github.com/fedimint/fedimint/actions/runs/10476502738/job/29015521632?pr=5889#step:6:1348

m1sterc001guy: I don't think we had a flake output for `gateway-cln-extension` in v0.2.1

m1sterc001guy: v0.4.0 is backwards compatible with < v0.4.0, which is why I set it like this.

bradleystachurski: Ahh, it's because we precompile without this check in test-ci-all

https://github.com/fedimint/fedimint/blob/f125d96061c8da593f5ec4362ebfef1c14e7bded/scripts/tests/test-ci-all.sh#L248-L256

It's a bit messy to add this specific conditional logic there. Perhaps we can update `nix_build_binary_for_version` to handle the version check logic?

bradleystachurski: A bit hacky, but something like

```bash
function nix_build_binary_for_version() {
  binary="$1"
  version="$2"

  # nix flake did not build gateway-cln-extension prior to v0.4.0
  if [[ ("$version" =~ ^v0\.[23]) && "$binary" == "gateway-cln-extension" ]] ; then
    version="v0.4.0"
  fi

  >&2 echo "Compiling ${binary} for version ${version} ..."
  echo "$(nix build 'github:fedimint/fedimint/'"$version"'#'"$binary" --no-link --print-out-paths)/bin/$binary"
}
export -f nix_build_binary_for_version
```

then we can remove the condition logic in `use_gateway_binaries_for_version`

m1sterc001guy: Done! Agree that is cleaner.

bradleystachurski (approved):

PR #5891, opened 2024-08-20
chore: remove redundant arguments

### LINKED ISSUE #5843: Remove superfluous `module_instance_id` argument
> Unrelated to the PR since `process_input` does the same: why do we have `module_instance_id` as an argument? `DynInput` already contains it:
> 
> ```rust
> #[doc = " An owned, immutable input to a [`Transaction`](fedimint_core::transaction::Transaction)"]
> pub struct DynInput {
>     module_instance_id: ::fedimint_core::core::ModuleInstanceId,
>     inner: Box<dyn IInput + 'static + Send + Sync>,
> }
> ```
> 
> I guess it's a sanity-check and supposed to be set to the *module's* module instance id?  In that case we should probably just use the same macro for `DynServerModule` as for `DynInput`, that includes the instance id in the dyn struct.

_Originally posted by @elsirion in https://github.com/fedimint/fedimint/pull/5836#discussion_r1714954064_

elsirion:
Fixes #5843

### DIFF
--- a/fedimint-core/src/core/server.rs
+++ b/fedimint-core/src/core/server.rs
@@ -55,11 +55,7 @@ pub trait IServerModule: Debug {
     // Use this function to parallelise stateless cryptographic verification of
     // inputs across a transaction. All inputs of a transaction are verified
     // before any input is processed.
-    fn verify_input(
-        &self,
-        input: &DynInput,
-        module_instance_id: ModuleInstanceId,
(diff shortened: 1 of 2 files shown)


tvolk131 (approved):
I'm not super familiar with this part of the codebase, but this seems like a fairly simple change. LGTM!

joschisan (approved):

PR #5899, opened 2024-08-21
chore: use a hardcoded regtest genesis block hash

dpc:
Instead of assuming that any unknown gensis block hash is a regtest network, use the fact the regtest genesis block is hardcoded.

This should protect us from any bugs that lead to invalid block hashes, like in #5813.


m1sterc001guy (approved):

bradleystachurski (approved):
Verified this catches the regression if it includes the fix in https://github.com/fedimint/fedimint/pull/5902

PR #5900, opened 2024-08-21
demo: verify regression is caught using hardcoded regtest genesis hash

bradleystachurski:
This passed locally, running in CI to make sure it's not my env. I would have expected this to fail after reverting the fix.


bradleystachurski:
Search the logs for `genesis_hash` and `encoded_as_hex` to see the failures

bradleystachurski:
Demonstrated what we needed, closing

PR #5901, opened 2024-08-21
chore: publish to falkehub

elsirion:


.github/workflows/flakebox-flakehub-publish.yml
```diff
@@ -0,0 +1,35 @@
+# THIS FILE IS AUTOGENERATED FROM FLAKEBOX CONFIGURATION
```
dpc, line 1: I see you copied it verbatim from some other repo. I guess that works. :D

dpc (approved):

.github/workflows/flakebox-flakehub-publish.yml
elsirion: Yep, idk how to patch in new features after the fact with flakebox otherwise. I even made it read-only (if that worked) :laughing:

PR #5902, opened 2024-08-21
fix(tests): use esplora and electrs backends

bradleystachurski:
Noticed while testing https://github.com/fedimint/fedimint/pull/5900

Introduces new env vars to override only the wallet server's Bitcoin RPC for esplora and electrs backend tests. We can't use `FM_FORCE_*` env vars to override because that will also override the wallet client, which fails since electrum can't support `get_txout_proof`.

### DIFF
--- a/scripts/tests/backend-test.sh
+++ b/scripts/tests/backend-test.sh
@@ -82,8 +82,8 @@ function run_tests() {
   fi
 
   # Switch to electrum and run wallet tests
-  export FM_BITCOIN_RPC_KIND="electrum"
-  export FM_BITCOIN_RPC_URL="tcp://127.0.0.1:$FM_PORT_ELECTRS"
+  export FM_TEST_BACKEND_BITCOIN_RPC_KIND="electrum"
+  export FM_TEST_BACKEND_BITCOIN_RPC_URL="tcp://127.0.0.1:$FM_PORT_ELECTRS"
 
   if [ -z "${FM_TEST_ONLY:-}" ] || [ "${FM_TEST_ONLY:-}" = "electrs" ]; then
     >&2 echo "### Testing against electrs"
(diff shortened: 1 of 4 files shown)


bradleystachurski:
Needs debugging to get tests to pass using esplora and electrs backends. I'll move out of draft once it's ready.

scripts/tests/backend-test.sh
```diff
@@ -95,14 +95,14 @@ function run_tests() {
...
-  export FM_BITCOIN_RPC_URL="http://127.0.0.1:$FM_PORT_ESPLORA"
+  export FM_TEST_BACKEND_BITCOIN_RPC_KIND="esplora"
+  export FM_TEST_BACKEND_BITCOIN_RPC_URL="http://127.0.0.1:$FM_PORT_ESPLORA"
 
   if [ -z "${FM_TEST_ONLY:-}" ] || [ "${FM_TEST_ONLY:-}" = "esplora" ]; then
     >&2 echo "### Testing against esplora"
     cargo nextest run --locked --workspace --all-targets \
       ${CARGO_PROFILE:+--cargo-profile ${CARGO_PROFILE}} ${CARGO_PROFILE:+--profile ${CARGO_PROFILE}} \
-      ${TEST_ARGS_THREADED} \
+      ${TEST_ARGS_SERIALIZED} \
```
bradleystachurski, line 105: Intentional. I'm not sure why, but some of the esplora tests can take a long time and since we lock bitcoind we can exceed the 120 second timeout if running all tests in parallel.

elsirion (approved):
Can we assert in a test that the correct blockchain API is used to avoid regressions? Do you think that would be useful?

dpc:
> Do you think that would be useful?

Seems veery unlikely to accidentally be using different one. All of signet, testnet and mainnet would not work in tests to begin with.

bradleystachurski:
> Seems veery unlikely to accidentally be using different one. All of signet, testnet and mainnet would not work in tests to begin with.

It might be useful since we've been accidentally using the wrong one for esplora and electrum for a long time 😃 

The changes in this PR were necessary since the esplora and electrum backend tests were still using bitcoind as an rpc and not testing the respective backends.

I'll put up a followup PR soon.

PR #5903, opened 2024-08-22
Don't merge: trying to simplify #5697

elsirion:

### DIFF
--- a/fedimint-client/src/meta.rs
+++ b/fedimint-client/src/meta.rs
@@ -227,6 +227,7 @@ impl<S: MetaSource + ?Sized> MetaService<S> {
 #[non_exhaustive]
 pub struct LegacyMetaSource {
     reqwest: reqwest::Client,
+    extra_override_keys: Vec<String>,
 }
 
 #[apply(async_trait_maybe_send!)]
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -394,14 +394,27 @@ impl ClientModule for WalletClientModule {
         // fetch consensus height first
         let session_count = self.client_ctx.global_api().session_count().await?;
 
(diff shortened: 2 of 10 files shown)


fedimint-client/src/meta.rs
```diff
@@ -227,6 +227,7 @@ impl<S: MetaSource + ?Sized> MetaService<S> {
 #[non_exhaustive]
 pub struct LegacyMetaSource {
     reqwest: reqwest::Client,
+    extra_override_keys: Vec<String>,
```
dpc, line 230: Unrelated, slipped in, I assume? @elsirion

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -1161,7 +1161,7 @@ mod tests {
             }
         );
         assert_eq!(
-            recover_scan_idxes_for_activity(TweakIdx(10), &BTreeSet::new(), |cur_idx| async move {
+            recover_scan_idxes_for_activity(TweakIdx(9), &BTreeSet::new(), |cur_idx| async move {
```
dpc, line 1164: The fact that the argument passed to this function, which acts somewhat like a start of a range is not inclusive is very weird, IMO. :D

fedimint-client/src/meta.rs
elsirion: Oh yeah :see_no_evil:

modules/fedimint-wallet-client/src/lib.rs
elsirion: Isn't it inclusive? Maybe I missed something, I just mechanically moved the `.prev()` call out of the `recover_scan_idxes_for_activity` fn into the callers.

dpc:
Accomodated these changes in the original PR.

PR #5904, opened 2024-08-22
fix: make spend ecash authenticated

m1sterc001guy:
Needs to be authenticated so that anyone can't just steal from the gateway


tvolk131 (approved):
Whoops, that could have ended badly 😆 good catch

PR #5905, opened 2024-08-22
Surface Bitcoin RPC used by fedimintd

bradleystachurski:
Followup addressing:

> Can we assert in a test that the correct blockchain API is used to avoid regressions? Do you think that would be useful?

https://github.com/fedimint/fedimint/pull/5902#pullrequestreview-2254638875

---

Adds two endpoints to wallet server:
- authed endpoint that returns bitcoin rpc kind and url
  - auth required since bitcoin url should not be leaked to public
  - useful for surfacing to guardians in UI
  - url is stripped of `user:pass` prior to sending over the wire
- no-auth endpoint that returns the bitcoin kind for a peer
  - useful for surfacing to clients (e.g. `fedimint-observer`)

```bash
bash-5.2$ fedimint-cli module wallet --help
Usage: wallet <COMMAND>

Commands:
  await-deposit           Await a deposit on a given deposit address
  get-bitcoin-rpc-kind    Returns the Bitcoin RPC kind
  get-bitcoin-rpc-config  Returns the Bitcoin RPC kind and URL, if authenticated
  help                    Print this message or the help of the given subcommand(s)

bash-5.2$ fedimint-cli module wallet get-bitcoin-rpc-kind 0
"bitcoind"

bash-5.2$ FM_OUR_ID=0 FM_PASSWORD=pass fedimint-cli module wallet get-bitcoin-rpc-config
{
  "kind": "bitcoind",
  "url": "http://127.0.0.1:11368/"
}
```

### DIFF
--- a/modules/fedimint-wallet-tests/tests/tests.rs
+++ b/modules/fedimint-wallet-tests/tests/tests.rs
@@ -1151,3 +1151,32 @@ mod fedimint_migration_tests {
         .await
     }
 }
+
+// Verify the correct Bitcoin RPC is used
+
+#[cfg(feature = "bitcoind")]
+#[test]
+fn verify_bitcoind_backend() {
+    let fixtures = fixtures();
+    let dyn_bitcoin_rpc = fixtures.dyn_bitcoin_rpc();
+    let bitcoin_rpc_kind = dyn_bitcoin_rpc.get_bitcoin_rpc_config().kind;
+    assert_eq!(bitcoin_rpc_kind, "bitcoind")
+}
+
+#[cfg(feature = "esplora")]
+#[test]
+fn verify_esplora_backend() {
+    let fixtures = fixtures();
+    let dyn_bitcoin_rpc = fixtures.dyn_bitcoin_rpc();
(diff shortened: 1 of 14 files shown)


modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -1053,3 +1053,32 @@ mod fedimint_migration_tests {
         .await
     }
 }
+
+// Verify the correct Bitcoin RPC is used
+
+#[cfg(feature = "bitcoind")]
```
bradleystachurski, line 1059: Another approach is using `#[ignore]`, but that makes the calls to `cargo nextest` messier since it requires multiple flags and filters.

dpc (dismissed):

dpc:
@bradleystachurski 

I thought that maybe we'll need `fedimint-cli dev get-bitcoin-backend` kind of thing.

elsirion:
I think @Kodylow could use the local config, especially the bitcoin backend in use, exposed as an admin API route anyway, maybe that way we could use an admin CLI tool like mentioned by @dpc to fetch and check it?

bradleystachurski:
Moving to draft until CI passes

dpc (approved):

Issue #5203, opened 2024-05-03
Remove usage of `assume_checked()`


tvolk131:
Follow-up to #2356

Bitcoin v30 introduced the concept of checked and unchecked addresses. While migrating over from v0.29 to v0.30, I called `assume_checked()` wherever necessary to make the compiler happy. Now that the migration is complete, we should remove these calls and instead use [`require_network()`](https://docs.rs/bitcoin/0.30.2/bitcoin/address/struct.Address.html#method.require_network).

Edit: See [this comment](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/CHANGELOG.md#changes-to-and-involving-the-network-type) in the changelog for `bitcoin` v0.32.0 which indicates that the work here might not be that important.

mubarak23:
@tvolk131 can you add more context to this, i want to work on it,

kindly assign

tvolk131:
> can you add more context to this

Sure! We currently rely on the rust [`bitcoin` crate v0.30.2](https://crates.io/crates/bitcoin/0.30.2). In this version and all newer versions (for now), the [`Address`](https://docs.rs/bitcoin/0.30.2/bitcoin/address/struct.Address.html) type has a [`NetworkValidation`](https://docs.rs/bitcoin/0.30.2/bitcoin/address/trait.NetworkValidation.html) generic parameter attached. The purpose of this is to ensure that bitcoin addresses from the wrong network are used (e.g. we don't want to accidentally send mainnet funds to a testnet address). Whenever addresses are deserialized (received over the wire, loaded from disk, etc...) they are unchecked. They can become checked either by calling [`assume_checked()`](https://docs.rs/bitcoin/0.30.2/bitcoin/address/struct.Address.html#method.assume_checked) or [`require_network()`](https://docs.rs/bitcoin/0.30.2/bitcoin/address/struct.Address.html#method.require_network). Currently we're calling `assume_checked()` everywhere since it's easier. But it's unsafe since we're not actually verifying that the address belongs to the correct network. We should instead call `require_network()` to assert that all bitcoin addresses used belong to the same network that the federation is running on.

Let me know if there's any else I can help you with!

mubarak23:
awesome, where is the require_network() function coming from ?

tvolk131:
I linked to it in my earlier comment

mubarak23:
i completed the work, everytime i try to add a commit message i get the following error 

`
[...]

PR #5907, opened 2024-08-22
feat: gateway can withdraw its on-chain funds

tvolk131:


gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -161,6 +163,13 @@ pub trait ILnRpcClient: Debug + Send + Sync {
     /// wallet.
     async fn get_funding_address(&self) -> Result<GetFundingAddressResponse, LightningRpcError>;
 
+    async fn sweep_funds(
```
m1sterc001guy, line 166: It is kind of confusing to have a `sweep_funds` function that takes an amount. Why not just `withdraw`?

tvolk131: Renamed everything to withdraw!

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -414,6 +414,29 @@ impl ILnRpcClient for GatewayLdkClient {
             })
     }
 
+    async fn withdraw_onchain(
+        &self,
+        address: Address,
+        amount: BitcoinAmountOrAll,
+        _fee_rate_sats_per_vbyte: u64, // TODO: Respect this fee rate.
```
tvolk131, line 421: Note for reviewers: `ldk-node` v0.3 doesn't support setting a fee-rate, but support for this is planned for v0.4 and is tracked here: https://github.com/lightningdevkit/ldk-node/issues/176

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -959,6 +960,50 @@ impl GatewayLightning for ClnRpcService {
...
+        let txid = self
+            .rpc_client()
+            .await
+            .map_err(|e| Status::internal(e.to_string()))?
+            .call(cln_rpc::Request::Withdraw(
+                model::requests::WithdrawRequest {
+                    feerate: Some(cln_rpc::primitives::Feerate::PerKw(
+                        // 1 vbyte = 4 weight units, so 250 vbytes = 1,000 weight units.
+                        request_inner.fee_rate_sats_per_vbyte as u32 * 250,
+                    )),
```
elsirion, line 978: Do we have any utility conversion fn that we could use here?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -959,6 +960,50 @@ impl GatewayLightning for ClnRpcService {
...
+                    minconf: Some(0),
+                    utxos: None,
+                    destination: request_inner.address,
+                    satoshi: if request_inner.amount_sats == 0 {
+                        cln_rpc::primitives::AmountOrAll::All
+                    } else {
```
elsirion, line 984: Would be nice not to use 0 as a magic value

gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -279,6 +281,23 @@ message GetLnOnchainAddressResponse {
   string address = 1;
 }
 
+message WithdrawOnchainRequest {
+  // The address to withdraw funds to.
+  string address = 1;
+
+  // The amount to withdraw in sats.
+  // If 0/empty, the entire balance will be swept.
+  uint64 amount_sats = 2;
```
elsirion, line 290: Should be optional if possible.

gateway/cli/src/lightning_commands.rs
```diff
@@ -69,6 +71,25 @@ pub enum LightningCommands {
...
+    /// Withdraw funds from the lightning node's on-chain wallet to a specified
+    /// address.
+    WithdrawOnchain {
+        /// The address to withdraw funds to.
+        #[clap(long)]
+        address: bitcoin::Address<NetworkUnchecked>,
+
+        /// The amount to withdraw, in satoshis.
+        #[clap(long)]
+        amount_sats: Option<u64>,
```
elsirion, line 83: Could we use an amount type here? Maybe even `AmountOrAll` so users can specify `all`

tvolk131: Done.

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: Done.

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: Done.

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: I can't find one. Want me to move this logic into `fedimint-core` somewhere?

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -959,6 +960,50 @@ impl GatewayLightning for ClnRpcService {
...
+            .rpc_client()
+            .await
+            .map_err(|e| Status::internal(e.to_string()))?
+            .call(cln_rpc::Request::Withdraw(
+                model::requests::WithdrawRequest {
+                    feerate: Some(cln_rpc::primitives::Feerate::PerKw(
+                        // 1 vbyte = 4 weight units, so 250 vbytes = 1,000 weight units.
+                        request_inner.fee_rate_sats_per_vbyte as u32 * 250,
+                    )),
+                    minconf: Some(0),
```
m1sterc001guy, line 979: Given that you're setting the feerate above, should this be `None`?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1313,6 +1313,27 @@ impl Gateway {
...
+            address,
+            amount: amount_sats,
+            fee_rate_sats_per_vbyte,
+        }: WithdrawOnchainPayload,
+    ) -> Result<Txid> {
+        let context = self.get_lightning_context().await?;
+        let response = context
+            .lnrpc
+            .withdraw_onchain(
+                address.assume_checked(),
```
m1sterc001guy, line 1329: Can we check the address here instead of using `assume_checked`? Not sure how cumbersome that will be

gateway/ln-gateway/src/lib.rs
```diff
@@ -1313,6 +1313,27 @@ impl Gateway {
...
+                amount_sats,
+                fee_rate_sats_per_vbyte,
+            )
+            .await?;
+        Ok(Txid::from_str(&response.txid).unwrap())
```
m1sterc001guy, line 1334: Let's use `expect`

gateway/ln-gateway/src/lightning/ldk.rs
m1sterc001guy: Maybe link the issue directly in the code?

gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -279,6 +281,23 @@ message GetLnOnchainAddressResponse {
   string address = 1;
 }
 
+message WithdrawOnchainRequest {
+  // The address to withdraw funds to.
+  string address = 1;
+
+  // The amount to withdraw in sats.
+  // If 0/empty, the entire balance will be swept.
+  optional uint64 amount_sats = 2;
```
m1sterc001guy, line 290: Nit: you can use protobuf's `oneof` which allows you to specify an enum. Might make a little more sense because you wouldn't need to use a magic value here. In the generated rust code, it still uses a `Option` though

gateway/ln-gateway/src/lightning/ldk.rs
m1sterc001guy: What feerate will it use instead? Something from esplora?

gateway/ln-gateway/src/bin/cln_extension.rs
tvolk131: This doesn't relate to the feerate. This determines UTXO selection for the sweep. I set this to 0 so that users can sweep funds of channels they just closed.

gateway/ln-gateway/src/lib.rs
tvolk131: Mind if I get around to this later? It does appear to be cumbersome. This work is being tracked through #5203 which I'm working on in the background.

gateway/ln-gateway/src/lib.rs
tvolk131: Decided to return an actually `Err` instead

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: Are you thinking of something like this?

```
message WithdrawOnchainRequest {
  // The address to withdraw funds to.
  string address = 1;

  // The amount to withdraw. Either `amount_sats` or `all` must be set.
  oneof amount {
    // The amount to withdraw in sats.
    uint64 amount_sats = 2;

    // Withdraw all funds.
    bool all = 3;
  }

  // The fee rate to use for the withdrawal transaction.
  uint64 fee_rate_sats_per_vbyte = 4;
}
```

This produces the following type:

```
    pub enum Amount {
        /// The amount to withdraw in sats.
        #[prost(uint64, tag = "2")]
        AmountSats(u64),
        /// Withdraw all funds.
        #[prost(bool, tag = "3")]
        All(bool),
    }
```

I don't like this as much, since the rust type changes from an `Option<u64>` to an `Option<Amount>`. The latter type has two invalid states (`None` and `Some(Amount::All(false))`) whereas the former has none. I lean towards keeping it as an `optional uint64`. WDYT?

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Sure

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Looks like `unwrap` is still used though?

tvolk131: Sorry, just pushed

gateway/ln-gateway/proto/gateway_lnrpc.proto
m1sterc001guy: That is fine. I don't really like magic values but I agree `oneof` is a bit cumbersome. Another reason to get rid of grpc and switch to axum.

gateway/ln-gateway/src/bin/cln_extension.rs
m1sterc001guy: Ah ok I misread the docs. Yeah 0 is the correct value here IMO

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Looks like it uses esplora to attempt a 6-block conf: https://github.com/lightningdevkit/ldk-node/blob/7eca7fbc66c7430ac488ca80b69db2e1dea40f34/src/fee_estimator.rs#L103-L126

And falls back to 20 sat/vbyte: https://github.com/lightningdevkit/ldk-node/blob/7eca7fbc66c7430ac488ca80b69db2e1dea40f34/src/fee_estimator.rs#L164

m1sterc001guy:
Looking good. Can we add a test before merging?

elsirion (dismissed):
Looks good to me, but requires rebase and ideally a second ACK by  @m1sterc001guy

tvolk131:
> Looks good to me, but requires rebase and ideally a second ACK by @m1sterc001guy

Sorry I still gotta add a test here before submitting. Got distracted by speeding up mprocs startup time. Thanks for the extra set of eyes on this!

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -1124,6 +1126,48 @@ impl ILnRpcClient for GatewayLndClient {
...
+        &self,
+        address: Address,
+        amount: BitcoinAmountOrAll,
+        fee_rate_sats_per_vbyte: u64,
+    ) -> Result<WithdrawOnchainResponse, LightningRpcError> {
+        #[allow(deprecated)]
+        let request = match amount {
+            BitcoinAmountOrAll::All => SendCoinsRequest {
+                addr: address.to_string(),
+                amount: 0,
```
Kodylow, line 1139: nit: can we do ::default here or add a note? this is confusing

Kodylow (dismissed):
needs rebase

Kodylow (approved):

m1sterc001guy:
@tvolk131 can you link the PR where you will be adding the test?

tvolk131:
> @tvolk131 can you link the PR where you will be adding the test?

We're adding tests for this behavior separately in #6086

PR #5910, opened 2024-08-23
chore: further small wallet recovery improvements

dpc:

### DIFF
--- a/modules/fedimint-wallet-client/src/backup.rs
+++ b/modules/fedimint-wallet-client/src/backup.rs
@@ -418,7 +418,13 @@ impl RecoveryFromHistory for WalletRecovery {
 
 /// We will check this many addresses after last actually used
 /// one before we give up
-pub(crate) const RECOVER_MAX_GAP: u64 = 10;
+pub(crate) const ONCHAIN_RECOVER_MAX_GAP: u64 = 10;
+
+/// When scanning the history of the Federation, there's no need to be
+/// so cautious about the privacy (as it's perfectly private), so might
+/// as well increase the gap limit.
+pub(crate) const FEDERATION_RECOVER_MAX_GAP: u64 = 50;
+
 /// New client will start deriving new addresses from last used one
 /// plus that many indexes. This should be less than
 /// `MAX_GAP`, but more than 0: We want to make sure we detect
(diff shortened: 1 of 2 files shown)


modules/fedimint-wallet-client/src/backup.rs
```diff
@@ -417,7 +418,13 @@ impl RecoveryFromHistory for WalletRecovery {
...
 /// We will check this many addresses after last actually used
 /// one before we give up
-pub(crate) const RECOVER_MAX_GAP: u64 = 10;
+pub(crate) const ONCHAIN_RECOVER_MAX_GAP: u64 = 10;
+
+/// When scanning the history of the Federation, there's no need to be
+/// so cautious about the privacy (as it's perfectly private), so might
+/// as well increase the gap limit.
+pub(crate) const FEDERATION_RECOVER_MAX_GAP: u64 = 50;
+
```
elsirion, line 427: It just occurred to me that as long as we skip, i.e. never give to user, addresses looked up on-chain during recovery we can also choose a rather high gap limit there. 

* If we find something: great! The user would have lost funds otherwise.
* If we don't: we correlated some pseudo-random addresses that will never be used again

We already skip `RECOVER_NUM_IDX_ADD_TO_LAST_USED`, but we should actually use the gap limit there imo.

modules/fedimint-wallet-client/src/backup.rs
```diff
@@ -466,13 +473,20 @@ where
...
             break;
         }
 
-        if !check_addr_history(cur_tweak_idx).await?.is_empty() {
+        let history = retry(
+            "Check address history",
+            backoff_util::background_backoff(),
+            || async { check_addr_history(cur_tweak_idx).await },
+        )
+        .await?;
```
elsirion, line 487: :+1:

modules/fedimint-wallet-client/src/backup.rs
dpc: @elsirion 
The problem with skipping addresses that were in the "gap" is that if user generates more addresses and then needs to recover again, these new deposits might not be detected anymore.

Because of that `RECOVER_NUM_IDX_ADD_TO_LAST_USED` is smaller than `ONCHAIN_RECOVER_MAX_GAP`. So `RECOVER_NUM_IDX_ADD_TO_LAST_USED` will be skipped, but it's less than the gap.

elsirion: Good point, the backup may be lost, which would make that a problem.

elsirion (approved):

PR #5915, opened 2024-08-24
Restore operation log entries for deposits

elsirion:
The functionality to list and observe deposits was accidentally removed in #5473.

It's a bit tricky because technically a single generated address can now cause multiple operations. My plan is to create a first operation on address generation, this will correspond to the first deposit seen on-chain. 

## Possible extension
Any successive deposit could result in another operation log entry. Since this will be hard to do 100% consistently (we don't model the "transaction seen but unconfirmed" state anymore) and the previous behavior was to ignore this anyway I'm not including this here.

### DIFF
--- a/modules/fedimint-wallet-client/src/lib.rs
+++ b/modules/fedimint-wallet-client/src/lib.rs
@@ -53,7 +53,8 @@ use fedimint_core::module::{
     ApiVersion, CommonModuleInit, ModuleCommon, ModuleInit, MultiApiVersion,
 };
 use fedimint_core::task::{MaybeSend, MaybeSync, TaskGroup};
-use fedimint_core::util::backoff_util;
+use fedimint_core::util::backoff_util::background_backoff;
+use fedimint_core::util::{backoff_util, retry};
 use fedimint_core::{
     apply, async_trait_maybe_send, push_db_pair_items, runtime, Amount, OutPoint, TransactionId,
 };
@@ -90,11 +91,27 @@ pub struct BitcoinTransactionData {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
-pub enum DepositState {
+pub enum DepositStateV1 {
     WaitingForTransaction,
     WaitingForConfirmation(BitcoinTransactionData),
     Confirmed(BitcoinTransactionData),
     Claimed(BitcoinTransactionData),
+    Failed(String),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub enum DepositStateV2 {
+    WaitingForTransaction,
+    WaitingForConfirmation {
+        #[serde(with = "bitcoin::amount::serde::as_sat")]
+        btc_deposited: bitcoin::Amount,
+        btc_out_point: bitcoin::OutPoint,
+    },
+    Claimed {
+        #[serde(with = "bitcoin::amount::serde::as_sat")]
+        btc_deposited: bitcoin::Amount,
+        btc_out_point: bitcoin::OutPoint,
+    },
     Failed(String),
 }
 
@@ -257,6 +274,10 @@ pub struct WalletOperationMeta {
 pub enum WalletOperationMetaVariant {
     Deposit {
         address: bitcoin::Address<NetworkUnchecked>,
+        /// Added in 0.4.2, can be `None` for old deposits or `Some` for ones
+        /// using the pegin monitor. The value is the child index of the key
+        /// used to generate the address, so we can re-generate the secret key
+        /// from our root secret.
         #[serde(default)]
         tweak_idx: Option<TweakIdx>,
         #[serde(default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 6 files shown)


modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -630,6 +669,85 @@ impl WalletClientModule {
         Ok((operation_id, address, tweak_idx))
     }
 
+    /// Returns a stream of updates about an ongoing deposit operation created
+    /// with [`WalletClientModule::allocate_deposit_address_expert_only`].
+    /// Returns an error for old deposit operations created prior to the 0.4
+    /// release and not driven to completion yet. This should be rare enough
+    /// that an indeterminate state is ok here.
```
elsirion, line 676: @maan2003 important to note for integrators

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -92,9 +93,21 @@ pub struct BitcoinTransactionData {
...
+    Confirmed {
+        // Caution: previously this variant contained `BitcoinTransactionData`, it should not have
+        // been persisted by the fedimint client though
+    },
+    Claimed {
+        // Caution: previously this variant contained `BitcoinTransactionData`, which may still be
+        // present in client DBs. If you want to re-use the btc_transaction and out_idx key make
+        // sure to remove old data cached in the operation log first.
+        btc_out_point: bitcoin::OutPoint,
+    },
```
elsirion, line 110: Add final amount.
Also for old tx.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -92,9 +93,21 @@ pub struct BitcoinTransactionData {
...
 pub enum DepositState {
     WaitingForTransaction,
-    WaitingForConfirmation(BitcoinTransactionData),
-    Confirmed(BitcoinTransactionData),
-    Claimed(BitcoinTransactionData),
+    WaitingForConfirmation {
+        // Caution: previously this variant contained `BitcoinTransactionData`, it should not have
+        // been persisted by the fedimint client though
+        btc_out_point: bitcoin::OutPoint,
+    },
```
elsirion, line 100: Maybe add confirmations?

elsirion: Out of scope for this PR imo

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -630,6 +673,113 @@ impl WalletClientModule {
...
+        let operation_meta = operation.meta::<WalletOperationMeta>();
+
+        let WalletOperationMetaVariant::Deposit {
+            address, tweak_idx, ..
+        } = operation_meta.variant
+        else {
+            bail!("Operation is not a deposit operation");
+        };
+
+        let Some(tweak_idx) = tweak_idx else {
```
elsirion, line 704: Document tweak idx being related to new oeprations

maan2003 (approved):
me and @elsirion did two review passes synchronously

fedimint-backports:

```bash
git worktree add -d .worktree/backport-5915-to-releases/v0.4 origin/releases/v0.4
cd .worktree/backport-5915-to-releases/v0.4
git switch --create backport-5915-to-releases/v0.4
git cherry-pick -x ceeec00b4c5a821791c39508ce31832261544ebe 136cb8b7665d43d3b686280cbca8c5f3e101afd1
```

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -90,14 +91,30 @@ pub struct BitcoinTransactionData {
...
+pub enum DepositStateV2 {
+    WaitingForTransaction,
+    WaitingForConfirmation {
+        #[serde(with = "bitcoin::amount::serde::as_sat")]
+        btc_deposited: bitcoin::Amount,
+        btc_out_point: bitcoin::OutPoint,
+    },
+    Claimed {
+        #[serde(with = "bitcoin::amount::serde::as_sat")]
+        btc_deposited: bitcoin::Amount,
```
shaurya947, line 112: @maan2003 and I were looking at this and I learned that this amount does not take into account the fee that we deducted by the federation.. maybe this is OK for now since the fee is a fixed amount.. but probably not the best pattern. I imagine an integrator would want to know the amount of BTC that was actually "claimed" to be used as their e-cash.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -630,6 +677,114 @@ impl WalletClientModule {
...
+                    btc_deposited,
+                    btc_out_point
+                };
+
+                stream_cient_ctx.module_db().wait_key_exists(&ClaimedPegInKey {
+                    peg_in_index: tweak_idx,
+                    btc_out_point,
+                }).await;
+
+                yield DepositStateV2::Claimed {
```
shaurya947, line 780: Wondering if we're missing the "await_primary_module_output" (or equivalent) here to actually indicate in the update stream that the e-cash has been issued and added to the balance..? Similar to https://github.com/fedimint/fedimint/blob/5e4247572386518809f5edb86397451a523edf3e/modules/fedimint-ln-client/src/lib.rs#L1624C43-L1624C71

modules/fedimint-wallet-client/src/lib.rs
elsirion: Yeah, unfortunately it's very hard to accurately show the amount received once e-cash fees are taken into account. So rather than showing a mostly correct value I opted for just supplying the raw deposit value.

This PR is in no way ideal, it's a stop-gap solution till we probably rebuild the pegin monitor to allow for better observeability.

modules/fedimint-wallet-client/src/lib.rs
elsirion: Yeah, missed that, but should be easy to fix.

PR #5921, opened 2024-08-26
chore: remove secp256k1-zkp in gateway

m1sterc001guy:
It is not necessary.


elsirion (approved):
I don't think it matters on which of both we depend (something in the workspace will always pull in zkp), but we certainly shouldn't depend on both.

dpc (approved):

PR #5923, opened 2024-08-26
chore: use typed errors in LNv2

m1sterc001guy:
Small ergonomic improvement, gives us some nice spans in the logs, can re-use `select_client` and get stack traces in the logs.

Also standardizes the responses from the gateway to prevent privacy leaks.


gateway/ln-gateway/src/lib.rs
```diff
@@ -1677,7 +1677,8 @@ impl Gateway {
     pub async fn routing_info_v2(
         &self,
         federation_id: &FederationId,
-    ) -> anyhow::Result<Option<RoutingInfo>> {
+    ) -> Result<Option<RoutingInfo>> {
+        info!("JUMOELL received request for routing info");
```
elsirion, line 1681: Do you want to keep your name in here?

elsirion (approved):
One nit

tvolk131 (approved):
Nice cleanups here! My only thought is that the PR name (and matching commit name) only reflect a subset of what's here. Could be worth renaming.

joschisan (changes requested):
Do not merge it like this. The gateways endpoints for the client return a string error since the user client will neither branch on those nor display the specifics to the user. The implication for the user is always to try a different gateway. Furthermore, the current GatewayError enum contains many variants that have nothing to do with a users ln payment.

The errors are only for debugging and need to maintain backwards compatibility here with the clients so we should use a string here.

m1sterc001guy:
From our call this morning. @joschisan add anything I missed.

1. Removal of wrapped `Result` is a good improvement, but the current state of error handling is still a bit confusing. After this PR we will:
 - Separate `GatewayError` into 3 error types: LNv1, LNv2, and GatewayAdmin error types. 
 - Each error type will log in `into_response` when returning the `Response` back to the client. 
 - Privacy sensitive errors like LNv1 and LNv2 will map error enums to error codes and messages like we do today. 
 - `GatewayAdmin` errors will return the full error back to the client for debugging purposes
 - Cleanup and consolidate error types.

joschisan (dismissed):

PR #5931, opened 2024-08-27
chore(ci): parameterize which upgrade tests to run

bradleystachurski:
Upgrade tests run all of the test kinds (`fedimint-cli`, `fedimintd`, and `gateway`) using the same upgrade paths. Now that `fedimintd` introduced coordinated upgrades, it's useful to trigger different upgrade paths for different test kinds. E.g. we know that `fedimintd` will fail upgrading from 0.2.1 -> 0.4.0, however it would be useful to test that upgrade path for `fedimint-cli` and `gateway`.

The changes to the GH workflow allows specifying the test kinds when manually triggering a run. If none are provided, all test kinds are run.

### DIFF
--- a/.github/workflows/upgrade-tests.yml
+++ b/.github/workflows/upgrade-tests.yml
@@ -18,6 +18,10 @@ on:
         description: 'Versions to test (if empty, uses same as daily run)'
         required: false
         default: ''
+      test_kinds:
+        description: 'Comma-separated tests to run. Options: fedimintd,fedimint-cli,gateway'
+        required: false
+        default: ''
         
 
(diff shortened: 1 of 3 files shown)


elsirion (approved):
Looks reasonable, feels like it's getting to a level of complexity where a rust rewrite is in order though :sweat_smile:

bradleystachurski:
> Looks reasonable, feels like it's getting to a level of complexity where a rust rewrite is in order though 😅

Yeah, the once minimal script is now doing a lot of things. I'm curious if calling `nix build` inside of rust works well

PR #5933, opened 2024-08-27
fix: deposit `Claimed` state waits for e-cash to be in wallet

elsirion:
Follow-up to #5915.

### DIFF
--- a/modules/fedimint-wallet-tests/tests/tests.rs
+++ b/modules/fedimint-wallet-tests/tests/tests.rs
@@ -27,6 +27,7 @@ use fedimint_wallet_common::{PegOutFees, Rbf};
 use fedimint_wallet_server::WalletInit;
 use futures::stream::StreamExt;
 use secp256k1::rand::rngs::OsRng;
+use tokio::select;
 use tracing::info;
 
 fn fixtures() -> Fixtures {
@@ -156,13 +157,21 @@ async fn on_chain_peg_in_and_peg_out_happy_case() -> anyhow::Result<()> {
     bitcoin.mine_blocks(finality_delay).await;
     await_consensus_to_catch_up(&client, 1).await?;
 
-    let mut balance_sub = initial_peg_in(&client, bitcoin.as_ref(), finality_delay).await?;
+    assert_eq!(client.get_balance().await, sats(0));
+    let (op, address, _) = wallet_module
+        .allocate_deposit_address_expert_only(())
+        .await?;
 
     // Test operation is created
     let operations = client.operation_log().list_operations(10, None).await;
     assert_eq!(operations.len(), 1, "Expecting only the peg-in operation");
+
     let deposit_operation_id = operations[0].0.operation_id;
     let deposit_operation = &operations[0].1;
+    assert_eq!(
+        deposit_operation_id, op,
+        "Operation ID should match the one returned by allocate_deposit_address"
+    );
 
     assert_eq!(
         deposit_operation.operation_module_kind(),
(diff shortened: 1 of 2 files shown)


modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -192,6 +192,10 @@ async fn on_chain_peg_in_and_peg_out_happy_case() -> anyhow::Result<()> {
         deposit_updates.next().await.unwrap(),
         DepositStateV2::WaitingForConfirmation { .. }
     ));
+    assert!(matches!(
+        deposit_updates.next().await.unwrap(),
+        DepositStateV2::Confirmed { .. }
+    ));
```
shaurya947, line 198: nit: would be nice if we can test that after `Claimed` state is emitted, the balance is correct (which would verify that all the e-cash has been issued)

shaurya947 (approved):

modules/fedimint-wallet-tests/tests/tests.rs
elsirion: I added a commit to that end

modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -188,14 +197,57 @@ async fn on_chain_peg_in_and_peg_out_happy_case() -> anyhow::Result<()> {
...
+    info!("Waiting for confirmation");
     assert!(matches!(
         deposit_updates.next().await.unwrap(),
-        DepositStateV2::WaitingForConfirmation { .. }
+        DepositStateV2::WaitingForConfirmation { btc_out_point, .. } if btc_out_point.txid == tx.txid()
     ));
+
+    bitcoin.mine_blocks(finality_delay).await;
+
+    // Afaik technically not necessary, but useful to speed up test (should probably
```
bradleystachurski, line 218: Hmm, I wouldn't have expected this to be necessary since we have a few checks if we're in a test and poll aggressively (1ms or 100ms depending on path).

I'm a bit suspicious of the behavior outside of a test environment now.

bradleystachurski (approved):
Should we backport and create another rc?

modules/fedimint-wallet-tests/tests/tests.rs
bradleystachurski: Interesting, looks like `is_running_in_test_env` returns false when running locally with `cargo test -p fedimint-wallet-tests`

elsirion: I read through the code that determines wake up times and it's wayyyy over-engineered imo. There are two layers of wake up logic:

1. The outer one is told by the lowest of all `next_check_time` when to wake up next. There is also a `min_sleep` that responds to being in a test env, but that only sets a minimum wait before anything can happen, so doesn't help.
2. The inner layer is triggered by the outer and determines which addresses actually need checking. This one checks more often if it runs in the dev env, but the outer layer blocks us from getting there.

I think the logic could be massively simplified, it's not that hard of a problem and the complexity here seems accidental.

fedimint-backports:

```bash
git worktree add -d .worktree/backport-5933-to-releases/v0.4 origin/releases/v0.4
cd .worktree/backport-5933-to-releases/v0.4
git switch --create backport-5933-to-releases/v0.4
git cherry-pick -x 5f0eedb4ce5387976f3f768e37ab4dc1a09a0dbf 9b65bbd7e638f2b93d0b32447270c4320c4c915c
```

PR #5935, opened 2024-08-27
cleanup: rename public_key to module_public_key

joschisan:


elsirion:
I don't see how this makes things more clear, from the user client's PoV the public key belonging to the gateway's LN module is an implementation detail. Maybe we can find a better description? Or just document what it's used for?

joschisan:
@elsirion Well its better then just public_key. I could not come up with a better name; I am aware its an implementation detail but when you are familiar with lnv2 it makes sense.

elsirion:
If there is no more self-descriptive name, can you please at least add some docs to the field? What is it used for?

joschisan:
@elsirion added documentation to all fields.

elsirion (approved):

PR #5936, opened 2024-08-27
chore: add and sort workspace deps

tvolk131:


elsirion (approved):
Would have been nice to do the sorting in a separate commit, but great to move more deps to the workspace!

dpc (approved):

tvolk131:
> Would have been nice to do the sorting in a separate commit, but great to move more deps to the workspace!

I thought about going back in and splitting that commit up but didn't feel like it was worth the time. Always trying to strike a balance between a clean commit history and efficiently utilizing my time. Apologies for the annoying review on this!

PR #5941, opened 2024-08-28
feat: fine grained fees specific to input and output variant

joschisan:
In case we ever want to fine tune this.

### DIFF
--- a/gateway/ln-gateway/src/gateway_module_v2/mod.rs
+++ b/gateway/ln-gateway/src/gateway_module_v2/mod.rs
@@ -28,7 +28,7 @@ use fedimint_lnv2_client::{LightningInvoice, SendPaymentPayload};
 use fedimint_lnv2_common::config::LightningClientConfig;
 use fedimint_lnv2_common::contracts::IncomingContract;
 use fedimint_lnv2_common::{
-    LightningCommonInit, LightningModuleTypes, LightningOutput, LightningOutputV0,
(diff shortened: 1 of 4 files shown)


elsirion (approved):

bradleystachurski:
I've narrowed it down to this commit that's causing upgrade tests to fail. I wouldn't have expected based on the diff. Any hunch why this causes an issue?

```
 Error: command: /nix/store/wzy7zgpqdi31jqfai9wg86k1n2q3pkwj-fedimint-cli/bin/fedimint-cli --data-dir=/tmp/nix-shell.Ovr9mh/devimint-3860923-100/clients/wait-session-client-0 dev session-count

Caused by:
    exit status: 1
    stdout:
    {
      "error": "redecode_raw: instance: 3, kind: lnv2"
    }
```
https://github.com/fedimint/fedimint/actions/runs/10601311644/job/29380818958#step:6:1448

This PR reverts and runs upgrade tests successfully https://github.com/fedimint/fedimint/pull/5952

m1sterc001guy:
I believe it is because this PR changes `FeeConsensus` to `LightningFees` in both `LightningClientConfig` and `LightningConsensusConfig`. This is one of the core module structs, which I believe is decoded when the module is initialized, regardless of its being using or not. I don't think at this point we're attempting API or consensus backwards compatibility for LNv2 yet, but I do think we always need to maintain db backwards compatibility since it will prevent upgrades like this.

IMO we should probably revert this and re-submit with a db migration.

PR #5949, opened 2024-08-28
chore(client): increase size of the session fetching cache LRUs

dpc:
I did some test recoveries on the real Federations and I observed, that due to small cache size wallet & mint recovery fetching diverged and was fetching sessions twice. Not the end of the world, but the whole point of this cache was to avoid doube-downloading, so a waste.

Increasing the size of the LRU, gives enough buffer for everything to reliably stay in sync.


elsirion (approved):

fedimint-backports:
- #5951

PR #5952, opened 2024-08-29
wip: debug upgrade failures

bradleystachurski:
https://github.com/fedimint/fedimint/issues/5948

Reverting fixes, but still investigating the root cause.

PR #5953, opened 2024-08-29
fix: upgrade test fixes

m1sterc001guy:
- Reverts https://github.com/fedimint/fedimint/pull/5941
 - Adds logic for using `get-funding-address` vs `get-ln-onchain-address`

Debugging the same issue as https://github.com/fedimint/fedimint/pull/5952


m1sterc001guy:
Upgrade tests pass https://github.com/fedimint/fedimint/actions/runs/10608262067/job/29402045618?pr=5953

bradleystachurski (approved):

elsirion:
Interesting, how can we deactivate upgrade tests for LNv2 while it doesn't guarantee compatibility yet? cc @joschisan

PR #5957, opened 2024-08-30
Removed the TODO about `balance_msat` field

Mirebella:
I addressed the TODO about renaming `balance_msat`

### DIFF
--- a/gateway/ln-gateway/src/rpc/mod.rs
+++ b/gateway/ln-gateway/src/rpc/mod.rs
@@ -67,7 +67,7 @@ pub struct WithdrawPayload {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct FederationInfo {
     pub federation_id: FederationId,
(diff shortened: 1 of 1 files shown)


elsirion (approved):
Looks good, do we break any of the frontend stuff too much with that change @Kodylow ?

elsirion:
Call: argument against: 
* possibly breaks monitoring systems
* JSON API doesn't have types, so context is useful

Kodylow:
seriously disagree with not including unit generally. Unless we switch everything in its entirety over to msats, then have in docs that every unit is always in msats, then I think we should pass around the unit like we do.

m1sterc001guy:
I would be in favor of just removing the todo, `balance_msat` seems fine to me

Mirebella:
I reverted the field renaming, so the field remains unchanged. Instead I just removed the TODO.

Ready for review.

Kodylow (approved):

PR #5958, opened 2024-08-30
Renamed `incoming_amount` to `incoming_amount_msats`

Mirebella:
I addressed the TODO about renaming `incoming_amount`

### DIFF
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -176,7 +176,7 @@ impl<Cap: Send> GatewayDbtxNcExt for DatabaseTransaction<'_, Cap> {
             &RegisteredIncomingContractKey(contract.commitment.payment_image.clone()),
             &RegisteredIncomingContract {
                 federation_id,
-                incoming_amount: incoming_amount.msats,
+                incoming_amount_msats: incoming_amount.msats,
                 contract,
             },
         )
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/db.rs
```diff
@@ -431,8 +431,7 @@ struct RegisteredIncomingContractKey(pub PaymentImage);
 pub struct RegisteredIncomingContract {
     pub federation_id: FederationId,
     /// The amount of the incoming contract, in msats.
-    /// TODO: Rename to `incoming_amount_msats` to be more explicit.
-    pub incoming_amount: u64,
```
elsirion, line 435: @joschisan why is this a u64 in the first place and not an `Amount`?

elsirion:
Thx for the contribution, trying to figure out if that's actually the correct move here :)

gateway/ln-gateway/src/db.rs
joschisan: Not sure why exactly here - generally we have some u64 msat amounts in the lnv2 module instead for easy interop with ln invoice amounts. Looks like in this case it should actually be an Amount though.

elsirion:
Looks reasonable to me, just needs a rebase :)

Kodylow (approved):
rebased

bradleystachurski:
Is this a safe rename? Is it possible to save the old struct in the db, then fail to read/decode the bytes after this change since the field name changed? If so, then this will require a migration.

bradleystachurski:
Capturing an out-of-band sync with @joschisan: this is safe since we'll decode the struct based on the ordering of the fields, not the field names.

PR #5959, opened 2024-08-30
refactor: remove panic in `get_first_module`

dpc:


maan2003:
love it

elsirion (approved):
That's quite a foot-gun indeed given that federations might just not have certain modules :grimacing:

PR #5963, opened 2024-08-31
Add transasction fees

elsirion:
Reimplementation of #5837 using #5914.

The idea is to phase in new config we may need using new API endpoints the content of which gets synced by the client regularly. That way we can evolve the config structure and values over time.

To activate transaction fees we need to do two things:
1. **Allow over-paying fees**: this is necessary to have a smooth transition where some clients aren't aware of transaction fees while others aren't. We'd be allowing transactions without fees for a transition period after the change kicks in.
2. **A way to activate transaction fees on the federation.**
    * This could be done using a consensus version upgrade as a static value, 
    * or preferably by implementing the server side of dynamic config, which would mean voting on values by peers. The voting would likely happen automatically in this case, but for other values we might want user input.

### DIFF
--- a/fedimint-client/src/extra_config.rs
+++ b/fedimint-client/src/extra_config.rs
@@ -1,0 +1,270 @@
+//! Helper module for handling syncing of dynamically updated config between
+//! client and server. See [`ExtraConfigService`] for more information.
+
+use std::collections::{HashMap, HashSet};
+use std::fmt::Debug;
+use std::sync::Arc;
+use std::time::Duration;
+
+use fedimint_api_client::api::{DynGlobalApi, FederationApiExt};
+use fedimint_core::config::ExtraConfig;
+use fedimint_core::core::ModuleInstanceId;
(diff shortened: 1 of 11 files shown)


fedimint-client/src/extra_config.rs
```diff
@@ -0,0 +1,270 @@
...
+use fedimint_core::task::TaskGroup;
+use fedimint_core::time::now;
+use futures::future::join_all;
+use tracing::warn;
+
+use crate::db::{ExtraConfigKey, ExtraConfigValue};
+
+#[derive(Debug, Default)]
+pub struct ExtraConfigServiceBuilder {
+    tracked_values: HashMap<ExtraConfigKey, (Duration, ValidationFn)>,
```
dpc, line 24: `V` could be just `TrackedConfigValue`?

dpc (approved):

dpc:
I get it now.

elsirion:
We can use this code later on to make everything dynamic, but not highest prio and too much opposition for now.

PR #5967, opened 2024-09-03
fix(recoverytool): don't generate unused change descriptors

elsirion:
We were generating one descriptor per session, as if in each session we had one change output generated. We should only generate change descriptors when peg-outs happened.


dpc (approved):

bradleystachurski (approved):
Shoot, good catch 👍

PR #5969, opened 2024-09-04
chore: improve mprocs startup speed

tvolk131:
Startup Time Before:
66.0s, 66.1s, 74.0s, 65.7s, 66.2s, 74.4s, 73.9s, 64.0s, 74.4s, 63.6s
Average: 68.83s

Startup Time After:
56.4s, 53.9s, 56.2s, 54.2s, 53.9s, 54.4s, 54.5s, 66.3s, 56.0s, 72.1s
Average: 57.79s

Speedup of ~16%

This speedup comes primarily from two things:
* Gateway wait-for-chain-sync command now also triggers a chain sync for LDK (which otherwise syncs every 10 seconds)
* Devimint now collects the TXIDs of channel funding transactions and waits for them all to appear in bitcoind's mempool before mining blocks (whereas before, we just blindly waited a few seconds)

As a minor improvement, we're also opening channels in parallel

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -946,6 +946,14 @@ pub async fn open_channels_between_gateways(
 
     bitcoind.mine_blocks(10).await?;
 
+    debug!(target: LOG_DEVIMINT, "Syncing gateway lightning nodes to chain tip...");
+    futures::future::try_join_all(
+        gateways
+            .iter()
+            .map(|(gw, _gw_name)| gw.wait_for_chain_sync(bitcoind)),
+    )
+    .await?;
+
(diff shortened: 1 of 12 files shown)


m1sterc001guy:
What is the main improvement here that makes mprocs faster? Do you have some numbers on how much faster it is?

tvolk131:
> What is the main improvement here that makes mprocs faster? Do you have some numbers on how much faster it is?

Updated description with answers to both questions!

Kodylow:
also needs to fix the mainnet issue we found where it's slamming the blockstream api every couple seconds

elsirion:
Call: let's close

PR #5971, opened 2024-09-04
refactor: rename mint_channel_id to federation_index

m1sterc001guy:
The concept of a `mint_channel_id` is going away with LNv1, but as https://github.com/fedimint/fedimint/pull/5852 showed it is still useful to have a unique identifier for a federation that is not the federation id, since that is rather long (32 bytes), whereas this identifier is a `u64` and can be variable length encoded.

This PR just renames `mint_channel_id` to `federation_index` to reflect its new use. It will still be used as the short channel id for LNv1 until LNv1 is removed.

### DIFF
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -267,7 +267,7 @@ struct FederationIdKeyV0 {
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable, Serialize, Deserialize)]
 pub struct FederationConfigV0 {
     pub invite_code: InviteCode,
-    pub mint_channel_id: u64,
+    pub federation_index: u64,
     pub timelock_delta: u64,
     #[serde(with = "serde_routing_fees")]
     pub fees: RoutingFees,
@@ -281,7 +281,10 @@ struct FederationIdKey {
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable, Serialize, Deserialize)]
 pub struct FederationConfig {
     pub invite_code: InviteCode,
-    pub mint_channel_id: u64,
+    // Unique integer identifier per-federation that is assigned when the gateways joins a
+    // federation.
+    #[serde(alias = "mint_channel_id")]
+    pub federation_index: u64,
     pub timelock_delta: u64,
     #[serde(with = "serde_routing_fees")]
     pub fees: RoutingFees,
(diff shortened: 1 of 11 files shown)


dpc:
How does it work during restore. How does the gw know to use the exact same mapping. I probably don't understand something simple here. @m1sterc001guy

gateway/ln-gateway/src/db.rs
```diff
@@ -281,7 +281,9 @@ struct FederationIdKey {
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable, Serialize, Deserialize)]
 pub struct FederationConfig {
     pub invite_code: InviteCode,
-    pub mint_channel_id: u64,
+    // Unique integer identifier per-federation that is assigned when the gateways joins a
+    // federation.
+    pub federation_index: u64,
```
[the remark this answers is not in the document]
m1sterc001guy: > Should we add #[serde(alias = "mint_channel_id")] above this to ensure seamless upgradeability? Or is this name change safe as-is?

Added

> Also, should we document that, until LNv1 is removed, this is also used as the SCID?

This is documented at the top of `federation_manager.rs`

modules/fedimint-ln-common/src/lib.rs
```diff
@@ -347,11 +347,12 @@ impl LightningGatewayAnnouncement {
 /// Information a gateway registers with a federation
 #[derive(Debug, Clone, Serialize, Deserialize, Encodable, Decodable, PartialEq, Eq, Hash)]
 pub struct LightningGateway {
-    /// Channel identifier assigned to the mint by the gateway.
+    /// Unique per-federation identifier assigned by the gateway.
     /// All clients in this federation should use this value as
     /// `short_channel_id` when creating invoices to be settled by this
     /// gateway.
-    pub mint_channel_id: u64,
+    #[serde(rename = "mint_channel_id", alias = "federation_index")]
```
tvolk131, line 354: Looks like we're renaming the rust struct property but using the `serde(rename)` to prevent renaming the serialized field, making the `serde(alias)` unneeded. However, I think it makes more sense to change the serialized field by using this instead:

`#[serde(alias = "mint_channel_id")]`

tvolk131 (changes requested):
One more small thing! Sorry, I should have caught this before.

modules/fedimint-ln-common/src/lib.rs
m1sterc001guy: This does not work. I did this initially, and the database backwards compatibility tests in `fedimint-ln-tests` fail. The renaming is necessary for these to succeed.

tvolk131: Gotcha. Then can we get rid of the `serde(alias)` and just have `#[serde(rename = "mint_channel_id")]`?

m1sterc001guy: Sure, that works. Done

tvolk131 (approved):

PR #5972, opened 2024-09-04
feat: database in fedimint wasm client

maan2003:
database taken from fedi pwa. I didn't use webimint database because it uses localstorage which is unavailable in web workers and can block ui for significant time to drop frames


maan2003:
how do we feel about merging this untested? In worse case we have a broken database in fedimint wasm client that nobody uses.

m1sterc001guy:
Will this be maintained here going forward? I just wouldn't want to merge code that won't be maintained here going forward and then have another app develop on top of it thinking that it works

maan2003:
> Will this be maintained here going forward? I just wouldn't want to merge code that won't be maintained here going forward and then have another app develop on top of it thinking that it works

Yes, we plan to maintain fedimint typescript client

m1sterc001guy (approved):
LGTM, I gave it a quick look, looks like this has very basic MVCC and conflict detection on writes. I assume if you're using in fedi pwa that it seems to be working. Would be good to add the tests like we have on the other db crates but I won't block on it

maan2003:
I am just going to merge this to unblock fedimint typescript client, ideally nobody external will use this database

PR #5975, opened 2024-09-04
chore: remove unused dependencies

tvolk131:


fedimint-client/Cargo.toml
```diff
@@ -25,7 +25,6 @@ path = "src/lib.rs"
 
 [dependencies]
 anyhow = { workspace = true }
-aquamarine = { workspace = true }
```
m1sterc001guy, line 28: This is still needed (its why your build is failing) but otherwise great cleanup!

tvolk131: Done! I was running `just check` and `just check-wasm` locally but forgot to run `just build-docs`.

fedimint-core/Cargo.toml
```diff
@@ -60,15 +60,12 @@ url = { workspace = true, features = ["serde"] }
 
 [dev-dependencies]
 test-log = { workspace = true }
-tokio-test = "0.4.4"
```
oleonardolima, line 63: I guess this one's still needed, check the CI failing step.

oleonardolima (commented):
Did you use any specific tool to find out the duplicated ones ?

I wasn't able to find it with udeps in the past (might've used wrongly)

dpc:
How is `cargo udeps` not detecting these?

tvolk131:
> Did you use any specific tool to find out the duplicated ones ?

Nope 😬 just happened to notice a few that weren't needed and decided to go through all `Cargo.toml` files by hand

> How is `cargo udeps` not detecting these?

... ¯\_(ツ)_/¯

fedimint-core/Cargo.toml
tvolk131: Fixed!

dpc (approved):
I'm somewhat amazed. :+1:

elsirion:
Is it possible that `udeps` doesn't detect them because ultimately they are still being pulled in by other deps? At least I can't see a net reduction in dependencies in `Cargo.lock`.

Issue #5954, opened 2024-08-29
Docker images should include basic utils


douglaz:
For instance, it should include `tar` to help remote backups

PR #5976, opened 2024-09-05
chore: add more utils to docker images

douglaz:
Closes https://github.com/fedimint/fedimint/issues/5954

Busybox is very complete. curl and rsync may be also be util.

### DIFF
--- a/nix/flakebox.nix
+++ b/nix/flakebox.nix
@@ -610,6 +610,9 @@ rec {
           pkgs.coreutils
           pkgs.sqlite
           pkgs.fakeNss
+          pkgs.busybox
+          pkgs.curl
+          pkgs.rsync
         ];
         config = {
           Cmd = [ ]; # entrypoint will handle empty vs non-empty cmd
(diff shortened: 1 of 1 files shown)


nix/flakebox.nix
```diff
@@ -610,6 +610,9 @@ rec {
           pkgs.coreutils
           pkgs.sqlite
           pkgs.fakeNss
+          pkgs.busybox
```
oleonardolima, line 613: I wonder if using `busybox` pkg can remove the need for `fakeNss`, need to take a look at it.

m1sterc001guy:
What does `busybox` add?

oleonardolima:
> What does `busybox` add?

@m1sterc001guy It adds a bunch of UNIX utilities (check below), but @douglaz do we need all of that ?

> I know from experience that we only find out we need it when most need it in debugging, so I might be biased. 😅

image

douglaz:
> > What does `busybox` add?
> 
> @m1sterc001guy It adds a bunch of UNIX utilities (check below), but @douglaz do we need all of that ?

It provide many useful commands but it's also very lightweight:
```
 ls -lh /nix/store/6723i29mnr4gib2qp5vahfr7xv7h8ad8-busybox-1.36.1/bin/busybox
-r-xr-xr-x 1 root root 1.2M Jan  1  1970 /nix/store/6723i29mnr4gib2qp5vahfr7xv7h8ad8-busybox-1.36.1/bin/busybox
```
So there is no reason to not add it.

m1sterc001guy (approved):

oleonardolima:
> It provide many useful commands but it's also very lightweight:
> 
> ```
>  ls -lh /nix/store/6723i29mnr4gib2qp5vahfr7xv7h8ad8-busybox-1.36.1/bin/busybox
> -r-xr-xr-x 1 root root 1.2M Jan  1  1970 /nix/store/6723i29mnr4gib2qp5vahfr7xv7h8ad8-busybox-1.36.1/bin/busybox
> ```
> 
> So there is no reason to not add it.

Oh, got it. Thanks!

PR #5981, opened 2024-09-06
chore(wallet): shutdown fedimintd if bitcoin backend is stale

bradleystachurski:
We observed issues with a public federation that had peers with a bitcoin backend that was unable to stay at tip, with some that went two months without contributing a block count vote. Instead of allowing the peers to keep progressing and contributing other consensus items, this change introduces a graceful shutdown if we haven't observed a block that other peers have seen `finality_delay` confirmations.

I'm opening as a draft since I'm still chewing on the implications of these changes. Please flag any concerns with this approach cc: @elsirion @dpc @joschisan

### DIFF
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -802,6 +802,7 @@ pub struct Wallet {
     block_count_rx: watch::Receiver<Option<u32>>,
     /// Fee rate updated periodically by a background task
     fee_rate_rx: watch::Receiver<Feerate>,
+    task_group: TaskGroup,
 }
 
 impl Wallet {
@@ -847,6 +848,7 @@ impl Wallet {
             our_peer_id,
             block_count_rx,
             fee_rate_rx,
+            task_group: task_group.clone(),
         };
 
         Ok(wallet)
@@ -1023,6 +1025,10 @@ impl Wallet {
             blocks_to_go = new_count - old_count,
             "New consensus count, syncing up",
         );
+
+        // Before we can safely call our bitcoin backend to process the new consensus
+        // count, we need to ensure we observed enough confirmations
+        self.wait_for_finality_confs_or_shutdown(new_count).await;
 
         for height in old_count..new_count {
             if height % 100 == 0 {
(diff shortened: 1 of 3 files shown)


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1299,6 +1304,42 @@ impl Wallet {
...
+    /// since we cannot participate in consensus if we're unable to observe the
+    /// chain.
+    ///
+    /// Our bitcoin backend is considered stale if:
+    ///   - it has not seen a block that our peers have observed
+    ///     `finality_delay` confirmations
+    ///   - it is unresponsive after 10 minutes
+    async fn shutdown_if_bitcoin_backend_is_stale(&self, new_consensus_block_count: u32) {
+        let strategy = backoff_util::custom_constant_backoff(Duration::from_secs(10), Some(6 * 10));
+        let get_block_count = || async { Ok(self.btc_rpc.get_block_count().await? as u32) };
```
dpc, line 1331: Ideally, we should not be doing rpcs in consensus processing. We should have our recent block count updated in the database / variable somewhere.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1299,6 +1304,42 @@ impl Wallet {
...
+
+        let backend_is_stale = match retry("get_block_count", strategy, get_block_count).await {
+            Ok(count) if count < new_consensus_block_count => true,
```
dpc, line 1334: For Federations with a small delay this might be too aggressive. It's not impossible to get 3 blocks in succession on mainnet.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1299,6 +1304,42 @@ impl Wallet {
...
+    /// Shutdown the task group shared throughout fedimintd, giving 60 seconds
+    /// for other services to gracefully shutdown.
+    async fn graceful_shutdown(&self) {
+        if let Err(e) = self
+            .task_group
+            .clone()
+            .shutdown_join_all(Some(Duration::from_secs(60)))
+            .await
+        {
+            error!("Error while shutting down fedimintd task group: {e}");
```
dpc, line 1317: Should we just panic?

modules/fedimint-wallet-server/src/lib.rs
bradleystachurski: Good point. I added logic that will consider a backend stale if it's 10 blocks behind the consensus tip regardless of the federation's finality delay. Let me know if you have something else in mind.

modules/fedimint-wallet-server/src/lib.rs
bradleystachurski: Yes we should :slightly_smiling_face:

bradleystachurski:
I added a separate commit (https://github.com/fedimint/fedimint/commit/df6a311acb6e7390cfb0b202516296b362173756) that takes a different approach. We have an issue where we don't guarantee that our bitcoin backend has observed finality delay confirmation prior to querying a block hash for the consensus block count. This commit polls our bitcoin backend until we observe finality delay confirmations and shuts down if we haven't observed after 1 hour.

- is 1 hour too aggressive? (24 hours better?)
- if we're okay with this approach, I'll remove the previous commit and move out of draft

bradleystachurski:
Moving out of draft, let's try to land in 0.5.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1380,6 +1386,53 @@ impl Wallet {
...
+        {
+            panic!("Error while shutting down fedimintd task group: {e}");
+        }
+    }
+
+    /// Returns once our bitcoin backend observes finality delay confirmations
+    /// of the consensus block count. If we don't observe enough confirmations
+    /// after one hour, we gracefully shutdown fedimintd. This is necessary
+    /// since we can no longer participate in consensus if our bitcoin backend
+    /// is unable to observe the same chain tip as our peers.
```
elsirion, line 1407: Just a note: This shouldn't lead to significant latency spikes since we only get consensus on a block height after t/n guardians proposed it. So at worst these `t/n` guardians have higher latency than the typical contributors, but the ordering latency of the height votes makes it very likely that the remaining `f` nodes catch up to the chain tip in the meantime.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1380,6 +1386,53 @@ impl Wallet {
...
+    async fn wait_for_finality_confs_or_shutdown(&self, consensus_block_count: u32) {
+        let backoff = if is_running_in_test_env() {
+            // every 100ms for 60s
+            backoff_util::custom_constant_backoff(Duration::from_millis(100), Some(10 * 60))
+        } else {
+            // every 10s for 1 hour
+            backoff_util::custom_constant_backoff(Duration::from_secs(10), Some(6 * 60))
+        };
```
elsirion, line 1415: nit: I'd prefer a Fibonacci backoff with 10s max to reduce expected latency.

modules/fedimint-wallet-tests/tests/tests.rs
```diff
@@ -617,17 +618,16 @@ async fn peg_ins_that_are_unconfirmed_are_rejected() -> anyhow::Result<()> {
...
 
-    let block_count = dyn_bitcoin_rpc.get_block_count().await?;
+    let block_count = dyn_bitcoin_rpc.get_block_count().await? as u32;
+    let consensus_block_count = block_count - finality_delay;
     sync_wallet_to_block(
         &mut dbtx
             .to_ref_with_prefix_module_id(module_instance_id)
             .into_nc(),
         &mut wallet,
-        block_count.try_into()?,
```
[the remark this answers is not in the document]
elsirion: (I know I somehow commented on the removed code, but the same goes for `as` casts)

elsirion (approved):
I had a pending review here, I deleted most comments, looks mostly goo imo.

Previously I was considering adding a 1 block grace period, but just waiting for the chain tip seems most robust/easiest to analyze. So let's go with that until we see performance problems.

PR #5983, opened 2024-09-06
feat: make gateway api optional

joschisan:
Currently LNV2 api requires you to specify the gateway api by url, makes it optional. Only really used in testing. In practice you would set this to none and module will choose the gateway for you.

Philosophy of LNv2: guardians set list of vetted gateways and module handles selection logic. All gatways are equal but some are more equal than others.

### DIFF
--- a/modules/fedimint-lnv2-client/src/lib.rs
+++ b/modules/fedimint-lnv2-client/src/lib.rs
@@ -410,7 +410,7 @@ impl LightningClientModule {
     async fn routing_info(
         &self,
         gateway_api: &SafeUrl,
-    ) -> Result<Option<RoutingInfo>, GatewayError> {
+    ) -> Result<Option<RoutingInfo>, GatewayConnectionError> {
         self.gateway_conn
             .routing_info(gateway_api.clone(), &self.federation_id)
             .await
@@ -469,7 +469,7 @@ impl LightningClientModule {
                 gateway_api.clone(),
                 self.routing_info(&gateway_api)
                     .await
-                    .map_err(SendPaymentError::GatewayError)?
+                    .map_err(SendPaymentError::GatewayConnectionError)?
                     .ok_or(SendPaymentError::UnknownFederation)?,
             ),
             None => self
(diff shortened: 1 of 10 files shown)


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -388,29 +387,60 @@ fn generate_ephemeral_tweak(static_pk: PublicKey) -> ([u8; 32], PublicKey) {
 }
 
 impl LightningClientModule {
+    async fn select_gateway(&self) -> Result<(SafeUrl, RoutingInfo), SelectGatewayError> {
```
m1sterc001guy, line 390: Might be a good idea to add a comment that this is a stateless command

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -547,6 +584,8 @@ impl LightningClientModule {
         panic!("We could not find an unused operation id for sending a lightning payment");
     }
 
+    /// Subscribe to all updates of the send operation. No specific order is not
```
m1sterc001guy, line 587: "No specific order is not guaranteed" is a little akward phrasing

m1sterc001guy (approved):
Just nits on some comments

elsirion:
> Philosophy of LNv2: guardians set list of vetted gateways and module handles selection logic. All gatways are equal but some are more equal than others.

Can gateways still register without guardian intervention btw? This is a very useful feature to be able to provide "fallback gateways" when guardians may not be around to fix the situation.

m1sterc001guy:
Not currently. The main issue is around DoS and user experience, when gateways can register without guardian approval, it is possible to either spam the guardian's DB with fake registrations or register gateways that have no inbound ecash, causing payments to fail.

@joschisan and I have chatted about some potential improvements. One idea was to pay for a registration with ecash, another was to submit a "proof" that proves you hold some threshold of ecash (proof here is very hand-wavy).

joschisan:
Guardians would have to set multiple gateways then the gatways fall back automatically.

PR #5991, opened 2024-09-09
Session tokens for admin api auth

### LINKED ISSUE #4128: Use session tokens for admin API auth
Users would acquire a session token through a login API endpoint. This token would stay valid for a limited time and be sent with every API request like the password is currently.

This allows:
* More elegant login attempt rate limiting (#4122)
* 2FA login

f3r10:
closes #4128


f3r10:
Hi @Kodylow sorry for the delay.  I was a little busy at work.
As I understand the PR is almost done. The flow should be something like this:
A client hits `/auth/challenge` and the gateway returns a challenge code.
A client sends back the previous challenge code to `/auth/session` and after validation a JWT returns.

For the private endpoints, a client should include the JWT in the request header, and if it is valid the client has access to the private content.

The only thing that I still do not understand is which `EncodingKey`  should be used for encoding the JWT :thinking: 

At first, I was confused about the issue :sweat_smile: but reading https://www.oauth.com/ helped me to understand how the flow should be.

Please correct me if this is the correct approach or if I misunderstood something.

Issue #5980, opened 2024-09-06
`GatewayBalances` should include inbound liquidity


m1sterc001guy:
This is arguably just as important as the `lightning_balance_msats`, since it determines if the gateway is able to service incoming payments. This can be deduced from `ListActiveChannels` but I think it would be useful to have an "at a glance" value like we do for `lightning_balance_msats`

```
#[derive(serde::Serialize, serde::Deserialize)]
pub struct GatewayBalances {
    pub onchain_balance_sats: u64,
    pub lightning_balance_msats: u64,
    pub ecash_balances: Vec<FederationBalanceInfo>,
}
```

m1sterc001guy:
Just send out a PR when you have something working. Be sure to test it locally using mprocs to make sure it works.

mubarak23:
@m1sterc001guy what is mprocs

PR #5994, opened 2024-09-10
feat(balance): add inbound lightning liquidity to GetBalancesResponse

mubarak23:
PR for this issue: #5980 
* add inbound lightning liquidity  to Get Balances Response

### DIFF
--- a/gateway/ln-gateway/proto/gateway_lnrpc.proto
+++ b/gateway/ln-gateway/proto/gateway_lnrpc.proto
@@ -332,4 +332,7 @@ message GetBalancesResponse {
 
     // The total balance of the node's lightning wallet, in millisats.
     uint64 lightning_balance_msats = 2;
+
+    // The total inbound Liquidity balance.
+    uint64 inbound_lightning_liquidity_msats = 3;
 }
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -1166,6 +1166,35 @@ impl GatewayLightning for ClnRpcService {
             })?
             .map_err(|e| tonic::Status::internal(e.to_string()))?;
 
+        let total_receivable_msat = self
+            .rpc_client()
+            .await
+            .map_err(|e| Status::internal(e.to_string()))?
+            .call(cln_rpc::Request::ListPeerChannels(
+                model::requests::ListpeerchannelsRequest { id: None },
+            ))
+            .await
+            .map(|response| match response {
+                cln_rpc::Response::ListPeerChannels(
+                    model::responses::ListpeerchannelsResponse { channels },
+                ) => Ok(channels
+                    .into_iter()
+                    .filter(|channel| {
+                        matches!(
+                            channel.state,
+                            model::responses::ListpeerchannelsChannelsState::CHANNELD_NORMAL
+                        )
+                    })
+                    .filter_map(|channel| channel.receivable_msat.map(|value| value.msat()))
+                    .sum::<u64>()), // Sum the receivable_msat values directly
+                _ => Err(ClnExtensionError::RpcWrongResponse),
+            })
+            .map_err(|e| {
+                error!("cln listchannels rpc returned error {:?}", e);
+                tonic::Status::internal(e.to_string())
+            })?
+            .map_err(|e| tonic::Status::internal(e.to_string()))?;
+
         let lightning_balance_msats = channels
             .into_iter()
             .fold(0, |acc, channel| acc + channel.our_amount_msat.msat());
@@ -1176,6 +1205,7 @@ impl GatewayLightning for ClnRpcService {
         Ok(tonic::Response::new(GetBalancesResponse {
             onchain_balance_sats,
             lightning_balance_msats,
+            inbound_lightning_liquidity_msats: total_receivable_msat,
         }))
     }
 }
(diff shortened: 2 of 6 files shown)


gateway/ln-gateway/proto/gateway_lnrpc.proto
```diff
@@ -332,4 +332,6 @@ message GetBalancesResponse {
 
     // The total balance of the node's lightning wallet, in millisats.
     uint64 lightning_balance_msats = 2;
+    // the total inbound Liquidity balance
```
m1sterc001guy, line 335: Nit: add newline above here to be consistent with the rest of the file

m1sterc001guy:
Looks pretty good to me. Can you add an assert in an existing test to verify that this works?

mubarak23:
> Looks pretty good to me. Can you add an assert in an existing test to verify that this works?

can you point me to direct test file where `handle_get_balances_msg` is tested

tvolk131:
> Can you add an assert in an existing test to verify that this works?

I don't think we currently have a test for the `get-balances` command. I can add one as a follow-up to this PR, but since we're not testing that yet anyway I don't think it's a blocker for this. I'll test this PR on my local machine before approving.

gateway/ln-gateway/proto/gateway_lnrpc.proto
tvolk131: > gateway_lnrpc.proto:335: trailing whitespace

You need to remove the trailing whitespace before committing. Your editor probably added it when you added a new line.

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -1165,6 +1165,9 @@ impl GatewayLightning for ClnRpcService {
                 tonic::Status::internal(e.to_string())
             })?
             .map_err(|e| tonic::Status::internal(e.to_string()))?;
+        let inbound_lightning_liquidity_msats = channels.clone().iter().fold(0, |acc, channel| {
+            acc + (channel.amount_msat.msat() - channel.our_amount_msat.msat())
+        });
```
elsirion, line 1170: * Why clone here? `.iter()` takes a reference to `self`, so that cloned version will be dropped right away.
* Why use `fold` instead of `map` and `sum`?
* You aren't taking into account the channels reserve afaik. Instead of calculating the total amount receivable yourself you could use `receivable_msat` returned by `listpeerchannels` (it's an additional API call, but more likely to give us a correct picture. thoughts @m1sterc001guy @tvolk131?).

tvolk131: Your third point makes sense. @mubarak23 sorry for the misdirection! You'll want to call this RPC: https://docs.corelightning.org/reference/lightning-listpeerchannels

m1sterc001guy: @mubarak23 any update on this? Would love to get it in soon

mubarak23: @m1sterc001guy yes, i work on this but am have some issue with `nix develop ` command

elsirion (approved):
Looks good!

gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -1166,6 +1166,35 @@ impl GatewayLightning for ClnRpcService {
...
+                ) => Ok(channels
+                    .into_iter()
+                    .filter(|channel| {
+                        matches!(
+                            channel.state,
+                            model::responses::ListpeerchannelsChannelsState::CHANNELD_NORMAL
+                        )
+                    })
+                    .filter_map(|channel| channel.receivable_msat.map(|value| value.msat()))
+                    .sum::<u64>()), // Sum the receivable_msat values directly
```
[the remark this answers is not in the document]
mubarak23: Yes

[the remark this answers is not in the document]
mubarak23: what could be the best data type to use here, u64 is the best option i can think of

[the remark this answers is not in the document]
elsirion: We could do it this way, but now that I noticed that the final API response uses a `u64` it's not super critical :man_shrugging: 

```diff
diff --git a/gateway/ln-gateway/src/bin/cln_extension.rs b/gateway/ln-gateway/src/bin/cln_extension.rs
index b66174de36..4533d5ff11 100644
--- a/gateway/ln-gateway/src/bin/cln_extension.rs
+++ b/gateway/ln-gateway/src/bin/cln_extension.rs
@@ -1185,8 +1185,8 @@ async fn get_balances(
                             model::responses::ListpeerchannelsChannelsState::CHANNELD_NORMAL
                         )
                     })
-                    .filter_map(|channel| channel.receivable_msat.map(|value| value.msat()))
-                    .sum::<u64>()), // Sum the receivable_msat values directly
+                    .filter_map(|channel| channel.receivable_msat)
+                    .sum::<Amount>()), // Sum the receivable_msat values directly
                 _ => Err(ClnExtensionError::RpcWrongResponse),
             })
             .map_err(|e| {
@@ -1205,7 +1205,7 @@ async fn get_balances(
         Ok(tonic::Response::new(GetBalancesResponse {
             onchain_balance_sats,
             lightning_balance_msats,
-            inbound_lightning_liquidity_msats: total_receivable_msat,
+            inbound_lightning_liquidity_msats: total_receivable_msat.msats,
         }))
     }
 }
```

Issue #5965, opened 2024-09-03
LDK Gateway spamming esplora and killing ldk-node


Kodylow:
Found while trying to run a mainnet ldk gateway:

```
gatewayd-1    | 2024-09-03T20:54:20.699562Z  WARN ln_gateway: Failed to retrieve Lightning info: LightningRpcError(FailedToGetNodeInfo { failure_reason: "Failed get chain tip block summary: Reqwest(reqwest::Error { kind: Status(429), url: Url { scheme: \"https\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"blockstream.info\")), port: None, path: \"/api/blocks\", query: None, fragment: None } })" })
gatewayd-1    | 2024-09-03T20:54:20.701770Z  INFO ln_gateway::lightning::ldk: Stopping LDK Node...
gatewayd-1    | thread 'tokio-runtime-worker' panicked at /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/tokio-1.40.0/src/runtime/scheduler/multi_thread/mod.rs:86:9:
gatewayd-1    | Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
gatewayd-1    | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Need to set the esplora calls to something reasonable when not on regtest, add a backoff, and not kill the ldk-node if the esplora calls fail

elsirion:
Is it about retries or the rate limits of e.g. mempool.space? I ended up using https://mempool.emzy.de/ for my LDK node GW and also set up https://mempool.sirion.io/ in the meantime to not be constrained by API rate limits anymore.

Kodylow:
I think it's both, blockstream's api limit + we're not doing any backoff / retry logic around it

oleonardolima:
It hasn't landed on master or any release yet, but this PR covers this problem: https://github.com/bitcoindevkit/rust-esplora-client/pull/71. It could be useful as a starting point.

oleonardolima:
Another reference that has just been posted @ https://github.com/bitcoindevkit/bdk/issues/1120#issuecomment-2331794726

> FYI
>
[...]

PR #5997, opened 2024-09-11
chore: add chain sync command to ILnRpcClient

tvolk131:
Allow gateway to trigger an on-chain wallet sync on the underlying lightning node. Only implemented for LDK since LND and CLN automatically sync.

This is needed for LDK since `ldk-node::Node` gets its on-chain data from esplora rather than bitcoind and therefore needs to avoid getting rate-limited by esplora by not pinging it too frequently.

This change speeds up `just mprocs` and should allow us to remove the following overrides which were put in place to prevent mprocs startup speed from almost tripling (and which I believe are the cause of #5965):

https://github.com/fedimint/fedimint/blob/06717c0a480f619feadb9b411dd57267fbb03e98/gateway/ln-gateway/src/lightning/ldk.rs#L76-L77

However, attempting to remove those overrides in this PR led to a doubling of mprocs startup time (~100s). I'm not sure why. I left a comment on this `ldk-node` issue with some more details: https://github.com/lightningdevkit/ldk-node/issues/339#issuecomment-2344230472

Before:
55.8s, 54.5s, 56.0s, 54.3s, 56.4s, 54.8s, 54.5s, 56.5s, 63.8s, 54.1s
Average: 56.07s

After:
51.3s, 48.6s, 49.5s, 55.3s, 52.5s, 45.4s, 56.3s, 52.2s, 50.1s, 52.6s
Average: 51.38s

Average speed improvement of ~8%

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -26,7 +26,9 @@ use tonic_lnd::lnrpc::{ChanInfoRequest, GetInfoRequest, ListChannelsRequest};
 use tonic_lnd::Client as LndClient;
 use tracing::{debug, info, trace, warn};
 
-use crate::util::{poll, ClnLightningCli, GatewayClnExtension, ProcessHandle, ProcessManager};
+use crate::util::{
+    poll, poll_with_timeout, ClnLightningCli, GatewayClnExtension, ProcessHandle, ProcessManager,
+};
 use crate::vars::utf8;
 use crate::version_constants::VERSION_0_4_0_ALPHA;
 use crate::{cmd, poll_eq, Gatewayd};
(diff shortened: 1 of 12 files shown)


devimint/src/external.rs
```diff
@@ -915,6 +915,14 @@ pub async fn open_channels_between_gateways(
 
     bitcoind.mine_blocks(10).await?;
 
+    debug!(target: LOG_DEVIMINT, "Syncing gateway lightning nodes to chain tip...");
```
tvolk131, line 918: Note for reviewers: We're adding this before waiting for channel readiness because it now triggers a chain sync, where as waiting for channel readiness does not.

dpc (approved):

PR #5999, opened 2024-09-11
refactor: move handle_htlc_stream() logic into route_htlcs()

tvolk131:
`Gateway::route_htlcs()` is the only place where `Gateway::handle_htlc_stream()` is called. Having both functions is a bit confusing, and `Gateway::handle_htlc_stream()` does very little. (A good chunk of its logic is to respect shutdown signals, but `Gateway::route_htlcs()` already calls it in a cancellable way anyway, so that's not even needed.) Let's inline it into `Gateway::route_htlcs()`.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -478,7 +478,7 @@ impl Gateway {
     async fn route_htlcs<'a>(
         &'a self,
         handle: &TaskHandle,
-        stream: RouteHtlcStream<'a>,
+        mut stream: RouteHtlcStream<'a>,
         ln_client: Arc<dyn ILnRpcClient>,
     ) -> ReceivePaymentStreamAction {
         let (lightning_public_key, lightning_alias, lightning_network) =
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -533,14 +533,31 @@ impl Gateway {
...
 
-        // Blocks until the connection to the lightning node breaks or we receive the
-        // shutdown signal
+        // Runs until the connection to the lightning node breaks or we receive the
+        // shutdown signal.
         if handle
-            .cancel_on_shutdown(self.handle_htlc_stream(stream, handle.clone()))
+            .cancel_on_shutdown(async move {
+                loop {
+                    let htlc_request = match stream.next().await {
```
m1sterc001guy, line 547: For the gateway safe stop feature to work, we still need to check if the gateway is in the running state and break out of this loop if it is not.

m1sterc001guy (changes requested):
Need to check gateway state inside of the loop

m1sterc001guy (approved):

gateway/ln-gateway/src/lib.rs
elsirion: I assume you refer to safe stopping where the last HTLCs are still handled but no new ones are accepted?

What about the following order of events:
1. Check here that the GW is still running
2. Wait for the next HTLC event from the stream
3. Initiate a safe shutdown
4. HTLC event is received from stream

Wouldn't we still start processing that HTLC? Maybe I'm misunderstanding the architecture/that feature, but I'd expect a `select!` over `stream.next()` and `is_shutting_down_safely`. Afaiu we cannot rely on the built-in shutdown behavior of `cancel_on_shutdown` because we need to stop HTLC processing first, then wait for in-flight HTLCs and finally shut down normally.

PR #6001, opened 2024-09-12
feat: relative fees for lnv2

joschisan:


bradleystachurski:
Kicked off upgrade tests in https://github.com/fedimint/fedimint/pull/6002

modules/fedimint-lnv2-common/src/config.rs
```diff
@@ -63,6 +63,34 @@ pub struct LightningConfigPrivate {
...
+
+impl FeeConsensus {
+    pub fn fee(&self, amount: Amount) -> Amount {
+        Amount::from_msats(self.fee_msats(amount.msats))
+    }
+
+    fn fee_msats(&self, msats: u64) -> u64 {
+        self.base.msats
+            + msats
+                .saturating_mul(self.parts_per_million)
```
bradleystachurski, line 80: It might be worth capturing in a comment our conversation from the deep dive about why we're okay using `saturating_mul` with `u64`. Also worth surfacing in the PR that @shaurya947 had a good suggestion for casting to `u128` then back to `u64`.

bradleystachurski (approved):
Upgrade tests look good :+1:

PR #6003, opened 2024-09-12
fix: find_by_prefix should guarantee order

dpc:
It's already the case, yet in the interface documentation we don't commit to it. I'm quite sure there are already parts of the code that depend on it anyway.

In a system like ours any form of non-determism is unwelcomed. I rather have the db impls have to do the sorting, than all users need to sort. Especially that for all kv stores, sorting should come naturally.

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -797,7 +797,7 @@ pub trait IDatabaseTransactionOpsCore: MaybeSend {
     async fn raw_remove_entry(&mut self, key: &[u8]) -> Result<Option<Vec<u8>>>;
 
     /// Returns an stream of key-value pairs with keys that start with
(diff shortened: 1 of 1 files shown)


m1sterc001guy (approved):

fedimint-core/src/db/mod.rs
```diff
@@ -797,7 +797,7 @@ pub trait IDatabaseTransactionOpsCore: MaybeSend {
     async fn raw_remove_entry(&mut self, key: &[u8]) -> Result<Option<Vec<u8>>>;
 
     /// Returns an stream of key-value pairs with keys that start with
-    /// `key_prefix`. No particular ordering is guaranteed.
+    /// `key_prefix`, sorted by key.
```
elsirion, line 800: Good catch, that was certainly a documentation bug (heck, there's even a version of this function that guarantees the reverse ordering :laughing:)

PR #6005, opened 2024-09-12
chore(db): `find_by_range`

dpc:


fedimint-core/src/db/mem_impl.rs
```diff
@@ -178,6 +179,18 @@ impl<'a> IDatabaseTransactionOpsCore for MemTransaction<'a> {
 
         Ok(Box::pin(stream::iter(data)))
     }
+
+    async fn raw_find_by_range(&mut self, range: Range<&[u8]>) -> Result<PrefixStream<'_>> {
```
elsirion, line 183: I assume it's implicit in the type, but would be nice to document inclusiveness/exclusiveness on both `raw_find_by_range` and `find_by_range`.

dpc: We seem to document only `raw_`, ATM, so I kept it this way.

fedimint-rocksdb/src/lib.rs
```diff
@@ -234,6 +235,42 @@ impl<'a> IDatabaseTransactionOpsCore for RocksDbTransaction<'a> {
         }))
     }
 
+    async fn raw_find_by_range(&mut self, range: Range<&[u8]>) -> Result<PrefixStream<'_>> {
+        // turn an `iter` into a `Stream` where every `next` is ran inside
+        // `block_in_place` to offload the blocking calls
+        fn convert_to_async_stream<'i, I>(iter: I) -> impl futures::Stream<Item = I::Item>
```
bradleystachurski, line 241: [nit] Good opportunity to move `convert_to_async_stream` higher then remove all the duplicate implementations

bradleystachurski (approved):

PR #6006, opened 2024-09-12
fix(ln-client): verify operation type when subscribing to LN payment updates

### LINKED ISSUE #6000: Panic in LN update stream
A user hit this panic in the wild:

https://github.com/fedimint/fedimint/blob/6e552f42657474325cc4b4980d21e577adbd5fd1/modules/fedimint-ln-client/src/lib.rs#L1223-L1231

It's unclear to me why there'd suddenly be a state that isn't expected (we check that we are dealing with the correct operation kind earlier), but fundamentally we should likely not panic on non-critical client programmer errors. Logging an error and returning none would be better here imo.

I still think that panics are great for critical code paths where broken invariants can cause loss of funds, but e.g. update streams are purely informational and inconsistencies are preferable to crashing the whole application.

elsirion:
Fixes #6000

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -1185,6 +1185,23 @@ impl LightningClientModule {
         operation_id: OperationId,
     ) -> anyhow::Result<UpdateStreamOrOutcome<InternalPayState>> {
         let operation = self.client_ctx.get_operation(operation_id).await?;
+
+        let LightningOperationMetaVariant::Pay(LightningOperationMetaPay {
+            out_point: _,
+            invoice: _,
+            change: _, // FIXME: why isn't this used here?
+            is_internal_payment,
+            ..
+        }) = operation.meta::<LightningOperationMeta>().variant
+        else {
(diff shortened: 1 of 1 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -1185,6 +1185,23 @@ impl LightningClientModule {
         operation_id: OperationId,
     ) -> anyhow::Result<UpdateStreamOrOutcome<InternalPayState>> {
         let operation = self.client_ctx.get_operation(operation_id).await?;
+
+        let LightningOperationMetaVariant::Pay(LightningOperationMetaPay {
+            out_point: _,
+            invoice: _,
+            change: _, // FIXME: why isn't this used here?
```
m1sterc001guy, line 1192: I believe the outpoint for this isn't used because it is passed through the `RefundSubmitted` state.

m1sterc001guy (approved):

elsirion:
This lead to quite some pain, so possibly worth backporting if it applies cleanly.

fedimint-backports:
- #6046

PR #6008, opened 2024-09-12
chore: devfed waits for channel funding txs before mining blocks

tvolk131:
Change `ILnRpcClient::open_channel()` to return the channel funding TXID, and set devfed to wait for those TXIDs to be present in the bitcoind mempool before mining blocks. Previously we were just blindly waiting, requiring that we waited significantly longer than needed in order to prevent flakiness.

Before:
66.0s, 74.0s, 64.2s, 64.3s, 66.2s, 66.7s, 63.8s, 64.0s, 64.0s, 66.4s
Average: 65.96s

After:
55.8s, 54.5s, 56.0s, 54.3s, 56.4s, 54.8s, 54.5s, 56.5s, 63.8s, 54.1s
Average: 56.07s

Average speed improvement of ~15%

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -900,18 +900,49 @@ pub async fn open_channels_between_gateways(
         gateways.iter().circular_tuple_windows::<(_, _)>().collect()
     };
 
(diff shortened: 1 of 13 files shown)


dpc (approved):
Very nice.

PR #6010, opened 2024-09-13
cleanup: cleanup lnv2 module client api

joschisan:
A bunch of API cleanup, methods not used yet become private for now. Most importantly I added the invoice to the receive operation meta and provided a method to calculate the fees paid to the gateway from the meta.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -1801,12 +1801,15 @@ impl Gateway {
                 // The base fee ensures that the gateway does not loose sats sending the payment due
                 // to fees paid on the transaction claiming the outgoing contract or
                 // subsequent transactions spending the newly issued ecash
-                send_fee_minimum: PaymentFee::SEND_FEE_MINIMUM,
+                send_fee_minimum: PaymentFee {
+                    base: Amount::from_sats(50),
+                    parts_per_million: 5_000,
+                },
+                expiration_delta_default: EXPIRATION_DELTA_LIMIT_DEFAULT,
+                expiration_delta_minimum: EXPIRATION_DELTA_MINIMUM_V2,
                 // The base fee ensures that the gateway does not loose sats receiving the payment
                 // due to fees paid on the transaction funding the incoming contract
                 receive_fee: PaymentFee::RECEIVE_FEE_LIMIT_DEFAULT,
-                expiration_delta_default: EXPIRATION_DELTA_LIMIT_DEFAULT,
-                expiration_delta_minimum: EXPIRATION_DELTA_MINIMUM_V2,
             }))
     }
 
@@ -1851,9 +1854,7 @@ impl Gateway {
             ));
         }
 
-        let contract_amount = payment_info
-            .receive_fee
-            .subtract_fee(payload.invoice_amount.msats);
+        let contract_amount = payment_info.receive_fee.subtract_from(payload.amount.msats);
 
         if contract_amount == Amount::ZERO {
             return Err(GatewayError::IncomingContractError(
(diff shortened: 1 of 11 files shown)


gateway/ln-gateway/src/lib.rs
```diff
@@ -1851,9 +1854,7 @@ impl Gateway {
             ));
         }
 
-        let contract_amount = payment_info
-            .receive_fee
-            .subtract_fee(payload.invoice_amount.msats);
+        let contract_amount = payment_info.receive_fee.subtract_fee(payload.amount.msats);
```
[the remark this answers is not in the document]
dpc: `fee.substract_from(amount)`  could work as a name too, I think.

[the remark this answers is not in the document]
joschisan: Renamed them according to @dpc's suggestion. Since these are used on the client and on the guardian I do not want to inline. Also saturation is required here.

dpc (approved):

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -217,44 +261,37 @@ impl RoutingInfo {
...
         parts_per_million: 15_000,
     };
 
+    /// This is the maximum receive fee of half of one percent plus fifty
+    /// satoshis a correct gateway may recommend as a default.
     pub const RECEIVE_FEE_LIMIT_DEFAULT: PaymentFee = PaymentFee {
-        base: DUST_LIMIT,
+        base: Amount::from_sats(50),
         parts_per_million: 5_000,
     };
```
elsirion, line 284: I assume all these defaults are adjustable? Seem way too high, but not worth discussing if operators can just lower them.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -697,69 +733,44 @@ impl LightningClientModule {
...
-    /// recipient's static module public key and fetches the corresponding
-    /// invoice. To receive the payment the recipient needs to call
-    /// [`Self::receive_incoming_contract`] on the returned contract. For
-    /// testing you can optionally specify a gateway to generate the
-    /// invoice, otherwise a gateway will be selected automatically.
-    pub async fn create_contract_and_fetch_invoice(
+    // Create an incoming contract locked to a public key derived from the
+    // recipient's static module public key and fetches the corresponding
+    // invoice.
+    async fn create_contract_and_fetch_invoice(
```
elsirion, line 765: Why isn't it a doc comment anymore?

modules/fedimint-lnv2-client/src/lib.rs
joschisan: operators can adjust them. its just a client side limit.

modules/fedimint-lnv2-client/src/lib.rs
joschisan: Method is not public anymore.

elsirion: It's still useful to have it be a docstring since it will integrate better with tooling that can display docs (e.g. https://docs.fedimint.org/ renders private items). Not blocking, but for future PRs.

elsirion (dismissed):

elsirion:
You need to make the `serde_json` dep in `ln_v2_client` non-optional for CI to succeed. For most builds it's activated implicitly by other deps (feature unification …), but for containers we seem to use a smaller set of crates.

Screenshot 2024-09-19 at 12-03-00 cleanup cleanup lnv2 module client api · fedimint_fedimint@ab70d09

elsirion (approved):

PR #6012, opened 2024-09-13
fix: enforce block height maximalism

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-client/src/api.rs
+++ b/modules/fedimint-lnv2-client/src/api.rs
@@ -14,7 +14,7 @@ use fedimint_core::{apply, async_trait_maybe_send, Amount, NumPeersExt, PeerId};
 use fedimint_lnv2_common::contracts::{IncomingContract, OutgoingContract};
 use fedimint_lnv2_common::endpoint_constants::{
     ADD_GATEWAY_ENDPOINT, AWAIT_INCOMING_CONTRACT_ENDPOINT, AWAIT_PREIMAGE_ENDPOINT,
-    CONSENSUS_BLOCK_COUNT_ENDPOINT, CREATE_BOLT11_INVOICE_ENDPOINT, GATEWAYS_ENDPOINT,
+    CONSENSUS_BLOCK_HEIGHT_ENDPOINT, CREATE_BOLT11_INVOICE_ENDPOINT, GATEWAYS_ENDPOINT,
     OUTGOING_CONTRACT_EXPIRATION_ENDPOINT, REMOVE_GATEWAY_ENDPOINT, ROUTING_INFO_ENDPOINT,
     SEND_PAYMENT_ENDPOINT,
 };
@@ -32,7 +32,7 @@ const RETRY_DELAY: Duration = Duration::from_secs(1);
 
 #[apply(async_trait_maybe_send!)]
 pub trait LnFederationApi {
-    async fn consensus_block_count(&self) -> FederationResult<u64>;
+    async fn consensus_block_height(&self) -> FederationResult<u64>;
 
     async fn await_incoming_contract(&self, contract_id: &ContractId, expiration: u64) -> bool;
 
@@ -57,9 +57,9 @@ impl<T: ?Sized> LnFederationApi for T
 where
     T: IModuleFederationApi + MaybeSend + MaybeSync + 'static,
 {
-    async fn consensus_block_count(&self) -> FederationResult<u64> {
+    async fn consensus_block_height(&self) -> FederationResult<u64> {
         self.request_current_consensus(
-            CONSENSUS_BLOCK_COUNT_ENDPOINT.to_string(),
+            CONSENSUS_BLOCK_HEIGHT_ENDPOINT.to_string(),
             ApiRequestErased::new(()),
         )
         .await
--- a/modules/fedimint-lnv2-client/src/lib.rs
+++ b/modules/fedimint-lnv2-client/src/lib.rs
@@ -499,16 +499,16 @@ impl LightningClientModule {
             ));
         }
 
-        let consensus_block_count = self
+        let consensus_block_height = self
             .module_api
-            .consensus_block_count()
+            .consensus_block_height()
             .await
             .map_err(|e| SendPaymentError::FederationError(e.to_string()))?;
 
         let contract = OutgoingContract {
             payment_image: PaymentImage::Hash(*invoice.payment_hash()),
             amount: send_fee.add_fee(amount),
-            expiration: consensus_block_count + expiration_delta + CONTRACT_CONFIRMATION_BUFFER,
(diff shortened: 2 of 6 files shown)


dpc:
Given that this just replaces names, either previous or current use is incorrect somewhere/partially.

To explain again the whole issue and rationale:

In Bitcoin Core "get block count" is retarded (I'm sorry, but that just the only word I can use):

```
❯ bitcoin-cli getblockcount
861164
❯ bitcoin-cli getblockhash 861164
000000000000000000000f2bbb43f689038ddb8767225cad09be4af02049f624
```

In normal software engineering:

```rust
fn main() {
  let blocks = vec!["a", "b", "c"];
  println!("block count: {}", blocks.len());
  println!("block at block count: {}", blocks[blocks.len()]);
}
```

```
block count: 3
thread 'main' panicked at src/main.rs:4:46:
index out of bounds: the len is 3 but the index is 3
```

Bitcoin Core is kind of stupid, here and in many other places.

So in Fedimint "block count" returns what `blocks.len()` would mean - the height/index of the first block that isn't there yet.

If you refer to the index of a particular block, it's OK to call it "height". This is the same between Bitcoin Core and Fedimnint.

```
❯ bitcoin-cli getblockhash 0
000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
```

Block at height/index `0` is the genesis block.

But if you refer to the number of blocks total, I strongly suggest using "count" in the classical sense.

The tradition of indexing from `0` and using half-open ranges everywhere (and thus "count" and not "index of last element" to track length/count of collections) is as old as as it gets, and was enshrined in the scriptes many times, notably including: https://www.cs.utexas.edu/~EWD/transcriptions/EWD08xx/EWD831.html . Inventing Bitcoin does not justify changing it, so Satoshi can suck an egg for all I care.

joschisan:
> @joschisan
> 
> Given that this just replaces names, either previous or current use is incorrect somewhere/partially.

We just need to lock the contract for a certain number of blocks as we are only working with deltas here. Actual value is irrelevant as long as it progresses on average as fast a the block height.

dpc:
> We just need to lock the contract for a certain number of blocks as we are only working with deltas here. Actual value is irrelevant as long as it progresses on average as fast a the block height.

The name needs to match the meaning of the value. Either it's the count of blocks on chain, or a height/index of some block.

joschisan:
Since block height is an established concept I would go with this on in this case. To be correct I subtract one from the block count as returned by the rpc. @dpc fine with you here?

bradleystachurski:
> The name needs to match the meaning of the value. Either it's the count of blocks on chain, or a height/index of some block.

I agree, that's my main concern and the issue that popped up with `sync_up_to_consensus_height`.

Out-of-scope for this PR, but perhaps it would be useful to define new types in `fedimint_core` for `BlockCount` and `BlockHeight`. Instead of using `u64` and `u32` throughout the codebase where we're expecting counts and heights, we could use the new types.

dpc:
> Since block height is an established concept I would go with this on in this case. To be correct I subtract one from the block count as returned by the rpc. @dpc fine with you here?

I'm not excited about it, but will not block it. :D

joschisan:
> issue that popped up with sync_up_to_consensus_height.

what issue?

bradleystachurski:
> what issue?

As discussed on the deep dive yesterday, we pass consensus block counts to `sync_up_to_consensus_height`, that then passes the block count to fetch the block hash, however the rpc expects a block height.

I don't believe this causes a bug in the wallet since we're stilling syncing.

dpc:
> I agree, that's my main concern and the issue that popped up with `sync_up_to_consensus_height`.

@bradleystachurski 

`sync_up_to_consensus_height` is a great example of where the silliness of using "block height" to describe number of elements in the sequence spills over into confusing code and potential bugs.

In normal software engineering, it would natural to use `count` (half-inclusive range) to do stuff. If you look at e.g. https://doc.rust-lang.org/std/primitive.slice.html#method.split_at , the `mid`, has the meaning of "count", not "height". 

That function should have been called `sync_up_to_consensus_count`, and `<` used everywhere. But because just "block height is an established concept", now it's necessary to write code out of character, to accommodate satoshisms.

> fedimint_core for BlockCount and BlockHeight

Newtype like that would be nice, but I'm more and more convinced that using `BlockCount` vs `BlockIndex` would be best, to avoid "height" confusion.

joschisan:
> As discussed on the deep dive yesterday, we pass consensus block counts to sync_up_to_consensus_height, that then 
> passes the block count to fetch the block hash, however the rpc expects a block height.

Does it not work out exactly as we are using a non-inclusive range so height in old_count..new_count?

dpc:
Yeah, I just looked briefly at this code and renaming this function and it's two arguments `s/height/count/` makes it look correct and natural?

bradleystachurski:
> Yeah, I just looked briefly at this code and renaming this function and it's two arguments `s/height/count/` makes it look correct and natural?

That makes it look correct and natural, except it then highlights we're passing a `count` instead of `height` to `btc_rpc.get_block_hash`, which is expecting a `height`. For that to be correct, we'd need to subtract 1 from the count then change the range we're looping over from

```rust
for height in old_height..new_height
```

to

```rust
for count in (old_count + 1)..=new_count
```

Since this is just a logical/name confusion without an underlying bug with our syncing logic, I'm in favor of not touching in the old wallet and instead cleaning up in wallet v2.

> Does it not work out exactly as we are using a non-inclusive range so height in old_count..new_count?

Yep :+1:

dpc:
> > Yeah, I just looked briefly at this code and renaming this function and it's two arguments `s/height/count/` makes it look correct and natural?
> 
> That makes it look correct and natural, except it then highlights we're passing a `count` instead of `height` to `btc_rpc.get_block_hash`, which is expecting a `height`. For that to be correct, we'd need to subtract 1 from the count then change the range we're looping over from

What? Noooo?

When you want to go through first 10 elements of the array, you do:

```rust
for i in 0..10 {
   // use `i`
}
```

The `i` is "height" in how both Bitcoin and Fedimint use it. 

In the same way, after the rename:

```
for height: u32 in old_count..new_count {
   self.btc_rpc.get_block_hash(height: u64::from(height));
}
```

is entirely correct. Fedimint does not use different numbering scheme than Bitcoin. Genesis block has index/height `0` in both of them.

So to "sync up to consensus count 5", you need to go through 5 blocks with height/indexes: `0, 1, 2, 3, 4`.

Only when referring to "total number of blocks in the blockchain", Fedimint would use "count is 5", while Bitcoin would say "height is 4".

Sorry if I misunderstand something. :) @bradleystachurski

bradleystachurski:
@dpc and I discussed further out-of-band. Let's land a rename change https://github.com/fedimint/fedimint/pull/6019

elsirion:
Superseded by #6019 afaik

PR #6019, opened 2024-09-16
refactor: rename sync_up_to_block_height and args to count

bradleystachurski:
We pass consensus block counts to `sync_up_to_consensus_height`, which is confusing. This is a simple rename so it's obvious we're using block counts. There are no logical changes.

More context in this PR discussion https://github.com/fedimint/fedimint/pull/6012


dpc (approved):

joschisan (approved):
That’s what I had in mind to. Seems like the most natural way to do it.

PR #6020, opened 2024-09-16
chore(core): `DatabaseTransaction::global_dbtx`

dpc:
This allows going back from prefixed dbtx, to global unprefixed dbtx, but only if one posseses the "access token" created during creation of the prefixed db.

In the future, this could help eliminate bunch of abstractions we introduced to juggle global vs module dbtx, though it might require long term effort and refactoring.

Currently  the goal is to unblock #5995

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -407,19 +407,27 @@ impl Database {
     }
 
     /// Create [`Database`] isolated to a partition with a given `prefix`
-    pub fn with_prefix(&self, prefix: Vec<u8>) -> Self {
-        Self {
-            inner: Arc::new(PrefixDatabase {
-                inner: self.inner.clone(),
-                prefix,
-            }),
-            module_decoders: self.module_decoders.clone(),
-        }
+    pub fn with_prefix(&self, prefix: Vec<u8>) -> (Self, GlobalDBTxAccessToken) {
+        let global_dbtx_access_token = GlobalDBTxAccessToken::from_prefix(&prefix);
+        (
+            Self {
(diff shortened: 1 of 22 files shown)


fedimint-core/src/db/mod.rs
```diff
@@ -1606,6 +1671,50 @@ impl<'tx, Cap> DatabaseTransaction<'tx, Cap> {
...
+    ///
+    /// Since we need to do it at runtime, we want the user modules not to be
+    /// able to call `global_dbtx` too easily. But at the same time we don't
+    /// need to be paranoid.
+    ///
+    /// This must be deterministic during whole instance of the software running
+    /// (because it's being rederived independently in multiple codepahs) , but
+    /// it could be somewhat randomized between different runs and releases.
+    fn from_prefix(prefix: &[u8]) -> Self {
+        Self(prefix.iter().fold(0, |acc, b| acc + u32::from(*b)) + 513)
```
elsirion, line 1716: Poor man's CRC :laughing: (or Java string hashing)

elsirion (approved):
Looks ok to me, wondering if we actually need the "hash" of the prefix or if it could be a ZST that can only be created when a tx is specialized. But no strong opinion either way, most of the complexity is elsewhere anyway.

fedimint-core/src/db/mod.rs
dpc: Creating such a ZST value needs to be available as a public method, which the user could get access to and use.

The whole mechanism is somewhat superficial, and we could possibly do only with `#[doc(hidden)]` really...

PR #6022, opened 2024-09-17
fix: GatewayLdkClient::info() doesn't rely on esplora

### LINKED ISSUE #5965: LDK Gateway spamming esplora and killing ldk-node

tvolk131:
Fixes #5965

Ever since adding the ldk gateway to mprocs, it's been significantly slower to fully initialize. To help speed it up, I made #5997 so that the `ldk-node::Node` instance would manually sync whenever we mined new blocks. This led to a ~8% improvement in `just mprocs` startup speed. Not bad, but less than I was hoping for. And strangely, even with that change, the startup time for `just mprocs` was heavily affected by the on-chain sync intervals:

https://github.com/fedimint/fedimint/blob/06717c0a480f619feadb9b411dd57267fbb03e98/gateway/ln-gateway/src/lightning/ldk.rs#L76-L77

It turns out the slowness was because the implementation of `ILnRpcClient::info()` for the LDK gateway set `synced_to_chain` to true only if the most recent sync time for the `ldk-node::Node` instance was more recent than the most recent esplora sync time. This led to `synced_to_chain` often staying false until _another_ sync was done, even if the node actually was fully synced. This effectively meant we had to wait for the background sync, which is why we couldn't remove the sync interval overrides above.

The solution is to simply use the block height provided by `ldk-node::Node::status()`. And with this change we can safely delete the sync overrides, preventing spam to the esplora server.

Before:
51.7s, 47.2s, 50.5s, 50.1s, 49.7s, 55.1s, 41.7s, 41.9s, 52.2s, 50.5s
Average: 49.06s

After:
35.5s, 35.3s, 35.7s, 34.6s, 34.4s, 37.9s, 36.6s, 33.9s, 36.1s, 35.4s
Average: 35.54s

Average speed improvement of ~28%

### DIFF
--- a/gateway/ln-gateway/src/lightning/ldk.rs
+++ b/gateway/ln-gateway/src/lightning/ldk.rs
@@ -453,22 +453,6 @@ impl ILnRpcClient for GatewayLdkClient {
         channel_size_sats: u64,
         push_amount_sats: u64,
     ) -> Result<OpenChannelResponse, LightningRpcError> {
-        let funding_txid_or = self
-            .node
-            .list_channels()
-            .iter()
-            .find(|channel| {
-                channel.counterparty_node_id == bitcoin30_to_bitcoin32_secp256k1_pubkey(&pubkey)
-            })
-            .and_then(|channel| channel.funding_txo)
-            .map(|funding_txo| funding_txo.txid);
-
-        if let Some(funding_txid) = funding_txid_or {
-            return Ok(OpenChannelResponse {
-                funding_txid: funding_txid.to_string(),
-            });
-        }
-
         let push_amount_msats_or = if push_amount_sats == 0 {
             None
         } else {
--- a/gateway/ln-gateway/src/lightning/ldk.rs
+++ b/gateway/ln-gateway/src/lightning/ldk.rs
@@ -99,16 +99,7 @@ impl GatewayLdkClient {
         });
         node_builder
             .set_entropy_bip39_mnemonic(mnemonic, None)
-            .set_chain_source_esplora(
-                esplora_server_url.to_string(),
-                Some(EsploraSyncConfig {
-                    // TODO: Remove these and rely on the default values.
-                    // See here for details: https://github.com/lightningdevkit/ldk-node/issues/339#issuecomment-2344230472
-                    onchain_wallet_sync_interval_secs: 10,
-                    lightning_wallet_sync_interval_secs: 10,
-                    ..Default::default()
-                }),
-            );
+            .set_chain_source_esplora(esplora_server_url.to_string(), None);
         let Some(data_dir_str) = data_dir.to_str() else {
             return Err(anyhow::anyhow!("Invalid data dir path"));
         };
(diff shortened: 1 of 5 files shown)


dpc (dismissed):

gateway/ln-gateway/src/lightning/lnd.rs
```diff
@@ -1134,7 +1134,7 @@ impl ILnRpcClient for GatewayLndClient {
         let mut client = self.connect().await?;
 
         // Connect to the peer first
-        client
+        let _ = client
```
m1sterc001guy, line 1137: Why is this being swallowed?

tvolk131: I'm still trying to figure out why CI is failing, this was one of the things I tweaked that I thought might fix it. The intention here was to make this function idempotent, since we retry channel opens in devimint:

https://github.com/fedimint/fedimint/blob/004c35d4b1caaa33c17f7640d5f17f93c3547154/devimint/src/external.rs#L915-L919

I believe calling `ConnectPeer` on a peer that's already connected returns an error.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -225,45 +215,15 @@ impl Drop for GatewayLdkClient {
...
             pub_key: self.node.node_id().serialize().to_vec(),
             alias: match self.node.node_alias() {
                 Some(alias) => alias.to_string(),
                 None => format!("LDK Fedimint Gateway Node {}", self.node.node_id()),
             },
             network: self.node.config().network.to_string(),
-            block_height,
-            synced_to_chain,
+            block_height: node_status.current_best_block.height,
+            synced_to_chain: node_status.is_running && node_status.is_listening,
```
m1sterc001guy, line 226: Does this mean it is synced to the chain? I would have thought you should check LDK's latest block vs the esplora's latest block?

m1sterc001guy:
Closing in favor of https://github.com/fedimint/fedimint/pull/6037

Issue #6014, opened 2024-09-13
Updating Lightning Gateway fees does not take effect in Guardian UI


f321x:
When updating the Gateway fees (by changing the FM_GATEWAY_FEES env var) from the default 1% to e.g. 0.1% it still shows the original 1% in the Guardian UI. 
As the UI only fetches the values from the gateway this could be a gateway related bug, maybe not taking the new fees in account or reporting a wrong fee rate to the UI?

elsirion:
Call: it's just a default fee rate, doesn't change anything for already-joined federations

elsirion:
To expand on my comment yesterday: the env var `FM_GATEWAY_FEES` is confusingly named and should be called `FM_DEFAULT_GATEWAY_FEES` or not exist at all. To change fees after the fact you can use the `gateway-cli set-configuration --per-federation-routing-fees "<federation_id>,<base>,<ppm>"` command.

```
$ gateway-cli --rpcpassword aebaege2Mi set-configuration --help
Usage: gateway-cli set-configuration [OPTIONS]

Options:
      --password <PASSWORD>
          
      --num-route-hints <NUM_ROUTE_HINTS>
          
      --routing-fees <ROUTING_FEES>
          Default routing fee for all new federations. Setting it won't affect existing federations
      --network <NETWORK>
          
      --per-federation-routing-fees <PER_FEDERATION_ROUTING_FEES>
          Format federation id,base msat,proportional to millionths part. Any other federations not given here will keep their current fees
  -h, --help
          Print help
```

f321x:
Thanks for the explanation, this resolves the issue.

PR #6023, opened 2024-09-17
fix(gateway): rename default fee env var

elsirion:
Related to #6014.

Its name now reflects that it doesn't affect already connected federations.


dpc (approved):
Do we need any sort of of backward compat here? (read old variable if new one is not set?)

m1sterc001guy (approved):

PR #6024, opened 2024-09-17
chore: add ldk logs to mprocs

m1sterc001guy:
LDK logs are emitted to a different directory than `ldk_gw`. It is useful for debugging to also have this pane in mprocs


tvolk131 (approved):
Awesome, thanks! Tested on my machine, works great.

PR #6032, opened 2024-09-18
feat(audit): allow adding of AuditItem to Audit

toyota-corolla0:
New Audit::add_item to allow adding of AuditItem directly. 

I have a case in my module where Audit::add_items does not work because I need to read multiple DB keys to find my millisat i64. Thanks!

### DIFF
--- a/fedimint-core/src/module/audit.rs
+++ b/fedimint-core/src/module/audit.rs
@@ -26,6 +26,10 @@ impl Audit {
         }
     }
 
+    pub fn add_item(&mut self, item: AuditItem) {
+        self.items.push(item);
+    }
+
     pub async fn add_items<KP, F>(
         &mut self,
         dbtx: &mut DatabaseTransaction<'_>,


elsirion (approved):
I *think* this is ok to do, the whole audit thing is not as useful as originally imagined anyway and depends on correct implementation of the module (see https://github.com/fedimint/fedimint/issues/4283).

The tight coupling to the DB was a mistake in retrospect imo and I probably just tried to be too clever about it.

@dpc @bradleystachurski thoughts? I'd just merge and revamp the whole thing eventually.

bradleystachurski (approved):
I believe this is okay :+1: 

@toyota-corolla0 when it's ready, could you please post a comment with a link to your use case?

toyota-corolla0:
@bradleystachurski @elsirion I realize now that I don't have a use case. I missed that `DatabaseTransaction::find_by_prefix` mutably borrows self so I cannot actually access other db entries during the stream. Sorry for the unneeded pull request.

toyota-corolla0:
There is a way to accomplish what I need which for some reason I didn't think of before. I just have to put all of the value I am going to need in a map before calling `Audit::add_items`:

```rust
let market_open_contracts_map: HashMap<OutPoint, ContractAmount> = dbtx
            .find_by_prefix(&db::MarketDynamicPrefixAll)
            .await
            .map(|(db::MarketDynamicKey(outpoint), market_dynamic)| {
                (outpoint, market_dynamic.open_contracts)
            })
            .collect()
            .await;
        audit
            .add_items(
                dbtx,
                module_instance_id,
                &db::MarketStaticPrefixAll,
                |db::MarketStaticKey(outpoint), market_static| {
                    let market_open_contracts = market_open_contracts_map.get(&outpoint).unwrap();
                    let milli_sat =
                        -((market_static.contract_price * market_open_contracts.0).msats as i64);

                    milli_sat
                },
            )
            .await;
```

PR #6034, opened 2024-09-18
fix: stop listening for payments when shutting down

m1sterc001guy:
Edge case of gateway's safe stop feature that was pointed out by @elsirion here: https://github.com/fedimint/fedimint/pull/5999/files#r1762742035

There's a race condition when stopping where if the gateway receives a payment request before `cancel_on_shutdown` has stopped the thread, the gateway would start processing the payment and potentially leave the gateway in an un-finished state. This PR just adds a `tokio::select!` to payment processing thread so that it knows to break when the gateway's state is `ShuttingDown`.

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -538,17 +538,27 @@ impl Gateway {
         if handle
             .cancel_on_shutdown(async move {
                 loop {
-                    let state = self.get_state().await;
-                    let GatewayState::Running { lightning_context } = state else {
+                    let payment_request = tokio::select! {
+                        payment_request = stream.next() => {
+                            payment_request
+                        }
+                        () = self.is_shutting_down_safely() => {
+                            break;
+                        }
+                    };
+
+                    // Hold the Gateway state's lock so that it doesn't change before `handle_htlc`.
+                    let state_guard = self.state.read().await;
+                    let GatewayState::Running { ref lightning_context } = *state_guard else {
                         warn!(
-                            ?state,
+                            ?state_guard,
                             "Gateway isn't in a running state, cannot handle incoming payments."
                         );
                         break;
                     };
 
-                    let htlc_request = match stream.next().await {
-                        Some(Ok(htlc_request)) => htlc_request,
+                    let payment_request = match payment_request {
+                        Some(Ok(payment_request)) => payment_request,
                         other => {
                             warn!(
                                 ?other,
@@ -558,7 +568,7 @@ impl Gateway {
                         }
                     };
 
-                    self.handle_htlc(htlc_request, &lightning_context).await;
+                    self.handle_htlc(payment_request, &lightning_context).await;
                 }
             })
             .await
@@ -569,6 +579,18 @@ impl Gateway {
         } else {
             info!("Received shutdown signal");
             ReceivePaymentStreamAction::NoRetry
+        }
+    }
+
+    /// Polls the Gateway's state waiting for it to shutdown so the thread
+    /// processing payment requests can exit.
+    async fn is_shutting_down_safely(&self) {
+        loop {
+            if let GatewayState::ShuttingDown { .. } = self.get_state().await {
+                return;
+            }
+
+            fedimint_core::task::sleep(Duration::from_secs(1)).await;
         }
     }
 
@@ -1538,9 +1560,11 @@ impl Gateway {
     }
 
     pub async fn handle_shutdown_msg(&self, task_group: TaskGroup) -> AdminResult<()> {
-        if let GatewayState::Running { lightning_context } = self.get_state().await {
-            self.set_gateway_state(GatewayState::ShuttingDown { lightning_context })
-                .await;
+        // Take the write lock on the state so that no additional payments are processed
+        let mut state_guard = self.state.write().await;
+        if let GatewayState::Running { lightning_context } = state_guard.clone() {
+            *state_guard = GatewayState::ShuttingDown { lightning_context };
+
             self.federation_manager
                 .read()
                 .await


gateway/ln-gateway/src/lib.rs
```diff
@@ -551,8 +551,17 @@ impl Gateway {
...
-                    let htlc_request = match stream.next().await {
-                        Some(Ok(htlc_request)) => htlc_request,
+                    let payment_request = tokio::select! {
+                        payment_request = stream.next() => {
+                            payment_request
+                        }
+                        () = self.is_shutting_down_safely() => {
+                            break;
+                        }
+                    };
```
elsirion, line 561: Not sure if we fully fix raciness, but at least the time where it might happen becomes much shorter. If we start shutdown after the select, but before calling `handle_htlc`, could that still be a problem?

elsirion (dismissed):
Not sure fully fixing it is feasible rn, but this is a step in the right direction (and wayyyy better than the behavior before safe shutdown).

elsirion:
Would like to get a second review though, maybe from @tvolk131 :)

gateway/ln-gateway/src/lib.rs
tvolk131: Currently according to `ILnRpcClient` documentation, not calling `ILnRpcClient::complete_htlc()` for an item sent to the `ILnRpcClient::route_htlcs()` stream is undefined behavior. The underlying behavior for ldk, lnd, and cln is to fail the HTLC after the CLTV expiry time has passed. Definitely not ideal.

@m1sterc001guy can we add a TODO comment in here stating the issue above? This should definitely be addressed down the road.

Also, would it be worth defining the above behavior in `ILnRpcClient`? If so, I can make a PR for that.

m1sterc001guy: > If we start shutdown after the select, but before calling handle_htlc, could that still be a problem?

Ah yes, this is still an issue. I think this is fixable by having the section of code after the `select!` protected by a lock, and also the `handle_shutdown_msg` function protected by the same lock. That way, the gateway's state cannot change after we've checked it and call `handle_htlc`

gateway/ln-gateway/src/lib.rs
```diff
@@ -551,8 +551,17 @@ impl Gateway {
                         break;
                     };
 
-                    let htlc_request = match stream.next().await {
-                        Some(Ok(htlc_request)) => htlc_request,
+                    let payment_request = tokio::select! {
+                        payment_request = stream.next() => {
+                            payment_request
+                        }
+                        () = self.is_shutting_down_safely() => {
```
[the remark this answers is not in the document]
m1sterc001guy: No. This case doesn't handle the case where the user presses Ctrl-C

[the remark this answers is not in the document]
m1sterc001guy: As discussed offline, I moved this to after the `select!` and after we take the lock, so that we can be sure that `GatewayState` doesn't change before `handle_htlc` is called.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: @elsirion I believe I fixed the remaining issue by holding the `GatewayState` RW Lock. The `handle_shutdown_msg` thread will hold the write lock and this thread will hold the read lock. Here, we now check the state after taking the lock, meaning if we are still in the `Running` state that the state should not change until `handle_htlc` is called (should be atomic now).

Thanks for double checking this!

gateway/ln-gateway/src/lib.rs
```diff
@@ -1493,9 +1516,11 @@ impl Gateway {
     }
 
     pub async fn handle_shutdown_msg(&self, task_group: TaskGroup) -> Result<()> {
-        if let GatewayState::Running { lightning_context } = self.get_state().await {
-            self.set_gateway_state(GatewayState::ShuttingDown { lightning_context })
-                .await;
+        // Take the write lock on the state so that no additional payments are processed
+        let mut state_guard = self.state.write().await;
+        if let GatewayState::Running { lightning_context } = state_guard.clone() {
```
tvolk131, line 1521: Can we do something similar here as I mentioned above?

gateway/ln-gateway/src/lib.rs
```diff
@@ -542,7 +542,18 @@ impl Gateway {
...
+                            payment_request
+                        }
+                        () = self.is_shutting_down_safely() => {
+                            break;
+                        }
+                    };
+
+                    // Hold the Gateway state's lock so that it doesn't change before `handle_htlc`.
+                    let state_guard = self.state.read().await;
+                    let state = state_guard.clone();
```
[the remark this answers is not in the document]
m1sterc001guy: That does not compile, you cannot deference the `state_guard` because that would move it, and it doesn't implement `Copy`. This is why I am cloning it, so that we can actually read the state.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: See above comment

gateway/ln-gateway/src/lib.rs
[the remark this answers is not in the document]
tvolk131: Did you try it? It runs on my machine. Using `ref lightning_context` and `*state_guard` performs a dereference rather than a clone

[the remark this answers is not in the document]
m1sterc001guy: ah I missed the `ref` in this. Yes this works, I will fix.

gateway/ln-gateway/src/lib.rs
m1sterc001guy: As discussed on the call, we need to clone here since we need the `LightningContext` in the `ShuttingDown` state.

tvolk131 (approved):

tvolk131:
Merge conflict 😭

m1sterc001guy:
> Merge conflict

Fixed!

tvolk131 (approved):

gateway/ln-gateway/src/lib.rs
elsirion: Looks good to me now!

PR #6037, opened 2024-09-18
fix: GatewayLdkClient::info() doesn't rely on timestamps

### LINKED ISSUE #5965: LDK Gateway spamming esplora and killing ldk-node

tvolk131:
Fixes #5965

### DIFF
--- a/devimint/src/gatewayd.rs
+++ b/devimint/src/gatewayd.rs
@@ -234,9 +234,14 @@ impl Gatewayd {
         let federation_id = fed.calculate_federation_id();
         let invite_code = fed.invite_code()?;
         info!("Recovering {federation_id}...");
-        cmd!(self, "connect-fed", invite_code.clone(), "--recover=true")
-            .run()
-            .await?;
+        poll("gateway connect-fed", || async {
+            cmd!(self, "connect-fed", invite_code.clone(), "--recover=true")
+                .run()
+                .await
+                .map_err(ControlFlow::Continue)?;
+            Ok(())
+        })
+        .await?;
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


m1sterc001guy:
Reverted with https://github.com/fedimint/fedimint/pull/6217, going to investigate flakiness

Issue #6030, opened 2024-09-17
Slow guardian recovery / synchronization


fparisotto:
For a federation that is running for months, with approx. 100k  sessions. The recovery process of a node from scratch (empty database) is taking 16k session/day. The CPU and network are mostly idle during the process.
It would be good to have this process optimized to avoid a longer downtime.

elsirion:
Idea from a call with DPC: to not mess with consensus too much, which doesn't lend itself to fetch parallelization right now, we could have a pre-fetching process. The API client has a cache already anyway, so we'd be fetching future sessions (up to some point) and pre-fill the cache with that and the consensus could just request sessions sequentially as it does now.

dpc:
Since it is using a custom api logic, it would require custom cache, or switching to use common api call with cache just like in other calls.

But for starter, I think the whole thing is bottlenecked on an artificial 5 second sleep for every session, which I believe is probably OK to remove, which should lead to 20x - 50x speedup right there.

joschisan:
@dpc I think I now the line you mean - feel free to remove it and ping me for review.

dpc:
I'm adding a backport label to the PR, but we should be careful rolling it out, as it's not impossible it might have some weird interactions w.r.t long requests and stuff like that will not show up in tests. Running it first for a day on some low profile federation would be a good idea.

dpc:
We can re-open if it ever looks too slow again.

PR #6043, opened 2024-09-18
chore(consensus): ask for signed session outcomes without a delay

dpc:
I think this just leads to a huge slowdown of #6030 and in a normal operation the cost of this API call is neglible.

### DIFF
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -797,11 +797,6 @@ impl ConsensusEngine {
         };
 
         loop {
-            // We only want to initiate the request if we have not ordered a unit in a
-            // while. This indicates that we have fallen behind and our peers
-            // have already switched sessions without us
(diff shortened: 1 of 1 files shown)


dpc:
Must get @joschisan 's review before landing.

joschisan:
@dpc why do you need to pin here?

dpc:
@joschisan Something something about that feature being moved if it doesn't get `&mut`, which then requires the whole thing to be pinned. I just do what the compiler tells me. :D

joschisan (approved):

fedimint-backports:
- #6044

elsirion:
Did you try using the [`pin!` macro](https://docs.rs/tokio/latest/tokio/macro.pin.html)? @dpc 

93zb4w

fedimint-backports:
Git push to origin failed for releases/v0.4 with exitcode 1

PR #6044, opened 2024-09-18
[Backport releases/v0.4] chore(consensus): ask for signed session outcomes without a delay

fedimint-backports:
# Description
Backport of #6043 to `releases/v0.4`.


elsirion:
I think we can live without it till 0.5. If anyone runs into the problem again let's re-open the backport.

Issue #6047, opened 2024-09-19
Obsolete environment variable warnings for FM_BITCOIN_RPC_URL and FM_BITCOIN_RPC_KIND on new NixOS setup


hensg:
```
2024-09-19T17:44:17.136537Z  INFO fedimintd::fedimintd: Starting fedimintd (version: 0.4.2 version_hash: 63a01dd4551157d40f559a5d38bc451db9d362dd)
2024-09-19T17:44:17.136580Z  WARN fm::core: FM_BITCOIN_RPC_KIND is obsolete, use FM_DEFAULT_BITCOIND_RPC_KIND instead
2024-09-19T17:44:17.136588Z  WARN fm::core: FM_BITCOIN_RPC_URL is obsolete, use FM_DEFAULT_BITCOIND_RPC_URL instead
2024-09-19T17:44:17.261241Z  INFO fm::consensus: Starting config gen
2024-09-19T17:44:17.261592Z  INFO fm::net::peer::dkg: Created new config gen Api

```

elsirion:
Fixed in #6048

PR #6048, opened 2024-09-19
chore: add default BTC RPC env vars in NixOS module

hensg:
This change is made to address deprecation warnings in the fedimint service logs. The environment variables `FM_BITCOIN_RPC_URL` and `FM_BITCOIN_RPC_KIND` are now considered obsolete, and having only these variables causes the following warnings:

```
2024-09-19T17:44:17.136580Z  WARN fm::core: FM_BITCOIN_RPC_KIND is obsolete, use FM_DEFAULT_BITCOIND_RPC_KIND instead
2024-09-19T17:44:17.136588Z  WARN fm::core: FM_BITCOIN_RPC_URL is obsolete, use FM_DEFAULT_BITCOIND_RPC_URL instead
```

To avoid these warnings, the corresponding environment variables with the `DEFAULT` prefix (`FM_DEFAULT_BITCOIND_RPC_URL` and `FM_DEFAULT_BITCOIND_RPC_KIND`) are added in the NixOS module, while keeping the deprecated ones for backward compatibility.

Ref: https://github.com/fedimint/fedimint/blob/master/fedimint-core/src/envs.rs#L80-L102

Related issue: https://github.com/fedimint/fedimint/issues/6047

### DIFF
--- a/nix/modules/fedimintd.nix
+++ b/nix/modules/fedimintd.nix
@@ -235,9 +235,9 @@ in
                 (if cfg.bitcoin.rpc.secretFile != null then
                   ''
                     secret=$(${pkgs.coreutils}/bin/head -n 1 "${cfg.bitcoin.rpc.secretFile}")
-                    prefix="''${FM_BITCOIN_RPC_URL%*@*}"  # Everything before the last '@'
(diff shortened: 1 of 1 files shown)


dpc:
:+1:

PR #6049, opened 2024-09-19
refactor: eliminate duplication in rocksdb impl

dpc:

### DIFF
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -24,6 +24,21 @@ use tracing::debug;
 
 use crate::envs::FM_ROCKSDB_WRITE_BUFFER_SIZE_ENV;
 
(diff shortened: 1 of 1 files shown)


fedimint-rocksdb/src/lib.rs
```diff
@@ -24,6 +24,21 @@ use tracing::debug;
 
 use crate::envs::FM_ROCKSDB_WRITE_BUFFER_SIZE_ENV;
 
+// turn an `iter` into a `Stream` where every `next` is ran inside
+// `block_in_place` to offload the blocking calls
```
maan2003, line 28: block_in_place doesn't offload anything

m1sterc001guy (approved):

fedimint-rocksdb/src/lib.rs
dpc: Right. But it offloads other async code to the other executor threads. :D

PR #6051, opened 2024-09-20
refactor: gateway error refactor

m1sterc001guy:
See https://github.com/fedimint/fedimint/pull/5923#issuecomment-2327284866 for additional context.

This PR refactors how we handle errors in the gateway. Previously, we just had a `GatewayError` type where we dumped all of our error types. Most of these would be redacted before returning to the client, which kind of defeats the purpose of using typed errors.

This PR adds two error types: `PublicGatewayError` and `AdminGatewayError`. `AdminGatewayError` error messages and full details are always returned to the client. The intention is that this error type is used for authenticated endpoints only. Unauthenticated endpoints, such as what is used in LNv1 and LNv2, should use `PublicGatewayError`, which has the specific details of the error redacted for privacy reasons.

Both error types always log so the gateway operator can check the logs for full description of an error.

Open to feedback on the design if anyone has strong opinions.

### DIFF
--- a/gateway/ln-gateway/src/error.rs
+++ b/gateway/ln-gateway/src/error.rs
@@ -1,0 +1,144 @@
+use std::fmt::Display;
+
+use axum::response::{IntoResponse, Response};
+use fedimint_core::config::{FederationId, FederationIdPrefix};
+use fedimint_core::fmt_utils::OptStacktrace;
+use reqwest::StatusCode;
+use thiserror::Error;
+use tracing::error;
+
+use crate::lightning::LightningRpcError;
+use crate::state_machine::pay::OutgoingPaymentError;
+
+/// Errors that unauthenticated endpoints can encounter. For privacy reasons,
+/// the error messages are intended to be redacted before returning to the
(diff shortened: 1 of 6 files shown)


elsirion (approved):
Concept ACK

gateway/ln-gateway/src/error.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        Response::builder()
+            .status(status_code)
+            .body(error_message.into())
+            .expect("Failed to create Response")
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum AdminGatewayError {
```
tvolk131, line 66: Nit: Please document

tvolk131 (changes requested):
A few nits. Also how difficult would it be to add a test that attempts to call an authenticated route without being authenticated and verify that the `AdminGatewayError` isn't returned? That would also ensure that the authentication middleware is working correctly.

tvolk131 (approved):

PR #6052, opened 2024-09-20
chore: document lnv2 fee configuration

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-common/src/config.rs
+++ b/modules/fedimint-lnv2-common/src/config.rs
@@ -22,6 +22,7 @@ impl LightningGenParams {
         Self {
             local: LightningGenParamsLocal { bitcoin_rpc },
             consensus: LightningGenParamsConsensus {
+                fee_consensus: FeeConsensus::default(),
                 network: Network::Regtest,
             },
         }
@@ -30,6 +31,7 @@ impl LightningGenParams {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct LightningGenParamsConsensus {
+    pub fee_consensus: FeeConsensus,
     pub network: Network,
 }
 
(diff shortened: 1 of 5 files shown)


modules/fedimint-lnv2-common/src/config.rs
```diff
@@ -65,11 +67,31 @@ pub struct LightningConfigPrivate {
...
+    /// that the federation may charge a additional relative fee per input and
+    /// output of up to one thousand parts per million which is equal to one
+    /// tenth of one percent.
+    ///
+    /// # Panics
+    ///
+    /// The method will panic when invoked with a relative fee in excess of one
+    /// thousand parts per million.
+    pub fn new(parts_per_million: u64) -> Self {
+        assert!(parts_per_million <= 1_000);
```
elsirion, line 87: Let's not make this an invariant/assert, but rather verify that the value provided as config gen params is in certain bounds. We can do that in the keygen functions of the `ServerModuleInit` impl. I'd also set the upper bound at 1% or something that becomes obviously ludicrous. 0.1% seems arbitrarily low.

joschisan: Where exactly would that be verified then?

Keep in mind that the fees add up so on an outgoing payment this will be paid by the gateway and the sender. If you  route a direct payment over the gateway the gateway will pay it twice and the receiver once. So imo 0.1% is a reasonable limit. I do not want to allow a value that may later be too high.

joschisan: I would do both then, keep the assert here and verify it before.

elsirion: Let's make the constructor return a `Result` and avoid redundant asserts so we have a single place to relax the limit later on. If you insist we can go with 0.1% for now, I just want to keep changeability.

elsirion (approved):

PR #6055, opened 2024-09-20
chore(client): retry getting api announcements from a single peer

dpc:
It's highly likely this query will sometimes fail.

We are seeing flakes caused by it in the CI. Eg.:

https://github.com/fedimint/fedimint/actions/runs/10964718284/job/30448946223?pr=6054

### DIFF
--- a/fedimint-client/src/api_announcements.rs
+++ b/fedimint-client/src/api_announcements.rs
@@ -9,7 +9,7 @@ use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::net::api_announcement::{override_api_urls, SignedApiAnnouncement};
 use fedimint_core::runtime::sleep;
 use fedimint_core::secp256k1::SECP256K1;
-use fedimint_core::util::SafeUrl;
+use fedimint_core::util::{backoff_util, retry, SafeUrl};
 use fedimint_core::{impl_db_lookup, impl_db_record, PeerId};
 use fedimint_logging::LOG_CLIENT;
 use futures::future::join_all;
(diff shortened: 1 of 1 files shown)


fedimint-client/src/api_announcements.rs
```diff
@@ -44,11 +44,21 @@ pub async fn run_api_announcement_sync(client_inner: Arc<Client>) {
...
-                let announcements = client_inner
-                    .api
-                    .api_announcements(peer_id)
-                    .await
-                    .with_context(move || format!("Fetching API announcements from peer {peer_id} failed"))?;
+                let announcements =
+
+                retry(
+                    "Fetch api announcement (sync)",
+                    backoff_util::aggressive_backoff(),
```
bradleystachurski, line 51: This will only retry for ~20s. Should use a strategy that retries longer?

fedimint-client/src/api_announcements.rs
```diff
@@ -44,11 +44,21 @@ pub async fn run_api_announcement_sync(client_inner: Arc<Client>) {
...
+                    || async {
+                        client_inner
+                            .api
+                            .api_announcements(peer_id)
+                            .await
+                            .with_context(move || format!("Fetching API announcements from peer {peer_id} failed"))
+                    },
+                )
+                .await
+                .map_err(|_| anyhow::anyhow!("Timed out waiting for chain sync"))?;
```
bradleystachurski, line 61: [nit] Chain sync?

bradleystachurski (approved):

fedimint-client/src/api_announcements.rs
dpc: I think 20s is plenty comparing to one time. :)

fedimint-client/src/api_announcements.rs
dpc: https://github.com/fedimint/fedimint/pull/6056

PR #6069, opened 2024-09-23
chore: benchmark DB access

### LINKED ISSUE #6050: Benchmark `find_by_prefix_sorted_descending`
`BackupWriter::append` uses it, which is called every AlephBFT unit afaik. We noticed a performance regression (multiples of CPU usage) from 0.3 to 0.4 that is likely related to it (`perf` showed lots of time spent in that function).

elsirion:
Fixes #6050, so we get some insight into RocksDB performance. I can't really see an obvious smoking gun here. While reverse iteration is about twice as slow as forward iteration, just getting the first element seems to be blazingly fast (less than 3us). If I combine it with an insert at the end it goes up to 462us, but that might just be the dbtx commit.

```
$ cargo bench -p fedimint-rocksdb

async_group/reverse_prefix_search_only_first
                        time:   [2.6789 µs 2.6957 µs 2.7144 µs]
                        change: [-0.6772% +1.1186% +3.1433%] (p = 0.28 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
Benchmarking async_group/reverse_prefix_search: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.0s, or reduce sample count to 80.
async_group/reverse_prefix_search
                        time:   [59.807 ms 60.265 ms 60.751 ms]
                        change: [-2.3671% -1.2342% -0.1487%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
async_group/prefix_search_only_first
                        time:   [1.5754 µs 1.5857 µs 1.5975 µs]
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
async_group/prefix_search
                        time:   [28.188 ms 28.392 ms 28.631 ms]
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
async_group/increment_last
                        time:   [457.40 µs 462.56 µs 468.48 µs]
                        change: [-4.2827% -2.5961% -0.8962%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe
```

TODO:
* Separate out write and commit in the `increment_last` test, so we only benchmark the reading (using [`iter_custom`](https://docs.rs/criterion/latest/criterion/struct.Bencher.html#method.iter_custom))
* Test with larger DB where most keys aren't queried

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1295,6 +1295,7 @@ dependencies = [
  "ciborium",
  "clap",
  "criterion-plot",
+ "futures",
  "is-terminal",
  "itertools 0.10.5",
  "num-traits",
@@ -1307,6 +1308,7 @@ dependencies = [
  "serde_derive",
  "serde_json",
  "tinytemplate",
+ "tokio",
  "walkdir",
 ]
 
@@ -2979,6 +2981,7 @@ version = "0.5.0-alpha"
 dependencies = [
  "anyhow",
  "async-trait",
+ "criterion",
  "fedimint-core",
  "futures",
  "rocksdb",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,7 @@ bls12_381 = "0.8.0"
 bytes = "1.7.2"
 clap = { version = "4.5.18", features = ["derive", "env"] }
 cln-rpc = "0.1.9"
-criterion = "0.5.1"
+criterion = { version = "0.5.1", features = ["async_tokio"] }
 devimint = { path = "./devimint", version = "=0.5.0-alpha" }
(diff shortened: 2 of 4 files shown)


Kodylow:
@joschisan

PR #6075, opened 2024-09-23
refactor: get rid of ClientDbTxContext

dpc:
It has been rendered redudndant, by the built-in DBTx global_tx.

### DIFF
--- a/fedimint-client/src/module/mod.rs
+++ b/fedimint-client/src/module/mod.rs
@@ -297,10 +297,18 @@ where
     ///
     /// Only intended for internal use (private).
     fn global_db(&self) -> fedimint_core::db::Database {
-        self.client.get().db().clone()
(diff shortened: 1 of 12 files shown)


fedimint-client/src/module/mod.rs
```diff
@@ -615,6 +456,83 @@ where
     {
         operation.outcome_or_updates(&self.global_db(), operation_id, stream_gen)
     }
+
+    pub async fn claim_input<I, S>(
```
elsirion, line 468: Any reason for moving the code? Makes the diff harder to read.

elsirion (approved):
Sweet :heart_eyes:

elsirion:
Feel free to merge, I'd just like @m1sterc001guy to take another look since he's most familiar with the DB architecture.

PR #6076, opened 2024-09-24
feat: cache gateways

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-client/src/api.rs
+++ b/modules/fedimint-lnv2-client/src/api.rs
@@ -21,6 +21,7 @@ use fedimint_lnv2_common::endpoint_constants::{
 use fedimint_lnv2_common::ContractId;
 use itertools::Itertools;
 use lightning_invoice::Bolt11Invoice;
+use rand::seq::SliceRandom;
 use secp256k1::schnorr::Signature;
 
 use crate::{
@@ -128,6 +129,10 @@ where
             .dedup()
             .cloned()
             .collect::<Vec<SafeUrl>>();
+
+        // Shuffling the gateways ensures that payments are distributed over the
+        // gateways evenly.
+        union.shuffle(&mut rand::thread_rng());
 
         union.sort_by_cached_key(|r| {
             gateways
--- a/modules/fedimint-lnv2-client/src/lib.rs
+++ b/modules/fedimint-lnv2-client/src/lib.rs
@@ -7,16 +7,19 @@
 pub mod api;
 #[cfg(feature = "cli")]
 mod cli;
+mod db;
 mod receive_sm;
 mod send_sm;
 
 use std::collections::BTreeMap;
 use std::sync::Arc;
+use std::time::Duration;
 
 use api::{GatewayConnection, RealGatewayConnection};
 use async_stream::stream;
 use bitcoin::hashes::{sha256, Hash};
 use bitcoin::secp256k1;
+use db::GatewayKey;
 use fedimint_api_client::api::DynModuleApi;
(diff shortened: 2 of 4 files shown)


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -433,7 +437,36 @@ fn generate_ephemeral_tweak(static_pk: PublicKey) -> ([u8; 32], PublicKey) {
...
+    /// This method updates the mapping from lightning node public keys to
+    /// gateway api endpoints maintained in the module database once a day. When
+    /// paying an invoice this enables the client to select the gateway that has
+    /// created the invoice, if possible, such that the payment does not go
+    /// over lightning, reducing fees and latency.
+    ///
+    /// Client integrators are expected to call this function in a spawned task.
+    pub async fn update_gateway_map(&self) -> ! {
+        loop {
+            if let Ok(gateways) = self.module_api.gateways().await {
```
elsirion, line 449: Could use let-else syntax to reduce indentation

```rust
let Ok(gateways) = self.module_api.gateways().await else { continue; };
```

on a second thought though: this should likely retry and never stop retrying till it gets an answer. That way the error case would be completely avoided (assuming we'll never hit 2^64 retries).

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -433,7 +437,36 @@ fn generate_ephemeral_tweak(static_pk: PublicKey) -> ([u8; 32], PublicKey) {
...
+                let mut dbtx = self.client_ctx.module_db().begin_transaction().await;
+
+                for gateway in gateways {
+                    if let Ok(Some(routing_info)) = self.routing_info(&gateway).await {
+                        dbtx.insert_entry(&GatewayKey(routing_info.lightning_public_key), &gateway)
+                            .await;
+                    }
+                }
```
elsirion, line 457: This code never invalidates gateways that got removed/are unreachable or am I missing something? That doesn't seem optimal.

I'd probably try to reach all of them first, then if any of them was reachable remove all DB entries and insert new ones. If none was reachable assume we are just went offline right after fetching the gateways, don't change the DB and wait for the next time the update is run.

modules/fedimint-lnv2-client/src/api.rs
```diff
@@ -129,6 +130,10 @@ where
             .cloned()
             .collect::<Vec<SafeUrl>>();
 
+        // Shuffling the gateways ensures that payments are distributed over the
+        // gateways evenly.
+        union.shuffle(&mut rand::thread_rng());
```
elsirion, line 135: This is useful because the `sort` called next is stable and we assume that most of the times guardians agree about the proposed gateways?

elsirion (commented):
Did you take a look at the existing GW syncing code? I don't remember all the pitfalls we ran into, but there might be some learnings embedded there.

modules/fedimint-lnv2-client/src/api.rs
joschisan: Exactly - They will be ordered by number of votes but gateways with the same number of votes are in random order.

modules/fedimint-lnv2-client/src/lib.rs
joschisan: When the gateway is selected to pay an invoice it will only consider the gateways returned by the federation at that moment, so an outdated entry will be ignored.

elsirion (approved):

modules/fedimint-lnv2-client/src/lib.rs
joschisan: I agree, sorry about that. Should have opened as draft.

PR #6084, opened 2024-09-24
chore(test): another mint module recovery test

dpc:

### DIFF
--- a/modules/fedimint-mint-tests/src/bin/mint-client-restore.rs
+++ b/modules/fedimint-mint-tests/src/bin/mint-client-restore.rs
@@ -1,0 +1,93 @@
+use anyhow::Result;
+use devimint::cmd;
+use devimint::federation::Federation;
+use devimint::util::FedimintCli;
+use devimint::version_constants::VERSION_0_5_0_ALPHA;
+use fedimint_logging::LOG_DEVIMINT;
+use rand::Rng;
+use tracing::{debug, info};
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    devimint::run_devfed_test(|fed, _process_mgr| async move {
+        let fed = fed.fed().await?;
+
+        test_restore_gap_test(fed).await?;
+        Ok(())
+    })
+    .await
+}
+
+pub async fn test_restore_gap_test(fed: &Federation) -> Result<()> {
+    let client = fed.new_joined_client("restore-gap-test").await?;
+    let fedimint_cli_version = FedimintCli::version_or_default().await;
+
+    if fedimint_cli_version < *VERSION_0_5_0_ALPHA {
+        return Ok(());
+    }
+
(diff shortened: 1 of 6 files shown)


elsirion (commented):
Is there a way to do this as an in-process/unit test? These CLI tests are very limited in their ability to explore weird scenarios that might only happen once in a blue moon in practice.

dpc:
@elsirion Maybe, but would need a description of a guess of a scenario that we're missing here.

The recovery logic does not have a lot of weird cases in my mind. You start from the empty state and rewind existing history. Any state that the user can hit, we should be able to hit with just the cli.

modules/fedimint-mint-tests/src/bin/mint-client-restore.rs
```diff
@@ -0,0 +1,91 @@
...
+    devimint::run_devfed_test(|fed, _process_mgr| async move {
+        let fed = fed.fed().await?;
+
+        test_restore_gap_test(fed).await?;
+        Ok(())
+    })
+    .await
+}
+
+pub async fn test_restore_gap_test(fed: &Federation) -> Result<()> {
```
elsirion, line 21: It says something about the gap limit, but it's unclear to me which guarantees about preconditions the code provides. The coin selection and reissuance is opaque, so we don't really know which scenario will be tested. Ideally we could test scenarios where transactions failed and e-cash outputs were cancelled, producing holes in the used indexes.

dpc: @elsirion So your worry is that on failed paymenets client allocates the note_idx, but never uses it, potentially creating holes of unused mode_idx larger than gap limit? In that case why don't we just increase the gap limit? I mean - there is no other way anyway.

elsirion: Yeah, that came to my mind too. I've been staring at the mint recovery code all afternoon and haven't seen anything suspicious.

elsirion: @dpc https://github.com/fedimint/fedimint/pull/6096

dpc: @elsirion Sir, I can test that to, see last commit.

dpc: I think we've found something.

dpc: @elsirion Actually, it was just artifact of multiple gaps adding up and crossing the gap limit, it seems. Everything seems to be working.

elsirion (approved):

elsirion:
More tests more good, but I wish this was a unit test. You built a beautifully testable recovery logic, that is very self-contained at its core and only relies on DB access at the very end. These CLI tests are really hard to debug should they ever become flaky.

PR #6092, opened 2024-09-25
chore: remove unused const strings

tvolk131:


fedimint-core/src/envs.rs
```diff
@@ -37,7 +33,7 @@ pub fn is_rbf_withdrawal_enabled() -> bool {
     is_env_var_set("FM_UNSAFE_ENABLE_RBF_WITHDRAWAL")
 }
 
-/// Get value of [`FEDIMINT_BUILD_CODE_VERSION_ENV`] at compile time
+/// Get value of `FEDIMINT_BUILD_CODE_VERSION` at compile time
```
tvolk131, line 36: Note for reviewer: We have to remove the brackets since we removed the rust const it's referencing. Rather than reference that rust const, now we just reference the environment variable.

dpc (approved):

fedimint-build/src/envs.rs
```diff
@@ -3,6 +3,4 @@ pub const FORCE_GIT_HASH_ENV: &str = "FEDIMINT_BUILD_FORCE_GIT_HASH";
 
 /// Env variable the cargo will set during crate build to pass the detected git
 /// hash to the binary itself.
-// Note: Keep in sync with `fedimint_core::envs`, which is not reused-to avoid
```
dpc, line 6: Why remove a comment?

dpc (commented):
Rest LGTM.

elsirion (approved):
The `pub` visibility makes it hard to automatically test against unused vars in CI I guess? Would need a custom script that greps the code base for usages.

fedimint-build/src/envs.rs
tvolk131: Because we removed the matching const in `fedimint-core/src/envs.rs`

PR #6096, opened 2024-09-26
Remove redundant recovery code

elsirion:

### DIFF
--- a/modules/fedimint-mint-client/src/backup/recovery.rs
+++ b/modules/fedimint-mint-client/src/backup/recovery.rs
@@ -432,12 +432,6 @@ impl MintRecoveryState {
                     (out_point, output.amount, issuance_request),
                 );
             } else {
-                // put it back, incorrect amount
-                self.pending_nonces.insert(
-                    output.blind_nonce.0.into(),
-                    (issuance_request, note_idx, pending_amount),
-                );
-
                 warn!(
                     output = ?out_point,
                     blind_nonce = ?output.blind_nonce.0,


modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -432,12 +437,6 @@ impl MintRecoveryState {
                     (out_point, output.amount, issuance_request),
                 );
             } else {
-                // put it back, incorrect amount
-                self.pending_nonces.insert(
-                    output.blind_nonce.0.into(),
-                    (issuance_request, note_idx, pending_amount),
-                );
-
```
elsirion, line 440: Happy to be corrected, but I don't see under which circumstance this would do anything

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -367,6 +367,11 @@ impl MintRecoveryState {
             )
             .is_none());
 
+        assert!(
+            self.pending_nonces.len() >= (self.gap_limit as usize) * self.tbs_pks.count_tiers(),
```
dpc, line 371: We remove nonces we've seen already, so this might fail, no?

dpc: Hmmmm... For every removed one we should add another one ...

dpc: Yeah, but initially we start with an empty one and fill it up one by one, so this is going to immediately fail.

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -65,7 +65,7 @@ impl RecoveryFromHistory for MintRecovery {
             MintRecovery {
                 state: MintRecoveryState::from_backup(
                     snapshot,
-                    30,
+                    200,
```
dpc, line 68: To start recovery we need to upfront derive this many * num_tiers of nonces, which was actually somewhat substantial. I'm planning to increase it to 100.

modules/fedimint-mint-client/src/backup/recovery.rs
elsirion: Ok, then not a viable assert unfortunately.

modules/fedimint-mint-client/src/backup/recovery.rs
elsirion: Already fixed as part of #6084

dpc (approved):

elsirion:
@dpc you are aware that merging this PR created a silent conflict with your PR https://github.com/fedimint/fedimint/pull/6099#discussion_r1780085544 ?

Maybe I should have closed it after seeing yours, didn't think of it in the moment.

Issue #6079, opened 2024-09-24
Gateway should query chain sync status of Lightning node when booting up


m1sterc001guy:
If its not synced, it should give a proper error message and quit.

m1sterc001guy:
@Kodylow would it actually be easier if the gateway waited for the chain to sync?

Kodylow:
yes

PR #6097, opened 2024-09-26
feat: gateway waits for chain sync

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/6079

Improvements in this PR:
1. LDK uses `TaskGroup` for background tasks instead of `tokio::spawn`
2. LDK syncs the wallet in the background more aggressively in devimint (but not in production)
3. Removes the `sync_to_chain` RPC
4. `wait_for_chain_sync` in devimint => `wait_for_block_height`
5. `gatewayd` will wait for the Lightning to sync to chain before transitioning to the `Running` state


gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -248,6 +250,21 @@ impl dyn ILnRpcClient {
...
+        loop {
+            let info = self.info().await?;
+            let synced = info.synced_to_chain;
+            let block_height = info.block_height;
+            if synced {
+                return Ok(());
+            }
+
+            warn!(?block_height, "Lightning node is not synced");
+            fedimint_core::task::sleep(Duration::from_secs(30)).await;
```
tvolk131, line 265: We'll be invoking this code from devimint, right? Won't this significantly slow down dev fed startup?

dpc: *heart attack*

elsirion: @m1sterc001guy thoughts?

m1sterc001guy: Haha appreciate the patience :)

This is not currently called from devimint, it's only called from the gateway. Devimint has its own loop that is tighter than this one, so this does not cause devimint to be slower.

All that said, I think this should refactor devimint. With this change devimint shouldn't need to explicitly wait for chain sync since the gateway will just do it itself. I will make this change.

devimint/src/gatewayd.rs
```diff
@@ -380,26 +386,37 @@ impl Gatewayd {
...
         Ok(())
     }
+
+    pub async fn wait_for_block_height(&self, target_block_height: u64) -> Result<()> {
+        poll("waiting for block height", || async {
+            let info = self.get_info().await.map_err(ControlFlow::Continue)?;
+            let gateway_info: GatewayInfo =
+                serde_json::from_value(info).expect("Failed to decode GatewayInfo");
+            if let Some(height) = gateway_info.block_height {
+                if height >= target_block_height as u32 && gateway_info.synced_to_chain {
```
elsirion, line 411: Is `synced_to_chain` strictly necessary? Is the `height` reported the synced height or the height that the node is aware of but not necessarily synced to yet? If it's the former then just syncing to the height should be sufficient.

In cases where parallel processes keep mining blocks the "synced to chain" state might not be reached for quite a while.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -113,32 +109,63 @@ impl GatewayLdkClient {
...
+        // LDK Node will sync the wallet's chain automatically in the background, but in
+        // the background it only sync every 10 seconds. In devimint we want
+        // this to sync faster, so we spawn a thread that manually syncs the chain.
+        if is_env_var_set(FM_IN_DEVIMINT_ENV) {
+            let node_clone = node.clone();
+            task_group.spawn("ldk node sync event handler", |handle| async move {
+                loop {
+                    if handle.is_shutting_down() {
+                        return;
+                    }
```
elsirion, line 123: You can just spawn the task as cancellable imo, then the `is_shutting_down` check becomes redundant.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -113,32 +109,63 @@ impl GatewayLdkClient {
...
+        handle: &TaskHandle,
     ) {
+        let event = tokio::select! {
+            event = node.next_event_async() => {
+                event
+            }
+            () = handle.make_shutdown_rx() => {
+                return;
+            }
+        };
```
elsirion, line 161: Nit: could you leave a comment please why the manual select over a shutdown receiver is used here over cancelling the entire future in case of shutdown? 

My read is that we *don't* want to shut down after receiving an even and before having sent it over the HTLC stream sender? Likely related to safe shutdowns? 

But I could imagine being like "that's silly, let's move the shutdown a level up" when looking at the code again (indeed that was my first reaction this time).

elsirion (approved):
Looks good to me, but someone with more gateway context should take another look

devimint/src/gatewayd.rs
m1sterc001guy: Agreed, I don't think it is necessary

gateway/ln-gateway/src/lightning/ldk.rs
m1sterc001guy: Yes, your read is correct, if we receive a payment we ideally want to process it before shutting down. This is the pattern we use in LND as well. I will add a comment

gateway/ln-gateway/src/lib.rs
```diff
@@ -532,6 +541,19 @@ impl Gateway {
             return ReceivePaymentStreamAction::ImmediatelyRetry;
         }
 
+        /*
+        TODO: Uncomment this
+        if synced_to_chain {
```
m1sterc001guy, line 546: This slows down devimint. I think I will make it optional via an environment variable.

devimint/src/gatewayd.rs
```diff
@@ -235,9 +235,14 @@ impl Gatewayd {
         let federation_id = fed.calculate_federation_id();
         let invite_code = fed.invite_code()?;
         info!("Recovering {federation_id}...");
-        cmd!(self, "connect-fed", invite_code.clone(), "--recover=true")
-            .run()
-            .await?;
+        poll("gateway recover-fed", || async {
```
tvolk131, line 238: Nit: `connect-fed --recover` since `gateway recover-fed` isn't a real command

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -232,6 +232,36 @@ impl dyn ILnRpcClient {
             })?;
         Ok((pub_key, alias, network, block_height, synced_to_chain))
     }
+
+    /// Waits for the Lightning node to be synced to the Bitcoin blockchain.
```
tvolk131, line 236: Nit: "Waits for the lightning node to be synced to the blockchain."

gateway/ln-gateway/src/lib.rs
tvolk131: I'm guessing this slows down devimint because `wait_for_chain_sync()` uses `backoff_util::background_backoff()`. Would it maybe make more sense to have an environment variable that makes the background backoff quicker rather than dealing with the consequences of it ad-hoc?

tvolk131 (changes requested):

devimint/src/gatewayd.rs
m1sterc001guy: its not meant to be a command, its just a string that indicates what went wrong

gateway/ln-gateway/src/lib.rs
m1sterc001guy: I don't think so. In production, it will take awhile to sync so its likely not a big deal. I think we want an environment variable to skip it anyway, there might be some scenarios where it is not desirable to wait for the chain to sync

m1sterc001guy: The maximum is 60s which is plenty fast for production

tvolk131 (approved):

m1sterc001guy:
Will debug further once https://github.com/fedimint/fedimint/pull/6203 is merged

tvolk131:
> Will debug further once https://github.com/fedimint/fedimint/pull/6203 is merged

It's merged! 😁

m1sterc001guy:
Appears that the flakes were due to https://github.com/fedimint/fedimint/pull/6203

tvolk131 (approved):

PR #6098, opened 2024-09-26
chore(deps): bump tonic_lnd

tvolk131:


m1sterc001guy:
Just FYI, `tonic` now has a vulnerability that is causing our `audit` to fail. The vulnerable version is being pulled in by `fedimint-tonic-lnd`. So this would be great to get in soon :)

https://github.com/fedimint/fedimint/issues/6117

PR #6099, opened 2024-09-27
chore(client): during recovery detect burned ecash due to nonce reuse

dpc:

### DIFF
--- a/modules/fedimint-mint-client/src/backup/recovery.rs
+++ b/modules/fedimint-mint-client/src/backup/recovery.rs
@@ -12,7 +12,7 @@ use fedimint_core::{
     apply, async_trait_maybe_send, Amount, NumPeersExt, OutPoint, PeerId, Tiered, TieredMulti,
 };
 use fedimint_derive_secret::DerivableSecret;
-use fedimint_logging::{LOG_CLIENT_MODULE_MINT, LOG_CLIENT_RECOVERY_MINT};
+use fedimint_logging::{LOG_CLIENT_MODULE_MINT, LOG_CLIENT_RECOVERY, LOG_CLIENT_RECOVERY_MINT};
 use fedimint_mint_common::{MintInput, MintOutput, Nonce};
 use serde::{Deserialize, Serialize};
 use tbs::{AggregatePublicKey, BlindedMessage, PublicKeyShare};
@@ -29,7 +29,7 @@ use crate::{MintClientInit, MintClientModule, MintClientStateMachines, NoteIndex
 
 #[derive(Clone, Debug)]
 pub struct MintRecovery {
-    state: MintRecoveryState,
+    state: MintRecoveryStateV0,
     secret: DerivableSecret,
     client_ctx: ClientContext<MintClientModule>,
 }
@@ -63,7 +63,7 @@ impl RecoveryFromHistory for MintRecovery {
 
         Ok((
             MintRecovery {
-                state: MintRecoveryState::from_backup(
+                state: MintRecoveryStateV0::from_backup(
                     snapshot,
                     100,
                     config.tbs_pks.clone(),
@@ -85,6 +85,14 @@ impl RecoveryFromHistory for MintRecovery {
         Ok(dbtx
             .get_value(&RecoveryStateKey)
             .await
+            .and_then(|(state, common)| {
+                if let MintRecoveryState::V0(state) = state {
+                    Some((state, common))
+                } else {
+                    warn!(target: LOG_CLIENT_RECOVERY, "Found unknown version recovery state. Ignoring");
+                    None
+                }
+            })
             .map(|(state, common)| {
                 (
                     MintRecovery {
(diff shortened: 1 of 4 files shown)


modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -411,21 +426,19 @@ impl MintRecoveryState {
...
             // the moment we see our blind nonce in the epoch history, correctly or
             // incorrectly used, we know that we must have used
             // already
             self.observe_nonce_idx_being_used(pending_amount, note_idx, secret);
 
             if pending_amount == output.amount {
-                assert!(self
-                    .pending_nonces
-                    .remove(&output.blind_nonce.0.into())
-                    .is_some());
```
elsirion, line 428: Moved into the `if` head. Valid transformation because we re-add in the else branch (which was a no-op previously).

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -247,6 +258,16 @@ impl From<CompressedBlindedMessage> for BlindedMessage {
...
 
+#[derive(Debug, Clone, Decodable, Encodable)]
+pub enum MintRecoveryState {
+    V0(MintRecoveryStateV0),
+    #[encodable_default]
+    Default {
+        variant: u64,
+        bytes: Vec<u8>,
+    },
+}
```
elsirion, line 269: Pretty sure this requires a DB migration

elsirion (dismissed):
Changing the recovery state struct without a DB migration would make clients that get upgraded during recovery crash afaik. Maybe there is a way to just bubble up an error in that case, but I wouldn't want a panic.

elsirion:
Maybe we can separate out the first commit? That one looks good. The latter requires some more discussion (as well as possibly applying it to other modules too).

dpc:
> Changing the recovery state struct without a DB migration would make clients that get upgraded during recovery crash afaik. 

Yeah, I know, but that's very unlikely and if they are recovering it's not the end of the world if they need to recover again. Thanks to turning the state into an enum, it can be easily done in the future. Other modules are already using it method, but I realized it after this one was already rolled out.

dpc:
I added a total burned amount field now.

modules/fedimint-mint-client/src/backup/recovery.rs
dpc: Adding a field already does unfortunately.

dpc:
@elsirion I added a "migration".

modules/fedimint-mint-client/src/backup/recovery.rs
elsirion: Not true anymore after merging #6096

modules/fedimint-mint-client/src/backup/recovery.rs
```diff
@@ -266,7 +266,7 @@ impl From<CompressedBlindedMessage> for BlindedMessage {
 
 #[derive(Debug, Clone, Decodable, Encodable)]
 pub enum MintRecoveryState {
-    V0(MintRecoveryStateV0),
+    V1(MintRecoveryStateV0),
```
elsirion, line 269: Inconsistent naming

modules/fedimint-mint-client/src/client_db.rs
```diff
@@ -98,3 +99,14 @@ impl_db_lookup!(
...
     query_prefix = CancelledOOBSpendKeyPrefix,
 );
+
+pub async fn migrate_to_v1(
+    dbtx: &mut DatabaseTransaction<'_>,
+) -> anyhow::Result<Option<(Vec<(Vec<u8>, OperationId)>, Vec<(Vec<u8>, OperationId)>)>> {
+    // between v0 and v1, we changed the format of `MintRecoveryState`, and instead
+    // of migrating it, we can just delete it, so the recovery will just start
+    // again, ignoring any existing state from before the migration
+    dbtx.remove_entry(&RecoveryStateKey).await;
```
elsirion, line 109: Seems safe to me since I can't find any code that derives the fact that we are recovering from its presence.

modules/fedimint-mint-client/src/backup/recovery.rs
dpc: It corresponds to the database version ... oh... I guess that doesn't really make sense.

modules/fedimint-mint-client/src/client_db.rs
dpc: It was all designed to support discarding the state.

modules/fedimint-mint-client/src/backup/recovery.rs
dpc: Yes. Added it back.

elsirion (approved):
LGTM. We don't really expose the lost amount afaik, but once the event log work lands this could be part of the "mint recovery finished" event, so apps can inform the user.

fedimint-backports:

```bash
git worktree add -d .worktree/backport-6099-to-releases/v0.4 origin/releases/v0.4
cd .worktree/backport-6099-to-releases/v0.4
git switch --create backport-6099-to-releases/v0.4
git cherry-pick -x 904ae4dd93c851a6d3402108011a39d57d9ae21d 2f91b3a6c012caddd139399d03cfb1e2c48d3eaa
```

PR #6109, opened 2024-09-30
fix: set lnv2 module consensus version to 1.0

joschisan:


Kodylow (approved):

dpc:
Major consensus versions were intended for upgrades like lnv1 to lnv2 - completely incompat in every way, one module impl replacing another.

Since we are not doing things this way we should only be incrementing minor version. Was lnv2 officially rollout out to require increments anyway?

joschisan:
This is the first module consensus version - since it is something we intend to support long term a 1.0 seemed like the right version to reflect that. Usually 1.0 implies higher stability then 0.x no?

dpc:
Oh, I see. I don't know. To me these are just counters and not attempt to be in any way related to semver. But I guess it's not unreasonable to emulate some meaning of these, where `0` is more of a "experimental".

PR #6119, opened 2024-10-02
chore: make meta source work without a client

maan2003:
so we can get meta before joining the federation


elsirion (approved):

dpc (approved):
Surprisingly tiny.

Issue #6108, opened 2024-09-30
Get rid of DATABASE_VERSION


dpc:
It just occurred to me that one could add a db migration, and then forget to increment `DATABASE_VERSION`. Why have both? A database version could be defined just as number of defined migrations? We could just have `fn database_version` that generically calls `get_database_migrations` and counts them, and checks if its contiguous.

dpc:
@m1sterc001guy  WDYT?

elsirion:
I'd just go with the max key, we already just warn if a key is missing. But agree with the general idea.

m1sterc001guy:
If I recall correctly, the original idea was that we could remove old db migrations for versions that we no longer maintain. In that case, `DATABASE_VERSION` would not be equal to the number of defined migrations. But that's such a small thing, and if we're cool with keeping that code around, then I see no problem with removing it.

dpc:
@m1sterc001guy In that case we could require that migrations are contiguous, but do not need to start at 0.

PR #6120, opened 2024-10-02
chore: remove DATABASE_VERSION

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/6108

Kind of a footgun so let's just remove it.

### DIFF
--- a/fedimint-client/src/db.rs
+++ b/fedimint-client/src/db.rs
@@ -6,9 +6,9 @@ use fedimint_api_client::api::ApiVersionSet;
 use fedimint_core::config::{ClientConfig, ClientConfigV0, FederationId, GlobalClientConfig};
 use fedimint_core::core::{ModuleInstanceId, OperationId};
 use fedimint_core::db::{
-    apply_migrations, create_database_version, CoreMigrationFn, Database, DatabaseTransaction,
-    DatabaseValue, DatabaseVersion, DatabaseVersionKey, IDatabaseTransactionOpsCore,
-    IDatabaseTransactionOpsCoreTyped, MODULE_GLOBAL_PREFIX,
+    apply_migrations, create_database_version, get_current_database_version, CoreMigrationFn,
+    Database, DatabaseTransaction, DatabaseValue, DatabaseVersion, DatabaseVersionKey,
+    IDatabaseTransactionOpsCore, IDatabaseTransactionOpsCoreTyped, MODULE_GLOBAL_PREFIX,
 };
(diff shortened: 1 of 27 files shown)


fedimint-client/src/db.rs
```diff
@@ -480,6 +475,13 @@ pub async fn apply_migrations_client(
         .await
         .is_none();
 
+    let last_key_value = migrations.last_key_value();
+    let target_version = if let Some((last_key, _)) = last_key_value {
+        DatabaseVersion(last_key.0 + 1)
+    } else {
+        DatabaseVersion(0)
+    };
+
```
dpc, line 484: Make a function (possibly generic of V, so can be re-used between server & client). In that function verify that migrations are contiguous.

m1sterc001guy: Will make a function.

I actually don't think we need to assert that migrations are continuous. `apply_migrations` and `apply_migrations_client` already skip over missing migrations, and since we're using the last key here (and not the size of the map), we can handle non-continuous migrations.

dpc: @m1sterc001guy Missing versions in db migration list does not seem useful for anything. It is probably a programming mistake that if not cough automatically, might get unnoticed.

m1sterc001guy: That is true, I also don't think it hurts too much. I can add the contiguous check, just wanted to point out that I don't think its strictly necessary. I'll add a comment.

fedimint-core/src/db/mod.rs
```diff
@@ -2184,6 +2184,31 @@ pub type CoreMigrationFn = for<'r, 'tx> fn(
...
 >;
 
+/// Verifies that all database migrations are defined contiguously and returns
+/// the "current" database version, which is one greater than the last key in
+/// the map.
+pub fn get_current_database_version<F>(
+    migrations: &BTreeMap<DatabaseVersion, F>,
+) -> DatabaseVersion {
+    let mut versions = migrations.keys().cloned().collect::<Vec<_>>();
+    versions.sort();
```
dpc, line 2194: Isn't btreemap already sorted?

dpc (commented):
LGTM, one question.

fedimint-core/src/db/mod.rs
m1sterc001guy: Didn't know if collect would guarantee the order, but quick googles indicates that its already sorted. Will remove `sort`

dpc (approved):

PR #6121, opened 2024-10-02
fix: fix gateway protoc warning

m1sterc001guy:
Fixes a warning that must have been introduced with a recent version bump.


tvolk131 (approved):
Thanks!

PR #6124, opened 2024-10-03
fix: use federation id for database prefix

m1sterc001guy:
We have a bug in the DB implementation of the gateway. Previously, we used the `federation_index` as the prefix for the database. However, this has an issue because the `federation_index` changes when the user leaves and then rejoins the federation.

This PR changes it so that when building the client, we use the federation id instead (which should not change). It turns out that [RocksDB does delta encoding, so common prefixes will not be stored redundantly](https://docs.rockset.com/documentation/docs/storage-architecture#data-compaction-in-rocksdb). So storage size shouldn't be a concern.


tvolk131 (approved):

tvolk131:
LGTM, but let's get one more set of eyes on this before merging

dpc (approved):

PR #6125, opened 2024-10-03
chore: update flakebox

dpc:
Notably nix community is switching to `nixfmt-rfc-style` utility/style for formatting.


Kodylow (approved):
gtg

Issue #6107, opened 2024-09-28
Gatway exits registration loop early if one federation is offline


elsirion:
If one of the federations the GW is trying to register with is offline, it will abort registration and skip the remaining ones. Instead, either each federation should have its own registration loop or failure to register one should not make the others fail too.

PR #6127, opened 2024-10-03
fix: gateway registration exits early

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/6107

### DIFF
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -95,7 +95,7 @@ use rpc::{
 };
 use state_machine::{GatewayClientModule, GatewayExtPayStates};
 use tokio::sync::RwLock;
-use tracing::{debug, error, info, info_span, warn, Instrument};
+use tracing::{debug, error, info, info_span, warn};
 
 use crate::config::LightningModuleMode;
 use crate::db::{get_gatewayd_database_migrations, FederationConfig};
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/state_machine/mod.rs
```diff
@@ -351,7 +351,7 @@ impl GatewayClientModule {
         time_to_live: Duration,
         fees: RoutingFees,
         lightning_context: LightningContext,
-    ) -> anyhow::Result<()> {
```
douglaz, line 354: It's weird that it will fail without notifying the caller. At least you can change this function name to `try_register_with_federation` to let this behavior be clear.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1636,30 +1636,20 @@ impl Gateway {
...
+                        .value()
+                        .get_first_module::<GatewayClientModule>()
+                        .expect("No GatewayClientModule exists")
+                        .register_with_federation(
+                            route_hints.clone(),
+                            GW_ANNOUNCEMENT_TTL,
+                            federation_config.fees,
+                            lightning_context.clone(),
+                        )
+                        .await;
```
elsirion, line 1649: Registration should run in parallel, so that timeouts of earlier federations don't block later federations.

gateway/ln-gateway/src/state_machine/mod.rs
m1sterc001guy: Done

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Fixed. Cancellable task is now spawn for each registration in the loop.

dpc (approved):

joschisan (approved):

PR #5803, opened 2024-08-05
chore: gateway-cli info cleanup

m1sterc001guy:
Cleans up a bunch of stuff about `gateway-cli info` that has been bugging me for awhile.

 - Removes `ClientConfig` from the `info` command. `ClientConfig` can still be retrieved using `gateway-cli config` command.
 - Fixes `gateway-cli config` command
 - Removes route hints and routing fees from `info`. Routing fees are already included in each individual federation. Route hints can be acquired by generating an invoice.
 - Adds the gateway's API URL
 - Adds the gateway's Lightning Mode for determining what type of node it is.

PR #6132, opened 2024-10-05
[Backport] chore: gateway-cli info cleanup

m1sterc001guy:
Backports: https://github.com/fedimint/fedimint/pull/5803


elsirion:
How urgent is that backport? Feels to me like we could just wait for 0.5?

m1sterc001guy:
Not urgent, let's wait for 0.5

PR #6149, opened 2024-10-09
feat: add tweaked invoice rpcs

Kodylow:
Adds the tweaked invoice methods for ln.handle_rpc so we can add lightning address receiving to the fedimint web sdk

### DIFF
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -80,7 +80,9 @@ use pay::PayInvoicePayload;
 use rand::rngs::OsRng;
 use rand::seq::IteratorRandom as _;
 use rand::{CryptoRng, Rng, RngCore};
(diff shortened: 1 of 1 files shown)


maan2003 (approved):

modules/fedimint-ln-client/src/lib.rs
```diff
@@ -476,6 +478,38 @@ impl ClientModule for LightningClientModule {
...
+                    let keypair = KeyPair::from_secret_key(&self.secp, &req.user_key);
+                    let operation_ids = self.scan_receive_for_user_tweaked(keypair, req.indices, req.extra_meta).await;
+                    yield serde_json::to_value(operation_ids)?;
+                }
+                "subscribe_ln_claim" => {
+                    let req: SubscribeLnClaimRequest = serde_json::from_value(payload)?;
+                    for await state in self.subscribe_ln_claim(req.operation_id).await?.into_stream() {
+                        yield serde_json::to_value(state)?;
+                    }
+                }
```
elsirion, line 512: Feels like we should start factoring these into separate functions.

PR #6152, opened 2024-10-11
fix(client): always adding retry delay, even on first connection

dpc:
While working on #6151 I've noticed that the duration of initial calls is suspiciously long.

Turns out we are always adding a delay on connecting, even the first time.


tvolk131 (approved):
I prefer the `_or` naming for `Option` types, but not a big deal! Otherwise LGTM

Issue #6153, opened 2024-10-11
LightningClientModule shouldn't attempt to update gateway cache during init


shaurya947:
https://github.com/fedimint/fedimint/blob/6ec7931f9fbb020406e01cb371be1d4f21a5e40e/modules/fedimint-ln-client/src/lib.rs#L627

The above line of code will cause module initialization to fail if:
- the device is offline AND
- the gateway cache is empty

On top of that, presently when module initialization fails, the whole client initialization also fails. We noticed this occurring on the Fedi side.

We can fix this in 3 phases:
1. Remove the offending lines from `LightningClientModule::new()`
2. Make client initialization more robust so that if a single module fails to initialize the client can still initialize
3. Potentially make module initialization infallible

I suspect (1) is straightforward but (2) and (3) might need some deliberation and consensus building from the community. So I'll do (1) right away for the time being..

PR #6154, opened 2024-10-11
fix: update GW cache outside LN module init

shaurya947:
Phase 1 of the 3-phase fix described in #6153 .. other phases will be their own separate PRs as and when those approaches are aligned on

### DIFF
--- a/fedimint-load-test-tool/src/main.rs
+++ b/fedimint-load-test-tool/src/main.rs
@@ -534,6 +534,10 @@ async fn get_user_client(
         invite_code.clone()
     };
     let (client, invite_code) = build_client(user_invite_code, user_db.as_ref()).await?;
+    // if lightning module is present, update the gateway cache
+    if let Ok(ln_client) = client.get_first_module::<LightningClientModule>() {
+        let _ = ln_client.update_gateway_cache().await;
+    }
     Ok((client, invite_code))
 }
 
--- a/modules/fedimint-ln-client/src/lib.rs
+++ b/modules/fedimint-ln-client/src/lib.rs
@@ -341,7 +341,7 @@ impl ClientModuleInit for LightningClientInit {
     }
 
     async fn init(&self, args: &ClientModuleInitArgs<Self>) -> anyhow::Result<Self::Module> {
-        Ok(LightningClientModule::new(args, self.gateway_conn.clone()).await?)
+        Ok(LightningClientModule::new(args, self.gateway_conn.clone()))
     }
 
     fn get_database_migrations(&self) -> BTreeMap<DatabaseVersion, ClientMigrationFn> {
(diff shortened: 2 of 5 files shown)


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -618,16 +618,8 @@ impl LightningClientModule {
             secp,
             client_ctx: args.context(),
             update_gateway_cache_merge: UpdateMerge::default(),
-            gateway_conn: gateway_conn.clone(),
-        };
-
-        // Only initialize the gateway cache if it is empty
```
dpc, line 624: Not doing it here is a significant change API change. I don't have a full understanding of the consequences.

But generally yes, this needs to be fixed, and doing rpc calls here slows the client initialization, and possibility of failing whole client due to a connectivity problem is terrible.

Alternatively, we should spawn a background task or something to keep doing it. Though the downside is that it racy and non-deterministic, so I'm a bit meh on it.

dpc:
@fedimint/lightning should look and discuss, but probably OK

Kodylow:
maanmeet says its okay

fedimint-load-test-tool/src/main.rs
```diff
@@ -534,6 +534,10 @@ async fn get_user_client(
         invite_code.clone()
     };
     let (client, invite_code) = build_client(user_invite_code, user_db.as_ref()).await?;
+    // if lightning module is present, update the gateway cache
+    if let Ok(ln_client) = client.get_first_module::<LightningClientModule>() {
+        let _ = ln_client.update_gateway_cache().await;
```
dpc, line 539: Is ignoring the error potential source of flakiness?

modules/fedimint-ln-tests/tests/tests.rs
```diff
@@ -48,6 +48,7 @@ async fn pay_invoice(
     gateway_id: Option<secp256k1::PublicKey>,
 ) -> anyhow::Result<OutgoingLightningPayment> {
     let ln_module = client.get_first_module::<LightningClientModule>()?;
+    ln_module.update_gateway_cache().await?;
```
dpc, line 51: And here we don't ignore?

shaurya947:
I ignored in places where lnv2 is also present. Lnv2 doesn’t need this
additional step to update cache right after client init. So I figured when
those tests deprecate lnv1 all this code would be removed anyway.

However I did not ignore the res value in tests that are specific to lnv1

dpc:
I bet lnv1 will stick around for a while. 


I haven't thought much about it, but it seems to me we need to handle possibility of a network failure, so neither ignoring nor failing are a good strategy. I think it would be more robust to have some kind of a retry loop function there and use that everywhere.

`await_gateway_cache_populated().await` or something?

shaurya947:
I don’t think retry logic is necessary since master doesn’t currently have
it (in lightning init) and AFAIK it hasn’t resulted in flakiness. But if we
want full code parity perhaps we should fail if the result is an err
everywhere

dpc:
I'm thinking more about the desired state than whatever is there now. Might as well steer everything in desired direction while at it.

I am guessing that test will want to make gateway cache to get populated at least once somewhere early, while  end user apps spawn a task that will keep populating it periodically(?). So we should probably have these right away.

shaurya947:
I think you are correct: tests will want to update cache once during setup,
whereas client integrators will want to call the function
update_gateway_cache_continously (might be remembering the name wrong but
it’s there in LightningClientModule).

In this PR, tests already take that approach. For fedimint-cli IDK if it
makes sense to call the “continuously” function. What do you think? We can
also iterate in follow up PRs if we want to deliberate more on the
fine-grained approach.

dpc (approved):

dpc:
We'll figure it out. :)

Issue #6155, opened 2024-10-11
Caugh a worrying consensus panic


dpc:
```
h: None, params: 2 }
2024-10-11T18:21:00.381787Z  INFO spawn{task="runway/loading"}: fm::consensus: Recovering from an in-session-shutdown units_len=539
thread 'tokio-runtime-worker' panicked at /home/dpc/lab/fedimint/fedimint/fedimint-server/src/consensus/engine.rs:421:29:
Consensus Failure: rejected item accepted by federation consensus
stack backtrace:
2024-10-11T18:21:00.390753Z DEBUG spawn{task="io-thread-peer-2"}:peer_io_thread{id=2}: fm::net::peer: Trying to reconnect our_id=PeerId(0) peer=PeerId(2) addr=fedimint://127.0.0.1:26871
```

Ooops.

I'm investigating CI flakes, and just ran into it locally. The code is not hacked or anything, very close to vanilla `master`.

dpc:
Ping @joschisan

dpc:
Caught it again, this time with extra panic details

```
thread 'tokio-runtime-worker' panicked at /home/dpc/lab/fedimint/fedimint/fedimint-server/src/consensus/engine.rs:421:29:
Consensus Failure: rejected item accepted by federation consensus: AcceptedItem { item: Module(DynModuleConsensusItem { id: 3, kind: Some(ModuleKind("lnv2")), inner: UnixTimeVote(172
8675433) }), peer: PeerId(0) }, items: 152+102, session_idx: 0
stack backtrace:
```

dpc:
Another one.

```
thread 'tokio-runtime-worker' panicked at /home/dpc/lab/fedimint/fedimint/fedimint-server/src/consensus/engine.rs:421:29:
Consensus Failure: rejected item accepted by federation consensus: AcceptedItem { item: Module(DynModuleConsensusItem { id: 3, kind: Some(ModuleKind("lnv2")), inner: UnixTimeVote(1728675826) }), peer: PeerId(3) }, items: 59+139, session_idx: 1
stack backtrace:
```

dpc:
BTW.

I'm using

```
> c b && ionice -c3 -n7 nix run nixpkgs#cpulimit -- -i -l 65 env RUST_LOG=fm=debug FM_DEVIMINT_CMD_INHERIT_STDERR=1 ./scripts/tests/reconnect-test.sh
```

[...]

PR #6156, opened 2024-10-11
chore: improve panic message

dpc:
Re https://github.com/fedimint/fedimint/issues/6155


fedimint-server/src/consensus/engine.rs
```diff
@@ -650,15 +654,15 @@ impl ConsensusEngine {
         // When we recover from a mid-session crash aleph bft will replay the units that
         // were already processed before the crash. We therefore skip all consensus
         // items until we have seen every previously accepted items again.
-        if let Some(accepted_item) = dbtx
```
dpc, line 653: Can revert. I think 'existing' carries the meaning better here, at least from a pov of outside investigator.

Kodylow (approved):

PR #6157, opened 2024-10-11
chore(ci): just shuffle to test order

dpc:
This should avoid host spots, etc. and distribute the load of tests somewhat more evenly.


Kodylow:
from eric: log the test order
dpc: it's logged in order by testing

Kodylow (approved):

PR #6158, opened 2024-10-11
fix: consensus panic

dpc:
Re #6155 

With this I'm no longer hitting panic in #6155 .


dpc:
Needs to go through @joschisan . There are multiple ways to write this fix, and i don't fully understand all the conditions and assumptions about this function.

joschisan:
This fix is incorrect, we need to use this instead: https://github.com/fedimint/fedimint/pull/6166.

This pr would cause previously processed items to be reprocessed.

PR #6172, opened 2024-10-15
chore: bump ldk-node

tvolk131:


joschisan (approved):

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -68,27 +75,48 @@ impl GatewayLdkClient {
         lightning_port: u16,
         mnemonic: Mnemonic,
     ) -> anyhow::Result<Self> {
-        let mut node_builder = ldk_node::Builder::from_config(ldk_node::Config {
-            network,
+        // In devimint, gateways must allow for other gateways to open channels to them.
+        // To ensure this works, we must set a node alias to signal to ldk-node that we
+        // should accept incoming public channels. However, on mainnet we can disable
+        // this for better privacy.
```
m1sterc001guy, line 81: Can you expand on this comment? Does this imply that if a gateway does not have an alias set, it cannot accept incoming public channels?

m1sterc001guy: I think ideally we should allow setting the alias from an environment variable

PR #6176, opened 2024-10-16
 chore: claim oob refunds with a single state machine 

### LINKED ISSUE #5069: Don't split e-cash when aborting OOB spends

dpc:
Fix #5069

Alternative to #5833

Instead of trying to reclaim every spendable note from oob spend, reclaim them all in one tx.

This requires first a refactoring commit that breaks the 1:1 input to sm relationship. Now multiple inputs can generate a single state machine that will reclaim them all.

One behavior change is that the new tx will fail or succeed together. But if the destination user claimed one note, is it still a reasonable thing to do to reclaim other ones? I don't think so.

### DIFF
--- a/modules/fedimint-mint-client/src/client_db.rs
+++ b/modules/fedimint-mint-client/src/client_db.rs
@@ -1,14 +1,18 @@
+use std::io::Cursor;
+
 use fedimint_client::module::init::recovery::RecoveryFromHistoryCommon;
 use fedimint_core::core::OperationId;
 use fedimint_core::db::{DatabaseTransaction, IDatabaseTransactionOpsCoreTyped as _};
 use fedimint_core::encoding::{Decodable, Encodable};
+use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::{impl_db_lookup, impl_db_record, Amount};
 use fedimint_mint_common::Nonce;
 use serde::Serialize;
 use strum_macros::EnumIter;
 
 use crate::backup::recovery::MintRecoveryState;
-use crate::SpendableNoteUndecoded;
+use crate::oob::{MintOOBStateMachine, MintOOBStateMachineV1, MintOOBStates, MintOOBStatesV1};
+use crate::{MintClientStateMachines, SpendableNoteUndecoded};
 
 #[repr(u8)]
 #[derive(Clone, EnumIter, Debug)]
@@ -109,4 +113,33 @@ pub async fn migrate_to_v1(
     dbtx.remove_entry(&RecoveryStateKey).await;
 
     Ok(None)
+}
+
+/// Maps all `Unreachable` states in the state machine to `OutputDone`
+pub(crate) fn migrate_state_to_v2(
+    operation_id: OperationId,
+    cursor: &mut Cursor<&[u8]>,
+) -> anyhow::Result<Option<(Vec<u8>, OperationId)>> {
+    let decoders = ModuleDecoderRegistry::default();
+
+    let mint_client_state_machine_variant = u16::consensus_decode(cursor, &decoders)?;
+
+    let bytes = match mint_client_state_machine_variant {
+        2 => {
+            let old_state = MintOOBStateMachineV1::consensus_decode(cursor, &decoders)?;
+
+            let new_state = match old_state.state {
+                MintOOBStatesV1::Created(created) => MintOOBStates::Created(created),
+                MintOOBStatesV1::UserRefund(refund) => MintOOBStates::UserRefund(refund),
+                MintOOBStatesV1::TimeoutRefund(refund) => MintOOBStates::TimeoutRefund(refund),
+            };
+            MintClientStateMachines::OOB(MintOOBStateMachine {
+                operation_id: old_state.operation_id,
+                state: new_state,
+            })
+            .consensus_encode_to_vec()
+        }
+        _ => return Ok(None),
+    };
+    Ok(Some((bytes, operation_id)))
 }
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -78,6 +78,7 @@ use fedimint_mint_common::config::MintClientConfig;
 pub use fedimint_mint_common::*;
 use futures::{pin_mut, StreamExt};
 use hex::ToHex;
+use oob::MintOOBStatesCreatedMulti;
 use serde::{Deserialize, Serialize};
 use strum::IntoEnumIterator;
 use tbs::{AggregatePublicKey, Signature};
@@ -92,7 +93,7 @@ use crate::client_db::{
 use crate::input::{
     MintInputCommon, MintInputStateCreated, MintInputStateMachine, MintInputStates,
 };
-use crate::oob::{MintOOBStateMachine, MintOOBStates, MintOOBStatesCreated};
+use crate::oob::{MintOOBStateMachine, MintOOBStates};
 use crate::output::{
     MintOutputCommon, MintOutputStateMachine, MintOutputStates, MintOutputStatesCreated,
     NoteIssuanceRequest,
@@ -1305,18 +1306,13 @@ impl MintClientModule {
             MintClientModule::delete_spendable_note(&self.client_ctx, dbtx, amount, note).await;
         }
 
-        let mut state_machines = Vec::new();
-
-        for (amount, spendable_note) in selected_notes.clone().into_iter_items() {
-            state_machines.push(MintClientStateMachines::OOB(MintOOBStateMachine {
-                operation_id,
-                state: MintOOBStates::Created(MintOOBStatesCreated {
-                    amount,
-                    spendable_note,
-                    timeout: fedimint_core::time::now() + try_cancel_after,
-                }),
-            }));
-        }
+        let state_machines = vec![MintClientStateMachines::OOB(MintOOBStateMachine {
+            operation_id,
+            state: MintOOBStates::CreatedMulti(MintOOBStatesCreatedMulti {
+                spendable_notes: selected_notes.clone().into_iter_items().collect(),
+                timeout: fedimint_core::time::now() + try_cancel_after,
+            }),
+        })];
 
         Ok((operation_id, state_machines, selected_notes))
     }
@@ -1334,13 +1330,17 @@ impl MintClientModule {
                     match state.state {
                         MintOOBStates::TimeoutRefund(refund) => Some(SpendOOBRefund {
                             user_triggered: false,
-                            transaction_id: refund.refund_txid,
+                            transaction_ids: vec![refund.refund_txid],
                         }),
                         MintOOBStates::UserRefund(refund) => Some(SpendOOBRefund {
                             user_triggered: true,
-                            transaction_id: refund.refund_txid,
+                            transaction_ids: vec![refund.refund_txid],
                         }),
-                        MintOOBStates::Created(_) => None,
+                        MintOOBStates::UserRefundMulti(refund) => Some(SpendOOBRefund {
+                            user_triggered: true,
+                            transaction_ids: vec![refund.refund_txid],
+                        }),
+                        MintOOBStates::Created(_) | MintOOBStates::CreatedMulti(_) => None,
                     }
                 }),
         )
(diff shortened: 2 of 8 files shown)


dpc:
State machine migration (trivial) is missing. Will add if this gets an ack.

fedimint-client/src/transaction/builder.rs
```diff
@@ -13,17 +13,20 @@ use crate::module::StateGenerator;
...
-pub struct ClientInput<I = DynInput, S = DynState> {
+pub struct ClientInput<I = DynInput> {
     pub input: I,
     pub keys: Vec<KeyPair>,
     pub amount: Amount,
+}
+
+#[derive(Clone)]
+pub struct ClientInputSM<S = DynState> {
     pub state_machines: StateGenerator<S>,
```
dpc, line 24: That's the root of refactoring. Now the state machine generator is not part of the `ClientInput` and is passed independently, so "one state machine driving a group of N inputs" is more natural to express.

Kodylow:
@elsirion : currently if you include an invalid ecash note then you can do things independently and it wont fail the transaction. We've seen this tons of times, people using the same seed for multiple wallets.

@dpc we can't compromise on good code for things that shouldn't be happening. The safety net of "if you recover you get the funds back anyway" is still there. Someone could get an oob payment, leave one note and spend the others, but we shouldn't handle that. There's a possibility that we could do "if something fails, binary search for the failing note and separate those out". @elsirion agrees

dpc:
We can land the refactoring that doesn't change the behavior (yet) in https://github.com/fedimint/fedimint/pull/6207

fedimint-client/src/lib.rs
```diff
@@ -289,13 +288,14 @@ pub trait IGlobalClientContext: Debug + MaybeSend + MaybeSync + 'static {
...
+    /// for [`DynGlobalClientContext::claim_inputs`].
     /// Returns transaction id of the funding transaction and an optional
     /// `OutPoint` that represents change if change was added.
-    async fn claim_input_dyn(
+    async fn claim_inputs_dyn(
         &self,
         dbtx: &mut ClientSMDatabaseTransaction<'_, '_>,
-        input: InstancelessDynClientInput,
+        inputs: Vec<InstancelessDynClientInput>,
+        input_sms: Vec<InstancelessDynClientInputSM>,
```
elsirion, line 298: Having the SMs be part of the input type was quite intentional: I don't want developers to be able to claim an input without spawning the required SMs. Maybe we could just ceate an `InstancelessDynClientInputs` type that contains a list of inputs and a list of SMs? That way users can't accidentally mess up. 

Just imagine someone getting the happy path to work by just handing in an empty vec and then forgetting to pass through the SMs.

elsirion (dismissed):

fedimint-client/src/lib.rs
dpc: Done in #6207

modules/fedimint-mint-client/src/oob.rs
```diff
@@ -25,9 +38,11 @@ use crate::{MintClientContext, MintClientStateMachines, SpendableNote};
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub enum MintOOBStates {
+    // Obsoleted, like [`Self::CratedMulti`] but for a single note only.
+    Created(MintOOBStatesCreated),
     /// The e-cash has been taken out of the wallet and we are waiting for the
     /// recipient to reissue it or the user to trigger a refund.
-    Created(MintOOBStatesCreated),
+    CreatedMulti(MintOOBStatesCreatedMulti),
```
elsirion, line 45: Why keep the old one around? If you add the state in the middle instead of the end you have to write a SM migration anyway and could change all `Created` states into `CreatedMulti`. Alternatively you could add the new variant at the end and we could add an empty SM migration, just to increment the DB version and detect going backwards in versions.

modules/fedimint-mint-client/src/oob.rs
```diff
@@ -44,11 +44,18 @@ pub enum MintOOBStates {
...
     /// The timeout of this out-of-band transaction was hit and we attempted to
     /// refund. This refund *failing* is the expected behavior since the
     /// recipient is supposed to have already reissued it.
     TimeoutRefund(MintOOBStatesTimeoutRefund),
+    /// Successfully refunded in a single tx after `UserRefundMulti`
+    RefundedMulti,
+    /// Refunding in a single tx after `UserRefundMulti` failed, started bunch
+    /// of refunds per every spendable note
+    RefundedMultiFallback(MintOOBStatesRefundedMultiFallback),
 }
```
elsirion, line 59: Same here: either only add states at the end or maybe just get rid of the old variant (preferable imo, less code).

Also, we might be getting to the point where we want manual annotation of enum variant indices and should adapt the `Encodable` and `Decodable` macros accordingly.

modules/fedimint-mint-client/src/oob.rs
```diff
@@ -170,6 +196,21 @@ impl MintOOBStatesCreatedMulti {
     }
 }
 
+impl MintOOBStatesUserRefundMulti {
+    fn transitions(
+        &self,
+        global_context: &DynGlobalClientContext,
+    ) -> Vec<StateTransition<MintOOBStateMachine>> {
+        let gc = global_context.clone();
+        let gc2 = global_context.clone();
```
elsirion, line 205: that name :raised_eyebrow:

modules/fedimint-mint-client/src/oob.rs
dpc: I might get rid of them. I haven't considered all the downstream consequences yet, especially w.r.t. notifications etc.

elsirion: If you choose to keep them around I'd advise to add the new variants to the end of the enum to avoid requiring migrations.

modules/fedimint-mint-client/src/oob.rs
dpc: The old state doesn't have all the info (list of notes) to upgrade it to the new state. @elsirion

modules/fedimint-mint-client/src/client_db.rs
```diff
@@ -110,3 +114,27 @@ pub async fn migrate_to_v1(
...
+/// Maps all `Unreachable` states in the state machine to `OutputDone`
+pub(crate) fn migrate_state_to_v2(
+    operation_id: OperationId,
+    cursor: &mut Cursor<&[u8]>,
+) -> anyhow::Result<Option<(Vec<u8>, OperationId)>> {
+    let decoders = ModuleDecoderRegistry::default();
+    let sm_variant = u16::consensus_decode(cursor, &decoders)?;
+
+    // TODO: where the hell does this come from?
+    if sm_variant != 5 {
```
dpc, line 127: @m1sterc001guy Halp!

dpc: OK, I just don't need to do it. Thank you.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1325,13 +1326,23 @@ impl MintClientModule {
...
+                        MintOOBStates::UserRefundMulti(_refund) => None,
+                        MintOOBStates::RefundedMulti => Some(SpendOOBRefund {
+                            user_triggered: true,
+                            // this state already waited for txid, it was a success
+                            transaction_ids: vec![],
+                        }),
+                        MintOOBStates::RefundedMultiFallback(refund) => Some(SpendOOBRefund {
+                            user_triggered: true,
+                            transaction_ids: refund.refund_txids,
+                        }),
```
elsirion, line 1344: In the old state machines the OOB spend SMs weren't responsible for reclaiming e-cash in case the original refund tx got rejected. Instead the e-cash input SMs of the refund tx would take care of reclaiming these notes. I think you actually implemented this redundantly here.

If we keep the old model and only make it multi-note then the code becomes much simpler and we could likely even get rid of old states because we wouldn't need to keep notes around inside `RefundMulti`.

elsirion (commented):
Let's discuss on Thursday or at another convenient time, the PR is mixing SM responsibilities (that likely ought to be documented better, I also forgot) and became unnecessarily complex due to that.

modules/fedimint-mint-client/src/lib.rs
dpc: > In the old state machines the OOB spend SMs weren't responsible for reclaiming e-cash in case the original refund tx got rejected. 

There was no "original refund". It would create bunch of `claim_input` and was done.

Do you mean that the failed `claim_input` should attempt splitting into claim-each-note-separately? Hmmm... That might be quite nice, and it would do it for everything ...

@elsirion

dpc: @elsirion So before, when we did per-note `claim_input` on refunding OOB, we would do another round try to reclaim the reclaim?

dpc:
Let's not merge just yet.

modules/fedimint-mint-client/src/client_db.rs
```diff
@@ -110,3 +114,21 @@ pub async fn migrate_to_v1(
...
 }
+
+/// Maps all `Unreachable` states in the state machine to `OutputDone`
+pub(crate) fn migrate_state_to_v2(
+    operation_id: OperationId,
+    cursor: &mut Cursor<&[u8]>,
+) -> anyhow::Result<Option<(Vec<u8>, OperationId)>> {
+    let decoders = ModuleDecoderRegistry::default();
+
+    let old_state = MintOOBStatesV1::consensus_decode(cursor, &decoders)?;
```
dpc, line 125: This is wrong, isn't it? Which type actually I need to be decoding here? @elsirion @m1sterc001guy

dpc: I see. So it would have to be `MintClientStateMachines`. That's why, instead it is decoding the variant, so it can only deal with one type of the one variant...

dpc:
Started working on continuation in #6264

modules/fedimint-mint-client/src/oob.rs
```diff
@@ -49,6 +69,20 @@ pub struct MintOOBStatesCreated {
...
+    pub(crate) spendable_notes: Vec<(Amount, SpendableNote)>,
+    pub(crate) timeout: SystemTime,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+pub struct MintOOBStatesUserRefundMulti {
+    /// The txid we are hoping succeeds refunding all notes in one go
+    pub(crate) refund_txid: TransactionId,
+    /// The notes we are going to refund individually if it doesn't
+    pub(crate) spendable_notes: Vec<(Amount, SpendableNote)>,
```
elsirion, line 90: Do we still need the spendable notes here now that we discovered the mint input SMs belonging to the refund tx can just take care of this?

If we can avoid that one field we don't need `UserRefundMulti` anymore because the semantics should be the same as the old one (just that now the refund tx contains multiple notes). As a second order effect we could also get rid of the now obsolete `Created` state. 

The reconciliation logic in the SM migration might be more than we want to stomach right now though :/ (we'd need to rewrite SM history such that the update stream still works properly). But we'll have to deal with that eventually anyway I fear. Thoughts @dpc?

modules/fedimint-mint-client/src/oob.rs
```diff
@@ -59,6 +93,11 @@ pub struct MintOOBStatesTimeoutRefund {
     pub(crate) refund_txid: TransactionId,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+pub struct MintOOBStatesRefundedMultiFallback {
+    pub(crate) refund_txids: Vec<TransactionId>,
+}
+
```
elsirion, line 107: After our last discussion this should be osolete

modules/fedimint-mint-client/src/oob.rs
maan2003: you clearly need let gc' :smile:

modules/fedimint-mint-client/src/oob.rs
```diff
@@ -32,6 +32,7 @@ pub enum MintOOBStatesV1 {
 /// ```mermaid
 /// graph LR
 ///     Created -- User triggered refund --> RefundU["User Refund"]
+///     RefundU -- Refund failed --> RefundedMultiFallback["User Refund Fallback"]
```
elsirion, line 35: This isn't relevant anymore, right? Might be still worth mentioning the legacy and the new states.

elsirion (approved):
Ok enough if you want to move forward, but we are definitely adding tech debt here.

modules/fedimint-mint-client/src/oob.rs
dpc: Leftover. Deleted.

modules/fedimint-mint-client/src/oob.rs
dpc: @elsirion 

I think it's just easier to let the old states be as they are, and then after like two releases replace their code with `Err("No longer supported")`. If we start doing complex migrations, then we should have some (probably also complex) tests, etc.

modules/fedimint-mint-client/src/oob.rs
dpc: Leftover. Deleted.

elsirion (approved):

PR #6180, opened 2024-10-18
chore(client): check peg-in addresses with no deposits more often

dpc:
Currently, if the user creates a deposit address, and then decides to make a peg-in after a week, they might wait 3.5 day for the client to notice.

Ultimately end-user apps needs to expose ability to check for deposits on demand (click a button or something), but irrespective of that, checking a 7 day old address arond once a day seems reasonable.


m1sterc001guy (approved):
Seems reasonable to me, would be good to get @bradleystachurski's ACK

bradleystachurski (approved):
I think this is a good improvement :+1: 

What do you think about capping the delay at one day regardless of the age?

dpc:
> What do you think about capping the delay at one day regardless of the age?

Then we'll be forever checking addresses that never got a deposit.

PR #6179, opened 2024-10-17
chore: replace our cargo wrappers with cargo-deluxe

dpc:
It discovered a problem causing needless rebuilds and slowing down our CI.

PR #6182, opened 2024-10-18
chore: update nix flake

tvolk131:


m1sterc001guy (approved):

dpc:
We should https://github.com/fedimint/fedimint/pull/6179 instead, I believe it does the same thing and will just conflict with this one.

PR #6195, opened 2024-10-21
chore: update rust toolchain

dpc:

### DIFF
--- a/fedimint-core/src/db/mod.rs
+++ b/fedimint-core/src/db/mod.rs
@@ -2215,6 +2215,9 @@ pub async fn apply_migrations_server(
     apply_migrations(db, kind, migrations, None, None).await
 }
 
+/// `apply_migrations` iterates from the on disk database version for the
+/// module.
+///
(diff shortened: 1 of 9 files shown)


Kodylow:
how's this only changing flake lock not the flake?

dpc:
> how's this only changing flake lock not the flake?

`fenix` input is not pinned to `rev`, so I just `nix flake lock --update-input fenix` and that updates the pinned version in the lock file, job done.

bradleystachurski (dismissed):
Looks like clippy is complaining after the update, but otherwise looks good :+1:

bradleystachurski (approved):

fedimint-core/src/db/mod.rs
```diff
@@ -2215,6 +2215,9 @@ pub async fn apply_migrations_server(
     apply_migrations(db, kind, migrations, None, None).await
 }
 
+/// `apply_migrations` iterates from the on disk database version for the
```
dpc, line 2218: Clippy now wants a nice and short title on everything.

Issue #5057, opened 2024-04-22
Extract protobuf definition and generated code into separate crate


m1sterc001guy:
We stuffed the protocol buffer definition into the Lightning Gateway crate and never moved it. Ideally this would be inside a lightweight crate by itself so only crates that need gRPC/protocol buffers depend on it. Currently anything that depends on `ln-gateway`, such as `fedimint-testing` needs to have `protoc` defined, which makes it hard for downstream crates to integrate `fedimint-testing`

justinmoon:
dev call: we could check-in protobuf definitions, but need CI to check it and just command to update.

Could also switch away from protobuf towards JSON-RPC which is used elsewhere (should be separate issue)

PR #6197, opened 2024-10-21
chore: remove CLN grpc server and generated types

m1sterc001guy:
gRPC from gatewayd to the cln extension doesn't give us much benefit currently and its generated types are hard to work with. We have talked about removing it for awhile. This PR removes gRPC and the generated types are replaces it with axum and native rust types.

Fixes: https://github.com/fedimint/fedimint/issues/5057


vincenzopalazzo (commented):
SGTM!

Kodylow:
@elsirion : need to look into the htlc streaming, if you're using a reverse proxy in front of a streaming http many of them will just close them off thinking that its a dos vector. Using a websocket directly is way better.

@dpc : if you can lose state on disconnect/reconnect the whole api is broken.

m1sterc001guy:
> need to look into the htlc streaming, if you're using a reverse proxy in front of a streaming http many of them will just close them off thinking that its a dos vector.

Is this any different from gRPC streaming? gRPC is also using a HTTP server underneath.

> if you can lose state on disconnect/reconnect the whole api is broken.

What does "lose state" mean in this context? Even with gRPC, if the connection between `gatewayd` and the extension is down and a HTLC arrives, we will not see this HTLC and the payment will fail. This refactor hasn't changed anything there, the behavior is the same as before. We also already have disconnect/reconnect tests that verify the lightning node can go down and come back and payments will continue to work once the lightning node is back online.

elsirion:
> > need to look into the htlc streaming, if you're using a reverse proxy in front of a streaming http many of them will just close them off thinking that its a dos vector.
> 
> Is this any different from gRPC streaming? gRPC is also using a HTTP server underneath.

Good point, I just fear the unknown unknowns when it comes to a less-used features like HTTP streaming. gRPC kinda handles it end-to end. But if you feel very comfortable with the choice then let's go for it, we also lack experience with actual WS JSON RPC notifications, which would be my choice otherwise.

> > if you can lose state on disconnect/reconnect the whole api is broken.
> 
> What does "lose state" mean in this context? Even with gRPC, if the connection between `gatewayd` and the extension is down and a HTLC arrives, we will not see this HTLC and the payment will fail. This refactor hasn't changed anything there, the behavior is the same as before. We also already have disconnect/reconnect tests that verify the lightning node can go down and come back and payments will continue to work once the lightning node is back online.

I was just mentioning that it would be bad if some middleware that is unaware of HTTP streaming would enforce a max connection time of a few minutes and then we'd lose the stream for a second every 5 minutes or so. The problem exists either way, but if we have a very reliable HTLC stream connection it's not as bad.

m1sterc001guy:
Good points. I don't have much experience with HTTP streaming (outside of gRPC), so I don't know if those things will be a problem or not. Given that we are planning on not supporting CLN in v0.6 (and the axum HTTP streaming stuff only affects CLN), I think its worth the risk just to remove the gRPC generated types.

m1sterc001guy:
Converting to draft until CI flakiness is resolved

joschisan (dismissed):

joschisan (approved):

joschisan:
lets get this merged.

PR #6198, opened 2024-10-21
chore: simplify and test liquidity management

joschisan:


gateway/ln-gateway/src/lib.rs
```diff
@@ -984,20 +985,28 @@ impl Gateway {
...
-        payload: PayInvoicePayload,
+        payload: PayInvoiceForOperatorPayload,
     ) -> AdminResult<Preimage> {
         let GatewayState::Running { lightning_context } = self.get_state().await else {
             return Err(AdminGatewayError::Lightning(
                 LightningRpcError::FailedToConnect,
             ));
         };
 
+        let max_fee = 50
```
m1sterc001guy, line 998: Can we make this a constant with an explanation for why 50 was chosen?

gateway/ln-gateway/src/lib.rs
```diff
@@ -984,20 +985,28 @@ impl Gateway {
...
+            + payload
+                .invoice
+                .amount_milli_satoshis()
+                .context("Invoice is missing amount")?
+                .saturating_div(100);
+
         let res = lightning_context
             .lnrpc
-            .pay(payload.invoice, payload.max_delay, payload.max_fee)
+            .pay(payload.invoice, 1000, Amount::from_msats(max_fee))
```
m1sterc001guy, line 1007: Let's make 1000 a constant too

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -148,17 +148,13 @@ pub struct SetConfigurationPayload {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-pub struct CreateInvoiceForSelfPayload {
+pub struct CreateInvoiceForOperatorPayload {
     pub amount_msats: u64,
-    pub expiry_secs: u32,
-    pub description: Option<String>,
```
m1sterc001guy, line 154: Operators might still want to attach a description? Do we need to remove this?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -194,7 +189,11 @@ fn v1_routes(gateway: Arc<Gateway>, task_group: TaskGroup) -> Router {
         .route(CONNECT_FED_ENDPOINT, post(connect_fed))
         .route(LEAVE_FED_ENDPOINT, post(leave_fed))
         .route(BACKUP_ENDPOINT, post(backup))
-        .route(PAY_INVOICE_SELF_ENDPOINT, post(pay_invoice_self))
+        .route(
+            CREATE_BOLT11_INVOICE_FOR_OPERATOR_ENDPOINT,
+            post(create_invoice_for_self),
```
m1sterc001guy, line 194: Why does this need to be authenticated? Do we care if "others" create invoices and send money to the gateway?

m1sterc001guy:
Concept ACK, would just like the small comments addressed. Are we still testing that we can create an receive a HOLD invoice in LND? We need to test this IMO, LNv2 uses HOLD invoices so we want to verify that there's no regressions due to the LNv2 implementation.

gateway/ln-gateway/src/rpc/mod.rs
joschisan: Since invoices are really only for yourself descriptions seemed unnecessary. We can always add this in later.

gateway/ln-gateway/src/rpc/rpc_server.rs
joschisan: As its for the operator to manage liquidity seems like this should be gated.

gateway/ln-gateway/src/lib.rs
joschisan: Its just the default of ldk, 1 % plus 50 sats.

joschisan: Personally, I would not choose a constant here as the value is not critical and it makes the code less localized.

joschisan:
@m1sterc001guy I added the check for the hold invoice.

gateway/ln-gateway/src/rpc/mod.rs
m1sterc001guy: They might not be for yourself, you might hand them out to someone else to pay you.

gateway/ln-gateway/src/rpc/rpc_server.rs
m1sterc001guy: But why? Other users might want to generate an invoice for paying the gateway too, doesn't seem like we should prevent that?

gateway/ln-gateway/src/lib.rs
m1sterc001guy: You can make the constant in the same function so that its not reused. I don't usually like directly inlining numbers in code because (usually after some time), people forget why it was there in the first place and it just becomes a magic number.

gateway/ln-gateway/src/rpc/rpc_server.rs
joschisan: In what situation you want users to just randomly send you money? Furthermore, if its not public we can return an admin error here.

gateway/ln-gateway/src/rpc/mod.rs
joschisan: In that case we should make the expiry_secs configurable as well. Added them back in.

gateway/ln-gateway/src/lib.rs
joschisan: Done.

gateway/ln-gateway/src/lib.rs
joschisan: Done.

modules/fedimint-lnv2-tests/src/bin/tests.rs
```diff
@@ -16,10 +16,6 @@ use tracing::info;
 #[derive(Parser)]
 enum TestOpts {
     All,
-    GatewayRegistration,
-    SelfPaymentsRefund,
-    SelfPaymentsSuccess,
-    LightningPayment,
     InterFederation,
```
tvolk131, line 19: Looks like both on master and in this PR, `TestOpts::All` is running all tests except for inter-federation tests. Is there a reason for that? If so, can you document it?

tvolk131 (changes requested):

modules/fedimint-lnv2-tests/src/bin/tests.rs
joschisan: The remaining two variants are tests that are run in CI in parallel, the other one are not used for that.

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -308,20 +310,22 @@ async fn withdraw(
...
+) -> Result<impl IntoResponse, AdminGatewayError> {
+    let invoice = gateway
+        .handle_create_invoice_for_operator_msg(payload)
+        .await?;
     Ok(Json(json!(invoice)))
 }
 
 #[instrument(skip_all, err, fields(?payload))]
-async fn pay_invoice_self(
+async fn pay_invoice_operator(
```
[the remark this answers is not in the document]
joschisan: Done

gateway/cli/src/lightning_commands.rs
```diff
@@ -23,25 +23,14 @@ const DEFAULT_WAIT_FOR_CHAIN_SYNC_RETRY_DELAY_SECONDS: u64 = 2;
 pub enum LightningCommands {
     /// Create an invoice to receive lightning funds to the gateway.
     CreateInvoice {
-        #[clap(long)]
         amount_msats: u64,
-
-        #[clap(long, default_value_t = 300)]
-        expiry_secs: u32,
-
+        #[clap(long)]
```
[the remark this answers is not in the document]
joschisan: Done

tvolk131 (dismissed):

tvolk131 (dismissed):

tvolk131 (approved):

scripts/tests/lnv2-inter-federation.sh
```diff
@@ -8,4 +8,5 @@ source scripts/_common.sh
 build_workspace
 add_target_dir_to_path
 
-tests inter-federation
\ No newline at end of file
+# TODO: fix flakiness and reactivate or remove test completely
+# tests inter-federation
```
m1sterc001guy, line 12: Flakiness should be better now after https://github.com/fedimint/fedimint/pull/6217. Please revert

m1sterc001guy (changes requested):
Looks good but inter-federation tests shouldn't be flaky anymore

scripts/tests/lnv2-inter-federation.sh
joschisan: Done

m1sterc001guy (approved):

Issue #5071, opened 2024-04-23
Gateway should have a mode where it returns payment errors to users


elsirion:
> dev call: gateway should have a mode where it gives more verbose error messages.

_Originally posted by @justinmoon in https://github.com/fedimint/fedimint/issues/4943#issuecomment-2070059520_
            

If `FM_DEBUG_GATEWAY` is set the cause of payment failures should be communicated to the client.

PR #6200, opened 2024-10-21
chore: debug gateway environment variable

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/5071

Might be useful in the future.


gateway/ln-gateway/src/error.rs
```diff
@@ -55,6 +56,12 @@ impl IntoResponse for PublicGatewayError {
             ),
         };
 
+        let error_message = if std::env::var(FM_DEBUG_GATEWAY_ENV).is_ok() {
```
bradleystachurski, line 59: [nit] There's a helper `is_env_var_set` that can be useful for boolean-like env vars since you can also set them to `0` or `false` to unset.

https://github.com/fedimint/fedimint/blob/c22a0b017d47824a076d88a32a9ab80cc4f2818a/fedimint-core/src/envs.rs#L19

bradleystachurski (dismissed):

gateway/ln-gateway/src/error.rs
m1sterc001guy: Fixed!

bradleystachurski (approved):

PR #6203, opened 2024-10-22
fix: devimint checks if txs are visible anywhere

tvolk131:
Currently we're calling `getrawtransaction` to check whether a particular transaction has been seen by bitcoind (for example, to wait for a channel open transaction to appear before mining blocks). Since we have `txindex` disabled in v0.4 and later, `getrawtransaction` will only return transactions in the mempool, not ones that have already been mined.

https://github.com/fedimint/fedimint/blob/5f56672f3c9934b9ebd8deaa6a3177f5108db7f5/devimint/src/external.rs#L52-L56

However, in devimint there are multiple parallel processes that each mine blocks on bitcoind, which can lead to a transaction being submitted in one process, confirmed in another, and then checked for existence in the first process. In this PR we add `get_transaction()` which checks both the mempool and the entire blockchain to see if a transaction exists. Then we expose `poll_get_transaction()` which repeatedly calls it. This should reduce flakiness in CI.

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -7,6 +7,7 @@ use std::time::Duration;
 use anyhow::{anyhow, bail, Context, Result};
 use bitcoincore_rpc::bitcoin::{Address, BlockHash};
 use bitcoincore_rpc::bitcoincore_rpc_json::{GetBalancesResult, GetBlockchainInfoResult};
+use bitcoincore_rpc::jsonrpc::error::RpcError;
 use bitcoincore_rpc::RpcApi;
 use cln_rpc::primitives::{Amount as ClnRpcAmount, AmountOrAny};
 use cln_rpc::ClnRpc;
@@ -17,6 +18,7 @@ use fedimint_core::util::write_overwrite_async;
 use fedimint_core::BitcoinHash;
 use fedimint_logging::LOG_DEVIMINT;
 use fedimint_testing::gateway::LightningNodeType;
+use futures::StreamExt;
 use hex::ToHex;
 use itertools::Itertools;
 use ln_gateway::envs::FM_CLN_EXTENSION_LISTEN_ADDRESS_ENV;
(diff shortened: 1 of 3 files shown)


devimint/src/external.rs
```diff
@@ -237,13 +239,53 @@ impl Bitcoind {
...
+    /// Get a transaction by its txid. Checks the mempool and all blocks.
+    pub async fn get_transaction(&self, txid: bitcoin::Txid) -> Result<String> {
+        // Check the mempool.
+        if let Ok(tx) = self.get_raw_transaction(txid, None).await {
+            return Ok(tx);
+        };
+
+        let block_height = self.get_block_count().await? - 1;
+
+        let mut per_block_check_futures = FuturesUnordered::new();
```
dpc, line 251: Even in a medium size test, this can lead to tens of thousands of futures, rpc calls and possibly start hitting some limits.

dpc: I'm thinking ... I would change this to `buffered` stream going in reverse from the highest height to the lowest with some windows like 128...

dpc:
```
futures::stream::iter((0..block_height).rev()).map(...).buffered(128)
```

tvolk131: Good call! Done.

dpc (approved):

m1sterc001guy (approved):

m1sterc001guy:
Let's hope this improves devimint's reliability, CI has been really flaky

dpc (approved):

devimint/src/external.rs
```diff
@@ -240,13 +241,54 @@ impl Bitcoind {
...
+        // the bitcoind with block requests.
+        let mut buffered_tx_stream = futures::stream::iter((0..block_height).rev())
+            .map(|height| async move {
+                let block_hash = self.get_block_hash(height).await?;
+                let tx = self.get_raw_transaction(txid, Some(block_hash)).await?;
+                Ok::<_, anyhow::Error>(tx)
+            })
+            .buffered(32);
+
+        while let Some(tx_or) = buffered_tx_stream.next().await {
```
m1sterc001guy, line 266: Do you need a loop here? Should one be one transaction right?

dpc: Oh, actually, shouldn't we be retrying on errors here anyway?

elsirion: I could imagine polling the mempool while scanning the chain as a performance improvement, but fundamentally the caller needs to re-try if nothing is found.

devimint/src/external.rs
```diff
@@ -240,13 +241,54 @@ impl Bitcoind {
...
+        // Check each block for the tx, starting at the chain tip.
+        // Buffer the requests to avoid spamming bitcoind.
+        // We're doing this after checking the mempool since the tx should
+        // usually be in the mempool, and we don't want to needlessly hit
+        // the bitcoind with block requests.
+        let mut buffered_tx_stream = futures::stream::iter((0..block_height).rev())
+            .map(|height| async move {
+                let block_hash = self.get_block_hash(height).await?;
+                let tx = self.get_raw_transaction(txid, Some(block_hash)).await?;
+                Ok::<_, anyhow::Error>(tx)
```
elsirion, line 262: Can we distinguish `Ok(None)` from `Err(_)` somehow? Seems to me that we should return on error (we expect `bitcoind` to respond correctly) and just continue scanning if we just couldn't find the tx.

devimint/src/external.rs
dpc: Well, the function is called `get_transaction`, and the caller might have a reason to know that the tx is already there, etc.

Anything OK, as long as we don't swallow errors.

devimint/src/external.rs
tvolk131: Good point! Done. It's a bit hacky, but I dug into the bitcoind source code and figured out that a JSON-RPC response with an error code of -5 indicates that the transaction doesn't exist. I linked to the relevant code in this PR.

devimint/src/external.rs
tvolk131: I changed this so we can now differentiate between the RPC call failing vs the transaction not being found. We now return an error if an RPC fails but continue checking more blocks if the transaction isn't found in a given block.

devimint/src/external.rs
```diff
@@ -240,11 +242,73 @@ impl Bitcoind {
...
+        // https://github.com/bitcoin/bitcoin/blob/25dacae9c7feb31308271e2fd5a127c1fc230c2f/src/rpc/rawtransaction.cpp#L360-L376
+        // https://github.com/bitcoin/bitcoin/blob/25dacae9c7feb31308271e2fd5a127c1fc230c2f/src/rpc/protocol.h#L42
+        let tx = match tx_or {
+            Ok(tx) => tx,
+            Err(err) => match err {
+                bitcoincore_rpc::Error::JsonRpc(bitcoincore_rpc::jsonrpc::Error::Rpc(
+                    RpcError { code: -5, .. },
+                )) => return Ok(None),
+                _ => return Err(err.into()),
+            },
```
elsirion, line 303: Nit: nested match is unnecessary

devimint/src/external.rs
```diff
@@ -240,11 +242,73 @@ impl Bitcoind {
...
+        // reaches bitcoind but is not found. See here:
+        // https://github.com/bitcoin/bitcoin/blob/25dacae9c7feb31308271e2fd5a127c1fc230c2f/src/rpc/rawtransaction.cpp#L360-L376
+        // https://github.com/bitcoin/bitcoin/blob/25dacae9c7feb31308271e2fd5a127c1fc230c2f/src/rpc/protocol.h#L42
+        let tx = match tx_or {
+            Ok(tx) => tx,
+            Err(err) => match err {
+                bitcoincore_rpc::Error::JsonRpc(bitcoincore_rpc::jsonrpc::Error::Rpc(
+                    RpcError { code: -5, .. },
+                )) => return Ok(None),
+                _ => return Err(err.into()),
```
elsirion, line 302: I wonder if we should log here. Some callers ignore the error case, which should never happen imo. It's probably good to not panic to avoid flakes, but if there is consistent failure reported from `bitcoind` we'd want to see it in the test logs.

devimint/src/external.rs
tvolk131: Fixed.

devimint/src/external.rs
tvolk131: I fixed this by adding `poll_get_transaction()` so all callers exit on the error case while continuing to poll on the `Ok(None)` case. I don't think we need to log here, since `poll_get_transaction()` calls `.context("getrawtransaction")` on the error case. That should be enough to see what's happening, right?

devimint/src/external.rs
```diff
@@ -240,11 +242,90 @@ impl Bitcoind {
...
+            match tx_or {
+                // The RPC succeeded, and the transaction was not found in the block. Continue to
+                // the next block.
+                Ok(None) => continue,
+                // The RPC failed, or the transaction was found in the block. Return the result.
+                other => return other,
+            };
+        }
+
+        Err(anyhow!("Transaction not found"))
```
elsirion, line 299: I think this needs to return `Ok(None)` for the retry logic to work.

tvolk131: Whoops! Good call, looks like that fixed CI

elsirion: Just verified that this fixes the devimint tests

PR #6211, opened 2024-10-23
chore: make read-only default in recoverytool

### LINKED ISSUE #5966: Make `recoverytool` always run in read-only mode
This will prevent accidentally creating an empty DB if the DB path is incorrect and we only read anyway.

elsirion:
Fixes #5966

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -2079,7 +2079,6 @@ pub async fn recoverytool_test(dev_fed: DevFed) -> Result<()> {
     info!("Recovering using utxos method");
     let output = cmd!(
         crate::util::Recoverytool,
-        "--readonly",
         "--cfg",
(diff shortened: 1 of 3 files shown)


m1sterc001guy (approved):
Failed tests just look to be flakes, we'll need to debug later, but this change is certainly not the cause

douglaz:
What's the syntax to disable it?

douglaz:
I don't think you can disable the `readonly` now. Perhaps better just remove the flag and let it be true?

elsirion:
> I don't think you can disable the `readonly` now. Perhaps better just remove the flag and let it be true?

Great point :see_no_evil:. I changed the PR to just always use readonly.

m1sterc001guy (approved):

dpc (approved):

Issue #5200, opened 2024-05-03
Upgrade to `bitcoin` v0.31


tvolk131:
Follow-up to #2356

This should be much easier than the upgrade from v0.29 to v0.30! I attempted to do this work but realized that a few dependencies we use that also depend on the `bitcoin` crate haven't yet published versions that depend on v0.31. At time of writing, these include:

* `cln-rpc`
* `lightning`
* `lightning-invoice`
* `lnurl-rs`

Some notes from my attempt at the upgrade:

* Bump `secp256k1` from v0.27 to v0.28
* Bump `secp256k1-zkp` from v0.9 to v0.10
* Bump `miniscript` from v10 to v11
* Bump `bitcoincore-rpc` from v0.17 to v0.18
* Bump `electrum-client` from v0.18 to v0.19
* Bump `esplora-client` from v0.6 to v0.7
* Rename `bitcoin::psbt::PartiallySignedTransaction` to `bitcoin::psbt::Psbt`
* Rename `bitcoin::network::Magic` to `bitcoin::p2p::Magic`
* Rename `bitcoin::key::KeyPair` to `bitcoin::key::Keypair`

tvolk131:
Looks like `lightning` and `lightning-invoice` bumped their versions of `bitcoin` straight from v0.30 to v0.32, so it looks like we'll need to do the same.

PR #6212, opened 2024-10-23
chore(deps): partially bump bitcoin to v0.32

tvolk131:
Part of #5200


fedimint-core/src/bitcoin_migration.rs
```diff
@@ -110,6 +110,15 @@ pub fn bitcoin30_to_bitcoin32_payment_preimage(
     lightning_types::payment::PaymentPreimage(preimage.0)
 }
 
+pub fn bitcoin30_to_bitcoin32_sha256_hash(
+    hash: &bitcoin30::hashes::sha256::Hash,
+) -> bitcoin::hashes::sha256::Hash {
+    bincode::deserialize(
+        &bincode::serialize(&hash).expect("Failed to serialize bitcoin30 sha256 hash"),
+    )
+    .expect("Failed to convert bitcoin30 sha256 hash to bitcoin32 sha256 hash")
```
elsirion, line 196: That's the third creative way you have found to transform hashes in rust-bitcoin :laughing: 

When you look at the [`sha256::Hash` impl](https://docs.rs/bitcoin_hashes/latest/bitcoin_hashes/sha256/struct.Hash.html) there are methods to converting from/to byte array references:

Screenshot 2024-10-24 at 10-22-57 Hash in bitcoin_hashes sha256 - Rust
Screenshot 2024-10-24 at 10-22-49 Hash in bitcoin_hashes sha256 - Rust

elsirion (changes requested):
I'm pretty sure you don't need `bincode` for any of these conversions and we should avoid introducing dependencies whenever posssible. Please look at the struct's docs, there should always be some from/to bytes conversion functions.

fedimint-core/src/bitcoin_migration.rs
tvolk131: Haha achievement unlocked...?

Fixed. Could probably optimize some of the existing migration functions, but will do that in a follow-up.

elsirion: Idk if necssary to touch other fns just yet, ideally we will get rid of them soonish anyway. Just introducing new ones seems very unnecessary. If we get stuck for some reason I'd do the refactor.

elsirion: > Fixed

Did you push?

tvolk131: > Did you push?

Whoops! Pushed.

fedimint-core/src/bitcoin_migration.rs
```diff
@@ -50,89 +50,74 @@ pub fn checked_address_to_unchecked_address(
...
 
 pub fn bitcoin30_to_bitcoin32_secp256k1_pubkey(
     pubkey: &bitcoin30::secp256k1::PublicKey,
 ) -> bitcoin::secp256k1::PublicKey {
-    bincode::deserialize(
-        &bincode::serialize(&pubkey).expect("Failed to serialize bitcoin30 secp256k1 pubkey"),
-    )
-    .expect("Failed to convert bitcoin30 secp256k1 pubkey to bitcoin32 secp256k1 pubkey")
+    bitcoin::secp256k1::PublicKey::from_str(&pubkey.to_string())
+        .expect("Failed to convert bitcoin30 secp256k1 pubkey to bitcoin32 secp256k1 pubkey")
```
elsirion, line 95: Screenshot 2024-10-27 at 09-01-41 PublicKey in secp256k1 - Rust
Screenshot 2024-10-27 at 09-01-04 PublicKey in secp256k1 - Rust

elsirion (commented):
Let's avoid stingification (it's ok for addresses, but keys have better methods for encoding/decoding)

fedimint-core/src/bitcoin_migration.rs
tvolk131: Done. However, `PublicKey::from_byte_array_compressed()` is introduced in secp256k1 v0.30, but bitcoin v0.30 relies on secp256k1 v0.29, so the best way to do this is using `PublicKey::from_slice()` and then calling `.expect()`.

elsirion (approved):

PR #6216, opened 2024-10-24
chore: bump Network and Amount en/decoding to bitcoin v0.32

tvolk131:
Part of #5200


elsirion:
There's likely a lot of deterministic, automated refactoring going on here. For the next PRs of that kind, could you try to separate out that part into a separate commit and include the command to generate it in the commit message? That way we can automate review a bit and focus on the parts where a human had to make decisions.

joschisan (approved):

PR #6222, opened 2024-10-25
chore: comment gateway environment variables

m1sterc001guy:
Just adding some comments. Likely helpful for deployments for figuring out how to configure the gateway.

### DIFF
--- a/gateway/ln-gateway/src/envs.rs
+++ b/gateway/ln-gateway/src/envs.rs
@@ -1,52 +1,79 @@
-// Env variable to TODO
-pub const FM_CLN_EXTENSION_LISTEN_ADDRESS_ENV: &str = "FM_CLN_EXTENSION_LISTEN_ADDRESS";
-
-// Env variable to TODO
+/// Environment variable that specifies the directory of the gateway's database.
 pub const FM_GATEWAY_DATA_DIR_ENV: &str = "FM_GATEWAY_DATA_DIR";
 
-// Env variable to TODO
+/// Environment variable that specifies the address the gateway's HTTP server
+/// should listen on.
 pub const FM_GATEWAY_LISTEN_ADDR_ENV: &str = "FM_GATEWAY_LISTEN_ADDR";
 
-// Env variable to TODO
+/// Environment variable that specifies the URL that clients can use to make
+/// requests to the gateway.
 pub const FM_GATEWAY_API_ADDR_ENV: &str = "FM_GATEWAY_API_ADDR";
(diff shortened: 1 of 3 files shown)


gateway/ln-gateway/src/envs.rs
```diff
@@ -1,50 +1,75 @@
...
-pub const FM_CLN_EXTENSION_LISTEN_ADDRESS_ENV: &str = "FM_CLN_EXTENSION_LISTEN_ADDRESS";
-
-// Env variable to TODO
+// Environment variable that specifies the directory of the gateway's database.
 pub const FM_GATEWAY_DATA_DIR_ENV: &str = "FM_GATEWAY_DATA_DIR";
 
-// Env variable to TODO
+// Environment variable that specifies the address the gateway's HTTP server
+// should listen on.
 pub const FM_GATEWAY_LISTEN_ADDR_ENV: &str = "FM_GATEWAY_LISTEN_ADDR";
```
elsirion, line 6: Since you are already touching the comments, can you make them doc comments please (three slashes)?

m1sterc001guy: Done

gateway/ln-gateway/src/envs.rs
```diff
@@ -1,50 +1,77 @@
...
+pub const FM_GATEWAY_LIGHTNING_ADDR_ENV: &str = "FM_GATEWAY_LIGHTNING_ADDR";
+
+/// Environment variable that specifies the mnemonic that the gateway should use
+/// for ecash and the LDK Node should use for onchain funds. If not set, a
+/// mnemonic will be generated. This environment variable can be used for
+/// recovering from an existing mnemonic.
 pub const FM_GATEWAY_MNEMONIC_ENV: &str = "FM_GATEWAY_MNEMONIC";
 
+/// Environment variable that specifies the "module mode" the gateway should run
+/// in. Options are "LNv1", "LNv2", or "All". It is not recommended to run "all"
```
bradleystachurski, line 69: [nit] Is "All" case sensitive? If so, then it may be more clear to `s/all/All`

bradleystachurski (dismissed):
Thanks for adding docs!

gateway/ln-gateway/src/envs.rs
m1sterc001guy: not case sensitive but good to be consistent. Fixed in the rebase :)

bradleystachurski (approved):

PR #6224, opened 2024-10-25
cleanup: remove inter federation test

joschisan:


modules/fedimint-lnv2-tests/src/bin/tests.rs
```diff
@@ -42,69 +32,36 @@ async fn main() -> anyhow::Result<()> {
...
-        .expect("Gateways of version 0.5.0 or higher support LDK");
-
-    federation
-        .pegin_gateways(1_000_000, vec![gw_lnd, gw_ldk])
-        .await?;
-
-    Ok(())
-}
-
 async fn test_gateway_registration(dev_fed: &DevJitFed) -> anyhow::Result<()> {
```
m1sterc001guy, line 58: nit: `add_gateway` can be done in parallel, might make it a bit faster

Issue #6123, opened 2024-10-03
`get_balances` inbound liquidity should only include active channels


m1sterc001guy:
Inactive channels cannot be used for receiving

m1sterc001guy:
Ah sorry @tvolk131, didn't realize you were working on this. I sent a PR

tvolk131:
All good! I've been busy with bitcoin migration and ldk gateway stuff and hadn't started working on this yet

PR #6225, opened 2024-10-26
feat: get_balances filters non-active channels

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/6123


tvolk131 (approved):
Nice, LGTM! And no worries about snagging that issue, I hadn't started on it yet. 🙂

PR #6230, opened 2024-10-27
fix: make `GatewayLdkClient::pay()` idempotent

tvolk131:
We recently added more fine-grained documentation for the API contract of `ILnRpcClient::pay()` that outlines the requirement that it behave idempotently. See the "Implementor restrictions" below:

https://github.com/fedimint/fedimint/blob/8826dc03a50b4053e2e116640ee398105711a03c/gateway/ln-gateway/src/lightning/mod.rs#L109-L130

This PR aligns the LDK gateway with the above spec.

Currently, `GatewayLdkClient::pay()` simply calls [`ldk-node::Bolt11Payment::send()`](https://docs.rs/ldk-node/0.4.2/ldk_node/payment/struct.Bolt11Payment.html#method.send), which does not follow any of the three idempotency rules above. If the payment is already in-flight or has been successfully paid, it returns an error stating that the invoice cannot be paid twice. And if a payment has been attempted and failed, a second call will retry it rather than returning an error. See [here](https://docs.rs/ldk-node/0.4.2/src/ldk_node/payment/bolt11.rs.html#102-109).

### DIFF
--- a/gateway/ln-gateway/src/lightning/ldk.rs
+++ b/gateway/ln-gateway/src/lightning/ldk.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::path::Path;
 use std::str::FromStr;
 use std::sync::Arc;
@@ -11,7 +12,9 @@ use fedimint_core::envs::is_env_var_set;
 use fedimint_core::task::{block_in_place, TaskGroup, TaskHandle};
 use fedimint_core::{Amount, BitcoinAmountOrAll};
 use fedimint_ln_common::contracts::Preimage;
+use futures::lock::Mutex;
 use ldk_node::config::EsploraSyncConfig;
+use ldk_node::lightning::events::PaymentFailureReason;
 use ldk_node::lightning::ln::msgs::SocketAddress;
 use ldk_node::lightning::ln::PaymentHash;
 use ldk_node::lightning::routing::gossip::NodeAlias;
@@ -32,6 +35,20 @@ use crate::lightning::{
 };
 use crate::rpc::{CloseChannelsWithPeerPayload, OpenChannelPayload, SendOnchainPayload};
 
+/// A collection of payment subscriptions, indexed by payment hash.
+/// Each payment hash maps to a list of oneshot `Sender`s that should be
+/// consumed when the payment completes, allowing for multiple subscribers to
+/// the same payment.
+type PaymentSubscriptions = BTreeMap<
+    bitcoin::hashes::sha256::Hash,
+    Vec<tokio::sync::oneshot::Sender<PaymentSubscriptionOutcome>>,
+>;
+
+enum PaymentSubscriptionOutcome {
+    Success(Preimage),
+    Failure(PaymentFailureReason),
+}
+
 pub struct GatewayLdkClient {
     /// The underlying lightning node.
     node: Arc<ldk_node::Node>,
@@ -44,6 +61,17 @@ pub struct GatewayLdkClient {
     /// The HTLC stream, until it is taken by calling
     /// `ILnRpcClient::route_htlcs`.
     htlc_stream_receiver_or: Option<tokio::sync::mpsc::Receiver<InterceptPaymentRequest>>,
+
+    /// The set of outgoing payments, by payment hash, that are currently
+    /// awaiting completion. This is used to allow `Self::handle_next_event()`
+    /// to notify `Self::pay()` when a payment completes.
+    ///
+    /// Note: Since this field is needed to process `PaymentSuccessful` and
+    /// `PaymentFailed` events from the LDK node, locking it will block all
+    /// event processing, including `PaymentClaimable` events, which are needed
+    /// to process inbound federation payments. Do not lock this field for
+    /// long periods of time.
+    payment_subscriptions: Arc<Mutex<PaymentSubscriptions>>,
 }
 
 impl std::fmt::Debug for GatewayLdkClient {
@@ -114,10 +142,19 @@ impl GatewayLdkClient {
         let (htlc_stream_sender, htlc_stream_receiver) = tokio::sync::mpsc::channel(1024);
         let task_group = TaskGroup::new();
 
+        let payment_subscriptions = Arc::new(Mutex::new(PaymentSubscriptions::new()));
+
         let node_clone = node.clone();
+        let payment_subscriptions_clone = payment_subscriptions.clone();
         task_group.spawn("ldk lightning node event handler", |handle| async move {
             loop {
-                Self::handle_next_event(&node_clone, &htlc_stream_sender, &handle).await;
+                Self::handle_next_event(
+                    &node_clone,
+                    &htlc_stream_sender,
+                    &handle,
+                    &payment_subscriptions_clone,
+                )
+                .await;
             }
         });
 
@@ -126,6 +163,7 @@ impl GatewayLdkClient {
             esplora_client: esplora_client::Builder::new(esplora_server_url).build_async()?,
             task_group,
             htlc_stream_receiver_or: Some(htlc_stream_receiver),
+            payment_subscriptions,
         })
     }
 
@@ -133,6 +171,7 @@ impl GatewayLdkClient {
         node: &ldk_node::Node,
         htlc_stream_sender: &Sender<InterceptPaymentRequest>,
         handle: &TaskHandle,
+        payment_subscriptions: &Arc<Mutex<PaymentSubscriptions>>,
     ) {
         // We manually check for task termination in case we receive a payment while the
         // task is shutting down. In that case, we want to finish the payment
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -269,8 +269,6 @@ impl ILnRpcClient for GatewayLdkClient {
         &self,
         _num_route_hints: usize,
     ) -> Result<GetRouteHintsResponse, LightningRpcError> {
-        // TODO: Return real route hints. Not strictly necessary but would be nice to
-        // have.
```
elsirion, line 273: Why remove? I think the LDK use-case will lend itself to adding route-hints: if your node only has one channel to an LSP you don't want to bother everyone by announcing it, you'd rather include it as a route hint.

tvolk131: `ILnRpcClient::routehints()` is only ever used for LNv1 payment receives. Specifically, the route hints are inserted into the gateway announcement, which can be used by clients when they generate invoices. The LDK gateway will never support LNv1 payment receives, it only supports LNv1 sends. LNv2 payment receives require that the gateway's lightning node generates invoices rather than the fedimint client, so it is able to insert the proper route hints on its own.

elsirion: Right, wasn't thinking about that!

tvolk131: I added this context to `GatewayLdkClient::routehints()` for future readers

m1sterc001guy:
Could you provide some more context here? The issue is the gateway going offline while an outgoing payment is happening? How do you propose we should test this?

m1sterc001guy:
I am confused if this is a simple refactor or if it is supposed to provide additional fault tolerance to LDK

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -54,6 +62,15 @@ pub struct GatewayLdkClient {
     /// The HTLC stream, until it is taken by calling
     /// `ILnRpcClient::route_htlcs`.
     htlc_stream_receiver_or: Option<tokio::sync::mpsc::Receiver<InterceptPaymentRequest>>,
+
+    /// The set of outgoing payments, by payment hash, that are currently
+    /// awaiting completion. This is used to allow `Self::handle_next_event()`
+    /// to notify `Self::pay()` when a payment completes.
+    payment_subscriptions: Arc<
+        Mutex<
+            HashMap<bitcoin30::hashes::sha256::Hash, tokio::sync::oneshot::Sender<PaymentOutcome>>,
```
m1sterc001guy, line 71: nit: I prefer `BTreeMap`

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -135,33 +160,74 @@ impl GatewayLdkClient {
...
         node: &ldk_node::Node,
         htlc_stream_sender: &Sender<InterceptPaymentRequest>,
+        payment_subscriptions: &Arc<
+            Mutex<
+                HashMap<
+                    bitcoin30::hashes::sha256::Hash,
+                    tokio::sync::oneshot::Sender<PaymentOutcome>,
+                >,
+            >,
+        >,
```
m1sterc001guy, line 177: This is ugly...can we make this a defined type?

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -269,8 +335,11 @@ impl ILnRpcClient for GatewayLdkClient {
         &self,
         _num_route_hints: usize,
     ) -> Result<GetRouteHintsResponse, LightningRpcError> {
-        // TODO: Return real route hints. Not strictly necessary but would be nice to
-        // have.
+        // `ILnRpcClient::routehints()` is currently only ever used for LNv1 payment
+        // receives and will likely be removed when we switch to LNv2. The LDK gateway
```
m1sterc001guy, line 339: nit: remove "likely"

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -282,7 +351,50 @@ impl ILnRpcClient for GatewayLdkClient {
...
+                        }))
+                    } else {
+                        None
+                    }
+                }
+                PaymentStatus::Failed => Some(Err(LightningRpcError::FailedPayment {
+                    failure_reason: "LDK payment failed".to_string(),
+                })),
+            }
+        } else if let Err(err) = self.node.bolt11_payment().send(
```
m1sterc001guy, line 397: I'm finding this section hard to read. Can we make this a `match` statement? I also think the nested match can perhaps be made into a separate function. It took me a few times reading this to understand what was going on.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -291,41 +403,30 @@ impl ILnRpcClient for GatewayLdkClient {
...
-                    }
-                    PaymentStatus::Failed => {
-                        return Err(LightningRpcError::FailedPayment {
-                            failure_reason: "LDK payment failed".to_string(),
-                        });
-                    }
-                }
-            }
-            fedimint_core::runtime::sleep(Duration::from_millis(100)).await;
+        if let Some(response) = response_or {
```
m1sterc001guy, line 413: Add comment here that this semantically means the payment has already been completed and we can just return the retrieved `PayInvoiceResponse`

m1sterc001guy:
Concept ACK but I would love a few comments and PR description. The reasoning behind doing this isn't obvious at first so it helps to have that context.

gateway/ln-gateway/src/lightning/ldk.rs
elsirion: It might also be worth making it a type def, given the length of the type

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Done

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Done and done

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Done

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -126,13 +150,22 @@ impl GatewayLdkClient {
...
         htlc_stream_sender: &Sender<InterceptPaymentRequest>,
         handle: &TaskHandle,
+        payment_subscriptions: &Arc<
+            Mutex<
+                HashMap<
+                    bitcoin::hashes::sha256::Hash,
+                    tokio::sync::oneshot::Sender<PaymentOutcome>,
+                >,
+            >,
+        >,
```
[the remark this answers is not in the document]
tvolk131: Done

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -290,7 +356,50 @@ impl ILnRpcClient for GatewayLdkClient {
...
+            .node
+            .list_payments_with_filter(|payment| {
+                if let PaymentKind::Bolt11 { hash, .. } = payment.kind {
+                    hash.0 == invoice.payment_hash().to_byte_array()
+                } else {
+                    false
+                }
+            })
+            .into_iter()
+            .next();
```
elsirion, line 380: This PR tries to achieve never paying an invoice twice, can we assert here that the iterator actually only had one item?

elsirion (commented):
Just some more comments, will leave it to @m1sterc001guy to make the final call.

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Good point! Done.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -135,33 +160,74 @@ impl GatewayLdkClient {
...
+            }
+            ldk_node::Event::PaymentFailed {
+                payment_hash: Some(payment_hash),
+                reason,
+                ..
+            } => {
+                if let Some(payment_outcome_sender) = payment_subscriptions.lock().await.remove(
+                    &bitcoin30::hashes::sha256::Hash::from_byte_array(payment_hash.0),
+                ) {
+                    let reason = reason.expect("Can only be `None` in ldk-node v0.2.1 and prior, which we've never used in a release");
```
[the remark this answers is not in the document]
tvolk131: Done

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Cleaned up and documented this section, and moved it into a new function called `pay_idempotent()`

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Done

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -290,7 +356,50 @@ impl ILnRpcClient for GatewayLdkClient {
         max_delay: u64,
         max_fee: Amount,
     ) -> Result<PayInvoiceResponse, LightningRpcError> {
-        let payment_id = match self.node.bolt11_payment().send(
+        // Subscribe to the `PaymentSuccessful`/`PaymentFailed` events for this invoice.
+        let (sender, receiver) = tokio::sync::oneshot::channel();
+        self.payment_subscriptions
+            .lock()
+            .await
+            .insert(*invoice.payment_hash(), sender);
```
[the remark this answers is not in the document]
tvolk131: You raise a good point: This should handle multiple concurrent calls while the payment is in-flight

I changed `PaymentSubscriptions` to map to a _list_ of oneshot senders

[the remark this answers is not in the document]
m1sterc001guy: Using a list here feels wrong. I think its just due to some implementation details, but let me know if I'm missing something more fundamental here. Can you do something like this:

1) Lock `payment_subscriptions` map
2) Check if the payment has already been made. You already do this in `pay_idempotent`
3) Only if it has not been made, then insert the sender into the map
4) Do payment as you do now

This would require some refactoring to pull out 2) from `pay_idempotent`, but I think it allows you to assert that there is only ever 1 thread waiting on a oneshot channel.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -44,6 +61,17 @@ pub struct GatewayLdkClient {
...
+    /// The set of outgoing payments, by payment hash, that are currently
+    /// awaiting completion. This is used to allow `Self::handle_next_event()`
+    /// to notify `Self::pay()` when a payment completes.
+    ///
+    /// Note: Since this field is needed to process `PaymentSuccessful` and
+    /// `PaymentFailed` events from the LDK node, locking it will block all
+    /// event processing, including `PaymentClaimable` events, which are needed
+    /// to process inbound federation payments. Do not lock this field for
+    /// long periods of time.
+    payment_subscriptions: Arc<Mutex<PaymentSubscriptions>>,
```
m1sterc001guy, line 74: Can we do better than a `Arc<Mutex<PaymentSubscriptions>>`? This essentially means that LDK can only process payments in sequence. What about a `Arc<BTreeMap<Mutex<Hash>, Sender>>`? This would only require you to lock each individual payment hash, which would allow for payments to process in parallel.

elsirion:
This PR is moving fairly slowly and might need more iterations, I think moving forward with 0.5 without it for now and then backporting is reasonable @bradleystachurski.

bradleystachurski:
> This PR is moving fairly slowly and might need more iterations, I think moving forward with 0.5 without it for now and then backporting is reasonable @bradleystachurski.

Sounds good, tracking in https://github.com/fedimint/fedimint/issues/6349

joschisan:
I think this is the right way to do it instead: https://github.com/fedimint/fedimint/pull/6352

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: This looks like it's essentially an async-native version of [`dashmap`](https://crates.io/crates/dashmap). Very cool! However, it's currently missing the ability to lock on a particular key regardless of whether there's already a value there. In `dashmap` we have [`DashMap::entry()`](https://docs.rs/dashmap/6.1.0/dashmap/struct.DashMap.html#method.entry). But in `whirlwind` all we could do is call [`ShardMap::insert()`](https://docs.rs/whirlwind/0.1.1/whirlwind/struct.ShardMap.html#method.insert) and hope that a previous value isn't returned, then call [`ShardMap::get()`](https://docs.rs/whirlwind/0.1.1/whirlwind/struct.ShardMap.html#method.get) to lock on the key.

tvolk131:
> Seems to me like this is the right way to do it instead: #6352

Part of the reason this PR is so complex is because we're also fixing this todo:

https://github.com/fedimint/fedimint/blob/5d84f93b3f3baff06f87e83eb4c8c11f828dd6c9/gateway/ln-gateway/src/lightning/ldk.rs#L310-L337

But perhaps this should be tackled separately

joschisan:
> Part of the reason this PR is so complex is because we're also fixing this todo

Why would you want to avoid looping / polling?

elsirion:
Closed in favor of https://github.com/fedimint/fedimint/pull/6369

PR #6231, opened 2024-10-27
fix: make `spend_ecash` not overspend by default

elsirion:
The default behavior of `spend_ecash` is a foot-gun: if no notes of the correct denominatons are available it will return the next bigger note without regard for the amount over-spent. This was ok for expert integrators, but is a massive foot-gun for people less familiar with the code base.

This PR makes `spend_ecash` fail if the exact amount cannot be provided and introduces a new `spend_ecash_expert` function that can be used by people relying on the old behavior.

This is mostly relevant to the Web SDK, but also the CLI. cc @maan2003 @alexlwn123

### DIFF
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -856,6 +856,17 @@ impl ClientModule for MintClientModule {
                         yield serde_json::to_value(state)?;
                     }
                 }
+                "spend_notes" => {
+                    let req: SpendNotesRequest = serde_json::from_value(request)?;
+                    let result = self.spend_notes_with_selector(
+                        &SelectNotesWithExactAmount,
+                        req.amount,
+                        req.try_cancel_after,
+                        req.include_invite,
+                        req.extra_meta
+                    ).await?;
+                    yield serde_json::to_value(result)?;
+                }
                 "spend_notes_expert" => {
                     let req: SpendNotesExpertRequest = serde_json::from_value(request)?;
                     let result = self.spend_notes(req.min_amount, req.try_cancel_after, req.include_invite, req.extra_meta).await?;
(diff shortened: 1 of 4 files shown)


maan2003:
> This PR makes spend_ecash fail if the exact amount cannot be provided and introduces a new spend_ecash_expert function that can be used by people relying on the old behavior.

but what would ui do in that case?

what is the path to resolve that issue.

elsirion:
> what is the path to resolve that error

A simple UI will not provide resolution, a fancy UI will try to come online and re-issue or use the expert mode to accept a certain degree of overspending. Ideally we'd have better support for creating exact-sized e-cash bags while online, but till we do I'd prefer integrators not paying attention to not shoot themselves in the foot.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -858,6 +858,17 @@ impl ClientModule for MintClientModule {
...
+                        req.amount,
+                        req.try_cancel_after,
+                        req.include_invite,
+                        req.extra_meta
+                    ).await?;
+                    yield serde_json::to_value(result)?;
+                }
+                "spend_notes_expert" => {
+                    let req: SpendNotesExpertRequest = serde_json::from_value(request)?;
                     let result = self.spend_notes(req.min_amount, req.try_cancel_after, req.include_invite, req.extra_meta).await?;
```
maan2003, line 872: should we change the default in self.spend_notes?

elsirion: good point!

elsirion: Thinking about it more, changing the behavior would be a bad move. Integrators might depend on the semantics. So instead we can a) remove it or b) deprecate it and remove it later. Deprecation is the right move for now imo.

maan2003, line 872: lets not call the deprecated function

elsirion: Should have run clippy :see_no_evil:

elsirion: All fixed

maan2003 (approved):

PR #6238, opened 2024-10-28
chore: remove duplicate CLN extension endpoint consts

tvolk131:
Looks like the CLN extension endpoint constants got mistakenly duplicated in #6197.


gateway/ln-gateway/src/bin/cln_extension.rs
```diff
@@ -29,20 +29,20 @@ use futures_util::stream::StreamExt;
 use hex::ToHex;
 use lightning_invoice::{Currency, InvoiceBuilder, PaymentSecret};
 use ln_gateway::envs::FM_CLN_EXTENSION_LISTEN_ADDRESS_ENV;
-use ln_gateway::lightning::extension::{
```
m1sterc001guy, line 32: Why the rename? I don't think these should be in `ln_gateway::rpc`, since this is a Lightning specific thing, not an "rpc" concept (which up until now we have used only for the HTTP webserver)

gateway/ln-gateway/src/rpc/extension_endpoints.rs
```diff
@@ -1,13 +0,0 @@
-pub const CLN_INFO_ENDPOINT: &str = "/cln_info";
```
m1sterc001guy, line 1: Thanks, this was a bad rebase I think

elsirion (approved):

m1sterc001guy (changes requested):
I would prefer to keep these endpoints in the `lightning` module since they should be only relevant to CLN

m1sterc001guy (approved):

PR #6247, opened 2024-10-29
feat: LNv2 allow user to verify payment parameters before paying

m1sterc001guy:
Currently when paying an LNv2 invoice, there are a number of parameters that the user does not have a chance to verify, such as the CLTV delta and fees that are paid to the gateway. This PR refactors LNv2's API and allows the CLI (and other integrators) to display relevant details to the user before executing the payment.

image

### DIFF
--- a/modules/fedimint-lnv2-client/src/cli.rs
+++ b/modules/fedimint-lnv2-client/src/cli.rs
@@ -20,6 +20,8 @@ enum Opts {
         invoice: Bolt11Invoice,
         #[arg(long)]
         gateway: Option<SafeUrl>,
+        #[clap(long, default_value_t = false)]
+        force: bool,
     },
     /// Await the final state of the send operation.
     AwaitSend { operation_id: OperationId },
@@ -63,7 +65,11 @@ pub(crate) async fn handle_cli_command(
     let opts = Opts::parse_from(iter::once(&ffi::OsString::from("lnv2")).chain(args.iter()));
 
     let value = match opts {
-        Opts::Send { gateway, invoice } => {
+        Opts::Send {
+            gateway,
+            invoice,
+            force,
+        } => {
             let (gateway_api, routing_info) = match gateway {
                 Some(gateway_api) => (
                     gateway_api.clone(),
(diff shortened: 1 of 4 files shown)


joschisan (changes requested):
Will need to look closer, but on first looks its a concept NACK. The client enforces sensible limits, and since the gateways are vetted otherwise just relies on the gateways recommendations for expiration delta and lightning fee. Users cannot decide that better than a vetted gateways which has access to the channel graph and routing history.

m1sterc001guy:
> Users cannot decide that better than a vetted gateway operator which has access to the channel graph and routing history.

Users should ultimately be the ones who are in control of their money. Currently, the user isn't even informed what fees they are paying, its just taken from their balance. You'd need to actually look at the code and dissect the constants to figure out what you're actually being charged. This is a horrible user experience, money just starts disappearing and users don't know why. 

This PR doesn't even change any of the sensible limits, it just displays them to the user and allows them to decide if they still want to do the payment.

Ultimately I think users should also decide what limits are sensible for them. We can give them a reasonable default, but I don't think we should assume that everyone has the same tolerance to fees.

elsirion:
We definitely need a way to show fees to users ahead of time. LNv1 allows that by making the fee announcement part of the gateway struct that you'll have to supply to the pay function, so integrators can do:

1. Call select gateway fn
2. Use fee information from that to calculate fee to show to users
3. Call pay fn with selected gateway if user accepts

I don't mind *how* exactly it's being implemented, but fees have to be known.

joschisan:
I am sorry there are multiple issues with this pr, and with the concept in general. Since there's so much to say on this I would propose a call with @elsirion and @m1sterc001guy to elaborate.

m1sterc001guy:
@joschisan and I had a chat about this today. I will try to summarize here.

I think there are two tensions here:

1) Keeping LNv2's API simple and easy to integrate
2) Whether or not the CLI should show these parameters.

I think regarding 1) we are mostly on the same page. The proposed API changes make it a bit more difficult to integrate, since now there is a `RoutingInfo` struct that is necessary. One option would be to change it so it takes a `Option<RoutingInfo>`, which would essentially just short-circuit the REST call to the gateway. I am open to feedback here.

On the second point I think we have a more fundamental disagreement. @joschisan sounded open to allowing app integrators display the fee and routing information but prefers to wait to hear from app integrators to see if that's what they desire before opening the API. Personally, I think `fedimint-cli` is a valid use case for using this API. I think the CLI should be a full-fledged "expert" wallet (similar to `lncli` or `lightning-cli` in the Lightning world) and not just a toy that is used for testing and in CI.

I am interested to hear others' opinions. I will change this to a draft for now.

joschisan:
Also, the cli right now is a close reflection of the module api such that an integrator can use it to get familiar with the module api before integration - this was a point mentioned by @maan2003 when reviewing the modules api. For now at least this is a property I would like to maintain.

m1sterc001guy:
If the module does not expose a way for any integrator (CLI or any other app) to compute fees and show them to the user, this is essentially an endorsement that they do not need to be shown at all. I do not mind making the CLI a reflection of the module API, but on this I simply disagree. I believe users should always be shown how their money is being used and have the option to continue or not.

maan2003:
> Also, the cli right now is a close reflection of the module api such that an integrator can use it to get familiar with the module api before integration - this was a property brought to my attention by @maan2003 when reviewing the modules api for integration

I don't recall my exact words. But, I use cli **code** (not cli itself) to see how to use the module api. This means cli should exercise all public functions of a module client. The cli interface doesn't have to be exactly same as rust api.

elsirion:
Last Thursday on the deep dive we agreed to close the PR for now and revisit the issue eventually. For now integrators can show:
* An upper fee limit enforced by the client
* An upper fee limit estimated by querying all the gateways upfront and taking the max fee from them (racy, but likely good enough unless a GW is actively malicious)
* Display the exact fee paid after the payment succeeded

PR #6249, opened 2024-10-29
feat: allow clients to manually add a gateway

m1sterc001guy:
LNv2 takes a different approach to gateway registration. Guardians need to manually "register" gateways. This allows the guardians to directly vet who provides liquidity to their federation via lightning, which allows for a tightly controlled user experience.

This, however, does put additional onus of the guardians to maintain a performant and highly available set of gateways. If guardians are not super responsive and there are no gateways available for payments....then LN payments just won't work.

This PR proposes a backup method that allows a user client to manually add a gateway to their cache. This way, if guardians cannot (or refuse) to update the list of registered lists, clients can manually do it. The connection info for a gateway (public key, url) can be shared via Nostr or another out of band medium.


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -467,13 +467,67 @@ impl LightningClientModule {
...
+    }
+
     async fn select_gateway(
         &self,
         invoice: Option<Bolt11Invoice>,
     ) -> Result<(SafeUrl, RoutingInfo), SelectGatewayError> {
         let gateways = self
             .module_api
-            .gateways()
+            .get_registered_gateways()
```
elsirion, line 530: Do we even use the DB gateways right now? Seems to me that we'd need to add them here somehow.

elsirion (approved):
I like the direction, but @joschisan will likely have an opinion on the exact semantics.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -467,13 +467,67 @@ impl LightningClientModule {
...
+                if gateways.contains(&url) {
+                    None
+                } else {
+                    Some(url)
+                }
+            })
+            .collect::<Vec<_>>()
+            .await;
+
+        gateways.append(&mut gateways_in_db);
```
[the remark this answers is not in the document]
m1sterc001guy: Fixed!

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: Integrators need to call `update_gateway_map` for the db to be updated currently. The CLI does not do this, so the CLI doesn't use the DB at all. I think this PR helps the CLI too

m1sterc001guy:
Will wait for @joschisan approval

m1sterc001guy:
Chatted offline with @joschisan and agreed to close this PR. If application integrators need the ability to directly manipulate the cache, that is either something they can do on their own or we can revisit in the future.

The cache is not currently being used in the CLI, we agreed it would be good to modify the API so it is able to be tested in the CLI.

PR #6251, opened 2024-10-30
fedimint client rpc ng

maan2003:
A json only protocol, no objects like in previous rpc.
This will help for new react native and stdio transports.

extracted a new crate that is independent of wasm.

moves more multi federation handling into the rpc backend.

overlay architecture:

image


fedimint-client-rpc/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+            }
+            fedimint_core::task::sleep(Duration::from_millis(100)).await;
+        }
+        Ok(())
+    }
+
+    fn handle_client_rpc(
+        self: Arc<Self>,
+        client_name: String,
+        module: String,
```
dpc, line 188: `Option<String>`? `""` feels icky. :D

dpc (commented):
Cool. One q.

fedimint-client-rpc/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+    End {},
+}
+
+pub trait RpcResponseHandler: MaybeSend + MaybeSync {
+    fn handle_response(&self, response: RpcResponse);
+}
+
+#[apply(async_trait_maybe_send!)]
+pub trait DatabaseFactory: MaybeSend + MaybeSync + 'static {
+    async fn create_database(&self, name: &str) -> anyhow::Result<Database>;
```
dpc, line 77: Is this "open"? I mean - it is expected to open previous database if one was used before under same name?

fedimint-client-rpc/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+            tracing::error!("RPC CLIENT ERROR: request id reuse detected");
+        }
+    }
+
+    fn remove_rpc_handle(&self, request_id: u64) -> Option<AbortHandle> {
+        let mut handles = self.rpc_handles.lock().unwrap();
+        handles.remove(&request_id)
+    }
+
+    async fn handle_join_federation(
```
dpc, line 121: I see, so the whole thing is stateful and multi-federation, so it keeps track of which federations were joined etc.?

maan2003: it doesn't keep track of the list, you need to keep track of list of joined clients somewhere else

but yes it is stateful

fedimint-client-rpc/src/lib.rs
maan2003: oh right this is open or create

fedimint-client-rpc/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+use futures::future::{AbortHandle, Abortable};
+use futures::StreamExt;
+use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
+use tracing::info;
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct RpcRequest {
+    pub request_id: u64,
```
dpc, line 26: PoI: describe why we have ids.

fedimint-client-rpc/src/lib.rs
maan2003: good idea

dpc (commented):
Huh. I see.

maan2003:
discussion from call: disallow joining same federation multiple times

fedimint-client-rpc/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+                        yield item?;
+                    }
+                }
+                "mint" => {
+                    let mint = client.get_first_module::<MintClientModule>()?.inner();
+                    let mut stream = mint.handle_rpc(method, payload).await;
+                    while let Some(item) = stream.next().await {
+                        yield item?;
+                    }
+                }
```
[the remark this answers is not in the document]
dpc: We should just scan list of modules and match the `kind`, there's no need to downcast them to their actual type - `IClientModule` can handle methods just fine.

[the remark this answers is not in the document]
dpc: BTW. Just like with the CLI, we should support `module` being both "kind" (meaning first one of that kind), and a number (`module_instance_id`). See `fedimint-cli module ...` handling.

dpc:
*Comment by Tau/GPT-5*

Triaging this stale draft: I think this can be closed as superseded. The main work here (extracting `fedimint-client-rpc` and wiring the WASM client through the JSON RPC layer) was later rebased/updated and merged in #7499, with follow-up client-rpc improvements merged since then. This PR now has conflicts and targets old APIs/structure (e.g. the old `DatabaseFactory`/IndexedDB split), while current `master` has `fedimint-client-rpc` integrated around the newer unified DB / `fedimint-cursed-redb` setup.

The only distinct piece I noticed is the experimental `fedimint-client-stdio` binary, which was not merged. If that is still wanted, it would be better as a new focused PR against the current `fedimint-client-rpc` API rather than rebasing this whole draft.

Closing to keep the PR queue tidy; this can always be reopened if I missed something.

PR #6253, opened 2024-10-30
fix: remove leading and trailing newlines from password

dpc:
If the file was created/modified manually, the text editor will likely add an extra newline.

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -94,6 +94,12 @@ impl ConfigGenApi {
     // Sets the auth and decryption key derived from the password
     pub async fn set_password(&self, auth: ApiAuth) -> ApiResult<()> {
         let mut state = self.require_status(ServerStatus::AwaitingPassword).await?;
+        let auth_trimmed = auth.0.trim();
+        if auth_trimmed != auth.0 {
+            return Err(ApiError::bad_request(
+                "Password contains leading/trailing whitespace".to_string(),
+            ));
+        }
         state.auth = Some(auth);
         state.status = ServerStatus::SharingConfigGenParams;
         info!(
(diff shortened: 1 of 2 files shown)


douglaz:
What if there is actually a space on the password? Perhaps remove only new lines?

dpc:
> What if there is actually a space on the password?

[Illegal](https://www.youtube.com/watch?v=eiyfwZVAzGw).


Note: `trim()` removes only leading and trailing whitespace, so spaces in between are fine.

douglaz:
Spaces are perfectly legal and reasonable in any part of the password. You have no right of enforcing your weak and authoritarian standards on the common people.

dpc:
Not only we have a right, but a moral duty to forbid leading and trailing whitespace in the passwords. *Salus populi suprema lex esto*

Leading and trailing whitespace is a sure way to make sure the user screws up copy&paste, sets a password with characters they are not aware of and then blames us that nothing works.

The passwords that Fedimint IO generates and suggests are 16-character long alpanumerical characters, and our flow heavily suggests copy&paste and backing up the password.

IroncladDev:
image

elsirion:
I wonder if we should at least emit a warning when `trim` did anything. Could be unintuitive to debug otherwise. Or just try the trimmed and the original version when trying to decrypt our secrets to maintain compatibility :laughing: 

Longer term I want to get rid of `password.secret` anyway #6258

dpc:
BTW. I do put spaces in passwords (well, passphrases) in between the words. That's fine. Messing with user passwords is a bad idea. Just the leading and trailing whitespace are 99% user error.

dpc:
OK, so to make it more digestable I changed it to:

* remove online leading/trailing newlines from the existing password
* `warn!` about existing leading/trailing whitespace (other than newlines) in the password file
* reject leading/trailing whitespace on `set_password` api request

elsirion (approved):

fedimint-backports:
- #6278

PR #6264, opened 2024-11-01
chore: allow state machine gen to know in/out_idx in a bundle

dpc:


fedimint-client/src/transaction/builder.rs
```diff
@@ -420,14 +450,60 @@ impl TransactionBuilder {
...
+            .flat_map(|(bundle_idx, bundle)| {
+                let input_idxs = input_idx_to_bundle_idx
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(input_idx, input_bundle_idx)| {
+                        (*input_bundle_idx == bundle_idx).then_some(input_idx as u64)
+                    })
+                    .fold(None, |cur: Option<(u64, u64)>, idx| {
+                        Some(cur.map_or((idx, idx), |cur| (cur.0.min(idx), cur.1.max(idx))))
+                    });
```
elsirion, line 466: There has to be a better way to recover input index ranges :laughing: This code was fairly hard to understand unfortunately. Maybe accepting mutation when adding inputs is the price we have to pay for not having to reconstruct the ranges like this.

Also, the algorithm scales quadratically with the number of inputs/bundles (likely not a real problem, but meh).

gateway/ln-gateway/src/gateway_module_v2/mod.rs
```diff
@@ -381,27 +381,32 @@ impl GatewayClientModuleV2 {
...
-                            incoming_chan_id,
-                            htlc_id,
-                        },
-                        state: CompleteSMState::Pending,
-                    }),
-                ]
+            state_machines: Arc::new(move |txid, out_idxs: std::ops::RangeInclusive<u64>| {
+                out_idxs
+                    .clone()
+                    .flat_map(|out_idx| {
```
elsirion, line 387: Should we assert that there's only ever one index in the range?

gateway/ln-gateway/src/gateway_module_v2/mod.rs
```diff
@@ -440,16 +445,21 @@ impl GatewayClientModuleV2 {
...
-                        contract: contract.clone(),
-                        out_point: OutPoint { txid, out_idx },
-                        refund_keypair: bitcoin30_to_bitcoin32_keypair(&refund_keypair),
-                    },
-                    state: ReceiveSMState::Funding,
-                })]
+            state_machines: Arc::new(move |txid, out_idxs| {
+                out_idxs
+                    .clone()
+                    .flat_map(|out_idx| {
```
elsirion, line 451: Same assert question here

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -2345,6 +2337,47 @@ pub fn represent_amount<K>(
...
+                vec![MintClientStateMachines::Input(MintInputStateMachine {
+                    common: MintInputCommon {
+                        operation_id,
+                        txid,
+                        input_idx,
+                    },
+                    state: MintInputStates::Created(input_state),
+                })]
+            })
+            .collect()
```
elsirion: `flat_map` + single element vec = NOP

modules/fedimint-mint-client/src/oob.rs
```diff
@@ -322,35 +322,35 @@ async fn try_cancel_oob_spend_multi(
...
+                .into_iter()
+                .flat_map(|input_idx| {
+                    vec![MintClientStateMachines::Input(MintInputStateMachine {
+                        common: MintInputCommon {
+                            operation_id,
+                            txid,
+                            input_idx,
+                        },
+                        state: MintInputStates::Refund(MintInputStateRefund { refund_txid: txid }),
+                    })]
```
elsirion: Same here

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -584,13 +585,18 @@ impl WalletClientModule {
...
-                operation_id,
-                state: WithdrawStates::Created(CreatedWithdrawState {
-                    fm_outpoint: OutPoint { txid, out_idx },
-                }),
-            })]
+        let sm_gen = move |txid, out_idxs: RangeInclusive<u64>| {
+            out_idxs
+                .clone()
+                .flat_map(|out_idx| {
+                    vec![WalletClientStates::Withdraw(WithdrawStateMachine {
```
elsirion, line 592: And here

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -610,13 +616,18 @@ impl WalletClientModule {
...
```
elsirion, line 623: and here

modules/fedimint-mint-client/src/input.rs
```diff
@@ -161,6 +176,104 @@ impl MintInputStateCreated {
...
+        MintInputStateMachine {
+            common: old_state.common,
+            state: match refund_txids.len() {
+                0 => panic!("Can't have nothing here"),
+                1 => MintInputStates::Refund(MintInputStateRefund {
+                    refund_txid: refund_txids[0],
+                }),
+                _ => MintInputStates::RefundMulti(MintInputStateRefundMulti { refund_txids }),
+            },
+        }
```
elsirion, line 286: Why not always make it a refund-multi from now on? We could even migrate old states into their multi equivalents this way, possibly avoiding having to maintain two versions of the update stream code.

elsirion (commented):
I wonder how much this breaks update streams, do we test them?

fedimint-client/src/transaction/builder.rs
dpc: I don't find this code offensive. Sure, it's well... immediately obvious, but I can add some comments, and the general model there is actually quite simple.

dpc: I actually found an elegant refactoring that makes it easier to look at.

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1083,28 +1083,22 @@ impl MintClientModule {
...
+            assert_eq!(out_idxs.clone().count(), issuance_requests.len());
+            vec![MintClientStateMachines::Output(MintOutputStateMachine {
+                common: MintOutputCommon {
+                    operation_id,
+                    txid,
+                    out_idxs: out_idxs.clone(),
+                },
+                state: MintOutputStates::CreatedMulti(MintOutputStatesCreatedMulti {
+                    issuance_requests: out_idxs.zip(issuance_requests.clone()).collect(),
+                }),
```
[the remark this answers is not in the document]
dpc: Hmmm... here it will just be easier to iterate over these, if it doesn't need to be zipped everywhere. No strong opinion.

modules/fedimint-mint-client/src/input.rs
```diff
@@ -250,7 +258,112 @@ impl MintInputStateCreatedMulti {
...
+                // if the bundle had 1 note, we did per-note refund already
+                MintInputStates::RefundedPerNote(MintInputStateRefundedPerNote {
+                    refund_txids: vec![refund_txid],
+                })
+            } else {
+                MintInputStates::RefundedBundle(MintInputStateRefundedBundle {
+                    refund_txid,
+                    spendable_notes,
+                })
+            },
```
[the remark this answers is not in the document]
dpc: @elsirion 

I don't get it.

We're in `MintInputStateCreatedBundle::refund` which means a general purpose transaction was submitted and it failed. Now we attempted to claim all the mint inputs from the bundle (code slightly above). If there was many of them, we should move to `RefundedBundle` which will attempt per-note refunds. But if the bundle had one input, there's no point in doing `RefundedBundle`, and we can just finish (transition to `RefundedPerNote` terminal state).

modules/fedimint-mint-client/src/output.rs
```diff
@@ -290,6 +295,205 @@ impl MintOutputStatesCreated {
...
+            let spendable_note = issuance_request.finalize(agg_blind_signature);
+
+            assert!(spendable_note.note().verify(*amount_key));
+
+            debug!(target: LOG_CLIENT_MODULE_MINT, amount = %amount, note=%spendable_note, "Adding new note from transaction output");
+
+            client_ctx
+                .log_event(
+                    &mut dbtx.module_tx(),
+                    NoteCreated {
```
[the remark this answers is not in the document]
dpc: I don't know. Doesn't seem all that useful to the external consumers,  but we can always add it later.

modules/fedimint-mint-client/src/output.rs
```diff
@@ -71,12 +71,15 @@ pub enum MintOutputStates {
...
+    CreatedMulti(MintOutputStatesCreatedMulti),
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Decodable, Encodable)]
 pub struct MintOutputCommon {
     pub(crate) operation_id: OperationId,
-    pub(crate) out_point: OutPoint,
+    pub(crate) txid: TransactionId,
+    pub(crate) out_idxs: std::ops::RangeInclusive<u64>,
```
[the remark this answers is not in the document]
dpc: @elsirion Another PR. :pray:  :D

modules/fedimint-mint-client/src/input.rs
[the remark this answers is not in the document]
elsirion: I was confused, I read your comment to mean "if there's only one note, then we split the previously failed refund tx already", but what you meant is "if there is one note, we don't need to split"

modules/fedimint-mint-client/src/lib.rs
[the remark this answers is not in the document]
elsirion: It's redundancy that could diverge, but ok to go ahead as is.

elsirion (approved):
Some comments were on commits accidentally pushed here, but belonging to another PR. LGTM.

PR #6265, opened 2024-11-01
feat: enable relative fees for the mint module

joschisan:


fedimint-client/src/module/mod.rs
```diff
@@ -805,7 +809,7 @@ pub trait IClientModule: Debug {
 
     fn input_fee(&self, amount: Amount, input: &DynInput) -> Option<Amount>;
 
-    fn output_fee(&self, output: &DynOutput) -> Option<Amount>;
+    fn output_fee(&self, amount: Amount, output: &DynOutput) -> Option<Amount>;
```
elsirion, line 812: I see we have the amount in `input_fee` too (I think because we already removed some data from the DB when this fn is called, so might not always know), but why can't the client figure out itself how much an output is worth?

fedimint-core/src/fee_consensus.rs
```diff
@@ -19,7 +19,7 @@ impl FeeConsensus {
     /// # Errors
     /// - This constructor returns an error if the relative fee is in excess of
     ///   one thousand parts per million.
-    pub fn new(parts_per_million: u64) -> anyhow::Result<Self> {
+    pub fn new_lnv2(parts_per_million: u64) -> anyhow::Result<Self> {
```
elsirion, line 22: I am **very** opposed to having any module-specific code in `fedimint-core`.

fedimint-core/src/fee_consensus.rs
```diff
@@ -31,6 +31,35 @@ impl FeeConsensus {
         })
     }
 
+    /// The mint module will charge a non-configurable base fee of one hundred
+    /// millisatoshis per transaction input and output too account for the costs
```
elsirion, line 35: `s/too/to/`

fedimintd/src/fedimintd.rs
```diff
@@ -327,7 +327,9 @@ impl Fedimintd {
                     local: EmptyGenParams::default(),
                     consensus: MintGenParamsConsensus::new(
                         2,
-                        fedimint_mint_server::common::config::FeeConsensus::default(),
+                        // we need to wait for clients to support the relative fees before we can
```
elsirion, line 330: Worth a TODO comment?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1097,8 +1103,6 @@ impl MintClientModule {
             }
         }
 
-        assert!(!outputs.is_empty());
```
elsirion, line 1100: I guess we want to support zero-amounts now?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -2306,25 +2320,28 @@ pub fn represent_amount<K>(
     for tier in tiers.tiers() {
         let notes = current_denominations.get(*tier);
         let missing_notes = u64::from(denomination_sets).saturating_sub(notes as u64);
-        let possible_notes = remaining_amount / *tier;
+        let possible_notes = remaining_amount / (*tier + fee_consensus.fee(*tier));
 
         let add_notes = min(possible_notes, missing_notes);
         denominations.inc(*tier, add_notes as usize);
-        remaining_amount -= *tier * add_notes;
+        remaining_amount -= (*tier + fee_consensus.fee(*tier)) * add_notes;
```
elsirion, line 2327: `represent_amount` can be used for inputs or outputs right now, both require opposite handling of fees (adding vs subtracting). 

Maybe the `fee_consensus` argument can be made optional?

fedimint-server/src/consensus/transaction.rs
```diff
@@ -97,7 +97,7 @@ impl FundingVerifier {
     }
 
     pub fn verify_funding(self) -> Result<(), TransactionError> {
-        if self.input_amount == (self.output_amount + self.fee_amount) {
+        if self.input_amount >= (self.output_amount + self.fee_amount) {
```
elsirion, line 100: Technically a consensus break.

fedimint-client/src/module/mod.rs
joschisan: True, not all inputs contain their amount but all outputs do. It was just a reflex on my part to mirror the input_fee. I would leave it like that for now and at some point remove the input_fee and output_fee altogether in favor of adding a fee field to ClientInput and ClientOutput just as we did with the amount previously. Already had started to work on the refactor and it was significantly cleaner but good stuck on the lnv1 fees as they are so weird.

fedimint-core/src/fee_consensus.rs
joschisan: I agree. I am not too happy with how this has turned out either. I would keep it like that for now - most important thing here is to make sure that the consensus fee calculation is correct. Eventually I want to refactor this though.

fedimint-core/src/fee_consensus.rs
joschisan: what?

fedimintd/src/fedimintd.rs
joschisan: Done.

fedimint-server/src/consensus/transaction.rs
joschisan: The second commit was only for testing and has been removed.

modules/fedimint-mint-client/src/lib.rs
joschisan: What do you mean? This has been removed as due to fees, a small amount, like less then 100 millisats, don create new outputs.

modules/fedimint-mint-client/src/lib.rs
joschisan: How is it used for inputs? I only see it used for outputs and in the tests.

fedimint-core/src/fee_consensus.rs
elsirion: Sorry, that's `sed`/`vim` syntax for replacing `too` with `to`. I think that would be the grammatically correct way to do it.

modules/fedimint-mint-client/src/lib.rs
elsirion: I got confused, I think in the beginning there was some overlap, but ever since we target specific denominations that's not possible anymore anyway. I should rename the fn.

modules/fedimint-mint-client/src/lib.rs
elsirion: That was an interaction between the functions I missed, glad I asked :laughing:

elsirion (dismissed):
Thx for the clarifications! Feel free to fix the one typo or merge :)

fedimint-core/src/fee_consensus.rs
joschisan: Done

elsirion (approved):

modules/fedimint-mint-server/src/lib.rs
```diff
@@ -734,7 +740,7 @@ mod test {
                     .unwrap()
                     .consensus
                     .peer_tbs_pks,
-                fee_consensus: FeeConsensus::default(),
+                fee_consensus: FeeConsensus::new_lnv2(1000).expect("Relative fee is within range"),
```
bradleystachurski, line 743: Should this be `new_mint` instead of `new_lnv2`? Seems out of place to call other module logic from within the mint server.

PR #6267, opened 2024-11-01
chore: improve documentation of fees

joschisan:
We remove the parameters from PaymentFeeExceedsLimit, ExpirationDeltaExceedsLimit and PaymentFeeExceedsLimit as they are not actionable, with the current implementation that is just considered invalid by the client and three no mechanism to accept them anyway.

### DIFF
--- a/modules/fedimint-lnv2-client/src/lib.rs
+++ b/modules/fedimint-lnv2-client/src/lib.rs
@@ -279,14 +279,14 @@ impl PaymentFee {
     /// This is the maximum send fee of one and a half percent plus one hundred
     /// satoshis a correct gateway may recommend as a default. It accounts for
     /// the fee required to reliably route this payment over lightning.
-    pub const SEND_FEE_LIMIT_DEFAULT: PaymentFee = PaymentFee {
+    pub const SEND_FEE_LIMIT: PaymentFee = PaymentFee {
         base: Amount::from_sats(100),
         parts_per_million: 15_000,
     };
 
     /// This is the maximum receive fee of half of one percent plus fifty
     /// satoshis a correct gateway may recommend as a default.
-    pub const RECEIVE_FEE_LIMIT_DEFAULT: PaymentFee = PaymentFee {
+    pub const RECEIVE_FEE_LIMIT: PaymentFee = PaymentFee {
         base: Amount::from_sats(50),
         parts_per_million: 5_000,
     };
@@ -515,8 +515,21 @@ impl LightningClientModule {
             .await
     }
 
-    /// Pay an invoice. For  testing  you can optionally specify a gateway to
-    /// route with, otherwise a gateway will be selected automatically.
+    /// Pay an invoice. For testing you can optionally specify a gateway to
+    /// route with, otherwise a gateway will be selected automatically. If the
+    /// invoice was created by a gateway connected to our federation, the same
+    /// gateway will be selected to allow for a direct ecash swap. Otherwise we
+    /// select a random online gateway.
+    ///
+    /// The fee for this payment may depend on the selected gateway but
+    /// will be limited to one and a half percent plus one hundred satoshis.
+    /// This fee accounts for the fee charged by the gateway as well as
+    /// the additional fee required to reliably route this payment over
+    /// lightning if necessary. Since the gateway has been vetted by at least
+    /// one guardian we trust it to set a reasonable fee.
+    ///
+    /// The absolute fee for a payment can be calculated from the operation meta
+    /// to be shown to the user in the transaction history.
     pub async fn send(
         &self,
         invoice: Bolt11Invoice,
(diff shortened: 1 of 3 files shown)


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -562,14 +574,12 @@ impl LightningClientModule {
 
         let (send_fee, expiration_delta) = routing_info.send_parameters(&invoice);
 
-        if !send_fee.le(&PaymentFee::SEND_FEE_LIMIT_DEFAULT) {
-            return Err(SendPaymentError::PaymentFeeExceedsLimit(send_fee));
+        if !send_fee.le(&PaymentFee::SEND_FEE_LIMIT) {
```
m1sterc001guy, line 577: As long as we're touching this, can we change this to use `gt` instead of "not less than or equal to"?

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -515,8 +515,20 @@ impl LightningClientModule {
...
+    /// route with, otherwise a gateway will be selected automatically. If the
+    /// invoice was created by a gateway connected to our federation this
+    /// gateway will be selected, otherwise we select a random online gateway.
+    ///
+    /// The fee for this payment may depend on the selected gateway but
+    /// will be limited to one and a half percent plus one hundred satoshis.
+    /// This fee accounts for the fee charged by the gateway as well as
+    /// the additional fee required to reliably route this payment over
+    /// lightning if necessary. Since the gateway has been vetted by at least
+    /// one guardian we trust it to set a reasonable fee.
```
m1sterc001guy, line 528: Does this mean that you expect gateways to not change fees after being vetted?

joschisan: No, but it does mean that guardians check up on the gateways fees (by using the federation themselves for example) according to how much they trust them. If they start out with our gateway for example, or have a gateway that had good service and fees for months, the check ups are less frequent.

modules/fedimint-lnv2-client/src/lib.rs
joschisan: The statements are are not necessarily identical in a partial ordering.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -278,15 +278,15 @@ pub struct PaymentFee {
 impl PaymentFee {
     /// This is the maximum send fee of one and a half percent plus one hundred
     /// satoshis a correct gateway may recommend as a default. It accounts for
-    /// the fee required to reliably route this payment over lightning.
-    pub const SEND_FEE_LIMIT_DEFAULT: PaymentFee = PaymentFee {
+    /// the fee required to reliably route this payment over lightningy.
```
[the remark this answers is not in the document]
joschisan: Done.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -515,8 +515,20 @@ impl LightningClientModule {
             .await
     }
 
-    /// Pay an invoice. For  testing  you can optionally specify a gateway to
-    /// route with, otherwise a gateway will be selected automatically.
+    /// Pay an invoice. For  testing you can optionally specify a gateway to
```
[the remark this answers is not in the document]
joschisan: Done.

modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -515,8 +515,20 @@ impl LightningClientModule {
             .await
     }
 
-    /// Pay an invoice. For  testing  you can optionally specify a gateway to
-    /// route with, otherwise a gateway will be selected automatically.
+    /// Pay an invoice. For testing you can optionally specify a gateway to
```
tvolk131, line 518: Please change "If the invoice was created by a gateway connected to our federation this gateway will be selected, otherwise we select a random online gateway."

to:

"If the invoice was created by a gateway connected to our federation, the same gateway will be selected to allow for a direct ecash swap. Otherwise we select a random online gateway."

tvolk131 (changes requested):
Just one more nit, otherwise LGTM!

Also, can you describe in the PR description why we are removing parameters from `PaymentFeeExceedsLimit`, `ExpirationDeltaExceedsLimit` and `PaymentFeeExceedsLimit`? It's not clear to me from looking at the code.

modules/fedimint-lnv2-client/src/lib.rs
joschisan: Done.

tvolk131 (approved):

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: Does partial ordering matter here? If it does we should really have a comment, it is not obvious to me in what situations that would be relevant. In general it is more readable if there are less negations

modules/fedimint-lnv2-client/src/lib.rs
m1sterc001guy: I just don't think in practice most guardians are going to be consistently verifying fees and other routing parameters. Clients can always route around vetted gateways by just providing the URL endpoint to `send`, and this may not have been vetted by the guardians, so the comment doesn't seem 100% applicable to me.

PR #6269, opened 2024-11-01
chore: cleanup gateway complete state machine

tvolk131:
No behavior change, just un-nesting and de-duplicating code where possible

### DIFF
--- a/gateway/ln-gateway/src/state_machine/complete.rs
+++ b/gateway/ln-gateway/src/state_machine/complete.rs
@@ -129,22 +129,24 @@ impl WaitForPreimageState {
         let mut stream = context.notifier.subscribe(common.operation_id).await;
         loop {
             debug!("Waiting for preimage for {common:?}");
-            if let Some(GatewayClientStateMachines::Receive(state)) = stream.next().await {
-                match state.state {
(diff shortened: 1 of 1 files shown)


gateway/ln-gateway/src/state_machine/complete.rs
```diff
@@ -199,56 +201,46 @@ impl CompleteHtlcState {
...
-
-                    lightning_context
-                        .lnrpc
-                        .complete_htlc(htlc)
-                        .await
-                        .map_err(|_| CompleteHtlcError::FailedToCompleteHtlc)?;
-                    return Ok(());
-                }
+        // Wait until the lightning node is online to complete the HTLC.
+        let lightning_context = loop {
```
m1sterc001guy, line 207: I don't think these are the same semantics as the deleted code. Before, if `complete_htlc` failed we would try again, now it only loops when getting the lightning context

tvolk131: We were looping only on lightning context previously, it was just a lot less clear by looking. We were using the question mark operator before, so `complete_htlc` actually was returning an error, not retrying.

m1sterc001guy (approved):

PR #6259, opened 2024-10-31
test: Gateway liquidity test (ecash, onchain, lightning)

m1sterc001guy:
The gateway has 3 different ways to interact with Bitcoin: ecash, onchain, or lightning payments. We have built up APIs for depositing/withdrawing onchain, spending/receiving ecash, and opening/closing channels, but a lot of these APIs are still not well covered in our test framework. This PR adds a dedicated test to "other" payments (i.e gateway payments that do not involve a fedimint client)

Covered:
 - Lightning payments (moved from lnv2 test suite)
 - OOB Ecash payments
 - Onchain pegin ecash
 - Onchain pegout ecash
 - Channel closure
 - Withdraw onchain funds from lightning wallet
 - Deposit onchain is implicitly tested in devimint startup

PR #6270, opened 2024-11-01
refactor: Gateway CLI improvements

m1sterc001guy:
I have been confused more than once which command is "pegin" and which one just directly interacts with the onchain wallet. This PR fixes this by refactoring the CLI so that there are separate "onchain" and "ecash" subcommands.

This PR also removes the `balance` command and endpoint. There is a better `get_balances` endpoint that should be used instead.

Also renames `withdraw_onchain` to `send_onchain`


m1sterc001guy:
Prefer to merge https://github.com/fedimint/fedimint/pull/6259 first, since they are going to conflict

gateway/cli/src/general_commands.rs
```diff
@@ -56,30 +50,8 @@ pub enum GeneralCommands {
         #[clap(long)]
         federation_id: Option<FederationId>,
     },
-    /// Check gateway's e-cash balance on the specified federation.
```
dpc, line 59: We might consider leaving the old commands for time being to help with the transition.
We don't have to do it. But we might consider it. :D

dpc (commented):
Looks like an improvement. Comment.

gateway/cli/src/general_commands.rs
m1sterc001guy: With the `gateway-cli` we don't try to maintain backwards compatibility, so I would prefer just to move them :)

m1sterc001guy: They're not! Which is nice, we added some logic in the backwards compatibility tests that always ensure `gateway-cli` version and `gatewayd` version are the same.

dpc: Go for it then. :)

gateway/cli/src/ecash_commands.rs
```diff
@@ -0,0 +1,121 @@
...
+                    .await?;
+
+                print_response(response);
+            }
+            Self::ReceiveEcash { notes, wait } => {
+                let response = create_client()
+                    .receive_ecash(ReceiveEcashPayload { notes, wait })
+                    .await?;
+                print_response(response);
+            }
```
elsirion, line 116: Since it's under the e-cash subcommand already, maybe just call it spend/receive (or deposit/withdraw) for e-cash to e-cash?

elsirion (approved):
Generally in favor, one nit

gateway/cli/src/ecash_commands.rs
m1sterc001guy: Good idea. I changed to `send` and `receive` which is consistent with LNv2

dpc (approved):

PR #6271, opened 2024-11-01
chore(deps): partially bump bitcoin v0.32

tvolk131:
Part of #5200 

I've added comments to all non-trivial changes

I recommend reviewing commit-by-commit

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -8,7 +8,9 @@ use std::{env, ffi};
 use anyhow::{anyhow, bail, Context, Result};
 use bitcoin::Txid;
 use clap::Subcommand;
-use fedimint_core::bitcoin_migration::bitcoin32_to_bitcoin30_tx;
+use fedimint_core::bitcoin_migration::{
+    bitcoin32_to_bitcoin30_script_buf, bitcoin32_to_bitcoin30_tx,
+};
 use fedimint_core::core::LEGACY_HARDCODED_INSTANCE_ID_WALLET;
 use fedimint_core::encoding::Decodable;
 use fedimint_core::envs::is_env_var_set;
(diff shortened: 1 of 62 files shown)


devimint/src/tests.rs
```diff
@@ -2088,25 +2092,26 @@ pub async fn recoverytool_test(dev_fed: DevFed) -> Result<()> {
...
+                .iter()
+                .map(|d| {
+                    json!({
+                        "desc": d,
+                        "timestamp": 0,
+                    })
+                })
+                .collect(),
+        ),
+    ]))?;
```
tvolk131, line 2107: Note for reviewer: The double-array here looks strange, but is intentional. With only a single array, `recoverytool_tests` fails due to a malformatted request. The change here is required due to [this](https://github.com/apoelstra/rust-jsonrpc/pull/108) breaking change that was introduced in v0.17 of the `jsonrpc` crate.

fedimint-core/src/encoding/mod.rs
```diff
@@ -970,6 +970,17 @@ impl Decodable for Cow<'static, str> {
...
+// Simple decoder implementing `bitcoin_io::Read` for `std::io::Read`.
+// This is needed because `bitcoin::consensus::Decodable` requires a
+// `bitcoin_io::Read`.
+pub struct SimpleBitcoinRead<R: std::io::Read>(R);
+
+impl<R: std::io::Read> bitcoin_io::Read for SimpleBitcoinRead<R> {
+    fn read(&mut self, buf: &mut [u8]) -> bitcoin_io::Result<usize> {
+        self.0.read(buf).map_err(bitcoin_io::Error::from)
+    }
+}
```
tvolk131, line 982: Note for reviewers: We're moving this to a more central place since it will be useful for implementing `Decodable` across the entire codebase later in the bitcoin v0.32 migration process

joschisan (approved):

PR #6284, opened 2024-11-04
chore: test gateway cache and selection

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-client/src/cli.rs
+++ b/modules/fedimint-lnv2-client/src/cli.rs
@@ -39,6 +39,9 @@ enum Opts {
 
 #[derive(Clone, Subcommand, Serialize)]
 enum GatewayOpts {
+    /// Update cache of gateway information to optimise gateway selection for a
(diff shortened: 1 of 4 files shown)


modules/fedimint-lnv2-client/src/cli.rs
```diff
@@ -39,6 +39,9 @@ enum Opts {
 
 #[derive(Clone, Subcommand, Serialize)]
 enum GatewayOpts {
+    /// Update cache of gateway information to optimise gateway selection for a
+    /// given invoice.
+    Cache,
```
m1sterc001guy, line 44: nit: `cache` is a noun whereas the rest of the commands are verbs

m1sterc001guy: eh I suppose cache is also a verb, nevermind

m1sterc001guy (dismissed):

m1sterc001guy (approved):

PR #6285, opened 2024-11-05
refactor(client): handle multiple outputs with a single state machine

dpc:
On top of #6264 .


elsirion (approved):
Looks good, I'm a bit suspicious if there are any other places than `await_output_finalized` where we need to wait for `CreatedMulti` now. I couldn't find any myself though.

PR #6286, opened 2024-11-05
Lnv2 api cleanup

joschisan:
Don't concern integrators with stuff that is not of their concern.


modules/fedimint-lnv2-client/src/lib.rs
```diff
@@ -34,33 +34,42 @@ use fedimint_core::bitcoin_migration::{
...
+use tpe::{derive_agg_decryption_key, AggregateDecryptionKey};
 
-use crate::api::LnFederationApi;
+use crate::api::LightningFederationApi;
 use crate::receive_sm::{ReceiveSMCommon, ReceiveSMState, ReceiveStateMachine};
 use crate::send_sm::{SendSMCommon, SendSMState, SendStateMachine};
 
+/// Number of blocks until outgoing lightning contracts times out and user
+/// client can refund it unilaterally
+const EXPIRATION_DELTA_LIMIT: u64 = 1440;
```
m1sterc001guy, line 68: Any reason for changing this away from 1008?

m1sterc001guy: IMO we should change the CLTV delta on the contact to be the `min_final_cltv_delta` of the invoice plus some constant, then check if that's higher than this limit. Can do this in a separate improvement though.

m1sterc001guy (approved):

modules/fedimint-lnv2-client/src/lib.rs
joschisan: I added just a bit more headroom here and its exactly 10 days now.

PR #6301, opened 2024-11-07
feat: make introduction of new ci variants consensus safe

joschisan:

### DIFF
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -769,7 +769,8 @@ impl ConsensusEngine {
                     target: LOG_CONSENSUS,
                     "Minor consensus version mismatch: unexpected consensus item type: {variant}"
                 );
-                bail!("Unexpected consensus item type: {variant}")
+
+                panic!("Unexpected consensus item type: {variant}")
             }
         }
     }
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -498,7 +498,7 @@ impl ServerModule for Wallet {
                 }
             }
             WalletConsensusItem::Default { variant, .. } => {
-                bail!("Received wallet consensus item with unknown variant {variant}");
+                panic!("Received wallet consensus item with unknown variant {variant}");
             }
         }
 


elsirion (commented):
We'd have to introduce panics in many more places, basically for every `Default` variant. Idk if that's worth it in general. 

Also, we lose BFT that way, a malicious guardian could trigger it, so I'd at least like a way to turn the panic into an error.

dpc:
> we lose BFT that way

My first thought.

joschisan:
@elsirion Why would we have to? these are the only places the transaction size limit would touch.

Off course we lose BFT, but we would do so with your intended upgrade method anyway and we reduce the damage from a database corruption to a shutdown - also in case of a mistake. 

We can always revert this once we have fixed the wallet transaction size limit.

Any questions? otherwise we should get this merged..

elsirion:
> @elsirion Why would we have to? these are the only places the transaction size limit would touch.

The title read more like a generalized proposal, if you are only talking about that one particular planned wallet upgrade then I agree (but only after thinking about it a lot, I expected us to at least need the same handling for wallet inputs, but that turns out to be unnecessary).

As so often a PR description would have been useful (even though we talked about it, I expected the PR to be about a more general case). 

> Off course we lose BFT, but we would do so with your intended upgrade method anyway and we reduce the damage from a database corruption to a shutdown - also in case of a mistake.

We'd lose BFT during the transition while everyone is upgrading, which hopefully would only take a week or so in most cases (the risk comes from some honest peers being upgraded and not others and a malicious one casting a vote). 

This proposal introduces the risk for the entire time till we do the upgrade. I don't think this is a major drawback though if there's a way to deactivate the panic one a guardian proves malicious.

> We can always revert this once we have fixed the wallet transaction size limit.
> Any questions? otherwise we should get this merged..

My reason for merging would be to make introducing version voting a bit safer. I'm not sold on kicking the can down the road any further. 

In the past there wasn't anything pressing that needed a consensus upgrade, so I agreed that not introducing voting yet was a good trade-off. But now implementing the wallet upgrade properly will require a bunch of custom/non-reusable logic to that makes honest guardians block new API calls until all peers are upgraded that I'd much rather see exist around consensus version voting, which we'll likely use 10s of times in the future.

This planned wallet upgrade is the *ideal* test run for consensus version voting:
* It's relevant enough to warrant us doing a consensus upgrade and not just ignoring it
* At the same time it's not an emergency, so we can do it properly and get experience for the next time 
* Having it in will make a lot of other (smaller) fixes, we'd otherwise ignore, more feasible. Ideally simple fixes (e.g. inverting coin selection) become as easy as DB migrations. *This one you might disagree with and it's the least important to me of these three.*

**My biggest fear with kicking the can further down the road is that we'll have an emergency consensus fix to deploy, have to merge version voting in a hurry and have no operational experience with it yet and no time to think through all the possible problems. This *will* happen eventually, software is always buggy, so it's just a question of time (and scrutiny of the code base). I want to be prepared.**

PR #6302, opened 2024-11-07
chore(deps): bump esplora-client to v0.10

tvolk131:


fedimint-bitcoind/src/esplora.rs
```diff
@@ -76,16 +74,15 @@ impl IBitcoindRpc for EsploraClient {
...
     }
 
     async fn get_fee_rate(&self, confirmation_target: u16) -> anyhow::Result<Option<Feerate>> {
-        let fee_estimates: HashMap<String, f64> = self.client.get_fee_estimates().await?;
+        let fee_estimates: HashMap<u16, f64> = self.client.get_fee_estimates().await?;
 
         let fee_rate_vb =
-            esplora_client::convert_fee_rate(confirmation_target.into(), fee_estimates)?;
+            esplora_client::convert_fee_rate(confirmation_target.into(), fee_estimates)
+                .unwrap_or(1.0);
```
tvolk131, line 85: Note for reviewers: The diffs here stem from two changes in `get_fee_estimates()` and `convert_fee_rate()` between `esplora-client` v0.6 and v0.10:

1. The key type of fee estimates is now a `u16` rather than a `String`, pushing the parsing logic from `convert_fee_rate()` into `get_fee_estimates()` and removing the need for `convert_fee_rate()` to return a `Result`
2. Rather than always returning a fee rate by defaulting to 1 sat/vbyte, `convert_fee_rate()` now returns an `Option`, which is `None` if there is no applicable fee rate to use

Here are the two implementations of `convert_fee_rate()` for reference:

https://docs.rs/esplora-client/0.6.0/src/esplora_client/lib.rs.html#83-97
https://docs.rs/esplora-client/0.10.0/src/esplora_client/lib.rs.html#105-111

Because of the second change mentioned above, we're adding an `unwrap_or(1.0)` to our code to maintain the behavior that was removed.

elsirion: Good to not do a behavior change here, but worth considering @joschisan @bradleystachurski

elsirion (approved):

PR #6306, opened 2024-11-07
refactor(client): newtype for the range of in/out idxs

dpc:

### DIFF
--- a/fedimint-client/src/module/mod.rs
+++ b/fedimint-client/src/module/mod.rs
@@ -993,6 +993,7 @@ impl IntoIterator for IdxRange {
         ops::RangeInclusive::new(self.start, self.end_inclusive)
     }
 }
+
 impl From<ops::RangeInclusive<u64>> for IdxRange {
     fn from(value: ops::RangeInclusive<u64>) -> Self {
         Self {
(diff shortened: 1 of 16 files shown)


fedimint-client/src/module/mod.rs
```diff
@@ -959,5 +960,47 @@ impl AsRef<maybe_add_send_sync!(dyn IClientModule + 'static)> for DynClientModul
     }
 }
 
+/// A contiguous range of input/output indexes
+#[derive(Copy, Clone, Encodable, Decodable, PartialEq, Eq, Hash, Debug)]
+pub struct IdxRange {
+    start: u64,
+    end_inclusive: u64,
+}
```
elsirion, line 968: Would it make sense to have the txid in here too? That way you could decide to iterate over it as `Iter<Item=OutPoint>` or `Iter<Item=u64>`. Afaik all places where we use the range now also have a `Txid`.

dpc: Can try `OutPointRange` type.

dpc: @elsirion Done.

fedimint-client/src/module/mod.rs
```diff
@@ -959,5 +960,111 @@ impl AsRef<maybe_add_send_sync!(dyn IClientModule + 'static)> for DynClientModul
...
+
+impl OutPointRange {
+    pub fn txid(&self) -> TransactionId {
+        self.txid
+    }
+}
+
+impl Iterator for OutPointRangeIter {
+    type Item = (TransactionId, u64);
```
elsirion, line 1063: Why not `OutPoint`?

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1084,16 +1084,19 @@ impl MintClientModule {
             }
         }
 
-        let state_generator = Arc::new(move |txid, out_idxs: RangeInclusive<u64>| {
-            assert_eq!(out_idxs.clone().count(), issuance_requests.len());
+        let state_generator = Arc::new(move |out_point_range: OutPointRange| {
+            assert_eq!(out_point_range.into_iter().count(), issuance_requests.len());
```
elsirion, line 1088: Why not `out_point_range.count()`?

elsirion (approved):
Very semantically pleasing! Although I had hoped for the code using the new range type to increase in pretty-ness more than it did :sweat_smile: I overestimated the number of places where we actually care about iterating over the `OutPoint`s.

PR #6308, opened 2024-11-07
chore: include federation name in `FederationInfo`

m1sterc001guy:
It is useful to have when managing multiple federations in a gateway.


bradleystachurski (dismissed):

elsirion (approved):
Concept ACK, but you had to touch 3 different places, looks like an opportunity to make it a function?

elsirion (approved):

PR #6321, opened 2024-11-09
chore: remove bitcoin v0.29 en/decoding of PSBTs

tvolk131:
Part of #5200

### DIFF
--- a/fedimint-core/src/encoding/btc.rs
+++ b/fedimint-core/src/encoding/btc.rs
@@ -7,11 +7,10 @@ use bitcoin::hashes::Hash as BitcoinHash;
 use hex::{FromHex, ToHex};
 use miniscript::{Descriptor, MiniscriptKey};
 
-use super::SimpleBitcoinRead;
+use super::{BufBitcoinReader, CountWrite, SimpleBitcoinRead};
 use crate::bitcoin_migration::{
-    bitcoin29_to_bitcoin32_network_magic, bitcoin29_to_bitcoin32_psbt,
-    bitcoin30_to_bitcoin32_network, bitcoin32_checked_address_to_unchecked_address,
-    bitcoin32_to_bitcoin29_network_magic, bitcoin32_to_bitcoin29_psbt,
+    bitcoin29_to_bitcoin32_network_magic, bitcoin30_to_bitcoin32_network,
+    bitcoin32_checked_address_to_unchecked_address, bitcoin32_to_bitcoin29_network_magic,
     bitcoin32_to_bitcoin30_address,
 };
 use crate::encoding::{Decodable, DecodeError, Encodable};
@@ -54,10 +53,7 @@ impl_encode_decode_bridge!(bitcoin::merkle_tree::PartialMerkleTree);
 
 impl crate::encoding::Encodable for bitcoin::psbt::Psbt {
     fn consensus_encode<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, std::io::Error> {
-        bitcoin29::consensus::Encodable::consensus_encode(
-            &bitcoin32_to_bitcoin29_psbt(self),
-            writer,
-        )
+        Ok(self.serialize_to_writer(&mut CountWrite::from(writer))?)
     }
 }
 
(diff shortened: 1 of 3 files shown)


fedimint-core/src/encoding/btc.rs
```diff
@@ -66,10 +62,8 @@ impl crate::encoding::Decodable for bitcoin::psbt::Psbt {
         d: &mut D,
         _modules: &ModuleDecoderRegistry,
     ) -> Result<Self, crate::encoding::DecodeError> {
-        Ok(bitcoin29_to_bitcoin32_psbt(
-            &bitcoin29::consensus::Decodable::consensus_decode_from_finite_reader(d)
-                .map_err(crate::encoding::DecodeError::from_err)?,
-        ))
+        Self::deserialize_from_reader(&mut std::io::BufReader::new(d))
```
elsirion, line 65: You need to use the same simple reader impl you used everywhere else. The `BufReader` will throw away leftover bytes.

fedimint-core/src/encoding/mod.rs
```diff
@@ -1004,11 +1004,70 @@ impl Decodable for Cow<'static, str> {
...
+    #[inline]
+    fn consume(&mut self, amount: usize) {
+        debug_assert!(false, "rust-bitcoin doesn't actually use this");
+        if amount >= 1 {
+            debug_assert_eq!(amount, 1, "Can only consume one byte");
+            debug_assert!(!self.is_consumed, "Cannot consume more than had been read");
+            self.is_consumed = true;
+        }
     }
 }
```
tvolk131, line 1072: Note for reviewers: This code comes from the lightning crate, which ran into the same problem we did with regard to `bitcoin_io::[Buf]Read`. See here: https://github.com/lightningdevkit/rust-lightning/blob/5718baaed947fcaa9c60d80cdf309040c0c68489/lightning/src/util/ser.rs#L72-L138

fedimint-core/src/encoding/mod.rs
```diff
@@ -1004,11 +1004,70 @@ impl Decodable for Cow<'static, str> {
...
+        Ok(self
+            .inner
+            .read(&mut output[offset..])
+            .map(|len| len + offset)?)
+    }
+}
 
-impl<R: std::io::Read> bitcoin_io::Read for SimpleBitcoinRead<R> {
-    fn read(&mut self, buf: &mut [u8]) -> bitcoin_io::Result<usize> {
-        self.0.read(buf).map_err(bitcoin_io::Error::from)
```
elsirion, line 1011: I wouldn't remove this impl, it's much better. The fake `BufRead` impl should only be used where absolutely necessary.

fedimint-core/src/encoding/btc.rs
tvolk131: Fixed

fedimint-core/src/encoding/mod.rs
tvolk131: Wouldn't that make it unsafe to switch between using `bitcoin_io::Read` and `bitcoin_io::BufRead`? I know we're not actually doing that, but if we ever needed to, that could lead to very wonky behavior.

elsirion: Why would it be unsafe? You can use `SimpleBitcoinRead` or `FakeBufRead` where `bitcoin_io::Read` is required and use `FakeBufRead` where `bitcoin_io::BufRead` is required. I'd just prefer to use `SimpleBitcoinRead` wherever possible since it allows reading multiple bytes at once if one knows how many bytes are expected. `FakeBufRead` always has to go byte by byte.

tvolk131: Discussed offline. Eric was thinking we add a separate implementation for `BufRead` whereas I thought we were extending `SimpleBitcoinRead`, in which case we would need the more complex implementation. We're going with the former.

fedimint-core/src/encoding/mod.rs
```diff
@@ -1012,6 +1012,80 @@ impl<R: std::io::Read> bitcoin_io::Read for SimpleBitcoinRead<R> {
     }
 }
 
+/// Wrap buffering support for implementations of Read.
+/// A reader which keeps an internal buffer to avoid hitting the underlying
+/// stream directly for every read.
+///
+/// In order to avoid reading bytes past the first object, and those bytes then
+/// ending up getting dropped, this BufBitcoinReader operates in
+/// one-byte-increments.
```
elsirion, line 1021: Worth pointing to the source where we copied it from.

tvolk131: Done

elsirion (approved):

PR #6322, opened 2024-11-09
chore: add vendored format_hex() from bitcoin_hashes v0.11

tvolk131:
Part of #5200

No functional change, just vendoring [this](https://docs.rs/bitcoin_hashes/0.11.0/bitcoin_hashes/hex/fn.format_hex.html) function so we can remove `bitcoin_hashes` v0.11


elsirion (commented):
Did you look into [`hex-conserative`](https://github.com/rust-bitcoin/hex-conservative)? I think that's what we should switch to longer term. Maybe worth for this already or maybe not.

elsirion (approved):

elsirion (approved):

PR #6326, opened 2024-11-09
chore: inline network en/decoding

tvolk131:


fedimint-core/src/encoding/btc.rs
```diff
@@ -154,10 +163,22 @@ impl Decodable for bitcoin::p2p::Magic {
...
+        // https://docs.rs/bitcoin/0.29.2/src/bitcoin/network/constants.rs.html#81-84
+        // https://docs.rs/bitcoin/0.32.2/src/bitcoin/p2p/mod.rs.html#216-223
+        let bytes = [
+            (num & 0xFF) as u8,
+            ((num >> 8) & 0xFF) as u8,
+            ((num >> 16) & 0xFF) as u8,
+            ((num >> 24) & 0xFF) as u8,
+        ];
+
+        Ok(Self::from_bytes(bytes))
```
elsirion, line 181: I guess we are stuck with the `u32` decoding because it's encoded as a varint and thus a bit longer than 4 bytes? Even if that's the case and we can't directly read the bytes, there is a better way to turn the `u32` into bytes: [`u32::to_le_bytes`](https://doc.rust-lang.org/std/primitive.u32.html#method.to_le_bytes) (it looks at least as if you needed little endian encoding).

tvolk131: Ahh much cleaner!

dpc (dismissed):

elsirion (approved):

PR #6332, opened 2024-11-11
chore: remove Encodable impl for checked addresses

tvolk131:


elsirion (approved):

elsirion:
@joschisan made me aware that any change in address encoding could be a consensus breaking change. Fortunately, the encoding/decoding for mainnet seems unaffected, only testnet/signet/regtest might change and lead to such federations not being able to recover from history anymore in the worst case. I think this is an ok trade-off. 

Do you know if this is breaking re-encodability in any way @tvolk131? I guess it was already broken, it might just be broken in a different way now and only for non-mainnet networks.

m1sterc001guy (approved):

dpc:
Why?

tvolk131:
> Why?

Because it's not being used, and Bitcoin v0.32 has a function to easily convert a checked address to an unchecked address: https://docs.rs/bitcoin/latest/bitcoin/struct.Address.html#method.as_unchecked

elsirion (approved):

PR #6335, opened 2024-11-12
cleanup: only check for mainnet

joschisan:
I propose this since we had issues around distinguishing between the non-main net types. Especially in the consensus code this is highly problematic. 

I would do the same for the walletv2 modules, getting completely rid of the distinction in the future.

@elsirion @dpc thoughts?

I think this will break the upgrade tests, though it shouldn't. We had this issue the last time we made changes to the lnv2 client config.

Since lnv2 is about to be shipped, this is the last chance to remove it.

### DIFF
--- a/devimint/src/federation/config.rs
+++ b/devimint/src/federation/config.rs
@@ -90,7 +90,7 @@ pub fn attach_default_module_init_params(
                 consensus: fedimint_lnv2_common::config::LightningGenParamsConsensus {
                     fee_consensus: fedimint_lnv2_common::config::FeeConsensus::new(1000)
                         .expect("Relative fee is within range"),
-                    network,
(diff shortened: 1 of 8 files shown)


elsirion:
:exploding_head: this actually works for LNv2 because now the LN node generates the invoice, so even though Fedimint doesn't know the exact network anymore the correct prefix will be used for invoices.

This won't be viable for Wallet v2 unfortunately though since we need to know the network to produce well-formed addresses for other networks. I agree that for anything in consensus we should either not care about the network at all (it's static after all and can be checked before submitting to consensus) or at least use a boolean like you do here.

Longer term I can imagine us just using a custom struct that contains all the data we need about a network to produce addresses, LN invoices, validate we are on the right one etc. and not rely on an enum anymore.

PR #6336, opened 2024-11-12
cleanup: sane network encoding for lnv2

joschisan:
Encode magic bytes directly instead of as u32 which was a historical accident.

Will also be done with walletv2. We can then retire the old encoding eventually with lnv1 and walletv1.

### DIFF
--- a/fedimint-core/src/encoding/btc.rs
+++ b/fedimint-core/src/encoding/btc.rs
@@ -6,6 +6,7 @@ use bitcoin::address::NetworkUnchecked;
 use bitcoin::hashes::Hash as BitcoinHash;
 use hex::{FromHex, ToHex};
 use miniscript::{Descriptor, MiniscriptKey};
+use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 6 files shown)


fedimint-core/src/encoding/btc.rs
```diff
@@ -170,6 +171,29 @@ impl Decodable for bitcoin::Network {
     }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct NetworkSaneEncodingWrapper(pub bitcoin::Network);
```
elsirion, line 175: Love the name :smile:

elsirion (approved):
Love it! Eventually I'd like to be annotate fields with custom encode `fn`s, but for now the newtype is great.

fedimint-core/src/encoding/btc.rs
dpc: Docstring comment would be useful.

PR #6340, opened 2024-11-12
fix: find_by_range Prefix DB Bug

m1sterc001guy:
`find_by_range` has a small bug that prevents `PrefixDatabaseTransaction` from finding elements. Its just a typo in the range. Added a test to cover this scenario.


dpc (approved):
Ah, shit. Good catch.

AFAICT we only use this in event log, so should have caused any damage yet.

PR #6344, opened 2024-11-13
dbtool: add dump for operation log and client config

maan2003:
very helpful for debugging, hex is annoying

hit bunch of decoder issues when logging active and inactive state machines, for some other time

### DIFF
--- a/fedimint-core/src/core.rs
+++ b/fedimint-core/src/core.rs
@@ -281,11 +281,11 @@ impl DecoderBuilder {
                 // pass empty `decoders`. But the client context uses nested `DynTypes` in
                 // `DynState`, so we special-case it with a flag.
                 let decoders = if is_transparent_decoder {
-                    Cow::Borrowed(decoders)
+                    decoders
                 } else {
-                    Cow::Owned(ModuleRegistry::default())
+                    &ModuleRegistry::default()
                 };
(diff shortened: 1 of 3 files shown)


fedimint-core/src/core.rs
```diff
@@ -281,11 +281,11 @@ impl DecoderBuilder {
                 // pass empty `decoders`. But the client context uses nested `DynTypes` in
                 // `DynState`, so we special-case it with a flag.
                 let decoders = if is_transparent_decoder {
-                    Cow::Borrowed(decoders)
+                    &decoders
                 } else {
-                    Cow::Owned(ModuleRegistry::default())
+                    &ModuleRegistry::default()
                 };
```
maan2003, line 287: the lifetime of expression gets auto extended

fedimint-dbtool/src/dump.rs
```diff
@@ -221,6 +223,9 @@ impl DatabaseDump {
         }
 
         if let Some(cfg) = self.client_cfg.clone() {
+            self.serialized
```
elsirion, line 226: Now we serialize module configs separately from the client config and also have the binary encoding of the module configs in the complete config. Instead we could use `JsonClientConfig` here.

elsirion (commented):
Clippy is still unhappy

dpc (approved):

m1sterc001guy (approved):

PR #6347, opened 2024-11-13
chore(rocksdb): lower memory footprint

dpc:
As per https://github.com/facebook/rocksdb/wiki/Block-Cache#caching-index-filter-and-compression-dictionary-blocks:

> By default index, filter, and compression dictionary blocks (with the exception of the partitions of [partitioned indexes/filters](https://github.com/facebook/rocksdb/wiki/Partitioned-Index-Filters)) are cached outside of block cache, and users won't be able to control how much memory should be used to cache these blocks, other than setting max_open_files.

### DIFF
--- a/fedimint-rocksdb/src/lib.rs
+++ b/fedimint-rocksdb/src/lib.rs
@@ -94,6 +94,14 @@ fn get_default_options() -> anyhow::Result<rocksdb::Options> {
         opts.set_db_write_buffer_size(16 * 1024 * 1024);
     }
     opts.create_if_missing(true);
+
+    let mut block_opts = rocksdb::BlockBasedOptions::default();
+    block_opts.set_optimize_filters_for_memory(true);
+    // counterintuitively, this limits amount of memory used as these will now be
+    // part of the block cache.
+    block_opts.set_cache_index_and_filter_blocks(true);
+    block_opts.set_ribbon_filter(10.0);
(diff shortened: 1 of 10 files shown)


fedimint-rocksdb/src/lib.rs
```diff
@@ -159,6 +161,28 @@ fn next_prefix(prefix: &[u8]) -> Option<Vec<u8>> {
 impl IRawDatabase for RocksDb {
     type Transaction<'a> = RocksDbTransaction<'a>;
     async fn begin_transaction<'a>(&'a self) -> RocksDbTransaction {
+        let count = self.1.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        if count % 100_000 == 0 {
```
elsirion, line 165: This is kinda hacky, do we even know how often that condition will be hit? I'd make it time based.

elsirion (approved):
Great optimizations! :rocket:

fedimint-rocksdb/src/lib.rs
dpc: Yeah, this one is purely a hack. I might need to clean it up into a proper abstraction.

dpc:
Turns out all these optimization don't matter AFAICT, and actually just switching to jemalloc behaves better. @elsirion

PR #6348, opened 2024-11-13
chore: swap NetworkSaneEncodingWrapper for Legacy

tvolk131:
Follow-up to #6336

Makes the default encoding for `bitcoin::Network` byte-encoded instead of int-encoded to prevent accidental use of the old encoding

To ensure safe conversion, this PR is split into two commits. One commit removes `bitcoin::Network` en/decoding and replaces it with a `NetworkLegacyEncodingWrapper`, and a second commit removes `NetworkSaneEncodingWrapper` and replaces it with `bitcoin::Network` en/decoding.

### DIFF
--- a/fedimint-core/src/encoding/btc.rs
+++ b/fedimint-core/src/encoding/btc.rs
@@ -168,27 +168,21 @@ impl Decodable for NetworkLegacyEncodingWrapper {
         Ok(Self(network))
     }
 }
-
-#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
-pub struct NetworkSaneEncodingWrapper(pub bitcoin::Network);
-
-impl Encodable for NetworkSaneEncodingWrapper {
+impl Encodable for bitcoin::Network {
     fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
-        self.0.magic().to_bytes().consensus_encode(writer)
+        self.magic().to_bytes().consensus_encode(writer)
     }
 }
 
-impl Decodable for NetworkSaneEncodingWrapper {
+impl Decodable for bitcoin::Network {
     fn consensus_decode<D: std::io::Read>(
         d: &mut D,
         modules: &ModuleDecoderRegistry,
     ) -> Result<Self, DecodeError> {
-        Ok(Self(
-            bitcoin::Network::from_magic(bitcoin::p2p::Magic::from_bytes(
-                Decodable::consensus_decode(d, modules)?,
-            ))
-            .ok_or_else(|| DecodeError::new_custom(format_err!("Unknown network magic")))?,
+        Self::from_magic(bitcoin::p2p::Magic::from_bytes(
+            Decodable::consensus_decode(d, modules)?,
         ))
+        .ok_or_else(|| DecodeError::new_custom(format_err!("Unknown network magic")))
     }
 }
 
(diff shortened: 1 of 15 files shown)


fedimint-core/src/encoding/btc.rs
```diff
@@ -281,30 +295,33 @@ mod tests {
...
         ];
 
-        for (network, magic_bytes, magic_sane_bytes) in networks {
-            let mut network_encoded = Vec::new();
-            network.consensus_encode(&mut network_encoded).unwrap();
+        for (network, magic_legacy_bytes, magic_sane_bytes) in networks {
+            let mut network_legacy_encoded = Vec::new();
+            NetworkLegacyEncodingWrapper(network)
+                .consensus_encode(&mut network_legacy_encoded)
+                .unwrap();
```
joschisan, line 296: You can use consensus_encode_to_vec for this.

fedimint-core/src/encoding/btc.rs
```diff
@@ -281,30 +295,33 @@ mod tests {
...
 
             let mut network_sane_encoded = Vec::new();
             NetworkSaneEncodingWrapper(network)
                 .consensus_encode(&mut network_sane_encoded)
                 .unwrap();
```
joschisan: You can use consensus_encode_to_vec for this.

fedimint-core/src/encoding/btc.rs
```diff
@@ -295,16 +289,14 @@ mod tests {
...
             NetworkLegacyEncodingWrapper(network)
                 .consensus_encode(&mut network_legacy_encoded)
                 .unwrap();
 
-            let mut network_sane_encoded = Vec::new();
-            NetworkSaneEncodingWrapper(network)
-                .consensus_encode(&mut network_sane_encoded)
-                .unwrap();
+            let mut network_encoded = Vec::new();
+            network.consensus_encode(&mut network_encoded).unwrap();
```
joschisan, line 299: You can use consensus_encode_to_vec for this.

joschisan:
Thats a good pr and well split into the two commits. It would merge it as is, however maybe we should maybe wait for the 0.5 cut with this as its not necessary to get in? @bradleystachurski

bradleystachurski:
> Thats a good pr and well split into the two commits. It would merge it as is, however maybe we should maybe wait for the 0.5 cut with this as its not necessary to get in? @bradleystachurski

I'm a bit confused by this, do we want this in 0.5? We have a [`releases/v0.5`](https://github.com/fedimint/fedimint/tree/releases/v0.5) branch so if we want this in 0.5 we'll need to backport after merging.

tvolk131:
> I'm a bit confused by this, do we want this in 0.5? We have a [`releases/v0.5`](https://github.com/fedimint/fedimint/tree/releases/v0.5) branch so if we want this in 0.5 we'll need to backport after merging.

No we don't need this in 0.5. Looks like this is safe to merge now that we have the first 0.5 rc out, right?

bradleystachurski:
>No we don't need this in 0.5. Looks like this is safe to merge now that we have the first 0.5 rc out, right?

Correct, this is safe to merge if we don't want it included in 0.5 since we already branched off `master` to create `releases/v0.5`.

joschisan:
I

fedimint-core/src/encoding/btc.rs
tvolk131: Looks like there are a bunch of places throughout the codebase where we could use `consensus_encode_to_vec()` instead of `consensus_encode().unwrap()`. Since this PR is already approved and ready to merge, I'll fix this in a follow-up.

fedimint-core/src/encoding/btc.rs
tvolk131: Looks like there are a bunch of places throughout the codebase where we could use `consensus_encode_to_vec()` instead of `consensus_encode().unwrap()`. Since this PR is already approved and ready to merge, I'll fix this in a follow-up.

fedimint-core/src/encoding/btc.rs
tvolk131: Done in #6364

fedimint-core/src/encoding/btc.rs
tvolk131: Done in #6364

PR #6351, opened 2024-11-14
feat: Gateway Transaction Log

m1sterc001guy:
Re-uses @dpc's client log for a "gateway transaction log".

Adds five LNv2 events `OutgoingPaymentStarted`, `OutgoingPaymentSucceeded`, `IncomingPaymentStarted`, `IncomingPaymentSucceeded`, and `CompleteLightningPaymentSucceeded`. Can tackle failure events in a follow up.

Also adds a `/payment_log` endpoint that allows querying for these events. The API contains very basic predicate filtering based on `FederationId` and `EventKind`, but I suspect the proper way to use this API will be to periodically query the data from an ETL task, insert it into a SQL database, and then execute more interesting queries there (e.g group by weekly, swaps vs LN, etc).

The added LNv2 events contain information required to compute fees, latency, and other payment information. Looking for feedback if there's other convenient data that we might want to include.

### DIFF
--- a/gateway/cli/Cargo.toml
+++ b/gateway/cli/Cargo.toml
@@ -24,6 +24,7 @@ bitcoin = { workspace = true }
 clap = { workspace = true }
 clap_complete = "4.5.38"
 fedimint-core = { workspace = true }
+fedimint-eventlog = { workspace = true }
 fedimint-logging = { workspace = true }
 fedimint-mint-client = { workspace = true }
 lightning-invoice = { workspace = true }
--- a/gateway/cli/src/general_commands.rs
+++ b/gateway/cli/src/general_commands.rs
@@ -2,9 +2,10 @@ use anyhow::bail;
 use clap::Subcommand;
 use fedimint_core::config::FederationId;
 use fedimint_core::fedimint_build_code_version_env;
+use fedimint_eventlog::{EventKind, EventLogId};
 use ln_gateway::rpc::rpc_client::GatewayRpcClient;
 use ln_gateway::rpc::{
-    ConfigPayload, ConnectFedPayload, FederationRoutingFees, LeaveFedPayload,
+    ConfigPayload, ConnectFedPayload, FederationRoutingFees, LeaveFedPayload, PaymentLogPayload,
     SetConfigurationPayload,
 };
 
@@ -93,6 +94,20 @@ pub enum GeneralCommands {
     },
     /// Safely stop the gateway
     Stop,
+    /// List the transactions that the gateway has processed
+    PaymentLog {
+        #[clap(long)]
+        end_position: Option<EventLogId>,
+
+        #[clap(long, default_value_t = 25)]
+        pagination_size: usize,
+
+        #[clap(long)]
+        federation_id: FederationId,
+
+        #[clap(long)]
+        event_kinds: Vec<EventKind>,
+    },
 }
 
 impl GeneralCommands {
@@ -177,6 +192,22 @@ impl GeneralCommands {
             }
             Self::Stop => {
                 create_client().stop().await?;
+            }
+            Self::PaymentLog {
+                end_position,
+                pagination_size,
+                federation_id,
+                event_kinds,
+            } => {
+                let payment_log = create_client()
+                    .payment_log(PaymentLogPayload {
+                        end_position,
+                        pagination_size,
+                        federation_id,
+                        event_kinds,
+                    })
+                    .await?;
+                print_response(payment_log);
             }
         }
 
--- a/gateway/ln-gateway/src/events.rs
+++ b/gateway/ln-gateway/src/events.rs
@@ -1,0 +1,112 @@
+use std::time::SystemTime;
+
+use fedimint_core::config::FederationId;
+use fedimint_core::core::ModuleKind;
+use fedimint_core::Amount;
+use fedimint_eventlog::{Event, EventKind};
+use fedimint_lnv2_common::contracts::{Commitment, OutgoingContract, PaymentImage};
+use serde::{Deserialize, Serialize};
+use serde_millis;
+
+/// All gateway events will be emitted using the same module kind.
+pub const GATEWAY_KIND: ModuleKind = ModuleKind::from_static_str("ln-gateway-core");
+
+pub const ALL_GATEWAY_EVENTS: [EventKind; 5] = [
+    OutgoingPaymentStarted::KIND,
+    OutgoingPaymentSucceeded::KIND,
+    IncomingPaymentStarted::KIND,
+    IncomingPaymentSucceeded::KIND,
+    CompleteLightningPaymentSucceeded::KIND,
+];
+
+/// Event that is emitted when an outgoing payment attempt is initiated.
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentStarted {
+    /// The timestamp that the operation begins, including the API calls to the
+    /// federation to get the consensus block height.
+    #[serde(with = "serde_millis")]
+    pub operation_start: SystemTime,
+
+    /// The outgoing contract for this payment.
+    pub outgoing_contract: OutgoingContract,
+
+    /// The minimum amount that must be escrowed for the payment (includes the
+    /// gateway's fee)
+    pub min_contract_amount: Amount,
+
+    /// The amount requested in the invoice.
+    pub invoice_amount: Amount,
+
+    /// The max delay of the payment in blocks.
+    pub max_delay: u64,
+}
+
+impl Event for OutgoingPaymentStarted {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("outgoing-payment-started");
+}
+
+/// Event that is emitted when an outgoing payment attempt has succeeded.
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentSucceeded {
+    /// The target federation ID if a swap was performed, otherwise `None`.
+    pub target_federation: Option<FederationId>,
+}
+
+impl Event for OutgoingPaymentSucceeded {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("outgoing-payment-succeeded");
+}
+
+/// Event that is emitted when an incoming payment attempt has started. Includes
+/// both internal swaps and outside LN payments.
+#[derive(Serialize, Deserialize)]
+pub struct IncomingPaymentStarted {
+    /// The timestamp that the operation begins, including any metadata checks
+    /// before the state machine has spawned.
+    #[serde(with = "serde_millis")]
+    pub operation_start: SystemTime,
+
+    /// The commitment for the incoming contract.
+    pub incoming_contract_commitment: Commitment,
+
+    /// The amount requested in the invoice.
+    pub invoice_amount: Amount,
+}
+
+impl Event for IncomingPaymentStarted {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("incoming-payment-started");
+}
+
+/// Event that is emitted when an incoming payment attempt has succeeded.
+/// Includes both internal swaps and outside LN payments.
+#[derive(Serialize, Deserialize)]
+pub struct IncomingPaymentSucceeded {
+    /// The payment hash of the invoice that was paid.
+    pub payment_hash: PaymentImage,
+}
+
+impl Event for IncomingPaymentSucceeded {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("incoming-payment-succeeded");
+}
+
+/// Event that is emitted when a preimage is revealed to the Lightning network.
+/// Only emitted for payments that are received from an external Lightning node,
+/// not internal swaps.
+#[derive(Serialize, Deserialize)]
+pub struct CompleteLightningPaymentSucceeded {
+    /// The payment hash of the invoice that was paid.
+    pub payment_hash: PaymentImage,
+}
+
+impl Event for CompleteLightningPaymentSucceeded {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("complete-lightning-payment-succeeded");
+}
--- a/gateway/ln-gateway/src/gateway_module_v2/receive_sm.rs
+++ b/gateway/ln-gateway/src/gateway_module_v2/receive_sm.rs
@@ -23,6 +23,7 @@ use fedimint_lnv2_common::{
 use tpe::{aggregate_decryption_shares, AggregatePublicKey, DecryptionKeyShare, PublicKeyShare};
 use tracing::{error, trace};
 
+use crate::events::IncomingPaymentSucceeded;
 use crate::gateway_module_v2::GatewayClientContextV2;
 
 const RETRY_DELAY: Duration = Duration::from_secs(1);
@@ -103,6 +104,7 @@ impl State for ReceiveStateMachine {
     ) -> Vec<StateTransition<Self>> {
         let gc = global_context.clone();
         let tpe_agg_pk = context.tpe_agg_pk;
+        let gateway_context = context.clone();
 
         match &self.state {
             ReceiveSMState::Funding => {
@@ -133,6 +135,7 @@ impl State for ReceiveStateMachine {
                                 old_state,
                                 gc.clone(),
                                 tpe_agg_pk,
+                                gateway_context.clone(),
                             ))
                         },
                     ),
@@ -229,6 +232,7 @@ impl ReceiveStateMachine {
         old_state: ReceiveStateMachine,
         global_context: DynGlobalClientContext,
         tpe_agg_pk: AggregatePublicKey,
+        client_ctx: GatewayClientContextV2,
     ) -> ReceiveStateMachine {
         let decryption_shares = decryption_shares
             .into_iter()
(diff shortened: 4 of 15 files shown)


gateway/cli/src/general_commands.rs
```diff
@@ -93,6 +94,20 @@ pub enum GeneralCommands {
     },
     /// Safely stop the gateway
     Stop,
+    /// List the transactions that the gateway has processed
+    Transactions {
```
joschisan, line 98: Should the command not be named Payments similar to the event names? Or even PaymentLog to be more descriptive?

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::sha256::Hash;
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentSuccess {
+    pub outgoing_contract: OutgoingContract,
+    pub min_contract_amount: Amount,
+    pub invoice_amount: Amount,
+    pub start_secs: u64,
+    pub end_secs: u64,
+    pub max_delay: u64,
```
joschisan, line 16: Seconds do not seem finegrained enough for latency tracking. How about Instant?

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    const KIND: EventKind = EventKind::from_static("outgoing-payment-success");
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct IncomingPaymentSuccess {
+    pub incoming_contract_commitment: Commitment,
+    pub amount: Amount,
+    pub start_secs: u64,
+    pub end_secs: u64,
```
joschisan, line 31: Seconds do not seem finegrained enough for latency tracking. How about Instant?

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
...
+    const MODULE: Option<ModuleKind> = Some(fedimint_lnv2_common::KIND);
+
+    const KIND: EventKind = EventKind::from_static("incoming-payment-success");
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct CompleteLightningPaymentSuccess {
+    pub payment_hash: Hash,
+    pub start_secs: u64,
+    pub end_secs: u64,
```
joschisan, line 44: Seconds do not seem finegrained enough for latency tracking. How about Instant?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1611,6 +1615,73 @@ impl Gateway {
...
+                Some(_) => None,
+                None => Some((
+                    fed_id,
+                    federation_manager
+                        .client(&fed_id)
+                        .expect("Invalid federation")
+                        .value(),
+                )),
+            })
+            .collect::<Vec<_>>();
```
joschisan, line 1653: Seems like this could be simplified by pulling out the match on federation id out of the iterator.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1611,6 +1615,73 @@ impl Gateway {
...
+                .filter_map(|e| {
+                    if event_kinds.contains(&e.1) {
+                        Some((e.0, e.1, e.2, e.3, fed_id, e.4))
+                    } else {
+                        None
+                    }
+                })
+                .collect::<Vec<_>>();
+            transactions.extend(filtered_transactions);
+        }
```
joschisan, line 1680: This could be written as a stream with flatten and filter_map.

joschisan:
Generally, how about instead of using two events instead of one, like InitiateOutgoingPayment and OutgoingPaymentSuccess for example then the time does not need to be persisted in the state machine?

gateway/cli/Cargo.toml
```diff
@@ -23,6 +23,7 @@ anyhow = { workspace = true }
 bitcoin = { workspace = true }
 clap = { workspace = true }
 clap_complete = "4.5.37"
+fedimint-client = { path = "../../fedimint-client", version = "=0.5.0-alpha", default-features = false }
```
dpc, line 26: Though: maybe we should split the log functionality into own crate so we don't need to import the whole client here?

dpc:
I need to look more carefully on this, but can't ATM.

gateway/cli/src/general_commands.rs
m1sterc001guy: `PaymentLog` is good, I can change it to that

m1sterc001guy:
> Generally, how about instead of using two events instead of one, like InitiateOutgoingPayment and OutgoingPaymentSuccess for example then the time does not need to be persisted in the state machine?

This is doable but personally I would like to avoid it. With two events, it now requires a join between `InitiateOutgoingPayment` and `OutgoingPaymentSuccess` in order to compute the latency for each transaction (which will increase the query time).

gateway/cli/Cargo.toml
m1sterc001guy: Sounds like a good idea to me. Made an issue https://github.com/fedimint/fedimint/issues/6358

gateway/ln-gateway/src/events.rs
m1sterc001guy: `Instant` also does not preserve the absolute time that the transaction happened, which I think is valuable

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Refactored this a bit, let me know if this is what you had in mind. I had trouble pushing the filter inside of the stream but otherwise I think it is an improvement over how it was written before

dpc:
#6361 , maybe you can rebase on top after it lands.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1612,6 +1616,75 @@ impl Gateway {
...
+            .collect::<BTreeMap<_, _>>();
+
+        let clients = match federation_id {
+            Some(id) if all_clients.contains_key(&id) => {
+                let client = *all_clients.get(&id).expect("already checked");
+                vec![(id, client)]
+            }
+            Some(_) => {
+                return PaymentLogResponse(vec![]);
+            }
```
joschisan, line 1655: Should this not be a request error?

gateway/ln-gateway/src/lib.rs
```diff
@@ -1612,6 +1616,75 @@ impl Gateway {
...
+                    .into_iter()
+                    .map(|e| (e.0, e.1, e.2, e.3, fed_id, e.4))
+                    .collect::<Vec<_>>()
+            })
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .flatten()
+            .filter(|e| event_kinds.contains(&e.1))
+            .collect::<Vec<_>>();
```
joschisan, line 1683: nicely written!

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentSuccess {
+    pub outgoing_contract: OutgoingContract,
+    pub min_contract_amount: Amount,
+    pub invoice_amount: Amount,
+    pub start_millis: u128,
+    pub end_millis: u128,
+    pub max_delay: u64,
+    pub is_swap: bool,
```
m1sterc001guy, line 17: Make this an `Option<FederationId>`

gateway/ln-gateway/src/lib.rs
dpc: Could flatten and filter be done on a stream to avoid one allocation?

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
...
+    const MODULE: Option<ModuleKind> = Some(fedimint_lnv2_common::KIND);
+
+    const KIND: EventKind = EventKind::from_static("outgoing-payment-success");
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct IncomingPaymentSuccess {
+    pub incoming_contract_commitment: Commitment,
+    pub amount: Amount,
+    pub start_millis: u128,
```
dpc, line 30: Json has numerical precision issues, so I'm not sure if there's a point of using `u128`

dpc: https://github.com/fedimint/fedimint/blob/33efb3408b2b08426da7dd9caf51e1efa164e738/fedimint-eventlog/src/lib.rs#L157

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
```diff
@@ -234,4 +255,32 @@ impl SendStateMachine {
...
+    ) -> OutgoingPaymentSuccess {
+        let invoice_amount = match common.invoice {
+            LightningInvoice::Bolt11(invoice) => Amount::from_msats(
+                invoice
+                    .amount_milli_satoshis()
+                    .expect("amountless invoices are not supported"),
+            ),
+        };
+
+        let end_millis = now()
```
dpc, line 271: Yes, since you're capturing `end_millis` at the moment the event is being added, it will virtually equal to the automatically generated timestamp value.

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
+use fedimint_core::core::ModuleKind;
+use fedimint_core::Amount;
+use fedimint_eventlog::{Event, EventKind};
+use fedimint_lnv2_common::contracts::{Commitment, OutgoingContract};
+use serde::{Deserialize, Serialize};
+
+use crate::sha256::Hash;
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentSuccess {
```
dpc, line 10: https://github.com/fedimint/fedimint/blob/master/modules/fedimint-mint-client/src/event.rs#L22

Let's keep the naming to "past verb" `OutgoingPayementSucceded`?

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub min_contract_amount: Amount,
+    pub invoice_amount: Amount,
+    pub start_millis: u128,
+    pub end_millis: u128,
+    pub max_delay: u64,
+    pub is_swap: bool,
+}
+
+impl Event for OutgoingPaymentSuccess {
+    const MODULE: Option<ModuleKind> = Some(fedimint_lnv2_common::KIND);
```
dpc, line 21: This events do not seem to be inside lnv2 module, so possibly it should be `None`. These are here mostly to avoid collisions between module authors naming their events the same by accident. 

Alternatively we should make up a face `ModuleKind` for `ln-gateway-core`.

dpc (commented):
Alright. Left some comments, was mostly focused on event log technicalities, not the LN part.

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,51 @@
...
+use fedimint_lnv2_common::contracts::{Commitment, OutgoingContract};
+use serde::{Deserialize, Serialize};
+
+use crate::sha256::Hash;
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentSuccess {
+    pub outgoing_contract: OutgoingContract,
+    pub min_contract_amount: Amount,
+    pub invoice_amount: Amount,
```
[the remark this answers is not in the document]
m1sterc001guy: Discussed this in the call today, since it is redundant data we won't include it. We already have the payment hash and can lookup LN transaction information using that.

gateway/ln-gateway/src/events.rs
m1sterc001guy: Using `millis` in rust returns a `u128`. What are the precision issues? Would using a `String` or something be better?

dpc: https://www.pullrequest.com/blog/safely-handling-large-integers-in-json-best-practices-and-pitfalls/ , JS numbers are floats, and JSON is fundamnetally a JS-based interoperability format.

`millis` return `u128`, but for real timestamps `u64` is enough for next thousands of years.

m1sterc001guy: I can use a `u64`

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Added request error.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1612,6 +1616,75 @@ impl Gateway {
...
+                        .client(&fed_id)
+                        .expect("invalid federation")
+                        .value(),
+                )
+            })
+            .collect::<BTreeMap<_, _>>();
+
+        let clients = match federation_id {
+            Some(id) if all_clients.contains_key(&id) => {
+                let client = *all_clients.get(&id).expect("already checked");
```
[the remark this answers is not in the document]
m1sterc001guy: Done

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
m1sterc001guy: Removed this. We can rely on the timestamp of the event

gateway/ln-gateway/src/events.rs
m1sterc001guy: I made a fake `ModuleKind` for `ln-gateway-core`

gateway/ln-gateway/src/lib.rs
```diff
@@ -1612,6 +1615,75 @@ impl Gateway {
         Ok(())
     }
 
+    /// Queries the client log for payment events and returns to the user.
+    pub async fn handle_payment_log_msg(
+        &self,
+        PaymentLogPayload {
+            position,
+            limit,
```
m1sterc001guy, line 1623: I'm rethinking this API. It is a bit weird to pass a global `position` and `limit` when each connected client will have a separate position in the log. @dpc any thoughts? Maybe the gateway API should take a timestamp instead and do `find_by_range` until it passes the supplied timestamp

m1sterc001guy: On second thought, I think `federation_id` should be required. Then `position` and `limit` actually make sense.

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentSucceeded {
+    pub target_federation: Option<FederationId>,
+}
+
+impl Event for OutgoingPaymentSucceeded {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("outgoing-payment-success");
```
dpc, line 42: `...-succeeded`

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct IncomingPaymentSucceeded {
+    pub payment_hash: PaymentImage,
+}
+
+impl Event for IncomingPaymentSucceeded {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("incoming-payment-success");
```
dpc, line 66: Same

gateway/ln-gateway/src/gateway_module_v2/receive_sm.rs
```diff
@@ -58,6 +59,7 @@ pub struct ReceiveSMCommon {
     pub contract: IncomingContract,
     pub out_point: OutPoint,
     pub refund_keypair: Keypair,
+    pub amount: Amount,
```
elsirion, line 62: Also a breaking change that needs a migration, is it worth it?

m1sterc001guy: Thanks, good catch. Forgot to remove this, its not needed now.

m1sterc001guy: Removed

gateway/ln-gateway/src/gateway_module_v2/send_sm.rs
```diff
@@ -143,7 +147,7 @@ impl SendStateMachine {
         min_contract_amount: Amount,
         invoice: LightningInvoice,
         contract: OutgoingContract,
-    ) -> Result<[u8; 32], Cancelled> {
+    ) -> Result<([u8; 32], Option<FederationId>), Cancelled> {
```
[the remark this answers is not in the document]
m1sterc001guy: Added a `PaymentResponse` struct that encapsulates these

gateway/ln-gateway/src/gateway_module_v2/receive_sm.rs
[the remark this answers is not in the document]
m1sterc001guy: This has been removed.

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,78 @@
...
+impl Event for IncomingPaymentSucceeded {
+    const MODULE: Option<ModuleKind> = Some(GATEWAY_KIND);
+
+    const KIND: EventKind = EventKind::from_static("incoming-payment-succeeded");
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct CompleteLightningPaymentSucceeded {
+    pub payment_hash: PaymentImage,
+}
```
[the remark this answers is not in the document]
m1sterc001guy: Added some docs to this event and the others

gateway/ln-gateway/src/events.rs
```diff
@@ -0,0 +1,78 @@
...
+    OutgoingPaymentStarted::KIND,
+    OutgoingPaymentSucceeded::KIND,
+    IncomingPaymentStarted::KIND,
+    IncomingPaymentSucceeded::KIND,
+    CompleteLightningPaymentSucceeded::KIND,
+];
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentStarted {
+    pub operation_start_millis: u64,
```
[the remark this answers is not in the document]
m1sterc001guy: Yup. Using `SystemTime` now as the time and serializing as millis

elsirion (approved):
LGTM, would like @dpc to take a look to make sure the event log abstraction is used as intended

elsirion:
Or actually, he already reviewed I just saw, will go ahead and merge.

PR #6352, opened 2024-11-14
fix: make ldk pay implementation idempotent

joschisan:

### DIFF
--- a/gateway/ln-gateway/src/lightning/ldk.rs
+++ b/gateway/ln-gateway/src/lightning/ldk.rs
@@ -16,6 +16,7 @@ use ldk_node::lightning::ln::msgs::SocketAddress;
 use ldk_node::lightning::ln::PaymentHash;
 use ldk_node::lightning::routing::gossip::NodeAlias;
 use ldk_node::payment::{PaymentKind, PaymentStatus, SendingParameters};
+use lightning::ln::channelmanager::PaymentId;
 use lightning::ln::PaymentPreimage;
 use lightning::util::scid_utils::scid_from_parts;
 use lightning_invoice::Bolt11Invoice;
@@ -290,22 +291,25 @@ impl ILnRpcClient for GatewayLdkClient {
         max_delay: u64,
         max_fee: Amount,
     ) -> Result<PayInvoiceResponse, LightningRpcError> {
-        let payment_id = match self.node.bolt11_payment().send(
-            &invoice,
-            Some(SendingParameters {
-                max_total_routing_fee_msat: Some(Some(max_fee.msats)),
-                max_total_cltv_expiry_delta: Some(max_delay as u32),
-                max_path_count: None,
-                max_channel_saturation_power_of_half: None,
-            }),
-        ) {
-            Ok(payment_id) => payment_id,
-            Err(e) => {
-                return Err(LightningRpcError::FailedPayment {
-                    failure_reason: format!("LDK payment failed to initialize: {e:?}"),
-                });
-            }
-        };
+        let payment_id = PaymentId(*invoice.payment_hash().as_byte_array());
+
+        if self.node.payment(&payment_id).is_none() {
+            assert_eq!(
+                self.node
+                    .bolt11_payment()
+                    .send(
+                        &invoice,
+                        Some(SendingParameters {
+                            max_total_routing_fee_msat: Some(Some(max_fee.msats)),
+                            max_total_cltv_expiry_delta: Some(max_delay as u32),
+                            max_path_count: None,
+                            max_channel_saturation_power_of_half: None,
+                        }),
+                    )
+                    .expect("LDK payment failed to initialize"),
+                payment_id
+            );
+        }
 
         // TODO: Find a way to avoid looping/polling to know when a payment is
         // completed. `ldk-node` provides `PaymentSuccessful` and `PaymentFailed`


gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -290,22 +291,25 @@ impl ILnRpcClient for GatewayLdkClient {
...
+                            max_total_routing_fee_msat: Some(Some(max_fee.msats)),
+                            max_total_cltv_expiry_delta: Some(max_delay as u32),
+                            max_path_count: None,
+                            max_channel_saturation_power_of_half: None,
+                        }),
+                    )
+                    .expect("LDK payment failed to initialize"),
+                payment_id
+            );
+        }
```
elsirion, line 312: Unfortunately this leads to a race condition afaict (if a second `pay` is executed while the first on is between L296 and L297). But I like how easy that was to spot in your PR vs the other one (idk if it has the same problem, but I remember thinking about it).

joschisan: The method does not have to be executed in parallel though I think.

tvolk131: In my PR I'm working on adding a [`DashMap`](https://docs.rs/dashmap/6.1.0/dashmap/struct.DashMap.html) to allow for payment-level locking. You could do something like that here too. Use the payment ID as the map key and call [`DashMap::entry()`](https://docs.rs/dashmap/6.1.0/dashmap/struct.DashMap.html#method.entry) at the very beginning of `pay()` to lock out all other concurrent calls. Hold the lock until the payment has succeeded or failed, then when the lock is released, we can guarantee that other concurrent calls will return right away since `self.node.payment(&payment_id)` will be `Some`.

> I like how easy that was to spot in your PR vs the other one

The reason my PR is so much more complicated is because it also fixes the TODO below about looping/polling. Much of the complexity in that PR is due to using `ldk-node::Node`'s event queue to watch for [`PaymentSuccessful`](https://docs.rs/ldk-node/0.4.2/ldk_node/enum.Event.html#variant.PaymentSuccessful) and [`PaymentFailed`](https://docs.rs/ldk-node/0.4.2/ldk_node/enum.Event.html#variant.PaymentFailed) event types.

tvolk131: > The method does not have to be executed in parallel though I think.

This might not always be the case though, and letting parallel calls be undefined behavior seems like a big foot-gun. I think we should add payment-level locking, but if you're strongly against that, we should document the single-threadedness constraint of this function in `ILnRpcClient::pay()`.

elsirion: > The method does not have to be executed in parallel though I think.

If that's the case we should just add a mutex/some other synchronization primitive that ensures there's always only one thread in the critical section. Otherwise someone will inevitable call it in parallel.

If we want to just prevent implementation errors we could even use `try_lock` and panic instead of waiting.

m1sterc001guy: > The method does not have to be executed in parallel though I think

Can you expand on this more? Why? Lightning payments have the possibility of becoming stuck (i.e if a node along the route does not respond). If this method is not done in parallel, all payments through LDK are essentially serialized, which means that payment 2 would be stuck waiting for payment 1 to finish, which would drastically increase the latency for payment 2.

joschisan: By design this method will be called with the same payment hash only from one state machine. You would have to modify the client to trigger two parallel payments to the same invoice, in which case that gateway might panic as one of the payments could fail to initialize. Seems to me that this would be fixed if ldk-node would change their implementation; they have a lock internally already but do return an error in case of a duplicate payment instead of proceeding for idempotency.

m1sterc001guy: > You would have to modify the client to trigger two parallel payments to the same invoice

IIRC correctly this is possible today. Since the gateway's `pay` endpoint is called from the trigger function inside the client state machine, it is possible the future is cancelled when the app stops and re-executed when the application restarts. This would result in 2 API calls to the gateway with the same payment hash. The Gateway should definitely not panic in this scenario.

elsirion: > You would have to modify the client to trigger two parallel payments to the same invoice, in which case that gateway might panic as one of the payments could fail to initialize

Is this still true if two user clients in different federations try to pay the same invoice? This is a real possibility, I like racing for being the one who gets the payment through with others.

tvolk131: I built #6369 on top of this PR. It adds payment hash level locking to ensure multithreaded idempotency. It doesn't introduce much additional complexity (only 26 extra lines of code excluding `Cargo.lock`).

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -290,22 +291,25 @@ impl ILnRpcClient for GatewayLdkClient {
...
+                    .send(
+                        &invoice,
+                        Some(SendingParameters {
+                            max_total_routing_fee_msat: Some(Some(max_fee.msats)),
+                            max_total_cltv_expiry_delta: Some(max_delay as u32),
+                            max_path_count: None,
+                            max_channel_saturation_power_of_half: None,
+                        }),
+                    )
+                    .expect("LDK payment failed to initialize"),
```
tvolk131, line 309: Are we sure it's safe to panic here rather than return an error like we did before?

gateway/ln-gateway/src/lightning/ldk.rs
joschisan: @m1sterc001guy the api call starts the state machine which in turn calls the pay method.

joschisan: @elsirion Okay you won, it is possible.

elsirion:
Can be merged all together in https://github.com/fedimint/fedimint/pull/6369

PR #6353, opened 2024-11-14
chore: use workspace dependencies instead of relative paths

bradleystachurski:
WIP

### DIFF
--- a/gateway/cli/Cargo.toml
+++ b/gateway/cli/Cargo.toml
@@ -27,7 +27,7 @@ fedimint-core = { workspace = true }
 fedimint-logging = { workspace = true }
 fedimint-mint-client = { workspace = true }
 lightning-invoice = { workspace = true }
-ln-gateway = { package = "fedimint-ln-gateway", path = "../ln-gateway", version = "=0.5.0-alpha", default-features = false }
+ln-gateway = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 tokio = { workspace = true }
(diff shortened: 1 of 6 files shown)


gateway/cli/Cargo.toml
```diff
@@ -27,7 +27,7 @@ fedimint-core = { workspace = true }
 fedimint-logging = { workspace = true }
 fedimint-mint-client = { workspace = true }
 lightning-invoice = { workspace = true }
-ln-gateway = { package = "fedimint-ln-gateway", path = "../ln-gateway", version = "=0.5.0-alpha", default-features = false }
+ln-gateway = { workspace = true }
```
bradleystachurski, line 30: @m1sterc001guy @tvolk131 do you know if removing `default-features = false` will cause problems? For context, switching to `workspace = true` doesn't respect `default-features = false` if any member uses `default-features = true` (or doesn't specify).

> If you want the flexibility of deciding whether or not a dependency enables the default-features of a dependency, be sure to set `default-features = false` in the workspace definition. Just beware that if you build multiple workspace members at the same time, the features will be unified so that if one member sets `default-features = true` (which is the default if not explicitly set), the default-features will be enabled for all members using that dependency.

https://doc.rust-lang.org/edition-guide/rust-2024/cargo-inherited-default-features.html

bradleystachurski: Ahh, I think CI might be answering this question :slightly_smiling_face:

PR #6354, opened 2024-11-14
chore: switch to using jemalloc on all major binaries. (v0.4)

dpc:
This is said to improve rocksdb memory usage
caused by memory fragmentation.

In my tests, it saves half of peak and sustained memory being used.
It needs more longer runtime testing, but the initial results are so
good there's no reason to wait landing it, IMO.

"Reverse-backported" it to https://github.com/fedimint/fedimint/pull/6357


fedimint-cli/src/main.rs
```diff
@@ -1,5 +1,12 @@
 use fedimint_cli::FedimintCli;
 use fedimint_core::fedimint_build_code_version_env;
+#[cfg(not(target_env = "msvc"))]
+use jemallocator::Jemalloc;
+
+#[cfg(not(target_env = "msvc"))]
+#[global_allocator]
+// rocksdb suffers from memory fragmentation when using standard allocator
+static GLOBAL: Jemalloc = Jemalloc;
```
maan2003, line 9: do even care windows? (msvc)

RCasatta: if yes, probably the dep has to be gated something like so:

```
[target.'cfg(not(target_env = "msvc"))'.dependencies]
jemallocator = "0.5"
```

dpc: We probably don't, we don't have any CI to verify anything about it. I just copied the boilerplate from jemalloc's README and I don't think it hurts.

dpc: Will do it in https://github.com/fedimint/fedimint/pull/6357

elsirion:
I'd like to merge to master first, then we can trigger backports to 0.5 and 0.4 from there.

dpc:
OK, closing this one as redundant.

elsirion:
I'm happy to use this PR, just didn't want to end up in a situation where we get it on 0.4 and master and e.g. forget 0.5 :see_no_evil: Since the automatic backport failed for 0.4 let's use this PR to fix #6384.

elsirion (approved):

Issue #6338, opened 2024-11-12
Failing Upgrade Tests


bradleystachurski:
We introduced a change that causes upgrade tests to fail with

```
Error: command: fedimint-cli --data-dir=/tmp/nix-shell.lh5Zqz/devimint-2050963-520/clients/reissue-client-0 deposit-address

Caused by:
    exit status: 1
    stdout:
    {
      "error": "Decoding named block MintOOBStateMachineV1 {} state"
    }
```

https://github.com/fedimint/fedimint/actions/runs/11790851677/job/32841950317#step:6:5015

I'm currently running `git bisect` to find the commit that introduced the failure.

bradleystachurski:
```
0fb234be560a10d9d3b2cd090c904e80565e4014 is the first bad commit
commit 0fb234be560a10d9d3b2cd090c904e80565e4014
Author: Dawid Ciężarkiewicz <dpc@dpc.pw>
Date:   Wed Oct 16 13:13:03 2024 -0700

    chore: claim oob refunds with a single state machine

 fedimint-client/src/sm/executor.rs         |  14 +--
 fedimint-core/src/db/mod.rs                |   2 +
 modules/fedimint-mint-client/src/input.rs  |   2 +-
 modules/fedimint-mint-client/src/lib.rs    |  92 +++++++++--------
 modules/fedimint-mint-client/src/oob.rs    | 158 +++++++++++++++++++++++++++--
 modules/fedimint-mint-tests/tests/tests.rs |  80 ++++++++++++++-
 6 files changed, 289 insertions(+), 59 deletions(-)
```

PR #6359, opened 2024-11-15
fix: mint client state machine migration

bradleystachurski:
Fixes https://github.com/fedimint/fedimint/issues/6338

This only fixes one failure of upgrade tests. Since `master` and `releases/v0.5` are still broken after including this fix, I haven't kicked off a run in CI, only local tests. To verify locally against these changes, first checkout the commit that introduced this failure and run upgrade tests for `fedimint-cli`

```
git checkout 8d156fd39049203fc134c2debe99e1182c3a6d60
TEST_KINDS=fedimint-cli just test-upgrades v0.4.0 current &> results.log
```

Expected failure message

```
Error: command: fedimint-cli --data-dir=/tmp/nix-shell.lh5Zqz/devimint-2050963-520/clients/reissue-client-0 deposit-address

Caused by:
    exit status: 1
    stdout:
    {
      "error": "Decoding named block MintOOBStateMachineV1 {} state"
    }
```

After applying the changes in this PR, `fedimint-cli` upgrade tests pass.

It's worth noting that I verified each variant of `MintClientStateMachines` that's included in this migration (`Output`, `Input`, and `OOB`) failed individually prior to adding the logic to parse the struct length, so I'm confident all paths are tested in the upgrade tests.

### DIFF
--- a/modules/fedimint-mint-client/src/client_db.rs
+++ b/modules/fedimint-mint-client/src/client_db.rs
@@ -117,7 +117,7 @@ pub async fn migrate_to_v1(
     Ok(None)
 }
 
-/// Maps all `Unreachable` states in the state machine to `OutputDone`
+/// Migrates `MintClientStateMachinesV1`
 pub(crate) fn migrate_state_to_v2(
     operation_id: OperationId,
     cursor: &mut Cursor<&[u8]>,
@@ -128,7 +128,9 @@ pub(crate) fn migrate_state_to_v2(
 
     let new_mint_state_machine = match mint_client_state_machine_variant {
         0 => {
+            let _output_sm_len = u16::consensus_decode(cursor, &decoders)?;
             let old_state = MintOutputStateMachineV1::consensus_decode(cursor, &decoders)?;
+
             MintClientStateMachines::Output(MintOutputStateMachine {
                 common: MintOutputCommon {
                     operation_id: old_state.common.operation_id,
@@ -140,6 +142,7 @@ pub(crate) fn migrate_state_to_v2(
             })
         }
         1 => {
+            let _input_sm_len = u16::consensus_decode(cursor, &decoders)?;
             let old_state = MintInputStateMachineV1::consensus_decode(cursor, &decoders)?;
 
             MintClientStateMachines::Input(MintInputStateMachine {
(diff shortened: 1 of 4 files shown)


modules/fedimint-mint-client/src/client_db.rs
```diff
@@ -128,7 +128,9 @@ pub(crate) fn migrate_state_to_v2(
 
     let new_mint_state_machine = match mint_client_state_machine_variant {
         0 => {
-            let old_state = MintOutputStateMachineV1::consensus_decode(cursor, &decoders)?;
+            let _output_sm_len = u16::consensus_decode(cursor, &decoders)?;
```
bradleystachurski, line 131: We're failing to decode the structs in the migration if we don't first read the length.

dpc: Makes sense. I always forget about this extra len in variants.

But, I copied the code here from `modules/fedimint-dummy-client/src/db.rs` `get_v1_migrated_state` and I don't see it decoding any len either, so does it mean it's broken as well, and just was never actually tested?

m1sterc001guy: Hmm we do test it in `test_client_db_migrations` of the dummy tests crate, but perhaps its decoding to the wrong values. We don't test that part. I'll check

bradleystachurski: We can address the dummy client migrations in a followup. Right now this is making it difficult to continue triaging the rest of the failures in the upgrade tests, so if there isn't a logical error let's land soon :slightly_smiling_face:

m1sterc001guy (approved):

modules/fedimint-mint-client/src/client_db.rs
m1sterc001guy: All the reason why the dummy migration tests work is just a happy accident lol. The `Unreachable` state decodes a `u16` and I guess originally I thought this was the `ModuleInstanceId`, but its actually the length. 

```
impl Decodable for Unreachable {
    fn consensus_decode<R: std::io::Read>(
        reader: &mut R,
        modules: &ModuleDecoderRegistry,
    ) -> Result<Self, fedimint_core::encoding::DecodeError> {
        let module_instance_id = ModuleInstanceId::consensus_decode(reader, modules)?;
        let operation_id = OperationId::consensus_decode(reader, modules)?;
        let txid = TransactionId::consensus_decode(reader, modules)?;
        let amount = Amount::consensus_decode(reader, modules)?;

        Ok(Unreachable {
            _module_instance_id: module_instance_id,
            operation_id,
            txid,
            amount,
        })
    }
}
```

I can submit a quick fix.

m1sterc001guy: https://github.com/fedimint/fedimint/pull/6362

fedimint-backports:
- #6363

PR #6360, opened 2024-11-15
chore: set devimint mint fees to non-zero

joschisan:


elsirion:
@douglaz activating fees breaks the load test tool, how hard do you think is it to fix?

@dpc how hard do you think is it to test note consolidation in this setting?

modules/fedimint-mint-tests/src/bin/mint-client-sanity.rs
```diff
@@ -1,80 +1,4 @@
...
-    // reissuance of last note will trigger consolidation
-    cmd!(receiver, "reissue")
-        .args(&all_notes[all_notes.len() - 1..])
-        .run()
-        .await?;
-
-    let info = cmd!(receiver, "info").out_json().await?;
-    info!(%info, "receiver info");
-    // receiver has the balance
-    assert_eq!(info["total_amount_msat"].as_i64().unwrap(), 20);
```
dpc, line 75: I'm guessing not comparing the balance, and just asserting that there is fewer than 20 notes, should make it pass? @joschisan

douglaz:
> @douglaz activating fees breaks the load test tool, how hard do you think is it to fix?

Probably we just need to provide more money to the tool, but I'll take a closer look to see if this is the best solution

douglaz:
I'm trying to test the load test tool, but devimint seems to be broken on this branch. Am I doing something wrong?

```
$ just mprocs
./scripts/dev/mprocs/run.sh 
2024-11-20T19:14:07.080722Z  INFO fm::devimint: Devimint data dir path=/tmp/nix-shell.DQgh9g/devimint-294998-228
2024-11-20T19:14:12.641235Z  INFO fm::devimint: DKG completed
2024-11-20T19:14:17.744862Z  INFO fm::devimint: Consensus running
2024-11-20T19:14:26.617816Z  WARN fm::devimint: Main process failed, will shutdown err=command: fedimint-cli --data-dir=/tmp/nix-shell.DQgh9g/devimint-294998-228/clients/default-0 await-deposit be60fe98e85218e1649d448f1da0e9881e3d09d97f2a515cf1e494eb959206ea
2024-11-20T19:14:26.619401Z  WARN fm::devimint: failed to terminate lightningd plugins: command: lightning-cli --network regtest --lightning-dir=/tmp/nix-shell.DQgh9g/devimint-294998-228/cln plugin stop gateway-cln-extension

Caused by:
    A Tokio 1.x context was found, but it is being shutdown.
Error: command: fedimint-cli --data-dir=/tmp/nix-shell.DQgh9g/devimint-294998-228/clients/default-0 await-deposit be60fe98e85218e1649d448f1da0e9881e3d09d97f2a515cf1e494eb959206ea

Caused by:
    exit status: 1
    stdout:
    {
      "error": "The transaction is unbalanced (in=1000000000 msat, out=997995070 msat, fee=2004876 msat)"
    }
    
    stderr:
    2024-11-20T19:14:24.425014Z  WARN fm::db: Database commit failed in an autocommit block - retrying curr_attempts=1 err=Resource busy:  delay_ms=28
        
error: Recipe `mprocs` failed on line 120 with exit code 1

```

modules/fedimint-mint-tests/src/bin/mint-client-sanity.rs
joschisan: Probably yes but not for the right reason I think. I you reissue a 1msat note by itself with fees its unpredictable what kind of output denominations you will get.

joschisan: That's why I removed the test as it seems to rely on a transaction being finalized with specific denominations.

elsirion:
Let's block on a 0.5 release

modules/fedimint-mint-tests/src/bin/mint-client-sanity.rs
dpc: Could we migrate this test to a different run of devimint where the fees are set to 0?

dpc: Or rather: if `run_devfed_test` accepted arguments that allow setting fees to `0` then we could test stuff that requires no fees with no fees.

PR #6369, opened 2024-11-16
fix: make ldk pay implementation idempotent

tvolk131:
Makes `GatewayLdkClient::pay()` idempotent even when called simultaneously by multiple threads.

To achieve single-threaded idempotency, we do two things:
* Use the invoice's payment hash as the payment ID
* Skip initiating a payment with the underlying `ldk-node::Node` instance if it has already saved a payment with the given payment ID, indicating that it has already initiated a payment attempt for a given invoice

To achieve multi-threaded idempotency, we use a [`lockable::LockPool`](https://docs.rs/lockable/0.1.1/lockable/struct.LockPool.html) to lock by payment hash for the entire duration of a call to `GatewayLdkClient::pay()`. This should have almost no performance impact, since as soon as the first call completes and returns, subsequent calls simply need to grab the lock, lookup the result that's stored in the `ldk-node::Node`, and return.

### DIFF
--- a/gateway/ln-gateway/src/lightning/ldk.rs
+++ b/gateway/ln-gateway/src/lightning/ldk.rs
@@ -15,6 +15,7 @@ use ldk_node::lightning::ln::msgs::SocketAddress;
 use ldk_node::lightning::ln::PaymentHash;
 use ldk_node::lightning::routing::gossip::NodeAlias;
 use ldk_node::payment::{PaymentKind, PaymentStatus, SendingParameters};
+use lightning::ln::channelmanager::PaymentId;
 use lightning::ln::PaymentPreimage;
 use lightning::util::scid_utils::scid_from_parts;
 use lightning_invoice::Bolt11Invoice;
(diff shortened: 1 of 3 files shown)


joschisan (dismissed):

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -321,20 +347,24 @@ impl ILnRpcClient for GatewayLdkClient {
...
                             failure_reason: "LDK payment failed".to_string(),
                         });
                     }
                 }
             }
             fedimint_core::runtime::sleep(Duration::from_millis(100)).await;
-        }
+        };
+
+        drop(payment_lock_guard);
```
m1sterc001guy, line 365: nit: You don't need to explicitly drop it here, it will drop when it falls out of scope.

m1sterc001guy:
Is the plan to handle the polling in a separate PR? I agree that this is more pressing

tvolk131:
> Is the plan to handle the polling in a separate PR? I agree that this is more pressing

Yep, exactly

gateway/ln-gateway/src/lightning/ldk.rs
tvolk131: Fixed.

m1sterc001guy (approved):

elsirion (approved):
Very elegant!

m1sterc001guy (approved):

tvolk131:
> Do we want this backported to `releases/v0.5`?

We shouldn't release an ldk gateway binary without backporting this. If it isn't too difficult let's backport, otherwise let's not include the ldk gateway in v0.5.

fedimint-backports:
- #6433

PR #6376, opened 2024-11-16
fix: upgrade warn to a panic

joschisan:


dpc (approved):
Time to man up and commit.

PR #6377, opened 2024-11-16
fix: outputs might be empoty due to fees

joschisan:
This is necessary to support federations with non-zero mint fees.

### DIFF
--- a/modules/fedimint-mint-client/src/lib.rs
+++ b/modules/fedimint-mint-client/src/lib.rs
@@ -1101,8 +1101,6 @@ impl MintClientModule {
             })]
         });
 
-        assert!(!outputs.is_empty());
-
         ClientOutputBundle::new(
             outputs,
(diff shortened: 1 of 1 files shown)


bradleystachurski (approved):
[nit] Typo in the commit message: `s/empoty/empty`

fedimint-backports:
- #6379

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1101,8 +1101,6 @@ impl MintClientModule {
             })]
         });
 
-        assert!(!outputs.is_empty());
```
elsirion, line 1104: Do we still spawn SMs?

elsirion: @dpc: the SM generator isn't called on empty bundles

PR #6382, opened 2024-11-17
Fix wallet tx size limit

joschisan:


fedimint-recoverytool/src/main.rs
```diff
@@ -264,15 +264,17 @@ fn input_tweaks_and_peg_out_count(
...
-                        .maybe_v0_ref()
-                        .expect("recoverytool only supports v0 wallet inputs")
-                        .0
-                        .tweak_contract_key()
-                        .serialize(),
+                    {
+                        WalletInput::V0(input) => input.tweak_contract_key().serialize(),
+                        WalletInput::V1(input) => input.tweak_contract_key.serialize(),
+                        WalletInput::Default { .. } => {
+                            panic!("recoverytool only supports v0 wallet inputs")
```
elsirion, line 275: nit: not correct anymore

modules/fedimint-wallet-common/src/lib.rs
```diff
@@ -358,6 +359,7 @@ pub struct WalletInputV0(pub Box<PegInProof>);
 pub struct WalletInputV1 {
     pub outpoint: bitcoin::OutPoint,
     pub tweak_contract_key: secp256k1::PublicKey,
+    pub script_pubkey: ScriptBuf,
```
elsirion, line 362: I guess by having some redundant data here we can make recovery easier? Might be worth it, but then we should check that the script pub key we submit is consistent with our DB when verifying inputs, so that later we don't accidentally depend on unverified data (when changing consensus in the future).

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -274,7 +284,11 @@ pub(crate) async fn transition_btc_tx_confirmed(
 
     let amount = pegin_proof.tx_output().value.into();
 
-    let wallet_input = WalletInput::new_v1(&pegin_proof);
+    let wallet_input = if consensus_version >= ModuleConsensusVersion::new(2, 2) {
+        WalletInput::new_v1(&pegin_proof)
+    } else {
+        WalletInput::new_v0(pegin_proof)
+    };
```
elsirion, line 291: Once we decommission the v0 input type in the client the fn signature should change such that it doesn't require a pegin proof anymore, but only the necessary data.

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -627,6 +628,10 @@ impl WalletClientModule {
         ))
     }
 
+    pub async fn btc_tx_has_no_size_limit(&self) -> FederationResult<bool> {
+        Ok(self.module_api.module_consensus_version().await? >= ModuleConsensusVersion::new(2, 2))
+    }
```
elsirion, line 633: Later optimization: cache result and use the cached version everywhere. That way this fn call becomes cheap enough to run whenever a user wants to receive money and integrators have to decide if they want to show an on-chain address or not.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -786,6 +803,22 @@ impl ServerModule for Wallet {
                     Ok(module.consensus_module_consensus_version(&mut context.dbtx().into_nc()).await)
                 }
             },
+            api_endpoint! {
+                ACTIVATE_CONSENSUS_VERSION_VOTING_ENDPOINT,
+                ApiVersion::new(0, 0),
```
elsirion, line 808: Probably not API version 0.0, but 0.1 or even higher?

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -439,9 +449,16 @@ impl ServerModule for Wallet {
...
+        if dbtx
+            .get_value(&ConsensusVersionVotingActivationKey)
+            .await
+            .is_some()
+            || is_running_in_test_env()
+        {
+            items.push(WalletConsensusItem::ModuleConsensusVersion(
+                MODULE_CONSENSUS_VERSION,
+            ));
+        }
```
elsirion, line 461: We should stop voting once consensus is achieved imo, otherwise we are just wasting consensus bandwidth.

elsirion (commented):
Looks great! Some comments are fixed in later commits.

One thing that I'd like to add: automatic voting start if all peers support the version. That way guardians will automatically get the feature once everyone upgrades and require less hand holding.

elsirion:
What do you think about also forbidding RBF tx with this consensus version update and removing the old hack?

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -512,6 +518,25 @@ impl ServerModule for Wallet {
                     dbtx.remove_entry(&UnsignedTransactionKey(txid)).await;
                 }
             }
+            WalletConsensusItem::ModuleConsensusVersion(module_consensus_version) => {
+                let current_vote = dbtx
+                    .get_value(&ConsensusVersionVoteKey(peer))
+                    .await
+                    .unwrap_or(ModuleConsensusVersion::new(2, 0));
```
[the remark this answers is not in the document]
joschisan: Already bumped in later commit.

fedimint-testing/src/btc/mock.rs
```diff
@@ -337,6 +337,18 @@ impl IBitcoindRpc for FakeBitcoinTest {
...
+    async fn get_block(&self, block_hash: &bitcoin::BlockHash) -> BitcoinRpcResult<bitcoin::Block> {
+        let blocks = &self.inner.read().unwrap().blocks;
+
+        for block in blocks {
+            if block.header.block_hash() == *block_hash {
+                return Ok(block.clone());
+            }
+        }
+
+        bail!("No block with that hash found");
```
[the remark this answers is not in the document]
joschisan: Done.

fedimint-testing/src/btc/mock.rs
```diff
@@ -337,6 +337,18 @@ impl IBitcoindRpc for FakeBitcoinTest {
             .block_hash())
     }
 
+    async fn get_block(&self, block_hash: &bitcoin::BlockHash) -> BitcoinRpcResult<bitcoin::Block> {
+        let blocks = &self.inner.read().unwrap().blocks;
```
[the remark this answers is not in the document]
joschisan: No idea. It works the same way in get_block_hash above.

modules/fedimint-wallet-common/src/lib.rs
joschisan: Yes.

modules/fedimint-wallet-server/src/lib.rs
joschisan: The bandwidth consumption is negligent. Optimizing this is would therefore be confusing as it suggests that its not.

joschisan:
> What do you think about also forbidding RBF tx with this consensus version update and removing the old hack?

Yes but different PR.

fedimint-testing/src/btc/mock.rs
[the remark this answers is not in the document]
elsirion: And that's why `expect`s are superior :rofl:

elsirion (approved):
> One thing that I'd like to add: automatic voting start if all peers support the version. That way guardians will automatically get the feature once everyone upgrades and require less hand holding.

Would still like to figure this out, but can be a separate PR.

@bradleystachurski could you take another look please since it's a consensus change?

modules/fedimint-wallet-server/src/lib.rs
dpc: Yeah, this does not look right. We should bump the minor api version and put that version here. @joschisan

dpc:
@joschisan @elsirion

We'll regret copy&pasting consensus voting inside every module, instead of making it general functionality. All it really saves is passing the exact version from the core to the module. Other than this the logic and all the constrains are exactly the same, simplifying nothing, just sitting (IMO) on the wrong side.

Having it here cements architectural decision that consensus voting is an internal of the module, and it will make near impossible to reconcile with general purpose voting in the future. It obscures the process of voting, and current consensus version, making it hard (impossible to do cleanly) to generically expose voting outside. 

The idea that this will be the only module that will ever need it, is just wishful thinking. From now on, every module that needs to do this, will basically need to copy&paste and maintain it separately. Possibly the first one will be Fedi's stabilitypool module.

It's fine with me, but I reserve the right to say "I told you so" in the future. :D

elsirion:
I think the moment we need it for a second module we should generalize. For me the argument for going this way for now is that we can try it out and see if there are any problems that we'd want to avoid in a final version implemented in core. 

It's pretty clear to me this version will be decommissioned sooner or later and I'd be fine with building a core one now. But balancing everything against each other I don't feel strongly enough to fight with @joschisan about it 😅

Implementing it in core would also mean some slightly annoying module API extensions, idk how much not wanting to deal with the module system plays into your aversion to generalized consensus version voting @joschisan? Maybe that's solvable by someone else building the necessary infra?

modules/fedimint-wallet-client/src/deposit.rs
```diff
@@ -249,15 +250,24 @@ async fn await_btc_transaction_confirmed(
 
         debug!(proof_block_hash = ?txout_proof.block_header.block_hash(), "Generated merkle proof");
 
-        return txout_proof;
+        let consensus_version = match global_context.module_api().module_consensus_version().await {
```
bradleystachurski, line 253: This codepath is no longer used once we introduced the peg-in monitor in v0.4.0. It existed only for allowing clients that upgraded with active state machines to gracefully finish, so we should remove this logic.

This is more of an FYI and I don't believe this change is harmful.

Tracking the cleanup in https://github.com/fedimint/fedimint/issues/6675

modules/fedimint-wallet-client/src/pegin_monitor.rs
```diff
@@ -421,7 +428,11 @@ async fn claim_peg_in(
 
         let amount = pegin_proof.tx_output().value.into();
 
-        let wallet_input = WalletInput::new_v0(pegin_proof);
+        let wallet_input = if consensus_version >= ModuleConsensusVersion::new(2, 2) {
```
bradleystachurski, line 431: Capturing conversations out-of-band: the upgrade tests detected an edge case. Relying only on the module consensus version doesn't guarantee that the federation is aware of the UTXO. More context and fix in https://github.com/fedimint/fedimint/pull/6687

bradleystachurski (commented):
Overall, looks great. Thanks for tackling this while I was occupied with the v0.5 debugging and release :slightly_smiling_face: 

I introduced a fix and a few cleanups in https://github.com/fedimint/fedimint/pull/6687.

> What do you think about also forbidding RBF tx with this consensus version update and removing the old hack?

Capturing in https://github.com/fedimint/fedimint/issues/6688

@joschisan Did you already have the RBF changes lined up? If not, I can grab it.

PR #6393, opened 2024-11-18
chore: inline collection en/decoding helpers

tvolk131:
Contains no functional changes, just inlining functions that are only used once

### DIFF
--- a/fedimint-core/src/encoding/mod.rs
+++ b/fedimint-core/src/encoding/mod.rs
@@ -341,95 +341,6 @@ macro_rules! impl_encode_decode_tuple {
     );
 }
 
-/// Specialized version of Encodable for bytes
-pub fn consensus_encode_bytes<W: std::io::Write>(
-    bytes: &[u8],
-    writer: &mut W,
-) -> Result<usize, Error> {
-    let mut len = 0;
-    len += (bytes.len() as u64).consensus_encode(writer)?;
-    writer.write_all(bytes)?;
-    len += bytes.len();
-    Ok(len)
-}
-
-/// Specialized version of Encodable for static byte arrays
-pub fn consensus_encode_bytes_static<const N: usize, W: std::io::Write>(
-    bytes: &[u8; N],
-    writer: &mut W,
-) -> Result<usize, Error> {
-    writer.write_all(bytes)?;
-    Ok(bytes.len())
-}
-
-struct ReadBytesFromFiniteReaderOpts {
-    len: usize,
-    chunk_size: usize,
-}
-
-/// Read `opts.len` bytes from reader, where `opts.len` could potentially be
-/// malicious.
-///
-/// Adapted from <https://github.com/rust-bitcoin/rust-bitcoin/blob/e2b9555070d9357fb552e56085fb6fb3f0274560/bitcoin/src/consensus/encode.rs#L659>
-#[inline]
-fn read_bytes_from_finite_reader<D: Read + ?Sized>(
-    d: &mut D,
-    mut opts: ReadBytesFromFiniteReaderOpts,
-) -> Result<Vec<u8>, io::Error> {
-    let mut ret = vec![];
-
-    assert_ne!(opts.chunk_size, 0);
-
-    while opts.len > 0 {
-        let chunk_start = ret.len();
-        let chunk_size = core::cmp::min(opts.len, opts.chunk_size);
-        let chunk_end = chunk_start + chunk_size;
-        ret.resize(chunk_end, 0u8);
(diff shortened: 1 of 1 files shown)


fedimint-core/src/encoding/mod.rs
```diff
@@ -341,95 +341,6 @@ macro_rules! impl_encode_decode_tuple {
...
-    writer.write_all(bytes)?;
-    Ok(bytes.len())
-}
-
-struct ReadBytesFromFiniteReaderOpts {
-    len: usize,
-    chunk_size: usize,
-}
-
-/// Read `opts.len` bytes from reader, where `opts.len` could potentially be
```
tvolk131, line 370: Note for reviewer: I removed the mention of potential maliciousness since it isn't relevant here because we're always decoding data that we previously encoded.

fedimint-core/src/encoding/mod.rs
```diff
@@ -341,95 +341,6 @@ macro_rules! impl_encode_decode_tuple {
...
-        ret.resize(chunk_end, 0u8);
-        d.read_exact(&mut ret[chunk_start..chunk_end])?;
-        opts.len -= chunk_size;
-    }
-
-    Ok(ret)
-}
-
-/// Specialized version of Decodable for bytes
-pub fn consensus_decode_bytes<D: std::io::Read>(r: &mut D) -> Result<Vec<u8>, DecodeError> {
```
tvolk131, line 396: Note for reviewer: This was removed, not inlined. It was not previously used anywhere.

fedimint-core/src/encoding/mod.rs
```diff
@@ -341,95 +341,6 @@ macro_rules! impl_encode_decode_tuple {
...
-    let opts = ReadBytesFromFiniteReaderOpts {
-        len,
-        chunk_size: 64 * 1024,
-    };
-
-    read_bytes_from_finite_reader(r, opts).map_err(DecodeError::from_err)
-}
-
-/// Specialized version of Decodable for fixed-size byte arrays
-pub fn consensus_decode_bytes_static<const N: usize, D: std::io::Read>(
```
tvolk131, line 418: Note for reviewer: This was removed, not inlined. It was not previously used anywhere.

dpc (commented):
I'm not sure if I like it. :D

tvolk131:
> I'm not sure if I like it. :D

Lol feel free to close if you're vehemently against this

dpc:
We need a second opinion.

elsirion:
What's the goal? I don't see a huge difference in readability (some become more and some less readable imo). Performance also shouldn't be much different.

tvolk131:
> What's the goal?

Less code! And I find it easler to understand how lists/vecs are encoded and decoded when I don't have to jump around through a bunch of helper functions and the unneeded `ReadBytesFromFiniteReaderOpts` struct.

fedimint-core/src/encoding/mod.rs
dpc: Removing unused code :+1:

fedimint-core/src/encoding/mod.rs
dpc: Removing unused code :+1:

fedimint-core/src/encoding/mod.rs
```diff
@@ -564,10 +499,9 @@ where
     fn consensus_encode<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, std::io::Error> {
         if TypeId::of::<T>() == TypeId::of::<u8>() {
             // unsafe: we've just checked that T is `u8` so the transmute here is a no-op
-            return consensus_encode_bytes_static(
-                unsafe { mem::transmute::<&[T; SIZE], &[u8; SIZE]>(self) },
-                writer,
-            );
+            let bytes = unsafe { mem::transmute::<&[T; SIZE], &[u8; SIZE]>(self) };
```
dpc, line 502: OK. Inlining might be better.

maan2003: FYI this might also be transmuting lifetimes

fedimint-core/src/encoding/mod.rs
```diff
@@ -578,14 +512,6 @@ where
     }
 }
 
-// From <https://github.com/rust-lang/rust/issues/61956>
-unsafe fn horribe_array_transmute_workaround<const N: usize, A, B>(mut arr: [A; N]) -> [B; N] {
```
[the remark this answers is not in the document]
tvolk131: Done.

dpc (approved):

Issue #6398, opened 2024-11-19
Nice way to profile cpu usage


dpc:
Run the test suite like this:

```
env FM_TEST_CI_ALL_TIMES=5 j test-ci-all
```

potentially use `CARGO_PROFILE=release` if you want to profile release build.

Make sure to install better `addr2line` with `cargo install addr2line --features=bin`.

When the tests are running for some time, use `perf record` on a highest cpu-usage `fedimintd` instance with:

```
perf record --call-graph dwarf -p $(ps -eo pid,cmd,%cpu --sort=-%cpu | grep fedimintd | grep -v grep | head -n 1 | awk '{print $1}')
```

Keep it running for at least couple of seconds, or until it ends and use `perf report --addr2line=$HOME/.cargo/bin/addr2line` to view the results.

dpc:
In case you wonder:

![image](https://github.com/user-attachments/assets/5f98de39-6c23-430a-a107-1cf734393cc8)

dpc:
Closing as non-actinable. Opened just to write it down for the future / to share.

PR #6399, opened 2024-11-19
refactor(tbs): paralellize aggregate_public_key_shares

dpc:
Re #6398

### DIFF
--- a/crypto/tbs/src/lib.rs
+++ b/crypto/tbs/src/lib.rs
@@ -199,7 +199,7 @@ pub fn aggregate_public_key_shares(shares: &BTreeMap<u64, PublicKeyShare>) -> Ag
 
 fn lagrange_multipliers(scalars: Vec<Scalar>) -> Vec<Scalar> {
     scalars
-        .iter()
+        .par_iter()
         .map(|i| {
             scalars
                 .iter()
--- a/crypto/tbs/src/lib.rs
+++ b/crypto/tbs/src/lib.rs
@@ -14,6 +14,7 @@ use hex::encode;
 use rand::rngs::OsRng;
 use rand::SeedableRng;
 use rand_chacha::ChaChaRng;
+use rayon::iter::{IntoParallelRefIterator as _, ParallelIterator as _};
 use serde::{Deserialize, Serialize};
 use sha3::Digest;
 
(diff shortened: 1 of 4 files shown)


dpc:
I need to bench it thorougly. 

But first, @joschisan does the math check out?

crypto/tbs/src/lib.rs
```diff
@@ -188,9 +189,10 @@ pub fn aggregate_public_key_shares(shares: &BTreeMap<u64, PublicKeyShare>) -> Ag
         lagrange_multipliers(shares.keys().cloned().map(Scalar::from).collect())
```
dpc, line 189: Hmmm... Might need to do something about this one too...

dpc:
I am dpc's dissapointment

```
> env CARGO_PROFILE=release ./scripts/tests/latency-test.sh fm-pay


Baseline (master):

 LATENCY FM PAY: min: 1.0s, avg: 1.1s, median: 1.1s, p90: 1.4s, max: 1.4s, sum: 11.5s



para aggregate_public_key_shares:

 LATENCY FM PAY: min: 0.9s, avg: 1.6s, median: 1.1s, p90: 6.2s, max: 6.2s, sum: 15.8s



para aggregate_public_key_shares + lagrange_multipliers:

 LATENCY FM PAY: min: 1.0s, avg: 1.1s, median: 1.1s, p90: 1.7s, max: 1.7s, sum: 11.2s



para lagrange_multipliers:

 LATENCY FM PAY: min: 1.0s, avg: 1.1s, median: 1.1s, p90: 1.4s, max: 1.4s, sum: 11.2s
```

dpc:
This is probably better done on a higher level (e.g. per note) and maybe already interferes with it.

elsirion:
We barely ever run this fn, afaik it's only required after DKG. I see that we also run it to convert mint server config to mint client config, but that could be easily cached. There's no need to optimize something that runs in a second and shouldn't be called more than once in the first place.

dpc:
> We barely ever run this fn, afaik it's only required after DKG

Hmmm... Possibly I've seen it a lot in the profiling, because we run DKG on start of every test.

elsirion:
If it would massively speed up CI might be worth it, but I don't think we start enough federations sequentially for that.

PR #6401, opened 2024-11-19
chore: require saturating or checked subtraction for Amount

tvolk131:
Remove implementation of `std::ops::Sub` for `Amount`, requiring the usage of `Amount::saturating_sub()` or `Amount::checked_sub()` instead


dpc (approved):
Great idea.

elsirion (approved):
Should we do the same for add etc?

tvolk131:
> Should we do the same for add etc?

I'll make a follow-up PR

PR #6407, opened 2024-11-19
fix(ci): some bundlers we use expect a missing `pname`

dpc:
Fix https://github.com/fedimint/fedimint/actions/runs/11922340790/job/33228414053


fedimint-backports:
- #6420

Issue #6218, opened 2024-10-24
Removing Core Lightning Gateway Support


m1sterc001guy:
I believe we have decided to not support a Core Lightning Gateway after v0.5. We will still be keeping `CLN` testing support in devimint. This issue outlines the steps to remove it from our codebase.

- [x] v0.5 Release
- [x] Remove dependence on gRPC generated types https://github.com/fedimint/fedimint/pull/6197
- [ ] Remove any gateway specific commands from devimint and replace them with Core Lightning RPCs
- [ ] Remove CLN Gateway and extension from devimint and from tests.

PR #6414, opened 2024-11-20
chore: Remove Core Lightning Gateway

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/6218

It is a sad day, but we have decided to retire the Core Lightning Gateway from devimint and discontinue supporting it. We are facing issues upgrading to the latest version on mac and with the lack of adoption, it is probably better focusing our efforts elsewhere.

This PR needs to make a decent amount of changes to the LNv1 tests to get them to compile. I don't believe the coverage is much worse than it was before. We tested almost everything with both the LND and CLN gateways, I believe with the one exception around HOLD invoices.

I personally think we should add another LDK Gateway in its place, but that can be a follow up.


joschisan:
Indeed, very sad!

elsirion:
🪦 :cry: 

For anyone running a CLN gateway: you can just stick with v0.5 for now, which should be good for another year. Eventually I want to implement a plugin that exposes the relevant gRPC APIs such that the gateway can connect to CLN as it would to LND.

dpc:
> Indeed, very sad!

image

joschisan (approved):

joschisan (approved):

PR #6424, opened 2024-11-20
chore: require checked/saturating operations on Amount

tvolk131:
Follow-up to #6401

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -1080,8 +1080,8 @@ impl Client {
                 "We only build transactions with input versions that are supported by the module",
             );
 
-            in_amount += input.amount;
-            fee_amount += item_fee;
+            in_amount.saturating_add_to(input.amount);
+            fee_amount.saturating_add_to(item_fee);
         }
 
         for output in builder.outputs() {
@@ -1091,11 +1091,11 @@ impl Client {
                 "We only build transactions with output versions that are supported by the module",
             );
 
-            out_amount += output.amount;
-            fee_amount += item_fee;
+            out_amount.saturating_add_to(output.amount);
+            fee_amount.saturating_add_to(item_fee);
         }
 
-        (in_amount, out_amount + fee_amount)
+        (in_amount, out_amount.saturating_add(fee_amount))
     }
 
     pub fn get_internal_payment_markers(&self) -> anyhow::Result<(PublicKey, u64)> {
@@ -1450,9 +1450,10 @@ impl Client {
         let mut amount = Amount::ZERO;
 
         for out_point in outputs {
-            amount += self
-                .await_primary_module_output(operation_id, out_point)
-                .await?;
+            amount.saturating_add_to(
+                self.await_primary_module_output(operation_id, out_point)
+                    .await?,
+            );
         }
 
         Ok(amount)
--- a/fedimint-core/src/amount.rs
+++ b/fedimint-core/src/amount.rs
@@ -71,18 +71,6 @@ impl Amount {
         Ok(Self::from(btc_amt))
     }
 
-    pub fn saturating_sub(self, other: Self) -> Self {
-        Self {
-            msats: self.msats.saturating_sub(other.msats),
-        }
-    }
-
-    pub fn mul_u64(self, other: u64) -> Self {
-        Self {
-            msats: self.msats * other,
-        }
-    }
-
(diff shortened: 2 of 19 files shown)


fedimint-core/src/amount.rs
tvolk131, line 1: Yikes... The diff looked much cleaner on my local machine. Not sure how to fix it...

fedimint-core/src/amount.rs
```diff
@@ -105,96 +93,58 @@ impl Amount {
...
-            msats: self.msats + rhs.msats,
+            msats: self.msats.saturating_mul(other),
         }
     }
-}
 
-impl std::ops::AddAssign for Amount {
-    fn add_assign(&mut self, rhs: Self) {
-        *self = *self + rhs;
+    pub fn checked_div(self, other: Self) -> Option<u64> {
```
tvolk131, line 136: Note for reviewer: We're not adding `saturating_div()` because, for unsigned integers, it would be impossible to overflow

fedimint-testing/src/btc/mock.rs
```diff
@@ -267,35 +267,33 @@ impl BitcoinTest for FakeBitcoinTest {
...
+                .expect("previous transaction should be known")
+        }));
 
-            return fee;
-        }
+        let total_output =
+            Amount::saturating_sum(tx.output.iter().map(|output| output.value.into()));
+
+        total_input.saturating_sub(total_output)
     }
```
tvolk131, line 297: Note for reviewer: I did some refactoring while I was here. The fee calculation always returns, so there's no need to have it be inside the loop.

fedimint-core/src/amount.rs
```diff
@@ -105,96 +93,58 @@ impl Amount {
...
-        self.msats %= rhs.msats;
-    }
-}
-
-impl std::ops::Div for Amount {
-    type Output = u64;
 
-    fn div(self, rhs: Self) -> Self::Output {
-        self.msats / rhs.msats
+    pub fn saturating_add_to(&mut self, other: Self) {
```
dpc, line 108: This variant seems unnecessary, standard library doesn't have anything like that. A bit more verbosity is OK, IMO.

fedimint-client/src/lib.rs
```diff
@@ -1080,8 +1080,8 @@ impl Client {
                 "We only build transactions with input versions that are supported by the module",
             );
 
-            in_amount += input.amount;
-            fee_amount += item_fee;
+            in_amount.saturating_add_to(input.amount);
+            fee_amount.saturating_add_to(item_fee);
```
elsirion, line 1084: Saturating is incorrect here imo. We can just make the fn fallible. If we ever overflow the tx will be rejected by the federation anyway.

fedimint-core/src/amount.rs
```diff
@@ -105,96 +93,58 @@ impl Amount {
...
-impl std::ops::Mul<u64> for Amount {
-    type Output = Self;
-
-    fn mul(self, rhs: u64) -> Self::Output {
+    pub fn saturating_sub(self, other: Self) -> Self {
         Self {
-            msats: self.msats * rhs,
+            msats: self.msats.saturating_sub(other.msats),
         }
     }
```
elsirion, line 122: Do we *ever* want saturating behavior for amounts?

elsirion: Going back to the previous PR there might be a place for it in tests, although even there panics would typically be better, just more verbose.

dpc: Yeah, when adding values, saturation is almost never what is needed.

We probably only need `checked_add` (result) and `expected_add` (panics)?

dpc:
*Comment by Tau/GPT-5*

Triage note: I think this PR should be closed. It was a follow-up to #6401, but the review feedback here converged that saturating arithmetic is generally not the right default for `Amount` additions; those call sites should be made fallible or explicitly panic/expect based on context.

The branch is also from 2024 and now conflicts with current `master`. Some pieces have moved on independently (for example, `Amount::checked_mul` now exists), while newer mintv2/walletv2/lnv2 code would need a fresh audit that this PR does not cover. Merging this as-is would encode the rejected saturating semantics and miss newer code.

If we still want this policy, I would suggest a fresh PR that removes unchecked `Amount` ops and updates call sites to checked/fallible or expect-style helpers according to context. Closing this stale PR for now; it can always be reopened if we missed something.

PR #6426, opened 2024-11-20
chore(ci): use determinate's system nix installer

dpc:
Maybe it will work better.

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -45,10 +45,11 @@ jobs:
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


elsirion:
Why exactly do we need this at all now that we are running on our own nix machines?

dpc:
> Why exactly do we need this at all now that we are running on our own nix machines?

If a step needs Nix, it's easier to just put it there, and have it do nothing when running on self-hosted runners. E.g. we run MacOS steps on Github Actions machines, and making stuff like that conditional is just extra work.

dpc:
Anyway, this action seems completely incompatible with self-hosted runners. I reported it upstream.

PR #6429, opened 2024-11-21
feat: initial implementation of walletv2

joschisan:


modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,497 @@
...
+
+#[derive(Error, Debug, Clone, Eq, PartialEq)]
+pub enum SendError {
+    #[error("Address is from a different network then the federation.")]
+    WrongNetwork,
+    #[error("The amount is to small to be send on-chain")]
+    DustAmount,
+    #[error("Federation returned an error: {0}")]
+    FederationError(String),
+    #[error("The federation has no vetted gateways")]
```
douglaz, line 448: wrong message?

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,497 @@
...
+    NoConsensusFeerateAvailable,
+    #[error("The client does not have sufficicent funds to send the payment")]
```
douglaz, line 450:
```suggestion
    #[error("The client does not have sufficient funds to send the payment")]
```

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,497 @@
...
+
+        Ok(operation_id)
+    }
+}
+
+#[derive(Error, Debug, Clone, Eq, PartialEq)]
+pub enum SendError {
+    #[error("Address is from a different network then the federation.")]
+    WrongNetwork,
+    #[error("The amount is to small to be send on-chain")]
```
douglaz, line 444:
```suggestion
    #[error("The amount is too small to be sent on-chain")]
```

modules/fedimint-walletv2-common/src/config.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// The mint module will charge a non-configurable base fee of one thousand
+    /// satoshis per transaction input and output to account for the costs
+    /// incurred by the federation for processing the transaction. On top of
+    /// that the federation may charge a additional relative fee per input and
+    /// output of up to ten thousands parts per million which equals one
+    /// percent.
+    ///
+    /// # Errors
+    /// - This constructor returns an error if the relative fee is in excess of
+    ///   one thousand parts per million.
```
douglaz, line 183: ten thousand?

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub value: bitcoin::Amount,
+    pub outpoint: bitcoin::OutPoint,
+    pub tweak: sha256::Hash,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
+pub struct SendFee {
+    // This index ensures that once a Fedimint transaction becomes invalid due to a
+    // change in feerate it stays invalid. Otherwise there exists a race condition
+    // where a client might consider a actually accepted transaction as rejected.
```
douglaz, line 121:
```suggestion
    // change in feerate, it remains invalid. Otherwise, there exists a race condition
    // where a client might consider an actually accepted transaction as rejected.
```

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub index: u64,
+    // The absolute fee currently required to send a payment on-chain.
+    pub value: bitcoin::Amount,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize, Encodable, Decodable)]
+pub struct ReceiveFee {
+    // This index ensures that once a Fedimint transaction becomes invalid due to a
+    // change in feerate it stays invalid. Otherwise there exists a race condition
```
douglaz, line 130: ditto

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    UnknownInputVariant(#[from] UnknownWalletInputVariantError),
+    #[error("The peg-in was already claimed")]
+    UnspentTxOutAlreadySpent,
+    #[error("Unknown UTXO")]
+    UnknownUnspentTxOut,
+    #[error("Wrong output script")]
+    WrongOutputScript,
+    #[error("No up to date feerate is available at the moment. Please try again later.")]
+    NoConsensusFeerateAvailable,
+    #[error("The fee index is outdated. Please construct an new transaction.")]
```
douglaz, line 263:
```suggestion
    #[error("The fee index is outdated. Please construct a new transaction.")]
```

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    IncorrectFeeRateIndex,
+    #[error("The total transaction fee is to low. Please construct an new transaction.")]
```
douglaz, line 265:
```suggestion
    #[error("The total transaction fee is too low. Please construct a new transaction.")]
```

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    InsufficientTotalFee,
+    #[error("Constructing the pegin transaction caused and arithmetic overflow")]
```
douglaz, line 267:
```suggestion
    #[error("Constructing the pegin transaction caused an arithmetic overflow")]
```

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    ArithmeticOverflow,
+}
+
+#[derive(Debug, Error, Encodable, Decodable, Hash, Clone, Eq, PartialEq)]
+pub enum WalletOutputError {
+    #[error("The wallet output version is not supported by this federation")]
+    UnknownOutputVariant(#[from] UnknownWalletOutputVariantError),
+    #[error("Connected bitcoind is on wrong network.")]
```
douglaz, line 275: wrong message

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    UnderDustLimit,
+    #[error("The federation does not have any funds yet")]
+    NoFederationUTXO,
+    #[error("No up to date feerate is available at the moment. Please try again later.")]
+    NoConsensusFeerateAvailable,
+    #[error("The fee index is outdated. Please construct an new transaction.")]
```
douglaz, line 281: ...

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    IncorrectFeeRateIndex,
+    #[error("The total transaction fee is to low. Please construct an new transaction.")]
```
douglaz, line 283: ...

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    InsufficicentTotalFee,
```
douglaz, line 284:
```suggestion
    InsufficientTotalFee,
```

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1147 @@
...
+                        }
+                    }
+                }
+            }
+            WalletConsensusItem::Feerate(feerate) => {
+                let old_median = self.consensus_feerate_median(dbtx).await;
+
+                if Some(feerate) == dbtx.insert_entry(&FeeRateVoteKey(peer), &feerate).await {
+                    bail!("Fee rate vote is redundant");
+                }
```
douglaz, line 390:
```suggestion
                ensure!(
                    Some(feerate) != dbtx.insert_entry(&FeeRateVoteKey(peer), &feerate).await,
                    "Fee rate vote is redundant"
                );
```

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1147 @@
...
+        let pending_count = dbtx
+            .find_by_prefix(&PendingTransactionPrefix)
+            .await
+            .count()
+            .await;
+
+        // Since the total number of transactions is bound the following addition
+        // and cast cannot overflow or truncate.
+
+        (unsigned_count + pending_count) as u32
```
douglaz, line 958:
```suggestion
        (unsigned_count + pending_count).try_into().expect("Since the total number of transactions is bound the addition cannot overflow or truncate")
```

douglaz:
Concept NACK if the goal is to replace the "wallet v1" module.

The primary issue here is that creating a single UTXO for the federation effectively and unavoidably turns the federation into a "mixer" by default.

While a mixer might be an interesting project from a cypherpunk perspective, it poses significant risks for guardians and gateway operators, which could hinder its adoption by most real-world communities. Additionally, it may attract unwanted attention to the project and its developers.

My main objection isn't to the existence of this module, but to it being the standard or default for fedimints. The standard module should have reasonable UTXO management suitable for all use cases. At a minimum, it should avoid unnecessary mixing of UTXOs.

nix/check-forbidden-deps.sh
```diff
@@ -6,7 +6,7 @@ if grep -E "(fedimint-mint|fedimint-wallet|fedimint-ln)" fedimint-server/Cargo.t
   >&2 echo "fedimint-server/Cargo.toml must not depend on modules"
   return 1
 fi
-if grep -E "(fedimint-mint|fedimint-wallet|fedimint-ln-(server|client))" fedimint-testing/Cargo.toml >&2 ; then
+if grep -E "(fedimint-mint|fedimint-wallet-|fedimint-ln-(server|client))" fedimint-testing/Cargo.toml >&2 ; then
```
dpc, line 9: v2 module is also a module ... IIRC this rule is to prevent circular deps.

devimint/src/federation/config.rs
```diff
@@ -101,6 +101,24 @@ pub fn attach_default_module_init_params(
         );
     }
 
+    if fedimintd_version >= &VERSION_0_5_0_ALPHA
+        && !is_env_var_set(FM_DEVIMINT_DISABLE_MODULE_LNV2_ENV)
```
dpc, line 105: s/LNV2/WALLETV2/?

fedimint-testing/src/btc/mock.rs
```diff
@@ -65,7 +68,7 @@ struct FakeBitcoinTestInner {
     /// Simulates the merkle tree proofs
     proofs: BTreeMap<Txid, TxOutProof>,
     /// Simulates the script history
-    scripts: BTreeMap<ScriptBuf, Vec<Transaction>>,
+    scripts: BTreeMap<ScriptBuf, Vec<(Transaction, u64)>>,
```
dpc, line 71: Oh, it's a block height? Hard to tell from looking at it. Either a comment, or maybe

`type BlockHeight = u64;` and `scripts: BTreeMap<ScriptBuf, Vec<(Transaction, BlockHeight)>>`?

fedimint-testing/src/btc/mock.rs
```diff
@@ -222,18 +225,26 @@ impl BitcoinTest for FakeBitcoinTest {
...
         };
+
         inner
             .proofs
             .insert(transaction.compute_txid(), proof.clone());
-        inner
-            .scripts
-            .insert(address.script_pubkey(), vec![transaction.clone()]);
+
+        if let Some(txs) = inner.scripts.get_mut(&address.script_pubkey()) {
```
dpc, line 240: Nit: I think it can be done cleaner with `inner.scripts.entry(address.script_pubkey()).or_default().push((transaction.clone(), block_height));

fedimintd/src/fedimintd.rs
```diff
@@ -368,6 +368,27 @@ impl Fedimintd {
             s
         };
 
+        let s = if is_env_var_set(FM_ENABLE_MODULE_LNV2_ENV) {
```
dpc, line 371: `WALLETV2`?

misc/git-hooks/pre-commit
```diff
@@ -68,7 +68,7 @@ function check_check_forbidden_dependencies() {
       >&2 echo "fedimint-server/Cargo.toml must not depend on modules"
       return 1
     fi
-    if grep -E "(fedimint-mint|fedimint-wallet|fedimint-ln-(server|client))" fedimint-testing/Cargo.toml >&2 ; then
+    if grep -E "(fedimint-mint|fedimint-wallet-|fedimint-ln-(server|client))" fedimint-testing/Cargo.toml >&2 ; then
```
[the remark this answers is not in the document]
dpc: Ehhh... OK, let's allow `-common`s. Someone is going to put a big fat dependency in one of the `-common`s eventually and ruin everything.

AFAICT, the correct one should be: `((fedimint-mint|fedimint-wallet|fedimint-ln)-(server|client))`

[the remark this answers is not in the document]
dpc: OK, let's just sort it out in another PR: https://github.com/fedimint/fedimint/pull/6573 @joschisan

modules/fedimint-walletv2-client/src/api.rs
```diff
@@ -0,0 +1,81 @@
...
+use fedimint_core::task::{MaybeSend, MaybeSync};
+use fedimint_core::{apply, async_trait_maybe_send};
+use fedimint_walletv2_common::endpoint_constants::{
+    CONSENSUS_BLOCK_COUNT_ENDPOINT, FEDERATION_WALLET_ENDPOINT, FILTER_UNSPENT_OUTPOINTS_ENDPOINT,
+    PENDING_TRANSACTIONS_ENDPOINT, RECEIVE_FEE_ENDPOINT, SEND_FEE_ENDPOINT,
+};
+use fedimint_walletv2_common::{FederationWallet, ReceiveFee, SendFee};
+
+#[apply(async_trait_maybe_send!)]
+pub trait WalletFederationApi {
```
dpc, line 15: Doccommment here would be extremely valuable.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+mod db;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum WalletOperationMeta {
+    Send(SendMeta),
+    Receive(ReceiveMeta),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SendMeta {
```
dpc, line 56: Do we have a consensus on Send/Receive nomenclature? I guess Send == Pegout?

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum FinalOperationState {
+    /// The operation has been successful.
+    Success,
+    /// The operation has been aborted due to a change of the consensus feerate.
+    Aborted,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UnspentDeposit {
```
dpc, line 81: Unspent == Unclaimed? And here it's a Deposit, but above "Receive"? :D

`DepositMeta` seems to make more sense than `ReceiveMeta` in that case. The operations can be "send" and "receive" but when talking about nouns, "deposit" and "withdrawal" sound better, IMO.

dpc: Writting docstring for these would clarify any doubts and help with the naming.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+    esplora_rpc: DynEsploraConnection,
+}
+
+#[derive(Debug, Clone)]
+pub struct WalletClientContext {
+    pub decoder: Decoder,
+}
+
+impl Context for WalletClientContext {
+    const KIND: Option<ModuleKind> = None;
```
dpc, line 109: None?! Shouldn't this be `Some(KIND)`? Other modules set it.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+        cli::handle_cli_command(self, args).await
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct WalletClientInit {
+    pub esplora_connection: DynEsploraConnection,
+}
+
+impl Default for WalletClientInit {
```
dpc, line 155: `impl Default` conceptually "making connections" is a bit odd.

modules/fedimint-walletv2-common/src/esplora_api.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub block_height: Option<u64>,
+}
+
+dyn_newtype_define! {
+    #[derive(Clone)]
+    pub DynEsploraConnection(Arc<IEsploraConnection>)
+}
+
+#[derive(Debug, Clone)]
+pub struct RealEsploraConnection;
```
dpc, line 37: We might want to stuff `reqwest::Client` here to reuse it.

> The Client holds a connection pool internally, so it is advised that you create one and reuse it.

https://docs.rs/reqwest/latest/reqwest/struct.Client.html

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+            .await
+            .map_err(|e| SendError::FederationError(e.to_string()))?
+            .ok_or(SendError::NoConsensusFeerateAvailable)
+            .map(|fee| fee.value)
+    }
+
+    /// Send an on-chain payment with the given fee.
+    pub async fn send(
+        &self,
+        address: &Address<NetworkUnchecked>,
```
dpc, line 234: Conceptually we should be passing `Address` (which is the same as `Address<NetworkChecked>` around. We should only need to deal with `NetworkUnchecked` on the boundaries of our system, check them fast and not have to worry about it anymore. Ping @tvolk131 if that's our approach, or are there any things preventing it.

dpc: Hmm... after noticing the check just below. I guess it makes sense to validate it here, since this is a public method that a downstream might use. Though... maybe we should pass that responsibilty to the user (that's what using `Address<NetworkChecked>` would do). Possibly they already validated it anyway, and they already have it checked.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+            Network::Bitcoin => "https://blockstream.info/api/",
+            Network::Testnet => "https://blockstream.info/testnet/api/",
+            Network::Signet => "https://blockstream.info/signet/api/",
+            _ => return None,
+        };
+
+        Some(SafeUrl::parse(url).expect("Failed to parse default esplora server"))
+    }
+
+    /// Check an address for unspent deposits and return the deposits in
```
dpc, line 369: The format of a docstring comment should be a "single sentence, newlines, longer description". I thought clippy nags if it isn't the case.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+
+                ReceiveFee {
+                    index: receive_fee.index,
+                    value,
+                }
+            }
+            None => receive_fee,
+        };
+
+        if unspent_deposit.value < receive_fee.value + bitcoin::Amount::from_sat(100) {
```
dpc, line 466: Magic constant, maybe make it a constant with a docstring.

modules/fedimint-walletv2-common/src/config.rs
```diff
@@ -0,0 +1,261 @@
...
+                    + change_output_weight
+                    + destination_output_weight,
+            ),
+            receive_tx_vbytes: weight_to_vbytes(
+                tx_overhead_weight
+                    + change_input_weight
+                    + change_input_weight
+                    + change_output_weight,
+            ),
+            dust_limit: bitcoin::Amount::from_sat(10_000),
```
dpc, line 160: Different magic constant.

modules/fedimint-walletv2-common/src/config.rs
```diff
@@ -0,0 +1,261 @@
...
+            + 4 // scriptPubKey length
+            + 34 * 4; // scriptPubKey
+
+        let destination_output_weight = 8 * 4 // nValue
+            + 4 // scriptPubKey length
+            + 34 * 4; // scriptPubKey
+
+        Self {
+            bitcoin_pks,
+            finality_delay: 6,
```
dpc, line 147: Magic constant.

modules/fedimint-walletv2-common/src/config.rs
dpc: Could we express/docuemnt in sats/vB? dpc has a too small brain to convert on the fly.

modules/fedimint-walletv2-common/src/config.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// Bitcoin network (e.g. testnet, bitcoin)
+    pub network: Network,
+}
+
+impl std::fmt::Display for WalletClientConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "WalletClientConfig {}",
+            serde_json::to_string(self).map_err(|_e| std::fmt::Error)?
```
dpc, line 258: Do we really need `Display` on this one? The implementation here looks like of like `Debug` anyway, not very usable for ... "output to the end user", if you ask me, and probably the whole struct shouldn't need it really in the first place. `{:?}` in formatting will produce roughly same thing without `serde_json`.

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+
+            if program.is_p2tr() {
+                return DestinationScript::P2TR(
+                    XOnlyPublicKey::from_slice(program.program().as_bytes())
+                        .expect("Witness program is 32 bytes"),
+                );
+            }
+        }
+
+        panic!("Failed to obtain Destination from address");
```
dpc, line 96: Maybe better as a `Result`?

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1156 @@
...
+                    .await;
+
+                let new_consensus_block_count = self.consensus_block_count(dbtx).await;
+
+                assert!(old_consensus_block_count <= new_consensus_block_count);
+
+                // We do not sync blocks that predate the federation itself.
+                if old_consensus_block_count == 0 {
+                    return Ok(());
+                }
```
tvolk131, line 340: I'm trying to wrap my head around why this is safe. I don't see anything preventing an on-chain receive from occurring before a consensus block is reached. And I see that in some of the tests you put this comment:

```
    // We need the consensus block count to reach a non-zero value before we send in
    // any funds such that the UTXO is tracked by the federation.
```

Are we assuming that the guardians will reach block consensus before inviting anyone to their federation? That seems like a reasonable assumption to make, but we should make that requirement clear.

modules/fedimint-walletv2-tests/bin/tests.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        if fedimint_cli_version < *VERSION_0_5_0_ALPHA {
+            info!(%fedimint_cli_version, "Version did not support walletv2 module, skipping");
+            return Ok(());
+        }
+
+        if fedimintd_version < *VERSION_0_5_0_ALPHA {
+            info!(%fedimintd_version, "Version did not support walletv2 module, skipping");
+            return Ok(());
+        }
```
tvolk131, line 28: Nit: Can we combine these?

```
        if fedimint_cli_version < *VERSION_0_5_0_ALPHA || fedimintd_version < *VERSION_0_5_0_ALPHA {
            info!(%fedimint_cli_version, "Version did not support walletv2 module, skipping");
            return Ok(());
        }
```

tvolk131 (dismissed):

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+impl ModuleInit for WalletInit {
+    type Common = WalletCommonInit;
+
+    async fn dump_database(
+        &self,
+        _dbtx: &mut DatabaseTransaction<'_>,
+        _prefix_names: Vec<String>,
+    ) -> Box<dyn Iterator<Item = (String, Box<dyn erased_serde::Serialize + Send>)> + '_> {
+        todo!()
```
elsirion: will panic

modules/fedimint-walletv2-tests/bin/tests.rs
```diff
@@ -0,0 +1,351 @@
...
+use devimint::{cmd, util};
+use fedimint_core::task::sleep_in_test;
+use fedimint_walletv2_common::TransactionInfo;
+use fedimint_walletv2_server::FINALITY_DELAY;
+use serde::Deserialize;
+use tokio::try_join;
+use tracing::info;
+
+fn bsats(satoshi: u64) -> u64 {
+    satoshi
```
elsirion, line 18: redundant

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+    async fn consensus_proposal<'a>(
+        &'a self,
+        dbtx: &mut DatabaseTransaction<'_>,
+    ) -> Vec<WalletConsensusItem> {
+        let mut items = dbtx
+            .find_by_prefix(&UnsignedTransactionPrefix)
+            .await
+            .map(|(key, unsigned_transaction)| {
+                let signatures = self.sign_transaction(&unsigned_transaction);
```
elsirion: Since `rust-secp256k1` uses RFC6979 we'll always generate the same signature by deriving the nonce using the secret key and the message.

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+            .await;
+
+        if let Some(status) = self.btc_rpc.status() {
+            assert_eq!(status.network, self.cfg.consensus.network);
+
+            let consensus_block_count = self.consensus_block_count(dbtx).await;
+
+            items.push(WalletConsensusItem::BlockCount(min(
+                status.block_count.saturating_sub(FINALITY_DELAY),
+                consensus_block_count + 5,
```
elsirion: nit: should be constant for documentation. the `+5` is to limit syncing too many blocks in one `process_consensus_item` step

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+    }
+
+    async fn process_consensus_item<'a, 'b>(
+        &'a self,
+        dbtx: &mut DatabaseTransaction<'b>,
+        consensus_item: WalletConsensusItem,
+        peer: PeerId,
+    ) -> anyhow::Result<()> {
+        match consensus_item {
+            WalletConsensusItem::BlockCount(block_count_vote) => {
```
elsirion, line 446: nit: could be broken into separate functions

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+                dbtx.insert_entry(&BlockCountVoteKey(peer), &block_count_vote)
+                    .await;
+
+                let new_consensus_block_count = self.consensus_block_count(dbtx).await;
+
+                assert!(old_consensus_block_count <= new_consensus_block_count);
+
+                // We do not sync blocks that predate the federation itself.
+                if old_consensus_block_count == 0 {
```
elsirion: nit: 0 as a special value is not idiomatic rust imo. `None` seems more appropriate.

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+                    return Ok(());
+                }
+
+                // Our bitcoin backend needs to be synced for the following calls to the
+                // get_block rpc to be safe for consensus.
+                self.await_local_sync_to_block_count(new_consensus_block_count + FINALITY_DELAY)
+                    .await;
+
+                for height in old_consensus_block_count..new_consensus_block_count {
```
elsirion: Won't this non-inclusive range lead to us always trailing by one more block than finality delay?

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+                    .await
+                    .expect("Bitcoind rpc to get_block_hash failed");
+
+                    let block = util::retry(
+                        "get_block",
+                        util::backoff_util::background_backoff(),
+                        || self.btc_rpc.get_block(&block_hash),
+                    )
+                    .await
+                    .expect("Bitcoind rpc to get_block failed");
```
elsirion: Why query both? Can't we just calculate the hash?

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+                    let pks_hash = self.cfg.consensus.bitcoin_pks.consensus_hash();
+
+                    for transaction in block.txdata {
+                        dbtx.remove_entry(&PendingTransactionKey(transaction.compute_txid()))
+                            .await;
+
+                        // We maintain an append-only log of valid P2WSH transaction
+                        // outputs created since the federation was established.
```
elsirion: Some more description what it is used for would be good.

From conversation:
1. For client recovery and generally detection of deposits
2. For verification of existence of a UTXO (spentness gets updated)

modules/fedimint-walletv2-server/src/lib.rs
joschisan: Fixed.

modules/fedimint-walletv2-tests/bin/tests.rs
joschisan: Fixed.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: We define the finality delay as the "Number of confirmations required for a transaction to be considered as confirmed by the federation. The block that mines the transaction does not count towards the number of confirmations"

modules/fedimint-walletv2-server/src/lib.rs
joschisan: Fixed.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: I agree in principle and have considered this, however in this specific case everywhere else the fallback to 0 make and is just easier to handle. This place is less idiomatic but overall the block count is easier to reason about.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: As far as I understand we can only get the hash by height but not the block directly.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: Fixed

devimint/src/cli.rs
```diff
@@ -334,11 +334,19 @@ pub async fn handle_command(cmd: Cmd, common_args: CommonArgs) -> Result<()> {
                         )?;
 
                         dev_fed.bitcoind().await?.mine_blocks_no_wait(11).await?;
-                        dev_fed
-                            .internal_client()
-                            .await?
-                            .await_deposit(&operation_id)
-                            .await?;
+                        if crate::util::supports_wallet_v2() {
```
dpc, line 337: Nit: `crate::util::supports_wallet_v2`does not give any clue _what_ is supposed to support walletv2.

devimint/src/federation.rs
```diff
@@ -671,9 +714,24 @@ impl Federation {
...
+        };
+
         let operation_id = self.pegin_client_no_wait(amount, client).await?;
 
-        client.await_deposit(&operation_id).await?;
+        if let Some(initial) = initial_balance {
+            // Walletv2: wait for balance to increase. We expect slightly less than
+            // `amount` due to mint module fees when creating ecash notes.
+            let expected_balance = initial + (amount * 1000 * 9 / 10);
+            client.await_balance(expected_balance).await?;
```
dpc, line 731: Awaiting aggregate balance feels very ... flaky, as an API. Not sure if that's a problem, but just thinking aloud.

devimint/src/federation.rs
```diff
@@ -898,6 +991,11 @@ impl Federation {
     }
 
     pub async fn await_all_peers(&self) -> Result<()> {
```
dpc, line 993: Not a problem with this PR, but `await_all_peers` is not very informative name. Await all peers _what_?

devimint/src/util.rs
```diff
@@ -1037,6 +1040,10 @@ pub fn supports_lnv2() -> bool {
         || is_env_var_set(FM_ENABLE_MODULE_LNV2_ENV)
 }
 
+pub fn supports_wallet_v2() -> bool {
```
dpc, line 1048: `is_wallet_v2_support_enabled` maybe?

fedimint-cli/src/client.rs
```diff
@@ -628,7 +628,18 @@ async fn get_note_summary(client: &ClientHandleArc) -> anyhow::Result<serde_json
     let mint_module_id = client
         .get_first_instance(&fedimint_mint_client::KIND)
         .context("Mint module not found")?;
-    let wallet_client = client.get_first_module::<WalletClientModule>()?;
+
+    // Try wallet v1 first, then walletv2
```
dpc, line 632: Would make this pattern a function. `get_available_wallet_module` or something?

dpc: Oh... but different type ... . Nvmd, I guess.

fedimint-walletv2-recoverytool/assets/bootstrap.bundle.min.js
```diff
@@ -0,0 +1,7 @@
+/*!
```
dpc: Large bundles of minimized hard to review js in a tool that can potentially steal money seem a bit meh.

We should make sure to use `integrity` tag on the `<script` and put a comment with a link to some official docs to make it easy to verify this is legit upstream versions.

fedimint-walletv2-recoverytool/src/components/key_input.rs
```diff
@@ -0,0 +1,231 @@
...
+                    keys: KeysConfig { descriptor },
+                });
+            }
+            Err(e) => {
+                set_error.set(Some(format!("Invalid keys: {e}")));
+            }
+        }
+    };
+
+    view! {
```
dpc: Note: these macros are not too bad. I still like maud more, but not too bad. :D

fedimint-rugtool/index.html
```diff
@@ -0,0 +1,17 @@
...
+    <title>Fedimint Wallet Recovery</title>
+    <link data-trunk rel="copy-file" href="assets/bootstrap.min.css"/>
+    <link rel="stylesheet" href="bootstrap.min.css"/>
+    <link data-trunk rel="css" href="style.css"/>
+</head>
+<body>
+    <div id="app"></div>
+    <link data-trunk rel="rust" data-wasm-opt="z"/>
+    <link data-trunk rel="copy-file" href="assets/bootstrap.bundle.min.js"/>
+    <script src="bootstrap.bundle.min.js"></script>
```
dpc: `integrity` + link as per comment above?

fedimint-server-ui/src/dashboard/mod.rs
dpc, line 235: This will need to be conditional now too, no?

devimint/src/cli.rs
joschisan: This follows existing naming convention with supports_lnv2

devimint/src/util.rs
joschisan: This follows existing naming convention with supports_lnv2

fedimint-server-ui/src/dashboard/mod.rs
joschisan: It already is no? The dashboard should work fine even if current core module are not present.

fedimint-walletv2-recoverytool/src/state.rs
```diff
@@ -0,0 +1,127 @@
+use bdk_wallet::Wallet;
+use bdk_wallet::bitcoin::{Network, Txid};
+use leptos::prelude::StoredValue;
```
elsirion: Should we make this a separate repo just to avoid pulling in the entire Leptos dep tree?

fedimint-rugtool/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+mod app;
```
elsirion: Please rename the folder 😆

modules/fedimint-walletv2-server/src/lib.rs
elsirion: Oh, sorry for the confusion

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1398 @@
...
+                            vout: u32::try_from(vout)
+                                .expect("Bitcoin transaction has more than u32::MAX outputs"),
+                        };
+
+                        let index = dbtx
+                            .find_by_prefix_sorted_descending(&DepositPrefix)
+                            .await
+                            .next()
+                            .await
+                            .map_or(0, |entry| entry.0.0 + 1);
```
elsirion, line 993: This pattern showed some pretty bad performance in the past, probably ok though since it's not run in a tight loop.

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+        secp256k1::SECP256K1,
+        &Scalar::from_be_bytes(tweak.to_byte_array()).expect("Hash is within field order"),
+    )
+    .expect("Failed to tweak bitcoin public key")
+}
+
+/// Returns true if the script pubkey is a valid deposit address for the
+/// federation. This uses a probabilistic filter - only ~1/65536 of P2WSH
+/// scripts pass.
+pub fn is_valid_script(script_pubkey: &ScriptBuf, pks_hash: &sha256::Hash) -> bool {
```
elsirion: Nit: Naming, it's less about being valid and more about potentially belonging to this federation.

fedimint-walletv2-recoverytool/assets/bootstrap.bundle.min.js
joschisan: Added the integrity tags.

fedimint-rugtool/index.html
joschisan: Added the integrity tags.

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1398 @@
...
+
+        let consensus_receive_fee = self
+            .receive_fee(dbtx)
+            .await
+            .ok_or(WalletInputError::NoConsensusFeerateAvailable)?;
+
+        // We allow for a higher fee such that a guardian could construct a CPFP
+        // transaction. This is the last line of defense should the federations
+        // transactions ever get stuck due to a critical failure of the feerate
+        // estimation.
```
elsirion, line 502: Do we currently limit the number of unconfirmed children? Would seem necessary if this is to work since there's some max package size even with package relay afaik. Idk if that's worth mitigating though (would severely limit throughput since we can only detect confirmations with finality delay).

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1398 @@
...
+
+        if let Some(wallet) = dbtx.remove_entry(&FederationWalletKey).await {
+            // The first pegin into the federation is confirmed hence its output value is
+            // over the P2WSH dust limit. By induction so is this change value.
+            let change_value = wallet
+                .value
+                .checked_add(deposit_value)
+                .ok_or(WalletInputError::ArithmeticOverflow)?;
+
+            assert!(change_value >= self.cfg.consensus.dust_limit);
```
elsirion: This is actually not technically correct since the dust limit is policy and a malicious miner (or one offering non-standard tx inclusion for money) could include 2 deposits of 1 sat in blocks, thriggering the assert.

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1398 @@
...
+                    TxIn {
+                        previous_output: tracked_outpoint,
+                        script_sig: Default::default(),
+                        sequence: Sequence::ENABLE_RBF_NO_LOCKTIME,
+                        witness: bitcoin::Witness::new(),
+                    },
+                ],
+                output: vec![TxOut {
+                    value: change_value,
+                    script_pubkey: self.descriptor(&wallet.consensus_hash()).script_pubkey(),
```
elsirion, line 540: It's probably fine, but hashing the entire wallet struct will make an alternative implementation of the "follow the UTXO chain" functionality somewhat more annoying imo (e.g. in a recovery tool that uses the static backup instead of data from the admin dashboard).

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -285,7 +288,6 @@ fn default_finality_delay(network: Network) -> u32 {
     match network {
         Network::Bitcoin | Network::Regtest => 10,
         Network::Testnet | Network::Signet | Network::Testnet4 => 2,
-        _ => panic!("Unsupported network"),
```
elsirion, line 288: Unrelated change. While good, should have been a separate PR.

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1398 @@
...
+        }
+    }
+
+    async fn tx_id(&self, dbtx: &mut DatabaseTransaction<'_>, outpoint: OutPoint) -> Option<Txid> {
+        let index = dbtx.get_value(&TxInfoIndexKey(outpoint)).await?;
+
+        dbtx.get_value(&TxInfoKey(index))
+            .await
+            .map(|entry| entry.txid)
+    }
```
elsirion, line 1273: What does this fn do? Docs …

It's a bit confusing because you are dealing with both Bitcoin on-chain as well as Fedimint TX IDs and OutPoints …

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1398 @@
...
+        dbtx: &mut DatabaseTransaction<'_>,
+        audit: &mut Audit,
+        module_instance_id: ModuleInstanceId,
+    ) {
+        audit
+            .add_items(
+                dbtx,
+                module_instance_id,
+                &FederationWalletPrefix,
+                |_, federation_utxo| 1000 * federation_utxo.value.to_sat() as i64,
```
elsirion: Should probably use try_into and fail if out of range. Otherwise might silently overflow in prod builds.

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1398 @@
...
+                    .collect::<Vec<FederationTx>>()
+                    .await;
+
+                for unconfirmed_tx in unconfirmed_txs {
+                    btc_rpc.submit_transaction(unconfirmed_tx.tx).await;
+                }
+
+                sleep(common::sleep_duration()).await;
+            }
+        });
```
elsirion, line 907: New signed tx should trigger this (i.e. select over sleep and a channel). Otherwise a withdrawal will on average wait 30s till it shows up in the mempool.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: Yes it is implicitly limited by the exponential feerate increase with the number pending transactions.

fedimint-rugtool/src/lib.rs
joschisan: Merged it into lib.rs

fedimint-walletv2-recoverytool/src/state.rs
joschisan: I got no opinion on this.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: This will give you the bitcoin TxId for the fedimint outpoint. It allows the state machine for a send to retrieve and store the bitcoin txid and log it in the send status update event.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: The msat amount of the federation utxo should be well within i64 range?

modules/fedimint-walletv2-server/src/lib.rs
joschisan: Its now broadcast immediately once signed.

modules/fedimint-walletv2-server/src/lib.rs
elsirion: Right, getting close but not too close. I mis-remembered, thought sats alsready got close and msat were technically over the line at 21M, but they aren't:

```
log2(21000000*10**8*1000)
Out[4]: 60.865095035863924
```

modules/fedimint-walletv2-server/src/lib.rs
elsirion: @joschisan the assert is gone, but I think the comment is still incorrect

fedimint-rugtool/src/lib.rs
elsirion: Sorry, I was actually talking about the `fedimint-rugtool` name, just didn't have a good way to put it :see_no_evil:

joschisan: Done.

modules/fedimint-walletv2-server/src/lib.rs
joschisan: Fixed.

fedimint-rugtool/src/lib.rs
joschisan: Its renamed.

modules/fedimint-walletv2-client/src/send_sm.rs
```diff
@@ -0,0 +1,135 @@
...
+                        SendPaymentStatusEvent {
+                            operation_id: old_state.common.operation_id,
+                            status: SendPaymentStatus::Aborted,
+                        },
+                    )
+                    .await;
+
+                old_state.update(SendSMState::Aborted(error))
+            }
+            AwaitFundingResult::Failure => old_state.update(SendSMState::Failure),
```
elsirion, line 132: Even unexpected failure states should log events, otherwise these are hard to debug when they happen (would appear stuck).

modules/fedimint-walletv2-client/src/send_sm.rs
```diff
@@ -0,0 +1,135 @@
...
+        outpoint: OutPoint,
+    ) -> AwaitFundingResult {
+        if let Err(error) = global_context.await_tx_accepted(outpoint.txid).await {
+            return AwaitFundingResult::Aborted(error);
+        }
+
+        match global_context.module_api().tx_id(outpoint).await {
+            Some(txid) => AwaitFundingResult::Success(txid),
+            None => AwaitFundingResult::Failure,
+        }
```
elsirion, line 94: Claude is moaning about a race condition here, but I can't see it. Once the tx is reported as accepted it's guaranteed that the tx id index has been updated in `process_output` imo.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+            .find_by_prefix_sorted_descending(&ValidAddressIndexPrefix)
+            .await
+            .next()
+            .await
+        {
+            self.derive_address(entry.0.0)
+        } else {
+            self.derive_address(self.next_valid_index(0))
+        }
+    }
```
elsirion, line 378: We discussed having a gap limit in the past, not necessary for MVP, but want to make sure that could be patched in easily later, right? As it is you have to wait for the previous deposit tx to be recognized by the federation before you can get a fresh address, that's an anti-pattern and really bad for privacy if you are receiving a lot of deposits.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+                    .receive_fee()
+                    .await?
+                    .ok_or(anyhow!("No consensus feerate is available"))?;
+
+                if tx_out.value > receive_fee && !deposit_range.spent.contains(&deposit_index) {
+                    let pending_txs = self.module_api.pending_tx_chain().await?;
+
+                    if pending_txs.len() >= 3 {
+                        return Ok(false);
+                    }
```
elsirion: We don't want to claim the deposit in this case because the fee would be higher than necessary I assume? Could use some comments what's happening (the fn in general).

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+                }
+            }
+
+            let mut dbtx = self.db.begin_transaction().await;
+
+            dbtx.insert_entry(&NextDepositIndexKey, &(deposit_index + 1))
+                .await;
+
+            dbtx.commit_tx_result().await?;
+        }
```
elsirion, line 595: Is a crash between these two commits fine? Using 3 different DB transactions here makes me a bit nervous. Could there be a double-claim that leads to confusing transaction list entries?

modules/fedimint-walletv2-client/src/send_sm.rs
joschisan: I am aware, but atm this is the consistent pattern across all modules for the send and receive events. This should never happen anyways.

modules/fedimint-walletv2-client/src/send_sm.rs
joschisan: My Claude had the same complaint but I also think ist wrong here.

modules/fedimint-walletv2-client/src/lib.rs
joschisan: A crash here should be fine - the index will be detected as spent just like in recovery.

modules/fedimint-walletv2-client/src/lib.rs
joschisan: > We don't want to claim the deposit in this case because the fee would be higher than necessary I assume?

Yes, we want to wait here as the congestion should clear up quickly.

modules/fedimint-walletv2-client/src/lib.rs
joschisan: Yes this is easy to patch in later, you just need to track more clientside. For the server nothing changes.

The client however would have to track which addressed have a ready been used, as we cannot infer that anymore.

modules/fedimint-walletv2-client/src/send_sm.rs
elsirion: I guess if something gets stuck we'll have to pull logs then …

modules/fedimint-walletv2-client/src/lib.rs
m1sterc001guy: If I'm understanding correctly, this is a problem for the CLI. The CLI does not run in the background to claim deposits and will just return the same address until it has seen a deposit. 

In my initial testing of the CLI, I sent some funds to a deposit address using bitcoin-cli and then I generated some blocks. I could not figure out how to get a new address, `receive` just always returned the same address.

CLI isn't just a testing tool, we need to be ready for the influx of agents that want to use fedimint as their wallet :)

modules/fedimint-walletv2-common/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        if program.is_p2tr() {
+            return Some(StandardScript::P2TR(
+                XOnlyPublicKey::from_slice(program.program().as_bytes())
+                    .expect("Witness program is 32 bytes"),
+            ));
+        }
+
+        None
+    }
```
m1sterc001guy, line 292: Why not make this return `Result`? Then we can remove the `expect` when parsing the program

modules/fedimint-walletv2-client/src/send_sm.rs
m1sterc001guy: What logs are there even to pull? From what I can tell if this is ever hit, it just silently fails and the user would need to contact the guardians and pull their logs to figure out if something went wrong.

Very much in favor of adding an event here, failure events should be the most important things to log. If it can truly never happen, then we should add an expect somewhere and remove the `Failure` state altogether.

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+    }
+
+    /// Fetch the current fee required to send an on-chain payment.
+    pub async fn send_fee(&self) -> Result<bitcoin::Amount, SendError> {
+        self.module_api
+            .send_fee()
+            .await
+            .map_err(|e| SendError::FederationError(e.to_string()))?
+            .ok_or(SendError::NoConsensusFeerateAvailable)
+    }
```
m1sterc001guy, line 249: I think this should be renamed `send_fee_absolute` so that it distinguishes that its not a feerate

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+
+        let fee = match fee {
+            Some(value) => value,
+            None => self
+                .module_api
+                .send_fee()
+                .await
+                .map_err(|e| SendError::FederationError(e.to_string()))?
+                .ok_or(SendError::NoConsensusFeerateAvailable)?,
+        };
```
m1sterc001guy, line 274: Why is `fee` even provided as an option? Shouldn't we always go and re-retrieve it?

modules/fedimint-walletv2-common/src/config.rs
```diff
@@ -0,0 +1,216 @@
...
+                    + destination_output_weight,
+            ),
+            receive_tx_vbytes: weight_to_vbytes(
+                tx_overhead_weight
+                    + change_input_weight
+                    + change_input_weight
+                    + change_output_weight,
+            ),
+            min_feerate: 1000,
+            dust_limit: bitcoin::Amount::from_sat(10_000),
```
m1sterc001guy, line 121: `10000` sats seems pretty high for a dust limit?

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+    ) -> (OperationId, TransactionId) {
+        let operation_id = OperationId::new_random();
+
+        let client_input = ClientInput::<WalletInput> {
+            input: WalletInput::V0(WalletInputV0 {
+                deposit_index,
+                fee,
+                tweak: self.derive_tweak(address_index).public_key(),
+            }),
+            keys: vec![self.derive_tweak(address_index)],
```
m1sterc001guy, line 420: Can save a derivation here by saving the output of 418 to a variable

modules/fedimint-walletv2-client/src/send_sm.rs
m1sterc001guy: > this is the consistent pattern across all modules for the send and receive events

I don't understand this comment. The receive state machine does not have a `Failure` event

modules/fedimint-walletv2-client/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+                .await
+                .next()
+                .await
+                .is_none()
+            {
+                dbtx.insert_new_entry(&ValidAddressIndexKey(module.next_valid_index(0)), &())
+                    .await;
+            }
+
+            dbtx.commit_tx().await;
```
m1sterc001guy, line 495: What is this doing? A comment would help.

modules/fedimint-walletv2-server/src/lib.rs
```diff
@@ -0,0 +1,1404 @@
...
+                dbtx.remove_entry(&UnconfirmedTxKey(tx.compute_txid()))
+                    .await;
+
+                // We maintain an append-only log of valid P2WSH transaction outputs created
+                // since the federation was established. This is downloaded by clients to
+                // detect pegins and claim them by index.
+
+                for (vout, tx_out) in tx.output.iter().enumerate() {
+                    if is_potential_receive(&tx_out.script_pubkey, &pks_hash)
+                        && tx_out.script_pubkey.is_p2wsh()
```
m1sterc001guy, line 980: Isn't this a privacy degradation compared to wallet v1? Now it is more obvious which addresses on chain are fedimint addresses because of the hash filter

modules/fedimint-lnv2-tests/bin/tests.rs
```diff
@@ -220,7 +220,7 @@ async fn test_payments(dev_fed: &DevJitFed) -> anyhow::Result<()> {
 
     federation.pegin_client(10_000, &client).await?;
 
-    almost_equal(client.balance().await?, 10_000 * 1000, 5_000).unwrap();
+    almost_equal(client.balance().await?, 10_000 * 1000, 500_000).unwrap();
```
m1sterc001guy, line 223: Woah, why was this change necessary?

devimint/src/cli.rs
m1sterc001guy: Should probably rename both to `devimint_supports_wallet_v2`, etc

modules/fedimint-walletv2-client/src/lib.rs
m1sterc001guy: Often times when doing a peg in, the client knows their TXID of the peg in. I think we should add a CLI command that polls your TX chain function and returns when their tx has been included.

devimint/Cargo.toml
```diff
@@ -38,6 +38,8 @@ fedimint-portalloc = { workspace = true }
 fedimint-server = { workspace = true }
 fedimint-testing-core = { workspace = true }
 fedimint-wallet-client = { workspace = true, features = ["cli"] }
+fedimint-walletv2-client = { workspace = true }
+fedimint-walletv2-server = { workspace = true }
```
m1sterc001guy, line 42: Hmm curious why the server is needed here

devimint/src/util.rs
m1sterc001guy: We can and should change naming conventions when there is > 1 use case :)

devimint/src/federation.rs
m1sterc001guy: +1

.gitignore
```diff
@@ -32,3 +32,7 @@
 fedimint-startos/docker-images
 fedimint-startos/fedimintd.s9pk
 fedimint-startos/scripts/embassy.js
+
+# Trunk build artifacts
```
m1sterc001guy, line 36: Trunk?

modules/fedimint-walletv2-common/src/config.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// | 11        | 379  | 618     |
+    /// | 12        | 406  | 672     |
+    /// | 13        | 414  | 689     |
+    /// | 14        | 441  | 742     |
+    /// | 15        | 468  | 796     |
+    /// | 16        | 476  | 813     |
+    /// | 17        | 503  | 867     |
+    /// | 18        | 530  | 920     |
+    /// | 19        | 539  | 937     |
+    /// | 20        | 565  | 991     |
```
m1sterc001guy, line 74: Can you include a formula here for how these numbers were computed?

PR #6438, opened 2024-11-21
fix(ci): flake under load due to too short wait

dpc:


fedimint-core/src/util/backoff_util.rs
```diff
@@ -25,6 +25,10 @@ pub fn aggressive_backoff() -> FibonacciBackoff {
     custom_backoff(Duration::from_millis(200), Duration::from_secs(5), Some(10))
```
dpc, line 25: This seems to give up after 22s, so I need to a fn that will try a bit longer.

elsirion (approved):

PR #6445, opened 2024-11-22
fix: use linear pagination algorithm for operation log

### LINKED ISSUE #6191: Make `list_operations` impl efficient
I left this comment a long time ago that we rediscovered it today:

https://github.com/fedimint/fedimint/blob/6249552fd55d599a4972d4c1e2bd8028e8c75b4b/fedimint-client/src/oplog.rs#L71-L78

The pagination of operation listing is horribly broken and requires a rewrite. Luckily, the second point is now possible after @dpc implemented `find_by_range` in https://github.com/fedimint/fedimint/pull/6005

elsirion:
Fixes #6191

### DIFF
--- a/fedimint-client/src/oplog.rs
+++ b/fedimint-client/src/oplog.rs
@@ -17,6 +17,7 @@ use fedimint_logging::LOG_CLIENT;
 use futures::{stream, Stream, StreamExt};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use tokio::sync::OnceCell;
 use tracing::{error, instrument, warn};
 
 use crate::db::{
@@ -26,11 +27,34 @@ use crate::db::{
 #[derive(Debug, Clone)]
 pub struct OperationLog {
     db: Database,
+    oldest_entry: OnceCell<ChronologicalOperationLogKey>,
 }
 
 impl OperationLog {
     pub fn new(db: Database) -> Self {
-        Self { db }
+        Self {
+            db,
+            oldest_entry: OnceCell::new(),
+        }
+    }
+
+    /// Will return the oldest operation log key in the database and cache the
+    /// result. If no entry exists yet the DB will be queried on each call till
+    /// an entry is present.
+    async fn get_oldest_operation_log_key(&self) -> Option<ChronologicalOperationLogKey> {
+        let mut dbtx = self.db.begin_transaction_nc().await;
+        self.oldest_entry
+            .get_or_try_init(move || async move {
+                dbtx.find_by_prefix(&ChronologicalOperationLogKeyPrefix)
+                    .await
+                    .map(|(key, ())| key)
+                    .next()
+                    .await
+                    .ok_or(())
+            })
+            .await
+            .ok()
+            .copied()
     }
 
     pub async fn add_operation_log_entry(
@@ -71,13 +95,7 @@ impl OperationLog {
 
         let mut dbtx = self.db.begin_transaction_nc().await;
         let operation_log_keys = if let Some(start_after) = last_seen {
-            let Some(last_key) = dbtx
-                .find_by_prefix(&ChronologicalOperationLogKeyPrefix)
-                .await
-                .map(|(key, ())| key)
-                .next()
-                .await
-            else {
+            let Some(oldest_entry_key) = self.get_oldest_operation_log_key().await else {
                 return vec![];
             };
 
@@ -90,7 +108,7 @@ impl OperationLog {
                 // We want to get all operation log keys in the range [last_key, start_after). Sp as
                 // long as the start time is greater than the last key's creation time, we have to
                 // keep going.
-                .take_while(|&start_time| start_time >= last_key.creation_time)
+                .take_while(|&start_time| start_time >= oldest_entry_key.creation_time)
                 .map(|start_time| {
                     let end_time = start_time - EPOCH_DURATION;
 
@@ -676,5 +694,23 @@ mod tests {
             ],
         )
         .await;
+    }
+
+    #[tokio::test]
+    async fn test_pagination_empty_then_not() {
+        let db = Database::new(MemDatabase::new(), ModuleRegistry::default());
+        let op_log = OperationLog::new(db.clone());
+
+        let page = op_log.list_operations(10, None).await;
+        assert!(page.is_empty());
+
+        let mut dbtx = db.begin_transaction().await;
+        op_log
+            .add_operation_log_entry(&mut dbtx.to_ref_nc(), OperationId([0; 32]), "foo", "bar")
+            .await;
+        dbtx.commit_tx().await;
+
+        let page = op_log.list_operations(10, None).await;
+        assert_eq!(page.len(), 1);
     }
 }
--- a/fedimint-client/src/oplog.rs
+++ b/fedimint-client/src/oplog.rs
@@ -93,86 +93,42 @@ impl OperationLog {
     ) -> Vec<(ChronologicalOperationLogKey, OperationLogEntry)> {
         const EPOCH_DURATION: Duration = Duration::from_secs(60 * 60 * 24 * 7);
 
+        let start_after_key = last_seen.unwrap_or_else(|| ChronologicalOperationLogKey {
+            creation_time: now(),
+            operation_id: OperationId([0; 32]),
+        });
+
+        let Some(oldest_entry_key) = self.get_oldest_operation_log_key().await else {
+            return vec![];
+        };
+
         let mut dbtx = self.db.begin_transaction_nc().await;
-        let operation_log_keys = if let Some(start_after) = last_seen {
-            let Some(oldest_entry_key) = self.get_oldest_operation_log_key().await else {
-                return vec![];
-            };
+        let mut operation_log_keys = Vec::with_capacity(limit);
 
-            // We want to fetch all operations that were created before `start_after`, going
-            // backwards in time. This means "start" generally means a later time than
-            // "end". Only when creating a rust Range we have to swap the terminology (see
-            // comment there).
-            let epochs_rev_ranges = (0..)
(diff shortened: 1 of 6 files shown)


fedimint-client/src/oplog.rs
```diff
@@ -64,45 +67,107 @@ impl OperationLog {
...
+            else {
+                return vec![];
+            };
+
+            // We want to fetch all operations that were created before `start_after`, going
+            // backwards in time. This means "start" generally means a later time than
+            // "end". Only when creating a rust Range we have to swap the terminology (see
+            // comment there).
+            let epochs_rev_ranges = (0..)
+                .map(|epoch| start_after.creation_time - epoch * EPOCH_DURATION)
```
elsirion, line 89: Are we ok with just not expecting people to run our app in 1970?

fedimint-client/src/oplog.rs
```diff
@@ -64,45 +67,107 @@ impl OperationLog {
         limit: usize,
         start_after: Option<ChronologicalOperationLogKey>,
```
dpc, line 68: "verbs" as variable names, confusing. Also, "after" suggests `<`, while for range start I would expect `start <= x`

dpc: `last_seen` maybe?

dpc: Would rename this function too while at it.

fedimint-client/src/oplog.rs
```diff
@@ -64,45 +67,107 @@ impl OperationLog {
...
-                    false
-                };
-
-                std::future::ready(skip)
-            })
-            .take(limit)
-            .collect::<Vec<_>>()
-            .await;
+        let operation_log_keys = if let Some(start_after) = start_after {
+            let Some(last_key) = dbtx
```
dpc, line 74: Would cache when `Some`. It can't ever change.

fedimint-client/src/oplog.rs
dpc: [They weren't ready for it anyway](https://github.com/user-attachments/assets/c79e271c-64f0-4125-8d05-485e9fe14e43)

fedimint-client/src/oplog.rs
dpc: Does it mean we just have to impl the same thing we implemented for event log (a single writer serializing the order)?

fedimint-client/src/oplog.rs
```diff
@@ -64,45 +67,107 @@ impl OperationLog {
...
+            // We want to fetch all operations that were created before `start_after`, going
+            // backwards in time. This means "start" generally means a later time than
+            // "end". Only when creating a rust Range we have to swap the terminology (see
+            // comment there).
+            let epochs_rev_ranges = (0..)
+                .map(|epoch| start_after.creation_time - epoch * EPOCH_DURATION)
+                // We want to get all operation log keys in the range [last_key, start_after). Sp as
+                // long as the start time is greater than the last key's creation time, we have to
+                // keep going.
+                .take_while(|&start_time| start_time >= last_key.creation_time)
```
dpc, line 93: I thought that it will be simpler.

IIUC, we want to paginate from the newest to the oldest?

In that case I would just always query `find_by_range(last_seen-EPOCH..last_seen)`, and use `start_key` (minimum know in the collection), to just terminate pagination when `last_seen < start_key`.

If querying `find_by_range(last_seen-EPOCH..last_seen)` returns too many entries, just throw away extra ones. If not enough, loop and query for more, the concat.

fedimint-client/src/oplog.rs
elsirion: > Would cache when Some. It can't ever change.

Is that really worth it? You aren't supposed to call this fn in a hot loop. Can add it in a separate commit and then we decide.

> Theoretically a new entry can get added with previous timestamp at any point, right?.

Sure, but that would also trigger some sort of event in the future I assume? We currently don't deal with the case "new operation was inserted, send diff/insert to frontend". Currently, you just have to fetch the log again.

In practice, DB transactions, that add operation log entries, aren't that long-running. So the time it takes a user to navigate from "send transaction" to "show transaction log" would typically be sufficient to commit the dbtx.

You might be thinking about it as an API to get updates, but for that it's entirely ill-suited because it goes backwards in time and not forwards. The event log is the right abstraction for that use case. `list_operations` is about historic data that gets lazy-loaded as you scroll through your transaction log.

Thinking about it, we might want to have a "forward" function too at some point, so a UI can drop more current data if a user scrolls too far back and re-fetch it when they scroll up again.

fedimint-client/src/oplog.rs
elsirion: > IIUC, we want to paginate from the newest to the oldest (reverse order)?

:+1: 

> In that case I would just always query `find_by_range(last_seen-EPOCH..last_seen)`, and use `start_key` (minimum know in the collection), to just terminate pagination when `last_seen < start_key`.

That's what happens here, isn't it? I just don't want the caller of `list_operations` to worry about that end timestamp, but rather abort pagination when `None` is returned.

fedimint-client/src/oplog.rs
elsirion: > Would rename this function too while at it.

Please make a recommendation in that case :laughing:

fedimint-client/src/oplog.rs
dpc: All it means is that this method can't be used for reliably noticing new operations, as it can miss new entries unless queried from beginning, and probably we should explicitly state it in the documentation. But this is probably not an issue.

fedimint-client/src/oplog.rs
dpc: We're using slower `find_by_prefix_sorted_descending` here and generally I'm confused about this function.

Pseudo-codishly:

```rust
let mut entries = vec![];
let mut last_seen = last_seen;

while log_entries.len() < limit {
  if last_seen < first_ever_event {
    // reached the beginning, no point in querying for more
    break;
  }
  let needs_num_entries = limit - entries.len();
  // get PERIOD worth of entries
  let more_entries = db.get_prefix(last_seen-PERIOD..last_seen);
  last_seen = more_entries.last();
  // take last `needs_num_entries`
  let more_entries = log_entries[more_entries.len().satuarating_sub(needs_num_entries)..more_entries.len()];
  log_entries = concat([more_events, entries]);
}

entries
```

fedimint-client/src/oplog.rs
dpc: `get_operations_before_op_log_key`? `paginate_operations_rev`? `paginate_oplog_desc`?

fedimint-client/src/oplog.rs
dpc: > Is that really worth it?

:shrug: Low gain, low effort.

fedimint-client/src/oplog.rs
dpc: For the records: I was explained that I'm just confused and the existing code works kind of like it already.

fedimint-client/src/oplog.rs
```diff
@@ -62,47 +74,103 @@ impl OperationLog {
...
+                return vec![];
+            };
+
+            // We want to fetch all operations that were created before `start_after`, going
+            // backwards in time. This means "start" generally means a later time than
+            // "end". Only when creating a rust Range we have to swap the terminology (see
+            // comment there).
+            let epochs_rev_ranges = (0..)
+                .map(|epoch| start_after.creation_time - epoch * EPOCH_DURATION)
+                // We want to get all operation log keys in the range [last_key, start_after). Sp as
```
m1sterc001guy, line 93: nit: typo on `Sp`

fedimint-client/src/oplog.rs
elsirion: As discussed on our call I split up the function to make it more readable.

fedimint-client/src/oplog.rs
```diff
@@ -23,11 +27,34 @@ use crate::db::{
...
+        Self {
+            db,
+            oldest_entry: OnceCell::new(),
+        }
+    }
+
+    /// Will return the oldest operation log key in the database and cache the
+    /// result. If no entry exists yet the DB will be queried on each call till
+    /// an entry is present.
+    async fn get_oldest_operation_log_key(&self) -> Option<ChronologicalOperationLogKey> {
```
dpc, line 44: :heart:

fedimint-client/src/oplog.rs
```diff
@@ -57,52 +84,73 @@ impl OperationLog {
...
+    pub async fn paginate_operations_rev(
         &self,
         limit: usize,
-        start_after: Option<ChronologicalOperationLogKey>,
+        last_seen: Option<ChronologicalOperationLogKey>,
     ) -> Vec<(ChronologicalOperationLogKey, OperationLogEntry)> {
+        const EPOCH_DURATION: Duration = Duration::from_secs(60 * 60 * 24 * 7);
+
+        let start_after_key = last_seen.unwrap_or_else(|| ChronologicalOperationLogKey {
+            creation_time: now(),
```
dpc, line 106: Double check: timestamps created at `now()` are included. Just to not skip them in a racy, hard to debug way.

dpc: Looks like we might because `Range` is not inclusive at the end? We could add a second, or maybe use `OperationId([0xff;32])`? Or maybe I'm confused.

My point is we should try to not theoretically skip anything that is already there if `last_seen == None`.

dpc (commented):
Even I understand it now. :)

One detail about the placeholder for None.

fedimint-client/src/oplog.rs
elsirion: You are right, since we deal with full precision time it's unlikely, but better safe than sorry, I'll add a commit to add some time.

dpc (dismissed):

elsirion:
CI was unhappy about a docstring :/

dpc (approved):

PR #6446, opened 2024-11-22
chore(ci): use newly added `fs-dir-cache exec`

dpc:
It has automatic lock management, which makes unlocking 100% reliable and faster.

### DIFF
--- a/.github/workflows/ci-backwards-compatibility.yml
+++ b/.github/workflows/ci-backwards-compatibility.yml
@@ -60,7 +60,7 @@ jobs:
             CARGO_PROFILE=ci \
             PERFIT_ACCESS_TOKEN="${{ secrets.PERFIT_ACCESS_TOKEN }}" \
             nix develop -c \
-            nix shell github:rustshop/fs-dir-cache -c \
+            nix shell 'github:rustshop/fs-dir-cache?rev=e9752d00ee16778c9d3d0b93a09c49c44013ac17' -c \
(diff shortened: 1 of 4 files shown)


dpc:
@bradleystachurski for awarness

bradleystachurski (commented):
Looks like back-compat is failing

```
error: unrecognized subcommand 'exec'
Usage: fs-dir-cache <COMMAND>
```

dpc:
Needed to pin it to the new revision, as it's already cached on the workers.

maan2003:
👏 for idea to use sockets as guard

you can probably also do it for non exec version, but 

```bash
# lock
fs-dir-cache lock ... &
LOCK_PID=$!


# unlock
kill $LOCK_PID
```

elsirion:
So either the old process cleans up the socket itself when exiting gracefully or when connecting to it the new process notices that the other end isn't listening anymore, so the new process can clean it up? That's a cool hack! Even avoids polling by just waiting for the socket to close!

dpc:
> 👏 for idea to use sockets as guard


Yeah, I felt quite clever about it. No polling, guaranteed unlocking.

> you can probably also do it for non exec version, but exec one is good enough already

Technically there's nothing guaranteeing that the script couldn't get killed by `kill -9` before it was able to run `kill $LOCK_PID` etc. Using sockets seems 100% guaranteed, to there is still possibility that it will simply hang and hold on to the lock until killed.

I don't think we ever had cases like that, but will see how well it works in practice.

bradleystachurski (approved):

PR #6430, opened 2024-11-21
fix: deactivate cross compile dev shell for aarch64-linux

elsirion:
The dev shell for this architecture seems broken anyway, which prevents flake evaluation. No harm in deactivating the shell till someone actually wants to develop fedimint with cross-compilation on ARM Linux.

Fixes #6428

PR #6452, opened 2024-11-23
chore: check nix formatting on pre-commit

elsirion:
I accidentally PRed an incorrectly formatted flake in #6430, which the pre-commit hook should have prevented. CI should also check it now since it just runs the pre-commit hooks.

### DIFF
--- a/misc/git-hooks/pre-commit
+++ b/misc/git-hooks/pre-commit
@@ -178,6 +178,13 @@ function check_typos() {
 }
 export -f check_typos
 
+function check_nix_fmt() {
+    # we actually rely on word splitting here
+    # shellcheck disable=SC2046
+    nixfmt -c $(git ls-files | grep "\.nix$")
+}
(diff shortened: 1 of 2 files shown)


misc/git-hooks/pre-commit
```diff
@@ -178,6 +178,13 @@ function check_typos() {
 }
 export -f check_typos
 
+function check_nix_fmt() {
```
dpc, line 181: I think this file is auto-generated by flakebox, so the PR would have to be there (happy to do that for you if you want).

Also `git ls` output is available for re-use to avoid running it multiple times: `$FLAKEBOX_GIT_LS` or `$FLAKEBOX_GIT_LS_TEXT` (only text files).

dpc: @elsirion more work was needed, done in https://github.com/rustshop/flakebox/pull/177

PR #6457, opened 2024-11-25
feature: repurpose bitcoin status endpoint

tvolk131:
Change the return type of `check_bitcoin_status` from an enum containing info on whether the bitcoin backend is synced to a struct containing the latest block height, block time, and optional sync percentage estimate. This allows for consistent behavior for different bitcoin backends.

We're able to change the return type of this endpoint since it is only used by the guardian UI, and we require that the UI and guardian are upgraded in lockstep. In the UI, we will need to compare the block height and block time to the consensus block height and local system time in order to determine whether the guardian's bitcoin backend is stalled.

### DIFF
--- a/fedimint-server/src/config/api.rs
+++ b/fedimint-server/src/config/api.rs
@@ -5,7 +5,6 @@ use std::time::Duration;
 
 use async_trait::async_trait;
 use bitcoin::hashes::sha256;
-use bitcoincore_rpc::RpcApi;
 use fedimint_api_client::api::{DynGlobalApi, StatusResponse};
 use fedimint_bitcoind::create_bitcoind;
 use fedimint_core::admin_client::{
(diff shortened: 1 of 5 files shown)


Kodylow:
I like this since it works nicely with esplora too, could we get the percentage if it's available from bitcoind? that's still useful to have instead of removing completely

fedimint-server/src/config/api.rs
```diff
@@ -511,66 +509,52 @@ impl ConfigGenApi {
...
-    Ok(BitcoinRpcConnectionStatus::Synced)
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, Copy)]
-pub enum BitcoinRpcConnectionStatus {
-    Synced,
-    Syncing(f64),
+pub struct BitcoinRpcConnectionStatus {
+    chain_tip_block_height: u64,
+    chain_tip_block_time: u32,
```
elsirion, line 557: Useful to make it a `SystemTime`?

tvolk131: I don't see the purpose of that if we're just going to serialize it afterwards. Can't the API caller convert it in whatever language it uses?

tvolk131:
> could we get the percentage if it's available from bitcoind?

Done. I added it as a method to the `IBitcoindRpc` trait so we don't need to build a bitcoind client separate from the generic client we already build.

fedimint-server/src/config/api.rs
elsirion: I guess it's enough if only passed through. Generally typing things is better though.

elsirion (approved):

PR #6468, opened 2024-11-26
 chore(wallet-server): improve logging 

dpc:

### DIFF
--- a/modules/fedimint-wallet-server/src/lib.rs
+++ b/modules/fedimint-wallet-server/src/lib.rs
@@ -426,23 +426,30 @@ impl ServerModule for Wallet {
         // node is delayed processing of change outputs for the federation, which is an
         // acceptable risk since subsequent rounds of consensus will reattempt to fetch
         // the latest block count.
-        if let Ok(block_count) = self.get_block_count() {
-            let block_count_vote = block_count.saturating_sub(self.cfg.consensus.finality_delay);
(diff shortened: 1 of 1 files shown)


modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1575,7 +1593,10 @@ impl Wallet {
                             let _ = fee_rate_tx.send(feerate_with_multiplier);
                         }
                         Ok(None) => {
-                            debug!(target: LOG_MODULE_WALLET, "Bitcoin node did not return a fee rate");
+                            // Regtest node never returns fee rate, so no point spamming about it
+                            if !is_running_in_test_env() {
+                                warn!(target: LOG_MODULE_WALLET, "Bitcoin node did not return a fee rate");
+                            }
```
elsirion, line 1599: Wouldn't that also be quite spammy on a newly started federation with a fresh bitcoind? Doesn't it take a moment to start returning a fee rate?

elsirion (approved):
More logs more good!

modules/fedimint-wallet-server/src/lib.rs
dpc: `bitcoind` should have been running for some time syncing up, no?

Also it's not that spammy. Just annoying when debugging having to scroll through tens of screens worth of it.

PR #6469, opened 2024-11-27
 chore(core): better logging 

dpc:


fedimint-server/src/consensus/mod.rs
```diff
@@ -233,7 +232,7 @@ fn submit_module_ci_proposals(
     });
 
     task_group.spawn(
-        "submit_module_ci_proposals_{module_id}",
+        format!("citem_proposals_{module_id}"),
```
dpc, line 235: Too long, especially since it goes in a span.

elsirion (approved):

PR #6471, opened 2024-11-27
refactor: move collection en/decodable impls to separate file

tvolk131:
Also beef up our testing of these impls

### DIFF
--- a/fedimint-core/src/encoding/collections.rs
+++ b/fedimint-core/src/encoding/collections.rs
@@ -1,0 +1,342 @@
+use std::any::TypeId;
+use std::collections::{BTreeMap, BTreeSet, VecDeque};
+use std::fmt::Debug;
+
+use crate::module::registry::ModuleRegistry;
+use crate::{Decodable, DecodeError, Encodable, ModuleDecoderRegistry};
+
+impl<T> Encodable for &[T]
+where
+    T: Encodable + 'static,
+{
+    fn consensus_encode<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<usize> {
+        if TypeId::of::<T>() == TypeId::of::<u8>() {
+            // unsafe: we've just checked that T is `u8` so the transmute here is a no-op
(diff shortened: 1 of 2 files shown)


fedimint-core/src/encoding/collections.rs
```diff
@@ -0,0 +1,342 @@
...
+        assert_eq!(
+            VecDeque::<u8>::consensus_decode(&mut buf.as_slice(), &ModuleRegistry::default())
+                .unwrap(),
+            vec![1u8, 2]
+        );
+        assert_eq!(
+            VecDeque::<u16>::consensus_decode(&mut buf.as_slice(), &ModuleRegistry::default())
+                .unwrap(),
+            vec![1u16, 2]
+        );
```
tvolk131, line 324: Note for reviewer: I'm not sure whether this is the behavior we want or not, but these test assertions are revealing behavior that was already there.

fedimint-core/src/encoding/mod.rs
```diff
@@ -345,209 +343,6 @@ impl_encode_decode_tuple!(T1, T2);
...
-#[test]
-fn vec_decode_sanity() {
-    let buf = [
-        0xffu8, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    ];
-
-    // On malicious large len, return an error instead of panicking.
-    assert!(Vec::<u8>::consensus_decode(&mut buf.as_slice(), &ModuleRegistry::default()).is_err());
-    assert!(Vec::<u16>::consensus_decode(&mut buf.as_slice(), &ModuleRegistry::default()).is_err());
-}
```
tvolk131, line 464: Note for reviewer: This test wasn't actually testing malicious lengths, it was testing a length prefix that is greater than the number of decodable elements. This test was replaced by `test_lists`.

elsirion (approved):

PR #6475, opened 2024-11-27
refactor: gateway owns its task group

tvolk131:
There are several methods on the `Gateway` struct that currently require a `TaskGroup` argument. In practice the same task group is always used when calling all of these methods, so let's have the `Gateway` hold it as a struct field to make the purpose/ownership more clear.


elsirion (approved):
Nice!

PR #6482, opened 2024-11-30
feat: gateway auth redesign

### LINKED ISSUE #6427: Make gateway password required on first startup
When starting the gateway, we call `Gateway::get_gateway_configuration()` which loads the `GatewayConfiguration` from the DB if it exists, and attempts to create one and save it to the DB if it doesn't exist yet:

https://github.com/fedimint/fedimint/blob/c660335d1e48f1d0909d66b4aaf0ac68731638a5/gateway/ln-gateway/src/lib.rs#L1737-L1772

It can return `None`, but only if a gateway config isn't present in the DB _and_ a password is not provided in the `GatewayParameters`:

https://github.com/fedimint/fedimint/blob/c660335d1e48f1d0909d66b4aaf0ac68731638a5/gateway/ln-gateway/src/lib.rs#L1749

If we instead panicked on this condition, requiring the gateway operator to provide a password when starting up a brand new gateway for the first time, we would be able to significantly simplify the gateway, including removing the [`GatewayState::Configuring`](https://github.com/fedimint/fedimint/blob/c660335d1e48f1d0909d66b4aaf0ac68731638a5/gateway/ln-gateway/src/lib.rs#L173) state and changing the type of [`Gateway::gateway_config`](https://github.com/fedimint/fedimint/blob/c660335d1e48f1d0909d66b4aaf0ac68731638a5/gateway/ln-gateway/src/lib.rs#L212) from `Arc<RwLock<Option<GatewayConfiguration>>>` to `Arc<RwLock<GatewayConfiguration>>` which allows for lots of downstream simplifications.

tvolk131:
Fixes #6427

Previously, the gateway stored its password in its DB and required that a password be provided via an environment variable or CLI argument on first-time startup, after which any provided password was ignored. In this PR, we remove the password from the gateway DB and require that a bcrypt password hash be provided on _every_ startup (either through an environment variable or a CLI argument).

This allows several downstream changes, including:
* Removing `GatewayState::Configuring` since all it was waiting on was for a password to be set, whereas now an unset bcrypt password hash will cause the gateway to panic and exit.
* Merging `GatewayState::Initializing` into `GatewayState::Disconnected` since without the `Configuring` state, the two behave identically.
* Remove the distinction between always-authenticated API routes and authenticated-after-config API routes

### DIFF
--- a/gateway/ln-gateway/Cargo.toml
+++ b/gateway/ln-gateway/Cargo.toml
@@ -30,6 +30,7 @@ aquamarine = { workspace = true }
 async-stream = { workspace = true }
 async-trait = { workspace = true }
 axum = { version = "0.7.9", features = [ "json" ] }
+bcrypt = { workspace = true }
 bitcoin = { workspace = true }
 clap = { workspace = true }
 erased-serde = { workspace = true }
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeMap;
 
-use bitcoin::hashes::sha256;
+use bitcoin::hashes::{sha256, Hash};
 use fedimint_api_client::api::net::Connector;
 use fedimint_core::config::FederationId;
 use fedimint_core::db::{
@@ -21,8 +21,6 @@ use secp256k1::{Keypair, Secp256k1};
 use serde::{Deserialize, Serialize};
 use strum::IntoEnumIterator;
 use strum_macros::EnumIter;
-
-use crate::rpc::rpc_server::hash_password;
 
 pub trait GatewayDbtxNcExt {
     async fn save_federation_config(&mut self, config: &FederationConfig);
@@ -330,10 +328,10 @@ struct GatewayConfigurationV0 {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable)]
-pub struct GatewayConfigurationKey;
+pub struct GatewayConfigurationKeyV1;
 
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable, Serialize, Deserialize)]
-pub struct GatewayConfiguration {
+pub struct GatewayConfigurationV1 {
     pub hashed_password: sha256::Hash,
     pub num_route_hints: u32,
     #[serde(with = "serde_routing_fees")]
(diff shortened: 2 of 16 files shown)


m1sterc001guy:
Can we update the gateway's documentation with an example of how to use `bcrypt` to get a strong passwork hash?

gateway/ln-gateway/Cargo.toml
```diff
@@ -30,6 +30,7 @@ aquamarine = { workspace = true }
 async-stream = { workspace = true }
 async-trait = { workspace = true }
 axum = { version = "0.7.9", features = [ "json" ] }
+bcrypt = "0.16.0"
```
m1sterc001guy, line 33: Make workspace dependency?

gateway/ln-gateway/src/db.rs
```diff
@@ -330,10 +328,10 @@ struct GatewayConfigurationV0 {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable)]
-pub struct GatewayConfigurationKey;
+pub struct GatewayConfigurationKeyV1;
```
m1sterc001guy, line 331: Use V0, not V1. We're programmers :)

gateway/ln-gateway/src/lib.rs
```diff
@@ -236,6 +229,10 @@ pub struct Gateway {
 
     /// The task group for all tasks related to the gateway.
     task_group: TaskGroup,
+
+    /// The bcrypt password hash used to authenticate the gateway.
+    /// This is an `Arc` because `bcrypt::HashParts` does not implement `Clone`.
+    bcrypt_password_hash: Arc<bcrypt::HashParts>,
```
m1sterc001guy, line 235: Is this necessary here? Can you just pass `bcrypt_password_hash` as a parameter to the rpc server?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -92,50 +90,8 @@ async fn auth_middleware(
...
-async fn authenticate(
-    gateway_hashed_password: sha256::Hash,
-    password_salt: [u8; 16],
-    request: Request,
-    next: Next,
-) -> Result<axum::response::Response, StatusCode> {
     let token = extract_bearer_token(&request)?;
-    let hashed_password = hash_password(&token, password_salt);
-    if gateway_hashed_password == hashed_password {
+    if bcrypt::verify(token, &gateway.bcrypt_password_hash.to_string()).unwrap() {
```
m1sterc001guy, line 94: Is `unwrap` intentional here? If it is we should at least use `expect`

m1sterc001guy:
Concept ACK, left some nits

gateway/ln-gateway/src/lib.rs
tvolk131: This would require moving where we parse the gateway options:

https://github.com/fedimint/fedimint/blob/7b7bf0f45574911e8348c68f2abc5c61b36b117c/gateway/ln-gateway/src/lib.rs#L296-L297

We'd need to move it up a level:

https://github.com/fedimint/fedimint/blob/7b7bf0f45574911e8348c68f2abc5c61b36b117c/gateway/ln-gateway/src/bin/gatewayd.rs#L26-L38

And pass it into `.run()`

I can do this as a follow-up, but this seems outside the scope of this PR.

gateway/ln-gateway/Cargo.toml
tvolk131: Done.

gateway/ln-gateway/src/rpc/rpc_server.rs
tvolk131: Good catch! Done.

m1sterc001guy (approved):

PR #6486, opened 2024-12-02
chore: conditionally coordinate fedimintd restart in upgrade tests

bradleystachurski:
Previously, devimint would coordinate a shutdown for `fedimintd` upgrade tests if the version we're upgrading to is `v0.4.*` or `v0.5.*`. This change considers if we're upgrading from `v0.3.3` or `v0.3.4-rc.1`, since those are the only two versions that require coordinating a shutdown.

I've been running upgrade tests locally with these changes and everything is passing for `v0.5.0` release candidates.

### DIFF
--- a/devimint/src/federation.rs
+++ b/devimint/src/federation.rs
@@ -664,13 +664,18 @@ impl Federation {
         process_mgr: &ProcessManager,
         bin_path: &PathBuf,
     ) -> Result<()> {
+        let current_fedimintd_version = crate::util::FedimintdCmd::version_or_default().await;
+
         // get the version we're upgrading to, temporarily updating the fedimintd path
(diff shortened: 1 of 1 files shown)


devimint/src/federation.rs
```diff
@@ -679,8 +684,19 @@ impl Federation {
...
+    /// v0.3.3 and v0.3.4-rc.1 to greater minor version (e.g. v0.4.0)
+    fn version_requires_coordinated_shutdown(
+        from_version: &semver::Version,
+        to_version: &semver::Version,
+    ) -> bool {
+        from_version.major == 0
+            && from_version.minor == 3
+            && from_version.patch >= 3
+            && to_version.major == 0
+            && to_version.minor >= 4
```
elsirion, line 699: Probably some easier to read way of expression exists

PR #6489, opened 2024-12-02
chore(ci): record hostname before build

dpc:
I've observered some segfaults in the CI, and it remainded me that it is often beneficial to know which one of the runner machines is a given build running on.


dpc:
Huh. For some reason the job doesn't have access to `hostname` command...

dpc:
Pointless. The runner job is in the `Setup Job` step anyway.

PR #6500, opened 2024-12-03
feat: LNv1 Payment Events

m1sterc001guy:
Similar to https://github.com/fedimint/fedimint/pull/6351, this PR adds payment events for LNv1. It allows us to compute profit, fees, and latency for LNv1.

The LNv2 events are a bit better structured, but this still gives us the necessary data. Since we will be supporting LNv1 for a bit I figured it would be good to add this as well (might even be fun to compare with LNv2).

### DIFF
--- a/gateway/ln-gateway/src/state_machine/events.rs
+++ b/gateway/ln-gateway/src/state_machine/events.rs
@@ -1,0 +1,46 @@
+use fedimint_core::core::ModuleKind;
+use fedimint_core::Amount;
+use fedimint_eventlog::{Event, EventKind};
+use fedimint_ln_common::contracts::outgoing::OutgoingContractAccount;
+use fedimint_ln_common::contracts::ContractId;
+use serde::{Deserialize, Serialize};
+
+use super::pay::OutgoingPaymentError;
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentStarted {
+    pub contract_id: ContractId,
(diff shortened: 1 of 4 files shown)


gateway/ln-gateway/src/state_machine/events.rs
```diff
@@ -0,0 +1,46 @@
...
+use fedimint_ln_common::contracts::ContractId;
+use serde::{Deserialize, Serialize};
+
+use super::pay::OutgoingPaymentError;
+
+#[derive(Serialize, Deserialize)]
+pub struct OutgoingPaymentStarted {
+    pub contract_id: ContractId,
+    pub invoice_amount: Amount,
+}
```
elsirion: Should contain operation id so e.g. invoice can be looked up from operation log.

elsirion (commented):
Looks good, but probably best to include the operation id so additional meta data can be looked up.

gateway/ln-gateway/src/state_machine/events.rs
m1sterc001guy: Fixed!

elsirion: You only added the operation id to the start event, which is ok I guess as long as consumers keep an index `payment_hash -> operation_id`. Worth considering the storage/ease of consumption trade-off though, cc @dpc.

elsirion (approved):

PR #6503, opened 2024-12-04
feat: fix missing loop break label

elsirion:
Likely related to flakiness reported in #6502, although I don't know how this *ever* worked the way it was :exploding_head:

### DIFF
--- a/fedimint-client/src/oplog.rs
+++ b/fedimint-client/src/oplog.rs
@@ -121,7 +121,9 @@ impl OperationLog {
         // not introduce a find_by_range_rev function we have to jump through some
         // hoops, see also the comments in rev_epoch_ranges.
         // TODO: Implement using find_by_range_rev if ever introduced
-        for key_range_rev in rev_epoch_ranges(start_after_key, oldest_entry_key, EPOCH_DURATION) {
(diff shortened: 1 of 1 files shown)


dpc (approved):

dpc:
@elsirion Given a random nature of the problem, my bet would be on something non-deterministic, like timestamp or ID comparison.

dpc:
@elsirion Would probably be good to have a test that generates like 100 random events, then paginates them 3 by 3, then compares that it got all the same events in same order. Just something heavy duty like that.

dpc:
Hmm... this test passes deterministically 100% of the time(?) locally?!

dpc:
So how did this pass https://github.com/fedimint/fedimint/actions/runs/12116242638/job/33776213978 . It says it ran the unit tests ...

dpc:
I have re-run it and it failed. O_O 

https://github.com/fedimint/fedimint/actions/runs/12116242638/job/33937424126

dpc:
It's going to be hard to tell now why did that one particular run lie to us and pass. Since everything seems to be correctly failing now, such a one-time event does not scare me, but making a mental note that it happened.

PR #6504, opened 2024-12-04
refactor: ClientHandle.inner is not wrapped in an Option

tvolk131:
Previously `ClientHandle.inner` was an `Option<Arc<Client>>`. I believe the reason here was to allow for `ClientHandle::shutdown_inner()` to call `Option::take()` on `ClientHandle.inner` and explicitly drop the `Arc<Client>`. However, this required sprinkling a lot of `.expect()`s around. By turning `ClientHandle.inner` into a `Arc<Client>`, we can remove those.

If this PR makes sense, then we should be able to merge `ClientHandle` and `ClientHandleArc`.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -651,22 +651,19 @@ fn states_add_instance(
 /// Put this in an Arc to clone it (see [`ClientHandleArc`]).
 #[derive(Debug)]
 pub struct ClientHandle {
-    inner: Option<Arc<Client>>,
+    inner: Arc<Client>,
 }
 
 /// An alias for a reference counted [`ClientHandle`]
 pub type ClientHandleArc = Arc<ClientHandle>;
 
 impl ClientHandle {
-    /// Create
     fn new(inner: Arc<Client>) -> Self {
-        ClientHandle {
-            inner: inner.into(),
-        }
+        ClientHandle { inner }
     }
 
     fn as_inner(&self) -> &Arc<Client> {
-        self.inner.as_ref().expect("Inner always set")
+        &self.inner
     }
 
     pub fn start_executor(&self) {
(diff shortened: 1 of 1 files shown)


fedimint-client/src/lib.rs
```diff
@@ -792,19 +776,14 @@ impl ClientWeak {
 /// `Arc<Client>`s such that at least one `Executor` never gets dropped.
 impl Drop for ClientHandle {
     fn drop(&mut self) {
-        if self.inner.is_none() {
```
tvolk131, line 795: Note for reviewer: Previously, `self.inner` being `None` indicated that the executor was already stopped. `Since `inner` is no longer an `Option` and we don't ever return early here anymore, the call to `stop_executor()` below can be called after the executor has already been stopped. This will result in the following code being hit:

https://github.com/fedimint/fedimint/blob/7b7bf0f45574911e8348c68f2abc5c61b36b117c/fedimint-client/src/sm/executor.rs#L126-L130

This will produce an extra log statement, but doesn't affect anything otherwise.

fedimint-client/src/lib.rs
```diff
@@ -678,18 +675,12 @@ impl ClientHandle {
         self.shutdown_inner().await;
     }
 
+    /// Shutdown the client. `self` MUST be dropped after this call.
```
tvolk131, line 678: Note for reviewer: The API contract documented here was already applicable before this PR, since it was expected that `self.inner` was always `Some` (hence all of the `expect()`s we're removing) except in the `Drop` impl of `ClientHandle`

dpc (changes requested):
IIRC. There was a reason we had an `Option` there and it had to do with being able to drop the inside and wait for clear shutdown of the last handle. For that we need ability to move out from `self.inner`. It's a mistake that it isn't better document, though it might be recoverable via `git blame` on relevant code.

Generally, this seems like a waste of time and asking for introducing bugs and problem. Since we messed with this code last time, everything was working perfectly fine w.r.t to shutdown, so unless there is some clear benefit/goal in mind other than just simplification for the sake of it, I don't want to spend time on it.

PR #6512, opened 2024-12-05
feat(wallet-server): ability to source fee rates from external sources

dpc:
Introduces a `FM_WALLET_FEERATE_SOURCES` env var which allows specifing a list of urls along with `jq` filter syntax to extract the sats/vB feerate.

A default value is (will be - the exact contents are to be determined) provided per-network.

All sources are queried, along with the bitcoind, and a median is used to get the actual feerate to vote on during consensus.

Can be tested with:

```
env RUST_LOG=fm=info,fm::module::wallet=debug FM_WALLET_FEERATE_SOURCES="https://mempool.space/api/v1/fees/recommended#.halfHourFee;https://blockstream.info/api/fee-estimates#.\"1\"" j devimint-env
```

### DIFF
--- a/modules/fedimint-wallet-common/src/lib.rs
+++ b/modules/fedimint-wallet-common/src/lib.rs
@@ -480,12 +480,14 @@ plugin_types_trait_impl_common!(
     WalletOutputError
 );
 
-#[derive(Debug, Error, Encodable, Decodable, Hash, Clone, Eq, PartialEq)]
+#[derive(Debug, Error, Clone)]
 pub enum WalletCreationError {
     #[error("Connected bitcoind is on wrong network, expected {0}, got {1}")]
     WrongNetwork(NetworkLegacyEncodingWrapper, NetworkLegacyEncodingWrapper),
     #[error("Error querying bitcoind: {0}")]
     RpcError(String),
+    #[error("Feerate source error: {0}")]
+    FeerateSourceError(String),
 }
 
 #[derive(Debug, Error, Encodable, Decodable, Hash, Clone, Eq, PartialEq)]
--- a/modules/fedimint-wallet-server/src/feerate_source.rs
+++ b/modules/fedimint-wallet-server/src/feerate_source.rs
@@ -1,0 +1,166 @@
+use std::str::FromStr;
+
+use anyhow::{anyhow, bail, Result};
+use fedimint_bitcoind::DynBitcoindRpc;
+use fedimint_core::util::SafeUrl;
+use fedimint_core::{apply, async_trait_maybe_send, Feerate};
+use fedimint_logging::LOG_MODULE_WALLET;
+use fedimint_wallet_common::CONFIRMATION_TARGET;
+use jaq_core::load::{Arena, File, Loader};
+use jaq_core::{Ctx, Native, RcIter};
+use jaq_json::Val;
+use tracing::{debug, trace};
+
+/// A feerate that we don't expect to ever happen in practice, that we are
+/// going to reject from a source to help catching mistakes and
+/// misconfigurations.
+const FEERATE_SOURCE_MAX_FEERATE_SATS_PER_VB: f64 = 10_000.0;
+
+/// Like [`FEERATE_SOURCE_MAX_FEERATE_SATS_PER_VB`], but minimum one we accept
+const FEERATE_SOURCE_MIN_FEERATE_SATS_PER_VB: f64 = 1.0;
+
+#[apply(async_trait_maybe_send!)]
+pub trait FeeRateSource: Send + Sync {
+    fn name(&self) -> String;
+    async fn fetch(&self) -> Result<Feerate>;
+}
+
+#[apply(async_trait_maybe_send!)]
+impl FeeRateSource for DynBitcoindRpc {
+    fn name(&self) -> String {
+        self.get_bitcoin_rpc_config().kind
+    }
+
+    async fn fetch(&self) -> Result<Feerate> {
+        self.get_fee_rate(CONFIRMATION_TARGET)
+            .await?
+            .ok_or_else(|| anyhow!("bitcoind did not return any feerate"))
+    }
+}
+
+pub struct FetchJson {
+    filter: jaq_core::Filter<Native<Val>>,
+    source_url: SafeUrl,
+}
+
+impl FetchJson {
+    pub fn from_str(source_str: &str) -> Result<Self> {
+        let (source_url, code) = {
+            let (url, code) = match source_str.split_once('#') {
+                Some(val) => val,
+                None => (source_str, "."),
+            };
+
+            (SafeUrl::parse(url)?, code)
+        };
+
+        debug!(target: LOG_MODULE_WALLET, url = %source_url, code = %code, "Setting fee rate json source");
+        let program = File { code, path: () };
+
+        let loader = Loader::new([]);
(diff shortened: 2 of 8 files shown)


modules/fedimint-wallet-server/src/feerate_source.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        trace!(target: LOG_MODULE_WALLET, name = %self.name(), resp = ?json_resp, "Got json response");
+
+        let inputs = RcIter::new(core::iter::empty());
+
+        let mut out = self
+            .filter
+            .run((Ctx::new([], &inputs), Val::from(json_resp)));
+
+        let val = out
```
dpc, line 94: This seems to return the original value, and not what I would expect after applying the filter. I must have some misconception here, and I'm asking for help from the jaq developers/community.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1547,17 +1556,31 @@ impl Wallet {
...
         task_group: &TaskGroup,
         bitcoind: &DynBitcoindRpc,
-    ) -> (watch::Receiver<Option<u32>>, watch::Receiver<Feerate>) {
+    ) -> anyhow::Result<(watch::Receiver<Option<u32>>, watch::Receiver<Feerate>)> {
         let (block_count_tx, block_count_rx) = watch::channel(None);
         let (fee_rate_tx, fee_rate_rx) = watch::channel(cfg.consensus.default_fee);
 
+        let sources = std::env::var(FM_WALLET_FEERATE_SOURCES_ENV)
+            .unwrap_or_else(|_| match cfg.consensus.network.0 {
+                Network::Bitcoin => "https://mempool.space/api/v1/fees/recommended#.;https://blockstream.info/api/fee-estimates#.%221%22".to_owned(),
```
dpc, line 1565: For mainnet, these would be the defaults.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1547,17 +1556,31 @@ impl Wallet {
...
+        let feerates = Arc::new(std::sync::Mutex::new(vec![None; sources.len()]));
+
         let mut desired_interval = tokio::time::interval(if is_running_in_test_env() {
             // In devimint, the setup is blocked by detecting block height changes,
             // and polling more often is not an issue.
             debug!(target: LOG_MODULE_WALLET, "Running in devimint, using fast node polling");
             Duration::from_millis(100)
         } else {
-            Duration::from_secs(10)
+            Duration::from_secs(30)
```
dpc, line 1583: We might need to query for block height and fee rate on different interverals, if this is a problem.

modules/fedimint-wallet-server/src/feerate_source.rs
elsirion: It works for me after refactoring the logic to make it testable: bc053cde23775c9848fc545cdf5a0fa91c6f80f8

The lack of unit-testability appeared to me when wondering what went wrong here, probably worth going even a step further than I did and extract the jq functionality into some pure functions that can be tested independently.

modules/fedimint-wallet-server/src/feerate_source.rs
```diff
@@ -0,0 +1,126 @@
...
+        };
+        debug!(target: LOG_MODULE_WALLET, name = %self.name(), rate_sats_vb = %rate, "Got fee rate");
+
+        if rate <= 0.0 {
+            bail!("Fee rate returned by source not positive: {rate}")
+        }
+
+        if 10_000.0 <= rate {
+            bail!("Fee rate returned by source too large: {rate}")
+        }
```
elsirion, line 118: Arbitrary, should be a constant with explanation of the choice

modules/fedimint-wallet-server/src/feerate_source.rs
```diff
@@ -0,0 +1,126 @@
...
+            Val::Num(rate) => FromStr::from_str(&rate)?,
+            _ => {
+                bail!("Value returned by feerate source has invalid type: {val:?}");
+            }
+        };
+        debug!(target: LOG_MODULE_WALLET, name = %self.name(), rate_sats_vb = %rate, "Got fee rate");
+
+        if rate <= 0.0 {
+            bail!("Fee rate returned by source not positive: {rate}")
+        }
```
elsirion, line 114: Isn't there even a standardness rule about min fees? We should probably set that as the minimum.

modules/fedimint-wallet-server/src/lib.rs
elsirion: ~I think I found the problem: you don't use `SafeUrl` for parsing out the fragment, so you don't URL-decode `%221` etc., so jaq can't interpret it. If that's the case it would also point towards not properly handling errors?~

Never mind, second commit changes this

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1547,10 +1556,24 @@ impl Wallet {
...
+                _ => String::new(),
+            })
+            .split(';')
+            .filter(|s| !s.is_empty())
+            .map(|s| Ok(Box::new(FetchJson::from_str(s)?) as Box<dyn FeeRateSource>))
+            .chain(iter::once(Ok(
+                Box::new(bitcoind.clone()) as Box<dyn FeeRateSource>
+            )))
+            .collect::<anyhow::Result<Vec<Box<dyn FeeRateSource>>>>()?;
+        let feerates = Arc::new(std::sync::Mutex::new(vec![None; sources.len()]));
```
elsirion, line 1575: Why does this need to be mutated at runtime?

modules/fedimint-wallet-server/src/feerate_source.rs
dpc: Oh, sorry. I have fixed it yesterday and forgot to update the description and the comment.

modules/fedimint-wallet-server/src/feerate_source.rs
```diff
@@ -0,0 +1,126 @@
...
+
+pub struct FetchJson {
+    filter: jaq_core::Filter<Native<Val>>,
+    source_url: SafeUrl,
+}
+
+impl FetchJson {
+    pub fn from_str(source_str: &str) -> Result<Self> {
+        let (source_url, code) = {
+            let (url, code) = match source_str.split_once('#') {
```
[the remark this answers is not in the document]
dpc: I did before, then got tired of dealing with escaping issues (fragment gets url-percent-encoded even though the docs say it doesn't).

I guess it doesn't  hurt to keep these extra functions, just like `Url` has them.

modules/fedimint-wallet-common/src/lib.rs
```diff
@@ -480,12 +480,15 @@ plugin_types_trait_impl_common!(
     WalletOutputError
 );
 
-#[derive(Debug, Error, Encodable, Decodable, Hash, Clone, Eq, PartialEq)]
+// #[derive(Debug, Error, Encodable, Decodable, Hash, Clone, Eq, PartialEq)]
```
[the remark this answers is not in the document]
elsirion: @dpc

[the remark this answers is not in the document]
dpc: Done now.

elsirion (approved):

PR #6514, opened 2024-12-05
fix: ignore RUSTSEC-2024-0402 which we don't care about

dpc:


m1sterc001guy (approved):

elsirion:
We should check if any consensus critical dep depends on it transitively.

dpc:
Doesn't look like it. AlephBFT does not for sure.

PR #6515, opened 2024-12-05
fix: bump hashbrown to v0.15.2

tvolk131:
Fixes this vulnerability which is causing cargo audit to fail: https://rustsec.org/advisories/RUSTSEC-2024-0402


m1sterc001guy:
This is probably a better solution https://github.com/fedimint/fedimint/pull/6514

PR #6516, opened 2024-12-05
feat: allow customization of LNv2 Lightning and Transaction Fees

m1sterc001guy:
I think it is important that we allow customization of the various fees in the gateway for LNv2. Not all federations will be run for profit (private federations, smaller communities, etc) and I think gateways should have the option to change their fees to also be cheaper. LNv2 lightning fees are currently very expensive (1.5%), which is often higher than Lightning.

This PR also adds customization for the "transaction fees" which are applied to every transaction (send or receive).

### DIFF
--- a/gateway/cli/src/config_commands.rs
+++ b/gateway/cli/src/config_commands.rs
@@ -1,0 +1,86 @@
+use clap::Subcommand;
+use fedimint_core::config::FederationId;
+use fedimint_core::Amount;
+use ln_gateway::rpc::rpc_client::GatewayRpcClient;
+use ln_gateway::rpc::{ConfigPayload, SetFeesPayload};
+
+use crate::print_response;
+
+#[derive(Subcommand)]
+pub enum ConfigCommands {
+    /// Gets each connected federation's JSON client config
+    ClientConfig {
+        #[clap(long)]
+        federation_id: Option<FederationId>,
+    },
+    /// Gets the Gateway's configured configuration for each federation
+    Display {
+        #[clap(long)]
+        federation_id: Option<FederationId>,
+    },
+    /// Set the gateway's lightning or transaction fees
+    SetFees {
+        #[clap(long)]
+        federation_id: Option<FederationId>,
+
+        #[clap(long)]
+        ln_base: Option<Amount>,
+
+        #[clap(long)]
+        ln_ppm: Option<u64>,
+
+        #[clap(long)]
+        tx_base: Option<Amount>,
+
+        #[clap(long)]
+        tx_ppm: Option<u64>,
+    },
+}
+
+impl ConfigCommands {
+    pub async fn handle(
+        self,
+        create_client: impl Fn() -> GatewayRpcClient + Send + Sync,
+    ) -> anyhow::Result<()> {
+        match self {
+            Self::ClientConfig { federation_id } => {
+                let response = create_client()
+                    .get_config(ConfigPayload { federation_id })
+                    .await?;
+
+                print_response(response);
+            }
+            Self::Display { federation_id } => {
+                let info = create_client().get_info().await?;
+                let federations = if let Some(id) = federation_id {
+                    info.federations
+                        .into_iter()
+                        .filter(|f| id == f.federation_id)
+                        .collect::<Vec<_>>()
+                } else {
+                    info.federations
+                };
+                print_response(federations);
+            }
+            Self::SetFees {
+                federation_id,
+                ln_base,
+                ln_ppm,
+                tx_base,
+                tx_ppm,
+            } => {
+                create_client()
+                    .set_fees(SetFeesPayload {
+                        federation_id,
+                        lightning_base: ln_base,
+                        lightning_parts_per_million: ln_ppm,
+                        transaction_base: tx_base,
+                        transaction_parts_per_million: tx_ppm,
+                    })
+                    .await?;
+            }
+        }
+
+        Ok(())
+    }
+}
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -13,6 +13,7 @@ use fedimint_core::invite_code::InviteCode;
 use fedimint_core::{impl_db_lookup, impl_db_record, push_db_pair_items, secp256k1, Amount};
 use fedimint_ln_common::serde_routing_fees;
 use fedimint_lnv2_common::contracts::{IncomingContract, PaymentImage};
+use fedimint_lnv2_common::gateway_api::PaymentFee;
 use futures::{FutureExt, StreamExt};
 use lightning_invoice::RoutingFees;
 use rand::rngs::OsRng;
@@ -45,10 +46,6 @@ pub trait GatewayDbtxNcExt {
     /// it does not exist. Remember to commit the transaction after calling this
     /// method.
     async fn load_or_create_gateway_keypair(&mut self) -> Keypair;
-
-    async fn load_gateway_config(&mut self) -> Option<GatewayConfiguration>;
-
-    async fn set_gateway_config(&mut self, gateway_config: &GatewayConfiguration);
 
     async fn save_new_preimage_authentication(
         &mut self,
@@ -139,15 +136,6 @@ impl<Cap: Send> GatewayDbtxNcExt for DatabaseTransaction<'_, Cap> {
         }
     }
 
-    async fn load_gateway_config(&mut self) -> Option<GatewayConfiguration> {
-        self.get_value(&GatewayConfigurationKey).await
-    }
-
-    async fn set_gateway_config(&mut self, gateway_config: &GatewayConfiguration) {
-        self.insert_entry(&GatewayConfigurationKey, gateway_config)
-            .await;
-    }
-
     async fn save_new_preimage_authentication(
         &mut self,
         payment_hash: sha256::Hash,
@@ -212,14 +200,6 @@ impl<Cap: Send> GatewayDbtxNcExt for DatabaseTransaction<'_, Cap> {
                         "Federation Config"
                     );
                 }
-                DbKeyPrefix::GatewayConfiguration => {
-                    if let Some(gateway_config) = self.load_gateway_config().await {
-                        gateway_items.insert(
-                            "Gateway Configuration".to_string(),
-                            Box::new(gateway_config),
-                        );
-                    }
-                }
                 DbKeyPrefix::GatewayPublicKey => {
                     if let Some(public_key) = self.load_gateway_keypair().await {
                         gateway_items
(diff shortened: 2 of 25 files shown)


elsirion (approved):
Second review by someone with more LN context would be useful

gateway/ln-gateway/src/lib.rs
```diff
@@ -1956,20 +1962,31 @@ impl Gateway {
...
-                send_fee_default: PaymentFee::SEND_FEE_LIMIT,
+                send_fee_default: routing_fees.into(),
                 // The base fee ensures that the gateway does not loose sats sending the payment due
                 // to fees paid on the transaction claiming the outgoing contract or
                 // subsequent transactions spending the newly issued ecash
-                send_fee_minimum: PaymentFee {
-                    base: Amount::from_sats(50),
-                    parts_per_million: 5_000,
-                },
+                send_fee_minimum: PaymentFee::SEND_FEE_DEFAULT,
```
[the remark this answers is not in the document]
m1sterc001guy: Ended up implementing this in the current PR

gateway/ln-gateway/src/lib.rs
```diff
@@ -1709,7 +1749,7 @@ impl Gateway {
         let routing_fees = gateway_parameters
             .fees
             .clone()
-            .unwrap_or(GatewayFee(PaymentFee::SEND_FEE_DEFAULT.into()));
+            .unwrap_or(GatewayFee(PaymentFee::TRANSACTION_FEE_DEFAULT.into()));
```
joschisan, line 1752: Wrapping this in a GatewayFee here seems unnecessary.

modules/fedimint-lnv2-common/src/gateway_api.rs
```diff
@@ -188,11 +192,9 @@ impl RoutingInfo {
...
         } else {
             (
-                PaymentFee {
-                    base: self.send_fee_default.base + self.send_fee_minimum.base,
-                    parts_per_million: self.send_fee_default.parts_per_million
-                        + self.send_fee_minimum.parts_per_million,
-                },
+                self.send_fee_default
+                    .clone()
+                    .add(self.send_fee_minimum.clone()),
```
joschisan, line 197: You can not change client side behavior like this. This addition needs to be done on the gateways side instead.

modules/fedimint-lnv2-common/src/gateway_api.rs
```diff
@@ -245,6 +247,16 @@ impl PaymentFee {
...
+impl Add for PaymentFee {
+    type Output = PaymentFee;
+    fn add(self, rhs: Self) -> Self::Output {
+        PaymentFee {
+            base: self.base + rhs.base,
+            parts_per_million: self.parts_per_million + rhs.parts_per_million,
+        }
+    }
+}
+
```
joschisan, line 259: Can this not be derived? If not, as its only used in one place please use it there directly instead of implementing a trait.

modules/fedimint-lnv2-common/src/gateway_api.rs
```diff
@@ -269,3 +281,20 @@ impl std::fmt::Display for PaymentFee {
...
+            .next()
+            .context("missing base fee in millisatoshis")?
+            .parse()?;
+        let parts_per_million = parts.next().context("missing parts per million")?.parse()?;
+        Ok(PaymentFee {
+            base,
+            parts_per_million,
+        })
+    }
+}
```
joschisan, line 300: Please avoid custom implementations like this. Instantiate directly where needed.

gateway/ln-gateway/src/db.rs
```diff
@@ -285,6 +307,8 @@ pub struct FederationConfig {
     pub timelock_delta: u64,
     #[serde(with = "serde_routing_fees")]
     pub fees: RoutingFees,
+    #[serde(with = "serde_routing_fees")]
+    pub transaction_fees: RoutingFees,
```
joschisan, line 311: Remove the RoutingFee struct here, consistently used PaymentFee everywhere instead. Ideally remove RoutingFee completely.

gateway/ln-gateway/src/lib.rs
```diff
@@ -1964,28 +2004,30 @@ impl Gateway {
...
-        } else {
-            PaymentFee::SEND_FEE_DEFAULT.into()
-        };
+        let routing_fees = fed_config
+            .routing_fees
+            .unwrap_or(PaymentFee::TRANSACTION_FEE_DEFAULT);
+
+        let transaction_fees = fed_config
+            .transaction_fees
+            .unwrap_or(PaymentFee::TRANSACTION_FEE_DEFAULT);
```
joschisan, line 2013: You can get this from the config directly instead of the FederationInfo.

joschisan (changes requested):

modules/fedimint-lnv2-common/src/gateway_api.rs
m1sterc001guy: It cannot be derived. It is used in multiple places (when returning the `RoutingInfo` and when checking if the newly configured fee is valid). Going to keep as is

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Fixed

modules/fedimint-lnv2-common/src/gateway_api.rs
m1sterc001guy: Fixed, moved to addition to the gateway

modules/fedimint-lnv2-common/src/gateway_api.rs
m1sterc001guy: Removed

gateway/ln-gateway/src/lib.rs
m1sterc001guy: Fixed

gateway/ln-gateway/src/db.rs
m1sterc001guy: I removed as much I reasonably could. There's two reasons completely removing it is difficult:

1) Its currently in DB structs necessary for db migrations
2) It is used in the `LightningGateway` struct that is also persisted on the server side, and I don't think doing a guardian db migration is worth it.

Right now `PaymentFee` is used everywhere except in some db migrations and the LNv1 code. So once LNv1 is removed it will be gone.

m1sterc001guy:
I removed the feature to set the default routing fees via the CLI or environment variable. I think it is better to keep things simple and it is not a critical feature.

gateway/ln-gateway/src/db.rs
```diff
@@ -341,16 +373,25 @@ pub struct GatewayConfigurationV1 {
...
     pub network: NetworkLegacyEncodingWrapper,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable)]
+pub struct GatewayConfigurationKey;
+
+#[derive(Debug, Clone, Eq, PartialEq, Encodable, Decodable, Serialize, Deserialize)]
+pub struct GatewayConfiguration {
+    pub num_route_hints: u32,
+    pub network: NetworkLegacyEncodingWrapper,
```
joschisan, line 392: Please migrate to the new network encoding while you are at it.

joschisan:
As PaymentFee is singular, for consistency we should use the names transaction_fee and routing_fee for fields and variables. Furthermore, maybe lightning_fee instead of routing_fee is clearer?

gateway/ln-gateway/src/db.rs
m1sterc001guy: Maybe we should just remove it altogether and make it an environment variable? It doesn't make sense to me to put this in the `GatewayConfiguration`, since that implies that you'd want to change it at runtime without restarting. To me, changing networks seems fine to require a restart.

Bonus is that `num_route_hints` would be the only thing left here and that will go away once LNv1 is gone

gateway/cli/src/config_commands.rs
```diff
@@ -0,0 +1,55 @@
...
+            Self::SetFees {
+                fed_id,
+                ln_base,
+                ln_ppm,
+                tx_base,
+                tx_ppm,
+            } => {
+                create_client()
+                    .set_fees(SetFeePayload {
+                        fed_id,
```
joschisan, line 43: Could rename in full federation_id for consistency.

m1sterc001guy:
> Also I cannot find what fees are set when connecting to a federation initially. Are we enabling the operator to configure this?

It is just set to a constant now, the operator can change it later. I don't think having the default customizable is worth it.

> cfg set-fees --fed-id ... --ln-base 50 --ln-ppm 1000 --tx-base 50 --tx-ppm 1000

This is done now.

I also took the opportunity to clean up the `set-configuration` endpoint (it is removed entirely). The `GatewayConfiguration` is now completely gone as well. The only two remaining parameters were `network` and `num_route_hints`, neither of which I think are worth being able to change at runtime (and `num_route_hints` will be removed with LNv1). These are now just environment variables.

gateway/ln-gateway/src/db.rs
m1sterc001guy: This is removed entirely

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -104,7 +103,8 @@ pub struct FederationInfo {
     pub federation_name: Option<String>,
     pub balance_msat: Amount,
     pub federation_index: u64,
-    pub routing_fees: Option<FederationRoutingFees>,
+    pub lightning_fee: Option<PaymentFee>,
+    pub transaction_fee: Option<PaymentFee>,
```
joschisan, line 107: Why are those optional?

gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -38,7 +37,7 @@ pub const CLOSE_CHANNELS_WITH_PEER_ENDPOINT: &str = "/close_channels_with_peer";
 pub const PAY_INVOICE_FOR_OPERATOR_ENDPOINT: &str = "/pay_invoice_for_operator";
 pub const PAYMENT_LOG_ENDPOINT: &str = "/payment_log";
 pub const RECEIVE_ECASH_ENDPOINT: &str = "/receive_ecash";
-pub const SET_CONFIGURATION_ENDPOINT: &str = "/set_configuration";
+pub const SET_FEE_ENDPOINT: &str = "/set_fee";
```
joschisan, line 40: Should be set_fees for consistency.

joschisan:
Seems like the only thing missing is to refactor the general config command to cfg display --fed-id ... instead

gateway/ln-gateway/src/rpc/mod.rs
m1sterc001guy: Fixed

gateway/ln-gateway/src/rpc/mod.rs
m1sterc001guy: Fixed

gateway/cli/src/config_commands.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use crate::print_response;
+
+#[derive(Subcommand)]
+pub enum ConfigCommands {
+    /// Gets each connected federation's JSON client config
+    JsonConfig {
+        #[clap(long)]
+        federation_id: Option<FederationId>,
+    },
```
joschisan, line 15: Maybe call this ClientConfig?

joschisan (approved):

joschisan (approved):

bradleystachurski (approved):

PR #6517, opened 2024-12-06
fix: remove usage of assume_checked()

### LINKED ISSUE #5203: Remove usage of `assume_checked()`

tvolk131:
Fixes #5203

Doesn't remove all usage of `assume_checked()`, since some uses are acceptable, such as in tests

### DIFF
--- a/fedimint-core/src/encoding/btc.rs
+++ b/fedimint-core/src/encoding/btc.rs
@@ -203,10 +203,13 @@ impl Decodable for bitcoin::Amount {
 impl Encodable for bitcoin::Address<NetworkUnchecked> {
     fn consensus_encode<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, Error> {
         let mut len = 0;
-        len += NetworkLegacyEncodingWrapper(get_network_for_address(self.as_unchecked()))
-            .consensus_encode(writer)?;
+        len +=
+            NetworkLegacyEncodingWrapper(get_network_for_address(self)).consensus_encode(writer)?;
         len += self
             .clone()
+            // We need an `Address<NetworkChecked>` in order to get the script pubkey.
+            // Calling `assume_checked` is generally a bad idea, but it's safe here where we're
+            // encoding the address because addresses are always decoded as unchecked.
             .assume_checked()
             .script_pubkey()
             .consensus_encode(writer)?;
@@ -225,7 +228,7 @@ impl Decodable for bitcoin::Address<NetworkUnchecked> {
         let address = bitcoin::Address::from_script(&script_pk, network)
             .map_err(|e| DecodeError::new_custom(e.into()))?;
 
-        Ok(address.as_unchecked().clone())
+        Ok(address.into_unchecked())
     }
 }
 
--- a/gateway/ln-gateway/src/lib.rs
+++ b/gateway/ln-gateway/src/lib.rs
@@ -63,7 +63,9 @@ use fedimint_core::secp256k1::PublicKey;
 use fedimint_core::task::{sleep, TaskGroup, TaskHandle, TaskShutdownToken};
 use fedimint_core::time::duration_since_epoch;
 use fedimint_core::util::{SafeUrl, Spanned};
-use fedimint_core::{fedimint_build_code_version_env, Amount, BitcoinAmountOrAll};
+use fedimint_core::{
+    fedimint_build_code_version_env, get_network_for_address, Amount, BitcoinAmountOrAll,
+};
 use fedimint_eventlog::{DBTransactionEventLogExt, EventLogId};
 use fedimint_ln_common::config::{GatewayFee, LightningClientConfig};
 use fedimint_ln_common::contracts::Preimage;
@@ -886,6 +888,15 @@ impl Gateway {
             address,
             federation_id,
         } = payload;
+
+        let address_network = get_network_for_address(&address);
+        let gateway_network = self.gateway_config.read().await.network.0;
+        let Ok(address) = address.require_network(gateway_network) else {
+            return Err(AdminGatewayError::WithdrawError {
+                failure_reason: format!("Gateway is running on network {gateway_network}, but provided withdraw address is for network {address_network}")
+            });
+        };
+
         let client = self.select_client(federation_id).await?;
         let wallet_module = client.value().get_first_module::<WalletClientModule>()?;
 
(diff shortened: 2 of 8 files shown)


fedimint-core/src/encoding/btc.rs
```diff
@@ -203,10 +203,13 @@ impl Decodable for bitcoin::Amount {
 impl Encodable for bitcoin::Address<NetworkUnchecked> {
     fn consensus_encode<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, Error> {
         let mut len = 0;
-        len += NetworkLegacyEncodingWrapper(get_network_for_address(self.as_unchecked()))
-            .consensus_encode(writer)?;
+        len +=
+            NetworkLegacyEncodingWrapper(get_network_for_address(self)).consensus_encode(writer)?;
         len += self
             .clone()
+            // We need an `Address<NetworkChecked>` in order to get the script pubkey.
```
dpc, line 210: `assume_checked` is only a risk if the address is untrusted (provided from the outsdie). If we're encoding something, it is "internal" data, and should have been validated already, so it's OK.

gateway/ln-gateway/src/lib.rs
```diff
@@ -889,6 +891,15 @@ impl Gateway {
             address,
             federation_id,
         } = payload;
+
+        let address_network = get_network_for_address(&address);
+        let gateway_network = self.gateway_config.read().await.network.0;
+        let Ok(address) = address.require_network(gateway_network) else {
+            return Err(AdminGatewayError::WithdrawError {
+                    failure_reason: format!("Gateway is running on network {gateway_network}, but provided withdraw address is for network {address_network}")
+                });
```
dpc, line 900: Nit: indent seems slightly off.

gateway/ln-gateway/src/lib.rs
```diff
@@ -900,7 +911,7 @@ impl Gateway {
                 let balance =
                     bitcoin::Amount::from_sat(client.value().get_balance().await.msats / 1000);
                 let fees = wallet_module
-                    .get_withdraw_fees(address.clone(), balance)
+                    .get_withdraw_fees(address.as_unchecked().clone(), balance)
```
dpc, line 914: Shouldn't we modify methods to take "checked" address instead? We should generally use `Checked` all over the code, except in deserialization / taking input from the user, where we sould make sure to validate them.

gateway/ln-gateway/src/lib.rs
tvolk131: Fixed

gateway/ln-gateway/src/lib.rs
tvolk131: Good point! Fixed.

fedimint-core/src/encoding/btc.rs
tvolk131: Currently we implement `Encodable` and `Decodable` for unchecked addresses but not checked addresses, requiring us to re-check the network for addresses that, as you mention, are trusted to be already checked. What do you think about changing the implementations to be for checked addresses instead? That way, we could remove the few places where we're calling `into_unchecked()` or `as_unchecked()` when encoding data and likewise remove the need to re-check addresses when decoding them.

tvolk131: I made #6550 to showcase this. However, since it requires the removal of `serde::Deserialize` for structs containing a checked address, it probably isn't worth it.

dpc: @tvolk131 Maybe we should parametrize with aliases?


```
struct SomeAggregateType<V = NetworkUnchecked> {
  ...
  addresss: Address<V>,
  ...
}

type CheckedSomeAggregateType = SomeAggregateType<NetworkChecked>;

impl SomeAggregateType<NetworkUnchecked> {
  fn validate_network(...) -> ... {

  }
}
```

Database definitions can just use `CheckedSomeAggregateType` everywhere, while networking code the (unchecked) `SomeAggregateType`?

@tvolk131

elsirion: Feels like too much overhead to me tbh.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1408,15 +1410,30 @@ impl Wallet {
...
+                    .clone()
+                    .require_network(self.cfg.consensus.network.0)
+                    .map_err(|_| {
+                        WalletOutputError::WrongNetwork(
+                            self.cfg.consensus.network.clone(),
+                            NetworkLegacyEncodingWrapper(get_network_for_address(
+                                &peg_out.recipient,
+                            )),
+                        )
+                    })?;
```
elsirion, line 1425: This is a consensus change unless proven otherwise. We can forbid submitting withdrawal addresses with incorrect network, but forbidding them in consensus may break guardian recovery from a peer's history if someone already submitted such a transaction.

We could phase it in through a consensus version vote, something we already implemented in the wallet module anyway. cc @bradleystachurski @joschisan

elsirion (changes requested):

modules/fedimint-wallet-server/src/lib.rs
tvolk131: Ahh good catch! I wonder if this is worth a consensus change... I think it would make sense to drop a comment describing why we're calling `assume_checked()` and otherwise leave this code alone. We're eventually going to replace it with wallet v2 anyway.

elsirion: It's also more of a user foot-gun, for everything that matters we use script anyway, so nothing should blow up even if someone submits a wrong network address on purpose (the client won't do that on its own).

tvolk131: Fixed! Also we already do reject transactions with the wrong network in `validate_tx()` which we call in `process_output()`:

https://github.com/fedimint/fedimint/blob/3b27f563b86ca3cd68e4f5b2399f9d10a79122bb/modules/fedimint-wallet-server/src/lib.rs#L713-L717

elsirion (approved):

PR #6518, opened 2024-12-06
fix: check that full balance becomes available to the client

joschisan:

### DIFF
--- a/modules/fedimint-lnv2-tests/bin/tests.rs
+++ b/modules/fedimint-lnv2-tests/bin/tests.rs
@@ -267,7 +267,7 @@ async fn test_payments(dev_fed: &DevJitFed) -> anyhow::Result<()> {
 
     while client.balance().await? < 9000 * 1000 {
         sleep_in_test(
-            format!("Waiting for the full balance to become available to the client"),
+            "Waiting for the full balance to become available to the client",
             Duration::from_secs(1),
         )
         .await;
--- a/modules/fedimint-lnv2-tests/bin/tests.rs
+++ b/modules/fedimint-lnv2-tests/bin/tests.rs
@@ -1,11 +1,10 @@
-use std::time::Duration;
-
+use anyhow::ensure;
 use devimint::devfed::DevJitFed;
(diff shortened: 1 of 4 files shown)


modules/fedimint-lnv2-tests/src/bin/tests.rs
```diff
@@ -260,7 +265,15 @@ async fn test_payments(dev_fed: &DevJitFed) -> anyhow::Result<()> {
...
     }
 
-    info!("Testing LDK Gateway can pay HOLD invoice...");
+    while client.balance().await? < 9000 * 1000 {
+        sleep_in_test(
+            format!("Waiting for the full balance to become available to the client"),
+            Duration::from_secs(1),
+        )
+        .await;
+    }
```
elsirion: You can use [`retry`](https://docs.fedimint.org/fedimint_core/util/fn.retry.html) here

scripts/tests/lnv2-module-test.sh
```diff
@@ -8,4 +7,4 @@ source scripts/_common.sh
 build_workspace
 add_target_dir_to_path
 
-tests all
```
elsirion, line 11: Did the `all` ever have any meaning?

joschisan: yes. we used to be able to run parts of the test suite seperately but that was removed.

modules/fedimint-lnv2-tests/bin/tests.rs
```diff
@@ -260,7 +264,18 @@ async fn test_payments(dev_fed: &DevJitFed) -> anyhow::Result<()> {
         await_receive_claimed(&client, receive_op).await?;
     }
 
-    info!("Testing LDK Gateway can pay HOLD invoice...");
+    retry(
```
m1sterc001guy, line 267: Was there a flake that prompted this change? I would think that `await_receive_claimed` would wait until the balance was been updated, so the retry wouldnt be necessary

elsirion (approved):

PR #6524, opened 2024-12-07
feat: allow specifying primary module kind

elsirion:
The primary module instance id cannot be known upfront, it may differ from federation to federation.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -2317,6 +2317,7 @@ impl ClientBuilder {
         ClientBuilder {
             module_inits: ModuleInitRegistry::new(),
             primary_module_instance: None,
+            primary_module_kind: None,
             connector: Connector::default(),
             admin_creds: None,
             db_no_decoders: db,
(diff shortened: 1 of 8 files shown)


fedimint-client/src/lib.rs
```diff
@@ -2358,6 +2361,10 @@ impl ClientBuilder {
     ///
     /// ## Panics
     /// If there was a primary module specified previously
+    #[deprecated(
+        since = "0.6.0",
+        note = "Use `with_primary_module_kind` instead, as the instance id can't be known upfront"
```
dpc, line 2366: Being able to select a specific primary module by id is theoretically useful. It's just not something that typically should be done. Maybe we should rename and document better, but ultimately keep?

dpc: Deprecating probably doesn't help much as a separate step anyway.

maan2003:
request to backport to 0.5

fedimint-client/src/lib.rs
elsirion: I'd say let's create a fn with a new name that retains the old behavior, but is strongly discouraged. I'd like to phase out the old one to get rid of old and likely incorrect usage.

dpc: Yes, we can just rugpull that name. :D

dpc (approved):

fedimint-backports:
- #6534

PR #6535, opened 2024-12-10
feat(gateway): add JWT authentication

### LINKED ISSUE #4128: Use session tokens for admin API auth

f3r10:
closes #4128 
# Motivation
Add support for JWT gateway authentication
# This PR adds one additional endpoint:
- `/auth/session` if the password is correct, this endpoint returns a JWT token that must be used in the other endpoints that require authentication.
# How to test it using the CLI:
`gateway-cli --address=http://127.0.0.1:16265/v1 --rpcpassword=theresnosecondbest get-session-jwt-auth` 

`gateway-cli --address=http://127.0.0.1:16265/v1 --rpcjwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpZCI6IjAzY2I1MTBjOWJiMzA5ODNkOTYzODk3ZWUyNjlkM2FiZmY3Nzg0YTVjMmVkYWM3YjQzMzg1NjIyMmMzZGIxYWQ1ZSIsImV4cCI6MTczMzg1ODUxN30.4FNxyn6Y-M_vfPRstxgrv4weK9MMRv31Inijw6JJo80 get-balances`

### DIFF
--- a/gateway/cli/src/main.rs
+++ b/gateway/cli/src/main.rs
@@ -29,6 +29,8 @@ struct Cli {
     /// WARNING: Passing in a password from the command line may be less secure!
     #[clap(long)]
     rpcpassword: Option<String>,
+    #[clap(long)]
+    rpcjwt: Option<String>,
 }
 
 #[derive(Subcommand)]
@@ -54,7 +56,13 @@ async fn main() -> anyhow::Result<()> {
 
     let cli = Cli::parse();
     let versioned_api = cli.address.join(V1_API_ENDPOINT)?;
-    let create_client = || GatewayRpcClient::new(versioned_api.clone(), cli.rpcpassword.clone());
+    let create_client = || {
+        GatewayRpcClient::new(
+            versioned_api.clone(),
+            cli.rpcpassword.clone(),
+            cli.rpcjwt.clone(),
+        )
+    };
 
     match cli.command {
         Commands::General(general_command) => general_command.handle(create_client).await?,
--- a/gateway/ln-gateway/src/auth_manager.rs
+++ b/gateway/ln-gateway/src/auth_manager.rs
@@ -1,42 +1,70 @@
 use std::time::UNIX_EPOCH;
 
 use fedimint_core::secp256k1::PublicKey;
-use jsonwebtoken::{encode, EncodingKey, Header};
+use jsonwebtoken::{encode, DecodingKey, EncodingKey, Header};
 use serde::{Deserialize, Serialize};
 
 const SESSION_EXPIRY_SECONDS: u64 = 60 * 30; // 30 minute
 
+#[derive(Clone)]
 pub struct AuthManager {
     ///gateway id
     gateway_id: PublicKey,
     /// A secret key to encode a JWT with
-    pub encoding_secret: [u8; 16],
+    encoding_key: EncodingKey,
+    /// A secret key to decode a JWT with
+    decoding_key: DecodingKey,
 }
 
 impl AuthManager {
     /// Create a new auth manager
-    pub fn new(encoding_secret: [u8; 16], gateway_id: PublicKey) -> Self {
+    pub fn new(
+        encoding_key: EncodingKey,
+        decoding_key: DecodingKey,
+        gateway_id: PublicKey,
+    ) -> Self {
         Self {
             gateway_id,
-            encoding_secret,
+            encoding_key,
+            decoding_key,
         }
     }
-    pub fn generate_session(&self) -> anyhow::Result<Session> {
+
+    /// generate a jwt token
+    pub fn generate_jwt(&self) -> anyhow::Result<String> {
         let session = Session::new(self.gateway_id, SESSION_EXPIRY_SECONDS);
-        Ok(session)
+        encode(&Header::default(), &session, &self.encoding_key)
+            .map_err(|_| anyhow::anyhow!("Unable to generate jwt token session"))
+        // session.encode_jwt(&self.encoding_key)
+    }
+
+    /// validate that a JWT is valid
+    pub fn is_jwt_valid(&self, token: &str) -> bool {
+        if let Ok(decoded_token_data) = jsonwebtoken::decode::<Session>(
+            token,
+            &self.decoding_key,
+            &jsonwebtoken::Validation::default(),
+        ) {
+            let now = fedimint_core::time::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("Time went backwards")
+                .as_secs();
+            return now < decoded_token_data.claims.exp;
+        }
+        false
     }
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-pub struct Session {
+struct Session {
     /// the unique identifier of the session.
-    pub id: PublicKey,
+    id: PublicKey,
     /// the expire time of the session
-    pub exp: u64,
+    exp: u64,
 }
 
 impl Session {
-    pub fn new(id: PublicKey, expiry: u64) -> Self {
+    fn new(id: PublicKey, expiry: u64) -> Self {
         let now = fedimint_core::time::now()
             .duration_since(UNIX_EPOCH)
             .expect("Time went backwards")
@@ -45,15 +73,5 @@ impl Session {
             id,
             exp: now + expiry,
         }
-    }
-
-    pub fn encode_jwt(self, encoding_secret: &[u8; 16]) -> anyhow::Result<String> {
-        let claim = self;
-        encode(
-            &Header::default(),
-            &claim,
-            &EncodingKey::from_secret(encoding_secret),
-        )
-        .map_err(|_| anyhow::anyhow!("Unable to generate jwt token session"))
     }
 }
--- a/gateway/ln-gateway/src/auth_manager.rs
+++ b/gateway/ln-gateway/src/auth_manager.rs
@@ -1,0 +1,59 @@
+use std::time::UNIX_EPOCH;
+
+use fedimint_core::secp256k1::PublicKey;
+use jsonwebtoken::{encode, EncodingKey, Header};
+use serde::{Deserialize, Serialize};
+
+const SESSION_EXPIRY_SECONDS: u64 = 60 * 30; // 30 minute
+
+pub struct AuthManager {
+    ///gateway id
+    gateway_id: PublicKey,
+    /// A secret key to encode a JWT with
+    pub encoding_secret: [u8; 16],
+}
+
+impl AuthManager {
+    /// Create a new auth manager
+    pub fn new(encoding_secret: [u8; 16], gateway_id: PublicKey) -> Self {
+        Self {
+            gateway_id,
+            encoding_secret,
+        }
+    }
+    pub fn generate_session(&self) -> anyhow::Result<Session> {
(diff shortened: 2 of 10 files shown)


gateway/cli/src/main.rs
```diff
@@ -50,7 +52,13 @@ async fn main() -> anyhow::Result<()> {
 
     let cli = Cli::parse();
     let versioned_api = cli.address.join(V1_API_ENDPOINT)?;
-    let create_client = || GatewayRpcClient::new(versioned_api.clone(), cli.rpcpassword.clone());
+    let create_client = || {
+        GatewayRpcClient::new(
+            versioned_api.clone(),
+            cli.rpcpassword.clone(),
+            cli.rpcjwt.clone(),
```
elsirion, line 59: ~Now needing both is unfortunate, ideally the password would just be an implementation detail of some function/struct providing fetching the session token. What's the plan to evolve this?~

I see, both are optional.

gateway/ln-gateway/src/lib.rs
```diff
@@ -368,13 +373,22 @@ impl Gateway {
         let task_group = TaskGroup::new();
         task_group.install_kill_handler();
 
+        let gateway_id = Self::load_or_create_gateway_id(&gateway_db).await;
+
+        // for JWT it is necessary to create an encoding secret that will be used each
+        // time a new JWT token is generated. The encoding secret ensures token's
+        // integrity and authenticity, so it is necessary to use a secure random
+        // number generator to generate strong keys.
+        let encoding_secret: [u8; 16] = rand::thread_rng().gen();
```
elsirion, line 382: The secret being genearted on the fly means a GW restart invalidates old tokens  I assume? Not a problem imo, just want to point it out so everyone is aware.

gateway/ln-gateway/src/rpc/rpc_client.rs
```diff
@@ -31,19 +32,26 @@ pub struct GatewayRpcClient {
     client: reqwest::Client,
     /// Optional gateway password
     password: Option<String>,
+    /// Optional gateway JWT
+    pub jwt_code: Option<String>,
 }
 
 impl GatewayRpcClient {
-    pub fn new(versioned_api: SafeUrl, password: Option<String>) -> Self {
+    pub fn new(versioned_api: SafeUrl, password: Option<String>, jwt_code: Option<String>) -> Self {
```
elsirion, line 40: What are the semantics of providing a password, a JWT or both? Please document. E.g. does providing the password lead to automatic requesting of a JWT whenever the previous expires?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -83,23 +85,60 @@ fn extract_bearer_token(request: &Request) -> Result<String, StatusCode> {
...
+        let now = fedimint_core::time::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_secs();
+        if now > decoded_token_data.claims.exp {
+            return Err(StatusCode::UNAUTHORIZED);
+        }
+        return Ok(next.run(request).await);
+    }
+    //TODO remove this fallback when all the callers support JWT auth
```
elsirion, line 112: Should create an issue once merged

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -83,23 +85,60 @@ fn extract_bearer_token(request: &Request) -> Result<String, StatusCode> {
...
+        &jsonwebtoken::DecodingKey::from_secret(auth_manager.encoding_secret.as_ref()),
+        &jsonwebtoken::Validation::default(),
+    ) {
+        let now = fedimint_core::time::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_secs();
+        if now > decoded_token_data.claims.exp {
+            return Err(StatusCode::UNAUTHORIZED);
+        }
```
elsirion, line 109: This could be a separate function for checking the JWT, would make the code cleaner. 

EDIT: should live inside `AuthManager` after reading the full diff

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -83,23 +85,60 @@ fn extract_bearer_token(request: &Request) -> Result<String, StatusCode> {
...
+        return Ok(next.run(request).await);
+    }
+    //TODO remove this fallback when all the callers support JWT auth
+    // fallback to check for password
     if bcrypt::verify(token, &gateway.bcrypt_password_hash.to_string())
         .expect("Bcrypt hash is valid since we just stringified it")
```
elsirion, line 115: Could also be a separate function

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -83,23 +85,60 @@ fn extract_bearer_token(request: &Request) -> Result<String, StatusCode> {
...
+        &jsonwebtoken::Validation::default(),
+    ) {
+        let now = fedimint_core::time::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_secs();
+        if now > decoded_token_data.claims.exp {
+            return Err(StatusCode::UNAUTHORIZED);
+        }
+        return Ok(next.run(request).await);
```
elsirion, line 110: This line is duplicated, we could do the following instead:

```rust
if self.check_jwt(…) || self.check_password(…) {
    Ok(next.run(request).await)
} else {
    Err(StatusCode::UNAUTHORIZED)
}
```

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -83,23 +85,60 @@ fn extract_bearer_token(request: &Request) -> Result<String, StatusCode> {
...
     Err(StatusCode::UNAUTHORIZED)
 }
 
+async fn auth_get_jwt_session(
+    Extension(gateway): Extension<Arc<Gateway>>,
+    request: Request,
+) -> Result<impl IntoResponse, AdminGatewayError> {
+    let token = extract_bearer_token(&request).map_err(|_| AdminGatewayError::Unauthorized)?;
+    if bcrypt::verify(token, &gateway.bcrypt_password_hash.to_string())
+        .expect("Bcrypt hash is valid since we just stringified it")
```
elsirion, line 128: By making password validation a fn this can also be deduplicated

gateway/ln-gateway/src/auth_manager.rs
```diff
@@ -0,0 +1,59 @@
...
+use jsonwebtoken::{encode, EncodingKey, Header};
+use serde::{Deserialize, Serialize};
+
+const SESSION_EXPIRY_SECONDS: u64 = 60 * 30; // 30 minute
+
+pub struct AuthManager {
+    ///gateway id
+    gateway_id: PublicKey,
+    /// A secret key to encode a JWT with
+    pub encoding_secret: [u8; 16],
```
elsirion, line 13: Why is this pub? Also could you make that an `EncodingKey`, so you don't have to run `EncodingKey::from_secret` each time a session is generated?

gateway/ln-gateway/src/lib.rs
```diff
@@ -197,6 +199,9 @@ pub struct Gateway {
     /// The gateway's federation manager.
     federation_manager: Arc<RwLock<FederationManager>>,
 
+    /// The gateway's authentication manager
+    auth_manager: Arc<Mutex<AuthManager>>,
```
elsirion, line 203: Why does the auth manager need to be mutated at all? Can we remove the `Acr<Mutex<_>>`?

gateway/ln-gateway/src/rpc/rpc_server.rs
```diff
@@ -83,23 +85,60 @@ fn extract_bearer_token(request: &Request) -> Result<String, StatusCode> {
...
+    {
+        let auth_manager = gateway.auth_manager.lock().await;
+        let session = auth_manager
+            .generate_session()
+            .map_err(|_| AdminGatewayError::Unauthorized)?;
+        let token = session
+            .encode_jwt(&auth_manager.encoding_secret)
+            .map_err(|_| AdminGatewayError::Unauthorized)?;
```
elsirion, line 136: Should be encapsulated as one function in the `AuthManager`. There's no reason to generate a session without signing it imo. Just call the fn something that makes it clear that the session tokens are signed.

gateway/ln-gateway/src/auth_manager.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Session {
+    /// the unique identifier of the session.
+    pub id: PublicKey,
+    /// the expire time of the session
+    pub exp: u64,
+}
```
elsirion, line 36: Member fields can be private afaik. Only the auth manager should access them to either create a session or check it.

elsirion (commented):
Thx for the PR, logic looks good, just code organization could be improved. Generally I'd recommend to think about how arguments to functions can be reduced, that way you typically get a good feeling for where a function should live (e.g. verification belongs into `AuthManager` because it has access to the secret and that's all you need to verify).

gateway/ln-gateway/src/lib.rs
f3r10: Yes, you are right. Where should I put a note about that? Would it be okay as an additional comment, or should I create an additional document for that?

gateway/ln-gateway/src/rpc/rpc_client.rs
f3r10: Providing both a password and a JWT does not make sense. 
Providing an initial password has the purpose of generating a JWT that has to be used to request private endpoints.

Maybe instead of providing optional parameters (JWT and password) would be better to provide an `enum` for each case,  WDYT?

gateway/ln-gateway/src/rpc/rpc_server.rs
f3r10: yes, @elsirion it would be better to create additional issues in this repository, and also I think it would be necessary to create an issue in the `ui` repository.

dpc:
*Comment by Tau/GPT-5*

Thanks for opening this. I did a stale-PR triage pass: the underlying feature still seems relevant (#4128 is still open and current master has no JWT/session-token admin API auth), but this particular patch is now obsolete.

The PR is based on the old `gateway/ln-gateway` and old gateway RPC client code. Current master has since split the gateway into `fedimint-gateway-{server,client,common}`, moved gateway requests through `fedimint-connectors` for both HTTP and Iroh, and added the liquidity-manager auth route split. There is no `jsonwebtoken`/`AUTH_SESSION_ENDPOINT` on master, so this is not already implemented, but landing it would require reworking the design against the current gateway/connectors architecture rather than resolving a simple conflict.

I am closing this stale PR to keep the queue actionable. The work can continue in a fresh PR, and this can always be reopened if we missed something.

Issue #6530, opened 2024-12-09
SM executor exponential backoff


dpc:

dpc:
Executor does not have visibility or control into state machines, so I guess we'll just have to improve the most common sm flows and their network retries.

dpc:
We can't do backoff, but we did bunch of api calls backoffs and improvements, so should be good enough now.

PR #6536, opened 2024-12-10
chore(client): don't await tx until it was submitted 

dpc:
Re #6530

Most of the original code was superseded/incorporated by #6538

The last optimization left is not starting await until first successful `submit`. This cuts number of requests in parallel at the beginning by half, which might be meaningful if the client starts with multiple pending txes to submit after period of being offline in a low bandwidth env.

I was a bit worried about extra latency, but it looks like it's OK - [the consensus has latency as well](https://github.com/fedimint/fedimint/discussions/6521), so client has more than enough time to make a another request. 

old:

```
### LATENCY FM PAY: min: 1.0s, avg: 1.1s, median: 1.1s, p90: 1.4s, max: 1.4s, sum: 11.2s
### LATENCY FM PAY: min: 1.0s, avg: 1.6s, median: 1.1s, p90: 6.1s, max: 6.1s, sum: 16.2s
```

new:

```
### LATENCY FM PAY: min: 1.0s, avg: 1.1s, median: 1.1s, p90: 1.2s, max: 1.2s, sum: 11.0s
### LATENCY FM PAY: min: 1.0s, avg: 1.1s, median: 1.1s, p90: 1.2s, max: 1.2s, sum: 11.2s
### LATENCY FM PAY: min: 1.0s, avg: 1.1s, median: 1.1s, p90: 1.7s, max: 1.7s, sum: 11.4s
```

### DIFF
--- a/fedimint-client/src/transaction/sm.rs
+++ b/fedimint-client/src/transaction/sm.rs
@@ -8,6 +8,7 @@ use fedimint_core::transaction::{Transaction, TransactionSubmissionOutcome};
 use fedimint_core::util::backoff_util::custom_backoff;
 use fedimint_core::util::retry;
 use fedimint_core::TransactionId;
+use tokio::sync::watch;
 
 use crate::sm::{Context, DynContext, State, StateTransition};
 use crate::{DynGlobalClientContext, DynState, TxAcceptedEvent, TxRejectedEvent};
@@ -77,6 +78,14 @@ impl State for TxSubmissionStatesSM {
         global_context: &DynGlobalClientContext,
     ) -> Vec<StateTransition<Self>> {
         let operation_id = self.operation_id;
+        // There is no point awaiting tx until it was submitted, so
+        // `trigger_created_rejected` which does the submitting will use this
+        // channel to let the `trigger_created_accepted` which does the awaiting
+        // know when it did the submission.
+        //
+        // Submitting tx does not guarantee that it will get into consensus, so the
+        // submitting need to continue.
+        let (tx_submitted_sender, tx_submitted_receiver) = watch::channel(false);
         match self.state.clone() {
             TxSubmissionStates::Created(transaction) => {
                 let txid = transaction.tx_hash();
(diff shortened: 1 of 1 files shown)


dpc:
@elsirion Seems like a likely cause of executor "getting stuck" on bad internet connections?

dpc:
Backcompat tests failing with:

```
00:00:20     2024-12-10T23:39:18.451651Z DEBUG fedimint_cli: Command failed err=Reissue failed: Transaction not accepted "The transaction had an invalid input: 1-The note is already spent"
```

for fedimintd 0.2.1.

I vaguely remember that we had an API race condition like that in the past, and possibly the timing changes in this PR are triggering it.

joschisan:
> I was a bit worried about not starting await before first succesful submit, but it looks like it's OK 

Why make this optimization? Why not already start the subscription?

dpc:
> Why make this optimization? Why not already start the subscription?

It cuts the number of calls running in parallel by two, at least initially.

fedimint-client/src/transaction/sm.rs
```diff
@@ -77,6 +78,14 @@ impl State for TxSubmissionStatesSM {
...
     ) -> Vec<StateTransition<Self>> {
         let operation_id = self.operation_id;
+        // There is no point awaiting tx until it was submitted, so
+        // `trigger_created_rejected` which does the submitting will use this
+        // channel to let the `trigger_created_accepted` which does the awaiting
+        // know when it did the submission.
+        //
+        // Submitting tx does not guarantee that it will get into consensus, so the
+        // submitting need to continue.
+        let (tx_submitted_sender, tx_submitted_receiver) = watch::channel(false);
```
elsirion, line 88: It never occurred to me that you could connect two trigger futures like that lol, slightly cursed :see_no_evil: 

Alternatively we could just make it one trigger that passes the result to the state transition fn.

elsirion (approved):
I like the general idea, the concrete impl is very clever, but also a bit scary :sweat_smile:

fedimint-client/src/transaction/sm.rs
dpc: @elsirion TBH it seemed perfectly normal and natural to me. :D

PR #6538, opened 2024-12-11
chore: api client retries on level of individual peers

joschisan:

### DIFF
--- a/fedimint-api-client/src/api/mod.rs
+++ b/fedimint-api-client/src/api/mod.rs
@@ -1,9 +1,9 @@
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::fmt::Debug;
 use std::pin::Pin;
+use std::result;
 use std::sync::Arc;
 use std::time::Duration;
-use std::{cmp, result};
 
 use anyhow::anyhow;
 #[cfg(all(feature = "tor", not(target_family = "wasm")))]
@@ -21,7 +21,6 @@ use fedimint_core::core::backup::SignedBackupRequest;
 use fedimint_core::core::{Decoder, DynOutputOutcome, ModuleInstanceId, OutputOutcome};
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::endpoint_constants::AWAIT_OUTPUT_OUTCOME_ENDPOINT;
-use fedimint_core::fmt_utils::AbbreviateDebug;
 use fedimint_core::invite_code::InviteCode;
 use fedimint_core::module::audit::AuditSummary;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
@@ -30,12 +29,14 @@ use fedimint_core::net::api_announcement::SignedApiAnnouncement;
 use fedimint_core::session_outcome::{SessionOutcome, SessionStatus};
 use fedimint_core::task::{MaybeSend, MaybeSync};
 use fedimint_core::transaction::{Transaction, TransactionSubmissionOutcome};
+use fedimint_core::util::backoff_util::api_networking_backoff;
 use fedimint_core::util::SafeUrl;
 use fedimint_core::{
-    apply, async_trait_maybe_send, dyn_newtype_define, runtime, NumPeersExt, OutPoint, PeerId,
+    apply, async_trait_maybe_send, dyn_newtype_define, util, NumPeersExt, OutPoint, PeerId,
     TransactionId,
 };
 use fedimint_logging::LOG_CLIENT_NET_API;
+use futures::future::pending;
 use futures::stream::FuturesUnordered;
 use futures::{Future, StreamExt};
 use itertools::Itertools;
@@ -55,7 +56,7 @@ use serde_json::Value;
 use tokio_rustls::rustls::RootCertStore;
 #[cfg(all(feature = "tor", not(target_family = "wasm")))]
 use tokio_rustls::{rustls::ClientConfig as TlsClientConfig, TlsConnector};
-use tracing::{debug, instrument, trace};
+use tracing::{debug, instrument, trace, warn};
 
 use crate::query::{QueryStep, QueryStrategy, ThresholdConsensus};
 mod error;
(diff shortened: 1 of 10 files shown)


fedimint-api-client/src/query.rs
```diff
@@ -2,20 +2,17 @@ use std::collections::{BTreeMap, BTreeSet};
...
-
 /// Fedimint query strategy
 ///
 /// Due to federated security model each Fedimint client API call to the
 /// Federation might require a different way to process one or more required
 /// responses from the Federation members. This trait abstracts away the details
 /// of each specific strategy for the generic client Api code.
 pub trait QueryStrategy<IR, OR = IR> {
-    fn process(&mut self, peer_id: PeerId, response: api::PeerResult<IR>) -> QueryStep<OR>;
+    fn process(&mut self, peer_id: PeerId, response: IR) -> QueryStep<OR>;
```
dpc, line 15: OK, so if I'm reading it right, we don't let the query strategies to see failures anymore, and instead a `retry` will just retry calls to any peer until they work.OK, so if I'm reading it right, we don't let the query strategies to see failures anymore, and instead a `retry` will just retry calls to any peer until they work.

fedimint-api-client/src/query.rs
```diff
@@ -177,7 +108,6 @@ pub struct ThresholdConsensus<R> {
 impl<R> ThresholdConsensus<R> {
     pub fn new(num_peers: NumPeers) -> Self {
         Self {
-            error_strategy: ErrorStrategy::new(num_peers.one_honest()),
```
dpc, line 180: This renders `error_strategy` not needed, but if no existing (and forseeable) query strategy was making use of it, that's OK.

fedimint-client/src/transaction/sm.rs
```diff
@@ -146,37 +145,35 @@ impl State for TxSubmissionStatesSM {
...
     async fn trigger_created_accepted(txid: TransactionId, context: DynGlobalClientContext) {
-        loop {
-            match context.api().await_transaction(txid).await {
-                Ok(..) => return,
-                Err(error) => error.report_if_important(),
-            }
-
-            sleep(RETRY_INTERVAL).await;
-        }
+        context.api().await_transaction(txid).await;
```
dpc, line 176: Huh. This is kind of good. We'll never need a higher level retry, and always just use lower level ones.

One thing that I'm worried about is that now it's a bit unclear which methods need a higher level loop and which don't. But I guess `await_...` might be a good enough clue?

fedimint-client/src/transaction/sm.rs
```diff
@@ -146,37 +145,35 @@ impl State for TxSubmissionStatesSM {
...
-            sleep(RETRY_INTERVAL).await;
+            sleep(backoff.next().expect("Number of retries is has no limit")).await;
         }
     }
 
     async fn trigger_created_accepted(txid: TransactionId, context: DynGlobalClientContext) {
-        loop {
-            match context.api().await_transaction(txid).await {
-                Ok(..) => return,
-                Err(error) => error.report_if_important(),
```
dpc, line 175: This reporting only when the error was not-expected and important was very helpful with debugging while avoiding spamming the logs on ever issue. We should add it to the new per-peer retry loop you've introduced, it would work quite well there I think.

Maybe we need to assign you more debugging weird crap issues and bugs so you get more sensitive about these. :P :D

fedimint-api-client/src/api/mod.rs
```diff
@@ -194,92 +195,155 @@ pub trait FederationApiExt: IRawFederationApi {
...
+                        }));
                     }
                 }
-                None => {
-                    panic!("Query strategy ran out of peers to query without returning a result");
+                QueryStep::Success(response) => return response,
+                QueryStep::Failure(e) => {
+                    warn!("Query strategy returned non-retryable failure for peer {peer}: {e}");
+
+                    futures.push(Box::pin(pending()));
```
dpc, line 344: This looks wrong, no? We are queering all peers on one future here, so we just want to do nothing and let the loop continue.

dpc:
Generally :+1: , though if you are including the changes from https://github.com/fedimint/fedimint/pull/6536, then lets already use `retry` there, and I think that channel to delay await until first submit suceeded is a worthwhile improvement (but I'm happy to argue about it :D ).

fedimint-client/src/transaction/sm.rs
joschisan: You see if the return a federation result or not.

fedimint-client/src/transaction/sm.rs
```diff
@@ -146,37 +145,35 @@ impl State for TxSubmissionStatesSM {
...
 
 impl TxSubmissionStates {
-    async fn trigger_created_rejected(tx: Transaction, context: DynGlobalClientContext) -> String {
+    async fn trigger_created_rejected(
+        transaction: Transaction,
+        context: DynGlobalClientContext,
+    ) -> String {
+        let mut backoff = custom_backoff(Duration::from_secs(2), Duration::from_secs(600), None);
+
         loop {
```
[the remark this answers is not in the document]
joschisan: I changed it to use retry already.

fedimint-api-client/src/api/mod.rs
joschisan: This is in fact correct so the futures unordered will never run out of futures.

fedimint-client/src/transaction/sm.rs
joschisan: Added the report_if_important.

fedimint-api-client/src/api/mod.rs
dpc: Ohhh... . I see it now. Why do we need to do that? We can clearly detect it (running out of futures) in:

```
            let (peer, response) = futures
                .next()
                .await
                .expect("Query strategy ran out of peers to query without returning a result");
```

and running out futures (no decision despite responses from all peers) used to be a programming error.

fedimint-api-client/src/api/mod.rs
```diff
@@ -194,92 +195,159 @@ pub trait FederationApiExt: IRawFederationApi {
...
-                PeerResponse {
-                    peer: *peer_id,
-                    result: request.await,
+                    (*peer, result)
                 }
             }));
         }
 
-        let mut peer_delay_ms = BTreeMap::new();
+        let mut peers = BTreeMap::new();
```
dpc, line 212: `peer_errors`?

fedimint-client/src/transaction/sm.rs
dpc: Huh... Yeah... I guess it's hard to make a retry loop around something that doesn't fail.

dpc (dismissed):

dpc (approved):

PR #6542, opened 2024-12-11
refactor(client): finalize_and_submit to return OutPointRange

dpc:
Notable changes:

* `await_primary_module_output` and similar do not return `Amount` as we no longer track per-output amount, and it didn't seem very used anywhere anyway, and it returned wrong value now as reported by @shaurya947 
* `OutPointRange` and `IdxRange` is a half-open range now, to allow expressing "txid without any change outputs"

If we're fine with this direction, I will attempt to backport, if it fails, we can backport just the `await_primary_module_output` side, I guess. I had to try the whole thing to ensure everything architecturally still works and makes sense.

### DIFF
--- a/fedimint-client/src/transaction/builder.rs
+++ b/fedimint-client/src/transaction/builder.rs
@@ -464,7 +464,7 @@ impl TransactionBuilder {
                     if let Some(input_idxs) = input_idxs.as_ref() {
                         (sm.state_machines)(OutPointRange::new(
                             txid,
-                            IdxRange::from(input_idxs.clone()),
+                            IdxRange::from_inclusive(input_idxs.clone()).expect("can't overflow"),
                         ))
                     } else {
                         vec![]
@@ -483,7 +483,8 @@ impl TransactionBuilder {
                         if let Some(output_idxs) = output_idxs.as_ref() {
                             (sm.state_machines)(OutPointRange::new(
                                 txid,
-                                IdxRange::from(output_idxs.clone()),
+                                IdxRange::from_inclusive(output_idxs.clone())
+                                    .expect("can't possibly overflow"),
                             ))
                         } else {
                             vec![]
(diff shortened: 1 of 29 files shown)


dpc:
First commit as a spearate PR: https://github.com/fedimint/fedimint/pull/6547

modules/fedimint-mint-client/src/lib.rs
```diff
@@ -1488,24 +1488,17 @@ impl MintClientModule {
...
-                    txid: Some(txid),
-                    out_point_indices: out_points
-                        .iter()
-                        .map(|out_point| out_point.out_idx)
-                        .collect(),
-                },
-                amount,
-                extra_meta: extra_meta.clone(),
-            }
+        let operation_meta_gen = |change_range: OutPointRange| MintOperationMeta {
```
elsirion, line 1491: I wonder if this is a bit confusing now: one can still get the txid from the change range, but it's not as obvious anymore

fedimint-client/src/transaction/builder.rs
```diff
@@ -467,6 +485,8 @@ impl TransactionBuilder {
                             IdxRange::from_inclusive(input_idxs.clone()).expect("can't overflow"),
                         ))
                     } else {
+                        // TODO: In the future `InputBundle` and `OutputBundle` should make empty
+                        // bundles impossible, so this should not be possible
```
elsirion, line 489: Empty bundles being impossible is ok, but inputless txns are an explicit feature that we use in e.g. the LN module.

elsirion (approved):
LGTM

elsirion:
Needs rebase :(

fedimint-client/src/transaction/builder.rs
dpc: Yeap. Inputless tx is just a tx without any bundles.

dpc:
@elsirion rebased

elsirion (approved):

PR #6550, opened 2024-12-12
chore: allow en/decoding only for checked networks

tvolk131:
Requires #6517

Replace implementations of `Encodable` and `Decodable` for _unchecked_ addresses with implementations for _checked_ addresses. This should be safe since all places where we previously encoded unchecked addresses, we simply casted a checked address to an unchecked address. However, since checked addresses don't implement `serde::Deserialize`, we must drop support for it in those structs. For that reason, this change probably isn't a good idea, I just wanted to see if it would work.


tvolk131:
Closing due to the `serde::Deserialize` issue mentioned in the description

PR #6553, opened 2024-12-12
fix(server): small submit_transaction logging improvements

dpc:

### DIFF
--- a/fedimint-server/src/consensus/api.rs
+++ b/fedimint-server/src/consensus/api.rs
@@ -50,7 +50,7 @@ use fedimint_core::{secp256k1, OutPoint, PeerId, TransactionId};
 use fedimint_logging::LOG_NET_API;
 use futures::StreamExt;
 use tokio::sync::{watch, RwLock};
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use crate::config::io::{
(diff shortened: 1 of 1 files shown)


fedimint-server/src/consensus/api.rs
```diff
@@ -128,12 +128,18 @@ impl ConsensusApi {
...
+            debug!(target: LOG_NET_API, %txid, %e, "Transaction rejected");
+        })?;
+
+        let _ = self
+            .submission_sender
+            .send(ConsensusItem::Transaction(transaction.clone()))
             .await
-            .ok();
+            .inspect_err(|e| {
+                warn!(target: LOG_NET_API, %txid, %e, "Unable to submit the tx into consensus");
```
dpc, line 141: Knowing that they are not in the logs helps with paranoia.

fedimint-server/src/consensus/api.rs
```diff
@@ -128,12 +128,18 @@ impl ConsensusApi {
             &transaction,
             self.cfg.consensus.version,
         )
-        .await?;
-
-        self.submission_sender
-            .send(ConsensusItem::Transaction(transaction))
+        .await
+        .inspect_err(|e| {
+            debug!(target: LOG_NET_API, %txid, %e, "Transaction rejected");
```
dpc, line 133: Helps with debugging and shouldn't be happening too often.

elsirion (approved):

PR #6554, opened 2024-12-12
Fibonacci retries for everyone

joschisan:


modules/fedimint-lnv2-client/src/send_sm.rs
```diff
@@ -172,36 +170,17 @@ impl SendStateMachine {
...
-                        ?federation_id,
-                        ?gateway_api,
-                        "Error while trying to send payment via gateway"
-                    );
-                }
-            }
+            ensure!(
+                contract.verify_gateway_response(&payment_result),
+                "Invalid gateway response: {payment_result:?}"
+            );
```
elsirion, line 178: Please preserve some of the logging, e.g. through an if+error+bail

modules/fedimint-ln-client/src/api.rs
```diff
@@ -238,8 +237,6 @@ where
         for peer in self.all_peers() {
             if let Ok(response) = self
                 .request_single_peer_federation::<bool>(
-                    // Only wait a second since removing a gateway is "best effort"
-                    Some(Duration::from_secs(1)),
```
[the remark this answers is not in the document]
joschisan: timeout is back.

modules/fedimint-wallet-client/src/withdraw.rs
```diff
@@ -68,44 +66,33 @@ impl State for WithdrawStateMachine {
...
+            outcome
+                .ensure_v0_ref()
+                .map(|outcome| outcome.0)
+                .map_err(|e| e.to_string())
+        }) {
+        Ok(txid) => Ok(txid),
+        Err(e) => {
+            warn!("Failed to process wallet output outcome: {e}");
+
+            pending().await
```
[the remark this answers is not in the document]
joschisan: Agree, this is just current behavior already.

modules/fedimint-lnv2-client/src/send_sm.rs
joschisan: I assumed the logging is / should be handled by util:retry?

gateway/ln-gateway/src/state_machine/pay.rs
```diff
@@ -328,13 +328,8 @@ impl GatewayPayInvoice {
...
             .module_api()
-            .wait_contract(contract_id)
-            .await
-            .map_err(|_| OutgoingPaymentError {
-                contract_id,
-                contract: None,
-                error_type: OutgoingPaymentErrorType::OutgoingContractDoesNotExist { contract_id },
-            })?;
+            .await_contract(contract_id)
+            .await;
```
joschisan, line 332: It appears that the previous behavior here was just wrong as an rpc error was interpreted at if the contract did not exist. Seems to me the gateway would then fail to cancel an existing contract.

modules/fedimint-ln-client/src/api.rs
```diff
@@ -272,42 +258,6 @@ where
...
-                amount: account.amount,
-                contract: c.contract,
-            }),
-            FundedContract::Outgoing(_) => Err(FederationError::general(
-                AWAIT_ACCOUNT_ENDPOINT,
-                id,
-                anyhow::anyhow!("WrongAccountType"),
-            )),
-        }
-    }
```
joschisan, line 284: This was just unused.

modules/fedimint-lnv2-client/src/send_sm.rs
elsirion: Good point, what I'm concerned about is that all the extra structured data (contract, invoice, etc.) won't be available anymore in logs, just some randomly formatted error string. But you are right that we should leave the logging to `retry` and there actually is an elegant way to inject all that data into the logs created by the `retry` closure: [spans](https://docs.rs/tracing/latest/tracing/span/index.html). 

Here we could do:

```rust
        util::retry("gateway-send-payment", api_networking_backoff(), || async {
            …
        })
        .instrument(info_span!(
            ?contract,
            ?invoice,
            ?federation_id,
            ?gateway_api,
        ))
        .await
        .expect("Number of retries has no limit")
```

Or alternatively add an instrument attribute to the whole function:

```rust
    #[instrument(
        skip_all,
        fields(
            ?contract,
            ?invoice,
            ?federation_id,
            ?gateway_api,
        ),
    )]
    async fn gateway_send_payment(
        gateway_api: SafeUrl,
        federation_id: FederationId,
        contract: OutgoingContract,
        invoice: LightningInvoice,
        refund_keypair: Keypair,
        context: LightningClientContext,
    ) -> Result<[u8; 32], Signature> {
```

<details><summary>More details about span log level interaction I was wondering about myself</summary>
<p>

Let me explain how span log levels interact with log statement levels in Rust's tracing framework.

The key thing to understand is that spans and events (log statements) have independent filtering, but spans can affect the visibility of events within them. Here's how it works:

```rust
use tracing::{info, debug, info_span, Level};

fn process_data() {
    // Span with INFO level
    let span = info_span!("process_data").entered();
    
    // These will be visible since their level >= span's level
    info!("starting processing");  // visible
    debug!("detailed step");      // visible
    
    // In a debug span
    let debug_span = debug_span!("detailed_processing").entered();
    info!("still visible");       // visible
    debug!("debug info");         // visible only if debug enabled
}
```

Key the interaction:

1. A span's level acts as a gate for whether the span itself is recorded/collected
2. Log statements (events) inside a span are independently filtered based on their own levels
3. If a span is disabled due to its level being too low for the current configuration, events within it may still be recorded
4. However, those events won't have the context (fields, metadata) from the filtered-out span

## Practical Example

Here's how different configurations affect visibility:

```rust
use tracing::{info, debug, info_span, Level};
use tracing_subscriber::filter::LevelFilter;

// Configure subscriber with INFO level
tracing_subscriber::fmt()
    .with_max_level(LevelFilter::INFO)
    .init();

fn example() {
    // This span will be recorded (INFO >= INFO)
    let span = info_span!("operation").entered();
    info!("visible message");     // visible
    debug!("invisible message");  // filtered out
    
    // This span won't be recorded (DEBUG < INFO)
    let detail_span = debug_span!("details").entered();
    info!("still visible");      // visible, but without debug_span context
    debug!("invisible");         // filtered out
}
```

## Best Practices

1. Set span levels to reflect their operational significance
2. Use consistent level hierarchies (don't put critical events in debug spans)
3. Remember that span levels affect metadata collection, not just visibility
4. Consider using different filters for spans vs events if needed through the `EnvFilter` or custom filters

This layered filtering system allows for flexible and granular control over your application's observability while maintaining performance.

</p>
</details>

modules/fedimint-ln-client/src/incoming.rs
```diff
@@ -177,41 +172,12 @@ impl FundingOfferState {
...
-            }
-            // give some time for other things to run
-            fedimint_core::runtime::sleep(Duration::from_secs(sleep)).await;
-        }
-
-        unreachable!("there is too many u64s to ever get here")
+        global_context
+            .await_tx_accepted(txid)
+            .await
+            .map_err(|error_message| IncomingSmError::FailedToFundContract { error_message })
```
elsirion, line 180: Seems we used `await_output_outcome` while it could have been the much more lightweight `await_tx_accepted` all along, worth considering in the future. cc @m1sterc001guy @tvolk131

modules/fedimint-wallet-client/src/withdraw.rs
```diff
@@ -68,44 +66,33 @@ impl State for WithdrawStateMachine {
...
 
-    loop {
-        match global_context
-            .api()
-            .await_output_outcome::<WalletOutputOutcome>(
-                created.fm_outpoint,
-                Duration::MAX,
-                &context.wallet_decoder,
-            )
-            .await
```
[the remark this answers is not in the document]
elsirion: Call: usage was inconsistent and often unnecessary, replaced with manual calls in the few places that require it.

elsirion (approved):

gateway/ln-gateway/src/state_machine/pay.rs
m1sterc001guy: Yes this was likely an oversight, there were no retries here.

PR #6556, opened 2024-12-12
feat: expose in-flight balance

elsirion:
Some operations may fail and can cause a refund (OOB sends) or succeed only after some time (Deposits). This PR exposes the total in-flight balance via a watch channel.

### DIFF
--- a/fedimint-client/src/lib.rs
+++ b/fedimint-client/src/lib.rs
@@ -3223,6 +3223,54 @@ impl ClientBuilder {
     }
 }
 
+/// Represents money we may be receiving, but haven't yet
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize)]
+pub struct InFlightAmounts {
+    /// Receives that are not yet reflected in our main balance
+    pub incoming: Amount,
+    /// Sends that may be refunded
+    pub outgoing: Amount,
+}
+
+impl Default for InFlightAmounts {
+    fn default() -> Self {
+        InFlightAmounts {
+            incoming: Amount::ZERO,
+            outgoing: Amount::ZERO,
+        }
+    }
+}
+
+impl InFlightAmounts {
+    pub fn incoming(incoming: Amount) -> Self {
+        InFlightAmounts {
+            incoming,
+            outgoing: Amount::ZERO,
+        }
+    }
+
+    pub fn outgoing(outgoing: Amount) -> Self {
+        InFlightAmounts {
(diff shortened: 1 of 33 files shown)


fedimint-client/src/lib.rs
```diff
@@ -3132,6 +3132,40 @@ impl ClientBuilder {
...
 }
 
+/// Represents money we may be receiving, but haven't yet
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize)]
+pub struct InFlightAmounts {
+    /// Receives that are not yet reflected in our main balance
+    pub incoming: Amount,
+    /// Sends that may be refunded
+    pub outgoing: Amount,
+}
```
elsirion, line 3142: Should maybe be an enum?

fedimint-client/src/sm/executor.rs
```diff
@@ -411,6 +419,12 @@ impl Executor {
     pub fn notifier(&self) -> &Notifier {
         &self.inner.notifier
     }
+
+    /// Estimated in-flight value represented by active state machines. Not all
+    /// modules fully support this feature.
```
elsirion, line 424: Add: "only use as a heuristic for users, not for programmatic decision-making"

elsirion (commented):
Feature idea: add in-flight balance querying per operation

elsirion:
Ahhhh, I think I discovered a missing piece: the on-chain deposits aren't SMs anymore, so need to be handled separately.

fedimint-client/src/lib.rs
```diff
@@ -3206,6 +3206,54 @@ impl ClientBuilder {
...
+impl AddAssign for InFlightAmounts {
+    fn add_assign(&mut self, rhs: Self) {
+        self.incoming += rhs.incoming;
+        self.outgoing += rhs.outgoing;
+    }
+}
+
+impl SubAssign for InFlightAmounts {
+    fn sub_assign(&mut self, rhs: Self) {
+        self.incoming -= rhs.incoming;
```
maan2003, line 3252: underflow waiting to happening

maan2003: I would rather store BTreeMap<OperationId, InFlightAmounts>, feels less error prone

elsirion:
@maan2003 do you think this is worth pursuing at all? I feel like closing for now, turns out to be quite ugly to implement.

maan2003:
no IMO, it is lot of more effort than any ux gains

PR #6561, opened 2024-12-13
feat: ldk gateway supports bitcoind as a backend

tvolk131:
Add the option to use bitcoind as a backend for the ldk gateway in addition to esplora.

Also use this new backend in devimint, since dev fed startup is usually blocked by channel opens, which are blocked by ldk gw startup, which is blocked by esplora startup which takes 10-15 seconds on my machine. By using bitcoind as the backend, channel opens no longer have to wait for esplora. On my machine, this changed mprocs startup time from ~25-30 seconds to ~19-20 seconds, making it faster and more consistent


dpc:
On my workstation:

```
2024-12-16T17:05:36.405940Z  INFO fm::devimint: Devfed ready elapsed_ms=12069
```

image

gateway/ln-gateway/src/envs.rs
```diff
@@ -37,10 +37,26 @@ pub const FM_LND_TLS_CERT_ENV: &str = "FM_LND_TLS_CERT";
...
 
-/// Environment variable that specifies the URL of an Esplora server. Necessary
-/// for LDK configuration.
+/// Environment variable that specifies the URL of an Esplora server.
+/// Necessary for LDK configuration if using esplora as the backend.
 pub const FM_LDK_ESPLORA_SERVER_URL: &str = "FM_LDK_ESPLORA_SERVER_URL";
 
+/// Environment variable that specifies the host of the bitcoind node.
+/// Necessary for LDK configuration if using bitcoind as the backend.
+pub const FM_LDK_BITCOIND_HOST: &str = "FM_LDK_BITCOIND_HOST";
```
dpc, line 46: Do we want to use a one URL env var instead of 4 separate part env? Just like we do in Fedimint? (`FM_BITCOIN_RPC_URL`)

We could possibly make it fallback to the value of `FM_BITCOIN_RPC_URL` then as well, as these would usually be the same.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -233,25 +286,14 @@ impl ILnRpcClient for GatewayLdkClient {
...
-        };
-
-        let esplora_chain_tip_block_height = esplora_chain_tip_block_summary.time.height;
-        let ldk_block_height: u32 = node_status.current_best_block.height;
+        let esplora_chain_tip_block_height =
+            self.bitcoind_rpc.get_block_count().await.map_err(|e| {
+                LightningRpcError::FailedToGetNodeInfo {
+                    failure_reason: format!("Failed to get block count from chain source: {e}"),
+                }
+            })? as u32
```
dpc, line 294: As a routine we should avoid `as` casts, and use `TryFrom`

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -94,9 +131,29 @@ impl GatewayLdkClient {
             node_alias,
             ..Default::default()
         });
-        node_builder
-            .set_entropy_bip39_mnemonic(mnemonic, None)
-            .set_chain_source_esplora(esplora_server_url.to_string(), None);
+
+        node_builder.set_entropy_bip39_mnemonic(mnemonic, None);
+
+        let bitcoind_rpc = create_bitcoind(&chain_source_config.bitcoin_rpc_config())?;
```
m1sterc001guy, line 137: We probably shouldn't even call `create_bitcoind` if the chain source is esplora right?

m1sterc001guy: Oh I see it is general between the two. Nevermind :)

tvolk131: Yeah, we should probably rename it. `create_bitcoind()` creates a generic bitcoin backend, not necessarily a bitcoind backend.

gateway/ln-gateway/src/envs.rs
m1sterc001guy: I would be in favor of one URL env var. It would make the code below in `LightningBuilder::build` a bit cleaner

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -34,12 +36,47 @@ use crate::lightning::{
 };
 use crate::rpc::{CloseChannelsWithPeerPayload, OpenChannelPayload, SendOnchainPayload};
 
+pub enum GatewayLdkChainSourceConfig {
+    Bitcoind {
+        rpc_host: String,
```
dpc, line 41: Looks inconsistent with the variant couple of lines below. :)

gateway/ln-gateway/src/lightning/mod.rs
```diff
@@ -365,19 +383,62 @@ impl LightningBuilder for GatewayLightningBuilder {
...
+                Box::new(
+                    ldk::GatewayLdkClient::new(
+                        &self.ldk_data_dir,
+                        chain_source_config,
+                        network,
+                        lightning_port,
+                        self.mnemonic.clone(),
+                        runtime,
+                    )
+                    .unwrap(),
```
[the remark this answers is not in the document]
tvolk131: Done.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -34,12 +36,47 @@ use crate::lightning::{
...
+    Esplora {
+        server_url: String,
+    },
+}
+
+impl GatewayLdkChainSourceConfig {
+    fn bitcoin_rpc_config(&self) -> BitcoinRpcConfig {
+        match self {
+            Self::Bitcoind {
+                rpc_host,
```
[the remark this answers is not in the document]
tvolk131: I'm not 100% sure what you're suggesting, but I changed `GatewayLdkChainSourceConfig` so it holds `SafeUrl`s instead of strings.

gateway/ln-gateway/src/lightning/ldk.rs
```diff
@@ -233,25 +286,14 @@ impl ILnRpcClient for GatewayLdkClient {
...
-            return Err(LightningRpcError::FailedToGetNodeInfo {
-                failure_reason:
-                    "Failed to get chain tip block summary (empty block list was returned)"
-                        .to_string(),
-            });
-        };
-
-        let esplora_chain_tip_block_height = esplora_chain_tip_block_summary.time.height;
-        let ldk_block_height: u32 = node_status.current_best_block.height;
+        let esplora_chain_tip_block_height =
```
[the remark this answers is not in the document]
tvolk131: Good catch! Done.

gateway/ln-gateway/src/lightning/ldk.rs

gateway/ln-gateway/src/envs.rs
tvolk131: Done.

dpc (approved):

PR #6569, opened 2024-12-16
refactor(client): clarify handling of empty bundles

dpc:
Trying to forbid "empty bundle" (no inputs/outputs) turned out to generate too much `Option`-noise downstream, so let's just embrace it.

### DIFF
--- a/fedimint-client/src/transaction/builder.rs
+++ b/fedimint-client/src/transaction/builder.rs
@@ -481,8 +481,8 @@ impl TransactionBuilder {
         let input_states = self
             .inputs
             .into_iter()
-            .filter(|bundle| !bundle.is_empty())
             .enumerate()
+            .filter(|(_, bundle)| !bundle.is_empty())
             .flat_map(|(bundle_idx, bundle)| {
                 let input_idxs = find_range_of_matching_items(&input_idx_to_bundle_idx, bundle_idx)
                     .expect("Non empty bundles must always have a match");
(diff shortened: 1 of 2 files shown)


fedimint-client/src/transaction/builder.rs
```diff
@@ -475,44 +481,36 @@ impl TransactionBuilder {
         let input_states = self
             .inputs
             .into_iter()
+            .filter(|bundle| !bundle.is_empty())
```
dpc, line 484: Core of the change: filter out "empty bundles" explicitly upfront, so we can simplify rest of the code.

fedimint-client/src/transaction/builder.rs
```diff
@@ -475,44 +481,36 @@ impl TransactionBuilder {
...
-                });
+        let output_states = self
+            .outputs
+            .into_iter()
+            .filter(|bundle| !bundle.is_empty())
+            .enumerate()
+            .flat_map(|(bundle_idx, bundle)| {
+                let output_idxs =
+                    find_range_of_matching_items(&output_idx_to_bundle_idx, bundle_idx)
+                        .expect("Non empty bundles must always have a match");
```
elsirion, line 505: Enumerate after filter feels fragile, wherever that idx is used has to have the same filtering in place.

dpc: @elsirion It looks fragile because it's actually a bug. The bundles are numbered without any filtering. :facepalm: It would lead to a panic if we actually used some empty bundles. Good catch.

dpc: Added a test and fix.

elsirion (approved):

PR #6570, opened 2024-12-16
refactor: cleanup FederationInfo

m1sterc001guy:
`FederationConfig` and `FederationInfo` have overlapping fields and its confusing when each one is used. This PR refactors it so that `FederationConfig` is a field inside of `FederationInfo`. `FederationInfo` is what is displayed to the gateway operator during the `info` command. `FederationConfig` is what is stored in the database.

`FederationConfig` can be printed using `gateway-cli cfg display`.

### DIFF
--- a/gateway/ln-gateway/src/federation_manager.rs
+++ b/gateway/ln-gateway/src/federation_manager.rs
@@ -174,16 +174,6 @@ impl FederationManager {
         }
     }
 
-    fn get_index_for_federation(&self, federation_id: FederationId) -> Option<u64> {
-        self.index_to_federation.iter().find_map(|(index, fid)| {
-            if *fid == federation_id {
-                Some(*index)
-            } else {
-                None
-            }
-        })
-    }
-
     pub fn get_client_for_federation_id_prefix(
         &self,
         federation_id_prefix: FederationIdPrefix,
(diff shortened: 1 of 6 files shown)


gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -231,9 +215,7 @@ impl FederationManager {
                     federation_id,
                     federation_name: self.federation_name(client).await,
                     balance_msat,
-                    federation_index,
-                    lightning_fee: fed_config.lightning_fee,
-                    transaction_fee: fed_config.transaction_fee,
+                    config: fed_config,
```
tvolk131, line 218: Nit: rename `fed_config` to `config` so this line can just be `config,`

gateway/ln-gateway/src/federation_manager.rs
```diff
@@ -251,8 +233,6 @@ impl FederationManager {
     ) -> Vec<FederationInfo> {
         let mut federation_infos = Vec::new();
         for (federation_id, client) in &self.clients {
-            let federation_index = self.get_index_for_federation(*federation_id).expect("`FederationManager.index_to_federation` is out of sync with `FederationManager.clients`! This is a bug.");
-
             let balance_msat = client.borrow().with(|client| client.get_balance()).await;
 
             let fed_config = dbtx.load_federation_config(*federation_id).await;
```
tvolk131, line 238: Same nit as above: Rename `fed_config` to `config`

tvolk131 (changes requested):
Just a couple nits, otherwise LGTM!

tvolk131 (approved):

PR #6571, opened 2024-12-16
chore: gateway returns federation info when not running

tvolk131:
Previously when the gateway was not in the running state, it didn't return information related to what federations it was connected to, or the bitcoin network it was connected to, or the lightning mode it was running on. I don't believe there's any reason why we shouldn't allow for this data to be shown. Doing this also means that `GatewayInfo.federation_fake_scids` and `GatewayInfo.network` are never `None`. However, simply changing them would lead to backwards compatibility issues when using the new CLI with older versions of the gateway.

### DIFF
--- a/gateway/ln-gateway/src/rpc/mod.rs
+++ b/gateway/ln-gateway/src/rpc/mod.rs
@@ -113,12 +113,14 @@ pub struct GatewayInfo {
     pub federations: Vec<FederationInfo>,
     /// Mapping from short channel id to the federation id that it belongs to.
     // TODO: Remove this alias once it no longer breaks backwards compatibility.
+    // TODO(support:v0.5): Remove the `Option` wrapper.
     #[serde(alias = "channels")]
     pub federation_fake_scids: Option<BTreeMap<u64, FederationId>>,
     pub lightning_pub_key: Option<String>,
     pub lightning_alias: Option<String>,
     pub gateway_id: secp256k1::PublicKey,
     pub gateway_state: String,
+    // TODO(support:v0.5): Remove the `Option` wrapper.
     pub network: Option<Network>,
     // TODO: This is here to allow for backwards compatibility with old versions of this struct. We
     // should be able to remove it once 0.4.0 is released.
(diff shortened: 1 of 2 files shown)


gateway/ln-gateway/src/rpc/mod.rs
```diff
@@ -113,12 +113,14 @@ pub struct GatewayInfo {
...
     /// Mapping from short channel id to the federation id that it belongs to.
     // TODO: Remove this alias once it no longer breaks backwards compatibility.
+    // TODO(support:v0.5): Remove the `Option` wrapper.
     #[serde(alias = "channels")]
     pub federation_fake_scids: Option<BTreeMap<u64, FederationId>>,
     pub lightning_pub_key: Option<String>,
     pub lightning_alias: Option<String>,
     pub gateway_id: secp256k1::PublicKey,
     pub gateway_state: String,
+    // TODO(support:v0.5): Remove the `Option` wrapper.
```
m1sterc001guy, line 123: This has already been fixed in https://github.com/fedimint/fedimint/pull/6516

m1sterc001guy:
> However, simply changing them would lead to backwards compatibility issues when using the new CLI with older versions of the gateway.

We do not maintain backwards compatibility for the gateway CLI. Whatever incompatibility issues you're running into, they can be fixed by conditional checks in devimint

dpc:
*Comment by Tau/GPT-5*

Thanks for opening this. I re-triaged this against current `master`.

The original diff is now stale/conflicting because the gateway crates and `GatewayInfo` shape have changed, and part of the motivation has since been handled differently: `lightning_mode` is now always present, and node connectivity details are represented by `LightningInfo` (`Connected`/`NotConnected`) instead of the old top-level optional fields.

The remaining core behavior still appears relevant: current `Gateway::handle_get_info` in `gateway/fedimint-gateway-server/src/lib.rs` still returns `federations: vec![]` and `federation_fake_scids: None` before it checks loaded federation info whenever the gateway is not in `Running` state. So this is not mergeable as-is, but it should be straightforward to revive by porting the small logic change to the current file: compute `federations`/`channels` before the `GatewayState::Running` early return, return them with `LightningInfo::NotConnected`, and avoid reintroducing the obsolete `network`/`lightning_mode` compatibility changes. A small regression test for `/info` while the LN node/gateway is not running would make the intended behavior clear.

Recommendation: leave open if someone wants to finish the port; otherwise a fresh small PR with the above change may be simpler than rebasing this draft.

PR #6581, opened 2024-12-18
feat: Gateway 24 hour payment summary

m1sterc001guy:
Adds a 24 hour payment summary command to the gateway. Useful to get an "at a glance" summary of what happened over the last day.

image

### DIFF
--- a/fedimint-eventlog/src/lib.rs
+++ b/fedimint-eventlog/src/lib.rs
@@ -92,20 +92,18 @@ impl UnordedEventLogId {
 pub struct EventLogId(u64);
 
 impl EventLogId {
+    pub const LOG_START: EventLogId = EventLogId(0);
+
     fn next(self) -> EventLogId {
         Self(self.0 + 1)
     }
 
-    fn saturating_add(self, rhs: u64) -> EventLogId {
+    pub fn saturating_add(self, rhs: u64) -> EventLogId {
         Self(self.0.saturating_add(rhs))
     }
 
     pub fn saturating_sub(self, rhs: u64) -> EventLogId {
         Self(self.0.saturating_sub(rhs))
-    }
-
-    pub fn new(log: u64) -> Self {
-        Self(log)
     }
 }
 
@@ -169,6 +167,16 @@ pub struct EventLogEntry {
     /// Event-kind specific payload, typically encoded as a json string for
     /// flexibility.
     pub payload: Vec<u8>,
+}
+
+/// Struct used for processing log entries after they have been persisted.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PersistedLogEntry {
+    pub event_id: EventLogId,
+    pub event_kind: EventKind,
+    pub module: Option<(ModuleKind, u16)>,
+    pub timestamp: u64,
+    pub value: serde_json::Value,
 }
 
 impl_db_record!(
@@ -247,13 +255,7 @@ pub trait DBTransactionEventLogExt {
         &mut self,
         pos: Option<EventLogId>,
         limit: u64,
-    ) -> Vec<(
-        EventLogId,
-        EventKind,
-        Option<(ModuleKind, ModuleInstanceId)>,
-        u64,
-        serde_json::Value,
-    )>;
+    ) -> Vec<PersistedLogEntry>;
 }
 
 #[apply(async_trait_maybe_send!)]
@@ -315,24 +317,16 @@ where
         &mut self,
         pos: Option<EventLogId>,
         limit: u64,
-    ) -> Vec<(
-        EventLogId,
-        EventKind,
-        Option<(ModuleKind, ModuleInstanceId)>,
-        u64,
-        serde_json::Value,
-    )> {
+    ) -> Vec<PersistedLogEntry> {
         let pos = pos.unwrap_or_default();
         self.find_by_range(pos..pos.saturating_add(limit))
             .await
-            .map(|(k, v)| {
-                (
-                    k,
-                    v.kind,
-                    v.module,
-                    v.ts_usecs,
-                    serde_json::from_slice(&v.payload).unwrap_or_default(),
-                )
+            .map(|(k, v)| PersistedLogEntry {
+                event_id: k,
+                event_kind: v.kind,
+                module: v.module,
+                timestamp: v.ts_usecs,
+                value: serde_json::from_slice(&v.payload).unwrap_or_default(),
             })
             .collect()
             .await
--- a/gateway/ln-gateway/src/events.rs
+++ b/gateway/ln-gateway/src/events.rs
@@ -1,12 +1,23 @@
-use fedimint_eventlog::{Event, EventKind};
+use std::sync::Arc;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use fedimint_client::ClientHandle;
+use fedimint_core::core::ModuleKind;
+use fedimint_core::util::{get_average, get_median};
+use fedimint_core::Amount;
+use fedimint_eventlog::{
+    DBTransactionEventLogExt, Event, EventKind, EventLogId, PersistedLogEntry,
+};
 use fedimint_mint_client::event::{OOBNotesReissued, OOBNotesSpent};
 use fedimint_wallet_client::events::{DepositConfirmed, WithdrawRequest};
+use itertools::Itertools;
 
 use crate::gateway_module_v2::events::{
     CompleteLightningPaymentSucceeded, IncomingPaymentFailed, IncomingPaymentStarted,
     IncomingPaymentSucceeded, OutgoingPaymentFailed, OutgoingPaymentStarted,
     OutgoingPaymentSucceeded,
 };
+use crate::rpc::PaymentStats;
 
 pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
     OutgoingPaymentStarted::KIND,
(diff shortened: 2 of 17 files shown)


gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +31,263 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
     DepositConfirmed::KIND,
 ];
 
-// TODO: Add Gateway specific events
+pub type LogEntry = (
```
dpc, line 34: That's... a weird way to write `struct LogEntry`... :D

dpc:
Seems like it wouldn't be a big lift now to just add `start` and `end` with defaults meaning "last 24 hours".

dpc:
No, just the `payement-summary` cli command could have `--start` and `--end`

m1sterc001guy:
Ah sure, can make that change. I would like to have a maximum range though. At some point if there's a lot of data, it could be a heavy computation.

gateway/ln-gateway/src/events.rs
m1sterc001guy: Fixed. Moved and renamed to `PersistedLogEntry` to not be confused with `EventLogEntry`

m1sterc001guy:
> No, just the payement-summary cli command could have --start and --end

Done

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+/// Binary searches the `PersistedLogEntry` slice for the start timestamp and
+/// end timestamp and returns the index of each. If an exact match is not found,
+/// the index where start or end could be inserted to maintain ordering
+/// is returned.
+fn get_index_bounds(
+    batch: &[PersistedLogEntry],
+    start: SystemTime,
+    end: SystemTime,
+) -> (usize, usize) {
+    let timestamps = batch.iter().map(|e| e.timestamp).collect::<Vec<_>>();
```
dpc, line 101: Words of warning: the exact timestamp is generated by the side that is inserting the event, and since they can be multiple of them, these don't need to be in exact order.

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
     DepositConfirmed::KIND,
 ];
 
-// TODO: Add Gateway specific events
+/// Searches through the event log for all events that occurred within the
+/// specified time bounds. Because it is inefficient to search the log
+/// backwards, instead this function traverses the log forwards, but in batches.
+/// All events are appended to an array until the cutoff event where the
+/// timestamp is greater than the start timestamp or the end of the log is hit.
+pub async fn get_events_for_duration(
```
dpc, line 40: I don't know how much should I stress about corectness here. Are people going to be using it for some accounting stuff? Maybe?

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+    start: SystemTime,
+    end: SystemTime,
+) -> Vec<PersistedLogEntry> {
+    const BATCH_SIZE: u64 = 10_000;
+    let mut end_position = {
+        let mut dbtx = client.db().begin_transaction_nc().await;
+        dbtx.get_next_event_log_id().await
+    };
+
+    let mut start_position = end_position.saturating_sub(BATCH_SIZE);
```
dpc, line 51: The mixing of `_index` and `_position` here is confusing. Should this one be `batch_end` and `batch_start`, and one below `time_bound_start` and `time_bound_end` or something?

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+    // Read the event log until we hit the end.
+    while start_position != log_start {
+        let batch = client.get_event_log(Some(start_position), BATCH_SIZE).await;
+        let (start_index, end_index) = get_index_bounds(&batch, start, end);
+
+        // If the `end_index` is non-zero, that means that all events < `end_index` are
+        // within our time bounds
+        if end_index > 0 {
+            let events = &batch[start_index..end_index];
+            all_events.append(&mut events.to_vec());
```
dpc, line 65: Since the rangers are iterated from end to start, the returned vec will have a weird ordering, which might not be immediately obvious.

IMO, the whole thing might be easier to impl, understand and also faster if it works like this:

* Find the rough start point of a range:
  * Go batch_size by batch size and read **1** event and check if it's older than requested `start`. This avoids reading whole batches just to figure out all entries in it are potentially useless.
* Go through from the "rought start" reading batch by batch, discard events that are too old, and stop when encountering event that is newer than `end`. Collected matched events will also be in the correct order.

If you want to avoid missing any slightly unordered events, add some (5 minutes?) gracefulness when terminating on `start` and `end`, but filter out the events older/newer being precise.

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+        Err(index) | Ok(index) => index,
+    };
+
+    (start_index, end_index)
+}
+
+/// Filters the given `PersistedLogEntry` slice by the `EventKind` and
+/// `ModuleKind`.
+pub(crate) fn filter_events(
+    all_events: &[PersistedLogEntry],
```
dpc, line 126: If you take it as a `IntoIter<Item = PersistedLogEntry>` and return `-> impl Iterator<Item = PersistedLogEntry` you can make the whole thing never even allocate extra.

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+/// This function computes a "nested loop join" by first computing the cross
+/// product of the start event vector and the success/failure event vectors. The
+/// resulting cartesian product is then filtered according to the join predicate
+/// supplied in the parameters.
+///
+/// This function is intended for small data sets. If the data set relations
+/// grow, this function should implement a different join algorithm or be moved
+/// out of the gateway.
+pub(crate) fn join_events<Start, Success, Failure>(
+    start_events: &Vec<&PersistedLogEntry>,
```
dpc, line 154: No function should ever need to take `&Vec<_>`. `&[_]` is always better, as `&vec` works for it, along with `&slice`, while it's not like it the other way around. Kind of like `&str` is always better arg type than `&String`.

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+
+/// This function computes a "nested loop join" by first computing the cross
+/// product of the start event vector and the success/failure event vectors. The
+/// resulting cartesian product is then filtered according to the join predicate
+/// supplied in the parameters.
+///
+/// This function is intended for small data sets. If the data set relations
+/// grow, this function should implement a different join algorithm or be moved
+/// out of the gateway.
+pub(crate) fn join_events<Start, Success, Failure>(
```
dpc, line 153: I was going to suggest making it use `const N : usize` to parametrize to make the success vs failure general, but I realized that this function is just the same thing repeated twice. The caller can just call it twice.


```rust
pub(crate) fn join_events<'a, L, R, Res>(
    events_l: &'a [&PersistedLogEntry],
    events_r: &'a [&PersistedLogEntry],
    predicate: impl Fn(L, R, u64) -> Option<Res> + 'a,
) -> impl Iterator<Item = Res> + 'a
where
    L: Event,
    R: Event,
{
    let cross_product = events_l
        .iter()
        .cartesian_product(events_r)
        .collect::<Vec<_>>();
    cross_product.into_iter().filter_map(move |(l, r)| {
        if let Some(latency) = r.timestamp.checked_sub(l.timestamp) {
            let event_l: L = serde_json::from_value(l.value.clone()).expect("could not parse JSON");
            let event_r: R = serde_json::from_value(r.value.clone()).expect("could not parse JSON");
            predicate(event_l, event_r, latency)
        } else {
            None
        }
    })
}
```

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+        })
+        .collect::<Vec<_>>();
+
+    (success_stats, failure_stats)
+}
+
+/// Helper struct for storing computed data about outgoing and incoming
+/// payments.
+#[derive(Debug, Default)]
+pub struct StructuredPaymentEvents {
```
dpc, line 209: Maybe this struct should have just `latencies`, `fees` and `latencies_failure`, so the caller can have two instances: outgoing and incoming? :shrug:

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+fn average(data: &[u64]) -> u64 {
+    let sum: u64 = data.iter().sum();
+    if data.is_empty() {
+        0
+    } else {
+        sum / data.len() as u64
+    }
+}
+
+/// Computes the median of the given `u64` slice.
```
dpc, line 307: *sorted slice, right?

We have a median in `modules/fedimint-wallet-server/src/lib.rs` maybe let's put in `core` somewhere?

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+fn median(data: &[u64]) -> u64 {
+    if data.is_empty() {
+        return 0;
+    }
+
+    let length = data.len();
+    if length % 2 == 0 {
+        data[length / 2 - 1]
```
dpc, line 315: Median of even is technically supposed to be average of two values around it.

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+        0
+    } else {
+        sum / data.len() as u64
+    }
+}
+
+/// Computes the median of the given `u64` slice.
+fn median(data: &[u64]) -> u64 {
+    if data.is_empty() {
+        return 0;
```
dpc, line 310: Either panic or `-> Option` and default to something on the caller side.

gateway/ln-gateway/src/events.rs
dpc: Why does Rust not have a median from sorted array in a stdlib?

gateway/ln-gateway/src/gateway_module_v2/events.rs
```diff
@@ -136,3 +137,97 @@ impl Event for CompleteLightningPaymentSucceeded {
 
     const KIND: EventKind = EventKind::from_static("complete-lightning-payment-succeeded");
 }
+
+/// Computes the `StructurePaymentEvents` for all LNv2 payments by filtering the
```
dpc, line 141: Nit: Rustdoc comments supposed to be one liner header, and then description after line break

```
/// Computes the `StructurePaymentEvents` for all LNv2 payments
///
/// Filters the ...
```

Yes, it kind of makes a difference in the UX e.g. when using LSPs and looking up what a function does.

fedimint-eventlog/src/lib.rs
```diff
@@ -315,24 +323,16 @@ where
         &mut self,
         pos: Option<EventLogId>,
         limit: u64,
-    ) -> Vec<(
```
dpc, line 318: Making it a proper struct is a big improv. . Should have done it before. :+1:

dpc (commented):
I have some more stuff (sorry!) but otherwise roughly good to go.

gateway/ln-gateway/src/events.rs
m1sterc001guy: I think it will mostly be used for an "at a glance" look at what happened in the gateway over some period of time. If there are obvious correctness issues we should fix them, but IMO this doesn't need to be down to the microsecond accurate

gateway/ln-gateway/src/events.rs
m1sterc001guy: Yes that is better naming

gateway/ln-gateway/src/events.rs
m1sterc001guy: I agree this code is hard to reason about, I had difficulty writing it. The reason I chose to iterate "backwards" though was because this function is most likely to be called on the most recent X timeframe (i.e 24 hours).

If you read starting from the beginning of the log to find the "rough start", in the common case you will have to read almost the entire log just to throw away most of the events. Isn't that somewhat wasteful?

m1sterc001guy: Oh maybe I misunderstood. I suppose you can find the rough start by going backwards, then iterate forwards again

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+
+    /// Computes the payment statistics for the given input data.
+    fn compute_payment_stats(
+        latencies: &[u64],
+        fees: &[Amount],
+        latencies_failure: &[u64],
+    ) -> PaymentStats {
+        PaymentStats {
+            average_latency_micros: average(latencies),
+            median_latency_micros: median(latencies),
```
joschisan, line 289: Can this be a Duration so it's rendered in the appropriate unit? Probably good to add 90th and 95th percentile as well as max latency.

modules/fedimint-lnv2-tests/bin/tests.rs
```diff
@@ -305,6 +305,19 @@ async fn test_payments(dev_fed: &DevJitFed) -> anyhow::Result<()> {
     // relative fee. Gateway receives: 1_000_000 payment.
     test_fees(fed_id, &client, gw_lnd, gw_ldk, 1_000_000 - 1_000 - 1_000).await?;
 
+    info!("Testing payment summary");
+    let lnd_payment_summary = gw_lnd.payment_summary().await?;
+    assert_eq!(lnd_payment_summary.outgoing.total_success, 4);
+    assert_eq!(lnd_payment_summary.outgoing.total_failure, 2);
+    assert_eq!(lnd_payment_summary.incoming.total_success, 3);
+    assert_eq!(lnd_payment_summary.incoming.total_failure, 0);
```
joschisan, line 313: Actually the tests fails incoming payments due to missing liquidity so this should not be 0.

modules/fedimint-lnv2-tests/bin/tests.rs
```diff
@@ -305,6 +305,19 @@ async fn test_payments(dev_fed: &DevJitFed) -> anyhow::Result<()> {
...
+
+    let ldk_payment_summary = gw_ldk.payment_summary().await?;
+    assert_eq!(ldk_payment_summary.outgoing.total_success, 4);
+    assert_eq!(ldk_payment_summary.outgoing.total_failure, 2);
+    assert_eq!(ldk_payment_summary.incoming.total_success, 4);
+    assert_eq!(ldk_payment_summary.incoming.total_failure, 0);
```
joschisan, line 319: Actually the tests fails incoming payments due to missing liquidity so this should not be 0.

joschisan: Maybe we should do a sanity check here on the latencies just to see that they are calculated correctly.

gateway/ln-gateway/src/events.rs
```diff
@@ -22,4 +32,288 @@ pub const ALL_GATEWAY_EVENTS: [EventKind; 11] = [
...
+
+/// Computes the average of the given `u64` slice.
+fn average(data: &[u64]) -> u64 {
+    let sum: u64 = data.iter().sum();
+    if data.is_empty() {
+        0
+    } else {
+        sum / data.len() as u64
+    }
+}
```
joschisan, line 305: Could also return an option.

gateway/ln-gateway/src/events.rs
m1sterc001guy: I am having difficulty with this. If I take `all_events` as `IntoIter<Item = PersistedLogEntry>`, won't that consume `all_events`? I want to call this multiple times over the same slice. I think copying it between calls would be expensive, since it contains all of the events from the time duration.

gateway/ln-gateway/src/events.rs
m1sterc001guy: Done

gateway/ln-gateway/src/events.rs
dpc: @m1sterc001guy https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fc0641fe97ed45728a95a163bd2b9faf

gateway/ln-gateway/src/events.rs
m1sterc001guy: Done

gateway/ln-gateway/src/events.rs
m1sterc001guy: Done

gateway/ln-gateway/src/gateway_module_v2/events.rs
m1sterc001guy: Done

gateway/ln-gateway/src/events.rs
m1sterc001guy: Done

gateway/ln-gateway/src/events.rs
m1sterc001guy: Changed to `Duration`. Will add more stats in a follow up.

modules/fedimint-lnv2-tests/bin/tests.rs
m1sterc001guy: The liquidity check for incoming payments happens before the state machine spawns....so we don't currently log it as a payment failure. That can be fixed if we want, but I think its out of scope for this PR.

modules/fedimint-lnv2-tests/bin/tests.rs
m1sterc001guy: Added sanity check

gateway/ln-gateway/src/events.rs
m1sterc001guy: Think I have a version of this working. Please double check since its a bit tricky, added some comments to help explain :)

gateway/ln-gateway/src/events.rs
m1sterc001guy: Yeah this is a rather naive implementation of a join. A better implementation would use a map and that would avoid N^2 (at the expense of some memory for the map). I can re-implement if you'd like, but I thought since the number of total events will be small, that this would be ok.

dpc (approved):

PR #6583, opened 2024-12-18
chore: update flakebox (speedup pre-commit git hook)

dpc:


nix/check-forbidden-deps.sh
```diff
@@ -2,14 +2,21 @@
 
 set -eo pipefail
 
-if grep -E "(fedimint-mint|fedimint-wallet|fedimint-ln)" fedimint-server/Cargo.toml >&2 ; then
+if grep -E "fedimint-[a-zA-Z0-9]+-(server|client|common)" fedimint-server/Cargo.toml | grep -v -E "fedimint-api-client|fedimint-dummy-common|fedimint-dummy-server" >&2 ; then
```
dpc, line 5: Had to move this changes that we did to auto-generated file. Now that file comes with ... warnings.

bradleystachurski (approved):

PR #6596, opened 2024-12-19
feat: gateway CLI can generate bcrypt password hashes

tvolk131:
Since we switched to using bcrypt password hashes in #6482, we should make it easier for to-be gateway operators to create password hashes to use in their deployments. Here we're adding a command to the gateway CLI to create a password hash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2556,6 +2556,7 @@ name = "fedimint-gateway-cli"
 version = "0.6.0-alpha"
 dependencies = [
  "anyhow",
+ "bcrypt",
  "bitcoin",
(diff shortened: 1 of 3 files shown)


m1sterc001guy (commented):
I am ok with this, it definitely helps with deployment.

I would like to see some documentation in the gateway docs for how to generate the hash, ideally with another example using `htpasswd` or another tool as well. Can we add that here too?

One issue I ran into is that these hashes usually have the `$` character and that character needs to be escaped (i.e `$$`) is docker compose deployments

dpc (approved):

PR #6603, opened 2024-12-20
chore(core): db should ensure whole key and value where consumed

dpc:


m1sterc001guy (approved):

m1sterc001guy (approved):

m1sterc001guy:
Awesome that this caught an issue with our migration tests

fedimint-core/src/encoding/mod.rs
```diff
@@ -183,6 +183,35 @@ pub trait Decodable: Sized {
...
+
+        if left != 0 {
+            return Err(fedimint_core::encoding::DecodeError::new_custom(
+                anyhow::anyhow!(
+                    "Type did not consume all bytes during decoding; expected={}; left={}; type={}",
+                    total_len,
+                    left,
+                    std::any::type_name::<Self>(),
+                ),
+            ));
```
elsirion, line 211: How sure are we that this isn't a problem for any existing DB entries?

dpc: Very unlikely, and we do have migration tests which did immediately catch the one place where this was the case. @elsirion

PR #6605, opened 2024-12-20
chore(consensus): improve a noisy log statement

dpc:
Timeouts here are normal and should go away after per-peer retries are released.

### DIFF
--- a/fedimint-server/src/consensus/engine.rs
+++ b/fedimint-server/src/consensus/engine.rs
@@ -822,8 +822,11 @@ impl ConsensusEngine {
 
             match result {
                 Ok(signed_session_outcome) => return signed_session_outcome,
-                Err(error) => {
-                    tracing::error!(target: LOG_CONSENSUS, "Error while requesting signed session outcome: {}", error);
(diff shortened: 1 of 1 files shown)


fedimint-server/src/consensus/engine.rs
```diff
@@ -822,8 +822,11 @@ impl ConsensusEngine {
 
             match result {
                 Ok(signed_session_outcome) => return signed_session_outcome,
-                Err(error) => {
-                    tracing::error!(target: LOG_CONSENSUS, "Error while requesting signed session outcome: {}", error);
+                Err(err) => {
+                    // Workaround: timeout messages here are annoying
+                    if !err.to_string().contains("Rpc error: Request timeout") {
+                        warn!(target: LOG_CONSENSUS, %err, "Error while requesting signed session outcome");
+                    }
```
elsirion, line 829: This is expected in case we are requesting a session that is still running, right? I'd probably make it a trace level log in case it's a timeout, but also ok with removing.

elsirion (approved):

PR #6612, opened 2024-12-21
chore: print causes, but not backtrace

dpc:
https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations


fedimint-core/src/db/mod.rs
```diff
@@ -2116,7 +2116,7 @@ pub enum DecodingError {
     WrongPrefix { expected: u8, found: u8 },
     #[error("Key had a wrong length, expected {expected} but got {found}")]
     WrongLength { expected: usize, found: usize },
-    #[error("Other decoding error: {0:?}")]
+    #[error("Other decoding error: {0:#}")]
```
elsirion, line 2119: TIL, thx! Please remind me to mention it during the monday call!

elsirion (approved):

PR #6616, opened 2024-12-21
fix: cargo audit

dpc:


bradleystachurski (commented):
I think we can drop the one related to `proc-macro-error` (RUSTSEC-2024-0370) and bump aquamarine to 0.6.0, since that's the only one sneaking it in as a transitive dep.

dpc:
@bradleystachurski This one is for v0.5 and looks like audit is happy now.

bradleystachurski (approved):

Issue #6579, opened 2024-12-18
Weird flake of fedimint-ln-gateway::gatewayd-tests 


dpc:
```
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:07 2024-12-18T00:17:33.923159Z  INFO ln_gateway::state_machine: Successfully registered gateway 020085cb4ddc0a92b4a88eaa1dfd1509c1c5d5206573abb8a9db22df20d23569bd with federation 0406afdc71a052d2787eab7e84c95803636d2a84c272eb81b4e01b27acb86c6f
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:07 2024-12-18T00:17:33.923583Z ERROR fedimint_core::db: error=Other decoding error: Decoding named block field: FederationConfig{ ... invite_code ... }
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:07 thread 'test_gateway_executes_swaps_between_connected_federations' panicked at /build/source/fedimint-core/src/db/mod.rs:1580:9:
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:07 Unrecoverable decoding DatabaseValue as ln_gateway::db::FederationConfig; err=Other decoding error: Decoding named block field: FederationConfig{ ... invite_code ... }, key_bytes=0406afdc71a052d2787eab7e84c95803636d2a84c272eb81b4e01b27acb86c6f2a, val_bytes=bbb18c0568ca7b59
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:07 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fedimint-test-all-wasm32-unknown-unknown-ci> 00:00:07
```

https://github.com/fedimint/fedimint/actions/runs/12383535974/job/34566407494

dpc:
@fedimint/lightning

dpc:
https://github.com/fedimint/fedimint/actions/runs/12417503245/job/34668605487

m1sterc001guy:
Likely due to https://github.com/fedimint/fedimint/pull/6570, will investigate as soon as I can

m1sterc001guy:
One observation, each time the test fails, the federation ID starts with `0x04`, which is the key prefix for `FederationConfig`

[...]

PR #6618, opened 2024-12-21
fix: Gateway isolated DBs clash with `FederationConfig`

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/6579

We have a nasty bug in the gateway's database. The crux of the issue is that when we added client databases to the same global database as the gateway's DB, we forgot to give them their own prefix and instead relied on the `FederationId` to provide uniqueness. This is fine....most of the time. As long as a `FederationId` doesn't start with `0x04`, then everything is ok. But if it does, then it actually clashes with `FederationIdKey` which is used to key the `FederationConfig` and the database cannot tell the difference. Trying to read the `FederationConfig` will result in a decoding error.

The fix is to migrate the isolated databases to be behind a prefix so that it doesn't clash with `FederationIdKey`. A database migration has been added.

I don't think this issue has been hit in the wild. The public federations I have checked do not have IDs that start with 0x04. In case there is a gateway that connected to a federation that has this issue, I have added a migration that tries to handle it.

Looking for ideas on how to test the migration. I think a new upgrade test is probably needed.

### DIFF
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -4,12 +4,13 @@ use bitcoin::hashes::{sha256, Hash};
 use fedimint_api_client::api::net::Connector;
 use fedimint_core::config::FederationId;
 use fedimint_core::db::{
-    CoreMigrationFn, Database, DatabaseTransaction, DatabaseVersion,
+    CoreMigrationFn, Database, DatabaseTransaction, DatabaseVersion, IDatabaseTransactionOpsCore,
     IDatabaseTransactionOpsCoreTyped, MigrationContext,
 };
 use fedimint_core::encoding::btc::NetworkLegacyEncodingWrapper;
 use fedimint_core::encoding::{Decodable, Encodable};
 use fedimint_core::invite_code::InviteCode;
+use fedimint_core::module::registry::ModuleDecoderRegistry;
 use fedimint_core::{impl_db_lookup, impl_db_record, push_db_pair_items, secp256k1, Amount};
 use fedimint_ln_common::serde_routing_fees;
 use fedimint_lnv2_common::contracts::{IncomingContract, PaymentImage};
@@ -417,6 +418,7 @@ pub fn get_gatewayd_database_migrations() -> BTreeMap<DatabaseVersion, CoreMigra
     migrations.insert(DatabaseVersion(1), |ctx| migrate_to_v2(ctx).boxed());
     migrations.insert(DatabaseVersion(2), |ctx| migrate_to_v3(ctx).boxed());
     migrations.insert(DatabaseVersion(3), |ctx| migrate_to_v4(ctx).boxed());
+    migrations.insert(DatabaseVersion(4), |ctx| migrate_to_v5(ctx).boxed());
     migrations
 }
 
@@ -518,6 +520,77 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
             dbtx.insert_new_entry(&new_key, &new_fed_config).await;
         }
     }
+    Ok(())
+}
+
+async fn migrate_to_v5(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Error> {
+    let mut dbtx = ctx.dbtx();
+
+    // We need to migrate all isolated database entries to be behind the 0x10
+    // prefix. The problem is, if there is a `FederationId` that starts with
+    // 0x04, we cannot read the `FederationId` because the database will be confused
+    // between the isolated DB and the `FederationIdKey` record. To solve this,
+    // what we do first to try and see if there are any entries that begin with
+    // 0x0404. This indicates a joined federation that will have a
+    // problem decoding the ID.
+    let problem_fed_configs = dbtx
+        .raw_find_by_prefix(&[0x04, 0x04])
+        .await?
+        .collect::<BTreeMap<_, _>>()
+        .await;
+    for (problem_key, problem_fed_config) in problem_fed_configs {
+        let federation_id = FederationId::consensus_decode_vec(
+            problem_key[1..33].to_vec(),
+            &ModuleDecoderRegistry::default(),
+        )?;
+        tracing::warn!(
+            ?federation_id,
+            "Found a FederationConfig entry that will cause issues decoding"
+        );
+        let federation_id_bytes = problem_key[1..3].to_vec();
+
+        // Remove the `FederationConfig` entry so that the migration of the isolated
+        // database doesn't accidentally migrate that too.
+        dbtx.raw_remove_entry(&problem_key).await?;
+
+        let isolated_entries = dbtx
+            .raw_find_by_prefix(&federation_id_bytes)
+            .await?
+            .collect::<BTreeMap<_, _>>()
+            .await;
+        for (mut key, value) in isolated_entries {
+            let mut new_key = vec![DbKeyPrefix::ClientDatabase as u8];
+            new_key.append(&mut key);
+            dbtx.raw_remove_entry(&key).await?;
+            dbtx.raw_insert_bytes(&new_key, &value).await?;
+        }
+
+        // Re-insert the `FederationConfig` entry so that it exists after the migration.
+        dbtx.raw_insert_bytes(&problem_key, &problem_fed_config)
+            .await?;
+    }
+
+    // Migrate the rest of the isolated databases that don't overlap with
+    // `FederationConfig`
+    let fed_ids = dbtx
+        .find_by_prefix(&FederationIdKeyPrefix)
+        .await
+        .collect::<BTreeMap<_, _>>()
+        .await;
+    for fed_id in fed_ids.keys() {
+        let isolated_entries = dbtx
+            .raw_find_by_prefix(&fed_id.id.consensus_encode_to_vec())
+            .await?
+            .collect::<Vec<_>>()
+            .await;
+        for (mut key, value) in isolated_entries {
+            let mut new_key = vec![DbKeyPrefix::ClientDatabase as u8];
+            new_key.append(&mut key);
+            dbtx.raw_remove_entry(&key).await?;
+            dbtx.raw_insert_bytes(&new_key, &value).await?;
+        }
+    }
+
     Ok(())
 }
 
--- a/gateway/ln-gateway/src/db.rs
+++ b/gateway/ln-gateway/src/db.rs
@@ -525,12 +525,36 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
 
 async fn migrate_to_v5(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Error> {
     let mut dbtx = ctx.dbtx();
+    migrate_federation_config(&mut dbtx).await
+}
+
+async fn migrate_federation_config(
+    dbtx: &mut DatabaseTransaction<'_>,
+) -> Result<(), anyhow::Error> {
+    async fn migrate_client_entries(
+        dbtx: &mut DatabaseTransaction<'_>,
+        prefix: &[u8],
+    ) -> Result<(), anyhow::Error> {
+        let isolated_entries = dbtx
+            .raw_find_by_prefix(prefix)
+            .await?
+            .collect::<BTreeMap<_, _>>()
+            .await;
+        for (mut key, value) in isolated_entries {
+            dbtx.raw_remove_entry(&key).await?;
+            let mut new_key = vec![DbKeyPrefix::ClientDatabase as u8];
+            new_key.append(&mut key);
+            dbtx.raw_insert_bytes(&new_key, &value).await?;
+        }
+
+        Ok(())
+    }
 
     // We need to migrate all isolated database entries to be behind the 0x10
     // prefix. The problem is, if there is a `FederationId` that starts with
     // 0x04, we cannot read the `FederationId` because the database will be confused
     // between the isolated DB and the `FederationIdKey` record. To solve this,
-    // what we do first to try and see if there are any entries that begin with
+    // we first try and see if there are any entries that begin with
     // 0x0404. This indicates a joined federation that will have a
     // problem decoding the ID.
     let problem_fed_configs = dbtx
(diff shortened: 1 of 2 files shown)


dpc (commented):
LGTM

Do we need something like a new db snapshot to test?

gateway/ln-gateway/src/db.rs
```diff
@@ -23,6 +24,18 @@ use serde::{Deserialize, Serialize};
...
 use strum_macros::EnumIter;
 
+pub trait GatewayDbExt {
+    fn get_client_database(&self, federation_id: &FederationId) -> Database;
+}
+
+impl GatewayDbExt for Database {
+    fn get_client_database(&self, federation_id: &FederationId) -> Database {
+        let mut prefix = vec![DbKeyPrefix::ClientDatabase as u8];
+        prefix.append(&mut federation_id.consensus_encode_to_vec());
```
maan2003, line 34: in fedi we did incremented number (fed db prefix id) as prefix to get a small prefix and save lot of space (because this will repeated a lot)

then we store a map from federation id => db prefix id (u64)

ref https://github.com/fedixyz/fedi/blob/master/bridge/fedi-ffi/src/storage.rs#L512-L520

dpc: > because this will repeated a lot

I have not verified, but I read that rocksdb impls diff encoding for keys, so any common prefixes are mostly free.

I still like the sequnce ids better. :)

gateway/ln-gateway/src/db.rs
```diff
@@ -508,6 +523,77 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
...
 }
 
+async fn migrate_to_v5(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Error> {
+    let mut dbtx = ctx.dbtx();
+
+    // We need to migrate all isolated database entries to be behind the 0x10
+    // prefix. The problem is, if there is a `FederationId` that starts with
+    // 0x04, we cannot read the `FederationId` because the database will be confused
+    // between the isolated DB and the `FederationIdKey` record. To solve this,
+    // what we do first to try and see if there are any entries that begin with
```
tvolk131, line 533: Nit: `what we do first to try and see` -> `we first try and see`

gateway/ln-gateway/src/db.rs
```diff
@@ -508,6 +523,77 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
...
+            .raw_find_by_prefix(&federation_id_bytes)
+            .await?
+            .collect::<BTreeMap<_, _>>()
+            .await;
+        for (mut key, value) in isolated_entries {
+            let mut new_key = vec![DbKeyPrefix::ClientDatabase as u8];
+            new_key.append(&mut key);
+            dbtx.raw_remove_entry(&key).await?;
+            dbtx.raw_insert_bytes(&new_key, &value).await?;
+        }
```
tvolk131, line 566: This chunk of code is almost identical to the one below, other than it being collected into a `BTreeMap` instead of a `Vec`. Is there a reason for that difference? If not, could we define a helper function at the top of `migrate_to_v5()`? Something like this:

```rust
    async fn migrate_client_entries(
        prefix: &[u8],
        let isolated_entries = dbtx
            .raw_find_by_prefix(prefix)
            .await?
            .collect::<BTreeMap<_, _>>()
            .await;

        for (mut key, value) in isolated_entries {
            let mut new_key = vec![DbKeyPrefix::ClientDatabase as u8];
            new_key.append(&mut key);
            dbtx.raw_remove_entry(&key).await?;
            dbtx.raw_insert_bytes(&new_key, &value).await?;
        }

        Ok(())
    }
```

tvolk131 (changes requested):

gateway/ln-gateway/src/db.rs
m1sterc001guy: Yes, @tvolk131 and I verified this awhile back. Rocksdb will compress the common prefixes so there is very little space cost.

gateway/ln-gateway/src/db.rs
m1sterc001guy: Done

m1sterc001guy:
> Do we need something like a new db snapshot to test?

Added a test that manually creates a problematic database and migrates that.

gateway/ln-gateway/src/db.rs
```diff
@@ -508,6 +523,82 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
...
+
+    // We need to migrate all isolated database entries to be behind the 0x10
+    // prefix. The problem is, if there is a `FederationId` that starts with
+    // 0x04, we cannot read the `FederationId` because the database will be confused
+    // between the isolated DB and the `FederationIdKey` record. To solve this,
+    // we first try and see if there are any entries that begin with
+    // 0x0404. This indicates a joined federation that will have a
+    // problem decoding the ID.
+    let problem_fed_configs = dbtx
+        .raw_find_by_prefix(&[0x04, 0x04])
```
dpc, line 561: Could `0x04` ever be a legitimate `FederationConfig`? I guess it would need to be an invite code with a len 4, which I guess guess is not possible, but just double checking.

Also, in the future this whole migration would not run at all on new databases (which should be empty anyway), so I guess we don't need to worry about it...

m1sterc001guy: Yes, exactly

gateway/ln-gateway/src/db.rs
```diff
@@ -508,6 +523,82 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
...
+
+        // Re-insert the `FederationConfig` entry so that it exists after the migration.
+        dbtx.raw_insert_bytes(&problem_key, &problem_fed_config)
+            .await?;
+    }
+
+    // Migrate the rest of the isolated databases that don't overlap with
+    // `FederationConfig`
+    let fed_ids = dbtx
+        .find_by_prefix(&FederationIdKeyPrefix)
```
dpc, line 590: This prefix ... is empty. Doesn't it ... match everything in the database? I'm confused.

dpc: Hmm...

```rust
impl_db_record!(
    key = FederationIdKey,
    value = FederationConfig,
    db_prefix = DbKeyPrefix::FederationConfig,
);
```

The naming here is not ideal. SHouldn't db_prefix and key match in the names?

`s/FederationIdKey/FederationConfigKey/`? The `FederationIdKey` _contains_ federation id, but it's actually a key to all `FederationConfig`s, right?

Also `s/FederationIdKeyPrefix/FederationConfigPrefixAll` or something?=

dpc: OK, so this is actually the bulk of the migration:

"Read all the federation ids using `FederationConfig`s, and migrate keys that match to the correct version". Might update the comment as "Migrate the rest" gives the impression like this is some kind of "minor part".

But given that the federation config was

m1sterc001guy: The `impl_db_lookup` will automatically add the `0x04` prefix (defined in the `DbKeyPrefix` enum). So it will be all entries beginning with `0x04`

m1sterc001guy: Yeah we need to use the `FederationConfig` records that are indexed by the `FederationID` to know what the IDs are. Agree that the naming here is not ideal. Would like to leave that to a separate improvement though

gateway/ln-gateway/src/db.rs
dpc: I keep thinking ... is this a wrong way to go about it ... ?

Maybe we should all the `0x04` as raw, and use only ones that decode correctly as `FederationConfig` as an id, while ones that don't mark as a prefix to migrate? That would probably kind of read better too and wouldn't be so tricky.

dpc: I don't know ... Probably current impl works too.

dpc (approved):
AFACIT this will work, but would be good to get a second opinion because it's tricky.

gateway/ln-gateway/src/db.rs
```diff
@@ -646,4 +739,94 @@ mod fedimint_migration_tests {
...
+                    &GatewayPublicKey,
+                    &Keypair::new(secp256k1::SECP256K1, &mut rand::thread_rng()),
+                )
+                .await;
+            isolated_dbtx.commit_tx().await;
+        }
+
+        let federation_id = FederationId::from_str(
+            "0406afdc71a052d2787eab7e84c95803636d2a84c272eb81b4e01b27acb86c6f",
+        )
```
tvolk131, line 762: Is `FederationId::dummy()` below meant to be a federation ID that _doesn't_ start with the clashing prefix? If so, can we declare that as a stringified ID like we do here, and name the two IDs something like `clashing_federation_id` and `non_clashing_federation_id`? Expecting `dummy()` to return a federation ID that doesn't start with this problematic prefix adds an implicit API contract that, while unlikely to be broken, would silently lose test coverage if it was.

gateway/ln-gateway/src/db.rs
```diff
@@ -508,6 +523,82 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
     Ok(())
 }
 
+async fn migrate_to_v5(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Error> {
+    let mut dbtx = ctx.dbtx();
+    migrate_federation_config(&mut dbtx).await
+}
+
+async fn migrate_federation_config(
```
tvolk131, line 531: Nit: Pluralize to `migrate_federation_configs()`

gateway/ln-gateway/src/db.rs
```diff
@@ -508,6 +523,82 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
     Ok(())
 }
 
+async fn migrate_to_v5(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Error> {
```
tvolk131, line 526: Your PR description explains the issue quite well, but it's not documented in the code. Should we add a link to this PR in the documentation for `migrate_to_v5()` so future readers can understand why this migration exists? I know this code is unlikely to substantially change, so the git diff could be used, but I think having the link here would be easier for future readers.

tvolk131 (changes requested):

gateway/ln-gateway/src/db.rs
m1sterc001guy: Done

gateway/ln-gateway/src/db.rs
m1sterc001guy: Done

gateway/ln-gateway/src/db.rs
m1sterc001guy: I added a comment that should help.

gateway/ln-gateway/src/db.rs
m1sterc001guy: As discussed on the call, I re-implemented this using decoding as a filter.

gateway/ln-gateway/src/db.rs
m1sterc001guy: Renamed `FederationIdKey` to `FederationConfigKey`

gateway/ln-gateway/src/db.rs
```diff
@@ -501,13 +520,89 @@ async fn migrate_to_v4(mut ctx: MigrationContext<'_>) -> Result<(), anyhow::Erro
...
+) -> Result<(), anyhow::Error> {
+    // We need to migrate all isolated database entries to be behind the 0x10
+    // prefix. The problem is, if there is a `FederationId` that starts with
+    // 0x04, we cannot read the `FederationId` because the database will be confused
+    // between the isolated DB and the `FederationConfigKey` record. To solve this,
+    // we try and decode each key as a Federation ID and each value as a
+    // FederationConfig. If that is successful and the federation ID in the
+    // config matches the key, then we skip that record and migrate the rest of
+    // the entries.
+    let problem_entries = dbtx
```
dpc, line 550: `s/problem_entries/raw_federation_configs`?

m1sterc001guy: They might not be federation configs. They could just be from the isolated database.

dpc (approved):

gateway/ln-gateway/src/db.rs
dpc: There might also not be "problem entries", but just valid entries. Most of them, most of the time will be just valid configs.


The way I think about it - is we read through "raw federation configs" to find "valid ones".


I would actually do it other way around, even more defensively:

* go through all (probably) raw configs
* filter out invalid ones, and collect valid ones, deserializing them along the way
* then for all the valid deserialized ones, do the migration
* after all is done, do a normal prefixed "read all federation configs"
  * for every federation config, maybe even read some federation-prefixed key from the database to validate that it migrated right

dpc:
Landing as is, because flakes in the CI are troubling me too much. We can always follow through with improvements.

tvolk131 (approved):

fedimint-backports:
Backport failed for `releases/v0.5`, because it was unable to cherry-pick the commit(s).

```bash
git fetch origin releases/v0.5
git worktree add -d .worktree/backport-6618-to-releases/v0.5 origin/releases/v0.5
cd .worktree/backport-6618-to-releases/v0.5
git switch --create backport-6618-to-releases/v0.5
git cherry-pick -x 36c81eb4e31dd7f3ac69bd95b18f64271c5f1c6f f3cc65307c04d7bbf918155b8521b3ea566e58c7 807461f785c4a99e78ce3171c57e0c9d8a66d526 28feb76e0fd83c21f30d1451574b2f2142329445
```

m1sterc001guy:
@elsirion @bradleystachurski we're going to be cutting v0.6 soon right? If possible I'd like to avoid backporting, since this would also need to include 2 other database migrations that apparently didn't make v0.5.

m1sterc001guy:
Discussed offline, going to wait for v0.6

bradleystachurski:
Yep, I'd anticipate in the next week or two we'll begin the release process for v0.6. Much simpler to avoid the backports for multiple migrations if they aren't strictly necessary for v0.5.

Issue #6282, opened 2024-11-04
Test upgrade paths on PRs


elsirion:
Brought up when talking about #6272

elsirion:
@dpc how do we generate new perfit metric ids?

dpc:
@elsirion In infra repo `just agenix-edit secrets/perfitd-info.age`, and and look at `E.g. to add a new metric for something run`

PR #6619, opened 2024-12-23
Run upgrade tests against PRs and reduce total test time

bradleystachurski:
Closes https://github.com/fedimint/fedimint/issues/6282

This PR enables running upgrade tests against PRs and reduces the total time to run upgrade tests in CI vs the last three major versions from ~90 minutes to ~12 minutes ([CI run](https://github.com/fedimint/fedimint/actions/runs/12780718825/job/35640806526?pr=6619)).

These gains are accomplished by:
- Reducing the number of iterations run per upgrade test from 20 to 1, which still tests the same coverage of database entries
- Skipping running another round of `fedimintd` tests with the federation degraded, which is unnecessary since we have degraded federation test coverage throughout the test suite
- Parallelizing the upgrade tests to handle multiple upgrade paths in the same run

Since the total time is ~12 minutes, I don't see an advantage to only running against the latest version on push and the latest three in the merge queue, which is the approach we need to take with backwards-compatibility tests.

### DIFF
--- a/devimint/src/tests.rs
+++ b/devimint/src/tests.rs
@@ -179,7 +179,11 @@ pub async fn latency_tests(
             let mut ln_sends = Vec::with_capacity(iterations);
             for i in 0..iterations {
                 let invoice = cln
-                    .invoice(1_000_000, format!("Description{i}"), format!("Label{i}"))
+                    .invoice(
+                        1_000_000,
+                        format!("Description{i}"),
+                        format!("Label{}", rand::random::<u64>()),
+                    )
                     .await?;
                 let start_time = Instant::now();
                 ln_pay(&client, invoice, lnd_gw_id.clone(), false).await?;
@@ -207,7 +211,7 @@ pub async fn latency_tests(
                 .invoice(
                     10_000_000,
                     "LnReceiveLatencyDesc".to_string(),
-                    "LatencyLabel".to_string(),
+                    rand::random::<u64>().to_string(),
                 )
                 .await?;
             ln_pay(&client, invoice, lnd_gw_id.clone(), false).await?;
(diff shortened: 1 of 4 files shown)


devimint/src/tests.rs
```diff
@@ -179,7 +179,11 @@ pub async fn latency_tests(
             let mut ln_sends = Vec::with_capacity(iterations);
             for i in 0..iterations {
                 let invoice = cln
-                    .invoice(1_000_000, format!("Description{i}"), format!("Label{i}"))
+                    .invoice(
+                        1_000_000,
+                        format!("Description{i}"),
+                        format!("Label{}", rand::random::<u64>()),
```
bradleystachurski, line 185: We need unique labels for cln otherwise we get the following error

```
 Error: Error code 900: Duplicate label 'Label0'
```

@dpc this is the mystery discussed in https://discord.com/channels/990354215060795454/1314093509640192000/1314323211051073566

bradleystachurski:
Once this merges, let's make the Upgrade Tests workflow required.

dpc (approved):

devimint/src/tests.rs
```diff
@@ -381,6 +385,10 @@ async fn stress_test_fed(dev_fed: &DevFed, clients: Option<&UpgradeClients>) ->
     // cause the upgrade test to fail
     let assert_thresholds = false;
 
+    // running only one iteration greatly improves the total test time while still
+    // testing the same types of database entries
+    let iterations = 1;
```
dpc, line 390: Probably easy to source it from env var and default to 1. @bradleystachurski

bradleystachurski: Capturing our chat out-of-band: let's keep as-is so it's not  easy to accidentally increase the runtime on the upgrade tests. If someone needs to reuse `stress_test_fed` outside of upgrade tests, we can refactor then.

PR #6623, opened 2024-12-23
chore: include incremental time change in test output

dpc:

### DIFF
--- a/scripts/dev/run-test/fm-run-test
+++ b/scripts/dev/run-test/fm-run-test
@@ -37,7 +37,7 @@ on_exit() {
 trap on_error ERR
 trap on_exit EXIT
 
-command time -q --format="$time_fmt" -o "$time_out_path" timeout -k 310 320 "$@" 2>&1 | ts -s > "$test_out_path"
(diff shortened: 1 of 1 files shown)


scripts/dev/run-test/fm-run-test
```diff
@@ -37,7 +37,7 @@ on_exit() {
 trap on_error ERR
 trap on_exit EXIT
 
-command time -q --format="$time_fmt" -o "$time_out_path" timeout -k 310 320 "$@" 2>&1 | ts -s > "$test_out_path"
+command time -q --format="$time_fmt" -o "$time_out_path" timeout -k 310 320 "$@" 2>&1 | ts -i "%.S" | ts -s "%M:%S" > "$test_out_path"
```
elsirion, line 40: Might be a bit noisy?

elsirion (approved):

scripts/dev/run-test/fm-run-test
dpc: We'll see.

We only print these when something fails and it's just 10 characters in the front.

PR #6626, opened 2024-12-25
chore: inherit more from workspace Cargo.toml

dpc:
I discovered that it's possible to inherit almost all fields in each `package` from the root `Cargo.toml`

This is just to show it by using in few packages.

### DIFF
--- a/fedimint-cli/Cargo.toml
+++ b/fedimint-cli/Cargo.toml
@@ -1,12 +1,13 @@
 [package]
 name = "fedimint-cli"
+edition = { workspace = true }
+homepage = { workspace = true }
+keywords = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
 version = { workspace = true }
-authors = ["The Fedimint Developers"]
-edition = "2021"
-description = "fedimint-cli is a command line interface wrapper for the client library."
-license = "MIT"
-readme = "../README.md"
-repository = "https://github.com/fedimint/fedimint"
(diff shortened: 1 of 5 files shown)


fedimint-cli/Cargo.toml
```diff
@@ -1,12 +1,13 @@
 [package]
 name = "fedimint-cli"
+edition = { workspace = true }
```
maan2003, line 3: `edition.workspace = true` maybe

dpc: That's a nice improvement, huh.

I tried it (`sed -i -e 's/ = { workspace = true }/.workspace = true/g' **.toml` in Fish) and unfortunately `cargo-sort` for some reason can't part it :/):

```
❯ j cargo-sort-check
cargo sort -w -g --order package,features,bin,lib,test,bench,dependencies,dev-dependencies,build-dependencies --check
Checking fedimint...
Checking aead...
thread 'main' panicked at src/sort.rs:113:46:
called `Result::unwrap()` on an `Err` value: TomlError { message: "TOML parse error at line 3, column 8\n  |\n3 | version.workspace = true\n  |        ^\nUnexpected `.`\nExpected `=`\n" }
error: Recipe `cargo-sort-check` failed on line 266 with exit code 101
```

tvolk131:
I like it! Let's do it this for of them!

maan2003 (approved):

PR #6628, opened 2024-12-26
chore: add Bitcoind::get_block_height()

tvolk131:
In all but one place where `get_block_count()` is called, we subtract 1. Let's add `get_block_height()` so we can get rid of that and make the code slightly easier on the eyes.

### DIFF
--- a/devimint/src/external.rs
+++ b/devimint/src/external.rs
@@ -169,7 +169,7 @@ impl Bitcoind {
     /// Poll until bitcoind rpc responds for basic commands
     async fn poll_ready(&self) -> anyhow::Result<()> {
         poll("bitcoind rpc ready", || async {
-            self.get_block_count()
+            self.get_block_height()
                 .await
                 .map_err(ControlFlow::Continue::<anyhow::Error, _>)?;
             Ok(())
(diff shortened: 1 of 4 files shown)


m1sterc001guy (approved):

dpc (commented):
Seems to me like `wait_for_block_height` on gateway should be `wait_for_block_count` instead.

The idea of "block height" as way to describe the state of the chain is a contagious disease.

dpc (changes requested):
Not so fast! A wild block-count-maximalist appeared. :D

dpc:
*Comment by Tau/GPT-5*

Triage note: I think we should close this stale PR rather than keep it open.

The original cleanup (adding `Bitcoind::get_block_height()` to avoid local `get_block_count() - 1` conversions) still applies in a few devimint spots, but the PR has had changes requested since Jan 2025: the requested direction was to avoid spreading “block height” terminology and instead prefer block-count semantics (e.g. `wait_for_block_count`). The branch is also now conflict/dirty against `master`, and the surrounding code has moved since the PR was opened.

If we still want this cleanup, it should probably come back as a fresh small PR in the block-count direction rather than reviving this branch. We can always reopen if I missed some reason to keep this exact PR alive.

PR #6635, opened 2025-01-01
Recovery data usage improvements

maan2003:
goal: reduce data usage for recovery

- changed from hex to base64: 33% saving
- client can only download from single peer and verify using signature: num peers times saving

for 4 guardian federation: overall 6x less data download

Happy new year!

### DIFF
--- a/fedimint-api-client/src/api/global_api.rs
+++ b/fedimint-api-client/src/api/global_api.rs
@@ -4,13 +4,14 @@ use std::num::NonZeroUsize;
 use std::result;
 use std::sync::Arc;
 
-use anyhow::anyhow;
+use anyhow::{anyhow, format_err};
 use bitcoin::hashes::sha256;
 use bitcoin::secp256k1;
 use fedimint_core::admin_client::{
     ConfigGenConnectionsRequest, ConfigGenParamsRequest, ConfigGenParamsResponse, PeerServerParams,
 };
 use fedimint_core::backup::ClientBackupSnapshot;
+use fedimint_core::config::ClientConfig;
 use fedimint_core::core::backup::SignedBackupRequest;
 use fedimint_core::core::ModuleInstanceId;
 use fedimint_core::endpoint_constants::{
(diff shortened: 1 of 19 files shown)


fedimint-core/src/module/mod.rs
```diff
@@ -964,6 +971,59 @@ impl<T: Encodable + Decodable + 'static> SerdeModuleEncoding<T> {
...
+
+    /// In cases where we know exactly which module kind we expect but don't
+    /// have access to all decoders this function can be used instead.
+    ///
+    /// Note that it just assumes the decoded module instance id to be valid
+    /// since it cannot validate against the decoder registry. The lack of
+    /// access to a decoder registry also makes decoding structs impossible that
+    /// themselves contain module dyn-types (e.g. a module output containing a
+    /// fedimint transaction).
+    pub fn try_into_inner_known_module_kind(&self, decoder: &Decoder) -> Result<T, DecodeError> {
```
dpc, line 1009: Sus. Why.

fedimint-core/src/module/mod.rs
```diff
@@ -911,6 +911,13 @@ pub struct SerdeModuleEncoding<T: Encodable + Decodable>(
     #[serde(skip)] PhantomData<T>,
 );
 
+/// Same as [`SerdeModuleEncoding`] but uses base64 instead of hex encoding.
+#[derive(Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct SerdeModuleEncoding2<T: Encodable + Decodable>(
```
dpc, line 916: `s/2/Base64/`?

fedimint-api-client/src/api/global_api.rs
```diff
@@ -129,12 +132,56 @@ where
...
+    }
+
     async fn get_session_status_raw(
         &self,
         block_index: u64,
         decoders: &ModuleDecoderRegistry,
     ) -> anyhow::Result<SessionStatus> {
         debug!(target: LOG_CLIENT_NET_API, block_index, "Fetching block's outcome from Federation");
+        let verify = |_: &SignedSessionOutcome| -> bool { todo!() };
+        // TODO: check api version
```
dpc, line 147: Probably just make it a new function, and let the caller verify. I think all over the code we do the api versioning checks in the higher layer. Though passing api version here wouldn't be bad either.

fedimint-api-client/src/api/global_api.rs
```diff
@@ -129,12 +132,56 @@ where
...
+                    // use fallback method
+                    break;
+                }
+                Err(_) => {
+                    // try next peer
+                    continue;
+                }
+            }
+        }
         self.request_current_consensus::<SerdeModuleEncoding<SessionStatus>>(
```
dpc, line 185: Is there any point keeping now?

fedimint-api-client/src/api/global_api.rs
```diff
@@ -129,12 +132,56 @@ where
...
         &self,
         block_index: u64,
         decoders: &ModuleDecoderRegistry,
     ) -> anyhow::Result<SessionStatus> {
         debug!(target: LOG_CLIENT_NET_API, block_index, "Fetching block's outcome from Federation");
+        let verify = |_: &SignedSessionOutcome| -> bool { todo!() };
+        // TODO: check api version
+        // fetch serially
+        for peer_id in self.select_peers_for_status().cycle() {
+            // TODO: maybe even multiple retries with same peer.
```
dpc, line 150: Doesn't higher layer already retry?

fedimint-api-client/src/api/global_api.rs
maan2003: we still need current consensus for pending and initial values because those don't have signature

fedimint-api-client/src/api/global_api.rs
maan2003: right, will just remove it

fedimint-api-client/src/api/global_api.rs
```diff
@@ -129,12 +132,56 @@ where
         .map_err(|e| anyhow!(e.to_string()))
     }
 
+    fn select_peers_for_status(&self) -> impl Iterator<Item = PeerId> + '_ + Clone {
+        // TODO: better selection
```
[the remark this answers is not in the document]
maan2003: i was thinking about avoiding the down guardian but random is still pretty good

fedimint-core/src/module/mod.rs
dpc: TODO: verify that the original wasn't written by me and silly.

fedimint-api-client/src/api/global_api.rs
dpc: Oh, right.

fedimint-api-client/src/api/global_api.rs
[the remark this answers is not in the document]
dpc: It's still a massive improvement and whatever is simple should work.

maan2003:
simplified the control flow

maan2003:
@dpc thoughts about https://github.com/fedimint/fedimint/pull/6635/commits/9bf85395af91865e1e223e258623d5444f8df5e9?

I don't think it will create any locking issues

dpc:
@maan2003 I might be off and not have time to fully bring it all in, but there were two caches because session status can have partial results, or something.

I would avoid adding too much changes to one PR for now unless absolutely necessary.

dpc:
```
    /// In theory these two LRUs have the same content, but one is locked by
    /// potentially long-blocking operation, while the other non-blocking one.
    /// Given how tiny they are, it's not worth complicating things to unify
    /// them.
```

Await blocks, while holding the lock there, right. So there is an issue, or at least at some point I believe there was (could have been wrong). Later this/next week I could investigate to fully understand the issue, but for now not worth blocking on.

dpc:
Right. So the goal of "await" is blocking until the result is there, while a concurrent "check status" for the same session wants to get a partial result, and not block, so would get stuck if there was a blocked "await", leading to random and unwanted hanging in certain cases (which we might add later). Even if there's a way to somehow unify these cashes, the complexity of it is probably not worth the tiny memory savings.

fedimint-core/src/module/mod.rs
dpc: OK, so it's used to decode session/output outcomes in client modules sometimes, doesn't panic. We always take it from current consensus of peers (except now, where we check the signature so effectively the same thing). Looks OK.

dpc (dismissed):

maan2003:
👀 looks like recovery tests is failing, will debug

dpc:
Rebased and force-pushed. Will look into failing test soon.

fedimint-core/src/encoding/mod.rs
```diff
@@ -590,7 +591,7 @@ impl DecodeError {
 
 impl std::fmt::Display for DecodeError {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Display::fmt(&self.0, f)
+        f.write_fmt(format_args!("{:#}", self.0))
```
dpc, line 594: This was soooo annoying.

dpc (dismissed):

dpc (dismissed):

dpc (approved):

Issue #6601, opened 2024-12-20
feat: `fedimint-cli module wallet check-address` and `gateway-cli check-address`


dpc:
We need a way to check peg-in address manually from CLI for both of them.

dpc:
Assigning to myself, but if someone from @fedimint/lightning wants to do at least the gw part, it might be much easier for them.

PR #6647, opened 2025-01-03
chore(client-wallet): new `fedimint-cli module wallet` cli commands

dpc:
Partially implements #6601 , plus moves a bit forward towards moving more stuff into module specific cli commands.

Notably I breaks backward compat of `fedimint-cli module wallet await-deposit` command, as I think it should default to `bitcoin::Address` argument, which the user probably cares more about than `OperationId`. Though both are still supported.

### DIFF
--- a/modules/fedimint-wallet-client/src/cli.rs
+++ b/modules/fedimint-wallet-client/src/cli.rs
@@ -1,17 +1,26 @@
+use std::str::FromStr as _;
 use std::{ffi, iter};
 
+use anyhow::bail;
+use bitcoin::address::NetworkUnchecked;
 use clap::Parser;
 use fedimint_core::core::OperationId;
 use serde::Serialize;
 
 use super::WalletClientModule;
 use crate::api::WalletFederationApi;
+use crate::client_db::TweakIdx;
 
 #[derive(Parser, Serialize)]
 enum Opts {
     /// Await a deposit on a given deposit address
     AwaitDeposit {
-        operation_id: OperationId,
+        addr: Option<String>,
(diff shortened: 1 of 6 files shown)


modules/fedimint-wallet-client/src/cli.rs
```diff
@@ -31,11 +50,40 @@ pub(crate) async fn handle_cli_command(
...
+            {
+                bail!("One and only one of the selector arguments must be set")
+            }
+            if let Some(tweak_idx) = tweak_idx {
+                module.await_num_deposits(tweak_idx, num).await?;
+            } else if let Some(operation_id) = operation_id {
+                module
+                    .await_num_deposits_by_operation_id(operation_id, num)
+                    .await?;
+            } else if let Some(addr) = addr {
```
elsirion, line 72: Wondering if there's a  nice abstraction we could use here. But would likely be a macro or very specific to the types here :/

modules/fedimint-wallet-client/src/lib.rs
```diff
@@ -830,6 +830,29 @@ impl WalletClientModule {
...
+            .begin_transaction_nc()
+            .await
+            .find_by_prefix(&PegInTweakIndexPrefix)
+            .await
+            .filter(|(k, _)| {
+                let (_, derived_address, _tweak_key, _) = data.derive_peg_in_script(k.0);
+                future::ready(derived_address.into_unchecked() == address)
+            })
+            .next()
+            .await
```
elsirion, line 849: Probably not a problem now, but less efficient than I'd like :grimacing:

PR #6652, opened 2025-01-06
chore: make Client::get_config sync

maan2003:
always use the correct lock for the job


elsirion:
Is there any reason we might want the fn to be async even though the lock is sync? E.g. needing to read something from the DB in the future? I'm inclined to make things async even if not strictly needed.

maan2003:
even if we read from database, we can always have in memory copy and updating it on writes

maan2003:
stale

PR #6655, opened 2025-01-06
chore: port (some) mprocs aliases to devimint-env

dpc:
I'm not using `just mprocs` anymore, but I want aliases too!

To test, `just mprocs` and/or `just devimint-env` and see that `fedimint-cli info`, `gateway-ldk info` and `gateway-lnd info` work.

### DIFF
--- a/scripts/dev/aliases.sh
+++ b/scripts/dev/aliases.sh
@@ -1,11 +1,22 @@
 #!/usr/bin/env bash
 
+# The original binaries from the environment, before the aliases potentially took them over
+# This is so that aliases can call actual commands, instead of each other/itself leading
+# to infinite recursion.
+export FM_ALIAS_ORIG_FEDIMINT_CLI
+FM_ALIAS_ORIG_FEDIMINT_CLI="$(which fedimint-cli)"
+export FM_ALIAS_ORIG_GATEWAY_CLI
+FM_ALIAS_ORIG_GATEWAY_CLI="$(which gateway-cli)"
+
+export PATH="${REPO_ROOT}/scripts/dev/devimint/aliases/:${PATH}"
+
+# Note: Please add new aliases as script to the directory above,
+# and migrate existing ones over time, so they work cross-shells
+# and cross-tools.
+# Also, please see https://github.com/fedimint/fedimint/issues/6658
 alias lightning-cli="\$FM_LIGHTNING_CLI"
 alias lncli="\$FM_LNCLI"
 alias bitcoin-cli="\$FM_BTC_CLIENT"
-alias fedimint-cli="\$FM_MINT_CLIENT"
-alias gateway-lnd="\$FM_GWCLI_LND"
-alias gateway-ldk="\$FM_GWCLI_LDK"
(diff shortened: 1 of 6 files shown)


scripts/dev/devimint/aliases/fedimint-cli
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+$CARGO_BUILD_TARGET_BIN_DIR/$FM_MINT_CLIENT "$@"
```
dpc, line 3: This concatenation is kind of clunky, and generally I think `devimint` should stop setting combined aliases like `FM_MINT_CLIENT`, and only set each thing that the caller can't  know about because `devimint` itself picks them (like data or tmp dirs).

dpc: Just created #6658 and came up with a more elegant approach.

scripts/dev/aliases.sh
```diff
@@ -1,11 +1,22 @@
...
+export PATH="${REPO_ROOT}/scripts/dev/devimint/aliases/:${PATH}"
+
+# Note: Please add new aliases as script to the directory above,
+# and migrate existing ones over time, so they work cross-shells
+# and cross-tools.
+# Also, please see https://github.com/fedimint/fedimint/issues/6658
 alias lightning-cli="\$FM_LIGHTNING_CLI"
 alias lncli="\$FM_LNCLI"
 alias bitcoin-cli="\$FM_BTC_CLIENT"
-alias fedimint-cli="\$FM_MINT_CLIENT"
```
dpc, line 6: Migrated just these 3 for now.

scripts/dev/devimint/aliases/fedimint-cli
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+$FM_ALIAS_ORIG_FEDIMINT_CLI --data-dir "$FM_CLIENT_DIR" "$@"
```
maan2003, line 3: i don't like having to manually do aliasing instead of using devimint environment variables :/

scripts/dev/devimint-env.sh
```diff
@@ -20,6 +20,8 @@ function devimint_env {
   # For starship users, we can actually make the prompt distinct so there's
   # no confusion.
   export STARSHIP_CONFIG="${REPO_ROOT}/scripts/dev/devimint-env/starship.toml"
+  source "${REPO_ROOT}/scripts/dev/aliases.sh"
```
maan2003, line 23: for fish we can do `fish --init-command 'source aliases.sh`  but bash can't do much.

scripts/dev/devimint/aliases/fedimint-cli
dpc: It's not like env variable shows up in the PATH magically, so something still needs to be done to actually make it usable.

scripts/dev/devimint-env.sh
dpc: PATH works for all shells, current and future ones.

maan2003 (approved):

scripts/dev/devimint/aliases/fedimint-cli
dpc: @maan2003 https://github.com/fedimint/fedimint/pull/6661

PR #6656, opened 2025-01-06
Activate wallet consensus rules

elsirion:
Automatically activates wallet module consensus upgrades if **every guardian** supports a newer consensus version than is currently active. If not, we still rely on manual upgrade votes, which can activate new consensus versions with a threshold of guardians, kicking out the remaining ones.

### DIFF
--- a/fedimint-server-core/src/init.rs
+++ b/fedimint-server-core/src/init.rs
@@ -6,6 +6,7 @@ use std::marker;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
+use fedimint_api_client::api::DynModuleApi;
 use fedimint_core::config::{
     ClientModuleConfig, CommonModuleInitRegistry, ConfigGenModuleParams, ModuleInitParams,
     ModuleInitRegistry, ServerModuleConfig, ServerModuleConsensusConfig,
@@ -44,6 +45,7 @@ pub trait IServerModuleInit: IDynCommonModuleInit {
         db: Database,
         task_group: &TaskGroup,
         our_peer_id: PeerId,
+        module_api: DynModuleApi,
     ) -> anyhow::Result<DynServerModule>;
 
     fn validate_params(&self, params: &ConfigGenModuleParams) -> anyhow::Result<()>;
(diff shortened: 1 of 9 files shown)


bradleystachurski (commented):
After a quick review it looks good! I'll give a more thorough review soon.

I'm considering if there's a reason to keep around manual activation. Perhaps a scenario where it's still useful is if only 3 peers are online in a 4 member federation and it's known the 4th peer is unable to ever come back online. Automatic voting requires a response from all peers where manual will allow the 3 peers to activate voting.

joschisan:
Good work on fixing the race condition in the tests.

elsirion:
> I'm considering if there's a reason to keep around manual activation

Note that I did not path out manual voting for the reason you mentioned. Just not a high prio to expose imo and needs some more work (right now we never stop voting afaik).

https://github.com/fedimint/fedimint/blob/453ce6a381efeaf7ed34f324c6edb3f3d4fd8d51/modules/fedimint-wallet-server/src/lib.rs#L467-L485

bradleystachurski:
> we might also want to vote on the version supported by everyone and not the highest one we support if we activate automatically

Are we already safe in that scenario since `self.consensus_module_consensus_version` calculates the threshold min supported?

elsirion:
> Are we already safe in that scenario since `self.consensus_module_consensus_version` calculates the threshold min supported?

I don't think so, assume the following situation:

```
current = 1
supported = [ 2, 3, 3, 3]
```

We'd want to vote for `2` since that's a more recent version supported by everyone, but before the fixup I just pushed the `3`s would vote for `3` instead of `2`.

bradleystachurski (approved):
Looks good :+1: 

I verified locally that the client still handles both v0 and v1 inputs based on the new timing in tests introduced with these changes.

Using this issue (https://github.com/fedimint/fedimint/issues/6736) to track some of the wallet specific TODOs from this PR. I don't think any of them are blockers for 0.6.

fedimint-server-core/src/init.rs
```diff
@@ -44,6 +45,7 @@ pub trait IServerModuleInit: IDynCommonModuleInit {
         db: Database,
         task_group: &TaskGroup,
         our_peer_id: PeerId,
+        module_api: DynModuleApi,
```
dpc, line 48: It was supposed to be simpler, now we need to inject DynModuleApi to modules. :P

elsirion: Yeah, I wish we had implemented the upgrade logic on the core level :cry:

bradleystachurski (approved):

PR #6661, opened 2025-01-07
chore(devimint): ship aliases with `devimint` package

dpc:
Superseeds #6655

Make aliases standalone binaries, so they work in any setup, tooling, shell.

Ship `aliases` as a part of "static data" of devimint package. The location of the "static data" defaults to `/usr/devimint`, but can be overridden during build time and runtime. On Nix packaging will take care of setting the default either through a shell script (our flake case), or by setting it during build time (probably easiest approach in NixOS).

### DIFF
--- a/nix/flakebox.nix
+++ b/nix/flakebox.nix
@@ -328,7 +328,18 @@ in
       }
     );
 
+    devimintStaticDataDir = pkgs.stdenv.mkDerivation {
+      name = "devimint-static-data-dir";
+      src = pkgs.emptyDirectory;
+
+      installPhase = ''
+        mkdir -p $out/share/
+        cp -r ${../devimint/share}/* $out/share/devimint/
+      '';
+    };
+
     # copied and modified from flakebox, to add `runLowPrio`, due to mistake in flakebox
+    # and also, it wraps the `devimint` into a shell scrip that points at the the static
     rawBuildPackageGroup =
       {
         pname ? null,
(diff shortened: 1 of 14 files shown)


nix/flakebox.nix
```diff
@@ -716,7 +747,11 @@ in
 
     devimint = flakeboxLib.pickBinary {
       pkg = devimint-pkgs;
-      bin = "devimint";
+      name = "devimint";
+      bins = [
+        "devimint"
```
dpc, line 752: Had to update `flakebox` to add support for copying-out multiple binaries.

scripts/dev/user-shell.sh
```diff
@@ -20,7 +20,7 @@ echo Done!
 echo
 echo "This shell provides the following aliases:"
 echo ""
-echo "  fedimint-cli   - cli client to interact with the federation"
+echo "  fm-cli   - cli client to interact with the federation"
```
dpc, line 23: Aliases should not override the original commands. That's confusing and prevents certain scenarios.

We might come up with an approach to unify the naming here or something, but this will be just a matter of renaming scripts in `./devimint/share`

dpc:
@maan2003 cool? :D

maan2003 (approved):

nix/flakebox.nix
```diff
@@ -328,7 +328,18 @@ in
...
+      src = pkgs.emptyDirectory;
+
+      installPhase = ''
+        mkdir -p $out/share/
+        cp -r ${../devimint/share}/* $out/share/devimint/
+      '';
+    };
+
     # copied and modified from flakebox, to add `runLowPrio`, due to mistake in flakebox
+    # and also, it wraps the `devimint` into a shell scrip that points at the the static
```
bradleystachurski, line 342: [nit] s/scrip/script

bradleystachurski (approved):

PR #6663, opened 2025-01-08
chore: rename consensus_decode to consensus_decode_partial

dpc:
Recently I discovered that we were not checking if the keys and values in the DB were fully decoded, so I introduced `Decodable::consensus_decode_whole` to make it more visible and upfront.

I was just writting https://github.com/rust-bitcoin/rust-bitcoin/issues/3887 to share this lesson with `rust-bitcoin` devs, and there I wrote:

> I was (and still kind of am) considering renaming `consensus_decode` to `consensus_decode_partial`, (and `consensus_decode_from_finite_reader` to `consensus_decode_partial_from_finite_reader`.

and inspired by it, I decide to check some stuff, and indeed the helpers `Decodable::consensus_decode_hex` and `Decodable::consensus_decode_vec` were forwarding to `Decodable::consensus_decode`, which means they ignore extra bytes. I checked where they were used and they were all used in places where extra bytes should not be happening, and possibly even an issue.

This was the last straw that made me convinced that `_partial` must there in the names, because 99% of the time the user will want the behavior of `_whole`. `Decodable::consensus_decode` ignoring bytes seems like a default when you are designing this trait and implementing things like derives that compose types together, but after that is done, downstream users will only ever (99%) be decoding already framed messages etc. and not stream multiple types via a single reader.

After going through the refactor I discover lots of places where we should be just using `_whole`, but habitually we've just defaulted to partial.

### DIFF
--- a/crypto/tbs/benches/tbs.rs
+++ b/crypto/tbs/benches/tbs.rs
@@ -1,5 +1,4 @@
 use std::collections::BTreeMap;
-use std::io::Cursor;
 
 use bls12_381::{G2Projective, Scalar};
 use criterion::{criterion_group, criterion_main, Criterion};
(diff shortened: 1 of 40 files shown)


tcharding:
Typo in the quote and original @dpc. "(and consensus_decode_from_finite_reader to consensus_decode_from_finite_reader" - one of these should have a different name.

elsirion (approved):

Issue #6220, opened 2024-10-24
Check against double-issued e-cash


elsirion:
Currently, the server doesn't care if clients issue the same e-cash twice and burning money that way. We could relatively easily create an index of all the issued blind nonces to

1. Check on transaction submission if a note was already issued
2. Provide an API which clients can use themselves to check the status of their wallet

This should not impact privacy unless clients hit a bug where e-cash is double-issued, but in that case privacy is compromised anyway and it's preferable to get the client into a good state again.

dpc:
Would be great if this was a consensus check eventually. That would be a relatively easy change... if we had something like ... consensus voting ... at the fedimint-core level. 🤦

elsirion:
Needs some testing of [#6665](https://github.com/fedimint/fedimint/pull/6665), likely good enough.

PR #6665, opened 2025-01-08
feat(wallet-server): reject reused blind nonces during tx submission

dpc:
Impl non-consensus version of #6220

### DIFF
--- a/modules/fedimint-mint-common/src/lib.rs
+++ b/modules/fedimint-mint-common/src/lib.rs
@@ -269,4 +269,6 @@ pub enum MintOutputError {
     InvalidAmountTier(Amount),
     #[error("The mint output version is not supported by this federation")]
     UnknownOutputVariant(#[from] UnknownMintOutputVariantError),
+    #[error("The mint output blind nonce was already used before")]
+    BlindNonceAlreadyUsed,
 }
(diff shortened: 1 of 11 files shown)


dpc:
@joschisan I think the API itself to make submission and processing itself is the most controversial, so let's agree or bikeshed it first, then I can figure out rest of the implementation details.

modules/fedimint-mint-common/src/lib.rs
```diff
@@ -269,4 +269,6 @@ pub enum MintOutputError {
     InvalidAmountTier(Amount),
     #[error("The mint output version is not supported by this federation")]
     UnknownOutputVariant(#[from] UnknownMintOutputVariantError),
+    #[error("The mint output blind nonce was already used before")]
+    BlindNonceAlreadyUsed,
```
dpc, line 273: New variant of the error. In practice it shouldn't happen, but when it does `SUBMIT_TRANSACTION_ENDPOINT.` on the client side will get confused.

Old clients will not understand it and probably fail to deserialize.

New clients in partially upgraded federation will not get `request_current_consensus_retry` and keep retrying.

Still better than burning money though. :shrug:

dpc:
@elsirion 

So when a tx fail due to nonce reuse:

* if the client is not upgraded it will keep failing to decode responses from peers that were upgraded
* if the federation is not fully upgraded, and client is, it will not be able to get consensus on submission result and keep submitting (funds kind of stuck)
* a client that is upgraded in a federation that is upgraded will just recognize the tx is rejected, inputs will get refunded, outputs abandoned, and since each time a nonce is allocated we increment the index in the db, it will try to use next nonces next time

All in all it's a hack, but all it rough edges are better than user burning money due to client malfunction/reuse.

dpc:
LOL. It looks like it's rejecting funds being burned in some tests? :D

elsirion (approved):
:pray:

elsirion:
Needs rebase :grimacing:

dpc:
Rebased.

elsirion (approved):

PR #6667, opened 2025-01-09
Cleanup cryptography around dealer dkg

joschisan:


dpc (approved):
LGTM, probably best to get Eric to take a look, even after the fact if you want to land.

PR #6669, opened 2025-01-09
chore: turn off tor by default

m1sterc001guy:
Tor pulls in a sqlite dependency, which requires downstream projects that depend on these crates to depend on sqlite in order to build, even if they're not using tor. Tor is already a feature, I think it makes sense to be optional.

### DIFF
--- a/fedimint-cli/Cargo.toml
+++ b/fedimint-cli/Cargo.toml
@@ -31,9 +31,9 @@ bitcoin = { workspace = true }
 clap = { workspace = true }
 clap_complete = "4.5.40"
 fedimint-aead = { workspace = true }
-fedimint-api-client = { path = "../fedimint-api-client", version = "=0.6.0-alpha", default-features = false }
+fedimint-api-client = { path = "../fedimint-api-client", version = "=0.6.0-alpha" }
 fedimint-bip39 = { workspace = true }
-fedimint-client = { path = "../fedimint-client", version = "=0.6.0-alpha", default-features = false }
+fedimint-client = { path = "../fedimint-client", version = "=0.6.0-alpha" }
 fedimint-core = { workspace = true }
 fedimint-eventlog = { workspace = true }
 fedimint-ln-client = { workspace = true, features = ["cli"] }
--- a/fedimint-client/Cargo.toml
+++ b/fedimint-client/Cargo.toml
@@ -17,7 +17,6 @@ normal = ["aquamarine"]
 rustc-args = ["--cfg", "tokio_unstable"]
 
 [features]
-default = ["tor"]
 tor = ["fedimint-api-client/tor"]
 
 [lib]
@@ -31,7 +30,7 @@ async-stream = { workspace = true }
 async-trait = { workspace = true }
 bitcoin = { workspace = true }
 fedimint-aead = { workspace = true }
-fedimint-api-client = { path = "../fedimint-api-client", version = "=0.6.0-alpha", default-features = false }
+fedimint-api-client = { path = "../fedimint-api-client", version = "=0.6.0-alpha" }
 fedimint-core = { workspace = true }
 fedimint-derive-secret = { workspace = true }
 fedimint-eventlog = { workspace = true }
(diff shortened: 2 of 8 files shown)


fedimint-cli/Cargo.toml
```diff
@@ -13,7 +13,6 @@ description = "Fedimint client CLI interface"
 rustc-args = ["--cfg", "tokio_unstable"]
 
 [features]
-default = ["tor"]
```
dpc, line 16: No one should need to import `fedimint-cli` as a dependency, so this one can enable tor by default and we probably want to ship `fedimint-cli` with tor support.

gateway/cli/Cargo.toml
```diff
@@ -13,7 +13,6 @@ description = "CLI tool to control lightning gateway"
 rustc-args = ["--cfg", "tokio_unstable"]
 
 [features]
-default = ["tor"]
```
dpc, line 16: Same here, I guess? Does anyone need to import `gateway-cli` as a cargo dep? I don't think so, but might be wrong.

fedimint-cli/Cargo.toml
m1sterc001guy: Fixed

gateway/cli/Cargo.toml
m1sterc001guy: Fixed

fedimint-client/Cargo.toml
```diff
@@ -17,7 +17,6 @@ normal = ["aquamarine"]
 rustc-args = ["--cfg", "tokio_unstable"]
 
 [features]
-default = ["tor"]
```
dpc, line 20: Probably some other crates (like `fedimint-cli`) will need to enable it when their feature is set. This was probably incorrectly handled in the first place.

dpc: The features need to cascade from outter-most to the innermost.

m1sterc001guy: Setting it as defaiult in `fedimint-cli` and not default here is the correct setup though right?

dpc: Yes. But fedimint-cli needs to have:

```
[features]
tor = [ "fedimint-cli/tor" ]
```

too, so when it's enabled on `fedimint-cli`, it will also get enabled on `fedimint-client`.

fedimint-cli/Cargo.toml
```diff
@@ -31,9 +31,9 @@ bitcoin = { workspace = true }
 clap = { workspace = true }
 clap_complete = "4.5.40"
 fedimint-aead = { workspace = true }
-fedimint-api-client = { path = "../fedimint-api-client", version = "=0.6.0-alpha", default-features = false }
+fedimint-api-client = { path = "../fedimint-api-client", version = "=0.6.0-alpha", features = ["tor"] }
```
dpc, line 34: Not needed. The `tor = ["fedimint-client/tor", "fedimint-api-client/tor"]` already does it.

fedimint-cli/Cargo.toml
```diff
@@ -31,9 +31,9 @@ bitcoin = { workspace = true }
...
 fedimint-bip39 = { workspace = true }
-fedimint-client = { path = "../fedimint-client", version = "=0.6.0-alpha", default-features = false }
+fedimint-client = { path = "../fedimint-client", version = "=0.6.0-alpha", features = ["tor"] }
```
dpc, line 36: same

gateway/cli/Cargo.toml
```diff
@@ -31,7 +31,7 @@ fedimint-eventlog = { workspace = true }
 fedimint-logging = { workspace = true }
 fedimint-mint-client = { workspace = true }
 lightning-invoice = { workspace = true }
-ln-gateway = { package = "fedimint-ln-gateway", path = "../ln-gateway", version = "=0.6.0-alpha", default-features = false }
+ln-gateway = { package = "fedimint-ln-gateway", path = "../ln-gateway", version = "=0.6.0-alpha", features = ["tor"] }
```
dpc, line 34: same

m1sterc001guy:
One consequence of having tor on by default in the cli crates is that devimint relies on the cli feature, so devimint needs to have the tor feature enabled in `fedimint-testing`. This is no worse than before, but any downstream crate that wants to depend on devimint will need to link against sqlite

dpc (approved):

Issue #6671, opened 2025-01-09
Deprecate concept of output outcome


dpc:

dpc:
@joschisan @elsirion Sounds OK?

joschisan:
Exactly my thoughts, are already working on the transition for lnv2. We should make it optional return an output outcome as well.

elsirion:
TODO: check that we don't depend on `OutputOutcome`s anywhere anymore, then leave the issue open (or open a new one) for eventual removal of the concept.

PR #6672, opened 2025-01-09
chore: deprecate concept of an "output outcome"

### LINKED ISSUE #6670: `output_status` documentation seems wrong
This is the current documentation for `output_status`

```rust
    /// Retrieve the current status of the output. Depending on the module this
    /// might contain data needed by the client to access funds or give an
    /// estimate of when funds will be available. Returns `None` if the
    /// output is unknown, **NOT** if it is just not ready yet.
```

This suggests that `None` shouldn't be used if the output status is not ready yet. However the current implementations all seem to rely on treating `None` as "not ready yet", and on the client side there is corresponding `retry` logic (I am guessing the API call fails when `output_status` returns `None`)
Screenshot 2025-01-09 at 1 43 16 PM
Screenshot 2025-01-09 at 1 43 25 PM
Screenshot 2025-01-09 at 1 43 34 PM

Either the documentation should be updated to align with current usages, or the usages should be updated to not rely on using `None` in this way, but instead having a DB key change wait or similar.

dpc:
See https://github.com/fedimint/fedimint/issues/6671

Also fix #6670


elsirion (approved):
To be able to actually remove it eventually we'd need to create API endpoints for the output types that still use it (iirc withdraws use it to communicate the transaction ID, e-cash *may* still use it to communicate blind signature shares, …) and then switch the client over to these using API versioning.

PR #6674, opened 2025-01-10
chore: ban HashMap on server side

maan2003:
use clippy disallowed types to ban hashmap and hashset from server and modules.

Looks like we are using it in many places

### DIFF
--- a/fedimint-server/clippy.toml
+++ b/fedimint-server/clippy.toml
@@ -1,0 +1,2 @@
+# Note: this file symlinked from multiple crates
+disallowed-types = ["std::collections::HashMap", "std::collections::HashSet"]
--- a/modules/fedimint-dummy-server/clippy.toml
+++ b/modules/fedimint-dummy-server/clippy.toml
@@ -1,0 +1,1 @@
+../../fedimint-server/clippy.toml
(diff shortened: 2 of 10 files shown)


maan2003:
👀 why is CI passing, I see warnings locally

dpc:
OK, now that it's working, we need to fix the existing uses of HashMap. @joschisan is doing some refactoring in some of the affected parts.

dpc:
Just got some big tasks assigned elsewhere, and this one can wait, so no commitment to rush it.

dpc:
*Comment by Tau/GPT-5*

I checked this against current `master` (`08e5d064`). This is still relevant and does not appear to have been done another way: there are no server-side `clippy.toml` files on `master`, and a few `std::collections::HashMap` uses remain in server crates (`fedimint-server/src/config/mod.rs`, `fedimint-server/src/consensus/api.rs`, `modules/fedimint-mint-server/src/lib.rs`, `modules/fedimint-wallet-server/src/lib.rs`).

The PR is very stale and currently conflicts, so it would need a refresh before merge: rebase, keep the Nix source-filter update, add the symlink for newer server crates such as `fedimint-mintv2-server`/`fedimint-walletv2-server` as appropriate, and either convert the remaining `HashMap` uses to deterministic maps or explicitly justify any exceptions. I’d leave it open for now since the goal still seems valid and the remaining work looks bounded.

dpc:
Redoing in https://github.com/fedimint/fedimint/pull/8601

PR #6677, opened 2025-01-10
chore(docs): generalize "setting up Fedimint" docs

dpc:
Currently from the README we link straight into very long and detailed and hands-on guide.

I think we need a single page that does some prep-talk, overviews and then points the reader at more detailed and specific guides, in which the existing guide will be just one of the items.


douglaz:
@dpc @elsirion images are broken now on https://github.com/fedimint/fedimint/blob/master/docs/deploying/docker-mutiny.md

dpc:
> @dpc @elsirion images are broken now on https://github.com/fedimint/fedimint/blob/master/docs/deploying/docker-mutiny.md

Fix in #6697

PR #6678, opened 2025-01-10
refactor: Move Lightning logic to `fedimint-lightning`

m1sterc001guy:
The `ln_gateway` crate is too big. It makes it hard to re-use and encourages bad practices by entangling dependencies between logically separate components. This is the first PR to try and refactor code out of `ln_gateway`

- [X] Move Lightning code into `fedimint-lightning` (this PR)
- [ ] Move `GatewayClientModuleV2` into separate crate
- [ ] Move `GatewayClientModule` into separate crate
- [ ] Create Gatewayd crate and gateway-server crate and attach modules to gateway-server inside of gatewayd (similar to fedimintd)

This PR accomplishes the first goal by moving the lightning code into its own crate. There is now a trait `LightningV2Manager` (open to suggestions on names) that dictates how other crates can interact with the gateway-server. When extracting `GatewayClientModuleV2`, additional functions can be added here to abstract away the dependency on the gateway (LNv1 will likely get its own trait, for clean separation).


Cargo.toml
```diff
@@ -29,7 +29,7 @@ members = [
     "fedimint-wasm-tests",
     "fedimintd",
     "fuzz",
-    "gateway/cli",
+    "gateway/cli", "gateway/fedimint-lightning",
```
dpc, line 32: on a new line

m1sterc001guy: ugh, `cargo new`.....

elsirion:
Good candidate for tomorrow's review club

dpc (commented):
LGTM, but I feel like someone more familiar with LN should scan through.

joschisan (approved):

PR #6684, opened 2025-01-11
Cleanup connector

joschisan:


joschisan:
An Iroh connection is more flexibel than a single bidirectional stream and it will map better on the custom trait.

Issue #6675, opened 2025-01-10
Remove legacy deposit state machine logic


bradleystachurski:
When we introduced the peg-in monitor, we kept the legacy deposit state machine to ensure existing deposit state machines would have a chance to complete. Now that the peg-in monitor has been running for several major versions, we should remove the legacy deposit state machine logic.

https://github.com/fedimint/fedimint/blob/0ec7044dfafc0f1e5d4361a5ead364f9fb9117c7/modules/fedimint-wallet-client/src/lib.rs#L15-L17

PR #6687, opened 2025-01-13
fix(wallet): verify utxo is known by server for v1 inputs

bradleystachurski:
The following edge case exists:
- Client sends funds to peg-in address
- Prior to claiming the peg-in, federation reaches consensus for 2.2
- Client attempts to claim peg-in, constructing V1 input since federation consensus is 2.2
- Federation rejects since UTXO is unknown, due to not persisting UTXO until 2.2 consensus is reached

This PR handles this by adding a new endpoint to check if the federation knows of the UTXO prior to submitting a V1 input. If the federation doesn't know of the UTXO, the client will fallback to V0 inputs.

Please note:
- The client no longer checks the consensus version since the federation will only know of the UTXO if the consensus version >= 2.2, thus the check is redundant
- This PR does not modify `modules/fedimint-wallet-client/src/deposit.rs` since the deposit flow is handled by the peg-in monitor since v0.4.0
  - To avoid further confusion, we should remove the legacy deposit logic https://github.com/fedimint/fedimint/issues/6675

### DIFF
--- a/fedimint-api-client/Cargo.toml
+++ b/fedimint-api-client/Cargo.toml
@@ -34,6 +34,7 @@ fedimint-logging = { workspace = true }
 futures = { workspace = true }
 itertools = { workspace = true }
 jsonrpsee-core = { version = "0.24.7" }
+jsonrpsee-types = { version = "0.24.7" }
 lru = "0.12.5"
 serde = { workspace = true }
 serde_json = { workspace = true }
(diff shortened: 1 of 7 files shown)


fedimint-api-client/Cargo.toml
```diff
@@ -34,6 +34,7 @@ fedimint-logging = { workspace = true }
 futures = { workspace = true }
 itertools = { workspace = true }
 jsonrpsee-core = { version = "0.24.7" }
+jsonrpsee-types = { version = "0.24.7" }
```
bradleystachurski, line 37: Needed to introduce so we can use `METHOD_NOT_FOUND_CODE` to check for the error type instead of using string matching.

modules/fedimint-wallet-server/src/lib.rs
```diff
@@ -1265,11 +1265,13 @@ impl Wallet {
                     }
 
                     for (vout, tx_out) in transaction.output.iter().enumerate() {
-                        if if self.cfg.consensus.peer_peg_in_keys.len() > 1 {
+                        let should_track_utxo = if self.cfg.consensus.peer_peg_in_keys.len() > 1 {
                             tx_out.script_pubkey.is_p2wsh()
                         } else {
                             tx_out.script_pubkey.is_p2wpkh()
-                        } {
```
elsirion, line 1272: That's the n(e|a)stiest if condition I've ever seen :laughing:

elsirion (approved):
Great fix and some beautiful cleanups along the way! :heart_eyes:

PR #6689, opened 2025-01-13
chore: expose get_prev_payment_result

maan2003:
we need to know "have we paid this invoice"


modules/fedimint-ln-client/src/lib.rs
```diff
@@ -632,7 +632,7 @@ impl LightningClientModule {
         }
     }
 
-    async fn get_prev_payment_result(
+    pub async fn get_prev_payment_result(
```
dpc, line 635: FWIW, making it public is a great opportunity to add a docstring.

PR #6690, opened 2025-01-13
[backport] fix: prefix isolated databases with `ClientDatabase` prefix

m1sterc001guy:
After the dev call (Jan 13), it occurred to me that upgrading from v0.4 -> v0.5 is not affected by https://github.com/fedimint/fedimint/issues/6579

This is because v0.4 handles client databases in the gateway differently. It uses a separate DB per client, which is not affected by the above issue. We have, and will always, still support the "legacy" way of opening databases.

Gateway operators can choose to migrate their joined clients if they want to by leaving and re-joining the federation. This behavior is tested in `mnemonic_upgrade_test`.

However, because v0.4 gateways are not affected, as long as v0.5 contains the fix, we do not need any database migrations or code that prevents database migrations from being run. 

v0.6 will include the database migrations that migrates previously affects v0.5 gateways.

### DIFF
--- a/gateway/ln-gateway/src/client.rs
+++ b/gateway/ln-gateway/src/client.rs
@@ -12,10 +12,9 @@ use fedimint_client::{Client, ClientBuilder};
 use fedimint_core::config::FederationId;
 use fedimint_core::core::ModuleKind;
 use fedimint_core::db::{Database, IDatabaseTransactionOpsCoreTyped};
-use fedimint_core::encoding::Encodable;
 use fedimint_core::module::registry::ModuleDecoderRegistry;
 
-use crate::db::FederationConfig;
+use crate::db::{FederationConfig, GatewayDbExt};
 use crate::error::AdminGatewayError;
 use crate::gateway_module_v2::GatewayClientInitV2;
 use crate::state_machine::GatewayClientInit;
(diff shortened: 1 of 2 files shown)


dpc:
Should we prevent joining federations starting with `0x04` by hardcoding it on server-side and return some error?

If so there's a case when someone wants to migrate on v0.4 from old style to new style, and discover they can not join.

m1sterc001guy:
> If so there's a case when someone wants to migrate on v0.4 from old style to new style, and discover they can not join.

This can only happen if the gateway is already on v0.5 and has joined an affected federation. But yes, this should be added for v0.5 -> v0.5.x

m1sterc001guy:
> Should we prevent joining federations starting with 0x04 by hardcoding it on server-side and return some error?

I don't see much value doing it that way. With this backport, gateways running v0.5.1 will be able to join federations starting with 0x04, since it has the fix.

The only affected gateways would be running v0.5 and have already joined an affected federation. The way to get these working again is to backport the database migrations, which we don't want to do. So v0.5 gateways should be advised to upgrade to v0.6 when it releases.

 - v0.4 gateways can safely upgrade to v0.5.1 or v0.6 (when released)
 - v0.5 gateways can upgrade to v0.5.1 or v0.6 (when released). Affected v0.5 gateways are advised to upgrade to v0.6.
 
 > If so there's a case when someone wants to migrate on v0.4 from old style to new style, and discover they can not join.

This is not possible, v0.4 does not support the new style of client databases.

I think that's a reasonable plan? Let me know if I'm missing something.

m1sterc001guy:
Sorry, think I got ahead of myself. The fix does not work without the database migration. Closing and will figure out a different solution

PR #6692, opened 2025-01-13
fix: backup size padding being 2B off sometimes

### LINKED ISSUE #6686: `ClientBackup` has weird sizes (padding) sometimes
Off by 2.

dpc:
Fix #6686

### DIFF
--- a/fedimint-client/src/backup.rs
+++ b/fedimint-client/src/backup.rs
@@ -1,6 +1,6 @@
 use std::cmp::Reverse;
 use std::collections::{BTreeMap, BTreeSet};
-use std::io::{Error, Read, Write};
+use std::io::{Cursor, Error, Read, Write};
 
 use anyhow::{bail, ensure, Context, Result};
 use bitcoin::secp256k1::{Keypair, PublicKey, Secp256k1, SignOnly};
@@ -102,7 +102,11 @@ impl ClientBackup {
 
     /// Encrypt with a key and turn into [`EncryptedClientBackup`]
     pub fn encrypt_to(&self, key: &fedimint_aead::LessSafeKey) -> Result<EncryptedClientBackup> {
-        let encoded = Encodable::consensus_encode_to_vec(self);
+        let mut encoded = Encodable::consensus_encode_to_vec(self);
+
+        let alignment_size = Self::get_alignment_size(encoded.len());
+        let padding_size = alignment_size - encoded.len();
+        encoded.write_all(&vec![0u8; padding_size])?;
 
         let encrypted = fedimint_aead::encrypt(encoded, key)?;
         Ok(EncryptedClientBackup(encrypted))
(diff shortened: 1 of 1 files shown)


elsirion:
CI doesn't like :grimacing:

fedimint-client/src/backup.rs
```diff
@@ -175,14 +179,18 @@ impl Encodable for ClientBackup {
...
+        // Older version of the client used a custom zero-filled vec to
+        // pad `ClientBackup` to alignment-size here. This has a problem
+        // that the size of encoding of the padding can have different sizes
+        // (due to var-ints).
+        //
+        // Conceptually serialization is also a the wrong place to do padding
+        // anyway, so we moved padding to encrypt/decryption. But for abundance of
+        // caution, we'll keep the old padding around, and always fill it
+        // with an empty Vec.
+        len += Vec::<u8>::new().consensus_encode(writer)?;
```
elsirion, line 193: We'd need that if FM supported downgrades, but since we don't it's just abundance of caution.

fedimint-client/src/backup.rs
```diff
@@ -102,7 +102,11 @@ impl ClientBackup {
 
     /// Encrypt with a key and turn into [`EncryptedClientBackup`]
     pub fn encrypt_to(&self, key: &fedimint_aead::LessSafeKey) -> Result<EncryptedClientBackup> {
-        let encoded = Encodable::consensus_encode_to_vec(self);
+        let mut encoded = Encodable::consensus_encode_to_vec(self);
+
+        let alignment_size = Self::get_alignment_size(encoded.len());
+        let padding_size = alignment_size - encoded.len();
+        encoded.write_all(&vec![0u8; padding_size])?;
```
elsirion, line 109: I think this would break old clients, but that's ok as we don't support downgrades anyway.

elsirion (approved):

fedimint-client/src/backup.rs
dpc: Well, someone might want to recover from backup made by the older version of the app, and then they would have to do full recovery.

And an empty vec is just one `0` byte.

dpc: Also - this backup is temporarily stored during init of a client with recovery in the database. Very unlikely that someone would upgrade during that exact window, but ... abundance of caution.

elsirion: Oh, I thought partial decode would fix that? But maybe this is wrapped again and partial decode on the outer layer wouldn't help?

fedimint-client/src/backup.rs
dpc: Yes, this would make the backup from the new client not work on an old one. Unsupported, unlikely and full recovery would still work.

fedimint-client/src/backup.rs
dpc: Old clients expect the vector there (see corresponding `consensus_decode`)

dpc: It was a stupid way to do padding, but oh well.

Issue #6679, opened 2025-01-10
LNv1 and LNv2 querying block count too often


m1sterc001guy:
I was recently debugging a federation that had stalled, and noticed that Blockstream's esplora API was throttling (error code 429) and failing (error code 403) the block height requests from the federation. My suspicion is that the guardians are querying the block height too often, leading to the IP being banned on the server side.

Both LNv2 and LNv1 query the block height once per `consensus_proposal` invocation which, according to the documentation, can be executed every few seconds:

```
async fn consensus_proposal(
        &self,
        _dbtx: &mut DatabaseTransaction<'_>,
    ) -> Vec<LightningConsensusItem> {
        let mut items = vec![LightningConsensusItem::UnixTimeVote(
            duration_since_epoch().as_secs(),
        )];

        if let Ok(block_count) = self.btc_rpc.get_block_count().await {
            items.push(LightningConsensusItem::BlockCountVote(block_count));
        }

        items
    }
```

The wallet module handles this correctly, spawning an infinite loop that queries the block height every 30 seconds and pushes the queried height to `consensus_proposal`. 

We should refactor LNv1 and LNv2 to use the same mechanism. Maybe its possible to even re-use the code from the wallet module?

PR #6693, opened 2025-01-13
refactor: lnv1 and lnv2 query block count every 30 seconds

m1sterc001guy:
Fixes: https://github.com/fedimint/fedimint/issues/6679

Re-uses the wallet module's approach to querying block count. Couldn't directly re-use the code since the wallet module also needs the fee rate, which the lightning modules don't need.


m1sterc001guy:
Second commit unifies the background tasks across LNv1, LNv2, and the wallet module. Its a little bit of overkill since the lightning modules don't need the feerate, but there is less repeated code. First commit only unifies it across the lightning modules.

fedimint-bitcoind/Cargo.toml
```diff
@@ -26,6 +26,11 @@ bitcoincore-rpc = { workspace = true, optional = true }
 esplora-client = { workspace = true, optional = true }
 fedimint-core = { workspace = true }
 fedimint-logging = { workspace = true }
+futures = { workspace = true }
 hex = { workspace = true }
+jaq-core = { workspace = true }
```
dpc, line 31: I wonder if we should move it to some `fedimint-bitcoind-feerate`, so that other `fedimint-bitcoind` depedents don't need this extra stuff now, but good enough I guess.

dpc (dismissed):

fedimint-bitcoind/src/lib.rs
```diff
@@ -183,3 +188,50 @@ dyn_newtype_define! {
...
+        let mut desired_interval = tokio::time::interval(if is_running_in_test_env() {
+            // In devimint, the setup is blocked by detecting block height changes,
+            // and polling more often is not an issue.
+            debug!(target: LOG_BITCOIND, "Running in devimint, using fast node polling");
+            Duration::from_millis(100)
+        } else {
+            Duration::from_secs(30)
+        });
+
+        task_group.spawn_cancellable("LNv2 module: background update", async move {
```
elsirion: Documentation/task description: Should not be limited to LNv2 if it lives in fedimint-bitcoin. 

Don't we also use something similar in the wallet module? Actually, we could have a `Mutex<Option<watch::Receiver<_>>>` field in `DynBitcoindRpc` that returns an existing handle and spawns a new one if none exists.

That way we could even share the polling between LNv2, LNv1 and the wallet.

fedimint-bitcoind/src/lib.rs
```diff
@@ -205,33 +227,85 @@ impl DynBitcoindRpc {
...
+                        if !is_running_in_test_env() {
+                            warn!(target: LOG_BITCOIND, "Unable to calculate any fee rate");
+                        }
+                    }
+                };
+
+                loop {
+                    let start = now();
+                    update_block_count().await;
+                    update_fee_rate().await;
```
elsirion, line 285: Do these belong together? Could also run independently.

fedimint-bitcoind/src/lib.rs
dpc: This PR does move the code from wallet module to bitcoind crate, as a sort of "common".

m1sterc001guy: Will change the task description and add some documentation.

As far as sharing the polling, that is possible to do but it would require each module to share the same `DynBitcoinRpc` struct. Currently each module creates their own based on the local config. I'm not sure if we want the client (as opposed to the modules) constructing the `DynBitcoinRpc` since it ties the Client a bit closer to Bitcoin. If we want to do that refactor I think it should be done in a follow up.

fedimint-bitcoind/src/lib.rs
m1sterc001guy: Separated into different tasks

fedimint-bitcoind/src/lib.rs
dpc: If we see that over and over many modules do the same thing, then we can think about merging it. Otherwise on an active federation there are just two modules that care.

dpc (approved):

PR #6694, opened 2025-01-14
chore(ci): switch linkchecking to lychee

dpc:
The output of existing link checker is super verbose and annoys me every time there's an issue, while it is no longer maintained:

https://github.com/gaurav-nelson/github-action-markdown-link-check/issues/193

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,4 @@
+http://your.ip.add.ress:3000/
+http://your.ip.add.ress:3001/
+# some fake links in rustdoc-index.md
+file://.*/\b\w+(_\w+)*\b/index.html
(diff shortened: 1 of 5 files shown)


.lycheeignore
```diff
@@ -0,0 +1,4 @@
+http://your.ip.add.ress:3000/
+http://your.ip.add.ress:3001/
```
bradleystachurski, line 2: I would have expected to also need `http://your.ip.add.ress.198:3003`.

Also, should we cleanup usages of 

```
<!-- markdown-link-check-disable -->
...
<!-- markdown-link-check-enable -->
```

bradleystachurski (approved):

.lycheeignore
dpc: No idea why it isn't shown. Maybe it's an invalid URL?

bradleystachurski:
Dang, looks like a new flake. Captured in https://github.com/fedimint/fedimint/issues/6696

PR #6695, opened 2025-01-14
Remove Parallelisation of DKG and some cleanups

joschisan:
This pr is part of a larger effort to simplify the p2p networking situation.

Removing the parallelization of the DKG across modules and across keys allows us to completely remove the multiplexing wrapper around the p2p networking layer and allows for further simplifications in  the server. As a downside the DKG is about 45x slower, primarily driven by the mint keys. Generating one keyset should take about 4x one way latency between peers hence about 200ms - 400ms or 4.5s - 9s for the entire DKG in a real deployment while it remains negligible in testing. 

We now explicitly rely on the assumption that the networking layer is reliable during the dkg, meaning all messages are delivered in order. The current design appears to try to work on an unreliable layer as well but fails to do so anyway as it's fundamentally impossible so we might as well exploit the reliability assumption for a cleaner design.

Before the DKG we wait for all p2p connections to be open and after the DKG we now await our outgoing message queues to drain in order for our fellow peers to complete the DKG as well; this should make the DKG more reliable especially with a high number of peers.


fedimint-core/src/config.rs
```diff
@@ -882,26 +880,6 @@ pub enum DkgPeerMsg {
...
-    #[error("Running DKG failed due to {0}")]
-    Failed(#[from] anyhow::Error),
-    #[error("The module was not found {0}")]
-    ModuleNotFound(ModuleKind),
-    #[error("Params for modules were not found {0:?}")]
-    ParamsNotFound(BTreeSet<ModuleKind>),
-    #[error("Failed to decode encodable message {0:?}")]
-    ModuleDecodeError(DecodeError),
-}
-
```
elsirion: My first reaction was "BUUUUT internationalization", then I realized that none of these errors should happen, even with invalid user input.

fedimint-server/src/config/distributedgen.rs
```diff
@@ -210,30 +229,15 @@ impl<G: DkgGroup> Dkg<G> {
...
-    /// Get a second generator by hashing the first one to the curve
-    fn gen_h(&self) -> G {
-        let mut hash_engine = sha3::Sha3_256::new();
-
-        hash_engine.update(self.generator.clone().to_bytes().as_ref());
-
-        G::random(&mut ChaChaRng::from_seed(hash_engine.finalize().into()))
-    }
+fn gen_h<G: DkgGroup>() -> G {
+    G::random(&mut ChaChaRng::from_seed([42; 32]))
```
elsirion, line 240: Can we just keep hashing, but use sha256 to get rid of that pretty random sha3 dep? Hashing the main generator is a very standard construction: https://bitcoin.stackexchange.com/a/116733

fedimint-server/src/config/mod.rs
```diff
@@ -490,50 +489,6 @@ impl ServerConfig {
...
-                        error!(target: LOG_NET_PEER_DKG, %peer_id, ?msg, "Received incorrect message after dkg was supposed to be finished. Probably dkg multiplexing bug.");
-                    },
-                    Err(Cancelled) => {/* ignore shutdown for time being, we'll timeout soon anyway */},
-                }
-            }
-        })
-        .await
-        {
-            error!(target: LOG_NET_PEER_DKG, "Timeout waiting for dkg completion confirmation from other peers");
-        };
```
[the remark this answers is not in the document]
elsirion: We talked about it: just need to wait for all messages to be sent as done in a later PR.

fedimint-server/src/config/distributedgen.rs
```diff
@@ -303,49 +299,37 @@ where
...
         DkgMessage<G>: ISupportedDkgMessage,
     {
-        // Use tokio channel to await on `recv` or we might block
-        let (send, mut receive) = tokio::sync::mpsc::channel(10_000);
+        // Collect every key, returning an error if any fails
+        let mut results: HashMap<T, DkgKeys<G>> = HashMap::new();
+
+        assert!(self.dkg_config.len() == 1);
 
         // For every `key` we run DKG in a new tokio task
```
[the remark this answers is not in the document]
elsirion: Removed later

elsirion (approved):

fedimint-server/src/config/mod.rs
```diff
@@ -446,119 +456,88 @@ impl ServerConfig {
...
-        conn.send(
-            peers,
-            (MODULE_INSTANCE_ID_GLOBAL, dkg_done.clone()),
-            DkgPeerMsg::Done,
-        )
-        .await?;
+        // We need to wait for out outgoing message queues to be fully transmitted
+        // before we move on in order for our peers to be able to complete the dkg.
+
+        connections.await_empty_outgoing_message_queues().await;
```
dpc, line 525: The fact that the outgoing message queue is empty does not guarantee that it was even sent. The peer thread had pick it up, but did it send it yet? The task group (and thus peer-connection tasks) is killed immediately afterwards, so it is actually quite possible that it didn't. Without some ability to "flush" the underlying thread and get an ack that it completed it, there's no way of knowing.

There was an exact same problem here before, that's why I added the loop below - to wait for acks from all peers, and fallback to some timeout. This problem was actually quite common because it's multiplied by each peer. It might be slightly better now because that 200ms polling timeout, but fundamentally this code is till buggy and can fail DKG for no fault of networking.



@elsirion @joschisan

PR #6702, opened 2025-01-14
Pedersen DKG without generics

joschisan:
We needed to remove the generics from the Dkg Message types such that we can derive the consensus encoding. Therefore we modify the dkg to run both groups G1 and G2 in parallel and then after the fact discard the public keys we don’t need. Hence we remain with only on type of dkg. 

Since this makes the dkg more complicated we use the opportunity to switch to a simpler dkg, with two rounds instead of four, aka Pedersen Dkg. This dkg does not guarantee an even probability across generated aggregated public keys but iirc this is not a problem in a simple signing scheme such as ours.

Finally this allows us to prepare the switch to consensus encoding in p2p networking in the next minor version.


dpc:
image


@joschisan Please help the reviewers present and future and write some description of what are you doing. You tend to be making changes in most core and delicate parts of the system.

fedimint-server/src/config/distributedgen.rs
```diff
@@ -317,11 +191,9 @@ fn eval_poly_scalar(coefficients: &[Scalar], x: &Scalar) -> Scalar {
...
 }
 
-#[derive(Debug, Clone)]
-pub enum DkgStep<G: DkgGroup> {
-    Broadcast(DkgMessage<G>),
-    Messages(Vec<(PeerId, DkgMessage<G>)>),
-    Result((Vec<G>, Scalar)),
+enum DkgStep {
+    Messages(Vec<(PeerId, DkgMessage)>),
+    Result((Vec<(G1Projective, G2Projective)>, Scalar)),
```
elsirion, line 230: Getting the public key on both groups is a fundamental change that could impact security proofs. My hunch is it won't (indeed, for proving BLS secure you give the adversary an oracle "translating"  arbitrary elements from G1 to G2 [possibly the other way around]), but is it really worth it to make an encoding impl a bit nicer?

Most of the other changes are net benefits imo, this one doesn't seem worth it imo.

elsirion: Implementing en-/decodable generically actually isn't that much code: https://github.com/fedimint/fedimint/commit/6989126480f13a0534dfe666ae15791e018c186f

joschisan: This DKG works over any group in which the discrete logarithm is hard and G1 x G2 is itself just a group again. No changes there.

joschisan: Also I consider manual implementation of encodings horrible style.

elsirion: > Also I consider manual implementation of encodings horrible style.

Meh, if you really want me to build a derive macro that works for generics … :zany_face: 

> This DKG works over any group in which the discrete logarithm is hard and G1 x G2 is itself

That's not the problem though. I trust the DKG will work just find, but we generate keys for (G1xG2) and then only use one of the two parts. That's non-obvious why that should be safe. I can likely come up with a scheme where it isn't.

dpc:
This looks like an improvement. I have no opinions (or rather knowledge) on the crypto parts.

elsirion (approved):
Since my concerns are mostly of theoretical nature and don't apply to our particular schemes let's get this merged (I don't have time to impl my competing proposal, so it's a do-ocracy). We can always change DKG without breaking compatibility.

